PR #1533, opened 2018-01-10
Fix #1521: Issue with Mutator position when nesting blocks

### LINKED ISSUE #1521: location of mutator makes it difficult to use.
### Problem statement
Mutators choose bad place when the blocks heap high.

### Expected Behavior

When I click the mutator button , the mutator display at a place easy to use.
image

### Actual Behavior
if  there are many blocks , the mutator display at the button of the blocks.
image

### Steps to Reproduce

drag and drop a  if-else block , and add much else-if sub block in mutator.

* Desktop Chrome

panreel:
## The basics

- [x] I branched from develop

### Resolves

[Issue #1521](https://github.com/google/blockly/issues/1521)

### Proposed Changes

Solves the issue related to wrong displaying of Mutator with multiple nested blocks: In the old version the bubble position (when sliding below the mutator) would have been calculated from the height of the bounding box surrounding the anchor item. 

`var bBox = (this.shape_).getBBox(); relativeTop = bBox.height;`

If multiple blocks were added, the height of such bounding box would have increased at each addition pushing the bubble position downward, and making the bubble unusable.

The proposed change doesn't calculate the bubble position from the bounding box height of the anchor item but from the viewHeight parameter of the workspace (which is not supposed to change when adding items to a mutator). 

`relativeTop = (metrics.viewHeight*0.1);`

### Reason for Changes

The reason to change is to fix the following wrong behavior:

#### Before

manyif-before

#### After

manyif

manyif-bottomup

### Test Coverage

ut

Tested on:
Desktop Chrome
Desktop Firefox

### DIFF
--- a/core/bubble.js
+++ b/core/bubble.js
@@ -382,7 +382,7 @@ Blockly.Bubble.prototype.setAnchorLocation = function(xy) {
  */
 Blockly.Bubble.prototype.layoutBubble_ = function() {
   // Compute the preferred bubble location.
-  var relativeLeft = -this.width_ / 4;
+  var relativeLeft = -this.width_ / 8;
   var relativeTop = -this.height_ - Blockly.BlockSvg.MIN_BLOCK_Y;
   // Prevent the bubble from being off-screen.
   var metrics = this.workspace_.getMetrics();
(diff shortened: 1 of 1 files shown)


panreel:
Test for CLA :)

rachel-fenichel:
I don't think this is the right fix.  The whole mutator is now visible, but the block is occluded and you can't see the results of your changes.  We would rather have the main body of the block visible while it's being edited.

Sorry for the delay getting back to you.

panreel:
@rachel-fenichel, looks like just a matter of pushing the mutator out of the block area while editing. 
I'll give it a try.

PR #1546, opened 2018-01-18
Fix #1499 and add some validation

### LINKED ISSUE #1499: Add default type option to variable fields
Variable fields have a default name option in JSON but no default type.  As a result, an unset variable field creates a variable with the default name but the `''` type.  This can be a problem if the field doesn't accept the `''` type.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

#1499

### Proposed Changes

Add a `defaultTypes` option to the JSON for a FieldVariable definition.  Parse it correctly, and make sure that the default type and the allowable types are compatible.

### Reason for Changes

There are cases when variable fields have to create new variables to be valid.  The type of the variable needs to match the intended type of the field.

### Test Coverage

JSUnit tests added to field_variable_test.js.

### Additional Information

@kchadha FYI

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -238,27 +238,30 @@ Blockly.FieldVariable.prototype.setTypes_ = function(opt_variableTypes,
     opt_defaultType) {
   // If you expected that the default type would be the same as the only entry
   // in the variable types array, tell the Blockly team by commenting on #1499.
-  this.defaultType_ = opt_defaultType || '';
+  var defaultType = opt_defaultType || '';
   // Set the allowable variable types.  Null means all types on the workspace.
   if (opt_variableTypes == null || opt_variableTypes == undefined) {
-    this.variableTypes = null;
+    var variableTypes = null;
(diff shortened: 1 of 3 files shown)


core/field_variable.js
```diff
@@ -224,6 +224,43 @@ Blockly.FieldVariable.prototype.getVariableTypes_ = function() {
...
+  } else if (Array.isArray(opt_variableTypes)) {
+    this.variableTypes = opt_variableTypes;
+    // Make sure the default type is valid.
+    var isInArray = false;
+    for (var i = 0; i < this.variableTypes.length; i++) {
+      if (this.variableTypes[i] == this.defaultType_) {
+        isInArray = true;
+      }
+    }
+    if (!isInArray) {
```
picklesrus: Do you think it matters whether you throw the error before setting this.variableTypes?  I guess I'd expect the error to get thrown without having the side effect.

rachel-fenichel: Updated.

PR #1557, opened 2018-01-20
Generator code problem fixed

Rgtemze:
## The basics

- [x] I branched from develop

### Resolves

In code generation, if no generator code is specified for any block in particular language error occurs (printed out on console) and generated code field remains unchanged leading to confusion.

### Proposed Changes

It does not generate the code if there is any block whose implementation is not known to the particular language and warns user.

I created a block 'length_of' whose code implementation for Python is not provided.
And here is the screenshot after clicking Python.
ss4git

### Reason for Changes

Blocks without implementation creates an error and changes in the workspace do not apply anymore.

### Test Coverage

Tested on:
* Desktop Chrome

* PC:
  * Device: Asus
  * OS: W10
  * Browser Chrome
  * Version 63

### DIFF
--- a/core/generator.js
+++ b/core/generator.js
@@ -106,10 +106,9 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       line = line[0];
     }
 
-    if(line == undefined){
+    if (line == undefined) {
       blocksNotGenerated.push(block);
-    }
-    else if (line) {
+    } else if (line) {
       if (block.outputConnection && this.scrubNakedValue) {
         // This block is a naked value.  Ask the language's code generator if
         // it wants to append a semicolon, or something.
@@ -118,7 +117,7 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       code.push(line);
     }
   }
-  if(blocksNotGenerated.length == 0){
+  if (blocksNotGenerated.length == 0) {
     code = code.join('\n');  // Blank line between each section.
     code = this.finish(code);
     // Final scrubbing of whitespace.
@@ -126,12 +125,12 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
     code = code.replace(/\n\s+$/, '\n');
     code = code.replace(/[ \t]+\n/g, '\n');
     return code;
-  }
-  else{
-    var textToBeLogged = 'The generator code for the following blocks not specified for ' + this.name_ + ': \n';
+  } else {
+    var textToBeLogged = 'The generator code for the following '
+    + 'blocks not specified for ' + this.name_ + ': \n';
     
     //Blocks not specified traversed
-    for(var x = 0; x < blocksNotGenerated.length; x++){
+    for (var x = 0; x < blocksNotGenerated.length; x++) {
       textToBeLogged +=  '* ' + blocksNotGenerated[x].type + '\n';
     }
     textToBeLogged += 'Fix those to be able to generate code';
@@ -197,11 +196,10 @@ Blockly.Generator.prototype.blockToCode = function(block) {
   // Prior to 24 September 2013 'this' was the only way to access the block.
   // The current prefered method of accessing the block is through the second
   // argument to func.call, which becomes the first parameter to the generator.
-  if(func == null || func == undefined){
+  if (func == null || func == undefined) {
     return undefined;
-  }else{
+   } else {
     var code = func.call(block, block);
-    
     if (goog.isArray(code)) {
       // Value blocks return tuples of code and operator order.
       goog.asserts.assert(block.outputConnection,
--- a/core/generator.js
+++ b/core/generator.js
@@ -135,10 +135,7 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       textToBeLogged +=  '* ' + blocksNotGenerated[x].type + '\n';
     }
     textToBeLogged += 'Fix those to be able to generate code';
-    alert(textToBeLogged);
-
-    //Code area remained empty
-    return '';
+    throw textToBeLogged;
   }
 };
 
--- a/core/generator.js
+++ b/core/generator.js
@@ -95,6 +95,7 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
     workspace = Blockly.getMainWorkspace();
   }
   var code = [];
+  var blocksNotGenerated = [];
   this.init(workspace);
   var blocks = workspace.getTopBlocks(true);
   for (var x = 0, block; block = blocks[x]; x++) {
@@ -104,7 +105,11 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       // Top-level blocks don't care about operator order.
       line = line[0];
     }
-    if (line) {
+
+    if(line == undefined){
+      blocksNotGenerated.push(block);
+    }
+    else if (line) {
       if (block.outputConnection && this.scrubNakedValue) {
         // This block is a naked value.  Ask the language's code generator if
         // it wants to append a semicolon, or something.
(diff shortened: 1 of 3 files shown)


NeilFraser:
I think this change would be better made in the code demo (or similar applications).  It seems unexpected that Blockly.Python.workspaceToCode would trigger an alert.  Better to throw an error, and the caller performs the alert.  Besides, what if this is a headless workspace on Node, which doesn't have alert?

Rgtemze:
Thank you for your immediate answer Neil, I think I should do what you suggest with try-catch clause and I will also try to fix the headless part. Which file would you suggest me to revise the headless?

demos/code/code.js
```diff
@@ -304,45 +306,63 @@ Code.renderContent = function() {
...
+    try{
+      var code = Blockly.JavaScript.workspaceToCode(Code.workspace);
+      content.textContent = code;
+      if (typeof PR.prettyPrintOne == 'function') {
+        code = content.textContent;
+        code = PR.prettyPrintOne(code, 'js');
+        content.innerHTML = code;
+      }
+    }catch(err){
+      alert(err);}
```
AnmAtAnm: Use Blockly.alert(..) so there is a clear override point for apps that use this demo with their own UI or in ways that cannot allow blocking the execution environment.

demos/code/code.js
```diff
@@ -304,45 +306,63 @@ Code.renderContent = function() {
...
     xmlTextarea.focus();
   } else if (content.id == 'content_javascript') {
-    var code = Blockly.JavaScript.workspaceToCode(Code.workspace);
-    content.textContent = code;
-    if (typeof PR.prettyPrintOne == 'function') {
-      code = content.textContent;
-      code = PR.prettyPrintOne(code, 'js');
-      content.innerHTML = code;
-    }
+    try{
```
AnmAtAnm: Google style: space before curly brace.
(similarly below).

demos/code/code.js
AnmAtAnm: Google Style:
 * spacing around catch: `} catch(err) {`
 * trailing curly brace on it's own line.

demos/code/code.js
```diff
@@ -93,7 +93,7 @@ Code.workspace = null;
  * Extracts a parameter from the URL.
  * If the parameter is absent default_value is returned.
  * @param {string} name The name of the parameter.
- * @param {string} defaultValue Value to return if parameter not found.
+ * @param {string} defaultValue Value to return if paramater not found.
```
AnmAtAnm: Revert this misspelling. The original is correct. Compare with [Wikipedia](https://en.wikipedia.org/wiki/Parameter_(computer_programming))

demos/code/code.js
```diff
@@ -158,8 +158,10 @@ Code.loadBlocks = function(defaultXml) {
  */
 Code.changeLanguage = function() {
   // Store the blocks for the duration of the reload.
+  // This should be skipped for the index page, which has no blocks and does
+  // not load Blockly.
   // MSIE 11 does not support sessionStorage on file:// URLs.
-  if (window.sessionStorage) {
+  if (typeof Blockly != 'undefined' && window.sessionStorage) {
```
AnmAtAnm: Revert this line. `Blockly` will never equal the string `'undefined'`.

core/generator.js
```diff
@@ -104,7 +105,11 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       // Top-level blocks don't care about operator order.
       line = line[0];
     }
-    if (line) {
+
+    if(line == undefined){
+      blocksNotGenerated.push(block);
+    }
+    else if (line) {
```
AnmAtAnm: Google style: put else on the same line as the trailing curly brace.
Here and below.

core/generator.js
```diff
@@ -104,7 +105,11 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       // Top-level blocks don't care about operator order.
       line = line[0];
     }
-    if (line) {
+
+    if(line == undefined){
```
AnmAtAnm: Google style: space after `if`, before curly brace.
Here and below.

core/generator.js
```diff
@@ -113,13 +118,25 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
...
+    code = code.join('\n');  // Blank line between each section.
+    code = this.finish(code);
+    // Final scrubbing of whitespace.
+    code = code.replace(/^\s+\n/, '');
+    code = code.replace(/\n\s+$/, '\n');
+    code = code.replace(/[ \t]+\n/g, '\n');
+    return code;
+  }
+  else{
+    var textToBeLogged = 'The generator code for the following blocks not specified for ' + this.name_ + ': \n';
```
AnmAtAnm: Google style:
 * `else` on the same line as both curly braces, with one space on each side: `} else {`
 * 80 character line length. You may need to break this into multiple concatenated strings.

core/generator.js
```diff
@@ -113,13 +118,25 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
...
+    
+    //Blocks not specified traversed
+    for(var x = 0; x < blocksNotGenerated.length; x++){
```
AnmAtAnm: Google style: Just like `if`, there should a space after `for` and before the curly brace `{`.

demos/blockfactory/block_exporter_controller.js
```diff
@@ -132,18 +132,16 @@ BlockExporterController.prototype.export = function() {
...
       // Get generator stub code in the selected language for the blocks.
       var genStubs = this.tools.getGeneratorCode(blockXmlMap,
           language);
-      // Get the correct file extension.
-      var fileType = (language == 'JavaScript') ? 'javascript' : 'plain';
+      
       // Download the file.
       FactoryUtils.createAndDownloadFile(
-          genStubs, generatorStub_filename, fileType);
+          genStubs, generatorStub_filename + '.js', 'javascript');
```
AnmAtAnm: Rebase onto the latest develop branch. These changes have already been pushed as part of the previous PR.

demos/code/code.js
AnmAtAnm: Instead of using several smaller `try` statements inside each `if` block, I think it makes sense to have one large `try` block encompassing all the `if` blocks.

core/generator.js
AnmAtAnm: What if the unsupported block was not a top block, whether a next block or a value block? It doesn't look like your code catches these cases.

core/generator.js
```diff
@@ -113,13 +118,25 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
...
+      textToBeLogged +=  '* ' + blocksNotGenerated[x].type + '\n';
+    }
+    textToBeLogged += 'Fix those to be able to generate code';
+    throw textToBeLogged;
```
AnmAtAnm: I have to discuss with others what the implication of this throw is. Specifically, what are the implications to an existing app that has not been erroring in the past and upgrades to this version of Blockly that now starts emitting errors.
@rachel-fenichel  ?

demos/code/code.js
```diff
@@ -158,8 +158,10 @@ Code.loadBlocks = function(defaultXml) {
  */
 Code.changeLanguage = function() {
   // Store the blocks for the duration of the reload.
+  // This should be skipped for the index page, which has no blocks and does
+  // not load Blockly.
```
AnmAtAnm: It is not clear to me what you are trying to say here. Which index page are you referring to?

Rgtemze: Actually, I did not change here and do not know why it is shown in my pull request.

demos/code/code.js
Rgtemze: I also have no knowledge of this change.

demos/code/code.js
Rgtemze: Thank you for all of your feedback, I will be working on it.

demos/blockfactory/block_exporter_controller.js
Rgtemze: Should I create new PR?

core/generator.js
Rgtemze: I could not understand what is implied from a 'top block'.

Rgtemze: And how can I make a distinction in the code?

core/generator.js
```diff
@@ -104,7 +105,10 @@ Blockly.Generator.prototype.workspaceToCode = function(workspace) {
       // Top-level blocks don't care about operator order.
       line = line[0];
     }
-    if (line) {
+
+    if (line == undefined) {
```
AnmAtAnm: The assumption here is that no generator would ever output an empty line. I don't think that is correct. A block might be disabled or otherwise intentionally not generating code. In some apps that have JavaScript-to-blocks functionality, there is a block that intentionally outputs an empty line to preserve the spacing in the code.

What you really want to test `this[block.type]` (or other language) undefined.
In fact, this is already tested inside `Generator.blockToCode()`, which I believe throws an error with the block name in the string (via `goog.assert`). What is doesn't do is traverse the block tree for all missing generator definitions.

Perhaps the best solution is to revert changes to the `generator.js` and add a preflight verification step inside each language `if` block of `Code.renderContent`. This should a function call that takes in the workspace and the language generator object, and in that function it should  tests each block.type in `workspace.getAllBlocks()` against the object keys of the generator. Then, if there is a missing block type, call `Blockly.alert(..)`.

Rgtemze: Okey, then, I will be working on it.

Rgtemze: How can I get  'language generator object'?

Rgtemze: Sorry for being late. My college semester has just started, I was preparing for it. What do you mean by 'language generator object' and 'object keys of the generator'? I could not understand. Thank you.

AnmAtAnm:
See #1585

PR #1562, opened 2018-01-25
Fix #1559

### LINKED ISSUE #1559: Don't let variables of different types have the same names.
Scratch allows this for backwards compatibility, so we have the infrastructure to support it in blockly.  But the default behaviour should be to require all variable names to be unique.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1559 

### Proposed Changes

Check whether any variable with the given name already exists when creating or renaming a variable.  
- Renaming to a name used for the same type is fine.  
- Renaming to a name used with a different type is a problem.  
- Creating a variable with a used name of the same type gets a warning and fails.
- Creating a variable with a used name of a different type gets a warning including the type of the conflicting variable and fails.

### Reason for Changes

The variable map has to support variables with the same name and different types internally (for Scratch), but we don't want to actually allow it in Blockly.

### Test Coverage
Tested in the playground.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -279,37 +279,37 @@ Blockly.Variables.createVariableButtonHandler = function(
   // This function needs to be named so it can be called recursively.
   var promptAndCheckWithAlert = function(defaultName) {
     Blockly.Variables.promptName(Blockly.Msg.NEW_VARIABLE_TITLE, defaultName,
-      function(text) {
-        if (text) {
-            var existing = Blockly.Variables.nameUsedWithAnyType_(text,
-                workspace);
+        function(text) {
+          if (text) {
+            var existing =
+                Blockly.Variables.nameUsedWithAnyType_(text, workspace);
             if (existing) {
               var lowerCase = text.toLowerCase();
               if (existing.type == type) {
-                var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS.replace('%1',
-                    lowerCase);
+                var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS.replace(
+                    '%1', lowerCase);
               } else {
                 var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS_FOR_ANOTHER_TYPE;
                 msg = msg.replace('%1', lowerCase).replace('%2', existing.type);
               }
               Blockly.alert(msg,
-                function() {
-                  promptAndCheckWithAlert(text);  // Recurse
-                });
+                  function() {
+                    promptAndCheckWithAlert(text);  // Recurse
+                  });
+            } else {
+              // No conflict
+              workspace.createVariable(text, type);
+              if (opt_callback) {
+                opt_callback(text);
+              }
+            }
           } else {
-            // No conflict
-            workspace.createVariable(text, type);
+            // User canceled prompt.
             if (opt_callback) {
-              opt_callback(text);
+              opt_callback(null);
             }
           }
-        } else {
-          // User canceled prompt without a value.
-          if (opt_callback) {
-            opt_callback(null);
-          }
-        }
-      });
+        });
   };
   promptAndCheckWithAlert('');
 };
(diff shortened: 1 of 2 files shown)


core/variables.js
```diff
@@ -275,19 +275,30 @@ Blockly.Variables.generateUniqueName = function(workspace) {
...
+            if (existing) {
+              var lowerCase = text.toLowerCase();
+              if (existing.type == type) {
+                var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS.replace('%1',
+                    lowerCase);
+              } else {
+                var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS_FOR_ANOTHER_TYPE;
+                msg = msg.replace('%1', lowerCase).replace('%2', existing.type);
+              }
+              Blockly.alert(msg,
```
AnmAtAnm: The above are all indented 2 spaces too many.

core/variables.js
```diff
@@ -275,19 +275,30 @@ Blockly.Variables.generateUniqueName = function(workspace) {
  */
 Blockly.Variables.createVariableButtonHandler = function(
     workspace, opt_callback, opt_type) {
+  var type = opt_type || '';
   // This function needs to be named so it can be called recursively.
   var promptAndCheckWithAlert = function(defaultName) {
     Blockly.Variables.promptName(Blockly.Msg.NEW_VARIABLE_TITLE, defaultName,
       function(text) {
```
AnmAtAnm: the function as a parameter should have four space indent.

core/variables.js
```diff
@@ -275,19 +275,30 @@ Blockly.Variables.generateUniqueName = function(workspace) {
...
         if (text) {
-          if (workspace.getVariable(text, opt_type)) {
-            Blockly.alert(Blockly.Msg.VARIABLE_ALREADY_EXISTS.replace('%1',
-                text.toLowerCase()),
+            var existing = Blockly.Variables.nameUsedWithAnyType_(text,
+                workspace);
```
AnmAtAnm: `text,` on next line, please

core/variables.js
```diff
@@ -275,19 +275,30 @@ Blockly.Variables.generateUniqueName = function(workspace) {
...
+            if (existing) {
+              var lowerCase = text.toLowerCase();
+              if (existing.type == type) {
+                var msg = Blockly.Msg.VARIABLE_ALREADY_EXISTS.replace('%1',
+                    lowerCase);
```
AnmAtAnm: I know you're repeating my prior code, but in retrospect it seems strange to repeat the variable name back to the user's input in the lowercased form.  Ideally, if `existing.name !== text` append something like `' (as "${existing.name}")'` to the variable name. That would make it clear that different name casings map to the same thing.

core/variables.js
```diff
@@ -340,9 +351,21 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
+                  promptAndCheckWithAlert(newName);  // Recurse
+                });
+            } else {
+              workspace.renameVariableById(variable.getId(), newName);
+              if (opt_callback) {
+                opt_callback(newName);
+              }
             }
           } else {
             // User canceled prompt without a value.
```
AnmAtAnm: Just `// User canceled prompt.`
(Same above, but I can't comment there.)

core/variables.js
```diff
@@ -340,9 +351,21 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
           if (newName) {
-            workspace.renameVariableById(variable.getId(), newName);
-            if (opt_callback) {
-              opt_callback(newName);
+            var existing = Blockly.Variables.nameUsedWithOtherType_(newName,
+                variable.type, workspace);
+            if (existing) {
+              var msg =
+                  Blockly.Msg.VARIABLE_ALREADY_EXISTS_FOR_ANOTHER_TYPE.replace(
+                  '%1', newName.toLowerCase()).replace('%2', existing.type);
```
AnmAtAnm: Indent parameter.

core/variables.js
```diff
@@ -340,9 +351,21 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
+              Blockly.alert(msg,
+                function() {
```
AnmAtAnm: +2 indent

core/variables.js
rachel-fenichel: Done

core/variables.js
rachel-fenichel: Done (moved the whole call to the next line)

core/variables.js
rachel-fenichel: Done

core/variables.js
rachel-fenichel: Done.

core/variables.js
rachel-fenichel: Done for both.

core/variables.js
rachel-fenichel: Created #1568

PR #1564, opened 2018-01-25
Fixed JS/PHP generators for math_number

### LINKED ISSUE #1563: Operator precedence for math_number (JS/PHP generators)
### Problem statement

**Bug Report**: The JavaScript/PHP generators for ``math_number`` currently return ``ORDER_ATOMIC`` for all numbers when they should be returning ``ORDER_UNARY_NEGATION`` for negative numbers.  In JavaScript, this can cause an issue if you e.g., extend ``Number.prototype`` by adding an ``abs`` method for absolute value and want to have a corresponding ``abs`` block (not for use in production, educational purposes only!)  For example:

image

### Expected Behavior

The generated code for the block above should be ``(-5).abs()`` so that it evaluates to ``5``.

### Actual Behavior

With ``ORDER_ATOMIC``, the above block would generate the code ``-5 .abs()``, which would incorrectly produce ``-5`` since the ``-5`` block is "ripped apart" as described in [Operator Precedence | Blockly | Google Developers](https://developers.google.com/blockly/guides/create-custom-blocks/operator-precedence).

All

### Additional Information

The ``math_number`` block generator for Dart, Python, and Lua account for negative numbers.

jschanker:
It was returning ``ORDER_ATOMIC`` for the block for any number; changed it so that it now returns ``ORDER_UNARY_NEGATION`` for negative numbers.

## The basics

- [x] I branched from develop

### Resolves #1563

### DIFF
--- a/generators/javascript/math.js
+++ b/generators/javascript/math.js
@@ -32,7 +32,9 @@ goog.require('Blockly.JavaScript');
 Blockly.JavaScript['math_number'] = function(block) {
   // Numeric value.
   var code = parseFloat(block.getFieldValue('NUM'));
-  return [code, Blockly.JavaScript.ORDER_ATOMIC];
+  var order = code >= 0 ? Blockly.JavaScript.ORDER_ATOMIC : 
+              Blockly.JavaScript.ORDER_UNARY_NEGATION;
+  return [code, order];
 };
 
 Blockly.JavaScript['math_arithmetic'] = function(block) {
--- a/generators/php/math.js
+++ b/generators/php/math.js
@@ -32,12 +32,14 @@ goog.require('Blockly.PHP');
 Blockly.PHP['math_number'] = function(block) {
   // Numeric value.
   var code = parseFloat(block.getFieldValue('NUM'));
+  var order = code >= 0 ? Blockly.PHP.ORDER_ATOMIC : 
+              Blockly.PHP.ORDER_UNARY_NEGATION;
   if (code == Infinity) {
     code = 'INF';
   } else if (code == -Infinity) {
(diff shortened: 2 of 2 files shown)


NeilFraser:
Very nice.  LGTM.
Just need the CLA signed.

jschanker:
I signed it!

AnmAtAnm:
Our previous generator tests were passing, so what specifically does it fail to capture?

rachel-fenichel:
And doe that mean that we should add a new test, that we fail without this change?

NeilFraser:
None of our existing blocks trigger this, so the tests (for our current blocks) are fine.

Dart already had this fix, since `abs` is a function on number, thus `3.abs()` and `-3.abs()` are legal.  In the latter case the implied order is `-(3.abs())`, which is not expected, so the fix was to generate `(-3).abs()` instead.

JavaScript uses `Math.abs(-3)` instead, so this is not an issue.  There are very few functions on JavaScript numbers themselves, none of which we use in Blockly's current generators.  One example would be `4.toFixed(10)` which should return the string `"4.0000000000"`.  However, this is syntactically illegal in JavaScript since the `.` is interpreted as a decimal.  So one needs to generate either `4 .toFixed(10)` or `(4).toFixed(10)`.  Assuming the former, then `-4 .toFixed(10)` does the unexpected result of `-(4 .toFixed(10))` which is `-4`, not `"-4.0000000000"`.  Personally I'd generate the latter, since it avoids the entire issue.

To summarize: not an issue with any of our generators or any generator we'd write.  But the fix is correct and consistent with other languages.

PR #1566, opened 2018-01-25
Improve support >ES5 and Node.

NeilFraser:
Adds ‘yield’ and ‘**’ to JS order list.
Build generates files (more) parsable by Node.
This PR should have no effect on existing code.

### DIFF
--- a/build.py
+++ b/build.py
@@ -104,16 +104,10 @@ class Gen_uncompressed(threading.Thread):
     f = open(self.target_filename, 'w')
     f.write(HEADER)
     f.write("""
-var isNodeJS = !!(typeof module !== 'undefined' && module.exports &&
-                  typeof window === 'undefined');
+this.IS_NODE_JS = !!(typeof module !== 'undefined' && module.exports);
 
-if (isNodeJS) {
-  var window = {};
-  require('closure-library');
-}
-
-window.BLOCKLY_DIR = (function() {
-  if (!isNodeJS) {
+this.BLOCKLY_DIR = (function(root) {
+  if (!root.IS_NODE_JS) {
     // Find name of current directory.
     var scripts = document.getElementsByTagName('script');
     var re = new RegExp('(.+)[\/]blockly_(.*)uncompressed\.js$');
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,12 +42,9 @@
...
   },
   "dependencies": {
+    "google-closure-library": "^20171203.0.0",
     "install": "^0.8.8",
     "npm": "^4.4.4",
-    "closure-library": "^1.43629075.2",
     "webdriverio": "^4.6.2"
-  },
-  "devDependencies": {
-    "eslint": "2.9.0"
```
NeilFraser: Appears to be a duplicate 'devDependencies'.  Merging with the one above.

package.json
```diff
@@ -41,12 +42,9 @@
     "unused": true
   },
   "dependencies": {
+    "google-closure-library": "^20171203.0.0",
     "install": "^0.8.8",
     "npm": "^4.4.4",
-    "closure-library": "^1.43629075.2",
```
NeilFraser: This is an abandoned unofficial 5 year old copy of Closure Library, not the real one.
https://github.com/gregrperkins/closure-library

generators/javascript.js
```diff
@@ -84,8 +84,9 @@ Blockly.JavaScript.ORDER_TYPEOF = 4.5;         // typeof
 Blockly.JavaScript.ORDER_VOID = 4.6;           // void
 Blockly.JavaScript.ORDER_DELETE = 4.7;         // delete
 Blockly.JavaScript.ORDER_AWAIT = 4.8;          // await
-Blockly.JavaScript.ORDER_DIVISION = 5.1;       // /
-Blockly.JavaScript.ORDER_MULTIPLICATION = 5.2; // *
```
NeilFraser: Mozilla now has / and * flipped on their page.  Makes no difference, but we might as well flip them accordingly.

build.py
```diff
@@ -104,16 +104,10 @@ def run(self):
     f = open(self.target_filename, 'w')
     f.write(HEADER)
     f.write("""
-var isNodeJS = !!(typeof module !== 'undefined' && module.exports &&
-                  typeof window === 'undefined');
+this.IS_NODE_JS = !!(typeof module !== 'undefined' && module.exports);
 
-if (isNodeJS) {
-  var window = {};
```
NeilFraser: Creating a fake global window object for Node is bad programming.

build.py
```diff
@@ -104,16 +104,10 @@ def run(self):
     f = open(self.target_filename, 'w')
     f.write(HEADER)
     f.write("""
-var isNodeJS = !!(typeof module !== 'undefined' && module.exports &&
```
NeilFraser: Creating a global variable is bad.  Make a property which can be deleted later.

package.json
```diff
@@ -24,7 +24,8 @@
   "license": "Apache-2.0",
   "private": true,
   "devDependencies": {
-    "jshint": "latest"
+    "jshint": "latest",
+    "eslint": "2.9.0"
```
AnmAtAnm: Aren't we in the process of moving this to 4.x from 3.x?

rachel-fenichel: Apparently from 2.9.0 -> 4.16.0.  But I'll handle that/any conflicts from this.

PR #1576, opened 2018-01-27
Add generators for dynamic variable blocks.

### LINKED ISSUE #1574: Add generators for variables_get_dynamic and variables_set_dynamic blocks
### Problem statement

We have blocks in our core library with no generators.

### Expected Behavior
All of our core library blocks should have generators.
The dynamic variables_get and variables_set blocks should have the same generators as the non-dynamic versions, since all of our provided languages are dynamic.

### Actual Behavior
An assertion fails when trying to generate code for a variables_get_dynamic block.

### Steps to Reproduce

1. Create a variable from the second variables category on the playground
2. Click "To JavaScript"
3. Check the console.

### Stack Traces

```
asserts.js:81 Uncaught goog.asserts.AssertionError {message: "Assertion failed: Language "JavaScript" does not k…rate code for block type "variables_get_dynamic".", reportErrorToServer: true, messagePattern: "Assertion failed: Language "%s" does not know how to generate code for block type "%s".", stack: "AssertionError: Assertion failed: Language "JavaSc…blockly/gblockly-rr/tests/playground.html:315:79)"}```

All

rachel-fenichel:
## The basics
Fixes #1574 

- [x] I branched from develop

### Resolves

Fixes #1574 

### Proposed Changes

Add generators for dynamic variable blocks.
They're just mappings to the normal variable generators, because all of our languages are dynamic.  That means this pull request is almost entirely licenses on new files.

### Reason for Changes

The blocks aren't that useful without the generators.

### Test Coverage
I checked that I can generate code for each language with the dynamic getters and setters (in the playground), and that it's the same as for the normal getters and setters.

### DIFF
--- a/generators/dart/variables_dynamic.js
+++ b/generators/dart/variables_dynamic.js
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Visual Blocks Language
+ *
+ * Copyright 2018 Google Inc.
+ * https://developers.google.com/blockly/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


AnmAtAnm:
Code is straight forward. Downloaded and tested locally, I'd suggest generator tests, but there isn't really anything here.  We'll need to watch out for that, make sure these generators don't diverge without tests.

PR #1243, opened 2017-08-01
Master to develop 07 31 17

rachel-fenichel:
Merging master into develop before another release.

@marisaleung please check that I have the right variables code.
@RoboErikG please sanity check.

PR #1578, opened 2018-01-29
Fix #1571

### LINKED ISSUE #1571: Class Gen_compressed in build.py has two gen_accessible method
As the https://github.com/google/blockly/blob/ed4c28adf67ce6b7935532ab0fe4be8e18580907/build.py#L245-L324 , the Class Gen_compressed in build.py has two gen_accessible method.

The problem is produce by the commit 2bd056ac357737e031e9d9da77cbe8d87cb34b96
and the PR #1243.

To fixed that the L266-L292 should be delete.

windwood-xmu:
the Class Gen_compressed in build.py has two gen_accessible method.
The problem was produced by the commit 2bd056a and the PR #1243.

Delete the accidental incoming L266-L292 to fix that problem.

### DIFF
--- a/build.py
+++ b/build.py
@@ -267,33 +267,6 @@ class Gen_compressed(threading.Thread):
 
     self.do_compile(params, target_filename, filenames, "")
 
-  def gen_accessible(self):
-    target_filename = "blockly_accessible_compressed.js"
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Thanks!

Issue #1584, opened 2018-02-01
Proposal for extensible input/field types from the JSON block definition


jollytoad:
### Problem statement

Adding or modifying an input or field type in the block JSON definition requires customisation of the `Blockly.Block.prototype.interpolate_` function - adding or changing a `case` in the `switch` statement ... https://github.com/google/blockly/blob/master/core/block.js#L1143

### Proposal

I'd like to suggest replacing the switch with a object of input/field handlers, into which we can add new types or override existing types.

```javascript
Blockly.Block.jsonInputTypes = {
  input_value: function(element) {
    return this.appendValueInput(element['name']);
  },
  ...
};

Blockly.Block.jsonFieldTypes = {
  field_label: function(element) {
    return Blockly.Block.newFieldLabelFromJson_(element);
  },
  ...
};
```

and replace the switch with:

```javascript
var inputType = Blockly.Block.jsonInputTypes[element['type']]
var fieldType = Blockly.Block.jsonFieldTypes[element['type']]

if (inputType) {
  input = inputType.call(this, element)
} else if (fieldType) {
  field = fieldType.call(this, element)
}

// The input/field fn may return null so it falls back to the 'alt' definition

if (!input && !field && element['alt']) {
  element = element['alt'];
  altRepeat = true;
}
```

if its something that you think could be accepted into the codebase I'd be more than happy to prepare a PR and sign the CLA.

rachel-fenichel:
Definitely a good idea to do this for fields, and we'd be happy to have a PR for that.  

My inclination is to leave inputs alone, with the few hard-coded types, because adding a new input type is not something we see people doing as frequently.  Do you have a use case for that in particular?

@NeilFraser do you have an opinion on whether this should be done by writing directly to a `Blockly.Block.fieldJsonTypes` object, vs registering the function some other way?

RoboErikG:
I have a preference for API calls so that it's less painful if we need to change the implementation in the future.

jollytoad:
Yeah I was thinking an API call to register/override rather than just assigning to the map, didn't want to overwhelm the proposal with too many details.

I included inputs as they are already handled in the same fn as fields, and it seemed appropriate to handle them in the same way and ditch the switch entirely rather than having some things handled in switch and others delegating.

[...]

PR #1594, opened 2018-02-07
Customising field types using a register of fields #1584

jollytoad:
## The basics

- [x] I branched from develop

### Resolves

#1584

### Proposed Changes

* Adds a register for Blockly.Field types
* Adds a Blockly.Field.fromJson to construct the appropriate field instance according to the register
* Existing fields now register themselves, allowing exclusion or replacement of the entire field

### Reason for Changes

Allow adding of new field types and overriding of the standard ones without having to hack the core of Blockly. See #1584 for more details.

### Test Coverage

Existing json tests provide some assurance, although they don't appear to have comprehensive coverage for all field types.

Tested on:

* Desktop Chrome

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1223,42 +1223,12 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
             case 'input_dummy':
               input = this.appendDummyInput(element['name']);
               break;
-            case 'field_label':
-              field = Blockly.FieldLabel.fromJson_(element);
-              break;
-            case 'field_input':
-              field = Blockly.FieldTextInput.fromJson_(element);
-              break;
-            case 'field_angle':
-              field = Blockly.FieldAngle.fromJson_(element);
-              break;
-            case 'field_checkbox':
-              field = Blockly.FieldCheckbox.fromJson_(element);
-              break;
-            case 'field_colour':
-              field = Blockly.FieldColour.fromJson_(element);
-              break;
-            case 'field_variable':
-              field = Blockly.FieldVariable.fromJson_(element);
-              break;
-            case 'field_dropdown':
-              field = Blockly.FieldDropdown.fromJson_(element);
-              break;
-            case 'field_image':
-              field = Blockly.FieldImage.fromJson_(element);
-              break;
-            case 'field_number':
-              field = Blockly.FieldNumber.fromJson_(element);
-              break;
-            case 'field_date':
-              if (Blockly.FieldDate) {
-                field = Blockly.FieldDate.fromJson_(element);
-                break;
-              }
-              // Fall through if FieldDate is not compiled in.
+
             default:
+              field = Blockly.Field.fromJson_(element);
+
               // Unknown field.
-              if (element['alt']) {
+              if (!field && element['alt']) {
                 element = element['alt'];
                 altRepeat = true;
               }
--- a/core/block.js
+++ b/core/block.js
@@ -1225,7 +1225,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
               break;
 
             default:
-              field = Blockly.Field.fromJson_(element);
+              field = Blockly.Field.fromJson(element);
 
               // Unknown field.
               if (!field && element['alt']) {
--- a/core/block.js
+++ b/core/block.js
@@ -1228,9 +1228,16 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
               field = Blockly.Field.fromJson(element);
 
               // Unknown field.
-              if (!field && element['alt']) {
-                element = element['alt'];
-                altRepeat = true;
+              if (!field) {
+                if (element['alt']) {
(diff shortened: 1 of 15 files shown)


jollytoad:
I'll add a unit for registration of a custom field type

core/field.js
```diff
@@ -52,6 +52,52 @@ Blockly.Field = function(text, opt_validator) {
...
+ * @private
+ */
+Blockly.Field.ALL_ = {};
+
+/**
+ * Registers a field type. May also override an existing field type.
+ * Blockly.Field.fromJson uses this registry to find the appropriate field.
+ * @param {!string} type The field type name as used in the JSON definition.
+ * @param {!{fromJson: Function}} fieldClass The field class containing a
+ *                                 fromJson function that can construct an
```
rachel-fenichel: Nit: indent line continuations +4 spaces for these descriptions (here and throughout).

core/field.js
```diff
@@ -52,6 +52,52 @@ Blockly.Field = function(text, opt_validator) {
   this.setValidator(opt_validator);
 };
 
+/**
+ * The set of all registered fields, keyed by field type as used in the JSON
+ * definition of a block.
+ * @private
```
rachel-fenichel: Add a type annotation.  I think it's:
`@type {string, !{fromJson: Function}}`

core/field.js
```diff
@@ -52,6 +52,52 @@ Blockly.Field = function(text, opt_validator) {
...
+ */
+Blockly.Field.ALL_ = {};
```
rachel-fenichel: Let's call this `Blockly.Field.TYPE_MAP_`

core/field.js
```diff
@@ -52,6 +52,52 @@ Blockly.Field = function(text, opt_validator) {
...
+ * @param {!string} type The field type name as used in the JSON definition.
+ * @param {!{fromJson: Function}} fieldClass The field class containing a
+ *                                 fromJson function that can construct an
+ *                                 instance of the field.
+ * @throws {Error} if the type name is empty, or the fieldClass is not an
+ *                 object containing a fromJson function.
+ */
+Blockly.Field.register = function(type, fieldClass) {
+  if (!goog.isString(type) || goog.string.isEmptyOrWhitespace(type)) {
+    throw new Error('Error: Invalid field type "' + type + '"');
```
rachel-fenichel: Nit: no need for the leading `'Error:` text, since the browser will prepend something like that anyway.  Here and below.

core/block.js
```diff
@@ -1223,42 +1223,12 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
...
+
             default:
+              field = Blockly.Field.fromJson(element);
+
               // Unknown field.
-              if (element['alt']) {
+              if (!field && element['alt']) {
                 element = element['alt'];
                 altRepeat = true;
               }
```
rachel-fenichel: This change is going to break some forks that rely on custom fields.  They should notice when they get a merge conflict in this function, but it's not guaranteed.

Add a warning if no field was created and no alt was found:
```
console.warn('Blockly could not create a field of type ' + element['type'] + 
    '. You may need to register your custom field.  See ' +
    'github.com/google/blockly/issues/1584');
```

core/field.js
jollytoad: So 4 spaces between * and fromJson?

    * @param {!{fromJson: Function}} fieldClass The field class containing a
    *    fromJson function that can construct an ...

rachel-fenichel: 5 spaces between * and `fromJson`, because there was already one space between * and `@param`.
```
* @param {!{fromJson: Function}} fieldClass The field class containing a
*     fromJson function that can construct an ...
```

core/field.js
jollytoad: I'm not that familiar with closure type system, I mostly work in Typescript, but from the docs... https://github.com/google/closure-compiler/wiki/A-word-about-the-type-Object#maps

I think its...

    @type {!Object<string, !{fromJson: Function}>}

I'll add this in

but I was wondering if the value type of the map should define that it inherits from Field, but i don't even know if that's possible with closure, imaginary example:

    @type {!Object<string, constructor extending Blockly.Field>}

core/field.js
jollytoad: The errors in Blockly.Extensions probably need updating too then

core/field.js
jollytoad: Ignore me, i see from other docs what you mean!

core/field.js
```diff
@@ -52,6 +52,52 @@ Blockly.Field = function(text, opt_validator) {
...
+  }
+  Blockly.Field.TYPE_MAP_[type] = fieldClass;
+};
+
+/**
+ * Construct a Field from a JSON arg object.
+ * Finds the appropriate registered field by the type name as registered using
+ * Blockly.Field.register.
+ * @param {!Object} options A JSON object with a type and options specific
+ *                          to the field type.
```
rachel-fenichel: Final nit: indentation on these parameter/return descriptions as well.

PR #1597, opened 2018-02-09
Make WorkspaceSvg.refreshToolboxSelection_ package instead of private

### LINKED ISSUE #1554: Make WorkspaceSvg.refreshToolboxSelection function package instead of private.
### Problem statement

In scratch-blocks, there is a use case for workspace.refreshToolboxSelection to be called from outside the WorkspaceSvg class. Make this function a package function to allow this.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1554 
### Proposed Changes

Makes a private function package instead.
### Reason for Changes

Scratch-blocks has a good use case for `refreshToolboxSelection` to be called from outside the WorkspaceSvg class.  It's new enough and private enough that we can just rename it.

### Test Coverage

Tested on:
* Desktop Chrome

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -929,9 +929,9 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
 
 /**
  * Refresh the toolbox unless there's a drag in progress.
- * @private
+ * @package
  */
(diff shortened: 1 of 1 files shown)


AnmAtAnm:
Lgtm

PR #1599, opened 2018-02-09
Don't allow block deletion in the flyout.  Make copy-paste always copy to the main workspace

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1373

### Proposed Changes

On keyboard paste, paste to the target workspace.
On keyboard cut, backspace, and delete, block the action if the selected block was in the flyout.

### Reason for Changes

See https://github.com/LLK/scratch-blocks/issues/1373

You could delete and paste blocks in the flyout, which was lovely.

### Test Coverage

Tested on:
* Desktop Chrome

### Additional Information

@kchadha fyi

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -274,6 +274,7 @@ Blockly.WorkspaceSvg.prototype.toolboxCategoryCallbacks_ = {};
  * @package
  */
 Blockly.WorkspaceSvg.prototype.targetWorkspace = null;
+
 /**
  * Inverted screen CTM, for use in mouseToSvg.
  * @type {SVGMatrix}
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -268,6 +268,13 @@ Blockly.WorkspaceSvg.prototype.flyoutButtonCallbacks_ = {};
 Blockly.WorkspaceSvg.prototype.toolboxCategoryCallbacks_ = {};
 
 /**
+ * In a flyout, the target workspace where blocks should be placed after a drag.
+ * Otherwise null.
+ * @type {?Blockly.WorkspaceSvg}
+ * @package
+ */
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
This bug is somehow new behaviour--it doesn't happen on blockly games.  But not too new, because it's on scratch-blocks.

amazinigmech2418:
I think the ctrl. + v is the problem. It should make a copy and place it at the mouse pointer if the mouse pointer isn't in the menu.

rachel-fenichel:
Blockly doesn't attach blocks to the pointer on paste.  Also, we don't know the pointer location during a key event (without additional work to track it continuously).

amazinigmech2418:
@rachel-fenichel You could use a mousemove event to read the mouse position when pasting

rachel-fenichel:
This issue came up because my dragging work made it possible to select a block in the flyout.  Previously `Blockly.selected` would not be set by interacting with a block in the flyout.

core/workspace_svg.js
```diff
@@ -267,6 +267,13 @@ Blockly.WorkspaceSvg.prototype.flyoutButtonCallbacks_ = {};
...
 Blockly.WorkspaceSvg.prototype.toolboxCategoryCallbacks_ = {};
 
+/**
+ * In a flyout, the target workspace where blocks should be placed after a drag.
+ * Otherwise null.
+ * @type {?Blockly.WorkspaceSvg}
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.targetWorkspace = null;
 /**
```
AnmAtAnm: Nit: empty line before documentation

rachel-fenichel: Fixed

Issue #1558, opened 2018-01-21
Disallow block type names that might match anything in Blockly.Generator.prototype


AnmAtAnm:
### Problem statement

We currently define block generators as:

```js
Blockly.JavaScript['asdf'] = function(block) {
  // TODO: Assemble JavaScript into code variable.
  var code = '...;\n';
  return code;
};
```

However, `Blockly.JavaScript` (and all other built in language implementations) are object instances of `Blockly.Generator` with existing fields and methods. That means the following should be considered invalid block type names:

```
"INFINITE_LOOP_TRAP", "STATEMENT_PREFIX", "INDENT", "COMMENT_WRAP",
"ORDER_OVERRIDES", "workspaceToCode", "prefixLines", "allNestedComments", "blockToCode",
"valueToCode", "statementToCode", "addLoopTrap", "RESERVED_WORDS_", "addReservedWords",
"FUNCTION_NAME_PLACEHOLDER_", "provideFunction_", "init", "scrub_", "finish",
"scrubNakedValue"
```

Assigning any of these may corrupt generator behavior.

AnmAtAnm:
TODO: Console warning ➔ thrown error

PR #1603, opened 2018-02-12
Warn conflicts with Blockly.Generator.prototype

AnmAtAnm:
Console warnings on block prototypeNames that conflict with Blockly.Generator.prototype

First pass at #1558

## The basics

- [x] I branched from develop

### Resolves

Warning on conflicting Block names. Implementing a warning first, since fully banning type names is a breaking change.

### Proposed Changes

Warn on conflict with Blockly.Generator.prototype members.

### Reason for Changes

By convention, we use the Generator instance objects as generator function maps. This will conflict with prototype lookup for name collisions between block names and said members, and needs to be disallowed.  However, this is a breaking changes, so emit a warning in this release.

### Test Coverage

Locally edited Playground test blocks to have conflicting block instances. Saw the expected the error.

Also ran the build script with no bad effect on existing demos.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -57,7 +57,8 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
     console.warn('FUTURE ERROR: Block prototypeName "' + prototypeName
         + '`" conflicts with Blockly.Generator members. Registering Generators '
         + 'for this block type will incur errors.'
-        + 'This name will be DISALLOWED in future versions of Blockly.');
+        + '\nThis name will be DISALLOWED (throwing an error) in future '
+        + 'versions of Blockly.');
   }
 
   /** @type {string} */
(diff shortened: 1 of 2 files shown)


core/block.js
```diff
@@ -52,6 +53,13 @@ goog.require('goog.string');
  * @constructor
  */
 Blockly.Block = function(workspace, prototypeName, opt_id) {
+  if (typeof Blockly.Generator.prototype[prototypeName] !== 'undefined') {
+    console.warn('FUTURE ERROR: Block prototypeName "' + prototypeName
+        + '`" conflicts with Blockly.Generator members. Registering Generators '
```
rachel-fenichel: There's an extra ` here that doesn't match the rest of your quote.

PR #1605, opened 2018-02-13
Bugfix/1600

### LINKED ISSUE #1600: Rename the dynamic variables category on the playground before release
The playground on develop now has two variable categories: "dynamic" and "not dynamic", both labeled "variables".  Either the "dynamic" one should get a better name, or we should have another option to switch between "dynamic" and "not dynamic".

Need to do this before updating the published playground.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fix #1600 

### Proposed Changes

Add a new option to the "simple/categories" toolbox dropdown in the playground.  The options are now "Categories (typed variables)", "Categories (untyped variables)", and "Simple".

I did this by adding a new chunk of XML because I want to keep the sample XML in the playground as simple as possible, such that someone can just copy the sample XML to get started.  A shorter way would have been to dynamically change the XML, but that would have spread the construction of the toolbox XML more than I want.

I added some comments to explain what each chunk of XML means, and links to the documentation for first-time developers who are looking through the playground code as a starting point.

### Reason for Changes

Mixing the typed and untyped variable blocks in the playground was confusing.
### Test Coverage
The playground still works.  I can open all three types of toolbox, by using the dropdown or by changing the URL.  The variable categories both open.

Tested on:
* Desktop Chrome

### Additional Information

In the second commit I cleaned up some indentation in the toolbox options struct.

### DIFF
--- a/tests/playground.html
+++ b/tests/playground.html
@@ -83,31 +83,36 @@ function start() {
   document.forms.options.elements.side.value = side;
   // Create main workspace.
   workspace = Blockly.inject('blocklyDiv',
-          {comments: true,
-           collapse: true,
-           disable: true,
-           grid:
-             {spacing: 25,
-              length: 3,
-              colour: '#ccc',
-              snap: true},
-           horizontalLayout: side == 'top' || side == 'bottom',
-           maxBlocks: Infinity,
-           media: '../media/',
-           oneBasedIndex: true,
-           readOnly: false,
-           rtl: rtl,
-           scrollbars: true,
-           toolbox: toolbox,
-           toolboxPosition: side == 'top' || side == 'start' ? 'start' : 'end',
-           zoom:
-             {controls: true,
-              wheel: true,
-              startScale: 1.0,
-              maxScale: 4,
-              minScale: .25,
-              scaleSpeed: 1.1}
-          });
+      {
+        comments: true,
+        collapse: true,
+        disable: true,
+        grid:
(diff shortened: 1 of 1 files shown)


tests/playground.html
```diff
@@ -688,6 +719,332 @@ <h1>Blockly Playground</h1>
     </category>
     <sep></sep>
     <category name="Variables" colour="330" custom="VARIABLE"></category>
+    <category name="Functions" colour="290" custom="PROCEDURE"></category>
+  </xml>
+
+  <!-- toolbox-categories-untyped-variables has a category menu and an
```
AnmAtAnm: untyped -> typed

tests/playground.html
```diff
@@ -137,7 +142,24 @@
 
 function getToolboxElement() {
   var match = location.search.match(/toolbox=([^&]+)/);
-  return document.getElementById('toolbox-' + (match ? match[1] : 'categories'));
+  // Default to the basic toolbox with categories and untyped variables,
+  // but override that if the toolbox type is set in the URL.
+  var toolboxSuffix = (match ? match[1] : 'categories');
```
AnmAtAnm: 'categories' should be 'toolbox-categories', right?

rachel-fenichel: No, because it's a suffix.  It's added to 'toolbox-' on line 150.

PR #1614, opened 2018-02-14
Don't register mouse events if pointer events are supported

rachel-fenichel:
See https://github.com/Microsoft/pxt-blockly/pull/96

@microsoftsam

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -53,7 +53,9 @@ goog.require('Blockly.WorkspaceSvg');
 goog.require('Blockly.constants');
 goog.require('Blockly.inject');
 goog.require('Blockly.utils');
+
 goog.require('goog.color');
+goog.require('goog.events.BrowserFeature');
 goog.require('goog.userAgent');
 
 
@@ -452,7 +454,8 @@ Blockly.bindEventWithChecks_ = function(node, name, thisObject, func,
 
   var bindData = [];
   // Don't register the mouse event if an equivalent pointer event is supported.
-  if (!window.PointerEvent || !(name in Blockly.Touch.TOUCH_MAP)) {
+  if (!goog.events.BrowserFeature.POINTER_EVENTS ||
+      !(name in Blockly.Touch.TOUCH_MAP)) {
     node.addEventListener(name, wrapFunc, false);
     bindData.push([node, name, wrapFunc]);
   }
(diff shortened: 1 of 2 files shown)


core/blockly.js
```diff
@@ -450,10 +450,14 @@ Blockly.bindEventWithChecks_ = function(node, name, thisObject, func,
     }
   };
 
-  node.addEventListener(name, wrapFunc, false);
-  var bindData = [[node, name, wrapFunc]];
+  var bindData = [];
+  // Don't register the mouse event if an equivalent pointer event is supported.
+  if (!window.PointerEvent || !(name in Blockly.Touch.TOUCH_MAP)) {
```
AnmAtAnm: window might not exist in headless mode. Are we certain this function will never be called in headless? I feel more comfortable checking `window` too.

core/blockly.js
```diff
@@ -495,10 +499,13 @@ Blockly.bindEvent_ = function(node, name, thisObject, func) {
     }
   };
 
-  node.addEventListener(name, wrapFunc, false);
-  var bindData = [[node, name, wrapFunc]];
-
-  // Add equivalent touch event.
+  var bindData = [];
+  // Don't register the mouse event if an equivalent pointer event is supported.
+  if (!window.PointerEvent || !(name in Blockly.Touch.TOUCH_MAP)) {
```
AnmAtAnm: Same concern with `window`

core/blockly.js
rachel-fenichel: Good point.  That also means that there should be a check for window in touch.js: 
```
Blockly.Touch.TOUCH_MAP = {};
if (window.PointerEvent) {
  Blockly.Touch.TOUCH_MAP = {
    'mousedown': ['pointerdown'],
    'mousemove': ['pointermove'],
    'mouseup': ['pointerup', 'pointercancel']
  };
} else if (goog.events.BrowserFeature.TOUCH_ENABLED) {
  Blockly.Touch.TOUCH_MAP = {
    'mousedown': ['touchstart'],
    'mousemove': ['touchmove'],
    'mouseup': ['touchend', 'touchcancel']
  };
}

```

In the particular case of `bindEvent_`, it's called from inject.js with `window` as the `node` argument, so we may have already failed.

It looks like we used to create a fake window object to solve this problem (not the best solution): https://github.com/google/blockly/pull/1566/files  Now that's gone.

rachel-fenichel:
This still doesn't work--now I get zero clicks in firefox and one click in chrome, instead of one click in firefox and two in chrome.  I'm disabling them for this release.

rachel-fenichel:
Merged, because @samelhusseini proved that the problem was with the firefox simulator.

Issue #1591, opened 2018-02-07
goog.date.fromIsoString not defined in compressed blockly file


kmasuyama:
As of a few hours ago, the calendar in the field>date block has stopped working. The console throws the errors below, and the block acts as if dragged (yellow highlight, movable). However, the block cannot be placed down, and no further action is possible. Using blockly_uncompressed.js behaves as expected, so the problem seems to be in the compression process. 

Error on clicking the calendar:
```
blockly_compressed.js:2168 Uncaught TypeError: goog.date.fromIsoString is not a function
```

Warnings on subsequent clicks:
```
blockly_compressed.js:1768 Trying to end a gesture recursively.
blockly_compressed.js:1953 tried to start the same gesture twice
Blockly.WorkspaceSvg.getGesture @ blockly_compressed.js:1953
Blockly.Field.onMouseDown_ @ blockly_compressed.js:1788
h @ blockly_compressed.js:2345
blockly_compressed.js:1953 tried to start the same gesture twice
Blockly.WorkspaceSvg.getGesture @ blockly_compressed.js:1953
Blockly.BlockSvg.onMouseDown_ @ blockly_compressed.js:2041
h @ blockly_compressed.js:2345
blockly_compressed.js:1768 Trying to end a gesture recursively.
```

### Expected Behavior

I can click the calendar and pick a date. The block stays fixed and I can continue using the app.

### Actual Behavior

The calendar opens, but I cannot pick a date or put the block down.

### Steps to Reproduce

run python build.py on a project with a date block.
Open the app and try clicking on the calendar

### Additional Info

The first few builds when I started seeing the issue showed server error messages during the build.py process (no longer showing up). I think this may have to do with the online clojure compiler.

I am using the following:
node 7.9.0
npm 4.2.0
electron 1.8.2
electron-packager 10.1.2
python 2.7.13
blockly 1.0.0
google-closure-library v20171203.0.0

AnmAtAnm:
Using Blockly Playground (blockly_uncompressed.js), I see similar behavior with a slightly different error message:

    Uncaught TypeError: goog.date.DateTime.fromIsoString is not a function
        at Blockly.FieldDate.createWidget_ (field_date.js:158)
        at Blockly.FieldDate.showEditor_ (field_date.js:123)
        at Blockly.Gesture.doFieldClick_ (gesture.js:707)
        at Blockly.Gesture.handleUp (gesture.js:573)
        at wrapFunc (blockly.js:447)
[...]

PR #1620, opened 2018-02-15
Updating Closure version

AnmAtAnm:
## The basics

- [x] I branched from develop

### Proposed Changes

Updating closure dependency.

### Reason for Changes

goog.date.fromIsoString was removed upstream and we need to stay up to date to be compatible with other closure apps/libraries. See #1591.

### Test Coverage

Build.py uses online compiler, thus latest Closure. That works, so I suspect this will too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "unused": true
   },
   "dependencies": {
-    "google-closure-library": "^20171203.0.0",
+    "google-closure-library": "^20180204.0.0",
     "install": "^0.8.8",
     "npm": "^4.4.4",
     "webdriverio": "^4.6.2"


rachel-fenichel:
The compiler version and the closure library version are different things.  Build.py uses the version of closure that you have on your computer and sends those files to the online closure compiler.

rachel-fenichel:
My previous comment was wrong.

gen_compressed [passes](https://github.com/google/blockly/blob/master/build.py#L223) "use_closure_library"= "true", so it (I think) gets whatever version the server is using. 

gen_uncompressed gets references to whatever version of closure library you have locally.

PR #1631, opened 2018-02-16
Delete extra variables at the end of a procedure argument edit

### LINKED ISSUE #1544: Too many variables get created when editing the name of a procedure argument.
pkdjkev9erc

### Problem statement
Every partial variable name is becoming a new variable.  So creating the argument "abcde" creates the variables "a", "ab", "abc", "abcd", and "abcde". None of the variables are autodeleted and all of them sort before the intended name, so it's a bad pollution of the variables category.

### Expected Behavior

Only the final variable name causes a new variable to be created.
### Actual Behavior
See above

### Steps to Reproduce

1. Create a procedure
2. Click the gear to open the mutator
3. Add an argument
4. Rename the argument
5. Open the variables category
6. Notice that there are too many variables

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1544 

### Proposed Changes

Update the procedures_mutatorarg block.  At the beginning of an edit to an argument name, start tracking variables that have been created as the user types.  At the end of the edit, delete all of the intermediate variables, saving only the one using the final variable name.  Don't delete variables that were already in use.

### Reason for Changes

See #1544.  We were creating one variable per key typed.
### Test Coverage
Tested on:
* Desktop Chrome

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -470,12 +470,12 @@ Blockly.Blocks['procedures_mutatorarg'] = {
     var field = new Blockly.FieldTextInput('x', this.validator_);
     // Hack: override showEditor to do just a little bit more work.
     // We don't have a good place to hook into the start of a text edit.
-    var oldShowEditorFn = field.showEditor_.bind(field);
+    field.oldShowEditorFn_ = field.showEditor_;
     var newShowEditorFn = function() {
       this.createdVariables_ = [];
-      oldShowEditorFn();
+      this.oldShowEditorFn_();
     };
-    field.showEditor_ = newShowEditorFn.bind(field);
+    field.showEditor_ = newShowEditorFn;
 
     this.appendDummyInput()
         .appendField(Blockly.Msg.PROCEDURES_MUTATORARG_TITLE)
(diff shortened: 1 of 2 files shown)


blocks/procedures.js
```diff
@@ -469,6 +468,15 @@ Blockly.Blocks['procedures_mutatorarg'] = {
...
   init: function() {
     var field = new Blockly.FieldTextInput('x', this.validator_);
+    // Hack: override showEditor to do just a little bit more work.
+    // We don't have a good place to hook into the start of a text edit.
+    var oldShowEditorFn = field.showEditor_.bind(field);
+    var newShowEditorFn = function() {
+      this.createdVariables_ = [];
+      oldShowEditorFn();
+    };
+    field.showEditor_ = newShowEditorFn.bind(field);
```
AnmAtAnm: Nit: Is the bind necessary?  Is there any chance it will be called without the `field` in `field.showEditor_()`?
Same could be said of oldShowEditorFn, as long as it is called via `field.oldShowEditorFn()` in the function expression.

blocks/procedures.js
```diff
@@ -478,43 +486,61 @@ Blockly.Blocks['procedures_mutatorarg'] = {
     this.setTooltip(Blockly.Msg.PROCEDURES_MUTATORARG_TOOLTIP);
     this.contextMenu = false;
 
+    // Get a handle on the parent workspace.
+    // TODO: Do I need to delete this during dispose?
+    this.outerWs_ = Blockly.Mutator.findParentWs(this.workspace);
```
AnmAtAnm: I would definite clear this.outWs_ during `dispose()`.

rachel-fenichel: Moved into the functions that use it.  I don't like the idea of overriding the block's dispose function for this.

blocks/procedures.js
rachel-fenichel: Done.

PR #1638, opened 2018-02-17
Fix #1619. buildTooltipWithFieldValue() => buildTooltipWithFieldText()

### LINKED ISSUE #1619: Var id instead of var name in tooltip
Loops: "Count with i from 1 to 10 by 1"

screen shot 2018-02-14 at 2 20 17 pm

Seen in rc/feb_2018

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

Fix #1619 

The "value" of FieldVariables was the variable id.
However, we want the field text for the human visible tooltip. Refactoring and renaming `buildTooltipWithFieldValue` to use the field text.

### Proposed Changes

 * Refactor `buildTooltipWithFieldValue()` to use field text, instead of value. Rename accordingly.
 * Make sure all use (the blocks in question) use the renamed function.
 * Remove custom functions that were not using this helper.

### Reason for Changes

#1619 

### Test Coverage

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/extensions.js
+++ b/core/extensions.js
@@ -422,8 +422,9 @@ Blockly.Extensions.buildTooltipWithFieldText = function(msgTemplate,
    */
   var extensionFn = function() {
     this.setTooltip(function() {
+      var field = this.getField(fieldName);
       return Blockly.utils.replaceMessageReferences(msgTemplate)
-          .replace('%1', this.getField(fieldName).getText());
+          .replace('%1', field ? field.getText() : '');
     }.bind(this));
   };
   return extensionFn;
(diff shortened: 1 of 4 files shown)


core/extensions.js
```diff
@@ -423,7 +423,7 @@ Blockly.Extensions.buildTooltipWithFieldValue = function(msgTemplate,
   var extensionFn = function() {
     this.setTooltip(function() {
       return Blockly.utils.replaceMessageReferences(msgTemplate)
-          .replace('%1', this.getFieldValue(fieldName));
+          .replace('%1', this.getField(fieldName).getText());
```
rachel-fenichel: Check if the field exists and use the empty string if it doesn't

PR #1646, opened 2018-02-21
Adding test blocks for Unicode and Emojis

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

Tests for critical infrastructure in light of target audience: emojis.

### Proposed Changes

Adds test blocks to demonstrate capabilities. Stopped short of adding BiDi override characters.
I added to the toolbox category "Images". It isn't really images, but it didn't seem worth making a separate category at this stage.

### Reason for Changes

CRITICAL INFRASTRUCTURE!!!

### Test Coverage

I opened the toolbox hosting the new blocks:

screen shot 2018-02-20 at 4 08 09 pm

Tested on:
 * Desktop Chrome


rachel-fenichel:
Note that we do not currently intend to fix zalgo.  But it's a reminder that our text height is not dynamically calculated (thanks, svg spec), so we may not handle accented characters as nicely as we would like.

Issue #1411, opened 2017-11-01
TEXT_CHARAT block: Ordinal translation issues in simplified mandarin and ukranian


AnmAtAnm:
Given..

    $ grep -R ORDINAL_NUMBER_SUFFIX msg/json 
    msg/json/uk.json:	"ORDINAL_NUMBER_SUFFIX": "-ий.",
    msg/json/zh-hans.json:	"ORDINAL_NUMBER_SUFFIX": "空白",

I believe the following are in error:

    $ grep -R TEXT_CHARAT_TAIL msg/json 
    msg/json/uk.json:	"TEXT_CHARAT_TAIL": "-ий.",
    msg/json/zh-hans.json:	"TEXT_CHARAT_TAIL": "空白",

Resulting in duplicate suffixes:
<img width="470" alt="screen shot 2017-11-01 at 4 15 35 pm" src="https://user-images.githubusercontent.com/9916202/32302665-17b89c92-bf21-11e7-9d7d-92fe6205a31a.png">
<img width="435" alt="screen shot 2017-11-01 at 4 14 05 pm" src="https://user-images.githubusercontent.com/9916202/32302661-1341399e-bf21-11e7-91ec-faa9e4abe0e0.png">

NeilFraser:
I don't see this.  Instead I get traceback:
```
blockly_compressed.js:1438 Uncaught TypeError: Cannot read property 'alt' of undefined
    at Blockly.FieldDropdown.trimOptions_ (blockly_compressed.js:1438)
    at new Blockly.FieldDropdown (blockly_compressed.js:1430)
    at Blockly.BlockSvg.Blockly.Block.interpolate_ (blockly_compressed.js:1271)
    at Blockly.BlockSvg.Blockly.Block.jsonInit (blockly_compressed.js:1264)
    at Blockly.BlockSvg.init (blockly_compressed.js:1598)
    at Blockly.BlockSvg.Blockly.Block (blockly_compressed.js:1238)
    at new Blockly.BlockSvg (blockly_compressed.js:1292)
    at Blockly.WorkspaceSvg.newBlock (blockly_compressed.js:1176)
    at Object.Blockly.Xml.domToBlockHeadless_ (blockly_compressed.js:1149)
    at Object.Blockly.Xml.domToBlock (blockly_compressed.js:1147)
```
All languages, all browsers, compiled only.

AnmAtAnm:
[...]

PR #1656, opened 2018-02-23
Move ordinal in block 'text_getSubstring'

zgtm:
On updating the block 'text_getSubstring', the ordinal string 'ORDINAL1' is moved when the corresponding corresponding input field is moved so that the ordinal string does not end up at the end of the block.

## The basics

- [x] I branched from develop

### Resolves

None

### Proposed Changes

Reordering the ordinal sign in the block to the right place, even after selecting different entries in the block input fields. Here in German:

image

### Reason for Changes

In languages with ordinal sign, the ordinal sign ends up on the wrong place in this block, if you first select "get substring from first letter" and then select "get substring from letter #". Here in German:

image

### Test Coverage

Tested on:
Desktop Chrome
Desktop Firefox

### DIFF
--- a/blocks/text.js
+++ b/blocks/text.js
@@ -313,6 +313,8 @@ Blockly.Blocks['text_getSubstring'] = {
         .appendField(menu, 'WHERE' + n);
     if (n == 1) {
       this.moveInputBefore('AT1', 'AT2');
+      if (this.getInput('ORDINAL1')) {
+        this.moveInputBefore('ORDINAL1', 'AT2');
     }
   }
 };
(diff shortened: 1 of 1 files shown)


zgtm:
Is it necessary or helpful for me to create an issue for this as well?

AnmAtAnm:
Can you also look at the `text_charat` block?

AnmAtAnm:
For context for English speakers, if we set `Blockly.Msg.ORDINAL_NUMBER_SUFFIX = 'th'`:

screen shot 2018-02-23 at 10 03 34 am

zgtm:
> Can you also look at the text_charat block? I ask because of known ordinal problems (#1411) in translation. We don't use the English ordinals, which means we don't test these features very often.

The `text_charat` block works fine for me, I think #1411 is a translation issue.

> I verified it works okay with the equivalent list_sublist block (I think that is its name).

Yes, that's where I took the patch from.

Actually, one could also think of replacing https://github.com/google/blockly/blob/develop/blocks/text.js#L290-L295
with 
https://github.com/google/blockly/blob/develop/blocks/lists.js#L748-L750
which I think is a bit more comprehensive.

Issue #1476, opened 2017-11-22
Workspace comments


rachel-fenichel:
Currently Blockly only has comments on blocks, which can be created through the right-click menu on a block.

Scratch has comments on the workspace and on blocks.  Comments can be dragged between blocks and the workspace.  We want the same thing.

The `comments` branch has some initial work here: workspace comments that can be freely dragged around the workspace.

More work we need to do:
- readonly
- events
- RTL
- scaling/scrolling
- block comments using new comment object
- workspace options: disabling comments
- dragging workspace comments to blocks
- dragging block comments to the workspace
- resize
- delete (#1662)
- collapse
- IE decisions
- selected state (how does that interact with Blockly.selected?)
- workspace comments in generators
- clean up dragging code/combine with block dragging code
  - Implemented in #1661 
- highlighting/unhighlighting
- disallowing comments in the flyout (may be based on workspace options)

testing:
- id conflict
- creation
- lookup
- deletion
- XML cycle

code cleanup!

Current plan is to:
- try to use the new comment object for block comments but not be able to drag between blocks and the workspace
- clean up dragging
- merge that into blockly
- then add dragging from blocks to workspace and back.

BeksOmega:
Superseded by https://github.com/google/blockly-samples/issues/827

PR #1661, opened 2018-02-27
Comments with bubble dragger

rachel-fenichel:
## The basics

- [ ] I branched from develop **I branched from comments_v2**
- [ ] My pull request is against develop **My pull request is against comments_v2**

### Resolves

Progress on #1476.

### Proposed Changes

After a lot of back and forth on design, I updated the comment work we did in November instead of fully integrating workspace comments and block comments.

This change uses the new BubbleDragger to drag workspace comments.  It already drags block comments, mutators, and warnings.  To do this I added a few functions that the BubbleDragger expects to exist on anything it's dragging, but it's not a lot of work overall.

### Reason for Changes

Make workspace comments draggable without needing separate draggers for ws comments and block comments.

### Test Coverage
Tested in the playground by creating some comments, editing them, dragging them around, serializing them, reloading, and deserializing them.

### Additional Information

@samelhusseini The good news is that the design we worked on in November is pretty solid, and I really like the way your ws comments look.  The bad news is that it took me a long time to come to that conclusion, and that it's not easy to update block comments to look like workspace comments.


msg/json/en.json
```diff
@@ -1,7 +1,7 @@
 {
 	"@metadata": {
 		"author": "Ellen Spertus <ellen.spertus@gmail.com>",
-		"lastupdated": "2018-02-08 15:48:36.801901",
+		"lastupdated": "2018-02-26 11:01:23.829885",
```
AnmAtAnm: Remove file from PR.

core/workspace_comment/workspace_comment_svg.js
```diff
@@ -503,3 +516,7 @@ Blockly.WorkspaceCommentSvg.prototype.setDeleteStyle = function(enable) {
         /** @type {!Element} */ (this.svgGroup_), 'blocklyDraggingDelete');
   }
 };
+
+Blockly.WorkspaceCommentSvg.prototype.setAutoLayout = function() {
```
RoboErikG: I feel like I'm missing something. I don't see where this would get called. Is there another change that's making this part of the bubble dragger path?

rachel-fenichel: The change to make this use the bubble dragger is right here: https://github.com/google/blockly/pull/1661/files#diff-2d76f92ba68d3e43c2c00ddf993751deR148

That's where I'm calling `gesture.handleBubbleStart(e, this);`

The `setAutoLayout` part is from when I added the bubble dragger initially, and I haven't yet sorted out if that's realy part of the API that I want to keep around.  But it's called from the bubble dragger.

Issue #1671, opened 2018-02-28
New field registration code does not work under advanced compilation


rachel-fenichel:
### Problem statement
Develop is currently broken under advanced compilation.  Master is fine.

### Steps to Reproduce
1. Check out develop
2. Go to `tests/compile`
3. Run `compile.sh`
4. Open `tests/compile/index.html` in your browser
5. Blockly does not load
6. Open developer tools
7. See an error in the console (stack trace below)

### Stack Traces

```
main_compressed.js:formatted:6932 Uncaught Error: Field "function mj(a,b){this.W=new Jd(0,17.5);this.Gj=b;this.aa(a)}" must have a fromJson function
    at ej (main_compressed.js:formatted:6932)
    at main_compressed.js:formatted:7151
```

### Additional Information

Master is fine--I followed the same steps and had no problems.

The error message comes from `Blockly.Field.register`: 
```

/**
 * Registers a field type. May also override an existing field type.
 * Blockly.Field.fromJson uses this registry to find the appropriate field.
 * @param {!string} type The field type name as used in the JSON definition.
 * @param {!{fromJson: Function}} fieldClass The field class containing a
 *     fromJson function that can construct an instance of the field.
 * @throws {Error} if the type name is empty, or the fieldClass is not an
 *     object containing a fromJson function.
 */
Blockly.Field.register = function(type, fieldClass) {
  if (!goog.isString(type) || goog.string.isEmptyOrWhitespace(type)) {
    throw new Error('Invalid field type "' + type + '"');
  }
  if (!goog.isObject(fieldClass) || !goog.isFunction(fieldClass.fromJson)) {
    throw new Error('Field "' + fieldClass +
        '" must have a fromJson function');
  }
  Blockly.Field.TYPE_MAP_[type] = fieldClass;
};

```

No code path directly results in a call to `Blockly.FieldName.fromJson`, so maybe they're getting compressed out?

PR #1672, opened 2018-02-28
Split event files

### LINKED ISSUE #1618: Split events.js by event type, and move into a subdirectory
Adding a new event type shouldn't cause merge conflicts in `events.js`.

TODOs:
- Split `events.js` into `events.js`, `event_abstract.js`, `event_block_create.js`, etc.
- Put these files into an `events` subdirectory inside `core`
- Verify no problems with building
- Verify no problems with advanced compilation
- Consider having all variable events inherit from a single class, all block events from a single subclass, etc (since they have common fields)

rachel-fenichel:
## The basics
- [x] I branched from develop

### Resolves

Fixes #1618 

### Proposed Changes

Split up the monolithic `events.js` file.
Create a new `events` folder, containing subfolders `block`, `variable`, and `ui`.

### Reason for Changes

See #1618.

The events file grows every time we add a new event type.  This will cause merge conflicts, e.g. when scratch team makes a new event type and we make a different new event type.

I want to make it easier to add a new event type.  We may also want to consider a registration mechanism for new event types, similar to what we did with fields.

I also want to make it easier to distinguish between creation events for blocks, variables, comments, etc.

### Test Coverage
Tested by running `build.py core` and opening up the playground and the demos.  In the playground I turned events on and dragged stuff around, opened a mutator, deleted blocks, etc.
In the demos, I opened the mirror demo and confirmed that block creation, dragging, and deletion all worked.

That tested uncompressed and compressed.  I was not able to test advanced compilation because of #1671.  I want to test that before I try to land this.

### Additional Information

@NeilFraser PTAL, this is similar to what we talked about last week.


rachel-fenichel:
It's almost the same.  The important difference is that there are explicit provides and requires for the various types of events.  I do think that's a meaningful difference, though testing it is blocked by #1671.

rachel-fenichel:
I just tested under advanced compilation after applying the fix for #1671.  Things seem fine.  To test I:
- added a logger to the workspace in the advanced compilation test
- made sure there were no errors when I interacted with the workspace
- deleted every call to `goog.require('Blockly.Events.Ui');`
- recompiled
- confirmed that there were errors when I tried to interact with the workspace

To add the logger, I replaced `tests/compile/main.js` with the following code:
```
goog.provide('Main');
// Messages (in some language)
goog.require('Blockly.Msg.en');
// Core
goog.require('Blockly');
// Blocks
goog.require('Blockly.Constants.Logic');
goog.require('Blockly.Constants.Loops');
goog.require('Blockly.Constants.Math');
goog.require('Blockly.Constants.Text');

function logger(e) {
  console.log(e);
}

Main.init = function() {
  var workspace = Blockly.inject('blocklyDiv', {
    'toolbox': document.getElementById('toolbox')
  });
  Blockly.Events.enable();
  workspace.addChangeListener(logger);
};
window.addEventListener('load', Main.init);
```

PR #1680, opened 2018-03-02
Mark bindEvent_ and bindEventWithChecks_ package instead of private

rachel-fenichel:
Fixes 50 compiler warnings.


rachel-fenichel:
Fifty!

Issue #1655, opened 2018-02-22
Tracking issue for pointer events


rachel-fenichel:
Enabling pointer events is simple:  Enable the commented-out code in `touch.js`.
```
// if (window.PointerEvent) {
//   Blockly.Touch.TOUCH_MAP = {
//     'mousedown': ['pointerdown'],
//     'mousemove': ['pointermove'],
//     'mouseup': ['pointerup', 'pointercancel']
//   };
// } else
if (goog.events.BrowserFeature.TOUCH_ENABLED) {
  Blockly.Touch.TOUCH_MAP = {
    'mousedown': ['touchstart'],
    'mousemove': ['touchmove'],
    'mouseup': ['touchend', 'touchcancel']
  };
}
```

This causes a variety of subtle bugs.
- Can't edit a comment after moving it or the workspace it's on
- Double event handling in iframes (handle both mouse and pointer events on a click)
- Mouse click-and-hold is treated as a longpress on the workspace, triggering the workspace context menu.

I disabled pointer events in the feb 2018 release, pending fixes for these issues.

In theory pointer events should fully replace mouse events on any platform where they are supported.

samelhusseini:
Regarding the double event firing on iframes, we know the source of that issue, and PR https://github.com/google/blockly/pull/1615 fixes that issue. 

Regarding click and hold, I have a fix for that. Issue is we were assuming pointer events are touch events but they're not necessarily, so we need to check before we call longStart that the pointerType is not 'mouse'. I'll get a PR out for that. 

Comments: 
- using mouseup for to focus the text field (has definitely regressed, seeing the same issue here, investigating).
- there's another issue, perhaps unrelated but I'm not able to focus or type on a comment field on IOS.

samelhusseini:
Changing the event to mousedown fixes both issues. I believe the issue here is perhaps related to the TODO in gesture.js (doBubbleClick_). 

The issue is that there's an overarching pointer event listener that's the workspace gesture.js (recognizer) that's listening to clicks / drags and classifying them. The recognizer discards any event that was meant for an input field (textarea, input, etc). so when we click on the textarea, it discards the event but in the process of doing that it's eating up the event that could have been passed to the input element itself. 

Reason why this worked, the mouseup event wasn't getting triggered at all actually, it's the browser that was focusing the textarea field in the bubble, which means bubble promotion wasn't happening when a user clicks on the comment field. (You can try this in blockly/master)

[...]

PR #1686, opened 2018-03-02
Pointer Event fixes

samelhusseini:
Fix pointer event (edge case) issues described in https://github.com/google/blockly/issues/1655

## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/1655

### Proposed Changes

** Pointer events **:
A couple of major changes with the way we're handling pointer events: 
- Always use pointer events, and no other mouse event. We were using a mixture of those, since we only had 'mousedown', 'mousemove' and 'mouseup' defined in the map. I added the rest of the mouse events, there were a couple of places where we attached events to mouseover, and so I just added mapping for the suite of mouse events. I also added a couple of mappings for touchend and touchcancel (this is really a workaround) since there are a couple of events that we register only for touch (eg: longstop) that we want to register for pointer events. 

Another major change is how we register pointer events. If they're available, they should be a replacement to mouse events. and in the same way that we don't add additionally hooks to preventDefault on mouse events (only for touch), we don't need to do that for pointer events. So I've shuffled the event registration code a little so that it registers pointer events with the wrapFunc function and not the touchWrapFunc function. 
(In other cases, we register both mouse and touch, if mappings exist). 

**Comments**:
Another fix to comments was to pass opt_noPreventDefault with the registered mouseup event. This was causing the gesture code to capture the event (even though the gesture code cancels on native elements like textarea) it was still capturing the event since the event (as far as it was concerned) was "handled". This was only affecting touch events, such as IOS.

Added an optimization to the bubble logic, only bumping it if it's not currently the top bubble. This also feeds into the comment focus logic where I only re-focus the text area if it has been moved within the bubble canvas.
 
** Context menu **:
Regarding the issue with the longStop: I just added a check to ensure the pointer type is not a mouse (since pointer events can be 'mouse', 'touch' or 'pen).

### Reason for Changes

Fixes issues described in https://github.com/google/blockly/issues/1655

### Test Coverage

Tested on:
Desktop Chrome on Mac (latest)
Desktop Firefox (latest) and (beta, Firefox 59) on Mac.
Desktop Safari on Mac.
Desktop Chrome on Surface Book (Windows).
Desktop Firefox on Sufrace Book (Windows).
IE 11 on Surface Book
Edge 16 on Surface Book
IOS (Safari latest). 
Chrome Android (although there is a known issue with pointer events on Android, we might want to either exclude that specific version of Android or Android altogether from using pointer events until the next version of Chrome is released). 
Firefox Android

### Additional Information

Note, with my changes all interactions with text areas inside comments are handled completely by the browser. Blockly usually tries to limit this behaviour so that the browser doesn't do some fancy zoom, but in this case, in order to allow things like selecting parts of the text or jumping to a different section in the comment, we need to let the browser take care of that.

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -452,26 +452,32 @@ Blockly.bindEventWithChecks_ = function(node, name, thisObject, func,
   };
 
   var bindData = [];
-  // Don't register the mouse event if an equivalent pointer event is supported.
-  if ((window && !window.PointerEvent) || !(name in Blockly.Touch.TOUCH_MAP)) {
+  var bindData = [];
+  if (window && window.PointerEvent &&
+      (name in Blockly.Touch.TOUCH_MAP)) {
+      for (var i = 0, type; type = Blockly.Touch.TOUCH_MAP[name][i]; i++) {
+        node.addEventListener(type, wrapFunc, false);
+        bindData.push([node, type, wrapFunc]);
+      }
+  } else {
     node.addEventListener(name, wrapFunc, false);
     bindData.push([node, name, wrapFunc]);
-  }
 
-  // Add equivalent touch or pointer event.
-  if (name in Blockly.Touch.TOUCH_MAP) {
-    var touchWrapFunc = function(e) {
-      wrapFunc(e);
-      // Calling preventDefault stops the browser from scrolling/zooming the
-      // page.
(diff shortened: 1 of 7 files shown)


samelhusseini:
Wrong source.

PR #1698, opened 2018-03-12
Expose block ID in SVG.

NeilFraser:
Useful for debugging.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -99,6 +99,9 @@ Blockly.BlockSvg = function(workspace, prototypeName, opt_id) {
   Blockly.Tooltip.bindMouseEvents(this.svgPath_);
   Blockly.BlockSvg.superClass_.constructor.call(this,
       workspace, prototypeName, opt_id);
+
+  // Expose this block's ID on its top-level SVG group.
+  this.svgGroup_.id = this.id;
 };
 goog.inherits(Blockly.BlockSvg, Blockly.Block);
 


core/block_svg.js
```diff
@@ -99,6 +99,9 @@ Blockly.BlockSvg = function(workspace, prototypeName, opt_id) {
   Blockly.Tooltip.bindMouseEvents(this.svgPath_);
   Blockly.BlockSvg.superClass_.constructor.call(this,
       workspace, prototypeName, opt_id);
+
+  // Expose this block's ID on its top-level SVG group.
+  this.svgGroup_.id = this.id;
```
AnmAtAnm: Don't these XML ids have strong potential for duplication in the document.  Consider a page with toolbox XML embedded in the HTML. There would be two elements with the same id: the toolbox `<block>` element in the XML and this new `<svg:g>` element.

I suggest using a [data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes): `svgGroup_.dataset.blockId = this.id;`

NeilFraser: Alternatively, prefix the SVG ID with 'block'.  Then it becomes clearer when browsing the DOM what that group actually is.

NeilFraser: Here's a solution that also addresses the issue that currently there's no way at all to associate a bubble with it's block by inspecting the DOM: https://github.com/google/blockly/pull/1700

AnmAtAnm: Testing with the playground (adding the id to the a block), this still isn't sufficient. If a toolbox block has an id, than the first copy of the that block in the workspace has the same id. Thus, the prefix differentiates the `<svg:g>` from the toolbox `<block>`, but does not uniquely identify the element in the document.

Looks like you may already be making this change in #1700.

PR #1700, opened 2018-03-13
Add IDs to bubbles.

NeilFraser:
Allows one to associate comments/mutators/warnings with their blocks while inspecting the DOM.

Also adds ‘block’ prefix to prevent possible ID collision between toolbox and workspace.

### DIFF
--- a/build.py
+++ b/build.py
@@ -223,6 +223,7 @@ class Gen_compressed(threading.Thread):
         ("output_info", "warnings"),
         ("output_info", "errors"),
         ("output_info", "statistics"),
+        ("warning_level", "DEFAULT"),
       ]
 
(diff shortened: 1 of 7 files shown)


rachel-fenichel:
Add an underscore after the prefix (`block_uuid`, etc) for all of the types.  Then lgtm.

rachel-fenichel:
We agreed to do `data-id` on the blocks and `data-blockId` on the bubbles.

This is the inverse.

AnmAtAnm:
> We agreed to do data-id on the blocks and data-blockId on the bubbles.

FYI: `el.dataset.blockId === el.getAttribute('data-block-id')`. `data-blockId` should never exist.

PR #1703, opened 2018-03-15
Fix more compiler warnings

rachel-fenichel:
In the first commit:
- In `block_render_svg` we pass around arrays that we use to build up the pieces of the SVG paths.  Once we have every piece, we concatenate everything in the arrays to build a string.  The items in the array can be numbers or strings, not just strings.
- In `Blockly.utils`, `hasClass` should be package instead of private.

Gets rid of 35 compiler warnings.

In the second commit:
- The block dragger can only drag BlockSvgs, not Blocks.
- Add a getter for toolboxes on the workspace, and use it.
- Mark some stuff protected in `icon.js` so that subclasses can access them.

Gets rid of about 30 more warnings.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -582,6 +582,15 @@ Blockly.WorkspaceSvg.prototype.getFlyout_ = function() {
 };
 
 /**
+ * Getter for the toolbox associated with this workspace, if one exists.
+ * @return {Blockly.Toolbox} The toolbox on this workspace.
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.getToolbox = function() {
+  return this.toolbox_;
+};
+
+/**
  * Update items that use screen coordinate calculations
  * because something has changed (e.g. scroll position, window size).
  * @private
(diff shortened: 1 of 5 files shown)


core/workspace_svg.js
```diff
@@ -581,6 +581,15 @@ Blockly.WorkspaceSvg.prototype.getFlyout_ = function() {
...
 };
 
+/**
+ * Getter for the toolbox associated with this workspace, if one exists.
+ * @return {Blockly.Toolbox} The toolbox on this workspace.
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.getToolbox = function() {
+  return this.toolbox_;
+};
```
picklesrus: Do you plan on using this elsewhere? Curious if it is better to expose a potentially null toolbox that folks have to check for, or just add a different method for doing this.  workspace.addToolboxStyle seems a little specific though too.

rachel-fenichel: No specific plans, but I know that getting the toolbox is reasonable and it would be nice to have a getter.  getFlyout does more work to figure out which flyout it is, but is otherwise similar.

picklesrus: ah right. We already have get flyout. LGTM

PR #1706, opened 2018-03-15
Make undoStack_ and redoStack_ protected

rachel-fenichel:
Resolve compiler warnings


AnmAtAnm:
Is there specific external code that calls these stacks?

rachel-fenichel:
They're used in WorkspaceSvg.

PR #1708, opened 2018-03-15
Warnings in drag surface classes

rachel-fenichel:


core/workspace_drag_surface_svg.js
```diff
@@ -132,11 +132,14 @@ Blockly.WorkspaceDragSurfaceSvg.prototype.getSurfaceTranslation = function() {
 /**
  * Move the blockCanvas and bubbleCanvas out of the surface SVG and on to
  * newSurface.
- * @param {!SVGElement} newSurface The element to put the drag surface contents
+ * @param {SVGElement} newSurface The element to put the drag surface contents
  *     into.
  * @package
  */
 Blockly.WorkspaceDragSurfaceSvg.prototype.clearAndHide = function(newSurface) {
+  if (!newSurface) {
```
AnmAtAnm: Why did you remove the non-nullable annotation, but then add a null/undefined check that throws?

rachel-fenichel: Because the callers pass in things that can be null, such as their SVG roots.  This is just explicitly handling the case here instead of in every call.

Why throw: I don't think there's a good way to keep going without possibly losing important state.

PR #1710, opened 2018-03-16
Add hook for custom context menus on the workspace

### LINKED ISSUE #391: Hook to add items in main context menu
Simularly to the context menu of custom blocks, it would be great to be able to manipulate the `menuOptions` array before it is being sent into `show`. Or maybe I just missed how to do that!

Our particular scenario is adding another context item.

https://github.com/google/blockly/blob/abe2533d7eedbd8546d53f3f193128489a785942/core/workspace_svg.js#L852

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #391 
Needs documentation, but so does everything about context menus.

### Proposed Changes

Adds a property to the workspace for a custom context menu function.

Like the one on blocks, it's null by default.  If it's defined it is called with an array of context menu options.  The function can edit the array, including removing items and adding items.

Each item in the array is an object containing:
`enabled`: boolean
`text`: string
`callback`: function

### Reason for Changes

#391 requested it, and so have other people.

### Test Coverage

Tested in the playground by defining this function:
```
function customContextMenuFn(options) {
  var option = {
    enabled: true,
    text: "Custom option",
    callback: function() {
      console.log('Custom context menu option called');
    }
  };
  options.push(option);
}
```
then setting `workspace.configureContextMenu = customContextMenuFn;`

Tested on:
* Desktop Chrome

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -269,6 +269,13 @@ Blockly.WorkspaceSvg.prototype.flyoutButtonCallbacks_ = {};
 Blockly.WorkspaceSvg.prototype.toolboxCategoryCallbacks_ = {};
 
 /**
+ * Developers may define this function to add custom menu options to the
+ * workspace's context menu or edit the workspace-created set of menu options.
+ * @param {!Array.<!Object>} options List of menu options to add to.
+ */
+Blockly.WorkspaceSvg.prototype.configureContextMenu = null;
+
+/**
  * In a flyout, the target workspace where blocks should be placed after a drag.
  * Otherwise null.
  * @type {?Blockly.WorkspaceSvg}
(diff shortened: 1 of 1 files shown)


core/workspace_svg.js
```diff
@@ -268,6 +268,13 @@ Blockly.WorkspaceSvg.prototype.flyoutButtonCallbacks_ = {};
  */
 Blockly.WorkspaceSvg.prototype.toolboxCategoryCallbacks_ = {};
 
+/**
+ * Developers may define this function to add custom menu options to the
+ * workspace's context menu or edit the workspace-created set of menu options.
+ * @param {!Array.<!Object>} options List of menu options to add to.
+ */
+Blockly.WorkspaceSvg.prototype.customContextMenu = null;
```
AnmAtAnm: CustomContextMenu is a noun phrase. As a function, I'd expect a verb phrase. configureWorkspaceMenu()?

rachel-fenichel: How about configureContextMenu, or configureContextMenuFn?

I was matching the API that we have for blocks, but I agree that it's not a great name.

alitaheri:
I had to dig a little deep to find out how this hook can be used. Might be a good idea to add a section in the documentation. The content of the array as well as the props on the array elements should be documented. Thanks 🙏 🙏 

For now a simple link [here](https://developers.google.com/blockly/reference/js/Blockly.WorkspaceSvg#configureContextMenu) to this particular issue will do good. The description is enough documentation. but I had to search through the issues to find this one.

Issue #1663, opened 2018-02-27
Workspace comments: right-click


rachel-fenichel:
### Problem statement

Workspace comments as implemented in `comments_v2` have two modes: focused and blurred.  

If a comment is focused, its text can be edited.  Right-clicking gives the standard text-editing right click menu.  Dragging on the comment selects text.  All of this is already implemented.

If a comment is blurred (not focused), its text cannot be edited.  Right-clicking should give a comment-specific right-click menu.  At minimum it should contain "Delete comment" and "Duplicate comment".  Dragging on the comment drags it around the workspace.  

### Expected Behavior

If a comment is blurred (not focused), its text cannot be edited.  
- Right-clicking should give a comment-specific right-click menu.  
- At minimum it should contain "Delete comment" and "Duplicate comment".  

### Actual Behavior

Right-clicking on a blurred comment goes straight through to the workspace.

### Steps to Reproduce

1. Check out comments_v2
2. Open the playground
3. Right click
4. Select "Add comment"
5. Unfocus the comment (drag the workspace, etc.)
6. Right-click on the comment

### Additional Information

@samelhusseini FYI

PR #1718, opened 2018-03-20
Right click support for workspace comments

samelhusseini:
Add support for a context menu on workspace comment showing delete and duplication options.
Added copy and paste support for workspace comments. 

PR into comments_v2 branch.

Resolves https://github.com/google/blockly/issues/1663

duplicateremovecomment

Notes: 
- Added an isComment attribute to workspace comment to allow us to differentiate between a WS comment and a block object
- Added showContextMenu_ method on bubble which handles context menu, only handling this in the workspace comment object right now, which means that we can lose right clicks on other bubbles (block comments for example, if you right click on the border)
- There's a TODO in there for propper bump login when duplicating a comment, I've just bumped it 50x units and 50y units.


core/blockly.js
```diff
@@ -252,7 +252,7 @@ Blockly.onKeyDown_ = function(e) {
  * @private
  */
 Blockly.copy_ = function(block) {
-  var xmlBlock = Blockly.Xml.blockToDom(block);
+  var xmlBlock = (block.isComment) ? block.toXmlWithXY() : Blockly.Xml.blockToDom(block);
```
rachel-fenichel: Update the parameter annotation so that this can be a block | workspaceComment.  Rename the param to `toCopy` to indicate that it's required to be a block.  Similarly, `xmlBlock` isn't guaranteed to represent a block anymore.

What does `deleteNext` do if you have a comment instead of a block?

rachel-fenichel: Same thing for `Blockly.duplicate_`, actually.

core/workspace_svg.js
```diff
@@ -991,6 +1004,36 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
   block.select();
 };
 
+/**
+ * Paste the provided block onto the workspace.
```
rachel-fenichel: Provided comment.

core/workspace_svg.js
```diff
@@ -991,6 +1004,36 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
+    var comment = Blockly.WorkspaceCommentSvg.fromXml(xmlComment, this);
+    // Move the duplicate to original position.
+    var commentX = parseInt(xmlComment.getAttribute('x'), 10);
+    var commentY = parseInt(xmlComment.getAttribute('y'), 10);
+    if (!isNaN(commentX) && !isNaN(commentY)) {
+      if (this.RTL) {
+        commentX = -commentX;
+      }
+      // Offset workspace comment.
+      // TODO: properly offset comment such that it's not interfereing with any blocks
```
rachel-fenichel: Make an issue for this and add a reference.  I'll add it to the project on github.  That way we don't lose track of it.

core/events/workspace_comment/create.js
```diff
@@ -0,0 +1,93 @@
+/**
```
rachel-fenichel: I'm also working on events right now.  Your code structure is similar to mine, though I created a base class for comment events.  I'm going to continue working on the other event types, and i'll handle any merge conflicts that result.

core/events/abstract.js
```diff
@@ -71,7 +71,7 @@ Blockly.Events.Abstract = function(elem) {
    */
   this.recordUndo = undefined;
 
-  if (elem instanceof Blockly.Block) {
+  if (elem instanceof Blockly.Block || elem instanceof Blockly.WorkspaceComment) {
```
rachel-fenichel: If it's a workspace comment save the id in `this.commentId` instead.  That way if we have comments on blocks (with similar events) we can store the block id correctly.

core/blockly.js
samelhusseini: deleteNext does nothing, since there's no next block.

core/events/workspace_comment/create.js
samelhusseini: Sounds good :)

core/blockly.js
```diff
@@ -248,34 +248,35 @@ Blockly.onKeyDown_ = function(e) {
...
-  Blockly.Xml.deleteNext(xmlBlock);
+  Blockly.Xml.deleteNext(xml);
   // Encode start position in XML.
-  var xy = block.getRelativeToSurfaceXY();
-  xmlBlock.setAttribute('x', block.RTL ? -xy.x : xy.x);
-  xmlBlock.setAttribute('y', xy.y);
-  Blockly.clipboardXml_ = xmlBlock;
-  Blockly.clipboardSource_ = block.workspace;
+  var xy = toCopy.getRelativeToSurfaceXY();
+  xml.setAttribute('x', toCopy.RTL ? -xy.x : xy.x);
```
rachel-fenichel: `toXmlWithXY` already does this for a comment.  Let's do

```
if (toCopy.isComment) {
  var xml = toCopy.toXmlWithXy();
} else {
  var xml = Blockly.Xml.blockToDom(toCopy);
  // Copy only the selected block and internal blocks.
  Blockly.Xml.deleteNext(xml);
  // Encode start position in XML.
  var xy = toCopy.getRelativeToSurfaceXY();
  xml.setAttribute('x', toCopy.RTL ? -xy.x : xy.x);
  xml.setAttribute('y', xy.y);
}

Blockly.clipboardXml_ = xml;
Blockly.clipboardSource_ = toCopy.workspace;
```

As a bonus, now we're not calling `deleteNext` on a WS comment, which didn't really make sense anyway.  (But worked because JS).

samelhusseini:
Yep, only shows the Blockly context menu when not selected, or when you right click on the bubble in both modes (and with my other PR, it would be on the top handle as well).
When in editing mode, you get the browser context menu that shows on a textarea.

samelhusseini:
Not too certain why the tests are failing. I tried running npm test locally and all passed on OSX, but 2 seem to be failing on travis. Have you run into this before?

rachel-fenichel:
If it's still failing after the next run, I'll download this and take a look.

core/events/workspace_comment/create.js
rachel-fenichel: Looking at this again, I think it's best if you remove this file + references from this PR.  I'll be landing #1723 tomorrow, barring surprises.  No point in landing two versions of the code and then deleting one.

samelhusseini: Makes sense. I'll get rid of it.

rachel-fenichel:
Aha, it's failing on Travis but not locally because you added a new require/provide pair and didn't push a new version of blockly_uncompressed.js.  

When I pulled but didn't build I saw this:
image

Building made it go away.

You're getting rid of that file anyway, so it should fix itself.

core/events/workspace_comment/create.js
samelhusseini: Done. Added a placeholder TODO to fire the event.

Issue #1662, opened 2018-02-27
Workspace comments: deletion


rachel-fenichel:
### Problem statement

Workspace comments in the `comments_v2` branch don't have a deletion mechanism in UI.

### Expected Behavior
Workspace comments should have an 'X' on the top right corner (top left in RTL).  Clicking on this X should delete the comment.

### Actual Behavior
There is no deletion UI.

### Steps to Reproduce

1. Check out `comments_v2`
2. Open the playground
3. Right click
4. Select "Add comment"
5. Try to delete the comment

### Additional Information

@samelhusseini FYI

rachel-fenichel:
Workspace comments should also be deleted when a drag ends over the toolbox/flyout or the trash can.

rachel-fenichel:
Closed by #1720

Issue #1667, opened 2018-02-27
Discussion: Possible values of Blockly.selected


rachel-fenichel:
As of Feb 2018, `Blockly.selected` can only be the currently selected **block**.

With the workspace comments we set `Blockly.selected` to be the currently focused/selected comment.  By extension it should be possible to select a block comment or warning, and possibly even the flyout or toolbox category.  When moving within a block, there could be a selected field or connection.

For accessibility and keyboard navigation we probably want some concept of "selected" that encompasses all of these things.

Questions:
- Would overloading `Blockly.selected` in this way cause problems?  e.g. if someone's code assumes that if `Blockly.selected` is set, it's a `Blockly.Block`.
- @quachtina96 do you have thoughts on accessibility?
- How does this play into keyboard navigation?
- Is it worth keeping track of a selected block separately from anything else that might be selected?
- Are there categories of selectable things?  e.g. blocks, block comments and workspace comments are mutually exclusive, but a textarea within a block or a comment may be selected at the same time as that block or comment.
- How does this interact with events?

samelhusseini:
For selection, I would say that Blockly's selection handling should be separate from the toolbox. As it's not a requirement to use Blockly's toolbox, as developers can implement their own custom toolbox.

I'm not sure if there's some other meaning for selection, but my understanding of selection is that there's a visual indicator in Blockly (not in scratch-blocks), when a block is selected. It's the last Block that a user clicked on. Perhaps in the future it would be the block that would move if you were to use the arrow keys (for accessibility). So with that assumption in mind, I'd say that selecting a block is no different to any other entity that can be moved around with arrow keys on the workspace. (ie: any other entity that lives on the workspace). That includes both workspace and block comments.

As an extension of that, the selected visual indicator is an indicator that the Block can be moved. (and I just checked, and we don't show a selected indicator for read only, so that's in alignment with my reasoning).

Perhaps down the line, all selectable objects (blocks, comments, warnings) should extend from one abstract class, and that is the type of Blockly.selected.

0199prabha:
I want help with my blockly code. how to remove all dragged blocks by clicking the reset button. my code only removes the child block.
Here's my code:
function reset() {
  if (Blockly.selected && !Blockly.selected.workspace.isFlyout) {
    Blockly.Events.setGroup(true);
    Blockly.hideChaff();
    Blockly.selected.dispose(false);
    Blockly.Events.setGroup(false);
}
  delete inputTextValue;
  redrawUi();
}

[...]

PR #1720, opened 2018-03-21
Workspace comments: delete targets and icon

samelhusseini:
Show an (X) icon on the comment when selected. Delete icon deletes the comment. Comment can be deleted if dragged onto the toolbox or the trash icon. A normal bubble cannot be deleted that way.

Branched of comments_v2

Resolves https://github.com/google/blockly/issues/1662

deletecomment

Notes: 
- Only showing the x icon when the comment is selected. (otherwise there was too much clutter)

### DIFF
--- a/core/bubble_dragger.js
+++ b/core/bubble_dragger.js
@@ -113,6 +113,13 @@ Blockly.BubbleDragger.prototype.startBubbleDrag = function() {
   }
 
   this.draggingBubble_.setDragging && this.draggingBubble_.setDragging(true);
+
+  var toolbox = this.workspace_.getToolbox();
+  if (toolbox) {
+    var style = this.draggingBubble_.isDeletable() ? 'blocklyToolboxDelete' :
+        'blocklyToolboxGrab';
+    toolbox.addStyle(style);
+  }
 };
 
 /**
@@ -209,6 +216,11 @@ Blockly.BubbleDragger.prototype.endBubbleDrag = function(
   }
   this.workspace_.setResizesEnabled(true);
 
+  if (this.workspace_.toolbox_) {
+    var style = this.draggingBubble_.isDeletable() ? 'blocklyToolboxDelete' :
+        'blocklyToolboxGrab';
+    this.workspace_.toolbox_.removeStyle(style);
+  }
   Blockly.Events.setGroup(false);
 };
 
--- a/core/bubble_dragger.js
+++ b/core/bubble_dragger.js
@@ -53,6 +53,22 @@ Blockly.BubbleDragger = function(bubble, workspace) {
   this.workspace_ = workspace;
 
   /**
+   * Which delete area the mouse pointer is over, if any.
+   * One of {@link Blockly.DELETE_AREA_TRASH},
+   * {@link Blockly.DELETE_AREA_TOOLBOX}, or {@link Blockly.DELETE_AREA_NONE}.
+   * @type {?number}
+   * @private
+   */
+  this.deleteArea_ = null;
+
+  /**
+   * Whether the bubble would be deleted if dropped immediately.
+   * @type {boolean}
+   * @private
+   */
+  this.wouldDeleteBubble_ = false;
+
+  /**
    * The location of the top left corner of the dragging bubble's body at the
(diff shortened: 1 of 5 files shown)


rachel-fenichel:
In no particular order:

I really like only showing the icon when the comment is selected.  It's also a nice way to show that the comment is currently active.

Needs to show the delete cursor when over the trashcan or toolbox.  I think that'll involve some digging in workspace_svg.js (unfortunately).  If you want me to handle that, file a bug and add a reference to it..

core/bubble_dragger.js
```diff
@@ -95,6 +111,8 @@ Blockly.BubbleDragger.prototype.startBubbleDrag = function() {
   if (this.dragSurface_) {
     this.moveToDragSurface_();
   }
+
+  this.draggingBubble_.setDragging && this.draggingBubble_.setDragging(true);
```
rachel-fenichel: This line isn't specific to deletion.  Was it just missing before?

samelhusseini: Yeah it's missing from the bubble class. Should I add it there instead?

samelhusseini:
It's actually really easy to add the delete cursor on there. But I didn't feel the hand delete was appropriate because you're not dragging and deleting (which is what it's used for today). 
There's also another cursor supported called not-allowed if you want to check that one out, but I wasn't a huge fan either. 
What do you think?

core/bubble_dragger.js
rachel-fenichel: No, this is fine.  Just wanted to make sure I understood.

rachel-fenichel:
It should have the same cursor as blocks do for deletion, to reinforce that it's going to disappear when you let go.  

What do you mean by "you're not dragging and deleting"?

samelhusseini:
The hand delete icon looks like a grabbing hand with a delete icon, so to me it looks like you're dragging something and it will delete if you let go. But i'm not too fussed on it, that's fine if that's what exists today. 

I think you raise a good point though about deleting only when you let go, which right now it's on mousedown, so I'll change that.

rachel-fenichel:
Oh I see, you're talking about the cursor on the icon.  I'm fine with leaving that alone.

I'm talking about the cursor during a drag, when you drag the block over the trash or toolbox.

samelhusseini:
Oh I just re-read your comment. Yes, so I did do that. Not reflected for some reason in my GIF (maybe it was a little old). It works on the trash icon, but not on the toolbox. I couldn't figure out why, even though I'm doing the same thing we do for blocks. so I'll file a bug to track.

samelhusseini:
Also highlighting on mouse down, only deleting on mouse out. 

deleteiconhighlight

samelhusseini:
I also noticed something, when we create a new comment, we currently select but not focus it which puts it in this weird state that you only get to when you first create it. 
We have a few options here, but what are your thoughts on what "selected" should mean? 
Should it be the same as focused? But that might be hard to enforce, as it's possible to click on the handle and go out of the text area focus but still be in a "selected" state. Also if we focus on select it might make it confusing if the first thing users do is move the comment to the right spot when they create it (because the focus would be inside the text area and you'd only be able to move it from the handle then). 
Thoughts?

rachel-fenichel:
For that particular issue, I think that creating a new comment should focus it for immediate editing.  If a user wants to edit immediately, it'll be possible.  If they want to move it immediately, they can grab the top bar.  If we play-test and find that it's an issue we can change it or try to make the top bar more prominent during editing.

For "selected" generally: yeah, I have an [issue](https://github.com/google/blockly/issues/1667) with my thoughts, because I'm not sure that Blockly.selected should be allowed to be a comment.  Can you go read that and put down any reactions you have?

If we do allowed Blockly.selected to be a comment, I think that if the comment is focused it must also be selected, but that it's allowed to be selected but not focused.  That means that it'll be selected at the end of a drag, even though it's not focused for editing.

rachel-fenichel:
LGTM, I'll merge it after lunch unless you have additional changes by then.

samelhusseini:
Okay I'm focusing on selection. PR is good to go.

PR #1723, opened 2018-03-21
Feature/comment events

rachel-fenichel:
Starting work on events for workspace comments.

This PR adds a new category of events (comment) and a set of classes within that category: base, change, create, delete, and move.

I've fleshed out `CommentCreate` and added uses of it, but no others.


rachel-fenichel:
@samelhusseini FYI

rachel-fenichel:
Locations are not being preserved across undo/redo combinations.  This mostly points at failures in the XML serialization deserialization ordering.

rachel-fenichel:
Can you elaborate on your second question?  I'm not sure I understand.

First question: that's reasonable.  I'm going to do it in a separate PR, and also make a note in my doc about how to add new event types.

Steps that I'll take in that other PR:
- Create a new base class for block events
- Make all block events extend the new base class
- Make the constructor in the new class set blockId
- Possibly move other duplicated code into the base class, as needed.
- Repeat for variable events.

rachel-fenichel:
https://github.com/google/blockly/pull/1729 is the new PR for setting up the base classes.

picklesrus:
I'm just wondering if it was a conscious decision that there are going to be 3 types of create events
Blockly.Events.VarCreate
Blockly.Events.Create
Blocky.Events.CommentCreate
vs.
one
Blockly.Events.Create  with information inside of the json to distinguish between blocks, variables, comments, etc.

rachel-fenichel:
I see.  No, that wasn't a conscious decision.  I'll think about whether I prefer one or the other.

Issue #1741, opened 2018-03-28
Blocks with variables shouldn't show "create var getter" option in context menu when in flyout


samelhusseini:
### Problem statement

Blocks with variables as inputs shouldn't show the create variable getter in their context menus when inside a flyout. (Just like variable getter / setter blocks)

![creategetvariable](https://user-images.githubusercontent.com/16690124/38012442-5b4066da-3216-11e8-8cd0-46592903e45d.gif)

<img width="396" alt="screen shot 2018-03-27 at 11 27 46 pm" src="https://user-images.githubusercontent.com/16690124/38012470-7a40ad10-3216-11e8-8e2b-77a11dc855f4.png">
<img width="197" alt="screen shot 2018-03-27 at 11 25 20 pm" src="https://user-images.githubusercontent.com/16690124/38012461-70434d4a-3216-11e8-93d8-4e3a4fb281c9.png">

### Expected Behavior

Context menu doesn't show the create getter option when inside the flyout.

### Actual Behavior

Context menu does show the option, and when clicking on the option, the getter is created and orphaned inside the flyout workspace. 

### Steps to Reproduce

1. Start by opening the Loops category in the Blockly playground.
2. Next, right click on the for block and select the 'Create "get VARIABLE_ID"' block (variable id should be variable name, see https://github.com/google/blockly/issues/1740)
3. Click on that option. (A new variable_get block is created but it is orphaned inside of the flyout workspace)

### Operating System and Browser

Tested on Chrome. 
Should repro on all.

samelhusseini:
Same issue with functions and their 'create function' option.

samelhusseini:
This is fixed.

PR #1742, opened 2018-03-28
Don't show create VAR and create Function context menu options when in the flyout

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/1741

### Proposed Changes

Don't show the 'create get VAR' and 'create Function' context menu options on blocks when in the flyout.

### Reason for Changes

Fixes https://github.com/google/blockly/issues/1741

### Test Coverage

Tested by right clicking on the for, count with, and function blocks in the flyout. Also tested that the option works when not in the flyout.

Tested on:
* Desktop Chrome.


rachel-fenichel:
Passing on the first build, and failing on macOS because Travis is unstable again.

PR #1762, opened 2018-04-04
Strip coordinates from storage of single block stacks in demos.

NeilFraser:
When loading, place blocks with no coordinates at 10,10 instead of 0,0.

### DIFF
--- a/appengine/storage.js
+++ b/appengine/storage.js
@@ -71,6 +71,15 @@ BlocklyStorage.restoreBlocks = function(opt_workspace) {
 BlocklyStorage.link = function(opt_workspace) {
   var workspace = opt_workspace || Blockly.getMainWorkspace();
   var xml = Blockly.Xml.workspaceToDom(workspace, true);
+  // Remove x/y coordinates from XML if there's only one block stack.
+  // There's no reason to store this, removing it helps with anonymity.
+  if (workspace.getTopBlocks(false).length == 1 && xml.querySelector) {
+    var block = xml.querySelector('block');
+    if (block) {
+      block.removeAttribute('x');
+      block.removeAttribute('y');
+    }
+  }
   var data = Blockly.Xml.domToText(xml);
   BlocklyStorage.makeRequest_('/storage', 'xml', data, workspace);
 };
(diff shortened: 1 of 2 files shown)


core/xml.js
```diff
@@ -407,8 +407,10 @@ Blockly.Xml.domToWorkspace = function(xml, workspace) {
         // to be moved to a nested destination in the next operation.
         var block = Blockly.Xml.domToBlock(xmlChild, workspace);
         newBlockIds.push(block.id);
-        var blockX = parseInt(xmlChild.getAttribute('x'), 10);
-        var blockY = parseInt(xmlChild.getAttribute('y'), 10);
+        var blockX = xmlChild.hasAttribute('x') ?
+            parseInt(xmlChild.getAttribute('x'), 10) : 10;
```
NeilFraser: This has been checked in RTL and it does the right thing (10 from the right).

appengine/storage.js
```diff
@@ -71,6 +71,15 @@ BlocklyStorage.restoreBlocks = function(opt_workspace) {
 BlocklyStorage.link = function(opt_workspace) {
   var workspace = opt_workspace || Blockly.getMainWorkspace();
   var xml = Blockly.Xml.workspaceToDom(workspace, true);
+  // Remove x/y coordinates from XML if there's only one block stack.
+  // There's no reason to store this, removing it helps with anonymity.
+  if (workspace.getTopBlocks(false).length == 1 && xml.querySelector) {
```
NeilFraser: I don't think the xml.querySelector is needed, but I'd rather add it defensively than have some browser (MSIE) blow up because it doesn't treat XML as a document.

appengine/storage.js
```diff
@@ -71,6 +71,15 @@ BlocklyStorage.restoreBlocks = function(opt_workspace) {
...
+    var block = xml.querySelector('block');
+    if (block) {
```
NeilFraser: The existence of a block would appear to be guaranteed (if workspace has one top block), but technically it might be a shadow.  Thus it's better to check than to blow up.  I don't consider it worth adding code to strip coordinates from a single top-level shadow block.

PR #1768, opened 2018-04-09
Fixing issue 1760: nulls in JSON array for block definitions.

### LINKED ISSUE #1760: Blockly.defineBlocksWithJsonArray should handle a null element in the array.
### Problem statement

`defineBlocksWithJsonArray` fails silently on a null element, by ending the loop.

### Expected Behavior

`defineBlocksWithJsonArray` defines all blocks in the passed in array, or provides an error if it can't.

TODOs:
- Change the annotation to allow null elements in the array.
- Change the for loop to check the length of the array.
- Add a warning if an element in the array was null.

Relevant code:
```
/**
 * Define blocks from an array of JSON block definitions, as might be generated
 * by the Blockly Developer Tools.
 * @param {!Array.<!Object>} jsonArray An array of JSON block definitions.
 */
Blockly.defineBlocksWithJsonArray = function(jsonArray) {
  for (var i = 0, elem; elem = jsonArray[i]; i++) {
    var typename = elem.type;
    if (typename == null || typename === '') {
      console.warn('Block definition #' + i +
        ' in JSON array is missing a type attribute. Skipping.');
    } else {
      if (Blockly.Blocks[typename]) {
        console.warn('Block definition #' + i +
          ' in JSON array overwrites prior definition of "' + typename + '".');
      }
      Blockly.Blocks[typename] = {
        init: Blockly.jsonInitFactory_(elem)
      };
    }
  }
};
```

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

#1760

### Proposed Changes

JSON array with null or undefined value will now skip the offending item, and proceed to load following items.

Added tests for null and undefined array elements.

Also testing the handling of block definitions with null or undefined type `id`.

### Reason for Changes

Fixing the silent failure when loading a JSON array with a null or undefined in it.

### Test Coverage

Added two new tests to cover this case and the related undefined case.

Tested on:
 * Desktop Chrome

### DIFF
--- a/tests/jsunit/json_test.js
+++ b/tests/jsunit/json_test.js
@@ -26,12 +26,17 @@ function test_json_minimal() {
   var workspace = new Blockly.Workspace();
   var block;
   try {
-    Blockly.defineBlocksWithJsonArray([{
-      "type": BLOCK_TYPE
-    }]);
+    var warnings = captureWarnings(function() {
+      Blockly.defineBlocksWithJsonArray([{
+        "type": BLOCK_TYPE
+      }]);
+      block = new Blockly.Block(workspace, BLOCK_TYPE);
+    });
 
-    block = new Blockly.Block(workspace, BLOCK_TYPE);
     assertEquals(BLOCK_TYPE, block.type);
+    assertEquals(
+        'Expecting no warnings when defining and creating a simple block.',
+        warnings.length, 0);
     // TODO: asserts
   } finally {
     block.dispose();


tests/jsunit/json_test.js
```diff
@@ -257,3 +257,57 @@ function test_json_dropdown_image() {
     delete Blockly.Msg['ALTTEXT'];
   }
 }
+
+function test_defineBlocksWithJsonArray_nullItem() {
```
picklesrus: if there isn't a test where the typename is empty  or null (didn't see one in the immediate vicinity of this test... but also didn't look very hard), would you add one?

core/blockly.js
```diff
@@ -392,21 +392,28 @@ Blockly.jsonInitFactory_ = function(jsonDef) {
  * @param {!Array.<!Object>} jsonArray An array of JSON block definitions.
  */
 Blockly.defineBlocksWithJsonArray = function(jsonArray) {
-  for (var i = 0, elem; elem = jsonArray[i]; i++) {
-    var typename = elem.type;
-    if (typename == null || typename === '') {
+  for (var i = 0; i < jsonArray.length; i++) {
```
picklesrus: ooc, why change the way the loop is defined?

rachel-fenichel: Defining it the other way means it ends when it finds a null elem, even if it's not at the end of the actual list.

picklesrus: where's the facepalm emoji when you need it?

tests/jsunit/json_test.js
AnmAtAnm: Added new test towards the top.

Issue #1759, opened 2018-04-03
Mutator dialog UI with empty sub-block list causes issues


Marclev78:
### Mutator dialog with empty flyout  causes exception and rendering issues

I am trying to define a mutator dialog that only has a single top level block, for example if I can configure the block using check boxes on that top level block, so don't need any others for the dialog.

In JavaScript, I try to initialise such a mutator like this:
```
Blockly.Extensions.registerMutator("precondition_configuration", {
  mutationToDom: function () {
    ...
  },

  domToMutation: function (xmlElement) {
    ...
  },
  compose: function (topBlock) {
    ...
  },

  decompose: function (workspace) {
    const topBlock = workspace.newBlock("mutator-config-block");

    topBlock.initSvg();
    topBlock.render();

    return topBlock;
  }
}, 
() => {
}, []); // <== Empty list of blocks for mutator toolbox
```

I would expect the resulting mutator dialog to just show me the top block "`mutator-config-block` without any flyout (or an empty toolbox, if necessary).

### Actual Behavior

Opening the mutator dialog causes this exception:

```
blockly_compressed.js:1227 Uncaught TypeError: Cannot read property 'contentHeight' of null
    at Blockly.Mutator.resizeBubble_ (blockly_compressed.js:1227)
    at Blockly.Mutator.setVisible (blockly_compressed.js:1230)
    at Blockly.Mutator.Blockly.Icon.iconClick_ (blockly_compressed.js:948)
    at Blockly.Mutator.iconClick_ (blockly_compressed.js:1223)
    at SVGGElement.h (blockly_compressed.js:1631)
```

A mutator dialog does then appear on screen, however it extends to the edges of the workspace (although it does contain the single top block as expected).

I have to put an artificial block into the mutator toolbox to work around this, which is both ugly, and causes user confusion as they don't know what to do with the artificial block.

AnmAtAnm:
@Marclev78 The toolbox area is also the area used to delete mutator blocks. Are you intending to introduce other ways of creating blocks that may need to be deleted?

rachel-fenichel:
This is a valid use case but high effort for us to fix.  We will keep the issue open so that we can find out if more people are running into it.

nederland074:
@Marclev78 Hi, I've also struggled with this problem recently and found a workaround, which is to translate this artificial block beyond the borders of the mutator dialog window, something like this:
```
Blockly.Blocks['dummyBlock'] = {
    init: function() {
[...]

PR #1774, opened 2018-04-11
Work around a problem with RTL mutators

### LINKED ISSUE #1425: Delete areas broken in RTL Mutators on FF
Demo:
https://www.youtube.com/watch?v=DIdfREakzpg

Try here:
https://blockly-demo.appspot.com/static/demos/rtl/index.html

rachel-fenichel:
## The basics
- [x] I branched from develop

### Resolves

Fixes #1425 
### Proposed Changes

Change the calculation of the delete area for flyouts in mutators in RTL in firefox.  Explicitly add an offset from the left side of the mutator.

### Reason for Changes

See #1425, but TLDR: Firefox is reporting the wrong value, and that bug has been around for years.  We surfaced it with a combination of changes to our DOM structure and changes to delete logic.

### Test Coverage

Tested in the playground in Firefox in RTL mode, with the if/else mutator.
Tested at a variety of scales.
Also tested in LTR mode to make sure there are no regressions.

Tested in Chrome to confirm no regressions.

Tested on:
* Desktop Chrome
* Desktop Firefox

### Additional Information

As soon as we merge this they're going to fix Firefox and this will all break again :(

This should unblock the Blockly Games release.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -155,8 +155,13 @@ Blockly.VerticalFlyout.prototype.position = function() {
   var x = targetWorkspaceMetrics.absoluteLeft;
   if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_RIGHT) {
     x += (targetWorkspaceMetrics.viewWidth - this.width_);
+    // Save the location of the left edge of the flyout, for use when firefox
+    // gets the bounding client rect wrong.
+    this.leftEdge_ = x;
   }
   this.positionAt_(this.width_, this.height_, x, y);
+
+
 };
 
 /**
@@ -315,6 +320,18 @@ Blockly.VerticalFlyout.prototype.getClientRect = function() {
     return new goog.math.Rect(x - BIG_NUM, -BIG_NUM, BIG_NUM + width,
         BIG_NUM * 2);
   } else {  // Right
+    // Firefox sometimes reports the wrong value for the client rect.
+    // See https://github.com/google/blockly/issues/1425 and
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1066435
+    if (goog.userAgent.GECKO &&
+        this.targetWorkspace_ && this.targetWorkspace_.isMutator) {
+      // If we're in a mutator, its scale is always 1, purely because of some
+      // oddities in our rendering optimizations.  The actual scale is the same
+      // as the scale on the parent workspace.
+      var scale = this.targetWorkspace_.options.parentWorkspace.scale;
+      x += this.leftEdge_ * scale;
+
+    }
     return new goog.math.Rect(x, -BIG_NUM, BIG_NUM + width, BIG_NUM * 2);
   }
 };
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Updated to use the browser's value if it looks reasonable.  This should protect us if Firefox eventually fixes the bug.

core/flyout_vertical.js
```diff
@@ -155,8 +155,13 @@ Blockly.VerticalFlyout.prototype.position = function() {
   var x = targetWorkspaceMetrics.absoluteLeft;
   if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_RIGHT) {
     x += (targetWorkspaceMetrics.viewWidth - this.width_);
+    // Save the location of the left edge of the flyout, for use when firefox
+    // gets the bounding client rect wrong.
+    this.leftEdge_ = x;
   }
   this.positionAt_(this.width_, this.height_, x, y);
+
+
```
AnmAtAnm: NIT: Excess newlines.

core/flyout_vertical.js
```diff
@@ -155,8 +155,13 @@ Blockly.VerticalFlyout.prototype.position = function() {
   var x = targetWorkspaceMetrics.absoluteLeft;
   if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_RIGHT) {
     x += (targetWorkspaceMetrics.viewWidth - this.width_);
+    // Save the location of the left edge of the flyout, for use when firefox
```
NeilFraser: Firefox is usually capitalized.

core/flyout_vertical.js
NeilFraser: Adding blank lines?

rachel-fenichel: Done.

rachel-fenichel:
Yeah, I'm not at all happy with it but it's been blocking for too long, and we *do* know exactly what conditions trigger it.

We may eventually fix the root cause the next time we work on the general DOM structure, but that's not going to be soon.

rachel-fenichel:
Updated the math to check whether the browser's value is right, and added a comment to explain what's going on. Thanks for the check @picklesrus!

Tested in Chrome by disabling the user-agent check.  In Chrome on RTL (when the browser isn't lying) it does the right thing as well.

core/flyout_vertical.js
```diff
@@ -323,15 +323,23 @@ Blockly.VerticalFlyout.prototype.getClientRect = function() {
...
+      var targetWsLeftPixels =
+          this.targetWorkspace_.svgGroup_.getBoundingClientRect().x;
+      // The client rect is in pixels relative to the window origin.  When the
+      // browser gets the wrong value it reports that the flyout left is the
+      // same as the mutator workspace left.
+      // We know that in a mutator workspace with the flyout on the right, the
+      // visible area of the workspace should be more than ten pixels wide.  If
+      // the browser reports that the flyout is within ten pixels of the left
+      // side of the workspace, ignore it and manually calculate the value.
+      if (Math.abs(targetWsLeftPixels - x) < 10) {
```
AnmAtAnm: Will the 10px assumption be invalidated if we allow mutators without flyouts/with empty flyouts, à la #1759. In practice, a mutator with an empty/invisible empty toolbox (no way to add new blocks), probably doesn't a need for a block delete area.  Technically, the issues are distinct.

rachel-fenichel: At that point there should be no flyout, so nothing to call `getClientRect` on.  Then this code won't be hit.

PR #1776, opened 2018-04-11
Remove rogue line breaks in translated messages.

NeilFraser:


rachel-fenichel:
image

AnmAtAnm:
Do we need to correct TranslateWiki upstream?

NeilFraser:
I've no idea.  Suggest we just watch what happens in the next couple of pushes from them.  If they revert it, then yes.

rachel-fenichel:
I manually edited all three.

PR #1786, opened 2018-04-16
Add 'ordered' option to descendant getting functions.

NeilFraser:
workspace.getTopBlocks has an ordered option, extend this option to workspace.getAllBlocks, block.getChildren, and block.getDescendants.

### DIFF
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -215,11 +215,11 @@ Blockly.ContextMenu.callbackFactory = function(block, xml) {
 Blockly.ContextMenu.blockDeleteOption = function(block) {
   // Option to delete this block but not blocks lower in the stack.
   // Count the number of blocks that are nested in this block.
-  var descendantCount = block.getDescendants(true).length;
+  var descendantCount = block.getDescendants(false).length;
   var nextBlock = block.getNextBlock();
   if (nextBlock) {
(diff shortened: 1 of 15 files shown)


core/contextmenu.js
```diff
@@ -215,11 +215,11 @@ Blockly.ContextMenu.callbackFactory = function(block, xml) {
 Blockly.ContextMenu.blockDeleteOption = function(block) {
   // Option to delete this block but not blocks lower in the stack.
   // Count the number of blocks that are nested in this block.
-  var descendantCount = block.getDescendants(true).length;
```
NeilFraser: This boolean previously did nothing and is just a mistake.  Same below.  This change probably fixes a couple of compiler warnings.

core/generator.js
```diff
@@ -142,7 +142,7 @@ Blockly.Generator.prototype.prefixLines = function(text, prefix) {
  */
 Blockly.Generator.prototype.allNestedComments = function(block) {
   var comments = [];
-  var blocks = block.getDescendants();
```
NeilFraser: This is actually a bug.  Comments used to be out of order.

AnmAtAnm:
LGTM, to the extent we are okay with adding the boolean parameter to a previously public function. Since the false maps to the prior behavior, I don't see a problem with it.

PR #1787, opened 2018-04-16
Move block animation code to a new file and rebuild

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

None
### Proposed Changes

Move code for block animations to a new file.

### Reason for Changes

Simplify merges with scratch-blocks.  Generally make it easier for forks to override our animations without merge problems.

### Test Coverage
Rebuilt, then checked that dispose, connection, and disconnect animations all play properly in the playground (uncompressed) and code demo (compressed).

### Additional Information

See https://github.com/LLK/scratch-blocks/pull/1449


rachel-fenichel:
I don't think this adds anything new, because everything is package or private.

core/block_animations.js
```diff
@@ -0,0 +1,213 @@
...
+ * Play some UI effects (sound, animation) when disposing of a block.
+ * @param {!Blockly.BlockSvg} block The block being disposed of.
+ * @package
+ */
+Blockly.BlockAnimations.disposeUiEffect = function(block) {
+  var workspace = block.workspace;
+  var svgGroup = block.getSvgRoot();
+  workspace.getAudioManager().play('delete');
+
+  var xy = workspace.getSvgXY(svgGroup);
```
picklesrus: how come you can drop the type annotation? I guess I don't quite know why it was needed before.

rachel-fenichel: I was a cast before (from the perspective of closure-compiler).  `getSvgRoot` is already annotated to return the correct type.  

It might eventually need a cast inside it, but that's out of scope for right now.  It gives warnings, not errors.

PR #1819, opened 2018-04-23
Stop handling keypresses when the workspace is hidden.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/LLK/scratch-blocks/issues/1019
### Proposed Changes

Add an `isVisible` property on `WorkspaceSvg`.
Make `rendered` specifically refer to whether a workspace is a `Workspace` or `WorkspaceSvg`.
Set `isVisible` when updating the visibility of a rendered workspace.
Check for `isVisible` when handling kepresses.

### Reason for Changes

Blockly shouldn't eat keyboard events when hidden.
Particularly annoying when trying to undo changes in the paint editor in scratch blocks accidentally deletes blocks instead.

### Test Coverage

Tested on my local playground and the hosted playground to make sure the behaviour has actually changed.

Added some blocks to the workspace.
Clicked "Hide"
Hit Ctrl-Z a few times
Clicked "Show"
Confirmed that all of the blocks are still there.

On the hosted playground I confirmed that some of the blocks had disappeared while hidden.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -173,9 +173,12 @@ Blockly.svgResize = function(workspace) {
  * @private
  */
 Blockly.onKeyDown_ = function(e) {
-  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e)) {
+  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e)
+      || (Blockly.mainWorkspace.rendered && !Blockly.mainWorkspace.isVisible)) {
     // No key actions on readonly workspaces.
     // When focused on an HTML text input widget, don't trap any keys.
+    // Ignore keypresses on rendered workspaces that have been explicitly
+    // hidden.
     return;
   }
   var deleteBlock = false;
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -128,11 +128,19 @@ Blockly.WorkspaceSvg.prototype.resizeHandlerWrapper_ = null;
 
 /**
  * The render status of an SVG workspace.
- * Returns `true` for visible workspaces and `false` for non-visible,
- * or headless, workspaces.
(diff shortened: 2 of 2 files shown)


core/workspace_svg.js
```diff
@@ -128,12 +128,20 @@ Blockly.WorkspaceSvg.prototype.resizeHandlerWrapper_ = null;
...
  */
 Blockly.WorkspaceSvg.prototype.rendered = true;
 
+/**
+ * Whether the workspace is visible.  False if the workspace has been hidden
+ * by calling `setVisisble(false)`.
+ * @type {boolean}
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.isVisible = true;
```
AnmAtAnm: I see the parallel with `setVisible(..)`, but wish it was named different, since it doesn't account for parent container `display` or viewport boundaries. Since it is `@package`, I suppose this is fine for now and adaptable in the future.

core/blockly.js
```diff
@@ -173,9 +173,12 @@ Blockly.svgResize = function(workspace) {
  * @private
  */
 Blockly.onKeyDown_ = function(e) {
-  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e)) {
+  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e)
+      || (Blockly.mainWorkspace.rendered && !Blockly.mainWorkspace.isVisible)) {
```
AnmAtAnm: This looks like it could break in situations with multiple workspaces. What if `mainWorkspace` is the read-only / headless workspace, while other workspaces are able to accept input. Consider the multiple workspaces in the developer tools; the assignment of `mainWorkspace` is somewhat arbitrary.

At minimum, add a note in the function documentation or add a TODO to address this in the future.

PR #1823, opened 2018-04-25
Simplify RTL handling of '<' and '>'.

NeilFraser:
Tested in Playground and RTL demo.

### DIFF
--- a/blocks/logic.js
+++ b/blocks/logic.js
@@ -133,10 +133,10 @@ Blockly.defineBlocksWithJsonArray([  // BEGIN JSON EXTRACT
         "options": [
           ["=", "EQ"],
           ["\u2260", "NEQ"],
-          ["<", "LT"],
-          ["\u2264", "LTE"],
-          [">", "GT"],
-          ["\u2265", "GTE"]
+          ["\u200F<", "LT"],
+          ["\u200F\u2264", "LTE"],
+          ["\u200F>", "GT"],
+          ["\u200F\u2265", "GTE"]
         ]
       },
       {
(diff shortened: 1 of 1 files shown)


AnmAtAnm:
I'd rather see a `Blockly.Msg` entry for the label. I know this adds four string to the string table for all languages, but it seems like the more appropriate solution.

NeilFraser:
Tested on Firefox and Chrome for Mac.

Adding these strings to the string table would solve the problem from an engineering point of view.  Then it just becomes an issue of educating non-technical translators when to type [right-to-left marks](https://en.wikipedia.org/wiki/Right-to-left_mark) in the translation console.  Oh, and we also need to get the Iraqi translators to agree with the Iranian translators about how math works.

This is a rabbit hole I don't think we want to go down.

NeilFraser:
BTW, here's a fun read if one does want to peek down the rabbit hole:
  https://www.w3.org/TR/arabic-math/

AnmAtAnm:
Issues such as Arabic variants are exactly why I think this should be a localization issue. It appears we already support regional variants (e.g., [be](https://github.com/google/blockly/blob/master/msg/json/be.json) vs [be-tarask](https://github.com/google/blockly/blob/master/msg/json/be-tarask.json)).

While your solution is simpler than the previous code, it still encodes a preference into the code that we shouldn't dictating.

Thankfully, we're not attempting to translate mathematical equations in whole, the way MathML is.

AnmAtAnm:
Staging site for testing other browsers and devices: http://anmatanm.github.io/blockly/demos/code/index.html?lang=fa

Expected rendering (screenshot of previous version):
screen shot 2018-05-07 at 2 36 18 pm
screen shot 2018-05-07 at 2 37 33 pm

AnmAtAnm:
Tested in Android Oreo Chrome, iOS 9.3.5 Safari, Windows 10 Edge.

PR #1826, opened 2018-04-25
Using %{BKY_} notation for toolbox category colours.

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

Use `Msg` constants as the source of truth for toolbox categories.

### Proposed Changes

In toolbox categories, replacing hard coded numbers with `%{BKY_..}` reference notation, adding the `colour` attribute when missing.

I also added `<sep></sep>` when missing above the variable and procedure categories.

### Reason for Changes

Ensure toolbox category colors match the block colors.

### Test Coverage

I opened all changed demos, ensuring the category colors did not load as black.

Tested on:
 * Desktop Chrome

### DIFF
--- a/demos/interpreter/step-execution.html
+++ b/demos/interpreter/step-execution.html
@@ -48,14 +48,14 @@
   </div>
 
   <xml id="toolbox" style="display: none">
-    <category name="Logic">
+    <category name="Logic" colour="%{BKY_LOGIC_HUE}">
       <block type="controls_if"></block>
       <block type="logic_compare"></block>
       <block type="logic_operation"></block>
       <block type="logic_negate"></block>
       <block type="logic_boolean"></block>
     </category>
-    <category name="Loops">
+    <category name="Loops" colour="%{BKY_LOOPS_HUE}">
       <block type="controls_repeat_ext">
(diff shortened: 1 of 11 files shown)


demos/interpreter/step-execution.html
```diff
@@ -82,8 +82,11 @@ <h1><a href="https://developers.google.com/blockly/">Blockly</a> &gt;
         </value>
       </block>
     </category>
-    <category name="Variables" custom="VARIABLE"></category>
-    <category name="Functions" custom="PROCEDURE"></category>
+    <sep></sep>
+    <category name="Variables" custom="VARIABLE" colour="%{BKY_VARIABLES_HUE}">
```
rachel-fenichel: Is the color attribute parsed when doing a custom category?  Should it be?

AnmAtAnm: This works.  The `colour` attribute is part of the total XML, not the XML generated by the custom category function.

I haven't done any test where a custom category function might return a `colour` attribute on its top-level XML element.  I'm guessing such an attribute will be ignored and no toolbox category label color will update.

PR #1828, opened 2018-04-26
Decompose some functions for creating zoom controls SVG

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

None
### Proposed Changes

Make some helper functions to create the zoom controls.
### Reason for Changes

More work for easing merges with scratch blocks.  Will need an accompanying change on that side.

### Test Coverage
Tested on chrome in the playground.

### DIFF
--- a/core/zoom_controls.js
+++ b/core/zoom_controls.js
@@ -160,7 +160,7 @@ Blockly.ZoomControls.prototype.position = function() {
 
 /**
  * Create the zoom in icon and its event handler.
- * @param {number} rnd The random number to use to identify the clip paths.
+ * @param {string} rnd The random string to use to identify the clip paths.
  * @private
  */
 Blockly.ZoomControls.prototype.createZoomOutSvg_ = function(rnd) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@AnmAtAnm @picklesrus I'm tagging both of you but only need one review, and it's not urgent.

core/zoom_controls.js
```diff
@@ -251,3 +157,144 @@ Blockly.ZoomControls.prototype.position = function() {
   this.svgGroup_.setAttribute('transform',
       'translate(' + this.left_ + ',' + this.top_ + ')');
 };
+
+/**
+ * Create the zoom in icon and its event handler.
+ * @param {string} rnd The random string to use to identify the clip paths.
```
AnmAtAnm: Maybe pass in a complete id, instead of a id suffix? At least clarify the comment to note this is a suffix appended to the id.

It's not clear to me why the id needs to be randomized in the first place.

rachel-fenichel: https://github.com/google/blockly/blob/master/core/inject.js#L118

rachel-fenichel: Added comments.

PR #1839, opened 2018-05-03
Move workspace context menu options into contextmenu.js

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

None
### Proposed Changes

Decompose pieces of the showContextMenu function on the workspace.

### Reason for Changes

Mergeability with Scratch Blocks, readability.

### Test Coverage

I checked that the playground still opens and runs.

Tested on:

### DIFF
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -294,20 +294,14 @@ Blockly.ContextMenu.blockDuplicateOption = function(block) {
  */
 Blockly.ContextMenu.blockCommentOption = function(block) {
   // If there's already a comment, add an option to delete it.
-  if (block.comment) {
-    var text = Blockly.Msg['REMOVE_COMMENT'];
-    var callback = function() {
-      block.setCommentText(null);
-    };
-  } else {
-    // If there's no comment, add an option to create a comment.
-    var text = Blockly.Msg['ADD_COMMENT'];
-    var callback = function() {
-      block.setCommentText('');
-    };
-  }
+  // If there's no comment, add an option to create a comment.
+  var msgIdentifier = block.comment ? 'REMOVE_COMMENT' : 'ADD_COMMENT';
+  var callback = function() {
+    block.setCommentText(block.comment ? null : '');
+  };
 
-  return Blockly.ContextMenu.createOption(text, !goog.userAgent.IE, callback);
+  return Blockly.ContextMenu.createOption(
+      Blockly.Msg[msgIdentifier], !goog.userAgent.IE, callback);
 };
 
 /**
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -174,6 +174,25 @@ Blockly.ContextMenu.hide = function() {
 };
 
 /**
+ * Create and return a option object that can be added to a context menu's list
+ * of options.
+ * @param {string} text The text to display to the user.
+ * @param {boolean} enabled True if the option should be enabled.  If false, the
+ *     option will be visible but grayed out.
+ * @param {Function} callback The function to be called when the option is
+ *     clicked.
+ * @return {!Object} A menu option, containing text, enabled, and a callback.
+ * @package
+ */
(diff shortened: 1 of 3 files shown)


core/workspace_svg.js
```diff
@@ -1307,57 +1294,18 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
...
-        addDeletableBlocks(children[i]);
-      }
-    }
-  }
-  for (var i = 0; i < topBlocks.length; i++) {
-    addDeletableBlocks(topBlocks[i]);
-  }
+  var deleteList = Blockly.WorkspaceSvg.buildDeleteList_(topBlocks);
 
+  var DELAY = 10;
```
picklesrus: This part of the diff is a little hard to follow - is this used?

core/contextmenu.js
```diff
@@ -301,6 +301,116 @@ Blockly.ContextMenu.blockCommentOption = function(block) {
...
+ * workspace.
+ * @param {!Blockly.WorkspaceSvg} ws The workspace where the right-click
+ *     originated.
+ * @return {!Object} A menu option, containing text, enabled, and a callback.
+ * @package
+ */
+Blockly.ContextMenu.wsRedoOption = function(ws) {
+  return {
+    text: Blockly.Msg['REDO'],
+    enabled: ws.hasRedoStack(),
```
picklesrus: Looking at these 3 options, seems like "enabled" is the the common thread, but you're getting at that in different ways (i.e. sometimes by passing the info in, other times by adding methods to workspace).  I'm wondering if it would be better to just pass the boolean in. That's almost what you're doing below with numTopBlocks. 
It also means you could just have a more generic addOption method that takes the text, enabled, and callback.

After reading a bit more code, it might work for collapse and expand too?

rachel-fenichel: That feels like it's pushing the code back into workspace, which is what I was trying to get rid of.  That's sort of the structure that the code has over there, just with inline creation of the objects.

core/contextmenu.js
```diff
@@ -173,6 +173,25 @@ Blockly.ContextMenu.hide = function() {
...
+ *     clicked.
+ * @return {!Object} A menu option, containing text, enabled, and a callback.
+ * @package
+ */
+Blockly.ContextMenu.createOption = function(text, enabled, callback) {
+  return {
+    text: text,
+    enabled: enabled,
+    callback: callback
+  };
```
AnmAtAnm: Personally, I find the inline object easier to read, since the keys describe the values, unlike this factory function with unlabelled arguments.  Further, the object literal promotes use of the even-smaller property initializers in ES2015 (if we ever support transpiling): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015

I understand there is a tradeoff in terms of making sure all necessary values are present (TypeScript is nice in this respect).

core/workspace_svg.js
```diff
@@ -1372,19 +1322,20 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
     Blockly.Events.setGroup(false);
   }
 
+  var deleteCount = deleteList.length;
   var deleteOption = {
-    text: deleteList.length == 1 ? Blockly.Msg['DELETE_BLOCK'] :
-        Blockly.Msg['DELETE_X_BLOCKS'].replace('%1', String(deleteList.length)),
-    enabled: deleteList.length > 0,
+    text: deleteCount == 1 ? Blockly.Msg['DELETE_BLOCK'] :
```
AnmAtAnm: Switch this to use Blockly.ContextMenu.createOption, like all the others (since you've switched all the others)?

rachel-fenichel:
Picklesrus and I also had some hesitations about the structural changes.  If I can't find something that we're all comfortable with I'll leave it as is.

PR #1841, opened 2018-05-04
Make ui events work when block is null; test

### LINKED ISSUE #1810: Category selected UI event has null element
Looks like this recent change makes the assumption that an event with no block will use fromJson to fill in its attributes
https://github.com/google/blockly/commit/453fa014c701a92db7f8989ac40fb525fd5673fa

but that's not the case here: https://github.com/google/blockly/blame/master/core/toolbox.js#L583

### Problem statement

Category UI events have no content (category name)

### Expected Behavior

event.element should be set to the old category name (i think)

### Actual Behavior

event.element is empty

### Steps to Reproduce

1. Open playgrounds and enable event logging
2. Select categories in the toolbox. 
3. You'll see empty ui events with just the workspaceId set. 

All

### Additional Information

Affects master and develop

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

#1810, I think.

### Proposed Changes

Don't cancel out of UI event creation if the block is null.  The block is essentially optional for UI events.
Add tests

### Reason for Changes

Fix bug #1810 

### Test Coverage
Tested in playground: opening one toolbox category and then another gives these events:

image

Tested on:

### DIFF
--- a/core/ui_events.js
+++ b/core/ui_events.js
@@ -34,6 +34,10 @@ goog.require('goog.math.Coordinate');
 
 /**
  * Class for a UI event.
+ * UI events are events that don't need to be sent over the wire for multi-user
+ * editing to work (e.g. scrolling the workspace, zooming, opening toolbox
+ * categories).
+ * UI events do not undo or redo.
  * @param {Blockly.Block} block The affected block.
  * @param {string} element One of 'selected', 'comment', 'mutator', etc.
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@samelhusseini @NeilFraser Please take a look.

samelhusseini:
Looks good, thanks!

PR #1844, opened 2018-05-04
Delay getting inverse screen CTM until needed

gnarf:
## The basics

- [x] I branched from develop

### Proposed Changes

* Wait until we are asked for this inverse screen CTM instead of immediately requesting the CTM from the browser right after we've mutated it, causing a forced layout / recalc styles.

### Reason for Changes

* to quote Chrome Debugger - "Forced reflow is a likely performance bottleneck"
* This forced reflow happens 2-3x per workspace update
* Nothing seems to read this value until drag events/other events are processed

Tested on:
* Desktop Chrome

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -298,7 +298,14 @@ Blockly.WorkspaceSvg.prototype.targetWorkspace = null;
  * @type {SVGMatrix}
  * @private
  */
-Blockly.WorkspaceSvg.prototype.inverseScreenCTM_ = Blockly.DIRTY;
+Blockly.WorkspaceSvg.prototype.inverseScreenCTM_ = null;
+
+/**
+ * Inverted screen CTM is dirty, recalculate it.
+ * @type {Boolean}
+ * @private
+ */
+Blockly.WorkspaceSvg.prototype.inverseScreenCTMDirty_ = true;
 
 /**
  * Getter for the inverted screen CTM.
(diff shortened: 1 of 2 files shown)


gnarf:
I signed it!

rachel-fenichel:
@picklesrus I just approved this for scratch-blocks.  PTAL for Blockly.  This should not be adding any external APIs, and we're using getInverseScreenCTM everywhere we access the CTM already.

core/workspace_svg.js
```diff
@@ -288,24 +288,34 @@ Blockly.WorkspaceSvg.prototype.targetWorkspace = null;
...
+
+  // Defer getting the screen CTM until we actually need it, this should
+  // avoid forced reflows from any calls to updateInverseScreenCTM.
+  if (this.inverseScreenCTM_ == Blockly.DIRTY) {
+    var ctm = this.getParentSvg().getScreenCTM();
+    if (ctm) {
+      this.inverseScreenCTM_ = ctm.inverse();
+    } else {
+      // When dirty, and we can't get a CTM, set it to null.
+      this.inverseScreenCTM_ = null;
```
picklesrus: ooc, is this a behavior change? Seems like before we just kept the old one, though that was likely to be null in this case anyway?  I'm fine with it this way, just wondering if I should try and keep this in my memory for testing of next release :)

gnarf: Nice catch - it is a minor (unnoticed by me) change, if for some reason the `getParentSvg().getScreenCTM()` is no longer working, we would return `null` instead of the last value (which is probably `null`)

I could "fix" it to not have that behavior change pretty easily by adding a @private boolean `WorkspaceSvg.inverseScreenCTMDirty_`, if you think this edge case is an edge we should be wary of.

picklesrus: I vaguely remember times when we've seen it be null... but can't quite remember @rachel-fenichel, do you remember?  I'm thinking maybe when Blockly gets loaded in a hidden iframe.  Does that ring a bell?

gnarf: I also wonder if return the "last time we had a ctm" is a better result than `null` - "no CTM exists right now"

Though it is a change in behavior, which I'm still happy to solve ;)

gnarf: > That being said, our consumer for the screen CTM appears to go do extra work if it's null, so we should keep that in mind.

It's interesting that the method calls the same `getScreenCTM().inverse()` directly without checking for nulls, of course it's using an svg element and not the `WorkspaceSvg`, and it is only happening for mouse events on the svg, so can pretty safely assume the svg is on screen, and the CTM wont be null?

I would feel like in the case where we returned `null` from the workspace, this would also get `null` and throw an error?

picklesrus: If it isn't hard (which is sounds like you think it is isn't based on your original suggestion?), I think we'd be best to preserve current behavior :-/

Issue #1806, opened 2018-04-18
RTL mutators broken on Edge and IE11


rachel-fenichel:
### Problem statement
RTL mutators are broken in Edge and IE11, on both master and rc/april_2018.

The mutator's flyout is on the left side instead of the right.  The block in the mutator workspace is where it should be.  Blocks being dragged out of the flyout appear to the right of the cursor.

### Expected Behavior

RTL mutator flyout behaves the same in all browsers.

### Operating System and Browser

* Windows Internet Explorer 11
* Windows Edge

### Additional Information

Since this is already on master, I'm not going to let it block this release.  But it is important, and hopefully a simple fix.

rachel-fenichel:
@timdawborn @samelhusseini Have you been running into this?  If so, do you know how long it's been here?

samelhusseini:
Yep looks to have been broken for a while.  https://makecode.microbit.org/?lang=ar (is running a ~ 4 month old blockly) and it repro's there.

Issue is CSS transforms on SVG (inner) elements are not supported in IE and Edge.
I've scanned the codebase and https://github.com/google/blockly/blob/master/core/flyout_base.js#L375 (when the tagname is a <g>) is the only instance where Blockly.utils.setCssTransform is used on an SVG <g> element. All other instances apply this on the SVG <svg> element and that's okay AFAIK. 

The fix is to use the transform attribute on the element itself when the flyout is using the <g> tag.

PR #1846, opened 2018-05-11
Fix RTL mutators in Edge and IE

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/1806

### Proposed Changes

CSS transforms on SVG (inner) elements are not supported in IE and Edge.

I've scanned the codebase and https://github.com/google/blockly/blob/master/core/flyout_base.js#L375 (when the tagname is a `<g>`) is the only instance where Blockly.utils.setCssTransform is used on an SVG (inner) element. All other instances apply this on the SVG `<svg>` element and that's okay.

### Reason for Changes

RTL mutators were broken on IE and Edge. The transform wasn't getting applied and so both the flyout and the workspace were left aligned (on top of each other)

### Test Coverage

Tested on: 
- Desktop Mac Chrome
- IE 11
- Edge 16

### Additional Information

Not sure how to avoid something like this in the future @rachel-fenichel any thoughts?

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -371,8 +371,15 @@ Blockly.Flyout.prototype.updateDisplay_ = function() {
 Blockly.Flyout.prototype.positionAt_ = function(width, height, x, y) {
   this.svgGroup_.setAttribute("width", width);
   this.svgGroup_.setAttribute("height", height);
-  var transform = 'translate(' + x + 'px,' + y + 'px)';
-  Blockly.utils.setCssTransform(this.svgGroup_, transform);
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
It's a similar issue to https://github.com/google/blockly/issues/1425.  It mostly comes down to using a <g> for the flyout in the mutator but otherwise using a <svg>.  

That happened when we were restructuring the SVGs for performance improvements.  Mutator flyouts were an edge case so we just did something that worked well enough--but obviously it didn't actually work that well.  The long term solution will be to make the flyout's SVG consistent regardless of where it's living.

PR #1857, opened 2018-05-15
Remove ES6 leaking into codebase

NeilFraser:
Breaks closure compiler


AnmAtAnm:
Allowing ES6 in Node scripts such as these. Corrected compilation documentation to not accidentally include these files in the glob.

PR #1889, opened 2018-05-23
Fix undoing a drag to connect

### LINKED ISSUE #1780: Input connection move events do not undo fully
### Problem statement

If a move event results in a connection, undoing the connection will not move the block back to its original location.

#### Before
screen shot 2018-04-13 at 3 54 27 pm

#### After Move
screen shot 2018-04-13 at 3 54 37 pm

#### After Undo
screen shot 2018-04-13 at 3 54 49 pm

### Expected Behavior

The moved block should not only disconnect, but move back to its original location (below the text block, in the above example).

### Actual Behavior

The moved block disconnects, and bumps slightly (less than a normal full bump) from the previously connected block.

### Steps to Reproduce

1. Start with two blocks on the workspace, one with an input and another that is a valid output.
2. Drag the output block into the input.
3. Immediately undo the move (Ctrl+Z or Cmd+Z).

The behavior is similar when undoing a connection between next and previous connections.

The behavior does not depend on the presence of a shadow block in the input.

### Playground event log

screen shot 2018-04-13 at 4 08 10 pm

Note that the new coordinate of the move triggered by the undo is identical to the old coordinate of the original move, even though the behavior on the workspace does not reflect this.

 * Desktop Chrome

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1780 

### Proposed Changes

Fire the move event at the end of a drag before firing the connect event.  This means that the events combine correctly.

Get rid of a check for `Blockly.selected` in `block.setParent`.

### Reason for Changes

For the change in block_dragger.js, see the [explanation](https://github.com/google/blockly/issues/1780#issuecomment-385087792) by @jlaurens in #1780.

The change in block_svg.js gets rid of a check against `Blockly.selected` that was causing problems if the block was actually selected.  I added that check as part of https://github.com/google/blockly/commit/df7f534ad6ed287ecc8d13c4e159138b71ed1e88.  It came in from scratch-blocks; I had noted that it was an odd check there, but didn't investigate further yet.  It appears to cause real problems here.

### Test Coverage
I imported two blocks (a math_number block and an addition block with no shadows).  I placed them in known locations, dragged to connect, and then used ctrl-z to undo.  I tested with the block selected and deselected.

This was all in the playground.  This would be a good candidate for testing with an event replay framework.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -224,9 +224,9 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
     // These are expensive and don't need to be done if we're deleting.
     this.draggingBlock_.moveConnections_(delta.x, delta.y);
     this.draggingBlock_.setDragging(false);
+    this.fireMoveEvent_();
     this.draggedConnectionManager_.applyConnections();
     this.draggingBlock_.render();
-    this.fireMoveEvent_();
     this.draggingBlock_.scheduleSnapAndBump();
   }
   this.workspace_.setResizesEnabled(true);
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -279,12 +279,8 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
   // If we are losing a parent, we want to move our DOM element to the
   // root of the workspace.
   else if (oldParent) {
-    // Avoid moving a block up the DOM if it's currently selected/dragging,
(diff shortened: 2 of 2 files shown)


rachel-fenichel:
On second thought, this isn't quite right.  It works just fine when there aren't any shadows, but when there are shadow blocks there are warnings if I move the `fireMoveEvent` call above the `applyConnections` call.

PR #1890, opened 2018-05-24
Removed incorrect escape sequence

Ellipsis753:
Hello. This line seems to cause any percents in strings to be turned into a literal \%.
This is because a Python percent cannot be escaped using a backslash. 

In a normal Python string I think that a % is meaningless and must not be escaped.

When a percent is used in the context of formatting it should be escaped as %%.

## The basics

- [ ] I branched from develop

### Test Coverage

Tested on:


rachel-fenichel:
Also needs to be against develop instead of master.

RoboErikG:
Hi Ellipsis, just a reminder we'll be happy to take this PR once it's against the develop branch and you've signed the CLA. Thanks for contributing!

AnmAtAnm:
No response from contributor. See #1982 instead.

Issue #1891, opened 2018-05-24
Performance and cost of filtering events


rachel-fenichel:
### Problem statement

The filter function in `Blockly.Events` combines related events (e.g. two moves of the same block) and removes events that do not need to be sent over the wire for simultaneous editing.  Events that are filtered out are never fired.

The goal is to reduce the number of events that needs to be sent over the wire, especially in cases when the two users were disconnected for long enough to build up a large queue.

This may be unnecessarily expensive when most of the events are not duplicates or when doing forseeably large batch operations, such as swapping sprites in Scratch.  We have previously improved the behaviour of the filter function to make it much faster, but 

The filter used to combine all move events without considering intervening events.  This can be a problem if the intermediate events relied on the move (e.g. connecting over a shadow block and then disconnecting).  This shows as a breakdown in undo/redo, but not in triggering action.

### Possible solutions
- Disable filtering entirely
- Allow the developer to enable or disable filtering
  - A developer who wants to send things over the wire or have heavyweight listeners would want filtering on
  - A developer with lightweight listeners and lots of batch actions might want filtering off

### To explore
- For some basic workflows, what percent of events are actually filtered out?
  - Load workspace
  - Clear workspace
- Are the current rules correct about which events get merged?

### Test code
To disable filtering, replace the first few lines of `Blockly.Events.fireNow_`.  Delete
```
  var queue = Blockly.Events.filter(Blockly.Events.FIRE_QUEUE_, true);
  Blockly.Events.FIRE_QUEUE_.length = 0;
```

and replace with

```
  var queue = Blockly.Events.FIRE_QUEUE_;
  Blockly.Events.FIRE_QUEUE_ = [];
```

rachel-fenichel:
@kchadha @gnarf @ewpatton This may be relevant for you.  Please let me know your thoughts, or any benchmarking you can provide for event filtering in your projects.

cpcallen:
The title of this issue is "Performance and cost of filtering events", but the major performance issue had already been dealt with in #1205.  It appears that the matters raised in this issue that were actually live at the time are

> The filter used to combine all move events without considering intervening events. This can be a problem if the intermediate events relied on the move (e.g. connecting over a shadow block and then disconnecting). This shows as a breakdown in undo/redo, but not in triggering action.

and 

> * For some basic workflows, what percent of events are actually filtered out?
> * Are the current rules correct about which events get merged?

[...]

PR #1892, opened 2018-05-25
Make event filtering stricter for move events

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves
The rest of https://github.com/google/blockly/issues/1780

### Proposed Changes

Make the filter function only merge consecutive move events on the same blocks.

### Reason for Changes

The fix in #1889 exposed a new problem, which was that events were getting merged in the wrong way.  The example is:

image

Dragging the "123" block to the addition block leads to four events:
- move "123" to an x,y position near the addition block
- move the shadow block from the parent block to the main workspace
- delete the shadow block
- move "123" to connect to the addition block

The middle two events are a result of calling `unplug` on the shadow block.

The previous merge algorithm then merged the first and 4th event, giving:
- move "123" to connect to the addition block
- move the shadow block from the parent block to the main workspace
- delete the shadow block

This worked fine, but undoing it tries to undelete and move a shadow block that doesn't exist.  That's because the first event actually caused the second and third events.

### Test Coverage
I added a new jsunit test to check exactly this case.  The other tests still pass.

In the playground, I dragged to connect a bunch of combinations of blocks (connect over a shadow, connect an input to an output, connect to a statement input, etc).  For each one I did the drag, then undid it, then redid it.  Everything ended up in the right place.  Again, this is a good candidate for the event replay framework.

### Additional Information

This issue affects scratch blocks as well, so they'll need the same fix.

This also led to #1891, which is a bigger question about event filtering in general.

### DIFF
--- a/core/dragged_connection_manager.js
+++ b/core/dragged_connection_manager.js
@@ -126,6 +126,16 @@ Blockly.DraggedConnectionManager.prototype.wouldDeleteBlock = function() {
 };
 
 /**
+ * Return whether the block would be deleted if dropped immediately, based on
+ * information from the most recent move event.
+ * @return {boolean} true if the block would be deleted if dropped immediately.
+ * @package
+ */
+Blockly.DraggedConnectionManager.prototype.wouldConnectBlock = function() {
+  return !!this.closestConnection_;
+};
+
+/**
  * Connect to the closest connection and render the results.
  * This should be called at the end of a drag.
  * @package
--- a/core/dragged_connection_manager.js
+++ b/core/dragged_connection_manager.js
@@ -126,9 +126,9 @@ Blockly.DraggedConnectionManager.prototype.wouldDeleteBlock = function() {
 };
 
 /**
- * Return whether the block would be deleted if dropped immediately, based on
+ * Return whether the block would be connected if dropped immediately, based on
  * information from the most recent move event.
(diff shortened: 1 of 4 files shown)


core/dragged_connection_manager.js
```diff
@@ -125,6 +125,16 @@ Blockly.DraggedConnectionManager.prototype.wouldDeleteBlock = function() {
   return this.wouldDeleteBlock_;
 };
 
+/**
+ * Return whether the block would be deleted if dropped immediately, based on
```
AnmAtAnm: Please correct comment.

tests/jsunit/event_test.js
```diff
@@ -790,3 +790,30 @@ function test_events_newblock_newvar_xml() {
...
+function test_events_filter_nomerge_move() {
+  // Move events should only merge if they refer to the same block and are
+  // consecutive.
+  eventTest_setUpWithMockBlocks();
+  try {
+    var block1 = createSimpleTestBlock(workspace);
+    var block2 = createSimpleTestBlock(workspace);
+
+    var events = [];
+    helper_addMoveEvent(events, block1, 1, 1);
```
AnmAtAnm: Can you add a comment about the shadow block case this attempts to emulate, giving the test some context?

PR #1900, opened 2018-06-01
Renaming injectionDiv as blocklyInjectionDiv, for namespace consistency

### LINKED ISSUE #1258: CSS namespace injectionDiv
One of the CSS rules falls outside the Blockly namespace:
https://github.com/google/blockly/blob/master/core/css.js#L129

This should be called blocklyInjectionDiv, or somesuch.  Otherwise we may collide with random elements on the host page.

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

#1258

### Proposed Changes

Renames the class "injectionDiv" as "blocklyInjectionDiv".

**Note:** This is technically a breaking change of a class name that has existed for many years. However, it is hard to imagine a app developer addressing this div specifically, instead of the parent div (the one the app developer named and passed into `Blockly.inject(...)`, and the one this `<div>` is sized to match).

If we choose not to accept this change, we should label #1258 with "wontfix".

### Reason for Changes

 * Consistency in the use of the "blockly" prefix in class names.
 * Reduce the change of class name collision with other CSS rules.

### Test Coverage

 * Opened the playground to verify it lays out correctly. Verified blocks on the workspace and during drag continue to clip at the workspace edges (an injection div CSS rule). Verified resizing the window continues to resize the workspace.
 * Rebuilt the code and tested the same on the fixed and resizable injection demos.
 * Ran jsunit tests.

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -126,7 +126,7 @@ Blockly.Css.CONTENT = [
     'z-index: 99999;', /* big value for bootstrap3 compatibility */
   '}',
 
-  '.injectionDiv {',
+  '.blocklyInjectionDiv {',
     'height: 100%;',
     'position: relative;',
     'overflow: hidden;', /* So blocks in drag surface disappear at edges */
--- a/core/inject.js
+++ b/core/inject.js
@@ -56,7 +56,7 @@ Blockly.inject = function(container, opt_options) {
     throw 'Error: container is not in current document.';
   }
   var options = new Blockly.Options(opt_options || {});
-  var subContainer = goog.dom.createDom(goog.dom.TagName.DIV, 'injectionDiv');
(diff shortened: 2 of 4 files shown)


rachel-fenichel:
I vote for leaving this alone and wontfix-ing #1258, or else adding "blocklyInjectionDiv" as a class with the same styling and leaving "injectionDiv" there.

I'm not that worried about the name collision, and I don't want to break anyone who *has* used this for their rules.

AnmAtAnm:
Duplicate style rules is just asking for trouble in CSS. I'm not sure how it will interact with people who have only overridden the older of the two class names, the people we are claiming to help by not submitting this.

I'm fine with marking the bug `wontfix`.

PR #1911, opened 2018-06-08
Fixing xml.js: Always throw XML I/O errors; Support parsing in Node

AnmAtAnm:
* Support JSDOM as an alternate parser under Node.js
 * Throw standard errors instead of goog.asserts.
   Asserts are ignore when not in goog.DEBUG (e.g., compiled).

## The basics

- [x] I branched from develop

### Resolves

`goog.asserts` would not throw if `goog.DEBUG` was true.

Node.js does not provide DOMParser. Add support for JSDOM.

### Proposed Changes

 * Removing `goog.asserts`, and replacing with `throw Error(..)` calls.
 * If DOMParser is not present, `require('jsdom')` and override the method that does the parsing.
 * Use `document.documentElement` instead of `document.firstChild`. A valid XML document can have more than one child, such as a comment. Relax the constraint.
 * Updated `textToDom(..)` comments to be more accurate.

### Test Coverage

Tested on:
 * Desktop Chrome (after `build.py` and `compile.sh`)
   * Playground, with export/import of workspace
   * Generator tests
 * Node.js (Somewhat... It does not resolve all issues, but does get past `DOMParser` issue.)


core/xml.js
```diff
@@ -434,8 +458,7 @@ Blockly.Xml.domToWorkspace = function(xml, workspace) {
         }
         variablesFirst = false;
       } else if (name == 'shadow') {
-        goog.asserts.fail('Shadow block cannot be a top-level block.');
-        variablesFirst = false;
```
AnmAtAnm: Removed unreachable statement (now that is always throws).

AnmAtAnm:
CC: @NeilFraser, Supporting XML parsing in Node.

rachel-fenichel:
LGTM

Issue #1905, opened 2018-06-06
Replace goog.Timer.callOnce(..) with setTimeout


AnmAtAnm:
### Problem statement

We use both `goog.Timer.callOnce(...)` and `setTimeout(..)` in various places. We have not had problems with `setTimeout(..)`, but a recent issue (#1901) highlighted a problem with `Timer.callOnce(..)` dependency on a `goog.global`, which is set incorrectly in Node.js (https://github.com/google/closure-library/issues/903).

### Expected Behavior

Consistently use one API.

Vimal-Raghubir:
Hi, I am interested in working on this issue. So based on my understanding, all instances of `goog.Timer.callOnce()` has to be replaced with the `setTimeout()` function?

NeilFraser:
If you want to work on this, please submit changes to the "Closure-die-die-die" branch.  Eliminating all instances of "goog.Timer" from the codebase would be great.

AnmAtAnm:
Sorry @NeilFraser . This one is important enough, as the behavior causes problems in Browserify and WebPack.

@Vimal-Raghubir Your understanding is correct. Please make a Pull Request against the `develop` branch.

AnmAtAnm:
Resolved in #1917.

Issue #1901, opened 2018-06-04
Cannot Delete Blocks (Develop Branch)


TheBrokenRail:
### Problem statement

When trying to delete a block it throws an error.

### Expected Behavior

The block is deleted.

### Actual Behavior

It throws an error.

```
Uncaught TypeError: goog.Timer.defaultTimerObject.setTimeout is not a function
    at Function.goog.Timer.callOnce (bundle.js:formatted:5591)
    at Blockly.BlockDragger.maybeDeleteBlock_ (bundle.js:formatted:17759)
    at Blockly.BlockDragger.endBlockDrag (bundle.js:formatted:17737)
    at Blockly.TouchGesture.Blockly.Gesture.handleUp (bundle.js:formatted:18391)
    at Blockly.TouchGesture.handleUp (bundle.js:formatted:19323)
    at HTMLDocument.h (bundle.js:formatted:26732)
```

### Steps to Reproduce

1. ```git clone https://github.com/TheBrokenRail/MicroBlocks.git```
2. ```npm install```
3. ```npm start```
4. Go to ```localhost/editor``` in a web browser
5. Try to delete a block

### Operating System and Browser

Chrome on Ubuntu 18.04

### Additional Information

This is on the Blockly develop branch for its ability to generate a Node.js compatible file.

AnmAtAnm:
Note for my own clarification: The above linked repo is using Blockly as a direct git dependency via NPM, calls `npm install` in that dependent module (concatenating the `core`/`blocks`/`javascript` files), and loads it into a client-side page via browserify.

FYI: The files that `npm install` concatenates are not rebuilt on every develop branch commit. They may be weeks out of date with the actual source.

AnmAtAnm:
The delete key and the context menu delete actions both work fine. It is is dragging to delete (either over the toolbox or the trash can) that causes the stack trace.

I do not see this error in the demos, either in a direct Blockly check out of the develop directory, or the demos in the checkout created by npm (`node_modules/blockly/demos/code/index.html`).

AnmAtAnm:
This bug is a weird interaction between the Closure compiled code and browserify.

When I run the code through a static instance (`demos/code/index.html`), `goog.Timer.defaultTimerObject` is a reference to `window`:

![screen shot 2018-06-04 at 1 28 52 pm](https://user-images.githubusercontent.com/9916202/40940016-4dbece96-67fb-11e8-8c1b-81fffee1b556.png)

[...]

PR #1916, opened 2018-06-12
Replace call once with set timeout

Vimal-Raghubir:
## The basics

- [x] I branched from develop

### Resolves
Issue #1905 

### Proposed Changes
This change is to replace the `goog.Timer.callOnce` function with the `setTimeout` function since the `callOnce` function wasn't working as expected. 

### Reason for Changes

In the issue linked above and the related issues such as #1901 and https://github.com/google/closure-library/issues/903, there has been numerous issues with the `callOnce` function. Therefore the proposed change was to remove the usage of it and replace with the `setTimeout` function.

### Test Coverage
Windows 7, Google Chrome.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -267,7 +267,7 @@ Blockly.BlockDragger.prototype.maybeDeleteBlock_ = function() {
 
   if (this.wouldDeleteBlock_) {
     if (trashcan) {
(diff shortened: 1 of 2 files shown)


AnmAtAnm:
Sorry.  This is incomplete.  See #1917.

PR #1930, opened 2018-06-19
Fix case of redeclared var (i) in Blockly.Block.prototype.jsonInit

jollytoad:
## The basics

- [x] I branched from develop

### Resolves

I'm experimenting with generation of Typescript defs using clutz, which uses the closure-compiler, but it complains that a var is redeclared.

### Proposed Changes

Rename redeclared var.

### Reason for Changes

To allow clutz to process core blockly js.

### Test Coverage

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1150,8 +1150,8 @@ Blockly.Block.prototype.jsonInit = function(json) {
 
   if (Array.isArray(json['extensions'])) {
     var extensionNames = json['extensions'];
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
I'd be surprised if this is the only redeclared var it complains about.  Our policy is to allow variable redeclaration for loop counters to avoid accidentally creating globals if the code gets rearranged.

That said, this solution is easy and fine.

Issue #1967, opened 2018-07-05
Hidden statement inputs seem broken


Marclev78:
### Problem
If I attach a block to a hidden statement input (reasonably expecting that block to be hidden), the block is programmatically attached to the input, but shown free-floating on the workspace.

### Expected behavior
The "visible" flag of the input should cascade down to anything attached to it.

### Actual Behavior

Actual behavior is basically "Weird".

I can verify programmatically that the block is attached to the input, but visually it just appears near the block with the input on the work-space and I drag it around free-form.

This makes hiding a statement input somewhat futile.

### Steps to reproduce
I have code like this:

```
        const summaryInput = this.appendStatementInput("summaryCollection");
        summaryInput.setVisible(false);
        summaryInput.connection.connect(blockToAttach.previousConnection);
```

blockToAttach ends up showing detached on the workspace (even though it is in fact attached in the block model).

### Workaround

As I can't see a way of hiding a block (looking at the Block API), I'm not sure what the workaround is.

RoboErikG:
setVisible() is used when collapsing a block stack and we assumed devs wouldn't be modifying a user's blocks while they were hidden. As a workaround, you can toggle setVisible again or make a function to run the code that [typically hides a connected block](https://github.com/google/blockly/blob/master/core/input.js#L182) manually before attaching it.

As an aside, what's your use case?

Marclev78:
I have some "advanced" stuff that I don't want to clutter up the main workspace with and only want to appear when the user opens the mutation dialog, so that they can configure it there.
As it happens, setCollapsed() better suits my use case as that gives some visual indication that there is more to the block than it seems.

Mabe the documentation for `setVisible` should be updated to reflect that the dev may have to call `hideAll` / `unhideAll` on the connection as well for it to work?

RoboErikG:
Ah, setCollapsed() is probably the least work code wise, but your use case would work better with a custom collapse/uncollapse icon that mutates the block to show the hidden state. Anything can trigger a block mutation (it doesn't have to use the mutator dialog), so you could add a field_image with your collapse/uncollapse icon and mutate the block whenever it's clicked. We'll add a note to the js docs for that API.

Note: You do have to set the image field to editable to get click events on it (#1887 )

RoboErikG:
This turns out to be more complicated than expected. There are quite a few cases that go through different paths and have differing state depending on the timing. #2003 is a start at a fix but needs a lot of tests to be sure all cases are covered correctly.

Notes from working on this:

[...]

PR #1968, opened 2018-07-09
Add visibility note to Input.setVisible docs

RoboErikG:
Adds a note to the setVisible docs that any blocks that are connected/disconnected while the input is not visible need to be manually updated.

Issue #1967

## The basics

- [x] I branched from develop

### Resolves

#1967 

### Proposed Changes

Adds a note to the JS Doc.

### Reason for Changes

Developers weren't aware of implications.

### Test Coverage

Tested on:

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/1968)

### DIFF
--- a/core/input.js
+++ b/core/input.js
@@ -164,7 +164,9 @@ Blockly.Input.prototype.isVisible = function() {
 
 /**
  * Sets whether this input is visible or not.
- * Used to collapse/uncollapse a block.
+ * Used to collapse/uncollapse a block. If connections are modified while
+ * an input is not visible, any connected/disconnected blocks must have their
+ * visibility updated manually.
  * @param {boolean} visible True if visible.
  * @return {!Array.<!Blockly.Block>} List of blocks to render.
  */


AnmAtAnm:
Is this really supposed to be a public method?  What is the use case for an app developer setting visibility of an Input, without using block.setCollapsed()? Maybe there is a use case in scratch-blocks I'm not thinking of, but I think this method should be tagged `@package`. If there are cases, like scratch-blocks, where this method is intended to be external, then I'd like to see documentation about what those other expected uses are.

More importantly, requiring manually updating visibility on well defined events sounds like a bug. It sounds like the connection/disconnection code isn't fully testing/setting appropriate state. We should be able to inspect the block tree and identify connecting/disconnection from a collapsed subtree, so we should be able to correct this.

That said, given that there is no user visible means of making such connections on a collapsed block, making me believe that the bug is low priority.  How did you run into this issue?

RoboErikG:
Whoops, looks like I still need to write tests on #2003 . Going to close this since there's a fix written that will make the comment obsolete once it's submitted.

PR #1975, opened 2018-07-16
Removing goog.* uses from blocks/*.js files

### LINKED ISSUE #1971: Procedures uses goog.array.equals()
### Problem statement

`blocks/procedures.js` uses `goog.array.equals()`, but `goog.array` has been removed from usage in `core/`.

### Expected Behavior

`blocks/*.js` should never use `goog` calls.

### Additional Information

This may mean older `blocks/procedures.js` and `blocks_compress.js` files are incompatible with the latest `develop` branch and future releases.  We may need to hack an update marker and log message into core to notify developers.

AnmAtAnm:
## The basics

- [x] I branched from develop

### Resolves

#1971

### Proposed Changes

Replace all uses of `goog.*` library calls, other than `goog.provide(..)` and `goog.require(..)`. Specifically, `goog.dom.createDom(..)` and one use of `goog.array.equals(..)`.

### Reason for Changes

`blocks/` files may not be compiled with the Closure compiler and should never use `goog.*` library calls.

This happened to work in the past because these functions happened to be called by `core/` files. However, as we are distancing ourselves from Closure library, the calls in `core/` were removed and now causes the crash identified in #1971

### Test Coverage

Performed actions in the playground that triggered each of the modified methods:
 * Performed numerous various procedure mutator interactions/modiciations.
 * Used variable block shortcuts in context menus.

Tested on:
 * Desktop Chrome

### DIFF
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -268,8 +268,9 @@ Blockly.Constants.Loops.CUSTOM_CONTEXT_MENU_CREATE_VARIABLES_GET_MIXIN = {
       option.text =
           Blockly.Msg['VARIABLES_SET_CREATE_GET'].replace('%1', varName);
       var xmlField = Blockly.Variables.generateVariableFieldDom(variable);
-      var xmlBlock = goog.dom.createDom('block', null, xmlField);
+      var xmlBlock = document.createElement('block');
       xmlBlock.setAttribute('type', 'variables_get');
+      xmlBlock.appendChild(xmlField);
       option.callback = Blockly.ContextMenu.callbackFactory(this, xmlBlock);
       options.push(option);
     }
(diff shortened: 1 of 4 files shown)


AnmAtAnm:
Re: Namespaces....
The use of `createDom()` fits our current practice of ignoring namespaces. That said, namespaces are probably the right thing with respect to all of our various uses of XML, especially those that get mixed into the HTML (toolboxes, etc.). That seems like a big change without immediate benefit with more implications in `core/` than this block code. (I can't think of anyone running HTML validators on blockly app code. If they are, they haven't complained yet.) Low priority feature request? We'd have to make sure it is sufficiently backward compatible, too.

NeilFraser:
Agreed.  I think the only issue with using no namespace is that every element is effectively HTML.  Thus 'block' is fine, since we just get a blank element of that name.  But if we were to use an 'img' element, it would come with a bunch of properties we weren't expecting.

PR #1522, opened 2017-12-21
Fix generator tests

rachel-fenichel:
- Add the concept of a "developer variable", which is in the generated code but never shown to the user.  We were implicitly doing this in generators before, but now it's explicit.  
- Update Blockly.Names to look up variables by ID instead of name.
- Update generators to set up their Blockly.Names objects correctly.
- Update block definitions used in the generator tests to define `getDeveloperVariables` instead of `getVars`.
- Update block generators used in the generator tests to refer to developer variables with a different constant than user variables.
- Add a utility function in Blockly.Variables to get all developer variables.

All generator tests now pass with variables referenced by ID, except for PHP lists.  That was broken before as well.

Calling getName with the result of block.getFieldValue returns a language-safe version of the variable name.  Without this change it would return a language-safe version of the variable ID instead.

PR #1979, opened 2018-07-17
Match code with docs

gokula-krishna-dev:
## The basics

- [x] I branched from develop

### Resolves
The documentation for allDeveloperVariables in https://developers.google.com/blockly/reference/js/Blockly.Variables does not match the variable in code.

### Test Coverage

Tested on:

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -106,8 +106,8 @@ Blockly.Variables.allDeveloperVariables = function(workspace) {
   var hash = {};
   for (var i = 0; i < blocks.length; i++) {
     var block = blocks[i];
-    if (block.getDeveloperVars) {
-      var devVars = block.getDeveloperVars();
+    if (block.getDeveloperVariables) {
+      var devVars = block.getDeveloperVariables();
       for (var j = 0; j < devVars.length; j++) {
         hash[devVars[j]] = devVars[j];
       }
--- a/core/variables.js
+++ b/core/variables.js
@@ -106,7 +106,7 @@ Blockly.Variables.allDeveloperVariables = function(workspace) {
   var hash = {};
   for (var i = 0; i < blocks.length; i++) {
     var block = blocks[i];
-    var getDeveloperVariables = block.getDeveloperVars || block.getDeveloperVariables;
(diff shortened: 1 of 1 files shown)


gokula-krishna-dev:
I signed it!

AnmAtAnm:
@gokula-krishna-dev  Since this is already released code, I think we need to update the docs to match the code, not vice versa.

Alternatively, we could support both, maintaining `getDeveloperVars()` for backward compatibility. It's unfortunate we were not consistent with `Variables.allDeveloperVariables()` and `block.getDeveloperVars()`. If so, you'll need to add a comment about the backward compatibility (with a date) and update `tests/generators/unittest.js`.

core/variables.js
```diff
@@ -106,8 +106,8 @@ Blockly.Variables.allDeveloperVariables = function(workspace) {
   var hash = {};
   for (var i = 0; i < blocks.length; i++) {
     var block = blocks[i];
-    if (block.getDeveloperVars) {
-      var devVars = block.getDeveloperVars();
+    if (block.getDeveloperVariables) {
```
AnmAtAnm: Changing this without some backward compatible support may break existing blocks, and will fail the generator tests.

rachel-fenichel:
How many releases has this been in for?  It might be worth changing it to be consistent (and updating blocks) if it's still very new.

AnmAtAnm:
Seven months (#1522). So that includes Feb, April, and June releases.

rachel-fenichel:
I think we should change it and have backward compatibility for a bit (marked @deprecated with the date), and then eventually remove the backward compatibility.  It's churn, and three releases is a bit much, but it's such an obvious mismatch in something that people do seem to want to use.

gokula-krishna-dev:
@AnmAtAnm @rachel-fenichel I have added the backward compatibility support. It seems that new tests should be introduced for this scenario.

rachel-fenichel:
Okay, can you add tests for that?

gokula-krishna-dev:
Sure.

AnmAtAnm:
Closing this for lack of updates in the past month. See #2021.

PR #1986, opened 2018-07-24
Change XML attribute for procedure from varId to var-id

gokula-krishna-dev:
## The basics

- [x] I branched from develop

### Resolves
When function blocks are exported as XML the attribute set for 'varId' in argument variables mutation list are exported as 'varid' with small 'i'. So when the block is loaded back as XML while parsing 'varId' is parsed as null, because getAttribute is case sensitive when parsed as XML.

For example:
Consider an XML exported from blockly where 'varId' is exported as 'varid' and imported with the inbuilt function where 'varid' is parsed as 'varId' which results in null.

`new DOMParser().parseFromString('<xml xmlns="http://www.w3.org/1999/xhtml"><a **varid**="seek"></a></xml>', "text/xml").getElementsByTagName('a')[0].getAttribute('**varId**')`

The above code results `null`

while changing it to 
`new DOMParser().parseFromString('<xml xmlns="http://www.w3.org/1999/xhtml"><a **varid**="seek"></a></xml>', "text/xml").getElementsByTagName('a')[0].getAttribute('**varid**')`

The above code results `seek`

Note: getAttribute is not case sensitive when parsed as HTML, but they are when parsed as XML.

Since it is unable to parse the 'varId' properly while loading when function block is imported separately they create a new variable. This new variable ID will be conflicted with variable ID which is present in the function block XML which causes the error.

`Uncaught (in promise) Error: Variable "x" is already in use and its id is "csN.asr2XahzV(d?*EYW" which conflicts with the passed in id, "=JE@nKuHj|$U]7@|b|+A".`

For the XML:
```
<xml xmlns="http://www.w3.org/1999/xhtml" collection="true">
  <block type="procedures_defnoreturn" id="~WI0tzsDy,J#ssf,3LfS">
    <mutation>
      <arg name="x" varid="=JE@nKuHj|$U]7@|b|+A"></arg>
    </mutation>
    <field name="NAME">sample</field>
    <comment pinned="false" h="80" w="160">Describe this function...</comment>
    <statement name="STACK">
      <block type="variables_set" id="^*]l5G!HH89h3aOvM,0U">
        <field name="VAR" id="=JE@nKuHj|$U]7@|b|+A" variabletype="">x</field>
        <value name="VALUE">
          <block type="variables_get" id="Zt272XF.p%@,?m_|2Rw9">
            <field name="VAR" id="=JE@nKuHj|$U]7@|b|+A" variabletype="">x</field>
          </block>
        </value>
      </block>
    </statement>
  </block>
</xml>
```

### Test Coverage

Tested on:

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -131,7 +131,7 @@ Blockly.Blocks['procedures_defnoreturn'] = {
       var parameter = document.createElement('arg');
       var argModel = this.argumentVarModels_[i];
       parameter.setAttribute('name', argModel.name);
-      parameter.setAttribute('var-id', argModel.getId());
+      parameter.setAttribute('varid', argModel.getId());
       if (opt_paramIds && this.paramIds_) {
         parameter.setAttribute('paramId', this.paramIds_[i]);
       }
@@ -155,7 +155,7 @@ Blockly.Blocks['procedures_defnoreturn'] = {
     for (var i = 0, childNode; childNode = xmlElement.childNodes[i]; i++) {
       if (childNode.nodeName.toLowerCase() == 'arg') {
         var varName = childNode.getAttribute('name');
-        var varId = childNode.getAttribute('var-id') || childNode.getAttribute('varId');
+        var varId = childNode.getAttribute('varid') || childNode.getAttribute('varId');
         this.arguments_.push(varName);
         var variable = Blockly.Variables.getOrCreateVariablePackage(
             this.workspace, varId, varName, '');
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Why switch to `var-id` instead of `varid`?

I remember something similar in scratch-blocks.  I'll check how we resolved it there.

More broadly, this is part of the mixup between XML and HTML elements in serialization and parsing.

gokula-krishna-dev:
@rachel-fenichel I named `var-id` because I came across the key `web-class` in core/flyout_button.js and thought of using the similar naming convention. I will change it to 'varid'.

rachel-fenichel:
Looks good to me.  @AnmAtAnm @RoboErikG do you see any backwards compatibility issues here?

RoboErikG:
Lgtm. That should also catch the cases where it was converted to lowercase during serialization.

PR #1987, opened 2018-07-27
Insertion markers in Blockly!

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Add the insertion marker manager class
Mark the dragged connection manager class deprecated, and remove the only use of it
(I propose leaving it in until we have a clear upgrade strategy for developers)
Add utility functions that are called by the insertion marker manager

### Reason for Changes

Blockly should have insertion markers!  They look snazzy on Scratch Blocks, and the plan was always to backport them to Blockly. 

### Test Coverage

Unit tests pass.
Tested on:
* Desktop Chrome

### Additional Information

You can play with this on [my demo site](https://rachel-fenichel.github.io/blockly/tests/playground.html) for now.  I reserve the right to update that site without warning.

### DIFF
--- a/blockly_uncompressed.js
+++ b/blockly_uncompressed.js
@@ -37,7 +37,7 @@ this.BLOCKLY_BOOT = function(root) {
 goog.addDependency("../../../" + dir + "/core/block.js", ['Blockly.Block'], ['Blockly.Blocks', 'Blockly.Comment', 'Blockly.Connection', 'Blockly.Events.BlockChange', 'Blockly.Events.BlockCreate', 'Blockly.Events.BlockDelete', 'Blockly.Events.BlockMove', 'Blockly.Extensions', 'Blockly.Input', 'Blockly.Mutator', 'Blockly.utils', 'Blockly.Warning', 'Blockly.Workspace', 'goog.math.Coordinate']);
 goog.addDependency("../../../" + dir + "/core/block_animations.js", ['Blockly.BlockAnimations'], ['Blockly.utils']);
 goog.addDependency("../../../" + dir + "/core/block_drag_surface.js", ['Blockly.BlockDragSurfaceSvg'], ['Blockly.utils', 'goog.math.Coordinate']);
-goog.addDependency("../../../" + dir + "/core/block_dragger.js", ['Blockly.BlockDragger'], ['Blockly.BlockAnimations', 'Blockly.DraggedConnectionManager', 'Blockly.Events.BlockMove', 'goog.math.Coordinate']);
+goog.addDependency("../../../" + dir + "/core/block_dragger.js", ['Blockly.BlockDragger'], ['Blockly.BlockAnimations', 'Blockly.InsertionMarkerManager', 'Blockly.Events.BlockMove', 'goog.math.Coordinate']);
 goog.addDependency("../../../" + dir + "/core/block_events.js", ['Blockly.Events.BlockBase', 'Blockly.Events.BlockChange', 'Blockly.Events.BlockCreate', 'Blockly.Events.BlockDelete', 'Blockly.Events.BlockMove', 'Blockly.Events.Change', 'Blockly.Events.Create', 'Blockly.Events.Delete', 'Blockly.Events.Move'], ['Blockly.Events', 'Blockly.Events.Abstract', 'Blockly.Xml.utils', 'goog.math.Coordinate']);
 goog.addDependency("../../../" + dir + "/core/block_render_svg.js", ['Blockly.BlockSvg.render'], ['Blockly.BlockSvg']);
 goog.addDependency("../../../" + dir + "/core/block_svg.js", ['Blockly.BlockSvg'], ['Blockly.Block', 'Blockly.BlockAnimations', 'Blockly.ContextMenu', 'Blockly.Events.Ui', 'Blockly.Events.BlockMove', 'Blockly.Grid', 'Blockly.RenderedConnection', 'Blockly.Tooltip', 'Blockly.Touch', 'Blockly.utils', 'goog.color', 'goog.math.Coordinate']);
@@ -77,6 +77,7 @@ goog.addDependency("../../../" + dir + "/core/grid.js", ['Blockly.Grid'], ['Bloc
 goog.addDependency("../../../" + dir + "/core/icon.js", ['Blockly.Icon'], ['Blockly.utils', 'goog.math.Coordinate']);
 goog.addDependency("../../../" + dir + "/core/inject.js", ['Blockly.inject'], ['Blockly.BlockDragSurfaceSvg', 'Blockly.Css', 'Blockly.Grid', 'Blockly.Options', 'Blockly.utils', 'Blockly.WorkspaceSvg', 'Blockly.WorkspaceDragSurfaceSvg', 'goog.ui.Component', 'goog.userAgent']);
 goog.addDependency("../../../" + dir + "/core/input.js", ['Blockly.Input'], ['Blockly.Connection', 'Blockly.FieldLabel']);
+goog.addDependency("../../../" + dir + "/core/insertion_marker_manager.js", ['Blockly.InsertionMarkerManager'], ['Blockly.BlockAnimations', 'Blockly.Events.BlockMove', 'Blockly.RenderedConnection', 'goog.math.Coordinate']);
 goog.addDependency("../../../" + dir + "/core/msg.js", ['Blockly.Msg'], []);
 goog.addDependency("../../../" + dir + "/core/mutator.js", ['Blockly.Mutator'], ['Blockly.Bubble', 'Blockly.Events.BlockChange', 'Blockly.Events.Ui', 'Blockly.Icon', 'Blockly.utils', 'Blockly.WorkspaceSvg', 'Blockly.Xml', 'Blockly.Xml.utils']);
 goog.addDependency("../../../" + dir + "/core/names.js", ['Blockly.Names'], []);
@@ -1786,6 +1787,7 @@ goog.require('Blockly.Grid');
 goog.require('Blockly.HorizontalFlyout');
 goog.require('Blockly.Icon');
 goog.require('Blockly.Input');
+goog.require('Blockly.InsertionMarkerManager');
 goog.require('Blockly.Msg');
 goog.require('Blockly.Mutator');
 goog.require('Blockly.Names');
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -162,6 +162,9 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
     this.draggingBlock_.bringToFront();
   }
 
+  // // During a drag there may be a lot of rerenders, but not field changes.
+  // // Turn the cache on so we don't do spurious remeasures during the drag.
+  // Blockly.Field.startCache();
   this.workspace_.setResizesEnabled(false);
   Blockly.BlockAnimations.disconnectUiStop();
 
@@ -221,6 +224,8 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   // Make sure internal state is fresh.
   this.dragBlock(e, currentDragDeltaXY);
   this.dragIconData_ = [];
+
+  Blockly.Field.stopCache();
 
   Blockly.BlockAnimations.disconnectUiStop();
 
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -27,7 +27,7 @@
 goog.provide('Blockly.BlockDragger');
 
 goog.require('Blockly.BlockAnimations');
-goog.require('Blockly.DraggedConnectionManager');
+goog.require('Blockly.InsertionMarkerManager');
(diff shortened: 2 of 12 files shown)


core/block.js
```diff
@@ -591,6 +616,31 @@ Blockly.Block.prototype.setShadow = function(shadow) {
...
+ * Set whether this block is an insertion marker block or not.
+ * @param {boolean} insertionMarker True if an insertion marker.
+ */
+Blockly.Block.prototype.setInsertionMarker = function(insertionMarker) {
+  if (this.isInsertionMarker_ == insertionMarker) {
+    return;  // No change.
+  }
+  this.isInsertionMarker_ = insertionMarker;
+  if (this.isInsertionMarker_) {
+    // TODO: Why is this in Block.js instead of block_svg.js?
```
AnmAtAnm: Move this `TODO`/question to the the setColour method in `block.js`?

core/block_render_svg.js
```diff
@@ -1144,3 +1151,45 @@ Blockly.BlockSvg.prototype.renderStatementInput_ = function(pathObject, row,
...
+    var dx = existingConnection.x_ - newConnection.x_;
+    var dy = existingConnection.y_ - newConnection.y_;
+
+    newBlock.moveBy(dx, dy);
+  }
+};
+
+/**
+ * Visual effect to show that if the dragging block is dropped, this block will
+ * be replaced.  If a shadow block it will disappear.  Otherwise it will bump.
```
AnmAtAnm: Comma after
> If a shadow block

blockly_uncompressed.js
```diff
@@ -34,82 +34,83 @@ this.BLOCKLY_BOOT = function(root) {
     dir = this.BLOCKLY_DIR.match(/[^\/]+$/)[0];
   }
   // Execute after Closure has loaded.
-goog.addDependency("../../../" + dir + "/core/block.js", ['Blockly.Block'], ['Blockly.Blocks', 'Blockly.Comment', 'Blockly.Connection', 'Blockly.Events.BlockChange', 'Blockly.Events.BlockCreate', 'Blockly.Events.BlockDelete', 'Blockly.Events.BlockMove', 'Blockly.Extensions', 'Blockly.Input', 'Blockly.Mutator', 'Blockly.Warning', 'Blockly.Workspace', 'Blockly.Xml', 'goog.array', 'goog.asserts', 'goog.math.Coordinate', 'goog.string']);
```
AnmAtAnm: I know this is the only compressed file that should be impacted by your changes, but it doesn't seem right to only update one of the files.  I would rather see all three updated in a single followup push.

ewpatton:
Two thoughts from my point of view:

1. Allowing for the old behavior would be nice. We will have to update videos and tutorial GIFs, which will take time, but I think App Inventor users would appreciate the new functionality. We did something similar with the grid and made it user toggleable.

2. The insertion marker is only the currently selected block. This doesn't seem quite right when dragging/inserting a stack of blocks since the to-be-displaced blocks move only for the size of the single block, not the whole stack. It would be nice if the blocks were displaced by the height of the whole stack.

blockly_uncompressed.js
rachel-fenichel: If I don't update this file it's not testable.

core/block.js
rachel-fenichel: As I said, this isn't for review yet.  I'm not done with cleanup.

RoboErikG:
@ewpatton 
Regarding 2. That was a trade off done mostly to reduce the cost of creating the insertion marker and re-rendering the stack. The insertion marker is actually a block, so it has to go through the whole rendering pass to figure out the size and shape. If there was a really big stack that you were putting inside another block it took a while to make the shadow of the entire stack.

It was also a little jarring when you dragged a stack down another stack and everything kept jumping off the bottom of the page. We found just using a single block was less jumpy. It may be confusing for some users, though, since at first they might think only one block will be connected.

rachel-fenichel:
This is a preliminary pull request to show what changes are required.  I'm not planning to land anything until I'm happy with behaviour and appearance in all cases *and* we have a good developer story.

Some information to consider:
- InsertionMarkerManager and DraggedConnectionManager are fully interchangeable--that was a deliberate design choice.
- All of the changes are additive except for the changes to block_dragger.js.  
- Put another way, you can get back to the old behaviour by reverting the [three lines that changed](https://github.com/google/blockly/pull/1987/files#diff-51c9823352b7d54bc4c1da592b0d136f) in block_dragger.js.
- The eventual goal is to fully switch to insertion markers, rather than maintaining both styles.
- The maintenance cost of two paths like this is extremely high, mostly in manual testing on every release.
- As a result, I do not want to add an option to the injection struct to choose between insertion markers and small highlights.
- We've been working with insertion markers on scratch blocks for ~2 years and have been very happy with how intuitive they look and feel.
- Insertion markers *can* cause performance problems for extremely deep block stacks or slow browsers.
- We can continue to fix performance problems as we find them, which usually boosts performance across the board.
- Performance fixes are worthwhile in their own right, and are still less work than ongoing maintenance of the `dragged_connection_manager`.

**My current plan** is to leave `dragged_connection_manager.js` in the repository, marked deprecated, for a few releases after we enable insertion markers.  Developers who want to switch over more slowly, or keep the old style, can revert the three lines in `block_dragger.js`.  

Developers who want it to be user-configurable would be responsible for including both styles (easy to do) and piping the user's decision to the right place in `block_dragger.js`.  Those developers would also be responsible for keeping the old `dragged_connection_manager.js` in working form for their project.  That's generally a smaller maintenance job than trying to keep it working for all possibly types of blockly projects.

Thoughts?

rachel-fenichel:
@ewpatton for a working example of why the single-block behaviour is the right way to do it, hit the "Spaghetti" button and drag one of the if statements around inside the stack.  It'll be slow, but you'll notice that you *can* move it down one block at a time, allowing you to insert it anywhere. 

If the insertion marker contains the entire stack, the next block would jump out of view.  The jumpiness isn't just annoying: it makes it impossible to move the insertion marker down by a single block.  Instead you have to drag it to the side until the marker disappears, then drag back to the stack at (hopefully) the correct place.

ewpatton:
@RoboErikG @rachel-fenichel Okay, that's sounds like a reasonable argument for the one blockly only approach. Thanks for the info.

core/block_dragger.js
```diff
@@ -162,6 +162,9 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
     this.draggingBlock_.bringToFront();
   }
 
+  // // During a drag there may be a lot of rerenders, but not field changes.
+  // // Turn the cache on so we don't do spurious remeasures during the drag.
+  // Blockly.Field.startCache();
```
RoboErikG: Did you forget to uncomment this line?

core/connection.js
```diff
@@ -245,6 +245,14 @@ Blockly.Connection.prototype.dispose = function() {
   this.dbOpposite_ = null;
 };
 
+/**
+ * @return {boolean} true if the connection is not connected or is connected to
+ *    an insertion marker, false otherwise.
+ */
+Blockly.Connection.prototype.isConnectedToNonInsertionMarker = function() {
```
RoboErikG: This doesn't seem to be used anywhere. Also, what about connections that would be connected except there's an insertion marker between blocks?

core/connection.js
```diff
@@ -334,12 +342,66 @@ Blockly.Connection.prototype.checkConnection_ = function(target) {
...
+
+  if (isNextConnection ||
+      (isFirstStatementConnection && !sourceHasPreviousConn)) {
+    // If the candidate is the first connection in a stack, we can connect.
+    if (!candidate.targetConnection) {
+      return true;
+    }
+
+    var targetBlock = candidate.targetBlock();
+    // If it is connected a real block, game over.
```
RoboErikG: missing 'to' "If it is connected __ a real block"

core/connection.js
```diff
@@ -334,12 +342,66 @@ Blockly.Connection.prototype.checkConnection_ = function(target) {
...
+    if (!targetBlock.isInsertionMarker()) {
```
RoboErikG: what about shadow blocks?

core/connection.js
```diff
@@ -334,12 +342,66 @@ Blockly.Connection.prototype.checkConnection_ = function(target) {
...
+      return false;
+    }
+    // If it's connected to an insertion marker but that insertion marker
+    // is the first block in a stack, it's still fine.  If that insertion
+    // marker is in the middle of a stack, it won't work.
+    return !targetBlock.getPreviousBlock();
+  }
+  // ???
```
RoboErikG: ????

core/connection.js
```diff
@@ -348,41 +410,29 @@ Blockly.Connection.prototype.isConnectionAllowed = function(candidate) {
...
-      // fine.  Similarly, replacing a terminal statement with another terminal
-      // statement is allowed.
-      if (candidate.isConnected() &&
-          !this.sourceBlock_.nextConnection &&
-          !candidate.targetBlock().isShadow() &&
-          candidate.targetBlock().nextConnection) {
-        return false;
-      }
+    }
+    case Blockly.NEXT_STATEMENT: {
```
RoboErikG: Did you remove all the checks for next_statement?

core/constants.js
```diff
@@ -42,7 +42,26 @@ Blockly.FLYOUT_DRAG_RADIUS = 10;
...
+ * when a connection is already highlighted.
+ */
+Blockly.CONNECTING_SNAP_RADIUS = 48;
+
+/**
+ * How much to prefer staying connected to the current connection over moving to
+ * a new connection.  The current previewed connection is considered to be this
+ * much closer to the matching connection on the block than it actually is.
+ */
+Blockly.CURRENT_CONNECTION_PREFERENCE = 0;
```
RoboErikG: Leaving this at 0 caused jumpiness before. Did you fix that some other way or do you still need to play around with these values?

core/block_dragger.js
rachel-fenichel: Whoops, fixed in a commit that I forgot to push.

core/connection.js
rachel-fenichel: I think this logic is correct.  Shadow blocks are real blocks.  We don't want to drag a real block to connect below a shadow block.

core/connection.js
rachel-fenichel: Added back in a commit that I had forgotten to push.  It should be there now.

core/constants.js
rachel-fenichel: I'm still planning to play with these values (and invite other people to play with them).

core/connection.js
rachel-fenichel: Done

core/connection.js
rachel-fenichel: Removed.

core/connection.js
rachel-fenichel: !!!

I'm unsure of what cases would end up here.  I think it's worth having a log statement here, which will let me debug until all cases are covered.  Any objections?

RoboErikG: sgtm

rachel-fenichel:
I added tests; PTAL.

core/insertion_marker_manager.js
```diff
@@ -0,0 +1,710 @@
...
+    this.highlightedBlock_ = closest.sourceBlock_;
+    // TODO: remove?
+    closest.sourceBlock_.highlightShapeForInput(closest, true);
+  }
+  this.highlightingBlock_ = true;
+};
+
+/**
+ * Get rid of the highlighting marking the block that will be replaced.
+ * Scratch-specific code, where "highlighting" applies to a block rather than
```
RoboErikG: "Scratch-specific code" then why is it here?

core/insertion_marker_manager.js
```diff
@@ -0,0 +1,710 @@
...
+  }
+};
+
+/**** End preview visibility functions ****/
+
+/**** Begin block highlighting functions ****/
+
+/**
+ * Add highlighting showing which block will be replaced.
+ * Scratch-specific code, where "highlighting" applies to a block rather than
```
RoboErikG: 'Scratch specific code'?

core/insertion_marker_manager.js
```diff
@@ -0,0 +1,710 @@
...
+   * Set in initAvailableConnections, if at all
+   * @type {Blockly.BlockSvg}
+   * @private
+   */
+  this.lastMarker_ = null;
+
+  /**
+   * The insertion marker that shows up between blocks to show where a block
+   * would go if dropped immediately.
+   * This is the scratch-blocks equivalent of connection highlighting.
```
RoboErikG: remove ref to scratch-blocks.

core/insertion_marker_manager.js
```diff
@@ -0,0 +1,710 @@
...
+   * Updated on every mouse move.
+   * @type {boolean}
+   * @private
+   */
+  this.wouldDeleteBlock_ = false;
+
+  /**
+   * Connection on the insertion marker block that corresponds to
+   * this.localConnection_ on the currently dragged block.
+   * This is part of the scratch-blocks equivalent of connection highlighting.
```
RoboErikG: remove scratch-blocks ref.

PR #1991, opened 2018-07-31
Some goog.dom removal

NeilFraser:
Another small step for deleting Closure.


core/inject.js
```diff
@@ -54,7 +53,7 @@ Blockly.inject = function(container, opt_options) {
         document.querySelector(container);
   }
   // Verify that the container is in document.
-  if (!goog.dom.contains(document, container)) {
+  if (!Blockly.utils.containsNode(document, container)) {
```
AnmAtAnm: `container` is an unfortunate variable name for the second parameter, given the container actually tested is `document`.

Issue #1994, opened 2018-08-03
Unplugging and healing stack blocks with mismatched types


rachel-fenichel:
### Problem statement

`block.unplug` takes in one argument, `opt_healStack`.  

If the block is in a stack of statements and `opt_healStack` is true, the block will be unplugged and the next and previous blocks will be connected.  
If `opt_healStack` is false, the block will be unplugged from the previous block, and its next block will stick with it.  

### Expected Behavior

If the next and previous blocks have incompatible connections, it behaves like `opt_healStack` is false.

### Actual Behavior

If the next and previous blocks have incompatible connections, the three blocks are all disconnected at the end of `unplug`.

PR #1995, opened 2018-08-03
Heal rows when unplugging, and add tests

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

None.

### Proposed Changes

Extend the `opt_healStack` behaviour in `block.unplug` to work for rows as well as stacks.
### Reason for Changes

Removing an insertion marker should heal the stack around it.

### Test Coverage
Added new JSUnit tests to cover various stack and row cases, including bad connection checks and bad numbers of connections.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -340,12 +340,6 @@ Blockly.Block.prototype.unplugFromRow_ = function(opt_healStack) {
     return;
   }
 
-  var parentBlock = parentConnection.getSourceBlock();
-  if (parentConnection != parentBlock.getOnlyValueConnection_()) {
-    // Too many or too few possible connections on the parent.
-    return;
-  }
-
   var thisConnection = this.getOnlyValueConnection_();
   if (!thisConnection || !thisConnection.isConnected()) {
     // Too many or too few possible connections on this block, or there's
--- a/core/block.js
+++ b/core/block.js
@@ -314,27 +314,98 @@ Blockly.Block.prototype.initModel = function() {
  */
 Blockly.Block.prototype.unplug = function(opt_healStack) {
   if (this.outputConnection) {
-    if (this.outputConnection.isConnected()) {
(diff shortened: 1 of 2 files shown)


core/block.js
```diff
@@ -314,27 +314,98 @@ Blockly.Block.prototype.initModel = function() {
...
+    this.outputConnection.disconnect();
+  }
+
+  // Return early in obvious cases.
+  if (!parentConnection || !opt_healStack) {
+    return;
+  }
+
+  var parentBlock = parentConnection.getSourceBlock();
+  if (parentConnection != parentBlock.getOnlyValueConnection_()) {
```
RoboErikG: This check doesn't make sense. Why does it matter if there are other connections when we know where the block was connected?

core/block.js
```diff
@@ -314,27 +314,98 @@ Blockly.Block.prototype.initModel = function() {
...
+    previousTarget = this.previousConnection.targetConnection;
+    // Detach this block from the parent's tree.
+    this.previousConnection.disconnect();
+  }
+  var nextBlock = this.getNextBlock();
+  if (opt_healStack && nextBlock) {
+    // Disconnect the next statement.
+    var nextTarget = this.nextConnection.targetConnection;
+    nextTarget.disconnect();
+    // Shouldn't checkType happen earlier?
```
RoboErikG: Yes it should if you're keeping the behavior here and in rows the same.

rachel-fenichel: https://github.com/google/blockly/issues/1994 (mentioned in a todo below, and I'll add the todo here as well).

This is the preexisting behaviour.  I think it's wrong, but I don't think it's a big deal right now and I want to deal with it in a separate PR either way.

core/block.js
rachel-fenichel: Fixed, and test updated.

PR #2019, opened 2018-08-23
Disable/enable function calls along with their definitions

### LINKED ISSUE #1344: Disable function calls when disabling the function definition
### Bug Report

Disabling a function definition in the workspace may leave the system in a broken state, where any calls to function remain enable and generate code with no matching definition for the identifier.

screen shot 2017-09-27 at 1 32 59 pm

### Expected Behavior

Similar to how deleting a function definition deletes the matching function calls, disabling a function definition should also disable the matching calls.

### Actual Behavior

Function call is left enabled, and any generated code is broken.

### Steps to Reproduce

1. Add a user defined function.
2. Add a call to the function.
3. Disable the definition, but not the call.

RoboErikG:
Fixes #1344 

## The basics

- [x] I branched from develop

### Resolves

Fixes #1344 

### Proposed Changes

Extends the event listener on procedure caller blocks to also check
for their definition being enabled/disabled and update their own
state in response.

### Reason for Changes

If you generate code for a function caller when the definition is disabled you can get errors in most generated languages.

### Test Coverage

Tested on:
Desktop Chrome

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2019)


rachel-fenichel:
Did you check how it works with undo/redo?

RoboErikG:
I did manually. I ran out of time to write tests for undo/redo.

blocks/procedures.js
```diff
@@ -840,6 +846,21 @@ Blockly.Blocks['procedures_callnoreturn'] = {
         this.dispose(true, false);
         Blockly.Events.setGroup(false);
       }
+    } else if (event.type == Blockly.Events.CHANGE && event.element == 'disabled') {
+      var name = this.getProcedureCall();
+      var def = Blockly.Procedures.getDefinition(name, this.workspace);
+      if (def && def.id == event.blockId) {
+        // in most cases the old group should be ''
```
AnmAtAnm: Elaborate on this assumption.  What are the side effects if not true, and maybe hint at the alternative solution of injecting to the undo stack before the other group.

tests/workspace_svg/procedure_svg_test.js
```diff
@@ -0,0 +1,373 @@
...
+            && 'g2' == firedEvents[2].group);
+
+
+  } finally {
+    procedureSvgTest_teardown();
+  }
+}
+
+
+function test_procedureReturnEnablingRemembersOldCallerState() {
```
AnmAtAnm: Rename (fixing name collision).  This is a no-return proc test.

PR #2020, opened 2018-08-23
Remove goog.ui.ColorPicker

NeilFraser:
Closure spells “Colour” wrong, so we shouldn’t be using this library.

Adds the ability to set the tooltip titles for each colour.

Note that Closure’s colour picker in Blockly used to have keyboard controls (as of > 3 years ago); but no longer seems to.  This replacement does not currently have keyboard controls either.

### DIFF
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -29,9 +29,7 @@ goog.provide('Blockly.FieldColour');
 goog.require('Blockly.Field');
 goog.require('Blockly.utils');
 
-goog.require('goog.events');
 goog.require('goog.style');
-goog.require('goog.ui.ColorPicker');
 
 
 /**
@@ -68,6 +66,13 @@ Blockly.FieldColour.fromJson = function(options) {
  * @private
  */
 Blockly.FieldColour.prototype.colours_ = null;
+
+/**
+ * By default use the global constants for colour tooltips.
+ * @type {Array.<string>}
+ * @private
+ */
+Blockly.FieldColour.prototype.titles_ = null;
 
 /**
  * By default use the global constants for columns.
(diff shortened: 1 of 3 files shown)


core/field_colour.js
```diff
@@ -69,6 +67,13 @@ Blockly.FieldColour.fromJson = function(options) {
  */
 Blockly.FieldColour.prototype.colours_ = null;
 
+/**
+ * By default use the global constants for colour tooltips.
```
rachel-fenichel: Clarify that null makes it use the global constants--I could see at a glance that this was null by default, but not what null meant.

core/field_colour.js
```diff
@@ -185,52 +222,79 @@ Blockly.FieldColour.prototype.showEditor_ = function() {
...
+      'mouseup', this, this.onClick);
 };
 
 /**
- * Create a color picker widget and render it inside the widget div.
- * @return {!goog.ui.ColorPicker} The newly created color picker.
+ * Handle a click on a colour cell.
+ * @param {!Event} e Mouse event.
+ */
+Blockly.FieldColour.prototype.onClick = function(e) {
```
rachel-fenichel: Can this be private?

core/field_colour.js
```diff
@@ -138,11 +143,38 @@ Blockly.FieldColour.prototype.getText = function() {
 
 /**
  * An array of colour strings for the palette.
- * See bottom of this page for the default:
- * http://docs.closure-library.googlecode.com/git/closure_goog_ui_colorpicker.js.source.html
+ * Copied from goog.ui.ColorPicker.SIMPLE_GRID_COLORS
```
rachel-fenichel: Add: all colour pickers will use this set of colours unless overridden with setColours.

core/field_colour.js
NeilFraser: Done.

core/field_colour.js
NeilFraser: Done.

core/field_colour.js
NeilFraser: Done.

PR #1754, opened 2018-03-31
Add support for transitioning the workspace block and bubble canvas. 

samelhusseini:
Add support for transitioning the workspace block and bubble canvas. 
Enable for scrollCenter.

It turns out only Chrome supports CSS transitions on SVG elements, so feel free not to pull in this change. Cool demo though!

transitioncanvas

## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

No current issue open. 

### Proposed Changes

Animate the workspace move when scrolling. Support for developers to begin and end the transition property on the block and bubble canvas.

### Reason for Changes

Animations are great!

### Test Coverage

Tested on:
Only works on Chrome

PR #2022, opened 2018-08-24
Add support for transitioning the workspace block and bubble canvases on scrollCenter

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

This is just a new version of #1754.  From that description:

> Animate the workspace move when scrolling. Support for developers to begin and end the transition property on the block and bubble canvas.

>Add support for transitioning the workspace block and bubble canvas.
>
> Enable for scrollCenter.
> 
> It turns out only Chrome supports CSS transitions on SVG elements, so feel free not to pull in this change. Cool demo though!

### Reason for Changes

> Animations are great!

### Test Coverage
Only works on chrome.

### Additional Information
I tested what happens if you pick up a block during this transition, and it can be significantly offset from the mouse (tested by setting the transition to 5 seconds).  However, with a .5s transition it's hard to make that happen, and the offset is only during that one drag--the blocks don't end up in an overall bad state.  I think we should take this change, since it makes things just a bit nicer on chrome.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -166,7 +166,7 @@ Blockly.Css.CONTENT = [
 
   '.blocklyBlockCanvas.blocklyCanvasTransitioning,',
   '.blocklyBubbleCanvas.blocklyCanvasTransitioning {',
-    'transition: transform 0.5s;',
+    'transition: transform .5s;',
   '}',
 
   '.blocklyTooltipDiv {',
--- a/core/css.js
+++ b/core/css.js
@@ -164,6 +164,11 @@ Blockly.Css.CONTENT = [
     'z-index: 50;', /* Display below toolbox, but above everything else. */
   '}',
 
+  '.blocklyBlockCanvas.blocklyCanvasTransitioning,',
+  '.blocklyBubbleCanvas.blocklyCanvasTransitioning {',
+    'transition: transform 0.5s;',
+  '}',
+
   '.blocklyTooltipDiv {',
     'background-color: #ffffc7;',
     'border: 1px solid #ddc;',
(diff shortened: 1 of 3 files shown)


core/zoom_controls.js
```diff
@@ -302,7 +302,11 @@ Blockly.ZoomControls.prototype.createZoomResetSvg_ = function(rnd) {
   Blockly.bindEventWithChecks_(zoomresetSvg, 'mousedown', null, function(e) {
     ws.markFocused();
     ws.setScale(ws.options.zoomOptions.startScale);
+    ws.beginCanvasTransition();
     ws.scrollCenter();
+    setTimeout(function() {
+      ws.endCanvasTransition();
+    }, 500);
```
NeilFraser: How does this behave in a race condition where the user clicks the button twice in less than half a second?  Change the timings to 10 seconds to see what happens.

You might need to keep the PID generated by setTimeout in a variable, and clear the timeout if another call is made.

core/css.js
```diff
@@ -164,6 +164,11 @@ Blockly.Css.CONTENT = [
     'z-index: 50;', /* Display below toolbox, but above everything else. */
   '}',
 
+  '.blocklyBlockCanvas.blocklyCanvasTransitioning,',
+  '.blocklyBubbleCanvas.blocklyCanvasTransitioning {',
+    'transition: transform 0.5s;',
```
NeilFraser: CSS style guide says to just use .5, whereas the JS style guide says to use 0.5.

core/zoom_controls.js
AnmAtAnm: Add transition to zoom-in and zoom-out buttons?

rachel-fenichel: It shouldn't ever end in a bad state, since it's not relying on the PID to know when to end the canvas transition.  So the worst case is that it jumps too fast, and that should hardly be noticeable.  The end state will be the correct CSS.

rachel-fenichel: Filed as #2023 after discussion--using the same code doesn't make it look smooth with multiple zooms.

Issue #2034, opened 2018-08-30
field_image opt_onClick not work


duzc2:
### Problem statement
when use a opt_onClick to field_image
it does not work when the image clicked.

### Expected Behavior

invoke the function when click the image

### Actual Behavior

nothing

### Steps to Reproduce

1. 
.appendField(
            new window.Blockly.FieldImage('https://www.gstatic.com/codesite/ph/images/star_on.gif', 15, 15, '*', function () {
                alert(1);
            }));
2. 
click the image
### Additional Information
to fix it
replace Blockly.FieldImage.prototype.maybeAddClickHandler_ by :
Blockly.FieldImage.prototype.maybeAddClickHandler_ = function () {
    if (this.clickHandler_) {
        this.mouseDownWrapper_ =
            Blockly.bindEventWithChecks_(
                this.fieldGroup_, 'mousedown', this, this.clickHandler_);
    }
};

AnmAtAnm:
Dupe of #1887

PR #2046, opened 2018-09-10
fix #1887 #2034

### LINKED ISSUE #1887: Image field: click handler gets ignored
### Problem statement

Image fields accept click handlers but do not invoke them.

### Expected Behavior

Clicking on the field's image should result in a call to the click handler.

### Actual Behavior

It does not.

### Steps to Reproduce

1. Create a block with an image field that has a click handler.
2. Try to click the block.

Desktop Chrome

### Additional Information

Setting the field's EDITABLE property to *true* makes it behave as expected.

duzc2:
## The basics

- [x ] I branched from develop

### Resolves

#1887
#2034

### Test Coverage

Tested on:
Desktop Chrome

### DIFF
--- a/core/field_image.js
+++ b/core/field_image.js
@@ -135,7 +135,7 @@ Blockly.FieldImage.prototype.maybeAddClickHandler_ = function() {
   if (this.clickHandler_) {
     this.mouseDownWrapper_ =
         Blockly.bindEventWithChecks_(
-            this.fieldGroup_, 'mousedown', this, this.onMouseDown_);
+            this.fieldGroup_, 'mousedown', this, this.clickHandler_);
   }
 };
 


rachel-fenichel:
There's a comment on #1887 that says 

"i am observing same behaviour in chrome and in firefox 60
works in firefox 52.8.1 (32-bit)"

Why would it work in one and not the other? 

This fix looks reasonable, but I wanted to make sure I understand what problem it's actually solving.

duzc2:
> There's a comment on #1887 that says
> 
> "i am observing same behaviour in chrome and in firefox 60
> works in firefox 52.8.1 (32-bit)"
> 
> Why would it work in one and not the other?
> 
> This fix looks reasonable, but I wanted to make sure I understand what problem it's actually solving.

i don't know why it works in ff52 , but I think it should not work...
sorry , I did not test ..

it just did not register the function as a callback .
there is no code reference the clickHandler_ , it has never been invoked.

PR #2069, opened 2018-09-25
Fix inconsistent return type warning

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Closure compiler warning


rachel-fenichel:
But it was always falsey so hey! it just worked.

PR #2078, opened 2018-10-11
Honour reductions to MAX_UNDO

NeilFraser:
If workspace.MAX_UNDO is reduced, we now throw away the extra records in the stack, as opposed to just not growing it anymore.

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -523,7 +523,7 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
   if (event.recordUndo) {
     this.undoStack_.push(event);
     this.redoStack_.length = 0;
-    if (this.undoStack_.length > this.MAX_UNDO) {
+    while (this.undoStack_.length > this.MAX_UNDO) {
       this.undoStack_.unshift();
     }
   }
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -524,7 +524,7 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
     this.undoStack_.push(event);
     this.redoStack_.length = 0;
     while (this.undoStack_.length > this.MAX_UNDO) {
-      this.undoStack_.unshift();
+      this.undoStack_.shift();
     }
   }
   for (var i = 0, func; func = this.listeners_[i]; i++) {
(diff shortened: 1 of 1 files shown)


core/workspace.js
```diff
@@ -523,8 +523,8 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
   if (event.recordUndo) {
```
RoboErikG: Is it worth also making this an && this.MAX_UNDO > 0?

core/workspace.js
```diff
@@ -523,8 +523,8 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
...
     this.undoStack_.push(event);
     this.redoStack_.length = 0;
-    if (this.undoStack_.length > this.MAX_UNDO) {
-      this.undoStack_.unshift();
+    while (this.undoStack_.length > this.MAX_UNDO) {
+      this.undoStack_.shift();
```
RoboErikG: Not for this PR, but we may want to improve this later to size it by the number of undo groups instead of individual events.

core/workspace.js
NeilFraser: Done, though it should be >= rather than > since 0 is valid.

core/workspace.js
NeilFraser: As far as I know, in the real world this setting will be either zero, or a "big number".  The "big number" is just a limit to prevent swamping memory.  I can't think of a use-case where someone would actually want the number of undo operations to be a specific number such as 6.  In the "big number" case, what we actually care about is bytes, not undo events or undo groups.

Thus I don't think that counting groups would help anyone.

core/workspace.js
RoboErikG: If it's 0 we shouldn't add anything to the stack.

NeilFraser: Yes, but we can't short-circuit as you suggest since we do need to empty the stack.

RoboErikG: Ah, good point. I'd been assuming this was only set at initialization when the undo stack would be empty anyway. In that case >= is still needed to keep from crashing on negative values.

RoboErikG:
It looks like we're targeting next week for a push to master, so should be available in a couple weeks unless we break something. ;)

PR #2080, opened 2018-10-12
Remove all calls to goog.dom.createDom

NeilFraser:
All XML nodes are now created using `Blockly.Xml.utils.createElement`.  All browser DOM nodes are now created using `document.createElement`.


NeilFraser:
FYI, when the dependencies change, Travis' error message is completely useless.  It's a random code dump from Travis' codebase.  Recompiling was a wild guess, and it worked.

core/mutator.js
```diff
@@ -33,8 +33,8 @@ goog.require('Blockly.Events.Ui');
 goog.require('Blockly.Icon');
 goog.require('Blockly.utils');
 goog.require('Blockly.WorkspaceSvg');
-
-goog.require('goog.dom');
+goog.require('Blockly.Xml');
```
rachel-fenichel: Why do you need both of these requires?

core/xml_utils.js
```diff
@@ -0,0 +1,77 @@
...
+goog.provide('Blockly.Xml.utils');
+
+
+/**
+ * Create DOM element for XML.
+ * @param {string} tagName Name of DOM element.
+ * @return {!Element} New DOM element.
+ */
+Blockly.Xml.utils.createElement = function(tagName) {
+  // TODO: Namespace this element.
```
rachel-fenichel: Todos should refer to github issue numbers.

core/xml_utils.js
```diff
@@ -0,0 +1,77 @@
...
+ * @name Blockly.Xml.utils
+ * @namespace
+ */
+goog.provide('Blockly.Xml.utils');
+
+
+/**
+ * Create DOM element for XML.
+ * @param {string} tagName Name of DOM element.
+ * @return {!Element} New DOM element.
```
rachel-fenichel: Mark all functions @public if you want to expose them fully, otherwise @package.

core/mutator.js
NeilFraser: This file contains two calls to `Blockly.Xml.utils.createElement` and two calls to `Blockly.Xml.domToText`.  I could just require `Blockly.Xml` and rely on the (current) transitive dependency, but that's fragile.

rachel-fenichel: Okay.  Leaving it as-is sounds good to me.

PR #2104, opened 2018-11-05
Immovable blocks should not move when workspace is cleaned

NeilFraser:
To recreate the issue reported by James Ma, go to the playground, and import this XML:
```
  <block type="controls_whileUntil" x="13" y="13"></block>
  <block movable="false" type="logic_negate" x="188" y="88"></block>
</xml>
```
Then right-click on workspace and cleanup blocks.  The 'not' block is immovable yet moves to a new location.

With this PR, the 'not' block does not move.


NeilFraser:
Googlebot, what have you been smoking this weekend?

NeilFraser:
> googlebot commented 2 minutes from now
> CLAs look good, thanks!

Thanks *future* Googlebot.

PR #2108, opened 2018-11-06
Remove all goog.math.Size uses

NeilFraser:
Reviewing the individual commits is probably better than the squashed whole.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -32,7 +32,6 @@ goog.require('Blockly.Events.BlockChange');
 goog.require('Blockly.Gesture');
 goog.require('Blockly.utils');
 
-goog.require('goog.math.Size');
 goog.require('goog.style');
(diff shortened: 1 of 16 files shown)


core/field.js
```diff
@@ -45,7 +44,8 @@ goog.require('goog.userAgent');
  * @constructor
  */
 Blockly.Field = function(text, opt_validator) {
-  this.size_ = new goog.math.Size(0, Blockly.BlockSvg.MIN_BLOCK_Y);
```
rachel-fenichel: Another option is to leave the .size_ property in here and just turn it into an object with .width_ and .height_.  Disadvantage: extra indirection.  Advantage: not breaking alternative renderers (e.g. makecode)

rachel-fenichel:
Can I close this?

Issue #1201, opened 2017-06-29
PHP generator test failure: Lists


rachel-fenichel:
Error message: `syntax error, unexpected ';', expecting :: (T_PAAMAYIM_NEKUDOTAYIM)`

TonyLianLong:
The generator writes code like this:
```
$list2 = array('Picard', 'Riker', 'Crusher');
  (true ? $list2 : null)[0] = 'Jean-Luc';
  assertEquals($list2, array('Jean-Luc', 'Riker', 'Crusher'), 'set first order list');
```
which would not work in PHP thus requiring modification.

AnmAtAnm:
These file is pared down to just the blocks in question. The second has the full `test set` test definition, just disabled.
 * http://anmatanm.github.io/blockly/blockly-issue1201-php-list-syntax-error.xml
 * http://anmatanm.github.io/blockly/blockly-issue1201-php-list-syntax-error-full.xml

AnmAtAnm:
I don't see an obvious solution to this, because I don't see a way to coerce an arbitrary expression into a reference variable in PHP.

When the `LIST` input is anything but a `variables_get` block, `cacheList()` generates this syntactically bad code. Thankfully, this `RANDOM` case is the only call to `cacheList()`.

    function test_set_random_with_syntax_error() {
      global $test_name, $number_of_calls, $list2, $changing_list, $list_copy, $text, $unittestResults;
      $list2 = array('Picard', 'Riker', 'Crusher');
      $tmp_list = &(true ? $list2 : null);                        // Error occurs here.
      $tmp_x = rand(0, count($tmp_list)-1);
      $tmp_list[$tmp_x] = 'Data';
      assertEquals(length($list2), 3, 'set random order list');
    }

However, we can pass the ternary into a helper function to perform the set, we get a much clearer message `"Only variables can be passed by reference"`:

    function set_random_helper(&$tmp_list) {
      $tmp_x = rand(0, count($tmp_list)-1);
      $tmp_list[$tmp_x] = 'Data';
    }

    function test_set_random() {
      global $test_name, $number_of_calls, $list2, $changing_list, $list_copy, $text, $unittestResults;
      $list2 = array('Picard', 'Riker', 'Crusher');
      set_random_helper(true ? $list2 : null);
      assertEquals(length($list2), 3, 'set random order list');
    }

For future reference, I used the following blocks to run these tests:

    <xml xmlns="http://www.w3.org/1999/xhtml">
      <variables>
[...]

PR #2111, opened 2018-11-09
Run generator tests on Travis

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Generator tests have not previously been scripted.

### Proposed Changes

For each language:
- run generator
- save generated code to a tmp file
- compare the generated file against a golden file

I also updated some of the test names to get rid of name conflicts in functions between text and list test, which were causing some tests to not be run when loading all of the tests with appendDomToWorkspace.

I verified this fix by checking that the same number of tests are run when loading test suites individually and when loading them all together.

### Reason for Changes

Generator tests are frustrating.

### Test Coverage
Tested by running the run_all_tests script locally.  I also ran the golden code for each language and verified that all tests pass.  For PHP I removed the known broken generated code (see #1201) and verified that the rest of the generated code works.


tests/generators/run_generators_in_browser.js
```diff
@@ -0,0 +1,145 @@
...
+      })
+      .url(url)
+      .then(function() {
+        console.log('about to load');
+        this.execute(function() {
+          checkAll();
+          loadSelected();
+        })
+      })
+      .pause(10000)
```
RoboErikG: Are 10 second pauses the only way to make sure this works?

tests/scripts/run_generators.sh
```diff
@@ -0,0 +1,56 @@
...
+TMP_DIR="tests/generators/tmp/"
+GOLDEN_DIR="tests/generators/golden/"
+
+FAILURE_COUNT=0
+check_result() {
+  local suffix=$1 # One of: js, py, dart, lua, php
+  local tmp_filename="${TMP_DIR}generated.$suffix"
+  if [ ! -f $tmp_filename ]; then
+    echo "File $tmp_filename not found!"
+    FAILURE_COUNT=$((FAILURE_COUNT+1))
```
RoboErikG: Should this return if the file doesn't exist?

tests/scripts/run_generators.sh
```diff
@@ -0,0 +1,56 @@
...
+  fi
+
+  local golden_filename="${GOLDEN_DIR}generated.$suffix"
+  if [ ! -f $golden_filename ]; then
+    echo "File $golden_filename not found!"
+    FAILURE_COUNT=$((FAILURE_COUNT+1))
```
RoboErikG: Ditto.

tests/scripts/run_generators.sh
rachel-fenichel: Changed to do the next steps if the file is found, and otherwise simply report the error.

tests/scripts/run_generators.sh
rachel-fenichel: Ditto.

tests/generators/run_generators_in_browser.js
rachel-fenichel: It shouldn't be.  I'll look into it.

rachel-fenichel:
Followup issues: #2114, #2115, #2116, #2117, #2118.

PR #2113, opened 2018-11-14
[rendering] update build script

samelhusseini:
Updated build script to support 'core' and 'scratch' rendering. 
Blocks are shared, 'core' will exclude the scratch rendering file and 'scratch' will exclude the core renderer (block_render_svg). 

'core' generates blockly_compressed.js
'scratch' generates blockly_scratch_compressed.js

Thoughts on what we want to call this? Can probably re-visit that later. 
I also added a scratch_playground which is a dupe of playground.html but references the blockly_scratch_compressed.js

Note: calling this script currently will fail as block_render_svg_scratch.js doesn't exist yet. 
I've left accessible as is.

### DIFF
--- a/build.py
+++ b/build.py
@@ -59,7 +59,7 @@ if sys.version_info[0] != 2:
 
 for arg in sys.argv[1:len(sys.argv)]:
   if (arg != 'core' and
-      arg != 'scratch' and
+      arg != '2018_rendering' and
       arg != 'accessible' and
       arg != 'generators' and
       arg != 'langfiles'):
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
Let's use `2018_rendering` as the temporary name, instead of `scratch`.  So `blockly_2018_rendering_compressed.js`, `2018_rendering` as the flag, etc.  We can revisit it to pick a proper name, but this can help us avoid some confusion in the interim.

The additional playground is fine.  Eventually we'll make that an option on the main playground.

samelhusseini:
Updated: 
build.py 2018_rendering
Playground: 2018_rendering_playground.html
Render SVG file: block_render_svg_2018_rendering.js
Output: blockly_2018_rendering_(un)compressed.js

Issue #2112, opened 2018-11-12
Upgrade build.py to python 3


rachel-fenichel:
`build.py` currently requires Python 2.  

We should support python 3, while still working for either.  Developers using python 2 should not see a breakage when we add a python 3 version.

This can be a separate script if that's easiest.

We also need a deprecation warning in the current script, and possibly a change to the developer documentation.  The developer documentation change needs to be done by a Blockly team member and can be done separately once we have the new script.

PR #2123, opened 2018-11-16
Make build.py compatible with Python 2 and 3

gomercin:
## The basics
- [x] I branched from develop

## The details
Following changes are made with minimal touches in the rest;
- version dependent module imports
- codecs.open instead of open (for encoding errors)
- string_is_ascii modified for better compatibility
- In create_messages.py, synonym_defs are sorted as well to increase
  testability, as Python 3 does not guarantee key order.
  Note that this will cause previously generated files to have different
  orders from the files generated without this patch.
- small changes; parenthesis around "print"s, has_key -> in,
  "except x, y" -> "except x as y", sorted -> sort

### Resolves
https://github.com/google/blockly/issues/2112

### Reason for Changes
Enabling the upgrade to Python3.x

### Test Coverage

Tested on:
Tested on Windows by comparing generated files using:
- original code with Python 2.7
- patched code with Python 2.7
- patched code with Python 3.6

There is a change of order due to sorting synonym_defs but there seems to be no easy
solution to keep the order from the original files.

### Additional Information
Other files need similar small changes but are not touched as they are not
referenced from build.py

### DIFF
--- a/build.py
+++ b/build.py
@@ -53,9 +53,6 @@
 #   msg/js/<LANG>.js for every language <LANG> defined in msg/js/<LANG>.json.
 
 import sys
-if sys.version_info[0] != 2:
-  raise Exception("Blockly build only compatible with Python 2.x.\n"
-                  "You are using: " + sys.version)
 
 for arg in sys.argv[1:len(sys.argv)]:
   if (arg != 'core' and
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
Thank you!  And feel free to make followup PRs to change any other python files that we have that only run in Python 2.

gomercin:
Thank you too, I will take a look at them as well

PR #2125, opened 2018-11-18
Fix issues related to Python 3 compatibility

gomercin:
## The basics
- [x] I branched from develop

## The details
Most changes are for except usages and paranthesis around print. 
decode had to be added for Python 3.4 compatibility, in build.py#341, because of changes in return type of response.read() and supported argument types for json.loads between 2.7, 3.4 and 3.6

### Resolves
Further updates following up https://github.com/google/blockly/issues/2112

### Reason for Changes
Making remaining python files compatible with Python 2 and 3 versions. 
This patch also covers Python 3.4 compatibility for build.py

### Test Coverage
All files are inspected by PyCharm

build.py and dedup_json.py is tested on Ubuntu 16.04 by comparing generated files using:

- original code with Python 2.7
- patched code with Python 2.7
- patched code with Python 3.4
- patched code with Python 3.6

json_to_js and create_messages are indirectly used by build. 

xliff_to_json.py is not tested but changes are not important to affect the code flow. 

### Additional Information

tests.py does not seem to be testing an existing function, so it might need to be removed or updated. 

Using "." for relative imports are encouraged (like "from .common import InputError" ) but it makes scripts trickier to run so they are not changed.

### DIFF
--- a/build.py
+++ b/build.py
@@ -336,7 +336,9 @@ class Gen_compressed(threading.Thread):
     conn = httplib.HTTPSConnection("closure-compiler.appspot.com")
     conn.request("POST", "/compile", urlencode(params), headers)
     response = conn.getresponse()
-    json_str = response.read()
+
+    # Decode is necessary for Python 3.4 compatibility
+    json_str = response.read().decode("utf-8")
     conn.close()
 
     # Parse the JSON response.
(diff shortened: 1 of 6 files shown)


gomercin:
Commenting for CLA check

gomercin:
Signed CLA with commit account as well, should hopefully work now

PR #2133, opened 2018-11-21
Generate Code For Single Block

### LINKED ISSUE #2067: Generate code for single block in stack only?
### Problem statement

I have a stack of blocks, but I only want to generate code for a single block in that stack. Blockly however appends the generated code of subsequent blocks to the one I call 
Blockly.JavaScript.blockToCode() for.

The only workaround I've found is to detach child blocks, generate the code for the block I actually want, and then reattach them. Is there an easier way to do this?

### Expected Behavior

I would like some way of getting `Blockly.JavaScript.blockToCode(block)` not to call the generators of all subsequent blocks in the same stack as the one I want to generate.

### Actual Behavior

All child block code is generated and automatically attached to the block's code that I'm interested in. This causes me all sorts of issues as in on my use case I only want the code for that very specific block even though it has multiple child blocks in its stack.

alschmiedt:
## The basics

### Resolves
#2067

### Proposed Changes
Allows a user to generate code for a single block by passing a parameter to the blockToCode method.

### Reason for Changes
Helpful for allowing the users to step through blocks one by one.

### Test Coverage
Tests can be found in tests/jsunit/generator_test.js. The tests check that the blockToCode method returns only a single block for all the generators. 

Tested on:

### DIFF
--- a/core/generator.js
+++ b/core/generator.js
@@ -158,11 +158,12 @@ Blockly.Generator.prototype.allNestedComments = function(block) {
 /**
  * Generate code for the specified block (and attached blocks).
  * @param {Blockly.Block} block The block to generate code for.
+ * @param {boolean} opt_thisOnly True to generate code for only this block.
  * @return {string|!Array} For statement blocks, the generated code.
  *     For value blocks, an array containing the generated code and an
  *     operator order value.  Returns '' if block is null.
  */
-Blockly.Generator.prototype.blockToCode = function(block) {
+Blockly.Generator.prototype.blockToCode = function(block, opt_thisOnly) {
   if (!block) {
     return '';
   }
@@ -186,13 +187,13 @@ Blockly.Generator.prototype.blockToCode = function(block) {
     if (!block.outputConnection) {
       throw TypeError('Expecting string from statement block: ' + block.type);
     }
-    return [this.scrub_(block, code[0]), code[1]];
+    return [this.scrub_(block, code[0], opt_thisOnly), code[1]];
   } else if (typeof code == 'string') {
     var id = block.id.replace(/\$/g, '$$$$');  // Issue 251.
     if (this.STATEMENT_PREFIX) {
       code = this.STATEMENT_PREFIX.replace(/%1/g, '\'' + id + '\'') + code;
     }
-    return this.scrub_(block, code);
+    return this.scrub_(block, code, opt_thisOnly);
   } else if (code === null) {
     // Block has handled code generation itself.
     return '';
--- a/core/generator.js
+++ b/core/generator.js
@@ -158,7 +158,7 @@ Blockly.Generator.prototype.allNestedComments = function(block) {
 /**
  * Generate code for the specified block (and attached blocks).
  * @param {Blockly.Block} block The block to generate code for.
- * @param {boolean} opt_thisOnly True to generate code for only this block.
+ * @param {boolean=} opt_thisOnly True to generate code for only this statement.
  * @return {string|!Array} For statement blocks, the generated code.
  *     For value blocks, an array containing the generated code and an
  *     operator order value.  Returns '' if block is null.
@@ -169,7 +169,7 @@ Blockly.Generator.prototype.blockToCode = function(block, opt_thisOnly) {
   }
   if (block.disabled) {
     // Skip past this block if it is disabled.
-    return this.blockToCode(block.getNextBlock());
+    return opt_thisOnly ? '' : this.blockToCode(block.getNextBlock());
   }
 
   var func = this[block.type];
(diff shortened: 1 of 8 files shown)


core/generator.js
```diff
@@ -158,11 +158,12 @@ Blockly.Generator.prototype.allNestedComments = function(block) {
 /**
  * Generate code for the specified block (and attached blocks).
  * @param {Blockly.Block} block The block to generate code for.
+ * @param {boolean} opt_thisOnly True to generate code for only this block.
```
NeilFraser: The JS Doc (here and below) would be {boolean=} to show it's optional.

generators/dart.js
```diff
@@ -221,7 +222,7 @@ Blockly.Dart.scrub_ = function(block, code) {
     }
   }
   var nextBlock = block.nextConnection && block.nextConnection.targetBlock();
-  var nextCode = Blockly.Dart.blockToCode(nextBlock);
+  var nextCode = opt_thisOnly ? "" : Blockly.Dart.blockToCode(nextBlock);
```
NeilFraser: Single quotes preferred (here and below).

core/generator.js
NeilFraser: Also, it should be "this statement" or something to that effect, since a value block would generate code for all its nested blocks, and a loop block would generate code for all its nested statements.

core/generator.js
```diff
@@ -158,11 +158,12 @@ Blockly.Generator.prototype.allNestedComments = function(block) {
...
  * @return {string|!Array} For statement blocks, the generated code.
  *     For value blocks, an array containing the generated code and an
  *     operator order value.  Returns '' if block is null.
  */
-Blockly.Generator.prototype.blockToCode = function(block) {
+Blockly.Generator.prototype.blockToCode = function(block, opt_thisOnly) {
   if (!block) {
     return '';
   }
```
NeilFraser: What happens if a statement block is disabled and blockToCode is called on it with opt_thisOnly set true?  I'd expect empty string, but my reading of the code indicates that the entire remainder of the stack is returned.

tests/jsunit/generator_test.js
```diff
@@ -19,10 +19,85 @@
...
+  workspace = new Blockly.Workspace();
+}
+
+function generatorTest_tearDown() {
+  undefineTestBlocks();
+  workspace.dispose();
+}
+
+function blockToCodeTest(generator, opt_thisOnly) {
+  var row_block = workspace.newBlock("row_block");
```
NeilFraser: Single quotes.  (The JSON above is fine, that does require double quotes.)

tests/jsunit/generator_test.js
```diff
@@ -19,10 +19,85 @@
...
+  var stack_block = workspace.newBlock("stack_block");
+
+  generator.row_block = (block) => {return "row_block"};
+  generator.stack_block = (block) => {return "stack_block"};
```
NeilFraser: To my knowledge, this would represent the only use of ES6 in Blockly.  Since this is a test, it probably won't break the compiler's settings.  None the less, it's probably safer to just use ES5 here.

tests/jsunit/generator_test.js
```diff
@@ -19,10 +19,85 @@
...
+
+  assertEquals(blockToCodeTest(Blockly.PHP, true), 'row_block');
+  assertEquals(blockToCodeTest(Blockly.PHP, false), 'row_blockstack_block');
+
+  assertEquals(blockToCodeTest(Blockly.Python, true), 'row_block');
+  assertEquals(blockToCodeTest(Blockly.Python, false), 'row_blockstack_block');
+
+  generatorTest_tearDown();
+}
+
```
RoboErikG: - Add a set of tests for the first block in a stack being disabled.
- Add a set of tests for nested statement and nested input blocks also being generated when opt_thisOnly is set to true.

tests/jsunit/generator_test.js
```diff
@@ -19,10 +19,169 @@
...
+    "args1": [{
+      "type": "input_statement",
+      "name": "DO"
+    }],
+    "previousStatement": null,
+    "nextStatement": null
+  }]);
+}
+
+function undefineTestBlocks() {
```
rachel-fenichel: Name this `undefineGeneratorTestBlocks`.  Otherwise it (unfortunately) conflicts with a previous declaration of `undefineTestBlocks` in block_test.js

alschmiedt:
Fixes #2067

Maclaon:
how should i use it correctly？
the `master` branch does not container this feature! when i apply [Advanced Compilation](https://developers.google.com/blockly/guides/modify/web/advanced) to `develop` branch, it won't work!

rachel-fenichel:
This is now live on master, since we just pushed our our December release.

Issue #1105, opened 2017-05-13
RFE: Delete variables without creating a new variable field block


AnmAtAnm:
Currently, variables do not delete themselves, even if there is no reference to them. In part, this is by design, because the creation of a variable occurs before a variable block is added to the workspace (via the toolbox "Create Variable" button).

Currently, the only way to delete such blocks is via create a block with a variable field, select the otherwise unused variable from the drop down, and then open the drop down again to select delete.

My intuitive expectation was the variable would autodelete after the last reference was removed (variable fields or function arguments). However, one would have to be careful this does not delete created variables before they are first used.

Alternatively, an affordance on the toolbox (perhaps only for unused variables).

rachel-fenichel:
We are definitely not autodeleting after the last reference is removed--that was an explicit decision in the new design.

Right now unused variables don't get serialized, I think.  That might change in the future.

Either way it sounds like this is a request for a 'delete variable' button.  It would either:
- let you pick the variable (or possibly variables) from a list, or
- clean up all unreferenced variables in one pass.

BeksOmega:
You can do this now by right clicking on the block in the toolbox.

PR #2143, opened 2018-11-28
Delete variable

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/1105 

### Proposed Changes
Adds a delete button to the default variables toolbox category.
screen shot 2018-11-28 at 2 51 34 pm

### Reason for Changes

There was no way before to delete a variable without dragging it into the workspace.

### Test Coverage

Tested on:

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -149,15 +149,19 @@ Blockly.Variables.allDeveloperVariables = function(workspace) {
  */
 Blockly.Variables.flyoutCategory = function(workspace) {
   var xmlList = [];
-  var button = document.createElement('button');
-  button.setAttribute('text', '%{BKY_NEW_VARIABLE}');
-  button.setAttribute('callbackKey', 'CREATE_VARIABLE');
+  xmlList.push(Blockly.Variables.createButton_(
+      '%{BKY_NEW_VARIABLE}', 'CREATE_VARIABLE'));
 
+  xmlList.push(Blockly.Variables.createButton_(
+      '%{BKY_DELETE_VARIABLE_BUTTON_TITLE}', 'DELETE_VARIABLE'));
+
+  workspace.registerButtonCallback('DELETE_VARIABLE', function(deleteButton) {
+    Blockly.Variables.deleteVariableButtonHandler(deleteButton.getTargetWorkspace());
+  });
   workspace.registerButtonCallback('CREATE_VARIABLE', function(button) {
     Blockly.Variables.createVariableButtonHandler(button.getTargetWorkspace());
   });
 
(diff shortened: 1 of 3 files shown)


core/variables.js
```diff
@@ -324,9 +328,44 @@ Blockly.Variables.createVariableButtonHandler = function(
...
+ * Handles "Delete Variable" button in the default variables toolbox category.
+ * It will prompt the user for a varibale name, including re-prompts if a name
+ * does not exist.
+ *
+ * @param {!Blockly.Workspace} workspace The workspace on which to delete the
+ *     variable.
+ */
+Blockly.Variables.deleteVariableButtonHandler = function(workspace) {
+  var promptAndCheckWithAlert = function(defaultName) {
+    Blockly.Variables.promptName(Blockly.Msg['DELETE_VARIABLE_PROMPT'], defaultName,
```
RoboErikG: Instead of using the existing text field prompt can you make one with a dropdown with all existing variables?

core/variables.js
```diff
@@ -581,6 +620,19 @@ Blockly.Variables.createVariable_ = function(workspace, id, opt_name,
   return variable;
 };
 
+/**
+ * Helper function to create a button for the flyout window.
```
rachel-fenichel: Note that this creates the XML for the button, not the button itself.  That's because the XML then gets passed into the flyout code, which creates the button onscreen.

samelhusseini:
Just throwing this out in case you find this helpful, but another way to achieve this without having another button in the flyout could be to use the variables_get context menu and have a delete option there. So a user could open the variables flyout and right click on the variable they're trying to delete and delete it. You already have the list of variables visible to the user to select which they're trying to delete.

rachel-fenichel:
@samelhusseini I like that plan, and I think it also matches what Scratch does.

@alschmiedt, @RoboErikG, thoughts?  It seems more elegenat than the alternatives.

samelhusseini:
Yep. I've added both the option to rename and delete. Not sure about @scratch

alschmiedt:
This sounds good to me as well. Thanks for the suggestion.

RoboErikG:
SGTM. Thanks, Sam. =)

alschmiedt:
Some thoughts for a more visible delete buttons. 
If we had all the time in the world we would want to create an x to the right of the variables in the flyout. 
Another (more time convenient) option would be to create a deleteAllUnusedVariables function. This would allow developers the ability to create a deleteVariables button in the flyout if necessary.

PR #2147, opened 2018-11-30
Allow statement connections to connect around block stacks during drags

### LINKED ISSUE #2131: C shaped blocks don't wrap around other blocks (insertion markers, develop)
### Problem statement
C shaped blocks don't wrap around other blocks.

### Expected Behavior
Wrap around the block, which I would say is the equivalent of the highlight marker in master. 

### Actual Behavior
Doesn't wrap around.

### Steps to Reproduce
1. Drag out any statement block
2. Drag out another C shaped block and try to wrap it around the initial block
3. Doesn't work.

Blockly master behaviour: 
current

Blockly develop behaviour:
new

Desktop Chrome

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2131 

### Proposed Changes

Allow statement inputs to connect around block stacks during drags.
Remove distinction between first statement connection and other statement connections.

### Reason for Changes

See #2131
### Test Coverage

Tested with the example in #2131, and with an if-else block to test the case with more than one statement input.

### Additional Information

This makes it possible to connect on a statement input, but it's still not a great experience--the insertion marker is a little too stubborn about sticking around, so you have to drag away from the target and then back to switch which of the statement inputs you're connecting to.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -355,29 +355,19 @@ Blockly.Connection.prototype.canConnectToPrevious_ = function(candidate) {
     return false;
   }
 
-  var firstStatementConnection =
-      this.sourceBlock_.getFirstStatementConnection();
-  var isFirstStatementConnection = this == firstStatementConnection;
-  var isNextConnection = this == this.sourceBlock_.nextConnection;
-
-  if (isNextConnection || isFirstStatementConnection) {
-    // If the candidate is the first connection in a stack, we can connect.
-    if (!candidate.targetConnection) {
-      return true;
-    }
-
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@alschmiedt fyi

rachel-fenichel:
Actually, the stickiness was because there was a check in InsertionMarkerManager.shouldUpdatePreviews_ that was checking if the proposed closest connection was the same as the old closest connection, but didn't check if the local connection had changed.  Fixed in the third commit.

PR #2149, opened 2018-12-03
Tweak insertion marker snap radii

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Insertion markers were too sticky when dragging a block with a statement input down through a stack of blocks.

### Proposed Changes

Decrease the radius to search for good connections.  This makes sense--the old radius was from scratch blocks, and their blocks are taller.

Also increase the preference for the current connection.

### Reason for Changes

It was not possible to drag an if block down one block through a stack of statement blocks.

### Test Coverage
Tested on the playground

### Additional Information

We can continue to tweak this, but these numbers feel okay to me.

### DIFF
--- a/core/constants.js
+++ b/core/constants.js
@@ -42,20 +42,20 @@ Blockly.FLYOUT_DRAG_RADIUS = 10;
 /**
  * Maximum misalignment between connections for them to snap together.
  */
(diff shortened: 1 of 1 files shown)


alschmiedt:
LGTM : )

Issue #2136, opened 2018-11-24
Trashcan Contents


BeksOmega:
### Problem statement

Feature Request: Trash Can Contents.

### Expected Behavior

From [Site](https://developers.google.com/blockly/guides/modify/future#improved_ui): Clicking on the trash can should bring up a bubble that contains everything that has been deleted. These blocks may be dragged out of the trash and back into the workspace.

### Actual Behavior

Logs a TODO message to console, saying that this should be added.

### Steps to Reproduce

1. Click the trash can.

### Stack Traces

Not Applicable.

### Operating System and Browser

All of them.

### Additional Information

I would like to take a crack at this. Here's the bad version:

1) Have the trashcan subscribe to BlockDelete events, and add the oldXml to a "contents" var.
2) Have the trashcan create a flyout in its init() function.
3) Have the trashcan call .show(contentsVar) on the flyout when it is clicked.

![img_20181124_091221](https://user-images.githubusercontent.com/25440652/48971002-36b38600-efc9-11e8-8d2d-113538980d5b.jpg)

I'm a bit unsure of how the contents should appear. Should they be in a bubble like mutators? A fullsidebar like the toolbox? If the workspace layout it set to horizontal, should the contents be horizontal as well?

RoboErikG:
Just a note, this is related to https://github.com/google/blockly/issues/225 

Internally, we've discussed doing some refactoring to support drop targets that could include a backpack. Listening to the delete event seems like a good short term option for just the trash, though I'm not sure how you'd remove things from the trash if the user undoes the delete (or if the list gets too big). Another option would be to scan the last N events in the undo stack each time the trash is opened, though this might be slower.

@rachel-fenichel any comments on design?

rachel-fenichel:
- Removing things from trash: do we need to?
- XML for contents, gotten from the deletion events, sounds good to me
- UI: a flyout that shows up on the right side of the workspace (for LTR) sounds like a good place to start.  We can worry about horizontal later.  
- App Inventor has something similar for their backpack. 
- Adding better drop targets (such as a backpack) is related, but not necessary to make this work--block deletion is easy to catch from events.

I think you should start by looking at the app inventor code for displaying and dragging out backpack contents, and try to make something equivalent happen in Blockly.  Once you have that, correctly populating the trashcan based on events should be easy.  

@ewpatton can you point to the relevant code in App Inventor for the backpack rendering and interactions?

Issue #2035, opened 2018-08-30
Horizontal bottom toolbar: Anchor zoom/trash buttons to top corner


AnmAtAnm:
### Problem statement

When the toolbox is a horizontal toolbar attached to the bottom, the zoom move as they are pressed. The zoom causes the blocks in the toolbox change size, and the toolbox to change height, and the buttons are currently positioned relative to the top of the toolbox.

Zooming in or out multiple times involves chasing the button.

### Expected Behavior

The buttons should not move when they are pressed.

Recommend anchoring the buttons to the top corner (right in LTR, left in RTL).

### Actual Behavior

The buttons move up and down as the toolbox height changes.

### Steps to Reproduce

1. Start by opening `tests/multi_playground.html`.
2. Set the toolbox style to simple.
3. On the last row of workspaces, find the two where the toolbox is attached to the bottom.
4. Press either zoom button in either workspace. Notice how the button moves up or down.

PR #2157, opened 2018-12-07
Added flyout to trashcan to "get back" deleted blocks.`

BeksOmega:
## The basics

- [x] I branched from develop

### Resolves

#2136

### Proposed Changes

Adds a flyout that appears ontop of the trashcan (and zoom controls) when the trashcan is clicked. You can then drag deleted blocks out of the trashcan.

Requirements:
1.There should be indication that the trashcan has blocks: When the trashcan has contents the lid is tilted & animates on hover.
2.Deleted blocks should render in an intuitive order: Newest at the top.
3.Deleted blocks should be easy to get: 1 click and 1 drag to retrieve, and the flyout always renders ontop of the trashcan.
4.The flyout should not be cluttered: Only 1 instances of every block (or group of blocks) is added to the flyout. A MAX_CONTENTS property limits the number of blocks (or groups of blocks).

### Reason for Changes

Using keyboard controls (ctrl z) could be unintuitive for young users. This creates a way that they can get back their accidentally deleted blocks purely through UI.

### Test Coverage

Requirement 1:
1. Dragged a 'controls_if' block into the trashcan. Did steps 2-5. (pass)
2. Observed how the lid is now tilted.
3. Hovered over the trashcan.
4. Observed how the lid animates (only animates when over the 'body' of the trashcan to avoid jitter).
5. Refreshed page.
6. Dragged a 'controls_if' block onto the workspace and pressed the delete key. Repeated 2-5. (pass)
7. Dragged a 'controls_if' block onto the workspace and deleted through the right-click context menu. Repeated 2-4. (pass)

Requirement 2:
1. Dragged a 'controls_if' block into the trashcan.
2. Dragged a 'logic_compare' block into the trashcan.
3. Clicked the trashcan to open the flyout.
4. Observed how the 'logic_compare' block was above the 'controls_if' block. (pass)

Requirement 3:
1. Added a block to the trashcan.
2. Clicked the trashcan.
3. Observed how the flyout opened over the trashcan, and that blocks could be dragged out of it.
4. Repeated for all combinations of RTL/LTR and Start/End. (pass)

Requirement 4:
1. Set the trashcan's MAX_CONTENTS var to 1.
1. Dragged 2 'controls_if' blocks to the trashcan.
2. Clicked the trashcan.
3. Observed how there was only 1 'control_if' block in the trashcan. (pass)
4. Dragged a 'logic_compare' block into the trashcan.
5. Clicked the trashcan.
6. Observed how there was no 'controls_if' block in the flyout, only a 'logic_compare' block. (pass)

Tested on:
* Desktop Chrome

### Note: Blocks Never get Removed From Trashcan (at least atm)

When you drag things out of the trashcan you'd think they should be removed from the trashcan's flyout. The problem is that (atm) there's no way to tell what flyout you grabbed the block from, or if it was un-deleted through events (as far as I know). If you do remove the blocks from the flyout whenever you receive a 'BLOCK_CREATED' event you end up in a situtation like this:

removecontents

Where you didn't drag the block out of the trashcan's flyout, (you created it from the toolbox) but it still gets removed.

A couple solutions I came up with (and why they suck):
1. Override a method on the flyout (possibly the .blockMouseDown_ method) and have it notify the trashcan when a block is created. This is not maintainable.
2. Don't allow 'primitive' (existing in the toolbox) blocks to be added to the flyout. A) I couldn't get workspace.options.languageTree.contains(myXML) to work when I tried it. B) A bit unintuitive.
3. Add an observer pattern to the flyout so it can notify you when a block is created. This is really only useful for the trashcan so it seems like overkill.

Erik's suggestion about looking through the undo stack for BLOCK_DELETE events would have worked, until you drag a block out of the trashcan's flyout which adds a create event to the undo stack. Then you're back to comparing creates vs deletes, which is the original problem.

So that's the thought proccess behind not removing blocks from the trashcan. If anyone has a better idea, or sees where my logic has gaps I would still love to add this feature!

### Note on Reviewing this PR:

I had to move around where some things like the trashcan, and zoom controls get created & initialized so that the trashcan's flyout could be initialized properly. So please double check that to make sure I didn't break anything.

### Additional Info
Hoping to work on #2035 and getting the contents flyout to work horizontally next.

### DIFF
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -121,6 +121,7 @@ Blockly.Trashcan.prototype.HAS_BLOCKS_LID_ANGLE = 0.1;
  * The maximum number of blocks that can go in the trashcan's flyout. '0' turns
  * turns off trashcan contents, 'Infinity' sets it to unlimited.
  * @type {number}
+ * @private
  */
 Blockly.Trashcan.prototype.MAX_CONTENTS = 32;
 
@@ -134,20 +135,23 @@ Blockly.Trashcan.prototype.isOpen = false;
  * The minimum openness of the lid. Used to indicate if the trashcan contains
  *  blocks.
  * @type {number}
+ * @private
  */
 Blockly.Trashcan.prototype.minOpenness_ = 0;
 
 /**
  * True if the trashcan contains blocks, otherwise false.
  * @type {boolean}
+ * @private
  */
-Blockly.Trashcan.prototype.hasBlocks = false;
+Blockly.Trashcan.prototype.hasBlocks_ = false;
 
 /**
  * A list of Xml representing blocks "inside" the trashcan.
  * @type {Array}
+ * @private
  */
-Blockly.Trashcan.prototype.contents = [];
+Blockly.Trashcan.prototype.contents_ = [];
 
 /**
  * The SVG group containing the trash can.
@@ -296,6 +300,7 @@ Blockly.Trashcan.prototype.dispose = function() {
  * Move the trash can to the bottom-right corner.
  */
 Blockly.Trashcan.prototype.position = function() {
+  // Not yet initialized.
   if (!this.bottom_) {
     return;
   }
@@ -402,15 +407,11 @@ Blockly.Trashcan.prototype.close = function() {
  * Inspect the contents of the trash.
  */
 Blockly.Trashcan.prototype.click = function() {
-  if (!this.hasBlocks) {
+  if (!this.hasBlocks_) {
     return;
   }
 
-  var xml = [];
-  for (var i = 0, text; text = this.contents[i]; i++) {
-    xml[i] = Blockly.Xml.textToDom(text).firstChild;
-  }
-  this.flyout_.show(xml);
+  this.flyout_.show(this.contents_);
 };
 
 /**
@@ -418,7 +419,7 @@ Blockly.Trashcan.prototype.click = function() {
  * @private
  */
 Blockly.Trashcan.prototype.mouseOver_ = function() {
-  if (!this.hasBlocks) {
+  if (!this.hasBlocks_) {
     return;
   }
   this.setOpen_(true);
@@ -447,17 +448,21 @@ Blockly.Trashcan.prototype.onDelete_ = function() {
       return;
     }
     if (event.type == Blockly.Events.BLOCK_DELETE) {
-      var textBlock = trashcan.blockXmlToText(event.oldXml);
-      if (trashcan.contents.indexOf(textBlock) != -1) {
+      var cleanedXML = trashcan.cleanBlockXML_(event.oldXml);
+      console.log('cleaned XML:');
+      console.log(cleanedXML);
+      if (trashcan.contents_.indexOf(cleanedXML) != -1) {
         return;
       }
-      trashcan.contents.unshift(textBlock);
-      if (trashcan.contents.length > trashcan.MAX_CONTENTS) {
-        trashcan.contents.splice(trashcan.MAX_CONTENTS,
-            trashcan.contents.length - trashcan.MAX_CONTENTS);
+      trashcan.contents_.unshift(cleanedXML);
+      console.log('added to contents: ');
+      console.log(trashcan.contents_);
+      if (trashcan.contents_.length > trashcan.MAX_CONTENTS) {
+        trashcan.contents_.splice(trashcan.MAX_CONTENTS,
+            trashcan.contents_.length - trashcan.MAX_CONTENTS);
       }
 
-      trashcan.hasBlocks = true;
+      trashcan.hasBlocks_ = true;
       trashcan.minOpenness_ = trashcan.HAS_BLOCKS_LID_ANGLE;
       trashcan.setLidAngle_(trashcan.minOpenness_ * 45);
     }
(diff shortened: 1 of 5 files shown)


core/trashcan.js
```diff
@@ -309,20 +368,28 @@ Blockly.Trashcan.prototype.setOpen_ = function(state) {
...
     this.lidTask_ = setTimeout(this.animateLid_.bind(this), 20);
   }
 };
 
+/**
+ * Set the angle of the trashcan's lid.
+ * @param {!number} lidAngle The angle at which to set the lid.
+ * @private
+ */
+Blockly.Trashcan.prototype.setLidAngle_ = function(lidAngle) {
```
BeksOmega: I'm not sure why this isn't updating the screen-image when called from onDelete_. From my testing the attribute is getting set, it's just not updating what you see.

BeksOmega: Fixed

rachel-fenichel:
Hey there--sorry for the delay on reviewing this.  It's taking a bit of extra time because it's large and I needed to pull the code down and play with the behaviour.  I'll have a review out shortly.

core/trashcan.js
```diff
@@ -335,10 +402,80 @@ Blockly.Trashcan.prototype.close = function() {
...
+};
+
+/**
+ * Converts xml representing a block into text that can be stored in the
+ *    content array.
+ * @param {!Element} xml An XML tree defining the block and any
+ *    connected child blocks.
+ * @returns {string} The XML tree converted into a usable text string.
+ */
+Blockly.Trashcan.prototype.blockXmlToText = function(xml) {
```
rachel-fenichel: A couple of comments here:
- You're removing the ID on the root block, but not on any descendants.  That means that you can get multiply copies of the same block group in the flyout, like this:
image

- You're deleting the next block(s).  We'd like to keep it.  Deleting a block stack and wanting to get the entire stack back seems like a common use case.

core/trashcan.js
```diff
@@ -335,10 +402,80 @@ Blockly.Trashcan.prototype.close = function() {
...
-  var dy = this.workspace_.startScrollY - this.workspace_.scrollY;
-  if (Math.sqrt(dx * dx + dy * dy) > Blockly.DRAG_RADIUS) {
+  if (!this.hasBlocks) {
     return;
   }
-  console.log('TODO: Inspect trash.');
+
+  var xml = [];
+  for (var i = 0, text; text = this.contents[i]; i++) {
+    xml[i] = Blockly.Xml.textToDom(text).firstChild;
```
rachel-fenichel: Here's the current flow:
- A block is deleted
- The block is serialized to XML on the event
- The trashcan change listener grabs that XML
- The trashcan change listener modifies the XML (remove coordinates, etc)
- The trashcan change listener converts the XML to text and stores it as text
- The trashcan is clicked
- This code converts the text to XML
- The flyout goes from XML to blocks

You can just store the contents as XML instead of as text, to get rid of a couple of conversion steps.

core/trashcan.js
```diff
@@ -96,12 +109,46 @@ Blockly.Trashcan.prototype.SPRITE_LEFT_ = 0;
...
+
+/**
+ * True if the trashcan contains blocks, otherwise false.
+ * @type {boolean}
+ */
+Blockly.Trashcan.prototype.hasBlocks = false;
+
+/**
+ * A list of Xml representing blocks "inside" the trashcan.
+ * @type {Array}
```
rachel-fenichel: This and `hasBlocks` can be private.  `private` in closure annotations [means](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#private) it can only be used in the same file, so your use of these in the delete handler will be fine.

core/trashcan.js
```diff
@@ -96,12 +109,46 @@ Blockly.Trashcan.prototype.SPRITE_LEFT_ = 0;
...
+ *    (0.0 = closed, 1.0 = open)
+ * @type {number}
+ * @private
+ */
+Blockly.Trashcan.prototype.HAS_BLOCKS_LID_ANGLE = 0.1;
+
+/**
+ * The maximum number of blocks that can go in the trashcan's flyout. '0' turns
+ * turns off trashcan contents, 'Infinity' sets it to unlimited.
+ * @type {number}
```
rachel-fenichel: Tag everything private where possible.

core/trashcan.js
```diff
@@ -96,12 +109,46 @@ Blockly.Trashcan.prototype.SPRITE_LEFT_ = 0;
...
+ */
+Blockly.Trashcan.prototype.MAX_CONTENTS = 32;
```
rachel-fenichel: File an issue to eventually set the trashcan max contents through the options struct, and you can do it as a followup PR.

core/trashcan.js
rachel-fenichel: Yes.  If anyone wants access later, we can add a getter function.

core/trashcan.js
BeksOmega: Ok I changed it over, and I'm getting some wierd behavior. Instead of saving the XML to the contents array, it seems like it's converting it into a block object? Which is cool but useless for comparing :/

adding to contents behavior

I did steal the converting to text thing from app inventor, so I assume they ran into the same problem and just decided to convert to text.

I left the logs in the latest commit so you can see if you so desire.

rachel-fenichel: Okay, let's go with text for simplicity.  Thanks!

kushal4:
Hi,
Could you please guide me as to how do i checkout this update

kushal4:
Found it  :)

Issue #1902, opened 2018-06-04
audio play() called before user interaction


AnmAtAnm:
New requirements in Chrome 67 require the user to interact with the page before Audio.play() will succeed.  The call to `play()` in `Blockly.WorkspaceAudio.preload()` now fails with the following message:

    DOMException: play() failed because the user didn't interact with the document first.

NeilFraser:
There is code already present which is supposed to handle that.
inject.js lines 403-409

I guess 'mousemove' is no longer considered an interaction.

AnmAtAnm:
> I guess 'mousemove' is no longer considered an interaction.
Correct.  Now it requires a click.

ewpatton:
We often embed assets (e.g., the cursors) as data URIs so that they are available immediately after the sources are loaded (although not in this particular case). Would that work for Blockly's use case? One concern might be bloating the compiled version, but you could have a separate JS file with just the base64 encoded assets.

AnmAtAnm:
@ewpatton Interesting idea.  We currently attempt to minimize the audio download size by offering multiple audio file formats (each platform has their own list of compatible of file format and compression formats). It probably wouldn't make sense to do that if we were embedding the files in the code, especially with the +33% increase that base64 encoding will likely incur.

If someone could do an updated evaluation of supported audio formats, going back to IE 11 and including mobile platforms, we could understand if we could reduce the number of required audio files (ideally down to three, one per audible event). (Note to self: make a test page of all nine audio files as data URIs.)

If this is true, we could use your proposed method and get audio on offline instances without requiring the app developer to define the `media/` directory option.

ewpatton:
@AnmAtAnm I would expect that the supported audio types will be fixed for a given (os, browser, version) tuple. One could have a JS file per audio type and detect which script to load at runtime. Here's a rough sketch of what I am thinking:

```
// sounds.mp3.js:
Blockly.SoundData["click"] = "data:audio/mp3;base64,...";
...
```

```
// sounds.ogg.js:
Blockly.SoundData["click"] = "data:audio/ogg;base64,...";
...
```

```
// sounds.wav.js:
Blockly.SoundData["click"] = "data:audio/wav;base64,...";
```

```
// workspace_audio.js:
Blockly.WorkspaceAudio.loadSoundData = function(preferredFormats) {
  for (var i = 0; i < preferredFormats.length; i++) {
    if (/* test for preferredFormat[i] support */) {
      // Add <script> tag for formatted sound data.
    }
[...]

PR #2162, opened 2018-12-10
Handle promise rejection for sound.play() in preload

epicfaace:
## The basics

- [x] I branched from develop

### Resolves

audio play() called before user interaction #1902

Handle promise rejection for sound.play() in preload. This removes the error, "play() failed because the user didn't interact with the document first."

### Test Coverage

Tested on:
* Desktop Chrome

Mac OS X

### DIFF
--- a/core/workspace_audio.js
+++ b/core/workspace_audio.js
@@ -111,7 +111,7 @@ Blockly.WorkspaceAudio.prototype.preload = function() {
   for (var name in this.SOUNDS_) {
     var sound = this.SOUNDS_[name];
     sound.volume = 0.01;
-    sound.play();
+    sound.play().catch(function() {});
     sound.pause();
     // iOS can only process one sound at a time.  Trying to load more than one
     // corrupts the earlier ones.  Just load one and leave the others uncached.


epicfaace:
@RoboErikG thanks, I've fixed those problems.

Why doesn't Blockly use Babel?

AnmAtAnm:
@RoboErikG  @rachel-fenichel 
I think this is still a problem for older browsers. The return value of `play()` will be undefined in older browsers, and that needs to be tested before calling `catch(..)`.
Unfortunately, this was pushed into the latest release.

rachel-fenichel:
@AnmAtAnm so does this now fail differently for older browsers, and in a way that we need to be concerned about?  Or is it just that this solution doesn't get rid of the warning in the older browsers?

tuxianchao:
In my project, [play-request-was-interrupted](https://developers.google.com/web/updates/2017/06/play-request-was-interrupted#error) this is how I solved the problem.

NeilFraser:
Any changes to sound need to be tested on all browsers (including IE), iOS, and Android.  Sound is a surprisingly immature and inconsistent beast.  The current Rube Goldberg method in which Blockly plays sound is a result of playing whack-a-mole until a working solution emerged.  Specs and logic are irrelevant, it's all driven by the random implementations of each browser.

tuxianchao:
> @RoboErikG @rachel-fenichel
> I think this is still a problem for older browsers. The return value of `play()` will be undefined in older browsers, and that needs to be tested before calling `catch(..)`. Fixing this would probably make this a solid solution.
> Unfortunately, this was pushed into the latest release.
> 
> @epicfaace I don't think Babel would fix this particular case.

may be like this,check play() method return promise before pause()
```
var playPromise =  sound.play();
if(playPromise !== undefined){
     playPromise.then(function(){
         sound.pause();
     }).catch(function(error){
          console.error(error);
         //....
     });
}

```

PR #2163, opened 2018-12-12
Propagate the event group when enforcing bounds on blocks

### LINKED ISSUE #1653: Dragging a block off the edge breaks undo in fixed size workspaces
### Problem statement

If you drag a block out of bounds in a fixed size workspace such that it is snapped back to the edge of Blockly Undo becomes unable to go past that event.

### Expected Behavior

Undo can revert changes back to the start.

### Actual Behavior

The undo stack gets stuck.

### Steps to Reproduce

1. go to https://blockly-demo.appspot.com/static/demos/fixed/index.html
2. Drag two blocks onto the workspace
3. Drag one of the blocks down past the edge of the browser window and release so it snaps back to the edge of the workspace.
4. Do any number of other changes.
5. Try to undo back past the snapped drag.

### Stack Traces

No error reported.

 * Desktop Chrome

RoboErikG:
Fixes #1653

Makes the bounds checker created in inject correctly set the event group
before moving blocks in bounds.

## The basics

- [x] I branched from develop

### Resolves

#1653 

### Proposed Changes

Propagate the event group in the bounds checker callback.

### Reason for Changes

When a workspace can't scroll the inject method sets up an event listener
to bump blocks back into the visible bounds when dropped off the edge. The
event group wasn't being used, though, causing the move to be a separate
undo step. When undone, this would get you stuck in a loop where the
workspace would keep pulling it back into the visible range.

### Test Coverage

Tested on:
Desktop Chrome

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2163)

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -224,7 +224,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
   Blockly.mainWorkspace = mainWorkspace;
 
   if (!options.readOnly && !options.hasScrollbars) {
-    var workspaceChanged = function() {
+    var workspaceChanged = function(e) {
       if (!mainWorkspace.isDragging()) {
         var metrics = mainWorkspace.getMetrics();
         var edgeLeft = metrics.viewLeft + metrics.absoluteLeft;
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Interesting.  So you're changing the group on the event after it's been fired?

RoboErikG:
No? I'm setting the group for the followup move events that are triggered by the event. This is the same way the bumping gets done.

rachel-fenichel:
I inverted the setGroup when I read it.  Whoops.

rachel-fenichel:
And if the previous event didn't have a group?

rachel-fenichel:
Yes, that would be good.  After that lgtm.

PR #2167, opened 2018-12-13
Reconnects child blocks when an undo is done on an else block

### LINKED ISSUE #2037: controls_if: Mutation actions do not undo correctly
### Problem statement

Undo does not work correctly when `controls_if` mutator edit actions are interleaved with normal actions.

### Expected Behavior

One undo step should undo to the previous step, maintaining connections unaffected by the edit.

### Actual Behavior

A block connected to an if block where an undo reverts a mutation may also disconnect the child, even when the connected input is not (supposed to be) affected by the mutation.

### Steps to Reproduce

1. Open the playground and create the following structure:
screen shot 2018-08-30 at 2 16 52 pm
2. Next, make the following edits:
  a. Add an "else if" in the mutator.
  b. Connect the loop block to the newly created second statement connection
  c. Add an "else" to the end, within the mutator.

The results should look like this:
screen shot 2018-08-30 at 2 27 54 pm

3. Undo one step, removing the last connection.
screen shot 2018-08-30 at 2 29 23 pm

Note that making what was supposed to be a single step, editing the last connection via the mutator workspace, it has also disconnected the loop block.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#2037 

### Proposed Changes
Reconnects the children of an if/else block after an undo is made.

### Reason for Changes

screen shot 2018-12-13 at 9 16 25 am
Before Undo

screen shot 2018-12-13 at 9 16 31 am
After Undo 

When a user tries to undo adding an else block it rebuilds the if/else block by removing all inputs and reconstructing with the appropriate number of elseif and else  inputs. When removing the inputs it disconnects any statements attached to the inputs.

### Test Coverage

Tested on:
Desktop Chrome 
Desktop Firefox
Desktop Opera

### DIFF
--- a/blocks/logic.js
+++ b/blocks/logic.js
@@ -424,7 +424,6 @@ Blockly.Constants.Logic.CONTROLS_IF_MUTATOR_MIXIN = {
   },
   /**
    * Reconstructs the block with all child blocks attached.
-   * Fixes issue #2037.
    */
   rebuildShape_: function() {
     var valueConnections = [null];
--- a/blocks/logic.js
+++ b/blocks/logic.js
@@ -331,7 +331,7 @@ Blockly.Constants.Logic.CONTROLS_IF_MUTATOR_MIXIN = {
   domToMutation: function(xmlElement) {
     this.elseifCount_ = parseInt(xmlElement.getAttribute('elseif'), 10) || 0;
     this.elseCount_ = parseInt(xmlElement.getAttribute('else'), 10) || 0;
-    this.updateShape_();
+    this.rebuildShape_();
   },
   /**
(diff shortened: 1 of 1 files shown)


blocks/logic.js
```diff
@@ -425,6 +422,30 @@ Blockly.Constants.Logic.CONTROLS_IF_MUTATOR_MIXIN = {
           clauseBlock.nextConnection.targetBlock();
     }
   },
+  /**
+   * Reconstructs the block with all child blocks attached.
+   * Fixes issue #2037.
```
RoboErikG: Don't need this comment about which issue the code fixes.

PR #2177, opened 2018-12-17
Prevent duplicate parameters in functions

phptuts:
## The basics

- [x] I branched from develop

### Resolves

Prevents the coder from using duplicate parameters names in functions.

### Proposed Changes

In the validator function check for parameter names that have been used.  If it finds a block with a name that matches return null so that it will revert the change.   Also change the name each time a new parameter is created to try to prevent duplicate name.

blockly-pr

### Reason for Changes

Duplicate parameter names do not exist in any language to my knowledge.  They cause syntax errors and are a pain.

### Test Coverage

I did not write any tests.

Tested on:

I tested this on chrome, firefox, opera, and safari.  I did not test this on mobile or on internet explorer.


phptuts:
I did sign the release, I am not sure how to get Github to refresh the check.

phptuts:
I think it's signed now.  👍

blocks/procedures.js
```diff
@@ -468,13 +453,28 @@ Blockly.Blocks['procedures_mutatorcontainer'] = {
...
+var parameterCounter = 1;
+
 Blockly.Blocks['procedures_mutatorarg'] = {
   /**
    * Mutator block for procedure argument.
    * @this Blockly.Block
    */
   init: function() {
-    var field = new Blockly.FieldTextInput('x', this.validator_);
+    var paramName = "param" + parameterCounter.toString();
```
RoboErikG: Instead of changing the default name of the block it should be changed in the copies of the block when they're added to the mutator workspace. This will keep the behavior consistent with how new functions get named as they're dragged onto the workspace.

blocks/procedures.js
```diff
@@ -468,13 +453,28 @@ Blockly.Blocks['procedures_mutatorcontainer'] = {
   }
 };
 
+var parameterCounter = 1;
```
RoboErikG: Using a global variable like this is going to create problems. Instead you'll want to do something similar to how new procedure names are generated, but just scanning the blocks in the mutator workspace. See [findLegalName](https://github.com/google/blockly/blob/7679d227df8f62a12316334ee708d4a9a800ef17/core/procedures.js#L89).

phptuts:
@RoboErikG 
Thank you for reviewing the pr.  I had to create a bunch of logic in a method called onchange to do what you wanted.  I also created a bunch of variables in the dialog box's workspace.

Here is what it looks like, let me know what you think.

blockly-pr-2

blocks/procedures.js
```diff
@@ -501,6 +487,58 @@ Blockly.Blocks['procedures_mutatorarg'] = {
...
     field.onFinishEditing_('x');
   },
+  /**
+   * This will create & delete variables and in dialogs workspace to ensure
+   * that when a new block is dragged out it will have a unique parameter name.
+   * @param {!Blockly.Events.Abstract} event Change event.
+   * @this Blockly.Block
+   */
+  onchange: function(event) {
+    if (!this.workspace || this.workspace.isFlyout ||
```
RoboErikG: The onchange method isn't actually doing anything. Mutators don't fire create or delete events for the sub blocks in the mutator. The global workspace fires a change event instead for the block being mutated. So this part of the code will always return immediately.

blocks/procedures.js
```diff
@@ -517,6 +555,16 @@ Blockly.Blocks['procedures_mutatorarg'] = {
     if (!varName) {
       return null;
     }
+    // Prevents duplicate parameter names in functions
```
RoboErikG: This bit of code is all you needed.

blocks/procedures.js
RoboErikG: Hrmm, I'm mistaken. It looks like it does fire those events, they just aren't logged. Investigating.

RoboErikG: Sorry about the confusion. When I tested it it seemed like variables weren't getting deleted and since the events don't get logged I thought this wasn't running. On further testing, it's actually that if you delete more than one variable at a time only the variable for the top-most block gets removed, which causes variables to leak in the mutator workspace. For delete events you'll want to go through the list of field tags and delete all the variables that are being deleted, not just the first one, and then this should work as expected.

Testing:
1. Drag out 3 input name blocks which should create 'x', 'i', 'j' variables.
2. Delete the entire stack.
3. Drag out 3 more input name blocks.

Expected: Creates 'x', 'i', 'j' variables.

Current: Creates 'x', 'k', 'm' variables

RoboErikG:
Oh! Almost missed it, please make your PR against Develop, not Master.

RoboErikG:
Hrm...it looks like you branched from master so you're picking up additional changes to the compressed files, xml, and package.json.

To fix this you'll want to update develop and then rebase:

1. Update develop
git fetch upstream
git checkout develop
git pull upstream develop
# optionally push back to your GitHub copy
git push origin HEAD:develop

2. rebase your change
git checkout duplicate-parameter
git rebase develop

phptuts:
@googlebot  I think I signed it again

phptuts:
@RoboErikG 
Thank you for working with me on this one.  

So I got it to work with that test case.  I also noticed that delete event was not being triggered.  To compensate for that, I just run the clean on every event which seems work.  Let me know what you think.  

Thank you 

Noah


blockly-pr-3

phptuts:
@googlebot I confirm that I am good with the commits being in here.

phptuts:
@RoboErikG 
Thank you for all your help and with the reviewing.  I think I got everything working and got the variables deleting.  I found that the delete event was not being triggered, you can see that in the gif I uploaded.  So I just run it every time as a clean up step and it seems work.  I should be good as far git stuff as well.  Made a bit of mess though.

Also, I am both people, glaserpower and phptuts.  I confirm to the agreement.

Thanks for everything and let me know what you think

Noah

blockly-pr-3

RoboErikG:
Something seemed wrong about delete events not firing, so I dug around a bit more. They are firing, but if you delete the entire stack of inner blocks there's no longer a procedures_mutatorarg block to handle the event. You actually want to move your old event handling code to the procedures_mutatorcontainer and have that single block handle adding and removing variables for the mutator workspace.

Then it should work to  just handle the create and delete events (and bonus, there will only be one handler instead of one for each block).

phptuts:
@RoboErikG 
I think it might be ready to go.  Let me know what you think.  Sorry to bug you like this. 

Thank you
Noah

RoboErikG:
Sorry for the delay, we were all out for the holidays. I'll take a look today or tomorrow. =)

rachel-fenichel:
:fireworks:

phptuts:
Sweet!!! Thank you

PR #2182, opened 2018-12-18
Avoid crashing when a procedure input has an empty string

### LINKED ISSUE #1958: Clearing the name of a procedure parameter (even temporarily) breaks everything
### Problem statement
image

Trying to rename `x` to another name by deleting `x` and then starting to type causes major problems.

This is because there's an intermediate state where it tries to create a variable with the empty name.  Then calls to `allUsedVariables` and fail, as does `mutateCallers`.

If the variable name is empty we should not attempt to create a variable with that name.  (Do we just keep the old name around instead, and revert to it if they end editing there?)

### Stack Traces

```
procedures.js:133 Uncaught TypeError: Cannot read property 'name' of null
    at Blockly.BlockSvg.mutationToDom (procedures.js:133)
    at Object.Blockly.Procedures.mutateCallers (procedures.js:281)
    at Blockly.BlockSvg.compose (procedures.js:225)
    at Blockly.Mutator.workspaceChanged_ (mutator.js:339)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (workspace.js:436)
    at Blockly.Events.fireNow_ (events.js:160)```

RoboErikG:
Fixes #1958

This will drop the input when it's in the empty string state. This
causes the block to remove the input and re-add it when the text
is reverted, but prevents the breakage. A followup fix should leave
the input alone instead of removing it, but this works for now.

## The basics

- [x] I branched from develop

### Test Coverage

Tested on:
 Desktop Chrome

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2182)

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -159,7 +159,11 @@ Blockly.Blocks['procedures_defnoreturn'] = {
         this.arguments_.push(varName);
         var variable = Blockly.Variables.getOrCreateVariablePackage(
             this.workspace, varId, varName, '');
-        this.argumentVarModels_.push(variable);
+        if (variable != null) {
+          this.argumentVarModels_.push(variable);
+        } else {
+          console.log('Failed to create a variable with name ' + varName + ', ignoring.');
+        }
       }
     }
     this.updateParams_();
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
I assume that the pre-existing logic that keeps blocks attached to the correct inputs when renaming/moving inputs around also works here.  If so, lgtm.  I agree that there are better long-term solutions, but this is a good fix for now.

Do you want me to cherry-pick this into the release, or leave it for february?

RoboErikG:
Yep, confirmed the reconnection logic still works (as does undo/redo).

Probably worth merging in since it's a somewhat painful crash.

PR #2184, opened 2018-12-19
Blockly colours

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Adds a style class
A style class currently holds a list of different block styles but in the future could hold different style attributes. Block styles contain a color for the block, the border, and the shadow block. 
Adds the ability to update the workspace and toolbox when a user switches between styles.

### Reason for Changes
Currently, we only give developers the ability to change the primary color of a block. All other colors are computed from the primary block color. Giving the developer the ability to change the different colors of the block will allow them to tailor the block colors to their specific needs. (ex create more accessible color palettes)

### Test Coverage
Tests can be found in style_test.js and block_test.js

Tested on:
Desktop Chrome
Desktop Firefox
Desktop Safari


tests/generators/index.html
```diff
@@ -89,6 +91,7 @@
        zoom: {controls: true, wheel: true}
        });
   changeIndex();
+  Blockly.setStyle(Blockly.Styles.Classic);
```
RoboErikG: Wow...the non-debug test error was really not helpful.

blocks/lists.js
```diff
@@ -133,7 +133,7 @@ Blockly.Blocks['lists_create_with'] = {
    */
   init: function() {
     this.setHelpUrl(Blockly.Msg['LISTS_CREATE_WITH_HELPURL']);
-    this.setColour(Blockly.Msg['LISTS_HUE']);
+    this.setStyle("lists");
```
RoboErikG: nit: Strings in JS code should use single quotes here and below (but leave the ones in the JSON blobs as double quotes).

blocks/procedures.js
```diff
@@ -50,7 +50,7 @@ Blockly.Blocks['procedures_defnoreturn'] = {
         Blockly.Msg['PROCEDURES_DEFNORETURN_COMMENT']) {
       this.setCommentText(Blockly.Msg['PROCEDURES_DEFNORETURN_COMMENT']);
     }
-    this.setColour(Blockly.Msg['PROCEDURES_HUE']);
+    this.setStyle("procedures");
```
RoboErikG: ditto on single quotes in this file

blocks/text.js
```diff
@@ -221,7 +221,7 @@ Blockly.Blocks['text_getSubstring'] = {
       [Blockly.Msg['TEXT_GET_SUBSTRING_END_LAST'], 'LAST']
     ];
     this.setHelpUrl(Blockly.Msg['TEXT_GET_SUBSTRING_HELPURL']);
-    this.setColour(Blockly.Msg['TEXTS_HUE']);
+    this.setStyle("text");
```
RoboErikG: And here too.

core/block.js
```diff
@@ -859,6 +922,33 @@ Blockly.Block.prototype.setColour = function(colour) {
...
+  if (blockStyle) {
+    this.style_ = blockStyle;
+    this.secondaryColour_ = blockStyle.secondaryColour;
+    this.tertiaryColour_ = blockStyle.tertiaryColour;
+    // Set colour will trigger an updateColour() on a block_svg
+    this.setColour(blockStyle.primaryColour);
+  }
+  else {
+    var errorMsg = 'Invalid style name: ' + blockStyleName;
+    throw errorMsg;
```
RoboErikG: nit this should be:
throw Error('message');

And you don't need the variable, you can assemble in the Error

core/block.js
```diff
@@ -859,6 +922,33 @@ Blockly.Block.prototype.setColour = function(colour) {
...
+/**
+ * Set the style and colour values of a block.
+ * @param {string} blockStyleName Name of the block style
+ * @throws {Error} if the block style does not exist.
+ */
+Blockly.Block.prototype.setStyle = function(blockStyleName) {
+  var dereferenced = Blockly.utils.replaceMessageReferences(blockStyleName);
+  var style = Blockly.getStyle();
+  if (!style) {
+    throw 'Trying to set block style before Blockly.setStyle() has been called';
```
RoboErikG: nit this should be:
throw Error('message');

core/style.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Overrides or adds a style to the blockStyles map.
+ * @param{String} blockStyleName The name of the block style.
+ * @param{Blockly.BlockStyle} blockStyle The block style
+*/
+Blockly.Style.prototype.setBlockStyle = function(blockStyleName, blockStyle) {
+  var event = new Blockly.Events.Ui(null, 'blockStyleChanged',
+      this.blockStyles_[blockStyleName], blockStyle);
+  this.blockStyles_[blockStyleName] = blockStyle;
+  Blockly.Events.fire(event);
```
RoboErikG: I'm not actually sure this event is useful. You were right to update all the blocks when the style is set instead of having them listen to style changes (as that would cause every block to have an event listener). We might want to just clearly document that styles will only be updated when Blockly.setStyle() is called.

core/style.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
```
RoboErikG: Copyright headers on all new files.

style/highcontrast.js
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "variables_dynamic": {
+    "primaryColour": "#880e4f",
+    "secondaryColour":"#FF73BE",
+    "tertiaryColour":"#FFD4EB"
+  }
+};
+
+
+
```
RoboErikG: Add a comment that this style is still being fleshed out and may change.

style/modern.js
```diff
@@ -0,0 +1,55 @@
...
+    "secondaryColour": "#dbbdd6",
+    "tertiaryColour": "#84497a"
+  },
+  "variables_dynamic": {
+    "primaryColour": "#a55b99",
+    "secondaryColour": "#dbbdd6",
+    "tertiaryColour": "#84497a"
+  }
+};
+
```
RoboErikG: Add a comment that this style is still being fleshed out and may change.

style/highcontrast.js
```diff
@@ -0,0 +1,57 @@
...
+    "tertiaryColour":"#FFD4EB"
+  },
+  "variables_dynamic": {
+    "primaryColour": "#880e4f",
+    "secondaryColour":"#FF73BE",
+    "tertiaryColour":"#FFD4EB"
+  }
+};
+
+
```
RoboErikG: Nit: too many newlines

style/classic.js
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Classic style.
+ * Contains multi colored border to create shadow effect.
+ * @author aschmiedt@google.com (Abby Schmiedt)
```
AnmAtAnm: Since this will be public, you don't need to include your email if you don't want to. You don't even need to include any @author line, though full name or just GitHub id might be balance between privacy and credit. Technically, most of this will be in the Git logs (there is a privacy flag for that, in the preferences), but keeping it out of the file makes it harder for people to extract or find its way into other projects.

style/style_script/create_styles.py
```diff
@@ -0,0 +1,182 @@
+# !/usr/bin/python2.7
```
AnmAtAnm: Probably worth making this new script compatible with 3, if possible:
https://python-future.org/compatible_idioms.html

style/style_script/styles_example.json
```diff
@@ -0,0 +1,31 @@
+/**
```
AnmAtAnm: Technically JSON does not allow comments, and several JSON parsers will (correctly) fail the parse when encountered.
Being a sample, it probably doesn't need a copyright header.

tests/multi_playground.html
```diff
@@ -56,6 +57,7 @@
 
   startBlocklyInstance('HorizontalEndLTR', false, true, 'end', toolbox);
   startBlocklyInstance('HorizontalEndRTL', true, true, 'end', toolbox);
+  Blockly.setStyle(Blockly.Styles.Classic);
```
AnmAtAnm: If this isn't set, will the workspace and blocks still load?  Ideally, we want this to be backward compatible with apps that upgrade their Blockly without adding any new code.

alschmiedt: This was left over from how I was setting the style before so I just deleted it. Currently, i'm setting the style in the inject method. If the user doesn't pass in a style then it will default to the Classic style.

core/block.js
```diff
@@ -224,11 +231,35 @@ Blockly.Block.prototype.data = null;
 Blockly.Block.prototype.colour_ = '#000000';
```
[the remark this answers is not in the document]
alschmiedt: So I think that it has to be in block because this is where we get all the information from the json object. I think we lose the colour value if we do not store it on the block. However, not a 100% sure so will investigate further.

blocks/colour.js
```diff
@@ -55,8 +55,8 @@ Blockly.defineBlocksWithJsonArray([  // BEGIN JSON EXTRACT
       }
     ],
     "output": "Colour",
-    "colour": "%{BKY_COLOUR_HUE}",
     "helpUrl": "%{BKY_COLOUR_PICKER_HELPURL}",
+    "style": "colour",
```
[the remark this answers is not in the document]
alschmiedt: Yeah I like that better. Especially since we are also adding category styles the extra clarification is nice.

core/block.js
```diff
@@ -224,11 +238,35 @@ Blockly.Block.prototype.data = null;
...
+ * @private
+ */
+Blockly.Block.prototype.tertiaryColour_ = null;
+
+/**
+ * Style of the block.
+ * @type {?Object} Map from style attribute name (string) to style value (string)
+ * @private
+ */
+Blockly.Block.prototype.style_ = null;
```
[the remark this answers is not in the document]
RoboErikG: Don't forget to remove this property.

Issue #2237, opened 2019-01-28
Trash hotspot wrong when trash is full.


NeilFraser:
1. Open this page:
  https://beta-dot-blockly-games.appspot.com/music
2. Drag a block over the trash.  Explore the top/bottom/left/right boundaries of the hotspot.  They are correct.
3. Drop the block into the trash.
4. Open the trash flyout, and drag the block out.
5. Drag this block over the trash.  Explore the top/bottom/left/right boundaries of the hotspot.  They are incorrect.

Affects both RTL and LTR.

NeilFraser:
@BeksOmega I think this is yours?

rachel-fenichel:
I agree, the root behaviour is bad, and fixing it would be better than adding in more special cases to mitigate it.  So go for it.

NeilFraser:
Fix confirmed.  Thanks!

PR #2240, opened 2019-01-28
Fixed Trashcan Hotspot & Changed Positioning

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2237 also similar to #2035 

### Proposed Changes

The trashcan and zoom controls are now positioned in the corner opposite the toolbox. (This just changes LTR, Vertical, End & RTL, Vertical, End)
toolboxesvertical
toolboxeshorizontal

Also replaces a call to workspace.resize() with a call to flyout.position() which corrects the trashcan hotspot bug.

### Reason for Changes

1) Removes the "chasing" of zoom controls that would occure when resizing the LTR, Vertical, End & RTL, Vertical, End workspaces when they had simple toolboxes.
2) Allows us to replace the workspace.resize() call with a call to flyout.position(), because the reason resize was called was specifically to move the trashcan (bad behavior stop it!). Replacing this call fixes the hotspot problem.

Plus the multiplayground looks very nice now =D

### Test Coverage

1. Opened the multiplayground, and set the toolbox mode to simple.
2. Observed how all of the trashcans & zoom controls were positioned correctly (pass).
3. Pressed the + & - zoom controls. Observed that the flyouts were being resized, but the trashcan & zoom controls were not moving. (pass)
4. Dragged a block from the flyout into the trashcan. Observed how the hotspot was correct. (pass)
5. Clicked the trashcan. Observed how the flyout was created in the correct location, over the trashcan. (pass)
6. Dragged a block out of the trashcan flyout back into the trashcan. Observed how the hotspot was still correct (pass)
7. Repeated steps 3-6 for every workspace on the multiplayground.

Tested on:
* Desktop Chrome

### Additional Information

Possible follow-up PR:
The trashcan lid opens on the left in LTR mode and on the right in RTL mode. This now feels weird for the Vertical, End workspaces because they open in the opposite direction you generally drag blocks into them from.
trashcanliddirection
Is this something you peoples would like changed?

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -366,8 +366,6 @@ Blockly.HorizontalFlyout.prototype.reflowInternal_ = function() {
     }
     // Record the height for .getMetrics_ and .position.
     this.height_ = flyoutHeight;
-    // Call this since it is possible the trash and zoom buttons need
-    // to move. e.g. on a bottom positioned flyout when zoom is clicked.
-    this.targetWorkspace_.resize();
+    this.position();
   }
 };
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -392,8 +392,6 @@ Blockly.VerticalFlyout.prototype.reflowInternal_ = function() {
     }
     // Record the width for .getMetrics_ and .position.
     this.width_ = flyoutWidth;
-    // Call this since it is possible the trash and zoom buttons need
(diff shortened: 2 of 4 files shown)


rachel-fenichel:
LGTM.  @NeilFraser for opinions on the direction the lid opens.  I say do it if it's easy but don't spend too much time on it.

PR #2243, opened 2019-01-29
Ignore insertion markers in getAllBlocks; add accessors for insertion markers

### LINKED ISSUE #2236: workspace.getAllBlocks shouldn't return insertion markers by default
workspace.getAllBlocks is used by developers to query the contents of the workspace.  Blockly Games has 10 such calls, used for evaluating answers, providing help, and checking if the user has completed certain steps.  None of these calls is expecting the presence of insertion markers.

My suggestion is to move the check added in PR #2233 from remainingCapacity to getAllBlocks.  If needed, add an optional boolean parameter which, if true,  will also return insertion markers.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2236

### Proposed Changes

Stop counting insertion markers in the results of getAllBlocks.  Add an accessor for getting the current live insertion markers from a gesture.

It also would have worked to have an optional parameter to getAllBlocks, but I decided to leave its signature alone and do something a little more complicated but only in internal code.  This way we don't add yet another thing to the external API.

### Reason for Changes

See #2236 
### Test Coverage
Tested in the max blocks demo with uncompressed blockly.

### Additional Information

This might also apply to `workspace.getBlocksByType`, `getTopBlocks`, etc.  It's kind of unfortunate: there are good reasons for the blocks to be in the top blocks list, mostly related to behaving like perfectly normal blocks for connection and disconnection purposes, but in all other cases they're unwanted.

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -328,7 +328,14 @@ Blockly.Workspace.prototype.getAllBlocks = function(ordered) {
       blocks.push.apply(blocks, blocks[i].getChildren(false));
     }
   }
-  return blocks;
+
+  // Insertion markers exist on the workspace for rendering reasons, but aren't
+  // "real" blocks from a developer perspective.
+  var filtered = blocks.filter(function(block) {
+    return !block.isInsertionMarker();
+  });
+
+  return filtered;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


core/workspace.js
```diff
@@ -328,7 +328,14 @@ Blockly.Workspace.prototype.getAllBlocks = function(ordered) {
       blocks.push.apply(blocks, blocks[i].getChildren(false));
     }
   }
-  return blocks;
+
+  // Insertion markers exist on the workspace for rendering reasons, but aren't
+  // "real" blocks from a developer perspective.
+  var filtered = blocks.filter(function(block) {
```
AnmAtAnm: Should these be filtered at the `getDescendent()` and `getChildren()` level? I imagine we don't want insertion markers returned via those methods, either.

rachel-fenichel: Hm, maybe?  I can't tell where the right place for this code is.

Issue #2238, opened 2019-01-28
Image dropdowns of disabled blocks don't render properly.


NeilFraser:
1. Open https://beta-dot-blockly-games.appspot.com/music
2. Drag note block from toolbox directly to trash.
3. Drag note block from toolbox to workspace as an orphan so it gets disabled.
4. Drag disabled block to trash.
5. Look in the trash.

<img width="190" alt="screen shot 2019-01-28 at 11 54 40" src="https://user-images.githubusercontent.com/250480/51862677-f7205780-22f3-11e9-9f2c-966c4aa3ebf0.png">

Note that the dropdown fails to render properly.

NeilFraser:
@BeksOmega I think this is yours?

NeilFraser:
This issue is unrelated to the trash.  It's actually a bug in the rendering of image dropdowns.

1. Go to: tests/playground.html?toolbox=test-blocks
2. From the 'Fields' category, create a block with the image dropdown.
3. Right-click, disable block.
4. Right-click, duplicate block.

![screen shot 2019-01-29 at 10 59 38](https://user-images.githubusercontent.com/250480/51932974-25b23700-23b5-11e9-89d3-d87368341b02.png)

BeksOmega:
I can still grab it if you're not working on it already.

NeilFraser:
Go for it!  At this point my knowledge of the update path is obsolete enough that it's actively dangerous.  :)

AnmAtAnm:
@rachel-fenichel @BeksOmega 
Resolved by #2246?

PR #2246, opened 2019-01-29
Fixed Image Dropdowns on Disabled Blocks not Rendering Correctly.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2238 

### Proposed Changes

So the problem is that the first time FieldDropdown.renderSelectedImage_ is called the dropdown arrow is not yet rendered, which causes the FieldDropdown to calculate its size incorrectly. By happy accident, if the block is enabled the field is forced to rerender because of updateColor(), but when the block is disabled this does not happen, so it is stuck with the incorrect size.

The incorrect first size calculation is triggered as follows: (function names link to the function, arrows link to the call)
[xml.domToBlock](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/xml.js#L535) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/xml.js#L555)
[block.initSvg](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_svg.js#L147) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_svg.js#L152)
[Input.init](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/input.js#L230) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/input.js#L235)
[Field.init](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field.js#L179) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field.js#L208)
[FieldDropdown.render_](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field_dropdown.js#L431) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field_dropdown.js#L450)
[FieldDropdown.renderSelectedImage_](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field_dropdown.js#L463)

#### I believe this can be corrected by removing the call to FieldDropdown.render_ in Field.init.

If block.render() is always called after Field.init / Input.init we can safely remove the render_ call in Field.init because block.render() triggers the following:
[block.render](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_render_svg.js#L318) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_render_svg.js#L336)
[block.renderCompute_](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_render_svg.js#L404) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_render_svg.js#L480)
[field.getSize](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field.js#L456) [>](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field.js#L458)
[FieldDropdown.render_](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/field_dropdown.js#L431)

Field.init is called in three places (as far as I can tell): 1) [block_svg.setCollapsed()](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_svg.js#L511) 2)[ input.insertFieldAt()](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/input.js#L111) 3) [input.init()](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/input.js#L235).

After block_svg.setCollapsed() calls Field.init() it calls block.render() [here](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/block_svg.js#L525). So if we remove the FieldDropdown.render_() call it will not cause problems here.

After input.insertFieldAt() calls Field.init() it calls block.render() [here](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/input.js#L128). So if we remove the FieldDropdown.render_() call it will not cause problems here.

After input.init() calls Field.init it does not call block.render(), but luckily the only place input.init() gets called (again as far as I can tell) is from BlockSvg.initSvg(), and anytime BlockSvg.initSvg() is called block.render() is also called. So if we remove the FieldDropdown.render_() call it will not cause problems here either.

#### TL:DR (or another way of thinking about it)
It does not make sense for render_ to be called inside Field.init because the calculations aren't ready to be performed. If they were ready to be performed we wouldn't have the problem of the width being incorrectly calculated in the first place.

### Reason for Changes

Fixes dropdowns being rendered incorrectly, also removes some extra calculations we don't need to do.

First 5 Spaghetti tests include the render_ call. Average time: 2856 ms
Last 5 Spaghetti tests do not include the render_ call. Average time: 2521 ms.
spaghetti tests

### Test Coverage

1 . Opened up the playground.
2 . Clicked through all of the categories.
3 . Observed how everything was rendering correctly. (pass)

1 . Went to: tests/playground.html?toolbox=test-blocks
2 . From the 'Fields' category, created a block with the image dropdown.
3 . Right-clicked, disabled block.
4 . Right-clicked, duplicated block.
5 . Observed how the block was duplicated correctly. (pass)

Tested on:
 * Desktop Chrome 

### Additional Information

I apologize for writing war and peace on your github.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -204,8 +204,6 @@ Blockly.Field.prototype.init = function() {
   this.mouseDownWrapper_ =
       Blockly.bindEventWithChecks_(
           this.fieldGroup_, 'mousedown', this, this.onMouseDown_);
-  // Force a render.
-  this.render_();
 };
 
 /**
@@ -375,7 +373,7 @@ Blockly.Field.prototype.render_ = function() {
 };
 
 /**
- * Updates thw width of the field. This calls getCachedWidth which won't cache
+ * Updates the width of the field. This calls getCachedWidth which won't cache
  * the approximated width on IE/Edge when `getComputedTextLength` fails. Once
  * it eventually does succeed, the result will be cached.
  */


rachel-fenichel:
I need to set aside some time to dig through this code.  Thanks for all of the pointers and descriptions of what's going on.

That render is weird enough that I want to look at it more.  It looks like it was put there to fix a specific problem.

Related: do you know about the blame view on github?  You can use it to see when a particular line of a file was changed, by whom, and in what commit.  I use it when I'm trying to trace the history of a line like this.  

Here's the [blame](https://github.com/google/blockly/blame/develop/core/field.js#L208) for that line, and here's the relevant [commit](https://github.com/google/blockly/commit/01ab0b4ab6bbae495c09d6ed6f553f9eb9e81b9f).

Issue #2160, opened 2018-12-10
Programatically calling workspace setScale doesn't properly scale scrollbars


samelhusseini:
### Problem statement

Calling workspace setScale doesn't properly scale the workspace scrollbars. 
Currently the only instance of setScale called internally is the zoom reset button, but that calls scrollCenter which in turn sets the state of the workspace scrollbar. 

Calling setScale on the workspace without following up with updating the workspace's scrollbar state causes the following issue, see screenshot.

### Expected Behavior

Calling setScale sets the scale of the workspace and in turn update / invalidates the scrollbar state.

### Actual Behavior

![setscale](https://user-images.githubusercontent.com/16690124/49711811-574d2400-fbf6-11e8-976b-dd2f57a1e1ed.gif)

### Steps to Reproduce

1. Open the Blockly playground with the simple toolbox (Although I've checked this repro's for category toolboxes as well)
2. Open the browser console and type: 
```
Blockly.mainWorkspace.setScale(0.5)
```
3. Drag any Block from the flyout and notice how the Block's scale is off.
4. This also causes issues with the Blockly

rachel-fenichel:
Weird.  `setScale` includes a call to `this.scrollbar.resize`, that I would expect to fix this.  
And I can't repro this on the hosted playground, either in simple or categories, with your repro steps you gave.

samelhusseini:
Hmm. I have a 100% repro on local develop Blockly. You're right this doesn't repro on the hosted version, not sure why, maybe bits have changed since. 

Can you repro locally?

rachel-fenichel:
Aha.  I can repro if there are no blocks on the workspace.  If there are blocks on the workspace, it correctly adjusts the scale and it's all good.

It's short-circuiting the scrollbar update because the host metrics haven't changed, [here](https://github.com/google/blockly/blob/develop/core/scrollbar.js#L99).  If there's a block on the screen the host metrics *do* change.

Issue #989, opened 2017-03-22
Zoom-center is wrong with an always-open flyout


rachel-fenichel:
In the multi-playground, with the same block dragged into each playground and then one click on the "zoom to center" button:

![image](https://cloud.githubusercontent.com/assets/13686399/24220758/d18e49c4-0f08-11e7-8383-a0e23451d24c.png)

![image](https://cloud.githubusercontent.com/assets/13686399/24220738/b9cc3530-0f08-11e7-9f76-6eee7d4745b0.png)

rachel-fenichel:
With categories there is no problem:

![image](https://cloud.githubusercontent.com/assets/13686399/24220811/0e3d104e-0f09-11e7-8741-f28090b286a7.png)

![image](https://cloud.githubusercontent.com/assets/13686399/24220821/1d5c4752-0f09-11e7-99db-51574e4a2a3f.png)

rachel-fenichel:
This exists on master right now.

alschmiedt:
Closing as I can no longer repro. From what I can tell it was probably fixed in #2247.

Issue #944, opened 2017-02-22
zoomToFit is broken when workspace has no scrollbars and blocks have negative coordinates


picklesrus:
zoomToFit calculates the incorrect blocks bounding box size when there are no scrollbars and blocks with negative coordinates. 

If metrics.contentLeft and metrics.contentTop are negative (which happens when blocks are placed up and to the left of the initial 0, 0 of the workspace), the blocksWidth and blocksHeight calculations are incorrect and the blocks are above or left of the visible area. So, there are two issues at play.

1) https://github.com/google/blockly/blob/master/core/workspace_svg.js#L1414 adds padding to the bounding box to ensure blocks are centered after the zoom is applied, but this only works if blocks are at positive coordinates.
2) There's currently no way to get blocks at negative coordinates into the frame.

Note that this isn't a common problem for a user to run into because most workspaces that have blocks in the negative coordinate space also have scrollbars.  The only other way to get them there is to put them there programatically.  This was noticed while working with multiple workspaces on a minimap demo (https://github.com/google/blockly/issues/924).

In order to solve both these problems we will probably need to either move all the blocks to originate at 0,0 or translate the workspace so that the top and left most edges are inside the view.

alschmiedt:
Closing as this looks like it was fixed in #2247.

PR #2247, opened 2019-01-31
Movement Updates

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2205 

### Proposed Changes

Adds a new "move" object configuration option with the following properties:
* scrollbars: Determines if the workspace has scrollbars. Defaults to true if the workspace has categories.
* drag: Determines if the workspace can be dragged with the mouse. Always false if scrollbars is false (at least in options parsing). Defaults to true if scrollbars is true.
* wheel: Determines if the workspace can be scrolled with the mouse wheel. Always false if scrollbars is false (at least in options parsing). Defaults to false.

Parsing does not currently allow these options to be independent, but you can make them independent by overriding the drag & wheel options after the workspace has been injected. This behavior is supported.

Changes the zoom function so that the zoom is "anchored" to the mouse position, even without scrollbars.

### Reason for Changes

People have requested scroll on wheel. #110

Also I think it may be useful to start separating the movement functionality from scrollbars to work towards #2145

### Test Coverage

  
Independent-ness of Scrollbars/Drag/Scrollwheel
1. Set move > scrollbars option to true. Set move > drag option to false. Set move > wheel option to false.
2. Observed that scrollbars alone worked. (pass)
3. Set move > scrollbars option to false. Overwrote move > drag to be true. Set move > wheel option to false. 
4. Observed that the dragging alone worked. And that scrollbars were not created. (pass)
5. Set move > scrollbars option to false. Set move > drag option to false. Overwrote move > wheel to be true.
6. Placed a block so I could observe the scrolling.
7. Observed that the scrollwheel alone worked. And that scrollbars were not created. (pass)

Scrolling and Zooming Interaction
1. Set move > scrollbars option to true. Set move > wheel option to false. Set zoom > wheel option to true.
2. Scrolled the mouse wheel. Observed how the workspace zoomed. (pass)
3. Set move > scrollbars option to true. Set move > wheel option to true. Set zoom > wheel option to true.
4. Scrolled the mouse wheel. Observed how the workspace scrolled vertical. (pass)
5. Hit shift and scrolled the mouse wheel. Observed how the workspace scrolled horizontally. (pass)
6. Hit ctrl and scrolled the mouse wheel. Observed how the workspace zoomed. (pass)
7. Set move > scrollbars option to true. Set move > wheel option to true. Set zoom > wheel option to false.
8. Scrolled the mouse wheel. Observed how the workspace scrolled. (pass)
9. Hit shift and scrolled the mouse wheel. Observed how the workspace scrolled horizontally. (pass)

Zooming With and Without Scrollbars
1. Set scrollbars option to true. Set zoom > wheel option to true. Placed a block to visualize zooming.
2. Scrolled the mouse wheel. Observed how the zoom was anchored to the mouse. (pass)
3. Pressed the zoom in control. Observed how it zoomed towards the center. (pass)
4. Changed scrollbars option to false.
5. Scrolled the mouse wheel. Observed how the zoom was anchored to the mouse. (pass)
6. Pressed the zoom in control. Observed how it zoomed towards the center. (pass)
7. Changed the zoom > wheel option to false. 
8. Pressed the zoom in control. Observed how it zoomed towards the top left corner. (pass)

Note: Steps 7&8 of the above make it so the content is no longer bounded. If you were to, for example, override drag to make the content bounded it would zoom towards the center.

Tested on:
* Desktop Chrome

### Additional Information

All of this works, but there are still some things I want to do (like clean up calls to scrollbars.set) so I consider it a little bit WIP. I just wanted to post a PR before I went too crazy doing things, so I can still change stuff easily.

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -288,10 +288,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
       // We always check isMovable_ again because the original
       // "not movable" state of isMovable_ could have been changed.
       if (!mainWorkspace.isDragging() && !mainWorkspace.isMovable() &&
-          (e.type == Blockly.Events.BLOCK_CREATE ||
-          e.type == Blockly.Events.BLOCK_MOVE ||
-          e.type == Blockly.Events.COMMENT_CREATE ||
-          e.type == Blockly.Events.COMMENT_MOVE)) {
+          (Blockly.Events.BUMP_EVENTS.indexOf(e.type) != -1)) {
         var metrics = getWorkspaceMetrics();
         if (metrics.contentTop < metrics.viewTop ||
             metrics.contentBottom > metrics.viewBottom ||
@@ -353,41 +350,6 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
     };
     mainWorkspace.addChangeListener(bumpObjects);
   }
-
-  // A reminder for developers to add any events that have to do with
-  // objects on/in the workspace to the bumpObjects function. Useful when
-  // testing often happens on a bounded workspace.
-  var bumpObjectsEventChecker = function(e) {
-    switch (e.type) {
-      case Blockly.Events.BLOCK_CREATE:
-      case Blockly.Events.BLOCK_MOVE:
-      case Blockly.Events.COMMENT_CREATE:
-      case Blockly.Events.COMMENT_MOVE:
-        // bumpObjects activates on these events, they are a known factor.
-        // Nothing to worry about here.
-        break;
-      case Blockly.Events.BLOCK_CHANGE:
-      case Blockly.Events.BLOCK_DELETE:
-      case Blockly.Events.COMMENT_CHANGE:
-      case Blockly.Events.COMMENT_DELETE:
-      case Blockly.Events.VAR_CREATE:
-      case Blockly.Events.VAR_DELETE:
-      case Blockly.Events.VAR_RENAME:
-      case Blockly.Events.UI:
-      case Blockly.Events.FINISHED_LOADING:
-        // bumpObjects ignores these events, they are also a know factor.
-        // Nothing to worry about here.
-        break;
-      default:
-        // We found an event we don't know about. It may or may not need
-        // to activate bumpObjects, notify the developer.
-        // If your event has to do with objects on/in the workspace, add it to
-        // the bumpObjects function. Either way add a check to this function.
-        console.warn('Found an event that may need to be added to' +
-          ' bumpObjects. Please check before submitting a PR.');
-    }
-  };
-  mainWorkspace.addChangeListener(bumpObjectsEventChecker);
 
   // The SVG is now fully assembled.
   Blockly.svgResize(mainWorkspace);
--- a/core/inject.js
+++ b/core/inject.js
@@ -237,7 +237,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
       // If the event is of a type that does not affect block placement,
       // return. In this case it is better return-if rather than do-if so
       // that if events are added that should activate this function to work,
-      // people don't ned to remember to update it.
+      // people don't need to remember to update it.
       if (e.type == Blockly.Events.BLOCK_DELETE ||
           e.type == Blockly.Events.VAR_CREATE ||
           e.type == Blockly.Events.VAR_DELETE ||
@@ -326,7 +326,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
           }
           if (e) {
             if (!e.group && movedBlock) {
-              console.log('WARNING: Moved blocks in bounds but there was no event group.'
+              console.log('WARNING: Moved block in bounds but there was no event group.'
                         + ' This may break undo.');
             }
             Blockly.Events.setGroup(oldGroup);
--- a/core/inject.js
+++ b/core/inject.js
@@ -250,6 +250,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
       // "not movable" state of isMovable_ could have been changed.
       if (!mainWorkspace.isDragging() && !mainWorkspace.isMovable_()) {
         var metrics = mainWorkspace.getMetrics();
+
         // Make sure we have exact dimensions even if the workspace is bounded.
         if (mainWorkspace.isContentBounded_()) {
           var blocksBoundingBox = mainWorkspace.getBlocksBoundingBox();
@@ -260,28 +261,33 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
           metrics.contentWidth = blocksBoundingBox.width * scale;
           metrics.contentHeight = blocksBoundingBox.height * scale;
         }
-        var edgeLeft = metrics.viewLeft + metrics.absoluteLeft;
-        var edgeTop = metrics.viewTop + metrics.absoluteTop;
-        if (metrics.contentTop < edgeTop ||
-            metrics.contentTop + metrics.contentHeight >
-            metrics.viewHeight + edgeTop ||
-            metrics.contentLeft <
-                (options.RTL ? metrics.viewLeft : edgeLeft) ||
-            metrics.contentLeft + metrics.contentWidth > (options.RTL ?
-                metrics.viewWidth : metrics.viewWidth + edgeLeft)) {
-          // The event block is out of bounds, bump it back in.
-          var MARGIN = 25;
+
+        // Check if the block is out of bounds.
+        var contentTop = metrics.contentTop;
+        var contentBottom = metrics.contentTop + metrics.contentHeight;
+        var contentLeft = metrics.contentLeft;
+        var contentRight = metrics.contentLeft + metrics.contentWidth;
+        var viewTop = metrics.viewTop;
+        var viewBottom = metrics.viewHeight + metrics.viewTop;
+        var viewLeft = metrics.viewLeft;
+        var viewRight = options.RTL ? metrics.viewWidth :
+            metrics.viewWidth + metrics.viewLeft;
+        if (contentTop < viewTop || contentBottom > viewBottom ||
+            contentLeft < viewLeft || contentRight > viewRight) {
+
+          // Handle undo.
           var oldGroup = null;
           if (e) {
             oldGroup = Blockly.Events.getGroup();
             Blockly.Events.setGroup(e.group);
           }
 
+          var MARGIN = 25;
           var movedBlock = false;
           var block = mainWorkspace.getBlockById(e.blockId);
 
-          // Get block info in pixel coordinates, since all of the metrics
-          // info is in pixel coordinates.
+          // Get block info in pixel units, since all of the metrics
+          // info is in pixel units.
           var blockXY = block.getRelativeToSurfaceXY();
           blockXY.x *= mainWorkspace.scale;
           blockXY.y *= mainWorkspace.scale;
@@ -290,24 +296,27 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
           blockHW.height *= mainWorkspace.scale;
 
           // Bump any block that's above the top back inside.
-          var overflowTop = edgeTop + MARGIN - blockHW.height - blockXY.y;
+          var overflowTop =
+              metrics.viewTop + MARGIN - blockHW.height - blockXY.y;
           if (overflowTop > 0) {
             // Convert to workspace units.
             overflowTop /= mainWorkspace.scale;
             block.moveBy(0, overflowTop);
             movedBlock = true;
           }
+
           // Bump any block that's below the bottom back inside.
           var overflowBottom =
-              edgeTop + metrics.viewHeight - MARGIN - blockXY.y;
+            metrics.viewTop + metrics.viewHeight - MARGIN - blockXY.y;
           if (overflowBottom < 0) {
             // Convert to workspace units.
             overflowBottom /= mainWorkspace.scale;
             block.moveBy(0, overflowBottom);
             movedBlock = true;
           }
+
           // Bump any block that's off the left back inside.
-          var overflowLeft = MARGIN + edgeLeft -
+          var overflowLeft = MARGIN + metrics.viewLeft -
               blockXY.x - (options.RTL ? 0 : blockHW.width);
           if (overflowLeft > 0) {
             // Convert to workspace units.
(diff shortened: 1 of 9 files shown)


core/inject.js
```diff
@@ -232,19 +232,19 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
-        var edgeLeft = metrics.viewLeft + metrics.absoluteLeft;
-        var edgeTop = metrics.viewTop + metrics.absoluteTop;
-        if (metrics.contentTop < edgeTop ||
+        if (metrics.contentTop < metrics.viewTop ||
             metrics.contentTop + metrics.contentHeight >
-            metrics.viewHeight + edgeTop ||
+            metrics.viewHeight + metrics.viewTop ||
             metrics.contentLeft <
-                (options.RTL ? metrics.viewLeft : edgeLeft) ||
+                (options.RTL ? metrics.viewLeft : metrics.viewLeft) ||
```
jogibear9988: why don't remove this condition?

BeksOmega: Yep, I need to take another look at this whole function. Coordinates man :/

Thanks for pointing it out to me!

BeksOmega: Turns out that I thought the "+ metrics.absoluteLeft" and "+ metrics.absoluteTop" were causing a problem, but the problem was actually that the scrollX and scrollY were not getting set after setMetrics was called, so viewTop and viewLeft were getting calculated incorrectly. This has been fixed.

Everything (except the mainworkspace.isContentBounded_() checks) has been reverted, and the workspaceChanged function is working once again.

rachel-fenichel:
Thanks for sending this in for review before diving deeper.  The scrollbar code has some important performance constraints, and has been optimized in some non-obvious ways to improve performance.  As you play with it you'll want to make sure your changes don't significantly change performance.

The most important thing is that `getMetrics()` isn't free, and we want to use cached values whenever possible.
GetMetrics also gets called in places you wouldn't expect, so it's worth digging through code to make sure they aren't being called unnecessarily.

The screenshot below is from the [performance](https://developers.google.com/web/tools/chrome-devtools/evaluate-performance/reference) tab in Chrome developer tools.  These tools are really helpful for keeping an eye on performance.  For this test I 
- loaded the playground
- hit airstrike once
- used the mouse wheel to scroll in

Here's what that looks like on master, for reference:
image

So far what you've got looks reasonable.  Just make sure you keep this stuff in mind.

rachel-fenichel:
Tag me on this when you're ready to have it reviewed again and I'll take another look.

BeksOmega:
Ok @rachel-fenichel I think it's ready for another look!

Pre Script : You know how much I value my PRs being easy to review, but I fell below my own standards on this one. I'll try my best to stick to the 1 PR 1 issue principle in the future, but I really wanted to leave the code in a good place, and that required more code than I first assumed.

### Updates!

The bulk of the changes went into making scrollbar-less zoomable workspaces work better. For example...

Old Zooming With Buttons:
oldzoom_buttons
New Zooming With Buttons:
newzoom_buttons

Old Zooming With Wheel:
oldzoom_wheel
New Zooming With Wheel:
newzoom_wheel

Blocks are now completely bumped into the workspace.
Old Block Bumping:
oldbumping_blocks
New Block Bumping:
newbumping_blocks

Workspace comments are also bumped now.
Old Comment Bumping:
oldbumping_comments
New Comment Bumping:
newbumping_comments

The scroll center zoom button was removed for un-movable workspaces (as you can see in the two "new" gifs above), because it could cause you to lose blocks at the edges of the workspace. 

The highlight definition button was also removed for un-movable workspaces, because it could cause you to lose blocks at the edges of the workspace.

All the movement functions now work correctly as well:

Fixed Issue #989 (incorrect scrollCenter() with flyouts)
newscrollcenter_1
newscrollcenter_2

Fixed zoomToFit with flyouts.
Old zoomToFit:
oldzoomtofit_1
oldzoomtofit_2
New zoomToFit:
newzoomtofit_1
newzoomtofit_2

Also fixed issue #944 (zoom to fit incorrect with negative block coords)
newzoomtofit_negativecoords

Post Script: I'm going to have to do 2000lbs of testing on this before I'm confident that all the edge cases are solid, but I wanted to give you another chance to look at it, and tell me what I should separate out if need be.

Post Post Script: Speaking of testing, if you have any ideas about how I could add unit tests for this I would be very interested in doing so.

Thanks for taking another look!
Beka

core/inject.js
```diff
@@ -232,71 +232,162 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
       }
     };
-    mainWorkspace.addChangeListener(workspaceChanged);
+    mainWorkspace.addChangeListener(bumpObjects);
   }
+
+  // A reminder for developers to add any events that have to do with
+  // objects on/in the workspace to the bumpObjects function. Useful when
+  // testing often happens on a bounded workspace.
+  var bumpObjectsEventChecker = function(e) {
```
BeksOmega: Do you know of any way to make closure not add this code to the minified files? Or to have it not run in compiled mode?

I added it because people will definitely forget to update this function, especially when most of the testing happens on a moveable workspace. But developers using blockly don't need to see this message.

BeksOmega:
This also fixes #2160:
programmaticresize

There are probably a couple of things that have been fixed because of the reorganization.

rachel-fenichel:
Thanks for the PR.  I'm a bit backlogged on reviews because I've been getting a release ready, but I'll give this a full review tomorrow.

core/workspace_svg.js
```diff
@@ -1516,36 +1628,39 @@ Blockly.WorkspaceSvg.prototype.setBrowserFocus = function() {
...
+  // Clamp scale within valid range.
   if (newScale > this.options.zoomOptions.maxScale) {
     scaleChange = this.options.zoomOptions.maxScale / this.scale;
   } else if (newScale < this.options.zoomOptions.minScale) {
     scaleChange = this.options.zoomOptions.minScale / this.scale;
   }
-  if (this.scale == newScale) {
-    return;  // No change in zoom.
-  }
-  if (this.scrollbar) {
```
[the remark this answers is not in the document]
BeksOmega: This check encapsulated centering the zoom on the mouse pointer, which was only possible before if there were scrollbars. Now we can center on the mouse pointer no matter what, so the check is unnecessary.

core/inject.js
```diff
@@ -232,71 +232,163 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
+        right: position.x + (options.RTL ? 0 : size.width),
+        top: position.y,
+        bottom: position.y + size.height,
+      };
+    };
+
+    var bumpObjects = function(e) {
+      // We always check isMovable_ again because the original
+      // "not movable" state of isMovable_ could have been changed.
+      if (!mainWorkspace.isDragging() && !mainWorkspace.isMovable_() &&
```
rachel-fenichel: See my comment on your bump object event checker, below.

core/inject.js
rachel-fenichel: I don't think we need this function.  Adding events isn't a super common thing yet, and this is the sort of information that we should include in the documentation on how to create a new event type instead of having it be a runtime check.

Rather than checking directly against the event types in your change handler, you can make a list in Blockly.Events of event types that should be bumped, and check the event type against that.  That way you don't have to enumerate all of the event types all over the place, and that code will be easier to read.

You can put a comment on that list saying that anyone adding a new event type that will cause bumping should add it to the list.

core/inject.js
```diff
@@ -232,71 +232,163 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
-                (options.RTL ? metrics.viewLeft : edgeLeft) ||
-            metrics.contentLeft + metrics.contentWidth > (options.RTL ?
-                metrics.viewWidth : metrics.viewWidth + edgeLeft)) {
-          // One or more blocks may be out of bounds.  Bump them back in.
-          var MARGIN = 25;
-          var blocks = mainWorkspace.getTopBlocks(false);
+  if (!options.readOnly && !mainWorkspace.isMovable_()) {
+    // Helper functions for the workspaceChanged callback.
+    var getWorkspaceMetrics = function() {
+      var workspaceMetrics = Object.create(null);
```
rachel-fenichel: This looks like it's doing a lot of work and checks that normally happen in `getMetrics`.  Why is all the new math here?

core/workspace_svg.js
```diff
@@ -1702,10 +1853,11 @@ Blockly.WorkspaceSvg.prototype.setScale = function(newScale) {
   if (this.grid_) {
     this.grid_.update(this.scale);
   }
+  this.scroll_(this.scrollX, this.scrollY);
```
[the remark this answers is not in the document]
BeksOmega: We call scroll_ instead of scrollbar.resize() so that we can center the zoom correctly without scrollbars, but scroll_ does not resize the scrollbars so we have to call resizeContent as well.

Also added the above as a comment.

core/inject.js
```diff
@@ -232,71 +232,163 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
+        workspaceMetrics.contentTop = defaultMetrics.contentTop / scale;
+        workspaceMetrics.contentRight =
+            (defaultMetrics.contentLeft + defaultMetrics.contentWidth) / scale;
+        workspaceMetrics.contentBottom =
+            (defaultMetrics.contentTop + defaultMetrics.contentHeight) / scale;
+      }
+
+      return workspaceMetrics;
+    };
+    var getWorkspaceObjectMetrics = function(object) {
```
rachel-fenichel: This is a useful function to define directly on blocks and comments, instead of as a helper function that lives inside the inject code.  I'm pretty sure that we're doing exactly this in other places in the code as well, and we can transition those to use the new function as we find them.  It'll be minor code duplication, but it'll make this function easier to find and use (and annotate so it shows up in our documentation).

core/inject.js
BeksOmega: The bumpObjects function needs a pretty uncommon set of metrics to work properly:

A) It needs everything in workspace units rather than pixel units.
B) It needs the un-bounded size of the content, even if the content is bounded.

Because the needs were so specific to the bumpObjects function, I wrote everything inside inject, but you're right it is more of a workspace thing. (Also I couldn't come up with a good name, getMetricsUnboundedWorkspaceUnits is a bit longwinded)

I can:
A) Leave as is.
B) Add a new function to the workspace_svg, but I would appreciate a name suggestion.
C) Change the signature of getTopLevelWorkspaceMetrics_() to (bool opt_workspaceUnits, bool opt_forceUnbounded), although this might be a bit confusing because I believe the signature of getMetrics would change for flyouts vs workspaces.

core/inject.js
BeksOmega: Turns out it was already a function: getBoundingRectangle ([block](https://developers.google.com/blockly/reference/js/Blockly.BlockSvg#getBoundingRectangle), [comment](https://developers.google.com/blockly/reference/js/Blockly.WorkspaceCommentSvg#getBoundingRectangle)) the more you know ¯\\\_(ツ)_/¯

core/inject.js
```diff
@@ -232,71 +232,163 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
-            }
-            // Bump any block that's off the right back inside.
-            var overflowRight = edgeLeft + metrics.viewWidth - MARGIN -
-                blockXY.x + (options.RTL ? blockHW.width : 0);
-            if (overflowRight < 0) {
-              block.moveBy(overflowRight, 0);
-              movedBlocks = true;
-            }
+
+          switch (e.type) {
```
BeksOmega: Should a getObjectById() function be added to the workspace so that we can remove this switch statement? Or do you think that's unnecessary?

core/workspace_svg.js
```diff
@@ -1714,6 +1866,57 @@ Blockly.WorkspaceSvg.prototype.setScale = function(newScale) {
...
+  y = Math.max(y, -maxOffsetOfViewFromOriginY);
+
+  // Hide the WidgetDiv without animation. This is to prevent a disposal
+  // animation from happening in the wrong location.
+  Blockly.WidgetDiv.hide(true);
+
+  this.scrollX = x;
+  this.scrollY = y;
+  if (this.scrollbar) {
+    // The negative x/y values are our new viewLeft and viewTop values.
```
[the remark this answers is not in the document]
BeksOmega: About 2, I changed it over scrollbar.set() but that causes two minor problems:

1. set() calls onScroll_ which calls setMetrics etc, adding unnecessary operations.
2. set() causes a little bit of jitter after zoom: (watch the lower right corner to observe how the scrollbars aren't placed correctly, then observe the upper left corner to observe the jitter)
scrollbarsset

The code is currently calling set(), but I can change it back if those problems are deal breakers.

BeksOmega:
Ok I have done 2000lbs of testing and it looks solid.

Here's everything I tested:

- Zooming with Buttons
  - Movable Workspaces
  - NonMovable Workspaces
- Zooming with Wheel
  - With Scrollbars
  - Without Scrollbars
- Scrolling with Wheel
  - With Scrollbars
  - Without Scrollbars
  - After Zooming in with Buttons
  - After Zooming out with Buttons
  - After Zooming in with Wheel
  - After Zooming out with Wheel
- Dragging the Workspace
  - With Scrollbars
  - Without Scrollbars
  - After Zooming in with Buttons
  - After Zooming out with Buttons
  - After Zooming in with Wheel
  - After Zooming out with Wheel
- Scrolling with Scrollbars
  - After Zooming in with Buttons
  - After Zooming out with Buttons
  - After Zooming in with Wheel
  - After Zooming out with Wheel
- Scroll Center
  - With Scrollbars
  - Without Scrollbars
  - After Zooming in
  - After Zooming out
- Zoom to Fit
  - With Scrollbars
  - Without Scrollbars
- Center on Block
  - With Scrollbars
  - Without Scrollbars
  - After Zooming in
  - After Zooming out
- Block Bumping
  - After Zooming in
  - After Zooming out
- Comment Bumping
  - After Zooming in
  - After Zooming out

And for every one of those I tested every toolbox position w/ categories and w/ simple toolboxes.

I just need replies to [this](https://github.com/google/blockly/pull/2247#discussion_r257051620), [this](https://github.com/google/blockly/pull/2247#discussion_r256634318), and [this](https://github.com/google/blockly/pull/2247#discussion_r256636907), then it should be good to go!

core/workspace_svg.js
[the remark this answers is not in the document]
rachel-fenichel: The setMetrics part wouldn't be a dealbreaker, only becuase I intend to make another pass to sort out metrics stuff.  But the jitter is a problem.  Go ahead and put it back to directly accessing the internals, but add a TODO and a github issue to try to sort those two out and not use the private properties in the future.

core/inject.js
rachel-fenichel: I think that's unnecessary--blocks and comments share a few APIs, but not many, so you'd still usually have to check the type of the object that you got back.

core/inject.js
rachel-fenichel: Leave as-is for this change, and add a TODO + github isue to circle back and see if we can move most of it out of inject.

rachel-fenichel:
Looks good!  I replied to the outstanding questions; after you handle those I can merge it.

For a few of these my answer is "we should fix it, but not in this PR."  We have until the next release to make sure everything is 100% correct, and those follow-ups are more about code structure than functionality, and can definitely follow in separate PRs.

PR #2254, opened 2019-02-04
Speed up the connection DB

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Significantly speeds up adding and removing connections from the connection DB.  This is especially important when dragging into or out of a large stack, because it's possible that all connections in the stack will need to be moved.  This also makes insertion markers faster.

Dragging a block out of the spaghetti stack is now about a second faster (!)

### Proposed Changes

Change the connection DB.  Instead of attempting to subclass Array, we made it contain an actual array.  

### Reason for Changes

This means that array operations like `splice` are optimized by the browser/hopefully happening in C, and we're getting massive speed-up when the connection DB is large (> 1000).

The connection DB was always supposed to be an Array, and the comments will tell you that it was one. `db instanceof Array` would return true, but `Array.isArray(db)` would return false.

More details on subclassing Array [here](http://perfectionkills.com/how-ecmascript-5-still-does-not-allow-to-subclass-an-array/) for the curious.

### Test Coverage

Tested in the playground on Chrome, with both spaghetti and normal dragging.  I also updated JSunit tests, which are looking at the db internals.

### DIFF
--- a/core/connection_db.js
+++ b/core/connection_db.js
@@ -36,14 +36,13 @@ goog.require('Blockly.Connection');
  * @constructor
  */
 Blockly.ConnectionDB = function() {
+  /**
+   * Array of connections sorted by y coordinate.
+   * @type {!Array.<!Blockly.Connection>}
+   * @private
+   */
+  this.connections_ = [];
 };
-
-Blockly.ConnectionDB.prototype = new Array();
-/**
- * Don't inherit the constructor from Array.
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
(Shout-out to @BeksOmega because I found this while getting screenshots explaining how I use chrome developer tools to keep an eye on performance)

Issue #1445, opened 2017-11-14
During horizontal window resize in RTL, scrollbar floats out of flyout bounds 


picklesrus:
### Problem statement

Flyout scrollbar should be attached to flyout edge during window resize, but it isn't. RTL only, toolbox at end. See animated gif attached.
- [x] Bug report
- [ ] Feature request

### Steps to Reproduce
1. Open simple flyout in rtl. (e.g. https://blockly-demo.appspot.com/static/tests/playground.html?dir=rtl&toolbox=simple&side=start)
2. shrink the window vertically so that the flyout has a scrollbar
3. shrink the window horizontally.
4. Notice the scrollbar doesn't stay pinned to the edge of the flyout.

![2b17dc51-e51a-4f0d-9c49-592a5f99f921](https://user-images.githubusercontent.com/15675877/32806995-573c5666-c943-11e7-8a10-832b88460ed4.gif)

### Operating System and Browser

_Remove this hint: these checkboxes can be checked like this: [x]_

- [x] Desktop:
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Opera
  - [ ] IE 10+
  - [ ] IE 11
  - [ ] EDGE

- [ ] Smartphone/Tablet/Chromebook (please complete the following information):

BeksOmega:
I think this is related to right-anchored flyouts rather than RTL flyouts. (Observe RTL vertical start & LTR vertical end)
![scrollbarwindowresize](https://user-images.githubusercontent.com/25440652/52448868-200dcd00-2aea-11e9-8bb9-91dcd75f7a37.gif)

BeksOmega:
This behavior also occures with bottom horizontal toolboxes.

Bottom LTR:
![scrollbarmovement_horizontal_playgrond_bottom_ltr](https://user-images.githubusercontent.com/25440652/52523545-84926e80-2c47-11e9-9601-da6f2fa63125.gif)

Bottom RTL:
![scrollbarmovement_horizontal_playgrond_bottom_rtl](https://user-images.githubusercontent.com/25440652/52523547-8c521300-2c47-11e9-8a15-968975f36698.gif)

It does not occure with top toolboxes:
![scrollbarmovement_horizontal_playgrond_top_ltr](https://user-images.githubusercontent.com/25440652/52523595-983dd500-2c47-11e9-8472-8529e09dee56.gif)

It is also not replicatable on the multiplayground (I assume because the workspaces don't actually resized vertically).
[...]

PR #2281, opened 2019-02-09
Fixed Flyout Scrollbars Not Being Repositioned on Window Resize

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#1445

### Proposed Changes

Adds an extra call to setPosition.

### Reason for Changes

Resize would fail because the metrics were the same. This means setPosition would never get called so the scrollbar would stay in the same position.

### Test Coverage

setposition_multiplayground
setposition_playground_ltr
setposition_playground_rtl

Tested on:
* Desktop Chrome

### Additional Information

N/A


rachel-fenichel:
LGTM, though we really need a better way of figuring out when metrics updates actually matter, and when scrollbars need to update.

Issue #2066, opened 2018-09-25
Clicking on document should call Blockly.hideChaff


ronniesir:
when i finish input text,i want to make the workspace get focus

NeilFraser:
Help me understand what you mean by focus (it's an overloaded term).  If a workspace had the focus, what would you be able to do?

ronniesir:
![image](https://user-images.githubusercontent.com/3816871/46051792-a645fa00-c16d-11e8-9f26-f8e6fd81495d.png)
![image](https://user-images.githubusercontent.com/3816871/46051855-076dcd80-c16e-11e8-8d57-770d9c69bfb8.png)

I input text '123' then switch another page.the text '123' still show. when the input lose focus the text show nomal.so i want to make the workspace get focus

NeilFraser:
I think I understand.  Allow me to rephrase the issue.

Problem:
WidgetDiv doesn't get unfocused when page elements outside Blockly are clicked.

Steps to recreate:
Click on a text or numeric input.  Field becomes editable as expected.
Click on the page body (not Blockly).  Field does not lose focus.
Click on workspace (Blockly).  Field does lose focus.

Sounds like there needs to be an event handler on the body to hide chaff.  I'm pretty sure this was the case a long time ago, and that this is a regression.  But if that's the case, it was so long ago that our old reference copy of Blockly Games doesn't show it.

ronniesir:
Your description is very exact.
I want to know if this problem will be fixed.

alschmiedt:
There were multiple attempts to fix this issue: #2359 and #2286. However, both were reverted because they caused other issues: #2312 and #2362 . We have decided that adding a 'mouseup' event listener at the document level has the potential to cause too many side effects. The current way around this is for the user to add an event listener on their own div that will call Blockly.hideChaff().

rachel-fenichel:
Closing this issue: developers should call Blockly.hideChaff() when the user starts doing things outside of the area that Blockly manages.

BeksOmega:
Do you think this could be fixable now since Blockly no longer depends on Closure for its context menus & other UI components? I'm wondering if it might be fixable now since you should have total control of the event handling. @rachel-fenichel @samelhusseini

Issue #2267, opened 2019-02-06
Block input box sticks on the screen when scrolled/navigated away while focused.


ljitendra:
### Problem statement

Bug: Block input box sticks on the screen when scrolled/navigated away while focused.

### Expected Behavior

When user scrolls the screen or navigates away from the page containing the blockly editor, input html element should be deleted from DOM

### Actual Behavior

Input box sticks to the screen.

### Steps to Reproduce

1. Include blockly editor as part of a form page such that you get scrollbar on the page.
2. Insert a block with input box.
3. Focus on input box to enter some text
4. While focused, scroll the page or navigate away from that page using link.

### Operating System and Browser
Test on latest chrome and firefox browser

### Additional Information

The reason the input box sticks to the page because the div is attached to the body adjacent to the application div.
Including a screenshot to better understand the problem.
Potential solution to this would be clear `<div class="blocklyWidgetDiv" ... >` innerHTML on `blur` and `scroll` event to remove the input box from DOM.

ljitendra:
![image](https://user-images.githubusercontent.com/15458935/52366902-d1443280-2a18-11e9-9e13-1d22e77177ed.png)

rachel-fenichel:
It sounds like there's a missing call to `hideChaff`.

samelhusseini:
Fixed by #2286

PR #2286, opened 2019-02-11
Fixes Block Input Sticking to the Screen When Navigated Away

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2066 and half of #2267 (I wasn't able to reproduce the sticking-on-scroll behavior)

### Proposed Changes

Calls hideChaff when the document is clicked. Also removes the hideChaff call from the toolbox, since that is now handled by the document listener.

### Reason for Changes

Bug fixes are good.

### Test Coverage

1 . Created a block with a text/number input field.
2 . Selected said field.
3 . Clicked the surface of the document. Observed how the field lost focus. (pass)
4 . Selected the field once again.
5 . Clicked on a category. Observed how the field lost focus. (pass)
6 . Selected the field once again.
7 . Click on a non-category part of the toolbox. Observed how the field lost focus (pass)
8 . Selected the field once again..
9 . Clicked the workspace. Observed how the field lost focus. (pass)

1 . Switched the toolbox view to simple.
2 . Created a block with a text/number input field.
3 . Selected said field.
4 . Clicked the flyout toolbox. Observed how the field lost focus. (pass)

Tested on:
* Desktop Chrome
* Desktop Firefox

* Windows Edge

### Additional Information

N/A

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -437,7 +437,10 @@ Blockly.init_ = function(mainWorkspace) {
  */
 Blockly.inject.bindDocumentEvents_ = function() {
   if (!Blockly.documentEventsBound_) {
-    Blockly.bindEventWithChecks_(document, 'mousedown', null, Blockly.hideChaff);
+    Blockly.bindEventWithChecks_(document, 'mousedown', null, function() {
+      Blockly.hideChaff();
+      Blockly.Touch.clearTouchIdentifier();  // Don't block future drags.
+    });
     Blockly.bindEventWithChecks_(document, 'scroll', null,
         Blockly.mainWorkspace.updateInverseScreenCTM
             .bind(Blockly.mainWorkspace));
--- a/core/inject.js
+++ b/core/inject.js
@@ -437,6 +437,7 @@ Blockly.init_ = function(mainWorkspace) {
  */
 Blockly.inject.bindDocumentEvents_ = function() {
   if (!Blockly.documentEventsBound_) {
+    Blockly.bindEventWithChecks_(document, 'mousedown', null, Blockly.hideChaff);
     Blockly.bindEventWithChecks_(document, 'scroll', null,
         Blockly.mainWorkspace.updateInverseScreenCTM
             .bind(Blockly.mainWorkspace));
--- a/core/toolbox.js
+++ b/core/toolbox.js
@@ -160,18 +160,6 @@ Blockly.Toolbox.prototype.init = function() {
   this.HtmlDiv.setAttribute('dir', workspace.RTL ? 'RTL' : 'LTR');
   svg.parentNode.insertBefore(this.HtmlDiv, svg);
 
-  // Clicking on toolbox closes popups.
-  Blockly.bindEventWithChecks_(this.HtmlDiv, 'mousedown', this,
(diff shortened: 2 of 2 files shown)


core/inject.js
```diff
@@ -380,6 +380,7 @@ Blockly.init_ = function(mainWorkspace) {
  */
 Blockly.inject.bindDocumentEvents_ = function() {
   if (!Blockly.documentEventsBound_) {
+    Blockly.bindEventWithChecks_(document, 'mousedown', null, Blockly.hideChaff);
```
AnmAtAnm: My intuition is this needs to clear the touch identifier also.

core/toolbox.js
```diff
@@ -160,18 +160,6 @@ Blockly.Toolbox.prototype.init = function() {
...
-  // Clicking on toolbox closes popups.
-  Blockly.bindEventWithChecks_(this.HtmlDiv, 'mousedown', this,
-      function(e) {
-        if (Blockly.utils.isRightButton(e) || e.target == this.HtmlDiv) {
-          // Close flyout.
-          Blockly.hideChaff(false);
-        } else {
-          // Just close popups.
-          Blockly.hideChaff(true);
-        }
```
AnmAtAnm: This looks like a change in behavior. By removing the if statement, the flyout will always close, even where it didn't before.

BeksOmega: That's what I thought too, but I think the first condition:
`if (Blockly.utils.isRightButton(e) || e.target == this.HtmlDiv)`
actually always returns true because e.target will always be this.HtmlDiv.

I'm honestly not exactly sure what this function was doing before. I tried clicking all over the categories but I couldn't figure it out.

AnmAtAnm:
Also, probably worth pointing out the [iframe test page](https://blockly-demo.appspot.com/static/tests/playgrounds/iframe.html).

BeksOmega:
> Also, probably worth pointing out the [iframe test page](https://blockly-demo.appspot.com/static/tests/playgrounds/iframe.html). Explore how this reacts when the drag hits the edge of the inner frame document.

If you click on the inner frame it works as expected. If you click on the outer frame nothing happens. If you drag the mouse from the outer frame to the inner frame, nothing happens.

I can do more testing, I just need a little bit more detail about what I'm looking for.

core/toolbox.js
rachel-fenichel: I'd like to keep the old behaviour.  I think you can add your new listener and also keep this one without any issues.

BeksOmega: > Here are the MDN docs for [Event.target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) and [Event.currentTarget](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget). And if you're curious, here's some information on [event bubbling and capture](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture), which is part of why the distinction matters.

Thanks for the links! That was a really interesting read :D

alschmiedt:
Going to revert this change since it is causing problem documented here #2312  but can revisit later

PR #2304, opened 2019-02-20
Disable orphaned mutator blocks.

NeilFraser:
Disables any block in a mutator that isn't connected to the root block.

Two-line change.  Not sure why this wasn't done years ago.


NeilFraser:
BTW, as a side effect, this PR adds our first call to `Blockly.Events.disableOrphans` to the core.  Previously this widely-used function was untestable in playground/demos.

Issue #2280, opened 2019-02-09
Q: Should the absoluteLeft and absoluteTop Metrics be Shifted to Avoid Simple Toolboxes?


BeksOmega:
### Problem statement

Simple toolboxes do not act the same as normal toolboxes with respect to metrics, but it seems like they should.

### Expected Behavior

Blocks shouldn't be created behind the toolbox: (the best example is in RTL)
![importingblocks_categories_rtl](https://user-images.githubusercontent.com/25440652/52524365-05099d00-2c51-11e9-9f6e-c9414c0b3e84.gif)

Scrollbars should only scroll the accessible region:
![accessibleregion_categories](https://user-images.githubusercontent.com/25440652/52524412-9bd65980-2c51-11e9-9d2e-c9d9668dcd0d.gif)

### Actual Behavior

Blocks are created behind the toolbox: (the best example is in RTL)
![importingblocks_simple_rtl](https://user-images.githubusercontent.com/25440652/52524370-0dfa6e80-2c51-11e9-80d1-76b1725de3d9.gif)

Scrollbars scroll blocks behind the toolbox:
![accessibleregion_simple](https://user-images.githubusercontent.com/25440652/52524414-a0027700-2c51-11e9-8b9e-236304472128.gif)

I've also seen not a small amount of code scattered about specifically for getting flyouts to work correctly.

For example, [this](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/inject.js#L328) which fixes the block creation in LTR, but not in RTL (as evidenced above):
![importingblocks_categories_ltr](https://user-images.githubusercontent.com/25440652/52524377-266a8900-2c51-11e9-8b18-188c133b9f20.gif)
![importingblocks_simple_ltr](https://user-images.githubusercontent.com/25440652/52524379-29fe1000-2c51-11e9-864b-6348a247a504.gif)

See also: [Bubble Adjustment](https://github.com/google/blockly/pull/2279/files), [scrollCenter Adjustment](https://github.com/google/blockly/pull/2247/files#diff-d66993062276fda309492ac385508aeaR1759), [zoomControls Positioning](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/zoom_controls.js#L134), [Trashcan Positioning](https://github.com/google/blockly/blob/3342a375ff0cee0be42062765d24922be2917170/core/trashcan.js#L242), etc.

### Proposal

[...]

PR #2311, opened 2019-02-23
Changed Absolute Metrics to Avoid Simple Toolboxes

BeksOmega:
## The basics

- [x] I branched from develop

### Resolves

#2280 

### Proposed Changes

You can now only scroll the visible portion of the workspace:
absmetrics_scroll_left
absmetrics_scroll_top

Blocks are always created in the correct position:
absmetrics_positions_1
absmetrics_positions_2

Otherwise everything looks the same, but the math is different. Now simple toolboxes and category toolboxes work the same way math-wise and UI-wise, except in the case of zooming where they act different (UI-wise and math-wise).

### Reason for Changes

Should (hopefully) make the coordinate system easier to use in the future. Now you won't have to write special code for flyouts (unless you're writing zooming code).

### Test Coverage

* Category Flyout Rendering
* Simple Toolbox Flyout Rendering
* Mutator Workspaces
* Bubble Positioning
* scrollCenter
* zoomCenter
* zoomToFit
* Zoom on Wheel
* Zoom then Scroll
* Scroll on Wheel
* Block/Comment Bumping (unmovable workspace bounds)

Tested on:
* Desktop Chrome

### Additional Information

Note to @rachel-fenichel : This change makes it so the absoluteLeft and absoluteTop metrics change after you zoom, with a simple flyout. This may or may not affect the performance PR for scroll/zoom you had planned.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -149,13 +149,21 @@ Blockly.VerticalFlyout.prototype.position = function() {
   var edgeHeight = targetWorkspaceMetrics.viewHeight - 2 * this.CORNER_RADIUS;
   this.setBackgroundPath_(edgeWidth, edgeHeight);
 
-  var y = targetWorkspaceMetrics.absoluteTop;
-  var x = targetWorkspaceMetrics.absoluteLeft;
-  if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_RIGHT) {
-    x += (targetWorkspaceMetrics.viewWidth - this.width_);
-    // Save the location of the left edge of the flyout, for use when Firefox
-    // gets the bounding client rect wrong.
-    this.leftEdge_ = x;
+  // Y is always 0 since this is a vertical flyout.
+  var y = 0;
+  // If there is a toolbox.
+  if (targetWorkspaceMetrics.toolboxWidth) {
+    if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_LEFT) {
+      var x = targetWorkspaceMetrics.toolboxWidth;
+    } else {
+      var x = targetWorkspaceMetrics.viewWidth - this.width_;
+    }
+  } else {
+    if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_LEFT) {
+      var x = 0;
+    } else {
+      var x = targetWorkspaceMetrics.viewWidth;
+    }
   }
   this.positionAt_(this.width_, this.height_, x, y);
 };
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -272,7 +272,7 @@ Blockly.Mutator.prototype.setVisible = function(visible) {
     this.rootBlock_.setDeletable(false);
     if (this.workspace_.flyout_) {
       var margin = this.workspace_.flyout_.CORNER_RADIUS * 2;
-      var x = this.workspace_.flyout_.width_ + margin;
+      var x = this.workspace_.getFlyout_().getWidth() + margin;
     } else {
       var margin = 16;
       var x = margin;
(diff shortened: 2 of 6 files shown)


core/mutator.js
```diff
@@ -385,9 +385,9 @@ Blockly.Mutator.prototype.workspaceChanged_ = function() {
 Blockly.Mutator.prototype.getFlyoutMetrics_ = function() {
   return {
     viewHeight: this.workspaceHeight_,
-    viewWidth: this.workspaceWidth_,
+    viewWidth: this.workspaceWidth_ - this.workspace_.flyout_.width_,
```
rachel-fenichel: Use `this.workspace_.getFlyout_().getWidth()`.

The goal is to not directly access private properties, and to use setters and getters instead, even if that makes it more verbose.  That way if we change something internally (say, `getWidth()` changes to add a constant value to the stored `width_`) we don't need to find every single place that directly accesses that property.

`getFlyout_` on workspaceSvg is `package`, even though it has an underscore in the name (historical problem).

Accessing `this.workspace_` is fine because it's a private property on the object we're currently on.

There's plenty of code that doesn't follow these rules, but we try to make sure new code does, and that we update old code as we touch it.

rachel-fenichel:
Also, I don't have plans to work on a performance PR in the short term.  Feel free to dig into it if you're interested.

BeksOmega:
> Have you tested it in mutators, and in mutators in RTL? In particular I'm thinking of the if/else mutator for testing. Errors in this code usually show up as incorrect deletion areas in those mutators, meaning that when you drag a block out of the flyout in the mutator, you can't drag it back to delete it (or the opposite: it deletes no matter where you drop it).

mutatortesting

Tested on all workspace configurations, and it looks good.

Only problem is the delete area isn't being calculated correctly after zoom. Going to look into it and see if it's a simple fix.

BeksOmega:
> Only problem is the delete area isn't being calculated correctly after zoom (also exists on develop). Going to look into it and see if it's a simple fix.

Added #2319 to track the delete areas issue. This PR should be GTG.

core/flyout_vertical.js
```diff
@@ -149,13 +149,21 @@ Blockly.VerticalFlyout.prototype.position = function() {
   var edgeHeight = targetWorkspaceMetrics.viewHeight - 2 * this.CORNER_RADIUS;
   this.setBackgroundPath_(edgeWidth, edgeHeight);
 
-  var y = targetWorkspaceMetrics.absoluteTop;
-  var x = targetWorkspaceMetrics.absoluteLeft;
-  if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_RIGHT) {
-    x += (targetWorkspaceMetrics.viewWidth - this.width_);
-    // Save the location of the left edge of the flyout, for use when Firefox
```
samelhusseini: @BeksOmega digging this up because I'm chasing why ``this.leftEdge_`` was removed here, but not [here](https://github.com/google/blockly/blob/master/core/flyout_vertical.js#L355). 

If you can remember, I'm wondering if you think this was deliberate or accidental?

BeksOmega: I can't remember, but I can't imagine it being deliberate.

samelhusseini: Okay, I'll look into whether this code is still needed.

BeksOmega: I couldn't reproduce #1425 on the latest firefox if that helps. And it looks like the root bug [1066435 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1066435)was closed.

samelhusseini: Cool, good to know. Thanks

PR #2313, opened 2019-02-26
Add dropdowndiv and use it for the colour picker and angle fields

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves
Part of rendering update.

### Proposed Changes

This pull request adds the dropdowndiv from pxt-blockly/scratch-blocks and uses it for the colour picker field.
New rendering:
image

Old rendering: 
image

This just puts the same HTML into the dropdown div instead of the widget div, and lets the dropdown div do the positioning.

### Reason for Changes

As part of rendering updates we're switching to using the dropdown div for any widgets that can be positioned as a dropdown.  Adding this (in addition to the current widget div) lets us have fields with dropdowns + inline editors, such as a mixed text input/dropdown.

The other goal is for fields to not need to calculate positioning for the dropdown, because there's a single set of rules for positioning that should be limited to one file.

### Test Coverage
Tested by opening the colour picker at a variety of zoom levels and in different locations on the screen.

### Additional Information

- There's a small pixel offset from center that mostly matters when zoomed far out.
- I'm not updating the actual dropdown field yet because its special positioning is actual relevant (aligns dropdown text to field text, instead of being centered).


rachel-fenichel:
@NeilFraser fyi

rachel-fenichel:
I just added the angle picker as well.  Old:
image

and new:
image

It doesn't work right on mobile, but that's a problem with the angle field itself and how it interacts with the text input field trying to prompt on mobile.  I will file a bug for that and look into it separately.

core/dropdowndiv.js
```diff
@@ -0,0 +1,519 @@
...
+
+/**
+ * Callback for when the drop-down is hidden.
+ * @type {Function}
+ */
+Blockly.DropDownDiv.onHide_ = 0;
+
+/**
+ * Create and insert the DOM element for this div.
+ * @param {Element} container Element that the div should be contained in.
```
RoboErikG: There's no param.

core/dropdowndiv.js
```diff
@@ -0,0 +1,519 @@
...
+  // Transition animation for transform: translate() and opacity.
+  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' +
+    Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
+    'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
+};
+
+/**
+ * Set an element to maintain bounds within. Drop-downs will appear
+ * within the box of this element if possible.
+ * @param {Element} boundsElement Element to bound drop-down to.
```
RoboErikG: s/bound/bind

core/dropdowndiv.js
```diff
@@ -0,0 +1,519 @@
...
+    renderX = primaryX;
+    renderY = primaryY + Blockly.DropDownDiv.PADDING_Y;
+    renderedSecondary = false;
+  }
+  // First calculate the absolute arrow X
+  // This needs to be done before positioning the div, since the arrow
+  // wants to be as close to the origin point as possible.
+  var arrowX = renderX - Blockly.DropDownDiv.ARROW_SIZE / 2;
+  // Keep in overall bounds
+  arrowX = Math.max(boundPosition.left, Math.min(arrowX, boundPosition.left + boundSize.width));
```
RoboErikG: Should the min call be "min(arrowX, boundPosition.left + boundSize.width - (divsize+Blockly.DropDownDiv.ARROW_SIZE) / 2)"? This would match the arrow's rightmost bound to the renderX bound calculated below.

An even better approach might be to adjust the renderX first and then base the arrowX on the already fixed renderX.

core/dropdowndiv.js
```diff
@@ -0,0 +1,519 @@
...
+  // Adjust the x-position of the drop-down so that the div is centered and within bounds.
+  var centerX = divSize.width / 2;
+  renderX -= centerX;
+  // Fit horizontally in the bounds.
+  renderX = Math.max(
+      boundPosition.left,
+      Math.min(renderX, boundPosition.left + boundSize.width - divSize.width)
+  );
+  // After we've finished caclulating renderX, adjust the arrow to be relative to it.
+  arrowX -= renderX;
```
RoboErikG: This is just removing renderX from the original positioning of the arrow, but after things have been adjusted. You might as well just set arrowX = -ARROW_SIZE/2. I think the ordering in this method is all wrong.

core/dropdowndiv.js
```diff
@@ -0,0 +1,519 @@
...
+  // Pad the arrow by some pixels, primarily so that it doesn't render on top of a rounded border.
+  arrowX = Math.max(
+      Blockly.DropDownDiv.ARROW_HORIZONTAL_PADDING,
+      Math.min(arrowX, divSize.width -
+        Blockly.DropDownDiv.ARROW_HORIZONTAL_PADDING - Blockly.DropDownDiv.ARROW_SIZE)
+  );
+
+  // Calculate arrow Y. If we rendered secondary, add on bottom.
+  // Extra pixels are added so that it covers the border of the div.
+  var arrowY = (renderedSecondary) ? divSize.height - Blockly.DropDownDiv.BORDER_SIZE : 0;
```
RoboErikG: Is this code ignoring the tertiary case?

core/field_colour.js
```diff
@@ -217,21 +261,15 @@ Blockly.FieldColour.prototype.setColumns = function(columns) {
...
-  var anchorBBox = this.getScaledBBox_();
+  Blockly.DropDownDiv.hideWithoutAnimation();
+  Blockly.DropDownDiv.clearContent();
 
-  // Create and add the colour picker, then record the size.
   var picker = this.createWidget_();
-  Blockly.WidgetDiv.DIV.appendChild(picker);
-  var paletteSize = goog.style.getSize(picker);
+  Blockly.DropDownDiv.getContentDiv().appendChild(picker);
+  Blockly.DropDownDiv.setColour('#ffffff', '#dddddd');
```
RoboErikG: Should these colour constants be pulled out? Needs at least a TODO.

core/dropdowndiv.js
rachel-fenichel: In the tertiary case the arrow doesn't need to be shown at all.

core/field_colour.js
RoboErikG: I was complaining more about the field having the colour set in the middle of a block of code without an obvious place to change it.

rachel-fenichel: I did wonder if I was missing the point.

rachel-fenichel:
I'm not sure a CSS class is the right solution, actually.  The dropdown div can get hidden by somebody other than the owner (e.g. from hideChaff) and the class change is on the dropdown div, not its content div.  That means that when closing it, we need to know which classes may have been set so that we can remove those classes, unless we want to delete the div and start over every time.  I don't like either of those options.  So I think constraining it to setColour and using constants makes more sense.

Thoughts?

RoboErikG:
Fine with me. I just want the constants pulled out somewhere obvious so they can be made themeable later.

rachel-fenichel:
Done.

Issue #2312, opened 2019-02-23
Context Menus not Receiving Clicks


BeksOmega:
### Problem statement

You know that [pull request](https://github.com/google/blockly/pull/2286) where I added code to receive document clicks? Turns out it breaks all context menus.

### Expected Behavior

Context menus should recieve clicks and call their callbacks.

### Actual Behavior

They don't.

### Steps to Reproduce

1. Add a block to the workspace.
2. Right click it.
3. Click the "add comment" option. Observe how the context menu disapears, and no comment box appears.

Or if you want to be more thorough:
1. Add the below code at ~ln 129 in playground:
```
workspace.configureContextMenu = function(options) {
    options.push({
      enabled: true,
      text: 'Test',
      callback: function() {
        console.trace('callback');
      }
    })
  }
```
2. Open the playground.
3. Right click on the workspace.
4. Click the "Test" option. Observe how the context menu disapears and no trace was logged to the console.

### Stack Traces

I added a console.trace() call to the document mouse down subscriber:

```
(anonymous) | @ | inject.js:441
-- | -- | --
  | wrapFunc  | @ | blockly.js:492
```

Then I disabled that subscriber and grabbed the trace from the above custom option.
```
callback | @ | playground.html:134
-- | -- | --
  | goog.events.EventTarget.fireListeners | @ | eventtarget.js:285
  | goog.events.EventTarget.dispatchEventInternal_ | @ | eventtarget.js:383
  | goog.events.EventTarget.dispatchEvent | @ | eventtarget.js:196
  | goog.ui.Control.performActionInternal | @ | control.js:1378
  | goog.ui.Control.handleMouseUp | @ | control.js:1321
  | childCtor.base | @ | base.js:2123
  | goog.ui.MenuItem.handleMouseUp | @ | menuitem.js:258
  | goog.ui.Container.handleChildMouseEvents | @ | container.js:672
  | goog.events.fireListener | @ | events.js:753
  | goog.events.handleBrowserEvent_ | @ | events.js:879
  | f | @ | events.js:296
```
### Operating System and Browser

I am assuming this is not browser specific.

### Additional Information

I was just going to fix this, and then I realized I didn't know how. I'm still willing to help if I can!

BeksOmega:
[...]

PR #2326, opened 2019-03-10
Fix issue 2061 (make list from text block throws uncaught Error)

### LINKED ISSUE #2061: make list from text block throws uncaught Error
### Problem statement
If you put a shadow block in the "make list from text" block and then select "text from list" in the dropdown,  the following error occurs.

```
connection.js:328 Uncaught Error: Connection checks failed. Input "INPUT" connection on "lists_split" block (id="+{VZiOpQGp#Fx~1R:F#R") expected Array, found String
```
listfromtext

### Expected Behavior
I would expect the shadow block to vanish when you switch to "text from list", and then reappear when you switch back to "list from text".  There should also be no `Uncaught Error`

### Actual Behavior
When you select "text from list" in the drop down, the shadow block disconnects and then you cannot interact with it.  You also see both items in the drop down as checked.
broken

### Steps to Reproduce
You can use this xml in the blockly playground to reproduce.
```
  <block type="lists_split" id="+{VZiOpQGp#Fx~1R:F#R" x="88" y="38">
    <mutation mode="SPLIT"></mutation>
    <field name="MODE">SPLIT</field>
    <value name="INPUT">
      <shadow type="text" id="2=*3(pV]Wqhs)y%Kutr)">
        <field name="TEXT"></field>
      </shadow>
    </value>
    <value name="DELIM">
      <shadow type="text" id="w]%su!7F(Ep@r_`^3#19">
        <field name="TEXT">,</field>
      </shadow>
    </value>
  </block>
</xml>
```
stepstoreproduce

### Stack Traces

```
Uncaught Error: Connection checks failed. Input "INPUT" connection on "lists_split" block (id="+{VZiOpQGp#Fx~1R:F#R") expected Array, found String
    at Blockly.RenderedConnection.Blockly.Connection.checkConnection_ (connection.js:328)
    at Blockly.RenderedConnection.Blockly.Connection.connect (connection.js:396)
    at Blockly.RenderedConnection.Blockly.Connection.respawnShadow_ (connection.js:531)
    at Blockly.RenderedConnection.respawnShadow_ (rendered_connection.js:354)
    at Blockly.RenderedConnection.Blockly.Connection.disconnect (connection.js:495)
    at Blockly.BlockSvg.Blockly.Block.unplugFromRow_ (block.js:334)
    at Blockly.BlockSvg.Blockly.Block.unplug (block.js:316)
    at Blockly.RenderedConnection.onCheckChanged_ (rendered_connection.js:420)
    at Blockly.RenderedConnection.Blockly.Connection.setCheck (connection.js:599)
    at Blockly.Input.setCheck (input.js:209)
```

cshung:
## The basics

- [x] I branched from develop

### Resolves

#2061 make list from text block throws uncaught Error

### Proposed Changes

When switching the operation, disconnect the input block first, if the input block is a shadow, dispose it and make sure the shadow dom is null, otherwise bump it away.

### Reason for Changes

The reason why I disconnect the input block is that if we change the operation, we change the type, and whatever got connected will have a type mismatch. Thus, to avoid the error, we need to make sure the block is disconnected.

The shadow is useless outside of the block, so I disposed the shadow.

If it is not a shadow, we want to keep it. After disconnecting, the input block stays at the same location, therefore we need to bump it away.

### Test Coverage

Because of #2312, the context menu is not working in the develop branch, so we cannot repro the bug in the develop branch.

Because of that, the bug was reproduced on master and fixed on master instead. [Here](https://github.com/google/blockly/compare/master...cshung:dev/andrewau/fix-issue-2061-master) is where the fix where originally developed. For that branch, the only platform I tested is Chrome on MacOSX.

I think we should wait until #2312 is fixed, and then test on the various platforms.

Tested on:
Desktop Chrome

### Additional Information

This is my first pull request on blockly. :)

### DIFF
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -829,6 +829,21 @@ Blockly.Blocks['lists_split'] = {
    * @this Blockly.Block
    */
   updateType_: function(newMode) {
+    var mode = this.getFieldValue('MODE');
+    if (mode != newMode) {
+      this.setFieldValue(newMode, 'MODE');
+      var inputConnection = this.getInput('INPUT').connection;
+      inputConnection.setShadowDom(null);
+      var inputBlock = inputConnection.targetBlock();
+      if (inputBlock) {
+        inputConnection.disconnect();
+        if (inputBlock.isShadow()) {
+          inputBlock.dispose();
+        } else {
+          this.bumpNeighbours_();
+        }
+      }
+    }
     if (newMode == 'SPLIT') {
       this.outputConnection.setCheck('Array');
       this.getInput('INPUT').setCheck('String');


RoboErikG:
Change looks good but please revert the compressed and uncompressed files. We usually don't submit a new build of those files until the release so we can keep the commit history a bit cleaner.

RoboErikG:
Thanks for the fix!

PR #2334, opened 2019-03-21
Keyboard nav cleanup

alschmiedt:
## The basics

- [ ] I branched from develop

### Proposed Changes
Add two layers for keyboard navigation. The bottom layer Cursor API will control walking through an AST and the top layer navigation will decide where to walk based on key bindings. 

### Reason for Changes
To further blockly accessibility.

### Test Coverage
Cursor is tested using mocha. 

Tested on:


core/workspace_svg.js
```diff
@@ -1054,6 +1053,11 @@ Blockly.WorkspaceSvg.prototype.pasteBlock_ = function(xmlBlock) {
   Blockly.Events.disable();
   try {
     var block = Blockly.Xml.domToBlock(xmlBlock, this);
+    if (Blockly.Navigation.insertionConnection_) {
+      Blockly.Navigation.insertBlock(block, Blockly.Navigation.insertionConnection_);
+      return;
```
rachel-fenichel: Does the finally run if you return early inside a try block?  @NeilFraser

core/blockly.js
```diff
@@ -276,7 +276,16 @@ Blockly.onKeyDown_ = function(e) {
     }
   } else if (Blockly.keyboardAccessibilityMode_
       && (e.keyCode === goog.events.KeyCodes.UP
-        || e.keyCode === goog.events.KeyCodes.DOWN)) {
+        || e.keyCode === goog.events.KeyCodes.DOWN
```
AnmAtAnm: A more common pattern for this is to return a "handled" boolean from the called method (`navigate()`, in this case).  In that way, you don't have to sync to locations of used keys. It does mean, if there is key mapping overlap, the function can override the main assignment.


```
   if (Blockly.keyboardAccessibilityMode_ && Blockly.Navigation.navigate(e)) {
     // return? maybe?  This block is really here to skip the following if else conditions.
   } else if ...

File: core/cursor.js [hunk 3]
```diff
@@ -2,7 +2,7 @@
  * @license
  * Visual Blocks Editor
  *
- * Copyright 2012 Google Inc.
+ * Copyright 2011 Google Inc.
```
AnmAtAnm: 2019? Since this is a new file.

core/workspace_svg.js
AnmAtAnm: Yes, it does.

core/gesture.js
```diff
@@ -760,10 +760,10 @@ Blockly.Gesture.prototype.doBlockClick_ = function() {
  */
 Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
   Blockly.keyboardAccessibilityMode_ = false;
-  Blockly.cursor.hide();
+  this.creatorWorkspace_.cursor.hide();
   if (e.shiftKey) {
     // Show the cursor at the specified position
-    Blockly.cursor.workspaceShow(e);
+    this.creatorWorkspace_.cursor.workspaceShow(e);
```
RoboErikG: It might be cleaner to call cursor.setLocation(this.creatorWorkspace_, false, [x, y]) where x and y are coordinates (possibly as workspace units).

This means you have one code path for moving the cursor. Using workspace units for x and y means you can move the cursor on a headless workspace as well.

core/workspace.js
```diff
@@ -133,6 +140,14 @@ Blockly.Workspace.prototype.isClearing = false;
  */
 Blockly.Workspace.prototype.MAX_UNDO = 1024;
 
+/**
+ * Adds cursor for keyboard navigation.
+ * @return {Blockly.Cursor} Cursor for keyboard navigation.
+ */
+Blockly.Workspace.prototype.addCursor = function() {
```
RoboErikG: nit: s/add/new. This method is just creating the cursor and returning it, it doesn't add it to anything.

Should also be protected.

PR #2335, opened 2019-03-22
Set theme fix

### LINKED ISSUE #2331: Blockly.setTheme errors out if Blockly.getMainWorkspace() does not return a value
### Problem statement

Blockly.setTheme errors out if Blockly.getMainWorkspace() does not return a value. (This can occur if using Blockly.setTheme with no workspace yet opened, such as when attempting to initialize Blockly information before injecting code.)

### Expected Behavior

If there is no main workspace yet,  Blockly.setTheme should just set Blockly.theme_ and move on.

### Actual Behavior

Blockly.setTheme errors out with "cannot read property getAllBlocks of undefined".

### Steps to Reproduce

Load up Blockly in any playground and attempt to call Blockly.setTheme before initializing the workspace.

* Desktop Chrome

(I expect it applies to all of them, though.)

### Additional Information

This is a less common use case, but certainly not an unthinkable one; I ran into it earlier today while adding Blockly.setTheme in order to update a nodejs Blockly project we've been using so as not to cause the blocks to error out when loaded up.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#2331 

### Proposed Changes
Check for a workspace before trying to refresh the theme on all the blocks and categories in the workspace.

### Reason for Changes
Blockly.setTheme was throwing an error when the workspace had not yet been created.

### Test Coverage
Tested on:
 Desktop Chrome

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -690,28 +690,30 @@ Blockly.setTheme = function(theme) {
   this.theme_ = theme;
   var ws = Blockly.getMainWorkspace();
 
-  //update all blocks in workspace that have a style name
-  this.updateBlockStyles_(ws.getAllBlocks().filter(
-      function(block){
-        return block.getStyleName() !== undefined;
-      }
-  ));
+  if (ws) {
+    //update all blocks in workspace that have a style name
+    this.updateBlockStyles_(ws.getAllBlocks().filter(
+        function(block){
+          return block.getStyleName() !== undefined;
+        }
+    ));
 
-  //update blocks in the flyout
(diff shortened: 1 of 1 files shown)


core/blockly.js
```diff
@@ -690,6 +690,16 @@ Blockly.setTheme = function(theme) {
...
 
+  if (ws) {
+    this.refreshTheme_(ws);
+  }
+};
+
+/**
+ * Refresh the theme for all items on the workspace.
+ * @param {Blockly.Workspace} ws Blockly workspace to refresh theme on.
+ */
```
RoboErikG: missing private annotation

PR #2344, opened 2019-03-27
Fix zoom bug while dragging in Mutators

NeilFraser:
If one opens a mutator, grabs a block, and scrolls the mouse wheel while dragging, coordinates got all messed up.  This PR fixes that.


rachel-fenichel:
LG, though it's hard to actually find the relevant change when looking at the PR as a whole, because there are so many lint changes.

NeilFraser:
The previous code had an unexpected side effect.  Mutators have no wheel explicit zoom/move options.  By short-circuiting mouse wheel events in mutators, the later check for a gesture was never checked.  Next the event bubbled up to the main workspace, where there was no gesture.

The new code is straightforward in that it checks gesture at each level and aborts if there is one.

RoboErikG:
Just a note, this change has a lint error that caused Travis to fail.

PR #2346, opened 2019-03-28
Random upgrades

NeilFraser:
Procrastination is a wonderful way to do cleaning...

### DIFF
--- a/appengine/storage.js
+++ b/appengine/storage.js
@@ -169,10 +169,10 @@ BlocklyStorage.monitorChanges_ = function(workspace) {
     var xmlText = Blockly.Xml.domToText(xmlDom);
     if (startXmlText != xmlText) {
       window.location.hash = '';
-      workspace.removeChangeListener(bindData);
+      workspace.removeChangeListener(change);
     }
   }
-  var bindData = workspace.addChangeListener(change);
+  workspace.addChangeListener(change);
 };
 
 /**
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -986,7 +986,7 @@ Blockly.BlockSvg.prototype.setBorderColour_ = function(rgb, colourTertiary) {
  * Sets the colour of shadow blocks.
  * @param {!string} rgb Primary colour of the block.
  * @param {?string} colourSecondary Colour for shadow block.
- * @return {!string} hexColour The background colour of the block.
+ * @return {!string} The background colour of the block.
  */
 Blockly.BlockSvg.prototype.setShadowColour_ = function(rgb, colourSecondary) {
   var hexColour;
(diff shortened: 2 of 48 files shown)


appengine/storage.js
```diff
@@ -169,10 +169,10 @@ BlocklyStorage.monitorChanges_ = function(workspace) {
     var xmlText = Blockly.Xml.domToText(xmlDom);
     if (startXmlText != xmlText) {
       window.location.hash = '';
-      workspace.removeChangeListener(bindData);
+      workspace.removeChangeListener(change);
     }
   }
-  var bindData = workspace.addChangeListener(change);
+  workspace.addChangeListener(change);
```
AnmAtAnm: While true in practice, the `addChangeListener(..)` documentation does not guarantee the returned function is the same as the passed function.

core/block_svg.js
```diff
@@ -986,10 +986,9 @@ Blockly.BlockSvg.prototype.setBorderColour_ = function(rgb, colourTertiary) {
  * Sets the colour of shadow blocks.
  * @param {!string} rgb Primary colour of the block.
  * @param {?string} colourSecondary Colour for shadow block.
- * @return {!string} hexColour The background color of the block.
+ * @return {!string} hexColour The background colour of the block.
```
AnmAtAnm: Remove `hexColour`. Return values don't have names.

NeilFraser: Fixed this and 65 other @return annotations.

core/gesture.js
```diff
@@ -954,3 +954,18 @@ Blockly.Gesture.prototype.getInsertionMarkers = function() {
   }
   return [];
 };
+
+/**
+ * Is a drag or other gesture currently in progress on any workspace?
+ * @return {boolean} True if gesture is occurring.
+ */
+Blockly.Gesture.inProgress = function() {
+  // If a drag is occurring, then it involves a selected block on a workspace.
```
rachel-fenichel: What about a workspace drag?

rachel-fenichel:
FYI we can enforce return vs returns with the `prefer` option in the `valid-jsdoc` rule: https://eslint.org/docs/rules/valid-jsdoc#options

If it's important we can probably enforce capitals at the start of param/return descriptions with `matchDescription`, documented in the same place.

core/gesture.js
NeilFraser: Fixed, now searching all workspaces.

appengine/storage.js
NeilFraser: Nowhere in Blockly are we still using the returned function (this was the last straggler).  I've changed the docs to indicate that the return value is obsolete and to be ignored.

PR #2355, opened 2019-04-02
Adds check for a targetConnection

### LINKED ISSUE #2336: Blocks unplugged during drag can be lost/deleted
### Problem statement

It appears that dragging some blocks with mixed external input types can result in losing already-created blocks that are unplugged by dragging over them. This can result in significant loss of work, and "undo" does not restore the lost block(s).

### Expected Behavior

Dragging a new block across existing blocks should never cause loss of existing blocks.

### Actual Behavior

Dragging, without dropping, can orphan and/or delete existing blocks.

### Steps to Reproduce

I couldn't get the problem to occur with any of the "standard" blocks, so I created a set of block definitions to demonstrate the problem. The issue seems to require a mixture of type-checked inputs, and seems to be related to the auto-healing/insertion logic, although I don't know what code is effected. I've created a branch in github to demonstrate the issue, with instructions at the top of the page.

https://github.com/dalnefre/blockly/tree/bugged-drag/demos/bugged

### Stack Traces

none

demonstrated on 
  * Desktop Chrome
  * macOS Mojave (10.14.3)

but probably not browser/OS-specific

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#2336 

### Proposed Changes
Currently, we are looking at all value inputs on a block. If there is more than one value input then we return null which means that we don't have a connection to reattach to the parent block. The proposed fix is to only look for value inputs that have something connected. In the case that something is connected we want to return that connection so that we can reattach this block to the parent. 

### Reason for Changes
When there was a block with more than one value input the block that was connected to the insertion marker would not get reattached to the parent. Therefore when we hid the insertion marker the block would disappear with it. 

### Test Coverage

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -397,8 +397,12 @@ Blockly.Block.prototype.unplugFromRow_ = function(opt_healStack) {
 };
 
 /**
- * Returns the connection on the only value input on the block, or null if the
- * number of value inputs is not one.
+ * Returns the connection on the value input that is connected to another block.
+ * When an insertion marker is connected to a connection with a block already
+ * attached, the connected block is attached to the insertion marker.
+ * Since only one block can be displaced and attached to the insertion marker
+ * this should only ever return one connection.
+ *
  * @return {Blockly.Connection} The connection on the value input, or null.
  * @private
  */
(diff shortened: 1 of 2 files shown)


RoboErikG:
Looks like you need to update the test_block_row_unplug_multi_inputs_middle test.

rachel-fenichel:
Why do this by checking for a target connection?  I expected the check to consider the types of the connections and return the only one that matches the type, since that's the root cause in the bug report.

alschmiedt:
See #2370 for update on this issue

Issue #2364, opened 2019-04-06
All mutators broken


NeilFraser:
Opening any mutator in 'develop' results in this console error:

mutator.js:397 Uncaught TypeError: this.workspace_.getFlyout_ is not a function
    at Blockly.Mutator.getFlyoutMetrics_ (mutator.js:397)
    at Blockly.VerticalFlyout.position (flyout_vertical.js:140)
    at Blockly.VerticalFlyout.reflowInternal_ (flyout_vertical.js:412)
    at Blockly.VerticalFlyout.Blockly.Flyout.reflow (flyout_base.js:764)
    at Blockly.VerticalFlyout.Blockly.Flyout.show (flyout_base.js:509)
    at Blockly.Mutator.setVisible (mutator.js:265)
    at Blockly.Mutator.Blockly.Icon.iconClick_ (icon.js:131)
    at Blockly.Mutator.iconClick_ (mutator.js:114)
    at SVGGElement.wrapFunc (blockly.js:491)

PR #2357, opened 2019-04-03
Rename WorkspaceSvg.getFlyout_ to WorkspaceSvg.getFlyout

TiaraIsTan:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

https://github.com/google/blockly/issues/1101

### Proposed Changes

Renames instances of ".getFlyout_" to ".getFlyout"

### Reason for Changes

Due to issue #1101

PR #2365, opened 2019-04-06
Fix getFlyout in mutator

NeilFraser:
PR #2357 was incomplete.  Fixes issue #2364


rachel-fenichel:
This still shows up as a diff but I don't see why--Abby also fixed this, in a different PR.  Either way, closing as fixed.

Issue #1623, opened 2018-02-15
Document the process of creating a custom field.


AnmAtAnm:
As of #1594, custom fields are now possible without modifying the core blockly code.

## Audit field APIs

Improve the `Field` class APIs to a level we are willing to support long term.

 * [ ] Use setText instead of setValue in fieldTextInput (#860).
 * [ ] Extract common Field support methods into new field_util.js (#1624).
 * [ ] Consistently call Field.init() from subclass .init() (or refactor to put ensure it is called in other ways).
 * [ ] Split `Field.init()` into `Field.initView()` and `Field.initModel()` (#1010). 
 * [ ] Consider splitting Field view and model more completely. For instance, `updateWidth()` is another view specific method.
 * [ ] Validation: Do we validate the text of the field, the value, or both (need separate APIs)?

## Document custom field development process and concerns

Write the documentation for the following aspects of the improved API.

 * [ ] Field class hierarchy, denoting where good extension points are.
 * [ ] Document required Field methods and life cycle of a Field object.
 * [ ] Difference between setText() vs setValue().
 * [ ] Timeline and expectations between `showEditor()` and registering event listeners and `hideChaff()`.
 * [ ] How to attach popup editors to a field.
 * [ ] Document the registration functions.

BeksOmega:
Would you peoples mind if I took a look at refactoring some of the field API stuffs?

If you don't I'll work up a little class diagram that you can look at before I get started. If you don't want me to touch it that's cool too =)

RoboErikG:
If you want to tackle some of the concrete issues or build out some more detail on others feel free.

I don't think we've fully discussed everything on this list, so there may be some suggestions and back and forth before you get the go ahead to implement some of them. =)

Also, Rachel is out this week and should have final say on any big changes.

BeksOmega:
So I read through a bunch of the field files and came up with a [WIP class diagram](https://www.lucidchart.com/invitations/accept/b049efd6-c5fe-49d6-a484-e6708a52f6a6). It has the main abstract field class and then three other concrete classes that I worked up to test if the abstract class made sense. The properties of the abstract field class that are blue are the ones I think should be optionally overriden by subclasses.

There are a couple things to note:
- This adds an isDirty_ property to the field so that [getSize ](https://github.com/google/blockly/blob/82871b3d24563388a6f113da8984694e9f37b53e/core/field.js#L454)can use that instead of checking if the width has been set to zero.
[...]

PR #2389, opened 2019-04-18
Added isSerializable and SERIALIZABLE to Field

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #1623 

### Proposed Changes

Adds a SERIALIZABLE property to fields. Adds an isSerializale function that handles warning & backwards compatibility logic.

I also reorganized the test-blocks toolbox so now it looks like this:
PlaygroundToolboxNew

### Reason for Changes

Sometimes you would want to programmatically modify a field, but not allow users to modify it, and save it to XML. Before you would have to create a mutator for this purpose. Now you can simply use a serializable field.

### Test Coverage

I added unit tests! I added tests for the isSerializable function to make sure it works in a backwards compatible way. I also added serialization tests to make sure the XML for different fields was comming out correctly.

Here's also a quick example of some fields being serialized and deserialized:
Serialization

Here's the generated XML for that example:
```
  <variables>
    <variable type="" id="o(EVLYWaZFZca|N:K.7I">not item</variable>
  </variables>
  <block type="test_fields_angle" id="mkWW9%.I_@;R#$W[H{PM" x="38" y="113">
    <field name="FIELDNAME">89</field>
  </block>
  <block type="test_fields_date" id="flqjVdY:q[i9i9quPeNH" x="37" y="163">
    <field name="FIELDNAME">2019-04-18</field>
  </block>
  <block type="test_fields_checkbox" id="ek{drzs2vnp^RIzb[~_." x="38" y="212">
    <field name="CHECKBOX">FALSE</field>
  </block>
  <block type="test_fields_colour" id="0kbOz)J7$}kH;aG$/7ku" x="38" y="263">
    <field name="COLOUR">#33ff33</field>
  </block>
  <block type="test_fields_text_input" id="f%yfBd#74I3,;do8vGAu" x="37" y="312">
    <field name="TEXT_INPUT">not default</field>
  </block>
  <block type="test_fields_variable" id="2![yICR}hguA1=0)-x`^" x="37" y="362">
    <field name="VARIABLE" id="o(EVLYWaZFZca|N:K.7I" variabletype="">not item</field>
  </block>
</xml>
```

Tested on:
* Desktop Chrome

### Additional Information

SERIALIZABLE is false by default. This is so that it is backwards compatible with any non-editable fields that people had created in the past. 
IMG_20190418_082838
The problem with this is that if you had created a custom editable field before you'll get a warning message when it goes to serialize, because EDITABLE is true and SERIALIZABLE is not, which is an invalid state. It still serializes! But you will get that warning. I don't necessarily /like/ this but it's the best solution I could come up with.

Also I'm planning on adding field_label_serializable and field_image_serializable fields, but the field_label_serializable stuff started cluttering up this PR and the fromXML toXML change needs to get in before field_image_serializable can be added.


core/field.js
```diff
@@ -268,15 +282,34 @@ Blockly.Field.prototype.updateEditable = function() {
...
 };
 
+/**
+ * Check whether this field should be serialized by the XML renderer.
+ * Handles the logic for backwards compatibility and incongruous states.
+ * @return {boolean} Whether this field should be serialized or not.
+ */
+Blockly.Field.prototype.isSerializable = function() {
+  var isSerializable = false;
+  if (this.name) {
```
rachel-fenichel: Why does the name matter?

core/field_angle.js
```diff
@@ -66,6 +66,22 @@ Blockly.FieldAngle.fromJson = function(options) {
   return new Blockly.FieldAngle(options['angle']);
 };
 
+/**
+ * Editable fields usually show some sort of UI indicating they are editable.
+ * They will also be saved by the XML renderer.
+ * @type {boolean}
+ * @const
+ */
+Blockly.FieldAngle.prototype.EDITABLE = true;
```
rachel-fenichel: Since EDITABLE defaults to true on fields, you don't need to redeclare it on each of these classes.  You only need to declare it for classes that set it to false.  The inverse is true for SERIALIZABLE.

rachel-fenichel:
One lint error on travis: https://travis-ci.org/google/blockly/builds/521780332#L536

core/field.js
NeilFraser: If there is no name, then serializing doesn't make sense since you can't refer to the field.

rachel-fenichel: Oh right.

NeilFraser:
> One lint error on travis: https://travis-ci.org/google/blockly/builds/521780332#L536

That console log looks appropriate and useful.  What's Travis' logic that flags this console usage and not others?

core/field_angle.js
BeksOmega: Fixed.

BeksOmega:
> We don't have mocha running on our server yet--are you running these tests locally?

Yes I added them to your handy-dandy index.html file. Here's a pic of my added tests (ignore the variable fields tests):
TestsPassing

rachel-fenichel:
> That console log looks appropriate and useful. What's Travis' logic that flags this console usage and not others?

Not sure, actually.  it should all be running with the same eslint rules, since our eslintrc is in our root directory, and that has the no-console rule disabled.  Weird.

BeksOmega:
Ok @rachel-fenichel I changed the mocha eslint config to extend the root one, and then I fixed the lint errors in the other mocha files. I can separate those changes into a separate PR if you like, otherwise I think this is good to go.

rachel-fenichel:
Awesome, thanks for tracking that down.

Issue #2169, opened 2018-12-13
Refactor Field validation


alschmiedt:
### Problem statement
Currently, `field.validator_` is called in some of the methods of classes that inherit from Field, but not called in the Field class itself (whether by `callValidator(..)` or more direct methods). When and where the validators are called in the field classes is not standardized and therefore makes coding and fixing bugs for all field types difficult. It would therefore be more consistent to have the `callValidator(..)` method called in the Field setValue method.

RoboErikG:
The base Field now calls the validator as part of the constructor. There is still additional cleanup to fields:

- [ ] Audit fields for cases where the individual field is calling validators instead of the base class.
- [ ] Where appropriate refactor to allow the base class to handle it.
- [ ] As a result of this change, field_number's class validator will to revalidate when changed. It should be updated to re-validate iff the constraints are changed after initializing.

BeksOmega:
I ran into this while working on the[ custom fields demo](https://github.com/google/blockly/issues/1623#issuecomment-486388067).

### Problem Statement
1) As mentioned in the OP field validation isn't consistent. This makes it hard to fix bugs.
2) setValue (in the base class) expects fields to only hold one value. This can be a problem if you're trying to validate something like a vector, a curve, or turtles.
3) setValue (in the base class) expects the value to be text. This can be a problem if you're trying to validate something like an image (not like a url, like a bitmap), or turtles.

### Proposal
Refactor setValue on the base class to be a [template](https://en.wikipedia.org/wiki/Template_method_pattern) that looks something like this:

```
// What the value contains (be it a string, object, etc) will only be
// known to the developer and the concrete field. To the abstract field
// it will be anonymous.
setValue(value) {
	// Validator defined on the class.
	var newValue = doClassValidation(value);
	// Validator defined by the developer.
	newValue = doUserValidation(newValue);
	// This is where the field stores the validated values however they want
	// (e.g. break an object up into separate variables).
	doValueUpdate(newValue);
	// This makes sure that rerendering is consistent.
	forceRerender();
}
```
(Plus null checks, sameness checks, events, etc.)

This way the developer could just define the 'do' methods, and know they aren't forgetting anything. And it would be easier to debug because we would know the path would be the same every time.

Also great is that if anyone has overriden setValue this is still backwards compatible because all of the changes would be on the abstract field class.

How does that sound to you peoples?

rachel-fenichel:
@NeilFraser do you have any thoughts on this?

[...]

PR #2392, opened 2019-04-18
Validate newValue parameter in setValue

asunwoo98:
## The basics

- [x] I branched from develop

### Resolves

#2169 

### Proposed Changes

Validates the input value in setValue of field.js by using callValidator function

### Reason for Changes

Proposed change in #2169 

### Test Coverage

Tested on:

### Additional Information

Might potentially introduce redundancies in subclasses?

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -574,6 +574,12 @@ Blockly.Field.prototype.setValue = function(newValue) {
     // No change if null.
     return;
   }
+  // Validate input.
+  var validated = this.callValidator(newValue);
+  if (validated !== null) {
+    newValue = validated;
+  }
+  // Check for change.
   var oldValue = this.getValue();
   if (oldValue == newValue) {
     return;
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -49,7 +49,6 @@ Blockly.FieldNumber = function(opt_value, opt_min, opt_max, opt_precision,
   opt_value = (opt_value && !isNaN(opt_value)) ? String(opt_value) : '0';
   Blockly.FieldNumber.superClass_.constructor.call(
       this, opt_value, opt_validator);
-  //this.setConstraints(opt_min, opt_max, opt_precision);
 };
 goog.inherits(Blockly.FieldNumber, Blockly.FieldTextInput);
 
@@ -84,7 +83,6 @@ Blockly.FieldNumber.prototype.setConstraints = function(min, max, precision) {
   this.min_ = isNaN(min) ? -Infinity : min;
   max = parseFloat(max);
   this.max_ = isNaN(max) ? Infinity : max;
-  //this.setValue(this.callValidator(this.getValue()));
 };
 
 /**
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -45,12 +45,11 @@ goog.require('Blockly.FieldTextInput');
  */
 Blockly.FieldNumber = function(opt_value, opt_min, opt_max, opt_precision,
     opt_validator) {
-  this.min_ = -Infinity;
-  this.max_ = Infinity;
+  this.setConstraints(opt_min, opt_max, opt_precision);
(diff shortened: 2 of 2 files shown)


RoboErikG:
Hrmm...it looks like the test failures are because Math.min returns NaN if one of the inputs is undefined, and field_number doesn't set min_ and max_ until after the initial value is set, which causes the value to be set as NaN and the test to fail.

To fix, either FieldNumber needs to set min_ to -Infinity and max_ to Infinity before doing anything else or setConstraints needs to be called before the super's constructor is called.

asunwoo98:
Hi,

Sorry for the slow fix. I made the constructor set the constraints before doing any calls as you suggested, but it came out with new errors with 2 similar traces, one copied below. 

`10:21:04.993  ERROR in test_fieldnumber_constructor
this.size_ is undefined
Blockly.Field.prototype.forceRerender@file:///Users/antonysunwoo/blockly/core/field.js:550:3
Blockly.Field.prototype.setText@file:///Users/antonysunwoo/blockly/core/field.js:538:3
Blockly.FieldTextInput.prototype.setText@file:///Users/antonysunwoo/blockly/core/field_textinput.js:141:3
Blockly.Field.prototype.setValue@file:///Users/antonysunwoo/blockly/core/field.js:591:3
Blockly.FieldTextInput.prototype.setValue@file:///Users/antonysunwoo/blockly/core/field_textinput.js:119:5
Blockly.FieldNumber.prototype.setConstraints@file:///Users/antonysunwoo/blockly/core/field_number.js:87:3
Blockly.FieldNumber@file:///Users/antonysunwoo/blockly/core/field_number.js:49:3
test_fieldnumber_constructor@file:///Users/antonysunwoo/blockly/tests/jsunit/field_number_test.js:29:15
10:21:04.993  test_fieldnumber_constructor : FAILED (run individually)`

I fixed this by defining this.size_ in FieldNumber's constructor like it is defined in Field, but maybe this could be refactored as well?

I also noticed that item in setValue are validated in FieldTextInput now as well as Field.js, that may be redundant as well?

Thank you for the help with debugging! Wasn't really sure what I was looking at.

RoboErikG:
I played around with it too and it looks like my second fix suggestion worked without breaking anything (defining min and max to -Infinity/Infinity). I updated the comment to remove the suggestion that doesn't work about 20 minutes after first making it, but it sounds like I should have sent an additional update. 

Could you go with just the min_/max_ fix instead? I'm not sure what other side effects there might be from requesting a rendering update before the field is fully initialized.

asunwoo98:
I implemented that change like you suggested and it works, but this seems very redundant. 

I found the issue with just using setConstraint first. Right now setConstraint revalidates the current value since currently the constraints are set after the value is set. However if that line is taken out from setConstraints, it never calls setValue and therefore never requests a render update before the field is fully initialized. What do you think?

RoboErikG:
Ditto. I'm trying to figure out why Travis is failing but locally the test is fine.

RoboErikG:
huh...after updating my local copy it now fails locally as well.

RoboErikG:
Oh, this PR added a new property dependency in the validator: https://github.com/google/blockly/pull/2384

Alright, time for a less brittle approach. Let's add a private constraintsSet_ property to field_number, make setConstraints set that property to true, and change classValidator to return if (text === null || !constrantsSet_)

asunwoo98:
Wanted to test out my idea

asunwoo98:
Seems to work as well. Should I implement your solution anyway?

RoboErikG:
Oh, that makes sense. I think I misunderstood your proposed solution. Your way is fine, just remove the commented lines. =)

asunwoo98:
Nice. Sorry I find explaining hard sometimes. Thanks for taking a look!

RoboErikG:
Small note, this will cause the current value to not be re-validated if the constraints change, but that seems like a less common issue so I went ahead and merged it.

PR #2405, opened 2019-04-24
Add tests for keyboard handling in the toolbox

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Proposed Changes

Add tests!
### Reason for Changes

Regression tests for navigation, since we'll be making lots of changes in that file.

### Test Coverage
Mocha tests added.

### DIFF
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -32,6 +32,17 @@
       <block type="basic_block"></block>
     </xml>
 
+    <xml id="toolbox-categories" style="display: none">
+      <category name="First">
+        <block type="basic_block">
+          <field name="TEXT">First</field>
+        </block>
+      </category>
+      <category name="Second">
+        <field name="TEXT">Second</field>
+      </category>
+    </xml>
+
     <script>
       mocha.run();
     </script>
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -39,7 +39,9 @@
         </block>
       </category>
       <category name="Second">
-        <field name="TEXT">Second</field>
+        <block type="basic_block">
+          <field name="TEXT">Second</field>
+        </block>
       </category>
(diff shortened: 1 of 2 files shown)


tests/mocha/index.html
```diff
@@ -32,6 +32,20 @@
       <block type="basic_block"></block>
     </xml>
 
+    <xml id="toolbox-simple" style="display: none">
```
alschmiedt: I don't think this is being used anywhere? Is the plan for this to be used for a future use case?

rachel-fenichel: Yes, there was a plan, but I'll remove it now and add it back later if I want it again.

tests/mocha/navigation_test.js
```diff
@@ -32,10 +130,12 @@ suite('Navigation', function() {
       chai.assert.equal(Blockly.Navigation.currentState_,
           Blockly.Navigation.STATE_WS);
     });
-
+    // More tests:
+    // - nested categories
     teardown(function() {
```
RoboErikG: Can you move the teardown to just below setup so it's easier to verify they're paired?

tests/mocha/index.html
```diff
@@ -32,6 +32,17 @@
...
     </xml>
 
+    <xml id="toolbox-categories" style="display: none">
+      <category name="First">
+        <block type="basic_block">
+          <field name="TEXT">First</field>
+        </block>
+      </category>
+      <category name="Second">
+        <field name="TEXT">Second</field>
```
RoboErikG: field isn't a valid tag inside a category. Did you mean to have another basic_block here or for it to be a label?

rachel-fenichel: Meant to have a basic_block.  Fixed.

Issue #1627, opened 2018-02-15
Add Field.prototype.toXml() & .fromXml()


AnmAtAnm:
### Problem statement

Some fields need more than the `value=` attribute of the `<field>` XML when serializing and deserializing data. For example, FieldVariable uses `id=` and `variabletype=`.

### Proposed Solution

 * [ ] Add `.fromXml(/* @type {!Element} */ fieldElement)` and `.toXml(!Element} */ fieldElement)` to the `Field` base class.
 * [ ] Modify Blockly.Xml.fieldToDom_(..) and .domToField_() to call new methods.

RoboErikG:
After internal discussion we'd like to keep writing the value as the text in the element instead of adding a value property to the tag.

PR #2407, opened 2019-04-25
Added fromXml and toXml to Fields

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#1627 

### Proposed Changes

Adds fromXml and toXml methods to fields that Blockly.Xml.domToField_ and fieldToDom_ can call.

Also changes the playground over to the new value-attribute rather than textContent based xml because... I thought it would be nice I guess ¯\\\_(ツ)_/¯

### Reason for Changes

Some fields (e.g. variable fields) need more information to be serialized to xml than just the field name and value. This adds functions that custom fields can override so people can impliment more advanced fields without having to change the core.

### Test Coverage

Added backwards compatibility mocha tests:
FromXMLToXML_Mocha

Also here's a gif of some things serializing and deserializing:
Serialization
That generated this XML:
```
  <variables>
    <variable type="" id="4=m(NP7VfKepgVGbq99y">am i item?</variable>
  </variables>
  <block type="test_fields_angle" id="pcasU-Ep,|#V,:HX]wN~" x="13" y="12">
    <field name="FIELDNAME" value="88"></field>
  </block>
  <block type="test_fields_date" id="wIadJFYd:h^^;oE1=XK}" x="13" y="62">
    <field name="FIELDNAME" value="2019-04-24"></field>
  </block>
  <block type="test_fields_checkbox" id="]b#S7?J!hd9]dLeUm_70" x="13" y="113">
    <field name="CHECKBOX" value="FALSE"></field>
  </block>
  <block type="test_fields_colour" id="~?Dg`]Hj=9Y_qjg?88*Z" x="13" y="163">
    <field name="COLOUR" value="#ffff00"></field>
  </block>
  <block type="test_fields_text_input" id="~(^KZ}SBbC?+Z0-0a/K6" x="13" y="213">
    <field name="TEXT_INPUT" value="am i default?"></field>
  </block>
  <block type="test_fields_variable" id="QEy@*.M-w4:^{(Y),k$%" x="12" y="262">
    <field name="VARIABLE" id="4=m(NP7VfKepgVGbq99y" variablename="am i item?" variabletype=""></field>
  </block>
  <block type="test_fields_label_serializable" id="A5)khRy04vFL~CPD*A`T" x="13" y="313">
    <field name="LABEL" value="BAAA"></field>
  </block>
</xml>
```

Switching the playground over also allowed me to test that the deserialization was working.

Tested on:
 * Desktop Chrome

### Additional Information

It may be better to call the 'variablename' attribute 'value' (even though it's more confusing) because I think having them be different is more prone to errors (i.e. when I was switching the playground over I messed it up a bunch).

Also I'm cool with removing the value  attribute stuff all together, that was just in the original solution proposal so I went with it.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -244,6 +244,30 @@ Blockly.Field.prototype.initModel = function() {
 };
 
 /**
+ * Sets the field's value based on the given XML element. Should only be
+ * called by Blockly.Xml.
+ * @param {!Element} fieldElement The element containing info about the
+ *    field's state.
+ * @package
+ */
+Blockly.Field.prototype.fromXml = function(fieldElement) {
+  this.setValue(fieldElement.textContent);
+};
+
+/**
+ * Serializes this field's value to XML. Should only be called by Blockly.Xml.
+ * @param {!Element} fieldElement The element to populate with info about the
+ *    field's state.
+ * @return {!Element} The element containing info about the field's state.
+ * @package
+ */
+Blockly.Field.prototype.toXml = function(fieldElement) {
+  fieldElement.setAttribute('name', this.name);
+  fieldElement.textContent = this.getValue();
+  return fieldElement;
+};
+
+/**
  * Dispose of all DOM objects belonging to this editable field.
  */
 Blockly.Field.prototype.dispose = function() {
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -126,6 +126,37 @@ Blockly.FieldVariable.prototype.initModel = function() {
   }
 };
 
+Blockly.FieldVariable.prototype.fromXml = function(fieldElement) {
+  var id = fieldElement.getAttribute('id');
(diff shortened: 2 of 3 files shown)


rachel-fenichel:
> Also I'm cool with removing the value attribute stuff all together, that was just in the original solution proposal so I went with it.

I'd like to undo this part.

I see that it's backwards compatible for loading, but changing the save format is still problematic for anyone doing things like diffing XML to see if two workspaces are the same.

BeksOmega:
@rachel-fenichel Do you want to revert the information stored in the variablename attribute to being stored in the textContent as well?

rachel-fenichel:
Yes--I want to keep the XML format entirely the same across this change, and only move the logic for parsing it.

BeksOmega:
All XML format changes have been reverted.

PR #2410, opened 2019-04-25
Fixed Buttons Being Rendered as Labels if They Come After a Label

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Makes it so buttons that come after labels in a flyout are correctly rendered as buttons and not labels.

### Reason for Changes

Before (meaning after I changed flyout parsing to a switch statement, but before this change) the flyout would be rendered like this:
Sep_Bad

### Test Coverage

Sep_Good

Tested on:
* Desktop Chrome

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -480,10 +480,8 @@ Blockly.Flyout.prototype.show = function(xmlList) {
         }
         break;
       case 'LABEL':
-        // Labels behave the same as buttons, but are styled differently.
-        var isLabel = true;
-        // Falls through.
       case 'BUTTON':
+        var isLabel = xml.tagName.toUpperCase() == 'LABEL';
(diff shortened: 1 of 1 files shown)


core/flyout_base.js
```diff
@@ -486,6 +486,7 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       case 'BUTTON':
         var curButton = new Blockly.FlyoutButton(this.workspace_,
```
RoboErikG: It might be a little easier to read if you explicitly set isLabel like this.
case 'LABEL':	
￼case 'BUTTON':
        // Labels behave the same as buttons, but are styled differently.	
        var isLabel = xml.tagName.toUpperCase() == 'LABEL';

BeksOmega: Changed!

PR #2003, opened 2018-08-14
Propagate the visible state when blocks connect

RoboErikG:
This fixes #1967.

In rendered connections when connecting:
- If the superior connection is hidden this hides the newly connected block.
- If the superior connection isn't hidden it makes sure the block is visible.

In rendered connections when disconnecting:
- If the superior connection is hidden, make the disconnected block stack visible.

TODO before review: write tests.

## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

#1967

### Proposed Changes

Adds logic to rendered_connection to hide or unhide blocks when they are connected to a hidden connection.

### Reason for Changes

If you programatically connect a block to a hidden connection the block remains visible and gets placed in a random spot. The block should be hidden when connected and unhidden when disconnected.

### Test Coverage

Tested on:
Desktop Chrome

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2003)

PR #2415, opened 2019-04-26
 Position correctly when inserting before another block (from the flyout)

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

Previously the block stack would move to 100, 100 on the workspace if the new block was being inserted at the start.

### Proposed Changes

Use the same code that we use to solve this problem for insertion markers.
I moved the helper function and changed its signature because it had an unnecessary param before.  That's the first commit.  The second commit uses it.
Remove extra `connect` calls in `insertBlock`--`connect` moves the `orphanBlock` around internally.

### Reason for Changes

Blocks need to stay in the same place as new blocks are connected.

### Test Coverage

Tested in the playground.

### Additional information
I had to add a call to `render` so that the new block's connection positions are correct internally.  That I expected.  I also had to explicitly set the connections to not be hidden, which I didn't expect.  The connections are hidden in `domToBlock` and unhidden in a timeout, for speed when creating large block stacks.  But if the connections are still hidden when I move the block, Erik's new code in #2003 makes the child block disappear.

### DIFF
--- a/core/block_render_svg.js
+++ b/core/block_render_svg.js
@@ -1167,31 +1167,6 @@ Blockly.BlockSvg.prototype.renderStatementInput_ = function(pathObject, row,
 };
 
 /**
- * Position an new block correctly, so that it doesn't move the existing block
- * when connected to it.
- * @param {!Blockly.Block} newBlock The block to position - either the first
- *     block in a dragged stack or an insertion marker.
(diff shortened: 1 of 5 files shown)


rachel-fenichel:
Lint errors in astnode_test, probably from recent rebase.  I'll fix indentation and then merge.

PR #2419, opened 2019-04-29
Add a marker that shows where the user has marked on the workspace

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

Progress on keyboard navigation.

### Proposed Changes

Add a `marker` to the workspace as well as the `cursor`.  A `marker` is a type of `cursorSvg` that cannot be moved with prev/next/in/out.  It can only be moved directly to known locations with `setLocation`.

When the user hits enter to mark a location on the workspace or a connection on a block, the marker renders.  It disappears again when the user inserts a block at that marker.

### Reason for Changes

It's important to know where your insertion location is when you're inserting blocks.

### Test Coverage
Tested in Chrome

### DIFF
--- a/core/navigation.js
+++ b/core/navigation.js
@@ -30,6 +30,12 @@ goog.require('Blockly.ASTNode');
 Blockly.Navigation.cursor_ = null;
 
 /**
+ * The marker that shows where a user has marked while navigating blocks.
+ * @type {!Blockly.CursorSvg}
+ */
+Blockly.Navigation.marker_ = null;
+
+/**
  * The current selected category if the toolbox is open or
  * last selected category if focus is on a different element.
  * @type {goog.ui.tree.BaseNode}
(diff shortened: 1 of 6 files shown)


core/navigation.js
```diff
@@ -297,6 +313,10 @@ Blockly.Navigation.insertFromFlyout = function() {
   //TODO: IF this is null then we need to add to the correct place on the
```
alschmiedt: Can we remove this TODO since I think it is taken care of in insertBlockToWs?

core/navigation.js
```diff
@@ -416,6 +436,45 @@ Blockly.Navigation.insertBlockFromWs = function() {
...
+      markedNode.getType() === Blockly.ASTNode.types.WORKSPACE;
+
+  if (isWsNode && flyout && flyout.isVisible()) {
+    var newBlock = flyout.createBlock(flyoutBlock);
+    // Render to get the sizing right.
+    newBlock.render();
+    var position = markedNode.getWsCoordinate();
+    newBlock.moveTo(position);
+
+    // Move the cursor toe the right place on the inserted block.
```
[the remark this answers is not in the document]
rachel-fenichel: Done

core/navigation.js
rachel-fenichel: Done

PR #2423, opened 2019-04-30
Cleanup for in functionality

alschmiedt:
## The basics

- [ ] I branched from develop

### Proposed Changes
Fixes: 
1. Going from empty workspace to stack 
2. Dealing with blocks that don't have editable field or input with connections
3. Simplifies finding the top ast node for a block 
### Test Coverage
Adds tests in mocha

Tested on:

### DIFF
--- a/core/ast_node.js
+++ b/core/ast_node.js
@@ -434,12 +434,9 @@ Blockly.ASTNode.prototype.navigateBetweenStacks_ = function(forward) {
  * @private
  */
 Blockly.ASTNode.prototype.findTopASTNodeForBlock_ = function(block) {
-  var previousConnection = block.previousConnection;
-  var outputConnection = block.outputConnection;
-  if (previousConnection) {
-    return Blockly.ASTNode.createConnectionNode(previousConnection);
-  } else if (outputConnection) {
-    return Blockly.ASTNode.createConnectionNode(outputConnection);
+  var topConnection = block.previousConnection || block.outputConnection;
+  if (topConnection) {
+    return Blockly.ASTNode.createConnectionNode(topConnection);
   } else {
     return Blockly.ASTNode.createBlockNode(block);
   }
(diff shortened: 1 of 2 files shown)


tests/mocha/astnode_test.js
```diff
@@ -316,51 +324,141 @@ suite('ASTNode', function() {
...
         assertEquals(inNode.getLocation(), outputConnection);
       });
-      test('blockToInput', function() {
+      test('fromInputToNull', function() {
+        var input = this.blocks.statementInput2.inputList[0];
+        var node = Blockly.ASTNode.createInputNode(input);
+        var inNode = node.in();
+        assertEquals(inNode, null);
+      });
+      test('fromInputToNext', function() {
```
rachel-fenichel: Is this actually fromInputToPrevious?  It looks like it's going from a statement input to the previous connection on the first block in the substack.

alschmiedt: Yeah it should be fromInputToPrevious. Just changed the name

core/ast_node.js
```diff
@@ -474,6 +471,31 @@ Blockly.ASTNode.prototype.getOutAstNodeForBlock_ = function(block) {
   }
 };
 
+/**
+ * Find the first editable field or input with a connection on a given block.
+ * @param {Blockly.BlockSvg} block The source block of the current location.
```
NeilFraser: block can't be null (block.inputList would fail), so this should be {!Blockly.BlockSvg}

core/ast_node.js
```diff
@@ -474,6 +471,31 @@ Blockly.ASTNode.prototype.getOutAstNodeForBlock_ = function(block) {
...
+ * @return {Blockly.ASTNode} An ast node pointing to the first field or input.
+ * Null if there are no editable fields or inputs with connections on the block.
+ * @private
+ */
+Blockly.ASTNode.prototype.findFirstFieldOrInput_ = function(block) {
+  var inputs = block.inputList;
+  for (var i = 0; i < inputs.length; i++) {
+    var input = inputs[i];
```
NeilFraser: Generally we use this form:
`  for (var i = 0, input; input = inputs[i]; i++) {`
Here and in the j loop below.

core/ast_node.js
alschmiedt: Fixed

core/ast_node.js
alschmiedt: Fixed

core/ast_node.js
```diff
@@ -250,7 +250,7 @@ Blockly.ASTNode.prototype.findNextEditableField_ = function(location,
   var fieldRow = parentInput.fieldRow;
   var fieldIdx = fieldRow.indexOf(location);
   var startIdx = opt_first ? 0 : fieldIdx + 1;
-  for (var i = startIdx; i < fieldRow.length; i++) {
+  for (var i = startIdx, field; field = fieldRow[i]; i++) {
     var field = fieldRow[i];
```
NeilFraser: This line is redundant.  :)

alschmiedt: Shoot missed that. Will fix in my next PR.

PR #1495, opened 2017-12-07
Act on variables by ID, not name + type

rachel-fenichel:
FieldVariable now always acts on a VariableModel.  

Tips for reviewing:
Changes in workspace.js are mostly consolidation of code/removal of lookup by name.
FieldVariable's initialization changed in a major way.  That's a good place to start.
Variables.js has a lot of extra utility functions now.

I changed many tests to use the new API.  I also deleted some tests for renaming variables that don't exist, which should no longer be necessary.

PR #2425, opened 2019-05-01
Removed Explicitly Setting Variable Type to Two Single Quotes

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Removes explicitly setting the variable type to two single quotes.

### Reason for Changes

Before the toXml/fromXml pr went in there was a check in domToFieldVariable_ to change an string containing two single quotes to an empty string, with a comment that maybe this wasn't necessary. I believe the only purpose of that chunk of code was to undo the thing that I am removing now (where the string containing two double quotes is created).

This all seems a bit weird though so I'm not sure if there was a secret purpose behind it. I did check the git blame and found the [original PR](https://github.com/google/blockly/pull/1495/files), but I didn't see any information about why it was set up this way.

### Test Coverage

I discovered this because field_variable's fromXml was throwing an errow when it was trying to deserialize the dynamic category XML, so I added tests for both the untyped and typed variable dynamic categories.

DynamicCategoryTests

I also manually tested creating variables through the flyout buttons, and that is working.
And I checked the test blocks' variable field block, and it is working. 

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -466,9 +466,6 @@ Blockly.Variables.generateVariableFieldXmlString = function(variableModel) {
   // The variable name may be user input, so it may contain characters that
   // need to be escaped to create valid XML.
   var typeString = variableModel.type;
-  if (typeString == '') {
-    typeString = '\'\'';
-  }
   var text = '<field name="VAR" id="' + variableModel.getId() +
       '" variabletype="' + goog.string.htmlEscape(typeString) +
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
This looks right.  The actual reason for that code is completely lost to me, but as long as serialization continues to work (and it looks like it does) I agree with getting rid of this.

Issue #1010, opened 2017-03-29
Split Field.init into Field.initView and Field.initModel


CoryDCode:
Currently, fields have one init function, which is called through the SVG flow. Variables currently do model logic in their init (adding variables to the list of variables, for reverse-compatibility.) That means headless versions of blockly don't correctly update their list of variables. I've added an initModel function to Field, so this can be solved manually, but we need to figure out where to call initModel so it gets called correctly in all flows.

BeksOmega:
@CoryDCode Could you provide steps to reproduce the old bad headless initialization (possibly w/ xml)? I'm doing a bit of digging into the field lifecycle and I want to make sure there's no regression. If not that's cool but I wanted to reach out!

Thanks!
Beka

rachel-fenichel:
@BeksOmega Done?

BeksOmega:
Code is done, still possibly needs docs.

rachel-fenichel:
Imo we're in good shape on docs, with the docs per field = the ones for creating a custom field.

PR #2427, opened 2019-05-01
Refactored field.init into field.initView and field.initModel

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#1010

### Proposed Changes

Moves init logic into an initView function. The init function now calls initView and initModel.

I also had to change the emoji testblocks category label because apparently IE just couldn't handle the beauty of it.

### Reason for Changes

Firstly makes the lifecycle more clear. Secondly the abstract init() function now creates the field group, so fields that want to do custom dom structures (e.g. not include the border rect or text element) don't need to remember to create a field group named 'fieldGroup_'.

### Test Coverage

Looked through test blocks and assured that everything was rendering correctly. RTL and LTR
Also ran mocha tests locally.

Tested on:
* Desktop Chrome
* Desktop Firefox

 * Windows Edge

### Additional Information

[Here](https://i.imgur.com/izzAxcV.jpg)'s a picture of how the flow works now.

If the field does not have a concrete init function (which for any future fields it should not!) any callers will hit the abstract init imediately and then the it'll just work how you'd expect. The sub-most class gets it's concrete initView called first and then that flows back up to the abstract class etc.

### DIFF
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -77,14 +77,12 @@ Blockly.FieldCheckbox.CHECK_CHAR = '\u2713';
 Blockly.FieldCheckbox.prototype.CURSOR = 'default';
 
 /**
- * Install this checkbox on a block.
+ * Create the block UI for this checkbox.
+ * @package
  */
-Blockly.FieldCheckbox.prototype.init = function() {
-  if (this.fieldGroup_) {
-    // Checkbox has already been initialized once.
-    return;
-  }
-  Blockly.FieldCheckbox.superClass_.init.call(this);
+Blockly.FieldCheckbox.prototype.initView = function() {
+  Blockly.FieldCheckbox.superClass_.initView.call(this);
+
   // The checkbox doesn't use the inherited text element.
   // Instead it uses a custom checkmark element that is either visible or not.
   this.checkElement_ = Blockly.utils.createSvgElement('text',
--- a/core/field.js
+++ b/core/field.js
@@ -202,15 +202,26 @@ Blockly.Field.prototype.setSourceBlock = function(block) {
 };
 
 /**
- * Install this field on a block.
+ * Initialize everything to render this field. Override
+ * methods initModel and initView rather than this method.
+ * @package
  */
 Blockly.Field.prototype.init = function() {
   if (this.fieldGroup_) {
     // Field has already been initialized once.
     return;
   }
-  // Build the DOM.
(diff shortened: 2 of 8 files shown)


BeksOmega:
Note: Just realized that there needs to be a follow up to this where the fields are audited to make sure they're initializing things correctly. FieldAngle is doing some pretty wack stuff, other fields may be as well.

core/field_checkbox.js
```diff
@@ -77,14 +77,22 @@ Blockly.FieldCheckbox.CHECK_CHAR = '\u2713';
...
- * Install this checkbox on a block.
+ * Create the block UI for this checkbox.
+ * @deprecated May 2019 Use FieldCheckbox.initView.
  */
 Blockly.FieldCheckbox.prototype.init = function() {
-  if (this.fieldGroup_) {
-    // Checkbox has already been initialized once.
-    return;
-  }
+  // Logging a warning about calling a deprecated method here would be
```
RoboErikG: Can you remove this declaration of init? Since it inherits from Field it should just call the super method instead if it's not declared.

Ditto for everything else that inherits from Field that you're updating to just call the super.

BeksOmega:
Derp fixed.

I also added package tags to all of the methods. initView could probably be protected (and renamed to initView_) but I wanted initView and initModel to be matchy-matchy. Thoughts?

RoboErikG:
@rachel-fenichel this looks good to me, but can you take a quick look in case I missed anything?

core/field_checkbox.js
```diff
@@ -77,14 +77,12 @@ Blockly.FieldCheckbox.CHECK_CHAR = '\u2713';
...
+ * @package
  */
-Blockly.FieldCheckbox.prototype.init = function() {
-  if (this.fieldGroup_) {
-    // Checkbox has already been initialized once.
-    return;
-  }
-  Blockly.FieldCheckbox.superClass_.init.call(this);
+Blockly.FieldCheckbox.prototype.initView = function() {
+  Blockly.FieldCheckbox.superClass_.initView.call(this);
```
rachel-fenichel: There's no longer anything to stop initView from happening twice.  But I think that's fine, because initializing the view twice would mean duplicate svg elements and it would be pretty visible during development of a custom field.

rachel-fenichel:
Package is fine for initView.  I agree that matching them is better.

PR #2434, opened 2019-05-02
Run mocha tests in the browser on travis

rachel-fenichel:
PR 2 of 2 to pull this stuff over from develop, where we've already reviewed it.

This should fail on travis with a message that two mocha tests are failing.  

There are 3 tests failing on keyboard_nav.  One is the test in block_test, which I've fixed here.  The other two are in keyboard-nav related work and Abby will have to take a look at them.


rachel-fenichel:
As expected, two tests failed.

Issue #2435, opened 2019-05-02
Q: How do you want text input validation to look?


BeksOmega:
### Problem statement

I'm trying to move the fields over to the new validation discussed [here](https://github.com/google/blockly/issues/2169#issuecomment-487407878) but the FieldInput is giving me trouble because of [this](https://github.com/google/blockly/blob/9c4f9d6c1693c3c25160f008d3eaa2539127c9e8/core/field_textinput.js#L113):

```
if (newValue !== null) { // No change if null.
    if (this.sourceBlock_) {
      var validated = this.callValidator(newValue);
      // If the new value is invalid, validation returns null.
      // In this case we still want to display the illegal result.
      if (validated !== null) {
        newValue = validated;
      }
    }
    Blockly.Field.prototype.setValue.call(this, newValue);
  }
```

I'm a bit unsure of what illegal results we want to display, where, and also when the input div should turn red.

Here's an example of a validator that always returns null:
![Validator_Null](https://user-images.githubusercontent.com/25440652/57113398-38604300-6cf9-11e9-8f2a-98039842e396.gif)
This makes sense, the value is invalid so it turns red, but it still keeps expanding the htmlInput so you can see what you're typing.

Here's an example of a validator that removes any a's:
![Validator_A](https://user-images.githubusercontent.com/25440652/57113471-98ef8000-6cf9-11e9-8dd4-9bbe7d47a120.gif)
As you can see it does neither of the above things.

I'm not sure if it should:
A) Remain white, but remove the a's from the htmlInput as well as the field (i.e. even if you typed an 'a' it wouldn't appear on screen).
B) Turn red, and leave the a's in both the htmlInput and the field until the user stops editting (replicating how the null validator works).

Or maybe you want both of them to work completely differently! I'm open to anything since I'm tearing out all of the validation stuff anyway.

### Steps to Reproduce

1. Define the following blocks in the test_blocks file:
```
Blockly.Blocks['test_text_null_validator'] = {
  init: function() {
    this.appendDummyInput()
        .appendField("null validator")
        .appendField(new Blockly.FieldTextInput("default", this.validate), "NAME");
    this.setColour(230);
  },
  validate: function(newValue) {
    return null;
  }
};

Blockly.Blocks['test_text_A_validator'] = {
  init: function() {
    this.appendDummyInput()
      .appendField("\'a\' validator")
      .appendField(new Blockly.FieldTextInput("default", this.validate), "NAME");
    this.setColour(230);
  },
[...]

Issue #2380, opened 2019-04-13
Degree symbol should be right of number in RTL


NeilFraser:
Blockly currently uses `90°` for LTR and `°90` for RTL.  My belief is that RTL should not be flipped, and that `90°` should be used universally.

This evidence for this claim is based on:
* The Hebrew wikipedia page on angles: https://yi.wikipedia.org/wiki/%D7%92%D7%A8%D7%90%D7%93_(%D7%95%D7%95%D7%99%D7%A0%D7%A7%D7%9C)
* Israeli weather site (note weather degrees are not angle degrees): https://www.israelweather.co.il/forecast/week_15.html
* Arabic weather site (note that the large panels use `90°` but the paragraph text uses `°90`): https://www.arabiaweather.com/%D8%AA%D9%88%D9%82%D8%B9%D8%A7%D8%AA-%D8%A7%D9%84%D8%B7%D9%82%D8%B3/%D8%A7%D9%84%D8%B1%D9%8A%D8%A7%D8%B6/344/sa

This is not extremely strong evidence.  But on balance, it appears that our degree symbol is on the wrong side.

chrisgarrity:
Unfortunately it's not quite that simple. Hebrew mathematical notation is LTR, but in Arabic script it's mixed. This article on [Arabic mathematical notation](https://www.w3.org/TR/2006/NOTE-arabic-math-20060131/#Comparison) is helpful in showing the different Arabic usage.

However, putting the degree symbol on the right for RTL in Blockly is likely to be correct more often than not.

Issue #1508, opened 2017-12-12
Renaming a procedure does not group events


RoboErikG:
### Problem statement
It looks like we lost a setGroup(true) somewhere in procedure renaming. This is causing each keystroke and each rename of a caller to be a separate undo action.

If you rename a procedure that has callers you will end up with an extra undo op for every block that was renamed. When you use ctrl+z to undo you then have to do an extra undo for each block. They are all renamed on the first undo, so the extra undos are just no-ops.

### Expected Behavior
One undo group for all the renames.

### Actual Behavior
The events for a procedure name change have no group.

### Steps to Reproduce

1. Create a procedure
2. Create two callers of the procedure
3. Change the name of the procedure.
4. Undo the name change.

### Stack Traces

```
Replace with error stack trace.
```

### Operating System and Browser

Desktop Chrome

BeksOmega:
For history's sake, this was actually reopened by #2441 ([apparently](https://github.com/google/blockly/pull/2441#issuecomment-489369438) I thought this was the intended behavior). And then fixed by #4952

PR #2441, opened 2019-05-04
Refactored Field Validation.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2169 

### Proposed Changes

Changes field validation to work like discussed [here](https://github.com/google/blockly/issues/2169#issuecomment-487407878).

### Reason for Changes

ATM fields only expect one value, and they expect it to be text, this is not true of all fields.
Also field validation is pretty all over the place, this should unify it to make it easier to debug.

### Test Coverage

For now N/A

Tested on:

### Additional Information

This isn't finished yet, but it's going to be quite a whopper of a PR, so I wanted to put it up so it could be reviewed in stages.


BeksOmega:
The first leg!

### Resolves
#2435

### Proposed Changes
Refactored FieldTextInput to conform to the new structure.

### Test Coverage

Added text validator test blocks to the playground.

* The "null validator" block always returns null (invalid) for all inputs
* The "remove 'a' validator" block removes any 'a' characters from the input value.
* The "'b' -> null validator" block returns null (invalid) if it detects a 'b' character in the input.

Validation_TextInput_Null
Validation_TextInput_A
Validation_TextInput_B

(To do the logging change the logger function on the playground to look like the below, and enable logging events)
```
function logger(e) {
  if (e.type == Blockly.Events.BLOCK_CHANGE && e.element == 'field') {
    console.log('new value: ', e.newValue);
  }
}
```

Tested setting value programmatically:
* Value = "ab"
Validator_TextInput_ab

* Value = "a"
Validator_TextInput_a

* Value = "b"
Validator_TextInput_b

Tested undo/redo and events (see #1508):
Validation_TextInput_Undo

There is one ctrl-z for each change in the workspace.

### Additional Information

I know all of the new text input stuff works, but I'm not sure if it's readable enough. Please tell me if there are places that are confusing!

Also @rachel-fenichel [here ](https://github.com/google/blockly/issues/2169#issuecomment-488471207)you mentioned

> tests that are the same before and after, to verify that existing implementations continue to work.

Do you mean like a test-field that overrides setValue or setText, and checks if it still gets called correctly? Or did you have something else you were thinking of?

BeksOmega:
Second leg!

### Proposed changes

Refactored FieldAngle to conform to the new structure.

### Test Coverage

Added angle validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force mult of 30" block returns the closest multiple of 30 to the input.
* The "not mult of 30 -> null" block returns null (invalid) if the input is not a multiple of thirty.
Validation_Angle_Null
Validation_Angle_Force30
Validation_Angle_Null30

Tested setting value programmatically:
* Value = 15:
Validator_Angle_15

### Additional Information

FieldAngle's value_ property now contains a number instead of text. This should act the same as before with regard to generators. It should also act the same with regard to validators because devs should already be converting the text into a number inside their validator. But now future peoples won't have to do that!

BeksOmega:
Third leg!

### Proposed Changes

Refactored FieldCheckbox to conform to the new structure.

### Test Coverage

Added checkbox validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force match" block has a validator on the end-most checkbox, which makes it match the start-most checkbox.
* The "not match -> null" block  has a validator on the end-most checkbox, which returns null (invalid) if the input does not match the start-most checkbox.
Validation_Checkbox_Null
Validation_Checkbox_Match
Validation_Checkbox_MatchNull

Tested setting value programatically:
* Original State was True then False
Validator_Checkbox_TF_Before

* Set to TRUE
Validator_Checkbox_TF_True

* Set to FALSE
Validator_Checkbox_TF_False

* Original State was False then True
Validator_Checkbox_FT_Before

* Set to True
Validator_Checkbox_FT_True

* Set to False
Validator_Checkbox_FT_False

### Additional Information

It may be worth considering deprecating the current checkbox in favor of one that contains an actual true/false value, probably not though.

BeksOmega:
Colour fields.

###Proposed Changes
Refactored FieldCheckbox to conform to the new structure.

###Test Coverage
Added checkbox validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force red" block changes the value to always have a full red value.
* The "not red -> null" block returns null (invalid) if the colour does not have a full red value.
Validation_Colour_Null
Validation_Colour_ForceRed
Validation_Colour_RedNull

Tested setting value programmatically:
* Set Value #ffffff
Validator_Color_FFFFFF

* Set Value #00ffff
Validator_Color_00FFFF

### Additional Information
N/A

BeksOmega:
Date fields.

### Proposed Changes
Refactored FieldDate to conform to the new structure.

### Test Coverage
Added date validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force day 20s" block changes the value to always have a day in the 20s.
* The "not 20s-> null" block returns null (invalid) if the date's day value is not in the 20s.
Validation_Date_Null
Validation_Date_Force20s
Validation_Date_20sNull

Tested setting the value programmatically:
Set value to '2020-02-15':
Validator_Date_2020-02-15

BeksOmega:
Date fields.

### Proposed Changes
Refactored FieldDate to conform to the new structure.

### Test Coverage
Added date validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force day 20s" block changes the value to always have a day in the 20s.
* The "not 20s-> null" block returns null (invalid) if the date's day value is not in the 20s.
Validation_Date_Null
Validation_Date_Force20s
Validation_Date_20sNull

Tested setting the value programmatically:
Set value to '2020-02-15':
Validator_Date_2020-02-15

BeksOmega:
Dropdown fields.

### Proposed Changes
Refactored FieldDropdown to conform to the new structure.

### Test Coverage
Added dropdown validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force1s" block changes the language-neutral value to always start with a one.
* The "not 1s-> null" block returns null (invalid) if the value does not start with a one.
Validation_Dropdown_Null
Validation_Dropdown_Force1s
Validation_Dropdown_1sNull

Tested setting value programmaticly:
* Value 1B
Validator_Dropdown_1B

* Value 2B
Validator_Dropdown_2B

* Value 3B
Validator_Dropdown_3B
Correctly got log message:
`Cannot set the dropdown's value to an unavailable option. Block type: test_validators_dropdown_null, Field name: INPUT`

rachel-fenichel:
Thanks for splitting this up by field type.  It's a lot of code!

This stuff looks like it should be testable in Mocha--you would be able to set up a field, call setValue with a variety of values, and then validate the resulting value and text.  Since this should be a no-op in terms of existing field behaviour, you would be able to run the tests before and after the change and get the same results.

BeksOmega:
> This stuff looks like it should be testable in Mocha--you would be able to set up a field, call setValue with a variety of values, and then validate the resulting value and text. Since this should be a no-op in terms of existing field behaviour, you would be able to run the tests before and after the change and get the same results.

Sweet! I think set up tests in a separate PR and then rebase on top of it before this gets merged, just to keep things a little more organized.

BeksOmega:
Image fields.

### Proposed Changes
Refactored FieldImage to conform to the new structure.

### Test Coverage
Note: Image fields don't take in validators, so we don't need to test those.

I added a default image field block to the playground, with a button to change the image src & alt text. I also added a "toggle collapsed" button so we can see how the blocks look when collapsed.
Validation_Image

### Additional Information

There's an issue with collapsing because setVisible is changing the field's private size_.width property to 0 when really that should just be handled inside getSize. So some fields like colour, image, etc have to override render to reset the field size. I'll come back and clean that up in a different PR.

Also the checkbox isn't handling collapsing properly, I'll have to check on that.

core/field.js
```diff
@@ -633,39 +638,114 @@ Blockly.Field.prototype.forceRerender = function() {
...
+  doLogging && console.log(this.value_);
+};
+
+/**
+ * A generic value possessed by the field.
+ * Should generally be non-null, only null when the field is created.
+ * @type {*}
+ * @protected
+ */
+Blockly.Field.prototype.value_ = null;
```
rachel-fenichel: Move this to the top of the file so it's not lost in here with the functions.

core/field.js
```diff
@@ -633,39 +638,114 @@ Blockly.Field.prototype.forceRerender = function() {
...
+
+/**
+ * Used to validate a value. Returns input by default. Can be overridden by
+ * subclasses, see FieldDropdown.
+ * @param {*} newValue The value to be validated.
+ * @return {*} The validated value, same as input by default.
```
rachel-fenichel: As I recall, the possible return values are null, undefined, or the validated value, where null means it's invalid and undefined means...something.  Is that accurate?

Either way, please update the comment to say what null means.

core/field.js
```diff
@@ -441,15 +446,24 @@ Blockly.Field.prototype.updateColour = function() {
...
+ * done here, and should be triggered by getSize().
  * @protected
  */
 Blockly.Field.prototype.render_ = function() {
   // Replace the text.
-  this.textElement_.textContent = this.getDisplayText_();
+  try {
+    this.textElement_.childNodes[0].nodeValue = this.getDisplayText_();
+  } catch (error) {
+    if (this.textElement_.childNodes.length == 0) {
```
rachel-fenichel: Why do this as a try-catch instead of checking the length of `childNodes` before trying to make the change?

core/field.js
```diff
@@ -633,39 +648,109 @@ Blockly.Field.prototype.forceRerender = function() {
...
+      this.forceRerender();
+    }
     return;
   }
-  // Validate input.
-  var validated = this.callValidator(newValue);
-  if (validated !== null) {
-    newValue = validated;
+
+  var userValidator = this.getValidator();
```
rachel-fenichel: Nit: maybe "localValidator" or "fieldValidator".  "userValidator" feels misleading.

core/field_angle.js
```diff
@@ -35,21 +35,17 @@ goog.require('goog.userAgent');
...
  * @param {Function=} opt_validator An optional function that is called
  *     to validate any constraints on what the user entered.  Takes the new
- *     text as an argument and returns the accepted text or null to abort
+ *     angle as an argument and returns the accepted angle or null to abort
  *     the change.
  * @extends {Blockly.FieldTextInput}
  * @constructor
  */
 Blockly.FieldAngle = function(opt_value, opt_validator) {
-  // Add degree symbol: '360°' (LTR) or '°360' (RTL)
```
rachel-fenichel: Making sure I understand:
Removing the degree symbol stuff here isn't specifically because of the change to the validator--it's because it was previously dynamic (depended on RTL) and is now always on the same side of the number, so it can just be an additional textNode that we don't change.  

Is that right?

core/field_angle.js
```diff
@@ -121,18 +117,23 @@ Blockly.FieldAngle.WRAP = 360;
...
 
 /**
- * Adds degree symbol and recalculates width.
- * Saves the computed width in a property.
+ * Creates the block UI for this field.
+ * @package
+ */
+Blockly.FieldAngle.prototype.init = function() {
+  Blockly.FieldAngle.superClass_.init.call(this);
+  var symbol = Blockly.utils.createSvgElement('tspan', {}, null);
```
rachel-fenichel: Add a comment: // Add the degree symbol to the left of the number, even in RTL (issue #2380)

core/field_angle.js
```diff
@@ -344,7 +335,14 @@ Blockly.FieldAngle.prototype.classValidator = function(text) {
...
   }
-  return String(n);
+  return n;
+};
+
+Blockly.FieldAngle.prototype.doValueUpdate_ = function(newValue) {
+  Blockly.FieldAngle.superClass_.doValueUpdate_.call(this, newValue);
+  // Always render the input angle.
+  this.text_ = String(newValue);
+  this.forceRerender();
```
rachel-fenichel: You call `forceRerender` in `Field.setValue`, just after `doValueUpdate_`.  Why the duplication?

core/field_checkbox.js
```diff
@@ -92,49 +93,72 @@ Blockly.FieldCheckbox.prototype.init = function() {
...
-    }
-    this.state_ = newState;
-    if (this.checkElement_) {
-      this.checkElement_.style.display = newState ? 'block' : 'none';
-    }
+Blockly.FieldCheckbox.prototype.doValueUpdate_ = function(newValue) {
+  this.value_ = this.convertValueToBool_(newValue);
+  // Update visual.
+  if (this.checkElement_) {
+    this.checkElement_.style.display = this.value_ ? 'block' : 'none';
```
rachel-fenichel: This UI update looks like it should happen in the rerender, with `doValueUpdate` just storing the new value and marking the field as dirty.

core/field_checkbox.js
```diff
@@ -92,49 +93,72 @@ Blockly.FieldCheckbox.prototype.init = function() {
...
   }
 };
 
 /**
- * Toggle the state of the checkbox.
+ * Get the value of this field, either 'TRUE' or 'FALSE'.
+ * @return {string} The value of this field.
+ */
+Blockly.FieldCheckbox.prototype.getValue = function() {
```
rachel-fenichel: While you're here, add a `getValueBoolean` method on `FieldCheckbox` that just returns the actual value, so that developers don't *have* to go back and forth checking string equality.

core/field.js
BeksOmega: Null means its invalid, which by default means the field reverts to its previous value.

Undefined doesn't mean anything at the moment, but most fields will revert to a default value (default meaning the value they would have if they had been constructed with no arguements, not any value they /were/ constructed with) when passed a value they can't parse (such as undefined, NaN, 'bad', etc).

This seems like a bug, since not all fields work this way. What I want to go back through and do is have fields' class validators validate anything they can't parse (undefined, NaN, 'bad', etc) to null (invalid). Any default values would be handled in the constructor.

But that's a bit besides the point, if we want to add more special cases besides 'invalid' I would prefer to move to using constants rather than keywords, so it's more extendable in the future.

core/field_checkbox.js
rachel-fenichel: Actually, I'm seeing this as a pattern across other fields as well.  More work is happening in doValueUpdate than I expect, and less in render, which often just sets the width.

core/field_dropdown.js
```diff
@@ -396,40 +389,49 @@ Blockly.FieldDropdown.prototype.getValue = function() {
...
+      isValueValid = true;
+      break;
+    }
   }
-  if (this.sourceBlock_ && Blockly.Events.isEnabled()) {
-    Blockly.Events.fire(new Blockly.Events.BlockChange(
-        this.sourceBlock_, 'field', this.name, this.value_, newValue));
+  if (!isValueValid) {
+    console.warn('Cannot set the dropdown\'s value to an unavailable option.' +
+      ' Block type: ' + this.sourceBlock_.type + ', Field name: ' + this.name);
```
rachel-fenichel: Add the value to the warning message.

core/field_textinput.js
```diff
@@ -234,16 +260,14 @@ Blockly.FieldTextInput.prototype.bindEvents_ = function(htmlInput) {
...
+  // Bind to keyup -- resize after every keystroke.
   htmlInput.onKeyUpWrapper_ =
       Blockly.bindEventWithChecks_(
           htmlInput, 'keyup', this, this.onHtmlInputChange_);
   // Bind to keyPress -- repeatedly resize when holding down a key.
   htmlInput.onKeyPressWrapper_ =
       Blockly.bindEventWithChecks_(
           htmlInput, 'keypress', this, this.onHtmlInputChange_);
-  htmlInput.onWorkspaceChangeWrapper_ = this.resizeEditor_.bind(this);
-  this.workspace_.addChangeListener(htmlInput.onWorkspaceChangeWrapper_);
```
rachel-fenichel: Why did you remove this?  I think you need it to make sure that the inline editor floats to the right place if you resize the window or workspace while you're editing this text.

core/field_textinput.js
```diff
@@ -234,16 +260,14 @@ Blockly.FieldTextInput.prototype.bindEvents_ = function(htmlInput) {
   htmlInput.onKeyDownWrapper_ =
       Blockly.bindEventWithChecks_(
           htmlInput, 'keydown', this, this.onHtmlInputKeyDown_);
-  // Bind to keyup -- trap Enter; resize after every keystroke.
```
rachel-fenichel: What about Enter?

core/field_textinput.js
```diff
@@ -371,12 +366,30 @@ Blockly.FieldTextInput.prototype.widgetDispose_ = function() {
...
+      // may not be if the input text is different than the validated text.
+      // We should fix that.
+      thisField.text_ = String(thisField.value_);
+      thisField.isTextValid_ = true;
+      thisField.forceRerender();
+    }
+    // Otherwise don't rerender.
+
+    // Call onFinishEditing
+    // TODO: Get rid of this or make it less of a hack.
```
rachel-fenichel: FWIW the case where this is really important is when editing the name of a parameter in a function.  That has code to create a variable with the given name at every keypress, and then it deletes intermediate variables when the user is done editing.

rachel-fenichel:
I just added more comments, and I've covered all of the files once so far.
Overall it looks good.  I'd like to straighten out when rendering actually happens, as I mentioned in a few comments.

core/field_checkbox.js
BeksOmega: Doing stuff inside doValueUpdate_ or doValueInvalid is actually a little bit of an optimization. Setting isDirty to true (and doing work inside render_) means the whole block has to rerender, which is unecessary if the fields size doesn't change.

The reason for setting the width inside render is because setVisible is broken at the moment. It's setting the field's size_.width to zero when that should really be handled inside getSize.

core/field_angle.js
BeksOmega: Not sure why this ended up down here, it should be inside the onMouseMove function. 

Fixed.

core/field_angle.js
BeksOmega: Correct! I had to switch up the rendering because of how render_ changed to handle htmlInput being red or not.

core/field.js
BeksOmega: So this happened because of how the angle field's degree symbol is static rather than reappended every time render is called. It used to be reappended everytime because when the textElement_'s textContent was set it would get kicked. So instead of setting the textContent it just sets the nodeValue of the textElement_'s first text node. Buuut if the text element doesn't have a first text node everything breaks. This can happen if a developer writes their initView function incorrectly.

TL:DR A series of changes meant I had to add this so developers know if they derp something up.

Really this needs a better solution though, how would you feel about some sort of textGroup_ that the textElement_ as well as things like degree symbols could be children of? Then render could go back to how it was before, and this could be removed.

core/field_textinput.js
BeksOmega: If that's what it was supposed to do it doesn't do that, the workspace doesn't support resize/move/zoom events yet, plus window resize calls hideChaff. As far as I can tell it's just code accretion.

I'll revert it and create an issue to further investigate.

core/field_textinput.js
BeksOmega: As far as I can tell onHtmlInputChange_ doesn't do anything with enter. I think this was just a typo so I removed it.

core/field_dropdown.js
BeksOmega: Fixed

core/field_checkbox.js

core/field_angle.js
BeksOmega: Fixed

core/field.js
BeksOmega: Fixed, I picked localValidator.

core/field.js
BeksOmega: Fixed

BeksOmega:
Number Fields!

### Proposed changes
Refactored FieldNumber to conform to the new structure.

### Test Coverage
Added number validator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force mult of 10" block returns the closest multiple of 30 to the input.
* The "not mult of 10-> null" block returns null (invalid) if the input is not a multiple of ten.
Validation_Number_Null
Validation_Number_Force10
Validation_Number_10Null

Tested setting value programmatically
* Value = 10
Validator_Number_10

* Value = 97
Validator_Number_97

Also the precision is still working:
Validator_Number_Precision

### Additional Information

Field number now contains a number value instead of text.

The label fields don't override setValue so those don't need to be refactored. This means I'm just down to variable fields. So my remaining tasks for this PR are:

- [ ]  Refactor Variable Fields
- [ ]  Fix Review Comments
- [ ]  Rebase onto Tests
- [ ]  Pass Tests (Eventually)
- [ ]  Clean up Logging, Formatting, etc
- [ ]  Celebrate!

BeksOmega:
Variable fields.

### Proposed Changes
Refactored FieldVariable to conform to the new structure.

### Test Coverage
Added variablevalidator blocks to the playground.

* The "always null" block always returns null (invalid) for all inputs.
* The "force1s" block changes the variable ID to always start with a one.
* The "not 1s-> null" block returns null (invalid) if the variable ID does not start with a one
Validation_Variable_Null
Validation_Variable_Force1s
Validation_Variable_1sNull

Tested setting value programmaticly:
* Value 1B
Validator_Variable_1B

* Value 2B
Validator_Variable_2B

* Value 3B
Validator_Variable_3B
& Throws a warning.

BeksOmega:
Besides the generator tests failing I think this should be good for another look @rachel-fenichel.

I'm having some trouble figuring out why the generator tests are failing, since they won't deserialize completely inside firefox, but I might take another look at it later.

core/field_checkbox.js
```diff
@@ -91,49 +94,98 @@ Blockly.FieldCheckbox.prototype.initView = function() {
...
 /**
- * Return 'TRUE' if the checkbox is checked, 'FALSE' otherwise.
- * @return {string} Current state.
+ * Checkboxes have a constant width.
+ * @private
  */
-Blockly.FieldCheckbox.prototype.getValue = function() {
-  return String(this.state_).toUpperCase();
+Blockly.FieldCheckbox.prototype.render_ = function() {
+  this.size_.width = 5;
```
rachel-fenichel: Make a constant (Blockly.FieldCheckbox.WIDTH) to use here and in the constructor.

core/field_checkbox.js
rachel-fenichel: Interesting.  Sounds like that's a separate issue, but one worth looking into (avoiding a rerender if the field's size didn't change).  Can you create an issue for it and note which fields will likely need investigation?

core/field_dropdown.js
```diff
@@ -386,48 +371,58 @@ Blockly.FieldDropdown.prototype.getOptions = function() {
 };
 
 /**
- * Get the language-neutral value from this dropdown menu.
- * @return {string} Current text.
+ * Check that the new value is an available option.
```
rachel-fenichel: Remove extra comment.

core/field_image.js
```diff
@@ -166,28 +167,28 @@ Blockly.FieldImage.prototype.setTooltip = function(newTip) {
 };
 
 /**
- * Get the source URL of this image.
- * @return {string} Current text.
- * @override
+ * Ensure that the input value is a string.
```
rachel-fenichel: Ensure that the input value _(the source URL)_ is a string.

core/field_number.js
```diff
@@ -31,22 +31,23 @@ goog.require('Blockly.FieldTextInput');
...
+ *    number, or null to abort the change.
  * @constructor
  */
 Blockly.FieldNumber = function(opt_value, opt_min, opt_max, opt_precision,
     opt_validator) {
   this.setConstraints(opt_min, opt_max, opt_precision);
-  opt_value = (opt_value && !isNaN(opt_value)) ? String(opt_value) : '0';
+  opt_value = this.doClassValidation_(opt_value);
+  if (opt_value === null) {
+    opt_value = 0;
```
rachel-fenichel: What if 0 is outside of the constraints?  Maybe set it to the nearest constraint (min if 0 is below min, max if 0 is above max).

@RoboErikG any preference?

rachel-fenichel: Or do we just say eh, creating it with an invalid initial value is a bad idea? I think there's a similar question for all fields--what if the default value for a text field is the empty string but the validator requires it to be nonempty, for instance?  Is there even a reasonable response?

core/field_angle.js
rachel-fenichel: Note to self: I still need to dig into the rendering change on textinput, and how it reflects back into other fields.

core/field_variable.js
```diff
@@ -96,28 +104,28 @@ Blockly.FieldVariable.prototype.initModel = function() {
...
 
   // Don't fire a change event for this setValue.  It would have null as the
   // old value, which is not valid.
   Blockly.Events.disable();
-  try {
-    this.setValue(variable.getId());
-  } finally {
-    Blockly.Events.enable();
-  }
+  this.setValue(variable.getId());
```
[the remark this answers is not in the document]
BeksOmega: Affirmative

core/field_variable.js
```diff
@@ -192,30 +207,52 @@ Blockly.FieldVariable.prototype.getVariable = function() {
...
   }
-  this.variable_ = variable;
-  this.value_ = id;
-  this.setText(variable.name);
+  return newId;
+};
+
+/**
+ * Update the value of this variable field, as well as its variable and text.
+ * @param {string} newId The id of the new variable.
```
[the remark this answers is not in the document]
BeksOmega: The class validator makes sure it's valid, but then any local validator could make it invalid again. This is a problem with all fields. The setValue template method could be changed so that the class validator runs twice if there is a local validator, but there are performance costs.

core/field_variable.js
```diff
@@ -248,9 +285,8 @@ Blockly.FieldVariable.prototype.getVariableTypes_ = function() {
   var variableTypes = this.variableTypes;
   if (variableTypes === null) {
     // If variableTypes is null, return all variable types.
-    if (this.sourceBlock_) {
-      var workspace = this.sourceBlock_.workspace;
-      return workspace.getVariableTypes();
+    if (this.workspace_) {
+      return this.workspace_.getVariableTypes();
     }
```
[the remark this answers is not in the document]
BeksOmega: I don't think so. If the field's variableTypes property isn't defined, all variable types are valid, correct? Based on the logic of typeIsAllowed_, getVariableTypes should just return something falsy (like null) if variableTypes isn't defined.

It might be better to investigate this separately, if you agree I can put up an issue for it.

core/field.js
```diff
@@ -644,39 +669,106 @@ Blockly.Field.prototype.forceRerender = function() {
...
+  this.value_ = newValue;
+  this.isDirty_ = true;
+  // For backwards compatibility.
+  this.text_ = String(newValue);
+};
+
+/**
+ * Used to notify the field an invalid value was input. Can be overiden by
+ * subclasses, see FieldTextInput.
+ * No-op by default.
```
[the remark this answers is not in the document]
BeksOmega: **With regard to the value:**
The field reverts to the value it had before the start of the "edit". For text fields the edit begins when the user opens the input, and ends when the user closes the input. For a colour field each selection is an individual edit.

This means that [this ](https://user-images.githubusercontent.com/25440652/57254058-2b976400-7005-11e9-9303-f3f16c76a737.gif)colour field /doesn't/ revert to its editor-opened value (red):

But [this ](https://user-images.githubusercontent.com/25440652/57185207-0c20ff80-6e7c-11e9-8bfd-8fb7491b4646.gif)text field /does/ revert to its editor-opened value (default) instead of reverting to a value of aaa.

I didn't realize this was inconsistent until you mentioned it, so this may or may not be what you want. Should be pretty easy to change though =)

**With regard to the text:**
Restated in a more general way, I believe you're asking if by default the display will represent the input value, even if it is invalid? This is not the case, because there is no way for blockly to know what the display is to be able to update it. By default fields will act like the colour field, displaying the last valid value, and having a value of the last valid value. If a developer wants to display invalid values (or put up warnings etc) they will need to override doValueInvalid_

core/field_checkbox.js

core/field_angle.js
```diff
@@ -268,24 +272,15 @@ Blockly.FieldAngle.prototype.onMouseMove = function(e) {
...
-  Blockly.FieldAngle.superClass_.setText.call(this, text);
-  if (!this.textElement_) {
-    // Not rendered yet.
-    return;
+  // Update value.
+  if (angle != this.text_) {
+    Blockly.FieldTextInput.htmlInput_.value = angle;
+    this.setValue(angle);
+    // Always render the input angle.
+    this.text_ = String(angle);
```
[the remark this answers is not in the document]
BeksOmega: Fixed

core/field_checkbox.js
BeksOmega: Can do, do you want everything moved to render then @rachel-fenichel ?

core/field_dropdown.js
BeksOmega: Fixed

core/field_image.js
BeksOmega: Fixed

core/field_number.js
BeksOmega: I'll wait for @RoboErikG 's response before I change things.

core/field_variable.js
```diff
@@ -192,30 +207,52 @@ Blockly.FieldVariable.prototype.getVariable = function() {
 };
 
 /**
- * Set the variable ID.
- * @param {string} id New variable ID, which must reference an existing
- *     variable.
+ * Gets the validation function for this field, or null if not set.
+ * Returns null if the variable is not set, because validators should not
+ * run on the initial setValue call.
```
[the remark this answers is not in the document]
BeksOmega: Fixed

core/field_variable.js
```diff
@@ -192,30 +207,52 @@ Blockly.FieldVariable.prototype.getVariable = function() {
...
-Blockly.FieldVariable.prototype.setValue = function(id) {
-  var workspace = this.sourceBlock_.workspace;
-  var variable = Blockly.Variables.getVariable(workspace, id);
+Blockly.FieldVariable.prototype.getValidator = function() {
+  // Validators shouldn't operate on the initial setValue call.
+  // Normally this is achieved by calling setValidator after setValue, but
+  // this is not a possibility with variable fields.
+  if (this.variable_) {
+    return this.validator_;
+  }
```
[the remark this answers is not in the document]
BeksOmega: Fixed

core/field.js
BeksOmega: Opinions @rachel-fenichel ?

BeksOmega:
Ok, I resolved conversations from the last review that had been dealt with, just to keep things a bit more tidy.

I also tagged anything that needs an opinion with @rachel-fenichel  so that it'll be easier to for you to find (maybe, hopefully, idk how to github).

core/field_variable.js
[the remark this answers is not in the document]
rachel-fenichel: Add a comment that it *should* be valid, and that if validators return bad IDs it could break.

core/field_variable.js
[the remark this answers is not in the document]
rachel-fenichel: Confirmed, separate issue but should investigate before the next release.

core/field_checkbox.js
rachel-fenichel: Leave it here, and then move it in a followup.

core/field.js
rachel-fenichel: Either that, or somehow move the special sauce back into field angle (even if it means putting it back to a dynamic state).  I'd like to avoid having a throw like this in the general field render path.

BeksOmega: I ended up reverting it because I changing it to have a new DOM structure would mean reworking some CSS, and I didn't want to do that in this PR.

I created an [issue ](https://github.com/google/blockly/issues/2484)to track fixing this (because it bugs me), but if you don't think it's a big deal I'm fine with closing it.

core/field_checkbox.js
BeksOmega: Sorry, I think I got myself a bit confused, and I'm unsure of exactly what you're looking for.

I thought I was supposed to remove the optimization, and then put up an issue to re-add the optimization once I could prove it was actually more efficient. So I asked:
> do you want everything moved to render then?  

To make sure I was understanding correctly.

But now it seems like I'm not supposed to remove the optimization? Am I'm still supposed to create an issue to investigate the optimization?

I'm cool with doing whatever! I just wanted to make sure I'm not creating any confusing issues. Sorry I had to bug you again, and thank you for the helps!

core/field_variable.js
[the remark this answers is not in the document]
BeksOmega: Fixed

core/field_variable.js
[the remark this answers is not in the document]
BeksOmega: Added Issue #2483

core/field_checkbox.js
rachel-fenichel: I think you've made a good case that it's more efficient, and my issue with it is just that the code reads in a confusing way--the meat of rendering the field doesn't always happen in render or a helper called by render.  So I meant 
- you should leave the code as it is right now (correct and efficient).
- we should (separately) add a code path that rerenders only the field and not the block if the field's size doesn't change.
- once that path exists, we should move this code (and similar for other fields) back into the render function or its helpers

Does that make sense?

rachel-fenichel:
Sweet, this is getting really close.

Re: generator test failures
Travis shows this error:
https://travis-ci.org/google/blockly/jobs/536494731#L944
" WARN webdriver: Request failed due to Dismissed user prompt dialog: Error parsing XML: text.xml
InternalError: too much recursion"

So I opened the generator playground and checked only "text", then hit load.
That failed with an error that included `Blockly.Constants.Text.TEXT_CHARAT_EXTENSION`, so I knew it had to do with the `text_charAt` block:
image

For some reason it works if I load from XML on the playground with this XML:
```
<block type="text_charAt">
        <value name="VALUE">
          <block type="variables_get">
            <field name="VAR">text</field>
          </block>
        </value>
      </block>
</xml>
```
but fails if I drag in from the flyout.

The error looks like this:
```
connection.js:39 Uncaught RangeError: Maximum call stack size exceeded
    at Blockly.RenderedConnection.Blockly.Connection (connection.js:39)
    at new Blockly.RenderedConnection (rendered_connection.js:43)
    at Blockly.BlockSvg.makeConnection_ (block_svg.js:1418)
    at Blockly.BlockSvg.Blockly.Block.appendInput_ (block.js:1681)
    at Blockly.BlockSvg.appendInput_ (block_svg.js:1370)
    at Blockly.BlockSvg.Blockly.Block.appendValueInput (block.js:1379)
    at Blockly.BlockSvg.updateAt_ (text.js:840)
    at Blockly.FieldDropdown.validator_ (text.js:866)
    at Blockly.FieldDropdown.Blockly.Field.setValue (field.js:687)
    at Blockly.BlockSvg.Blockly.Block.setFieldValue (block.js:1129)
```


Anyway, that's not a solution, but should hopefully point you in the right direction for debugging.

Cheers,
Rachel

core/field.js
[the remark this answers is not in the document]
RoboErikG: We weren't very consistent before and I don't think there's a big difference between the two behaviors. I'm happy with this as is given that this makes it consistent and adds a way to override the default behavior.

core/field_checkbox.js
BeksOmega: > * you should leave the code as it is right now (correct and efficient).
> * we should (separately) add a code path that rerenders only the field and not the block if the field's size doesn't change.
> * once that path exists, we should move this code (and similar for other fields) back into the render function or its helpers
> 
> Does that make sense?

Makes sense! I put up #2490 with both proposed solutions.

BeksOmega:
So the issues with generator tests were validators calling setValue when they shouldn't.

With regard to the text char block, the comment was actually incorrect, the dropdown never gets removed. You can see this if you external-ize the inputs.

GenTests_TextAt

So that could safely be removed.

And with regard to the list_split block.
A) the setValue call wasn't included until #2326 
B) updateType_ only gets called when the dropdown is being validated (so the value will get updated anyway) or when domToMutation is called, and in that case the value should be set through the XML parsing.

GenTests_ListSplit

(same behavior as on develop - the shadow block was included to show #2326 functioning)

so that could be safely removed.

rachel-fenichel:
Makes sense.  Ready to merge?

BeksOmega:
All review comments have been dealt with! So It should be!

rachel-fenichel:
Here we go!  Thanks again for all your careful code and followups.

PR #2444, opened 2019-05-06
Mark the cursor location when opening the toolbox; cleanup

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

Inserting a block when no location was marked was previously a no-op

### Proposed Changes

If no location is marked when opening the toolbox/flyout, mark the current cursor location.
Also unify some code for inserting on ws vs at a connection.

### Reason for Changes

This seems like a reasonable guess at the user's intent.

### Test Coverage
Tested in the playground.

### DIFF
--- a/core/navigation.js
+++ b/core/navigation.js
@@ -108,6 +108,15 @@ Blockly.Navigation.markAtCursor = function() {
 };
 
 /**
+ * Remove the marker from its current location and hide it.
+ * @package
+ */
+Blockly.Navigation.removeMark = function() {
+  Blockly.Navigation.marker_.setLocation(null);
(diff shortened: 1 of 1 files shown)


core/navigation.js
```diff
@@ -551,8 +577,7 @@ Blockly.Navigation.keyboardOut = function() {
  * Mark the current location of the cursor.
  */
 Blockly.Navigation.markConnection = function() {
```
alschmiedt: Could we get rid of this and just use the markAtCursor() function?

alschmiedt: other than that lgtm

PR #2445, opened 2019-05-06
Make the marker blue on both blocks and input/output connections.

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

Marker not being distinct.
### Reason for Changes

Set the colour of the cursor differently when it's a marker.

### Test Coverage
Checked in the playground.


rachel-fenichel:
Travis failures appear to be intermittent and only on one node version, and I can't repro locally.  If it continues I will investigate more.

PR #2446, opened 2019-05-06
Disconnect blocks with X

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

Couldn't disconnect blocks before.

### Proposed Changes

Capture the X key and break the connection that the cursor is pointing to.

### Test Coverage
Tested in the playground, by disconnecting inputs, outputs, and prev connections.

### Additional Information
- The marker doesn't blink correctly after disconnecting an input/output connection.  Moving the marker with the keyboard shows that it is in the right place, but not showing up.
- There are lots of cases where this won't do anything.  We need to find a way to surface that, probably.

### DIFF
--- a/core/navigation.js
+++ b/core/navigation.js
@@ -497,6 +497,43 @@ Blockly.Navigation.insertAtConnection = function(newBlock, connection) {
   Blockly.Navigation.insertBlock(newBlock, connection);
 };
 
+/**
+ * Disconnect the connection that the cursor is pointing to, and bump blocks.
+ * This is a no-op if the connection cannot be broken or if the cursor is not
+ * pointing to a connection.
+ * @package
+ */
+Blockly.Navigation.disconnectBlocks = function() {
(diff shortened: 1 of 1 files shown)


core/navigation.js
```diff
@@ -497,6 +497,43 @@ Blockly.Navigation.insertAtConnection = function(newBlock, connection) {
...
 
+/**
+ * Disconnect the connection that the cursor is pointing to, and bump blocks.
+ * This is a no-op if the connection cannot be broken or if the cursor is not
+ * pointing to a connection.
+ * @package
+ */
+Blockly.Navigation.disconnectBlocks = function() {
+  var curNode = Blockly.Navigation.cursor_.getCurNode();
+  if (!curNode.isConnection()) {
```
RoboErikG: Should we allow disconnecting if you're on a block and it has a prev/output connection?

rachel-fenichel: I don't think so--it's not obvious at that point which connection you're going for.

PR #2452, opened 2019-05-08
Changes the cursor look for a stack

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Having the cursor for a stack and a block be the same can be pretty confusing when trying to navigate through the blocks. 

### Proposed Changes
Make the cursor for a stack cover the entire stack of blocks as well as make it slightly cover a slightly larger area than the block. 

### Test Coverage

Tested on:

### Additional Information

Screen Shot 2019-05-08 at 2 44 31 PM

### DIFF
--- a/core/cursor_svg.js
+++ b/core/cursor_svg.js
@@ -27,7 +27,7 @@
 goog.provide('Blockly.CursorSvg');
 
 goog.require('Blockly.Cursor');
-goog.require('Blockly.BlockSvg');
+goog.require('Blockly.BlockSvg.render');
 
 
 /**
--- a/core/cursor_svg.js
+++ b/core/cursor_svg.js
@@ -250,20 +250,21 @@ Blockly.CursorSvg.prototype.showWithField_ = function() {
 Blockly.CursorSvg.prototype.showWithStack_ = function() {
   var block = this.getCurNode().getLocation();
 
-  //Gets the height and width of entire stack
+  // Gets the height and width of entire stack.
   var heightWidth = block.getHeightWidth();
 
-  //Add padding so that being on a stack looks different than being on a block
+  // Add padding so that being on a stack looks different than being on a block.
   var width = heightWidth.width + Blockly.CursorSvg.STACK_PADDING;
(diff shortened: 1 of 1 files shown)


core/cursor_svg.js
```diff
@@ -237,6 +243,37 @@ Blockly.CursorSvg.prototype.showWithField_ = function() {
...
 };
 
+/**
+ * Show the cursor using a stack.
+ * @private
+ */
+Blockly.CursorSvg.prototype.showWithStack_ = function() {
+  var block = this.getCurNode().getLocation();
+
+  //Gets the height and width of entire stack
```
rachel-fenichel: General formatting comment: single line comments should have a space after the second slash, and a period at the end.  As usual for this sort of stuff, apply it to the lines you touched here and change existing ones as you touch them in other PRs.

core/cursor_svg.js
```diff
@@ -237,6 +243,37 @@ Blockly.CursorSvg.prototype.showWithField_ = function() {
...
+  var width = heightWidth.width + Blockly.CursorSvg.STACK_PADDING;
+  var height = heightWidth.height + Blockly.CursorSvg.STACK_PADDING;
+
+  //Shift the rectangle slightly to upper left so padding is equal on all sides
+  var x =  -1 * Blockly.CursorSvg.STACK_PADDING / 2;
+  var y =  -1 * Blockly.CursorSvg.STACK_PADDING / 2;
+
+  //If the block has an output connection it needs more padding
+  if (block.outputConnection) {
+    x = -1 * Blockly.CursorSvg.STACK_PADDING;
```
RoboErikG: Should this be x -= Blockly.BlockSvg.TAB_WIDTH; and do you need to add the tab width to the cursor width or is it calculated in the block's width but drawn with a - offset?

alschmiedt: Yes to the first part, but I think the TAB_WIDTH is included when we get the width for the entire block so we just need to change where it is positioned.

PR #2453, opened 2019-05-08
TDD: tests for navigation insert/modify, currently skipped

rachel-fenichel:
I'm working on unifying all of the `insertBlockToWs`, `insertBlockFromWs`, etc. code into a single function, `Blockly.Navigation.modify`.  That function acts based on the state of the cursor and the marker.  I'm currently assuming that it will return a boolean specifying whether it actually made a change (or whether it made the expected change?) but that's still flexible.

This is a start on tests for this, so that we can lay out the desired behaviour.  I'm still adding more, but want to get this out for comment.

The first commit has tests for the marker being on a previous or next connection, and the cursor being on a workspace, block, previous connection, next connection, or output connection.

### DIFF
--- a/core/navigation.js
+++ b/core/navigation.js
@@ -405,13 +405,24 @@ Blockly.Navigation.modify = function() {
     Blockly.Navigation.warn('Cannot insert with no cursor node.');
     return false;
   }
+  var markerType = markerNode.getType();
+  var cursorType = cursorNode.getType();
 
-  if (cursorNode.getType() == Blockly.ASTNode.types.FIELD) {
+  if (markerType == Blockly.ASTNode.types.FIELD) {
+    Blockly.Navigation.warn('Should not have been able to mark a field.');
+    return false;
+  }
+  if (markerType == Blockly.ASTNode.types.BLOCK) {
+    Blockly.Navigation.warn('Should not have been able to mark a block.');
+    return false;
+  }
+
+  if (cursorType == Blockly.ASTNode.types.FIELD) {
     Blockly.Navigation.warn('Cannot attach a field to anything else.');
     return false;
   }
 
-  if (cursorNode.getType() == Blockly.ASTNode.types.WORKSPACE) {
+  if (cursorType == Blockly.ASTNode.types.WORKSPACE) {
     Blockly.Navigation.warn('Cannot attach a workspace to anything else.');
     return false;
   }
@@ -423,22 +434,29 @@ Blockly.Navigation.modify = function() {
     // TODO: Handle the case when one or both are already connected.
     if (cursorNode.isConnection()) {
       return Blockly.Navigation.connect(cursorLoc, markerLoc);
-    } else if (cursorNode.getType() == Blockly.ASTNode.types.BLOCK ||
-        cursorNode.getType() == Blockly.ASTNode.types.STACK) {
+    } else if (cursorType == Blockly.ASTNode.types.BLOCK ||
+        cursorType == Blockly.ASTNode.types.STACK) {
       // TODO: Make insertFromFlyout code similar.
       return Blockly.Navigation.insertBlock(cursorLoc, markerLoc);
     }
-  } else if (markerNode.getType() == Blockly.ASTNode.types.WORKSPACE) {
+  } else if (markerType == Blockly.ASTNode.types.WORKSPACE) {
     // TODO: Make insertFromFlyout code similar.
     if (cursorNode.isConnection()) {
+      if (cursorType == Blockly.ASTNode.types.INPUT ||
+          cursorType == Blockly.ASTNode.types.NEXT) {
+        Blockly.Navigation.warn(
+            'Cannot move a next or input connection to the workspace.');
+        return false;
+      }
       var block = cursorLoc.getSourceBlock();
-    } else if (cursorNode.getType() == Blockly.ASTNode.types.BLOCK ||
-        cursorNode.getType() == Blockly.ASTNode.types.STACK) {
+    } else if (cursorType == Blockly.ASTNode.types.BLOCK ||
+        cursorType == Blockly.ASTNode.types.STACK) {
       var block = cursorLoc;
     } else {
       return false;
     }
     if (block.isShadow()) {
+      Blockly.Navigation.warn('Cannot move a shadow block to the workspace.');
       return false;
     }
     if (block.getParent()) {
@@ -446,10 +464,9 @@ Blockly.Navigation.modify = function() {
     }
     block.moveTo(markerNode.getWsCoordinate());
     return true;
-  } else if (markerNode.getType() == Blockly.ASTNode.types.BLOCK) {
-    return false;
   }
-  return true;
+  Blockly.Navigation.warn('Unexpected state in Blockly.Navigation.modify.');
+  return false;
   // TODO: Make sure the cursor and marker end up in the right places.
 };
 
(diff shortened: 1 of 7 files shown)


tests/mocha/navigation_modify_test.js
```diff
@@ -0,0 +1,157 @@
...
+  teardown(function() {
+    delete Blockly.Blocks['stack_block'];
+    delete Blockly.Blocks['row_block'];
+    this.workspace.dispose();
+    // Does disposing of the workspace dispose of cursors and markers
+    // correctly?
+  });
+
+  suite('Marked Connection', function() {
+    test('Cursor on workspace', function() {
```
RoboErikG: This can be moved into the Marker on next suite

rachel-fenichel:
From my discussion with Abby yesterday, here's a high-level description of other cases I want to cover:

- cursor is on a field or workspace: should fail
- no marker: should fail
- cursor is on a block, marker is on a connection or stack: try to find the "best" connection?  and connect if possible.
- cursor is on a block or stack, marker is on the workspace: block or stack moves to the workspace, disconnecting itself as needed.
- various cases with the marker on a block, mostly involving guessing at the "best" connection.

And other stuff about the outcome that I will need to test:
- where does the marker end up?
- where does the cursor end up?
- if the modification failed, do the blocks end up nearby, or do they go back to their original locations?

rachel-fenichel:
Lots more tests added.  If you have strong opinions on what the right behaviour is for any of these cases, please comment on the appropriate test and we can discuss.

rachel-fenichel:
Updated with the tests turned on, and with a central `modify` function in `Blockly.Navigation` that decides what to do based on the combination of cursor and marker types.

I'm not yet using this new function to handle insertion from the flyout.  That comes next.

All mocha tests pass, though eslint will fail.

I *think* that the flakiness in mocha tests that we're seeing right now is because workspaces bump blocks after loading from XML, with a call to `setTimeout`.  But I'm not 100% sure of that yet.

core/navigation.js
```diff
@@ -387,6 +387,103 @@ Blockly.Navigation.resetFlyout = function(shouldHide) {
...
+    Blockly.Navigation.warn('Cannot insert with no marked node.');
+    return false;
+  }
+
+  if (!cursorNode) {
+    Blockly.Navigation.warn('Cannot insert with no cursor node.');
+    return false;
+  }
+
+  if (cursorNode.getType() == Blockly.ASTNode.types.FIELD) {
```
alschmiedt: Should we also have this check for a field marker node?

core/navigation.js
```diff
@@ -387,6 +387,103 @@ Blockly.Navigation.resetFlyout = function(shouldHide) {
...
+    // TODO: Make insertFromFlyout code similar.
+    if (cursorNode.isConnection()) {
+      var block = cursorLoc.getSourceBlock();
+    } else if (cursorNode.getType() == Blockly.ASTNode.types.BLOCK ||
+        cursorNode.getType() == Blockly.ASTNode.types.STACK) {
+      var block = cursorLoc;
+    } else {
+      return false;
+    }
+    if (block.isShadow()) {
```
alschmiedt: Should we add a warning here with something like "Cannot move shadow blocks." ?

core/navigation.js
```diff
@@ -425,43 +522,16 @@ Blockly.Navigation.findBestConnection = function(block, connection) {
...
-  var targetConnection = Blockly.Navigation.getInsertionConnection();
-  var sourceConnection = Blockly.Navigation.cursor_.getCurNode().getLocation();
-  if (targetConnection && sourceConnection) {
-    try {
-      sourceConnection.connect(targetConnection);
-    } catch (Error) {
-      Blockly.Navigation.warn('The connection block is not the right type');
-    }
+  if (bestConnection.isConnected() &&
+      !bestConnection.targetConnection.getSourceBlock().isShadow()) {
```
alschmiedt: nit: could we use besetConnection.targetBlock() instead or bestConnection.targetConnection.getSourceBlock()?

tests/mocha/navigation_modify_test.js
```diff
@@ -0,0 +1,328 @@
...
+      });
+      test('Cursor on workspace', function() {
+        Blockly.Navigation.cursor_.setLocation(
+            Blockly.ASTNode.createWorkspaceNode(
+                this.workspace, new goog.math.Coordinate(100, 100)));
+        chai.assert.isFalse(Blockly.Navigation.modify());
+
+      });
+    });
+    suite('Marked stack block', function() {
```
RoboErikG: Just a note: This and the 'Marked row block' suites are behavior I think we'll want to hide higher up. We should try to do something reasonable if the user marks a block and tries to insert another block.

core/navigation.js
rachel-fenichel: Per previous discussion, we aren't going to let the marker be on a field.  But I'm not sure where we're going to block that, so yeah--it's probably worth having this check and fail if the marker is on a field.  I'll add it.

core/navigation.js
```diff
@@ -387,6 +387,103 @@ Blockly.Navigation.resetFlyout = function(shouldHide) {
...
+    if (cursorNode.isConnection()) {
+      return Blockly.Navigation.connect(cursorLoc, markerLoc);
+    } else if (cursorNode.getType() == Blockly.ASTNode.types.BLOCK ||
+        cursorNode.getType() == Blockly.ASTNode.types.STACK) {
+      // TODO: Make insertFromFlyout code similar.
+      return Blockly.Navigation.insertBlock(cursorLoc, markerLoc);
+    }
+  } else if (markerNode.getType() == Blockly.ASTNode.types.WORKSPACE) {
+    // TODO: Make insertFromFlyout code similar.
+    if (cursorNode.isConnection()) {
```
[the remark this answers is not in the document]
rachel-fenichel: Limit to previous and output types, I think.

tests/mocha/navigation_modify_test.js
rachel-fenichel: Agreed.  These are here partly because I was just walking through a matrix of cursor/marker combinations, and partly to point out that even if we don't want to do anything here, we need some way to signal that if it's at all possible for the user to get here.

tests/mocha/navigation_modify_test.js
```diff
@@ -0,0 +1,328 @@
...
+    test('Cursor on child block (row)', function() {
+      this.row_block_1.inputList[0].connection.connect(
+          this.row_block_2.outputConnection);
+
+      Blockly.Navigation.cursor_.setLocation(
+          Blockly.ASTNode.createBlockNode(
+              this.row_block_2));
+
+      chai.assert.isTrue(Blockly.Navigation.modify());
+      chai.assert.isNull(this.row_block_2.getParent());
```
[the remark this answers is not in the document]
rachel-fenichel: Checking that the parent was null was supposed to be the check whether it's at the top level.  I suppose I could check if it's in the list of top blocks, but if it has a null parent and isn't in the list of top blocks, I think that's an issue with code in block.js.

Added asserts on location, here and for 'Cursor on child block (stack)'.

tests/mocha/navigation_modify_test.js
```diff
@@ -0,0 +1,328 @@
...
+        chai.assert.isNull(this.stack_block_1.getPreviousBlock());
+      });
+      test('Cursor on really incompatible connection', function() {
+        Blockly.Navigation.cursor_.setLocation(
+            Blockly.ASTNode.createConnectionNode(
+                this.row_block_1.outputConnection));
+        chai.assert.isFalse(Blockly.Navigation.modify());
+        chai.assert.isNull(this.stack_block_1.getNextBlock());
+      });
+      test('Cursor on block', function() {
```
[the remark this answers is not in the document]
rachel-fenichel: Good call-that one caught a bug.

rachel-fenichel:
- lint fixed
- mocha tests all pass, but there's still that flakiness, so Travis may say that they fail
- I may just turn off linting for mocha tests.  It's off for the other tests, and I'm much less concerned about style in test code.

PR #2456, opened 2019-05-10
Maybe fix mocha tests

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Proposed Changes

In the bump function that happens on all simple workspaces, don't bump if the block has a parent.
### Reason for Changes

Tests are failing intermittently.  I think that's because the bump function runs on some delay/setTimeout after loading the blocks from XML, and the blocks getting bumped have already been attached by the time that setTimeout runs.  One solution is to switch to a workspace that scrolls infinitely for these tests, but I think this is safer.

### Test Coverage
Tested locally in the browser.

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -293,17 +293,22 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
             Blockly.Events.setGroup(e.group);
           }
 
+          var shouldBump = false;
           switch (e.type) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
```
============Blockly Mocha Test Summary=================
null
null tests failed
============Blockly Mocha Test Summary=================
```
Well I guess this didn't fix them, then.

PR #2457, opened 2019-05-10
Remove goog.userAgent dependency.

NeilFraser:


core/field_dropdown.js
```diff
@@ -97,7 +97,8 @@ Blockly.FieldDropdown.MAX_MENU_HEIGHT_VH = 0.45;
 /**
  * Android can't (in 2014) display "▾", so use "▼" instead.
```
rachel-fenichel: Out of curiosity, is this still true?

PR #2458, opened 2019-05-11
Add STATEMENT_SUFFIX

NeilFraser:
Also remove need to manually indent INFINITE_LOOP_TRAP.  This is a breaking change for Python users of this property.  However, very few users of this property exist, given that the existing code breaks if the block ID has a $ in it (also fixed in this PR).


blocks/loops.js
```diff
@@ -303,7 +303,24 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
    * Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN.LOOP_TYPES.push('custom_loop');
    */
   LOOP_TYPES: ['controls_repeat', 'controls_repeat_ext', 'controls_forEach',
-    'controls_for', 'controls_whileUntil'],
+      'controls_for', 'controls_whileUntil'],
```
RoboErikG: This line is triggering a lint error. Should only be 4 spaces.

blocks/loops.js
```diff
@@ -315,17 +332,8 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
...
-    var block = this;
-    do {
-      if (this.LOOP_TYPES.indexOf(block.type) != -1) {
-        legal = true;
-        break;
-      }
-      block = block.getSurroundParent();
-    } while (block);
-    if (legal) {
+    if (Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN
```
RoboErikG: Can this just be this.getSurroundLoop(this) since both methods are part of the same mixin?

NeilFraser:
Closing due to a catastrophic Git rebase.  Apparently I forgot how to computer.

PR #2459, opened 2019-05-11
Added Field Value Tests

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2169 specifically [this](https://github.com/google/blockly/issues/2169#issuecomment-488471207).

### Proposed Changes

Adds tests to document how fields should react to different values.

Also clarifies which fields have default values and which do not.

### Reason for Changes

Tests needed to be added before #2441 possibly breaks the world.

### Test Coverage

This is tests silly!

### Additional Information

So... a lot of these tests are actually failing! But that's not because the tests are broken, it's because the fields are broken.

Examples:
* Most validator tests fail because atm most fields don't validate unless they are attached to a block. ([link](https://github.com/google/blockly/issues/2169#issuecomment-491080867))
* All checkbox field tests are failing because checkboxes don't handle their text_ properly.
* Lots of fields have less severe checks on setValue than on their constructors, so several setValue tests are failing (e.g. most fields fail the non-parsable string test).

Plus a smattering of other field-specific bugs.

### Important information

I have some TODO notes in the tests, these are requests for information on what the intended behavior is.

For example, most fields will revert to a default value (meaning the value they would have had had they been constructed no parameters) when passed a non-parsable value. I believe this is a bug, and non-parsable values should validated to null inside a field's doClassValidation_ function, but I wanted confirmation.

### DIFF
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -45,7 +45,10 @@ goog.require('goog.math.Size');
 Blockly.FieldLabel = function(text, opt_class) {
   this.size_ = new goog.math.Size(0, 17.5);
   this.class_ = opt_class;
-  this.setValue(text || '');
+  if (text === null || text === undefined) {
+    text = '';
+  }
+  this.setValue(String(text));
   this.tooltip_ = '';
 };
 goog.inherits(Blockly.FieldLabel, Blockly.Field);
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -36,7 +36,8 @@ goog.require('goog.math.Size');
 
 /**
  * Class for a non-editable, non-serializable text field.
- * @param {string} text The initial content of the field.
+ * @param {string=} text The initial content of the field, defaults to an
+ *    empty string.
  * @param {string=} opt_class Optional CSS class for the field's text.
  * @extends {Blockly.Field}
  * @constructor
@@ -44,7 +45,7 @@ goog.require('goog.math.Size');
 Blockly.FieldLabel = function(text, opt_class) {
   this.size_ = new goog.math.Size(0, 17.5);
   this.class_ = opt_class;
-  this.setValue(text);
+  this.setValue(text || '');
   this.tooltip_ = '';
 };
 goog.inherits(Blockly.FieldLabel, Blockly.Field);
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -37,7 +37,8 @@ goog.require('goog.math.Coordinate');
 
 /**
  * Class for an editable text field.
- * @param {string} text The initial content of the field.
+ * @param {string=} text The initial content of the field, defaults to an
+ *    empty string.
  * @param {Function=} opt_validator An optional function that is called
  *     to validate any constraints on what the user entered.  Takes the new
  *     text as an argument and returns either the accepted text, a replacement
@@ -46,7 +47,7 @@ goog.require('goog.math.Coordinate');
  * @constructor
  */
 Blockly.FieldTextInput = function(text, opt_validator) {
-  Blockly.FieldTextInput.superClass_.constructor.call(this, text,
+  Blockly.FieldTextInput.superClass_.constructor.call(this, text || '',
       opt_validator);
 };
 goog.inherits(Blockly.FieldTextInput, Blockly.Field);
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -47,7 +47,10 @@ goog.require('goog.math.Coordinate');
  * @constructor
  */
 Blockly.FieldTextInput = function(text, opt_validator) {
-  Blockly.FieldTextInput.superClass_.constructor.call(this, text || '',
+  if (text === null || text === undefined) {
+    text = '';
+  }
+  Blockly.FieldTextInput.superClass_.constructor.call(this, String(text),
       opt_validator);
 };
 goog.inherits(Blockly.FieldTextInput, Blockly.Field);
--- a/tests/mocha/field_angle_test.js
+++ b/tests/mocha/field_angle_test.js
@@ -115,95 +115,90 @@ suite ('Angle Fields', function() {
     });
   });
   suite('setValue', function() {
-    test('Empty -> Null', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue(null);
-      assertValueDefault(angleField);
+    suite('Empty -> New Value', function() {
+      var angleField;
+      setup(function() {
+        angleField = new Blockly.FieldAngle();
+      });
+      test('Null', function() {
+        var angleField = new Blockly.FieldAngle();
+        angleField.setValue(null);
+        assertValueDefault(angleField);
+      });
+      test('Undefined', function() {
+        angleField.setValue(undefined);
+        assertValueDefault(angleField);
+      });
+      test.skip('Non-Parsable String', function() {
+        angleField.setValue('bad');
+        assertValueDefault(angleField);
+      });
+      test('NaN', function() {
+        angleField.setValue(NaN);
+        assertValueDefault(angleField);
+      });
+      test('Integer', function() {
+        angleField.setValue(2);
+        assertValue(angleField, 2);
+      });
+      test('Float', function() {
+        angleField.setValue(2.5);
+        assertValue(angleField, 2.5);
+      });
+      test('Integer String', function() {
+        angleField.setValue('2');
+        assertValue(angleField, 2);
+      });
+      test('Float', function() {
+        angleField.setValue('2.5');
+        assertValue(angleField, 2.5);
+      });
+      test('>360°', function() {
+        angleField.setValue(361);
+        assertValue(angleField, 1);
+      });
     });
-    test('Value -> Null', function() {
-      var angleField = new Blockly.FieldAngle(1);
-      angleField.setValue(null);
-      assertValue(angleField, 1);
-    });
-    test('Empty -> Undefined', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue(undefined);
-      assertValueDefault(angleField);
-    });
-    test.skip('Value -> Undefined', function() {
-      var angleField = new Blockly.FieldAngle(1);
-      angleField.setValue(undefined);
-      assertValue(angleField, 1);
-    });
-    test.skip('Empty -> Non-Parsable String', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue('bad');
-      assertValueDefault(angleField);
-    });
-    test.skip('Value -> Non-Parsable String', function() {
-      var angleField = new Blockly.FieldAngle(1);
-      angleField.setValue('bad');
-      assertValue(angleField, 1);
-    });
-    test('Empty -> NaN', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue(NaN);
-      assertValueDefault(angleField);
-    });
-    test.skip('Value -> NaN', function() {
-      var angleField = new Blockly.FieldAngle(1);
-      angleField.setValue(NaN);
-      assertValue(angleField, 1);
-    });
-    test('Empty -> Integer', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue(2);
-      assertValue(angleField, 2);
-    });
-    test('Value -> Integer', function() {
-      var angleField = new Blockly.FieldAngle(1);
-      angleField.setValue(2);
-      assertValue(angleField, 2);
-    });
-    test('Empty -> Float', function() {
-      var angleField = new Blockly.FieldAngle();
-      angleField.setValue(2.5);
-      assertValue(angleField, 2.5);
-    });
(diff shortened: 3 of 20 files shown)


core/field_label.js
```diff
@@ -36,12 +36,16 @@ goog.require('goog.math.Size');
...
 /**
  * Class for a non-editable, non-serializable text field.
- * @param {string} text The initial content of the field.
+ * @param {!string} text The initial content of the field.
  * @param {string=} opt_class Optional CSS class for the field's text.
  * @extends {Blockly.Field}
  * @constructor
  */
 Blockly.FieldLabel = function(text, opt_class) {
+  if (!text) {
```
RoboErikG: This will be false for the empty string as well, which I don't think is a behavior change you intended to make. 

In general, why are you adding this check? It looks like setText will handled it by casting to a String if it's not null. Is it a problem with validators?

core/field_textinput.js
```diff
@@ -46,6 +46,9 @@ goog.require('goog.userAgent');
  * @constructor
  */
 Blockly.FieldTextInput = function(text, opt_validator) {
+  if (!text) {
```
RoboErikG: Same issue, and empty string should be a valid starting point for text input fields.

core/field_label.js
BeksOmega: You're right, Fixed =)

core/field_textinput.js
BeksOmega: Fixed

tests/mocha/field_angle_test.js
```diff
@@ -0,0 +1,234 @@
...
+    });
+    test('Empty -> Undefined', function() {
+      var angleField = new Blockly.FieldAngle();
+      angleField.setValue(undefined);
+      assertValueDefault(angleField);
+    });
+    test('Value -> Undefined', function() {
+      var angleField = new Blockly.FieldAngle(1);
+      angleField.setValue(undefined);
+      // TODO: Is this how we want this to work?
```
RoboErikG: I think this is a bug in field angle. Should probably be treated the same as null.

tests/mocha/field_angle_test.js
```diff
@@ -0,0 +1,234 @@
...
+    });
+    test('Empty -> Non-Parsable String', function() {
+      var angleField = new Blockly.FieldAngle();
+      angleField.setValue('bad');
+      assertValueDefault(angleField);
+    });
+    test('Value -> Non-Parsable String', function() {
+      var angleField = new Blockly.FieldAngle(1);
+      angleField.setValue('bad');
+      // TODO: Is this how we want this to work?
```
RoboErikG: Ditto. Should probably not change the value if the input is invalid.

tests/mocha/field_checkbox_test.js
```diff
@@ -0,0 +1,159 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+suite ('Checkbox Fields', function() {
+  function assertValue(checkboxField, expectedValue, expectedText) {
+    var actualValue = checkboxField.getValue();
+    var actualText = checkboxField.getText();
+    assertEquals(actualValue, expectedValue);
+    assertEquals(actualText, expectedText);
```
RoboErikG: +1 for checking both the value and the text. That'll make supporting screen readers easier.

tests/mocha/field_checkbox_test.js
```diff
@@ -0,0 +1,159 @@
...
+        checkboxField.setValidator(function() {
+          return 'TRUE';
+        });
+      });
+      test('New Value', function() {
+        checkboxField.setValue('FALSE');
+        assertValue(checkboxField, 'TRUE', 'true');
+      });
+    });
+    suite('Always True Validator', function() {
```
RoboErikG: s/True/False

RoboErikG:
Filed https://github.com/google/blockly/issues/2462 for to track bad value handling in fields.

tests/mocha/field_colour_test.js
```diff
@@ -0,0 +1,179 @@
...
+      assertValue(colourField, '#bcbcbc', '#bcbcbc');
+    });
+    test('#bcbcbc', function() {
+      var colourField = Blockly.FieldColour.fromJson({ colour: '#bcbcbc' });
+      assertValue(colourField, '#bcbcbc', '#bcbcbc');
+    });
+  });
+  suite('setValue', function() {
+    test('#ffffff -> Null', function() {
+      var colourField = new Blockly.FieldColour('#ffffff');
```
RoboErikG: Creating the colourField can be moved into setup for this suite. Also, you should use a value other than the default so you can distinguish when it's falling back to the default vs the current value.

tests/mocha/field_colour_test.js
```diff
@@ -0,0 +1,179 @@
...
+    test('#ffffff -> #bcbcbc', function() {
+      var colourField = new Blockly.FieldColour('#ffffff');
+      colourField.setValue('#bcbcbc');
+      assertValue(colourField, '#bcbcbc', '#bcbcbc');
+    });
+  });
+  suite('Validators', function() {
+    var colourField;
+    setup(function() {
+      colourField = new Blockly.FieldColour('#ffffff');
```
RoboErikG: Use a colour string other than the default so you can distinguish default vs previous value.

tests/mocha/field_colour_test.js
```diff
@@ -0,0 +1,179 @@
...
+
+suite ('Colour Fields', function() {
+  function assertValue(colourField, expectedValue, expectedText) {
+    var actualValue = colourField.getValue();
+    var actualText = colourField.getText();
+    assertEquals(actualValue, expectedValue);
+    assertEquals(actualText, expectedText);
+  }
+  function assertValueDefault(colourField) {
+    var expectedValue = Blockly.FieldColour.COLOURS[0];
```
RoboErikG: For testing consistency you might want to change Blockly.FieldColour.COLOURS to a specific set of colours in setup and change it back in teardown.

tests/mocha/field_date_test.js
```diff
@@ -0,0 +1,176 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+suite ('Date Fields', function() {
+  function assertValue(dateField, expectedValue) {
+    var actualValue = dateField.getValue();
+    var actualText = dateField.getText();
+    var stringExpectedValue = String(expectedValue);
+    assertEquals(String(actualValue), stringExpectedValue);
```
RoboErikG: Why do you compare the values as strings?

tests/mocha/field_date_test.js
```diff
@@ -0,0 +1,176 @@
...
+    test('Non-Parsable String', function() {
+      var dateField = new Blockly.FieldDate('bad');
+      assertValueDefault(dateField);
+    });
+    test('2020-02-20', function() {
+      var dateField = new Blockly.FieldDate('2020-02-20');
+      assertValue(dateField, '2020-02-20');
+    });
+    test('Invalid Date - Month(2020-13-20)', function() {
+      // TODO: What do we want it to do in this case?
```
RoboErikG: Test as written seems reasonable to me.

tests/mocha/field_date_test.js
```diff
@@ -0,0 +1,176 @@
...
+      var dateField = new Blockly.FieldDate('2020-13-20');
+      assertValueDefault(dateField);
+    });
+    test('Invalid Date - Day(2020-02-32)', function() {
+      // TODO: What do we want it to do in this case?
```
RoboErikG: ditto.

tests/mocha/field_date_test.js
```diff
@@ -0,0 +1,176 @@
...
+    });
+    test('Empty -> Undefined', function() {
+      var dateField = new Blockly.FieldDate();
+      dateField.setValue(undefined);
+      assertValueDefault(dateField);
+    });
+    test('Value -> Undefined', function() {
+      var dateField = new Blockly.FieldDate('2020-02-20');
+      dateField.setValue(undefined);
+      // TODO: Is this how we want this to work?
```
RoboErikG: As above, probably want it to leave the current value instead of reverting to the default.

rachel-fenichel:
For tests that fail now because the fields do the wrong thing--you can write them as `test.skip('Name', function() {});` or `suite.skip('Name', function() {});`

The tests will show up on the index page, but won't be run.  Then you can enable them as you fix the bugs.

tests/mocha/field_number_test.js
```diff
@@ -0,0 +1,352 @@
...
+      });
+      test('When Editing', function() {
+        numberFieldField.isBeingEdited_ = true;
+        numberFieldField.setValue(2);
+        assertValue(numberFieldField, 1, '2');
+        numberFieldField.isBeingEdited_ = false;
+      });
+      test('When Not Editing', function() {
+        numberFieldField.setValue(2);
+        assertValue(numberFieldField, 2);
```
RoboErikG: Shouldn't a null validator cause the value to be reverted?

tests/mocha/field_textinput_test.js
```diff
@@ -0,0 +1,142 @@
...
+      assertValue(textInputField, '1');
+    });
+    test('Boolean', function() {
+      var textInputField = Blockly.FieldTextInput.fromJson({ text:true });
+      assertValue(textInputField, 'true');
+    });
+  });
+  suite('setValue', function() {
+    test('Null', function() {
+      var textInputField = new Blockly.FieldTextInput('value');
```
RoboErikG: pull out into setup?

tests/mocha/field_variable_test.js
```diff
@@ -8,54 +8,46 @@ suite('Variable Fields', function() {
       }
```
RoboErikG: Looks like we forgot to add a license header. Would you mind adding one?

tests/mocha/field_variable_test.js
```diff
@@ -8,54 +8,46 @@ suite('Variable Fields', function() {
...
+  }
+  function assertValueDefault(variableField) {
+    var actualValue = variableField.getValue();
+    var actualText = variableField.getText();
+    assertNotNull(actualValue);
+    assertNotUndefined(actualValue);
+    assertNotNull(actualText);
+    assertNotUndefined(actualText);
+  }
+  function assertValueName(variableField, expectedText) {
```
RoboErikG: Ideally you'd want to assert the id and the text are as expected. You may need to replace the method that generates ids for variables with a mock to do this, though. Specificially getOrCreateVariablePackage.

tests/mocha/field_variable_test.js
```diff
@@ -8,54 +8,46 @@ suite('Variable Fields', function() {
...
-
+  function createAndInitFieldConstructor(workspace, variableName) {
+    return initField(new Blockly.FieldVariable(variableName), workspace);
+  }
+  function createAndInitFieldJson(workspace, variableName) {
+    return initField(Blockly.FieldVariable.fromJson(
+        { variable:variableName }), workspace);
+  }
+  function assertValueDefault(variableField) {
+    var actualValue = variableField.getValue();
```
RoboErikG: From code, the default name will be 'i' and the default value will be a random UUID. It might be worth mocking some of the methods in Blockly.Variables so you can control the defaults and make the tests for default clearer.

RoboErikG: Note: It looks like sinon is the proper way to do mocking in mocha.

Example: https://github.com/google/blockly/blob/develop/tests/mocha/event_test.js#L84
Docs: https://sinonjs.org/

tests/mocha/field_variable_test.js
BeksOmega: I ended up stubbing Blockly.utils.genUid and Blockly.Variables.generateUniqueName instead. It should work the same.

BeksOmega:
@RoboErikG  should be good to go.

core/field_label.js
```diff
@@ -36,15 +36,16 @@ goog.require('goog.math.Size');
...
+ *    empty string.
  * @param {string=} opt_class Optional CSS class for the field's text.
  * @extends {Blockly.Field}
  * @constructor
  */
 Blockly.FieldLabel = function(text, opt_class) {
   this.size_ = new goog.math.Size(0, 17.5);
   this.class_ = opt_class;
-  this.setValue(text);
+  this.setValue(text || '');
```
RoboErikG: I didn't think enough about this before. In your tests you require 1 -> '1' and true -> 'true', but not 0 -> '0' and false -> 'false' . It looks like those values will result in the empty string instead.

Here and elsewhere can we update this to explicitly check for null/undefined instead and let the majority of things be auto-cast to strings?

BeksOmega: Fixed for all applicable fields.

rachel-fenichel:
I'm taking a look now, will comment when I'm done.  @RoboErikG thanks for the thorough review, and thanks @BeksOmega for the thorough tests!

rachel-fenichel:
I'm happy with this!  @RoboErikG please confirm and then merge.

PR #2463, opened 2019-05-14
Fraser suffix2

NeilFraser:
Adding ability to insert suffix code after every statement.  All prefix code is guaranteed to have a suffix, and vice versa.

### DIFF
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -302,8 +302,19 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
    * To add a new loop type add this to your code:
    * Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN.LOOP_TYPES.push('custom_loop');
    */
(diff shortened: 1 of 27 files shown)


blocks/loops.js
```diff
@@ -302,8 +302,36 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
...
+
+  /**
+   * Is the given block enclosed (at any level) by a loop?
+   * @param {!Blockly.Block} block Current block.
+   * @return {Blockly.Block} The nearest surrounding loop, or null if none.
+   */
+  getSurroundLoop: function(block) {
+    // Is the block nested in a loop?
+    do {
+      if (Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN.LOOP_TYPES
```
NeilFraser: @RoboErikG Using 'this' worries me, is the array on the block going to be === to the array on CONTROL_FLOW_IN_LOOP_CHECK_MIXIN after developers push to the latter?
If they are (and I *think* they are), semantically we are checking a central list, not a customizable list on each block.
Honestly this whole architecture is rather poor.  I'd rather have an 'isLoop' property in the XML of loop blocks.

PR #2467, opened 2019-05-17
Fix bug with in bounds bumping

RoboErikG:
Traced one cause of Mocha tests sometimes failing to blocks being
bumped back into bounds. It looks like a block being moved to be
a child of another block ended up out of bounds probably
because the size of the workspace isn't set yet. When the change
handler tried to move it back in it would throw an error since you
can't move child blocks with that method.

## The basics

- [ ] I branched from develop

### Test Coverage

Tested on:

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2467)


rachel-fenichel:
https://github.com/google/blockly/pull/2456 was my previous attempt at this.

rachel-fenichel:
LG if travis says okay.

PR #2469, opened 2019-05-20
Load a list of tests to run from a file; connect test output to dif-viewer

rachel-fenichel:
## The basics

- [ ] I branched from develop

^ None of these, but the style guide doesn't currently apply in this directory.

### Proposed Changes

- Use a `test_cases.json` file to specify which tests to run.
- Use a custom reporter for mocha to output test results in a way that's easy to use.
- Use those results in `diff_viewer.html`.
- Make sure folders exist
- Automatically open `diff_viewer.html` after running tests

### DIFF
--- a/tests/screenshot/diff-reporter.js
+++ b/tests/screenshot/diff-reporter.js
@@ -23,7 +23,7 @@
 /**
  * @fileoverview Reporter that prints results to the console with the same
  * format as the spec reporter, but also saves a test_output.js file with a
- * that just wraps a json object, for use in diff_viewer.html.
+ * variable that just wraps a json object, for use in diff_viewer.html.
  */
 var mocha = require('mocha');
 var fs = require("fs");
--- a/tests/screenshot/diff-reporter.js
+++ b/tests/screenshot/diff-reporter.js
@@ -1,0 +1,131 @@
+// diff-reporter.js
+
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2019 Google Inc.
+ * https://developers.google.com/blockly/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


tests/screenshot/diff-reporter.js
```diff
@@ -0,0 +1,131 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Reporter that prints results to the console with the same
+ * format as the spec reporter, but also saves a test_output.js file with a
```
RoboErikG: "with a / that just wraps" you accidentally a noun.

tests/screenshot/diff_screenshots.js
```diff
@@ -31,35 +31,55 @@ var old_dir = 'tests/screenshot/outputs/old/';
...
-// - iterate over all test cases
-// - use the name of the test case file as the name of the test
-// - improve doneReading to not rely on globals like this
-// - write or find basic html page that lets you open pairs of outputs + diffs
+if (!fs.existsSync(diff_dir)){
+  fs.mkdirSync(diff_dir);
+}
+
+function getTestList() {
+  var file = fs.readFileSync('tests/screenshot/test_cases/test_cases.json');
```
RoboErikG: Should this path be up with the new dir?

tests/screenshot/gen_screenshots.js
```diff
@@ -34,25 +40,39 @@ module.exports = genScreenshots;
...
+  for (var i = 0, testName; testName = test_list[i]; i++) {
+    await genSingleScreenshot(browser_new, 'new', testName);
+    await genSingleScreenshot(browser_old, 'old', testName);
+  }
 
   await cleanUp(browser_new, browser_old);
   return 0;
 }
 
+function getTestList() {
```
RoboErikG: This looks like the same code as in diff_screenshots.js. Should some of this get pulled out into a utils file?

tests/screenshot/diff_screenshots.js
rachel-fenichel: Done.

tests/screenshot/diff-reporter.js
rachel-fenichel: Done.

tests/screenshot/gen_screenshots.js
rachel-fenichel: Maybe?  It's only this function.  Definitely if I end up adding more shared code.

rachel-fenichel:
Added a few things I remembered during lunch:
- gitignore
- Check if the old screenshot exists, and skip if so

PR #2470, opened 2019-05-20
Add a checkbox to toggle rendering debug visualizations in the playground

rachel-fenichel:
Like the logEvents checkbox, it persists its state across reloads.

It controls both rectangles and connection dots.


rachel-fenichel:
@alschmiedt FYI

PR #2472, opened 2019-05-21
Refactor top row

alschmiedt:
## The basics

- [ ] I branched from develop

### Proposed Changes
Refactors the top row to inherit from Row instead of Measurable. Top row is now treated like a row and has an elements list with measurables such as corners, previous connection, and hats. 

### Reason for Changes
It makes more sense for the top row to be treated like a row. 

### Test Coverage

Tested on:


core/block_rendering_rewrite/block_render_info.js
```diff
@@ -127,6 +127,8 @@ Blockly.BlockRendering.RenderInfo.prototype.measure_ = function() {
  */
 Blockly.BlockRendering.RenderInfo.prototype.createRows_ = function() {
   var activeRow = new Blockly.BlockRendering.Row();
+  this.topRow = this.createTopRow_();
```
rachel-fenichel: Move these above the `activeRow` line, so that it's clear that the `activeRow` will come after these.

Do we still need `this.topRow` now that it's not as distinct?

core/block_rendering_rewrite/block_render_info.js
```diff
@@ -160,6 +162,37 @@ Blockly.BlockRendering.RenderInfo.prototype.createRows_ = function() {
   }
 };
 
+/**
+ * Create the top row and fill the elements list with the correct measurables
+ * based on values from the block.
+ * @return {Blockly.BlockRendering.TopRow} The top row with all the elements
```
rachel-fenichel: Why return?  I expected you to just set `this.topRow` inside this function.

rachel-fenichel: With all non-spacer elements created.

core/block_rendering_rewrite/block_render_info.js
```diff
@@ -292,6 +337,23 @@ Blockly.BlockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
     }
   }
 
+  // Spacing between a hat and a corner
+  if (prev.isRoundedCorner() || prev.isSquareCorner()) {
```
rachel-fenichel: Shouldn't even have a rounded corner preceding a hat.

core/block_rendering_rewrite/block_render_info.js
```diff
@@ -356,7 +418,11 @@ Blockly.BlockRendering.RenderInfo.prototype.alignRowElements_ = function() {
       var desiredWidth = this.width;
       var missingSpace = desiredWidth - currentWidth;
       if (missingSpace) {
-        this.addAlignmentPadding_(row, missingSpace);
+        if (row.type === 'top row') {
+          row.getLastSpacer().width += missingSpace;
```
rachel-fenichel: This looks like it belongs in `addAlignmentPadding`.

core/block_rendering_rewrite/block_render_info.js
alschmiedt: We use it in the draw function and draw highlight function so I think it is probably worth it to keep it around? It seems cleaner than getting the first element in the list of rows every time.

core/block_rendering_rewrite/measurables.js
```diff
@@ -26,6 +26,15 @@ Blockly.BlockRendering.Measurable.prototype.isField = function() {
   return this.type == 'field';
 };
 
+/**
```
rachel-fenichel: Note for later: we *may* remove these helper functions if all of them end up being direct checks against types.

core/block_rendering_rewrite/measurables.js
```diff
@@ -208,6 +263,86 @@ Blockly.BlockRendering.ExternalValueInput = function(input) {
...
+ * An object containing information about the space a previous connection takes
+ * up during rendering.
+ * @package
+ * @constructor
+ */
+Blockly.BlockRendering.PreviousConnection = function() {
+  Blockly.BlockRendering.PreviousConnection.superClass_.constructor.call(this);
+  this.renderRect = null;
```
rachel-fenichel: If you're not using `renderRect`, remove it.  On `render/collab` it's only ever set to null, and never used.

core/block_rendering_rewrite/measurables.js
```diff
@@ -208,6 +263,86 @@ Blockly.BlockRendering.ExternalValueInput = function(input) {
...
+  this.type = 'previous connection';
+  this.height = BRC.MEDIUM_PADDING;
```
rachel-fenichel: Add a new constant for this height.  `BRC.STACK_CONNECTION_HEIGHT`, maybe?  Use it here and for NextConnection, and set it equal to `BRC.MEDIUM_PADDING`.  Changing connection height/shapes is a common tweak for rendering.

core/block_rendering_rewrite/measurables.js
alschmiedt: Made a gnosis ticket in cleanup section and added todo

core/block_rendering_rewrite/measurables.js
rachel-fenichel: Or `NOTCH_HEIGHT, since we seem to be calling those 'notches'

core/block_rendering_rewrite/measurables.js
```diff
@@ -208,6 +263,86 @@ Blockly.BlockRendering.ExternalValueInput = function(input) {
...
+ * An object containing information about the space a hat takes up during
+ * rendering.
+ * @package
+ * @constructor
+ */
+Blockly.BlockRendering.Hat = function() {
+  Blockly.BlockRendering.Hat.superClass_.constructor.call(this);
+  this.renderRect = null;
+  this.type = 'hat';
+  this.height = BRC.MEDIUM_PADDING;
```
rachel-fenichel: Add a separate constant for this height as well.

rachel-fenichel: ...though actually, this isn't the height up, right?  It's the height down into the row?  If so, does this need to have any height at all, or can we say its height is 0?

core/block_rendering_rewrite/measurables.js
```diff
@@ -208,6 +263,86 @@ Blockly.BlockRendering.ExternalValueInput = function(input) {
...
+ * An object containing information about the space a rounded corner takes up
+ * during rendering.
+ * @package
+ * @constructor
+ */
+Blockly.BlockRendering.RoundCorner = function() {
+  Blockly.BlockRendering.RoundCorner.superClass_.constructor.call(this);
+  this.renderRect = null;
+  this.type = 'round corner';
+  this.width = 8;
```
rachel-fenichel: Use constants.

core/block_rendering_rewrite/measurables.js
```diff
@@ -273,38 +412,36 @@ Blockly.BlockRendering.InRowSpacer = function(width) {
...
+ * An object containing information about what elements are in the top row of a
+ * block as well as spacing information for the top row.
+ * Elements in a top row can consist of corners, hats and previous connections.
+ * @param {[type]} block [description]
+ * @package
+ */
+Blockly.BlockRendering.TopRow = function(block) {
+  Blockly.BlockRendering.TopRow.superClass_.constructor.call(this);
+
+  this.elements = [];
```
rachel-fenichel: `this.elements` has already been created in the `Row` constructor.

core/block_rendering_rewrite/block_render_info.js
```diff
@@ -160,6 +163,35 @@ Blockly.BlockRendering.RenderInfo.prototype.createRows_ = function() {
...
+  var hasHat = this.block_.hat ? this.block_.hat === 'cap' : Blockly.BlockSvg.START_HAT;
+  var hasPrevious = !!this.block_.previousConnection;
+  var prevBlock = this.block_.getPreviousBlock();
+  var squareCorner = !!this.block_.outputConnection ||
+      hasHat || (prevBlock && prevBlock.getNextBlock() == this.block_);
+  this.topRow = new Blockly.BlockRendering.TopRow(this.block_);
+
+  if (squareCorner) {
+    this.topRow.elements.push(
+        new Blockly.BlockRendering.SquareCorner(this.width));
```
rachel-fenichel: Get rid of unused width, here and on other measurables

Issue #2482, opened 2019-05-23
Firefox 67.0 when block drag from toolbox, block not centred to mouse


YannCaron:
After updated by Firefox to 67.0 I experienced a strange behaviour :
When picking block from toolbox, the block is displaced but not centred to the cursor. It move far to the left of it (I estimate about 250px).
I don't know if it is a blockly bug or a firefox one.

NeilFraser:
Confirmed.  Works as expected on FF 66, breaks badly in FF 67.

This affects category toolboxes, not simple toolboxes.  Master and Develop.

Pretty nasty bug.  Thanks for reporting!

NeilFraser:
The root issue is that this is failing to match:
```
'display: block; transform: translate(107px);'.match(Blockly.utils.getRelativeXY.XY_2D_REGEX_)
```

NeilFraser:
PR #2485 fixes this issue.  I appear to be alone in the office today (is there a Blockly party I'm not invited to?), so this might take a little while to land in develop and master.

YannCaron:
Wahou ! Many thanks for the reactivity.
Great, I will wait until the fix is applied to master branch :-) :+1:

NeilFraser:
Patch applied to both develop and master.
* The Blockly team needs to push to the demo server
* I'll handle a recompile and push to Blockly Games later today.
* We also need to ensure that larger users (e.g. Scratch) know about this issue since it affects them and they need to update.

Keeping this issue open.

RoboErikG:
@thisandagain @pelikhan

rachel-fenichel:
@ewpatton

YannCaron:
I've tested on my Ubuntu's Firefox v67.0 (64bits) and it works perfectly now.
Many thanks for your great job :-)

PR #2487, opened 2019-05-23
Fix style matching in Firefox 67 (master cherrypick)

NeilFraser:
Cherrypicks XY regex fix for Firefox 67 (issue #2482) onto master.  Recompile.


RoboErikG:
The travis failure is due to an update on one of the testing dependencies. Having trouble finding the PR that fixed it though. This PR should be fine to merge.

PR #2497, opened 2019-05-24
Changed Colour Field Class Validator to Accept all Documented Values

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Makes it so the class validator accepts any string that can be parsed by [goog.color](https://google.github.io/closure-library/api/goog.color.html).parse. If it can be parsed it converts it to #rrggbb format.

### Reason for Changes

I noticed that the [documentation ](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#colour_picker_field)stated any css colour string was valid, so I wanted to make sure the colour field still worked if people had defined their colours that way.

I did not change the documentation of the field constructor, because I think it's best if people input strings in #rrggbb format, because that's the format validators will need to parse.

### Test Coverage

Added tests for strings: '#AA0', '#aa0', 'rgb(170, 170, 0)', and 'red'.

### Additional Information

N/A

### DIFF
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -32,6 +32,7 @@ goog.require('Blockly.Events.BlockChange');
 goog.require('Blockly.Field');
 
 goog.require('goog.math.Size');
+goog.require('goog.color');
 
 
 /**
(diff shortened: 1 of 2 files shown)


core/field_colour.js
```diff
@@ -176,8 +177,11 @@ Blockly.FieldColour.prototype.render_ = function() {
  * @protected
  */
 Blockly.FieldColour.prototype.doClassValidation_ = function(newValue) {
-  if (Blockly.FieldColour.COLOUR_REGEX.test(newValue)) {
```
NeilFraser: FYI, this changes leaves `Blockly.FieldColour.COLOUR_REGEX` orphaned.  Don't worry about it, I'm removing it in a PR I'm working on.

BeksOmega: Gah! Thank you for noticing!

Issue #2428, opened 2019-05-01
Change Access Level of Field.updateWidth


BeksOmega:
### Problem statement

The updateWidth function is called as part of the rendering of a field. After all of the parts of the field have been shuffled around it figures out how much room they take up and assigns that value to the field's size_ property.

It doesn't really make sense to call this from anywhere besides the render_ function, and I can't think of a reason a developer would want to call it, so having it be public seems unecessary.

This also makes it confusing when you want to create a custom field that has variable /height/ (which is supported!) but you only see an "updateWidth" function.

[Here](https://github.com/google/blockly/search?q=updateWidth&unscoped_q=updateWidth)'s a link to all the places updateWidth gets called, incase anyone is interested.

### Proposal

Since it wouldn't make sense for a developer to call this function externally, I believe it can safely be changed to a private function.

RoboErikG:
Since it's overridden by a few subclasses it'd need to be package or protected, but otherwise sounds reasonable.

Issue #2499, opened 2019-05-25
Some Browsers Don't Properly Import XML with Sub-Block Nodes


BeksOmega:
### Problem statement

If the XML parser tries to deserialize some xml that has nodes inside block tags (i.e. shadow blocks, values, etc) it will sometimes fail, depending on the browser.

This can be seen when opening certain toolbox categories, as well as when importing straight XML through the playground.

### Steps to Reproduce

1. Try to import the below code in the playground.
```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <block type="controls_if" id="L:ym~c^IZzI,G#/PaR(^" x="38" y="88">
    <value name="IF0">
      <block type="logic_boolean" id="JQdsdXtYxW}j~W8k%ke`">
        <field name="BOOL">TRUE</field>
      </block>
    </value>
  </block>
</xml>
```
2. Observe the error.

### Stack Traces

```
0: Object doesn't support property or method 'getAttribute'
xml.js (630,6)
```

### Operating System and Browser

* Windows Internet Explorer 11
* Windows Edge

### Additional Information

I think the problem originates with #2468 If you try and log the value of `Element.TEXT_NODE` inside of the `domToBlockHeadless_` function when you're running on internet explorer or edge, it will return undefined. If you log it when on chrome or firefox it returns 3 (correct).

NeilFraser:
Good grief.  Thanks for this.

Can you check to see if Microsoft browsers know about Node.TEXT_NODE?

NeilFraser:
I've prepared this commit: https://github.com/google/blockly/commit/2c0983911ffcb937bf58f618f1868e46e4fa2e8b  However, I'll wait until we hear whether Microsoft browsers support Node constants before creating a PR.

What's odd is that Element is a child of Node.  Which means that if a property is on Node, then it should also be on Element.

BeksOmega:
> Can you check to see if Microsoft browsers know about Node.TEXT_NODE?

Screen Cap of Edge Console:
![TEXTNODE_Edge](https://user-images.githubusercontent.com/25440652/58382993-284d2380-7f86-11e9-90e8-dbca407e858e.jpg)

Screen Cap of IE 11 Console:
![TEXTNODE_IE](https://user-images.githubusercontent.com/25440652/58383001-30a55e80-7f86-11e9-80ad-92eca66029c5.jpg)

So it looks like they do!

NeilFraser:
Thanks!  PR #2503 created.

[...]

PR #2500, opened 2019-05-25
Fixed How Fields Handle Size

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2428 and some collapse/expand problems

### Proposed Changes

1. Makes it so visibility (return a size of 0) is handled inside getSize instead of inside setVisible.
2. Changes (public) updateWidth to (protected) updateSize_.

### Reason for Changes

1. There were problems with collapsing when it was done the previous way. Now all blocks are expanded and collapsed correctly.
2. Makes it less confusing for outside developers who want to create a field with a dynamic height.

Note that the dropdown field's updateWidth function wasn't getting called from anywhere, so it could be safely removed.

### Test Coverage

FieldsHandleSize

I also tested collapsing and collapsing all of the default categories, and it all looks good.

Tested on:
* Desktop Chrome
* Desktop Firefox

Was not able to test on internet explorer or edge because of #2499

### Additional Information

N/A

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -230,6 +230,9 @@ Blockly.Field.prototype.init = function() {
     return;
   }
   this.fieldGroup_ = Blockly.utils.createSvgElement('g', {}, null);
+  if (!this.isVisible()) {
+    this.fieldGroup_.style.display = 'none';
+  }
   this.sourceBlock_.getSvgRoot().appendChild(this.fieldGroup_);
   this.initView();
   this.initModel();
@@ -240,9 +243,6 @@ Blockly.Field.prototype.init = function() {
  * @package
  */
 Blockly.Field.prototype.initView = function() {
-  if (!this.visible_) {
-    this.fieldGroup_.style.display = 'none';
-  }
   this.borderRect_ = Blockly.utils.createSvgElement('rect',
       {
         'rx': 4,
@@ -386,7 +386,6 @@ Blockly.Field.prototype.setVisible = function(visible) {
   var root = this.getSvgRoot();
   if (root) {
     root.style.display = visible ? 'block' : 'none';
-    this.size_.width = 0;
   }
 };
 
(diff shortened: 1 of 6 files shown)


core/field.js
```diff
@@ -475,15 +474,24 @@ Blockly.Field.prototype.updateColour = function() {
...
-  this.updateWidth();
+  this.updateSize_();
 };
 
 /**
- * Updates the width of the field. This calls getCachedWidth which won't cache
- * the approximated width on IE/Edge when `getComputedTextLength` fails. Once
- * it eventually does succeed, the result will be cached.
+ * Updates the width of the field. Redirects to updateSize_().
+ * @deprecated May 2019 Use Blockly.Field.getCachedWidth() directly instead.
```
rachel-fenichel: Why does this say to use `getCachedWidth` instead of `updateSize_`?

core/field_checkbox.js
```diff
@@ -91,6 +91,14 @@ Blockly.FieldCheckbox.prototype.SERIALIZABLE = true;
  */
 Blockly.FieldCheckbox.prototype.CURSOR = 'default';
 
+/**
+ * Used to tell if the field needs to render. Checkbox fields are statically
```
rachel-fenichel: "Used to tell if the field needs to be rendered the next time the block is rendered.  Checkbox fields are statically sized, and only need to be rendered at initialization."

^ If this is inaccurate, let's find a more accurate way to say it.

rachel-fenichel: Similar/the same wording for other places where you're setting isDirty to false.

core/field_dropdown.js
```diff
@@ -510,28 +510,6 @@ Blockly.FieldDropdown.prototype.renderSelectedText_ = function() {
   this.size_.width = Blockly.Field.getCachedWidth(this.textElement_);
 };
 
-/**
- * Updates the width of the field. Overrides field.prototype.updateWidth to
- * deal with image selections on IE and Edge. If the selected item is not an
- * image, or if the browser is not IE / Edge, this simply calls the parent
- * implementation.
- */
-Blockly.FieldDropdown.prototype.updateWidth = function() {
```
rachel-fenichel: Why did you remove this?

BeksOmega: It doesn't actually get called anywhere, all of the sizing logic is handled in the image and text render functions.

core/field.js
BeksOmega: render_ only supports text fields. updateSize_ is only called by the default render_ function. updateSize_ also only supports text fields.

If you have a text field, there is no need to override render_, and no need to call updateSize_.

If you have an image, or mixed field you need to override render_. In this case updateSize_ will not be helpful, because it only supports text fields. But if it is a mixed field, you may still want to know the size of your text, so you should call getCachedWidth().

core/field_checkbox.js

core/field.js
rachel-fenichel: I see your point, but the redirect still feels wrong to me:
- `updateWidth` and `updateSize` have no return value, but have side effects
- `getCachedWidths` has a return value and no side effects

I suggest:
@deprecated May 2019 Use Blockly.Field.updateSize_() to force an update to the size of the field, or Blockly.Field.getCachedWidth() to check the size of the field..

BeksOmega: Sounds good, Fixed.

PR #2501, opened 2019-05-26
Angle Field Fixes

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

1. Fixes the angle offset property.
2. Angle field now displays wrapped values correctly.
3. Fixed the gauge not displaying correctly at high angle values (like the millions or something, only applies to text input)
4. Angle field now displays zero if given empty text input.

### Reason for Changes

1. It was broken. (The below is a clockwise angle field)
Angle_Clockwise_Offset90

2. It was always displaying between 0 and 360. It's pretty confusing if it changes when you close th editor.
3. This just bugged me.
4. Angle fields should always display a number value (this is different from pure text inputs, which can be empty).

### Test Coverage

I tested lots of combinations of CLOCKWISE, OFFSET, and WRAP. Here are some highlights:
* Clockwise, Offset 90
Angle_Clockwise_Offset90_Good!

* Widdershins, Offset 90
Angle_Wittershins_Offset90

* Widdershins, Wrap 180
Angle_Wittershins_Wrap180

* Widdershins, Wrap 180, Offset 90
Angle_Wittershins_Wrap180_Offset90

And here's it now working with the empty string:
Angle_EmptyString

Tested on:
* Desktop Chrome

### Additional Information

N/A

### DIFF
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -264,6 +264,8 @@ Blockly.FieldAngle.prototype.onMouseMove = function(e) {
   } else if (dy > 0) {
     angle += 360;
   }
+
+  // Do offsetting.
   if (Blockly.FieldAngle.CLOCKWISE) {
     angle = Blockly.FieldAngle.OFFSET + 360 - angle;
   } else {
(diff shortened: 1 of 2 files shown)


core/field_angle.js
```diff
@@ -295,6 +307,7 @@ Blockly.FieldAngle.prototype.updateGraph_ = function() {
   }
   // Always display the input (i.e. getText) even if it is invalid.
   var angleDegrees = Number(this.getText()) + Blockly.FieldAngle.OFFSET;
+  angleDegrees = angleDegrees % 360;
```
NeilFraser: This is a rare opportunity to use the `%=` operator.
```
angleDegrees %= 360;
```

While you're at it, there's another opportunity for a `%=` in `Blockly.FieldAngle.prototype.doClassValidation_`.

BeksOmega: Oooo fun! :D

Fixed

rachel-fenichel:
@NeilFraser does this look good to you now?

PR #2503, opened 2019-05-26
Replace Element constants with Node constants.

NeilFraser:
Element constants don’t exist in IE or Edge.  Node constants do.

(Note that old versions of some browser doesn't have Node constants either.  Can't remember which browser (Netscape 4, IE 5?) but doesn't affect us.  This is why older front-end devs just use raw numbers; we all know that 1 and 3 refer to elements and text respectively.)


rachel-fenichel:
Why was this working before #2468?

NeilFraser:
On Tue, 28 May 2019, 16:14 Rachel Fenichel, <notifications@github.com>
wrote:

> Why was this working before #2468
> <https://github.com/google/blockly/pull/2468>?
>

It wasn't.  #2468 wasn't the source.

The most recent source of these references is here:
https://github.com/google/blockly/commit/834fa6467478ee96a9bfe2ab93c39d57ab5dfd60
And there's another source here:
https://github.com/google/blockly/commit/59a4db61b228563f342fb01de8ab2ed050c58c6a

>

rachel-fenichel:
Okay.  The most recent source hasn't hit master, so that's no surprise.  The other has been on master for a while so I'm a bit surprised that we haven't seen it before.

LGTM, approval.

Issue #2484, opened 2019-05-23
Make Adding Children to a Field's textElement_ Easier


BeksOmega:
### Problem statement

You cannot use the field's abstract render_ function, and also append children to the field's textElement_ (such as the angle field does), without having to re-append that child every time the field renders.

### Expected Behavior

You should be able to structure all of the DOM elements inside initView, and then never worry about them again.

### Actual Behavior

You have to re-structure the DOM whenever the field renders.

### Additional Information

My idea for getting around this would be to create a textGroup_ element that the textElement_ and auxiliary elements could be appended to.

The problem with this is that some of the CSS would also need to be reworked, since it selects the text based on DOM structure instead of class. (.blocklyEditableText>text)

Other ideas would be appreciated as well!

BeksOmega:
I mentioned [here ](https://github.com/google/blockly/pull/2441#discussion_r287131143)that this could be closed if desired, but I was wrong this does need a solution. The angle field needs to be able to safely call its parent's render function so that the htmlInput can be resized. Currently the behavior looks like this:

![TextElement_Angle](https://user-images.githubusercontent.com/25440652/58374029-cdff8480-7eeb-11e9-815f-23ef0d6e623c.gif)

PR #2509, opened 2019-05-27
Reorganized Field View Initialization

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2484 plus some other changes.

### Proposed Changes

1. Adds separate functions for creating border rects and text elements.
2. Adds fieldBindEvents_ and doEventBinding_ functions to the field.
4. Adds tooltip support to all fields.
3. Moves updateEditable, setTooltip, and event binding calls out of initView and into init.
5. Removes the field is editable check from isFieldClick inside gesture.

Field Specific Stuff:
1. Fixes the angle field, with regard to #2484
1. Changes the checkbox to use the textElement_.
2. Removes the textElement_ from the colour field.
3. Reorganizes dropdown initialization so that dom creation happens inside initView and rendering just modifies it.
4. Removes tooltip handling from image.
5. Image onClick handlers use the showEditor_ function instead of having their own event binding.
6. Removes tooltip handling from label.
7. Label fields now call the createTextElement_ funtion.

### Reason for Changes

1. Allows fields that /only/ need a borderRect or /only/ need a textElement to create them without duplicating code.
2. If outside developers override initView they won't have to remember to include the event bindings boilerplate.
4. Eliminates duplicate code, plus it's more fun!
3. See number 2.
5. Image fields aren't editable, but they receive clicks. No reason to have this check.

Field Specific Stuff:
1. Bugs are bad.
1. No reason not to.
2. Useless DOM is bad.
3. More efficient!
4. Eliminates duplicate code.
5. All fields should do things the same way.
6. Eliminates duplicate code.
7. Eliminates duplicate code.

### Test Coverage

Angle Fields: (The odd guage thing is fixed by #2501)
ViewInitialization_Angle

Enabled:
ViewInitialization_Enabled

Shadow:
ViewInitialization_Shadow

Collapsed:
ViewInitialization_Collapsed

Editable:
ViewInitialization_Editable

Tested on:
* Desktop Chrome
* Desktop Firefox

### Additional Information

If anyone has unit tests suggestions I'd love to add some!

Also before the custom field documentation goes up, and probably before the new rendering stuff goes out, I'd love to finalize how you want field sizing to work.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -235,6 +235,9 @@ Blockly.Field.prototype.init = function() {
   }
   this.sourceBlock_.getSvgRoot().appendChild(this.fieldGroup_);
   this.initView();
+  this.updateEditable();
+  this.setTooltip();
+  this.bindEvents_();
   this.initModel();
 };
 
@@ -243,26 +246,54 @@ Blockly.Field.prototype.init = function() {
  * @package
  */
 Blockly.Field.prototype.initView = function() {
+  this.createBorderRect_();
+  this.createTextElement_();
+};
+
+/**
+ * Create a field border rect element. Not to be overridden by subclasses.
+ * Instead modify the result of the function inside initView, or create a
+ * separate function to call.
+ * @protected
+ */
+Blockly.Field.prototype.createBorderRect_ = function() {
   this.borderRect_ = Blockly.utils.createSvgElement('rect',
       {
         'rx': 4,
         'ry': 4,
         'x': -Blockly.BlockSvg.SEP_SPACE_X / 2,
         'y': 0,
-        'height': 16
+        'height': 16,
+        'width': this.size_.width + Blockly.BlockSvg.SEP_SPACE_X
       }, this.fieldGroup_);
+};
+
+/**
+ * Create a field text element. Not to be overridden by subclasses. Instead
+ * modify the result of the function inside initView, or create a separate
+ * function to call.
+ * @protected
+ */
+Blockly.Field.prototype.createTextElement_ = function() {
   this.textElement_ = Blockly.utils.createSvgElement('text',
-      {'class': 'blocklyText', 'y': this.size_.height - 12.5},
-      this.fieldGroup_);
-  var textNode = document.createTextNode('');
-  this.textElement_.appendChild(textNode);
+      {
+        'class': 'blocklyText',
+        'y': this.size_.height - 12.5
+      }, this.fieldGroup_);
+  this.textContent_ = document.createTextNode('');
+  this.textElement_.appendChild(this.textContent_);
+};
 
-  this.updateEditable();
-
-  this.clickTarget_ = this.getSvgRoot();
+/**
+ * Bind events to the field. Can be overridden by subclasses if they need to do
+ * custom input handling.
+ * @protected
+ */
+Blockly.Field.prototype.bindEvents_ = function() {
+  Blockly.Tooltip.bindMouseEvents(this.getClickTarget_());
   this.mouseDownWrapper_ =
       Blockly.bindEventWithChecks_(
-          this.clickTarget_, 'mousedown', this, this.onMouseDown_);
+          this.getClickTarget_(), 'mousedown', this, this.onMouseDown_);
 };
 
 /**
@@ -332,6 +363,14 @@ Blockly.Field.prototype.updateEditable = function() {
     Blockly.utils.removeClass(group, 'blocklyEditableText');
     group.style.cursor = '';
   }
+};
+
+/**
+ * Check whether this field defines the showEditor_ function.
+ * @return {boolean} Whether this field is clickable.
+ */
+Blockly.Field.prototype.isClickable = function() {
+  return !!this.showEditor_ && (typeof this.showEditor_ === 'function');
 };
 
 /**
@@ -481,7 +520,7 @@ Blockly.Field.prototype.updateColour = function() {
  * @protected
  */
 Blockly.Field.prototype.render_ = function() {
-  this.textElement_.textContent = this.getDisplayText_();
+  this.textContent_.nodeValue = this.getDisplayText_();
   this.updateSize_();
 };
 
@@ -806,11 +845,15 @@ Blockly.Field.prototype.onMouseDown_ = function(e) {
 
 /**
  * Change the tooltip text for this field.
- * @param {string|!Element} _newTip Text for tooltip or a parent element to
- *     link to for its tooltip.
+ * @param {string|function|!Element} newTip Text for tooltip or a parent
+ *    element to link to for its tooltip.
  */
-Blockly.Field.prototype.setTooltip = function(_newTip) {
-  // Non-abstract sub-classes may wish to implement this.  See FieldLabel.
+Blockly.Field.prototype.setTooltip = function(newTip) {
+  if (!newTip && newTip !== '') {  // If null or undefined.
+    this.getClickTarget_().tooltip = this.sourceBlock_;
+  } else {
+    this.getClickTarget_().tooltip = newTip;
+  }
 };
 
 /**
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -79,6 +79,20 @@ Blockly.FieldCheckbox.WIDTH = 5;
 Blockly.FieldCheckbox.CHECK_CHAR = '\u2713';
 
 /**
+ * Used to correctly position the check mark.
+ * @type {number}
+ * @const
+ */
+Blockly.FieldCheckbox.CHECK_X_OFFSET = -3;
+
+/**
+ * Used to correctly position the check mark.
+ * @type {number}
+ * @const
+ */
+Blockly.FieldCheckbox.CHECK_Y_OFFSET = 14;
+
+/**
  * Serializable fields are saved by the XML renderer, non-serializable fields
  * are not. Editable fields should also be serializable.
  * @type {boolean}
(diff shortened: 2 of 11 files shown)


core/field.js
```diff
@@ -235,26 +238,64 @@ Blockly.Field.prototype.initView = function() {
...
+          this.getClickTarget_(), 'mousedown', this, this.onMouseDown_);
+  this.doEventBinding_();
+};
+
+/**
+ * Bind events to the field. Can be overridden by subclasses.
+ * No-op by default.
+ * @protected
+ */
+Blockly.Field.prototype.doEventBinding_ = function() {
```
BeksOmega: Thinking about this again, this doesn't really need to be two functions. It'd probably be just as easy for developers to override a protected bindEvents_ function and call the parent (which would do the tooltip and click stuff). That would also make it easier for devs to remove the tooltip and click events if they interfere with the input they're trying to recieve.

Does changing this to one protected function sound ok?

core/field.js
```diff
@@ -474,7 +515,7 @@ Blockly.Field.prototype.updateColour = function() {
  * @protected
  */
 Blockly.Field.prototype.render_ = function() {
-  this.textElement_.textContent = this.getDisplayText_();
```
rachel-fenichel: What does this change do?

core/field_checkbox.js
```diff
@@ -98,19 +98,13 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
...
 
-  // The checkbox doesn't use the inherited text element.
-  // Instead it uses a custom checkmark element that is either visible or not.
-  this.checkElement_ = Blockly.utils.createSvgElement('text',
-      {'class': 'blocklyText blocklyCheckbox', 'x': -3, 'y': 14},
-      this.fieldGroup_);
-  var textNode = document.createTextNode(Blockly.FieldCheckbox.CHECK_CHAR);
-  this.checkElement_.appendChild(textNode);
-  this.checkElement_.style.display = this.value_ ? 'block' : 'none';
+  this.textElement_.setAttribute('x', -3);
```
rachel-fenichel: Turn these two into constants in this file.

core/field_colour.js
```diff
@@ -138,13 +138,10 @@ Blockly.FieldColour.prototype.DROPDOWN_BACKGROUND_COLOUR = 'white';
  * @package
  */
 Blockly.FieldColour.prototype.initView = function() {
-  Blockly.FieldColour.superClass_.initView.call(this);
```
rachel-fenichel: The superclass `initView` function also sets `display: none` if the field is not visible.  That needs to happen for this field as well.

core/field_dropdown.js
```diff
@@ -128,11 +128,17 @@ Blockly.FieldDropdown.prototype.imageJson_ = null;
 Blockly.FieldDropdown.prototype.initView = function() {
   Blockly.FieldDropdown.superClass_.initView.call(this);
 
-  // Add dropdown arrow: "option ▾" (LTR) or "▾ אופציה" (RTL)
-  this.arrow_ = Blockly.utils.createSvgElement('tspan', {}, null);
-  this.arrow_.appendChild(document.createTextNode(this.sourceBlock_.RTL ?
-      Blockly.FieldDropdown.ARROW_CHAR + ' ' :
-      ' ' + Blockly.FieldDropdown.ARROW_CHAR));
+  this.imageElement_ = Blockly.utils.createSvgElement(
+      'image', { 'y':5 }, this.fieldGroup_);
```
rachel-fenichel: Make this 5 a constant.

core/field_dropdown.js
```diff
@@ -441,15 +447,6 @@ Blockly.FieldDropdown.prototype.updateColour = function() {
  * @private
  */
 Blockly.FieldDropdown.prototype.render_ = function() {
-  var child;
```
rachel-fenichel: This goes away because you're creating the image and arrow elements in initView and keeping them around, right?

core/field_dropdown.js
```diff
@@ -465,20 +462,16 @@ Blockly.FieldDropdown.prototype.render_ = function() {
  * @private
  */
 Blockly.FieldDropdown.prototype.renderSelectedImage_ = function() {
-  // Image option is selected.
-  this.imageElement_ = Blockly.utils.createSvgElement('image',
-      {
-        'y': 5,
-        'height': this.imageJson_.height + 'px',
```
rachel-fenichel: Where is the height of the image element getting set now?  Does it ever change?  (I assume width is set/modified in the same places.)

core/gesture.js
```diff
@@ -899,9 +899,7 @@ Blockly.Gesture.prototype.isBlockClick_ = function() {
  * @private
  */
 Blockly.Gesture.prototype.isFieldClick_ = function() {
-  var fieldEditable = this.startField_ ?
-      this.startField_.isCurrentlyEditable() : false;
-  return fieldEditable && !this.hasExceededDragRadius_ && (!this.flyout_ ||
+  return this.startField_ && !this.hasExceededDragRadius_ && (!this.flyout_ ||
```
rachel-fenichel: I'm suspicious of this change.  What happens now if you click on a non-editable field (such as a label)?  It should be a block click, not a field click, but I think that's no longer the case with this change.

rachel-fenichel:
Future field sizing: 
- `getSize` should return the size of the bounding box containing all visible parts of the field.  
- Field bounding boxes are the black boxes here.
- Notice that labels have a tight box, just around the text, while editable things (like the checkbox, colour picker, and angle picker) have a larger box that includes the border rect.
- You can see that the date picker isn't respecting that right now: its reported size doesn't include the size of its bounding box, and the height looks wrong too.  
- The `getCorrectedSize` functions in the `render/collab` branch show places where the reported size of the field doesn't currently match the actual size of the field.
- When we get rid of the old rendering code, I will go through the fields removing `getCorrectedSize` functions and making the fields report their sizes correctly.
- I'm not doing that until we start swapping in the new rendering, because I don't want to break the old rendering (and it's kind of fragile).
- Fields are placed by moving the `svgRoot` of the field (code is [here](https://github.com/google/blockly/blob/render/collab/core/block_rendering_rewrite/block_render_draw.js#L323)).

image

core/field.js
BeksOmega: Fixes #2484

core/field_colour.js
BeksOmega: Fixed by #2500. It moves that check up to the init method. (Sorry, I probably shouldn't have put these up all at once. I was just really bored over memorial day)

core/field_dropdown.js
BeksOmega: Correct.

core/gesture.js
BeksOmega: You are correct.

Options:
1) Revert change. If the image field's onClick parameter is defined set this.EDITABLE to true.
2) Revert change.  Have image field override bindEvents_ instead.

core/field_dropdown.js
BeksOmega: I didn't think there was a need to set the height and width of the image element. It looks like it handles different sizes just fine:

DropdownSizeExample

But I checked the[ docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image#Specific_attributes) since you mentioned it, and they do say a height and width is required, so it's probably just a gap in my knowledge.

Readded setting height and width for safety.

core/field_dropdown.js
BeksOmega: Fixed

core/field_checkbox.js

core/field.js
BeksOmega: Fixed

BeksOmega:
> Future field sizing:  ...

That is super awesome because I believe it means the field's size property can be gotten rid of entirely! Fields should be able to return based on the fieldGroup_ (svgRoot)'s bounding box instead.

Here's an [example branch](https://github.com/google/blockly/compare/render/collab...BeksOmega:fixes/getCorrectedSize).
getCorrectedSize

(Note that the list of jackass fields is now a list of all editable fields)

Again, it would be totally awesome if the size property could be removed. One less thing for outside developers to have to deal with :D This makes me so excited about the new rendering!

core/field_textinput.js
```diff
@@ -275,7 +275,7 @@ Blockly.FieldTextInput.prototype.showInlineEditor_ = function(quietInput) {
  *     which event handlers will be bound.
  * @private
  */
-Blockly.FieldTextInput.prototype.bindEvents_ = function(htmlInput) {
+Blockly.FieldTextInput.prototype.bindInputEvents = function(htmlInput) {
```
rachel-fenichel: If this is private, add an undescore.

core/gesture.js
rachel-fenichel: Option 3 sounds good to me.

rachel-fenichel:
> That is super awesome because I believe it means the field's size property can be gotten rid of entirely! Fields should be able to return based on the fieldGroup_ (svgRoot)'s bounding box instead.

I think you're right!  Size might stick around for caching purposes, but it would be completely internal and the developer wouldn't have to mess with it.

@alschmiedt this might resolve that one pixel you were poking, or at least take the guesswork out of who should own that pixel.

core/gesture.js
BeksOmega: Done!

core/field_textinput.js
BeksOmega: Derp, Fixed

BeksOmega:
Will rebase and fix conflicts.

rachel-fenichel:
Looks good other than the conflicts.  I'll take another look once you rebase and then I'll merge.

core/field_textinput.js
rachel-fenichel: Still need to remove the @private tag :)

BeksOmega: bindInputEvents_ is different than bindEvents_. bindInputEvents_ is just used by the textInput for its htmlInput, while bindEvents_ is for binding to the field proper. Maybe they need better names?

rachel-fenichel: It's fine--I was looking at an outdated copy of the code there.

PR #2515, opened 2019-05-28
Make tests pass by setting up requires/provides correctly

rachel-fenichel:
## The basics

- [ ] I branched from develop

### Resolves

All tests were failing.

### Proposed Changes

- Add appropriate calls to `goog.requires` and `goog.provides`
- Remove explicit `<script>` tags in `playground.html`
- Rebuild

### Reason for Changes

`blockly-uncompressed` didn't contain all necessary files, which caused the test pages to error when loading.

### Test Coverage
Tests should run on travis.  Mocha, generator, and jsunit tests should pass.

Eslint tests still fail because of references to `BRC`.  Renaming that is a future todo.


rachel-fenichel:
Update: I added comments to force eslint to accept `BRC` as a valid name.  We will remove these when we rename BRC.  I've added an issue on our list about it.

PR #2517, opened 2019-05-28
Moved the Date Field to the Dropdown Div, and Restyled It

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

1) Moves the date picker from the widget div to the dropdown div.
2) Removes the automatic closing of the picker upon selection (this comes with some tradeoffs).
3) Makes the picker look butiful (maybe?)
Date_New

Note: The darker colour denotes "Today" and the lighter colour denotes the selected day. It is based on how G-cal's mini-calendar looks:
Date_GCal

### Reason for Changes

1) The date field was sad because it wasn't in the dropdown div club.
2) Makes it so the dropdown div works consistently.
3) Sticking with the old UI meant one of two options had to be picked:
Date_BorderDefault
Date_BorderlessDefault
Both of which look ugly (imo).

### Test Coverage

The closure date picker is missing some functionality that we need, which is the ability to undo a date selection. Because there's no perfect way to do this, the date field has some quirky behavior when dealing with null validators. But as no one uses date fields, and especially no one validates them, I think this is ok.

Closing and Reopening:
Date_CloseReopen

Today Button:
Date_TodayButton

Correctly Updating When Month Changes 1:
Date_HidingMonthChanges

Correctly Updating when Month Changes 2:
Date_ChangingDayMonthChanges

Not Updating for Nulls:
Date_NullsNotReacting

Validated Values Updating:
Date_DifUpdating

And here's the bit of odd behavior I mentioned. If you are looking at a different month than the one containing the currently selected date, and you select a new date that validates to null, it will bump you back to the old month:
Date_NullsChangesMonth

There's no reasonable way to get around this with closure's[ current DatePicker API](https://google.github.io/closure-library/api/goog.ui.DatePicker.html).

Tested on:
 * Desktop Chrome

### Additional Information

N/A

### DIFF
--- a/core/field_date.js
+++ b/core/field_date.js
@@ -83,6 +83,22 @@ Blockly.FieldDate.prototype.SERIALIZABLE = true;
 Blockly.FieldDate.prototype.CURSOR = 'text';
 
 /**
+ * Border colour for the dropdown div showing the date picker. Must be a CSS
+ * string.
+ * @type {string}
+ * @private
+ */
+Blockly.FieldDate.prototype.DROPDOWN_BORDER_COLOUR = 'silver';
+
+/**
+ * Background colour for the dropdown div showing the date picker. Must be a
+ * CSS string.
+ * @type {string}
+ * @private
+ */
+Blockly.FieldDate.prototype.DROPDOWN_BACKGROUND_COLOUR = 'white';
+
+/**
  * Close the colour picker if this input is being deleted.
  */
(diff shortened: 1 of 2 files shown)


alschmiedt:
Looks so much better than the previous date picker : )

BeksOmega:
> Looks so much better than the previous date picker : )

Hehe thank you so much! Normally I hate doing the css stuff, but being able to copy off the google maps made it bearable lol

PR #2520, opened 2019-05-29
Call getSize in getCorrectedSize

rachel-fenichel:
## The basics

- [x] I branched from `render/collab`
- [x] My pull request is against `render/collab`

### Resolves

Changing the field value didn't update the appearance of the field, even though the value was changed internally.

### Proposed Changes

Call `getSize` from `getCorrectedSize`

### Reason for Changes

`getSize` rerenders the field if necessary.  I don't want to duplicate its internal logic, especially since that's an active area of code, so I'm just calling it here.

Another option would be to call `getSize` and then `getCorrectedSize` in sequence in `measurables.js`.  That would put it in only one place, which would be nice, but it's also a weird place for it.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -563,6 +563,11 @@ Blockly.Field.prototype.getSize = function() {
   return this.size_;
 };
 
+/**
+ * Get the size of the visible field, as used in new rendering.
+ * @return {!goog.math.Size} The size of the visible field.
+ * @package
+ */
 Blockly.Field.prototype.getCorrectedSize = function() {
   return this.getSize();
 };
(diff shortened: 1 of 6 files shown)


RoboErikG:
What's the plan for getSize/getCorrectedSize once the new renderer is the default? What does the eventual cleanup look like?

rachel-fenichel:
Eventual cleanup plan:
For every field that I needed to write `getCorrectedSize`, there's something wonky in the current rendering that leads to a weird offset.  I will go through one field at a time and simultaneously remove the offset and the call to `getCorrectedSize`.

PR #2521, opened 2019-05-29
Fix checkbox rendering

rachel-fenichel:
### Resolves
It looked like this before:
image

### Proposed Changes

- Add checkboxes to the list of jackass fields.
- Override `getComputedSize` to include the size of the border rect.

### Reason for Changes

Checkboxes were allocated too little space, and offset when rendered.

### Test Coverage
test_fields_checkbox screenshot diff now passes.


core/block_rendering_rewrite/block_render_draw.js
```diff
@@ -288,7 +288,8 @@ Blockly.BlockRendering.Drawer.prototype.drawInternals_ = function() {
 Blockly.BlockRendering.Drawer.prototype.dealWithJackassFields_ = function(field) {
```
RoboErikG: Wasn't there suppose to be a TODO to remove this method before it goes into master?

PR #2522, opened 2019-05-29
Adds minimum block height

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Problem with some rows not having the correct padding on the bottom row.

### Proposed Changes
Add height to the bottom row if the current height does not meet the minimum heigh requirement.

### Test Coverage

Tested on:


core/block_rendering_rewrite/block_render_info.js
```diff
@@ -639,6 +639,12 @@ Blockly.BlockRendering.RenderInfo.prototype.finalize_ = function() {
   this.blockBottom = yCursor;
```
rachel-fenichel: Check whether we're still using `blockBottom`.  If we are, your change to the `yCursor` needs to be above this line.  If not, add an item to the gnosis to delete it.

alschmiedt: Doesn't look like we are using it. Will add a task for cleanup in gnosis.

rachel-fenichel:
Looks good, please check why the tests are failing.

alschmiedt:
Ran tests locally and both passed. Looks like it is not related to this change so will merge and continue investigating.

PR #2523, opened 2019-05-30
Fix 19 warnings in theme-related code.

### LINKED ISSUE #2414: getAllBlockStyles Annotation Error?
### Expected Behavior

The annotation and the return type match.
A) getAllBlockStyles would return an {Array.\<String\>} as annotated.
or B) Annotation would say getAllBlockStyles returns {Object.<string, Blockly.BlockStyle>}.

### Actual Behavior

The annotation and the return type do not match.

### Steps to Reproduce

See [here](https://github.com/google/blockly/blob/9c4f9d6c1693c3c25160f008d3eaa2539127c9e8/core/theme.js#L56).

NeilFraser:
Resolves #2414.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -228,6 +228,13 @@ Blockly.Block.obtain = function(workspace, prototypeName) {
 Blockly.Block.prototype.data = null;
 
 /**
+ * Don't automatically add STATEMENT_PREFIX and STATEMENT_SUFFIX to generated
+ * code.  These will be handled manually in this block's generators.
+ * @type {boolean}
+ */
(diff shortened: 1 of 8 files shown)


specc:
@NeilFraser your changes made Blockly.Themes.Classic to become undefined . Can you please take a look at it ?

NeilFraser:
@specc When I open tests/playground.html in a browser I can still switch between the three themes.  And from the console `Blockly.Themes.Classic` returns an object.

Can you file an issue with steps to recreate what you are seeing?  Thanks.

PR #2537, opened 2019-06-04
Block min width; notch offsets

rachel-fenichel:
Tests passing: 65/109 -> 77/109 in LTR

- Increase padding between an icon and end of row (changes min width)
- Set padding between a notch (prev or next) and the end of a row (changes min width)
- Adjust padding between a notch (prev or next) and the beginning of the row (addresses misalignment)

### DIFF
--- a/core/block_rendering_rewrite/block_render_info.js
+++ b/core/block_rendering_rewrite/block_render_info.js
@@ -316,7 +316,7 @@ Blockly.BlockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
     if (prev.isField() && prev.isEditable) {
       return BRC.MEDIUM_PADDING;
     }
-    // A block with just an icon is skinny.  Make it more substantial.
+    // Padding at the end of an icon-only row to make the block shape clearer.
     if (prev.isIcon()) {
       return (BRC.LARGE_PADDING * 2) + 1;
     }
@@ -383,10 +383,14 @@ Blockly.BlockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
     }
   }
 
-  // Spacing between a hat and a corner
   if (prev.isSquareCorner()) {
+    // Spacing between a hat and a corner
     if (next.isHat()) {
       return BRC.NO_PADDING;
+    }
+    // Spacing between a square corner and a previous or next connection
+    if (next.isPreviousConnection() || next.isNextConnection()) {
+      return BRC.NOTCH_OFFSET_LEFT;
     }
   }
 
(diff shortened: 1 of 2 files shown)


core/block_rendering_rewrite/block_render_info.js
```diff
@@ -316,13 +316,16 @@ Blockly.BlockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
     if (prev.isField() && prev.isEditable) {
       return BRC.MEDIUM_PADDING;
     }
+    // A block with just an icon is skinny.  Make it more substantial.
```
RoboErikG: What if it's two or three icons? Is this adding necessary padding after the last icon or just bumping up the minimum block length? If the latter, should the block minimum width be bigger even for empty blocks?

core/block_rendering_rewrite/block_render_info.js
```diff
@@ -389,8 +392,10 @@ Blockly.BlockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
 
   // Spacing between a rounded corner and a previous or next connection
   if (prev.isRoundedCorner()){
```
RoboErikG: Would it make sense to change the squared corners so they're the same size as a rounded corner? It seems odd to have special spacing for one and not the other.

Also, this block is splitting two checks for prev.isSquareCorner() that should be combined.

rachel-fenichel: I just tried making squared corners have width (previously they had no width, and all width was taken up in the spacer instead).
The problem is that a square corner can come before a hat (in which case the corner should have zero width) or before a notch (in which case it should have non-zero width).  The corner's width should not depend on what comes before or after it, because that moves the spacing logic out of this single function.

I'll combine the isSquareCorner checks.

I think that once I start testing with more squared corners (stacked blocks) I'll have to change the square corner part of this as well, to have the slight difference between previous and next offsets.

RoboErikG: sgtm.

core/block_rendering_rewrite/block_render_info.js
rachel-fenichel: This is adding padding after the last icon no matter what.  I think that's the right thing to do, because then a block with lots of icons still noticeably looks like a block instead of a blob.

Here's (some of) the code that controls this from the old rendering:
```
  inputRows.rightEdge = iconWidth + Blockly.BlockSvg.SEP_SPACE_X * 2;
  if (this.previousConnection || this.nextConnection) {
    inputRows.rightEdge = Math.max(inputRows.rightEdge,
        Blockly.BlockSvg.NOTCH_WIDTH + Blockly.BlockSvg.SEP_SPACE_X);
  }
```

Min width for empty blocks: Empty blocks look like this right now:
image

which matches the old rendering and (imo) has enough width to show that it's a block and not a sliver.

RoboErikG: sgtm. Just update the comment, then.

rachel-fenichel: Done.

PR #2539, opened 2019-06-04
Fix text y location on statement inputs

rachel-fenichel:
In current rendering, text on inline inputs and statement inputs is placed a set distance below the top of the row.  I had two different constants for this, but that was masking a different error.

If/when we switch to vertically centering the fields in a row, this goes away entirely.


rachel-fenichel:
Tests passing: 81/109

Issue #2535, opened 2019-06-03
Shadow blocks shouldn't hide blocklyPathDark


rachel-fenichel:
Shadow blocks look too small because their dark path is hidden (`display: none`) instead of displaying with the shadow color.

This shows up on develop but not master.

I don't know how this will interact with themes.

RoboErikG:
My vote is for setting the border colour using [getColourShadow](https://github.com/google/blockly/blob/develop/core/block.js#L889) for now.

PR #2541, opened 2019-06-04
Account for shadow in block size.

rachel-fenichel:
Changes:
- Add the shadow/dark path size to the reported size of all blocks (in block_render_draw)
- Vertically align inline inputs to match classic rendering (in block_render_info)
- Account for the shadow/dark path in the size of an inline input (in measurables.js)
- Use a constant for the size of the dark path, to make it easier to later remove it

One additional test passes.

Diffs looked like this:
image

Now they look like this:
image

Note that the remaining error here is an instance of #2535:
image


rachel-fenichel:
Tests passing: 82/109

PR #2542, opened 2019-06-04
Add tests with external and inline inputs; set shadow colour

rachel-fenichel:
Changes:
- Add two new test cases (see below)
- Set shadow colour of block correctly (makes 3 tests pass, total 85/111 passing)

@alschmiedt if this looks like a good fix for the shadow block colour, I can make a similar PR against develop.

New test cases: 
image

image


alschmiedt:
Yup looks good to me!

PR #2546, opened 2019-06-05
Remove goog.color

NeilFraser:
Replace `goog.color` with our own `Blockly.utils.colour` in order to use the correct spelling.

Each commit in this PR can be viewed separately.


RoboErikG:
Something in block_test -> test_set_style() broke. I don't see where the problem is from the code though...

NeilFraser:
> Something in block_test -> test_set_style() broke. I don't see where the problem is from the code though...

When Blockly round-trips colours, they now come back lowercase hex.  Updated test.

PR #2547, opened 2019-06-05
Extract connection placement into separate functions

rachel-fenichel:
Part of making multi-block tests pass is getting the connections to be placed right.

This mostly preserves the previous code, but pulled out so that we can easily compare (e.g.) previous and next placement.


rachel-fenichel:
Actually moving the connections and tightening connections is done with an existing call to renderMoveConnections in BlockSvg, so this just has to tell them where to be relative to the top left corner of the block.

PR #2549, opened 2019-06-06
Fix tree traversal to work for multiple stacks

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Tree traversal was not working for multiple stacks when a stack was not valid. 

### Proposed Changes
If the sibling or parent node is not valid traverse the sub tree to look for a valid node. We don't want to traverse the tree if we are at the top workspace node. 

### Reason for Changes
Before if the node was not valid then we skipped the sub tree underneath it and just went to the next out node. 

### Test Coverage

Tested on:

### DIFF
--- a/demos/keyboard-nav/index.html
+++ b/demos/keyboard-nav/index.html
@@ -215,12 +215,14 @@
       } else if (newNode) {
         return treeTraversal(newNode);
       } else {
-        var siblingOrParent = findSiblingOrParent(node);
+        var siblingOrParent = findSiblingOrParent(node.out());
         if (validNode(siblingOrParent)) {
           return siblingOrParent;
         } else if (siblingOrParent
           && siblingOrParent.getType() !== Blockly.ASTNode.types.WORKSPACE) {
           return treeTraversal(siblingOrParent);
+        } else {
+          return null;
         }
       }
     }
--- a/demos/keyboard-nav/index.html
+++ b/demos/keyboard-nav/index.html
@@ -190,8 +190,9 @@
       if (!node) {
         return null;
       }
-      if (validNode(node.next())) {
-        return node.next();
+      var nextNode = node.next();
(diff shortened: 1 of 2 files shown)


demos/keyboard-nav/index.html
```diff
@@ -214,7 +215,13 @@ <h1><a href="https://developers.google.com/blockly/">Blockly</a> &gt;
       } else if (newNode) {
         return treeTraversal(newNode);
       } else {
-        return findSiblingOrParent(node);
+        var siblingOrParent = findSiblingOrParent(node);
```
RoboErikG: Hmm...Just a note, to get to this line I think node.in() and node.next() had to return null, so you could pass in node.out() to save an iteration.

demos/keyboard-nav/index.html
```diff
@@ -214,7 +215,13 @@ <h1><a href="https://developers.google.com/blockly/">Blockly</a> &gt;
...
+        if (validNode(siblingOrParent)) {
+          return siblingOrParent;
+        } else if (siblingOrParent
+          && siblingOrParent.getType() !== Blockly.ASTNode.types.WORKSPACE) {
+          return treeTraversal(siblingOrParent);
+        }
```
RoboErikG: Should add an explicit else return null.

tests/mocha/navigation_test.js
```diff
@@ -209,17 +209,6 @@ suite('Navigation', function() {
           Blockly.Navigation.STATE_WS);
     });
 
-    test('Add block to workspace with a marked workspace node', function() {
```
RoboErikG: Why did this test break?

alschmiedt: We removed the function that this test was testing a while ago and I think we just never removed this test.

PR #2552, opened 2019-06-06
Remove a bunch more Closure

NeilFraser:
Each commit may be viewed individually.


core/flyout_base.js
```diff
@@ -817,25 +817,23 @@ Blockly.Flyout.prototype.placeNewBlock_ = function(oldBlock) {
   var flyoutOffsetPixels = this.workspace_.getOriginOffsetInPixels();
 
   // The position of the old block in flyout workspace coordinates.
-  var oldBlockPosWs = oldBlock.getRelativeToSurfaceXY();
-
+  var oldBlockPos = oldBlock.getRelativeToSurfaceXY();
   // The position of the old block in pixels relative to the flyout
   // workspace's origin.
-  var oldBlockPosPixels = oldBlockPosWs.scale(this.workspace_.scale);
+  oldBlockPos.scale(this.workspace_.scale);
```
rachel-fenichel: Please put the explicit variable names back.  This code is deliberately overly clear, because debugging it goes wrong so quickly.

Here, and in contextmenu.js.

core/scrollbar.js
```diff
@@ -27,10 +27,10 @@
 goog.provide('Blockly.Scrollbar');
 goog.provide('Blockly.ScrollbarPair');
 
+goog.require('Blockly.utils.Coordinate');
```
rachel-fenichel: Nit: fix order

core/workspace_comment_svg.js
```diff
@@ -26,16 +26,16 @@
 
 goog.provide('Blockly.WorkspaceCommentSvg');
 
+goog.require('Blockly.utils.Coordinate');
```
rachel-fenichel: Nit: fix order

core/workspace_svg.js
```diff
@@ -30,6 +30,7 @@ goog.provide('Blockly.WorkspaceSvg');
 //goog.require('Blockly.BlockSvg');
 goog.require('Blockly.ConnectionDB');
 goog.require('Blockly.constants');
+goog.require('Blockly.utils.Coordinate');
```
rachel-fenichel: Nit: fix order

core/ws_comment_events.js
```diff
@@ -30,13 +30,12 @@ goog.provide('Blockly.Events.CommentCreate');
 goog.provide('Blockly.Events.CommentDelete');
 goog.provide('Blockly.Events.CommentMove');
 
+goog.require('Blockly.utils.Coordinate');
```
rachel-fenichel: Nit: fix order.

NeilFraser:
Here's a summary of the rectangle changes that span across four commits.  Previously we were using three different types of data structures to describe rectangles:
1) Edges: top, bottom, left, right.
2) Size: top, left, height, width.
3) Corners: top-left coordinate, bottom-right coordinate. 

As a result of these commits, we are now using only option 1.  This means that Blockly.utils.Rect is used in more places.

core/comment.js
```diff
@@ -31,7 +31,7 @@ goog.require('Blockly.Events');
 goog.require('Blockly.Events.BlockChange');
 goog.require('Blockly.Events.Ui');
 goog.require('Blockly.Icon');
-goog.require('Blockly.utils');
+goog.provide('Blockly.utils.dom');
```
rachel-fenichel: This should not be a provide.

core/field_angle.js
```diff
@@ -28,7 +28,7 @@ goog.provide('Blockly.FieldAngle');
 
 goog.require('Blockly.DropDownDiv');
 goog.require('Blockly.FieldTextInput');
-goog.require('Blockly.utils');
+goog.provide('Blockly.utils.dom');
```
rachel-fenichel: Also shouldn't be a provide

rachel-fenichel:
Note: provides were fixed in the following commit.

LGTM

BeksOmega:
By the way based on my [conversation with Amber](https://groups.google.com/forum/#!topic/blockly/kWZDQD7-A9w), I think moving svg utils into utils.dom may break a not insignificant amount of people.

PR #2554, opened 2019-06-07
Render/bottom inline highlight

rachel-fenichel:
- Fix highlight on ltr  inline inputs (broken in my last commit)
- Fix height and remove offset constant
- Move next connection based on correct height

Tested in both ltr and rtl.


core/block_rendering_rewrite/block_render_draw.js
```diff
@@ -413,8 +413,8 @@ Blockly.BlockRendering.Drawer.prototype.positionNextConnection_ = function() {
 
   if (bottomRow.hasNextConnection) {
     var connX = (this.info_.RTL ? -BRC.NOTCH_OFFSET_LEFT : BRC.NOTCH_OFFSET_LEFT);
```
alschmiedt: nit : remove parenthesis here for consistency.

PR #2560, opened 2019-06-10
Change alignment on previous and next connections

rachel-fenichel:
Makes the logic similar for square and round corners.

Removes extra constant.

I don't know where the 0.5 comes from.

Tests passing:
87/112 in LTR
77/112 in RTL

### DIFF
--- a/core/icon.js
+++ b/core/icon.js
@@ -207,7 +207,10 @@ Blockly.Icon.prototype.getIconLocation = function() {
 };
 
 /**
- * Get the size of the icon, as used for rendering.
+ * Get the size of the icon as used for rendering.
+ * This differs from the actual size of the icon, because it bulges slightly
+ * out of its row rather than increasing the height of its row.
+ * TODO (#2562): Remove getCorrectedSize.
  * @return {!goog.math.Size} Height and width.
  */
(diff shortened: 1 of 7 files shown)


rachel-fenichel:
Other changes:
- field sizing tweaks 
- get rid of some magic numbers
- add some functions to icons and inputs so that measurement code can call them

Tests passing:
- 96/112 LTR
- 85/112 RTL
- Most of the remaining non-dummy-input problems are acceptable changes
- I started a column in gnosis to describe test output differences that I don't plan to fix, where the come from, and why

core/icon.js
```diff
@@ -205,3 +205,12 @@ Blockly.Icon.prototype.computeIconLocation = function() {
 Blockly.Icon.prototype.getIconLocation = function() {
   return this.iconXY_;
 };
+
+/**
+ * Get the size of the icon, as used for rendering.
+ * @return {!goog.math.Size} Height and width.
+ */
+Blockly.Icon.prototype.getCorrectedSize = function() {
+  return new goog.math.Size(
```
alschmiedt: Can we add an explanation for why we have the -2?

PR #2567, opened 2019-06-11
Add a row after a dummy input.

rachel-fenichel:
Before:

image
 
After:
image


core/block_rendering_rewrite/block_render_info.js
```diff
@@ -221,6 +221,14 @@ Blockly.BlockRendering.RenderInfo.prototype.createBottomRow_ = function() {
 };
 
 
+/**
+ * Add an input element to the active row, if needed, and record the type of the
+ * input on the row.
+ * @param {Blockly.Input} input The input to record information about.
```
NeilFraser: Can't be null, should have a !
Same applies to the param below.

rachel-fenichel: Done

PR #2568, opened 2019-06-11
Added Trashcan Tests

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Adds tests to documenta trashcan behavior.

### Reason for Changes

To keep Beka from breaking things.

### Test Coverage

This is tests silly!

### Additional Information

The tests document the current behavior of the trashcan, but I don't think all of that behavior is good. I left TODO's in places where I think behavior should be changed.

Also this might not cover everything that needs to be tested. If you have suggestions I'd love to add more tests.

### DIFF
--- a/tests/mocha/trashcan_test.js
+++ b/tests/mocha/trashcan_test.js
@@ -102,19 +102,6 @@ suite("Trashcan", function() {
       );
       chai.assert.equal(this.trashcan.contents_.length, 1);
     });
-    test("No Coords - Coords", function() {
-      sendDeleteEvent(
-          '<xml>' +
-          '  <block type="dummy_type"/>' +
-          '</xml>'
-      );
-      sendDeleteEvent(
-          '<xml>' +
-          '  <block type="dummy_type" x="10" y="10"/>' +
-          '</xml>'
-      );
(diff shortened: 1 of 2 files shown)


tests/mocha/trashcan_test.js
```diff
@@ -0,0 +1,561 @@
...
+    this.setLidStub = sinon.stub(this.trashcan, 'setLidAngle_');
+  });
+  teardown(function() {
+    this.setLidStub.restore();
+    this.trashcan = null;
+  });
+
+  suite("Events", function() {
+    test("Delete", function() {
+      sendDeleteEvent(
```
rachel-fenichel: `sendDeleteEvent` can add the `<xml>` and `</xml>` tags to remove duplication.

BeksOmega: Fixed.

BeksOmega:
> Create an issue for those TODOs and tag it in the comment as `TODO (#issueNumber)` so we don't completely lose track of them.

Done. When I was creating the issue I noticed that the problems with the `disabled` `editable` and `movable` attributes do not actually occure, because a block will never serialize to `disabled="false"` when deleted, so I removed those test cases.

PR #2571, opened 2019-06-12
Resolve 13 warnings

NeilFraser:
Some type annotations, some redeclared variables.  Note the removal of copy-pasted validation code in Field.prototype.setValue.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -736,7 +736,7 @@ Blockly.Field.prototype.setValue = function(newValue) {
 
   var validatedValue = this.doClassValidation_(newValue);
   // Class validators might accidentally forget to return, we'll ignore that.
-  newValue = this.validate_(newValue, validatedValue);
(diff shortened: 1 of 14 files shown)


core/field.js
```diff
@@ -782,6 +770,27 @@ Blockly.Field.prototype.setValue = function(newValue) {
...
 };
 
+/**
+ * Process the result of validation.
+ * @param {*} newValue New value.
+ * @param {*} validatedValue Validated value.
+ * @return {*} New value, or an Error object.
+ * @private
+ */
+Blockly.Field.prototype.validate_ = function(newValue, validatedValue) {
```
BeksOmega: Looks like a good change!

I might rename this to something like "processValidation_" though since I believe validate_ used to be used inside the text input field, and may conflict.

NeilFraser: Good call.  Done.

PR #2577, opened 2019-06-14
Added Sending the Bad Value to doValueInvalid

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

doValueInvalid_() now recieves the invalid value.

### Reason for Changes

Makes it easier for fields to display invalid values if they wish.

This was not added previously because the only core field that uses doValueInvalid is the text input field, and its display value is stored on the htmlInput. But when I was working on the custom fields demo I found that this was useful.

### Test Coverage

N/A

### Additional Information

N/A

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -115,9 +115,12 @@ Blockly.FieldTextInput.prototype.doClassValidation_ = function(newValue) {
  * Called by setValue if the text input is not valid. If the field is
  * currently being edited it reverts value of the field to the previous
  * value while allowing the display text to be handled by the htmlInput_.
+ * @param {*} _newValue The input value that was determined to be invalid.
+ *    This is not used by the text input because its display value is stored on
+ *    the htmlInput_.
  * @protected
  */
-Blockly.FieldTextInput.prototype.doValueInvalid_ = function() {
+Blockly.FieldTextInput.prototype.doValueInvalid_ = function(_newValue) {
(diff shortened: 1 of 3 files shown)


core/field_textinput.js
```diff
@@ -115,9 +115,12 @@ Blockly.FieldTextInput.prototype.doClassValidation_ = function(newValue) {
  * Called by setValue if the text input is not valid. If the field is
  * currently being edited it reverts value of the field to the previous
  * value while allowing the display text to be handled by the htmlInput_.
+ * @param {*} _newValue The input value that was determined to be invalid.
+ *    This is not used by the text input because its display value is stored on
+ *    the htmlInput_.
  * @protected
  */
-Blockly.FieldTextInput.prototype.doValueInvalid_ = function() {
+Blockly.FieldTextInput.prototype.doValueInvalid_ = function(_newValue) {
```
NeilFraser: What does this underscore signify?  I'm not familiar with such a naming scheme.

BeksOmega: I believe that it's just a thing for the eslint, so that it doesn't freak out about an unused variable.

alschmiedt: Actually thinking about this could you change _newValue to something like _invalidValue? I think it is a bit more clear. sorry for the late comment!

BeksOmega: Makes sense! Can do =)

PR #2580, opened 2019-06-20
Fixed bug block is not unhighlighted

NagaoKyota:
## The basics

- [x] I branched from develop

### Resolves

N/A

### Proposed Changes

The only current-executed blocks should be highlighted but are not cleared so finally all the blocks are lit on only when they run the program on IE10/11.

after

### Reason for Changes

In Japan many online materials of STEM classes for elementary schools are based on Google Blockly, so the case where children use Goolgle Blockly is not rare. Moreover most of Japanese schools are not equipped with Chrome/ Microsoft Edge nor Safari. They use Internet Explorer, even though IE has not supported. 
We would like to make it clear to children that which code is to debag. 

https://blockly-demo.appspot.com/static/demos/interpreter/step-execution.html
before

### Test Coverage

Tested on:
* Desktop Chrome
* Desktop Firefox
* Desktop Safari
* Desktop Opera
* Windows Internet Explorer 10
* Windows Edge

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1191,11 +1191,10 @@ Blockly.BlockSvg.prototype.setHighlighted = function(highlighted) {
   if (highlighted) {
     this.svgPath_.setAttribute('filter',
         'url(#' + this.workspace.options.embossFilterId + ')');
-    this.svgPathLight_.style.display = 'none';
   } else {
-    this.svgPath_.removeAttribute('filter');
-    delete this.svgPathLight_.style.display;
+    this.svgPath_.setAttribute('filter', 'none');
   }
+  this.svgPathLight_.style.display = 'none';
 };
 
 /**
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1194,7 +1194,7 @@ Blockly.BlockSvg.prototype.setHighlighted = function(highlighted) {
   } else {
     this.svgPath_.setAttribute('filter', 'none');
   }
-  this.svgPathLight_.style.display = 'none';
+  this.svgPathLight_.style.display = 'inline';
 };
 
 /**


NagaoKyota:
I signed it!

core/block_svg.js
```diff
@@ -1191,11 +1191,10 @@ Blockly.BlockSvg.prototype.setHighlighted = function(highlighted) {
   if (highlighted) {
     this.svgPath_.setAttribute('filter',
         'url(#' + this.workspace.options.embossFilterId + ')');
-    this.svgPathLight_.style.display = 'none';
   } else {
-    this.svgPath_.removeAttribute('filter');
-    delete this.svgPathLight_.style.display;
+    this.svgPath_.setAttribute('filter', 'none');
   }
+  this.svgPathLight_.style.display = 'none';
```
NeilFraser: This line deletes the top-left highlight on every block when executed and never restores it -- as can be seen in your demo.

If 'delete' isn't working, then set the display to 'inline'.

NagaoKyota: Thanks for the review 😄

> If 'delete' isn't working, then set the display to 'inline'.

In https://github.com/google/blockly/pull/2580/commits/09b8389d7c9e1ed47f042a5f141e24bb0b0ceac8, change the display style 'none' to 'inline'.

NagaoKyota:
https://github.com/google/blockly/pull/2580/files#diff-fd04c06f310c3ed628154e399e570c9cR1197
changed the display style `none` to `inline`.

NeilFraser:
Thanks for pointing out the bug in IE, and providing a solution.  I've submitted PR #2598 which properly handles making the highlight path appear and disappear.

PR #2581, opened 2019-06-21
Pull Request: Split showContextMenu_

amber-cd:
## The basics

- [ x ] I branched from develop
- [ x ] My pull request is against develop
- [ x ] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

## The details
Split showContextMenu_ into generateContextMenu_ and showContextMenu_.

### Resolves
No issue created, but I can create one if that would make for easier tracking.

### Proposed Changes
Split showContextMenu_ into generateContextMenu_ and showContextMenu_.

### Reason for Changes

This allows for custom Blockly forks to easily extend the context menu options available on all blocks by just doing something like

```
var oldGenerateContextMenu_ = Blockly.BlockSvg.prototype.generateContextMenu_;
Blockly.BlockSvg.prototype.generateContextMenu_ = function() {
  var menuOptions = oldGenerateContextMenu_() || [];
  // Push other options into menuOptions
  return menuOptions;
}
```

Rather than having to modify the core code in such a way as to cause unnecessary maintenance overhead when upgrading Blockly versions, which is our use case.

### Test Coverage

Tested on:
* Desktop Chrome

### Additional Information

This is a fairly minor change but one which will make it a lot easier when developers need to extend the context menu options for all blocks, rather than only for individual blocks.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -633,12 +633,8 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
   }
 };
 
-/**
- * Show the context menu for this block.
- * @param {!Event} e Mouse event.
- * @private
- */
-Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
+Blockly.BlockSvg.prototype.generateContextMenu_ = function() {
+  
   if (this.workspace.options.readOnly || !this.contextMenu) {
     return;
   }
@@ -729,8 +725,21 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
     this.customContextMenu(menuOptions);
   }
 
-  Blockly.ContextMenu.show(e, menuOptions, this.RTL);
-  Blockly.ContextMenu.currentBlock = this;
+  return menuOptions
+}
+
+/**
+ * Show the context menu for this block.
+ * @param {!Event} e Mouse event.
+ * @private
+ */
+Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
+  var menuOptions = this.generateContextMenu_();
+  
+  if(menuOptions && menuOptions.length) {
+    Blockly.ContextMenu.show(e, menuOptions, this.RTL);
+    Blockly.ContextMenu.currentBlock = this;  
+  }
 };
 
 /**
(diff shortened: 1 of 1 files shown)


core/block_svg.js
```diff
@@ -634,11 +634,12 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
 };
 
 /**
- * Show the context menu for this block.
- * @param {!Event} e Mouse event.
+ * Generate the context menu for this block.
  * @private
```
RoboErikG: This should be protected instead of private since it's intended to be overwritten.

amber-cd: Hrm. I'm actually not sure if protected method standard is to have an underscore or not... is this what you were expecting?

RoboErikG: We pretty consistently leave the underscore on there for protected methods in Blockly. I'm not sure what the wider style norms are, though.

rachel-fenichel: Sorry to butt in, but that's not quite true.

If a protected method has an underscore in Blockly, it means it used to be marked private but was obviously part of our API, so we changed the annotation without changing the name.

New protected methods have no underscore.

RoboErikG: Thanks for the correction, Rachel. =) Merging as is, then.

core/block_svg.js
```diff
@@ -634,11 +634,12 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
 };
 
 /**
- * Show the context menu for this block.
- * @param {!Event} e Mouse event.
- * @private
+ * Generate the context menu for this block.
+ * @protected
+ * @returns {Array} Context menu options
```
NeilFraser: The JSDoc doesn't match the return value.  Either change this annotation to
```
{!Array.<!Object>|undefined}
```
or (in my opinion this would be better) change the annotation to 
```
{Array.<!Object>}
```
and change the return on 644 to be
```
return null;
```

Also, drop the blank line on 642.

Otherwise, looks good to me.

amber-cd: This PR looks like it was already merged and closed; would you like me to make a new PR with those changes, or?

rachel-fenichel: New PR, please.

NeilFraser: I've just folded these minor nits into PR #2598 which also edits this file.

PR #2589, opened 2019-06-26
Build script updates

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
1. We were not able to have folders inside of core.
2. We want to be able to change between different renderers at build time.

### Proposed Changes
1. Move all files in the subfolders to a temp folder when running compile.sh.
2. Pass in a flag to the build.py file (-renderer RENDERNAME) and the compile.sh file (RENDERNAME).

### Reason for Changes
1. The closure compiler treats different sub directories as different packages.
2. We want to add new renderers so we need a way to choose between the different renderers at build time.
 

### Test Coverage

Tested on:


build.py
```diff
@@ -523,9 +516,68 @@ def run(self):
...
+    self.langfiles = False
+    self.render_name = "block_rendering_rewrite"
+
+# Add all directories to the path list except for any extra renderers.
+def find_path(args, directories):
+  new_list = []
+  render_path = 'core/renderers/' + args.render_name
+  core_search_paths = calcdeps.ExpandDirectories(directories)
+  for path in core_search_paths:
+    #If it is the desired renderer
```
RoboErikG: nit space after # here and below

tests/compile/compile.sh
```diff
@@ -73,9 +73,47 @@ if [ -f "$BLOCKLY_ROOT/tests/compile/main_compressed.js" ]; then
...
+    done
+  else
+    find $dir -type f -exec cp {} $tempPath \;
+  fi
+done
+
 echo "Compiling Blockly..."
 COMPILATION_COMMAND="java -jar $COMPILER --js='$BLOCKLY_ROOT/tests/compile/main.js' \
-  --js='$BLOCKLY_ROOT/core/**.js' \
+  --js='$BLOCKLY_ROOT/tests/compile/temp_core/**.js' \
```
RoboErikG: This should reference the tempPath variable -> '$tempPath/**.js'

tests/compile/compile.sh
```diff
@@ -73,9 +73,47 @@ if [ -f "$BLOCKLY_ROOT/tests/compile/main_compressed.js" ]; then
...
+    echo "A renderer with the name $1 does not exist"
+    echo "Please provide a valid renderer name"
+    exit 1
+  fi
+else
+  renderName="block_rendering_rewrite"
+fi
+
+
+tempPath="$BLOCKLY_ROOT/tests/compile/temp_core"
```
RoboErikG: Why is this under tests? It's more common to see an output or a build directory at the root of the project.

PR #2594, opened 2019-06-27
Added custom fields demo.

BeksOmega:
## The basics

- [X] I branched from develop

I added some comments in places where comments are generally not supposed to go.

### Resolves

Work on #1623 

### Proposed Changes

Adds a custom fields demo with turtle fields.
DemoPage

### Reason for Changes

Demos are fun.

### Test Coverage

Tested using all of the buttons and all of the blocks, everything looks good.

Tested on:
* Desktop Chrome

### Additional Information

To test out the demo I think you need to rebuild the core because of the utils name change.

I added inline comments as was suggested, but I may have gone a bit overboard.


BeksOmega:
> There seems to be a bad interaction between the no-matches validator and the default values.

Thank you for noticing this! Sadly this issue is unfixable with the current API, so I'll come up with a different example and add it in.

demos/custom-fields/field_turtle.js
```diff
@@ -0,0 +1,705 @@
...
+  // the parameters into an object.
+  var value = {};
+  value.pattern = opt_pattern;
+  value.hat = opt_hat;
+  value.turtleName = opt_turtleName;
+
+  var valid = this.doClassValidation_(value);
+  if (valid === null) {
+    // See the doClassValidation_ function for information on the
+    // cachedValidatedValue_ property.
```
RoboErikG: Why return null and create this cachedValidatedValue_? You'd get the same effect by setting the properties on newValue to the default when it's invalid and returning the newValue in doClassValidation_.

demos/custom-fields/field_turtle.js
```diff
@@ -0,0 +1,705 @@
...
+// Used to create the DOM of our field.
+CustomFields.FieldTurtle.prototype.initView = function() {
+  // Because we want to have both a borderRect_ (background) and a
+  // textElement_ (text) we can call the super-function. If we only wanted
+  // one or the other, we could call their individual createX functions.
+  CustomFields.FieldTurtle.superClass_.initView.call(this);
+
+  // The field group is created by the abstract field's init_ function. *All
+  // elements* should be children of the fieldGroup_.
+  this.movableGroup_ = Blockly.utils.dom.createSvgElement('g',
```
RoboErikG: You might want to pull out the svg code into a separate function and put it at the bottom of the file. It's not really important to the concepts so moving it out of the way will make it easier for folks to follow the rest of the code.

demos/custom-fields/field_turtle.js
BeksOmega: You're correct that in this case it does give you the same effect, but I decided to do it this way because that's what's recommended in the [documentation](https://docs.google.com/document/d/1ONGGy0A_w24L2fDH5AqI5yQG1iyl91hEbAjMiRf3268/edit#heading=h.wcex0ucfeenc).

This is probably easier to explain with an example use case:
Say you want to create a Vector2 field that takes in text input. You would want to be able to validate the text of each number separately, and turn the text inputs red separately. To do this you would need to use the cachedValidatedValue_ pattern to save the mixed-validated state.

As you noticed the turtle field isn't a great example of this pattern, but I figured it was better to include it anyway.

RoboErikG: Sounds reasonable. As a followup change (separate PR) if you'd like to make it clearer how doValueInvalid_ can make use of this you could color the parts of the turtle that are invalid red (hat, shell, name) in addition to tinting the background. ;)

RoboErikG:
Oh, looks like Neil also submitted a change requiring a rebuild. You'll need to rebase and build again, sorry about that.

BeksOmega:
> You'll need to rebase and build again.

Done! :D

demos/custom-fields/field_turtle.js
RoboErikG: Thinking about it more, styling the text of the picker or adding a warning icon next to the text would probably be clearer. That way it's easy to tell what part to change to fix the invalid selection. =)

PR #2596, opened 2019-06-28
Rename block_render_* files

rachel-fenichel:
Simple rename of files to remove the block_render_ prefix, now that we've decided that the files should live in a subdirectory.


rachel-fenichel:
Actually, I'm not going to do this yet.  I still need to think a bit about subdirectories.

Issue #2102, opened 2018-11-02
Package name casing inconsistencies


AnmAtAnm:
The `Blockly` namespace is capitalized.
The `Blockly.Xml` namespace is also capitalized.
The `Blockly.utils` and `Blockly.Xml.utils` used a mix of capitalized and lowercase.

[Google Style Guide recommends](https://google.github.io/styleguide/jsguide.html#naming-package-names) `lowerCamelCase` for each part of a package name.

(The likelihood of fixing these are probably zero, due to backward compatibility needs. Still, it is worth highlighting to prevent further inconsistencies.)

NeilFraser:
To clarify, a name should start with lowercase if it is just a namespace, but should start uppercase if it is a constructor.  Thus `goog` and `goog.math` are lowercase, but `goog.math.Coordinate` starts uppercase since the latter is a constructor.

In our case `Blockly` should have been lowercase, but that's impossible to change now.  `Blockly.inject` is correctly lowercase, `Blockly.FieldAngle` is correctly uppercase, and `Blockly.Touch` is incorrectly uppercase.

A list of currently incorrect names in core are:
```
Blockly
Blockly.BlockAnimations
Blockly.Blocks
Blockly.ContextMenu
Blockly.Css
Blockly.DropDownDiv
Blockly.Events
Blockly.Extensions
Blockly.Msg
Blockly.Procedures
Blockly.Themes.Classic
Blockly.Themes.HighContrast
Blockly.Themes.Modern
Blockly.Tooltip
Blockly.Touch
Blockly.Variables
Blockly.VariablesDynamic
Blockly.WidgetDiv
Blockly.Xml
```
Not sure if there are any here that can be easily renamed without breaking 3rd party stuff.

NeilFraser:
After discussion with the team, the only item on this list we felt comfortable changing was Blockly.BlockAnimations > Blockly.blockAnimations.  The rest failed the cost-benefit analysis.

NeilFraser:
For the record:
* we all agreed that Blockly.Themes should be Blockly.themes
* we could not agree on whether Blockly.Themes.Classic should become Blockly.themes.classic or Blockly.themes.CLASSIC
* we all agreed that the question was moot since any change to themes would be too disruptive.

Closing issue.

PR #2606, opened 2019-07-01
Rename Blockly.Xml.utils to Blockly.utils.xml

NeilFraser:
If we are renaming a bunch of utils in this release, then we might as well rename this one too.


NeilFraser:
One side effect of this PR is that our XML is now case-sensitive (because it's now actual XML, not HTML).  We were already pretty good, the exception was `variabletype` (in variables.js) vs `variableType` (in field_variable.js).  I've normalized to lowercase.

rachel-fenichel:
I know the `variableType` thing has come up repeatedly, though I'd have to dig to find the issue numbers.

But doesn't think mean that we no longer accept all previously generated XML?  That seems like an issue.

BeksOmega:
Related to #2102

PR #2608, opened 2019-07-02
Create bug and feature templates

alschmiedt:


alschmiedt:
This is mainly the default github templates. I think the only thing I added was the stack trace for the bugs template.

PR #2623, opened 2019-07-09
Remove accessibility demo from Blockly repo

### LINKED ISSUE #2565: Move accessible directory to an experiments repo
Per offline discussion--the accessible directory has a demo that is out of date, but an interesting experiment that impacted how we think about screenreaders and accessibility.  We should move it to a `blockly-experiments` repo with an old version of any files it needs and remove it from the main blockly repo.

rachel-fenichel:
The accessibility demo now lives in a new repository, [blockly-experimental](https://github.com/google/blockly-experimental).

The Blockly team will no longer maintain compatibility with the demo code, but we want to keep it around as an interesting artifact.

Our accessibility work is now focused on keyboard navigation in Blockly itself, rather than in a fully distinct renderer for use with screen readers.  As of July 8 2019 that code is in the keyboard_nav branch, and it will land in develop in the next release.

The first commit deletes the accessible folder and sets up the demo to redirect.
The second commit removes it from the build process.

@alschmiedt to look at build.py

Fixes #2565


build.py
```diff
@@ -16,11 +16,9 @@
...
 # limitations under the License.
 
-# Usage: build.py <0 or more of accessible, core, generators, langfiles>
+# Usage: build.py <0 or more of core, generators, langfiles>
 # build.py with no parameters builds all files.
 # core builds blockly_compressed, blockly_uncompressed, and blocks_compressed.
-# accessible builds blockly_accessible_compressed,
-#  blockly_accessible_uncompressed, and blocks_compressed.
 # generators builds every <language>_compressed.js.
 # langfiles builds every msg/js/<LANG>.js file.
```
alschmiedt: Should the below say the script generates two versions? And remove the "first pair" part?

rachel-fenichel: Done.

Issue #1308, opened 2017-09-05
Create DTD for xml format


RoboErikG:
We should create a DTD for Blockly's XML format and integrate it with xml_test.js.

NeilFraser:
After getting a basic implementation working, I discovered that DTD has fallen out of favour, and that XSD is the new hotness.  Nearly all of the online tooling around today uses XSD.

NeilFraser:
XSD created for Blockly's XML format.

PR #2633, opened 2019-07-12
Add XSD to validate Blockly’s XML

NeilFraser:
This XSD might be used in tests or other tooling.

Resolves issue #1308.


tests/xml/blockly.xsd
```diff
@@ -0,0 +1,178 @@
...
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:element name="variable">
+    <xs:complexType>
+      <xs:simpleContent>
+        <xs:extension base="xs:NCName">
+          <xs:attribute name="id"/>
+          <xs:attribute name="type" use="required"/>
```
RoboErikG: Note: I think this means old xml formats from before types are won't pass as valid..

NeilFraser: Currently we throw an error if we try to load a variable that doesn't have a type attribute:
`Error: Variable with id, -~F#HnYc~^iM+NrU/(#S is without a type xml.js:579:13`

RoboErikG: Oh, wait, I'm confused. This only applies to the list of variables at the top, which has always required a type. variableType is optional for fields.

Issue #2640, opened 2019-07-15
DropDownDiv Doesn't Handle Small Workspaces Well


BeksOmega:
### Problem statement

The way the dropdown div handles being too tall for a workspace looks buggy.

### Expected Behavior

If the dropdown div is too big it will stay connected to the field, but render outside the bounds of the workspace.

### Actual Behavior

It renders at the top of the screen. The relevant logic is [here](https://github.com/google/blockly/blob/16a1a747e31b6eb28675f591cb9e6ae0a3e85a30/core/dropdowndiv.js#L344).

### Steps to Reproduce

1. Navigate to the fixed size workspace demo.
2. Change the hight of the blocklyDiv to 100px.
3. Add a "test_fields_angle" block to the toolbox.
4. Load the demo.
5. Create an angle field block.
6. Open the dropdown div.
7. Observe how the angle field is at the top of the screen.

![DropdownDiv_Position](https://user-images.githubusercontent.com/25440652/61246290-5b4ca180-a703-11e9-88f3-50f802caebb8.png)

### Additional Information

This looks especially bad if the dropdown div is far from the top of the screen.

samelhusseini:
Fixed by #2641

PR #2641, opened 2019-07-15
Fixed DropDownDiv Rendering On Small Workspaces

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2640 

### Proposed Changes

The dropdowndiv will now try to render outside the bounds of the workspace if it can't render inside the bounds.

It first tries to render below, but if that goes outside the /window/ bounds it renders above.

### Reason for Changes

The old behavior looked kinda buggy.
DropdownDiv_Position

### Test Coverage

Rendering Above:
DropdownDiv_Position_Bottom

Rendering Below:
DropdownDiv_Position_Top

Tested on:
* Desktop Chrome

### Additional Information

N/A

### DIFF
--- a/core/constants.js
+++ b/core/constants.js
@@ -238,24 +238,6 @@ Blockly.TOOLBOX_AT_LEFT = 2;
 Blockly.TOOLBOX_AT_RIGHT = 3;
 
 /**
- * ENUM for positioning the DropDownDiv directly below the target.
- * @const
- */
-Blockly.DROPDOWNDIV_BELOW = 1;
-
-/**
- * ENUM for positioning the DropDownDiv directly above the target.
- * @const
- */
-Blockly.DROPDOWNDIV_ABOVE = 2;
-
-/**
- * ENUM for positioning the DropDownDiv at the top of the page.
- * @const
- */
-Blockly.DROPDOWNDIV_TOP_OF_PAGE = 3;
-
-/**
  * ENUM representing that an event is not in any delete areas.
  * Null for backwards compatibility reasons.
  * @const
--- a/core/constants.js
+++ b/core/constants.js
@@ -238,6 +238,24 @@ Blockly.TOOLBOX_AT_LEFT = 2;
 Blockly.TOOLBOX_AT_RIGHT = 3;
 
 /**
+ * ENUM for positioning the DropDownDiv directly below the target.
+ * @const
+ */
+Blockly.DROPDOWNDIV_BELOW = 1;
+
+/**
+ * ENUM for positioning the DropDownDiv directly above the target.
+ * @const
+ */
+Blockly.DROPDOWNDIV_ABOVE = 2;
+
+/**
+ * ENUM for positioning the DropDownDiv at the top of the page.
+ * @const
+ */
+Blockly.DROPDOWNDIV_TOP_OF_PAGE = 3;
+
+/**
  * ENUM representing that an event is not in any delete areas.
  * Null for backwards compatibility reasons.
  * @const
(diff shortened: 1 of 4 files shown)


samelhusseini:
Hey Beka, there's one scenario where that this change doesn't cover, and that is the renderTertiary. If you don't have enough room to render above or below (ie: not primary or secondary), the final fallback option is to render at the top of the page, which will render on top of the field. 
Today, that introduces a conflict for fields that make use of both the widget and dropdown div (eg: angle field), as the widget div will show up above the dropdown div blocking certain input from reaching the dropdowndiv. eg: 
Screen Shot 2019-07-26 at 2 28 39 PM

Ideally in that case I think we could render the widgetdiv at the top of the page, and the dropdowndiv right below it in a flow layout, but we can file a bug to track that change. 
I think with this PR if let's include the tertiary fallback option in the logic.

samelhusseini:
Also, if we can't render below (outside of workspace), why not try render above (outside of workspace) as well? 
ie: 
- try below (primary)
- try above (secondary)
- try below including outside of workspace (primary)
- try above including outside of workspace (secondary)
- fallback to on top of the field (tertiary), ie: render it at Y = 0.

core/constants.js
```diff
@@ -237,6 +237,24 @@ Blockly.TOOLBOX_AT_LEFT = 2;
  */
 Blockly.TOOLBOX_AT_RIGHT = 3;
 
+/**
+ * ENUM for positioning the DropDownDiv directly below the target.
+ * @const
+ */
+Blockly.DROPDOWNDIV_BELOW = 1;
+
```
samelhusseini: Since there's not really a need for these enums to be globally accessible and available to developers. You might consider creating the enum in the dropdown div file, something like: 
```

/**
 * The types for positioning a dropdown.
 * @enum {number}
 */
Blockly.DropDownDiv.PositionType = {
  ABOVE: 1,
  BELOW: 2,
  TOP_OF_PAGE: 3,
};
```

BeksOmega:
Ok @samelhusseini  this is ready for another look whenever you're available.

I added some tests because I messed with a bunch of stuff. They don't test that all the calculations are correct, but they do test that the dropdowndiv is picking the correct option.

samelhusseini:
Thanks Beka. The logic looks good. Just a little bit of cleanup and I think we're good :)

core/dropdowndiv.js
```diff
@@ -309,6 +311,8 @@ Blockly.DropDownDiv.show = function(owner, primaryX, primaryY,
  * @private
  */
 Blockly.DropDownDiv.getBoundsInfo_ = function() {
+  // TODO: Account for toolboxes.
```
samelhusseini: Is this new? Where we accounting for toolboxes earlier?

core/dropdowndiv.js
```diff
@@ -224,7 +224,6 @@ Blockly.DropDownDiv.showPositionedByBlock = function(field, block,
   return Blockly.DropDownDiv.show(field, primaryX, primaryY, secondaryX, secondaryY, opt_onHide);
 };
 
-
```
samelhusseini: nit: Remove new line

core/dropdowndiv.js
```diff
@@ -257,7 +256,6 @@ Blockly.DropDownDiv.showPositionedByField = function(owner,
       owner, primaryX, primaryY, secondaryX, secondaryY, opt_onHide);
 };
 
-
```
samelhusseini: nit: Remove new line

core/dropdowndiv.js
BeksOmega: This is new, see #2744 for details.

core/dropdowndiv.js
BeksOmega: ?

samelhusseini: woops never mind.

PR #2651, opened 2019-07-17
Add getCorrectedSize functions to fields

rachel-fenichel:
## The basics

Note: This is a chunk of code from `render/collab` which I'm pulling in piecewise, because the rebase is terrible.

- [x] I branched from develop

### Resolves

Part of new rendering.

### Proposed Changes

Adds `getCorrectedSize` functions to fields and icons.

### Reason for Changes

`getSize` is not guaranteed to return a bounding box on the field, and I don't want to touch it now because that will break old rendering.
New rendering uses `getCorrectedSize`, which calls `getSize` if `getSize` already does the correct thing.

As we get rid of the old rendering I will change `getSize` functions to be correct, and eventually remove `getCorrectedSize` entirely.

### Additional Information

@BeksOmega because this touches fields.

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -556,6 +556,9 @@ Blockly.FieldDropdown.prototype.getCorrectedSize = function() {
   // getSize also renders and updates the size if needed.  Rather than duplicate
   // the logic to figure out whether to rerender, just call getSize.
   this.getSize();
+  // This extra 9 was probably to add padding between rows.
+  // It's also found in render_, renderSelectedImage_, and renderSelectedText_.
+  // TODO (#2562): Remove getCorrectedSize.
   return new goog.math.Size(this.size_.width + Blockly.BlockSvg.SEP_SPACE_X,
       this.size_.height - 9);
 };
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -547,4 +547,17 @@ Blockly.FieldDropdown.validateOptions_ = function(options) {
   }
 };
 
+/**
+ * Get the size of the visible field, as used in new rendering.
+ * @return {!goog.math.Size} The size of the visible field.
+ * @package
+ */
+Blockly.FieldDropdown.prototype.getCorrectedSize = function() {
+  // getSize also renders and updates the size if needed.  Rather than duplicate
+  // the logic to figure out whether to rerender, just call getSize.
+  this.getSize();
+  return new goog.math.Size(this.size_.width + Blockly.BlockSvg.SEP_SPACE_X,
+      this.size_.height - 9);
+};
+
 Blockly.Field.register('field_dropdown', Blockly.FieldDropdown);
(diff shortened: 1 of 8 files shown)


BeksOmega:
@rachel-fenichel  LGTM fields-wise, I can't think of any reason this would break stuffs.

core/field_dropdown.js
```diff
@@ -547,4 +547,17 @@ Blockly.FieldDropdown.validateOptions_ = function(options) {
...
+ * Get the size of the visible field, as used in new rendering.
+ * @return {!goog.math.Size} The size of the visible field.
+ * @package
+ */
+Blockly.FieldDropdown.prototype.getCorrectedSize = function() {
+  // getSize also renders and updates the size if needed.  Rather than duplicate
+  // the logic to figure out whether to rerender, just call getSize.
+  this.getSize();
+  return new goog.math.Size(this.size_.width + Blockly.BlockSvg.SEP_SPACE_X,
+      this.size_.height - 9);
```
RoboErikG: Why is this - 9?

core/field_label.js
```diff
@@ -100,4 +100,16 @@ Blockly.FieldLabel.prototype.doClassValidation_ = function(newValue) {
...
+/**
+ * Get the size of the visible field, as used in new rendering.
+ * @return {!goog.math.Size} The size of the visible field.
+ * @package
+ */
+Blockly.FieldLabel.prototype.getCorrectedSize = function() {
+  // getSize also renders and updates the size if needed.  Rather than duplicate
+  // the logic to figure out whether to rerender, just call getSize.
+  this.getSize();
+  return new goog.math.Size(this.size_.width, this.size_.height - 5);
```
RoboErikG: Why - 5?

core/field_textinput.js
```diff
@@ -432,4 +432,16 @@ Blockly.FieldTextInput.nonnegativeIntegerValidator = function(text) {
...
+/**
+ * Get the size of the visible field, as used in new rendering.
+ * @return {!goog.math.Size} The size of the visible field.
+ * @package
+ */
+Blockly.FieldTextInput.prototype.getCorrectedSize = function() {
+  // getSize also renders and updates the size if needed.  Rather than duplicate
+  // the logic to figure out whether to rerender, just call getSize.
+  this.getSize();
+  return new goog.math.Size(this.size_.width + Blockly.BlockSvg.SEP_SPACE_X, 16);
```
RoboErikG: I'm surprised the height isn't dependent on the text. What happens if you change the css for fonts?

rachel-fenichel: The height doesn't change.  This is partly because SVG has `getComputedTextWidth` but no matching `getComputedTextHeight`

core/field_dropdown.js
rachel-fenichel: TODOs with issue references added, here and everywhere else.

Issue #2315, opened 2019-02-26
FieldAngle CLOCKWISE and OFFSET apply to all angle pickers simultaneously


rachel-fenichel:
### Problem statement
We allow angle pickers to work in both protractor and compass mode, but whatever mode is picked applies to all angle pickers on the screen.
```
/* The following two settings work together to set the behaviour of the angle
 * picker.  While many combinations are possible, two modes are typical:
 * Math mode.
 *   0 deg is right, 90 is up.  This is the style used by protractors.
 *   Blockly.FieldAngle.CLOCKWISE = false;
 *   Blockly.FieldAngle.OFFSET = 0;
 * Compass mode.
 *   0 deg is up, 90 is right.  This is the style used by maps.
 *   Blockly.FieldAngle.CLOCKWISE = true;
 *   Blockly.FieldAngle.OFFSET = 90;
 */

/**
 * Angle increases clockwise (true) or counterclockwise (false).
 */
Blockly.FieldAngle.CLOCKWISE = false;

/**
 * Offset the location of 0 degrees (and all angles) by a constant.
 * Usually either 0 (0 = right) or 90 (0 = up).
 */
Blockly.FieldAngle.OFFSET = 0;
```

### Expected Behavior

You can pick on a per-field basis.

### Additional Information
Blockly's angle picker doesn't have labels to help orient the user, so it might be confusing to allow these to mix.  But if we upgrade the rendering, we may also want to change these properties.

abhinavsagar:
@rachel-fenichel Is this issue still open. Could I work on this?

RoboErikG:
@abhinavsagar yep, should be fine for you to work on. As a note, this should be done by adding a property on the field to override the default behavior with a setter and a getter. The getter should handle checking the local property first and falling back to the global property if it's not set.

@alschmiedt do you have any thoughts on how this should be added to the JSON definition?

alschmiedt:
Talked with Rachel and I think adding a 'mode' field to the JSON definition with the options for 'compass' and 'math' would be ideal. There aren't a lot of requests for other configurations, and most other configurations are pretty confusing anyway.

abhinavsagar:
@alschmiedt @RoboErikG Got it i am taking on this work. I'll open a PR in a couple of days time.

abhinavsagar:
Sorry for the late response. I am unable to solve this with my limited knowledge. Would look forward to solve other issues in this repository.

rachel-fenichel:
[...]

PR #2654, opened 2019-07-17
Added JSON Customization Support For Angle Fields

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2315 plus some

### Proposed Changes

* Adds modes: Compass and Protractor to the angle field JSON definition.
* Also adds properties: clockwise, offset, wrap, and round to the angle field JSON definition.
* Adds functions for setting the individual properties programmatically.
* Adds functions for getting the value of the property (with respect for globals).

### Reason for Changes

* Requested.
* More info below.
* Good practice.
* Good practice.

#### Why did I add individual properties?

I absolutely love love love angle fields. I think they're probably the coolest field in Blockly and I think they're super under utilized. I would hate to see them hamstrung by the JSON.

For example, if there were only the two modes available you couldn't make super cool blocks like this:
AngelJSON

which have built-in affordances for directionally-challenged users.

Providing the modes & the properties (with the properties given override power since they're more specific) is the best of both ease of use, and power.

### Test Coverage

Added test blocks for all of the modes and all of the properties.

Tested all of the blocks to make sure they were functioning properly.

Tested on:
* Desktop Chrome

### Additional Information

- [ ] Doc ticket should be created once this is merged.

Setting round to 0 looks crazy lol:
AngelJSON_Round0

But devs can fix it by setting round to 1 instead, so it doesn't need a fix.

### DIFF
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -61,7 +61,47 @@ goog.inherits(Blockly.FieldAngle, Blockly.FieldTextInput);
  * @nocollapse
  */
 Blockly.FieldAngle.fromJson = function(options) {
-  return new Blockly.FieldAngle(options['angle']);
+  var field = new Blockly.FieldAngle(options['angle']);
+
+  var clockwise = null;
+  var offset = null;
+  var wrap = null;
+  var round = null;
+
+  var mode = options['mode'];
+  switch (mode) {
+    case 'compass':
+      clockwise = true;
+      offset = 90;
+      break;
+    case 'protractor':
+      // This is the default mode, so we could do nothing. But just to
+      // future-proof we'll set it anyway.
+      clockwise = false;
+      offset = 0;
+      break;
+  }
+
+  // Allow individual settings to override the mode setting.
+  if (typeof options['clockwise'] == 'boolean') {
+    clockwise = options['clockwise'];
+  }
+  if (typeof options['offset'] == 'number') {
+    offset = options['offset'];
+  }
+  if (typeof options['wrap'] == 'number') {
+    wrap = options['wrap'];
+  }
+  if (typeof options['round'] == 'number') {
+    round = options['round'];
+  }
+
+  field.setClockwise((clockwise));
+  field.setOffset(offset);
+  field.setWrap(wrap);
+  field.setRound(round);
+
+  return field;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


core/field_angle.js
```diff
@@ -72,51 +112,83 @@ Blockly.FieldAngle.fromJson = function(options) {
...
- *   Blockly.FieldAngle.OFFSET = 90;
+/**
+ * Radius of protractor circle.  Slightly smaller than protractor size since
+ * otherwise SVG crops off half the border at the edges.
+ * @type {number}
+ * @const
  */
+Blockly.FieldAngle.RADIUS = Blockly.FieldAngle.HALF - 1;
+
+/* See <DOCS LINK HERE> for more information about these properties. */
```
BeksOmega: Needs a link.

BeksOmega:
Do you want any block factory support for this? Maybe a mutator with the extra options?

core/field_angle.js
```diff
@@ -347,10 +423,104 @@ Blockly.FieldAngle.prototype.doClassValidation_ = function(newValue) {
...
 };
 
+/**
+ * Set which direction should make the graphical angle editor increase.
+ * @param {?boolean} clockwise Whether the graphical angle editor should
+ *    increase as it is moved clockwise (true) or counterclockwise (false)
+ *    or the global setting (null).
+ * @return {!Blockly.FieldAngle} Returns itself (for method chaining).
+ */
+Blockly.FieldAngle.prototype.setClockwise = function(clockwise) {
```
rachel-fenichel: @RoboErikG thoughts on returning itself for method chaining?  It's a reasonable pattern, but 1) we don't use it anywhere else and 2) we probably want to be consistent.

RoboErikG: It's nice for convenience, but it's pretty non-standard for instances. I think I'd rather start using an opt_config parameter with a map of the configuration options. The keys should be the same as the JSON config which will also let you use the same code for both.

BeksOmega: Do you want me to go ahead and update it? or are you waiting for feedback from Rachel?

rachel-fenichel: +1 to Erik.  Go ahead and update it.

BeksOmega: I went to go update this, and then I realized I had more questions and wasn't sure how far reaching the change should be. I [created an issue](https://github.com/google/blockly/issues/2722) to make discussion easier.

rachel-fenichel: If you're happy with the clarification on #2722, then just remove the method chaining and this will be good to go.

BeksOmega: I was planning on getting the abstract field changes in first (just to make sure how I'm thinking about it in my head is actually going to work). But don't worry I haven't forgotten about this little guy!

Also atm #2722 has this guy down as no hooks. This is b/c, for example,  changing the WRAP value at runtime means the field would have to do some extra work recalculating it's value. I think I could get it to work though.

Did you want hooks for the angle field or no?

BeksOmega:

PR #2656, opened 2019-07-17
Rename BRC to Blockly.renderingConstants

rachel-fenichel:
### Proposed Changes

Rename `BRC` to `Blockly.renderingConstants`.

I'm open to other suggestions for the name.

Lowercase because it's a namespace.

Some things in here may move out into a separate file--particularly the objects for the puzzle tab, etc.  But those don't exist yet.  I think they'll be `Blockly.renderingConstants.paths` if I create them.

I'm open to other suggestions on names.

And I think `Blockly.utils.Path` should actually be `Blockly.utils.svgPath`, per the convention of lowercase for namespaces.

### Reason for Changes

`BRC` was just shorthand to reduce typing.

### Test Coverage

I ran screenshot tests.  96/112 still pass, which is the same as before.


rachel-fenichel:
After discussion with Abby, I'm going to make a `Blockly.blockRendering` namespace and then make this `Blockly.blockRendering.constants`

PR #2663, opened 2019-07-18
Textarea Field, Multiline Block (from acbart)

rachel-fenichel:
This is a version of #2584 from @acbart, with two changes:
- fixed call `Blockly.Xml.utils`, which we renamed while the other PR was open
- removed language files from the PR

@acbart I made a new PR because I didn't want to push to your develop branch. 

## The basics

- [x] I branched from develop 

## The details
See #2584 for details and discussion.


acbart:
I consent.

acbart:
Looks like there's a couple unit tests still failing with this. Do you have it under handled or does this need to be tackled again?

rachel-fenichel:
If you want to take a look, that would be awesome.  The failing unit test does seem to be directly related to the XML.

acbart:
Just as an update, I've tracked it down to the XML unit test changing slightly. There's now an XML element without a closing tag. @NeilFraser added it a few weeks ago: https://github.com/google/blockly/blame/develop/tests/jsunit/xml_test.js#L38

Assuming this is desirable, I think I'll need to update the regex to handle this case. Which... doesn't look super fun, but I'll try to take a stab at it.

acbart:
I believe the new regex should be:

```javascript
var regexp = /(<[^/][^<]*>[^<]*[^/])\n([^<]*<\/)/;
```

In `core/xml.js`: https://github.com/google/blockly/blob/3f661160ce95693b5fbe12ab1f95fae53b6a8538/core/xml.js#L299

But I am not entirely certain. It does pass the tests and my experimentation with a regex tool.

rachel-fenichel:
We haven't forgotten this--Neil will take a look at the regex when he gets a chance.

NeilFraser:
> I believe the new regex should be:
> 
> ```js
> var regexp = /(<[^/][^<]*>[^<]*[^/])\n([^<]*<\/)/;
> ```
> 
> In `core/xml.js`:
> 
> https://github.com/google/blockly/blob/3f661160ce95693b5fbe12ab1f95fae53b6a8538/core/xml.js#L299
> 
> But I am not entirely certain. It does pass the tests and my experimentation with a regex tool.

Oh great.  This is the *exact* issue raised in the most-cited Stack Overflow question of all time:
https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454
Every engineer needs to read the first answer in its entirety.

But accepting our fate that we can't solve the general case, but can solve the Blockly-specific one, the following example shows that your proposed regexp isn't correct:
```
'<foo>\n</foo>'.replace(/(<[^/][^<]*>[^<]*[^/])\n([^<]*<\/)/, '$1&#10;$2');
```

The most straight-forward regexp I can come up with is:
```
var regexp = /(<[^/](?:[^>]*[^/])?>[^<]*)\n([^<]*<\/)/;
```

This would be easier if browsers supported lookbehind assertions, but so far only Chrome supports them.

Definitely also add another example to the comment:
  // E.g. <foo>\n<bar/>\n</foo> is unchanged.

samelhusseini:
An update on this. I have gone through and updated the field so that: 
- It uses ES5
- The closure typings are all correct
- Better alignment of the field
- Use CSS for static style properties
- Works in Edge and IE (Use text with a group instead of tspan).

This is ready for a final review. There's a CLA Issue where it's unable to verify concent. @rachel-fenichel do you have any idea how to resolve?

rachel-fenichel:
CLA is fine--he consented above: https://github.com/google/blockly/pull/2663#issuecomment-512956488

core/field_multilineinput.js
```diff
@@ -0,0 +1,279 @@
...
+  var fontSize =
+      (Blockly.FieldTextInput.FONTSIZE * scale) + 'pt';
+  div.style.fontSize = fontSize;
+  htmlInput.style.fontSize = fontSize;
+  var borderRadius =
+      (Blockly.FieldTextInput.BORDERRADIUS * scale) + 'px';
+  htmlInput.style.borderRadius = borderRadius;
+  var padding = Blockly.Field.DEFAULT_TEXT_OFFSET * scale;
+  htmlInput.style.paddingLeft = padding + 'px';
+  htmlInput.style.width = 'calc(100% - ' + padding + 'px)';
```
rachel-fenichel: I didn't know you could do this.

samelhusseini:
Also, two jsunit tests are failing which I'm looking into.

samelhusseini:
Fixed jsunit tests by using Neil's regex instead.

rachel-fenichel:
Field structure lgtm.

PR #2668, opened 2019-07-19
Deprecated Field.setText

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

* Deprecates the field's setText function.
* Adds a setAlt function to the image field.
* Adds an updateVariableName function to the variable field.

### Reason for Changes

* Because the only access point should be setValue.
* So that developers can still manipulate the image field's alt.
* So that updating the variable name without updating the variable model still works & is efficient.

### Test Coverage

* Added unit tests for setting an image field's alt value (w/ backwards compatibility tests.
* Added unit tests for renaming a variable (keeping same id, and getting a new id).

Tested on:
* Desktop Chrome

### Additional Information

Important: I'm not sure if the new variable field function should be called updateVariableName or updateVarName, there is inconsistency.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1078,7 +1078,7 @@ Blockly.Block.prototype.updateVarName = function(variable) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
       if (field.referencesVariables() &&
           variable.getId() == field.getValue()) {
-        field.updateVariableName();
+        field.refreshVariableName();
       }
     }
   }
--- a/core/block.js
+++ b/core/block.js
@@ -1078,7 +1078,7 @@ Blockly.Block.prototype.updateVarName = function(variable) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
       if (field.referencesVariables() &&
           variable.getId() == field.getValue()) {
-        field.setText(variable.name);
+        field.updateVariableName();
       }
     }
   }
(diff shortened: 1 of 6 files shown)


core/block.js
```diff
@@ -1078,7 +1078,7 @@ Blockly.Block.prototype.updateVarName = function(variable) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
       if (field.referencesVariables() &&
           variable.getId() == field.getValue()) {
-        field.setText(variable.name);
+        field.updateVariableName();
```
BeksOmega: BTW would you rather this be updateVariableName or updateVarName? There are some places in the code base where it is spelled out explicitly, and other places where it is abbreviated (e.g. the function that makes this call).

core/field_variable.js
```diff
@@ -341,6 +341,16 @@ Blockly.FieldVariable.prototype.setTypes_ = function(opt_variableTypes,
   this.variableTypes = variableTypes;
 };
 
+/**
+ * A callback to update the name of a variable. Should only be called by
+ * the block.
+ * @package
+ */
+Blockly.FieldVariable.prototype.updateVariableName = function() {
+  this.text_ = this.variable_.name;
```
rachel-fenichel: Can you rename this to `refreshVariableName`?  I want to indicate that it isn't actually changing the name on the model, just re-grabbing it because the model changed out from underneath the field.  And update the comment accordingly.

core/block.js
rachel-fenichel: See comment below.

PR #2678, opened 2019-07-21
Remove Dependency on dart:html

### LINKED ISSUE #2662: Generate Dart Code that Does Not Require dart:html
The generated Dart code for ```text_prompt``` requires ```Html.window.prompt``` which is not available on the standalone Dart VM.

TheBrokenRail:
Fixes #2662

### DIFF
--- a/generators/dart/text.js
+++ b/generators/dart/text.js
@@ -274,8 +274,14 @@ Blockly.Dart['text_print'] = function(block) {
 
 Blockly.Dart['text_prompt_ext'] = function(block) {
   // Prompt function.
-  Blockly.Dart.definitions_['import_dart_html'] =
-      'import \'dart:html\' as Html;';
+  Blockly.Dart.definitions_['import_dart_io'] =
+      'import \'dart:io\';';
+  var functionName = Blockly.Dart.provideFunction_(
+      'text_prompt',
+      [Blockly.Dart.FUNCTION_NAME_PLACEHOLDER_ + '(msg) {',
+       '  print(msg);',
+       '  return stdin.readLineSync((;',
+       '}']);
   if (block.getField('TEXT')) {
     // Internal message.
     var msg = Blockly.Dart.quote_(block.getFieldValue('TEXT'));
@@ -284,7 +290,7 @@ Blockly.Dart['text_prompt_ext'] = function(block) {
     var msg = Blockly.Dart.valueToCode(block, 'TEXT',
         Blockly.Dart.ORDER_NONE) || '\'\'';
   }
-  var code = 'Html.window.prompt(' + msg + ', \'\')';
+  var code = functionName + '(' + msg + ', \'\')';
   var toNumber = block.getFieldValue('TYPE') == 'NUMBER';
   if (toNumber) {
     Blockly.Dart.definitions_['import_dart_math'] =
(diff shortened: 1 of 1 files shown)


generators/dart/text.js
```diff
@@ -274,8 +274,14 @@ Blockly.Dart['text_print'] = function(block) {
...
   // Prompt function.
-  Blockly.Dart.definitions_['import_dart_html'] =
-      'import \'dart:html\' as Html;';
+  Blockly.Dart.definitions_['import_dart_io'] =
+      'import \'dart:io\';';
+  var functionName = Blockly.Dart.provideFunction_(
+      'text_prompt',
+      [Blockly.Dart.FUNCTION_NAME_PLACEHOLDER_ + '(msg) {',
+       '  print(msg);',
+       '  return stdin.readLineSync((;',
```
NeilFraser: This should be `()`, not `((`.

generators/dart/text.js
```diff
@@ -284,7 +290,7 @@ Blockly.Dart['text_prompt_ext'] = function(block) {
     var msg = Blockly.Dart.valueToCode(block, 'TEXT',
         Blockly.Dart.ORDER_NONE) || '\'\'';
   }
-  var code = 'Html.window.prompt(' + msg + ', \'\')';
+  var code = functionName + '(' + msg + ', \'\')';
```
NeilFraser: The second argument is unused and creates an error.

generators/dart/text.js
```diff
@@ -274,8 +274,14 @@ Blockly.Dart['text_print'] = function(block) {
 
 Blockly.Dart['text_prompt_ext'] = function(block) {
   // Prompt function.
-  Blockly.Dart.definitions_['import_dart_html'] =
-      'import \'dart:html\' as Html;';
+  Blockly.Dart.definitions_['import_dart_io'] =
+      'import \'dart:io\';';
```
NeilFraser: When this runs in DartPad it generates:
```
Uncaught exception:
Unsupported operation: StdIOUtils._getStdioInputStream
```

generators/dart/text.js
TheBrokenRail: Fixed!

generators/dart/text.js
TheBrokenRail: Fixed!

generators/dart/text.js
TheBrokenRail: This is because Dart for Web does not have ```dart:io``` while the standalone Dart VM does not have ```dart:html```. However due to the fact that the rest of Blockly has no web specific function related to the web, I assumed most users would be using the standalone VM to execute their code, and if they were adding web based blocks, they would just override the prompt block's generator.

NeilFraser: Hmm, I'm not sure that we can assume that all of Blockly's Dart users are server-side.  Is it possible to make the code smart enough to detect what's available?

TheBrokenRail: As far as I know Dart does not support conditional imports.

TheBrokenRail: That seems to require both modules having the same API, or creating separate modules, that have the same API. ```dart:html``` and ```dart:io``` have different APIs for user input, and Blockly does not have the ability to create multiple modules.

BeksOmega:
Closing this pull request as stale. Sorry we didn't get back to you after your last reply :/ If you're still interested in working on this feel free to reopen and comment.

PR #2690, opened 2019-07-23
Define BORDER_RECT_DEFAULT_HEIGHT constant

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Missing definition of this property.

### Proposed Changes

Define it, and use it in two places.

### Reason for Changes

Remove magic numbers.  Also, I was already using this in the new rendering code, but didn't have the definition in.


core/field_textinput.js
```diff
@@ -442,7 +442,8 @@ Blockly.FieldTextInput.prototype.getCorrectedSize = function() {
   // the logic to figure out whether to rerender, just call getSize.
   this.getSize();
   // TODO (#2562): Remove getCorrectedSize.
```
alschmiedt: Can we get rid of this todo?

PR #2700, opened 2019-07-24
Adds helper functions for keyboard nav

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
This adds in the helper functions necessary for keyboard navigation.

### Test Coverage

Tested on:

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -391,6 +391,28 @@ Blockly.BlockSvg.prototype.moveToDragSurface_ = function() {
 };
 
 /**
+ * Move a block to a position.
+ * @param {goog.math.Coordinate} xy The position to move to in workspace units.
+ */
+Blockly.BlockSvg.prototype.moveTo = function(xy) {
+  if (this.parentBlock_) {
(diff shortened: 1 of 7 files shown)


rachel-fenichel:
Run the rename script, then lgtm.

core/block_svg.js
```diff
@@ -390,6 +390,28 @@ Blockly.BlockSvg.prototype.moveToDragSurface_ = function() {
   this.workspace.blockDragSurface_.setBlocksAndShow(this.getSvgRoot());
 };
 
+/**
+ * Move a block to a position.
+ * @param {goog.math.Coordinate} xy The position to move to in workspace units.
+ */
+Blockly.BlockSvg.prototype.moveTo = function(xy) {
```
RoboErikG: This is duplicating most of the code in moveBy. Unless there's a reason not to can we make one of them just convert the units and then call the other?

alschmiedt: I added a todo to the cleanup for keyboard nav list but would prefer to keep that separate from the keyboard nav to develop move.

PR #2701, opened 2019-07-24
Fix some jsdoc types

samelhusseini:
## The basics

In preparation for the typings generation script. Fixing a few JSDoc types that have caused conflicts. 

- [x] I branched from develop

### Resolves

Fixes 10 Closure compiler warnings, but more importantly fixes type issues that produce invalid typings. 

### Reason for Changes

To enable generating typings from jsdoc comments.

### Test Coverage

I tested that this doesn't break compilation using the closure compiler. I also tested that this creates valid typings when using the typescript closure generator.

### Additional Information

I'll add additional comments in the PR explaining some changes.


core/field_angle.js
```diff
@@ -334,8 +334,8 @@ Blockly.FieldAngle.prototype.updateGraph_ = function() {
 
 /**
  * Ensure that the input value is a valid angle.
- * @param {string|number=} newValue The input value.
- * @return {?number} A valid angle, or null if invalid.
```
RoboErikG: Why is this valid in field_number but not field_angle?

core/block_svg.js
```diff
@@ -872,10 +872,10 @@ Blockly.BlockSvg.prototype.getSvgRoot = function() {
 
 /**
  * Dispose of this block.
- * @param {boolean} healStack If true, then try to heal any gap by connecting
+ * @param {boolean=} healStack If true, then try to heal any gap by connecting
  *     the next statement with the previous statement.  Otherwise, dispose of
  *     all children of this block.
- * @param {boolean} animate If true, show a disposal animation and sound.
+ * @param {boolean=} animate If true, show a disposal animation and sound.
  */
```
samelhusseini: Blockly.Block defines dispose without any parameters, the subclass (Blockly.BlockSvg) can't have dispose define required parameters.

core/dropdowndiv.js
```diff
@@ -199,7 +199,7 @@ Blockly.DropDownDiv.setCategory = function(category) {
  * @param {!Blockly.Block} block Block to position the drop-down around.
  * @param {Function=} opt_onHide Optional callback for when the drop-down is
  *   hidden.
- * @param {number} opt_secondaryYOffset Optional Y offset for above-block
+ * @param {number=} opt_secondaryYOffset Optional Y offset for above-block
```
samelhusseini: It's optional, also can't have a required field appear after an optional field.

core/field.js
```diff
@@ -864,7 +864,7 @@ Blockly.Field.prototype.onMouseDown_ = function(e) {
 
 /**
  * Change the tooltip text for this field.
- * @param {string|function|!Element} newTip Text for tooltip or a parent
+ * @param {string|Function|!Element} newTip Text for tooltip or a parent
  *    element to link to for its tooltip.
```
samelhusseini: typo

core/field_image.js
```diff
@@ -37,8 +37,8 @@ goog.require('goog.math.Size');
 /**
  * Class for an image on a block.
  * @param {string=} src The URL of the image. Defaults to an empty string.
- * @param {!(string|number)} width Width of the image.
- * @param {!(string|number)} height Height of the image.
+ * @param {!(string|number)=} width Width of the image.
+ * @param {!(string|number)=} height Height of the image.
  * @param {string=} opt_alt Optional alt text for when block is collapsed.
  * @param {Function=} opt_onClick Optional function to be called when the image
```
samelhusseini: Can't have required fields after the optional field (src). 
Alternately we can make src required. 

If we decide to go with this change we should probably have defaults for height and width.

core/theme.js
```diff
@@ -27,16 +27,29 @@ goog.provide('Blockly.Theme');
 
 
 /**
- * A block style or a category style.
- * @typedef {!Object.<string, string>}
- */
-Blockly.Theme.Style;
+ * A block style.
+ * @typedef {{
+  *            colourPrimary:string,
```
samelhusseini: You can define interfaces like this. I've split the interfaces for style into two: BlockStyle and CategoryStyle after chatting with Abby as the two accept different fields.

core/block_svg.js
```diff
@@ -262,6 +262,7 @@ Blockly.BlockSvg.prototype.getIcons = function() {
 /**
  * Set parent of this block to be a new block or null.
  * @param {Blockly.BlockSvg} newParent New parent block.
+ * @override
```
samelhusseini: Challenge here is the superclass (Block) and subclass (BlockSvg) define different parameter types for newParent, one is Block the other is BlockSvg and that conflicts. 

more details about override field:
https://github.com/google/closure-compiler/wiki/@abstract-classes-and-methods

core/field_angle.js
samelhusseini: Because field_angle inherits from field_text which defines this newValue as string.

samelhusseini: Turns out field_number didn't have the @extends jsdoc, but it is also a problem there. Thanks for pointing this out.

core/field_image.js
BeksOmega: Currently not passing a src is still /technically/ invalid (or at least bad behavior). The thing is that it just shows a missing image icon instead of throwing an error.

I think it would be better to make src required and throw errors if it is not provided correctly (null, undefined, empty string).

samelhusseini: That makes sense. I'll change it back to required, and add the check

BeksOmega: Looks perfect thank you!

PR #2710, opened 2019-07-24
Integrates cursor into blockly

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Implements the cursor.

### Test Coverage

Tested on:


core/gesture.js
```diff
@@ -752,15 +755,24 @@ Blockly.Gesture.prototype.doBlockClick_ = function() {
 
 /**
  * Execute a workspace click.
+ * @param {!Event} e A mouse up or touch end event.
  * @private
  */
-Blockly.Gesture.prototype.doWorkspaceClick_ = function() {
-  if (Blockly.selected) {
+Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
+  Blockly.keyboardAccessibilityMode = false;
```
rachel-fenichel: Add a comment: Shift-clicking puts the workspace in accessibility mode.

core/blockly.js
```diff
@@ -46,6 +46,8 @@ goog.require('Blockly.FieldTextInput');
 goog.require('Blockly.FieldNumber');
 goog.require('Blockly.FieldVariable');
 goog.require('Blockly.Generator');
+// TODO: Decide if this require actually goes here.
```
rachel-fenichel: As opposed to having it on a per-workspace basis?

And is this recorded in gnosis?  If not, put it there or as a gitub issue so we don't lose track of it.

core/workspace.js
```diff
@@ -31,6 +31,7 @@ goog.require('Blockly.utils');
 goog.require('Blockly.utils.math');
 goog.require('Blockly.VariableMap');
 goog.require('Blockly.WorkspaceComment');
+goog.require('Blockly.Cursor');
```
rachel-fenichel: Move up (alphabetize)

core/workspace_svg.js
```diff
@@ -597,6 +627,11 @@ Blockly.WorkspaceSvg.prototype.dispose = function() {
     this.zoomControls_ = null;
   }
 
+  if (this.cursor_) {
```
rachel-fenichel: Also dispose of the marker if it exists.

And you don't have to set the marker or cursor to null (we're not going to do that for new stuff, and will phase it out in older stuff).

rachel-fenichel: Actually--if the cursor exists, call dispose on it in the Workspace dispose function and not here.  And the marker dispose has to be called here.

core/blockly.js
alschmiedt: I think this is from when the navigation file was a lot different: https://github.com/google/blockly/pull/2219/files. So I think we can get rid of this comment.

core/keyboard_nav/cursor.js
```diff
@@ -13,7 +13,6 @@
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
rachel-fenichel: ???

core/blockly.js
```diff
@@ -247,8 +263,11 @@ Blockly.onKeyDown_ = function(e) {
     } else if (e.keyCode == 90) {
       // 'z' for undo 'Z' is for redo.
       Blockly.hideChaff();
-      mainWorkspace.undo(e.shiftKey);
+      workspace.undo(e.shiftKey);
```
samelhusseini: @alschmiedt What's the reason for this line of change? It seems to have broken undo with Ctrl + Z.

alschmiedt: I think this just got incorrectly copied over from the rebase. I'll make a PR to fix now.

PR #2723, opened 2019-07-29
Fix mocha

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Mocha tests were giving null tests failed and failing intermittently on CI. 

### Proposed Changes
Adds tests_failed="unset" to the failureCount div.

### Reason for Changes
The browser's wait until method was checking that the tests_failed text != unset. The value was not initialized with "unset" so it would only check once before moving on. This resulted was null tests failed if the tests did not finish fast enough.
### Test Coverage

Tested on:

### DIFF
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -12,7 +12,7 @@
   <body>
 
     <div id="mocha"></div>
-    <div id="failureCount" style="display:none"></div>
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Awesome.  Thanks for hunting this down!

PR #2739, opened 2019-07-30
Remove all calls to goog.math

samelhusseini:
Remove calls to goog.math.
It turns out we're only using goog.math.Size and goog.math.Coordinate and we're using them as a struct and not utilizing any of their helper methods.
Unlike https://github.com/google/blockly/pull/2108, the proposal here is to keep the structs the way they are and just reference our version instead of closure's. 
This way the shape doesn't change and won't break anyone depending on these fields having certain properties.


core/utils/math.js
```diff
@@ -68,3 +68,47 @@ Blockly.utils.math.clamp = function(lowerBound, number, upperBound) {
...
+ * @param {number=} opt_y Top, defaults to 0.
+ * @struct
+ * @constructor
+ */
+Blockly.utils.math.Coordinate = function(opt_x, opt_y) {
+  /**
+   * X-value
+   * @type {number}
+   */
+  this.x = goog.isDef(opt_x) ? opt_x : 0;
```
NeilFraser: Please don't introduce new dependencies on Closure.

samelhusseini:
No difference, my bad looks like that was already defined.  Fixed.

core/utils/size.js
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Utility methods for size calculation.
+ * These methods are not specific to Blockly, and could be factored out into
+ * a JavaScript framework such as Closure.
+ * @author fraser@google.com (Neil Fraser)
```
NeilFraser: Are you the new me?

core/gesture.js
```diff
@@ -764,7 +764,7 @@ Blockly.Gesture.prototype.doWorkspaceClick_ = function(e) {
   var ws = this.creatorWorkspace_;
   ws.cursor.hide();
   if (e.shiftKey) {
-    var screenCoord = new goog.math.Coordinate(e.clientX, e.clientY);
```
NeilFraser: I'd love to know how this one slipped through my efforts last month...
Ah, @alschmiedt added it last week.  It's not in master.

tests/jsunit/event_test.js
```diff
@@ -238,7 +238,7 @@ function test_move_constructorCoordinate() {
   setUpMockMethod(mockControl_, Blockly.utils, 'genUid', null, ['1', '2']);
   try {
     var block1 = createSimpleTestBlock(workspace);
-    var coordinate = new goog.math.Coordinate(3,4);
+    var coordinate = new Blockly.utils.Coordinate(3,4);
```
NeilFraser: Not the fault of this PR, but can we get some spaces between arguments here and below?

PR #2740, opened 2019-07-30
Fixes copy paste to work with the marker

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Copy paste was broken for keyboard navigation.

### Proposed Changes
If we are in accessibility mode and a connection has been marked then try to insert the copied block at the marked location. If the marker is not at a valid connection point, then warn the user. 

### Test Coverage

Tested on:

### DIFF
--- a/core/keyboard_nav/navigation.js
+++ b/core/keyboard_nav/navigation.js
@@ -542,6 +542,9 @@ Blockly.Navigation.insertBlock = function(block, targetConnection) {
   if (bestConnection && bestConnection.isConnected() &&
       !bestConnection.targetBlock().isShadow()) {
     bestConnection.disconnect();
+  } else if (!bestConnection) {
+    Blockly.Navigation.warn(
+    'This block can not be inserted at the marked location. Moved to workspace.');
   }
   return Blockly.Navigation.connect(bestConnection, targetConnection);
 };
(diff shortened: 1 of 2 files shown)


core/keyboard_nav/navigation.js
```diff
@@ -542,6 +542,9 @@ Blockly.Navigation.insertBlock = function(block, targetConnection) {
   if (bestConnection && bestConnection.isConnected() &&
       !bestConnection.targetBlock().isShadow()) {
     bestConnection.disconnect();
+  } else if (!bestConnection) {
+    Blockly.Navigation.warn(
+        'This block can not be inserted at the marked location. Moved to workspace.');
```
RoboErikG: Does this move the target connection to the workspace in this case or leave it where it is?

alschmiedt: It leaves the marked connection where it is.

RoboErikG: Update the warning to say it's being left where it is then.

alschmiedt: The block is moved to the workspace which is what I was trying to communicate with the "Moved to workspace". Would it make more sense for it to be "Block has been moved to the top level workspace?"

RoboErikG: In the copy paste case it is, but this method also gets called if you select a block in a stack and press 'i' to insert it. In that case, if the marked location is invalid does it stay in the stack or get moved to the workspace?

alschmiedt: ah yup forgot about that. I will take that part out since it doesn't apply to all scenarios.

PR #2745, opened 2019-07-30
Position notches

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Some work to position previous and next notches.
Check whether the highlighter exists before using it.

I think this is just moving 0.5s around, unfortunately.


rachel-fenichel:
Yes, and it doesn't crash.  That doesn't mean everything looks great, though.  There's an interaction between the highlighting and the dark path offset that will eventually need fixing.

PR #2746, opened 2019-07-30
Fix references to createSvgElement

samelhusseini:
A couple of references to createSvgElement weren't moved when it moved under Blockly.utils.dom.


NeilFraser:
This duplicates #2727.

samelhusseini:
Yep Thanks Neil.

PR #2750, opened 2019-07-30
Inject Blockly version into binaries

samelhusseini:
Inject the Blockly version from package.json into compressed Blockly builds.

This makes it easier for developers to let us know which version of Blockly they are running.

The way the variable is defined (Jsdoc), it lets us also use the --define flag when compiling locally. see --define flag for more info: https://github.com/google/closure-compiler/wiki/Flags-and-Options

### DIFF
--- a/build.py
+++ b/build.py
@@ -56,7 +56,7 @@ else:
   from urllib.parse import urlencode
   from importlib import reload
 
-# Read package.json and extract the current Blockly version
+# Read package.json and extract the current Blockly version.
 blocklyVersion = json.loads(open('package.json', 'r').read())['version']
 
 def import_path(fullpath):
@@ -224,9 +224,12 @@ class Gen_compressed(threading.Thread):
       if filename.startswith(os.pardir + os.sep):  # '../'
         continue
       f = codecs.open(filename, encoding="utf-8")
-      params.append(("js_code", "".join(f.readlines()).encode("utf-8")))
+      code = "".join(f.readlines()).encode("utf-8")
+      # Inject the blockly version.
+      if filename == "core/blockly.js":
+        code = code.replace("Blockly.VERSION = 'unset';", "Blockly.VERSION = '" + blocklyVersion + "';")
+      params.append(("js_code", code))
       f.close()
-
     self.do_compile(params, target_filename, filenames, "")
 
   def gen_blocks(self):
@@ -348,7 +351,6 @@ class Gen_compressed(threading.Thread):
 
       code = HEADER + "\n" + json_data["compiledCode"]
       code = code.replace(remove, "")
-      code = code.replace('Blockly.version="unset"', 'Blockly.version="' + blocklyVersion + '"')
       code = self.trim_licence(code)
 
       stats = json_data["statistics"]
--- a/build.py
+++ b/build.py
@@ -56,6 +56,9 @@ else:
   from urllib.parse import urlencode
   from importlib import reload
 
+# Read package.json and extract the current Blockly version
+blocklyVersion = json.loads(open('package.json', 'r').read())['version']
+
 def import_path(fullpath):
   """Import a file with full path specification.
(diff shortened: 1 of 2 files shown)


core/blockly.js
```diff
@@ -132,6 +132,12 @@ Blockly.cache3dSupported_ = null;
  */
 Blockly.theme_ = null;
 
+/**
+ * Blockly core version.
+ * @define {string}
+ */
+Blockly.version = "unset";
```
NeilFraser: I think this should be a constant.  Thus `Blockly.VERSION`
Also, Google prefers single quoted strings unless there's a reason otherwise (e.g. JSON, or strings with single quotes)
https://google.github.io/styleguide/javascriptguide.xml?showone=Strings#Strings

core/blockly.js
```diff
@@ -132,6 +132,12 @@ Blockly.cache3dSupported_ = null;
  */
 Blockly.theme_ = null;
 
+/**
+ * Blockly core version.
+ * @define {string}
```
NeilFraser: What's `define`?  Did you mean `type`?
https://google.github.io/styleguide/jsguide.html#appendices-jsdoc-tag-reference

build.py
```diff
@@ -345,6 +348,7 @@ def file_lookup(name):
 
       code = HEADER + "\n" + json_data["compiledCode"]
       code = code.replace(remove, "")
+      code = code.replace('Blockly.version="unset"', 'Blockly.version="' + blocklyVersion + '"')
```
NeilFraser: This is rather fragile.  I would do the replacement on the code *before* compilation, since you've got now guarantees of what you'll get out the other side.  Any of these (and more) are possible:
```
Blockly.version="unset"
Blockly.version='unset'
Blockly.c="unset"
f.c="unset"
var fc="unset"
var Blockly={version:"unset",...}
```

build.py
```diff
@@ -56,6 +56,9 @@
   from urllib.parse import urlencode
   from importlib import reload
 
+# Read package.json and extract the current Blockly version
```
NeilFraser: Prefer to end with a period.

"Complete sentences are recommended but not required. Complete sentences should use appropriate capitalization and punctuation."
https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments

core/blockly.js
samelhusseini: Nope, it's handy for replacing that value for local compilations. (although not using it now, it seemed like the right usecase for it). see https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#define-type-description

core/blockly.js
samelhusseini: Cool, thanks for the tip.

build.py
samelhusseini: Yeah that makes sense.

core/blockly.js
NeilFraser: Oh, neat.  Is there a reason not to use that feature by passing a flag to the compiler rather than editing the code?

samelhusseini: Unfortunately the online compiler doesn't support it :(

NeilFraser: Sounds like a good GTI project.  :)

samelhusseini:
@NeilFraser this is ready for another look.

core/blockly.js
```diff
@@ -65,6 +65,12 @@ goog.require('Blockly.Xml');
 var CLOSURE_DEFINES = {'goog.DEBUG': false};
 /* eslint-enable no-unused-vars */
 
+/**
+ * Blockly core version.
+ * @define {string}
+ */
+Blockly.VERSION = 'unset';
```
NeilFraser: I wonder if this should be 'uncompiled'...  Just a thought.  No opinion on this.

PR #2755, opened 2019-07-31
Work on connection locations

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

- rename `renderMoveConnections` to `updateConnectionLocations_`, move it from `block_render_svg.js` to `block_svg.js`, and call it in the same place for both renderers.
- Use the new constants in `rendered_connection.js` for connection highlighting
- Fix some connection locations in new code 
- Change connection placement in old code to make the new highlights work in the old code

### Reason for Changes

- code dedup
- fixing connection highlights in new rendering
- fixing block positioning in new code, maybe
- the old code doesn't mind a few more constants sprinkled in, and this makes it so that I don't have to check which renderer I'm using when highlighting connections.

### Test Coverage
Tested in the playground in both LTR and RTL, in classic and compatibility rendering modes.

### Additional Information
Future work: connection location for previous connection should be from the connection measurable, not from the constant, when it's being set in `positionPreviousConnection`.

### DIFF
--- a/core/block_render_svg.js
+++ b/core/block_render_svg.js
@@ -102,6 +102,21 @@ Blockly.BlockSvg.TAB_WIDTH = 8;
  * @const
  */
 Blockly.BlockSvg.NOTCH_WIDTH = 30;
+
+/**
+ * Offset of the notch from the left side of the block.
+ * @type {number}
+ * @const
+ */
+Blockly.BlockSvg.NOTCH_OFFSET_X = 15;
+
+/**
+ * Offset of the puzzle tab from the top of the block.
+ * @type {number}
+ * @const
+ */
+Blockly.BlockSvg.TAB_OFFSET_Y = 5;
+
 /**
  * Rounded corner radius.
  * @const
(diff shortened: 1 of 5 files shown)


core/block_render_svg.js
```diff
@@ -687,8 +651,8 @@ Blockly.BlockSvg.prototype.renderDrawTop_ = function(pathObject, rightEdge) {
     steps.push(Blockly.BlockSvg.NOTCH_PATH_LEFT);
     highlightSteps.push(Blockly.BlockSvg.NOTCH_PATH_LEFT_HIGHLIGHT);
 
-    var connectionX = (this.RTL ?
-        -Blockly.BlockSvg.NOTCH_WIDTH : Blockly.BlockSvg.NOTCH_WIDTH);
+    var notchOffsetStart = 15;
```
RoboErikG: Why was the constant replaced with a number? It also appears in two other spots below, so worth pulling back out into a constant.

core/block_render_svg.js
```diff
@@ -813,7 +773,7 @@ Blockly.BlockSvg.prototype.renderDrawLeft_ = function(pathObject) {
   var highlightSteps = pathObject.highlightSteps;
   if (this.outputConnection) {
     // Create output connection.
-    this.outputConnection.setOffsetInBlock(0, 0);
+    this.outputConnection.setOffsetInBlock(0, 5);
```
RoboErikG: Add a constant for the y offset?

core/block_svg.js
```diff
@@ -1547,6 +1547,8 @@ Blockly.BlockSvg.prototype.render = function(opt_bubble) {
   } else {
     this.renderInternal();
   }
+  // No matter how we rendered, connections locations should now be correct.
```
RoboErikG: nit, s/connections/connection/ (no s, since it's modifying locations)

core/block_render_svg.js
rachel-fenichel: Done, here and for the other constant.

This change moves the connection's position from the right of the notch to the left of the notch.  The notch width constant used above is *not* the width of the notch.  It's the distance from the left of the block to the right of the notch.

The other change moves the connection position from 5 pixels above the puzzle tab to the top of the puzzle tab.

These changes are safe as long as I always pair changes to matching types of connections.

core/block_render_svg.js
rachel-fenichel: Done

core/block_svg.js
rachel-fenichel: Done

PR #2758, opened 2019-07-31
Fix angle picker when 'Infinity' is typed in.

NeilFraser:
Previously 'NaN' would be generated, which is not a legal value.

### DIFF
--- a/core/block_events.js
+++ b/core/block_events.js
@@ -429,7 +429,7 @@ Blockly.Events.Move.prototype.fromJson = function(json) {
   if (json['newCoordinate']) {
     var xy = json['newCoordinate'].split(',');
     this.newCoordinate =
-        new Blockly.utils.Coordinate(parseFloat(xy[0]), parseFloat(xy[1]));
+        new Blockly.utils.Coordinate(Number(xy[0]), Number(xy[1]));
   }
 };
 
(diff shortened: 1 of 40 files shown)


BeksOmega:
Does this same fix apply to the number field? If it does would you mind adding it there as well? ([link to relevant line](https://github.com/google/blockly/blob/5029f2d3a04ca0e666e9080dbc3f3022f82b73b3/core/field_number.js#L122))

NeilFraser:
@BeksOmega Infinity (and -Infinity) are accepted by number fields, and I think that's ok.  We even have an Infinity constant on one of the Math blocks.  The issue is that Infinity is not a valid angle since it turns to NaN when modded by 360.  Does that make sense?

However, I will do an audit of all 19 instances of parseFloat() and consider whether each one should be changed to the more restrictive Number().

BeksOmega:
Thanks for the extra work Neil! Sorry I went and turned your lovely little PR into a bigger PR :/

> Infinity (and -Infinity) are accepted by number fields, and I think that's ok. We even have an Infinity constant on one of the Math blocks. The issue is that Infinity is not a valid angle since it turns to NaN when modded by 360. Does that make sense?

Makes sense! Thanks for explaining!

NeilFraser:
> Why did you remove all the constructor null param tests?

From the description of that commit:

Null is explicitly disallowed in most field constructors.  Thus the random results of passing null are undefined and may change.  These changes should not be tested against.

Exceptions:  Variable and serializable label fields actually do take null.

This commit is due to the upcoming parseFloat->Number commit which happens to change how null behaves.

PR #2764, opened 2019-08-01
Make the TabList exclude hidden fields

royvs33:
## The basics

- [x] I branched from develop

### Resolves

When using the tab button and pressing tab from a visible field, the code does not skip over hidden fields.

### Proposed Changes

Added a check if the field is visible

### Reason for Changes

Behavior not logical

### Test Coverage

None

Tested on:
* Desktop Chrome
* Desktop Firefox

* Windows Edge

### Documentation

None

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -605,7 +605,7 @@ Blockly.BlockSvg.prototype.createTabList_ = function() {
   var list = [];
   for (var i = 0, input; input = this.inputList[i]; i++) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
-      if (field instanceof Blockly.FieldTextInput) {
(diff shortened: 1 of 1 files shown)


core/block_svg.js
```diff
@@ -605,7 +605,7 @@ Blockly.BlockSvg.prototype.createTabList_ = function() {
   var list = [];
   for (var i = 0, input; input = this.inputList[i]; i++) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
-      if (field instanceof Blockly.FieldTextInput) {
+      if (field instanceof Blockly.FieldTextInput && field.visible_) {
```
alschmiedt: Could you use the field.isVisible() method instead of accessing the private variable directly? Otherwise lgtm : )

Issue #756, opened 2016-12-02
Block Factory injection vulnerability


NeilFraser:
Failure to escape user data.  Look at the JSON for this block:
https://blockly-demo.appspot.com/static/demos/blockfactory/index.html#xd6zrv

Obviously check other places than just this one field.  And don't rely on the validation of number fields, since one could disable the validators locally, save arbitrary XML, then send the URL to someone else.

rachel-fenichel:
The problem appears to be in how we're escaping the input to prettyPrintOne.

PR #2768, opened 2019-08-01
 Fix injection vulnerability in Block Factories.

NeilFraser:
Fixes issue #756.

### DIFF
--- a/demos/blockfactory/factory_utils.js
+++ b/demos/blockfactory/factory_utils.js
@@ -902,9 +902,9 @@ FactoryUtils.defineAndGetBlockTypes = function(blockDefsString, format) {
 FactoryUtils.injectCode = function(code, id) {
   var pre = document.getElementById(id);
   pre.textContent = code;
-  code = pre.textContent;
-  code = PR.prettyPrintOne(code, 'js');
-  pre.innerHTML = code;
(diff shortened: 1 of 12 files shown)


demos/blockfactory/factory_utils.js
```diff
@@ -902,9 +902,9 @@ FactoryUtils.defineAndGetBlockTypes = function(blockDefsString, format) {
 FactoryUtils.injectCode = function(code, id) {
   var pre = document.getElementById(id);
   pre.textContent = code;
-  code = pre.textContent;
-  code = PR.prettyPrintOne(code, 'js');
-  pre.innerHTML = code;
+  // Remove the 'prettyprinted' class, so that Prettify will recalculate.
+  pre.className = pre.className.replace('prettyprinted', '');
```
RoboErikG: I don't think this is sufficient. If I name it 'prettyprettyprintedprinted' the result of the replace is 'prettyprinted'

NeilFraser: This is not a user-definable field.  The class is hardcoded in our HTML, and the pretty printer appends 'prettyprinted' when the text is syntax highlighted.

RoboErikG: Got it. I was trying to figure out how the injection worked but was unclear where the input came from. =)

PR #2769, opened 2019-08-02
Remove self-closing XML tags in workspace factory

NeilFraser:
Workspace Factory generates self-closing XML tags that are incompatible with being embedded in an HTML document.

### DIFF
--- a/demos/blockfactory/workspacefactory/wfactory_controller.js
+++ b/demos/blockfactory/workspacefactory/wfactory_controller.js
@@ -343,6 +343,10 @@ WorkspaceFactoryController.prototype.exportXmlFile = function(exportMode) {
     throw new Error(msg);
   }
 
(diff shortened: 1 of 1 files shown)


demos/blockfactory/workspacefactory/wfactory_controller.js
```diff
@@ -343,6 +343,10 @@ WorkspaceFactoryController.prototype.exportXmlFile = function(exportMode) {
     throw new Error(msg);
   }
 
+  // Unpack self-closing tags.  These tags fail when embedded in HTML.
+  // <block name="foo"/> -> <block name="foo"></block>
+  configXml = configXml.replace(/<(\w+)([^<]*)\/>/g, '<$1$2></$1>');
+
```
samelhusseini: A block name of "foo/>" would break this, but I checked and we already sanitize characters in the block factory.

Issue #2152, opened 2018-12-04
Dispose vs Garbage Collector


NeilFraser:
Most of our .dispose() functions set all the properties of the doomed object to null.  Such as:
https://github.com/google/blockly/blob/master/core/workspace_svg.js#L507

This used to be critically important for garbage collection reasons.  Circular data structures were immortal with the early reference counting garbage collectors.  However, as of IE 8 all browsers now use a mark and sweep garbage collector which doesn't have this limitation.

Additionally, early browsers had a problem where JS and the DOM had separate garbage collectors.  Thus if JS referenced a DOM node, the DOM could not be garbage collected.  I can't find any indications that this is still a problem.

Accordingly, our dispose functions probably don't need to set these properties to null.  Removing these lines should not introduce any memory leaks.  To be sure, check the browser's memory footprint, then airstrike, spaghetti, delete all blocks, repeat 1000 times, and check the browser's memory footprint, for every browser.

Additionally, there's an issue with block.dispose in that it disposes of all inputs, which disposes of the connections on each input, then block.dispose disposes of all connections a second time.  This second pass ought to only dispose of next/previous/output connections since the connections on inputs have already been disposed of.

And finally, @rachel-fenichel would like a .disposed property on these objects, so that we don't need to be sniffing block.workspace or input.sourceBlock_ to see if the object is disposed.

BeksOmega:
> Additionally, early browsers had a problem where JS and the DOM had separate garbage collectors. Thus if JS referenced a DOM node, the DOM could not be garbage collected. I can't find any indications that this is still a problem.

I realized I never clarified this before, but it seems this is indeed still a problem. From the [Fix memory problems chrome devtools page](https://developer.chrome.com/docs/devtools/memory-problems/#discover_detached_dom_tree_memory_leaks_with_heap_snapshots):

> A DOM node can only be garbage collected when there are no references to it from either the page's DOM tree or JavaScript code. A node is said to be "detached" when it's removed from the DOM tree but some JavaScript still references it. Detached DOM nodes are a common cause of memory leaks.

PR #2770, opened 2019-08-02
Updated How Connections Dispose of Themselves

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2152

### Proposed Changes

* Connections now have a default behavior for dealing with connected blocks on dispose.
  * This matches the previous behavior which was stored in block.js.
  * This does not affect how blocks dispose of their children (which is different than the default behavior depending on the healstack property), because all disposal of children is handled before the disposal of connections.
* Connections now have a disposed property.
* Blocks also have a disposed property, because it was needed for unit tests.

### Reason for Changes

Code simplification.

#### Added unit tests for:
* Unplugging with connected shadow blocks.
* Dispose tests that correspond to all of the unplug tests (including the new shadow ones).
* Remove input tests for:
  * No connected block.
  * Connected shadow block.
  * Connected corporeal block.

#### I ran a memory test for disposing following the below procedure:
1. Defined the below function.
```
function memoryStrike(n) {
  for (var i = 0; i < n; i++) {
    spaghetti(8);
    workspace.getTopBlocks()[0].dispose(false);
  }
}
```
2. Attached the function to a button with an `n` value of 30.
3. Started a Timeline recording following the procedure described [here](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#visualize_memory_leaks_with_timeline_recordings).
4. Ran the memoryStrike.
5. Ended the Timeline recording following the procedure described [here](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#visualize_memory_leaks_with_timeline_recordings).

Repeating the same test on develop showed that the memory consumption was comparable.
[DisposeProfiles.zip](https://github.com/google/blockly/files/3462706/DisposeProfiles.zip)

#### I ran a smaller scale detached-nodes check following the below procedure:
1. Ran spaghetti stress test.
2. Deleted all blocks.
3. Created a heap snapshot following the procedure described [here](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#discover_detached_dom_tree_memory_leaks_with_heap_snapshots).
4. Observed how all detached nodes were related to events (undo/redo) or workspace audio.

Repeating the same test on develop yielded the same number of detached nodes (17072)

#### I ran a similar memory test for removing inputs:
1. Modified the memoryStrike to look like so:
```
function memoryStrike(n) {
        for (var i = 0; i < n; i++) {
          spaghetti(8);
          var blocks = workspace.getBlocksByType('logic_compare');
          for(var j = 0, block; block = blocks[j]; j++) {
            block.removeInput('A');
            block.removeInput('B');
          }
        }
      }
```
2. Attached the function to a button with an `n` value of 1.
3. Started a Timeline recording following the procedure described [here](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#visualize_memory_leaks_with_timeline_recordings).
4. Ran the memoryStrike.
5. Ended the Timeline recording following the procedure described [here](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#visualize_memory_leaks_with_timeline_recordings).

Repeating the same test on develop showed that the memory consumption was comparable.
[RemoveInputProfiles.zip](https://github.com/google/blockly/files/3462708/RemoveInputProfiles.zip)

#### I also created a heap snapshot to check for detached nodes.

I observed how all detached nodes were related to events (undo/redo) or workspace audio.

Repeating the same test on develop yielded the same number of detached nodes (8686)

Tested on:
* Desktop Chrome

### Documentation

Nope!

### Additional Information

Each commit can be reviewed separately.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -316,12 +316,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
     this.inputList.length = 0;
     // Dispose of any remaining connections (next/previous/output).
     var connections = this.getConnections_(true);
-    for (var i = 0; i < connections.length; i++) {
-      var connection = connections[i];
-      if (connection.isConnected()) {
-        connection.disconnect();
-      }
-      connections[i].dispose();
+    for (var i = 0, connection; connection = connections[i]; i++) {
+      connection.dispose();
     }
   } finally {
     Blockly.Events.enable();
@@ -1759,17 +1755,6 @@ Blockly.Block.prototype.moveNumberedInputBefore = function(
 Blockly.Block.prototype.removeInput = function(name, opt_quiet) {
   for (var i = 0, input; input = this.inputList[i]; i++) {
     if (input.name == name) {
-      if (input.connection && input.connection.isConnected()) {
-        input.connection.setShadowDom(null);
-        var block = input.connection.targetBlock();
-        if (block.isShadow()) {
-          // Destroy any attached shadow block.
-          block.dispose();
-        } else {
-          // Disconnect any attached normal block.
-          block.unplug();
-        }
-      }
       input.dispose();
       this.inputList.splice(i, 1);
       return;
--- a/core/block.js
+++ b/core/block.js
@@ -321,9 +321,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
     }
   } finally {
     Blockly.Events.enable();
+    this.disposed = true;
   }
-
-  this.disposed = true;
 };
 
 /**
--- a/core/block.js
+++ b/core/block.js
@@ -322,6 +322,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
   } finally {
     Blockly.Events.enable();
   }
+
+  this.disposed = true;
 };
 
 /**
--- a/core/block.js
+++ b/core/block.js
@@ -227,6 +227,13 @@ Blockly.Block.obtain = function(workspace, prototypeName) {
 Blockly.Block.prototype.data = null;
 
 /**
+ * Has this block been disposed of?
+ * @type {boolean}
+ * @package
+ */
+Blockly.Block.prototype.disposed = false;
+
+/**
  * Colour of the block as HSV hue value (0-360)
  * This may be null if the block colour was not set via a hue number.
  * @type {?number}
--- a/tests/mocha/block_test.js
+++ b/tests/mocha/block_test.js
@@ -104,27 +104,35 @@ suite('Blocks', function() {
         blocks.B.unplug(true);
         assertUnpluggedNoheal(blocks);
       });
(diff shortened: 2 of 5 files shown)


rachel-fenichel:
That sounds like a lot of detached nodes related to events.  Possibly worth a separate issue for tracking and investigating.

core/block.js
```diff
@@ -316,16 +316,14 @@ Blockly.Block.prototype.dispose = function(healStack) {
...
-      }
-      connections[i].dispose();
+    for (var i = 0, connection; connection = connections[i]; i++) {
+      connection.dispose();
     }
   } finally {
     Blockly.Events.enable();
   }
+
+  this.disposed = true;
```
RoboErikG: Should this be inside the finally? I'm not sure what would happen with a half-disposed block, but I suspect it'd be better to mark it disposed.

tests/mocha/block_test.js
```diff
@@ -56,142 +45,285 @@ suite('Blocks', function() {
...
-
-        this.blocks = {
-          A: blockA,
-          B: blockB,
-          C: blockC
-        };
-      });
+    suite('Dispose', function() {
+      function assertDisposedNoheal(blocks) {
+        chai.assert.isNotOk(blocks.A.disposed);
```
rachel-fenichel: Why check `isNotOk` if you want it to be `false`, but `isTrue` if you want it to be `true`?  

Is it because blocks don't have a `disposed` property until they've been disposed of?  If so, why not have that property (`Blockly.Block.prototype.disposed = false`) and then set it to `true` on the block when dispose is called?

Edit: and if the answer is that you'll get to it when you get to block disposing, that's fine too.

core/block.js
BeksOmega: I think you're right, but I'm not sure. I'm planning to investigate this more once I get to block disposal.

The thing is I'm not sure exactly what that try block is catching; I'm thinking it might be for if removeInput throws? In which case it would be redundant (opt_quite could just be set to true instead).

For now I'm going to leave it (since it's just for testing), but this is the first thing I'm planning to scrub once I get to blocks!

tests/mocha/block_test.js
BeksOmega: > Is it because blocks don't have a `disposed` property until they've been disposed of?

You are correct!

> If so, why not have that property (`Blockly.Block.prototype.disposed = false`) and then set it to `true` on the block when dispose is called?

Sounds good, I'll go ahead and add it, plus update the connections and fields as well. First thing tomorrow :D

BeksOmega: Fixed!

core/block.js
BeksOmega: Fixed!

PR #2772, opened 2019-08-02
Don't try to zoom flyouts and mutators

### LINKED ISSUE #2752: Exception on doing pinch zoom over blocklyDiv
**Describe the bug**

Doing pinch zoom over blocklyDiv (Div that contains blocks) throws an exception in the developer console.

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/demos/fixed/index.html
2. Open developer console
3. Do pinch zoom over the blocklyDiv 
4. See the exception in the developer console

**Expected behavior**

Doing pinch zoom over blocklyDiv should not throw any excpetion

**Screenshots**

Screenshot 2019-07-31 at 12 05 34 PM

ChromeOS : 74.0.3729.159 (Official Build) (64-bit) 
Platform: 11895.118.0 (Official Build)
JavaScript: V8 7.4.288.28 
Flash: 32.0.0.171 

**Stack Traces**

```
blockly_compressed.js:1350 Uncaught TypeError: Cannot read property 'scaleSpeed' of undefined
    at Blockly.WorkspaceSvg.zoom (blockly_compressed.js:1350)
    at Blockly.TouchGesture.handleTouchMove (blockly_compressed.js:1234)
    at Blockly.TouchGesture.handleMove (blockly_compressed.js:1230)
    at HTMLDocument.h (blockly_compressed.js:1818)
```

**Additional context**

I think this issue will be happening on other touch devices as well.

RoboErikG:
Flyouts and mutator workspaces don't have enough info to handle
zooming correctly, so they were throwing errors if you pinched them.
This adds a check to quit a zoom early if in either of those. We
might want to revisit later if we should support zooming on them.

## The basics

- [x] I branched from develop

### Resolves

Fixes #2752

### Test Coverage

Tested on:
Desktop Chrome

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/2772)

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1690,6 +1690,10 @@ Blockly.WorkspaceSvg.prototype.setBrowserFocus = function() {
  *     amount values zoom in.
  */
 Blockly.WorkspaceSvg.prototype.zoom = function(x, y, amount) {
+  // Mutators and flyouts don't support zooming
+  if (this.isFlyout || this.isMutator) {
+    return;
+  }
   // Scale factor.
   var speed = this.options.zoomOptions.scaleSpeed;
   var scaleChange = Math.pow(speed, amount);
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1690,7 +1690,9 @@ Blockly.WorkspaceSvg.prototype.setBrowserFocus = function() {
  *     amount values zoom in.
  */
 Blockly.WorkspaceSvg.prototype.zoom = function(x, y, amount) {
-  // Mutators and flyouts don't support zooming
+  // TODO (#2782): Consider removing once pinch understands zoom configuration
+  // Mutators and flyouts don't support zooming, and pinch doesn't understand
+  // that.
   if (this.isFlyout || this.isMutator) {
     return;
   }


BeksOmega:
I'm not sure that this is the root of the problem.

Flyouts and mutators shouldn't be zooming anyway because they're not passed a 'zoom' struct in their workspace options when they're created. ([toolbox flyout](https://github.com/google/blockly/blob/d471f5104539843cb2912c2749600c8a80f0cdbe/core/toolbox.js#L177), [workspace flyout](https://github.com/google/blockly/blob/d471f5104539843cb2912c2749600c8a80f0cdbe/core/workspace_svg.js#L670), [mutator](https://github.com/google/blockly/blob/d471f5104539843cb2912c2749600c8a80f0cdbe/core/mutator.js#L145))

I think the problem is related to touch gestures not checking if the workspace is zoomable before zooming ([link](https://github.com/google/blockly/blob/d471f5104539843cb2912c2749600c8a80f0cdbe/core/touch_gesture.js#L254)). You can see that the workspace's mouse wheel handling does check before calling zoom ([link](https://github.com/google/blockly/blob/d471f5104539843cb2912c2749600c8a80f0cdbe/core/workspace_svg.js#L1314)).

So checks should either be added to the touch gestures, or moved into the workspace zoom function.

I could be wrong about this though, that's just my read : )

samelhusseini:
@BeksOmega has a good point, the problem is that we're not respecting the zoom options with pinch to zoom. 

Currently there is no zoom option for that though. We could overload wheel, we could create a new option, or we could just check for the existence of the zoom option. I've filed a bug tracking that https://github.com/google/blockly/issues/2782

I think this is still a good change and a good place to do it, as in even if pinch to zoom is enabled, we should not zoom in mutators or flyouts for the reasons above.

BeksOmega:
> even if pinch to zoom is enabled, we should not zoom in mutators or flyouts for the reasons above.

Makes sense!

> I've filed a bug tracking that #2782

Would it be ok to add a TODO (#2782) to this PR and add comment on that issue? Once that gets fixed these checks will become redundant (since zoom structs don't get passed to flyouts or mutators), so they should probably get cleaned up.

Also note that #2752 should probably stay open since that's related to the main workspace and not mutators/flyouts.

RoboErikG:
TODO sounds good.

@BeksOmega #2752 is for the exception spam. When I tested that spam only happens if you pinch on the flyout or a mutator workspace. I think the filer just wasn't sure what exactly was triggering it. Is there a case where it happens on the main workspace?

BeksOmega:
> Is there a case where it happens on the main workspace?

Nope you're right : )

I was assuming it was throwing b/c scaleSpeed wouldn't get set if a zoom struct doesn't get passed, but apparently the options parsing handles this and sets the default scaleSpeed to 1. The options of flyouts and mutators just don't get parsed so defaults aren't handled, and it ends up throwing.

Someday I'll learn all the quirks of this system hehe

RoboErikG:
I don't think any of us understand all the quirks of the system, so when you figure them out let us know. ;)

PR #2774, opened 2019-08-03
Add jsdoc to messages

samelhusseini:
Add jsdoc annotations to messages so that the typescript generator tool picks it up.

Disclaimer, I don't know a huge amount about how our translation service looks at this file. I assume the comments are formatted in a specific way for a reason, eg: {{Notranslate}}. 

In order to avoid breaking any automated tool, I added the jsdoc at the top of the comment block for each message. 

Ran ``npm run typings`` to generate typings.
Also ran advanced compilation and it doesn't introduce any more warnings (partly because I think this file isn't included, and is only used for generating other message files).

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -70,6 +70,7 @@ Blockly.ASTNode = function(type, location, opt_params) {
 
 /**
  * Object holding different types for an AST node.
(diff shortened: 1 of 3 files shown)


RoboErikG:
@NeilFraser Know of any issues with annotating the message strings like this?

PR #2776, opened 2019-08-03
Center text field input when editing

samelhusseini:
Make field text centered and use input instead of keyup/press.

Old behaviour: 
old
Notice how the text jumps left when editing.

New behaviour:
new

The keypress / keydown event combo was used to support IE9- browsers. Since we've dropped compatibility, we can use the input event which is available on all browsers we support: https://caniuse.com/input-event.

## The basics

- [x] I branched from develop

### Resolves

N/A

### Proposed Changes

Center the text in the input widget and use the input change event to detect changes instead of the keypress / keyup event.

### Reason for Changes

It looks better when the text doesn't jump out of place when you're editing.

### Test Coverage

Tested typing characters, clearing input, select all and delete, escape, enter, and pressing and holding (all with expected behaviour)

Tested on:
- Desktop Chrome
- Desktop Safari
- Edge 18
- IE 11

### Additional Information

The input even is available on all the browsers we support: 
https://caniuse.com/#feat=input-event

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -538,7 +538,8 @@ Blockly.Css.CONTENT = [
     'margin: 0;',
     'outline: none;',
     'padding: 0 1px;',
-    'width: 100%',
+    'width: 100%;',
+    'text-align: center;',
   '}',
 
   '.blocklyMainBackground {',
--- a/core/css.js
+++ b/core/css.js
@@ -537,7 +537,7 @@ Blockly.Css.CONTENT = [
     'height: 100%;',
     'margin: 0;',
     'outline: none;',
-    'padding: 0 1px;',
+    'padding: 0;',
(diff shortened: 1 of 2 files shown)


BeksOmega:
I love this change! Text has looked odd for a long time and this looks so much better.

I'm wondering though, how does this look for angle fields (since they have that extra degree symbol)?

samelhusseini:
Yes, good question. It's not going to have the degree symbol in the input text and so the size is going to be different. There will be a slight offset when editing the angle picker.
However that offset is already there, and I would argue is more pronounced today. Here's a comparison: 

Old: 
old

New: 
new

BeksOmega:
Definitely looks better! Thanks for satisfying my curiosity! =)

Issue #2530, opened 2019-06-02
New Field Rendering


BeksOmega:
### Problem statement

Before the new field creation documentation goes up, I think there needs to be some guidelines for how to set up a field's SVG structure. Things like padding, spacing, etc so that people can know their fields will look good and be handled correctly.

### Example

**Note: I think this is why editable fields are acting like jackasses.**

Right now the field rendering is a bit messy and unclear. For example the [createBorderRect_ ](https://github.com/google/blockly/blob/7c5a94d0d9642991abb0de48db0049d8d7fc3d1a/core/field.js#L259)function:
```
Blockly.Field.prototype.createBorderRect_ = function() {
  this.borderRect_ = Blockly.utils.createSvgElement('rect',
      {
        'rx': 4,
        'ry': 4,
        'x': -Blockly.BlockSvg.SEP_SPACE_X / 2,
        'y': 0,
        'height': 16,
        'width': this.size_.width + Blockly.BlockSvg.SEP_SPACE_X
      }, this.fieldGroup_);
};
```
References [Blockly.BlockSvg.SEP_SPACE_X](https://developers.google.com/blockly/reference/js/Blockly.BlockSvg#.SEP_SPACE_X), which is documented as "Horizontal space between elements." But if you remove references to Blockly.BlockSvg.SEP_SPACE_X:
```
Blockly.Field.prototype.createBorderRect_ = function() {
  this.borderRect_ = Blockly.utils.createSvgElement('rect',
      {
        'rx': 4,
        'ry': 4,
        'x': 0/*-Blockly.BlockSvg.SEP_SPACE_X / 2*/,
        'y': 0,
        'height': 16,
        'width': this.size_.width/* + Blockly.BlockSvg.SEP_SPACE_X*/
      }, this.fieldGroup_);
};

Blockly.Field.prototype.updateSize_ = function() {
  var width = Blockly.Field.getCachedWidth(this.textElement_);
  if (this.borderRect_) {
    this.borderRect_.setAttribute('width',
        width/* + Blockly.BlockSvg.SEP_SPACE_X*/);
  }
  this.size_.width = width;
};
```
You can see that it is /actually/ being used as padding:
![Rendering_BorderX](https://user-images.githubusercontent.com/25440652/58762861-ec2b3d00-8508-11e9-974b-9db5a7acc167.jpg)

### Problem Statement Continued

Here's an initial list of things I think need to be decided:
- [ ] Horizontal Padding
- [ ] Vertical Padding
[...]

PR #2783, opened 2019-08-06
Remove getCorrectedSize for fields

### LINKED ISSUE #2562: Remove getCorrectedSize
`getCorrectedSize` is a function defined on fields and icons to bridge the gap between old and new rendering: we cannot make changes to `getSize` to make new rendering work without breaking old rendering.

Many implementations of `getCorrectedSize` return the size of the field with some offset in X and Y.  We should be able to remove those offsets in pairs: in the code that made them necessary and in `getCorrectedSize`.

In the best case, we'll entirely remove `getCorrectedSize` and just use the bounding box of the `svgRoot`, as @BeksOmega prototyped here: https://github.com/google/blockly/compare/render/collab...BeksOmega:fixes/getCorrectedSize

This issue should not be worked on by external contributors until new rendering has landed.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

#2562

### Proposed Changes

Update field sizing to always describe all visible parts of the field, including the border rect if it exists.

### Reason for Changes

Make field sizing more consistent.
Note that this breaks the old renderer.

### Test Coverage
71/112 tests pass in LTR; 86/112 in RTL.  All field types pass, and the remaining test failures are due to other known bugs.

### Documentation

After I finish this + the followup work, will need to update the documentation on fields to include a description of what the size represents.

### Additional Information

Also related to #2530

Followup work: 
- Fix all fields that actually position themselves at -5 px, which may be all fields with a border rect, and remove `dealWithOffsetFields`
- Remove `getCorrectedSize` for icons
- Update documentation

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -136,9 +136,12 @@ Blockly.Field.BORDER_RECT_DEFAULT_HEIGHT = 16;
  */
 Blockly.Field.TEXT_DEFAULT_HEIGHT = 12.5;
 
-Blockly.Field.WIDTH_PADDING = 10;
-
-Blockly.Field.HEIGHT_PADDING = 10;
+/**
+ * The padding added to the width by the border rect, if it exists.
+ * @type {number}
+ * @package
+ */
+Blockly.Field.X_PADDING = 10;
 
 /**
  * Name of field.  Unique within each block.
@@ -297,11 +300,13 @@ Blockly.Field.prototype.initView = function() {
 Blockly.Field.prototype.createBorderRect_ = function() {
   this.size_.height =
       Math.max(this.size_.height, Blockly.Field.BORDER_RECT_DEFAULT_HEIGHT);
+  this.size_.width =
+      Math.max(this.size_.width, Blockly.Field.X_PADDING);
   this.borderRect_ = Blockly.utils.dom.createSvgElement('rect',
       {
         'rx': 4,
         'ry': 4,
-        'x': -Blockly.Field.WIDTH_PADDING / 2,
+        'x': -Blockly.Field.X_PADDING / 2,
         'y': 0,
         'height': this.size_.height,
         'width': this.size_.width
@@ -318,7 +323,7 @@ Blockly.Field.prototype.createTextElement_ = function() {
   this.textElement_ = Blockly.utils.dom.createSvgElement('text',
       {
         'class': 'blocklyText',
-        // This may just be trying to vertically center the text?
+        // The y position is the baseline of the text.
         'y': Blockly.Field.TEXT_DEFAULT_HEIGHT,
         'x': 0
       }, this.fieldGroup_);
@@ -587,7 +592,7 @@ Blockly.Field.prototype.updateSize_ = function() {
   var textWidth = Blockly.Field.getCachedWidth(this.textElement_);
   var totalWidth = textWidth;
   if (this.borderRect_) {
-    totalWidth += Blockly.Field.WIDTH_PADDING;
+    totalWidth += Blockly.Field.X_PADDING;
     this.borderRect_.setAttribute('width', totalWidth);
   }
   this.size_.width = totalWidth;
--- a/core/field.js
+++ b/core/field.js
@@ -144,6 +144,12 @@ Blockly.Field.TEXT_DEFAULT_HEIGHT = 12.5;
 Blockly.Field.X_PADDING = 10;
 
 /**
+ * The default offset between the left of the text element and the left of the
+ * border rect, if the border rect exists.
+ * @type {[type]}
+ */
+Blockly.Field.DEFAULT_TEXT_OFFSET = Blockly.Field.X_PADDING / 2;
+/**
  * Name of field.  Unique within each block.
  * Static labels are usually unnamed.
  * @type {string|undefined}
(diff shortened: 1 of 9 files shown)


core/field.js
```diff
@@ -124,6 +129,20 @@ Blockly.Field.cacheReference_ = 0;
  */
 Blockly.Field.BORDER_RECT_DEFAULT_HEIGHT = 16;
 
+/**
+ * The default height of the text element on any field.
+ * @type {number}
+ * @package
+ */
+Blockly.Field.TEXT_DEFAULT_HEIGHT = 12.5;
+
```
samelhusseini: What is this number based on? Does it have any relationship with Blockly.FieldText.FONTSIZE?

samelhusseini: ie: if a developer changes the font size from 11pt to 12pt, would they have to fiddle around with this number?

core/field_colour.js
```diff
@@ -78,15 +78,15 @@ Blockly.FieldColour.fromJson = function(options) {
...
+Blockly.FieldColour.DEFAULT_WIDTH = 16 + Blockly.Field.X_PADDING;
 
 /**
  * Default height of a colour field.
  * @type {number}
  * @private
  * @const
  */
-Blockly.FieldColour.DEFAULT_HEIGHT = 12;
+Blockly.FieldColour.DEFAULT_HEIGHT = Blockly.Field.BORDER_RECT_DEFAULT_HEIGHT;
```
RoboErikG: Is losing the -1 intentional?

core/field.js
rachel-fenichel: Yes, they would have to fiddle with it. As mentioned in one of the other bugs, we don't handle font size changes correctly.

Right now this just matches the previous value, though it often showed up as an implicit part of another sizing term.

core/field_image.js
```diff
@@ -114,6 +116,14 @@ Blockly.FieldImage.prototype.EDITABLE = false;
  */
 Blockly.FieldImage.prototype.isDirty_ = false;
 
+/**
+ * Vertical padding below the image, which is included in the reported height of
+ * the field.
+ * @type {number}
+ * @private
+ */
+Blockly.FieldImage.Y_PADDING = 1;
```
RoboErikG: Should constants be at the top of the file?

core/field_image.js
```diff
@@ -59,14 +59,16 @@ Blockly.FieldImage = function(src, width, height,
...
+  if (imageHeight <= 0 || imageWidth <= 0) {
     throw Error('Height and width values of an image field must be greater' +
       ' than 0.');
   }
-  this.size_ = new Blockly.utils.Size(this.width_,
-      this.height_ + 2 * Blockly.BlockSvg.INLINE_PADDING_Y);
+  // Store the image height, since it is different from the field height.
+  this.imageHeight_ = imageHeight;
+  this.size_ = new Blockly.utils.Size(imageWidth,
+      imageHeight + Blockly.FieldImage.Y_PADDING);
```
RoboErikG: Does this mean there's only 1/2 a pixel above and below an image or should this be 2*Y_PADDING?

core/field_colour.js
rachel-fenichel: Yes.  The field was reporting its height to be smaller than its actual bounds, in order to make the spacer below it smaller.  I don't want to do that--the height should be accurate--so I'm allowing it to have a slightly taller spacer below.

core/field_image.js
rachel-fenichel: Only below the image.

core/field_image.js
rachel-fenichel: We seem to have standardized on 
- constructor
- fromJson
- properties on the class
- properties on the prototype

I'm moving this above the properties on the prototype to match.  I think the goal was to keep fromJson as close to the constructor it calls as possible, but I'm okay with changing that for all fields at the same time.

RoboErikG: Yeah, that order seems right. This just needs to be above the prototype declarations.

rachel-fenichel: Done.

rachel-fenichel:
I made the change Erik requested, and also removed dealWithOffsetFields, because it turned out to be easy.  PTAL the newest commit.

core/field_colour.js
```diff
@@ -78,15 +78,15 @@ Blockly.FieldColour.fromJson = function(options) {
  * @private
  * @const
  */
-Blockly.FieldColour.DEFAULT_WIDTH = 16;
+Blockly.FieldColour.DEFAULT_WIDTH = 16 + Blockly.Field.X_PADDING;
```
BeksOmega: Since there's no padding on a colour field, I think it would make more sense to set `DEFAULT_WIDTH = 26`.

BeksOmega:
It seems confusing to me that elements get sized based on size_ and the size_ is set based on the size of elements. Is there a plan to reconcile this? Or is that further down the road?

core/field_colour.js
rachel-fenichel: Reasonable.  Done.

rachel-fenichel:
@BeksOmega For the most part sizing is
- get the desired size of the element that is varying (e.g. text element or image element), often based on calling `getCachedWidth` 
- set the size, x and y attributes, etc. of the actual element
- update the border box size if needed
- if border box size is different from reported size, store border box size internally
- calculate reported size based on the values computed above and store in `this.size_`

So the values in `this.size_` should be based on the sizes of the elements, but not vice versa.

The exception is in `initView`, when I set the `size_` and then use it immediately, but that only happens once, not on every change

BeksOmega:
> @BeksOmega For the most part sizing is...

Rad! If it's that easy to explain then I think it's great =)

PR #2787, opened 2019-08-08
Initial commit for changing key mappings

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Currently, there is no easy way to change the key mapping for keyboard navigation. 

### Proposed Changes
Add a key mapping that can be easily switched out so that it is much easier to test with and users can potentially choose their own key mappings. 

### Test Coverage
Working on tests now but thought I would get initial feedback while I am working on tests.

Tested on:

### Additional Information
A demo using this functionality will be added in a different PR.


core/keyboard_nav/key_action.js
```diff
@@ -0,0 +1,42 @@
...
+
+goog.provide('Blockly.KeyAction');
+
+/**
+ * Class for a single action.
+ * There can be one action for each key. If the action only applies to a
+ * single state (toolbox, flyout, workspace) then the function should handle this.
+ * @param {string} name The name of the action.
+ * @param {string} desc The description of the action.
+ * @param {string} func The function to be called when the key is pressed.
```
RoboErikG: This is a {function} not a {string}.

core/keyboard_nav/key_action.js
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview The class representing a keyboard action.
+ * Used primarily for keyboard navigation.
+ */
+'use strict';
+
+goog.provide('Blockly.KeyAction');
```
RoboErikG: s/KeyAction/Action/

These can be triggered independent of keyboard keys.

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * Update the key map and actions array to contain the new action.
+ * @param {!string} keyCode The key code serialized by the serializeKeyEvent.
+ * @param {!Blockly.KeyAction} action The action to be executed when the keys
+ *     corresponding to the serialized key code is pressed.
+ * @package
+ */
+Blockly.user.keyMap.setActionForKey = function(keyCode, action) {
+  var key = Blockly.user.keyMap.getKeyByActionName(action.name);
```
RoboErikG: nit, oldKey is a clearer name here.

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+/**
+ * Gets the current key map.
+ * @return {Object<string,Blockly.KeyAction>} The object holding the key to
+ *     action mapping.
+ * @package
+ */
+Blockly.user.keyMap.getKeyMap = function() {
+  return Blockly.user.keyMap.map_;
```
RoboErikG: Should this return a copy of the map to make it clear you shouldn't edit it directly?

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+ * Holds the serialized key to key action mapping.
+ * @type {Object<string, Blockly.KeyAction>}
+ */
+Blockly.user.keyMap.map_ = {};
+
+/**
+ * Holds the list of actions that can be executed using the keyboard.
+ * @type {Array<Blockly.KeyAction>}
+ */
+Blockly.user.keyMap.actions_ = [];
```
RoboErikG: You aren't reading this anywhere. I don't think you need it.

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+/**
+ * Get the serialized key that corresponds to the action name.
+ * @param {!string} actionName The name of the action.
+ * @return {string} The serialized key or null if the action does not have
+ *     a key mapping.
+ * @package
+ */
+Blockly.user.keyMap.getKeyByActionName = function(actionName) {
```
RoboErikG: Why are you using the name of the action instead of the action object itself?

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+  return null;
+};
+
+/**
+ * Serialize the key event.
+ * @param {!Event} e A key up event holding the key code.
+ * @return {!string} A string containing the serialized key event.
+ */
+Blockly.user.keyMap.serializeKeyEvent = function(e) {
+  var key = "";
```
RoboErikG: nit, single quotes

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+/**
+ * Creates the default key map.
+ * @return {!Object<string,Blockly.KeyAction>} An object holding the default key
+ *     to action mapping.
+ */
+Blockly.user.keyMap.createDefaultKeyMap = function() {
+  //Set W
+  var wAction = new Blockly.KeyAction('previous', 'Goes to the previous location', function() {
```
RoboErikG: These actions should be public constants on the navigation namespace called things like "ACTION_PREVIOUS". That way, developers can call setActionForKey(keycode, Blockly.navigation.ACTION_PREVIOUS).

core/keyboard_nav/key_map.js
```diff
@@ -0,0 +1,250 @@
...
+  });
+
+  //Set ESC
+  var escAction = new Blockly.KeyAction('escape', 'Exit the toolbox', function() {
+    if (Blockly.Navigation.currentState_ === Blockly.Navigation.STATE_TOOLBOX) {
+      Blockly.Navigation.focusWorkspace();
+    }
+  });
+
+  var map = {};
```
RoboErikG: Building this map is the only thing this function should do, using publicly defined actions.

core/keyboard_nav/navigation.js
```diff
@@ -653,127 +654,10 @@ Blockly.Navigation.handleEnterForWS = function() {
  * @return {!boolean} True if the key was handled false otherwise.
  */
 Blockly.Navigation.navigate = function(e) {
```
RoboErikG: Should probably rename this to onKeypress or something similar since it's handling keypress events.

core/keyboard_nav/key_map.js
alschmiedt: Because there are situations in the UI where we only have the action name not the action. Also if you have the action you will always have the action name.

core/keyboard_nav/key_action.js
samelhusseini: {Function}*

core/keyboard_nav/navigation.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_action.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_map.js
alschmiedt: done

core/keyboard_nav/key_action.js
alschmiedt: It seems to appear both ways in code, is there a preference?

rachel-fenichel: https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System

> Also note the difference between `function()` and `Function`. The latter is an instance type and is nullable by default. `function(...)` should be used instead of `Function` whenever possible because it provides more type information about its parameters and return value.

core/keyboard_nav/key_map.js
```diff
@@ -99,18 +93,17 @@ Blockly.user.keyMap.getActionByKeyCode = function(keyCode) {
...
  */
-Blockly.user.keyMap.getKeyByActionName = function(actionName) {
+Blockly.user.keyMap.getKeyByAction = function(action) {
   var keys = Object.keys(Blockly.user.keyMap.map_);
-  for (var i = 0; i < keys.length; i++) {
-    var key = keys[i];
-    var action = Blockly.user.keyMap.map_[key];
-    if (actionName === action.name) {
+  for (var i = 0, key; key = keys[i]; i++) {
+    if (Blockly.user.keyMap.map_[key] === action) {
```
RoboErikG: This won't work with the actions returned by getKeyMap currently since it's making copies of the action objects. For checking equality I think you do want to compare the names.

core/keyboard_nav/key_map.js
```diff
@@ -29,68 +29,62 @@ goog.provide('Blockly.user.keyMap');
...
  *     action mapping.
  * @package
  */
 Blockly.user.keyMap.getKeyMap = function() {
-  return Blockly.user.keyMap.map_;
+  var mapClone = {};
+  var keys = Object.keys(Blockly.user.keyMap.map_);
+  for (var i = 0, key; key = keys[i]; i++) {
+    var action = Blockly.user.keyMap.map_[key];
+    mapClone[key] = new Blockly.Action(action.name, action.desc, action.func);
```
RoboErikG: It's probably sufficient to return the original action in a copy of the map. If they want to modify the action directly they have access to it on the namespace, so it's less likely they'll get it here for that purpose.

core/keyboard_nav/key_map.js
```diff
@@ -123,128 +116,34 @@ Blockly.user.keyMap.getKeyByActionName = function(actionName) {
...
-    key += 'meta';
+  var modifierKeys = Blockly.user.keyMap.modifierKeys;
+  var key = '';
+  for (var i = 0, keyName; keyName = modifierKeys[i]; i++) {
+    if (e.getModifierState(keyName)) {
+      key += keyName;
+    }
   }
   key += e.keyCode;
+  console.log(key);
```
RoboErikG: Nit: remove or wrap with a debug flag.

Issue #2720, opened 2019-07-27
Field.text_: Documentation vs use


NeilFraser:
Our documentation is very clear about not using Field's `.text_` property.

> "Warning: Never access the text_ property directly, instead access it through getText. It is maintained for backwards compatibility."
> "Caution: Never use the text_ property to access or edit the text of your field. It is maintained for backwards compatibility."
https://devsite.googleplex.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating

> "We recommend that you never access or update the text_ property of your field directly."
https://devsite.googleplex.com/blockly/guides/create-custom-blocks/fields/customizing-fields/upgrading

Yet our codebase uses it all over the place:
* Blockly.Block.prototype.toString
* Blockly.FieldAngle.prototype.onMouseMove
* Blockly.FieldDropdown.prototype.doValueUpdate_
* Blockly.FieldImage
* Blockly.FieldImage.prototype.initView
* Blockly.FieldImage.prototype.setAlt
* Blockly.FieldTextInput.prototype.doValueUpdate_
* Blockly.FieldTextInput.prototype.showPromptEditor_
* Blockly.FieldTextInput.prototype.widgetDispose_
* Blockly.FieldVariable.prototype.doValueUpdate_
* Blockly.FieldVariable.prototype.refreshVariableName

Is `.text_` a fundamental part of fields that needs to be used whenever one creates a new field?

BeksOmega:
So here was my thought process (which may or may not be correct):

Setting the text_ property every time a new value gets set is inefficient. For fields that display text it's necessary, but for fields that only display text when they're collapsed you're doing extra work you don't need to do. The [turtle field](https://developers.google.com/blockly/guides/create-custom-blocks/fields/anatomy-of-a-field#other_display_modes) is a good example of this.

Moving outside devs to use getText is also good because it keeps peoples code more organized. Instead of having text_ updates split between doValueUpdate_, doValueInvalid_, and render_ we can tell people, "just handle that inside getText".

Now the reason text_ wasn't removed anywhere in this codebase is because we can't know how, if, where, or why any outside developers are accessing the property and I didn't want to be the one to break everyone. (It remains to be seen whether I did or did not break everyone lol)

If you want to pull text_ out of this codebase, I'm totally down for that (as long as you agree to take part of the blame :P ).

And if you have a different opinion about text_'s use I'd love to discuss =) This just explains what I was thinking when I made these decisions.

samelhusseini:
[...]

Issue #2544, opened 2019-06-05
Colour picker doesn't close on selection


NeilFraser:
Click on a colour picker.
Click on a colour.

Expect colour picker to close (like angle picker, or dropdown).  Instead the picker remains open.

BeksOmega:
I think the default behavior for editors that are rendered inside the DropDownDiv is that they stay open until the user clicks outside of them. So it seems like the angle picker may be the one with deviant behavior.

NeilFraser:
Until now, all widgets have closed immediately upon a selection.  For reference:
  https://3-dot-blockly-games.appspot.com/
(compiled from Blockly 'develop' on 19 Feb 2019)

Colour (movie), angle (bird), notes (music), dropdowns, context menus, and flyouts are all consistent in how they behave.

BeksOmega:
I agree that this is how all widgets acted before, and I also agree that there should be consistency in this area. But I don't believe forcing the DropDownDiv to act like the WidgetDiv is the best way to create consistency.

Firstly, forcing the DropDownDiv to act like the WidgetDiv would not work for all types of editors. For example this turtle field works very well with the default DropDownDiv behavior, and wouldn't work at all with WidgetDiv behavior:

![DropDownDiv_Turtle](https://user-images.githubusercontent.com/25440652/58971460-5d205e00-8770-11e9-91b9-e2183c8be75b.gif)

Secondly, forcing the DropDownDiv to act like the WidgetDiv would /create/ inconsistency whenever a new field was created. Outside developers would have to spend time writing custom code to make it conform to Blockly's design, whereas if Blockly's DropDownDivs used the default behavior, outside developers wouldn't have to worry about this issue.

I believe that since the DropDownDiv has a very distinct look, it is reasonable to give those editors their own internally consistent behavior, and that it will not be confusing to users.

I think that the angle field could be easily changed so that you click and drag the selector, rather than having it activate on hover, which would make it easy to move to the default DropDownDiv behavior. Overrall it may be a small thing, but I believe it will make life easier for future developers.

(Ok I'll drop this now, but if you do decide to force the DropDownDiv to act like the WidgetDiv you may want to update the Date field as well)

RoboErikG:
For reference, the angle field still closes on click because Rachel [added a click handler that hides the DropDownDiv](https://github.com/google/blockly/blame/86bba8127603d57e533041c0808b55e58bb88049/core/field_angle.js#L217)

A few thoughts:

[...]

PR #2796, opened 2019-08-09
text_ Refactor Demo

BeksOmega:
This is a demonstration of one method for refactoring the text_ property of fields, as discussed in this issue: https://github.com/google/blockly/issues/2720

I have posted it as a PR because I figured that was the easiest way to discuss the changes. If there is a better place to put this please let me know and I will change it :D

I have also added explanatory comments throughout the code that are meant to be read _in order_. Please look through each commit in this PR in the order that it was committed. Namely:
1) Turtle field explanation.
2) Added support for displaying invalid values to the date field.
3) Added display value support to the abstract field class.
4) Fixed the text input field.

Note for @samelhusseini  I do believe we are on the same page about how the Turtle field works, but I wanted to include an explanation of it just in case. I also wanted to provide it as a place to ask questions.

Thank you for taking the time to look at this*! I really appreciate everyone's willingness to hear my ideas!

*Don't feel like you have to look at it today though, I do realize it's Friday :P

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -725,7 +725,10 @@ Blockly.Field.prototype.getScaledBBox_ = function() {
  * @protected
  */
 Blockly.Field.prototype.getDisplayText_ = function() {
-  var text = this.text_;
+  // The field will now generate the display text based on the display value.
+  // The text no longer needs to be set explicitly. Booyah!
+  var text = this.displayValue_.toString();
+
   if (!text) {
     // Prevent the field from disappearing if empty.
     return Blockly.Field.NBSP;
@@ -885,6 +888,10 @@ Blockly.Field.prototype.doClassValidation_ = function(newValue) {
  * @protected
  */
 Blockly.Field.prototype.doValueUpdate_ = function(newValue) {
+  // If the value is valid that is the value we want to be displayed. So set
+  // the display value to the new value.
+  this.displayValue_ = newValue;
+
   this.value_ = newValue;
   this.isDirty_ = true;
   // For backwards compatibility.
@@ -900,7 +907,14 @@ Blockly.Field.prototype.doValueUpdate_ = function(newValue) {
  */
 Blockly.Field.prototype.doValueInvalid_ = function(_invalidValue) {
   // NOP
+
+  // This function still does nothing. This means that by default fields
+  // will still do nothing when they recieve an invalid input. Meaning that
+  // they continue to display the last valid value.
 };
+
+// Now our date picker should work as expected. And it does:
+// https://i.imgur.com/nLTyqBc.gif
 
 /**
  * Handle a mouse down event on a field.
--- a/core/field_date.js
+++ b/core/field_date.js
@@ -116,18 +116,64 @@ Blockly.FieldDate.prototype.doClassValidation_ = function(opt_newValue) {
   return opt_newValue;
 };
 
+// Now let's see if we can implement this on a field that uses the default UI.
+// To demonstrate this I'm going to add support for showing invalid values
+// to the date field.
+
+Blockly.FieldDate.prototype.doValueUpdate_ = function(newValue) {
+  Blockly.FieldDate.superClass_.doValueUpdate_.call(this, newValue);
+
+  // This is used to tell if the borderRect should be rendered in red.
+  this.isValueValid_ = true;
+
+  // As an aside, isValueValid_ could be moved to the abstract field class if
+  // desired. Moving it there would not affect behavior because the default
+  // rendering would not take it into account. But it would be available if
+  // subclasses wanted to use it.
+};
+
+Blockly.FieldDate.prototype.doValueInvalid_ = function(invalidValue) {
+  // This is what will allow us to display invalid values.
+  this.displayValue_ = invalidValue;
+
+  // This is used to tell if the borderRect should be rendered in red.
+  this.isValueValid_ = false;
+
+  // Confirm that we want a rerender.
+  this.isDirty_ = true;
+};
+
 /**
  * Render the field. If the picker is shown make sure it has the current
  * date selected.
  * @protected
  */
 Blockly.FieldDate.prototype.render_ = function() {
+  // See final comment.
   Blockly.FieldDate.superClass_.render_.call(this);
+
+  // Update the borderRect.
+  if (!this.isValueValid_) {
+    this.borderRect_.style.fill = '#faa';
+    this.borderRect_.style.fillOpacity = 1;
+  } else {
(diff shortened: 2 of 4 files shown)


core/field_textinput.js
```diff
@@ -278,11 +311,17 @@ Blockly.FieldTextInput.prototype.widgetDispose_ = function() {
...
   // No need to call setValue because if the widget is being closed the
   // latest input text has already been validated.
-  if (this.value_ !== this.text_) {
+
+  // This handles the cleaning up of the text input editor. If the display
+  // value is currently different than the actual value, we want to rectify
+  // that now.
+  // As stated above setValue has already been called based on the final
+  // input from the user.
+  if (this.value_ !== this.displayValue_) {
```
RoboErikG: This is almost the right behavior. Instead we can just set displayValue_ to null as the dispose and handle it in getDisplayText as commented above.

core/field_textinput.js
```diff
@@ -365,12 +404,77 @@ Blockly.FieldTextInput.prototype.onHtmlInputChange_ = function(_e) {
...
+ *   value_ = newValue;
+ *   tempDisplayValue_ = null;
+ * }
+ *
+ * getDisplayText() {
+ *   var text = tempDisplayValue || value;
+ * }
+ */
+
+// I would support this =)
```
RoboErikG: This was our original proposal. You may have found a better place to clear the tempDisplayValue_ in dropdownDispose_. If we set tempDisplayValue_ to null there it's a nice default of clearing it when the editor disappears. =)

core/field_textinput.js
```diff
@@ -365,12 +404,77 @@ Blockly.FieldTextInput.prototype.onHtmlInputChange_ = function(_e) {
...
+// Let me go ahead and speak to that for a moment.
+
+// Most of the problems with text_ have already been fixed prior to this PR.
+//   1) Setting the text_ property (for use in collapse text) was inefficient.
+//   This has been solved by moving people to the getText property.
+//   2) Moving peoples code to getText keeps it more organized, instead of
+//   having it split between doValueUpdate, doValueInvalid_ and render_.
+// As you know I also noted these facts in my first comment:
+// https://github.com/google/blockly/issues/2720#issuecomment-515688519
+
```
RoboErikG: The third reason for getting rid of text_ is to make it clear to developers this is an intentional change in behavior and usage. We think using text_ as a way to override what's being displayed was broken previously and we are now providing an explicit way to do this override.

BeksOmega:
Ope, sorry @samelhusseini I didn't see that you still wanted to take a look.

samelhusseini:
Yeah, thanks for the demo @BeksOmega. I haven't forgotten about this, I'm just having a think about it a little. Part of the problem is there are multiple ways of achieving the same thing, so trying to think of developers and what they would prefer, but also trying to think of it from an architectural stand point and how we can ensure each field class is only accessing things that belong to it.
What you've done here is create a property ``displayValue_`` that is a reflection of what gets shown on the screen. You point out yourself that it can be repetitive for fields that don't need the value that's different. My problem with that is that it's easy to lose track of the value, and for developers extending fields, its easy not to set and reset it in the right place. 
Let me have a little more of a think about it, and get back to you. My inclination is to figure out what the role of each field class/subclass is. One thing to point out, in the turtle field it doesn't make use of getDisplayText_, to me that's an indication that the field subclass has diverged away from using the built in functionality and is managing its own rendering. Even though the field makes use of initView (to create the text element), it's mostly using that out of convenience. 

If the field class is to be in charge of creating it's own DOM elements, and managing their lifecycle, then that should be more explicit and obvious to developers what these lifecycle methods are.
doValueUpdate, and doValueInvalid are steps in the right direction.

BeksOmega:
>what these lifecycle methods are

I could try and give you my current understanding if you want. (Which is not necessarily the correct understanding lol) I know you asked before:

> why doesn't setValue set this.isDirty and the this.value and instead rely on doValueUpdate to do it?

which I haven't gotten around to answering yet. (I promise there is a reason! I'm just terrible at having two conversations at once)

BeksOmega:
>what these lifecycle methods are

I could try and give you my current understanding if you want. (Which is not necessarily the correct understanding lol) I know you asked before:

>why doesn't setValue set this.isDirty and the this.value and instead rely on doValueUpdate to do it?

which I haven't gotten around to answering yet. (I promise there is a reason! I'm just terrible at having two conversations at once)

BeksOmega:
Here's what I'm trying to solve for:
text_

I'm trying to solve for the blue part of the diagram.

I think you're trying to (primarily) solve for the purple part of the diagram.

In this demo I show a way that the purple part can be built ontop of the blue part.

Note: That the blue part has already mostly been solved by adding the doValueInvalid_ method. This just completes it.

core/field.js
```diff
@@ -885,6 +888,10 @@ Blockly.Field.prototype.doClassValidation_ = function(newValue) {
  * @protected
  */
 Blockly.Field.prototype.doValueUpdate_ = function(newValue) {
+  // If the value is valid that is the value we want to be displayed. So set
+  // the display value to the new value.
+  this.displayValue_ = newValue;
```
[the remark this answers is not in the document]
BeksOmega: As shown in the diagram I posted below, this is solving for fields that want to show invalid values.

> case where the validator strips half the text out
That case involves a changed value.

Most fields (i.e. all fields except the text input) should accept changed values, since they are valid.

Ignoring changed values can be built ontop of this system, as I show in this demo.

core/field_date.js
```diff
@@ -116,19 +116,65 @@ Blockly.FieldDate.prototype.doClassValidation_ = function(opt_newValue) {
...
+
+  // As an aside, isValueValid_ could be moved to the abstract field class if
+  // desired. Moving it there would not affect behavior because the default
+  // rendering would not take it into account. But it would be available if
+  // subclasses wanted to use it.
+};
+
+Blockly.FieldDate.prototype.doValueInvalid_ = function(invalidValue) {
+  // This is what will allow us to display invalid values.
+  this.displayValue_ = invalidValue;
```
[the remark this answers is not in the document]
BeksOmega: Date fields close automatically:
https://i.imgur.com/nLTyqBc.gif

As was decided here: https://github.com/google/blockly/issues/2544

But that functionality could be built ontop of this system.

Fields that want to wipe the display value when the editor closes (i.e. text input fields) can do so, as shown in this demo.

samelhusseini:
@BeksOmega just to let you know. I'm doing a deep dive into how each of the fields are using this.text_ today. I'll report back with my findings.

PR #2798, opened 2019-08-10
new feature: support hex number like 0xfff #1306

### LINKED ISSUE #1306: Blockly.FieldNumber does not support HEX
I made a optimization to allow use input hex in FieldNumber

Blockly.FieldNumber.prototype.HEXExp = /^0[xX]([0-9A-Fa-f])+$/g;
/**
 * Ensure that only a number in the correct range may be entered.
 * @param {string} text The user's text.
 * @return {?string} A string representing a valid number, or null if invalid.
 */
Blockly.FieldNumber.prototype.classValidator = function(text) {
    if (text === null) {
        return null;
    }
    text = String(text);
    // TODO: Handle cases like 'ten', '1.203,14', etc.
    // 'O' is sometimes mistaken for '0' by inexperienced users.
    text = text.replace(/O/ig, '0');
    // Strip out thousands separators.
    text = text.replace(/,/g, '');
    var hex = text.match(Blockly.FieldNumber.prototype.HEXExp);
    var n = hex ? parseInt(text, 16) : parseFloat(text || 0);
    if (isNaN(n)) {
        // Invalid number.
        return null;
    }
    // Round to nearest multiple of precision.
    if (this.precision_ && isFinite(n)) {
        n = Math.round(n / this.precision_) * this.precision_;
    }
    // Get the value in range.
    n = goog.math.clamp(n, this.min_, this.max_);
    return hex ? text : String(n);
};

duzc2:
## The basics

- [x] I branched from develop

### Resolves

#1306

### Reason for Changes

#1306

### Test Coverage

Tested on:
* Desktop Chrome


duzc2:
#1306

core/field_number.js
```diff
@@ -119,9 +119,8 @@ Blockly.FieldNumber.prototype.doClassValidation_ = function(opt_newValue) {
   newValue = newValue.replace(/O/ig, '0');
   // Strip out thousands separators.
   newValue = newValue.replace(/,/g, '');
-
-  // Clean up number.
-  var n = Number(newValue || 0);
+  var hex = newValue.match(/^0xX+$/g);
```
RoboErikG: This doesn't do what you think it does. This match is looking for start of the line followed by '0x' followed by 1 or more 'X' characters followed by the end of the line. So it would match '0xXXXXXXXXXXXX' but not '0xabc'.

core/field_number.js
```diff
@@ -119,9 +119,8 @@ Blockly.FieldNumber.prototype.doClassValidation_ = function(opt_newValue) {
...
+  var n = hex ? parseInt(newValue, 16) : Number(newValue || 0);
```
RoboErikG: This also isn't necessary, Number('0xabc') and parseInt('0xabc', 16) both return the same value, 2748. Converting a 3 letter hex color to a 6 letter number is a bit trickier than this.

RoboErikG:
This was already fixed by https://github.com/google/blockly/pull/2758

Issue #636, opened 2016-09-23
BlocklyFactory: When a block has an output type, blockly factory doesn't recognize a save


picklesrus:
When you attach an output type to the factory base block in blockly factory, and click the update button to save changes to the block, it doesn't recognize the save and the update button stays highlighted as if you never clicked it.

Looking at the xml of the workspace, it seems like clicking the update button is causing the id of the "any" shadow block to change. i.e. if you do a workspaceToDom then click the update button and do another workspaceToDom, the ids of the "any" shadow block are different.  Not sure why yet though.

picklesrus:
This turns out to be a rather odd interaction between shadow blocks and the mutating factory_base block trying to spawn shadow blocks as it mutates.
When you add a connection in the container block and replace the "type" shadow block with a regular block, the xml you get from blockToDom for the output portion looks something like this (this is what we put in local storage):
<value name="OUTPUTTYPE">
  <block type="type_boolean" id="-Gj,l7rh(YYFU:1CfK:I"></block>
  <shadow type="type_null" id="teL,2l7V+X;c+%%O7{ev"></shadow>
</value>

There is also code in the factory base block that creates a new shadow block using workspace.newBlock.  That code is there to spawn a shadow block when the user adds an output connection in the base block.  See https://github.com/google/blockly/blob/develop/demos/blockfactory/blocks.js#L101

So, when you load into block factory workspace from xml, the shadow block mentioned above is the one that ends up winning and the code to compare the current workspace with what's in storage fails reports a difference.

There are a few ways I can think to side-step this problem, but it is kind of interesting because any of manipulation like this in the block definition is likely to alter the xml on load so perhaps worth discussion.

picklesrus:
Fixed in develop

NielsH12:
I'm still experiencing this issue. Didn't it get fixed?

RoboErikG:
Looks like it's back. Possibly related to removal of ids in the stored xml, but needs investigation.

samelhusseini:
Fixed by #2801

PR #2801, opened 2019-08-12
Fixed Block Factory XML not Getting Properly Cleaned

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#636 

### Proposed Changes

Makes it so that all unimportant attributes (x, y, id, xmlns) are removed from the xml before comparing.

### Reason for Changes / Underlying Problem

The reason diffing was failing before is that an output shadow block was being created with a different ID before the one in XML could be loaded.

This is because validators are now called whenever a value is loaded from XML, which makes [this](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/demos/blockfactory/blocks.js#L48) trigger the creation of a new shadow.

The changes proposed negate this problem by ignoring the ID when diffing.

### Test Coverage

Could not replicated #636

Tested on:
* Desktop Chrome

### Documentation

N/A

### Additional Information

The [trashcan ](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/core/trashcan.js#L503)and the blockfactory are now both doing this cleaning/diffing combo. It may be worth it at some point in the future to create explicit workspace diffing functions. If there was such a system cross-save-format diffing could also be added (e.g. is this JSON the same as this XML) if other save formats are ever added.

### DIFF
--- a/demos/blockfactory/factory_utils.js
+++ b/demos/blockfactory/factory_utils.js
@@ -921,7 +921,7 @@ FactoryUtils.sameBlockXml = function(blockXml1, blockXml2) {
   // Each XML element should contain a single child element with a 'block' tag
   if (blockXml1.tagName.toLowerCase() != 'xml' ||
       blockXml2.tagName.toLowerCase() != 'xml') {
-    throw new Error('Expected two XML elements, recieved elements with tag ' +
+    throw new Error('Expected two XML elements, received elements with tag ' +
         'names: ' + blockXml1.tagName + ' and ' + blockXml2.tagName + '.');
   }
 
@@ -948,6 +948,12 @@ FactoryUtils.sameBlockXml = function(blockXml1, blockXml2) {
   return blockXmlText1 == blockXmlText2;
 };
 
+/**
+ * Strips the provided xml of any attributes that don't describe the
+ * 'structure' of the blocks (i.e. block order, field values, etc).
+ * @param {Node} xml The xml to clean.
+ * @return {Node}
+ */
 FactoryUtils.cleanXml = function(xml) {
   var newXml = xml.cloneNode(true);
   var node = newXml;
--- a/demos/blockfactory/factory_utils.js
+++ b/demos/blockfactory/factory_utils.js
@@ -926,7 +926,7 @@ FactoryUtils.sameBlockXml = function(blockXml1, blockXml2) {
   }
 
   // Compare the block elements directly. The XML tags may include other meta
-  // information we want to igrore.
+  // information we want to ignore.
   var blockElement1 = blockXml1.getElementsByTagName('block')[0];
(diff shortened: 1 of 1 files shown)


RoboErikG:
@NeilFraser had been doing some work on this. He's out this week but I'd like to get his opinion before committing so expect a response next week.

demos/blockfactory/factory_utils.js
```diff
@@ -926,16 +926,19 @@ FactoryUtils.sameBlockXml = function(blockXml1, blockXml2) {
...
-  var blockXmlText1 = Blockly.Xml.domToText(blockElement1);
-  var blockXmlText2 = Blockly.Xml.domToText(blockElement2);
+  var cleanBlockXml1 = FactoryUtils.cleanXml(blockElement1);
+  var cleanBlockXml2 = FactoryUtils.cleanXml(blockElement2);
+
+  var blockXmlText1 = Blockly.Xml.domToText(cleanBlockXml1);
+  var blockXmlText2 = Blockly.Xml.domToText(cleanBlockXml2);
 
   // Strip white space.
   blockXmlText1 = blockXmlText1.replace(/\s+/g, '');
```
NeilFraser: Hmm, this is a bug.  Not in your PR, but pre-existing.  It means that a text block (e.g. tooltip) with the value " " is the same as a text block with the value " &nbsp; &nbsp; &nbsp; ".

Is stripping whitespace even needed?

BeksOmega: Dug into it, it seems that the stripping of whitespace is covering a different bug.

The blockXml1 is the xml currently stored by the blockfactory. It is stored as pretty text to make it easy to export (I assume).

When you do:
* block/workspaceToDom
* domToPrettyText
* textToDom <- I think the DOM retains whitespace characters here
* domToText

You end up with pretty text at the end, instead of non-pretty text as you'd expect.

This means that the blockXml2, which goes:
* workspaceToDom
* domToText

is not equivalent.

I tried converting the blockXml2 into pretty text instead of normal text, but that doesn't work either (I assume something about the round tripping messes up the saved text's indentation).

There are a couple of options I've come up with:

1) Do a "direct" equality check by walking both dom trees and checking for attribute equality. This gets rid of the need to worry about whitespace (hopefully).
2) Store the block's xml as plain text/actual xml instead of pretty text. Convert it to pretty text only when getBlockLibrary is called.
3) Try to fix round tripping pretty text (probably better saved for another PR).

Or some combination of the three.

BeksOmega:
Also got that JSDoc added!

NeilFraser:
Looks good.

Don't worry about the pre-existing whitespace bug.  It is extremely unlikely to occur, and easy to work around if it does.  Thanks for looking into that and discovering it's not trivial to fix.

PR #2802, opened 2019-08-12
Math.sign -> Math.sin

samelhusseini:
There was a typo at some point that introduced this. 
Note for all,`` Math.sign`` is not available on IE, and needs a polyfill if we want to use it.

In this case, it was just a typo, and ``sin`` was the intended method.

## The basics

- [x] I branched from develop


rachel-fenichel:
d'oh

PR #2806, opened 2019-08-12
Hats and tabs

rachel-fenichel:
## The basics
This is an updated version of https://github.com/google/blockly/pull/2790
I split some changes out into https://github.com/google/blockly/pull/2804 and https://github.com/google/blockly/pull/2805, so you should review those first.

- [x] I branched from develop

### Proposed Changes
- Makes all constants and highlight constants use relative svg commands instead of absolute ones.
- Adds startX and startY values to the RenderInfo object and uses them in drawing and highlighting.
- Uses startX and startY to make all parts of a block (including tabs and hats) draw down and to the right from the block's position on the workspace.
- Adds some debug rendering (see #2805)
- Fixes calculation of block width when blocks have tabs.
- Adds an `xPos` property to rows.  I decided this is worth it to make sure that a row has all of the information needed to position it and its contents.

### Reason for Changes
The constant shapes should not assume that they know where they'll be placed in absolute terms. For instance, a hat should not assume that it starts at 0,0. An output connection should not assume that either.

### Additional information
There are a lot of different values of 'width' in play here:
- The width of the block including child blocks attached to value inputs
- The width of the block including child blocks attached to statement inputs
- The width of the block including the output tab
- The width of the block except for the output tab

### DIFF
--- a/core/renderers/block_rendering_rewrite/block_render_draw.js
+++ b/core/renderers/block_rendering_rewrite/block_render_draw.js
@@ -133,7 +133,7 @@ Blockly.blockRendering.Drawer.prototype.drawTop_ = function() {
   }
   this.positionPreviousConnection_();
   this.steps_.push(
-      Blockly.utils.svgPaths.moveBy(this.info_.startX, 0));
+      Blockly.utils.svgPaths.moveBy(this.info_.startX, this.info_.startY));
   for (var i = 0, elem; elem = elements[i]; i++) {
     if (elem.type === 'round corner') {
       this.steps_.push(Blockly.blockRendering.constants.OUTSIDE_CORNERS.topLeft);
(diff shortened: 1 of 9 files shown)


core/renderers/block_rendering_rewrite/block_render_draw.js
```diff
@@ -132,20 +132,19 @@ Blockly.blockRendering.Drawer.prototype.drawTop_ = function() {
...
+      this.steps_.push(Blockly.blockRendering.constants.OUTSIDE_CORNERS.topLeft);
     } else if (elem.type === 'previous connection') {
-      this.steps_.push(Blockly.blockRendering.constants.NOTCH.pathLeft);
+      this.steps_.push(topRow.notchShape.pathLeft);
     } else if (elem.type === 'hat') {
       this.steps_.push(Blockly.blockRendering.constants.START_HAT.path);
     } else if (elem.isSpacer()) {
       this.steps_.push('h', elem.width);
     }
+    // No branch for a square corner, because it's a no-op.
```
RoboErikG: I'm a little worried about losing information by removing the branch. Maybe put 'square corner' in quotes so it's clear that's the type that is dropped?

rachel-fenichel: Added the quotes in my changes for my next PR.

I may be able to get rid of the square corner entirely!  I think it was mostly holding the startX and startY information

PR #2807, opened 2019-08-13
package.json no longer private

samelhusseini:
Remove private attribute from package.json as the package is now published on ``npm``.


rachel-fenichel:
lgtm

PR #2809, opened 2019-08-13
Include next connection in block's bounding box

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Next connection was not correctly part of the height of the last row.

### Proposed Changes

- Give the bottom row the concept of overhang, and adjust drawing of the bottom of the block accordingly.
- Remove more explicit references to connection shape objects, and use the connection measurable's connectionShape property instead.  Tested by replacing the puzzle tab with a different shape in measurables.js:
```
Blockly.blockRendering.Measurable.prototype.connectionShape =
    Blockly.blockRendering.constants.TRIANGLE;
```
### Reason for Changes

- Block height and width should include all parts of the block.
- All input and output connections should be drawn the same way right now.

### Test Coverage
Tested on chrome in the playground.

### Additional Information

Followup: look into getHeightWidth in block_render_svg, and move it block_svg.

### DIFF
--- a/core/renderers/block_rendering_rewrite/measurables.js
+++ b/core/renderers/block_rendering_rewrite/measurables.js
@@ -311,7 +311,7 @@ Blockly.blockRendering.ExternalValueInput = function(input) {
   this.type = 'external value input';
 
   if (!this.connectedBlock) {
-    this.height = Blockly.blockRendering.constants.TAB_HEIGHT;
+    this.height = this.connectionShape.height;
   } else {
     this.height =
         this.connectedBlockHeight - 2 * Blockly.blockRendering.constants.TAB_OFFSET_FROM_TOP;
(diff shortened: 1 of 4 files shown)


core/renderers/block_rendering_rewrite/measurables.js
```diff
@@ -587,3 +588,18 @@ Blockly.blockRendering.BottomRow.prototype.getNextConnection = function() {
...
 };
+
+Blockly.blockRendering.BottomRow.prototype.measure = function() {
+  for (var e = 0; e < this.elements.length; e++) {
+    var elem = this.elements[e];
+    this.width += elem.width;
+    if (!(elem.isSpacer())) {
+      if (elem.type == 'next connection') {
+        this.height = this.height + elem.height;
+        this.overhangY = elem.height;
```
RoboErikG: This assumes the next connection only extends down from the bottom. What if it extends up into the block as well?

rachel-fenichel: Nothing good.  Right now prev/next connections have to go down and output/input connections have to go left.  If we want to allow anything else I think I should make an issue for that so we can discuss.

RoboErikG: Yeah, please file that for discussion. I think it wouldn't be unreasonable to allow.

PR #2813, opened 2019-08-13
Add Blockly aria utils

samelhusseini:
Adding Blockly utils file for accessibility roles and state.

Similar to https://github.com/google/closure-library/blob/master/closure/goog/a11y/aria/aria.js

FYI @alschmiedt 

The two public methods to be used here are: 
``Blockly.utils.aria.setRole`` and ``Blockly.utils.aria.setState``.

### DIFF
--- a/core/utils/aria.js
+++ b/core/utils/aria.js
@@ -143,13 +143,13 @@ Blockly.utils.aria.Role = {
   MENUBAR: 'menubar',
 
   // ARIA role for menu item elements.
-  MENU_ITEM: 'menuitem',
+  MENUITEM: 'menuitem',
 
   // ARIA role for a checkbox box element inside a menu.
-  MENU_ITEM_CHECKBOX: 'menuitemcheckbox',
+  MENUITEMCHECKBOX: 'menuitemcheckbox',
 
   // ARIA role for a radio button element inside a menu.
-  MENU_ITEM_RADIO: 'menuitemradio',
+  MENUITEMRADIO: 'menuitemradio',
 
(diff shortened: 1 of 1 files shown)


core/utils/aria.js
```diff
@@ -0,0 +1,470 @@
...
+ * ARIA role values.
+ * Copied from Closure's goog.a11y.aria.Role
+ * @enum {string}
+ */
+Blockly.utils.aria.Role = {
+  // ARIA role for an alert element that doesn't need to be explicitly closed.
+  ALERT: 'alert',
+
+  // ARIA role for an alert dialog element that takes focus and must be closed.
+  ALERTDIALOG: 'alertdialog',
```
moniika: Although these were copied from closure, some of the constants aren't following the style guide because constants should use CONSTANT_CASE (separating words with underscore).
Since we are redefining them, I think changing the names should be fine, unless there's a reason why some, like MENU_ITEM, use constant case and others don't.

samelhusseini: Makes sense, I'll update.

samelhusseini: Actually I think the style guide is for constants, and not Enums. I think there's something to be said here about not updating them. These are meant to be 1:1 mappings to strings, just all caps, making sure the compiler catches any typos. 
If we change them, then you constantly need to think is checkbox, CHECK_BOX or CHECKBOX.

moniika: technically guide defines the values inside an enum as constants, but 1:1 mapping makes sense and as long as we apply decision consistently (like with MENU_ITEM) I am happy with change.

PR #2815, opened 2019-08-13
Add a singleton id generator util

samelhusseini:
This is very similar to closure's id generator. 

It's a great way to get a unique id for any components, just call
``Blockly.utils.IdGenerator.getInstance().getNextUniqueId()``

Similar to closure's https://github.com/google/closure-library/blob/master/closure/goog/ui/idgenerator.js
(We just use the single prefix for blockly)

### DIFF
--- a/core/utils/idgenerator.js
+++ b/core/utils/idgenerator.js
@@ -1,0 +1,64 @@
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2019 Google Inc.
+ * https://developers.google.com/blockly/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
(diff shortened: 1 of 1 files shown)


RoboErikG:
For clarification, is this for naming dom elements uniquely? For UUIds in general there's already Blockly.utils.genUid.

samelhusseini:
This is for generating unique DOM ids. Although, I can't think of a reason we couldn't use UUIDs instead (except that they're long).

RoboErikG:
I guess I was asking is there a reason to have the blockly prefix or can we just reuse the existing UID generator? Is this being used to make ids that would appear alongside other named dom elements when inspecting a page or is this only used in code?

If it's generally useful, lgtm. If it's just what Closure was using I'd prefer we switch to the existing genUid method.

samelhusseini:
Yep, I get the concern. I hadn't seen that method, and this is me just doing what closure does, but looking at this more, here are my thoughts: 

These ids that the id generator generates end up in the dom and are referred to by aria attributes.
HTML4 limitation here restricts ids to the following set:  ASCII letters, digits, '_', '-' and '.'
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

Blockly.utils.genUid.soup_ has a much wider set than that, so we'd have to adapt that to make this work. 

In addition, since this ends up in the dom. It will make debugging easier if we the id is legible, and not some UUID. 

My thoughts are we need this, but I think we can move the Blockly.utils.genUid into this idGenerator file and add comments to indicate the difference between
``genUid`` and ``getNextUniqueId``.

RoboErikG:
genUid is a public API, so I think we'll need to leave it where it is. As a followup I'd move this next to genUid, but this PR is fine for now so we can get the toolbox in.

RoboErikG:
Err, typed too fast, this file is already in utils, so just explaining the difference is good.

PR #2818, opened 2019-08-13
Foundation for no-closure toolbox

samelhusseini:
A subset of closure tree components for building a no-closure toolbox.

Note, the build will fail until the following PRs are checked in: 
IdGenerator: https://github.com/google/blockly/pull/2815
Aria: https://github.com/google/blockly/pull/2813
KeyCode: https://github.com/google/blockly/pull/2814
Style: https://github.com/google/blockly/pull/2816

These are the files needed to create a no-closure toolbox. This doesn't integrate with toolbox.js yet, that will be a different PR. 

This is a subset of the closure components (goog.ui.component, goog.ui.tree.BaseNode, goog.ui.tree.TreeNode, and goog.ui.tree.TreeControl) stripped down to only support functionality that's used in Blockly. 

Another key difference with how closure renders the toolbox and us. In order to not have to add functionality to parse HTML. We build the DOM tree up using JavaScript. This is more secure than using innerHTML, and we use textContent to set the label. ie: we don't support any HTML content for the tree labels, and we don't run the risk of bad html being injected. see https://github.com/google/blockly/pull/2818/files#diff-69d2bd0b6082349c1924b57268aea82aR708
 

- [x] Run the closure compiler and make sure there are no errors / warnings.

### DIFF
--- a/core/components/component.js
+++ b/core/components/component.js
@@ -71,7 +71,7 @@ Blockly.Component = function() {
   /**
    * Parent component to which events will be propagated.  This property is
    * strictly private and must not be accessed directly outside of this class!
-   * @private {Blockly.Component?}
+   * @private {?Blockly.Component}
    */
   this.parent_ = null;
 
@@ -157,6 +157,7 @@ Blockly.Component.Error = {
  * right to left determinations.
  * @param {?boolean} rightToLeft Whether the components should be rendered
  *     right-to-left. Null iff components should determine their directionality.
+ * @protected
  */
 Blockly.Component.setDefaultRightToLeft = function(rightToLeft) {
   Blockly.Component.defaultRightToLeft_ = rightToLeft;
@@ -166,6 +167,7 @@ Blockly.Component.setDefaultRightToLeft = function(rightToLeft) {
  * Gets the unique ID for the instance of this component.  If the instance
  * doesn't already have an ID, generates one on the fly.
  * @return {string} Unique component ID.
+ * @protected
  */
 Blockly.Component.prototype.getId = function() {
   return this.id_ || (this.id_ = this.idGenerator_.getNextUniqueId());
@@ -178,6 +180,7 @@ Blockly.Component.prototype.getId = function() {
  * new ID; this may throw an error if the parent already has a child with an ID
  * that conflicts with the new ID.
  * @param {string} id Unique component ID.
+ * @protected
  */
 Blockly.Component.prototype.setId = function(id) {
   if (this.parent_ && this.parent_.childIndex_) {
@@ -193,6 +196,7 @@ Blockly.Component.prototype.setId = function(id) {
 /**
  * Gets the component's element.
  * @return {Element} The element for the component.
+ * @package
  */
 Blockly.Component.prototype.getElement = function() {
   return this.element_;
@@ -209,6 +213,7 @@ Blockly.Component.prototype.getElement = function() {
  * This should only be used by subclasses and its associated renderers.
  *
  * @param {Element} element Root element for the component.
+ * @protected
  */
 Blockly.Component.prototype.setElementInternal = function(element) {
   this.element_ = element;
@@ -223,6 +228,7 @@ Blockly.Component.prototype.setElementInternal = function(element) {
  * @see Blockly.Component#removeChild
  * @see Blockly.Component#removeChildAt
  * @param {Blockly.Component} parent The parent component.
+ * @protected
  */
 Blockly.Component.prototype.setParent = function(parent) {
   if (this == parent) {
@@ -242,7 +248,8 @@ Blockly.Component.prototype.setParent = function(parent) {
 
 /**
  * Returns the component's parent, if any.
- * @return {Blockly.Component?} The parent component.
+ * @return {?Blockly.Component} The parent component.
+ * @protected
  */
 Blockly.Component.prototype.getParent = function() {
   return this.parent_;
@@ -251,6 +258,7 @@ Blockly.Component.prototype.getParent = function() {
 /**
  * Determines whether the component has been added to the document.
  * @return {boolean} TRUE if rendered. Otherwise, FALSE.
+ * @protected
  */
 Blockly.Component.prototype.isInDocument = function() {
   return this.inDocument_;
@@ -259,6 +267,7 @@ Blockly.Component.prototype.isInDocument = function() {
 /**
  * Creates the initial DOM representation for the component.  The default
  * implementation is to set this.element_ = div.
+ * @protected
  */
 Blockly.Component.prototype.createDom = function() {
   this.element_ = document.createElement('div');
@@ -278,6 +287,7 @@ Blockly.Component.prototype.createDom = function() {
  *
  * @param {Element=} opt_parentElement Optional parent element to render the
  *    component into.
+ * @protected
  */
 Blockly.Component.prototype.render = function(opt_parentElement) {
   this.render_(opt_parentElement);
@@ -290,6 +300,7 @@ Blockly.Component.prototype.render = function(opt_parentElement) {
  * Throws an Error if the component is already rendered.
  *
  * @param {Node} sibling Node to render the component before.
+ * @protected
  */
 Blockly.Component.prototype.renderBefore = function(sibling) {
   this.render_(/** @type {Element} */ (sibling.parentNode), sibling);
@@ -345,6 +356,7 @@ Blockly.Component.prototype.render_ = function(
  *
  * If the component contains child components, this call is propagated to its
  * children.
+ * @protected
  */
 Blockly.Component.prototype.enterDocument = function() {
   this.inDocument_ = true;
@@ -370,6 +382,7 @@ Blockly.Component.prototype.enterDocument = function() {
  *
  * It should be possible for the component to be rendered again once this method
  * has been called.
+ * @protected
  */
 Blockly.Component.prototype.exitDocument = function() {
   // Propagate exitDocument to child components that have been rendered, if any.
@@ -387,6 +400,7 @@ Blockly.Component.prototype.exitDocument = function() {
  * {@link #disposeInternal}.
  *
  * @return {void} Nothing.
+ * @protected
  */
 Blockly.Component.prototype.dispose = function() {
   if (!this.disposed_) {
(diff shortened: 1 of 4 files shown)


core/components/component.js
```diff
@@ -0,0 +1,816 @@
...
+    // to make sure we render the child component's element into the correct
+    // parent element (otherwise render_ with a null first argument would
+    // render the child into the document body, which is almost certainly not
+    // what we want).
+    if (!this.element_) {
+      this.createDom();
+    }
+    // Render the child into the parent at the appropriate location.  Note that
+    // getChildAt(index + 1) returns undefined if inserting at the end.
+    // TODO(attila): We should have a renderer with a renderChildAt API.
```
rachel-fenichel: Remove somebody else's todo, here and on 601.  We're not going to fix them.

core/components/tree/basenode.js
```diff
@@ -0,0 +1,1151 @@
...
+ * @type {!Array<!Blockly.tree.BaseNode>}
+ * @private
+ */
+Blockly.tree.BaseNode.EMPTY_CHILDREN_ = [];
+
+/**
+ * @param {number} index 0-based index.
+ * @return {Blockly.tree.BaseNode} The child at the given index; null if none.
+ */
+Blockly.tree.BaseNode.prototype.getChildAt;
```
rachel-fenichel: ???

core/components/tree/basenode.js
```diff
@@ -0,0 +1,1151 @@
...
+  }
+  return false;
+};
+
+/**
+ * An array of empty children to return for nodes that have no children.
+ * @type {!Array<!Blockly.tree.BaseNode>}
+ * @private
+ */
+Blockly.tree.BaseNode.EMPTY_CHILDREN_ = [];
```
rachel-fenichel: Not used anywhere else.

core/components/tree/basenode.js
```diff
@@ -0,0 +1,1151 @@
...
+  if (this.selected_ == selected) {
+    return;
+  }
+  this.selected_ = selected;
+
+  this.updateRow();
+
+  var el = this.getElement();
+  if (el) {
+    Blockly.utils.aria.setState(el, 'selected', selected);
```
rachel-fenichel: I haven't looked at the utils.aria stuff.  Do you also need to unset selected and activedescendant on the previously selected element.

core/components/tree/basenode.js
```diff
@@ -0,0 +1,1151 @@
...
+ * Creates HTML Element for the node.
+ * @return {!Element} html element
+ * @protected
+ */
+Blockly.tree.BaseNode.prototype.toDom = function() {
+  var childClass = this.config_.cssChildrenNoLines;
+
+  var nonEmptyAndExpanded = this.getExpanded() && this.hasChildren();
+
+  var children = document.createElement('div');
```
rachel-fenichel: This (and the previous file) are introducing some new uses of `document`.  Is that going to be an issue when running headless/on node?

core/components/tree/basenode.js
```diff
@@ -0,0 +1,1151 @@
...
+Blockly.tree.BaseNode.prototype.updateIcon_ = function() {
+  this.getIconElement().className = this.getCalculatedIconClass();
+};
+
+/**
+ * Handles mouse down event.
+ * @param {!Event} e The browser event.
+ * @protected
+ */
+Blockly.tree.BaseNode.prototype.onMouseDown = function(e) {
```
rachel-fenichel: It looks like if we want to we'll be able to wire this into gesture, instead of having click handling be completely separate for the toolbox.  Could be interesting to explore.

core/components/tree/treecontrol.js
```diff
@@ -0,0 +1,445 @@
...
+  // which is a generic component
+};
+
+
+/**
+ * Handles focus on the tree.
+ * @param {!Event} e The browser event.
+ * @private
+ */
+Blockly.tree.TreeControl.prototype.handleFocus_ = function(/* e */) {
```
rachel-fenichel: You can use `_e` to have an unused variable without commenting it out (closure compiler may complain otherwise)

core/components/tree/treecontrol.js
```diff
@@ -0,0 +1,445 @@
...
+};
+
+/** @override */
+Blockly.tree.TreeControl.prototype.updateExpandIcon = function() {
+  // no expand icon
+};
+
+/** @override */
+Blockly.tree.TreeControl.prototype.getRowClassName = function() {
+  // TODO: look into this
```
rachel-fenichel: Remove TODO or file an issue.

core/components/tree/treecontrol.js
```diff
@@ -0,0 +1,445 @@
...
+/**
+ * Add roles and states.
+ * @protected
+ * @override
+ */
+Blockly.tree.TreeControl.prototype.initAccessibility = function() {
+  Blockly.tree.TreeControl.superClass_.initAccessibility.call(this);
+
+  var elt = this.getElement();
+  // goog.asserts.assert(elt, 'The DOM element for the tree cannot be null.');
```
rachel-fenichel: Delete.

core/components/tree/basenode.js
rachel-fenichel: And I see below that you're using `bindEventWithChecks`.  Cool.

core/components/tree/treenode.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+  return null;
+};
+
+/**
+ * Returns the source for the icon.
+ * @return {string} Src for the icon.
+ * @override
+ */
+Blockly.tree.TreeNode.prototype.getCalculatedIconClass = function() {
```
rachel-fenichel: Is this the same as the one for the tree control?

samelhusseini: Nope, different CSS used in each case.

core/components/tree/basenode.js
samelhusseini: There's no need for that, I don't know why closure does that, will remove.

core/components/tree/basenode.js
samelhusseini: Yep, when a new item is selected, the currently selected item gets a setSelected(false) before the new one gets setSelected(true). 
Login in Blockly.tree.TreeControl.prototype.setSelectedItem

core/components/component.js
```diff
@@ -0,0 +1,816 @@
...
+ */
+Blockly.Component.prototype.getElementByClass = function(className) {
+  return this.element_ ? document.getElementByClass(className, this.element_) :
+                         null;
+};
+
+/**
+ * Similar to `getElementByClass` except that it expects the
+ * element to be present in the dom thus returning a required value. Otherwise,
+ * will assert.
```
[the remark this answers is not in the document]
samelhusseini: I don't think there's a need for all of this, removing.

core/components/tree/treecontrol.js
samelhusseini: Cool, thanks.

core/components/component.js
```diff
@@ -0,0 +1,816 @@
...
+
+/**
+ * Returns an array of all the elements in this component's DOM with the
+ * provided className.
+ * @param {string} className The name of the class to look for.
+ * @return {!IArrayLike<!Element>} The items found with the class name provided.
+ */
+Blockly.Component.prototype.getElementsByClass = function(className) {
+  return this.element_ ?
+      document.getElementsByClass(className, this.element_) :
```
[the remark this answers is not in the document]
samelhusseini: this.element_ would be null in that case, but I think this is just a nice helper method. No reason you couldn't getElement and run this elsewhere. Since we don't use it, removing..

core/components/tree/basenode.js
samelhusseini: Component.getChildAt sets it's return type as Blockly.Component. This is just to please the closure compiler so it doesn't throw warnings.

samelhusseini:
I ran the closure compiler on this section of code, and I got it down to 0 errors and warnings for this!

core/components/component.js
```diff
@@ -0,0 +1,816 @@
...
+  /**
+   * The DOM element for the component.
+   * @private {?Element}
+   */
+  this.element_ = null;
+
+  /**
+   * Parent component to which events will be propagated.  This property is
+   * strictly private and must not be accessed directly outside of this class!
+   * @private {Blockly.Component?}
```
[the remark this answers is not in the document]
samelhusseini: https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System says the dot before the < in Array is optional, but happy to convert these to be consistent with the rest of Blockly.
Apparently ? at the end is the same, but it's deprecated. Good to know. I'll move them all to the front.

core/components/component.js
```diff
@@ -0,0 +1,815 @@
...
+   *
+   * @private {?Object}
+   */
+  this.childIndex_ = null;
+};
+
+
+/**
+ * Generator for unique IDs.
+ * @type {Blockly.utils.IdGenerator}
```
rachel-fenichel: Non-nullable?

core/components/component.js
```diff
@@ -0,0 +1,815 @@
...
+ * Finally, this method also throws an error if the new child already has a
+ * different parent, or the given index is out of bounds.
+ *
+ * @see Blockly.Component#addChild
+ * @param {Blockly.Component} child The new child component.
+ * @param {number} index 0-based index at which the new child component is to be
+ *    added; must be between 0 and the current child count (inclusive).
+ * @param {boolean=} opt_render If true, the child component will be rendered
+ *    into the parent.
+ * @return {void} Nada.
```
rachel-fenichel: Why have an `@return` annotation at all, then?

core/components/component.js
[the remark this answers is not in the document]
rachel-fenichel: Correct, this is just about consistency.

core/components/component.js
samelhusseini: Yep

core/components/component.js
samelhusseini: Fixed.

core/components/tree/basenode.js
samelhusseini: It shouldn't. We don't initialize a toolbox in headless mode, and all these methods are dom methods.

PR #2819, opened 2019-08-13
No closure toolbox

samelhusseini:
Use built in Blockly.tree control instead of closures.

This PR depends on the toolbox foundations here: https://github.com/google/blockly/pull/2818
It also depends on aria utils: https://github.com/google/blockly/pull/2813

Here, we make use of the new closure-like built-in Blockly tree component to build up the Blockly toolbox.

When all PRs are checked in, the toolbox no longer has any closure dependencies.

Testing: 
- [x] Toolbox in LTR/RTL.
- [x] Toolbox in top / bottom.
- [x] Navigation in all modes.
- [x] Checked DOM matches exactly
- [x] Flat toolbox and nested toolbox
- [x] Touch navigation / Mobile
- [x] Typings passing
- [x] Tests passing
- [x] Keyboard nav accessibility works
- [ ] Please add more to this list of things to test...


core/toolbox.js
```diff
@@ -470,23 +533,23 @@ Blockly.Toolbox.prototype.updateColourFromTheme = function() {
...
  * @private
  */
 Blockly.Toolbox.prototype.updateSelectedItemColour_ = function(tree) {
-  var selectedItem = tree.selectedItem_;
+  var selectedItem = tree.getSelectedItem();
   if (selectedItem) {
     var hexColour = selectedItem.hexColour || '#57e';
     selectedItem.getRowElement().style.backgroundColor = hexColour;
-    tree.toolbox_.addColour_(selectedItem);
+    tree.getToolbox().addColour_(selectedItem);
```
rachel-fenichel: Is `tree.getToolbox()` the same as `this`?

core/toolbox.js
```diff
@@ -470,23 +533,23 @@ Blockly.Toolbox.prototype.updateColourFromTheme = function() {
...
   }
 };
 
 
 /**
  * Recursively add colours to this toolbox.
- * @param {Blockly.Toolbox.TreeNode=} opt_tree Starting point of tree.
+ * @param {Blockly.tree.BaseNode=} opt_tree Starting point of tree.
```
rachel-fenichel: This is a `BaseNode` but on line 505 you have a `TreeNode` for the exact same description of the parameter.  What's the difference, and why is this one not a `TreeNode`?

samelhusseini: TreeControl and TreeNode both inherit from BaseNode. For methods that begin at the root node (TreeControl) and recurse through their children calling the same method (a common theme here), they should use BaseNode as either subclass could be the input.
I wonder why closure didn't complain about the other one, but yes the other one should also be updated to BaseNode.

core/toolbox.js
samelhusseini: Yeah haha I wonder why we were doing tree.toolbox_

Issue #2821, opened 2019-08-13
Block + children bounding boxes aren't always right


rachel-fenichel:
**Describe the bug**

This bug is with new rendering.

Sometimes the bounding boxes aren't quite right, which mostly shows up when putting blocks into inline inputs.  That probably also means that they're wrong for computing the bounds of the workspace.

![image](https://user-images.githubusercontent.com/13686399/62984727-23f81000-bde9-11e9-9537-990063ff166f.png)
![image](https://user-images.githubusercontent.com/13686399/62984754-41c57500-bde9-11e9-8e5f-77dfd12858f3.png)

**Expected behavior**

Inline inputs should tightly bound blocks

**Additional information**
The first step to debugging will be to add debug rendering rectangles to show the bounds of the block including children.

PR #2825, opened 2019-08-14
Correctly calculate widths of blocks with children on external inputs.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #2821 

### Proposed Changes

In render_draw_debug:
- Adds rendering rectangles for the width of a row with connected blocks and the width of a block with connected blocks
- Adds a config object so it's easy to turn on and off parts of the debug rendering--there's too much going on otherwise.
- Uses random colours for some parts, to distinguish between blocks at boundaries

In the other three files:
- Improves calculation of widthWithConnectedBlocks on rows
- Defines widthWithConnectedBlocks on rows
- When adding alignment padding to a row, records the size change in widthWithConnectedBlocks (only for external value inputs)

### Reason for Changes

Block width was wrong before, and the missing amount was the amount that had been added in alignment padding on the given row (so it was showing up for external value inputs but not otherwise).

### Test Coverage
Tested on the playground with the new debug rendering.

### Documentation
Description of debug rendering, at some point.

### Additional Information

SO MANY COLORED RECTANGLES

### DIFF
--- a/core/renderers/block_rendering_rewrite/block_render_draw_debug.js
+++ b/core/renderers/block_rendering_rewrite/block_render_draw_debug.js
@@ -47,7 +47,21 @@ Blockly.blockRendering.Debug = function() {
    * @type {!SVGElement}
    */
   this.svgRoot_ = null;
+
+  /**
+   * @type {Object} Configuration object containing booleans to enable and
+   *     disable debug rendering of specific rendering components.
+   */
+  this.config_ = {
+    rowSpacers: true,
+    elemSpacers: true,
+    rows: true,
+    elems: true,
+    connections: true,
+    blockBounds: true
+  };
 };
+
 
 /**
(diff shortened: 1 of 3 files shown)


core/renderers/block_rendering_rewrite/block_render_draw_debug.js
```diff
@@ -239,9 +311,12 @@ Blockly.blockRendering.Debug.prototype.drawBoundingBox = function(width, height,
  * @package
  */
 Blockly.blockRendering.Debug.prototype.drawDebug = function(block, info) {
+  this.config_.rowSpacers = false;
   this.clearElems();
   this.svgRoot_ = block.getSvgRoot();
 
+  this.randomColour_ = '#' + Math.floor(Math.random() * 16777215).toString(16);
```
samelhusseini: love it

PR #2830, opened 2019-08-14
Minor refactor in drawing statement and value inputs

moniika:
… and more approriate constants.

## The basics

- [x] I branched from develop

### Proposed Changes

block_render_draw:
* Updated drawValueInput and drawStatementInput to use the element xPos
* Updated constants used in drawStatementInput to more accurately reflect what is happening.
block_render_draw_highlight:
* Using xPos for drawing highlight for value input and statement input
* Using svgPaths function and appending to this.steps_ rather than creating a local variable to follow patterns used in other functions in file.

### Reason for Changes

The change in constants more accurately represent what is happening in computation.
Using xPos and the highlight minor refactor more closely matches pattern used in other functions and does not change output.

### Test Coverage

Tested on playground in RTL and LTR.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/block_rendering_rewrite/block_render_draw.js
+++ b/core/renderers/block_rendering_rewrite/block_render_draw.js
@@ -191,7 +191,7 @@ Blockly.blockRendering.Drawer.prototype.drawValueInput_ = function(row) {
   this.positionExternalValueConnection_(row);
 
   this.outlinePath_ +=
-      Blockly.utils.svgPaths.lineOnAxis('H', row.xPos + row.width) +
+      Blockly.utils.svgPaths.lineOnAxis('H', input.xPos + input.width) +
       input.connectionShape.pathDown +
       Blockly.utils.svgPaths.lineOnAxis('v', row.height - input.connectionHeight);
 };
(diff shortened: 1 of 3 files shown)


core/renderers/block_rendering_rewrite/block_render_draw_highlight.js
```diff
@@ -108,40 +108,35 @@ Blockly.blockRendering.Highlighter.prototype.drawJaggedEdge_ = function(row) {
...
-  var steps = '';
   if (this.RTL_) {
     var belowTabHeight = row.height - input.connectionHeight;
 
-    steps =
-        Blockly.utils.svgPaths.moveTo(
-            row.xPos + row.width - this.highlightOffset_, row.yPos) +
-        this.puzzleTabPaths_.pathDown(this.RTL_) +
-        Blockly.utils.svgPaths.lineOnAxis('v', belowTabHeight);
+    this.steps_.push(Blockly.utils.svgPaths.moveTo(
```
rachel-fenichel: > Using svgPaths function and appending to this.steps_ rather than creating a local variable to follow patterns used in other functions in file.

I'm going to move us away from the steps array in this file soon.  Similar to https://github.com/google/blockly/pull/2817

core/renderers/block_rendering_rewrite/block_render_draw.js
```diff
@@ -203,15 +203,18 @@ Blockly.blockRendering.Drawer.prototype.drawStatementInput_ = function(row) {
   var x = input.xPos + input.width;
 
   var innerTopLeftCorner =
-      Blockly.blockRendering.constants.NOTCH.pathRight + ' h -' +
-      (Blockly.blockRendering.constants.NOTCH_WIDTH -
-          Blockly.blockRendering.constants.CORNER_RADIUS) +
+      input.notchShape.pathRight +
+      Blockly.utils.svgPaths.lineOnAxis('h',
+          -(Blockly.blockRendering.constants.NOTCH_OFFSET_LEFT -
```
rachel-fenichel: Can you get this notch_offset_left value to be stored on the measurable instead?

rachel-fenichel: Could also probably store a pointer to INSIDE_CORNERS, since I think that's used in the measurable already for sizing information.

moniika: I like the change of storing notch_offset_left on the measurable. I was thinking of storing it on the notchShape object. Since with that change I should also update the other instances of notch_offset_left, I'd prefer to do that in a follow up PR.
As for inside_corners, I don't think it should be stored on a measurable since i don't anticipate different measurables having different corner shapes from each other.

core/renderers/block_rendering_rewrite/block_render_draw_highlight.js
moniika: Got it, I will undo the changes that refactor to push onto the array.

Issue #2734, opened 2019-07-29
Remove remaining closure ui components


RoboErikG:
**Describe the bug**

We're almost done removing our dependency on the Closure UI library. We still need to cleanup any remaining references to goog.* other than the requires/provides pieces used by the compiler.

samelhusseini:
Done.

PR #2831, opened 2019-08-15
Blockly menus without closure

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2734

### Proposed Changes

Introduces a Blockly Menu and MenuItem that don't rely on the closure library.
Both components inherit from Blockly.Component.

- Blockly.MenuItem is mostly in charge of rendering.
- Blockly.Menu is a container of MenuItem's and registers all events, passing some for the child menuitem to handle.

A bunch of highlighting code in Blockly.Menu is borrowed from goog.ui.Container, but other than that, this is mostly new code.

Tested that all APIs surfaced with @package are sufficient for Blockly's use in the context menu and field_dropdown. 

Ran the closure compiler with 0 errors and warnings.

### Reason for Changes

Remove closure dependency.


core/components/menu/menu.js
```diff
@@ -0,0 +1,543 @@
...
+/**
+ * A basic menu class.
+ * @constructor
+ * @extends {Blockly.Component}
+ */
+Blockly.Menu = function() {
+  Blockly.Component.call(this);
+
+  /**
+   * This is the element that we will listen to the real focus events on.
```
rachel-fenichel: Add an explanation of what -1 means.

core/components/menu/menu.js
```diff
@@ -0,0 +1,543 @@
...
+Blockly.Menu.prototype.createDom = function() {
+  var element = document.createElement('div');
+  element.setAttribute('id', this.getId());
+  this.setElementInternal(element);
+
+  // Set class
+  element.setAttribute('class', 'goog-menu goog-menu-vertical');
+  element.setAttribute('style', 'user-select: none;');
+
+  element.tabIndex = "0";
```
rachel-fenichel: Why is this a string?

core/components/menu/menu.js
```diff
@@ -0,0 +1,543 @@
...
+
+  // Lazily create the child element ID map on first use.
+  if (!this.childElementIdMap_) {
+    this.childElementIdMap_ = {};
+  }
+  this.childElementIdMap_[id] = child;
+};
+
+/**
+ * Returns the child menu item that owns the given DOM node, or null if no such
```
rachel-fenichel: Is there a difference between a "menu item" and a "control"?

core/components/menu/menu.js
```diff
@@ -0,0 +1,543 @@
...
+    }
+  }
+  return null;
+};
+
+/** @override */
+Blockly.Menu.prototype.addChildAt = function(control, index, opt_render) {
+
+  // Disable mouse event handling by child controls.
+  // control.setHandleMouseEvents(false);
```
rachel-fenichel: ?

core/components/menu/menu.js
```diff
@@ -0,0 +1,543 @@
...
+
+/**
+ * Updates the highlighted index when children are added or moved.
+ * @param {number} fromIndex Index of the child before it was moved, or -1 if
+ *     the child was added.
+ * @param {number} toIndex Index of the child after it was moved or added.
+ * @private
+ */
+Blockly.Menu.prototype.updateHighlightedIndex_ = function(
+    fromIndex, toIndex) {
```
rachel-fenichel: I don't understand this function.  Will get an explanation in person.

core/components/menu/menu.js
samelhusseini: shouldn't be. Fixed.

core/components/menu/menu.js
samelhusseini: Removing.

core/components/menu/menu.js
samelhusseini: No, I'll update all references to control with menu item, since that's the only use case here.

core/components/menu/menu.js
samelhusseini: Done.

core/components/menu/menu.js
samelhusseini: This was to support dynamic adding of menu item elements. Since we don't need that built in, I'll remove for now.

samelhusseini: I should clarify, that this is while the menu item is open. I know we have a dynamic dropdown in the playground, but that the menu there is created every time the dropdown is open.

samelhusseini:
This is ready for a final look :)

Issue #1346, opened 2017-09-27
Align first row of fields/inputs for connected blocks


vicng:
This is better explained with screenshots.

### Example 1
Problem:
![screen shot 2017-09-27 at 2 16 30 pm](https://user-images.githubusercontent.com/13488389/30938325-8019ff0a-a38e-11e7-943f-97ea75d01293.png)

Desired behaviour:
![screen shot 2017-09-27 at 2 14 37 pm](https://user-images.githubusercontent.com/13488389/30938301-6921e600-a38e-11e7-81a2-a7f715c9e890.png)

### Example 2

Problem:
![screen shot 2017-09-27 at 2 41 44 pm](https://user-images.githubusercontent.com/13488389/30939364-090a7f26-a392-11e7-8ae1-4cf4ebcdbad6.png)

Desired behaviour (note how the "do" label is also aligned with the first row):
![screen shot 2017-09-27 at 2 40 56 pm](https://user-images.githubusercontent.com/13488389/30939325-e9851b5c-a391-11e7-9fc2-fb2b1ebbbaf6.png)

### Example 3

Problem:
![screen shot 2017-09-27 at 2 47 09 pm](https://user-images.githubusercontent.com/13488389/30939580-c23185bc-a392-11e7-8b94-8695a1d299a4.png)

Desired behaviour (note how alignment still occurs for the bottom row with its children's first rows):
![screen shot 2017-09-27 at 2 46 50 pm](https://user-images.githubusercontent.com/13488389/30939571-b82ae40a-a392-11e7-9796-3230e4a75190.png)

rachel-fenichel:
Fixed by #2835

Issue #333, opened 2016-04-14
Horizontal baseline alignment across text, fields, inline values, and external values.


AnmAtAnm:
Right now (picture attached), Blockly does not attempt to calculate a baseline for all text in an input (or adjacent inputs), even though an input is often designed to be read across multiple fields.  Inline fields are the worst offender, often leading to text the appears to curve down.

Blockly would appear much more polished if it calculated a baseline for all text in an input, and adjacent inputs, and aligned the textual components to that.

<img width="835" alt="screen shot 2016-04-13 at 5 41 03 pm" src="https://cloud.githubusercontent.com/assets/9916202/14513956/1e776f02-01a1-11e6-8c9c-7b84f3809609.png">

I offer bonus style points for aligning the jigsaw connectors to the same baseline.

rachel-fenichel:
#2835 significantly improves this.  @samelhusseini has an upcoming change to center text inside fields better, and @rachel-fenichel has an upcoming change to center the jigsaw connectors vertically.

Note that most changes here are vertical centering within the row, not baseline alignment.

MNSDivya:
![image](https://user-images.githubusercontent.com/56336353/81166256-0db26580-8fb1-11ea-84c0-157a44dbdb46.png)
Issue still persists

PR #2835, opened 2019-08-15
Center some fields and inputs vertically

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/1346
https://github.com/google/blockly/issues/333

### Proposed Changes

Center fields and inputs vertically within a row.

### Reason for Changes

Alignment

### Test Coverage
Checked in the playground.
New:
image

Old:

image


rachel-fenichel:
It's not perfect ("get" is better aligned in the old than in the new, probably because of the vertical alignment of text within a dropdown field next to it), but still an improvement.

Also, notice how little code has to change to make it so!

rachel-fenichel:
Possible followup: vertically center the input and output connections as well.  If that's what we want.

And we need to decide if any of these are flags but I'm leaning toward nope.  It's hard to argue for the old version on any grounds other than familiarity.

PR #2836, opened 2019-08-15
Colour field accessibility

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Colour field accessibility.

### Proposed Changes

Add keyboard navigation support using the arrow keys to move around the colour field. 
Focus the color field table when it's shown.
Add aria attributes for: 
- expanded
- rowcount
- colcount
- selected (for gridcells)

Update css so it looks like this:

Screen Shot 2019-08-15 at 11 09 41 AM

### Reason for Changes

Accessibility!

### Test Coverage

Tested on:
* Desktop Chrome
* Desktop Firefox
* Desktop Safari
* IE 11
* Edge 18

### Additional Information

Up: Moves up one grid cell until it reaches the top
Down: Moves down one grid cell until it reaches the top
Left: Moves left one grid cell until it reaches the left edge, then it wraps around the end of previous row if it exists.
Right: Moves right one grid cell until it reaches the right edge, then it moves to the start of following row if it exists
Enter: Selects the highlighted colour.

Passes closure compiler with no errors and no (new) warnings.


core/field_colour.js
```diff
@@ -314,26 +453,51 @@ Blockly.FieldColour.prototype.dropdownCreate_ = function() {
   // Create the palette.
   var table = document.createElement('table');
   table.className = 'blocklyColourTable';
+  table.tabIndex = "0";
```
alschmiedt: nit: Should be ' '

core/field_colour.js
```diff
@@ -314,26 +453,51 @@ Blockly.FieldColour.prototype.dropdownCreate_ = function() {
...
+  Blockly.utils.aria.setRole(table,
```
alschmiedt: Could we maybe add a label to the table? I think currently it just starts to list off all of the values in the cells, which might be confusing.

core/field_colour.js
samelhusseini: actually should be a number haha, but thanks for spotting this.

core/field_colour.js
samelhusseini: We should definitely add a label. Since that requires the addition of new localized messages, and since all fields are going to need a similar change. I'll leave that for a separate PR.

Issue #2838, opened 2019-08-15
Stop filtering subdirectories in the compile test script


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**
The compile test script (`scripts/compile.sh`) filters out subdirectories because it was designed to only include one rendering directory.

This doesn't interact well with the new pile of subdirectories in core.

It is also unnecessary now that we've decided to choose the renderer at runtime instead of at compile time.

rachel-fenichel:
And also in build.py, actually.

alschmiedt:
Fixed in #2842

PR #2840, opened 2019-08-15
Move rows to a new file and annotate

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

- Move the three Row types (Row, TopRow, BottomRow) to a new file: measurables/rows.js
- Update the build.py to find the files
  - I filed #2838 to discuss doing this properly, in both build.py and tests/compile
- Add lots of annotations
- Add annotations in other places to make this file's annotations work
- Add requires and provides as needed
- Rebuild

### Reason for Changes

Documenting this code is part of getting it ready for release.  Also, measurables.js was too crowded.

### Test Coverage

Checked in the playground.

### Additional Information

@RoboErikG I'm ready to discuss design and APIs around this file, and it seems like a good size chunk to discuss.
I've also put a few TODOs in that I want to resolve before merge, but want to discuss first.


core/renderers/measurables/rows.js
```diff
@@ -0,0 +1,389 @@
...
+   * @type {Blockly.RenderedConnection}
+   */
+  this.connection = block.previousConnection;
+
+  var precedesStatement = block.inputList.length &&
+      block.inputList[0].type == Blockly.NEXT_STATEMENT;
+
+  // This is the minimum height for the row. If one of its elements has a
+  // greater height it will be overwritten in the compute pass.
+  if (precedesStatement && !block.isCollapsed()) {
```
rachel-fenichel: TODO (fenichel): See if I can move the minimum height and maximum width stuff into measure, for both top and bottom rows.

core/renderers/measurables/rows.js
```diff
@@ -0,0 +1,389 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview An object representing a single row on a rendered block and all
+ * of its subcomponents.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
+
+goog.provide('Blockly.blockRendering.Rows');
```
moniika: I believe this is supposed to match Blockly.blockRendering.Row.

core/renderers/block_rendering_rewrite/measurables.js
```diff
@@ -419,78 +436,14 @@ Blockly.blockRendering.RoundCorner = function() {
...
-
-Blockly.blockRendering.Row.prototype.getFirstSpacer = function() {
-  return this.elements[0];
-};
-
-Blockly.blockRendering.Row.prototype.getLastSpacer = function() {
-  return this.elements[this.elements.length - 1];
-};
-
 Blockly.blockRendering.BetweenRowSpacer = function(height, width) {
```
moniika: I prefer the between row spacer to inherit from Row, because of its usage as an element in the rows array in block_render_info. There's also a few instances in code where it is referred to as a row, like in makeSpacerRow_ (the name implies it is a row and the jsdoc refers to it as a row).

core/renderers/measurables/rows.js
```diff
@@ -0,0 +1,389 @@
...
+
+/**
+ * Inspect all subcomponents and populate all size properties on the row.
+ * @package
+ */
+Blockly.blockRendering.Row.prototype.measure = function() {
+  var connectedBlockWidths = 0;
+  for (var e = 0; e < this.elements.length; e++) {
+    var elem = this.elements[e];
+    this.width += elem.width;
```
moniika: This is a preexisting issue, but since this.width is not reset, I think the width is incorrect if the method is ever accidentally called twice. Potentially we could introduce a minWidth property that this.width is reset to in the beginning.

core/renderers/block_rendering_rewrite/measurables.js
rachel-fenichel: Agreed.  I just did some poking and my current proposal is:
- Row is the parent class
- There are four subclasses:
  - TopRow
  - BottomRow
  - SpacerRow
  - InputRow
- No one ever instantiates a Row directly (always a subclass)
- SpacerRow contains an array of elements, which contains one InRowSpacer.  This gets rid of the need to check if a row is a spacer row before iterating across its element array.

This lets me make many of the annotations tighter (albeit with some casts) and removes some unnecessary distinctions between spacer and non spacer rows.

core/renderers/measurables/rows.js
```diff
@@ -0,0 +1,389 @@
...
+
+/**
+ * @override
+ */
+Blockly.blockRendering.TopRow.prototype.measure = function() {
+  for (var e = 0; e < this.elements.length; e++) {
+    var elem = this.elements[e];
+    this.width += elem.width;
+    if (!(elem.isSpacer())) {
+      if (elem.type == 'hat') {
```
moniika: Similar idea for bottom row. You could call super.measure() and then just loop through the elements to get the hat if isHat is true and adjust the startY and height. Since the number of elements in this row is expected to be very small, looping through the elements a second time shouldn't be a performance concern.
An idea to go with this is that topRow could have an additional field that is a reference to the hat element so you don't need to loop though.

A related question, why is startY for the row also set in createTopRow? should that be removed?

core/renderers/block_rendering_rewrite/block_render_draw.js
```diff
@@ -296,14 +302,12 @@ Blockly.blockRendering.Drawer.prototype.drawLeft_ = function() {
...
     var row = this.info_.rows[r];
-    if (!(row.isSpacer())) {
-      for (var e = 0; e < row.elements.length; e++) {
-        var elem = row.elements[e];
-        if (elem.isInlineInput()) {
-          this.drawInlineInput_(elem);
-        } else if (elem.isIcon() || elem.isField()) {
-          this.layoutField_(elem);
-        }
+    for (var e = 0; e < row.elements.length; e++) {
```
rachel-fenichel: Removed isSpacer check because spacer rows now contain an elements array.

core/renderers/block_rendering_rewrite/block_render_info.js
```diff
@@ -718,14 +723,12 @@ Blockly.blockRendering.RenderInfo.prototype.finalize_ = function() {
...
     }
-    if (!(row.isSpacer())) {
-      var xCursor = row.xPos;
-      for (var e = 0; e < row.elements.length; e++) {
-        var elem = row.elements[e];
-        elem.xPos = xCursor;
-        elem.centerline = this.getElemCenterline_(row, elem);
-        xCursor += elem.width;
-      }
+    var xCursor = row.xPos;
```
rachel-fenichel: Removed spacer check because spacer rows now contain an elements array.

core/renderers/measurables/rows.js
rachel-fenichel: `startY` for the row isn't set in `createTopRow` (now `populateTopRow`).  That sets `startY` on the block instead.

The row doesn't know about the hat until this measure pass, because the hat is added by pushing directly to the elements array.  But I could definitely change that.

I don't think that I can just figure out the correct height in a later loop, because if the hat was the tallest thing in the block it would override the height of anything else (with the Math.max call) and we wouldn't know what the second-tallest thing was to reset the height to.

core/renderers/measurables/rows.js
```diff
@@ -0,0 +1,441 @@
...
+  }
+  return null;
+};
+
+/**
+ * @override
+ */
+Blockly.blockRendering.BottomRow.prototype.measure = function() {
+  this.width = 0;
+  this.height = 0;
```
moniika: There is some logic for Bottom row for the minimum height of the row. Doesn't the reset of width and height need to take that into account?

rachel-fenichel: Good catch.  Fixed.

PR #2845, opened 2019-08-16
Fixing basic warnings in block_render_draw_debug.

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

Updates annotation for svgRoot to allow for null.
Minor refactor of for loop to declare variable inside the loop block.

### Reason for Changes

Cleans up warnings, code health.

### Test Coverage

Checked things still work in playground.

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/renderers/block_rendering_rewrite/block_render_draw_debug.js
+++ b/core/renderers/block_rendering_rewrite/block_render_draw_debug.js
@@ -44,7 +44,7 @@ Blockly.blockRendering.Debug = function() {
   /**
    * The SVG root of the block that is being rendered.  Debug elements will
    * be attached to this root.
-   * @type {!SVGElement}
+   * @type {SVGElement}
    */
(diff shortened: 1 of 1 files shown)


core/renderers/block_rendering_rewrite/block_render_draw_debug.js
```diff
@@ -44,7 +44,7 @@ Blockly.blockRendering.Debug = function() {
   /**
    * The SVG root of the block that is being rendered.  Debug elements will
    * be attached to this root.
-   * @type {!SVGElement}
```
moniika: This change I'm not 100% certain is the right change to make.
Alternatively, I could also set this.svgRoot_ to undefined. The variable is never expected to be set to null so that annotation makes sense. But I'm not sure about setting something to undefined because I don't see it often.

PR #2853, opened 2019-08-17
Field angle keyboard navigation

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Treating the angle picker like a slider when using the arrow keys:

UP: increments 15
DOWN: decrements 15
LEFT: decrements 15 (increments in RTL)
RIGHT: increments 15 (decrements in RTL)

### Reason for Changes

Accessibility


samelhusseini:
@chrisgarrity what are your thoughts (from an RTL perspective) on treating the angle picker as a slider and flipping the left and right arrow keys only in RTL.

core/field_angle.js
```diff
@@ -333,6 +341,38 @@ Blockly.FieldAngle.prototype.updateGraph_ = function() {
...
+ */
+Blockly.FieldAngle.prototype.onHtmlInputKeyDown_ = function(e) {
+  Blockly.FieldAngle.superClass_.onHtmlInputKeyDown_.call(this, e);
+
+  var multiplier;
+  if (e.keyCode === Blockly.utils.KeyCodes.LEFT) {
+    // decrement (increment in RTL)
+    multiplier = this.sourceBlock_.RTL ? 1 : -1;
+  } else if (e.keyCode === Blockly.utils.KeyCodes.RIGHT ||
+      e.keyCode === Blockly.utils.KeyCodes.UP) {
```
alschmiedt: I don't think we want Blockly.utils.KeyCodes.UP here

core/field_angle.js
```diff
@@ -333,6 +341,38 @@ Blockly.FieldAngle.prototype.updateGraph_ = function() {
...
+    multiplier = this.sourceBlock_.RTL ? -1 : 1;
+  } else if (e.keyCode === Blockly.utils.KeyCodes.DOWN) {
+    // decrement
+    multiplier = -1;
+  } else if (e.keyCode === Blockly.utils.KeyCodes.UP) {
+    // increment
+    multiplier = 1;
+  }
+  if (multiplier) {
+    this.setAngle(Number(this.getValue()) +
```
alschmiedt: This is passing in -15 when wrapping instead of 345. I think we need to handle that either here or in setAngle

core/field_angle.js
samelhusseini: Yep, good spot!

core/field_angle.js
samelhusseini: Yep, i'll add wrapping for negative values.

PR #2865, opened 2019-08-19
Cleanup of last fortnight’s commits.

NeilFraser:
No functional changes.  Ran junit and mocha tests.

PR #2867, opened 2019-08-19
Fix reference error in block_render_draw

samelhusseini:
Fix ``Uncaught ReferenceError: e is not defined`` introduced by:
https://github.com/google/blockly/pull/2865


NeilFraser:
Sorry, that's the result of me trying to resolve conflicts using the web interface.

Issue #2722, opened 2019-07-27
Standardize Field Configuration


BeksOmega:
**Is your feature request related to a problem? Please describe.**

@RoboErikG  mentioned [here ](https://github.com/google/blockly/pull/2654#discussion_r307908212)that he would like to move to a new system for configuring field customizations.

I think this is a good idea, but I have some questions so I thought I'd move it to an issue.

**Describe the solution you'd like**

> I think I'd rather start using an opt_config parameter on the constructor with a map of the configuration options. The keys should be the same as the JSON config which will also let you use the same code for both.

**Describe alternatives you've considered**

@RoboErikG 

**Open questions**
Should this change only to apply to future fields? Or should the currently available field constructors change to support this?

Should the opt_config accept tooltips and other parameters that apply to all fields?

Should there be a hook so you can change the options at runtime (e.g. setOptions(config))?

Should the currently available hooks (e.g. setSpellcheck) be removed, deprecated, or left?

RoboErikG:
What makes the most sense to me (and @rachel-fenichel may overrule me):

- Use this pattern for future fields that have non-trivial configuration. We can discuss what counts as non-trivial on a field by field basis.
- For existing fields that have 2+ config parameters adding an opt_config to the constructor makes sense, but is low priority.
- Existing configuration methods should be left. Slightly fewer lines of code in block inits doesn't really justify the churn of asking developers to update.
- I'm slightly against adding setOptions. Configuring fields outside of init adds a lot of complexity. Do you have a good use case in mind?

rachel-fenichel:
+1 to @RoboErikG

BeksOmega:
Ok I have a revised proposal! Some things are added and some things are changed, so tell me what you think of it.

Important: I have highlighted things that I think need discussion with italics.

### Proposal: Field Configuration Paradigm

Field configuration is not consistent. There are four places where field configuration is possible: JSON definition, JavaScript constructor, JavaScript constants, and JavaScript hooks.

#### Definitions and Constructors

The JSON definition and the JavaScript constructor should be equivalent in terms of function. Anything you can configure with one you should be able to configure with the other. They should both cover all configurations exposed by constants and hooks.

#### Constants

JavaScript constants modify all field instances of a given type and cannot be changed while Blockly is running. They do not need to cover all configurations available in the definition or constructor. They should only be available when there is a logical default value for a configuration, and it would be reasonable for an outside developer to set this property globally.

[...]

PR #2876, opened 2019-08-20
Changed Checkbox Field to New Configuration Paradigm

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

* Adds an opt_config property to the constructor that accepts checkCharacter
* JSON Definition now also accepts checkCharacter
* Added a setCheckCharacter function to the checkmark field.

### Reason for Changes

### Test Coverage

Added tests for:
* Setting the character via the constant.
* Setting the character via the JS Constructor.
* Setting the character via the JSON Definition.
* Setting the character via setCheckCharacter.

Tested on:
* Desktop Chrome

### Documentation

Yes, the [checkmark customization](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox#checkmark_character) section should be updated with new information.

### Additional Information

N/A

### DIFF
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -162,7 +162,8 @@ Blockly.FieldCheckbox.prototype.initView = function() {
 
 /**
  * Set the character used for the check mark.
- * @param {string} character The character to use for the check mark.
+ * @param {?string} character The character to use for the check mark, or
+ *    null to use the default.
  */
 Blockly.FieldCheckbox.prototype.setCheckCharacter = function(character) {
   this.checkChar_ = character;
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -162,12 +162,12 @@ Blockly.FieldCheckbox.prototype.initView = function() {
 
 /**
  * Set the character used for the check mark.
- * @param {string} character The character to use for the checkmark.
+ * @param {string} character The character to use for the check mark.
  */
 Blockly.FieldCheckbox.prototype.setCheckCharacter = function(character) {
   this.checkChar_ = character;
   if (this.textContent_) {
-    this.textContent_.nodeValue = character;
+    this.textContent_.nodeValue = character || Blockly.FieldCheckbox.CHECK_CHAR;
   }
 };
 
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -49,6 +49,15 @@ goog.require('Blockly.utils.Size');
  * @constructor
  */
 Blockly.FieldCheckbox = function(opt_value, opt_validator, opt_config) {
+  opt_value = this.doClassValidation_(opt_value);
+  if (opt_value === null) {
+    opt_value = 'FALSE';
+  }
+  Blockly.FieldCheckbox.superClass_.constructor.call(
+      this, opt_value, opt_validator, opt_config);
+
+  this.size_.width = Blockly.FieldCheckbox.WIDTH;
+
   /**
    * Character for the check mark. Used to apply a different check mark
    * character to individual fields.
@@ -57,19 +66,7 @@ Blockly.FieldCheckbox = function(opt_value, opt_validator, opt_config) {
    */
   this.checkChar_ = null;
 
-  if (opt_config) {
-    if (opt_config['checkCharacter']) {
-      this.checkChar_ = opt_config['checkCharacter'];
-    }
-  }
-
-  opt_value = this.doClassValidation_(opt_value);
-  if (opt_value === null) {
-    opt_value = 'FALSE';
-  }
-  Blockly.FieldCheckbox.superClass_.constructor.call(
-      this, opt_value, opt_validator, opt_config);
-  this.size_.width = Blockly.FieldCheckbox.WIDTH;
+  this.configure_(opt_config);
 };
 goog.inherits(Blockly.FieldCheckbox, Blockly.Field);
 
@@ -133,6 +130,19 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
  * @private
  */
 Blockly.FieldCheckbox.prototype.isDirty_ = false;
+
+/**
+ * Configure the field based on the given map of options.
+ * @param {Object} opt_config A map of options to configure the field based on.
+ * @private
+ */
+Blockly.FieldCheckbox.prototype.configure_ = function(opt_config) {
+  if (opt_config) {
+    if (opt_config['checkCharacter']) {
+      this.checkChar_ = opt_config['checkCharacter'];
+    }
+  }
+};
 
 /**
  * Create the block UI for this checkbox.
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -43,8 +43,8 @@ goog.require('Blockly.utils.Size');
  *    returns a validated value ('TRUE' or 'FALSE'), or null to abort the
  *    change.
  * @param {Object=} opt_config A map of options used to configure the field.
- *    See the documentation for a list of properties this parameter supports.
- *    https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox#creation
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox#creation}
+ *    for a list of properties this parameter supports.
  * @extends {Blockly.Field}
  * @constructor
  */
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -49,6 +49,9 @@ goog.require('Blockly.utils.Size');
  * @constructor
  */
 Blockly.FieldCheckbox = function(opt_value, opt_validator, opt_config) {
+  /** @type {?string} @private */
+  Blockly.FieldCheckbox.prototype.checkChar_ = null;
+
   if (opt_config) {
     if (opt_config['checkCharacter']) {
       this.checkChar_ = opt_config['checkCharacter'];
@@ -127,14 +130,6 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
 Blockly.FieldCheckbox.prototype.isDirty_ = false;
 
 /**
- * Character for the check mark. Used to apply a different check mark
- * character to individual fields.
- * @type {?string}
- * @private
- */
-Blockly.FieldCheckbox.prototype.checkChar_ = null;
-
-/**
  * Create the block UI for this checkbox.
  * @package
  */
@@ -156,7 +151,9 @@ Blockly.FieldCheckbox.prototype.initView = function() {
  */
 Blockly.FieldCheckbox.prototype.setCheckCharacter = function(character) {
   this.checkChar_ = character;
-  this.textContent_.nodeValue = character;
+  if (this.textContent_) {
+    this.textContent_.nodeValue = character;
+  }
 };
 
 /**
--- a/core/field_checkbox.js
+++ b/core/field_checkbox.js
@@ -42,15 +42,25 @@ goog.require('Blockly.utils.Size');
  *    changes to the field's value. Takes in a value ('TRUE' or 'FALSE') &
  *    returns a validated value ('TRUE' or 'FALSE'), or null to abort the
  *    change.
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the documentation for a list of properties this parameter supports.
+ *    https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox#creation
  * @extends {Blockly.Field}
  * @constructor
  */
-Blockly.FieldCheckbox = function(opt_value, opt_validator) {
+Blockly.FieldCheckbox = function(opt_value, opt_validator, opt_config) {
+  if (opt_config) {
+    if (opt_config['checkCharacter']) {
+      this.checkChar_ = opt_config['checkCharacter'];
+    }
+  }
+
   opt_value = this.doClassValidation_(opt_value);
   if (opt_value === null) {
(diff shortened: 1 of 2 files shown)


core/field_checkbox.js
```diff
@@ -127,11 +145,20 @@ Blockly.FieldCheckbox.prototype.initView = function() {
...
-  var textNode = document.createTextNode(Blockly.FieldCheckbox.CHECK_CHAR);
-  this.textElement_.appendChild(textNode);
+  this.textContent_.nodeValue =
+      this.checkChar_ || Blockly.FieldCheckbox.CHECK_CHAR;
   this.textElement_.style.display = this.value_ ? 'block' : 'none';
 };
 
+/**
+ * Set the character used for the check mark.
+ * @param {string} character The character to use for the checkmark.
```
samelhusseini: @package for method visibility.

core/field_checkbox.js
```diff
@@ -116,6 +126,14 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
  */
 Blockly.FieldCheckbox.prototype.isDirty_ = false;
 
+/**
+ * Character for the check mark. Used to apply a different check mark
+ * character to individual fields.
+ * @type {?string}
+ * @private
+ */
+Blockly.FieldCheckbox.prototype.checkChar_ = null;
```
samelhusseini: Moving forward we are no longer setting fields on the prototype as per the Google JS style guide: https://google.github.io/styleguide/jsguide.html#features-classes-fields
Let's move this to the constructor.

core/field_checkbox.js
```diff
@@ -127,11 +145,20 @@ Blockly.FieldCheckbox.prototype.initView = function() {
...
   this.textElement_.style.display = this.value_ ? 'block' : 'none';
 };
 
+/**
+ * Set the character used for the check mark.
+ * @param {string} character The character to use for the checkmark.
+ */
+Blockly.FieldCheckbox.prototype.setCheckCharacter = function(character) {
+  this.checkChar_ = character;
+  this.textContent_.nodeValue = character;
```
samelhusseini: Only if the field is initialized will ``this.textContent_`` have a value, do we need to check for that here? is it possible to call this before ``init``.

BeksOmega: Oh gosh golly, thanks for the catch! Fixed and added a test.

core/field_checkbox.js
BeksOmega: Ok I moved it but I'm not sure if I did it correctly. Questions:

* Initialize to undefined? or null?
* @private first? or @type first?
* I removed the property description, is that correct?

core/field_checkbox.js
BeksOmega: It's a developer hook, so I think it should be public. Should it be tagged @public?

public: Definitely.

core/field_checkbox.js
```diff
@@ -42,15 +42,28 @@ goog.require('Blockly.utils.Size');
...
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the documentation for a list of properties this parameter supports.
+ *    https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox#creation
  * @extends {Blockly.Field}
  * @constructor
  */
-Blockly.FieldCheckbox = function(opt_value, opt_validator) {
+Blockly.FieldCheckbox = function(opt_value, opt_validator, opt_config) {
+  /** @type {?string} @private */
+  Blockly.FieldCheckbox.prototype.checkChar_ = null;
```
samelhusseini: Let's move this to the bottom of the constructor (especially after calling the superclass constructor).
and defined the type with multi line comments like so: 
```

/**
 * Character for the check mark. Used to apply a different check mark
 * character to individual fields.
 * @type {?string}
 * @private
 */
```

core/field_checkbox.js
samelhusseini: Oh yes, so leave as is.

core/field_checkbox.js
samelhusseini: - Initialize to null.
- Type then private. 
- Keep property description, use multiline jsdoc rather than single line jsdoc format.

core/field_checkbox.js
samelhusseini: Also, when it's in the constructor. It's simply: 
``this.checkChar_ = null;``

BeksOmega: > Also, when it's in the constructor. It's simply:
> `this.checkChar_ = null;`

Derp. Fixed.

> Let's move this to the bottom of the constructor

I don't think it can be moved to the bottom b/c then it would override the configured option.

Option 1:
We could move it to the bottom and then call a config function after it. Problem is some fields (like the number field) need to be configured before the super-constructor is called so that the class validator works properly.

Option 2:
Remove the intial `this.checkChar_ = null; ` line and instead document the `this.checkChar_ = opt_config['checkCharacter']` line.

Option 3:
Leave as is.

core/field_checkbox.js
BeksOmega: Done!

core/field_checkbox.js
samelhusseini: My recommendation would be to do: 
- super
- initialization
- config

I recognize that in some cases (number field), you might have to do initialization first, and that's okay as an exception to the rule.

BeksOmega: > My recommendation would be to do:
> 
> * super
> * initialization
> * config

Ok, can I fix as a separate PR since this affects all fields?

core/field_checkbox.js
```diff
@@ -42,15 +42,33 @@ goog.require('Blockly.utils.Size');
...
+  /**
+   * Character for the check mark. Used to apply a different check mark
+   * character to individual fields.
+   * @type {?string}
+   * @private
+   */
+  this.checkChar_ = null;
+
+  if (opt_config) {
+    if (opt_config['checkCharacter']) {
```
samelhusseini: Just so we're consistent, lets move this to a separate configure method like the other PR, and I think this is good to go. We can fix the ordering of the constructor in a separate PR.

core/field_checkbox.js
samelhusseini: Yep

core/field_checkbox.js
```diff
@@ -116,6 +131,19 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
...
 
+/**
+ * Configure the field based on the given map of options.
+ * @param {Object} opt_config A map of options to configure the field based on.
+ * @private
+ */
+Blockly.FieldCheckbox.prototype.configure_ = function(opt_config) {
+  if (opt_config) {
+    if (opt_config['checkCharacter']) {
+      this.checkChar_ = opt_config['checkCharacter'];
```
samelhusseini: Okay, one last thing. Can we call setCheckCharacter here so it all leads to the same code path. This method becomes a property bag to call a bunch of configuration methods. ie: 
```js
new FieldCheckbox(val, null, {
   checkCharacter: '\u2713'
});
```
is shorthand for having done: 
```js
var field = new FieldCheckbox(val);
field.setCheckCharacter('\u2713');
```

core/field_checkbox.js
```diff
@@ -127,11 +155,22 @@ Blockly.FieldCheckbox.prototype.initView = function() {
...
 };
 
+/**
+ * Set the character used for the check mark.
+ * @param {string} character The character to use for the checkmark.
+ */
+Blockly.FieldCheckbox.prototype.setCheckCharacter = function(character) {
+  this.checkChar_ = character;
+  if (this.textContent_) {
+    this.textContent_.nodeValue = character;
```
samelhusseini: If you pass this null, it sets the nodeValue to null. but the desired outcome is to set it to CHECK_CHAR just like the above so I think you need ``character || Blockly.FieldCheckbox.CHECK_CHAR`` here

core/field_checkbox.js
```diff
@@ -116,6 +131,19 @@ Blockly.FieldCheckbox.prototype.CURSOR = 'default';
  */
 Blockly.FieldCheckbox.prototype.isDirty_ = false;
 
+/**
+ * Configure the field based on the given map of options.
+ * @param {Object} opt_config A map of options to configure the field based on.
+ * @private
+ */
+Blockly.FieldCheckbox.prototype.configure_ = function(opt_config) {
+  if (opt_config) {
```
samelhusseini: As I think about this more. Perhaps the configure method should be public. 
This allows a user to either pass the property bag in the constructor, or call it at any point after.

samelhusseini:
@BeksOmega let's focus on getting this PR in first, and then we can move on to the rest of them. This is the most straightforward so we can focus on the desired architecture here.

My general recommendation would be to:
- keep constructor arguments as is (to not break backwards compat). I think it's more important that we keep the existing constructors than break them to fit into this new pattern. As you see in the Image field PR, it gets messy very quickly trying to change constructor arguments
- add an opt_config to each constructor argument
- every constructor calls configure(opt_config) with the config object
- configure is a public method on the field that can be called independently with a property bag
- configure goes through each config option and calls the respective setXYZ option. 
- a number of public setXYZ options for each configuration option.

RoboErikG:
And to expand on Sam's first point, there's a cost to code changes in general and especially for API changes. Both in merge conflicts for developers that modify Blockly and risk of introducing new bugs. It's one of the reasons we'd avoiding breaking public APIs for ~8 years. ;)

We did some breaking changes in the last release and there's a few more coming this release for the rendering updates in order to fix the biggest pain points. We'd still like to minimize additional changes if they aren't causing issues to keep the delta as small as possible for developers.

In this particular case, we'd like to adopt an opt_config map as the preferred pattern going forward, and fields that don't have a bunch of config parameters in the constructor can have an opt_config added easily. But for complex constructors like field_image we don't think it's worth the cost of changing the API to match the new pattern since so many developers would break.

BeksOmega:
Sorry, I'm a little bit confused I thought I got [approval](https://github.com/google/blockly/issues/2722#issuecomment-519300741) to do things [this way](https://github.com/google/blockly/issues/2722#issuecomment-519294538).

Did something change since then?

If you would like to rediscuss that's totally cool. I'll take the PR's down and we can chat over on #2722 .

I had just worked ahead because I was having some trouble with #2789, and I wanted to make sure I got all the time-sensative stuff in before school starts.

RoboErikG:
Yeah, that's my fault, sorry. I think I didn't realize how significantly some of the field constructors would need to change to match the new pattern until I saw it in code. To do the type checking on the parameters for image and number would add a decent chunk of code as well to handle the backwards compatibility so let's hold off on those two.

I think it's still okay to move most of the fields over, but we'll take a closer look at all of them to double check.

Sorry for the confusion Beka and thanks for working through all of these.

BeksOmega:
> * configure is a public method on the field that can be called independently with a property bag
> * configure goes through each config option and calls the respective setXYZ option.
> * a number of public setXYZ options for each configuration option.

This is a good idea, but it was decided [here](https://github.com/google/blockly/issues/2722#issuecomment-516052316) and [here](https://github.com/google/blockly/issues/2722#issuecomment-516481707) not to go with that method because some configuration options are hard to change at runtime.

For example: It was decided that a `setWrap` function would not be added to the angle field because that would change the value of the field, which was deamed too complex.

> * every constructor calls configure_(opt_config) with the config object

I did have a reason for not doing this originally, but I can see where it keeps the constructor more organized, so I will move the fields over to this pattern.

<details>
  <summary>Original reasoning</summary>
<p>

As you can see [here](https://github.com/google/blockly/issues/2722#issuecomment-519294538) the original plan was:
* Add a doConfiguration_ function to the base field.
* doConfiguration_ would then be called by the base field in its constructor.
* Subclasses could override then doConfiguration_ method, and then call the parent function so that config inheritance was achieved.

This did not end up working because number fields need to be configured before the super-constructor is called.

Since that option was off the table I decided the best course of action would be to include configuration directly in the constructor.

This is because adding identically named `configure_` functions to each field makes it appear like `configure_` is a protected function, rather than a private one.

Which encourages future developers to call the "base" `configure_` method when they should not, because doing so would cause code to run multiple times when it does not need to.

This is not a defense of my reasoning, simply an explanation of it. As I said above I will move the fields over to the new organization.
</p>
</details>

samelhusseini:
Thanks for the added context @BeksOmega. 

I wanted to clarify one thing, in everything I've seen in these PRs so far the set of options configured with ``opt_config`` is the same as the ones surfaced with setters (eg: ``setXYZ``, but based on my understanding of the proposal, that's not always true. ie: because opt_config configures things on initialization, then it can configure options that we won't necessarily surface through setters.

If the above is true, then having ``configure_`` be only a function of initialization (ie: an extension of the constructor) and each of the ``setXYZ`` methods be public methods for configuration options that can be changed during runtime (ie: at any point during the Blockly lifecycle) sounds good to me.

core/field_checkbox.js
samelhusseini: Last unresolved comment, and then we're good to go!!

BeksOmega: Fixed and added test.

BeksOmega:
Hehe some PRs just take more work than others.

I wanted to tell you, I really do appreciate your feedback! I love it when I get comments on my PRs cuz it lets me know someone actually read it, which is like "yay someone appreciates what I did!" and also like "yay maybe it won't break everything now!".

And you're really good at doing PR reviews! You're always really nice about things, and you're super thorough.

So thank you for dealing with _me_ on this one @samelhusseini :P

samelhusseini:
Thanks @BeksOmega, I appreciate that. I also very much appreciate all the work you've been putting in to getting these changes honed out. Thank you!

PR #2878, opened 2019-08-20
Changed Image Field to New Configuration Paradigm.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

* Changed JS Constructor to look like: function(src, width, height, opt_onClick, opt_config) where the opt_config accepts opt_alt and opt_flipRTL properties.
* Also fixed the block factory's image field JS (JSON was already correct).

### Reason for Changes

### Test Coverage

Added tests for:
* Backwards compatibility with the old constructor signature.
* Setting the click listener via JS Constructor.
* Setting the click listener via setOnClick function.
* Setting the flipRtl property via JS Constructor.
* Setting the flipRtl property via JSON Definition.

Also moved the setAlt tests into the new customization suite.

Tested on:
* Desktop Chrome

### Documentation

Yes, the [image creation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#creation), and [on click](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#click_handler) sections should be updated with new information.

### Additional Information

N/A

### DIFF
--- a/core/field_image.js
+++ b/core/field_image.js
@@ -147,7 +147,7 @@ Blockly.FieldImage.prototype.configure_ = function(
     config = opt_config;
   }
   if (doOldParams) {
-    var config = {};
+    config = {};
     // opt_onClick used to be opt_alt.
     config['alt'] = opt_onClick;
     // opt_config used to be opt_onClick.
--- a/core/field_image.js
+++ b/core/field_image.js
@@ -47,16 +47,17 @@ goog.require('Blockly.utils.Size');
  * @constructor
  */
 Blockly.FieldImage = function(src, width, height, opt_onClick, opt_config) {
-  this.sourceBlock_ = null;
-
   if (!src) {
     throw Error('Src value of an image field is required');
   }
+  width = Blockly.utils.replaceMessageReferences(width);
+  height = Blockly.utils.replaceMessageReferences(height);
   if (isNaN(height) || isNaN(width)) {
     throw Error('Height and width values of an image field must cast to' +
       ' numbers.');
   }
 
+  src = Blockly.utils.replaceMessageReferences(src);
   this.setValue(src);
 
   // Ensure height and width are numbers.  Strings are bad at math.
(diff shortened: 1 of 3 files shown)


core/field_image.js
```diff
@@ -124,6 +139,13 @@ Blockly.FieldImage.prototype.EDITABLE = false;
  */
 Blockly.FieldImage.prototype.isDirty_ = false;
 
+/**
+ *
+ * @type {boolean}
+ * @private
+ */
+Blockly.FieldImage.prototype.flipRtl_ = false;
```
samelhusseini: Move to the constructor, and add description.

core/field_image.js
```diff
@@ -90,13 +110,8 @@ goog.inherits(Blockly.FieldImage, Blockly.Field);
  * @nocollapse
  */
 Blockly.FieldImage.fromJson = function(options) {
-  var src = Blockly.utils.replaceMessageReferences(options['src']);
```
samelhusseini: How come we no longer need to pass these through ``replaceMessageReferences``?

core/field_image.js
```diff
@@ -70,13 +71,32 @@ Blockly.FieldImage = function(src, width, height,
   this.size_ = new Blockly.utils.Size(imageWidth,
       imageHeight + Blockly.FieldImage.Y_PADDING);
 
-  this.flipRtl_ = opt_flipRtl;
-  this.text_ = opt_alt || '';
-  this.setValue(src || '');
-
+  // Deal with config options and param reordering.
```
samelhusseini: Since all this functionality is related, maybe we can move it to its own private method. something like ``configure_(options)``

core/field_image.js
BeksOmega: Fixed, probably. See [comment on checkbox pr](https://github.com/google/blockly/pull/2876#discussion_r316258252). (I put this one more towards the bottom of the constructor).

core/field_image.js
BeksOmega: Oh I removed this because I thought I added it accidentally in a previous PR, looking at the git blame that is not the case.

Readded.

core/field_image.js
BeksOmega: I changed it, but it feels like the backwards compatibility code makes less sense when it's extracted from its context.

core/field_image.js
```diff
@@ -124,6 +122,47 @@ Blockly.FieldImage.prototype.EDITABLE = false;
...
+  if (typeof opt_onClick == 'function') {
+    doOldParams = false;
+    this.clickHandler_ = opt_onClick;
+  }
+  if (opt_config && typeof opt_config == 'object') {
+    doOldParams = false;
+    config = opt_config;
+  }
+  if (doOldParams) {
+    var config = {};
```
samelhusseini: redeclaring variable config.

core/field_image.js
```diff
@@ -124,6 +122,47 @@ Blockly.FieldImage.prototype.EDITABLE = false;
  */
 Blockly.FieldImage.prototype.isDirty_ = false;
 
+/**
+ * Configure this field's customization options. Taking into account old
+ * parameter ordering on the constructor.
+ * @param {Function|string|null} opt_onClick Option function to be called
+ *    when the image is clicked. Or alt text.
+ * @param {Object|Function|null} opt_config A map of options used to
```
samelhusseini: are these all optional at this point? if not, rename opt_

BeksOmega: Sort of optional? You can pass null just fine. I think it's confusing because this function is handling backwards compatibility for a different function.

Option 1:
Move the backwards compat code back to the constructor. Handle only configuration inside config_.

Option 2:
Come up with a different naming scheme for this particular config_ function. Currently I do not have any ideas.

Option 3:
Leave as is.

core/field_image.js
BeksOmega: Fixed.

core/field_image.js
samelhusseini: Why not keep the constructor parameters as is, and just provide a way to override them in opt_config. 
I'm wondering why the decision was made to move rtl and alt into config, but everything else is still there?
What makes rtl a configuration option more than say something like ``onClick``?

BeksOmega:
Closing this as it was decided [here](https://github.com/google/blockly/pull/2876#issuecomment-523713777) to leave the image field as is for the time being.

I think adding an opt_config parameter to the image field to support tooltips (and possibly future generic config options) can be discussed separately (possibly on #2722).

PR #2880, opened 2019-08-20
Storing notch_left_offset on measurable or row.

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

Storing notch_left_offset on measurable or row instead of using constant directly

### Reason for Changes

This allows for easier future customization.

### Test Coverage

Tested locally in playground.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/measurables/base.js
+++ b/core/renderers/measurables/base.js
@@ -63,6 +63,14 @@ Blockly.blockRendering.Measurable.prototype.connectionShape =
 Blockly.blockRendering.Measurable.prototype.notchShape =
     Blockly.blockRendering.constants.NOTCH;
 
+/**
+ * The offset from the left side of a block or the inside of a statement input
+ * to the left side of the connection notch.
+ * @type {number}
+ */
+Blockly.blockRendering.Measurable.prototype.notchOffset =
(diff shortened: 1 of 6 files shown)


core/renderers/measurables/base.js
```diff
@@ -63,6 +63,14 @@ Blockly.blockRendering.Measurable.prototype.connectionShape =
 Blockly.blockRendering.Measurable.prototype.notchShape =
     Blockly.blockRendering.constants.NOTCH;
 
+/**
+ * The offset from the left side of a block or the inside of a statement input
+ * to the left side of the connection notch.
+ * @type {number}
+ */
+Blockly.blockRendering.Measurable.prototype.notchOffset =
```
rachel-fenichel: Can we define this on just the Measurables that will need it?  That is, previous and next connections, and statement inputs.

moniika: I think notchOffset should be defined on all measurables that have a notchShape defined. So, I'd like to keep it there, but I wouldn't be against moving it along with notchShape if that refactor happens.

rachel-fenichel: +1

PR #2881, opened 2019-08-20
Change Label Fields to New Configuration Pardigm

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

* Added an opt_config property to the JS Constructor, which accepts a 'class' property.
* Added a setClass function.

### Reason for Changes

### Test Coverage

Added tests for:
* Backwards compatibility with the old constructor signature.
* Setting the class via JS Constructor.
* Setting the class via JSON Definition.
* Setting the class via setClass().
* Removing the class via setClass().

Tested on:
* Desktop Chrome

### Documentation

Yes, the [label creation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation), and [label serializable creation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label-serializable#creation) sections should both be updated.

Both pages should also be given a customization section that describes the setClass method.

### Additional Information

N/A

### DIFF
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -45,13 +45,7 @@ goog.require('Blockly.utils.Size');
  * @constructor
  */
 Blockly.FieldLabel = function(opt_value, opt_config) {
-  /**
-   * The css class name to apply to the field's textElement_.
-   * @type {?string}
-   * @private
-   */
-  this.class_ = null;
-
+  // Construction.
   opt_value = this.doClassValidation_(opt_value);
   if (opt_value === null) {
     opt_value = '';
@@ -59,9 +53,24 @@ Blockly.FieldLabel = function(opt_value, opt_config) {
   Blockly.FieldLabel.superClass_.constructor.call(
       this, opt_value, null, opt_config);
 
+  // Initialization.
+  /**
+   * The css class name to apply to the field's textElement_.
+   * @type {?string}
+   * @private
+   */
+  this.class_ = null;
+
+  /**
+   * The size of the area rendered by the field.
+   * @type {Blockly.utils.Size}
+   * @private
+   */
+  this.size_ = new Blockly.utils.Size(0, Blockly.Field.TEXT_DEFAULT_HEIGHT);
+
+  // Configuration.
   this.configure_(opt_config);
 
-  this.size_ = new Blockly.utils.Size(0, Blockly.Field.TEXT_DEFAULT_HEIGHT);
 };
 goog.inherits(Blockly.FieldLabel, Blockly.Field);
 
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -64,7 +64,7 @@ Blockly.FieldLabel = function(opt_value, opt_config) {
   /**
    * The size of the area rendered by the field.
    * @type {Blockly.utils.Size}
-   * @private
+   * @protected
    */
   this.size_ = new Blockly.utils.Size(0, Blockly.Field.TEXT_DEFAULT_HEIGHT);
 
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -54,22 +54,14 @@ Blockly.FieldLabel = function(opt_value, opt_config) {
    */
   this.class_ = null;
 
-  // Handle backwards-compat.
-  if (opt_config) {
-    if (typeof opt_config == 'string') {
-      // opt_config used to be opt_class.
-      this.class_ = opt_config;
-    } else if (opt_config['class']) {
-      this.class_ = opt_config['class'];
-    }
-  }
-
   opt_value = this.doClassValidation_(opt_value);
   if (opt_value === null) {
     opt_value = '';
   }
   Blockly.FieldLabel.superClass_.constructor.call(
       this, opt_value, null, opt_config);
+
+  this.configure_(opt_config);
 };
 goog.inherits(Blockly.FieldLabel, Blockly.Field);
 
@@ -93,6 +85,23 @@ Blockly.FieldLabel.fromJson = function(options) {
  * @const
  */
 Blockly.FieldLabel.prototype.EDITABLE = false;
+
+/**
+ * Configure the field based on the given map of options.
+ * @param {Object} opt_config A map of options to configure the field based on.
+ * @private
+ */
+Blockly.FieldLabel.prototype.configure_ = function(opt_config) {
+  // Handle backwards-compat.
+  if (opt_config) {
+    if (typeof opt_config == 'string') {
+      // opt_config used to be opt_class.
+      this.class_ = opt_config;
+    } else if (opt_config['class']) {
+      this.class_ = opt_config['class'];
+    }
+  }
+};
 
 /**
  * Create block UI for this label.
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -39,8 +39,8 @@ goog.require('Blockly.utils.Size');
  * @param {string=} opt_value The initial value of the field. Should cast to a
  *    string. Defaults to an empty string if null or undefined.
  * @param {Object=} opt_config A map of options used to configure the field.
- *    See the documentation for a list of properties this parameter supports.
- *    https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation}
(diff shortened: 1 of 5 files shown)


core/field_label.js
```diff
@@ -38,32 +38,45 @@ goog.require('Blockly.utils.Size');
...
-  this.setValue(opt_value);
+  Blockly.FieldLabel.superClass_.constructor.call(
+      this, opt_value, null, opt_config);
 };
 goog.inherits(Blockly.FieldLabel, Blockly.Field);
 
 /**
  * Construct a FieldLabel from a JSON arg object,
- * dereferencing any string table references.
+ * de-referencing any string table references.
```
samelhusseini: I believe dereferencing is one word.

core/field_label.js
```diff
@@ -74,6 +87,13 @@ Blockly.FieldLabel.fromJson = function(options) {
  */
 Blockly.FieldLabel.prototype.EDITABLE = false;
 
+/**
+ * The css class name to apply to the field's textElement_.
+ * @type {?string}
+ * @private
+ */
+Blockly.FieldLabel.prototype.class_ = null;
```
samelhusseini: Same as other PRs, please move to the constructor.

core/field_label.js
```diff
@@ -100,4 +120,16 @@ Blockly.FieldLabel.prototype.doClassValidation_ = function(opt_newValue) {
   return String(opt_newValue);
 };
 
+/**
+ * Set the css class applied to the field's textElement_.
+ * @param {?string} cssClass The new css class name, or null to remove.
+ */
+Blockly.FieldLabel.prototype.setClass = function(cssClass) {
+  Blockly.utils.dom.removeClass(this.textElement_, this.class_);
```
samelhusseini: ``removeClass`` expects an non-null element. this.textElement_ can be null if this field isn't initialized.

core/field_label.js
BeksOmega: Done.

core/field_label.js
BeksOmega: Fixed, I think. See [comment on checkbox pr](https://github.com/google/blockly/pull/2876#discussion_r316258252).

core/field_label.js
BeksOmega: Fixed and added test.

core/field_label_serializable.js
```diff
@@ -36,14 +36,16 @@ goog.require('Blockly.utils');
  * Class for a non-editable, serializable text field.
  * @param {*} opt_value The initial value of the field. Should cast to a
  *    string. Defaults to an empty string if null or undefined.
- * @param {string=} opt_class Optional CSS class for the field's text.
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the documentation for a list of properties this parameter supports.
+ *    https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label-serializable#creation
```
samelhusseini: If you use {@link http....} it will add a link when this is used to generate documentation.

BeksOmega: Fixed. I will fix for all other fields separately.

core/field_label.js
```diff
@@ -38,18 +38,30 @@ goog.require('Blockly.utils.Size');
...
+   * @type {?string}
+   * @private
+   */
+  this.class_ = null;
+
   opt_value = this.doClassValidation_(opt_value);
   if (opt_value === null) {
     opt_value = '';
   }
-  this.setValue(opt_value);
```
samelhusseini: I think this was removed by mistake.

core/field_label.js
```diff
@@ -38,18 +38,30 @@ goog.require('Blockly.utils.Size');
...
+  Blockly.FieldLabel.superClass_.constructor.call(
+      this, opt_value, null, opt_config);
```
samelhusseini: this.size_ is going to be overridden by the parent constructor with this call.
What's the reason for ``doClassValidation_`` going before the parent constructor? If none, then we can easily move this to the top, so that configuration inheritance is achieved.

core/field_label.js
```diff
@@ -38,18 +38,30 @@ goog.require('Blockly.utils.Size');
...
  *    string. Defaults to an empty string if null or undefined.
- * @param {string=} opt_class Optional CSS class for the field's text.
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation}
+ *    for a list of properties this parameter supports.
  * @extends {Blockly.Field}
  * @constructor
  */
-Blockly.FieldLabel = function(opt_value, opt_class) {
+Blockly.FieldLabel = function(opt_value, opt_config) {
```
samelhusseini: I'm not a fan of this. I know the goal here is to move away from setting configuration options in the constructor, and I'm totally okay with that plan moving forward with new fields. Unfortunately due to the legacy nature of these constructors, we don't want to break existing developers.
I recognize we can do that by doing a check on the constructor parameters but for the same reasons that were outlined in FieldImage, the code gets messy very quickly and hard to parse. Here are our options, and I'll see what others think: 
1. Append opt_config to existing fields, for some configuration options like ``class`` above, it might mean there are multiple ways to configure the same option (constructor or property bag), and code like this:
```
new Blockly.FieldLabel(val, 'test');
new Blockly.FieldLabel(val, null, { class: 'test'});
```
Not great for developers using the new config, as they have to skip parameters, but clean in code.
2. Do what you did here, which is check for the value of the parameter, and switch constructor paths based on that.
Not clean in code, compiler types are inconsistent (because now you're saying opt_config can be a string or an Object but you only handle that internally, and don't surface it in jsdoc to developers (docs) and the closure compiler. 
3. We could make handle both constructors, but in a way the compiler understands:
```
/**
 * Class for a non-editable, non-serializable text field.
 * @param {string=} opt_value The initial value of the field. Should cast to a
 *    string. Defaults to an empty string if null or undefined.
 * @param {string|Object=} opt_classOrConfig Either an optional CSS class, or a
 * map of options used to configure the field.
 * @extends {Blockly.Field}
 * @constructor
 */
Blockly.FieldLabel = function (opt_value, opt_classOrConfig) {
  if (typeof opt_classOrConfig == 'string') {
    return Blockly.FieldLabel.constructWithClass_(opt_value,
      /** @type {string} */ opt_classOrConfig);
  } else {
    return Blockly.FieldLabel.constructWithConfig_(opt_value,
      /** @type {Object} */ opt_classOrConfig)
  }
};

/**
 * Class for a non-editable, non-serializable text field.
 * @param {string=} opt_value The initial value of the field. Should cast to a
 *    string. Defaults to an empty string if null or undefined.
 * @constructor
 */
Blockly.FieldLabel.constructWithClass_ = function (opt_value, opt_class) {
  ...
};

/**
 * Class for a non-editable, non-serializable text field.
 * @param {string=} opt_value The initial value of the field. Should cast to a
 *    string. Defaults to an empty string if null or undefined.
 * @param {Object=} opt_config A map of options used to configure the field.
 *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation}
 *    for a list of properties this parameter supports.
 * @constructor
 */
Blockly.FieldLabel.constructWithConfig_ = function (opt_value, opt_config) {
  ...
};
```
Notice how the constructor is the one branching out the two constructor paths. It's also type casting the parameter, and each constructor has the correct type description.

@rachel-fenichel @RoboErikG what are your thoughts on dealing with this backwards compat issue with all these fields?

core/field_label.js
BeksOmega: Class validation must be handled before passing the value to the constructor. Otherwise if the developer passes an invalid value to the constructor or JSON the `value_` ends up being null.

This is the course of action recommended in the [docs](https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating).

Moved this.size_ set to after the super-constructor call.

samelhusseini: Sounds good. Thanks for moving that. Let's try: 
- super class (as close as possible to the top)
- initializations (this.class, this.size)
- finally the configure call.

BeksOmega: Fixed!

BeksOmega:

BeksOmega:

PR #2887, opened 2019-08-20
Moves the cursor to correct location when block is deleted

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
When the cursor is on a block that is deleted the cursor will also be deleted.

### Proposed Changes
Move the cursor to the appropriate location before the block is deleted. 
The appropriate location are one of the below:
* The connection point of the deleted block and the parent 
* In the case of no parent or the parent has also been deleted to the workspace

### Test Coverage

Tested on:

### DIFF
--- a/core/keyboard_nav/navigation.js
+++ b/core/keyboard_nav/navigation.js
@@ -718,6 +718,24 @@ Blockly.navigation.moveCursorOnBlockDelete = function(deletedBlock) {
 };
 
 /**
+ * When a block that the cursor is on is mutated move the cursor to the block
+ * level.
+ * @param {!Blockly.Block} mutatedBlock The block that is being mutated.
+ * @package
+ */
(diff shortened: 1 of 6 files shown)


core/keyboard_nav/navigation.js
```diff
@@ -662,6 +662,60 @@ Blockly.navigation.handleEnterForWS = function() {
...
+  var cursor = Blockly.navigation.cursor_;
+  if (cursor) {
+    var curNode = cursor.getCurNode();
+    var block = Blockly.navigation.getSourceBlock_(curNode);
+
+    if (block === deletedBlock) {
+      // If the block has a parent move the cursor to their connection point.
+      if (block.getParent()) {
+        var topConnection = block.previousConnection
+          ? block.previousConnection : block.outputConnection;
```
rachel-fenichel: Move the ? to the end of the previous line ("When a line is broken at an operator the break comes after the symbol. ")

PR #2895, opened 2019-08-21
Add local build tasks to gulp

samelhusseini:
##  The basics

- [x] I branched from develop

### Proposed Changes

Adds a build tasks to the gulpfile which make use of the npm closure compiler. It uses whichever platform it has available to it beginning with ``native`` then ``java`` then ``javascript``. All options used here work or any of the platforms. (Tested). 

The build tasks are split in a similar fashion to the build.py file into: 
- blockly (core)
- blocks
- generators (each in a separate task)
- uncompressed. 

The one thing here that isn't available that is in build.py is message parsing and generation of the the locale files.

### Reason for Changes

Ability to build locally without having to acquire the the closure compiler, and just be able to run: 
```
npm install
npm run build
```

### Test Coverage

Tested on:
- Mac

### Documentation

We need to add documentation into how to build, but I think that should be tied with any documentation to deprecate the current build.py. 

We're not there yet (deprecating), so going to keep this as an option for now.

### Additional information

The ``uncompressed`` task still has a dependency on ``python``. This is because I haven't found a non-python replacement to`` calcdeps.py``. Note this actually uses ``depswriter.py`` which is the recommended tool to use as ``calcdeps.py`` has been deprecated.

Given you only need to run uncompressed for development, I think it's okay to keep python as a dependency for that component only.


gulpfile.js
```diff
@@ -36,15 +36,295 @@ var fs = require('fs');
...
+  // Remove these if they belong to Google or MIT.
+  // MIT's permission to do this is logged in Blockly issue #2412.
+  return gulp.replace(new RegExp(licenseRegex, "g"), '');
+}
+
+/**
+ * Helper method for prepending the auto-generated header text.
+ */
+function prependHeader() {
+  return gulp.insert.prepend(`// Do not edit this file; automatically generated by build.py.\n`);
```
alschmiedt: Is the build.py part of this comment still valid?

samelhusseini: Yes, I wanted to keep the two compatible for now, but as I think about it more, maybe we should know the source of truth.

PR #2896, opened 2019-08-21
Split render info

rachel-fenichel:
##  The basics

- [x] I branched from develop

### Proposed Changes

Split render_info into a base class and a few different classes that extend it, for three different renderers that we will publish:
- **geras** is for compatibility with the old renderer.  It keeps everything the same.
- **thrasos** is for the primary renderer moving forward.  It's going to diverge from geras over time.  Right now the primary difference is in vertical centering of fields and icons.  Next I want to get rid of the highlight and shadow and have it use the modern theme.
- **zelos** is for the makecode/scratch style renderer.  Right now it'll just run the base renderer, but this gives us a place to put makecode/scratch style changes.

block_render_info is the base class (and will be renamed eventually).


core/renderers/measurables/rows.js
```diff
@@ -481,3 +481,12 @@ Blockly.blockRendering.InputRow.prototype.measure = function() {
   }
   this.widthWithConnectedBlocks = this.width + connectedBlockWidths;
 };
+
+
+Blockly.blockRendering.InputRow.prototype.getLastSpacer = function() {
+  if (this.hasExternalInput || this.hasStatement) {
+    var spacer = this.elements[this.elements.length - 3];
```
RoboErikG: This seems very fragile.

rachel-fenichel: I made it very very very slightly less fragile.

PR #2897, opened 2019-08-21
Dropdown scrolls to selected element

### LINKED ISSUE #2613: Dropdown with long list does not scroll to selected item when re-opened
### Problem statement

Problem occurred with drop down that have a long list of items (for example images that force the select zone to be scrollable).
A value is selected a first time. And the second time the dropdown is open, the item is correctly selected but the list is not scrolled to show the item.
User must scroll itself to find the selected item.

### Expected Behavior

Expected Behavior; each time the dropdown is opened, the selected item is shown and list scrolled

image

### Actual Behavior

Actual behaviour; when opened a second time, the list does not show the selected item, because list was not scrolled

image

### Steps to Reproduce

1. Create a long list of image.
2. Select an image after scroll to find it
3. Re-open the dropdown to select another image

### Additional Information

Many thanks for the great job around blockly and visual development.

mirandago:
## The basics

- [x] I branched from develop

### Resolves

#2613

### Proposed Changes

Created a private variable to calculate the y-value to scroll the div element.

### Reason for Changes

The div would not scroll to the selected element when the dropdown opened.

### Test Coverage

Tested on:

### Documentation

No documentation needs to be changed.

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -58,6 +58,7 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
     Blockly.FieldDropdown.validateOptions_(menuGenerator);
   }
   this.menuGenerator_ = menuGenerator;
+  this.selectedMenuItem_ = this.getOptions()[0];
 
   this.trimOptions_();
   var firstTuple = this.getOptions()[0];
@@ -179,7 +180,7 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
       /** @type {!Element} */ (this.menu_.getElement()), 'blocklyDropdownMenu');
 
   this.positionMenu_(this.menu_);
-
+  Blockly.utils.style.scrollIntoContainerView(this.selectedMenuItem_.getElement(), this.menu_.getElement());
   // Focusing needs to be handled after the menu is rendered and positioned.
   // Otherwise it will cause a page scroll to get the misplaced menu in
   // view. See issue #1329.
@@ -217,6 +218,7 @@ Blockly.FieldDropdown.prototype.widgetCreate_ = function() {
     menuItem.setChecked(value == this.value_);
     if (value == this.value_) {
       selectedMenuItem = menuItem;
+      this.selectedMenuItem_ = menuItem;
     }
     menuItem.onAction(this.handleMenuActionEvent_, this);
   }


core/field_dropdown.js
```diff
@@ -58,6 +58,7 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
     Blockly.FieldDropdown.validateOptions_(menuGenerator);
   }
   this.menuGenerator_ = menuGenerator;
+  this.selectedMenuItem_ = this.getOptions()[0];
```
samelhusseini: It's possible to not have a selected menu item. I know this is harmless because you're just using it to scroll, but just so it's not confusing for anyone reading the code, set this here to null, and check for this.selectedMenuItem_ before scrolling.

core/field_dropdown.js
```diff
@@ -179,7 +180,7 @@ Blockly.FieldDropdown.prototype.showEditor_ = function() {
       /** @type {!Element} */ (this.menu_.getElement()), 'blocklyDropdownMenu');
 
   this.positionMenu_(this.menu_);
-
+  Blockly.utils.style.scrollIntoContainerView(this.selectedMenuItem_.getElement(), this.menu_.getElement());
```
samelhusseini:
```
if (this.selectedMenuItem_) {
...
```

core/field_dropdown.js
```diff
@@ -217,6 +218,7 @@ Blockly.FieldDropdown.prototype.widgetCreate_ = function() {
     menuItem.setChecked(value == this.value_);
     if (value == this.value_) {
       selectedMenuItem = menuItem;
```
samelhusseini: No need to keep the local selectedMenuItem. You can replace it with this.selectedMenuItem_

core/field_dropdown.js
samelhusseini: Be sure to annotate it to meet the style guide: 
```
  /**
   * A reference to the currently selected menu item.
   * @type {Blockly.MenuItem}
   * @private
   */
  this.selectedMenuItem_ = null;
```

core/field_dropdown.js
samelhusseini: Travis is also failing on eslint. That's because this exceeds the line length, change to this: 
```
  // Scroll the dropdown to show the selected menu item.
  if (this.selectedMenuItem_) {
    Blockly.utils.style.scrollIntoContainerView(
        this.selectedMenuItem_.getElement(), this.menu_.getElement());
  }
```

core/field_dropdown.js
mirandago: This makes sense! Thank you. I was confused on whether that commend should be within the  function or if it should be on the outside below all the Blockly.FieldDropdown.prototypes.

core/field_dropdown.js
mirandago: Done!

core/field_dropdown.js
mirandago: Made this change as well!

core/field_dropdown.js
```diff
@@ -140,6 +140,13 @@ Blockly.FieldDropdown.prototype.imageElement_ = null;
  */
 Blockly.FieldDropdown.prototype.imageJson_ = null;
 
+/**
+ * A reference to the currently selected menu item.
+ * @type {Blockly.MenuItem}
+ * @private
+ */
+this.selectedMenuItem_ = null;
```
samelhusseini: You'll need to set this at the bottom of the constructor. It won't work like that.

samelhusseini:
One last thing and then lgtm.

core/field_dropdown.js
mirandago: Done!

YannCaron:
Many thanks guys. :-)

PR #2901, opened 2019-08-23
Changed text input field to new configuration paradigm

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

Adds an opt_config property to the text input field, which accepts `spellcheck` as an option.

Also changed the setSpellcheck function so that it works while the editor is opened. Not sure how often this actually occures, so I can revert if you want.

### Reason for Changes

Fix possible bug.

### Test Coverage

Added tests for:
* Default spellcheck value.
* JS Constructor.
* JSON Definition.
* setSpellcheck while the editor is hidden.
* setSpellcheck while the editor is open.

Tested on:
* Desktop Chrome

### Documentation

The text input field [creation ](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/text-input#creation)and [customization ](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/text-input#spellcheck)section should be updated.

### Additional Information

N/A

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -117,7 +117,7 @@ Blockly.FieldTextInput.prototype.configure_ = function(opt_config) {
     return;
   }
   if (typeof opt_config['spellcheck'] == 'boolean') {
-    this.setSpellcheck(opt_config['spellcheck']);
+    this.spellcheck_ = opt_config['spellcheck'];
   }
 };
 
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -116,10 +116,7 @@ Blockly.FieldTextInput.prototype.configure_ = function(opt_config) {
   if (!opt_config) {
     return;
   }
-  if (typeof opt_config == 'boolean') {
-    // opt_config used to be spellcheck.
-    this.setSpellcheck(opt_config);
-  } else if (typeof opt_config['spellcheck'] == 'boolean') {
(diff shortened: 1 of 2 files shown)


core/field_textinput.js
```diff
@@ -106,6 +107,20 @@ Blockly.FieldTextInput.prototype.CURSOR = 'text';
...
+ * Configure the field based on the given map of options.
+ * @param {Object} opt_config A map of options to configure the field based on.
+ * @private
+ */
+Blockly.FieldTextInput.prototype.configure_ = function(opt_config) {
+  if (!opt_config) {
+    return;
+  }
+  if (typeof opt_config['spellcheck'] == 'boolean') {
+    this.setSpellcheck(opt_config['spellcheck']);
```
samelhusseini: You're calling setSpellcheck here, but this in inconsistent with what is being done elsewhere (eg: [checkbox field](https://github.com/google/blockly/blob/develop/core/field_checkbox.js#L142)). Why is that?

samelhusseini:
One clarification question, otherwise lgtm.

core/field_textinput.js
BeksOmega: No reason, just a derp on my part. Thanks for the catch!

BeksOmega:

PR #2903, opened 2019-08-23
Constant provider objects

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Cleanup
### Proposed Changes

Make constants and highlightconstants into objects that can be instantiated.

### Reason for Changes

Allowing for an inheritance structure for constants for different renderers.

### Test Coverage
Rebuild and used the playground.


core/renderers/block_rendering_rewrite/block_rendering_constants.js
```diff
@@ -19,125 +19,167 @@
...
-/**
- * Width of SVG path for jagged teeth at the end of collapsed blocks.
- * @const
- */
-Blockly.blockRendering.constants.JAGGED_TEETH_WIDTH = 6;
+  /**
+   * Width of SVG path for jagged teeth at the end of collapsed blocks.
+   * @const
+   */
+  this.JAGGED_TEETH_WIDTH = 6;
```
RoboErikG: As discussed in person, I think everything that's a fixed constant (up to this point) should also be pulled out into one overridable method, but that's something to play with in a separate PR.

rachel-fenichel:
Mocha failed.  I'm investigating.

rachel-fenichel:
The mocha issue was because I put the rendering init call into inject, and the tests were creating a workspaceSvg without calling inject.  I moved the call into workspaceSvg.

RoboErikG:
Makes sense. Might make things easier if we ever decide we want to render two instances of Blockly differently on the same page, too.

Issue #2803, opened 2019-08-12
Formalize type declarations/annotations for shape objects for rendering.


rachel-fenichel:
Objects like `Blockly.blockRendering.constants.PUZZLE_TAB` are used for rendering now, and should be swappable as long as we have a good description of what properties each one needs to have.  Right now that's lacking.

Here is a sample object:
```
Blockly.blockRendering.constants.NOTCH = (function() {
  var width = Blockly.blockRendering.constants.NOTCH_WIDTH;
  var height = Blockly.blockRendering.constants.NOTCH_HEIGHT;
  var innerWidth = 3;
  var outerWidth = (width - innerWidth) / 2;
  function makeMainPath(dir) {
    return Blockly.utils.svgPaths.line(
        [
          Blockly.utils.svgPaths.point(dir * outerWidth, height),
          Blockly.utils.svgPaths.point(dir * innerWidth, 0),
          Blockly.utils.svgPaths.point(dir * outerWidth, -height)
        ]);
  }
  // TODO: Find a relationship between width and path
  var pathLeft = makeMainPath(1);
  var pathRight = makeMainPath(-1);

  return {
    width: width,
    height: height,
    pathLeft: pathLeft,
    pathRight: pathRight
  };
})();
```

Notice that it evaluates immediately, so the actual object just has `width`, `height`, `pathLeft`, and `pathRight` properties for the rest of the rendering code to use.

rachel-fenichel:
Additional note from the rendering hackathon: some shape objects call `makeMainPath` with `-1` or `1`, while others use `true` or `false`.

We should be consistent.

gonfunko:
Closing since these are all now full Typescript interfaces in renderers/common/constants.ts.

PR #2909, opened 2019-08-24
Add a function to get the shape object of a connection, and use it

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Makes it possible to have different shapes for different connection types!

### Proposed Changes

- Add a shapeFor method on the constant provider
- Get rid of the `connectionShape` and `notchShape` properties on measurables and rows
- Add a `shape` property for connection measurables
- Populate the `shape` property based on the connection model
- Change over all references to those properties
- Add a demo in Zelos of using this to make boolean connections a different shape

### Reason for Changes

This means there's a single place where you can specify what shape a connection is based on the connection itself.  That lets you specify based on the connection's checks, or on the parent block, or most other things.

In the base case you just use this to say that previous and next connections get notches and input and output connections get puzzle tabs.  But this opens up an API that developers have been requesting for a long time, and one that is necessary for Zelos.

Note that there are a few places where we were relying on the height of the notch despite not having a connection (in top and bottom row) which may need to just refer to a different constant.

### Additional Information

The drawers are still assuming that the shape objects have the properties that they want (pathUp, pathLeft, etc).  #2803 covers standardizing the set of properties on these objects.  (Notice that the return type is just !Object, which is not especially useful.)


core/renderers/block_rendering_rewrite/constants.js
```diff
@@ -354,4 +354,24 @@ Blockly.blockRendering.ConstantProvider.prototype.makeOutsideCorners = function(
...
+    connection) {
+  switch (connection.type) {
+    case Blockly.INPUT_VALUE:
+    case Blockly.OUTPUT_VALUE:
+      return this.PUZZLE_TAB;
+    case Blockly.PREVIOUS_STATEMENT:
+    case Blockly.NEXT_STATEMENT:
+      return this.NOTCH;
+    default:
+      throw new Error('Unknown type');
```
samelhusseini: Unknown connection type?

PR #2915, opened 2019-08-26
Fix configure method

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

The issue here is the ``configure_`` method is overriding the parent field method and is being called twice. By having the same name and on the prototype it's overriding the superclass method which isn't the desired behaviour here. 

@BeksOmega apologies for not noticing this sooner. The closure compiler pointed this out to me!

### Proposed Changes

This is by no means the solution we must go with, there are a few ways to fix this. This is just a proposal that uses a developer hook to solve the configuration problem. 
@BeksOmega I'd love your feedback here.

The proposal adds a developer hook that is called during initialization. This was done in order to avoid having each subclass having to call configure, and also to avoid having to make sure the call to the super (in the subclass constructor) is the last thing called.

### Reason for Changes

Configure is broken the way it is. The intention here is to avoid adding configuration options in the constructor, but the way it's done overrides instead of extends.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -163,7 +163,7 @@ Blockly.Field.prototype.visible_ = true;
 
 /**
  * The element the click handler is bound to.
- * @type {!Element}
+ * @type {Element}
  * @private
  */
 Blockly.Field.prototype.clickTarget_ = null;
@@ -769,15 +769,18 @@ Blockly.Field.prototype.getValue = function() {
 /**
  * Used to validate a value. Returns input by default. Can be overridden by
  * subclasses, see FieldDropdown.
- * @param {*} newValue The value to be validated.
+ * @param {*=} opt_newValue The value to be validated.
  * @return {*} The validated value, same as input by default.
  * @protected
  * @suppress {deprecated}
  */
-Blockly.Field.prototype.doClassValidation_ = function(newValue) {
+Blockly.Field.prototype.doClassValidation_ = function(opt_newValue) {
+  if (opt_newValue === null || opt_newValue === undefined) {
+    return null;
+  }
   // For backwards compatibility.
-  newValue = this.classValidator(/** @type {string} */ (newValue));
-  return newValue;
+  opt_newValue = this.classValidator(/** @type {string} */ (opt_newValue));
+  return opt_newValue;
 };
 
 /**
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -436,16 +436,16 @@ Blockly.FieldDropdown.prototype.getOptions = function() {
 
 /**
  * Ensure that the input value is a valid language-neutral option.
- * @param {string} newValue The input value.
+ * @param {*=} opt_newValue The input value.
  * @return {?string} A valid language-neutral option, or null if invalid.
  * @protected
  */
-Blockly.FieldDropdown.prototype.doClassValidation_ = function(newValue) {
+Blockly.FieldDropdown.prototype.doClassValidation_ = function(opt_newValue) {
   var isValueValid = false;
   var options = this.getOptions();
   for (var i = 0, option; option = options[i]; i++) {
     // Options are tuples of human-readable text and language-neutral values.
-    if (option[1] == newValue) {
+    if (option[1] == opt_newValue) {
       isValueValid = true;
       break;
     }
@@ -454,11 +454,11 @@ Blockly.FieldDropdown.prototype.doClassValidation_ = function(newValue) {
     if (this.sourceBlock_) {
       console.warn('Cannot set the dropdown\'s value to an unavailable option.' +
         ' Block type: ' + this.sourceBlock_.type + ', Field name: ' + this.name +
-        ', Value: ' + newValue);
+        ', Value: ' + opt_newValue);
     }
     return null;
   }
-  return newValue;
+  return /** @type {string} */ (opt_newValue);
 };
 
 /**
--- a/core/field_image.js
+++ b/core/field_image.js
@@ -168,15 +168,15 @@ Blockly.FieldImage.prototype.initView = function() {
 
 /**
  * Ensure that the input value (the source URL) is a string.
- * @param {string} newValue The input value.
+ * @param {*=} opt_newValue The input value.
  * @return {?string} A string, or null if invalid.
  * @protected
  */
-Blockly.FieldImage.prototype.doClassValidation_ = function(newValue) {
-  if (typeof newValue != 'string') {
+Blockly.FieldImage.prototype.doClassValidation_ = function(opt_newValue) {
+  if (typeof opt_newValue != 'string') {
     return null;
   }
-  return newValue;
+  return opt_newValue;
 };
 
 /**
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -103,15 +103,15 @@ Blockly.FieldLabel.prototype.initView = function() {
 
 /**
  * Ensure that the input value casts to a valid string.
- * @param {string} newValue The input value.
+ * @param {*=} opt_newValue The input value.
  * @return {?string} A valid string, or null if invalid.
  * @protected
  */
-Blockly.FieldLabel.prototype.doClassValidation_ = function(newValue) {
-  if (newValue === null) {
+Blockly.FieldLabel.prototype.doClassValidation_ = function(opt_newValue) {
+  if (opt_newValue === null || opt_newValue === undefined) {
     return null;
   }
-  return String(newValue);
+  return String(opt_newValue);
 };
 
 Blockly.fieldRegistry.register('field_label', Blockly.FieldLabel);
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -109,15 +109,15 @@ Blockly.FieldTextInput.prototype.spellcheck_ = true;
 
 /**
  * Ensure that the input value casts to a valid string.
- * @param {string} newValue The input value.
- * @return {?string} A valid string, or null if invalid.
+ * @param {*=} opt_newValue The input value.
+ * @return {*} A valid string, or null if invalid.
  * @protected
  */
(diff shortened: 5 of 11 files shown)


samelhusseini:
After chatting with @BeksOmega offline and with the rest of the team, we have decided to revert the changes that validate the value passed in the constructor. We shall keep the subclass field value optional and have each of the fields resort to a default value there.

Another change we're going to do is move setValue out of the field superclass constructor. This lets us configure elements that affect validation with ``opt_config``.

I've also gone for a standard inheritance model for configure_. Any field that overrides that, should call super so that it can do its configuration. 

Also fixed all the tests, it was mostly calling ``initModel`` so that the initial value is set, and the field is configured.

samelhusseini:
Okay this is ready for review. Unfortunately it's a little bigger than I'd hoped. A lot of tests needed updating, but it's pretty much the same pattern everywhere.

core/field.js
```diff
@@ -51,15 +51,55 @@ goog.require('Blockly.utils.style');
...
+
+  /**
+   * The initial value to be set when the field is initialized.
+   * @type {*}
+   * @private
+   */
+  this.initialValue_ = value;
+
+  /**
+   * Validation function called when user edits an editable field.
```
BeksOmega: Nit: Does this need doc updates to match initialValue_?

core/field_dropdown.js
```diff
@@ -63,6 +63,10 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
   this.trimOptions_();
   var firstTuple = this.getOptions()[0];
 
+  // Call parent's constructor.
+  Blockly.FieldDropdown.superClass_.constructor.call(this, firstTuple[1],
```
BeksOmega: Q: Do you think moving this super to the top of the constructor needs looking into (after this PR)?

core/field_image.js
```diff
@@ -47,6 +47,9 @@ goog.require('Blockly.utils.Size');
  */
 Blockly.FieldImage = function(src, width, height,
     opt_alt, opt_onClick, opt_flipRtl) {
+  Blockly.FieldImage.superClass_.constructor.call(
+      this, src || '', null);
```
BeksOmega: I'm not sure how this should be handled, but it feels weird to pass `src || ''` when `src` is required.

I don't have a better solution though :/

core/field_label.js
```diff
@@ -43,13 +43,26 @@ goog.require('Blockly.utils.Size');
...
-  this.size_ = new Blockly.utils.Size(0, Blockly.Field.TEXT_DEFAULT_HEIGHT);
-  this.class_ = opt_class;
-  opt_value = this.doClassValidation_(opt_value);
-  if (opt_value === null) {
+  if (opt_value == null) {
     opt_value = '';
   }
-  this.setValue(opt_value);
+  Blockly.FieldLabel.superClass_.constructor.call(
+      this, opt_value, null);
```
BeksOmega: Nit: `opt_value || ''` instead? - Unless we want to accept false and convert it to a string.

core/field_textinput.js
```diff
@@ -50,12 +50,12 @@ goog.require('Blockly.utils.userAgent');
...
 Blockly.FieldTextInput = function(opt_value, opt_validator) {
-  opt_value = this.doClassValidation_(opt_value);
-  if (opt_value === null) {
+  if (opt_value == null) {
     opt_value = '';
   }
-  Blockly.FieldTextInput.superClass_.constructor.call(this, opt_value,
-      opt_validator);
+  Blockly.FieldTextInput.superClass_.constructor.call(this,
+      opt_value, opt_validator);
```
BeksOmega: Ditto label fields.

demos/custom-fields/field_turtle.js
```diff
@@ -46,24 +46,21 @@ CustomFields.FieldTurtle = function(
...
   CustomFields.FieldTurtle.superClass_.constructor.call(
       this, value, opt_validator);
+
+  /**
+   * The size of the area rendered by the field.
+   * @type {Blockly.utils.Size}
+   * @protected
+   * @override
+   */
   this.size_ = new Blockly.utils.Size(0, 0);
```
BeksOmega: Not a problem with your PR, but can this be removed since (0, 0) is the default size?

tests/blocks/test_blocks.js
```diff
@@ -964,55 +964,6 @@ Blockly.Blocks['test_validators_colour_red_null'] = {
   }
 };
 
-Blockly.Blocks['test_validators_date_null'] = {
```
BeksOmega: Thank you!

tests/playground.html
```diff
@@ -1453,13 +1453,6 @@ <h1>Blockly Playground</h1>
         <block type="test_validators_colour_force_red"></block>
         <sep gap="12"></sep>
         <block type="test_validators_colour_red_null"></block>
-        <label text="Dates"></label>
```
BeksOmega: And thank you!

core/field.js
```diff
@@ -51,15 +51,55 @@ goog.require('Blockly.utils.style');
...
+   * @protected
+   */
+  this.value_ = null;
+
+  /**
+   * Whether or not the field's value has been initialized.
+   * @type {!boolean}
+   * @private
+   */
+  this.valueInitialized_ = false;
```
BeksOmega: Oh my gosh you read my mind.

core/input.js
```diff
@@ -118,6 +118,7 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
   if (this.sourceBlock_.rendered) {
     field.init();
   }
+  field.initValue();
```
BeksOmega: Just a heads up for you: I'm was thinking about changing this area a little bit (even before you put up this PR). It works perfectly for now, I just wanted to warn you so I wasn't stepping on toes.
And I'll put up an issue before I do anything so that we can discuss.

core/field_image.js
samelhusseini: src is a required field, but nothing about the constructor description says you can't pass null. ie: there's a difference between 
```
 * @param {string} src The URL of the image. Defaults to an empty string. 
``` and
```
 * @param {!string} src The URL of the image. Defaults to an empty string.
```

The former means you can still pass it null, hence the fallback to the empty string.

File: core/field_dropdown.js (hunk 2 above)
Comment (samelhusseini) [reply core/field_dropdown.js]: Yep, and there's an obvious place for it to go (hint: initValue).

**Comment by samelhusseini** [comment]
Yes, sorry I realized I didn't update my explanation with why I chose ``initValue`` over ``initModel``.
We want initValue to run in all cases, including headless. Currently initModel doesn't get initialized in headless, and upgrading that functionality breaks variables as it expects a very peculiar initialization order.

File: core/field_label.js (hunk 4 above)
Comment (samelhusseini) [reply core/field_label.js]: Hmm, I think this is intentional, so you can do things like: 
``new Blockly.FieldLabel(false)`` and have the value be ``false`` (which is then converted to the string "false" when rendered).
Otherwise the above would yield an empty string.

File: core/field.js (hunk 1 above)
Comment (samelhusseini) [reply core/field.js]: Yep!

**Comment by BeksOmega** [comment]
> Currently initModel doesn't get initialized in headless.

It does (mostly) get called in headless, it just gets called by the block instead of the input:
```
Blockly.FieldVariable.initModel | @ | field_variable.js:108
  | Blockly.Block.initModel | @ | block.js:355
  | Blockly.Xml.domToBlock | @ | xml.js:548
  | Blockly.Xml.domToWorkspace | @ | xml.js:403
  | fromXml | @ | my_playground.html:383
  | onclick
```
So it works, at least when deserializing from xml.

Luckily I think this was just a [quick-fix for headless](https://github.com/google/blockly/pull/1520), and not some necessary life cycle thing. This is part of what I'm looking at fixing with the inputs.

But! None of that actually affects configuration (plus as you mentioned variable fields are cray) so initValue sounds good for now :D

File: core/field.js [hunk 11]
```diff
@@ -196,20 +223,17 @@ Blockly.Field.prototype.EDITABLE = true;
 Blockly.Field.prototype.SERIALIZABLE = false;
 
 /**
- * Configure the field based on the given map of options.
- * @param {Object} opt_config The map of options to configure the field
- *    based on.
- * @private
+ * Process the configuration map passed to the field.
+ * @param {Object} config A map of options used to configure the field. See
```
NeilFraser: Looks like config can't be null.  {!Object}

samelhusseini: Yep

NeilFraser:
For what it is worth, it appears that calling 'super' first is *not* a pattern in any of the dynamic languages which JavaScript was inspired by.  Here are some examples of each language I found in the wild:

Smalltalk (super):
```
Dog class»new 
    count := count +1. 
    ^ super new
```

Scheme (super-new):
```
(send
    (new
      (class object%
        (init-field a)
        (super-new)
        (define/public (hello) a))
      [a 22])
    hello)
```

Lisp (super-new):
```
(define rectangle%
  (class object%
    (init width)
    (init height)
    (super-new)
    (define curr-height height)
    (define curr-width width)))
```

Python (super):
```
def __init__(self, enum: Enum, value: int, name: Text = ''):
  del value  # Unused.
  super().__init__()
  self._dtype = enum
  self._name = keywords.error_if_invalid_name(name) if name else ''
```

Lambda Moo (pass):
```
if (parent(this) == $player)
  $wiz_utils:mentor_alert(this.name, " types a bad command: " + callers()[$][2]);
endif
return pass(@args);
```

Self (resend):
```
c _Define: (|
        p*<- a.
        b = b.
        f = ('c' printLine. resend.f).

|)
```

Perl (SUPER):
```
sub save {
    my $self = shift;
    say 'Prepare to rock';
    $self->SUPER::save();
}
```

Lua (named call):
```
function KCCScene:onEnter()
  doThis()
  CCScene.onEnter(self)
  doThat()
end
```

The only languages I can find that prefer or enforce a super-first policy are typed languages.

PR #2923, opened 2019-08-27
Fixed comment ownership.

### LINKED ISSUE #1969: Block comments should not be owned by the icon in a rendered workspace
### Problem statement

On a rendered workspace the icon for opening the comment also owns the text for that comment. This means the block model depends on the rendered icon for information about the comment text (and position). This information should be independent of the icon so that it isn't lost when loaded/saved in a headless workspace.

### Expected Behavior

- Comment data is preserved in headless workspaces.
- There should not be any special cases for serializing comments.

### Actual Behavior

- Only the text is preserved in a headless workspace.
- We have to special case delete events to avoid losing the comment when its icon is destroyed.

### Steps to Reproduce

See discussion in #1960 

### Stack Traces

```
```

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #1969 

### Proposed Changes

* Moved comment icons to use a model-based system. The block holds the model of the comment, and the comment icon holds a reference to it.
* Reorganized the setVisible function.
* Changed how xml.js serializes and deserializes comments.

Some other little things:
* Removed blockSvg's getCommentText function.
* Changed bubble.getSize to return a Blockly.utils.Size object.
* Removed a rendered_ check from the bubble.
* Changed icons to properly call the parent updateEditable function.

### Reason for Changes

* Fixed block disposal/delete events.
* The previous organization caused erroneous events to be fired.
* Fixed round-tripping through a headless workspace.
* This can now be handled by the headless block.
* Type specificity is good.
* This was unnecessary and causes icons to have to set the size of the bubble, even though they just passed it to the constructor.
* Code health.

### Test Coverage

Added tests for:
* Comment Serialization (headless and rendered)
* Comment Deserialization (headless and rendered)
* Round Tripping: Rendered -> Xml -> Headless -> Xml
* Round Tripping: Headless-> Xml -> Rendered-> Xml
* GetCommentText (headless and rendered)
* SetCommentText (headless and rendered)
* Comment.setVisible
* Comment Editability
* Comment.setBubbleSize

Tested on:
* Desktop Chrome

### Documentation

I don't believe so.

### Additional Information

N/A

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -25,7 +25,6 @@
 'use strict';
 
 goog.provide('Blockly.Block');
-goog.provide('Blockly.Block.CommentModel');
 
 goog.require('Blockly.Blocks');
 goog.require('Blockly.Connection');
@@ -135,6 +134,7 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
    * @deprecated August 2019. Use getCommentText or getCommentIcon instead.
    */
   this.comment = null;
+
   /**
    * A model of the comment attached to this block.
    * @type {!Blockly.Block.CommentModel}
@@ -221,7 +221,6 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
 };
 
 /**
- * A model of a comment.
  * @typedef {{
  *            text:?string,
  *            pinned:boolean
--- a/core/block.js
+++ b/core/block.js
@@ -223,7 +223,7 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
 /**
  * @typedef {{
  *            text:?string,
- *            pinned:boolean
+ *            pinned:boolean,
  *            size:Blockly.utils.Size
  *          }}
  */
--- a/core/block.js
+++ b/core/block.js
@@ -134,20 +134,9 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
    * @deprecated August 2019. Use getCommentText or getCommentIcon instead.
    */
   this.comment = null;
-
-  // TODO: There may be a better place for this typedef, or it could simply
-  //  be removed.
-  /**
-   * A model of a comment.
-   * @typedef {Object} TextBubbleModel
-   * @property {?string} text - The text of the comment.
-   * @property {boolean} pinned - Whether the comment is open or not.
-   * @property {Blockly.utils.Size} size - The height and width of the bubble.
-   */
-
   /**
    * A model of the comment attached to this block.
-   * @type {TextBubbleModel}
+   * @type {!Blockly.Block.CommentModel}
    * @package
    */
   this.commentModel = {
@@ -229,6 +218,18 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
     this.setOnChange(this.onchange);
   }
 };
+
+/**
+
+/**
+ * A model of a comment.
+ * @typedef {{
+ *            text:?string,
+ *            pinned:boolean
+ *            size:Blockly.utils.Size
+ *          }}
+ */
+Blockly.Block.CommentModel;
 
 /**
  * Optional text data that round-trips between blocks and XML.
--- a/core/block.js
+++ b/core/block.js
@@ -25,6 +25,7 @@
 'use strict';
 
 goog.provide('Blockly.Block');
+goog.provide('Blockly.Block.CommentModel');
 
 goog.require('Blockly.Blocks');
 goog.require('Blockly.Connection');
@@ -218,8 +219,6 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
     this.setOnChange(this.onchange);
   }
 };
-
-/**
 
 /**
  * A model of a comment.
--- a/core/block.js
+++ b/core/block.js
@@ -128,8 +128,33 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
    */
   this.collapsed_ = false;
 
-  /** @type {string|Blockly.Comment} */
+  /**
+   * A string/Blockly.Comment representing the comment attached to this block.
+   * @type {string|Blockly.Comment}
+   * @deprecated August 2019. Use getCommentText or getCommentIcon instead.
+   */
   this.comment = null;
+
+  // TODO: There may be a better place for this typedef, or it could simply
+  //  be removed.
+  /**
+   * A model of a comment.
+   * @typedef {Object} TextBubbleModel
+   * @property {?string} text - The text of the comment.
+   * @property {boolean} pinned - Whether the comment is open or not.
+   * @property {Blockly.utils.Size} size - The height and width of the bubble.
+   */
+
+  /**
+   * A model of the comment attached to this block.
+   * @type {TextBubbleModel}
+   * @package
+   */
+  this.commentModel = {
+    text: null,
+    pinned: false,
+    size: new Blockly.utils.Size(160, 80)
+  };
 
   /**
    * The block's position in workspace units.  (0, 0) is at the workspace's
@@ -1795,11 +1820,13 @@ Blockly.Block.prototype.getCommentText = function() {
  * @param {?string} text The text, or null to delete.
  */
 Blockly.Block.prototype.setCommentText = function(text) {
-  if (this.comment != text) {
-    Blockly.Events.fire(new Blockly.Events.BlockChange(
-        this, 'comment', null, this.comment, text || ''));
-    this.comment = text;
+  if (this.commentModel.text == text) {
+    return;
   }
+  Blockly.Events.fire(new Blockly.Events.BlockChange(
+      this, 'comment', null, this.commentModel.text, text || ''));
+  this.commentModel.text = text;
+  this.comment = text;  // For backwards compatibility.
 };
 
 /**
--- a/core/block.js
+++ b/core/block.js
@@ -40,7 +40,6 @@ goog.require('Blockly.utils.Coordinate');
 goog.require('Blockly.utils.object');
 goog.require('Blockly.fieldRegistry');
 goog.require('Blockly.utils.string');
-goog.require('Blockly.Warning');
 goog.require('Blockly.Workspace');
 
 
--- a/core/block.js
+++ b/core/block.js
@@ -1812,7 +1812,7 @@ Blockly.Block.prototype.getInputTargetBlock = function(name) {
  * @return {string} Block's comment.
  */
 Blockly.Block.prototype.getCommentText = function() {
-  return this.comment || '';
+  return this.commentModel.text || '';
 };
 
 /**
--- a/core/block.js
+++ b/core/block.js
@@ -129,9 +129,9 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
   this.collapsed_ = false;
 
   /**
-   * A string/Blockly.Comment representing the comment attached to this block.
-   * @type {string|Blockly.Comment}
-   * @deprecated August 2019. Use getCommentText or getCommentIcon instead.
+   * A string representing the comment attached to this block.
+   * @type {string}
+   * @deprecated August 2019. Use getCommentText instead.
    */
   this.comment = null;
 
--- a/core/block.js
+++ b/core/block.js
@@ -130,7 +130,7 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
 
   /**
    * A string representing the comment attached to this block.
-   * @type {string}
+   * @type {string|Blockly.Comment}
    * @deprecated August 2019. Use getCommentText instead.
    */
   this.comment = null;
--- a/core/block.js
+++ b/core/block.js
@@ -1808,11 +1808,11 @@ Blockly.Block.prototype.getInputTargetBlock = function(name) {
 };
 
 /**
- * Returns the comment on this block (or '' if none).
+ * Returns the comment on this block (or null if there is no comment).
  * @return {string} Block's comment.
  */
 Blockly.Block.prototype.getCommentText = function() {
-  return this.commentModel.text || '';
+  return this.commentModel.text;
 };
 
 /**
@@ -1824,7 +1824,7 @@ Blockly.Block.prototype.setCommentText = function(text) {
     return;
   }
   Blockly.Events.fire(new Blockly.Events.BlockChange(
-      this, 'comment', null, this.commentModel.text, text || ''));
+      this, 'comment', null, this.commentModel.text, text));
   this.commentModel.text = text;
   this.comment = text;  // For backwards compatibility.
 };
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -258,8 +258,16 @@ Blockly.BlockSvg.prototype.mutator = null;
 /**
  * Block's comment icon (if any).
  * @type {Blockly.Comment}
+ * @deprecated August 2019. Use getCommentIcon instead.
  */
 Blockly.BlockSvg.prototype.comment = null;
+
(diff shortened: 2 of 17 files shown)


core/comment.js
```diff
@@ -250,44 +309,60 @@ Blockly.Comment.prototype.getBubbleSize = function() {
...
  */
 Blockly.Comment.prototype.getText = function() {
-  return this.textarea_ ? this.textarea_.value : this.text_;
+  return this.model_.text;
 };
 
 /**
  * Set this comment's text.
+ *
+ * If you want to get the event, then this should not be called directly.
```
rachel-fenichel: Get what event?

core/warning.js
```diff
@@ -115,38 +115,55 @@ Blockly.Warning.textToDom_ = function(text) {
...
+    this.disposeBubble_();
+  }
+};
+
+/**
+ * Show the bubble.
+ * @package
+ */
+Blockly.Warning.prototype.createBubble = function() {
+  // TODO: This is package because comments steal this UI for non-editable
```
rachel-fenichel: Make a numbered issue for this and the next TODO

BeksOmega:
@samelhusseini  If you have time could I have some help with typing?

Problem:
The block owns the model of the comment (an object).
And the comment holds a reference to that model.

I figured it would be good to add a typedef for the model, so that it's more descriptive, but I'm not sure what the best place for the typedef to live is.

Travis doesn't seem to like the way I have it set up currently, but maybe it just needs a recompile?

I'd really appreciate your input on this! Who knows, maybe it doesn't need a typedef at all ¯\\\_(ツ)_/¯
Thank you for any help you can give me!
Beka

core/block.js
```diff
@@ -206,6 +221,15 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
   }
 };
 
+/**
+ * @typedef {{
+ *            text:?string,
+ *            pinned:boolean
```
samelhusseini: You just need a comma after ``pinned:boolean`` to fix typings ;)

BeksOmega: Oh gosh (－‸ლ) thank you for the help.

core/comment.js
```diff
@@ -26,6 +26,7 @@
 
 goog.provide('Blockly.Comment');
 
+goog.require('Blockly.Block');
```
samelhusseini: There's a circular dependency here that creates a compiler error:
Blockly.Comment requires Blockly.Block, and Blockly.Block requires Blockly.Comment..

core/icon.js
```diff
@@ -26,6 +26,7 @@
 
 goog.provide('Blockly.Icon');
 
+goog.require('Blockly.BlockSvg');
```
samelhusseini: Same here.

BeksOmega: Q about typing: So if the icon is required by the block svg, the icon can reference block svg members?

core/block.js
```diff
@@ -27,7 +27,6 @@
 goog.provide('Blockly.Block');
 
 goog.require('Blockly.Blocks');
-goog.require('Blockly.Comment');
```
samelhusseini: It's okay to remove it from here, but somewhere in Blockly we need to include it, otherwise compressed won't have it. It looks like block_svg is the appropriate place to place that, as it instantiates a comment in ``setCommentText``.

core/icon.js
samelhusseini: Not a typing question per se, but a closure compiler question, and the answer is a little complex: 
You've identified there's a circular dependency there, ``Block`` requires ``Icon``, and ``Icon`` requires ``Block``.

We know that we will never instantiate an icon in an environment that doesn't have ``Blockly.BlockSvg`` included, so you can choose to just leave it out, and we do that in much of Blockly.
However, the same reason why you thought you needed it, it looks a little odd using and expecting it in ``Icon``, but not requiring it.

The closure compiler does have a term for this: forward declaration. 
Here, you can use: ``goog.forwardDeclare('Blockly.BlockSvg')`` basically stating that you need it for this "module" but it will be declared in the future.

core/block.js
```diff
@@ -129,9 +128,24 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
    */
   this.collapsed_ = false;
 
-  /** @type {string|Blockly.Comment} */
```
samelhusseini: Even if this is deprecated, you want to keep the type as string|Blockly.Comment. 
Otherwise, ``this.comment`` is of type ``string`` in the superclass (Blockly.Block)
but ``this.comment`` is of type ``Blockly.Comment`` in the subclass (Blockly.BlockSvg)
And subclasses don't have the ability to just change the type of a property. They can only restrict.

core/comment.js
```diff
@@ -177,71 +208,99 @@ Blockly.Comment.prototype.resizeBubble_ = function() {
...
-  // However doing so in Firefox swallows the cursor for unknown reasons.
-  // So this is hooked to mouseup instead.  No big deal.
+Blockly.Comment.prototype.startEdit_ = function(_e) {
   if (this.bubble_.promote_()) {
     // Since the act of moving this node within the DOM causes a loss of focus,
     // we need to reapply the focus.
     this.textarea_.focus();
   }
+
+  this.cachedText_ = this.model_.text;
```
samelhusseini: type mismatch here. this.model_.text is string or null. cachedText is expecting a string.

core/comment.js
```diff
@@ -177,71 +208,99 @@ Blockly.Comment.prototype.resizeBubble_ = function() {
...
+ * Show a non-editable bubble.
+ * @private
+ */
+Blockly.Comment.prototype.createNonEditableBubble_ = function() {
+  // Steal the code from warnings to make an uneditable text bubble.
+  // MSIE does not support foreignobject; textareas are impossible.
+  // https://docs.microsoft.com/en-us/openspecs/ie_standards/ms-svg/56e6e04c-7c8c-44dd-8100-bd745ee42034
+  // Always treat comments in IE as uneditable.
+  // TODO (#2917): It would be great if the comment could support line breaks.
+  Blockly.Warning.prototype.createBubble.call(this, true);
```
samelhusseini: Calling createBubble with an additional parameter?

core/comment.js
```diff
@@ -177,71 +208,99 @@ Blockly.Comment.prototype.resizeBubble_ = function() {
...
+};
+
+/**
+ * Dispose of the bubble.
+ * @private
+ */
+Blockly.Comment.prototype.disposeBubble_ = function() {
+  if (this.paragraphElement_) {
+    // We're using the warning UI so we have to let it dispose.
+    Blockly.Warning.prototype.disposeBubble.call(this);
```
samelhusseini: The compiler really isn't going to like this.. but we can fix as part of #2943

core/icon.js
```diff
@@ -34,10 +35,15 @@ goog.require('Blockly.utils.Size');
...
  * Class for an icon.
- * @param {Blockly.Block} block The block associated with this icon.
+ * @param {Blockly.BlockSvg} block The block associated with this icon.
  * @constructor
  */
 Blockly.Icon = function(block) {
+  /**
+   * The block this icon is attached to.
+   * @type {Blockly.BlockSvg}
+   * @private
```
samelhusseini: protected so the Comment / Warning can access.

BeksOmega:
Ok, so everything is passing but now Blockly.Block has a Blockly.Comment type declaration, even though it doesn't require Blockly.Comment. Is that cool with the compiler? because it feels kinda weird.

I feel like I really need to learn more about the relationship between typing and requiring and how the compiler works. Is there documentation you would recommend? I'm having no luck finding any :/ (I tried looking up forwardDeclare and got nowhere.)

samelhusseini:
It's fine since it's only a type declaration and there's no actual use of it there. 
Closure's requires/provides are a way to create separate files and have them depend on each other (pre ES6), perhaps you could refer to them as modules (but that isn't technically corrrect).
https://github.com/google/closure-compiler/wiki/Managing-Dependencies

For forward declare, read up about it here: (I could't find much else about it either)
https://github.com/google/blockly/blob/develop/closure/goog/base.js#L708

There's not a huge amount of relationship between typing and requiring. Types are used by the compiler to type check everything, and find mistakes, optimize, and do a bunch of other magic! Requiring is used for dependency management.

I'm sure you've seen these two before, but I'll link them for completeness: 
https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler
https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System

And finally, there's this blog post that I think would be particularly helpful to you. (I know this because I know there's a few issues with the typings of doValueUpdate / Invalid, and it's because of a misunderstanding that this blog post covers)
http://closuretools.blogspot.com/2012/06/subtyping-functions-without-poking-your.html

Let me know if you have any further questions, I'm happy to answer questions along the way :)

BeksOmega:
Thank you for all of the links! I'm excited to read up on it! :D

RoboErikG:
Does this need another review?

BeksOmega:
> Does this need another review?

I believe all of the current comments have been resolved. I just need to fix merge conflicts.

If you want to give it another review I'd love that :D

core/comment.js
```diff
@@ -44,27 +44,28 @@ goog.require('Blockly.utils.userAgent');
...
-Blockly.utils.object.inherits(Blockly.Comment, Blockly.Icon);
 
-/**
- * Comment text (if bubble is not visible).
- * @private
- */
-Blockly.Comment.prototype.text_ = '';
+  /**
+   * The model for this comment.
+   * @type {Blockly.Block.CommentModel}
```
samelhusseini: ``@type {!Blockly.Block.CommentModel}``

core/comment.js
```diff
@@ -104,41 +105,58 @@ Blockly.Comment.prototype.drawIcon_ = function(group) {
...
+  // However doing so in Firefox swallows the cursor for unknown reasons.
+  // So this is hooked to mouseup instead.  No big deal.
+  Blockly.bindEventWithChecks_(textarea, 'mouseup', this, this.startEdit_,
       true, true);
   // Don't zoom with mousewheel.
   Blockly.bindEventWithChecks_(textarea, 'wheel', this, function(e) {
     e.stopPropagation();
   });
   Blockly.bindEventWithChecks_(textarea, 'change', this, function(_e) {
-    if (this.text_ != textarea.value) {
```
samelhusseini: Do we still need the onchange event?

core/comment.js
```diff
@@ -104,41 +105,58 @@ Blockly.Comment.prototype.drawIcon_ = function(group) {
...
+    if (this.cachedText_ != this.model_.text) {
       Blockly.Events.fire(new Blockly.Events.BlockChange(
-          this.block_, 'comment', null, this.text_, textarea.value));
-      this.text_ = textarea.value;
+          this.block_, 'comment', null, this.cachedText_, this.model_.text));
     }
   });
+  Blockly.bindEventWithChecks_(textarea, 'input', this, function(_e) {
+    this.model_.text = textarea.value;
```
samelhusseini: Can we just use an ``input`` event and fire the Blockly event here?

core/comment.js
```diff
@@ -147,29 +165,41 @@ Blockly.Comment.prototype.createEditor_ = function() {
  * @override
  */
 Blockly.Comment.prototype.updateEditable = function() {
+  Blockly.Comment.superClass_.updateEditable.call(this);
   if (this.isVisible()) {
-    // Toggling visibility will force a rerendering.
-    this.setVisible(false);
-    this.setVisible(true);
+    // Recreate the bubble with the correct UI.
```
samelhusseini: What's the reason for this change?

core/comment.js
```diff
@@ -178,71 +208,99 @@ Blockly.Comment.prototype.resizeBubble_ = function() {
...
   if (visible == this.isVisible()) {
-    // No change.
     return;
   }
   Blockly.Events.fire(
       new Blockly.Events.Ui(this.block_, 'commentOpen', !visible, visible));
-  if ((!this.block_.isEditable() && !this.textarea_) ||
-      Blockly.utils.userAgent.IE) {
-    // Steal the code from warnings to make an uneditable text bubble.
-    // MSIE does not support foreignobject; textareas are impossible.
```
samelhusseini: I would keep the comment about IE not supporting foreign objects in the same place we check for the useragent. ie: in this new code it's createBubble_

core/block_svg.js
```diff
@@ -29,6 +29,7 @@ goog.provide('Blockly.BlockSvg');
 goog.require('Blockly.Block');
 goog.require('Blockly.blockAnimations');
 goog.require('Blockly.blockRendering.IPathObject');
+goog.require('Blockly.Comment');
```
samelhusseini: This sounds good to me, can we move ``
goog.require('Blockly.Warning');`` into block_svg as part of this change as well. 
Since you're making Icon only accept BlockSvg and Warning is a subclass of Icon.

core/comment.js
```diff
@@ -251,44 +309,60 @@ Blockly.Comment.prototype.getBubbleSize = function() {
...
 };
 
 /**
  * Returns this comment's text.
  * @return {string} Comment text.
+ * @deprecated August 2019 Use block.getCommentText() instead.
  */
 Blockly.Comment.prototype.getText = function() {
-  return this.textarea_ ? this.textarea_.value : this.text_;
+  return this.model_.text;
```
samelhusseini: ``this.model_.text`` can be null, ``return this.model_.text || ''``

core/comment.js
BeksOmega: It avoids firing erroneous events.

core/comment.js
BeksOmega: We could, but it would give slightly different behavior.

'input' occures on every keypress. 'change' occures when the textarea is deselected.

That's a design decision so it's up to you. (┏－‿ ◕)┏

core/comment.js
BeksOmega: That shouldn't happen in this context because a `null` value signals that there is no comment (and by extension no comment icon) but I'll add an `||` for safety.

core/block_svg.js
BeksOmega: I moved warning as well as mutator.

RoboErikG: Note, Neil just removed this require as part of making it optional: https://github.com/google/blockly/pull/3053

core/comment.js
samelhusseini: I see, so previously we weren't changing the ``this.text_`` property on every key press?

core/block_svg.js
samelhusseini: Yeah so you can remove all three from here.

core/comment.js
BeksOmega: That is correct.

BeksOmega:
I'll finish rebasing on Neil's change later this afternoon.

core/block.js
```diff
@@ -40,7 +40,6 @@ goog.require('Blockly.utils.Coordinate');
 goog.require('Blockly.utils.object');
 goog.require('Blockly.fieldRegistry');
 goog.require('Blockly.utils.string');
-goog.require('Blockly.Warning');
```
samelhusseini: Maybe leave this one here for now. If you remove it you'll have to take care of checking if it exists when we need it.

BeksOmega:
> Revert Blockly.Warning require change

Done.

But just to be clear: Deleting the mutator require is ok?

samelhusseini:
Yeah, that was removed by Neil's change. I assume Neil will also be removing the Warning require soon.

PR #2930, opened 2019-08-29
Field Text transformation methods

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

There's no obvious extension path for fields that show a different value than their stored value when editing.

### Proposed Changes

- Add transform methods in ``FieldTextInput`` that allow field subclasses to override to accept a different value than the one stored.
- Removed the ``htmlInputValue_`` cache, because Neil pointed out that the html input is still around when widgetDispose is called.
- Add the Blockly Games ``FieldPitch`` into demos/customblocks. Updated the ``FieldPitch`` to match this architecture.
- Moved the turtle field into a directory within custom fields, and updated references so it all works.
- Moved ``updateGraph`` in ``FieldAngle`` from the render method and into the doUpdate/Invalid methods to match other fields. 

### Test Coverage

* Desktop Chrome

### Documentation

Field customization documentation needs to be updated once all this settles.

### Additional Information

@BeksOmega @NeilFraser would love your feedback on this.


demos/custom-fields/pitch/field_pitch.js
```diff
@@ -0,0 +1,257 @@
...
+};
+
+/** 
+ * Get the text to be displayed on the field node.
+ * @return {?string} The html value if we're editing, otherwise null. Null means
+ *   the super class will handle it, likely a string cast of value.
+ * @protected
+ */
+Blockly.FieldPitch.prototype.getText_ = function() {
+  if (this.isBeingEdited_) {
```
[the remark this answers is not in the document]
samelhusseini: Yeah will revisit later.

demos/custom-fields/pitch/index.html
```diff
@@ -0,0 +1,141 @@
...
+        output.select();
+      }
+
+      function fromXml() {
+        var input = document.getElementById('importExport');
+        var xml = Blockly.Xml.textToDom(input.value);
+        Blockly.Xml.domToWorkspace(xml, workspace);
+      }
+
+      function setRandomStyle() {
```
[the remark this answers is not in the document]
samelhusseini: Yep.

core/field_angle.js
```diff
@@ -132,11 +132,18 @@ Blockly.FieldAngle.prototype.initView = function() {
 };
 
 /**
- * Updates the graph when the field rerenders.
- * @private
+ * @override
+ */
+Blockly.FieldAngle.prototype.doValueUpdate_ = function(newValue) {
+  Blockly.FieldAngle.superClass_.doValueUpdate_.call(this, newValue);
+  this.updateGraph_();
```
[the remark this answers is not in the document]
samelhusseini: Thanks for linking the discussion, I think that makes sense. I'll move view related stuff to render_.
I also updated the ``setEditorValue_`` method to just set the isDirty flag and not forceRerender as setValue will take care of that for us.

core/field_textinput.js
```diff
@@ -294,14 +288,13 @@ Blockly.FieldTextInput.prototype.widgetDispose_ = function() {
   this.isTextValid_ = true;
   // No need to call setValue because if the widget is being closed the
   // latest input text has already been validated.
-  if (this.value_ != this.htmlInputValue_) {
+  if (this.value_ != this.getValueFromEditorText_(this.htmlInput_.value)) {
```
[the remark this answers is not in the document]
samelhusseini: Yep, I'll get rid of that.

demos/custom-fields/pitch/field_pitch.js
```diff
@@ -0,0 +1,257 @@
...
+};
+
+/**
+ * Transform the text received from the html input (note) into a value
+ *    to store in this field.
+ * @param {string} text Text received from the html input.
+ * @returns {string} The value to store.
+ */
+Blockly.FieldPitch.prototype.getValueFromEditorText_ = function(text) {
+  return this.noteToValue(text) || '';
```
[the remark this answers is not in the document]
samelhusseini: I was dealing with the scenario where you could ``setValue`` while the input is open. Since we won't support that, I'll get rid of that.

demos/custom-fields/pitch/field_pitch.js
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Called by setValue if the text input is not valid. If the field is
+ * currently being edited it reverts value of the field to the previous
+ * value while allowing the display text to be handled by the htmlInput_.
+ * @param {*} _invalidValue The input value that was determined to be invalid.
+ *    This is not used by the text input because its display value is stored on
+ *    the htmlInput_.
+ * @protected
+ */
+Blockly.FieldPitch.prototype.doValueInvalid_ = function(_invalidValue) {
```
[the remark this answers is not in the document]
samelhusseini: I'll be removing both these and instead use ``render_``, but good catch!

demos/custom-fields/pitch/field_pitch.js
```diff
@@ -0,0 +1,257 @@
...
+ * All notes available for the picker.
+ */
+Blockly.FieldPitch.NOTES = 'C3 D3 E3 F3 G3 A3 B3 C4 D4 E4 F4 G4 A4'.split(/ /);
+
+/**
+ * Clean up this FieldPitch, as well as the inherited FieldTextInput.
+ * @return {!Function} Closure to call on destruction of the WidgetDiv.
+ * @private
+ */
+Blockly.FieldPitch.prototype.dispose_ = function() {
```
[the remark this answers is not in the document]
samelhusseini: Yeah this was copied over from Blockly Games, I'll move to dropdownDispose_.

samelhusseini:
Thanks for reviewing @BeksOmega. Updated with PR comments. 
I wanted to address your question about the pitch field's class validator accepting both notes and values (numbers).
I think since this is a sample, it would complicate it to accept both, and it would only be valuable for setting the value programatically, so I think let's leave it out.

demos/custom-fields/index.html
```diff
@@ -1,153 +1,54 @@
...
-      <block type="turtle_basic"></block>
-      <block type="turtle_nullifier" y="120"></block>
-      <block type="turtle_changer" y="230"></block>
-    </xml>
-  </body>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Blockly Demos</title>
+  <style>
```
NeilFraser: This CSS ought to be present in both demos, since it's pretty jarring as the font changes between this page and the two linked pages.

demos/custom-fields/index.html
```diff
@@ -1,153 +1,54 @@
...
-    <xml xmlns="https://developers.google.com/blockly/xml" id="workspace-blocks" style="display: none">
-      <block type="turtle_basic"></block>
-      <block type="turtle_nullifier" y="120"></block>
-      <block type="turtle_changer" y="230"></block>
-    </xml>
-  </body>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Blockly Demos</title>
```
NeilFraser: Previous title on this file was more informative.

demos/custom-fields/pitch/custom-fields.css
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+ #notePicker {
```
NeilFraser: This line is indented by one space.

demos/custom-fields/pitch/custom-fields.css
```diff
@@ -0,0 +1,27 @@
...
+    background-image: url(media/notes.png);
```
NeilFraser: Style guide says to use two space indents in CSS:
https://google.github.io/styleguide/htmlcssguide.html#Indentation
Same issue present in turtle/custom-fields.css

demos/custom-fields/pitch/custom-fields.css
```diff
@@ -0,0 +1,27 @@
...
+    border: 1px solid #ccc;
+    height: 109px;
+    width: 46px;
+}
```
NeilFraser: This file's name doesn't reflect that it is specific to the note picker.  Thus it should be called `pitch.css`.
Same issue present with turtle/custom-fields.css

demos/custom-fields/pitch/field_pitch.js
```diff
@@ -0,0 +1,230 @@
...
+ * @author fraser@google.com (Neil Fraser)
+ * @author samelh@google.com (Sam El-Husseini)
+ */
+'use strict';
+
+goog.provide('Blockly.FieldPitch');
+
+goog.require('Blockly.FieldTextInput');
+goog.require('Blockly.utils.math');
+goog.require('Blockly.utils.userAgent');
```
NeilFraser: This require is never used.  Originally showEditor_ had this snippet:
```
  // Mobile browsers have issues with in-line textareas (focus & keyboards).
  var noFocus =
      Blockly.utils.userAgent.MOBILE ||
      Blockly.utils.userAgent.ANDROID ||
      Blockly.utils.userAgent.IPAD;
  Blockly.FieldPitch.superClass_.showEditor_.call(this, noFocus);
```
Has the mobile issue been resolved?

demos/custom-fields/pitch/index.html
```diff
@@ -0,0 +1,103 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>Blockly Demo: Custom Fields</title>
```
NeilFraser: `Blockly Demo: Custom Pitch Field`

demos/custom-fields/turtle/index.html
```diff
@@ -0,0 +1,155 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>Blockly Demo: Custom Fields</title>
```
NeilFraser: `Blockly Demo: Custom Turtle Field`

NeilFraser:
> One thing I'm wondering about: How would you feel about changing the pitch field's class validator to accept both numbers and strings (i.e. 'C4', 'A3', 'G4', etc.) and convert the strings to numbers? Here's a [little example](https://github.com/google/blockly/commit/5f963c796ae48395a26a76552e6aa04680dbfc78) I worked up.
> 
> I think this makes the code look a little simpler, and it means that someone could call `setValue('C4')` on the pitch field. I think saying that `setValue` should accept the same values as those passed through the editor is a good precedent to set.

Yes, this would make things simpler for specific case of the note picker where the numbers (0-12) happen to be human-readable and the text (C3-A4) doesn't change based on language.  Indeed, for this case, if we wanted to make our lives easier, I'd have just used C3-A4 as the value and had a really simple field, with slightly more complex decoding.

But the intent here was to demonstrate the worst-case field where the text and the value were completely different and in no way interchangeable.

BeksOmega:
> Yes, this would make things simpler for specific case of the note picker where the numbers (0-12) happen to be human-readable and the text (C3-A4) doesn't change based on language. Indeed, for this case, if we wanted to make our lives easier, I'd have just used C3-A4 as the value and had a really simple field, with slightly more complex decoding.
> 
> But the intent here was to demonstrate the worst-case field where the text and the value were completely different and in no way interchangeable.

The text being displayed is definitely /different/ than the value being stored, but to me at least they seem pretty interchangable, because the `noteToValue` and `valueToNote` functions allow you to convert between them.

The method* I demonstrated basically just moves `getValueFromEditorText_`'s logic into the class validator, allowing `getValueFromEditorText_` to be removed, and allowing programmatic value setting to work intuitively.

The same logic is happening in both methods*, they're just organized differently.

Your point about language-neutrality demonstrates this perfectly. Using the current method `getValueFromEditorText_` would have to deal with converting language-specific values to notes. Using my proposed method, the class validator would do the same.

 *method meaning "way of doin' stuff", not "function" hehe

demos/custom-fields/pitch/field_pitch.js
samelhusseini: Yep, it's moved up into the text field's showEditor. Will remove the require.

samelhusseini:
Addressed Neil's PR comments.
@BeksOmega I think both Neil and I agree that this can be done by moving the logic to the class validator, but we want this field to be a sample for how to use the converter methods if the value stored in the field is different to value shown to the user in the html input.

samelhusseini:
Thanks all!

demos/custom-fields/pitch/custom-fields.css
samelhusseini: Fixed.

PR #2931, opened 2019-08-29
Fixes toolbox and menu dropdown on mobile.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Toolbox doesn't click on IOS.
Long dropdown's can't be scrolled. 

### Proposed Changes

I can't find a reason for using mousedown on these elements since all we're after is detecting a click. A click works universally, so going with that instead. 

@rachel-fenichel can you think of a reason to do mouseover > click?

### Reason for Changes

Fix mobile

### Test Coverage

Tested on:
* Desktop Chrome
* Desktop Safari

* IOS


rachel-fenichel:
Note that bindEventWithChecks doesn't have anything in the touch map for clicks.  But that doesn't seem like an issue.

This looks reasonable, though mocha is failing.

PR #2938, opened 2019-08-30
Render round right corners in zelos.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Right round corners in zelos.

### Proposed Changes

In order to allow extending of info to add a different element for top / bottom rows, I've had to pull out measure from the constructor.
Extends the top and bottom row to add an additional (right corner) element.
Extends constants to describe how to render the top right and bottom right corners.
Extends drawer to draw the right corners when drawing the top and bottom rows. 
Extends info to add a zero spacing (NO_PADDING) after the right corner element
Extends info to skip the right element and the spacing after it when aligning the top and bottom rows (so as to not add spacing after the right corner elements)

Screen Shot 2019-08-30 at 10 16 10 AM

@rachel-fenichel I think this is most of the way there. 

There is one problem with this, I'm not sure where but the change in height of the top (or maybe bottom) row is not being accounted for, you can see that issue here: 

Screen Shot 2019-08-30 at 10 22 36 AM


core/renderers/zelos/drawer.js
```diff
@@ -45,3 +45,63 @@ Blockly.zelos.Drawer = function(block, info) {
...
+ */
+Blockly.zelos.Drawer.prototype.drawTop_ = function() {
+  var topRow = this.info_.topRow;
+  var elements = topRow.elements;
+
+  this.positionPreviousConnection_();
+  this.outlinePath_ +=
+      Blockly.utils.svgPaths.moveBy(topRow.xPos, this.info_.startY);
+  for (var i = 0, elem; (elem = elements[i]); i++) {
+    if (elem.type == 'round corner') {
```
rachel-fenichel: I prefer 'left round corner' and 'right round corner' to 'round corner' and 'right round corner'.

If you've got ideas about better ways of handling types (that doesn't involve instanceof) lmk.

core/renderers/zelos/drawer.js
```diff
@@ -45,3 +45,63 @@ Blockly.zelos.Drawer = function(block, info) {
 };
 goog.inherits(Blockly.zelos.Drawer, Blockly.blockRendering.Drawer);
 
+
+/**
+ * Add steps for the top corner of the block, taking into account
+ * details such as hats and rounded corners.
+ * @protected
+ */
+Blockly.zelos.Drawer.prototype.drawTop_ = function() {
```
rachel-fenichel: What's the reason to put this stuff in the zelos drawer instead of in the common drawer?

core/renderers/zelos/drawer.js
```diff
@@ -45,3 +45,63 @@ Blockly.zelos.Drawer = function(block, info) {
...
+
+  this.outlinePath_ +=
+    Blockly.utils.svgPaths.lineOnAxis('v', bottomRow.height - bottomRow.overhangY);
+
+  for (var i = elems.length - 1, elem; (elem = elems[i]); i--) {
+    if (elem.isNextConnection()) {
+      this.outlinePath_ += elem.shape.pathRight;
+    } else if (elem.isSquareCorner()) {
+      this.outlinePath_ += Blockly.utils.svgPaths.lineOnAxis('H', bottomRow.xPos);
+    } else if (elem.isRoundedCorner()) {
```
rachel-fenichel: Again, this check and the check against 'round corner right' feels mismatched.

core/renderers/zelos/measurables/rows.js
```diff
@@ -0,0 +1,131 @@
...
+};
+goog.inherits(Blockly.zelos.TopRow, Blockly.blockRendering.TopRow);
+
+/**
+ * Create all non-spacer elements that belong on the top row.
+ * @param {!Blockly.BlockSvg} block The block whose top row this represents.
+ * @package
+ * @override
+ */
+Blockly.zelos.TopRow.prototype.populate = function(block) {
```
rachel-fenichel: Do you have opinions on whether this and the bottomRow `populate` function belong in info instead?

core/renderers/zelos/drawer.js
samelhusseini: I would prefer that as well. I wanted to leave the types as is ``round corner`` instead of adding the left, as there's a few places that check the type directly rather than using the base method ``isRoundCorner``.

In terms of dealing with types, here are a few options:
- do what we're doing now, con with that is that the check method is on the measurable, ie: to add additional types you have to do them in base measurable, otherwise ``elem.isSpecialType()`` might not exist on all elements.
- factory type checker, something like: ``Blockly.isSpecialType(elem)`` that gets passed in an element and checks the type. This can be extended independently in each renderer to types that only matter there.
- Use a enum flag and bitwise operations with types. This is the option I prefer, but it might be an overkill. We could have a central place to register types, which basically internally appends new types to a enum flag. Then you can do things like: 
```
if (elem.type & Blockly.Measurable.Types['left row']) {
}
```
The nice thing about enum flags is you can have union types, so you can do bitwise checks to find if an element is a corner (not really caring about the specific type of corner it is), but then again you could always just do indexOf('corner') haha.

A combination of 2 and 3 would be nice too.

samelhusseini:
The height issue was unrelated, and fixed in https://github.com/google/blockly/pull/2942.

core/renderers/zelos/measurables/rows.js
samelhusseini: Yeah I think that makes sense. Top and bottom rows are the only measurables that are also defining what elements go inside of them in the class definition. InputRows on the other hand rely on ``info.createRows_`` to define what goes inside of them.

core/renderers/zelos/drawer.js
samelhusseini: I was avoiding adding functionality in base that was only relevant to zelos. Ideally I think the base adds corners on either side, and thrasos / geras can have right square corners. 
I tried to do that, but I broke the highlight logic.
If you'd prefer I continue down that path, I can refactor this such that the base renderer supports corners on all sides, and deal with the issues that arise in the other renderers.

rachel-fenichel: Gotcha.  You can leave it this way, and I'll file a bug to follow up and unify it.

core/renderers/zelos/measurables/rows.js
rachel-fenichel: https://github.com/google/blockly/issues/2945

You can do it now or I can do it as a followup once this lands.

core/renderers/zelos/drawer.js
rachel-fenichel: I created a tracking bug: #2946

samelhusseini: Thanks

samelhusseini:
Fixed the height of the bottom row by not running drawOutline_ on the bottom row (similar to the top row).

samelhusseini:
Screen Shot 2019-08-30 at 6 08 20 PM

Issue #2946, opened 2019-08-31
Type checking for Measurables


rachel-fenichel:
Right now Measurables are checked with a combination of string equality tests against the `type` property and calls to `isCorner()`, `isInput()`, etc.

We need to pick a standard approach and use it everywhere.

From discussion with @samelhusseini :

> I would prefer that as well. I wanted to leave the types as is round corner instead of adding the left, as there's a few places that check the type directly rather than using the base method isRoundCorner.
> 
> In terms of dealing with types, here are a few options:
> 
> - do what we're doing now, con with that is that the check method is on the measurable, ie: to add additional types you have to do them in base measurable, otherwise elem.isSpecialType() might not exist on all elements.
> - factory type checker, something like: Blockly.isSpecialType(elem) that gets passed in an element and checks the type. This can be extended independently in each renderer to types that only matter there.
> - Use a enum flag and bitwise operations with types. This is the option I prefer, but it might be an overkill. We could have a central place to register types, which basically internally appends new types to a enum flag. Then you can do things like:
> if (elem.type & Blockly.Measurable.Types['left row']) {
> }
> The nice thing about enum flags is you can have union types, so you can do bitwise checks to find if an element is a corner (not really caring about the specific type of corner it is), but then again you could always just do indexOf('corner') haha.
> 
> A combination of 2 and 3 would be nice too.

samelhusseini:
Fixed.

PR #2950, opened 2019-09-03
Rendering type checker

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2946

### Proposed Changes

Use an enum for rendering element types. Use enum flags so an element can have multiple types embedded in the type field. eg: a top row is a ``ROW`` and a ``TOP_ROW``.

Convert all type checks in the code to use the checks defined in ``Blockly.blockRendering.Types``.

I also added the ability for elements defined by customer renderers to create their own types using the ``Blockly.blockRendering.getType`` helper.

### Reason for Changes

Central place for defining and checking measurable (rendering elements) types.


core/renderers/measurables/types.js
```diff
@@ -0,0 +1,353 @@
...
+ */
+Blockly.blockRendering.Types.nextTypeValue_ = 1 << 24;
+
+/**
+ * Get the enum flag value of an existing type or register a new type.
+ * @param {!string} type The name of the type.
+ * @return {!number} The enum flag value assosiated with that type.
+ * @package
+ */
+Blockly.blockRendering.Types.getType = function(type) {
```
rachel-fenichel: When is this used?

core/renderers/thrasos/info.js
```diff
@@ -89,14 +90,14 @@ Blockly.thrasos.RenderInfo.prototype.shouldStartNewRow_ = function(input, lastIn
 Blockly.thrasos.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
   if (!prev) {
     // Between an editable field and the beginning of the row.
-    if (next.isField() && next.isEditable) {
+    if (next && Blockly.blockRendering.Types.isField(next) && next.isEditable) {
       return this.constants_.MEDIUM_PADDING;
     }
     // Inline input at the beginning of the row.
-    if (next.isInput && next.isInlineInput()) {
+    if (next && next.isInput && Blockly.blockRendering.Types.isInlineInput(next)) {
```
rachel-fenichel: Do we need the isInput check if the next check is against isInlineInput?

core/renderers/thrasos/info.js
```diff
@@ -134,43 +135,43 @@ Blockly.thrasos.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
...
       return this.constants_.LARGE_PADDING;
-    } else if (prev.isStatementInput()) {
+    } else if (Blockly.blockRendering.Types.isStatementInput(prev)) {
       return this.constants_.NO_PADDING;
     }
   }
 
   // Spacing between a non-input and an input.
-  if (!prev.isInput && next.isInput) {
+  if (!prev.isInput && next && next.isInput) {
```
rachel-fenichel: Should isInput be part of the type checker?  If not, why not?

core/renderers/zelos/drawer.js
```diff
@@ -59,20 +60,20 @@ Blockly.zelos.Drawer.prototype.drawTop_ = function() {
   this.outlinePath_ +=
       Blockly.utils.svgPaths.moveBy(topRow.xPos, this.info_.startY);
   for (var i = 0, elem; (elem = elements[i]); i++) {
-    if (elem.type == 'round corner') {
+    if (elem.type & Blockly.blockRendering.Types.LEFT_ROUND_CORNER) {
```
rachel-fenichel: Why do you sometimes use `&` directly and sometimes use it through the helper methods on `Blockly.blockRendering.Types`?

core/renderers/measurables/inputs.js
```diff
@@ -142,7 +144,7 @@ goog.inherits(Blockly.blockRendering.StatementInput,
 Blockly.blockRendering.ExternalValueInput = function(input) {
   Blockly.blockRendering.ExternalValueInput.superClass_.constructor.call(this,
       input);
-  this.type = 'external value input';
+  this.type |= Blockly.blockRendering.Types.EXTERNAL_VALUE_INPUT;
```
rachel-fenichel: What's the rule for deciding whether to use `|=` or just `=` to set the type?

core/renderers/measurables/types.js
rachel-fenichel: Also, why is the name a string here?  Can you index into an enum with a string?

core/renderers/measurables/types.js
```diff
@@ -0,0 +1,353 @@
...
+ */
+
+/**
+ * @fileoverview Measurable types.
+ * @author samelh@google.com (Sam El-Husseini)
+ */
+
+'use strict';
+
+goog.provide('Blockly.blockRendering.Types');
```
rachel-fenichel: Some thoughts
- It would be nice if we could find a way to make this name shorter, since we're using it explicitly everywhere.
- If this is a namespace, `types` should not be capitalized.

core/renderers/measurables/types.js
samelhusseini: An enum is just an object so yes. and it's not used anywhere right now but it's for custom (dynamic) rendering types. I already have a few cases for this, but in essence, this avoids polluting these types with element types that are specific to some renderers.
Instead they can use create / register a type by doing something like: 
``
this.type = Blockly.blockRendering.Types.getType('CUSTOM_ELEMENT_TYPE');
```
and then to check in their code if an element is of that type: 
```
elem.type & Blockly.blockRendering.Types.getType('CUSTOM_ELEMENT_TYPE')
```

core/renderers/thrasos/info.js
samelhusseini: I'm not sure, I left the code as is without any functional changes. I only added some checks for nullability where the closure compiler was complaining.

core/renderers/thrasos/info.js
samelhusseini: Hmm, looking closer at this. it looks like Blockly.blockRendering.InputConnection extends measurable but doesn't update its type. Instead it relies on isInput to determine if a measurable is of an input type. 
With this, we can have it set it's type to ``INPUT`` and then each subclass can extend the type to also include the specific type enum flag.
I like that solution better, can fix if that makes sense.

core/renderers/zelos/drawer.js
samelhusseini: I shouldn't have.. Will check and update all instances of this.

core/renderers/measurables/inputs.js
samelhusseini: It should always be |= since the base is 0 and it's okay to bitwise OR zero.

samelhusseini: It should only be an equal in the two base classes: 
- Measurable
- Row

core/renderers/zelos/drawer.js
samelhusseini: Fixed.

core/renderers/measurables/types.js
rachel-fenichel: SGTM

core/renderers/thrasos/info.js
rachel-fenichel: Yup, that's what I expected based on the pattern elsewhere.

Though InputConnection extends Connection, which extends Measurable, rather than extending Measurable directly.

core/renderers/measurables/types.js
samelhusseini: Types is a singleton class, similar to ``Blockly.utils.IdGenerator``, so I think capitalized makes sense. Open to ideas on shortening the name :)

rachel-fenichel: Okay.  maybe we'll find a better name than Blockly.blockRendering for the first half.

rachel-fenichel:
LGTM once you resolve a few conflicts.

core/renderers/thrasos/info.js
samelhusseini: Yeah so in this design an InputConnection is an INPUT and a CONNECTION. do you see any problem with that?

Issue #2952, opened 2019-09-03
Slight difference in statement input connection location


samelhusseini:
**Describe the bug**

There's a slight difference in location of a statement input's connection between the original renderer and the geras.

**To Reproduce**

Original:
<img width="830" alt="Screen Shot 2019-09-03 at 10 38 10 AM" src="https://user-images.githubusercontent.com/16690124/64195809-376e2800-ce37-11e9-8933-91c9959507ae.png">

Geras:
<img width="557" alt="Screen Shot 2019-09-03 at 10 38 53 AM" src="https://user-images.githubusercontent.com/16690124/64195816-3a691880-ce37-11e9-8be4-62545c7ed0e3.png">

**Expected behavior**

Notice the highlight is ever so slightly offset to the left.

samelhusseini:
Fixed.

PR #2953, opened 2019-09-03
Fix slight offset in statement input connection X location.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2952

### Proposed Changes

Add additional X coordinate space when positioning the statement input. Enough for the dark path and the light path (so it matches the original renderer). 

Use the DARK_PATH_OFFSET constants instead of 0.5 when positioning the statement input on the Y axis.

### Reason for Changes

Match original renderer.

### DIFF
--- a/core/renderers/common/drawer.js
+++ b/core/renderers/common/drawer.js
@@ -394,8 +394,9 @@ Blockly.blockRendering.Drawer.prototype.positionStatementInputConnection_ = func
     var connX = row.xPos + row.statementEdge + input.notchOffset;
     if (this.info_.RTL) {
       connX *= -1;
+    } else {
+      connX += this.constants_.DARK_PATH_OFFSET;
     }
-    connX += this.constants_.DARK_PATH_OFFSET;
     input.connection.setOffsetInBlock(connX,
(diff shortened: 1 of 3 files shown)


samelhusseini:
RTL has a number of issues: https://github.com/google/blockly/issues/2954, will address in this PR too.

samelhusseini:
Fixed https://github.com/google/blockly/issues/2954

core/renderers/common/drawer.js
```diff
@@ -394,8 +394,9 @@ Blockly.blockRendering.Drawer.prototype.positionStatementInputConnection_ = func
     var connX = row.xPos + row.statementEdge + input.notchOffset;
     if (this.info_.RTL) {
       connX *= -1;
+    } else {
```
rachel-fenichel: This works for now--longer term, I think I want to move this into the geras-specific drawer and get rid of the dark path offset for other renderers.

samelhusseini: I'm down with that plan!

PR #2955, opened 2019-09-03
Refactors modify and add tests

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Refactors modify and cleans up some TODOs. 

### Proposed Changes
The only change in functionality is if the workspace is marked and the cursor is on a next or an input connection it will move the source block to the workspace. Previously, it did not move the block and warned the user that this action was not allowed. We changed insert to be a bit more aggressive in trying to find a way to connect two blocks given a connection, so trying to do the same thing here.

### Test Coverage

Tested on:

### DIFF
--- a/core/keyboard_nav/navigation.js
+++ b/core/keyboard_nav/navigation.js
@@ -487,7 +487,7 @@ Blockly.navigation.moveBlockToWorkspace_ = function(block, wsNode) {
 Blockly.navigation.modify_ = function() {
   var markerNode = Blockly.navigation.marker_.getCurNode();
   var cursorNode = Blockly.navigation.cursor_.getCurNode();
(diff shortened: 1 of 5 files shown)


core/keyboard_nav/navigation.js
```diff
@@ -432,68 +433,82 @@ Blockly.navigation.modify = function() {
...
   }
+  return true;
+};
 
-  if (cursorType == Blockly.ASTNode.types.WORKSPACE) {
-    Blockly.navigation.warn('Cannot attach a workspace to anything else.');
+/**
+ * Disconnect the block from its parent and move to the position of the
+ * workspace node.
+ * @param {!Blockly.Block} block [description]
```
rachel-fenichel: Missing descriptions

PR #2959, opened 2019-09-03
Action list

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Keyboard navigation did not work in readonly mode. 

### Proposed Changes
If we are in read only mode and keyboard accessibility mode is turned on then we will allow a certain subset of actions that do not modify the workspace, but allow the user to move around the workspace.

### Test Coverage

Tested on:

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -216,9 +216,8 @@ Blockly.onKeyDown_ = function(e) {
 
   if (mainWorkspace.options.readOnly) {
     // When in read only mode handle key actions for keyboard navigation.
-    if (Blockly.keyboardAccessibilityMode &&
-      Blockly.navigation.onKeyPress(e)) {
-      return true;
+    if (Blockly.keyboardAccessibilityMode) {
+      Blockly.navigation.onKeyPress(e);
     }
     return;
   }
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -206,20 +206,23 @@ Blockly.svgResize = function(workspace) {
 Blockly.onKeyDown_ = function(e) {
   var mainWorkspace = Blockly.mainWorkspace;
 
-  if (mainWorkspace.options.readOnly &&
-      Blockly.keyboardAccessibilityMode &&
-      Blockly.navigation.onKeyPress(e)) {
-    return true;
-  }
-
-  if (mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e) ||
+  if (Blockly.utils.isTargetInput(e) ||
       (mainWorkspace.rendered && !mainWorkspace.isVisible())) {
-    // No key actions on readonly workspaces.
     // When focused on an HTML text input widget, don't trap any keys.
     // Ignore keypresses on rendered workspaces that have been explicitly
     // hidden.
     return;
   }
+
+  if (mainWorkspace.options.readOnly) {
+    // When in read only mode handle key actions for keyboard navigation.
+    if (Blockly.keyboardAccessibilityMode &&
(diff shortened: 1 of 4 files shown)


core/keyboard_nav/ast_node.js
```diff
@@ -288,7 +288,7 @@ Blockly.ASTNode.prototype.findPreviousEditableField_ = function(location,
   var previousField = null;
   var startIdx = opt_last ? fieldRow.length - 1 : fieldIdx - 1;
   for (var i = startIdx, field; field = fieldRow[i]; i--) {
-    if (field.isCurrentlyEditable()) {
```
RoboErikG: Why this change?

core/blockly.js
```diff
@@ -205,6 +205,13 @@ Blockly.svgResize = function(workspace) {
 // are multiple workspaces and non-main workspaces are able to accept input.
 Blockly.onKeyDown_ = function(e) {
   var mainWorkspace = Blockly.mainWorkspace;
+
+  if (mainWorkspace.options.readOnly &&
```
RoboErikG: I think you're conflicting a bit with the if statement immediate below. I'd actually move this below, take out the readOnly check in the other if so it reads like:

```
if (Blockly.utils.isTargetInput(e) || (...rendered && ... isVisible)) {
  ...
}
if (mainWorkspace.options.readOnly) {
  if (Blockly.keyboardAccessibilityMode) {
    Blockly.navigation.onKeyPress(e);
  }
  return; // Doesn't return a value
}
```

core/keyboard_nav/navigation.js
```diff
@@ -837,14 +837,19 @@ Blockly.navigation.onKeyPress = function(e) {
...
   var curNode = Blockly.navigation.cursor_.getCurNode();
+  var readOnly = Blockly.getMainWorkspace().options.readOnly;
   var actionHandled = false;
 
   if (action) {
-    if (curNode && curNode.getType() === Blockly.ASTNode.types.FIELD) {
-      actionHandled = curNode.getLocation().onBlocklyAction(action);
-    }
-    if (!actionHandled) {
+    if (readOnly && Blockly.navigation.READONLY_ACTION_LIST.indexOf(action) > -1) {
```
RoboErikG: If you swap these two it becomes
if (!readonly) {
  ...
} else if (Blockly.navigation.READONLY_ACTION_LIST.indexOf(action) > -1) {
  ...
}

core/keyboard_nav/ast_node.js
alschmiedt: Because isCurrentlyEditable will always be false for fields when in readOnly mode. So the user wouldn't be able to navigate to the fields.

core/blockly.js
alschmiedt: fixed

core/keyboard_nav/navigation.js
alschmiedt: fixed

core/blockly.js
```diff
@@ -206,20 +206,23 @@ Blockly.svgResize = function(workspace) {
...
     // Ignore keypresses on rendered workspaces that have been explicitly
     // hidden.
     return;
   }
+
+  if (mainWorkspace.options.readOnly) {
+    // When in read only mode handle key actions for keyboard navigation.
+    if (Blockly.keyboardAccessibilityMode &&
+      Blockly.navigation.onKeyPress(e)) {
+      return true;
```
RoboErikG: This method has no return. You can just check keyboardAccessibilityMode and then call onKeyPress without the return true.

PR #2969, opened 2019-09-04
Tweak top and bottom rows

### LINKED ISSUE #2866: StartY and overhangY in top and bottom rows
I added startY and overhangY properties in top and bottom rows in the new rendering to distinguish between the bounding box of a block and the "top edge" and "bottom edge".  The naming and use of the values is pretty clumsy, and I want to figure out what they're good for, annotate them fully, and be able to explain them to anyone else.

### LINKED ISSUE #2945: topRow.populate and bottomRow.populate belong in info
Everything else is populated in info, and Sam has made a good case for wanting to change what's in those rows by renderer type, which also points at putting population in info.

Argument against: mostly that there's already lots of code there.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves
#2866, which helps to fix #2945 

### Proposed Changes

Naming: 
 - overhangY -> descenderHeight (on bottom row)
 - startY -> ascenderHeight/capline (on top row)

Everything else:
 - getElemCenterline now does different things for top and bottom rows.  This might generalize into respecting ascender and descender heights, but I'm not certain.  This makes the debug rectangles render more accurately.
 - Rewrite top- and bottom-row measure to (hopefully) respect min height and min width

### Reason for Changes

- naming we agreed on
- render debug should be accurate
- height and width were often wrong

### Test Coverage
All three renderers work in the playground with a variety of blocks.

### Additional Information
Please scrutinize top row and bottom row measure to make sure min height and width are being respected.


core/renderers/measurables/row_elements.js
```diff
@@ -125,7 +125,7 @@ Blockly.blockRendering.RoundCorner = function(opt_position) {
   this.width = this.constants_.CORNER_RADIUS;
   // The rounded corner extends into the next row by 4 so we only take the
   // height that is aligned with this row.
-  this.height = this.constants_.NOTCH.height;
+  this.height = this.constants_.CORNER_RADIUS / 2;
```
samelhusseini: It would be nice if we could get to a point where the corners store their true height value. 
The corner extending past its row, that statement isn't true for zelos. What's our plan here? Is there value in having this report its real height if the notch height is greater?

rachel-fenichel: Yeah, that gets back to whether anyone other than a notch and hat get ascenders and descenders.

In either case, the notch height should have nothing to do with the corner height.  That was just a convenient way of getting a 4 in the past.

Issue #2956, opened 2019-09-03
Move DARK_PATH_OFFSET into Geras


rachel-fenichel:
The Geras renderer has a dark path, which impacts sizing and placement of connections, among other things, and is accounted for with the DARK_PATH_OFFSET constant.  Other renderers set this constant to zero, but it should just be removed for anyone other than Geras.

RoboErikG:
Could be punted. Low priority cleanup for this release.

PR #2971, opened 2019-09-05
Remove references to dark path offset; fix highlights while dragging

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #2956 
### Proposed Changes

Remove most references to dark path offset outside of geras.

It's still being used for inline inputs in thrasos, and in the root drawer.  thrasos may need to set it to zero, or (preferred) the root drawer needs to not refer to it at all (all references can be in info instead).

### Reason for Changes

Cleanup


samelhusseini:
Zelos constant sets the DARK_PATH_OFFSET to 0. Should we remove it from there as well?

rachel-fenichel:
Not yet, because it's still used in the InlineInput measurable.

PR #2973, opened 2019-09-05
Use strings instead of arrays for highlight paths

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves
Cleanup, toward defining the API for a renderer to know what paths it can change.

### Proposed Changes
Use string concatenation instead of array.push to build up the highlight strings.

### Reason for Changes
String concatentation is faster.  We're already doing this in the drawer.

### Test Coverage
Checked that all three renderers work.  Also rebuilt to add in some missing requires.


core/renderers/geras/highlighter.js
```diff
@@ -81,92 +79,95 @@ Blockly.geras.Highlighter = function(info, pathObject) {
       this.highlightConstants_.JAGGED_TEETH;
 };
 
+Blockly.geras.Highlighter.prototype.getSteps = function() {
```
samelhusseini: jsdoc?

core/renderers/geras/highlighter.js
```diff
@@ -81,92 +79,95 @@ Blockly.geras.Highlighter = function(info, pathObject) {
...
+  return this.steps_;
+};
+
+Blockly.geras.Highlighter.prototype.getInlineSteps = function() {
```
samelhusseini: jsdoc?

rachel-fenichel:
Yes, we're making some changes to PathObject (bigger changes) so that weirdness won't stick around.

PR #2976, opened 2019-09-06
Remove closure base file dependency

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Shave off **28KB** from Blockly compressed.

### Proposed Changes

Because we made use of a number of closure's base methods, we had a need to include closure's base file. Since we only use very few of these methods, we can trim down our binary size even further by providing alternatives for the following: 
- goog.inherits
- goog.exportSymbol
- goog.mixin
- goog.require
- goog.provide

``goog.exportSymbol`` and ``goog.mixin`` can be moved to Blockly.utils and re-implemented.

``goog.provide`` and ``goog.require`` are built in and the compiler knows how to handle those. We do need to stub those though, because our [message files](https://github.com/google/blockly/blob/master/msg/js/en.js#L7) include a require and provide call. 

``goog.inherits`` I've toyed with a few ideas of what to do with this one, but the least intrusive solution I came up with was to just inject an implementation of it.

@NeilFraser thoughts?

### Reason for Changes

Less closure (save 28KB)

### Test Coverage

Tested the demos.


core/utils.js
```diff
@@ -622,3 +622,37 @@ Blockly.utils.screenToWsCoordinates = function(ws, screenCoordinates) {
...
 };
+
+/**
+ * Builds an object structure for the provided namespace path, ensuring that
+ * names that already exist are not overwritten. For example:
+ * "a.b.c" -> a = {};a.b={};a.b.c={}
+ * @param {string} name name of the object that this file defines.
+ * @param {*=} opt_object the object to expose at the end of the path.
+ */
+Blockly.utils.exportSymbol = function(name, opt_object) {
```
NeilFraser: Find these comments in our codebase:
// Export symbols that would otherwise be renamed by Closure compiler.

And either replace them with exportSymbol, or delete this function and use the existing approach.  Whatever we do, it should be consistent.

build.py
```diff
@@ -210,6 +210,10 @@ def gen_core(self):
         ("warning_level", "DEFAULT"),
       ]
 
+    # Add Closure provide, require and inherits methods.
+    params.append(("js_code", """var goog=goog||{};goog.provide=function(){};goog.require=function(){};
+goog.inherits=function(a,b){function c(){}c.prototype=b.prototype;a.superClass_=b.prototype;a.prototype=new c;a.prototype.constructor=a;};"""))
```
NeilFraser: Neat idea, but this moves some of our core code into a very unexpected location, build.py.  And it means everyone's build process needs a similar hack added.

build.py
```diff
@@ -210,6 +210,10 @@ def gen_core(self):
         ("warning_level", "DEFAULT"),
       ]
 
+    # Add Closure provide, require and inherits methods.
+    params.append(("js_code", """var goog=goog||{};goog.provide=function(){};goog.require=function(){};
```
NeilFraser: I don't see any goog.provide or goog.require in the compressed files.  Why are these needed?

samelhusseini: It's there: https://github.com/google/blockly/blob/master/blockly_compressed.js#L7
We need it for our message files as they have provide/require calls eg: [here](https://github.com/google/blockly/blob/master/msg/js/en.js). and are included as is in html files eg: [here](https://github.com/google/blockly/blob/master/demos/fixed/index.html#L8)

samelhusseini: We also have calls to require/provide in our custom fields in demos, but I know we can get rid of those. I'm just not sure about the message files, does Blockly games need the call to require/provide in order to bundle message files?

core/utils.js
samelhusseini: Nice, done.

core/utils.js
```diff
@@ -622,3 +622,37 @@ Blockly.utils.screenToWsCoordinates = function(ws, screenCoordinates) {
...
+      cur = cur[part] = {};
+    }
+  }
+};
+
+
+/**
+ * Copies all the members of a source object to a target object.
+ * @param {Object} target Target.
+ * @param {Object} source Source.
```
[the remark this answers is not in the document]
samelhusseini: Done.

build.py
samelhusseini: I'm happy moving these to a `base.min.js` file or something so they're on their own and under `closure/goog/`, but I can't think of a way to have them defined in `core/` and not have to:
- reimplement provide/require (as we need it in there for the reasons outlined in the PR description)
- have a need to depend (ie: require) on a "base" file anywhere we do a `goog.inherit`.
If we think of this patch as a way to make our code work with the closure compiler, then the build script is the appropriate place to patch them in.
For example, we have similar code [here](https://github.com/google/blockly/blob/master/build.py#L255) and [here](https://github.com/google/blockly/blob/master/build.py#L283).

core/utils.js
```diff
@@ -622,3 +622,37 @@ Blockly.utils.screenToWsCoordinates = function(ws, screenCoordinates) {
   var finalOffsetMainWs = finalOffsetPixels.scale(1 / ws.scale);
   return finalOffsetMainWs;
 };
+
+/**
+ * Builds an object structure for the provided namespace path, ensuring that
+ * names that already exist are not overwritten. For example:
+ * "a.b.c" -> a = {};a.b={};a.b.c={}
+ * @param {string} name name of the object that this file defines.
+ * @param {*=} opt_object the object to expose at the end of the path.
```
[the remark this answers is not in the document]
samelhusseini: Fixed.

gulpfile.js
```diff
@@ -83,6 +83,14 @@ function prependHeader() {
   return gulp.insert.prepend(`// Do not edit this file; automatically generated by gulp.\n`);
 }
 
+/**
+ * Helper method for prepending Google closure methods used in Blockly.
```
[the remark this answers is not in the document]
samelhusseini: Fixed.

demos/custom-fields/pitch/field_pitch.js
```diff
@@ -29,6 +29,7 @@ goog.provide('CustomFields.FieldPitch');
 goog.require('Blockly.FieldTextInput');
 goog.require('Blockly.utils.math');
 
+var CustomFields = {};
```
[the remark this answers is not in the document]
samelhusseini: Fixed.

build.py
samelhusseini: Discussed offline, we'll go ahead and remove the require/provide calls in the built message files.

core/workspace.js
```diff
@@ -791,10 +791,11 @@ Blockly.Workspace.getAll = function() {
...
 
 // Export symbols that would otherwise be renamed by Closure compiler.
-Blockly.Workspace.prototype['clear'] = Blockly.Workspace.prototype.clear;
-Blockly.Workspace.prototype['clearUndo'] =
-    Blockly.Workspace.prototype.clearUndo;
-Blockly.Workspace.prototype['addChangeListener'] =
-    Blockly.Workspace.prototype.addChangeListener;
-Blockly.Workspace.prototype['removeChangeListener'] =
-    Blockly.Workspace.prototype.removeChangeListener;
+Blockly.utils.exportSymbol('Blockly.Workspace.prototype.clear',
```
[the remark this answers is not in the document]
samelhusseini: Discussed offline, will use @export instead.

build.py
samelhusseini: Going to go ahead with Blockly.utils.object.inherits....

samelhusseini:
@NeilFraser this is ready for a final review.

tests/compile/main.js
```diff
@@ -1,6 +1,6 @@
 goog.provide('Main');
-// Messages (in some language)
-goog.require('Blockly.Msg.en');
+// Messages
+goog.require('Blockly.Msg');
```
NeilFraser: Why is this needed?

samelhusseini: Yep, removed.

PR #2981, opened 2019-09-06
Renderers define which paths a block has.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

- Create a base class for a PathObject and a geras-specific implementation.
- Add that to each renderer
- Update the block_svg constructor to ask the renderer for a path object

### Reason for Changes

Renderers should be able to specify which paths they can expect to have on the block.

### Test Coverage

Loaded blocks with all three renderers in the playground.

### Additional Information

- @samelhusseini Let's chat about inheritance, typing, etc.  This does what I want, but it has the typing problems we discussed yesterday.
- Also, let's find a better name.
- The dark and light paths have to be kept around until we get a colourer sorted out that also takes in a pathObject
- All documentation (including `@fileoverview` and copyright headers) needs a pass once we have a design we're happy with.


samelhusseini:
Looking into this, the closure library doesn't have support for type checking constructors. You can set that each of those properties is a constructor, but subclasses overriding this won't be bound to any type restrictions on what they set that property to. 
Instead you can make this a factory method, and you already have that with the path object with ``makePathObject``. Override that in geras and you we can ensure the return type of these is consistent. 
Now for what the return type should be. The two PathObjects don't share a lot of commonality, except that they have the same constructor, but each of the ``setPaths`` takes a different number of arguments. That's okay, it just means the path object's shape is just any object that takes in a ``root`` svg element. We can add further restrictions. So what we need to do here is create an interface that each of the path objects ``@implements``, and use that interface as the return type for ``makePathObject``. 
```js
/**
 * @interface
 */
Blockly.blockRendering.IPathObject = function() {};


/**
 * @param {SVGElement} root Root element.
 * @constructor
 * @implements {Blockly.blockRendering.IPathObject}
 */
Blockly.blockRendering.PathObject = function(root) {
  this.svgRoot = root;
.... 

```

A couple of other tips. For files that you want to require only the interface use: 
```
goog.requireType('Blockly.blockRendering.IPathObject');
```
You will need to also ``goog.provide`` that interface in the file where it's defined.

samelhusseini:
@googlebot I consent.

core/renderers/common/path_object.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview TODO
```
RoboErikG: Missing fileoverview

core/renderers/common/path_object.js
```diff
@@ -0,0 +1,93 @@
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2016 Google Inc.
```
RoboErikG: Update copyright

core/renderers/common/path_object.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param {!SVGElement} root The root SVG element.
+ * @constructor
+ * @implements {Blockly.blockRendering.IPathObject}
+ * @package
+ */
+Blockly.blockRendering.PathObject = function(root) {
+  this.svgRoot = root;
+
+  /**
+   * @type {SVGElement}
```
RoboErikG: Add comment

core/renderers/common/path_object.js
```diff
@@ -0,0 +1,93 @@
...
+   * @package
+   */
+  this.svgPath = Blockly.utils.dom.createSvgElement('path',
+      {'class': 'blocklyPath'}, this.svgRoot);
+
+  // The light and dark paths need to exist (for now) because there is colouring
```
RoboErikG: Is this still true?

core/renderers/geras/path_object.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2016 Google Inc.
```
RoboErikG: Update license and fileoverview.

core/renderers/geras/path_object.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+Blockly.geras.PathObject = function(root) {
+  this.svgRoot = root;
+
+  // The order of creation for these next three matters, because that
+  // effectively sets their z-indices.
+
+  /**
+   * @type {SVGElement}
+   * @package
```
RoboErikG: @Override ?

core/renderers/common/path_object.js
samelhusseini: Yeah. We're still assume these paths exist in [block_svg::setBorderColor_](https://github.com/google/blockly/blob/develop/core/block_svg.js#L995)

core/renderers/geras/path_object.js
samelhusseini: path_object doesn't inherit from a single superclass file. Instead we use an interface, and so no need for @override.

samelhusseini:
This is ready for another look :)

PR #2983, opened 2019-09-08
Image Field Config

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

Adds an opt_config option that can be passed to the image field constructor. Supports flipRtl, alt, and tooltip.

Also fixed block factory.

### Reason for Changes

Working is good!

### Test Coverage

Added tests for:
 * onClick handler
 * Alt
 * FlipRtl

### Documentation

Updates [here ](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#creation)and [here](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#click_handler).

### Additional Information

### DIFF
--- a/core/field_image.js
+++ b/core/field_image.js
@@ -40,31 +40,68 @@ goog.require('Blockly.utils.Size');
  * @param {!(string|number)} width Width of the image.
  * @param {!(string|number)} height Height of the image.
  * @param {string=} opt_alt Optional alt text for when block is collapsed.
- * @param {Function=} opt_onClick Optional function to be called when the image
- *     is clicked.  If opt_onClick is defined, opt_alt must also be defined.
+ * @param {function(!Blockly.FieldImage)=} opt_onClick Optional function to be
+ *     called when the image is clicked. If opt_onClick is defined, opt_alt must
+ *     also be defined.
  * @param {boolean=} opt_flipRtl Whether to flip the icon in RTL.
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#creation}
+ *    for a list of properties this parameter supports.
  * @extends {Blockly.Field}
  * @constructor
  */
 Blockly.FieldImage = function(src, width, height,
-    opt_alt, opt_onClick, opt_flipRtl) {
-
+    opt_alt, opt_onClick, opt_flipRtl, opt_config) {
+  /* Return early. */
   if (!src) {
     throw Error('Src value of an image field is required');
   }
-
   if (isNaN(height) || isNaN(width)) {
     throw Error('Height and width values of an image field must cast to' +
       ' numbers.');
   }
-
-  // Ensure height and width are numbers.  Strings are bad at math.
   var imageHeight = Number(height);
   var imageWidth = Number(width);
   if (imageHeight <= 0 || imageWidth <= 0) {
     throw Error('Height and width values of an image field must be greater' +
       ' than 0.');
   }
+
+  /* Initialize configurable properties. */
+  /**
+   * Whether to flip this image in RTL.
+   * @type {boolean}
+   * @private
+   */
+  this.flipRtl_ = false;
+
+  /**
+   * Alt text of this image.
+   * @type {string}
+   * @private
+   */
+  this.altText_ = '';
+
+  Blockly.FieldImage.superClass_.constructor.call(
+      this, src || '', null, opt_config);
+
+  /* Give constructor parameters priority. */
+  if (opt_flipRtl != null) {
+    this.flipRtl_ = !!opt_flipRtl;
+  }
+  if (opt_alt != null) {
+    this.altText_ = opt_alt;
+  }
+
+  /* Initialize other properties. */
+  /**
+   * The size of the area rendered by the field.
+   * @type {Blockly.utils.Size}
+   * @protected
+   * @override
+   */
+  this.size_ = new Blockly.utils.Size(imageWidth,
+      imageHeight + Blockly.FieldImage.Y_PADDING);
 
   /**
    * Store the image height, since it is different from the field height.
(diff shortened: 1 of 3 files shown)


BeksOmega:
I think this one is good for a look @samelhusseini

core/field_image.js
```diff
@@ -40,60 +40,60 @@ goog.require('Blockly.utils.Size');
  * @param {!(string|number)} width Width of the image.
  * @param {!(string|number)} height Height of the image.
  * @param {string=} opt_alt Optional alt text for when block is collapsed.
- * @param {Function=} opt_onClick Optional function to be called when the image
- *     is clicked.  If opt_onClick is defined, opt_alt must also be defined.
+ * @param {function(!Blockly.FieldImage)=} opt_onClick Optional function to be
```
samelhusseini: Nice change, I'd keep Blockly.FieldImage as optional. ie: they don't have to define a function that takes in a FieldImage if they didn't want to. A function that we call when the the image is clicked would suffice: 
```
@param {function(Blockly.FieldImage)=}
```

core/field_image.js
```diff
@@ -40,60 +40,60 @@ goog.require('Blockly.utils.Size');
...
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/image#creation}
+ *    for a list of properties this parameter supports.
  * @extends {Blockly.Field}
  * @constructor
  */
 Blockly.FieldImage = function(src, width, height,
-    opt_alt, opt_onClick, opt_flipRtl) {
-
+    opt_alt, opt_onClick, opt_flipRtl, opt_config) {
+  /* Return early. */
```
samelhusseini: Why multi-line comment?

core/field_image.js
```diff
@@ -40,60 +40,60 @@ goog.require('Blockly.utils.Size');
...
-    this.clickHandler_ = opt_onClick;
-  }
+  this.altText_ = '';
 
   Blockly.FieldImage.superClass_.constructor.call(
-      this, src || '', null);
+      this, src || '', null, opt_config);
 
+  /* Give constructor parameters priority. */
+  if (opt_flipRtl != null) {
```
samelhusseini: Simply, 
```
if (opt_flipRtl)
```
as if it's false we don't need to re-set it.

core/field_image.js
```diff
@@ -40,60 +40,60 @@ goog.require('Blockly.utils.Size');
...
+    this.flipRtl_ = !!opt_flipRtl;
+  }
+  if (opt_alt != null) {
```
samelhusseini: Same as above.

core/field_image.js
BeksOmega: Should it be `{function(Blockly.FieldImage)=}` or `{function(!Blockly.FieldImage=)=}`? The former is definitely more readable hehe.

core/field_image.js
BeksOmega: They were kinda acting like organization headers, so I wanted them to stand out a bit. I'll switch it to normal comments.

core/field_image.js
BeksOmega: I was thinking if you had something like:

```
var defaultOptions = {
  alt: 'someAlt',
  flipRtl: true
}

// Override flipRtl to false.
var field  = new Blockly.FieldImage('src', 10, 10, null, null, false, defaultOptions);
```

You'd want the field's flipRtl value to be false. Similar thing with the alt field. (Admittedly, that's a pretty silly thing for people to do though.)

I'm cool with having the above result in true, or false, I just want confirmation first.

If we decide to have it result in true, I'll switch it and add tests to document (^-^)b

core/field_image.js
samelhusseini: Yep, you're right! ``{function(!Blockly.FieldImage=)=}`` is the right way to describe the above.
On second thoughts, in the interest of simplicity, maybe just keep it as ``{function(!Blockly.FieldImage)=}``.

core/field_image.js
samelhusseini: Ah I see. Here are some thoughts on how the two ways of configuring interface together. 
- Currently, in some cases where the config param is configuring something that used to be configured via a constructor parameter, it's getting configured twice (as we're doing the old way first, and then overriding it). 
- In the interest of avoiding that, and avoiding developers having to guess which will win, we should only support one way or the other. ie: if the opt_config property doesn't exist, go ahead and use the constructor parameters and configuring properties in the constructor "the old way". Otherwise, ``configure_`` will take care of configuring parameters. 

What are your thoughts on that? Do you see reasons for supporting both simultaneously?

BeksOmega: Sounds like a great plan! I'll update everything accordingly :D 

Just gotta make sure it gets added to the docs.

core/field_image.js
```diff
@@ -116,12 +131,14 @@ Blockly.utils.object.inherits(Blockly.FieldImage, Blockly.Field);
...
-      Number(Blockly.utils.replaceMessageReferences(options['height']));
+  var width = Number(Blockly.utils.replaceMessageReferences(
+      options['width']));
+  var height = Number(Blockly.utils.replaceMessageReferences(
+      options['height']));
   var alt = Blockly.utils.replaceMessageReferences(options['alt']);
   var flipRtl = !!options['flipRtl'];
-  return new Blockly.FieldImage(src, width, height, alt, null, flipRtl);
+  return new Blockly.FieldImage(src, width, height,
+      alt, null, flipRtl, options);
```
samelhusseini: There's something funny going on here. If you pass in the options property to the constructor you're essentially saying use the new configuration. So we don't "really" need to pass in alt, flipRtl as they're configurable arguments.
Also, interestingly alt here goes through replaceMessageReferences in this path, but it doesn't in the configure_ path, I think that's a bug.

BeksOmega: > There's something funny going on here. If you pass in the options property to the constructor you're essentially saying use the new configuration. So we don't "really" need to pass in alt, flipRtl as they're configurable arguments.

Yep missed that, I caught it on the number field though so that's good.

>Also, interestingly alt here goes through replaceMessageReferences in this path, but it doesn't in the configure_ path, I think that's a bug.

Good catch, can fix.

PR #2984, opened 2019-09-08
Label Field Config

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

Adds an opt_config parameter to the label field and serializable label field constructors. It accepts class and tooltip.

### Reason for Changes

### Test Coverage

Added tests for setting the class via js constructor, json, programmatic, etc. Tests run on label field and serializable label field.

### Documentation

Updates [here ](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation)and [here](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label-serializable#creation).

Label fields also need a customization section, since setClass is now a thing.

### Additional Information

### DIFF
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -53,16 +53,16 @@ Blockly.FieldLabel = function(opt_value, opt_class, opt_config) {
 
   /**
    * The html class name to use for this field.
-   * @type {string}
+   * @type {?string}
    * @private
    */
-  this.class_ = '';
+  this.class_ = null;
 
   Blockly.FieldLabel.superClass_.constructor.call(
       this, opt_value, null, opt_config);
 
   if (!opt_config) {  // If the config was not passed use old configuration.
-    this.class_ = opt_class || '';
+    this.class_ = opt_class || null;
   }
 
   /**
@@ -101,7 +101,7 @@ Blockly.FieldLabel.prototype.EDITABLE = false;
  */
 Blockly.FieldLabel.prototype.configure_ = function(config) {
   Blockly.FieldLabel.superClass_.configure_.call(this, config);
-  this.class_ = config['class'] || '';
+  this.class_ = config['class'] || null;
 };
 
 /**
--- a/core/field_label.js
+++ b/core/field_label.js
@@ -61,7 +61,7 @@ Blockly.FieldLabel = function(opt_value, opt_class, opt_config) {
       this, opt_value, null, opt_config);
 
   if (!opt_config) {  // If the config was not passed use old configuration.
-    this.class_ = opt_class || null;
+    this.class_ = opt_class;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


core/field_label.js
```diff
@@ -40,22 +40,30 @@ goog.require('Blockly.utils.Size');
...
+  }  // Else it gets cast to a string (e.g. false -> 'false').
+
   /**
    * The html class name to use for this field.
-   * @type {?string}
+   * @type {string}
    * @private
    */
-  this.class_ = opt_class || null;
+  this.class_ = '';
```
samelhusseini: Why not keep this as a nullable string?
That way we don't write an empty class to the dom element.

samelhusseini: ie: set it to ``null`` here.

samelhusseini: Oh nevermind, your check won't write a class if it's empty.
Still I don't see any reason to make it an empty string.

core/field_label.js
```diff
@@ -40,22 +40,30 @@ goog.require('Blockly.utils.Size');
...
  * @param {string=} opt_class Optional CSS class for the field's text.
+ * @param {Object=} opt_config A map of options used to configure the field.
+ *    See the [field creation documentation]{@link https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/label#creation}
+ *    for a list of properties this parameter supports.
  * @extends {Blockly.Field}
  * @constructor
  */
-Blockly.FieldLabel = function(opt_value, opt_class) {
+Blockly.FieldLabel = function(opt_value, opt_class, opt_config) {
+  if (opt_value == null) {
```
samelhusseini: same thing here, if we can keep it close to the superclass constructor call.

core/field_label.js
```diff
@@ -88,6 +96,14 @@ Blockly.FieldLabel.fromJson = function(options) {
  */
 Blockly.FieldLabel.prototype.EDITABLE = false;
 
+/**
+ * @override
+ */
+Blockly.FieldLabel.prototype.configure_ = function(config) {
+  Blockly.FieldLabel.superClass_.configure_.call(this, config);
+  this.class_ = config['class'] || null;
```
samelhusseini: Theoretically you can set a class name to `0`?

BeksOmega: Theoretically, yes.

`utils.dom.addClass` can handle that through implicit casts (tested), but we can caste explicitly or throw errors if you want.

samelhusseini: Yes, so I think we just remove the null fallback here, because passing config['class'] currently would set ``this.class_`` to null.

BeksOmega: Ohhh sorry I thought you were pointing out something about typing, fixed.

PR #2985, opened 2019-09-08
Number Field Configuration 

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

Adds an opt_config parameter to the number field construct. Accepts min, max, precision, and tooltip.

Also added a getConstraints function.

### Reason for Changes

So people can change individual constraints at runtime without having to store the constraints separately.

### Test Coverage

Added tests for setting:
* Min
* Max
* And precision

### Documentation

Updates [here](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/number#creation).

### Additional Information

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -138,19 +138,6 @@ Blockly.FieldNumber.prototype.setConstraints = function(min, max, precision) {
 };
 
 /**
- * Returns the value constraints on this number field.
- * @return {{min: number, max: number, precision: number}} The constraints
- *    on the number field.
- */
-Blockly.FieldNumber.prototype.getConstraints = function() {
-  return {
-    min: this.min_,
-    max: this.max_,
-    precision: this.precision_
-  };
-};
-
-/**
  * Sets the minimum value this field can contain. Updates the value to reflect.
  * @param {number|string|undefined} min Minimum value.
  */
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -28,6 +28,7 @@ goog.provide('Blockly.FieldNumber');
 
 goog.require('Blockly.fieldRegistry');
 goog.require('Blockly.FieldTextInput');
+goog.require('Blockly.utils.object');
 
 
 /**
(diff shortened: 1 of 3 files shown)


core/field_number.js
```diff
@@ -41,35 +41,46 @@ goog.require('Blockly.utils.object');
...
    * @type {number}
    * @protected
    */
   this.precision_ = 0;
 
+  /**
+   * The number of decimal places to allow, or null to allow any number of
+   * decimal digits.
+   * @type {?number}
+   * @protected
```
samelhusseini: I think this one is private.

samelhusseini:
@googlebot I concent

core/field_number.js
```diff
@@ -41,35 +41,46 @@ goog.require('Blockly.utils.object');
...
+   */
+  this.decimalPlaces_ = null;
+
   Blockly.FieldNumber.superClass_.constructor.call(
-      this, opt_value || 0, opt_validator);
+      this, opt_value || 0, opt_validator, opt_config);
 
   this.setConstraints(opt_min, opt_max, opt_precision);
```
BeksOmega: Do you think we can have this call the "internal" methods instead?

I'd hope that the value developers are passing to the constructor is within their constraints.

core/field_number.js
```diff
@@ -108,25 +131,132 @@ Blockly.FieldNumber.prototype.SERIALIZABLE = true;
...
+  this.setPrecisionInternal_(precision);
+  this.setValue(this.getValue());
+};
+
+/**
+ * Returns the value constraints on this number field.
+ * @return {{min: number, max: number, precision: number}} The constraints
+ *    on the number field.
+ */
+Blockly.FieldNumber.prototype.getConstraints = function() {
```
samelhusseini: Can we get rid of this now?

core/field_number.js
samelhusseini: I think this is fine.

BeksOmega: Sounds good, it does mean setValue gets called twice when the field is constructed though.

PR #2993, opened 2019-09-09
Fix npm js generator path in npm readme.

samelhusseini:
Update npm readme to use correct path for the Javascript generator.


samelhusseini:
Thanks!

Issue #2870, opened 2019-08-19
Convert all JSunit tests to mocha


samelhusseini:
The jsunit tests still depend on closure for a bunch of mock functionality. As we move away from closure, we can re-implement that functionality. Instead we can move away from jsunit altogether, and adopt mocha tests for all the coverage in jsunit. 

Conversion between the two should be straightforward. We just need to make sure we don't lose test coverage. Using this issue to track this conversion: 

- [x] block_test.js
- [x] connection_db_test.js https://github.com/google/blockly/pull/2872
- [x] connection_test.js https://github.com/google/blockly/commit/1baf7d1dcf6f8e54f30b90dc6f17c061e3a7e862#diff-e69b0acbbba2a2b3df6e74941d868499
- [x] event_test.js
- [x] extensions_test.js
- [x] generator_test.js
- [x] gesture_test.js https://github.com/google/blockly/pull/2875
- [x] json_test.js
- [x] metrics_test.js https://github.com/google/blockly/pull/2871
- [x] names_test.js https://github.com/google/blockly/pull/2873
- [x] theme_test.js https://github.com/google/blockly/pull/2874
- [x] utils_dom_test.js
- [x] utils_math_test.js
- [x] utils_string_test.js
- [x] utils_test.js
- [x] variable_map_test.js
- [x] variable_model_test.js
- [x] variables_test.js
- [x] widget_div_test.js
- [x] workspace_comment_test.js
- [x] workspace_test.js
- [x] workspace_undo_redo_test.js
- [x] tests/workspace_svg/procedure_svg.js
- [x] tests/workspace_svg/event_svg.js
- [x] tests/workspace_svg/workspace_svg_test.js
- [x] tests/blocks/logic_ternary_test.js

moniika:
We should also review the jsunit tests under tests/workspace_svg/ and either convert to mocha or remove.

PR #2994, opened 2019-09-10
Combined connection's inD_ and hidden_ properties into a single tracked_ property

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A
[EDIT also #2870]

### Proposed Changes

* Combines Connection's inDb_ and hidden_ properties into a single tracked_ property.
* Removes all database related functions from Connection (they now live on RenderedConnection).
* Removes all database related functions from Block (they now live on BlockSvg).
* Adds explicit functions for making the BlockSvg wait before asking its connections to add themselves to the database (used when deserializing from XML).
* Renames some private ConnectionDB functions.

### Reason for Changes

* These meant /slightly/ different things before, but they were similar enough that it seemed weird to have two properties. It also seemed confusing to have a property named 'hidden_' that didn't always reflect the visibility of the connection (since it had to do with database tracking, not visibility).
* The database is only used in rendered mode.
* The database is only used in rendered mode.
* I had to rename these functions anyway to get rid of references to 'hidden'. I thought I might as well make them more descriptive.
* Makes the names more descriptive of what the functions are doing.

### Test Coverage

* Reorganized the connection db tests a bit (I got permission from Sam first).
* Moved connection tests to mocha:
  * Can connect with reason.
    * Target Null.
    * Target self.
    * Different workspaces.
    * Type (previous, next, output, input) match/mismatch.
    * Shadows.
  * Type (string, number, etc) checks.
* Added block tests for:
  * Removing inputs.
  * Tracking connections.
  * Managing icons (when collapsing and things).

### Documentation

Nada.

### Additional Information

It's maybe best to review this commit by commit (although there are a lot of commits). That way it's easier to understand my thought process. And easier to make sure I didn't forget to move something hehe.

I also understand this is a bit of a large PR, so don't feel the need to look at it any time soon.


RoboErikG:
Rachel's out until next Wednesday, but I think she's the best person to take a look at this.

rachel-fenichel:
Notes from my reading so far:
-  `Input.setVisible` still updates whether an input's connections are being tracked
- `startTrackingAll` is on a `RenderedConnection`, but `startTrackingConnections` is on `BlockSvg`.
- There's a `waitToTrackConnections` on a block, and a `startTrackingConnections`, but no `stopTrackingConnections`.  That was unexpected.
- The db is no longer responsible for updating the connections--instead, rendered connections are responsible for updating the DB.  +1 for clarifying who is in charge.

Overall, I expect that a disable function would have a matching enable function, and they're not matching up the way I expect.  I think `startTrackingAll` and `stopTrackingAll` on `RenderedConnection` should just call `this.setTracking(true)` and then `this.targetBlock.startTrackingAll()` or something equivalent.

BeksOmega:
> * Input.setVisible still updates whether an input's connections are being tracked

Yes

> * `startTrackingAll` is on a `RenderedConnection`, but `startTrackingConnections` is on `BlockSvg`.
> * There's a `waitToTrackConnections` on a block, and a `startTrackingConnections`, but no `stopTrackingConnections`.  That was unexpected.
> 
> Overall, I expect that a disable function would have a matching enable function, and they're not matching up the way I expect. I think `startTrackingAll` and `stopTrackingAll` on `RenderedConnection` should just call `this.setTracking(true)` and then `this.targetBlock.startTrackingAll()` or something equivalent.

So this confusion arose because XML accessed the wait/start blockSvg functions, while inputs would access the start/stop connection functions.

In the context of XML it would be really weird to stop tracking connections, which is why blockSvg didn't get a stop function.

But I think you're right that the two can be unified, I'm gonna give it a shot!

BeksOmega:
So I'm having a bit of internal confict about this. On one hand I want to fix collapsing properly, and on the other hand I don't want this PR to get any crazier than it already is.

Unifiying `startTrackingConnections`,  `startTrackingAll` and `stopTrackingAll` turned out to be trickier than I thought because `startTrackingAll` and `stopTrackingAll` are handling some odd parts of the collapsing logic.

`stopTrackingAll` handles closing icons (like comment and mutator bubbles), which really belongs at the block level.

and `startTrackingAll` handles populating the render list with "leaf" blocks. Something that `startTrackingConnections` doesn't need to do, because in that case the blocks are already rendered.

Optimally I would want to at least add a block-level `setVisible`/`setCollapsedInternal` method, and possibly reorganize more of the collapse pipeline.

**In Conclusion I can either:**
A) Shuffle things out of `startTrackingAll` and `stopTrackingAll` and into other places temporarily. But to warn you, I would want to go back and rewrite everything in another PR (cuz it's gonna look gross).
B) Fix collapsing properly in this PR, which may make it bigger, and also possibly harder to review.
C) Fix collapsing in a follow up PR.

rachel-fenichel:
Would it help if `startTrackingAll` returned a render list of leaf blocks that went unused in some cases?

I choose C from those options.  I like the movement of database related functions to clarify that they don't matter for non-rendered connections and blocks.

BeksOmega:
> Would it help if `startTrackingAll` returned a render list of leaf blocks that went unused in some cases?

Yeah *sigh*,  I think you're right, that's probably what it'll end up being. I just don't like it >.<

BeksOmega:
Thinking about it, I could probably split this into a few different PRs without too much trouble. It's pretty well split up by commit already.

Maybe like:
* Tests.
* Move database stuff out of rendered.
* Move over to 'tracked' naming.

@rachel-fenichel ?

rachel-fenichel:
Yes please!

BeksOmega:
Gonna put up as separate PRs this afternoon.

PR #3001, opened 2019-09-11
Fix travis by running browser tests in headless mode

samelhusseini:
##  The basics

- [x] I branched from develop

### Resolves

Attempt to Fix travis

### Proposed Changes

Attempt to Fix travis

### Reason for Changes

Attempt to Fix travis

### Test Coverage

Let's see if travis works.


RoboErikG:
Looks like stable is grabbing an old version of Chrome for some reason. tests are timing out because it's older than Chrome 69

RoboErikG:
Could also be an issue with dpkg https://askubuntu.com/questions/1065231/dpkg-deb-error-archive-has-premature-member-control-tar-xz-before-contr/1100361

RoboErikG:
This looks like the issue: https://github.com/travis-ci/travis-ci/issues/9361

samelhusseini:
Should we just switch distro until the issue is fixed?

samelhusseini:
## Here's a summary of what happened: (I think)

Travis was initially complaining about not being able to acquire google-chrome-stable from apt-get version 77. A public key not valid error. After some research we found [this](https://travis-ci.community/t/travis-is-downloading-older-chrome-versions-in-place-of-chrome-stable/5040) hinting to the issue being an issue with the dist we were using (trust). We switched to xenial, and that got us around that issue. We then found out that there was a mismatch between the chrome driver's supported version and the version we were acquiring. 
apt-get google-chrome-stable was getting us a later version (73) than what the version of chrome-driver we were using (2.44).
After upgrading our chrome driver to 77.0.3865.40 which supports Chrome 77, we figured out that google-chrome-stable on the xenial dist was only getting us Chrome 73.
I then switched to acquiring chrome using the travis addon which got us version 77. 

At this point we had compatible chrome driver + browser combination. We were still hitting an error ``DevToolsActivePort file doesn't exist.``. Further detail on this issue [here](https://github.com/heroku/heroku-buildpack-google-chrome/issues/46). 
I still don't know why that issue is there, perhaps so far we've avoided this as we were using an old version of the driver, but we had to configure the browsers to run in headless mode on travis to workaround the issue.

I added a new env variable TRAVIS_CI, and used it to determine when to run the tests in headless mode. (so we can locally test them without them running in headless mode).

## Updated selenium + driver acquisition:
We were acquiring selenium, chrome and gecko drivers manually using scripts under ``test/scripts``. Instead we're able to use the CLI npm tool ``selenium-standalone``. This makes configuration of selenium and driver versions cleaner and in the one file ``tests/scripts/selenium-config.js``.

Issue #2828, opened 2019-08-14
Blocks with outputs misaligned in flyout


rachel-fenichel:
**Describe the bug**
Blocks with output tabs are supposed to be moved left so that the blocks align based on the visible "left edge" of the block.  New rendering messed this up.

![image](https://user-images.githubusercontent.com/13686399/63062718-ba423980-beae-11e9-9219-572cebd0b458.png)

**Expected behavior**

![image](https://user-images.githubusercontent.com/13686399/63062696-abf41d80-beae-11e9-9d2a-00fb46eaaab6.png)

**Additional context**

There's a question here about whether I should offset the block by the connection width or by some startX value that can be independent from the connection width.

Something similar holds true for hats.

RoboErikG:
https://github.com/google/blockly/issues/2800 is related to your comment about hats.

rachel-fenichel:
#2800 is fixed by my current code, but I don't want to close it until I finish this up.

moniika:
Fixed

PR #3003, opened 2019-09-11
Fixing x positioning of blocks with output connections in vertical fl…

moniika:
## The basics

- [x] I branched from develop

### Resolves

#2828 

### Proposed Changes

1. Changes constant used for output connection tab width from constant in BlocklySvg to constant from ConstantProvider which is used in current rendering for output tab width.

2. Updated layout calculations for blocks with output connections correctly compute the edge after the output tab as the edge for alignment in flyouts.

#### Before:
old
#### After:
new

3. Increased x offset of first block in horizontal flyout in RTL (to match offset in LTR).

4. Added missing goog.require calls in touched files

### Reason for Changes

1. This is the correct constant for tab width and would reflect appropriate changes if the tab width was changed.

2. The bounds and width definition for blocks with output has changed with new rendering changes to now include output tab which unintentionally changed how the blocks were now aligned in flyout. This improves the layout of blocks and is consistent with old behavior.

3. This improves the layout and although is not consistent with old behavior, makes the offset of the first block in the horizontal flyout consistent between LTR and RTL.

#### horizontal LTR layout (unchanged):
newLTR

#### horizontal RTL layout new:
newRTL

#### horizontal RTL layout old:
oldRTL

4. Cleanup

### Test Coverage

Tested on playground in all combinations of Start/End/Top/Bottom toolbox and LTR/RTL.

Tested on:
* Desktop Chrome 

### Additional Information

This change supports a constant output tab width among blocks with output tabs. If we were to support a similar alignment for blocks with varying output tab widths, then the following changes would need to occur:
1. Instead of offsetting blocks by TAB_WIDTH, the maximum width among output tabs would need to be used.
2. Instead of shifting blocks with output tabs by TAB_WIDTH, they would need to be shifted by their respective tab width. This value could either by added to the BlockSvg object, as an offset value, in Blockly.blockRendering.Drawer.prototype.recordSizeOnBlock_ since this information is available on the RenderInfo object.


core/flyout_base.js
```diff
@@ -32,8 +32,10 @@ goog.require('Blockly.Events.BlockCreate');
 goog.require('Blockly.Events.VarCreate');
 goog.require('Blockly.FlyoutButton');
 goog.require('Blockly.Gesture');
+goog.require('Blockly.Scrollbar');
 goog.require('Blockly.Tooltip');
 goog.require('Blockly.Touch');
+goog.require('Blockly.blockRendering');
```
RoboErikG: Move this up. Should be alphabetical ignoring case.

core/flyout_horizontal.js
```diff
@@ -268,17 +270,17 @@ Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
...
   var margin = this.MARGIN;
-  var cursorX = this.RTL ? margin : margin + Blockly.BlockSvg.TAB_WIDTH;
+  var cursorX = margin + this.tabWidth_;
   var cursorY = margin;
   if (this.RTL) {
     contents = contents.reverse();
   }
 
-  for (var i = 0, item; item = contents[i]; i++) {
+  for (var i = 0, item; (item = contents[i]); i++) {
```
RoboErikG: Why did you add these parenthesis here and elsewhere?

core/flyout_horizontal.js
```diff
@@ -289,7 +291,7 @@ Blockly.HorizontalFlyout.prototype.layout_ = function(contents, gaps) {
       var blockHW = block.getHeightWidth();
 
       // Figure out where to place the block.
-      var tab = block.outputConnection ? Blockly.BlockSvg.TAB_WIDTH : 0;
+      var tab = block.outputConnection ? this.tabWidth_ : 0;
```
RoboErikG: I think this is backwards. 
In RTL the margin on the left side is bigger for blocks with outputs compared to non-outputs
image
image


In both cases, we're missing margin on the right side when in rtl:
image

core/flyout_horizontal.js
```diff
@@ -29,6 +29,8 @@ goog.provide('Blockly.HorizontalFlyout');
 goog.require('Blockly.Block');
 goog.require('Blockly.Flyout');
 goog.require('Blockly.FlyoutButton');
+goog.require('Blockly.Scrollbar');
+goog.require('Blockly.WidgetDiv');
```
RoboErikG: WidgetDiv below utils

core/flyout_vertical.js
```diff
@@ -29,6 +29,8 @@ goog.provide('Blockly.VerticalFlyout');
 goog.require('Blockly.Block');
 goog.require('Blockly.Flyout');
 goog.require('Blockly.FlyoutButton');
+goog.require('Blockly.Scrollbar');
+goog.require('Blockly.WidgetDiv');
```
RoboErikG: ditto

core/flyout_horizontal.js
RoboErikG: Sorry, first two images are LTR. The right side margin is missing in both RTL and LTR.

moniika: I think you are right. I'll fix it since it's likely the intended behavior.

core/flyout_horizontal.js
moniika: I added this updates to layout_ and reflowInternal_ because I was already looking very closely in those functions.
Adding the parenthesis prevents this line from being flagged as a warning for some debuggers and IDEs. I did a similar change in the renderers directory in #2883 where I also included some detail in the PR about the warning.

core/flyout_base.js
moniika: Updated.

PR #3009, opened 2019-09-12
Update APIs and Add New Cursor Look

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
**Navigation / Navigation Test**
  Change most of  the navigation apis to be private
**Ast Node**
  Remove the parent Input for the input node. We now have a function that can get the input off of 
  the connection so we weren't using this function anymore.
**Cursor_Svg**
  Change the cursor look 
  Update the setParent method so that when we move the svg group of the cursor to the block, workspace, or field that object is in charge of positioning the cursor svg group.
**Cursor**
  Change the logic to match the new cursor look 

### Test Coverage

Tested on:


alschmiedt:
Ok fixed the relevant closure complaints, but will open another pr for the remaining ones.

core/workspace.js
```diff
@@ -120,16 +120,14 @@ Blockly.Workspace = function(opt_options) {
...
-  this.cursor_ = null;
+  this.cursor = null;
 
   /**
    * The marker used to mark a location for keyboard navigation.
    * @type {Blockly.MarkerCursor}
-   * @private
    */
-  this.marker_ = null;
+  this.marker = null;
```
samelhusseini: Why no longer private?

core/field.js
```diff
@@ -906,3 +923,39 @@ Blockly.Field.prototype.getParentInput = function() {
...
+ * @package
+ */
+Blockly.Field.prototype.setMarkerSvg = function(markerSvg) {
+  if (!markerSvg) {
+    this.markerSvg_ = null;
+    return;
+  }
+
+  if (this.fieldGroup_) {
+    this.fieldGroup_.appendChild(markerSvg);
```
samelhusseini: This check here hints to me that this can be called before initialization. In which case, I don't see a code path in initialization that appends the ``this.markerSvg_`` to the field group during initialization.

alschmiedt: Hmm, yeah I don't think we actually need this or the check in the block_svg. Both should already be initialized by the time the user starts moving around.

Issue #2702, opened 2019-07-24
Decide how developers will choose their renderer


rachel-fenichel:
Things to think about:
- at inject vs compile time
- registry of renderers, as with fields?

Issue #3012, opened 2019-09-13
Update documentation with how to configure a renderer


samelhusseini:
**Where**
A new page under Configure Blockly.

**What**

How to choose which renderer to use.

rachel-fenichel:
Added to the description of the configuration struct.

PR #3013, opened 2019-09-13
Support for registering renderers and switching between them

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2702

### Proposed Changes

Added support for registering multiple renderers. I went the way of fields by having a renderer map and each renderer registers their existence on load so those that compile Blockly with a custom entry point are able to pick and choose which renderers they want to include. 

Added a dropdown in the playground to switch between them.

I also made sure that the block_rendering was closure compiler compliant.

### Reason for Changes

No way to switch between renderers.

### Test Coverage

Tested the playground
Closure compiler
Custom compilation without all the renderers.
Demos.

Tested on:
* Desktop Chrome.

### Documentation

New workspace option to choose the renderer, once the dust settles we should update the documentation. Filed an issue to track. https://github.com/google/blockly/issues/3012


core/renderers/common/block_rendering.js
```diff
@@ -30,20 +30,79 @@
...
+  /**
+   * @constructor
+   * @extends {Blockly.blockRendering.Renderer}
+   */
+  var rendererCtor = function() {
+    rendererCtor.superClass_.constructor.call(this);
+  };
+  Blockly.utils.object.inherits(rendererCtor,
+      Blockly.blockRendering.rendererMap_[name]);
+  Blockly.blockRendering.renderer = new rendererCtor();
```
RoboErikG: missing an _ on renderer_

core/renderers/common/block_rendering.js
```diff
@@ -57,10 +116,18 @@ Blockly.blockRendering.render = function(block) {
...
 };
 
-Blockly.blockRendering.getConstants = function() {
-  return Blockly.blockRendering.renderer.constants;
+/**
+ * Get the current renderer.
+ * @return {Blockly.blockRendering.Renderer} The current renderer.
+ */
+Blockly.blockRendering.getRenderer = function() {
+  return Blockly.blockRendering.renderer;
```
RoboErikG: missing the _

core/renderers/common/block_rendering.js
```diff
@@ -57,10 +116,18 @@ Blockly.blockRendering.render = function(block) {
...
 };
 
-Blockly.blockRendering.getHighlightConstants = function() {
-  return Blockly.blockRendering.renderer.highlightConstants;
+/**
+ * Get the current renderer's constant provider.
+ * @return {Blockly.blockRendering.ConstantProvider} The constant provider.
+ */
+Blockly.blockRendering.getConstants = function() {
+  return Blockly.blockRendering.renderer.constants;
```
RoboErikG: here too

core/renderers/geras/renderer.js
```diff
@@ -59,3 +60,16 @@ Blockly.geras.Renderer.prototype.init = function() {
...
+
+/**
+ * Initialize the renderer.  Geras has a highlight provider in addition to
+ * the normal constant provider.
+ * @return {Blockly.geras.HighlightConstantProvider} The highlight constant
+ *     provider.
+ * @package
+ */
+Blockly.geras.Renderer.prototype.getHighlightConstants = function() {
+  return Blockly.blockRendering.renderer.highlightConstants;
```
RoboErikG: should probably use this.highlightConstants instead of getting the current renderer which either is this one or may be incompatible.

core/renderers/geras/renderer.js
```diff
@@ -59,3 +60,16 @@ Blockly.geras.Renderer.prototype.init = function() {
   Blockly.geras.Renderer.superClass_.init.call(this);
   this.highlightConstants = new Blockly.geras.HighlightConstantProvider();
 };
+
+/**
+ * Initialize the renderer.  Geras has a highlight provider in addition to
```
RoboErikG: This isn't initializing. Clean up comment.

core/workspace_svg.js
```diff
@@ -126,7 +126,7 @@ Blockly.WorkspaceSvg = function(options,
         Blockly.Procedures.flyoutCategory);
   }
 
-  Blockly.blockRendering.init();
+  Blockly.blockRendering.init(this.options.renderer || 'geras');
```
RoboErikG: Should this happen in inject instead since the rendering is shared by the workspace, the toolbox flyout, and the trash flyout and this will cause it to be configured three times? Alternately, init could return a renderer that gets set on the workspace_svg. That would allow rendering blocks in the toolbox  differently than on the workspace which some developers do...

samelhusseini: Discussed offline, will tackle this in a separate PR, but agreed to move away from a global singleton rendering instance.

samelhusseini:
Resolved PR comments, ready for final review.

PR #3017, opened 2019-09-13
Adding checkboxes for debug rendering options to playground.

### LINKED ISSUE #2904: Debug rendering isn't easy to turn on and off
You currently have to comment/uncomment it in code.

I was using css to turn it on and off, but then if you didn't have the css you just got black boxes (and it was rendering lots of extra rects all the time for no reason).

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2904 

### Proposed Changes

Adds checkboxes to playground for toggling debug rendering options that are dynamically generated (based on defined config object on debugger) and hidden when debug rendering is disabled. The debug rendering options persist in session storage and changing any debug render option triggers re-render.
In this change, the debug config object is changed to be shared across all debuggers, rather than each instance of the debug renderer for each block storing its own object. This simplified logic, but no longer allows for different instances of the debug renderer to have different options (which is expected to be a rare case).

#### Debug rendering disabled
debugDisabled

#### Debug rendering enabled with options
debugEnabled

### Reason for Changes

Improves debugging experience.

### Test Coverage

Tested on playground and checked without session storage for options and with options saved.

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/renderers/common/debugger.js
+++ b/core/renderers/common/debugger.js
@@ -54,24 +54,20 @@ Blockly.blockRendering.Debug = function() {
    * @type {SVGElement}
    */
   this.svgRoot_ = null;
-
-  /**
-   * @type {Object} Configuration object containing booleans to enable and
-   *     disable debug rendering of specific rendering components.
-   */
-  this.config_ = Blockly.blockRendering.Debug.getDebugConfig();
 };
 
-Blockly.blockRendering.Debug.getDebugConfig = function() {
-  return {
-    // rowSpacers: true,
-    // elemSpacers: true,
-    // rows: true,
-    elems: true,
-    // connections: true,
-    blockBounds: true,
-    // connectedBlockBounds: true
-  };
+/**
+ * @type {!Object.<string, boolean>} Configuration object containing booleans
+ *    to enable and disable debug rendering of specific rendering components.
+ */
+Blockly.blockRendering.Debug.config = {
+  rowSpacers: true,
+  elemSpacers: true,
+  rows: true,
+  elems: true,
+  connections: true,
+  blockBounds: true,
+  connectedBlockBounds: true
 };
 
 /**
@@ -94,7 +90,7 @@ Blockly.blockRendering.Debug.prototype.clearElems = function() {
  * @package
  */
 Blockly.blockRendering.Debug.prototype.drawSpacerRow = function(row, cursorY, isRtl) {
-  if (!this.config_.rowSpacers) {
+  if (!Blockly.blockRendering.Debug.config.rowSpacers) {
     return;
   }
 
@@ -121,7 +117,7 @@ Blockly.blockRendering.Debug.prototype.drawSpacerRow = function(row, cursorY, is
  * @package
  */
 Blockly.blockRendering.Debug.prototype.drawSpacerElem = function(elem, rowHeight, isRtl) {
-  if (!this.config_.elemSpacers) {
+  if (!Blockly.blockRendering.Debug.config.elemSpacers) {
     return;
   }
 
@@ -153,7 +149,7 @@ Blockly.blockRendering.Debug.prototype.drawSpacerElem = function(elem, rowHeight
  * @package
  */
 Blockly.blockRendering.Debug.prototype.drawRenderedElem = function(elem, isRtl) {
-  if (this.config_.elems) {
+  if (Blockly.blockRendering.Debug.config.elems) {
     var xPos = elem.xPos;
     if (isRtl) {
       xPos = -(xPos + elem.width);
@@ -174,7 +170,8 @@ Blockly.blockRendering.Debug.prototype.drawRenderedElem = function(elem, isRtl) 
   }
 
 
-  if (Blockly.blockRendering.Types.isInput(elem) && this.config_.connections) {
+  if (Blockly.blockRendering.Types.isInput(elem) &&
+      Blockly.blockRendering.Debug.config.connections) {
     this.drawConnection(elem.connection);
   }
 };
@@ -187,7 +184,7 @@ Blockly.blockRendering.Debug.prototype.drawRenderedElem = function(elem, isRtl) 
  * @package
  */
 Blockly.blockRendering.Debug.prototype.drawConnection = function(conn) {
-  if (!this.config_.connections) {
+  if (!Blockly.blockRendering.Debug.config.connections) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


core/renderers/common/debugger.js
```diff
@@ -54,24 +54,20 @@ Blockly.blockRendering.Debug = function() {
...
-    // rowSpacers: true,
-    // elemSpacers: true,
-    // rows: true,
-    elems: true,
-    // connections: true,
-    blockBounds: true,
-    // connectedBlockBounds: true
-  };
+/**
+ * @type {!Object.<string, boolean>} Configuration object containing booleans
```
samelhusseini: Place description first, then @type so we're consistent, [see](https://github.com/google/blockly/blob/master/core/field.js#L56)

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+    optionInput.checked = !!overrideOptions[optionName];
+  }
+}
+
+function updateRenderDebugOptions() {
+  var renderDebugOptionsListEl = document.getElementById('renderDebugOptions');
+  var renderDebugOptionInputs =
+      renderDebugOptionsListEl.getElementsByTagName('INPUT');
+  var debugOptions = {};
+  for (var optionInput of renderDebugOptionInputs) {
```
samelhusseini: ``for ... of`` introduced in ES6 which won't work in IE:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

tests/playground.html
```diff
@@ -260,6 +265,57 @@
       'removeDynamicOption', Blockly.TestBlocks.removeDynamicDropdownOption_);
 }
 
+function setRenderDebugOptionCheckboxState(overrideOptions) {
+  Blockly.blockRendering.Debug.config = overrideOptions || {};
+  if(!overrideOptions) return;
```
samelhusseini: Space after if, return statement on a new line.

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+  sessionStorage.setItem(
+      'blockRenderDebugOptions',JSON.stringify(debugOptions));
+}
+
+function addRenderDebugOptionsCheckboxes() {
+  var renderDebugConfig = Blockly.blockRendering.Debug.config;
+  var renderDebugOptionsListEl = document.getElementById('renderDebugOptions');
+  for (var optionName in renderDebugConfig) {
+    var optionCheckId = 'RenderDebug' + optionName + 'Check';
+    var optionLabel = document.createElement('LABEL');
```
samelhusseini: lowercase element name `label` for consistency with rest of Blockly.

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+    optionLabel.setAttribute('htmlFor', optionCheckId);
+    optionLabel.innerHTML = optionName;
+    var optionCheck = document.createElement('INPUT');
```
samelhusseini: lowercase element name `input` for consistency with rest of Blockly.

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+  for (var optionInput of renderDebugOptionInputs) {
+    var optionId = optionInput.getAttribute('id');
+    var checkboxMatch = optionId.match('^RenderDebug(.*)Check$');
+    if (!checkboxMatch) continue;
+    var optionName = checkboxMatch[1];
+    debugOptions[optionName] = optionInput.checked;
+  }
+  Blockly.blockRendering.Debug.config = debugOptions;
+  sessionStorage.setItem(
+      'blockRenderDebugOptions',JSON.stringify(debugOptions));
```
samelhusseini: space before ``JSON.stringify``

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+      renderDebugOptionsListEl.getElementsByTagName('INPUT');
+  var debugOptions = {};
+  for (var optionInput of renderDebugOptionInputs) {
+    var optionId = optionInput.getAttribute('id');
+    var checkboxMatch = optionId.match('^RenderDebug(.*)Check$');
+    if (!checkboxMatch) continue;
+    var optionName = checkboxMatch[1];
+    debugOptions[optionName] = optionInput.checked;
+  }
+  Blockly.blockRendering.Debug.config = debugOptions;
```
samelhusseini: Why are we updating all the options every time. Can't we just use ``e.target`` to determine which option was update and update that individually?

tests/playground.html
```diff
@@ -260,6 +265,57 @@
...
+  sessionStorage.setItem(
+      'blockRenderDebugOptions',JSON.stringify(debugOptions));
+}
```
samelhusseini: Would it also be possible to re-render the workspace when an option is updated: 
``workspace.render()``

core/renderers/common/debugger.js
moniika: Done

tests/playground.html
moniika: Done.

tests/playground.html
moniika: Thanks for catching, updated

tests/playground.html
moniika: Updated.

tests/playground.html
moniika: Good point, I was constructing the whole object originally to pass as return for config function, but now that its an object it's can more easily be handled. I've updated to use e.target as suggested.

tests/playground.html
moniika: I would also prefer for the workspace to be re-rendered. I've added the call.

tests/playground.html
```diff
@@ -260,6 +265,51 @@
...
+  config[optionName] = !!target.checked;
+  sessionStorage.setItem(
+      'blockRenderDebugOptions', JSON.stringify(config));
+  workspace.render();
+}
+
+function addRenderDebugOptionsCheckboxes() {
+  var renderDebugConfig = Blockly.blockRendering.Debug.config;
+  var renderDebugOptionsListEl = document.getElementById('renderDebugOptions');
+  Object.keys(renderDebugConfig).forEach(function(optionName) {
```
samelhusseini: .forEach is ES6 :(

moniika: Done.

tests/playground.html
```diff
@@ -260,6 +265,52 @@
...
+      'blockRenderDebugOptions', JSON.stringify(config));
+  workspace.render();
+}
+
+function addRenderDebugOptionsCheckboxes() {
+  var renderDebugConfig = Blockly.blockRendering.Debug.config;
+  var renderDebugOptionsListEl = document.getElementById('renderDebugOptions');
+  var optionNames = Object.keys(renderDebugConfig);
+  for (var i = 0, optionName; (optionName = optionNames[i]); i++) {
+    var optionCheckId = 'RenderDebug' + optionName + 'Check';
```
samelhusseini: do we need the id anymore?

moniika: Yes so that I can associate the label with the input. This association is convenient because you can click on the label to check the checkbox (but in testing I realize I should have setattribute for 'for' instead of 'htmlfor', I fixed and pushed and now that is the behavior)

Issue #3018, opened 2019-09-13
Have a way to choose which renderers are included in your bundle


samelhusseini:
We currently don't have a way to pick and choose which fields or renderers are included in a bundle for those building with the closure compiler (Blockly Games, App Inventor, etc). 

We should pull out any Blockly "modules" into a separate entry point that we then use as our main entry point for building compressed Blockly. 

Folks wanting to customize what's included in their bundle are then able to goog.require whatever fields / renderers they want.

RoboErikG:
Nice to have for this release but could be punted if necessary.

samelhusseini:
Fixed.

PR #3019, opened 2019-09-13
Blockly entry point

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3018

### Proposed Changes

Move Blockly's fields and renderers into a separate entry point used for building Blockly compressed. 
This makes it so folks building Blockly with closure can selectively choose which renderers and which fields are included in their bundle. 

### Reason for Changes

More modularity.

@NeilFraser FYI


NeilFraser:
Can a few more requires be moved?  A couple that stand out to me are:

Many Blockly applications don't use the toolbox (e.g. Blockly Games: Puzzle, Maze, Bird).  The toolbox should definitely be optional given the tree of dependencies that hangs off of it.

Workspace comments are off by default anyway (I can't figure out how to turn them on), and seem to total 1439 lines of code across three files.

There's also generators (Puzzle (and maybe Scratch?) doesn't use them), zoom controls, variables, procedures, grid, mutators, and more.  Each is a cost-benefit analysis of how easy each is to modularize vs the size savings.

samelhusseini:
I've filed a new issue to track some of the other modules that can be moved out: https://github.com/google/blockly/issues/3023

PR #3020, opened 2019-09-15
Simplify build script

NeilFraser:
* Remove unused imports and reorder.
* Compute default values during argument parsing, not at every end-point.
* Always rebuild JSON messages, since it just takes a split second.  Removes a complicated file-timestamp check.
* Remove support for “old argument style” (whatever that was).
* Restore Python 3 compatibility (injecting Blockly version number only works on strings, not bytes).

### DIFF
--- a/build.py
+++ b/build.py
@@ -460,6 +460,13 @@ class Gen_langfiles(threading.Thread):
       else:
         print("FAILED to create " + f)
 
+# Class to hold arguments if user passes in old argument style.
+class Arguments:
+  def __init__(self):
+    self.core = False
+    self.generators = False
+    self.langfiles = False
+
 # Gets the command line arguments.
 def get_args():
   parser = argparse.ArgumentParser(description="Decide which files to build.")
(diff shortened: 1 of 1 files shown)


alschmiedt:
Just for context, old argument style was passing in the arguments as core, generators, or langfiles instead of -core, -generators, -langfiles. As far as I could tell the python parser we switched to needs some kind of symbol before the arguments. We left both because we didn't want to break people's build scripts. Not sure if we still want to support both, but we should at least print out some kind of error message saying what has happened if we switch to the new version.

NeilFraser:
> Just for context, old argument style was passing in the arguments as core, generators, or langfiles instead of -core, -generators, -langfiles.

Ah, that's helpful.  I couldn't figure out what that was.

> we should at least print out some kind of error message saying what has happened if we switch to the new version.

It does have a useful message:
```
./build.py core
usage: build.py [-h] [-core] [-generators] [-langfiles]
build.py: error: unrecognized arguments: core
```

Your call if we want to keep the old arguments in or not.

alschmiedt:
Hmm, I would probably keep it in for right now. Rachel and I talked about this and I think the consensus was to keep support for both right now and start deprecating the old argument style.

PR #3031, opened 2019-09-17
Check if a type exists before testing instanceof

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Check if a type exists before instance of in case a field doesn't exist.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1338,7 +1338,8 @@ Blockly.Block.prototype.toString = function(opt_maxLength, opt_emptyToken) {
   } else {
     for (var i = 0, input; input = this.inputList[i]; i++) {
       for (var j = 0, field; field = input.fieldRow[j]; j++) {
-        if (field instanceof Blockly.FieldDropdown && !field.getValue()) {
+        if (Blockly.FieldDropdown &&
+            field instanceof Blockly.FieldDropdown && !field.getValue()) {
           text.push(emptyFieldPlaceholder);
         } else {
           text.push(field.getText());
--- a/core/block.js
+++ b/core/block.js
@@ -1338,12 +1338,7 @@ Blockly.Block.prototype.toString = function(opt_maxLength, opt_emptyToken) {
   } else {
     for (var i = 0, input; input = this.inputList[i]; i++) {
       for (var j = 0, field; field = input.fieldRow[j]; j++) {
-        if (Blockly.FieldDropdown &&
-            field instanceof Blockly.FieldDropdown && !field.getValue()) {
-          text.push(emptyFieldPlaceholder);
-        } else {
-          text.push(field.getText());
-        }
+        text.push(field.getText());
       }
       if (input.connection) {
         var child = input.connection.targetBlock();
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -632,9 +632,7 @@ Blockly.BlockSvg.prototype.createTabList_ = function() {
   var list = [];
   for (var i = 0, input; input = this.inputList[i]; i++) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
-      if (Blockly.FieldTextInput &&
-          field instanceof Blockly.FieldTextInput && field.isVisible()) {
-        // TODO (#1276): Also support dropdown fields.
+      if (field.isTabNavigable() && field.isVisible()) {
         list.push(field);
       }
     }
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -632,7 +632,8 @@ Blockly.BlockSvg.prototype.createTabList_ = function() {
   var list = [];
   for (var i = 0, input; input = this.inputList[i]; i++) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
-      if (field instanceof Blockly.FieldTextInput && field.isVisible()) {
+      if (Blockly.FieldTextInput &&
+          field instanceof Blockly.FieldTextInput && field.isVisible()) {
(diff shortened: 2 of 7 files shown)


core/renderers/measurables/row_elements.js
```diff
@@ -95,7 +95,7 @@ Blockly.blockRendering.Field = function(constants, field, parentInput) {
   Blockly.blockRendering.Field.superClass_.constructor.call(this, constants);
   this.field = field;
   this.isEditable = field.isCurrentlyEditable();
-  this.flipRtl = field instanceof Blockly.FieldImage && field.getFlipRtl();
+  this.flipRtl = field.getFlipRtl && field.getFlipRtl();
```
NeilFraser: .getFlipRtl() should be moved to Field, and Field.prototype.flipRtl_ should just be set to false.

core/block_svg.js
```diff
@@ -632,7 +632,8 @@ Blockly.BlockSvg.prototype.createTabList_ = function() {
   var list = [];
   for (var i = 0, input; input = this.inputList[i]; i++) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
-      if (field instanceof Blockly.FieldTextInput && field.isVisible()) {
+      if (Blockly.FieldTextInput &&
+          field instanceof Blockly.FieldTextInput && field.isVisible()) {
```
NeilFraser: This smells pretty bad, and the TODO on the next line confirm that this is the wrong approach since more and more subclass code needs to be injected here if we continue this approach.  Suggest moving this code to the subclasses.

core/block.js
```diff
@@ -1338,7 +1338,8 @@ Blockly.Block.prototype.toString = function(opt_maxLength, opt_emptyToken) {
   } else {
     for (var i = 0, input; input = this.inputList[i]; i++) {
       for (var j = 0, field; field = input.fieldRow[j]; j++) {
-        if (field instanceof Blockly.FieldDropdown && !field.getValue()) {
+        if (Blockly.FieldDropdown &&
+            field instanceof Blockly.FieldDropdown && !field.getValue()) {
```
NeilFraser: I'm scratching my head regarding how a dropdown field would have no value, and why a '?' would be better than '' in that case.

core/renderers/measurables/row_elements.js
BeksOmega: If it gets moved it might need a doc issue as well.

core/block.js
RoboErikG: Dynamic dropdowns sometimes don't have a value if they haven't had anything added yet.

NeilFraser: Ah, that makes this easier.  My interpretation of the current code is that the value in this case would be null.  If that's the case, then this line should just be:
`if (field.getValue() === null) {`

BeksOmega: > Dynamic dropdowns sometimes don't have a value if they haven't had anything added yet.

Doesn't Blockly throw errors if this occures?

<details>
<summary>Steps to Reproduce</summary>
<p>

1.  In the test blocks file change:
```
Blockly.Blocks['test_dropdowns_dynamic'] = {
  init: function() {
    var dropdown = new Blockly.FieldDropdown(this.dynamicOptions);
    this.appendDummyInput()
      .appendField('dynamic')
      .appendField(dropdown, 'OPTIONS');
  },

  dynamicOptions: function() {
    if (!Blockly.TestBlocks.dynamicDropdownOptions_.length) {
      return [['', 'OPTION0']];
    }
    return Blockly.TestBlocks.dynamicDropdownOptions_;
  }
};
```
to:
```
Blockly.Blocks['test_dropdowns_dynamic'] = {
  init: function() {
    var dropdown = new Blockly.FieldDropdown(this.dynamicOptions);
    this.appendDummyInput()
      .appendField('dynamic')
      .appendField(dropdown, 'OPTIONS');
  },

  dynamicOptions: function() {
    return Blockly.TestBlocks.dynamicDropdownOptions_;
  }
};
```
2. Import the following xml to the playground:
```
  <block type="test_dropdowns_dynamic"></block>
</xml>
```
3. Observe the error  in console.
```
Uncaught TypeError: Cannot read property '1' of undefined
    at new Blockly.FieldDropdown (field_dropdown.js:82)
    at Blockly.BlockSvg.init (my_test_blocks.js:1657)
    at Blockly.BlockSvg.Blockly.Block (block.js:183)
    at new Blockly.BlockSvg (block_svg.js:103)
    at Blockly.WorkspaceSvg.newBlock (workspace_svg.js:703)
    at Object.Blockly.Xml.domToBlockHeadless_ (xml.js:618)
    at Object.Blockly.Xml.domToBlock (xml.js:536)
    at Object.Blockly.Xml.domToWorkspace (xml.js:417)
    at fromXml (my_playground.html?dir=ltr&toolbox=test-blocks&side=start&theme=classic&export=clean-xml:384)
    at HTMLInputElement.onclick (my_playground.html?dir=ltr&toolbox=test-blocks&side=start&theme=classic&export=clean-xml:132)
```

</p>
</details>

samelhusseini: [This](https://github.com/google/blockly/commit/2c72590a0199474790d8ba7248fa7331210cbb57) is when this code was introduced. I think we no longer allow fields to have null values, so going to remove this code path.

NeilFraser: Turns out this was added several years ago during the creation of Accessible Blockly, to support non-graphical blocks.  That project is dead, and as Beka points out the behaviour has changed, so we agreed to remove this code.

PR #3034, opened 2019-09-17
Add require statements to all block files.

NeilFraser:
Allows each set of blocks to pull in the correct fields when they are used in a compiled application.

### DIFF
--- a/blocks/colour.js
+++ b/blocks/colour.js
@@ -32,8 +32,10 @@
 goog.provide('Blockly.Blocks.colour');  // Deprecated
 goog.provide('Blockly.Constants.Colour');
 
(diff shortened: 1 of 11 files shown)


samelhusseini:
Assigning this to myself as I'd like to take a look at how this affects our NPM build of blocks, making sure it doesn't pull in those files twice.

samelhusseini:
@googlebot I consent.

samelhusseini:
I've taken a look at this and it LGTM, I've applied a similar patch to the gulp build script: 
https://github.com/google/blockly/pull/3034/commits/e904ec179ca7b25d66bdbf7b664c8c67f49f32e5

Issue #3032, opened 2019-09-17
npm: new blocks instances appear below menuFlyout while dragging


microspace:
Blocks always go under menuflyout when I drag clock to create new block instance.
Issue appears when I use **npm** package of blockly

1. Go to https://github.com/google/blockly-samples/tree/master/blockly-webpack
2. clone, npm install, npm run start
3. Drag any of block from toolbox.
4. See error: block appears below menuFlyout

![Снимок](https://user-images.githubusercontent.com/2958584/65040348-966e8980-d96d-11e9-87a9-19b54fe5305f.PNG)

**Desktop:**
 - OS: Windows 10
 - Chrome browser
 - Version 76

samelhusseini:
Fixed in develop. We are planning an npm release in a few weeks.

microspace:
@samelhusseini , thank you! Can I use the develop version for now, or it is not possible?

RoboErikG:
@microspace We don't publish our develop branch to npm, so if you would like to use it you'll have to clone it from the GitHub repo.

microspace:
@samelhusseini , sorry to bother, but do I have to build develop version myself?
I get error "no module named blockly" when I install develop version of blckly from github

PR #3035, opened 2019-09-17
Fix Blockly.utils.global when wrapped in a module.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3032
An any bugs that arise from Blockly.utils.global not pointing to the global object.

### Proposed Changes

The previous logic in Blockly.utils.global was referencing ``this`` and falling back onto ``self`` if this didn't exist. When Blockly is wrapped in a UMD module, ``this`` references the function scope of the module. 

``self`` points to window in browsers and web-workers. Not generally implemented outside of browsers.
``window`` points to the window object. Not generally implemented outside of in-page browsers.
``global`` points to the global object in Node. Not implemented outside of Node.

I left out ``globalThis`` as it's not widely implemented outside of V8.

### Test Coverage

Tested Blockly samples, webpack, react and node.

Tested on:
* Desktop Chrome
* Desktop Firefox

### Additional Information

This closure library doc has further context on implementation: 
https://docs.google.com/document/d/1NAeW4Wk7I7FV0Y2tcUFvQdGMc89k2vdgSXInw8_nvCI/edit#

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@ blockly_node_javascript_en.js
 gulpfile.js
 /msg/*
 /core/css.js
+/core/utils/global.js
 /tests/blocks/*
 /tests/compile/*
 /tests/jsunit/*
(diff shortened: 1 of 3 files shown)


samelhusseini:
@NeilFraser FYI

Issue #2493, opened 2019-05-24
Setting the Value of a Dropdown is Inefficient


BeksOmega:
### Problem statement

validateOptions_, doClassValidation_, and doValueUpdate_ all loop through the dropdown's options. This is inefficient.

### Proposal

validateOptions_ should create a map of Language-Neutral keys to Human-Readable options. That way it could be queried when selecting a value, instead of having to loop through all of the options.

### Additional Information

If this change is made option trimming (prefix/suffix matching) would need to be reworked so that the map contains the trimmed options.

samelhusseini:
Yeah, I agree, it doesn't make sense that we constantly regenerate the options list, especially that the returned values could be different. 
I think it makes sense to re-generate the options list when the dropdown is opened.

BeksOmega:
Relevant Discussion: https://github.com/google/blockly/pull/3036#discussion_r325908979

Issue #2926, opened 2019-08-28
During domToWorkspace() child block with FieldDropdown is in workspace's topblock


ArfyFR:
I updated blockly in a project from
1.20190419.0
to "npm": "blockly": "2.20190722.1"

In order to restore a view, I use
`    Blockly.Xml.domToWorkspace(this.contentData.content, this.workspace);
`

In our project we only have one top/root block (with childs) by design.
Some child blocks have FieldDropdown, which are filled with a function like
```
	new Blockly.FieldDropdown(function createSelectedDataTableField_func() {
		let selectedTables = [];
		...
		return selectedTables;
```
with FieldDropdown values **depending on what is set in the root block**.

With Blockly 1.20190419.0 => **OK**

But with Blockly 2.20190722.1 after domToWorkspace, I faced the following.  Traces in console like:
> Cannot set the dropdown's value to an unavailable option. Block type: dataset_field_table_column, Field name: FIELD_TABLE, Value: mapchartbycust

In the createSelectedDataTableField_func() call, the child block's getRootBlock() is itself **O_o**
And the workspace has two+ top blocks, the good one and the child(s).  So If I call getRootBlock() to do a
`	let xyz = rootBlock.getFieldValue('SELECTED_TABLE');
`
call it won't work anymore.

I had to (bad) hack our code with
```
	if(rootBlock.type == 'dataset_field_table_column') {
	  rootBlock = rootBlock.workspace.getTopBlocks()[0];
	}
```
to select the "good" root block.

Later, I get the Blockly.Events.BLOCK_CREATE event, and the child(s) block(s) are (back) to root block childs => OK

**Expected behavior**

Like before

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 76.0.3809.100

BeksOmega:
> @BeksOmega any ideas before I dig in?

The only change I can think of (without digging into it) is that dropdown validators are now called on XML values. This problem doesn't seem to involve validators though, so idk :/

ArfyFR:
Hi,

My problem is more: **in previous version**, the root block DURING `Blockly.Xml.domToWorkspace()` was the good one ;)
(I noticed I wasn't perhaps clear, I use "**Dynamic dropdowns**" with a function cf [doc](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#dynamic_dropdowns) )

[...]

PR #3036, opened 2019-09-17
Cache dynamic dropdown options

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2493
https://github.com/google/blockly/issues/2926

### Proposed Changes
Only generate options during inialization (as we need the first tuple) and when a use clicks on the field to open it. (ie: when the editor is shown / widget is created). 

### Reason for Changes

Be more efficient about calling a dynamic dropdown function and not break in places where we expect certain options to be available. 

### Test Coverage

Tested the playground, and added a new test dropdown block that returns a random number of items every time the generate method is called. 

Tested on:
* Desktop Chrome

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -433,7 +433,6 @@ Blockly.FieldDropdown.prototype.isOptionListDynamic = function() {
  *     cached options or to re-generate them.
  * @return {!Array.<!Array>} A non-empty array of option tuples:
  *     (human-readable text or image, language-neutral name).
- * @throws If generated options are incorrectly structured.
  */
 Blockly.FieldDropdown.prototype.getOptions = function(opt_useCache) {
   if (this.isOptionListDynamic()) {
@@ -599,7 +598,7 @@ Blockly.FieldDropdown.prototype.getText_ = function() {
 /**
  * Validates the data structure to be processed as an options list.
  * @param {?} options The proposed dropdown options.
- * @throws If proposed options are incorrectly structured.
+ * @throws {TypeError} If proposed options are incorrectly structured.
  * @private
  */
 Blockly.FieldDropdown.validateOptions_ = function(options) {
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -83,7 +83,7 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
   this.selectedIndex_ = 0;
 
   this.trimOptions_();
-  var options = this.getOptions(true);
+  var options = this.getOptions(false);
   var firstTuple = options[0];
 
   // Call parent's constructor.
@@ -238,7 +238,7 @@ Blockly.FieldDropdown.prototype.widgetCreate_ = function() {
   menu.setRightToLeft(this.sourceBlock_.RTL);
   menu.setRole('listbox');
 
-  var options = this.getOptions(true);
+  var options = this.getOptions(false);
   this.selectedMenuItem_ = null;
   for (var i = 0; i < options.length; i++) {
     var content = options[i][0]; // Human-readable text or image.
@@ -438,7 +438,7 @@ Blockly.FieldDropdown.prototype.isOptionListDynamic = function() {
  */
 Blockly.FieldDropdown.prototype.getOptions = function(opt_useCache) {
   if (this.isOptionListDynamic()) {
-    if (!opt_useCache || !this.generatedOptions_) {
+    if (!this.generatedOptions_ || !opt_useCache) {
       this.generatedOptions_ = this.menuGenerator_.call(this);
       Blockly.FieldDropdown.validateOptions_(this.generatedOptions_);
     }
@@ -455,7 +455,7 @@ Blockly.FieldDropdown.prototype.getOptions = function(opt_useCache) {
  */
 Blockly.FieldDropdown.prototype.doClassValidation_ = function(opt_newValue) {
   var isValueValid = false;
-  var options = this.getOptions(false);
+  var options = this.getOptions(true);
   for (var i = 0, option; option = options[i]; i++) {
     // Options are tuples of human-readable text and language-neutral values.
     if (option[1] == opt_newValue) {
@@ -481,7 +481,7 @@ Blockly.FieldDropdown.prototype.doClassValidation_ = function(opt_newValue) {
  */
 Blockly.FieldDropdown.prototype.doValueUpdate_ = function(newValue) {
   Blockly.FieldDropdown.superClass_.doValueUpdate_.call(this, newValue);
-  var options = this.getOptions(false);
+  var options = this.getOptions(true);
   for (var i = 0, option; option = options[i]; i++) {
     if (option[1] == this.value_) {
       this.selectedIndex_ = i;
@@ -514,7 +514,7 @@ Blockly.FieldDropdown.prototype.render_ = function() {
   this.imageElement_.style.display = 'none';
 
   // Show correct element.
-  var options = this.getOptions(false);
+  var options = this.getOptions(true);
   var selectedOption = this.selectedIndex_ >= 0 &&
       options[this.selectedIndex_][0];
   if (selectedOption && typeof selectedOption == 'object') {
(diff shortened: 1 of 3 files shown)


samelhusseini:
@BeksOmega FYI.

I'm also a little surprised that we don't support an empty options list, but that can be dealt with in a different change.

core/field_dropdown.js
```diff
@@ -421,15 +430,18 @@ Blockly.FieldDropdown.prototype.isOptionListDynamic = function() {
 
 /**
  * Return a list of the options for this dropdown.
+ * @param {boolean=} opt_regenerate Whether or not to re-generate the options.
  * @return {!Array.<!Array>} Array of option tuples:
  *     (human-readable text or image, language-neutral name).
  * @throws If generated options are incorrectly structured.
  */
-Blockly.FieldDropdown.prototype.getOptions = function() {
+Blockly.FieldDropdown.prototype.getOptions = function(opt_regenerate) {
```
BeksOmega: `opt_regenerate` probably needs to default to true for backwards compat.

core/field_dropdown.js
```diff
@@ -617,6 +629,9 @@ Blockly.FieldDropdown.validateOptions_ = function(options) {
           tuple);
     }
   }
+  if (!options.length) {
```
BeksOmega: Thank you! Uncaught errors are the worst.

core/field_dropdown.js
```diff
@@ -68,15 +68,23 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
   this.menuGenerator_ = menuGenerator;
 
   /**
-   * The currently selected index. A value of -1 indicates no option
-   * has been selected.
+   * A cache of the most recently generated options.
+   * @type {Array.<!Array.<string>>}
+   * @private
+   */
+  this.generatedOptions_ = null;
```
BeksOmega: Did you look into having this be a map/dictionary? I think there are some loops that could be elliminated that way.

core/field_dropdown.js
```diff
@@ -68,15 +68,23 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator) {
   this.menuGenerator_ = menuGenerator;
 
   /**
-   * The currently selected index. A value of -1 indicates no option
-   * has been selected.
+   * A cache of the most recently generated options.
+   * @type {Array.<!Array.<string>>}
```
BeksOmega: I think `.` in array declarations is deprecated.

tests/blocks/test_blocks.js
```diff
@@ -1210,3 +1210,21 @@ Blockly.TestBlocks.removeDynamicDropdownOption_ = function() {
     }
   })
 };
+
+Blockly.Blocks['test_dropdowns_dynamic_random'] = {
```
BeksOmega: Looks fun hehe.

BeksOmega:
> I'm also a little surprised that we don't support an empty options list, but that can be dealt with in a different change.

Do you have a use case in mind? I think that's it generally good to oppose that behavior so that you avoid dealing with nulls in code generation. I'm having trouble thinking of a situation where the limitation can't be designed around (e.g. variables), but I could be missing something.

RoboErikG:
> Do you have a use case in mind? I think that's it generally good to oppose that behavior so that you avoid dealing with nulls in code generation. I'm having trouble thinking of a situation where the limitation can't be designed around (e.g. variables), but I could be missing something.

I mentioned to @samelhusseini that I think our APIs made this easy to do with dynamic dropdown functions. Looking at our existing code, an empty options list is allowed by validateOptions_ but then we assume it's non-empty a few lines later in the constructor. It's probably better if we handle an empty list in one spot instead of expecting every option generator to check for it.

core/field_dropdown.js
samelhusseini: It is, but we're keeping it in Blockly for consistency.

BeksOmega:
>  It's probably better if we handle an empty list in one spot instead of expecting every option generator to check for it.

I agree that it is a bit annoying to have to design around this limitation (as with most limitations hehe) but I think allowing empty lists and nulls would actually end up creating /more/ work (for us, and outside developers) in the longrun.

Firstly, if outside developers don't handle this at option-generation-time they'll end up having to handle it at code-generation-time. That is a much trickier spot to find bugs. I think we're actually helping them by notifying them before it gets to that point.

Secondly, allowing empty lists and nulls creates a lot more work for us, because now we have to check for that.

We'd have to start checking for empty lists all over the dropdown field, and special case them. For example:
  * Constructor would have to pass something different to the super.
  * getText_ would have to return `''` or `BLANK` or some placeholder.
  * And render_ would have to special-case somehow.
  
We'd also no longer be able to confidently call `field.getValue()` in a generic context, because it could return null. Making getValue() non-null saves us a lot of work, because it means you can remove `if(field.getValue())` checks.

Two other small points:
1) That `getOptions()[0]` line has been around since the [dawn of Blockly](https://github.com/google/blockly/blame/89af53a893185b269d9db635ce5cde29237cf5f6/core/field_dropdown.js#L46), and people seem mostly cool with it.
2) Displaying an empty dropdown is pretty buggy looking behavior anyway. If I were using a block programming language, I'd report that.

In Conclusion:
I think that throwing errors for empty lists is the proper way to handle things.

It saves outside developers the headache in the long run. It makes Blockly less fragile and more elegant. And I can't think of a usecase for empty lists (at least one that doesn't have a better solution).

RoboErikG:
@BeksOmega good points, especially about it never working with an empty list. If @samelhusseini agrees then I'm okay with requiring a non-empty list of options and updating validateOptions_ to explicitly check for an empty list and throw a meaningful error in that case.

samelhusseini:
I agree with the points made here. This PR adds a check in ``validateOptions`` and throws if the list returned is empty. I'll add further code comments so that we're more explicit about this.

core/field_dropdown.js
samelhusseini: What would the key of this map be? If it's the value, it means you can't have multiple options with the same value, label, is the same. I don't think we set that restrictions right now.

core/field_dropdown.js
samelhusseini: Yeah that makes sense.

core/field_dropdown.js
```diff
@@ -421,15 +429,19 @@ Blockly.FieldDropdown.prototype.isOptionListDynamic = function() {
 
 /**
  * Return a list of the options for this dropdown.
- * @return {!Array.<!Array>} Array of option tuples:
+ * @param {boolean=} opt_useCache For dynamic options, whether or not to use the
+ *     cached options or to re-generate them.
+ * @return {!Array.<!Array>} A non-empty array of option tuples:
  *     (human-readable text or image, language-neutral name).
  * @throws If generated options are incorrectly structured.
```
alschmiedt: nit: Could we add the type here? @throws {Type}

samelhusseini: Done.

core/field_dropdown.js
BeksOmega: > What would the key of this map be?

It would be keyed by value. This is what the dropdown currently uses, it's just inefficient and uses for-loops instead of a map.

We may not currently throw errors, but the language-neutral keys of your dropdown have to be unique, otherwise your dropdown breaks:

SameValues

<details><summary>Steps to reproduce</summary>
<p>

1. Add the following block definition to the playground.
```
{
    "type": "test_dropdowns_same_value",
    "message0": "same value %1",
    "args0": [
      {
        "type": "field_dropdown",
        "name": "FIELDNAME",
        "options": [
          [ "first item", "ITEM1" ],
          [ "second item", "ITEM2" ],
          [ "third item", "ITEM2" ],
          [ "fourth item", "ITEM2" ],
          [ "fifth item", "ITEM2" ],
        ]
      }
    ],
    "style": "fields_dropdowns"
  }
```
2. Add the block to the workspace.
3. Use the dropdown.
4. Observe the bad behavior.
</p>
</details>

If we were to switch it to a map we could definitely get rid of the loop in `doClassValidation_`. We could also probably get rid of the `selectedIndex_` property and just use the value instead, elliminating the loop in `doValueUpdate_`.

PR #3037, opened 2019-09-17
Creating cursordrawer in renderer.

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

Moves initialization of drawer for cursor to the renderer, rather than in the WorkspaceSvg constructor.
Additionally, in order to have access to the renderer when setCursor and setMarker are called, the initialization of the renderer was moved to getRenderer(), so the renderer is now initialized at the first time getRenderer is called, rather than in the WorkspaceSvg constructor. At this time, this means that the renderer is initialized at the overridden setCursor call for WorkspaceSvg (inside of the super constructor call).

### Reason for Changes

We expect some people to customize their cursor, so moving the initialization to the renderer gives more flexibility for changing what implementation of the cursor drawer to call.

### Test Coverage

Tested in playground.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/common/renderer.js
+++ b/core/renderers/common/renderer.js
@@ -32,6 +32,7 @@ goog.require('Blockly.blockRendering.Drawer');
 goog.require('Blockly.blockRendering.IPathObject');
 goog.require('Blockly.blockRendering.PathObject');
 goog.require('Blockly.blockRendering.RenderInfo');
+goog.require('Blockly.CursorSvg');
 
 
 /**
@@ -50,7 +51,7 @@ Blockly.blockRendering.Renderer = function() {
 };
 
 /**
- * Initialize the renderer
+ * Initialize the renderer.
  * @package
  */
 Blockly.blockRendering.Renderer.prototype.init = function() {
@@ -99,6 +100,19 @@ Blockly.blockRendering.Renderer.prototype.makeDebugger_ = function() {
 };
 
 /**
+ * Create a new instance of the renderer's cursor drawer
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace the cursor belongs to.
+ * @param {boolean=} opt_marker True if the cursor is a marker. A marker is used
+ *     to save a location and is an immovable cursor. False or undefined if the
+ *     cursor is not a marker.
+ * @return {!Blockly.CursorSvg} The cursor drawer.
+ * @package
+ */
+Blockly.blockRendering.Renderer.prototype.makeCursorDrawer = function(workspace, opt_marker) {
+  return new Blockly.CursorSvg(workspace, opt_marker);
+};
+
+/**
  * Create a new instance of a renderer path object.
  * @param {!SVGElement} root The root SVG element.
  * @return {!Blockly.blockRendering.IPathObject} The renderer path object.
@@ -133,4 +147,3 @@ Blockly.blockRendering.Renderer.prototype.render = function(block) {
   info.measure();
   this.makeDrawer_(block, info).draw();
 };
-
--- a/core/renderers/common/renderer.js
+++ b/core/renderers/common/renderer.js
@@ -100,7 +100,7 @@ Blockly.blockRendering.Renderer.prototype.makeDebugger_ = function() {
 };
 
 /**
- * Create a new instance of the renderer's cursor drawer
+ * Create a new instance of the renderer's cursor drawer.
  * @param {!Blockly.WorkspaceSvg} workspace The workspace the cursor belongs to.
  * @param {boolean=} opt_marker True if the cursor is a marker. A marker is used
  *     to save a location and is an immovable cursor. False or undefined if the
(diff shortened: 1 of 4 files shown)


core/workspace_svg.js
```diff
@@ -141,16 +141,16 @@ Blockly.WorkspaceSvg = function(options,
...
-  this.renderer_ = Blockly.blockRendering.init(this.options.renderer || 'geras');
 };
 Blockly.utils.object.inherits(Blockly.WorkspaceSvg, Blockly.Workspace);
 
+/**
+ * The block renderer used for rendering blocks on this workspace.
+ * @type {!Blockly.blockRendering.Renderer}
+ * @private
+ */
+Blockly.WorkspaceSvg.prototype.renderer_ = undefined;
```
samelhusseini: Moving forward we are no longer setting properties on the prototype to match the Google JS Style guide, and it's more performance with today's JS engines.

core/workspace_svg.js
```diff
@@ -421,6 +421,10 @@ Blockly.WorkspaceSvg.prototype.inverseScreenCTMDirty_ = true;
  * @return {!Blockly.blockRendering.Renderer} The renderer attached to this workspace.
  */
 Blockly.WorkspaceSvg.prototype.getRenderer = function() {
+  if (!this.renderer_) {
+    this.renderer_ =
+        Blockly.blockRendering.init(this.options.renderer || 'geras');
```
samelhusseini: I don't mind the lazy load, but what's the reasoning?

samelhusseini: It looks like there's some funny load order going on with Workspace and WorkspaceSvg when setting the cursor.
If we reserve the setCursor / setMarker methods for dynamic setting of the cursor / marker (ie: after the workspace has been initialized), we can then set ``this.cursor_ = new Blockly.Cursor()`` in the constructor of Workspace instead of calling ``setCursor``. 

This moves the initialization of the drawer over to Workspace_svg's createDom method where the cursor is being injected into the DOM.

moniika: I explained a bit in the PR description, but essentially the issue is that the renderer is normally initialized after the setCursor/setMarker methods are called, but I need to use it inside of the methods.
Your suggestion of reserving setCursor/setMarker for dynamically changing them works. It'll also make it more clear exactly when the renderer is initialized.

samelhusseini: Yeah I recognize the issue you're running into here, and I think the problem lies in the fact we were calling overridden methods from the constructor (making it hard to trace the initialization order) but more importantly assuming the renderer has been initialized.

core/workspace_svg.js
moniika: No longer needed due to removing lazy loading.

core/workspace_svg.js
moniika: Applied fix

core/renderers/common/renderer.js
```diff
@@ -98,6 +99,19 @@ Blockly.blockRendering.Renderer.prototype.makeDebugger_ = function() {
   return new Blockly.blockRendering.Debug();
 };
 
+/**
+ * Create a new instance of the renderer's cursor drawer
```
samelhusseini: nit: end with ``.``

core/workspace.js
```diff
@@ -122,24 +122,20 @@ Blockly.Workspace = function(opt_options) {
    * @type {Blockly.Cursor}
```
samelhusseini: ``!Blockly.Cursor``

core/workspace.js
```diff
@@ -122,24 +122,20 @@ Blockly.Workspace = function(opt_options) {
...
    * @protected
    */
-  this.cursor_ = null;
+  this.cursor_ = new Blockly.Cursor();
 
   /**
    * The marker used to mark a location for keyboard navigation.
    * @type {Blockly.MarkerCursor}
```
samelhusseini: ``!Blockly. MarkerCursor``

core/renderers/common/renderer.js
```diff
@@ -98,6 +99,19 @@ Blockly.blockRendering.Renderer.prototype.makeDebugger_ = function() {
...
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace the cursor belongs to.
+ * @param {boolean=} opt_marker True if the cursor is a marker. A marker is used
+ *     to save a location and is an immovable cursor. False or undefined if the
+ *     cursor is not a marker.
+ * @return {!Blockly.CursorSvg} The cursor drawer.
+ * @package
+ */
+Blockly.blockRendering.Renderer.prototype.makeCursorDrawer = function(workspace, opt_marker) {
```
samelhusseini: More than 80 chars.

core/renderers/common/renderer.js
moniika: Done.

core/workspace.js
moniika: Done

core/workspace.js
moniika: Done

core/renderers/common/renderer.js
moniika: Done.

PR #3039, opened 2019-09-18
Remove now unneeded requires from compile test.

NeilFraser:
The block files now require these fields.


NeilFraser:
Branch collision.

Issue #2811, opened 2019-08-13
Text fields re-sizing is broken after the workspace regains visibility


samelhusseini:
**Describe the bug**

Text fields don't adjust when the Blockly workspace regains visibility. 

**To Reproduce**

1. Go to the Blockly playground.
2. Hide the workspace ``Blockly.mainWorkspace.setVisible(false);``
3. Import a block to the workspace that contains a text field.
4. Show the workspace ``Blockly.mainWorkspace.setVisible(true);``
5. Notice how even after a re-render, the text field doesn't adjust its size accordingly. 

**Expected behavior**

The text field should adjust it's size after a re-render, just like other fields. 

**Screenshots**

<img width="307" alt="Screen Shot 2019-08-13 at 1 43 15 PM" src="https://user-images.githubusercontent.com/16690124/62976049-a1635680-bdd0-11e9-93a5-5599bc8ed4fc.png">

**Additional context**

This problem only occurs on develop, so something must've changed that broke this. 

This bug here is more or less the same, https://github.com/google/blockly/issues/1688
However it defines the problem in master. Don't have full context here, but I don't repro this on master.

BeksOmega:
Problem is [this line](https://github.com/google/blockly/blob/4efa0dad1f2c882234759d3ba7ecd0ebd95a4c77/core/xml.js#L533) is calling render when it shouldn't (b/c the workspace isn't visible yet).

It could be wrapped in an isVisible() check:
```
if (workspace.isVisible()) {
  for (var i = blocks.length - 1; i >= 0; i--) {
    blocks[i].render(false);
  }
}
```
But then scrollbars don't seem to resize correctly.

Another option would be to:
* First pass the visibility down to all the fields, connections, etc before rendering.
* Then don't allow the fields to cache their widths if render is called while they are invisible.

But eliminating the bad render call all together would be more efficient.

samelhusseini:
Fixed.

RoboErikG:
This is still happening on the code demo if you paste blocks into the xml tab then switch back to the blocks tab:
![image](https://user-images.githubusercontent.com/11066279/66451127-ecd06480-ea0f-11e9-95c4-8e82e5f8cfc4.png)

RoboErikG:
Repro steps:
1. Open demos/code/index.html
2. Switch to the xml tab
3. Paste the below xml
4. Switch back to the blocks tab

```
<xml xmlns="https://developers.google.com/blockly/xml">
[...]

PR #3041, opened 2019-09-18
Fix field sizing issue when workspace is hidden

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2811

### Proposed Changes

When a workspace is hidden, hide the blocks too, which in turn hides its inputs.
Vice versa when the workspace is shown. 

This makes it so that when blocks enter the workspace when it's not visible, they are not rendered. When the workspace restores visibility, they are then rendered.

For blocks that are currently on the workspace and rendered, they are merely hidden, and their visibility is restored when the workspace is set to visible. 

### Reason for Changes

Bug in field sizing when blocks enter the workspace while it's hidden. 

### Test Coverage

Hiding the playground, importing blocks and then showing it.

Tested on:
* Desktop Chrome
* Desktop Safari

### Additional Information

@BeksOmega thoughts on this solution? It's pretty similar to what you proposed in the bug.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -532,6 +532,22 @@ Blockly.BlockSvg.prototype.getBoundingRectangle = function() {
   return new Blockly.utils.Rect(top, bottom, left, right);
 };
 
+
+/**
+ * Toggles the visibility of the block.
+ * @param {boolean} isVisible True if block should be visible.
+ */
+Blockly.BlockSvg.prototype.setVisible = function(isVisible) {
+  var display = isVisible ? 'block' : 'none';
+
+  this.getSvgRoot().style.display = display;
+
+  // Show/hide the inputs.
+  for (var i = 0, input; input = this.inputList[i]; i++) {
+    input.setVisible(isVisible);
+  }
+};
+
 /**
  * Set whether the block is collapsed or not.
  * @param {boolean} collapsed True if collapsed.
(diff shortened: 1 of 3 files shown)


core/block_svg.js
```diff
@@ -532,6 +532,22 @@ Blockly.BlockSvg.prototype.getBoundingRectangle = function() {
   return new Blockly.utils.Rect(top, bottom, left, right);
 };
 
+
+/**
+ * Toggles the visibility of the block.
+ * @param {boolean} isVisible True if block should be visible.
+ */
```
BeksOmega: Currently I believe the other setVisible functions are package, because there are issues about connecting to invisible blocks and things.

core/block_svg.js
```diff
@@ -532,6 +532,22 @@ Blockly.BlockSvg.prototype.getBoundingRectangle = function() {
...
+Blockly.BlockSvg.prototype.setVisible = function(isVisible) {
+  var display = isVisible ? 'block' : 'none';
+
+  this.getSvgRoot().style.display = display;
+
+  // Show/hide the inputs.
+  for (var i = 0, input; input = this.inputList[i]; i++) {
+    input.setVisible(isVisible);
```
RoboErikG: Does this work with collapsed blocks? Collapsing a block also calls setVisible(false) on all of its inputs, if you make them all visible again weird things might happen.

samelhusseini: Good point, I'll check how the two interface together.

samelhusseini:
Circling back on this, I tried this out with collapsed blocks (as per @RoboErikG's suggestion) and it's very broken. I'll have to think of a different approach that is compatible with collapsed blocks. 

Closing for now.

PR #3044, opened 2019-09-18
Changed multiline field to use new configuration

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2722 

### Proposed Changes

Gives the multinline input field an opt_config parameter.

Also I fixed a little text input doc bug.

### Reason for Changes

Currently the field doesn't have any of its own config options, but now it can support tooltips and spellcheck correctly.

### Test Coverage

I didn't add any configuration tests because this field doesn't have any options of its own, but I did add a test block, and everything seems to be working.

Tested on:
* Desktop Chrome

### Documentation

When the multiline input field gets written up, its creation section should include info about its config properties.

### Additional Information

N/A

### DIFF
--- a/core/field_multilineinput.js
+++ b/core/field_multilineinput.js
@@ -26,7 +26,7 @@
  */
 'use strict';
 
-goog.provide('Blockly.FieldMultilineCode');
+goog.provide('Blockly.FieldMultilineInput');
 
 goog.require('Blockly.DropDownDiv');
 goog.require('Blockly.FieldTextInput');
--- a/core/field_multilineinput.js
+++ b/core/field_multilineinput.js
@@ -26,7 +26,7 @@
  */
 'use strict';
 
-goog.provide('Blockly.FieldMultilineInput');
+goog.provide('Blockly.FieldMultilineCode');
 
 goog.require('Blockly.DropDownDiv');
 goog.require('Blockly.FieldTextInput');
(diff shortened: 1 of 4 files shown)


core/field_multilineinput.js
```diff
@@ -26,7 +26,7 @@
  */
 'use strict';
 
-goog.provide('Blockly.FieldMultilineInput');
+goog.provide('Blockly.FieldMultilineCode');
```
samelhusseini: ??

core/field_multilineinput.js
```diff
@@ -70,19 +75,14 @@ Blockly.FieldMultilineInput.LINE_HEIGHT = 20;
...
  * @nocollapse
  */
 Blockly.FieldMultilineInput.fromJson = function(options) {
   var text = Blockly.utils.replaceMessageReferences(options['text']);
-  var field = new Blockly.FieldMultilineInput(text);
-  if (typeof options['spellcheck'] === 'boolean') {
-    field.setSpellcheck(options['spellcheck']);
-  }
-  return field;
+  return new Blockly.FieldMultilineInput(text, null, options);
```
samelhusseini: Just FYI, and this affects all fields not just this one (I just haven't gotten around to fixing it up), but passing null here is wrong as the constructor is expecting the type ``Function=``, ie: ``Function|undefined``. Instead this can be changed to ``undefined``.

samelhusseini:
Thank you!

PR #3048, opened 2019-09-19
Make mutator an optional module.

NeilFraser:


NeilFraser:
Confirming a 3 KB drop in compiled applications that don't use a mutator.

PR #3053, opened 2019-09-19
Make block comments an optional module.

NeilFraser:
Measured as a 5 KB *increase* in code size, but that's because some other commit just landed between my tests.  It's like running down an up escalator.

### DIFF
--- a/core/requires.js
+++ b/core/requires.js
@@ -29,6 +29,7 @@ goog.provide('Blockly.requires');
 
 // Blockly Core.
 goog.require('Blockly');
+goog.require('Blockly.Comment');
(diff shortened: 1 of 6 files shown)


core/requires.js
```diff
@@ -29,6 +29,7 @@ goog.provide('Blockly.requires');
 
 // Blockly Core.
 goog.require('Blockly');
+goog.require('Blockly.Comment');
```
RoboErikG: Might be nice to add a comment that comments should be disabled in the inject options if removing this: https://developers.google.com/blockly/guides/get-started/web#configuration

NeilFraser: Done

PR #3054, opened 2019-09-19
Merge toolbox navigation 

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Merge both types of toolbox navigation (when in accessibility mode) and when we're not. 
The arrow keys can be used when we're not in accessibility mode and focused on the toolbox.

The logic was duplicated into Blockly.navigation. This moves it all into the one places and makes use of the same methods in base_node.js

I've also moved the handling of a blockly action on a toolbox into toolbox.js.

### Reason for Changes

The two navigation methods were tracking the selected category differently causing conflicts when using both interchangeably.

### Test Coverage

Tested playground.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/components/tree/basenode.js
+++ b/core/components/tree/basenode.js
@@ -1206,7 +1206,7 @@ Blockly.tree.BaseNode.prototype.selectParent = function() {
  * Expand the current node if it's not already expanded, or select the
  * child node.
  * @return {boolean} True if the action has been handled, false otherwise.
- * @protected
+ * @package
  */
 Blockly.tree.BaseNode.prototype.selectChild = function() {
   if (this.hasChildren()) {
--- a/core/components/tree/basenode.js
+++ b/core/components/tree/basenode.js
@@ -1127,43 +1127,22 @@ Blockly.tree.BaseNode.prototype.onKeyDown = function(e) {
       if (e.altKey) {
         break;
       }
-      if (this.hasChildren()) {
-        if (!this.getExpanded()) {
-          this.setExpanded(true);
-        } else {
-          this.getFirstChild().select();
-        }
-      }
+      handled = this.selectChild();
       break;
 
     case Blockly.utils.KeyCodes.LEFT:
       if (e.altKey) {
         break;
       }
-      if (this.hasChildren() && this.getExpanded() && this.isUserCollapsible_) {
-        this.setExpanded(false);
-      } else {
-        var parent = this.getParent();
-        var tree = this.getTree();
-        // don't go to root if hidden
-        if (parent && (parent != tree)) {
-          parent.select();
-        }
-      }
+      handled = this.selectParent();
       break;
 
     case Blockly.utils.KeyCodes.DOWN:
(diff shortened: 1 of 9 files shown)


core/toolbox.js
```diff
@@ -648,9 +679,20 @@ Blockly.Toolbox.prototype.refreshSelection = function() {
   }
 };
 
+/**
+ * Select the first toolbox category if no category is selected.
+ * @package
```
alschmiedt: Just curious why this only selects the first category if no category is selected? Also could we change the name to selectFirstCategory?

core/components/tree/basenode.js
```diff
@@ -1166,6 +1156,70 @@ Blockly.tree.BaseNode.prototype.onKeyDown = function(e) {
   return handled;
 };
 
+
+/**
+ * Select the next node.
+ * @return {boolean} True if the field handled the action, false otherwise.
```
alschmiedt: Update the @return comments

core/toolbox.js
```diff
@@ -281,6 +286,32 @@ Blockly.Toolbox.prototype.handleNodeSizeChanged_ = function() {
   Blockly.svgResize(this.workspace_);
 };
 
+/**
+ * Handles the given Blockly action on a toolbox.
+ * This is only triggered when keyboard accessibility mode is enabled.
+ * @param {!Blockly.Action} action The action to be handled.
+ * @return {boolean} True if the field handled the action, false otherwise.
```
alschmiedt: Update @return comment

core/toolbox.js
samelhusseini: That's the desired behaviour from Blockly.navigation. If we're going from WS -> Toolbox, select first category. If from Flyout -> Toolbox, no need to select a category as there would have had to be a category selected for the flyout to be open.

samelhusseini: As for name change, it's not necessarily a "category" that we're selecting although that's how we're using it. But since this is on BaseNode, I'll try to keep it generic like the rest of the methods here.

core/components/tree/basenode.js
```diff
@@ -1166,6 +1156,70 @@ Blockly.tree.BaseNode.prototype.onKeyDown = function(e) {
...
+    }
+  }
+  return true;
+};
+
+/**
+ * Expand the current node if it's not already expanded, or select the
+ * child node.
+ * @return {boolean} True if the action has been handled, false otherwise.
+ * @protected
```
alschmiedt: should this be @package?

core/toolbox.js
alschmiedt: But this is on the Toolbox not BaseNode right? So the name could be more specific?

core/components/tree/basenode.js
samelhusseini: Yep!

core/toolbox.js
samelhusseini: Yes, you're right! Will fix.

PR #3055, opened 2019-09-19
Make toolbox an optional module.

NeilFraser:
This drops the compiled size by 25 kb if there are no categories.  Which also indicates that the toolbox itself suffers from really bad bloat.  The toolbox should not represent 1/6th of our codebase.

### DIFF
--- a/core/components/tree/basenode.js
+++ b/core/components/tree/basenode.js
@@ -178,13 +178,7 @@ Blockly.tree.BaseNode.prototype.initAccessibility = function() {
 
     var img = this.getIconElement();
     if (img) {
-      Blockly.utils.aria.setRole(img,
-          Blockly.utils.aria.Role.PRESENTATION + '1');
-    }
-    var ei = this.getExpandIconElement();
-    if (ei) {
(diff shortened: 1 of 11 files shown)


core/components/tree/basenode.js
```diff
@@ -892,22 +728,13 @@ Blockly.tree.BaseNode.prototype.getRowElement = function() {
...
-};
-
 /**
  * @return {Element} The icon element.
  * @protected
  */
 Blockly.tree.BaseNode.prototype.getIconElement = function() {
   var el = this.getRowElement();
-  return el ? /** @type {Element} */ (el.childNodes[1]) : null;
+  return el ? /** @type {Element} */ (el.firstChild) : null;
```
samelhusseini: This changes the DOM structure of the toolbox which might break some legacy CSS out there that assumes the icon element to be the second element.

NeilFraser: Given that the icon element has both a class and an id, I think that any reasonable developer would use one of those rather than child order.

PR #3057, opened 2019-09-20
Move populate for topRow and bottomRow to info

### LINKED ISSUE #2945: topRow.populate and bottomRow.populate belong in info

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2945 

### Proposed Changes

Refactors populate method from Row Measurable to RenderInfo.

### Reason for Changes

Other rows are populated in RenderInfo, so moving the logic for top and bottom row seems consistent.

### Test Coverage

Tested manually on playground.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -28,17 +28,21 @@ goog.provide('Blockly.blockRendering.RenderInfo');
 
 goog.require('Blockly.blockRendering.BottomRow');
 goog.require('Blockly.blockRendering.ExternalValueInput');
+goog.require('Blockly.blockRendering.Hat');
 goog.require('Blockly.blockRendering.InlineInput');
 goog.require('Blockly.blockRendering.InputRow');
 goog.require('Blockly.blockRendering.Measurable');
 goog.require('Blockly.blockRendering.NextConnection');
 goog.require('Blockly.blockRendering.OutputConnection');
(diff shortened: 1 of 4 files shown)


core/renderers/common/info.js
```diff
@@ -28,17 +28,21 @@ goog.provide('Blockly.blockRendering.RenderInfo');
...
 goog.require('Blockly.blockRendering.OutputConnection');
 goog.require('Blockly.blockRendering.PreviousConnection');
+goog.require('Blockly.blockRendering.RoundCorner');
 goog.require('Blockly.blockRendering.Row');
 goog.require('Blockly.blockRendering.SpacerRow');
 goog.require('Blockly.blockRendering.StatementInput');
+goog.require('Blockly.blockRendering.SquareCorner');
 goog.require('Blockly.blockRendering.TopRow');
 goog.require('Blockly.blockRendering.Types');
+goog.require('Blockly.BlockSvg');
```
samelhusseini: I think this introduces a circular dependency. I don't think rendering code should require ``Blockly.BlockSvg``. That constant being used should eventually just be moved into the rendering space.

PR #3060, opened 2019-09-20
Adding checks for whether to add elem spacer on row ends

### LINKED ISSUE #2964: Spacer calculation

In the new rendering code, rows have `getFirstSpacer` and `getLastSpacer` functions, but they're often wrong or make bad assumptions.

Rows should instead set `firstSpacer` and `lastSpacer` properties during `measure()`.

Fix the getter functions.

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2964 

### Proposed Changes

Adds functions to row to check whether spacer element should be added to start or end of row and removing adjustments in other places of code that assume that the last spacer is not the correct one.

### Reason for Changes

In certain cases, like with statement input, external input, and end of top/bottom row (zelos), the last spacer in the row is not actually a spacer that we want to modify and instead get the second to last spacer of that row. Not adding a spacer to the ends in some cases simplifies logic.

### Test Coverage

Tested on playground by checking that flyout blocks in different categories render correctly (especially for cases for statement input and external input) with all 3 renderer types.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/measurables/rows.js
+++ b/core/renderers/measurables/rows.js
@@ -185,6 +185,22 @@ Blockly.blockRendering.Row.prototype.getLastInput = function() {
 };
 
 /**
+ * Determines whether this row should start with an element spacer.
+ * @return {boolean} Whether the row should start with a spacer.
+ */
+Blockly.blockRendering.Row.prototype.startsWithElemSpacer = function() {
+  return true;
+};
+
+/**
+ * Determines whether this row should end with an element spacer.
+ * @return {boolean} Whether the row should end with a spacer.
+ */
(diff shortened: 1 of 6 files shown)


core/renderers/measurables/rows.js
```diff
@@ -184,6 +184,22 @@ Blockly.blockRendering.Row.prototype.getLastInput = function() {
   return null;
 };
 
+/**
+ * Determines whether this row should start with an element spacer.
+ * @return {boolean} Whether the row should start with a spacer.
+ */
```
samelhusseini: @package

moniika: Done.

PR #3065, opened 2019-09-20
Changed variable field to use configuration

### LINKED ISSUE #2722: Standardize Field Configuration

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #2722

### Proposed Changes

Adds an opt_config option to the variable field.

I added the option to configure types using the opt_config, but I did it on a separate commit so it can be easily reverted.

The initial commit just adds an opt_config property so variables can support tooltips.

### Reason for Changes

### Test Coverage

Added tests for configuring types using different methods.

And for the initial commit I manually tested setting tooltips via json.

Tested on:
* Desktop Chrome

### Documentation

If you want to keep configuring types via the opt_config the variable field's [creation ](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/variable#creation)section may need updates.

### Additional Information

Configuring types via opt_config is on a separate commit, so it'll be easy to revert if you don't want it.

Also: Woot! this is the last field!

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -75,7 +75,7 @@ Blockly.FieldVariable = function(varName, opt_validator, opt_variableTypes,
    * The initial variable name passed to this field's constructor, or an
    * empty string if a name wasn't provided. Used to create the initial
    * variable.
-   * @type {!string}
+   * @type {string}
    */
   this.defaultVariableName = varName || '';
 
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -90,7 +90,9 @@ Blockly.FieldVariable = function(varName, opt_validator, opt_variableTypes,
   opt_config && this.configure_(opt_config);
   opt_validator && this.setValidator(opt_validator);
 
-  this.setTypes_(opt_variableTypes, opt_defaultType);
+  if (!opt_config) {  // Only do one kind of configuration or the other.
+    this.setTypes_(opt_variableTypes, opt_defaultType);
+  }
 };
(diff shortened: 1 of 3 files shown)


core/field_variable.js
```diff
@@ -51,19 +51,33 @@ goog.require('Blockly.Xml');
...
-  opt_validator && this.setValidator(opt_validator);
-  this.defaultVariableName = varname || '';
 
-  this.setTypes_(opt_variableTypes, opt_defaultType);
-  this.value_ = null;
+  /**
+   * The initial variable name passed to this field's constructor, or an
+   * empty string if a name wasn't provided. Used to create the initial
+   * variable.
+   * @type {!string}
```
samelhusseini: string is implicitly not null, so no need for ``!``.

PR #3068, opened 2019-09-21
Strip out more bloat from toolbox

NeilFraser:
Also removes three private accesses from workspaceSvg into toolbox tree.

The strategy for workspace.updateToolbox is no longer to clear and repopulate the toolbox, but rather to throw away the old toolbox and rebuild a new one.  This is simpler and more reliable.

This commit trims off another KB from the compiled code.

### DIFF
--- a/core/components/component.js
+++ b/core/components/component.js
@@ -47,7 +47,7 @@ Blockly.Component = function() {
    * been set by calling {@link #setRightToLeft} explicitly.
    * @private {?boolean}
    */
-  this.rightToLeft_ = Blockly.Component.defaultRightToLeft_;
+  this.rightToLeft_ = Blockly.Component.defaultRightToLeft;
 
   /**
    * Unique ID of the component, lazily initialized in {@link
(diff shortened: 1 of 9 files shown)


core/css.js
```diff
@@ -136,7 +136,6 @@ Blockly.Css.CONTENT = [
 
   '.blocklyNonSelectable {',
     'user-select: none;',
-    '-moz-user-select: none;',
```
NeilFraser: FYI: The current version of Mozilla now supports `user-select` without the `-moz` prefix.  And previous versions of Mozilla started supporting `-webkit-user-select` (!).  So these lines aren't needed any more.

core/components/component.js
```diff
@@ -77,42 +77,36 @@ Blockly.Component = function() {
...
-  this.childIndex_ = null;
+  this.childIndex_ = {};
 };
 
 
 /**
  * The default right to left value.
- * @type {?boolean}
- * @private
+ * @type {boolean}
```
samelhusseini: Should this be @package?

core/css.js
samelhusseini: Should we hold off on this one until the next release? Firefox 68 still has 4% market share in September 2019 according to https://www.stetic.com/market-share/browser/

NeilFraser: Public support in Firefox for `-webkit-user-select` (which we are keeping) was added in version 49.  That's from 2016.
https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

Additionally, the failure mode is benign, it's just the ability to select text that probably shouldn't be selectable.

core/components/component.js
NeilFraser: Good call.  Done.

core/css.js
samelhusseini: Sounds good.

core/toolbox.js
```diff
@@ -198,15 +198,37 @@ Blockly.Toolbox.prototype.init = function() {
   this.config_['cleardotPath'] = workspace.options.pathToMedia + '1x1.gif';
   this.config_['cssCollapsedFolderIcon'] =
       'blocklyTreeIconClosed' + (workspace.RTL ? 'Rtl' : 'Ltr');
+  this.renderTree(workspace.options.languageTree);
+};
+
+/**
+ * Fill the toolbox with categories and blocks.
+ * @param {!Node} languageTree DOM tree of blocks.
+ */
```
samelhusseini: Just checking if this is intended to be a public API?

NeilFraser: Now @package

PR #3070, opened 2019-09-23
Fix interaction between mouse and keyboard

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
1. If a user tried to use both the keyboard and the mouse for keyboard navigation the states wouldn't be in sync and everything would break. 
2. Hitting out on a flyout without a toolbox was causing an error.

### Proposed Changes
Keep the navigation state updated when the user moves around using the mouse.
Fixed Scenarios: 
1. User clicks on the toolbox and then uses keyboard navigation to move around.
2. User is in the toolbox using keyboard navigation and then clicks on the workspace. 
3. User is in the flyout using keyboard navigation and then drags a block to the workspace. 
4. User is in the toolbox using keyboard navigation and then clicks on the category to close the toolbox.

### Test Coverage

Tested on:

### DIFF
--- a/core/gesture.js
+++ b/core/gesture.js
@@ -663,6 +663,9 @@ Blockly.Gesture.prototype.handleWsStart = function(e, ws) {
   this.setStartWorkspace_(ws);
   this.mostRecentEvent_ = e;
   this.doStart(e);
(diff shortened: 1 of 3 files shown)


alschmiedt:
Fixes #3071

PR #3073, opened 2019-09-23
Disconnect child block always for block.dispose(/* healStack */ true)

### LINKED ISSUE #2991: block.dispose(/* healStack */ true) does not properly handle rows
**Describe the bug**

If you run block.dispose(true) on a "row" block, and that row block's next block is not compatible with the row block's parent, it gets destroyed instead of disconnected.

**To Reproduce**

1. Load the following XML to the playground:
```
  <block type="math_trig">
    <field name="OP">SIN</field>
    <value name="NUM">
      <shadow type="math_number">
        <field name="NUM">0</field>
      </shadow>
      <block type="math_on_list" id="toDispose">
        <mutation op="SUM"/>
        <field name="OP">SUM</field>
        <value name="LIST">
          <block type="lists_create_with">
            <mutation items="3"/>
          </block>
        </value>
      </block>
    </value>
  </block>
</xml>
```
2. Run the following code in the console:
```
Blockly.mainWorkspace.getBlockById('toDispose').dispose(true);
```
3. Observe how both the "math_on_list" block and the "lists_create_with" block are disposed.

**Expected behavior**

The "list_create_with" block should survive, but be disconnected.

**This is not browser specific**

**Additional context**

This does not occure with statement blocks.

This occures because [unplugFromRow_](https://github.com/google/blockly/blob/4efa0dad1f2c882234759d3ba7ecd0ebd95a4c77/core/block.js#L365) does not work as described:
> Unplug this block's output from an input on another block.  Optionally reconnect the block's parent to the only child block, if possible.

There is [a comment](https://github.com/google/blockly/blob/4efa0dad1f2c882234759d3ba7ecd0ebd95a4c77/core/block.js#L393) later in that function which reads:
> Only disconnect the child if it's possible to move it to the parent.

Which is not the desired behavior.

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixed #2991 

### Proposed Changes
Always disconnects child block with healStack true in unplugFromRow_. Bumps child block away if it cannot be connected.

1d1fb01e-2a02-41a9-b04a-70312608cb26

### Reason for Changes

Desired behavior.

### Test Coverage

Blockly playground

Tested on:
 * Desktop Chrome

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -417,7 +417,7 @@ Blockly.Block.prototype.unplugFromRow_ = function(opt_healStack) {
   if (childConnection.checkType_(parentConnection)) {
     parentConnection.connect(childConnection);
   } else {
-    childConnection.bumpAwayFrom_(parentConnection);
+    childConnection.onFailedConnect(parentConnection);
   }
 };
 
--- a/core/block.js
+++ b/core/block.js
@@ -413,9 +413,11 @@ Blockly.Block.prototype.unplugFromRow_ = function(opt_healStack) {
   var childConnection = thisConnection.targetConnection;
   // Disconnect the child block.
   childConnection.disconnect();
-  // Move the child to the parent, if possible.
+  // Connect child to the parent if possible, otherwise bump away.
   if (childConnection.checkType_(parentConnection)) {
     parentConnection.connect(childConnection);
+  } else {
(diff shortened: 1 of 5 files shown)


RoboErikG:
Fix travis then lgtm

PR #3075, opened 2019-09-23
Add a test for built file size

### LINKED ISSUE #3052: Build a test for checking built size
There are two things that need to be done for this.

First is just building a test for Blockly's compiled size and gzipped size so we can track the overall size of the library over time.

Second, we now have a way to remove optional components from Blockly at build time via the [requires.js](https://github.com/google/blockly/blob/develop/core/requires.js) file. We should build a test that generates the bytes saved by removing each component so developers can decide which components are worth using for their application.

See #3023 for a partial list of components that may be optional over time.

RoboErikG:
Adds a script test for building Blockly's compressed files and
checking their size to prevent unexpected increases.

## The basics

- [x] I branched from develop

### Resolves

Fixes #3052 

### Proposed Changes

Adds a script test for building Blockly and comparing the size against a previous value. This test will need to be updated each release. We could also compare against the files currently on the master branch, but we may want to manually update the size when appropriate instead of allowing continuous growth over multiple releases.

### Test Coverage

Tested on:

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/3075)

### DIFF
--- a/tests/run_all_tests.sh
+++ b/tests/run_all_tests.sh
@@ -67,7 +67,7 @@ run_test_command "generators" "tests/scripts/run_generators.sh"
 # Generate TypeScript typings and ensure there are no errors.
 run_test_command "typings" "tests/scripts/compile_typings.sh"
 
-# Check the sizes of built files for unexpected growth
+# Check the sizes of built files for unexpected growth.
 run_test_command "metadata" "tests/scripts/check_metadata.sh"
 
 # # Attempt advanced compilation of a Blockly app.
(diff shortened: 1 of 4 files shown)


RoboErikG:
Note: Sam says the package-lock file should be checked in, but we were missing it before.

tests/run_all_tests.sh
```diff
@@ -67,6 +67,9 @@ run_test_command "generators" "tests/scripts/run_generators.sh"
 # Generate TypeScript typings and ensure there are no errors.
 run_test_command "typings" "tests/scripts/compile_typings.sh"
 
+# Check the sizes of built files for unexpected growth
```
samelhusseini: nit: end comment with full stop.

Issue #2642, opened 2019-07-15
Angle Field Doesn't Handle Touch Input Correctly


BeksOmega:
### Problem statement

It seems like it's not capturing the input correctly so it's allowing scrolls? Not sure.

### Expected Behavior

I should be able to drag the angle picker without it getting "stuck"

### Actual Behavior

The angle picker gets stuck.
![AngleFieldInput](https://user-images.githubusercontent.com/25440652/61251321-f139f980-a70e-11e9-8b90-b701ea814920.gif)

### Steps to Reproduce

1. Load up an angle picker on mobile (preferably the blockly developer tools).
2. Try to drag it.
3. Observe how it sometimes drags the window instead of dragging the angle picker.

### Operating System and Browser

* Tablet:
  * Device: Samsung Galaxy Tab S2
  * OS: Android 7
  * Browser: Chrome
  * Version: 75

### Additional Information

Maybe related to #2316 ?

alschmiedt:
I am not having this problem when I tested on safari or chrome for ios. However, tested on android and can confirm it is getting stuck.

RoboErikG:
Note, Chrome 75 for Android came out on June 5th this year. We may be getting hit with some changes to gesture handling in Chrome itself.

RoboErikG:
Possibly related to https://bugs.chromium.org/p/chromium/issues/detail?id=974823

samelhusseini:
Fixed.

PR #3079, opened 2019-09-24
Angle field touch fixes

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/2642

### Proposed Changes

There were a number of issues with the angle picker on touch devices.
- The primary fix for the bug was that default browser touch events were being handled on the drag surface that were cancelling the pointer event. This is resolved by adding the CSS ``touch-action: none``. More info [here](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) and [here](https://docs.w3cub.com/dom_events/pointercancel/)
- on IOS, which uses touch events, bindEvents_ was blocking a click event from bubbling, so it wasn't being closed on a click
- I couldn't find a reason to have the mouse events set on the gauge as well as the circle (drag surface), so I removed that and disabled pointer events on the gauge.
- I added a click handler on the drag surface for touch devices to do a final update on click, as on touch devices, the value is only updated on drag. (added comment).

### Reason for Changes

Fix angle picker on touch devices (IOS and Android).

### Test Coverage

Playground

Tested on:
* Desktop Chrome
* Desktop Chrome with mobile emulator.
* Desktop Safari
* IOS Safari
* Android Chrome


RoboErikG:
This lgtm, but @rachel-fenichel to double check

core/field_angle.js
```diff
@@ -280,15 +281,16 @@ Blockly.FieldAngle.prototype.dropdownCreate_ = function() {
...
-  // bindEventWithChecks_ allows it to work without a mousedown/touchstart.
+  // change this behaviour.
   this.clickWrapper_ =
-      Blockly.bindEvent_(svg, 'click', this, this.hide_);
-  this.moveWrapper1_ =
-      Blockly.bindEvent_(circle, 'mousemove', this, this.onMouseMove);
-  this.moveWrapper2_ =
-      Blockly.bindEvent_(this.gauge_, 'mousemove', this, this.onMouseMove);
-
+      Blockly.bindEventWithChecks_(svg, 'click', this, this.hide_);
```
rachel-fenichel: Does the change to bindEventWithChecks work for the case of changing it on mousemove without a drag?

samelhusseini: Yes, when using mouseevents / pointerevents.

samelhusseini: But I assume you're talking about the 'mousemove' two lines below? This one just hides on click.

rachel-fenichel: That's right.  The comment above references it as well.

PR #3080, opened 2019-09-24
Removing require for Blockly.FlyoutButton

NeilFraser:
Should have been in the previous PR, since that’s the point.

Also, a recompile given all the recent dependency changes.


core/flyout_base.js
```diff
@@ -31,7 +31,6 @@ goog.require('Blockly.blockRendering');
 goog.require('Blockly.Events');
 goog.require('Blockly.Events.BlockCreate');
 goog.require('Blockly.Events.VarCreate');
-goog.require('Blockly.FlyoutButton');
```
NeilFraser: This is the only functional change in this PR.

PR #3082, opened 2019-09-25
Code cleanup, drop some CSS.

NeilFraser:
Random improvements to require statements.

Removal of some obviously dead CSS.

.blocklyDropDownDiv no longer suppresses text selection (see comment in PR).

This PR drops the compiled size by 4 KB, which is larger than I was expecting.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -131,7 +131,7 @@ Blockly.Css.CONTENT = [
     'height: 100%;',
     'position: relative;',
     'overflow: hidden;', /* So blocks in drag surface disappear at edges */
-    'touch-action: none',
+    'touch-action: none;',
   '}',
 
   '.blocklyNonSelectable {',
@@ -192,7 +192,6 @@ Blockly.Css.CONTENT = [
     'border: 1px solid;',
     'border-radius: 2px;',
     'padding: 4px;',
-    '-webkit-user-select: none;',
     'box-shadow: 0px 0px 3px 1px rgba(0,0,0,.3);',
   '}',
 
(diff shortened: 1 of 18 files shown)


core/css.js
```diff
@@ -572,10 +571,6 @@ Blockly.Css.CONTENT = [
     'fill-opacity: .8;',
   '}',
 
-  '.blocklyTransparentBackground {',
```
NeilFraser: No idea what this was, appears orphaned.

core/css.js
```diff
@@ -696,9 +691,6 @@ Blockly.Css.CONTENT = [
     'overflow-x: visible;',
     'overflow-y: auto;',
     'position: absolute;',
-    'user-select: none;',
-    '-ms-user-select: none;',
-    '-webkit-user-select: none;',
```
NeilFraser: Added blocklyNonSelectable to toolbox class, so this isn't needed any more.

core/css.js
```diff
@@ -192,7 +192,6 @@ Blockly.Css.CONTENT = [
     'border: 1px solid;',
     'border-radius: 2px;',
     'padding: 4px;',
-    '-webkit-user-select: none;',
```
NeilFraser: This line is the only functional change.  First, it's broken, only applying to webkit.  Second, it is inappropriate for the dropdowndiv to declare that any contents be unselectable; we have no idea what's in dropdowndiv.  Currently all content in existing dropdowndivs is non-text, with the exception of the Turtle demo.  I've added blocklyNonSelectable to that demo.

samelhusseini:
I only saw a reduction of < 1KB for this one if compared to ``develop``.
I agree a 4KB is larger than expected for this.

PR #3085, opened 2019-09-25
Updating behavior of register functions and adding unregister.

### LINKED ISSUE #2788: Overwriting a JSON Key is Not Consistent

Overwriting a JSON key does not have a consistent feedback mechanism.

[Blockly.defineBlocksWithJsonArray](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/core/blockly.js#L423) logs a warning.
[Blockly.Extensions.register](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/core/extensions.js#L56) throws an error.
[Blockly.Field.register](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/core/field.js#L72) does nothing.

I prefer the log a warning option.

* It was stated explicitly on the [draggable number field issue](https://github.com/google/blockly/issues/1701#issuecomment-385534439) that overriding fields was an option, so I think that should be kept.
* I also wished I could override extensions when I was working on an [alternative mutators demo](https://github.com/google/blockly/issues/1845) recently, so I think that should be allowed.
* I think a similar principle applies to blocks.

* Throwing errors. This is annoying, if I want to override something I think I should be allowed to.
* Not doing anything. This would be fine... but I don't mind the extra "hey be careful there" of a warning either.

**Additional context**

[Applicable field registry test](https://github.com/google/blockly/blob/33cce3c8693dfd2d9a52f1d9039e4ef4eabb7250/tests/mocha/field_registry_test.js#L61)

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2788

### Proposed Changes

Adds unregister methods for extensions, fields, and renderers.
Changes field register to throw error if another is registered with same name.

### Reason for Changes

Makes the behavior of register functions more consistent with extensions, fields, and renderer all throwing an error when something is registered with same name as existing and adding an unregister functionality for all of them.

### Test Coverage

Ran tests.

### DIFF
--- a/core/field_registry.js
+++ b/core/field_registry.js
@@ -38,7 +38,7 @@ goog.provide('Blockly.fieldRegistry');
 Blockly.fieldRegistry.typeMap_ = {};
 
 /**
- * Registers a field type. May also override an existing field type.
+ * Registers a field type.
  * Blockly.fieldRegistry.fromJson uses this registry to
  * find the appropriate field type.
  * @param {string} type The field type name as used in the JSON definition.
@@ -64,9 +64,7 @@ Blockly.fieldRegistry.register = function(type, fieldClass) {
 };
 
 /**
- * Registers a field type. May also override an existing field type.
- * Blockly.fieldRegistry.fromJson uses this registry to
- * find the appropriate field type.
+ * Unregisters the field registered with the given type.
  * @param {string} type The field type name as used in the JSON definition.
  */
 Blockly.fieldRegistry.unregister = function(type) {
--- a/core/field_registry.js
+++ b/core/field_registry.js
@@ -44,19 +44,38 @@ Blockly.fieldRegistry.typeMap_ = {};
  * @param {string} type The field type name as used in the JSON definition.
  * @param {!{fromJson: Function}} fieldClass The field class containing a
  *     fromJson function that can construct an instance of the field.
- * @throws {Error} if the type name is empty, or the fieldClass is not an
- *     object containing a fromJson function.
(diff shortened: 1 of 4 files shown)


core/field_registry.js
```diff
@@ -44,21 +44,40 @@ Blockly.fieldRegistry.typeMap_ = {};
...
+  }
   if (!fieldClass || (typeof fieldClass.fromJson != 'function')) {
     throw Error('Field "' + fieldClass + '" must have a fromJson function');
   }
   type = type.toLowerCase();
   Blockly.fieldRegistry.typeMap_[type] = fieldClass;
 };
 
+/**
+ * Registers a field type. May also override an existing field type.
```
RoboErikG: s/Registers/Unregisters/

core/field_registry.js
```diff
@@ -44,21 +44,40 @@ Blockly.fieldRegistry.typeMap_ = {};
  * @param {string} type The field type name as used in the JSON definition.
  * @param {!{fromJson: Function}} fieldClass The field class containing a
  *     fromJson function that can construct an instance of the field.
- * @throws {Error} if the type name is empty, or the fieldClass is not an
- *     object containing a fromJson function.
+ * @throws {Error} if the type name is empty, the field is already
```
RoboErikG: Update method description, no longer allows overriding an existing field type without unregistering first.

core/field_registry.js
moniika: Done

PR #3086, opened 2019-09-25
Remove @license tag

NeilFraser:
The @license tag is used by only one tool, the Closure Compiler.  It uses this tag to find licenses so that it may preserve them in the compiled output — something we actively don’t want.  This forces us to add an epic regex hack to the build to strip out these licenses.

A simpler solution is just to remove these tags in the first place.  We still have a root-level LICENSE file that’s used by project-level analysis (including GitHub).

This also drops the compiled test file by 3 KB.


NeilFraser:
This is a speculative PR, what do other people think?

If approved, then I'll do the same over on Blockly Games.

NeilFraser:
Lawyer says no.

PR #3074, opened 2019-09-23
Cleanup navigation file

alschmiedt:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves
Comments and organization of the navigation file was out of date. 

### Proposed Changes
Group similar methods together and add comments.

### Test Coverage

Tested on:

PR #3091, opened 2019-09-25
Partial code review of nav

NeilFraser:
Had a few minutes free.

There are also some post-submit comments on https://github.com/google/blockly/pull/3074

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -31,12 +31,12 @@ goog.provide('Blockly.ASTNode');
  * Class for an AST node.
  * It is recommended that you use one of the createNode methods instead of
  * creating a node directly.
- * @constructor
  * @param {string} type The type of the location.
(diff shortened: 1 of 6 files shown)


core/keyboard_nav/ast_node.js
```diff
@@ -491,26 +490,6 @@ Blockly.ASTNode.prototype.findFirstFieldOrInput_ = function(block) {
   return null;
 };
 
-/**
- * Walk backwards from the given block up through the stack of blocks to find
- * the top block of the sub stack. If we are nested in a statement input only
- * find the top most nested block. Do not go all the way to the top of the
```
rachel-fenichel: This says it doesn't go all the way to the top of the stack, unlike `getRootBlock`.

NeilFraser: Yup, caught that and then fell down the rabbit hole of trying to figure out what the neighbouring `getFirstStatementConnection` was for.

Issue #2733, opened 2019-07-29
Themes: Background Colour


alschmiedt:
**Is your feature request related to a problem? Please describe.**
For some people the white background of the workspace can be uncomfortable to look at. ([section 4.4](https://ksiresearchorg.ipage.com/vlss/journal/VLSS2017/vlss-2017-ludi-spencer.pdf))

**Describe the solution you'd like**
As a part of [themes](https://developers.google.com/blockly/guides/configure/web/themes) add a way for developers to change the workspace color. 

**Describe alternatives you've considered**

**Additional context**
This would probably involve adding a new optional parameter to the Theme constructor. This is currently a CSS value so we would need to find a way to overwrite or change this when we change themes.

hilts-vaughan:
Sign me up, why not? :)

hilts-vaughan:
Proposing the following:

1. Add the parameters to the constructor, expose trivial config
2. In the SVG renderer, do some refactoring of the class option and instead make it a "style specifier" which is a string that is either a. a color b. a class name. We will introduce a new util that will check if the string is a color and use that. This type of overloading is pretty typical, for example: https://phaser.io/docs/2.6.2/Phaser.Stage.html#backgroundColor

This should not break the API and give enough flexibility. Thoughts? And of course, tests.

Open q's:

1. Are you guys writing tests in Mocha now?
2. Should we specify an explicit value on the themes in source control? It would probably serve to document them better.

alschmiedt:
Hi there!
This sounds like a good plan. 
1. We still need to include the styles in the blocklySvg class. Are you proposing leaving the class and just overwriting the background-color attribute? 
2. Yes we are testing in Mocha right now!
3. Yes we should add a value to the example themes. Themes are documented here: https://developers.google.com/blockly/guides/configure/web/themes but feel free to add more comments and documentation in the example themes if you think it would be useful : )

hilts-vaughan:
1. More like the fill, since it's an SVG attribute, but yes!

2, 3: Sounds good :)

alschmiedt:
I'm not sure how setting the fill will work with our current setup. 
Currently we are using the background-color found here: https://github.com/google/blockly/blob/master/core/css.js#L117 to control the background color of the workspace.
I think this has to do with how we are creating the background grid.

hilts-vaughan:
Thanks for the info. I'll do a bit more search before I start on the actual implementation. Are you open to draft PRs to get feedback on the design? I would be a first time contributor.

[...]

PR #2605, opened 2019-07-01
Fixes theme error for headless workspaces

alschmiedt:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves
#2575 & #2540

### Proposed Changes
Set a default theme in the workspace constructor.

### Reason for Changes
We were getting an error when creating a block on a headless workspace because the theme was not set. This is because we only set the default theme in Blockly.Inject which does not get called for a headless workspace.

### Test Coverage

Tested on:

PR #3093, opened 2019-09-26
Workspace theme

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/2733

### Proposed Changes

Move the theme so it's attached to the parent workspace (and no longer a global). This allows us to have multiple instances of Blockly with different themes.

Created a new object ``Blockly.ThemeManager`` that stores the current workspace theme and is in charge of registering subscribed workspaces, and registering subscribed UI components. 

Subscribed workspaces are refreshed when the theme changes.
Throughout our codebase we can also register Blockly UI components with the theme, letting the theme manager know what component name to use, and what style property to set to. Whenever the theme changes, the UI component is updated with the new value for the component name it's interested in.

This happens at runtime, so developers are able to switch themes, and have all the styles reflected.

Added Dark theme for testing.

### Reason for Changes

No way to configure UI component styles.

### Test Coverage

Screen Shot 2019-09-25 at 5 31 57 PM

Testing with Dark Theme.
Tested playground and multi-playground.
All tests pass.

Tested on:
* Desktop Chrome
* Desktop Safari

### Documentation

Need to add documentation for how to configure themes.


core/workspace.js
```diff
@@ -198,16 +203,74 @@ Blockly.Workspace.prototype.getMarker = function() {
...
+
+  var event = new Blockly.Events.Ui(null, 'theme', null, null);
+  event.workspaceId = this.id;
+  Blockly.Events.fire(event);
+};
+
+/**
+ * Updates all the blocks with new style.
+ * @param {!Array.<!Blockly.Block>} blocks List of blocks to update the style
+ * on.
```
alschmiedt: Indent

core/toolbox.js
```diff
@@ -175,7 +178,8 @@ Blockly.Toolbox.prototype.init = function() {
     oneBasedIndex: workspace.options.oneBasedIndex,
     horizontalLayout: workspace.horizontalLayout,
     toolboxPosition: workspace.options.toolboxPosition,
-    renderer: workspace.options.renderer
+    renderer: workspace.options.renderer,
+    theme: workspace.options.theme
```
alschmiedt: Do we need this here? I think the workspace is just going to end up using the parent's themeManager

core/block.js
```diff
@@ -990,10 +990,10 @@ Blockly.Block.prototype.setColour = function(colour) {
  * @throws {Error} if the block style does not exist.
  */
 Blockly.Block.prototype.setStyle = function(blockStyleName) {
-  var theme = Blockly.getTheme();
+  var theme = this.workspace.getTheme();
   if (!theme) {
```
alschmiedt: Do we need this check anymore? Since theme should never be able to be null?

core/workspace.js
```diff
@@ -131,11 +131,16 @@ Blockly.Workspace = function(opt_options) {
...
-    Blockly.setTheme(Blockly.Themes.Classic);
-  }
+  /**
+   * Object in charge of storing and updating the workspace theme.
+   * @type {!Blockly.ThemeManager}
+   * @protected
+   */
+  this.themeManager_ = this.options.parentWorkspace ?
+      this.options.parentWorkspace.getThemeManager() :
+      new Blockly.ThemeManager(this.options.theme);
```
alschmiedt: Should we set a default theme here if there are none are on the options? There was a bug that the code we deleted fixed for headless workspaces. Issue #2605

core/toolbox.js
samelhusseini: Yep, I got rid of all the others, this one snuck through.

core/workspace.js
samelhusseini: Yep, agreed.

core/block.js
samelhusseini: Removing.

PR #3101, opened 2019-09-26
Make dynamic variables be an optional module.

NeilFraser:
Also straighten out some dependencies.


NeilFraser:
This reduces the compile test by 1 KB.
Though 3 KB have been added since yesterday, so the net difference is -2 KB.

rachel-fenichel:
That 3 KB number doesn't make sense.  Here's what;s landed in the last day:
image

NeilFraser:
Ah, you're right Rachel.  The 3 KB came from switching of the as-yet unmerged fraser-license branch, where #3086 dropped the compiled version by that amount.  My mistake.

Issue #3023, opened 2019-09-16
Modularlize more of Blockly


samelhusseini:
Using this issue to track work for splitting out Blockly into more modular components. Renderers and fields are already fairly modular, we'll need to run a cost-benefit analysis of how easy it is to modularize more of Blockly. 
Some candidates: 
- Toolbox
- Workspace comments
- Generators
- Variables
- Procedures
- Grid
- Mutators

RoboErikG:
Note on the help wanted tag, some of these should be easy like #3053 but not all of them.

NeilFraser:
Related PRs so far:
* Trash can and zoom controls: #3043
* Mutators: #3048
* Block comments: #3053
* Toolbox: #3055
* Flyout Buttons: #3076 and #3080
* CSS: #3084
* Dynamic Variables: #3100

cpcallen:
Most or all of these (notably including `Blockly.Trashcan`*) have ceased to be optional due to the work done to re-export `Blockly.*` in `core/blockly.js`.  🤔

(* Noted while investigating issues with building Blockly Games.)

NeilFraser:
Unchecking all the progress we've made (14/17) due to the TypeScript migration.  We're back to square one.

rachel-fenichel:
Hang on, I consider it still modularized if it's registrable. There are two parts we're talking about:
- Make pieces registrable
- Make pieces removable

The first one focuses on modularization for API boundaries, and we've made a lot of progress there. The second one focuses on size.

PR #3105, opened 2019-09-26
Make generator an optional module

NeilFraser:
Issue #3023

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -56,7 +56,8 @@ goog.require('Blockly.Workspace');
  * @throws When block is not valid or block name is not allowed.
  */
 Blockly.Block = function(workspace, prototypeName, opt_id) {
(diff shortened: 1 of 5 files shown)


NeilFraser:
This has virtually no effect on size, since if there's no generator, the contents of the generator file is tree-shaken out.

More interestingly, the code is technically buggy and non-deterministic.
`typeof Blockly.Generator.prototype[prototypeName]`
In this case `prototypeName` is a block name, like `"logic_boolean"`, and the properties on `Blockly.Generator.prototype` are methods like `workspaceToCode`.  However, during compilation, `workspaceToCode` will get compiled down to `d`, `$2`, or whatever the compiler currently feels like.  If the compiler ends up coincidentally choosing `logic_boolean` or one of the other block names, then that compiled Blockly instance will start failing.  Don't know how to fix this.  But not something I'm worried about.

PR #3108, opened 2019-09-26
Fix bugs

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
1. Null pointer exception when we have an output connection marked and we try inserting a block with a dummy input.
2. Demo used to not allow traversing to the workspace.
3. Cursor was skipping output connection // left over from our previous cursor
4. Toolbox was closing after we navigated to a tree separator node. This would put us into the workspace navigation state, making it impossible to reach the nodes below it.
5. If an output connection is marked we should look at all possible ways to connect the inserted block. This includes the output connection. 
Example: When we have the below marked connection and we try inserting a number block it should connect.
Screen Shot 2019-09-26 at 3 43 05 PM

### Proposed Changes
1. Check that inputConnection is not null.
2. Add workspace type as a valid node for tree traversal.
3. Make it not skip the output connection
4. Not ideal, but if the newnode is a tree separator then don't put us in workspace navigation mode. I think ideally we would not close the flyout since that should signify being back on the workspace, but don't want to mess up current behavior.
5. If we have a marked output connection check the inserted block for all input connections. If none exist then try connecting the block's output connection with the destination connection. The connect_ function does more logic around connecting and will end up connecting to the destination's target connection if it exists.

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox.js
+++ b/core/toolbox.js
@@ -293,7 +293,8 @@ Blockly.Toolbox.prototype.handleAfterTreeSelected_ = function(
   } else {
     // Hide the flyout.
     this.flyout_.hide();
-    if (Blockly.keyboardAccessibilityMode && !(newNode instanceof Blockly.Toolbox.TreeSeparator)) {
+    if (Blockly.keyboardAccessibilityMode &&
+        !(newNode instanceof Blockly.Toolbox.TreeSeparator)) {
       Blockly.navigation.setState(Blockly.navigation.STATE_WS);
     }
   }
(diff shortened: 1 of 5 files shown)


core/toolbox.js
```diff
@@ -293,7 +293,7 @@ Blockly.Toolbox.prototype.handleAfterTreeSelected_ = function(
   } else {
     // Hide the flyout.
     this.flyout_.hide();
-    if (Blockly.keyboardAccessibilityMode) {
+    if (Blockly.keyboardAccessibilityMode && !(newNode instanceof Blockly.Toolbox.TreeSeparator)) {
       Blockly.navigation.setState(Blockly.navigation.STATE_WS);
```
samelhusseini: 99 chars.

PR #3109, opened 2019-09-26
Fixed field editor disposal

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Sort of #2152 - but here we're actually adding places where the property /should/ be set to null.

### Proposed Changes

Makes it so the angle field and text input field properly dispose of their editors.

Also reorganized the text input field's disposal a bit.
Also also did some picky doc changes.

### Reason for Changes

Detached nodes are bad.

Cleaner code.
I can be picky!

### Test Coverage

No functional changes.

### Documentation

Actually yes, [Creating an editor > Cleaning up](https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating#cleaning_up) should tell developers to release any dom references (my bad).

Tell me if you want me to file a separate issue.

### Additional Information

N/A

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -103,6 +103,7 @@ Blockly.DropDownDiv.PADDING_Y = 16;
 Blockly.DropDownDiv.ANIMATION_TIME = 0.25;
 
 /**
+<<<<<<< HEAD
  * The default dropdown div border color.
  * @type {string}
  * @const
@@ -115,6 +116,14 @@ Blockly.DropDownDiv.DEFAULT_DROPDOWN_BORDER_COLOR = '#dadce0';
  * @const
  */
 Blockly.DropDownDiv.DEFAULT_DROPDOWN_COLOR = '#fff';
+
+/**
+ * Transition animation for transform and opacity.
+ * @const {string}
+ */
+Blockly.DropDownDiv.TRANSITION = 'transform ' +
+    Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
+    'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
 
 /**
  * Timer for animation out, to be cleared if we need to immediately hide
@@ -154,11 +163,7 @@ Blockly.DropDownDiv.createDom = function() {
   Blockly.DropDownDiv.arrow_ = arrow;
 
   Blockly.DropDownDiv.DIV_.style.opacity = 0;
-
-  // Transition animation for transform: translate() and opacity.
-  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' +
-    Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
-    'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
+  Blockly.DropDownDiv.DIV_.style.transition = Blockly.DropDownDiv.TRANSITION;
 
   // Handle focusin/out events to add a visual indicator when
   // a child is focused or blurred.
@@ -192,7 +197,23 @@ Blockly.DropDownDiv.getContentDiv = function() {
  */
 Blockly.DropDownDiv.clearContent = function() {
   Blockly.DropDownDiv.content_.innerHTML = '';
-  Blockly.DropDownDiv.content_.style.width = '';
+  Blockly.DropDownDiv.content_.style.cssText = '';
+};
+
+/**
+ * Resets the CSS of the div.
+ * @param {boolean} persistDisplay True to persist the display mode,
+ *    otherwise it will be set to 'none'.
+ */
+Blockly.DropDownDiv.resetCss = function(persistDisplay) {
+  var style = Blockly.DropDownDiv.DIV_.style;
+  var displayMode = style.display;
+  style.cssText = '';
+  style.opacity = 0;
+  style.transition = Blockly.DropDownDiv.TRANSITION;
+  style.backgroundColor = Blockly.DropDownDiv.DEFAULT_DROPDOWN_COLOR;
+  style.borderColor = Blockly.DropDownDiv.DEFAULT_DROPDOWN_BORDER_COLOR;
+  style.display = persistDisplay ? displayMode : 'none';
 };
 
 /**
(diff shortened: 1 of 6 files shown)


core/field_textinput.js
```diff
@@ -295,34 +295,33 @@ Blockly.FieldTextInput.prototype.widgetCreate_ = function() {
...
+ * Closes the editor, saves the results, and disposes of any events or
+ * dom-references belonging to the editor.
  * @private
  */
 Blockly.FieldTextInput.prototype.widgetDispose_ = function() {
-  // Finalize value.
+  this.onEditorClose_();
+  this.unbindInputEvents_();
+  var style = Blockly.WidgetDiv.DIV.style;
+  style.width = 'auto';
```
samelhusseini: The widget div should probably do that, similar to how the dropdown div clears itself between uses.

core/field_angle.js
```diff
@@ -294,13 +294,15 @@ Blockly.FieldAngle.prototype.dropdownCreate_ = function() {
...
 /**
- * Dispose of events belonging to the angle editor.
+ * Disposes of events and dom-references belonging to the angle editor.
  * @private
  */
 Blockly.FieldAngle.prototype.dropdownDispose_ = function() {
   Blockly.unbindEvent_(this.clickWrapper_);
   Blockly.unbindEvent_(this.moveWrapper1_);
   Blockly.unbindEvent_(this.moveWrapper2_);
+  this.gauge_ = null;
```
samelhusseini: This isn't necessary.

BeksOmega: I think in this case it is.

If the angle field was itself being destroyed, it wouldn't need to set properties to null, but since we are releasing something else we need to set the properties to null.

Otherwise the dropdown div elements will be kept around as detached nodes. Learned that the hard way haha.

You can verify by running the following test:
```
      function memoryStrike() {
        for (var i = 0; i < 1000; i++) {
          var field = Blockly.Xml.domToBlock(Blockly.Xml.textToDom(
            '<block type="test_fields_angle"/>'), workspace).getField('FIELDNAME');
          field.showEditor_();
          Blockly.hideChaff();
        }
      }
```
And then looking at the heap's detached nodes; with and without setting the properties to null.

I was going to send you some heap snapshots I took, but github doesn't want to process my zip files --.--

core/field_textinput.js
BeksOmega: Ok I moved the responsibility for css into dropdown div and widget div.

It may break people who are depending on how we currently handle the editor css though, so I'd love opinions.

Also I need to verify that none of the dropdown div issues from #2551 have come back.

core/dropdowndiv.js
```diff
@@ -192,7 +197,23 @@ Blockly.DropDownDiv.getContentDiv = function() {
  */
 Blockly.DropDownDiv.clearContent = function() {
   Blockly.DropDownDiv.content_.innerHTML = '';
-  Blockly.DropDownDiv.content_.style.width = '';
+  Blockly.DropDownDiv.content_.style.cssText = '';
```
samelhusseini: Can we check on the compatibility of cssText in IE11? https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText

core/dropdowndiv.js
```diff
@@ -103,6 +103,7 @@ Blockly.DropDownDiv.PADDING_Y = 16;
 Blockly.DropDownDiv.ANIMATION_TIME = 0.25;
 
 /**
+<<<<<<< HEAD
```
samelhusseini: merge issue

core/dropdowndiv.js
```diff
@@ -192,7 +197,23 @@ Blockly.DropDownDiv.getContentDiv = function() {
...
+ * Resets the CSS of the div.
+ * @param {boolean} persistDisplay True to persist the display mode,
+ *    otherwise it will be set to 'none'.
+ */
+Blockly.DropDownDiv.resetCss = function(persistDisplay) {
+  var style = Blockly.DropDownDiv.DIV_.style;
+  var displayMode = style.display;
+  style.cssText = '';
+  style.opacity = 0;
+  style.transition = Blockly.DropDownDiv.TRANSITION;
```
samelhusseini: Is this just for cleanliness? We weren't setting resetting the transition property.

core/dropdowndiv.js
```diff
@@ -584,17 +605,9 @@ Blockly.DropDownDiv.hideWithoutAnimation = function() {
...
-  div.style.left = '';
-  div.style.top = '';
-  div.style.opacity = 0;
-  div.style.display = 'none';
-  div.style.backgroundColor = Blockly.DropDownDiv.DEFAULT_DROPDOWN_COLOR;
-  div.style.borderColor = Blockly.DropDownDiv.DEFAULT_DROPDOWN_BORDER_COLOR;
-
+  Blockly.DropDownDiv.clearContent();
+  Blockly.DropDownDiv.resetCss();
+  Blockly.DropDownDiv.owner_ = null;
```
samelhusseini: Note: I recently moved this so that's it after the onHide to give the onHide method the ability to interface with the dom element (eg: React to unmount). 
This will probably be resolved by a re-base but leaving a note to verify.

core/widgetdiv.js
```diff
@@ -85,18 +85,30 @@ Blockly.WidgetDiv.show = function(newOwner, rtl, dispose) {
...
+  Blockly.WidgetDiv.dispose_ = null;
+};
+
+/**
+ * Resets the css of the div.
+ * @private
+ */
+Blockly.WidgetDiv.resetCss_ = function() {
+  var style = Blockly.WidgetDiv.DIV.style;
+  style.cssText = '';
```
samelhusseini: I don't like this method of clearing the CSS. Best if we only add and remove what we care about.
That way if a developer really wants to add style elements to the div, we don't interfere with those.

BeksOmega: Wouldn't it be easier to just let the text input field clean up then?

samelhusseini: Yeah I see your point. Following on my previous comment, I would say let each field clean up anything it added to the dropdown div's style.

BeksOmega: Sounds good, I'll revert the previous commit.

core/field_textinput.js
```diff
@@ -292,34 +292,33 @@ Blockly.FieldTextInput.prototype.widgetCreate_ = function() {
...
+  style.height = 'auto';
+  style.fontSize = '';
+  this.htmlInput_ = null;
+};
+
+/**
+ * Deals with the not-disposal parts of closing the text input field's editor.
+ * @private
+ */
+Blockly.FieldTextInput.prototype.onEditorClose_ = function() {
```
samelhusseini: I don't see a reason to split this. 
There's already a divide here between widgetCreate and widgetDispose. 
I would however be in favour of adding a disposeView_ to match initView_ and dispose of any UI element references there.

samelhusseini: Just add comments in widgetDispose_ for a cleaner structure.

PR #3110, opened 2019-09-27
Simplify trashcan code

NeilFraser:
Also delete unused props in CSS.

### DIFF
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -45,15 +45,8 @@ Blockly.Trashcan = function(workspace) {
   this.workspace_ = workspace;
 
   /**
-   * True if the trashcan contains blocks, otherwise false.
-   * @type {boolean}
-   * @private
-   */
-  this.hasBlocks_ = false;
-
-  /**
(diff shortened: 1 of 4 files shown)


core/trashcan.js
```diff
@@ -467,35 +459,28 @@ Blockly.Trashcan.prototype.mouseOut_ = function() {
...
-    if (event.type == Blockly.Events.BLOCK_DELETE &&
-        event.oldXml.tagName.toLowerCase() != 'shadow') {
-      var cleanedXML = trashcan.cleanBlockXML_(event.oldXml);
-      if (trashcan.contents_.indexOf(cleanedXML) != -1) {
-        return;
-      }
-      trashcan.contents_.unshift(cleanedXML);
-      if (trashcan.contents_.length >
-          trashcan.workspace_.options.maxTrashcanContents) {
-        trashcan.contents_.splice(
```
NeilFraser: A single splice is more efficient than using a while loop and popping off multiple unwanted items.  But in Blockly's trashcan, n will always be 1.  So if/pop is simpler and while/pop adds a level of safety.

Issue #3106, opened 2019-09-26
Uncaught typeError in workspaceCommentRenderSvg


rachel-fenichel:
Uncompressed blockly is broken in the playground after a rebuild--I suspect one of the recent modularization changes.

To replicate:
- rebuild
- open the playground
- check the console

```
workspace_comment_render_svg.js:72 Uncaught TypeError: Cannot set property 'getHeightWidth' of undefined
    at workspace_comment_render_svg.js:72
(anonymous) @ workspace_comment_render_svg.js:72
```

It looks like WorkspaceCommentSvg is no longer being included, but WorkspaceCommentSvg.render is, and it's trying to set things on the prototype of WorkspaceCommentSvg.

PR #3115, opened 2019-09-27
Make workspace comments an optional module

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3106

### Proposed Changes

Fixes the dependency issue causing uncompressed to throw an error.
Also makes workspace comments an optional module, and removes them from this release as there's no way currently to surface them through workspace options.

I decided to go with a silent (warning) fail with the XML, as (if for whatever reason you are using workspace comments today and not requiring the module) it's not detrimental for the workspace to just not import them since the module doesn't exist.

### Reason for Changes

Shaves off 7 KB


NeilFraser:
Related to #3023.

PR #3120, opened 2019-09-27
Don't require everything in uncompressed.

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Blockly uncompressed requires everything in the directory regardless if it's in the dependency tree. It's okay to ``goog.addDependency`` of everything in uncompressed since that's just a representation of what depends on what, but we shouldn't require all dependencies explicitly. 

Leaving it as is could lead us to bugs that we only catch when we test compressed, and uncompressed should be a representation of what we plan to ship.

### Reason for Changes

Avoiding bugs.

### Test Coverage

Tested playground.

### Additional Information

@NeilFraser @rachel-fenichel 
Let me know if I'm missing any context.


rachel-fenichel:
Looks reasonable to use with the new entry point.

This means that debug rendering won't work in the playground until we add it back explicitly.

samelhusseini:
Added custom requires in the playground.html for block rendering debug and the other two renderers.

PR #3122, opened 2019-09-27
Move the dropdown field into the dropdown div.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Last (dropdown) field still using the widget div.

### Proposed Changes

Render the dropdown field in the dropdown div.
Add default colors to the dropdown div, so that each owner that doesn't care about setting the color doesn't need to set default colors. (We should think about how this ties into themes).
Added support for inverted text (white text with colored background), by default resorting to black text for the white default background.

### Test Coverage

Tested playground.

Tested on:
* Desktop Chrome

### Additional Information

Screen Shot 2019-09-27 at 2 02 55 PM


core/css.js
```diff
@@ -730,14 +733,9 @@ Blockly.Css.CONTENT = [
...
 
   /* State: hover. */
-  '.blocklyWidgetDiv .goog-menuitem-highlight,',
-  '.blocklyWidgetDiv .goog-menuitem-hover {',
-    'background-color: #f1f3f4;',
-  '}',
-
-  '.blocklyDropDownDiv .goog-menuitem-highlight,',
-  '.blocklyDropDownDiv .goog-menuitem-hover {',
-    'background-color: rgba(0,0,0,.2);',
```
samelhusseini: We don't use ``goog-menuitem-hover``

core/dropdowndiv.js
```diff
@@ -185,10 +201,16 @@ Blockly.DropDownDiv.clearContent = function() {
  * Set the colour for the drop-down.
  * @param {string} backgroundColour Any CSS colour for the background.
  * @param {string} borderColour Any CSS colour for the border.
+ * @param {boolean=} opt_invertedText Whether or not the text color should be
```
rachel-fenichel: Why `opt_invertedText` instead of `opt_className`?

samelhusseini: I was hoping to just provide the two options, and that was a way to indicate whether or not the color needs inverted text or not.
I will leave this out of this PR though, and think about it a little more.

Issue #2892, opened 2019-08-21
Set all object fields in the constructor instead of the prototype


samelhusseini:
The Google JS style guide says to set all concrete object fields (i.e. all properties other than methods) in the constructor as opposed to placing them on the prototype. [see](https://google.github.io/styleguide/jsguide.html#features-classes-fields).

The style guide sites the reason being a performance gain. Looking more into this, the performance gain comes from engines (such as V8) optimizing for the shape of the object that it gets from what the constructor instantiates. 

I had a chat to Neil and he ran some performance tests and found that there was a performance improvement on V8, but the two had very similar performance on Firefox. 
I then ran the test myself on Safari, and found that there was a performance improvement on Safari as well.

Because of this, moving forward all new concrete object fields (ie: all properties other than methods) should go in the constructor.

BeksOmega:
No longer an issue now that we've moved to typescript.

PR #3133, opened 2019-10-02
Connection Tracking Pt 3: Changed connections to use 'tracked_' property

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Changes all reference to `hidden_` or `inDb_` to `tracked_`.

There were also quite a few picky doc changes that went with this.

### Reason for Changes

These meant /slightly/ different things before, but they were similar enough that it seemed weird to have two properties. It also seemed confusing to have a property named 'hidden_' that didn't always reflect the visibility of the connection (since it had to do with database tracking, not visibility).

### Test Coverage

Already has tests coverage.

Reenabed two disable tests.

Tested on:

### Documentation

N/A

### Additional Information

Related to #2994

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -141,6 +141,7 @@ Blockly.utils.object.inherits(Blockly.BlockSvg, Blockly.Block);
  * Height is in workspace units.
  */
 Blockly.BlockSvg.prototype.height = 0;
+
 /**
  * Width of this block, including any connected value blocks.
  * Width is in workspace units.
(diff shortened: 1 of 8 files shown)


rachel-fenichel:
Holding until after we cut the release (tomorrow)

core/block_svg.js
```diff
@@ -162,6 +163,14 @@ Blockly.BlockSvg.prototype.dragStartXY_ = null;
  */
 Blockly.BlockSvg.prototype.warningTextDb_ = null;
 
+/**
```
rachel-fenichel: Put this in the constructor, per #2892

rachel-fenichel:
Good call on the naming.  It feels less awkward.

Issue #3132, opened 2019-10-02
Base renderer issues


samelhusseini:
**Describe the bug**

A couple of issues with the base renderer: 
- The next connection seems to shift left when a block is connected to it. 
- The notch is the statement input and the previous connection notch don't align. 

**To Reproduce**

Create a new renderer with no changes from the common renderer.

Notice how the child block notch doesn't align with the parent statement input notch.
![Screen Shot 2019-10-01 at 5 18 07 PM](https://user-images.githubusercontent.com/16690124/66009481-75c72900-e46f-11e9-909e-6e843440226d.png)

Notice how the connected block is shifted left: 
![Screen Shot 2019-10-01 at 5 18 11 PM](https://user-images.githubusercontent.com/16690124/66009540-b6bf3d80-e46f-11e9-98d2-4127fb32e92a.png)

samelhusseini:
This affects the Summit Hackathon.

PR #3135, opened 2019-10-02
Base renderer next / prev notch alignment issues

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Resolves https://github.com/google/blockly/issues/3132

### Proposed Changes

Add spacing before next / previous connection notches in the base renderer to account for the notch offset. 
Take into account square corners when there is a next connection.

Move baseline alignment of rows into the base renderer to take into account top and bottom row baseline. 

### Reason for Changes

Couple of issues with the base renderer.

### Test Coverage

Tested all renderers in the playground.

Tested on:

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -438,6 +438,22 @@ Blockly.blockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
     }
   }
 
+  // Spacing between a square corner and a previous or next connection
+  if (prev && Blockly.blockRendering.Types.isLeftSquareCorner(prev) && next) {
+    if (Blockly.blockRendering.Types.isPreviousConnection(next) ||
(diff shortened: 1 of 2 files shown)


core/renderers/common/info.js
```diff
@@ -612,16 +628,30 @@ Blockly.blockRendering.RenderInfo.prototype.getSpacerRowHeight_ = function(
...
  * @return {number} The desired centerline of the given element, as an offset
  *     from the top left of the block.
  * @protected
  */
 Blockly.blockRendering.RenderInfo.prototype.getElemCenterline_ = function(row,
-    _elem) {
-  var result = row.yPos;
-  result += (row.height / 2);
-  return result;
+    elem) {
```
rachel-fenichel: Next step: check if the other renderers can have their own logic and then just call the superclass function in the default case (at least for thrasos, maybe also for geras)

PR #3136, opened 2019-10-02
Add a sample renderer instance.

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Add a sample renderer instance that's only included in the playground. This allows us to test the ``common`` renderer.

### Test Coverage

Tested in playground.

Tested on:


samelhusseini:
I thought about calling it ``base`` but I worry that sends the wrong message since ``common`` is the real base.

rachel-fenichel:
`minimalist`?

PR #3144, opened 2019-10-03
Fix compiler warnings related to fields.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes 34 warnings related to fields.

### Proposed Changes

``bumpNeighbours`` is @package.
Fix unregister calls to delete from map rather than set to undefined.
Make ``SERIALIZABLE`` and ``EDITABLE`` not constants.
Fix ``doValueUpdate_`` parameter to be consistent with super class. 

### Reason for Changes

Compiler warnings.

### Test Coverage

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -518,11 +518,11 @@ Blockly.Block.prototype.lastConnectionInStack = function() {
 /**
  * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
  * connected should not coincidentally line up on screen.
- * @protected
+ * @package
  */
-Blockly.Block.prototype.bumpNeighbours_ = function() {
-  console.warn('Not expected to reach Block.bumpNeighbours_ function. ' +
-      'BlockSvg.bumpNeighbours_ was expected to be called instead.');
+Blockly.Block.prototype.bumpNeighbours = function() {
+  console.warn('Not expected to reach Block.bumpNeighbours function. ' +
+      'BlockSvg.bumpNeighbours was expected to be called instead.');
 };
 
 /**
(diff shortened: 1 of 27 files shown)


core/block.js
```diff
@@ -1106,7 +1107,7 @@ Blockly.Block.prototype.renameVarById = function(oldId, newId) {
 Blockly.Block.prototype.getFieldValue = function(name) {
   var field = this.getField(name);
   if (field) {
-    return field.getValue();
+    return String(field.getValue());
```
samelhusseini: This is an interesting one. field.getValue() can be of any type internally, but getFieldValue promises a string. Casting to string might not be desired, we could instead return a ``{*}``. Thoughts @rachel-fenichel?

BeksOmega: I think it needs to return `{*}` because `String()` will break object values.

samelhusseini: Yep, that’s a fair point. I’ll change it to return {string} instead

core/field_colour.js
```diff
@@ -196,7 +195,7 @@ Blockly.FieldColour.prototype.doClassValidation_ = function(opt_newValue) {
 
 /**
  * Update the value of this colour field, and update the displayed colour.
- * @param {string} newValue The new colour in '#rrggbb' format.
+ * @param {*} newValue The value to be saved. A colour in '#rrggbb' format.
```
rachel-fenichel: I think these should say something like "the default validator guarantees that this is a colour in '#rrggbb' format", and similar for the other `doValueUpdate_` functions.

core/field_variable.js
```diff
@@ -144,7 +143,8 @@ Blockly.FieldVariable.prototype.initModel = function() {
     return; // Initialization already happened.
   }
   var variable = Blockly.Variables.getOrCreateVariablePackage(
-      this.workspace_, null, this.defaultVariableName, this.defaultType_);
+      this.sourceBlock_.workspace, null,
```
rachel-fenichel: Is this ever a behaviour change?

rachel-fenichel: Ah never mind, I see.

core/field_colour.js
samelhusseini: Done.

NeilFraser:
This appears to break the compiler: #3152

PR #3147, opened 2019-10-03
Misc cleanup

rachel-fenichel:
Fix some annotations and remove setting things to null in dispose for flyout buttons.

I played around with the memory tab in chrome dev tools to make sure this wasn't retaining anything extra as a result.

There are some event listeners attached to the dom nodes but they don't keep anything else alive, and no one else references them, so they disappear with the nodes.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -947,9 +947,9 @@ Blockly.WorkspaceSvg.prototype.getParentSvg = function() {
   while (element) {
     if (element.tagName == 'svg') {
       this.cachedParentSvg_ = element;
-      return /** @type {!SVGElement} */ (element);
+      return element;
     }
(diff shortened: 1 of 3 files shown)


core/workspace_svg.js
```diff
@@ -947,7 +947,7 @@ Blockly.WorkspaceSvg.prototype.getParentSvg = function() {
   while (element) {
     if (element.tagName == 'svg') {
       this.cachedParentSvg_ = element;
-      return element;
+      return /** @type {!SVGElement} */ (element);
     }
     element = element.parentNode;
```
samelhusseini: Best to do the cast on the source, which is here: 
```
element = /** @type {!SVGElement} */ (element.parentNode);
```

PR #3154, opened 2019-10-04
Get rid of unnecessary nulls in block dragger, insertion marker manag…

rachel-fenichel:
…er, and gesture dispose functions.

## The basics

- [x] I branched from develop

### Resolves

Work on #2152

### Proposed Changes

Stop setting some properties to null in the dispose functions for the block dragger, insertion marker manager, and gesture classes.

### Reason for Changes

Garbage collector can handle it just fine.

### Test Coverage
Tested in the playground:
- refresh
- load a single block from xml
- force run the garbage collector from chrome dev tools
- take a heap snapshot
- drag the block around
- force run the garbage collector
- take a heap snapshot
- look at detached nodes in the second snapshot, and compare sizes of first and second snapshots (should be the same or almost the same)

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -100,14 +100,10 @@ Blockly.BlockDragger = function(block, workspace) {
  * @package
  */
 Blockly.BlockDragger.prototype.dispose = function() {
-  this.draggingBlock_ = null;
-  this.workspace_ = null;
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
This one should wait until after we cut the release.

Issue #3152, opened 2019-10-04
Visibility of bumpNeighbours breaks compiler


NeilFraser:
```
../../blocks/lists.js:842: ERROR - Access to package-private property bumpNeighbours of Blockly.Block not allowed here.
          this.bumpNeighbours();
          ^^^^^^^^^^^^^^^^^^^

../../blocks/logic.js:577: ERROR - Access to package-private property bumpNeighbours of Blockly.Block not allowed here.
      this.bumpNeighbours();
      ^^^^^^^^^^^^^^^^^^^

../../blocks/logic.js:632: ERROR - Access to package-private property bumpNeighbours of Blockly.Block not allowed here.
            block.bumpNeighbours();
            ^^^^^^^^^^^^^^^^^^^^

../../blocks/procedures.js:256: ERROR - Access to package-private property bumpNeighbours of Blockly.Block not allowed here.
            stackBlock.bumpNeighbours();
            ^^^^^^^^^^^^^^^^^^^^^^^^^

../../blocks/procedures.js:755: ERROR - Access to package-private property bumpNeighbours of Blockly.Block not allowed here.
          connection.getSourceBlock().bumpNeighbours();
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

5 error(s), 445 warning(s), 95.2% typed
```

NeilFraser:
Cause appears to be: #3144

samelhusseini:
Interestingly @package is more restrictive than @private as it causes a compiler failure if used outside of a folder. but @private would just cause a warning.

samelhusseini:
Fixed.

PR #3156, opened 2019-10-04
Make bumpNeighbours public.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3152

### Proposed Changes

Turns out @package is more restrictively. Making ``bumpNeighbours`` public since it's used in blocks.
This is a bit lame, because what we really want is @package. A way to say this is okay to be used within our codebase but not outside, but blocks falls in a different directory, so we can't use @package. 

### Test Coverage

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -539,7 +539,6 @@ Blockly.Block.prototype.lastConnectionInStack = function() {
 /**
  * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
  * connected should not coincidentally line up on screen.
- * @package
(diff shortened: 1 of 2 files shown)


RoboErikG:
Separate from this change, why was @package causing an error in the compiler? Our compilation script should be putting everything in the same folder.

RoboErikG:
Ah, right. Blocks are built separately from core. Making it public makes sense since specific blocks are using it.

Issue #3153, opened 2019-10-04
Object.assign (ES6) in codebase


NeilFraser:
[Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) does not exist in any version of IE (according to MDN).
The following instances exist:

- [ ] [/core/keyboard_nav/key_map.js:81](https://github.com/google/blockly/blob/develop/core/keyboard_nav/key_map.js#L81)
- [ ] [/demos/keyboard_nav/index.html:328](https://github.com/google/blockly/blob/develop/demos/keyboard_nav/index.html#L328)
- [ ] [/demos/custom-fields/turtle/blocks.js:51](https://github.com/google/blockly/blob/develop/demos/custom-fields/turtle/blocks.js#L51) 

There are also three more in /package, but I don't know if they matter.

rachel-fenichel:
@alschmiedt for the ones in keyboard nav

@BeksOmega for the one in the turtle demo

I think the ones in /package just derive from the ones in core and demos.

samelhusseini:
The ones in package are used by the npm package. I’ll take a look at those. I remember testing IE with the npm package, so I wonder why that wasn’t caught.

samelhusseini:
Turns out each framework patches in a polyfill. Except for our UMD sample, which is broken in IE. Will fix.

samelhusseini:
Fixed.

PR #3158, opened 2019-10-04
Fix UMD build in IE11

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3153

### Proposed Changes

IE11 is broken on the UMD build. Other frameworks insert polyfills when transpiling to ES6. 
UMD sample loads our package using script tags. 

### Test Coverage

Tested the UMD sample in IE11. 
https://github.com/google/blockly-samples/tree/master/blockly-umd

Tested on:


NeilFraser:
Why doesn't this PR just use the existing function: Blockly.utils.object.mixin

samelhusseini:
I think it could. Will clean up after the release.

PR #3160, opened 2019-10-04
Add a lint plugin to ensure we only use ES5 syntax.

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3153

### Proposed Changes

Adds an eslint plugin that ensure we don't use ES6 syntax in our codebase.
Rename ``Blockly.utils.object.values`` to ``Blockly.utils.object.valuesShim``.
Use ``Blockly.utils.object.mixin`` for the last instance of Object.assign in our codebase.

### Reason for Changes

No Object.assign in IE. Avoid having to constantly deal with this in the future.

### Test Coverage
Playground.
eslint.

Tested on:


RoboErikG:
Let's hold off on submitting until we finish discussing options.

rachel-fenichel:
@RoboErikG I'm fine with this (with the plugin set as warn rather than error)

RoboErikG:
Leaving as warnings sgtm.

samelhusseini:
Updated, leaving ``es5/no-es6-methods`` as warn.

tests/node/run_node_test.js
```diff
@@ -22,28 +22,28 @@
 var assert = require('chai').assert;
 var Blockly = require('../../dist/');
 
-var xmlText = `<xml xmlns="https://developers.google.com/blockly/xml">
```
rachel-fenichel: You could also add this file to the eslintignore and leave it as es6.

PR #3171, opened 2019-10-05
Fix some calls with the wrong number of arguments.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Calls with wrong numbers of arguments.

Also calls clearAllBlocks in flyout dispose.

### Reason for Changes
- warnings
- blocks, mats, etc. are not cleaned up during hide

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -262,7 +262,7 @@ Blockly.onKeyDown_ = function(e) {
   if (deleteBlock && !Blockly.selected.workspace.isFlyout) {
     Blockly.Events.setGroup(true);
     Blockly.hideChaff();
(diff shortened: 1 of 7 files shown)


core/blockly.js
```diff
@@ -262,7 +262,7 @@ Blockly.onKeyDown_ = function(e) {
   if (deleteBlock && !Blockly.selected.workspace.isFlyout) {
     Blockly.Events.setGroup(true);
     Blockly.hideChaff();
-    Blockly.selected.dispose(/* heal */ true, true);
+    Blockly.selected.dispose(/* heal */ true);
```
samelhusseini: This is functionally different as the second argument for a rendererd block (BlockSvg) is to animate the disposal

PR #3179, opened 2019-10-07
Set correct defaults for Field min/max/precision

### LINKED ISSUE #3177: FieldNumber interprets null min/max as 0.
**Describe the bug**

Converting to Number(value) in FieldNumber had the effect of causing nulls passed to the constructor for min/max to be interpreted as 0. We need to explicitly check for null and set them to +/- infinity as appropriate.

**To Reproduce**

1. Select test blocks in the playground
2. Drag out a rounding to 10 block from the validators category
3. Try to set the value to anything other than 0

**Expected behavior**

value gets rounded to the nearest 10

**Screenshots**

**Stack Traces**

```
```

**Additional context**

RoboErikG:
null was being converted to 0 by Number() when it should cause the
default value to be set instead. This updates FieldNumber to handle
nulls correctly.

Fixes #3177

Should be cherry-picked

## The basics

- [x] I branched from develop

### Resolves

Fixes #3177

### Test Coverage

Tested on:

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/3179)

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -32,10 +32,10 @@ goog.require('Blockly.utils.object');
  * Class for an editable number field.
  * @param {string|number=} opt_value The initial value of the field. Should cast
  *    to a number. Defaults to 0.
- * @param {(string|number)=} opt_min Minimum value.
- * @param {(string|number)=} opt_max Maximum value.
- * @param {(string|number)=} opt_precision Precision for value.
- * @param {Function=} opt_validator A function that is called to validate
+ * @param {?(string|number)=} opt_min Minimum value.
+ * @param {?(string|number)=} opt_max Maximum value.
+ * @param {?(string|number)=} opt_precision Precision for value.
+ * @param {?Function=} opt_validator A function that is called to validate
  *    changes to the field's value. Takes in a number & returns a validated
  *    number, or null to abort the change.
  * @param {Object=} opt_config A map of options used to configure the field.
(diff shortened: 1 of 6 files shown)


RoboErikG:
@BeksOmega can you think of any other fields that might be affected like this?

samelhusseini:
This lgtm, but can we have the types reflect that input values. Update each of the min, max, precision @param types to accept null. ie: ``?(number|string|undefined)``
Both for the setter (eg: setMin) and the internal set method (eg: setMinInternal_).

samelhusseini:
Field angle looks like it's also affected by this, when it casts the config offset / wrap / round to Number: https://github.com/google/blockly/blob/develop/core/field_angle.js#L175

core/field_number.js
```diff
@@ -32,10 +32,10 @@ goog.require('Blockly.utils.object');
...
  * @param {string|number=} opt_value The initial value of the field. Should cast
  *    to a number. Defaults to 0.
- * @param {(string|number)=} opt_min Minimum value.
- * @param {(string|number)=} opt_max Maximum value.
- * @param {(string|number)=} opt_precision Precision for value.
- * @param {Function=} opt_validator A function that is called to validate
+ * @param {?(string|number)=} opt_min Minimum value.
+ * @param {?(string|number)=} opt_max Maximum value.
+ * @param {?(string|number)=} opt_precision Precision for value.
+ * @param {?Function=} opt_validator A function that is called to validate
```
samelhusseini: I'd leave this one out unless the super class is ready to handle null in it's types. FieldTextInput doesn't say that it handles a null validator (currently).

RoboErikG: It has to handle null if the dev is using the config object, as it's not reasonable to ask devs to pass undefined instead of null to get to config.

RoboErikG:
Updated field angle as well and propagated the ? annotation for the validator back up the stack.

Issue #3190, opened 2019-10-07
Dropdown field ltr options in RTL


samelhusseini:
**Describe the bug**

Any dropdown field shows options in an ltr direction when the workspace direction is RTL.

**Screenshots**

![Screen Shot 2019-10-07 at 4 32 10 PM](https://user-images.githubusercontent.com/16690124/66356538-26c23d80-e920-11e9-89a8-196284563101.png)

samelhusseini:
Fixed.

PR #3165, opened 2019-10-04
October 2019 release candidate (Q3 2019)

rachel-fenichel:
Issues closed:

- Object.assign (ES6) in codebase (https://github.com/google/blockly/issues/3153)
- Visibility of bumpNeighbours breaks compiler (https://github.com/google/blockly/issues/3152)
- Base renderer issues (https://github.com/google/blockly/issues/3132)
- Uncaught typeError in workspaceCommentRenderSvg (https://github.com/google/blockly/issues/3106)
- Inline image dropdown too low. (https://github.com/google/blockly/issues/3098)
- Remove render debug and non-default renderers from default build (https://github.com/google/blockly/issues/3089)
- +/- buttons for dynamically extending value input list (https://github.com/google/blockly/issues/3078)
- Keyboard Nav: Interaction Between Mouse and Keyboard (https://github.com/google/blockly/issues/3071)
- Add the ability to make a read-only workspace writable (https://github.com/google/blockly/issues/3049)
- npm: new blocks instances appear below menuFlyout while dragging (https://github.com/google/blockly/issues/3032)
- Keyboard Nav: Cursor not showing correct location (https://github.com/google/blockly/issues/3028)
- Compile test is broken (https://github.com/google/blockly/issues/3027)
- Keyboard Nav: Cursor in RTL not working (https://github.com/google/blockly/issues/3024)
- Link in `ReadME` points to an invalid URL (https://github.com/google/blockly/issues/3021)
- Have a way to choose which renderers are included in your bundle (https://github.com/google/blockly/issues/3018)
- Unable to scroll the trash flyout without a mouse scroll (https://github.com/google/blockly/issues/3008)
- Drag in trash flyout causes error (https://github.com/google/blockly/issues/3007)
- Object.values not available in IE (https://github.com/google/blockly/issues/2999)
- Missing hideChaff in custom field demo? (https://github.com/google/blockly/issues/2998)
- block.dispose(/* healStack */ true) does not properly handle rows (https://github.com/google/blockly/issues/2991)
- Endmost value connections are not considered available (https://github.com/google/blockly/issues/2989)
- how to build package published on NPM (https://github.com/google/blockly/issues/2987)
- [zelos] Dynamic output shapes (https://github.com/google/blockly/issues/2977)
- How to compile project with local closure compiler? (https://github.com/google/blockly/issues/2975)
- Define a BlockRenderer object that Geras, Thrasos, and Zelos extend (https://github.com/google/blockly/issues/2966)
[...]

PR #3192, opened 2019-10-08
Dropdowndiv RTL support

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3190

### Proposed Changes

Add support for setting the direction in RTL.
Fields that want to always render in LTR can explicitly set that in their own divs.

### Reason for Changes
RTL support.

### Test Coverage

Tested on playground in: 

Tested on:
* Desktop Chrome
* Desktop Safari

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -62,7 +62,7 @@ Blockly.DropDownDiv.owner_ = null;
 
 /**
  * Whether the dropdown was positioned to a field or the source block.
- * @type {boolean}
+ * @type {?boolean}
  * @private
  */
 Blockly.DropDownDiv.positionToField_ = null;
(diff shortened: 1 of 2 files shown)


samelhusseini:
@rachel-fenichel I think we should cherry-pick this.

core/dropdowndiv.js
```diff
@@ -290,17 +292,19 @@ Blockly.DropDownDiv.showPositionedByField = function(owner,
...
  * secondary point and position above.
  * @param {Object} owner The object showing the drop-down
+ * @param {boolean} rtl Right-to-left (true) or left-to-right (false).
  * @param {number} primaryX Desired origin point x, in absolute px
  * @param {number} primaryY Desired origin point y, in absolute px
  * @param {number} secondaryX Secondary/alternative origin point x, in absolute px
  * @param {number} secondaryY Secondary/alternative origin point y, in absolute px
  * @param {Function=} opt_onHide Optional callback for when the drop-down is hidden
  * @return {boolean} True if the menu rendered at the primary origin point.
+ * @package
```
samelhusseini: I don't see a reason for developers to use this directly, so I changed it to @package. If we think this should be kept public, happy to revert that.

core/dropdowndiv.js
```diff
@@ -290,17 +292,19 @@ Blockly.DropDownDiv.showPositionedByField = function(owner,
  * If we can't maintain the container bounds at the primary point, fall-back to the
  * secondary point and position above.
  * @param {Object} owner The object showing the drop-down
+ * @param {boolean} rtl Right-to-left (true) or left-to-right (false).
```
samelhusseini: I recognize that this is breaking for any developers using this directly. I checked, and both Makecode and Scratch don't call this directly, since this is new, having the RTL here would match widgetdiv (and would be my preferred api), but if we think it's too late to change this API, then happy to throw this at the end.

samelhusseini:
Cherry-picked to release https://github.com/google/blockly/pull/3165/commits/60d69b9dbaafd1d06c8a973a5e402c1e57d171a1

PR #3195, opened 2019-10-08
Set correct defaults for Fields (#3179)

### LINKED ISSUE #3177: FieldNumber interprets null min/max as 0.

RoboErikG:
null was being converted to 0 by Number() when it should cause the
default value to be set instead. This updates FieldNumber and
FieldAngle to handle nulls correctly. Also update jsdoc.

Fixes #3177

## The basics

- [ ] I branched from develop

### Test Coverage

Tested on:

### Additional Information

---
This change is [Reviewable](https://reviewable.io/reviews/google/blockly/3195)

### DIFF
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -172,19 +172,31 @@ Blockly.FieldAngle.prototype.configure_ = function(config) {
   if (typeof clockwise == 'boolean') {
     this.clockwise_ = clockwise;
   }
-  var offset = Number(config['offset']);
-  if (!isNaN(offset)) {
-    this.offset_ = offset;
+
+  var offset = config['offset'];
+  if (offset != null) {
+    offset = Number(offset);
+    if (!isNaN(offset)) {
+      this.offset_ = offset;
+    }
   }
-  var wrap = Number(config['wrap']);
-  if (!isNaN(wrap)) {
-    this.wrap_ = wrap;
+  var wrap = config['wrap'];
+  if (wrap != null) {
+    wrap = Number(wrap);
+    if (!isNaN(wrap)) {
+      this.wrap_ = wrap;
+    }
   }
-  var round = Number(config['round']);
-  if (!isNaN(round)) {
-    this.round_ = round;
+  var round = config['round'];
+  if (round != null) {
+    round = Number(round);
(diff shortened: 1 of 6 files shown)


tests/mocha/field_number_test.js
```diff
@@ -247,6 +247,12 @@ suite('Number Fields', function() {
           numberField.setValue(123.456);
           assertValue(numberField, 123);
         });
+        test('null', function() {
```
rachel-fenichel: It's weird to me that you're testing whether the number is set instead of checking what the max/min/precision values on the field are.  Why?

core/field_angle.js
```diff
@@ -172,20 +172,32 @@ Blockly.FieldAngle.prototype.configure_ = function(config) {
   if (typeof clockwise == 'boolean') {
     this.clockwise_ = clockwise;
   }
-  var offset = Number(config['offset']);
-  if (!isNaN(offset)) {
-    this.offset_ = offset;
+
```
rachel-fenichel: If null -> do nothing -> use the value set on Blockly.FieldAngle, correct?

Worth a comment if so.

tests/mocha/field_number_test.js
RoboErikG: Whoops. I was copying the way the previous tests work and didn't think too much about it. Probably makes sense to check the values directly, though that can be a followup since this is a cherry-pick.

core/field_angle.js
RoboErikG: Correct. I'll add that in a followup pr as well (though not worth cherry-picking).

rachel-fenichel:
Ah, I didn't see that this is a cherry-pick.

Approve.

PR #3196, opened 2019-10-08
Fix compilier warnings for keyboard nav

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes a bunch of compiler warnings.

### Proposed Changes
Less trivial changes: 
1. Changed canConnectWithReason and checkConnection from private to package so I could use in navigation.
2. Changed getTopStackBlock in block to package. 
3. Changed getSourceBlock in connection to return !Blockly.Block since it doesn't look like it can be set to null anywhere. 
4. Create methods for ASTNodes handle being given a null value and return null if this is the case.
5. Never let the cursor_ or marker_ on the workspace be null.

### Test Coverage

Tested on:

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -28,9 +28,9 @@ goog.provide('Blockly.ASTNode');
  * Class for an AST node.
  * It is recommended that you use one of the createNode methods instead of
  * creating a node directly.
- * @param {string} type The type of the location.
+ * @param {!string} type The type of the location.
  *     Must be in Bockly.ASTNode.types.
- * @param {Blockly.Block|Blockly.Connection|Blockly.Field|Blockly.Workspace}
+ * @param {!(Blockly.Block|Blockly.Connection|Blockly.Field|Blockly.Workspace)}
  *     location The position in the AST.
  * @param {!Object=} opt_params Optional dictionary of options.
  * @constructor
@@ -107,10 +107,13 @@ Blockly.ASTNode.isConnectionType_ = function(type) {
 
 /**
  * Create an AST node pointing to a field.
- * @param {!Blockly.Field} field The location of the AST node.
- * @return {!Blockly.ASTNode} An AST node pointing to a field.
+ * @param {Blockly.Field} field The location of the AST node.
+ * @return {Blockly.ASTNode} An AST node pointing to a field.
  */
 Blockly.ASTNode.createFieldNode = function(field) {
+  if (!field) {
+    return null;
+  }
   return new Blockly.ASTNode(Blockly.ASTNode.types.FIELD, field);
 };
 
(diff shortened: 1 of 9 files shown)


core/keyboard_nav/ast_node.js
```diff
@@ -28,9 +28,9 @@ goog.provide('Blockly.ASTNode');
  * Class for an AST node.
  * It is recommended that you use one of the createNode methods instead of
  * creating a node directly.
- * @param {string} type The type of the location.
+ * @param {!string} type The type of the location.
```
samelhusseini: string is ``!string`` by default. No need.

core/keyboard_nav/ast_node.js
```diff
@@ -398,9 +410,10 @@ Blockly.ASTNode.prototype.navigateBetweenStacks_ = function(forward) {
 Blockly.ASTNode.prototype.findTopASTNodeForBlock_ = function(block) {
   var topConnection = block.previousConnection || block.outputConnection;
   if (topConnection) {
-    return Blockly.ASTNode.createConnectionNode(topConnection);
+    return /** @type {!Blockly.ASTNode} */ (Blockly.ASTNode.createConnectionNode(
+        topConnection));
   } else {
-    return Blockly.ASTNode.createBlockNode(block);
+    return /** @type {!Blockly.ASTNode} */ (Blockly.ASTNode.createBlockNode(block));
```
samelhusseini: more than 80 chars.

core/keyboard_nav/cursor_svg.js
```diff
@@ -274,7 +274,7 @@ Blockly.CursorSvg.prototype.showWithField_ = function(curNode) {
 Blockly.CursorSvg.prototype.showWithInput_ = function(curNode) {
   var connection = /** @type {Blockly.Connection} */
       (curNode.getLocation());
-  var sourceBlock = /** @type {Blockly.BlockSvg} */ (connection.getSourceBlock());
+  var sourceBlock = /** @type {!Blockly.BlockSvg} */ (connection.getSourceBlock());
```
samelhusseini: is this cast needed anymore? Since getSourceBlock is now ``!Blockly.BlockSvg``

alschmiedt: Yes because setParent takes in a Blockly.BlockSvg not a Blockly.Block so it complains when I remove it.

samelhusseini: But you just changed getSourceBlock to return a ``!Blockly.BlockSvg``.

alschmiedt: I changed to !Blockly.Block. I was originally trying to fix this problem but when it didn't I just left the change since it seems more correct since the sourceBlock never gets set to null.

samelhusseini: Yep makes sense.

PR #3201, opened 2019-10-09
More field dispose

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Memory leaks with fields when closing fields, but also when disposing of fields.

### Proposed Changes

Dispose of references to all elements created as part of the render pass in field ``dispose``. Dispose of all references to elements created in widget/dropdown create methods in widget/dropdown dispose.

### Reason for Changes

Plugging leaks

### Test Coverage

Tested using chrome perf tools.
Creating and disposing fields + opening / closing fields.

Tested on:
* Desktop Chrome

### Additional Information

@BeksOmega FYI

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -94,6 +94,41 @@ Blockly.Field = function(value, opt_validator, opt_config) {
    */
   this.markerSvg_ = null;
 
+  /**
+   * The rendered field's SVG group element.
+   * @type {SVGGElement}
+   * @protected
+   */
+  this.fieldGroup_ = null;
+
+  /**
+   * The rendered field's SVG border element.
+   * @type {SVGRectElement}
+   * @protected
+   */
+  this.borderRect_ = null;
+
+  /**
+   * The rendered field's SVG text element.
+   * @type {SVGTextElement}
+   * @protected
+   */
+  this.textElement_ = null;
+
+  /**
+   * The rendered field's text content element.
+   * @type {Text}
+   * @protected
+   */
+  this.textContent_ = null;
+
+  /**
+   * Mouse down event listener data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.mouseDownWrapper_ = null;
+
   opt_config && this.configure_(opt_config);
   this.setValue(value);
   opt_validator && this.setValidator(opt_validator);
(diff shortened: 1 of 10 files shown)


core/field_dropdown.js
```diff
@@ -611,4 +633,15 @@ Blockly.FieldDropdown.prototype.onBlocklyAction = function(action) {
...
 };
 
+/**
+ * Dispose of all DOM objects and events belonging to this editable field.
+ * @package
+ * @override
+ */
+Blockly.FieldDropdown.prototype.dispose = function() {
+  Blockly.FieldDropdown.superClass_.dispose.call(this);
+  this.imageElement_ = null;
```
BeksOmega: As far as I know, you don't need to set these to null in this case.

Since fields don't get disposed of directly, and are only disposed through input.removeField the input handles dropping the reference to the field. Because the field is then unreachable, these properties become unreachable as well, so the elements get disposed of, instead of sticking around as detached nodes.

Might be useful to @NeilFraser for an opinion (and also to check my reasoning).

samelhusseini: I think you might be right, did a quick test and it doesn't appear to leave any detached nodes behind when not set to null.

core/field.js
```diff
@@ -186,11 +221,21 @@ Blockly.Field.prototype.clickTarget_ = null;
...
 Blockly.Field.prototype.getText_;
 
+/**
+ * An optional method that can be defined to show an editor when the field is
+ *     clicked. Blockly will automatically set the field as clickable if this
+ *     method is defined.
+ * @return {void}
+ * @protected
+ */
+Blockly.Field.prototype.showEditor_;
```
BeksOmega: Could you check that `isClickable` works w/ this change? And verify that the field clicking works? I think `isClickable` can maybe be simplified now.

samelhusseini: When it's defined like this, showEditor_ is undefined in JS, and so it doesn't affect any of the logic in isClickable.
This is purely just a hint to the closure compiler.

BeksOmega: Oh sweet! Good to know.

core/field.js
```diff
@@ -260,11 +305,13 @@ Blockly.Field.prototype.init = function() {
...
     return;
   }
-  this.fieldGroup_ = Blockly.utils.dom.createSvgElement('g', {}, null);
+  this.fieldGroup_ = /** @type {!SVGGElement} **/
+      (Blockly.utils.dom.createSvgElement('g', {}, null));
   if (!this.isVisible()) {
     this.fieldGroup_.style.display = 'none';
   }
-  this.sourceBlock_.getSvgRoot().appendChild(this.fieldGroup_);
+  var sourceBlockSvg = (/** @type {!Blockly.BlockSvg} **/ (this.sourceBlock_));
```
moniika: You don't need to wrap this in parenthesis, removing them will make in more consistent with other places.

PR #3214, opened 2019-10-09
bringToFront optimization for large workspaces

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

The ``appendChild`` is causing a reflow takes much longer with larger workspaces.
Something as simple as a field click causes calls to bringToFront.

### Proposed Changes

Minor optimization to skip moving a block if it's already in the right position.

### Reason for Changes

Perf.

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1375,7 +1375,12 @@ Blockly.BlockSvg.prototype.bringToFront = function() {
   var block = this;
   do {
     var root = block.getSvgRoot();
-    root.parentNode.appendChild(root);
+    var parent = root.parentNode;
+    var childNodes = parent.childNodes;
+    // Avoid moving the block if it's already at the bottom.
+    if (childNodes[childNodes.length - 1] !== root) {
+      root.parentNode.appendChild(root);
+    }
     block = block.getParent();
   } while (block);
 };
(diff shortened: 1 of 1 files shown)


core/block_svg.js
```diff
@@ -1375,7 +1375,12 @@ Blockly.BlockSvg.prototype.bringToFront = function() {
   var block = this;
   do {
     var root = block.getSvgRoot();
-    root.parentNode.appendChild(root);
+    var parent = root.parentNode;
+    var childNodes = parent.childNodes;
+    // Avoid moving the block if it's already at the bottom.
+    if (childNodes[childNodes.length - 1] !== root) {
+      root.parentNode.appendChild(root);
```
RoboErikG: nit, this can be parent.appendChild(root)

rachel-fenichel:
Is this big enough that we want to put it in this release, or okay to wait?

samelhusseini:
Not big enough I don’t think. It’s really just skipping moving blocks around an in turn a reflow if we don’t need to.
It’s safe enough to take in though

rachel-fenichel:
Okay, no cherry-pick, since that's the default.

PR #3224, opened 2019-10-13
Rephrase "variable already exists" message

### LINKED ISSUE #1568: Phrasing of "variable already exists" error
### Problem statement

When a user tries to create a variable with an in-use name we alert: "A variable named "%1" already exists."

Currently we always lower-case the variable name for use in the message, presumably to indicate that variable names are not case sensitive.

### Expected Behavior

@AnmAtAnm says:
>I know you're repeating my prior code, but in retrospect it seems strange to repeat the variable name back to the user's input in the lowercased form. Ideally, if existing.name !== text append something like ' (as "${existing.name}")' to the variable name. That would make it clear that different name casings map to the same thing.

### Actual Behavior

Name is always lowercased, regardless of what the user typed or what the conflicting name's casing is.

### Steps to Reproduce

1. Go to the playground
2. Create a variable named "fOOf"
3. Try to create a variable named "Foof"
4. Message shows "A variable named "foof" already exists"

All

### Additional Information

FOOF is [exciting](http://blogs.sciencemag.org/pipeline/archives/2010/02/23/things_i_wont_work_with_dioxygen_difluoride).

zochris:
## The basics

- [x] I branched from develop

### Resolves

Fix #1568

### Proposed Changes
The message that pops up when trying to create a variable with a name that already exists now includes the name with the casing of the already existing variable.

### Reason for Changes

See #1568 for details.

### Test Coverage

Tested on:
* Desktop Chrome Version 77.0.3865.90 (Official Build) (64-bit)

* Windows Edge Version 44.18362.387.0

Tested by manually creating a variable and then trying to add another with the same name but different casing.

### Additional Information

This will probably break all non-english translations, so please treat this as a proposal.


NeilFraser:
This would mean that the most common case (where there is no case change) becomes:
> A variable named "index" already exists (as "index").
This seems overly wordy and redundant.  Also, as the number of '%' substitutions go up in a message, the failure rate for our non-programmer volunteer translators rises.

I think it would be fine to just print `existing.name` in the current message and drop the lowercasing.

zochris:
Yes, that makes more sense.

PR #3226, opened 2019-10-13
restart cursor animation on cursor move

### LINKED ISSUE #3225: Keyboard shortcut cursor sometimes feels unresponsive
**Describe the bug**

When "accessibility mode" is turned on at demos/keyboard_nav/index.html, the cursor animate fades in out similar to the blinking caret in a text input. However unlike a text input blinking caret, the fading animation in Blockly does not reset when the cursor moves. Because of this, the effect of keypresses that occur while the cursor is not visible appears delayed to the user.

**To Reproduce**

1. Open demos/keyboard_nav/index.html in a browser
2. Click the 'Enable Accessibility Mode' checkbox
3. Press 'd' or 's' to move the cursor
4. Try moving the cursor while the cursor is not visible

**Expected behavior**

Like a text edit caret, the cursor should immediately become visible when it moves.

**Screenshots**

fixhighlight

 - OS: OSX 10.14
 - Browser: Firefox + Chrome
 - Version: Firefox Nightly 71.0a1 (2019-10-02) + Chrome Version 77.0.3865.90 (Official Build) (64-bit)

**Additional context**

I discussed this in-person with Abby at the Blockly Summit Hackathon on Oct 11.

thomasballinger:
## The basics

- [x] I branched from develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide) *mostly*
  - [ ] I have not confirmed that this works in IE11. It's possible that SMIL animations for SVG don't work at all on IE11.
  - [ ] I have not run eslint

### Resolves

Resolves #3225 in which movement of the accessibility cursor appears delayed for movement which occurs while the cursor is not visible.

### Proposed Changes

Adds a check for an animate (in a way that assumes animate is the only type of child the cursor ever has) element in the current cursor and calls `beginElement()` on it if found to restart the animation.

### Reason for Changes

Before patch
fixhighlight

With patch applied
fixedhighlight

### Test Coverage

No automated tests changed or added.

Manually tested on:
* Desktop Chrome
* Desktop Firefox

### Documentation

No doc updates

### Additional Information

I discussed this in person with Abby at the Blockly Summit Hackathon on Oct 11.

### DIFF
--- a/core/keyboard_nav/cursor_svg.js
+++ b/core/keyboard_nav/cursor_svg.js
@@ -511,6 +511,12 @@ Blockly.CursorSvg.prototype.draw = function(curNode) {
   } else if (curNode.getType() === Blockly.ASTNode.types.STACK) {
     this.showWithStack_(curNode);
   }
+
+  // Ensures the cursor will be visible immediately after the move.
+  var animate = this.currentCursorSvg.childNodes[0];
+  if (animate !== undefined) {
+    animate.beginElement && animate.beginElement();
+  }
 };
 
 /**


core/keyboard_nav/cursor_svg.js
```diff
@@ -511,6 +511,12 @@ Blockly.CursorSvg.prototype.draw = function(curNode) {
   } else if (curNode.getType() === Blockly.ASTNode.types.STACK) {
     this.showWithStack_(curNode);
   }
+
+  // Ensures the cursor will be visible immediately after the move.
+  var animate = this.currentCursorSvg.children[0];
```
samelhusseini: Node.children is unavailable on IE 11. Use childNodes instead.

thomasballinger:
@googlebot I signed it!

core/keyboard_nav/cursor_svg.js
thomasballinger: ah thanks good to know!

core/keyboard_nav/cursor_svg.js
```diff
@@ -511,6 +511,12 @@ Blockly.CursorSvg.prototype.draw = function(curNode) {
   } else if (curNode.getType() === Blockly.ASTNode.types.STACK) {
     this.showWithStack_(curNode);
   }
+
+  // Ensures the cursor will be visible immediately after the move.
+  var animate = this.currentCursorSvg.childNodes[0];
+  if (animate !== undefined) {
+    animate.beginElement();
```
alschmiedt: It looks like this line will break things in ie and edge since they don't support animate. So change this to animate.beginElement && animate.beginElement() and then this should be good to go!

thomasballinger: Thanks @alschmiedt!

alschmiedt: Thanks for contributing!

PR #3239, opened 2019-10-14
Update version numbers

rachel-fenichel:


samelhusseini:
The version in the blockly_compressed is pulled from the package.json.
We need a system for that for release candidates, but for now, we’ll need to rebuild here to get that latest version bundled in.

rachel-fenichel:
Right, I'll rebuild.

PR #3252, opened 2019-10-15
Fix blocks tests

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Blocks tests are broken because no goog. in blockly_compressed.

``/tests/blocks/index.html``

### Proposed Changes

Use uncompressed instead.

### Reason for Changes

Fix test

### Test Coverage

Tested on:

### DIFF
--- a/tests/blocks/index.html
+++ b/tests/blocks/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>Unit Tests for Blockly Blocks</title>
-    <script src="../../blockly_compressed.js"></script>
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
LG, though the reason confuses me--didn't we get rid of goog in uncompressed as well?

samelhusseini:
No, good.provide / require are needed for the compiler to do its thing.
When running uncompressed files we still need closure’s base.js which has definitions for good.provide and require.
Since this test includes the blocks files directly, it needs those methods defined and hence why just switching to uncompressed fixes things.

Issue #2239, opened 2019-01-28
Disabled blocks can't be retrieved from the trash.


NeilFraser:
1. Open https://beta-dot-blockly-games.appspot.com/music
2. Drag note block from toolbox to workspace as an orphan so it gets disabled.
3. Drag disabled block to trash.
4. Open trash and attempt to retrieve the disabled block.

Disabled blocks can't be pulled out of the trash.

NeilFraser:
@BeksOmega I think this is yours?

BeksOmega:
Before I fix this (and #2238 ) I wanted to lay out another option: disabled blocks could be re-enabled when they are added to the trashcan.

I've never asked, but I assume the main usecases for disabled blocks are A) user debugging and B) blocks in the incorrect context? If this is true I don't think it'd be confusing that when the blocks are taken out of their debugging/incorrect context (i.e. put into the trashcan) they would be reenabled.

Opinions?

rachel-fenichel:
Yeah, we discussed that briefly but were split on whether we wanted re-enabling to happen.  For now let's leave them disabled.

NeilFraser:
Fix confirmed.  Thanks!

PR #3255, opened 2019-10-15
Fixed disabled blocks in trashcan. Also fixed block comments.

### LINKED ISSUE #2574: Trashcan Does Not Properly Ignore Comment Properties
### Problem statement

Comments with different height, width, and pinned statuses will cause duplicate blocks in the trashcan, even if the text of the comment is identical.

### Expected Behavior

No duplicate blocks in the trashcan.

### Actual Behavior

Sometimes duplicate blocks.

### Steps to Reproduce

1. Load the following xml into the playground
```
  <block type="controls_if">
    <comment pinned="false" h="80" w="160">Comment Text</comment>
  </block>
  <block type="controls_if">
    <comment pinned="true" h="80" w="160">Comment Text</comment>
  </block>
  <block type="controls_if">
    <comment pinned="false" h="160" w="320">Comment Text</comment>
  </block>
</xml>
```
2. Delete all blocks.
3. Open the trashcan and observe how there are multiple identical looking blocks.

### Additional Information

N/A

### LINKED ISSUE #3243: Trashcan duplicates disabled blocks when maxBlocks is hit
- Go to https://blockly.games/maze?level=3
- Drag a 'move forward' block from the toolbox to the trashcan.
- Drag a 'move forward' block from the toolbox to the workspace (this trips maxBlocks).
- Open the trashcan and drag the 'move forward' block back into the trash.
- Open the trashcan.

Expected result: one disabled 'move forward' block.

Actual result: two disabled 'move forward' blocks.

Notes: One of the blocks in he trash flyout is disabled, the other is enabled, but as a result of maxBlocks getting hit everything is disabled.

More importantly, this application does not have user-selectable enabling and disabling of blocks.  Which means a disabled block pulled from the trashcan is forever useless.

Suggestion: If disabled:false is in the config, then one should not be able to drag disabled blocks out of the trashcan.  Currently one can't drag disabled blocks out of a toolbox flyout.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #3243
Closes #2574 

### Proposed Changes

* Disabled tags are removed from blocks when they are added to the trashcan.
* h, w, and pinned tags are removed from block-comments when they are added to the trashcan.

### Reason for Changes

* See reasoning [here](https://github.com/google/blockly/issues/2239#issuecomment-458290090). I was going to go with the alternative suggestion [here](https://github.com/google/blockly/issues/3243#issuecomment-541941991), but I believe that would allow users to get around the maxBlocks/maxInstances if disable/enable was available.
* This allowed "duplicate" blocks that looked the same to exist in the trashcan.

### Test Coverage

Modified unit tests to reflect changes.

#### Testing disableOrphans
1. Added disableOrphans listener to playground.
2. Dragged blocks onto the workspace.
3. Dragged them into the trashcan.
4. Observed how they were re-enabled in the trashcan.
5. Dragged them out of the trashcan.
6. Observed how they were re-disabled.

#### Testing maxBlocks
1. Set maxBlocks to 1.
2. Dragged a block from the toolbox to the trashcan.
3. Added a block to the workspace. This trips maxBlocks.
4. Opened the trashcan.
5. Observed how I was unable to add the block to the workspace.

#### Testing Comments
1. Followed the procedure from #2574.
2. Was not able to reproduce the issue.

Tested on:
 * Desktop Chrome 

### Documentation

N/A

### Additional Information

@NeilFraser does the solution to #3243 look ok to you?

### DIFF
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -497,6 +497,11 @@ Blockly.Trashcan.prototype.cleanBlockXML_ = function(xml) {
       node.removeAttribute('y');
       node.removeAttribute('id');
       node.removeAttribute('disabled');
+      if (node.nodeName == 'comment') {  // Future proof just in case.
+        node.removeAttribute('h');
+        node.removeAttribute('w');
+        node.removeAttribute('pinned');
+      }
     }
 
     // Try to go down the tree
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -287,11 +287,6 @@ Blockly.Trashcan.prototype.init = function(verticalSpacing) {
     Blockly.utils.dom.insertAfter(this.flyout_.createDom('svg'),
         this.workspace_.getParentSvg());
     this.flyout_.init(this.workspace_);
-    this.flyout_.isBlockCreatable_ = function() {
-      // All blocks, including disabled ones, can be dragged from the
-      // trashcan flyout.
-      return true;
-    };
   }
 
   this.verticalSpacing_ = this.MARGIN_BOTTOM_ + verticalSpacing;
@@ -501,6 +496,7 @@ Blockly.Trashcan.prototype.cleanBlockXML_ = function(xml) {
       node.removeAttribute('x');
       node.removeAttribute('y');
       node.removeAttribute('id');
+      node.removeAttribute('disabled');
     }
 
     // Try to go down the tree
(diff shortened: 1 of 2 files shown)


NeilFraser:
Looks good.

For the record, the only risk here (already identified by @BeksOmega) is an application with *all* of the following conditions:

* a disabled block on the workspace that the user can delete and recreate in the enabled state
* disable:false in the config (since otherwise the user could perform the action themselves using the context menu).
* the disabled state of the block was not set by disableOrphan or maxBlocks or any other event handler (since those would re-disable the block from the trash).
* the disabled block was pre-placed on the workspace (since one can't create a disabled block from the toolbox).
* the disabled block was deletable.

I can't think of any case for such an application.  And if this comes up, it is simple enough to add an event handler to the offending block that re-disables it.

PR #3261, opened 2019-10-16
Cleanup aria usage.

NeilFraser:
Use enums instead of arbitrary strings.
Comment out unused enum values.
Add missing requires.
Delete unused code.

### DIFF
--- a/core/components/menu/menu.js
+++ b/core/components/menu/menu.js
@@ -92,7 +92,7 @@ Blockly.Menu.prototype.blur = function() {
 
 /**
  * Set the menu accessibility role.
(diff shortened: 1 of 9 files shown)


samelhusseini:
Eslint complaining about the commented out code, just add an extra space after the //

NeilFraser:
It is worth noting that these are not [all the Aria properties](https://a11yproject.com/posts/ARIA-properties/), just the subset that Closure uses.  Since this set is pretty arbitrary from our point of view, I think I'd rather delete the ones that Closure uses but we don't, rather than comment them out.  Sound good?

samelhusseini:
Yep, that sounds fine by me.

Issue #3260, opened 2019-10-15
Proposal: Standard dynamic plugin system for Blockly.


jollytoad:
**Is your feature request related to a problem? Please describe.**

Many customisations of Blockly require hacking of the core, either by forking or monkey patching.
There is also no standard way to write and publish a customisation (which I'll refer to as 'plugin' herein) for Blockly.

**Describe the solution you'd like**

A standard API within Blockly which can be used to register and unregister plugins into Blockly dynamically.
Blockly core and other plugins can define named hooks that any plugin can supply.
Plugins may also run initialisation and disposal code.
Plugins may supply Aspects (AOP) to augment behaviour of Blockly core features that don't leverage hooks.
Plugins can be implemented with any build system or transpiled language so long as they return a standard plugin spec object.
Plugins can be agnostic to the origin of Blockly itself - ie. we pass Blockly to the plugin.

**Describe alternatives you've considered**

**Additional context**

I've already started thinking about this, so i'll post some ideas later.

jollytoad:
I've hacked an initial stab at the plugin system API, I've used TypeScript to define it, as I'm more familiar with this, although I'm not suggesting it will be implemented in TS.

I've put this in a gist so it can be evolved... https://gist.github.com/jollytoad/841545a2e55cee8c5b56d53f464a2b09

I'm not sure everything i've added would definitely be necessary. I'm going to try to implement the basics of the API and some example plugins.

I'd love to get some early feedback on this and whether it's something that would be considered for core Blockly, and whether other Blockly users would make use of it.

jollytoad:
Take a look at my branch in the PR linked above for examples.

jollytoad:
This could help with https://github.com/google/blockly/issues/3023 allowing features to be loaded/unloaded on demand

RoboErikG:
Quick update: We've been discussing this internally and using your PR as one option for making this possible. Here's the summary of our internal discussions so far.

- We aren't really comfortable with making whole chunks of the library arbitrarily replaceable. This could make compatibility and maintenance a big problem as it's harder to define clear API boundaries with this approach.
- We do like the way Rendering is being done, using an API for registering/replacing components that conform to a specific API and extend a base class. We will probably try this approach with a few more components to see how well it generalizes.
[...]

PR #3266, opened 2019-10-16
Experimental plugin system. For demo purposes only, not for merging.

jollytoad:
This PR is meant to demonstrate the idea of a plugin system.

I've added the basic plugin system API, and migrated all Block ContextMenu items to each be individual plugins. So you can potentially add more items in the same way or even easily disable an item by simply unregistering the plugin, for example:

`Blockly.Plugins.unregister('block_menu_comment')`

I've rebuilt only the uncompressed js in this branch so the playground for this should work out of the box.

See https://github.com/google/blockly/issues/3260


jollytoad:
So, there are now some plugin examples in the playground of:

* block context menu item - right click on a block to see the example menu item
* aspects with before & after advices - create a variable to see log messages generated from the before/after advices

To see how a plugin can be dynamically unregistered, from the console try:

`Blockly.Plugins.unregister('block_menu_example')`
`Blockly.Plugins.unregister('aspect_example')`

typings/plugins.d.ts
```diff
@@ -0,0 +1,163 @@
+import * as Blockly from "./blockly";
```
jollytoad: Please note: This file does not represent what is implemented so far, it's just a potential API.

rachel-fenichel:
Closing since we picked a plugin system and are running with it.

PR #3267, opened 2019-10-16
Fix #1714

### LINKED ISSUE #1714: Get rid of code that handles Safari 7.1.3
In xml.js, there's a [comment](https://github.com/google/blockly/blob/develop/core/xml.js#L385):
```
  // Safari 7.1.3 is known to provide node lists with extra references to
  // children beyond the lists' length.  Trust the length, do not use the
  // looping pattern of checking the index for an object.
  var childCount = xml.childNodes.length;
```

Safari 7.1.3 is effectively deprecated now.  We're also not consistent with this approach across the rest of Blockly.  We should just look at the actual node list.

@NeilFraser if you object, speak now.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1714 
### Proposed Changes

Get rid of code that handles a Safari 7 bug.

### Reason for Changes

Bug has since been fixed.

### Test Coverage
Tested serialization and deserialization in the playground in chrome, but not safai.

Neil confirmed that the bug was no longer present in Safari 11.

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -389,10 +389,6 @@ Blockly.Xml.domToWorkspace = function(xml, workspace) {
   }
   var newBlockIds = [];  // A list of block IDs added by this call.
   Blockly.utils.dom.startTextWidthCache();
-  // Safari 7.1.3 is known to provide node lists with extra references to
-  // children beyond the lists' length.  Trust the length, do not use the
-  // looping pattern of checking the index for an object.
-  var childCount = xml.childNodes.length;
   var existingGroup = Blockly.Events.getGroup();
   if (!existingGroup) {
     Blockly.Events.setGroup(true);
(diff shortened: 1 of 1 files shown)


core/xml.js
```diff
@@ -404,8 +400,7 @@ Blockly.Xml.domToWorkspace = function(xml, workspace) {
   }
   var variablesFirst = true;
   try {
-    for (var i = 0; i < childCount; i++) {
-      var xmlChild = xml.childNodes[i];
+    for (var i = 0, xmlChild; xmlChild = xml.childNodes[i]; i++) {
```
mattbishop: Isn't this supposed to be `(xmlChild = xml.childNodes[I];)` now?

samelhusseini: Yeah. Neil took care of that in #3381

PR #3269, opened 2019-10-16
Add node’s global objects to reserved list

NeilFraser:
I skipped adding DTRACE_NET_SERVER_CONNECTION, DTRACE_NET_STREAM_END, DTRACE_HTTP_SERVER_REQUEST, DTRACE_HTTP_SERVER_RESPONSE, DTRACE_HTTP_CLIENT_REQUEST, DTRACE_HTTP_CLIENT_RESPONSE since they are unlikely to collide with a user variable, they don’t look important if they are clobbered, and I found [an indication](https://gitlab.swt-2-2.dsalab.de/cherbib/node/commit/f9ddbb6b2f2ff61bb62a4ee8a8819d561322c9e8) that these may be going away soon.


generators/javascript.js
```diff
@@ -55,14 +55,16 @@ Blockly.JavaScript.addReservedWords(
...
-    'SVGAnimatedPoints,SVGColorProfileRule,SVGCSSRule,SVGExternalResourcesRequired,SVGFitToViewBox,SVGLangSpace,SVGLocatable,SVGRenderingIntent,SVGStylable,SVGTests,SVGTextContentElement,SVGTextPositioningElement,SVGTransformable,SVGUnitTypes,SVGURIReference,SVGViewSpec,SVGZoomAndPan');
+    'SVGAnimatedPoints,SVGColorProfileRule,SVGCSSRule,SVGExternalResourcesRequired,SVGFitToViewBox,SVGLangSpace,SVGLocatable,SVGRenderingIntent,SVGStylable,SVGTests,SVGTextContentElement,SVGTextPositioningElement,SVGTransformable,SVGUnitTypes,SVGURIReference,SVGViewSpec,SVGZoomAndPan,' +
+    // Node.js
+    'global,process,Buffer,module,require');
```
samelhusseini: This looks good, while you're here if you wanted to add ES6 future reserved words: 
http://www.ecma-international.org/ecma-262/6.0/index.html#sec-future-reserved-words
``await``

NeilFraser:
Change of plan.

Mozilla updated their site, so I went through the links to update our list.  But the more I poked at it, the more I realized that it's impossible to compile a list of all globals in all environments.  There's too much variation and custom extensions.

So I ripped out all but the grammar keywords, and populated the rest of the list by scanning the global object's properties.  This has three effects:
* The size is much smaller (7.4 KB to 0.3 KB).
* All properties in the current environment will be picked up, including developer-created global variables and every weird vendor-specific extension in the user's browser.
* Code is no longer portable.  JS code generated on Node may collide with global variables when shipped to a user for eval().

This last effect is why I refrained from doing this in the first place.  However, today's discovery that we can't make a definitive list means that our existing strategy isn't even portable to the current machine, let alone any others.

NeilFraser:
It's also worth noting that if one is using the JS Interpreter, there is no problem with code portability.  JS Interpreter has no global namespace pollution, so generating in any browser or any version of Node will work just fine.

And if one is generating on Node, and targeting some other system, there's nothing stopping a developer from providing their own list of reserved words through the existing API.

NeilFraser:
> Looks good. Would there perhaps be a way to run the 'Object.getOwnPropertyNames' at the point of eval, just in case they're not evaling their code in the same context as Blockly (eg: JSInterpreter).

It's technically possible.  Every program would be rendered as a JavaScript string with all variables and names prefixed and suffixed by some Unicode character.  Then there would be a snippet of code that scans the local environment and uses regexes to find and rename all the variables, before evaling the user's code.  Certainly would go against the goal of generating readable code.

A survey globals in various environments (in 2019):

-   JS Interpreter: 36
-   Node: 104
-   Safari: 778
-   Firefox: 785
-   Chrome: 835

The good news is that the JS Interpreter has a smaller global variable footprint than any other environment, and that footprint is a subset of all the other environments.  So generation anywhere will produce perfectly good names for the JS Interpreter.

PR #3271, opened 2019-10-16
Fix jsdoc. @return instead of @returns

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Clean up.

### Proposed Changes

Fix jsdoc. @return instead of @returns

### Test Coverage

Tested on:


samelhusseini:
Yep, will fix. thanks for pointing that out.

PR #3276, opened 2019-10-17
Fix 11 warnings related to block_svg

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fix 11 warnings related to block_svg

### Proposed Changes

Compiler warning fixes.

### Test Coverage

Tested on:

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -613,13 +613,11 @@ Blockly.Field.prototype.render_ = function() {
 };
 
 /**
- * Show an editor when the field is clicked only if the field is clickable.
+ * Show an editor when the field is clicked.
  * @package
  */
 Blockly.Field.prototype.showEditor = function() {
-  if (this.isClickable()) {
-    this.showEditor_();
-  }
+  this.showEditor_ && this.showEditor_();
 };
 
 /**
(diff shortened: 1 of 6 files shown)


core/field.js
```diff
@@ -232,7 +232,7 @@ Blockly.Field.prototype.getText_;
  *     clicked. Blockly will automatically set the field as clickable if this
  *     method is defined.
  * @return {void}
- * @protected
+ * @package
  */
 Blockly.Field.prototype.showEditor_;
```
alschmiedt: Remove the _

samelhusseini: I believe this one will have to fall under an exception as too many fields out there use this as public API, so changing it will break a lot of people.
I propose either: 
- a, doing the above. Leave the name as is, and change it to @package.
- b, Leave it as @protected, and add a wrapper. something like showEditor that is public and just calls showEditor_ internally.

alschmiedt: Ah, ok. It might be nice to have a wrapper so new people can use something that doesn't look private.

samelhusseini: Done.

PR #3277, opened 2019-10-17
Fix compiler warnings related to XML

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fix compiler warnings.

### Proposed Changes

Use Node instead of Element for XML elements. Fix 5 warnings.

### Test Coverage

Tested on:

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -392,10 +392,10 @@ Blockly.Field.prototype.bindEvents_ = function() {
 };
 
 /**
- * Sets the field's value based on the given XML element. Should only be called
- * by Blockly.Xml.
- * @param {!Element} fieldElement The XML element containing info about the
- *     field's state.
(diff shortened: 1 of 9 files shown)


core/field.js
```diff
@@ -394,8 +394,8 @@ Blockly.Field.prototype.bindEvents_ = function() {
 /**
  * Sets the field's value based on the given XML element. Should only be
  * called by Blockly.Xml.
- * @param {!Element} fieldElement The element containing info about the
- *    field's state.
+ * @param {!Node} fieldElement The element containing info about the field's
```
rachel-fenichel: Update comments to stop calling it an element if it's not one.  Here and below.

samelhusseini:
Change of plans. Element is indeed the right type to use, we just need to cast child nodes (that are elements) to the Element type before passing them down the chain.

PR #3278, opened 2019-10-17
Fix 11 warnings related to variables and variable_model.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fix compiler warnings.

### Proposed Changes

Fix 11 warnings related to variables and variable_model.

### Test Coverage

Tested on:


core/variables.js
```diff
@@ -500,10 +500,11 @@ Blockly.Variables.getOrCreateVariablePackage = function(workspace, id, opt_name,
  */
```
moniika: Since the return is explicitly null-able, I think should be changed to {?Blockly.VariableModel} (although I know its not a new thing)

samelhusseini: ``Blockly.VariableModel`` is the same as ``?Blockly.VariableModel``, so there's no need to add that

Issue #3279, opened 2019-10-17
Blockly.BlockSvg.START_HAT is broken


duzc2:
**Describe the bug**
if I set Blockly.BlockSvg.START_HAT = true
all blocks have a hat.

**To Reproduce**
just 
Blockly.BlockSvg.START_HAT = true

**Expected behavior**

only blocks with no previous or output connections have a 'hat' on top

**Screenshots**
![bug](https://user-images.githubusercontent.com/564778/66994479-e7ee6d80-f0ff-11e9-945a-65e72365a187.png)

**Desktop (please complete the following information):**
 - OS: windows 10
 - Browser chrome
 - Version 75

**Additional context**

I found someting like :
![bug1](https://user-images.githubusercontent.com/564778/66994809-79f67600-f100-11e9-8a7a-a5bf34b73f35.png)
it must think about hasPrevious and output.

rachel-fenichel:
![image](https://user-images.githubusercontent.com/13686399/67052825-3f86dc00-f0f4-11e9-841d-4318e27fc798.png)

samelhusseini:
This has been fixed and cherry-picked to master. Thanks for filing!

PR #3280, opened 2019-10-17
Only add a hat if there's no output connection

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3279

### Proposed Changes
Only add a hat if there's no output connection

### Reason for Changes

Rendering fix.

### Test Coverage

Tested on the playground by changing the ``Blockly.BlockSvg.START_HAT = true``

Tested on:
* Desktop Chrome

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -253,10 +253,12 @@ Blockly.blockRendering.RenderInfo.prototype.createRows_ = function() {
  * @package
  */
 Blockly.blockRendering.RenderInfo.prototype.populateTopRow_ = function() {
-  var hasHat = this.block_.hat ?
-      this.block_.hat === 'cap' : Blockly.BlockSvg.START_HAT;
   var hasPrevious = !!this.block_.previousConnection;
-  var hasOutputConnection = !!this.outputConnection;
+  var hasPreviousBlock = hasPrevious &&
+    !!this.block_.previousConnection.targetConnection;
+  var hasHat = (this.block_.hat ?
+    this.block_.hat === 'cap' : Blockly.BlockSvg.START_HAT) &&
+    !this.outputConnection && !hasPreviousBlock;
   var leftSquareCorner = this.topRow.hasLeftSquareCorner(this.block_);
 
   if (leftSquareCorner) {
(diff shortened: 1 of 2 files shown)


core/renderers/common/info.js
```diff
@@ -266,7 +267,7 @@ Blockly.blockRendering.RenderInfo.prototype.populateTopRow_ = function() {
         new Blockly.blockRendering.RoundCorner(this.constants_));
   }
 
-  if (hasHat) {
+  if (hasHat && !hasOutputConnection) {
```
rachel-fenichel: I would fold the !hasOutputConnection into the definition of hasHat.

samelhusseini: There's actually more going on here. We should also consider whether we're connected to a block. eg, this is what you get: 

Screen Shot 2019-10-17 at 9 22 26 AM

rachel-fenichel: no prev, no output, and start hat is set?

samelhusseini: Fixed. Since we're checking for start hat in multiple places, thoughts on moving that check onto the block? ``block.hasHat()``?

Screen Shot 2019-10-17 at 9 41 32 AM

samelhusseini: Actually if we want renderers to be able to change this behaviour we probably shouldn't.

rachel-fenichel: This is still wrong--the top repeat block should not have a hat.  The check is whether there's a previous connection, not whether there's a previous block.

RoboErikG:
Why do we want unconnected blocks with previous connections to have a hat? That's still a change in behavior.

core/renderers/common/info.js
samelhusseini: Yep, misunderstanding on my part. Updated to only show when there's no previous connection and no output connection.

core/renderers/measurables/rows.js
```diff
@@ -283,9 +283,10 @@ Blockly.utils.object.inherits(Blockly.blockRendering.TopRow,
  * @return {boolean} Whether or not the top row has a left square corner.
  */
 Blockly.blockRendering.TopRow.prototype.hasLeftSquareCorner = function(block) {
-  var hasHat = block.hat ? block.hat === 'cap' : Blockly.BlockSvg.START_HAT;
+  var hasHat = (block.hat ? block.hat === 'cap' : Blockly.BlockSvg.START_HAT) &&
+    !block.outputConnection && !block.previousConnection;
   var prevBlock = block.getPreviousBlock();
-
+
```
RoboErikG: nit, extra spaces.

Issue #3216, opened 2019-10-11
Add drag start/end events


RoboErikG:
**Is your feature request related to a problem? Please describe.**

You might want to update some of the UI in response to a drag starting. Adding drag start/end events would be one way to allow this.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #3282, opened 2019-10-17
Add dragStart and dragEnd UI events

navilperez:
## The basics

- [x ] I branched from develop

### Resolves

Addresses #3216, creating UI events for the start and end of a drag.

### Proposed Changes
Creates UI events "dragStart" and "dragEnd" which are fired at the start and end of a drag respectively.
"dragStart"

### Reason for Changes
Provide support for updating the UI in response to a block being dragged.

### Test Coverage

Tested on:
Desktop Chrome

### Documentation

The Events document needs to be updated to include "dragStart" and "dragEnd" to the list of valid element values.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -195,7 +195,7 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
 Blockly.BlockDragger.prototype.fireDragStartEvent_ = function() {
   var event = new Blockly.Events.Ui(this.draggingBlock_, 'dragStart',
       null, this.draggingBlock_.getDescendants(false));
-  Blockly.Events.fire(event)
+  Blockly.Events.fire(event);
 };
 
 /**
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -189,7 +189,7 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
 };
 
 /**
- * Fire a ui event at the start of a block drag.
+ * Fire a UI event at the start of a block drag.
  * @private
  */
 Blockly.BlockDragger.prototype.fireDragStartEvent_ = function() {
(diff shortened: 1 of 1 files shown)


RoboErikG:
@googlebot check again

core/block_dragger.js
```diff
@@ -186,6 +188,16 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
   }
 };
 
+/**
+ * Fire a ui event at the start of a block drag.
+ * @private
+ */
+Blockly.BlockDragger.prototype.fireDragStartEvent_ = function() {
+  var event = new Blockly.Events.Ui(this.draggingBlock_, 'dragStart',
+      null, this.draggingBlock_.getDescendants(false));
```
samelhusseini: Couple of questions: 
- I'm curious why we're passing newValue here with the list of blocks, whereas oldValue on the end event below. Looking at other examples of UI events we generally only pass values to old / new if some value has changed.
- I'm also curious why we're passing this.draggingBlock_.getDescendants(false) at all? We don't seem to do that in any other events.

Also, could we perhaps do this with the one event name? ``drag`` and pass ``false,true`` as we start the drag, and ``true, false`` as we end the event.

RoboErikG: It's passing the list of blocks that are part of the drag. If we want to use this event for highlighting or locking blocks that are currently being dragged we'll want the full list of ids.

It's the new value when the drag starts, because those are the blocks being dragged, and the old value at the end because they are no longer being dragged. We could do this as a single drag event and just update the list of ids that are being dragged instead of different start and end events, though if we ever re-implement multi-drag that would be a problem.

samelhusseini: That makes sense. Thanks @RoboErikG

core/block_dragger.js
```diff
@@ -186,6 +188,16 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
   }
 };
 
+/**
+ * Fire a ui event at the start of a block drag.
```
samelhusseini: ui events are referenced as ``UI events`` elsewhere. Let's keep that consistent.

core/block_dragger.js
```diff
@@ -252,6 +265,16 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   Blockly.Events.setGroup(false);
 };
 
+/**
+ * Fire a ui event at the end of a block drag.
```
samelhusseini: nit: same, replace ui with ``UI``.

navilperez:
Thanks Team :)

PR #3286, opened 2019-10-17
Add missing utils.global

NeilFraser:
Build is breaking due to the JS environment sweep (#3269).


samelhusseini:
Also add the replace in the local build script (gulpfile.js):
https://github.com/google/blockly/blob/develop/gulpfile.js#L188

PR #3287, opened 2019-10-17
Make finished-loading events be UI events

NeilFraser:
Currently the playground throws when a block is deleted and the delete is undone.

### DIFF
--- a/core/events.js
+++ b/core/events.js
@@ -390,6 +390,9 @@ Blockly.Events.fromJson = function(json, workspace) {
     case Blockly.Events.COMMENT_DELETE:
       event = new Blockly.Events.CommentDelete(null);
       break;
+    case Blockly.Events.FINISHED_LOADING:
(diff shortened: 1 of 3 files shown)


RoboErikG:
Change looks fine, but where are you seeing an error when undoing a delete?

NeilFraser:
> where are you seeing an error when undoing a delete?

Ah, the mirror demo, not playground.  Create a block, delete it, undo:
```
blockly_compressed.js:43 Uncaught Error: Unknown event type.
    at Object.Blockly.Events.fromJson (blockly_compressed.js:43)
    at mirrorEvent (index.html:75)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (blockly_compressed.js:239)
    at Blockly.Events.fireNow_ (blockly_compressed.js:38)
```

This PR solves the error twice, once by supporting the JSON parsing of the finished_loading event, and once by not invoking the JSON round-trip for finished_loading in the first place.

PR #3315, opened 2019-10-22
Handle 0x hex values in color parsing

shakao:
## The basics

- [x] I branched from develop

### Resolves

Pulls https://github.com/microsoft/pxt-blockly/pull/255. (Sorry for resubmitting, I messed something up in the other one!)

### Proposed Changes

Allows Blockly.utils.colour.parse to handle "0xffffff" hex formatting. Calls "parse" in the hexToRgb to handle cases of "#fff", "0xfff" etc.

### Reason for Changes

Additional flexibility in defining colors.

### Test Coverage

Modified the classic theme with color values of "0x######" format, ran it in 

Tested on:
Desktop Chrome
Desktop Firefox
Windows Internet Explorer 11
Windows Edge

### Documentation

I updated the function description to include the supported format.

### DIFF
--- a/core/utils/colour.js
+++ b/core/utils/colour.js
@@ -89,12 +89,13 @@ Blockly.utils.colour.rgbToHex = function(r, g, b) {
 };
 
 /**
- * Converts a hex representation of a colour to RGB.
- * @param {string} hexColour Colour in '#ff0000' format.
+ * Converts a colour to RGB.
+ * @param {string} colour String representing colour in any
+ *     colour format ('#ff0000', 'red', '0xff000', etc).
  * @return {!Array.<number>} RGB representation of the colour.
  */
-Blockly.utils.colour.hexToRgb = function(hexColour) {
-  var hex = Blockly.utils.colour.parse(hexColour);
+Blockly.utils.colour.hexToRgb = function(colour) {
+  var hex = Blockly.utils.colour.parse(colour);
   if (!hex) {
     return [0, 0, 0];
   }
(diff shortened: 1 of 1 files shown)


core/utils/colour.js
```diff
@@ -92,7 +94,12 @@ Blockly.utils.colour.rgbToHex = function(r, g, b) {
  * @return {!Array.<number>} RGB representation of the colour.
  */
 Blockly.utils.colour.hexToRgb = function(hexColour) {
-  var rgb = parseInt(hexColour.substr(1), 16);
+  var hex = Blockly.utils.colour.parse(hexColour);
```
samelhusseini: Given this function converts a hex to rgb, I wouldn't want to pass it through colour.parse, instead we could do the following: 
```js
  hexColour = hexColour.substring(0, 2) == '0x' ?
    hexColour.substring(2) : hexColour.substring(1);
```
and then set rgb to:
```js
  var rgb = parseInt(hexColour, 16);
```

I would also update the @param to:
```js
 * @param {string} hexColour Colour in '#ff0000' or '0xff0000' format.
```

shakao: hm, this means hex shorthand ("#fff") would parse incorrectly though. i guess the documentation could be updated to exclude that input format?

samelhusseini: Fair enough. Let's update the jsdoc and parameter name to accept a colour. (instead of hexColour), and pass it through the parse method as you're doing.

PR #3317, opened 2019-10-22
Adding pinch zoom config option.

### LINKED ISSUE #2782: Pinch to zoom can't be configured / turned off
**Describe the bug**

Pinch to zoom doesn't respect zoom options in order to figure out whether or not to zoom the workspace. 

**To Reproduce**

1. Configure the playground without any zoom options.
2. Browse to the playground on a touch device. 
3. Try to pinch to zoom the workspace
4. It works, when it shouldn't..

**Expected behavior**

I expect a developer can turn off pinch to zoom with the zoom options.

 - Device: Iphone X
 - OS: IOS 12
 - Browser: Safari

**Additional context**

Currently there is no zoom option for pinch to zoom. We have a few options: 
- We could overload wheel. 
- We could create a new option.
- We could just check for the existence of the zoom option (ie: any zoom option, enables it). 

I feel if we create a zoom option for it now, we'd have to communicate to people that they have to now opt into it through that option.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#2782

### Proposed Changes

Adds pinch zoom option.

### Reason for Changes

Allows for pinch zoom to be configured.

### Test Coverage

Tested on:
* Desktop Chrome

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -247,6 +247,11 @@ Blockly.Options.parseZoomOptions_ = function(options) {
   } else {
     zoomOptions.scaleSpeed = Number(zoom['scaleSpeed']);
   }
+  if (zoom['pinch'] === undefined) {
+    zoomOptions.pinch = zoomOptions.wheel || zoomOptions.controls;
+  } else {
+    zoomOptions.pinch =!! zoom['pinch'];
+  }
   return zoomOptions;
 };
 
--- a/core/options.js
+++ b/core/options.js
@@ -250,7 +250,7 @@ Blockly.Options.parseZoomOptions_ = function(options) {
   if (zoom['pinch'] === undefined) {
     zoomOptions.pinch = zoomOptions.wheel || zoomOptions.controls;
   } else {
-    zoomOptions.pinch =!! zoom['pinch'];
+    zoomOptions.pinch = !!zoom['pinch'];
   }
   return zoomOptions;
 };
--- a/core/touch_gesture.js
+++ b/core/touch_gesture.js
@@ -266,7 +266,7 @@ Blockly.TouchGesture.prototype.handleTouchMove = function(e) {
 
   var pointers = Object.keys(this.cachedPoints_);
   if (this.isPinchZoomEnabled_ && pointers.length === 2) {
-    this.handlePinch(e);
+    this.handlePinch_(e);
   } else {
     Blockly.TouchGesture.superClass_.handleMove.call(this, e);
   }
@@ -275,9 +275,9 @@ Blockly.TouchGesture.prototype.handleTouchMove = function(e) {
 /**
 * Handle pinch zoom gesture.
 * @param {!Event} e A touch move, or pointer move event.
-* @package
+* @private
 */
-Blockly.TouchGesture.prototype.handlePinch = function(e) {
+Blockly.TouchGesture.prototype.handlePinch_ = function(e) {
   var pointers = Object.keys(this.cachedPoints_);
   // Calculate the distance between the two pointers
   var point0 = /** @type {!Blockly.utils.Coordinate} */ (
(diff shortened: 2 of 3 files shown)


core/workspace_svg.js
```diff
@@ -1811,12 +1813,6 @@ Blockly.WorkspaceSvg.prototype.setBrowserFocus = function() {
  *     amount values zoom in.
  */
 Blockly.WorkspaceSvg.prototype.zoom = function(x, y, amount) {
-  // TODO (#2782): Consider removing once pinch understands zoom configuration
```
moniika: Currently testing mutators.

core/options.js
```diff
@@ -247,6 +247,11 @@ Blockly.Options.parseZoomOptions_ = function(options) {
   } else {
     zoomOptions.scaleSpeed = Number(zoom['scaleSpeed']);
   }
+  if (zoom['pinch'] === undefined) {
+    zoomOptions.pinch = zoomOptions.wheel || zoomOptions.controls;
+  } else {
+    zoomOptions.pinch =!! zoom['pinch'];
```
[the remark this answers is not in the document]
moniika: Done.

core/workspace_svg.js
moniika: tested with breakpoints.

RoboErikG:
@moniika is this ready for another review pass?

samelhusseini:
This mostly looks good, I would just cache the whether or not to handlePinch in the constructor of the TouchGesture object.

core/touch_gesture.js
```diff
@@ -256,33 +256,47 @@ Blockly.TouchGesture.prototype.handleTouchMove = function(e) {
   // Update the cache
   this.cachedPoints_[pointerId] = this.getTouchPoint(e);
 
+  var workspace = this.startWorkspace_;
   var pointers = Object.keys(this.cachedPoints_);
-  // If two pointers are down, check for pinch gestures
-  if (pointers.length == 2) {
-    // Calculate the distance between the two pointers
-    var point0 = /** @type {!Blockly.utils.Coordinate} */ (
+  if (workspace.options.zoomOptions && workspace.options.zoomOptions.pinch &&
```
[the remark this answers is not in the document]
moniika: My concern with caching is that it wouldn't be reflected when the value changes, but I'll take a look at where I could potentially add the cache and see what it looks like.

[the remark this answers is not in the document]
samelhusseini: I don't believe we supporting changing workspace options on the fly.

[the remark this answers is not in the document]
moniika: Good to know, it also looks like new gesture is created every time.

core/touch_gesture.js
```diff
@@ -257,32 +265,44 @@ Blockly.TouchGesture.prototype.handleTouchMove = function(e) {
...
+    this.handlePinch(e);
+  } else {
+    Blockly.TouchGesture.superClass_.handleMove.call(this, e);
+  }
+};
+
+/**
+* Handle pinch zoom gesture.
+* @param {!Event} e A touch move, or pointer move event.
+* @package
```
samelhusseini: Why is this @package? And not private

core/touch_gesture.js
```diff
@@ -257,32 +265,44 @@ Blockly.TouchGesture.prototype.handleTouchMove = function(e) {
...
+    this.cachedPoints_[pointers[1]]);
+  var moveDistance = Blockly.utils.Coordinate.distance(point0, point1);
+  var startDistance = this.startDistance_;
+  this.touchScale_ = moveDistance / startDistance;
 
-    if (this.previousScale_ > 0 && this.previousScale_ < Infinity) {
-      var gestureScale = scale - this.previousScale_;
-      var delta = gestureScale > 0 ?
+  if (this.previousScale_ > 0 && this.previousScale_ < Infinity) {
+    var gestureScale = this.touchScale_ - this.previousScale_;
```
samelhusseini: Curious why scale is now stored on the object? I don’t see it read anywhere outside this method

core/touch_gesture.js
moniika: Could be private, changing it now.

core/touch_gesture.js
moniika: It was always stored on the object (see old L269). Given its a private field and I also don't see it anywhere outside of this method, I'm going to change it to a local variable and if we need it in future, we can update it then.

PR #3322, opened 2019-10-23
Color blind themes

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Add color palettes for three types of color blindness: Deuteranopia, Protanopia, and Tritanopia. The Deuteranopia color palette works for Protanopia as well. Also removed modern from the themes in require. Doesn't seem necessary since we still have classic.

Deuteranopia/Protanopia
Screen Shot 2019-10-23 at 12 42 01 PM

Tritanopia
Screen Shot 2019-10-23 at 12 42 15 PM

### Test Coverage

Tested on:


core/requires.js
```diff
@@ -88,5 +88,6 @@ goog.require('Blockly.geras.Renderer');
 // Classic is the default theme.
 goog.require('Blockly.Themes.Classic');
 goog.require('Blockly.Themes.Dark');
+goog.require('Blockly.Themes.Deuteranopia');
 goog.require('Blockly.Themes.HighContrast');
-goog.require('Blockly.Themes.Modern');
```
samelhusseini: Intentionally removing modern?

samelhusseini: I would comment it out instead, similar to requires that aren't included in the rest of this file.

tests/playground.html
```diff
@@ -584,9 +587,10 @@ <h1>Blockly Playground</h1>
   </form>
   <select id="themeChanger" name="theme" onchange="changeTheme()">
     <option value="classic">Classic</option>
-    <option value="modern">Modern</option>
```
samelhusseini: Why remove modern from the playground?

samelhusseini: You can require it separately in the playground.html file

core/requires.js
alschmiedt: done

tests/playground.html
alschmiedt: ok sounds good

PR #3323, opened 2019-10-23
Fix cloud compile build

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fix cloud compile build.

### Proposed Changes

calcdeps.py doesn't understand reqiureType. Stripping it out before sending files up to the closure cloud compiler.

### Test Coverage

Tested on:


samelhusseini:
No, this just means the cloud build doesn’t know about certain types that we don’t include. If we were getting / reading the compile log, it would just give a warning about the missing type information.
But the result is the same.

PR #3324, opened 2019-10-23
[Zelos] Fix dynamic connection shape size

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Dynamic shapes not reporting the correct size information.
Fixes https://github.com/google/blockly/issues/3325

### Proposed Changes

Each connection is able to query its connectionWidth or Height that will in turn call the shape's height / width methods if the shape is dynamic.

Change all references to shape.width and shape.height to getConnectionWidth and getConnectionHeight for consistency.

In Zelos, recompute bounds after setting the connection height and width (once we've figured out what the final height is going to be)

### Test Coverage

Tested on:

### Additional Information

Before:
Screen Shot 2019-10-23 at 2 12 35 PM

After:
Screen Shot 2019-10-23 at 2 11 18 PM


samelhusseini:
@rachel-fenichel this is ready for another look. This ensures all info is there before we do the draw pass, and the draw pass is only reading pre-computed info.
For the dynamic output shapes, I set their initial shape dimensions to 0. Once we've done a full pass in info and have the height of the connection handy, we re-set the shape dimensions on the connections. 
I've moved setting the shape dimensions into a method call so that any dependencies on the height / width can be updated accordingly.

rachel-fenichel:
Thanks for updating this.

LGTM

PR #3334, opened 2019-10-24
Resolve remaining compiler type warnings

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Resolve remaining type mismatch warnings.

### Proposed Changes

Misc fixes. Marked the functional changes in comments.

### Test Coverage

Tested on:

### DIFF
--- a/core/block_drag_surface.js
+++ b/core/block_drag_surface.js
@@ -173,7 +173,7 @@ Blockly.BlockDragSurfaceSvg.prototype.translateSurface = function(x, y) {
  * @return {!Blockly.utils.Coordinate} Current translation of the surface.
  */
 Blockly.BlockDragSurfaceSvg.prototype.getSurfaceTranslation = function() {
-  var xy = Blockly.utils.getRelativeXY(this.SVG_);
+  var xy = Blockly.utils.getRelativeXY(/** @type {!SVGElement} */ (this.SVG_));
   return new Blockly.utils.Coordinate(xy.x / this.scale_, xy.y / this.scale_);
 };
 
@@ -189,11 +189,11 @@ Blockly.BlockDragSurfaceSvg.prototype.getGroup = function() {
 /**
  * Get the current blocks on the drag surface, if any (primarily
  * for BlockSvg.getRelativeToSurfaceXY).
- * @return {!Element|undefined} Drag surface block DOM element, or undefined
- * if no blocks exist.
+ * @return {Element} Drag surface block DOM element, or undefined if no blocks
+ * exist.
  */
 Blockly.BlockDragSurfaceSvg.prototype.getCurrentBlock = function() {
-  return this.dragGroup_.firstChild;
+  return /** @type {Element} */ (this.dragGroup_.firstChild);
 };
 
 /**
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -279,7 +279,7 @@ Blockly.ContextMenu.blockCommentOption = function(block) {
     enabled: !Blockly.utils.userAgent.IE
   };
   // If there's already a comment, add an option to delete it.
-  if (block.comment) {
+  if (block.getCommentIcon()) {
     commentOption.text = Blockly.Msg['REMOVE_COMMENT'];
     commentOption.callback = function() {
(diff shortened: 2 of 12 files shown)


core/contextmenu.js
```diff
@@ -279,7 +279,7 @@ Blockly.ContextMenu.blockCommentOption = function(block) {
     enabled: !Blockly.utils.userAgent.IE
   };
   // If there's already a comment, add an option to delete it.
-  if (block.comment) {
+  if (block.getCommentIcon()) {
```
samelhusseini: block.comment is deprecated.

core/extensions.js
```diff
@@ -114,7 +114,7 @@ Blockly.Extensions.registerMutator = function(name, mixinObj, opt_helperFn,
       if (!Blockly.Mutator) {
         throw Error(errorPrefix + 'Missing require for Blockly.Mutator');
       }
-      this.setMutator(new Blockly.Mutator(opt_blockList));
+      this.setMutator(new Blockly.Mutator(opt_blockList || []));
```
samelhusseini: Functional change, Mutator does not accept null or undefined

core/workspace.js
```diff
@@ -246,9 +246,11 @@ Blockly.Workspace.prototype.refreshTheme = function() {
 Blockly.Workspace.prototype.updateBlockStyles_ = function(blocks) {
   for (var i = 0, block; block = blocks[i]; i++) {
     var blockStyleName = block.getStyleName();
-    block.setStyle(blockStyleName);
-    if (block.mutator) {
-      block.mutator.updateBlockStyle(blockStyleName);
+    if (blockStyleName) {
```
samelhusseini: setStyle doesn't accept null, checking for null before setting.

core/extensions.js
rachel-fenichel: What did it do before?

core/keyboard_nav/ast_node.js
```diff
@@ -293,7 +293,7 @@ Blockly.ASTNode.prototype.findNextForField_ = function() {
   var location = /** @type {!Blockly.Field} */ (this.location_);
   var input = location.getParentInput();
   var block = location.getSourceBlock();
-  var curIdx = block.inputList.indexOf(input);
+  var curIdx = block.inputList.indexOf(/** @type {!Blockly.Input} */ (input));
```
rachel-fenichel: Why isn't the result of getParentInput already of type Blockly.Input?

alschmiedt: It is of type Blockly.Input but not !Blockly.Input. We might be able to change that though? I can't think of a situation where the field is attached to a block, but doesn't have a parent input?

BeksOmega: Yeah that's imposible.

core/extensions.js
samelhusseini: Well it didn't handle not passing a block list. It would have thrown [here](https://github.com/google/blockly/blob/master/core/mutator.js#L134). However, a call to ``checkMutatorDialog_`` guards this code, and that checks to see if the compose / decompose methods exist, if you register a mutator with a compose / decompose and don't pass a block list it will crash.

Tried it and indeed it does.

core/keyboard_nav/ast_node.js
alschmiedt: ok! well looking at this again it looks like getParentInput should still return null in the case that the sourceBlock_ is null so I don't think we can return !Blockly.Input from there.

core/block_drag_surface.js
```diff
@@ -189,11 +189,11 @@ Blockly.BlockDragSurfaceSvg.prototype.getGroup = function() {
 /**
  * Get the current blocks on the drag surface, if any (primarily
  * for BlockSvg.getRelativeToSurfaceXY).
- * @return {!Element|undefined} Drag surface block DOM element, or undefined
- * if no blocks exist.
+ * @return {Element} Drag surface block DOM element, or undefined if no blocks
```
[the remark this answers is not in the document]
samelhusseini: Yeah, Element is just shorthand for !Element|undefined

Issue #3330, opened 2019-10-24
Dummy input aligns only left


pkestefran:
**Describe the bug**

Dummy input currently does not align right (or center).
Last working in 2.20190722.1

**To Reproduce**

1. In blockly factory create a block similar to :

```javascript
Blockly.Blocks['block_type'] = {
  init: function() {
    this.appendDummyInput()
        .appendField("something long");
    this.appendDummyInput()
        .setAlign(Blockly.ALIGN_CENTRE)
        .appendField("a");
    this.appendDummyInput()
        .setAlign(Blockly.ALIGN_RIGHT)
        .appendField("a");
    this.setColour(230);
 this.setTooltip("");
 this.setHelpUrl("");
  }
};
```

**Expected behavior**

Fields in Dummy input should respect selected alignment.

**Screenshots**

Current master:
![2019-10-24_09h47_36](https://user-images.githubusercontent.com/18596276/67464257-54f38d00-f643-11e9-9e23-96c53eee7376.png)

2.20190722.1:
![2019-10-24_09h50_17](https://user-images.githubusercontent.com/18596276/67464581-da773d00-f643-11e9-9f25-6a95dfe512fa.png)

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser Firefox
 - Version 70.0

RoboErikG:
@rachel-fenichel looks like we're missing some logic for dummy input rows. Value inputs and statement inputs do center and right alignment correctly.

@pkestefran are you depending on the alignment in a live app or is this just something you noticed in the developer tools?

![image](https://user-images.githubusercontent.com/11066279/67502446-bb27e200-f63a-11e9-82b0-c36a520b2830.png)

pkestefran:
@RoboErikG The behavior is the same in the live app. I have noticed this while testing new changes in blockly with custom blocks we have in our app. We have multiple right aligned dummy inputs within our blocks. For now we will stay with 2.20190722.1

RoboErikG:
Thanks, that means we may need to do another patch then.

samelhusseini:
We've fixed this and released a patch v3.20191014.4.
Thanks for reporting.

pkestefran:
@samelhusseini I played a bit with test blocks in playground and noticed that the alignment is unfortunately still not respected in Zelos and Minimalist renderers. Geras and Thrasos are ok.

[...]

Issue #338, opened 2016-04-20
Field alignment appears broken when internal inputs are enabled


tmickel:
Example:
https://blockly-demo.appspot.com/static/demos/blockfactory/index.html#kk43hy

Try switching the alignment on the star. No matter the selection, the image is aligned left. Now switch to external inputs - the image is aligned properly.

tmickel:
@NeilFraser I noticed in the code that this is "TODO". Is this something you plan to implement?

rachel-fenichel:
Fixed in scratch-blocks, at least for the basic case:
https://github.com/LLK/scratch-blocks/commit/4ae2bf455042442e95fb660c31040b3945823612v

Next steps:
- Investigate whether we can just pull in the scratch-blocks fix.
- If so, pull the fix in.
- If not, mark as low priority.

rachel-fenichel:
There's one code path for an inline row, and different code paths for external value input rows and dummy input rows.  The branch is here:
```
    if (this.isCollapsed()) {
      this.renderJaggedEdge_(stepObject, row, cursor);
    } else if (row.type == Blockly.BlockSvg.INLINE) {
      this.renderInlineRow_(
          stepObject, row, cursor, connectionPos, inputRows.rightEdge);
    } else if (row.type == Blockly.INPUT_VALUE) {
      this.renderExternalValueInput_(
          stepObject, row, cursor, connectionPos, inputRows.rightEdge);
    } else if (row.type == Blockly.DUMMY_INPUT) {
      this.renderDummyInput_(
          stepObject, row, cursor, inputRows.rightEdge, inputRows.hasValue);
    } else if (row.type == Blockly.NEXT_STATEMENT) {
      this.renderStatementInput_(
          stepObject, row, cursor, connectionPos, inputRows, y);
    }
```

A dummy row (no actual input connection) on a block with inline inputs seems to still be rendered as an inline row, which means that it's missing out on the alignment code that's normally used for fields on dummy inputs.

The blocks look like this:
![image](https://user-images.githubusercontent.com/13686399/40442022-78ddd65e-5e77-11e8-970d-a724fb46e9fd.png)

The left has inline inputs; the right has external inputs.

MNSDivya:
![image](https://user-images.githubusercontent.com/56336353/81170523-781ad400-8fb8-11ea-89b2-a8217b173b0d.png)
Dummpy input seems changed to check star mark

rachel-fenichel:
Closing because I can't reproduce.

PR #3337, opened 2019-10-24
Fix dummy input alignment. Fix for #3330

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3330

### Proposed Changes

Wrap dummy inputs in a measurable so that it can be aligned appropriately by ``addAlignmentPadding_``.

### Reason for Changes

Fix rendering.

### Test Coverage

Tested with the following block in the block factory: 
```
{
  "type": "test",
  "message0": "s %1 some %2 s %3 ss %4",
  "args0": [
    {
      "type": "input_dummy",
      "align": "RIGHT"
    },
    {
      "type": "input_statement",
      "name": "NAME",
      "align": "RIGHT"
    },
    {
      "type": "input_statement",
      "name": "NAME",
      "align": "RIGHT"
    },
    {
      "type": "input_value",
      "name": "NAME",
      "align": "RIGHT"
    }
  ],
  "colour": 230,
  "tooltip": "",
  "helpUrl": ""
}
```

### Additional Information

@rachel-fenichel FYI.

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -374,8 +374,8 @@ Blockly.blockRendering.RenderInfo.prototype.addInput_ = function(input, activeRo
         new Blockly.blockRendering.ExternalValueInput(this.constants_, input));
     activeRow.hasExternalInput = true;
   } else if (input.type == Blockly.DUMMY_INPUT) {
-    // Dummy inputs have no visual representation, but the information is still
-    // important.
+    activeRow.elements.push(
+        new Blockly.blockRendering.DummyInput(this.constants_, input));
     activeRow.hasDummyInput = true;
   }
 };
--- a/core/renderers/geras/info.js
+++ b/core/renderers/geras/info.js
@@ -113,8 +113,8 @@ Blockly.geras.RenderInfo.prototype.addInput_ = function(input, activeRow) {
         new Blockly.blockRendering.ExternalValueInput(this.constants_, input));
     activeRow.hasExternalInput = true;
   } else if (input.type == Blockly.DUMMY_INPUT) {
-    // Dummy inputs have no visual representation, but the information is still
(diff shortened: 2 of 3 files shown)


core/renderers/common/info.js
```diff
@@ -374,8 +374,8 @@ Blockly.blockRendering.RenderInfo.prototype.addInput_ = function(input, activeRo
         new Blockly.blockRendering.ExternalValueInput(this.constants_, input));
     activeRow.hasExternalInput = true;
   } else if (input.type == Blockly.DUMMY_INPUT) {
-    // Dummy inputs have no visual representation, but the information is still
-    // important.
+    activeRow.elements.push(
+        new Blockly.blockRendering.DummyInput(this.constants_, input));
     activeRow.hasDummyInput = true;
```
RoboErikG: Do we still need the boolean with this change?

samelhusseini: yeah it's still used for deriving inter-row padding.

rachel-fenichel:
LG.  May also fix https://github.com/google/blockly/issues/338

It looks like I noticed this in https://github.com/google/blockly/pull/2947 and thought that it matched existing behaviour.  Whoops.

samelhusseini:
Should we cherry pick?

samelhusseini:
There's actually more to this. I hadn't realized this but this adds extra padding because it treats the dummy input as an input.
Screen Shot 2019-10-24 at 9 51 36 PM

Reverting to re-work it.

RoboErikG:
I suspect you're getting extra padding because having the dummy input adds some padding by itself, and then the check for hasDummyInput in info.js getSpacerRowHeight_ adds some more (in both geras and thrasos)

Issue #3341, opened 2019-10-25
Value input before/after dummy input row height inconsistencies


samelhusseini:
Value input after dummy input.

Original rendering: 
<img width="180" alt="Screen Shot 2019-10-25 at 1 13 25 PM" src="https://user-images.githubusercontent.com/16690124/67602215-015c6e80-f72b-11e9-91e3-2f171a81451e.png">

Geras rendering:
<img width="148" alt="Screen Shot 2019-10-25 at 1 26 42 PM" src="https://user-images.githubusercontent.com/16690124/67602270-1f29d380-f72b-11e9-8b2b-679436a1aaec.png">

Value input before dummy input.

Original rendering:
<img width="162" alt="Screen Shot 2019-10-25 at 1 27 47 PM" src="https://user-images.githubusercontent.com/16690124/67602378-4aacbe00-f72b-11e9-85f5-8c0c235634ee.png">

Geras rendering:
<img width="151" alt="Screen Shot 2019-10-25 at 1 27 55 PM" src="https://user-images.githubusercontent.com/16690124/67602389-4ed8db80-f72b-11e9-8daf-cb214372b2c9.png">

RoboErikG:
Somewhat related to #3330

samelhusseini:
Fixed.

PR #3342, opened 2019-10-25
Dummy input right alignment

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3341
Fixes https://github.com/google/blockly/issues/3330

### Proposed Changes

Adds a measurable for a dummy input and updates inRowSpacing to deal with the new measurable. 

Adds a special case to deal with blocks that have any external inputs, to add a spacer to the right that is the tab width in order to align items like this: (all text on the right aligns)

Screen Shot 2019-10-25 at 3 00 28 PM

Add a constant for a minimum height for dummy inputs.

### Reason for Changes

Rendering.

### Test Coverage

Tested by comparing rendering with results in the block factory here: https://20190419-dot-blockly-demo.appspot.com/static/demos/blockfactory/index.html

Tested this block: 
Screen Shot 2019-10-25 at 3 00 31 PM

Tested on:

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -24,7 +24,6 @@
 goog.provide('Blockly.blockRendering.RenderInfo');
 
 goog.require('Blockly.blockRendering.BottomRow');
-goog.require('Blockly.blockRendering.DummyInput');
 goog.require('Blockly.blockRendering.ExternalValueInput');
 goog.require('Blockly.blockRendering.Hat');
 goog.require('Blockly.blockRendering.InlineInput');
@@ -375,10 +374,11 @@ Blockly.blockRendering.RenderInfo.prototype.addInput_ = function(input, activeRo
         new Blockly.blockRendering.ExternalValueInput(this.constants_, input));
     activeRow.hasExternalInput = true;
   } else if (input.type == Blockly.DUMMY_INPUT) {
-    activeRow.elements.push(
-        new Blockly.blockRendering.DummyInput(this.constants_, input));
+    // Dummy inputs have no visual representation, but the information is still
+    // important.
     activeRow.hasDummyInput = true;
   }
+  activeRow.align = input.align;
 };
 
 /**
@@ -411,12 +411,6 @@ Blockly.blockRendering.RenderInfo.prototype.shouldStartNewRow_ = function(input,
  * @protected
  */
 Blockly.blockRendering.RenderInfo.prototype.addElemSpacing_ = function() {
-  var hasExternalInputs = false;
-  for (var i = 0, row; (row = this.rows[i]); i++) {
-    if (row.hasExternalInput) {
-      hasExternalInputs = true;
-    }
-  }
   for (var i = 0, row; (row = this.rows[i]); i++) {
     var oldElems = row.elements;
     row.elements = [];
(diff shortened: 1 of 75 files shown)


core/renderers/geras/info.js
```diff
@@ -139,8 +139,9 @@ Blockly.geras.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
     return this.constants_.LARGE_PADDING;
   }
 
-  // Spacing between a non-input and the end of the row.
-  if (!Blockly.blockRendering.Types.isInput(prev) && !next) {
+  // Spacing between a non-input and the end of the row or a dummy input.
+  if (!Blockly.blockRendering.Types.isInput(prev) && (!next ||
+      (next && Blockly.blockRendering.Types.isDummyInput(next)))) {
```
RoboErikG: nit, can remove the next before && (!next || (next && ...)) -> (!next || ...)

core/renderers/measurables/types.js
```diff
@@ -185,6 +186,17 @@ Blockly.blockRendering.Types.isInlineInput = function(elem) {
   return elem.type & Blockly.blockRendering.Types.INLINE_INPUT;
 };
 
+/**
+ * Whether a measurable stores information about a dummy input.
+ * @param {!Blockly.blockRendering.Measurable} elem The element to check.
+ * @return {number} 1 if the object stores information about an
```
RoboErikG: nit 'an' -> 'a'

samelhusseini:
@rachel-fenichel I've been pondering this one for a while, if you could please take a look and ask all the questions! Hopefully I'll be able to defend my decisions, especially around this one: https://github.com/google/blockly/pull/3342/files#diff-a8035c3fa36575e91cde165d25330a9eR432

core/renderers/common/info.js
```diff
@@ -472,6 +487,11 @@ Blockly.blockRendering.RenderInfo.prototype.getInRowSpacing_ = function(prev, ne
       return next.notchOffset - this.constants_.CORNER_RADIUS;
     }
   }
+  
+  // Spacing between a dummy input and anything.
+  if (prev && Blockly.blockRendering.Types.isDummyInput(prev)) {
```
rachel-fenichel: This will still add spacing after a dummy input and before the next thing, which will come up in an inline render of the blocks you tested.

core/renderers/common/info.js
```diff
@@ -410,6 +411,12 @@ Blockly.blockRendering.RenderInfo.prototype.shouldStartNewRow_ = function(input,
  * @protected
  */
 Blockly.blockRendering.RenderInfo.prototype.addElemSpacing_ = function() {
+  var hasExternalInputs = false;
```
rachel-fenichel: I don't think this belongs in the base renderer.  I could see it in Thrasos.

core/renderers/geras/info.js
```diff
@@ -113,8 +113,8 @@ Blockly.geras.RenderInfo.prototype.addInput_ = function(input, activeRow) {
         new Blockly.blockRendering.ExternalValueInput(this.constants_, input));
     activeRow.hasExternalInput = true;
   } else if (input.type == Blockly.DUMMY_INPUT) {
-    // Dummy inputs have no visual representation, but the information is still
-    // important.
+    activeRow.elements.push(
```
rachel-fenichel: High-level question: why add an element for the dummy input at all, given that it's zero width and not supposed to have padding around it, instead of just adding a .align property on the row and setting it based on the input (s) on that row?

samelhusseini: Although dummy inputs don't have a width, they do have a MIN_HEIGHT that should be taken into consideration (partially what this is tackling see https://github.com/google/blockly/issues/3341).
That being said, we could set a MIN_HEIGHT on the row if we see a dummy input.

samelhusseini:
@rachel-fenichel updated with your feedback. This is ready for another look.

PR #3352, opened 2019-10-28
Add keyboard navigation support for multiple workspaces

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Adds support for navigating to multiple workspaces using the keyboard.

### Proposed Changes
1. Add keyboardAccessibilityMode onto the workspace
2. Add a tabindex and label onto the injectionDiv
3. onFocus set the mainworkspace
4. onBlur set the mainWorkspace to null

### Reason for Changes
Want to be able to test with screen readers so we need to add a tabindex, and also multi workspaces should work with keyboard navigation.

### Test Coverage

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -325,8 +325,7 @@ Blockly.Block.prototype.dispose = function(healStack) {
     this.workspace.removeChangeListener(this.onchangeWrapper_);
   }
 
-  if (Blockly.getMainWorkspace() &&
-      Blockly.getMainWorkspace().keyboardAccessibilityMode) {
+  if (this.workspace.keyboardAccessibilityMode) {
     // No-op if this is called from the block_svg class.
     Blockly.navigation.moveCursorOnBlockDelete(this);
   }
--- a/core/block.js
+++ b/core/block.js
@@ -325,7 +325,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
     this.workspace.removeChangeListener(this.onchangeWrapper_);
   }
 
-  if (Blockly.keyboardAccessibilityMode) {
+  if (Blockly.getMainWorkspace() &&
+      Blockly.getMainWorkspace().keyboardAccessibilityMode) {
     // No-op if this is called from the block_svg class.
     Blockly.navigation.moveCursorOnBlockDelete(this);
   }
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1031,7 +1031,7 @@ Blockly.BlockSvg.prototype.dispose = function(healStack, animate) {
     Blockly.ContextMenu.hide();
   }
 
-  if (Blockly.keyboardAccessibilityMode) {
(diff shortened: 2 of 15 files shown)


core/block.js
```diff
@@ -325,7 +325,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
     this.workspace.removeChangeListener(this.onchangeWrapper_);
   }
 
-  if (Blockly.keyboardAccessibilityMode) {
+  if (Blockly.getMainWorkspace() &&
```
samelhusseini: The block has a workspace associated with it. why not use that?
This would affect blocks across workspaces if the mainWorkspace changes with a focus on the other workspace.

core/block_svg.js
```diff
@@ -1031,7 +1031,7 @@ Blockly.BlockSvg.prototype.dispose = function(healStack, animate) {
     Blockly.ContextMenu.hide();
   }
 
-  if (Blockly.keyboardAccessibilityMode) {
+  if (Blockly.getMainWorkspace().keyboardAccessibilityMode) {
```
samelhusseini: Similar to above, why not use this.workspace?

core/inject.js
```diff
@@ -78,6 +82,14 @@ Blockly.inject = function(container, opt_options) {
 
   Blockly.svgResize(workspace);
 
+  subContainer.onfocus = function() {
```
samelhusseini: Use addEventListener instead (more generally, so others are able to add other listeners if they want).

core/mutator.js
```diff
@@ -374,7 +374,8 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
       block.render();
     }
 
-    if (oldMutation != newMutation && Blockly.keyboardAccessibilityMode) {
+    if (oldMutation != newMutation &&
+        Blockly.getMainWorkspace().keyboardAccessibilityMode) {
```
samelhusseini: Use the workspace reference here.

core/block_svg.js
alschmiedt: fixed

core/block.js
alschmiedt: fixed

core/inject.js
```diff
@@ -61,6 +61,10 @@ Blockly.inject = function(container, opt_options) {
     (/** @type {!Blockly.BlocklyOptions} */ ({})));
   var subContainer = document.createElement('div');
   subContainer.className = 'injectionDiv';
+  subContainer.tabIndex = 0;
+  Blockly.utils.aria.setState(subContainer,
+      Blockly.utils.aria.State.LABEL, Blockly.Msg.WORKSPACE_ARIA_LABEL);
```
samelhusseini: Blockly.Msg is of type object, so for advanced compilation, you'll need to use string lookup instead ``Blockly.Msg['WORKSPACE_ARIA_LABEL']``

Issue #3233, opened 2019-10-14
More control over how themes are applied to path_objects


BeksOmega:
### Is your feature request related to a problem? Please describe.

For the Blockly Summit 2019 Hackathon I wanted to create a block that had different border widths on the top-left than on the bottom-right.

To do this I created a custom path_object (which held the top-left border, the inside, and the bottom-left border), but then themes did not know how to apply colours to my new path_object.

This is because updating themes is handled by the blockSvg, which expects specific properties on the path_object.

#### Allow the Path_Object to handle colour updates
The path_object interface gains the following functions:
* setStyle(style)
* setShadow(style)
* setDisabled(style)

The path_object can then apply colours (and other css properties for that matter) to its svg elements in any way it desires.

I also think the following properties should be added to the style interface. They currently exist on the block (I believe) but those could easily be made into facade-functions that just pass the responsibility to the style.
* getColourPrimary
* getColourSecondary
* getColourTertiary
* getColourShadow
* getColourBorder
* getColourMain <- Currently exists as getColour

Giving the responsibility to the style would keep the path_object from having knowledge of the block.

#### Set colours automatically based on css class names
Path_objects would be responsible for applying class names to the different elements, which would tell the themes how to apply colours to the different svg elements.

E.g. The path_object would apply a `blocklyBorderColourStroke` class to every svg element that you want to have a stroke value of the border colour.

I don't like this solution as much because you'd end up needing a class for everything a developer would want to do.

For example: What if a developer wanted shadow blocks to have borders?
* `blocklyShadowBorderColourStroke`

Then what if they wanted shadow blocks to have the primary colour as the border instead of the normal border colour?
* `blocklyShadowPrimaryColourStroke`

And as things get added to styles (e.g. stroke width, stroke dashes, opacity, etc) It would just get more and more complicated to maintain these.

### Additional context

N/A

rachel-fenichel:
@BeksOmega we've made a bunch of changes to how path objects work and interact with styles, including that they go through constants, and that constants get refreshed when themes change.  Can you take a look and comment on those APIs?

BeksOmega:
Confirmed! It is now possible to apply colours to a pathobject in a render-specific fasion. Thanks for the fix!

PR #3357, opened 2019-10-28
Move block colouring code from block to the pathObject

rachel-fenichel:
##  The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/3233

### Proposed Changes

Move block colouring code from block/block_svg to path_object.

### Reason for Changes

Figuring out primary, secondary, and tertiary from the style or single block colour is now handled in a central location.
Applying that to the paths in a way that's aware of which paths exist is now handled by the path object.

### Test Coverage
Tested in the playground.  You can now load any theme with any renderer.  

### Additional Information

We may want to change the name of the path object.


core/block.js
```diff
@@ -890,63 +874,10 @@ Blockly.Block.prototype.setTooltip = function(newTip) {
  * @return {string} #RRGGBB string.
  */
 Blockly.Block.prototype.getColour = function() {
+  // TODO: Can we remove this?
```
rachel-fenichel: Open question: can I remove colour and hue from block entirely?

core/block_svg.js
```diff
@@ -1061,75 +1053,35 @@ Blockly.BlockSvg.prototype.dispose = function(healStack, animate) {
...
-  } else {
-    this.setBorderColour_();
-    this.svgPath_.setAttribute('fill', this.getColour());
-  }
+  this.pathObject.applyColour(this.isShadow());
 
   var icons = this.getIcons();
   for (var i = 0; i < icons.length; i++) {
-    icons[i].updateColour();
+    // TODO: Decide whether to make icon and field applyColour take in a
```
rachel-fenichel: I'm open to suggestions here.  I'll make an actual issue if it's a change that we want.

Also, note that I changed a name: `field.updateColour` -> `field.applyColour`. 

This was to make names match, and to indicate that the colour model isn't changing, it's that the change is being applied to the rendered block.  I could be talked into changing it back if we want to leave the name alone for field.

core/block.js
samelhusseini: It doesn't need to be stored internally on the block, but a Field could need to call into the block to ask it for it's colour. So I think we still need references to all three of these.

rachel-fenichel: I had those fields looking at the block's colourer--now they look at the path object.

core/block_svg.js
```diff
@@ -1374,15 +1326,40 @@ Blockly.BlockSvg.prototype.setDeleteStyle = function(enable) {
 // Overrides of functions on Blockly.Block that take into account whether the
 // block has been rendered.
 
+/**
+ * Get the colour of a block.
+ * @return {string} #RRGGBB string.
+ */
+Blockly.BlockSvg.prototype.getColour = function() {
+  return this.pathObject.primaryColour;
+};
```
samelhusseini: I would also add ``getColourSecondary`` and ``getColourTertiary`` wrapper methods.

core/block.js
samelhusseini: Yeah, I'd rather not surface that path object outside of the renderer. It makes it public domain if fields are accessing it.

core/utils/colour.js
```diff
@@ -211,3 +211,39 @@ Blockly.utils.colour.names = {
   'white': '#ffffff',
   'yellow': '#ffff00'
 };
+
+/**
+ * Parse a block colour from a number or string, as provided in a block
+ * definition.
+ * @param {number|string} colour HSV hue value (0 to 360), #RRGGBB string,
+ *     or a message reference string pointing to one of those two values.
+ * @return {{hue: ?number, colour: string}} An object containing the colour as
```
samelhusseini: update @return type to ``{{hue: ?number, hex: string}}`` to reflect update below.

core/renderers/geras/renderer.js
```diff
@@ -106,6 +106,7 @@ Blockly.geras.Renderer.prototype.makePathObject = function(root) {
   return new Blockly.geras.PathObject(root);
 };
 
+
```
samelhusseini: nit: remove newline

core/utils/colour.js
```diff
@@ -211,3 +211,39 @@ Blockly.utils.colour.names = {
...
+ *     a #RRGGBB string, and the hue if the input was an HSV hue value.
```
samelhusseini: Add @throws

core/renderers/geras/path_object.js
```diff
@@ -66,7 +67,40 @@ Blockly.geras.PathObject = function(root) {
...
+  this.tertiaryColour = '#000000';
+
+  /**
+   * The colour of the dark path on the block in '#RRGGBB' format.
+   * @type {string}
+   * @package
+   */
+  this.darkColour = '#000000';
 };
+Blockly.utils.object.inherits(Blockly.geras.PathObject,
```
samelhusseini: Add require for ``Blockly.utils.object``

core/renderers/geras/path_object.js
```diff
@@ -66,7 +67,40 @@ Blockly.geras.PathObject = function(root) {
...
   this.svgPathLight = Blockly.utils.dom.createSvgElement('path',
       {'class': 'blocklyPathLight'}, this.svgRoot);
+
+
+  /**
+   * Primary colour of the block in '#RRGGBB' format.
+   * @type {string}
+   * @package
+   */
+  this.primaryColour = '#000000';
```
samelhusseini: Why do you need to re-declare these here? If they're inherited.

core/renderers/common/path_object.js
```diff
@@ -95,3 +109,62 @@ Blockly.blockRendering.PathObject.prototype.flipRTL = function() {
...
+  }
+};
+
+/**
+ * Update colour properties based on a triplet of colours.
+ * @param {string} primary The primary colour.
+ * @param {string} secondary The secondary colour, or null to have the colourer
+ *     generate it.
+ * @param {string} tertiary The tertiary colour, or null to have the colourer
+ *     generate it.
```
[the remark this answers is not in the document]
samelhusseini: Yeah good catch. ``{?string}`` not because of subclasses but because you're checking below if secondary is null.

core/block.js
rachel-fenichel: Moving back to field getting colours from the block.

rachel-fenichel: And I've filed #3360 about removing colour from block and having it only on blockSvg.

core/block_svg.js
rachel-fenichel: Added back.

core/renderers/common/path_object.js
```diff
@@ -74,6 +67,27 @@ Blockly.blockRendering.PathObject = function(root) {
...
+   * @package
+   */
+  this.secondaryColour = '#000000';
+
+  /**
+   * Tertiary colour of the block in '#RRGGBB' format.
+   * @type {string}
+   * @package
+   */
+  this.tertiaryColour = '#000000';
```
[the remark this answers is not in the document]
rachel-fenichel: Yes, because the path object is passed the style to set these, so we could add more properties.

That said, it might be good to store the style as well.  I'll noodle on that.

core/renderers/common/path_object.js
```diff
@@ -95,3 +109,62 @@ Blockly.blockRendering.PathObject.prototype.flipRTL = function() {
...
   this.svgPath.setAttribute('transform', 'scale(-1 1)');
 };
+
+/**
+ * Apply the stored colours to the block's path, taking into account whether
+ * the paths belong to a shadow block.
+ * @param {boolean} isShadow True if the block is a shadow block.
+ * @package
+ */
+Blockly.blockRendering.PathObject.prototype.applyColour = function(isShadow) {
```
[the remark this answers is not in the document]
rachel-fenichel: You are one step ahead of my code! Yes, `isShadow` isn't the only use case, it's just the one that comes directly from `block.updateColour`.  After this PR I'm going to take a look at highlighting, disabling, and insertion markers, and I may split it out into more functions or add an enum.

core/renderers/geras/path_object.js
rachel-fenichel: I'm not calling the superclass constructor here, because the order of creating the svg path objects matters.

The old pattern I would use puts these declarations outside of the constructor--they would be on the prototype and left below.  We agreed to stop doing that, but it seems relevant here.

core/renderers/geras/path_object.js
rachel-fenichel: Done.

core/renderers/geras/renderer.js
rachel-fenichel: Done.

core/utils/colour.js
rachel-fenichel: Done.

core/utils/colour.js
rachel-fenichel: Done.

rachel-fenichel:
@BeksOmega I'll think about the suggestion of moving the inference of additional colours into the style constructor (for single-colour styles).  I may also keep a handle on the style object in the path object.

The way I envisioned this being used, you would extend it, making your own path object, to add the sorts of styling that you wanted.  Then you have unlimited access to your svg paths and you also have access to the style object, in setFromStyle.

(The name will also change from pathObject to something more accurate, but that's waiting until I figure out what other funcitonality moves to this object.)

BeksOmega:
> The way I envisioned this being used, you would extend it, making your own path object, to add the sorts of styling that you wanted. Then you have unlimited access to your svg paths and you also have access to the style object, in setFromStyle.

Yes this makes sense! You want to outside developers to be able to style their custom paths in whatever way they want. And to be able to do that they need access to all of their custom paths.

The one thing I'm a little bit fuzzy on is why the pathObject has to /store/ all of the style data. It seems like it could just retrieve it when it needs it.

My mental picture looks something like this:
StyleClassDiagram

BlockSvg: Acts as a facade for all of the objects that compose a block. Importantly in this case, the `Style` and `PathObject` objects.

Style: Holds all the style properties of the block. Handles returning default properties if they were not specified explicitly.

PathObject: Handles creating and styling the paths of the block. (Bonus! You don't have to rename the PathObject)

---

One major benefit of this is that there is a well defined point of access for getting the style properties of the block. `Field`s, `Icon`s, `PathObject`s, etc all call `block.getStyle`.

Calling `block.getColourPrimary` or `pathObject.getColourPrimary` gives those objects responsibility that could easily be handled by the `Style`.

This keeps everything organized and everything encapsulated.

rachel-fenichel:
Your design sounds good.  I'll update the code to make all blocks have styles.

rachel-fenichel:
@alschmiedt fyi

core/renderers/common/path_object.js
```diff
@@ -67,13 +61,11 @@ Blockly.blockRendering.PathObject = function(root) {
       {'class': 'blocklyPathLight'}, this.svgRoot);
 
   /**
-   * The dark path of the block.
-   * @type {SVGElement}
-   * @package
+   * The style object to use when colouring block paths.
+   * @type {!Blockly.Theme.BlockStyle}
+   * @public
```
samelhusseini: ``@public`` is implicit

core/block_svg.js
```diff
@@ -1332,13 +1328,16 @@ Blockly.BlockSvg.prototype.setColour = function(colour) {
...
     this.hat = blockStyle.hat;
-    this.pathObject.setColourFromStyle(blockStyle);
+    this.pathObject.setStyle(blockStyle);
+    // Set colour to match Block.
+    this.colour_ = blockStyle.colourPrimary;
+    this.style = blockStyle;
+
     this.applyColour();
   } else {
     throw Error('Invalid style name: ' + blockStyleName);
```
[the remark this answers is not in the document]
rachel-fenichel: I think so, because later changes to the styles may mean that a valid style for that name starts to exist.

core/renderers/common/path_object.js
```diff
@@ -60,34 +61,11 @@ Blockly.blockRendering.PathObject = function(root) {
...
-      {'class': 'blocklyPathDark', 'transform': 'translate(1,1)'},
-      this.svgRoot);
-
-  /**
-   * Primary colour of the block in '#RRGGBB' format.
-   * @type {string}
-   * @package
+   * The style object to use when colouring block paths.
+   * @type {!Blockly.Theme.BlockStyle}
+   * @public
```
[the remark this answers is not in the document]
rachel-fenichel: Actually, I'll go with package until I need it to be public, since it's also stored on the block.

core/renderers/geras/path_object.js
```diff
@@ -69,39 +68,20 @@ Blockly.geras.PathObject = function(root) {
...
-
   /**
-   * Tertiary colour of the block in '#RRGGBB' format.
-   * Used for the light path (highlight) in Geras.
+   * The colour of the dark path on the block in '#RRGGBB' format.
    * @type {string}
    * @package
    */
-  this.tertiaryColour = '#000000';
+  this.darkColour = '#000000';
```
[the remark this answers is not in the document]
rachel-fenichel: Good catch.

core/renderers/geras/path_object.js
```diff
@@ -135,26 +115,26 @@ Blockly.geras.PathObject.prototype.flipRTL = function() {
...
+    this.svgPathDark.setAttribute('fill', this.colourDark);
+    this.svgPath.setAttribute('fill', this.style.colourPrimary);
   }
 };
 
 /**
- * @override
+ * Set the style.
+ * @param {!Blockly.Theme.BlockStyle} blockStyle The block style to use.
+ * @package
```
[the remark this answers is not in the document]
rachel-fenichel: No, because I'm no longer extending `Blockly.blockRendering.PathObject`.

core/theme.js
```diff
@@ -97,22 +101,90 @@ Blockly.Theme.prototype.getAllBlockStyles = function() {
...
  * Gets the BlockStyle for the given block style name.
- * @param {string} blockStyleName The name of the block style.
- * @return {Blockly.Theme.BlockStyle|undefined} The named block style.
+ * @param {?string} blockStyleName The name of the block style.
+ * @return {!Blockly.Theme.BlockStyle} The named block style, or a default style
+ *     if no style with the given name was found.
  */
 Blockly.Theme.prototype.getBlockStyle = function(blockStyleName) {
-  return this.blockStyles_[blockStyleName];
+  var defaultStyle = Blockly.Theme.createBlockStyle('#000000');
```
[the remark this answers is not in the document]
rachel-fenichel: Yeah, that's another piece of upcoming work.  I need to decide how default styles work, and whether that's on a theme, on Blockly.Theme, or on the renderer.

core/theme.js
```diff
@@ -97,22 +101,90 @@ Blockly.Theme.prototype.getAllBlockStyles = function() {
...
 };
 
 /**
  * Overrides or adds a style to the blockStyles map.
  * @param {string} blockStyleName The name of the block style.
  * @param {Blockly.Theme.BlockStyle} blockStyle The block style.
-*/
+ * @package
+ */
 Blockly.Theme.prototype.setBlockStyle = function(blockStyleName, blockStyle) {
```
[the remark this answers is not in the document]
rachel-fenichel: Nope.  Good catch.

PR #3359, opened 2019-10-29
Accept lowercase alignment

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

We weren't accepting lowercase alignment values in block config.

### Proposed Changes

Accept lowercase version of LEFT, RIGHT, and CENTRE.
Accept alternate spelling of CENTER.
The first input decides the alignment.
Add input type information on the row.

### Reason for Changes

More flexibility in what we author.

### Test Coverage

This block now looks like it was authored (with a right aligned value input)
Screen Shot 2019-10-28 at 5 13 02 PM

Tested on:
* Desktop Chrome

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1678,7 +1678,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign,
         if (element['align']) {
           var align = !!element['align'] && element['align'].toUpperCase();
           var alignment = alignmentLookup[align];
-          if (alignment != null) {
+          if (alignment != undefined) {
             input.setAlign(alignment);
           } else {
             console.warn(warningPrefix + 'Illegal align value: ',
--- a/core/block.js
+++ b/core/block.js
@@ -1454,7 +1454,7 @@ Blockly.Block.prototype.jsonInit = function(json) {
   var i = 0;
   while (json['message' + i] !== undefined) {
     this.interpolate_(json['message' + i], json['args' + i] || [],
-        json['lastDummyAlign' + i]);
+        json['lastDummyAlign' + i], warningPrefix);
     i++;
   }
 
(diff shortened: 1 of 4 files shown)


core/block.js
```diff
@@ -1625,6 +1625,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
   var alignmentLookup = {
     'LEFT': Blockly.ALIGN_LEFT,
     'RIGHT': Blockly.ALIGN_RIGHT,
+    'CENTRE': Blockly.ALIGN_CENTRE,
```
RoboErikG: I think you mean 'CENTER'

core/block.js
```diff
@@ -1673,7 +1674,8 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
           input.setCheck(element['check']);
         }
         if (element['align']) {
-          input.setAlign(alignmentLookup[element['align']]);
+          var alignment = !!element['align'] && element['align'].toUpperCase();
+          input.setAlign(alignmentLookup[alignment]);
```
RoboErikG: If that alignment doesn't exist this will set the input's alignment to undefined. Should this throw if the lookup doesn't exist?

core/block.js
samelhusseini: Lol yes

core/block.js
samelhusseini: I propose we default to Blockly.ALIGN_LEFT in that case and warn.

RoboErikG: SGTM.

core/renderers/measurables/rows.js
```diff
@@ -154,6 +154,13 @@ Blockly.blockRendering.Row = function(constants) {
   this.constants_ = constants;
 
   this.notchOffset = this.constants_.NOTCH_OFFSET_LEFT;
+
+  /**
+   * Alignment of the row.
+   * @package
+   * @type {number}
```
moniika: Since this is null-able, should add ?

core/block.js
```diff
@@ -1673,7 +1676,14 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
           input.setCheck(element['check']);
         }
         if (element['align']) {
-          input.setAlign(alignmentLookup[element['align']]);
+          var align = !!element['align'] && element['align'].toUpperCase();
+          var alignment = alignmentLookup[align];
+          if (alignment != null) {
```
RoboErikG: This will be undefined if it doesn't exist in the lookup, not null. Should be if (alignment)

samelhusseini: undefined == null, so this works. However if your suggestion is ``alignment != undefined`` is more appropriate here, then I think you're right. I initially avoided it because I know undefined can be re-assigned in old browsers, so the more appropriate check here would be ``typeof alignment !== "undefined"``, but that's a few more characters than ``undefined == null``. Looking into this further, it turns out ES5 is when the issue of re-assignment of undefined was fixed, and since we support only ES5+ browsers, then going with ``alignment != undefined`` sounds good.

PR #3362, opened 2019-10-30
Fix warnings for code used by blocks

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Compiler warnings.

### Proposed Changes

Fix warnings for code used by blocks.

### Test Coverage

Tested on:

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -239,13 +239,6 @@ Blockly.BlockSvg.prototype.decompose;
 Blockly.BlockSvg.prototype.compose;
 
 /**
- * An optional boolean for determining whether or not the context menu is
- * shown on a block.
- * @type {?boolean}
- */
-Blockly.BlockSvg.prototype.contextMenu;
-
-/**
  * An optional method for defining custom block context menu items.
  * @type {?function(!Array.<!Object>)}
  */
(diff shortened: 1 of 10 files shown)


core/block_svg.js
```diff
@@ -238,6 +238,19 @@ Blockly.BlockSvg.prototype.decompose;
  */
 Blockly.BlockSvg.prototype.compose;
 
+/**
+ * An optional boolean for determining whether or not the context menu is
+ * shown on a block.
+ * @type {?boolean}
+ */
+Blockly.BlockSvg.prototype.contextMenu;
```
alschmiedt: Just curious why we need this here if it is already defined on the block?

samelhusseini: Good point.

core/renderers/measurables/connections.js
```diff
@@ -47,7 +47,7 @@ Blockly.blockRendering.Connection = function(constants, connectionModel) {
       constants);
   this.connectionModel = connectionModel;
   this.shape = this.constants_.shapeFor(connectionModel);
-  this.isDynamicShape = !!this.shape.isDynamic;
+  this.isDynamicShape = !!this.shape['isDynamic'];
```
rachel-fenichel: Why the change in style?

samelhusseini: shapeFor is returning an object, we don't have any idea of its "shape" (no pun intended). 
When we use an interface for shapes this will go away.

PR #3383, opened 2019-11-01
Make warnings an optional module.

NeilFraser:
Issue #3023


core/block_svg.js
```diff
@@ -1107,7 +1106,7 @@ Blockly.BlockSvg.prototype.setCommentText = function(text) {
   }
   if (shouldHaveComment) {
     this.commentIcon_ = new Blockly.Comment(this);
```
samelhusseini: Unrelated to this, but this should have a check if ``Blockly.Comment`` exists.

NeilFraser: It does, top of that function.

samelhusseini: Ah yes I see it.

Issue #3382, opened 2019-11-01
Advanced compilation test is now black and white.


NeilFraser:
Similar in effect to #2215, but not the same cause.
<img width="175" alt="Screen Shot 2019-10-31 at 22 23 15" src="https://user-images.githubusercontent.com/250480/68004011-0fc7f000-fc2d-11e9-9e1c-1273b2fcb3d6.png">

samelhusseini:
Fixed.

PR #3384, opened 2019-11-01
Fix block colours in advanced compilation

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3382

### Proposed Changes

Access the theme object properties with bracket notation so advanced compilation doesn't rename them.

@rachel-fenichel FYI

### Reason for Changes

Fixes colours in advanced compilation.

### Test Coverage

Tested on:

### DIFF
--- a/core/theme.js
+++ b/core/theme.js
@@ -170,18 +170,18 @@ Blockly.Theme.validatedBlockStyle = function(blockStyle) {
 
   // Validate required properties.
   var parsedColour = Blockly.utils.colour.parseBlockColour(
-      valid.colourPrimary || '#000');
+      valid['colourPrimary'] || '#000');
   valid.colourPrimary = parsedColour.hex;
-  valid.colourSecondary = valid.colourSecondary ?
-      Blockly.utils.colour.parseBlockColour(valid.colourSecondary).hex :
+  valid.colourSecondary = valid['colourSecondary'] ?
+      Blockly.utils.colour.parseBlockColour(valid['colourSecondary']).hex :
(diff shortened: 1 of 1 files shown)


BeksOmega:
This is a me question: Aren't renderers and themes included in the advanced compilation? So why don't the style properties get renamed in their files as well?

samelhusseini:
@BeksOmega. Advanced compilation would go around renaming everything that is the same to shorter characters but it doesn't rename strings. This is because if we're accepting users to give us input in the form of ``{readonly: boolean, comments: boolean, toolbox: string}`` then renaming it internally to be looking for ``{r, c, t}`` wouldn't work. Hence why if you look at ``options.js`` all the inputs it looks for use bracket notation.
In this case, it's okay to be using a different name internally to the external facing name for ``colourPrimary``, secondary, etc. Since we pass themes through the validation method, we're renaming those three values. Now there's a couple of interesting points here: 
- What if the renderer is accessing style values directly that are not the four that we just validated? The answer there it would need to use bracket notation to access those as well.
- Another thing to point out, is that validation method is going through and setting an internal "renamed" property on the style option but not removing the external one, so in fact in advanced compilation you'd end up with the style object looking more like this:
```
{
'colourPrimary': '#HEX',
'colourSecondary': '#HEX',
'colourTertiary': '#HEX',
'r': '#HEX',
'x': '#HEX',
'f': '#HEX'
} 
```
That's not a terrible side effect IMO. Internally we'd only refer to the above ``r``, ``x``, or ``f`` but the others still exist. And in the case where there are more properties on the style object, like say ``colourMain`` then that would continue to exist on the style object all throughout the code.

PR #3385, opened 2019-11-01
Add changes for focusing on the workspace

### LINKED ISSUE #3364: Clicking workspace after comment resize causes a crash
**Describe the bug**

**To Reproduce**

1. Go to the playground
2. Drag any block onto the workspace and add a comment.
3. Resize the comment
4. Click on the workspace
5. See error in console.

**Expected behavior**

No error.

**Stack Traces**

```
blockly.js:343 Uncaught TypeError: Cannot read property 'trashcan' of null
    at Object.Blockly.hideChaff (blockly.js:343)
    at Blockly.TouchGesture.Blockly.Gesture.doStart (gesture.js:491)
    at Blockly.TouchGesture.doStart (touch_gesture.js:119)
    at Blockly.TouchGesture.Blockly.Gesture.handleWsStart (gesture.js:659)
    at Blockly.WorkspaceSvg.onMouseDown_ (workspace_svg.js:1392)
    at SVGGElement.wrapFunc (blockly.js:479)
```
**Additional context**

This is because workspace.onBlur is clearning Blockly.mainworkspace and a hideChaff is getting called when we click on the workspace.

The blur is getting called not because the injection div is blurred but one of its children is blurred (the comment text area).

@alschmiedt FYI

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3364.

### Proposed Changes
1. use focusin and focusout
2. Add tabindex to blocklySvg.
3. Remove where the mainWorkspace is set automatically.

### Reason for Changes
1. Focusin and focusout bubble where as focus and blur do not.

2.  If we don't add the tabIndex to one of the children then the focus event does not get dispatched. Adding this means we have tabindex on the injectionDiv, blocklySvg, and the toolbox. 

3. This makes me a bit nervous, but it doesn't makes sense to have the workspace get focus without ever clicking it. Also, for multi workspaces the mainworkspace was just getting set to the last workspace on the page which could be confusing for keyboard navigation.

### Test Coverage

Tested on:


samelhusseini:
Couple of extra thoughts: 
- comments create a text area which by default has a tabindex of 0, if fields have to be navigated to via Blockly keyboard nav, it seems weird that comments would be an exception. 
- Should we add aria-labels for each of these tabbable elements: 
 - blocklyTreeRoot
 - blocklySvg
 - injectionDiv
- When we hit ``t`` in keyboard nav mode, should we focus the blocklyTreeRoot div, similar to how we focus the WidgetDiv when field's ``showEditor`` is called.
- Now that we're clearing Blockly.mainWorkspace when is out. It means there's no way for a user to get back to the workspace without tabbing to the element, after they're done editing a field. ie: keyboard nav got field -> showEditor -> WidgetDiv -> ENTER -> ??? Lost focus. That scenario was supported because Blockly.mainWorkspace was always set. I'm okay with this change, we just need to come up with a way for fields to return focus to the workspace.

samelhusseini:
Following on to point 4. The other issue here is ``showEditor`` now clears Blockly.mainWorkspace as focus is being moved out of the workspace and into Widgetdiv, so navigation won't work there. As in ``Blockly.onAction`` won't work in fields. One proposal there is to move the WidgetDiv under the injectionDiv and thus a field would stay in the workspace focus scope.

alschmiedt:
1. Yeah I agree but will probably look at more when we get to keyboard nav opening and closing comments. 
2. Yup! good call will add those. 
3. Yes, I was planning on looking at where and when keyboard nav needed to deal with focus in a different PR.
4. Ok I will investigate a bit more!

alschmiedt:
Separated out into #3428 and #3425

PR #3388, opened 2019-11-01
Move filters and patterns into constants

### LINKED ISSUE #3377: Move filter and pattern creation into renderer
Emboss filters (for highlighting) and patterns (for disabled blocks and grids) are currently created in `Blockly.createDom_` and saved on the options struct, but they should be in the renderer instead.

These filters and patterns are per-workspace, with unique IDs.

They are children of the `<defs>` element on the root Blockly SVG.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

#3377 

Moves more code into path object.

### Proposed Changes

- Move filter and pattern creation into the constants object
- Move `setHighlighted` and `setDisabled` into the path object
- Remove `svgPath_` and `svgPathLight_` from blockSvg
- Remove `embossFilterId` and `disabledPatternId` from the options struct
- Add `createDom` and `dispose` to the constants object
- Give the path object a reference to the constants object

### Reason for Changes

In general, this is moving block-rendering-related code into the rendering files.

Highlighted and disabled are both functions that act directly on the paths, and thus depend on whether the light path exists.

The filters and patterns were being created in inject, which didn't make that much sense.  The only benefit that I can see is the child workspaces (such as mutators) don't need to create their own patterns, and this is not much of a benefit.

The IDs seemed to be on the options struct just as a way of passing information around, and not because they really belonged there.  I checked and we can have multiple <defs> tags, and they don't have to be at the root level.

### Test Coverage
In the playground in Chrome I loaded blocks and toggled disable and highlight status.  I also opened up a mutator and did the same to make sure it worked, and I did this with both Geras and the minimalist renderer.


core/renderers/common/constants.js
```diff
@@ -398,3 +441,91 @@ Blockly.blockRendering.ConstantProvider.prototype.shapeFor = function(
...
+ * @param {!SVGElement} svg The root of the workspace's SVG.
+ * @package
+ */
+Blockly.blockRendering.ConstantProvider.prototype.createDom = function(svg) {
+  /*
+  <defs>
+    ... filters go here ...
+  </defs>
+  */
+  var defs = Blockly.utils.dom.createSvgElement('defs', {}, svg);
```
samelhusseini: This makes it hard for subclasses to extends this and add additional items to the same ``<defs>`` tag. I don't know if that's important since SVG supports multiple ``<defs>`` tags.

core/renderers/geras/path_object.js
```diff
@@ -34,11 +36,19 @@ goog.require('Blockly.utils.object');
  * An object that handles creating and setting each of the SVG elements
  * used by the renderer.
  * @param {!SVGElement} root The root SVG element.
+ * @param {!Blockly.blockRendering.ConstantProvider} constants The renderer's
```
samelhusseini: This should be ``Blockly.geras.ConstantProvider`` no?

core/renderers/geras/path_object.js
```diff
@@ -24,7 +24,9 @@
 
 goog.provide('Blockly.geras.PathObject');
 
+goog.require('Blockly.blockRendering.ConstantProvider');
```
samelhusseini: We should use ``Blockly.geras.ConstantProvider`` only here (that in turn would require ``Blockly.blockRendering.ConstantProvider``). so there's no need for this.

core/renderers/common/constants.js
rachel-fenichel: Not an issue.

core/renderers/geras/path_object.js
rachel-fenichel: No, because it's a `blockRendering.ConstantProvider` on the interface.  But I will eventually make an interface for the constants provider as well, and then can change this.

PR #3390, opened 2019-11-02
Move block selection into the renderer.

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Move block selection into the renderer and extend zelos rendering to use a block selection filter instead of a stroke.

@rachel-fenichel the motivation here was to use the ``<defs>`` extending and implement the selection filter in zelos.
Doing so I realized the reason we were avoiding PathObject inheritance was because of setPaths parameters, and then I came to the realization that we can have a ``setPath`` method that is on the interface that only takes a single path param, and for geras we can have a separate method ``setHighlightPath`` for the highlight path! Let me know if you had other concerns, but doing so avoids a lot of code duplication. 

Also going through this exercise I realized that constants need to access the workspace theme. At first thought, I figured we can pipe it through when the renderer is initialized (which is when constants is initialized), but that wouldn't work if themes can dynamically change throughout the workspace lifetime. It all gets kinda complicated when we're initializing a ``<defs>`` element that depends on the theme.. What component will be responsible for updating it? Perhaps at that point we need to resort to something like ``rendererOptions`` and make those static options that can't be dynamically updated like the theme.

### Reason for Changes

Extending selection highlight in zelos.

### Test Coverage

Tested geras and zelos in the playground.

Tested on:
* Desktop Chrome
* Desktop Firefox
* Desktop Safari


BeksOmega:
> It all gets kinda complicated when we're initializing a `<defs>` element that depends on the theme

If you want the `<def>`'s to be changable with the theme, could they live on the theme?

samelhusseini:
> If you want the `<def>`'s to be changable with the theme, could they live on the theme?
I think a more extensible design would be to use the theme manager. It already calls back into the workspace when the theme has changed. Now the workspace can call into its renderer / constants and tell it that the theme has changed in order to update any theme dependent constants / components.

BeksOmega:
> I think a more extensible design would be to use the theme manager. It already calls back into the workspace when the theme has changed. Now the workspace can call into its renderer / constants and tell it that the theme has changed in order to update any theme dependent constants / components.

So it seems like you're suggesting a sequence that looks something like [this ](https://app.creately.com/diagram/NRw9PbuIb6B/view)(if we're building off the current architecture)
ThemeSequenceA

I'm a little confused about how it would be more extensible, because it seems like there's a bottle neck where the constants go to create the defs.

The constants will need some way to figure out /what/ they're supposed to be creating, which I imagine will require a theme ID and a switch statement. That doesn't seem very extensible to me, because it will need to be modified every time you add a theme.

I think that the optimal sequence would look something like [this](https://app.creately.com/diagram/5QCHywIzZM6/view):
ThemeSequenceB
* The theme handles the creation of the def elements, keeping the system extensible by avoiding switch statements.
* Default defs can easily be provided through theme validation.
* All the theme info is contained in the theme, so fields and icons can access the defs if needed.
* The state parameter passed to the `updateStyle` property allows the theme to be changed while a block is highlighted or disabled.

samelhusseini:
The theme object is really meant to be a struct of colours, so that you can do something like [this](https://github.com/google/blockly/blob/develop/core/options.js#L289). Additional methods on the Theme object are helpers for getting / setting and validating these colours.
Moving a dom object onto the theme would be replicating some of the design we already have by having the renderer be **the** object you extend, in this case to modify dom elements.

My point about it being more extensible is that if the constants object (or really more the renderer) gets a call when the theme has changed, then it can respond to that however it sees fit. In the ``<defs>`` case, the constants object doesn't need to dispose of the element to update it. It can simply update the specific subelement that references that colour with the new theme's colour.

Happy continuing this discussion but I just want to point out that some of this is premature because the renderer doesn't know about the theme object today, only the style object which corresponds to a block. The style object is enough for colouring a specific block, but the theme object contains colours for more than just block styles including overall "component" styles (eg: workspace background colour). Things like ``<defs>`` are generally used to apply a filter for all blocks, eg: this PR is adding it for selected blocks, so we somehow need somewhere to define these options. The theme seems appropriate (either "component" colours or a new separate category of colours), but I'm open to other ideas such as a `rendererOptions` object.

BeksOmega:
> My point about it being more extensible is that if the constants object (or really more the renderer) gets a call when the theme has changed, then it can respond to that however it sees fit. In the <defs> case, the constants object doesn't need to dispose of the element to update it. It can simply update the specific subelement that references that colour with the new theme's colour.

Ok, I think I'm starting to understand. It sounds like you're saying that renderers will be responsible for creating the patterns (embossed, crosshatch, etc) while the theme will be responsible for the colours. So everything about the patterns besides the colours will be static.

Is that in the ballpark?

ewpatton:
If the theme is changed, is there a corresponding event that gets fired? It seems to me that when the theme changes you’d probably need to rerender the whole workspace (in the worst case), so the tenderer object could update any internal state in response to the event.

BeksOmega:
> If the theme is changed, is there a corresponding event that gets fired? It seems to me that when the theme changes you’d probably need to rerender the whole workspace (in the worst case), so the tenderer object could update any internal state in response to the event.

That would be rad cuz it would decouple the systems. Would the [events timeout](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/events.js#L185) be a problem though? I'm not sure if it's actually a noticable delay when you're doing things like rendering.

But you could always create a separate observer pattern for themes if you wanted to go down that route. Since they're UI events anyway and shouldn't need to be considered for multiplayer.

core/block_svg.js
```diff
@@ -1242,16 +1242,14 @@ Blockly.BlockSvg.prototype.setHighlighted = function(highlighted) {
  * Select this block.  Highlight it visually.
  */
 Blockly.BlockSvg.prototype.addSelect = function() {
-  Blockly.utils.dom.addClass(
```
samelhusseini: Might want to think about whether we keep this as a "framework" level task, and let each renderer actually implement the details of what rendering selected looks like.

ewpatton:
+1. For real-time collaboration we need to be able to represent different users with different selection colors, so it’d be nice if we could also associate a selection with a particular “user”. Alternatively, I guess we could subclass the existing renderers to provide RTC variants.

samelhusseini:
@ewpatton for that we expect to extend the [markers](https://github.com/google/blockly/blob/master/core/workspace.js#L129) API to support multiple markers. A marker would be used to represent the "other" users and their respective locations. That's a better API than selection as it's not just limited to Blocks. A user could have their marker set to a field (say if they are editing a field). We are also planning on moving the [CursorSvg](https://github.com/google/blockly/blob/master/core/keyboard_nav/cursor_svg.js) code into the renderer, which would in turn determine what the UI for a marker set on a block, field, etc. would look like.

samelhusseini:
> Ok, I think I'm starting to understand. It sounds like you're saying that renderers will be responsible for creating the patterns (embossed, crosshatch, etc) while the theme will be responsible for the colours. So everything about the patterns besides the colours will be static.
> 
> Is that in the ballpark?

Kinda. In this case yes, but say there's a renderer out there that wants to dynamically respond to themes. When it receives a theme changed callback it can choose to destroy a ``<defs>`` element and recreate it if it isn't simply a color change.

BeksOmega:
> Kinda. In this case yes, but say there's a renderer out there that wants to dynamically respond to themes. When it receives a theme changed callback it can choose to destroy a `<defs>` element and recreate it if it isn't simply a color / property change.

It just seems weird to me to have the renderer respond to themes in this way. If you want to have logic that only depends on the theme, wouldn't it be easier to have it live on the theme? Putting the patterns/textures on the renderer seems like it just makes them harder to access.

I do hear your point about that making the JSON theme definitions trickier to impliment. Defs are /not/ something that can be easily JSON-ified.

But doesn't definining the patterns on the renderer also make the JSON less useful?

It seems like the main advantage of the JSON is that you can just crank out themes without touching the javascript. But if you have to write javascript to define your textures, it doesn't really matter whether you're putting the javascript in a renderer file or a themes file, you're still writing javascript.

There's probably something about the JSON that I don't know that makes it really useful in this context! And I'm totally willing to be made wrong.

But from my current perspective, if the only advantage of defining defs on the renderer is to preserve the JSON, that doesn't seem like enough of a benefit to outweigh the added complexity and code duplication.

samelhusseini:
Have a think about the simple 'geras' case here. We want setSelected to do something different in geras vs zelos. In geras you want to set the stroke width and colour, and in zelos you want to set a filter. If we forget for a sec where the zelos filters are defined. That renderer in geras still needs to [set](https://github.com/google/blockly/pull/3390/files#diff-489a0a18d0727d233f4670195c168665R162) the stoke to a specific colour which I imagine would come from the Theme object.
One way or another as the Theme changes (since it can be dynamic), the renderer would have to be called to reflect the update. 
@rachel-fenichel have been discussing this and our current thoughts are that when a renderer is initialized and when the theme changes, the constants object has an opportunity to look at the Theme object and copy over / set any internal properties on the constants object to reflect the theme configuration.
In this case, you can imagine that the Theme object would have a configuration "selectedColour", when the constants object is initialized / when the theme changes, the constants object would receive a Theme object (or a subset of it), and it can update it's property ``this.SELECTED_COLOUR = theme.components['selectedColour'];``, then when setSelected is called on a block's path object, it will reach into the constants object and query the SELECTED_COLOUR property for setting the stroke of the block to.

BeksOmega:
Ok yes that makes sense! The renderer controls the patterns/textures and the theme defines the colours/stroke widths/other values.

I thought from your last reply that you wanted the renderer to be able to change the pattern based on the theme (e.g. theme X has a polka dot disabled pattern, while theme Y has a crosshatch disabled pattern), which would get super messy.

But I this separation of responsibilities seems like it will keep everything nice and clean!

Issue #2929, opened 2019-08-29
Exception when try to move blocks within blockly div when blocks start shaking


ankuraga1508:
**Describe the bug**

Exception in the developer console when trying to move blocks after blocks start shaking

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/demos/fixed/index.html
2. Open developer console
3. Drag blocks in the blockly area over each other until there is no space left or blocks start shaking.
4. Now remove any block by putting it back
5. Notice this exception : 
blockly_compressed.js:1802 Uncaught TypeError: Cannot read property 'getRootBlock' of null
    at blockly_compressed.js:1802
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (blockly_compressed.js:1016)
    at Blockly.Events.fireNow_ (blockly_compressed.js:917)

**Expected behavior**

There should not be any error

![Capture](https://user-images.githubusercontent.com/15721876/63949623-50926580-ca40-11e9-869a-bbeb3b04d328.PNG)

**Video showing the steps**
[Video.zip](https://github.com/google/blockly/files/3556041/Video.zip)

ChromeOS : 74.0.3729.159 (Official Build) (64-bit)
Platform: 11895.118.0 (Official Build)
JavaScript: V8 7.4.288.28
Flash: 32.0.0.171

RoboErikG:
Definitely a bug, but things seem to keep working after the error so lower priority to fix.

ankuraga1508:
@RoboErikG Can we just handle the exception instead of throwing it?

RoboErikG:
@ankuraga1508 that would just hide the bug without fixing it and could cause more issues later on.

BeksOmega:
@NeilFraser I thought I fixed this when I fixed the shaking, but that must not have been the root problem. I can take a look in an hour or two.

samelhusseini:
Source of the exception: 
https://github.com/google/blockly/blob/master/core/inject.js#L306

samelhusseini:
Here are some simpler repro steps:
1. Load the Fixed demo: ``/demos/fixed/index.html``
2. Open the developer console, and run the following: 
```Blockly.Xml.domToWorkspace(Blockly.mainWorkspace, Blockly.Xml.textToDom(`<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_repeat_ext" x="11" y="0">
    <statement name="DO">
      <block type="controls_repeat_ext">
        <next>
[...]

Issue #3393, opened 2019-11-04
Deleting block throws error on small screens


NeilFraser:
* Edit playground's options to simply be:
```
      {
        trashcan: true,
        toolbox: toolbox
      }
```
* Open playground in simple mode.
* Open console so that the screen is small.
* Create a stack of two if-else blocks.
* Drag a block from the toolbox to the trash.

<img width="1287" alt="Screen Shot 2019-11-04 at 11 59 50" src="https://user-images.githubusercontent.com/250480/68153766-f78cf500-fefa-11e9-9d80-a6add21f9352.png">

In the console is this error:
```
inject.js:306 Uncaught TypeError: Cannot read property 'getRootBlock' of null
    at bumpObjects (inject.js:306)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (workspace.js:783)
    at Blockly.Events.fireNow_ (events.js:203)
```

Affects master and develop.  Found on Blockly Games.

NeilFraser:
Duplicate of https://github.com/google/blockly/issues/2929

PR #3394, opened 2019-11-04
Fixed getRootBlock error in bumpObjects

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2929 

### Proposed Changes

Wraps the object.getRootBlock call in a null check.

### Reason for Changes

It seems like there's a race condition. Deleting a block via drag fires a move condition. And then the block gets disposed. It seems like sometimes listeners receive the move event after the block has already been disposed, so they can't access it.

### Test Coverage

No test coverage! I wasn't able to reproduce the error, so someone else should take a look as well.

Tested on:

### Documentation

N/A

### Additional Information

I can close this since I wasn't able to reproduce if you want to take a look @samelhusseini 

I do think this should fix it though.

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -253,7 +253,9 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
             case Blockly.Events.BLOCK_CREATE:
             case Blockly.Events.BLOCK_MOVE:
               var object = mainWorkspace.getBlockById(e.blockId);
-              object = object.getRootBlock();
+              if (object) {
+                object = object.getRootBlock();
+              }
               break;
             case Blockly.Events.COMMENT_CREATE:
             case Blockly.Events.COMMENT_MOVE:


samelhusseini:
@BeksOmega I agree this "fixes" things, but it's rather just avoiding the exception. I would like someone (doesn't have to be me) to understand the underlying issue in further detail. I was able to re-produce this a little while ago, so I can add some more detail to the repro steps. @NeilFraser was also able to repro this today.

BeksOmega:
> @BeksOmega I agree this "fixes" things, but it's rather just avoiding the exception. I would like someone (doesn't have to be me) to understand the underlying issue in further detail. I was able to re-produce this a little while ago, so I can add some more detail to the repro steps. @NeilFraser was also able to repro this today.

I would love to dig into this, so more detail would be appreciated.

I tried the steps in both #3393 and #2929 on chrome and firefox and I couldn't get it to break. So I think that lends itself to the race condition theory.

I did trace down the "problem" move call to [here ](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_dragger.js#L284)(problem in quotes b/c it is used for undo). Assuming, again, that it is a race condition problem.

samelhusseini:
Thanks @BeksOmega.
Here's some repro steps that I've been able to consistently repro in Chrome: 
1. Load the Fixed demo: ``/demos/fixed/index.html``
2. Open the developer console, and run the following: 
```Blockly.Xml.domToWorkspace(Blockly.mainWorkspace, Blockly.Xml.textToDom(`<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_repeat_ext" x="11" y="0">
    <statement name="DO">
      <block type="controls_repeat_ext">
        <next>
          <block type="controls_repeat_ext">
            <next>
              <block type="controls_repeat_ext">
                <next>
                  <block type="controls_repeat_ext">
                    <next>
                      <block type="controls_repeat_ext">
                        <next>
                          <block type="controls_repeat_ext">
                            <statement name="DO">
                              <block type="controls_repeat_ext"/>
                            </statement>
                            <next>
                              <block type="controls_repeat_ext"/>
                            </next>
                          </block>
                        </next>
                      </block>
                    </next>
                  </block>
                </next>
              </block>
            </next>
          </block>
        </next>
      </block>
    </statement>
  </block>
</xml>`))
```
3. Drag the inner most repeat block into the flyout to delete it.
4. See exception in console.

Let me know if you still can't repro with the above.

BeksOmega:
Ok, It looks like it is as I suspected. The block is getting deleted before the move event fires. (Though admittedly I was wrong about it being a race condition b/c it does happen consistently)

I added a `console.log(e)` call at the beginning of bumpObjects, and a  `console.trace(event)` call inside Blockly.Events.fire. That gave me the following log:

<details><summary>Log</summary>
<p>

```
events.js:183 Blockly.Events.Move {
  workspaceId: "X6w3!yoq]d:WNPPk_j($",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  recordUndo: true,
  blockId: "f:_*bh:9^Z,Np^(9}O+Y",
  oldParentId: undefined,
  blockId: "f:_*bh:9^Z,Np^(9}O+Y",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  newCoordinate: Blockly.utils.Coordinate {x: -110.65393829345703, y: 9.066680908203125},
  newInputName: undefined,
  newParentId: undefined,
  oldCoordinate: Blockly.utils.Coordinate {x: 85.61273193359375, y: 72},
  oldInputName: undefined,
  oldParentId: undefined,
  recordUndo: true,
  workspaceId: "X6w3!yoq]d:WNPPk_j($" …}
    Blockly.Events.BlockBase
    Blockly.Events.fire @ events.js:183
    Blockly.BlockDragger.fireMoveEvent_ @ block_dragger.js:286
    Blockly.BlockDragger.maybeDeleteBlock_ @ block_dragger.js:303
    Blockly.BlockDragger.endBlockDrag @ block_dragger.js:243
    Blockly.Gesture.handleUp @ gesture.js:578
    Blockly.TouchGesture.handleUp @ touch_gesture.js:204
    wrapFunc @ blockly.js:481
events.js:183 Blockly.Events.Ui {
  workspaceId: "X6w3!yoq]d:WNPPk_j($",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  recordUndo: false,
  blockId: null,
  element: "selected", …}
    Blockly.Events.fire @ events.js:183
    Blockly.BlockSvg.unselect @ block_svg.js:305
    Blockly.BlockSvg.dispose @ block_svg.js:973
    Blockly.BlockDragger.maybeDeleteBlock_ @ block_dragger.js:304
    Blockly.BlockDragger.endBlockDrag @ block_dragger.js:243
    Blockly.Gesture.handleUp @ gesture.js:578
    Blockly.TouchGesture.handleUp @ touch_gesture.js:204
    wrapFunc @ blockly.js:481
events.js:183 Blockly.Events.Delete {
  workspaceId: "X6w3!yoq]d:WNPPk_j($",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  recordUndo: true,
  blockId: "f:_*bh:9^Z,Np^(9}O+Y",
  oldXml: block#f:_*bh:9^Z,Np^(9}O+Y, …}
    Blockly.Events.fire @ events.js:183
    Blockly.Block.dispose @ block.js:333
    Blockly.BlockSvg.dispose @ block_svg.js:1004
    Blockly.BlockDragger.maybeDeleteBlock_ @ block_dragger.js:304
    Blockly.BlockDragger.endBlockDrag @ block_dragger.js:243
    Blockly.Gesture.handleUp @ gesture.js:578
    Blockly.TouchGesture.handleUp @ touch_gesture.js:204
    wrapFunc @ blockly.js:481
inject.js:231 Blockly.Events.Ui {
  workspaceId: "X6w3!yoq]d:WNPPk_j($",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  recordUndo: false,
  blockId: "f:_*bh:9^Z,Np^(9}O+Y",
  element: "dragStop", …}
inject.js:231 Blockly.Events.Move {
  workspaceId: "X6w3!yoq]d:WNPPk_j($",
  group: "Nc7`1B1+PgHlBE*g^gNf",
  recordUndo: true,
  blockId: "f:_*bh:9^Z,Np^(9}O+Y",
  oldParentId: undefined, …}
inject.js:252 Uncaught TypeError: Cannot read property 'getRootBlock' of null
    at bumpObjects (inject.js:252)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (workspace.js:792)
    at Blockly.Events.fireNow_ (events.js:204)
```

</p>
</details>

This shows that the Delete event is getting fired (aka the block is being deleted) before the move event is received.

Now! The reason this only happens when blocks go beyond the bounds of the workspace is because of [this check](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/inject.js#L290). If there are only a few blocks in the workspace, that check doesn't pass.

I think that optimally this would be fixed inside the [BlockDragger](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_dragger.js#L276) (because move events firing after a block has been deleted can affect outside developers). But [the culprit move event](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_dragger.js#L283) is needed for undo, so we can't avoid firing it.

Maybe we could set the delete to happen after the events delay? But that seems a little hacky.

I'm not currently seeing a lot of other options. Opinions would be appreciated!

[EDIT: I'll look into how undo and move events are handled after I finish school, maybe there's some wiggle room there]

BeksOmega:
Ok I put up an alternative fix. This one fixes things at the BlockDragger level by encoding the block location in the delete event. It does so by moving the block to the start location before deleting.

This shouldn't be merged either because it has two problems:
1. moveTo seems pretty expensive, especially for deleting large stacks of blocks (it updates all connection locations).
2. The delete animation is now in the wrong location:
MoveToDelete

But I think this is maybe closer?

BeksOmega:
Ok, problem. It turns out this is not just an issue with dragging, it's also an issue when you delete the selected block.

<details><summary>Log</summary>
<p>

```
events.js:183 Blockly.Events.Move {
  workspaceId: "*Lgtg/0SAN)_}-B9,7l~",
  group: "2?*6`_:JwoG+RUzK/GnF",
  recordUndo: true,
  blockId: "x.Ej:,OvQx?M*kFTl5mc",
  oldParentId: "jsODwG?kW_QZB)aBFN}O",
  blockId: "x.Ej:,OvQx?M*kFTl5mc",
  group: "2?*6`_:JwoG+RUzK/GnF",
  newCoordinate: Blockly.utils.Coordinate {x: 85.61273193359375, y: 422},
  newInputName: undefined,
  newParentId: undefined,
  oldCoordinate: undefined,
  oldInputName: "DO",
  oldParentId: "jsODwG?kW_QZB)aBFN}O",
  recordUndo: true,
  workspaceId: "*Lgtg/0SAN)_}-B9,7l~",
  __proto__: Blockly.Events.BlockBase …}
    Blockly.Events.fire @ events.js:183
    Blockly.Connection.disconnectInternal_ @ connection.js:568
    Blockly.RenderedConnection.disconnectInternal_ @ rendered_connection.js:458
    Blockly.Connection.disconnect @ connection.js:543
    Blockly.Block.unplugFromStack_ @ block.js:488
    Blockly.Block.unplug @ block.js:406
    Blockly.BlockSvg.dispose @ block_svg.js:986
    Blockly.onKeyDown @ blockly.js:270
    wrapFunc @ blockly.js:481
events.js:183 Blockly.Events.Delete {
  workspaceId: "*Lgtg/0SAN)_}-B9,7l~",
  group: "2?*6`_:JwoG+RUzK/GnF",
  recordUndo: true, blockId: "x.Ej:,OvQx?M*kFTl5mc",
  oldXml: block#x.Ej:,OvQx?M*kFTl5mc, …}
    Blockly.Events.fire @ events.js:183
    Blockly.Block.dispose @ block.js:333
    Blockly.BlockSvg.dispose @ block_svg.js:1004
    Blockly.onKeyDown @ blockly.js:270
    wrapFunc @ blockly.js:481
inject.js:231 Blockly.Events.Ui {
  workspaceId: "*Lgtg/0SAN)_}-B9,7l~",
  group: "2?*6`_:JwoG+RUzK/GnF",
  recordUndo: false,
  blockId: null,
  element: "selected", …}
inject.js:231 Blockly.Events.Move {
  workspaceId: "*Lgtg/0SAN)_}-B9,7l~",
  group: "2?*6`_:JwoG+RUzK/GnF",
  recordUndo: true,
  blockId: "x.Ej:,OvQx?M*kFTl5mc",
  oldParentId: "jsODwG?kW_QZB)aBFN}O", …}
inject.js:253 Uncaught TypeError: Cannot read property 'getRootBlock' of null
    at bumpObjects (inject.js:253)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (workspace.js:792)
    at Blockly.Events.fireNow_ (events.js:204)
```

</p>
</details>

This time the move event is fired by `disconnectInternal_` which is called by `unplug` which is sadly a public function, so we can't really mess with that move event. We could maybe pass info about "This is a delete!" through the whole call stack, but like... I'd rather not.

Maybe we could try collapsing move events into delete events? Going to check that next.

BeksOmega:
So I think that merging moves into deletes /could/ work. But it will get really messy.

Delete events will need to duplicate a bunch of functionality from move events, basically [this function](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_events.js#L481) and associated properties.

The [event filtering ](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/events.js#L214)is also not /particularly/ built for this. Merging move and delete events will need to be special cased outside of the normal `!lastEntry` logic chain. Move events will need to be removed from the queue, which means indexing of all the events will need to be updated, which in turn means a drop in efficiency.

And we'll just have to hope that Events.filter keeps getting called exactly as it is currently (only on events in the fire queue, or events in a single group) otherwise we could get unrelated move and delete events merging.

Just to reiterate, it will get messy.

So is this something you guys want me to pursue? I'm personally a little bit ambivalent about it. But I also really hate this bug! I want to squash it! So I wanted to get your opinions on it.

rachel-fenichel:
I just dug around at this more, and I think that your original null check is the correct fix.  That change is basically an extension of the previous null check.

The line to look for the root block was [added](https://github.com/google/blockly/commit/23860381389669ded5519787befec34f3f2d1dbf) after the original null check was [added](https://github.com/google/blockly/commit/c43988873ee4660d85cc1790ae8623c0187837ff).  Since it only triggers in a nested block, it makes sense that we wouldn't hit it during normal drag testing.

It also makes sense to me because a delay in processing that event is plausible (say there's a lot going on on the main thread).  Removing the bump-move means that undo works correctly--the block goes back to being connected--so it never ends up in a bad state.  On the other hand, a move-delete-move sequence would cause issues during the undo.

BeksOmega:
Reverted to le original null check.

Issue #3397, opened 2019-11-05
All stock blocks in Blockly Games now black.


NeilFraser:
<img width="251" alt="Screen Shot 2019-11-04 at 19 02 53" src="https://user-images.githubusercontent.com/250480/68175370-e0b7c400-ff35-11e9-9d1c-c7fa5211dd4e.png">

All blocks defined by Blockly are now black in Blockly Games.  Both in compiled mode and uncompiled mode.  There are no errors in the console.  The relevant includes are:
```
goog.require('Blockly');
goog.require('Blockly.Constants.Logic');
goog.require('Blockly.geras.Renderer');
```
These includes appear to be consistent with those in https://github.com/google/blockly/blob/master/tests/compile/main.js yet I can't recreate this behaviour in Blockly develop's demos or tests.  It is a new behaviour since Blockly Games renders properly on Blockly master.

I'm rather puzzled since there's doesn't seem to be any code on Blockly Games' side to break.  All we do is `require` the block category and then use it in XML.  I thought we were done with this sort of breaking changes.

samelhusseini:
Intriguing. Let me take a look.

NeilFraser:
Yes, both Monica and I can see this when pulling from develop.  Master is fine.

NeilFraser:
And we are black again.  This time only in compressed mode.  Blockly games synced to Blockly develop is broken, master is fine.
![puzzle](https://user-images.githubusercontent.com/250480/71326716-e18bc180-24b3-11ea-99da-b52f33f4b261.png)

BeksOmega:
I believe this is because the Theme constructor's signature changed again. The [new constructor](https://github.com/google/blockly/blob/54cb686dcdca1ecfe66358706fb2cb73f7037fb1/core/theme.js#L41) also takes in a name parameter as well as the old things.

I [tagged this for docs updates](https://github.com/google/blockly/pull/3413#issuecomment-552646883). And it got added to issue #3138 .

I hope that fixes it for you! It's just an instinct though.

rachel-fenichel:
The fact that this keeps happening means we need to add a test for it, for both compressed and uncompressed.

samelhusseini:
Fixed.

alschmiedt:
Added #3585 to track creating an automated test for this issue.

PR #3399, opened 2019-11-05
Move parseBlockColour into Blockly.utils

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

@NeilFraser I chased the problem you're running into to a missing require in colours.

Fixes https://github.com/google/blockly/issues/3397

### Proposed Changes

https://github.com/google/blockly/blob/develop/core/utils/colour.js#L232 depends on ``Blockly.HSV_SATURATION`` which comes from ``Blockly.constants``. Even though this is required in ``blockly.js``, the Theme object [runs through and validates](https://github.com/google/blockly/blob/develop/core/theme.js#L164) all of the colours in the theme during initialization which in turn calls this method. 

This doesn't show up if you require `Blockly` first, but in the case of Blockly games, Blockly isn't the first thing that's required, which is why you end up in with the Theme object calling the validate method without ``Blockly.constants`` available.

Add missing require in colour utils.

### Test Coverage

Tested on:


NeilFraser:
Just to expand on why having that function here is a bad idea, the single call to `Blockly.utils.replaceMessageReferences` means we must also require `Blockly.utils` in this file, which implicitly requires:
```
goog.require('Blockly.Msg');
goog.require('Blockly.utils.Coordinate');
goog.require('Blockly.utils.global');
goog.require('Blockly.utils.string');
goog.require('Blockly.utils.style');
goog.require('Blockly.utils.userAgent');
```
Which chains on to require:
```
goog.require('Blockly.utils.Size');
```

It happens to work out at the moment, but this is how one gets recursive dependencies.

samelhusseini:
I agree in moving this to ``Blockly.utils``. It was actually originally in [block.js](https://github.com/google/blockly/blob/23a78c89e4c0f2801768b5c55c7f91ac261f4bc6/core/block.js#L963).

samelhusseini:
@NeilFraser updated.

samelhusseini:
No worries :)

rachel-fenichel:
Makes sense.  Thanks for the explanations.

PR #3401, opened 2019-11-05
Update JS-Interpreter to latest version

NeilFraser:
And fix alerting of falsy values.

The last version bump was committed on Oct 26, 2016.  Before Trump...


NeilFraser:
Heh, three years of development, and we are down 30 lines of code.

demos/interpreter/step-execution.html
```diff
@@ -153,21 +153,20 @@ <h1><a href="https://developers.google.com/blockly/">Blockly</a> &gt;
       // Add an API function for the alert() block, generated for "text_print" blocks.
       interpreter.setProperty(scope, 'alert',
           interpreter.createNativeFunction(function(text) {
-        text = text ? text.toString() : '';
+        text = arguments.length ? text : '';
```
samelhusseini: Out of curiosity why not `text != null``?

NeilFraser:
```
alert(null) => 'null'
alert(undefined) => 'undefined'
alert() => ''
```
I need to print a stringified version of every value -- except if there is no argument.  If `text === undefined` that's not enough to know what to print.

samelhusseini: Ah I see. Makes sense.

Issue #3344, opened 2019-10-25
AST should navigate to non editable fields


alschmiedt:
**Is your feature request related to a problem? Please describe.**
The ast should be able to traverse non editable fields.

**Describe the solution you'd like**
Set a flag to indicate whether or not we want the AST to return non editable fields. For now at least it seems like there are a decent amount of use cases that would not want to navigate to non editable fields, so having this check at the AST level could be useful.

**Describe alternatives you've considered**

**Additional context**

alschmiedt:
Feature added in #3404

PR #3404, opened 2019-11-05
Add flag to navigate to all fields

### LINKED ISSUE #3339: FieldImage cannot be navigated to when opt_clickHandler is passed
Currently, field click is determined whether or not the block is editable and the showEditor_ method is implemented.
Keyboard Navigation is determined whether or not the field is EDITABLE. This means that FieldImage that has an optional click handler passed in, cannot be navigated to.

Perhaps FieldImage should set EDITABLE to true, but then it would be SERIALIZABLE due to this code: https://github.com/google/blockly/blob/master/core/field.js#L434 that doesn't let fields be EDITABLE and SERIALIZABLE at the same time.

If we don't see a reason for a field to be EDITABLE and not SERIALIZABLE then we should at least fix keyboard nav to use isFieldClickable instead.

@alschmiedt @BeksOmega thoughts?

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3339.
And adds support for #3344

### Proposed Changes
Use isClickable() instead of EDITABLE when navigating around the AST. 
Add a flag on ASTNode to support going through all fields or only going through clickable fields.

### Test Coverage

Tested on:

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -83,6 +83,12 @@ Blockly.ASTNode.types = {
 };
 
 /**
+ * True to navigate to all fields. False to only navigate to clickable fields.
+ * @type {boolean}
+ */
+Blockly.ASTNode.NAVIGATE_ALL_FIELDS = false;
+
+/**
  * The default y offset to use when moving the cursor from a stack to the
  * workspace.
  * @type {number}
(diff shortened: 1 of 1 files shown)


samelhusseini:
My initial observation is that it seems this should be dealt with on the Cursor level. It's up to the cursor to choose to skip non-editable fields, and maybe the default one does. But the AST would include all fields. Thoughts?

alschmiedt:
We can do it at the cursor level however this seems like a pretty common use case and having it at the cursor level makes quite a bit more code (it has to be in methods for going in, out, previous, and next). This approach seems to simplify it a bit more for developers.

PR #3405, opened 2019-11-06
Move lots of update style functions from block_svg to path_object

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Moves more responsibility to the path object.

### Proposed Changes

Move some functions that apply styles to the block's svg root into the path object.

Note that unlike #3390 this doesn't address moving styling into the renderer code.  This just moves the existing addition/removal of CSS styles.

### Reason for Changes
Pre-work for different renderers implementing each of these differently.
Remove the pattern of having lots of different functions that really just decide whether to add or remove a class on the svg root.

### DIFF
--- a/core/renderers/common/path_object.js
+++ b/core/renderers/common/path_object.js
@@ -129,7 +129,7 @@ Blockly.blockRendering.PathObject.prototype.setStyle = function(blockStyle) {
  *     be removed.
  * @protected
  */
-Blockly.blockRendering.PathObject.prototype.setClass = function(
+Blockly.blockRendering.PathObject.prototype.setClass_ = function(
     className, add) {
   if (add) {
     Blockly.utils.dom.addClass(/** @type {!Element} */ (this.svgRoot),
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
fyi @BeksOmega @ewpatton

core/renderers/common/path_object.js
```diff
@@ -121,14 +122,33 @@ Blockly.blockRendering.PathObject.prototype.setStyle = function(blockStyle) {
...
 };
 
+/**
+ * Add or remove the given CSS class on the path object's root SVG element.
+ * @param {string} className The name of the class to add or remove
+ * @param {boolean} add True if the class should be added.  False if it should
+ *     be removed.
+ * @protected
+ */
+Blockly.blockRendering.PathObject.prototype.setClass = function(
```
samelhusseini: underscore in the method name for @protected methods.

samelhusseini:
Compiler complaining that the geras path object also needs to be updated with these changes.

BeksOmega:
LGTM. I really like the setClass addition!

Issue #2979, opened 2019-09-06
[zelos] Render empty inline inputs in separate path


samelhusseini:
**Is your feature request related to a problem? Please describe.**
Zelos renders empty "placeholder" shapes for inputs without a shadow block: 
ie, instead of: 
<img width="420" alt="Screen Shot 2019-09-06 at 9 54 09 AM" src="https://user-images.githubusercontent.com/16690124/64445898-51547880-d08c-11e9-9e87-bb812aaaae7e.png">
we should render:
<img width="420" alt="Screen Shot 2019-09-06 at 9 54 09 AM" src="https://user-images.githubusercontent.com/16690124/64445888-4d285b00-d08c-11e9-847c-ad7a09753e9f.png">

**Describe the solution you'd like**
The renderer needs to be able to add custom paths onto the block. Once we have the ability to do that, we can render default shapes depending on the type of input and add them to the block such that they show above the `svgPath`.

**Additional context**

I'm not sure who should be in charge of hiding / showing them.

samelhusseini:
Fixed.

PR #3410, opened 2019-11-07
[zelos] Add input outlines on the path object.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/2979

### Proposed Changes

Create input outlines on the path object and maintain a reference to them. 
When a block is drawn, go through and remove all input outlines (as some inputs may have been removed).
When drawing an inline input, create a new outline path. Generate it's shape from the input shape, and send the path object the path to set on the outline svg path.

In this change, I also allow the configuration of the FONTWEIGHT in JS. This should probably eventually end up on the theme object. (This can be pulled out into a separate change if you like)

I also realized the we had the height of the inline input all along and didn't have to wait till the finalize phase to set it's dimensions, so I moved the back into the constructor. 

### Reason for Changes

Zelos rendering

### Test Coverage

Tested both geras and zelos.

Tested on:
* Desktop Chrome

### Additional Information

Screen Shot 2019-11-06 at 3 47 12 PM
Screen Shot 2019-11-06 at 3 46 58 PM


core/renderers/measurables/inputs.js
```diff
@@ -87,36 +87,26 @@ Blockly.blockRendering.InlineInput = function(constants, input) {
...
     this.width = this.connectedBlockWidth;
     this.height = this.connectedBlockHeight;
   }
-  this.connectionHeight = height;
-  this.connectionWidth = width;
+
+  this.connectionHeight = this.shape.height;
+  this.connectionWidth = !this.isDynamicShape ? this.shape.width :
+    this.shape.width(this.height);
+  if (!this.connectedBlock) {
```
rachel-fenichel: Merge this with the if statement on line 92.

core/renderers/zelos/drawer.js
```diff
@@ -143,8 +150,31 @@ Blockly.zelos.Drawer.prototype.drawFlatBottom_ = function() {
  * @override
  */
 Blockly.zelos.Drawer.prototype.drawInlineInput_ = function(input) {
-  // Don't draw an inline input.
   this.positionInlineInputConnection_(input);
+
+  var hasConnection = !!input.connection.targetConnection;
+  var isInsertionMarker = input.connection.getSourceBlock().isInsertionMarker();
```
rachel-fenichel: Is this information already on the info object? And if not, should it be?  This is reaching out of rendering and into getting properties from the block directly.

core/renderers/zelos/drawer.js
```diff
@@ -45,6 +45,13 @@ Blockly.zelos.Drawer = function(block, info) {
 Blockly.utils.object.inherits(Blockly.zelos.Drawer,
     Blockly.blockRendering.Drawer);
 
+/**
+ * @override
+ */
+Blockly.zelos.Drawer.prototype.draw = function() {
+  this.block_.pathObject.clearOutlines();
```
rachel-fenichel: Is there a way to avoid clearing and recreating these every time?  As operations go, creating and deleting DOM nodes seems expensive.

core/renderers/zelos/drawer.js
```diff
@@ -143,8 +150,31 @@ Blockly.zelos.Drawer.prototype.drawFlatBottom_ = function() {
  * @override
  */
 Blockly.zelos.Drawer.prototype.drawInlineInput_ = function(input) {
-  // Don't draw an inline input.
   this.positionInlineInputConnection_(input);
+
+  var hasConnection = !!input.connection.targetConnection;
```
rachel-fenichel: What's the case when an inline input doesn't have a connection?

core/renderers/zelos/info.js
```diff
@@ -346,16 +346,6 @@ Blockly.zelos.RenderInfo.prototype.finalize_ = function() {
   for (var i = 0, row; (row = this.rows[i]); i++) {
     row.xPos = this.startX;
 
-    for (var j = 0, elem; (elem = row.elements[j]); j++) {
```
rachel-fenichel: +1

core/renderers/zelos/path_object.js
```diff
@@ -87,3 +94,47 @@ Blockly.zelos.PathObject.prototype.updateSelected = function(enable) {
...
+ * Create's an outline path for the specified input.
+ * @param {string} name The input name.
+ * @return {!SVGElement} The SVG outline path.
+ * @package
+ */
+Blockly.zelos.PathObject.prototype.createOutlinePath = function(name) {
+  if (!this.outlines_[name]) {
+    this.outlines_[name] = Blockly.utils.dom.createSvgElement('path', {
+      'class': 'blocklyOutlinePath',
+      // IE doesn't like paths without the data definition, set empty default
```
rachel-fenichel: Do I also need to do this for the calls to create the main, light, and dark paths?

core/renderers/measurables/inputs.js
samelhusseini: The connectionWidth depends on the``this.height``. And the hight is only calculated once we're through with that statement.

core/renderers/zelos/drawer.js
samelhusseini: There should be. The main problem is not knowing when an input has been deleted of the block. What we could do is mark that the path object at the start of the draw to watch what inputs get used, and remove any remaining ones at the end of the draw.

core/renderers/zelos/drawer.js
samelhusseini: Unlike geras we don't want to keep around these outlines if we have a connection.

core/renderers/zelos/drawer.js
```diff
@@ -143,8 +150,31 @@ Blockly.zelos.Drawer.prototype.drawFlatBottom_ = function() {
...
+  var isInsertionMarker = input.connection.getSourceBlock().isInsertionMarker();
+
+  var inputName = input.input.name;
+  if (!!hasConnection || isInsertionMarker) {
```
samelhusseini: lol this is funny. This should be hasConnection

core/renderers/zelos/drawer.js
samelhusseini: True, let me look into that.

core/renderers/zelos/path_object.js
samelhusseini: Possibly, will dig up the bug that introduced this and try to repro.

core/renderers/zelos/drawer.js
rachel-fenichel: Got it.

core/renderers/zelos/drawer.js
samelhusseini: hasConnection can be replaced with ``input.connectedBlock``, but I don't see us using this elsewhere. I can add the meta data for whether the connectedBlock onto the input measurable.

samelhusseini:
@rachel-fenichel Updated to keep the outlines around and not recreated them during every draw pass.

core/renderers/common/path_object.js
```diff
@@ -220,3 +220,19 @@ Blockly.blockRendering.PathObject.prototype.updateInsertionMarker = function(
 Blockly.blockRendering.PathObject.prototype.updateMovable = function(enable) {
   this.setClass_('blocklyDraggable', enable);
 };
+
+/**
+ * Method that's called when the drawer is about to draw the block.
+ * @package
+ */
+Blockly.blockRendering.PathObject.prototype.beginDrawing = function() {
```
rachel-fenichel: I don't think you need these on the base path object if you're not putting them onto the interface.  You're only using them in the zelos case, and at that point you know you have a zelos path object.

core/renderers/zelos/drawer.js
```diff
@@ -143,8 +143,26 @@ Blockly.zelos.Drawer.prototype.drawFlatBottom_ = function() {
...
+  }
+
+  var width = input.width - input.connectionWidth;
+  var height = input.height;
+  var yPos = input.centerline - height / 2;
+
+  var connectionRight = input.xPos + input.connectionWidth;
+
+  var outlinePath = Blockly.utils.svgPaths.moveTo(connectionRight, yPos) +
+      Blockly.utils.svgPaths.lineOnAxis('h', width - input.connectionWidth) +
```
rachel-fenichel: You're defining width up above:
`var width = input.width - input.connectionWidth`
then only using it with an extra subtraction:
`width - input.connectionWidth`

So line 153 can become
`var width - input.width - 2 * input.connectionWidth`
 and the later references can just use `width`.

samelhusseini: Good catch.

core/renderers/common/path_object.js
samelhusseini: Removed.

core/renderers/zelos/path_object.js
```diff
@@ -87,3 +104,75 @@ Blockly.zelos.PathObject.prototype.updateSelected = function(enable) {
...
+    // The value set here isn't used anywhere, we are just using the
+    // object as a Set data structure.
+    this.remainingOutlines_[key] = 1;
+  }
+};
+
+/**
+ * @override
+ */
+Blockly.zelos.PathObject.prototype.endDrawing = function() {
```
rachel-fenichel: No longer an override, here and for `beginDrawing`.

LGTM after you update those comments.

Issue #3407, opened 2019-11-06
Add the current theme to the DOM


samelhusseini:
**Is your feature request related to a problem? Please describe.**

It's currently not possible to add CSS that is only relevant for a specific theme used. Say you have a default theme, and a high contrast theme. If you wanted to add additional properties to your CSS that only applied to the high contrast theme, that wouldn't be possible.

**Describe the solution you'd like**

Simple solution, set the current theme name as a class on the injection div.

**Describe alternatives you've considered**

**Additional context**

samelhusseini:
Fixed.

BeksOmega:
I've been really dying to ask this, but I've honestly been scared to. Why did you decide to go with themes over css? Don't get me wrong, I love themes, but it just feels like css would be the first method to try. Is there maybe an IE thing you ran into that makes css unusable?

samelhusseini:
Hmm. I'm not sure what you mean?
Do you mean why we originally used JS to set component colours as opposed to doing this from the start?

BeksOmega:
> Hmm. I'm not sure what you mean?
> Do you mean why we originally used JS to set component colours as opposed to doing something like this from the start?
> ps: Always feel free to always ask questions!

Yeah, like go through and make sure all dom elements were tagged with classes, and then allow styling with those.

samelhusseini:
Yeah they are, and what we've settled on is: 
- Use CSS for styling
- For commonly configured colours use the themes object (workspace, toolbox, flyout). This makes it easier for developers to create dynamic themes (such as high contrast) that changes block colours, category colours, and major components of Blockly.
- For more complex theme adjustments (such as hover), we fallback on using CSS with the theme selector. 

The downside of using CSS for colours is that it cannot be (easily) dynamically adjusted. I say easily because the caveat there is you could dump out the CSS from a div, and re-inject new CSS that has the updated colours.
Doing the same thing with CSS would mean selectors like this: 
```
.dark-theme .blocklySvg {
 ....
}
.dark-theme .blocklyToolboxDiv {
 ....
}
```
etc. 

The above has a couple of issues. In order to adjust the theme, developers need to know of our CSS selectors. eg: we have the following selector ``.blocklyTreeRow:not(.blocklyTreeSelected):hover``. First of all, blocklyTreeRow isn't a great name for toolbox category, and with a toolbox re-write we'll probably change those to better names. Second of all, if you are adjusting the theme, without full knowledge of our CSS, you might think to add: 
```
[...]

PR #3413, opened 2019-11-07
Add the theme name to the injection div

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3407

### Proposed Changes

Each theme now has a name.
Set the name of the theme as a class onto the injection div. 

### Reason for Changes

So we can use CSS to customize a theme further.

### Test Coverage

Tested playground.
Ran build.

Tested on:
* Desktop Chrome


tests/mocha/trashcan_test.js
```diff
@@ -16,7 +16,7 @@
  */
 
 suite("Trashcan", function() {
-  var themeManager = new Blockly.ThemeManager(Blockly.Themes.Classic);
+  var themeManager = new Blockly.ThemeManager(null, Blockly.Themes.Classic);
```
rachel-fenichel: Not allowed to pass null here.

core/theme/dark.js
```diff
@@ -115,7 +116,7 @@ Blockly.Themes.Dark.categoryStyles = {
 
 // This style is still being fleshed out and may change.
 Blockly.Themes.Dark =
-    new Blockly.Theme(Blockly.Themes.Dark.defaultBlockStyles,
+    new Blockly.Theme('dark', Blockly.Themes.Dark.defaultBlockStyles,
```
rachel-fenichel: `.dark` is very vague.  I suggest that we either auto-append `'Theme'` to their name, or add it just for our built-in themes.

samelhusseini: Fair enough.

tests/mocha/trashcan_test.js
samelhusseini: Fixed.

core/theme/dark.js
samelhusseini: Done.

BeksOmega:
This change should be tagged for doc updates, as Blockly.Theme is a public constructor. The section on [creating a theme](https://developers.google.com/blockly/guides/configure/web/themes#create_a_theme) needs to be updated to include the name parameter.

Also important for upgrade notes (if you're doing that for the next release).

PR #3417, opened 2019-11-08
Add extern for goog.

NeilFraser:
As of closure-compiler-v20191027.jar the compile now breaks due to a lack of ‘goog’ being defined.

```
../../temp_core/constants.js:24: ERROR - [JSC_UNDEFINED_VARIABLE] variable goog is undeclared
goog.provide('Blockly.constants');
^^^^

1 error(s), 0 warning(s)
Compiler exit code: 1
Compilation FAIL.
```

This is despite all usages of `goog` being compiler directives, and never being in the output.

PR #3418, opened 2019-11-08
Upgrade closure compiler to v20191027

samelhusseini:
Use this opportunity to upgrade the closure compiler to v20191027.

Release notes: https://github.com/google/closure-compiler/wiki/Releases

@NeilFraser I'm checking this into your branch as it has the same dependency on that goog-externs file.

Requires https://github.com/google/blockly/pull/3417


package-lock.json
```diff
@@ -5468,6 +5642,12 @@
       "resolved": "https://registry.npmjs.org/safer-buffer/-/safer-buffer-2.1.2.tgz",
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg=="
     },
+    "samsam": {
```
NeilFraser: Are all these intentional?

samelhusseini: Yeah we should update package-lock.json whenever updating package.json.

PR #3424, opened 2019-11-11
Full-block editors for single-field reporters.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of Zelos
### Proposed Changes

Adds a switch to text input fields that makes the editor fully cover the block for Zelos reporters.

image

Add a constant on the rendering constants to control this.
Add rounded corners to the widget div when needed.

### Reason for Changes

Fields on simple reporters act like they cover the entire block, both when being clicked and when being edited.

### Test Coverage
I added a test block in the playground, and generally used this XML:

```
  <block type="text_isEmpty" id="MEP:T0V^)nwI%G.*$:X(" x="12" y="12">
    <value name="VALUE">
      <shadow type="test_fields_only_text_input" id="t#1Ay*.2;,(nc~xDT7x7">
        <field name="TEXT_INPUT">default</field>
      </shadow>
    </value>
  </block>
  <block type="test_fields_only_text_input" id=".}E2;kV]}J`I:Zb_erHc" x="87" y="63">
    <field name="TEXT_INPUT">default</field>
  </block>
  <block type="controls_repeat" id="(x!H5h!]mTROLgfyLZy]" x="38" y="113">
    <field name="TIMES">10</field>
  </block>
</xml>
```

Notice that the number field on the controls_repeat block is handled correctly.

### Additional Information

We'll need to do something similar for the colour field, but this handles all fields that inherit from textinput, including number fields.

It's nowhere near pixel perfect.


core/field_textinput.js
```diff
@@ -128,6 +135,50 @@ Blockly.FieldTextInput.prototype.configure_ = function(config) {
   }
 };
 
+/**
+ * Create the block UI for this field.
+ * @package
+ */
```
alschmiedt: Should this be override from field?

core/field_textinput.js
```diff
@@ -128,6 +135,50 @@ Blockly.FieldTextInput.prototype.configure_ = function(config) {
...
+Blockly.FieldTextInput.prototype.initView = function() {
+  var renderer = this.sourceBlock_.workspace.getRenderer();
+  if (renderer.getConstants().FULL_BLOCK_FIELDS) {
+    // Step one: figure out if this is the only field on this block.
```
alschmiedt: Not necessary, but might be nice to put some of the logic for figuring out if it is a fullBlockClickTarget in its own method.

core/field_textinput.js
rachel-fenichel: Yup.  Done.

core/field_textinput.js
rachel-fenichel: Agreed.  I'm going to see how this works out with the colour field (which may be the only other field that needs this function directly) and then see if I can move all of this logic onto the renderer instead.

PR #3427, opened 2019-11-12
Fixed comment event listener disposal.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2152

### Proposed Changes

Saves event data for bound events, and properly disposes of it.

### Reason for Changes

Proper memory handling.

### Test Coverage

I ran the following stress test while doing a performance profile:
```
      function memoryStrike(n) {
        var toggleCount = 0;
        var block = /** @type Blockly.BlockSvg */Blockly.Xml.domToBlock(Blockly.Xml.textToDom(
          '<block type="controls_if">' +
          '</block>'), workspace);
        for (var i = 0; i < 1000; i++) {
          block.mutator.setVisible(true);
          block.mutator.setVisible(false);
          toggleCount++;
        }
        console.log(toggleCount);
      }
```

The following are the profiles from my tests. You can view them by opening the chrome console, going to the performance tab, and clicking "Load profile...".

#### Before changes:
[Profile-Orig.zip](https://github.com/google/blockly/files/3833694/Profile-Orig.zip)
6190 Listeners
#### After comment updates:
[Profile-Comment.zip](https://github.com/google/blockly/files/3833697/Profile-Comment.zip)
1814 Listeners
#### After bubble updates:
[Profile-Bubble.zip](https://github.com/google/blockly/files/3833699/Profile-Bubble.zip)
82 Listeners Woot!

Tested on:
* Desktop Chrome

### Documentation

N/A

### Additional Information

Slowly making icons dispose correctly!

### DIFF
--- a/core/comment.js
+++ b/core/comment.js
@@ -61,6 +61,34 @@ Blockly.Comment = function(block) {
    */
   this.cachedText_ = '';
 
+  /**
+   * Mouse up event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onMouseUpWrapper_ = null;
+
+  /**
+   * Wheel event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onWheelWrapper_ = null;
+
+  /**
+   * Change event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onChangeWrapper_ = null;
+
+  /**
+   * Input event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onInputWrapper_ = null;
+
   this.createIcon();
 };
 Blockly.utils.object.inherits(Blockly.Comment, Blockly.Icon);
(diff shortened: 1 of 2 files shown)


core/comment.js
```diff
@@ -271,7 +302,22 @@ Blockly.Comment.prototype.disposeBubble_ = function() {
     Blockly.Warning.prototype.disposeBubble.call(this);
     return;
   }
-
+  if (this.onMouseUpWrapper_) {
+    Blockly.unbindEvent_(this.onMouseUpWrapper_);
+    this.onMouseUpWrapper_ = null;
```
samelhusseini: Do we need to set these to null?

BeksOmega: Since we're disposing of the whole comment, I think it's ok to remove that line. If it were inside a something like a field editor dispose it's probably safer to set it to null, but that's probably matter of opinion.

Fixin'!

samelhusseini: Yep!

BeksOmega: My bad, completely read this wrong. I think it is probably safer to leave this here because the comment is still alive, we're just disposing of the bubble. This way it works the same as when fields dispose of their editor listeners.

samelhusseini: Okay that sounds good to me. Note to self, we should do the same in all the fields (eg: [here](https://github.com/google/blockly/blob/develop/core/field_textinput.js#L363)) since as you said we're not disposing of the entire object when widgetDispose_ (or the like) is called .

BeksOmega: Already on it =D

PR #3428, opened 2019-11-12
Move dropdown and widget divs inside injectionDiv

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Move dropdown div and widget div inside injection div.

### Test Coverage

Tested on:


alschmiedt:
Moving the widgetDiv and dropdownDiv inside the injectionDiv limits the size of the divs because they can no longer go outside of the injectionDiv. This is different than our current behavior where the size is not limited. Closing because this would negatively impact users with small screens and large widget/dropdown divs.

PR #3429, opened 2019-11-12
Cleaned up unbindEvent_ calls

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Every time there was an `unbindEvent_` call that wasn't part of disposing of the entire object I set the event data object to null. This pattern was already used some places, like scrollbars, and the workspace comment renderer.

### Reason for Changes

This makes sure that if the function which contains the unbinding is called, we don't have unnecessary calls to `unbindEvent_`.

If we are disposing of the entire object, setting the event data object to null is unnecessary because we'll be getting rid of that reference anyway.

### Test Coverage

These changes shouldn't affect behavior.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -84,6 +84,13 @@ Blockly.FieldTextInput = function(opt_value, opt_validator, opt_config) {
    * @private
    */
   this.onKeyInputWrapper_ = null;
+
+  /**
+   * Blur input event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onBlurInputWrapper_ = null;
 };
 Blockly.utils.object.inherits(Blockly.FieldTextInput, Blockly.Field);
 
@@ -349,7 +356,6 @@ Blockly.FieldTextInput.prototype.bindInputEvents_ = function(htmlInput) {
   this.onKeyInputWrapper_ =
       Blockly.bindEventWithChecks_(
           htmlInput, 'input', this, this.onHtmlInputChange_);
-
   this.onBlurInputWrapper_ =
       Blockly.bindEventWithChecks_(
           htmlInput, 'blur', this, this.onHtmlInputBlur_);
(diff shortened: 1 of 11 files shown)


core/field_textinput.js
```diff
@@ -362,9 +362,11 @@ Blockly.FieldTextInput.prototype.bindInputEvents_ = function(htmlInput) {
...
   if (this.onKeyDownWrapper_) {
     Blockly.unbindEvent_(this.onKeyDownWrapper_);
+    this.onKeyDownWrapper_ = null;
   }
   if (this.onKeyInputWrapper_) {
     Blockly.unbindEvent_(this.onKeyInputWrapper_);
+    this.onKeyInputWrapper_ = null;
   }
   if (this.onBlurInputWrapper_) {
     Blockly.unbindEvent_(this.onBlurInputWrapper_);
```
samelhusseini: Missed one?

samelhusseini: ``onBlurInputWrapper_``

core/workspace_comment_svg.js
```diff
@@ -49,6 +49,21 @@ goog.require('Blockly.WorkspaceComment');
...
+   */
+  this.onMouseUpWrapper_ = null;
+
+  /**
+   * Mouse move event data.
+   * @type {?Blockly.EventData}
+   * @private
+   */
+  this.onMouseMoveWrapper_ = null;
+
```
samelhusseini: Looks like these were just added, but not set to anything.

core/field_textinput.js
BeksOmega: Haha I think this may have appeared when I rebased, will fix!

core/workspace_comment_svg.js
BeksOmega: They're [used inside the workspace_comment_render_svg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/workspace_comment_render_svg.js#L266) file, so I believe that means they need to be declared here. Honestly I'm not sure though.

samelhusseini: Sounds good.

PR #3433, opened 2019-11-12
Pull comment CSS into comment modules.

NeilFraser:


core/comment.js
```diff
@@ -413,3 +413,21 @@ Blockly.Comment.prototype.dispose = function() {
   this.block_.comment = null;
   Blockly.Icon.prototype.dispose.call(this);
 };
+
```
samelhusseini: should goog.require('Blockly.Css')

core/workspace_comment_svg.js
```diff
@@ -610,3 +610,68 @@ Blockly.WorkspaceCommentSvg.prototype.toXmlWithXY = function(opt_noId) {
   element.setAttribute('w', this.getWidth());
   return element;
 };
+
```
samelhusseini: same here.

PR #3439, opened 2019-11-13
Move cursorSvg and markerSvg to pathObject

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Move cursor svg and marker svg into the path object.

### Reason for Changes

Moving markers and cursors into rendering generally.

### Test Coverage
Tested in playground with zelos and thrasos.


rachel-fenichel:
Fields don't have the equivalent of a path object--they hold their own svg directly, and the cursor will stay on the field.

Issue #2694, opened 2019-07-23
Labels With Different Font Size are Not Rendered Correctly


BeksOmega:
**Describe the bug**

Labels can be passed a class. If this class gives the label a different font size, it renders outside the bounds of the block.

**To Reproduce**

1. Add the following block definitions to the test blocks file:
```
{
    "type": "example_text_class",
    "message0": "%1",
    "args0": [
      {
        "type": "field_label",
        "name": "NAME",
        "text": "default",
        "class": "custom-text-class"
      }
    ],
    "colour": 230,
  },
  {
    "type": "example_text_class_image",
    "message0": "%2 %1",
    "args0": [
      {
        "type": "field_image",
        "src": "https://www.gstatic.com/codesite/ph/images/star_on.gif",
        "width": 50,
        "height": 50,
        "alt": "*",
        "flipRtl": false
      },
      {
        "type": "field_label",
        "name": "NAME",
        "text": "default",
        "class": "custom-text-class"
      }
    ],
    "colour": 230,
  },
```
2. Add the following css to the playground:
```
.custom-text-class {
  fill: black;  /* for visibility*/
  font-size: 50px;
}
```
3. Load the playground.
4. Import the following XML.
```
<xml>
  <block type="example_text_class"/>
  <block type="example_text_class_image" y="50"/>
</xml>
```
5. Observe how the text is rendered oddly. Especially on the block with the image.

**Expected behavior**

I expect the top of the text to be aligned with the top of the block (as that seems to be the [current official behavior](https://github.com/google/blockly/issues/1430#issuecomment-342672979)). The block should encapsulate all of the text.

**Screenshots**

![Text_Classes](https://user-images.githubusercontent.com/25440652/61754229-38efff00-ad67-11e9-90ba-58f0ca21efc7.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 75

(I believe this is not browser specific)

**Stack Traces**

N/A

**Additional context**

I think the problem is probably caused by [this line](https://github.com/google/blockly/blob/bb8fb10ef7c3eea9a83a96b2144199b4a967b02b/core/field_label.js#L84) being a constant (since [size_.height is constant](https://github.com/google/blockly/blob/bb8fb10ef7c3eea9a83a96b2144199b4a967b02b/core/field_label.js#L47)).

[...]

PR #3443, opened 2019-11-14
Field textinput pixel perfect input

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Pixel perfect input.
Related to https://github.com/google/blockly/issues/2694. We still need to know the font height.

### Proposed Changes

Fix sizing of text input field in zelos. 
Fix sizing issues in other browsers. 

### Reason for Changes

Rendering.

### Test Coverage

Tested the playground with both geras, thrasos and zelos.

Tested on:
* Desktop Chrome
* Desktop Firefox
* Desktop Safari

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -360,7 +360,7 @@ Blockly.Css.CONTENT = [
     'fill: #000;',
   '}',
 
-  '.blocklyEditableText:hover>rect {',
+  '.blocklyEditableText:not(.editing):hover>rect {',
     'stroke: #fff;',
     'stroke-width: 2;',
   '}',
@@ -427,6 +427,7 @@ Blockly.Css.CONTENT = [
     'padding: 0;',
     'width: 100%;',
     'text-align: center;',
+    'display: block;',
   '}',
 
   /* Edge and IE introduce a close icon when the input value is longer than a
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -248,14 +248,7 @@ Blockly.FieldTextInput.prototype.render_ = function() {
   // This logic is done in render_ rather than doValueInvalid_ or
   // doValueUpdate_ so that the code is more centralized.
   if (this.isBeingEdited_) {
-    if (this.sourceBlock_.RTL) {
-      // in RTL, we need to let the browser reflow before resizing
-      // in order to get the correct bounding box of the borderRect
-      // avoiding issue #2777.
-      setTimeout(this.resizeEditor_.bind(this), 0);
-    } else {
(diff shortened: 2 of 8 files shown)


core/css.js
```diff
@@ -360,7 +360,7 @@ Blockly.Css.CONTENT = [
     'fill: #000;',
   '}',
 
-  '.blocklyEditableText:hover>rect {',
+  '.blocklyEditableText:not(.editing):hover>rect {',
```
samelhusseini: Firefox, Safari include the stroke when calculating sizing info. This ensures that when we calculate the size, there's not stroke.

core/field_textinput.js
```diff
@@ -638,21 +619,15 @@ Blockly.FieldTextInput.prototype.getValueFromEditorText_ = function(text) {
  */
 Blockly.FieldTextInput.prototype.getScaledBBox = function() {
   if (this.fullBlockClickTarget_) {
-    var heightWidth = this.sourceBlock_.getHeightWidth();
+    var xy = this.getClickTarget_().getBoundingClientRect();
   } else {
-    var heightWidth = this.borderRect_.getBBox();
+    var xy = this.borderRect_.getBoundingClientRect();
```
samelhusseini: The trouble with the previous method (bbox + absoluteXY), is that absolute XY gets the size info of the click target / field svg root. This includes both the border rect and the text element. The text element (currently in geras) is slightly larger than the border rect which introduces the discrepancy when calculating positioning. The goal of sizing the text input is to match the dimensions of the border rect, hence why I'm switching it here to only use those dimensions.

tests/mocha/field_angle_test.js
```diff
@@ -201,10 +201,14 @@ suite('Angle Fields', function() {
       this.angleField.htmlInput_ = Object.create(null);
       this.angleField.htmlInput_.oldValue_ = '1';
       this.angleField.htmlInput_.untypedDefaultValue_ = 1;
+      this.stub = sinon.stub(this.angleField, 'resizeEditor_');
```
rachel-fenichel: Why do you need to stub this?  The FieldAngle shouldn't be rendered, so resizeEditor shouldn't be called.

samelhusseini: ``onHtmlInputChange_`` is called, which in turn now calls ``resizeEditor_``

core/field_textinput.js
rachel-fenichel: Cool.  Removing the workarounds certainly seems positive to me.

PR #3444, opened 2019-11-14
Highlight input shape in Zelos

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Highlight input shape in zelos.

### Proposed Changes

Highlight input shape on drag in Zelos.

Add a method on the block and the renderer that determines whether or not to insert a dragged block into a stack. We could consider moving these into an object that defines how the insertion marker does its work if we get more of these.

### Reason for Changes

Screen Shot 2019-11-14 at 11 27 09 AM

### Test Coverage
Tested on:

### DIFF
--- a/core/insertion_marker_manager.js
+++ b/core/insertion_marker_manager.js
@@ -409,7 +409,7 @@ Blockly.InsertionMarkerManager.prototype.shouldReplace_ = function() {
   if (local.type == Blockly.OUTPUT_VALUE) {
     // Insert the dragged block into the stack if possible.
     if (closest &&
-        local.getSourceBlock()
+        this.workspace_.getRenderer()
             .shouldInsertDraggedBlock(this.topBlock_, closest)) {
       return false; // Insert.
     }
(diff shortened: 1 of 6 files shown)


core/insertion_marker_manager.js
```diff
@@ -408,9 +408,9 @@ Blockly.InsertionMarkerManager.prototype.shouldReplace_ = function() {
   // Dragging a block over an existing block in an input.
   if (local.type == Blockly.OUTPUT_VALUE) {
     // Insert the dragged block into the stack if possible.
-    if (!closest.isConnected() ||
-      Blockly.Connection.lastConnectionInRow(this.topBlock_,
-          closest.targetConnection.getSourceBlock())) {
+    if (closest &&
+        local.getSourceBlock()
+            .shouldInsertDraggedBlock(this.topBlock_, closest)) {
```
rachel-fenichel: What if we just gave the insertion marker manager a pointer to the renderer?  Passing these all the way through the block svg seems wrong.

samelhusseini: Yep, that is indeed a round about way of getting there. I don't like this either. 
I just realized there was already a reference to the workspace on the insertion marker manager. Will up to go directly to the renderer.

core/renderers/common/path_object.js
```diff
@@ -265,3 +265,16 @@ Blockly.blockRendering.PathObject.prototype.updateReplacementHighlight =
...
+
+/**
+ * Add or remove styling that shows that if the dragging block is dropped, this
+ * block will be connected to the input.
+ * @param {Blockly.Connection} _conn The connection on the input to highlight.
+ * @param {boolean} _enable True if styling should be added.
+ * @package
+ */
+Blockly.blockRendering.PathObject.prototype.updateShapeForInputHighlight =
+    function(_conn, _enable) {
```
alschmiedt: nit: for consistency should this be _add ?

samelhusseini: Yep, thanks!

samelhusseini: Actually all the others use enable so _enable would be being consistent.

PR #3446, opened 2019-11-15
CSS in renderers

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

CSS in renderers.

### Proposed Changes

Have each renderer declare some CSS that is specific to the renderer.
Renderers can override to change the CSS declared. The declared CSS is prefixed with the renderer selector to ensure it only applies to that renderer if multiple different renderers are on the same page. 

Make blocklyText 12 pt and bold in Zelos.

### Reason for Changes

Renderers need a way to define their own CSS.

### Test Coverage

Tested geras and zelos in playground, and in multi-playground.

Tested on:


samelhusseini:
@rachel-fenichel thoughts on this design?

core/renderers/common/renderer.js
```diff
@@ -139,6 +148,64 @@ Blockly.blockRendering.Renderer.prototype.getConstants = function() {
...
+ * @protected
+ */
+Blockly.blockRendering.Renderer.prototype.getCSS_ = function() {
+  var selector = '.' + this.name_ + '-renderer';
+  return [
+    /* eslint-disable indent */
+    selector + ' .blocklyText {',
+      'cursor: default;',
+      'fill: #fff;',
+      'font-family: sans-serif;',
```
rachel-fenichel: This can now use the constants for font-family and size.

rachel-fenichel:
Yeah this looks right.

core/renderers/common/renderer.js
samelhusseini: True!!

samelhusseini: Fixed.

samelhusseini:
@rachel-fenichel ready for another review.

core/renderers/common/renderer.js
```diff
@@ -55,6 +63,7 @@ Blockly.blockRendering.Renderer = function() {
 Blockly.blockRendering.Renderer.prototype.init = function() {
   this.constants_ = this.makeConstants_();
   this.constants_.init();
+  this.injectCSS_(this.getCSS_(this.constants_));
```
rachel-fenichel: You don't need to pass in the constants object since the function is on the renderer and can access it directly.

samelhusseini: This was so ``this.constants_`` can remain private, making it protected would also resolve this. It just depends on how we see the constants object being used by subclass renderers.I see everything seems to use getConstants. so perhaps that's the solution. Have getCSS_ use ``getConstants``.

samelhusseini: Done.

PR #3465, opened 2019-11-22
[Zelos] apply Colour

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes
Only apply colour on a path object from a single entry point. Avoid passing extra parameters to applyColour from more than one entry point, instead call applyColour after setting disabled.

Pass in parent for zelos applyColour. Shadow blocks take on parent colour. We might want to consider passing in the entire block object.

### Reason for Changes

Zelos rendering.

### Test Coverage

Tested geras and zelos rendering.

Tested on:

### Additional Information

Screen Shot 2019-11-22 at 1 19 34 PM


core/block_svg.js
```diff
@@ -975,7 +977,7 @@ Blockly.BlockSvg.prototype.applyColour = function() {
     return;
   }
 
-  this.pathObject.applyColour(this.isShadow());
+  this.pathObject.applyColour(this.isShadow(), this.getParent());
```
rachel-fenichel: How does this pass type checks?  Only some versions of `applyColour` take two arguments now.

samelhusseini:
@rachel-fenichel I've updated this PR to make disabled go through an ``applyColour`` pass. I didn't modify selected and highlighted at this point, as we would need to add state on the block for when these are changed.
Let me know your thoughts on this.

core/renderers/common/i_path_object.js
```diff
@@ -90,14 +90,6 @@ Blockly.blockRendering.IPathObject.prototype.setMarkerSvg;
  */
 Blockly.blockRendering.IPathObject.prototype.updateHighlighted;
 
-/**
- * Set whether the block shows a disable pattern or not.
- * @param {boolean} disabled True if disabled.
- * @param {boolean} isShadow True if the block is a shadow block.
- * @package
```
rachel-fenichel: I think this should still be here, just marked protected.

samelhusseini: The reason I removed this is because I dont think it is a must that each renderer define these methods. For example, a renderer might choose to override the ``applyColours`` method and use it's own internal set of methods to determine how colours are applied, or not use helpers at all. It's not important for the public interface to require these "helper" methods to be present / implemented.

samelhusseini:
I agree on moving towards maintaining state for selected, and highlighted.

core/renderers/common/i_path_object.js
rachel-fenichel: I'm torn, because the base `applyColours` is calling them.  But I see your point for a renderer that overrides `applyColours` entirely.  So lgtm, approval.

PR #3468, opened 2019-11-22
Fixed adding a variable field programmatically when it has a following field.

### LINKED ISSUE #3458: Blocks with a field following a variable field break when a mutator is open
**Describe the bug**

In blocks that have mutators that create and delete inputs (dummy or value), if those blocks contain a variable field followed by another field break when the mutator is opened.

**To Reproduce**

1. Create a mutating block with a field layout described in above:
```
Blockly.Blocks['mutator_variable_test'] = {
    init: function () {
        this.itemCount_ = 2;
        this.appendDummyInput().appendField('mutator test block');
        this.updateShape_();
        this.setMutator(new Blockly.Mutator(['mutator_variable_test_item']));
    },
    mutationToDom: function () {
        var container = Blockly.utils.xml.createElement('mutation');
        container.setAttribute('items', this.itemCount_);
        return container;
    },
    domToMutation: function (xmlElement) {
        this.itemCount_ = parseInt(xmlElement.getAttribute('items'), 10);
        this.updateShape_();
    },
    decompose: function (workspace) {
        var containerBlock = workspace.newBlock('mutator_variable_test_container');
        containerBlock.initSvg();
        var connection = containerBlock.getInput('STACK').connection;
        for (var i = 0; i < this.itemCount_; i++) {
            var itemBlock = workspace.newBlock('mutator_variable_test_item');
            itemBlock.initSvg();
            connection.connect(itemBlock.previousConnection);
            connection = itemBlock.nextConnection;
        }
        return containerBlock;
    },
    compose: function (containerBlock) {
        var itemBlock = containerBlock.getInputTargetBlock('STACK');
        // Count number of inputs.
        var connections = [];
        while (itemBlock) {
            connections.push(itemBlock.valueConnection_);
            itemBlock = itemBlock.nextConnection &&
                itemBlock.nextConnection.targetBlock();
        }
        this.itemCount_ = connections.length;
        this.updateShape_();
    },
    updateShape_: function () {
        // Add new inputs.
        for (var i = 0; i < this.itemCount_; i++) {
            if (!this.getInput('ADD' + i)) {
                this.appendValueInput('ADD' + i)
                    .appendField('variable')
                    .appendField(new Blockly.FieldVariable('item'), 'VAR' + i)  // <-- THIS IS THE SOURCE OF THE BUG WHEN FOLLOWED BY ANOTHER FIELD.
                    .appendField('that stores stuff');
            }
        }
        // Remove deleted inputs.
        while (this.getInput('ADD' + i)) {
            this.removeInput('ADD' + i);
            i++;
        }
    }
};

Blockly.Blocks['mutator_variable_test_container'] = {
    init: function () {
        this.appendDummyInput().appendField('add things');
        this.appendStatementInput('STACK');
    }
};

Blockly.Blocks['mutator_variable_test_item'] = {
    init: function () {
        this.appendDummyInput().appendField('thing');
        this.setNextStatement(true);
    }
};
```
2. Click on the mutator icon
3. Try mutating the block
image

**Expected behavior**

The block should render correctly and the fields should be editable even when the mutator is open.

 - OS: MacOS 10.13.2, CrOS
 - Browser: Chrome 78.0.3904.97, Firefox 70.0.1, Safari 11.0.2

**Stack Traces**

```
field.js:567 Uncaught TypeError: Cannot set property 'nodeValue' of undefined
    at Blockly.FieldLabel.Blockly.Field.render_ (field.js:567)
    at Blockly.FieldLabel.Blockly.Field.getSize (field.js:611)
    at new Blockly.blockRendering.Field (row_elements.js:101)
    at Blockly.minimalist.RenderInfo.Blockly.blockRendering.RenderInfo.createRows_ (info.js:235)
    at Blockly.minimalist.RenderInfo.Blockly.blockRendering.RenderInfo.measure (info.js:188)
    at rendererCtor.Blockly.blockRendering.Renderer.render (renderer.js:150)
    at Blockly.BlockSvg.render (block_svg.js:1616)
    at Blockly.FieldVariable.Blockly.Field.forceRerender (field.js:712)
    at Blockly.FieldVariable.Blockly.Field.setValue (field.js:762)
    at Blockly.FieldVariable.initModel (field_variable.js:155)
```

**Additional context**

We noticed this error within our implementation (solo.parallax.com), and tried replacing the blockly core with the 2019 Q3 patch 4 release, which still showed the same error.  I then built the most minimal mutating block that still demonstrated the error in the playground (release 2019 Q2) to verify that it wasn't caused by something within our codebase.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #3458 

### Proposed Changes

1) Changes the mutator to no longer "disable" rendering.
2) Changes the variable field to call doValueUpdate_ directly instead of calling setValue.

### Reason for Changes

So the problem wasn't exactly what I thought it was. The problem was that mutators set the block's `rendered` [property to false](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/mutator.js#L351), which meant the [fields were not getting initialized](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/input.js#L116) when they were appended. Then mutator would then [call initSvg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/mutator.js#L357), which would eventually call the variable field's initMethod, which would force a rerender. This rerender would cause an error to be thrown because the renderer would try to render the fields following the variable field, which had not been initialized.

So 1) Makes it so that fields are properly initialized when they are appended during `compose`.
And 2) Makes it so variable fields no longer cause unnecessary render calls.

### Test Coverage

Could not reproduce #3458.

Tested on:
* Desktop Chrome

### Documentation

Nada

### Additional Information

Whoever reviews this, please check my thinking! I don't think changing mutators like this will break things, but I would really appreciate a second opinion.

Also @MatzElectronics

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -147,11 +147,8 @@ Blockly.FieldVariable.prototype.initModel = function() {
       this.sourceBlock_.workspace, null,
       this.defaultVariableName, this.defaultType_);
 
-  // Don't fire a change event for this setValue.  It would have null as the
-  // old value, which is not valid.
-  Blockly.Events.disable();
-  this.setValue(variable.getId());
-  Blockly.Events.enable();
+  // Don't call setValue because we don't want to cause a rerender.
+  this.doValueUpdate_(variable.getId());
 };
 
 /**
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -374,6 +374,7 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
     var oldMutation = oldMutationDom && Blockly.Xml.domToText(oldMutationDom);
     // Allow the source block to rebuild itself.
     block.compose(this.rootBlock_);
+    block.initSvg();
     block.render();
     var newMutationDom = block.mutationToDom();
     var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -372,15 +372,9 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
     var block = this.block_;
     var oldMutationDom = block.mutationToDom();
     var oldMutation = oldMutationDom && Blockly.Xml.domToText(oldMutationDom);
-    // Switch off rendering while the source block is rebuilt.
-    var savedRendered = block.rendered;
-    block.rendered = false;
     // Allow the source block to rebuild itself.
     block.compose(this.rootBlock_);
-    // Restore rendering and show the changes.
-    block.rendered = savedRendered;
-    // Mutation may have added some elements that need initializing.
-    block.initSvg();
+    block.render();
     var newMutationDom = block.mutationToDom();
     var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
     if (oldMutation != newMutation) {
@@ -393,9 +387,6 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
         block.bumpNeighbours();
         Blockly.Events.setGroup(false);
       }, Blockly.BUMP_DELAY);
-    }
-    if (block.rendered) {
-      block.render();
     }
 
     if (oldMutation != newMutation &&


samelhusseini:
I would go with the ``doValueUpdate_`` solution, it's easier to reason about (skip setValue and directly update the internal value during initialization). 
Refactoring the mutator code to re-render instead of initialize + render appears to work, but I'm concerned there are edge cases where it breaks. I couldn't trace back to why that code was added initially, but just going by the comments, it seems it's there to initialize any icons / fields that are added as part of the mutation: 
```
// Mutation may have added some elements that need initializing.
```

BeksOmega:
### To init or Not to init
>It seems it's there to initialize any icons / fields that are added as part of the mutation.

So I did some looking into it, and I believe it is safe to get rid of the init. There are two reasons for this:

#### 1) Init Happens Immediately

All fields and icons initialize and render themselves immediately upon being appended, with the exception of the mutator icon.

* Comment ([creation](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1156), [initSvg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/comment.js#L64), [render](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1164)).
* Warning ([creation](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1227), [initSvg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/warning.js#L42), [render](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1247)).
* Mutator ([creation](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1257), [initSvg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/block_svg.js#L1264), no render :/) <- Bug! but not a problem in this case.
* Field ([creation](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/input.js#L101), [initSvg](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/input.js#L117), [render](https://github.com/google/blockly/blob/ba6dfd812552e0823ed75f3dc5133a267cc6fbc9/core/input.js#L134))

This shows us that not only can we get rid of the `block.initSvg` call, but we can also get rid of the `block.render` call. It will already have been called several times over.

This makes sense if you think about:

#### 2) If We Need to Call `initSvg` Or `render` That's a Bug

Every call inside a mutator `compose` method should also work when it is *not* inside a `compose` method.

If you add a field/icon programmatically it should always intialize it's svg, and rerender the block. If it doesn't then you end up with a broken block, and that sucks.

So really if we do call `initSvg` and `render` after `compose` we're just going to be hiding bugs.

### Next Steps

What I would like to do:
1) Add initSvg back to this PR for safety. Get this PR merged.
2) Create issues addressing the above and the below.
3) I or other people can take them from there.

Also of note: I'd like to get some tests in that document how fields and icons are supposed to initialize themselves immediately.

How does that plan sound to you @samelhusseini ?

### A Note About the `rendered` Property

I think it would be nice to do a two things:

1) Better document what `rendered` means. Does it just mean that we're not in headless mode? Does it mean that the block/workspace is currently visible? How does this change `rendered` checks that are currently in the code base? Can some of them be removed/changed?
2) Maybe add some sort of batch rendering functionality? So that if we are for example composing the block, we can wait to re-render until all of the UI elements have been added. There's obviously one use case for it, but idk if there are enough other use cases to make it worth it.

samelhusseini:
> Add initSvg back to this PR for safety. Get this PR merged.
Create issues addressing the above and the below.
I or other people can take them from there.

@BeksOmega that plan sounds good.

PR #3469, opened 2019-11-22
Support adjusting dropdowndiv default colours through CSS

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Support setting the dropdown theme default colours through CSS.
Update dark theme to adjust dropdown and widget div colours. 

### Test Coverage

Multiple themes in playground.

Tested on:

### Additional Information

Screen Shot 2019-11-22 at 3 47 28 PM
Screen Shot 2019-11-22 at 3 47 25 PM

### DIFF
--- a/core/theme/dark.js
+++ b/core/theme/dark.js
@@ -130,6 +130,9 @@ Blockly.Themes.Dark.setComponentStyle('scrollbarOpacity', 0.4);
 
 /**
  * CSS for the dark theme.
+ * This registers CSS that is specific to this theme. It does so by prepending a
+ * ``.dark-theme`` selector before every CSS rule that we wish to override by
+ * this theme.
  */
 (function() {
   var selector = '.dark-theme';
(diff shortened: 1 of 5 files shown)


core/theme/dark.js
```diff
@@ -131,10 +131,32 @@ Blockly.Themes.Dark.setComponentStyle('scrollbarOpacity', 0.4);
 /**
  * CSS for the dark theme.
```
alschmiedt: Since this is our main example currently for how to use themes could we update the comment to explain a bit more about what this function is doing?

alschmiedt:
It looks like the dropdown doesn't get updated when we switch between themes. It only gets updated once a user opens and closes the dropdown. Can we re draw on theme change?

samelhusseini:
Maybe we can hideChaff on Theme change?

core/theme/dark.js
samelhusseini: Done. Good idea!

Issue #3475, opened 2019-11-26
JetBrains artifacts aren't gitignored


dwesty17:
**Describe the bug**

The `.idea` directory (JetBrains artifacts) are monitored by git causing them to be included in diffs.

**To Reproduce**

Try to make a commit using IntelliJ/WebStorm

**Expected behavior**

I want that directory to be ignored.

dwesty17:
Here's a PR: https://github.com/google/blockly/pull/3476

dwesty17:
Closing per convo on PR

PR #3476, opened 2019-11-26
gitignore the .idea directory which contains jetbrains artifacts

dwesty17:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3475

### Proposed Changes

Add `.idea` to `.gitignore`

### Reason for Changes

So that the `.idea` directory isn't accidentally committed/to make it easier for people who just JetBrains products to contribute to the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ build-debug.log
 .DS_Store
 .settings
 .project
+.idea
 *.gz
 *.pyc
 *.komodoproject


dwesty17:
@googlebot I signed it!

BeksOmega:
I think this might fit better in your personal git exclude, since not everyone uses jetbrains IDE's. [Info about git excludes](https://help.github.com/en/github/using-git/ignoring-files#explicit-repository-excludes).

This is how I manage my .idea files and I think it works great. Do you think it could work for your situation?
--Beka

dwesty17:
Cool! I actually didn't know that was a thing. This works for me.

I think it probably still makes sense to have given that it's a pretty popular IDE, in the same way that not everyone uses Macs, but `.DS_Store` is in most `.gitignore` files. Reduces friction for getting started ever so slightly.

Thanks for the info about `git excludes` though! I'll go ahead and close this out.

NeilFraser:
I didn't know that was a thing either.  Which is why `*.komodoproject` is in Blockly's .gitignore (I'm a Komodo user).

dwesty17:
Got it 🙂

PR #3484, opened 2019-12-04
Refactor marker and cursor

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Marker should be the parent of the cursor.

### Proposed Changes
1. Move marker to be parent of a cursor
2. Add a type and colour to a marker
3. Pass the marker to the drawer so it can use the colour and the type to figure out what to draw.

### Test Coverage

Tested on:


core/renderers/common/marker_svg.js
```diff
@@ -16,113 +16,117 @@
...
+  /**
+   * The colour of the marker.
+   * @type {string}
+   */
+  this.colour_ = marker.colour || defaultColour;
 };
 
 /**
- * The name of the CSS class for a cursor.
+ * The name of the CSS class for a marker.
```
samelhusseini: This should probably stay as cursor.

Issue #3414, opened 2019-11-07
Decide what to do about output shape on the block


samelhusseini:
Zelos blocks sometimes set the output shape in JS (by calling block.setOutputShape). We should decide what to do about those blocks that want to specify the output shape explicitly and not use the ``shapeFor`` method to dynamically decide what shape to use.

rachel-fenichel:
The task here is to add `setOutputShape` as a function on a block

samelhusseini:
Fixed.

PR #3485, opened 2019-12-04
Move blockStyles into constants

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Zelos uses darker shades for both secondary / tertiary colours. This is in conflict with the defaults in Blockly. We need to provide a way to extend that.

### Proposed Changes

Refactor theme object to add protected generators for secondary and tertiary colours moving the validate method from a static method to an instance method.
Pass in style object when creating the path object, rather than re-creating it in the path object.
Add classic and modern zelos themes, that use ``Blockly.zelos.Theme`` which overrides the secondary/tertiary generator methods and uses dark derivatives of the primary colour for both:
- Classic, a duplicate of ``Blockly.Themes.Classic`` but uses ``Blockly.zelos.Theme`` instead.
- Modern, a new theme that uses ``Blockly.zelos.Theme`` with the MakeCode / Scratch colours for blocks (Not required, good for demos, happy to leave out).

### Reason for Changes

Zelos rendering / theming.

### Test Coverage

Tested zelos renderering playground.
Ran mocha tests.

Tested on:

### Additional Information

Modern Zelos theme sample:
Screen Shot 2019-12-04 at 1 44 34 PM


core/renderers/common/path_object.js
```diff
@@ -60,10 +62,10 @@ Blockly.blockRendering.PathObject = function(root, constants) {
 
   /**
    * The style object to use when colouring block paths.
-   * @type {!Blockly.Theme.BlockStyle}
+   * @type {Blockly.Theme.BlockStyle}
```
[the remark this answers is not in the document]
samelhusseini: Yeah my mistake.

core/theme/zelos/theme.js
```diff
@@ -0,0 +1,62 @@
...
+  Blockly.zelos.Theme.superClass_.constructor.call(this, name,
+      blockStyles, categoryStyles, opt_componentStyles);
+};
+Blockly.utils.object.inherits(Blockly.zelos.Theme, Blockly.Theme);
+
+
+/**
+ * @override
+ */
+Blockly.Theme.prototype.generateSecondaryColour_ = function(colour) {
```
rachel-fenichel: These should be on the `zelos.Theme` prototype, not on `Blockly.Theme.prototype`.

tests/rendering/zelos/zelos.html
```diff
@@ -32,7 +32,11 @@
   <script>
     goog.require('Blockly.blockRendering.Debug');
     goog.require('Blockly.zelos.Renderer');
+    goog.require('Blockly.zelos.Themes.Classic');
     // Blockly.blockRendering.startDebugger();
+
+    Blockly.BlockSvg.prototype.setOutputShape = function() { };
```
rachel-fenichel: ?

core/theme/zelos/classic.js
```diff
@@ -0,0 +1,99 @@
...
+    "colour": "330"
+  },
+  "variable_dynamic_category": {
+    "colour": "310"
+  }
+};
+
+Blockly.zelos.Themes.Classic =
+    new Blockly.zelos.Theme('zelosclassic',
+        Blockly.zelos.Themes.Classic.defaultBlockStyles,
```
rachel-fenichel: Instead of having all of the block and category styles above, can you do
`new Blockly.zelos.Theme('zelosclassic', Blockly.Themes.Classic.defaultBlockStyles, Blockly.Themes.Classic.categoryStyles)`?

samelhusseini: No because Blockly.Themes.Classic.defaultBlockStyles is on a temporary holding object that gets replaced with the Blockly.Theme object when created. We could instead do ``Blockly.Themes.Classic.getAllBlockStyles()`` but that would have passed through the previous theme's validation methods which would have already generated the secondary / tertiary colours..

tests/rendering/zelos/zelos.html
samelhusseini: There as a stub for pxt-blockly blocks. We might want to consider allowing blocks to choose their output shape (see https://github.com/google/blockly/issues/3414) but for now a stub will get around that for testing.

core/theme/zelos/theme.js
samelhusseini: Fixed.

samelhusseini:
After speaking to @rachel-fenichel offline, we've decided to move block Styles and the associated validation into constants.

tests/rendering/zelos/zelos.html
samelhusseini: Added comment.

PR #3492, opened 2019-12-06
Fix full text input fields with label fields

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes blocks that have a text input field and only label fields.

### Proposed Changes

Count all fields when deciding whether or not a full text field should be used.
@rachel-fenichel any idea why we weren't including label fields?

### Reason for Changes

Before: 
Screen Shot 2019-12-06 at 10 59 17 AM

After: 
Screen Shot 2019-12-06 at 10 59 52 AM

### Test Coverage

Tested test blocks playground.

Tested on:


rachel-fenichel:
Not sure why we weren't including labels--I believe I copied that from Scratch but I don't remember why it was there.

Issue #3494, opened 2019-12-07
[zelos rendering] Multi-row reporters


samelhusseini:
Implement multi-row reporters in zelos rendering like below:

![Screen Shot 2019-12-06 at 4 46 01 PM](https://user-images.githubusercontent.com/16690124/70365758-eb70bc00-1847-11ea-8916-7c503252b9b8.png)

samelhusseini:
Fixed.

Issue #2978, opened 2019-09-06
[zelos] Determine what to do with inline inputs.


samelhusseini:
**Is your feature request related to a problem? Please describe.**

Geras rendering supports both inline inputs and external inputs. 
We need to determine what to do about the inlineInputs flag in zelos rendering.

**Describe the solution you'd like**

We can think of inline inputs as a way to determine whether or not to lay the input horizontally or vertically. Even though geras displays the input on the outside of the block when displayed vertically we can use the same flag in zelos to rendering inputs on multiple lines (vertically) when setInputsInline is false, alternately we lay them out horizontally. In either case the input is inside the block, and not external like geras.

For example:

"Inline"
<img width="637" alt="Screen Shot 2019-09-06 at 9 38 53 AM" src="https://user-images.githubusercontent.com/16690124/64444951-29641580-d08a-11e9-80fa-8d0fda26cd3b.png">

"External"
<img width="416" alt="Screen Shot 2019-09-06 at 9 37 58 AM" src="https://user-images.githubusercontent.com/16690124/64444930-18b39f80-d08a-11e9-9d69-db288cac3813.png">

**Additional context**

We should also think about how we want to lay out multi-line inputs. Left / right alignment will also take play in how it renders.

samelhusseini:
Done.

PR #3501, opened 2019-12-10
[zelos] Multi-line rendering

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3494
Fixes https://github.com/google/blockly/issues/2978

### Proposed Changes

Add a new ``isMultiLine`` property on the info object, and based on that decide whether or not to create a new row.

Size the negative spacing for blocks with output connections based on the radius created by the height of the block in order to maintain a ``SMALL_PADDING`` padding around.

We could use the isInline property, we would just need override code to instead render inline inputs instead of external inputs all over, both in the info object and drawer. I thought it would be cleaner to create a new property.

### Reason for Changes

Zelos rendering

### Test Coverage

Tested test blocks in playground.

Tested on:

### Additional Information

Screen Shot 2019-12-10 at 10 30 04 AM

Screen Shot 2019-12-10 at 10 34 01 AM


core/renderers/common/info.js
```diff
@@ -138,6 +138,13 @@ Blockly.blockRendering.RenderInfo = function(renderer, block) {
    */
   this.rows = [];
 
+  /**
+   * The number of active rows on a block.
```
rachel-fenichel: Update the comment to say that while creating rows, it' the currently active one, and after creating rows it's the total number of rows on the block.

samelhusseini: I think instead I'm going to call it inputRowCount as I'm only interested in the number of input rows.

samelhusseini: I'm not sure about that comment. It's not the currently active one, it's not even the index of the row of the active row. It's the current count of input rows created.

samelhusseini: How does ``The total number of input rows added onto the block.`` sound?

rachel-fenichel: Sounds good. I thought you were using the value to determine the negative padding for each row as you went, but you're only using whether it's greater than 1.

PR #3516, opened 2019-12-13
[zelos] Output connection position

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Zelos connection is currently positioned in the top left corner. Because the connection shape is circular / triangular, this is positioned off the block.  

### Proposed Changes

Position the connection point to the left and center of the block: 
Screen Shot 2019-12-12 at 7 23 44 PM

Added methods on the shape for the shape developer to decide where to position it. 
@rachel-fenichel just a thought, but maybe we should make all shapes dynamic and simplify the code, in the cases where the height / width are static they would just return the static height / width not taking into account the block height that's passed in.

### Reason for Changes

Zelos rendering.

### Test Coverage

Tested RTL.

Tested on:

### DIFF
--- a/core/renderers/common/drawer.js
+++ b/core/renderers/common/drawer.js
@@ -396,7 +396,8 @@ Blockly.blockRendering.Drawer.prototype.positionInlineInputConnection_ = functio
     if (this.info_.RTL) {
       connX *= -1;
     }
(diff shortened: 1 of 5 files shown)


rachel-fenichel:
LGTM

And yes, making everything dynamic sounds like a good idea.

PR #3525, opened 2019-12-24
test: Added connection test blocks.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Adds blocks that test the connection system (typed connections, "missing" connections, and non-movable blocks).

I also moved the field blocks to the bottom of the test toolbox.

Toolbox
Row
Stack
Statement

The "insert" buttons insert the set of blocks shown to the left (for each respective screen shot).

And as the toolbox suggests all of the blocks have tooltips to explain how they work. But simply:
* Greys:
  * Have permiscuous (untyped) connections so they can connect to anything.
* Blues:
  * Top/Left: None.
  * Bottom/Right: Only yellow.
* Yellows:
  * Top/Left: Blue, yellow.
  * Bottom/Right: Yellow, red.
* Reds:
  * Top/Left: Only yellow.
  * Bottom/Right: None.

### Reason for Changes

I've been using these for a while, and I think they're useful for testing!

### Test Coverage

This is tests silly!

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/tests/playground.html
+++ b/tests/playground.html
@@ -1443,6 +1443,11 @@ var spaghettiXml = [
   the standard predefined blocks, and so test alternative block rendering
   code paths. -->
   <xml xmlns="https://developers.google.com/blockly/xml" id="toolbox-test-blocks" style="display: none">
+    <category name="Align">
+      <block type="test_align_dummy_right"></block>
+      <block type="test_align_all"></block>
+      <block type="test_align_with_external_input"></block>
+    </category>
     <category name="Basic">
       <block type="test_basic_empty"></block>
       <block type="test_basic_empty_with_mutator"></block>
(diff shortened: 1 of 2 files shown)


tests/playground.html
```diff
@@ -1460,7 +1542,45 @@ <h1>Blockly Playground</h1>
         </statement>
       </block>
     </category>
-    <category name="Fields" expanded="true">
+    <category name="Mutators">
```
samelhusseini: I think these (Mutator and Style categories) are dups.

BeksOmega: I'm not sure what you mean, so I'll just try to cover everything.

I checked and the categories are only showing up once:
Categories
so that's rad.

But if you're wondering if the categories are no longer necessary, I'm not sure. You'd have to ask @alschmiedt

BeksOmega: I'm not sure what you mean so I'll just try to cover everything.

I checked and the categories are only showing up once:
Categories
so that's rad.

But if you're wondering if the categories are no longer necessary, I'm not sure. You'd have to ask @alschmiedt

samelhusseini: I see what's going on. They're just being moved around. Mutators and Style categories were after Fields.
There was probably an intention to keep these alphabetically sorted, but I messed that up by adding Align second.
We could fix that here?

BeksOmega:
Categories now look like this:
NewCategories

tests/playground.html
```diff
@@ -1460,7 +1610,7 @@ <h1>Blockly Playground</h1>
         </statement>
       </block>
     </category>
-    <category name="Fields" expanded="true">
+    <category name="Fields">
```
samelhusseini: I think you need to leave ``expanded="true"`` to make sure it's expanded by default.

tests/playground.html
```diff
@@ -1374,10 +1461,73 @@ <h1>Blockly Playground</h1>
       <block type="test_basic_tooltips"></block>
       <block type="test_basic_javascript"></block>
     </category>
-    <category name="Align">
-      <block type="test_align_dummy_right"></block>
-      <block type="test_align_all"></block>
-      <block type="test_align_with_external_input"></block>
+    <category name="Connections">
```
samelhusseini: Add ``expanded="true"`` to make sure it's expanded by default.

samelhusseini:
Thanks @BeksOmega!

PR #3527, opened 2019-12-27
feat: Improved procedure arg interaction.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Changes the prevent-duplicate-args behavior so that instead of updating the arg when the block is placed, the flyout is updated instead.

I also changed it so that it starts at whatever character you want, instead of always at 'i'.

New behavior:
ProcedureArgs
Old behavior:
ProcedureArgsOld

### Reason for Changes

... Mostly I was just bored hehe. But I do think there are a couple of nice things about the new behavior.

* The arg value changes in the flyout. This makes the behavior more clear.
* The args go in alphabetical order, instead of jumping to i. I think that's more pleasant.
* In the second behavior gif, I drag out the second arg and its value turns to "j" instead of "i". The new behavior fixes that, which I think is nice.

### Test Coverage

I don't think this really needs unit tests. It's basically just two event listeners and a unique string generator, so it should be pretty unbreakable. But if people have test recommendations I'd be happy to write them!

In addition to the behavior demo above, I'd like to show that this works after the mutator has been closed and reopened:
ProcedureArgs2

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

If you guys don't like this behavior, that's cool. I was just bored and wanted something to tinker with.

Also I'll be out of town for the next week, so I may not be super quick to respond.

### DIFF
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -72,6 +72,10 @@ Blockly.Mutator.prototype.setBlock = function(block) {
   this.block_ = block;
 };
 
+Blockly.Mutator.prototype.getWorkspace = function() {
+  return this.workspace_;
+};
+
 /**
  * Draw the mutator icon.
  * @param {!Element} group The icon group.
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -72,6 +72,11 @@ Blockly.Mutator.prototype.setBlock = function(block) {
   this.block_ = block;
 };
 
+/**
+ * Returns the workspace inside this mutator icon's bubble.
+ * @return {Blockly.WorkspaceSvg} The workspace inside this mutator icon's
+ *     bubble.
+ */
 Blockly.Mutator.prototype.getWorkspace = function() {
   return this.workspace_;
 };
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -46,6 +46,10 @@ goog.require('Blockly.Xml');
  */
 Blockly.Procedures.NAME_TYPE = Blockly.PROCEDURE_CATEGORY_NAME;
 
+/**
+ * The default argument for a procedures_mutatorarg block.
+ * @type {string}
+ */
 Blockly.Procedures.DEFAULT_ARG = 'x';
 
 /**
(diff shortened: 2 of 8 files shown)


core/variables.js
```diff
@@ -67,11 +67,11 @@ Blockly.Variables.allUsedVarModels = function(ws) {
     }
   }
   // Flatten the hash into a list.
-  var variableList = [];
+  var VariableList = [];
```
mattbishop: Curious, why the capitalization of `variableList`?

BeksOmega: Cuz typos are my best friends :P hehe had some accidental renamings and looks like I didn't quite fix this one. Thanks for the catch!

rachel-fenichel:
+1 to this approach.

core/procedures.js
```diff
@@ -271,6 +277,77 @@ Blockly.Procedures.flyoutCategory = function(workspace) {
...
+  var workspace = block.mutator.getWorkspace();
+  Blockly.Procedures.updateMutatorFlyout(workspace);
+  workspace.addChangeListener(Blockly.Procedures.mutatorChangeListener);
+};
+
+/**
+ * Listens for changes in a procedure mutator and triggers flyout updates when
+ * necessary.
+ * @param {!Blockly.Events.Abstract} e The event that triggered this listener.
+ * @package
```
alschmiedt: @private

core/procedures.js
```diff
@@ -271,6 +277,77 @@ Blockly.Procedures.flyoutCategory = function(workspace) {
...
 };
 
+/**
+ * Updates the procedure mutator's flyout so that the arg block is not a
+ * duplicate of another arg.
+ * @param {!Blockly.Workspace} workspace The procedure mutator's workspace. This
+ *     workspace's flyout is what is being updated.
+ * @package
+ */
+Blockly.Procedures.updateMutatorFlyout = function(workspace) {
```
alschmiedt: Should be @private

core/mutator.js
```diff
@@ -72,6 +72,15 @@ Blockly.Mutator.prototype.setBlock = function(block) {
   this.block_ = block;
 };
 
+/**
+ * Returns the workspace inside this mutator icon's bubble.
+ * @return {Blockly.WorkspaceSvg} The workspace inside this mutator icon's
+ *     bubble.
+ */
```
samelhusseini: Any reason for this to be public? If not, use ``@package`` instead.

BeksOmega: I'll create a separate PR (・◡・)b

Issue #3532, opened 2020-01-02
Scrolled text input widget div is positioned incorrectly


samelhusseini:
**Describe the bug**
Scrolled text input widget div is positioned incorrectly

**To Reproduce**

1. Go to the multiplayground
2. Scroll the page
3. Drag any text input field and click on it to open it.
4. See how it's positioned incorrectly.

**Expected behavior**

Positioned over the field even when the page is scrolled.

**Screenshots**
<img width="501" alt="Screen Shot 2020-01-02 at 12 25 21 PM" src="https://user-images.githubusercontent.com/16690124/71691018-28eb3b00-2d5b-11ea-9cc3-92a5ca88c269.png">

samelhusseini:
Fixed.

PR #3533, opened 2020-01-02
Fix scrolled text input widgetdiv positioning.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3532

### Proposed Changes

Take into account page scroll with field scaled bbox calculation.

### Reason for Changes

Fix scrolled text input widgetdiv positioning.

### Test Coverage
Tested playground and multi playground.

Tested on:

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -377,16 +377,16 @@ Blockly.DropDownDiv.show = function(owner, rtl, primaryX, primaryY,
  * @private
  */
 Blockly.DropDownDiv.getBoundsInfo_ = function() {
-  var boundPosition = Blockly.DropDownDiv.boundsElement_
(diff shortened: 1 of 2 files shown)


alschmiedt:
So testing it out it looks like this causes a problem with the dropdown div. If you open the dropdown in any of the workspaces other than the top one then it will always render above the field even if there is room below.

samelhusseini:
Good find, but I think that's a separate issue with the dropdown div's bounding box. Currently in develop this is what it looks like in multi_playground:
Screen Shot 2020-01-02 at 2 39 16 PM

I'll try to fix within this same PR.

alschmiedt:
lgtm once the compile error is fixed

PR #3540, opened 2020-01-03
[zelos] Left/right aligned dummy inputs

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Deal with the case where we have two dummy inputs on the same row, one with left alignment and the other with right.

Zelos rendering specific fix. We keep track of the right aligned dummy input and add the missing space when adding alignment padding.

### Reason for Changes

Zelos rendering.

### Test Coverage

Tested with pxt-blockly's controls_if block.

Screen Shot 2020-01-03 at 1 28 51 PM
Screen Shot 2020-01-03 at 1 28 38 PM

Tested on:

### DIFF
--- a/core/renderers/zelos/info.js
+++ b/core/renderers/zelos/info.js
@@ -233,6 +233,47 @@ Blockly.zelos.RenderInfo.prototype.getElemCenterline_ = function(row, elem) {
 };
 
 /**
+ * @override
+ */
+Blockly.zelos.RenderInfo.prototype.addInput_ = function(input, activeRow) {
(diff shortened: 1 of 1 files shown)


alschmiedt:
Just curious, how does this work if we have more than two dummy inputs? For example, if we have Left aligned/right aligned/left aligned/right aligned?

samelhusseini:
It's only supposed to support that last input, so in the case that it's LRLR, it would add the spacing between the last LR dummy inputs, otherwise if it's something like LCR, it would default to using the first dummy input's alignment for the entire row, ie: L aligned in that case.

PR #3541, opened 2020-01-03
[zelos] Bump / tight nest for all rows that are preceded and followed by a notch

samelhusseini:
## The basics

- [x] I branched from develop

### Proposed Changes

Instead of only applying the bump X logic (for bumping inputs past the notch horizontally), apply this logic for every row that has a notch before it and after it (with the exception of the first row where we always apply the bump).
When bumping horizontally, unless we're bumping the first row, use the prev statement row's left width to determine the full extent to which to bump the next row, ie: making sure we always bump past the notch on every row.
We use this same logic to determine whether we should tightly nest vertically.

Add small padding after a notch so the notch doesn't clash with inputs / fields.

### Reason for Changes

Zelos rendering.

### Test Coverage

Screen Shot 2020-01-02 at 2 30 31 PM

Screen Shot 2020-01-03 at 2 06 28 PM

Tested in zelos rendering playground.

Tested on:

### DIFF
--- a/core/renderers/zelos/info.js
+++ b/core/renderers/zelos/info.js
@@ -247,7 +247,7 @@ Blockly.zelos.RenderInfo.prototype.adjustXPosition_ = function() {
     var row = this.rows[i];
     var nextSpacer = this.rows[i + 1];
     var hasPrevNotch = i == 2 ?
-        !!this.topRow.hasPreviousConnection : prevSpacer.followsStatement;
+        !!this.topRow.hasPreviousConnection : !!prevSpacer.followsStatement;
     var hasNextNotch = i + 2 >= this.rows.length - 1 ?
         !!this.bottomRow.hasNextConnection : !!nextSpacer.precedesStatement;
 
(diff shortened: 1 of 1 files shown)


core/renderers/zelos/info.js
```diff
@@ -239,31 +239,42 @@ Blockly.zelos.RenderInfo.prototype.getElemCenterline_ = function(row, elem) {
...
-  if (!this.topRow.hasPreviousConnection) {
-    return;
-  }
-  var minXPos = this.constants_.NOTCH_OFFSET_LEFT +
+  var notchTotalWidth = this.constants_.NOTCH_OFFSET_LEFT +
       this.constants_.NOTCH_WIDTH;
-  for (var i = 0, row; (row = this.rows[i]); i++) {
-    if (Blockly.blockRendering.Types.isInputRow(row)) {
+  var minXPos = notchTotalWidth;
+  for (var i = 2; i < this.rows.length - 1; i += 2) {
```
alschmiedt: Might help readability to have a comment explaining the goal of this for loop. Specifically, might be nice to know why we are counting by 2s.

Issue #2890, opened 2019-08-20
Field date closure dependency


samelhusseini:
``field_date.js`` still has a dependency on closure. 
In the short term, we are going to exclude it from the build script as it's not included in blockly_compressed.js

There are a few options being floated around about what to do about it moving forward. Here are some ideas:
- Convert it to use an open source date picker library that isn't closure
- Once that's converted, we can move it out of core into demos.

Filing this issue to keep track of that work.

We should also update our documentation to provide instructions of how to include the date field. Now that closure is gone, there will be a few extra tasks one needs to complete.

kushthedude:
Hello @samelhusseini , I am new to the Google Open Source Community, I think I can try to fix this issue up. Please reply if it `up for grabs`

samelhusseini:
Hey @kushthedude, sure did you have a specific date picker in mind to integrate?

kushthedude:
> Hey @kushthedude, sure did you have a specific date picker in mind to integrate?

@samelhusseini, Sorry for the late reply, I was getting a hang of blockly that's why couldn't reply you. I don't have any specific date picker, But if you have thought of anything I can move ahead with it

samelhusseini:
Here are a few options: 
- Bootstrap: https://github.com/Eonasdan/bootstrap-datetimepicker
- Jquery: https://jqueryui.com/datepicker/
- React: https://github.com/Hacker0x01/react-datepicker

samelhusseini:
More details on this change:
- We want the date field to become a demo field, so we'd move it to https://github.com/google/blockly/tree/develop/demos/custom-fields (similar to turtle and pitch)
- Since it's a demo, you'd want to use CDNs to include any frameworks, as we wouldn't be able to ship any dependencies for this.
- Clean up field_date from core/

Let me know if you have any questions. Thanks for picking this up!

samelhusseini:
https://github.com/google/blockly-samples/pull/28

kushthedude:
I did completely forget about it ,sorry about that.
I had done some changes and will definitely use the given reference. A PR
will be ready in a day.

samelhusseini:
With #3554 developers have the ability to build Blockly with the closure library in order to make use of ``FieldDate``. I've also updated the FieldDate documentation to reflect the steps required for including FieldDate in their build (Those changes will be live when the release is published).

PR #3554, opened 2020-01-07
Upgrade closure + support for closure-library dependency

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Add hint for how to include FieldDate into build.
https://github.com/google/blockly/issues/2890

### Proposed Changes

Upgrade the closure compiler and the compiler-deps to ``20200101.0.0``.
Add scripting for including the closure library in a build like so: 
```
gulp build-core --closure-library
```
Add hints on how to include the FieldDate into a build (by uncommenting out the require, adding google-closure-library to package.json and running the build comment with the ``--closure-library`` flag.

### Reason for Changes

Allow users to include the closure library if they want to.

### Test Coverage

Tested building compressed and uncompressed versions.

Tested on:


alschmiedt:
lgtm. Might be nice to also add comment about this in the field date file, but probably fine either way.

Issue #3546, opened 2020-01-06
[zelos] Overlap in block height when a statement block with no next connection is attached into a statement input.


samelhusseini:
**Describe the bug**

Statement block with no next connection doesn't connect properly to the rest of the blocks. 

**To Reproduce**

1. Go to the playground
2. Drag a repeat block and a break out of loop block.
3. Drag the breakout block inside the repeat block.
4. Notice how there's an overlap in spacing on the bottom of the block.

**Expected behavior**

Expect there to be room to add the block with no overlap.

**Screenshots**

![Screen Shot 2020-01-06 at 11 30 10 AM](https://user-images.githubusercontent.com/16690124/71843206-3284e380-3078-11ea-9c07-9757f41bfa78.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**

samelhusseini:
Fixed.

PR #3562, opened 2020-01-07
[zelos] Don't add next notch in a statement input if the connection block doesn't have a next connection

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3546

### Proposed Changes

Create a custom zelos statement input that checks if the last connected block in a statement input's stack has a next connection and if not, doesn't add a next notch.

### Reason for Changes

Zelos rendering.

### Test Coverage

Tested the break out of loop block in a repeat, if and procedure statement in the zelos renderer.

Screen Shot 2020-01-07 at 2 41 37 PM

Tested on:


alschmiedt:
Found this when playing with the feature: 
Screen Shot 2020-01-08 at 7 50 41 PM
It looks like it is only the last statement input that is having the problem. I tried it in develop and it seems fine so I think it is because of this change.

samelhusseini:
Yep, that was fixed in https://github.com/google/blockly/pull/3565. I just hadn't merged that back into this PR, but can do now.

petervanderwalt:
> Screen Shot 2020-01-07 at 2 41 37 PM

@samelhusseini which theme is this?  Looks nice and flat compared to default theme 

image

BeksOmega:
Hello @petervanderwalt !

This is using the [zelos renderer](https://developers.google.com/blockly/guides/configure/web/appearance/renderers#built-in) (which controls the shape of the block) with the [modern theme](https://www.npmjs.com/package/@blockly/theme-modern) (which controls the colors of the block.

In the future, please direct questions like this to the [developer forum](https://groups.google.com/g/blockly)! We try to make sure the github is just projec-tracking stuff :P Appreciate the question though!

petervanderwalt:
Thank you, and sorry!

PR #3569, opened 2020-01-08
Give bottom row a fixed with after statement inputs

### LINKED ISSUE #3367: Statement input rendering: bar below input got longer
Blocks with statement inputs are rendered differently with the new renderer.
The horizontal bar below the statement input has the same width as the bar above it. The old renderer made the bar as short as possible. This is especially noticable in the for-loop block.

I tested with Q3 patch 3.

statementInputBar

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3367

### Proposed Changes
geras/Info.js
Override alignRowElements_ and getSpacerRowWidth_ for geras so that it can check if it is the bottom row and the bottom row has a fixed width.

geras/drawer.js
Moves drawing the horizontal 'lineOnAxis('H', row.xPos + row.width)' from drawStatementInput to drawRightSideRow.

I had to move addRowSpacing_ back to below alignRowElement because I needed the bottom rows width to get the correct width of the spacer. This was originally changed in #3463 for zelos, however zelos now overrides the measure function so it can have a different order. 

### Test Coverage

Tested on:

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -191,9 +191,9 @@ Blockly.blockRendering.RenderInfo.prototype.getRenderer = function() {
 Blockly.blockRendering.RenderInfo.prototype.measure = function() {
   this.createRows_();
   this.addElemSpacing_();
-  this.addRowSpacing_();
   this.computeBounds_();
   this.alignRowElements_();
+  this.addRowSpacing_();
   this.finalize_();
 };
 
--- a/core/renderers/geras/info.js
+++ b/core/renderers/geras/info.js
@@ -83,7 +83,7 @@ Blockly.geras.RenderInfo.prototype.populateBottomRow_ = function() {
       this.block_.inputList.length &&
       this.block_.inputList[this.block_.inputList.length - 1]
           .type == Blockly.NEXT_STATEMENT;
-  this.bottomRow.hasFixedWidth = followsStatement && this.block_.getInputsInline();
+  this.bottomRow.hasFixedWidth = !!followsStatement && this.block_.getInputsInline();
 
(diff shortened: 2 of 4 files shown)


core/renderers/common/info.js
```diff
@@ -191,9 +191,9 @@ Blockly.blockRendering.RenderInfo.prototype.getRenderer = function() {
 Blockly.blockRendering.RenderInfo.prototype.measure = function() {
   this.createRows_();
   this.addElemSpacing_();
-  this.addRowSpacing_();
   this.computeBounds_();
   this.alignRowElements_();
+  this.addRowSpacing_();
```
samelhusseini: This was moved up deliberately before computeBounds, what's the reasoning to move it back down?

core/renderers/geras/info.js
```diff
@@ -339,6 +340,41 @@ Blockly.geras.RenderInfo.prototype.getSpacerRowHeight_ = function(prev, next) {
...
+  }
+};
+
+/**
+ * @override
+ */
+Blockly.geras.RenderInfo.prototype.getSpacerRowWidth_ = function(prev, next) {
+  // The width of the spacer before the bottom row should be the same as the
+  // bottom row.
+  if (Blockly.blockRendering.Types.isBottomRow(next) && next.hasFixedWidth) {
```
samelhusseini: I feel like the reason we need to do this is because the spacer is added after computeBounds. If the bottom row width is fixed, why does it matter what the spacer width is.

samelhusseini:
Closing in favour of https://github.com/google/blockly/pull/3575. Thanks @alschmiedt

PR #3575, opened 2020-01-10
Geras inline alignment of rows + max bottom row width

### LINKED ISSUE #3367: Statement input rendering: bar below input got longer

samelhusseini:
##  The basics
- [x] I branched from develop

### Resolves

Fixes #3367

### Proposed Changes

Add hook for getting desired width of a row.
Move MAX_BOTTOM_ROW constants to geras as it only applies there.
Don't take row alignment into account if inline.
Override alignRowElements_ in geras to align all rows (top, bottom, spacer) to the width of the max of the prev / next input.

### Reason for Changes

Geras rendering fix.

### Test Coverage

Tested a bunch of blocks in the blocks factory matching spec of the old renderer.

Screen Shot 2020-01-10 at 1 15 54 PM

Tested on:

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -191,8 +191,8 @@ Blockly.blockRendering.RenderInfo.prototype.getRenderer = function() {
 Blockly.blockRendering.RenderInfo.prototype.measure = function() {
   this.createRows_();
   this.addElemSpacing_();
-  this.addRowSpacing_();
   this.computeBounds_();
+  this.addRowSpacing_();
   this.alignRowElements_();
   this.finalize_();
 };
@@ -553,7 +553,7 @@ Blockly.blockRendering.RenderInfo.prototype.alignRowElements_ = function() {
       var currentWidth = row.width;
       var desiredWidth = this.getDesiredRowWidth_(row);
       var missingSpace = desiredWidth - currentWidth;
-      if (missingSpace) {
+      if (missingSpace > 0) {
         this.addAlignmentPadding_(row, missingSpace);
       }
     }
@@ -617,7 +617,7 @@ Blockly.blockRendering.RenderInfo.prototype.alignStatementRow_ = function(row) {
   var desiredWidth = this.statementEdge;
   // Add padding before the statement input.
   var missingSpace = desiredWidth - currentWidth;
-  if (missingSpace) {
+  if (missingSpace > 0) {
     this.addAlignmentPadding_(row, missingSpace);
   }
   // Also widen the statement input to reach to the right side of the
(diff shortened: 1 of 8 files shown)


core/renderers/common/info.js
```diff
@@ -551,15 +551,26 @@ Blockly.blockRendering.RenderInfo.prototype.alignRowElements_ = function() {
...
   }
 };
 
+/**
+ * Calculate the desired width of an input row.
+ * @param {!Blockly.blockRendering.Row} _row The input row.
+ * @return {number} The desired width of the input row.
+ * @protected
+ */
+Blockly.blockRendering.RenderInfo.prototype.getDesiredRowWidth_ = function(
```
alschmiedt: should we change the name of this to get getInputRowDesiredWidth? Or something with input since that is the only rows it applies to?

core/renderers/geras/info.js
```diff
@@ -376,6 +377,67 @@ Blockly.geras.RenderInfo.prototype.getElemCenterline_ = function(row, elem) {
...
+/**
+ * @override
+ */
+Blockly.geras.RenderInfo.prototype.alignRowElements_ = function() {
+  if (!this.isInline) {
+    Blockly.geras.RenderInfo.superClass_.alignRowElements_.call(this);
+    return;
+  }
+
+  // Walk up each row from the bottom, keeping track of the right input edge.
```
alschmiedt: Walk backwards through elements in a row, ...

core/renderers/geras/info.js
```diff
@@ -376,6 +377,67 @@ Blockly.geras.RenderInfo.prototype.getElemCenterline_ = function(row, elem) {
...
+      if (row.hasStatement) {
+        this.alignStatementRow_(
+            /** @type {!Blockly.blockRendering.InputRow} */ (row));
+      }
+      if (prevInput && prevInput.hasStatement && row.width < prevInput.width) {
+        row.nextRightEdge = prevInput.width;
+      } else {
+        nextRightEdge = row.width;
+      }
+      prevInput = row;
```
alschmiedt: Is this technically just prevRow? Since it is not just going through the input rows?

core/renderers/common/info.js
samelhusseini: Yep.

core/renderers/geras/info.js
samelhusseini: Maybe "Walk backgrounds through rows on the block"? Because we're not going through the elements in the row.

core/renderers/geras/info.js
samelhusseini: No prevInput, because of the if statement only inputs go down that codepath.

PR #3579, opened 2020-01-12
Fixes blockly games blocks being black

### LINKED ISSUE #3397: All stock blocks in Blockly Games now black.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3397

### Proposed Changes
Use valid['colourPrimary'] instead of valid.colourPrimary

### Test Coverage
Working on it.

Tested on:

### Additional Information
Still not sure why this was happening for blockly games but I couldn't reproduce in blockly demos where we use compressed.

### DIFF
--- a/core/renderers/common/constants.js
+++ b/core/renderers/common/constants.js
@@ -634,7 +634,7 @@ Blockly.blockRendering.ConstantProvider.prototype.validatedBlockStyle_ =
   valid['colourSecondary'] = valid['colourSecondary'] ?
       Blockly.utils.parseBlockColour(valid['colourSecondary']).hex :
       this.generateSecondaryColour_(valid.colourPrimary);
-  valid['colourTertiary'] = valid['colourTertiary'] ?
+  valid['colourTertiary'] = valid.colourTertiary ?
       Blockly.utils.parseBlockColour(valid['colourTertiary']).hex :
       this.generateTertiaryColour_(valid.colourPrimary);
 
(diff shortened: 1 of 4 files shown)


samelhusseini:
Blockly games uses ADVANCED compilation whereas we use SIMPLE_OPTIMIZATIONS for the compressed version we ship / what's used in the demos. 

I'm not sure this is the right fix though.  If we use qualified names for those "internal" properties of the style object, then we'd have to use qualified names everywhere we reference colourPrimary, secondary etc.

samelhusseini:
You should be able to repro this locally in Blockly with the compile script in: ``tests/compile/compile.sh``

alschmiedt:
For future reference: The difference between blockly games and blockly was that blockly games uses setColour while blockly uses setStyle. I am working on adding a block to our compile test that uses setColour.

PR #3580, opened 2020-01-13
Start hat support in zelos.

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Override start hat constants in zelos to rendering of: 

Screen Shot 2020-01-13 at 11 03 08 AM

### Reason for Changes

Zelos rendering.

### Test Coverage

Tested in rendering playground.

Tested on:

### DIFF
--- a/core/renderers/zelos/measurables/rows.js
+++ b/core/renderers/zelos/measurables/rows.js
@@ -62,7 +62,10 @@ Blockly.zelos.TopRow.prototype.endsWithElemSpacer = function() {
  * @override
  */
 Blockly.zelos.TopRow.prototype.hasLeftSquareCorner = function(block) {
-  return !!block.outputConnection;
+  var hasHat = (typeof block.hat !== 'undefined' ?
+      block.hat === 'cap' : this.constants_.ADD_START_HATS) &&
+      !block.outputConnection && !block.previousConnection;
+  return !!block.outputConnection || hasHat;
 };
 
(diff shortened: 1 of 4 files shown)


core/renderers/zelos/measurables/rows.js
```diff
@@ -62,7 +62,10 @@ Blockly.zelos.TopRow.prototype.endsWithElemSpacer = function() {
  * @override
  */
 Blockly.zelos.TopRow.prototype.hasLeftSquareCorner = function(block) {
-  return !!block.outputConnection;
+  var hasHat = (typeof block.hat !== 'undefined' ?
```
alschmiedt: Just curious why you are using typeof instead of a.hat != undefined?

samelhusseini: Mostly for consistency with https://github.com/google/blockly/blob/develop/core/renderers/measurables/rows.js#L293
but in general this type of check is no longer necessary (it used to be back in the day, when browsers let you change the value of ``undefined``)
Another subtle difference between the two is the typeof check makes sure it had never been defined before, rather than block.hat being set to undefined.

I'll defer the question over to @moniika as she was the one that changed it from ``block.hat`` to ``typeof block.hat !== 'undefined'`` in https://github.com/google/blockly/commit/26db9167db6da15c5ba463e345727243054bea64

PR #3590, opened 2020-01-14
Added optional data property to variable model with storing into/retrieving from XML

radovan76:
## The basics

- [x] I branched from develop

### Proposed Changes
Added data property to variable model same way as is in the block.

### Reason for Changes
May be used by 3rd parties for meta information. And I need it in my project :-)

### Test Coverage
Tried to create variable, set data, export to xml, retrieve from xml and compare. All in /tests/playground.html

Tested on:
 Chrome

### Additional Information
Had to change reading of variable name from XML, because textContent of parent element is changed by adding childNode with textContent.
For discussion is if it is not problematic about backward compatibility. If yes, better could be using attribute than element (but strange if anybody stores large data).

Be careful, this is my first pull request in this project :-)

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -602,15 +602,17 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
     var type = xmlChild.getAttribute('type');
     var id = xmlChild.getAttribute('id');
     var name = '';
+    var data = null;
     for (var i = 0; i < xmlChild.childNodes.length; i++) {
-      if (xmlChild.childNodes[i].nodeType === Node.TEXT_NODE) {
-        name += xmlChild.childNodes[i].textContent;
+      if (xmlChild.childNodes[i].nodeType == Blockly.utils.dom.Node.TEXT_NODE) {
+        name = name+xmlChild.childNodes[i].textContent;
+      } else if (xmlChild.childNodes[i].nodeName == 'data') {
+        data = xmlChild.childNodes[i].textContent;
       }
     }
     var variable = workspace.createVariable(name.trim(), type, id);
-    var dataXmlArray = xmlChild.getElementsByTagName('data');
-    variable.data = dataXmlArray.length ? dataXmlArray[0].textContent : null;
-  }
+    variable.data = data;
+    }
 };
 
 /**
--- a/core/xml.js
+++ b/core/xml.js
@@ -603,12 +603,13 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
     var id = xmlChild.getAttribute('id');
     var name = '';
     for (var i = 0; i < xmlChild.childNodes.length; i++) {
-      if (xmlChild.childNodes[i].nodeType === Node.TEXT_NODE)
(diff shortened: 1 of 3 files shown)


radovan76:
It should be also stored in VAR_CREATE and VAR_DELETE events.

core/xml.js
```diff
@@ -596,9 +601,15 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
     }
     var type = xmlChild.getAttribute('type');
     var id = xmlChild.getAttribute('id');
-    var name = xmlChild.textContent;
-
-    workspace.createVariable(name, type, id);
+    var name = '';
+    for (var i = 0; i < xmlChild.childNodes.length; i++) {
+      if (xmlChild.childNodes[i].nodeType === Node.TEXT_NODE) {
```
samelhusseini: Use ``Blockly.utils.dom.Node.TEXT_NODE`` instead for Node.js compatibility. Why not check for ``childNodes[i].nodeName`` == 'data'`` here and populate your data attribute.

Since we're only writing a single name / data attribute, I think it's fair to override rather than append when populating each attribute.

core/xml.js
```diff
@@ -596,10 +601,18 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
...
     var type = xmlChild.getAttribute('type');
     var id = xmlChild.getAttribute('id');
-    var name = xmlChild.textContent;
-
-    workspace.createVariable(name, type, id);
-  }
+    var name = '';
+    var data = null;
+    for (var i = 0; i < xmlChild.childNodes.length; i++) {
+      if (xmlChild.childNodes[i].nodeType == Blockly.utils.dom.Node.TEXT_NODE) {
```
samelhusseini: store xmlChild.childNodes[i] in a var, perhaps: ``xmlVariableChild``

core/xml.js
radovan76: Okay, changed. Variable node can have more textNodes when is XML prettified (spaces before and after data node). So I have to append name in my opinion.

mattbishop:
I have a block-oriented Map in my app where the key is the blockId and the value is a "metadata" object, or domain object. I am always using the blockIds from events, child blocks or parents to find the metadata and respond. If you are attaching these to the XML, cool! It would be really helpful to have a top-level method on the workspace to get the metadata by blockId, as the block may not be handy all the time.

radovan76:
If I want to add it to VAR_CREATE event, I have to add it as optional parameter (opt_data) to:
```
VariableModel constructor
Blockly.VariableMap.prototype.createVariable 
Blockly.Workspace.prototype.createVariable
Blockly.WorkspaceSvg.prototype.createVariable
```
without it is event fired before data assign.

....will you be ok with that changes?

core/xml.js
radovan76: okay, changed and replaced "i" by"j"

BeksOmega:
Sorry we never got around to merging this @radovan76 :/ I'm going to go ahead and close this because we're iceboxing the XML system.

PR #3592, opened 2020-01-14
Setting default options in workspace constructor.

### LINKED ISSUE #3459: Workspace options should always be parsed
**Describe the bug**

Because options are parsed in inject instead of the workspace constructor, you can get headless workspaces the opperate differently from svg workspaces. [https://groups.google.com/forum/#!topic/blockly/tfHCA7afoC0](https://groups.google.com/forum/#!topic/blockly/tfHCA7afoC0)

This also makes working with flyouts confusing, because while they do have workspaces they may not work as you first assume.

**To Reproduce**

1. Create a rendered workspace and a headless workspace. Pass options to neither.
2. Deserialize the following xml to both workspaces.

```
  <variables>
    <variable>i</variable>
  </variables>
  <block type="variables_set">
    <field name="VAR">i</field>
    <value name="VALUE">
      <block type="lists_create_with">
        <mutation items="3"/>
        <value name="ADD0">
          <block type="text">
            <field name="TEXT">1</field>
          </block>
        </value>
        <value name="ADD1">
          <block type="text">
            <field name="TEXT">2</field>
          </block>
        </value>
        <value name="ADD2">
          <block type="text">
            <field name="TEXT">3</field>
          </block>
        </value>
      </block>
    </value>
    <next>
      <block type="text_print">
        <value name="TEXT">
          <shadow type="text">
            <field name="TEXT"/>
          </shadow>
          <block type="lists_getIndex">
            <mutation statement="false" at="true"/>
            <field name="MODE">GET</field>
            <field name="WHERE">FROM_START</field>
            <value name="VALUE">
              <block type="variables_get">
                <field name="VAR">i</field>
              </block>
            </value>
            <value name="AT">
              <block type="math_number">
                <field name="NUM">1</field>
              </block>
            </value>
          </block>
        </value>
      </block>
    </next>
  </block>
</xml>
```
3. Export both workspaces to javascript.

```
// Rendered.
var i;


i = ['1', '2', '3'];
window.alert(i[0]);
```

```
// Headless
var i;


i = ['1', '2', '3'];
window.alert(i[1]);  //Bad!
```
4. Observe how the headless workspace behaves differently.

**Expected behavior**

Both workspaces should result in `window.alert(i[0]);`.

**Additional context**

Flyouts, mutators, and mutator flyouts should get a thorough testing if this is fixed, because there may be code relying on their current behavior.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#3459

### Proposed Changes

Uses default `Blockly.Options` if it is not provided to the Workspace constructor.

### Reason for Changes

Before this change, the options set for workspaces through the constructor vs inject when no options were provided were different. This especially caused confusion for the option for oneBasedIndexing which can result in different code being generated.

### Test Coverage

Compared code generation between headless and non-headless workspaces without parameters provided.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -77,7 +77,7 @@ Blockly.inject = function(container, opt_options) {
 
   var workspace = Blockly.createMainWorkspace_(svg, options, blockDragSurface,
       workspaceDragSurface);
-  Blockly.user.keyMap.setKeyMap(workspace.options.keyMap);
+  Blockly.user.keyMap.setKeyMap(options.keyMap);
 
   Blockly.init_(workspace);
 
--- a/core/inject.js
+++ b/core/inject.js
@@ -77,7 +77,7 @@ Blockly.inject = function(container, opt_options) {
 
   var workspace = Blockly.createMainWorkspace_(svg, options, blockDragSurface,
       workspaceDragSurface);
-  Blockly.user.keyMap.setKeyMap(options.keyMap);
+  Blockly.user.keyMap.setKeyMap(workspace.options.keyMap);
 
   Blockly.init_(workspace);
 
@@ -161,24 +161,25 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
   options.parentWorkspace = null;
   var mainWorkspace =
       new Blockly.WorkspaceSvg(options, blockDragSurface, workspaceDragSurface);
-  mainWorkspace.scale = options.zoomOptions.startScale;
+  var wsOptions = mainWorkspace.options;
+  mainWorkspace.scale = wsOptions.zoomOptions.startScale;
   svg.appendChild(mainWorkspace.createDom('blocklyMainBackground'));
 
   // Set the theme name and renderer name onto the injection div.
   Blockly.utils.dom.addClass(mainWorkspace.getInjectionDiv(),
-      (mainWorkspace.options.renderer || 'geras') + '-renderer');
+      (wsOptions.renderer || 'geras') + '-renderer');
   Blockly.utils.dom.addClass(mainWorkspace.getInjectionDiv(),
       mainWorkspace.getTheme().name + '-theme');
 
-  if (!options.hasCategories && options.languageTree) {
+  if (!wsOptions.hasCategories && wsOptions.languageTree) {
     // Add flyout as an <svg> that is a sibling of the workspace svg.
     var flyout = mainWorkspace.addFlyout('svg');
     Blockly.utils.dom.insertAfter(flyout, svg);
   }
-  if (options.hasTrashcan) {
+  if (wsOptions.hasTrashcan) {
     mainWorkspace.addTrashcan();
   }
-  if (options.zoomOptions && options.zoomOptions.controls) {
+  if (wsOptions.zoomOptions && wsOptions.zoomOptions.controls) {
     mainWorkspace.addZoomControls();
   }
   // Register the workspace svg as a UI component.
@@ -188,7 +189,7 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
   // A null translation will also apply the correct initial scale.
   mainWorkspace.translate(0, 0);
 
-  if (!options.readOnly && !mainWorkspace.isMovable()) {
+  if (!wsOptions.readOnly && !mainWorkspace.isMovable()) {
     // Helper function for the workspaceChanged callback.
     // TODO (#2300): Move metrics math back to the WorkspaceSvg.
     var getWorkspaceMetrics = function() {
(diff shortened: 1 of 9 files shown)


core/workspace.js
```diff
@@ -40,7 +41,7 @@ Blockly.Workspace = function(opt_options) {
   this.id = Blockly.utils.genUid();
   Blockly.Workspace.WorkspaceDB_[this.id] = this;
   /** @type {!Blockly.Options} */
-  this.options = opt_options || /** @type {!Blockly.Options} */ ({});
+  this.options = opt_options || new Blockly.Options({});
```
samelhusseini: Not sure this fixes the issue. This might use a better default if no options are passed, but what if a ``{}`` is passed in. It wouldn't go through Blockly.Options, and thus ``options.oneBasedIndex`` would be set to false.

moniika: I've updated the logic to be the same as in inject.

core/workspace.js
```diff
@@ -40,7 +41,7 @@ Blockly.Workspace = function(opt_options) {
   this.id = Blockly.utils.genUid();
   Blockly.Workspace.WorkspaceDB_[this.id] = this;
   /** @type {!Blockly.Options} */
-  this.options = opt_options || /** @type {!Blockly.Options} */ ({});
+  this.options = new Blockly.Options(opt_options || {});
```
samelhusseini: This makes sense for headless, but in rendered mode a Blockly.WorkspaceSvg instance calls it's super constructor with a options object of type ``Blockly.Options``. This would now pipe that Blockly.Options object into another Blockly.Options object.

moniika: Good point. Looking more thoroughly through the code, changing the reference of the options object causes issues with non-headless and also I didn't realize there was a `Blockly.Options` and a `Blockly.BlocklyOptions`.

Given this, I think the earlier solution of:
`opt_options || new Blockly.Options({});`
makes more sense here.

core/workspace.js
moniika: Changed back after noticing there is `Blockly.Options` and a `Blockly.BlocklyOptions`,
I think this makes sense here because if someone was to pass an empty object, it would actually be the wrong type and I think its fine that it isn't parsed.
This fixes the inconsistency between the constructor with no option passed and inject with no option passed.

samelhusseini: Okay that makes sense. In which case, I would update any of our tests that pass in an object in the constructor to instead wrap that object with a Blockly.Options object.

core/inject.js
```diff
@@ -77,7 +77,7 @@ Blockly.inject = function(container, opt_options) {
 
   var workspace = Blockly.createMainWorkspace_(svg, options, blockDragSurface,
       workspaceDragSurface);
-  Blockly.user.keyMap.setKeyMap(options.keyMap);
+  Blockly.user.keyMap.setKeyMap(workspace.options.keyMap);
```
samelhusseini: Why not continue using options? workspace.options should be == options.

core/options.js
```diff
@@ -155,6 +155,12 @@ Blockly.Options = function(options) {
  */
 Blockly.BlocklyOptions = function() {};
 
+/**
+ * The SVG element for the grid pattern.
+ * @type {SVGElement}
+ */
+Blockly.Options.prototype.gridPattern = null;
```
samelhusseini: Can we move this to the constructor.

core/workspace.js
moniika: Test have been updated.

core/options.js
samelhusseini: Even if it's not passed into the options and you just want to set it to ``null``. I would set it in the constructor as per https://github.com/google/blockly/issues/2892

moniika: Updated.

core/inject.js
moniika: Changed back.

core/mutator.js
```diff
@@ -161,21 +161,23 @@ Blockly.Mutator.prototype.createEditor_ = function() {
...
-    toolboxPosition: this.block_.RTL ? Blockly.TOOLBOX_AT_RIGHT :
-        Blockly.TOOLBOX_AT_LEFT,
-    horizontalLayout: false,
-    getMetrics: this.getFlyoutMetrics_.bind(this),
-    setMetrics: null,
-    renderer: this.block_.workspace.options.renderer
-  });
+  var workspaceOptions = new Blockly.Options(
+      /** @type {!Blockly.BlocklyOptions} */
+      ({
```
mattbishop: Just curious, why are you wrapping the options object in a closure?

alschmiedt: I think she is just casting the options to the correct type so that it passes our type checks.

Issue #3553, opened 2020-01-07
Investigate how fields cache options


alschmiedt:
**Describe the bug**
When a dropdown field list refreshes the selected item does not match the real selection. 
This occurs when an item in the list is renamed and causes a resort of the options list.
Bug was introduced in #3036. 

**To Reproduce**
1. Navigate to test blocks
2. Choose Dynamic dropdown 
3. Add 'aa', 'bb', 'cc' as options
4. Choose 'cc'
5. In the console rename 'cc' to be 'ab'
6. Observe 
<img width="206" alt="Screen Shot 2020-01-06 at 4 13 00 PM" src="https://user-images.githubusercontent.com/23059043/71858351-b7372800-309f-11ea-83cd-23c318631959.png">

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**
The forum post found [here](https://groups.google.com/forum/#!topic/blockly/h34eayeVOzM) describes the issue further.

samelhusseini:
Fixed.

Issue #3099, opened 2019-09-26
Dynamic dropdowns are not responding to programmatic setValue calls correctly


BeksOmega:
**Describe the bug**

If you try to set a dynamic dropdown's value programmatically it won't regenerate its options, causing it to sometimes throw the 'Cannot set the dropdown's value to an unavailable option.' warning when it should not.

**To Reproduce**

1. Add the following block to the playground:
```
Blockly.Blocks['test_block'] = {
  init: function() {
    this.appendDummyInput()
      .appendField(new Blockly.FieldDropdown([
        ['a', 'A'],
        ['b', 'B'],
        ['c', 'C'],
        // This is your validator.
      ], this.firstDropdownListener), 'FIRST')
      // This is your dynamic option generator.
      .appendField(new Blockly.FieldDropdown(this.generateOptions), 'SECOND');
  },

  firstValue: 'A',

  firstDropdownListener: function(newValue) {
    var sourceBlock = this.getSourceBlock();
    // We have to cache the value to the block because setFieldValue() calls
    // generateOptions(), and generateOptions() wouldn't be able to retrieve
    // the newValue using getFieldValue() at this point (because the new
    // value hasn't been applied yet).
    sourceBlock.firstValue = newValue;
    sourceBlock.setFieldValue(sourceBlock.getOptions(newValue)[0][1], 'SECOND');
  },

  generateOptions: function() {
    var sourceBlock = this.getSourceBlock();
    // The menu generation function is called in the constructor, at which
    // point the source block has not yet been applied, so we have to handle
    // that case.
    if (sourceBlock) {
      return sourceBlock.getOptions(sourceBlock.firstValue);
    } else {
      return [
        ['alligator', 'ALLIGATOR'],
        ['apple', 'APPLE'],
        ['airplane', 'AIRPLANE']
      ];
    }
  },

  getOptions: function(firstValue) {
    switch (firstValue) {
      case 'A':
        return [
          ['alligator', 'ALLIGATOR'],
          ['apple', 'APPLE'],
          ['airplane', 'AIRPLANE']
        ];
      case 'B':
        return [
          ['ball', 'BALL'],
          ['basket', 'BASKET'],
          ['bottle', 'BOTTLE']
        ];
      case 'C':
        return [
          ['coffee', 'COFFEE'],
          ['captain', 'CAPTAIN'],
          ['carrot', 'CARROT']
        ];
    }
  }
};
```
2. Load the following xml:
```
[...]

PR #3599, opened 2020-01-15
Use an is dirty flag to determine whether or not to re-generate dynamic field dropdown options

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Proposal for fixing https://github.com/google/blockly/issues/3553
and https://github.com/google/blockly/issues/3099

### Proposed Changes

Instead of relying on the lifecycle of the block to determine whether or not to re-generate dynamic field dropdown options, use an is dirty flag to determine that. 
Developers would call ``setOptionListDirty`` on the field to set the dynamic option list to dirty which would in turn re-generate the dynamic options the next time ``getOptions`` is called.

### Reason for Changes

Bug fix.

### Test Coverage

Tested with example given in https://github.com/google/blockly/issues/3099

Tested on:

### Additional Information

@BeksOmega thoughts on this proposal?

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -79,6 +79,13 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator, opt_config) {
   this.generatedOptions_ = null;
 
   /**
+   * Whether or not we should re-generate dynamic options.
+   * @type {boolean}
+   * @private
+   */
+  this.isOptionListDirty_ = true;
+
+  /**
    * The currently selected index. The field is initialized with the
    * first option selected.
    * @type {number}
@@ -87,7 +94,7 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator, opt_config) {
   this.selectedIndex_ = 0;
 
   this.trimOptions_();
-  var firstTuple = this.getOptions(false)[0];
+  var firstTuple = this.getOptions()[0];
 
   // Call parent's constructor.
   Blockly.FieldDropdown.superClass_.constructor.call(
@@ -321,7 +328,7 @@ Blockly.FieldDropdown.prototype.dropdownCreate_ = function() {
   menu.setRightToLeft(this.sourceBlock_.RTL);
   menu.setRole(Blockly.utils.aria.Role.LISTBOX);
 
-  var options = this.getOptions(false);
+  var options = this.getOptions();
   this.selectedMenuItem_ = null;
   for (var i = 0; i < options.length; i++) {
     var content = options[i][0]; // Human-readable text or image.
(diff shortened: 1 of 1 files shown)


moniika:
I'm not sure if this is the correct way to solve the problem, but I also needed to add a call to `this.setOptionListDirty();` inside of `Blockly.FieldDropdown.prototype.doClassValidation_` in order to fix the issue in #3099. Is that something we should add?

moniika:
> I'm not sure if this is the correct way to solve the problem, but I also needed to add a call to `this.setOptionListDirty();` inside of `Blockly.FieldDropdown.prototype.doClassValidation_` in order to fix the issue in #3099. Is that something we should add?

Discussed offline. Developer would be responsible for calling setOptionsListDirty. In case of example in #3099 it would be called on the second dropdown within the first dropdown listener.

samelhusseini:
Discussed with @moniika offline, and the https://github.com/google/blockly/issues/3099#issuecomment-574792577 does fix the issue outlined in #3099. We'll wait to hear from @BeksOmega about what she thinks of this proposal, but otherwise we agree that this should resolve these edge cases with caching based on the lifecycle of the field.

mattbishop:
I'm not in favour of _"Developer would be responsible for calling setOptionsListDirty"_ because as a dynamic list author, I don't want to keep track of the before/after list to know if it's dirty. The way it is now, the dynamic list provider is called whenever a list needs to be shown, so I populate it from another data model.

mattbishop:
The key issue is the selected value in that dynamic list does not change, but it's position in the dynamic list does. The dropdown needs to change the selected index to match the new position.

BeksOmega:
Sorry for the late reply - busy school day.

Gahhh I really like this solution, but I don't know if it's ok to change this so late in the game :/

I think I agree with @mattbishop It seems like it would cause outside developers to have to do a significant amount of refactoring. And unlike other breaking changes (like the new rendering system) this doesn't add any new functionality, it just makes things run more efficiently.

So while I do really like this solution (and if I was rewriting blockly I'd probably use it), I think I'd have to pick a more backwards-compatible route and go with something like this:
1) Change the selectedIndex_ to a selectedOption_/object/whatever. Basically instead of storing the index it stores what's actually going to get displayed. This way we don't need to grab the options inside render anymore.
2) Remove the setOptionListDirty function and instead set `this.isOptionListDirty = true` at the end of doValueUpdate.
  This makes it so that the next time the menu is opened /or/ the next time setValue is called, the options will be regenereated.
3) Change getOptions to a wrapper function. It is now only called by outside developers, and ensures that it always returns an up-to-date set of options, while not messing up the dropdown's state.

I think that should cover everything, but I'm not sure. It's been a while hehe. You can view an example of my proposal on my [test branch](https://github.com/google/blockly/compare/develop...BeksOmega:DropdownFix) (it's an extension of this branch). I tested against #3099 and it worked.

Alternative 1: Just have doClassValidation call getOptions(false).

Alternative 2: Just rip out all of the cacheing stuff. I know I proposed it originally, but I think I /drastically/ underestimated the complexity of it when I did.

samelhusseini:
So there's a couple of things at play here, and I recognize that wasn't clear from the start of this proposal. The first is how do better do caching, allowing developers with dynamic dropdowns to setValue's programmatically https://github.com/google/blockly/issues/3099. The second has to do with how developers that "know" that the underlying model has change (ie: generateOptions returns a different set of options), can safely re-render the block https://github.com/google/blockly/issues/3553.

For the first point, I agree with both you guys that perhaps this proposal to have developer tell us when the underlying model is dirty puts the onus on the developer in tracking their model's changes, and you can basically achieve the same functionality by just setting ``field.generatedOptions_ = null``, or calling ``field.getOptions(false)``. Stripping out caching could be worth considering too. 

The selectedIndex_ issue is a little more intricate. The underlying issue is that we are storing some value that we retrieve in the different field lifecycle methods, and the issue @mattbishop is running into is because you're calling methods out of the field's lifecycle method order. In this example, it's selectedIndex_ but for example in FieldVariable that might be the variable_ property. We can't rely on developers in updating these cached values, instead it's best if they call public methods that refresh these properties. Can't remember who mentioned it, but a doValueUpdate_ and a forceRerender does fix the problem, so does a doClassValidation_ as @BeksOmega mentioned. The problem with doValueUpdate_ is that there's an assumption that the underlying value is still valid which might not be the case if say the value has been removed from the generated list.

Proposal: How about we allow setValue to accept a value equal to the field's value? This way calling setValue(field.getValue()) will trigger a validation, a value update, and re-render the field.

samelhusseini:
An alternative to supporting ``setValue(field.getValue())`` might be to call ``doValueUpdate_`` from forceRerender.

samelhusseini:
Closing in favour of https://github.com/google/blockly/pull/3602.

PR #3602, opened 2020-01-16
Use selectedOption instead of selectedIndex in field dropdown

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3553

### Proposed Changes

Reference selected option instead of selected index in field dropdown so re-names in the underlying model are reflected in the field.

### Reason for Changes

Bug fix.

### Test Coverage

Tested with sample block in bug. 

Tested on:

### Additional Information

@mattbishop thoughts?

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -78,20 +78,19 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator, opt_config) {
    */
   this.generatedOptions_ = null;
 
+  this.trimOptions_();
+
   /**
-   * The currently selected index. The field is initialized with the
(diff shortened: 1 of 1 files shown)


core/field_dropdown.js
```diff
@@ -78,20 +78,19 @@ Blockly.FieldDropdown = function(menuGenerator, opt_validator, opt_config) {
    */
   this.generatedOptions_ = null;
 
+  this.trimOptions_();
```
mattbishop: What does trimOptions_() do?

mattbishop: NM, existing code.

Issue #3614, opened 2020-01-21
Toolbox category not deselected after block is dragged out


moniika:
After a block is dragged out of a category in toolbox, the category is not un-selected (although the flyout is closed) and it takes 2 clicks to re-open that category.

**To Reproduce**

1. Drag a bug out of toolbox from category, observe how flyout closes, but the category is still highlighted.
2. Click on the **same** category to open the flyout,  observe flyout does not open but is no longer highlighted.
4. Click category again to open flyout.

**Expected behavior**

Category should be deselected after block is dragged out and it should only take one click to open flyout.

**Screenshots**
![5BFHo0RSFn9](https://user-images.githubusercontent.com/6621618/72835108-f5574e80-3c3e-11ea-9348-cd5d01305645.gif)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: chrome

PR #3616, opened 2020-01-21
Fix parsing and passing of workspace options around Blockly

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3614

### Proposed Changes

Workspace options were moved to all pass through the Blockly.Options parser. This means that we have to update the options sent to reflect the differences in the property names.

There's a number of side effects to this issue, RTL, mutators, etc.
This also fixes issues with Advanced compilation where the options object properties are renamed if not fully qualified.

### Reason for Changes

Bug fix.

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/mutator.js
+++ b/core/mutator.js
@@ -166,18 +166,17 @@ Blockly.Mutator.prototype.createEditor_ = function() {
       ({
         // If you want to enable disabling, also remove the
         // event filter from workspaceChanged_ .
-        disable: false,
-        languageTree: quarkXml,
-        parentWorkspace: this.block_.workspace,
-        pathToMedia: this.block_.workspace.options.pathToMedia,
-        RTL: this.block_.RTL,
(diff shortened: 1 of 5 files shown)


core/mutator.js
```diff
@@ -166,18 +166,17 @@ Blockly.Mutator.prototype.createEditor_ = function() {
       ({
         // If you want to enable disabling, also remove the
         // event filter from workspaceChanged_ .
-        disable: false,
-        languageTree: quarkXml,
-        parentWorkspace: this.block_.workspace,
-        pathToMedia: this.block_.workspace.options.pathToMedia,
-        RTL: this.block_.RTL,
-        toolboxPosition: this.block_.RTL ? Blockly.TOOLBOX_AT_RIGHT :
+        'disable': false,
```
mattbishop: Why quote the keys?

samelhusseini: For advanced compilation. Otherwise they’ll get renamed.

Issue #3621, opened 2020-01-21
Note picker broken (getColourBorder not a function)


NeilFraser:
Note picker in Blockly Games is suddenly broken on develop.
Video:  https://youtu.be/xA2b9efPuAM

```
Uncaught TypeError: this.sourceBlock_.getColourBorder is not a function
    at CustomFields.FieldPitch.showEditor_ (field_pitch.js:78)
    at CustomFields.FieldPitch.Blockly.Field.showEditor (field.js:613)
    at Blockly.TouchGesture.Blockly.Gesture.doFieldClick_ (gesture.js:734)
    at Blockly.TouchGesture.Blockly.Gesture.handleUp (gesture.js:590)
    at Blockly.TouchGesture.handleUp (touch_gesture.js:204)
    at HTMLDocument.wrapFunc (blockly.js:475)
```

alschmiedt:
Fixed blockly games and added getColourBorder back with @deprecated.

PR #3626, opened 2020-01-22
Add colour methods back to blockSvg

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
getColourBorder is used in a lot of fields. In an attempt to not break people we don't need to we are adding back the method and deprecating it. Also adding back helper methods for getting the different colours.
Original issue reported here: #3621

### Proposed Changes
Added back the methods and removed some unnecessary checks if the colours don't exist since they should all be going through validatedBlockStyle_ which makes sure primary, secondary, and tertiary all have colours. 

### Test Coverage

Tested on:

### Additional Information
Not sure if we also wanted to deprecate the getColourSecondary, getColourTertiary and getColourShadow ?

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -219,6 +219,48 @@ Blockly.BlockSvg.prototype.initSvg = function() {
 };
 
 /**
+ * Get the secondary colour of a block.
+ * @return {?string} #RRGGBB string.
+ */
+Blockly.BlockSvg.prototype.getColourSecondary = function() {
+  return this.style.colourSecondary;
+};
+
+/**
+ * Get the tertiary colour of a block.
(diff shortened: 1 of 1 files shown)


core/block_svg.js
```diff
@@ -218,6 +218,48 @@ Blockly.BlockSvg.prototype.initSvg = function() {
...
+ */
+Blockly.BlockSvg.prototype.getColourTertiary = function() {
+  return this.style.colourTertiary;
+};
+
+/**
+ * Get the shadow colour of a block.
+ * @return {?string} #RRGGBB string.
+ */
+Blockly.BlockSvg.prototype.getColourShadow = function() {
```
samelhusseini: Please deprecate this one as well.

PR #3635, opened 2020-01-22
Editable field spacing

### LINKED ISSUE #3632: TextInput fields have extra padding in readonly
**Describe the bug**
It looks like textinputs have more padding around them in readonly mode.

**To Reproduce**
1. Import following xml in to a playground that is in readonly mode 

` <xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_repeat_ext" id="6B|/UvKRbPaWKPGeVR._" x="137" y="112">
    <value name="TIMES">
      <shadow type="math_number" id="HR-}CEhizODrNSY,ubD%">
        <field name="NUM">10</field>
      </shadow>
    </value>
  </block>
</xml> `

**Expected behavior**

**Screenshots**
**Readonly true**
Screen Shot 2020-01-21 at 8 31 01 PM

**Readonly false**
Screen Shot 2020-01-21 at 8 31 10 PM

**Stack Traces**

```
```

**Additional context**

moniika:
## The basics

- [x] I branched from develop

### Resolves

#3632

### Proposed Changes

Subsituting checks for isEditable for field EDITABLE property for determining element spacing.

### Reason for Changes

Fields that can be editable like `TextInput` have property `isEditable = false` in `readonly`  and element spacing calculation uses the `isEditable` property, which causes a difference in padding depending on `readonly` state.

### Test Coverage

Opened all blocks in playground in geras and thasos.

### DIFF
--- a/core/renderers/geras/info.js
+++ b/core/renderers/geras/info.js
@@ -303,7 +303,7 @@ Blockly.geras.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
   // Spacing between two fields of the same editability.
   if (Blockly.blockRendering.Types.isField(prev) &&
       next && Blockly.blockRendering.Types.isField(next) &&
-      ((/** @type Blockly.blockRendering.Field */ (prev)).isEditable ===
+      ((/** @type Blockly.blockRendering.Field */ (prev)).isEditable ==
           (/** @type Blockly.blockRendering.Field */ (next)).isEditable)) {
     return this.constants_.LARGE_PADDING;
   }
--- a/core/renderers/geras/info.js
+++ b/core/renderers/geras/info.js
@@ -172,7 +172,8 @@ Blockly.geras.RenderInfo.prototype.addElemSpacing_ = function() {
 Blockly.geras.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
   if (!prev) {
     // Between an editable field and the beginning of the row.
-    if (next && Blockly.blockRendering.Types.isField(next) && next.isEditable) {
+    if (next && Blockly.blockRendering.Types.isField(next) &&
+        (/** @type Blockly.blockRendering.Field */ next).field.EDITABLE) {
       return this.constants_.MEDIUM_PADDING;
     }
     // Inline input at the beginning of the row.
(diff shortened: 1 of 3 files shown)


samelhusseini:
I'm not opposed to the extra Field type checks, and the type casts, but the field measurable's isEditable property is only used in this context (getInRowSpacing_) why not update that from 
```
this.isEditable = field.isCurrentlyEditable();
```
to
```
this.isEditable = field.EDITABLE;
```

moniika:
> I'm not opposed to the extra Field type checks, and the type casts, but the field measurable's isEditable property is only used in this context (getInRowSpacing_) why not update that from
> 
> ```
> this.isEditable = field.isCurrentlyEditable();
> ```
> 
> to
> 
> ```
> this.isEditable = field.EDITABLE;
> ```
> 
> https://github.com/google/blockly/blob/develop/core/renderers/measurables/row_elements.js#L94

Makes sense, I'll update it.

core/renderers/geras/info.js
```diff
@@ -298,9 +301,10 @@ Blockly.geras.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
   }
 
   // Spacing between two fields of the same editability.
-  if (!Blockly.blockRendering.Types.isInput(prev) &&
-      next && !Blockly.blockRendering.Types.isInput(next) &&
-      (prev.isEditable == next.isEditable)) {
+  if (Blockly.blockRendering.Types.isField(prev) &&
+      next && Blockly.blockRendering.Types.isField(next) &&
+      ((/** @type Blockly.blockRendering.Field */ (prev)).isEditable ===
```
samelhusseini: Only double equals.

core/renderers/thrasos/info.js
```diff
@@ -226,9 +229,10 @@ Blockly.thrasos.RenderInfo.prototype.getInRowSpacing_ = function(prev, next) {
   }
 
   // Spacing between two fields of the same editability.
-  if (!Blockly.blockRendering.Types.isInput(prev) &&
-      next && !Blockly.blockRendering.Types.isInput(next) &&
-      (prev.isEditable == next.isEditable)) {
+  if (Blockly.blockRendering.Types.isField(prev) &&
+      next && Blockly.blockRendering.Types.isField(next) &&
+      ((/** @type Blockly.blockRendering.Field */ (prev)).isEditable ===
```
samelhusseini: same here.

core/renderers/geras/info.js
moniika: done

rachel-fenichel:
Does this mean that `isCurrentlyEditable` should only be used to determine if a field can be clicked, and not for any other checks against editability?

BeksOmega:
> Does this mean that `isCurrentlyEditable` should only be used to determine if a field can be clicked, and not for any other checks against editability?

I think it depends on the context. I think of EDITABLE as like an intrinsic property of the field. While isCurrentlyEditable has to do with its current state.

( Put another way: If fields had a more robust inheritance structure I think fields would implement an IEditable interface, which would have an isCurrentlyEditable property. EDITABLE is like the IEditable interface)

So if you're checking if the field "implements" EDITABLE check EDITABLE; if you want to know if the field is currently editable check isCurrentlyEditable.

(Obligatory disclaimer: My knowledge could be out of date hehe)

PR #3668, opened 2020-01-29
Decompose text constants in constant provider construction

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Create `getTextConstants_` and `setTextConstants_` functions.  Use them to set up text constants.

### Reason for Changes
Make it easier to override a single set of values in the constructor.

This is the internal structure I was talking about for bucketing.

### Test Coverage
Tested Zelos and Geras in the playground.

### Additional Information

Note that I had to call `setTextConstants_` again in the zelos constructor, because it depends on the grid unit, which is not set before the super constructor is called.  I chose this over moving part of the configuration above the constructor.

I decided that values must be provided for everything in the text constants struct, and that child classes should call the super class to get a fully populated struct and then modify it.

### DIFF
--- a/core/renderers/common/constants.js
+++ b/core/renderers/common/constants.js
@@ -242,30 +242,6 @@ Blockly.blockRendering.ConstantProvider = function() {
   this.JAGGED_TEETH_WIDTH = 6;
 
   /**
-   * Point size of text.
-   * @type {number}
-   */
-  this.FIELD_TEXT_FONTSIZE = 11;
-
-  /**
-   * Height of text.
-   * @type {number}
-   */
-  this.FIELD_TEXT_HEIGHT = 16;
-
-  /**
-   * Text font weight.
-   * @type {string}
-   */
-  this.FIELD_TEXT_FONTWEIGHT = 'normal';
-
-  /**
-   * Text font family.
-   * @type {string}
-   */
-  this.FIELD_TEXT_FONTFAMILY = 'sans-serif';
-
-  /**
    * A field's border rect corner radius.
    * @type {number}
    */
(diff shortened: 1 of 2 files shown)


core/renderers/common/constants.js
```diff
@@ -702,6 +660,87 @@ Blockly.blockRendering.ConstantProvider.prototype.dispose = function() {
...
+ *  baselineCenter: boolean,
+ *  yOffset: number
+ * }} An object containing information about text sizing for fields.
+ * @protected
+ */
+Blockly.blockRendering.ConstantProvider.prototype.getTextConstants_ = function() {
+  return {
+    fontSize: 11,
+    height: 16,
+    weight: 'normal',
```
samelhusseini: It's hard to tell here what weight and height refer to without having it prefixed with ``font``.

core/renderers/common/constants.js
```diff
@@ -515,6 +471,8 @@ Blockly.blockRendering.ConstantProvider = function() {
     PUZZLE: 1,
     NOTCH: 2
   };
+
+  this.setTextConstants_();
```
samelhusseini: Why not do this in ``init``?

samelhusseini:
Correct me if I'm wrong, but it seems that we're giving a renderer multiple opportunities to adjust text constants. The first is with ``getTextConstants_``, where a renderer can change default text related rendering constants, the second is with ``setTextConstants_``, a renderer would override this method to adjust any internal properties that are dependent on text constants.

rachel-fenichel:
> Correct me if I'm wrong, but it seems that we're giving a renderer multiple opportunities to adjust text constants. The first is with getTextConstants_, where a renderer can change default text related rendering constants, the second is with setTextConstants_, a renderer would override this method to adjust any internal properties that are dependent on text constants.

That's correct.  We could make setTextConstants_ private if we want them to only be able to change this by modifying the default text related rendering constants.

Also once we wire in something from themes, does that mean we have three ways to do this?

core/renderers/common/constants.js
rachel-fenichel: Not sure what the correct order for these calls is.   No other constants depend on this one, so it could go in init, but lots of things depend on the padding one so it has to be done much earlier.  I chose to leave it in the same place as the old code.

samelhusseini:
So if we step back use the style guide requirement for fields to be [initialized in the constructor](https://google.github.io/styleguide/jsguide.html#features-classes-fields) to guide us, we could: 
- In common/constants: Initialize all properties on the constants object in the constructor with their defaults (as is today)
- Add a ``setConstants_`` method to constants and call it from init. This can then in turn call methods like ``setPaddingConstants_``, ``setTextConstants_`` in order that they matter, and we can divvy them up as needed. The purpose of this is to allow renderers to override fields.
- In zelos|geras/common: Any field overrides now go in ``setConstants_`` or one of the sub methods after extending these methods (override + call super). Constants introduced by the renderer like ``GRID_UNIT`` would still be added in the constructor (as per style guide). 
- When we pipe in themes, we can override certain properties like ``FIELD_TEXT_FONTFAMILY`` to the value of ``theme.font.family`` (or the like) if it exists. If the value isn't set in the field, it's safe to assume a default has already been set for the property in the constructor.

PR #3681, opened 2020-02-04
Improve App Engine deployment

NeilFraser:
Make the gcloudignore directives more specific by using a leading /.
Update instructions.


appengine/README.txt
```diff
@@ -11,20 +11,21 @@ structure:
...
- |- closure-library/  (Optional)
  `- static/
      |- blocks/
      |- core/
      |- demos/
      |- generators/
      |- media/
      |- msg/
-     |- tests/
+     |- tests/  (Optional)
```
rachel-fenichel: Why is this optional?

NeilFraser: Historically we've indicated that the uncompiled code is optional for an App Engine deploy (not ours, but other people's).  Thus `closure-library/` and `blockly_uncompressed.js` were both labeled as "Optional".  Since the tests only run in uncompiled mode, they should be in that bucket too.  I guess `blocks/`, `core/` , and `generators/` should all be optional too.  Will add these.

rachel-fenichel: I prefer to keep the playground required.

Issue #3684, opened 2020-02-07
Zelos multi-line blocks with output connections grow without a limit.


samelhusseini:
**Describe the bug**
https://groups.google.com/g/blockly/c/-EAGCXtq99Q describes community thoughts on the obnoxious and never-ending size of the list block's connection shape.

**To Reproduce**
1. Go to 'the playground
2. Click on the list category and add a create list block to the playground.
3. Use the mutator to add a few more items to the list.
4. Notice how the list block continues to grow.

**Expected behavior**

It would be good to limit the size of the list block's connection shape.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**

samelhusseini:
Fixed.

Issue #3676, opened 2020-01-31
Zelos: Reporter block with statement input


rachel-fenichel:
What it looks like in Zelos:
![blockError](https://user-images.githubusercontent.com/13686399/73582443-13446080-4442-11ea-8f75-872ff1b7db78.png)

What it looks like in Geras:
![image](https://user-images.githubusercontent.com/13686399/73582451-193a4180-4442-11ea-8e83-32917014374d.png)

What it should look like:
I'm not sure.  This is the function definition, not the call.  On the other hand, making this just get taller and taller means adding more dead space in the horizontal direction, which I don't like.

Navaneethan-B:
I'm facing this issue in the new Zelos rendering.

samelhusseini:
I think for this sort of block (used for things like Promises), we should borrow a design choice from SNAP and only render the left side of the connection shape like so: 

![Screen Shot 2020-02-06 at 5 12 14 PM](https://user-images.githubusercontent.com/16690124/73992782-0376c180-4905-11ea-8d0f-3d8c6bf49970.png)

samelhusseini:
So it would look something like this: 

<img width="449" alt="Screen Shot 2020-02-06 at 5 41 31 PM" src="https://user-images.githubusercontent.com/16690124/73993731-f27b7f80-4907-11ea-8c0b-6e40a44cfb79.png">

samelhusseini:
Fixed.

PR #3685, opened 2020-02-07
[zelos] Set a maximum width for dynamic connection shapes.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3684 and https://github.com/google/blockly/issues/3676

### Proposed Changes

Set a maximum for the width of a dynamic connection shape, thus getting to a shape like this: 

Screen Shot 2020-02-07 at 1 29 16 PM
and
Screen Shot 2020-02-07 at 1 34 28 PM

For dealing with the statement input + output connection. We keep track if we have any statement inputs in our input rows, and only render the left side of the connection if that is the case, like so: 

Screen Shot 2020-02-07 at 1 46 29 PM or 
Screen Shot 2020-02-07 at 1 47 39 PM

### Reason for Changes

Size of multi-line blocks with an output connection grows without a limit creating obnoxiously sized blocks.

### Test Coverage

Tested in playground.

Tested on:

### Additional Information

As a side effect of this change, when blocks with output connections are deeply nested, at some point (after hitting the max width), the blocks don't visually align exactly. I think that's a fair side effect. 

Thoughts @rachel-fenichel?

Screen Shot 2020-02-07 at 1 15 25 PM

### DIFF
--- a/core/renderers/common/info.js
+++ b/core/renderers/common/info.js
@@ -139,11 +139,10 @@ Blockly.blockRendering.RenderInfo = function(renderer, block) {
   this.rows = [];
 
   /**
-   * The total number of input rows added onto the block.
-   * @type {number}
-   * @protected
+   * An array of input rows on the block.
+   * @type {!Array.<!Blockly.blockRendering.InputRow>}
    */
-  this.inputRowNum_ = 1;
+  this.inputRows = [];
 
   /**
    * An array of measurable objects containing hidden icons.
@@ -206,6 +205,7 @@ Blockly.blockRendering.RenderInfo.prototype.createRows_ = function() {
   this.populateTopRow_();
   this.rows.push(this.topRow);
   var activeRow = new Blockly.blockRendering.InputRow(this.constants_);
+  this.inputRows.push(activeRow);
 
   // Icons always go on the first row, before anything else.
   var icons = this.block_.getIcons();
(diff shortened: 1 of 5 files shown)


core/renderers/zelos/constants.js
```diff
@@ -492,22 +503,31 @@ Blockly.zelos.ConstantProvider.prototype.makeHexagonal = function() {
  * @package
  */
 Blockly.zelos.ConstantProvider.prototype.makeRounded = function() {
+  var maxWidth = this.MAX_DYNAMIC_CONNECTION_SHAPE_WIDTH;
+  var maxHeight = maxWidth * 2;
+
   // The main path for the rounded connection shape is made out of a single arc.
```
alschmiedt: Update comment

core/renderers/common/info.js
```diff
@@ -139,11 +139,10 @@ Blockly.blockRendering.RenderInfo = function(renderer, block) {
...
 
   /**
-   * The total number of input rows added onto the block.
-   * @type {number}
-   * @protected
+   * An array of input rows on the block.
+   * @type {!Array.<!Blockly.blockRendering.InputRow>}
    */
-  this.inputRowNum_ = 1;
+  this.inputRows = [];
```
alschmiedt: Just curious why the change from just storing the number of inputs?

samelhusseini: I think that way we can update our "not so good looking" for loops from looping through all the rows to just the input rows.

PR #3689, opened 2020-02-13
Move connection utilities to new connUtils namespace

### LINKED ISSUE #3656: lastConnectionInRow inconsistent with lastConnectionInStack
**Describe the bug**

[lastConnectionInStack](https://github.com/google/blockly/blob/249d48477385f7ac389cb6f0dd9b2c06e9bc82dc/core/block.js#L533) lives on the block, while [lastConnectionInRow](https://github.com/google/blockly/blob/249d48477385f7ac389cb6f0dd9b2c06e9bc82dc/core/connection.js#L502) lives on the connection.

**Expected behavior**

Connection manipulation functions should be stored together. This makes them easier to use. And it makes it less likely that they will be accidentally duplicated.

**Additional context**

Both of these functions are package, so they can be safely moved.

In addition to the two above location options (connection or block) I would like to raise the option of moving these functions into their own "connection manipulation" file, as they don't access any private state. (lastConnectionInRow isn't even on the prototype)

[Todo link](https://github.com/google/blockly/blob/e2b510662cf60e4dba1e228f3b83549ea694e951/core/renderers/common/renderer.js#L192)

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #3656 

### Proposed Changes

Moved lastConnectionInStack, lastConnectionInRow, and singleConnection to a new connUtils namespace.

I also got rid of connectReciprocally, as it was only being called in one place.

### Reason for Changes

This will hopefullly make these functions easier to use and find. I think it also makes it less likely that they will be accidentally duplicated in the future.

### Test Coverage

Added unit tests for:
* Single Compatible Row Connection
  * Single Compatible Row Connection
  * Multiple Compatible Row Connections
  * Single Compatible Statement Connection
  * Multiple Compatible Input Connections
  * No Compatible Connections
  * No Connections
* Last Compatible Row Connection
  * All Compatible
  * Middle Not Compatible
  * End Not Compatible
* Last Stack Connection
  * Has Last Connection
  * Has No Last Connection

### Documentation

Noup

### Additional Information

This is just my interpretation of #3656 so I'm happy to make any changes!


rachel-fenichel:
I think I wasn't clear in my comments on the bug--I want to hold off on creating a new file for this at all.  Sorry for the extra work.

Issue #3613, opened 2020-01-21
Filename in compile error message is off-by-one (OBO)


vhermecz:
**Describe the bug**

The filename printed in do_compile on error is one before the filename that actually has the error.
The problem is that the code assumes that the closure compiler returns 1-based indexing for filenames, but it is actually 0 based.

samelhusseini:
Fixed.

PR #3692, opened 2020-02-14
#3613 fixing off-by-one error on compiler errors

vhermecz:
As both the python filename array and the Input_X indexing is 0 based, the `-1` operation is not needed.

fromFileArray in closure-compiler uses 0 based indexing. see:

https://github.com/google/closure-compiler/blob/09cca3b536923dfb86d4e2ea34c8ee97d3ab471c/src/com/google/javascript/jscomp/gwt/client/JsRunnerMain.java#L816

## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3613

### DIFF
--- a/build.py
+++ b/build.py
@@ -309,7 +309,7 @@ goog.provide('Blockly.utils.string');
     def file_lookup(name):
       if not name.startswith("Input_"):
         return "???"
-      n = int(name[6:]) - 1
+      n = int(name[6:])
       return filenames[n]
 
     if "serverErrors" in json_data:


vhermecz:
Originally created a [PR against my own fork.](https://github.com/vhermecz/blockly/pull/1)

alschmiedt:
@vhermecz Thanks for the fix! Just a heads up, we are moving away from build.py in favor of using gulp and npm (more info found [here](https://developers.google.com/blockly/guides/modify/web/building)). I checked and it doesn't look like this bug is replicated there.

PR #3695, opened 2020-02-18
Security update to JS-Interpreter

NeilFraser:
The past week has been ‘exciting’ over in the JS-Interpreter repo.


NeilFraser:
> lgtm, just curious why the rename to globalObject?

In JavaScript every function has a scope object that has properties for all variables at that level.  This object is not accessible, you can't get a handle on it, but it exists.  The exception is the global scope object which manifests itself as the tangible object 'window'.  So when JS does a variable look-up, it looks at the current function's scope object, and if there's no property of that name, looks at that scope object's parent scope, which is the scope object of the enclosing function, and so on, back to the global scope.  Thus each of these invisible scope objects has a list of properties and a link to its parent scope object.  All fine.

The 'with' construct messes with this a bit.  If one uses `with (foo) {...}` then the enclosed code gets executed with a new scope object, in this case `foo`.  So `foo.bar` can then be referred to as just `bar`.  'with' is banned at Google, but it is part of the ES5 spec.

Last week this bug was filed:
```
// global scope
function foo() {
  // function->global scopes
  var o = {}
  with (o) {
    // o->function->global scopes
    with (o) {
      // o->o->function->global scopes
      alert(bar);
    }
  }
}
```
In this case by the time we get to the alert, the current scope is `o`, and its parentScope is set to `o`.  To do a variable look-up, we need to walk back, and we correctly get to `o`.  But walking back again gets us to `o`, whereas we should get to `function`.  The two `o`s need to be the same object, yet need to have different parentScopes.

The solution was to split all scopes in the JS-Interpreter into two objects.  One object is a scope object that unique to every scope creation.  The other is a JavaScript dictionary of variables which might be common to more than one scope object.

Thus whereas before we were passing in the global `scope` object to the initialization function, now we are passing in that scope's data object, otherwise known as the global object.

Hope this makes sense.

NeilFraser:
For reference:  https://github.com/NeilFraser/JS-Interpreter/issues/183

alschmiedt:
Makes sense! Thanks for the explanation!

PR #3696, opened 2020-02-18
fix: Break breaking undo.

### LINKED ISSUE #2321: Disabling orphan break blocks breaks undo
### Problem statement

When a break block is auto-disabled because it's not in a loop it breaks undo. This is probably because the disable event isn't getting properly grouped with the events that triggered it (you have to manually add the event group when things are triggered by an event).

Side note: We should consider setting the event group while events are firing so that any new events are automatically grouped with the firing event.

### Expected Behavior

Can drag out an orphaned break block and then undo it.

### Actual Behavior

Undo gets stuck at the point where the block was disabled.

### Steps to Reproduce

1. Drag out a break block.
2. Try to undo it.

### Stack Traces

```
```

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #2321 

### Proposed Changes

This pull request makes it so that all events fired inside a listener to an event are automatically grouped with the firing event.

### Reason for Changes

This makes it so that behavior resulting from an event gets grouped with that event for undo purposes. I think that this generally makes sense.

The breaking of undo was actually related to the onchange listener not properly filtering events. The disable event was getting undone, which fires an event, then the change listener would re-disable the block. Grouping the move event eliminates that problem. But I can definitely fix the event filtering on the onchange listener if you want as well.

### Test Coverage

BreakBlocks

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

No clue what this might break. But with events, you know it's probably going to break something hehe.

[Edit: I'm going to test all of the existing change listeners]

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -653,9 +653,12 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
       this.undoStack_.shift();
     }
   }
+  var group = Blockly.Events.getGroup();
+  Blockly.Events.setGroup(event.group);
   for (var i = 0, func; (func = this.listeners_[i]); i++) {
     func(event);
   }
+  Blockly.Events.setGroup(group);
 };
 
 /**


BeksOmega:
Closing in favor of #3697  Totally cool with reopening it if you guys like this better!

PR #3697, opened 2020-02-19
fix: Break breaking undo.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2321 

### Proposed Changes

Fixes the event filtering and grouping in the break block so that it doesn't break undo.

Also makes it so that update disabled is called only if the block is not inherently disabled.

### Reason for Changes

Bugs are the worst.

This matches the current behavior, it's just called in a more centralized place now.

### Test Coverage

BreakBlocks2

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

I like this much better than #3696 This may not fix the wider problem, but I think it's safer.

### DIFF
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -327,24 +327,23 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
   /**
    * Called whenever anything on the workspace changes.
    * Add warning if this flow block is not nested inside a loop.
-   * @param {!Blockly.Events.Abstract} _e Change event.
+   * @param {!Blockly.Events.Abstract} e Change event.
    * @this {Blockly.Block}
    */
-  onchange: function(_e) {
-    if (!this.workspace.isDragging || this.workspace.isDragging()) {
+  onchange: function(e) {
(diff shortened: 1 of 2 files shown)


blocks/loops.js
```diff
@@ -327,24 +327,23 @@ Blockly.Constants.Loops.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
...
    * Add warning if this flow block is not nested inside a loop.
-   * @param {!Blockly.Events.Abstract} _e Change event.
+   * @param {!Blockly.Events.Abstract} e Change event.
    * @this {Blockly.Block}
    */
-  onchange: function(_e) {
-    if (!this.workspace.isDragging || this.workspace.isDragging()) {
+  onchange: function(e) {
+    if (!this.workspace.isDragging || this.workspace.isDragging() ||
+        e.type != Blockly.Events.BLOCK_MOVE || e.blockId != this.id) {
```
rachel-fenichel: This updates it to:
- don't change state at the start of a drag
- don't change state if it's not a move
- don't change state if the moving block isn't this block

rachel-fenichel: Update the comment--but otherwise looks good.

Issue #2793, opened 2019-08-09
Flyout Buttons with Different Font Size are Not Rendered Correctly


BeksOmega:
**Describe the bug**

Flyout buttons accept classes. If you give this class a font-size that is different from the standard one, the button does not render correctly.

**To Reproduce**

1. Add the following css to the playground:
```
.customFlyoutButton .blocklyText {
    font-size: 30px;
}
```
2. Add the following button to the playground (any toolbox category is fine):
`<button text="test" callbackKey="" web-class="customFlyoutButton"></button>`
3. Open the flyout.
4. Observe how the button is not rendered correctly.

**Expected behavior**

The flyout button should render to have the correct padding around the text, regardless of the height of the text.

**Screenshots**

![FlyoutButton_Text](https://user-images.githubusercontent.com/25440652/62794538-37be1200-ba89-11e9-89b3-968fb13ed133.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 76

**Additional context**

Thematically related to #2649, but the code is unrelated.

[This line](https://github.com/google/blockly/blob/de39d5f2d056f0f380dbbfb8f86cfa82ea449f55/core/flyout_button.js#L158) is the root of the problem.

samelhusseini:
Thanks for filing @BeksOmega. I wanted to add that this is more of a general problem with sizing based on font size that's throughout the project. You can find similar bugs with changing the font size of the text input or the ``.blocklyText`` class.

samelhusseini:
Fixed

PR #3698, opened 2020-02-19
Dynamic fonts

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/2793
Partially addresses https://github.com/google/blockly/issues/2694

### Proposed Changes

Use theme options in the renderer.
Support custom font styles either by extending the renderer and overriding ``getDefaultFontStyle_`` or by setting the font style in a theme.
Support the ability for flyout buttons to declare in CSS their font styles, and measure the height when the flyout opens.
Computing text height during initializes and anytime a theme changes. Font metrics include: height and baseline, baseline is used when ``FIELD_TEXT_BASELINE_CENTER`` is set to false (geras and Edge / IE).
Added debugger element to show text baseline of field labels.
**Text constants are all computed now, no more browser specific heights / offsets.**

### Reason for Changes

Dynamic font sizes.

### Test Coverage

Tested all blocks in the playground in all supported browsers.
Tested both ``FIELD_TEXT_BASELINE_CENTER`` true and false.

Tested on:

### Additional Information

Screen Shot 2020-02-18 at 5 44 08 PM

Screen Shot 2020-02-18 at 5 43 55 PM


rachel-fenichel:
@BeksOmega fyi

core/field.js
```diff
@@ -334,24 +326,12 @@ Blockly.Field.prototype.createBorderRect_ = function() {
...
   this.textElement_ = /** @type {!SVGTextElement} **/
       (Blockly.utils.dom.createSvgElement('text',
           {
             'class': 'blocklyText',
-            'y': baselineCenter ? this.size_.height / 2 : baselineY,
-            'dy': this.constants_.FIELD_TEXT_Y_OFFSET,
-            'x': xOffset
           }, this.fieldGroup_));
-  if (baselineCenter) {
+  if (this.constants_.FIELD_TEXT_BASELINE_CENTER) {
```
rachel-fenichel: Why is this still here, since you got rid of all other uses of constants?

core/field.js
```diff
@@ -619,22 +599,72 @@ Blockly.Field.prototype.updateWidth = function() {
...
+ * @protected
+ */
+Blockly.Field.prototype.positionTextElement_ = function(xOffset, contentWidth) {
+  if (!this.textElement_) {
+    return;
+  }
+  var constants = this.constants_;
+  var halfHeight = this.size_.height / 2;
+
+  this.textElement_.setAttribute('x', this.sourceBlock_.RTL ?
```
rachel-fenichel: Since you're doing updateSize on every render, it seems worth checking whether anything has changed before calling setAttribute (probably in updateSize, not here).

tests/playground.html
```diff
@@ -347,6 +347,18 @@
   }
 }
 
+function changeRenderingConstant(value) {
+  var type = document.getElementById('rendering-constant-selector').value;
+  if (type == 'fontSize') {
```
alschmiedt: Might be nice to check if there is already a fontStyle on the theme so when someone changes the size the font family will not change back to the default.

core/theme.js
```diff
@@ -57,27 +57,44 @@ Blockly.Theme = function(name, blockStyles, categoryStyles,
    * @private
    */
   this.componentStyles_ = opt_componentStyles || Object.create(null);
+
+  /**
+   * The font style.
+   * @type {?Blockly.Theme.FontStyle}
+   * @package
+   */
+  this.fontStyle = null;
```
alschmiedt: Should we add a getter for this? So someone could change just a single value on a theme instead of having to change the entire fontStyle?

core/field_checkbox.js
```diff
@@ -82,15 +82,6 @@ Blockly.FieldCheckbox.prototype.SERIALIZABLE = true;
...
 Blockly.FieldCheckbox.prototype.CURSOR = 'default';
 
-/**
- * Used to tell if the field needs to be rendered the next time the block is
- * rendered. Checkbox fields are statically sized, and only need to be
- * rendered at initialization.
- * @type {boolean}
- * @protected
- */
-Blockly.FieldCheckbox.prototype.isDirty_ = false;
```
rachel-fenichel: You removed isDirty but not any uses of it.

core/field_checkbox.js
```diff
@@ -108,19 +99,30 @@ Blockly.FieldCheckbox.prototype.configure_ = function(config) {
...
 };
 
+/**
+ * @override
+ */
+Blockly.FieldCheckbox.prototype.render_ = function() {
+  if (this.textContent_) {
+    this.textContent_.nodeValue = this.getDisplayText_();
+  }
+  this.updateSize_(this.constants_.FIELD_CHECKBOX_X_OFFSET);
```
rachel-fenichel: Why call the parent updateSize?  Checkboxes barely need anything from it.

core/theme.js
samelhusseini: I can make it public instead of package, and then users can access it directly.

tests/playground.html
samelhusseini: This is for playground testing, reloading the page will get you back to the defaults.

core/renderers/common/constants.js
```diff
@@ -571,6 +569,60 @@ Blockly.blockRendering.ConstantProvider.prototype.refreshTheme = function(
...
 };
 
+/**
+ * Set dynamic constants that depent on other values or theme properties.
+ * @param {!Blockly.Theme} theme The current workspace theme.
+ * @protected
+ */
+Blockly.blockRendering.ConstantProvider.prototype.setDynamicConstants_ =
+    function(theme) {
+    /* eslint-disable indent */
```
rachel-fenichel: You need an `eslint-enable` line for each of these, or it applies to the rest of the file.

core/field_colour.js
```diff
@@ -120,15 +120,6 @@ Blockly.FieldColour.prototype.SERIALIZABLE = true;
...
 Blockly.FieldColour.prototype.CURSOR = 'default';
 
-/**
- * Used to tell if the field needs to be rendered the next time the block is
- * rendered. Colour fields are statically sized, and only need to be
- * rendered at initialization.
- * @type {boolean}
- * @protected
- */
-Blockly.FieldColour.prototype.isDirty_ = false;
```
BeksOmega: Did you decide to remove the static rendering just for readability reasons?

Honestly I don't really mind b/c I don't think it was that much of an efficiency improvement. I'm just curious.

core/field_dropdown.js
```diff
@@ -564,10 +564,8 @@ Blockly.FieldDropdown.prototype.render_ = function() {
   } else {
     this.renderSelectedText_();
   }
-  if (this.borderRect_) {
-    this.borderRect_.setAttribute('height', this.size_.height);
-    this.borderRect_.setAttribute('width', this.size_.width);
-  }
+
+  this.positionBorderRect_();
```
BeksOmega: I like this change!

core/utils/dom.js
```diff
@@ -273,7 +273,7 @@ Blockly.utils.dom.getTextWidth = function(textElement) {
  * This method requires that we know the text element's font family and size in
  * advance. Similar to `getTextWidth`, we cache the width we compute.
  * @param {!Element} textElement An SVG 'text' element.
- * @param {number} fontSize The font size to use.
+ * @param {string} fontSize The font size to use.
```
BeksOmega: This looks like a breaking change, so it should be marked for doc updates.

core/field.js
```diff
@@ -619,22 +599,72 @@ Blockly.Field.prototype.updateWidth = function() {
...
+  this.positionBorderRect_();
 };
 
+/**
+ * Position a field's text element after a size change.
+ * @param {number} xOffset x offset to use when positioning the text element.
+ * @param {number} contentWidth The content width.
+ * @protected
+ */
+Blockly.Field.prototype.positionTextElement_ = function(xOffset, contentWidth) {
```
BeksOmega: Maybe add a note that this handles RTL and LTR automatically, up to you.

core/renderers/common/constants.js
```diff
@@ -571,6 +569,60 @@ Blockly.blockRendering.ConstantProvider.prototype.refreshTheme = function(
   }
 };
 
+/**
+ * Set dynamic constants that depent on other values or theme properties.
```
BeksOmega: I wouldn't worry about updating this PR but I think the [latest style recommendation](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments) is third person verbs for method descriptions.

core/renderers/common/constants.js
```diff
@@ -1001,7 +1062,7 @@ Blockly.blockRendering.ConstantProvider.prototype.createDom = function(svg) {
  * @param {string} name Name of the renderer.
  * @package
  */
-Blockly.blockRendering.ConstantProvider.prototype.injectCSS = function(
+Blockly.blockRendering.ConstantProvider.prototype.injectCSS_ = function(
```
BeksOmega: Tag and function name mismatch.

core/renderers/common/constants.js
```diff
@@ -508,9 +502,13 @@ Blockly.blockRendering.ConstantProvider = function() {
 
 /**
  * Initialize shape objects based on the constants set in the constructor.
+ * @param {!Blockly.Theme} theme The workspace theme object.
  * @package
  */
-Blockly.blockRendering.ConstantProvider.prototype.init = function() {
+Blockly.blockRendering.ConstantProvider.prototype.init = function(theme) {
```
BeksOmega: I think this needs to be flagged for doc updates as well. Subclasses need to pass the theme up to this parent, and I think currently they will not.

core/renderers/common/renderer.js
```diff
@@ -48,11 +48,25 @@ Blockly.blockRendering.Renderer = function(name) {
...
+  this.constants_.init(theme);
+};
+
+/**
+ * Refresh the renderer after a theme change.
+ * @param {!SVGElement} svg The root of the workspace's SVG.
+ * @param {!Blockly.Theme} theme The workspace theme object.
+ * @package
+ */
+Blockly.blockRendering.Renderer.prototype.refresh = function(svg, theme) {
```
BeksOmega: I think this needs to be protected since geras overrides it.

core/theme.js
```diff
@@ -57,27 +57,44 @@ Blockly.Theme = function(name, blockStyles, categoryStyles,
...
+ *          }}
+ */
 Blockly.Theme.CategoryStyle;
 
+/**
+ * A category style.
+ * @typedef {{
+ *            family:string,
+ *            weight:string,
+ *            size:number
```
BeksOmega: From the way these are used I think it should be `string?` and `number?`.

core/theme.js
```diff
@@ -57,27 +57,44 @@ Blockly.Theme = function(name, blockStyles, categoryStyles,
...
-  *            colour:string
-  *          }}
-  */
+ *            colour:string
+ *          }}
+ */
 Blockly.Theme.CategoryStyle;
 
+/**
+ * A category style.
```
BeksOmega:
```suggestion
 * A font style.
```

core/flyout_base.js
```diff
@@ -46,6 +46,7 @@ Blockly.Flyout = function(workspaceOptions) {
    */
   this.workspace_ = new Blockly.WorkspaceSvg(workspaceOptions);
   this.workspace_.isFlyout = true;
+  this.workspace_.setVisible(false);
```
BeksOmega: I think this is to keep the flyout from rerendering on theme changes right? If so please correct me. I might consider leaving a comment, but its your call.

tests/playground.html
```diff
@@ -631,6 +643,15 @@ <h1>Blockly Playground</h1>
...
     <input type="button" value="Spaghetti!" onclick="spaghetti(8)">
   </p>
+  <p>
+    Rendering: &nbsp;
+    <select id="rendering-constant-selector">
+      <option value="fontSize">Font Size</option>
+    </select>
+    <input type="range" min="3" max="50" value="4"
+      oninput="changeRenderingConstant(this.value)"
+      onchange="changeRenderingConstant(this.value)" />
```
BeksOmega: I like this addition!

core/field.js
samelhusseini: This is a static decision that can be made during init. A renderer decides whether the text is baseline centered and that isn't affected by any theme changes.

core/field_checkbox.js
samelhusseini: ``isDirty_`` is used by field.js to determine whether or not to re-render. Initializing a field with isDirty_ = false was a way for us to mark certain fields "static fields", ie: only init gets called for those fields and not render_.

Removing this because the checkbox field needs to render every time, as the font may have changed and the checkmark is a ``blocklyText`` element which can be resized with the font size.

core/field_checkbox.js
samelhusseini: Same as previous comment, checkboxes have a checkmark which is a ``blocklyText`` element and that needs updating if the font size changes.

core/field_colour.js
samelhusseini: I think you might be onto something in the case of field_colour. Will double check now, but for field_checkbox we needed to re-render as the checkmark's font size may have changed so removed the isDirty_ flag as it's no longer a static field.

core/field.js
samelhusseini: Good idea.

core/renderers/common/renderer.js
samelhusseini: ``package`` can be overridden.

core/theme.js
samelhusseini: Ideally it would be string|undefined but closure doesn't support that in typedefs :(

samelhusseini: I can make it nullable though.

core/theme.js
samelhusseini: Oh I didn't know you can make suggestions!!!

samelhusseini: but I already changed it, thanks though!

core/flyout_base.js
samelhusseini: Yes, and more importantly so the flyout representation of visibility is reflected in the workspace. https://github.com/google/blockly/blob/master/core/flyout_base.js#L130 is set to false

tests/playground.html
samelhusseini: Thanks!

core/field.js
rachel-fenichel: I see.  So it's back to the same question of how we indicate that some "constants" can be changed by themes, but with a clear reason why this one is static.

core/renderers/common/constants.js
```diff
@@ -508,9 +502,13 @@ Blockly.blockRendering.ConstantProvider = function() {
...
+
+  this.setTheme_(theme);
+  this.setDynamicConstants_(theme);
```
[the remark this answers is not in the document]
samelhusseini: Load order wise it I don't mind either way. Init is the entry point the rest are private methods. Happy to move setDynamicConstants to setTheme.

core/theme.js
rachel-fenichel: But then we wouldn't know about changes to it -> wouldn't know to update the constant provider.

core/renderers/common/constants.js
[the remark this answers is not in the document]
samelhusseini: It's now private. We re-inject the CSS when we createDom. A theme change calls dispose and createDom, thus removing the old renderer specific CSS and re-injecting the renderer specific CSS with new constant values.

core/renderers/common/constants.js
samelhusseini: Thanks

core/field.js
samelhusseini: I've done this here by setting the constant value in the constructor to ``this.DYNAMICALLY_SET`` which it's value is -1 but that doesn't matter as it's value is updated when ``setDynamicConstants`` is called. 
Course, that info isn't easily recognizable for developers.

samelhusseini: I could also make that clear in the description of a dynamic property. That its real value is set later on.

core/utils/dom.js
rachel-fenichel: The change here is just whether you expect the 'pt' on the font size to be added here or by the calling code, right?  What's the advantage of changing it?

core/field.js
samelhusseini: Agreed, we would be getting into the territory of keeping a cache of these values on the field itself (which I'm not opposed to).

core/field.js
rachel-fenichel: I like having that information in the description, since that's what's going to be exposed by generated jsdocs.

core/theme.js
samelhusseini: So right now the way we know the theme has changed at all is by the user calling ``workspace.setTheme(workspace.getTheme())``. Trigger a re-fresh.

We could call a refresh when the theme.setX is called, trouble with that is what if you're making multiple changes, ie: setX, setY ..., we'd be calling a refresh on every set.

core/utils/dom.js
samelhusseini: The flyout button dynamically reads from CSS what the font size should be, which generates the font size in pixels. I needed to change this in order to support both ``pt`` for fields and ``px`` for flyout buttons. 

As I think about this, there's a way to do this without making it a breaking change. I'll just add another util method that uses pixels.

core/theme.js
rachel-fenichel: Ah right, we're not trying to be magical about detecting these changes (and don't need to be).

core/utils/dom.js
samelhusseini: I'm struggling to come up with the other method name, how do we feel about ``getFastTextWidthWithSizeString``.

core/renderers/common/constants.js
rachel-fenichel: This is why I was asking about this change--it requires anyone who has already started tinkering with it to update their constructor code, and I'm not sure we really gain anything over calling setTheme after init.

core/field.js
samelhusseini: Done.

core/renderers/common/constants.js
```diff
@@ -569,6 +566,62 @@ Blockly.blockRendering.ConstantProvider.prototype.refreshTheme = function(
...
+
+  this.setDynamicConstants_(theme);
+};
+
+/**
+ * Sets dynamic constants that depent on other values or theme properties.
+ * @param {!Blockly.Theme} theme The current workspace theme.
+ * @protected
+ */
+Blockly.blockRendering.ConstantProvider.prototype.setDynamicConstants_ =
```
rachel-fenichel: Naming nit: setDynamicValues?  setDynamicProperties?  setDynamicConstants just feels like an oxymoron.

core/renderers/common/renderer.js
```diff
@@ -48,11 +48,25 @@ Blockly.blockRendering.Renderer = function(name) {
...
+  var constants = this.getConstants();
+  constants.dispose();
```
rachel-fenichel: Oh I see.  Now you call refresh on the renderer instead of on the constant provider.  Which is good because then the outside code doesn't need to know about the constant provider, but also weird because most of the renderer isn't being refreshed.

core/renderers/common/constants.js
samelhusseini: Thanks for pointing that out, I like that we're being diligent about unnecessary breaking changes.
I actually think setTheme needs to be called before init in case any dynamic constants are relied upon in any of the init methods, eg: makeNotch, but I agree we don't gain much, will refactor to call setTheme before init.

core/renderers/common/constants.js
samelhusseini: I like setDynamicProperties :)

core/renderers/common/constants.js
samelhusseini: Done.

BeksOmega:
I pulled down the branch and did some testing and it looks really good! I did find a smattering of small issues though so I'm just going to smatter them at you :P 

Multiline text field is a bit broken.
Size-CodeField

Geras inline inputs are kinda weird. But I don't think this is a bug.
Size-InlineInputs

Checkbox needs to re-render when the character is changed (this one is probably mine hehe)
Size-CheckCharacter

Dropdown carrots act differently in geras and zelos. In geras they match text in zelos they do not. Not sure if this is a bug, or which way it goes if it is. Might want to ask devs what they want.
Size-Dropdown1
Size-Dropdown2

Warnings are a bit broken. They do resize pretty well if you close & reopen them though.
Size-Warnings

Comments (block & workspace) don't reflect text size. Not sure if they should.
Size-BlockComment

Zelos flyout buttons don't reflect text size.
Size-ZelosButtons

I think the text category flyout label should automatically match the text size since it only overrides the font family, which it is not doing. But I'm not sure, this could be intended.
Size-FlyoutLabel

Label gaps act a bit oddly.
Size-FlyoutLables2
```
      <label text="test" web-class="customButton"></label>
      <sep gap="8"></sep>
      <label text="test" web-class="customButton"></label>
      <label text="test" web-class="customButton"></label>
      <spe gap="32"></spe>
      <label text="test"></label>
      <sep gap="8"></sep>
      <label text="test"></label>
      <label text="test"></label>
```
```
.customButton > text {
      font-family: Impact !important;
      font-size: 40px !important;
    }
```
I expect both `<sep gap="8"></sep>` gaps to be the same. But I'll admit this looks nice.

And I think that #2694 should be left open, as applying a class directly to the field still doesn't quite work.
Size-BlockLabel

If you want me to create separate issues for any of these I'm totally cool with that! (┏－‿ ◕)┏

Thanks for all your hard work!

samelhusseini:
> Multiline text field is a bit broken.
Will fix.

> Geras inline inputs are kinda weird. But I don't think this is a bug.
Yeah, that's expected. Thrasos doesn't have that problem as it vertically aligns everything. 
 
> Checkbox needs to re-render when the character is changed (this one is probably mine hehe)
I think a forceRerender would fix this in ``setCheckCharacter``, but I'm going to leave this one out as it's unrelated (Feel free to file an issue for this).

> Dropdown carrots act differently in geras and zelos. In geras they match text in zelos they do not. Not sure if this is a bug, or which way it goes if it is. Might want to ask devs what they want.
True, the background here is that zelos uses SVG and we don't resize the SVG based on the text font size. Whereas dropdowns are text in geras and are resized when the font changes.
Could look into this as a separate change.

> Warnings are a bit broken. They do resize pretty well if you close & reopen them though.
This one's a little tricky as icons don't have render loops. 
- We could add render loops, and re-render on a theme change.
- reset visibility for warnings that are open. (ie: set to false, and back to true, which would dispose and recreate the bubble)
- close all warnings when we need to re-render the workspace.

> Comments (block & workspace) don't reflect text size. Not sure if they should.
Also simple to make them reflect the size, currently renderers aren't applying it's font properties to ``.blocklyCommentTextarea``. 
Will have a think about this one, but will keep out of this change.


> Zelos flyout buttons don't reflect text size.
That's by design, I added an override in the [playground](https://github.com/google/blockly/pull/3698/files#diff-e408cc0526c5f52eeba282b9a74d2998R579) in order to test custom CSS.

> I think the text category flyout label should automatically match the text size since it only overrides the font family, which it is not doing. But I'm not sure, this could be intended.
Similar to comments, we can have a think about applying the renderer font properties to that as well.

> Label gaps act a bit oddly.
> Size-FlyoutLables2
> 
> ```
>       <label text="test" web-class="customButton"></label>
>       <sep gap="8"></sep>
>       <label text="test" web-class="customButton"></label>
>       <label text="test" web-class="customButton"></label>
>       <spe gap="32"></spe>
>       <label text="test"></label>
>       <sep gap="8"></sep>
>       <label text="test"></label>
>       <label text="test"></label>
> ```
> 
> ```
> .customButton > text {
>       font-family: Impact !important;
>       font-size: 40px !important;
>     }
> ```
> 
> I expect both `<sep gap="8"></sep>` gaps to be the same. But I'll admit this looks nice.
> 
> And I think that #2694 should be left open, as applying a class directly to the field still doesn't quite work.
> Size-BlockLabel
> 
> If you want me to create separate issues for any of these I'm totally cool with that! (┏－‿ ◕)┏
> 
> Thanks for all your hard work!

core/field_colour.js
samelhusseini: Revert field colour to a static field :)

core/field.js
samelhusseini: I'm going to take care of this in a separate change.

samelhusseini: https://github.com/google/blockly/issues/3701

samelhusseini:
Thanks all for reviewing and for your feedback! Apologies that this ended up being on the larger end of changes that I usually like. But hey dynamic fonts woo!

PR #3702, opened 2020-02-20
Support ES6 classes in renderer initialization

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

We are currently unable to do ``Blockly.blockRendering.register('test', Renderer);`` where Renderer is an ES6 class.

### Proposed Changes

Remove wrapper class during initialization (which was only really there to make the compiler happy).

### Reason for Changes

Make the renderer compatible with ES6 classes.

### Test Coverage

Tested on:


samelhusseini:
@googlebot I fixed

PR #3703, opened 2020-02-21
[zelos] Fix centered textinputs

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

https://github.com/google/blockly/pull/3685 introduced a bug where blocks with only output connections had right corners added to their elements resulting in an offset when centering the text of text fields: 
Screen Shot 2020-02-20 at 4 57 30 PM

### Proposed Changes

Only add a right corner if we don't have an output connection or if we do but we have a statement input to account for:

Screen Shot 2020-02-20 at 4 56 59 PM

### Reason for Changes

Rendering bug fix.

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -59,6 +59,8 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
   this.nextConnection = null;
   /** @type {Blockly.Connection} */
   this.previousConnection = null;
+  /** @type {boolean} */
+  this.hasStatementInput = false;
   /** @type {!Array.<!Blockly.Input>} */
   this.inputList = [];
   /** @type {boolean|undefined} */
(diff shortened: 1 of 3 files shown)


core/block.js
```diff
@@ -1635,6 +1637,9 @@ Blockly.Block.prototype.appendInput_ = function(type, name) {
   if (type == Blockly.INPUT_VALUE || type == Blockly.NEXT_STATEMENT) {
     connection = this.makeConnection_(type);
   }
+  if (type == Blockly.NEXT_STATEMENT) {
```
rachel-fenichel: If you're doing this here, you need to also check again and possibly set it to false in `removeInput`

The other option is to figure out whether there are statement inputs in the renderer.  iirc the rows know it, but the top-level info object doesn't.

samelhusseini: Agree to using a counter instead of a boolean.
Renderer option won't work without a look-ahead as the info we need is in the top row before the other rows and inputs are created / added.

core/block.js
```diff
@@ -1711,6 +1721,9 @@ Blockly.Block.prototype.moveNumberedInputBefore = function(
  */
 Blockly.Block.prototype.removeInput = function(name, opt_quiet) {
   for (var i = 0, input; (input = this.inputList[i]); i++) {
+    if (input.type == Blockly.NEXT_STATEMENT) {
+      this.statementInputs--;
+    }
     if (input.name == name) {
       input.dispose();
       this.inputList.splice(i, 1);
```
alschmiedt: Should we only decrease statementInputs count when we dispose of an input?

samelhusseini: Whoops, yes! Good catch

PR #3708, opened 2020-02-21
Add workspace comment context menu option to playground

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Adds a context menu option to the playground so that you can test workspace comments.

### Reason for Changes

I was doing some testing with workspace comments earlier and I think it is handy to be able to create one through right-clicking instead of having to load xml.

### Test Coverage

This is for testing silly!

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/workspace_comment_svg.js
+++ b/core/workspace_comment_svg.js
@@ -24,7 +24,6 @@ goog.require('Blockly.utils.dom');
 goog.require('Blockly.utils.object');
 goog.require('Blockly.utils.Rect');
 goog.require('Blockly.WorkspaceComment');
-goog.require('Blockly.WorkspaceCommentSvg.render');
 
 
 /**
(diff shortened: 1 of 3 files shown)


tests/playground.html
```diff
@@ -465,6 +470,21 @@
     }
   };
   menuOptions.push(screenshotOption);
+
+  // Adds a default-sized workspace comment to the workspace.
+  var workspaceCommentOption = {
```
samelhusseini: Instead of creating a completely new option for the playground, can we make use of: 
```
Blockly.ContextMenu.workspaceCommentOption
```

core/workspace_comment_svg.js
```diff
@@ -24,6 +24,7 @@ goog.require('Blockly.utils.dom');
 goog.require('Blockly.utils.object');
 goog.require('Blockly.utils.Rect');
 goog.require('Blockly.WorkspaceComment');
+goog.require('Blockly.WorkspaceCommentSvg.render');
```
samelhusseini: This was deliberately left out so that others can extend the rendering part of a workspace comment. Previously, BlockSvg had the same structure.
You can just require the WorkspaceCommentSvg.render in the playground.

tests/playground.html
BeksOmega: Totally didn't know this existed! Thanks for the tip :D

Issue #3700, opened 2020-02-19
Rendering RTL collapsed blocks in geras error


moniika:
**Describe the bug**

Path error during rendering of collapsed RTL blocks in geras.

**To Reproduce**

1. Go to playground
2. Set renderer to geras and blocks to RTL
3. Pull out a block and collapse it
4. See error in console

**Expected behavior**

There should not be any error and in some cases the highlight path is rendered very incorrectly.

**Screenshots**

![image](https://user-images.githubusercontent.com/6621618/74878142-93603680-531b-11ea-98be-a8131ddb1bda.png)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: chrome

**Stack Traces**

```
Error: <path> attribute d: Expected number, "…8  l 5.1,2.6  v NaN  H 82.399490…".
  | Blockly.geras.PathObject.setHighlightPath | @ | path_object.js:113
  | Blockly.geras.Drawer.draw | @ | drawer.js:64
  | Blockly.blockRendering.Renderer.render | @ | renderer.js:242
  | Blockly.BlockSvg.render | @ | block_svg.js:1659
  | Blockly.BlockSvg.setCollapsed | @ | block_svg.js:660
...
```

samelhusseini:
Fixed.

PR #3709, opened 2020-02-21
Fix geras collapsed rendering incorrect in RTL

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#3700 

### Proposed Changes

* Add height and width to geras jaggedTeeth constants object.
* Remove horizontal-line-to step.

### Reason for Changes

* These being undefined causes NaN errors if/when they are accessed.
* This step was causing errors for blocks with output notches, and seems to be unneccessary.

### Test Coverage

CollapsedBlocksGeras

Tested on:
 * Desktop Chrome 

### Documentation

N/A

### Additional Information

N/A


samelhusseini:
Thanks @BeksOmega

Issue #3710, opened 2020-02-22
Display flex breaks measureFontMetrics


BeksOmega:
**Describe the bug**

I was working in my custom playground and I noticed that the text was slightly off:
![TextOff](https://user-images.githubusercontent.com/25440652/75099132-09120f80-5573-11ea-9462-1cad6271d6a0.png)
It turned out this was because the baseline was being calculated incorrectly by measureFontMetrics.

I believe this is because my body is set to `display: flex` (as that makes it reproduce in the playground) but I'm not sure.

**To Reproduce**

Option A:
1. Set body to display:flex.
2. Open playground and view blocks.
![Flex1](https://user-images.githubusercontent.com/25440652/75099059-2397b900-5572-11ea-81a3-82c2b2112239.png)

Option B:
1. Grab [my custom playground](https://pastebin.com/fXQeBGwN).
2. Open the playground and view blocks.
![flex2](https://user-images.githubusercontent.com/25440652/75099089-707b8f80-5572-11ea-98be-5b9d28d5102a.png)

**Expected behavior**

Text should be positioned correctly no matter what.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser Chrome
 - Version 80

**Additional context**

I use a custom playground because I have a tiny screen, and the default one doesn't flex.

samelhusseini:
Fixed in #3714.

PR #3714, opened 2020-02-25
Use flex to compute measure font metrics

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3710

### Proposed Changes

Use ``display:flex`` and align-items when computing the height and baseline of text (based on a specific font) rather than inline-block with vertical-align. 

### Reason for Changes

Fix bug when user is using flex.

### Test Coverage

Tested on all major browsers with both display:block and display:flex

Tested on:
* Desktop Chrome
* Desktop Firefox
* Desktop Safari
* Windows IE 11

### Additional Information

@BeksOmega FYI

### DIFF
--- a/core/utils/dom.js
+++ b/core/utils/dom.js
@@ -347,25 +347,25 @@ Blockly.utils.dom.measureFontMetrics = function(text, fontSize, fontWeight,
     fontFamily) {
 
   var span = document.createElement('span');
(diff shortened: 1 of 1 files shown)


BeksOmega:
Works perfectly! Thank you so much!

PR #3721, opened 2020-03-05
Flash on render

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Flash a block red for 100ms when the debugger is included and enabled so we can tell when a block is re-rendered.

### Reason for Changes

Help with render performance testing.

### Test Coverage

Help with testing.

Tested on:

### Additional Information

flash_on_render


samelhusseini:
True that is also useful, but I find this to be more specific to rendering the block. Paint flashing is more verbose.

PR #3724, opened 2020-03-06
[DEMO NO MERGE] Decouple procedure callers from mutator UI implemenation

BeksOmega:
### Basics

This is just a demo showing how the procedure callers could be generalized so that they are no longer tightly coupled to Blockly's current mutator-ui implimentation.

This is not for merging, and will be closed immediately after it is opened.

#### Background

The code I am talking about is inside [setProcedureParameters_](https://github.com/google/blockly/blob/master/blocks/procedures.js#L650).

Procedure callers currently rely on the IDs of blocks in the mutator, which means that if the mutator UI is not the default UI procedure callers break.

Procedure callers use the IDs of the blocks as keys in a dictionary. Because each block is associated with a parameter of the function, they are by extension associated with inputs on the caller. Because of this the caller can use the IDs of mutator blocks as keys for the blocks connected to its inputs.

But again, if the mutator UI doesn't have blocks, it has no way to keep track of what blocks should be connected to what inputs.

#### Solution

Instead of using block IDs as keys this solution uses variable IDs as keys. This works because each variable ID is also associated with a parameter of the function, and by extension an input on the caller.

This implimentation also has the advantage that we can get rid of the onFinishedEditing hack. Because each mutatorarg sets the id of its variable explicitly, it can simply rename its var as its text input is edited, instead of creating a bunch of new variables that have to be deleted in onFinishEditing.

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -99,19 +99,13 @@ Blockly.Blocks['procedures_defnoreturn'] = {
    * @return {!Element} XML storage element.
    * @this {Blockly.Block}
    */
-  mutationToDom: function(opt_paramIds) {
+  mutationToDom: function() {
     var container = Blockly.utils.xml.createElement('mutation');
-    if (opt_paramIds) {
-      container.setAttribute('name', this.getFieldValue('NAME'));
-    }
-    for (var i = 0; i < this.argumentVarModels_.length; i++) {
+    container.setAttribute('name', this.getFieldValue('NAME'));
+    for (var i = 0; i < this.arguments_.length; i++) {
       var parameter = Blockly.utils.xml.createElement('arg');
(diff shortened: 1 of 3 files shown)


BeksOmega:
This is not for merging, it is just a demo.

rachel-fenichel:
Do you want me to take a look at the code in here, even though there's a reason this approach doesn't work on its own?

BeksOmega:
> Do you want me to take a look at the code in here, even though there's a reason this approach doesn't work on its own?

You don't need to. I think the general idea of having a param ID that is separate from the mutatorarg block ID is /probably/ sound. But the implementation would need to be a lot different to preserve the create-a-bunch-of-vars behavior.

If you think there's something to be learned from it feel free to take a look. But otherwise it was just an idea hehe.

Issue #3113, opened 2019-09-27
Theme configuration.


samelhusseini:
**What UI components should be configurable by the theme, and what should the configuration name be.**

- [x] `workspace`: The workspace background color
- [x] `toolbox`: Toolbox background color
- [x] `toolboxText`: Toolbox category text color
- [x] `flyout`: Flyout background color
- [x] `flyoutText`: Flyout label text color
- [x] `flyoutOpacity`: Flyout opacity
- [x] `scrollbar`: Scrollbar color
- [x] `scrollbarOpacity`: Scrollbar opacity
- [x] `insertionMarker`: Insertion marker color
- [x] `insertionMarkerOpacity`: Insertion marker opacity
- [ ] `flyoutButton`: Background color for flyout buttons
- [ ] `fieldOpacity`: The opacity of the background of a field.
- [ ] `fieldColor`: The background color for the field.
- [ ] `highlight`: The highlight color of the selected block **(in progress)**
- [ ] `dropdown`: Default dropdown div background color
- [ ] `dropdownBorder`: Default dropdown div border
- [ ] `dropdownText`: Default dropdown div text. (or whether or not it's inverted)
- [x] `font`: The font for toolbox and blocks. (Could be useful for people with dyslexia)
- [x] `startHats`: Whether or not to enable start hats for blocks with no previous or output connections.

Using this issue to track proposals, please feel free to add to this list.

@alschmiedt FYI

samelhusseini:
Another extension point to think about is the ability to add custom filters. Currently all our filters are just injected in inject.js. Developers can't extend that though. It would be good to standardize that.
Even if that's not taken care of on a theme by theme basis.

rachel-fenichel:
Do we really want to encourage the use of custom filters?  They're very browser-dependent and can really slow things down if we're not careful.

alschmiedt:
Per discussion we have new and improved names:

workspace -> workspaceBackgroundColour
toolboxText -> toolboxForegroundColour
flyout -> flyoutBackgroundColour
flyoutText -> flyoutForegroundColour
scrollbar -> scrollbarColour
insertionMarker -> insertionMarkerColour
flyoutButton -> flyoutButtonColour
dropdown -> dropdownBackgroundColour
dropdownText -> dropdownForegroundColour

*Also discussed in the meeting, we are mostly sticking with our CSS as default except for a few cases where it makes sense to have javascript as default. (ex. selected)

rachel-fenichel:
[...]

PR #3731, opened 2020-03-10
Support extending themes with a base theme property.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Related to https://github.com/google/blockly/issues/3113

### Proposed Changes

```
Blockly.Theme.defineTheme(name, {
   'base': Blockly.Themes.Classic,
   'blockStyles': ...,
   'categoryStyles': ...,
   'fontStyles': ...
});
```

Developers can call the factory themselves when defining a theme, before passing in the theme object to the workspace options. In addition, if a developer is authoring a theme in workspace options in JSON, we would pass those options to this factory instead of doing this.

The defineTheme method would read the base property if it exists, and complete a deep merge of the base theme object's properties with the properties passed into the method and return a new Blockly.Theme.

Removed Dark theme CSS, and got around the need to specify a hover color by using an rgba value instead of a specific colour.

### Reason for Changes

Support developers in extending themes.

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/theme.js
+++ b/core/theme.js
@@ -40,24 +40,28 @@ Blockly.Theme = function(name, opt_blockStyles, opt_categoryStyles,
   /**
    * The block styles map.
    * @type {!Object.<string, !Blockly.Theme.BlockStyle>}
+   * @pacakge
    */
   this.blockStyles = opt_blockStyles || Object.create(null);
 
   /**
    * The category styles map.
    * @type {!Object.<string, Blockly.Theme.CategoryStyle>}
+   * @pacakge
    */
   this.categoryStyles = opt_categoryStyles || Object.create(null);
 
   /**
    * The UI components styles map.
    * @type {!Object.<string, *>}
+   * @pacakge
    */
   this.componentStyles = opt_componentStyles || Object.create(null);
 
   /**
    * The font style.
    * @type {Blockly.Theme.FontStyle}
+   * @pacakge
    */
   this.fontStyle = /** @type {Blockly.Theme.FontStyle} */ (Object.create(null));
 
@@ -65,6 +69,7 @@ Blockly.Theme = function(name, opt_blockStyles, opt_categoryStyles,
    * Whether or not to add a 'hat' on top of all blocks with no previous or
    * output connections.
    * @type {boolean}
+   * @pacakge
    */
   this.startHats = false;
 };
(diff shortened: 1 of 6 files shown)


core/theme.js
```diff
@@ -13,65 +13,60 @@ goog.provide('Blockly.Theme');
...
-  this.categoryStyles = categoryStyles;
+  this.categoryStyles = opt_categoryStyles || Object.create(null);
 
   /**
    * The UI components styles map.
    * @type {!Object.<string, *>}
-   * @private
    */
-  this.componentStyles_ = opt_componentStyles || Object.create(null);
+  this.componentStyles = opt_componentStyles || Object.create(null);
```
alschmiedt: Why are we making this public?

core/theme.js
```diff
@@ -161,5 +156,32 @@ Blockly.Theme.prototype.getComponentStyle = function(componentName) {
...
 Blockly.Theme.prototype.setComponentStyle = function(componentName,
     styleValue) {
-  this.componentStyles_[componentName] = styleValue;
+  this.componentStyles[componentName] = styleValue;
+};
+
+/**
+ * Define a new Blockly theme.
+ * @param {string} name The name of the theme.
+ * @param {Object} themeObj An object containing theme properties.
```
alschmiedt: Should be !Object, or add a check for null themeObj.

core/theme/dark.js
```diff
@@ -116,39 +26,3 @@ Blockly.Themes.Dark.setComponentStyle('flyoutForegroundColour', '#ccc');
 Blockly.Themes.Dark.setComponentStyle('flyoutOpacity', 1);
 Blockly.Themes.Dark.setComponentStyle('scrollbarColour', '#797979');
 Blockly.Themes.Dark.setComponentStyle('scrollbarOpacity', 0.4);
```
alschmiedt: Could we add these to the object we pass into the define theme so we have an example of best way to use it?

samelhusseini: Great idea!

core/theme.js
samelhusseini: I guess it could technically remain package, but certainly not protected because its being accessed from Blockly.Theme.defineTheme which is a static class method.

samelhusseini: Done.

PR #3732, opened 2020-03-10
Support multiple renderer / theme combos on the same page.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

If we have multiple renderers with different themes, they begin to interfere with each other as all renderers of the same type share the CSS.

### Proposed Changes

Making it so that each renderer/theme combo share CSS on the page, and no renderer disposes of the CSS when it is disposed (only clears reference to it). This is because we don't know if another workspace is making use of it.

### Reason for Changes

Noticed in design meetings that multi workspaces with differing renderers / themes was broken.

### Test Coverage

Tested in multi playground.

Tested on:

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -155,10 +155,12 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
   svg.appendChild(mainWorkspace.createDom('blocklyMainBackground'));
 
   // Set the theme name and renderer name onto the injection div.
+  var rendererName = mainWorkspace.getRenderer().name;
+  var themeName = mainWorkspace.getTheme().name;
   Blockly.utils.dom.addClass(mainWorkspace.getInjectionDiv(),
-      (wsOptions.renderer || 'geras') + '-renderer');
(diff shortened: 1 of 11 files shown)


samelhusseini:
Ping.

core/inject.js
```diff
@@ -155,10 +155,12 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
   svg.appendChild(mainWorkspace.createDom('blocklyMainBackground'));
 
   // Set the theme name and renderer name onto the injection div.
+  var rendererName = mainWorkspace.getRenderer().name;
+  var themeName = mainWorkspace.getTheme().name;
   Blockly.utils.dom.addClass(mainWorkspace.getInjectionDiv(),
-      (wsOptions.renderer || 'geras') + '-renderer');
+      rendererName + '-' + themeName + '-renderer');
```
moniika: With this change, do we need to add additional logic to ThemeManager.setTheme to update class combo?

samelhusseini: Good point, I'll look into that.

Issue #3653, opened 2020-01-24
Add hat to block definition


alschmiedt:
**Is your feature request related to a problem? Please describe.**
The only way to add a hat to a block is either overriding the renderer (adds a hat to all blocks that fit the criteria) or add it to a style. The problem with having it in style is that you have to add two styles with the same colours. 

**Describe the solution you'd like**
Add hat to the block definition.

**Describe alternatives you've considered**

**Additional context**
https://groups.google.com/g/blockly/c/IE4197b9LMY

mattbishop:
StartHat is a feature of all renderers, and it needs a nice, simple way to turn it on and off. Perhaps it could be added with WorkspaceOptions, or better yet, add a new RenderOptions to workspaceOptions to override all the options that ConstantsProvider supplies? I really like Zelos, but there are a few of the sizes I'm going to want to change. I shouldn't have to create a new Renderer class just to tune an existing Renderer.

samelhusseini:
Very much agree with your sentiment @mattbishop, and we should be focusing on ways to make it easier to configure renderers for the next release :)

moniika:
Start Hats can also be added through themes, which is a bit easier.
However, as Sam mentioned, we are actively working on making it easier to configure the rendering constants.

rachel-fenichel:
We resolved this with Sam's PR #3734, which makes it possible to pass renderer overrides in options. For the broader case ("I shouldn't have to create a new Renderer class just to tune an existing Renderer"), we decided that you *do* have to extend the renderer to tune it, and added a [codelab](https://blocklycodelabs.dev/codelabs/custom-renderer/index.html?index=..%2F..index#0) explaining how to do it. There are too many ways to customize a renderer for us to add APIs to satisfy everyone, and we do not want to add them piecewise.

PR #3734, opened 2020-03-10
Support passing renderer overrides in options

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Request on the forums and in https://github.com/google/blockly/issues/3653

### Proposed Changes

The ability to pass in workspace options ``rendererOverrides``. 
The properties passed in are applied when a new renderer constants is created. Before initialization, the property bag is mixed in to override whichever properties the developers wants.

As part of this change also: 
- Renderers create a new constants provider instance when refreshing a theme
- Fields no longer keep a reference to the rendering constants, instead call into the workspace for that.

### Reason for Changes

Allow developers to override rendering constants without the need to create a new renderer.

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/renderers/common/renderer.js
+++ b/core/renderers/common/renderer.js
@@ -44,6 +44,13 @@ Blockly.blockRendering.Renderer = function(name) {
    * @private
    */
   this.constants_ = null;
+
+  /**
+   * Rendering constant overrides, passed in through options.
+   * @type {?Object}
+   * @package
+   */
+  this.overrides = null;
 };
 
 /**
@@ -56,7 +63,7 @@ Blockly.blockRendering.Renderer.prototype.init = function(theme,
     opt_rendererOverrides) {
   this.constants_ = this.makeConstants_();
   if (opt_rendererOverrides) {
-    this.constants_.overrides = opt_rendererOverrides;
+    this.overrides = opt_rendererOverrides;
     Blockly.utils.object.mixin(this.constants_, opt_rendererOverrides);
   }
   this.constants_.setTheme(theme);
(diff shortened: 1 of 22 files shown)


core/renderers/common/renderer.js
```diff
@@ -49,10 +49,16 @@ Blockly.blockRendering.Renderer = function(name) {
...
  * @param {!Blockly.Theme} theme The workspace theme object.
+ * @param {Object=} opt_rendererOverrides Rendering constant overrides.
  * @package
  */
-Blockly.blockRendering.Renderer.prototype.init = function(theme) {
+Blockly.blockRendering.Renderer.prototype.init = function(theme,
+    opt_rendererOverrides) {
   this.constants_ = this.makeConstants_();
+  if (opt_rendererOverrides) {
+    this.constants_.overrides = opt_rendererOverrides;
```
rachel-fenichel: You're setting this directly from out here instead of changing `makeConstants_`.  Is that to avoid/minimize API changes?

rachel-fenichel: And if so, is it actually helping?  The developer still needs to modify their override of `renderer.init` and `renderer.refresh`, and now  needs to remember to add the overrides in the correct order.

samelhusseini: Yes

samelhusseini: To your second comment, init is our entry point from the workspace, and we don't have other any other option for passing renderer overrides. 

I don't imagine overriding the init method would be common, but if a developer does (as we do in geras), I would expect a super call. 
Moving this logic to ``makeConstants_`` would have the same problem if not worse, since its a factory method for creating the constants provider and that is way more common of an override.

rachel-fenichel: Reasoning sounds good.

LGTM to this PR

PR #3735, opened 2020-03-11
Adjust marker and cursor colours from the theme

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/3113

### Proposed Changes

Add an accessibilityStyle property bag in the theme that adjusts the respective properties in renderer constants when the theme is set.

Redraw the attached marker / cursor when a field / block / workspace is re-rendered.

### Reason for Changes

Support theme configuration for accessibility marker / cursor colours. That's quite important if we're allowing developers to change the workspace background.

We weren't currently updating the a field when its size changed. Similarly we weren't updating a block's marker if it is re-rendered. 

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1664,14 +1664,14 @@ Blockly.BlockSvg.prototype.render = function(opt_bubble) {
   }
   Blockly.utils.dom.stopTextWidthCache();
 
-  this.updateAccessibility_();
+  this.updateMarkers_();
 };
 
 /**
  * Redraw any attached marker or cursor svgs if needed.
  * @protected
  */
-Blockly.BlockSvg.prototype.updateAccessibility_ = function() {
+Blockly.BlockSvg.prototype.updateMarkers_ = function() {
   if (this.workspace.keyboardAccessibilityMode && this.pathObject.cursorSvg) {
     this.workspace.getCursor().draw();
   }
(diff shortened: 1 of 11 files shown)


core/block_svg.js
```diff
@@ -1664,9 +1664,19 @@ Blockly.BlockSvg.prototype.render = function(opt_bubble) {
...
-  if (this.workspace.keyboardAccessibilityMode && this.pathObject.cursorSvg_) {
-    cursor.draw();
+  this.updateAccessibility_();
+};
+
+/**
+ * Redraw any attached marker or cursor svgs if needed.
+ * @protected
+ */
+Blockly.BlockSvg.prototype.updateAccessibility_ = function() {
```
alschmiedt: I would rather we use updateMarkers_ instead of updateAccessibility. Markers can be used outside of accessibility purposes and I don't feel like the name is very descriptive of what is actually happening.

core/theme.js
```diff
@@ -60,18 +60,26 @@ Blockly.Theme = function(name, opt_blockStyles, opt_categoryStyles,
...
    */
-  this.startHats = false;
+  this.startHats = null;
+
+  /**
+   * The accessibility style.
+   * @type {!Blockly.Theme.AccessibilityStyle}
+   * @package
+   */
+  this.accessibilityStyle =
```
alschmiedt: Same idea as the above comment.

samelhusseini: In which case maybe it can live in componentStyle then.

core/block_svg.js
samelhusseini: Done.

PR #3740, opened 2020-03-12
Fix block connected bounds for Zelos and minimalist renderers.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Connected block bounds are incorrect for Zelos and minimalist renderers.
Screen Shot 2020-03-12 at 12 18 57 am

This causes all sorts of issues with keyboard nav markers and types of blocks that factor in a statement block's width (eg: value to statement blocks).

### Proposed Changes

Fix right connection measurable width addition logic in Zelos. Revert finalize_ measurement of widthWithConnection to be consistent with other renderers. 

I also began setting the widthWithConnectedBlocks of the top and bottom row to be the same as the row's width so that the entire block width is taken into consideration when measuring widthWithConnectedBlocks.

### Reason for Changes

Fix rendering.

### Test Coverage

Tested in playground with repeat and value to statement blocks.

Tested on:

### Additional Information

Fixed: 
Screen Shot 2020-03-12 at 12 17 13 am

### DIFF
--- a/core/renderers/zelos/info.js
+++ b/core/renderers/zelos/info.js
@@ -377,14 +377,12 @@ Blockly.zelos.RenderInfo.prototype.finalizeOutputConnection_ = function() {
       this.outputConnection.shape.connectionOffsetX(connectionWidth);
 
   // Adjust right side measurable.
-  var rightConnection = false;
   this.startX = connectionWidth;
   if (!this.hasStatementInput && !this.bottomRow.hasNextConnection) {
     this.rightSide.height = connectionHeight;
     this.rightSide.width = connectionWidth;
     this.rightSide.centerline = connectionHeight / 2;
     this.rightSide.xPos = this.width + connectionWidth;
-    rightConnection = true;
     connectionWidth *= 2;
   }
   this.width += connectionWidth;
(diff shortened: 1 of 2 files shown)


core/renderers/zelos/info.js
```diff
@@ -376,16 +377,16 @@ Blockly.zelos.RenderInfo.prototype.finalizeOutputConnection_ = function() {
...
 
   // Adjust right side measurable.
-  if (!this.hasStatementInput) {
+  this.startX = connectionWidth;
+  if (!this.hasStatementInput && !this.bottomRow.hasNextConnection) {
     this.rightSide.height = connectionHeight;
     this.rightSide.width = connectionWidth;
     this.rightSide.centerline = connectionHeight / 2;
     this.rightSide.xPos = this.width + connectionWidth;
+    connectionWidth *= 2;
```
alschmiedt: Could we add some comments for this section? For example why are we multiplying connectionWidth by 2? And why do we care if it has a next connection or not?

samelhusseini: Okay and will simplify the logic to make it clearer.

PR #3742, opened 2020-03-12
Add non-nullable modifier to return type of functions never returning null

vrana:
- [x] I branched from develop


core/flyout_base.js
```diff
@@ -629,7 +629,7 @@ Blockly.Flyout.prototype.isBlockCreatable_ = function(block) {
 /**
  * Create a copy of this block on the workspace.
  * @param {!Blockly.BlockSvg} originalBlock The block to copy from the flyout.
- * @return {Blockly.BlockSvg} The newly created block, or null if something
+ * @return {!Blockly.BlockSvg} The newly created block, or null if something
```
samelhusseini: Hmm, the description says it returns null if something goes wrong in deserialization. Looking at the code, if the try fails around ``placeNewBlock_``, the method returns null.

PR #3747, opened 2020-03-13
Fix two memory leaks

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

2 memory leaks

### Proposed Changes

Connection.js: 
Generally connections are disconnected in block_svg's dispose method before they are disposed (eg: dragging a block into the trash can). Only setting the shadowDom_ to null if its connected does not suffice, and leaves around detached elements.

Block_svg.js
A path object maintains a reference to a block's svgRoot. Not disposing of the path object leaves around detached nodes.

### Reason for Changes

Performance.

### Test Coverage

Tested with the chrome memory tools.

Tested on:

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -291,6 +291,7 @@ Blockly.BlockDragger.prototype.maybeDeleteBlock_ = function() {
     // Fire a move event, so we know where to go back to for an undo.
     this.fireMoveEvent_();
     this.draggingBlock_.dispose(false, true);
+    Blockly.draggingConnections = [];
   } else if (trashcan) {
     // Make sure the trash can is closed.
     trashcan.close();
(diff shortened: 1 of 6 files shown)


rachel-fenichel:
I thought that garbage disposal was supposed to handle this cleanly.  If all references to a block disappear, any reference it holds to a bit of the DOM should also disappear.  Do you know why that's not happening in this case?

samelhusseini:
Okay I dug deeper and this was only dealing with the side effects of the memory leak but not the actual cause. The cause was two part: 
- **DraggedConnections:** When we delete blocks, we don't clear Blockly.draggedConnections which maintains a reference to the source block and everything else.
- **Tooltips:**: 
  - Firstly when tooltips are disposed we should clear the element_ and poisonedElement_ references otherwise they keep a reference to the svg path element. 
  - When we create a block, we create a cyclic reference block -> svgPath -> block when we do [this](https://github.com/google/blockly/blob/master/core/block_svg.js#L104), we should remove that reference when we dispose
  - Additionally, we create a similar reference in [here](https://github.com/google/blockly/blob/master/core/flyout_base.js#L723) in the flyout which needs to also be removed.

samelhusseini:
Just for the record, the tooltip issue was actually that the tooltip listeners were not being unbound when a block / field / flyout mat were disposed.

PR #3750, opened 2020-03-16
[DEMO NO MERGE] Decouple procedure callers from mutator UI implemenation Mk2

BeksOmega:
## Basics
This is just a demo showing how the procedure callers could be generalized so that they are no longer tightly coupled to Blockly's current mutator-ui implimentation.

This is not for merging, and will be closed immediately after it is opened.

### Background
See #3724 

### Solution
This solution gives each parameter an ID that is separate from both the variable ID and the mutator block ID. It is very similar to the previous demo, but this one avoids coalescing vars accidentally.

This implementation brings back onFinishEditing. It's not necessarily necessary, you could handle deleting the previous var inside the validator. But I thought this was easier.

---
I think this is a pretty good for a demo, but if this was actually for PR there are still things I would want to clean up. Deleting arguments sometimes leaves unused vars in the flyout (although I think this happens on master). And I would want to explore refactoring `mutateCallers`.

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -99,19 +99,13 @@ Blockly.Blocks['procedures_defnoreturn'] = {
    * @return {!Element} XML storage element.
    * @this {Blockly.Block}
    */
-  mutationToDom: function(opt_paramIds) {
+  mutationToDom: function() {
     var container = Blockly.utils.xml.createElement('mutation');
-    if (opt_paramIds) {
-      container.setAttribute('name', this.getFieldValue('NAME'));
-    }
-    for (var i = 0; i < this.argumentVarModels_.length; i++) {
+    container.setAttribute('name', this.getFieldValue('NAME'));
+    for (var i = 0; i < this.arguments_.length; i++) {
       var parameter = Blockly.utils.xml.createElement('arg');
-      var argModel = this.argumentVarModels_[i];
-      parameter.setAttribute('name', argModel.name);
-      parameter.setAttribute('varid', argModel.getId());
-      if (opt_paramIds && this.paramIds_) {
-        parameter.setAttribute('paramId', this.paramIds_[i]);
-      }
+      parameter.setAttribute('name', this.arguments_[i]);
+      parameter.setAttribute('varid', this.varIds_[i]);
       container.appendChild(parameter);
     }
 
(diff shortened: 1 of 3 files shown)


BeksOmega:
This is not for merging, it is just a demo.

blocks/procedures.js
```diff
@@ -128,19 +127,27 @@ Blockly.Blocks['procedures_defnoreturn'] = {
...
         var varId = childNode.getAttribute('varid') || childNode.getAttribute('varId');
+        this.varIds_.push(varId);
         this.arguments_.push(varName);
         var variable = Blockly.Variables.getOrCreateVariablePackage(
             this.workspace, varId, varName, '');
         if (variable != null) {
           this.argumentVarModels_.push(variable);
         } else {
-          console.log('Failed to create a variable with name ' + varName + ', ignoring.');
+          console.log('Failed to create a variable with name ' +
```
rachel-fenichel: There's a pre-existing problem here, but it's worth addressing if you're changing this code: If no variable can be created we still add to most of the arrays (`arguments_`, `varIds_`, `paramIds_`) but not to `argumentVarModels_`.  That means they're out of sync permanently.

core/field.js
```diff
@@ -852,7 +852,10 @@ Blockly.Field.prototype.setValue = function(newValue) {
     return;
   }
 
-  if (this.sourceBlock_ && Blockly.Events.isEnabled()) {
+  if (this.sourceBlock_ &&
+      !this.sourceBlock_.disposed &&
```
rachel-fenichel: Is this a change we need no matter what?

BeksOmega: Ah, yep I forgot I wanted to do that. Pull request at #3766

BeksOmega:
> Can I get a walkthrough on this? My understanding so far:

I think you've got it already haha

> * the goal is to change `setProcedureParameters` to not assume that the mutator UI has a stack of blocks

Correct : )

> * `paramId` is now separate from `varId`, but parallel

Yesh! This is a bit confusing because there's no concrete object that paramId is identifying. Really it's used in two cases:
1) We need to associate a new var with a position.
2) We need to change the positions of the vars.

If either one of these was static, we could use it as the ID instead. For example: If the position was static we could use that to associate new vars with each position. Or if the vars were static we could pass their IDs to the caller for reordering (which is what #3724 tried to do).

Because neither is static we need something else that is.

It would be nice if there was an clearer way to explain it though.

> * added `idsToConnections` for procedure caller blocks
> * `paramId` and `idsToConnections` are similar to the old `quarkIds` and `quarkConnections`

Yes and yes! All of this is so that when we handle case 2 the child blocks can get reconnected correctly.

blocks/procedures.js
BeksOmega: Good idea!

rachel-fenichel:
I'm interested in fixing this, but also really wary of PRs because we don't currently have tests. I just opened an issue (#3781) for procedure block tests, and I want to have tests in place before changing the actual block implementaiton.

Issue #3166, opened 2019-10-04
Sizing of text inputs fields in mutator workspaces is broken


samelhusseini:
**Describe the bug**
Text input fields in mutator workspace doesn't resize to the right size of the block. 

**To Reproduce**

1. Go to the playground.
2. Drag out a procedure block
3. Open the mutator dialog
4. Drag out an input x block
5. Click on the x field to change its name.

**Expected behavior**

The field's input box should resize to the size of the field. 

**Screenshots**

<img width="371" alt="Screen Shot 2019-10-04 at 2 36 18 PM" src="https://user-images.githubusercontent.com/16690124/66241801-bb286800-e6b4-11e9-9b1f-77e86bc12562.png">

Repro's on Safari and Chrome, probably others too.

**Additional context**

Affects RC/Oct_2019

samelhusseini:
Looks like you need to be zoomed in to repro this, so an issue with zoom as well.
Not a regression.

rachel-fenichel:
Makes sense--mutators aren't actually zoomed because they're inside the parent svg that was zoomed, and I can see that getting out of hand.

samelhusseini:
Fixed.

PR #3763, opened 2020-03-23
Fix mutator field scale

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3166

### Proposed Changes

Fix the scale used by fields inside of the mutator dialog. Mutators should borrow their scale from their parent.

### Reason for Changes

Bug fix.

### Test Coverage

Tested in playground.

Tested on:


rachel-fenichel:
This is kind of unfortunate--now we have a difference between the value of `scale` and `getScale`, and it's not always correct to use the getter instead of the value directly.

The conflict is because anything in the mutator svg is automatically scaled by the scale on the parent workspace, but text input widgets actually live outside of the SVG.

We don't have to fix it right now, but the `getScale` function needs a comment explaining when you would use it vs just getting the workspace's `scale` directly.

Issue #3758, opened 2020-03-20
Input fields unresponsive in fullscreen mode


bartbutenaers:
Hello everybody,

Last year I [integrated ](https://flows.nodered.org/node/node-red-contrib-blockly)blockly into the Node-RED IOT framework.  This works fine but our users have asked to be able to show the Blockly editor in _**full-screen**_ mode.  I have implemented that feature request in [plain](https://www.w3schools.com/howto/howto_js_fullscreen.asp) Javascript, and it works!  

But the problem is that (only in fullscreen mode) the input fields don't respond: 
+ When clicking on a dropdown, the dropdown doesn’t appear
+ When clicking on a color picker, the picker doesn’t appear
+ When clicking on an edit box, the text is not selected and you can not enter new characters
+ ...

In had this problem in my old Blockly version, but after installing the *current version from the Github master* branch it is still the same behaviour.  So I assume it is in all Blockly versions...

**To Reproduce**

I used the following standard code snippet to goto fullscreen mode, when an expand button is clicked:
```
var blocklyArea = $("#node-red-contrib-blockly-expandable");
var blocklyFullScreenRequest = blocklyArea[0].requestFullscreen || blocklyArea[0].mozRequestFullScreen || blocklyArea[0].webkitRequestFullscreen || blocklyArea[0].msRequestFullScreen;
// Check whether the browser supports fullscreen
if (blocklyFullScreenRequest) {
   // Expand the Blockly editor when the expand button has been clicked
   expandblocklyAreaButton.click(function(e) {
      // Make sure (on all browsers) that the blocklyArea DIV element has 100% AFTER it has become fullscreen
      // If document.fullscreenElement is null then full-screen mode has been quitted...
      blocklyArea.on('webkitfullscreenchange mozfullscreenchange msRequestFullscreen fullscreencha nge', function(e) {
         if (document.fullscreenElement) {
            // Make sure the blockly area fills the entire screen
            document.fullscreenElement.style.height = "100%";
         }
      });
                    
      // Expand the blocklyArea element to fullscreen
      blocklyFullScreenRequest.call(blocklyArea[0]);
   })
}
```
1. Show the Blockly editor in fullscreen mode.
2. Try to enter a text in a text input field, or try to open a dropdown input.
3. Nothing will happen ...

**Expected behavior**

I expect in fullscreen mode that I can enter data into my blocks.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome, Firefox, Edge

[...]

PR #3772, opened 2020-03-26
Support a custom container element for the dropdowndiv, widgetdiv and tooltip.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Resolves https://github.com/google/blockly/issues/3758

### Proposed Changes

Support a custom container element for the dropdowndiv, widgetdiv and tooltip.

### Reason for Changes

Allow developers to change where these divs get placed.

### Test Coverage

Tested in playground.

Tested on:

### Additional Information

@bartbutenaers

For full screen, you can support your scenario by: 

```
var blocklyArea = document.getElementById('blocklyArea');
Blockly.WidgetDiv.CONTAINER_ELEMENT = blocklyArea;
Blockly.DropDownDiv.CONTAINER_ELEMENT = blocklyArea;
Blockly.Tooltip.CONTAINER_ELEMENT = blocklyArea;
```

And just make sure you inject your workspace into ``blocklyArea``.
This makes sure everything ends up there, and then you can call ``fullscreen`` on the blockly area element.

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -95,7 +95,7 @@ Blockly.cache3dSupported_ = null;
  * @type {?Element}
  * @package
  */
-Blockly.PARENT_CONTAINER = null;
+Blockly.parentContainer = null;
 
 /**
  * Blockly opaque event data used to unbind events when using
@@ -673,8 +673,9 @@ Blockly.checkBlockColourConstant_ = function(
  * Set the parent container.  This is the container element that the WidgetDiv,
  * DropDownDiv, and Tooltip are rendered into the first time `Blockly.inject`
  * is called.
+ * This method is a NOP if called after the first ``Blockly.inject``.
  * @param {!Element} container The container element.
  */
 Blockly.setParentContainer = function(container) {
-  Blockly.PARENT_CONTAINER = container;
+  Blockly.parentContainer = container;
 };
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -91,6 +91,13 @@ Blockly.clipboardTypeCounts_ = null;
 Blockly.cache3dSupported_ = null;
 
 /**
+ * Container element to render the WidgetDiv, DropDownDiv and Tooltip.
+ * @type {?Element}
+ * @package
+ */
+Blockly.PARENT_CONTAINER = null;
+
+/**
  * Blockly opaque event data used to unbind events when using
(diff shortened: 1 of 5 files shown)


core/tooltip.js
```diff
@@ -111,6 +111,13 @@ Blockly.Tooltip.MARGINS = 5;
  */
 Blockly.Tooltip.DIV = null;
 
+/**
+ * Container element to render the tooltip into.
+ * @type {?Element}
+ * @const
```
rachel-fenichel: These aren't const if we expect them to get set externally.

samelhusseini: Removed.

samelhusseini:
I don't see why it wouldn't work for multiple instances: 

```
<BlocklyArea>
 <BlocklyWorkspace1 />
 <BlocklyWorkspace2 />
  <WidgetDiv />
  <DropdownDiv />
  <Tooltip />
</div>
```

This is just providing developers with some flexibility as to where they want us to render these supplemental components. 
I would agree that it would be better if we can find a solution that renders these components under the ``injectionDiv``. Although I see this as a good short term solution.

samelhusseini:
Discussed offline, decided to hoist the container up a level to the Blockly namespace.
Add a setter, eg: ``setParentContainer``. and pass that along to the dropdown / widget divs.

core/blockly.js
```diff
@@ -661,3 +668,13 @@ Blockly.checkBlockColourConstant_ = function(
     console.warn(warning);
   }
 };
+
+/**
+ * Set the parent container.  This is the container element that the WidgetDiv,
+ * DropDownDiv, and Tooltip are rendered into the first time `Blockly.inject`
+ * is called.
```
rachel-fenichel: Add a comment: this is a no-op if called after the first call to `inject`.

core/blockly.js
```diff
@@ -90,6 +90,13 @@ Blockly.clipboardTypeCounts_ = null;
  */
 Blockly.cache3dSupported_ = null;
 
+/**
+ * Container element to render the WidgetDiv, DropDownDiv and Tooltip.
```
rachel-fenichel: This should be lower case.  It's not a constant.

core/blockly.js
samelhusseini: Done.

core/blockly.js
samelhusseini: True.

Issue #3778, opened 2020-03-30
Flyout labels don't respect font size


samelhusseini:
**Describe the bug**

Flyout labels don't respect font size.

**To Reproduce**

1. Go to https://google.github.io/blockly/tests/playground.html?dir=ltr&toolbox=test-blocks&side=start&renderer=geras
2. Change the font using the font slider
3. Click on the Dropdown category
4. Notice how the flyout label hasnt changed font.

**Expected behavior**

Flyout label changes font to match theme font, same as flyout button.

**Screenshots**
![Screen Shot 2020-03-30 at 10 59 44 AM](https://user-images.githubusercontent.com/16690124/77946285-ead8b700-7276-11ea-8ba6-26cceaaad096.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**

PR #3780, opened 2020-03-30
Move flyout label css into the renderer

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3778

### Proposed Changes

Move flyout label css into the renderer and set the theme font options in CSS. 
This still allows users (in CSS) font properties they want to override, like so: 
https://github.com/google/blockly/blob/develop/tests/playground.html#L607
but it defaults to using the same properties as set on ``blocklyText``

### Reason for Changes

Bug fix, expect theme font changes to be reflected in flyout labels (similar to flyout buttons). 

### Test Coverage

Tested in playground.

Tested on:

### DIFF
--- a/core/renderers/common/constants.js
+++ b/core/renderers/common/constants.js
@@ -1165,13 +1165,17 @@ Blockly.blockRendering.ConstantProvider.prototype.injectCSS_ = function(
 Blockly.blockRendering.ConstantProvider.prototype.getCSS_ = function(selector) {
   return [
     /* eslint-disable indent */
-    // Fields.
+    // Text.
     selector + ' .blocklyText, ',
     selector + ' .blocklyFlyoutLabelText {',
-      'fill: #fff;',
       'font-family: ' + this.FIELD_TEXT_FONTFAMILY + ';',
       'font-size: ' + this.FIELD_TEXT_FONTSIZE + 'pt;',
       'font-weight: ' + this.FIELD_TEXT_FONTWEIGHT + ';',
+    '}',
+
+    // Fields.
+    selector + ' .blocklyText {',
+      'fill: #fff;',
     '}',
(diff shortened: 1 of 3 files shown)


core/renderers/common/constants.js
```diff
@@ -1166,7 +1166,8 @@ Blockly.blockRendering.ConstantProvider.prototype.getCSS_ = function(selector) {
   return [
     /* eslint-disable indent */
     // Fields.
-    selector + ' .blocklyText {',
+    selector + ' .blocklyText, ',
+    selector + ' .blocklyFlyoutLabelText {',
```
moniika: I would prefer that this additional selector `selector + ' .blocklyFlyoutLabelText` not be added here and instead
add font-family, font-size, font-weight to the block with the flyout labels selector you added on L1187. 

This would not change the css that is applied, but it would make the css rules cleaner. Currently the flyout label has fill colors #fff and #000 in the rules (with the second declared color being what is actually applied due to ordering). So it's harder to scan to see which the css applied. Also I think it would match the comments better.

Same comment applies to the `Zelos.getCSS_`

samelhusseini: Alternately I could do something like this: 
```
.blocklyText, .blocklyFlyoutLabelText { ... TEXT FONTS ... }

.blocklyText { fill: #000 }
.blocklyFlyoutLabelText { fill: #fff }
```

moniika: That seems fine.
With the alternative you're suggesting I would also recommend updating the comments in the CSS because at the moment the comments separate sections of CSS rules like `Fields` and I don't think that `blocklyFlyoutLabelText` would fit in `Fields`.

PR #3783, opened 2020-03-31
Remove mutator bump timeout

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Description

Anything that changes the shape of a block should be calling bumpNeighbors directly. For example:
* Append Input
* Add Comment
* Insert Field
* Etc

So this call is unnecessary. If this call /is/ making something bump that normally wouldn't, that should be treated as a bug, and the root should be made to call bumpNieghbours.

### Test Coverage

BumpingMutators

Tested on:
 * Desktop Chrome

### Additional Information

N/A


rachel-fenichel:
Sounds good.  We'll merge this after sending the release out the door at the end of the week.

PR #3805, opened 2020-04-09
Updating chrome version in selenium config.

moniika:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Proposed Changes

Update chrome version in selenium config.

### Reason for Changes

Fixes issue with running tests on newer versions of chrome.

PR #3808, opened 2020-04-10
Update ast node

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
While writing keyboard navigation documentation I found some parts of AST Node and Marker_Svg that I wanted to change.

### Proposed Changes
1. Blockly.navigation.getTopNode -> Blockly.ASTNode.createTopNode(block)
-- We already have a bunch of create methods in Blockly.ASTNode so seems like this is a better fit
2. Changed methods in marker_svg to be showWithBlock_, showWithPrevious, showWithOutput_ so that they are easier to override. They were previously all one function, which was an... interesting decision by past me.
3. Passed the curNode to applyColour_. Seems like it would be nice to have the option to change the colour of the cursor depending on the type of curNode. 
4. Random cleanup: Made sure everything was less than 80 characters, changed some wording in comments.

### Test Coverage
Added tests for createTopNode. 

Tested on:

### DIFF
--- a/core/renderers/common/marker_svg.js
+++ b/core/renderers/common/marker_svg.js
@@ -85,8 +85,7 @@ Blockly.blockRendering.MarkerSvg.MARKER_CLASS = 'blocklyMarker';
 /**
  * What we multiply the height by to get the height of the marker.
  * Only used for the block and block connections.
- * @type {number}
- * @const
+ * @const {number}
  */
 Blockly.blockRendering.MarkerSvg.HEIGHT_MULTIPLIER = 3 / 4;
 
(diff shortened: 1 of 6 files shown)


samelhusseini:
I think we're going to need to integrate https://github.com/google/blockly/pull/3805 to satisfy the travis god.

core/renderers/common/marker_svg.js
```diff
@@ -323,10 +411,12 @@ Blockly.blockRendering.MarkerSvg.prototype.positionBlock_ = function(
 /**
  * Position the marker for an input connection.
  * Displays a filled in puzzle piece.
- * @param {!Blockly.RenderedConnection} connection The connection to position marker around.
+ * @param {!Blockly.RenderedConnection} connection The connection to position
+ *     marker around.
  * @private
```
samelhusseini, line 416: Why is this one private and the rest protected.

alschmiedt: There were reasons but rethinking it seems better to make them all protected.

PR #3810, opened 2020-04-10
Fixing event tests and removing duplicate tests.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#2870

### Proposed Changes
Fixes bug with existing event tests and removes jsunit tests that are already covered by existing mocha tests.

### Reason for Changes
Explained in linked issue.

### Additional Information
Changes for converting existing jsunit tests to mocha with follow in a separate PR.


tests/mocha/event_test.js
```diff
@@ -332,14 +326,15 @@ suite('Events', function() {
       });
     });
 
-    suite.skip('Run Forward', function() {
+    suite('Run Forward', function() {
```
rachel-fenichel, line 329: Why did changing to new variable names fix this test and the next one?

moniika: Before, the tests "Run Forward":"Var create" and "Run Backward":"Var delete" were failing because they did not take into account that a variable is created in the setup method (L227 for all the tests in the "Variable events" suite).
They were always failing at the `assertNull` check that checked that a variable with "id1" does not exist.

The other tests depend on the instantiation of the variable in the setup method (set as `this.variable`), which is why I just instantiated a variable with a different id/name/type so there was no collision for these 2 tests.

PR #3812, opened 2020-04-10
Removed duplicate block tests and renamed tests.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#2870

### Proposed Changes
Removed jsunit tests for blocks as they were already covered in mocha tests. Additionally, renamed unit tests to be more clear in test results.

### Reason for Changes
Explained in linked issue.

### Test Coverage
Ran mocha tests.


tests/mocha/block_test.js
```diff
@@ -116,29 +116,29 @@ suite('Blocks', function() {
         blocks.B.unplug(true);
         assertUnpluggedHealFailed(blocks);
       });
-      test('A has multiple inputs', function() {
+      test('Parent has multiple inputs', function() {
```
alschmiedt, line 119: If we are going to change to talk about these as Parent, Middle, Child might be worth it to add a comment above explaining which ones are which.

tests/mocha/block_test.js
```diff
@@ -49,6 +49,29 @@ suite('Blocks', function() {
     delete Blockly.Blocks['statement_block'];
   });
 
+  function createTestBlocks(workspace, isRow) {
```
alschmiedt, line 52: I like the helper method. If all the tests use this.blocks could we just set that in the parent setup method?

moniika: I tried setting this.blocks, but it seems like the "this" in the function is not the same "this" as in the test suites.

alschmiedt: I think I meant just having this.block = createTestBlocks() in the parent setup function.

Issue #3213, opened 2019-10-09
Typings not split into files


samelhusseini:
The TypeScript typings aren't split into separate files, and so when a user imports ``'blockly/core'`` there isn't a typings files that's associated with it.

We should split up our blockly.d.ts file into a similar structure to our modules, thus that when a user imports a specific module, the typings are included for that subset of Blockly.

rachel-fenichel:
Fixed by #3821

PR #3821, opened 2020-04-15
Add typings modules to match UMD modules.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3213

### Proposed Changes

Add typings files for the rest of the blockly submodules: 
- blocks
- langs (javascript / python / etc)
- locales
- core
- index

### Reason for Changes

Someone doing: 
```
import * as Blockly from 'blockly/core';
```

Would still get the proper Typescript typings.

### Test Coverage

Tested with the angular sample.

Tested on:


typings/blocks.d.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
alschmiedt, line 3: Should this be updated to 2020

samelhusseini: Fixed.

yukiya383 (commented):
Since javascript.js exports Blockly.JavaScript, an **instance** of Blockly.Generator, javascript.d.ts should declare instance too.
Current declaration is below:
```typescript
import * as Blockly from './core';
export = Blockly.Generator;
```
Correct declaration is below:
```typescript
import * as Blockly from './core';
declare var tmp:Blockly.Generator;
export = tmp;
```
Other lang files should be fixed in the same way.
See #4742 for detail.

PR #3823, opened 2020-04-15
Add Selectable interface

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Add an interface for what a "Selectable" object is.
Right now Selectable and Copyable are the same thing, we could split them up but we don't really have much use for a selectable object that is not copyable.

### Test Coverage

Build passes.

Tested on:

### DIFF
--- a/core/interfaces/i_selectable.js
+++ b/core/interfaces/i_selectable.js
@@ -1,0 +1,65 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview The interface for an object that is selectable.
+ * @author samelh@google.com (Sam El-Husseini)
+ */
+
+'use strict';
+
+goog.provide('Blockly.ISelectable');
+
+goog.requireType('Blockly.WorkspaceSvg');
+
+
+/** @interface */
+Blockly.ISelectable = function() {};
+
+/**
+ * @type {string}
+ */
+Blockly.ISelectable.prototype.id;
+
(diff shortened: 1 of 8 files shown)


BeksOmega:
Related? https://github.com/google/blockly/issues/1667

core/interfaces/i_selectable.js
```diff
@@ -0,0 +1,65 @@
...
+ * @return {!Blockly.ISelectable.CopyData} Copy metadata.
+ */
+Blockly.ISelectable.prototype.toCopyData;
+
+/**
+ * Copy Metadata.
+ * @typedef {{
+ *            xml:!Element,
+ *            source:Blockly.WorkspaceSvg,
+ *            typeCounts:Object
```
rachel-fenichel, line 62: `typeCounts` doesn't fit with the general copyable/selectable interface.  it's specific to blocks.

samelhusseini: Hmm. I can make it optional, but the ``copy_`` method expects typeCounts. (even if it is in some cases).
Not saying the copy / pasting logic doesn't need refactoring to make it more extensible.

Any ideas on what else can be done here?

rachel-fenichel: Yeah, I think it should have been optional there too--and it feels very strange to encode that into the defined interface.

But LGTM on this change, with a note to look at the copy/paste logic more generally.

samelhusseini: Sounds good. Filed https://github.com/google/blockly/issues/3833 to track.

PR #3831, opened 2020-04-17
Converting jsunit asserts to chai.

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of changes with #2870

### Proposed Changes

- Converts usages of jsunit syntax to chai
- Moves jsunit syntax asserts out of mocha/test_helpers.js and into jsunit/test_utilities.js and removes corresponding eslint exceptions
- Fixes order of assert function calls to be ordered in actual then expected

### Reason for Changes

- Cleaning up the jsunit syntax makes the tests more consistent and in some cases easier to read, instead of relying on a mix of chai an jsunit
- Removing the helpers disincentivizes usage of jsunit syntax and makes other helper functions easier to find
- Ordering in actual then equal gives better test error messages

### Test Coverage

Ran mocha tests, manually reviewed changes, and conducted regex searches to look for common issues (like param order with optional message).

### Additional Information

Regex is magic.


rachel-fenichel:
Nice!

Issue #3834, opened 2020-04-17
Field Dropdown and Field Variable options opening mispositioned when inside a compose/decompose mutator


joseluiz98:
Hi guys!

I have a compose/decompose mutator with sub blocks that have either FieldDropdown or FieldVariable so user can select the correct option. But when I open this dropdowns, they are positioned ackward on the screen, always on the left side, instead of below its trigger like the html select.

**To Reproduce**

Just create a mutator with compose and decompose methods so the button trigger that opens the bubble will appear, like its mentioned on the docs. Then, pass a block that have a fieldDropdown in the args to this mutator, and open this dropdown inside the bubble.

**Expected behavior**

I expect the field dropdown options to appear below the trigger, the same behavior as the html select with the option array inside it.

**Screenshots**

![image](https://user-images.githubusercontent.com/24555974/79599494-f7a13b80-80bb-11ea-9ed0-579ce3a82afc.png)

samelhusseini:
Thanks for filing the bug @joseluiz98. We've been able to reproduce your problem and have a fix: https://github.com/google/blockly/pull/3835

PR #3835, opened 2020-04-17
Fix bug with dropdown divs in mutator dialogs.

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/3834

### Proposed Changes

Dropdown divs were using the tiny mutator workspace as their bounds. Use the root workspace as the bound instead.

### Reason for Changes

Bug fix.

### Test Coverage

Added test dropdown in mutator block and tested in playground.

Tested on:


joseluiz98:
Ok, great work, thank you guys! So users that uses a npm-based blockly installed in their projects like me, would have this fix by the next releases?

samelhusseini:
Yep, you can expect it in the upcoming release.

PR #3872, opened 2020-04-29
Add sourcemaps

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Adds sourcemaps to Blockly's compressed files and the UMD modules published to npm.
Removed hacks in build for adding goog.provides and then removing them using regex, instead using closure externs (see: ``externs/block-externs``, and ``externs/generator-externs``)
Made all our *_compressed files UMD modules themselves.

### Reason for Changes

Easier debugging for developers in general, and more specifically plugin developers.

### Test Coverage

Tested in demos/
Tested in plugins in blockly-samples/ with ``source-map-loader``.
Ran node tests

### Additional Information

The gulp plugin for the closure compiler supports gulp-sourcemaps but not all gulp plugins support them, this is why I've removed stripping out apache licenses from the build output as I wasn't able to find a gulp plugin that does regex replacing that also supported sourcemaps.
Delta is an additional 978 bytes for the license comments.
Here's my reasoning:
 - Sourcemaps are way more useful then stripping these comments out
 - If a user doesn't care about sourcemaps, they can easily write a script to strip these themselves.
 - If we had ∞ resources we could write a gulp plugin ourselves that did this and supported sourcemaps.
- Thanks to SPDX, our copyright content is much smaller.


samelhusseini:
Remaining tasks: 
- [ ] Look into closure compiler warning when using externs
- [ ] Look into whether or not we need to do anything so that ``source-map-loader`` picks up variable names.

scripts/gulpfiles/build_tasks.js
```diff
@@ -35,22 +35,16 @@ const licenseRegex = `\\/\\*\\*
 )? \\* SPDX-License-Identifier: Apache-2.0
  \\*\\/`;
 
-/**
- * Helper method for stripping the Google's and MIT's Apache Licenses.
- */
+/**	
+ * Helper method for stripping the Google's and MIT's Apache Licenses.	
+ */
```
alschmiedt, line 40: nit: Remove extra spaces

externs/generator-externs.js
```diff
@@ -0,0 +1,17 @@
...
+ * @fileoverview Externs for Blockly generators.
+ * @externs
+ */
+
+goog.provide('Blockly');
+goog.provide('Blockly.Generator');
+goog.provide('Blockly.utils.global');
+goog.provide('Blockly.utils.string');
+
+var Blockly;
```
alschmiedt, line 17: nit: new line

samelhusseini:
Note: I was able to strip the licenses afterall. Neat trick to add new lines in its place so it doesn't affect source maps lines.

Issue #2322, opened 2019-02-28
WorkspaceToDOM Includes Insertion Markers


amber-cd:
### Problem statement

If workspaceToDom is called while an insertion marker is visible, the insertion marker is included and can become a 'real' block.

### Expected Behavior

workspaceToDom should ignore insertion markers, as they are UI only.

### Actual Behavior

If workspaceToDom is called while an insertion marker is visible, the insertion marker is included and can become a 'real' block.

### Steps to Reproduce

This is probably a bit of a tough one to run into 'in the wild'... we ran into it due to a timing issue with Blockly as nested in a React component where we run `workspaceToDom` in order to save the XML to the store, then repopulate the workspace from that 'changed' XML before the insertion marker can be deleted. However, to verify the issue, I believe you can just do:

1. Add a listener to the workspace to run workspaceToDom when it changes.
2. Begin dragging the child out of a top-level block
3. While the insertion marker is still visible, check the output of workspaceToDom.

(This may require a bit of playing around... we specifically encountered it by dragging the block and holding it for a little while.)

### Operating System and Browser

* Desktop Chrome

I expect it applies to other browsers, but haven't tested it.

### Additional Information

I realize this is a pretty obscure scenario that only really occurs in scenarios like ours wherein we use React to try to make Blockly behave like a controlled component; however, I do believe this still constitutes a legitimate bug.

I nearly just made a PR to fix this, but I wanted advice on whether it would be accepted/whether this was an appropriate solution. For the record, my fix is just a simple fix to `workspaceToDom` to skip insertion markers:

```
Blockly.Xml.workspaceToDom = function(workspace, opt_noId) {
  var xml = Blockly.Xml.utils.createElement('xml');
  var variablesElement = Blockly.Xml.variablesToDom(
      Blockly.Variables.allUsedVarModels(workspace));
  if (variablesElement.hasChildNodes()) {
    xml.appendChild(variablesElement);
  }
  var comments = workspace.getTopComments(true);
  for (var i = 0, comment; comment = comments[i]; i++) {
    xml.appendChild(comment.toXmlWithXY(opt_noId));
  }
  var blocks = workspace.getTopBlocks(true);
  for (var i = 0, block; block = blocks[i]; i++) {
    if (!block.isInsertionMarker_) {
      xml.appendChild(Blockly.Xml.blockToDomWithXY(block, opt_noId));
    }
  }
  return xml;
};
```

[...]

PR #3874, opened 2020-04-30
Add ignoring insertion markers in code/xml gen

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#2322
#2646

### Proposed Changes

Makes it so that code generation and serialization skip over insertion markers.

### Reason for Changes

Supports real time code generation and real time serialization.

### Test Coverage

Tested generation & serialization for the following cases:
* Marker Surrounds
* Marker Enclosed
* Marker Enclosed and Surrounds
* Marker Prev
* Marker Next
* Marker Middle of Stack
* Marker On Output
* Marker On Input
* Marker Middle of Row
* Marker Detatched

I also check that this matches the pre-insertion marker generation.
Pre-Insertion Markers:
OldGen1

This Branch:
NewGen1

Tested on:
 * Desktop Chrome

### Documentation

Nope, I checked and the docs still show the "old" method, so we should be good!

### Additional Information

I went ahead and fixed both things but if you want me to remove the serialization fix that's cool!

### DIFF
--- a/tests/mocha/insertion_marker_test.js
+++ b/tests/mocha/insertion_marker_test.js
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2019 Google LLC
+ * Copyright 2020 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -46,17 +46,17 @@ suite('InsertionMarkers', function() {
   });
   suite('Code Generation', function() {
     setup(function() {
-      Blockly.JavaScript['stack_block'] = function(_) {
-        return 'stack;\n';
+      Blockly.JavaScript['stack_block'] = function(block) {
+        return 'stack[' + block.id + '];\n';
       };
       Blockly.JavaScript['row_block'] = function(block) {
         var value = Blockly.JavaScript
             .valueToCode(block, 'INPUT', Blockly.JavaScript.ORDER_NONE);
-        var code = 'row(' + value + ')';
+        var code = 'row[' + block.id + '](' + value + ')';
         return [code, Blockly.JavaScript.ORDER_NONE];
       };
       Blockly.JavaScript['statement_block'] = function(block) {
-        return 'statement{\n' + Blockly.JavaScript
+        return 'statement[' + block.id + ']{\n' + Blockly.JavaScript
             .statementToCode(block, 'STATEMENT') + '};\n';
       };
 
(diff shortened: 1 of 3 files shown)


BeksOmega:
I think I fixed part of the compile issues, but I'm not sure how to fix the other part.

Namely the bit causing:
```
gulp-google-closure-compiler: core-block_events.js:235: ERROR - [JSC_TYPE_MISMATCH]
    actual parameter 1 of Blockly.Xml.domToText does not match formal parameter
found   : (Element|null)
required: Node
  json['xml'] = Blockly.Xml.domToText(this.xml);
```
Here is what I think is happening:
Create and delete events call blockToDom() which could (now) return null if the block is an insertion marker. The compiler doesn't like that because the xml is later used as if it doesn't have a possibility of being null.

One important point is that (for create and delete events) the xml *will never actually* be null because insertion markers do not fire these events.

How do I tell the compiler that this code is fine? Do I need a cast, or is there a different method? Or maybe the code isn't fine - in that case, how would I fix it?

The relevant part of the delete event can be found [here](https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/core/block_events.js#L210).

tests/mocha/insertion_marker_test.js
```diff
@@ -0,0 +1,388 @@
...
+        "nextStatement": null
+      }]);
+  });
+  teardown(function() {
+    this.workspace.dispose();
+    delete Blockly.Blocks['stack_block'];
+    delete Blockly.Blocks['row_block'];
+    delete Blockly.Blocks['statement_block'];
+  });
+  suite('Code Generation', function() {
```
rachel-fenichel, line 47: To make sure that you're serializing the non-insertion marker block, add the ID to the code generation. Use a simple ID for the non-insertion marker blocks. Then you can also search for "insertion" in the generated code to make sure that block wasn't serialized.

tests/mocha/insertion_marker_test.js
```diff
@@ -0,0 +1,388 @@
...
+          '<xml xmlns="https://developers.google.com/blockly/xml">' +
+          '  <block type="stack_block" id="insertion"/>' +
+          '  <block type="stack_block"/>' +
+          '</xml>');
+      this.assertGen(xml, 'stack;\n');
+    });
+  });
+  suite('Serialization', function() {
+    setup(function() {
+      this.assertGen = function(xml, expectXml) {
```
rachel-fenichel, line 205: Rename to `assertXmlGen`--I got very confused while reviewing.

core/xml.js
```diff
@@ -128,9 +134,20 @@ Blockly.Xml.allFieldsToDom_ = function(block, element) {
...
+ * @return {Element} Tree of XML elements.
  */
 Blockly.Xml.blockToDom = function(block, opt_noId) {
+  // Skip over insertion markers.
+  if (block.isInsertionMarker()) {
+    var child = block.getChildren(false)[0];
+    if (child) {
+      return Blockly.Xml.blockToDom(child);
+    } else {
+      return null;
```
samelhusseini, line 146: If i'm understanding what you're saying correctly, this ``null`` would never actually happen?
If so, can we just cast child to non-nullable and maintain that blockToDom returns ``!Element``

BeksOmega: So this function /will/ return null if it is passed an insertion marker.

But it will never return null from the "create" event, because the create event always passes it a real block.

-----

The problem this is solving is making sure insertion markers don't get added to xml if the developer happens to serialize the workspace while the user is dragging. But maybe there's a better way to handle this than returning null (since it seems to really mess up typing).

I'm going to split this off into a second PR b/c I don't want it to hold up fixing the code gen. Maybe we can talk about it more over on #2322

tests/mocha/insertion_marker_test.js
BeksOmega: Ok it should now be asserting that correct blocks were serialized! (not just that the correct /number/ of blocks were serialized).

PR #3878, opened 2020-05-02
Include async in JS reserved words

seldomU:
## The basics

- [ x] My pull request is against develop

### Resolves

There is no issue for this.

### Proposed Changes

Javascript's `async` keyword was missing from the generator's reserved word list.

### DIFF
--- a/generators/javascript.js
+++ b/generators/javascript.js
@@ -35,7 +35,7 @@ Blockly.JavaScript.addReservedWords(
     'break,case,catch,class,const,continue,debugger,default,delete,do,else,export,extends,finally,for,function,if,import,in,instanceof,new,return,super,switch,this,throw,try,typeof,var,void,while,with,yield,' +
     'enum,' +
     'implements,interface,let,package,private,protected,public,static,' +
-    'await,' +
+    'async,await,' +
     'null,true,false,' +
     // Magic variable.
     'arguments,' +


NeilFraser:
Can you elaborate on why this needs to be a reserved word?  It's not in [Mozilla's list of reserved words](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords).  Indeed the [TC39 spec](https://tc39.es/ecma262/#sec-keywords-and-reserved-words) explicitly states:

> **async** is not reserved; it can be used as a variable name or statement label without restriction.

Testing the following code works just fine:
```
var async = 42;
alert(async);
```

seldomU:
I didn't consider that it can be both a language keyword and a valid identifier name. You're right, there is not need to reserve it.

The readability of the generated code might be an argument to reserve it anyway.

NeilFraser:
I'm not sure we want to get into the business of interfering with generated code when not needed.  Otherwise we'll start getting requests to ban 'l', 'I', 'True', swear words...

BeksOmega:
I wanted to add: if you want to implement this for your own project (or if anyone in the future does) I believe you can pass 'async' to the [addReservedWords](https://developers.google.com/blockly/reference/js/Blockly.Generator#addReservedWords) function without modifying the source. See the docs on [Generating and Running Javascript](https://developers.google.com/blockly/guides/app-integration/running-javascript). I hope this works for your project!

PR #3880, opened 2020-05-03
Simplify Closure-sourced code for menus

NeilFraser:
Bunch of changes relating to menus.  Highlights:
* Deleted a bunch of unused CSS rules and classes.
* Put the menu code on a diet, so it only supports features that we actually use.
* Improved menu UX with support for more keyboard navigation.
* Remove Component as a dependency of Menu/MenuItem.

### DIFF
--- a/core/components/menu/menu.js
+++ b/core/components/menu/menu.js
@@ -342,7 +342,7 @@ Blockly.Menu.prototype.highlightLast = function() {
  * Helper function that manages the details of moving the highlight among
  * child menuitems in response to keyboard events.
  * @param {number} startIndex Start index.
- * @param {boolean} delta Step direction: 1 to go down, -1 to go up.
+ * @param {number} delta Step direction: 1 to go down, -1 to go up.
  * @protected
  */
 Blockly.Menu.prototype.highlightHelper = function(startIndex, delta) {
--- a/core/components/menu/menu.js
+++ b/core/components/menu/menu.js
@@ -78,6 +78,15 @@ Blockly.Menu = function() {
    * @private
    */
   this.onKeyDownWrapper_ = null;
+
+  /**
+   * Map of DOM IDs to child menuitems. Each key is the DOM ID of a child
+   * menuitems's root element; each value is a reference to the child menu
+   * item itself.
+   * @type {!Object}
+   * @private
+   */
+  this.childElementIdMap_ = {};
 };
 Blockly.utils.object.inherits(Blockly.Menu, Blockly.Component);
 
@@ -215,17 +224,8 @@ Blockly.Menu.prototype.detachEvents_ = function() {
 // Child component management.
 
 /**
- * Map of DOM IDs to child menuitems. Each key is the DOM ID of a child
- * menuitems's root element; each value is a reference to the child menu
- * item itself.
- * @type {?Object}
- * @private
- */
-Blockly.Menu.prototype.childElementIdMap_ = null;
-
-/**
  * Creates a DOM ID for the child menuitem and registers it to an internal
- * hash table to be able to find it fast by id.
+ * hash table to be able to find it fast by ID.
  * @param {Blockly.Component} child The child menuitem. Its root element has
  *     to be created yet.
  * @private
@@ -233,14 +233,8 @@ Blockly.Menu.prototype.childElementIdMap_ = null;
 Blockly.Menu.prototype.registerChildId_ = function(child) {
   // Map the DOM ID of the menuitem's root element to the menuitem itself.
   var childElem = child.getElement();
-
   // If the menuitem's root element doesn't have a DOM ID assign one.
   var id = childElem.id || (childElem.id = child.getId());
-
-  // Lazily create the child element ID map on first use.
-  if (!this.childElementIdMap_) {
-    this.childElementIdMap_ = {};
-  }
   this.childElementIdMap_[id] = child;
 };
 
@@ -252,17 +246,13 @@ Blockly.Menu.prototype.registerChildId_ = function(child) {
  * @protected
  */
 Blockly.Menu.prototype.getMenuItem = function(node) {
-  // Ensure that this menu actually has child menuitems before
-  // looking up the menuitem.
-  if (this.childElementIdMap_) {
-    var elem = this.getElement();
-    while (node && node !== elem) {
-      var id = node.id;
-      if (id in this.childElementIdMap_) {
-        return this.childElementIdMap_[id];
-      }
-      node = node.parentNode;
+  var elem = this.getElement();
+  while (node && node !== elem) {
+    var id = node.id;
+    if (id in this.childElementIdMap_) {
+      return this.childElementIdMap_[id];
     }
+    node = node.parentNode;
   }
   return null;
 };
@@ -372,8 +362,7 @@ Blockly.Menu.prototype.highlightPrevious = function() {
 Blockly.Menu.prototype.highlightHelper = function(fn, startIndex) {
   // If the start index is -1 (meaning there's nothing currently highlighted),
   // try starting from the currently open item, if any.
-  var curIndex =
-      startIndex < 0 ? -1 : startIndex;
+  var curIndex = startIndex < 0 ? -1 : startIndex;
   var numItems = this.getChildCount();
 
   curIndex = fn.call(this, curIndex, numItems);
@@ -486,8 +475,7 @@ Blockly.Menu.prototype.handleMouseLeave_ = function(_e) {
  * @protected
  */
 Blockly.Menu.prototype.handleKeyEvent = function(e) {
-  if (this.getChildCount() != 0 &&
-      this.handleKeyEventInternal(e)) {
+  if (this.getChildCount() != 0 && this.handleKeyEventInternal(e)) {
     e.preventDefault();
     e.stopPropagation();
     return true;
--- a/core/components/menu/menu.js
+++ b/core/components/menu/menu.js
@@ -15,7 +15,8 @@ goog.provide('Blockly.Menu');
 goog.require('Blockly.utils.aria');
 goog.require('Blockly.utils.Coordinate');
 goog.require('Blockly.utils.dom');
-goog.require('Blockly.utils.object');
+goog.require('Blockly.utils.KeyCodes');
+goog.require('Blockly.utils.style');
 
 
 /**
--- a/core/components/menu/menu.js
+++ b/core/components/menu/menu.js
@@ -243,7 +243,7 @@ Blockly.Menu.prototype.getMenuItem_ = function(elem) {
         }
       }
     }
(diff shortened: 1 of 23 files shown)


core/components/menu/menu.js
```diff
@@ -210,92 +191,43 @@ Blockly.Menu.prototype.detachEvents_ = function() {
...
+  this.element_ = null;
 };
 
+// Child component management.
+
 /**
- * Returns the child menuitem that owns the given DOM node, or null if no such
- * menuitem is found.
+ * Returns the child menu item that owns the given DOM node, or null if no such
+ * menui tem is found.
```
samelhusseini, line 206: nit: fix spacing.

core/css.js
```diff
@@ -150,8 +150,7 @@ Blockly.Css.CONTENT = [
     'box-shadow: 4px 4px 20px 1px rgba(0,0,0,.15);',
```
samelhusseini, line 150: This is independent of any other css, in the spirit to keep module and css together, maybe this could be moved to ``tooltip.js``.

NeilFraser: True.  Though it is only one CSS rule, and tooltips aren't an optional module.

samelhusseini:
@NeilFraser, looks like you're still working on this, let me know when it's ready.

blockly:
It's ready.

Sent (awkwardly) from an Android device.

core/components/menu/menu.js
```diff
@@ -352,52 +262,43 @@ Blockly.Menu.prototype.highlightNext = function() {
...
-  var numItems = this.getChildCount();
-
-  curIndex = fn.call(this, curIndex, numItems);
-  var visited = 0;
-  while (visited <= numItems) {
-    var menuItem = /** @type {Blockly.MenuItem} */ (this.getChildAt(curIndex));
-    if (menuItem && this.canHighlightItem(menuItem)) {
-      this.setHighlightedIndex(curIndex);
-      return true;
+Blockly.Menu.prototype.highlightHelper = function(startIndex, delta) {
```
samelhusseini, line 292: Since this method is @protected, it should have an underscore at the end.

core/components/menu/menu.js
```diff
@@ -471,59 +367,46 @@ Blockly.Menu.prototype.handleMouseEnter_ = function(_e) {
...
- * calls the child menuitem's `handleKeyEvent` method to give the menuitem
- * a chance to handle the event first.
- * @param {Event} e Key event to handle.
+ * false otherwise.
+ * @param {!Event} e Key event to handle.
  * @return {boolean} Whether the event was handled by the container (or one of
  *     its children).
  * @protected
  */
 Blockly.Menu.prototype.handleKeyEventInternal = function(e) {
```
samelhusseini, line 398: Can this method be collapsed into the above ``handleKeyEvent``, with the preventDefault/stopPropagation at the end if it's been handled.

core/components/menu/menu.js
```diff
@@ -77,27 +83,57 @@ Blockly.Menu = function() {
    * @type {?Blockly.EventData}
    * @private
    */
-  this.onKeyDownWrapper_ = null;
+  this.onKeyDownHandler_ = null;
 };
```
samelhusseini, line 87: Add definition for ``roleName_`` and ``element``

core/components/menu/menu.js
```diff
@@ -77,27 +83,57 @@ Blockly.Menu = function() {
...
   var element = document.createElement('div');
-  element.id = this.getId();
-  this.setElementInternal(element);
-
-  // Set class
-  element.className = 'goog-menu goog-menu-vertical blocklyNonSelectable';
+  // goog-menu is deprecated, use blocklyMenu.  May 2020.
+  element.className = 'blocklyMenu goog-menu blocklyNonSelectable';
   element.tabIndex = 0;
+  Blockly.utils.aria.setRole(element, this.roleName_);
```
samelhusseini, line 107: Any reason to not default to a role name of menu here?

core/components/menu/menuitem.js
```diff
@@ -25,121 +24,108 @@ goog.require('Blockly.utils.object');
...
 
   /**
-   * @type {Blockly.MenuItem}
-   * @private
+   * Whether the menu item is rendered right-to-left.
+   * @type {boolean}
+   * @protected
    */
-  this.nextSibling_;
+  this.rightToLeft_ = false;
```
samelhusseini, line 62: Same here (and this doesn't need to be done in this PR, since its not a regression), but we should add definitions for: 
``this.checked_``, ``this.checkable_``, ``this.roleName``, and ``this.actionHandler_``.

core/components/menu/menuitem.js
```diff
@@ -25,121 +24,108 @@ goog.require('Blockly.utils.object');
...
  * Creates the menuitem's DOM.
- * @override
+ * @return {!Element} Completed DOM.
  */
 Blockly.MenuItem.prototype.createDom = function() {
   var element = document.createElement('div');
-  element.id = this.getId();
-  this.setElementInternal(element);
+  element.id = Blockly.utils.IdGenerator.getNextUniqueId();
+  element.blocklyMenuItem = this;  // Link DOM back to this data structure.
```
samelhusseini, line 73: Frankly, I'm not a huge fan of this pattern, as it shortcuts typing. There is precedence for this pattern in Blockly (eg: tooltips), so not saying it has to be changed. But here are my thoughts on the pattern:
- It makes it harder to debug for a someone new to the team / area, as there's hidden context (traditionally a DOM element doesn't have additional properties that are not DOM properties)
- The closure typechecker is not complaining (currently) because strictTypeChecks are not on, but when we do eventually get there, this will introduce issues. There are ways around that, ie: Create a new type that subclasses HTMLElement, and adds a blocklyMenuItem of type MenuItem as a property, but its clunky. 

That being said, if the perf benefit is worth it for you, then by all means keep as is.

core/components/menu/menuitem.js
```diff
@@ -176,23 +171,7 @@ Blockly.MenuItem.prototype.setCheckable = function(checkable) {
  * @package
  */
 Blockly.MenuItem.prototype.setChecked = function(checked) {
```
samelhusseini, line 173: Why not do the same thing with ``setEnabled``? If I'm reading between the lines, the suggestions here is that you shouldn't be able to call setChecked on a menu item after it's been rendered (unless you call render) yourself.

I'm okay with that as we don't have a use for that in Blockly, but I don't think we have a use for that for setEnabled either.

core/components/menu/menuitem.js
```diff
@@ -232,51 +217,39 @@ Blockly.MenuItem.prototype.setEnabled = function(enabled) {
 
   var el = this.getElement();
   if (el) {
-    if (!this.enabled_) {
-      Blockly.utils.dom.addClass(el, 'goog-menuitem-disabled');
+    // goog-menuitem-disabled is deprecated, use blocklyMenuItemDisabled.
```
samelhusseini, line 220: A link back to the comment above, this is what I'm referring to.https://github.com/google/blockly/pull/3880/files#r421074552

core/components/tree/basenode.js
```diff
@@ -731,14 +722,14 @@ Blockly.tree.BaseNode.prototype.onKeyDown = function(e) {
   var handled = true;
   switch (e.keyCode) {
     case Blockly.utils.KeyCodes.RIGHT:
-      if (e.altKey) {
+      if (e.altKey) {  // Why?
```
samelhusseini, line 725: No clue, doing a blame on the closure code didn't yield any insight, that bit of code was there from 11 years ago for when it was moved over to git.

I would say remove.

samelhusseini:
Adding ``api_change`` label to mention the deprecation of ``goog-menu`` CSS in the upcoming release.

core/components/menu/menu.js
NeilFraser: Ok, this is new.  None of the current @protected methods in core/components have an underscore.  I'll change them all.

samelhusseini: Yes, new as in it wasn't done because core/components mostly was inherited from closure, but course not new to Blockly (as the standard).

core/components/menu/menu.js
NeilFraser: Agreed.  I believe that Closure split these up so that they could write cleaner tests against the `handleKeyEventInternal` part.  But we don't have a requirement for that level of testing.

core/components/menu/menu.js
NeilFraser: Done.

core/components/menu/menu.js
NeilFraser: We currently use this code for two things, menus and dropdowns, each of which have completely different roles.  I'd rather the role be set explicitly since otherwise developers copying the menu code won't see the setRole method and may end up with UI that quietly has the wrong role.

samelhusseini: Fair enough.

core/components/menu/menuitem.js
NeilFraser: Done.

core/components/menu/menuitem.js
NeilFraser: Oh, good catch.  Deleting all that junk in setEnabled.

core/components/tree/basenode.js
NeilFraser: Removed.

core/components/menu/menuitem.js
```diff
@@ -25,121 +24,138 @@ goog.require('Blockly.utils.object');
...
+  /**
+   * Whether the menu item is rendered right-to-left.
+   * @type {boolean}
+   * @private
+   */
+  this.rightToLeft_ = false;
+
+  /**
+   * ARIA name for this menu.
+   * @type {Blockly.utils.aria.Role}
```
samelhusseini, line 66: Role is not an object type but an enum, so this is actually ``{!Blockly.utils.aria.Role}`` similar to ``{string}``. 

Since it's nullable, you'll have to explicitly state that, ie: ``{?Blockly.utils.aria.Role}``

core/components/menu/menuitem.js
NeilFraser: The original Closure code maintained a database for efficient look-ups.  But it was a lot of code and complexity undeserving for this task.  The DOM-to-JS links were simple and efficient, but have the problems you cite.  I've replaced this with an O(n) search of the menu's options.  Shouldn't be a problem given that our menus/dropdowns aren't intended to have hundreds or thousands of items.

NeilFraser:
Finally got the compiler happy.  Nodes vs Elements are annoying.
Assuming you are ok with this PR as it is, do you prefer the 30 commits to be rebased or squashed?

samelhusseini:
Awesome, thanks!
Squashed please.

PR #3883, opened 2020-05-05
Add ignoring insertion markers in xml gen

### LINKED ISSUE #2322: WorkspaceToDOM Includes Insertion Markers

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #2322 

### Proposed Changes

Skips over insertion markers, returning the xml of the insertion marker's child instead. If the insertion marker has no children we return a DocumentFragment instead (as suggested by @NeilFraser ).

### Reason for Changes

If the developer happens to serialize the workspace while the user is dragging we don't want insertion markers to be included in the xml.

### Test Coverage

Unit tests for all of the following:
* Marker Surrounds
* Marker Enclosed
* Marker Enclosed and Surrounds
* Marker Prev
* Marker Next
* Marker Middle of Stack
* Marker On Output
* Marker On Input
* Marker Middle of Row
* Marker Detatched

### Documentation

Noup

### Additional Information

Hopefully the compiler likes this! :crossed_fingers:

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -80,8 +80,9 @@ Blockly.Xml.blockToDomWithXY = function(block, opt_noId) {
   if (block.isInsertionMarker()) {  // Skip over insertion markers.
     block = block.getChildren(false)[0];
     if (!block) {
-      // Disappears when appended.
-      return /** @type{!Element} */ (new DocumentFragment());
+      // Disappears when appended. Cast to ANY b/c DocumentFragment -> Element
+      // is invalid. We have to cast to ANY in between.
+      return /** @type{?} */ (new DocumentFragment());
     }
   }
 
@@ -146,8 +147,9 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
     if (child) {
       return Blockly.Xml.blockToDom(child);
     } else {
-      // Disappears when appended.
-      return /** @type{!Element} */ (new DocumentFragment());
+      // Disappears when appended. Cast to ANY b/c DocumentFragment -> Element
+      // is invalid. We have to cast to ANY in between.
+      return /** @type{?} */ (new DocumentFragment());
     }
   }
 
--- a/core/xml.js
+++ b/core/xml.js
@@ -45,7 +45,10 @@ Blockly.Xml.workspaceToDom = function(workspace, opt_noId) {
   }
   var blocks = workspace.getTopBlocks(true);
   for (var i = 0, block; (block = blocks[i]); i++) {
-    xml.appendChild(Blockly.Xml.blockToDomWithXY(block, opt_noId));
+    var dom = Blockly.Xml.blockToDomWithXY(block, opt_noId);
+    if (dom) {
+      xml.appendChild(dom);
+    }
   }
   return xml;
 };
@@ -82,6 +85,9 @@ Blockly.Xml.blockToDomWithXY = function(block, opt_noId) {
     width = block.workspace.getWidth();
   }
   var element = Blockly.Xml.blockToDom(block, opt_noId);
+  if (!element) {
+    return null;
+  }
   var xy = block.getRelativeToSurfaceXY();
   element.setAttribute('x',
       Math.round(block.workspace.RTL ? width - xy.x : xy.x));
@@ -128,9 +134,20 @@ Blockly.Xml.allFieldsToDom_ = function(block, element) {
  * Encode a block subtree as XML.
  * @param {!Blockly.Block} block The root block to encode.
  * @param {boolean=} opt_noId True if the encoder should skip the block ID.
- * @return {!Element} Tree of XML elements.
+ * @return {Element} Tree of XML elements.
  */
 Blockly.Xml.blockToDom = function(block, opt_noId) {
+  // Skip over insertion markers.
+  if (block.isInsertionMarker()) {
+    var child = block.getChildren(false)[0];
+    if (child) {
+      return Blockly.Xml.blockToDom(child);
+    } else {
+      return null;
+    }
+  }
+
+
   var element =
       Blockly.utils.xml.createElement(block.isShadow() ? 'shadow' : 'block');
   element.setAttribute('type', block.type);
@@ -186,8 +203,11 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
         container.appendChild(Blockly.Xml.cloneShadow_(shadow, opt_noId));
       }
       if (childBlock) {
-        container.appendChild(Blockly.Xml.blockToDom(childBlock, opt_noId));
-        empty = false;
+        var dom = Blockly.Xml.blockToDom(childBlock, opt_noId);
+        if (dom) {
+          container.appendChild(dom);
+          empty = false;
+        }
       }
     }
     container.setAttribute('name', input.name);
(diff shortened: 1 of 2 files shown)


core/xml.js
```diff
@@ -186,8 +200,11 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
         container.appendChild(Blockly.Xml.cloneShadow_(shadow, opt_noId));
       }
       if (childBlock) {
-        container.appendChild(Blockly.Xml.blockToDom(childBlock, opt_noId));
-        empty = false;
+        var elem = Blockly.Xml.blockToDom(childBlock, opt_noId);
+        if (elem instanceof Element) {
```
samelhusseini, line 204: Instead compare the nodeType: 
`elem.nodeType == Blockly.utils.dom.NodeType.ELEMENT_NODE`

core/xml.js
```diff
@@ -217,9 +234,12 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
 
   var nextBlock = block.getNextBlock();
   if (nextBlock) {
-    var container = Blockly.utils.xml.createElement('next');
-    container.appendChild(Blockly.Xml.blockToDom(nextBlock, opt_noId));
-    element.appendChild(container);
+    var elem = Blockly.Xml.blockToDom(nextBlock, opt_noId);
+    if (elem instanceof Element) {
```
samelhusseini, line 238: Same as above, use ``nodeType``.

tests/mocha/insertion_marker_test.js
```diff
@@ -200,4 +200,189 @@ suite('InsertionMarkers', function() {
       this.assertGen(xml, 'stack[a];\n');
     });
   });
+  suite('Serialization', function() {
+    setup(function() {
+      this.assertXml = function(xml, expectXml) {
+        Blockly.Xml.domToWorkspace(xml, this.workspace);
+        var block = this.workspace.getBlockById('insertion');
+        block.isInsertionMarker_ = true;
```
samelhusseini, line 208: Any reason not to use ``setInsertionMarker(true)`` here?

core/xml.js
```diff
@@ -131,6 +133,18 @@ Blockly.Xml.allFieldsToDom_ = function(block, element) {
...
  */
 Blockly.Xml.blockToDom = function(block, opt_noId) {
+  // Skip over insertion markers.
+  if (block.isInsertionMarker()) {
+    var child = block.getChildren(false)[0];
+    if (child) {
+      return Blockly.Xml.blockToDom(child);
+    } else {
+      // Disappears when appended.
+      return /** @type{!Element} */ (new DocumentFragment());
```
samelhusseini, line 143: ``domToBlockHeadless_`` is only excluding text nodes, what happens when this round-trips?

samelhusseini:
Something funny is happening with the XY positions of the block when an insertion marker is wrapped around it, here's a screengrab of a mirror workspace that takes the output of `workspaceToDom` and inputs it into a second read only workspace: 

May-05-2020 17-17-09

BeksOmega:
> Something funny is happening with the XY positions of the block when an insertion marker is wrapped around it

Thanks for the gif and the testing! The problem is that the blockToDomWithXY_ is attaching the coords of the insertion marker to the xml for the "if" block. This happens whenever the insertion marker is the top level parent.

At the time I was coding this I couldn't think of a way around it, but thinking about it again I should be able to use the same "skip over the insertion marker" trick. I'll give it a shot and report back!

core/xml.js
BeksOmega: Thank you! I really didn't want to use `instanceof` but I couldn't come up with a different check.

core/xml.js
BeksOmega: The document fragment kinda seems to disappear when it's appended? I'm not exactly sure what's going on, you'll have to ask @NeilFraser The document fragment doesn't exist if you log the xml, and it doesn't get added to the string when you run domToText.

I added a call to domToWorkspace in the assertXml function. It won't test the structure of the xml (toString should do that) but it will make sure that round-tripping the xml doesn't throw errors.

BeksOmega:
Hmm... I was hoping it would let me cast to an Element because DocumentFragment inherits from Node. I was looking at [this](https://groups.google.com/forum/#!topic/closure-compiler-discuss/A1hIQJwNxtA) discussion.

Any advice @samelhusseini? It looks like we might need to go back to the drawing board.

samelhusseini:
Yeah the type cast should be fine, you just need to add parentheses around the thing you're casting. eg: 

```
return /** @type{!Element} */ (new DocumentFragment());
```

BeksOmega:
@samelhusseini would you be able to give me some advice? I still can't get it to cast :/

samelhusseini:
Yep, I'll take a look later today :)

samelhusseini:
Took a look and DocumentFragment -> Element is an invalid type cast. In order to make the compiler happy, just cast it to the any type. `?`: 
```js
    if (!block) {
      // Disappears when appended.
      return /** @type {?} */ (new DocumentFragment());
    }
```
Leave a comment for why we're casting to ANY, something along the lines of it's an invalid typecast from DocumentFragment to Element so we have to cast to ANY in between.

BeksOmega:
Thanks @samelhusseini ! That worked, mocha just timed out.

PR #3884, opened 2020-05-05
Hide DropdownDiv on flyout scroll

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Also hide the Dropdown div when the flyout is scrolled.

### Reason for Changes

We do it for the widget div, looks weird to have a hanging dropdown div when the flyout is scrolled.

### Test Coverage

Tested on simple flyout, variable dropdown.

Tested on:

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -239,8 +239,9 @@ Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
     pos = Math.min(pos, limit);
     pos = Math.max(pos, 0);
     this.scrollbar_.set(pos);
-    // When the flyout moves from a wheel event, hide WidgetDiv.
+    // When the flyout moves from a wheel event, hide WidgetDiv and DropDownDiv.
     Blockly.WidgetDiv.hide();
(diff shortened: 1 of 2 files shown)


BeksOmega:
Would calling `hideChaff(true)` work? ([link](https://github.com/google/blockly/blob/master/core/blockly.js#L325))

samelhusseini:
No because it would also hide the flyout on scroll (which we dont want)

samelhusseini:
It could, only thing is it would also clear the tooltip. Which I'm not sure if we also want to lump that in.

BeksOmega:
Testing it, it looks like the tooltip gets hidden anyway (no clue where that's happening haha). But it's up to you whether you want to switch over or not.
TooltipHide

samelhusseini:
In the above screengrab (although not very clear), the tooltip only disappeared after you scrolled past a single block, if you try to go slower you'll notice that the tooltip doesn't disappear on a single block until you move past that block as you scroll.

Given that browser tooltips don't disappear immediately when you scroll, let's stick with this change.

PR #3891, opened 2020-05-07
Toolbox definition

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Add support for defining a toolbox in json.

### Proposed Changes
* Refactors toolbox syncTrees_ => createTrees_ and has it work with json instead of xml
* Converts the xml, NodeList, string, or Array that we get in into JSON in Blockly.Options.parseToolbox.

Not in this PR: 
* Support for defining a type of block and filling in the shadow blocks. 
* Combining simple toolbox with toolbox definition.
### Reason for Changes
xml can be difficult

### Test Coverage
To be Tested: 
* Mutators
* updateToolbox
* simple toolbox
* category toolbox
* custom category

Tested on:

### Additional Information
We can merge this into a toolbox branch if people are uncomfortable with it being in develop just yet.


core/flyout_base.js
```diff
@@ -426,77 +426,60 @@ Blockly.Flyout.prototype.hide = function() {
 
 /**
  * Show and populate the flyout.
- * @param {!Array|!NodeList|string} xmlList List of blocks to show.
+ * @param {!Array|!NodeList|string} contentInfo List of contents to display in
```
samelhusseini, line 429: Can this cover the different possible options for contentInfo. I know it's mentioned further below in the comments, but it's good to have that stated in the parameter.
Also, an Array of what? Array of NodeList only, Array of JSON, both?

core/flyout_base.js
```diff
@@ -426,77 +426,60 @@ Blockly.Flyout.prototype.hide = function() {
...
-          // Record blocks that were initially disabled.
-          // Do not enable these blocks as a result of capacity filtering.
-          this.permanentlyDisabled_.push(curBlock);
-        }
-        contents.push({type: 'block', block: curBlock});
-        // This is a deprecated method for adding gap to a block.
-        // <block type="math_arithmetic" gap="8"></block>
-        var gap = parseInt(xml.getAttribute('gap'), 10);
-        gaps.push(isNaN(gap) ? default_gap : gap);
+        var blockInfo = /** @type {Blockly.utils.toolbox.BlockInfo} */ (contentItem);
```
samelhusseini, line 468: Let's keep the naming consistent, blockItem from contentItem, or a blockInfo from a contentInfo

core/flyout_base.js
```diff
@@ -426,77 +426,60 @@ Blockly.Flyout.prototype.hide = function() {
...
-        } else {
-          gaps.push(default_gap);
-        }
+        var sepInfo = /** @type {Blockly.utils.toolbox.SeparatorInfo} */ (contentItem);
+        this.addSeparator_(sepInfo, gaps, defaultGap);
         break;
       case 'LABEL':
+        var labelInfo = /** @type {Blockly.utils.toolbox.LabelInfo} */ (contentItem);
+        // A label is a button with different styling.
+        this.addButton_(labelInfo, contents, gaps, defaultGap, true);
```
samelhusseini, line 478: Any reason we couldn't pass an object to that consists of: {labelInfo, contents, gaps, etc.. } rather than passing all these parameters.

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
+/**
```
samelhusseini, line 1: All utilities under /utils are currently generic non-blockly specific utilities. This would change that, just a thought but we should discuss whether this is the right place for this.

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+'use strict';
+
+goog.provide('Blockly.utils.toolbox');
+
+
+/**
+ * A block definition.
+ * @typedef {{
+ *            contenttype:string,
```
samelhusseini, line 19: camelcase?

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+Blockly.utils.toolbox.ToolboxItem;
+
+/**
+ * A category definition.
+ * @typedef {{
+ *            contenttype:string,
+ *            name:string,
+ *            categorystyle:?string,
+ *            colour:?string,
+ *            contents:Array<Blockly.utils.toolbox.ToolboxItem>
```
samelhusseini, line 69: ``Array.``.

core/workspace_svg.js
```diff
@@ -1815,11 +1816,15 @@ Blockly.WorkspaceSvg.prototype.showContextMenu = function(e) {
 
 /**
  * Modify the block tree on the existing toolbox.
- * @param {Node|string} tree DOM tree of blocks, or text representation of same.
+ * @param {Node|Array<Object>|string} toolboxDef DOM tree of toolbox contents,
```
samelhusseini, line 1819: Why is the type of toolbox contents Object here?

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+ * @param {!NodeList|!Node|!Array<Blockly.utils.toolbox.ToolboxItem>} toolboxDef The
+ *     definition of the toolbox in one of its many forms.
+ * @return {!Array<Blockly.utils.toolbox.ToolboxItem>} A list of objects in the
+ *    toolbox.
+ * @private
+ */
+Blockly.utils.toolbox.toolboxXmlToJson_ = function(toolboxDef) {
+  var arr = [];
+  // If it is a node it will have children.
+  var childNodes = toolboxDef.children;
```
samelhusseini, line 107: Use of ``children`` is fine in IE as this is an XML Node, but if it was SVG it wouldn't be supported (just FYI). Any reason not to use childNodes, and just avoid having the reader guess whether or not this would break IE?

core/options.js
```diff
@@ -37,10 +38,9 @@ Blockly.Options = function(options) {
     var hasDisable = false;
     var hasSounds = false;
   } else {
-    var languageTree =
-        Blockly.Options.parseToolboxTree(options['toolbox'] || null);
-    var hasCategories = Boolean(languageTree &&
-        languageTree.getElementsByTagName('category').length);
+    var languageTree = Blockly.Options.parseToolboxTree(options['toolbox'] || null);
+    languageTree = Blockly.utils.toolbox.parseToolbox(languageTree);
```
samelhusseini, line 42: Hard to tell from this naming what the difference is between parseToolboxTree and toolbox.parseToolbox

core/options.js
```diff
@@ -312,7 +312,7 @@ Blockly.Options.parseThemeOptions_ = function(options) {
 
 /**
  * Parse the provided toolbox tree into a consistent DOM format.
- * @param {Node|string} tree DOM tree of blocks, or text representation of same.
+ * @param {Node|NodeList|string|null} tree DOM tree of blocks, or text representation of same.
```
samelhusseini, line 315: Use ``Node|NodeList|?string`` instead of null.

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+      obj['xmldef'] = Blockly.utils.xml.domToText(child);
+    }
+    // Get the contents for a category.
+    if (child.tagName.toUpperCase() === 'CATEGORY') {
+      for (var k = 0; k < child.children.length; k++) {
+        obj['contents'] = Blockly.utils.toolbox.toolboxXmlToJson_(child);
+      }
+    }
+    // Add xml attributes to object
+    for (var j = 0; j < child.attributes.length; j++) {
```
samelhusseini, line 129: `.attributes` unsupported on IE: https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+ * @param {Node|Array<Blockly.utils.toolbox.ToolboxItem>} toolboxDef The definition
+ *    of the toolbox. Either in xml or JSON.
+ * @return {boolean} True if the toolbox input has categories.
+ * @package
+ */
+Blockly.utils.toolbox.hasCategories = function(toolboxDef) {
+  if (Array.isArray(toolboxDef)) {
+    // Search for categories
+    for (var i = 0; i < toolboxDef.length; i++) {
+      if (toolboxDef[i]['contenttype'].toUpperCase() === 'CATEGORY') {
```
samelhusseini, line 149: ``==`` unless otherwise needed (In Blockly standard, I know that's not what the Goog style guide says)

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+        return true;
+      }
+    }
+  } else {
+    return Boolean(toolboxDef &&
```
samelhusseini, line 154: You can also do ``!!(toolboxDef && toolboxDef.getElementsByTagName('category').length)``

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Handle the before tree item selected action.
+ * @param {Node|Array<Blockly.utils.toolbox.ToolboxItem>} toolboxDef The definition
+ *    of the toolbox. Either in xml or JSON.
+ * @return {boolean} True if the toolbox input has categories.
+ * @package
+ */
+Blockly.utils.toolbox.hasCategories = function(toolboxDef) {
+  if (Array.isArray(toolboxDef)) {
+    // Search for categories
```
samelhusseini, line 147: Do we really need to search for categories here? If it's a category toolbox, then every item in the array needs to be a category, simply ``!!(toolboxDef.length && toolboxDef[0]['contenttype'].toUpperCase() == 'CATEGORY')``.

core/utils/toolbox.js
```diff
@@ -0,0 +1,158 @@
...
+'use strict';
+
+goog.provide('Blockly.utils.toolbox');
+
+
+/**
+ * A block definition.
+ * @typedef {{
+ *            contenttype:string,
+ *            xmldef:string
```
samelhusseini, line 20: blockXml?

core/utils/toolbox.js
samelhusseini: I assume you don't want to go with type here as it could be confused with a block's type.
How about tag? or kind?

core/flyout_base.js
```diff
@@ -426,77 +426,60 @@ Blockly.Flyout.prototype.hide = function() {
...
   // Create the blocks to be shown in this flyout.
   var contents = [];
   var gaps = [];
   this.permanentlyDisabled_.length = 0;
-  var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
-  for (var i = 0, xml; (xml = xmlList[i]); i++) {
-    if (!xml.tagName) {
+  var defaultGap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
+  for (var i = 0, contentItem; (contentItem = parsedContent[i]); i++) {
+    if (!contentItem.contenttype) {
```
samelhusseini, line 463: Is this still possible? if we've parsed through the content

core/utils/toolbox.js
alschmiedt: done

core/flyout_base.js
alschmiedt: done

core/flyout_base.js
```diff
@@ -426,77 +426,58 @@ Blockly.Flyout.prototype.hide = function() {
...
           ' a toolbox category.');
     }
-    xmlList = fnToApply(this.workspace_.targetWorkspace);
-    if (!Array.isArray(xmlList)) {
+    flyoutDef = fnToApply(this.workspace_.targetWorkspace);
+    if (!Array.isArray(flyoutDef)) {
       throw TypeError('Result of toolbox category callback must be an array.');
     }
   }
+  // contentInfo is either an array with JSON, an Array with xml, or a NodeList.
```
rachel-fenichel, line 454: This comment is confusing not because contentInfo -> flyoutDef, but because I expected it to be explaining the variable being declared here (parsedContent) instead of the one being used (flyoutDef).

core/options.js
```diff
@@ -37,10 +38,9 @@ Blockly.Options = function(options) {
     var hasDisable = false;
     var hasSounds = false;
   } else {
-    var languageTree =
-        Blockly.Options.parseToolboxTree(options['toolbox'] || null);
-    var hasCategories = Boolean(languageTree &&
-        languageTree.getElementsByTagName('category').length);
+    var languageTree = Blockly.Options.parseToolboxTree(options['toolbox'] || null);
+    languageTree = Blockly.utils.toolbox.parseToolboxContents(languageTree);
```
rachel-fenichel, line 42: I have a similar concern here about reusing a variable: if the first and second lines are doing different things, and return different types, you should declare a new variable instead of reusing the old one.

core/toolbox.js
```diff
@@ -252,6 +252,163 @@ Blockly.Toolbox.prototype.renderTree = function(languageTree) {
   }
 };
 
+/**
+ * Create the toolbox tree.
+ * @param {Array.<Blockly.utils.toolbox.ToolboxInfo>} toolboxDef List of objects
+ *    holding information on toolbox contents.
+ * @param {!Blockly.tree.BaseNode} treeOut The TreeControl or TreeNode
```
rachel-fenichel, line 259: I don't understand the descriptions of treeOut and the return value. Both are nodes--are they the *root nodes* of the things being described?

core/toolbox.js
```diff
@@ -252,6 +252,163 @@ Blockly.Toolbox.prototype.renderTree = function(languageTree) {
...
+  } else {
+    openNode = this.createTree_(categoryInfo.contents, childOut) || openNode;
+  }
+  this.addColourOrStyle_(categoryInfo, childOut, categoryName);
+  openNode = this.setExpanded_(categoryInfo, childOut) || openNode;
+  return openNode;
+};
+
+/**
+ * Create the TreeNode for the category.
```
rachel-fenichel, line 323: What's the difference between `addCategory` and `addCategoryChild`?

core/toolbox.js
```diff
@@ -252,6 +252,163 @@ Blockly.Toolbox.prototype.renderTree = function(languageTree) {
...
+ */
+Blockly.Toolbox.prototype.addColourOrStyle_ = function(
+    categoryInfo, childOut, categoryName) {
+  var styleName = categoryInfo['categorystyle'];
+  var colour = categoryInfo['colour'];
+
+  if (colour && styleName) {
+    childOut.hexColour = '';
+    console.warn('Toolbox category "' + categoryName +
+        '" can not have both a style and a colour');
```
rachel-fenichel, line 354: can not -> must not

core/utils/toolbox.js
rachel-fenichel: Yes, things under /utils are supposed to not be blockly specific.

Also, half of this file is types. I know I said the types should be in the same file as the rest of the utils, but what if we moved all the typedefs into another file, and possibly created a directory for typedef-only files?

core/utils/toolbox.js
```diff
@@ -0,0 +1,152 @@
...
+      return /** @type {!Array.<Blockly.utils.toolbox.ToolboxInfo>} */ (toolboxDef);
+    }
+  }
+
+  return Blockly.utils.toolbox.toolboxXmlToJson_(toolboxDef);
+};
+
+/**
+ * Convert the xml for a toolbox to JSON.
+ * @param {!NodeList|!Node|!Array.<Blockly.utils.toolbox.ToolboxInfo>} toolboxDef The
```
rachel-fenichel, line 98: Based on the types in the previous function, this can't be an array of ToolboxInfo.

core/utils/toolbox.js
```diff
@@ -0,0 +1,152 @@
...
+ * @return {!Array.<Blockly.utils.toolbox.ToolboxInfo>} Array of JSON holding
+ *    information on toolbox contents.
+ * @package
+ */
+Blockly.utils.toolbox.parseToolboxContents = function(toolboxDef) {
+  if (!toolboxDef) {
+    return [];
+  }
+  // The array can be either an array of xml or an array of JSON.
+  if (Array.isArray(toolboxDef)) {
```
rachel-fenichel, line 87: Combine the ifs with `&&`

core/utils/toolbox.js
```diff
@@ -0,0 +1,152 @@
...
+      var attr = child.attributes[j];
+      obj[attr.nodeName] = attr.value;
+    }
+    arr.push(obj);
+  }
+  return arr;
+};
+
+/**
+ * Handle the before tree item selected action.
```
rachel-fenichel, line 139: Comment is out of date?

core/workspace_svg.js
```diff
@@ -1815,11 +1816,16 @@ Blockly.WorkspaceSvg.prototype.showContextMenu = function(e) {
 
 /**
  * Modify the block tree on the existing toolbox.
- * @param {Node|string} tree DOM tree of blocks, or text representation of same.
+ * @param {Node|Array.<Blockly.utils.toolbox.ToolboxInfo>|Array.<Node>|string} toolboxDef
```
rachel-fenichel, line 1819: Since this type (the combination of all of the others) is popping up all over the place, consider making a typedef for it.

core/toolbox.js
alschmiedt: done

core/utils/toolbox.js
alschmiedt: done

core/utils/toolbox.js
alschmiedt: done

core/utils/toolbox.js
alschmiedt: done

core/toolbox.js
alschmiedt: removed addCategoryChild. And just put the logic in addCategory.

core/flyout_base.js
```diff
@@ -426,77 +426,58 @@ Blockly.Flyout.prototype.hide = function() {
 
 /**
  * Show and populate the flyout.
- * @param {!Array|!NodeList|string} xmlList List of blocks to show.
- *     Variables and procedures have a custom set of blocks.
+ * @param {!Array.<Blockly.utils.toolbox.ToolboxInfo>|!Array.<Node>|!NodeList|string} flyoutDef
+ *    List of contents to display in the flyout as either an array of xml an
```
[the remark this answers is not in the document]
alschmiedt: And an array of Blockly.utils.toolbox.ToolboxInfo.

core/workspace_svg.js

core/utils/toolbox.js
samelhusseini: The other issue with that is what would the "namespace" be, one possibility is that we do a ``Blockly.Typedefs`` provide, and require it from `blockly.js`.
I agree with the concern of discoverability, it's more discoverable if in the same file.

core/options.js
samelhusseini: Are they both parsing, or is one parsing and the other converting? Also, why not add the parse util within the ``parseToolboxTree`` definition?

alschmiedt: Oh I like the word converting, will use that! An parseToolboxTree is a public method, so I was trying to not change the public API.

core/utils/toolbox.js
alschmiedt: done

core/flyout_button.js
```diff
@@ -23,11 +23,12 @@ goog.require('Blockly.utils.dom');
...
  *     button.
  * @param {!Blockly.WorkspaceSvg} targetWorkspace The flyout's target workspace.
- * @param {!Element} xml The XML specifying the label/button.
+ * @param {!Blockly.utils.toolbox.ButtonInfo|!Blockly.utils.toolbox.LabelInfo} json
+ *    The JSON specifying the label/button.
  * @param {boolean} isLabel Whether this button should be styled as a label.
  * @constructor
  */
-Blockly.FlyoutButton = function(workspace, targetWorkspace, xml, isLabel) {
+Blockly.FlyoutButton = function(workspace, targetWorkspace, json, isLabel) {
```
samelhusseini, line 31: Is this API breaking if we don't also accept ``Element``?

core/utils/toolbox.js
```diff
@@ -0,0 +1,161 @@
...
+
+
+/**
+ * The information needed to create a block in the toolbox.
+ * @typedef {{
+ *            kind:string,
+ *            blockxml:(string|Node)
+ *          }}
+ */
+Blockly.utils.toolbox.BlockInfo;
```
samelhusseini, line 23: Given the name space is already explicit enough. I think it's fair to cut out the ``Info`` part of each of these.

core/flyout_button.js
alschmiedt: Yes. However, I couldn't think of a reason why someone would be making a Blockly.FlyoutButton without going through flyout.show_.

samelhusseini: Perhaps it should be package then?

core/flyout_base.js
```diff
@@ -426,82 +426,39 @@ Blockly.Flyout.prototype.hide = function() {
...
           ' a toolbox category.');
     }
-    xmlList = fnToApply(this.workspace_.targetWorkspace);
-    if (!Array.isArray(xmlList)) {
+    flyoutDef = fnToApply(this.workspace_.targetWorkspace);
+    if (!Array.isArray(flyoutDef)) {
       throw TypeError('Result of toolbox category callback must be an array.');
     }
   }
+  // Parse the Array or NodeList passed in into an Array of Blockly.utils.toolbox.Toolbox.
```
samelhusseini, line 454: nit: 80 char length

core/flyout_base.js
```diff
@@ -532,6 +489,129 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   this.workspace_.addChangeListener(this.reflowWrapper_);
 };
 
+/**
+ * Create the contents array and gaps array necessary to create the layout for
+ * the flyout.
+ * @param {Array.<Blockly.utils.toolbox.Toolbox>} parsedContent The array
+ *   of objects to show in the flyout.
+ * @return {{contents:Array<Object>, gaps:Array<number>}} The list of contents
+ *   and gaps needed to lay out the flyout.
```
samelhusseini, line 498: Add visibility of this method

core/toolbox.js
```diff
@@ -252,6 +252,156 @@ Blockly.Toolbox.prototype.renderTree = function(languageTree) {
...
+
+/**
+ * Add either the colour or the style for a category.
+ * @param {!Blockly.utils.toolbox.Category} categoryInfo The object holding
+ *    information on the category.
+ * @param {!Blockly.tree.TreeNode} childOut The TreeNode for a category.
+ * @param {string} categoryName The name of the category.
+ * @private
+ */
+Blockly.Toolbox.prototype.addColourOrStyle_ = function(
```
samelhusseini, line 339: Are you adding or setting the colour here?
For the sake of consistency with the methods below, should this be setColourOrStyle_ ?

core/utils/toolbox.js
samelhusseini: Did you figure out what was going on here?

core/flyout_base.js
alschmiedt: ddone

core/flyout_base.js
alschmiedt: done

core/flyout_button.js
alschmiedt: done

core/toolbox.js
alschmiedt: done

core/utils/toolbox.js
alschmiedt: Not yet :/ having trouble testing with ie. Working on it next.

alschmiedt: Ok, so I ran all my tests on ie11 and it looks like they are all working. `.attributes` seems to be supported for Elements? You brought up the point about Node vs Element. Since we check that a tagName exists before looking for the attributes I think we are essentially ensuring that anything that gets to this point will be an Element.

samelhusseini: Okay, perhaps mdn is outdated on this particular. Thanks, I feel like we've done our due diligence here.

PR #3893, opened 2020-05-07
Move Menu + MenuItem out of components to utils

NeilFraser:
Blockly.Menu and Blockly.MenuItem no longer depend on Blockly.Component and thus should not be in /core/component.

Menu and MenuItem do not depend on anything other than utils, which would seem to also make them belong in utils.


samelhusseini:
Hmm, many other scripts only depend on utils, such as: `Blockly.blockAnimations`, `Blockly.BlockDrragSurfaceSvg`. 

Last we spoke about the purpose of utils, I thought it's supported (public) re-usable components.
I'm not sure I feel the menu is that, although you could argue that it could be re-used in another application, I'd rather have the flexibility to keep it as internal Blockly code that could change its APIs in the future (especially given the recent changes to the API).

NeilFraser:
Actually, what triggered the refactoring of Blockly's menu was a desire to use them in Code City.  However, CC's feature requirements are different (e.g. we require multiple-selection, we don't require checking), so I won't be using Blockly's menus after all.

I've amended the PR to move the menus into core Blockly.  What do you think of this option?

samelhusseini:
Yep, that makes sense to me. They're more a "part" of Blockly now by moving them out of components. That being said if we ever wanted to restructure core into folders, we could revisit the placement then (as it is today, some things are in folders, and others are not).

Issue #57, opened 2015-02-01
Keep the operator precedence when collapse a block


carloslfu:
![operator-precedence](https://cloud.githubusercontent.com/assets/5993168/5993040/4f6fde08-aa10-11e4-9e50-83d09041706d.jpg)

NeilFraser:
Do you have suggestions on how to achieve this?  In this example, parentheses would be nice, but in most cases that would be awful (e.g. "if ((x) = (true)) do".  I don't see a solution here that doesn't make things worse.

carloslfu:
I have tested this solution and should be good:
```
Blockly.Block.prototype.toString = function(opt_maxLength) {
  var text = [];
  for (var i = 0, input; input = this.inputList[i]; i++) {
    for (var j = 0, field; field = input.fieldRow[j]; j++) {
      text.push(field.getText());
    }
    if (input.connection) {
      var child = input.connection.targetBlock();
      if (child) {
        /////------------------------------
        // could be declared in block definition like (in 'math_arithmetic' case):
        /*
          this.setPrecedence(true); //default false
          this.setPrecedenceOrder({'ADD' : 0,
                             'MINUS': 0,
                             'MULTIPLY': 1,
                             'DIVIDE': 1,
                             'POWER': 2
                             });
          this.getPrecedence = function(){
            return this.precedenceOrder[this.getFieldValue('OP')];
          };
        */
        //simulate blocks using precedence (for testing)
        this.precedence = false; //default value
        if (this.type == "math_arithmetic") {
          this.precedence = true;
          this.precedenceOrder = {'ADD' : 0,
                             'MINUS': 0,
                             'MULTIPLY': 1,
                             'DIVIDE': 1,
                             'POWER': 2
                             };
          this.getPrecedence = function(){
            return this.precedenceOrder[this.getFieldValue('OP')];
          };
        }
        if(child.type == "math_number") {
          child.precedence = false; //if not precedence , it have 99 order
        } else if (child.type == "math_arithmetic") {
          child.precedence = true;
          child.precedenceOrder = {'ADD' : 0,
                             'MINUS': 0,
                             'MULTIPLY': 1,
                             'DIVIDE': 1,
                             'POWER': 2
                             };
          child.getPrecedence = function(){
            return this.precedenceOrder[this.getFieldValue('OP')];
          };
        } else {
          //others like 'remainder of' block, this may be declared in each math block
          child.precedence = true;
          child.getPrecedence = function() {
            return -1;
          } 
        } 
[...]

PR #3895, opened 2020-05-09
Use the Blockly AST for block toString

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/57

### Proposed Changes

The block ``toString`` method which is what's shown to users when a block is collapsed doesn't allow users to tell the difference between the block configuration: 
Screen Shot 2020-05-09 at 9 52 40 AM
Screen Shot 2020-05-09 at 9 53 14 AM

Instead, now that we have a Blockly AST, we can go through and build up a block's textual representation by traversing the tree. In addition, we can add parentheses around `Number` and `Boolean` connections to make it clearer to user the order of precedence. 
The previous example turns into (when collapsed):
Screen Shot 2020-05-09 at 9 52 48 AM

Finally, I run through the text array and remove any simple parentheses around single input blocks. eg: ``((10) * (50)) + (10)`` turns into ``(10 * 50) + 10``.

### Reason for Changes

Fix ambiguity around order of precedence of collapsed blocks. 

### Test Coverage

Tested in playground.

Tested on:

### Additional Information

Screen Shot 2020-05-09 at 9 30 11 AM
turns into
Screen Shot 2020-05-09 at 9 30 16 AM

Screen Shot 2020-05-09 at 9 33 25 AM
turns into
Screen Shot 2020-05-09 at 9 33 29 AM

@NeilFraser FYI, I know you had some thoughts about parenthesis in the bug, do you think this design makes things better in most cases?

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -12,6 +12,7 @@
 
 goog.provide('Blockly.Block');
 
+goog.require('Blockly.ASTNode');
 goog.require('Blockly.Blocks');
 goog.require('Blockly.Connection');
 goog.require('Blockly.Events');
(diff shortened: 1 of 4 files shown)


core/block.js
```diff
@@ -1306,23 +1307,93 @@ Blockly.Block.prototype.setCollapsed = function(collapsed) {
...
+  // around single field blocks.
+  for (var i = 2, l = text.length; i < l; i++) {
+    if (text[i - 2] == '(' && text[i] == ')') {
+      text[i - 2] = text[i - 1];
+      text.splice(i - 1, 2);
+      l -= 2;
+    }
+  }
+
+  text = text.join(' ').replace(/(\() | (\))/gmi, '$1$2').trim() || '???';
```
alschmiedt, line 1396: Add a comment explaining what this is doing

samelhusseini: Done.

samelhusseini:
Added tests. Added some thoughts inline.

tests/mocha/block_test.js
```diff
@@ -1680,4 +1680,116 @@ suite('Blocks', function() {
       });
     });
   });
+  suite('toString', function() {
+    var toStringTests = [
```
samelhusseini, line 1684: @moniika had some thoughts as I was adding these tests.
The problem at hand was that I needed tests for toString, I could go ahead and recreate block definitions of the blocks I want to tests, similar to what is done [here](https://github.com/google/blockly/blob/master/tests/mocha/astnode_test.js#L9) but that had two drawbacks: 
- In terms of coverage, that screams to me as re-inventing the wheel (surely there's a bunch of tests that we want to cover a whole group of baseline blocks), and we don't want to re-generate what that baseline is every time. 
- Built-in blocks would be much easier to use as I can use the playground to generate the xml needed.

That got me here, where I used a similar pattern adopted in blockly-samples, to create the tests in some intermediate format, and generate them in a forEach. I like this method in that it's less verbose, but this still doesn't address the fact that it would be nice if I didn't have to generate what the baseline set of blocks and their respective xml is every time a test of this nature needs to be added.

As an aside, it was also clunky to have to add the XML without multi-line comments, that's just wasted time there, where we can just enable multi-line comments in mocha tests (after all, we're not running our unit tests in IE).

I think we need some sort of baseline xml definitions, similar to those of screen shot [tests](https://github.com/google/blockly/tree/master/tests/screenshot/test_cases), that we can run actions on top of. In this case, I would add the expected output for each of these tests, and 
update the test runner to run my test for each of these baseline blocks.

@rachel-fenichel FYI

rachel-fenichel: +1 to multi-line comments; no comment yet on the rest.

rachel-fenichel: For test block definitions, all of the common ones (empty block, value block, statement block) should definitely move into test_helpers.js.  We can either define and undefine them each time (calling helpers) or define them globally for all tests in a wrapper setup function.

I slightly prefer defining and undefining in each suite that needs them, with something like `helpers_define_blocks('empty_block', 'value_block', ...)` and `helpers_undefine_blocks('empty_block', 'value_block', ...)'.

moniika: Definitely if it seems useful for multiple tests then adding block_name and xml (or function that generates xml allowing for customization of values) could be added to helpers.

We currently have some helpers for defining blocks. I would be in favor of a pattern where you could use helper to define blocks and that would keep track of the blocks that were defined during the test (and then you call a cleanup method that undefines the blocks without needing to remember what you defined). They aren't being used consistently, so cleanup is needed there (and also a convenient way to undefine isn't implemented yet).

PR #3904, opened 2020-05-18
Blockly registry

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
This is just an initial PR to get people's opinion and start a discussion about plugins. I tried to include a piece from all the different parts we have talked about.

### Proposed Changes
There are a few main parts of this PR

- Using a common registry for things like fields & renderers
- Using this registry and an 'plugins' option to replace built in components like toolbox & block_dragger
- A separate 'plugins' registry on workspace that would allow a user to add external plugins like workspace search
  - This fixes two problems with workspace search: resize properly when window changes & hide itself when setVisible(false) called on workspace.
  - I added an interface for a ui plugin and a normal plugin that workspace search implements. We might want more or different interfaces, but this is just to get an idea.

I tried to separate the commits so that they are easy to follow

Some Notes:
-  Actually checking if a class implements an interface seems like a non trivial task in js
 - I can look into this more, but is this something we want to include in our registry?
- I've included some examples of plugins in the tests/plugin folder. 
### Test Coverage

Tested on:

### DIFF
--- a/core/field_registry.js
+++ b/core/field_registry.js
@@ -25,7 +25,7 @@ Blockly.fieldRegistry.typeMap_ = {};
 
 /**
  * Registers a field type.
- * Blockly.fieldRegistry.fromJson uses this registry to
+ * Blockly.registry.fieldFromJson uses this registry to
  * find the appropriate field type.
  * @param {string} type The field type name as used in the JSON definition.
  * @param {!{fromJson: Function}} fieldClass The field class containing a
--- a/core/field_registry.js
+++ b/core/field_registry.js
@@ -14,18 +14,12 @@
 
 goog.provide('Blockly.fieldRegistry');
 
+goog.require('Blockly.registry');
 
-/**
- * The set of all registered fields, keyed by field type as used in the JSON
- * definition of a block.
- * @type {!Object<string, !{fromJson: Function}>}
- * @private
- */
-Blockly.fieldRegistry.typeMap_ = {};
 
 /**
  * Registers a field type.
- * Blockly.registry.fieldFromJson uses this registry to
+ * Blockly.fieldRegistry.fromJson uses this registry to
  * find the appropriate field type.
  * @param {string} type The field type name as used in the JSON definition.
  * @param {!{fromJson: Function}} fieldClass The field class containing a
@@ -35,18 +29,7 @@ Blockly.fieldRegistry.typeMap_ = {};
  *     function.
  */
 Blockly.fieldRegistry.register = function(type, fieldClass) {
-  if ((typeof type != 'string') || (type.trim() == '')) {
-    throw Error('Invalid field type "' + type + '". The type must be a' +
-      ' non-empty string.');
-  }
-  if (Blockly.fieldRegistry.typeMap_[type]) {
-    throw Error('Error: Field "' + type + '" is already registered.');
-  }
-  if (!fieldClass || (typeof fieldClass.fromJson != 'function')) {
-    throw Error('Field "' + fieldClass + '" must have a fromJson function');
-  }
-  type = type.toLowerCase();
-  Blockly.fieldRegistry.typeMap_[type] = fieldClass;
+  Blockly.registry.register('field', type, fieldClass);
 };
 
 /**
@@ -54,12 +37,7 @@ Blockly.fieldRegistry.register = function(type, fieldClass) {
  * @param {string} type The field type name as used in the JSON definition.
  */
 Blockly.fieldRegistry.unregister = function(type) {
-  if (Blockly.fieldRegistry.typeMap_[type]) {
-    delete Blockly.fieldRegistry.typeMap_[type];
-  } else {
-    console.warn('No field mapping for type "' + type +
-        '" found to unregister');
-  }
+  Blockly.registry.unregister('field', type);
 };
 
 /**
@@ -73,8 +51,7 @@ Blockly.fieldRegistry.unregister = function(type) {
  * @package
  */
 Blockly.fieldRegistry.fromJson = function(options) {
-  var type = options['type'].toLowerCase();
-  var fieldClass = Blockly.fieldRegistry.typeMap_[type];
+  var fieldClass = Blockly.registry.getClass('field', options['type']);
   if (!fieldClass) {
     console.warn('Blockly could not create a field of type ' + options['type'] +
       '. The field is probably not being registered. This could be because' +
--- a/core/interfaces/i_plugin.js
+++ b/core/interfaces/i_plugin.js
@@ -1,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview The interface for a plugin.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IPlugin');
+
+
+/**
+ * @interface
+ */
+Blockly.IPlugin = function() {};
+
+/**
+ * Initializes the plugin.
+ * @return {void}
+ */
+Blockly.IPlugin.prototype.init;
+
+/**
+ * Disposes of the plugin. Should completely clean up after the plugin.
+ * @return {void}
+ */
+Blockly.IPlugin.prototype.dispose;
--- a/core/interfaces/i_toolbox.js
+++ b/core/interfaces/i_toolbox.js
@@ -1,0 +1,92 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview The interface for a toolbox.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IToolbox');
+
+goog.require('Blockly.IUIPlugin');
+
+/**
+ * @extends {Blockly.IUIPlugin}
+ * @interface
+ */
+Blockly.IToolbox = function() {};
+
+/**
+ * Get the toolbox flyout.
+ * @return {Blockly.Flyout} The toolbox flyout.
+ */
+Blockly.IToolbox.prototype.getFlyout;
+
+/**
+ * Unhighlight any previously specified option.
+ * @return {void}
+ */
+Blockly.IToolbox.prototype.clearSelection;
+
+/**
+ * Updates the category colours and background colour of selected categories.
+ */
+Blockly.IToolbox.prototype.updateColourFromTheme;
+
+/**
+ * Get the width of the toolbox.
+ * @return {number} The width of the toolbox.
+ */
+Blockly.IToolbox.prototype.getWidth;
+
+/**
+ * Get the height of the toolbox.
+ * @return {number} The width of the toolbox.
+ */
+Blockly.IToolbox.prototype.getHeight;
+
+/**
+ * Return the deletion rectangle for this toolbox.
+ * @return {Blockly.utils.Rect} Rectangle in which to delete.
+ * TODO: Should be part of deletable?
+ */
+Blockly.IToolbox.prototype.getClientRect;
+
+/**
+ * Handles the given Blockly action on a toolbox.
+ * This is only triggered when keyboard accessibility mode is enabled.
+ * @param {!Blockly.Action} action The action to be handled.
+ * @return {boolean} True if the field handled the action, false otherwise.
+ */
+Blockly.IToolbox.prototype.onBlocklyAction;
+
+/**
+ * Adds a style on the toolbox. Usually used to change the cursor.
+ * @param {string} style The name of the class to add.
+ */
+Blockly.IToolbox.prototype.addStyle;
+
+/**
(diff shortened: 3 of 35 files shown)


core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+Blockly.IToolbox.prototype.getClientRect;
+
+/**
+ * Handles the given Blockly action on a toolbox.
+ * This is only triggered when keyboard accessibility mode is enabled.
+ * @param {!Blockly.Action} action The action to be handled.
+ * @return {boolean} True if the field handled the action, false otherwise.
+ */
+Blockly.IToolbox.prototype.onBlocklyAction;
```
samelhusseini, line 66: This should be lifted into a separate interface, one that implements `IBlocklyAccessibility` or something like that.

core/interfaces/i_plugin.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+/**
+ * @fileoverview The interface for a plugin.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IPlugin');
```
samelhusseini, line 14: I think we should have one more interface hierarchically that's above this that is something like `IRegistrable`. This would be a shallow interface in that it doesn't have any methods you need to implement, but it's a head nod to any object such as Field / Insertion marker / Renderer that we expect users to register. That can also be the abstract type of the object being registered in `Blockly.registry.register`

core/renderers/common/block_rendering.js
```diff
@@ -33,33 +33,6 @@ Blockly.blockRendering.rendererMap_ = {};
...
 Blockly.blockRendering.useDebugger = false;
 
-/**
- * Registers a new renderer.
- * @param {string} name The name of the renderer.
- * @param {!Function} rendererClass The new renderer class
- *     to register.
- * @throws {Error} if a renderer with the same name has already been registered.
- */
-Blockly.blockRendering.register = function(name, rendererClass) {
```
samelhusseini, line 43: Don't remove, deprecate and internally use the new registry.

core/registry.js
```diff
@@ -0,0 +1,106 @@
...
+  delete Blockly.registry.typeMap_[type][name];
+};
+
+/**
+ * Get the class for the given name and type.
+ * @param {string} type The type of the plugin. (Ex: Field, Toolbox)
+ * @param {string} name The name of the plugin. (Ex: field_angle)
+ * @return {Object} The class with the given name and type or null if none exists.
+ */
+Blockly.registry.getClass = function(type, name) {
```
samelhusseini, line 81: Should this be a factory instead? ie: create a new instance of the class rather than return a reference to the class type?
If so, maybe ``Blockly.registry.createInstanceOf(type, name)``

core/workspace_svg.js
```diff
@@ -449,6 +458,25 @@ Blockly.WorkspaceSvg.prototype.setMarkerSvg = function(markerSvg) {
...
+Blockly.WorkspaceSvg.prototype.addPlugin = function(type, pluginClass, params) {
+  if (!this.plugins[type]) {
+    this.plugins[type] = [];
+  }
+  var newPlugin = new pluginClass(this, params);
+  newPlugin.init();
+  this.plugins[type].push(newPlugin);
+
+  // List of all plugins for dispose.
+  this.plugins['all'].push(newPlugin);
```
samelhusseini, line 477: Why is this needed, wouldn't it be enough to go through `Object.keys` on `this.plugins` to get all of them? or am i missing something?

core/workspace_svg.js
```diff
@@ -993,6 +1024,11 @@ Blockly.WorkspaceSvg.prototype.resize = function() {
   if (this.scrollbar) {
     this.scrollbar.resize();
   }
+  var uiPlugins = this.plugins['uiplugin'];
+  for (var i = 0; uiPlugins && i < uiPlugins.length; i++) {
+    var plugin = uiPlugins[i];
+    plugin.position();
+  }
```
samelhusseini, line 1031: This is interesting, because only plugins that are registered as type `uiplugin` will be positioned here. In the case of the Toolbox, it's of type 'toolbox' and 'uiplugin' (given both the inheritance tree in the interface, and the fact that it implements `position()`), so perhaps what you're really after here, is get me all registered components that implement position?
If so, we should think about how the registry can solve for that.

tests/playground.html
```diff
@@ -67,6 +67,9 @@
 <script src="blocks/test_blocks.js"></script>
 <script src="themes/test_themes.js"></script>
 <script src="./playgrounds/screenshot.js"></script>
+<script src="./plugins/toolbox_new.js"></script>
```
samelhusseini, line 70: Maybe this should live under playgrounds.

core/registry.js
alschmiedt: I think the problem with this is that we don't know what parameters are going to come in for what we are creating.

tests/playground.html
```diff
@@ -142,8 +146,14 @@
             maxScale: 4,
             minScale: 0.25,
             scaleSpeed: 1.1
-          }
+          },
+        plugins: {
+          'toolbox': 'newToolbox',
+          'connectionManager': 'draggedConnections',
```
samelhusseini, line 152: Okay if the goal was to demonstrate the use of `workspace.addPlugin`, but would it have been possible to register workspace search in options, eg: 
```
{
  plugins: {
      'workspacesearch': Blockly.WorkspaceSearch
  }
}
```

core/workspace_svg.js
alschmiedt: Yeah I think I was trying to communicate something else and failed lol 
I was trying to get at that plugins can have some kind of hierarchy:
plugin ->  uiPlugin -> some_other_plugin
If we have some_other_plugin then we want it to go in it's own list as well as the uiPlugin list so that we can easily get all the things that need to change without having to search through.

core/workspace_svg.js
alschmiedt: I am not convinced I would want to get all registered components that implement position. It seems like that can get messy if people have methods on their classes that they don't realize are part of a different interfaces.

tests/playground.html
alschmiedt: They can, I'm not sure I was planning on even including these in final pr, just seemed useful for discussion purposes.

core/interfaces/i_plugin.js
```diff
@@ -0,0 +1,32 @@
...
+ * Initializes the plugin.
+ * @return {void}
+ */
+Blockly.IPlugin.prototype.init;
+
+/**
+ * Disposes of the plugin. Should completely clean up after the plugin.
+ * @return {void}
+ */
+Blockly.IPlugin.prototype.dispose;
```
samelhusseini, line 32: An `IDisposable` candidate.

core/interfaces/i_plugin.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @interface
+ */
+Blockly.IPlugin = function() {};
+
+/**
+ * Initializes the plugin.
+ * @return {void}
+ */
+Blockly.IPlugin.prototype.init;
```
samelhusseini, line 26: Many UI plugin candidates (eg: trashcan, zoomcontrols) have both init and createDom. We should try and formalize when to use which and create a subtype of UIPlugin that implements createDom.

core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Get the height of the toolbox.
+ * @return {number} The width of the toolbox.
+ */
+Blockly.IToolbox.prototype.getHeight;
+
+/**
+ * Return the deletion rectangle for this toolbox.
+ * @return {Blockly.utils.Rect} Rectangle in which to delete.
+ * TODO: Should be part of deletable?
```
samelhusseini, line 56: This is used for delete areas rather than deletable objects which is what IDeletable is intended for. I would say an `IDeleteArea` interface that Toolbox / Flyout / Trashcan can implement.

tests/playground.html
samelhusseini: Totally, then please ignore.

core/registry.js
samelhusseini: Ah makes sense. Technically we could use variable arguments here and pass them along, but I think you're right it cleaner to just get the class by type and then call it with the arguments separately. 
With that in my I would recommend you cast the type before calling the constructor. This is so the typechecker can check the arguments and their type.

core/workspace_svg.js
samelhusseini: I see, so in that case you would register a toolbox as both: 
``workspace.addPlugin('toolbox', Blockly.Toolbox)``
and 
``workspace.addPlugin('uiplugin', Blockly.Toolbox)``

or perhaps a variation of the above with an array: 
``workspace.addPlugin(['toolbox', 'uiplugin'], Blockly.Toolbox)``
?

core/workspace_svg.js
samelhusseini: Well there are other ways to tell if a class implements a specific interface. Unfortunately JS doesnt have that built in, but perhaps that interface can require a property such as:
`Blockly.UIPlugin.prototype.IS_UI_PLUGIN = true` and all plugins that have that bit set to true will also have position implemented (as we know they implement the UIPlugin interface and hence can safely call position on them)

core/workspace_svg.js
alschmiedt: Yeah basically. I kind of wanted to take away that work from the developer and do it for them in the cases that we know. For example, toolbox is always going to be a uiplugin and we can add it for them. But this might be too convoluted on our end.

core/workspace_svg.js
alschmiedt: Ah, yeah that makes sense. We can potentially use that information for figuring out what lists to add it to on the workspace?

core/renderers/common/block_rendering.js
alschmiedt: done

core/workspace_svg.js
samelhusseini: I was thinking more along the lines that all plugins are in one list. You can use a property like that to filter for plugins that implement a specific interface.

core/field_registry.js
```diff
@@ -73,8 +51,7 @@ Blockly.fieldRegistry.unregister = function(type) {
  * @package
  */
 Blockly.fieldRegistry.fromJson = function(options) {
-  var type = options['type'].toLowerCase();
-  var fieldClass = Blockly.fieldRegistry.typeMap_[type];
+  var fieldClass = Blockly.registry.getClass('field', options['type']);
```
samelhusseini, line 54: Why not call ``Blockly.registry.getClassFromOptions``?

alschmiedt: Because options here is not actually Blockly.Options. The getClassFromOptions is specifically looking for if someone has passed in the type to the plugins object.

core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @fileoverview The interface for a toolbox.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IToolbox');
+
+goog.require('Blockly.IUIPlugin');
```
rachel-fenichel, line 16: RequireType?

rachel-fenichel:
After reading through this, all of my questions are about interfaces:
- How do we decide when to add a new interface?
- Is it reasonable to add tons of small interfaces? (e.g. copyable, bounding box, deletable, etc)
- What if there's a function that's part of multiple interfaces?
- How does a developer explore the interfaces we have and know what to implement?
- As you mentioned above, how can we check whether a developer actually implemented a specific interface?

BeksOmega:
> As you mentioned above, how can we check whether a developer actually implemented a specific interface?

I don't know if this will work here, but maybe you could try using [go-like interfaces](https://tour.golang.org/methods/10). In golang, if the object has matching functions for all of the functions defined in the interface, then it fulfills the interface.

From the docs: "Implicit interfaces decouple the definition of an interface from its implementation"

Probably won't work here, but I thought it would be cool to mention it =)

samelhusseini:
Closure’s @record (structural interface) has similar behavior to go interfaces: https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler#structural-interfaces

rachel-fenichel:
Closure compiler interfaces can protect us within the library, but my understanding is that they don't help in plugin-land unless the plugin developer is using typescript.

samelhusseini:
That is correct.
If the developer is using TypeScript, they can simply `implements` an interface and the type checker will ensure the interface is implemented correctly. 
A JavaScript developer has a little less assistance, but:
 - In some editors, (eg: vscode), the language service will load the typescript definitions and run similar checks on the code, although ES6 doesn't have `implements`, the language service would read the `@implements` JSDoc tag.
- A JS developer can read the interfaces directly to see what methods need implementing (old school way), but still relevant / useful.

alschmiedt:
If we want to provide more error checking up front we could look at doing something like [this](https://stackoverflow.com/questions/13041168/a-running-example-or-working-demo-of-interface-pattern-in-javascript).

PR #3912, opened 2020-05-21
Strict types in input

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Strict types in input.js.
Added prefixField and suffixField properties to FieldDropdown 
Add rendered property to Blockly.Block and set to null such that:
 - Blockly.Block, rendered = null
 - Blockly.BlockSvg, rendered = false
 - Blockly.BlockSvg (after render is called), rendered = true

### Test Coverage

Tested on:

### DIFF
--- a/core/input.js
+++ b/core/input.js
@@ -109,19 +109,21 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
   field.name = opt_name;
   field.setVisible(this.isVisible());
 
-  if (field.prefixField) {
+  var fieldDropdown = /** @type {Blockly.FieldDropdown} */ (field);
+  if (fieldDropdown.prefixField) {
     // Add any prefix.
-    index = this.insertFieldAt(index, field.prefixField);
+    index = this.insertFieldAt(index, fieldDropdown.prefixField);
   }
   // Add the field to the field row.
   this.fieldRow.splice(index, 0, field);
   ++index;
-  if (field.suffixField) {
+  if (fieldDropdown.suffixField) {
     // Add any suffix.
-    index = this.insertFieldAt(index, field.suffixField);
+    index = this.insertFieldAt(index, fieldDropdown.suffixField);
   }
 
   if (this.sourceBlock_.rendered) {
+    this.sourceBlock_ = /** @type {!Blockly.BlockSvg} */ (this.sourceBlock_);
     this.sourceBlock_.render();
     // Adding a field will cause the block to change shape.
     this.sourceBlock_.bumpNeighbours();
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
`rendered` already is overloaded. Is there a different name that we can use to start fresh and always give the meanings you listed above?

Also, what is the value after `dispose` has been called?

samelhusseini:
Hmm, that's not a choice I'm making. In places where we do things like: 
```
if (this.sourceBlock_.rendered) {
 ...
}
```
If the type of sourceBlock_ is `Blockly.Block`, then the compiler is going to complain because that property doesn't exist on Blockly.Block, only on Blockly.BlockSvg, so by adding it to the Block and setting it null, it's a non-functional change (since we don't check for the nullability of that property, only for its value), it lets the compiler know that that property can exist.

For Blockly.Block, it's value is always null. 
For Blockly.BlockSvg, it's value is false after dispose.

rachel-fenichel:
Oh, you're just adding it to Block and it's already on BlockSvg, and you're not changing value.  Never mind.

core/input.js
```diff
@@ -109,19 +109,21 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
   field.name = opt_name;
   field.setVisible(this.isVisible());
 
-  if (field.prefixField) {
+  var fieldDropdown = /** @type {Blockly.FieldDropdown} */ (field);
```
rachel-fenichel, line 112: I don't think it's safe to cast this to `FieldDropdown`.

samelhusseini: Not safe to cast the `field` var, but I don't see why its not safe to create a new var that is the cast of field as a `FieldDropdown`.

rachel-fenichel: You're casting so that you can access the field without the type checker warning, I assume. I guess it's technically safe but normally we cast when we already know the type and it's just that we've lost that information along the way, which isn't true here.

samelhusseini: Not quite, a type cast is not always an assertion that we know what the type is. In this case it's necessary to type cast because `prefixField` does not exist on type `Blockly.Field`, so from the compiler's perspective how can you say ``if (field.prefixField) `` if there is no such thing, instead what the cast is saying is assume that the field is a `Blockly.FieldDropdown` then proceed with that check as follows.

Perhaps the reason this seemed weird to you is because (you're right), we haven't made that assertion in the code. The above is a non-functional change, perhaps a cleaner change would be to say: 
- either cast it to an interface that implements prefixField, that way the code reads: any field that implements prefixField fits in here and should proceed in this code path, which is functionally what it's doing, but we're making assumptions here that only the DropdownField ever goes down this path.
- another functional alternative here is to check whether or not the field is an instanceof `Blockly.FieldDropdown` and only then proceed down the path, the compiler will implicitly cast the field to a `FieldDropdown` and "know" that prefixField exists.

The interface option here is overhead I feel, but an instanceof check would work. Only problem there is I don't think Blockly.FieldDropdown is one of the "required" fields, and so we can't assume that module exists, so we're back at "assuming".

rachel-fenichel: I see.  In that case, the cast is fine--it's not any worse than what we were doing before--and I'll keep this in mind as another corner.

Issue #3746, opened 2020-03-12
Create a Public Metrics Method 


alschmiedt:
**Is your feature request related to a problem? Please describe.**
Currently the only way to get metrics is by using the same method we use internally for processes that need to know just about everything about the workspace. Therefore, the method returns a lot of information that is not well organized or very well documented. 

**Describe the solution you'd like**
A metrics method that is meant to be public and provides a subset of the current values returned from getMetrics with easier to understand and better documented values. 

**Describe alternatives you've considered**

**Additional context**
Possible Use Cases: 
- Where to position something relative to the toolbars. 
- For scrolling to blocks

alschmiedt:
So far it seems like the following methods would be most useful: 

**getAbsoluteMetrics** (or different name)

Gets the metrics for the workspace without the toolbox.
This would be used for anything that needs to be positioned on top of the workspace. 
**Examples**: WorkspaceSearch, Trashcan, ZoomControl, Flyouts, Modals*
returns {
   absoluteLeft  
   absoluteTop 
   absoluteWidth
   absoluteHeight
}
*Modals would have to be used with workspace.getParentSvg().getClientBoundingRect() since the modal is outside of the blockly div(because it uses the widgetDiv)

**getContentMetrics** (probably different name. I don't like this name.)

This would hold what is currently the content and view metrics from our current getMetrics method.
This would be used for anything that needs to be positioned on the workspace or need information on how the workspace is scrolled.
**Examples**: Bubble, Scrolling on workspace

**getToolboxMetrics**

Contain the metrics for both the flyout and toolbox
Not currently used internally anywhere except flyout_vertical and flyout_horizontal. And I think that might be able to be factored out.

alschmiedt:
Related concerns:
* Having some way to display debug rectangles similar to rendering_debug would be nice.
* I think making these methods public would fix the problem where users who want to create their own `getMetrics` method but don't want to have to copy and paste all of the private `getTopLevelWorkspaceMetrics_`.

maribethb:
Another use case would be preventing blocks from going behind an always-open flyout (more specifically, a flyout on a toolbox where the flyout has autoClose = true. This already works when the workspace contains just a flyout and no toolbox). Currently to achieve this, most of getTopLevelWorkspaceMetrics is fine, but we need to change some properties to take into account the width of both toolbox + flyout. So having this method public would allow me to call it from a custom getMetrics method without having to copy/paste a bunch of code I don't need to change.

[...]

PR #3913, opened 2020-05-21
Add Blockly.utils.Metrics @record

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Partly https://github.com/google/blockly/issues/3746

### Proposed Changes

Add a Blockly.utils.Metrics @record and use that instead of {Object} when referring to metrics.

### Test Coverage

Passes strict type checks
Passes build
Tested playground.

Tested on:


core/inject.js
```diff
@@ -29,6 +29,8 @@ goog.require('Blockly.utils.userAgent');
 goog.require('Blockly.WorkspaceDragSurfaceSvg');
 goog.require('Blockly.WorkspaceSvg');
 
+goog.requireType('Blockly.utils.Metrics');
```
alschmiedt, line 32: Does this need to be here?

core/mutator.js
```diff
@@ -404,14 +406,20 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
...
 Blockly.Mutator.prototype.getFlyoutMetrics_ = function() {
   return {
     viewHeight: this.workspaceHeight_,
     viewWidth: this.workspaceWidth_ - this.workspace_.getFlyout().getWidth(),
+    contentHeight: 0,
+    contentWidth: 0,
+    viewTop: 0,
+    viewLeft: 0,
+    contentTop: 0,
+    contentLeft: 0,
```
alschmiedt, line 422: Can we make these optional? Setting something like contentHeight to 0 seems a bit strange to me

core/utils/metrics.js
```diff
@@ -0,0 +1,124 @@
...
+ * Height of the flyout if it is always open.  Otherwise zero.
+ * @type {number|undefined}
+ */
+Blockly.utils.Metrics.prototype.flyoutHeight;
+
+/**
+ * Top, bottom, left or right. Use TOOLBOX_AT constants to compare.
+ * @type {number|undefined}
+ */
+Blockly.utils.Metrics.prototype.toolboxPosition;
```
alschmiedt, line 124: nit: move this up to be with the toolbox group

core/blockly.js
```diff
@@ -107,13 +108,11 @@ Blockly.EventData;
...
  * @param {!SVGElement} svg SVG image.
- * @return {!Object} Contains width and height properties.
+ * @return {!Blockly.utils.Size} Contains width and height properties.
  */
 Blockly.svgSize = function(svg) {
-  return {
-    width: svg.cachedWidth_,
-    height: svg.cachedHeight_
-  };
+  svg = /** @type {?} */ (svg);
```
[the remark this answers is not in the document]
samelhusseini: Because an SVG element doesn't have `cachedHeight_` and `cachedHeight_`, these are properties we've added onto the element. It is for this reason that I'm not a fan of adding properties onto DOM elements.

core/flyout_base.js
```diff
@@ -932,3 +949,44 @@ Blockly.Flyout.prototype.onBlocklyAction = function(action) {
   var cursor = this.workspace_.getCursor();
   return cursor.onBlocklyAction(action);
 };
+
```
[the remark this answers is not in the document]
samelhusseini: Since we're talking about interfaces, this is an `abstract` class. We refer to Blockly.Flyout everywhere in WorkspaceSvg / Toolbox as the abstract Blockly.Flyout, but we never actually instantiate a Blockly.Flyout, it's always either Vertical or Horizontal. These methods need to be implemented by the subclass for everything to function properly (ie: we make assumptions in other parts of Blockly that these methods exist on a Flyout).

Which leads me to, we should probably mark this class as `@abstract`. Will look into that.

core/inject.js
samelhusseini: Yes, it's used in the getMetrics definition.

core/mutator.js
samelhusseini: We could but they create a bigger headache elsewhere where we make assumptions that they exist, I will give it a go and see how bad it is.

core/utils/metrics.js
samelhusseini: I wanted to keep the ones that could be undefined separately and ordered as they are ordered in getMetrics, ie: conceptually.
I could be order them alphabetically into two groups if you feel that would help?

core/flyout_base.js
```diff
@@ -232,11 +243,15 @@ Blockly.Flyout.prototype.createDom = function(tagName) {
  *     create new blocks.
  */
 Blockly.Flyout.prototype.init = function(targetWorkspace) {
-  this.targetWorkspace_ = targetWorkspace;
+  this.targetWorkspace = targetWorkspace;
   this.workspace_.targetWorkspace = targetWorkspace;
 
   // Add scrollbar.
```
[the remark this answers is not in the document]
samelhusseini: noted.

core/bubble.js
```diff
@@ -535,10 +537,10 @@ Blockly.Bubble.prototype.getOverlap_ = function(relativeMin, metrics) {
  * Calculate what the optimal horizontal position of the top-left corner of the
  * bubble is (relative to the anchor point) so that the most area of the
  * bubble is shown.
- * @param {!Object} metrics The metrics of the workspace the bubble will
- *    appear in.
+ * @param {!Blockly.utils.Metrics} metrics The metrics of the workspace the
```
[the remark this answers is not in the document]
samelhusseini: I think that would be leading, getOptimalRelativeTop_ and getOptimalRelativeLeft_ methods that take that parameter don't do any of the positioning themselves, they just calculate where the positioning is.

core/mutator.js
samelhusseini: I looked into this, and I think it's not worth having to assert whether or not these values exist all of the place.

core/flyout_base.js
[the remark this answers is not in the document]
samelhusseini: Done, made Blockly.Flyout abstract.

core/mutator.js
alschmiedt: Ok, could we add a comment explaining why we are setting these values to 0?

core/utils/metrics.js
alschmiedt: Yeah that could help, I just want them to be ordered in some way that I understand where to put something new when looking at it.

core/mutator.js
samelhusseini: Sounds good.

samelhusseini: The mutator workspace only uses a subset of Blockly.utils.Metrics as features such as scroll and zoom are unsupported.

core/utils/metrics.js
samelhusseini: Totally, they are currently grouped as they are in the "main" getMetrics method https://github.com/google/blockly/blob/develop/core/workspace_svg.js#L2425. I don't like alphabetical grouping here because its nice having width and height of the same category coupled together. 

Happy to bump toolboxPosition so that it's closer to toolboxWidth and toolboxHeight if that makes more sense to you, but I would say we should apply the same order in `Blockly.WorkspaceSvg.getTopLevelWorkspaceMetrics_`.

samelhusseini: Also, updated ordering of metrics to match everywhere.

core/mutator.js
alschmiedt: That looks great, thanks!

PR #3914, opened 2020-05-21
Registry

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Add a global registry. Refactor field and renderer registries to use this registry. 

### Reason for Changes
Nice to not have to revalidate everything every single time

### Test Coverage

Tested on:


core/registry.js
```diff
@@ -0,0 +1,174 @@
...
+ * @private
+ */
+Blockly.registry.validateProperties_ = function(type, requiredProperties, registryClass) {
+  var unimplemented = requiredProperties.filter(function(property) {
+    return !registryClass.hasOwnProperty(property) ||
+        (typeof registryClass[property] != 'function');
+  });
+  if (unimplemented.length) {
+    throw Error('Type "' + type + '" requires the following properties "' +
+        unimplemented + '"');
```
samelhusseini, line 128: Should we join the array here?

core/registry.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+};
+
+/**
+ * Unregisters the class with the given type and name.
+ * @param {string|Blockly.registry.Type<T>} type The type of the plugin.
+ *     (eg: Field, Renderer)
+ * @param {string} name The plugin's name. (Ex. field_angle, geras)
```
samelhusseini, line 136: nit: using eg: and Ex. interchangeably. Choose one

core/registry.js
```diff
@@ -0,0 +1,174 @@
...
+ * @template T
+ */
+Blockly.registry.unregister = function(type, name) {
+  type = String(type);
+  var typeRegistry = Blockly.registry.typeMap_[type.toLowerCase()];
+  if (!typeRegistry) {
+    console.warn('No type "' + type + '" found');
+    return;
+  }
+  if (!typeRegistry[name]) {
```
samelhusseini, line 146: Name is not normalized before this check, why not just:
```
type = String(type).toLowerCase()
name = name.toLowerCase()
```
at the top? If it's so the error message reads the value that the developer gave, I would suggest otherwise (better that the developer see what value we're checking against).

core/renderers/common/block_rendering.js
```diff
@@ -1,4 +1,5 @@
 /**
+ *
```
samelhusseini, line 2: nit: remove

tests/mocha/registry_test.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
samelhusseini, line 3: nit: 2020

tests/mocha/registry_test.js
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Tests for Blockly.fieldRegistry
```
samelhusseini, line 8: nit: Blockly.registry

tests/mocha/registry_test.js
```diff
@@ -0,0 +1,55 @@
...
+ * @author bekawestberg@gmail.com (Beka Westberg)
```
samelhusseini, line 9: nit: update author to you

core/registry.js
alschmiedt: done

core/registry.js
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Returns the name of the type.
+ * @return {string} The name.
+ * @override
+ */
+Blockly.registry.Type.prototype.toString = function() {
+  return this.name_;
+};
+
+/** @type {!Blockly.registry.Type<Blockly.blockRendering.Renderer>} */
```
rachel-fenichel, line 46: Are you doing it this way instead of with an enum so that it's easy for someone else to add a type without changing core?

rachel-fenichel:
This looks really solid :+1:

core/registry.js
samelhusseini: Using generics this way allows the closure compiler to infer the type of the class in `register` and `getClass`, when the `Blockly.registry.Type` wrapper is passed in. ie: 
If you do: 
```
Blockly.registry.getClass(Blockly.registry.Type.RENDERER, 'zelos');
```
Using generics the closure compiler knows that the return type of that method is going to be a `Blockly.blockRendering.Renderer` type. 
However, if you do instead: 
```
Blockly.registry.getClass('renderer', 'zelos')
```
It wont. Since we cast that type to string, the two methods yield the same value, it's just the first one lets us internally infer the typechecks and: 
- a, not have to cast everywhere
- b, ensure that a type within the registry is consistent across, ie: all registered field types 
inherit from `Blockly.Field`.

This method avoids us having to add ``registerRenderer`, `registerField``, etc wrapper functions, and instead use the single register method but still type check.

Unfortunately, the parameter args of the class we're instantiating are still not type checked.

More context here: https://github.com/alschmiedt/blockly/pull/1

rachel-fenichel: That's awesome. Gets around a lot of the problems we were discussing. Very nice. :+1:

samelhusseini: Yep, it takes a little to get your head around it when reading the code, but I think it's worth it in this case.

PR #3918, opened 2020-05-26
Toolbox interface

alschmiedt:
## The basics

- [x] I branched from develop

## The details
I left the toolbox interface in mainly one piece other than the items that were optional for the toolbox (delete are & accessibility). 
Planning on separating out other interface pieces from the toolbox in separate PR
### Proposed Changes
1. Adds a toolbox interface
2. Adds a delete area interface
3. Adds a registrable interface (For objects that can be registered)

### Reason for Changes
Working on making a toolbox that can be used instead of this toolbox.

### Test Coverage

Tested on:

### DIFF
--- a/core/interfaces/i_deletearea.js
+++ b/core/interfaces/i_deletearea.js
@@ -1,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview The interface for a component that can delete a block that is
+ *   dropped on top of it.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IDeleteArea');
+
+
+/** @interface */
+Blockly.IDeleteArea = function() {};
+
+/**
+ * Return the deletion rectangle.
+ * @return {Blockly.utils.Rect} Rectangle in which to delete.
+ */
+Blockly.IDeleteArea.prototype.getClientRect;
--- a/core/interfaces/i_deletearea.js
+++ b/core/interfaces/i_deletearea.js
@@ -15,7 +15,10 @@
 goog.provide('Blockly.IDeleteArea');
 
 
-/** @interface */
+/**
+ * Interface for a component that can delete a block that is dropped on top of it.
+ * @interface
+ */
 Blockly.IDeleteArea = function() {};
 
 /**
--- a/core/interfaces/i_toolbox.js
+++ b/core/interfaces/i_toolbox.js
@@ -13,10 +13,11 @@
 
 goog.provide('Blockly.IToolbox');
 
-goog.require('Blockly.IRegistrable');
+goog.requireType('Blockly.IRegistrable');
 
 
 /**
+ * Interface for a toolbox.
  * @extends {Blockly.IRegistrable}
  * @interface
  */
@@ -24,6 +25,7 @@ Blockly.IToolbox = function() {};
 
 /**
  * Initializes the toolbox.
+ * @return {void}
  */
 Blockly.IToolbox.prototype.init;
 
@@ -31,11 +33,13 @@ Blockly.IToolbox.prototype.init;
  * Fill the toolbox with categories and blocks.
  * @param {Array.<Blockly.utils.toolbox.Toolbox>} toolboxDef Array holding objects
  *    containing information on the contents of the toolbox.
+ * @return {void}
  */
 Blockly.IToolbox.prototype.render;
 
 /**
  * Dispose of this toolbox.
+ * @return {void}
  */
 Blockly.IToolbox.prototype.dispose;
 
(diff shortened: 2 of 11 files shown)


core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,106 @@
...
+'use strict';
+
+goog.provide('Blockly.IToolbox');
+
+goog.require('Blockly.IDisposable');
+goog.require('Blockly.IRegistrable');
+
+
+/**
+ * @extends {Blockly.IDisposable}
```
samelhusseini, line 21: Where is `Blockly.IDisposable` defined?

core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,106 @@
...
+ * Fill the toolbox with categories and blocks.
+ * @param {Array.<Blockly.utils.toolbox.Toolbox>} toolboxDef Array holding objects
+ *    containing information on the contents of the toolbox.
+ */
+Blockly.IToolbox.prototype.render;
+
+/**
+ * Dispose of this toolbox.
+ */
+Blockly.IToolbox.prototype.dispose;
```
samelhusseini, line 42: I would imagine this would already be in `Blockly.IDisposable`, so there's no need to repeat it here.

core/interfaces/i_deletearea.js
```diff
@@ -0,0 +1,25 @@
...
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IDeleteArea');
+
+
+/** @interface */
+Blockly.IDeleteArea = function() {};
```
samelhusseini, line 19: Not necessarily in this PR, but I imagine the goal here would be to turn `Blockly.WorkspaceSvg.prototype.recordDeleteAreas` and `Blockly.WorkspaceSvg.prototype.isDeleteArea` into an API that supports ANY delete area where areas get "registered" onto the workspace during startup, and these two methods just iterate through the list of delete areas completing the action at hand.

core/interfaces/i_toolbox.js
alschmiedt: Left in on accident. Should be removed from this PR.

core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,104 @@
...
+ * to a change in one of the dynamic categories, such as variables or
+ * procedures.
+ */
+Blockly.IToolbox.prototype.refreshSelection;
+
+/**
+ * Toggles the visibility of the toolbox.
+ * @param {boolean} isVisible True if the toolbox should be visible.
+ */
+Blockly.IToolbox.prototype.setVisible;
```
samelhusseini, line 99: toolbox.js doesn't implement this method?

core/interfaces/i_toolbox.js
samelhusseini: No worries, and glad you're thinking about moving that into its own interface (in a separate PR).

core/interfaces/i_toolbox.js
```diff
@@ -0,0 +1,111 @@
...
+'use strict';
+
+goog.provide('Blockly.IToolbox');
+
+goog.requireType('Blockly.IRegistrable');
+
+
+/**
+ * Interface for a toolbox.
+ * @extends {Blockly.IRegistrable}
```
samelhusseini, line 21: WorkspaceSvg makes the assumption that the toolbox is also a deletearea. Should we make it a requirement that all toolboxes be delete areas, in which case `@extends {Blockly.IDeleteArea}`

core/toolbox.js
```diff
@@ -37,6 +39,9 @@ goog.requireType('Blockly.IBlocklyActionable');
  *     blocks.
  * @constructor
  * @implements {Blockly.IBlocklyActionable}
+ * @implements {Blockly.IDeleteArea}
+ * @implements {Blockly.IRegistrable}
```
samelhusseini, line 43: `Blockly.IRegistrable` is redundant here, leave out. Just implements `Blockly.IToolbox`

core/interfaces/i_toolbox.js
alschmiedt: good catch, must have stashed when i changed branches.

core/interfaces/i_toolbox.js
alschmiedt: This pr removes this assumption by adding `typeof this.toolbox_.getClientRect == 'function'`before calling it. I don't really want to require that for someone creating an external toolbox

core/interfaces/i_deletearea.js
```diff
@@ -0,0 +1,25 @@
...
+ *   dropped on top of it.
+ * @author aschmiedt@google.com (Abby Schmiedt)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IDeleteArea');
+
+
+/** @interface */
```
samelhusseini, line 18: Add interface description.

core/toolbox.js
```diff
@@ -28,6 +28,9 @@ goog.require('Blockly.utils.Rect');
 goog.require('Blockly.utils.toolbox');
 
 goog.requireType('Blockly.IBlocklyActionable');
+goog.requireType('Blockly.IDeleteArea');
+goog.requireType('Blockly.IRegistrable');
```
samelhusseini, line 32: Remove IRegistrable from requires.

PR #3922, opened 2020-05-28
Add plugin options

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Add a way to override the toolbox in options. 
 
** I left out any extra validation for the toolbox. I still think it would be nice to have some kind of check that will warn people if their toolbox doesn't have the required methods, but don't have a great way to do this yet. 

### Test Coverage

Tested on:

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -179,7 +179,7 @@ Blockly.Options = function(options) {
 
   /**
    * Map of plugin type to name of registered plugin or plugin class.
-   * @type {!Object<string, (function(new:?, ...?)|string)>}
+   * @type {!Object.<string, (function(new:?, ...?)|string)>}
    */
   this.plugins = plugins;
 };
--- a/core/options.js
+++ b/core/options.js
@@ -74,7 +74,6 @@ Blockly.Options = function(options) {
       hasSounds = true;
     }
   }
-  var plugins = options['plugins'] || {};
   var rtl = !!options['rtl'];
   var horizontalLayout = options['horizontalLayout'];
   if (horizontalLayout === undefined) {
@@ -110,6 +109,8 @@ Blockly.Options = function(options) {
   var keyMap = options['keyMap'] || Blockly.user.keyMap.createDefaultKeyMap();
 
   var renderer = options['renderer'] || 'geras';
+
+  var plugins = options['plugins'] || {};
 
   /** @type {boolean} */
   this.RTL = rtl;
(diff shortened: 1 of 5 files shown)


core/options.js
```diff
@@ -110,6 +110,8 @@ Blockly.Options = function(options) {
 
   var renderer = options['renderer'] || 'geras';
 
+  var plugins = options['plugins'] || {};
```
samelhusseini, line 113: || Object.create(null)

core/options.js
```diff
@@ -174,6 +176,12 @@ Blockly.Options = function(options) {
    * @type {Blockly.Workspace}
    */
   this.parentWorkspace = options['parentWorkspace'];
+
+  /**
+   * Map of plugin type to name of registered plugin or plugin class.
+   * @type {!Object<string, (function(new:?, ...?)|string)>}
```
samelhusseini, line 182: nit: ``Object.<``

core/registry.js
```diff
@@ -152,3 +165,23 @@ Blockly.registry.getClass = function(type, name) {
...
+ * @param {!Blockly.Options} options The option object to check for the given
+ *     plugin.
+ * @return {?function(new:T, ...?)} The class for the plugin.
+ * @template T
+ */
+Blockly.registry.getClassFromOptions = function(type, options) {
+  var typeName = type.toString();
+  var plugin = options.plugins[typeName] || Blockly.registry.DEFAULT;
+
+  // If the user passed in a plugin class instead of a registered plugin name
```
samelhusseini, line 182: nit: end with `.`

core/workspace_svg.js
```diff
@@ -749,7 +750,9 @@ Blockly.WorkspaceSvg.prototype.createDom = function(opt_backgroundClass) {
     if (!Blockly.Toolbox) {
       throw Error('Missing require for Blockly.Toolbox');
     }
-    this.toolbox_ = new Blockly.Toolbox(this);
+    var toolboxClass = Blockly.registry.getClassFromOptions(
```
samelhusseini, line 753: Since this is a constructor we're returning, the style guide suggests its name be in `UpperCamelCase`. so this should be `var ToolboxClass`

core/options.js
alschmiedt: done

core/registry.js
alschmiedt: done

core/workspace_svg.js

PR #3923, opened 2020-05-28
Update build scripts

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Will bump the version number in package.json and blockly_compressed.js whenever we build core.

For releasing: 
1. npm run recompile 
2. Merge created PR into develop
3. npm run release
4. Merge created PR into master

### Proposed Changes
1. Remove the part that creates a pull request that bumps the version since we are rolling this into the rebuild process.
2. Added the bump whenever we build core
3. Renamed prerelease -> recompile

### Test Coverage

Tested on:

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,36 +10,12 @@
  */
 
 var gulp = require('gulp');
-var execSync = require('child_process').execSync;
-
-var packageJson = require('./package.json');
 
 var typings = require('./scripts/gulpfiles/typings');
 var buildTasks = require('./scripts/gulpfiles/build_tasks');
 var packageTasks = require('./scripts/gulpfiles/package_tasks');
 var gitTasks = require('./scripts/gulpfiles/git_tasks');
 var licenseTasks = require('./scripts/gulpfiles/license_tasks');
-
-// See https://docs.npmjs.com/cli/version.
-const preversion = gulp.series(
-  gitTasks.syncMaster,
-  function(done) {
-    // Create a branch named bump_version for the bump and rebuild.
-    execSync('git checkout -b bump_version', { stdio: 'inherit' });
-    done();
-  },
-);
-
-// See https://docs.npmjs.com/cli/version
-function postversion(done) {
-  // Push both the branch and tag to google/blockly.
-  execSync('git push ' + upstream_url + ' bump_version',
-      { stdio: 'inherit' });
-  var tagName = 'v' + packageJson.version;
-  execSync('git push ' + upstream_url + ' ' + tagName,
-      { stdio: 'inherit' });
-  done();
-};
 
 module.exports = {
   default: buildTasks.build,
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -26,8 +26,6 @@ module.exports = {
   buildUncompressed: buildTasks.uncompressed,
   buildCompressed: buildTasks.compressed,
   buildGenerators: buildTasks.generators,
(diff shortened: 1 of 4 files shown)


scripts/gulpfiles/git_tasks.js
```diff
@@ -88,7 +93,7 @@ const recompile = gulp.series(
 // Create and push an RC branch.
 // Note that this pushes to google/blockly.
 const createRC = gulp.series(
-  syncDevelop,
+  syncDevelop(),
```
rachel-fenichel, line 96: ?

This is supposed to be the function, not the call to it. gulp.series should call it.

gulpfile.js
```diff
@@ -10,37 +10,13 @@
...
-var packageJson = require('./package.json');
 
 var typings = require('./scripts/gulpfiles/typings');
 var buildTasks = require('./scripts/gulpfiles/build_tasks');
 var packageTasks = require('./scripts/gulpfiles/package_tasks');
 var gitTasks = require('./scripts/gulpfiles/git_tasks');
 var licenseTasks = require('./scripts/gulpfiles/license_tasks');
 
-// See https://docs.npmjs.com/cli/version.
-const preversion = gulp.series(
```
rachel-fenichel, line 24: You removed these two (and their entries in package.json) because now we can use the builtin versions, right?

package.json
```diff
@@ -26,23 +26,20 @@
...
     "build:uncompressed": "gulp buildUncompressed",
-    "bump": "npm version 3.$(date +'%Y%m%d').0",
+    "bump": "npm --no-git-tag-version version 3.$(date +'%Y%m%d').0",
     "license": "gulp checkLicenses",
     "lint": "eslint .",
     "package": "gulp package",
-    "postversion": "gulp postversion",
     "prepare": "npm run package",
-    "preversion": "gulp preversion",
-    "prerelease": "gulp gitRecompile",
```
rachel-fenichel, line 36: Are there any references to postversion, preversion, version, or prelease that you need to update?

scripts/gulpfiles/git_tasks.js
alschmiedt: I was getting an error because done wasn't getting called. I think the syncDevelop function is returning the function that needs to be called.

gulpfile.js
alschmiedt: I did it because these are creating a bump_version branch that I don't think we will need anymore since we are combining bumping the version with our recompile script.

package.json
alschmiedt: I don't think so. These are automagically called when someone runs npm run version

rachel-fenichel: Even prerelease?

scripts/gulpfiles/git_tasks.js
rachel-fenichel: Hm, I guess it just wasn't working before and no one noticed? Does that mean this wasn't ever being used?

alschmiedt: I used it in the January release where it was working because it looked like this: `gulp.task('git-sync-develop', syncBranch('develop'));` And then I think it got refactored and I didn't use it in the last release because we changed how we were releasing.

package.json
alschmiedt: I think so. Running npm prerelease was our first step in our old release instructions, so I don't think it got used internally.

PR #3924, opened 2020-05-28
Enable unusedPrivateMembers check

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves
Enables unusedPrivateMembers check

### Proposed Changes

Remove unused private members
Fix up the use of private on members the override a property.
Make Blockly.Icon abstract

### Test Coverage

Tested on:

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -410,18 +410,6 @@ Blockly.Workspace.prototype.deleteVariableById = function(id) {
 };
 
 /**
- * Deletes a variable and all of its uses from this workspace without asking
- * the user for confirmation.
- * @param {!Blockly.VariableModel} variable Variable to delete.
- * @param {!Array.<!Blockly.Block>} uses An array of uses of the variable.
- * @private
- * @suppress {unusedPrivateMembers}
- */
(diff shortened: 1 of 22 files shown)


samelhusseini:
@googlebot I consent

core/workspace.js
```diff
@@ -415,6 +415,7 @@ Blockly.Workspace.prototype.deleteVariableById = function(id) {
  * @param {!Blockly.VariableModel} variable Variable to delete.
  * @param {!Array.<!Blockly.Block>} uses An array of uses of the variable.
  * @private
+ * @suppress {unusedPrivateMembers}
  */
 Blockly.Workspace.prototype.deleteVariableInternal_ = function(variable, uses) {
```
rachel-fenichel, line 420: Why did you keep this function around?

samelhusseini: It wasn't clear to me whether it was one of those APIs that is set as private but is actually used, and if we wanted to deprecate before removing.
Based on your comment, it sounds like you think this can be removed.

rachel-fenichel: I just took a look at the history and we're good to remove it :+1:

samelhusseini: Done.

PR #3925, opened 2020-05-29
Update gh pages script

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Adds step for rebuilding before pushing to github pages.
Remove automatically pushing to github pages after a release. 
I am concerned if there is a merge conflict in the update github pages step it will be confusing what part of the script has failed if I put it with the release step. So I would rather keep the tasks separate.

### Test Coverage

Tested on:


alschmiedt:
I think so. It still seems beneficial to have a PR showing when we merge everything from develop into master.

PR #3927, opened 2020-06-01
Add support for adding objects to the registry

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Able to register and add a category.

### Proposed Changes
Changes the registry to handle storing object as well as constructor functions. 

### Reason for Changes
We want to be able to register a category so that users can use it in xml.

### Test Coverage

Tested on:


core/registry.js
```diff
@@ -16,12 +16,13 @@ goog.provide('Blockly.registry');
...
 goog.requireType('Blockly.Field');
 goog.requireType('Blockly.IToolbox');
+goog.requireType('Blockly.utils.toolbox');
 
 
 /**
  * A map of maps. With the keys being the type and name of the class we are
  * registering and the value being the constructor function.
- * Ex: {'field': {'field_angle': Blockly.FieldAngle}}
+ * ex: {'field': {'field_angle': Blockly.FieldAngle}}
```
samelhusseini, line 25: I think what you're after here is `e.g.`
https://english.stackexchange.com/questions/16197/whats-the-difference-between-e-g-and-ex

core/registry.js
```diff
@@ -62,17 +63,21 @@ Blockly.registry.Type.FIELD = new Blockly.registry.Type('field');
 /** @type {!Blockly.registry.Type<Blockly.IToolbox>} */
 Blockly.registry.Type.TOOLBOX = new Blockly.registry.Type('toolbox');
 
+/** @type {!Blockly.registry.Type<Blockly.utils.toolbox.Category>} */
+Blockly.registry.Type.TOOLBOX_CATEGORY = new Blockly.registry.Type('category');
```
samelhusseini, line 67: Since this is a generic registry. I feel category is not specific enough. How about: 
`toolboxCategory` or `toolbox.category`?

tests/playground.html
```diff
@@ -617,15 +618,7 @@ <h1>Blockly Playground</h1>
...
-    <category name="Logic" categorystyle="logic_category">
-      <block type="controls_if"></block>
-      <block type="logic_compare"></block>
-      <block type="logic_operation"></block>
-      <block type="logic_negate"></block>
-      <block type="logic_boolean"></block>
-      <block type="logic_null" disabled="true"></block>
-      <block type="logic_ternary"></block>
-    </category>
+    <category type="logic_category"></category>
```
samelhusseini, line 621: I worry, since this is the "main" playground toolbox, that folks would copy it without including the `playgrounds/toolbox_categories.js` file. Is there some other category we could use as an example?
If not, perhaps we could only use the registered logic category for toolbox-categories-typed-variables?

tests/playgrounds/toolbox_categories.js
```diff
@@ -0,0 +1,37 @@
...
+      "blockxml": "<block type=\"logic_null\" disabled=\"true\"></block>",
+    },
+    {
+      "kind": "BLOCK",
+      "blockxml": "<block type=\"logic_ternary\"></block>",
+    }
+  ],
+};
+
+Blockly.registry.register(Blockly.registry.Type.TOOLBOX_CATEGORY, 'logic_category', logic_category);
```
samelhusseini, line 37: I feel like this might be a little confusing. We're telling developers to register blocks (used in the toolbox xml) with: 
``Blockly.Blocks[TYPE] = ...`` whereas they register categories in the same toolbox xml using: 
``Blockly.registry.register(Blockly.registry.Type.TOOLBOX_CATEGORY, TYPE, ...)``.

I feel like a more appropriate pattern would be to use ``Blockly.Categories[TYPE] = ...``, I recognize that goes against the idea of creating a registry, but I feel in this case it would be more confusing to switch the pattern given the same toolbox XML attribute `type` would end up referring to different patterns depending on the xml tag (block / category).

core/registry.js
```diff
@@ -167,8 +174,33 @@ Blockly.registry.getClass = function(type, name) {
...
+
+/**
+ * Gets the object for the given name and type.
+ * @param {string|Blockly.registry.Type<T>} type The type of the plugin.
+ *     (ex: Category)
+ * @param {string} name The plugin's name. (Ex. logic_category)
+ * @returns {T} The object with the given name and type or null if none exists.
+ * @template T
+ */
+Blockly.registry.getObject = function(type, name) {
```
samelhusseini, line 197: Just wanting to get an idea of what other objects we have in mind here?

Ones I can think of: 
- Blocks
- Block generators
- Generators
- Themes

I can see a solid use case for Themes, and maybe Generators, but unclear whether we would want to change the pattern for registering blocks / block generators (functions).

core/registry.js
```diff
@@ -143,15 +149,16 @@ Blockly.registry.unregister = function(type, name) {
...
+ *     (ex: Field, Renderer)
  * @param {string} name The plugin's name. (Ex. field_angle, geras)
- * @return {?function(new:T, ...?)} The class with the given name and type or
- *     null if none exists.
+ * @return {?function(new:T, ...?)|Object} The class or object with the given
+ *     name and type or null if none exists.
  * @template T
  */
-Blockly.registry.getClass = function(type, name) {
+Blockly.registry.getRegistryItem_ = function(type, name) {
```
samelhusseini, line 161: `get_` or `getItem_`? registry is already in the namespace.

tests/playground.html
alschmiedt: I moved it to the typed-variables toolbox and added a comment above it. If you still think it is confusing lmk and I can try to come up with a better solution.

tests/playgrounds/toolbox_categories.js
alschmiedt: Talked about offline. Removing this for now, and will see if it needs to be added later on.

PR #3928, opened 2020-06-03
Add ability to add blocks by type

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Before if a user wanted to define a simple block in JSON they would need to specify the blockxml, which can be a bit of a pain. This will allow users to just give the type. 

### Proposed Changes
Add ability to have a block with the following definition.
{
'kind':'block',
'type': 'type_name',
'disabled': 'true'
}

### Reason for Changes
Makes it easier to add a simple block. 

### Test Coverage

Tested on:

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -612,13 +612,15 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
  * @private
  */
 Blockly.Flyout.prototype.getBlockXml_ = function(blockInfo) {
-  if (blockInfo['type']) {
+  var blockXml = null;
+  // All blockInfo will have type, so check for blockxml first.
+  if (blockInfo['blockxml']) {
+    blockXml = Blockly.Xml.textToDom(blockInfo['blockxml']);
+  } else if (blockInfo['type']) {
     var blockXml = Blockly.utils.xml.createElement('xml');
     blockXml.setAttribute('type', blockInfo['type']);
     blockXml.setAttribute('gap', blockInfo['gap']);
     blockXml.setAttribute('disabled', blockInfo['disabled']);
-  } else if (blockInfo['blockxml']) {
-    blockXml = Blockly.Xml.textToDom(blockInfo['blockxml']);
   } else {
     throw Error('Error: Invalid block definition. Block definition must have blockxml or type.');
   }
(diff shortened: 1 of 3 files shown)


core/flyout_base.js
```diff
@@ -612,11 +612,17 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
  * @private
  */
 Blockly.Flyout.prototype.getBlockXml_ = function(blockInfo) {
-  var blockXml = blockInfo['blockxml'];
-  if (blockXml) {
+  var blockXml = null;
+  // All blockInfo will have type, so check for blockxml first.
+  if (blockInfo['blockxml']) {
     blockXml = Blockly.Xml.textToDom(blockInfo['blockxml']);
+  } else if (blockInfo['type']) {
```
samelhusseini, line 619: We should update `Blockly.utils.toolbox.Block` to include type, and make blockxml optional.

core/flyout_base.js
```diff
@@ -612,11 +612,17 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
...
+    var blockXml = Blockly.utils.xml.createElement('xml');
+    blockXml.setAttribute('type', blockInfo['type']);
+    blockXml.setAttribute('gap', blockInfo['gap']);
```
samelhusseini, line 622: `gap` and `disabled` are also not on the blockInfo type ``Blockly.utils.toolbox.Block``.

tests/playgrounds/test_blocks_toolbox.js
```diff
@@ -889,35 +889,35 @@ var style = {
...
     },
     {
       "kind": "BLOCK",
-      "blockxml": "<block type=\"test_style_hex4\"></block>",
+      "type": "test_style_hex4",
     },
     {
       "kind": "BLOCK",
-      "blockxml": "<block type=\"test_style_hex5\"></block>",
+      "type": "test_style_hex5",
```
rachel-fenichel, line 920: Based on the code, you can have blocks with both `type` and `gap`.  Add something to test that.

core/flyout_base.js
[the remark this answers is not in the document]
alschmiedt: This was an attempt to simplify the code so that when we get the blockxml object back everything has converged to single object, instead of still having information in two spots. I can update if you think it would be better to have `blockInfo['gap'] || blockXml.getAttribute('gap')`

PR #3931, opened 2020-06-04
Fix performance

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
1. Remove call to  Blockly.utils.xml.domToText. I was converting to Text in `convertToolboxXml` and then having to convert back from textToDom later on. 
2. Removed the two branches because I read that helps performance. Testing it, it seems to improve a it by ~6ms on average.

### Test Coverage

Tested on:

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -612,14 +612,11 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
  * @private
  */
 Blockly.Flyout.prototype.getBlockXml_ = function(blockInfo) {
-  var potBlockXml = blockInfo['blockxml'];
-  var blockXml = null;
+  var blockXml = blockInfo['blockxml'];
 
   // All blockInfo will have type, so check for blockxml first.
-  if (potBlockXml && typeof potBlockXml == 'string') {
-    blockXml = Blockly.Xml.textToDom(potBlockXml);
-  } else if (potBlockXml) {
-    blockXml = potBlockXml;
+  if (blockXml && typeof blockXml == 'string') {
+    blockXml = Blockly.Xml.textToDom(blockXml);
   } else if (blockInfo['type']) {
     blockXml = Blockly.utils.xml.createElement('xml');
     blockXml.setAttribute('type', blockInfo['type']);
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -612,19 +612,23 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
  * @private
  */
 Blockly.Flyout.prototype.getBlockXml_ = function(blockInfo) {
-  var blockXml = /** @type {!Element} */ (blockInfo['blockxml']);
+  var blockElement = null;
+  var blockXml = blockInfo['blockxml'];
 
-  // All blockInfo will have type, so check for blockxml first.
-  if (blockXml && typeof blockXml == 'string') {
-    blockXml = Blockly.Xml.textToDom(blockXml);
-  } else if (blockInfo['type'] && !blockXml) {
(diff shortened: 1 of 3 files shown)


core/utils/toolbox.js
```diff
@@ -126,15 +126,16 @@ Blockly.utils.toolbox.toolboxXmlToJson_ = function(toolboxDef) {
...
+
     // Store the xml for a block
-    if (child.tagName.toUpperCase() == 'BLOCK') {
-      obj['blockxml'] = Blockly.utils.xml.domToText(child);
-    }
-    // Get the contents for a category.
-    if (child.tagName.toUpperCase() == 'CATEGORY') {
-      obj['contents'] = Blockly.utils.toolbox.toolboxXmlToJson_(child);
-    }
+    (tagName == 'BLOCK' && (obj['blockxml'] = child));
```
samelhusseini, line 133: Please use an if statement.

core/flyout_base.js
```diff
@@ -612,10 +612,11 @@ Blockly.Flyout.prototype.createBlock_ = function(blockXml) {
  * @private
  */
 Blockly.Flyout.prototype.getBlockXml_ = function(blockInfo) {
-  var blockXml = null;
+  var blockXml = blockInfo['blockxml'];
```
samelhusseini, line 615: Can you cast this to a Node for clarity of type.

core/utils/toolbox.js
alschmiedt: The reason I am doing it this way is to increase performance. It seems to shave off ~5ms for a toolbox with 20 categories and 100 blocks each.

core/flyout_base.js
alschmiedt: At this point it could be either an element or a string, I can cast it when returning?

core/utils/toolbox.js
samelhusseini: That's very surprising. Do you know why it increases performance?

core/flyout_base.js
samelhusseini: Yeah, the desired type of blockXml is to be of type `Element` here so it's okay to cast at the top. You'd be saying, assume the type is Element.
You then check if the type is string below, and if so, convert it into an Element. 

The goal from a readability perspective, is to determine what the type of blockXml is, and that's not clear until further reading, so although the type cast isn't necessary, better to add if it helps readability

core/utils/toolbox.js
samelhusseini: I can see that you've cached the call to tagName, we can still do that, but the comparison is between running: 
```
if (condition) {
  statement
}
```
and 
```
(condition) && (statement)
```
Increasing performance doesn't make sense to me.

core/flyout_base.js
alschmiedt: This makes the compiler angry, since I am passing something that is an element to textToDom. I updated to try and make it more clear, but lmk if it doesn't and I can try again.

core/utils/toolbox.js
```diff
@@ -126,15 +126,18 @@ Blockly.utils.toolbox.toolboxXmlToJson_ = function(toolboxDef) {
...
+
     // Store the xml for a block
-    if (child.tagName.toUpperCase() == 'BLOCK') {
-      obj['blockxml'] = Blockly.utils.xml.domToText(child);
+    if (tagName == 'BLOCK') {
+      obj['blockxml'] = child;
     }
     // Get the contents for a category.
-    if (child.tagName.toUpperCase() == 'CATEGORY') {
+    if (tagName == 'CATEGORY') {
```
samelhusseini, line 137: else if?

PR #3932, opened 2020-06-05
Register themes

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Register themes so that they can be referenced by name in options, eg: 
```js
{
   theme: 'dark'
}
```
or can be referenced by name in the base theme: 
```js
{
    theme: {
         base: 'dark',
         fontStyle: {
               size: 30,
         }
    }
}
```

This also allows us to have a central repository of themes by name for the use of the playground. 

Because themes already are constructed by name, we can register them internally in the code, and developers don't have to worry about registering themes themselves.

### Reason for Changes

Playground listing themes.
Referencing themes by name.

### Test Coverage

Tested on:


core/utils/idgenerator.js
```diff
@@ -29,5 +29,5 @@ Blockly.utils.IdGenerator.nextId_ = 0;
  * @return {string} The next unique identifier.
  */
 Blockly.utils.IdGenerator.getNextUniqueId = function() {
-  return 'blockly:' + (Blockly.utils.IdGenerator.nextId_++).toString(36);
+  return 'blockly-' + (Blockly.utils.IdGenerator.nextId_++).toString(36);
```
alschmiedt, line 32: What is the reasoning behind this change?

samelhusseini: The comment above says: "Use only ASCII letters, digits, '_', '-' and '.'".
In addition, css class names can not have `:` in the name.

alschmiedt: Ok, I think we should add the api change label for this then. From my understanding since we don't have a way to add classes to the toolbox categories, people sometimes rely on some hack around using the category ids. It's not really supported but I think would still be a nice heads up.

samelhusseini: Done.

Issue #3905, opened 2020-05-18
Convert workspace.resizesEnabled_ into a counter


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**

Right now it's a boolean, which limits its usefulness since it is set and unset in multiple places.

**Describe the solution you'd like**

Change it from a boolean to a counter, following the model we use for events.

adroitwhiz:
I got started on this, and in the process uncovered two places in the codebase where resizes are disabled and never re-enabled, both in `flyout_base.js`: [in `createBlock`](https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/core/flyout_base.js#L641) and [in `placeNewBlock_`](https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/core/flyout_base.js#L801) (a search for `terminateDrag` reveals nothing other than that one comment).

rachel-fenichel:
`createBlock` is called in [Blockly.Gesture.updateIsDraggingFromFlyout_](https://github.com/google/blockly/blob/develop/core/gesture.js#L326).  That's called in `updateIsDraggingBlock_`, and if it returned true, `startDraggingBlock_` is called.

`terminateDrag` is probably referring to [Blockly.BlockDragger.prototype.endBlockDrag](https://github.com/google/blockly/blob/develop/core/block_dragger.js#L246), which does re-enable resizes.

Having the disable in both `createBlock` and `placeNewBlock_` does seem wrong, since `placeNewBlock_` is only called from `createBlock`.

`createBlock is also called in one other place, [Blockly.navigation.insertFromFlyout](https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/core/keyboard_nav/navigation.js#L219), and that one is definitely not balanced correctly.

With that extra information, how do you suggest balancing enable and disable calls here?

adroitwhiz:
I'm iffy on the idea of tracking a resizes-enabled counter across functions-- as far as I can tell, all `Blockly.Events.disable()` and `Blockly.Events.enable()` pairs (the existing usages of a counter-based system) occur within the same function.

Already, many calls to disable/enable events are [wrapped in try-finally blocks](https://github.com/google/blockly/pull/348) to prevent events from being permanently disabled just in case any exceptions are thrown.

[...]

PR #3933, opened 2020-06-05
Convert workspace.resizesEnabled_ into a counter

adroitwhiz:
## The basics

- [x] I branched from develop

## The details

This PR converts `WorkspaceSvg.setResizesEnabled` to use a counter internally to check whether resizes are enabled, in the same way events currently do.

### Resolves

Resolves https://github.com/google/blockly/issues/3905

### Proposed Changes

There are four parts here:
- Make `setResizesEnabled` use a counter instead of a simple boolean.
- Modify methods in `flyout_base.js` to properly reenable resizes after disabling them.
- Wrap all code before `setResizesEnabled(true)` in `try` blocks, and place `setResizesEnabled(true)` inside `finally` clauses
- Disable then reenable resizes inside `Blockly.Gesture.updateIsDraggingBlock_`

### Reason for Changes

- Make `setResizesEnabled` use a counter instead of a simple boolean.
  - Currently, many call sites disable and then reenable workspace resizing. This limits its usefulness and can lead to bugs-- for instance, if an outer function disables resizing, calls an inner function, then does some other stuff and reenables resizing, but the inner function also disables and reenables resizing, then everything that happens after the inner function is called could break if it expects resizing to be disabled.
  - The new behavior is more "stack-like"-- if workspace resizing is disabled twice, for instance, it must also be reenabled twice.
- Modify methods in `flyout_base.js` to properly reenable resizes after disabling them.
  - Previously, both `Flyout.createBlock` and `Flyout.placeNewBlock_` disabled resizes and then never reenabled them, expecting it to be done by `endBlockDrag`.
  - If you placed a block with keyboard navigation, or clicked a block on an autoclosing flyout to place it without dragging it, this would never happen. This is worse now because if you forget to match every `setResizesEnabled(false)` with a `setResizesEnabled(true)`, resizes will *never* be reenabled.
  - Now, `Flyout.createBlock` doesn't disable resizes, and `Flyout.placeNewBlock_` reenables resizes itself.
- Wrap all code before `setResizesEnabled(true)` in `try` blocks, and place `setResizesEnabled(true)` inside `finally` clauses
  - This ensures that even if code throws, resizes will always be reenabled.
- Disable then reenable resizes inside `Blockly.Gesture.updateIsDraggingBlock_`
  - This is the one I'm the least happy about. Since `Flyout.createBlock` now reenables resizes, we need a way to make sure that the workspace doesn't get resized between `createBlock` and `startBlockDrag`. This new behavior of disabling workspace resizes before the block is created and reenabling them after `startBlockDrag` is called does the job, but it relies on the new counter behavior: it knows that after `startBlockDrag` is called, `setResizesEnabled(true)` won't *actually* resize the workspace since `startBlockDrag` also disables resizes--it only "balances out" the `setResizesEnabled(false)` from earlier.
  - If you have an idea for how to do this more cleanly, I'd be glad to do that instead.

### Test Coverage

Tested on:

* Desktop Firefox

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -383,10 +383,13 @@ Blockly.Xml.textToDom = function(text) {
  */
 Blockly.Xml.clearWorkspaceAndLoadFromXml = function(xml, workspace) {
   workspace.setResizesEnabled(false);
-  workspace.clear();
-  var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
-  workspace.setResizesEnabled(true);
-  return blockIds;
+  try {
+    workspace.clear();
+    var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
+    return blockIds;
+  } finally {
+    workspace.setResizesEnabled(true);
+  }
 };
 
 /**
@@ -478,11 +481,12 @@ Blockly.Xml.domToWorkspace = function(xml, workspace) {
       Blockly.Events.setGroup(false);
     }
     Blockly.utils.dom.stopTextWidthCache();
+    // Re-enable workspace resizing.
+    if (workspace.setResizesEnabled) {
+      workspace.setResizesEnabled(true);
+    }
   }
-  // Re-enable workspace resizing.
-  if (workspace.setResizesEnabled) {
-    workspace.setResizesEnabled(true);
-  }
+
   Blockly.Events.fire(new Blockly.Events.FinishedLoading(workspace));
   return newBlockIds;
 };
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -217,34 +217,37 @@ Blockly.BlockDragger.prototype.dragBlock = function(e, currentDragDeltaXY) {
  * @package
  */
 Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
-  // Make sure internal state is fresh.
-  this.dragBlock(e, currentDragDeltaXY);
-  this.dragIconData_ = [];
-  this.fireDragEndEvent_();
+  try {
+    // Make sure internal state is fresh.
+    this.dragBlock(e, currentDragDeltaXY);
+    this.dragIconData_ = [];
+    this.fireDragEndEvent_();
 
-  Blockly.utils.dom.stopTextWidthCache();
+    Blockly.utils.dom.stopTextWidthCache();
 
-  Blockly.blockAnimations.disconnectUiStop();
+    Blockly.blockAnimations.disconnectUiStop();
 
-  var delta = this.pixelsToWorkspaceUnits_(currentDragDeltaXY);
-  var newLoc = Blockly.utils.Coordinate.sum(this.startXY_, delta);
-  this.draggingBlock_.moveOffDragSurface(newLoc);
+    var delta = this.pixelsToWorkspaceUnits_(currentDragDeltaXY);
+    var newLoc = Blockly.utils.Coordinate.sum(this.startXY_, delta);
+    this.draggingBlock_.moveOffDragSurface(newLoc);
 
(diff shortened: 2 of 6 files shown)


core/xml.js
```diff
@@ -383,9 +383,13 @@ Blockly.Xml.textToDom = function(text) {
  */
 Blockly.Xml.clearWorkspaceAndLoadFromXml = function(xml, workspace) {
   workspace.setResizesEnabled(false);
-  workspace.clear();
-  var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
-  workspace.setResizesEnabled(true);
+  try {
+    workspace.clear();
+    var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
```
rachel-fenichel, line 388: Travis is complaining because this change the return type:

https://travis-ci.org/github/google/blockly/jobs/694866312#L9283

rachel-fenichel:
Disable then reenable resizes inside Blockly.Gesture.updateIsDraggingBlock_
> This is the one I'm the least happy about. Since Flyout.createBlock now reenables resizes, we need a way to make sure that the workspace doesn't get resized between createBlock and startBlockDrag. This new behavior of disabling workspace resizes before the block is created and reenabling them after startBlockDrag is called does the job, but it relies on the new counter behavior: it knows that after startBlockDrag is called, setResizesEnabled(true) won't actually resize the workspace since startBlockDrag also disables resizes--it only "balances out" the setResizesEnabled(false) from earlier.
> If you have an idea for how to do this more cleanly, I'd be glad to do that instead.

I think this isn't inherently problematic. Even if the outer set of disable/enable calls were to disappear, we would avoid one unnecessary resize from the createBlock.

Another option is just to disable resizes earlier in the gesture code, and reenable them later. Are there any paths through a gesture where resizing actually should be enabled? If not, what about disabling at start and reenabling at end?

adroitwhiz:
> Even if the outer set of disable/enable calls were to disappear, we would avoid one unnecessary resize from the createBlock.

> Another option is just to disable resizes earlier in the gesture code, and reenable them later. Are there any paths through a gesture where resizing actually should be enabled? If not, what about disabling at start and reenabling at end?

The issue is that the block is created (and hence resizes reenabled) before the gesture begins, so if resizes aren't disabled, the workspace will resize once when a block is dragged out of the flyout, causing the scrollbars to jump a bit:
Peek 2020-06-12 04-13

It looks like there's a `handleUp` function in `Gesture` which could be a place to reenable resizes, but no corresponding `handleDown` function, which means that ensuring that disabling/enabling resizes happens in pairs could be difficult. Is there a pair of matching functions in the gesture code that handles both the start and end of a drag?

rachel-fenichel:
Where do you see the block being created before the gesture begins? I know the block is created before the block dragger is created and used--is that what you're referring to?

Functions for starting gestures:
- [doStart](https://github.com/google/blockly/blob/master/core/gesture.js#L463) in gesture.js
- [handleFlyoutStart](https://github.com/google/blockly/blob/master/core/gesture.js#L665) in gesture.js
- [handleBlockStart](https://github.com/google/blockly/blob/master/core/gesture.js#L680) in gesture.js

If the gesture is in charge of resizes instead of the block dragger, you'll also need to reenable resizes in [cancel](https://github.com/google/blockly/blob/master/core/gesture.js#L597).

The way the various start functions work is that the mouse down handlers on fields, blocks, workspaces, etc call the corresponding handleSomethingStart function. So if you click on a field, you get a setStartField, handleBlockStart, handleWorkspaceStart. Mostly those just record which block, field, etc. was clicked--no action happens until it gets down to the workspace. When it gets down to the workspace, it calls `doStart`, and then registers event handlers for mouse moves and ups, and uses the recorded information to decide what the gesture was supposed to be.

Issue #3941, opened 2020-06-08
Add a function "workspace.refresh()"  


NavpreetDevpuri:
i was unable to refresh workspace after reinitialize a block by `Blockly.Blocks['a']`
then i found out that blocks inside workspace don't update but blocks inside toolbox are updating fine so, we need to refresh workspace after reinitializing a block.
As mentioned at https://github.com/google/blockly/issues/3940#issuecomment-640267246

1) A solution for refreshing workspace
    1) export xml of blocks 
    2) clear the workspace 
    3) reimport blocks from xml 
![HSgjRh9DEn](https://user-images.githubusercontent.com/30471072/83977682-0e2b7c80-a920-11ea-84f1-0e6166e440d4.gif)

2) A solution to refresh toolbox by using `workspace.refreshToolboxSelection()`
![YroRl89Pja](https://user-images.githubusercontent.com/30471072/83978042-5fd50680-a922-11ea-9e35-7d53052a607b.gif)

Now we can add a function `workspace.refresh()` something like 
```javascript
Blockly.Workspace.prototype.refresh() = function() {
    var xml = Blockly.Xml.workspaceToDom(this);
    this.clear();
    Blockly.Xml.domToWorkspace(xml, this);
    this.refreshToolboxSelection();
};
```

NavpreetDevpuri:
i don't know how to contribute means i don't know that putting this function into `blockly/core/workspace.js` is enough or i need to add it somewhere else also ?

genemars:
However, this can be achieved by triggering the window `resize` event:

```
window.dispatchEvent(new Event('resize'));
```
I have to use this as a workaround because Blockly is not resizing properly sometimes when used in Angular templates.

![Schermata del 2022-11-19 19-27-46](https://user-images.githubusercontent.com/6654265/202866060-f7a29d57-0917-46cc-aa3a-0e13ff869f44.png)

BeksOmega:
Hello @genemars ! This issue is slightly different from resizing the workspace. The person in this issue wanted all of the blocks on the workspace to update when they modified the block definition, which we do not generally want to support.

However, if you want to resize the workspace that is definitely supported. You can call `Blockly.svgResize(myWorkspace)`. I hope that helps!

genemars:
[...]

Issue #3940, opened 2020-06-07
Error while using workspace.removeBlockById()


NavpreetDevpuri:
i have a custom block with type `a`
```javascript 
Blockly.Blocks['a'] = {
        init: function() {
            this.appendValueInput("a1")
                .setCheck(null)
                .appendField(new Blockly.FieldDropdown([["aaa","0"], ["b","1"], ["c","2"]]), "a2");
            this.setColour(130);
            this.setTooltip("");
            this.setHelpUrl("");
        }
    };
    Blockly.JavaScript['a'] = function(block) {
        var dropdown_a2 = block.getFieldValue('a2');
        var value_a1 = Blockly.JavaScript.valueToCode(block, 'a1', Blockly.JavaScript.ORDER_ATOMIC);
        // TODO: Assemble JavaScript into code variable.
        var code = '...;\n';
        return code;
    };
```
i run `workspace.removeBlockById(workspace.getBlocksByType("a")[0].id)` in console and it show following error as shown in following gif 
![1tks3hbAD8](https://user-images.githubusercontent.com/30471072/83965004-37bbb800-a8ce-11ea-9a02-63d254079d7f.gif)

It should delete block

BeksOmega:
Hi! When filing an issue please fill out all of the pre-made sections. That helps people figure out what could be going on =)

Lucky for you someone else had the same problem a few weeks ago =) You want to use `block.dispose()` ([docs](https://developers.google.com/blockly/reference/js/Blockly.Block#dispose)) not `.removeBlockById`. As you can see [that function is package](https://developers.google.com/blockly/reference/js/Blockly.Workspace#removeBlockById).

I hope that helps! If you have any further problems (specifically related to disposing) I think it might be best to redirect your questions to the forums. Thank you :D

NavpreetDevpuri:
thanks for replying 
I want to achive something like "delete a block but don't want this deleted block goes into recyclebin"
thats why is used `.removeBlockById` 

mainly my main motive was to achive something like "update dropdown list of a block programmically" 
i was able to do that by simply reinitialize block by `Blockly.Blocks['a']` but this creates a problem that the blocks that are inside workspace those blocks just don't update in any way but those blocks that are inside toolbox are updated then i tried to delete those unchanged blocks and try to add new ones it works but then these deleted blocks just appears inside recyclebin.


now i just 
1) export xml of blocks 
2) clear the workspace 
3) reimport blocks from xml 
it works fine it update blocks.

[...]

PR #3945, opened 2020-06-09
added workspace.refresh()

NavpreetDevpuri:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/3941
https://github.com/google/blockly/issues/3940

### Proposed Changes

Now we can refresh workspace after changing some blocks 
VgbMcgh2y5

### Reason for Changes

There should be a way to refresh workspace after reinitialized a block by `Blockly.Blocks['block_name']`

### Test Coverage
Tested by using playground

Tested on:
Desktop Chrome
Desktop Firefox
Desktop Opera

### Documentation

Added new function into `Workspace` class
#### refresh
refresh() 
Refresh workspace after changing some blolcks.

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -132,6 +132,15 @@ Blockly.Workspace.prototype.MAX_UNDO = 1024;
 Blockly.Workspace.prototype.connectionDBList = null;
 
 /**
+ * Refresh workspace after changing some blolcks.
+ */
+Blockly.Workspace.prototype.refresh = function() {
+  this.clear();
+  Blockly.Xml.domToWorkspace(Blockly.Xml.workspaceToDom(this), this);
+  this.refreshToolboxSelection();
+};
+
+/**
  * Dispose of this workspace.
  * Unlink from all DOM elements to prevent memory leaks.
  * @suppress {checkTypes}


core/workspace.js
```diff
@@ -131,6 +131,15 @@ Blockly.Workspace.prototype.MAX_UNDO = 1024;
  */
 Blockly.Workspace.prototype.connectionDBList = null;
 
+/**
+ * Refresh workspace after changing some blolcks.
+ */
+Blockly.Workspace.prototype.refresh = function() {
+  this.clear();
+  Blockly.Xml.domToWorkspace(Blockly.Xml.workspaceToDom(this), this);
```
samelhusseini, line 139: `workspaceToDom` after the workspace is cleared won't show anything. I believe what you're after is this: 
```
var xml = Blockly.Xml.workspaceToDom(this);
this.clear();
Blockly.Xml.domToWorkspace(xml, this);
```

Also, we need to handle the refresh of the trashcan if it is currently open.

@rachel-fenichel this seems reasonable to me, and is similar to what I'm doing in the new playground to refresh the workspace.
I wonder though, if there's a reason this functionality hadn't been added before, is there some piece of context we're missing here?

rachel-fenichel: We didn't have this function because clearing and reloading the workspace with the exact same blocks is almost never the correct thing to do (unless you're changing something core like whether there's a toolbox), and it's expensive.

If you need to rerender all of the blocks, you can get the list of blocks and call render on every one.

If you want to update a bunch of dropdowns, you can get the blocks by type and then update them.

If you want dynamic dropdowns, you can do that with the built-in dynamic dropdown functionality.

And so on.

In this [case](https://github.com/google/blockly/issues/3940), the goal appears to be to do dynamic dropdowns. Redefining the block is an extremely heavy-handed way to accomplish this, and the library provides several other ways to do it, including dynamic dropdowns and mutators.

samelhusseini: Thanks for the context @rachel-fenichel. 
I agree that, although this appears harmless, it encourages behaviour that yields to poor performance. 
Happy to add it when we get a case where there isn't a better alternative, but for now, with all the cases we know of listed above, there shouldn't be a reason for clearing all blocks and re-adding them.

NavpreetDevpuri: @rachel-fenichel i tried as you suggested before thinking about workspace.refresh() 
Here is an example when i get blocks by id and try to call render on it and nothing happened 
RTWcUxbflM

NavpreetDevpuri: By using `workspace.refresh()` we don't need to worry about anything like getting blocks by types and changing them and then try to rerender those blocks in toolbox, workspace's area and trashcan also.

samelhusseini: Have you tried using a function to populate the list of options in the `FieldDropdown`, also known as dynamic dropdowns: 
https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#dynamic_dropdowns

NavpreetDevpuri: actually i want full control of it i want to change full appearance of blocks. I want to update those blocks dynamically may be need to change color or may be want to change full functionality just need to redefine blocks.

rachel-fenichel: https://developers.google.com/blockly/guides/create-custom-blocks/web/mutators

Mutators give you even deeper control over your blocks.

If you want this functionality instead of using any of the other ways the library can help you out, you should implement it in your own app, since the functions you need to call are public.

NavpreetDevpuri: thanku
i will try it.

PR #3956, opened 2020-06-11
Emit workspace click event

### LINKED ISSUE #3947: Close Mutator on Workspace Click
Developers want to be able to close the mutator editor when a user clicks on the workspace. Previously people were trying to use the select and deselect events. For example: #3865

I think a hook in `onMouseDown_` for the workspace would solve the problem.

Making onMouseDown_ public so that it can be overridden. We could also do this, but it seems like people have already been overriding this and so the name change from onMouseDown_ -> onMouseDown would break them. 

**Additional context**

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3947

### Proposed Changes
Add an event that is emitted when someone clicks on the workspace.

### Test Coverage

Tested on:

### DIFF
--- a/core/gesture.js
+++ b/core/gesture.js
@@ -651,9 +651,23 @@ Blockly.Gesture.prototype.handleWsStart = function(e, ws) {
   this.setStartWorkspace_(ws);
   this.mostRecentEvent_ = e;
   this.doStart(e);
+  if (!this.startBlock_ && !this.startBubble_) {
+    this.fireWorkspaceClick_(ws);
+  }
   if (this.startWorkspace_.keyboardAccessibilityMode) {
     Blockly.navigation.setState(Blockly.navigation.STATE_WS);
   }
+};
+
+/**
+ * Fires a workspace click event.
+ * @param {!Blockly.WorkspaceSvg} ws The workspace that a user clicks on.
+ * @private
+ */
+Blockly.Gesture.prototype.fireWorkspaceClick_ = function(ws) {
+  var clickEvent = new Blockly.Events.Ui(null, "workspaceClick", null, null);
+  clickEvent.workspaceId = ws.id;
+  Blockly.Events.fire(clickEvent);
 };
 
 /**


core/gesture.js
```diff
@@ -651,11 +651,25 @@ Blockly.Gesture.prototype.handleWsStart = function(e, ws) {
...
   }
 };
 
+/**
+ * Fires a workspace click event.
+ * @param {!Blockly.WorkspaceSvg} ws The workspace that a user clicks on.
+ * @private
+ */
+Blockly.Gesture.prototype.fireWorkspaceClick_ = function(ws) {
+  var clickEvent = new Blockly.Events.Ui(null, "workspaceClick", null, null);
```
samelhusseini, line 668: nit: Use `'` instead of `"` https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes

core/gesture.js
```diff
@@ -651,11 +651,25 @@ Blockly.Gesture.prototype.handleWsStart = function(e, ws) {
   this.setStartWorkspace_(ws);
   this.mostRecentEvent_ = e;
   this.doStart(e);
+  if (!this.startBlock_ && !this.startBubble_) {
+    this.fireWorkspaceClick_(ws);
```
samelhusseini, line 655: Curious why we don't do this in `doWorkspaceClick_`?

core/gesture.js
alschmiedt: Updated in #3959. Just curious, should the eslint catch this?

core/gesture.js
alschmiedt: woops! Just didn't know it was there.

core/gesture.js
samelhusseini: Yeah it totally could, although we use `"` in a lot of our test files, so might require a little refactoring at first.
https://eslint.org/docs/rules/quotes
From the goog-eslint-config: 
```js
    'quotes': ['error', 'single', {allowTemplateLiterals: true}],
```

Issue #3966, opened 2020-06-12
Input.removeField(name) throws unsubstituted Error message


mattbishop:
**Describe the bug**

When attempting to remove a field from an input that does not exist, Input will throw the following Error: `Field "%s" not found.`

**To Reproduce**

const anInput = aBlock.appendDummyInput()
anInput.removeField('not_a_field')

**Expected behavior**

An Error stating `Field "not_a_field" not found.`

**Desktop (please complete the following information):**
 - OS: MacOS 15.5.5
 - Browser Safari and Chrome
 - Version Safari 13.1.1, Chrome 83

**Stack Traces**

```
Uncaught Error: Field "%s" not found.
    at Blockly.Input.removeField (input.js:148)
    at Blockly.BlockSvg.updateShape (boolean-blocks.js:93)
    at Blockly.BlockSvg.compose (boolean-blocks.js:87)
    at Blockly.Mutator.workspaceChanged_ (mutator.js:377)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (workspace.js:657)
    at Blockly.Events.fireNow_ (events.js:192)
```

PR #3968, opened 2020-06-12
Add opt quiet to remove field

### LINKED ISSUE #3967: A quietable version of Input.removeField()

I am creating a mutator that will add and remove fields in a named input of a block. When the state of the mutator dictates the the fields need to be removed from the block, the code has to call theInput.removeField(name). 

This remove() method throws an Error if the field name is not found, so I have to either wrap the call in try/catch or first check for the field before removing it.

Block has a similar removeInput(name) method, with an `opt_quiet` property that does not throw errors. I'd like something similar on Input, so I could call anInput.removeField(aName, true) and it would not throw an error if the field is not found.

My own utility method that would remove the fields in a try/catch or field test.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #3967 and #3966

### Proposed Changes
Add opt_quiet to removeField. Similar to how we have opt_quiet for removeInput.

### Test Coverage

Tested on:

### DIFF
--- a/core/input.js
+++ b/core/input.js
@@ -134,7 +134,7 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
 /**
  * Remove a field from this input.
  * @param {string} name The name of the field.
- * @param {boolean=} opt_quiet True to prevent error if field is not present.
+ * @param {boolean=} opt_quiet True to prevent an error if field is not present.
  * @return {boolean} True if operation succeeds, false if field is not present
  *     and opt_quiet is true.
  * @throws {Error} if the field is not present and opt_quiet is false.
--- a/core/input.js
+++ b/core/input.js
@@ -156,7 +156,7 @@ Blockly.Input.prototype.removeField = function(name, opt_quiet) {
   if (opt_quiet) {
     return false;
   } else {
-    throw Error('Field "%s" not found.', name);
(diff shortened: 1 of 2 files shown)


core/input.js
```diff
@@ -147,10 +150,14 @@ Blockly.Input.prototype.removeField = function(name) {
...
       }
-      return;
+      return true;
     }
   }
-  throw Error('Field "%s" not found.', name);
+  if (opt_quiet) {
+    return false;
+  } else {
+    throw Error('Field "%s" not found.', name);
```
mattbishop, line 159: BTW I have a bug report against this error. The string doesn't format like one would think: https://github.com/google/blockly/issues/3966

alschmiedt: Thanks for letting me know. Just updated.

core/input.js
```diff
@@ -134,9 +134,12 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
 /**
  * Remove a field from this input.
  * @param {string} name The name of the field.
- * @throws {Error} if the field is not present.
+ * @param {boolean=} opt_quiet True to prevent error if field is not present.
```
samelhusseini, line 137: nit: True to prevent /an/ error if field is not present

PR #3988, opened 2020-06-23
Fix theme and add example

### LINKED ISSUE #3987: Variable "propertyValue" of unknown origin
https://github.com/google/blockly/blob/52b818ed67c5bfa7fd73edf2b067649b8b1f447b/core/theme.js#L164-L171

`propertyValue `from where? Is it a mistake？

#3093

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#3987

### Proposed Changes
propertyValue -> style

### Reason for Changes
This allows users to use a value in the components in multiple places in their component. I added an example in dark.js in order to test it easier. I think we didn't realize it because it is a smaller feature that a lot of people might not use. Might be fine for this release since it is so small.

### Test Coverage

Tested on:

### DIFF
--- a/core/theme.js
+++ b/core/theme.js
@@ -167,7 +167,7 @@ Blockly.Theme.prototype.setCategoryStyle = function(categoryStyleName,
  */
 Blockly.Theme.prototype.getComponentStyle = function(componentName) {
   var style = this.componentStyles[componentName];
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
Approved, but hold until after release.

PR #3999, opened 2020-06-25
Fix #3998

### LINKED ISSUE #3998: Multiline text field generator broken on IE 11
Because it uses `.includes`.

Fix incoming

rachel-fenichel:
Avoid `.includes` in the JavaScript generator.

Not used in any other generators.


rachel-fenichel:
This should go into the june release.

Issue #4015, opened 2020-07-06
External images not working on dropdown input


FabioOliveira1:
**Describe the bug**

Images from external sources are not being displayed on dropdown options on production environment - neither on my project nor on blockly demos.
Instead of displaying it, a message object HTMLImageElement is shown.

**To Reproduce**

Click here: [demo](https://blockly-demo.appspot.com/static/demos/blockfactory/index.html#mw4oh2)

or

1. Go to https://blockly-demo.appspot.com/static/demos/blockfactory/index.html
2. Add a dummy input inside the main block
3. Add a dropdown field inside the dummy input
4. Change the dropdown items to show only one image - instead of 3 text field
5. Click on dropdown options on the preview
6. See the message [object HTMLImageElement] instead of showing the image

**Expected behavior**

It was supposed to show a image on the dropdown options

**Screenshots**

![image](https://user-images.githubusercontent.com/13422600/86609643-93876880-bf82-11ea-9280-5a85ed814419.png)

**Desktop:**
 - OS: Ubuntu 19.04
 - Browser: Chrome
 - Version: 83.0.4103.106 (Official Build) (64-bit)

**Additional context**

It works on local environment

samelhusseini:
A git bisect leads to this PR: https://github.com/google/blockly/pull/3880

samelhusseini:
Fixed and released patch 3.20200625.2

PR #4016, opened 2020-07-06
Fix field dropdown with images

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/4015

### Proposed Changes

Handle cases where the content of the menu item is a HTML element.

### Reason for Changes

Fix regression.

### Test Coverage

Tested with the test field dropdown image block.

Tested on:

### DIFF
--- a/core/menuitem.js
+++ b/core/menuitem.js
@@ -21,13 +21,13 @@ goog.require('Blockly.utils.IdGenerator');
  * Class representing an item in a menu.
  *
  * @param {string|!HTMLElement} content Text caption to display as the content
- *     of the item.
+ *     of the item, or a HTML element to display.
  * @param {string=} opt_value Data/model associated with the menu item.
  * @constructor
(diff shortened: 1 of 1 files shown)


core/menuitem.js
```diff
@@ -20,15 +20,15 @@ goog.require('Blockly.utils.IdGenerator');
 /**
  * Class representing an item in a menu.
  *
- * @param {string} content Text caption to display as the content of
- *     the item.
+ * @param {string|!HTMLElement} content Text caption to display as the content
```
alschmiedt, line 23: Update comment from "Text caption"

samelhusseini:
@rachel-fenichel I think this should head out as a patch release, please take a look.

rachel-fenichel:
@samelhusseini I agree, we should have a patch release with this.

PR #4024, opened 2020-07-08
Update click event handling for workspace and trashcan.

### LINKED ISSUE #3975: Trashcan Fires Workspace Click Event
**Describe the bug**

If you click on the trashcan to open it a workspace click event is fired. Not sure if this is the way it should work or not.

**To Reproduce**
1. Turn on event logging.
1. Delete a block.
3. Click on the trashcan to open it.
4. Observe the fired click event.

**Expected behavior**

Probably shouldn't fire an event? Although in the case of closing mutators maybe it should fire. I'm not sure.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#3975 

### Proposed Changes

- Changes ui events emitted on workspace and on block click to both be 'click' and adding string in newValue to differentiate between what has been clicked.
- Stops event propagation on mouse down when trashcan is clicked (and has contents) to prevent workspace click from being emitted.
- Emits ui event when trashcan is opened (mouseUp) 

### Reason for Changes

- Consistency in naming of events
- Clicking on trashcan is not expected to trigger a workspace click event

### Test Coverage

Playground

Tested on:
* Desktop Chrome

### Documentation

We may also want to update https://developers.google.com/blockly/guides/configure/web/events to include the new possible value for element (i.e. 'trashcan').

### Additional Information

Any support for emitting an event on trashcan flyout closing will be addressed in future PR.

### DIFF
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -301,6 +301,8 @@ Blockly.Trashcan.prototype.createDom = function() {
   this.svgLid_.setAttributeNS(Blockly.utils.dom.XLINK_NS, 'xlink:href',
       this.workspace_.options.pathToMedia + Blockly.SPRITE.url);
 
+  Blockly.bindEventWithChecks_(
+      this.svgGroup_, 'mousedown', this, this.blockMouseDownWhenFull_);
   Blockly.bindEventWithChecks_(this.svgGroup_, 'mouseup', this, this.click);
   // bindEventWithChecks_ quashes events too aggressively. See:
   // https://groups.google.com/forum/#!topic/blockly/QF4yB9Wx00s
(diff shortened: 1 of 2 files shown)


core/trashcan.js
```diff
@@ -510,14 +521,38 @@ Blockly.Trashcan.prototype.click = function() {
...
+
+  this.fireUiEvent_();
+};
+
+/**
+ * Fires a ui event for trashcan click.
+ * @private
+ */
+Blockly.Trashcan.prototype.fireUiEvent_ = function() {
+  var uiEvent = new Blockly.Events.Ui(null, 'trashcan', null, null);
```
moniika, line 533: I originally had "trashcanOpen" to follow the pattern of "commentOpen" and "mutatorOpen". However, using the naming "trashcan" is consistent with "category" ui event, which also opens flyout and I'm also not sure we want to emit an event every time a flyout is closed.

moniika: I've updated back to 'trashcanOpen', but I'm open to discussion.

rachel-fenichel:
LGTM. I think this deserves the 'api change' label.

PR #4025, opened 2020-07-08
Emit UI event on zoom control clicks.

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

Emits a ui event when zoom control is clicked including old and new workspace scale.

### Reason for Changes

Since clicking on zoom controls does not emit a workspace click event, it makes sense to emit some kind of event instead to describe what happened.

### Test Coverage

Playground

Tested on:
* Desktop Chrome

### Documentation

We may also want to update https://developers.google.com/blockly/guides/configure/web/events to include the new possible value for element (i.e. 'zoom').

### Additional Information

This PR also does some refactoring cleanup of the mouseDown event handlers for clicking on zoom controls.

### DIFF
--- a/core/zoom_controls.js
+++ b/core/zoom_controls.js
@@ -183,7 +183,6 @@ Blockly.ZoomControls.prototype.createZoomOutSvg_ = function(rnd) {
         clip-path="url(#blocklyZoomoutClipPath837493)"></image>
   </g>
   */
-  var ws = this.workspace_;
   this.zoomOutGroup_ = Blockly.utils.dom.createSvgElement('g',
       {'class': 'blocklyZoom'}, this.svgGroup_);
   var clip = Blockly.utils.dom.createSvgElement('clipPath',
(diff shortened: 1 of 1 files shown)


core/zoom_controls.js
```diff
@@ -207,16 +206,11 @@ Blockly.ZoomControls.prototype.createZoomOutSvg_ = function(rnd) {
...
 
   // Attach listener.
-  Blockly.bindEventWithChecks_(zoomoutSvg, 'mousedown', null, function(e) {
-    ws.markFocused();
-    ws.zoomCenter(-1);
-    Blockly.Touch.clearTouchIdentifier();  // Don't block future drags.
-    e.stopPropagation();  // Don't start a workspace scroll.
-    e.preventDefault();  // Stop double-clicking from selecting text.
-  });
+  Blockly.bindEventWithChecks_(
```
samelhusseini, line 212: Not necessarily in this PR, but ideally we should be maintaining a reference to this and disposing of the event handler in the zoom control `dispose` method.

core/zoom_controls.js
```diff
@@ -260,16 +253,28 @@ Blockly.ZoomControls.prototype.createZoomInSvg_ = function(rnd) {
...
+      zoominSvg, 'mousedown', null, this.zoom_.bind(this, 1));
+};
+
+/**
+ * Zooms in or out of the workspace.
```
samelhusseini, line 264: I think this is too generic of a description for this method. This isn't a method for zooming in / out, it's a method that handles a mouse event for both zoom in and out buttons, perhaps something like: 

"Handles a mouse-down on the Zoom in or Zoom out buttons on the workspace."

core/zoom_controls.js
```diff
@@ -312,19 +316,40 @@ Blockly.ZoomControls.prototype.createZoomResetSvg_ = function(rnd) {
...
-    e.stopPropagation();  // Don't start a workspace scroll.
-    e.preventDefault();  // Stop double-clicking from selecting text.
-  });
+  Blockly.bindEventWithChecks_(
+      zoomresetSvg, 'mousedown', null, this.resetZoom_.bind(this));
+};
+
+/**
+ * Resets zoom on the workspace.
+ * @param {!Event} e A mouse down event.
```
samelhusseini, line 328: Same here,

"Handles a mouse-down on the reset zoom button on the workspace."

core/zoom_controls.js
moniika: sgtm I'll make a follow up PR to address this and add you to it

PR #4035, opened 2020-07-11
Add listener for flyout events to playground.

### LINKED ISSUE #3976: Flyout Fires Workspace Click Event (parent workspace id)
**Describe the bug**

Very similar to #3975. I'm not sure what the desired behavior is here.

When you click on the background of an open flyout it fires a workspace click event. The workspaceId in the event is not the id of the flyout itself, but the id of the parent workspace.

**To Reproduce**

1. Turn on event logging.
2. Open the flyout.
3. Click the background of the flyout.
4. Observe the fired click event.

**Expected behavior**

Again I'm just not sure how it should work in this situation. I think my instinct is that it would fire, but it would use the flyout's id and not the id of the parent. But I'm not sure that that's the most useful way to structure it.

moniika:
## The basics

- [x] I branched from develop

### Resolves

#3976

### Proposed Changes

- Changes workspace linked to event to `startWorkspace_` for firing workspace click events.
- Adds option to listen to flyout workspace events

### Reason for Changes

- This changes makes it so the workspace id for a click event made on a flyout references the flyout workspace.
- Adding an option to listen to flyout workspace events is useful for debugging. Being able to only listen to main workspace events is valuable because it reduces the amount of events.

### Test Coverage

Tested on playground.

Tested on:
* Desktop Chrome

### Screenshots
The log events checkbox:
image
was changed to log main workspace events and log flyout events:
image

### DIFF
--- a/tests/playground.html
+++ b/tests/playground.html
@@ -147,10 +147,10 @@ function start() {
       document.getElementById('importExport').value = text;
     }
     // Restore event logging state.
-    var state = sessionStorage.getItem('logEvents');
-    logEvents(Boolean(Number(state)));
-    state = sessionStorage.getItem('logFlyoutEvents');
-    logFlyoutEvents(Boolean(Number(state)));
+    var logMainEventsState = sessionStorage.getItem('logEvents');
+    logEvents(Boolean(Number(logMainEventsState)));
+    var logToolboxFlyoutEventsState = sessionStorage.getItem('logFlyoutEvents');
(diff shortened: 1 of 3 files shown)


tests/playground.html
```diff
@@ -149,6 +149,8 @@
     // Restore event logging state.
     var state = sessionStorage.getItem('logEvents');
     logEvents(Boolean(Number(state)));
+    state = sessionStorage.getItem('logFlyoutEvents');
```
alschmiedt, line 152: nit: Use a different variable.

PR #4036, opened 2020-07-11
Use generics to derive SVG element type

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Use generics to automatically derive the specific SVG element type, rather than have to cast it everywhere.

### Reason for Changes

Cleaner code.

### Test Coverage

Tested with the playground.
Build compiles successfully.

Tested on:

### DIFF
--- a/core/utils/dom.js
+++ b/core/utils/dom.js
@@ -157,7 +157,7 @@ Blockly.utils.dom.SVGElementType.SVGLineElement =
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGPathElement>} */
 Blockly.utils.dom.SVGElementType.SVGPathElement =
-  new Blockly.utils.dom.SVGElementType('path');
+    new Blockly.utils.dom.SVGElementType('path');
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGPatternElement>} */
 Blockly.utils.dom.SVGElementType.SVGPatternElement =
@@ -169,7 +169,7 @@ Blockly.utils.dom.SVGElementType.SVGPolygonElement =
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGRectElement>} */
 Blockly.utils.dom.SVGElementType.SVGRectElement =
-  new Blockly.utils.dom.SVGElementType('rect');
+    new Blockly.utils.dom.SVGElementType('rect');
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGSVGElement>} */
 Blockly.utils.dom.SVGElementType.SVGSVGElement =
@@ -177,11 +177,11 @@ Blockly.utils.dom.SVGElementType.SVGSVGElement =
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGTextElement>} */
 Blockly.utils.dom.SVGElementType.SVGTextElement =
-  new Blockly.utils.dom.SVGElementType('text');
+    new Blockly.utils.dom.SVGElementType('text');
 
 /** @type {!Blockly.utils.dom.SVGElementType<!SVGTSpanElement>} */
 Blockly.utils.dom.SVGElementType.SVGTSpanElement =
-  new Blockly.utils.dom.SVGElementType('tspan');
+    new Blockly.utils.dom.SVGElementType('tspan');
 
 
 /**
--- a/core/utils/dom.js
+++ b/core/utils/dom.js
@@ -78,7 +78,7 @@ Blockly.utils.dom.canvasContext_ = null;
  * @constructor
  * @template T
  */
-Blockly.utils.dom.SVGElementType = function(tagName) {
+Blockly.utils.dom.SvgElementType = function(tagName) {
   /**
    * @type {string}
    * @private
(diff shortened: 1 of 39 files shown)


core/zoom_controls.js
```diff
@@ -30,7 +30,7 @@ Blockly.ZoomControls = function(workspace) {
    */
   this.workspace_ = workspace;
 
-  /**
+  /*
```
maribethb, line 33: nit, but I think you need this *

core/zoom_controls.js
```diff
@@ -53,6 +53,35 @@ Blockly.ZoomControls = function(workspace) {
...
+   * The vertical distance between the workspace bottom edge and the control.
+   * The value is initialized during `init`.
+   * @type {?number}
+   * @private
+   */
+  this.verticalSpacing_ = null;
+
+  /**
+   * The zoom in svg <g> element.
+   * @type {SVGGElement}
```
maribethb, line 67: These should have explicit `?` if they are nullable, I think.

core/utils/dom.js
```diff
@@ -72,16 +72,132 @@ Blockly.utils.dom.cacheReference_ = 0;
...
+ * Returns the SVG element tag name.
+ * @return {string} The name.
+ * @override
+ */
+Blockly.utils.dom.SVGElementType.prototype.toString = function() {
+  return this.tagName_;
+};
+
+/** @type {!Blockly.utils.dom.SVGElementType<!SVGAnimateElement>} */
+Blockly.utils.dom.SVGElementType.SVGAnimateElement =
```
maribethb, line 99: Since  'SVGElement' is already in the name, do you think it might be more clear to have final part of the path be just the element name? e.g. `Blockly.utils.dom.SvgElementType.Animate`?
I find 'SVGSVGElement` and `SVGGElement` both really hard to read and this would solve it for both of them and make all of it just a bit less verbose.

core/utils/dom.js
```diff
@@ -72,16 +72,132 @@ Blockly.utils.dom.cacheReference_ = 0;
...
+/** @type {!Blockly.utils.dom.SVGElementType<!SVGClipPathElement>} */
+Blockly.utils.dom.SVGElementType.SVGClipPathElement =
+    new Blockly.utils.dom.SVGElementType('clipPath');
+
+/** @type {!Blockly.utils.dom.SVGElementType<!SVGDefsElement>} */
+Blockly.utils.dom.SVGElementType.SVGDefsElement =
+    new Blockly.utils.dom.SVGElementType('defs');
+
+/** @type {!Blockly.utils.dom.SVGElementType<!SVGFECompositeElement>} */
+Blockly.utils.dom.SVGElementType.SVGFECompositeElement =
```
maribethb, line 115: I think the camel casing for all of these should be
`SvgElementType` (keeping with the pre-existing function style like `createSvgElement` and https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined)

For all of the ones that start with `fe` I think the rules might be a little ambiguous but personally I think `SvgFeCompositeElement` (or `FeComposite`) is a bit more readable but this one's more up to interpretation I think.

core/zoom_controls.js
samelhusseini: Good catch, that was a product of a merge.

core/zoom_controls.js
samelhusseini: Generally the rule is to not use redundant types. In this case, SVGGElement is an object, and objects are implicitly nullable, so no need to add the `?`. 
Literals on the other hand (string, boolean, number) are non-nullable implicitly. I know it's all a little confusing!

core/utils/dom.js
samelhusseini: Agreed it's long. My concern is it's not easily derivable at the call site what type of element you are getting back. This way, you know that calling `Blockly.createSvgElement` with name `Blockly.utils.dom.SVGElementType.SVGAnimateElement` is going to yield an `SVGAnimateElement` type.

I would be open to changing the `Blockly.utils.dom.SVGElementType` to something else if you think of an alternative there.

core/utils/dom.js
samelhusseini: Similarly to the previous comment, the value of the end of the type there is to use the name as an indicator to the exact type you're getting. Being able to infer that `Blockly.utils.dom.SVGElementType.SVGFECompositeElement` returns an element of type `SVGFECompositeElement`. 

I agree that `Blockly.utils.dom.SVGElementType` should change to `Blockly.utils.dom.SvgElementType`.

core/utils/dom.js
samelhusseini: Oh. On second thought, this doesn't make sense at all.

If you were used to doing: 
```Blockly.utils.dom.createSvgElement('g'`` Now you need to know that g is in fact an SVGGElement. That's bizarre, so good point. Changing them all to use the name.

core/zoom_controls.js
maribethb: According to this, reference types should always be annotated. https://google.github.io/styleguide/jsguide.html#jsdoc-nullability
I think this changed at some point, so maybe Blockly is using an older version of the style guide? Elsewhere in this file some classes are annotated with `?` though

core/utils/dom.js
maribethb: Sorry, I think I confused myself, but here I was only nitpicking the capitalization of `Svg` and `Fe` in the interest of readability. But I think I didn't realize that these types already existed so you can't change their names, so this makes a lot more sense now :)

But while I am on the topic of nitpicking capitalization, if you are choosing constant case for the names now, then should there be underscores between words? e.g. `CLIP_PATH` instead of `CLIPPATH` ?

I find a lot of capital letters in a row hard to read esp. since I'm not super familiar with SVG types so the words don't stand out easily :)

maribethb:
By the way, I'm mostly familiar with the internal JS style guide which I am not sure is identical to the one I linked to here, so if they differ in relevant areas or if there is an alternate source for the style Blockly follows, please let me know!

core/utils/dom.js
samelhusseini: Normally I would agree with sticking with constant case, but I think in this case, there's an advantage of being able to directly convert from something like the tag name (eg: `clipPath`) to type name (eg: CLIPPATH), without having to wonder if there is an underscore.

samelhusseini:
Will do, all of them differ in some way shape or form. It's mostly a legacy thing. At some point the Blockly style matched the Google style guide, but the Google style guide grew and left parts of Blockly behind.
In general, local file style takes precedence over style guide. If you find things in Blockly you'd like to "upgrade", you're welcome to, as long as it's done holistically, and the team is onboard (just to avoid inconsistencies).

core/zoom_controls.js
samelhusseini: Yeah I think this has changed over time.

Elsewhere in the file, I think you're referring to `Blockly.EventData`. That is not an Object, but a record type, and they are non-nullable by default.
This is a case in point that what we are doing in Blockly is confusing. We really _should_ be doing what the style guide suggests there, and always add the modified for anything that's not a literal to be clear about its nullability.

I think that's going to be a heavy lift to switch over. Let's discuss at the meeting next week, and see if we should look into doing this as a big PR (with some find and replace), and switch over the Blockly style to match the Google style.

For this PR, I'll keep as is.

maribethb: Sounds good!

core/utils/dom.js
maribethb: Personally, I think using a non-standard format of constant case actually makes it harder, because following the normal algorithm I would expect there to be an underscore at the camelCase boundaries. But I will defer to your judgment :)

PR #4046, opened 2020-07-14
Perf: Move multiline xml text encoding out of domToText

### LINKED ISSUE #4030: domToText: Performance issue in large projects
I'm investigating performance issues when working with large blocks programs (in MakeCode Arcade), and Blockly's `domToText` method came up as a hot path due to a regexp replace it executes. This regexp was added in https://github.com/google/blockly/pull/2663 (Textarea Field, Multiline Block (from acbart)). Here's the line in question: https://github.com/google/blockly/blob/ba68081d8f0f8e81d76baf370716d5955bae82a1/core/xml.js#L319

This regexp runs repeatedly on the document, causing `domToText` to take a significant amount of time when the xml is very large. *Without this regexp, saving takes practically no time at all.*

If you could help me understand the purpose of this search/replace, perhaps I could come up with another approach. And if you have a suggestion for how to improve or remove it, I would be happy to try it and report back.

image

OS: Windows 10
Browser: Chrome 83

eanders-MS:
## The basics

- [x] I branched from develop

## The details

A complex regular expression in `Blockly.Xml.domToText` was running on every save, causing performance problems on large projects. The regexp progressively replaces all `\n` with `&#10;` in content fields, and was added along with `FieldMultilineInput`, in #2663.

### Proposed Changes

Override `toXml` and `fromXml` on `FieldMultilineInput` and do the conversion there. This simplifies the regexp, and it only runs when there is multiline content in the program, improving performance - especially on large programs.

### Performance Comparison

Testing with this MakeCode Arcade game: https://arcade.makecode.com/74548-67333-76127-33508

Time spent in `Blockly.Xml.domToText`:

| Before | After |
| -- | -- |
| 10s | 250ms |

### Resolves

Resolves #4030

### Test Coverage

I tested this change in playground.html. Steps I followed:
1. Add a few paragraph blocks and populate them with multiline content.
2. Export to Xml. Inspect the Xml to verify content encoding.
3. Delete all blocks.
4. Import from Xml. Verify (visually) that the content looks correct.

Also: In the dev console, set breakpoints in the `toXml` and `fromXml` methods of `FieldMultilineInput` to verify they're getting called at the right times.

Tested on:
* Desktop Chrome

### Additional information

**Of note:** Another document-wide regexp has appeared in `domToText`, and will likely become a performance issue. You may want to fix it: https://github.com/google/blockly/blob/ba68081d8f0f8e81d76baf370716d5955bae82a1/core/xml.js#L327

I can see how `domToText` would be a tempting place to fixup formatting issues. It might be worth adding a comment in that method stating: "For performance reasons, don't run regular expressions on the whole document."

### DIFF
--- a/core/field_multilineinput.js
+++ b/core/field_multilineinput.js
@@ -77,9 +77,10 @@ Blockly.FieldMultilineInput.fromJson = function(options) {
  * @package
  */
 Blockly.FieldMultilineInput.prototype.toXml = function(fieldElement) {
-  // Replace '\n' characters with html-escaped equivalent '&#10'. This is needed
-  // so the plain-text representation of the xml produced by `Blockly.Xml.domToText`
-  // will appear on a single line (this is a limitation of the plain-text format).
+  // Replace '\n' characters with html-escaped equivalent '&#10'. This is
+  // needed so the plain-text representation of the xml produced by
+  // `Blockly.Xml.domToText` will appear on a single line (this is a limitation
+  // of the plain-text format).
   fieldElement.textContent = this.getValue().replace(/\n/g, '&#10;');
   return fieldElement;
 };
--- a/core/field_multilineinput.js
+++ b/core/field_multilineinput.js
@@ -70,6 +70,32 @@ Blockly.FieldMultilineInput.fromJson = function(options) {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


eanders-MS:
@googlebot I signed it!

samelhusseini:
Thanks!

Issue #1357, opened 2017-10-11
Add a 'strict' flag on connections


rachel-fenichel:
### Problem statement

- [ ] Bug report
- [x] Feature request

A connection with the `null` type will connect to any other connection, regardless of the other connection's type.  As a result, making a block that doesn't allow default blocks to connect to it is difficult: you have to change the connection types on all existing blocks.

### Requested Behavior

I'm considering adding an optional `strict` flag to connections.  

A connection with the `strict` flag set will only connect to another connection if they both have at least one type in common in their `checks`.

Compatible:
`['Zebra', 'Mongoose']` and `['Cobra', 'Mongoose']`
`null` and `null`

Incompatible:
`['Zebra', 'Mongoose']` and `['Lion', 'Cheetah']`
`['Zebra', 'Mongoose']` and `null`

This would let a developer easily add a few strict checks without needing to update all of their blocks.

### Additional Information
Related to https://github.com/LLK/scratch-blocks/issues/1127

rachel-fenichel:
@vicng @AnmAtAnm 

Do you have thoughts about this, particularly with respect to Android and iOS?

bkisdi:
I needed it in my projects too. I changed all the standard blocks to have 'Statement' type - this is to be done once (well, once after each download), and it solved the problem.

1. In all files of blockly/blocks replace all
setPreviousStatement(true);
with
setPreviousStatement(true, 'Statement');

2. In all files of blockly/blocks replace all
setNextStatement(true);
with
setNextStatement(true, 'Statement');

Please.

vicng:
@rachel-fenichel Can you add a potential JSON spec to the proposal?

rachel-fenichel:
## Options
- Don't add a strict flag at all, but make it easier to modify types of connections (in developer tools). 
  - No changes in core Blockly.
- Add a strict flag at the block level (we would choose a more descriptive name).
  - Adding a flag at the block level would be the simplest from an API perspective, but means that developers need to recognize that the block-level flag has an impact on connection types.  The block-level strict flag makes all connections on the block strict.  To make a non-strict connection on a strict block, a developer would set the type of that connection to the wildcard (null).

Block JSON would look like this:
```
{
  //...
  previousStatement: null,
  nextStatement: ["Panda", "Koala"],
  strict: true
}
```

- Add a strict flag at the connection level, by making connections into objects in JSON.
  - This keeps the strict specifier as close as possible to the connection definition, but means we need to keep backwards compatibility.

[...]

PR #4050, opened 2020-07-16
Connection checker object

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #1357

### Proposed Changes

- Create a new `IConnectionChecker` interface and `ConnectionChecker` class
- Store the connections checker on the workspace and in the connection db.
- Use the new checker object everywhere
- Update tests

### Reason for Changes

Extract logic so that alternate type systems can be implemented as extensions.

### Test Coverage

The tests in connection_tests.js and connection_db_tests.js were actually just checking this logic in a lot of cases. I updated them to point to the correct objects.

### Documentation

Eventually we will want to have a codelab or example for implementing your own connection checks.

### Additional Information

Next steps:
- Add deprecation warnings for old functions
- Add tests for old functions 
- Register the checker through the registry, instead of always creating an instance of `Blockly.ConnectionChecker`. (separate PR)

The first two I'm working on now, but wanted to kick off the PR for general review first.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -16,6 +16,7 @@ goog.require('Blockly.Events');
 goog.require('Blockly.Events.BlockMove');
 goog.require('Blockly.Xml');
 
+goog.requireType('Blockly.ConnectionTypeChecker');
 goog.requireType('Blockly.IASTNodeLocationWithBlock');
 
 
@@ -145,8 +146,10 @@ Blockly.Connection.prototype.connect_ = function(childConnection) {
         if (nextBlock && !nextBlock.isShadow()) {
           newBlock = nextBlock;
         } else {
-          if (orphanBlock.previousConnection.checkType(
-              newBlock.nextConnection)) {
+          if (orphanBlock.workspace.connectionTypeChecker.checkType(
+              orphanBlock.previousConnection, newBlock.nextConnection)) {
+            //orphanBlock.previousConnection.checkType(
+              //newBlock.nextConnection)) {
             newBlock.nextConnection.connect(orphanBlock.previousConnection);
             orphanBlock = null;
           }
@@ -247,28 +250,29 @@ Blockly.Connection.prototype.isConnected = function() {
  *    an error code otherwise.
  */
 Blockly.Connection.prototype.canConnectWithReason = function(target) {
-  if (!target) {
-    return Blockly.Connection.REASON_TARGET_NULL;
-  }
-  if (this.isSuperior()) {
-    var blockA = this.sourceBlock_;
-    var blockB = target.getSourceBlock();
-  } else {
-    var blockB = this.sourceBlock_;
-    var blockA = target.getSourceBlock();
-  }
-  if (blockA && blockA == blockB) {
-    return Blockly.Connection.REASON_SELF_CONNECTION;
-  } else if (target.type != Blockly.OPPOSITE_TYPE[this.type]) {
-    return Blockly.Connection.REASON_WRONG_TYPE;
-  } else if (blockA && blockB && blockA.workspace !== blockB.workspace) {
-    return Blockly.Connection.REASON_DIFFERENT_WORKSPACES;
-  } else if (!this.checkType(target)) {
-    return Blockly.Connection.REASON_CHECKS_FAILED;
-  } else if (blockA.isShadow() && !blockB.isShadow()) {
-    return Blockly.Connection.REASON_SHADOW_PARENT;
-  }
-  return Blockly.Connection.CAN_CONNECT;
+  return this.sourceBlock_.workspace.connectionTypeChecker.canConnectWithReason(this, target);
+  // if (!target) {
+  //   return Blockly.Connection.REASON_TARGET_NULL;
+  // }
+  // if (this.isSuperior()) {
+  //   var blockA = this.sourceBlock_;
+  //   var blockB = target.getSourceBlock();
+  // } else {
+  //   var blockB = this.sourceBlock_;
+  //   var blockA = target.getSourceBlock();
+  // }
+  // if (blockA && blockA == blockB) {
+  //   return Blockly.Connection.REASON_SELF_CONNECTION;
+  // } else if (target.type != Blockly.OPPOSITE_TYPE[this.type]) {
+  //   return Blockly.Connection.REASON_WRONG_TYPE;
+  // } else if (blockA && blockB && blockA.workspace !== blockB.workspace) {
+  //   return Blockly.Connection.REASON_DIFFERENT_WORKSPACES;
+  // } else if (!this.checkType(target)) {
+  //   return Blockly.Connection.REASON_CHECKS_FAILED;
+  // } else if (blockA.isShadow() && !blockB.isShadow()) {
+  //   return Blockly.Connection.REASON_SHADOW_PARENT;
+  // }
+  // return Blockly.Connection.CAN_CONNECT;
 };
 
 /**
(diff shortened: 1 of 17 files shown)


core/connection_db.js
```diff
@@ -235,14 +247,17 @@ Blockly.ConnectionDB.prototype.searchForClosest = function(conn, maxRadius,
...
   var temp;
+  var curDistance;
 
   // Walk forward and back on the y axis looking for the closest x,y point.
   var pointerMin = closestIndex - 1;
   while (pointerMin >= 0 && this.isInYRange_(pointerMin, conn.y, maxRadius)) {
     temp = this.connections_[pointerMin];
-    if (conn.isConnectionAllowed(temp, bestRadius)) {
+    curDistance = temp.distanceFrom(conn);
+    if (curDistance <= bestRadius &&
```
samelhusseini, line 257: This moves the logic for distance based connection checking outside of the connection checker. I'm trying to think of a specific example where someone would want to make a connection decision based on the distance. Perhaps using a different `bestRadius` value based on the type of connection (input / output vs next/previous).

Similar to how we have a drag check, could we have a check that also takes into account the distance (when `canConnect` is called with a distance).

core/interfaces/i_connection_checker.js
```diff
@@ -0,0 +1,87 @@
...
+ * @fileoverview The interface for an object that encapsulates logic for
+ * checking whether a potential connection is safe and valid.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
+'use strict';
+
+goog.provide('Blockly.IConnectionChecker');
+
+goog.requireType('Blockly.Connection');
+
```
samelhusseini, line 17: nit: new line

core/interfaces/i_connection_checker.js
```diff
@@ -0,0 +1,87 @@
...
+ * @public
+ */
+Blockly.IConnectionChecker.prototype.canConnect;
+
+/**
+ * Checks whether the current connection can connect with the target
+ * connection, and return an error code if there are problems.
+ * @param {Blockly.Connection} one Connection to check compatibility with.
+ * @param {Blockly.Connection} two Connection to check compatibility with.
+ * @param {boolean} isDragging [description]
```
samelhusseini, line 41: Add description

core/keyboard_nav/navigation.js
```diff
@@ -499,13 +500,11 @@ Blockly.navigation.connect_ = function(movingConnection, destConnection) {
...
     return true;
   } else {
-    try {
-      destConnection.checkConnection(movingConnection);
-    }
-    catch (e) {
-      // If nothing worked report the error from the original connections.
-      Blockly.navigation.warn_('Connection failed with error: ' + e);
-    }
+    var checker = movingConnection.getConnectionChecker();
```
samelhusseini, line 503: This part doesn't make sense, it used to be a try/check, now we're warning every time?

core/connection_checker.js
```diff
@@ -0,0 +1,265 @@
...
+  if (Blockly.draggingConnections.indexOf(two) != -1) {
+    return false;
+  }
+
+  return true;
+};
+
+/**
+ * Helper function for drag checking.
+ * @param {!Blockly.Connection} one The connection to check, which must be a
```
samelhusseini, line 233: Just a suggestion, instead of `one` / `two`, we could use `a` and `b`. This matches other util comparison methods we have, and is shorter. There's also a couple of places where we're using things like `blockA` / `blockB` already, so that would match there.
Your call.

core/connection.js
```diff
@@ -16,6 +16,7 @@ goog.require('Blockly.Events');
 goog.require('Blockly.Events.BlockMove');
 goog.require('Blockly.Xml');
 
+goog.requireType('Blockly.IConnectionChecker');
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

core/connection_checker.js
```diff
@@ -0,0 +1,265 @@
...
+ * @fileoverview An object that encapsulates logic for checking whether a potential
+ * connection is safe and valid.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
+'use strict';
+
+goog.provide('Blockly.ConnectionChecker');
+
+goog.requireType('Blockly.IConnectionChecker');
+goog.requireType('Blockly.Connection');
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

core/connection_checker.js
```diff
@@ -0,0 +1,265 @@
...
+
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

core/connection_db.js
rachel-fenichel: The big advantage of the current setup is that they types are clean: everything is a Connection.

I *could* make it so that we cast to RenderedConnection if `isDragging` is true, but that still means I need another param for the distance.

Essentially anything I do to `canConnect` I also need to do to `canConnectWithReason`, which is frustrating.

core/interfaces/i_connection_checker.js
rachel-fenichel: Done.

core/interfaces/i_connection_checker.js
rachel-fenichel: Done, here and in connection_checker.

core/keyboard_nav/navigation.js
rachel-fenichel: This part used to be a pretty janky way of getting the error message out. It was only ever called when we knew that the connection would fail and an error would be thrown.

core/connection_checker.js
rachel-fenichel: Done.

rachel-fenichel:
Updated with deprecation warnings and tests, PTAL at the new files.

core/connection_db.js
samelhusseini: Ah, what if we passed in a scope object that describes the state.
```
{
  isDragging: boolean?,
  distance: number?,
  ... 
}
```

core/connection.js
```diff
@@ -598,18 +497,9 @@ Blockly.Connection.prototype.targetBlock = function() {
...
-  }
-  // Find any intersection in the check lists.
-  for (var i = 0; i < this.check_.length; i++) {
-    if (otherConnection.check_.indexOf(this.check_[i]) != -1) {
-      return true;
-    }
-  }
-  // No intersection.
-  return false;
+  // TODO (fenichel): Add deprecation warnings.
```
samelhusseini, line 500: One more deprecation warning here.

core/keyboard_nav/navigation.js
samelhusseini: Oh okay.

samelhusseini: Could we add a comment, as it does seem weird to assume the connection failed after checking it.

core/connection_db.js
rachel-fenichel: What's the advantage of having it be a scope object instead of just adding another arg? It doesn't change the typing/casting issue.

samelhusseini: Saves you from having to pass parameters around, that's all.
I think what you said about casting when there's a drag makes sense. Most checks would be done on the `Connection` type, with the exception of drag checks.

core/connection.js
rachel-fenichel: Done.

PR #4053, opened 2020-07-17
Move events into the global registry

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Register all events in the global registry to make it easier to add new events that can be deserialized by type in `Blockly.Events.fromJson`. 

We now have two `fromJson` methods, one that is static, and one one the instance level. Open to ideas there.

I wanted to give events the flexibility to do whatever is needed in the `fromJson` static method, and so I haven't abstracted away the call to the local `fromJson` or setting `workspaceId`. 

Given `fromJson` and `workspaceId` are on the Abstract event, I'm open to moving these two calls to `Blockly.Events.fromJson`.

### Reason for Changes

Simplify event extensions

### Test Coverage
Tested compressed with the mirror demo.

Tested on:


samelhusseini:
Yeah, that would instead make all arguments for events optional. @rachel-fenichel thoughts?

rachel-fenichel:
lgtm once tests pass

samelhusseini:
Tests pass locally, looks to be a travis issue with running our mocha tests.. looking

core/block_events.js
```diff
@@ -31,19 +32,26 @@ goog.require('Blockly.utils.xml');
 
 /**
  * Abstract class for a block event.
- * @param {Blockly.Block} block The block this event corresponds to.
+ * @param {!Blockly.Block=} opt_block The block this event corresponds to.
```
moniika, line 35: Does this mean that we are requiring that if a block is not passed, then we need to pass undefined, instead of null? I believe currently we either pass null or undefined.
This case (when block is not passed) happens often for `Blockly.Events.UI` (since some UI events do not have an associated block, but block is the first param for the constructor).

samelhusseini: We were currently only allowing Block or null for Block events. With the switch to make everything optional, I don't see a need for allowing null there. Null was being used to signal a blank event, and that has been taken care of my making everything optional.

In the case of UI events, we still allow `null`, as null for the block parameter means that the event is not associated with a block. See jsdoc for UI events:
```
 * @param {?Blockly.Block=} opt_block The affected block.  Null for UI events
 *     that do not have an associated block.  Undefined for a blank event.
```

moniika: I see, I got confused because I mistook this for the `Blockly.Events.Abstract` class. That makes sense, thanks for clarifying.

PR #4059, opened 2020-07-22
Add Flyout as a Plugin

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
This adds an interface for a flyout and the ability to add a toolbox flyout. 

### Proposed Changes
The registry doesn't really support nesting the name of the plugin. 
I don't love the way I implemented this, but was having issues coming up with better ways to do this.

### Test Coverage

Tested on:

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -42,6 +42,7 @@ Blockly.utils.object.inherits(Blockly.HorizontalFlyout, Blockly.Flyout);
  * @type {string}
  */
 Blockly.HorizontalFlyout.registryName = 'horizontalFlyout';
+
 /**
  * Return an object with all the metrics required to size scrollbars for the
  * flyout.  The following properties are computed:
@@ -378,5 +379,5 @@ Blockly.HorizontalFlyout.prototype.reflowInternal_ = function() {
   }
 };
 
-Blockly.registry.register(Blockly.registry.Type.TOOLBOX_FLYOUT,
+Blockly.registry.register(Blockly.registry.Type.FLYOUTS_TOOLBOX,
     Blockly.HorizontalFlyout.registryName, Blockly.HorizontalFlyout);
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -14,6 +14,7 @@ goog.provide('Blockly.HorizontalFlyout');
 
 goog.require('Blockly.Block');
 goog.require('Blockly.Flyout');
+goog.require('Blockly.registry');
 goog.require('Blockly.Scrollbar');
 goog.require('Blockly.utils');
 goog.require('Blockly.utils.object');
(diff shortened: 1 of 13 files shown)


core/flyout_horizontal.js
```diff
@@ -32,11 +33,16 @@ goog.requireType('Blockly.utils.Metrics');
...
-  
   this.horizontalLayout = true;
 };
 Blockly.utils.object.inherits(Blockly.HorizontalFlyout, Blockly.Flyout);
 
+/**
+ * The name of the horizontal flyout in the registry.
+ * @type {string}
+ */
+Blockly.HorizontalFlyout.registryName = 'horizontalFlyout';
```
samelhusseini, line 44: We should standardize on one name for this. This is what Rachel has in her strict-connection-checker plugin: 
https://github.com/google/blockly-samples/pull/333/files#diff-459bd1d34908031cb468abe7c18d912bR55

Not saying that's what we should go with, just that we should align.

core/registry.js
```diff
@@ -220,6 +231,10 @@ Blockly.registry.getClassFromOptions = function(type, options) {
   var typeName = type.toString();
   var plugin = options.plugins[typeName] || Blockly.registry.DEFAULT;
 
+  if (Blockly.registry.flyoutTypes_.indexOf(type) > -1 && options.plugins.flyouts) {
+    plugin = options.plugins.flyouts[typeName] || Blockly.registry.DEFAULT;
```
samelhusseini, line 235: Not a huge fan of this approach, as it means whenever we want to add another plugin group, we'd need to add additional login in the global registry to support that group.
Here are a couple of alternatives: 

Only support a single flat list in plugins, for semantically similar groups, you can have a shared prefix, eg: 
```
{
   ...
   'plugins': {
         'toolbox': ...,
         'flyouts-toolbox': ...,
         'flyouts-trashcan': ...,
         'flyouts-mutator': ...,
    }
}
```

Alternately, you can support a nested list in plugins, but flatten it when parsed. such that this:
```
{
    'plugins': {
          'flyouts': {
               'toolbox': ...,
               'trashcan': ...,
           }
     }
}
```
turns into: 
```
{
     'plugins': {
         'flyouts.toolbox': ...,
         'flyouts.trashcan': ...,
     }
}
```

core/toolbox.js
```diff
@@ -184,22 +185,24 @@ Blockly.Toolbox.prototype.init = function() {
...
-    this.flyout_ = new Blockly.VerticalFlyout(workspaceOptions);
+
+  var FlyoutClass = Blockly.registry.getClassFromOptions(
+      Blockly.registry.Type.FLYOUTS_TOOLBOX, this.workspace_.options);
+  var defaultClass = Blockly.registry.getClass(
+      Blockly.registry.Type.FLYOUTS_TOOLBOX, Blockly.registry.DEFAULT)
+
+  if (FlyoutClass == defaultClass && workspace.horizontalLayout) {
+    FlyoutClass = Blockly.registry.getClass(Blockly.registry.Type.FLYOUTS_TOOLBOX,
+        Blockly.HorizontalFlyout.registryName);
```
samelhusseini, line 196: I can see why this is necessary (we want to use a different default based on the horizontalLayout flag). Do you think doing this would make the logic clearer:

- Instead of registering the two flyouts (vertical and horizontal) under the flyouts_toolbox type, register each under a different type (eg: flyouts_vertical_toolbox, flyouts_horizontal_toolbox), that way each has its own default and you're able to keep similar logic that bases which flyout to use on whether the workspace.horizontalLayout is present.

core/flyout_horizontal.js
alschmiedt: Ok, will keep in mind for the future. I ended up going with having flyout-vertical-toolbox and flyout-horizontal-toolbox so ended up removing this.

core/registry.js
alschmiedt: ok, went with option 1 for this pr. If we decide we really want a nested structure I can update in the future.

core/toolbox.js
alschmiedt: Yeah this definitely makes the logic much nicer. If we get to the point where we can move the horizontal flyout into a plugin we can probably get rid of the two names and just use flyout-toolbox.

samelhusseini: That sounds good.

Issue #4065, opened 2020-07-23
Add missing tests for events triggered by clicks on workspace


moniika:
Recent changes logic of `Blockly.Events.UI` event emition should be tested in mocha tests.
Ideally tests would also cover checking that events are not propagated where appropriate (i.e. click on trashcan does not also trigger workspace click)

**The following scenarios should be tested:**
- [x] workspace click (should emit workspace click)
- [x] click on non-empty trashcan (should emit "trashcanOpen" true)
- [x] click on empty trashcan (should emit workspace click)
- [ ] closing trashcan flyout (should emit "trashcanOpen" false) (scenarios: workspace click and dragging block onto workspace)
- [x] click on zoom controls (zoom in, zoom out, zoom reset)

**On hold while Toolbox changes are being made (changes are being made to events emitted):**
- [ ] toolbox flyout workspace click (should emit workspace click, with the correct workspace id)
- [ ] opening toolbox flyout/category
- [ ] closing toolbox flyout/category

Issue #4064, opened 2020-07-23
Add cleanup for Events in mocha tests


moniika:
Many tests trigger Events fired on a timeout, which can end up firing outside of the tests (if that test didn't specifically stub events to fire immediately or may have some other method called on a timeout that then triggers an event).

This can cause unexpected tests failures in tests that check for what events were fired during that test.
Additionally, there is a lot of duplicated code relating the clearing the Event state (this may resolve the issue with event firing, but would need some investigation) that should be refactored and called in most tests.

moniika:
Event cleanup added to all current mocha tests (through the use of `sharedTestSetup` `sharedTestTeardown`

PR #4066, opened 2020-07-23
Refactor shared

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4065

### Proposed Changes

- Refactored logic for stubbing `Blockly.utils.genUid` into shared_helpers.js and updated usages.
- Refactored helper method for asserting Event values into shared_helpers (`assertEventEquals`)
- Created helper methods for assertions on spys for event firing
- Refactored `comment_test.js` `event_test.js` and `trashcan_test.js` to use new event test helpers
- Added test for testing emitting UI event for "trashcanOpen"

### Test Coverage

Ran mocha tests.

Tested on:
* Desktop Chrome

### Additional Information

Test methods were refactored as helpers were developed to better fit the needs of multiple tests, causing multiple files to be touched. 

#4062 fixes eslint errors on the PR.

Initially after this refactor the test failed when they were running concurrently, revealing that some tests cause events to be fired on a timeout which causes failures in other event tests. A bug was filed #4064 to further investigate and resolve this issue. As a temporary fix, the event queue is cleared when a stub is created for `Blockly.Events.fire` but it is not the ideal fix.


tests/mocha/test_helpers.js
```diff
@@ -42,13 +45,37 @@ function captureWarnings(innerFunc) {
...
  * Creates stub for Blockly.Events.fire that fires events immediately instead of
  * with timeout.
- * @return {sinon.stub} The created stub.
+ * @return {!SinonStub} The created stub.
  */
 function createEventsFireStub() {
+  // TODO(#4064): Remove clearing of event clear here in favor of adding cleanup
+  // to other tests that cause events to be added to the queue even after they
+  // end.
+  Blockly.Events.FIRE_QUEUE_.length = 0;
```
moniika, line 78: This fix **seems** to work.

However, I **thought** I had tested with this change at some point and it didn't work, but it works now.
I'm a little worried that this fixes most failures, but there's still "sometimes" failures, which is why I want to address this better in a follow up as part of #4064.

moniika:
Upon investigation, there are other instances where Blockly.Events.fire is stubbed/spied such as in `block_test.js` and `toolbox_test.js`. I plan to address in a follow up PR, to reduce the amount of changes in this PR, but I expect the changes to be very similar to what I've done here.
These other tests I've found that are not stubbing them to fire immediately may also be part of the issue in #4064.

PR #4078, opened 2020-07-28
Adds methods migrated from old toolbox

alschmiedt:
## The basics

- [] I branched from develop

### Resolves
This PR mainly adds back in methods from the old toolbox that are necessary.
The exceptions to this are: 
getToolboxItemById
getSelectedItem

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox/toolbox.js
+++ b/core/toolbox/toolbox.js
@@ -509,7 +509,7 @@ Blockly.Toolbox.prototype.clearSelection = function() {
 
 /**
  * Updates the category colours and background colour of selected categories.
- * @package
+ * @protected
  */
 Blockly.Toolbox.prototype.refreshTheme = function() {
   for (var i = 0; i < this.contents_.length; i++) {
--- a/core/toolbox/toolbox.js
+++ b/core/toolbox/toolbox.js
@@ -349,7 +349,7 @@ Blockly.Toolbox.prototype.addToolboxItem_ = function(toolboxItem) {
 /**
  * Adds a style on the toolbox. Usually used to change the cursor.
  * @param {string} style The name of the class to add.
- * @package
+ * @protected
  */
 Blockly.Toolbox.prototype.addStyle = function(style) {
   Blockly.utils.dom.addClass(/** @type {!Element} */ (this.HtmlDiv), style);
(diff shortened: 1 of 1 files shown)


core/toolbox/toolbox.js
```diff
@@ -346,6 +346,202 @@ Blockly.Toolbox.prototype.addToolboxItem_ = function(toolboxItem) {
   }
 };
 
+/**
+ * Adds a style on the toolbox. Usually used to change the cursor.
+ * @param {string} style The name of the class to add.
+ * @protected
+ */
+Blockly.Toolbox.prototype.addStyle = function(style) {
```
samelhusseini, line 354: In the interest of simplicity, can we leave these two out. I don't see it adding any additional value to the interface since HtmlDiv is public, if we have a need to add styles to it, this can be done where-ever this is called. WDYT?

core/toolbox/toolbox.js
```diff
@@ -346,6 +346,202 @@ Blockly.Toolbox.prototype.addToolboxItem_ = function(toolboxItem) {
...
+ * Unhighlights any previously selected item.
+ * @public
+ */
+Blockly.Toolbox.prototype.clearSelection = function() {
+  this.setSelectedItem(null);
+};
+
+/**
+ * Updates the category colours and background colour of selected categories.
+ * @protected
```
samelhusseini, line 512: I think this is meant to be package.

core/toolbox/toolbox.js
```diff
@@ -346,6 +346,202 @@ Blockly.Toolbox.prototype.addToolboxItem_ = function(toolboxItem) {
...
+ */
+Blockly.Toolbox.prototype.addStyle = function(style) {
+  Blockly.utils.dom.addClass(/** @type {!Element} */ (this.HtmlDiv), style);
+};
+
+/**
+ * Removes a style from the toolbox. Usually used to change the cursor.
+ * @param {string} style The name of the class to remove.
+ * @protected
```
samelhusseini, line 361: If we do keep this, this should be package

core/toolbox/toolbox.js
alschmiedt: My main concern is that if we pull these out then developers lose the ability to easily change this functionality. I personally don't think it is worth the small cleanup, but let me know if I am missing something.

core/toolbox/toolbox.js
alschmiedt: done

core/toolbox/toolbox.js
alschmiedt: done

PR #4082, opened 2020-07-28
Add last accessed time to Xml model, and update it when getting from the db

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of data retention.

### Proposed Changes

- Add a `last_accessed` field to the `Xml` model
- Use `auto_now` to [make it update on every put](https://googleapis.dev/python/python-ndb/latest/model.html#google.cloud.ndb.model.DateTimeProperty)
- Update when accessed

### Reason for Changes

Data retention

### Test Coverage

Not yet tested. I think my best way to test would be to:
- deploy this on staging
- save some blocks
- inspect entities to find the correct one and confirm that the datetime is correct
- open the saved blocks from the link
- inspect entities to confirm that the last_accessed time has been updated

### Additional Information

I can also set the property to be required, but I'm not sure what value I would end up with. The documentation says "And unlike the legacy google.appengine.ext.db, auto_now does not supply a default value."

I will need to add a last+accessed time to all entities, but I will probably just do that in a separate batch job.

### DIFF
--- a/appengine/storage.py
+++ b/appengine/storage.py
@@ -23,6 +23,7 @@ __author__ = "q.neutron@gmail.com (Quynh Neutron)"
 import cgi
 import hashlib
 from random import randint
+from datetime import timezone
 from google.cloud import ndb
 
 
@@ -30,6 +31,7 @@ class Xml(ndb.Model):
   # A row in the database.
   xml_hash = ndb.IntegerProperty()
   xml_content = ndb.TextProperty()
+  last_accessed = ndb.DateTimeProperty(auto_now = true)
 
 
 def keyGen():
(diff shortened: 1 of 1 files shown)


appengine/storage.py
```diff
@@ -72,10 +74,19 @@ def keyToXml(key_provided):
...
     xml = ""
   else:
     xml = result.xml_content
+    # Update row to include last accessed time--because last_accessed is set to
+    # auto_now, any write updates the time.
+    with client.context():
+      row = Xml(id = key_provided,
+        xml_hash = result.xml_hash,
+        xml_content = result.xml_content)
+      row.put();
```
moniika, line 88: would it work if you just did `result.put()` (instead of creating/copying result)?

What's the benefit of `auto_now` vs explicitly updating the `last_accesed` time here? It seems more straightforward to me if we just updated `last_accessed`.

rachel-fenichel: Switched to just doing `result.put()`.

I like using `auto_now` because it ensures consistency in datetime formatting, timezones, etc.

rachel-fenichel: I tested locally with `dev_appserver.py`. I still haven't been able to view the datastore directly, but I verified that the last updated time is getting updated correctly on read.

PR #4089, opened 2020-07-29
Add separator for toolbox

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Adds a toolbox separator class.

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox/separator.js
+++ b/core/toolbox/separator.js
@@ -30,7 +30,7 @@ Blockly.ToolboxSeparator = function(toolboxSeparatorDef, toolbox) {
   Blockly.ToolboxSeparator.superClass_.constructor.call(
       this, toolboxSeparatorDef, toolbox);
   /**
-   * The config for all the separator classes.
+   * All the css class names that are used to create a separator.
    * @type {!Blockly.ToolboxSeparator.ClassConfig}
    * @protected
    */
@@ -44,6 +44,7 @@ Blockly.ToolboxSeparator = function(toolboxSeparatorDef, toolbox) {
 Blockly.utils.object.inherits(Blockly.ToolboxSeparator, Blockly.ToolboxItem);
 
 /**
+ * All the css class names that are used to create a separator.
  * @typedef {{
  *            container:?string,
  *          }}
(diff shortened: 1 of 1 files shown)


core/toolbox/separator.js
```diff
@@ -0,0 +1,75 @@
...
+Blockly.ToolboxSeparator = function(toolboxSeparatorDef, toolbox) {
+
+  Blockly.ToolboxSeparator.superClass_.constructor.call(
+      this, toolboxSeparatorDef, toolbox);
+  /**
+   * The config for all the separator classes.
+   * @type {!Blockly.ToolboxSeparator.ClassConfig}
+   * @protected
+   */
+  this.classConfig_ = {
```
moniika, line 37: I think some confusion I'm having is with usage of the term classes. Class could be either referring to JavaScript class or css class, after looking a bit longer it seems like this is about css. Maybe mentioning css in the JsDoc would make this clearer.

core/toolbox/separator.js
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  Blockly.utils.object.mixin(this.classConfig_,
+      toolboxSeparatorDef['classConfig']);
+};
+Blockly.utils.object.inherits(Blockly.ToolboxSeparator, Blockly.ToolboxItem);
+
+/**
+ * @typedef {{
+ *            container:?string,
```
moniika, line 48: Is there a way that you can document what the properties of the typedef mean? This by itself doesn't explain what it is, it only provides typings.

core/toolbox/separator.js
alschmiedt: I added what I think is a better comment, but would it also help to name these cssConfig_ instead of classConfig_?

core/toolbox/separator.js
alschmiedt: I don't know of a way to do this just yet. I have seen @property used in some places online, so I might be able to use that but will keep looking into it. If nothing else I'll add a line in the comment explaining what the container is.

core/toolbox/separator.js
moniika: I think `cssConfig_` could also help. I do like the jsdoc comment changes.

core/toolbox/separator.js
moniika: sgtm.

PR #4090, opened 2020-07-29
Adds create dom for the category

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Adds create dom methods for a category.

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox/category.js
+++ b/core/toolbox/category.js
@@ -165,6 +165,12 @@ Blockly.ToolboxCategory.ClassConfig;
 Blockly.ToolboxCategory.nestedPadding = 19;
 
 /**
+ * The width in pixels of the strip of colour next to each category.
+ * @type {number}
+ */
+Blockly.ToolboxCategory.borderWidth = 8;
+
+/**
  * Parses the contents array depending on if the category has children, is a
  * dynamic category, or if its contents are meant to be shown in the flyout.
  * @param {!Blockly.utils.toolbox.Category} categoryDef The information needed
@@ -243,6 +249,7 @@ Blockly.ToolboxCategory.prototype.createContainer_ = function() {
   Blockly.utils.dom.addClass(container, this.classConfig_['container']);
   return container;
 };
+
 /**
  * Creates the container for the icon and the label.
  * @return {!Element} The div that holds the icon and the label.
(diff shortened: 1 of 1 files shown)


core/toolbox/category.js
```diff
@@ -194,6 +194,147 @@ Blockly.ToolboxCategory.prototype.parseContents_ = function(categoryDef,
...
+      Blockly.utils.aria.State.SELECTED,false);
+  Blockly.utils.aria.setState(/** @type {!Element} */ (this.htmlDiv_),
+      Blockly.utils.aria.State.LEVEL, this.level_);
+  this.htmlDiv_.setAttribute('id', this.id_);
+
+  this.rowDiv_ = this.createRowContainer_();
+  this.htmlDiv_.appendChild(this.rowDiv_);
+
+  this.iconSpan_ = this.createIconSpan_();
+  Blockly.utils.aria.setRole(this.iconSpan_, Blockly.utils.aria.Role.PRESENTATION);
```
rachel-fenichel, line 213: Can you set roles inside your helper functions, at least for this, the label, and the group?

alschmiedt: I went back and forth with this a bit. By keeping these in the `createDom` method someone can overwrite the create methods without losing the accessibility labels. If we put them in their individual create methods we lose this, but increase the ability for the developer to create their own accessibility labels. Thoughts?

core/toolbox/category.js
```diff
@@ -194,6 +194,147 @@ Blockly.ToolboxCategory.prototype.parseContents_ = function(categoryDef,
...
+ * Creates the container that holds the row and any sub categories.
+ * @return {!Element} The div that holds the icon and the label.
+ * @protected
+ */
+Blockly.ToolboxCategory.prototype.createContainer_ = function() {
+  var container = document.createElement('div');
+  Blockly.utils.dom.addClass(container, this.classConfig_['container']);
+  return container;
+};
+/**
```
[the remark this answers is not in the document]
alschmiedt: fixed.

core/toolbox/category.js
```diff
@@ -194,6 +194,147 @@ Blockly.ToolboxCategory.prototype.parseContents_ = function(categoryDef,
...
+};
+
+/**
+ * Add the strip of colour to the toolbox category.
+ * @param {string} colour The category colour.
+ * @protected
+ */
+Blockly.ToolboxCategory.prototype.addColour_ = function(colour) {
+  if (colour) {
+    var border = '8px solid ' + (colour || '#ddd');
```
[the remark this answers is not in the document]
alschmiedt: Just updated.

core/toolbox/category.js
rachel-fenichel: Do we expect all changes to the create methods to end up needing the same accessibility labels?

I see your point, though. There's a balance between code structure and making sure that accessibility is the default, and I think the latter is more important here.

If the accessibility labels end up being more than a few lines, let's put them in helper methods going forward.

alschmiedt: Yeah, I think it depends a lot on the use case. They would definitely need to change the labels if they go away from the tree structure, but hard to say otherwise.

sgtm

Issue #4091, opened 2020-07-31
Error calling dispose on non-injected WorkspaceSvg


moniika:
**Describe the bug**

Cannot safely call `dispose` on a `Blockly.WorkspaceSvg` that has not been injected.

**To Reproduce**

1. Go to playground.
2. Input the following code into the console:
```
var ws = new Blockly.WorkspaceSvg(new Blockly.Options({}));
ws.dispose();
```
3. See error

**Expected behavior**

Should be able to call `dispose` without error.

**Stack Traces**

```
Uncaught TypeError: Cannot read property 'dispose' of null
    at Blockly.MarkerManager.dispose (marker_manager.js:179)
    at Blockly.WorkspaceSvg.dispose (workspace_svg.js:821)
    at <anonymous>:1:4
```

**Additional context**

Uncovered while writing tests.

Issue #4080, opened 2020-07-28
Workspace is not always disposed in mocha tests


moniika:
**Problem**

Some mocha tests make the incorrect assumption that the before/after of an outer test suite has access to objects defined on `this` from inner suites and due to this assumption, some tests do not dispose the created workspace.

Common example:
```
suite('Outer suite', function() {
  setup(function() {
    // Setup workspace that most tests use.
    this.workspace = new Workspace();
  });
  teardown(function() {
    this.workspace.dispose();
  });
  ...
  suite('Inner suite', function() {
    setup(function() {
      this.workspace.dispose(); // dispose of workspace because we're not using it
      // Create a new workspace and ASSUME it'll be cleaned up by outer suite's teardown but be WRONG
      this.workspace = new Workspace(some extra options);
    });
    ...
  });
});
```

The inner suite has a copy of the outer context, but when references are changed, then it is not reflected in the outer suite context (at teardown). In the above example, what actually happens is that the first workspace created in the outer suite is disposed of twice (once in inner suite setup, once in outer suite teardown) and the second workspace created in the inner suite is never disposed.

**Proposed solution**

Should review existing mocha tests and fix teardown so that workspaces all end up being disposed of properly. Potentially also investigate if there were any other bugs in tests due to this assumption.

**Additional context**

Chrome developer tools and breakpoints were used to confirm that this is the scenario and checking the mocha source code appears to support this as well.

moniika:
All tests addressed, except for `procedure_tests.js`

Issue #4070, opened 2020-07-24
Cleanup for asynchronous calls in Blockly tests


moniika:
### Problem

Logic in Blockly core triggers some methods asynchronously through `setTimeout` calls.
Many of our mocha tests trigger these asynchronous calls, which may end up being called after the test has completed resulting in issues such as:
- console warnings when tests are run in browser
- console errors when tests are run in browser
- unexpected behavior in unrelated tests, potentially causing false positives. For instance, in the case of events, addition events may be fired caused by previous tests (#4064)

Some approaches that we have tried have been to use a [sinon fake timer](https://github.com/sinonjs/fake-timers) or make use of mocha's built-in `done()` call. These approaches has been applied in some of the tests, but there are more tests that are still missing cleanup.

### Proposed solution

Since there are a lot of sources for asynchronous calls in Blockly, a shared cleanup method that any of the mocha tests can use would be ideal.

### Additional context

Upon investigation of sinon fake timer library, the `runAll` and `reset` methods (separate from the `restore` which only restores the faked methods)  are both options that we could use to clear the event queue (which as of now we have not yet taken advantage of).

moniika:
Cleanup for async calls now in all current mocha tests (through `sharedTestSetup` and `sharedTestTeardown` that use sinon fake timer).

PR #4092, opened 2020-08-01
Shared events cleanup

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4070
Part of #4064
Partially addresses #4080

### Proposed Changes

- Adds shared setup/cleanup to handle resolving all setTimeout calls triggered in tests in teardown, disposes of workspace, and resets global properties on Blockly.Events (`sharedTestSetup` `sharedTestTeardown`)
- Adds warnings in cleanup to help with detecting leakages from other tests
- Adds try/catch/finally in cleanup that logs an errors on Blockly code so that it does not prevent other tests in the suite from running and also still performs some of the cleanup needed.
- Moves logic for stubbing `Blockly.Events.fire` into shared setup
- Adds usages of `sharedTestSetup` `sharedTestTeardown` in tests that were stubbing/spying `Events.fire`

### Reason for Changes

- Shared code makes for easier cleanup
- Cleaning up async calls reduces possibilities of test failures that are solely due to bug in how test was written.
- Warnings help with detecting leakage from tests and are useful to log to console because they could potentially be causing an unexpected test failure.

### Test Coverage

Ran mocha tests

Tested on:
* Desktop Chrome

### Additional Information

This does not add cleanup to all mocha tests, additional PRs to follow.

Testing these changes revealed a bug with disposing of a non-injected instance of `WorkspaceSvg` #4091 and also revealed that there are leakages from other tests that make it so that we cannot call the exact implementation of `Blockly.Events.fire` (comment in code explaining issue) for these tests until the leakages are resolved.

### DIFF
--- a/tests/mocha/block_test.js
+++ b/tests/mocha/block_test.js
@@ -393,8 +393,6 @@ suite('Blocks', function() {
   });
   suite('Connection Tracking', function() {
     setup(function() {
-      this.workspace.dispose();
-      // The new rendered workspace will get disposed by the parent teardown.
       this.workspace = Blockly.inject('blocklyDiv');
 
       this.getInputs = function() {
@@ -420,6 +418,11 @@ suite('Blocks', function() {
         chai.assert.isEmpty(this.getNext());
         chai.assert.isEmpty(this.getPrevious());
       };
+    });
+    teardown(function() {
+      this.clock.runAll();
+      this.workspace.dispose();
+      this.clock.runAll();
     });
 
     suite('Deserialization', function() {
--- a/tests/mocha/block_test.js
+++ b/tests/mocha/block_test.js
@@ -420,9 +420,7 @@ suite('Blocks', function() {
       };
     });
     teardown(function() {
-      this.clock.runAll();
-      this.workspace.dispose();
-      this.clock.runAll();
+      workspaceTeardown.call(this, workspace);
     });
 
     suite('Deserialization', function() {
--- a/tests/mocha/block_test.js
+++ b/tests/mocha/block_test.js
@@ -44,7 +44,7 @@ suite('Blocks', function() {
       }]);
   });
   teardown(function() {
-    sharedTestCleanup.call(this);
+    sharedTestTeardown.call(this);
     delete Blockly.Blocks['empty_block'];
     delete Blockly.Blocks['stack_block'];
     delete Blockly.Blocks['row_block'];
@@ -420,7 +420,7 @@ suite('Blocks', function() {
       };
     });
     teardown(function() {
-      workspaceTeardown.call(this, workspace);
+      workspaceTeardown.call(this, this.workspace);
     });
 
     suite('Deserialization', function() {
@@ -1113,7 +1113,7 @@ suite('Blocks', function() {
         });
         test('Set While Visible - NonEditable', function() {
           this.block.setCommentText('test1');
-          // Restored up by call to sinon.restore() in sharedTestCleanup()
+          // Restored up by call to sinon.restore() in sharedTestTeardown()
           sinon.stub(this.block, 'isEditable').returns(false);
           var icon = this.block.getCommentIcon();
           icon.setVisible(true);
--- a/tests/mocha/block_test.js
+++ b/tests/mocha/block_test.js
@@ -6,6 +6,7 @@
 
 suite('Blocks', function() {
   setup(function() {
+    sharedTestSetup.call(this, {fireEventsNow: false});
     this.workspace = new Blockly.Workspace();
     Blockly.defineBlocksWithJsonArray([
       {
@@ -43,7 +44,7 @@ suite('Blocks', function() {
       }]);
   });
   teardown(function() {
-    this.workspace.dispose();
+    sharedTestCleanup.call(this);
     delete Blockly.Blocks['empty_block'];
     delete Blockly.Blocks['stack_block'];
     delete Blockly.Blocks['row_block'];
@@ -419,12 +420,8 @@ suite('Blocks', function() {
         chai.assert.isEmpty(this.getNext());
         chai.assert.isEmpty(this.getPrevious());
       };
+    });
 
-      this.clock = sinon.useFakeTimers();
-    });
-    teardown(function() {
-      this.clock.restore();
-    });
     suite('Deserialization', function() {
       test('Stack', function() {
         Blockly.Xml.appendDomToWorkspace(Blockly.Xml.textToDom(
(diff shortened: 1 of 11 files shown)


tests/mocha/test_helpers.js
```diff
@@ -81,6 +84,86 @@ function createEventsFireStub() {
...
+ * stubs. Should be called in setup of outermost suite using
+ * sharedTestSetup.call(this).
+ * @param {Object<string, boolean>} options Options to enable/disable setup
+ *    of certain stubs.
+ */
+function sharedTestSetup(options = {}) {
+  this.sharedSetupCalled_ = true;
+  // Sandbox created for greater control when certain stubs are cleared.
+  this.sharedSetupSandbox_ = sinon.createSandbox();
+  this.clock = this.sharedSetupSandbox_.useFakeTimers();
```
moniika, line 99: Creating this sandbox allows for other tests to use `sinon.restore()` and not affect the stubbing of the clock.
This is important for preventing leakage of `setTimeout` calls between tests.

tests/mocha/test_helpers.js
```diff
@@ -40,38 +40,41 @@ function captureWarnings(innerFunc) {
...
+    // TODO(#4070) Replace the fake function content with the following code
+    // that uses wrappedMethod after cleanup is added to ALL tests.
+    // Calling the original method will not consistently work if other tests
+    // add things to the event queue because the setTimeout call will not be
+    // added to the stubbed clock (so runAll cannot be used to control).
+    // // Call original method.
+    // stub.wrappedMethod.call(this, ...arguments);
+    // // Advance clock forward to run any queued events.
+    // clock.runAll();
+    //
```
moniika, line 77: I would like to note, that this was **"fun"** to debug.

tests/mocha/test_helpers.js
```diff
@@ -40,38 +40,41 @@ function captureWarnings(innerFunc) {
...
- * Creates stub for Blockly.utils.genUid that returns the provided id or ids.
- *    Recommended to also assert that the stub is called the expected number of
- *    times.
- * @param {string|!Array<string>} returnIds The return values to use for the
- *    created stub. If a single value is passed, then the stub always returns
- *    that value.
- * @return {!SinonStub} The created stub.
+ * Safely disposes of Blockly workspace, logging any errors.
+ * Assumes that sharedTestSetup has also been called. This should be called
+ * using workspaceTeardown.call(this).
```
moniika, line 45: Here, as well as in `sharedTestSetup` and `sharedTestTeardown`, I'm not sure of a better way to explain how the correct `this` should be bound to the method.

tests/mocha/test_helpers.js
```diff
@@ -81,6 +84,86 @@ function createEventsFireStub() {
...
+    Blockly.Events.setGroup(false);
+    Blockly.Events.disabled_ = 0;
+    if (Blockly.Events.FIRE_QUEUE_.length) {
+      // If this happens, it may mean that some previous test is missing cleanup
+      // (i.e. a previous test added an event to the queue on a timeout that
+      // did not use a stubbed clock).
+      Blockly.Events.FIRE_QUEUE_.length = 0;
+      console.warn(this.currentTest.fullTitle() +
+          '" needed cleanup of Blockly.Events.FIRE_QUEUE_. This may indicate ' +
+          'leakage from an earlier test');
```
moniika, line 137: This does currently cause warnings on the console, but I think they are valuable as it does indicate an actual issue that we should fix.

tests/mocha/test_helpers.js
```diff
@@ -81,6 +84,86 @@ function createEventsFireStub() {
...
+      workspaceTeardown.call(this, this.workspace);
+      this.workspace = null;
+    } else {
+      this.clock.runAll();  // Run all queued setTimeout calls.
+    }
+  } catch (e) {
+    console.error(this.currentTest.fullTitle() + '\n', e);
+  } finally {
+    // Clear Blockly.Event state.
+    Blockly.Events.setGroup(false);
```
rachel-fenichel, line 128: I think you should also warn if event groups are still set, or if events are disabled. Both are signs that something is not getting reset.

However, you can just file a bug to add that--adding it right now would probably pollute the console even more.

tests/mocha/test_helpers.js
```diff
@@ -81,6 +84,86 @@ function createEventsFireStub() {
...
+      // did not use a stubbed clock).
+      Blockly.Events.FIRE_QUEUE_.length = 0;
+      console.warn(this.currentTest.fullTitle() +
+          '" needed cleanup of Blockly.Events.FIRE_QUEUE_. This may indicate ' +
+          'leakage from an earlier test');
+    }
+
+    // Restore all stubbed methods.
+    this.sharedSetupSandbox_.restore();
+    sinon.restore();
```
rachel-fenichel, line 142: Why also call `sinon.restore`? It looks like everything done in this shared setup used the sandbox.

tests/mocha/block_test.js
```diff
@@ -433,7 +431,7 @@ suite('Blocks', function() {
             '</xml>'
         ), this.workspace);
         this.assertConnectionsEmpty();
-        this.clock.tick(1);
```
rachel-fenichel, line 436: Another one to file as a bug for future cleanup (and not handle in this PR): these calls are a good candidate for a helper function, which:
- takes in the text
- does the text to dom call
- appends dom to workspace
- asserts connections empty
- calls runAll

tests/mocha/test_helpers.js
moniika: The only stub on `sharedSetupSandbox_` is `this.clock`. All the stubs defined in the tests define themselves on `sinon` directly, which is why I need to also do `sinon.restore`. I have the clock on a separate sandbox so that if anywhere else `sinon.restore` is called, it doesn't affect the stubbed clock.

rachel-fenichel: Ah, I see--the events stub is not on the sandbox.

tests/mocha/block_test.js
moniika: Filed #4101

tests/mocha/test_helpers.js
moniika: That, and also other stubs and spies in individual tests.

tests/mocha/test_helpers.js
moniika: Filed #4102

tests/mocha/test_helpers.js
rachel-fenichel: Right--the stubs and spies in individual tests are an argument for calling restore in the same place where they set up the stub or spy, though. My concern is around the asymmetry, where we can sometimes expect to be cleaning up state in this teardown that the matching setup function did not create.

rachel-fenichel:
LGTM

PR #4096, opened 2020-08-03
Add option sub types

samelhusseini:
## The basics
- [x] I branched from develop

### Proposed Changes

Add typedefs for MoveOptions, GridOptions and ZoomOptions.

### Reason for Changes

Strict types

### Test Coverage

Tested with build.

Tested on:

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -134,7 +134,7 @@ Blockly.Options = function(options) {
   this.pathToMedia = pathToMedia;
   /** @type {boolean} */
   this.hasCategories = hasCategories;
-  /** @type {!Object} */
+  /** @type {!Blockly.Options.MoveOptions} */
   this.moveOptions = Blockly.Options.parseMoveOptions(options, hasCategories);
   /** @deprecated  January 2019 */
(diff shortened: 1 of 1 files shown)


core/options.js
```diff
@@ -213,7 +248,7 @@ Blockly.Options.prototype.getMetrics;
  *    behaviour is unspecified.
  * @param {!Object} options Dictionary of options.
  * @param {boolean} hasCategories Whether the workspace has categories or not.
- * @return {!Object} A dictionary of normalized options.
+ * @return {!Blockly.Options.MoveOptions} Normalized move options.
  * @private
  */
 Blockly.Options.parseMoveOptions = function(options, hasCategories) {
```
alschmiedt, line 254: Slightly unrelated, but should be Blockly.Options.parseMoveOptions_

samelhusseini: I saw that and was tempted to change, but I imagined there was a good reason to not rename.
I don't imagine developers using this API call, so I'm going to rename in this PR.

samelhusseini: Marking as api_change just in case.

PR #4100, opened 2020-08-03
Cleanup of chai asserts

moniika:
## The basics

- [x] I branched from develop

### Resolves

None

### Proposed Changes

Fixes asserts in tests to follow the form: (actual, expected) so that error messages are correct.

### Reason for Changes

This fixes test assert messages so the values aren't swapped, making for a better debugging experience.

### Test Coverage

Ran tests.

Tested on:
* Desktop Chrome

### Additional Information

Files updated using the regex:
Find:
`chai.assert.equal\((\d+), ([^\s\)]+)\);`
Replace:
`chai.assert.equal\($2, $1\);`

### DIFF
--- a/tests/mocha/event_test.js
+++ b/tests/mocha/event_test.js
@@ -469,7 +469,7 @@ suite('Events', function() {
         new Blockly.Events.Ui(block, 'click', undefined, undefined)
       ];
       var filteredEvents = Blockly.Events.filter(events, true);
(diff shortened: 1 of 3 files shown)


moniika:
> LGTM. I assume the reason that they're swapped is because the jsunit helpers had them in the opposite order.

I don't believe so. It's a common error as its hard to catch (until a test fails and the error message is only slightly wrong).

PR #4103, opened 2020-08-04
Adding async cleanup to tests

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4070
Part of #4064

### Proposed Changes

- Adds calls for `sharedTestSetup` and `sharedTestTeardown` to handle running all async Blockly code in teardown
- Updates calls in setup/teardown for disposing of workspace to use `workspaceTeardown` in order to handle runnning any async tasks and catching if any error (so remaining tests still run).

### Reason for Changes

- Preventing leakage between tests.

### Test Coverage

Ran mocha tests

Tested on:
* Desktop Chrome

### Additional Information

`navigation_modify_test.js`, `insertion_marker_test.js` and `theme_test.js` not included in this PR. To be addressed in follow up as there were errors that need to be investigated when shared setup/teardown was tried.

### DIFF
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -29,7 +29,8 @@
     <script src="https://unpkg.com/@blockly/dev-tools/dist/index.js"></script>
     <script>
       mocha.setup({
-        ui: 'tdd'
+        ui: 'tdd',
+        timeout: 1000000
       });
     </script>
 
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -29,8 +29,7 @@
     <script src="https://unpkg.com/@blockly/dev-tools/dist/index.js"></script>
     <script>
       mocha.setup({
-        ui: 'tdd',
-        timeout: 1000000
+        ui: 'tdd'
       });
     </script>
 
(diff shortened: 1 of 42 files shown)


tests/mocha/index.html
```diff
@@ -29,7 +29,8 @@
     <script src="https://unpkg.com/@blockly/dev-tools/dist/index.js"></script>
     <script>
       mocha.setup({
-        ui: 'tdd'
+        ui: 'tdd',
+        timeout: 1000000
```
alschmiedt, line 33: just curious what this timeout is for?

tests/mocha/navigation_test.js
```diff
@@ -50,8 +57,8 @@ suite('Navigation', function() {
     });
 
     teardown(function() {
+      workspaceTeardown.call(this, this.workspace);
```
alschmiedt, line 60: Could this also go in the parent teardown function?

tests/mocha/index.html
moniika: for debugging, should be removed

tests/mocha/navigation_test.js
moniika: Can't be moved, parent teardown doesn't have access to `this.workspace`. Could be moved if `this.workspace` was re-defined in outer scope, but there's a lot of other similar instances and I'm not yet convinced that's the best plan. See #4080 for more explanation (if I do move things, I'd probably do it as part of addressing that bug).

PR #4109, opened 2020-08-05
Apply dependabot change to develop

rachel-fenichel:
See #4054


rachel-fenichel:
No--I think because lodash is a transitive dependency, not directly mentioned in package.json.

Issue #4105, opened 2020-08-04
FieldNumber does not respect precision for small precision values


mfallihee:
When instantiating a FieldNumber with a very small precision value, they are handled incorrectly.

Repro:
1. Create FieldNumber with precision of a very small number value (e.g. .0000001, 1e-8, etc)

Expected Behavior:
The field respects the precision and rounds values appropriately

Actual Behavior:
The precision value gets converted to exponential notation (e.g. 1e-8) and when trying to parse the number of decimal places in Blockly.FieldNumber.prototype.setPrecisionInternal_ it ends up with decimalPlaces_ = 0. This causes all numbers to be rounded to integer values.

~~This is easily worked around by using a string representation for the precision (e.g. ".0000001")~~ but this used to work in older versions and is a regression. 

Edit: Upon further testing, a string representation is not a workaround and fails for the same reason.

mfallihee:
My suggestion is to switch from using Number.prototype.toString() to do the parsing and instead use Number.prototype.toLocaleString()
```
precision.toLocaleString("en-US", { maximumFractionDigits: 20 });
```
This should always give a consistent result when trying to count decimal places. 

I'm happy to submit a PR for this.

moniika:
Definitely is a bug, thanks for catching it! I can't think of a reason off the top of my head why `toLocaleString` wouldn't be an appropriate fix. We would definitely welcome a PR. I've also added a test case that catches this bug.

PR #4124, opened 2020-08-06
Use toLocaleString to count decimal places consistently

mfallihee:
## The basics

- [x] I branched from develop

### Resolves

Fixes issue #4105 

### Proposed Changes

Use toLocaleString when counting decimal places for consistent results around small values. Numbers like 1e-7 automatically get converted to exponential notation with .toString()

### Test Coverage

Test coverage added in separate PR #4115 

Tested on:
* Desktop Chrome 
* Desktop Firefox 
* Desktop Safari 

### Additional Info
There is potentially issues with IE 10 support here since the "locale" and "options" arguments for .toLocaleString are not supported for IE 10. I don't have an easy way to test IE 10 at the moment.

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -231,7 +231,7 @@ Blockly.FieldNumber.prototype.setPrecisionInternal_ = function(precision) {
     }
   }
 
-  var precisionString = this.precision_.toString();
+  var precisionString = this.precision_.toLocaleString("en-US", {maximumFractionDigits: 20});
   var decimalIndex = precisionString.indexOf('.');
   if (decimalIndex == -1) {
     // If the precision is 0 (float) allow any number of decimals,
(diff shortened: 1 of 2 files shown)


moniika:
Thanks for creating the PR.
[CanIUse](https://caniuse.com/#search=toLocaleString) seems to show that IE11 is supported. We aren't supporting IE10, so that's fine.
Only thing missing is now that #4115 is merged, you can remove "skip: true" and the TODO comment as well as signing the CLA.

mfallihee:
@googlebot I signed it!

mfallihee:
@moniika I have updated the PR to unskip the test and signed the cla. 
Good to hear about IE 10 support. The [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide) still references IE 10 support, not sure if there is an issue tracking cleaning up IE 10 references.

rachel-fenichel:
I updated the style guide to say IE11 instead of IE 10

NeilFraser:
FYI, this change results in a 4.6% slowdown of Blockly (as measured when loading the Spaghetti test in the playground).  PR #4870 corrects this.

PR #4144, opened 2020-08-10
Toolbox Events

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Before we were using Blockly.Events.Ui with an element type of 'category'. Since we are no longer assuming that all items in a toolbox are categories it would make sense to update element to be the type of the toolbox item. However, if we do this there is not a great way for someone to listen to only toolbox change events. 

### Proposed Changes
Add a ToolboxChange event. 
This would have the newId and oldId of the toolbox item as well as the new and old type. We would no longer supply the name of the toolbox item since a developer can get the name of the toolbox item by using toolbox.getToolboxItemById(item_id). 

### Test Coverage

Tested on:


core/toolbox/toolbox_events.js
```diff
@@ -0,0 +1,99 @@
...
+ * Used to notify the developer when a user has clicked on a new toolbox item.
+ * @param {?string} oldType The type of the previously selected toolbox item.
+ * @param {?string} oldId The id of the previously selected toolbox item.
+ * @param {?string} newType The type of the currently selected toolbox item.
+ * @param {?string} newId The id of the currently selected toolbox item.
+ * @param {Blockly.Workspace} workspace The workspace the toolbox is on.
+ * @extends {Blockly.Events.Abstract}
+ * @constructor
+ */
+Blockly.Events.ToolboxChange = function(oldType, oldId, newType, newId, workspace) {
```
samelhusseini, line 32: Are there going to be other toolbox events. It looks like this is more accurately a `ToolboxCategoryChange` event?

core/toolbox/toolbox_events.js
```diff
@@ -0,0 +1,99 @@
...
+  /**
```
samelhusseini, line 33: Will also have to add a way to determine that the block is indeed blank, checking that the first parameter is undefined should suffice: 
```
  this.isBlank = typeof opt_oldId == 'undefined';
```

Could also consider passing in the toolbox as the first parameter since this is a toolbox event.

core/toolbox/toolbox_events.js
```diff
@@ -0,0 +1,99 @@
...
+goog.require('Blockly.Events');
+goog.require('Blockly.Events.Abstract');
+goog.require('Blockly.registry');
+goog.require('Blockly.utils.object');
+
+
+/**
+ * Class for a toolbox changed event.
+ * Used to notify the developer when a user has clicked on a new toolbox item.
+ * @param {?string} oldType The type of the previously selected toolbox item.
```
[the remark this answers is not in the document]
alschmiedt: done

core/toolbox/toolbox_events.js
alschmiedt: Updated to ToolboxItemChange since they aren't guaranteed to be categories.

core/toolbox/toolbox_events.js
alschmiedt: done and also passed in the toolbox.

alschmiedt:
Closing

PR #4158, opened 2020-08-11
Hide category

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Adds ability to hide and disable a category. 

### Proposed Changes
Add isHidden_ and isDisabled properties.

### Test Coverage

Tested on:

### Additional Information
If isHidden is set to true on the currently selected item, this relies on the developer to call toolbox.clearSelection() to close the flyout.

### DIFF
--- a/core/toolbox/category.js
+++ b/core/toolbox/category.js
@@ -513,16 +513,31 @@ Blockly.ToolboxCategory.prototype.hasChildren = function() {
  * Sets whether the category is visible or not.
  * For a category to be visible its parent category must also be expanded.
  * @param {boolean} isVisible True if category should be visible.
- * @public
+ * @private
  */
-Blockly.ToolboxCategory.prototype.setVisible = function(isVisible) {
+Blockly.ToolboxCategory.prototype.setVisible_ = function(isVisible) {
   this.htmlDiv_.style.display = isVisible ? 'block' : 'none';
   if (this.hasChildren()) {
     for (var i = 0, child; (child = this.contents_[i]); i++) {
-      child.setVisible(isVisible);
+      child.setVisible_(isVisible);
     }
   }
   this.isHidden_ = !isVisible;
+};
+
+/**
+ * Hide the category.
+ */
+Blockly.ToolboxCategory.prototype.hide = function() {
+  this.setVisible_(false);
+};
+
+/**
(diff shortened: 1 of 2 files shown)


core/toolbox/category.js
```diff
@@ -126,11 +126,27 @@ Blockly.ToolboxCategory = function(categoryDef, toolbox, opt_parent) {
...
+   * True if the parent category is expanded, false otherwise.
+   * Children categories can only be visible if their parent category is
+   * expanded.
+   * @type {boolean}
+   * @private
+   */
+  this.isParentExpanded_ = true;
+
+  /**
+   * True if this category is not disabled, false otherwise.
```
maribethb, line 145: -not

core/toolbox/category.js
```diff
@@ -494,24 +510,44 @@ Blockly.ToolboxCategory.prototype.hasChildren = function() {
...
+  if (this.hasChildren()) {
+    for (var i = 0, child; (child = this.contents_[i]); i++) {
+      child.setVisible_(isVisible);
+    }
+  }
+  this.isHidden_ = !isVisible;
+};
+
+/**
+ * Hide the category.
```
maribethb, line 529: Question: In the PR description you said in certain cases the developer would need to call toolbox.clearSelection() to close the flyout; is that the case when they call this hide function on the currently selected item? If so, maybe add that instruction here so they know they need to call it?

Or is it possible to detect that this is being called on the selected item and call it for them? I wasn't sure if that was a requirement for them to call the function or if you were clarifying that if they wanted the flyout to close, they would still have to do it themselves.

alschmiedt: So, it doesn't technically break anything but it is definitely a bit weird to have the flyout open when the category has been hidden. 

We could do something like: 
```
if (this.parentToolbox_.getSelectedItem() == this) {
   this.parentToolbox_.clearSelection();
}
```
It seemed a bit messy and wasn't sure if we wanted to leave this up to the developer. But thinking about it more I can't think of a situation where a developer wouldn't want to do this. So I think I will add back in.

core/toolbox/category.js
alschmiedt: done

PR #4161, opened 2020-08-12
Small fixes

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
1. Allows developer to set hidden="true" on categories
2. Adds fix for the registering a toolbox item in the registry. We usually throw an error when someone registers something that already exists, however this is expected in the case of toolbox_items since we expect users to register different categories and separators.
3. Stores the blockxml when we convert it from a string so we don't have to call textToDom every time the flyout opens.

### Test Coverage

Tested on:

### DIFF
--- a/core/registry.js
+++ b/core/registry.js
@@ -126,8 +126,9 @@ Blockly.registry.register = function(type, name, registryItem) {
   // Validate that the given class has all the required properties.
   Blockly.registry.validate_(type, registryItem);
 
-  // If the name already exists throw an error.
-  if (typeRegistry[name]) {
+  // Don't throw an error if it is a toolbox item. We expect developers to override
+  // categories and separators.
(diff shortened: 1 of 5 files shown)


core/registry.js
```diff
@@ -126,8 +126,9 @@ Blockly.registry.register = function(type, name, registryItem) {
   // Validate that the given class has all the required properties.
   Blockly.registry.validate_(type, registryItem);
 
-  // If the name already exists throw an error.
-  if (typeRegistry[name]) {
+  // Don't throw an error if it is a toolbox item. We expect developers to override
+  // categories and separators.
+  if (type != Blockly.registry.Type.TOOLBOX_ITEM && typeRegistry[name]) {
```
rachel-fenichel, line 131: This seems reasonable, but if it ends up being a pattern we should find a good way to distinguish between things we expect to be overridden vs things that should be unique. (That is, if this turns into a list of types that can be overridden, let's find a better structure for the check).

PR #4170, opened 2020-08-13
Add checks to WorkspaceSvg dispose and fix shared Workspace tests

### LINKED ISSUE #4091: Error calling dispose on non-injected WorkspaceSvg

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #4091
Part of #4080

### Proposed Changes

- Adds checks that prevent errors when `dispose` is called on `WorkspaceSvg` before dom is created
- Adds test for `dispose` call on `WorkspaceSvg` without `inject` or `createDom` call
- Changes workspace `dispose` calls to use `workspaceTeardown`
- Adds checks in shared workspace test to create `WorkspaceSvg` if main workspace is rendered

### Reason for Changes

- Preventing errors in `dispose` call seems reasonable and cleans up errors logged by tests
- Using `workspaceTeardown` instead of calling dispose directly prevents possibility of `dispose` call errors masking errors from test assertions
- Updating shared tests ensures that we are testing `WorkspaceSvg` when `testAWorkspace` is called for `WorkspaceSvg` tests

### Test Coverage

Ran mocha tests

Tested on:
 * Desktop Chrome

### Additional Information

Shared test cleanup depends on fix for `WorkspaceSvg` `dispose` call because test "Workspace base class getById after dispose" would otherwise fail.


tests/mocha/workspace_test.js
```diff
@@ -247,7 +247,9 @@ function testAWorkspace() {
 
   suite('addTopBlock', function() {
     setup(function() {
-      this.targetWorkspace = new Blockly.Workspace();
+      this.targetWorkspace = this.workspace.rendered ?
```
rachel-fenichel, line 250: When would the workspace not be a workspace svg, if it's a flyout?

moniika: I suppose in headless it might not make sense to test flyout. Is that the case?
I could instead:

- Move this to WorkspaceSvg tests 
- OR do something like:

```
this.targetWorkspace = this.workspace.rendered ? this.workspace : new Blockly.Workspace();
if (this.workspace.rendered) {
  this.workspace.isFlyout = true;
  this.workspace.targetWorkspace = this.targetWorkspace;
}
```

rachel-fenichel: Moving it to workspace_svg sounds most reasonable to me.

moniika: Done.

Issue #4178, opened 2020-08-14
Dropdown inputs suddenly getting a 'black' background after value changes


joseluiz98:
**Describe the bug**

Currently, dropdown fields are getting a 'black' background after certain value changes, I've noted that some default blocks have this behavior either, e.g 'lists' blocks.
I think it got something to do with the mutator of the block, cause if the option selected doesn't result in a major change on the block shape (e.g another input being added), this background 'lost' doesn't occur.

I've made a gif to show it, see how the field keep it's color on the first change, but on the second one he turns on a black background with no explanation but the input appearing.

https://media.giphy.com/media/YlYIhj6p3Kk6TNCsYk/giphy.gif

A discussion about this was opened on our developers forum's, and a user (rajuga) has a possible solution to this problem, i'm quoting he below:

'I also got this issue , basically they forgot to update the color,  in  Blockly.Blocks.lists_getIndex() , updateAt function , there we have to add one more exxtra line of code "block.setStyle('list_blocks');"'

Our discussion: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/blockly/c6SpqvTKflM/ZUOfrr2rAAAJ

**To Reproduce**

1. Add the 'lists_setIndex' block to your blockly clean installation.
2. Drag this block to your workspace
3- On the '#' (index) dropdown, select 'random' option.
4- Note that this dropdown you've just changed now have a 'black' background, and one input disappeared from the block end (a correct behavior if analyzed separately, but maybe a trigger for this problem?)

**Expected behavior**

I expect that the dropdown maintain its background according to the block style, regardless of the value selected or any major changes on the block.

**Screenshots**

I have made a gif to exemplify this, you can find out on the below link:
https://www.google.com/url?q=https%3A%2F%2Fmedia.giphy.com%2Fmedia%2FYlYIhj6p3Kk6TNCsYk%2Fgiphy.gif&sa=D&sntz=1&usg=AFQjCNFGibiKVGAb8wjRSsQGYIBuYLbbSQ

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.1 LTS
 - Browser Chrome
 - Version 84.0.4147.125 (official version) 64 bits

samelhusseini:
Fixed.

PR #4183, opened 2020-08-17
Call apply colour when inserting field into rendered block

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/4178

### Proposed Changes

Call applyColour when inserting a field to a block that has already been rendered.

### Reason for Changes

Bug fix.

### Test Coverage

Tested with the get_list block.

Tested on:

### DIFF
--- a/core/input.js
+++ b/core/input.js
@@ -105,6 +105,7 @@ Blockly.Input.prototype.insertFieldAt = function(index, field, opt_name) {
   field.setSourceBlock(this.sourceBlock_);
   if (this.sourceBlock_.rendered) {
     field.init();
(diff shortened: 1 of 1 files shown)


samelhusseini:
Yeah, it's a separate code path: 
- Normally `applyColour` is called: render -> initSvg -> field.init / field.initView, and applyColour
- When a field is added dynamically, we go through the above code path. If we were to add `applyColour` to the field `init` path, we'll be calling applyColour multiple times.
Worth keeping that in mind if we want to clean up our field applyColour lifecycle.

Issue #3899, opened 2020-05-13
Data retention enhancements needed for 3 blockly demos (code, cloud storage and block)


mmathrani:
**Is your feature request related to a problem? Please describe.**
Today three demo's (code, cloud storage and block) provide an option to store user created samples. These samples are stored in appengine datastore. They are not tied to any user and the link to the sampleas can be shared between users. We have no defined data retention policy around this and we need to build that to adhere to company privacy guidelines.

**Describe the solution you'd like**
1. Add a timestamp to the appengine data store that stores this data in XML format. This will help us delete the records after a specific time interval (currently planned for 1 year).
2. Create a cron job that will run once a week and clear records > time interval
3. Update the message on 'Save' to be 'Share your blocks with the below public link. We\'ll store them for a year. Please be sure not to include any private information.' from 'Share your blocks with the below public link.'
4. Backfill all existing records with today (the day we release) timestamp.
5. Provide a note in FAQ or help section to allow deletion of any saved sample before the data retention cron job deletes it. This will send an email to the core Google group who can do the needful.

NeilFraser:
They tried to get us to do expiry in Blockly Games too.  I considered this utterly unacceptable (there are lesson plans, [bug demos](https://github.com/google/blockly/issues/216), official [example links](https://blockly-demo.appspot.com/static/demos/code/index.html?lang=tlh#ortpyd), and the entirety of the block factory data).  After a closer look they backed down on the basis that this data was not tied to any individual user.

Push back.  Don't break a decade's worth of material.

mmathrani:
Based on internal discussions within the team, there is a change in direction wrt to the logic of deletion of developer created samples.
Point 1 from the PR description
"Add a timestamp to the appengine data store that stores this data in XML format. This will help us delete the records after a specific time interval (currently planned for 1 year)."
**changed to ** 
"Add a timestamp to the appengine data store that stores the last access time. This will help us delete the records after a specific time interval (currently planned for 1 year). The get/setter for this last_access_time will be updated from the code that loads/saves the three samples"

rachel-fenichel:
Done.

PR #4191, opened 2020-08-18
Add expiration handler

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #3899 

### Proposed Changes

Add an expiration handler that gets keys for expired items (last accessed more than a year ago) and deletes them.

I'm going to set up a cron job to hit this endpoint. But we don't need to start running that yet, since nothing expires for at least a year from now.

### Reason for Changes

Data retention.

### Test Coverage
Tested locally:
- dev_appserver
- local datastore
- set retention window to a few hours instead of a year

### Documentation

None

### Additional Information

Remaining work:
- handler to add dates to all existing entries
- cron.yaml

### DIFF
--- a/appengine/expiration.py
+++ b/appengine/expiration.py
@@ -25,11 +25,13 @@ import datetime
 
 from google.cloud import ndb
 
+
+EXPIRATION_DAYS = 365
+# Limit the query to avoid timeouts.
+QUERY_LIMIT = 1000
+
 def delete_expired():
-  # Delete entries that have not been accessed in more than a year.
-  EXPIRATION_DAYS = 365
-  # Limit the query to avoid timeouts.
-  QUERY_LIMIT = 1000
+  """Deletes entries that have not been accessed in more than a year."""
   bestBefore = datetime.datetime.utcnow() - datetime.timedelta(days=EXPIRATION_DAYS)
   client = ndb.Client()
   with client.context():
--- a/appengine/expiration.py
+++ b/appengine/expiration.py
@@ -1,0 +1,49 @@
+"""
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Delete expired XML.
+"""
+
+__author__ = "fenichel@google.com (Rachel Fenichel)"
+
+
+import storage
+import datetime
+
+from google.cloud import ndb
+
+def delete_expired():
+  # Delete entries that have not been accessed in more than a year.
+  EXPIRATION_DAYS = 365
+  # Limit the query to avoid timeouts.
+  QUERY_LIMIT = 1000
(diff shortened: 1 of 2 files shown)


appengine/expiration.py
```diff
@@ -0,0 +1,49 @@
...
+__author__ = "fenichel@google.com (Rachel Fenichel)"
+
+
+import storage
+import datetime
+
+from google.cloud import ndb
+
+def delete_expired():
+  # Delete entries that have not been accessed in more than a year.
```
moniika, line 29: This seems like it's meant to be a method comment. In which case, you should use the format for docstring:
`"""Delete entries that have not been accessed in more than a year."""`

Related nit: Deletes instead of Delete (coming from style guide for wording)

appengine/expiration.py
```diff
@@ -0,0 +1,49 @@
...
+  EXPIRATION_DAYS = 365
+  # Limit the query to avoid timeouts.
+  QUERY_LIMIT = 1000
+  bestBefore = datetime.datetime.utcnow() - datetime.timedelta(days=EXPIRATION_DAYS)
+  client = ndb.Client()
+  with client.context():
+    query = storage.Xml.query(storage.Xml.last_accessed < bestBefore)
+    results = query.fetch(limit=QUERY_LIMIT, keys_only=True)
+    for x in results:
+      x.delete()
```
moniika, line 39: Purely a suggestion, that I don't know if we want.
We could add logging here, which would give us an idea of how often/when things are deleted.
```
import logging
...
logging.info('Cleaned up entry with key :%s', x.urlsafe())
```

appengine/expiration.py
```diff
@@ -0,0 +1,49 @@
...
+import datetime
+
+from google.cloud import ndb
+
+def delete_expired():
+  # Delete entries that have not been accessed in more than a year.
+  EXPIRATION_DAYS = 365
+  # Limit the query to avoid timeouts.
+  QUERY_LIMIT = 1000
+  bestBefore = datetime.datetime.utcnow() - datetime.timedelta(days=EXPIRATION_DAYS)
```
moniika, line 33: It isn't necessary, but we could also add an expiration time on the entities as a computed property and then you could query for expiry time. 
```
 expiry_time = ndb.ComputedProperty(
      lambda self: self.last_accesssed + datetime.timedelta(days=EXPIRATION_DAYS))
```
Note: ComputedProperty is computed on `put()`, similar to what needed to be done with adding `last_accessed`, if we added the property we would need `put()` the entities before we could query them by that property.

appengine/expiration.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+import storage
+import datetime
+
+from google.cloud import ndb
+
+def delete_expired():
+  # Delete entries that have not been accessed in more than a year.
+  EXPIRATION_DAYS = 365
```
samelhusseini, line 30: nit: move constants out of method for clarity.

rachel-fenichel: Done.

appengine/expiration.py
rachel-fenichel: Done.

appengine/expiration.py
rachel-fenichel: Not necessary--it doesn't change what we're doing but is slightly more complicated.

appengine/expiration.py
rachel-fenichel: I don't think it's worth the logging here either. I expect that this cron job will delete a lot of entries at the beginning, and then we'll take a look at the number of remaining entities and make decisions from there.

samelhusseini:
Test failure is intermittent.

PR #4193, opened 2020-08-19
Bug Fixes for Toolbox

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
1. Allows uppercase and lowercase versions of cssConfig and categoryStyle
2. Fixes hovering for categories.

### Proposed Changes
Change the div that handles a click to be the blocklyTreeRow div instead of blocklyToolboxCategory div. The blocklyToolboxCategory div held all of its children, so when there was a nested category and we hovered over the parent category all the children would appear hovered.

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox/category.js
+++ b/core/toolbox/category.js
@@ -121,7 +121,7 @@ Blockly.ToolboxCategory = function(categoryDef, toolbox, opt_parent) {
     'closedIcon': 'blocklyTreeIconClosed',
   };
 
-  var cssConfig = categoryDef['cssConfig'] || categoryDef['cssconfig'];
+  var cssConfig = categoryDef['cssconfig'] || categoryDef['cssConfig'];
   Blockly.utils.object.mixin(this.cssConfig_, cssConfig);
 
   /**
--- a/core/toolbox/category.js
+++ b/core/toolbox/category.js
@@ -258,6 +258,7 @@ Blockly.ToolboxCategory.prototype.createDom = function() {
   this.htmlDiv_.appendChild(this.rowDiv_);
 
   this.rowContents_ = this.createRowContentsContainer_();
+  this.rowContents_.style.pointerEvents = 'none';
   this.rowDiv_.appendChild(this.rowContents_);
 
   this.iconDom_ = this.createIconDom_();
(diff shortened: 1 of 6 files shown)


core/toolbox/category.js
```diff
@@ -108,15 +115,17 @@ Blockly.ToolboxCategory = function(categoryDef, toolbox, opt_parent) {
...
     'label': 'blocklyTreeLabel',
     'contents': 'blocklyToolboxContents',
     'selected': 'blocklyTreeSelected',
     'openIcon': 'blocklyTreeIconOpen',
-    'closedIcon': 'blocklyTreeIconClosed'
+    'closedIcon': 'blocklyTreeIconClosed',
   };
 
-  Blockly.utils.object.mixin(this.cssConfig_, categoryDef['cssConfig']);
+  var cssConfig = categoryDef['cssConfig'] || categoryDef['cssconfig'];
```
samelhusseini, line 127: Curious why this is needed?

core/toolbox/category.js
```diff
@@ -389,7 +415,7 @@ Blockly.ToolboxCategory.prototype.addColourBorder_ = function(colour) {
  * @protected
  */
 Blockly.ToolboxCategory.prototype.getColour_ = function(categoryDef) {
-  var styleName = categoryDef['categorystyle'];
+  var styleName = categoryDef['categorystyle'] || categoryDef['categoryStyle'];
```
samelhusseini, line 418: Why is the order of check different here than above? lowercase then uppercase, vs uppercase then lowercase.

alschmiedt: No good reason. Changed them all to be the same.

core/toolbox/category.js
alschmiedt: It's not needed just nice to have. We currently have json and xml for the toolbox definitions and so all of our definitions are lowercase, however I have accidentally done camel case a lot in the JSON when testing and figure developers will as well. So just adding it as an option.

samelhusseini: Is this something that can be normalized in the xml to json toolbox method? That way we're not checking for both cases all over the place?

alschmiedt: Not the way it currently is. We only go through the given toolbox definition if it is not the correct format (ie xml, Array<Node>, string). If it is JSON we just return it. I could create a validate method to go through all of the JSON to validate it but I am unsure if this is worth it or not.

samelhusseini: Ok. Good to keep that option available, but yeah not worth it at this point.

Issue #4194, opened 2020-08-19
Cleanup warnings and errors logged in mocha tests


moniika:
Mocha tests trigger a variety of warnings and errors.
These warnings and errors logs are caused by a variety of things:
- expected warnings (like deprecation warnings)
- missing cleanup in tests or improper setup of objects used in tests
- bugs that should be filed

We should cleanup the tests so it is easier to use the console when debugging mocha tests and so that any warnings or errors we see in the console when running mocha tests are more meaningful.

moniika:
Remaining logs/warnings triggered by tests:
- [ ] `Comment.prototype.getText` deprecation warning - #4186
- [ ] `Field.prototype.classValidator` deprecation warning - #4197
- [ ] Not expected to reach Block.bumpNeighbours function. BlockSvg.bumpNeighbours was expected to be called instead. - from `logic_ternary_test.js` triggered because test runs on headless.
- [ ] Saw an existing group while responding to a definition change - `from procedures_test.js`

moniika:
- [ ] More recent regression for generator mocha tests emitting warning during `blockToCode` suite is tracked in https://github.com/google/blockly/issues/5029

moniika:
Closing this bug and tracking warning triggered by procedures in https://github.com/google/blockly/issues/5530

PR #4208, opened 2020-08-20
Capture warning logged by calling convertToolboxToJSON in tests.

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4194 

### Proposed Changes

Captures warning logged by calling `convertToolboxToJSON` in toolbox tests.

### Reason for Changes

Cleans up console when running tests.

### Test Coverage

Ran mocha tests.

Tested on:
 * Desktop Chrome

### DIFF
--- a/tests/mocha/toolbox_test.js
+++ b/tests/mocha/toolbox_test.js
@@ -8,16 +8,9 @@ suite('Toolbox', function() {
   function convertToolBoxToJSONCaptureWarnings(toolboxXml) {
     // TODO(#3985): Remove after warning log has been removed.
     var toolboxJson;
-    var warnings = captureWarnings(() => {
+    filterWarnings(() => {
       toolboxJson = Blockly.utils.toolbox.convertToolboxToJSON(toolboxXml)
-    });
-    // convertToolboxToJSON doesn't always log a warning.
-    chai.assert.isTrue(warnings.length <= 1);
-    if (warnings.length) {
-      // Only allow expected warning.
-      chai.assert.match(warnings[0],
-          /defining a toolbox using JSON is not ready yet/);
-    }
+    }, /defining a toolbox using JSON is not ready yet/);
     return toolboxJson;
   }
   setup(function() {
--- a/tests/mocha/toolbox_test.js
+++ b/tests/mocha/toolbox_test.js
@@ -9,7 +9,7 @@ suite('Toolbox', function() {
     // TODO(#3985): Remove after warning log has been removed.
     var toolboxJson;
     filterWarnings(() => {
-      toolboxJson = Blockly.utils.toolbox.convertToolboxToJSON(toolboxXml)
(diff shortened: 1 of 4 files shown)


tests/mocha/toolbox_test.js
```diff
@@ -4,7 +4,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-suite('Toolbox', function() {
+suite.only('Toolbox', function() {
+  function convertToolBoxToJSONCaptureWarnings(toolboxXml) {
+    // TODO(#3985): Remove after warning log has been removed.
+    var toolboxJson;
+    captureWarnings(() => {
```
rachel-fenichel, line 11: Do you have a way to make this only capture warnings matching a specific message (or with a specific prefix)? Capturing all warnings seems like a way to miss real warnings.

moniika: I wanted to be more lenient so that that tests wouldn't fail after warning log was removed, but I can check the number of warnings and optionally use regex. For other tests I haven't been using regex to check the content of the warnings, but that can be changed (those that are capturing are only checking the count of warnings).

rachel-fenichel: I don't think you need to check the number of warnings, just that if the warning doesn't match a regex you call `nativeConsoleWarn` to pass it through.

moniika: I found that the warning is not always logged. I added a check to only allow up to 1 warning, and to check the contents of that warning message. Since in working on this I created a helper for asserting a specific warning was logged, that wasn't used, I'll add support for that in a separate PR (this is a special case, but in other cases we always expect a warning).

tests/mocha/toolbox_test.js
```diff
@@ -5,6 +5,21 @@
...
 
 suite('Toolbox', function() {
+  function convertToolBoxToJSONCaptureWarnings(toolboxXml) {
+    // TODO(#3985): Remove after warning log has been removed.
+    var toolboxJson;
+    var warnings = captureWarnings(() => {
+      toolboxJson = Blockly.utils.toolbox.convertToolboxToJSON(toolboxXml)
+    });
+    // convertToolboxToJSON doesn't always log a warning.
+    chai.assert.isTrue(warnings.length <= 1);
```
rachel-fenichel, line 15: @alschmiedt to comment on why this doesn't always log.

alschmiedt: It only throws a warning if the test is testing defining a toolbox using JSON.

alschmiedt:
I was planning on taking this warning out this quarter since we have looked into performance for JSON and are current plan is to leave as is. I think we are going to detail the problems in our documentation so this won't be needed. 

I would rather we didn't change the toolbox-test file if possible. I have changed it substantially in the toolbox branch and this would cause some unnecessary merge conflicts since I am going to take the warning out anyway.

moniika:
Since the warning will be removed this quarter, closing PR

PR #4211, opened 2020-08-21
Fixing issues with workspace dispose in event tests.

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4194 

### Proposed Changes

Modifying setup in event tests to not disable/enable events when creating a block with variable.

### Reason for Changes

Disabling/enabling events was causing the error:
```
Events Constructors With variable getter blocks Block change
 Error: Tried to call dropdownCreate on a variable field with no variable selected.
    at Blockly.FieldVariable.dropdownCreate [as menuGenerator_] (field_variable.js:390)
    at Blockly.FieldVariable.Blockly.FieldDropdown.getOptions (field_dropdown.js:478)
    at Blockly.FieldVariable.Blockly.FieldDropdown.doValueUpdate_ (field_dropdown.js:521)
    at Blockly.FieldVariable.doValueUpdate_ (field_variable.js:285)
    at Blockly.FieldVariable.initModel (field_variable.js:133)
    at Blockly.FieldVariable.toXml (field_variable.js:180)
    at Object.Blockly.Xml.fieldToDom_ (xml.js:113)
    at Object.Blockly.Xml.allFieldsToDom_ (xml.js:129)
    at Object.Blockly.Xml.blockToDom (xml.js:173)
    at new Blockly.Events.Delete (block_events.js:309)
```
to happen at workspace dispose

### Test Coverage

Ran mocha tests.

Tested on:
 * Desktop Chrome

### DIFF
--- a/tests/mocha/event_test.js
+++ b/tests/mocha/event_test.js
@@ -221,17 +221,19 @@ suite('Events', function() {
       });
     });
 
-    suite('With variable getter blocks', function() {
+    suite.only('With variable getter blocks', function() {
       setup(function() {
-        this.genUidStub = createGenUidStubWithReturns(this.TEST_BLOCK_ID);
-        this.block =
-            createSimpleTestBlock(this.workspace, 'field_variable_test_block');
+        this.genUidStub = createGenUidStubWithReturns(
+            [this.TEST_BLOCK_ID, 'test_var_id', 'test_group_id']);
+        // Disabling events when creating a block with variable can cause issues
+        // at workspace dispose.
+        this.block = new Blockly.Block(
+            this.workspace, 'field_variable_test_block');
       });
 
       test('Change', function() {
         var event = new Blockly.Events.Change(
             this.block, 'field', 'VAR', 'id1', 'id2');
-        sinon.assert.calledOnce(this.genUidStub);
         assertEventEquals(event, Blockly.Events.CHANGE, this.workspace.id,
             this.TEST_BLOCK_ID,
             {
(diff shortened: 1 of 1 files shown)


tests/mocha/event_test.js
```diff
@@ -223,15 +223,17 @@ suite('Events', function() {
...
     suite('With variable getter blocks', function() {
       setup(function() {
-        this.genUidStub = createGenUidStubWithReturns(this.TEST_BLOCK_ID);
-        this.block =
-            createSimpleTestBlock(this.workspace, 'field_variable_test_block');
+        this.genUidStub = createGenUidStubWithReturns(
+            [this.TEST_BLOCK_ID, 'test_var_id', 'test_group_id']);
+        // Disabling events when creating a block with variable can cause issues
+        // at workspace dispose.
+        this.block = new Blockly.Block(
```
samelhusseini, line 230: Would it make more sense to instead just pass an optional parameter (to not disable events for `createSimpleTestBlock`). In other words, is there value in keeping these tests consistent in how they're creating the Block (by going through `createSimpleTestBlock`), and for the cases where you don't want to disable events during creation, that can be accommodated through an optional parameter.

moniika: I prefer not to add a flag to `createSimpleTestBlock` because I don't think the naming makes sense to use `createSimpleTestBlock` to create a block other than `simple_test_block`. I could rename, but this is the only case that's different, so it makes more sense to me to not re-purpose the helper.

Issue #3929, opened 2020-06-04
Undo'ing a delete of a block with a shadow, loses shadow reference.


samelhusseini:
**Describe the bug**

If you delete a block that has a connected shadow, and undo that. Blockly is dropping the shadow block entirely.

**To Reproduce**

1. Go to the playground
2. Drag a block that has a shadow block (eg: repeat do)
3. Delete the block
4. Undo (ctrl z - or right click -> undo)

**Expected behavior**

The undo to return the same block.

**Screenshots**

<img width="212" alt="Screen Shot 2020-06-04 at 3 36 41 PM" src="https://user-images.githubusercontent.com/16690124/83817329-4d846f80-a679-11ea-9ef7-256c6b06b1b6.png">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**

Possibility related to: https://github.com/google/blockly/pull/3902 (I only suggest that as a good starting point because it's a recent change related to shadow blocks), we should run a git bisect to determine the real root cause.

samelhusseini:
Ran a git bisect, and indeed #3902 introduced this issue, will dig deeper to figure out why.

samelhusseini:
@BeksOmega looking further into this, shadow blocks are not respawn when `Blockly.Events.recordUndo` is false.

Another interesting side effect that I came across while digging into this is (and this is not a regression), if `Blockly.Events.recordUndo = false`, we don't spawn the shadow if there's a connected block. A way to repro that is: 
Take this Xml: 
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_repeat_ext">
    <value name="TIMES">
      <shadow type="math_number">
        <field name="NUM">10</field>
      </shadow>
      <block type="math_arithmetic">
        <field name="OP">ADD</field>
        <value name="A">
          <shadow type="math_number">
            <field name="NUM">1</field>
          </shadow>
        </value>
        <value name="B">
          <shadow type="math_number">
            <field name="NUM">1</field>
          </shadow>
        </value>
      </block>
    </value>
  </block>
</xml>
```
If you import the above, there's a shadow block behind the connected `math_arithmetic` block. If you delete the above, and undo on https://blockly-demo.appspot.com/static/tests/playground.html, you'll notice you lose the shadow block behind it.

samelhusseini:
[...]

PR #3902, opened 2020-05-15
Add programmatically setting shadows

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

N/A

### Proposed Changes

This PR allows developers to assign shadow blocks to connections/inputs programmatically.

### Reason for Changes

This functionality is necessary if you want to assign shadow blocks based on the mutation of a block. For example: if you have a mutable addition block and you want to add a shadow to each input.

PlusBlock

### Test Coverage

NewTests

I also flipped through the toolbox, and everything seems fine!

Tested on:
 * Desktop Chrome

### Documentation

None as far as I know.

### Additional Information

This was actually really easy to add, which kind of scares me.

If anyone can think of other areas that need testing I'm happy to test.

PR #4215, opened 2020-08-21
Add programmatically setting shadows Take 2

### LINKED ISSUE #3964: Revert back programatic setShadow after release
Undo https://github.com/google/blockly/pull/3963 after release. 

Original PR: https://github.com/google/blockly/pull/3902

Release Blockling Bug: https://github.com/google/blockly/issues/3929

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #3964 
Addresses #3929

### Proposed Changes

See #3902 for more details.

To address #3929:
I looked into the `Blockly.Events.recordUndo` issue mentioned [here](https://github.com/google/blockly/issues/3929#issuecomment-641419970). I tracked it back to [this commit](https://github.com/google/blockly/commit/2a1ffa11c4c167e013cb1505afcd6e4cde9291f5#diff-1a1453e5905aa4f669088c9cfc5e6951R510).

After some testing it seems like the reason this was added is because it was allowing create, move, and delete events to handle deleting/respawning shadows in the context of undoing/redoing events (recordUndo is always false when undoing/redoing events). I've changed it so that if those events are recording info about a shadow block, they no longer get applied to the undo stack. This means that shadow spawning/deleting is now handled invariantly by the connect / disconnect functions, which should make this system easier to debug in the future.

### Test Coverage

1) #3929 
  Could not reproduce given the steps in the given issue.

2) Connecting & disconnecting Blocks.
ConnectAndDisconnect

3) Undoing a connect.
UndoConnect

4) Undoing a disconnect.
UndoDisconnect

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

I would really love to add unit tests for this. @moniika do you have any ideas for unit testing undo & redo?

### DIFF
--- a/core/block_events.js
+++ b/core/block_events.js
@@ -218,6 +218,10 @@ Blockly.Events.Create = function(opt_block) {
   if (!opt_block) {
     return;  // Blank event to be populated by fromJson.
   }
+  if (opt_block.isShadow()) {
+    // Moving shadow blocks is handled via disconnection.
+    this.recordUndo = false;
+  }
 
   if (opt_block.workspace.rendered) {
     this.xml = Blockly.Xml.blockToDomWithXY(opt_block);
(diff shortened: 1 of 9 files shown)


BeksOmega:
> Additionally, workspace_tests.js has tests for Redo/Undo, it would be nice to add a test there that tests the case mentioned in the issue (assert that shadow block is not dropped). I don't want to add to much work here, so could also just file an issue to create that test.

I went ahead and added the workspace tests here b/c I wanted to make sure this change was thoroughly tested. But I can move those changes into a different PR to make things easier to review if ya like :D

BeksOmega:
@moniika thanks for testing! This should be gtg now I think :D

PR #4226, opened 2020-08-26
Toolbox parse

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Toolbox parsing was messy. 

### Proposed Changes
Breaks it into two methods: 
1. For parsing the flyout def which can be: NodeList, Array<Node>, Array<json>, string(name of custom category). This returns a list of json for the flyout.
2. For parsing the toolbox def which can be: Node, string(of xml), Json. This returns toolbox json.

### Test Coverage

Tested on:

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -472,8 +472,8 @@ Blockly.Flyout.prototype.show = function(flyoutDef) {
   }
   this.setVisible(true);
 
-  // Parse the Array, Node, NodeList into a a list of flyout items.
-  var parsedContent = Blockly.utils.toolbox.parseFlyoutDef(flyoutDef);
+  // Parse the Array, Node or NodeList into a a list of flyout items.
+  var parsedContent = Blockly.utils.toolbox.convertFlyoutDefToJsonArray(flyoutDef);
   parsedContent = parsedContent.slice(); // Shallow copy of parsedContent.
   var flyoutInfo =
     /** @type {{contents:!Array.<!Object>, gaps:!Array.<number>}} */ (
@@ -531,7 +531,7 @@ Blockly.Flyout.prototype.createFlyoutInfo_ = function(parsedContent) {
       var categoryName = customInfo['custom'];
       var flyoutDef = this.getDynamicCategoryContents_(categoryName);
       var parsedDynamicContent = /** @type {!Array<Blockly.utils.toolbox.FlyoutItemJson>} */
-        (Blockly.utils.toolbox.parseFlyoutDef(flyoutDef));
+        (Blockly.utils.toolbox.convertFlyoutDefToJsonArray(flyoutDef));
       parsedContent.splice.apply(parsedContent, [i, 1].concat(parsedDynamicContent));
       contentInfo = parsedContent[i];
     }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -457,8 +457,8 @@ Blockly.Flyout.prototype.hide = function() {
 
 /**
  * Show and populate the flyout.
- * @param {!Blockly.utils.toolbox.FlyoutDefinition} flyoutDef Contents to display
- *     in the flyout. This is either an array of Nodes, a Node, a
+ * @param {!Blockly.utils.toolbox.FlyoutDefinition|string} flyoutDef Contents to display
+ *     in the flyout. This is either an array of Nodes, a NodeList, a
  *     toolbox definition, or a string with the name of the dynamic category.
  */
 Blockly.Flyout.prototype.show = function(flyoutDef) {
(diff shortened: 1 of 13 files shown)


core/flyout_base.js
```diff
@@ -457,8 +457,8 @@ Blockly.Flyout.prototype.hide = function() {
 
 /**
  * Show and populate the flyout.
- * @param {!Blockly.utils.toolbox.FlyoutDefinition} flyoutDef Contents to display
- *     in the flyout. This is either an array of Nodes, a Node, a
+ * @param {!Blockly.utils.toolbox.FlyoutDefinition|string} flyoutDef Contents to display
```
rachel-fenichel, line 460: Why add a string as part of the type here, instead of making `FlyoutDefinition` include `string`?

core/flyout_base.js
```diff
@@ -514,7 +513,7 @@ Blockly.Flyout.prototype.show = function(flyoutDef) {
 /**
  * Create the contents array and gaps array necessary to create the layout for
  * the flyout.
- * @param {!Array<Blockly.utils.toolbox.FlyoutItem>} parsedContent The array
+ * @param {!Array<Blockly.utils.toolbox.FlyoutItemJson>} parsedContent The array
```
rachel-fenichel, line 516: You have a type for `ToolboxJson` that's an array of `ToolboxItemJson`. Make a parallel type, `FlyoutJson`, instead of explicitly having this be an array.

core/toolbox/category.js
```diff
@@ -697,7 +701,7 @@ Blockly.ToolboxCategory.prototype.getChildToolboxItems = function() {
  * Updates the contents to be displayed in the flyout.
  * If the flyout is open when the contents are updated, refreshSelection on the
  * toolbox must also be called.
- * @param {!Blockly.utils.toolbox.ToolboxDefinition} contents The contents
+ * @param {!Blockly.utils.toolbox.FlyoutDefinition} contents The contents
```
rachel-fenichel, line 704: If `FlyoutDefinition` doesn't include `string`, this needs an explicit `|string`

core/utils/toolbox.js
```diff
@@ -135,75 +136,126 @@ Blockly.utils.toolbox.CATEGORY_TOOLBOX_KIND = 'categoryToolbox';
...
+    return flyoutDef['contents'];
+  }
+
+  // If it is already in the correct format return the flyoutDef.
+  if (Array.isArray(flyoutDef) && flyoutDef.length > 0 &&
+      !flyoutDef[0].nodeType) {
+    return flyoutDef;
   }
 
   return Blockly.utils.toolbox.toolboxXmlToJson_(
```
rachel-fenichel, line 208: Why is this still `toolboxXmlToJson_` if this is guaranteed to be a flyout?

core/flyout_base.js
alschmiedt: Ahh, I found the culprit. So, `convertFlyoutDefToJsonArray` actually does not take in a string. We handle any custom categories before hand in the show method using `getDynamicCategoryContents_` before passing to the `convertFlyoutDefToJsonArray`. We could move the `getDynamicCategoryContents_` to the utils file, we would just need to pass in the workspace anywhere we call `convertFlyoutDefToJsonArray`.

core/flyout_base.js
alschmiedt: The ToolboxJson is a an object with contents `{'contents': []}` having FlyoutJson seems a bit misleading since it is actually an array. Do any of these sound ok?`FlyoutJsonArray` or `ContentsArray` or `FlyoutItemsArray`.

core/utils/toolbox.js
alschmiedt: Updated to xmlToJsonArray_

core/toolbox/category.js
alschmiedt: done

core/flyout_base.js
rachel-fenichel: :+1: 

No opinion on whether to move it into utils.

core/flyout_base.js
rachel-fenichel: Hm I think `FlyoutItemJsonArray`--it's longer but most descriptive.

`ContentsArray` would get overloaded too easily I think.

rachel-fenichel:
Per previous discussion, I approve and will continue looking at types.

core/flyout_base.js
alschmiedt: Ok planning on leaving it for now then.

core/flyout_base.js
alschmiedt: Updated

PR #4232, opened 2020-08-28
Toolbox bugs

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
1. Separators were always being added to the toolbox and not to the flyout because I added items to the flyout based on if they did not exist in the registry.
2. Going from a custom category to a category that has contents using updateFlyoutItems was not working. 
3. Refresh selection on a category that has both categories and blocks was not working.
4. There was extra padding on nested categories. 

### Proposed Changes
1. From our old implementation the position of the separator (either in the flyout or in the toolbox) depends on the item directly before the separator. In the case that the separator is the first item in the list it is considered to be in the flyout.
2. Delete the old customs field on the toolboxItemDef and clear out the flyoutItems list.
3. Remove check for isCollapsible and instead look at whether the category has any contents.
4. Move the padding from toolboxContents to toolboxDiv and move the padding on a category to blocklyRowDiv

### Test Coverage

Tested on:

### DIFF
--- a/core/toolbox/category.js
+++ b/core/toolbox/category.js
@@ -247,7 +247,7 @@ Blockly.ToolboxCategory.prototype.parseContents_ = function(categoryDef) {
         this.flyoutItems_.push(flyoutItem);
         prevIsFlyoutItem = true;
       } else {
-        this.addToolboxItem_(itemDef);
+        this.createToolboxItem_(itemDef);
         prevIsFlyoutItem = false;
       }
     }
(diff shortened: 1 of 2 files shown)


core/toolbox/category.js
```diff
@@ -724,14 +732,15 @@ Blockly.ToolboxCategory.prototype.getChildToolboxItems = function() {
...
-    console.warn('Category can not have both flyout contents and subcategories');
-    return;
-  }
+  this.flyoutItems_ = [];
+
   if (typeof contents == 'string') {
     this.toolboxItemDef_['custom'] = contents;
   } else {
-    this.toolboxItemDef_['contents'] = Blockly.utils.toolbox.convertFlyoutDefToJsonArray(contents);
+    // Removes old custom field when contents is updated.
```
rachel-fenichel, line 740: You're trying to clear any state that's left over from the last update. Why not throw away the whole toolboxItemDef and replace it with a new one, so you're guaranteed to have clean state?

alschmiedt: I think there are pieces on the toolboxItemDef that we still need. For example, we need the categoryStyle on it in refreshTheme.

alschmiedt: Going to merge but can take another look if we decide it is possible.

PR #4241, opened 2020-08-31
Update some inline docs in block.js

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Just updates some docs, no functional changes.

### Reason for Changes

Noticed that the docs for setFieldValue were weird over the weekend, and I figured I would fix them up.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -38,13 +38,13 @@ goog.requireType('Blockly.IASTNodeLocation');
  * Class for one block.
  * Not normally called directly, workspace.newBlock() is preferred.
  * @param {!Blockly.Workspace} workspace The block's workspace.
- * @param {?string} prototypeName Name of the language object containing
+ * @param {!string} prototypeName Name of the language object containing
  *     type-specific functions for this block.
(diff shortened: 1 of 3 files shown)


core/block.js
```diff
@@ -38,13 +38,13 @@ goog.requireType('Blockly.IASTNodeLocation');
  * Class for one block.
  * Not normally called directly, workspace.newBlock() is preferred.
  * @param {!Blockly.Workspace} workspace The block's workspace.
- * @param {?string} prototypeName Name of the language object containing
+ * @param {!string} prototypeName Name of the language object containing
```
moniika, line 41: Changing this to non-nullable breaks compile test. 
```
found   : (null|string)
required: string
  502|   return new Blockly.Block(this, prototypeName, opt_id);
```

It is caused by `Blockly.Workspace.prototype.newBlock` having `prototypeName` be `{?string}`

BeksOmega: Thanks for the ping! I just figured it would work so I didn't watch the build hehe. Should be fixed now.

PR #4262, opened 2020-09-09
Add check for custom components

### LINKED ISSUE #4157: Custom Elements - isTargetInput fails
**Describe the bug**

When using customElements (web-components), some events (i.e. 'input' with BACKSPACE) are not propagated because they are not recognised as an input by [Blockly.utils.isTargetInput](https://github.com/google/blockly/blob/93b5be40b6229cf164e83b63fe5a6cc4d44a9b5f/core/utils.js#L47).

**To Reproduce**

1. Add the sample Blockly workspace
2. Add a custom element input that does not have a `type` property
3. Type in something and press BACKSPACE

**Expected behavior**

Be able to delete the value when pressing BACKSPACE

 - OS: macOS
 - Browser: Chrome
 - Version: 84.0.4147.105 (Official Build) (64-bit)

**Additional context**

Developers building custom elements do not have to add a `type` property to an input element if they want to enforce rules around the input. `Type` can also be used for other purposes such as the `kind` of component - in that case the method would fail without extra information.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#4157

### Proposed Changes
Allows developers to add `data-is-text-input` to their custom elements to signal that it is a text input.

### Reason for Changes
There is not currently a way to tell if custom components are a text input or not. 

### Test Coverage

Tested on:

### Additional Information
Some alternatives: 
1. Could add a different attribute such as kind. But we would have to make sure it does not interfere with attributes that are already used such as type.

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -49,7 +49,7 @@ Blockly.utils.isTargetInput = function(e) {
          e.target.type == 'number' || e.target.type == 'email' ||
          e.target.type == 'password' || e.target.type == 'search' ||
          e.target.type == 'tel' || e.target.type == 'url' ||
-         e.target.isContentEditable;
+         e.target.isContentEditable || e.target.dataset.isTextInput == 'true';
 };
 
 /**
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
I'm fine with the naming. Compilation fails because `isTextInput` is never defined on `dataset`.

Documentation: add something on this page: https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating#input_events

alschmiedt:
Travis: It looks like it just needed a null check. Should be fixed now. 
Documentation: Sounds good will look for an appropriate place to add it.

rachel-fenichel:
lgtm

Issue #3784, opened 2020-04-01
Collapsed blocks do not update when modified


BeksOmega:
**Describe the bug**

Collapsed blocks do not update when they are modified. This includes:
* Inputs added
* Fields added
* Field values changed
* etc

This affects all collapsed blocks, not just the examples below.

**To Reproduce**

1. Drag out a procedure def and a caller.
2. Collapse the caller.
3. Rename the procedure.
4. Observe how the collapsed caller does not update.

**Expected behavior**

Collapsed blocks should update.

**Screenshots**

![CollapsedNotUpdating2](https://user-images.githubusercontent.com/25440652/78167639-a4fd2980-7403-11ea-9615-3824b8585553.gif)
![CollapsedNotUpdating](https://user-images.githubusercontent.com/25440652/78167646-a62e5680-7403-11ea-8860-d32c47ce6839.gif)

**Additional context**

I fixed this in AppInventor's repo in https://github.com/mit-cml/appinventor-sources/pull/2074

Here is the description of that pull request:
1. Adds an extra rendering step to the rendering pipeline in the case that the block is collapsed. This step makes sure that the block is in a proper collapsed state (e.g. all inputs are hidden) before re-rendering the block.
2. Fixes some bugs from core like google/blockly#1967
3. Optimizes toString to make it super zippy. (Probably overkill, but it was fun)

I think the same strategy could work in Core as well. I would be happy to put up a pull request if you are interested.

rachel-fenichel:
Yes, we're interested in a fix.

@moniika who has looked into a variety of related rendering issues to discuss and review the associated PR.

moniika:
+1 would love fix.
Here's a related bug #3619 (probably partial dupe).

BeksOmega:
@moniika Does the solution in the other PR look good to you? I wasn't sure how you guys would feel about the `if (isRendering) return;` solution to the stack overflow problem.

I just wanna check before I start trying to update it. Totally cool with whatever you guys decide.

moniika:
> > @moniika Does the solution in the other PR look good to you? I wasn't sure how you guys would feel about the `if (isRendering) return;` solution to the stack overflow problem.
> > I just wanna check before I start trying to update it. Totally cool with whatever you guys decide.
> 
> I'll take a closer look at that and give you a reply by end of day.

[...]

PR #4264, opened 2020-09-09
Don't render children of collapsed blocks

### LINKED ISSUE #4229: Collapsed block performance
**Describe the bug**

Hidden child blocks of collapsed blocks are being re-rendered when a block is being dragged to connect to it.
This causes unnecessary performance issues.

**To Reproduce**

1. Go to the Blockly playground, and use the Spaghetti option to add a long list of connected blocks to the playground.
2. Click on each of the top level if statements and collapse them.
3. Drag another block and attempt to place it in between these blocks.
4. Notice the lag.

**Expected behavior**

Even though there aren't that many visible blocks, we are still re-rendering the block and all of it's (hidden) children as we drag a block to connect to it.

If we can't easily avoid rendering all blocks (even if they're hidden) on first pass, we can probably skip re-rendering blocks if they're hidden during re-render.

**Additional context**

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4229

### Proposed Changes

- Sets rendered state of children of collapsed blocks to false, so they are not re-rendered when the block is moved.
- Checks if child is rendered before calling `updateDisabled ` in `updateDisabled`.

### Reason for Changes

Improves performance.

### Test Coverage

Tested scenario descibed in #4229 and saw improvement in performance while testing with chrome performance tool.

Tested that issues #3784 was still fixed.

Tested issue #1967 in playground using xml:
```
  <block type="controls_repeat_ext" x="88" y="138">
    <value name="TIMES">
      <shadow type="math_number">
        <field name="NUM">10</field>
      </shadow>
    </value>
    <statement name="DO">
      <block type="text_print">
        <value name="TEXT">
          <shadow type="text">
            <field name="TEXT">already here 1</field>
          </shadow>
        </value>
        <next>
          <block type="text_print">
            <value name="TEXT">
              <shadow type="text">
                <field name="TEXT">already here 2</field>
              </shadow>
            </value>
          </block>
        </next>
      </block>
    </statement>
  </block>
  <block type="text_print" x="88" y="288">
    <value name="TEXT">
      <shadow type="text">
        <field name="TEXT">connected while collapsed</field>
      </shadow>
    </value>
  </block>
</xml>
```
And repro code:
```
  var ws = Blockly.getMainWorkspace();
  var blocks = ws.getTopBlocks();
  blocks[0].setCollapsed(true);
  blocks[0].getInput('DO').connection.connect(blocks[1].previousConnection);
  // Check that attached block is hidden/not rendered
  blocks[0].setCollapsed(false);
  // Check that all blocks are un-hidden/rendered
```

Tested on:
* Desktop Chrome

### Additional Information

n/a


BeksOmega:
I think I might have changed this because of https://github.com/google/blockly/issues/1967 but I can't exactly remember.

core/rendered_connection.js
```diff
@@ -459,6 +459,12 @@ Blockly.RenderedConnection.prototype.onFailedConnect = function(
  */
 Blockly.RenderedConnection.prototype.disconnectInternal_ = function(parentBlock,
     childBlock) {
+  // The input the child block was connected to (if any).
+  var parentInput = parentBlock.getInputWithBlock(childBlock);
+  if (parentInput && !parentInput.isVisible()) {
+    childBlock.rendered = true;
+
+  }
```
moniika, line 467: I'm worried about whether this may have unintended side effects.

core/rendered_connection.js
```diff
@@ -459,6 +459,11 @@ Blockly.RenderedConnection.prototype.onFailedConnect = function(
  */
 Blockly.RenderedConnection.prototype.disconnectInternal_ = function(parentBlock,
     childBlock) {
+  // The input the child block was connected to (if any).
+  var parentInput = parentBlock.getInputWithBlock(childBlock);
+  if (parentInput && !parentInput.isVisible()) {
+    childBlock.rendered = true;
+  }
```
moniika, line 466: I'm worried about whether this may have unintended side effects.

BeksOmega: I think it looks good! Because if the block is top-level it should definitely be visible right? As long as the [nested disconnection test](https://github.com/google/blockly/blob/9ab6273e4c2fb0cd25004650d0020490294eed50/tests/mocha/block_test.js#L1444) is passing it should be fine.

I'm going to tag this with #4288 though cuz I know that `.rendered` is kinda crazy wrt collapsed blocks.

rachel-fenichel (approved):

PR #4296, opened 2020-09-18
Call initSvg for blocks with mutator after loading field xml

### LINKED ISSUE #3542: When block has mutation, initModel is called before fromXml
**Description**
When block has mutation, initModel is called before fromXml. If block contains field_variable, default variable is created before initialization from XML and stays in variableMap after fromXml.

**To Reproduce**
1. try this simple HTML:
```
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <script src="https://blockly-demo.appspot.com/static/blockly_compressed.js"></script>
  <script src=".https://blockly-demo.appspot.com/static/blocks_compressed.js"></script>
  <script src="https://blockly-demo.appspot.com/static/msg/js/en.js"></script>
</head>
<body>
  <button onclick="logit()">log variables</button>
  <div id="blocklyDiv" style="height: 480px; width: 600px;"></div>
  <xml xmlns="https://developers.google.com/blockly/xml" id="toolbox" style="display: none">
    <block type="var"><mutation/></block>
  </xml>
<script>
var logit=function(){console.log(workspace.getAllVariables())}
Blockly.Blocks.var = {
  init: function () {
    this.jsonInit({
      message0: "variable %1",
        args0: [{
          type: "field_variable",
          name: "VAR",
      }],
    });
  },
  mutationToDom: function () {
    var m= Blockly.utils.xml.createElement('mutation');
    m.setAttribute('any', 'any');
    return m
  },
  domToMutation: function (xmlElement) {return;},
};
var workspace = Blockly.inject('blocklyDiv',{media: 'https://blockly-demo.appspot.com/static/media/',toolbox: document.getElementById('toolbox')});
</script>
</body>
</html>
```
2. Move almost two blocks from toolbox to workspace
3. click on button
4. see console (you will see equal number of variables as number of blocks, but every block show same variable)

**Expected behavior**
4. see only one (used) variable

moniika:
## The basics

- [x] I branched from develop

### Resolves

#3542

### Proposed Changes

Calls `initSvg` for block with mutator after fields' `fromXml` are called.

### Reason for Changes

Fixes issue with variable fields with a mutator creating extra variable models.

### Test Coverage

- Tested repro steps in #3542 and saw issue was fixed.
- Ran mocha tests.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -651,6 +651,7 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
       }
     }
 
+    var callInitSvg = false;
     var name = xmlChild.getAttribute('name');
     var xmlChildElement = /** @type {!Element} */ (xmlChild);
     switch (xmlChild.nodeName.toLowerCase()) {
@@ -660,7 +661,7 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
           block.domToMutation(xmlChildElement);
           if (block.initSvg) {
             // Mutation may have added some elements that need initializing.
-            block.initSvg();
+            callInitSvg = true;
           }
         }
         break;
@@ -746,6 +747,11 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
         // Unknown tag; ignore.  Same principle as HTML parsers.
         console.warn('Ignoring unknown tag: ' + xmlChild.nodeName);
     }
+  }
+
+  if (callInitSvg) {
+    // InitSvg needs to be called after variable fields are loaded.
+    block.initSvg();
   }
 
   var inline = xmlBlock.getAttribute('inline');


rachel-fenichel (approved):
I'm approving for the bugfix, but also--why do we call initSvg in domToBlockHeadless?

BeksOmega:
> I'm approving for the bugfix, but also--why do we call initSvg in domToBlockHeadless?

I don't think we need to. The issue is that before `domToBlock` [used to be a giant function](https://github.com/google/blockly/commit/0f8d01209c3217cd6d66a8468606f0e8083c12f1) that handled both rendered and non-rendered blocks. And for some reason `initSvg` was called near the beginning of the function instead of the end, which required another call to it after the mutator was added.

Then `domToBlock` [got split](https://github.com/google/blockly/commit/2844da768f00661c5b2886f73a5be427013673dc) into `domToBlock` and `domToBlockHeadless_` but the `initSvg` call for mutators didn't get removed/moved.

The reason this isn't causing errors is because the call was always wrapped in a check for the existence of `initSvg` (before it was [here](https://github.com/google/blockly/commit/0f8d01209c3217cd6d66a8468606f0e8083c12f1#diff-d242a6098fe7c5d9c01ae8f7ef924a7dc4598545564226fbf823e2a6bcd5b9eeR315), and now it is [here](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L706)).

In modern Blockly mutators [create their icons when they are added](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/block_svg.js#L1152), so this call should be unnecessary.

PR #4299, opened 2020-09-19
scroll perfomance fix

Evgeniy331:
Fix for scroll large DOM tree in the latest Google Chrome

## The basics

- [X] I branched from develop

### Resolves

When we have a large DOM tree scroll performance is very poor in the latest Google Chrome

Here is the link - https://blockly-demo.appspot.com/static/tests/playground.html

Click on spaghetti and try to scroll

And I deployed this fix you can see a difference - http://ydubr-blockly-perfomance-fix.surge.sh/

### Proposed Changes

Disable workspace drag surface and use requestAnimationFrame in onScroll handler

### Reason for Changes

Performance improvements

### Test Coverage

Tested via the playground

Tested on Mac and Windows:

* Desktop Chrome
* Desktop Firefox
* Desktop Safari

### Additional Information

I tested in all modern browser and seems the problem with scroll performance exist only in Google Chrome on Mac and Windows

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -65,7 +65,10 @@ Blockly.inject = function(container, opt_options) {
   // Create surfaces for dragging things. These are optimizations
   // so that the browser does not repaint during the drag.
   var blockDragSurface = new Blockly.BlockDragSurfaceSvg(subContainer);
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
Thanks for the pull request, but we avoid adding browser+version specific code, especially for browsers that are in active development.

PR #4311, opened 2020-09-24
Procedures test workspace cleanup

### LINKED ISSUE #4080: Workspace is not always disposed in mocha tests

moniika:
## The basics

- [x] I branched from develop

### Resolves

Closes #4080

### Proposed Changes

- Refactors tests to generate separate tests for `procedures_callreturn` and `procedures_callnoreturn` rather than use `this.callForAllTypes`.
- Adds `try`/`finally` blocks with `workspaceTeardown` where appropriate
- Removes unnecessary `block.dispose` calls (handled in `Workspace.dispose`)
- Removes unnecessary `deleteVariable` calls (each test has its own `Workspace`)

### Reason for Changes

- Without separating logic into separate test calls, if test assert failed for `procedures_callnoreturn` (was run first) then test was not run for `procedures_callreturn. This makes it easier to debug if there is a bug specific to `procedures_callreturn` or `procedures_callnoreturn`
- Workspaces are now always disposed even if assert fails
- Removing unnecessary logic helps simplify tests

### Test Coverage

Ran mocha tests

Tested on:
* Desktop Chrome

### Documentation

n/a

### Additional Information

Behavior of tests unchanged.

### DIFF
--- a/tests/mocha/procedures_test.js
+++ b/tests/mocha/procedures_test.js
@@ -631,46 +631,33 @@ suite('Procedures', function() {
                 chai.assert.equal(this.callBlock.arguments_[i], argArray[i]);
               }
             }
-            function clearVariables() {
-              // TODO: Update this for typed vars.
-              var variables = this.workspace.getVariablesOfType('');
-              var variableMap = this.workspace.getVariableMap();
-              for (var i = 0, variable; (variable = variables[i]); i++) {
-                variableMap.deleteVariable(variable);
-              }
-            }
             test('Simple Add Arg', function() {
               var args = ['arg1'];
               createMutator.call(this, args);
               assertArgs.call(this, args);
-              clearVariables.call(this);
             });
             test('Multiple Args', function() {
               var args = ['arg1', 'arg2', 'arg3'];
               createMutator.call(this, args);
               assertArgs.call(this, args);
-              clearVariables.call(this);
             });
             test('Simple Change Arg', function() {
               createMutator.call(this, ['arg1']);
               this.argBlock.setFieldValue('arg2', 'NAME');
               this.defBlock.compose(this.containerBlock);
               assertArgs.call(this, ['arg2']);
-              clearVariables.call(this);
             });
             test('lower -> CAPS', function() {
               createMutator.call(this, ['arg']);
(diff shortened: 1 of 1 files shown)


crypt096 (approved):

tests/mocha/procedures_test.js
```diff
@@ -505,7 +484,7 @@ suite('Procedures', function() {
           this.workspaceSvg.clearUndo();
           Blockly.Events.setGroup('g1');
           this.barDef.setEnabled(false);
-          Blockly.Events.setGroup(false);
+          // Blockly.Events.setGroup(false);
```
alschmiedt: Can this be removed?

alschmiedt (approved):

tests/mocha/procedures_test.js
moniika: I commented this temporarily as i was debugging why a certain warning was happening.

PR #4327, opened 2020-09-28
Block factory fix

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Block factory was broken due to: 
1. Using Blockly.Options.parseToolbox instead of Blockly.utils.toolbox.parseToolbox
2. Toolbox Items use the given id on the `<category>` element when creating the html for the toolbox. This causes problems since the `<category>` xml has the same id.

### Proposed Changes
1. Use Blockly.utils.toolbox.parseToolbox.
2. .  Use a different name for the id so that there are not two elements on the page with the same id.

### Test Coverage

Tested on:

### Additional Information
I could potentially keep it as `id` for JSON and change it for xml but I thought this would be more confusing. So changing the name for both.

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -363,3 +363,19 @@ Blockly.Options.parseThemeOptions_ = function(options) {
   return Blockly.Theme.defineTheme(theme.name ||
       ('builtin' + Blockly.utils.IdGenerator.getNextUniqueId()), theme);
 };
+
+/**
+ * Parse the provided toolbox tree into a consistent DOM format.
+ * @param {?Node|?string} toolboxDef DOM tree of blocks, or text representation
+ *    of same.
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
LGTM, but should we be concerned about breaking others in the same way? We could deprecate Blockly.Options.parseToolboxTree (mark as deprecated with warnings) and have it call through to the new function.

alschmiedt:
At the time I couldn't think of any use cases outside of core, but I think this proves there are some so happy to add back in.

moniika (approved):

KlemenDEV:
The patch release is called "Q4 2020 Patch 1" but should probably be "Q3 2020 Patch 1"

PR #4317, opened 2020-09-24
Adding tests for multiline block

moniika:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

n/a

### Proposed Changes

-  Adds tests for multiline block, including blockToCode tests
- Adds generator tests for multiline block

### Reason for Changes

- blockToCode tests were necessary to sufficiently test code generation for multiline text (cannot write tests using blocks for multiline block expected output without using multiline block) , but basic tests were added to generator tests to test order (parentheses and operator precedence).
- Better test coverage.

### Test Coverage

Ran mocha tests

Tested on:
* Desktop Chrome

### Additional Information

Rebased on #4329

Created with issues #3978 and #3979 in mind.

PR #4329, opened 2020-09-28
Fix order errors in text generators

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

- Fixes text block generators logic to use the correct order
- Fixes unit test blocks logic to use the correct order (didn't seem to have caused any issues before, but were incorrect)
- Moves multiline string helper into `text.js` files

### Reason for Changes

- Fixes errors in code generation for text blocks
- multiline string helper is only used in `text.js` files and moving it makes it easier to understand the logic in code for block `text_multiline`

### Test Coverage

Generated and ran unit tests for text in interpreters for all languages.

### Additional Information

First noticed error in order after creating tests for `text_multiline` in  #4317 and saw that generated code was invalid for python for the new tests added.

In creation of this PR, also learned that `ORDER_COMMA` was being misused. In most of our use cases, `ORDER_COMMA` should have been `ORDER_NONE` because although there may be a comma, it is not a [comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator). This also helped fix issues with extra parenthesis.

### DIFF
--- a/generators/dart/text.js
+++ b/generators/dart/text.js
@@ -23,6 +23,20 @@ Blockly.Dart['text'] = function(block) {
   return [code, Blockly.Dart.ORDER_ATOMIC];
 };
 
+/**
+ * Encode a string as a properly escaped multiline Dart string, complete with
+ * quotes.
+ * @param {string} string Text to encode.
+ * @return {string} Dart string.
+ * @private
+ */
+Blockly.Dart.multiline_quote_ = function (string) {
+  var lines = string.split(/\n/g).map(Blockly.Dart.quote_);
+  // Join with the following, plus a newline:
+  // + '\n' +
+  return lines.join(' + \'\\n\' + \n');
+};
+
 Blockly.Dart['text_multiline'] = function(block) {
   // Text value.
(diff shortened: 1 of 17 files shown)


crypt096 (approved):

generators/dart/text.js
```diff
@@ -92,8 +108,9 @@ Blockly.Dart['text_charAt'] = function(block) {
   // Get letter at index.
   // Note: Until January 2013 this block did not have the WHERE input.
   var where = block.getFieldValue('WHERE') || 'FROM_START';
-  var text = Blockly.Dart.valueToCode(block, 'VALUE',
-      Blockly.Dart.ORDER_UNARY_POSTFIX) || '\'\'';
+  var textOrder = (where == 'FIRST' | where == 'FROM_START') ?
```
rachel-fenichel, line 111: Is the single `|` instead of `||` intentional?

moniika: Not intentional, good catch.

rachel-fenichel (approved):

PR #4342, opened 2020-09-30
Clean up some text generator logic

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

- Removed extra set of text in `Blockly.JavaScript['text_getSubstring']`
- Minor refactor of set of text in `Blockly.PHP['text_getSubstring']`

### Reason for Changes

Cleans up code.

### Test Coverage

Ran tests.

### Additional Information

Follow up after #4329. Noticed that some logic that should have been deleted wasn't. Should have no effect on code generation.

### DIFF
--- a/generators/php/text.js
+++ b/generators/php/text.js
@@ -153,14 +153,12 @@ Blockly.PHP['text_getSubstring'] = function(block) {
   // Get substring.
   var where1 = block.getFieldValue('WHERE1');
   var where2 = block.getFieldValue('WHERE2');
+  var text = Blockly.PHP.valueToCode(block, 'STRING',
+      Blockly.PHP.ORDER_NONE) || '\'\'';
   if (where1 == 'FIRST' && where2 == 'LAST') {
-    var text = Blockly.PHP.valueToCode(block, 'STRING',
-        Blockly.PHP.ORDER_NONE) || '\'\'';
     var code = text;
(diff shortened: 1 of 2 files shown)


generators/php/text.js
```diff
@@ -153,14 +153,12 @@ Blockly.PHP['text_getSubstring'] = function(block) {
...
+  var text = Blockly.PHP.valueToCode(block, 'STRING',
+      Blockly.PHP.ORDER_NONE) || '\'\'';
   if (where1 == 'FIRST' && where2 == 'LAST') {
-    var text = Blockly.PHP.valueToCode(block, 'STRING',
-        Blockly.PHP.ORDER_NONE) || '\'\'';
     var code = text;
     return [code, Blockly.PHP.ORDER_NONE];
   } else {
-    var text = Blockly.PHP.valueToCode(block, 'STRING',
-        Blockly.PHP.ORDER_NONE) || '\'\'';
```
alschmiedt, line 163: If you delete text here do you need to delete below in `var code = functionName + '(' + text + ', \''`

alschmiedt (approved):

generators/php/text.js
moniika: I didn't just remove it, I moved it to earlier line, since logic in both if and else both set it to the same value (and use the same order).

Issue #1863, opened 2018-05-15
Default text in comment should probably get selected when the comment is created for easy editing


rachel-fenichel:
### Expected Behavior

Text is selected when the comment is first created.

### Actual Behavior
Comment is selected, but not the comment's text in the textarea.

PR #4349, opened 2020-10-05
select text on comment open

halitanildonmez:
## The basics

- [ X] I branched from develop
- [ X] My pull request is against develop
- [ X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

https://github.com/google/blockly/issues/1863

### Proposed Changes

Selects the text inside a comment box when it is opened

### Reason for Changes

To solve the issue at hand

### Test Coverage
For each platform I have used tests/playground.html and created a function to open comment

Tested on:
Desktop Chrome
Desktop Firefox

### Documentation

No it was not, though I would be willing to update it if need be.


alschmiedt:
I think the original issue has to do with workspace comments instead of block comments. You can create a workspace comment by right clicking on the workspace and selecting "Add Comment" from the context menu. 

If you are still interested in working on this issue I think looking in [workspace_comment_svg.js](https://github.com/google/blockly/blob/master/core/workspace_comment_svg.js) would be a good starting place. I would also mention, that it should only highlight the workspace comment if it has the default text.

Thanks!

halitanildonmez:
Wow I have misunderstood this one :smile: 

Going to fix it asap! Thanks for the input!

halitanildonmez:
done! :smile:

core/workspace_comment_svg.js
```diff
@@ -163,6 +163,10 @@ Blockly.WorkspaceCommentSvg.prototype.initSvg = function() {
   if (!this.getSvgRoot().parentNode) {
     this.workspace.getBubbleCanvas().appendChild(this.getSvgRoot());
   }
+
+  if (this.textarea_) {
+    this.textarea_.select();
+  }
```
alschmiedt, line 169: I think this is on the right track! 

My only concern is that this `initSvg` method will get called in [fromXml](https://github.com/google/blockly/blob/11987dc6d6aa16138b36237e8959f2a3f91eb4e2/core/workspace_comment_svg.js#L584) which will cause all of the comments to be selected when someone loads a workspace from xml. This can be tested by going to the [playground](https://blockly-demo.appspot.com/static/tests/playground.html), creating a workspace comment, changing the text, and then hitting the "Export to XML" and the "Import from XML" buttons.

I think this can be fixed if we add an optional parameter to this method that can be passed in as true in `fromXml`. We can then check that the optional parameter is false before selecting the area.

Thanks! And let me know if you have any questions!

halitanildonmez: Thanks for the help! 

I am not sure about the final status: :

I created a new boolean parameter `deSelectText` if it is false I am selecting the text
From `fromXml` method I set the flag to true. 

It feels a bit strange to me, especially with the variable name, let me know what you think! 

again thanks for the help!

core/workspace_comment_svg.js
```diff
@@ -145,9 +145,12 @@ Blockly.WorkspaceCommentSvg.prototype.dispose = function() {
 /**
  * Create and initialize the SVG representation of a workspace comment.
  * May be called more than once.
+ *
+ * @param {boolean} deSelectText Text inside text area will be selected if false
+ *
  * @package
  */
-Blockly.WorkspaceCommentSvg.prototype.initSvg = function() {
+Blockly.WorkspaceCommentSvg.prototype.initSvg = function(deSelectText) {
```
alschmiedt, line 153: How about opt_noSelect for the variable name?

core/workspace_comment_svg.js
```diff
@@ -145,9 +145,12 @@ Blockly.WorkspaceCommentSvg.prototype.dispose = function() {
 /**
  * Create and initialize the SVG representation of a workspace comment.
  * May be called more than once.
+ *
+ * @param {boolean} deSelectText Text inside text area will be selected if false
```
alschmiedt, line 149: {boolean} can be {boolean=}

The = means it is an optional parameter.

alschmiedt (commented):
It looks like a few build files got added that need to be removed.

Other than that lgtm once the comments are addressed : )

halitanildonmez:
It should be fixed now

halitanildonmez:
Hmm looks like the build failed due to a missing file ` /home/travis/build/google/blockly/blockly_compressed.js ` but that is generated with the build right ? 

I deleted it since it was not intended to be committed.

core/workspace_comment_svg.js
```diff
@@ -145,9 +145,12 @@ Blockly.WorkspaceCommentSvg.prototype.dispose = function() {
 /**
  * Create and initialize the SVG representation of a workspace comment.
  * May be called more than once.
+ *
+ * @param {boolean=} deSelectText Text inside text area will be selected if false
```
alschmiedt, line 149: This comment needs to be updated to use `opt_noSelect`

alschmiedt (commented):
The files that were accidentally added shouldn't be deleted, they should just be reverted to their previous state. You should be able to do something like: 
`git checkout <commit> filename`.
The only change for the pr should be in `workspace_comment_svg.js`.

Thanks for the updates! I know getting the compressed files wrapped up in a pr can be a pain : ) I have definitely done it a few times.

core/workspace_comment_svg.js
halitanildonmez: Ok now things looks better 😄

alschmiedt:
This looks good, but there is still one file that needs to be reverted `msg/js/ja.js`.

alschmiedt (approved):
lgtm thanks @halitanildonmez!

Issue #4345, opened 2020-10-02
Text fields are red sometimes on valid input


moniika:
**Describe the bug**

When a text field with a validator has a character changed that makes it invalid and changed back, the field will still be highlighted red (partially due to the value not having changed). 

**To Reproduce**

![8bd568fc-07f2-47dc-85f4-62d79acd9d6c](https://user-images.githubusercontent.com/6621618/94973210-e8724680-04bf-11eb-90d8-394f8993ee93.gif)

**Expected behavior**

Fields should not be red on valid input

**Additional context**

This issue is reproducible with number fields and date fields.

alschmiedt:
Fixed.

PR #4351, opened 2020-10-05
do value update if the value is valid again

halitanildonmez:
## The basics

- [X] I branched from develop

### Resolves

https://github.com/google/blockly/issues/4345

### Proposed Changes

The issue was, in the field, if the user has entered invalid once the flag `isTextValid_;` was being set to false at `doValueInvalid_` which is fine. If however the user enters correct value, then the flag is never se to true.

I made it so that at field.js `processValidation_` we do `doValueUpdate_` so that we update the flag.

### Reason for Changes

To solve the bug that was opened.

### Test Coverage

Tested on:
Desktop Chrome
Desktop Firefox

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -931,6 +931,8 @@ Blockly.Field.prototype.processValidation_ = function(newValue,
   }
   if (validatedValue !== undefined) {
     newValue = validatedValue;
+    this.doValueUpdate_(newValue);
+    this.forceRerender();
   }
   return newValue;
 };
--- a/core/field.js
+++ b/core/field.js
@@ -899,6 +899,7 @@ Blockly.Field.prototype.setValue = function(newValue) {
   var oldValue = this.getValue();
   if (oldValue === newValue) {
     doLogging && console.log('same, return');
+    this.doValueUpdate_(newValue);
     return;
   }
 
@@ -931,8 +932,6 @@ Blockly.Field.prototype.processValidation_ = function(newValue,
   }
   if (validatedValue !== undefined) {
     newValue = validatedValue;
-    this.doValueUpdate_(newValue);
-    this.forceRerender();
   }
   return newValue;
 };
--- a/core/field.js
+++ b/core/field.js
@@ -898,7 +898,7 @@ Blockly.Field.prototype.setValue = function(newValue) {
   }
   var oldValue = this.getValue();
   if (oldValue === newValue) {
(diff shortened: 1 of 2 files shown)


BeksOmega:
Hello @halitanildonmez Thanks for submitting a fix for this :D

I think that you're definitely looking in the right spot wrt the bug you mentioned! But I think it might be better to modify the logic of the setValue function, rather than modify the processValidation_ function. With the current fix it might run into problems with doValueUpdate_ and forceRerender getting called on values that pass the class validator, but not the local validator. I don't think this would cause bad behavior or anything, but rerenders can be pretty expensive, so it's best to minimize them when possible. Do you feel like that's workable?

And it seems like the travis build failed because some mocha tests need to be updated for the new behavior. These tests are located in your my-blockly-clone/tests/mocha directory. There is also an index.html file in that directory, and if you open that in your browser it will give you more details about what tests are passing/failing.

Specifically it looks like some sinon assertions need updates, so the [sinon assertions](https://sinonjs.org/releases/latest/assertions/) docs might be helpful.

Thank you again for the submission! I know the fields stuff can be kind of a mess hehe so thanks for digging through it :D If you have any questions/comments obviously please feel free to reply!

halitanildonmez:
Hello @BeksOmega ! 

Thank you for the feedback and the help! :smile: I have fixed it with the last two commits (i hope) and hope it looks better now.

Let me know if you want me to change something else :smile:

halitanildonmez:
one note, I changed the tests to calledOnce for value update, hope that is ok

BeksOmega:
Looks perfect! Thank you for the updates @halitanildonmez :D

halitanildonmez:
Awesome! So glad to hear 😄

core/field.js
```diff
@@ -899,6 +899,7 @@ Blockly.Field.prototype.setValue = function(newValue) {
   var oldValue = this.getValue();
   if (oldValue === newValue) {
     doLogging && console.log('same, return');
```
alschmiedt, line 901: I think this needs to be updated to reflect the change. Something like 'same, doValueUpdate_ then return'.

alschmiedt (approved):
This looks great! 

Just one comment and then lgtm.

core/field.js
halitanildonmez: Fixed it! :smile: thank you for the input!

Issue #4049, opened 2020-07-16
Deprecation warning util


rachel-fenichel:
When we deprecate functions we usually add a console.warn about it, often with an explanation of what to do instead.

I propose standardizing onto a function that prints
- the name of the deprecated function
- when it was deprecated
- when it will be deleted
- (optionally) which function to call instead

Detailed explanations of how to replace the call can go in comments instead of in the log.

Motivation:
- standardize messaging
- always have a planned deletion date in deprecation messages

rachel-fenichel:
Followup work: go through all deprecation warnings in functions and make them use the new helper.

halitanildonmez:
I have been going through the code and as far as I can see all the deprecation for functions have warnings except for the constant variable, for example blocks/loops.js Blockly.Constants.Loops.HUE = 120; 

Question is, is there a plan to add warnings for constants or is there anything else to do here ?

alschmiedt:
There is not currently a plan to add warnings for constants.

I think there is [one case](https://github.com/google/blockly/blob/11987dc6d6aa16138b36237e8959f2a3f91eb4e2/core/field_textinput.js#L534) in field_textinput.js that needs to be updated to use `Blockly.utils.deprecation.warn` but other than that I think this is finished.

Thanks for commenting!

halitanildonmez:
No worries 😄 

I can fix that one if you wish ?

alschmiedt:
If you would like to that would be great :)

halitanildonmez:
Done!

alschmiedt:
This looks to be finished.

PR #4356, opened 2020-10-07
add deprecation warning

halitanildonmez:
## The basics

- [X] I branched from develop

### Resolves

https://github.com/google/blockly/issues/4049

### Proposed Changes

Using the deprecation utils to add warning for numberValidator method

### Reason for Changes

To use deprecation warning utils instead of console logging

### Test Coverage

Not applicable in this case I think.

Tested on:

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -531,8 +531,11 @@ Blockly.FieldTextInput.prototype.resizeEditor_ = function() {
  * @deprecated
  */
 Blockly.FieldTextInput.numberValidator = function(text) {
-  console.warn('Blockly.FieldTextInput.numberValidator is deprecated. ' +
-               'Use Blockly.FieldNumber instead.');
+  Blockly.utils.deprecation.warn(
+    'FieldTextInput.numberValidator',
+    'May 2019',
+    'October 2020',
+    'Blockly.FieldNumber');
   if (text === null) {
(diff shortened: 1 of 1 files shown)


core/field_textinput.js
```diff
@@ -531,8 +531,11 @@ Blockly.FieldTextInput.prototype.resizeEditor_ = function() {
  * @deprecated
  */
 Blockly.FieldTextInput.numberValidator = function(text) {
-  console.warn('Blockly.FieldTextInput.numberValidator is deprecated. ' +
-               'Use Blockly.FieldNumber instead.');
+  Blockly.utils.deprecation.warn(
+    'FieldTextInput.numberValidator',
+    'May 2019',
+    'October 2020',
+    'Blockly.FieldNumber');
```
alschmiedt, line 538: Thanks!

Just a few comments then should be good : ) 

I think eslint is unhappy because these should be indented two more spaces. Just as a heads up you can run eslint locally by using `npm run lint`.

Since the second date is the deletion date, can we make it December 2020? I think there are quite a few other field methods that are scheduled for deletion in December so we can go back and delete in same PR.

halitanildonmez: Done!

halitanildonmez: Now it looks good 😄

Issue #4203, opened 2020-08-20
Clean up Blockly UI Events


moniika:
**Is your feature request related to a problem? Please describe.**

UI events in Blockly don’t fit the `Blockly.Events.Ui` pattern well. 
- properties for events don’t fit the names `element`, `newValue`, and `oldValue` well
- properties of Ui events cannot be typed well (for `newValue` and `oldValue`)
- for some events, `oldValue` doesn’t add any information (like in mutatorOpen, warningOpen, commentOpen)
- some newer UI events, like viewport change, have multiple “new” properties (viewport coordinates and scale)

**Describe the solution you'd like**

Split ui events into different events, setting the type based on logical grouping and building in some other way to distinguish these events as ui events.
These events could inherit from a UI events interface and have a property `isUiEvent=true` and have `AbstractEvent` first declare this property (false by default, but true for ui events)

Current UI Events and properties:
| element | oldValue | newValue |
| --- | --- | --- |
| dragStart | null | this.draggingBlock_.getDescendants(false) |
| dragStop | this.draggingBlock_.getDescendants(false) | null | 
| selected | null or id | null or id | 
| click | null | 'workspace' or 'block' | 
| cursorMove | oldNode | newNode | 
| markerMove | oldNode | newNode | 
| commentOpen | true or false | true or false | 
| mutatorOpen | true or false | true or false | 
| warningOpen | true or false | true or false | 
| trashcanOpen | true or false | true or false | 
| category | old category name or null | new category name or null | 
| theme | null | null | 

New grouping of events and properties:

| class | type | properties | old ui events |
| --- | --- | --- | --- |
| Blockly.Events.BlockDrag | drag | blocks, isStart | dragStart, dragStop |
| Blockly.Events.Selected | selected | oldElementId, newElementId | select | 
| Blockly.Events.Click | click | targetType | click | 
| Blockly.Events.MarkerMove | marker_move | oldNode, newNode, isCursor | cursorMove, markerMove | 
| Blockly.Events.BubbleOpen | bubble_open | isOpen, bubbleType | commentOpen, mutatorOpen, warningOpen | 
| Blockly.Events.TrashcanOpen | trashcanOpen | isOpen | trashcanOpen | 
| Blockly.Events.ToolboxItemSelect | toolbox_item_select | oldItem, newItem | category | 
| Blockly.Events.ThemeChange | theme_change | themeName | theme | 
[...]

PR #4370, opened 2020-10-14
Ui events base

moniika:
## The basics

- [ ] I branched from develop

### Resolves

 Part of #4203

### Proposed Changes

- Adds constants for new UI types
- Adds property `IS_UI_EVENT` to `Blockly.Events.Abstract`
- Updates checks for UI events in core to use new `IS_UI_EVENT` property instead of checking `type`

### Reason for Changes

- Preparation for later refactoring of UI events
- Property added as constant on prototype as it does not change per-instance

### Documentation

Not yet

### DIFF
--- a/core/ui_events.js
+++ b/core/ui_events.js
@@ -49,6 +49,12 @@ Blockly.Events.Ui = function(opt_block, opt_element, opt_oldValue,
 Blockly.utils.object.inherits(Blockly.Events.Ui, Blockly.Events.Abstract);
 
 /**
+ * Whether or not the event is a UI event.
+ * @type {boolean}
+ */
+Blockly.Events.Ui.prototype.IS_UI_EVENT = true;
+
+/**
  * Type of this event.
  * @type {string}
  */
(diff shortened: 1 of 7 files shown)


core/ui_events.js
```diff
@@ -48,6 +48,12 @@ Blockly.Events.Ui = function(opt_block, opt_element, opt_oldValue,
 };
 Blockly.utils.object.inherits(Blockly.Events.Ui, Blockly.Events.Abstract);
 
+/**
+ * Whether or not the event is a UI event.
+ * @type {boolean}
+ */
+Blockly.Events.Ui.prototype.IS_UI_EVENT = true;
```
rachel-fenichel, line 55: Why is this a constant?

rachel-fenichel (approved):

core/ui_events.js
moniika: I made it a constant on the prototype because its the same for every instance of the class.

PR #4387, opened 2020-10-19
Refactor selected event

moniika:
## The basics

- [ ] I branched from develop

### Resolves

Part of #4203

### Proposed Changes

Creates dedicated class (`Blockly.Events.Selected`) for selected event

### Reason for Changes

- better typing
- clearer events

### Test Coverage

Tested by logging events on playground and observing that expected events were logged.

Tested on:
 * Desktop Chrome

### Additional Information

Selected events are triggered on blocks AND workspace comments.


core/ui_events.js
```diff
@@ -113,5 +114,63 @@ Blockly.Events.Click.prototype.fromJson = function(json) {
   this.blockId = json['blockId'];
 };
 
+/**
+ * Class for a Selected event.
+ * @param {?string=} opt_oldElementId The id of the previously selected
```
maribethb, line 119: Do null and undefined mean different things for these two parameters (e.g. null if no last selected element, undefined if...?)? It's slightly confusing that the last one is only optional, but the first two allow both undefined and null. The Click event in this file doesn't allow the optional string parameters to be null, so I'd favor treating them the same unless null and undefined have different meanings.

maribethb (approved):

core/ui_events.js
moniika: I've updated the param description. It is different yes. Null means that there was no last selected element and undefined means that it's a blank event.

maribethb: Cool, thanks for the explanation!

PR #4390, opened 2020-10-19
Refactor bubble open events

moniika:
## The basics

- [ ] I branched from develop

### Resolves

Part of #4203

### Proposed Changes

Creates dedicated class (`Blockly.Events.BubbleOpen`) for bubble open events.

### Reason for Changes

- better typing
- clearer events

### Test Coverage

Tested by logging events on playground and observing that expected events were logged.

Tested on:
 * Desktop Chrome


core/events_bubble_open.js
```diff
@@ -0,0 +1,79 @@
...
+   */
+  this.element = opt_element;
+};
+Blockly.utils.object.inherits(Blockly.Events.BubbleOpen, Blockly.Events.Ui);
+
+/**
+ * Type of this event.
+ * @type {string}
+ */
+Blockly.Events.BubbleOpen.prototype.type = Blockly.Events.BUBBLE_OPEN;
```
samelhusseini, line 53: Why are we inheriting from UI events, if we're overriding the type. 

If someone wants to filter for all events that are not UI events, then they won't be able to just do this no more: 
```
if (event.type !== 'ui') {
 ...
}
```

core/events_bubble_open.js
```diff
@@ -0,0 +1,79 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Events fired as a result of bubble open.
+ * @author kozbial@google.com (Monica Kozbial)
+ */
+'use strict';
+
+goog.provide('Blockly.Events.BubbleOpen');
```
samelhusseini, line 13: We could consider moving all these events into their own folder for better organization.

samelhusseini (commented):
Any reason to check in the _compressed files?

core/events_bubble_open.js
moniika: Yes, I was planning on moving the files after I get most PRs in.

moniika:
> Any reason to check in the _compressed files?

I need to rebuild because I added new files and edited provide/requires

core/events_bubble_open.js
moniika: UI events have a new property `IS_UI_EVENT`, in order to filter UI events, one can:
```
if (event.IS_UI_EVENT) {
  ...
}
```

samelhusseini: Ah got it.

samelhusseini (approved):

Issue #1052, opened 2017-04-19
Events for viewport changes


rachel-fenichel:
We need to sort out the format of the event.

These events would be fired when scrolling and zooming the workspace.

The event would contain:
- scale of the workspace
- upper left corner of the visible workspace, in workspace coordinates

moniika:
Enabled as part of #4418

PR #4395, opened 2020-10-22
Viewport events

moniika:
## The basics

- [ ] I branched from develop

### Resolves

Part of #4203
Part of #1052

### Proposed Changes

Creates dedicated class (`Blockly.Events.ViewportChange`) for viewport change events.

### Reason for Changes

- better typing
- clearer events

### Test Coverage

Tested by logging events on playground and observing that expected events were logged and running mocha tests.

Tested on:
 * Desktop Chrome


alschmiedt (approved):
this lgtm! I just have a few things I am curious about: 
1. Is there a use case in mind for this event?
2. Depending on the use case, would it be helpful to have the width and size of the view port for this event? 
3. Do we nee the blockly_compressed file? I know we need blockly_uncompressed since you added a file, but not sure about blockly_compressed.

moniika:
> this lgtm! I just have a few things I am curious about:
> 
> 1. Is there a use case in mind for this event?
> 2. Depending on the use case, would it be helpful to have the width and size of the view port for this event?
> 3. Do we nee the blockly_compressed file? I know we need blockly_uncompressed since you added a file, but not sure about blockly_compressed.

1. I can't remember the specific use case, but I've added the related bug
2. I was also thinking width/height (size) might be useful, Maybe to follow up I can add.
3. I had just made the assumption that it was also needed, but looks like it isn't necessary.

I'm going to leave for now, but may add viewport size information in a follow up.

PR #4398, opened 2020-10-23
Adds shortcut registry and removes action and key map

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Initial commit of keyboard shortcuts. 
Note: I've broken the keyboard shortcuts registry up into multiple PRs to make things easier. For the full change see: #4397. Because of this I expect travis to fail.

### Proposed Changes
Add a keyboard shortcut registry. Remove `action.js` and `key_map.js` in favor of the new shortcut registry.

### Test Coverage

Tested on:

### Additional Information
- The registry was broken up into a registry_ of shortcuts and a keyMap_. This is done to make it easier to save a key map. 
- I went back and forth a bit between KeyboardShortcut and KeyboardAction as the name. I stuck with KeyboardShortcuts since I think it makes the most sense. However, KeyboardAction works better with `onBlocklyAction`. Lmk if you have any opinions.
- I used clang to format since it is a new file. It was great : )


core/keyboardshortcut_registry.js
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Gets the current key map.
+ * @return {!Object<string,!Array<!Blockly.KeyboardShortcutRegistry.KeyboardShortcut>>}
+ *     The object holding key codes to
+ *     Blockly.KeyboardShortcutRegistry.KeyboardShortcut.
+ * @public
+ */
+Blockly.KeyboardShortcutRegistry.prototype.getKeyMap = function() {
+  var keyMap = {};
```
samelhusseini, line 198: Curious why we're doing a copy?
Also just an FYI, mixin doesn't do a deep copy.

core/keyboardshortcut_registry.js
```diff
@@ -0,0 +1,351 @@
...
+Blockly.KeyboardShortcutRegistry = function() {
+  // Singleton instance should be registered once.
+  Blockly.KeyboardShortcutRegistry.registry = this;
+
+  /**
+   * Registry of all keyboard shortcuts, keyed by name of shortcut.
+   * @type {!Object<string, !Blockly.KeyboardShortcutRegistry.KeyboardShortcut>}
+   * @private
+   */
+  this.registry_ = {};
```
samelhusseini, line 36: Safest to use Object.create(null) here and we won't have to worry about inherited properties.

core/keyboardshortcut_registry.js
```diff
@@ -0,0 +1,351 @@
...
+
+  /**
+   * Map of key codes to an array of shortcut names.
+   * @type {!Object<string, !Array<string>>}
+   * @private
+   */
+  this.keyMap_ = {};
```
samelhusseini, line 43: Same here, use Object.create(null)

core/keyboardshortcut_registry.js
```diff
@@ -0,0 +1,351 @@
...
+ * @throws {Error} if the modifier is not in the valid modifiers list.
+ * @private
+ */
+Blockly.KeyboardShortcutRegistry.prototype.checkModifiers_ = function(
+    modifiers) {
+  var validModifiers = Blockly.utils.object.values(
+      Blockly.KeyboardShortcutRegistry.modifierKeys);
+  for (var i = 0, modifier; (modifier = modifiers[i]); i++) {
+    if (validModifiers.indexOf(modifier) < 0) {
+      throw Error(modifier + ' is not a valid modifier key.');
```
samelhusseini, line 310: Always throw with `new` according to the styleguide: https://google.github.io/styleguide/jsguide.html#features-exceptions

but there's a number of instances of this in Blockly, so happy to keep it optional.

samelhusseini (approved):

core/keyboardshortcut_registry.js
```diff
@@ -0,0 +1,351 @@
...
+goog.require('Blockly.utils.object');
+
+
+/**
+ * Class for the registry of keyboard shortcuts. This is intended to be a
+ * singleton. You should not create a new instance, and only access this class
+ * from Blockly.KeyboardShortcutRegistry.registry.
+ * @constructor
+ */
+Blockly.KeyboardShortcutRegistry = function() {
```
samelhusseini, line 27: Do you forsee a different type of Shortcut registry. Curious if we could just call it the ShortcutRegistry?

core/keyboardshortcut_registry.js
alschmiedt: done.

core/keyboardshortcut_registry.js
alschmiedt: From what I can remember this was to prevent people from accidentally changing the key map. They would need to use `setKeyMap` to update it. I don't have super strong feelings either way, but it seems nice to have. 

I think I can accomplish this by calling `Blockly.utils.object.deepMerge` instead of `Blockly.utils.object.mixin`.

core/keyboardshortcut_registry.js
alschmiedt: Not that I can really think of. I stuck with KeyboardShortcutRegistry because it was more descriptive but it is pretty cumbersome so happy to move to Blockly.ShortcutRegistry.

alschmiedt: done

core/keyboardshortcut_registry.js
alschmiedt: done

core/keyboardshortcut_registry.js
alschmiedt: done

PR #4400, opened 2020-10-24
Move events to new directory and rename Ui events base

moniika:
## The basics

- [ ] I branched from develop

### Resolves

Part of #4203

### Proposed Changes

- Move events into events directory
- Rename base class for UI to `UiBase` 
- Rename old ui events class back to `Ui`
- Moves `Ui` class into same file as `UiBase`
Creates dedicated class (`Blockly.Events.ViewportChange`) for viewport change events.

### Reason for Changes

- Organization
- Backwards compatibility

### Test Coverage

Tested by logging events on playground and observing that expected events were logged and running mocha tests.

Tested on:
 * Desktop Chrome


core/events/ui_events.js
```diff
@@ -10,13 +10,50 @@
...
+goog.provide('Blockly.Events.UiBase');
 
 goog.require('Blockly.Events');
-goog.require('Blockly.Events.Ui');
+goog.require('Blockly.Events.Abstract');
 goog.require('Blockly.registry');
 goog.require('Blockly.utils.object');
 
+/**
+ * Class for a UI event.
```
maribethb, line 22: assuming i understand these changes correctly, i might add 'base class for a ui event, use the more specific event type in core/events if possible' or some other text explaining. because if i understand correctly this isn't meant to be used directly (?)

and same below for the plain Ui event, maybe add that it's there for compatibility but you should use the new event classes

maribethb (approved):

core/events/ui_events.js
moniika: I've update Jsdoc for both and added `@deprecated` annotation to `Events.Ui`

moniika: Adding `@deprecated` is an issue for travis until I also fix zoom event that is still using old class. Since I want to handle that in a seperate PR, I'm just going to update the jsdoc here.

PR #4412, opened 2020-10-27
Shortcut tests and refactored navigation tests

alschmiedt:
## The basics

- [ ] I branched from develop

### Proposed Changes
Adds tests for keyboard shortcuts.
Refactors tests for navigation.
Removes key map tests. 

### Test Coverage

Tested on:

### Additional Information
This is a partial PR and will therefore fail on Travis. For the full PR see #4397.


tests/mocha/shortcut_registry_test.js
```diff
@@ -0,0 +1,352 @@
...
+  suite('removeKeyMapping', function() {
+    test('Removes a key mapping', function() {
+      this.registry.registry_['test_shortcut'] = {'name': 'test_shortcut'};
+      this.registry.keyMap_['keyCode'] = ['test_shortcut', 'test_shortcut_2'];
+
+      var isRemoved =
+          this.registry.removeKeyMapping('keyCode', 'test_shortcut');
+
+      var shortcutNames = this.registry.keyMap_['keyCode'];
+      chai.assert.lengthOf(shortcutNames, 1);
```
moniika, line 140: I would probably also check that the only shortcut in list is 'test_shortcut_2'

tests/mocha/shortcut_registry_test.js
```diff
@@ -0,0 +1,352 @@
...
+      this.registry.registry_['test_shortcut'] = {'name': 'test_shortcut'};
+      this.registry.keyMap_['keyCode'] = ['test_shortcut'];
+
+      this.registry.removeKeyMapping('keyCode', 'test_shortcut');
+
+      var shortcutNames = this.registry.keyMap_['keyCode'];
+      chai.assert.isUndefined(shortcutNames);
+    });
+    test('Removes a key map that does not exist opt_quiet=false', function() {
+      var consoleSpy = sinon.spy(console, 'warn');
```
moniika, line 153: It would be better if we could prevent expected console logs in order to keep the console clear (for better debugging). 

For that reason, I would instead recommend using a stub (and also checking that it was only called once). `sinon.assert.calledWith` also works with stubs (since it is considered a spy).

(similar thinking with other tests in this suite that trigger console (and spy on warn).

moniika (approved):
Looks good. My main concern is just about how some of the tests seem to trigger console logs. As part of #4194, been trying to clean them up.

tests/mocha/shortcut_registry_test.js
alschmiedt: done

tests/mocha/shortcut_registry_test.js
alschmiedt: Updated!

Issue #4260, opened 2020-09-08
Refactor field tests in core


moniika:
Refactor field tests in core to take advantage of test helpers available in dev-tools from Blockly Samples in order to improve readability and overall test coverage.

PR #4417, opened 2020-10-29
Refactor field_variable_tests to use dev-tools

moniika:
## The basics

- [x] I branched from develop

### Resolves
 Part of #4260

### Proposed Changes

- Refactors `FieldVariable` tests to use dev-tools helpers
- Modifies handling of invalid values passed to `FieldVariable` constructor (for truthy non-string values like number or true)

### Reason for Changes

- Expands coverage
- Better handling of invalid values (to use default value rather than causing error later down the line.

### Test Coverage

Ran mocha tests (using local version of dev-tools).

### Additional Information

Depends on change https://github.com/google/blockly-samples/pull/428

Old tests:
image

New tests:
image

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -65,7 +65,8 @@ Blockly.FieldVariable = function(varName, opt_validator, opt_variableTypes,
    * variable.
    * @type {string}
    */
-  this.defaultVariableName = varName || '';
+  this.defaultVariableName =
+      typeof varName === 'string' || varName instanceof String ? varName : '';
 
   /**
    * The size of the area rendered by the field.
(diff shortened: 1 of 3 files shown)


core/field_variable.js
```diff
@@ -65,7 +65,8 @@ Blockly.FieldVariable = function(varName, opt_validator, opt_variableTypes,
    * variable.
    * @type {string}
    */
-  this.defaultVariableName = varName || '';
+  this.defaultVariableName =
+      typeof varName === 'string' || varName instanceof String ? varName : '';
```
maribethb, line 69: I'm surprised you need both of these conditions, or either of them really based on the types here. varName can only be null or string, so the existing code theoretically should have been fine. (or maybe varName !== null)... it looks like this line is upsetting closure though.

maribethb (approved):

core/field_variable.js
moniika: Updated.

moniika:
I'll be waiting for the next Blockly-Samples release before merging this PR.

PR #4432, opened 2020-11-04
Renaming ui event properties and updating documentation for consistency

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #4203

### Proposed Changes

- Updates to documentation wording.
- Rename of some properties and class name (`Drag` -> `BlockDrag`) for consistency

### Reason for Changes

Cleanup for new api and for better autogenerated documentation.

### Test Coverage

Ran mocha tests and tested in playground.

Tested on:
* Desktop Chrome


core/gesture.js
```diff
@@ -665,7 +665,7 @@ Blockly.Gesture.prototype.handleWsStart = function(e, ws) {
  * @private
  */
 Blockly.Gesture.prototype.fireWorkspaceClick_ = function(ws) {
-  var clickEvent = new Blockly.Events.Click(null, ws.id);
+  var clickEvent = new Blockly.Events.Click(null, ws.id, 'workspace');
```
alschmiedt, line 668: Could we make this a constant somewhere?

alschmiedt (approved):

core/events/events_click.js
```diff
@@ -33,10 +34,6 @@ Blockly.Events.Click = function(opt_block, opt_workspaceId, opt_targetType) {
   Blockly.Events.Click.superClass_.constructor.call(this, workspaceId);
   this.blockId = opt_block ? opt_block.id : null;
 
-  if (!opt_targetType && !this.isBlank) {
-    opt_targetType = opt_block ? 'block' : 'workspace';
-  }
-
```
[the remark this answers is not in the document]
moniika: At first the only 2 types were 'block' and 'workspace', but now it also used for zoom control ui click and it seems confusing to have it try to do something smart rather than explicitly passing it. Removing this seems like a better idea to me so there's no confusion of whether you need to pass the targetType.

core/gesture.js
moniika: For `'workspace'`? I'd rather add that in a separate PR.  I would do so for all the strings passed for `targetType` and probably also do a similar thing for `bubbleType` on `BubbleOpen`.

PR #4461, opened 2020-11-18
Connecting block to parent in domToBlockHeadless_

### LINKED ISSUE #2135: Dynamic dropdown "forgets" option keys when reloaded
Given a dynamic dropdown, i.e. one where `options` is set to a function instead of a fixed array, if that function returns a list of options where the option key is different to the option value, the option key is lost upon reloading the workspace.

### Expected Behavior

A dynamic dropdown should always show the option key to the user after a workspace using it is reloaded, not the option value.

### Actual Behavior

Given a dynamic dropdown, if the workspace containing it is persisted to XML and then later reloaded, the option keys are lost and the option value is shown to the user in the workspace.

### Steps to Reproduce

Given a dynamic dropdown function like:

```
function dynamicOptions() {
   return [["a", "b"]];
```

If you have a block on your workspace that uses it, it will show `a` as the option to the user upon being dragged onto the workspace.

If you then write the workspace to XML with `Blockly.Xml`, create a new empty workspace and populate it from the XML (again with `Blockly.Xml`) it will show the same block with the same dropdown, but this time the option will be shown as `b` to the user, i.e. Blockly is treating the option value as the key for display purposes.

### Additional Information

From what I can see the issue is that the option keys are not represented in the workspace XML and the dropdown does not infer them from the dynamic option function when deserialized.

Strangely, if you open the dropdown to see the options, it highlights the correct one by key.

In my real life use case, the option values are a JSON datastructure, so this looks pretty ugly. Is there any way to force the dropdown to re-initialize the options upon deserialization?

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #2135

### Proposed Changes

- Connects block to it's parent in domToBlockHeadless_ rather than in the parent block's instantiation (allows for parent to be connected before field tag is processed)
  - Before: Block was connected it it's parent after it was fully loaded (which caused issues with fields validators that depended on the connected state of the block)
  - After: Block is connected to parent as it is being loaded, before it's field values are loaded

### Reason for Changes

- Solves issues with loading blocks with field validators that depend on the parent being connected (and accessing parent fields) from xml
- Solves issues with loading dynamic dropdowns that depend on a connected parent from xml

### Test Coverage

Ran tests and tested in advanced playground with the following block:
```
  <variables>
    <variable>list</variable>
  </variables>
  <block type="controls_repeat_ext">
    <value name="TIMES">
      <block type="math_number">
        <field name="NUM">123</field>
      </block>
    </value>
    <statement name="DO">
      <block type="test_dropdowns_dynamic_connect_dependant">
        <field name="OPTIONS">wcjS31{oxnJqsPUgQ+_m_type_key</field>
        <next>
          <block type="lists_getIndex">
            <mutation statement="true" at="true"></mutation>
            <field name="MODE">REMOVE</field>
            <field name="WHERE">FROM_START</field>
            <value name="VALUE">
              <block type="variables_get">
                <field name="VAR">list</field>
              </block>
            </value>
            <value name="AT">
              <block type="lists_getIndex">
                <mutation statement="false" at="true"></mutation>
                <field name="MODE">GET</field>
                <field name="WHERE">FROM_START</field>
                <value name="VALUE">
                  <block type="variables_get">
                    <field name="VAR">list</field>
                  </block>
                </value>
                <value name="AT">
                  <block type="math_number">
                    <field name="NUM">123</field>
                  </block>
                </value>
              </block>
            </value>
            <next>
              <block type="test_dropdowns_dynamic_connect_dependant">
                <field name="OPTIONS">T1_y8l67gg$DXQ=u8Q~2_type_key</field>
              </block>
            </next>
          </block>
        </next>
      </block>
    </statement>
  </block>
</xml>
```
### Additional Information

The generator tests were very helpful for testing this change as it caught many edge cases of xml generation.

Depends on https://github.com/google/blockly/pull/4571

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -797,9 +797,8 @@ Blockly.Xml.applyInputTagNodes_ = function(xmlChildren, workspace, block,
     }
     var childBlockInfo = Blockly.Xml.findChildBlocks_(xmlChild);
     if (childBlockInfo.childBlockElement) {
-      // Create child block.
-      Blockly.Xml.domToBlockHeadless_(
-          childBlockInfo.childBlockElement, workspace, input.connection);
+      Blockly.Xml.domToBlockHeadless_(childBlockInfo.childBlockElement,
+          workspace, input.connection, false);
     }
     // Set shadow after so we don't create a shadow we delete immediately.
     if (childBlockInfo.childShadowElement) {
@@ -828,8 +827,8 @@ Blockly.Xml.applyNextTagNodes_ = function(xmlChildren, workspace, block) {
         throw TypeError('Next statement is already connected.');
       }
       // Create child block.
-      Blockly.Xml.domToBlockHeadless_(
-          childBlockInfo.childBlockElement, workspace, block.nextConnection);
+      Blockly.Xml.domToBlockHeadless_(childBlockInfo.childBlockElement,
+          workspace, block.nextConnection, true);
     }
     // Set shadow after so we don't create a shadow we delete immediately.
     if (childBlockInfo.childShadowElement && block.nextConnection) {
@@ -846,11 +845,13 @@ Blockly.Xml.applyNextTagNodes_ = function(xmlChildren, workspace, block) {
  * @param {!Blockly.Workspace} workspace The workspace.
  * @param {!Blockly.Connection=} parentConnection The parent connection to
  *    to connect this block to after instantiating.
+ * @param {boolean} connectedToParentNext Whether the provided parent connection
+ *    is a next connection, rather than output or statement.
  * @return {!Blockly.Block} The root block created.
  * @private
  */
 Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace,
-    parentConnection) {
+    parentConnection, connectedToParentNext) {
   var block = null;
   var prototypeName = xmlBlock.getAttribute('type');
   if (!prototypeName) {
@@ -869,13 +870,22 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace,
 
   // Connect parent after processing mutation and before setting fields.
   if (parentConnection) {
-    if (block.outputConnection) {
-      parentConnection.connect(block.outputConnection);
-    } else if (block.previousConnection) {
-      parentConnection.connect(block.previousConnection);
+    if (connectedToParentNext) {
+      if (block.previousConnection) {
+        parentConnection.connect(block.previousConnection);
+      } else {
+        throw TypeError(
+            'Next block does not have previous statement.');
+      }
     } else {
-      throw TypeError(
-          'Child block does not have output or previous statement.');
+      if (block.outputConnection) {
+        parentConnection.connect(block.outputConnection);
+      } else if (block.previousConnection) {
+        parentConnection.connect(block.previousConnection);
+      } else {
+        throw TypeError(
+            'Child block does not have output or previous statement.');
+      }
     }
   }
 
--- a/core/xml.js
+++ b/core/xml.js
@@ -797,9 +797,11 @@ Blockly.Xml.applyInputTagNodes_ = function(xmlChildren, workspace, block,
     }
     var childBlockInfo = Blockly.Xml.findChildBlocks_(xmlChild);
     if (childBlockInfo.childBlockElement) {
+      if (!input.connection) {
+        throw TypeError('Input connection does not exist.');
+      }
       Blockly.Xml.domToBlockHeadless_(childBlockInfo.childBlockElement,
-          workspace, /** @type {!Blockly.Connection} */ (input.connection),
-          false);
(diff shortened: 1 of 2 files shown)


BeksOmega:
With the new connection checker being more flexible, I wonder about connection checks that depend on the state of a parent block's fields. Wouldn't this mess with that?

core/xml.js
```diff
@@ -631,125 +634,164 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
...
-        }
+        xmlChildNameMap.next.push(xmlChild);
         break;
       default:
         // Unknown tag; ignore.  Same principle as HTML parsers.
         console.warn('Ignoring unknown tag: ' + xmlChild.nodeName);
     }
   }
 
+  var callInitSvg = false;
```
rachel-fenichel, line 687: Break this out into helper functions where possible.

core/xml.js
```diff
@@ -631,125 +634,164 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
...
+   * A mapping of nodeName to node for child nodes of xmlBlock.
+   * @type {{next: Array<Element>, mutation: Array<Element>,
+   *    input: Array<Element>, data: Array<Element>, field: Array<Element>,
+   *    comment: Array<Element>}}
+   */
+  var xmlChildNameMap = {
+    mutation: [], comment: [], data: [], field: [], input: [],
+    next: []
+  };
   for (var i = 0, xmlChild; (xmlChild = xmlBlock.childNodes[i]); i++) {
```
rachel-fenichel, line 648: This loop can all be pulled out into a helper function that returns the generated map.

package.json
```diff
@@ -37,6 +37,7 @@
     "recompile": "gulp gitPreCompile && npm run bump && gulp gitPostCompile",
     "release": "gulp gitCreateRC",
     "test": "concurrently 'npm run test:prepare' 'sleep 5 && npm run test:run'",
+    "test:generators": "concurrently 'npm run test:prepare' 'sleep 25 && tests/scripts/run_generators.sh'",
```
rachel-fenichel, line 40: Why is the sleep now 25?

rachel-fenichel (commented):
Some comments.

I'm nervous about this because of backwards compatibility, but agree that making our parsing more clearly defined is a good move.

core/xml.js
moniika: Done.

core/xml.js
```diff
@@ -631,125 +634,164 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
   var id = xmlBlock.getAttribute('id');
   block = workspace.newBlock(prototypeName, id);
 
-  var blockChild = null;
+  // Preprocess childNodes so tags can be processed in a consistent order.
+  /**
+   * A mapping of nodeName to node for child nodes of xmlBlock.
+   * @type {{next: Array<Element>, mutation: Array<Element>,
```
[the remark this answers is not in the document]
moniika: Done

core/xml.js
moniika: Done

package.json
moniika: At some point while i was testing the extra sleep was necessary for running generator tests. However, it looks like 5 seconds works for me locally, so I will change to 5.

core/xml.js
```diff
@@ -619,10 +619,13 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
  * workspace.
  * @param {!Element} xmlBlock XML block element.
  * @param {!Blockly.Workspace} workspace The workspace.
+ * @param {?Blockly.Connection=} parentConnection The parent connection to
```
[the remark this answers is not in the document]
moniika: Updated.

moniika:
Rebased after merging change for parsing tags in explicit order.

moniika:
> With the new connection checker being more flexible, I wonder about connection checks that depend on the state of a parent block's fields. Wouldn't this mess with that?

I don't think this would be an issue if the connection checker depends on the **parent** block's fields.
The xml (#4571) is loaded in an explicit order, such that, the input/statement/value fields that create the child block are processed after the field tags are processed.

However, if the connection depends on the **child** block's fields, then those wont be loaded yet. 
(Ex: a block with boolean field that must be true in order for it's output to be connected to parent)
(Ex: a block whose field value must match the field value of the parent in order for it to be connected <- in previous release, this case only worked if the field in question was listed before the child block in xml)

Other potential **new** issue (introduced in #4571):
- field validator depends on connected child blocks (used to work **only if** the field tag was listed before the statement/value/input tag in the xml)

Other Remaining unsolved issues (not new issues):
- child field validator depends on parent's other connected statement/input/value child blocks
- field validator depends on all blocks in the workspace being loaded

rachel-fenichel (approved):

PR #4484, opened 2020-11-30
Remove warning from Block bumpNeighbors

### LINKED ISSUE #4312: Library blocks call bumpNeighbors in headless mode which triggers warning
**Describe the bug**

Multiple library blocks, including:
- `procedures_defreturn`, `procedures_defnoreturn`, `procedures_callreturn`, `procedures_callnoreturn`
- `list_split`
- `logic_ternary`, `logic_compare`

Have calls to `bumpNeighbours` which causes a warning to be triggered when in headless.

**To Reproduce**

1.
```
var ws = new Blockly.Workspace();
var block = ws.newBlock('logic_ternary');
var otherBlock = ws.newBlock('math_number');
block.getInput('THEN').connection.connect(otherBlock.outputConnection);
var parent = ws.newBlock('text_trim');
// input mismatch triggering disconnect of math block from THEN and bumpNeighbors() call
parent.getInput('TEXT').connection.connect(block.outputConnection);
```
2. Observe warning:
```
Not expected to reach Block.bumpNeighbours function. BlockSvg.bumpNeighbours was expected to be called instead.
```

**Expected behavior**

There should not be a warning when there is not an issue.

**Additional context**

Causes warning in headless tests, tracked in #4194

moniika:
## The basics

- [x] I branched from develop

### Resolves
 Fixes #4312

### Proposed Changes

Removes warning log from `Block` implementation of `bumpNeighbors`.

### Reason for Changes
Without this check, calling `bumpNeighbors` in headless cases an unnecessary warning. Also, cleans up console logs for our tests that triggered this warning.

### Test Coverage

Ran mocha tests.

Tested on:
* Desktop Chrome

### DIFF
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -829,7 +829,9 @@ Blockly.Blocks['lists_split'] = {
         if (inputBlock.isShadow()) {
           inputBlock.dispose();
         } else {
-          this.bumpNeighbours();
+          if (this.rendered) {
+            this.bumpNeighbours();
+          }
         }
       }
     }
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -829,9 +829,7 @@ Blockly.Blocks['lists_split'] = {
         if (inputBlock.isShadow()) {
           inputBlock.dispose();
         } else {
-          if (this.rendered) {
-            this.bumpNeighbours();
-          }
+          this.bumpNeighbours();
         }
       }
     }
--- a/blocks/logic.js
+++ b/blocks/logic.js
@@ -564,7 +564,9 @@ Blockly.Constants.Logic.LOGIC_COMPARE_ONCHANGE_MIXIN = {
           this.getInput('B').connection.connect(prevB.outputConnection);
         }
       }
-      this.bumpNeighbours();
+      if (this.rendered) {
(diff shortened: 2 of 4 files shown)


BeksOmega:
I'm just curious, so don't feel the need to respond hehe, but has moving all of the rendered checks into bumpNeighbours itself (and removing the relevant warning) ever been considered?

moniika:
> I'm just curious, so don't feel the need to respond hehe, but has moving all of the rendered checks into bumpNeighbours itself (and removing the relevant warning) ever been considered?

I was actually using `rendered` to check if the block was a `Block` or `BlockSvg`, though upon reflecting I realize `rendered` can also be false for `BlockSvg` and I'm not intending to change the behavior of `BlockSvg` instances.

Since `bumpNeighbors` is implemented by `Block` and `BlockSvg`, I could just remove the warning from the `Block` implementation (and make it a no-op), but the warning might be useful in some cases.
@samelhusseini what do you think about removing the warning?

samelhusseini:
It sounds reasonable to me to remove the warning from bumpNeighbours in block.js

alschmiedt (approved):

Issue #4481, opened 2020-11-30
Remove all December 2020 deprecated functions


alschmiedt:
Remove all functions that we said we would delete before December 2020.

PR #4489, opened 2020-12-02
Remove deprecated field functions

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Part of #4481

### Proposed Changes
Removes all deprecated functions in fields.

### Test Coverage

Tested on:

### Additional Information
`Blockly.FieldTextInput.nonnegativeIntegerValidator` doesn't look like it ever got a `Blockly.utils.deprecation.warn`. So we can either: 
- Add the deprecation warning now, and add back `Blockly.FieldTextInput.numberValidator` since `nonnegativeIntegerValidator`  relies on it.
- Just remove it now without the deprecation warning. 
- Add a deprecate warn to `nonnegativeIntegerValidator` now, and remove all the functions together when they have been deleted.

I prefer option 2 or 3, so that people only have to go through and update their fields once.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -19,7 +19,6 @@ goog.require('Blockly.Events.BlockChange');
 goog.require('Blockly.Gesture');
 goog.require('Blockly.Tooltip');
 goog.require('Blockly.utils');
-goog.require('Blockly.utils.deprecation');
 goog.require('Blockly.utils.dom');
 goog.require('Blockly.utils.Rect');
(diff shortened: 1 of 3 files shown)


samelhusseini (approved):

rachel-fenichel:
I am okay with just removing everything, as long as we communicate clearly in our release notes.

I think we have enough API changes that we should do a major release now anyway.

Issue #2153, opened 2018-12-05
Needing to Initialize Generator Before Calling .blockToCode is Opaque.


BeksOmega:
### Problem statement

Calling [.blockToCode](https://github.com/google/blockly/blob/ca17e4019530c044ac828de4b3c57515572c67f4/core/generator.js#L165) does not automatically initialize the generator like [.workspaceToCode](https://github.com/google/blockly/blob/ca17e4019530c044ac828de4b3c57515572c67f4/core/generator.js#L90) does. 

### Expected Behavior

A) .blockToCode should initialize the generator.
or
B) .blockToCode should throw an informative error.

### Actual Behavior

You get opaque errors like: [Uncaught TypeError: Cannot read property 'getName' of undefined at Blockly.BlockSvg.Blockly.JavaScript.procedures_callnoreturn](https://groups.google.com/forum/#!topic/blockly/tnr9pxj81H4)

### Steps to Reproduce

1. Add below code ~ln 366 in playground.
`<xml xmlns="http://www.w3.org/1999/xhtml" id="workspaceBlocks" style="display:none">
    <block type="procedures_callnoreturn" deletable="false" id="4Tu3_y[h.I1HOr0m3.CE" x="12" y="13"></block>
  </xml>`
2. Add below code ~ln 124 (after workspace has been injected) in playground.
`var workspaceBlocks = document.getElementById("workspaceBlocks");
  Blockly.Xml.domToWorkspace(workspaceBlocks, workspace);
  Blockly.JavaScript.blockToCode(workspace.getBlockById('4Tu3_y[h.I1HOr0m3.CE'));`
3. Refresh the playground and observe the unhelpful error.

### Stack Traces

```
Uncaught TypeError: Cannot read property 'getName' of undefined
    at Blockly.BlockSvg.Blockly.JavaScript.procedures_callnoreturn (procedures.js:85)
    at Blockly.Generator.blockToCode (generator.js:185)
    at start (playground.html:126)
    at onload (playground.html:307)
```

### Operating System and Browser

Assuming this is not browser specific.

### Additional Information

This doesn't occure with all blocks, only blocks that require the generator to be initialized (e.g. proceedures).

RoboErikG:
A better error is probably safer at this point. workspaceToCode calls blockToCode for every top level block and it gets called again from scrub_ in the language specific generators and a few other places, which would cause init() to be called repeatedly if it was in both methods.

Some alternatives would be:

[...]

PR #4492, opened 2020-12-03
Add warning to blockToCode

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/2153

### Proposed Changes

Adds property to `Generator` to allow for checking whether it has been initialized and logging a warning in `blockToCode` if `init` was not called.

### Reason for Changes

Provides developers with a helpful warning when `blockToCode` is called before `init`.

### Test Coverage

Tested manually on playground with repro steps described in bug.

Tested on:
* Desktop Chrome

### Additional Information

`isInitialized` is set to `null` so that it does not affect custom generators (and incorrectly log a warning when `blockToCode` is called).

### DIFF
--- a/core/generator.js
+++ b/core/generator.js
@@ -78,6 +78,12 @@ Blockly.Generator.prototype.COMMENT_WRAP = 60;
 Blockly.Generator.prototype.ORDER_OVERRIDES = [];
 
 /**
+ * Whether the init method has been called.
+ * @type {?boolean}
+ */
+Blockly.Generator.prototype.isInitialized = null;
+
+/**
  * Generate code for all blocks in the workspace to the specified language.
  * @param {Blockly.Workspace} workspace Workspace to generate code from.
  * @return {string} Generated code.
@@ -167,6 +173,10 @@ Blockly.Generator.prototype.allNestedComments = function(block) {
  *     operator order value.  Returns '' if block is null.
  */
 Blockly.Generator.prototype.blockToCode = function(block, opt_thisOnly) {
+  if (this.isInitialized === false) {
+    console.warn(
+        'Generator init was not called before blockToCode was called.');
+  }
   if (!block) {
     return '';
   }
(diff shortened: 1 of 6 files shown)


core/generator.js
```diff
@@ -77,6 +77,12 @@ Blockly.Generator.prototype.COMMENT_WRAP = 60;
  */
 Blockly.Generator.prototype.ORDER_OVERRIDES = [];
 
+/**
+ * Whether the init method has been called.
```
moniika, line 81: I'm not sure if there's additional information I could add to help clarify how this is used.

Our core custom generators set this to `false` after creation of the singleton and then to `true` in `init`. 
`blockToCode` uses this to emit a helpful warning only if it has been set to `false`.

generators/dart.js
```diff
@@ -71,11 +71,18 @@ Blockly.Dart.ORDER_CASCADE = 15;       // ..
...
+ * @type {?boolean}
+ */
+Blockly.Dart.isInitialized = false;
+
 /**
  * Initialise the database of variable names.
  * @param {!Blockly.Workspace} workspace Workspace to generate code from.
  */
 Blockly.Dart.init = function(workspace) {
+  this.isInitialized = true;
```
moniika, line 85: Just want to call out that I'm not sure whether this ought to be at the end of the init function, rather than the start.

core/generator.js
alschmiedt: I think adding what you have is good but maybe change: only if it has been set to false --> when the generator has not been initialized.

alschmiedt (approved):

generators/dart.js
moniika: Done.

core/generator.js
moniika: Updated as per discussion offline.

Issue #4186, opened 2020-08-17
Comment calls deprecated getText method


moniika:
**Describe the bug**

Non-editable comment bubble on block emits deprecation warning.

**To Reproduce**

1. Import the folowing code in the playground:
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_if" id="L8cs$OY5`*V^_9(*FUwY" editable="false" x="88" y="38">
    <comment pinned="false" h="80" w="160">test</comment>
  </block>
</xml>
```
2. Open comment and observe console warning

**Expected behavior**

Blockly library should not call deprecated method.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser chrome

**Stack Traces**

```
Comment.prototype.getText was deprecated on August 2019 and will be deleted on December 2020.
Use block.getCommentText instead.
```

**Additional context**

`Comment.prototype.createNonEditableBubble_` calls `Warning.prototype.createBubble(this)` which uses `this.getText` which calls `Comment.getText` (and emits deprecation warning). 

Related to #4197
Causes console warnings in tests #4194

Issue #2943, opened 2019-08-30
Comments Access Warning Members They Probably Shouldn't


BeksOmega:
**Describe the bug**

Comments use warnings' UI when displaying in non-editable mode.

This means that comments need to access, and have knowledge of, things that they really shouldn't.

**Slightly Prefered Solution**

Comments should have their own UI for displaying non-editable mode.

**Alternative Solutions**
* The UI for both warnings and non-editable comments should be put into a single class so that they're less tightly coupled.
* Comments don't have special UI for non-editable mode. Instead changes to the comment aren't saved.

**Additional context**

Related: #2917

PR #4495, opened 2020-12-04
Move where a non editable bubble is created

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes: #4186 and #2943.

### Proposed Changes
There are two methods in warning that comments need in order to create a non editable bubble. Since these methods primary goal is to create a bubble this moves those methods on to the Bubble class. 

#2943 says that the comments should have their own UI which I agree with. However, since we can not use `<foreignobject>` to create a non editable bubble (see [here](https://github.com/google/blockly/blob/85874cd6962d6baf81c806a190a05be0b45f2838/core/comment.js#L254)) then whatever UI we create will most likely rely on these methods in some capacity.

### Reason for Changes
1. We want to be able to remove `getText` on the comment. We can't do that if the warning UI is using getText. (#4186)
2. General cleanup. (#2943)

### Test Coverage

Tested on:

### DIFF
--- a/core/bubble.js
+++ b/core/bubble.js
@@ -917,7 +917,7 @@ Blockly.Bubble.createNonEditableBubble = function(paragraphElement, block, iconX
       /** @type {!Blockly.WorkspaceSvg} */ (block.workspace),
       paragraphElement, block.pathObject.svgPath,
       /** @type {!Blockly.utils.Coordinate} */ (iconXY), null, null);
-  // Expose this warning's block's ID on its top-level SVG group.
+  // Expose this bubble's block's ID on its top-level SVG group.
   bubble.setSvgId(block.id);
   if (block.RTL) {
     // Right-align the paragraph.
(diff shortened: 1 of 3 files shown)


core/bubble.js
```diff
@@ -878,3 +878,57 @@ Blockly.Bubble.prototype.getRelativeToSurfaceXY = function() {
...
+ * @param {!Blockly.utils.Coordinate} iconXY The coordinate of the icon.
+ * @return {!Blockly.Bubble} The non editable bubble.
+ * @package
+ */
+Blockly.Bubble.createNonEditableBubble = function(paragraphElement, block, iconXY) {
+  var bubble = new Blockly.Bubble(
+      /** @type {!Blockly.WorkspaceSvg} */ (block.workspace),
+      paragraphElement, block.pathObject.svgPath,
+      /** @type {!Blockly.utils.Coordinate} */ (iconXY), null, null);
+  // Expose this warning's block's ID on its top-level SVG group.
```
rachel-fenichel, line 920: Change comment (warning -> bubble)

rachel-fenichel (approved):

PR #4498, opened 2020-12-04
Add a remove key map function

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Adds a function to remove key maps without needing to know the key code for the shortcut.

### Test Coverage

Tested on:

### DIFF
--- a/core/shortcut_registry.js
+++ b/core/shortcut_registry.js
@@ -104,10 +104,7 @@ Blockly.ShortcutRegistry.prototype.unregister = function(shortcutName) {
     return false;
   }
 
-  // Remove all key mappings with this shortcut.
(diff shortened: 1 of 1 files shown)


maribethb (approved):
when would you want to call this but not unregister? maybe adding an explanation in the jsdoc would help explain the api (or maybe i am missing something obvious)?

alschmiedt:
I think there are two use cases around changing the default key mappings: 
* You want to change the default key mappings and don't know/care what the old key mappings are. 
* You want to change the default key mapping for something that has multiple keys registered to it and it would be a pain to call removeKeyMappping multiple times. (ex: copy)
Happy to update the comment to explain it better : )

PR #4512, opened 2020-12-08
Fix ui event checks

### LINKED ISSUE #4499: Function mutator flyouts to not update (new events?)
**Describe the bug**

In previous versions the variable name in the functions' mutator flyout would update so that you couldn't get duplicate variable names. This is no longer working:
BadMutator
I think this is related to the new events? But that's just an instinct, I'm not sure.

**To Reproduce**

1. Add a function definition block to the workspace (either type).
2. Drag parameters into the workspace.
3. Observe how the flyout does not update.

**Expected behavior**

The variable name should change as you add new blocks, going from x->z and a -> w

moniika:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #4499
Part of #4203

### Proposed Changes

Fixes ui event checks.

### Reason for Changes

Checks were incorrectly updated before (when ui events refactor was merged).

### Test Coverage

Tested in playground. Confirmed that procedure input names updated such that there would be no duplicate variable names.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -293,7 +293,8 @@ Blockly.Procedures.updateMutatorFlyout_ = function(workspace) {
  * @package
  */
 Blockly.Procedures.mutatorOpenListener = function(e) {
-  if (!e.isUiEvent || e.element != 'mutatorOpen' || !e.newValue) {
+  if (!(e.type == Blockly.Events.BUBBLE_OPEN && e.bubbleType === 'mutator' &&
+      e.isOpen)) {
     return;
   }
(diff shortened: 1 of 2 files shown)


alschmiedt (approved):
This PR lgtm
Unrelated but it looks like mutatorOpen is still used once in the events comment. Since this event no longer exists we should update the comment.
https://github.com/google/blockly/blob/b21fa5de6317df0a6463a0dfa7109ca33ca292a2/core/events/ui_events.js#L62

PR #4517, opened 2020-12-10
Fix ios12 bug

### LINKED ISSUE #4510: Blockly playground doesn't work in iOS 12
**Describe the bug**
Blockly workspace is not working in iOS 12. I repro this in the playground as well as in our application using the latest Blockly release.

**To Reproduce**

Open https://blockly-demo.appspot.com/static/tests/playground.html on a device running iOS 12
image

**Expected behavior**
Expect to be able to drag/drop blocks, open the toolbox, scroll the workspace, etc
Observed: Unable to interact at all with the playground

**Screenshots**

All devices running iOS 12

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes #4510

### Proposed Changes
Changes 'mousedown' --> 'click'.

### Reason for Changes
**There were two problems for this bug:** 
1. [e.srcElement](https://github.com/google/blockly/blob/85874cd6962d6baf81c806a190a05be0b45f2838/core/toolbox/toolbox.js#L260) is undefined for a touchstart event. The [onClick_  method in toolbox](https://github.com/google/blockly/blob/85874cd6962d6baf81c806a190a05be0b45f2838/core/toolbox/toolbox.js#L255) relies on this and would therefore throw an error.
2. After this was fixed, the new bug became that the toolbox would close shortly after opening. 

**Why this bug was happening:**
1. We add a listener for a touchstart event when binding a mousedown event. 
2. Because we call [clearTouchIdentifier](https://github.com/google/blockly/blob/85874cd6962d6baf81c806a190a05be0b45f2838/core/toolbox/toolbox.js#L272) on the toolbox it does not throw out one of these as it should. 
3. The toolbox onClick_  method will therefore get called twice. The first event will open the category, the second will close it.

### Test Coverage
Tested on chrome.
Tested on ios12 using a simulator. 
Tested on ios14 using an iphone.

Tested on:

### Additional Information
I'm not sure why we chose to use 'mousedown' but it looks like it has been that way from when the toolbox was first create. Changing this to 'click' will change the behavior slightly when someone clicks on the toolbox to close it.

### DIFF
--- a/core/toolbox/toolbox.js
+++ b/core/toolbox/toolbox.js
@@ -236,7 +236,7 @@ Blockly.Toolbox.prototype.createContentsContainer_ = function() {
 Blockly.Toolbox.prototype.attachEvents_ = function(container,
     contentsContainer) {
   // Clicking on toolbox closes popups.
-  var clickEvent = Blockly.bindEventWithChecks_(container, 'mousedown', this,
+  var clickEvent = Blockly.bindEventWithChecks_(container, 'click', this,
       this.onClick_, /* opt_noCaptureIdentifier */ false,
       /* opt_noPreventDefault */ true);
   this.boundEvents_.push(clickEvent);
--- a/core/toolbox/toolbox.js
+++ b/core/toolbox/toolbox.js
@@ -257,8 +257,8 @@ Blockly.Toolbox.prototype.onClick_ = function(e) {
     // Close flyout.
     Blockly.hideChaff(false);
   } else {
-    var srcElement = e.srcElement;
-    var itemId = srcElement.getAttribute('id');
+    var targetElement = e.target;
+    var itemId = targetElement.getAttribute('id');
     if (itemId) {
       var item = this.getToolboxItemById(itemId);
       if (item.isSelectable()) {
--- a/tests/mocha/toolbox_test.js
+++ b/tests/mocha/toolbox_test.js
@@ -130,14 +130,14 @@ suite('Toolbox', function() {
 
     test('Toolbox clicked -> Should close flyout', function() {
       var hideChaffStub = sinon.stub(Blockly, "hideChaff");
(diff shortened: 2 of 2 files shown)


rachel-fenichel (approved):

breville:
IIRC, in the past, `mousedown` has happened instantly, while `click` has introduced two delays: waiting for the corresponding `mouseup` to complete the click, and also waiting longer to make sure that a double-click isn't happening.  The "300 ms delay" was pretty notorious and led to libraries like [this](https://github.com/ftlabs/fastclick).  Things have become more nuanced now, but is this possibly the reason why `mousedown` was used before?  Is there a chance that on some devices, this will lead to longer response times for this click?

alschmiedt:
Hello!
Thanks for the question. I think my original PR message is a bit misleading. In the original toolbox before the rewrite we listened for a [click event](https://github.com/google/blockly/blob/74260102c433032aad870458813ca0f258761676/core/components/tree/treecontrol.js#L285) to handle when a user clicked on a category and [listened for a mousedown event](https://github.com/google/blockly/blob/74260102c433032aad870458813ca0f258761676/core/toolbox.js#L160) to close the toolbox when a user clicked on the space below the categories.

In the rewrite we combined our listeners so both of the above events were now on a [mousedown event](https://github.com/google/blockly/blob/11987dc6d6aa16138b36237e8959f2a3f91eb4e2/core/toolbox/toolbox.js#L238). I think this is where the problem came from. 

In this PR I changed it so that both events happen on a click event. From our original toolbox, the only change is that closing the toolbox when a user clicks on the space below the categories, now happens on a `click` event instead of a `mousedown` event. Since we have been using the click event in the past for opening and closing the toolbox I don't think this will be a problem.

breville:
@alschmiedt Thanks for the detailed information!

Issue #4509, opened 2020-12-08
Demos > Custom Fields > Pitch Field: textAreaChange is not defined


4n70w4:
https://blockly-demo.appspot.com/static/demos/custom-fields/pitch/index.html

```
index.html:56 Uncaught ReferenceError: textAreaChange is not defined
    at HTMLTextAreaElement.onchange (index.html:56)
    at Blockly.WorkspaceSvg.setBrowserFocus (workspace_svg.js:1817)
    at Blockly.WorkspaceSvg.markFocused (workspace_svg.js:1805)
    at Blockly.TouchGesture.Blockly.Gesture.doStart (gesture.js:485)
    at Blockly.TouchGesture.doStart (touch_gesture.js:108)
    at Blockly.TouchGesture.Blockly.Gesture.handleWsStart (gesture.js:656)
    at Blockly.WorkspaceSvg.onMouseDown_ (workspace_svg.js:1552)
    at SVGGElement.wrapFunc (blockly.js:462)
```

https://github.com/google/blockly/blob/f8fc748055515e30eb6ae63106b85313824c837f/demos/custom-fields/turtle/index.html#L67

alschmiedt:
As far as I can tell this function never existed. I think the fix to this is to just remove the onchange listener for this.

PR #4518, opened 2020-12-10
Remove unused textArea from field demos

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/4509

### Proposed Changes

Removes unused textArea with unimplemented reference to `textAreaChange` and unused CSS from pitch and turtle custom field demos.

### Reason for Changes

Text area is not relevant as part of the demo and typing in the text area would cause console errors.

### Test Coverage

Ran demos locally.

Tested on:
* Desktop Chrome


alschmiedt (commented):
I think the textarea can still be used for importing/exporting code, I just think we should remove the onchange listener for it since the textAreaChange does not exist.

moniika:
> I think the textarea can still be used for importing/exporting code, I just think we should remove the onchange listener for it since the textAreaChange does not exist.

It could, but there is no implementation for importing/exporting code here. I think the textarea's can be added back along with implementation of importing/exporting code in future.

moniika:
> I think the textarea can still be used for importing/exporting code, I just think we should remove the onchange listener for it since the textAreaChange does not exist.

I hadn't noticed there was an import/export code button in the demo. I have reverted to only removing the onchange listener. Thanks for catching.

alschmiedt (approved):

PR #4541, opened 2020-12-17
Update typings

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Issues while testing the blockly-angular example due to bad typings led to realizing that the typings script was no longer running due to a reference to the deleted components directory.

I noticed that the toolbox and events subdirectories were not included in the second list of temp directories, so I added them.

### Proposed Changes

Removes missing src directory and adds toolbox and events.
Ran `npm run typings`.

### Reason for Changes

Script breaks on missing directory, so our typings are not getting updated.

### Test Coverage
Ran `npm run typings`

### Additional Information

The fact that we published a beta without seeing this problem means we probably didn't run typings automatically during release, and should.

### DIFF
--- a/scripts/gulpfiles/typings.js
+++ b/scripts/gulpfiles/typings.js
@@ -26,7 +26,6 @@ function typings() {
   const tmpDir = './typings/tmp';
   const blocklySrcs = [
     "core/",
-    "core/components",
     "core/events",
     "core/keyboard_nav",
     "core/renderers/common",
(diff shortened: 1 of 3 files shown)


scripts/gulpfiles/typings.js
```diff
@@ -67,12 +66,13 @@ function typings() {
     'typings/templates/blockly-interfaces.template',
     `${tmpDir}/core/**`,
     `${tmpDir}/core/interfaces/**`,
-    `${tmpDir}/core/components/**`,
+    `${tmpDir}/core/events/**`,
     `${tmpDir}/core/components/tree/**`,
```
samelhusseini, line 70: You can remove this one also.

Issue #4507, opened 2020-12-08
Script to release beta version


alschmiedt:
**Is your feature request related to a problem? Please describe.**
Releasing a beta version of blockly should be easy and have very little possibility of human error.

**Describe the solution you'd like**
A script that we can run using something like "npm run release:beta". 
The script should: 
1. Change the version number in package.json to newReleaseNumber-beta.0
2. npm run build
3. npm run package
4. cd dist
5. npm run publish --tag beta
** Note to do this we will need to add a publishConfig similar to how blockly samples does this.

**Describe alternatives you've considered**

**Additional context**

PR #4558, opened 2020-12-23
Update build scripts

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
- Adds a script for releasing beta (#4507)
- Adds a script for publishing
- Updates recompile script for releasing

### Proposed Changes
1. Adds a script for publishing beta version. This relies on user input about the version since I **think** we want this to be next.release.version-beta.0. However, it checks that -beta.<some number> is present and that the version doesn't already exist.
2. Publish script -- Packages, checks with the user they are on the correct branch, checks the dist folder for certain files, and then will login and publish.
3. recompile script -- Updated to ask the user whether to update, patch, minor, or major version.
4. Delete package from require cache whenever we get the version number. Not deleting it was causing us to get the incorrect version in `blockly_compressed.js`.

### Reason for Changes
1. There are few extra steps for releasing a beta version that we do not want to mess up.
2. Basically same.
3. I almost forgot to update the major version this release, which would have been a medium ish headache.

### Test Coverage

Tested on:

### Additional Information
I need to uncomment the line that will actually publish.
I will do that and merge after I have a chance to test releasing beta version, or patch, or something with new script.
I can also add a dryrun version similar to samples if people think that would be helpful.


scripts/gulpfiles/release_tasks.js
```diff
@@ -0,0 +1,182 @@
...
+// Check with the user that the version number is correct, then login and publish to npm.
+function loginAndPublish_(done, isBeta) {
+  delete require.cache[require.resolve('../../dist/package.json')]
+  var { version } = require('../../dist/package.json');
+  if(readlineSync.keyInYN(`You are about to publish blockly with the version number:${version}. Do you want to continue?`)) {
+    execSync(`npm login --registry https://wombat-dressing-room.appspot.com`, {stdio: 'inherit'});
+    execSync('npm pack', {cwd: RELEASE_DIR, stdio: 'inherit'});
+    console.log(`npm publish --registry https://wombat-dressing-room.appspot.com ${isBeta ? '--tag beta' : ''}`);
+    // TODO: Uncomment this line and test.
+    // execSync('npm publish --registry https://wombat-dressing-room.appspot.com', {cwd: RELEASE_DIR, stdio: 'inherit'});
```
moniika, line 110: Can this TODO be removed or is this something that still needs to be tested?

scripts/gulpfiles/build_tasks.js
```diff
@@ -184,6 +183,8 @@ return ${namespace};
  *     blockly_compressed.js
  */
 function buildCompressed() {
+  delete require.cache[require.resolve('../../package.json')]
+  var packageJson = require('../../package.json');
```
moniika, line 187: Is this necessary only if requiring within a function? I ask because I recently added a script for deploy demos that also uses require package.json to get version. 

Also, although this is only 2 lines, I'm thinking we might want to consider also making a helper function for gulp scripts to get the package version.

scripts/gulpfiles/git_tasks.js
```diff
@@ -63,32 +62,6 @@ function checkoutBranch(branchName) {
...
-    console.log('push-rebuild-branch: committing rebuild');
-    execSync('git commit -am "Rebuild"', { stdio: 'inherit' });
-    var branchName = getRebuildBranchName();
-    execSync('git push origin ' + branchName, { stdio: 'inherit' });
-    console.log('Branch ' + branchName + ' pushed to GitHub.');
-    console.log('Next step: create a pull request against develop.');
-    done();
-  }
-);
-
```
moniika, line 91: I just wanted to note that I'm a fan of this cleanup.

scripts/gulpfiles/release_tasks.js
```diff
@@ -0,0 +1,182 @@
...
+    case 'minor':
+      execSync(`npm --no-git-tag-version version ${majorVersion}.$(date +'%Y%m%d').0`, {stdio: 'inherit'});
+      done();
+      break;
+    case 'patch':
+      execSync(`npm --no-git-tag-version version patch`, {stdio: 'inherit'});
+      done();
+      break;
+    default:
+      done(new Error('In unexpected place in switch statement.'))
```
moniika, line 62: Instead of "In inexpected place in switch", maybe something referencing "unexpected version update type" to make it clearer what caused the error.

scripts/gulpfiles/build_tasks.js
alschmiedt: If you are not changing the version number in your script I *think* you should be fine. The problem I was running into was when I updated the version number in my script and then tried to build it would use the old cached version number in `blockly_compressed.js`.

scripts/gulpfiles/release_tasks.js
alschmiedt: This still needs to be here. I am waiting for when we need to publish a patch, or beta version so that I can test.

scripts/gulpfiles/build_tasks.js
alschmiedt: Moved this into its own function. It is used across multiple files so added a helper function file.

scripts/gulpfiles/release_tasks.js
alschmiedt: Done.

scripts/gulpfiles/release_tasks.js
moniika: ok. Could you change to "Uncomment this line and test the next time we publish a patch, or beta version." and/or maybe create an issue that can be linked.
Then others with less context know what that's in there for.

moniika (approved):

scripts/gulpfiles/release_tasks.js
alschmiedt: Just tested, so uncommenting this line and removed the npm pack above.

PR #4566, opened 2021-01-06
Update demo deploy script and fix gcloudignore file

### LINKED ISSUE #4565: Fix .gcloudignore file
**Describe the bug**

It has come to attention that our process for deploying demos does not upload the .gcloudignore file. Recently, after trying to use the .gcloudignore, it was found that it is out of date and excludes some files that are necessary for the playground demo to work.

**To Reproduce**
 
1. Include the .gcloudignore file when uploading demos
2. Open playground demo
3. Blockly workspace does not load and console logs indicate that there are missing files (from goog/base)

**Expected behavior**

The gcloudignore file should be used in order to not upload certain unecessary files (such as dev scripts).

**Additional context**

This issue was found after creating a script to automate demo deployment and including gcloudignore as part of the deployment.

moniika:
## The basics

- [x] I branched from develop

### Resolves
 
Fixes demo deploy (previously did not work for playground demo because .gcloudignore was wrong) and Fixes #4565

### Proposed Changes

- Updates .gcloudignore
- Udates logic in deploy demos to use appengine directory from HEAD rather than copying local files

### Reason for Changes

Prevents unnecessary files from being uploaded to appengine and fixes automated demo deployment for playground.

### Test Coverage

1. Ran locally with minor local updates to code:
   - deploying to version appended with "-test" 
   - copying local .glcloudignore (since the .gcloudignore copied from HEAD would not have the corrected updates)
2. Confirmed that playground loads.

### Additional Information

Upon further investigation, it was also found that the test blocks toolbox and advanced playground do not work deployed and bug #4567 was filed to track it.


appengine/.gcloudignore
```diff
@@ -2,14 +2,12 @@
...
 *.komodoproject
-/deploy
+deploy
 /static/appengine/
-/static/closure/
 /static/demos/plane/soy/*.jar
 /static/demos/plane/xlf/
 /static/externs/
 /static/msg/json/
-/static/node_modules/
```
alschmiedt, line 12: Why remove the git ignore for the node_modules?

alschmiedt (approved):

appengine/.gcloudignore
moniika: The files copied using the command `git archive HEAD` are those committed to the repository and do not include the node_modules directory (removing it reduces confusion since these files don't exist to be ignored in the first place).

PR #4571, opened 2021-01-08
Process xml tags in an explicit order in domToBlockHeadless_

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #2135

### Proposed Changes

- Processes xml tags in an explicit order.
- Adds npm script `npm run test:generators` to allow for running only generator tests (which were used heavily in testing this change)

### Reason for Changes

- This change allows for us to guarantee that tags are processed in a specific order and ensures that a difference in order of xml tags does not affect the loaded block.
- After this change, we have the ability to add logic between processing specific tags.

### Test Coverage

Ran tests.

Tested on:

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -631,40 +631,25 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
   var id = xmlBlock.getAttribute('id');
   block = workspace.newBlock(prototypeName, id);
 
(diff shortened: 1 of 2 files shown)


moniika:
Applied comments from https://github.com/google/blockly/pull/4461

rachel-fenichel (approved):

PR #4574, opened 2021-01-08
Be consistent in how we access global properties on Blockly.utils.global

samelhusseini:
## The basics

- [x] I branched from develop

### Resolves

This doesn't fix a specific problem per se, but it makes things consistent. This is ensuring we always access properties / methods on `Blockly.utils.global` using fully qualified strings (rather than dot syntax), that way the closure compiler will never rename these fields.

The reason these instances haven't caused an issue is:
- navigator: Elsewhere in the code we use the fully qualified name, so the closure compiler must pick up that that field shouldn't be renamed. 
- getComputedStyle: Unsure why that is. I would have expected it to but it didn't, maybe closure is doing something special for an object with the name `global`?

### Proposed Changes

Use fully qualified names.

### Reason for Changes

Consistency 

### Test Coverage

Tested on:

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -434,7 +434,7 @@ Blockly.utils.is3dSupported = function() {
   }
   // CC-BY-SA Lorenzo Polidori
   // stackoverflow.com/questions/5661671/detecting-transform-translate3d-support
-  if (!Blockly.utils.global.getComputedStyle) {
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):
Approve if travis is okay.

Issue #4575, opened 2021-01-08
Block.interpolate_ calls the FieldLabel constructor directly instead of through the registry


rachel-fenichel:
**Describe the bug**

The `interpolate_` function in block.js [calls](https://github.com/google/blockly/blob/master/core/block.js#L1684) `new Blockly.FieldLabel(element)` instead of using `Blockly.FieldRegistry.fromJson.

All other fields are created through the registry.

This is only in the case where there is a bare string in the block definition, which is then interpreted as a label.

For consistency, and in case someone provides their own implementation of field label, we should use the registry.

Found by looking at missing requires.

**Expected behavior**

All fields are instantiated through the registry.

**Additional context**

Desired fix: call `fieldRegistry.fromJson`, passing in an object that has the correct information. I think it would need to have the contents of `element` as `'text'` and `'label'` as `'type'`.

BeksOmega:
I can grab this =) And I'll write some tests for the interpolate_ while I'm at it.

PR #4585, opened 2021-01-13
Refactor interpolate_ into multiple functions & remove direct new Field call

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #4575

### Proposed Changes

Refactors the interpolate_ function to be more testable and consistent. Removes the call to new Blockly.FieldLabel() in the process.

This PR breaks the interpolate_ function into 5 new functions, so that each piece can be tested individually:
  * validateTokens_(): Ensures that all of the args are referenced, that all references are in the correct range, and that there are no duplicate references.
  * interpolateArgs_(): Replaces all arg references and strings with field/input json definitions. This means that `interpolate_` can invariantly expect all elements of the `elements` array to be objects, which simplifies the logic in `interpolate_` greatly.
  * inputFromJson_(): Appends an input to the block and returns it, given a json representation of an input. 
  * fieldFromJson_(): Returns a field instance given a json representation of a field. Note that every element whose type is not one of the hard-coded input types gets routed here. In the case that a field of the given `element['type']` is not registered it calls, itself recursively with the `element['alt']` value to try to create a field. If a field cannot be created, this returns null.
  * stringToFieldJson_(): Returns a JSON representation of a label field, with the given string as its text. If when `trim()`ed the string evaluates to an empty string, this returns null. The null value works its way back up to `interpolate_` where the element is ignored.

### Reason for Changes

After looking at #4575 I realized that the `interpolate_` function could use some serious cleanup.

Firstly the original function was very hard to read. For example the original call pointed out by the issue isn't even the real place that strings were being handled. That location was only handling strings in `'alt'` properties. Normally strings were being handled by [input.insertFieldAt](https://github.com/google/blockly/blob/edd475f37b02888819b87ddfca1d6eeebe68898a/core/input.js#L103) (which still needs to be fixed) via the logic [here](https://github.com/google/blockly/blob/edd475f37b02888819b87ddfca1d6eeebe68898a/core/block.js#L1677) and [here](https://github.com/google/blockly/blob/edd475f37b02888819b87ddfca1d6eeebe68898a/core/block.js#L1725). This is just one example of the logic being messy, but the whole thing was basically a rats-nest of nesting that needed to be detangled.

And secondly, the entire function was based on changing state, which made it very hard to test. Breaking the large function up in to smaller stateless functions makes it unit-testable.

### Breaking Changes

I did an intense review over two days to make sure I didn't make any changes that would break block initialization unintentionally.

Here is a list of the things I *intentionally* broke:
1) **You can no longer create inputs as alts.** I believe this is fine because the [documentation](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#args) only refers to fields as alts, not inputs, so there is no "contract" between us and the developer to support this.
```
  {
    'type': 'test1',
    'message0': 'test: %1',
    'args0': [
      {
        'type': 'field_undefined',
        'alt': {
          'type': 'input_value',
          'name': 'ALT_INPUT'
        },
      }
    ]
  },
```

Before:
alt-input

After:
no-alt-input

2) **String args are now trimmed.** I believe this is fine because it makes them more consistent with normal strings in the message.

```
  {
    'type': 'test2',
    'message0': 'test: %1',
    'args0': [
      '    arg string    '
    ]
  },
```

Before:
string-no-trimming

After:
string-trimming

3) **If a string arg evaluates to the null string after being trimmed it is ignored.** Again this aids in consistency.
```
  {
    'type': 'test3',
    'message0': 'test: %1',
    'args0': [
      '        '
    ]
  },
```

Before:
string-no-eliminating

After:
string-eliminating

4) **Field alt strings are now trimmed.** Again this aids in consistency.
```
  {
    'type': 'test4',
    'message0': 'test: %1',
    'args0': [
      {
        'type': 'field_undefined',
        'alt': '    arg string    ',
      }
    ]
  },
```

Before:
alt-no-trimming

After:
alt-trimming

5) **If a field alt string evaluates to the null string after being trimmed it is ignored.** Again this aids in consistency.
```
  {
    'type': 'test5',
    'message0': 'test: %1',
    'args0': [
      {
        'type': 'field_undefined',
        'alt': '        ',
      }
    ]
  }
```

Before:
alt-no-eliminating

After:
alt-eliminating

6) **The call to new Blockly.FieldLabel() has been removed.** This was the intention of the original issue, but could cause problems if someone is relying on this behavior to get the default label implementation instead of their custom one in the case of an alt-field specifically. However I believe this is fine because that use-case is extremely unlikely.

### Test Coverage

I wrote tests for all of the new functions, excepting stringToFieldJson, because that is covered by the other tests =)

I also went through all the test blocks and built-in blocks and made sure they intialized themselves correctly.

Tested on:
 * Desktop Chrome

### Additional Information

There is still a call in [insertFieldAt](https://github.com/google/blockly/blob/edd475f37b02888819b87ddfca1d6eeebe68898a/core/input.js#L103) that needs to be changed to use the field registry.

There is a bug where registered fields that don't begin with the `'field_'` prefix will not make the block add a dummy input at the end of the block. This obviously isn't affecting anyone currently, because they would have figured it out when their fields didn't show up, but it would be good to fix. I've created a skipped test to document this.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1641,6 +1641,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign,
       case 'input_statement':
       case 'input_dummy':
         var input = this.inputFromJson_(element, warningPrefix);
+        // Should never be null, but just in case.
         if (input) {
           for (var j = 0, tuple; (tuple = fieldStack[j]); j++) {
             input.appendField(tuple[0], tuple[1]);
@@ -1658,6 +1659,14 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign,
   }
 };
 
+/**
+ * Validates that the tokens are within the correct bounds, with no duplicates,
+ * and that all of the arguments are referred to. Throws errors if any of these
+ * things are not true.
+ * @param {!Array<string|number>} tokens An array of tokens to validate
+ * @param {number} argsCount The number of args that need to be referred to.
+ * @private
+ */
 Blockly.Block.prototype.validateTokens_ = function(tokens, argsCount) {
   var visitedArgsHash = [];
   var visitedArgsCount = 0;
@@ -1683,6 +1692,18 @@ Blockly.Block.prototype.validateTokens_ = function(tokens, argsCount) {
   }
 };
 
+/**
+ * Inserts args in place of numerical tokens. String args are converted to json
+ * that defines a label field. If necessary an extra dummy input is added to
+ * the end of the elements.
+ * @param {!Array<!string|number>} tokens The tokens to interpolate
+ * @param {!Array<!Object|string>} args The arguments to insert.
+ * @param {string} lastDummyAlign The alignment the added dummy input should
+ *     have, if we are required to add one.
+ * @return {!Array<!Object>} The JSON definitions of field and inputs to add
+ *     to the block.
+ * @private
+ */
 Blockly.Block.prototype.interpolateArguments_ =
     function(tokens, args, lastDummyAlign) {
       var elements = [];
@@ -1716,6 +1737,16 @@ Blockly.Block.prototype.interpolateArguments_ =
       return elements;
     };
 
+/**
+ * Creates a field from the json definition of a field. If a field with the
+ * given type cannot be found, this attempts to create a different field using
+ * the 'alt' property of the json definition (if it exists).
+ * @param {{alt:(string|undefined)}} element The element to try to turn into a
+ *     field.
+ * @return {!Blockly.Field|null} The field defined by the JSON, or null if one
+ *     couldn't be created.
+ * @private
+ */
 Blockly.Block.prototype.fieldFromJson_ = function(element) {
   var field = Blockly.fieldRegistry.fromJson(element);
   if (!field && element['alt']) {
@@ -1728,6 +1759,16 @@ Blockly.Block.prototype.fieldFromJson_ = function(element) {
   return field;
 };
 
+/**
+ * Creates an input from the json definition of an input. Sets the input's check
+ * and alignment if they are provided.
+ * @param {!Object} element The JSON to turn into an input.
+ * @param {string} warningPrefix The prefix to add to warnings to help the
+ *     developer debug.
+ * @return {!Blockly.Input|null} The input that has been created, or null if one
+ *     could not be created for some reason (should never happen).
+ * @private
+ */
 Blockly.Block.prototype.inputFromJson_ = function(element, warningPrefix) {
   var alignmentLookup = {
     'LEFT': Blockly.ALIGN_LEFT,
@@ -1748,6 +1789,7 @@ Blockly.Block.prototype.inputFromJson_ = function(element, warningPrefix) {
       input = this.appendDummyInput(element['name']);
       break;
   }
+  // Should never be hit because of interpolate_'s checks, but just in case.
   if (!input) {
     return null;
   }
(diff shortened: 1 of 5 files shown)


tests/mocha/utils_test.js
```diff
@@ -23,73 +23,124 @@ suite('Utils', function() {
...
+    setup(function() {
+      this.assertInterpolation = function(string, tokens) {
+        chai.assert.deepEqual(
+            Blockly.utils.tokenizeInterpolation(string), tokens);
+      };
+    });
 
-      tokens = Blockly.utils.tokenizeInterpolation('Hello%%World');
-      chai.assert.deepEqual(tokens, ['Hello%World'], 'Escaped %.');
+    suite('Basic', function() {
```
maribethb, line 33: I like that these are broken into separate tests now, but I don't think you are gaining much by using the assertInterpolation function. In general, unless the series of assertions are really complex, I think tests are better off being very plain and explicit about their assertions for ease of reading and so that you can tell a test is correct at first glance. And that way you can add individual error messages (though I will say that some of the original error messages of "Interpolations." are comically unhelpful).

tests/mocha/utils_test.js
```diff
@@ -23,73 +23,124 @@ suite('Utils', function() {
...
+        this.assertInterpolation('Hello%1World', ['Hello', 1, 'World']);
+      });
+
+      test('Multi-digit number interpolation', function() {
+        this.assertInterpolation(
+            '%123Hello%456World%789', [123, 'Hello', 456, 'World', 789]);
+      });
+
+      test('Crazy interpolation', function() {
+        // No idea what this is supposed to tell you if it breaks. But might
```
maribethb, line 62: Percent signs in interpolations, I think? I could see a legitimate need for a % in a message

maribethb (commented):
Wow, you're right, this was a bit of a mess. I'd rather someone more familiar with the existing behavior signed off on the breaking changes, but I have some minor comments for the code itself.

core/block.js
```diff
@@ -1630,111 +1630,201 @@ Blockly.Block.prototype.mixin = function(mixinObj, opt_disableCheck) {
...
-  var indexDup = [];
-  var indexCount = 0;
-  var elements = [];
+  this.validateTokens_(tokens, args.length);
+  var elements = this.interpolateArguments_(tokens, args, lastDummyAlign);
+
+  // An array of [field, fieldName] tuples.
+  var fieldStack = [];
+  for (var i = 0, element; (element = elements[i]); i++) {
+    switch (element['type']) {
```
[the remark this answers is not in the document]
BeksOmega: I'm not sure either haha. My understanding is that since this is parsing JSON (which comes from outside the "compiled" blockly implementation) it's supposed to use the bracket notation. If it doesn't closure could rename this to something like `a.a` which wouldn't work with a `{ 'type': 'input_statement'}` JSON object. But I'm only an 80% on that =)

tests/mocha/utils_test.js
BeksOmega: The individual error messages thing is a good point! I'll have to keep that in mind in the future.

tests/mocha/utils_test.js
BeksOmega: Ohhh you're right. There is a test for an escaped %, but not one where the escaped percent preceeds a number! I'll add a separate test for that so that it's easier to debug hehe

maribethb (approved):

core/block.js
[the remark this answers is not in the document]
maribethb: I think you're right! Now that you mention it, I think I had the exact same conversation with Abby about toolbox json, so maybe I will remember this next time. Thanks!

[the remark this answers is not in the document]
BeksOmega: Hopefully I can remember next time too haha. After working on interpolate_ I realized I forgot to use the bracket notation in the type hierarchy D: Shouldn't be too hard to fix though hehe

rachel-fenichel:
The changes to alt and whitespace behaviour look reasonable to me.

Issue #4567, opened 2021-01-06
Advanced Playground and Test Blocks toolbox hosted on appengine do not load


moniika:
**Describe the bug**
The block test toolbox (on the playground) and advanced playground rely on having run `npm install` in order to load. When opening either of these on apengine hosted demo site, they do not load and emit the alert "You need to run 'npm install' in order to use the test blocks." or "You need to run 'npm install' in order to use this playground.".

**To Reproduce**

Go to either:
- https://blockly-demo.appspot.com/static/tests/playground.html?dir=ltr&toolbox=test-blocks
- https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html

**Expected behavior**

The advanced playground should not be uploaded to appengine.
The playground should either not be uploaded to appengine or the test toolbox should not be allowed to be selected from the dropdown of toolboxes (this second option could be tricky and would be best if it only affected appengine hosted instances of the page).

**Additional context**

We already use github.io for testing on playground and the playgrounds do not seem linked in the index page, so not uploading them to appengine entirely seems like the most straightforward option.

alschmiedt:
Alternatives to removing it are: 
1. Upload the node module needed for block tests.
2. Include the script tag for test blocks and try to load that way before trying to load from node_modules.
I think 1 might be the easiest solution.

samelhusseini:
Remaining tasks: 
- [ ] Retrieve package.json from HEAD
- [ ] Don't rely on local node_modules for playground files, instead fetch them from unpkg.

PR #4591, opened 2021-01-15
Update appengine_tasks to include playground dependencies

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/4567

### Proposed Changes

Include playground deps from node_modules in the set of files deployed to appengine.

### Reason for Changes

Fix deployed playgrounds.

### Test Coverage

Tested on:

### DIFF
--- a/scripts/gulpfiles/appengine_tasks.js
+++ b/scripts/gulpfiles/appengine_tasks.js
@@ -62,7 +62,7 @@ function copyPlaygroundDeps() {
       './node_modules/@blockly/theme-modern/dist/index.js',
       './node_modules/@blockly/block-test/dist/index.js',
   ];
-  return gulp.src(playgroundDeps).pipe(gulp.dest(demoStaticTmpDir));
+  return gulp.src(playgroundDeps, {base: '.'}).pipe(gulp.dest(demoStaticTmpDir));
 }
 
 /**
--- a/scripts/gulpfiles/appengine_tasks.js
+++ b/scripts/gulpfiles/appengine_tasks.js
@@ -54,6 +54,18 @@ function copyAppengineSrc() {
 }
 
 /**
+ * Copies playground deps into deploy directory.
+ */
+function copyPlaygroundDeps() {
+  const playgroundDeps = [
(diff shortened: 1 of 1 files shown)


scripts/gulpfiles/appengine_tasks.js
```diff
@@ -53,6 +53,18 @@ function copyAppengineSrc() {
   return gulp.src(appengineSrc).pipe(gulp.dest(demoTmpDir));
 }
 
+/**
+ * Copies playground deps into deploy directory.
```
moniika, line 57: This copy differs from the other `copyStaticSrc` and `copyAppengineSrc` since it copies uncommited local files. I think it's worth mentioning in the JsDoc about that (and perhaps mention that npm install needs to have been run).

If possible, I'd like for us to check that there is node_modules file somewhere in this step because otherwise it will just silently "fail" if someone was to have cleaned their node modules before running.

moniika (approved):

scripts/gulpfiles/appengine_tasks.js
samelhusseini: Hmm, all good points. If the goal is to not have it depend on the current state of the repo, perhaps we can directly acquire the files we're interested in from unpkg.

samelhusseini: Also I noticed that the script reads from the local package.json to determine the version, rather than the one in HEAD. Should that also be switched over?

moniika: The current deploy demo script uses files in HEAD, rather than local files because it was a direct translation of the steps we have done in the past. I want to note that we could consider changing it.
Assuming we're not:
- unpkg: +1 to using unpkg, assuming it's not too complicated (not necessary for this PR, but we could file an issue to add later)
- `package.json`: good catch. I think it probably would be better to use the version in HEAD

samelhusseini: In the interest of not letting this sit, I'm going to throw this in, and update the associated but with the remaining tasks.

PR #4601, opened 2021-01-23
Remove last FieldLabel call and fix adding dummy inputs to json

### LINKED ISSUE #4575: Block.interpolate_ calls the FieldLabel constructor directly instead of through the registry

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #4575 

### Proposed Changes

1) Changes input.insertFieldAt to use the field registry.
2) Fixes last dummy inputs only being added for fields where the key is prefixed with 'field_'

Changes are kept in separate commits to make reviewing easier =)

### Reason for Changes

1) Using the registry allows people to override the default field for strings, as referenced in #4575 
2) Not all fields are prefixed with 'field_' but we should add the last dummy input regardless.

### Test Coverage

Most of the tests were already written :D But for insertFieldAt I made a test to make sure if field_label gets overwritten in the registry, the new field gets used. And for interpolation I added a test that fields prefixed with input_ also force dummy inputs to be created.

Again, I went through and checked all of the core blocks and test blocks to make sure they get created correctly.

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1635,25 +1635,21 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign,
   // An array of [field, fieldName] tuples.
   var fieldStack = [];
   for (var i = 0, element; (element = elements[i]); i++) {
-    switch (element['type']) {
-      case 'input_value':
-      case 'input_statement':
-      case 'input_dummy':
-        var input = this.inputFromJson_(element, warningPrefix);
-        // Should never be null, but just in case.
-        if (input) {
-          for (var j = 0, tuple; (tuple = fieldStack[j]); j++) {
-            input.appendField(tuple[0], tuple[1]);
-          }
-          fieldStack.length = 0;
+    if (this.isInputKeyword_(element['type'])) {
+      var input = this.inputFromJson_(element, warningPrefix);
+      // Should never be null, but just in case.
+      if (input) {
+        for (var j = 0, tuple; (tuple = fieldStack[j]); j++) {
+          input.appendField(tuple[0], tuple[1]);
         }
-        break;
+        fieldStack.length = 0;
+      }
+    } else {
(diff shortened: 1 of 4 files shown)


BeksOmega:
Weird, it looks like the build never ran.

maribethb:
> Weird, it looks like the build never ran.

Sorry, forgot to update you. We are having issues with Travis, but on Monday if the problems persist I will manually run the build on your PR and then merge it for you.

maribethb (approved):

maribethb:
Hey @BeksOmega we have GitHub Actions set up now to replace travis. Could you either rebase or merge from develop so that it pulls in the new workflow stuff (should have a folder `.github/workflows` after the merge) so that the automated tests will run on this PR? Thanks, and sorry for the hassle! 

(I tried to rebase your pr myself but didn't want to overwrite your commits)

BeksOmega:
Cool should be gtg now! And don't every worry about overwriting my commits hehe, just do whatever is easiest for you. If it's easiest for you to rebase, go right ahead. Or if it's easiest for me to rebase, I'm always happy to =)

PR #4619, opened 2021-02-05
Migrates the blocks directory to ES6 with goog.module

samelhusseini:
I've migrated the blocks directory to ES6 with goog.module.
Use this PR to have a conversation / ask questions about various aspects of the migration.

Note: 
- like ES6 modules, goog.module means no IE 11 support for uncompressed / playground
- Also, goog.modules are loaded asynchronously (as needed), and like ES6 modules are subject to CORS, meaning our file:/// access will have to end, and we'll have to start a simple web server to run the playground. @blockly/dev-scripts can come in handy here.
- The blocks modules don't export anything, hence no `exports = ...`, we can expect other modules in core to export fields / methods / classes

Additional follow-up tasks:
- Strip out and clean up the deprecated provide (eg: Blockly.Blocks.colour) and the deprecated HUE export (eg: Blockly.Constants.Colour.HUE = 20;)
- Update eslint config so it reflects an ES6 environment for blocks, and ES5 for core (during migration)
- Integrate @bloockly/dev-scripts to run a simple server to load the playground


BeksOmega:
Hi Sam! This looks really cool :D Is this prep for moving the block definitions to blockly-samples?

rachel-fenichel:
Some high level questions/thoughts:
- Can I get a link to the documentation where it says that module doesn't work/will not work in IE11 or uncompressed mode? My understanding was that using the debug loader would still make it load synchronously.
- How did you pick the module names? In particular, why are procedures different?
- How did we end up with an extra 4K lines of closure code from this?
- Is the compiler now compiling from es6 into es5?

@BeksOmega -- we're investigating closure modules as a possible intermediate step on the way to es6 or typescript.

PR #4627, opened 2021-02-09
Metrics refactor

### LINKED ISSUE #3746: Create a Public Metrics Method

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#3746

### Proposed Changes
Pulls out getMetrics logic into a metrics manager and splits up calculating metrics into smaller methods. 

### Reason for Changes
- Once we add the ability to register metrics, developers can easily override certain methods instead of having to override all of getMetrics.(Like in the case of the continuous toolbox).
- Internal methods can call the appropriate method on the metrics manager allowing them to only calculate the metrics they need. 
- Increases readability by having well documented methods.

### Test Coverage

Tested on:

### Additional Information
Follow up prs include: 
- Metrics tests
- Refactoring blockly to use the new methods where applicable.
- Add the metrics manager to the registry.


core/metrics_manager.js
```diff
@@ -0,0 +1,385 @@
...
+  var flyoutDimensions = this.getDimensionsPx_(this.workspace_.getFlyout(opt_own));
+  return new Blockly.utils.Size(
+      flyoutDimensions.width, flyoutDimensions.height);
+};
+
+/**
+ * Gets the width, height and position of the toolbox on the workspace in pixel
+ * coordinates. Returns 0 for the width and height if the workspace has a simple
+ * toolbox instead of a category toolbox. To get the width and height of a
+ * simple toolbox see getFlyoutMetrics.
```
moniika, line 187: I think there's an annotation you can use like `@see` and/or `@link` that you can use for this reference to `getFlyoutMetrics`

core/metrics_manager.js
```diff
@@ -0,0 +1,385 @@
...
+   * @protected
+   */
+  this.workspace_ = workspace;
+};
+
+/**
+ * Describes the width, height and location of the toolbox on the main
+ * workspace.
+ * @typedef {{
+ *            width:number,
```
moniika, line 40: We aren't consistent in our code base, but I think there should be a space after the `:` in these typedefs. So, for example, this would change to `width: number`.

core/metrics_manager.js
```diff
@@ -0,0 +1,385 @@
...
+ * @return {!Blockly.MetricsManager.ContainerRegion} The
+ *     metrics for the content container.
+ * @public
+ */
+Blockly.MetricsManager.prototype.getContentMetrics = function(
+    opt_getWorkspaceCoordinates) {
+  var scale = opt_getWorkspaceCoordinates ? this.workspace_.scale : 1;
+  var viewSize = this.getViewMetrics(false);
+  var contentDimensions = this.getContentDimensions_(viewSize);
+
```
moniika, line 313: Since we're allowing for getting only content metrics, I think we should refactor so that we only get view metrics if it's needed to compute content dimensions.

core/metrics_manager.js
```diff
@@ -0,0 +1,385 @@
...
+ * .flyoutHeight: Height of the flyout if it is always open.  Otherwise zero.
+ * .toolboxPosition: Top, bottom, left or right. Use TOOLBOX_AT constants to
+ *     compare.
+ * @return {!Blockly.utils.Metrics} Contains size and position metrics of a top
+ *     level workspace.
+ * @public
+ */
+Blockly.MetricsManager.prototype.getMetrics = function() {
+  var contentMetrics = this.getContentMetrics();
+  var viewMetrics = this.getViewMetrics();
```
moniika, line 356: One concern I have is that we compute `viewMetrics` twice in `getMetrics` since there's a call for `getViewMetrics` inside of `getContentMetrics`. 
Maybe we can add an optional parameter to contentMetrics so that we can pass viewmetrics if we already have that computed.

moniika (approved):

core/bubble.js
```diff
@@ -517,12 +513,12 @@ Blockly.Bubble.prototype.layoutBubble_ = function() {
  * workspace (what percentage of the bubble is visible).
  * @param {!{x: number, y: number}} relativeMin The position of the top-left
  *     corner of the bubble relative to the anchor point.
- * @param {!Blockly.utils.Metrics} metrics The metrics of the workspace the
- *     bubble will appear in.
+ * @param {!Blockly.MetricsManager.ContainerRegion} viewMetrics The metrics of the
```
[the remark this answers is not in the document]
alschmiedt: done

core/metrics_manager.js
alschmiedt: Ok, I added it but couldn't find a lot of information on how it shows up in the docs so I am not sure if I did it correctly. Let me know if you think it should be changed.

core/metrics_manager.js
alschmiedt: Ok done.

core/metrics_manager.js
alschmiedt: done.

PR #4642, opened 2021-02-19
Move bind/unbind events functions to new file, with example usage.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Resolves some dependency issues.

### Proposed Changes

Moves three functions and one typedef onto a newly created namespace:
- `Blockly.bindEvent_` -> `Blockly.browserEvents.bind`
- `Blockly.unbindEvent_` -> `Blockly.browserEvents.unbind`
- `Blockly.bindEventWithChecks_` -> `Blockly.browserEvents.conditionalBind`
- `Blockly.EventData` -> `Blockly.browserEvents.Data`

For backwards compatibility, I left aliases in blockly.js for these three functions. I plan to move all internal uses to the fully qualified name.

#### Behavior Change?

No change in behaviour

### Reason for Changes

We have circular dependecies because blockly.js requires most parts of blockly (workspaces, etc) but defines key utility functions that are used throughout the codebase.

The functions are used all over the place and only require one thing (`Blockly.Touch`). Moving the functions to their own namespace allows users to require them directly, rather than requiring the whole of Blockly.

### Additional Information

This is a sample to show what the change will look like. I'd like to nail down names for the namespace and function before I change all of the uses.


rachel-fenichel:
@maribethb I've updated the names and comments, PTAL

samelhusseini (approved):

PR #4648, opened 2021-02-23
Refactor flyout and toolbox

### LINKED ISSUE #4594: Toolbox and flyout metrics should report appropriate height and width
getMetrics returns the following information for a flyout and a toolbox: 

- flyoutWidth,
- flyoutHeight,
- toolboxWidth,
- toolboxHeight,

When there is a vertical toolbox or flyout the height is equal to 0. Similarly, when there is a horizontal toolbox or flyout the width is equal to 0.

For the toolbox and flyout metrics to accurately report the actual height and width of the flyout.

Keep it as is and continue to return 0.

**Additional context**
This would mostly include looking into where this value is being used and making sure we don't break anything by updating this.
I don't think this would break a lot of people, but it is worth it to note that this is an api change.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#4594

### Proposed Changes
- Changes the toolbox metrics to properly report it's height and width, no matter the orientation. 
- Updates the code to not rely on the toolbox returning 0 for the height or width.

#### Behavior Before Change
workspace.getMetrics for a horizontal toolbox:
`toolboxHeight`: HEIGHT_OF_TOOLBOX
`toolboxWidth`: 0

workspace.getMetrics for a vertical toolbox: 
`toolboxHeight`: 0
`toolboxWidth`: WIDTH_OF_TOOLBOX

#### Behavior After Change
workspace.getMetrics for a horizontal toolbox:
`toolboxHeight`: HEIGHT_OF_TOOLBOX,
`toolboxWidth`: WIDTH_OF_TOOLBOX

workspace.getMetrics for a vertical toolbox: 
`toolboxHeight`: HEIGHT_OF_TOOLBOX,
`toolboxWidth`: WIDTH_OF_TOOLBOX


core/flyout_horizontal.js
```diff
@@ -145,9 +145,9 @@ Blockly.HorizontalFlyout.prototype.getY = function() {
   // If this flyout is not the trashcan flyout (e.g. toolbox or mutator).
   if (this.targetWorkspace.toolboxPosition == this.toolboxPosition_) {
     // If there is a category toolbox.
-    if (targetWorkspaceMetrics.toolboxHeight) {
+    if (this.targetWorkspace.getToolbox()) {
       if (this.toolboxPosition_ == Blockly.TOOLBOX_AT_TOP) {
-        y = targetWorkspaceMetrics.toolboxHeight;
+        y = /** @type {number} */ (targetWorkspaceMetrics.toolboxHeight);
```
maribethb, line 150: This cast seemed weird to me, and it's because the type for toolbox width/height in Blockly.utils.metrics says

```
/**
 * Width of the toolbox, if it exists.  Otherwise zero.
 * @type {number|undefined}
 */
```

The comment and the type don't match up. If it is indeed 0 if the toolbox doesn't exist, then we could remove the undefined from the type, and you wouldn't have to cast here. That seems better to me.

If it really can be undefined sometimes, then you could also check to make sure that the value is defined which would remove the need for a cast (the old code was checking it was defined, so the type refinement meant it could only be a number at this point). Maybe the cast is fine, but in the end we are basically converting this to a runtime check. It *shouldn't* be undefined, if `getToolbox` isn't undefined, but technically we haven't actually verified that at this point. I can see arguments for it either way, I just got immediately suspicious when I see a cast to a number for something that definitely should have already been a number.

maribethb (approved):

core/flyout_horizontal.js
alschmiedt: The problem here was that the flyout and the mutator were returning a subset of the metrics object and so we had to say certain fields on the metrics type were `number|undefined`. In this case we could assume that they defined because we were getting the metrics for the main workspace and not the flyout or mutator workspace. However, I agree that is not ideal. I just updated flyouts and mutators so they return the same metrics as the main workspace. This means that we can get rid of this cast.

maribethb (approved):

PR #4650, opened 2021-02-23
Erratic scroll fix

### LINKED ISSUE #4636: Block moves erratically when dragged a small amount
**Describe the bug**

Sometimes when I drag a block a small amount, it jumps erratically. I think this has to do with a weird interaction between the block drag surface and the workspace drag surface.

**To Reproduce**
1. Open the Playground
2. Drag a block to roughly the middle of the space
3. Scroll the workspace so that the block is in the top left corner of the workspace
4. Drag the block a small amount and release

**Expected behavior**

Expected: the block should stay where I release it.
Observed: the block jumps a seemingly random distance towards the center of the workspace.

**Screenshots**
Feb-16-2021 10-26-21

 - OS: Mac Mojave
 - Browser: Chrome
 - Version 88

**Stack Traces**
No errors

**Additional context**

moniika:
## The basics

- [x] I branched from develop

### Resolves

Fixes #4636

### Proposed Changes

Removes insertion marker from content size computation (cause of erratic scroll bug)

#### Behavior Before Change

Erratic scroll behavior in `contentResize` logic, as described in #4636, where no scroll was expected.

#### Behavior After Change

No erratic jump in scenario described by #4636.

### Reason for Changes

Fixes incorrect behavior (insertion marker was causing issues in the math for scrolling)

### Test Coverage

Extensive manual testing in playground.

* Desktop Chrome

### Additional Information

Broken off from #4643
This was a tricky change.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1726,6 +1726,9 @@ Blockly.WorkspaceSvg.prototype.getBlocksBoundingBox = function() {
 
   // Start at 1 since the 0th block was used for initialization.
   for (var i = 1; i < topElements.length; i++) {
+    if (topElements[i].isInsertionMarker_) {
+      continue;
+    }
     var blockBoundary = topElements[i].getBoundingRectangle();
     if (blockBoundary.top < boundary.top) {
       boundary.top = blockBoundary.top;
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1726,10 +1726,11 @@ Blockly.WorkspaceSvg.prototype.getBlocksBoundingBox = function() {
 
   // Start at 1 since the 0th block was used for initialization.
   for (var i = 1; i < topElements.length; i++) {
-    if (topElements[i].isInsertionMarker_) {
+    var topElement = topElements[i];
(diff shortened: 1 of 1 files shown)


core/workspace_svg.js
```diff
@@ -1726,7 +1726,11 @@ Blockly.WorkspaceSvg.prototype.getBlocksBoundingBox = function() {
 
   // Start at 1 since the 0th block was used for initialization.
   for (var i = 1; i < topElements.length; i++) {
-    var blockBoundary = topElements[i].getBoundingRectangle();
+    var topElement = topElements[i];
+    if (topElement.isInsertionMarker && topElement.isInsertionMarker()) {
```
moniika, line 1730: Choosing to check here, rather than not allowing insertion markers to be added to top bounded elements, in order to avoid any performance implications of needing to remove the block from top bounded elements whenever an insertion marker is created.
An insertion marker is created as a regular block (would be added to top elements) and then set as an insertion maker (where we would need to remove it from top elements). Removing from top elements would require traversing through all the top elements (with the `remove` call).

alschmiedt (approved):

Issue #1292, opened 2017-08-23
Enable scrolling in only one direction


rachel-fenichel:
E.g. only horizontal scrolling in the workspace, or only vertical scrolling.  The other position would be fixed.

Open questions:
- Where to specify which direction of scrolling to allow?  Right now scrolling is specified in the options that are passed into the inject call, but is just a boolean.  
We could change it to be: "both", "none", "vertical", and "horizontal".
Leaving it undefined would keep the current behaviour: "both" if there are categories, otherwise "none".
"true" would be parsed as "both".
"false" would be parsed as "none".

- Will we allow vertical-only scrolling with a horizontal toolbox?  Or the other way around?
- What happens if the user zooms?  Or is that disallowed?

In the non-scrolling direction we would preserve the current behaviour:
- Blocks placed outside of the bounds would be [bumped back in](https://github.com/google/blockly/blob/master/core/inject.js#L203).
- The content dimensions would be the size of the blocks bounding box in that dimension, which is not allowed to be bigger than the viewport in that direction.

In the scrolling direction we would preserve the current scrolling behaviour:
- The content dimensions would include a buffer on each side, equal to half of the size of the viewport in that direction.

RoboErikG:
> Will we allow vertical-only scrolling with a horizontal toolbox? Or the other way around?
> What happens if the user zooms? Or is that disallowed?
Would these just be checks on the injection params? I think you're right that those are probably bad configurations, but in general I think we should be moving away from APIs that need to check other parts of the UI to figure out what they can do. It's much easier to maintain an API for each component if you don't try to handle all the possible configurations for components, even if it lets devs do weird combinations.

microsoftsam:
Will this be a mode (ie: an option)? or will that be the default, dependent on whether or not you're using horizontal or vertical workspaces?

RoboErikG:
@Sam, this would be a new option, not the default. It was a feature we wanted to have for [Code a Snowflake](https://santatracker.google.com/snowflake.html) but didn't have time to implement.

rachel-fenichel:
Here's another fun use case: setting a fixed height to the workspace, but still allowing vertical scrolling.
That would be for horizontal mode, and an equivalent vertical version would exist.  

That would let the user have more than a single stack of blocks, while not letting them get too lost in the workspace.

brokenpylons:
I managed to hack something together. I added a single vertical scrollbar with:

```
[...]

PR #4651, opened 2021-02-23
Adding single-direction scroll options to inject options

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of #1292

### Proposed Changes

Adds inject options to specify whether workspace should have single-directions scrollbars.

#### Behavior After Change

Single direction scrollbars can be enabled for main workspace by passing options `move: {hasVerticalScrollbars: true}` or `move: {hasHorizontalScrollbars: true}`.
If `hasHorizontalScrollbars` and `hasVerticalScrollbars` are not used in options, the behavior of parsing workspace options remains the same as before this PR.

### Reason for Changes

Adding support for single-direction scroll.

### Test Coverage

Tested with playground.

* Desktop Chrome

### Documentation

Inject options on devsite should be updated to reflect this addition.

### Additional Information

Depends on #4652

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -381,7 +381,10 @@ Blockly.init_ = function(mainWorkspace) {
   if (options.moveOptions && options.moveOptions.scrollbars) {
     mainWorkspace.scrollbar =
         new Blockly.ScrollbarPair(
-            mainWorkspace, true, true, 'blocklyMainWorkspaceScrollbar');
+            mainWorkspace,
+            options.moveOptions.hasHorizontalScrollbars,
+            options.moveOptions.hasVerticalScrollbars,
+            'blocklyMainWorkspaceScrollbar');
     mainWorkspace.scrollbar.resize();
   } else {
     mainWorkspace.setMetrics({x: 0.5, y: 0.5});
--- a/core/inject.js
+++ b/core/inject.js
@@ -378,14 +378,20 @@ Blockly.init_ = function(mainWorkspace) {
     mainWorkspace.zoomControls_.init(verticalSpacing);
   }
 
-  if (options.moveOptions && options.moveOptions.scrollbars) {
-    mainWorkspace.scrollbar =
-        new Blockly.ScrollbarPair(
-            mainWorkspace,
-            options.moveOptions.hasHorizontalScrollbars,
-            options.moveOptions.hasVerticalScrollbars,
-            'blocklyMainWorkspaceScrollbar');
-    mainWorkspace.scrollbar.resize();
+  if (options.moveOptions) {
+    var horizontalScroll = options.moveOptions.scrollbars === true ||
+        options.moveOptions.scrollbars.horizontal;
+    var verticalScroll = options.moveOptions.scrollbars === true ||
+        options.moveOptions.scrollbars.vertical;
+    if (horizontalScroll || verticalScroll) {
+      mainWorkspace.scrollbar =
+          new Blockly.ScrollbarPair(
+              mainWorkspace, horizontalScroll, verticalScroll,
+              'blocklyMainWorkspaceScrollbar');
+      mainWorkspace.scrollbar.resize();
+    } else {
+      mainWorkspace.setMetrics({x: 0.5, y: 0.5});
+    }
   } else {
     mainWorkspace.setMetrics({x: 0.5, y: 0.5});
   }
--- a/core/inject.js
+++ b/core/inject.js
@@ -380,9 +380,9 @@ Blockly.init_ = function(mainWorkspace) {
 
   if (options.moveOptions && options.moveOptions.scrollbars) {
     var horizontalScroll = options.moveOptions.scrollbars === true ||
-        options.moveOptions.scrollbars.horizontal;
+        !!options.moveOptions.scrollbars.horizontal;
     var verticalScroll = options.moveOptions.scrollbars === true ||
-        options.moveOptions.scrollbars.vertical;
+        !!options.moveOptions.scrollbars.vertical;
     mainWorkspace.scrollbar =
         new Blockly.ScrollbarPair(
             mainWorkspace, horizontalScroll, verticalScroll,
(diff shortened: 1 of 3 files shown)


core/options.js
```diff
@@ -250,10 +252,19 @@ Blockly.Options.prototype.getMetrics;
 Blockly.Options.parseMoveOptions_ = function(options, hasCategories) {
   var move = options['move'] || {};
   var moveOptions = {};
-  if (move['scrollbars'] === undefined && options['scrollbars'] === undefined) {
+  if (move['hasVerticalScrollbars'] !== undefined || move['hasHorizontalScrollbars'] !== undefined) {
+    moveOptions.hasVerticalScrollbars = !!move['hasVerticalScrollbars'];
```
samelhusseini, line 256: In the interest of keeping the options consistent, let's not have `has` in the name of the option. (e.g.: to determine whether or not the wheel is enabled, we just use the term `wheel` rather than `hasWheel`).
Another thing to consider is splitting this into an object such that: 
- if scrollbars is a boolean and true, enable both
- if scrollbar is an object, look further for `horizontal` and `vertical` fields to determine if each axis is enabled.
The resulting interface would look something like this:

```
interface MoveOptions {
   drag: boolean;
   scrollbars: boolean | ScrollbarOptions;
   wheel: boolean
}

interface ScrollbarOptions {
    horizontal: boolean;
    vertical: boolean;
}
```

moniika: Done.

core/inject.js
```diff
@@ -379,9 +379,14 @@ Blockly.init_ = function(mainWorkspace) {
   }
 
   if (options.moveOptions && options.moveOptions.scrollbars) {
+    var horizontalScroll = options.moveOptions.scrollbars === true ||
+        options.moveOptions.scrollbars.horizontal;
```
samelhusseini, line 383: if `scrollbars = false` horziontalScroll will be equal to undefined. Which is okay the way we're using it and happen to not lie about its type (ScrollbarPair param type is `boolean=`), however I think it's been using as a boolean in this context.

Prefer if you did `!!options.moveOptions.scrollbars.horizontal` here and below.

core/options.js
```diff
@@ -205,12 +205,21 @@ Blockly.Options.GridOptions;
...
  *     drag: boolean,
- *     scrollbars: boolean,
+ *     scrollbars: (boolean | !Blockly.Options.ScrollbarOptions),
  *     wheel: boolean
  * }}
  */
 Blockly.Options.MoveOptions;
 
+/**
+ * Move Options.
```
samelhusseini, line 215: Scrollbar Options

core/options.js
```diff
@@ -250,11 +259,23 @@ Blockly.Options.prototype.getMetrics;
 Blockly.Options.parseMoveOptions_ = function(options, hasCategories) {
   var move = options['move'] || {};
   var moveOptions = {};
-  if (move['scrollbars'] === undefined && options['scrollbars'] === undefined) {
+  if (move['scrollbars'] === undefined) {
```
samelhusseini, line 262: Just confirming that you're intentionally removing support for this: 

```
const options = {
   scrollbars: true
}
```

core/options.js
```diff
@@ -250,11 +259,23 @@ Blockly.Options.prototype.getMetrics;
...
     moveOptions.scrollbars = hasCategories;
+  } else if (move['scrollbars'] &&
+      move['scrollbars']['horizontal'] !== undefined &&
+      move['scrollbars']['vertical'] !== undefined) {
+    moveOptions.scrollbars.horizontal = !!move['scrollbars']['horizontal'];
+    moveOptions.scrollbars.vertical = !!move['scrollbars']['vertical'];
```
samelhusseini, line 268: There's an opportunity here to normalize the internal options from the external options. 
Yes, we accept options in a number of various ways (scrollbar can be boolean or an object), but internally since we're setting the values here, is there a reason we couldn't normalize everything to use the new scrollbaroptions format?

core/inject.js
moniika: sgtm

core/options.js
moniika: whoops, definitely will update.

core/options.js
moniika: I started doing that at first, but I was worried about how moveOptions is public. So, I wanted to make it so that the type didn't change if the new single-scrollbar feature wasn't being used.

core/options.js
moniika: Wasn't my intensions, must have just gotten lost in the refactor as I was working on it. I will add it back.

core/options.js
samelhusseini: Ok, two things: 
- This will only accept an object with both horizontal and vertical defined, can we default to false? if only one is defined.
- Let's see if we can simplify it a little, how about:

```js
if (move['scrollbars'] === undefined && options['scrollbars'] === undefined) {
   moveOptions.scrollbars = hasCategories;  
} else {
   if (typeof moveOptions['scrollbars'] === 'object') {
      moveOptions.scrollbars. horizontal = !!moveOptions['scrollbars']['horizontal'];
      moveOptions.scrollbars. vertical = !!moveOptions['scrollbars']['vertical'];
   } else {
       moveOptions.scrollbars = !!move['scrollbars'] || !!options['scrollbars'];
   }
}
```

moniika: I've simplified a bit, but I've left in the logic that changes `scrollbars` to a `boolean` if `horizontal` and `vertical` are the same value. Leaving that logic in allows me to simplify checks for for whether scrollbars exist because I can do `!!moveOptions.scrollbars` (otherwise if both `horizontal` and `vertical` are set `false` in the object, this would return `true` incorrectly).

samelhusseini: Sounds good, please add a comment for why this is needed.

samelhusseini (approved):

core/options.js
moniika: added

PR #4618, opened 2021-02-04
Updating bump logic to support single-direction scrollbars

moniika:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

 Part of #1292

### Proposed Changes

- Metrics changes:
  - Adds scroll metrics. (`getScrollMetrics` and adds scroll metrics to the metrics object `Blockly.utils.Metrics`)
  - Changes content metrics computation/usage
- Bump logic (bumping blocks/workspace comments within fixed edges of workspace):
  - Adds `oldScale` property to `ViewportChange` event (in order to easily detect zoom changes)
  - Bumps objects if they are created/moved outside of scroll area
  - Bumps top objects on zoom in/out and window resize (where scroll area may have changed)

### Reason for Changes

Necessary to support single-direction scrollbar

### Test Coverage

Extensive manual testing in playground. Mocha tests.

Tested on:
* Desktop Chrome

### Additional Information

Expecting merge conflicts with #4643

PR #4652, opened 2021-02-24
Updating bump logic to support single-direction scrollbars

moniika:
## The basics

- [x] I branched from develop

### Resolves

 Part of #1292

### Proposed Changes

- Metrics changes:
  - Adds scroll metrics. (`getScrollMetrics` and adds scroll metrics to the metrics object `Blockly.utils.Metrics`)
  - Changes content metrics computation/usage
- Bump logic (bumping blocks/workspace comments within fixed edges of workspace):
  - Bumps objects if they are created/moved outside of scroll area
  - Bumps top objects on window resize (where scroll area may have changed)

### Reason for Changes

Necessary to support single-direction scrollbar

### Test Coverage

Extensive manual testing in playground. Mocha tests.

Tested on:
* Desktop Chrome

### Additional Information

Hosted at: https://moniika.github.io/blockly/tests/playground.html
Remake of #4618


core/inject.js
```diff
@@ -381,12 +376,13 @@ Blockly.init_ = function(mainWorkspace) {
   }
 
   if (options.moveOptions && options.moveOptions.scrollbars) {
+    // TODO change back to addHorizontal=true before merge.
     mainWorkspace.scrollbar =
         new Blockly.ScrollbarPair(
-            mainWorkspace, true, true, 'blocklyMainWorkspaceScrollbar');
+            mainWorkspace, false, true, 'blocklyMainWorkspaceScrollbar');
```
moniika, line 382: This was set temporarily so the hosted playground has single scrollbars.
This line is also the cause of the mocha test failure.

core/inject.js
```diff
@@ -329,6 +189,140 @@ Blockly.createMainWorkspace_ = function(svg, options, blockDragSurface,
...
+    case Blockly.Events.COMMENT_MOVE:
+      object = workspace.getCommentById(e.commentId);
+      break;
+  }
+  return object;
+};
+
+/**
+ * Bumps the top objects in the given workspace into bounds.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace.
```
alschmiedt, line 221: Do these methods need to be public?

core/inject.js
```diff
@@ -352,6 +346,7 @@ Blockly.init_ = function(mainWorkspace) {
       function() {
         Blockly.hideChaff(true);
         Blockly.svgResize(mainWorkspace);
+        Blockly.bumpTopObjectsIntoBounds(mainWorkspace);
```
alschmiedt, line 349: Do you know if there are any performance implications with this change?

alschmiedt (commented):
Not necessarily in this PR, but I am curious what your thoughts are on pulling out the bump logic from inject? I think this would give us the ability to register the class later on, so someone could create their own bump blocks logic.

core/inject.js
moniika: I don't expect any performance regressions as part of this change because `bumpTopObjectsIntoBounds` checks `hasScrollEdges` and returns early before any metrics calls for workspace with scrollbar pairs.

core/scrollbar.js
```diff
@@ -571,7 +570,7 @@ Blockly.Scrollbar.prototype.constrainPosition_ = function(value) {
     value = 0;
   } else {
     // Handle length should never be greater than this.scrollViewSize_.
-    // If the viewSize is greater than or equal to the contentSize, the
+    // If the viewSize is greater than or equal to the scrollSize, the
     // handleLength will end up equal to this.scrollViewSize_.
     value = Math.min(value, this.scrollViewSize_ - this.handleLength_);
```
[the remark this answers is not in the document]
moniika: I would be in favor of renaming `scrollViewSize_`. I'll think a little about other potential names.

core/inject.js
moniika: No, I think you're right. private or protected would be more appropriate (protected if it ends up being refactored out).

core/metrics_manager.js
```diff
@@ -279,41 +219,160 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+  }
+
+  var hScrollEnabled = this.workspace_.isMovableHorizontally();
+  var vScrollEnabled = this.workspace_.isMovableVertically();
+
+  var viewMetrics = opt_viewMetrics || this.getViewMetrics(false);
+
+  var edges = {};
+  if (!vScrollEnabled) {
+    if (edges.top !== undefined) {
```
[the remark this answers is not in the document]
moniika: This logic is leftover from when I was testing how fixed edges might work (where the top and the left edge doesn't go past 0). I want to talk with you about it, maybe offline, about leaving some of this logic in and providing a hook to make it easier to set this kind of an edge and settle on terminology for what we want to call a "fixed edge".
As it is now, this block could be simplified because edges properties are always undefined.

core/scrollbar.js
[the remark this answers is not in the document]
moniika: What do you think about `scrollbarLength_`?

core/inject.js
moniika: Removed

core/metrics_manager.js
```diff
@@ -279,41 +219,160 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+ * Returns the "fixed" edges of the scroll area.
+ * @param {!Blockly.MetricsManager.ContainerRegion=} opt_viewMetrics The view
+ *     metrics if they have been previously computed. Passing in null may cause
+ *     the view metrics to be computed again, if it is needed.
+ * @return {{top: (number|undefined), bottom: (number|undefined),
+ *     left: (number|undefined), right: (number|undefined)}} The fixed edges of
+ *     the scroll area.
+ * @protected
+ */
+Blockly.MetricsManager.prototype.getScrollEdges_ = function(
```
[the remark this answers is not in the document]
moniika: Changing to `computeFixedEdges_`.

core/metrics_manager.js
```diff
@@ -279,41 +219,160 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+    top: blockBox.top * scale,
+    left: blockBox.left * scale,
+  };
+};
+
+/**
+ * Returns whether the scroll area has "fixed" edges.
+ * @return {boolean} Whether the scroll area has "fixed" edges.
+ */
+Blockly.MetricsManager.prototype.hasScrollEdges = function() {
```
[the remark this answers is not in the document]
moniika: I think that works.

core/metrics_manager.js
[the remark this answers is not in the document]
moniika: Simplified logic.

core/utils/metrics.js
```diff
@@ -42,6 +42,18 @@ Blockly.utils.Metrics.prototype.contentHeight;
  */
 Blockly.utils.Metrics.prototype.contentWidth;
 
+/**
+ * Height of the scroll area.
```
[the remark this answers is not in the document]
moniika: Updating scrollable area to scroll area in MetricsManager for consistency.

core/metrics_manager.js
```diff
@@ -279,41 +219,160 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+      !this.workspace_.isMovableVertically();
+};
+
+/**
+ * Returns the "fixed" edges of the scroll area.
+ * @param {!Blockly.MetricsManager.ContainerRegion=} opt_viewMetrics The view
+ *     metrics if they have been previously computed. Passing in null may cause
+ *     the view metrics to be computed again, if it is needed.
+ * @return {{top: (number|undefined), bottom: (number|undefined),
+ *     left: (number|undefined), right: (number|undefined)}} The fixed edges of
```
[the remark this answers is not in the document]
moniika: Done.

core/metrics_manager.js
```diff
@@ -280,40 +231,145 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+    left: blockBox.left * scale,
+  };
+};
+
+/**
+ * Returns whether the scroll area has fixed edges.
+ * @return {boolean} Whether the scroll area has fixed edges.
+ * @package
+ */
+Blockly.MetricsManager.prototype.hasScrollEdges = function() {
```
alschmiedt, line 262: Since we updated a few of the other names to be `fixedEdges` should this also be `hasFixedEdges`?

core/metrics_manager.js
```diff
@@ -280,40 +231,145 @@ Blockly.MetricsManager.prototype.getViewMetrics = function(
...
+/**
+ * Computes the fixed edges of the scroll area.
+ * @param {!Blockly.MetricsManager.ContainerRegion=} opt_viewMetrics The view
+ *     metrics if they have been previously computed. Passing in null may cause
+ *     the view metrics to be computed again, if it is needed.
+ * @return {Blockly.MetricsManager.FixedEdges} The fixed edges of the scroll
+ *     area.
+ * @protected
+ */
+Blockly.MetricsManager.prototype.computeFixedEdges_ = function(
```
alschmiedt, line 277: Since we are still getting something maybe `getComputedFixedEdges_`?

core/inject.js
```diff
@@ -387,7 +384,7 @@ Blockly.init_ = function(mainWorkspace) {
             mainWorkspace, true, true, 'blocklyMainWorkspaceScrollbar');
     mainWorkspace.scrollbar.resize();
   } else {
-    mainWorkspace.setMetrics({x: 0.5, y: 0.5});
+    mainWorkspace.setMetrics({x: 0, y: 0});
```
alschmiedt, line 387: Just curious, why the change?

alschmiedt (approved):
A few small comments left, but then lgtm!

core/metrics_manager.js
moniika: Yeah, agree. I'll update

core/metrics_manager.js
moniika: sgtm

core/inject.js
moniika: I believe I changed it because it fixed unit tests. However it looks like it doesn't make a difference now, so I'll change it back.
I think it's possible that the scrollbar logic refactor PR may have something to do with it.

PR #4653, opened 2021-02-25
Add zoom support for single-direction-scroll

moniika:
## The basics

- [x] I branched from develop

### Resolves

 Part of #1292

### Proposed Changes

- ZoomReset
  - Updates zoom reset implementation to work for single-direction scrollbars 
- Bump logic (bumping blocks/workspace comments within fixed edges of workspace):
  - Adds `oldScale` property to `ViewportChange` event (in order to easily detect zoom changes)
  - Bumps top objects on zoom in/out (where scroll area may have changed)

### Reason for Changes

Adds logic to support zooming in main workspace with single scrollbars.

### Test Coverage

Extensive manual testing in playground. Mocha tests.

Tested on:
* Desktop Chrome

### Additional Information

Depends on #4652


core/zoom_controls.js
```diff
@@ -453,7 +453,26 @@ Blockly.ZoomControls.prototype.createZoomResetSvg_ = function(rnd) {
  */
 Blockly.ZoomControls.prototype.resetZoom_ = function(e) {
   this.workspace_.markFocused();
-  this.workspace_.setScale(this.workspace_.options.zoomOptions.startScale);
+
+  // Update scrollX/scrollY so that the setScale translates the workspace
+  // correctly. This is needed for single-direction scrollbars.
+  var oldScale = this.workspace_.scale;
```
rachel-fenichel, line 459: Some questions:

- Why is this the responsibility of the zoom controls and not the workspace or the scrollbars?
  - Could it just be part of the scrollCenter function?
  - Are the zoom controls directly setting scrollX and scrollY anywhere else?
- Can you check the scrollbars and only do this when there are single-direction scrollbars?

core/inject.js
```diff
@@ -246,32 +246,38 @@ Blockly.bumpTopObjectsIntoBounds_ = function(workspace) {
...
+      if (wasBumped && !e.group) {
+        console.warn('Moved object in bounds but there was no' +
+            ' event group. This may break undo.');
+      }
+      if (oldGroup !== null) {
+        Blockly.Events.setGroup(oldGroup);
+      }
+    } else if (e.type === Blockly.Events.VIEWPORT_CHANGE) {
+      var viewportEvent = /** @type {!Blockly.Events.ViewportChange} */ (e);
+      if (viewportEvent.scale !== viewportEvent.oldScale) {
```
rachel-fenichel, line 278: In addition to checking whether the scale has changed at all, you can check whether the viewport got smaller (and thus things might need to be bumped). If the viewport grew you can just skip the bump.

core/zoom_controls.js
moniika: My issues stem from how `setScale` does a translation of the workspace that ends up moving the "center" of the workspace.
This wasn't a problem before because the scroll area used to be equivalent to the contentArea and scrolling the workspace wouldn't change the center.

In order to get around this problem, I need something like `zoomCenter` but where I pass in the exact scale.

I was initially intimidated by trying to compute the correct amount to pass to `zoomCenter`, so I went this route where I used logic from `zoom` to translate the workspace (by setting `scrollX` and `scrollY`) before calling `setScale` and I figured we may end up moving this logic into workspace.

HOWEVER, I tried my hand at computing the amount to pass to `zoomCenter` and with the extra comments I've added I think it's not too hard to follow.

core/inject.js
moniika: done.

rachel-fenichel (approved):

PR #4656, opened 2021-02-25
Fix event related missing requires

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes circular dependency between `Blockly.Events` and `Blockly.Xml`.
Fix 46 warnings when `stringMissingRequires` is enabled related to events, registry and xml.

### Proposed Changes

Use the Blockly registry when instantiating a Blockly event.

### Reason for Changes

Path towards enabling strictMissingRequires

### DIFF
--- a/core/zoom_controls.js
+++ b/core/zoom_controls.js
@@ -14,6 +14,8 @@ goog.provide('Blockly.ZoomControls');
 
 goog.require('Blockly.constants');
 goog.require('Blockly.Css');
+goog.require('Blockly.Events');
+goog.require('Blockly.Events.Click');
 goog.require('Blockly.Scrollbar');
 goog.require('Blockly.Touch');
 goog.require('Blockly.utils.dom');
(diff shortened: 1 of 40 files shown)


core/zoom_controls.js
```diff
@@ -14,6 +14,8 @@ goog.provide('Blockly.ZoomControls');
 
 goog.require('Blockly.constants');
 goog.require('Blockly.Css');
+goog.require('Blockly.Events');
+goog.require('Blockly.Events.Click');
```
rachel-fenichel, line 18: I think you don't need to require `Events.Click` now, because you're no longer calling the constructor directly.

rachel-fenichel (commented):
LGTM

core/zoom_controls.js
samelhusseini: I went back and forth on this, so let me explain my thinking.
You're right, this is no longer needed since the Blockly.events using the registry takes care of grabbing the right event. 
However, the only way for events to be registered is for them to be included in the bundle. In other words, all our event files (block_events, ws_comment_events, etc), have to be required somewhere.

To draw parallels, we require all themes built-in themes and renderers in the requires.js entry point, the difference here is that there IS an inherit dependency here that needs to be captured somehow: 
- Option A: Use goog.require to indicate a dependency on a specific set of events at the call site (hard to enforce as the compiler won't help here)
- Option B: require all built-in events in some central location. The problem with this option is take for example trash can events. Say you build a version of blockly (using closure -- in say blockly games, and you don't need the trashcan. Why should the trashcan event be included in your bundle? So having it associated with the trashcan.js file makes more sense here.

Thoughts?

rachel-fenichel: Okay. I think leaving it this way is reasonable, and that we will need to take a look at the event requires more generally at some point in the future. I see why B is a problem.

PR #4659, opened 2021-02-26
Use registry for creating the no-category flyout

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes 6 strict missing requires errors.

### Proposed Changes

Use the registry for creating the vertical / horizontal flyout on a no-category workspace.

### Additional Information

@rachel-fenichel  FYI

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -82,7 +82,7 @@ Blockly.WorkspaceSvg = function(
   Blockly.WorkspaceSvg.superClass_.constructor.call(this, options);
 
   var MetricsManagerClass = Blockly.registry.getClassFromOptions(
-      Blockly.registry.Type.METRICS_MANAGER, options);
+      Blockly.registry.Type.METRICS_MANAGER, options, true);
   /**
    * Object in charge of calculating metrics for the workspace.
    * @type {!Blockly.IMetricsManager}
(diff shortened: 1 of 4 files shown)


core/workspace_svg.js
```diff
@@ -992,15 +990,13 @@ Blockly.WorkspaceSvg.prototype.addFlyout = function(tagName) {
       }));
   workspaceOptions.toolboxPosition = this.options.toolboxPosition;
   if (this.horizontalLayout) {
-    if (!Blockly.HorizontalFlyout) {
-      throw Error('Missing require for Blockly.HorizontalFlyout');
-    }
-    this.flyout_ = new Blockly.HorizontalFlyout(workspaceOptions);
+    var HorizontalFlyout = Blockly.registry.getClassFromOptions(
```
rachel-fenichel, line 993: This no longer throws a clear error for the missing require. Are we okay with that?

samelhusseini: That’s true, so it’s not exactly the right check to check if Blockly.HorizontalFlyout exists any more since it can be some other object that is registered, but we can check if getClassFromOptions returns nothing and throw. 
I would rather deal with this in the registry. It is currently logging a warning. This sounds like something that applies anywhere we request a class from the registry (the whole warn vs throw dilemma). If we want to throw or warn on a case by case basis then we can add another param to the registry call to request a throw if it yields nothing. I would just rather not have to always check if exists and throw.

rachel-fenichel: I like having it optionally throw on failure. At the call site we often know whether a failure would be catastrophic--like here--and a centralized throw would be way better.

samelhusseini: Ok I’ll add that functionality to the registry, and make it throw from this call site

samelhusseini: Done.

rachel-fenichel (approved):

PR #4660, opened 2021-02-26
Fix missing require in contextmenu items

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Fixes 71 warnings (Big win)

### Proposed Changes

Fix circular dependency between context menu registry and context menu registry items.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -20,6 +20,7 @@ goog.require('Blockly.browserEvents');
 goog.require('Blockly.constants');
 goog.require('Blockly.ContextMenu');
 goog.require('Blockly.ContextMenuRegistry');
+goog.require('Blockly.ContextMenuItems');
 goog.require('Blockly.Events');
 goog.require('Blockly.Events.BlockMove');
 goog.require('Blockly.Events.Selected');
--- a/core/contextmenu_items.js
+++ b/core/contextmenu_items.js
@@ -17,6 +17,7 @@
 goog.provide('Blockly.ContextMenuItems');
 
 goog.require('Blockly.constants');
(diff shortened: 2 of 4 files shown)


core/block_svg.js
```diff
@@ -20,6 +20,7 @@ goog.require('Blockly.browserEvents');
 goog.require('Blockly.constants');
 goog.require('Blockly.ContextMenu');
 goog.require('Blockly.ContextMenuRegistry');
+goog.require('Blockly.ContextMenuItems');
```
maribethb, line 23: Is this actually used in the class? I don't see that it is. I don't love relying on tricks like this to force the `registerDefaultOptions` function to be called (which I think is what's happening here?). If we had all the tools we have in google3 for example, this would be an unused import and removed by the autoformatter. I know we should remove the circular dependency, but this feels fragile since it's not obvious that this is necessary and could get removed in a "cleanup" at some point.

maribethb (commented):
also looks like some mocha tests are failing, but i can't figure out how to see which ones

core/block_svg.js
samelhusseini: I see your point, but it’s no different to having to require fields just so they can be included in the bundle and “registered”. In most cases throughout Blockly we require things because we need them in the file, but there are exceptions for “registrables” that we need to require the file somewhere.
This is usually reserved for the requires.js file, so I’m not opposed to moving the ContextMenuItems out to that file, it just felt like a better fit here (but I could be convinced).
You’re right in calling them extraRequires though.

samelhusseini:
Closing in favour of https://github.com/google/blockly/pull/4681

PR #4662, opened 2021-02-26
Align constants -> Blockly.constants.ALIGN enum

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes 16 warnings.

### Proposed Changes

Move the align constants into an enum on the Blockly.constants namespace, and then add aliases in blockly.js.

### Reason for Changes

Resolves compiler warnings related to setting properties on `Blockly` from `constants.js`.
Allows files to require `constants` rather than `blockly` to get the correct values, to reduce circular dependencies.

### Documentation

- Will need to make sure the jsdoc is correct.
- Constants are referenced in the [define blocks page](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks)

### Additional Information

There are still uses of the old names in demos and in samples.
We will likely never get rid of the old constants for external use, because they are part of block definitions. But all internal uses should now use the constants file.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1765,10 +1765,10 @@ Blockly.Block.prototype.fieldFromJson_ = function(element) {
  */
 Blockly.Block.prototype.inputFromJson_ = function(element, warningPrefix) {
   var alignmentLookup = {
-    'LEFT': Blockly.ALIGN_LEFT,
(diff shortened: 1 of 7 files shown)


samelhusseini (approved):
Looking good!

PR #4669, opened 2021-03-02
Adding Positionable interface

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

Adds interface for positionable ui elements.

#### Behavior After Change

no change in behavior

### Reason for Changes

Adds formal API for ui elements that are positioned on the workspace.

### Test Coverage

Manual testing in playground.

 * Desktop Chrome

### Additional Information

I plan on refactoring collision detection into a helper method in a future PR.


core/trashcan.js
```diff
@@ -428,38 +431,74 @@ Blockly.Trashcan.prototype.emptyContents = function() {
...
  * It is positioned in the opposite corner to the corner the
  * categories/toolbox starts at.
- */
-Blockly.Trashcan.prototype.position = function() {
+ * @param {!Blockly.MetricsManager.ContainerRegion} viewMetrics The workspace
+ *     viewMetrics.
+ * @param {!Blockly.MetricsManager.AbsoluteMetrics} absoluteMetrics The absolute
+ *     metrics for the workspace.
+ * @param {!Blockly.MetricsManager.ToolboxMetrics} toolboxMetrics The toolbox
+ *     metrics for the workspace.
```
alschmiedt, line 439: Should we just pass in all the metrics? In case other positionables end up needing them?

core/trashcan.js
```diff
@@ -428,38 +431,74 @@ Blockly.Trashcan.prototype.emptyContents = function() {
...
+  this.top_ = Blockly.utils.math.clamp(minTop, this.top_, maxTop);
 
   this.svgGroup_.setAttribute('transform',
       'translate(' + this.left_ + ',' + this.top_ + ')');
 };
 
+/**
+ * Returns the bounding rectangle of the UI element in pixel units relative to
+ * the Blockly injection div.
+ * @returns {Blockly.utils.Rect} The plugin’s bounding box.
```
alschmiedt, line 494: Should be `!Blockly.utils.Rect`

core/utils/rect.js
```diff
@@ -52,3 +52,15 @@ Blockly.utils.Rect = function(top, bottom, left, right) {
 Blockly.utils.Rect.prototype.contains = function(x, y) {
   return x >= this.left && x <= this.right && y >= this.top && y <= this.bottom;
 };
+
+/**
+ * Tests whether this rectangle intersects the provided rectangle.
+ * Assumes that the coordinate system increases going down and left.
+ * @param {Blockly.utils.Rect} other The other rectangle to check for
```
alschmiedt, line 59: `!Blockly.utils.Rect`

core/zoom_controls.js
```diff
@@ -194,44 +197,82 @@ Blockly.ZoomControls.prototype.dispose = function() {
...
+  // Upper corner placement
+  var minTop = this.top_ = this.verticalSpacing_;
+  // Bottom corner placement
+  var maxTop = viewMetrics.height + absoluteMetrics.top -
+      this.HEIGHT_ - this.verticalSpacing_;
+  var placeBottom = toolboxMetrics.position !== Blockly.TOOLBOX_AT_BOTTOM;
+  this.top_ = placeBottom ? maxTop : minTop;
+  if (placeBottom) {
+    this.zoomInGroup_.setAttribute('transform', 'translate(0, 43)');
+    this.zoomOutGroup_.setAttribute('transform', 'translate(0, 77)');
```
alschmiedt, line 250: Probably not for this PR because it looks like these were already there. But this looks pretty suspicious. We should probably at least create local variables and add comments so we know where these numbers are coming from.

core/zoom_controls.js
```diff
@@ -194,44 +197,82 @@ Blockly.ZoomControls.prototype.dispose = function() {
...
+    if (boundingRect.intersects(otherEl)) {
+      if (placeBottom) {
+        // Bump up
+        this.top_ = otherEl.top - this.HEIGHT_ - this.MARGIN_BOTTOM_;
+      } else {
+        this.top_ = otherEl.bottom + this.MARGIN_BOTTOM_;
+      }
+      // Recheck other savedPositions
+      boundingRect = this.getBoundingRectangle();
+      i = -1;
```
alschmiedt, line 270: Do we handle the case where bumping makes the positionable overlap with another positionable? And if no, then is that a case we want to handle?

alschmiedt (commented):
Overall lgtm! Most of the comments are type related. Waiting for travis to pass before approving : )

core/trashcan.js
moniika: The downside to all the metrics is that contentMetrics is somewhat expensive and not something I would expect (or want...) a positionable to need.
That being said, I'm still open to passing a single metrics object. Going that route, I could also make a custom type that is passed so we have flexibility to adding what is included on the metrics object without needing to change the method signature.

core/zoom_controls.js
moniika: Yes. That's why I set `i=-1`, it forces rechecking of all positionables, starting with the first in the list.

core/zoom_controls.js
moniika: I left this logic unchanged. I don't fully understand what it's doing, but I think it has to do with rearranging the order from top to bottom of zoom in, zoom out, and zoom reset button.

core/zoom_controls.js
alschmiedt: Ahh, read this wrong. Ok thanks for clarifying : )

core/trashcan.js
alschmiedt: Ok, that sounds reasonable. And yes, I like the idea of passing a single metrics object.

moniika: As long as it's fine with you. I'm going to look at making a single metrics object in next PR.

alschmiedt (approved):

PR #4672, opened 2021-03-03
Add PluginManager

moniika:
## The basics

- [x] I branched from develop

### Proposed Changes

- Adds `PluginManager` to hold and organize plugins and core elements by type so and external developers can have access to them.
- Adds trashcan and zoom controls instances to `PluginManager`
- Creates metrics object for to use with `IPositionable.position` method

### Reason for Changes

Exposes API that allows external plugins to check where other positionable elements are placed.

### Test Coverage

Tested on playground.

* Desktop Chrome


core/plugin_manager.js
```diff
@@ -0,0 +1,91 @@
...
+  return this.pluginData_[id] && this.pluginData_[id].plugin;
+};
+
+/**
+ * Gets all the plugins of the specified type.
+ * @param {typeof Blockly.IPlugin} type The type of the plugin.
+ * @param {boolean} sorted Whether to return list ordered by weights.
+ * @return {!Array<!Blockly.IPlugin>} The plugins that match the specified type.
+ */
+Blockly.PluginManager.prototype.getPlugins = function(type, sorted) {
```
moniika, line 79: I wanted to do this typing with generics, but it seems like I can't use `typeof` in the JSDoc with generics.

core/plugin_manager.js
```diff
@@ -0,0 +1,91 @@
...
+goog.provide('Blockly.PluginManager');
+
+goog.require('Blockly.IPlugin');
+
+
+/**
+ * Manager for all items registered with the workspace.
+ * @constructor
+ */
+Blockly.PluginManager = function() {
```
moniika, line 23: I'd like to rename this and `IPlugin` (but I don't yet have a good alternative in mind) away from using the word "plugin". 
We are currently using plugins to refer to objects/overrides that developers can add on top of core Blockly (such as where we pass in inject plugin options classes to override), but this manager is also used to track core Blockly elements (like trashcan and zoom).

core/workspace_svg.js
```diff
@@ -511,6 +518,15 @@ Blockly.WorkspaceSvg.prototype.getMetricsManager = function() {
   return this.metricsManager_;
 };
 
+/**
+ * Gets the plugin manager for this workspace.
+ * @return {!Blockly.PluginManager} The metrics manager.
```
samelhusseini, line 523: The plugin manager.

core/workspace_svg.js
```diff
@@ -953,6 +969,12 @@ Blockly.WorkspaceSvg.prototype.addTrashcan = function() {
   this.trashcan = new Blockly.Trashcan(this);
   var svgTrashcan = this.trashcan.createDom();
   this.svgGroup_.insertBefore(svgTrashcan, this.svgBlockCanvas_);
+  this.pluginManager_.addPlugin({
+    id: 'trashcan',
+    plugin: this.trashcan,
+    weight: 0,
```
samelhusseini, line 975: rather than weight, is this more of a zindex?

core/workspace_svg.js
```diff
@@ -1078,25 +1106,14 @@ Blockly.WorkspaceSvg.prototype.resize = function() {
...
-    var toolboxMetrics = metricsManager.getToolboxMetrics();
-    var savedPositions = [];
-    for (var i = 0, uiElement; (uiElement = positionableEls[i]); i++) {
-      uiElement.position(
-          viewMetrics, absoluteMetrics, toolboxMetrics, savedPositions);
-      savedPositions.push(uiElement.getBoundingRectangle());
-    }
+  
+  var positionables = /** @type {!Array<!Blockly.IPositionable>} */
+      (this.pluginManager_.getPlugins(Blockly.IPositionable, true));
```
samelhusseini, line 1111: You might be able to do this cast automatically with a generic.

core/workspace_svg.js
```diff
@@ -967,6 +989,12 @@ Blockly.WorkspaceSvg.prototype.addZoomControls = function() {
   this.zoomControls_ = new Blockly.ZoomControls(this);
   var svgZoomControls = this.zoomControls_.createDom();
   this.svgGroup_.appendChild(svgZoomControls);
+  this.pluginManager_.addPlugin({
+    id: 'zoomControls',
+    plugin: this.zoomControls_,
+    weight: 0,
+    types: [Blockly.IPositionable]
```
samelhusseini, line 996: Can you check this with compiled mode, and advanced compiled mode? I wonder if interfaces are filtered out completely out of existence that they cause issues here.

samelhusseini (commented):
Really good direction here! Great work!

core/workspace_svg.js
moniika: yes

core/plugin_manager.js
```diff
@@ -0,0 +1,91 @@
...
+Blockly.PluginManager.PluginData;
+
+/**
+ * Adds a plugin.
+ * @param {!Blockly.PluginManager.PluginData} pluginDataObject The plugin.
+ */
+Blockly.PluginManager.prototype.addPlugin = function(pluginDataObject) {
+  this.pluginData_[pluginDataObject.id] = pluginDataObject;
+  for (var i = 0, type; (type = pluginDataObject.types[i]); i++) {
+    var typeKey = type.name;  // The map key needs to be a string
```
samelhusseini, line 54: I would be careful with using Function.name. It's not supported in IE 11:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

and it's also not consistent across browsers, e.g: I tried this on my machine, and it's an empty string for `Blockly.IPositionable.name`

core/workspace_svg.js
moniika: Refactored to use generics

core/plugin_manager.js
moniika: code updated

core/workspace_svg.js
```diff
@@ -1078,25 +1106,15 @@ Blockly.WorkspaceSvg.prototype.resize = function() {
...
-    var absoluteMetrics = metricsManager.getAbsoluteMetrics();
-    var toolboxMetrics = metricsManager.getToolboxMetrics();
-    var savedPositions = [];
-    for (var i = 0, uiElement; (uiElement = positionableEls[i]); i++) {
-      uiElement.position(
-          viewMetrics, absoluteMetrics, toolboxMetrics, savedPositions);
-      savedPositions.push(uiElement.getBoundingRectangle());
-    }
+
+  var positionables = /** @type {!Array<!Blockly.IPositionable>} */
```
samelhusseini: This case should no longer be necessary.

moniika: just removed it, right as you commented

core/plugin_manager.js
samelhusseini: Any thoughts on FeatureManager?

samelhusseini (approved):

core/plugin_manager.js
moniika: I'm not sold on `FeatureManager` because I think it's not specific enough. Feature is a very encompassing term, and there's plenty of features that this doesn't manage.
I think it might get easier to find a name after we try to use this a bit and get a better idea of what things we would add to the manager, other than `IPositionable`.

samelhusseini (approved):

PR #4679, opened 2021-03-05
Mutator metrics

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
- The mutator's `getMetrics` method returned a subset of metrics. 
- The mutator's `getMetrics` method would return null if the workspace was not visible. Our type we return from getMetrics is {!Blockly.utils.Metrics}, so this was incorrect.
- Storing random values on an object (svg.cachedWidth_) is not recommended.

### Proposed Changes
The main workspace's metrics manager is used for both the mutator's workspace and the main workspace. 
The main change needed to do this, was to store the cached svg parent size for both a mutator's workspace as well as the main workspace. 

#### Behavior Before Change
Calling `getMetrics` on a mutator's workspace only returns view and absolute metrics.

#### Behavior After Change
Calling `getMetrics` on a mutator's workspace returns all the metrics.

### Reason for Changes
Mutator workspace's had their own `getMetrics` method that only returned a subset of the metrics that the main workspace returned. This made typing difficult since we couldn't assume some of the fields would be returned from getMetrics. It also added complexity since we were computing these metrics differently.


core/workspace_svg.js
```diff
@@ -725,6 +733,17 @@ Blockly.WorkspaceSvg.prototype.getSvgXY = function(element) {
...
 };
 
+/**
+ * Gets the size of the workspace's parent svg element.
+ * @return {!Blockly.utils.Size} The cached width and height of the workspace's
+ *     parent svg element.
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.getCachedParentSvgSize = function() {
+  var size = this.cachedParentSvgSize_;
```
rachel-fenichel, line 743: Copying to avoid handing out the actual object, right?

core/workspace_svg.js
```diff
@@ -1146,6 +1165,27 @@ Blockly.WorkspaceSvg.prototype.getCanvas = function() {
...
+ * with getSvgMetrics.
+ * @param {?number} width The width of the parent svg element.
+ * @param {?number} height The height of the parent svg element
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.setCachedParentSvgSize = function(width, height) {
+  var svg = this.getParentSvg();
+  if (width) {
+    this.cachedParentSvgSize_.width = width;
+    // This is set to support the public Blockly.svgSize method.
```
rachel-fenichel, line 1179: "public (but deprecated) Blockly.svgSize method"

rachel-fenichel (approved):

core/workspace_svg.js
alschmiedt: Yup! This was causing a super fun bug this morning : )

core/workspace_svg.js

core/blockly.js
```diff
@@ -114,8 +114,16 @@ Blockly.parentContainer = null;
  * Returns the dimensions of the specified SVG image.
  * @param {!SVGElement} svg SVG image.
  * @return {!Blockly.utils.Size} Contains width and height properties.
+ * @deprecated Use workspace.setCachedParentSvgSize. (2021 March 5)
  */
 Blockly.svgSize = function(svg) {
+  // When removing this function, remove svg.cachedWidth_ and svg.cachedHeight_
+  // from setCachedParentSvgSize.
+  Blockly.utils.deprecation.warn(
```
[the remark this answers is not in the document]
alschmiedt: done

PR #4681, opened 2021-03-05
Fix circular dependencies in shortcut + contextmenu registries

samelhusseini:
## The basics
- [x] I branched from develop

### Resolves

Resolves 115 strict missing require warnings.

### Proposed Changes

Switch registry property to a `getRegistry` method that will create the singleton if it's not there yet.
Require the registry in the item registration file. Require the item registration file from blockly.js / workspace_svg.js as an extra require.

### DIFF
--- a/core/shortcut_registry.js
+++ b/core/shortcut_registry.js
@@ -122,6 +122,7 @@ Blockly.ShortcutRegistry.prototype.unregister = function(shortcutName) {
  */
 Blockly.ShortcutRegistry.prototype.addKeyMapping = function(
     keyCode, shortcutName, opt_allowCollision) {
+  keyCode = String(keyCode);
   var shortcutNames = this.keyMap_[keyCode];
   if (shortcutNames && !opt_allowCollision) {
     throw new Error(
--- a/core/shortcut_registry.js
+++ b/core/shortcut_registry.js
@@ -124,9 +124,9 @@ Blockly.ShortcutRegistry.prototype.unregister = function(shortcutName) {
 
 /**
  * Adds a mapping between a keycode and a keyboard shortcut.
- * @param {string|Blockly.utils.KeyCodes<number>} keyCode The key code for the
- *     keyboard shortcut. If registering a key code with a modifier (ex: ctrl+c)
- *     use Blockly.ShortcutRegistry.getRegistry().createSerializedKey;
+ * @param {string|Blockly.utils.KeyCodes} keyCode The key code for the keyboard
+ *     shortcut. If registering a key code with a modifier (ex: ctrl+c) use
+ *     Blockly.ShortcutRegistry.getRegistry().createSerializedKey;
  * @param {string} shortcutName The name of the shortcut to execute when the
  *     given keycode is pressed.
  * @param {boolean=} opt_allowCollision True to prevent an error when adding a
--- a/core/shortcut_registry.js
+++ b/core/shortcut_registry.js
@@ -43,7 +43,7 @@ Blockly.ShortcutRegistry = function() {
 
 /**
  * Singleton instance of this class.
- * @type {?Blockly.ShortcutItems}
+ * @type {?Blockly.ShortcutRegistry}
  */
 Blockly.ShortcutRegistry.instance_ = null;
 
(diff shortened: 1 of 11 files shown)


core/workspace_svg.js
```diff
@@ -19,6 +19,8 @@ goog.require('Blockly.ConnectionDB');
 /** @suppress {extraRequire} */
 goog.require('Blockly.constants');
 goog.require('Blockly.ContextMenu');
+/** @suppress {extraRequire} */
+goog.require('Blockly.ContextMenuItems');
```
maribethb, line 23: I still don't really love that this magic line is doing the initialization, and a different line somewhere else is doing it for the ShortcutRegistry (and iiuc, shouldn't there also be a similar require in block_svg? I didn't see it). The extraRequire helps, as in the tooling won't remove it, but if you're ever trying to figure out how the registry gets initialized it's confusing (and the current design also somewhat suffers from this too I think, except at least the require is not unused).

I would rather do it in requires.js so we have all of the bootstrapping-type magic requires all in the same place. Alternatively, because it seems like we definitely have other files that do this kind of thing, we could have some kind of bootstrap.js that gets called immediately and could call both `new ContextMenuRegistry()` (so that you wouldn't have to do the api change here, can still access the registry through a property instead) and `ContextMenuItems.registerDefaultItems()` and same for ShortcutRegistry. As it is this will be a breaking api change and that seems like a lot of work just to fix a dependency issue.

Sorry for being stubborn about this change, I just find this pattern really confusing, and would rather not add more of it if we can think of a better/clearer way of doing it.

samelhusseini: This is really just a bi-product of having self registering files, for fields since there’s no requirement to have any specific field included in the bundle we created requires.js as a place to declare what’s included in the bundle.

I’m trying to understand how registering context menu / shortcut items is different from that?

My reasoning for placing the extra require in something like blockly.js / workspace_svg is because requires in requires.js are optional you can pick and choose which you want. I think of the Blockly.js entry point as what are the things needed for Blockly to run, so I’m open to moving these to Blockly.js.

as for the idea of a bootstrap method, we have inject as one, but my question is why not keep them self-registrable? Since they are not attached to any specific workspace but global.

samelhusseini: I can also get behind adding comments to explain what each of the extra requires does. I don’t like that you can tell “why” it’s being required.

rachel-fenichel: > Sorry for being stubborn about this change, I just find this pattern really confusing, and would rather not add more of it if we can think of a better/clearer way of doing it.

Your concern seems warranted here.

Moving the requires into requires.js makes sense to me. Other things in there essentially do this bootstrapping, since plenty of the files end by registering themselves in the base registry.

samelhusseini: On second thought maybe requires.js isn’t really just for optional requires (eg: vertical / horizontal fly out), but more serves as a documented (through comments) of the various modules available to be included. Someone can always exclude something like the connection checker from their bundle, as long as they replace it with their own variation.

moving these to requires.js

maribethb: I understand how having the self-registrable bit makes it so we have to have a magic goog.require *somewhere*, I just don't like having to do it in (seemingly) random files like workspace_svg and would prefer we do it all in the same place. Having them still be self-registrable and moving the requires to either requires.js or Blockly.js would be fine with me. The distinction about requires.js being for optional features makes sense so yes maybe Blockly.js is better.

core/shortcut_registry.js
```diff
@@ -43,8 +39,23 @@ Blockly.ShortcutRegistry = function() {
...
+ * @type {?Blockly.ShortcutRegistry}
+ */
+Blockly.ShortcutRegistry.instance_ = null;
 
-  Blockly.ShortcutItems.registerDefaultShortcuts();
+/**
+ * Gets the shortcut registry registry instance.
+ * @return {!Blockly.ShortcutRegistry} The shortcut registry.
+ */
+Blockly.ShortcutRegistry.getRegistry = function() {
```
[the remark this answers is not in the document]
samelhusseini: It's only used in a test, but just creates a copy of the shortcuts. Renaming to getShortcuts and updating jsdoc to reduce confusion between the two.

core/workspace_svg.js
samelhusseini: Done.

maribethb (approved):
Thanks, Sam, I like this better. 

I do still wonder if we really need to go through the churn of an API change for this (it would work just as well with the previous way, I think?) but I think this is probably a more standard way of dealing with a singleton so if you think it's better, I'm okay with it, just not totally convinced it's worth the hassle.

rachel-fenichel:
+1 to the question about the API change--where possible, fixes to requires should not involve API changes.

samelhusseini:
Agreed, I had tried to avoid the api_change but I made a mistake (missing a rebuild of uncompressed) that left me thinking it was unavoidable.

PR #4685, opened 2021-03-06
Positionables bugfix

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

- Fixes bugs in `getPlugins` logic where plugins are not returned in correct order.
- Updates some comments for clarity.
- Assigns different weights to zoom controls and trashcan

#### Behavior Before Change

`getPlugins` did not order plugins by weight.

#### Behavior After Change

`getPlugins` now returns the plugins correctly ordered.

### Reason for Changes

bugfix of new logic

### Test Coverage

Tested on playground by observing how changing weight of positionable affected how they were positioned.

### DIFF
--- a/core/plugin_manager.js
+++ b/core/plugin_manager.js
@@ -79,9 +79,12 @@ Blockly.PluginManager.prototype.getPlugin = function(id) {
  * @template T
  */
 Blockly.PluginManager.prototype.getPlugins = function(type, sorted) {
-  var plugins = [];
   var typeKey = String(type).toLowerCase();
   var pluginIds = this.typeToPluginId_[typeKey];
+  if (!pluginIds) {
+    return [];
+  }
+  var plugins = [];
   if (sorted) {
     var pluginDataList = [];
     var pluginData = this.pluginData_;
--- a/core/plugin_manager.js
+++ b/core/plugin_manager.js
@@ -21,7 +21,7 @@ goog.provide('Blockly.PluginManager');
 Blockly.PluginManager = function() {
   /**
    * A map of the plugins registered with the workspace, mapped to id.
-   * @type {!Object<string, !Blockly.PluginManager.PluginData>}
+   * @type {!Object<string, !Blockly.PluginManager.PluginDatum>}
    */
   this.pluginData_ = {};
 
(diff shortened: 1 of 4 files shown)


core/plugin_manager.js
```diff
@@ -79,16 +79,29 @@ Blockly.PluginManager.prototype.getPlugin = function(id) {
  * @template T
  */
 Blockly.PluginManager.prototype.getPlugins = function(type, sorted) {
-  var plugins = [];
   var typeKey = String(type).toLowerCase();
   var pluginIds = this.typeToPluginId_[typeKey];
```
alschmiedt, line 83: nit: Could we make `typeToPluginId_` be `typeToPluginId**s**`? The 's' implies returning multiply ids instead of a single id.

core/plugin_manager.js
```diff
@@ -79,16 +79,29 @@ Blockly.PluginManager.prototype.getPlugin = function(id) {
...
-  for (var i = 0, id; pluginIds && (id = pluginIds[i]); i++) {
-    plugins.push(this.pluginData_[id].plugin);
+  if (!pluginIds) {
+    return [];
   }
+  var plugins = [];
   if (sorted) {
-    plugins.sort(function(a, b) {
+    var pluginDataList = [];
```
alschmiedt, line 89: Same question as above, if we changed the type should we change this as well?

core/plugin_manager.js
```diff
@@ -21,7 +21,7 @@ goog.provide('Blockly.PluginManager');
 Blockly.PluginManager = function() {
   /**
    * A map of the plugins registered with the workspace, mapped to id.
-   * @type {!Object<string, !Blockly.PluginManager.PluginData>}
+   * @type {!Object<string, !Blockly.PluginManager.PluginDatum>}
    */
   this.pluginData_ = {};
```
alschmiedt, line 26: Since we changed the type, should we change this as well?

alschmiedt (approved):

core/plugin_manager.js
moniika: sgtm

core/plugin_manager.js
moniika: I changed the type from `PluginData` to `PluginDatum` because datum is the singular of data. My intention was to use datum for a single plugin object bag and data to refer to a collection of them.

PR #4709, opened 2021-03-17
Fix advanced comp

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Advanced compilation test was failing due to 
- Us moving the "test_style_hex1" block out of core
- Because of a type cast to `Blockly.FieldDropdown`.

This bug was a bit weird and I am not entirely sure I understand what is happening but I will try to document at least what I know:
- We were getting an error because somehow [field.suffixField](https://github.com/google/blockly/blob/d7e9eaff0fd533e8bfda905c8eb02a16722af89d/core/input.js#L122) was getting set to true when it shouldn't be(only in advanced compilation).
- We do not get the error when we remove the [type cast](https://github.com/google/blockly/blob/d7e9eaff0fd533e8bfda905c8eb02a16722af89d/core/input.js#L114) OR when we remove [these three lines](https://github.com/google/blockly/blob/82a2dd90a7f965205882875bf7325dd4428fd0e9/core/field_dropdown.js#L162) in FieldDropdown.fromXml.
- I have no idea why, removing those three lines in fromXml works since I put console logs there, and it doesn't look like we actually ever get to that code.
### Proposed Changes
- Add back in the test block. 
- Remove the cast to a field dropdown. From what I can tell this does not break the compiler, and it seems similar to what we do when we check for an optional method on a class before calling it.

### Additional Information
The [original comment](https://github.com/google/blockly/pull/3912/files#r428912607) discussing this.
Alternatives: 
1. Create an interface for prefixField and suffixField
2. Check the field is an instance of Blockly.FieldDropdown. (I tried this and the compiler complained. I can look more into this, however in the comment we say that Blockly.FieldDropdown isn't a required field so we shouldn't assume it is available?)
3. Leave as is.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -449,6 +449,7 @@ function buildAdvancedCompilationTest() {
     'tests/compile/main.js',
     'core/**/**/*.js',
     'blocks/*.js',
+    'tests/compile/test_blocks.js',
     'generators/**/*.js'];
   return gulp.src(maybeAddClosureLibrary(srcs), {base: './'})
     .pipe(stripApacheLicense())
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -447,9 +447,9 @@ function buildLangfiles(done) {
 function buildAdvancedCompilationTest() {
   const srcs = [
     'tests/compile/main.js',
+    'tests/compile/test_blocks.js',
     'core/**/**/*.js',
     'blocks/*.js',
-    'tests/compile/test_blocks.js',
     'generators/**/*.js'];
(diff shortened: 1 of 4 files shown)


scripts/gulpfiles/build_tasks.js
```diff
@@ -449,6 +449,7 @@ function buildAdvancedCompilationTest() {
     'tests/compile/main.js',
     'core/**/**/*.js',
     'blocks/*.js',
+    'tests/compile/test_blocks.js',
```
samelhusseini, line 452: Maybe move up so this is right after main.js (for association)

samelhusseini (approved):

scripts/gulpfiles/build_tasks.js
alschmiedt: done

PR #4722, opened 2021-03-23
Fix missing dependency 

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Blockly games was failing in the maze game. It looks like it was because of [this line](https://github.com/google/blockly/blob/62dc2ff6fa8f211710d3d82f5ab68032792bc65c/core/input.js#L110). It was returning null because `Blockly.FieldLabel` was not registered yet. I'm still trying to figure out why when I run this in core it works fine.

### Proposed Changes
To get this to work I had to add a require for `Blockly.FieldLabel` in `Blockly.Input`. This was removed in our current cleanup (I think) because the constructor is never explicitly used in this file. 

When I add it back, however, I get the below circular dependency: 
Input -> field_label -> field -> gesture -> block_dragger -> insertion_marker_manager -> block -> input

In order to fix this, I moved `Blockly.Block.COLLAPSED_INPUT_NAME` to `Blockly.constants.COLLAPSED_INPUT_NAME`.


rachel-fenichel:
Can you avoid the circular dependency by moving the extra require to requires.js instead of input.js?

alschmiedt:
So my understanding of requires.js (which could definitely be incorrect) is that it is for any files that we don't **need** in core. Any extra fields, themes, optional components, etc. It seems like we need field_label for the inputs to work because we are essentially calling the constructor for a new `Blockly.FieldLabel` by calling `Blockly.fieldRegistry.fromJson({'type:'field_label',...})`.

rachel-fenichel:
Approved, because the bigger question of requires.js is not one we will solve in the next three days.

PR #4725, opened 2021-03-23
Always scroll center on zoom reset

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

Updates zoom reset logic to scroll to center after resetting zoom level.

#### Behavior Before Change

Workspace doesn't scroll to center when zoom reset control is clicked.

#### Behavior After Change

Workspace scrolla to center when zoom reset control is clicked.

### Reason for Changes

This behavior is consistent with how the zoom controls worked in previous releases.

### Test Coverage

* Desktop Chrome

### DIFF
--- a/tests/playground.html
+++ b/tests/playground.html
@@ -110,7 +110,9 @@ function start() {
         readOnly: false,
         rtl: rtl,
         move: {
-          scrollbars: true,
+          scrollbars: {
+            vertical: true
+          },
           drag: true,
           wheel: false,
         },
--- a/tests/playground.html
+++ b/tests/playground.html
@@ -110,9 +110,7 @@ function start() {
         readOnly: false,
         rtl: rtl,
         move: {
-          scrollbars: {
-            vertical: true
-          },
+          scrollbars: true,
           drag: true,
           wheel: false,
         },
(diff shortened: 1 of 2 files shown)


alschmiedt:
I think there still might be a very small change in behavior here. I think the master version still centers the workspace, but develop does not.

**Master**
master_version

**Develop**
develop

moniika:
> I think there still might be a very small change in behavior here. I think the master version still centers the workspace after it scrolls out, but develop does not.
> 
> **Master**
> master_version
> 
> **Develop**
> develop

I've updated the logic to always call `scrollCenter` now, so I think it should be consistent with master.

tests/playground.html
```diff
@@ -110,7 +110,9 @@
         readOnly: false,
         rtl: rtl,
         move: {
-          scrollbars: true,
+          scrollbars: {
+            vertical: true
+          },
```
alschmiedt, line 115: Is this meant to be updated?

alschmiedt (approved):

tests/playground.html
moniika: no. Reverted

PR #4726, opened 2021-03-23
Fix flyout scroll metrics

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

Updates flyout metrics computation.

#### Behavior Before Change

1. Switching between text and logic category flyouts in the playground and scrolling in them caused a weird "bump" behavior.

f52f29e7-1e0a-44f8-a04e-6bc87b014f97

2. The horizontal scrollbar did not have enough padding to the left of the first block in the flyout, if that block was not a value block, when compared to master.

Expected:
4H7P2UATEfQAoo3

Actual:
BTjdHbESghsJWsT

3. (on master) The left edge padding for a horizontal flyout is not adjusted if the first block is not a value block. 
BhUwdWpYwQzsibK

4. (on master) The bottom edge padding for a vertical flyout and the left edge padding for a horizontal flyout is not scaled with zoom as it should.

#### Behavior After Change

1. Flyout scrolling now behaves as expected, without "bump" behavior.

2. Padding to the left of the first block in a horizontal flyout on develop is adjusted so that blocks with and without an output connection are aligned.

3. Padding to the right of blocks in horizontal flyout is now adjusted so that there is the expected padding if the first block  had an output connection.
9E4qcA6PeBWLZZE

4. The bottom and right edge padding of flyouts now scale with zoom.

### Reason for Changes

Bugfix.

### Test Coverage

Tested on multi_playground.

* Desktop Chrome

### DIFF
--- a/core/metrics_manager.js
+++ b/core/metrics_manager.js
@@ -541,16 +541,19 @@ Blockly.FlyoutMetricsManager.prototype.getContentMetrics = function(
 Blockly.FlyoutMetricsManager.prototype.getScrollMetrics = function(
     opt_getWorkspaceCoordinates, opt_viewMetrics, opt_contentMetrics) {
   var contentMetrics = opt_contentMetrics || this.getContentMetrics();
-  var absoluteMetrics = this.getAbsoluteMetrics();
-  var margin = this.flyout_.MARGIN;
+  var margin = this.flyout_.MARGIN * this.workspace_.scale;
   var scale = opt_getWorkspaceCoordinates ? this.workspace_.scale : 1;
-  var leftPadding = contentMetrics.left - absoluteMetrics.left + margin;
+
+  // The left padding isn't just the margin. Some blocks are also offset by
+  // tabWidth so that value and statement blocks line up.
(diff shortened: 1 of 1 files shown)


core/metrics_manager.js
```diff
@@ -541,14 +541,16 @@ Blockly.FlyoutMetricsManager.prototype.getContentMetrics = function(
 Blockly.FlyoutMetricsManager.prototype.getScrollMetrics = function(
     opt_getWorkspaceCoordinates, opt_viewMetrics, opt_contentMetrics) {
   var contentMetrics = opt_contentMetrics || this.getContentMetrics();
+  var absoluteMetrics = this.getAbsoluteMetrics();
   var margin = this.flyout_.MARGIN;
   var scale = opt_getWorkspaceCoordinates ? this.workspace_.scale : 1;
+  var leftPadding = contentMetrics.left - absoluteMetrics.left + margin;
```
alschmiedt, line 547: I would just add a comment explaining why we have to add extra left padding.

alschmiedt (approved):

alschmiedt (approved):

PR #4753, opened 2021-04-04
fix: Fix typing error of generators

ywwhack:
## The basics

- [x] I branched from develop

#### Behavior Before Change
`workspaceToCode` method  is not inferred automatic

image

image

#### Behavior After Change
`workspaceToCode` method  is inferred automatic

image

### Reason for Changes

Fix wrong ts definition for exported generator.

### DIFF
--- a/typings/dart.d.ts
+++ b/typings/dart.d.ts
@@ -12,4 +12,5 @@
 /// <reference path="core.d.ts" />
 
 import * as Blockly from './core';
-export = Blockly.Generator;
+declare const dart: Blockly.Generator;
+export = dart;
--- a/typings/javascript.d.ts
+++ b/typings/javascript.d.ts
@@ -12,4 +12,5 @@
 /// <reference path="core.d.ts" />
 
 import * as Blockly from './core';
-export = Blockly.Generator;
+declare const javascript: Blockly.Generator;
+export = javascript;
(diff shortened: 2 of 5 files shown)


ywwhack:
@googlebot I signed it!

typings/blockly.d.ts
```diff
@@ -6635,6 +6635,31 @@ declare module Blockly.Generator {
      * Category to separate generated function names from variables and procedures.
      */
     var NAME_TYPE: any /*missing*/;
+
+    /**
+     * instance of javascript generator
+     */
+     var javascript: Blockly.Generator;
```
samelhusseini, line 6642: Can you move these instances into the generator typings file itself?

That way if one only includes some generators but not others, they don't get them all

typings/dart.d.ts
```diff
@@ -12,4 +12,4 @@
 /// <reference path="core.d.ts" />
 
 import * as Blockly from './core';
-export = Blockly.Generator;
+export = Blockly.Generator.dart;
```
samelhusseini, line 15: instead: 
```
const dart: Blockly.Generator;
export = dart;
```

typings/blockly.d.ts
ywwhack: Done

samelhusseini (approved):

samelhusseini:
Thank you for the PR!

PR #4781, opened 2021-04-14
Init block dragger

alschmiedt:
- Adds the block drag surface to the registry
- Adds `moveBlockWhileDragging` to the block dragger
- Adds `translateBy` to the block drag surface


core/block_dragger.js
```diff
@@ -216,6 +216,30 @@ Blockly.BlockDragger.prototype.dragBlock = function(e, currentDragDeltaXY) {
...
 };
 
+/**
+ * Updates the location of the block that is being moved by a certain amount.
+ * TODO: Figure out if this is the correct place for this.
+ * @param {number} deltaX Horizontal offset in pixel units.
+ * @param {number} deltaY Vertical offset in pixel units.
+ */
+Blockly.BlockDragger.prototype.moveBlockWhileDragging = function(deltaX, deltaY) {
+  // Negative because we are trying to offset the amount child block dragger
```
maribethb, line 226: I think it would be helpful to add a bit more description to this method. A year from now we are going to look back on this like "what the heck is this even doing" and I think describing what we found out this quarter, and explaining the motivation behind the method and what it's doing (e.g. "When the workspace is dragged, x happens, but when a block is dragged, y happens and you need to account for z"). Maybe we should even add those details to other methods in this class. I'm also happy to work on this separately / not blocking this change. I already need to review what x, y, and z are to make sure I remember it correctly :)

core/interfaces/i_block_drag_surface.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+Blockly.IBlockDragSurfaceSvg.prototype.getGroup;
+
+/**
+ * Get the current blocks on the drag surface, if any (primarily
+ * for BlockSvg.getRelativeToSurfaceXY).
+ * @return {Element} Drag surface block DOM element, or undefined if no blocks
+ * exist.
+ */
+Blockly.IBlockDragSurfaceSvg.prototype.getGroup;
```
maribethb, line 77: I think this should be `getCurrentBlock`

maribethb (approved):

core/block_dragger.js
alschmiedt: Good point and I completely agree. I definitely don't want to have to figure out the block dragger logic again. I will plan to add more documentation once I am sure where this function should live since I am not 100% convinced this is the right place for it just yet.

core/interfaces/i_block_drag_surface.js
alschmiedt: Whoops, good catch :)

core/block_dragger.js
maribethb: Sounds good, we can do a documentation pass before we merge this branch to develop.

PR #4796, opened 2021-04-23
Register block dragger

alschmiedt:
- Updates translateBy so it will work with zoom.
- Removes `moveBlockWhileDragging` since that is going into the plugin
- Adds `BlockDragger` to the registry.
- Adds `getSvgRoot` since I need that SVG in the plugin.


core/gesture.js
```diff
@@ -180,7 +182,7 @@ Blockly.Gesture = function(e, creatorWorkspace) {
 
   /**
    * The object tracking a block drag, or null if none is in progress.
-   * @type {Blockly.BlockDragger}
+   * @type {Blockly.IBlockDragger}
```
maribethb, line 185: This needs a ? since it is nullable

maribethb (approved):

Issue #4826, opened 2021-05-12
Connecting orphan block does not do type checks in correct order


BeksOmega:
### Problem

I'm working on type checking generic params for the nominal-connection-checker, and there's an issue where if you insert a generic param block between two other blocks it blows up. In the bellow case I can no longer interact with the workspace because of errors:
![blows-up](https://user-images.githubusercontent.com/25440652/118054633-76a68900-b33b-11eb-9707-f683f48add3f.gif)

First of all the insertion marker shouldn't even show up there, but I'll file a separate issue for that. This is a different issue because you also get errors when you connect the block programmatically. In the below case I can't set down the block because of errors:
![blows-up-2](https://user-images.githubusercontent.com/25440652/118054644-79a17980-b33b-11eb-8442-478c831f1d65.gif)

I believe this is caused because of how [connect_](https://github.com/google/blockly/blob/3fe56b083c62549cb2bdb17270846bf17ced5d7a/core/connection.js#L106) tries to insert a block X between two blocks A and B:
1) Checks that X and A are compatible
2) Disconnects B from A
3) Tries to connect B to X
    3.1) Checks that B and X are compatible
    3.2) Connects B to X
4) Connects X to A

So in the case above A is the `SelectRandom` block, B is the `GetterList[Dog]` block, and X is the `GetterList[T]` block. We'll call the other `GetterList[Dog]` block C.
1) A is compatible with X because `GetterList[T]` has a common super type with C's `GetterList[Dog]` type.
2) B gets disconnected from A
3) Tries to connect B to X
    3.1) B is compatible with X because `GetterList[Dog]` is compatible with the simple generic `T`
    3.2) Connects B to X
4) Connects X to A

But step 4 is now erroneous. After step 3.2 the type of X is now `GetterList[GetterList[Dog]]`, which is not compatible with the `GetterList[Dog]` check from block C.

### Solution

I think that what needs to happen is steps 3 and 4 need to be swapped. We should only try to reconnect the orphan *after* the new parent has been connected, incase the state of the new parent changes when it is connected.

As you can see, if we do this manually, the nominal-connection-checker performs as expected:
![manually-disconnect](https://user-images.githubusercontent.com/25440652/118054652-7c9c6a00-b33b-11eb-856d-22db08bd74d7.gif)

[...]

Issue #3658, opened 2020-01-25
lastConnectionInRow cause unintuitive re-connection behavior


BeksOmega:
**Describe the bug**

This is actually a lot of little things that I think are weird about how lastConnectionInRow decides if the orphaned block should be reconnected.

This is how the logic works
1. If any block in the chain has no matching connections, the block cannot reconnect.
![BadMiddleType](https://user-images.githubusercontent.com/25440652/73125725-f8b04a00-3f5e-11ea-9078-d6cbfe0907d0.gif)

2. If any block in the chain has multiple matching connections, the block cannot reconnect.
![NoConnectAtEnd](https://user-images.githubusercontent.com/25440652/73125785-ade30200-3f5f-11ea-9eae-a078feecc8a8.gif)
![NoConnectInside](https://user-images.githubusercontent.com/25440652/73125786-b0455c00-3f5f-11ea-9dac-181adab4b259.gif)

3. For 2 filled connections are still seen as available.
![FilledIgnored](https://user-images.githubusercontent.com/25440652/73125758-4f1d8880-3f5f-11ea-931f-03a4691100d6.gif)

But these cases work:
![ConnectAtEnd](https://user-images.githubusercontent.com/25440652/73125759-517fe280-3f5f-11ea-9df4-50ac899d55de.gif)
![ConnectInside](https://user-images.githubusercontent.com/25440652/73125760-547ad300-3f5f-11ea-8531-b4af611fa2df.gif)

**To Reproduce**

1. You can replicate the above scenarios using the connection test blocks on the playground.

**Expected behavior**

I think behavior 3 should definitely be changed because it makes doing things like this inconvenient:
![Math](https://user-images.githubusercontent.com/25440652/73125842-56916180-3f60-11ea-9358-d9f575163c28.gif)

The other behaviors still feel unintuitive, but I'm not sure what the most intuitive behavior would be.

**Additional context**

[TODO link](https://github.com/google/blockly/blob/e2b510662cf60e4dba1e228f3b83549ea694e951/core/renderers/common/renderer.js#L185)

PR #4831, opened 2021-05-15
Change lastConnectionInRow to getPlaceForOrphanedOutput

### LINKED ISSUE #3656: lastConnectionInRow inconsistent with lastConnectionInStack

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #3656

### Proposed Changes

Renames `lastConnectionInRow` and `singleConnection_` to better reflect what they're doing. But the main point of this was to add tests for `lastConnectionInRow`/`getPlaceForOrphanedOutput`.

No behavioral changes.

### Reason for Changes

Before I can tackle #4826 I need to write tests for everything, this is part of that.

### Test Coverage

Manually tested all of the behavior documented by #3658 to ensure nothing changed.

Tested on:
* Desktop Chrome

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -134,7 +134,7 @@ Blockly.Connection.prototype.connect_ = function(childConnection) {
       // Attempt to reattach the orphan at the end of the newly inserted
       // block.  Since this block may be a row, walk down to the end
       // or to the first (and only) shadow block.
-      var connection = Blockly.Connection.lastConnectionInRow(
+      var connection = Blockly.Connection.getPlaceForOrphanedOutput(
           childBlock, orphanBlock);
       if (connection) {
         orphanBlock.outputConnection.connect(connection);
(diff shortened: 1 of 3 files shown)


core/connection.js
```diff
@@ -410,18 +411,19 @@ Blockly.Connection.singleConnection_ = function(block, orphanBlock) {
...
-  while ((connection = Blockly.Connection.singleConnection_(
-      /** @type {!Blockly.Block} */ (newBlock), orphanBlock))) {
-    newBlock = connection.targetBlock();
-    if (!newBlock || newBlock.isShadow()) {
-      return connection;
-    }
-  }
-  return null;
-};
+Blockly.Connection.getPlaceForOrphanedOutput =
```
moniika, line 414: Why this rename? I prefer the name to include "Connection" in it rather than "Place", but I do like adding the "get" in front.
What about `getLastConnectionInRow`?

tests/mocha/connection_test.js
```diff
@@ -784,4 +784,348 @@ suite('Connection', function() {
...
+          'message0': '',
+          'output': 'check',
+        },
+      ]);
+    });
+
+    teardown(function() {
+      workspaceTeardown.call(this, this.workspace);
+      delete Blockly.Blocks['input'];
+      delete Blockly.Blocks['output'];
```
moniika, line 815: These deletes are actually taken care of in the `sharedTestTeardown` call.
How it works is that the `defineBlocksWithJsonArray` is stubbed in `sharedTestSetup` to keep track of the block definitions that need deleted.

Same comment for the other definition deletes that were created through `defineBlocksWithJsonArray`.

core/connection.js
BeksOmega: I'm totally cool with swithcing `Place` to `Connection` but I don't think anything with `Last` or `Row` is an accurate representation of what this function is doing.

For example, pretend you did not know the logic this function is running. Would you expect something called `getLastConnectionInRow` to return the highlighted connection in these stacks of blocks:
middle
end

Or no connection for this stack of blocks:
none

Imo it's not a general use function. It's only used for finding a good connection to reattach an orphan to, so it should be named as such.

This is opposed to `lastConnectionInStack`, which ignores statement inputs and connection checks. This makes it consistant and general purpose, so I think the name is great in that situation.

tests/mocha/connection_test.js
BeksOmega: Stubbing `defineBlocksWithJsonArray` to handle that is so smart! Deletions removed :D

moniika (approved):

BeksOmega:
Sorry I accidently put this against master instead of develop :/ If there's anything I can do to fix it please let me know.

moniika:
> Sorry I accidently put this against master instead of develop :/ If there's anything I can do to fix it please let me know.

My bad. I should have also double checked.
If you could make a new PR against develop. I will revert this change on master.

PR #4841, opened 2021-05-20
Updates block drag surface

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes

- Removes the block drag surface from the registry, since we did not end up needing it for the scroll options plugin.
- Caches the translation of the child drag surface and adds a getter for it. This enables us to not have to get the information directly from the dom every time we scroll in the scroll options plugin.


core/block_drag_surface.js
```diff
@@ -209,6 +222,16 @@ Blockly.BlockDragSurfaceSvg.prototype.getCurrentBlock = function() {
   return /** @type {Element} */ (this.dragGroup_.firstChild);
 };
 
+/**
+ * Gets the translation of the child block surface
+ * This surface is in charge of keeping track of how much the workspace has
+ * moved.
+ * @return {!Blockly.utils.Coordinate} The amount the workspace has been moved.
+ */
+Blockly.BlockDragSurfaceSvg.prototype.getWSTranslation = function() {
```
maribethb, line 231: I think in other places where we abbreviate workspace, we do it as `Ws` e.g. `Blockly.Mutator.findParentWs`, `getWsCoordinate`, etc.

maribethb (approved):

core/block_drag_surface.js
alschmiedt: Yup, good catch. Done.

PR #4842, opened 2021-05-20
Update block dragger

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes

- Changes private --> protected
- Changes package --> public
- Formats the document using clang
- Breaks up larger functions to make overriding easier
- Renames block dragger methods to make them more generic in anticipation of a common i_dragger interface.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -184,9 +184,9 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(
  * @protected
  */
 Blockly.BlockDragger.prototype.shouldDisconnect_ = function(healStack) {
-  return this.draggingBlock_.getParent() ||
+  return !!(this.draggingBlock_.getParent() ||
       (healStack && this.draggingBlock_.nextConnection &&
-       this.draggingBlock_.nextConnection.targetBlock());
+       this.draggingBlock_.nextConnection.targetBlock()));
 };
 
 /**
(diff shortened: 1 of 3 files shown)


core/block_dragger.js
```diff
@@ -242,46 +262,73 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
...
+};
+
+/**
+ * Updates the necessary information to place a block at a certain location.
+ * @param {!Blockly.utils.Coordinate} delta The change in location from where
+ *     the block started the drag to where it ended the drag.
+ * @protected
+ */
+Blockly.BlockDragger.prototype.updateBlockAfterMove_ = function(delta) {
+  // These are expensive and don't need to be done if we're deleting.
```
maribethb, line 279: nit: this comment makes more sense where it was before imo (explaining why we are only calling this when not deleting)

maribethb (approved):
We are currently discussing offline how to handle the name changes... since this isn't going into develop right now I'm fine with however you choose to handle it, we should just make sure to revisit this conversation before merging to develop

I like the new functions! Makes it easier to read as well as override now

core/block_dragger.js
alschmiedt: Updated.

alschmiedt:
I'm going to leave the names for right now so that we can get this in, but we can revisit next week.

PR #4845, opened 2021-05-21
Rename Blockly.<language>.variableDB_ to .nameDB_ and add .nameDB_.getUserNames('VARIABLE') and .nameDB_.getUserNames('PROCEDURE')

NeilFraser:
Should be no user-visible changes.

### DIFF
--- a/core/generator.js
+++ b/core/generator.js
@@ -513,8 +513,8 @@ Blockly.Generator.prototype.scrub_ = function(_block, code, _opt_thisOnly) {
 Blockly.Generator.prototype.finish = function(code) {
   // Optionally override
   // Clean up temporary data.
-  delete Blockly.JavaScript.definitions_;
-  delete Blockly.JavaScript.functionNames_;
+  delete this.definitions_;
+  delete this.functionNames_;
   return code;
 };
 
--- a/core/generator.js
+++ b/core/generator.js
@@ -33,11 +33,6 @@ Blockly.Generator = function(name) {
 };
 
 /**
- * Category to separate generated function names from variables and procedures.
- */
-Blockly.Generator.NAME_TYPE = 'generated_function';
-
-/**
  * Arbitrary code to inject into locations that risk causing infinite loops.
  * Any instances of '%1' will be replaced by the block ID that failed.
  * E.g. '  checkTimeout(%1);\n'
--- a/core/generator.js
+++ b/core/generator.js
@@ -505,8 +505,8 @@ Blockly.Generator.prototype.scrub_ = function(_block, code, _opt_thisOnly) {
 
 /**
  * Hook for code to run at end of code generation.
- * Subclasses may override this, e.g. to prepend the generated code with the
- * variable definitions.
+ * Subclasses may override this, e.g. to prepend the generated code with import
+ * statements or variable definitions.
  * @param {string} code Generated code.
  * @return {string} Completed code.
  */
--- a/core/generator.js
+++ b/core/generator.js
@@ -402,24 +402,42 @@ Blockly.Generator.prototype.definitions_;
 Blockly.Generator.prototype.functionNames_;
 
 /**
- * A database of variable names.
+ * A database of variable and procedure names.
  * @type {Blockly.Names}
  * @protected
  */
-Blockly.Generator.prototype.variableDB_;
+Blockly.Generator.prototype.nameDB_;
 
 /**
- * Define a function to be included in the generated code.
+ * @deprecated 'variableDB_' was renamed to 'nameDB_' (May 2021).
+ */
+Object.defineProperty(Blockly.Generator.prototype, 'variableDB_', {
+  get: function() {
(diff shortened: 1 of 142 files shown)


NeilFraser:
The net result of all of these commits is that in any block's generator one can now call either one of:
```
Blockly.JavaScript.nameDB_.getUserNames('VARIABLE')
Blockly.JavaScript.nameDB_.getUserNames('PROCEDURE')
```
and get an array of all the user's variable or procedure names.

core/generator.js
```diff
@@ -402,32 +402,50 @@ Blockly.Generator.prototype.definitions_;
...
+Object.defineProperty(Blockly.Generator.prototype, 'variableDB_', {
+  get: function() {
+    console.warn('Deprecated use of "variableDB_"; change to "nameDB_"');
+    return this.nameDB_;
+  },
+  set: function(x) {
+    console.warn('Deprecated use of "variableDB_"; change to "nameDB_"');
+    this.nameDB_ = x;
+  }
+});
```
BeksOmega: This seems like a nice way to deprecate something! I'll have to keep this in mind in the future.

core/generator.js
```diff
@@ -402,32 +397,59 @@ Blockly.Generator.prototype.definitions_;
...
+
+Object.defineProperty(Blockly.Generator.prototype, 'variableDB_', {
+  /**
+   * Getter.
+   * @deprecated 'variableDB_' was renamed to 'nameDB_' (May 2021).
+   * @this {Blockly.Generator}
+   * @return {Blockly.Names} Name database.
+   */
+  get: function() {
+    console.warn('Deprecated use of "variableDB_"; change to "nameDB_"');
```
maribethb, line 414: We have the Blockly.utils.deprecation.warn function, which would be good to use for consistency and so we can more easily track what we have scheduled for deletion

maribethb (approved):
It was helpful that the cleanup changes were in separate commits, but it'd be even easier to review if they were in separate PRs. This also matters because your PR title will be in the release notes, and having variable and procedure names available on demand might be something devs are interested in seeing at a glance from the notes.

core/generator.js
```diff
@@ -402,32 +397,59 @@ Blockly.Generator.prototype.definitions_;
...
  */
-Blockly.Generator.prototype.variableDB_;
+Blockly.Generator.prototype.nameDB_;
+
+Object.defineProperty(Blockly.Generator.prototype, 'variableDB_', {
+  /**
+   * Getter.
+   * @deprecated 'variableDB_' was renamed to 'nameDB_' (May 2021).
+   * @this {Blockly.Generator}
+   * @return {Blockly.Names} Name database.
```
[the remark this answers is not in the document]
NeilFraser: Fixed here, and a few dozen other places.

core/generator.js
NeilFraser: Fixed, and changed Blockly.utils.deprecation.warn to also handle properties, not just functions.

NeilFraser:
> It was helpful that the cleanup changes were in separate commits, but it'd be even easier to review if they were in separate PRs. This also matters because your PR title will be in the release notes, and having variable and procedure names available on demand might be something devs are interested in seeing at a glance from the notes.

The flip side is that splitting this into multiple PRs would result in one of:
1) A set of mostly serial PRs that each block the next and takes much longer
2) An unmanageable set of merge collisions
3) Pressure to not fix stuff all over the codebase.

I'm concerned that the current process around release notes is resulting in lower quality code since there's a penalty for resolving side issues on the spot upon discovery.  Is there a way we can build informative release notes without discouraging off-topic edits?

maribethb:
> > It was helpful that the cleanup changes were in separate commits, but it'd be even easier to review if they were in separate PRs. This also matters because your PR title will be in the release notes, and having variable and procedure names available on demand might be something devs are interested in seeing at a glance from the notes.
> 
> The flip side is that splitting this into multiple PRs would result in one of:
> 
> 1. A set of mostly serial PRs that each block the next and takes much longer
> 2. An unmanageable set of merge collisions
> 3. Pressure to not fix stuff all over the codebase.
> 
> I'm concerned that the current process around release notes is resulting in lower quality code since there's a penalty for resolving side issues on the spot upon discovery. Is there a way we can build informative release notes without discouraging off-topic edits?

Makes sense. This is one of those times I look longingly over at internal-Google systems and remember how easy it is to have multiple changes going at once. The good ol' days.

I probably would have moved all of the cleanup changes to their own PR, and have a separate PR for the non-cleanup changes around name generation to get that balance between clarity/ease of reviewing and workflow, but I realize everyone sets that threshold differently, so it's not a big deal. 

I do think it's a good idea to update the PR title to mention the generator change though, as it's most relevant from a release notes perspective. I'm not sure how best to improve them, since it needs to be an automatic process, and grabbing every commit title is too granular, imo. We could look into it in our next "non-feature week" though! There's probably some tooling out there for it, or we could have a section in the PR template for release notes and parse that out later.

PR #4855, opened 2021-05-27
Add support for IAutoHideable

moniika:
## The basics

- [x] I branched from develop

### Resolves

Part of adding support in core for Backpack.
https://github.com/google/blockly-samples/issues/233

### Proposed Changes

- Adds new plugin interface `IAutoHideable` for components that should be automatically hidden when `Blockly.hideChaff` is called.
- Updates jsDoc for `Blockly.hideChaff` to be more accurate
- Registers trashcan and toolbox as `IAutoHideable`
- Moves registration of zoom controls and trashcan to the init method

#### Behavior Before Change

no change in behavior

#### Behavior After Change

no change in behavior

### Reason for Changes

Adds support for Backpack (otherwise Backpack would need to monkeypatch).

### Test Coverage

Tested manually in playground.

* Desktop Chrome


core/blockly.js
```diff
@@ -300,20 +300,14 @@ Blockly.hideChaff = function(opt_allowToolbox) {
   Blockly.Tooltip.hide();
   Blockly.WidgetDiv.hide();
   Blockly.DropDownDiv.hideWithoutAnimation();
+
   if (!opt_allowToolbox) {
```
moniika, line 304: I just noticed this was a parameter in hideChaff, which introduces some complication.
I'm not sure how to document that autoCloseables are only hidden it this is false. The documentation for this parameter isn't entirely accurate either because it ALSO prevents the trashcan from automatically closing.

moniika: Updated to pass parameter through and jsDoc updated.

core/blockly.js
```diff
@@ -294,27 +294,20 @@ Blockly.onContextMenu_ = function(e) {
 
 /**
  * Close tooltips, context menus, dropdown selections, etc.
- * @param {boolean=} opt_allowToolbox If true, don't close the toolbox.
+ * @param {boolean=} opt_onlyClosePopups Whether only popups should be closed.
```
maribethb, line 297: I would also add something to say specifically that flyouts won't be closed. After learning that people do actually read the jsdoc I'm trying to be more explicit in providing directions/info for developers :)

core/interfaces/i_autohideable.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Interface for a component that can be automatically hidden.
+ * @extends {Blockly.IPlugin}
+ * @interface
+ */
+Blockly.IAutoHideable = function() {};
+
+/**
+ * Hides the component. Called in Blockly.hideChaff.
```
maribethb, line 29: Similarly I think this could use some more documentation, about how to use/register it, and what the implementer should do in this method. But I think if you wanted to do that separately, that's fine too since perhaps the PluginManager api is changing.

core/workspace_svg.js
```diff
@@ -2261,7 +2249,7 @@ Blockly.WorkspaceSvg.prototype.getScale = function() {
  * @package
  */
 Blockly.WorkspaceSvg.prototype.scroll = function(x, y) {
-  Blockly.hideChaff(/* opt_allowToolbox */ true);
+  Blockly.hideChaff(true);
```
maribethb, line 2252: I would leave the jsdoc comment with the parameter name, personally, just update it to be correct. so that you don't have to go look at the method signature for this to figure out what true means.

maribethb (approved):

core/interfaces/i_autohideable.js
moniika: I expect to add documentation, probably during bugbash, when adding documentation for the `PluginManager`.

core/workspace_svg.js
moniika: Done.

core/blockly.js
moniika: Updated.

PR #4863, opened 2021-06-01
Fix drag surface

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
- Returns a copy of the coordinate in block drag surface
- Runs clan format on the block drag surface.


core/block_drag_surface.js
```diff
@@ -229,7 +229,7 @@ Blockly.BlockDragSurfaceSvg.prototype.getCurrentBlock = function() {
  * @return {!Blockly.utils.Coordinate} The amount the workspace has been moved.
```
maribethb, line 232: I would add a comment that you're returning a copy on purpose so someone doesn't try to come "fix" it like I did haha

maribethb (approved):

core/block_drag_surface.js
alschmiedt: Done.

PR #4873, opened 2021-06-04
Remove row healing for block deletion #4832

### LINKED ISSUE #4832: Context menu Delete X blocks preview does not match behavior
**Describe the bug**

When you open up the context menu and select the "Delete X blocks" option, X blocks may not actually be deleted, because the stack is healed.
delete-blocks

**To Reproduce**

1. Load the following XML into the playground:
```
  <variables>
    <variable>test</variable>
  </variables>
  <block type="variables_set">
    <field name="VAR">test</field>
    <value name="VALUE">
      <block type="text_join">
        <value name="ADD0">
          <block type="text">
            <field name="TEXT"></field>
          </block>
        </value>
      </block>
    </value>
  </block>
</xml>
```
2) Right click on the middle block and select "Delete 2 Blocks"
3) Observe how only one block is deleted due to healing the stack.

**Expected behavior**

Either both blocks should be deleted, or the context menu should show the correct preview. I recommend the former, because you should be able to achieve that easily by passing healStack as false to dispose.

**Additional context**

Reported here: https://groups.google.com/g/blockly/c/rRB1IJy_iWo

jschanker:
## The basics

- [X] I branched from develop

### Proposed Changes

Selecting Delete X blocks from the context menu will no longer result in an attempt to heal a row of blocks.  Attempted stack healing will persist.

#### Behavior Before Change

If a row could be healed when selecting a block for deletion from its context menu, it would.  However, the number in the context menu delete message reflected that row healing would not occur.

#### Behavior After Change

Rows are no longer healed, and the number X in Delete X blocks on the context menu now correctly reflects the number of blocks that are actually deleted.

### Reason for Changes

From the discussion in #4832, it was decided that we should continue to try to heal vertical stacks but not rows.

### Test Coverage

Tested locally by creating and connecting blocks where both rows/stacks had the opportunity to be healed and then deleted them to make sure that rows were no longer healed but stacks were.  Should I add tests to https://github.com/google/blockly/blob/master/tests/mocha/contextmenu_items_test.js#L304 ?

### Documentation

No additional documentation required.

### Additional Information

#4832


alschmiedt (commented):
Thanks for the PR!

I think we actually want this logic in [Blockly.deleteBlock](https://github.com/google/blockly/blob/fc6a1be3d982349292b630d2818b3b5509688d5f/core/blockly.js#L215) function and then we can call that from here. This way both [deleting with the keyboard](https://github.com/google/blockly/blob/fc6a1be3d982349292b630d2818b3b5509688d5f/core/shortcut_items.js#L79) and deleting with the context menu works the same way.

alschmiedt (approved):
Lgtm

Thanks!

PR #4878, opened 2021-06-08
Fix removal of spaces near parens inside strings

NeilFraser:
Extra spaces should only be stripped from the inside of paren tokens.  Parens in strings (and other static content) should not be edited.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1424,16 +1424,19 @@ Blockly.Block.prototype.toString = function(opt_maxLength, opt_emptyToken) {
 
   // Run through our text array and simplify expression to remove parentheses
   // around single field blocks.
-  for (var i = 2, l = text.length; i < l; i++) {
+  // E.g. ['repeat', '(', '10', ')', 'times', 'do', '?']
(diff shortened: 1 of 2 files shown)


core/block.js
```diff
@@ -1424,16 +1424,19 @@ Blockly.Block.prototype.toString = function(opt_maxLength, opt_emptyToken) {
...
       text[i - 2] = text[i - 1];
       text.splice(i - 1, 2);
-      l -= 2;
     }
   }
 
-  // Join the text array, removing spaces around added paranthesis.
-  text = text.join(' ').replace(/(\() | (\))/gmi, '$1$2').trim() || '???';
+  // Join the text array, removing spaces around added parentheses.
+  text = text.reduce(function(acc, value) {
```
BeksOmega, line 1436: Are we allowed to use reduce in core? If so, I am super excited about that :D

NeilFraser: I don't see why not.  Supported in IE9.  Already used in Blockly.utils.string.shortestStringLength

alschmiedt (approved):

PR #4884, opened 2021-06-09
Add removeComponent to ComponentManager

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

- adds functionality to `ComponentManager` to remove registered components
- adds calls in `dispose` to remove components from `ComponentManager`
- fixes requires
- updates some jsdoc types


core/flyout_base.js
```diff
@@ -321,6 +322,8 @@ Blockly.Flyout.prototype.init = function(targetWorkspace) {
  */
 Blockly.Flyout.prototype.dispose = function() {
   this.hide();
+  this.workspace_.getComponentManager().removeComponent(
+      'flyout' + this.workspace_.id);
```
maribethb, line 326: Since you are creating this id twice, I'd rather it be a property. If anyone ever changes the id when you add the capability there's a decent chance they might forget and break it here (or if someone overrides the corresponding `add` method to change the id, they'd also have to override this just to change the name)

maribethb (approved):

core/component_manager.js
```diff
@@ -70,20 +70,39 @@ Blockly.ComponentManager.prototype.addComponent = function(
...
+ * Removes a component.
+ * @param {string} id The ID of the component to remove.
+ * @template T
+ */
+Blockly.ComponentManager.prototype.removeComponent = function(id) {
+  var component = this.componentData_[id];
+  if (!component) {
+    return;
+  }
+  for (var i = 0, capability; (capability = component.capabilities[i]); i++) {
```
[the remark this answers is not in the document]
moniika: I'll update it because changing it also makes the line shorter.

core/flyout_base.js
moniika: Done.

PR #4893, opened 2021-06-10
Block drag

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves

Adds necessary changes to core for the blockly samples scroll-options plugin.

### Proposed Changes
 Changes are: 
- Cache the XY value of the child drag surface on the block_drag_surface and provide a public getter for it. 
- Adds a `translateBy` public method to translate the parent drag surface by a delta.
- Adds block dragger to the registry.
- Refactors block dragger.


core/block_dragger.js
```diff
@@ -69,24 +72,24 @@ Blockly.BlockDragger = function(block, workspace) {
...
    */
   this.startXY_ = this.draggingBlock_.getRelativeToSurfaceXY();
 
   /**
    * A list of all of the icons (comment, warning, and mutator) that are
    * on this block and its descendants.  Moving an icon moves the bubble that
    * extends from it if that bubble is open.
-   * @type {Array<!Object>}
-   * @private
+   * @type {Array.<!Object>}
```
maribethb, line 91: This `.` should be removed

core/block_dragger.js
```diff
@@ -159,34 +163,53 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY,
...
-    toolbox.addStyle(style);
-  }
+  this.updateToolboxStyle_(false);
+};
+
+/**
+ * Whether or not we should disconnect the block when a drag is started.
+ * @param {boolean} healStack Whether or not to heal the stack after
+ *     disconnecting.
+ * @returns {boolean} True to disconnect the block, false otherwise.
```
maribethb, line 182: should be `@return`

core/block_dragger.js
```diff
@@ -256,46 +279,73 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
...
+ * Adds or removes the style of the cursor for the toolbox.
+ * This is what changes the cursor to display an x when a deletable block is
+ * held over the toolbox.
+ * @param {boolean} isEnd True if we are at the end of a drag, false otherwise.
+ * @protected
+ */
+Blockly.BlockDragger.prototype.updateToolboxStyle_ = function(isEnd) {
+  var toolbox = this.workspace_.getToolbox();
+
+  if (toolbox) {
```
maribethb, line 329: I know this was already here and you just moved it into a function. So we don't have to change anything rn. But it seems weird that the BlockDragger does this. It should just notify the toolbox and let the toolbox worry about its own styles.

Once we have delete area components, would this change to accommodate all delete areas? Then it makes even more sense, for each delete area, call some method on it, and let them figure out what to do. I'm not actually sure if this does matter for all delete areas though. There is the somewhat similar `updateCursorDuringBlockDrag` and I'm not sure how they are related/different.

maribethb (approved):
small changes and a question/comment for the future

core/block_dragger.js
moniika: Do you need this method to be `protected` (do you override it)? I think it could be potentially refactored into the toolbox using the `IDragTarget` API.

moniika: I was able to get this logic factored out into the `Toolbox` class using the changes in https://github.com/google/blockly/pull/4888. I can make a follow up PR after this one so we can discuss pulling it out.

core/block_dragger.js
alschmiedt: done

core/block_dragger.js
alschmiedt: done

PR #4898, opened 2021-06-11
Update the registered component ids

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

Updates the registered component id for `zoom_controls`, `trashcan` and `toolbox`.

### Reason for Changes

Since there is only ever one of each of these kinds of components, these ids do not need to be generated. 
These ids are easier to use to get the component by an external developer.

### Additional Information

This change would mean that if someone wanted to have multiple of these kind of components, then they would need to change the ids of the two instances so that they would be unique (ids must be unique).

It might make sense to also add a constant for a default id for developers to use to get these components from the `ComponentManager`.

### DIFF
--- a/core/toolbox/toolbox.js
+++ b/core/toolbox/toolbox.js
@@ -65,7 +65,8 @@ Blockly.Toolbox = function(workspace) {
   this.workspace_ = workspace;
 
   /**
-   * The unique id for this component.
+   * The unique id for this component that is used to register with the
+   * ComponentManager.
    * @type {string}
    */
   this.id = 'toolbox';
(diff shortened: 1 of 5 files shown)


core/toolbox/toolbox.js
```diff
@@ -68,7 +68,7 @@ Blockly.Toolbox = function(workspace) {
    * The unique id for this component.
    * @type {string}
    */
-  this.id = Blockly.utils.genUid();
+  this.id = 'toolbox';
```
alschmiedt, line 71: I think I would prefer either this was updated to `componentId` or we update the comment to say this is used to register the item with the component manager.

alschmiedt (approved):

core/toolbox/toolbox.js
moniika: I've updated the jsdoc to reference the `ComponentManager`.

Issue #2114, opened 2018-11-14
Debug and reenable macOS CI on GitHub Actions


rachel-fenichel:
Generator tests require firefox instead of chrome, but our current setup has issues with geckodriver on osx.  Low priority because we do regularly build on osx, and because ubuntu tests are running.

cpcallen:
Working on this in PR #4903.  Looks like the generator tests are fine on macOS, but mocha tests are failing in CI (not locally).

cpcallen:
I looked at this a bit but I am unable to figure out what is going wrong.  Comparing test logs from my own machine vs. [our CI servers](https://github.com/google/blockly/pull/4903/checks?check_run_id=2802943075) it looks like the CI servers are actually running MORE mocha tests than get run locally.  Here are the test output files, trimmed to start with the mocha tests and modified to remove timestamps and element IDs for easier diffing:

* [test-local.txt](https://github.com/google/blockly/files/6639547/test-local.txt) - output from `npm run test:run` on my machine.
* [test-ci.txt](https://github.com/google/blockly/files/6639548/test-ci.txt) - taken from [the CI logs for PR #4903)](https://github.com/google/blockly/pull/4903/checks?check_run_id=2802943075) doing the same thing.

Could it be that tests are passing locally *only because we are not running all of them*?

I'm going to leave this for the time being because I do not understand mocha, selenium, webdriver, etc.

cpcallen:
One further possible clue about what is going on: it seems that the mocha tests are flaky on macOS, and sometimes succeed and sometimes not; e.g., for the same commit:

* [v12 passed](https://github.com/google/blockly/pull/4903/checks?check_run_id=2804483638), and
* [v14 passed](https://github.com/google/blockly/pull/4903/checks?check_run_id=2804483673), but
* [v16 failed](https://github.com/google/blockly/pull/4903/checks?check_run_id=2804483713)

I don't think its v16 specific, however, because at earlier commits v12 passed but v14 and v16 failed.

rachel-fenichel:
To work on this bug:
- Add `macos-latest` back in [`browser_test.yml`](https://github.com/google/blockly/blob/osd/.github/workflows/browser_test.yml#L19)
- Create a pull request
- [Look at logs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/using-workflow-run-logs) to verify that the tests ran on both ubuntu and macos.
  - Usually creating a PR kicks off the actions, and there will be links to the logs next to each check.

[...]

PR #4903, opened 2021-06-11
CI updates: bump node, prepare to reenable macOS

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

~~This should fix the CI issue with PR #4812~~, and also does some preparatory work for #2114.

### Proposed Changes

Update our CI configuration:

- Bump node.js versions: ~~remove v10 and~~ add v16, per https://nodejs.org/en/about/releases/
- Update the configuration for macOS, but leave it disabled for now.

### Test Coverage

Will be improved.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
         # TODO (#2114): re-enable osx build.
         # os: [ubuntu-latest, macos-latest]
         os: [ubuntu-latest]
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [10.x, 12.x, 14.x, 16.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,12 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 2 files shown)


cpcallen:
I'm going to give up fixing #2114 for now and re-disable CI on macOS.

rachel-fenichel:
Per discussion in team meeting, add 16.x but leave 10.x in place--we don't need to deprecate it immediately.

cpcallen:
> Per discussion in team meeting, add 16.x but leave 10.x in place--we don't need to deprecate it immediately.

Done.

alschmiedt:
Sorry wasn't here for the team discussion so might be a bit behind :) 

If we do not remove node v10 does this still fix #4812? It looks like the problem there is one of the dependencies doesn't support v10 anymore.

cpcallen:
> If we do not remove node v10 does this still fix #4812? It looks like the problem there is one of the dependencies doesn't support v10 anymore.

It does not (and I've updated the description accordingly).  I'm unclear on what the plan is for #4812, but my understanding of the discussion was that we do not (yet) want to drop support for node v10.

rachel-fenichel (approved):

rachel-fenichel:
The plan for #4812 is to just let it sit for now--it's not a security change, just standard dependabot bumping, so updating yargs isn't urgent. Keeping node v10 is more important.

PR #4906, opened 2021-06-14
Use selenium-standalone-service to automatically start and stop Selenium when running the test suite.

gonfunko:
This change also migrates the generator tests to run in Chrome instead of Firefox.

## The basics

- [X] I branched from develop

### Proposed Changes

This PR uses the selenium-standalone-service module to automatically start and stop Selenium when running test cases, instead of the old commands in package.json to coordinate doing so. It also uses Chrome to run the generator tests (consistent with the Mocha tests) instead of Firefox.

#### Behavior Before Change

`npm run test` would run the test suite, but hang and never terminate since it was blocking on a child Selenium process.

#### Behavior After Change

`npm run test` runs the test suite and terminates when it is done.

### Reason for Changes

The test suite had to be manually stopped with ^C.

### Test Coverage

Change is to the test suite itself; ran the test suite and confirmed that all tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,12 +84,12 @@
     "rimraf": "^3.0.2",
     "selenium-standalone": "^6.17.0",
     "through2": "^4.0.2",
+    "typescript": "^4.3.2",
     "typescript-closure-tools": "^0.0.7",
     "webdriverio": "^7.0.3",
     "yargs": "^16.0.3"
   },
   "dependencies": {
-    "jsdom": "15.2.1",
-    "typescript": "^4.3.2"
+    "jsdom": "15.2.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -40,12 +40,8 @@
     "publish:beta": "gulp publishBeta",
     "recompile": "gulp recompile",
     "release": "gulp gitCreateRC",
-    "test": "concurrently 'npm run test:prepare' 'sleep 5 && npm run test:run'",
-    "test:generators": "concurrently 'npm run test:prepare' 'sleep 5 && tests/scripts/run_generators.sh'",
-    "test:prepare": "npm run test:setupselenium && npm run test:startselenium",
-    "test:run": "tests/run_all_tests.sh",
-    "test:setupselenium": "selenium-standalone install --config=./tests/scripts/selenium-config.js",
-    "test:startselenium": "selenium-standalone start --config=./tests/scripts/selenium-config.js",
+    "test": "tests/run_all_tests.sh",
+    "test:generators": "tests/scripts/run_generators.sh",
     "test:compile:advanced": "gulp buildAdvancedCompilationTest",
     "typings": "gulp typings",
     "updateGithubPages": "gulp gitUpdateGithubPages"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -92,6 +89,7 @@
     "yargs": "^16.0.3"
   },
   "dependencies": {
-    "jsdom": "15.2.1"
+    "jsdom": "15.2.1",
+    "typescript": "^4.3.2"
```
samelhusseini, line 93: Did this make it in by accident? or is typescript a required dependency here?

samelhusseini (approved):

package.json
gonfunko: It was added automatically by the tooling; the dependencies tab on the package's page on npm doesn't indicate typescript is a dependency, but I'm not sure how I'd otherwise tell?

samelhusseini: Yeah I'm curious how that was added. Try removing it and re-running `npm install` and seeing if everything works fine without it.

tests/generators/run_generators_in_browser.js
```diff
@@ -40,14 +40,18 @@ async function runLangGeneratorInBrowser(browser, filename, codegenFn) {
 async function runGeneratorsInBrowser() {
   var options = {
     capabilities: {
-      browserName: 'firefox'
+      browserName: 'chrome',
```
rachel-fenichel, line 43: Update the comments for this function to not say that it's using firefox.

Looking forward, should this entire file/set of functions move into gulp scripts? Is there any reason we should keep it as a separate thing? It looks kind of like it went manual testing -> bash script -> node script.

package.json
gonfunko: I tried removing it, npm install was happy, but re-running the tests added it back.

tests/generators/run_generators_in_browser.js
gonfunko: Done, and perhaps? I'm not familiar enough with the tooling to have a useful opinion, but simplifying/using the standard approach SGTM.

package.json
cpcallen: I suspect this is related to `npm test` regenerating the TypeScript typing files in `typings/`.

I'm unclear on whether rebuilding these during test is desirable or not, but in any case if `typescript` is needed it ought to be installed in the `devDependencies` section rather than `dependencies` and—and probably permanently, so that we can be alerted to vulnerabilities, rather than being added ad-hoc by the build/test scripts.

cpcallen:
@googlebot I consent.

package.json
gonfunko: Yeah, moving it to devDependencies seems to do the trick - it doesn't get added back to dependencies that way. Updated accordingly!

samelhusseini: Oh yeah that makes sense @cpcallen.
Re-building is important as it tells you whether there is a problem with the closure to typescript conversion. I'm ok with adding that in devDependencies given that requirement.

cpcallen: It would probably be good to find the code that was previously adding it, and remove it.

Issue #4818, opened 2021-05-10
Editing text field impossible on iPad while external keyboard connected


torenord:
**Describe the bug**

Touching a text field on iPad while external keyboard is connected causes the field to immediately lose focus making editing impossible.

A workaround is making `Blockly.utils.userAgent.IPAD` true by, for instance, turning the Safari setting "Request Desktop Website" off or setting it manually. A prompt then opens instead of the inline editor and editing is possible. (`Blockly.utils.userAgent.IPAD` is false on iPad due to the user agent string sent from Safari on iPad.)

**To Reproduce**

1. Open Safari on iPad with external keyboard connected
2. Go to https://blockly-demo.appspot.com/static/demos/fixed/index.html
3. Touch a block's text field
4. The field receives focus and then immediately loses it

**Expected behavior**

The text field should keep focus after touch.

**Smartphone:**
 - Device: iPad (8th generation)
 - OS: 14.4
 - Browser: Safari
 - Version: 14.0.3
 - User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Safari/605.1.15

PR #4911, opened 2021-06-14
Updated user agent parsing to detect iPads posing as desktop Safari.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/4818

### Proposed Changes

Updates user agent parsing to detect iPads posing as desktop Safari. Since iOS 13, non-mini iPads use the same user agent as desktop Safari. This change checks for multitouch (which doesn't exist on Macs) to distinguish iPads. I looked into using the desktop behavior (inline fields instead of a modal text field), but that appears to be bordering on infeasible due to mobile Safari playing games with calls to focus().

#### Behavior Before Change
Editing values in text fields in blocks on an iPad running iOS 13 or greater with an external keyboard was not possible due to focus being lost as soon as the text field was tapped.

#### Behavior After Change
The standard modal text input dialog appears.

### Test Coverage
I have manually verified the fix on an iPad running iOS 14.6 with an external keyboard attached. I also verified that it still worked on desktop Safari and an iPhone.

### DIFF
--- a/core/utils/useragent.js
+++ b/core/utils/useragent.js
@@ -93,9 +93,11 @@ Blockly.utils.userAgent.MOBILE;
       !Blockly.utils.userAgent.EDGE;
 
   // Platforms.  Logic from:
-  // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/platform.js
+  // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/platform.js and
+  // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/extra.js
   Blockly.utils.userAgent.ANDROID = has('Android');
-  Blockly.utils.userAgent.IPAD = has('iPad');
+  Blockly.utils.userAgent.IPAD = has('iPad') ||
+      has('Macintosh') && navigator.maxTouchPoints > 0;
   Blockly.utils.userAgent.IPOD = has('iPod');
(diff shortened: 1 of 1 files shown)


NeilFraser (approved):
I added a commit to this PR, but it was written blind and is untested.  If this looks good to you, LGTM.

core/utils/useragent.js
```diff
@@ -96,8 +96,8 @@ Blockly.utils.userAgent.MOBILE;
   // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/platform.js and
   // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/extra.js
   Blockly.utils.userAgent.ANDROID = has('Android');
-  Blockly.utils.userAgent.IPAD = has('iPad') ||
-      has('Macintosh') && navigator.maxTouchPoints > 0;
+  var maxTouchPoints = Blockly.utils.global['navigator'] && Blockly.utils.global['navigator']['maxTouchPoints'];
+  Blockly.utils.userAgent.IPAD = has('iPad') || has('Macintosh') && maxTouchPoints > 0;
```
cpcallen, line 100: Line length (x2).

cpcallen (changes requested):

core/utils/useragent.js
gonfunko: Done.

PR #4923, opened 2021-06-16
Add option to hide the inline/external menu option

### LINKED ISSUE #3077: Prevent blocks from toggling inline/external
It used to be that if a block was `this.isDeletable() && this.isMovable()` then the inline/external menu row would appear (assuming the block's topology qualified).  This original behaviour can be seen in the archived version of Blockly Games where the green puzzle pieces do *not* have an inline context menu option:
https://2-dot-blockly-games.appspot.com/puzzle

The current code is that the block need only be `this.isMovable()` for the menu row to appear.  This new behaviour can be seen in the live version of Blockly Games where the green puzzle pieces do have an inline context menu option:
https://blockly.games/puzzle

It is desirable to be able to eliminate this menu option for simple Blockly instances such as Puzzle.  The original mechanism worked, but doesn't make a lot of sense.  The new mechanism doesn't have any options for removing the menu option.  What's the best way to restore this functionality?

I think the key question is does it need to be per block, per block type, or can it just be across the workspace?  Personally I think workspace is fine, but don't care one way or another.

NeilFraser:
Fixes #3077

New option will need to be added to this page upon release:
https://developers.google.com/blockly/guides/configure/web/configuration_struct

```
inlineExternal
boolean
Allows block inputs to be toggled between inline or external. Defaults to `true` if the toolbox has categories, `false` otherwise.
```

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -523,7 +523,8 @@ Blockly.BlockSvg.prototype.moveToDragSurface = function() {
 
 /**
  * Move a block to a position.
- * @param {Blockly.utils.Coordinate} xy The position to move to in workspace units.
+ * @param {!Blockly.utils.Coordinate} xy The position to move to in workspace
+ *     units.
  */
 Blockly.BlockSvg.prototype.moveTo = function(xy) {
   var curXY = this.getRelativeToSurfaceXY();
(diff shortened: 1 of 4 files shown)


alschmiedt:
Since we have the new context menu registry this can be accomplished by doing the below.

```
Blockly.ContextMenuRegistry.registry.unregister('blockInline')
```

I think this is preferable to adding another item to the options struct. Would this work for your use case or is there a piece I am missing?

BeksOmega:
Closing this as stale

Issue #4912, opened 2021-06-14
Flyouts do not recalculate delete area


BeksOmega:
**Describe the bug**

If you have a simple toolbox (ie just a flyout) you can modify the blocks within the flyout before dragging them out. When you modify the blocks, this can cause the size of the flyout to change. However, the delete area is still the old size.

**To Reproduce**

1. Go to the advanced playground and open the simple toolbox.
2. Drag any block out of the flyout.
3. Go down to the last block in the flyout and change the dropdown from "letter #" to "letter # from end" to make the flyout wider.
4. Drag the block you created back over the flyout, and observe how the delete area is still the old size.

Note: if you switch steps 2 and 3, it doesn't reproduce. No clue what's up with that.

**Expected behavior**

The delete area should update based on the size of the flyout.

**Screenshots**

![delete-area-old](https://user-images.githubusercontent.com/25440652/121970516-93afec80-cd2b-11eb-93fa-992705864f83.gif)

moniika:
Drag target areas are re-calculated for calls to `WorkspaceSvg.prototype.recordDragTargets`. 
In this case, we would need to trigger re-calculating the drag targets when the flyout resizes with a call to `recordDragTargets`.

PR #4927, opened 2021-06-16
Fixed bug that prevented delete areas from updating when flyout width changed.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/4912

### Proposed Changes
Updates the delete areas when the flyout width changes.

#### Behavior Before Change
Delete area bounds were not updated, so the width of the delete area did not align with the width of the flyout.

#### Behavior After Change
Delete area and flyout bounds are kept in sync.

### Test Coverage
Manually tested and confirmed all test suites pass.

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -367,6 +367,7 @@ Blockly.HorizontalFlyout.prototype.reflowInternal_ = function() {
     // Record the height for workspace metrics and .position.
     this.height_ = flyoutHeight;
     this.position();
+    this.targetWorkspace.recordDragTargets();
   }
 };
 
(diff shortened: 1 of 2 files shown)


alschmiedt:
I think there is the same problem with horizontal flyouts so if you could also add this fix to `horizontal_flyout.js` that would be great. If you want to test this, I think the best way would be to add a `test_fields_multilinetext` block to a simple toolbox and set the `horizontalLayout` in the options struct to be `true`. lmk if you have any questions : )

gonfunko:
Good call, thanks for catching that! Updated accordingly.

alschmiedt (approved):

Issue #4847, opened 2021-05-24
Detached shadow block in workspace + Uncaught TypeError: Cannot read property 'isShadow' of null


4n70w4:
**Describe the bug**

Detached shadow block in workspace. These blocks cannot be moved, attached, or deleted.

**To Reproduce**

1. Add to toolbox:
```
'blockxml' => '<block type="lists_create_with">
          <mutation items="2"></mutation>
                    <value name="ADD0">
            <shadow type="text">
              <field name="TEXT">abc</field>
            </shadow>
          </value>
                              <value name="ADD1">
            <shadow type="text">
              <field name="TEXT">def</field>
            </shadow>
          </value>
        </block>',
```

2. Add this `lists_create_with` to workspace
3.  Click `gear` icon to change mutations
4. Delete any `item` in `'list` in mutation popup
5. Close mutation popup
6. See detached shadow block in workspace
7. Click at this detached shadow block
8. See in console:

```
Uncaught TypeError: Cannot read property 'isShadow' of null
```

**Expected behavior**

For example: these blocks can be moved, attached or removed.

**Screenshots**

![image](https://user-images.githubusercontent.com/38257723/119337574-a9d00d00-bc97-11eb-8e1b-ad1de81627b3.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 90.0.4430.212

**Stack Traces**

```
Uncaught TypeError: Cannot read property 'isShadow' of null
    at t.TouchGesture.t.Gesture.setTargetBlock_
    at t.TouchGesture.t.Gesture.setTargetBlock_
    at t.TouchGesture.t.Gesture.setStartBlock
    at t.TouchGesture.t.Gesture.handleBlockStart
    at t.BlockSvg.onMouseDown_
    at SVGGElement.l
```

**Additional context**

blockly@5.20210325.1

moniika:
Issue can also be repro'd without adding to toolbox and adding the block with the following code:
```
Blockly.Xml.domToBlock(Blockly.Xml.textToDom(`<block type="lists_create_with">
          <mutation items="2"></mutation>
                    <value name="ADD0">
            <shadow type="text">
              <field name="TEXT">abc</field>
            </shadow>
          </value>
                              <value name="ADD1">
            <shadow type="text">
              <field name="TEXT">def</field>
            </shadow>
          </value>
        </block>`), Blockly.getMainWorkspace());
```

moniika:
[...]

PR #4929, opened 2021-06-17
Add check for removing shadow

moniika:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/4847

### Proposed Changes

Adds check in disconnect to dispose of shadow blocks that are disconnected in logic for list and test blocks.

#### Behavior Before Change

Shadow blocks are not disposed of when they are disconnected (triggered by mutator).

#### Behavior After Change

Shadow blocks are not disposed of when they are disconnected (triggered by mutator).

### Reason for Changes

Fixes bug where workspace could get in invalid state with a shadow block with no parent.

### Test Coverage

1. Added a `lists_create_with` block with shadow blocks attached.
```
  <block type="lists_create_with" id="ValgeK+i,*14r-$Y/(Mf" x="137" y="288">
    <mutation items="3"></mutation>
    <value name="ADD0">
      <shadow type="text" id="$7.W0Q.$C0EN_gvhj!O1">
        <field name="TEXT"></field>
      </shadow>
    </value>
    <value name="ADD1">
      <shadow type="text_join" id="Y]Ey@~iE1)CyROK:2S;Y">
        <value name="ADD0">
          <shadow type="text" id="8d.%rw85YL({LZTf5iR;">
            <field name="TEXT"></field>
          </shadow>
        </value>
      </shadow>
    </value>
    <value name="ADD2">
      <block type="text" id="PV!ud*`c-FK6?-|3$74o">
        <field name="TEXT"></field>
      </block>
    </value>
  </block>
</xml>
```
2. Try each of the following scenarios (recreating the above block before each attempt):
  - Remove the second connection on the outer list_create_with block (should dispose of the 2 shadow blocks connected)
  - Remove the first connection on the outer list_create_with block (should shift the shadow blocks connected in the second connections up one)
  - Removed the first connection on the inner list_create_with block (should dispose of the connected shadow block)
3. Observed that the shadow block(s) was/were properly disposed of.

### DIFF
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -184,7 +184,11 @@ Blockly.Blocks['lists_create_with'] = {
     for (var i = 0; i < this.itemCount_; i++) {
       var connection = this.getInput('ADD' + i).connection.targetConnection;
       if (connection && connections.indexOf(connection) == -1) {
+        var childBlock = connection.getSourceBlock();
         connection.disconnect();
+        if (childBlock.isShadow()) {
+          childBlock.dispose(true);
+        }
       }
     }
     this.itemCount_ = connections.length;
--- a/blocks/text.js
+++ b/blocks/text.js
@@ -731,7 +731,11 @@ Blockly.Constants.Text.TEXT_JOIN_MUTATOR_MIXIN = {
     for (var i = 0; i < this.itemCount_; i++) {
       var connection = this.getInput('ADD' + i).connection.targetConnection;
       if (connection && connections.indexOf(connection) == -1) {
+        var childBlock = connection.getSourceBlock();
         connection.disconnect();
+        if (childBlock.isShadow()) {
+          childBlock.dispose(true);
+        }
       }
     }
     this.itemCount_ = connections.length;
--- a/core/connection.js
+++ b/core/connection.js
@@ -435,7 +435,7 @@ Blockly.Connection.prototype.disconnect = function() {
   this.disconnectInternal_(parentBlock, childBlock);
   if (childBlock.isShadow()) {
     // Displaced shadow blocks dissolve.
-    childBlock.dispose();
+    childBlock.dispose(true);
   } else {
     // If we were disconnecting a shadow, no need to spawn a new one.
     parentConnection.respawnShadow_();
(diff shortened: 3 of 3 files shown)


moniika:
This fixes the logic in the list and test blocks so that they properly handle shadow block.
We may, however, want to address how it is possible to get into the state where there is a shadow block without a parent (disconnect does not dispose of the shadow block).

moniika:
This fixes the logic in the list and test blocks so that they properly handle shadow block.
We may, however, want to address how it is possible to get into the state where there is a shadow block without a parent (disconnect does not dispose of the shadow block).

BeksOmega:
This fix looks good, but I also really liked your previous fix where you disposed of the shadow block on disconnect! I can't see what tests were failing, but do you think it could be related to how [connect_ also disposes of shadows](https://github.com/google/blockly/blob/13bb9f5bf6e242b5360b4f56653fbc656d081ada/core/connection.js#L124) (previously [this line](https://github.com/google/blockly/blob/fc6a1be3d982349292b630d2818b3b5509688d5f/core/connection.js#L126))? If so, do you think that your previous solution could work? Or were there other issues with it?

If you do decide to go with the dispose-on-disconnect option, could you add some unit tests for it too?

And independent of which solution you go with, could you check how it handles undo/redo? I can't remember if there should be an event group around the dispose call, or if it will automatically get grouped into the mutation/change event.

moniika:
> This fix looks good, but I also really liked your previous fix where you disposed of the shadow block on disconnect! I can't see what tests were failing, but do you think it could be related to how [connect_ also disposes of shadows](https://github.com/google/blockly/blob/13bb9f5bf6e242b5360b4f56653fbc656d081ada/core/connection.js#L124) (previously [this line](https://github.com/google/blockly/blob/fc6a1be3d982349292b630d2818b3b5509688d5f/core/connection.js#L126))? If so, do you think that your previous solution could work? Or were there other issues with it?
> 
> If you do decide to go with the dispose-on-disconnect option, could you add some unit tests for it too?
> 
> And independent of which solution you go with, could you check how it handles undo/redo? I can't remember if there should be an event group around the dispose call, or if it will automatically get grouped into the mutation/change event.

Thanks for pointing out the undo/redo. This change does have issues with undo/redo.
The previous fix that adds a dispose on disconnect also had issues, I think particularly when there were nested shadow blocks. It also was a problem for the behavior of the list/text blocks where it disconnects the input and shifts it to the previous one. 

Since this fix is not as straightforward as I had hoped and we are at release weel, I will leave this bug to be addressed in a future bug bash.

PR #4934, opened 2021-06-18
Fix escape key on all fields.

### LINKED ISSUE #4910: Escape no longer reverts field edits
Open this ancient version of Blockly:
  https://2-dot-blockly-games.appspot.com/turtle?level=10
Edit a number field, type in a big number, then press escape.
The number reverts to its original value.

Open any modern version of Blockly:
  https://blockly.games/turtle?level=10
Edit a number field, type in a big number, then press escape.
The number no longer reverts to its original value.

Same with text inputs.

NeilFraser:
Currently the escape key fails to revert the value.  This is the inherited root of all type-in fields, including number and angle.

Fixes #4910

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -456,7 +456,7 @@ Blockly.FieldTextInput.prototype.onHtmlInputKeyDown_ = function(e) {
     Blockly.WidgetDiv.hide();
     Blockly.DropDownDiv.hideWithoutAnimation();
   } else if (e.keyCode == Blockly.utils.KeyCodes.ESC) {
-    this.setValue(this.htmlInput_.defaultValue);
+    this.setValue(this.htmlInput_.untypedDefaultValue_);
     Blockly.WidgetDiv.hide();
     Blockly.DropDownDiv.hideWithoutAnimation();
   } else if (e.keyCode == Blockly.utils.KeyCodes.TAB) {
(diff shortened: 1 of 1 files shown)


core/field_textinput.js
```diff
@@ -456,7 +456,7 @@ Blockly.FieldTextInput.prototype.onHtmlInputKeyDown_ = function(e) {
     Blockly.WidgetDiv.hide();
     Blockly.DropDownDiv.hideWithoutAnimation();
   } else if (e.keyCode == Blockly.utils.KeyCodes.ESC) {
-    this.htmlInput_.value = this.htmlInput_.defaultValue;
+    this.setValue(this.htmlInput_.defaultValue);
```
BeksOmega, line 459: I think this should actually use `untypedDefaultValue` (which, admittedly, could use a better name and/or docs). `defaultValue` is the editor text, which may not be accepted by `setValue`.

Might be worth testing this fix with the note demo field if possible, because I think that one has the most complicated editor text behavior.

Otherwise, I think this is the correct way to fix the issue. LGTM!

BeksOmega (approved):

core/field_textinput.js
NeilFraser: Good call.

The note field also tests correctly (before this PR esc would close the note picker and select the current node, not revert to the previous note).

Issue #4931, opened 2021-06-17
Non autoclosing flyouts do not update zoom level correctly


alschmiedt:
**Describe the bug**
When a toolbox has a flyout where `autoclose` is set to `false` it does not update its zoom level while the workspace is zoomed.

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/tests/playground.html
2. Run `Blockly.getMainWorkspace().getToolbox().getFlyout().autoClose = false` in the console.
3. Zoom the workspace
4. Click on the flyout.
5. Notice how the flyout zoom level jumps

![flyout-bug](https://user-images.githubusercontent.com/23059043/122480218-488f1700-cf81-11eb-82ff-c4943031f28e.gif)

**Expected behavior**
The flyout should zoom in as the workspace zooms in.
![flyout-expected](https://user-images.githubusercontent.com/23059043/122480227-4d53cb00-cf81-11eb-8e5a-060f4e8344c0.gif)

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Stack Traces**

```
Replace with error stack trace.
```

**Additional context**

BeksOmega:
Closed by #4938

PR #4938, opened 2021-06-18
Reflow flyouts when zoom level changes to keep block zoom level in sync.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/4931

### Proposed Changes
Updates flyouts to listen for zoom events and re-render themselves so that their blocks reflect the new zoom level.

#### Behavior Before Change
Zoom level changes are not reflected in an open flyout until a click occurs inside the flyout itself.

#### Behavior After Change
Zoom level changes are immediately reflected in an open flyout.

### Test Coverage
Manually verified and ensured all test suites pass.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -296,9 +296,6 @@ Blockly.Flyout.prototype.init = function(targetWorkspace) {
     this.targetWorkspace.addChangeListener(this.filterWrapper_);
   }
 
-  var zoomHandler = this.handleZoom_.bind(this);
-  this.targetWorkspace.addChangeListener(zoomHandler);
-
   // Dragging the flyout up and down.
   Array.prototype.push.apply(
       this.eventWrappers_,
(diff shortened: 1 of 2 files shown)


alschmiedt (commented):
So I think this might be even simpler than this change. 

We use flyouts in two different ways on the workspace: 
- When we have a category toolbox we have flyouts that the toolbox owns. `(workspace.getToolbox().getFlyout()`
- When we have a "simple toolbox" (a flyout that sits on the workspace always open) then the workspace owns the flyout. `workspace.flyout_`

We actually do the reflow for the flyouts that live on the workspace on [this line](https://github.com/google/blockly/blob/fc6a1be3d982349292b630d2818b3b5509688d5f/core/workspace_svg.js#L2213). I think we could leverage this for flyouts that are open as well by using `workspace.getFlyout()` instead of `workspace.flyout_` and checking that the flyout `isVisible` before doing the reflow.

How we deal with the two kinds of flyouts is not very clear so let me know if you have any questions!

gonfunko:
Oh nice, good call! That appears to work fine too, and is a lot simpler. Updated the PR accordingly - thanks!

alschmiedt (approved):
Awesome, thanks!

Approved, but just a reminder to wait until after release to merge in : )

Issue #2763, opened 2019-08-01
Make a single field uneditable without making the whole block uneditable


Tereshka:
**Is your feature request related to a problem? Please describe.**

Need an opportunity to make a single field uneditable without making the whole block uneditable

**Describe the solution you'd like**
There was a topic in Blockly developers group with solution:
[https://groups.google.com/d/msg/blockly/OiUL3gzGATQ/WUadyexeBwAJ](https://groups.google.com/d/msg/blockly/OiUL3gzGATQ/WUadyexeBwAJ)

**Additional context**

It will be great to have an implementation of above solution in blockly core itself.

BeksOmega:
Closing as a dupe of #4932 as that one has more detail.

Issue #4932, opened 2021-06-18
Disabling only one field on a block


jschanker:
**Is your feature request related to a problem? Please describe.**

While you can disable all fields on a block by calling [setEditable](https://github.com/google/blockly/blob/master/core/block.js#L882) on it, there doesn't appear to be a method to disable only one of its fields.

**Describe the solution you'd like**

Introduce a `setEnabled` method for fields so you can toggle whether the editor will display or not.

**Describe alternatives you've considered**

* Have the user extend a Field subclass to provide this method manually for the specific type of field he/she wants to have the ability to disable.  Not good for overriding multiple types of Fields.

**Additional context**

Related to the question in the post on the Blockly forum here: [https://groups.google.com/g/blockly/c/tnWc8-WM1lk](https://groups.google.com/g/blockly/c/tnWc8-WM1lk)

[Beka] And this topic as well: https://groups.google.com/d/msg/blockly/OiUL3gzGATQ/WUadyexeBwAJ

jschanker:
Stumbled upon duplicate https://github.com/google/blockly/issues/2763 while looking at the Icebox issues.  Close this as duplicate?

BeksOmega:
I think this one has more detail so I'm going to keep this one and close the other =) Plus this issue is already linked to your PR. I've added the forum link from that issue to this one. Thanks for finding the dupe!

jschanker:
Resolved by #4941.

PR #4941, opened 2021-06-20
Permit single field disabling (#4932)

jschanker:
## The basics

- [x] I branched from develop

### Resolves

#2763, #4932

### Proposed Changes

Allows the developer to specify which fields on an editable block should be disabled.

#### Behavior Before Change

Editors for fields would appear when clicked exactly when their source blocks were editable.  This meant that the developer could not call a core Blockly method to disable a block's field without disabling all other fields on that block.   To enable/disable all fields on a block, he/she could call the Block's [setEditable](https://github.com/google/blockly/blob/master/core/block.js#L882) method.

#### Behavior After Change

The developer can now choose which fields to disable via the newly added `Field` method `setEnabled`.  To be clickable, the field must be enabled **and** its source block must be editable.  A field can be enabled even if its source block is non-editable, it just won't be clickable.

### Reason for Changes

Developers may want to disable some fields without disabling all of them.

### Test Coverage

Tested locally on block with multiple fields using the different combinations (editable block/enabled field, editable block/disabled field, non-editable block/enabled field, non-editable block/disabled field). 

### Documentation

Probably not beyond the property/function annotations that were added/updated, which I'm assuming are autogenerated.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -203,7 +203,8 @@ Blockly.Field.prototype.isDirty_ = true;
 Blockly.Field.prototype.visible_ = true;
 
 /**
- * Can the field value be changed using the editor on an editable block?
+ * Whether the field value can be changed using the editor on an editable
+ *     block.
  * @type {boolean}
  * @protected
  */
@@ -399,8 +400,8 @@ Blockly.Field.prototype.bindEvents_ = function() {
 };
 
 /**
- * Sets the field's value based on the given XML element. Should only be
- * called by Blockly.Xml.
+ * Sets the field's value based on the given XML element. Should only be called
+ *     by Blockly.Xml.
  * @param {!Element} fieldElement The element containing info about the
  *    field's state.
  * @package
@@ -460,7 +461,7 @@ Blockly.Field.prototype.updateEditable = function() {
 
 /**
  * Set whether this field's value can be changed using the editor when the
- * source block is editable.
+ *     source block is editable.
  * @param {boolean} enabled True if enabled.
  */
 Blockly.Field.prototype.setEnabled = function(enabled) {
@@ -470,7 +471,7 @@ Blockly.Field.prototype.setEnabled = function(enabled) {
 
 /**
  * Check whether this field's value can be changed using the editor when the
- * source block is editable.
+ *     source block is editable.
  * @return {boolean} Whether this field is enabled.
  */
 Blockly.Field.prototype.isEnabled = function() {
--- a/core/field.js
+++ b/core/field.js
@@ -399,39 +399,14 @@ Blockly.Field.prototype.bindEvents_ = function() {
 };
 
 /**
- * Disables field if the given XML element specifies this.
+ * Sets the field's value based on the given XML element. Should only be
+ * called by Blockly.Xml.
  * @param {!Element} fieldElement The element containing info about the
- *     field's state.
- * @protected
- */
-Blockly.Field.prototype.setEnabledFromXml_ = function(fieldElement) {
-  if (fieldElement.getAttribute('enabled') === 'false') {
-    this.setEnabled(false);
-  }
-};
-
-/**
- * Sets the field's value and possibly disables field based on the given XML
- *     element. Should only be called by Blockly.Xml.
- * @param {!Element} fieldElement The element containing info about the
- *     field's state.
+ *    field's state.
  * @package
  */
 Blockly.Field.prototype.fromXml = function(fieldElement) {
   this.setValue(fieldElement.textContent);
-  this.setEnabledFromXml_(fieldElement);
-};
-
-/**
- * Adds false enabled attribute to the given XML element if field is disabled.
- * @param {!Element} fieldElement The element containing info about the
- *     field's state.
- * @protected
- */
-Blockly.Field.prototype.addDisabledToXml_ = function(fieldElement) {
-  if (!this.enabled_) {
-    fieldElement.setAttribute('enabled', 'false');
-  }
 };
 
 /**
@@ -443,7 +418,6 @@ Blockly.Field.prototype.addDisabledToXml_ = function(fieldElement) {
  */
 Blockly.Field.prototype.toXml = function(fieldElement) {
   fieldElement.textContent = this.getValue();
-  this.addDisabledToXml_(fieldElement);
   return fieldElement;
 };
 
@@ -486,7 +460,7 @@ Blockly.Field.prototype.updateEditable = function() {
 
 /**
  * Set whether this field's value can be changed using the editor when the
- *     source block is editable.
+ * source block is editable.
  * @param {boolean} enabled True if enabled.
  */
 Blockly.Field.prototype.setEnabled = function(enabled) {
@@ -496,7 +470,7 @@ Blockly.Field.prototype.setEnabled = function(enabled) {
 
 /**
  * Check whether this field's value can be changed using the editor when the
- *     source block is editable.
+ * source block is editable.
  * @return {boolean} Whether this field is enabled.
  */
 Blockly.Field.prototype.isEnabled = function() {
--- a/core/field.js
+++ b/core/field.js
@@ -203,8 +203,7 @@ Blockly.Field.prototype.isDirty_ = true;
 Blockly.Field.prototype.visible_ = true;
 
 /**
- * Whether the field value can be changed using the editor on an editable
- *     block.
+ * Can the field value be changed using the editor on an editable block?
  * @type {boolean}
  * @protected
  */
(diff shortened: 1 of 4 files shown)


jschanker:
Thinking about this more, I'm assuming we'd want to include if a given field was disabled in the block serialization.  I can make additional commits later, accordingly.  However, I'll wait for comments on whether the approach I've taken is how we'd want to handle disabling individual fields.

alschmiedt:
Thanks for your PR! 

I have assigned someone to look at it but just as a heads up we have an internal event this week so we might be a little bit slower than normal reviewing it. Thanks for your patience! Also, since we are getting close to releasing this will most likely go into the next release. Thanks again!

rachel-fenichel:
Bump @maribethb now that we're done with releasing.

core/field.js
```diff
@@ -202,6 +202,13 @@ Blockly.Field.prototype.isDirty_ = true;
  */
 Blockly.Field.prototype.visible_ = true;
 
+/**
+ * Can the field value be changed using the editor on an editable block?
```
maribethb, line 206: Nit: For consistency, "Whether the field value can be changed using the editor on an editable block."

core/field.js
```diff
@@ -392,14 +399,39 @@ Blockly.Field.prototype.bindEvents_ = function() {
...
+  this.setEnabledFromXml_(fieldElement);
+};
+
+/**
+ * Adds false enabled attribute to the given XML element if field is disabled.
+ * @param {!Element} fieldElement The element containing info about the
+ *     field's state.
+ * @protected
+ */
+Blockly.Field.prototype.addDisabledToXml_ = function(fieldElement) {
```
maribethb, line 431: This name is slightly confusing because you're adding an attribute called "enabled". But you're only adding it if the field is disabled so "addEnabledToXml_" might be just as confusing. What do you think?

maribethb (approved):
This looks good to me (with some minor comments). Sorry for the long delay in reviewing. Thanks for your work on this!

BeksOmega:
Ope sorry! Somehow I missed the XML stuff before. This doesn't actually need to be serialized because it's not user-facing. Since it's only triggered by devs programmatically, we should just expect devs to re-trigger it if necessary (like how warnings work).

core/field.js
jschanker: Sounds good, I made the change.  I'm holding off on making another commit before the serialization questions are resolved.

jschanker: Actually one follow-up: Should other comments be changed as well?  I used a question here because some other properties were commented that way.

core/field.js
jschanker: Perhaps `addEnabledFalseToXml_`?  I'm not sure.  After I submitted the pull request, I was also wondering if it'd be better to use something like `addAttributesToXml_` to make it more forward-thinking for potential future additions.  Of course this assumes that the `Field` subclasses would generally add them in the same way.  I'm open to whichever option you think is best.

jschanker:
> Ope sorry! Somehow I missed the XML stuff before. This doesn't actually need to be serialized because it's not user-facing. Since it's only triggered by devs programmatically, we should just expect devs to re-trigger it if necessary (like how warnings work).

I added code to serialize because I noticed that if a developer sets a block's `editable_` property to `false`, then this would be reflected in the XML.  I thought this would be analogous?

BeksOmega:
> > Ope sorry! Somehow I missed the XML stuff before. This doesn't actually need to be serialized because it's not user-facing. Since it's only triggered by devs programmatically, we should just expect devs to re-trigger it if necessary (like how warnings work).
> 
> I added code to serialize because I noticed that if a developer sets a block's `editable_` property to `false` via `setEditable`, then this would be reflected in the XML. I thought this would be analogous?

Hmm I thought editablility was surfaced through the context menu, but you're totally right it's only available to the developer. In that case I'm of the opinion that neither of those things should be serialized. Either (a) the developer is triggering this setting based on other state, in which case all of the relevant information is already encoded in XML, or (b) the developer is surfacing this toggle to the user themselves, in which case it is the developer's responsibility to serialize it, not Blockly's.

Does that feel logical? or does it feel like I'm off the mark?

Sadly we can't get rid of the existing `editable` save information because devs could be relying on it. But I think that we don't want to add to the XML unless we have to.

jschanker:
> it's only available to the developer. In that case I'm of the opinion that neither of those things should be serialized...Does that feel logical? or does it feel like I'm off the mark?

That seems reasonable to me and is consistent with how other developer-only facing attributes such as a block's color are handled.  Serializing these attributes may be unnecessary or even undesirable in certain applications.  It can also make maintenance of the core a little more difficult. 

That being said, it's possible some developers may decide to make one or more of these attributes user-facing.  So if there's sufficient demand, perhaps a plugin could be created to detect and serialize attributes that can be changed by a user (or allows the developer to choose some subset of them for serialization).

> Sadly we can't get rid of the existing editable save information because devs could be relying on it.

For consistency, we could always add a context menu item allowing users to toggle the editability of a block.  But this is probably not a good idea since it might be confused with disabling a block and adds a likely seldomly used option.

core/field.js
jschanker: No longer applicable after serialization was reverted.

jschanker:
I reverted the (de)serialization code additions.

BeksOmega:
> That seems reasonable to me and is consistent with how other developer-only facing attributes such as a block's color are handled. Serializing these attributes may be unnecessary or even undesirable in certain applications. It can also make maintenance of the core a little more difficult.

Fwew I'm glad it sounds ok to you. Honestly I was going back and forth on it after posting my previous message. I do like consistency, and there's no really consistent solution here :/

> That being said, as you noted in (b), it's possible some developers may decide to make one or more of these attributes user-facing. So if there's sufficient demand, perhaps a plugin could be created to detect and serialize attributes that can be changed by a user (or allows the developer to choose some subset of them for serialization).

Yeah definitely! And if enough people ask for it we can always add it directly into core too.

I'm going to leave the final review of this to @maribethb =) Thanks for hashing this out with me hehe.

core/field.js
maribethb: You're right, this is consistent with how this file does it. I wasn't familiar with this style inside of Blockly before. You can just leave it like you had it. Sorry for the churn.

core/field_dropdown.js
```diff
@@ -153,10 +153,10 @@ Blockly.FieldDropdown.fromJson = function(options) {
 };
 
 /**
- * Sets the field's value based on the given XML element. Should only be
- * called by Blockly.Xml.
- * @param {!Element} fieldElement The element containing info about the
- *    field's state.
+ * Sets the field's value based on the given XML element. Should only be called
```
maribethb, line 156: I think you could probably remove this file from the PR now right?

maribethb (commented):
Thank you and Beka for working through the xml details :)

core/field_dropdown.js
jschanker: I removed the file.  It's not related to the pull request anymore, but I thought I would adjust the styling of the comment by (a) moving up "field's" to the previous line since the line would have 80 characters and (b) indenting the wrapped line with 4 spaces.

Regarding (b), I noticed that there is some inconsistency with comments' styling in general as some are indented by 4 spaces when they wrap to the next line and others are not.

jschanker:
> You're right, this is consistent with how this file does it. I wasn't familiar with this style inside of Blockly before. You can just leave it like you had it. Sorry for the churn.

I reverted to the question form, but as was the case with the indentation of comments, I'm not sure what the expected style is.  I see that Boolean properties seem to have question annotations while Boolean parameters and return values use statements?

core/field_dropdown.js
maribethb: They should ideally be indented, but there are many places they aren't currently. However, we are Soon (TM) going to be using clang-format for indentation so we will never have to think about it (or line wrapping) ever again :D So for now it's fine as long as eslint doesn't complain.

maribethb:
> > You're right, this is consistent with how this file does it. I wasn't familiar with this style inside of Blockly before. You can just leave it like you had it. Sorry for the churn.
> 
> I reverted to the question form, but as was the case with the indentation of comments, I'm not sure what the expected style is. I see that Boolean properties seem to have question annotations while Boolean parameters and return values use statements?

It does look like we have a mix of both, and the style guide seems to have no opinion, so we should just be consistent with what's nearby. `menuitem.js` splits the difference with `Is this menu item currently highlighted.` which is probably the worst of all options... so we're not consistent and I probably should have checked this file before asking you to change it, but I hadn't realized how prevalent it was.

PR #4945, opened 2021-06-22
Fix serializing non-printable characters to XML

### LINKED ISSUE #4590: Blockly does not serialize non-printable characters correctly
**Describe the bug**

Blockly does not properly escape non-printable characters when serializing field inputs to XML.

XML does not accept all unicode characters. It only accepts the characters defined in [this range](https://www.w3.org/TR/xml/#charsets), and others need to be escaped. Currently Blockly is not properly escaping them.

**To Reproduce**

1. Add a block with a text input to the workspace.
2. Enter a non-printable character into the input. Eg on ubuntu you do `ctrl + shift + u` then `0001` to get a non-printable character.
3. Refresh to reload from XML.
4. Observe how the XML was unable to be parsed.

**Expected behavior**

Any text that it is possible for the user to enter into a field should serialize and deserialize properly.
[Edit: non-printable characters can also be input by the user via comments, so those values should be escaped as well]

**Additional context**

Related to: https://github.com/google/blockly/pull/1628 and https://github.com/mit-cml/appinventor-sources/pull/2370

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #4590

### Proposed Changes

Makes it so that control chars are properly handled when serializing and deserializing XML.

#### Behavior Before Change

Control chars would not be "escaped" (replaced with their numerical character reference) when serializing. And they would cause errors to be thrown when deserializing.

#### Behavior After Change

Now they are properly escaped (for XML 1.1) and properly deserialized.

### Reason for Changes

1) Accidentally typing a control char into your project should not brick your project.
2) Sometimes programmers need to interact with control chars, so Blockly should support that.

### Test Coverage

Added tests for serialization and deserialization of control chars. Also added a test for escaping ampersands because I was in the area.

Manually tested that control chars can be serialized and deserialized when typed into a text input field using the steps from #4590.

Tested on:
 * Desktop Chrome

### Additional Information

Based on @ewpatton's changes from https://github.com/mit-cml/appinventor-sources/pull/2370. I decided to do the escaping when serializing rather than deserializing to conform with XML 1.1.

I also attempted using an xml document for serialization as proposed [here](https://github.com/google/blockly/pull/1628#discussion_r168647450), but the built-in serializer does not properly escape control characters :/

And I also attempted to escape the control chars within `Blockly.utils.xml.createTextNode`, but this caused the `&` to be escaped to `&amp;` inside `domToText`, which is not what we wanted.

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -93,20 +93,10 @@ Blockly.utils.xml.domToText = function(dom) {
  * reference).
  * @param {string} text The text to sanitize.
  * @return {string} The sanitized text.
+ * @public
  */
 Blockly.utils.xml.sanitizeText = function(text) {
-  if (!text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS)) {
-    return text;
-  }
-  var match;
-  var newText = '';
-  while ((match = text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS))) {
-    newText += text.substring(0, match.index);
-    newText += '&#x';
-    newText += text.charCodeAt(match.index).toString(16);
-    newText += ';';
-    text = text.substr(match.index + 1);
-  }
-  newText += text;
-  return newText;
+  return text.replace(Blockly.utils.xml.INVALID_CONTROL_CHARS, function(match) {
+    return '&#x' + match.charCodeAt(0).toString(16) + ';';
+  });
 };
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -29,7 +29,7 @@ Blockly.utils.xml.NAME_SPACE = 'https://developers.google.com/blockly/xml';
  * character references) before serializing to XML.
  */
 // eslint-disable-next-line no-control-regex
-Blockly.utils.xml.INVALID_CONTROL_CHARS = /[\u0000-\u0009\u000B\u000C\u000E-\u001F]/;
+Blockly.utils.xml.INVALID_CONTROL_CHARS = /[\x00-\x09\x0B\x0C\x0E-\x1F]/;
 
 /**
  * Get the document object.  This method is overridden in the Node.js build of
@@ -54,7 +54,6 @@ Blockly.utils.xml.createElement = function(tagName) {
 
 /**
  * Create text element for XML.
- * serialized to XML.
  * @param {string} text Text content.
  * @return {!Text} New DOM text node.
  * @public
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -25,6 +25,13 @@ goog.provide('Blockly.utils.xml');
 Blockly.utils.xml.NAME_SPACE = 'https://developers.google.com/blockly/xml';
 
 /**
+ * Matches control characters tha need to be replaced with NCRs (numerical
+ * character references) before serializing to XML.
+ */
+// eslint-disable-next-line no-control-regex
+Blockly.utils.xml.INVALID_CONTROL_CHARS = /[\u0000-\u0009\u000B\u000C\u000E-\u001F]/;
+
+/**
  * Get the document object.  This method is overridden in the Node.js build of
  * Blockly. See gulpfile.js, package-blockly-node task.
  * @return {!Document} The document object.
@@ -47,6 +54,7 @@ Blockly.utils.xml.createElement = function(tagName) {
 
 /**
  * Create text element for XML.
+ * serialized to XML.
  * @param {string} text Text content.
  * @return {!Text} New DOM text node.
  * @public
@@ -76,5 +84,29 @@ Blockly.utils.xml.textToDomDocument = function(text) {
  */
 Blockly.utils.xml.domToText = function(dom) {
   var oSerializer = new XMLSerializer();
-  return oSerializer.serializeToString(dom);
+  return Blockly.utils.xml.sanitizeText(oSerializer.serializeToString(dom));
+};
+
+/**
+ * Sanitizes a text string so that it can be serialized to XML. Specifically,
+ * this replaces control characters with their NCR (numerical character
+ * reference).
+ * @param {string} text The text to sanitize.
+ * @return {string} The sanitized text.
+ */
+Blockly.utils.xml.sanitizeText = function(text) {
+  if (!text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS)) {
+    return text;
+  }
+  var match;
+  var newText = '';
+  while ((match = text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS))) {
+    newText += text.substring(0, match.index);
+    newText += '&#x';
+    newText += text.charCodeAt(match.index).toString(16);
+    newText += ';';
+    text = text.substr(match.index + 1);
+  }
+  newText += text;
+  return newText;
 };
--- a/core/xml.js
+++ b/core/xml.js
@@ -366,11 +366,11 @@ Blockly.Xml.textToDom = function(text) {
   if (!doc || !doc.documentElement ||
       doc.getElementsByTagName('parsererror').length) {
     var oParser = new DOMParser();
-    // Attempt to parse as html to deserialize control chars.
+    // Attempt to parse as HTML to deserialize control chars.
     doc = oParser.parseFromString(text, 'text/html');
     if (!doc || !doc.body.firstChild ||
         doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
-      throw Error('textToDom was unable to parse: ' + text);
+      throw Error('DOMParser was unable to parse: ' + text);
     }
     return /** @type {!Element} */ (doc.body.firstChild);
   }
--- a/core/xml.js
+++ b/core/xml.js
@@ -278,7 +278,7 @@ Blockly.Xml.cloneShadow_ = function(shadow, opt_noId) {
   var node = shadow;
   var textNode;
   while (node) {
-    if (opt_noId && node.nodeName == 'shadow') {
+    if (opt_noId && node.nodeName.toLowerCase() == 'shadow') {
       // Strip off IDs from shadow blocks.  There should never be a 'block' as
       // a child of a 'shadow', so no need to check that.
       node.removeAttribute('id');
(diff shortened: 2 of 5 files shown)


core/utils/xml.js
```diff
@@ -76,5 +84,29 @@ Blockly.utils.xml.textToDomDocument = function(text) {
...
+
+/**
+ * Sanitizes a text string so that it can be serialized to XML. Specifically,
+ * this replaces control characters with their NCR (numerical character
+ * reference).
+ * @param {string} text The text to sanitize.
+ * @return {string} The sanitized text.
+ */
+Blockly.utils.xml.sanitizeText = function(text) {
+  if (!text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS)) {
```
BeksOmega, line 98: "[Do you need this? B]ecause without it, the function should still return the input text when it contains no invalid control characters."

core/utils/xml.js
```diff
@@ -76,5 +84,29 @@ Blockly.utils.xml.textToDomDocument = function(text) {
...
+  }
+  var match;
+  var newText = '';
+  while ((match = text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS))) {
+    newText += text.substring(0, match.index);
+    newText += '&#x';
+    newText += text.charCodeAt(match.index).toString(16);
+    newText += ';';
+    text = text.substr(match.index + 1);
+  }
```
BeksOmega, line 109: "Instead of [this] while loop, could you just use the String's built-in [replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) method as follows, or is this less readable?"

```
return text.replace(RegExp(Blockly.utils.xml.INVALID_CONTROL_CHARS, "g"), function(match) {
  return '&#x' + match.charCodeAt(0).toString(16) + ';'
});
```

core/utils/xml.js
```diff
@@ -76,5 +84,29 @@ Blockly.utils.xml.textToDomDocument = function(text) {
...
+    return text;
+  }
+  var match;
+  var newText = '';
+  while ((match = text.match(Blockly.utils.xml.INVALID_CONTROL_CHARS))) {
+    newText += text.substring(0, match.index);
+    newText += '&#x';
+    newText += text.charCodeAt(match.index).toString(16);
+    newText += ';';
+    text = text.substr(match.index + 1);
```
BeksOmega, line 108: "Is there a reason you used `substr` instead of `substring` here as `substr` technically doesn't seem to be part of the core ECMAScript language?"

BeksOmega (commented):
Some comments that @jschanker sent me privately, which I'm just transfering to github. Thank you again for the comments @jschanker :D

core/utils/xml.js
BeksOmega: Good point! There is one argument for it that I can think of for it. Which is that [returning early](https://medium.com/swlh/return-early-pattern-3d18a41bba8) makes it more clear this function only cares about strings w/ control characters. But it is definitely not necessary.

core/utils/xml.js
BeksOmega: That looks sooo much more readable. I'll switch it to that on Monday :D Thank you for the recommendation!

core/utils/xml.js
BeksOmega: It looks like the ECMAScript people have backed off of that? I don't see any warning on the [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) anymore. But I'm not sure.

cpcallen: `String.prototype.substr` is in Appendix B, which is to say that it is required for web compatibility.  [Specifically](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-additional-ecmascript-features-for-web-browsers):

> These features are not considered part of the core ECMAScript language. Programmers should not use or assume the existence of these features and behaviours when writing new ECMAScript code. ECMAScript implementations are discouraged from implementing these features unless the implementation is part of a web browser or is required to run the same legacy ECMAScript code that web browsers encounter.

I'm not sure why MDN would have removed their warning, though perhaps it is fair to say that on the one hand `substr` is for all practical purposes here to stay (even in non-browser engines like node.js) while on the other hand is not deprecated (because it does not officially exist at all).  🤷‍♂️

core/utils/xml.js
```diff
@@ -47,6 +54,7 @@ Blockly.utils.xml.createElement = function(tagName) {
 
 /**
  * Create text element for XML.
+ * serialized to XML.
```
NeilFraser, line 57: Not sure I understand the comment.  Seems like a sentence fragment.

core/utils/xml.js
```diff
@@ -24,6 +24,13 @@ goog.provide('Blockly.utils.xml');
  */
 Blockly.utils.xml.NAME_SPACE = 'https://developers.google.com/blockly/xml';
 
+/**
+ * Matches control characters tha need to be replaced with NCRs (numerical
+ * character references) before serializing to XML.
+ */
+// eslint-disable-next-line no-control-regex
+Blockly.utils.xml.INVALID_CONTROL_CHARS = /[\u0000-\u0009\u000B\u000C\u000E-\u001F]/;
```
NeilFraser, line 32: Suggest: [\x00-\x09\x0B\x0C\x0E-\x1F]
It's shorter, and it also reduces future temptations to add Unicode ranges above FF which may break sanitizeText which can't handle multi-byte Unicode characters.

core/xml.js
```diff
@@ -365,7 +365,14 @@ Blockly.Xml.textToDom = function(text) {
   var doc = Blockly.utils.xml.textToDomDocument(text);
   if (!doc || !doc.documentElement ||
       doc.getElementsByTagName('parsererror').length) {
-    throw Error('textToDom was unable to parse: ' + text);
+    var oParser = new DOMParser();
+    // Attempt to parse as html to deserialize control chars.
```
NeilFraser, line 369: Capitalize HTML please.

core/xml.js
```diff
@@ -365,7 +365,14 @@ Blockly.Xml.textToDom = function(text) {
...
+    doc = oParser.parseFromString(text, 'text/html');
+    if (!doc || !doc.body.firstChild ||
+        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
+      throw Error('textToDom was unable to parse: ' + text);
```
NeilFraser, line 373: I think this message is out of date.  `textToDom` should probably be `DOMParser`.

NeilFraser (approved):
I'm shocked that XMLSerializer will create XML that DOMParser can't parse.  After a bunch of research, I can confirm that this PR appears to be the best approach.

core/xml.js
```diff
@@ -365,7 +365,14 @@ Blockly.Xml.textToDom = function(text) {
   var doc = Blockly.utils.xml.textToDomDocument(text);
   if (!doc || !doc.documentElement ||
       doc.getElementsByTagName('parsererror').length) {
-    throw Error('textToDom was unable to parse: ' + text);
+    var oParser = new DOMParser();
+    // Attempt to parse as html to deserialize control chars.
+    doc = oParser.parseFromString(text, 'text/html');
+    if (!doc || !doc.body.firstChild ||
+        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
```
NeilFraser, line 372: Note that Blockly.Xml.textToDom has a different way of detecting the same thing.  Can you choose one method and make them consistent?  Maybe even factor it out as a util function if that makes sense.  Your call, just pointing it out.

NeilFraser: Follow up on the failure detection, does this even work?  When I execute the following valid XML:
(new DOMParser).parseFromString('<xml xmlns=\"https://developers.google.com/blockly/xml\">ab</xml>', 'text/xml').body
The result is always null.  There doesn't seem to be a .body property on an XML.

Furthermore, when I execute the following invalid XML ad skip the body:
(new DOMParser).parseFromString('<xml xmlns=\"https://developers.google.com/blockly/xml\">a\u0001b</xml>', 'text/xml').firstChild
I get an 'xml' node, that contains a 'parsererror' node.

So it looks like this method of detecting an error is broken in two different ways.  Assuming that's correct, the follow-up question would be why do the tests pass?

Also, I think that's why the equivalent test in Blockly.Xml.textToDom checks for the existence of a 'parsererror' node.

core/utils/xml.js
NeilFraser: `Blockly.utils.xml.INVALID_CONTROL_CHARS.test(text)` should be much faster than `match` since it only returns true or false, and doesn't have to build up a reporting object.

core/xml.js
BeksOmega: It looks like github cut off the relevant top part of this function haha. The whole thing looks like this:
```javascript
Blockly.Xml.textToDom = function(text) {
  var doc = Blockly.utils.xml.textToDomDocument(text);
  if (!doc || !doc.documentElement ||
      doc.getElementsByTagName('parsererror').length) {
    var oParser = new DOMParser();
    // Attempt to parse as HTML to deserialize control chars.
    doc = oParser.parseFromString(text, 'text/html');
    if (!doc || !doc.body.firstChild ||
        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
      throw Error('DOMParser was unable to parse: ' + text);
    }
    return /** @type {!Element} */ (doc.body.firstChild);
  }
  return doc.documentElement;
};
```

So the first bit that uses `'text/xml'` parsing does check for a `parsererror`. But that doesn't work for HTML parsing because it doesn't (seem to?) give you parser errors. For example if I run `oParser.parseFromString('<xml</xml>', 'text/html');` It gives me:
```xml
<html>
  <head></head>
  <body>
    <xml< xml=""></xml<>
  </body>
</html>
```
So we have to run that other weird `nodeName` check instead.

Also note that I wanted to add this logic to `Blockly.utils.xml.textToDomDocument`, but the HTML parsing results in an `Element` and not a `Document`, which would be a breaking change on a public API.

BeksOmega:
> I'm shocked that XMLSerializer will create XML that DOMParser can't parse.

I know right!? I don't get it lol.

BeksOmega:
Closing this as the new JSON serializer doesn't have this problem.

PR #4952, opened 2021-06-28
Group field edits together for undo/redo

### LINKED ISSUE #4189: Fields don't group edits
**To Reproduce**

1. Choose a new angle with the mouse.
2. Undo several times

**Actual behavior**

The angle steps back in 15 degree increments one by one.

**Expected behavior**

The angle should return to 0 (or whatever the initial value was) in a single undo step.

NeilFraser:
Fixes #4189.
Also use 0 instead of 0px, in accordance with the rest of Blockly and Google’s CSS style guide.

### DIFF
--- a/core/block_drag_surface.js
+++ b/core/block_drag_surface.js
@@ -159,7 +159,7 @@ Blockly.BlockDragSurfaceSvg.prototype.translateSurfaceInternal_ = function() {
   this.SVG_.style.display = 'block';
 
   Blockly.utils.dom.setCssTransform(
-      this.SVG_, 'translate3d(' + x + 'px, ' + y + 'px, 0px)');
+      this.SVG_, 'translate3d(' + x + 'px, ' + y + 'px, 0)');
 };
 
 /**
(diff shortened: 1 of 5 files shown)


BeksOmega (approved):
LGTM, and procedure renaming works correctly again!

One non-code question: do you think that the [custom field docs](https://developers.google.com/blockly/guides/create-custom-blocks/fields/customizing-fields/creating) should be updated to include info about how to handle event grouping properly? It looks like outside developers need to handle this themselves because the group needs to be killed when their arbitrary widget closes.

NeilFraser:
If developers of custom fields (that aren't descendants of textinput) don't do anything, then they will simply get a lot of ungrouped events, so it is safe.  They can reduce the spam by adding `Blockly.Events.setGroup(true);` in their equivalent of `widgetCreate_` and a matching `Blockly.Events.setGroup(false);` in their equivalent of `widgetDispose_`.  But it is optional (core Blockly was missing this for about a year).

rachel-fenichel:
Since it's release week, we should have held off on merging. I think we should revert it and hold for a week.

NeilFraser:
Reverted.

PR #4958, opened 2021-06-28
Add command to deploy beta version of demos

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

Adds command to deploy beta version of appengine without requiring local edits to package.json.

#### Behavior Before Change

Deploying a test version of appengine demos required either editing gulpscript or package.json to avoid overwriting currently deployed version of demos.

#### Behavior After Change

Command `npm run deployDemosBeta` can be used to quickly deploy demos and reduce error in unintentionally overwriting the non-beta appengine instances.

### Reason for Changes

Improve developer scripts.

### Test Coverage

Ran command to deploy new version of demos.


scripts/gulpfiles/appengine_tasks.js
```diff
@@ -84,6 +78,50 @@ function deployAndClean(done) {
...
+  const demoVersion = getDemosVersion();
+  deployToAndClean(demoVersion);
+  done();
+}
+
+/**
+ * Constructs a beta demo version name based on the current date.
+ */
+function getDemosBetaVersion() {
+  var date = new Date();
```
maribethb, line 111: My only concern with this is that it makes it impossible to do multiple deployments on the same day (unless having multiple versions is not a problem? I'm not sure how it works)

So having a command line argument to specify the version number, or a prefix, or something, so that we can specify a unique one if needed, would be helpful. I'm not sure how often we do this or if we plan to stick with the App Engine approach anyway so feel free to ignore this or wait until we decide

maribethb (approved):

scripts/gulpfiles/appengine_tasks.js
moniika: You can do multiple deployments on the same day, it will overwrite the existing version that is deployed.
Given that we use beta for testing, I think overwriting it is the desired behavior.
I am a fan of this automatically generating the name, because it makes it much more convenient than typing in the version name, but open to discussion. I wont merge this in until after release.

cpcallen (commented):
I wonder if it might be cleaner just to provide a command line flag (say, `--beta`) that would control which algorithm is used to generate the version number, rather than having two separate gulp tasks?  You could still have two separate npm scripts, if you wanted to:

```
    "deployDemos": "gulp deployDemos",
    "deployDemos:beta": "gulp deployDemos --beta",
```
Or you could provide the flag manually:
```
$ npm run deployDemos -- --beta
```
(The extra `--` required so that `npm` knows the following arguments are for `gulp`, not for it.)

There's already an example of how to do this using [yargs](https://yargs.js.org/) in [`build_tasks.js`](https://github.com/google/blockly/blob/master/scripts/gulpfiles/build_tasks.js).  See also [gulp-cli](https://github.com/gulpjs/gulp-cli#custom-metadata) for info about documenting task command line arguments (which `build_tasks.js` doesn't do, but probably should.)

scripts/gulpfiles/appengine_tasks.js
maribethb: Cool, overwriting makes sense then.

I was thinking it would create by default, unless an option was passed in. But it sounds like we don't actually need it.

scripts/gulpfiles/appengine_tasks.js
```diff
@@ -98,6 +136,15 @@ const deployDemos = gulp.series(
     deployAndClean
 );
 
+const deployDemosBeta = gulp.series(
+    prepareDeployDir,
+    copyStaticSrc,
+    copyAppengineSrc,
+    copyPlaygroundDeps,
```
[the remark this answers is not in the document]
moniika: I'm going to file a bug to use a command line flag instead of having two separate gulp tasks.
I think it sounds like a good idea, but there isn't time to change it right now.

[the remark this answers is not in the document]
moniika: https://github.com/google/blockly/issues/5155

PR #4967, opened 2021-06-30
Enable static method inheritance.

NeilFraser:
In ES5 static methods on child classes don't inherit static methods on parent classes.  In ES6 they do.

For example, a new field in core (written in ES5) which doesn't define `.fromJson` will not have this method.  Whereas a new field in a plugin (written in ES6) which doesn't define `.fromJson` will inherit from `Blockly.FieldTextInput.fromJson` or `Blockly.FieldNumber.fromJson`

This change modifies our ES5 inheritance to be consistent with the ES6 pattern.

### DIFF
--- a/core/utils/object.js
+++ b/core/utils/object.js
@@ -19,15 +19,25 @@ goog.provide('Blockly.utils.object');
 
 /**
  * Inherit the prototype methods from one constructor into another.
- *
  * @param {!Function} childCtor Child class.
  * @param {!Function} parentCtor Parent class.
  * @suppress {strictMissingProperties} superClass_ is not defined on Function.
  */
 Blockly.utils.object.inherits = function(childCtor, parentCtor) {
+  // Set a .superClass_ property so that methods can call parent methods
+  // without hard-coding the parent class name.
+  // Could be replaced by ES6's super().
   childCtor.superClass_ = parentCtor.prototype;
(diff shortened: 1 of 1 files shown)


core/utils/object.js
```diff
@@ -19,15 +19,25 @@ goog.provide('Blockly.utils.object');
...
  * @suppress {strictMissingProperties} superClass_ is not defined on Function.
  */
 Blockly.utils.object.inherits = function(childCtor, parentCtor) {
+  // Set a .superClass_ property so that methods can call parent methods
+  // without hard-coding the parent class name.
+  // Could be replaced by ES6's super().
   childCtor.superClass_ = parentCtor.prototype;
+
+  // Link the child class to the parent class so that static methods inherit.
+  Object.setPrototypeOf(childCtor, parentCtor);
```
cpcallen, line 33: Observation: `.setPrototypeOf` was not formally introduced until ES6, but AIUI you have confirmed that it is available on all our target platforms.

cpcallen (approved):

core/utils/object.js
NeilFraser: Yes, it was introduced in IE11.

rachel-fenichel:
This is ready to merge now that the release is out.

PR #4968, opened 2021-06-30
Build in a temporary directory

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

1. Modify the `build`, `typings` and `test` npm scripts so that they do not overwrite any checked-in file(s).  Specifically, separate `build` into three separate scripts:

   * `npm run generate:langfiles`, to be invoked manually after making changes to `msg/messages.js`, will run `js_to_json.py` to update `msg/json/*.json`.
   * `npm run build` will do the rest of the build process but write the output to `BUILD_DIR` (configured to be `blockly/built/`, but could be anywhere).
   * `npm run checkin:built` will copy the built files from `BUILD_DIR` to the repository root so they can be committed.

   The `typings` npm script is similarly revised to write to `BUILD_DIR`, and a `checkin:typings` script provided to  copy the results back to `typings/`.  A `checkin` script will run `checkin:built` and `checkin:typings`.

0. Modify the scripts run by `npm test` so that, where they depend on built files, they use the copies in `BUILD_DIR` instead of the the checked-in ones in the repository root.

0. Modify the `recompile` script to run `checkin:built` and `checkin:typings`.

0. Provide `clean`, `clean:build` and `clean:release` scripts to wipe the `build/` and `dist/` temporary directories, and arrange for these to be executed at appropriate points by the 

#### Behavior Before Change

`npm run build` and `npm test` would modify committed files.

#### Behavior After Change

`npm run build` and `npm test` should not modify committed files, but `npm checkin` can be used to request such modifications.

### Reason for Changes

1. Developer experience.  Having `npm run build` or `npm test` modify tracked files is surprising and makes it difficult to understand what changes are intentional and which are fallout from previous commits.
0. Moves us toward being able to remove checked-in build products like `blockly_compressed.js` and `msg/js/*.js` from the repository in future if we want to.

### Test Coverage

This has been manually tested locally, including by removing all the checked-in build products (that I know about!) via 
```
rm -rf *_compressed.js* msg/js/*.js typings/blockly.d.ts typings/msg/*.d.ts
npm test
```
but I'm not in a position to fully verify the changes to `scripts/gulpfiles/release_tasks.js`.

### Documentation

TBD.  These changes ought to be documented somewhere.  Is there documentation for our existing npm scripts somewhere?

#### Release process

There is a question about how packaging and releasing ought to work.  I initially modified the `package` script to pull from `BUILD_DIR` instead of the repository root, but after looking at the `recompile` and `publish` tasks I changed this back so as to not change how the release process works.  I'm not sure which is preferable:

* The advantage of the present arrangement, where we package and release checked-in build products, is that it very easy to see (and test) exactly what will be packaged/released in advance of doing the release.
* The disadvantage is that we run the risk of packaging and releasing out-of-date build products.  This is especially a risk as none of the automated tests actually test the checked-in build products (with one minor exception: prior to this PR, `tests/scripts/compile_typings.sh` did actually check the generated `.d.ts` files in `typings/` would compile).

Overall I think it would be better to move to a process that does build, test, package, and release all in a single step, but I do not have much experience in this area to refer to and would appreciate input.

#### Appengine

I've not looked in detail at `scripts/gulpfiles/appengine_tasks.js`; AFAICT it will still use checked-in build products.  I would appreciate input from anyone more familiar with the Appengine deploy procedure about how this should be handled.


scripts/gulpfiles/appengine_tasks.js
```diff
@@ -47,8 +47,8 @@ function copyStaticSrc(done) {
  */
 function copyAppengineSrc() {
   const appengineSrc = [
-      path.join(demoStaticTmpDir, 'appengine/**/*'),
-      path.join(demoStaticTmpDir, 'appengine/.gcloudignore'),
+      `${demoStaticTmpDir}/appengine/**/*`,
+      `${demoStaticTmpDir}/appengine/.gcloudignore`,
```
moniika, line 51: Based on my reading https://gulpjs.com/docs/en/api/vinyl/#normalization-and-concatenation,
is that why the call to `path.join` here isn't necessary?

scripts/gulpfiles/config.js
```diff
@@ -0,0 +1,22 @@
...
+ * @fileoverview Common configuration for Gulp scripts.
+ */
+
+var path = require('path');
+
+// Paths are all relative to the repository root.  Do not include
+// trailing slash.
+module.exports = {
+  // Directory to write compiled output to.
+  BUILD_DIR: 'built',
```
moniika, line 17: Why not use "build" here? I thought using build was a common pattern, but it could just be that we happen to be using it in Blockly Samples.

scripts/gulpfiles/build_tasks.js
```diff
@@ -409,30 +411,58 @@ goog.require('Blockly.requires');
...
+function generateLangfiles(done) {
   // Run js_to_json.py
-  const jsToJsonCmd = `python ./scripts/i18n/js_to_json.py \
---input_file ${path.join('msg', 'messages.js')} \
---output_dir ${path.join('msg', 'json')} \
---quiet`;
+  const jsToJsonCmd = `python scripts/i18n/js_to_json.py \
+      --input_file ${path.join('msg', 'messages.js')} \
+      --output_dir ${path.join('msg', 'json')} \
+      --quiet`;
```
moniika, line 422: Minor note: I don't think these lines needed to be indented, if it was for the purpose of following the style guide.
Template literals specifically don't have to follow the indentation rules.
https://google.github.io/styleguide/jsguide.html#features-strings-template-strings

I don't think it hurts to change it, it'll just add extra spaces between the command line flags.

moniika (commented):
Re: Test Coverage Section:
- I have an alias for untracking build produces using `git update-index`. I also have listed blockly_uncompressed.js, msg/js/* and msg/json/*. I'm not sure if those should also be considered.

Re: Is there documentation on npm release scripts?
- We don't have documentation on our release scripts in core, aside from the jsdoc in the source files. Blockly Samples does have documentation (https://github.com/google/blockly-samples/blob/master/scripts.md) and I think it would be worth it to add something similar to core.

Appengine:
- Feel free to ping me about question about the `appengine_tasks.js` script. I have the most familiarity with that script on the team.
- The appengine script copies the checked in files and also uses local files.
The local files used are: version from package.json and playground dependencies in ./node_modules. (some discussion on that here: https://github.com/google/blockly/pull/4591#discussion_r558751306)

Misc notes:
- I like the addition of a config file, rather than having constants at the top of the scripts.
- This may change in the upcoming quarter, but we have needed to update compressed during the quarter, so we might want to document the case when build artifacts should be updated more officially and that `npm checkin` should be used (assuming that is the solution)

scripts/gulpfiles/build_tasks.js
cpcallen: Agreed.  I indented it here because that style rule continues with "though it may if the added whitespace does not matter", and we're not in any danger of exceeding the maximum command line length (8kiB on Windows and >200kB on macOS / Linux).

scripts/gulpfiles/appengine_tasks.js
cpcallen: More specifically because of [the documentation for using globs with `src`](https://gulpjs.com/docs/en/getting-started/explaining-globs/), though I suspect the two are connected.

scripts/gulpfiles/config.js
cpcallen: I don't have a strong opinion about this.  I picked "built" over "build" (after also considering "bin", "out", "compiled", etc.) because "build" sounded like it might contain build _tools_, whereas "built" seemed like output.  I wasn't able to find any info about what conventions are typically used here—but if we're using "build" in other packages for ephemeral compiler output then that's great and I'll go with that.

I was going to write output directly to `dist/`, but the files in `built/msg/` differ from the ones in `builld/msg/js/` (and `msg/js/`) for reasons I am not yet  clear on, so I've held off on that for now.

cpcallen:
> * I have an alias for untracking build produces using `git update-index`. I also have listed blockly_uncompressed.js, msg/js/* and msg/json/*. I'm not sure if those should also be considered.

The files in `msg/json/` will now _only_ be rebuilt upon manual request via `npm run generate:langfiles` (see a03cd29).

The files in `msg/js/` are now built to `built/msg/js/` by `npm run build:langfiles` (or `npm run build` or any of the other scripts that invoke that one) and copied back to `msg/js/` upon request by `npm run checkin` (and by some of the release scripts which, run that automatically).

I was wondering about whether the tests depended on `msg/js/` and whether they should run against `built/msg/js` instead, but doing an `rm -rf msg/js` before running the tests didn't break anything (!) so I don't think there is an issue here (other than a complete absence of test coverage of these `.js` files).

`blockly_uncompressed.js` is still being build by `build:uncompressed` which is run by `build`, but this this will be dealt with in a separate PR (the file will soon no longer be generated at all).

> * We don't have documentation on our release scripts in core, aside from the jsdoc in the source files. Blockly Samples does have documentation (https://github.com/google/blockly-samples/blob/master/scripts.md) and I think it would be worth it to add something similar to core.

I can add some.  Where should it such an `.md` file live?

>   The local files used are: version from package.json and playground dependencies in ./node_modules. (some discussion on that here: [Update appengine_tasks to include playground dependencies #4591 (comment)](https://github.com/google/blockly/pull/4591#discussion_r558751306))

That is… funky.  What are we doing that is not compatible with App Engine's automatic installation of depedencies (in the standard node environment, at least)?

> * I like the addition of a config file, rather than having constants at the top of the scripts.

One minor issue: that there are a few shell scripts that need the same configuration constant, and I've not thought of a good way to have them read from `config.js`.

> * This may change in the upcoming quarter, but we have needed to update compressed during the quarter, so we might want to document the case when build artifacts should be updated more officially and that `npm checkin` should be used (assuming that is the solution)

Yes.  I am hoping someone can help me update the release documentation appropriately.

moniika:
> 
> > * We don't have documentation on our release scripts in core, aside from the jsdoc in the source files. Blockly Samples does have documentation (https://github.com/google/blockly-samples/blob/master/scripts.md) and I think it would be worth it to add something similar to core.
> 
> I can add some. Where should it such an `.md` file live?
I've filed a bug to track adding the `scripts.md`. My preference would be at the same level as the package.json.

> 
> > The local files used are: version from package.json and playground dependencies in ./node_modules. (some discussion on that here: [Update appengine_tasks to include playground dependencies #4591 (comment)](https://github.com/google/blockly/pull/4591#discussion_r558751306))
> 
> That is… funky. What are we doing that is not compatible with App Engine's automatic installation of depedencies (in the standard node environment, at least)?
> 

The project deployed on appengine is not a node.js service (there is no package.json), so the dependencies aren't automatically installed.

The structure on appengine is that the core repository is copied over into a `static/` directory and everything that was under [`appengine/`](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/appengine) is moved up to the top level of the directory structure.

If you want to take a better look at what the generated directory and included files for appengine look like, you can comment out the the [deployAndClean](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/scripts/gulpfiles/appengine_tasks.js#L98) step of the deployDemos script and look at the `deploy` directory that is generated.
**Keeping in mind there is also a [`.gitignore`](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/appengine/.gcloudignore) used to skip some files and directories.**

scripts/gulpfiles/build_tasks.js
```diff
@@ -409,30 +411,58 @@ goog.require('Blockly.requires');
...
+  done();
+};
+
+/**
+ * This task builds Blockly's lang files.
+ *     msg/*.js
+ */
+function buildLangfiles(done) {
+  // Create output directory.
+  // BUG(cpcallen): this probably doesn't work on Windows.
```
rachel-fenichel, line 449: When checking in, use TODO(#issuenumber)

scripts/gulpfiles/build_tasks.js
```diff
@@ -409,30 +411,58 @@ goog.require('Blockly.requires');
...
+Regenerated several flies in msg/json/.  Now run
+
+    git diff msg/json/*.json
+
+and check that operation has not overwritten any modifications made to
+hints, etc. by the TranslateWiki volunteers.  If it has, backport
+their changes to msg/messages.js and re-run 'npm run generate:langfiles'.
+
+Once you are satisfied that any new hints have been backported you may
+go ahead and commit the changes, but note that the generate script
```
rachel-fenichel, line 435: Possible future work: no longer remove the translator credits, so that we don't have to have this admonishment.

rachel-fenichel (approved):
I have a few small comments, but it looks like Monica has been very thorough. 

> There is a question about how packaging and releasing ought to work. I initially modified the package script to pull from BUILD_DIR instead of the repository root, but after looking at the recompile and publish tasks I changed this back so as to not change how the release process works. I'm not sure which is preferable:
> 
> The advantage of the present arrangement, where we package and release checked-in build products, is that it very easy to see (and test) exactly what will be packaged/released in advance of doing the release.
> The disadvantage is that we run the risk of packaging and releasing out-of-date build products. This is especially a risk as none of the automated tests actually test the checked-in build products (with one minor exception: prior to this PR, tests/scripts/compile_typings.sh did actually check the generated .d.ts files in typings/ would compile).
> Overall I think it would be better to move to a process that does build, test, package, and release all in a single step, but I do not have much experience in this area to refer to and would appreciate input.

I agree that build->test->package->release is ideal, and also that it doesn't have to be fixed in this PR. I propose that we update the documentation for our day of releasing tasks to include the command needed to rebuild and check in the generated files, which at least reduces the risk of packaging the wrong files.

scripts/gulpfiles/config.js
cpcallen: I've renamed this to "build", since it seems that is more commonly used.

scripts/gulpfiles/build_tasks.js
cpcallen: Done.

scripts/gulpfiles/build_tasks.js
cpcallen: Fixing js_to_json to backport translation hints is probably not a good use of our time, but I've filed #5001 to request that at least it not delete/overwrite certain bits of metadata when (in the usual case) there are no hints to backport and no changes to `messages.js`.

cpcallen:
Because the node tests depend on the contents of `RELEASE_DIR` (`dist/`), the only way to actually make sure that they are testing the current original sources is to have the `package` gulp task take input from `BUILD_DIR` (renamed `build`, was `built` in an earlier version of this branch).  See ec14cc3 (and the commit comment thereto) for the gory details, but the most notable changes are:

* You have to do `npm run build` and `npm run typings` before you can run `npm package`, and there is now a check to ensure you have.  `package` never uses (or at least should never use) any checked-in copies of files that are generated by `build` or `typings`.
* `npm run recompile` does `build`, `typings` and `checkin` (amongst other things).
* `npm run publish` (and `publishBeta`) now runs `build` and `typings` before running `package`.

This all means it's not strictly necessary to run `recompile` before `publish` (but it's probably a good idea to, since it gives the team a chance to review what is about to be published).  (For good measure it also runs `checkin`, so at least there will be evidence after the fact if there were changes in the source not reflected in the checked-in build output.)

@maribethb: Can you have a look at the changes to in `release_tasks.js` and verify that no changes to the documented release process are required?

@samelhusseini: I'd appreciate you have a close look at this whole PR but especially the changes to `package_tasks.js` and `release_tasks.js`.

I believe that this it otherwise ready to merge.

cpcallen:
Realised I'd not finished the `built` -> `build` rename, and (worse) that some tests had been passing even though they should have been failing as result of this oversight, so I've pushed a commit (9e72378) to fix this as well as another (9e72378) merging in the `develop` branch again so as to resolve a minor merge conflict in `package.json`.

@rachel-fenichel and @moniika: would you like to review the additional changes since your previous reviews? 
 @samelhusseini has had a quick look and has said he will do a full review but hasn't had time yet.  It would be nice to land this today if at all possible.

scripts/gulpfiles/package_tasks.js
```diff
@@ -369,9 +397,18 @@ function packageReadme() {
...
 function packageDTS() {
-  return gulp.src(['typings/*.d.ts', 'typings/msg/*.d.ts'],
-                  {base: './typings'})
-    .pipe(gulp.dest(`${RELEASE_DIR}`));
+  const handwrittenSrcs = [
+    'typings/*.d.ts',
+    '!typings/blockly.d.ts',  // Exclude checked-in copy of blockly.d.ts.
+    'typings/msg/msg.d.ts',
+  ];
+  const builtSrcs = [
```
rachel-fenichel, line 405: Does this mean that there are actually two versions of typings being included?

scripts/gulpfiles/release_tasks.js
```diff
@@ -160,30 +170,25 @@ const publish = gulp.series(
...
-  buildTasks.checkinBuilt,
-  typings.typings,
-  typings.checkinTypings,
+  recompile,
   gitTasks.pushRebuildBranch
   );
 
 module.exports = {
-  recompile: recompile,
+  recompile: recompileDevelop,
```
rachel-fenichel, line 191: Nit: now we have an exported command named `recompile` and an internal function named `recompile` and they don't match. Please change the internal one to `recompileInternal` to avoid confusion down the line.

scripts/gulpfiles/package_tasks.js
cpcallen: There are two sets of typings: handwritten ones, and generated ones.  The latter includes (only) `blockly.d.ts` and the various `msg/*.d.ts` files; all the rest are handwritten.

Previously they were all pulled from the same place—`typings/`—but now the generated ones are pulled from `build/` instead.  But there's still checked-in copies of the generated ones in `typings/`, and we want to exclude those (because they might be out-of-date if `npm run checkin:typings` hasn't been run recently) in favour of the generated ones.

So there are two versions, but only one is being included in the npm.

rachel-fenichel (approved):

scripts/gulpfiles/release_tasks.js
cpcallen: Good point.  I opted for `rebuildAll` instead of `recompileInternal`, because it is a little more clear about the scope of what is done.

samelhusseini (commented):
I realize this is late, but LGTM.

PR #4972, opened 2021-06-30
Allow inheriting of fromJson in fields

NeilFraser:


cpcallen (approved):
A little lint to deal with but looks OK otherwise.

core/field.js
```diff
@@ -154,6 +154,14 @@ Blockly.Field = function(value, opt_validator, opt_config) {
   opt_validator && this.setValidator(opt_validator);
 };
 
+/**
+ * Construct a Field from a JSON arg object.
+ * @param {!Object} _options A JSON object with options.
+ */
+Blockly.Field.fromJson = function(_options) {
+  throw Error("Field subclass doesn't define fromJson");
```
BeksOmega, line 162: Is there anyway we can tell people which field subclass is causing the problem?

Also style nit: single quotes and period in error msg. (I'm surprised lint doesn't catch the quotes).

BeksOmega, line 162: Is there anyway we can tell people which field subclass is causing the problem?

Also style nit: single quotes and period in error msg. (I'm surprised lint doesn't catch the quotes).

core/field_angle.js
```diff
@@ -135,7 +135,9 @@ Blockly.FieldAngle.prototype.DEFAULT_VALUE = 0;
  * @nocollapse
  */
 Blockly.FieldAngle.fromJson = function(options) {
-  return new Blockly.FieldAngle(options['angle'], undefined, options);
+  // `this` might be a subclass of FieldAngle if that class doesn't override
+  // the static fromJson method.
+  return new this(options['angle'], undefined, options);
```
BeksOmega, line 140: I'm a little concerned about this. There's nothing guaranteeing that a subclass constructor has the same method signature right? Eg say someone wants to define a FieldAngle subclass that has a default angle of 90 degrees:
```
// No value parameter to constructor
Blockly.FieldAngleNinety = function(opt_validator, opt_config) { ... }
```
And they also forget to define a fromJson method. It might be tricky for them to debug what's going on. But I'm not sure of a better way to handle this either... thoughts?

core/field_dropdown.js
```diff
@@ -149,6 +149,8 @@ Blockly.FieldDropdown.ImageProperties;
  * @nocollapse
  */
 Blockly.FieldDropdown.fromJson = function(options) {
+  // `this` might be a subclass of FieldDropdown if that class doesn't override
+  // the static fromJson method.
   return new Blockly.FieldDropdown(options['options'], undefined, options);
```
BeksOmega, line 154: Should this one get switched over to?

core/field_multilineinput.js
```diff
@@ -89,7 +89,9 @@ Blockly.FieldMultilineInput.prototype.configure_ = function(config) {
  */
 Blockly.FieldMultilineInput.fromJson = function(options) {
   var text = Blockly.utils.replaceMessageReferences(options['text']);
-  return new Blockly.FieldMultilineInput(text, undefined, options);
+  // `this` might be a subclass of FieldLabel if that class doesn't override
```
BeksOmega, line 92: FieldLabel -> FieldMultilineInput

core/field_dropdown.js
NeilFraser: Oops.  Thanks!

core/field_multilineinput.js
NeilFraser: Oops.  Thanks!

core/field.js
NeilFraser: 'this' is the field subclass, but there's no name I can find that's available to print.

I'm leaning towards deleting this function completely.  It will still throw, and the error would be more informative.  It's more of an architectural issue in that 'fromJson' is expected on all Fields, so Field should really define something.  This is a bit of an academic debate between a good design and a good implementation.  What do you think?

Regarding the string, it contains an apostrophe, thus the double-quotes.

BeksOmega: > I'm leaning towards deleting this function completely. It will still throw, and the error would be more informative. It's more of an architectural issue in that 'fromJson' is expected on all Fields, so Field should really define something. This is a bit of an academic debate between a good design and a good implementation. What do you think?

If you can get the type system to like it I'd say go ahead and delete it :D Otherwise this totally works, I just wanted to bring it up using a different message in case it was a possibility.

> Regarding the string, it contains an apostrophe, thus the double-quotes.

Oh whoops I totally missed that haha. I think the new styleguide recommends a template string in that case.

core/field_angle.js
NeilFraser: At the moment if FieldAngleNinety forgets to define a fromJson, in ES6 a FieldAngle will be unexpectedly created instead.  This is hilariously challenging to debug (took me nearly two days to figure out what was going on).  So this PR will at least create the right object.

I'm less concerned about a change in signature.  After all, what we are describing is someone creating a new field, inventing a new JSON signature out of thin air, but forgetting to actually code that signature anywhere.  It's a bit like calling Blockly.inject with some new options, but not actually adding those options to the codebase.

core/field.js
NeilFraser: Chris and I just had a 30 minute ultra-geek conversation about the options here.  The best code is no code.  Deleted.  :)

core/field_angle.js
BeksOmega: Good point. LGTM then :D

core/field.js
cpcallen: > style nit: ... period in error msg

Oh, curious.  Is that a Blockly-specific convention?  I can't find mention of this in the Google styleguide.  I've usually avoided omitted them as the error message will normally be followed by "at filename:line:column…" when displayed.

BeksOmega: Oh nope that's just me being confused. I misread the diff and thought this PR added a period to the other error (actually removed it). I was going for consistency and instead caused confusion haha.

rachel-fenichel:
I think this one is ready to merge, now that the release is done.

Issue #4988, opened 2021-07-02
Github Action to deploy appengine site


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**

Deploying to AppEngine during quarterly releases is done locally and requires installing gcloud.

**Describe the solution you'd like**

Deploy through a Github Action, using [deploy-appengine](https://github.com/google-github-actions/deploy-appengine).

Use a flag to indicate whether to deploy to a staging site or the main site.

**Describe alternatives you've considered**

Continue to do this manually.

gonfunko:
Closing since we're moving away from App Engine.

PR #4998, opened 2021-07-08
Initial commit for appengine deploy action

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/4988

### Proposed Changes

- Adds a gulp action that prepares the demos but doesn't deploy or delete the directory
- Adds a github workflow with steps to prepare and deploy demos

#### Behavior Before Change

Deploying had to be done manually.

#### Behavior After Change

Deploying will be done by clicking a button in github.

### Reason for Changes

Automate tasks

### Test Coverage

Not yet tested

### Documentation

Internal documentation on how to release will need to be updated.

### Additional Information

- I'm following [these steps](https://github.com/google-github-actions/deploy-appengine#setup) to use the deploy-appengine action. I need to set up a service account and some secrets before I can actually run this.
- I still need to set the version number based on package.json
- @cpcallen after your build changes are merged, I'll need to use them and make sure that I'm copying over the correct built files
- I'm using [uploading and downloading artifacts](https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow) to share information between steps
- I could pass in a flag to determine whether to switch over traffic to the new version or not (`promote: false` or `promote: true`)
- The prepare step puts files in `../_deploy`. I don't know if that's actually accessible from a workflow

### DIFF
--- a/.github/workflows/appengine_deploy.yml
+++ b/.github/workflows/appengine_deploy.yml
@@ -1,0 +1,54 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Deploy to App Engine
+
+# Controls when the workflow will run
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  prepare:
+    name: Prepare
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Prepare demo files
+        run: |
+          npm install
+          npm run prepareDemos
+
+      - name: Upload
+        uses: actions/upload-artifact@v2
+        with:
+          name: appengine_files
+          path: _deploy/
+
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    needs: prepare
+    steps:
+      - name: Download prepared files
+        uses: actions/download-artifact@v2
+        with:
+          name: appengine_files
+          path: _deploy/
+
+      - name: Deploy to App Engine
+        uses: google-github-actions/deploy-appengine@v0.2.0
+        with:
+          deliverables: app.yaml
+          working_directory: _deploy/
+          # TODO: Set up project id and credentials secrets
+          project_id: ${{ secrets.GCP_PROJECT }}
+          credentials: ${{ secrets.GCP_SA_KEY }}
+          promote: false
+          # TODO: Generate a version string based on package.json
+          version: vtest
--- a/.github/workflows/appengine_deploy.yml
+++ b/.github/workflows/appengine_deploy.yml
@@ -1,4 +1,4 @@
-# This is a basic workflow to help you get started with Actions
+# Workflow that prepares files and deploys to appengine
 
 name: Deploy to App Engine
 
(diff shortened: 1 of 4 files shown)


.github/workflows/appengine_deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: Prepare demo files
+        run: |
+          npm install
```
moniika, line 24: I believe the reason we need to `npm install` is because we need the dependencies for the advanced playground to be installed so we can copy them over.

We could optionally add a comment here about that because it's something we've forgotten to do in the past.

.github/workflows/appengine_deploy.yml
```diff
@@ -0,0 +1,54 @@
+# This is a basic workflow to help you get started with Actions
```
moniika, line 1: I'd prefer this description be updated to describe this specific workflow.

.github/workflows/appengine_deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+
+name: Deploy to App Engine
+
+# Controls when the workflow will run
+on:
+  # Allows you to run this workflow manually from the Actions tab
```
moniika, line 7: I like this comment on `workspace_dispatch`. It's not immediately apparent from the syntax that this is what it means so I think it'll be helpful for people newer to Github actions reading the code.

.github/workflows/appengine_deploy.yml
```diff
@@ -0,0 +1,54 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Deploy to App Engine
+
+# Controls when the workflow will run
```
moniika, line 5: Some of these comments feel like they're not adding much value because they're obvious. Such as this one and also I feel like: L10, L14, and L17.
I assume they originated from a template, like the top description.

moniika (approved):
I understand we're probably going to test this after it's merged in, so I'm going to approve.
I only have a few straightforward comments on the solely on the comments in the workflow file.

moniika:
Re: the promote flag (pros and cons). Our current system where we don't promote when we deploy is nice in that we can test the deploy works fine before migrating traffic. I like that extra protection, but admit that it is a little inconvenient to need to go to patheon and navigate the site to do the migration.

cpcallen:
> it is a little inconvenient to need to go to patheon and navigate the site to do the migration.

Since the migrate can be done with the `gcloud` command line tool, perhaps having the appropriate command presented to the user at a suitable point in time could make life easier?

.github/workflows/appengine_deploy.yml
rachel-fenichel: Updated.

rachel-fenichel:
Another option is to have separate workflows for deploying and promoting--that way we run them all from github actions, but we still have time to test in before promoting.

> Since the migrate can be done with the gcloud command line tool, perhaps having the appropriate command presented to the user at a suitable point in time could make life easier?

One of the goals is to not need to have gcloud installed locally (and set up, and authed) in order to do this.

rachel-fenichel:
I merged this to develop and tried to test it, but manually running a workflow only works if it's defined in the default branch (master for us) so I'll need to make a PR with these changes against master instead.

The new PR is https://github.com/google/blockly/pull/5005

cpcallen:
> One of the goals is to not need to have gcloud installed locally (and set up, and authed) in order to do this.

Then my suggestion was terrible and you are right to ignore it!

PR #4999, opened 2021-07-08
Enforce connection preconditions for setParent

### LINKED ISSUE #4989: Calling a.setParent(b) without connecting blocks a and b can cause major rendering issues and errors
**Describe the bug**

Calling `a.setParent(b)` on blocks a and b without connecting them can cause rendering issues where the parent and unconnected child block can move together with varying degrees of space in between them.  If `a` had no parent originally, then attempting to delete `b` results in an unmovable `b` and error messages on any event.

**To Reproduce (2 possibilities)**

Steps to reproduce the behavior (two different types):
1. Go to the Blockly Playground.
2. Import the following XML:
```
  <block type="text_print" id="s+b@l4y{Nw=CeqI3/qzW" x="38" y="38">
    <value name="TEXT">
      <shadow type="text" id="^QF~!eeGDNIEJE8Ye7ji">
        <field name="TEXT">abc</field>
      </shadow>
      <block type="text_join" id="M$%c|%@HCw^J:hg7*J%t">
        <value name="ADD0">
          <block type="text" id="w,TRg]2*bD`k[5+`I$VE">
            <field name="TEXT"></field>
          </block>
        </value>
      </block>
    </value>
  </block>
</xml>
```
3a. Enter the following in the Developer console:
`workspace.getAllBlocks().find(block => block.type === "text").setParent(workspace.getAllBlocks().find(block => block.type === "text_print"))`
4a. There will be a "Still connected to parent block error."  Move the `text_join` block around the workspace and watch the text block jump all around.  

OR...

3b. Disconnect the text block and then enter the code from 3a in the console.
4b. No error this time, but if you move the print block, you will notice that the disconnected text block moves together with the print block.  Attempt to delete the print block and the errors ensue.

**Expected behavior**

Should `setParent` be a `package` method?  I'm not sure why it should be called for any other purpose than connecting/disconnecting blocks.  In this case (outside of `Blockly.Block` and `Blockly.BlockSvg`), it should not be needed outside of the `Connection` class.  The only other places I could find it being called were in testing to set the parent to `null` here:
 https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/tests/mocha/event_test.js#L975

and (indirectly) here:

https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/tests/mocha/event_test.js#L977

jschanker:
## The basics

- [x] I branched from develop

### Proposed Changes

Throws error when attempting to set a block's parent to a non-null block `b` if `b` is `not` properly connected to it without changing state (i.e., if its previous/output connection is *not* connected to one of `b`'s input/next connections or when it doesn't have such connections).  Also throws error when attempting to set a block's parent to `null` if it has a connection to a superior block.

#### Behavior Before Change

Setting a parent of a block already connected to some other superior block would remove the block from that old parent's child list before throwing, leading to potential rendering issues.  Setting the parent of a disconnected block could cause rendering issues and errors when attempting to delete that parent.

#### Behavior After Change

Errors are thrown in these cases without changing state so none of these problems ensue.

### Reason for Changes

To enforce the desired precondition for `setParent` that a non-`null` inputted parent should already be a superior block via the proper connections and that a block should not be connected to a superior one when attempting to set its new parent to `null`.

### Test Coverage

Tested on Blockly Playground to make sure that it resolved the desired bugs.

### Documentation

No additional documentation required.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -766,7 +766,7 @@ Blockly.Block.prototype.setParent = function(newParent) {
     // This block hasn't actually moved on-screen, so there's no need to update
     //     its connection locations.
   } else {
-    // New parent must be non-null so remove this block from the workspace's 
+    // New parent must be non-null so remove this block from the workspace's
     //     list of top-most blocks.
     this.workspace.removeTopBlock(this);
   }
--- a/core/block.js
+++ b/core/block.js
@@ -747,7 +747,7 @@ Blockly.Block.prototype.setParent = function(newParent) {
 
   // Check that block is connected to new parent if new parent is not null.
   //    After todo is addressed, throw error if block is not disconnected from
-  //    superior block when setting newParent to null. 
+  //    superior block when setting newParent to null.
   var connection = this.previousConnection || this.outputConnection;
   var isConnected = !!(connection && connection.targetBlock());
 
@@ -756,8 +756,8 @@ Blockly.Block.prototype.setParent = function(newParent) {
   } else if (!isConnected && newParent) {
     throw Error('Block not connected to new parent.');
   } else if (isConnected && !newParent) {
-    throw Error('Cannot set parent to null while block is still connected to'
-        + ' superior block.');
+    throw Error('Cannot set parent to null while block is still connected to' +
+        ' superior block.');
   }
 
   if (this.parentBlock_) {
--- a/core/block.js
+++ b/core/block.js
@@ -745,9 +745,8 @@ Blockly.Block.prototype.setParent = function(newParent) {
     return;
   }
 
-  // Check that block is connected to new parent if new parent is not null.
-  //    After todo is addressed, throw error if block is not disconnected from
-  //    superior block when setting newParent to null.
+  // Check that block is connected to new parent if new parent is not null and
+  //    that block is not connected to superior one if new parent is null.
   var connection = this.previousConnection || this.outputConnection;
   var isConnected = !!(connection && connection.targetBlock());
 
--- a/core/block.js
+++ b/core/block.js
@@ -755,11 +755,9 @@ Blockly.Block.prototype.setParent = function(newParent) {
     throw Error('Block connected to superior one that is not new parent.');
   } else if (!isConnected && newParent) {
     throw Error('Block not connected to new parent.');
-  } else if (connection && !newParent) {
-    // TODO #4989: Clean up code so that error can be thrown in this case 
-    //     without breaking things.
-    // throw Error('Cannot set parent to null while block is still connected
-    //    to superior block.');
+  } else if (isConnected && !newParent) {
+    throw Error('Cannot set parent to null while block is still connected to'
+        + ' superior block.');
   }
 
   if (this.parentBlock_) {
--- a/core/block.js
+++ b/core/block.js
@@ -763,13 +763,11 @@ Blockly.Block.prototype.setParent = function(newParent) {
     // Remove this block from the old parent's child list.
     Blockly.utils.arrayRemove(this.parentBlock_.childBlocks_, this);
 
-    // Reflect disconnection from superior blocks.
-    this.parentBlock_ = null;
-
     // This block hasn't actually moved on-screen, so there's no need to update
-    // its connection locations.
+    //     its connection locations.
   } else {
-    // Remove this block from the workspace's list of top-most blocks.
+    // New parent must be non-null so remove this block from the workspace's 
+    //     list of top-most blocks.
     this.workspace.removeTopBlock(this);
   }
 
--- a/core/block.js
+++ b/core/block.js
@@ -741,21 +741,34 @@ Blockly.Block.prototype.getChildren = function(ordered) {
  * @param {Blockly.Block} newParent New parent block.
  */
 Blockly.Block.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  if (newParent === this.parentBlock_) {
     return;
   }
+
+  // Check that block is connected to new parent if new parent is not null.
+  //    After todo is addressed, throw error if block is not disconnected from
+  //    superior block when setting newParent to null. 
+  var connection = this.previousConnection || this.outputConnection;
+  var isConnected = !!(connection && connection.targetBlock());
+
+  if (isConnected && newParent && connection.targetBlock() !== newParent) {
+    throw Error('Block connected to superior one that is not new parent.');
+  } else if (!isConnected && newParent) {
+    throw Error('Block not connected to new parent.');
+  } else if (connection && !newParent) {
+    // TODO #4989: Clean up code so that error can be thrown in this case 
+    //     without breaking things.
+    // throw Error('Cannot set parent to null while block is still connected
+    //    to superior block.');
+  }
+
   if (this.parentBlock_) {
     // Remove this block from the old parent's child list.
     Blockly.utils.arrayRemove(this.parentBlock_.childBlocks_, this);
 
-    // Disconnect from superior blocks.
-    if (this.previousConnection && this.previousConnection.isConnected()) {
-      throw Error('Still connected to previous block.');
-    }
-    if (this.outputConnection && this.outputConnection.isConnected()) {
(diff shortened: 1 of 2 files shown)


BeksOmega:
Hi jschanker, thanks for posting this! I'm just about to head out for today, so I'll get a review back to you on Monday =)

Also sorry about creating all of those linked issues when I was testing some automations stuff (whoops) I'll see if I can get those cleaned up.

core/block.js
```diff
@@ -741,21 +741,31 @@ Blockly.Block.prototype.getChildren = function(ordered) {
  * @param {Blockly.Block} newParent New parent block.
  */
 Blockly.Block.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  if (newParent === this.parentBlock_) {
```
BeksOmega, line 744: Thanks for fixing this while you were in the area :D

core/block.js
```diff
@@ -741,21 +741,31 @@ Blockly.Block.prototype.getChildren = function(ordered) {
...
+  if (isConnected && newParent && connection.targetBlock() !== newParent) {
+    throw Error('Block connected to superior one that is not new parent.');
+  } else if (!isConnected && newParent) {
+    throw Error('Block not connected to new parent.');
+  } else if (isConnected && !newParent) {
+    throw Error('Cannot set parent to null while block is still connected to' +
+        ' superior block.');
+  }
+
   if (this.parentBlock_) {
```
BeksOmega, line 762: I know this isn't exactly related to what you're doing, but do you think this if-statement and the below one could be combined? It seems like they're doing the same things (updating child blocks and top blocks).

BeksOmega (changes requested):
Thanks for putting up this fix :D I think it's good to enforce these changes programmatically in addition to fixing the visibility.

I think the only thing this change really needs is some tests. You can either choose to add them, or [file an issue](https://github.com/google/blockly/issues/new?assignees=&labels=type%3A+bug%2C+triage&template=bug_report.md) so that someone can add them later. I think there are 5 cases that need to be covered:
  * Properly setting the parent to null
  * Properly setting the parent to a new not-null parent
  * Cases for each of your error situations

But if you think of others that's rad.

The best place to add these tests is probably in the [block_test.js mocha file](https://github.com/google/blockly/blob/master/tests/mocha/block_test.js). I can't remember if you've written tests before, so here are the [mocha docs](https://mochajs.org/#tdd), and the [chai docs](https://www.chaijs.com/api/assert/) just in case. You will probably want to assert [equal](https://www.chaijs.com/api/assert/#method_equal), [throws](https://www.chaijs.com/api/assert/#method_throws), and maybe [doesNotThrow](https://www.chaijs.com/api/assert/#method_throws).

Thank you again for sending this in! The code looks great :D

core/block.js
jschanker: Happy to do so!

core/block.js
jschanker: I could be missing something, but I'm not sure how to cleanly group those two if statements given the overlap of the following conditional tasks.

```
                             Remove from old parent  Add to new parent Remove top block  Add top block
Has old/new parent block              Y                     Y                 N               N
Only has old parent block             Y                     N                 N               Y
Only has new parent block             N                     Y                 Y               N               
```

So for an example, having an if statement selecting only for the first case would mean duplicating code to remove/add parents.  But maybe that's fine?  I could also remove the second else and instead conditionally add a top block if there's no new parent by nesting an if statement inside the first, but I'm not sure that makes the code cleaner?  Thoughts?

I did however remove the unnecessary statement setting `this.parentBlock_` to `null` (and associated comment) as it unconditionally gets set to `newParent` later.

jschanker:
> Thanks for putting up this fix :D I think it's good to enforce these changes programmatically in addition to fixing the visibility.

Happy to help!

> I think the only thing this change really needs is some tests. 

Absolutely!  Added the tests.

> I can't remember if you've written tests before

I have for https://github.com/google/blockly/pull/4876, and I hope it's okay that I cleaned up the ones I wrote in this pull request (creating the workspace in the setup and tearing it down upon completion).

> Thank you again for sending this in! The code looks great :D

My pleasure again and thanks for reviewing!

tests/mocha/block_test.js
```diff
@@ -890,6 +890,96 @@ suite('Blocks', function() {
...
+            .bind(this.textJoinBlock, this.printBlock));
+        assertNonParentAndOrphan(this.printBlock, this.textJoinBlock, 'TEXT');
+      });
+      test('Setting to new parent when orphan', function() {
+        this.textBlock.outputConnection.disconnect();
+        // when new parent has no available input connection
+        chai.assert.throws(this.textBlock.setParent
+            .bind(this.textBlock, this.printBlock));
+        this.textJoinBlock.outputConnection.disconnect();
+        // when new parent has available input connection
```
BeksOmega, line 963:
```suggestion
        // When new parent has available input connection.
```

tests/mocha/block_test.js
```diff
@@ -890,6 +890,96 @@ suite('Blocks', function() {
...
+      });
+      test('Setting to same parent after disconnecting from it', function() {
+        this.textJoinBlock.outputConnection.disconnect();
+        chai.assert.throws(this.textJoinBlock.setParent
+            .bind(this.textJoinBlock, this.printBlock));
+        assertNonParentAndOrphan(this.printBlock, this.textJoinBlock, 'TEXT');
+      });
+      test('Setting to new parent when orphan', function() {
+        this.textBlock.outputConnection.disconnect();
+        // when new parent has no available input connection
```
BeksOmega, line 959:
```suggestion
        // When new parent has no available input connection.
```

tests/mocha/block_test.js
```diff
@@ -890,6 +890,96 @@ suite('Blocks', function() {
...
+        assertNonParentAndOrphan(this.printBlock, this.textJoinBlock, 'TEXT');
+        assertBlockIsOnlyChild(this.textJoinBlock, this.textBlock, 'ADD0');
+      });
+      test('Setting to same parent after disconnecting from it', function() {
+        this.textJoinBlock.outputConnection.disconnect();
+        chai.assert.throws(this.textJoinBlock.setParent
+            .bind(this.textJoinBlock, this.printBlock));
+        assertNonParentAndOrphan(this.printBlock, this.textJoinBlock, 'TEXT');
+      });
+      test('Setting to new parent when orphan', function() {
```
BeksOmega, line 957: I like how you tested both with available inputs and without :D

tests/mocha/block_test.js
```diff
@@ -890,6 +890,96 @@ suite('Blocks', function() {
...
+        chai.assert.doesNotThrow(this.textBlock.setParent
+            .bind(this.textBlock, this.printBlock));
+        assertBlockIsOnlyChild(this.printBlock, this.textBlock, 'TEXT');
+      });
+      test('Setting to new parent while connected to other block', function() {
+        // setting to grandparent with no available input connection
+        chai.assert.throws(this.textBlock.setParent
+            .bind(this.textBlock, this.printBlock));
+        this.textJoinBlock.outputConnection.disconnect();
+        // setting to block with available input connection
```
BeksOmega, line 945:
```suggestion
        // Setting to block with available input connection.
```

BeksOmega (commented):
Thanks for adding tests, they look great! Just a few formatting changes, then should be gtg :D I think there should be some UI for you to commit them automagically.

> I hope it's okay that I cleaned up the ones I wrote in this pull request (creating the workspace in the setup and tearing it down upon completion).

Oh yeah definitely! In my opinion it's way better to clean up code than to worry about being sticklery on PRs.

tests/mocha/block_test.js
jschanker: Thanks!

jschanker:
Great, made the formatting changes.  Thanks again!

BeksOmega (approved):
LGTM! Thanks again for adding these checks :D

PR #5002, opened 2021-07-08
Add a github action to generate release forum post content

### LINKED ISSUE #4973: Github Action to collect contributors list

When we post the release notes to the forum we want to be able to shout out contributors. If we had a way to automatically create a list of people, that would make sure we don't miss anyone.

A list of all users who created a pull request that got merged in the past quarter. Each user in the list should have a list of of the pull requests they contributed, for easy referencing.

Something like:
```
* UserName ([#num](link)) ([#num2](link2))
* UserName2 ([#num3](link3))
```

Or possibly:
```
* UserName
  * ([#num](link)) Title
  * ([#num2](link2)) Title
* UserName2
  * ([#num3](link3)) Title
```

Depending on what we can get github to spit out and where we want to route it.

N/A because we're not certain how we want it to work yet

**Additional context**

This is part of making the release process more automated.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Closes #4973 

### Proposed Changes

Adds a gulp script that generates the release forum post template. Then posts an issue telling someone to fill in the blanks and put it on the forums. [Here](https://github.com/BeksOmega/blockly/issues/127) is an example I created in my fork.

The template includes:
  * All of the pull requests created (and not closed) by contributors since the previous major/minor release.
  * The description of the latest release, which is generated by Abby's script: https://github.com/google/blockly/pull/5009

### Reason for Changes

Make creating the release forum post easier/smoother.

### Testing
Currently you can trigger this using:
```
npm run listReleaseNotes -- --key <your github personal access token>
```


scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+
+**Issues Closed**:
+${await getClosedIssues(headers)}
+
+**Pull Requests**:
+${await getMergedPrs(headers)}`;
+
+  console.log(note);
+
+  done();
```
cpcallen, line 61: As noted above: don't call `done` here.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+Full release notes are below. As always, we welcome bug reports and pull requests!
+
+Cheers,
+{Your Name}
+
+**Issues Closed**:
+${await getClosedIssues(headers)}
+
+**Pull Requests**:
+${await getMergedPrs(headers)}`;
```
cpcallen, line 57: Newline before the closing backtick, both so output ends with newline and to make it easier to spot the closing delimiter of the long multi-line template literal.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+ */
+
+/**
+ * @fileoverview Gulp scripts for creating release notes.
+ */
+
+var gulp = require('gulp');
+var fetch = require('node-fetch');
+
+const baseUrl = "https://api.github.com/graphql";
```
cpcallen, line 14: [Style guide for constants](https://google.github.io/styleguide/jsguide.html#naming-constant-names): `BASE_URL`.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+        const pull = pulls[i];
+        const createdDate = Date.parse(pull.createdAt);
+        if (createdDate < goalDate) {
+          cursor = '';
+          break;
+        }
+
+        if (collaborators.has(pull.author.login)) {
+          continue;
+        }
```
cpcallen, line 91: The style guide [allows this to be a single line if you want](https://google.github.io/styleguide/jsguide.html#formatting-braces-all):

```JS
if (collaborators.has(pull.author.login)) continue;
```

I often find this more readable for "escape clauses" at the top of a loop, but I will leave it to your judgment whether you think this improves readability in this particular case or not.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+        contributions.push({
+          title: pull.title,
+          url: pull.url,
+          number: pull.number
+        })
+      }
+
+    } catch (error) {
+        console.log(error);
```
cpcallen, line 105: Indentation: should be +2, not +4.  (Maybe just run `clang-format` on the whole file?)

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+  } while (cursor);
+
+  let notes = '';
+  for (let [name, contributions] of contributors) {
+    notes += `* ${name}\n`;
+    for (let contribution of contributions) {
+      let {title, url, number} = contribution;
```
cpcallen, line 113: [Prefer `const` to `let`](https://google.github.io/styleguide/jsguide.html#features-use-const-and-let)—and I believe both these lines can use `const`.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+          contributors.set(pull.author.login, contributions);
+        }
+        contributions.push({
+          title: pull.title,
+          url: pull.url,
+          number: pull.number
+        })
+      }
+
+    } catch (error) {
```
cpcallen, line 104: Wrapping such a big block of code in `try`/`catch` seems like  code smell.  Do you really want to continue iterating if _anything_ in the block throws?

* If the idea is just to ensure that errors are correctly reported to/by gulp, then I'd move the `try`/`catch` to the top-level Gulp task (in this case `generateReleaseNotes`—but note that you will want to report to Gulp that the task failed, and the easiest way to do that is to have the promise reject, which can be achieved simply by _not handling the exception_ in the top-level `async function`.  (Do verify this works as you expect, obviously.)
* If the intention is to retry failed queries then I'd try to reduce the scope of the `try` block to contain only the statement(s) that you are expecting might fail and whose failure should not unwind the stack.  Add a `continue` to the `catch` block if you want to skip the post-query processing that would otherwise follow.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+
+async function getClosedIssues(headers) {
+  return '';
+}
+
+async function getMergedPrs(headers) {
+  return '';
+}
+
+function makeContributorQuery(repo, baseRef, cursor) {
```
cpcallen, line 150: It looks like this query will fetch a list of pull requests / "contributions" (rather than "contributors"), so perhaps `makePRQuery` or `makeContributionsQuery`?

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+async function getCollaborators(repo, headers) {
+  try {
+    const response = await fetch(baseUrl, {
+      method: "POST",
+      headers: headers,
+      body: makeCollaboratorQuery(repo),
+    });
+    const json = await response.json();
+    const collaborators =  json.data.repository.collaborators.nodes
+        .map((node) => node.login);
```
cpcallen, line 129: [Prefer to break lines at a higher syntactic level](https://google.github.io/styleguide/jsguide.html#formatting-where-to-break) (and N.B. parens not required for single-argument arrow functions [but see tip](https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions) about why you might want them):

```JS
    const collaborators =
        json.data.repository.collaborators.nodes.map(node => node.login);
```
(Better double-check my line lengths, though!)

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+    collaborators.push('dependabot');
+    return new Set(collaborators);
+  } catch (error) {
```
cpcallen, line 132: See comment above re: large `try`/`catch` blocks.

scripts/gulpfiles/release_notes.js
```diff
@@ -0,0 +1,201 @@
...
+}
+
+function processContributors(goalDate, pullRequestData) {
+}
+
+const contributors = gulp.series(generateReleaseNotes);
+
+module.exports = {
+  contributors: contributors,
+}
```
cpcallen, line 201: Are you expecting the `contributors` task to have other sub-tasks?  If not, just write:

```JS
module.exports = {
  contributors: generateReleaseNotes,
}
```

cpcallen (commented):
I'm not really competent to review the graphql query code, but I've provided some other comments on gulp-related matters as well as general JavaScript style.

package.json
```diff
@@ -44,7 +44,8 @@
     "test:generators": "tests/scripts/run_generators.sh",
     "test:compile:advanced": "gulp buildAdvancedCompilationTest",
     "typings": "gulp typings",
-    "updateGithubPages": "gulp gitUpdateGithubPages"
+    "updateGithubPages": "gulp gitUpdateGithubPages",
+    "contributors": "gulp contributors"
```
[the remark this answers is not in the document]
BeksOmega: Hmmm I can't think of anything else we'd be listing/posting so I'll go with `listReleaseNotes` for now =)

scripts/gulpfiles/release_notes.js
BeksOmega: Thanks for mentioning this, it should definitely be reorganized. My intention is to retry failed queries, but I'm confused about what you mean by:
> Add a `continue` to the `catch` block if you want to skip the post-query processing that would otherwise follow.

Could you give a code sample for what this would look like? I think that would really help clear things up for me :D

cpcallen: Sure.  A very simple example might look like this:

```JS
while (true) {
  let input = prepareInput();
  let output;
  try {
    output = thisMightFail(input);
  } catch (err) {
    handleError();
    continue;
  }
  postprocess(output);
}
```

This will cause exceptions thrown by `prepareInput` or `postprocess` (which we do not expect to throw) to terminate the loop, while exceptions from `thisMightFail` (which we do expect to throw) are dealt with—say, by logging them.  The `continue` ensures that `postprocess` only gets called if `thisMightFail` succeeds.

Compare to this version which also only runs `postprocess` if `thisMightFail` succeeds, but will swallow unexpected exceptions from `postprocess` along with the expected ones from `thisMightFail`:

```JS
while (true) {
  let input = prepareInput();
  let output;
  try {
    output = thisMightFail(input);
    postprocess(output);
  } catch (err) {
    handleError();
  }
}
```

.github/workflows/create_release_notes.yml
```diff
@@ -0,0 +1,16 @@
...
+name: 'Release Notes'
+on:
+  release:
+    types: [published]
+
+jobs:
+  notes:
+    steps:
+      - name: Create Notes
+        run: npm run listReleaseNotes -- --key ${{ secrets.GITHUB_TOKEN }}
```
cpcallen, line 16: GitHub notes no trailing newline.  Probably not strictly necessary, but could help avoid finding a bug in the yaml parser…

scripts/gulpfiles/release_notes.js
cpcallen: > Also do you have a recommendation for how best to retry failed queries? Should everything go in a loop?

I'm not an expert in this area but I'd say your general approach here looks pretty good.  The things I would check are:

* What happens if the first request fails?  If I am reading correctly, it looks like `cursor` won't get set and the loop will terminate without retrying.

* What happens if a subsequent request fails?  It looks like cursor will be unchanged and so it will retry at the same position on the next loop iteration (provided nothing after `cursor = …` throws).

Note that you may want to put an upper limit on the number of failures.  This could be global (give up after N failed requests in total), or per-request (give up if any individual request fails N times); in the latter case putting the request in an inner `do {request} while (!success && !tooManyAttempts)` loop could be helpful.

cpcallen: (There's also the small matter of what to do if the request succeeds but the returned data is invalid / truncated.  Retrying the same request may or may not help.)

BeksOmega:
Turns out that whenever this creates an issue, the issue adds references to all of the issues/pull requests linked in the body (slipped my mind when I was making it). And there's no way to remove those references, not even programmatically using the graphql or REST apis. So it may be better to not have it create the issue.

BeksOmega:
Closing this for now. If you peoples still want this later I can reopen it.

PR #5005, opened 2021-07-09
Appengine deploy workflow on master for testing

rachel-fenichel:
This is a version of https://github.com/google/blockly/pull/4998 against master so that the action is available to be triggered manually.


moniika (approved):
Same comments as from https://github.com/google/blockly/pull/4998

PR #5011, opened 2021-07-09
Update positionable jsdoc

moniika:
## The basics

- [x] I branched from develop

### Resolves

n/a

### Proposed Changes

Update JsDoc for positionables to allow for null bounding box to be returned, indicating that this positionables position should be ignored.

### Reason for Changes

Feature for positionables. Useful for UI elements like in workspace-search that are meant to not affect positioning of other positioanbles.

### Additional Information
I forgot I had this branch. Creating this PR also to test workflow that comments on PR.

### DIFF
--- a/core/trashcan.js
+++ b/core/trashcan.js
@@ -506,7 +506,8 @@ Blockly.Trashcan.prototype.position = function(metrics, savedPositions) {
 /**
  * Returns the bounding rectangle of the UI element in pixel units relative to
  * the Blockly injection div.
- * @return {!Blockly.utils.Rect} The UI elements’s bounding box.
+ * @return {?Blockly.utils.Rect} The UI elements’s bounding box. Null if
+ *   bounding box should be ignored by other UI elements.
  */
 Blockly.Trashcan.prototype.getBoundingRectangle = function() {
   var bottom = this.top_ + this.BODY_HEIGHT_ + this.LID_HEIGHT_;
--- a/core/interfaces/i_positionable.js
+++ b/core/interfaces/i_positionable.js
@@ -37,6 +37,7 @@ Blockly.IPositionable.prototype.position;
 /**
  * Returns the bounding rectangle of the UI element in pixel units relative to
  * the Blockly injection div.
- * @return {!Blockly.utils.Rect} The UI elements’s bounding box.
+ * @return {?Blockly.utils.Rect} The UI elements’s bounding box. Null if
+ *   bounding box should be ignored by other UI elements.
  */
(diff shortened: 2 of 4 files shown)


core/trashcan.js
```diff
@@ -506,7 +506,8 @@ Blockly.Trashcan.prototype.position = function(metrics, savedPositions) {
 /**
  * Returns the bounding rectangle of the UI element in pixel units relative to
  * the Blockly injection div.
- * @return {!Blockly.utils.Rect} The UI elements’s bounding box.
+ * @return {?Blockly.utils.Rect} The UI elements’s bounding box. Null if
+ *   bounding box should be ignored by other UI elements.
```
cpcallen, line 510: Note: I _think_ you could have left the type signature alone here, since `{!Type}` is a subtype of `{?Type}`.  It would avoid breaking someone's code in the (unlikely!) event that they were depending on the return value of `Trashcan.p.getBoundingRectangle` being non-nullable but don't care about the interface.

(Same for `ZoomControls`, obviously.)

cpcallen (approved):
> Feature for positionables. Useful for UI elements like in workspace-search that are meant to not affect positioning of other positioanbles.

I am not familiar enough with the codebase to know if it is a reasonable concern or not, but the first question that pops into my head is "will there be cases where one wants to query the `boundingRect` of a positionable even though we don't want it to affect positioning of other positionables?"

If the answer is yes then it might be better to signal this a different way, such as by having a boolean flag (or method returning such) on the positionable to indicate whether it should affect others or not.

core/trashcan.js
moniika: I didn't realize it was a subtype, that's interesting.
This method was introduced as part of the `IPositionable` api (which hasn't yet been widely advertised), so its *probably* unlikely that someone is depending on.

I'd like to make the argument that making this nullable gives some extra flexibility if we, or a plugin, wanted to make the trashcan hide-able (where it might make sense for them to have this return null).

moniika:
> I am not familiar enough with the codebase to know if it is a reasonable concern or not, but the first question that pops into my head is "will there be cases where one wants to query the `boundingRect` of a positionable even though we don't want it to affect positioning of other positionables?"
> 
> If the answer is yes then it might be better to signal this a different way, such as by having a boolean flag (or method returning such) on the positionable to indicate whether it should affect others or not.

We don't have very many example use cases at the moment as this is a relatively new api. 
A flag seems like it could also work and still be filtered out from the array before it's passed in `position`.

I'm hesitant to introduce a new property vs changing the jsdoc as it feels like a bigger change.
I'll file a bug and reference this PR so we can address it later.

PR #5015, opened 2021-07-09
Move deploy temp dir into the Blockly directory so that it can be use…

rachel-fenichel:
…d from github actions

## The basics

- [x] I branched from develop

### Resolves

Part of #4988 

### Proposed Changes

Make the prepareDemos script put files in a `_deploy` directory in the repository instead of up a level. This is necessary because github actions cannot access outputs that are up a level.

This PR is deliberately against develop--changes to files other than the workflow itself can be on develop, because I can choose which branch to run the workflow against.

#### Behavior Before Change

Files were placed in `../_deploy`

#### Behavior After Change

Files are placed in `_deploy`

### Test Coverage

Tested by running `npm run prepareDemos` from the command line.

### DIFF
--- a/scripts/gulpfiles/appengine_tasks.js
+++ b/scripts/gulpfiles/appengine_tasks.js
@@ -17,8 +17,8 @@ var execSync = require('child_process').execSync;
 
 var packageJson = require('../../package.json');
 
-const demoTmpDir = '../_deploy';
-const demoStaticTmpDir = '../_deploy/static';
+const demoTmpDir = '_deploy';
+const demoStaticTmpDir = '_deploy/static';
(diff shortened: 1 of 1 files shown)


samelhusseini:
Can we make use of the new `build` directory for this. Just have a `deploy` folder in that?

rachel-fenichel:
@samelhusseini Yes, but my goal is to get the github actions + appengine side of this working with minimal changes to the rest of blockly. Changing the directory here is much smaller than redoing the deploy script entirely, and would both simplify a step of our release process and be a proof of concept for using github actions to deploy to appengine.

PR #5018, opened 2021-07-11
Build system changes cleanup

cpcallen:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Proposed Changes

* Fixes possible Windows incompatibility (see #5000)
* Fixes issue with eslint tests failing due to (incorrectly) linting build output.

#### Behavior Before Change

* Build depends on existence of `mkdir -p` command.
* `npm test` will fail if `npm run clean` (or `npm run clean:build`) isn't run first.

#### Behavior After Change

* Tests succeed even if build directory not cleaned first.

PR #5019, opened 2021-07-12
Preparation for goog.module transition: base.js, deps.js

cpcallen:
## The basics

- ~~[ ] I branched from develop~~
- ~~[ ] My pull request is against develop~~
- [X] I branched from goog_module
- [X] My pull request is against goog_module

### Proposed Changes

* Use the current standard Closure Library version of `closure/goog/base.js`.
* In uncompiled mode, load dependency graph from `tests/deps.js`.
  * New gulp task `buildDeps` (`npm run build:deps`) to create `tests/deps.js`.
  * `blockly_uncompressed.js` is now handwritten.
    * And simplified; in particular, BLOCKLY_BOOT is gone.
    * And linted!
  * Old gulp task `buildUncompressed` is deleted.
* For consistency with the Closure documentation and base.js, consistently refer to what blockly_uncompressed.js is used for as "uncompiled mode" rather than "uncompressed mode" (but don't yet rename it to blockly_uncompiled.js).
* Have `eslint` ignore "undefined" global variable `exports`, used to specify exports from `goog.module` modules.

### Reason for Changes

The version of base.js we have been using is a cut-down copy of an older version, stripped of everything that wasn't strictly needed for Blockly.  We now need things that were previously stripped out, so restore the current version from https://github.com/google/closure-library/blob/master/closure/goog/base.js.  We can always strip out unused sections again in future - once we know which sections those are.

Loading dependency graph info from `tests/deps.js` (instead of having it embedded in `blockly_uncompressed.js` helps maintain a clear separation between generated and handwritten code, avoids having code duplicated between the generator script and its output, and helped find some actual errors the handwritten parts because they can now be linted.  It also more closely follows the Closure Library convention.

### Additional Information

This PR (or another to the same effect) must be merged to the `goog_module` branch before any others that actually  carrying out `goog.provides` -> `goog.module` migration.


core/contextmenu_items.js
```diff
@@ -191,7 +191,9 @@ Blockly.ContextMenuItems.addDeletableBlocks_ = function(block, deleteList) {
   if (block.isDeletable()) {
     Array.prototype.push.apply(deleteList, block.getDescendants(false));
   } else {
-    var children = /** @type !Array<!Blockly.BlockSvg> */ (block.getChildren(false));
+    /* eslint-disable indent */
```
rachel-fenichel, line 194: This disables the rule for the rest of the file. Use `eslint-disable-line` or `eslint-disable-next-line` instead.

https://eslint.org/docs/user-guide/configuring/rules#disabling-rules

rachel-fenichel (approved):
Approved after lint fix.

cpcallen:
Blocked on #5018.

core/contextmenu_items.js
cpcallen: TIL.  Ta!

PR #5020, opened 2021-07-12
Fix vertical button padding in horizontal flyout

### LINKED ISSUE #4914: Buttons in horiz flyouts are missing vertical padding
**Describe the bug**

I think that horizontal flyouts should give buttons a bit of vertical padding on the bottom, which currently they do not.

**To Reproduce**

1. Go to the advanced playground, and set it to a horizontal layout, with the category toolbox.
2. Open up the variables category.
3. Observe how the create variable button is right against the bottom of the flyout.

**Expected behavior**

There should probably be some padding below the button (ie space between the bottom of the button and the bottom of the flyout).

**Screenshots**

variable-button

**Additional Info**
It seems like this might be a good candidate for a "help-wanted" issue. If you want to work on this issue, I think you should probably start by checking out [layout_](https://github.com/google/blockly/blob/8d8309e7e76a7c1d0aa5427fc4d1e0b0f9b0d011/core/flyout_horizontal.js#L238) and [reflowInternal_](https://github.com/google/blockly/blob/8d8309e7e76a7c1d0aa5427fc4d1e0b0f9b0d011/core/flyout_horizontal.js#L339) for the horizontal flyout. My guess is that the [flyoutHeight here](https://github.com/google/blockly/blob/8d8309e7e76a7c1d0aa5427fc4d1e0b0f9b0d011/core/flyout_horizontal.js#L343) needs to take into account the sizes of buttons and labels in addition to blocks, but I'm not sure =)

hpnrep6:
## The basics

- [x] I branched from develop

### Resolves

Fixes #4914.

### Proposed Changes

Uses the height of flyout buttons in the calculation to find the height of the flyout.

#### Behavior Before Change

Screenshot from 2021-07-12 01-32-00

#### Behavior After Change

Screenshot from 2021-07-12 01-00-32

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -343,6 +343,10 @@ Blockly.HorizontalFlyout.prototype.reflowInternal_ = function() {
   for (var i = 0, block; (block = blocks[i]); i++) {
     flyoutHeight = Math.max(flyoutHeight, block.getHeightWidth().height);
   }
+  var buttons = this.buttons_;
+  for (var i = 0, button; (button = buttons[i]); i++) {
+    flyoutHeight = Math.max(flyoutHeight, button.height);
+  }
   flyoutHeight += this.MARGIN * 1.5;
   flyoutHeight *= this.workspace_.scale;
   flyoutHeight += Blockly.Scrollbar.scrollbarThickness;


hpnrep6:
@googlebot I signed it!

gonfunko (approved):
This looks great, thank you for tracking it down and fixing it!

PR #5021, opened 2021-07-12
Group field edits together for undo/redo (attempt #2)

NeilFraser:
Reverts google/blockly#4965


cpcallen:
I was wondering why you were reverting, then I saw that the title was

> Revert "Revert "Group field edits together for undo/redo"" #5021

cpcallen (approved):
Approving this on the basis that it is identical to previous Group field edits together for undo/redo
#4952

Issue #4977, opened 2021-07-01
Create Github Action to calculate release bundle size delta


gonfunko:
We should create a Github action to calculate the change in size of the compressed Blockly distribution between the current release and the previous one. This would be incorporated into the release notes.

cpcallen:
Related prior art: [`check_metadata.sh`](https://github.com/google/blockly/blob/master/tests/scripts/check_metadata.sh)

PR #5033, opened 2021-07-12
Added script and workflow to automatically update build artifact sizes in check_metadata.sh

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
https://github.com/google/blockly/issues/4977

### Proposed Changes
Adds a script that calculates the size of the various Blockly JS files and updates check_metadata.sh with the new values.

#### Behavior Before Change
As part of the release process, a developer would need to manually determine the build artifact sizes and update the check_metadata.sh script.

#### Behavior After Change
Running the workflow will automatically calculate the new build artifact sizes, update the script, and create a PR with the updated script.

### Reason for Changes
Makes this step in the release process more reliable and reduces the effort on the part of the dev team.

### DIFF
--- a/.github/workflows/update_metadata.yml
+++ b/.github/workflows/update_metadata.yml
@@ -6,13 +6,8 @@ name: Update Metadata
 on: [workflow_dispatch]
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        node-version: [16.x]        
+  update-metadata:
+    runs-on: ubuntu-latest        
 
     steps:
     - name: Check Out Blockly
@@ -20,10 +15,22 @@ jobs:
       with:
         ref: 'develop'
 
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js 16.x
       uses: actions/setup-node@v1
       with:
-        node-version: ${{ matrix.node-version }}        
+        node-version: 16.x
+
+    - name: Build Blockly
+      run: npm run build:compressed
+
+    - name: Build Blockly blocks
+      run: npm run build:blocks
+
+    - name: Gzip Blockly
+      run: gzip -k build/blockly_compressed.js
+
+    - name: Gzip Blockly blocks
+      run: gzip -k build/blocks_compressed.js
 
     - name: Update Metadata
       run: source ./tests/scripts/update_metadata.sh
@@ -31,9 +38,9 @@ jobs:
     - name: Create Pull Request
       uses: peter-evans/create-pull-request@9825ae65b1cb54b543b938503728b432a0176d29
       with:
-        commit-message: Updated build artifact sizes in check_metadata.sh
+        commit-message: Update build artifact sizes in check_metadata.sh
         delete-branch: true
-        title: Updated build artifact sizes in check_metadata.sh
+        title: Update build artifact sizes in check_metadata.sh
 
     - name: View Pull Request
       run: echo "View Pull Request - ${{ steps.cpr.outputs.pull-request-url }}"
--- a/.github/workflows/update_metadata.yml
+++ b/.github/workflows/update_metadata.yml
@@ -1,0 +1,39 @@
+# This workflow updates the check_metadata.sh script, which compares the current
+# size of build artifacts against their size in the previous version of Blockly.
+
+name: Update Metadata
+
+on: [workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [16.x]        
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/update_metadata.yml
```diff
@@ -0,0 +1,39 @@
...
+on: [workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [16.x]
```
cpcallen, line 15: As you're just running on a single OS / `node` version combo there's no need for a `matrix` and thus no need for for a `strategy` either.

.github/workflows/update_metadata.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+    - name: Check Out Blockly
+      uses: actions/checkout@v2
+      with:
+        ref: 'develop'
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
```
cpcallen, line 26: Is this needed?  It looks like you're directly running a shell script which doesn't rely on `node` or `npm` at all (but see note below).

.github/workflows/update_metadata.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}        
+
+    - name: Update Metadata
+      run: source ./tests/scripts/update_metadata.sh
+
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@9825ae65b1cb54b543b938503728b432a0176d29
+      with:
+        commit-message: Updated build artifact sizes in check_metadata.sh
```
cpcallen, line 34: Suggestion: I think commit messages usually read better when they are in the present tense ("Update build artifact sizes…").

.github/workflows/update_metadata.yml
```diff
@@ -0,0 +1,39 @@
...
+        delete-branch: true
+        title: Updated build artifact sizes in check_metadata.sh
```
cpcallen, line 36: Ditto pull request titles.

tests/scripts/update_metadata.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Determines the size of generated files and updates check_metadata.sh to
+# reflect the new values.
+
+blockly_size=$(wc -c < "blockly_compressed.js")
+blocks_size=$(wc -c < "blocks_compressed.js")
+blockly_gz_size=$(wc -c < "blockly_compressed.js.gz")
+blocks_gz_size=$(wc -c < "blocks_compressed.js.gz")
```
cpcallen, line 9: One note: you are measuring the sizes of the checked-in build products, rather than rebuilding from scratch.  That's fine (at least until we stop checking in build products), but it might be worth making that clear in appropriate places to avoid confusion down the line.

(Alternatively, since you are installing `node.js` anyway, you could `npm run build:compressed` and then measure the sizes of the copies in `build/` instead.)

.github/workflows/update_metadata.yml
gonfunko: The script does rely on `npx` for determining the package version.

.github/workflows/update_metadata.yml
gonfunko: Done.

.github/workflows/update_metadata.yml
gonfunko: Updated accordingly.

.github/workflows/update_metadata.yml
gonfunko: Here too.

tests/scripts/update_metadata.sh
gonfunko: Updated to build and calculate the size of the new artifacts.

gonfunko (commented):
Thanks for the review! Incorporated your suggested changes, PTAL!

.github/workflows/update_metadata.yml
```diff
@@ -6,34 +6,41 @@ name: Update Metadata
...
+      run: npm run build:compressed
+
+    - name: Build Blockly blocks
+      run: npm run build:blocks
+
+    - name: Gzip Blockly
+      run: gzip -k build/blockly_compressed.js
+
+    - name: Gzip Blockly blocks
+      run: gzip -k build/blocks_compressed.js
```
cpcallen, line 33: Having these steps logged in the CI output is nice, but I think it might be better to put these in `update_metadata.sh`—both to be consistent with `check_metadata.sh`, and because it allows the script to be run locally without having to do the `gzip` manually.

gonfunko: Good call, moved them to update_metadata.sh.

cpcallen (approved):

PR #5034, opened 2021-07-12
Adds a serializer test suite - project cereal

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on Project Cereal

### Proposed Changes

Adds a suite for testing serializers, as outlined [here](https://docs.google.com/document/d/1Dbpezxsd4H1S0p31ncrZ1hLaMfFYBeanv8pqjkpxoHI/edit?resourcekey=0-hYVVlo828cnLmnlj30Lysg#heading=h.g8wz2wuda80j). Once we're sure that this suite covers enough cases, this will be moved into the [dev-tools test helpers](https://github.com/google/blockly-samples/tree/master/plugins/dev-tools/src) (probably around the end of my internship).

Currently this just tests that the XML round-trips correctly, but the idea is that you can plug in other serializers (including the JSON serializer).

### Reason for Changes

Gives us confidence that the new JSON serializer will properly round-trip state. Also once this is a test helper it will make it easy for devs to test their custom serializers.

### Documentation

When this moves into the dev-tools package this will need to be added to the readme. But for now no docs are needed.

### Additional Information

@moniika you don't need to give this a full review, but I would appreciate your expert eye giving it a quick look to make sure there are no glaring errors hehe

### DIFF
--- a/tests/mocha/serializer_test.js
+++ b/tests/mocha/serializer_test.js
@@ -1733,17 +1733,10 @@ var runSerializerTestSuite = (serializer, deserializer, testSuite) => {
     return createTestFunction(suiteOrTest);
   };
 
-  // TODO: Fix after suites method is being exported.
-  var runTests = function(suite) {
-    if (suite.testSuites && suite.testSuites.length) {
-      suite.testSuites.forEach(runTests);
-    }
-    if (suite.testCases && suite.testCases.length) {
-      testHelpers.runTestCases(suite.testCases, createTestFunction);
-    }
-  };
+  let suiteCall = (testSuite.skip ? suite.skip : suite);
+  suiteCall = (testSuite.only ? suite.only : suiteCall);
 
-  suite(testSuite.title, function() {
+  suiteCall(testSuite.title, function() {
     setup(function() {
       this.workspace = new Blockly.Workspace();
     });
@@ -1752,11 +1745,9 @@ var runSerializerTestSuite = (serializer, deserializer, testSuite) => {
       this.workspace.dispose();
     });
 
-    runTests(testSuite);
-    // TODO: Fix after this method is being exported.
-    // testHelpers.runTestSuites(
-    //     testSuite.testSuites, createSuiteOrTestFunction);
-    // testHelpers.runTestCases(testSuite.testCases, createTestFunction);
+    testHelpers.runTestSuites(
+        testSuite.testSuites, createSuiteOrTestFunction);
+    testHelpers.runTestCases(testSuite.testCases, createTestFunction);
   });
 };
 
--- a/tests/mocha/serializer_test.js
+++ b/tests/mocha/serializer_test.js
@@ -49,8 +49,8 @@ Serializer.Attributes.Basic = new SerializerTestCase('Basic',
     '</xml>');
 Serializer.Attributes.IdChars = new SerializerTestCase('IdChars',
     '<xml xmlns="https://developers.google.com/blockly/xml">' +
-    '<block type="logic_negate" id="!#$%()*+,-./:;=?@[]^_" x="42" y="42"></block>' +
-    '<block type="logic_negate" id="`{|}~ABCDEFGHIJKLMNO" x="42" y="42"></block>' +
+    '<block type="logic_negate" id="!#$%()*+,-./:;=?@[]^" x="42" y="42"></block>' +
+    '<block type="logic_negate" id="_`{|}~ABCDEFGHIJKLMN" x="42" y="42"></block>' +
     '<block type="logic_negate" id="OPQRSTUVWXYZabcdefgh" x="42" y="42"></block>' +
     '<block type="logic_negate" id="ijklmnopqrstuvwxyz01" x="42" y="42"></block>' +
     '<block type="logic_negate" id="23456789!#$%()*+,-./" x="42" y="42"></block>' +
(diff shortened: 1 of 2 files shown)


tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+
+Serializer.Empty = new SerializerTestCase('Empty',
+    '<xml xmlns="https://developers.google.com/blockly/xml"></xml>'
+);
+Serializer.Data = new SerializerTestCase('Data',
+    '<xml xmlns="https://developers.google.com/blockly/xml">' +
+    '<block type="logic_negate" id="id******************" editable="false" x="42" y="42">' +
+    '<data>test data</data>' +
+    '</block>' +
+    '</xml>');
```
moniika, line 46: This seems a little weird that these test cases (and the test cases defined later) are both on the TestSuite object directly and in the test cases. I would think it would only be necessary to add on `testCases` and i would just use a local variable when you are creating the test case.

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+
+Serializer.testSuites = [
+  Serializer.Attributes,
+  Serializer.Fields,
+  Serializer.Icons,
+  Serializer.Connections,
+  Serializer.Mutations,
+];
+
+var runSerializerTestSuite = (serializer, deserializer, testSuite) => {
```
moniika, line 1715: I know you're planning on refactoring this code out of core later, but for readability I would create a `serializer_test_helpers.js` which has the definitions for:
- `SerializerTestCase`
- `SerializerTestSuite`
- `runSerializerTestSuite`

and in this file:
- declare/create the test suites and test cases
- call `runSerializerTestSuite`

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+    setup(function() {
+      this.workspace = new Blockly.Workspace();
+    });
+
+    teardown(function() {
+      this.workspace.dispose();
+    });
+
+    testHelpers.runTestSuites(
+        testSuite.testSuites, createSuiteOrTestFunction);
```
moniika, line 1749: Why do you need to check if it is a test suite or a test function? the contents of testSuites should only be of type `TestSuite`

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+];
+
+var runSerializerTestSuite = (serializer, deserializer, testSuite) => {
+
+  const createTestFunction = function(test) {
+    return function() {
+      Blockly.Xml.domToWorkspace(
+          Blockly.Xml.textToDom(test.xml), this.workspace);
+      if (serializer && deserializer) {
+        // do custom serialization and deserialization.
```
moniika, line 1722: I think it would be nice to make this a TODO to note that the serializer and deserializer aren't yet used in these tests.

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+      chai.assert.equal(Blockly.Xml.domToText(newXml), test.xml);
+    };
+  };
+
+  const createSuiteOrTestFunction = function(suiteOrTest) {
+    if (suiteOrTest instanceof testHelpers.TestSuite) {
+      return createSuiteOrTestFunction;
+    }
+    return createTestFunction(suiteOrTest);
+  };
```
moniika, line 1734: Looking at how `runTestSuites` currently works, I'm not sure this is correct.
The expected function passed to `runTestSuites` is:
`function(!TestSuite<T>):(function(T):!Function`
It expects a `TestSuite` to be passed and the returned function should be a function that takes a `TestCase`.

In the case of nested `TestSuite`s, I think it should still work because this callback is only [invoked on `testCases`](https://github.com/google/blockly-samples/blob/ea55892df2097ac35dc3b3f8b83ab1103bdd51b5/plugins/dev-tools/src/common_test_helpers.mocha.js#L89).

tests/mocha/serializer_test.js
BeksOmega: Yeah it is a bit weird haha. But the intention was to make it easy for external developers to skip/only suites/tests. If you only include things in the `testSuites` or `testCases` arrays then code for skipping becomes:
```
Serializer.testSuites[1].testSuites[0].testCases[4].skip = true;
```
as opposed to
```
Serializer.Fields.Angle.SmallestNumber.skip = true;
```

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1754 @@
...
+
+  let suiteCall = (testSuite.skip ? suite.skip : suite);
+  suiteCall = (testSuite.only ? suite.only : suiteCall);
+
+  suiteCall(testSuite.title, function() {
+    setup(function() {
+      this.workspace = new Blockly.Workspace();
```
BeksOmega, line 1741: Should this be calling `sharedTestSetup` @moniika  ?

tests/mocha/serializer_test.js
BeksOmega: I like this idea, but I think I'm going to leave it for now because everything is going to be moving into samples (including all of the test cases). I think if we wait we'll have a better chance of structuring it so that it makes the most sense for that package.

tests/mocha/serializer_test.js
moniika: Yeah, I would add `sharedTestSetup` and `sharedTestTeardown`.
Although it's not in Samples, it's not much work to add a call here and It's helprf me catch bugs with tests before, so it could be useful.

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1757 @@
...
+/**
+ * The XML we want to ensure round-trips through the serializer.
+ */
+SerializerTestCase.prototype.xml = '';
+
+
+/**
+ * Constructs a serializer test suite.
+ * @param {string} title The title of this test suite.
+ * @extends {TestSuite}
```
moniika, line 31: `@extends {TestSuite<SerializerTestCase>}`

There is a also second template type defined on `TestCase` but I think it should be removed. I'll link you in the PR for discussion.

moniika (changes requested):

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1757 @@
...
+ */
+
+
+// TODO: Move this into samples as part of the dev-tools package.
+/**
+ * Constructs a serializer test.
+ * @param {string} title The title of this testcase.
+ * @param {string} xml The XML to use for the round-trip test.
+ * @constructor
+ * @extends {TestCase}
```
moniika, line 14: Correction on my previous comment, based on my reading, I think this is supposed to be `@implements`. The other one needs to be `@exports` because there is a generic template type involved.

tests/mocha/serializer_test.js
BeksOmega: Good idea! Because I'm not sure how this works with exporting either, and that'll be easier to figure out once it's actually in samples.

BeksOmega:
@moniika I think that everything's fixed! @maribethb Do you want to take a look before this goes in or does Monica's review cover it? Abby took a look yesterday to stay up-to-date but didn't review.

moniika (approved):
LGTM

PR #5037, opened 2021-07-13
Fix code in code demo executing twice on mobile

### LINKED ISSUE #1949: Code Demo on iOS: Double execution upon tapping run
### Problem statement

On iOS Safari and Chrome, in the Blockly Code Demo, tapping the run button will execute the code twice.

### Expected Behavior

Tapping the run button should only execute the code once.

### Steps to Reproduce

1. On an iPad, open the Code Demo in either Safari or Chrome
2. Add one print block from the Text category.
3. Tap the run button.

Observe the alert dialog shows twice.

* Device: iPad mini, 1st gen
  * OS: 9.3.5
  * Browser Safari, Chrome

hpnrep6:
## The basics

- [x] I branched from develop

### Resolves

Fixes #1949

### Proposed Changes

Only bind one event listener depending on platform.

#### Behavior Before Change

Code executes twice, one after the other because of the two event listeners bound to the "run" button.

#### Behavior After Change

Code executes only once.

### Test Coverage

Tested on chromium devtools mobile viewport mode, where the bug was also present.

### DIFF
--- a/demos/code/code.js
+++ b/demos/code/code.js
@@ -187,7 +187,7 @@ Code.bindClick = function(el, func) {
   if (typeof el == 'string') {
     el = document.getElementById(el);
   }
-  if (Blockly.utils.userAgent.MOBILE) {
+  if (Blockly.utils.userAgent.MOBILE || Blockly.utils.userAgent.TABLET) {
     el.addEventListener('touchend', func, true);
   } else {
     el.addEventListener('click', func, true);
--- a/demos/code/code.js
+++ b/demos/code/code.js
@@ -549,7 +549,7 @@ Code.initLanguage = function() {
 /**
  * Execute the user's code.
  * Just a quick and dirty eval.  Catch infinite loops.
- * @param {UIEvent} event Event created from listener bound to the function.
+ * @param {Event} event Event created from listener bound to the function.
  */
 Code.runJS = function(event) {
   // Prevent code from being executed twice on touchscreens.
--- a/demos/code/code.js
+++ b/demos/code/code.js
@@ -187,11 +187,8 @@ Code.bindClick = function(el, func) {
   if (typeof el == 'string') {
     el = document.getElementById(el);
   }
-  if (Blockly.utils.userAgent.MOBILE || Blockly.utils.userAgent.TABLET) {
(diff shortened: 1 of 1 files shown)


NeilFraser:
I'm not sure that mobile/tablet detection is the right way to resolve this issue, since it would appear to disable touch actions on devices such as Chromebooks.  There's more discussions here:
https://stackoverflow.com/questions/25572070/javascript-touchend-versus-click-dilemma

Sadly I don't personally have any iOS devices at my disposal to test, so I'm not the right person to offer definitive opinions.  Just raising a concern.

This is also relevant to Blockly Games since it has identical code:
https://github.com/google/blockly-games/blob/master/appengine/js/lib-games.js#L317

hpnrep6:
Yes, that will indeed a problem with the changes. I've made a second (more like third) commit that will stop the ClickEvent from executing if a TouchEvent is being executed simultaneously.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent

Although one thing is that I've been unable to reproduce this issue on neither my iPad air 3rd gen running iPadOS 14.6 (tested on firefox and safari browser) or pixel 4a 5G running android 11 (tested on firefox browser), and I was only able to reproduce this issue using the browser devtools mobile mode, so perhaps some things have changed in the way iOS browsers handle touch events since when this issue was first reported.

demos/code/code.js
```diff
@@ -549,8 +549,14 @@ Code.initLanguage = function() {
 /**
  * Execute the user's code.
  * Just a quick and dirty eval.  Catch infinite loops.
+ * @param {UIEvent} event Event created from listener bound to the function.
```
moniika, line 552: I think this type is supposed to be `Event`.

moniika (commented):
Just a comment about type in the JSDoc, otherwise lgtm.
Thanks for the PR and looking into supporting both TouchEvent and MouseEevent.

demos/code/code.js
hpnrep6: Changed it to `Event`!

moniika (approved):

PR #5038, opened 2021-07-13
Update package-lock.json to lockfileVersion 2

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Update `package-lock.json` from `lockfileVersion: 1` to `lockfileVersion: 2`.

### Reason for Changes

Version 7 of `npm` does this automatically; as a consequence, most of the team are using `npm` v6 to avoid this update.  But I believe the update is harmless:

* `lockfileVersion: 2` is backward-compatible with `lockfileVersion: 1`, so `npm` v6 can work fine with new file (although it does print a warning, because it can't know that this file-format-from-the-futrure is fine).
* `npm` v7 is backwards compatible as far as `node.js` v10, so this update does not prevent anyone from using `npm` v7 with any of the older versions of `node.js` we wish to continue supporting.

### Test Coverage

Awaiting CI tests on GitHub.

### Additional Information

You can update `npm` to the latest (v7.x) by running `npm install -g npm`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,12510 @@
 {
   "name": "blockly",
   "version": "6.20210701.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "6.20210701.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "jsdom": "15.2.1"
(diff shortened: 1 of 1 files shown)


maribethb (approved):
Thanks for the explanation of why this is safe! We should probably do this for all of blockly-samples as well.

Issue #5026, opened 2021-07-12
Migrate Blockly from goog.provide to goog.module


cpcallen:
This is a tracking bug for the work the Blockly team is doing in Q3 2021 to transition the codebase from `goog.provide` to [`goog.module`](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide), [as recently announced](https://groups.google.com/g/blockly/c/J_-2EMaYBXg).

This bug will track any outstanding tasks not covered by other more specific bugs, provide documentation for how the migration will be done and so on.

**If you have questions about why we are doing this work and what the consequences will be for developers who use Blockly [please ask them in the Blockly forum](https://groups.google.com/g/blockly/c/J_-2EMaYBXg)** rather than commenting on this bug.

This work will be carried out on the `goog_module` branch.  When working on migration, please branch from and submit PRs to merge back to `goog_module`, not `develop`.

# Module Conversion Guide

Most of the work required to migrate from `goog.provide` to `goog.module` can be done on a module-by-module (and often file-by-file) basis.

The end goal is that each `.js` file will contain a single module, with a clearly defined interface, that can be imported independently of any other modules—except of course for that module's own dependencies, which it will import for its own internal use.  Modules should normally only reexport identifiers from other modules when they are aggregating a number of submodules together (possibly for legacy compatibility reasons).

Since we're visiting every file in the project, we're also going to use this opportunity to begin the migration to ES6 and the [new JavaScript styleguide](https://google.github.io/styleguide/jsguide.html).

## Converting a single `.js` file

There are several mostly-independent steps needed to convert a single `.js` source file.  In most cases it will be preferable to start conversion with leaf modules—i.e., ones which do not `require` any others—but within any given module that does have imports it will probably be easier to start by dealing with its `goog.require`s before dealing with its `goog.provide`s.

Where possible you should present the conversion of a single `.js` file as a pull against the `goog_module` branch, consisting of four commits, as follows:

1. Update file to ES6 const/let.
2. Rewrite the `goog.provide` statement as `goog.module` and explicitly enumerate exports.
3. Rewrite`goog.requires` statements and add missing requires.
4. Run `clang-format` on the whole file.

In situations where step 2 requires combining or splitting several related `.js` files (see details below) then it is appropriate that they be presented together in a single PR.

### 1. Update the file to ES6 const/let

[...]

PR #5049, opened 2021-07-14
Migrate core/utils/keycodes.js to goog.module

moniika:
## The basics

- ~~[ ] I branched from develop~~
- ~~[ ] My pull request is against develop~~
- [X] I branched from goog_module
- [X] My pull request is against goog_module

### Resolves

Part of https://github.com/google/blockly/issues/5026

### Proposed Changes

Converts `core/utils/keycodes.js` to `goog.module` syntax. 

### Test Coverage

Ran `npm run start` and `npm run test`

### Additional Information

No instances of var to update.
No changes to the file after running clang-format.

### DIFF
--- a/core/utils/keycodes.js
+++ b/core/utils/keycodes.js
@@ -16,7 +16,8 @@
  * @name Blockly.utils.KeyCodes
  * @namespace
  */
-goog.provide('Blockly.utils.KeyCodes');
+goog.module('Blockly.utils.KeyCodes');
+goog.module.declareLegacyNamespace();
 
 
 /**
@@ -29,7 +30,7 @@ goog.provide('Blockly.utils.KeyCodes');
  *
  * @enum {number}
  */
-Blockly.utils.KeyCodes = {
+const KeyCodes = {
   WIN_KEY_FF_LINUX: 0,
   MAC_ENTER: 3,
   BACKSPACE: 8,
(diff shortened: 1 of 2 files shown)


core/utils/keycodes.js
```diff
@@ -166,3 +167,5 @@ Blockly.utils.KeyCodes = {
   // http://en.community.dell.com/support-forums/laptop/f/3518/p/19285957/19523128.aspx
   PHANTOM: 255
 };
+
+exports = KeyCodes;
```
BeksOmega, line 171: Is this exporting `Blockly.utils.KeyCodes.KeyCodes` now instead of `Blockly.utils.KeyCodes`?

Christopher's write-up makes me think so, but I'm really not sure:
> ```
> // Following goog.provides('MyModule'):
> /** @const */
> MyModule.BLUE = '#0000ff';
> /** @param ...  @return ... */
> MyModule.colourNameToHex(name) { ... };
> ```
> becomes:
> ```
> /** @const */
> const BLUE = '#0000ff';
> /** @param ...  @return ... */
> function colourNameToHex(name) { ... };
> 
> exports = {BLUE, colourNameToHex};
> ```

cpcallen: My example may be a bit misleading, because I was treating `BLUE` as just an arbitrary constant rather than a member of an enum.  I will try to improve it.

@moniika's code looks correct.  As written, the return value of `goog.require('Blockly.utils.KeyCodes')` will be the `KeyCodes` enum object, which will be compatible with what happened before this commit.  That means code that uses this module can either do `const KeyCodes = require('...KeyCodes');  let code = KeyCodes.BACKSPACE` or `const {BACKSPACE} = require('...KeyCodes'); let code = BACKSPACE;`.

BeksOmega (approved):
LGTM!

PR #5053, opened 2021-07-14
Add serialization of basic block attributes to JSOs

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/basic-hooks...BeksOmega:cereal/block-serialization

### Resolves

Work on project cereal
Dependent on #5052 

### Proposed Changes

Adds serialization of basic block attributes: type, id, x, y, collapsed, deletable, etc.

Does not serialize children, fields, mutators, or icons. Does not handle deserialization.

### Reason for Changes

Serializing state is good.

### Test Coverage

Added unit tests to make sure the serializer is returning an object with the structure we expect.


core/serialization/blocks.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+/**
+ * Adds attributes to the given state object based on the state of the block.
+ * Eg collapsed, disabled, editable, etc.
+ * @param {!Blockly.Block} block The block to base the attributes on.
+ * @param {!Blockly.serialization.blocks.State} state The state object to append
+ *     to.
+ */
+const addAttributes = function(block, state) {
```
BeksOmega, line 76: Update private functions w/ whatever the preferred syntax is before merging.

core/serialization/blocks.js
```diff
@@ -0,0 +1,117 @@
...
+  if (block.isCollapsed()) {
+    state['collapsed'] = true;
+  }
+  if (!block.isEnabled()) {
+    state['disabled'] = true;
+  }
```
BeksOmega, line 82: Do we want to change this to be `state['enabled']` to match the new API?

cpcallen: In general I think 'enabled' flags are easier to understand than 'disabled' ones (and therefore preferable), but given that `undefined` is falsey there might be an argument to using `disabled` if 'enabled' is the default state allowing the flat to be omitted in most cases.

maribethb (approved):

PR #5054, opened 2021-07-14
Add serialization of mutators

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/block-serialization...BeksOmega:cereal/mutator-serialization

### Resolves

Work on project cereal.
Dependent on #5053 

### Proposed Changes

Adds serialization of mutators on blocks. ~~Handles backwards compatibility with the XML hooks by converting them to a string before appending them to the JSO.~~ That was removed in favor of investigating upgrading developers' files directly using a script.

Does not handle deserialization.

### Reason for Changes

Serializing state is good.

### Test Coverage

Added tests for the different categories of objects the save function could return. Also added a test for wrapping the `mutationToDom` hook.

### DIFF
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -127,6 +127,7 @@ function addExtraState(block, state) {
   if (block.saveExtraState) {
     state['extra-state'] = block.saveExtraState();
   } else if (block.mutationToDom) {  // Backwards compatibility.
-    state['extra-state'] = Blockly.Xml.domToText(block.mutationToDom());
+    state['extra-state'] = Blockly.Xml.domToText(block.mutationToDom())
+        .replace('xmlns="https://developers.google.com/blockly/xml"', '');
   }
 }
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -129,12 +129,5 @@ const addExtraState = function(block, state) {
     if (extraState !== null) {
       state['extraState'] = extraState;
     }
-  } else if (block.mutationToDom) {  // Backwards compatibility.
-    const mutation = block.mutationToDom();
-    if (mutation !== null) {
-      state['extraState'] = Blockly.Xml.domToText(mutation)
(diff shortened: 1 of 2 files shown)


core/serialization/blocks.js
```diff
@@ -0,0 +1,143 @@
...
+    if (extraState !== null) {
+      state['extraState'] = extraState;
+    }
+  } else if (block.mutationToDom) {  // Backwards compatibility.
+    const mutation = block.mutationToDom();
+    if (mutation !== null) {
+      state['extraState'] = Blockly.Xml.domToText(mutation)
+          .replace('xmlns="https://developers.google.com/blockly/xml"', '');
+ 
+    }
```
BeksOmega, line 139: Instead of handle the backwawrds compatibility here, we could provide a script w/ the release that automatically add wrapper implementations to people's block definitions.

This might be better because it would simplify this logic, and make it so the JSO system doesn't rely on the XML system. (Plus I might get to break out my python skillz :D )

BeksOmega: Discussed this on the other PR and we decided to remove backwards compatibility for now.

BeksOmega:
@NeilFraser do you want to grab this one?

core/serialization/blocks.js
```diff
@@ -112,3 +114,18 @@ const addCoords = function(block, state) {
   state['x'] = Math.round(workspace.RTL ? workspace.getWidth() - xy.x : xy.x);
   state['y'] = Math.round(xy.y);
 };
+
+/**
+ * Adds any extra state the block may provide to the given state object.
+ * @param {!Blockly.Block} block The block to serialize the extra state of.
+ * @param {!State} state The state object to append
+ *     to.
```
NeilFraser, line 122: This comment can be one line.  Same for line 74 whence this was copied.

NeilFraser (approved):

PR #5059, opened 2021-07-14
Migrate core/utils/string.js to goog.module syntax

gonfunko:
## The basics

### Resolves

Part of #5026

### DIFF
--- a/core/utils/string.js
+++ b/core/utils/string.js
@@ -64,7 +64,8 @@ const commonWordPrefix = function(array, opt_shortest) {
   }
   let wordPrefix = 0;
   const max = opt_shortest || shortestStringLength(array);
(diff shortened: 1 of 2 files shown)


core/utils/string.js
```diff
@@ -135,44 +138,44 @@ Blockly.utils.string.wrap = function(text, limit) {
  * @return {string} Wrapped text.
  * @private
```
maribethb, line 139: Based on the style guide, I don't think this should be marked private. It just doesn't get exported. However, we can wait for Chris to confirm that.

gonfunko: Updated since this has been long since confirmed, and also updated the exports style to reflect our current convention.

maribethb (approved):

PR #5065, opened 2021-07-15
Migrate core/extensions.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/extensions.js` to `goog.module` syntax.


core/extensions.js
```diff
@@ -17,7 +17,8 @@
  * @name Blockly.Extensions
  * @namespace
  */
-goog.provide('Blockly.Extensions');
+goog.module('Blockly.Extensions');
```
cpcallen, line 20: Since this doesn't export a class it should (eventually) be renamed `Blockly.extensions`.  I've added it to the list in #5073.

core/extensions.js
```diff
@@ -28,7 +29,7 @@ goog.requireType('Blockly.Block');
  * The set of all registered extensions, keyed by extension name/id.
  * @private
  */
-Blockly.Extensions.ALL_ = Object.create(null);
+const ALL = Object.create(null);
```
cpcallen, line 31: This object is not "deeply immutable" (it gets used as a mutable dictionary) so the variable should not be named as if it were a constant.  `allExtensions` would probably better name.

Also—as has been discussed at length recently—declaring this `@private` is disallowed by the style guide (though I'm fine with leaving it for now, until we have consensus about the issue).

core/extensions.js
```diff
@@ -440,13 +441,15 @@ Blockly.Extensions.buildTooltipWithFieldText = function(msgTemplate,
...
-        this.tooltipWhenNotConnected_;
+        this.tooltipWhenNotConnected;
   }.bind(this));
 };
-Blockly.Extensions.register('parent_tooltip_when_inline',
-    Blockly.Extensions.extensionParentTooltip_);
+register('parent_tooltip_when_inline',
+    extensionParentTooltip);
+
+exports = {ALL_: ALL, register, registerMixin, registerMutator, unregister, apply, buildTooltipForDropdown, buildTooltipWithFieldText};
```
cpcallen: So this will be something like `exports = {ALL_: allExtensions, …`.

core/extensions.js
```diff
@@ -20,9 +20,8 @@
 goog.module('Blockly.Extensions');
 goog.module.declareLegacyNamespace();
 
-goog.require('Blockly.utils');
-
-goog.requireType('Blockly.Block');
+const Block = goog.requireType('Blockly.Block');
```
cpcallen, line 23: Evidently this is causing spurious eslint errors.

cpcallen (changes requested):

BeksOmega:
Just flagging this is one of the files I touched. See #5052

core/extensions.js
moniika: Thanks for pointing that out.
I'll update the name so it's no longer using CONSTANT_CASE, which is supposed to be reserved for true "constants".

core/extensions.js
moniika: Yes, I brought this up in the team chat. I think we resolved to ignore it. It's an issue with eslint not recognizing that the value is used in the closure annotations.

PR #5063 makes this a warning instead of an error and after we clang-format every file we will revisit updating the eslint config.

cpcallen (approved):

PR #5066, opened 2021-07-15
Migrate core/css.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/css.js` to `goog.module` syntax.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -77,7 +77,7 @@ const inject = function(hasCss, pathToMedia) {
  * Array making up the CSS content for Blockly.
  */
 const CONTENT = [
-`.blocklySvg {
+  `.blocklySvg {
   background-color: #fff;
   outline: none;
   overflow: hidden;  /* IE overflows by default. */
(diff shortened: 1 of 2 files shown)


core/css.js
```diff
@@ -14,28 +14,29 @@
...
+goog.module.declareLegacyNamespace();
 
 
 /**
  * Has CSS already been injected?
  * @type {boolean}
  * @private
  */
-Blockly.Css.injected_ = false;
+let injected_ = false;
```
gonfunko, line 26: I was under the impression that we also wanted to remove underscores (and possibly the @private? A bit unclear if we've decided on that yet)

core/css.js
```diff
@@ -78,478 +78,273 @@ const inject = function(hasCss, pathToMedia) {
  */
 const CONTENT = [
```
gonfunko, line 79: I understand why clang-format did this, but ooph...possibly we want to avoid it here?

core/css.js
moniika: Fixed.

core/css.js
moniika: Yikes. I'm reverting that commit and I'll see if there's a better way to do this.
I'm thinking either to disable clang for these lines in some way or with a template string.

gonfunko (approved):

PR #5069, opened 2021-07-15
Migrate core/blocks.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/blocks.js` to `goog.module` syntax.


core/blocks.js
```diff
@@ -14,10 +14,13 @@
  * A mapping of block type names to block prototype objects.
  * @name Blockly.Blocks
  */
-goog.provide('Blockly.Blocks');
+goog.module('Blockly.Blocks');
+goog.module.declareLegacyNamespace();
 
 /**
  * A mapping of block type names to block prototype objects.
  * @type {!Object<string,Object>}
```
BeksOmega, line 22: Could you update the type annotation while you're here? I think it should be `!Object<string, !Object>`. Also I think this needs the [dict](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#dict) annotation since outside devs add properties to it.

BeksOmega (approved):
LGTM! Just a suggested JSDoc fix while you're in the area.

core/blocks.js
moniika: I've updated the type, but it seems like the `@dict` annotation isn't working when I try to build compressed. 
I'm not going to add the `@dict` for now and I'll see about adding it in a future PR.

PR #5072, opened 2021-07-15
Add serialization of field values

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/mutator-serialization...BeksOmega:cereal/field-serialiation

### Resolves

Work on project cereal.
Dependent on #5054 

### Proposed Changes

Adds serializing field values to JSOs.

Does not handle deserialization.

### Reason for Changes

Serializing state is good.

### Test Coverage

Added tests for different categories of types the field could return. Also added a test for wrapping the default XML implementation.

### Additional Information

I got rid of one of the outer suites, so the test file looks like I made a lot of changes when I didn't. The new field tests are at the bottom of the file.

I'm also planning on adding tests for individual fields as I add JSO serialization to them.


core/serialization/blocks.js
```diff
@@ -0,0 +1,165 @@
...
+ *     y: ?number,
+ *     collapsed: ?boolean,
+ *     disabled: ?boolean,
+ *     editable: ?boolean,
+ *     deletable: ?boolean,
+ *     movable: ?boolean,
+ *     inline: ?boolean,
+ *     data: ?string,
+ *     extra-state: *
+ *     fields: ?Object<string, *>
```
alschmiedt, line 40: Should this be `!Object<string, *>|undefined`? since I think if we don't have any fields this does not get set?

alschmiedt, line 40: Should this be `!Object<string, *>|undefined`? since I think if we don't have any fields this does not get set?

alschmiedt (approved):

PR #5076, opened 2021-07-15
Migrate core/utils/toolbox.js to goog.module syntax

alschmiedt:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/utils/toolbox.js` to `goog.module` syntax. 

### Additional Information
This is failing [this](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/tests/mocha/workspace_svg_test.js#L115) test case. It is because `Blockly.utils.toolbox.parseToolboxTree` !== `parseToolbox`. I'm not entirely sure what to do about this. I tried updating the test, but couldn't find a way to require the module with the current set up.

### DIFF
--- a/core/utils/toolbox.js
+++ b/core/utils/toolbox.js
@@ -14,7 +14,8 @@
  * @name Blockly.utils.toolbox
  * @namespace
  */
-goog.provide('Blockly.utils.toolbox');
+goog.module('Blockly.utils.toolbox');
+goog.module.declareLegacyNamespace();
 
 /** @suppress {extraRequire} */
 goog.require('Blockly.constants');
@@ -34,7 +35,7 @@ goog.requireType('Blockly.ToolboxSeparator');
  *            disabled: (string|boolean|undefined)
  *          }}
  */
-Blockly.utils.toolbox.BlockInfo;
+let BlockInfo;
 
 /**
  * The information needed to create a separator in the toolbox.
@@ -45,7 +46,7 @@ Blockly.utils.toolbox.BlockInfo;
  *            cssconfig:(!Blockly.ToolboxSeparator.CssConfig|undefined)
  *          }}
  */
-Blockly.utils.toolbox.SeparatorInfo;
+let SeparatorInfo;
 
 /**
  * The information needed to create a button in the toolbox.
@@ -55,7 +56,7 @@ Blockly.utils.toolbox.SeparatorInfo;
  *            callbackkey:string
  *          }}
  */
-Blockly.utils.toolbox.ButtonInfo;
+let ButtonInfo;
 
 /**
  * The information needed to create a label in the toolbox.
@@ -65,21 +66,21 @@ Blockly.utils.toolbox.ButtonInfo;
  *            id:(string|undefined)
  *          }}
  */
-Blockly.utils.toolbox.LabelInfo;
+let LabelInfo;
 
 /**
  * The information needed to create either a button or a label in the flyout.
- * @typedef {Blockly.utils.toolbox.ButtonInfo|
- *           Blockly.utils.toolbox.LabelInfo}
+ * @typedef {ButtonInfo|
+ *           LabelInfo}
  */
-Blockly.utils.toolbox.ButtonOrLabelInfo;
+let ButtonOrLabelInfo;
 
 /**
  * The information needed to create a category in the toolbox.
  * @typedef {{
  *            kind:string,
  *            name:string,
- *            contents:!Array<!Blockly.utils.toolbox.ToolboxItemInfo>,
+ *            contents:!Array<!ToolboxItemInfo>,
  *            id:(string|undefined),
  *            categorystyle:(string|undefined),
  *            colour:(string|undefined),
@@ -87,7 +88,7 @@ Blockly.utils.toolbox.ButtonOrLabelInfo;
  *            hidden:(string|undefined)
  *          }}
  */
-Blockly.utils.toolbox.StaticCategoryInfo;
+let StaticCategoryInfo;
 
 /**
  * The information needed to create a custom category.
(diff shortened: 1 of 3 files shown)


core/utils/toolbox.js
```diff
@@ -274,12 +274,12 @@ Blockly.utils.toolbox.hasCategories = function(toolboxJson) {
     return false;
   }
 
-  var toolboxKind = toolboxJson['kind'];
+  let toolboxKind = toolboxJson['kind'];
```
cpcallen, line 276: Why is this `let` rather than `var`?

tests/deps.js
```diff
@@ -170,7 +171,7 @@ goog.addDependency('../../core/touch_gesture.js', ['Blockly.TouchGesture'], ['Bl
 goog.addDependency('../../core/trashcan.js', ['Blockly.Trashcan'], ['Blockly.ComponentManager', 'Blockly.DeleteArea', 'Blockly.Events', 'Blockly.Events.TrashcanOpen', 'Blockly.IAutoHideable', 'Blockly.IPositionable', 'Blockly.Options', 'Blockly.Xml', 'Blockly.browserEvents', 'Blockly.constants', 'Blockly.registry', 'Blockly.uiPosition', 'Blockly.utils', 'Blockly.utils.Rect', 'Blockly.utils.Svg', 'Blockly.utils.dom', 'Blockly.utils.toolbox'], {'lang': 'es5'});
 goog.addDependency('../../core/utils.js', ['Blockly.utils'], ['Blockly.Msg', 'Blockly.constants', 'Blockly.utils.Coordinate', 'Blockly.utils.Rect', 'Blockly.utils.colour', 'Blockly.utils.global', 'Blockly.utils.string', 'Blockly.utils.style', 'Blockly.utils.userAgent']);
 goog.addDependency('../../core/utils/aria.js', ['Blockly.utils.aria'], []);
-goog.addDependency('../../core/utils/colour.js', ['Blockly.utils.colour'], [], {'lang': 'es6', 'module': 'goog'});
+goog.addDependency('../../core/utils/colour.js', ['Blockly.utils.colour'], []);
```
cpcallen: There are a bunch of changes to `deps.js` in this commit which seem manifestly wrong—e.g., this commit does not convert `colour.js` back to `goog.provide` so I do not understand why the `'module': 'goog'` option has been deleted here.

I see that most of these erroneous changes were undone by the final commit in this PR, but have you any insight in to how this happened?  Maybe we need to make sure to give `npm run build:deps` a nice hot cup of coffee before putting it to work next time?

tests/deps.js
```diff
@@ -180,13 +181,13 @@ goog.addDependency('../../core/utils/keycodes.js', ['Blockly.utils.KeyCodes'], [
...
-goog.addDependency('../../core/utils/rect.js', ['Blockly.utils.Rect'], [], {'lang': 'es6', 'module': 'goog'});
+goog.addDependency('../../core/utils/rect.js', ['Blockly.utils.Rect'], []);
 goog.addDependency('../../core/utils/size.js', ['Blockly.utils.Size'], []);
 goog.addDependency('../../core/utils/string.js', ['Blockly.utils.string'], []);
 goog.addDependency('../../core/utils/style.js', ['Blockly.utils.style'], ['Blockly.utils.Coordinate', 'Blockly.utils.Size']);
 goog.addDependency('../../core/utils/svg.js', ['Blockly.utils.Svg'], []);
-goog.addDependency('../../core/utils/svg_paths.js', ['Blockly.utils.svgPaths'], [], {'lang': 'es6', 'module': 'goog'});
-goog.addDependency('../../core/utils/toolbox.js', ['Blockly.utils.toolbox'], ['Blockly.Xml', 'Blockly.constants']);
+goog.addDependency('../../core/utils/svg_paths.js', ['Blockly.utils.svgPaths'], []);
+goog.addDependency('../../core/utils/toolbox.js', ['Blockly.utils.toolbox'], ['Blockly.Xml', 'Blockly.constants'], {'lang': 'es6', 'module': 'goog'});
```
cpcallen: (This line is the only change that I see to this file that looks correct.)

core/utils/toolbox.js
```diff
@@ -387,7 +386,7 @@ const addAttributes = function(node, obj) {
 const parseToolboxTree = function(toolboxDef) {
   if (toolboxDef) {
     if (typeof toolboxDef != 'string') {
-      if (Blockly.utils.userAgent.IE && toolboxDef.outerHTML) {
+      if (IE && toolboxDef.outerHTML) {
```
cpcallen, line 390: This (and the corresponding change to the `require` above) is certainly mechanically correct, but I don't think the result is very readable.  Probably the API of the `Blocly.utils.userAgent` module could be improved, but for now I suggest importing all of `useragent` and then writing `useragent.IE`.  That requires no changes outside this file but would help reduce reader confusion.

core/utils/toolbox.js
```diff
@@ -408,4 +409,25 @@ const parseToolboxTree = function(toolboxDef) {
...
+  FlyoutItemInfoArray,
+  ToolboxDefinition,
+  FlyoutDefinition,
+  Position,
+  convertToolboxDefToJson,
+  convertFlyoutDefToJsonArray,
+  hasCategories,
+  isCategoryCollapsible,
+  parseToolboxTree
+}
```
cpcallen, line 433: `eslint` notes the missing semicolon.

tests/deps.js
```diff
@@ -4,6 +4,7 @@ goog.addDependency('../../blocks/logic.js', ['Blockly.Blocks.logic', 'Blockly.Co
 goog.addDependency('../../blocks/loops.js', ['Blockly.Blocks.loops', 'Blockly.Constants.Loops'], ['Blockly', 'Blockly.Blocks', 'Blockly.FieldDropdown', 'Blockly.FieldLabel', 'Blockly.FieldNumber', 'Blockly.FieldVariable', 'Blockly.Warning']);
 goog.addDependency('../../blocks/math.js', ['Blockly.Blocks.math', 'Blockly.Constants.Math'], ['Blockly', 'Blockly.Blocks', 'Blockly.FieldDropdown', 'Blockly.FieldLabel', 'Blockly.FieldNumber', 'Blockly.FieldVariable']);
 goog.addDependency('../../blocks/procedures.js', ['Blockly.Blocks.procedures'], ['Blockly', 'Blockly.Blocks', 'Blockly.Comment', 'Blockly.FieldCheckbox', 'Blockly.FieldLabel', 'Blockly.FieldTextInput', 'Blockly.Mutator', 'Blockly.Warning'], {'lang': 'es5'});
+goog.addDependency('../../blocks/test_blocks.js', ['Blockly.TestBlocks'], ['Blockly', 'Blockly.Blocks'], {'lang': 'es5'});
```
cpcallen, line 7: The other weird changes to `deps.js` got fixed in the final commit, but I'm unclear on why this line has been added.

cpcallen (approved):
I've tweaked the PR title for consistency.

Overall looks good, but the test failure will need to be sorted before merging.

I'm not sure what the best way to do this is: moving to properly encapsulated modules makes the kind of monkey patching that I presume the test is doing no longer work, unfortunately.

core/utils/toolbox.js
cpcallen: (Oh, good: you correctly guessed that I actually meant `const`.)

cpcallen:
Any luck figuring out how to modify the tests?

alschmiedt:
No, not yet. I think the correct way is to import the module ([reference here](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide#how-do-i-write-a-unit-test-for-a-googmodule)), however we would then have to convert our tests to use `goog.module` which we can't load in `index.html` through a scripts tag. You mentioned looking at the `blockly_uncompressed.js` file, but I have a few questions before moving on.

alschmiedt:
Removed the test since it should probably be in a separate test suite with the rest of `Blockly.utils.toolbox`. Filed issue #5139 to track.

cpcallen (approved):
Seems like a reasonable solution.

PR #5077, opened 2021-07-15
feat: upgrade fields to use new JSO hooks

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/field-serialiation...BeksOmega:cereal/field-upgrades

### Resolves

Work on project cereal
Dependent on #5072 

### Proposed Changes

Adds implementations of the new JSO serialization hooks to all of the built-in fields. Fields in blockly-samples will be handled later.

### Test Coverage

Added basic tests to ensure that fields serialize correctly. More edge cases are handled by the round-trip serializer tests.

### DIFF
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -261,7 +261,7 @@ Blockly.FieldAngle.prototype.initView = function() {
  * @package
  */
 Blockly.FieldAngle.prototype.saveState = function() {
-  return /** @type{number} */ (this.value_);
+  return /** @type {number} */ (this.getValue());
 };
 
 /**
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -261,7 +261,7 @@ Blockly.FieldAngle.prototype.initView = function() {
  * @package
  */
 Blockly.FieldAngle.prototype.saveState = function() {
-  return this.value_;
+  return /** @type{number} */ (this.value_);
 };
 
 /**
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -241,8 +241,6 @@ Blockly.FieldAngle.prototype.configure_ = function(config) {
   }
 };
 
-
-
 /**
  * Create the block UI for this field.
  * @package
@@ -254,6 +252,26 @@ Blockly.FieldAngle.prototype.initView = function() {
       Blockly.utils.Svg.TSPAN, {}, null);
   this.symbol_.appendChild(document.createTextNode('\u00B0'));
   this.textElement_.appendChild(this.symbol_);
+};
+
+/**
+ * Saves this field's value.
+ * @return {number} The angle value held by this field.
+ * @override
+ * @package
+ */
+Blockly.FieldAngle.prototype.saveState = function() {
+  return this.value_;
+};
+
+/**
+ * Sets the field's value based on the given state.
+ * @param {*} state The state to apply to the angle field.
+ * @override
+ * @package
+ */
+Blockly.FieldAngle.prototype.loadState = function(state) {
+  this.setValue(state);
 };
 
 /**
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -192,7 +192,7 @@ Blockly.FieldColour.prototype.initView = function() {
  * @package
  */
 Blockly.FieldColour.prototype.saveState = function() {
-  return this.value_;
+  return /** @type{string} */ (this.value_);
 };
 
 /**
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -192,7 +192,7 @@ Blockly.FieldColour.prototype.initView = function() {
  * @package
  */
 Blockly.FieldColour.prototype.saveState = function() {
-  return /** @type{string} */ (this.value_);
+  return /** @type {string} */ (this.getValue());
 };
 
 /**
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -186,6 +186,26 @@ Blockly.FieldColour.prototype.initView = function() {
 };
 
 /**
+ * Saves this field's value.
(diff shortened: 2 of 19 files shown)


core/field_variable.js
```diff
@@ -194,6 +194,28 @@ Blockly.FieldVariable.prototype.toXml = function(fieldElement) {
...
+ * Saves this field's value.
+ * @return {string} The id of the variable referenced by this field.
+ * @override
+ * @package
+ */
+Blockly.FieldVariable.prototype.saveState = function() {
+  // Make sure the variable is initialized.
+  this.initModel();
+  return this.variable_.getId();
+};
```
BeksOmega, line 207: In the `toXml` function we serialize a lot more state (variable names and types). But since that info is now stored separately (and has been for a long time) I don't think it's necessary. All variables should be created before this field is deserialized from XML, so all we need is a reference to it (ie its ID).

If this is wrong, it should come up in the round-trip tests, but please check my logic! I definitely don't want to break variables haha.

BeksOmega:
@maribethb I'd like your opinion on this:
Since I removed the default backwards compatible implementations of `saveState` and `loadState` from the base field class, we could remove the implementations in the sub classes (except for variable). However, leaving them in has two advantages
1) It's less work if we want to add the backwards compatible implementation back in.
2) We get a tiny bit more type safety via explicit casts.

Do you think it's better lo leave them in or remove them?

tests/mocha/field_dropdown_test.js
```diff
@@ -152,4 +152,29 @@ suite('Dropdown Fields', function() {
...
+        const jso = Blockly.serialization.blocks.save(block);
+        chai.assert.deepEqual(jso['fields'], {'DROPDOWN': value});
+      };
+    });
+
+    teardown(function() {
+      workspaceTeardown.call(this, this.workspace);
+    });
+
+    test('Simple', function() {
```
alschmiedt, line 176: Would it be worth it to add a test for dynamic dropdowns?

core/field_colour.js
```diff
@@ -185,6 +185,26 @@ Blockly.FieldColour.prototype.initView = function() {
...
+ * @override
+ * @package
+ */
+Blockly.FieldColour.prototype.saveState = function() {
+  return /** @type{string} */ (this.value_);
+};
+
+/**
+ * Sets the field's value based on the given state.
+ * @param {*} state The state to apply to the colour field.
```
alschmiedt, line 200: Could this type be `string`? If we know that we save a string and anything other than a string is invalid for this field?

alschmiedt (approved):

tests/mocha/field_dropdown_test.js
BeksOmega: Yes good call, I should also add something for that to the round-trip tests (if I haven't already)

core/field_colour.js
BeksOmega: Sadly this cannot be a string :/ The super class accepts `*` so all subclasses must also accept `*`. Here's an example of how this works type-system wise:
```typescript
class AnimalPen {
  constructor() { }

  feed (animal: Animal) {

  }
}

class CowPen extends AnimalPen {
  constructor() { }

  // This is not allowed b/c it is more restrictive.
  feed (cow: Cow) {

  }
}

// CowPen is an AnimalPen so this is legal (ignoring the bad type of feed).
const animalPen: AnimalPen = new CowPen();

const turkey: Turkey = new Turkey();

// Animal pen accepts any animal, so this is legal.
// However, animalPen is actually a CowPen and as such, would blow up.
// Which is why you are not allowed to restrict parameter types in subtypes.
animalPen.feed(turkey);
```

If `Field` was a parameterized type, as outlined in [this issue](https://github.com/google/blockly/issues/5258), we would be able to fix this. But it's not so we can't :/

BeksOmega:
@alschmiedt are you cool with leaving the explicit implementations in? or do you think it's better to pull them? See [this comment](https://github.com/google/blockly/pull/5077#issuecomment-890002660).

> Since I removed the default backwards compatible implementations of saveState and loadState from the base field class, we could remove the implementations in the sub classes (except for variable). However, leaving them in has two advantages
>   1) It's less work if we want to add the backwards compatible implementation back in.
>   2) We get a tiny bit more type safety via explicit casts.
>
> Do you think it's better lo leave them in or remove them?

core/field_colour.js
alschmiedt: Ah, yup. Sorry wishful thinking : )

core/field_angle.js
```diff
@@ -256,6 +254,26 @@ Blockly.FieldAngle.prototype.initView = function() {
...
 };
 
+/**
+ * Saves this field's value.
+ * @return {number} The angle value held by this field.
+ * @override
+ * @package
+ */
+Blockly.FieldAngle.prototype.saveState = function() {
+  return /** @type{number} */ (this.value_);
```
alschmiedt, line 264: Can these use `getValue()` like the `field.saveState`?

alschmiedt:
I think it is fine in the interim while we are still trying to figure out the best story for backwards compatibility. However, I would add this to the list of things to revisit before releasing. My only problem with keeping it in is we are essentially duplicating this code everywhere, so any change to saveState and loadState would have to be updated in every field.

BeksOmega:
> I think it is fine in the interim while we are still trying to figure out the best story for backwards compatibility. However, I would add this to the list of things to revisit before releasing. My only problem with keeping it in is we are essentially duplicating this code everywhere, so any change to saveState and loadState would have to be updated in every field.

Cool cool, I've added that to my task backlog. I'll get this moved to `getValue()` and merge this tomorrow!

PR #5085, opened 2021-07-16
Migrate core/interfaces/i_accessibility.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/interfaces/i_accessibility.js` to `goog.module` syntax.

### Additional Information

This PR breaks splits `core/interfaces/i_accessibility.js` into multiple files for each interface that it provided.


core/interfaces/i_ast_node_location.js
```diff
@@ -0,0 +1,20 @@
...
+
+'use strict';
+
+goog.provide('Blockly.IASTNodeLocation');
+
+/**
+ * An AST node location interface.
+ * @interface
+ */
+Blockly.IASTNodeLocation = function() {};
```
cpcallen: Aside: I am skeptical about this empty interface.  Knowing that `X` implements `IASTNodeLocation` tells us precisely nothing about `X`.  Also, TypeScript is mostly structurally typed, so (as I understand it) _any_ (object) value could be assigned to a variable of type `IASTNodeLocation` without triggering a type error.

This leads me to wonder: is this type actually needed?  This PR would not be the place to remove it, but perhaps an issue should be filed to consider its removal.

core/interfaces/i_ast_node_location_svg.js
```diff
@@ -0,0 +1,38 @@
...
+ * @author samelh@google.com (Sam El-Husseini)
+ */
+
+'use strict';
+
+goog.provide('Blockly.IASTNodeLocationSvg');
+
+goog.require('Blockly.IASTNodeLocation');
+
+
```
cpcallen, line 19: N.B.: multiple blank lines [allowed but not encouraged](https://google.github.io/styleguide/jsguide.html#formatting-vertical-whitespace).  (Here and a few other places.)

cpcallen (approved):

cpcallen:
Oh: one note on presentation: personally, I'd have preferred that this PR have been organised with one commit to split the file, one to convert all the split files to `goog.module(...)`, and a third for all the named `goog.require(...)`s.

core/interfaces/i_ast_node_location_svg.js
moniika: We use 2 newlines after our requires in all of our files. I would rather address this separately in a different pass.

cpcallen: Oh, OK: I think that's a great convention.  It just stood out in this PR because you don't add two blank lines between before the body of the module when the module has no `require`s, so it looked a little inconsistent.

PR #5098, opened 2021-07-16
Migrate core/comment.js to goog.module syntax

gonfunko:
## The basics

### Resolves

Part of #5026


core/comment.js
```diff
@@ -10,43 +10,44 @@
...
-goog.require('Blockly.browserEvents');
-goog.require('Blockly.Bubble');
-goog.require('Blockly.Css');
-goog.require('Blockly.Events');
+goog.module('Blockly.Comment');
+goog.module.declareLegacyNamespace();
+
+const Block = goog.requireType('Blockly.Block');
+const BlockSvg = goog.requireType('Blockly.BlockSvg');
+const browserEvents = goog.require('Blockly.browserEvents');
```
moniika, line 18: I think this is something that should be destructured.
```
const {conditionalBind, data, unbind} = goog.require('Blockly.browserEvents');
```

Same for `dom`, `userAgent`, and `utilsObject`.

core/comment.js
```diff
@@ -10,43 +10,44 @@
...
+const Bubble = goog.require('Blockly.Bubble');
+const Coordinate = goog.requireType('Blockly.utils.Coordinate');
+const Css = goog.require('Blockly.Css');
```
moniika, line 21: `Blockly.Css.register` is the only method used on this namespace, so this would also be an appropriate candidate for destructuring.
```
const {register} = goog.require('Blockly.Css');
```

core/comment.js
gonfunko: Done!

core/comment.js
gonfunko: Done

core/comment.js
```diff
@@ -10,43 +10,42 @@
...
+const Bubble = goog.require('Blockly.Bubble');
+const Coordinate = goog.requireType('Blockly.utils.Coordinate');
+const Events = goog.require('Blockly.Events');
+const Icon = goog.require('Blockly.Icon');
+const Size = goog.requireType('Blockly.utils.Size');
+const Svg = goog.require('Blockly.utils.Svg');
+const WorkspaceSvg = goog.requireType('Blockly.WorkspaceSvg');
+const {CommentModel} = goog.requireType('Blockly.Block');
+const {conditionalBind, Data, unbind} = goog.require('Blockly.browserEvents');
+const {dom, userAgent, object: utilsObject} = goog.require('Blockly.utils');
```
moniika, line 27: I am not sure that this is allowed because these are all provided in different files.

In any case, my comment about destructuring was not about combining these to the same require, but rather, doing a similar process as when requireing `browserEvents`.

```
const {createSvgElement, HTML_NS} = goog.require('Blockly.utils.dom');
const {IE} = goog.require('Blockly.utils.userAgent');
const {inherits} = goog.require('Blockly.utils.object');
```

core/comment.js
```diff
@@ -10,43 +10,42 @@
...
+const Block = goog.requireType('Blockly.Block');
+const BlockSvg = goog.requireType('Blockly.BlockSvg');
+const Bubble = goog.require('Blockly.Bubble');
+const Coordinate = goog.requireType('Blockly.utils.Coordinate');
+const Events = goog.require('Blockly.Events');
+const Icon = goog.require('Blockly.Icon');
+const Size = goog.requireType('Blockly.utils.Size');
+const Svg = goog.require('Blockly.utils.Svg');
+const WorkspaceSvg = goog.requireType('Blockly.WorkspaceSvg');
+const {CommentModel} = goog.requireType('Blockly.Block');
```
moniika, line 25: This shouldn't be required twice, instead you should use `Block.CommentModel` in the JsDoc reference.

core/comment.js
gonfunko: Ah, got it (I think) - sorry for the confusion and thanks for clarifying!

core/comment.js
gonfunko: Done

core/comment.js
```diff
@@ -10,43 +10,43 @@
...
+const Events = goog.require('Blockly.Events');
+const Icon = goog.require('Blockly.Icon');
+const Size = goog.requireType('Blockly.utils.Size');
+const Svg = goog.require('Blockly.utils.Svg');
+const WorkspaceSvg = goog.requireType('Blockly.WorkspaceSvg');
+const {conditionalBind, Data, unbind} = goog.require('Blockly.browserEvents');
+const {createSvgElement, HTML_NS} = goog.require('Blockly.utils.dom');
+const {inherits} = goog.require('Blockly.utils.object');
+const {register} = goog.require('Blockly.Css');
+const {IE} = goog.require('Blockly.utils.userAgent');
```
moniika, line 29: The most minor of nits, `const {IE} = ...` should be moved up so it is alphabetized.
 I THINK this should be before inherits, or at least before register. I'm not how sure how capitalization affects alphabetization.

moniika (approved):

core/comment.js
gonfunko: Yeah, the styleguide is sort of vague on that point since it assumes they get auto-sorted. I think capital/lowercase are sorted separately, but I realize the capitals should probably come first; any objections to that (so {IE} before {conditionalBind...})?

alschmiedt: I had this same import in a file I was working on and Chris had me require the entire module. [Comment found here](https://github.com/google/blockly/pull/5076#discussion_r671414763).

core/comment.js
```diff
@@ -260,8 +253,8 @@ Blockly.Comment.prototype.setVisible = function(visible) {
  * Show the bubble. Handles deciding if it should be editable or not.
  * @private
  */
-Blockly.Comment.prototype.createBubble_ = function() {
-  if (!this.block_.isEditable() || Blockly.utils.userAgent.IE) {
+Comment.prototype.createBubble_ = function() {
+  if (!this.block_.isEditable() || IE) {
```
cpcallen, line 257: Yes, as @alschmiedt notes above, I think this line is easier to understand if it reads:
```
  if (!this.block_.isEditable() || userAgent.IE) {
```
(Indeed, it could be even more readable if the constant were renamed: `userAgent.isIE`.)

gonfunko: Updated accordingly. Renaming the constant SGTM as well, but is beyond the scope of this PR.

core/comment.js
gonfunko: Since the consensus seems to be requiring the module (which seems a bit more readable to me as well) I went ahead and did that, which also resolves the sort issue.

Issue #5073, opened 2021-07-15
Review structure and style of goog.modules


cpcallen:
## Background

The first pass converting `goog.provide` to `goog.module` is for #5026 is intentionally being done quite mechanically so as to not change the API of any individual module and therefore to allow the work to proceed file-by-file.

In the process, there are going to be situations where it becomes clear that the name and/or structure of the exports of a module are inelegant, no longer well-structured or in violation of the style guide.  Possible examples include:

1. Modules that export a single class constructor, where the class constructor is encumbered with a number properties whose values seem unrelated to the class.  Examples:
   - A class `MyClass` where `MyClass.CONST` was some value that was not an instance of MyClass, or
   - …where `MyClass.someFunction` did not accept or return a value of type `MyClass`.

   Not examples—these are fine:
   - Classes with static methods (like `Array.from` on `Array`).
   - Enums with their corresponding value properties: `MyEnum.value1`, `MyEnum.value2`.

0. Modules named with a capital letter that do not have as their sole export the correspondingly-named class constructor.  Examples: 
   - A `goog.module('….Name')` that has `exports = {id1, id2, id3}`, or
   - …which has `exports = Identifier` where `Identifier` is not a class constructor.

[Add similar situations here.]

Fixing these issues naturally entails modifying the exported API of the module which consequently involves modifying the code which imports the module.  We therefore wish to defer this work until the first pass, file-by-file conversion to `goog.module` is complete.

This bug is to track instances of the above issues so we can come back to them once the first pass is complete.

This bug is _not_ intended to track cases where it is merely the contents of the API which is in need of review—i.e., where we think we wish to add items to (or remove them from) the `exports` list.

## Items in need of further attention

Items should be formatted ```* [ ] `path/to/filename.js` (`Current.module.Name`):``` followed by an explanatory note or link to existing discussion about the problem.

### Needing triage:

Add to this list any files which you discover which you think may need to be reviewed for inclusion in any of the other categories below.  @cpcallen will review and tick completed items off.

* [X] `core/utils/coordinate.js` (`Blockly.utils.Coordinate`): see  #5047.
* [ ] …

### Exports a class constructor with unrelated static properties:

* [ ] …

### Named with a capital letter but does not export a class constructor:

* [ ] `core/blockly.js` (`Blockly`): not a class constructor (at least not yet).  Users can import `const Blockly = goog.require('blockly');` to avoid having to rename throughout existing codebases.
* [ ] …

PR #5102, opened 2021-07-16
Migrate core/connection.js to goog.module syntax

gonfunko:
## The basics

### Resolves

Part of #5026


core/connection.js
```diff
@@ -47,79 +47,79 @@ Blockly.Connection = function(source, type) {
...
  */
-Blockly.Connection.CAN_CONNECT = 0;
-Blockly.Connection.REASON_SELF_CONNECTION = 1;
-Blockly.Connection.REASON_WRONG_TYPE = 2;
-Blockly.Connection.REASON_TARGET_NULL = 3;
-Blockly.Connection.REASON_CHECKS_FAILED = 4;
-Blockly.Connection.REASON_DIFFERENT_WORKSPACES = 5;
-Blockly.Connection.REASON_SHADOW_PARENT = 6;
-Blockly.Connection.REASON_DRAG_CHECKS_FAILED = 7;
+Connection.CAN_CONNECT = 0;
```
rachel-fenichel, line 50: This looks like an instance of "Exports a class constructor with unrelated static properties", which is one of the triage categories in https://github.com/google/blockly/issues/5073

Intuitively, I expected that these would be exported directly.

core/connection.js
```diff
@@ -305,18 +298,18 @@ Blockly.Connection.prototype.connect = function(otherConnection) {
...
 /**
  * Update two connections to target each other.
- * @param {Blockly.Connection} first The first connection to update.
- * @param {Blockly.Connection} second The second connection to update.
+ * @param {Connection} first The first connection to update.
+ * @param {Connection} second The second connection to update.
  * @private
  */
-Blockly.Connection.connectReciprocally_ = function(first, second) {
+Connection.connectReciprocally_ = function(first, second) {
```
rachel-fenichel, line 312: Since this one is private and static, I think it can just become a module-internal function named `connectReciprocally`.

rachel-fenichel: Same for the next few private static functions.

core/connection.js
cpcallen: Yes: not for this PR, but this looks like it should be an enum.  In that form I think it is probably closely-related enough that it could be exported either as a static property of `Connection` (e.g. `Connection.Reason`) or beside it (e.g., `exports = {Connection, ConnectionReason}`).  Added to list in #5073.

gonfunko: Is there any action needed here beyond adding to the triage list?

core/connection.js
gonfunko: Done!

core/connection.js
rachel-fenichel: Nope.

rachel-fenichel (approved):

BeksOmega:
As mentioned in the team meeting, this is one of the files I'm going to be modifying. In this case, to deal with shadows.

PR #5106, opened 2021-07-16
Migrate core/connection_checker.js to goog.module

gonfunko:
## The basics

### Resolves

Part of #5026

### DIFF
--- a/core/connection_checker.js
+++ b/core/connection_checker.js
@@ -64,14 +64,14 @@ Blockly.ConnectionChecker.prototype.canConnect = function(a, b,
  */
 Blockly.ConnectionChecker.prototype.canConnectWithReason = function(
     a, b, isDragging, opt_distance) {
-  var safety = this.doSafetyChecks(a, b);
+  const safety = this.doSafetyChecks(a, b);
   if (safety != Blockly.Connection.CAN_CONNECT) {
     return safety;
   }
 
   // If the safety checks passed, both connections are non-null.
-  var connOne = /** @type {!Blockly.Connection} **/ (a);
-  var connTwo = /** @type {!Blockly.Connection} **/ (b);
+  const connOne = /** @type {!Blockly.Connection} **/ (a);
+  const connTwo = /** @type {!Blockly.Connection} **/ (b);
   if (!this.doTypeChecks(connOne, connTwo)) {
     return Blockly.Connection.REASON_CHECKS_FAILED;
   }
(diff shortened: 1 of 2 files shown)


core/connection_checker.js
```diff
@@ -108,12 +108,13 @@ Blockly.ConnectionChecker.prototype.getErrorMessage = function(errorCode,
       return 'Attempt to connect incompatible types.';
     case Blockly.Connection.REASON_TARGET_NULL:
       return 'Target connection is null.';
-    case Blockly.Connection.REASON_CHECKS_FAILED:
-      var connOne = /** @type {!Blockly.Connection} **/ (a);
-      var connTwo = /** @type {!Blockly.Connection} **/ (b);
-      var msg = 'Connection checks failed. ';
+    case Blockly.Connection.REASON_CHECKS_FAILED: {
```
BeksOmega: Just to clarify, this scope is to get rid of a lint error? Or is it functionally necessary? I see it's used later in this file, but I'm not really familiar with this syntax.

core/connection_checker.js
```diff
@@ -14,19 +14,18 @@
 goog.module('Blockly.ConnectionChecker');
 goog.module.declareLegacyNamespace();
 
-goog.require('Blockly.Connection');
-goog.require('Blockly.connectionTypes');
+const Connection = goog.require('Blockly.Connection');
+const connectionTypes = goog.require('Blockly.connectionTypes');
+const IConnectionChecker = goog.require('Blockly.IConnectionChecker');
+const registry = goog.require('Blockly.registry');
+const RenderedConnection = goog.requireType('Blockly.RenderedConnection');
```
BeksOmega, line 21: I think these are supposed to be alphabetized with capitals first? I know that's how it was in an old style guide, and that matches the example [here](go/js-style#file-goog-require), but it doesn't say explicitly so up to you.

BeksOmega (approved):
LGTM! Just a few style questions.

core/connection_checker.js
gonfunko: Yeah, it's to get rid of a lint error. I believe the motivation is to avoid surprises/allow you to use the same variable name in multiple case statements, since without the braces those are the same scope.

core/connection_checker.js
gonfunko: Yeah I was also a bit unclear on this, updated to reflect that convention.

PR #5107, opened 2021-07-16
Migrate core/renderers/common/info.js to goog.module syntax

rachel-fenichel:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/renderers/common/info.js` to `goog.module` with ES6 `const`/`let`.


core/renderers/common/info.js
```diff
@@ -10,37 +10,37 @@
...
+const BottomRow = goog.require('Blockly.blockRendering.BottomRow');
+const ConstantProvider = goog.requireType('Blockly.blockRendering.ConstantProvider');
+const ExternalValueInput = goog.require('Blockly.blockRendering.ExternalValueInput');
+const Field = goog.require('Blockly.blockRendering.Field');
+const Hat = goog.require('Blockly.blockRendering.Hat');
+const Icon = goog.require('Blockly.blockRendering.Icon');
+const InlineInput = goog.require('Blockly.blockRendering.InlineInput');
+const Input = goog.requireType('Blockly.Input');
+const InputRow = goog.require('Blockly.blockRendering.InputRow');
+const inputTypes = goog.require('Blockly.inputTypes');
```
moniika, line 26: There's only 3 constants from this namespace, we can destructure this require and import them directly.
```
const {DUMMY, INPUT, STATEMENT} = goog.require('Blockly.inputTypes');
```
This would also move the order of this import to the end.

core/renderers/common/info.js
```diff
@@ -10,37 +10,37 @@
...
-goog.require('Blockly.constants');
-goog.require('Blockly.inputTypes');
-
-goog.requireType('Blockly.blockRendering.ConstantProvider');
-goog.requireType('Blockly.blockRendering.Icon');
-goog.requireType('Blockly.blockRendering.Renderer');
-goog.requireType('Blockly.BlockSvg');
-goog.requireType('Blockly.Input');
-goog.requireType('Blockly.RenderedConnection');
+const blocklyConstants = goog.require('Blockly.constants');
```
moniika, line 43: Non-blocking: There's been a lot of discussion on this.

I think it would be ideal if this was:
```
const {ALIGN} = goog.require('Blockly.constants'); 
```
For the usage in this file.

However, as discussion offline, this file has side effects so I'm ok with it being left as is and changed later when we deal with the `core/constants.js` file.

## Alternative approach
I'm not sure if this is valid, but maybe we could do something like:
```
const {ALIGN} = goog.require('Blockly.constants');
// ...
```

That way we ensure we ensure we don't forget to do the destructuring and still ensure we have the extra require.

core/renderers/common/info.js
rachel-fenichel: Done.

core/renderers/common/info.js
rachel-fenichel: Moved to the first version.

rachel-fenichel:
I talked to Chris about constants in general. I'm going to tackle that file next, which includes:
- Moving constants.js to explicit exports
- Moving all side-effect assignments onto an additional namespace (currently thinking `internalConstants`) that we can take apart later
  - We do not want to export these more generally, so I hope to finish the renaming this quarter, but I will file a bug about it in either case
- Adding aliases in Blockly.js for all constants that were set as side effects in `constants.js`
- Updating every place that requires `constants.js` for its side effects, to make it require the new file directly.
  - I think this can be done in separate PRs as we switch to named imports

core/renderers/common/info.js
```diff
@@ -10,37 +10,37 @@
...
+const Renderer = goog.requireType('Blockly.blockRendering.Renderer');
+const RoundCorner = goog.require('Blockly.blockRendering.RoundCorner');
+const Row = goog.require('Blockly.blockRendering.Row');
+const SpacerRow = goog.require('Blockly.blockRendering.SpacerRow');
+const SquareCorner = goog.require('Blockly.blockRendering.SquareCorner');
+const StatementInput = goog.require('Blockly.blockRendering.StatementInput');
+const TopRow = goog.require('Blockly.blockRendering.TopRow');
+const Types = goog.require('Blockly.blockRendering.Types');
+const {ALIGN} = goog.require('Blockly.constants');
+const {VALUE, STATEMENT, DUMMY} = goog.require('Blockly.inputTypes');
```
moniika, line 42: These should be alphabetized `{DUMMY, STATEMENT, VALUE}`

moniika (approved):
One minor comment, otherwise lgtm

PR #5122, opened 2021-07-19
Migrate core/interfaces/i_toolbox_item.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/interfaces/i_toolbox_item.js` to `goog.module` syntax. 

### Additional Information

This PR splits up `core/interfaces/i_toolbox_item.js` into multiple files for each interface that it provided.

### DIFF
--- a/core/interfaces/i_collapsible_toolbox_item.js
+++ b/core/interfaces/i_collapsible_toolbox_item.js
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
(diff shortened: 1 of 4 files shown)


cpcallen (approved):
LGTM.

My only comment is that it would have been good to run `build:deps` after dividing the file and included the result in the first commit (and then run it again for the `goog.module` conversion in the second PR, as you did do).

Since the divided file only contains interface declarations, failing to have done so probably doesn't cause tests to fail (or playground not to load), but had the file contained executable code things would have been broken—and even though the next commit would have fixed it,  I think it's preferable to avoid this when possible.

Issue #5124, opened 2021-07-19
IRegistrableField uses interface naming convention but is not an interface


moniika:
**Describe the bug**

`IRegistrableField` uses interface naming convention but is not an interface.

**Expected behavior**

Appending "I" to the start of a classname should be reserved for interfaces.
`IRegistrableField` should be renamed (or be an interface or record).

**Additional context**

gonfunko:
IRegistrableField is no more as of #6603.

PR #5060, opened 2021-07-14
Migrate core/utils/style.js to goog.module syntax

gonfunko:
## The basics

- [x] I branched from `goog_module`
- [x] My pull request is against `goog_module`
- [x] My code follows the [style guide](
      https://developers.google.com/blockly/guides/modify/web/style-guide)
- [x] My code is presented in the form suggested in the [module
      conversion guide](https://github.com/google/blockly/issues/5026)
- [x] I have run `npm test` locally already.

### Resolves

Part of #5026

PR #5125, opened 2021-07-19
Migrate core/interfaces/i_registrable_field.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/interfaces/i_registrable_field.js` to `goog.module` syntax. 

### Additional Information

https://github.com/google/blockly/issues/5124

### DIFF
--- a/core/interfaces/i_registrable_field.js
+++ b/core/interfaces/i_registrable_field.js
@@ -11,21 +11,25 @@
 
 'use strict';
 
-goog.provide('Blockly.IRegistrableField');
+goog.module('Blockly.IRegistrableField');
+goog.module.declareLegacyNamespace();
 
 goog.requireType('Blockly.Field');
+
 
 /**
  * A registrable field.
  * Note: We are not using an interface here as we are interested in defining the
  * static methods of a field rather than the instance methods.
  * @typedef {{
- *     fromJson:Blockly.IRegistrableField.fromJson
+ *     fromJson:IRegistrableField.fromJson
  * }}
  */
-Blockly.IRegistrableField;
+let IRegistrableField;
 
 /**
  * @typedef {function(!Object): Blockly.Field}
  */
-Blockly.IRegistrableField.fromJson;
+IRegistrableField.fromJson;
+
+exports = IRegistrableField;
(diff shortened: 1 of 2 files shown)


core/interfaces/i_registrable_field.js
```diff
@@ -11,21 +11,25 @@
 
 'use strict';
 
-goog.provide('Blockly.IRegistrableField');
+goog.module('Blockly.IRegistrableField');
+goog.module.declareLegacyNamespace();
 
 goog.requireType('Blockly.Field');
 
+
```
cpcallen, line 19: Whoop whoop!

core/interfaces/i_registrable_field.js
```diff
@@ -11,21 +11,25 @@
...
-Blockly.IRegistrableField;
+let IRegistrableField;
 
 /**
  * @typedef {function(!Object): Blockly.Field}
  */
-Blockly.IRegistrableField.fromJson;
+IRegistrableField.fromJson;
+
+exports = IRegistrableField;
```
cpcallen, line 35: I see in #5060 and #5079 @gonfunko has omitted the blank line between the definition and the export.  I like that because it makes clear the export is directly related to the definition above.

core/interfaces/i_registrable_field.js
moniika: I added the newline here because all but 3 of our files use 2 newlines after our requires except for:
- `core/interfaces/i_registrable_field.js`
- `core/delete_area.js`
- `core/input_types.js`

For that reason I think it makes sense to add the newline here for consistency.

core/interfaces/i_registrable_field.js
moniika: Those files differ in that they aren't exporting a single class.

I agree that it makes sense to put the export directly below the function definition, but not in the case of a class.
I think that the newline here makes it easier to read and removing it would make it seem like it is directly related to the function definition for `IRegistrableField.fromJson;`

core/interfaces/i_registrable_field.js
cpcallen: Yeah, nice.

core/interfaces/i_registrable_field.js
cpcallen: Yes, I see that now, and I agree with you.

Issue #4992, opened 2021-07-05
Values of fields defined in the toolbox should use the message localization system


ndahn:
When defining comments or default values for fields in the toolbox definition, e.g. for example programs, these strings will usually be localized to the user's language. However, AFAIK the values of fields cannot be localized. At the very least a description of how this can be done is missing from the toolbox documentation at [https://developers.google.com/blockly/guides/configure/web/toolbox](https://developers.google.com/blockly/guides/configure/web/toolbox).

More specifically, the following was expected to work, but does not:
```xml
<block type="myblock">
    <field name="FIELD_1">%{BKY_MY_VALUE}</field>
</block>
```
This will display `%{BKY_MY_VALUE}` verbatim and not the value defined for `MY_VALUE` as defined in the loaded localization.

**Preferred solution**
To me the ideal solution would be if the above syntax worked, passing the field's value through `Blockly.utils.replaceMessageReferences`. 
Alternatively the field could also accept a `value="%{BKY_MY_VALUE}"` attribute, although this begs the question whether it should override the value between the `<field>` tags or be overridden by it.

BeksOmega:
Hi @ndahn so I brought up your [proposal](https://github.com/google/blockly/pull/5126#issuecomment-883564468) with the team, and one of them mentioned that if all you want to do is localize field values for your custom blocks, you can do that using [JSON block definitions](https://developers.google.com/blockly/guides/create-custom-blocks/localize-blocks). As you can see [here](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/field_textinput.js#L112), when fields are created via JSON they do dereference the messages.

But that only solves the problem of fields, and only if you only want to localize values in your custom blocks. Given that, do you think this could work for you?

(sorry for jumping back to your issue rather than your PR, but I want future peeps to be able to find this if it does work for you)

ndahn:
Hey @BeksOmega! Unfortunately field values are not enough, I also need to localize some comments. However, since I'm in full control of my project and aware of the limitations, monkey patching or your suggestion from #5126* work fine for me.

*) convert XML to text, localize text; will result in invalid XML with special symbols like "©" (should be "&copy;")

Should we close this and continue on the PR?

BeksOmega:
> Should we close this and continue on the PR?

I think we can leave this open for now, but definitely continue on the PR :D

BeksOmega:
Closing this because #5238 has more info.

PR #5126, opened 2021-07-19
Localization for toolbox XML

ndahn:
## The basics

- [x] I branched from develop

### Resolves

[#4492](https://github.com/google/blockly/issues/4992)

### Proposed Changes

Allows to use the %{BKY_*} message reference system for field values when loading a toolbox XML. The second (actually first) commit also allows localization for block comments, variable names and block data.

#### Behavior Before Change

Field values, comments, variable names and block data could not be localized.

#### Behavior After Change

Field values, comments, variable names and block data can now be localized.

### Reason for Changes

When adding example programs to the toolbox these may often contain field values, comments, variable names, etc. to make the program easier to follow. It would be at least odd if these were not localized in a multi-lingual application.

### Test Coverage

* Desktop Firefox

* Smartphone/Tablet/Chromebook (please complete the following information):
  * Device: Samsung Galaxy M31
  * OS: Android 11
  * Browser Chrome / Firefox
  * Version 91.0.4472.164 / 90.1.1

### Documentation

No additional documentation is required as this was (to me) expected behavior and is kind of covered in the localization guide.

### Additional Information

That's it :)

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -400,7 +400,8 @@ Blockly.Field.prototype.bindEvents_ = function() {
  * @package
  */
 Blockly.Field.prototype.fromXml = function(fieldElement) {
-  this.setValue(fieldElement.textContent);
+  let value = Blockly.utils.replaceMessageReferences(fieldElement.textContent);
+  this.setValue(value);
 };
 
 /**
--- a/core/xml.js
+++ b/core/xml.js
@@ -617,7 +617,7 @@ Blockly.Xml.domToVariables = function(xmlVariables, workspace) {
     }
     var type = xmlChild.getAttribute('type');
     var id = xmlChild.getAttribute('id');
-    var name = xmlChild.textContent;
+    var name = Blockly.utils.replaceMessageReferences(xmlChild.textContent);
 
     workspace.createVariable(name, type, id);
   }
@@ -720,7 +720,7 @@ Blockly.Xml.applyMutationTagNodes_ = function(xmlChildren, block) {
  */
 Blockly.Xml.applyCommentTagNodes_ = function(xmlChildren, block) {
   for (var i = 0, xmlChild; (xmlChild = xmlChildren[i]); i++) {
-    var text = xmlChild.textContent;
+    var text = Blockly.utils.replaceMessageReferences(xmlChild.textContent);
     var pinned = xmlChild.getAttribute('pinned') == 'true';
     var width = parseInt(xmlChild.getAttribute('w'), 10);
     var height = parseInt(xmlChild.getAttribute('h'), 10);
@@ -747,7 +747,7 @@ Blockly.Xml.applyCommentTagNodes_ = function(xmlChildren, block) {
  */
 Blockly.Xml.applyDataTagNodes_ = function(xmlChildren, block) {
   for (var i = 0, xmlChild; (xmlChild = xmlChildren[i]); i++) {
-    block.data = xmlChild.textContent;
+    block.data = Blockly.utils.replaceMessageReferences(xmlChild.textContent);
   }
 };
 


ndahn:
@googlebot I signed it!

BeksOmega:
Hi @ndahn Thanks for the PR =) The code for achieving this looks good, but I'm not sure if this is a behavior we want. Sorry for not commenting on your issue :/ I didn't see it until now.

I think that generally, a serialization system should contain minimal business logic. If it's a really high-stakes system (like serialization) you want to keep it pretty simple/dummy proof so that you don't introduce bugs. Or at least that's my feeling, totally open to other opinions.

Do you think that converting your toolbox to text and passing that through `replaceMessageReferences` before deserializing it could work as an alternative? If that doesn't work for you, do you think there's something else Blockly could provide to make this easier?

Thank you again for taking the time to put up a PR! I do really appreciate it because I know the XML system can be a rats nest to dig through. I'm just hesitant about adding to that rats nest hehe.

ndahn:
Hey @BeksOmega, thanks for taking the time to reply in so much detail! And don't worry, I like rats 🐀 :)

I agree with the general feeling that doing message localization in the toolbox XML is a bit odd when everything else happens in the block definitions. On the other hand there is no other place for doing that when loading a toolbox where values should be localized on a per-block basis.

My fix for now is to monkey patch the relevant functions, but that's not a good general solution since I will have to check if the functions changed on every update. A viable alternative could be to add a flag to `Blockly.inject`, e.g. localizeToolbox, which does the conversion you mentioned.

cpcallen:
Hi @ndahn,

Thanks for your PR.  I'll let @BeksOmega and the rest of the team do the reviewing, but in the mean time: would you be so kind as to edit the title of the PR so that people browsing the list of PRs will be able to know at a glance what changes you are proposing to make?

Thanks!

ndahn:
Sure @cpcallen! I don't even know where that came from 🤔

BeksOmega:
> Hey @BeksOmega, thanks for taking the time to reply in so much detail! And don't worry, I like rats 🐀 :)

Yeah absolutely! Thank you for taking the time to reply back :D

> My fix for now is to monkey patch the relevant functions, but that's not a good general solution since I will have to check if the functions changed on every update.

Totally feel you, monkey patching is the worst to maintain, especially when you're dealing with internal functions.

> A viable alternative could be to add a flag to `Blockly.inject`, e.g. localizeToolbox, which does the conversion you mentioned.

Personally, I would be much more comfortable doing something like that. I think the function for conversion would be something like this:
```
function replaceMessageReferencesInToolbox(toolboxXml) {
  var toolboxText = Blockly.Xml.domToText(toolboxXml);
  var newText = Blockly.utils.replaceMessageReferences(toolboxText);
  return Blockly.Xml.textToDom(newText);
}
```
I did a bit of testing, and this seems to work well. But I'm not super familiar with how the `replaceMessageReferences` algorithm works, and if there might be edge cases. Do you think you could test this with your toolbox to see if there are any references it's not replacing?

If not that't totally cool, it just seemed like since you've got a real use case it might also be a good test case hehe. Thank you again for your reply! This is definitely a use-case I want to solve, and discussing it with you is really helpful =)

ndahn:
@BeksOmega That solution works great for most cases! 

However, I found a pitfall in that special characters like "&" will result in ill-formed XML (e.g. "&" should be "&amp;") and thus results in a parsing error. Although `Blockly.inject` accepts a toolbox string, it internally converts it back to XML, so the same will happen. I see two solutions to this:

- Make users aware that for toolbox localization they have to use SGML entity names. Not very user friendly, but simple.
- Use a library like [https://github.com/mathiasbynens/he](he).encode to encode the string before turning it into XML. Nice and clean, but introduces a new dependency.

What do you think?

ndahn:
> Darn that sucks. I was really hoping the code posted above would work out-of-the-box lol.

Yeah, it was close... The chances of me noticing were actually quite slim, there's only a single random "&". Sorry about that :)

> Wow that looks like a sweet library, but I'm not sure how the team rest of the team feels about external libraries. I'm going to bring it up on our internal chat and report back. If we decide to go with this we could probably add a flag to [tokenizeInterpolation](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/utils.js#L276) that deals with escaping the bad chars?

I feel like this belongs more to the DOM/XML system, possibly only element creation, but I agree that `Blockly.inject` is definitely too high level to be the right place. Thanks for checking!

> The other issue I want to mention is that this runs up against that fact that our i18n system is not... great. It works perfectly for translating the core blocks b/c of our integration with translatewiki, but it's not great for external developers or for plugins. I still want to try to get a fix in for this, but it's something to keep in mind in case this whole system gets refactored later.

Haha, can't argue with that :) I was trying before to bundle all my localization in one place but couldn't quite get Blockly to cooperate. It would be nice if there was a clean way of injecting functions here and there into the flow.

BeksOmega:
It turns out this is a bit of a tricky problem that's going to need more internal discussion :/ I've added notes on this to the team-meeting for next week. If there's anything in particular you'd like to highlight about the options I'd love to copy that over. And either way I'll keep you in the loop about what we discuss!

Thank you for bearing with me! I really appreciate you going back and forth with me on this.

ndahn:
> It turns out this is a bit of a tricky problem that's going to need more internal discussion :/ I've added notes on this to the team-meeting for next week. If there's anything in particular you'd like to highlight about the options I'd love to copy that over. And either way I'll keep you in the loop about what we discuss!
> 
> Thank you for bearing with me! I really appreciate you going back and forth with me on this.

Of course! Thank you for taking so much time to discuss possible solutions and keeping me in the loop, that's quite something :)

> Yeah, it does feel weird to add the flag to `tokenizeInterpolation` when you might not be dealing with XML. I'm really what the best place to put this is :/

Hmm.... How about [utils.xml.createTextNode](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/utils/xml.js#L55)? This might well be the only place where it matters. 
I actually just checked again (and a bit more thoroughly). There's a bunch of stuff that will cause issues:
- A lonely `&` causes issues,
- As well as any string of the form `&[\w\d]*;` that doesn't match a valid SGML entity name,
- Also as any `<` (but not `>`).
- `©` is fine because it has no special meaning for XML.
- etc.

So the library mentioned above doesn't solve this issue. I'm very sorry about that :/ 
This now feels like adding a check and an exception would be a better option than trying to cover every corner case...

> Ah gosh that sucks, I'm sorry. Blockly's i18n system definitely needs some TLC haha. Could I ask what other systems you're working with? Maybe we can ~steal~ borrow some of their ideas :D

Sneaky :D This is my first time doing localization, but I settled on [jQuery.i18n](https://github.com/wikimedia/jquery.i18n). I tried to monkey patch Blockly to use that function for looking up strings instead of its own system, but couldn't quite get it to work. In the end I just accepted having translations in several files.

BeksOmega:
Heyo @ndahn After two meetings, we decided to stick with what we've got now, and work on making this better once our schedule is more open (probably not for a good while :/ ). We want to do more research into how other people do localization, and create a better plan for plugin localization before we start reworking everything.

Thank you for bringing this up and all of your help with my questions! If there's anything I can do for you in return just ask :D

I filed a [new tracking issue](https://github.com/google/blockly/issues/5238) which includes a bunch of the stuff we discussed internally, as well as a bunch of info from you. So now it's all in one place. I'm going to go ahead and close this, and your related issue, because the other one has more info.

Thank you again! --Beka

ndahn:
Hey @BeksOmega! I don't quite agree with the decision since there's a perfectly fine PR, but I'll relent :) Thank you again for getting back and giving all these updates (and also for the very concise tracking issue). Hope you'll find a solution soon!

All the best --Nikolas

PR #5131, opened 2021-07-20
feat: add JSO serialization and deserialization of variables

BeksOmega:
## The basics

- [X] I branched from **project-cereal**
- [X] My pull request is against **project-cereal**
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/connnected-blocks...BeksOmega:cereal/variables

### Resolves

Work on project cereal
Dependent on #5120 

### Proposed Changes

Adds serialization and deserialization of variables as JSOs.

### Reason for Changes

Saving state is good.

### Test Coverage

Added tests to ensure that serialized variables have the correct format.

PR #5132, opened 2021-07-20
feat: add top level `save` and `load` functions for JSO serialization

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/variables...BeksOmega:cereal/top-level

### Resolves

Work on project cereal
Dependent on #5131 

### Proposed Changes

Adds the top-level `save` and `load` functions to the `Blockly.serialization` namespace.

### Reason for Changes

Saving state is good. Also this allows us to test round-tripping.

### Test Coverage

Added using the JSO hooks in the serializer tests. Currently they are all skipped because deserialization hasn't been added to the blocks system.


core/serialization/serialization.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+};
+ 
+exports = {
+  blocks,
+  load,
+  save,
+  variables,
+};
```
BeksOmega, line 87: Does this look correct?

BeksOmega: I think this should not export variables, because we want the save and load for variables to remain internal.

alschmiedt: The part I am not sure about is having this file export items from a separate module. This is something we do in blockly samples (index.js usually exports everything), but I am not sure how this fits into the `goog.module` conversion in core. @rachel-fenichel any thoughts?

tests/mocha/serializer_test.js
```diff
@@ -0,0 +1,1758 @@
...
+    });
+
+    testHelpers.runTestSuites(
+        testSuite.testSuites, createTestCaseFunction);
+    testHelpers.runTestCases(testSuite.testCases, createTestFunction);
+  });
+};
+
+runSerializerTestSuite(null, null, Serializer);
+Serializer.skip = true;
```
alschmiedt, line 1757: Just curious why we are skipping here?

alschmiedt (approved):

tests/mocha/serializer_test.js
BeksOmega: Skipping because this work was done before deserialization of blocks was added. That happens in the next PR & this gets unskipped :D

core/serialization/serialization.js
rachel-fenichel: Two things here:
1) Let's avoid creating non-leaf namespaces going forward. In this case, that's `Blockly.serialization` given that you are also creating `Blockly.serialization.variables`, etc.
2) What is the purpose of re-exporting these functions, given that you can require the serialization code and access it from anywhere in core?

BeksOmega: > Let's avoid creating non-leaf namespaces going forward. In this case, that's Blockly.serialization given that you are also creating Blockly.serialization.variables, etc.

Cool, we can totally change this around since it isn't published yet. If I'm reading this right, this relates to [3.3.1](https://google.github.io/styleguide/jsguide.html#naming-hierarchy) in the styleguide? So instead of `Blockly.serialization` and `Blockly.serialization.variables`, we would have `Blockly.serialization.<something>` and `Blockly.serialization.variables`? I'm thinking `Blockly.serialization.workspace`.

> What is the purpose of re-exporting these functions, given that you can require the serialization code and access it from anywhere in core?

It wasn't letting me access the nested namespaces (eg `Blockly.serialization.variables`) from inside tests. But I'm thinking now that I was probably requiring them in requires.js incorrectly. I'll investigate that when I switch over the top-level namespace.

PR #5140, opened 2021-07-20
Make constants.js have no side effects

rachel-fenichel:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

- Converts `core/constants.js` to a module with es6 const/let.
- Removes all side-effect assignments from `constants.js` (any assignments to properties on the `Blockly` object).
- Moves previously side-effect assigned properties onto a new namespace, `Blockly.internalConstants`.
- Adds aliases in `blockly.js` for all of these properties.
- Updates all references in the core library to require `Blockly.internalConstants` and use properties exported there.
- Does *not* update anything in `blocks/` or `generators/`, as those are outside of core.
- Removes all lines of `goog.require('Blockly.constants')` that were tagged with `@suppress extraRequire`

### Additional Information

Follow-up work includes:
- Investigate each of the properties (now in the list of aliases at the bottom of `blockly.js`) to decide which ones are not actually public, if any.
- Add deprecation warnings to any properties that are public but that we will remove in an upcoming release.
- Add setters for any properties that we will keep visible past this quarter.


core/generator.js
```diff
@@ -14,8 +14,7 @@
 goog.provide('Blockly.Generator');
 
 goog.require('Blockly.Block');
-/** @suppress {extraRequire} */
-goog.require('Blockly.constants');
+goog.require('Blockly.internalConstants');
```
cpcallen, line 17: Require ordering.

core/gesture.js
```diff
@@ -18,11 +18,10 @@ goog.require('Blockly.blockAnimations');
 goog.require('Blockly.BlockDragger');
 goog.require('Blockly.browserEvents');
 goog.require('Blockly.BubbleDragger');
-/** @suppress {extraRequire} */
-goog.require('Blockly.constants');
 goog.require('Blockly.Events');
 /** @suppress {extraRequire} */
 goog.require('Blockly.Events.Click');
+goog.require('Blockly.internalConstants');
```
cpcallen, line 24: Require ordering.

cpcallen (approved):
LGTM, except:

* Some of these files now definitely have [invalid `goog.requires` ordering](https://google.github.io/styleguide/jsguide.html#file-goog-require).  I've noted a couple of these below, but there are several more.  Do you want to fix them in this PR?  An open issue to fix this problem globally would be an acceptable alternative.
* Have you run `clang-format` on the affected files (at least `core/constants.js` and `core/internal_constants.js`)?

I've filed #5158 to track breaking up and removing `core/internal_constants.js` in due course.

Aside: I found this PR a bit difficult to review.  Part of this was just that it is a bit larger than the usual migrations, but it didn't help that some of the intermediate commits left it in an obviously-broken state.  I think I'd have preferred if it had been presented with the first commit creating `internal_constants.js` and doing _all_ of the associated changes to other files (including updating `deps.js`) without changing line breaks / indentation, followed by the usual sequence of commits for updating `constants.js` to `goog.module` and named requires, then a final clang-format of everything.

rachel-fenichel:
> Aside: I found this PR a bit difficult to review. Part of this was just that it is a bit larger than the usual migrations, but it didn't help that some of the intermediate commits left it in an obviously-broken state. I think I'd have preferred if it had been presented with the first commit creating internal_constants.js and doing all of the associated changes to other files (including updating deps.js) without changing line breaks / indentation, followed by the usual sequence of commits for updating constants.js to goog.module and named requires, then a final clang-format of everything.

I have clang-format set to run on changed lines, which is something we as a team were doing before this big migration, but it sounds like that made things more confusing here. Since we're applying formatting to all files as we go, maybe I should turn it off for per-line changes and only run it in bulk on each file.

I did *not* do a bulk clang-format of all of the files i touched--I left that to be done on a per-file basis as we do updates.

core/internal_constants.js
```diff
@@ -0,0 +1,217 @@
...
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Module that provides constants for use inside Blockly. Do not
+ * use these constants outside of the core library.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

rachel-fenichel (commented):
I rebased and made requested changes. I did not try to fix require orders in any file that had not yet been converted to named requires and did not clang-format those files, since that will all be done in per-file PRs.

I did run clang-format on `constants.js` and `internal_constants.js`.

cpcallen:
> I have clang-format set to run on changed lines, which is something we as a team were doing before this big migration, but it sounds like that made things more confusing here.

Ordinarily I think that's a great idea (I have emacs configured to `gofmt` ay time is saves a `.go` file)—but yes: where the file is going to be `clang-format`ed in a later step I find it easier to review the non-formatting changes separately.

> I did _not_ do a bulk clang-format of all of the files i touched--I left that to be done on a per-file basis as we do updates.

Yes, sensible enough (and a good cause for me to pipe down about the format changes that did get made in these files, as I'd otherwise be complaining about them being left with overly-long lines).  Can you confirm that what is left of `constants.js` is correctly formatted, since that is now migrated (and you didn't touch every line of it)?

cpcallen:
(Oh, sorry: overnight my browser window had updated with your first reply but not the second one, and I forgot to hit reload.  Thanks for confirming that the two migrated files are formatted!)

PR #5141, opened 2021-07-20
Migrate core/keyboard_nav/ast_node.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/keyboard_nav/ast_node.js` to `goog.module` syntax.

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -13,13 +13,20 @@
 goog.module('Blockly.ASTNode');
 goog.module.declareLegacyNamespace();
 
+/* eslint-disable-next-line no-unused-vars */
 const Block = goog.requireType('Blockly.Block');
+/* eslint-disable-next-line no-unused-vars */
 const Connection = goog.requireType('Blockly.Connection');
 const Coordinate = goog.require('Blockly.utils.Coordinate');
+/* eslint-disable-next-line no-unused-vars */
 const Field = goog.requireType('Blockly.Field');
+/* eslint-disable-next-line no-unused-vars */
 const IASTNodeLocation = goog.requireType('Blockly.IASTNodeLocation');
+/* eslint-disable-next-line no-unused-vars */
 const IASTNodeLocationWithBlock = goog.requireType('Blockly.IASTNodeLocationWithBlock');
+/* eslint-disable-next-line no-unused-vars */
 const Input = goog.requireType('Blockly.Input');
+/* eslint-disable-next-line no-unused-vars */
 const Workspace = goog.requireType('Blockly.Workspace');
 const connectionTypes = goog.require('Blockly.connectionTypes');
 
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -150,8 +150,8 @@ ASTNode.createConnectionNode = function(connection) {
   const type = connection.type;
   if (type == connectionTypes.INPUT_VALUE) {
     return ASTNode.createInputNode(connection.getParentInput());
-  } else if (type == connectionTypes.NEXT_STATEMENT &&
-      connection.getParentInput()) {
+  } else if (
+      type == connectionTypes.NEXT_STATEMENT && connection.getParentInput()) {
     return ASTNode.createInputNode(connection.getParentInput());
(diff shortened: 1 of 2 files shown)


core/keyboard_nav/ast_node.js
```diff
@@ -10,38 +10,38 @@
...
-goog.require('Blockly.utils.Coordinate');
-
-goog.requireType('Blockly.Block');
-goog.requireType('Blockly.Connection');
-goog.requireType('Blockly.Field');
-goog.requireType('Blockly.IASTNodeLocation');
-goog.requireType('Blockly.IASTNodeLocationWithBlock');
-goog.requireType('Blockly.Input');
-goog.requireType('Blockly.Workspace');
+const Block = goog.requireType('Blockly.Block');
```
maribethb, line 16: I think for now you need to add the line disabling the lint warning (sorry, I'm working on it)

core/keyboard_nav/ast_node.js
```diff
@@ -412,29 +415,30 @@ Blockly.ASTNode.prototype.findPrevForField_ = function() {
...
   if (!curLocation || !curLocation.workspace) {
     return null;
   }
-  var curRoot = curLocation.getRootBlock();
-  var topBlocks = curRoot.workspace.getTopBlocks(true);
-  for (var i = 0, topBlock; (topBlock = topBlocks[i]); i++) {
+  const curRoot = curLocation.getRootBlock();
+  const topBlocks = curRoot.workspace.getTopBlocks(true);
+  let i = 0, topBlock;
+  for (; (topBlock = topBlocks[i]); i++) {
```
maribethb, line 434: Is this what it's supposed to be? I know this technically works, but this is really weird looking to me though and I thought `for (let i = 0, topBlock; topBlock = topBlocks[i]; i++)` was supposed to work? (there are several like this in this file)

core/keyboard_nav/ast_node.js
```diff
@@ -444,125 +448,128 @@ Blockly.ASTNode.prototype.navigateBetweenStacks_ = function(forward) {
...
 
 /**
  * Find the element to the right of the current element in the AST.
- * @return {Blockly.ASTNode} An AST node that wraps the next field, connection,
+ * @return {ASTNode} An AST node that wraps the next field, connection,
  *     block, or workspace. Or null if there is no node to the right.
  */
-Blockly.ASTNode.prototype.next = function() {
+ASTNode.prototype.next = function() {
+  let connection, block, nextConnection, targetConnection;
```
maribethb, line 545: I think declaring multiple variables in one line is disallowed: https://google.github.io/styleguide/jsguide.html#features-one-variable-per-declaration (which might conflict with the array advice..... in which case I would probably convert those arrays to more normal style of arrays `for (let i = 0, i < a.length; i++) {`)

core/keyboard_nav/ast_node.js
moniika: This is what my IDE automatically updated. I'll look into it.

core/keyboard_nav/ast_node.js
moniika: Noted. I try something else that follows the style guide. I was getting a different warning about declaring variables in a case block before, which is why i put them all up here.

core/keyboard_nav/ast_node.js
cpcallen: The IDE is probably worried that `i` or `topBlock` might be used after the end of the loop body—which is possible with a `var` (function scoped), but not with `let` (scoped only to the loop).

A bit disappointing it doesn't actually check, though.

core/keyboard_nav/ast_node.js
cpcallen: If it helps, you can (and [are allowed](https://google.github.io/styleguide/jsguide.html#formatting-switch-statements)) to put a block in each `case`:
```JS
switch (x) {
  case y: {
    // body;
    break;  // or return;
  }
  // etc.
}
```
…though admittedly it does look a little weird.

moniika: I've added blocks to each `case` that require lexical scoping, but I'm not sure if I should add it to all cases if some of them need it.

For now I've only added it to those that need it, but I could add to the rest.

maribethb (approved):

PR #5143, opened 2021-07-20
Fix edge cases for convert-script

moniika:
Fixes bugs for the following edge cases:
* No missing requires -  corrects logic for detecting missing requires by checking for non-empty string instead of counting lines for the regex match return
* Checks for overlapping requires, such as if `Blockly.utils` and `Blockly.utils.dom` are both required
* Checks for overlap with module name and require, such as if `Blockly` is a require

### DIFF
--- a/scripts/goog_module/convert-file.sh
+++ b/scripts/goog_module/convert-file.sh
@@ -141,7 +141,6 @@ step2 () {
 
   npm run build:deps
   success "Completed automation for step 3. Please manually review and add exports for non-private top-level functions."
-
 }
 #######################################
 # Runs step 3 of the automated conversion.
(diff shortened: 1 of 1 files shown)


scripts/goog_module/convert-file.sh
```diff
@@ -201,12 +215,11 @@ step3() {
...
   local missing_requires=$(perl -nle'print $& while m{(?<!'\'')Blockly(\.\w+)+}g' "${filepath}")
-  local missing_require_lines=$(echo "${missing_requires}" | wc -l)
-  if [[ "${missing_require_lines}" -gt "0" ]]; then
+  if [[ ! -z "${missing_require_lines}" ]]; then
     err "Missing requires for: ${missing_requires} Please manually fix."
   fi
 
-  inf "Add missing nullability modifier..."
+  inf "Add missing nullability modifiers..."
   perl -pi -e 's/@(param|return) \{([A-Z])/@\1 \{?\2/g' "${filepath}"
```
maribethb, line 223: I think this might not work because there are a lot of `Function` types which are considered primitives and don't need a modifier. And we decided not to do it as we go anyway so you could probably take this part out rather than solve it. I've noticed some cases where there is no modifier so by default it's nullable but logically they should probably not be nullable. Those cases will be easier to notice if we haven't mechanically added the `?` without checking each case by hand

maribethb (approved):
LGTM with the caveat that I am underqualified to review perl :P

scripts/goog_module/convert-file.sh
moniika: I can remove it.

PR #5149, opened 2021-07-21
Migrate core/contextmenu_registry.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/contextmenu_registry.js` to `goog.module` with ES6 `const`/`let`.


core/contextmenu_registry.js
```diff
@@ -160,4 +169,6 @@ Blockly.ContextMenuRegistry.prototype.getContextMenuOptions = function(scopeType
 };
 
 // Creates and assigns the singleton instance.
-new Blockly.ContextMenuRegistry();
+new ContextMenuRegistry();
```
rachel-fenichel, line 172: @maribethb given that we only access this as `ContextMenuRegistry.registry` (and the comment says not to call the constructor directly), should the exports be just an object with the named `registry` property instead?

maribethb (approved):

core/contextmenu_registry.js
maribethb: We still have to export the type for use elsewhere and so the other class methods can be used / understood by the compiler. I talked to Chris and he suggested exporting it the normal way (so, this) but adding a `@private` annotation to the constructor, assuming that passes compilation.

gonfunko: Updated to make the constructor private, which seems to work fine.

PR #5157, opened 2021-07-21
Remove need for require instead of requireType for Block in ASTNode

moniika:
Replacing check `instance of Blockly.Block` in order to fix circular dependency between `Blockly.Block` and `Blockly.ASTNode`.

### DIFF
--- a/core/keyboard_nav/ast_node.js
+++ b/core/keyboard_nav/ast_node.js
@@ -420,7 +420,7 @@ Blockly.ASTNode.prototype.findPrevForField_ = function() {
  */
 Blockly.ASTNode.prototype.navigateBetweenStacks_ = function(forward) {
   var curLocation = this.getLocation();
-  if (!(curLocation instanceof Blockly.Block)) {
+  if (curLocation.getSourceBlock) {
     curLocation = /** @type {!Blockly.IASTNodeLocationWithBlock} */ (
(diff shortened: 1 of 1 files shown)


core/keyboard_nav/ast_node.js
```diff
@@ -420,7 +420,7 @@ Blockly.ASTNode.prototype.findPrevForField_ = function() {
  */
 Blockly.ASTNode.prototype.navigateBetweenStacks_ = function(forward) {
   var curLocation = this.getLocation();
-  if (!(curLocation instanceof Blockly.Block)) {
+  if (curLocation.getSourceBlock) {
```
alschmiedt, line 423: I think the best way to fix this would be below, but if you are going for simplest possible change right now then this works. 
```
  var sourceBlock = this.getSourceBlock();
  if (!sourceBlock || !sourceBlock.workspace) {
    return null;
  }
  var curRoot = sourceBlock.getRootBlock();
```

alschmiedt (approved):

core/keyboard_nav/ast_node.js
moniika: I can add that block of code inside of this `if block`.
We'll still need to check for the existence of the `getSourceBlock` method since it's not on the `Blockly.ASTNode` interface and some of the objects passed in may not have it defined.

PR #5137, opened 2021-07-20
feat: add deserialization of JSO block state

BeksOmega:
## The basics

- [X] I branched from **project-cereal**
- [X] My pull request is against **project-cereal**
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

Work on project cereal
Dependent on #5132 

### Proposed Changes

Adds deserialization of blocks. This handles everything except icons (ie comments), because I'm going to investigate their serialization and deserialization separately. Namely, type, id, coords, attributes, extra state, fields, and children.

Does not handle initialization of the deserialized blocks.

### Reason for Changes

Saving state is good.

### Test Coverage

Enabled all of the serializer tests for the JSO system except for icons, which aren't handled yet, and mutators, which aren't upgraded yet.

Issue #387, opened 2016-05-27
When adding multiple blocks with domToBlock, avoid resize calls until the end


rachel-fenichel:
Right now domToBlock always causes the workspace to resize--this is inefficient when adding lots of blocks at a time, including when loading a workspace or opening the flyout.

PR #5166, opened 2021-07-21
feat: add initialization of blocks and event firing

BeksOmega:
## The basics

### Resolves

Work on project cereal
Dependent on #5137 

### Proposed Changes

Adds (hopefully) properly initializing blocks for rendered workspaces, and firing block create events. Note that variable events are already [fired by the variable model](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/core/variable_model.js#L69).

This also modifies the playground (not the advanced playground) to use JSO serialization for testing.

### Reason for Changes

Properly rendering blocks is important. Also being consistent with our events system.

### Test Coverage

Added tests for properly firing events. If you have recommendations for testing the other parts of initialization I'd love to do that as well!

### Additional Information

NOTE FOR REVIEWERS: Please look over the logic in [domToBlock](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L543) and check if you think the logic added to the JSO serializer covers it. This uses a different (hopefully nicer) organization, but it should be equivalent.

### DIFF
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -262,10 +262,31 @@ const saveConnection = function(connection) {
  * Loads the block represented by the given state into the given workspace.
  * @param {!State} state The state of a block to deserialize into the workspace.
  * @param {!Workspace} workspace The workspace to add the block to.
+ * @return {!Block} The block that was just loaded.
  */
 const load = function(state, workspace) {
-  loadInternal(state, workspace);
+  // We only want to fire an event for the top block.
+  Blockly.Events.disable();
+
+  const block = loadInternal(state, workspace);
+
+  Blockly.Events.enable();
+  Blockly.Events.fire(
+      new (Blockly.Events.get(Blockly.Events.BLOCK_CREATE))(block));
+  
+  // Adding connections to the connection db is expensive. This defers that
+  // operation to decrease load time.
+  if (block instanceof Blockly.BlockSvg) {
+    setTimeout(() => {
+      if (!block.disposed) {
+        block.setConnectionTracking(true);
+      }
+    }, 1);
+  }
+
+  return block;
 };
+exports.load = load;
 
 /**
  * Loads the block represented by the given state into the given workspace.
@@ -275,6 +296,7 @@ const load = function(state, workspace) {
  * @param {!Workspace} workspace The workspace to add the block to.
  * @param {!Connection} parentConnection The optional parent connection to
  *     attach the block to.
+ * @return {!Block} The block that was just loaded.
  */
 const loadInternal = function(state, workspace, parentConnection = undefined) {
   const block = workspace.newBlock(state['type'], state['id']);
@@ -291,6 +313,8 @@ const loadInternal = function(state, workspace, parentConnection = undefined) {
   loadFields(block, state);
   loadInputBlocks(block, state);
   loadNextBlocks(block, state);
+  initBlock(block);
+  return block;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


BeksOmega:
Note that I didn't include [this resize call](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L580), because #387 has been closed for a long time, but someone who did the recent metrics refactor may want to take a look.

core/serialization/blocks.js
```diff
@@ -275,6 +296,7 @@ const load = function(state, workspace) {
  * @param {!Workspace} workspace The workspace to add the block to.
  * @param {!Connection} parentConnection The optional parent connection to
```
maribethb, line 297: I know it's not in this PR, but optional types need a `=` as in `!Connection=`

core/serialization/blocks.js
```diff
@@ -424,4 +448,22 @@ const loadConnection = function(connection, connectionState) {
         connection);
   }
 };
-exports.load = load;
+
+// TODO (#5146): Remove this from the serialization system.
```
maribethb, line 452: I think you need to remove the space after `TODO` in order to take advantage of automatic linking and the like

tests/mocha/jso_deserialization_test.js
```diff
@@ -0,0 +1,157 @@
...
+    this.workspace = new Blockly.Workspace();
+  });
+
+  teardown(function() {
+    workspaceTeardown.call(this, this.workspace);
+    sharedTestTeardown.call(this);
+  });
+
+  suite('Events', function() {
+    test.skip('Finished loading', function() {
```
maribethb, line 19: Will this be unskipped in an upcoming PR?

tests/mocha/jso_deserialization_test.js
```diff
@@ -0,0 +1,157 @@
...
+        Blockly.serialization.workspaces.load(state, this.workspace);
+        assertEventFired(
+            this.eventsFireStub,
+            Blockly.Events.BlockCreate,
+            {},
+            this.workspace.id,
+            'testId');
+      });
+
+      test('with children', function() {
```
maribethb, line 120: We don't really have a consistent style for tests imo... so feel free to ignore. But I personally like to add which behavior I expect to happen in the test so you can see at a glance. For some tests if it's really obvious then saying something like "Var creation causes var create event to be fired" might be obvious/redundant but for some of them I think "only top block event is fired" or something can make it obvious what should have happened, when a test fails. e.g. https://github.com/google/blockly/blame/master/tests/mocha/block_test.js#L1636

tests/playground.html
```diff
@@ -204,22 +204,22 @@
   }
 }
 
-function toXml() {
+function save() {
```
maribethb, line 207: I think these should probably be separate buttons instead of replacing the xml buttons, at least for now. We still use the non-advanced playground for some QA test cases and this includes exporting the XML and using the Code demo to import the XML. It could still be removed at some point in the future but I'm hesitant about removing it for now even though this is mostly a test page for us.

maribethb (approved):
Can we have a test where you deserialize some json and then inspect the workspace to make sure those blocks were created correctly? or is that covered somewhere else?

core/serialization/blocks.js
BeksOmega: Oh whoops! Thank you for the catch

tests/mocha/jso_deserialization_test.js
BeksOmega: Yesh this gets unskipped in the next one.

BeksOmega:
> Can we have a test where you deserialize some json and then inspect the workspace to make sure those blocks were created correctly? or is that covered somewhere else?

That is covered by the round-trip tests. We:
  1) Deserialize some XML.
  2) Serialize that to a JSO
  3) Deserialize the JSO
  4) Serialize that to XML
  5) Check that the start XML and end XML are identical

Because we know that the end XML will only be identical to the start XML if the blocks deserialized from XML are identical to the blocks deserialized from JSO *and* because we know that the XML deserializes properly, this proves that the JSO system deserializes properly.

But I've also checked it manually just in case hehe.

tests/mocha/jso_deserialization_test.js
BeksOmega: I think that's a good suggestion! I think it would also be nice to have a styleguide for tests, or at least some more consistency in them haha.

maribethb (approved):

PR #5204, opened 2021-07-23
Migrate core/renderers/common/constants.js to goog.module syntax

rachel-fenichel:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/renderers/common/constants.js` to `goog.module` with ES6 `const`/`let`.

### Additional Information

I did the change to `const`/`let` at the end--sorry about that.

The first commit changes how the debug filter is created. Previously it was checking `if (Blockly.blockRendering.Debug)` to see if the debug renderer existed, which breaks with named requires + optionally requiring the debug renderer. I changed it to instead create the debug filter regardless, if it didn't already exist. This happens at most once per workspace and I don't expect the existence of the filter to be expensive.

I propose that I change the debug renderer in dev-tools to call this new function (`createDebugFilter`), rather than having the constants provider do it. If that's approved, I'll file an issue for that and set it up to happen when we next release core.

Note that there is a debug renderer in core, but there's already an issue open to delete it this quarter.


rachel-fenichel:
https://github.com/google/blockly-samples/issues/826 tracks followup work.

core/renderers/common/constants.js
```diff
@@ -401,6 +400,14 @@ Blockly.blockRendering.ConstantProvider = function() {
    */
   this.randomIdentifier = String(Math.random()).substring(2);
 
+  /**
+   * The defs tag that contains all filters and patterns for this Blockly
+   * instance.
+   * @type {SVGElement}
+   * @private
+   */
+  this.defs = null;
```
alschmiedt, line 409: I think we are keeping `_` on class properties.

alschmiedt (approved):

core/renderers/common/constants.js
rachel-fenichel: Done.

PR #5207, opened 2021-07-23
Migrate dropdown to goog.module syntax

alschmiedt:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/dropdowndiv.js` to `goog.module` with ES6 syntax.

### Additional Information
I'm keeping this as a draft right now because I'm not sure that I did this correctly, or even if we want to go this direction. I was more curious what it might look like for this class.


core/dropdowndiv.js
```diff
@@ -744,5 +744,11 @@ DropDownDiv.repositionForWindowResize = function() {
     DropDownDiv.hide();
   }
 };
-
 exports = DropDownDiv;
+
+exports.testOnly_setGetBoundsInfo = function(getBoundsInfoMock) {
+  goog.setTestOnly();
+  getBoundsInfo = getBoundsInfoMock;
```
moniika, line 751: I think this needs to be changed slightly in order to work for other modules that destructure `getPositionMetrics_`. Also, and you may need to check this, I'm not sure this would affect the export if the variable `getPositionMetrics` is modified after the module is loaded.

In the case of destructuring, since the module under test will have already been loaded, this set for `getBoundsInfo` wont affect the local copy of `getBoundsInfo` that the module being tested has destructured.
In order to resolve this, `getPositionMetrics` should call the mock if it exists.

```
let mockGetBoundsInfo = null;

const getBoundsInfo = function(...) {
  if (mockGetBoundsInfo) {
    mockGetBoundsInfo();
  }
  // ...
}

exports.testOnly_setGetBoundsInfo = function(mock) {
  goog.setTestOnly();
  mockGetBoundsInfo = mock;
}
```

cpcallen: @moniika: I'm not sure what you mean by "In the case of destructuring, since the module under test will have already been loaded, this set for getBoundsInfo wont affect the local copy of getBoundsInfo that the module being tested has destructured."

I note that `getBoundsInfo` is a local variable in this module, and each module only gets loaded once (no matter how many times it is required), so there is only `getBoundsInfo` and it can have only one value (at a time).

Perhaps the confusion is that when we write `const {func} = goog.require('module');` we sometimes say that "the destructuring gives us a copy of func"—but this is not so: it gives us only _a copy of the reference_ to the  function object in question.  That _can_ cause problems if our tests want to replace  this reference with a reference to a different function, because the importer will continue to use their unmodified copy, which still points at the original implementation.  But destructuring a reference to `getPositionMetrics_` (were it exported) won't affect what function is called when `getPositionMetrics_` invokes `getBoundsInfo`: it will call whatever value was most recently passed to `testOnly_setGetBoundsInfo`, or the original implementation if `testOnly_setGetBoundsInfo` has never been invoked.

core/dropdowndiv.js
```diff
@@ -744,5 +744,11 @@ DropDownDiv.repositionForWindowResize = function() {
...
+};
```
cpcallen, line 752: Ought we not to provide a mechanism for the test harness to restore the original value of `getBoundsInfo` once the test is over?  Perhaps doing so if `getBoundsInfoMock` is `null`?

Or does each test completely reload Blockly, such that a change to `getBoundsInfo` will not persist past the end of that specific test?

core/dropdowndiv.js
```diff
@@ -744,5 +744,11 @@ DropDownDiv.repositionForWindowResize = function() {
...
+
+exports.testOnly_getPositionMetrics = getPositionMetrics;
```
cpcallen, line 754: Pedantically, this should probably be a wrapper:

```JS
/** ... */
exports.testOnly_getPositionMetrics = function (/* ... */) {
  goog.setTestOnly();
  return getPositionMetrics(arguments);
}
```

Since we may have many such wrappers it might be worth defining a higher-order function to create them, so you could just write something like:

```JS
exports.testOnly_foo = testUtils.testOnly(foo);
```

scripts/gulpfiles/build_tasks.js
```diff
@@ -342,6 +342,8 @@ function buildDeps(done) {
     closurePath,
     'core',
     'blocks',
+      'generators',
+      'tests/mocha'
```
cpcallen, line 346: Indentation.

cpcallen (changes requested):
Seems solid but there are a few small issues.  Nice work with the test fixes!

core/dropdowndiv.js
moniika: Thanks for the clarification Chris, that it's a copy of the reference and not a copy of the function. I think I did make that mistake.

I might have also mixed up function names (as I was writing this) and was considering the situation where if `getBoundsInfo` was exported (which is not actually the case) then it could cause issues if `getBoundsInfo` was destructured.

core/dropdowndiv.js
```diff
@@ -195,9 +194,9 @@ DropDownDiv.createDom = function() {
   divInternal.style.opacity = 0;
 
   // Transition animation for transform: translate() and opacity.
-  divInternal.style.transition = 'transform ' +
-    DropDownDiv.ANIMATION_TIME + 's, ' +
-    'opacity ' + DropDownDiv.ANIMATION_TIME + 's';
+  divInternal.style.transition = 'transform ' + DropDownDiv.ANIMATION_TIME +
+      's, ' +
+      'opacity ' + DropDownDiv.ANIMATION_TIME + 's';
```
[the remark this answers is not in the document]
alschmiedt: Unfortunately, it will not.

core/dropdowndiv.js
```diff
@@ -31,97 +32,108 @@ goog.requireType('Blockly.WorkspaceSvg');
...
 
 /**
  * Drop-downs will appear within the bounds of this element if possible.
- * Set in Blockly.DropDownDiv.setBoundsElement.
+ * Set in DropDownDiv.setBoundsElement.
  * @type {Element}
- * @private
  */
-Blockly.DropDownDiv.boundsElement_ = null;
+let boundsElementInternal = null;
```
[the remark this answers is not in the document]
alschmiedt: Ok. So I tried to go down the road of giving everything better names but it ended up being I think a larger change than I wanted to make right now, so I went back to having everything as a `@private` static property.

core/dropdowndiv.js
alschmiedt: Ah, this is a good point. AFAICT the tests do not completely reload Blockly. We would need something to restore the old value. However, I'm not entirely sure the best way to do this. I went spelunking through code and couldn't find any instances where they do this. Any ideas?

alschmiedt: Monica mentioned a way to do this would be to set a module-local variable and inside `getBoundsInfo` do
```
if (mockBoundsInfo) {
   mockBoundsInfo();
   return;
}
```
I'm a bit hesitant to add test logic to our code like this, so if we can't find a better solution I plan on chalking this up to a fun experiment and going with the original plan to just export the small number of private methods that we need in order to make our tests work.

core/dropdowndiv.js
```diff
@@ -13,115 +13,139 @@
...
+ * The arrow element.
+ * @type {!Element}
+ * @private
+ */
+DropDownDiv.arrow_;
+
+/**
+ * Drop-downs will appear within the bounds of this element if possible.
+ * Set in DropDownDiv.setBoundsElement.
+ * @type {Element}
```
NeilFraser, line 131: I think this should be {?Element} (and {?Object} below) in accordance with the new style guide.

core/dropdowndiv.js
```diff
@@ -150,61 +174,47 @@ Blockly.DropDownDiv.BoundsInfo;
...
-  Blockly.DropDownDiv.DIV_.style.opacity = 0;
+  DropDownDiv.DIV_.style.opacity = 0;
 
   // Transition animation for transform: translate() and opacity.
-  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' +
-    Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
-    'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
+  DropDownDiv.DIV_.style.transition = 'transform ' + DropDownDiv.ANIMATION_TIME +
+      's, ' +
+      'opacity ' + DropDownDiv.ANIMATION_TIME + 's';
```
NeilFraser, line 209: Unneeded line break.

core/dropdowndiv.js
cpcallen: I will leave this to your judgement, as I think you understand the tests much better than I do.

cpcallen (approved):
This looks fine now, except for the question about restoring functions that have been mocked for the tests, and the two nits that Neil mentions.

core/dropdowndiv.js
alschmiedt: done

NeilFraser (approved):

PR #5217, opened 2021-07-26
Migrate core/registry.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/registry.js` to `goog.module` with ES6 `const`/`let`.


core/registry.js
```diff
@@ -120,13 +132,12 @@ Blockly.registry.Type.BLOCK_DRAGGER =
...
-Blockly.registry.register = function(
-    type, name, registryItem, opt_allowOverrides) {
-  if ((!(type instanceof Blockly.registry.Type) && typeof type != 'string') ||
+const register = function(type, name, registryItem, opt_allowOverrides) {
+  if ((!(type instanceof Type) && typeof type != 'string') ||
       String(type).trim() == '') {
     throw Error(
         'Invalid type "' + type + '". The type must be a' +
-        ' non-empty string or a Blockly.registry.Type.');
+        ' non-empty string or a Type.');
```
maribethb, line 140: I think this error message should retain the namespace for legibility.

core/registry.js
```diff
@@ -34,81 +47,80 @@ goog.requireType('Blockly.ToolboxItem');
...
  */
-Blockly.registry.Type = function(name) {
+const Type = function(name) {
   /**
    * @type {string}
    * @private
    */
   this.name_ = name;
 };
+exports.Type = Type;
```
maribethb, line 74: This seems like it may be one of the cases where Type ought to be a class in its own file and moved out of this one, and we should add it to the list of weird files to follow up on. Because this isn't just a static function on the registry namespace like many of the other files have.

maribethb (approved):

core/registry.js
gonfunko: Good call, updated.

core/registry.js
gonfunko: Added a note about this to the tracking issue, I agree it seems reasonable to split it out into its own file.

PR #5219, opened 2021-07-26
Migrate core/shortcut_registry.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/shortcut_registry.js` to `goog.module` with ES6 `const`/`let`.


BeksOmega (approved):
LGTM! But I'm not sure if we need to make any changes wrt how singleton instances are handled. I think we should get @cpcallen 's opinion before merging.

core/interfaces/i_keyboard_accessible.js
```diff
@@ -15,7 +15,7 @@ goog.module('Blockly.IKeyboardAccessible');
 goog.module.declareLegacyNamespace();
 
 /* eslint-disable-next-line no-unused-vars */
-const {KeyboardShortcut} = goog.requireType('Blockly.ShortcutRegistry');
+const ShortcutRegistry = goog.requireType('Blockly.ShortcutRegistry');
```
cpcallen, line 18: This de-destructuring probably wasn't *really* necessary: since this is a `requireType` (and not just `require`) it can't be necessary to stub/mock this dependency in tests—and the name `KeyboardShortcut` seems adequately unambiguous on its own.

core/shortcut_registry.js
```diff
@@ -22,16 +23,16 @@ goog.requireType('Blockly.Workspace');
...
  * singleton. You should not create a new instance, and only access this class
- * from Blockly.ShortcutRegistry.registry.
+ * from ShortcutRegistry.registry.
  * @constructor
  */
-Blockly.ShortcutRegistry = function() {
+const ShortcutRegistry = function() {
   // Singleton instance should be registered once.
-  Blockly.ShortcutRegistry.registry = this;
+  ShortcutRegistry.registry = this;
```
cpcallen, line 31: Having the constructor overwrite the singleton seems like a bad idea (especially given the constructor is exported!)  No fix required now, but see below.

core/interfaces/i_keyboard_accessible.js
cpcallen, line 18: This de-destructuring probably wasn't *really* necessary: since this is a `requireType` (and not just `require`) it can't be necessary to stub/mock this dependency in tests—and the name `KeyboardShortcut` seems adequately unambiguous on its own.

core/shortcut_registry.js
```diff
@@ -352,4 +353,6 @@ Blockly.ShortcutRegistry.prototype.createSerializedKey = function(
 };
 
 // Creates and assigns the singleton instance.
-new Blockly.ShortcutRegistry();
+new ShortcutRegistry();
+
+exports = ShortcutRegistry;
```
cpcallen, line 353: Fine for now (to keep API compatibility until the first pass is finished), but it should probably be changed to something like:
```JS
exports.shortcutRegistry = new ShortcutRegistry();
```
i.e., only exporting the singleton, not the constructor (and not having the constructor overwrite any previously-created singleton).

Added to the list in #5073.

cpcallen (approved):
LGTM, but see comments below re: future work.

core/interfaces/i_keyboard_accessible.js
gonfunko: The compiler yelled without it, I believe about not being able to destructure a default export.

core/shortcut_registry.js
gonfunko: Yeah that does seem much nicer, good call!

core/interfaces/i_keyboard_accessible.js
cpcallen: Oh, yeah.  Right.  I forgot it has _opinions_ about this sort of thing.

Moving away from default exports will help solve this.

PR #5222, opened 2021-07-26
Migrate core/tooltip.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/tooltip.js` to `goog.module` with ES6 `const`/`let`.


core/tooltip.js
```diff
@@ -32,102 +33,104 @@ goog.require('Blockly.utils.string');
...
  */
-Blockly.Tooltip.blocked_ = false;
+let blocked = false;
 
 /**
  * Maximum width (in characters) of a tooltip.
  */
-Blockly.Tooltip.LIMIT = 50;
+const LIMIT = 50;
+exports.LIMIT = LIMIT;
```
cpcallen, line 55: Are we expecting that developers will want to modify this value?

core/tooltip.js
```diff
@@ -32,102 +33,104 @@ goog.require('Blockly.utils.string');
...
-Blockly.Tooltip.TipInfo;
+let TipInfo;
+exports.TipInfo = TipInfo;
 
 /**
  * Is a tooltip currently showing?
  */
-Blockly.Tooltip.visible = false;
+let visible = false;
+exports.visible = visible;
```
cpcallen, line 43: This doesn't look correct.  If we're exporting `visible`, presumably external code wants to know the _current_ value of the flag, not just the initial value.

See [the style guide](https://google.github.io/styleguide/jsguide.html#es-module-exports) under "Mutability of exports" for how to deal with this (but maybe for now `exports.visible` needs to be an actual getter, for backwards compatibility.)

(Also: perhaps add `@type {boolean}` to the JSDoc?)

cpcallen (commented):
Generally fine, but I have a few questions.

cpcallen:
I'm un-assigning this from myself for now but feel free to reassign when you want a re-review.

core/tooltip.js
gonfunko: Checked with Beka and she wasn't aware of and couldn't find anyone that was. I take the all-caps to imply that it was intended as a constant as well.

core/tooltip.js
gonfunko: Updated this (and DIV) following the pattern we've settled on.

core/tooltip.js
cpcallen: I guess the question then is: does it need to be exported?  I don't see reference to it anywhere outside this file (but maybe it's referenced outside this repo?)

core/tooltip.js
```diff
@@ -39,10 +40,28 @@ exports.TipInfo = TipInfo;
 
 /**
  * Is a tooltip currently showing?
+ * @type {boolean}
  */
 let visible = false;
+/** @deprecated September 2021 */
 exports.visible = visible;
```
cpcallen, line 47: This is being overwritten by a call to `Object.defineProperties` below, and somehow I dislike that, though I can't articulate exactly why.

It this the easiest way to get the correct documentation and type info associated with the exported name?  I think I'd prefer a pure declaration, even though we'd be repeating the JSDoc:

```JS
/**
 * Is a tooltip currently showing?
 * @type {boolean}
 * @deprecated September 2021
 */
exports.visible;
```
Or even—since it's deprecated and a compiler warning might encourage people to migrate—just:

```JS
/** @deprecated September 2021 */
exports.visible;
```

("This is not important; wontfix" might be reasonable answer.)

cpcallen (approved):
I think this is fine, but I did have one niggle.

core/tooltip.js
gonfunko: Looks like it's used in blockly-samples: https://github.com/google/blockly-samples/blob/1be2b748414ef147c8f7d657fa440442c0ef49a8/plugins/block-extension-tooltip/src/tooltip_monkey_patch.ts

BeksOmega: That's kind of annoying. First party plugins aren't supposed to monkey patch according to [our own criteria](https://developers.google.com/blockly/guides/plugins/overview#first-party-criteria). Sucks that this has to be made public because of that :/

core/tooltip.js
gonfunko: Done!

PR #5223, opened 2021-07-26
Add comment to top of convert-file script for reference to syntax used

moniika:
* Adds comment to top of file as reference for syntax used.
* Converts some `perl -nle` commands to a simplified `perl -ne` for commands that are used to extract only a single match
* Adds space to perl commands (`perl -nle'command'` => `perl -nle 'command'`)
* Removed an unnecessary `tr ' ' '\n'` that was noticed on L207 (old L172)

### DIFF
--- a/scripts/goog_module/convert-file.sh
+++ b/scripts/goog_module/convert-file.sh
@@ -4,11 +4,11 @@
 # replacing string (regex) patterns in a way that is both GNU and macOS
 # compatible.
 #
-# Common perl flags used (also decribed at https://perldoc.perl.org/perlrun):
+# Common perl flags used (also described at https://perldoc.perl.org/perlrun):
 #   -e : Used to execute perl programs on the command line
 #   -p : Assumes an input loop around script. Prints every processed line.
 #   -n : Assumes an input loop around script. Does not print every line.
-#   -i : Used for in-place editing. Used for commands for find/replace.
+#   -i : Used for in-place editing. Used in commands for find/replace.
 #   -l[octnum] : Assigns the output record separator "$/" as an octal number. If
 #        octnum is not present, sets output record separator to the current
 #        value of the input record separator "$\".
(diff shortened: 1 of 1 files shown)


scripts/goog_module/convert-file.sh
```diff
@@ -1,5 +1,39 @@
...
+#   -l[octnum] : Assigns the output record separator "$/" as an octal number. If
+#        octnum is not present, sets output record separator to the current
+#        value of the input record separator "$\".
+#
+# Common perl commands found:
+# 1.  perl -pi -e 's/regex/replacement/modifiers'
+#   This command does an in-place search-and-replace. The global ("/g") modifier
+#   causes it to replace all occurrences, rather than only the first match.
+# 2.  perl -ne 'print m/regex/modifiers'
+#   This command returns the all regex matches. If the global ("/g") modifier is
```
rachel-fenichel, line 21: This explanation is confusing. Please rewrite.

> This command returns the all regex matches.

Extra "the", but I understand that it returns all matches.

> If the global ("/g") modifier is specified, then the capture group "()" is not necessary and it will return all matches, rather than only the first match.

Wait, if it returns all matches if global is specified, then what does it return if not specified? Where is the capture group in the example? Is the first sentence in the explanation incorrect?

moniika: I've updated the wording for what the match contains. It returns the first match if the global modifier is not specified and It returns a string containing all matches if the global modifier is specified.

The capture group is referring to the regex, so I've updated it to say that the regex group must specify a capture group. I'm going to do a little reading about that detail because although it's something I encountered as I was testing my commands, there could be a more accurate thing to say here.

moniika: I've updated the comment further based on my understanding.

rachel-fenichel (approved):

Issue #4600, opened 2021-01-21
Getters for copy information


alschmiedt:
**Is your feature request related to a problem? Please describe.**
In the keyboard navigation demo I override the default paste functionality so I can paste the copied block to the current location of the cursor. In order to do this I need to use know what block was copied and from what workspace. This requires using `Blockly.clipboardSource_` and `Blockly.clipboardXml_`.

**Describe the solution you'd like**
Getters for these fields. 

**Describe alternatives you've considered**
Making the fields public. I'm not sure this is necessary. Since we have `Blockly.copy`, I don't see a reason why someone would want to set these fields manually.

**Additional context**

George-the-1st:
I should be mentioning that I'm almost finished with this issue.

cpcallen:
Closed by #5237.

Issue #5208, opened 2021-07-24
Review contents of `blockly.js`


rachel-fenichel:
`blockly.js` is full of public stuff that doesn't really fit on any namespace. This is a problem because internal code requiring it causes circular dependencies.

The fix is to move everything in `blockly.js` out into separate files and require those from inside the codebase, while leaving aliases behind in `blockly.js` for anything that should stay publicly accessible.

In some cases, we also need setters so that external code can change those values.

In other cases, properties should just not be exported.

This issue is for tracking decisions about these properties and whether those decisions have been implemented.

Needs decision:
- [ ] `svgResize`
- [ ] `onKeyDown`
- [ ] `deleteBlock`
- [ ] `onContextMenu_`
- [x] `hideChaff`
- [x] `alert`
- [x] `confirm`
- [x] `prompt`
- [x] `hueToHex`

Can be internal:
- [ ] `Blockly.jsonInitFactory_`
- [x] `clipboardXml_` -> `clipboard.js` in #5237
- [x] `clipboardSource_` -> `clipboard.js` in #5237
- [x] `clipboardTypeCounts_` -> `clipboard.js` in #5237
- [x] `parentContainer` -> `common.js` in #5262
- [x] `draggingConnections` -> `common.js` in #5262

Exported function, but moved to a different file and aliased.
- [x] `bindEvent_` -> `Blockly.browserEvents.bind` in #4642 (2021Q1)
- [x] `getMainWorkspace` -> `Blockly.common.getMainWorkspace` in #5244
- [x] `copy` -> `Blockly.clipboard.copy` in #5237, and tracking remaining work in https://github.com/google/blockly-samples/issues/833
- [x] `paste -> `Blockly.clipboard.paste` in #5237, and tracking remaining work in https://github.com/google/blockly-samples/issues/833
- [x] `duplicate` -> `Blockly.clipboard.duplicate` in #5237, and tracking remaining work in https://github.com/google/blockly-samples/issues/833
- [x] `setParentContainer` -> `Blockly.common.setParentContainer` in #5262
- [x] `hueToHex` -> `Blockly.utils.colour.hueToHex` in #5462
- [x] `alert` -> `Blockly.dialog.alert` in #5457 
- [x] `confirm` -> `Blockly.dialog.confirm` in #5457 
- [x] `prompt` -> `Blockly.dialog.prompt` in #5457

Exported function, used only externally and therefore left in `blockly.js`:
- [ ] `isNumber`: Used in generators
[...]

PR #5237, opened 2021-07-28
Move clipboard functions to a separate namespace

rachel-fenichel:
- Moves clipboard functions `copy`, `paste`, and `duplicate` out of `blockly.js`.
- Moves clipboard state `clipboardXml_`, `clipboardTypeCounts_`, and `clipboardSource_` out of `blockly.js`
- Aliases old functions and properties to refer to new ones.
- Updates tests that spy on specific functions.
- Updates callers inside core.

This is part of https://github.com/google/blockly/issues/5208

There's an API question here: should clipboard xml, type counts, and source be exported? The keyboard nav plugin accesses them directly, but with a note that it needs a better way to do that.

I propose that there should instead be a getter for `clipboardInfo` that returns an object containing all three of those properties, but as copies (and effectively read-only).

I also propose that we remove `clipboardXml_`, `clipboardTypeCounts_`, and `clipboardSource_` from `blockly.js` entirely and not export those three from `clipboard`.


alschmiedt (commented):
I agree with everything proposed in the PR description. We just also need to make an issue to update keyboard nav to use the new API before releasing.

rachel-fenichel:
Tracking -samples update in https://github.com/google/blockly-samples/issues/833

This PR also closes https://github.com/google/blockly/issues/4600

core/blockly.js
```diff
@@ -402,9 +352,7 @@ Blockly.defineBlocksWithJsonArray = function(jsonArray) {
               'Block definition #' + i + ' in JSON array' +
               ' overwrites prior definition of "' + typename + '".');
         }
-        Blockly.Blocks[typename] = {
-          init: Blockly.jsonInitFactory_(elem)
-        };
+        Blockly.Blocks[typename] = {init: Blockly.jsonInitFactory_(elem)};
```
rachel-fenichel, line 355: Clang-format did this.

core/clipboard.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+const copy = function(toCopy) {
+  var data = toCopy.toCopyData();
+  if (data) {
+    xml = data.xml;
+    source = data.source;
+    typeCounts = data.typeCounts;
+  }
+};
+exports.copy = copy;
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

core/blockly.js
```diff
@@ -220,7 +206,8 @@ Blockly.deleteBlock = function(selected) {
     Blockly.Events.setGroup(true);
     Blockly.hideChaff();
     if (selected.outputConnection) {
-      // Do not attempt to heal rows (https://github.com/google/blockly/issues/4832)
+      // Do not attempt to heal rows
```
rachel-fenichel, line 209: clang-format did this.

core/blockly.js
```diff
@@ -136,9 +124,7 @@ Blockly.svgSize = function(svg) {
   // When removing this function, remove svg.cachedWidth_ and svg.cachedHeight_
   // from setCachedParentSvgSize.
   Blockly.utils.deprecation.warn(
-      'Blockly.svgSize',
-      'March 2021',
-      'March 2022',
+      'Blockly.svgSize', 'March 2021', 'March 2022',
```
rachel-fenichel, line 127: clang-format did this.

core/blockly.js
```diff
@@ -92,25 +93,12 @@ Blockly.selected = null;
...
- * @private
- */
-Blockly.clipboardXml_ = null;
-
-/**
- * Source of the local clipboard.
- * @type {Blockly.WorkspaceSvg}
- * @private
+ * Get the current contents of the clipboard and associated metadata.
+ * @return {{xml: ?Element, source: ?Blockly.WorkspaceSvg, typeCounts: ?Object}}
```
alschmiedt, line 97: If we could provide a better type for `typeCounts` now that this is public I think that would be really helpful.

core/blockly.js
```diff
@@ -92,25 +93,12 @@ Blockly.selected = null;
...
+ *     An object containing the clipboard contents and associated metadata.
+ * @public
```
alschmiedt, line 99: I think we aren't adding `@public`.

core/blockly.js
```diff
@@ -92,25 +93,12 @@ Blockly.selected = null;
...
  */
-Blockly.clipboardSource_ = null;
-
-/**
- * Map of types to type counts for the clipboard object and descendants.
- * @type {Object}
- * @private
- */
-Blockly.clipboardTypeCounts_ = null;
+Blockly.getClipboardInfo = Blockly.clipboard.getClipboardInfo;
```
alschmiedt, line 101: Do we need to add this here since it is a new API?

alschmiedt (approved):

BeksOmega:
Just tagging this as something I was planning on switching over to the JSO system. But since this is such a big change it may be better to wait.

core/blockly.js
rachel-fenichel: I switched all of this to `ICopyable.CopyData`, which is an existing type.

core/blockly.js
rachel-fenichel: I removed this for now. However, we will have to revisit when we remove `declareLegacyNamespace` in `clipboard.js`. Since the clipboard didn't exist at all before this quarter, and I prefer to keep it internal, I may end up adding this single function at the top level.

core/blockly.js
rachel-fenichel: Deleted function.

rachel-fenichel:
@alschmiedt I made some significant changes, so please take another look.

Changes:
- Switched to storing data as an `ICopyable.CopyData` instead of storing each component individually.
- Got rid of the exported function `getClipboardInfo` and also removed it from `blockly.js`--it's not clear that it's the right API, and we have an issue tracking that.

alschmiedt (approved):

PR #5243, opened 2021-07-29
Move Ok/Cancel messages into Blockly.

NeilFraser:


NeilFraser:
These translations are coming from Blockly Games and are being removed from that project with this companion PR: https://github.com/google/blockly-games/pull/225

moniika (approved):

PR #5246, opened 2021-07-29
Add serializing shadows as JSOs for the JSO system

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/comments...BeksOmega:cereal/shadows
### Resolves

Work on project cereal
Dependent on #5216

### Proposed Changes
Makes it so that shadow blocks are encoded as JSOs rather than stringified XML when using the JSO serialization system.

### Reason for Changes
Stringified XML takes up more space than the normal JSO representation.

### Test Coverage
Added tests to make sure that the new setShadowState hook works correctly, and that the state serializes to XML and JSO properly, regardless of whether you use the setShadowDom or setShadowState hooks.

Also adds two simple tests for recycling vs not recycling blocks.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -744,7 +744,7 @@ Blockly.Connection.prototype.setShadowStateInternal_ =
           this.serializeShadow_(this.targetBlock());
         }
       } else {
-        var shadow = this.createShadowBlock_();
+        var shadow = this.createShadowBlock_(false);
         this.serializeShadow_(shadow);
         if (shadow) {
           shadow.dispose(false);
@@ -791,6 +791,7 @@ Blockly.Connection.prototype.createShadowBlock_ = function(attemptToConnect) {
     }
     return blockShadow;
   }
+  return null;
 };
 
 /**
--- a/core/connection.js
+++ b/core/connection.js
@@ -603,8 +603,8 @@ Blockly.Connection.prototype.getShadowDom = function(returnCurrent) {
 
 /**
  * Changes the connection's shadow block.
- * @param {?blocks.State} shadowState An state represetation of the block or
- *     null.
+ * @param {?Blockly.serialization.blocks.State} shadowState An state
+ *     represetation of the block or null.
  */
 Blockly.Connection.prototype.setShadowState = function(shadowState) {
   this.setShadowStateInternal_({shadowState: shadowState});
@@ -617,13 +617,14 @@ Blockly.Connection.prototype.setShadowState = function(shadowState) {
  *     attached to this connection, this serializes the state of that block
  *     and returns it (so that field values are correct). Otherwise the saved
  *     state is just returned.
- * @return {?blocks.State} Serialized object representation of the block, or
- *     null.
+ * @return {?Blockly.serialization.blocks.State} Serialized object
+ *     representation of the block, or null.
  */
 Blockly.Connection.prototype.getShadowState = function(returnCurrent) {
   if (returnCurrent && this.targetBlock() && this.targetBlock().isShadow()) {
-    return Blockly.serialization.blocks
-        .save(this.targetBlock(), {addNextBlocks: true});
+    return Blockly.serialization.blocks.save(
+        /** @type {!Blockly.Block} */ (this.targetBlock()),
+        {addNextBlocks: true});
   }
   return this.shadowState_;
 };
@@ -700,7 +701,8 @@ Blockly.Connection.prototype.toString = function() {
 /**
  * Returns the state of the shadowDom_ and shadowState_ properties, then
  * temporarily sets those properties to null so no shadow respawns.
- * @return {{shadowDom: ?Element, shadowState: ?Object}} The state of both the
+ * @return {{shadowDom: ?Element,
+ *     shadowState: ?Blockly.serialization.blocks.State}} The state of both the
  *     shadowDom_ and shadowState_ properties.
  * @private
  */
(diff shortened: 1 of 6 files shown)


BeksOmega:
I'm not sure how to add RealChildOfShadow errors to this. (These errors would be thrown if you're trying to deserialize a JSO specifies a real block should be the child of a shadow block)

Options that don't work:
  * Adding a "shadow" parameter to the `load` function. This does not work because we don't want that to be part of the external API.
  * Adding a "shadow" parameter to the `loadInternal` function. This does not work because shadow deserialization has to be triggered by the connection so that it can store the serialized version of the shadow as a property. And connections will only have access to the exported interface of the serialization, not any internal functions.

So it looks like we have to add a "shadow" property to the object defining the shadow block. Eg:
```js
{
  'type': 'text_print',
  'next': {
    'shadow': {
      'type': 'text_print',
      'kind': 'shadow'  // We have to add this.
    }
  }
}
```

The question is whether we want to:
  * Always add this property when serializing, and require it on all blocks when deserializing.
  * Always add this property to shadow blocks when serializing, and require it on shadow blocks when deserializing.
  * Not add this property when serializing, and add it _during_ deserialization to track the shadow/non-shadowness.

Or maybe there's another even better alternative :D

BeksOmega:
Remove references to `addNextBlocks: true` because it is now true by default.

core/connection.js
```diff
@@ -675,3 +697,103 @@ Blockly.Connection.prototype.toString = function() {
   }
   return msg + block.toDevString();
 };
+
+/**
+ * Returns the state of the shadowDom_ and shadowState_ properties, then
+ * temporarily sets those properties to null so no shadow respawns.
+ * @return {{shadowDom: ?Element, shadowState: ?Object}} The state of both the
+ *     shadowDom_ and shadowState_ properties.
+ */
```
maribethb, line 706: This function and many of the ones below it look like they're intended to be private/package, but they don't have access modifiers.

core/connection.js
```diff
@@ -675,3 +697,103 @@ Blockly.Connection.prototype.toString = function() {
...
+Blockly.Connection.prototype.setShadowStateInternal_ =
+    function({shadowDom = null, shadowState = null} = {}) {
+      // One or both of these should always be null.
+      // If neither is null, the shadowState will get priority.
+      this.shadowDom_ = shadowDom;
+      this.shadowState_ = shadowState;
+
+      var target = this.targetBlock();
+      if (!target) {
```
maribethb, line 742: Could you explain these 3 different cases? Not necessarily in code comments yet but just I don't understand why each case does what it does.

maribethb (commented):
It's hard to tell how many lines you actually added to connection_test vs how many were changed and the diffing tool shows an untidy diff. But that file is huge now (big enough GitHub doesn't load it by default) so I think if you added a lot of lines/tests, it's worth considering moving them to a new test file for ease of use/reading them/finding the specific behavior's tests in a giant file.

core/connection.js
BeksOmega: I can give it my best shot haha

1) `!target` means that there is no connected block (no shadow and no real). In this case we want to try to respawn a shadow. If one is created, we serialize it.
2) `target.isShadow()` means there is a connected shadow block. In this case we want to dispose of that and try to respawn a new shadow. If one is created, we serialize it.
3) Otherwise we know we have a target block (b/c `!target` was false) and we know it's not a shadow, so we must have a connected real block. In this case we just want to create the shadow block (without trying to connect it) and serialize it.

The serializing is to make sure that our `shadowDom_` *and* `shadowState_` properties both have the correct value so that both serialization systems work properly.

Tell me if there's anything I didn't explain well. Cuz honestly all of this logic is a bit of a (hopefully necessary?) mess.

And if there's anything I can do to clean this up, definitely tell me that as well.

BeksOmega:
> It's hard to tell how many lines you actually added to connection_test vs how many were changed and the diffing tool shows an untidy diff. But that file is huge now (big enough GitHub doesn't load it by default) so I think if you added a lot of lines/tests, it's worth considering moving them to a new test file for ease of use/reading them/finding the specific behavior's tests in a giant file.

If you check out [this commit](https://github.com/BeksOmega/blockly/commit/42f89440d6f304537e90cdf6700940d66f8c616e) is has just the test edits/additions without the reorganization. There were some tests added, but I'm not sure if it's worth moving stuff into a new file.

BeksOmega:
> I'm not sure how to add RealChildOfShadow errors to this. (These errors would be thrown if you're trying to deserialize a JSO specifies a real block should be the child of a shadow block)
> 
> Options that don't work:
> 
> * Adding a "shadow" parameter to the `load` function. This does not work because we don't want that to be part of the external API.
> * Adding a "shadow" parameter to the `loadInternal` function. This does not work because shadow deserialization has to be triggered by the connection so that it can store the serialized version of the shadow as a property. And connections will only have access to the exported interface of the serialization, not any internal functions.
> 
> So it looks like we have to add a "shadow" property to the object defining the shadow block. Eg:
> 
> ```js
> {
>   'type': 'text_print',
>   'next': {
>     'shadow': {
>       'type': 'text_print',
>       'kind': 'shadow'  // We have to add this.
>     }
>   }
> }
> ```
> 
> The question is whether we want to:
> 
> * Always add this property when serializing, and require it on all blocks when deserializing.
> * Always add this property to shadow blocks when serializing, and require it on shadow blocks when deserializing.
> * Not add this property when serializing, and add it _during_ deserialization to track the shadow/non-shadowness.
> 
> Or maybe there's another even better alternative :D
> 
> [Edit: Maybe when we set a block's shadow state we could also traverse all of the children to see if one of them is a non-shadow... Also note that the main reason this is a problem because blocks are connected to each other from leaf nodes up to child nodes, so blocks are connected to their parents before we know if the parent is a shadow]

maribethb (approved):

PR #5262, opened 2021-07-30
More cleanup of blockly.js

rachel-fenichel:
## The basics

- [x] I branched from goog_module
- [x] My pull request is against goog_module

### Resolves

Part of #5208 and #5026.

### Proposed Changes

- Move `Blockly.draggingConnections` into `common.js` with a getter and a setter.
  - Update all uses
  - This was previously package, so I am not leaving an alias in `blockly.js`.
-  Add deprecation warnings on setters and getters for `Blockly.mainWorkspace` (followup from #5244)
- Move `Blockly.parentContainer` into `common.js` with a getter and setter
  - Update all internal uses
  - This was previously public, and I think is necessary for injection inside some frameworks, so I left an alias behind in `blockly.js`.
- Delete two old functions that we were using to warn about an old deprecation: `checkBlockColourConstant_` and `checkBlockColourConstants`
  - Delete the only call, in `inject.js`
- Remove an unused property, `Blockly.cache3dSupported_`. It was internal and I don't see any uses of it in core.


core/common.js
```diff
@@ -44,3 +46,58 @@ const setMainWorkspace = function(workspace) {
...
+ * @param {!Element} newParent The container element.
+ */
+const setParentContainer = function(newParent) {
+  parentContainer = newParent;
+};
+exports.setParentContainer = setParentContainer;
+
+/**
+ * All of the connections on blocks that are currently being dragged.
+ * @type {!Array.<!Connection>}
```
NeilFraser, line 80: We use {!Array<!Connection>} now, rather than {!Array.<!Connection>}
There was a mass change a couple of months ago as the dots were eradicated across the codebase.
Here and twice below.

NeilFraser (approved):

core/common.js
```diff
@@ -72,3 +74,30 @@ const setParentContainer = function(newParent) {
...
+
+/**
+ * Set the list of dragging connections.
+ * @param {!Array.<!Connection>} connectionList The new list of connections on
+ *     blocks that are currently being dragged.
+ */
+const setDraggingConnections = function(connectionList) {
+  draggingConnections = connectionList;
+};
+exports.setDraggingConnections = setDraggingConnections;
```
cpcallen, line 103: This API is fine, but—noting that the styleguide does _not_ prohibit _modifying the properties of an exported object_ (only modifying `exports` itself)—a simpler alternative would be to export the array itself:

```
/**
 * All of the connections on blocks that are currently being dragged.
 * @type {!Array<!Connection>}
 */
exports.draggingConnections = [];
```

And then replace calls to `Blockly.common.setDraggingConnections([])` with a simple assignment `Blockly.common.draggingConnections.length = 0;`, and modify the code in `Blockly.BlockSvg.prototype.setDragging` to append, i.e., replace:

```JS
    var newList =
        Blockly.common.getDraggingConnections().concat(this.getConnections_(true));
    Blockly.common.setDraggingConnections(newList);
```

with:
```JS
    Blockly.common.draggingConnections.push(...this.getConenctions_(true));
```
This is marginally less encapsulated, but it's definitely less complex, and the reduction in number of function calls and array allocations might make it a bit faster too.

cpcallen (approved):
LGTM as-is, but I have made one suggestion about an alternative approach which you may wish to consider.

tests/deps.js
```diff
@@ -172,8 +172,13 @@ goog.addDependency('../../core/toolbox/category.js', ['Blockly.ToolboxCategory']
...
 goog.addDependency('../../core/toolbox/toolbox.js', ['Blockly.Toolbox'], ['Blockly.BlockSvg', 'Blockly.CollapsibleToolboxCategory', 'Blockly.ComponentManager', 'Blockly.Css', 'Blockly.DeleteArea', 'Blockly.Events', 'Blockly.Events.ToolboxItemSelect', 'Blockly.IAutoHideable', 'Blockly.IKeyboardAccessible', 'Blockly.IStyleable', 'Blockly.IToolbox', 'Blockly.Options', 'Blockly.Touch', 'Blockly.browserEvents', 'Blockly.constants', 'Blockly.registry', 'Blockly.utils', 'Blockly.utils.Rect', 'Blockly.utils.aria', 'Blockly.utils.dom', 'Blockly.utils.toolbox'], {'lang': 'es5'});
+<<<<<<< HEAD
 goog.addDependency('../../core/toolbox/toolbox_item.js', ['Blockly.ToolboxItem'], ['Blockly.IToolboxItem', 'Blockly.utils.IdGenerator'], {'lang': 'es6', 'module': 'goog'});
 goog.addDependency('../../core/tooltip.js', ['Blockly.Tooltip'], ['Blockly.browserEvents', 'Blockly.utils.string']);
+=======
+goog.addDependency('../../core/toolbox/toolbox_item.js', ['Blockly.ToolboxItem'], ['Blockly.IToolboxItem']);
+goog.addDependency('../../core/tooltip.js', ['Blockly.Tooltip'], ['Blockly.browserEvents', 'Blockly.common', 'Blockly.utils.string']);
+>>>>>>> c62fba04 (Move set/getParentContainer into common)
```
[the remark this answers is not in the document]
rachel-fenichel: >.< Regenerated.

core/common.js
rachel-fenichel: Done.

core/common.js
```diff
@@ -44,3 +44,31 @@ const setMainWorkspace = function(workspace) {
...
+const getParentContainer = function() {
+  return parentContainer;
+};
+exports.getParentContainer = getParentContainer;
+
+/**
+ * Set the parent container.  This is the container element that the WidgetDiv,
+ * DropDownDiv, and Tooltip are rendered into the first time `Blockly.inject`
+ * is called.
+ * This method is a NOP if called after the first ``Blockly.inject``.
```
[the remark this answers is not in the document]
rachel-fenichel: Each div is created once--subsequent calls to their `createDom` functions just return immediately.

Philosophically, I think there's a question about whether it's a NOP if it *does* change something, but that something is never read again.

So, yes. The API is kludgy. `parentContainer` is modified but won't be looked at again.

PR #5276, opened 2021-08-02
feat: adds hooks for serializing plugins

BeksOmega:
## The basics

### Resolves

Work on project cereal
Dependent on #5246

### Proposed Changes
Adds the ability to register plugin serializers so that the serialization format is extensible.

### Reason for Changes
Support for cool future plugins like https://github.com/google/blockly-samples/issues/827

### Test Coverage
Added a bunch of tests to the registry because it was missing coverage. Also added a test to ensure that plugin state is cleared and loaded in the correct order based on priority.

Other functionality (like the block and variable serializers being called correctly) is covered by existing tests.


core/serialization/blocks.js
```diff
@@ -0,0 +1,649 @@
...
+   */
+  save(workspace) {
+    const blockState = [];
+    for (const block of workspace.getTopBlocks()) {
+      // TODO: JavaScript class semantics make this call confusing. Any way to
+      //    make this clearly not recursive?
+      const state = save(block, {addCoordinates: true, addNextBlocks: true});
+      if (state) {
+        blockState.push(state);
+      }
```
BeksOmega, line 610: Looking for opinions on this.

core/serialization/serialization.js
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Returns the state of the workspace as a plain JavaScript object.
+ * @param {!Blockly.Workspace} workspace The workspace to serialize.
+ * @return {!Object<string, *>} The serialized state of the workspace.
+ */
+const save = function(workspace) {
+  const state = Object.create(null);
+
+  const serializerMap = registry.getAllItems(registry.Type.PLUGIN_SERIALIZER);
```
BeksOmega, line 37: It may be a good idea to add some wrappers instead of using the registry directly. Eg `serialization.workspaces.registerSerializer`. That way if we decide the registry doesn't make sense for plugin serializers we can easily change it.

core/serialization/variables.js
```diff
@@ -0,0 +1,132 @@
...
+const save = function(variableModel) {
+  const state = Object.create(null);
+  state['name'] = variableModel.name;
+  state['id'] = variableModel.getId();
+  if (variableModel.type) {
+    state['type'] = variableModel.type;
+  }
+  return state;
+};
+exports.save = save;
```
BeksOmega, line 56: I think I can stop exporting this now, because it's wrapped in the plugin serializer. This is perfect because I didn't want these to be public anyway.

core/registry.js
```diff
@@ -273,6 +278,31 @@ Blockly.registry.getObject = function(type, name, opt_throwIfMissing) {
...
+ *     (e.g. Category)
+ * @param {boolean=} opt_throwIfMissing Whether or not to throw an error if we
+ *     are unable to find the object.
+ * @return {?Object<string, ?T|?function(new:T, ...?)>} A map of objects with
+ *     the given type, or null if none exists.
+ * @template T
+ */
+Blockly.registry.getAllItems = function(type, opt_throwIfMissing) {
+  type = String(type).toLowerCase();
+  var typeRegistry = Blockly.registry.typeMap_[type];
```
BeksOmega, line 293: Hmm there's a bit of a problem with this. When you register an item the case is removed (not sure why), so all of the keys in this map are caseless. This is incorrect for the JSON serializer because according to the [styleguide](https://google.github.io/styleguide/jsoncstyleguide.xml#Property_Name_Format), key names should be camelCased.

We could:
  1) Stop removing the case from keys.
  2) Keep a separate map the maintains casing.
  3) Require developers to supply the names of their serializers twice. Once in registration, and once in the definition of their serializer.

core/serialization/i_plugin_serializer.js
```diff
@@ -0,0 +1,70 @@
...
+  /* eslint-disable no-unused-vars, valid-jsdoc */
+
+  /**
+   * Saves the state of the plugin.
+   * @param {!Workspace} workspace The workspace the plugin to serialize is
+   *     associated with.
+   * @return {?Object} A JS object containing the plugin's state, or null if
+   *     there is no state to record.
+   */
+  save(workspace) { }
```
alschmiedt, line 48: I think since this is a record you don't need the `{}`

core/serialization/workspaces.js
```diff
@@ -28,31 +27,13 @@ const variables = goog.require('Blockly.serialization.variables');
...
-  const blockStates = [];
-  for (let block of workspace.getTopBlocks(false)) {
-    const blockState = blocks.save(block, {addCoordinates: true});
-    if (blockState) {
-      blockStates.push(blockState);
+  const serializerMap = registry.getAllItems(registry.Type.PLUGIN_SERIALIZER);
+  for (const key in serializerMap) {
+    const save = serializerMap[key].save(workspace);
+    if (save) {
+      state[key] = save;
```
alschmiedt, line 34: Does this mean for the blocks this would look like
```
'blocks': {
   languageVersion: 0, 
   blocks: [...]
}

**Review by alschmiedt — APPROVED** [review]

File: core/registry.js (hunk 4 above)
Comment (BeksOmega) [reply core/registry.js:293]: > I think one of the reasons we did this was that it slightly decreases the chance of mistakes for developers who have to type in the plugin name to the options struct.

Ah that makes sense. And I guess it also means that we can't go with 1, without breaking people who *are* accessing registered items using the caseless name. And we don't even throw errors by default, so no help there.

But I'm not sure if breaking people here is a big deal or not.

File: core/serialization/i_plugin_serializer.js [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+
+// eslint-disable-next-line no-unused-vars
+const Workspace = goog.requireType('Blockly.Workspace');
+
+
+/**
+ * Serializes and deserializes a plugin.
+ * @record
+ */
+class IPluginSerializer {
```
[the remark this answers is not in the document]
BeksOmega: I think that's a great idea!

core/serialization/i_plugin_serializer.js
```diff
@@ -0,0 +1,70 @@
...
+goog.module('Blockly.serialization.IPluginSerializer');
+goog.module.declareLegacyNamespace();
+
+// eslint-disable-next-line no-unused-vars
+const Workspace = goog.requireType('Blockly.Workspace');
+
+
+/**
+ * Serializes and deserializes a plugin.
+ * @record
```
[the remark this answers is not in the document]
BeksOmega: Yesh, will do o7

core/serialization/i_plugin_serializer.js
```diff
@@ -0,0 +1,70 @@
...
+     * priorities. Eg if you have priorities (0, -10, 10, 100) the order of
+     * deserialiation will be (100, 10, 0, -10).
+     * If two plugins have the same priority, they are deserialized in an
+     * arbitrary order relative to eachother.
+     * @type {number}
+     */
+    this.priority;
+  }
+
+  /* eslint-disable no-unused-vars, valid-jsdoc */
```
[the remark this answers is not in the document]
BeksOmega: `no-unused-vars` is because of the unused function parameters. `valid-jsdoc` is because `save` has an `@return` annotation, but no `return` statement.

core/serialization/blocks.js
```diff
@@ -570,3 +574,70 @@ const initBlock = function(block, rendered) {
...
+        'blocks': blockState
+      };
+    }
+    return null;
+  }
+
+  /**
+   * Deserializes the blocks defined by the given state into the given
+   * workspace.
+   * @param {{languageVersion: number, blocks:!Array<!State>}} state The state
```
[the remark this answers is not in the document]
BeksOmega: I think literals are considered [non-null by default](https://google.github.io/styleguide/jsguide.html#jsdoc-nullability).

core/serialization/i_plugin_serializer.js
BeksOmega: Hmm it looks like you need them for non-static methods, but not static methods ([5.4.9](https://google.github.io/styleguide/jsguide.html#features-classes-interfaces)) And also there's not supposed to be an extra space, so thanks for the catch :D

core/serialization/workspaces.js
BeksOmega: Yes it does.

core/serialization/blocks.js
```diff
@@ -273,7 +277,7 @@ const saveConnection = function(connection) {
   }
   var state = Object.create(null);
   if (shadow) {
-    state['shadow'] = Xml.domToText(shadow)
+    state['shadow'] = Blockly.Xml.domToText(shadow)
```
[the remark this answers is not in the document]
BeksOmega: Fixed by rebase o7

BeksOmega:
I'm getting some errors from the closure compiler about my `@interface` annotations, and I'm not sure why:
```
gulp-google-closure-compiler: core-serialization-blocks.js:594:0: ERROR - [JSC_RESOLVED_TAG_EMPTY]
    Could not resolve type in @implements tag of module$contents$Blockly$serialization$blocks_BlockSerializer
  594| class BlockSerializer {
       ^^^^^^^^^^^^^^^^^^^^^^^
  595|   constructor() {
       ^^^^^^^^^^^^^^^^^
...
  649|   }
       ^^^
  650| }
       ^

core-serialization-variables.js:82:0: ERROR - [JSC_RESOLVED_TAG_EMPTY]
    Could not resolve type in @implements tag of module$contents$Blockly$serialization$variables_VariableSerializer
   82| class VariableSerializer {
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
   83|   constructor() {
       ^^^^^^^^^^^^^^^^^
...
  124|   }
       ^^^
  125| }

```

The tags look like this:
```js
// eslint-disable-next-line no-unused-vars
const ISerializer = goog.requireType('Blockly.serialization.ISerializer');

// ...

/**
 * Serializer for saving and loading variable state.
 * @implements {ISerializer}
 */
class VariableSerializer { /* ... */ }
```

core/registry.js
alschmiedt: I'm not sure it is a big deal. I think probably the subset of people that are using this feature and that have incorrectly capitalized is probably pretty small. We should add it to the release notes since it is a change in the public API, but I don't think it is worth complicating the code (2) or adding extra work for the developer (3).

core/serialization/serialization.js
alschmiedt: Sorry I didn't see this earlier. Yes, I think that is a great idea.

alschmiedt:
Ok so I think the interface problem is coming from how `ISerializer` is getting imported. If you want to do named exports instead of default exports then I think you will need to import using the below.
``` 
const {ISerializer} = goog.requireType('Blockly.serialization.ISerializer');`.
```
I think this solved the original problem you posted about, but it uncovered a few more type errors that will need to be addressed.

BeksOmega:
> Ok so I think the interface problem is coming from how `ISerializer` is getting imported. If you want to do named exports instead of default exports then I think you will need to import using the below.
> 
> ```
> const {ISerializer} = goog.requireType('Blockly.serialization.ISerializer');`.
> ```
> 
> I think this solved the original problem you posted about, but it uncovered a few more type errors that will need to be addressed.

Oh dur! That makes sense. Thank you for the help :D

core/serialization/i_serializer.js
```diff
@@ -0,0 +1,71 @@
...
+  /* eslint-disable no-unused-vars, valid-jsdoc */
+
+  /**
+   * Saves the state of the plugin.
+   * @param {!Workspace} workspace The workspace the plugin to serialize is
+   *     associated with.
+   * @return {?} A JS object containing the plugin's state, or null if
+   *     there is no state to record.
+   */
+  save(workspace) {}
```
alschmiedt, line 48: > Hmm it looks like you need them for non-static methods, but not static methods (5.4.9) 

Ah, I didn't realize! I think so far we have been following [this example](https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler#nominal-interfaces)(possibly outdated?). I tried something similar with the class syntax and the compiler was unthrilled. I think it is fine for right now, but if we end up going back to using es5 syntax I would prefer this matched the style of our other interfaces.

alschmiedt (commented):
I missed it the first time around, but we are using some es6 classes here. I think we plan on eventually making that jump, but according to [this issue](https://github.com/google/blockly/issues/5026) "We are not proposing to convert ES5 classes to ES6 class syntax at the present time.]". However, I'm not sure how we feel about entirely new classes using this syntax. Could you add it to your list to check in with Chris or Rachel when they return?

core/serialization/i_serializer.js
```diff
@@ -0,0 +1,70 @@
+/**
```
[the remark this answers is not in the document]
BeksOmega: We could, but to conform to the [styleguide](https://google.github.io/styleguide/jsguide.html#naming-hierarchy) we'd have to rename it to `Blockly.interfaces.ISerializer`. Personally I think it makes more sense to keep all of the serialization stuff in the same module, but there's probably something I'm unaware of here haha.

BeksOmega:
> Could you add it to your list to check in with Chris or Rachel when they return?

I know I asked about building it as a module or using requires, and they said to use a module. But I'm not sure if I asked about the class syntax or not. I'll definitely bug Christopher about that on Monday. And if I have to switch it around that's totally cool =)

core/registry.js
BeksOmega: So I tried to implement (1) and I don't think it's going to be workable :/ The issue is that the toolbox uses tag names from XML to access things in the registry, and we can't rely on the case of tag names due to IE ([relevant code](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/utils/toolbox.js#L345)).

So given that, I'm going to attempt to implement (2) instead.

BeksOmega: Got (2) working!

core/serialization/i_serializer.js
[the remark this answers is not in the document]
alschmiedt: Yeah that is good point. I think this a place where we currently deviate from the style guide. I guess it depends on whether we plan on moving towards the pattern the style guide recommends in the future or not. I'm not entirely sure what the plan for interfaces is with the ongoing conversion. We can leave this and add it to the list of things to check with Chris or Rachel on. If we don't plan on changing anything this quarter my vote would be to place it in the interface directory and ignore that style guide rule. (Since that is what we are doing for the other interfaces.) Then when we decide if we want to update the interfaces we can change them all at the same time.

alschmiedt (approved):

cpcallen:
> I missed it the first time around, but we are using some es6 classes here. I think we plan on eventually making that jump, but according to [this issue](https://github.com/google/blockly/issues/5026) "We are not proposing to convert ES5 classes to ES6 class syntax at the present time."

I wrote that to clarify that we are not planning to refactor existing ES5 class definitions to ES6 `class` syntax as part of the first phase of the migration.  As noted in conversation previously, it is my understanding that new code is to use the new styleguide—and `class`—but it would indeed be wise to confirm this with @rachel-fenichel.

PR #5278, opened 2021-08-02
Update build artifact sizes in check_metadata.sh

gonfunko:
This updates the build artifact sizes in check_metadata.sh, since (due to the goog.module conversion) these have now grown by more than 10% and are causing test failures.


rachel-fenichel (approved):
Please add a note that this was a mid-quarter change due to modules.

PR #5281, opened 2021-08-02
Migrate core/events/events_selected.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/events/events_selected.js` to `goog.module` with ES6 `const`/`let`.


core/events/events_selected.js
```diff
@@ -26,12 +27,11 @@ goog.require('Blockly.utils.object');
...
  * @param {string=} opt_workspaceId The workspace identifier for this event.
  *    Null if no element previously selected. Undefined for a blank event.
- * @extends {Blockly.Events.UiBase}
+ * @extends {UiBase}
  * @constructor
  */
-Blockly.Events.Selected = function(opt_oldElementId, opt_newElementId,
-    opt_workspaceId) {
-  Blockly.Events.Selected.superClass_.constructor.call(this, opt_workspaceId);
+const Selected = function(opt_oldElementId, opt_newElementId, opt_workspaceId) {
```
alschmiedt, line 33: We don't have to do it right now, but I think this falls under the category of classes that could use a rename. I think `SelectedEvent` makes a lot more sense.

alschmiedt (approved):

core/events/events_selected.js
gonfunko: +1, that would definitely make it clearer. Added a note to #5073 so we don't forget to revisit this.

PR #5287, opened 2021-08-02
Migrate core/options.js to goog.module syntax 

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/options.js` to `goog.module` syntax. 

### Additional Information

This PR also refactors `Blockly.BlocklyOptions` our of `Blockly.Options` into a separate module in `core/blockly_options.js` and adds missing requires to files that use `Blockly.BlocklyOptions`.
In some cases, unused `Blockly` requires were removed (under the assumption that they were added by mistake thinking they were needed for `Blockly.BlocklyOptions`


core/options.js
```diff
@@ -31,57 +31,57 @@ goog.requireType('Blockly.WorkspaceSvg');
...
+  let maxTrashcanContents = options['maxTrashcanContents'];
+  if (hasTrashcan) {
+    if (maxTrashcanContents === undefined) {
+      maxTrashcanContents = 32;
+    }
+  } else {
+    maxTrashcanContents = 0;
+  }
+  const rtl = !!options['rtl'];
+  let horizontalLayout = options['horizontalLayout'];
```
cpcallen, line 81: Did you move `maxTrashcanContents` in order to alphabetise the setting of the various local variables?  If so then shouldn't it go _after_ `horizontalLayout`?

core/blockly_options.js
```diff
@@ -10,12 +10,15 @@
  */
 'use strict';
 
-goog.provide('Blockly.BlocklyOptions');
+goog.module('Blockly.BlocklyOptions');
+goog.module.declareLegacyNamespace();
 
 /**
```
cpcallen: Shouldn't there be two blank lines here?

cpcallen (approved):
Two small nits.

core/options.js
moniika: I moved it out of the if statement so that `maxTrashcanContents` would always be defined. 

Before, if `readOnly = true`, then `maxTrashcanContents = undefined`.
Now, if `readOnly = true`, then `maxTrashcanContents = 0` (because `hasTrashcan = false`.

I don't expect this to change any behavior, but it fixes typing issues with `maxTrashcanContents`.

cpcallen: Oh yes, quite so.  Apologies for my misreading.

PR #5296, opened 2021-08-03
Migrate core/contextmenu.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/contextmenu.js` to `goog.module` syntax. 

 ### Additional Information

Adds JS accessor properties for `Blockly.ContextMenu.


core/block_svg.js
```diff
@@ -921,7 +921,7 @@ Blockly.BlockSvg.prototype.dispose = function(healStack, animate) {
     this.workspace.cancelCurrentGesture();
   }
   // If this block has a context menu open, close it.
-  if (Blockly.ContextMenu.currentBlock == this) {
+  if (Blockly.ContextMenu.getCurrentBlock == this) {
```
NeilFraser, line 924: Shouldn't this be a function call?  As it stands, it will always return false.

core/contextmenu.js
```diff
@@ -10,93 +10,118 @@
...
-goog.require('Blockly.Xml');
-
-goog.requireType('Blockly.Block');
-goog.requireType('Blockly.WorkspaceSvg');
 
 
 /**
  * Which block is the context menu attached to?
- * @type {Blockly.Block}
+ * @type {Block}
```
NeilFraser, line 40: I think all three of these should be {?Block} since the '?' is no longer optional under the new style guide.

NeilFraser (approved):
LGTM after fixing the function call.

PR #5302, opened 2021-08-04
Migrate core/events/events.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/events/events.js` to `goog.module` with ES6 `const`/`let`.

### Additional Information
This uses goog.forwardDeclare/goog.module.get to work around circular requires with Blockly.Workspace.


core/events/events.js
```diff
@@ -218,101 +303,99 @@ Blockly.Events.BumpEvent;
...
  * @const
  */
-Blockly.Events.BUMP_EVENTS = [
-  Blockly.Events.BLOCK_CREATE,
-  Blockly.Events.BLOCK_MOVE,
-  Blockly.Events.COMMENT_CREATE,
-  Blockly.Events.COMMENT_MOVE
-];
+const BUMP_EVENTS = [BLOCK_CREATE, BLOCK_MOVE, COMMENT_CREATE, COMMENT_MOVE];
+exports.BUMP_EVENTS = BUMP_EVENTS;
```
NeilFraser, line 307: Would it be evil to just say:
```
exports.BUMP_EVENTS = [exports.BLOCK_CREATE, exports.BLOCK_MOVE, exports.COMMENT_CREATE, exports.COMMENT_MOVE]
```
Then 27 different `const` statements could disappear.

core/events/events.js
```diff
@@ -218,101 +303,99 @@ Blockly.Events.BumpEvent;
...
- * @param {!Array<!Blockly.Events.Abstract>} queueIn Array of events.
+ * @param {!Array<!Abstract>} queueIn Array of events.
  * @param {boolean} forward True if forward (redo), false if backward (undo).
- * @return {!Array<!Blockly.Events.Abstract>} Array of filtered events.
+ * @return {!Array<!Abstract>} Array of filtered events.
  */
-Blockly.Events.filter = function(queueIn, forward) {
-  var queue = queueIn.slice();  // Shallow copy of queue.
+const filter = function(queueIn, forward) {
+  let queue = queueIn.slice();  // Shallow copy of queue.
```
NeilFraser, line 359: I think `queue` is `const`.  The reverse is an in place operation.

core/events/events.js
```diff
@@ -14,201 +14,286 @@
...
 
 /**
  * Allow change events to be created and fired.
  * @type {number}
- * @private
  */
-Blockly.Events.disabled_ = 0;
+let disabled = 0;
+/** @private */
+exports.disabled_ = disabled;
```
NeilFraser, line 94: This line appears to be clobbered by the setter below.  Delete?

core/events/events.js
```diff
@@ -14,201 +14,286 @@
...
+
+
+Object.defineProperties(exports, {
+  disabled_: {
+    set: function(newValue) {
```
NeilFraser, line 99: Who calls this setter?  Seems odd to export a private property.  I don't see anyone using this in our codebase in `develop`.

core/events/events.js
```diff
@@ -14,201 +14,286 @@
...
  * Not to be confused with bumping so that disconnected connections do not
  * appear connected.
- * @typedef {!Blockly.Events.BlockCreate|!Blockly.Events.BlockMove|
- * !Blockly.Events.CommentCreate|!Blockly.Events.CommentMove}
+ * @typedef {!BlockCreate|!BlockMove|
+ * !CommentCreate|!CommentMove}
  */
-Blockly.Events.BumpEvent;
+let BumpEvent;
+exports.BumpEvent = BumpEvent;
```
NeilFraser, line 296: This is a typedef, does this need to be a `let` variable?

NeilFraser (approved):

core/events/events.js
gonfunko: It's used by test_helpers.js and workspace_test.js unfortunately.

core/events/events.js
gonfunko: The compiler complains if it's made a `const` because it isn't being assigned a value.

core/events/events.js
gonfunko: It gets reassigned on line 409.

core/events/events.js
gonfunko: Closure isn't able to see properties that get added by Object.defineProperties, so this is here solely to inform the compiler that a .disabled_ property exists.

core/events/events.js
gonfunko: I don't think I quite understand the suggestion here, could you clarify?

PR #5306, opened 2021-08-04
Migrate core/renderers/geras/path_object.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/renderers/geras/path_object.js` to `goog.module` with ES6 `const`/`let`.


core/renderers/geras/path_object.js
```diff
@@ -14,30 +14,32 @@
...
 goog.module.declareLegacyNamespace();
 
-goog.require('Blockly.blockRendering.PathObject');
-goog.require('Blockly.geras.ConstantProvider');
-goog.require('Blockly.Theme');
-goog.require('Blockly.utils.colour');
-goog.require('Blockly.utils.dom');
-goog.require('Blockly.utils.object');
-goog.require('Blockly.utils.Svg');
+const BasePathObject = goog.require('Blockly.blockRendering.PathObject');
```
cpcallen, line 17: Nit: [the styleguide says](https://google.github.io/styleguide/jsguide.html#file-goog-require):

> Aliases must match the final dot-separated component of the imported module's namespace.
>
> **Exception:** In certain cases, additional components of the namespace can be used to form a longer alias.

Strictly speaking, "base" is not a component of "Blockly.blockRendering.PathObject"— but I'm not convinced that `BlockRenderingPathObject` would actually be preferable, so I'll let you decide whether you think it improves readability to change it or not.

cpcallen (approved):

core/renderers/geras/path_object.js
gonfunko: Yeah, I think I prefer it as-is - I agree that "BlockRendering" doesn't really clarify it, whereas this makes it clear it's just a shared base class.

PR #5310, opened 2021-08-04
[DONT MERGE] Project cereal upgrader script

BeksOmega:
## The basics

- [X] My code follows the python styleguide (I think)

### Resolves
Work on project cereal.

### Proposed Changes
Adds a python script which walks over all of the files in a director and adds default implementations of the new JSO serialization hooks where it finds old XML hooks.

### Reason for Changes
Our previous idea for backwards compatibility was to build it into core itself, but that ran into two problems:
1) For fields, the plan was to add default backwards-compatibles implementations of the new hooks to the base Field class. These implementations would just call the corresponding xml hooks, and stringify the result.
    The problem with this method is that if you subclassed a field besides Field (eg FieldDropdown, FieldNumber, or FieldTextInput) you would hit *that* field's implementation of `saveState`/`loadState` rather than the base backwards-compatible implementation.
2) For blocks, the plan was to add special handling inside the block serialization system to stringify the values returned by XML hooks if necessary.
    The problem with this method is that it ties the new system to the old system, making the old system harder to remove in the future. It also makes the new system messy.

### Testing
I created a test_defs.js file which includes different methods for specifying blocks and fields. This includes js definitions and typescript definitions.

The upgrader included in the PR handles *almost* all of the cases. The cases it misses are:
```typescript
// JS Block def that points to other def.
Blockly.Bocks['test2'] = {
  mutationToDom: Blockly.Blocks['text_prompt_ext'].mutationToDom,
  domToMutation: Blockly.Blocks['text_prompt_ext'].domToMutation
}

// Typescript interface definition.
interface Whatever {
  mutationToDom(this: Whatever): () => Element,
  domtoMutation(this: Whatever): (Element) => void,
} 
```

### Additional Info
If anyone has other suggestions for how to handle backwards compatibility, I'm totally open to investigating them as well!

This code should not be merged into core. Instead the script should be included with the release notes when the new serializer is published.

### Reading this PR
The first commit contains the test_defs.js file with its initial definitions.
The second commit contains the upgrader script.
The third commit contains the modifications the upgrader script made to the test_defs.js file.

### DIFF
--- a/core/serialization/test_defs.js
+++ b/core/serialization/test_defs.js
@@ -5,8 +5,14 @@ Blockly.Blocks['test1'] = {
   mutationToDom: function() {
 
   },
+  saveExtraState: function() {
+    return Blockly.Xml.domToText(this.mutationToDom());
+  },
   domToMutation: function(mutation) {
 
+  },
+  loadExtraState: function(state) {
+    return this.domToMutation(Blockly.Xml.textToDom(state));
   },
 }
 
@@ -15,8 +21,14 @@ Blockly.Blocks['test3'] = {
   'mutationToDom': function() {
 
   },
+  'saveExtraState': function() {
+    return Blockly.Xml.domToText(this.mutationToDom());
+  },
   'domToMutation': function(mutation) {
 
+  },
+  'loadExtraState': function(state) {
+    return this.domToMutation(Blockly.Xml.textToDom(state));
   },
 };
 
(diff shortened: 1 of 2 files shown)


BeksOmega:
@alschmiedt and @maribethb for opinions on if this is the right way to handle backwards compatibility.

If you peoples think it will work, then Aaron said he will review the python code for me :D

PR #5313, opened 2021-08-04
cleanup: Change type defs for state types to records

BeksOmega:
## The basics

Link for Diff: https://github.com/BeksOmega/blockly/compare/cereal/plugin-hooks...BeksOmega:cereal/types
### Resolves

Work on project cereal
Dependent on #5276

### Proposed Changes
Changes the type declarations for state types (ie block state and variable state) to use class declarations and `@record` annotations, as recommended by the styleguide. See [5.4.9](https://google.github.io/styleguide/jsguide.html#features-classes-interfaces) and the bottom of [7.7](https://google.github.io/styleguide/jsguide.html#jsdoc-enum-and-typedef-comments)

### Reason for Changes
Styleguide conformity. Plus it allows for more explicit docs.

### Test Coverage
Not necessary in this case.


core/registry.js
```diff
@@ -117,7 +122,7 @@ Blockly.registry.Type.BLOCK_DRAGGER =
  *     an already registered item.
  * @throws {Error} if the type or name is empty, a name with the given type has
  *     already been registered, or if the given class or object is not valid for
- * it's type.
+ *     it's type.
```
maribethb, line 125: the nitpickiest nit ever: `its type` (i think this commit is in a different PR, sorry)

core/serialization/blocks.js
```diff
@@ -19,47 +19,154 @@ const Block = goog.requireType('Blockly.Block');
...
+    this.x;
+
+    /**
+     * The position of the block along the y axis.
+     * @type {(number|undefined)}
+     */
+    this.y;
+
+    /**
+     * Whether the block so collapsed or not. True if it is collapsed, undefined
```
maribethb, line 87: s/so/is

core/serialization/blocks.js
```diff
@@ -19,47 +19,154 @@ const Block = goog.requireType('Blockly.Block');
...
+    /**
+     * Whether the block is inline or not. True if it is inlined, false if it
+     * is not. Undefined if the inlined state matches the block's default
+     * inlined state (depends on the particular block).
+     * @type {(boolean|undefined)}
+     */
+    this.inline;
+
+    /**
+     * An extra string that gets round-tripped through serialiation. Often used
```
maribethb, line 130: typo: serialization

core/serialization/blocks.js
```diff
@@ -19,47 +19,154 @@ const Block = goog.requireType('Blockly.Block');
...
+     * for associating blocks with external assets.
+     * @type {(string|undefined)}
+     */
+    this.data;
+
+    /**
+     * The extra state associated with the block.
+     * In the past this was called a "mutation".
+     * @type {(*|undefined)}
```
maribethb, line 139: Don't really think you need the `|undefined` since `*` means `ALL` which includes `undefined`.

core/serialization/blocks.js
```diff
@@ -19,47 +19,154 @@ const Block = goog.requireType('Blockly.Block');
...
+     */
+    this.extraState;
+
+    /**
+     * The state of the icons attached to this block.
+     * @type {(!Object<string, *>|undfined)}
```
maribethb, line 145: typo: undefined

One would hope that this kind of incorrect type would have emitted an error somewhere... :/

core/serialization/blocks.js
```diff
@@ -19,47 +19,154 @@ const Block = goog.requireType('Blockly.Block');
...
+     */
+    this.icons;
+
+    /**
+     * The state of the fields attached to this block.
+     * @type {(!Objec<string, *>|undefined)}
```
maribethb, line 151: typo: Object

core/serialization/blocks.js
maribethb: Okay, phew, the tests did fail. I was worried this somehow would have gotten through silently!

core/serialization/blocks.js
BeksOmega: Weird, what you said makes sense so I removed it, but it gave me a compiler error. Maybe `|undefined` isn't required in general, but it is when specifying optional properties?

BeksOmega:
Ok so, this has run into a bit of an issue. Closure does not allow you to access properties of `@struct`s using `['']` notation. And it seems all `@record`s are also `@struct`s and there's no way to fix that.

I want to use an `@record` type so that we can provide detailed type information in the JSDoc and typescript types. But I also want to use `['']` notation so that the closure compiler doesn't rename the properties (since they could be coming from JSON).

I'm really not sure what to do here. I did find [one recommendation](https://stackoverflow.com/a/7834912), which was to basically wrap any JSON you recieve in a new object with new rename-able properties. But I think that for that to work, we would have to change the parameter to the `load` function to be an `!Object`, which doesn't give us the benefits of the JSDoc or typescript types.
```js
/**
 * @param {!Object} state // No benefit from types.
 */
const load = function(state, workspace) {
  const typedState = {
    // this property could be renamed by the compiler.
    // but it is still compatible with JSON
    id: state['id'],
    // etc...
  }
}
```
The only option I can think of is to not merge this, and stick with the workable, but less informative, typedefs. @maribethb  thoughts?

maribethb (commented):
So, yes, I think for now the best bet is to stick with typedefs. (I wouldn't take the advice of that StackOverflow post without further research because it's from 2011 (!) and `@struct/@dict` annotations weren't added at all until 2012.) I was hoping you could just throw a `@dict` or an `@unrestricted` annotation onto the record, but apparently that's not the case :/

I think the actual correct thing to do in an ideal world is to have our external API defined in externs (here is where I would add a link to the internal JS Best Practices site) so that the compiler does not rename these properties. After all, these objects aren't really behaving like dictionaries. We just treat them that way as a convenient way to turn off property renaming by the compiler. But I don't think we currently supply any externs and I don't think you should have to figure that out as part of this project so I think going back to a typedef and using bracket access is fine and matches what we already have in the xml serializers.

BeksOmega:
Closing since records don't work with the compiler in this case and externs are hard.

PR #3335, opened 2019-10-24
Resolve remaining compiler warnings with visibility issues

samelhusseini:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

Resolve remaining compiler warnings with visibility issues

### Proposed Changes

Added wrappers around blockDb_ to set a block and remove it.
Renamed block / comment.``moveToDragSurface_`` to ``moveToDragSurface`` and made it @package.
Renamed block.``showHelp_`` to ``showHelp`` and made it @package.
Renamed Blockly.``duplicate_`` to ``duplicate`` and made it @package.
Renamed Blockly.``longStart_`` to ``longStart`` and made it @package.

### Test Coverage

Tested on:

PR #1817, opened 2018-04-23
Comments v3

rachel-fenichel:
## The basics

- [ ] I branched from develop
- [ ] My pull request is against develop
- [ ] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Test Coverage

Tested on:

### Additional Information

There's an open PR for all of the message files that should make this one have fewer files.

PR #5316, opened 2021-08-05
Add goog.module.get calls for WorkspaceComment and WorkspaceCommentSvg

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Adds `requireType` and `goog.module.get` to retrieve references for optional modules `Blockly.WorkspaceComment` and `Blockly.WorkspaceCommentSvg` in code.

 ### Additional Information

Adding `requireType` does not cause the module to be loaded.
This PR also fixes build compile errors that appear after adding `requireType` for these modules to files.


core/workspace_comment_svg.js
```diff
@@ -173,9 +178,6 @@ WorkspaceCommentSvg.prototype.dispose = function() {
   }
 
   dom.removeNode(this.svgGroup_);
-  // Sever JavaScript to DOM connections.
-  this.svgGroup_ = null;
-  this.svgRect_ = null;
```
moniika, line 178: I'm not sure about this change.

core/workspace_comment_svg.js
```diff
@@ -173,9 +177,6 @@ WorkspaceCommentSvg.prototype.dispose = function() {
...
```
alschmiedt, line 178: Why remove these?

core/workspace_comment_svg.js
```diff
@@ -414,26 +415,9 @@ WorkspaceCommentSvg.prototype.moveToDragSurface = function() {
...
-
-/**
- * Move this comment back to the workspace block canvas.
- * Generally should be called at the same time as setDragging(false).
- * Does nothing if useDragSurface_ is false.
- * @param {!Coordinate} newXY The position the comment should take
- *     on on the workspace canvas, in workspace coordinates.
- * @private
- */
-WorkspaceCommentSvg.prototype.moveOffDragSurface = function(newXY) {
```
alschmiedt, line 430: I see that this is not getting used, but I'm a little uncomfortable removing it without knowing the original reason for leaving it around. 

Also, I would rather we removed this in a PR with a description that contains more information on why we are removing it.

core/workspace_svg.js
```diff
@@ -71,6 +71,8 @@ goog.requireType('Blockly.ScrollbarPair');
 goog.requireType('Blockly.Theme');
 goog.requireType('Blockly.Trashcan');
 goog.requireType('Blockly.VariableModel');
+goog.requireType('Blockly.WorkspaceCommentSvg');
+goog.requireType('Blockly.WorkspaceComment');
```
alschmiedt, line 75: nit: I think this need to be flipped

core/workspace_comment.js
```diff
@@ -103,6 +103,12 @@ const WorkspaceComment = function(workspace, content, height, width, opt_id) {
    */
   this.content_ = content;
 
+  /**
+   * @protected
```
alschmiedt, line 107: Add a description

alschmiedt (approved):

core/workspace_comment_svg.js
moniika: These sets are not necessary and cause compile issues because the type for these two properties is non-nullable (`!`).

Background on why these sets is not necessary is here: https://github.com/google/blockly/issues/2152

core/workspace_comment_svg.js
moniika: This function needs to be removed in order to fix closure-compiler errors.
Leaving it in causes `ERROR - [JSC_UNUSED_PRIVATE_PROPERTY]`.

I looked into it and found that this method was introduced in https://github.com/google/blockly/pull/1817 along with `WorkspaceCommentSvg.moveToDragSurface` and browsing the files at that commit, it seems like neither of those method was ever being called. I think it introduced initially because `BlockDragger` requires `BlockSvg` has those methods, but `BubbleDragger` implements its own version of `moveToDragSurface` and `moveOffDragSurface`.

In addition to this method, `moveToDragSurface` and `clearTransformAttributes_` is dead code. I can remove those two methods as well and see that it's never used. `moveToDragSurface` doesn't cause compile errors because it is `@package` as of https://github.com/google/blockly/pull/3335 (however, I think that change only happened because it shares a name with the method on `BlockSvg`).

Issue #5298, opened 2021-08-03
Update Tag module cleanup workflow to add instead of set labels


moniika:
**Is your feature request related to a problem? Please describe.**

The tag_module_cleanup github action workflow sets the list of labels to an explicit list which clears any existing labels, including those added by the `google-cla` bot.
This is convenient for 2 main reasons:
- This clears any labels added when the PR is created or between pushes and it can be easy to forget to re-add them
- The `google-cla` bot comments on the PR every time the `cla: yes` label is removed

![image](https://user-images.githubusercontent.com/6621618/128097295-78a0a7d1-7f3a-453b-92cd-754cc699e8d9.png)

**Describe the solution you'd like**

Update the github action script to add to the list of labels instead of setting.

**Describe alternatives you've considered**

Remove the set labels functionality and have the workflow only set the milestone. We would have to remember to manually add the `type: cleanup` label, but it would prevent conflicts with the `google-cla` bot.

rachel-fenichel:
Robot fight!
https://media2.giphy.com/media/qCRhaWtcExo9W/giphy.gif?cid=ecf05e472qvvvxzq2w3wkq010jyjazu4465y5m3rqyg62g46&rid=giphy.gif&ct=g

PR #5324, opened 2021-08-06
Add labels instead of setting in github action

moniika:
Update `tag_module_cleanup.yml` logic to add labels instead of setting labels list.

Fixes: https://github.com/google/blockly/issues/5298

### DIFF
--- a/.github/workflows/tag_module_cleanup.yml
+++ b/.github/workflows/tag_module_cleanup.yml
@@ -30,10 +30,9 @@ jobs:
               // https://github.com/google/blockly/milestone/18
               milestone: 18
             })
-            
-            //await github.issues.addLabels({
-            //  owner: context.repo.owner,
-            //  repo: context.repo.repo,
-            //  issue_number: context.issue.number,
-            //  labels: ['type: cleanup']
-            //})
+            await github.issues.addLabels({
+              owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


.github/workflows/tag_module_cleanup.yml
```diff
@@ -19,17 +19,20 @@ jobs:
...
+            // Note that pull requests are considered issues and "shared"
+            // actions for both features, like manipulating labels and
+            // milestones are provided within the issues API.
+            await github.issues.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              // 2021 q3 release milestone.
+              // https://github.com/google/blockly/milestone/18
+              milestone: 18
```
alschmiedt, line 31: I think this should still be `milestoneNumber`

alschmiedt (approved):

.github/workflows/tag_module_cleanup.yml
moniika: `milestoneNumber` is a local variable that doesn't exist in this version.

alschmiedt: Yeah, I was trying to say that I think we should keep the local variable `milestoneNumber` around. Probably not worth opening a new pr though so it is fine : )

PR #5325, opened 2021-08-07
Migrate core/events/variable_events.js to goog.module syntax

moniika:
## The basics

### Resolves

Part of #5026

### Proposed Changes

Converts `core/events/variable_events.js` to `goog.module` syntax. 

### Additional Information

The modules in `core/events/variable_events.js` were split out into the files:
- `core/events/events_var_base.js`
- `core/events/events_var_create.js`
- `core/events/events_var_delete.js`
- `core/events/events_var_rename.js`


core/events/events_var_create.js
```diff
@@ -0,0 +1,82 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Class for a variable creation event.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
+'use strict';
+
+goog.provide('Blockly.Events.VarCreate');
```
cpcallen: Not a problem with this PR, but: might it be worth filing an issue to reconsider the naming of these modules?

In particular, I think `Blockly.Events.Variable.Create` seems more readable—though I admit that a require of the form

```JS
const Create = goog.require('Blockly.Events.Variable.Create');
```

might be undesirable because `Create` on its own would then be pretty ambiguous, so one might end up with

```JS
const VarCreate = goog.require('Blockly.Events.Variable.Create');  // or VariableCreate
```

…which means the change wouldn't necessarily buy us much.

You probably have a better perspective than I for considering this question.

cpcallen (approved):
This would make an excellent model PR for how to present a migration that involves splitting a file.  👍

core/events/events_var_create.js
moniika: If we did add an extra namespace to variable events, I would recommend we update all of our events as they all fall into a category of one of:
- "Workspace"
- "Block"
- "WorkspaceComment"
- "Variable"
- "Ui"

I lean towards not changing it so there's less api changes introduced.

Additionally, the constants defined on `Blockly.Events` for the variable event types start with `VAR_` (as well as the string values starting with "var_") which makes this naming consistent.

PR #5329, opened 2021-08-09
feat: upgrade block defs to have JSO serialization hooks

BeksOmega:
## The basics

### Resolves
Work on project cereal

### Proposed Changes
Adds JSO serialization hooks to block definitions which need them. See the [deserialization](https://docs.google.com/document/d/1Dbpezxsd4H1S0p31ncrZ1hLaMfFYBeanv8pqjkpxoHI/edit?resourcekey=0-hYVVlo828cnLmnlj30Lysg#heading=h.gk17j63tj8rg) section of the design document for information about which blocks do not need hooks.

### Reason for Changes
Makes the JSO format smaller by not using stringified XML for mutators.

### Test Coverage
Added a few more mutator tests, but mostly just enabled existing tests.


blocks/procedures.js
```diff
@@ -150,6 +152,54 @@ Blockly.Blocks['procedures_defnoreturn'] = {
     // Show or hide the statement input.
     this.setStatements_(xmlElement.getAttribute('statements') !== 'false');
   },
+  /**
+   * Returns the state of this block as a JSON serializable object.
+   * @return {{params: (!Array<{name: string, id: string}>|undefined),
```
maribethb, line 157: This needs a `?` before the record type definition, since this function may return null.

blocks/procedures.js
```diff
@@ -1026,14 +1110,20 @@ Blockly.Blocks['procedures_ifreturn'] = {
...
+
   /**
    * Called whenever anything on the workspace changes.
    * Add warning if this flow block is not nested inside a loop.
    * @param {!Blockly.Events.Abstract} _e Change event.
    * @this {Blockly.Block}
    */
   onchange: function(_e) {
-    if (!this.workspace.isDragging || this.workspace.isDragging()) {
+    if (this.workspace.isDragging && this.workspace.isDragging()) {
```
maribethb, line 1126: Are you sure this is intended? The old effect would basically short circuit early if we had a Workspace instead of a WorkspaceSvg. In WorkspaceSvg `isDragging` would never be undefined, it looks like to me. So this is changing the behavior to have this function run on non-svg workspaces and I don't know enough to know if it should or not.

blocks/procedures.js
```diff
@@ -150,6 +152,54 @@ Blockly.Blocks['procedures_defnoreturn'] = {
...
+        var param = state['params'][i];
+        var variable = Blockly.Variables.getOrCreateVariablePackage(
+            this.workspace, param['id'], param['name'], '');
+        this.arguments_.push(variable.name);
+        this.argumentVarModels_.push(variable);
+      }
+    }
+    this.updateParams_();
+    Blockly.Procedures.mutateCallers(this);
+    this.setStatements_(state['hasStatements'] === undefined ? true : false);
```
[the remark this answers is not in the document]
BeksOmega: Yeah that seems much safer. Thank you for the catch!

blocks/procedures.js
```diff
@@ -807,6 +861,34 @@ Blockly.Blocks['procedures_callnoreturn'] = {
     }
     this.setProcedureParameters_(args, paramIds);
   },
+  /**
+   * Returns the state of this block as a JSON serializable object.
+   * @return {{itemCount: number}} The state of this block, ie the params and
+   *     procedure name.
```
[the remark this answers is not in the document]
BeksOmega: Whoops yep thank you!

blocks/procedures.js
BeksOmega: Yeah I changed it because I do want it to run on headless workspaces. Before this wasn't as much of a problem because we were serializing the state (ie whether the input exists or not). But because I didn't add jso serialization hooks, now this relies on the change listener to remove the input.

Plus this really should have been running in headless anyway, because if devs programmatically move their blocks we want the block to react how it would in a rendered workspace. Eg, this is important for automated testing.

blocks/lists.js
```diff
@@ -141,13 +142,31 @@ Blockly.Blocks['lists_create_with'] = {
...
+      'itemCount': this.itemCount_,
+    };
+  },
+  /**
+   * Applies the given state to this block.
+   * @param {*} state The state to apply to this block, ie the item count.
+   */
+  loadExtraState: function(state) {
+    this.itemCount_ = state['itemCount'];
+    this.updateShape_();
```
cpcallen, line 168: @BeksOmega : any particular reason you call `updateShape_` here instead of `rebuildShape_`, as `domToMutation` does?

(See [this comment reopening issue #2037](https://github.com/google/blockly/issues/2037#issuecomment-2209637735) for why I am curious.)

BeksOmega: Best guess is that I didn't understand what `rebuildShape_` was trying to work around, and I figured `updateShape_` was more efficient. I don't think there's any behavioral reason to prefer `updateShape_`

PR #5331, opened 2021-08-09
Migrate core/widgetdiv.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/widgetdiv.js` to `goog.module` with ES6 `const`/`let`.


core/widgetdiv.js
```diff
@@ -16,201 +16,232 @@
...
+let DIV;
+/** @deprecated September 2021 */
+exports.DIV = DIV;
+
+const getDiv = function() {
+  return DIV;
+};
+exports.getDiv = getDiv;
+
+const setDiv = function(newDiv) {
```
moniika, line 70: We should add jsdoc for these two methods.

core/widgetdiv.js
```diff
@@ -16,201 +16,232 @@
...
+  DIV = newDiv;
+};
```
moniika, line 72: This seems like a place where we could make a decision make a setter only for tests.
I don't think the original intention was to allow this property to be reassigned. It is set by the `createDom` method that is called in `inject`.
However, I'm not sure whether anyone is currently reassigning this property.

tests/mocha/field_textinput_test.js
```diff
@@ -168,7 +168,7 @@ suite('Text Input Fields', function() {
           };
           field.clickTarget_ = document.createElement('div');
           Blockly.mainWorkspace = workspace;
-          Blockly.WidgetDiv.DIV = document.createElement('div');
+          Blockly.WidgetDiv.setDiv(document.createElement('div'));
```
moniika, line 171: These tests didn't fail before because the mocha tests don't have any cleanup to reset the state of `Blockly.WidgetDiv.DIV` (even though the original set didn't work).

Could you please add that cleanup to `setTestTeardown` or file an issue. It's possible that adding the cleanup may reveal other test failures, so if that's the case then an issue would be better.

core/widgetdiv.js
gonfunko: Done

core/widgetdiv.js
gonfunko: Yeah, I agree that seems better - updated accordingly.

tests/mocha/field_textinput_test.js
gonfunko: Done, and also refactored the test to use the public createDom() for setup.

tests/mocha/field_textinput_test.js
```diff
@@ -183,6 +183,7 @@ suite('Text Input Fields', function() {
         if (this.stub) {
           this.stub.restore();
         }
+        Blockly.WidgetDiv.testOnly_setDiv(null);
```
moniika, line 186: We want this clearing of the global variable to happen for all tests, not just this suite.
I think it should go in `sharedTestTeardown`

gonfunko: Done

moniika (approved):

PR #5332, opened 2021-08-09
Migrate core/workspace.js to goog.module syntax

moniika:
## The basics

- [x] I have run `npm run test` locally already.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/workspace.js` to `goog.module` syntax.


core/workspace.js
```diff
@@ -13,62 +13,69 @@
 goog.module('Blockly.Workspace');
 goog.module.declareLegacyNamespace();
 
+/* eslint-disable-next-line no-unused-vars */
+const Abstract = goog.requireType('Blockly.Events.Abstract');
```
gonfunko, line 17: This may be clearer as EventsAbstract, but up to you.

core/workspace.js
```diff
@@ -770,25 +783,27 @@ Blockly.Workspace.prototype.setVariableMap = function(variableMap) {
  * Database of all workspaces.
  * @private
  */
-Blockly.Workspace.WorkspaceDB_ = Object.create(null);
+Workspace.WorkspaceDB_ = Object.create(null);
```
gonfunko, line 786: Since this is both private and static, I think you can just do `const workspaceDB = Object.create(null);` and remove the private annotation/replace uses of it in this file.

gonfunko (approved):

core/workspace.js
moniika: I like that change, but we're not doing that in other files (and I prefer to be consistent).
I'll create a bug for updating the require in all the files at once and leave this for now.

moniika: https://github.com/google/blockly/issues/5336

PR #5344, opened 2021-08-10
Fix issue that prevented Mocha and generator tests from running on certain configurations

gonfunko:
This PR adds the --disable-gpu flag to the run mocha/generator test scripts, which works around an issue where Chrome would hang when running the tests on Linux with older versions of the NVIDIA driver.

This fixes https://github.com/google/blockly/issues/5345.

### DIFF
--- a/tests/generators/run_generators_in_browser.js
+++ b/tests/generators/run_generators_in_browser.js
@@ -52,7 +52,8 @@ async function runGeneratorsInBrowser() {
     };
   } else {
     // --disable-gpu is needed to prevent Chrome from hanging on Linux with
-    // NVIDIA drivers older than v295.20.    
+    // NVIDIA drivers older than v295.20. See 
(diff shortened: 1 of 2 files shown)


tests/generators/run_generators_in_browser.js
```diff
@@ -51,8 +51,10 @@ async function runGeneratorsInBrowser() {
       args: ['--headless', '--no-sandbox', '--disable-dev-shm-usage', '--allow-file-access-from-files']
     };
   } else {
+    // --disable-gpu is needed to prevent Chrome from hanging on Linux with
+    // NVIDIA drivers older than v295.20.
```
maribethb, line 55: Could you add a link here and below to #5345 so we have the future context, please?

maribethb (approved):

tests/generators/run_generators_in_browser.js
gonfunko: Done!

PR #5367, opened 2021-08-12
Migrate core/renderers/thrasos/info.js to goog.module syntax

moniika:
## The basics

- [x] I have run `npm run test` locally already.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/renderers/thrasos/info.js` to `goog.module` syntax.


core/renderers/thrasos/info.js
```diff
@@ -11,8 +11,8 @@
  */
 'use strict';
 
-goog.provide('Blockly.thrasos');
```
cpcallen, line 14: Just out of curiosity: what's with this?  Any idea why this file was providing `Blockly.thrasos` when it didn't actually export any properties other than `Blockly.thrasos.RenderInfo`?

cpcallen (approved):

core/renderers/thrasos/info.js
moniika: I talked with Rachel and she said that it was fine to remove.
It's possible it was added because it was thought that there needed to be a provide for the intermediate namespace.
It's also not needed as an entry point for getting all the requires for the renderer because we use `Blockly.thrasos.Renderer` for that.

PR #5006, opened 2021-07-09
Create Github Action to comment on PR while develop is frozen

moniika:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

https://github.com/google/blockly/issues/4971

### Proposed Changes

Adds GitHub action to comment on new pull requests while develop is frozen.

#### Behavior After Change

While this workflow is enabled, new pull requests without the label 'ignore-freeze' is commented on to let the author know that they may experience a delay in getting a review.

### Test Coverage

Not tested.

### Additional Information

This workflow requires manual enabling/disabling to track when develop is frozen.

PR #5369, opened 2021-08-13
Fix blocks with no next connections overlapping statement input in Zelos

### LINKED ISSUE #5170: Stacks with no next connection overlap statement input in Zelos
**Describe the bug**

If a stack block does not have a next connection, and you add it to a statement input, the bottom of the block will overlap the bottom of the statement input.
Screenshot 2021-07-22 07 09 50

**To Reproduce**

1. Open up the playground, and change to the Zelos renderer.
2. Create blocks like seen above.
3. Observe how the stack block overlaps the bottom of the statement input.

**Expected behavior**

Block should snuggly fit within the statement input, without overlapping.

**Additional context**

It looks like zelos is actually trying to handle this case [here](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/renderers/zelos/measurables/inputs.js#L43)? But for some reason it's not working.

hpnrep6:
## The basics
- [x] I branched from develop

### Resolves
Fixes #5170

### Proposed Changes

Use the zelos.StatementInput function to handle statements in addInput_ instead of the blockRendering.StatementInput function used by the superclass' addInput_.

#### Behavior Before Change

Screenshot from 2021-08-12 23-17-46

#### Behavior After Change

Screenshot from 2021-08-12 22-29-40

### Reason for Changes

Blockly.zelos.StatementInput was unused and is required to correctly render using Zelos.

### Test Coverage

Chromium and Firefox on Ubuntu 21.04

### DIFF
--- a/core/renderers/zelos/info.js
+++ b/core/renderers/zelos/info.js
@@ -276,6 +276,10 @@ Blockly.zelos.RenderInfo.prototype.addInput_ = function(input, activeRow) {
     activeRow.elements.push(
         new Blockly.zelos.StatementInput(this.constants_, input));
     activeRow.hasStatement = true;
+
+    if (activeRow.align == null) {
+      activeRow.align = input.align;
+    }
     return;
   }
   Blockly.zelos.RenderInfo.superClass_.addInput_.call(this, input, activeRow);
--- a/core/renderers/zelos/info.js
+++ b/core/renderers/zelos/info.js
@@ -28,6 +28,7 @@ goog.require('Blockly.utils.object');
 goog.require('Blockly.zelos.BottomRow');
 goog.require('Blockly.zelos.RightConnectionShape');
 goog.require('Blockly.zelos.TopRow');
+goog.require('Blockly.zelos.StatementInput');
 
 goog.requireType('Blockly.BlockSvg');
 goog.requireType('Blockly.zelos.ConstantProvider');
@@ -270,6 +271,12 @@ Blockly.zelos.RenderInfo.prototype.addInput_ = function(input, activeRow) {
       activeRow.align == Blockly.constants.ALIGN.LEFT &&
       input.align == Blockly.constants.ALIGN.RIGHT) {
     activeRow.rightAlignedDummyInput = input;
+  } else if (input.type == Blockly.inputTypes.STATEMENT) {
+    // Handle statements without next connections correctly.
+    activeRow.elements.push(
+        new Blockly.zelos.StatementInput(this.constants_, input));
+    activeRow.hasStatement = true;
+    return;
   }
   Blockly.zelos.RenderInfo.superClass_.addInput_.call(this, input, activeRow);
 };


hpnrep6:
Would rewriting all the `if` `else` statements be required (or adding the following lines to the `else if (input.type == Blockly.inputTypes.STATEMENT)` block) since the lines:
https://github.com/google/blockly/blob/79522a07713839d8943c900f35d5c36f867b7456/core/renderers/common/info.js#L381-L383
will not be executed with this change.

BeksOmega:
Thank you so much for digging into this @hpnrep6 This is a really tricky bug! I would have never suspected that the correct class wasn't actually being used lol.

> Would rewriting all the if else statements from the superclass' addInput_ function be required...

Good catch on this! I think that the way [super class function](https://github.com/google/blockly/blob/79522a07713839d8943c900f35d5c36f867b7456/core/renderers/common/info.js#L358) is written is not ideal. The direct references to the input classes mean we have to duplicate logic in the subclasses. Eg how the [geras renderer's code](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/renderers/geras/info.js#L86) has basically the same structure, but it had to be rewritten for the different geras classes. Ideally it would probably use something like dependency injection, but refactoring the code to do that is quite the undertaking haha.

tl:dr I think in this case the best thing to do is just duplicate the alignment check. Eg:
```javascript
  } else if (input.type == Blockly.inputTypes.STATEMENT) {
    // Handle statements without next connections correctly.
    activeRow.elements.push(
        new Blockly.zelos.StatementInput(this.constants_, input));
    activeRow.hasStatement = true;

    // These lines added!
    if (activeRow.align == null) {
      activeRow.align = input.align;
    }

    return;
  }
```

Once that's in I'll pull this code down and smoke test it a bit. So just so you know, it may be Monday before this gets merged =)

I also noticed you've got a bit of a streak of contributions going :D Thank you so much for helping us out! We always really appreciate people fixing issues like this which are hard to get around to. If you ever want to work on stuff and are having trouble finding something feel free to ping me! I know digging through the issues yourself can be a bit of a hassle haha.

But yeah, all that's just to say thank you so much :D and feel free to message us whenever!

hpnrep6:
Thanks! I've added the duplicate check to the changes.

BeksOmega:
Ope, I just noticed, could you also rebase this so that this is against `develop` @hpnrep6 ?

BeksOmega:
Other than that I just smoke tested a whole bunch of stuff and everything looks good!

For future people, I checked that:
   * All of the built-in and test blocks continue to render as they do on master.
   * Alignment of inputs is still respected
   * Alignment of statement inputs is still respected
   * Connected child blocks render as expected (with and without next connections)
   * Connected shadow blocks render as expected (with and without next connections)

So once this is rebased it should be good to merge =)

BeksOmega:
Thanks for the rebase @hpnrep6 but it looks like we've got some merge conflicts :/

Here's what you'll want to do to fix those:
1) `git checkout develop`
2) `git pull upstream develop` replace `upstream` with whatever you name for his repo is.
3) `git checkout zelos-statement`
4) `git rebase -i develop`
    This starts an "interactive rebase", which allows you to change your commit history. You're going to use this to get rid of those bad commits from master, but be careful because this is one of the only ways you can lose work with git.
    You'll see something that looks like this:
    ```
    pick 5e4c650d Bump ws from 7.4.4 to 7.5.1
    pick 71e7f125 Bump lodash from 4.17.19 to 4.17.21
    pick 615abab1 Get deploy files from the correct directory
    pick 8c635b5f Create Github Action to comment on PR while develop is frozen (#5006)
    pick f2591438 Revert "Create Github Action to comment on PR while develop is frozen (#5006)" (#5013)
    pick bd46d8a0 Bump hosted-git-info from 2.8.4 to 2.8.9 (#4980)
    pick 16ca3786 Revert "Get deploy files from the correct directory"
    pick 36fa8604 Pull Request template for goog.module PRs
    pick d4196c11 Update goog_module.md with suggested fixes
    pick 79522a07 Use zelos' StatementInput function in addInput_
    pick 4ab506da Add align check to addInput_
    ```
5) Change that to look like this:
    ```
    d 5e4c650d Bump ws from 7.4.4 to 7.5.1
    d 71e7f125 Bump lodash from 4.17.19 to 4.17.21
    d 615abab1 Get deploy files from the correct directory
    d 8c635b5f Create Github Action to comment on PR while develop is frozen (#5006)
    d f2591438 Revert "Create Github Action to comment on PR while develop is frozen (#5006)" (#5013)
    d bd46d8a0 Bump hosted-git-info from 2.8.4 to 2.8.9 (#4980)
    d 16ca3786 Revert "Get deploy files from the correct directory"
    d 36fa8604 Pull Request template for goog.module PRs
    d d4196c11 Update goog_module.md with suggested fixes
    pick 79522a07 Use zelos' StatementInput function in addInput_
    pick 4ab506da Add align check to addInput_
    ```
    I think this uses your console's default text-editor which could be vim, emacs, nano, etc. If it's vim you can move between lines using the arrow keys. Then type `ce` to remove the 'pick' word. Then type 'd'. And then hit `esc`. Then repeat that for all the lines you need to edit.
6) Exit your text editor. If you are using vim you do that by hitting `esc` and then typing `:wq`.
7) `git push --force origin zelos-statement` replace `origin` with whatever your name for your fork of Blockly is.
8) Celebrate because that should be it!

Best of luck! If that doesn't work I'll see if I can remove those commits from my end =)

hpnrep6:
Thanks again! Saved me there since I was kind of struggling to get it working (since I've only really used the basic git commands).

BeksOmega:
> Thanks again! Saved me there since I was kind of struggling to get it working (since I've only really used the basic git commands before).

Haha yeah I totally get you. The docs for it really suck as well. I usually rely on stack overflow, but that's only useful if you already know what to google haha.

Thank you for doing that extra work for me! And thank you again for putting up this fix. It feels so good to have it working :D

BeksOmega (approved):

PR #5378, opened 2021-08-16
fix: insertion markers and change events to work with JSO hooks

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Proposed Changes
Updates the logic for insertion markers and change events to properly handle JSO serialization hooks.

### Reason for Changes
Bugs are bad.

### Test Coverage
 * Added tests for mutators properly firing (or not firing) events. 
 * Added tests for undoing and redoing mutator change events, for both XML and JSO hooks
 * Manually tested that insertion markers are properly constructed for blocks with XML hooks, JSO hooks, or both.

### DIFF
--- a/core/events/block_events.js
+++ b/core/events/block_events.js
@@ -193,7 +193,7 @@ Blockly.Events.BlockChange.prototype.run = function(forward) {
       block.setInputsInline(!!value);
       break;
     case 'mutation':
-      var oldState = this.getExtraBlockState_(
+      var oldState = Blockly.Events.BlockChange.getExtraBlockState_(
           /** @type {!Blockly.BlockSvg} */ (block));
       if (block.loadExtraState) {
         block.loadExtraState(JSON.parse(/** @type {string} */ (value) || '{}'));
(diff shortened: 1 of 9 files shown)


core/events/block_events.js
```diff
@@ -193,23 +193,40 @@ Blockly.Events.BlockChange.prototype.run = function(forward) {
...
     default:
       console.warn('Unknown change type: ' + this.element);
   }
 };
 
+/**
+ * Returns the extra state of the given block (either as XML or a JSO, depending
+ * on the block's definition).
+ * @param {!Blockly.BlockSvg} block The block to get the extra state of.
+ * @return {string} A strigified version of the extra state of the given block.
```
NeilFraser, line 217: strigified?  :)

core/mutator.js
```diff
@@ -456,6 +454,23 @@ Blockly.Mutator.prototype.workspaceChanged_ = function(e) {
   }
 };
 
+/**
+ * Returns the extra state of the given block (either as XML or a JSO, depending
+ * on the block's definition).
+ * @param {!Blockly.BlockSvg} block The block to get the extra state of.
+ * @return {string} A strigified version of the extra state of the given block.
+ */
+Blockly.Mutator.prototype.getExtraBlockState_ = function(block) {
```
NeilFraser, line 463: This function and Blockly.Events.BlockChange.prototype.getExtraBlockState_ appear to be identical (right down to the strigification).  Copied code is generally a sign of sub-optimal code (bloat and harder to maintain, though easier to write).  Is it practical to extract these out somewhere, or have one call the other?

Sometimes the answer is no -- but do take a look at this.

NeilFraser (approved):

core/mutator.js
BeksOmega: Yeah this is defnitely something I copy-pasta'ed.

I think the root cause of this duplication is that the `BlockChange` type should use subtyping instead of a switch statement. Eg we would have a `BlockFieldChange` event, `BlockMutatorChange` event, etc. This would allow us to encapsulate the logic for getting the extra state in the `BlockMutatorChange` event. For example, we could have the event do a "recording" like move events do.

But right now, because mutator events are not their own type, every time a caller wants to construct a `BlockChange` event for a mutation, they have to use this logic to get the extra state so they can pass it to the constructor :/

I could add this logic as a static function on the `Blockly.Events.BlockChange` type (i.e. `Blockly.Events.BlockChange.getExtraBlockState`) but based on [this comment from Christopher](https://github.com/google/blockly/issues/5026#issuecomment-880806386) about the goog.module conversion, I'm not sure if that's a good idea. What's your opinion @NeilFraser ? Leave the duplication, or move to a static method? (both being temporary solutions)

NeilFraser: Entirely your call.  Your instincts here are better than mine.  Just highlighting the issue.

PR #5380, opened 2021-08-16
Migrate core/events/ws_comment_events.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/events/ws_comment_events.js` to `goog.module` with ES6 `const`/`let`.


tests/mocha/event_test.js
```diff
@@ -538,8 +543,12 @@ suite('Events', function() {
...
           testSuite.testCases.forEach((testCase) => {
             test(testCase.title, function() {
-              var event = new testCase.class(...testCase.getArgs(this));
-              var event2 = new testCase.class();
+              // Some classes that are optional requires need to be dynamically
+              // loaded via goog.module.get.
+              const testClass = testCase.classModule ?
+                  goog.module.get(testCase.classModule) : testCase.class;
+              var event = new testClass(...testCase.getArgs(this));
+              var event2 = new testClass();
```
alschmiedt, line 551: So this has caused a bit of an adventure haha

I originally thought this should not be needed so I commented it out and all my tests passed. I then looked into it more and I am now confused why commenting out those lines did not break everything. Is this fixing an error that you were getting? If yes, what was the error? I *think* we might be able to do this by including the appropriate files in the `mocha/index.html` file instead of having to do this bit, but I'm not entirely sure.

core/events/events_comment_delete.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Classes for all comment events.
```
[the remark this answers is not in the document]
gonfunko: Done

tests/mocha/event_test.js
gonfunko: Sorry for sending you down that path! When I wrote this the test was failing without these changes, but it now appears that the test passes as it was, presumably since fixing up the dependencies in other files in the meantime has sorted things out here. Reverted the test file to its original state.

alschmiedt (approved):

Issue #5160, opened 2021-07-21
Use new eslint config for es6 / goog.module


maribethb:
## Describe the solution you'd like

After the round of goog.module conversion is finished, and we've converted all of the `var`s to `let/const`s, and clang-formatted all the things... we need to update the eslint config. We can't do it beforehand, because all of the unconverted code will have a lot of errors. Afterwards, there will probably still be some errors that we need to fix to actually be in agreement with the Google style guide.

Ideally, we'd be able to use the same library that we use in blockly-samples that understands the Google style guide: https://github.com/google/eslint-config-google. I think this is feasible, and while we will have some real problems to clean up afterwards, we'd be
a) conforming to Google style
b) using the same style in both blockly and blockly-samples.

## Describe alternatives you've considered

We might be able to use the same eslint config as blockly-samples, or extend that config (as opposed to both extending from the same config, but not each other). I'm not sure how feasible this will be yet, so when we're making this change, we should consider how similar they are and if this makes sense.

rachel-fenichel:
There are a lot of issues in `tests/mocha/*`, including use of `this` and `var`.

Setting those aside and looking primarily at core and blocks, there are 492 violations of `comma-dangle`, which I'm fixing in https://github.com/google/blockly/pull/5625.

There are 512 violations of `camelcase`. We use `opt_argument` to indicate optional args, so we need to allow a few prefixes. This configuration of the rule yields 16 problems, all of which are legitimate:
```
"camelcase": ["error", {
    "properties": "never",
    "allow": ["^opt_", "^_opt_", "^testOnly_"]
}]
```

rachel-fenichel:
Rules that we are currently violating outside of `tests/mocha/*`:

- [ ] `no-invalid-this`
- [ ] `comma-spacing`
- [ ] `key-spacing
- [ ] `camelcase`
- [ ] `new-cap`
- [ ] `padded-blocks`
- [ ] `prefer-const`
- [ ] `brace-style`
- [ ] `guard-for-in`
- [ ] `one-var`
- [ ] `quote-props`
- [ ] `prefer-spread`
- [ ] `require-jsdoc`
- [ ] `comma-dangle`

PR #5398, opened 2021-08-19
fix: project cereal cleanup

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Description

This is a running PR for small changes I need to make as my project finishes up and things get merged in.

Each commit will be stand alone, and can be reviewed individually.


alschmiedt (approved):

core/serialization/blocks.js
```diff
@@ -52,9 +52,6 @@ exports.ConnectionState = ConnectionState;
  *     y: (number|undefined),
```
alschmiedt, line 52: Not from this PR, but can we remove the TODO and the `/* eslint-disable no-use-before-define */` above?

BeksOmega: We can, but lint still isn't fixed, so we'd have to reorganize the file :/ Right now higher level functions are defined before lower level procedures, and we'd have to flip it.

alschmiedt: Ahhh, makes sense. Do you plan on fixing before merging in to master? If not, then we should just make an issue to update it so we have something to reference in the TODO.

BeksOmega: I can link it to https://github.com/google/blockly/issues/5160

alschmiedt: Yup, sounds great : )

PR #5419, opened 2021-09-01
Allow command key zoom on mac devices

### LINKED ISSUE #4993: Zoom in/out with CMD on Mac
Using JS code:

Blockly.inject(element, {
      zoom: {
        wheel: true,
      },
      move: {
        scrollbars: true,
        drag: true,
        wheel: true,
      },
    });

we can zoom in/out on CTRL, but it doesn't work on CMD on Macs.

hpnrep6:
## The basics

- [x] I branched from develop

### Resolves

Fixes #4993.

### Proposed Changes

Also check if the command key has been pressed on mac devices when deciding if the user is able to zoom. 

#### Behavior Before Change

The workspace can't be zoomed when the command key is pressed and the mouse wheel is scrolled (when move with wheel is enabled).

#### Behavior After Change

The command key can be used to zoom.

### Test Coverage

Tested on:
- OS: macOS 11.4, Ubuntu 21.04 (to check that zoom doesn't work with the super key)

Browsers:
- Google Chrome on macOS
- Safari on macOS
- Chromium on Ubuntu

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1801,7 +1801,8 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
 
   var scrollDelta = Blockly.utils.getScrollDeltaPixels(e);
 
-  // Zoom should also be enabled by the command key on Mac devices.
+  // Zoom should also be enabled by the command key on Mac devices,
+  // but not super on Unix.
   var commandKey;
   if (Blockly.utils.userAgent.MAC) {
     commandKey = e.metaKey;
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1800,7 +1800,14 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
   }
 
   var scrollDelta = Blockly.utils.getScrollDeltaPixels(e);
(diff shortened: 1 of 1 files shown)


core/workspace_svg.js
```diff
@@ -1800,7 +1800,14 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
   }
 
   var scrollDelta = Blockly.utils.getScrollDeltaPixels(e);
-  if (canWheelZoom && (e.ctrlKey || !canWheelMove)) {
+
+  // Zoom should also be enabled by the command key on Mac devices.
```
BeksOmega, line 1804:
```suggestion
  // Zoom should also be enabled by the command key on Mac devices,
  // but not super on Unix.
```
Could you modify this comment to clarify why we need the `userAgent` check? =)

BeksOmega (commented):
Hi @hpnrep6 this PR looks great! Just one note below about clarifying the comment.

Also, thanks for doing all of the different browser/OS testing :D I know it takes a bunch of time to load stuff up on different devices, so I really appreciate it!

hpnrep6:
Thanks! I've applied the changes for the comment to the PR.

BeksOmega (approved):

PR #5421, opened 2021-09-02
fix: cereal backwards compatibility

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Proposed Changes
Changes the JSO system to automatically convert values returned from XML hooks to strings, so that the data can be encoded in the JSO state.

This allows us to serialize and deserialize blocks/fields that were built for the XML system using the JSO system. This will in turn allow us to move copy paste, the flyout, shadows, etc to the JSO system without breaking people.

### Reason for Changes
No breaking people!

### Test Coverage
Added tests for all of the following cases, for both serialization and deserialization, for both systems:
  * A field with no overrides.
  * A field with XML hooks.
  * A field with XML hooks that calls super.
  * A field with JSO hooks.
  * A field with JSO hooks that calls super.
  * A field with both sets of hooks.

### Additional Info

To clarify for Christopher: Yesterday when we discussed this plan, we were mostly talking about the Dropdown field. I forgot that the dropdown field has custom *deserialization* not *serialization* :/

But the same issue still applies.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -436,9 +436,10 @@ Blockly.Field.prototype.saveState = function() {
  * @package
  */
 Blockly.Field.prototype.loadState = function(state) {
-  if (!this.loadLegacyState(Blockly.Field, state)) {
-    this.setValue(state);
+  if (this.loadLegacyState(Blockly.Field, state)) {
+    return;
   }
+  this.setValue(state);
 };
 
 // eslint-disable-next-line valid-jsdoc
--- a/core/field.js
+++ b/core/field.js
@@ -444,7 +444,7 @@ Blockly.Field.prototype.saveState = function() {
 Blockly.Field.prototype.loadState = function(state) {
   if (Blockly.Field.prototype.loadState === this.loadState &&
       Blockly.Field.prototype.fromXml !== this.fromXml) {
-    this.fromXml(Blockly.Xml.textToDom(state));
+    this.fromXml(Blockly.Xml.textToDom(/** @type {string} */ (state)));
     return;
   }
   // Either they called this on purpose from their loadState, or they have
--- a/core/field.js
+++ b/core/field.js
@@ -422,16 +422,10 @@ Blockly.Field.prototype.toXml = function(fieldElement) {
  * @package
  */
 Blockly.Field.prototype.saveState = function() {
-  if (Blockly.Field.prototype.saveState === this.saveState &&
-      Blockly.Field.prototype.toXml !== this.toXml) {
-    var elem = Blockly.utils.xml.createElement("field");
-    elem.setAttribute("name", this.name || '');
-    var text = Blockly.Xml.domToText(this.toXml(elem));
-    return text.replace(
-        'xmlns="https://developers.google.com/blockly/xml" ', '');
+  var legacyState = this.saveLegacyState(Blockly.Field);
+  if (legacyState !== null) {
+    return legacyState;
   }
-  // Either they called this on purpose from their saveState, or they have
-  // no implementations of either hook. Just do our thing.
   return this.getValue();
 };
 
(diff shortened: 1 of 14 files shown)


core/serialization/blocks.js
```diff
@@ -162,6 +163,11 @@ const saveExtraState = function(block, state) {
     if (extraState !== null) {
       state['extraState'] = extraState;
     }
+  } else if (block.mutationToDom) {
+    const extraState = block.mutationToDom();
+    if (extraState !== null) {
+      state ['extraState'] = Xml.domToText(extraState);
```
cpcallen, line 169: No space before `[`.

core/field.js
```diff
@@ -422,6 +422,16 @@ Blockly.Field.prototype.toXml = function(fieldElement) {
...
+  if (Blockly.Field.prototype.saveState === this.saveState &&
+      Blockly.Field.prototype.toXml !== this.toXml) {
+    var elem = Blockly.utils.xml.createElement("field");
+    elem.setAttribute("name", this.name || '');
+    var text = Blockly.Xml.domToText(this.toXml(elem));
+    return text.replace(
+        'xmlns="https://developers.google.com/blockly/xml" ', '');
+  }
+  // Either they called this on purpose from their saveState, or they have
+  // no implementations of either hook. Just do our thing.
```
cpcallen, line 434: Implementation looks fine (here and, AFAICT, in the other files too), but it'd be nice if the common code could be factored out, leaving just a call and check like:
```JS
  var xml = this.saveLegacyState(Blockly.Field);
  if (xml) return xml;  // Subclass overrides toXml but not saveState.
```
in each class, using a using a (`@protected`)
```
Blockly.Field.prototype.saveLegacyState = function(superclass) {
  if (superclass.prototype.saveState !== this.saveState ||
      superclass.prototype.toXml === this.toXml) {
    return null;
  }
  var elem = Blockly.utils.xml.createElement("field");
  elem.setAttribute("name", this.name || '');
  var text = Blockly.Xml.domToText(this.toXml(elem));
  return text.replace(
        'xmlns="https://developers.google.com/blockly/xml" ', '');
}
```
(Maybe verify that `saveLegacyState` can't return the empty string!)

core/serialization/blocks.js
```diff
@@ -162,6 +163,11 @@ const saveExtraState = function(block, state) {
     if (extraState !== null) {
       state['extraState'] = extraState;
     }
+  } else if (block.mutationToDom) {
+    const extraState = block.mutationToDom();
```
cpcallen, line 167: Can `extraState` here end up duplicating some of the non-extra state on `state`?

core/field.js
BeksOmega: That sounds perfect. I wanted to factor this out, but I wasn't sure how to do it without creating a static function. Thank you :D

core/serialization/blocks.js
BeksOmega: It shouldn't unless developers explicitly want it to. `mutationToDom` doesn't have a default implementation. All we're getting from it is whatever developers define.

I'm worried I'm missing something here. Could you expound?

cpcallen: I was just wondering if any of the blocks we provide implement both `saveExtraState` and `mutationToDom`, and so if a developer overrode (only) the with an implementation that did a super call whether this might result in the same information being recorded both in fields in the JSO (by our block's `saveExtraState`) and as XML in the JSO (via override's super call).

But it looks like if a block implements `saveExtraState` then its `mutationToDom` is completely ignored.  Is that correct?

core/field_variable.js
```diff
@@ -225,20 +219,15 @@ Blockly.FieldVariable.prototype.saveState = function() {
...
-    return;
+  if (!this.loadLegacyState(Blockly.Field, state)) {
+    // This is necessary so that blocks in the flyout can have custom var names.
+    var variable = Blockly.Variables.getOrCreateVariablePackage(
+        this.sourceBlock_.workspace,
+        state['id'] || null,
+        state['name'],
+        state['type'] || '');
+    this.setValue(variable.getId());
   }
```
cpcallen, line 230: Something I learned from the Go styleists which I think is a generally applicable principle is that it's usually easier to read a function or loop that [gets rid of the special cases at the top and leaves the main body unindented](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow):
```JS
function foo() {
  if (specialCondition)  return;
  doUsualThing();
}
```
This convention is not so universally observed in JavaScript, but I think it can often be helpful nonetheless.

cpcallen (approved):
I'm not 100% certain I fully understand how this PR interacts with all the rest of Blockly (or even Project Cereal); you may wish to have someone more familiar have a look.  But the code to handle backward-compatibility checks looks like I'd expect it to, and there are no serious style issues.

core/serialization/blocks.js
BeksOmega: > But it looks like if a block implements saveExtraState then its mutationToDom is completely ignored. Is that correct?

That is correct! We ignore `mutationToDom` because we assume that if a dev has implemented the equivalent JSO hook (`saveExtraState`) they want us to use that.

> ...did a super call ...

Also, all block types are (basically) implemented as mixins, so we don't have to worry about subclassing afaik, because it's not supported. Eg the [lists_create_with](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/blocks/lists.js#L118) block.

cpcallen: > Also, all block types are (basically) implemented as mixins, so we don't have to worry about subclassing afaik

TIL (though possibly not for the first time…)

PR #5422, opened 2021-09-02
fix: dragging blocks from the flyout that only have XML hooks

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Description

1) Fixes a blocks not being positioned correctly in the flyout
2) Dragging blocks/fields out of the flyout that only have XML hooks

### Testing
Tested dragging blocks/fields that only have XML hooks out of the flyout using https://github.com/google/blockly-samples/pull/875

### DIFF
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -399,8 +399,8 @@ const loadPrivate = function(
  * @param {!State} state The state object to reference.
  */
 const loadCoords = function(block, state) {
-  const x = state['x'] === undefined ? 10 : state['x'];
-  const y = state['y'] === undefined ? 10 : state['y'];
+  const x = state['x'] === undefined ? 0 : state['x'];
+  const y = state['y'] === undefined ? 0 : state['y'];
   block.moveBy(x, y);
 };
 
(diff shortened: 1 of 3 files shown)


core/serialization/blocks.js
```diff
@@ -399,8 +399,8 @@ const loadPrivate = function(
  * @param {!State} state The state object to reference.
  */
 const loadCoords = function(block, state) {
-  const x = state['x'] === undefined ? 10 : state['x'];
-  const y = state['y'] === undefined ? 10 : state['y'];
+  const x = state['x'] === undefined ? 0 : state['x'];
+  const y = state['y'] === undefined ? 0 : state['y'];
```
alschmiedt, line 403: I think paste has a default of 10, just curious why this is different?

alschmiedt (approved):

core/serialization/blocks.js
BeksOmega: I thought the same thing! But this messes with the alignment of the flyout :/

And thinking about it now, I think it makes more sense that if coordinates aren't provided the block should just go to `(0, 0)`. Less likely to cause bugs that way I think.

PR #5423, opened 2021-09-03
fix: copy paste with json system

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Description

Fixes copying and pasting blocks with different sets of serialization hooks.

### Testing
* Tested copying and pasting blocks with different sets of hooks using the blocks from https://github.com/google/blockly-samples/pull/875
* Tested that workspace comments can be duplicated.
* Tested that the instance limit for blocks is still respected.

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -92,24 +92,10 @@ Blockly.draggingConnections = [];
 
 /**
  * Contents of the local clipboard.
- * @type {Element}
+ * @type {?Blockly.ICopyable.CopyData}
  * @private
  */
-Blockly.clipboardXml_ = null;
-
-/**
- * Source of the local clipboard.
- * @type {Blockly.WorkspaceSvg}
- * @private
- */
-Blockly.clipboardSource_ = null;
-
-/**
- * Map of types to type counts for the clipboard object and descendants.
- * @type {Object}
- * @private
- */
-Blockly.clipboardTypeCounts_ = null;
+Blockly.clipboardData_ = null;
 
 /**
  * Cached value for whether 3D is supported.
@@ -234,12 +220,7 @@ Blockly.deleteBlock = function(selected) {
  * @package
  */
 Blockly.copy = function(toCopy) {
-  var data = toCopy.toCopyData();
-  if (data) {
-    Blockly.clipboardXml_ = data.xml;
-    Blockly.clipboardSource_ = data.source;
-    Blockly.clipboardTypeCounts_ = data.typeCounts;
-  }
+  Blockly.clipboardData_ = toCopy.toCopyData();
 };
 
 /**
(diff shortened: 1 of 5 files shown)


core/workspace_svg.js
```diff
@@ -1467,40 +1467,56 @@ Blockly.WorkspaceSvg.prototype.highlightBlock = function(id, opt_state) {
...
     return;
   }
-  // The check above for tagName rules out the possibility of this being a DocumentFragment.
-  xmlBlock = /** @type {!Element} */ (xmlBlock);
   if (this.currentGesture_) {
     this.currentGesture_.cancel();  // Dragging while pasting?  No.
   }
-  if (xmlBlock.tagName.toLowerCase() == 'comment') {
-    this.pasteWorkspaceComment_(xmlBlock);
+  if (state['type']) {
```
alschmiedt, line 1483: I think this is just to check if we are in the jso system? If that is the case, could you add a comment explaining this above the if statement?

core/workspace_svg.js
```diff
@@ -1539,7 +1555,7 @@ Blockly.WorkspaceSvg.prototype.pasteBlock_ = function(xmlBlock) {
           blockY += Blockly.SNAP_RADIUS * 2;
         }
       } while (collide);
-      block.moveBy(blockX, blockY);
+      block.moveTo({x: blockX, y: blockY});
```
alschmiedt, line 1558: I think this is technically expecting a `Blockly.utils.Coordinate`. Would you mind updating to use that instead? Also, I'm a bit surprised closure isn't complaining.

alschmiedt (approved):

core/blockly.js
```diff
@@ -274,17 +255,10 @@ Blockly.paste = function() {
...
-  // Create a duplicate via a copy/paste operation.
+  var data = Blockly.clipboardData_;
   Blockly.copy(toDuplicate);
-  toDuplicate.workspace.paste(Blockly.clipboardXml_);
-
-  // Restore the clipboard.
-  Blockly.clipboardXml_ = clipboardXml;
-  Blockly.clipboardSource_ = clipboardSource;
+  toDuplicate.workspace.paste(Blockly.clipboardData_.saveInfo);
+  Blockly.clipboardData_ = data;
```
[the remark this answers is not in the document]
BeksOmega: The above `Blockly.copy` call reassigns `clipboardData_` so we have to bring it back.

I guess you could change it to just `toDuplicate.workspace.paste(toDuplicate.toCopydata())`

I don't think it's worth it to change this since since the [new implementation](https://github.com/google/blockly/pull/5237/files#diff-96b05a61c1487616cf85c353d633e196e2aa643a7d00a7c0b2ca1877652d79a0R72) works.

core/workspace_svg.js
BeksOmega: I think this works because [coordinate](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/utils/coordinate.js#L26) is a struct. But you're right, using the constructor is more future proof.

PR #5425, opened 2021-09-03
fix: create and delete events, and the trashcan

BeksOmega:
## The basics

### Resolves

Work on project cereal

### Description

* Fixes undoing and redoing copy and paste events for blocks with JSON hooks.
* Fixes adding blocks with JSON hooks to the trashcan.

### Testing
Using the blocks from https://github.com/google/blockly-samples/pull/875
* Tested undoing and redoing create events
* Tested undoing and redoing delete events
* Tested adding blocks to the trashcan

### DIFF
--- a/core/events/block_events.js
+++ b/core/events/block_events.js
@@ -253,9 +253,8 @@ Blockly.Events.Create = function(opt_block) {
   /**
    * JSON representation of the block that was just created.
    * @type {!Blockly.serialization.blocks.State}
-   * @private
    */
-  this.blockJson_ = /** @type {!Blockly.serialization.blocks.State} */
+  this.json = /** @type {!Blockly.serialization.blocks.State} */
       (Blockly.serialization.blocks.save(opt_block, {addCoordinates: true}));
 };
 Blockly.utils.object.inherits(Blockly.Events.Create, Blockly.Events.BlockBase);
@@ -283,7 +282,7 @@ Blockly.Events.Create.prototype.toJson = function() {
   var json = Blockly.Events.Create.superClass_.toJson.call(this);
   json['xml'] = Blockly.Xml.domToText(this.xml);
   json['ids'] = this.ids;
-  json['blockJson_'] = this.blockJson_;
+  json['json'] = this.json;
   if (!this.recordUndo) {
     json['recordUndo'] = this.recordUndo;
   }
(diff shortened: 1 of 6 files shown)


BeksOmega:
TODO: update events docs if this gets approved.

core/events/block_events.js
```diff
@@ -247,12 +247,16 @@ Blockly.Events.Create = function(opt_block) {
     this.recordUndo = false;
   }
 
-  if (opt_block.workspace.rendered) {
```
alschmiedt, line 250: Why are we removing this?

BeksOmega: `blockToDomWithXY` [accepts a Blockly.Block](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/xml.js#L87). This works b/c Blockly.Block has a default implementation of `getRelativeToSurfaceXY` which just returns (0, 0) which is perfectly fine in this case.

It's not a necessary change, so I can remove it from this PR, but I just did the pruning while I was in the area.

core/trashcan.js
```diff
@@ -426,8 +425,8 @@ Blockly.Trashcan.prototype.openFlyout = function() {
   if (this.contentsIsOpen()) {
     return;
   }
-  var xml = this.contents_.map(Blockly.Xml.textToDom);
-  this.flyout.show(xml);
+  var contents = this.contents_.map((str) => JSON.parse(str));
```
alschmiedt, line 428: Arrow functions are not allowed in ie11. https://caniuse.com/?search=arrow%20functions

core/events/block_events.js
```diff
@@ -247,12 +247,16 @@ Blockly.Events.Create = function(opt_block) {
...
+  this.xml = Blockly.Xml.blockToDomWithXY(opt_block);
   this.ids = Blockly.Events.getDescendantIds(opt_block);
+
+  /**
+   * JSON representation of the block that was just created.
+   * @type {!Blockly.serialization.blocks.State}
+   * @private
+   */
+  this.blockJson_ = /** @type {!Blockly.serialization.blocks.State} */
+      (Blockly.serialization.blocks.save(opt_block, {addCoordinates: true}));
```
alschmiedt, line 259: As discussed offline, I think I would prefer if this was public because it makes it more consistent with the delete event and with the xml system.

alschmiedt (approved):

PR #5430, opened 2021-09-06
maxInstance fix

### LINKED ISSUE #5374: Block with maxInstances 'x' already disabled in toolbox when block 'x-1' is dragged on workspace
**Describe the bug**

When the maxInstances option is set for a block (for example to 5), the corresponding block in the toolbox gets the 'blocklyDisabled' class after all the instances have been used in the workspace as expected. 
The strange thing, however, is that this block already temporarily gets the 'blocklyDisabled' status while dragging around the second-last block on the workspace (for example block number 4). When the block is released, the status is flipped again...

**To Reproduce**

- Set the maxInstances option on one of the blocks in the toolbox, for example to 5
- Then drag around the 4th block around on the workspace

**Expected behavior**

The block in the toolbox should not temporarily receive the 'blocklyDisabled' status while dragging around the second-last block around on the workspace.

 - OS: iOS
 - Browser: chrome
 - Version: 91.0.4472.114

Apoorvgarg-creator:
## The basics

- [x] I branched from develop

### Resolves
Issue Fix #5374 

### Proposed Changes
This PR fixes the issue #5374, Filtering out the insertion marker in the function [getsBlockByType](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/workspace.js#L257) will fix the issue.

#### Behavior Before Change
As @kvanthienen stated-
When the maxInstances option is set for a block (for example to 5), the corresponding block in the toolbox gets the 'blocklyDisabled' class after all the instances have been used in the workspace as expected.
The strange thing, however, is that this block already temporarily gets the 'blocklyDisabled' status while dragging around the second-last block on the workspace (for example block number 4). When the block is released, the status is flipped again.

#### Behavior After Change
This is a short term fix where blocks which are not real ` Insertion marker ` are filtered out. Hence, on dragging the block it is not counted twice.Thus solving the issue when Maxinstance is reached when the **block haven't placed yet**.

https://user-images.githubusercontent.com/57873504/132156519-8a06f96c-7e8f-49dd-a840-3e1e2fd6173e.mov

### Reason for Changes
Whenever we were picking a block set with a maxInstance, [getsBlockByType](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/workspace.js#L257) returns the value with insertion marker + number of blocks, which was causing the problem as an virtual object `(Insertion marker)` got counted as well as.
So filtering out insertion marker gives the correct count.

### Test Coverage

 Tested on: 
 * Desktop Chrome: Yes, Worked perfectly

* Desktop Safari: Yes, Worked perfectly

### Documentation
No changes in documentation 

@BeksOmega

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -266,7 +266,12 @@ Blockly.Workspace.prototype.getBlocksByType = function(type, ordered) {
     }
     blocks.sort(this.sortObjects_);
   }
-  return blocks;
+
+  var filtered = blocks.filter(function(block) {
+    return !block.isInsertionMarker();
+  });
+
+  return filtered;
 };
 
 /**
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -267,11 +267,9 @@ Blockly.Workspace.prototype.getBlocksByType = function(type, ordered) {
     blocks.sort(this.sortObjects_);
   }
 
-  var filtered = blocks.filter(function(block) {
+  return blocks.filter(function(block) {
     return !block.isInsertionMarker();
   });
-
-  return filtered;
 };
 
 /**


Apoorvgarg-creator:
@googlebot I signed it!

core/workspace.js
```diff
@@ -266,7 +266,12 @@ Blockly.Workspace.prototype.getBlocksByType = function(type, ordered) {
     }
     blocks.sort(this.sortObjects_);
   }
-  return blocks;
+
+  var filtered = blocks.filter(function(block) {
+    return !block.isInsertionMarker();
+  });
+
+  return filtered;
```
BeksOmega, line 274:
```suggestion
  return blocks.filter(function(block) {
    return !block.isInsertionMarker();
  });
```
This can be turned into one return statement =)

BeksOmega (approved):
Hi @Apoorvgarg-creator ! This fix looks great. Thank you for putting this PR up and filling out the description. Seeing what you did to test was really helpful for reviewing this PR =)

I left a comment with one little code simplification, but otherwise LGTM! Once that's fixed I'll get this merged.

core/workspace.js
Apoorvgarg-creator: Thank you for reviewing. I have committed the suggested changes. 
Thank you @BeksOmega

BeksOmega: My pleasure! Thank you for the PR :D

PR #5440, opened 2021-09-08
Migrate tests goog module

alschmiedt:
### Proposed Changes
- Adds a `mocha_deps.js` file that contains everything in core, generators, blocks, and mocha/tests.
- Updates `test_helpers.js` to be a goog.module. 
- Updating `test_helpers.js` causes other tests to fail, so I updated a few as an example for how to fix the other ones.

We were uncertain if stubbing something like `Blockly.utilx.genUid` would work when we require it through `const utils = goog.require('Blockly.utils'). I tried the below and it seemed to work, but I'm not sure it is a comparable situation. 

```
sinon.stub(utilsToolbox, 'convertFlyoutDefToJsonArray').returns('AHHHH');
console.log(Blockly.utils.toolbox.convertFlyoutDefToJsonArray({})); <-- Prints out AHHHH
console.log(utilsToolbox.convertFlyoutDefToJsonArray({}));<-- Prints out AHHHH
```

I also tried
```
In toolbox_test.js
sinon.stub(utilsToolbox, 'convertFlyoutDefToJsonArray').returns('AHHHH');
this.toolbox.aMethod();

In Toolbox.js
Toolbox.prototype.aMethod = function() {
  console.log(Blockly.utils.toolbox.convertFlyoutDefToJsonArray({})); <-- Prints out AHHH
};
```

### Additional Information
I factored out `testAWorkspace` function into its own `workspace_helpers.js` file. It was being used in both `workspace_test.js` and `workspace_svg_test.js` so it made sense to me to create a workspace helper file same as we have procedure and toolbox helpers.


cpcallen:
OK, this looks really promising as far as mocking goes.

As far as the issues you noted in chat:
> Tests are still failing because I exported everything from `test_helpers.js` and they then need to be imported to the other tests where the helpers are used...I didn't know a way around this, but if there is one lmk and I can try to fix.

The only way around is through: all the test files that depend on the helpers need to become `goog.module`s too.  Actually, that's not strictly true; we have a few choices:

1. We could make all the test files into `goog.module`s.
0. We could move the helpers that mock/stub things into a separate file, and only `module`ify that file and the particular test files that depend on those particular functions.
0. We could have `test_helpers.js` do `goog.module.declareLegacyNamespace()`, and then refer to the helpers by their fully-qualified names (suggestion: `Blockly.test.helpers.*`)

But I think that #2 and #3 are probably only procrastinating from doing #1.

scripts/gulpfiles/build_tasks.js
```diff
@@ -342,8 +342,18 @@ function buildDeps(done) {
...
     'blocks',
   ];
+
+  const testRoots = [
+    closurePath,
+    'core',
+    'blocks',
+    'generators',
+    'tests/mocha'
+  ];
```
cpcallen, line 352:
```suggestion
  const testRoots = [
    ...roots,
    'generators',
    'tests/mocha'
  ];
```

scripts/gulpfiles/build_tasks.js
```diff
@@ -342,8 +342,18 @@ function buildDeps(done) {
...
   const args = roots.map(root => `--root '${root}' `).join('');
+  const testArgs = testRoots.map(root => `--root '${root}' `).join('');
   execSync(`closure-make-deps ${args} > tests/deps.js`, {stdio: 'inherit'});
+  execSync(`closure-make-deps ${testArgs} > tests/mocha/mocha_deps.js`, {stdio: 'inherit'});
```
cpcallen, line 356: What do you think about having this be `tests/deps.mocha.js`?  It seems tidier to me to keep all the generated code together, and not intermixed with the hand-written code.

tests/mocha/astnode_test.js
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+goog.module('ASTTest');
+
+const {sharedTestSetup, sharedTestTeardown} = goog.require('TestHelpers');
+
```
cpcallen, line 10: I believe that @moniika has been promulgating a standard of putting two blank lines between the imports (or module declaration, if there are no imports) and the main body of the file.
```suggestion


```

tests/mocha/astnode_test.js
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+goog.module('ASTTest');
```
cpcallen, line 7: Is there a good reason not to name test modules consistently with how core modules are named?  So this could be `Blockly.test.astTest`, for example.

tests/mocha/test_helpers.js
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+goog.module('TestHelpers');
```
cpcallen, line 7: Perhaps `Blockly.test.helpers`?

cpcallen (commented):
Looks good, but for some small stylistic issues.

I have commits in my local branch to apply the name changes (and convert a few more files to `goog.module`) that I can push to this PR if you like.

alschmiedt:
Happy to have you push whatever changes you have so far, and then I can finish up converting the rest of the files and applying any other comments you have on top of that.

cpcallen:
@googlebot I consent.

scripts/gulpfiles/build_tasks.js
alschmiedt: done

scripts/gulpfiles/build_tasks.js
alschmiedt: Sounds good to me. Done.

cpcallen (approved):
I think "Fix imports and workspace tests" might be a little under-descriptive of the changes actually contained in that commit: I was fairly confused about why you were moving most of the contents of `workspace_test.js` to `workspace_helpers.js` until I saw that `testAWorkspace` is now _also_ being called from `workspace_svg_test.js`!

PR #5441, opened 2021-09-08
Rename `Blockly.utils.IdGenerator` to `idGenerator` and move `Blockly.utils.genUid` to there

cpcallen:
## The basics

- [x] I have run `npm test`.

(Tests will fail until the `dev-tools` package from blockly-samples is published.)

### Resolves

Part of #5026

### Proposed Changes

* Rename the `Blockly.utils.IdGenerator` module to `Blockly.utils.idGenerator` (because does not have a class as its default export).
* Move `genUid` from `Blockly.utils` to `Blockly.utils.idGenerator`, and provide a mechanism to allow it to be stubbed for tests in a way that should continue to work despite ongoing module migration work.

### Test Coverage

Awaiting push of `dev-tools` package from blockly-samples.  Tests will fail until then.

### Additional Information

See also google/blockly-samples#876.


core/utils.js
```diff
@@ -402,29 +403,38 @@ Blockly.utils.tokenizeInterpolation_ = function(message,
...
  */
-Blockly.utils.genUid = function() {
-  var length = 20;
-  var soupLength = Blockly.utils.genUid.soup_.length;
-  var id = [];
-  for (var i = 0; i < length; i++) {
-    id[i] = Blockly.utils.genUid.soup_.charAt(Math.random() * soupLength);
+Blockly.utils.genUid = function() {};  // For type declaration.
+
+Object.defineProperties(Blockly.utils, {
```
rachel-fenichel, line 420: We no longer need the getter/setter pair. `genUid` can just do a `deprecation.warn` and then call `utils.idGenerator.genUid`.

core/utils/idgenerator.js
```diff
@@ -5,34 +5,74 @@
...
+  const soupLength = soup.length;
+  const id = [];
+  for (let i = 0; i < length; i++) {
+    id[i] = soup.charAt(Math.random() * soupLength);
+  }
+  return id.join('');
+};
+
+/**
+ * Generate a random unique ID.  This should be globally unique.
```
rachel-fenichel, line 71: Move this documentation to `internal.genUid` and then either duplicate it or use `@see`. Better to have it attached to the full implementation.

core/utils.js
cpcallen: Fixed.

core/utils/idgenerator.js
cpcallen: OK.  I'd put it here so that someone browsing auto-generated API docs for the module would see it attached to the part of the API we want them to use, but I've re-arranged it as you suggested.

rachel-fenichel (approved):

PR #5448, opened 2021-09-09
Migrate core/msg.js to goog.module syntax

cpcallen:
## The basics

- [X] My code is presented in the form suggested in the [module
- [X] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/msg.js` to `goog.module` with ES6 `const`/`let`.

### Additional Information

There is some doubt about whether this PR will cause issues with localisation because:

* The `exports` object of a `goog.module` gets frozen upon being loaded.
* This will prevent any additional properties from being added to it.
* `msg.js` doesn't actually add any message properties to it at all.
* How is the playground successfully loading messages after the fact?
* Will this continue to work with advanced compilation?

### DIFF
--- a/core/msg.js
+++ b/core/msg.js
@@ -17,16 +17,16 @@
 goog.module('Blockly.Msg');
 goog.module.declareLegacyNamespace();
 
-goog.require('Blockly.utils.global');
+const global = goog.require('Blockly.utils.global');
 
 
 /**
  * Exported so that if Blockly is compiled with ADVANCED_COMPILATION,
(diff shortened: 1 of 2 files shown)


cpcallen:
@NeilFraser Neil: I've assigned this PR to you with a view to testing to make sure that the code in this branch will work successfully with Blockly Games before this PR gets merged.  (Though if the problem is caused by code already in the `goog_module` branch then that will need to be fixed on Blockly Games' end.)

rachel-fenichel (approved):

PR #5453, opened 2021-09-10
Move bump objects function into a separate module

rachel-fenichel:
## The basics

- [x] I branched from develop
- [x] My pull request is against *goog_module*

### Resolves

Part of #5026.

### Proposed Changes

Moves the `bumpObjectsIntoBounds_` function out of inject and into its own module, and fixed its uses.
Removes `Blockly.bumpObjectsIntoBounds_`, which was package.

#### Behavior Before Change

No change

#### Behavior After Change

No change

### Reason for Changes

The function wasn't actually on the inject namespace--it was on the top level namespace--but it didn't need to be there.

However, having it in inject.js made converting inject to module quite difficult.

### Test Coverage
Tested by opening both the normal playground and the advanced playground, and by setting the advanced playground to not have scroll bars to manually check the bump behaviour.

### Additional Information

This uses a combination of `goog.module.get` and `goog.require` to use a module in a provide file. That will go away when I go on to convert `inject.js`, but I wanted this to be a separate change for reviewing.

I *think* I'm going to move other functions from `inject.js` into `bump_objects.js`, but this was the minimum change to create the module.


core/bump_objects.js
```diff
@@ -0,0 +1,72 @@
...
+goog.module('Blockly.bumpObjects');
+
+/* eslint-disable-next-line no-unused-vars */
+const IBoundedElement = goog.requireType('Blockly.IBoundedElement');
+/* eslint-disable-next-line no-unused-vars */
+const MetricsManager = goog.requireType('Blockly.MetricsManager');
+/* eslint-disable-next-line no-unused-vars */
+const WorkspaceSvg = goog.requireType('Blockly.WorkspaceSvg');
+const mathUtils = goog.require('Blockly.utils.math');
+
```
cpcallen, line 22:
```suggestion


```

core/inject.js
```diff
@@ -239,7 +239,7 @@ Blockly.bumpTopObjectsIntoBounds_ = function(workspace) {
   var scrollMetricsInWsCoords = metricsManager.getScrollMetrics(true);
   var topBlocks = workspace.getTopBoundedElements();
   for (var i = 0, block; (block = topBlocks[i]); i++) {
-    Blockly.bumpObjectIntoBounds_(workspace, scrollMetricsInWsCoords, block);
+    goog.module.get('Blockly.bumpObjects').bumpIntoBounds(workspace, scrollMetricsInWsCoords, block);
```
cpcallen, line 242: This is indeed correct usage of `get`, but I think I'd have opted to do a `goog.module.declareLegacyNamespace` at the top of `bump_objects.js` rather than this `goog.module.get` call, because the former is less likely to be overlooked in future cleanup passes.

cpcallen (approved):
Any particular reason not to also move `bumpTopObjectsIntoBounds_` and `bumpIntoBoundsHandler_`?

core/inject.js
cpcallen: > because the former is less likely to be overlooked in future cleanup passes.

(Well: I suspect that _I would be less likely to overlook the former in future cleanup passes_, anyway…)

cpcallen:
> Any particular reason not to also move `bumpTopObjectsIntoBounds_` and `bumpIntoBoundsHandler_`?

Oh, I see you've done this in the next PR.

rachel-fenichel:
> Any particular reason not to also move bumpTopObjectsIntoBounds_ and bumpIntoBoundsHandler_?

This is the only function that is called from outside of inject. The others are actually file-internal and were on the Blockly namespace to avoid making them completely global, so they were not directly blocking. But yes, I'm moving them in #5456.

PR #5454, opened 2021-09-10
chore: sets polyfill flag to true

### LINKED ISSUE #5039: Remove rewrite_polyfills=false
**Describe the bug**

We have the closure compiler `rewrite_polyfills` flag set to false.

https://github.com/google/blockly/blob/master/scripts/gulpfiles/build_tasks.js#L120

The flag documentation says:
"--rewrite_polyfills Rewrite ES6 library calls to use polyfills provided by the compiler's runtime. (default: true)"

Setting this to true may increase the size of our generated files.

Flag documentation: https://github.com/google/closure-compiler/wiki/Flags-and-Options

alschmiedt:
### Resolves
#5039

### Additional Details

Updating this flag caused the following changes in the compressed files:

**Old**
blockly_compressed: **795798**
blockly_compressed.js.gz: **151736**

**New**
blockly_compressed: **800683**
blockly_compressed.js.gz: **153167**

**Deltas** 
blockly_compressed: **4885** (~5kb)
blockly_compressed.js.gz: **1431**(~1.5kb)

**Note**: this is for about 6 polyfills(as far as I could tell). As we increase the number of methods that need polyfills, the size will increase. 

**How I tested?** 
- Changed the flag in the `build_tasks.js` file
- `npm run build`
- `./tests/scripts/check_metadata.sh` to get the new size.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -119,7 +119,7 @@ function compile(compilerOptions, opt_verbose, opt_warnings_as_error,
   options.warning_level = opt_verbose ? 'VERBOSE' : 'DEFAULT';
   options.language_in = 'ECMASCRIPT6_STRICT',
   options.language_out = 'ECMASCRIPT5_STRICT';
-  options.rewrite_polyfills = false;
+  options.rewrite_polyfills = true;
   options.hide_warnings_for = 'node_modules';
   if (opt_warnings_as_error || opt_strict_typechecker) {
     options.jscomp_error = JSCOMP_ERROR;


rachel-fenichel:
Your tests checked the size, but probably didn't check that the polyfills actually worked. Please use browserstack or similar to test with IE11, which is most likely to break with some of the new language stuff we're using.

alschmiedt:
**Tl;dr I think the polyfills and transpiling are working fine, but we need a better way to test ie11 going forward that will allow us to test workspace comments and test blocks.**

Here are my findings:

1. We can no longer use `blockly_uncompressed.js` on ie11. We already knew this was coming, just thought I would reiterate for context. 
2. I think we need a better way to test ie11. Even if we decide to switch our playground to use `blockly_compressed.js` every time we want to test on ie11 (I don't think we want to do that) it still does not work because we have a template string in `playground.html`. We can take that out, but if we aren't going to be using the playground for testing ie11 then I'm not sure I see the point.
3. Even if we take out the string literal right clicking does not work because we try to include workspace comment using `goog.require`. `goog` is not available when we load the compressed version, so if this is something we want to test on ie11 then we need to find a different way to include it.
4. Test blocks `index.js` has a template string in it that is causing it to not load on ie11. This is a bit troubling because I thought we were transpiling and in our dist folder when run locally it does look like it is appropriately transpiled (no template strings).

cpcallen:
> 1. We can no longer use `blockly_uncompressed.js` on ie11. We already knew this was coming, just thought I would reiterate for context.

That is as expected.

> 2. I think we need a better way to test ie11. Even if we decide to switch our playground to use `blockly_compressed.js` every time we want to test on ie11 (I don't think we want to do that) it still does not work because we have a template string in `playground.html`. We can take that out, but if we aren't going to be using the playground for testing ie11 then I'm not sure I see the point.

Making sure that `playground.html` can be opened in ie11 doesn't seem like a bad idea.  It ought to be possible to have it selectively load either `blockly_compressed.js` or `blockly_uncompressed.js` depending on what browser you are using, too.

> 3. Even if we take out the string literal right clicking does not work because we try to include workspace comment using `goog.require`. `goog` is not available when we load the compressed version, so if this is something we want to test on ie11 then we need to find a different way to include it.

Probably by using a chunked compile, I suspect.  🤔

> 4. Test blocks `index.js` has a template string in it that is causing it to not load on ie11. This is a bit troubling because I thought we were transpiling and in our dist folder when run locally it does look like it is appropriately transpiled (no template strings).

That merits further investigation.  Can you file a bug, please?

cpcallen (commented):
Did you mean to push commits 6dc53f9 and 927de5d?

I don't think we want to move to using `blockly_compressed.js` by default.

alschmiedt:
> Did you mean to push commits [6dc53f9](https://github.com/google/blockly/commit/6dc53f9d9c9cf3dec32b0e3d4f99efed7ea6c734) and [927de5d](https://github.com/google/blockly/commit/927de5d1d37664affdfbc2b265f3776b2393aea7)?

This was part of my attempts at finding a good way to test ie11. Reverted. 

> It ought to be possible to have it selectively load either `blockly_compressed.js` or `blockly_uncompressed.js` depending on what browser you are using, too.

Filed #5482 to track this. I mention workspace comments and using a chunked compile in the additional context section.

> That merits further investigation. Can you file a bug, please?

Issue filed google/blockly-samples#887.

cpcallen (approved):

PR #5457, opened 2021-09-13
Move alert/confirm/prompt to a new file, dialog.js

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5208

### Proposed Changes

This PR moves Blockly.alert/confirm/prompt into a new Blockly.dialog namespace in core/dialog.js. It also migrates uses in core to use Blockly.dialog.* instead of Blockly.*, and adds (deprecated) setters and getters under the old name in Blockly for backwards compatibility.


core/dialog.js
```diff
@@ -0,0 +1,81 @@
...
+  callback(window.prompt(message, defaultValue));
+};
+
+/**
+ * Wrapper to window.alert() that app developers may override to
+ * provide alternatives to the modal browser window.
+ * @param {string} message The message to display to the user.
+ * @param {function()=} opt_callback The callback when the alert is dismissed.
+ */
+const alert = function(message, opt_callback) {
```
rachel-fenichel, line 38: Why create a wrapper here, instead of just doing `exports.alert = alertImplementation`?

gonfunko: Because that exports the initial value of alertImplementation, so even if alertImplementation gets modified via the setter, the initial/default implementation will continue to be used.

rachel-fenichel: Right, that makes sense.

The comment is no longer correct (for this and the others) because to override you would call `setAlert` rather than overriding `alert`.

gonfunko: Updated the JSDoc to clarify, thanks!

rachel-fenichel (approved):

PR #5459, opened 2021-09-14
Migrate core/utils/xml.js to goog.module syntax

cpcallen:
## The basics

- [X] My code is presented in the form suggested in the [module
- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/utils/xml.js` to `goog.module` with ES6 `const`/`let`.

### Additional Information

The `.documents` export is monkey-patched in `scripts/package/node/core.js`; removing this monkey-patching will be addressed in a separate PR.

### DIFF
--- a/core/utils/xml.js
+++ b/core/utils/xml.js
@@ -63,7 +63,7 @@ Blockly.utils.xml.createTextNode = function(text) {
  * @public
  */
 Blockly.utils.xml.textToDomDocument = function(text) {
-  var oParser = new DOMParser();
+  const oParser = new DOMParser();
   return oParser.parseFromString(text, 'text/xml');
 };
 
@@ -75,6 +75,6 @@ Blockly.utils.xml.textToDomDocument = function(text) {
  * @public
  */
 Blockly.utils.xml.domToText = function(dom) {
-  var oSerializer = new XMLSerializer();
+  const oSerializer = new XMLSerializer();
   return oSerializer.serializeToString(dom);
 };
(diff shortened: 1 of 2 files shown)


core/utils/xml.js
```diff
@@ -16,65 +16,71 @@
...
+const NAME_SPACE = 'https://developers.google.com/blockly/xml';
+exports.NAME_SPACE = NAME_SPACE;
 
 /**
  * Get the document object.  This method is overridden in the Node.js build of
  * Blockly. See gulpfile.js, package-blockly-node task.
+ *
+ * Note that this function is named myDocument so as to not shadow the
+ * global of the same name, but (for now) exported as .document to not
+ * break existing importers.
```
BeksOmega, line 35: Three things:
  1) myDocument -> getDocument
  2) Does this break [node packaging](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/scripts/package/node/core.js#L25)?
  3) Do you want to export `getDocument` as well?

cpcallen:
Hmm, apparently I didn't run `npm test` because CI tests are failing.

BeksOmega (approved):
LGTM but I'm not sure if changing `.document` causes a problem with node packaging or not.

core/utils/xml.js
cpcallen: This renaming _should_ have been entirely internal and totally fine from the point of view of any users of this package (and therefore should not have been able to break the node packaging), _except that I modified `createElement` and `createTextNode` to call `getDocument()` instead of `exports.document()`, making it impossible to monkey-patch these calls_.  Ooops.

I want to fix the money patching properly but for now I will just modify these two internal calls to call `exports.document` so that everything works the way it previously did.

cpcallen: 1. Fixed.
2. See previous comment.
3. Yes, but not in this PR.  In the next PR I'll export `getDocument` and also a `setDocument`, and use the latter instead of monkey-patching.

BeksOmega (approved):

PR #5462, opened 2021-09-14
Migrate Blockly.hueToHex to Blockly.utils.colour.hueToHex

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5208

### Proposed Changes

This PR migrates Blockly.hueToHex to Blockly.utils.colour.hueToHex and updates core and demos to use the new function.

### DIFF
--- a/demos/blockfactory/workspacefactory/wfactory_controller.js
+++ b/demos/blockfactory/workspacefactory/wfactory_controller.js
@@ -590,7 +590,8 @@ WorkspaceFactoryController.prototype.loadCategoryByName = function(name) {
   }
   if (!standardCategory.colour && standardCategory.hue !== undefined) {
     // Calculate the hex colour based on the hue.
-    standardCategory.colour = Blockly.hueToHex(standardCategory.hue);
+    standardCategory.colour = Blockly.utils.colour.hueToHex(
+      standardCategory.hue);
   }
   // Transfers current flyout blocks to a category if it's the first category
   // created.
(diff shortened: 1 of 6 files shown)


maribethb (approved):

demos/blockfactory/workspacefactory/wfactory_controller.js
```diff
@@ -590,7 +590,8 @@ WorkspaceFactoryController.prototype.loadCategoryByName = function(name) {
   }
   if (!standardCategory.colour && standardCategory.hue !== undefined) {
     // Calculate the hex colour based on the hue.
-    standardCategory.colour = Blockly.hueToHex(standardCategory.hue);
+    standardCategory.colour = Blockly.utils.colour.hueToHex(
```
maribethb, line 593: This is actually a breaking change in demos, at least until we check in an updated `blockly_compressed.js`. I'm running into this now with my similar change, and not sure if we should a) make the change anyway which breaks the demo but hope it works whenever we do check in blockly_compressed b) check in blockly_compressed now or c) hold off on making the change to demos until after we check in blockly_compressed

we've done option a here but not sure if that was intentional and if I should continue doing that with other changes. any thoughts?

BeksOmega: When I made changes to the code demo to add JSON serialization I tested w/ compressed to make sure it worked, but didn't check it in.

gonfunko: I was looking at this from the perspective of it being easy to forget to backport these changes later, and assuming that we'll wind up needing to do a round of regression testing/fixing regardless. Happy to revert the changes to the files in demo if you think that's a better idea though!

Issue #5428, opened 2021-09-03
Review contents of `utils.js`


cpcallen:
`core/utils.js`, which provides `Blockly.utils`.  The contents of `core/utils/*.js` formerly provided leaf modules `Blockly.utils.*`; they have now been migrated to `goog.module` and (provisionally) `declareLegacyNamespace`; ~~this is unfortunately incompatible with having `utils.js` be a `goog.module`~~.

We were considering evacuating and removing `Blockly.utils` entirely, but this is probably not the correct course of action.  Instead, we will probably remove the calls to `declareLegacyNamespace` from `core/utils/*.js` and have this module import and reexport those leaf modules instead.

It would nevertheless be sensible to consider whether any/all of the contents of this module should be moved to other module(s) in any case.

Needs decision:
* [x] `runAfterPageLoad`
* [x] `arrayRemove`
* [x] `getBlockTypeCounts`
* [x] `noEvent`
* [x] `isTargetInput`
* [x] `getRelativeXY`
* [x] `getInjectionDivXY_`
* [x] `isRightButton`
* [x] `mouseToSvg`
* [x] `getScrollDeltaPixels`
* [x] `tokenizeInterpolation`
* [x] `replaceMessageReferences`
* [x] `checkMessageReferences`
* [x] `genUid`
* [x] `is3dSupported`
* [x] `getViewportBBox`
* [x] `getDocumentScroll`
* [x] `screenToWsCoordinates`
* [x] `parseBlockColour`

rachel-fenichel:
Groupings that I see so far:

### Browser events
These take in an `Event` and do some checks or extract some information. I propose that we move these to `browserEvents`

- `isTargetInput`
- `noEvent` -- this is also only used in one place (contextmenu.js, to prevent the system context menu), so maybe just move the definition in there and get rid of this function entirely?
- `isRightButton`
- `mouseToSvg`
- `getScrollDeltaPixels`

### Coordinates
These could go in a new file for coordinate/svg math specifically--or perhaps there's an existing file that they could live in? I would say `Coordinate`, but that's a class. The common thing I see here is that these inspect the dom to figure out where something is/how to compute where something is. There are some regexes that go along with these but can be module-internal in whichever file they end up in.

- `getRelativeXY` -- used quite a bit, depends on the XY regex and on `Blockly.utils.Coordinate`
- `getInjectionDivXY_` -- only used internally
- `getViewportBBox`
- `getDocumentScroll`
- `is3dSupported`
[...]

PR #5464, opened 2021-09-14
Move browser event utils from utils.js to browser_events.js

BeksOmega:
## The basics

- [ ] My code is presented in the form suggested in the [module
- [x] I have run `npm test`.

### Resolves

Part of #5428 

### Proposed Changes

Moves `isTargetInput`, `isRightButton`, `mouseToSvg`, and `getScrollDeltaPixels` to the `browserEvents` namespace.

Deprecates `noEvent` and adds a private implementation in `context_menu.js`, since this is the only place it's used.


core/utils.js
```diff
@@ -36,31 +37,21 @@ const style = goog.require('Blockly.utils.style');
...
- */
-const noEvent = function(e) {
-  // This event has been handled.  No need to bubble up to the document.
-  e.preventDefault();
-  e.stopPropagation();
-};
-exports.noEvent = noEvent;
+Object.defineProperty(exports, 'noEvent', {
+  get: function() {
+    deprecation.warn(
```
gonfunko, line 52: Here and elsewhere in this file, it might be useful to include the 4th arg to warn that specifies what you should use instead.

core/utils.js
```diff
@@ -36,31 +37,21 @@ const style = goog.require('Blockly.utils.style');
...
- * Don't do anything for this event, just halt propagation.
- * @param {!Event} e An event.
- */
-const noEvent = function(e) {
-  // This event has been handled.  No need to bubble up to the document.
-  e.preventDefault();
-  e.stopPropagation();
-};
-exports.noEvent = noEvent;
+Object.defineProperty(exports, 'noEvent', {
```
gonfunko, line 50: Check with Chris about this, but I vaguely recall him mentioning that we should be using defineProperties instead (I think due to quoted field names not playing nicely with obfuscation? I also realized I might need to change this in a couple of my own recent PRs)

BeksOmega: @cpcallen

cpcallen: Yes: `Object.defineProperties` is preferred to `Object.defineProperty` because the Closure Compiler in advanced optimization mode will successfully rename the properties supplied as keys in the argument object, whereas it will not touch a property name supplied as a string.

But to step back a level here: why are you providing  a get accessor here when it looks like a stub function would work just as well?:

```
/**
 * Don't do anything for this event, just halt propagation.
 * @deprecated Use Blockly.browserEvents.noEvent() instead.
 * @param {!Event} e An event.
 */
const noEvent = function(e) {
  deprecation.warn(
      'Blockly.utils.noEvent', 'September 2021', 'September 2022', /* etc */);
  browserEvents.noEvent(e);
};
exports.noEvent = noEvent;
```

(Or have I misunderstood what is going on here?)

cpcallen (changes requested):
Please add entries in `scripts/migration/renamings.js` for the functions that have been moved.

core/utils.js
BeksOmega: > why are you providing a get accessor here when it looks like a stub function would work just as well?

No reason, I just thought that's what I was supposed to do haha. I'll switch them over 👍

core/contextmenu.js
```diff
@@ -184,12 +184,21 @@ const createWidget_ = function(menu) {
...
+      /** @type {!EventTarget} */ (menuDom), 'contextmenu', null, noEvent);
   // Focus only after the initial render to avoid issue #1329.
   menu.focus();
 };
 
+/**
+ * Halts the propagation of the event without doing anything else.
+ * @param {!Event} e An event.
+ */
+const noEvent = function(e) {
```
cpcallen, line 196: I'm wondering if `haltPropagation` might be a better name for this function?

core/utils.js
```diff
@@ -37,28 +38,23 @@ const userAgent = goog.require('Blockly.utils.userAgent');
 
 
 /**
- * Don't do anything for this event, just halt propagation.
+ * Halts the propagation of the event without doing anything else.
```
cpcallen, line 41: Perhaps a bit belt-and-braces, but why not mark this as `@deprecated`?  (Here and everywhere below that calls `deprecation.warn`.)

core/utils.js
```diff
@@ -37,28 +38,23 @@ const userAgent = goog.require('Blockly.utils.userAgent');
...
-  return e.target.type == 'textarea' || e.target.type == 'text' ||
-      e.target.type == 'number' || e.target.type == 'email' ||
-      e.target.type == 'password' || e.target.type == 'search' ||
-      e.target.type == 'tel' || e.target.type == 'url' ||
-      e.target.isContentEditable ||
-      (e.target.dataset && e.target.dataset.isTextInput == 'true');
+  deprecation.warn(
+    'Blockly.utils.isTargetInput', 'September 2021', 'September 2022',
+    'Blockly.browserEvets.isTargetInput');
+  return browserEvents.isTargetInput;
```
cpcallen, line 57: Doesn't this need parens?:

```suggestion
  return browserEvents.isTargetInput();
```

core/utils.js
```diff
@@ -37,28 +38,23 @@ const userAgent = goog.require('Blockly.utils.userAgent');
...
   e.preventDefault();
   e.stopPropagation();
 };
 exports.noEvent = noEvent;
 
-/**
- * Is this event targeting a text input widget?
- * @param {!Event} e An event.
- * @return {boolean} True if text input.
- */
```
cpcallen, line 54: I'd preserve the docs until the function is deleted.

cpcallen (changes requested):

cpcallen (approved):

PR #5465, opened 2021-09-14
Re-export `Blockly.utils.*` submodules from `Blockly.utils`

cpcallen:
## The basics

- [X] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Remove the Have `utils.js` reexport all of the `Blockly.utils.*` modules in `core/utils/*.js`

### Additional Information

I'm going to mark this as draft until the issues surrounding versioning / unpkging of `dev-tools` are clarified a bit.


alschmiedt:
One of the questions that came up from this PR was: 

**Should test helpers be in core instead of separate plugin that core has to import?** 
Yes, it makes more sense for them to be in core eventually. However, because we have not figured out multiple entry points having these in core causes a size concern. Once we figure out multiple entry points this should be moved into core from blockly samples. All of the plugins that rely on the test helpers would then be able to get the test helpers from the blockly version they already install. I added #5466 to track the issue.

cpcallen:
OK, this just needs a code review.

rachel-fenichel (approved):
LGTM

PR #5467, opened 2021-09-15
test: Changes mocha tests to use node_modules instead of unpkg

alschmiedt:
### Proposed Changes
Changes mocha tests to use node modules instead of unpkg.

#### Behavior Before Change
Tests use latest version on unpkg.

#### Behavior After Change
Tests use version that is installed in node modules.

### Reason for Changes
It is confusing to install a version of mocha and then use a different version from unpkg.

### DIFF
--- a/tests/mocha/index.html
+++ b/tests/mocha/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <title>Mocha Tests for Blockly</title>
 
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
This works, but please file a bug to track updating mocha tests to use imports rather than script tags.

PR #5469, opened 2021-09-15
Reinstall to update package-lock.json

rachel-fenichel:
- Branched from `goog_module`
- Created by updating my npm version to 7.23.0 and running `npm install`
- This is also a test for whether the build issues we're seeing on #5465 are actually because of this update


cpcallen:
Closing this as I don't think it will be needed: #5465 has now passed CI.

PR #5470, opened 2021-09-15
fix: migrate svg math utils to new svgMath namespace

BeksOmega:
## The basics

- [x] I branched from `goog_module`
- [x] My pull request is against `goog_module`
- [x] My code follows the [style guide](
      https://developers.google.com/blockly/guides/modify/web/style-guide)
- [ ] My code is presented in the form suggested in the [module
      conversion guide](https://github.com/google/blockly/issues/5026)
- [x] I have run `npm test`.

### Resolves

Part of #5428 

### Proposed Changes

Moves getRelativeXY, getInjectionDivXY, getViewportBBox, getDocumentScroll, and is3dSupported to a new namespace.

Moves screenToWsCoordinates to a property of the workspace, and renames it to screenToWorkspaceXY, for consistency.

PR #5479, opened 2021-09-16
Make several exported private fields/methods TEST_ONLY

### LINKED ISSUE #5135: Remove or deprecate exported names ending in `_`
### Background

During the first pass converting `goog.provide` to `goog.module`, we are going export most or all existing exported identifiers, even ones marked `@private` and ending in underscore (`_`).  This is being done to ensure that the initial conversion does not break other code within our own codebase (as well as in those of Blockly embedders).

### Issue

Since:
* it doesn't really make sense to export identifiers that are supposed to be private, and
* we do not want external developers using these not-intended-to-be-public exports.

### Suggested fix

We should either:
* remove these exports (where doing does not and is not expected to break anything), or
* mark them as `@deprecated` and wrapped with a call to the `warn` method from `Blockly.utils.deprecation`.

### Alternatives considered

Doing nothing, and leaving them exported and not marked for deprecation.  This runs the risk of them becoming de-facto parts of the API we need to continue to support in the future.

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

#5135

### Proposed Changes
This PR addresses exported private fields and methods in Blockly. They have been resolved as follows:

* `Blockly.Events.disabled_` -> Removed, and tests rewritten to use public API to accomplish the same thing
* `Blockly.Events.FIRE_QUEUE_`, `Blockly.Events.fireNow()` -> Moved into TEST_ONLY export and tests updated accordingly
* `Blockly.Extensions.ALL_` -> Moved into TEST_ONLY export (as allExtensions) and tests updated accordingly
* `Blockly.registry.typeMap_` -> Moved into TEST_ONLY export and tests updated accordingly

Because all these private exports were used exclusively in tests, I did not include deprecation wrappers; if we have reason to think external developers were using them I'm happy to add those.

The only other "private" export I found was `Blockly.utils.getInjectionDivXY_()`, but despite ending in an underscore, this was not marked @private pre-conversion, is not used internally in utils or in tests, and is used in workspace and scrollbar. Beka has dealt with it in #5470, so I'm ignoring it here.

### DIFF
--- a/core/events/events.js
+++ b/core/events/events.js
@@ -90,17 +90,6 @@ Object.defineProperties(exports, {
  * @type {number}
  */
 let disabled = 0;
-/** @private */
-exports.disabled_ = disabled;
-
-
-Object.defineProperties(exports, {
-  disabled_: {
-    set: function(newValue) {
-      disabled = newValue;
-    },
-  },
-});
 
 /**
  * Name of event that creates a block. Will be deprecated for BLOCK_CREATE.
@@ -310,8 +299,6 @@ exports.BUMP_EVENTS = BUMP_EVENTS;
  * List of events queued for firing.
  */
 const FIRE_QUEUE = [];
-/** @private */
-exports.FIRE_QUEUE_ = FIRE_QUEUE;
 
 /**
  * Create a custom event and fire it.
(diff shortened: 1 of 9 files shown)


NeilFraser (approved):
The TEST_ONLY in the title made me think this wasn't a real PR to commit.  :)

PR #5484, opened 2021-09-16
fix: move message utils to a new file

BeksOmega:
## The basics

- [ ] My code is presented in the form suggested in the [module conversion guide](https://github.com/google/blockly/issues/5026)
- [x] I have run `npm test`. -> But I ran into test errors that also repro on goog_module

### Resolves

Part of #5428 

### Proposed Changes

Moves tokenizeInterpolation, replaceMessageReferences, and checkMessageReferences to a new utils.messages namespace. Also renames
  * replaceMessageReferences -> messages.replaceReferences
  * checkMessageReferences -> messages.checkReferences.


core/utils/messages.js
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Internal implementation of the message reference and interpolation token
+ * parsing used by tokenizeInterpolation() and replaceMessageReferences().
+ * @param {string} message Text which might contain string table references and
+ *     interpolation tokens.
+ * @param {boolean} parseInterpolationTokens Option to parse numeric
+ *     interpolation tokens (%1, %2, ...) when true.
+ * @return {!Array<string|number>} Array of strings and numbers.
+ */
+const tokenizeInterpolation_ = function(message, parseInterpolationTokens) {
```
maribethb, line 88: This is not marked private, but it has the underscore. The old function was also not marked private, but I don't have context on if that's intentional and this is another `bindEvent_` scenario where we made this available for external folks.

I think either we:
a) Intentionally make this public, remove the underscore (and rename it to avoid clashing), and add a deprecation notice to the old function (and add it to renames file)
b) make this private and keep the underscore

depending on the intent with this function. I don't have context on that though, maybe Rachel?

BeksOmega: I think it's private by default because it's not exported. How would you feel about renaming it to `tokenizeInterpolationInternal` but otherwise leaving as is?

core/block.js
```diff
@@ -44,6 +44,7 @@ const constants = goog.require('Blockly.constants');
 const fieldRegistry = goog.require('Blockly.fieldRegistry');
 const idGenerator = goog.require('Blockly.utils.idGenerator');
 const inputTypes = goog.require('Blockly.inputTypes');
+const messages = goog.require('Blockly.utils.messages');
```
BeksOmega, line 47: @cpcallen Since this is new, does this need an `exports.messages` assignment? or no?

cpcallen: If it is intended to be part of the Blockly API then yes: it should be reexported in `utils.js`.  (Same for `browserEvents` added in #5464.)

BeksOmega: Cool cool 👍 I think browser events are fine though because it's just `Blockly.browserEvents`, and it already existed.

cpcallen: You are quite correct.  Evidently it is getting late here…

BeksOmega: node did not like it and I don't know why :/ Last commit passes tests just fine but after adding `exports.messages = messages` it gets angry.

core/utils/messages.js
maribethb: Oh, duh, forgot to check if it's exported. But good call on renaming, since I think we moved away from underscores on non-exported private methods. Thanks!

cpcallen:
@maribethb: Will you finish this for Beka, or would you like to reassign it?

rachel-fenichel:
This does not have to go in for the release--let's leave it alone for now.

rachel-fenichel:
Closing in favor of #5706 (same changes but on develop).

PR #5489, opened 2021-09-17
Migrate selected from Blockly to Blockly.common

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5208

### Proposed Changes
Moves Blockly.selected to Blockly.common.getSelected/setSelected, and refactors core to use the new method.


core/shortcut_items.js
```diff
@@ -63,8 +63,8 @@ const registerDelete = function() {
   const deleteShortcut = {
     name: names.DELETE,
     preconditionFn: function(workspace) {
-      return !workspace.options.readOnly && Blockly.selected &&
-          Blockly.selected.isDeletable();
+      return !workspace.options.readOnly && common.getSelected() &&
+          common.getSelected().isDeletable();
```
cpcallen, line 67: Not sure if it's OK to use yet, but I observe this would be a perfect place to use the optional chaining operator: `common.getSelected()?.isDeletable()`.

core/shortcut_items.js
```diff
@@ -94,15 +94,15 @@ const registerCopy = function() {
     name: names.COPY,
     preconditionFn: function(workspace) {
       return !workspace.options.readOnly && !Gesture.inProgress() &&
-          Blockly.selected && Blockly.selected.isDeletable() &&
-          Blockly.selected.isMovable();
+          common.getSelected() && common.getSelected().isDeletable() &&
+          common.getSelected().isMovable();
```
cpcallen, line 98: I wonder if this is the point at which, even if we can't use optional chaining, we ought instead to do `const selected = common.getSelected();` and then just check `selected.isDeletable() && selected.isMovable()`.

core/shortcut_items.js
```diff
@@ -129,16 +129,18 @@ const registerCut = function() {
     name: names.CUT,
     preconditionFn: function(workspace) {
       return !workspace.options.readOnly && !Gesture.inProgress() &&
-          Blockly.selected && Blockly.selected.isDeletable() &&
-          Blockly.selected.isMovable() && !Blockly.selected.workspace.isFlyout;
+          common.getSelected() && common.getSelected().isDeletable() &&
+          common.getSelected().isMovable() &&
+          !common.getSelected().workspace.isFlyout;
```
cpcallen, line 134: OK: maybe *this* is the point where we do that?

core/blockly.js
```diff
@@ -80,12 +80,6 @@ Object.defineProperty(Blockly, 'mainWorkspace', {
   }
 });
 
-/**
- * Currently selected block.
- * @type {?Blockly.ICopyable}
- */
-Blockly.selected = null;
```
cpcallen, line 87: I believe we will be able to do get/set accessors (despite the current difficulties 🤞), and this probably should have a pair.

cpcallen:
Oh, and there should be an entry in `renamings.js` too.

core/blockly.js
gonfunko: Done!

core/shortcut_items.js
gonfunko: Updated

core/shortcut_items.js
gonfunko: Done

core/shortcut_items.js
gonfunko: Also not sure if this is OK yet so leaving it as-is; agreed this will be nice once we can though!

cpcallen (approved):

PR #5494, opened 2021-09-17
Migrate core/blockly.js to goog.module syntax

gonfunko:
## The basics

- [x] I have run `npm test`.

### Resolves

Part of #5026

### Proposed Changes

Converts `core/blockly.js` to `goog.module` with ES6 `const`/`let`.


core/blockly.js
```diff
@@ -63,37 +64,41 @@ goog.requireType('Blockly.Workspace');
...
-      'Blockly.common.getSelected()');
-    return Blockly.common.getSelected();
-  },
-  set: function(newSelection) {
-    Blockly.utils.deprecation.warn(
-      'Blockly.selected', 'September 2021', 'September 2022',
-      'Blockly.common.setSelected()');
-    Blockly.common.setSelected(newSelection);
-  }
+Object.defineProperties(exports, {
```
cpcallen, line 89: Since `defineProperties` can do more than one, this could be combined with the previous into a single call.  (And another similar case below.)

core/blockly.js
```diff
@@ -245,28 +256,32 @@ Blockly.defineBlocksWithJsonArray = function(jsonArray) {
...
+  hueToHex: {
+    get: function() {
+      deprecation.warn(
+          'Blockly.hueToHex()', 'September 2021', 'September 2022',
+          'Blockly.utils.colour.hueToHex()');
+      return colour.hueToHex;
+    }
   }
 });
```
cpcallen, line 286: Only tangentially related to this PR, but since you're touching the accessors anyway:

Since there's no `set` accessor I'm sure that this accessor property does _not_ exist for the purpose of allowing library users to supply their own `hueToHex` implementation.  This raises the question: since `hueToHex` is just an ordinary function (rather than a formerly mutable export), why not provide a simple wrapper for it?:

```suggestion
/**
 * Convert a hue (HSV model) into an RGB hex triplet.
 * @deprecated Use Blockly.utils.co.our.hueToHex instead.
 * @param {number} hue Hue on a colour wheel (0-360).
 * @return {string} RGB code, e.g. '#5ba65b'.
 */
const hueToHex = function(hue) {
  Blockly.utils.deprecation.warn(
      'Blockly.hueToHex()', 'September 2021', 'September 2022',
      'Blockly.utils.colour.hueToHex()');
  return colour.hueToHex(hue);
};
exports.hueToHex = hueToHex;
```

cpcallen (approved):

core/blockly.js
gonfunko: Done

core/blockly.js
gonfunko: Updated accordingly.

PR #5500, opened 2021-09-18
feat: add `Blockly.Extensions.isRegistered` function

aaimio:
## The basics

- [x] I branched from develop

### Proposed Changes

Adds a `Blockly.Extensions.isRegistered` function.

I'm using the [block-plus-mins](https://github.com/google/blockly-samples/tree/master/plugins/block-plus-minus) plugin which attempts to [`Blockly.Extensions.unregister`](https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/if.js#L178) an extension called `controls_if_mutator`. That extension only exists if you import everything from Blockly (default blocks etc.). However, In my project I'm importing Blockly as:
 
```JS
import * as Blockly from "blockly/core";
```

which means the `controls_if_mutator` extension is not there.

Currently, any call to `Blockly.Extensions.unregister` will put a warning in the developer console.

Screenshot from 2021-09-18 21-57-07

#### Behavior Before Change

If attempting to unregister an extension that isn't registered, it will emit a warning to the developer's console.

#### Behavior After Change

This change will allow us to update the  [block-plus-mins](https://github.com/google/blockly-samples/tree/master/plugins/block-plus-minus) plugin to check whether an extension is registered before unregistering it, so we can prevent the `console.warn` from happening.

### Reason for Changes

Please see **Proposed changes**.

### Test Coverage

N/A


aaimio:
@googlebot I signed it!

moniika (changes requested):
Thanks for the PR!

This change looks good to me, it just needs to be updated with the latest changes from develop. This quarter we did a large refactor to migrate to closure modules and the syntax and style of declaring functions has changed slightly.

If it easier to create a new PR, I can also review that instead.

aaimio:
Thank you @moniika!

Issue #5436, opened 2021-09-07
Collapse context menu option does not group events


BeksOmega:
**Describe the bug**

When you select the "Collapse Blocks" context menu option, all of the blocks are properly collapsed. However, if you try to undo this, you have to undo collapsing each block individually, because the events are not grouped.

**To Reproduce**

1. Go to the playground.
2. Select the spawn random blocks action, or drag some blocks into the workspace.
3. Right click the workspace and select the "Collapse Blocks" option.
4. Attempt to undo the action using ctrl + z
5. Observe how you have to undo the actions individually.

**Additional context**

The relevant code of this context menu item lives [here](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/contextmenu_items.js#L104).

I think the easiest way to solve this is to remove the [timeout](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/contextmenu_items.js#L110), and add [setGroup](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/events/events.js#L388) calls surrounding the for-loop.

PR #5501, opened 2021-09-19
fix: group context menu collapse blocks events

aaimio:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/5436

### Proposed Changes

Groups collapse and expand events originating from the workspace context menu option.

#### Behavior Before Change

> When you select the "Collapse Blocks" context menu option, all of the blocks are properly collapsed. However, if you try to undo this, you have to undo collapsing each block individually, because the events are not grouped.

#### Behavior After Change

The events are grouped so an undo/redo will correctly revert or apply all collapse events.

### Reason for Changes

To address https://github.com/google/blockly/issues/5436

### Test Coverage

* Desktop Chrome
* Desktop Firefox

1. Launch a bunch of airstrikes
2. Collapse all blocks
3. Undo

### Additional Information

I've tried to maintain the [original delay](https://github.com/aaimio/blockly/blob/develop/core/contextmenu_items.js#L105-L106) in between collapsing blocks to ensure the original behaviour is maintained, however it adds a bit of bloat compared to just removing the timeouts completely as suggested by @BeksOmega in https://github.com/google/blockly/issues/5436:

> I think the easiest way to solve this is to remove the timeout, and add setGroup calls surrounding the for-loop.

Feel free to let me know which approach would be better.


alschmiedt:
Hello!

Thanks for the fix and welcome to the Blockly community! We are working on getting a new release out so we might be a bit slower than normal in getting back to you, but we are working on getting someone to review your PR. 

Thanks for your contribution!

core/contextmenu_items.js
```diff
@@ -104,10 +104,28 @@ Blockly.ContextMenuItems.registerCleanup = function() {
...
+  var createTimeoutFn = function(collapseFn, timerIdx) {
+    var timeoutFn = function() {
+      collapseFn();
+      timeouts.splice(timerIdx, 1, null);
+      for (var i = 0; i < timeouts.length; i++) {
+        if (timeouts[i] != null) {
+          return;
+        }
+      }
+      Blockly.Events.setGroup(false);
```
moniika, line 118: Rather than using the state of the `timeouts` array to track whether all the `setCollapsed` calls have completed, this could be simplified by using a counter.
The counter would be incremented before the` setTimeout` call and decremented within `createTimeoutFn`. When the counter is 0, then all the `setCollapsed` calls have been made. This would remove the need to also store an array and simplify some of the checks.

aaimio: That makes much more sense, thank you @moniika. I will add these changes and update the PR.

alschmiedt (approved):

alschmiedt:
Thanks for all your work on this @aaimio! 

If you are need help looking for new issues or would like some help getting started on something, please don't hesitate to reach out!

Issue #5153, opened 2021-07-21
Migrate to named exports


cpcallen:
### Background

The [JavaScript](https://google.github.io/styleguide/jsguide.html#es-module-exports) and [TypeScript](https://google.github.io/styleguide/tsguide.html#exports) style guides specifically prohibit default exports for ES/TS Modules:

> ```JS
> // Do not use default exports:
> export default class Foo { ... } // BAD!
> 
> // Use named exports:
> export class Foo { ... }
> 
> // Alternate style named exports:
> class Foo { ... }
> export {Foo};
> ```

For `goog.module` modules, the JavaScript style guide does not specifically prohibit default exports (`exports = MyClass;`), but it gives only examples of named exports.  (It does give examples of `goog.require`ing the default export, however.)

### Proposal

Migrate all modules which have a default export to use only named exports, so

```JS
const MyClass = function(...) {...};
exports = MyClass
```
becomes
```JS
const MyClass = function(...) {...};
exports.MyClass = MyClass;
```
and the corresponding import is (simultaneously) changed from
```JS
const MyClass = goog.require('MyClass');
```
to
```
const {MyClass} = goog.require('MyClass');
```

### Discussion

There's nothing wrong with default exports in the short term, but this is something we will want to fix eventually as we continue the migration from `goog.provide` -> `goog.module` -> ES modules -> TypeScript.  Issue #5073 records a number of cases where we will probably decide to fix this sooner rather than later.  This bug is to ensure that "later" arrives eventually.

moniika:
When the time comes, we can automate changing all this with the following commands (swapping out core for each directory we want to fix):
```
# Change all the default exports to named exports
grep -rlE "^exports = \w+" core | xargs sed -i -E "s/^exports = (\w+)/exports = \{\1\}/"
# Change all the default imports to named imports
grep -rlE "^const \w+ = goog\.require" core | xargs sed -i -E "s/^const (\w+) = goog\.require/const \{\1\} = goog\.require/"
```

cpcallen:
> When the time comes, we can automate changing all this with the following commands (swapping out core for each directory we want to fix):
This one is great:
> ```
> # Change all the default exports to named exports
> grep -rlE "^exports = \w+" core | xargs sed -i -E "s/^exports = (\w+)/exports = \{\1\}/"
> ```
But I believe this catch _all_ the cases where we do non-destructuring imports, even the ones where there wasn't a default export:
> ```
> # Change all the default imports to named imports
[...]

PR #5515, opened 2021-09-22
refactor: Rename `Blockly.Blocks`, migrate to named exports

cpcallen:
## The basics

- [X] I have run `npm test`.

### Resolves

Part of #5073, #5153.

### Proposed Changes

* Rename  the `Blockly.Blocks` module to `Blockly.blocks` and name its default export ~~`definitions`~~ `Blocks`.
* Remove the `goog.module.declareLegacyNamespace` call from that module.
* Have `blockly.js` reexport the default export as ~~both~~ `Blockly.Blocks` ~~(marked @deprecated, for backwards compatibility) and `Blockly.blocks`~~.
* Add an appropriate entry to `renamings.js`, along with documentation explaining the format of that file.

### Additional Information

This PR is presented as a series of commits which sort of show an evolution (or possibly devolution) of my thoughts about how to handle reexporting formerly-default exports.

~~I am a bit on the fence about how best to handle these:~~

* ~~Reexporting the `definitions` export from `Blockly.blocks` as `Blockly.Blocks` in `blockly.js` seems like obviously the right thing to do, as it will keep existing code working.  But I think this should be marked deprecated, because the name does not agree with our style guide.~~

* ~~But what alternative should we provide?~~

  * ~~We could do the straight-forward thing—`exports.blocks = goog.require('Blockly.blocks');`—that's basically where I was heading in e33b736.  Developers could update their code by renaming `Blockly.Blocks` => `Blockly.blocks.definitions`. 
 It ensures that the name `Blockly.blocks` will be the object returned by `goog.require('Blockly.blocks')`.  But it's only possible to do this because I am both naming the default export _and_ renaming the module.~~
  * ~~We could instead do `exports.blocks = goog.require('Blockly.blocks').definitions;`, which is what I end up doing in 6261cf2.  This is potentially confusing (because `Blockly.blocks` now refers to two different things (`Blockly.blocks === goog.require('Blockly.blocks').definitions`).~~
  * ~~We could not export any new name at all, and instead suggest the migration path of (for now) using a `goog.require` of `Blockly.blocks` (as we do in `core`) or (later) an ES module `import` of `blockly/core/blocks.js` once we have migrated to ES modules.~~

~~Thoughts?~~


rachel-fenichel:
There's another option here: ignore the style guide and leave the name as `Blocks`. Use it as a named export; re-export in `blockly.js`.

Style is good, but consistency usually is more important than style. The `Blockly.Blocks` namespace is used by every developer who uses Blockly. Letting them continue with their existing block definitions is more important than having a lowercase name for this namespace.

Using a named export internally and re-exporting is still a good idea because we're trying to get to no default exports. 

But changing the name here creates a lot of work and I don't think it's worth it.

I still think many of the other renames are worth doing, because the names are only used externally by people who are already mucking around in Blockly's code in some way--whether by defining custom fields and renderers, extending existing classes, or registering new stuff.

cpcallen:
When you put it like that it seems so straight forward that I am wondering why I spent so much time fretting over other possibilities.

cpcallen:
@rachel-fenichel: OK, I have fixed all the test failures with this PR and it is ready for a proper review.  I suggest ignoring the individual commits and just looking at the net changes in the files changed tab.

I have some more changes to clean up the node packaging (which no longer does what the original author intended it to, if indeed it ever did), but I will save them for a follow-up PR because this one has been enough trouble.

cpcallen:
@rachel-fenichel: OK, this will require one more merge commit to resolve conflicts but is otherwise ready for re-review.

rachel-fenichel (approved):

PR #5521, opened 2021-09-23
Migrate core/field_angle.js to to goog.module syntax

NeilFraser:
Move angle picker field to goog module and ES6.
Previous branch got corrupted and had to rebuild it.


cpcallen (approved):
Looks like you've run `clang-format` on this.  Is that correct?

PR #5549, opened 2021-09-27
fix: infinite loop when using defineProperties

### LINKED ISSUE #5538: Blockly Developer Tools Test URL not loading with "Uncaught Range error:Maximum call stack size exceeded"
**Describe the bug**

Blockly Developer Tools Test URL not loading with "Uncaught Range error:Maximum call stack size exceeded"

**To Reproduce**

1. Go to 'http://google.github.io/blockly/demos/blockfactory/index.html'
2. Check the console error

**Expected behavior**

Page should be loaded successfully

**Screenshots**
Screenshot 2021-09-27 10 25 13 AM

 - OS: [ChromeOS]
 - Browser [chrome]
 - Version [93.0.4577.85 (Official Build) (64-bit)]

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#5538

### Background
When in compiled mode, the return value for `getDiv` was getting set to `exports.DIV`. This was creating an infinite loop since the getter for exports.DIV calls `getDiv`. 

I thought we were adding `exports.DIV = DIV` because the compiler does not understand that defineProperties adds a property named `DIV` on to the exports object. However, when I remove this line the compiler does not complain, so I am wondering if it is easier to just remove this line all together? 

### Additional Information
- We were not being consistent with how we were defining the property. Some did `exports.X = X` and some did `exports.X;` I moved to the second way of doing it since that way does not cause infinite loops.


rachel-fenichel (approved):
Approved because I see that in blockly.js we just add properties but don't export them, and the same makes sense here.

The pitch field uses the widget div's DIV (in -samples), so that's an easy test.

PR #5624, opened 2021-10-20
Clang formatter

### LINKED ISSUE #5604: Enable clang formatting when PR is opened
After Q3 2021 all of our files have been formatted using [clang format](https://clang.llvm.org/docs/ClangFormat.html). We want our files to continue to be formatted correctly.

Ideally we would want a Github Action to format files correctly when a developer pushes changes to their PR. Developers would therefore not need to know about clang format when contributing. This is one less barrier for new developers wanted to contribute to Blockly. There is a [github action](https://github.com/marketplace/actions/clang-format-lint) that claims to do this, however I have not had a chance to look at it in depth. Part of this bug is doing some research on what is out there.

The other option is to make sure that the files submitted in a PR have been formatted using clang format, but not necessarily make the changes for the user. This is not ideal, since the developer has to run clang format on their end, but it is still better than nothing if we can't get the desired solution working. Here is a [potential github action](https://github.com/marketplace/actions/clang-format-check) that might be helpful for this.

**Additional context**

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#5604

### Proposed Changes
- Adds a GitHub action that will fail when files have not been clang formatted.
- Adds a GitHub action that will comment on the PR when the files have not been formatted correctly.
- Adds a script to run clang format on all files in core. `npm run format`.

### Reason for Changes
We just formatted all of our files and want to make sure they stay that way.

### Additional Information
One thing we might want to do is to limit the amount of times that this comments if clang format is broken. We can do this by using a third party github action like this [action](https://github.com/marketplace/actions/add-pr-comment) for adding a comment.


alschmiedt:
@deploy

PR #5626, opened 2021-10-21
chore: replace var with const and let in blocks directory

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/5160

### Proposed Changes

Replace uses of `var` with `const` and `let` as appropriate.

#### Behavior After Change

No change in behaviour.

### Reason for Changes

Get closer to google style guide.

### Test Coverage
Standard tests + opened all categories in the playground + played around with mutators on list blocks to test updateShape logic.

### DIFF
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -134,7 +134,7 @@ Blockly.Blocks['lists_create_with'] = {
    * @this {Blockly.Block}
    */
   mutationToDom: function() {
-    var container = Blockly.utils.xml.createElement('mutation');
+    const container = Blockly.utils.xml.createElement('mutation');
     container.setAttribute('items', this.itemCount_);
     return container;
   },
@@ -172,11 +172,11 @@ Blockly.Blocks['lists_create_with'] = {
    * @this {Blockly.Block}
    */
   decompose: function(workspace) {
-    var containerBlock = workspace.newBlock('lists_create_with_container');
+    const containerBlock = workspace.newBlock('lists_create_with_container');
     containerBlock.initSvg();
-    var connection = containerBlock.getInput('STACK').connection;
-    for (var i = 0; i < this.itemCount_; i++) {
-      var itemBlock = workspace.newBlock('lists_create_with_item');
+    let connection = containerBlock.getInput('STACK').connection;
+    for (let i = 0; i < this.itemCount_; i++) {
+      const itemBlock = workspace.newBlock('lists_create_with_item');
       itemBlock.initSvg();
       connection.connect(itemBlock.previousConnection);
       connection = itemBlock.nextConnection;
@@ -189,17 +189,17 @@ Blockly.Blocks['lists_create_with'] = {
    * @this {Blockly.Block}
    */
   compose: function(containerBlock) {
-    var itemBlock = containerBlock.getInputTargetBlock('STACK');
+    let itemBlock = containerBlock.getInputTargetBlock('STACK');
     // Count number of inputs.
-    var connections = [];
+    const connections = [];
     while (itemBlock && !itemBlock.isInsertionMarker()) {
       connections.push(itemBlock.valueConnection_);
       itemBlock = itemBlock.nextConnection &&
           itemBlock.nextConnection.targetBlock();
     }
     // Disconnect any children that don't belong.
-    for (var i = 0; i < this.itemCount_; i++) {
-      var connection = this.getInput('ADD' + i).connection.targetConnection;
+    for (let i = 0; i < this.itemCount_; i++) {
+      const connection = this.getInput('ADD' + i).connection.targetConnection;
       if (connection && connections.indexOf(connection) === -1) {
         connection.disconnect();
       }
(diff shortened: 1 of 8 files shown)


blocks/procedures.js
```diff
@@ -783,8 +784,8 @@ Blockly.Blocks['procedures_callnoreturn'] = {
    * @this {Blockly.Block}
    */
   updateShape_: function() {
-    for (var i = 0; i < this.arguments_.length; i++) {
-      var field = this.getField('ARGNAME' + i);
+    for (let i = 0; i < this.arguments_.length; i++) {
+      let field = this.getField('ARGNAME' + i);
```
cpcallen, line 788: I think this should be `const`, because the "mutation" of `field` below appears to just be variable reuse.

blocks/lists.js
```diff
@@ -217,10 +217,10 @@ Blockly.Blocks['lists_create_with'] = {
    * @this {Blockly.Block}
    */
   saveConnections: function(containerBlock) {
-    var itemBlock = containerBlock.getInputTargetBlock('STACK');
-    var i = 0;
+    let itemBlock = containerBlock.getInputTargetBlock('STACK');
+    let i = 0;
     while (itemBlock) {
```
cpcallen, line 222: There is potential here as well to replace this with a `for` loop which might be more readable.

blocks/logic.js
```diff
@@ -447,17 +449,17 @@ Blockly.Constants.Logic.CONTROLS_IF_MUTATOR_MIXIN = {
...
+    const valueConnections = [null];
+    const statementConnections = [null];
+    let elseStatementConnection = null;
 
     if (this.getInput('ELSE')) {
       elseStatementConnection = this.getInput('ELSE').connection.targetConnection;
     }
-    var i = 1;
+    let i = 1;
     while (this.getInput('IF' + i)) {
```
cpcallen, line 460: This again would be more readable as a `for` loop

blocks/procedures.js
```diff
@@ -798,19 +799,20 @@ Blockly.Blocks['procedures_callnoreturn'] = {
...
-        var input = this.appendValueInput('ARG' + i)
+        const input = this.appendValueInput('ARG' + i)
             .setAlign(Blockly.ALIGN_RIGHT)
             .appendField(field, 'ARGNAME' + i);
         input.init();
       }
     }
     // Remove deleted inputs.
+    let i = this.itemCount_;
     while (this.getInput('ARG' + i)) {
```
cpcallen, line 810: And again.

blocks/text.js
```diff
@@ -779,10 +779,10 @@ Blockly.Constants.Text.TEXT_JOIN_MUTATOR_MIXIN = {
    * @this {Blockly.Block}
    */
   saveConnections: function(containerBlock) {
-    var itemBlock = containerBlock.getInputTargetBlock('STACK');
-    var i = 0;
+    let itemBlock = containerBlock.getInputTargetBlock('STACK');
+    let i = 0;
     while (itemBlock) {
```
cpcallen, line 784: Possibly here too, thought the loop update condition is possibly a bit verbose unless you leave the `itemBlock` assignment in the body and only move the `i++` to the `for` line.

blocks/text.js
```diff
@@ -803,16 +803,17 @@ Blockly.Constants.Text.TEXT_JOIN_MUTATOR_MIXIN = {
...
+        const input = this.appendValueInput('ADD' + i)
             .setAlign(Blockly.ALIGN_RIGHT);
         if (i === 0) {
           input.appendField(Blockly.Msg['TEXT_JOIN_TITLE_CREATEWITH']);
         }
       }
     }
     // Remove deleted inputs.
+    let i = this.itemCount_;
     while (this.getInput('ADD' + i)) {
```
cpcallen, line 817: And again.

cpcallen (approved):
This is probably excessively picky, but where one or more `let` statements is/are followed by a loop, and the declared variable(s is/are used only in the loop, it is probably preferable to move the `let` declarations into the loop initialiser (changing `while` to `for` if required).

It reduces the visibility of the loop-related variables, and can help make it more clear what the loop is iterating over.  Of course, some degree of judgement is required: it probably isn't helping readability if it results in the loop header being so long it has to be spread over multiple lines.

blocks/text.js
rachel-fenichel: I changed this one around but agree that it looks weird.

blocks/text.js
rachel-fenichel: Done.

blocks/lists.js
rachel-fenichel: Done

blocks/lists.js
```diff
@@ -240,16 +240,17 @@ Blockly.Blocks['lists_create_with'] = {
...
-        var input = this.appendValueInput('ADD' + i)
+        const input = this.appendValueInput('ADD' + i)
             .setAlign(Blockly.ALIGN_RIGHT);
         if (i === 0) {
           input.appendField(Blockly.Msg['LISTS_CREATE_WITH_INPUT_WITH']);
         }
       }
     }
     // Remove deleted inputs.
+    let i = this.itemCount_;
```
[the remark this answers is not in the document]
rachel-fenichel: Done

blocks/text.js
rachel-fenichel: The `no-cond-assign` rule does not like this (for readability) so I'm going to put it back. I don't think that `for (let i = 0; itemBlock; i++)` does as good a job of making it obvious that the loop depends on `itemBlock` instead of just being a standard for loop.

Issue #5438, opened 2021-09-07
Undo/Redo with create/delete events does not trigger animations


BeksOmega:
**Describe the bug**

When you undo a create event, or redo a delete event, it does not trigger the delete animation.

**To Reproduce**
Create event:
1. Go to the playground.
2. Drag out a block.
3. Use ctrl + z to undo
4. Observe how there is no animation for the delete.

Delete event:
1. Go to the playground.
2. Drag out a block.
3. Delete the block
3. Use ctrl + z to undo
4. Use ctr + shift + z to redo
4. Observe how there is no animation for the delete.

**Expected behavior**

There should be fun animations :D

**Fix**
To fix this you should just need to add an extra `true` parameter [here](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/events/block_events.js#L295) and [here](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/events/block_events.js#L384) to trigger [this](https://github.com/google/blockly/blob/f94b1db85018b5a7261fa01f7d04cd5c8367fe2c/core/block_svg.js#L924).

BeksOmega:
Closed by #5919

BeksOmega:
Reopened due to #6001

PR #5640, opened 2021-10-26
Added true parameter to trigger animations

VedanthB:
## The basics

- [X] I branched from develop

### Resolves

I have implemented the fix suggested in the [issue] (https://github.com/google/blockly/issues/5438) 

### Proposed Changes

When you undo a create event, or redo a delete event, it does not trigger the delete animation. I have implemented the fix suggested in the [issue](https://github.com/google/blockly/issues/5438) 

### Test Coverage

I have reproduced the bug following the instructions in the issue and then implemented the fix suggested in the issue , I have not run any tests (I dont know how)

### Additional Information

Ive tried to make the changes to the best of my understanding of the issue , apologise in advance if the changes implemented are wrong.

### DIFF
--- a/core/events/events_block_create.js
+++ b/core/events/events_block_create.js
@@ -103,7 +103,7 @@ BlockCreate.prototype.run = function(forward) {
       const id = this.ids[i];
       const block = workspace.getBlockById(id);
       if (block) {
-        block.dispose(false);
+        block.dispose(true);
       } else if (id === this.blockId) {
         // Only complain about root-level block.
         console.warn('Can\'t uncreate non-existent block: ' + id);
--- a/core/events/events_block_delete.js
+++ b/core/events/events_block_delete.js
@@ -113,7 +113,7 @@ BlockDelete.prototype.run = function(forward) {
       const id = this.ids[i];
       const block = workspace.getBlockById(id);
       if (block) {
-        block.dispose(false);
+        block.dispose(true);
       } else if (id === this.blockId) {
         // Only complain about root-level block.
         console.warn('Can\'t delete non-existent block: ' + id);


core/events/events_block_create.js
```diff
@@ -103,7 +103,7 @@ BlockCreate.prototype.run = function(forward) {
       const id = this.ids[i];
       const block = workspace.getBlockById(id);
       if (block) {
-        block.dispose(false);
+        block.dispose(true);
```
clukito, line 106: This will actually heal the stack rather than triggering the animations. Another `true` argument should be added to trigger the animation.

core/events/events_block_delete.js
```diff
@@ -113,7 +113,7 @@ BlockDelete.prototype.run = function(forward) {
       const id = this.ids[i];
       const block = workspace.getBlockById(id);
       if (block) {
-        block.dispose(false);
+        block.dispose(true);
```
clukito, line 116: Same here.

jschanker:
@VedanthB To follow up on what @clukito wrote, you'll want to change the first argument back to `false` to preserve the behavior from before (not attempting to heal the stack) and then add a second argument of `true` to trigger the delete animation when you call `dispose` as per the JSDoc annotation here:
https://github.com/google/blockly/blob/f14a1c8034eb0e5e16ae89951841090110a4a532/core/block_svg.js#L880-L888

But to implement Neil's suggestion, you can add a `ui` parameter variable to the `run` method of the `BlockDelete` event here:
https://github.com/google/blockly/blob/e661c92a360e83c0bf1d2055b81355c0312cb08d/core/events/block_events.js#L378

and use it instead of passing `true` for the second argument you'll be adding here:
https://github.com/google/blockly/blob/e661c92a360e83c0bf1d2055b81355c0312cb08d/core/events/block_events.js#L384
This way, the deletion animation runs exactly when `run` is called with a second argument of `true` (for your newly added `ui` parameter).  Since this will be an optional argument that defaults to `true`, you will also want to set it to `true` when it's not provided (e.g., analogous to the code excerpt below):
https://github.com/google/blockly/blob/d919b0af21f57c33f5e0d0761477ec033fc83e97/core/options.js#L70-L72

By doing this, you allow the application to decide whether to show the deletion animation.  So, for example, for the Mirror Blockly Demo, we'd want the deletion animation to show in the primary workspace for which it occurred on the left but not for the mirrored one on the right.  To accomplish this, we could call `run` on the `secondaryEvent` by replacing the below line:

https://github.com/google/blockly/blob/3fe56b083c62549cb2bdb17270846bf17ced5d7a/demos/mirror/index.html#L76

with `secondaryEvent.run(true, false);`  (Here, `secondaryEvent` is essentially a copy of the most recent event that took place in the main workspace on the left for the mirrored workspace on the right.)

Remember that you'll also want to update the JSDoc annotation to document this change to the method's signature:

https://github.com/google/blockly/blob/e661c92a360e83c0bf1d2055b81355c0312cb08d/core/events/block_events.js#L376-L378

by inserting a line on 377 analogous to the one here https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/core/block_svg.js#L900 

for the addition of the optional `ui` parameter variable.

@NeilFraser please correct me if I misrepresented what you were suggesting.

VedanthB:
@jschanker @clukito thanks for explaining it so well, ill make the changes and submit a PR 😄

Issue #5004, opened 2021-07-08
Blockly.Events.Abstract is missing type property


moniika:
The `type` property is not documented by the `Blockly.Events.Abstract` class which all Events inherit from.
This is evident from the generated [documentation](https://developers.google.com/blockly/reference/js/Blockly.Events.Abstract?hl=en) and leads to some confusion for developers.

The `type` property should be declared on `Blockly.Events.Abstract` or there should be an event interface that `Blockly.Events.Abstract` implements that declares the property.

maribethb:
This has been fixed

PR #5641, opened 2021-10-26
Documenting the Type Property

Devesh21700Kumar:
## The basics

- [x] I branched from develop

### Resolves

 #5004 

### Proposed Changes

Declared *type* property should be declared on `Blockly.Events.Abstract`

### DIFF
--- a/core/events/events_abstract.js
+++ b/core/events/events_abstract.js
@@ -53,6 +53,12 @@ const Abstract = function() {
    * @type {boolean}
    */
   this.recordUndo = eventUtils.getRecordUndo();
+
+  /**
+   * Specifies the type of the event
+   * @type {string|undefined}
+   */
+  this.type = undefined;
 };
 
 /**


alschmiedt:
Thanks for the PR! 

This property needs to be set on the prototype like below: 
```
/**
 * Type of this event.
 * @type {string|undefined}
 */
Abstract.prototype.type = undefined;
```

The reason the tests are failing is because the events [set the type on the prototype](https://github.com/google/blockly/blob/bc7329f34a61f389c31e4b3992e84d1d5ce43ded/core/events/events_block_change.js#L58) instead of in the constructor. When we try to set `type = undefined;` in the constructor of the Abstract Event it is clobbering the value on the prototype of the subclass. This was making the type of all events undefined.

Devesh21700Kumar (commented):
Thanks for helping me in finding why the errors occurred @alschmiedt .
Will try to fix this asap.

BeksOmega:
Hello! I'm going to go ahead and close this due to it being stale. If you ever want to contribute in the future definitely feel welcome to! We always appreciate help =)

PR #5648, opened 2021-10-28
chore: Add feedback on expiry page.

NeilFraser:
Blank pages are bad for confirming anything happened.

Can be temporarily tested at:
https://fraserexpire-dot-blockly-demo.appspot.com/expiration
This shows that the production version is actually working fine, so there's no bug (just a lot of data).

For reference, our current DB size is 1.88 GB, and every execution kills 1000 records.
I've increased the hit from once a week to every two minutes, and will let that run until the above URL stops saying 1000.

### DIFF
--- a/appengine/expiration.py
+++ b/appengine/expiration.py
@@ -39,13 +39,14 @@ def delete_expired():
     results = query.fetch(limit=QUERY_LIMIT, keys_only=True)
     for x in results:
       x.delete()
+  return len(results)
 
 
 def app(environ, start_response):
-  out = ""
   headers = [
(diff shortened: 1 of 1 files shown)


NeilFraser:
Follow-up, the size shrank from 1.88GB to 462.31 MB.  That's a _lot_ of data.

cpcallen:
> the size shrank from 1.88GB to 462.31 MB. That's a _lot_ of data.

You call that a lot of data?  [I just want to serve 5 terabytes.](https://youtu.be/3t6L-FlfeaI)

PR #5653, opened 2021-10-28
chore: update dependabot messages and labels

rachel-fenichel:
The same as #5618, but against master because dependabot ignores changes on develop.
See https://github.com/google/blockly/pull/5618


cpcallen:
> The same as #5618, but against master because dependabot ignores changes on develop.

Ooh, oops!

PR #5656, opened 2021-11-01
Remove usage of potentialVariableMap

### LINKED ISSUE #5273: Cleanup handling of potential variables.

Potential variables are variables that we don't want to exist on the main workspace. For example, if you want blocks in your flyout to reference variables, but you don't want those variables to be accessible in the main workspace until the block is dragged into the main workspace.

The problem is, we shouldn't need a separate [potentialVariableMap_](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/workspace.js#L122) which is only available on workspaces that are flyouts. We should be able to use the normal [variableMap_](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/workspace.js#L110) of the workspace that is a flyout.

Use the normal variableMap_ of the workspace that is the flyout, instead of creating a separate property that is only used by flyouts. This would require modifying the [flyout](https://github.com/google/blockly/blob/master/core/flyout_base.js), [workspace](https://github.com/google/blockly/blob/master/core/workspace.js), and [variables namespace](https://github.com/google/blockly/blob/master/core/variables.js).

**Additional context**

The [potential variable map is package](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/workspace.js#L735), so this shouldn't be a breaking change.

JerryHue:
## The basics

- [X] I branched from develop

### Resolves

Fixes #5273.

### Proposed Changes

To avoid triggering a refresh, we will not create a variable like main workspaces do. Instead, we create a variable by accessing the variable map directly.

#### Behavior Before Change

Before, when creating a new variable, the flyout workspace would use potentialVariableMap to keep track of such variables that are not being used.

#### Behavior After Change

Now, flyout workspace is using its own variableMap, removing the unnecessary references to potentialVariableMap.

### Reason for Changes

Simplify the definition of what a workspace is (potentialVariableMap was a property used in the special case of a flyout workspace).

### Test Coverage

I tested on the following browsers:

* Desktop Firefox

The way I tested is by creating and deleting variables and making sure that it behaved the same way as before the changes.

### Documentation

I believe some sparse documentation needs to be updated, since the potentialVariableMap was heavily involved on the usage of flyout workspaces. I tried to reword some comments that I could find.

### Additional Information

No additional information has to be added.

### DIFF
--- a/core/variable_map.js
+++ b/core/variable_map.js
@@ -358,9 +358,6 @@ VariableMap.prototype.getVariablesOfType = function(type) {
 VariableMap.prototype.getVariableTypes = function(ws) {
   const variableMap = {};
   object.mixin(variableMap, this.variableMap_);
-  if (ws && ws.getPotentialVariableMap()) {
-    object.mixin(variableMap, ws.getPotentialVariableMap().variableMap_);
-  }
   const types = Object.keys(variableMap);
   let hasEmpty = false;
   for (let i = 0; i < types.length; i++) {
--- a/core/variables.js
+++ b/core/variables.js
@@ -472,15 +472,11 @@ exports.getOrCreateVariablePackage = getOrCreateVariablePackage;
  * @alias Blockly.Variables.getVariable
  */
 const getVariable = function(workspace, id, opt_name, opt_type) {
-  const potentialVariableMap = workspace.getPotentialVariableMap();
   let variable = null;
   // Try to just get the variable, by ID if possible.
   if (id) {
-    // Look in the real variable map before checking the potential variable map.
+    // Look in the variable map
     variable = workspace.getVariableById(id);
-    if (!variable && potentialVariableMap) {
-      variable = potentialVariableMap.getVariableById(id);
-    }
     if (variable) {
       return variable;
     }
@@ -493,9 +489,6 @@ const getVariable = function(workspace, id, opt_name, opt_type) {
     }
     // Otherwise look up by name and type.
     variable = workspace.getVariable(opt_name, opt_type);
-    if (!variable && potentialVariableMap) {
-      variable = potentialVariableMap.getVariable(opt_name, opt_type);
-    }
   }
   return variable;
 };
@@ -512,18 +505,17 @@ exports.getVariable = getVariable;
  *     or name + type combination.
  */
 const createVariable = function(workspace, id, opt_name, opt_type) {
-  const potentialVariableMap = workspace.getPotentialVariableMap();
   // Variables without names get uniquely named for this workspace.
   if (!opt_name) {
     const ws = workspace.isFlyout ? workspace.targetWorkspace : workspace;
     opt_name = exports.generateUniqueName(ws);
   }
 
-  // Create a potential variable if in the flyout.
+  // Create a variable directly in the variable map if in the flyout.
   let variable = null;
-  if (potentialVariableMap) {
-    variable = potentialVariableMap.createVariable(opt_name, opt_type, id);
-  } else {  // In the main workspace, create a real variable.
+  if (workspace.isFlyout) {
+    variable = workspace.getVariableMap().createVariable(opt_name, opt_type, id);
+  } else {  // In the main workspace, create a variable to trigger a refresh.
     variable = workspace.createVariable(opt_name, opt_type, id);
   }
   return variable;
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -322,11 +322,6 @@ Flyout.prototype.init = function(targetWorkspace) {
   this.workspace_.getGesture =
       this.targetWorkspace.getGesture.bind(this.targetWorkspace);
 
-  // Get variables from the main workspace rather than the target workspace.
-  this.workspace_.setVariableMap(this.targetWorkspace.getVariableMap());
-
-  this.workspace_.createPotentialVariableMap();
-
   targetWorkspace.getComponentManager().addComponent({
     component: this,
     weight: 1,
@@ -800,9 +795,6 @@ Flyout.prototype.clearOldBlocks_ = function() {
     button.dispose();
   }
   this.buttons_.length = 0;
-
-  // Clear potential variables from the previous showing.
-  this.workspace_.getPotentialVariableMap().clear();
 };
 
 /**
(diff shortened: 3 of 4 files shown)


JerryHue:
@googlebot I signed it!

core/variables.js
```diff
@@ -472,15 +472,11 @@ exports.getOrCreateVariablePackage = getOrCreateVariablePackage;
  * @alias Blockly.Variables.getVariable
  */
 const getVariable = function(workspace, id, opt_name, opt_type) {
-  const potentialVariableMap = workspace.getPotentialVariableMap();
   let variable = null;
   // Try to just get the variable, by ID if possible.
   if (id) {
-    // Look in the real variable map before checking the potential variable map.
+    // Look in the variable map
```
maribethb, line 478: The main idea behind this refactor is that for any workspace, we have two properties: variableMap and potentialVariableMap. "Real" workspaces (not flyouts) use the variableMap but not the potentialVariableMap and flyout workspaces use the potentialVariableMap but not the variableMap (mostly, see below). We could simplify this logic by just using the variableMap for both use cases. We still need the concept of "potential" variables that exist in the flyout but not the main workspace, but we can store those in the same property to simplify things.

The flyout was using the variableMap property to store the variableMap for the target workspace. However, you can just access that via `workspace.targetWorkspace.getVariableMap()` which frees you up to use the `variableMap` property for the "potential" variables.

For this part, you still need to do what the comment you deleted says: look in the real variable map before checking for potential variables. In our new code, that means if we're in a flyout, we need to check the target workspace's variable map first (real variables) and then check the flyout's variable map (potential variables) if needed. If we're not in a flyout, just check the variable map (real variables).

So you can do something like

```js
if (workspace.isFlyout) {
  // Check for real variables in the target workspace first.
  variable = workspace.targetWorkspace.getVariableById(id);
}
if (!variable) {
  // Either this is a flyout and the main workspace didn't contain the variable, so check for potential variables, or this wasn't a flyout workspace so just check for real ones.
  variable = workspace.getVariableById(id);
}
```

So looking at this logic, I'm not sure that we should follow through with this refactor. It simplifies things in one respect, because we are eliminating the `potentialVariableMap` property which is unused for main workspaces. However, now we are using one property to mean two different things depending on whether it's a flyout workspace or not, which is a different kind of complexity that comes through on the comment for that second `if` statement.

core/variables.js
```diff
@@ -493,9 +489,6 @@ const getVariable = function(workspace, id, opt_name, opt_type) {
     }
     // Otherwise look up by name and type.
     variable = workspace.getVariable(opt_name, opt_type);
```
maribethb, line 491: We would need to do the same kind of logic here to check the target workspace first.

core/variables.js
```diff
@@ -512,18 +505,17 @@ exports.getVariable = getVariable;
...
 const createVariable = function(workspace, id, opt_name, opt_type) {
-  const potentialVariableMap = workspace.getPotentialVariableMap();
   // Variables without names get uniquely named for this workspace.
   if (!opt_name) {
     const ws = workspace.isFlyout ? workspace.targetWorkspace : workspace;
     opt_name = exports.generateUniqueName(ws);
   }
 
-  // Create a potential variable if in the flyout.
+  // Create a variable directly in the variable map if in the flyout.
```
maribethb, line 514: This section is where the refactor would simplify things. `workspace.getVariableMap().createVariable` is equivalent to `workspace.createVariable`. No matter if we are in a flyout or a main workspace, after this refactor, we would just create it in the variable map so we could eliminate the if statements here and just do

```js
variable = workspace.createVariable(...)
```

However again, looking at this simplification vs the additional complexity introduced when looking for a variable, I'm not sure it's worth doing.

maribethb (commented):
Hello, thank you for the PR! I have some comments that hopefully help clarify what would be needed here, but after looking at how the logic changes, I'm not sure that we should continue forward with this refactor. I'm going to post in the bug about it in case there is further discussion. Thank you, and sorry both that it took us a while to respond to your PR and that this refactor may not be needed after all.

JerryHue:
Hello, @maribethb! I understand the points that you have made, and I agree with them.

As a side note, I chose this specific issue to work on since it seems less complex and it helped me dive into the codebase and understand things a little bit better. I appreciate your response. Thank you.

core/variables.js
cpcallen, line 478: I think this vestigial comment can be deleted.

core/variable_map.js
```diff
@@ -358,9 +358,6 @@ VariableMap.prototype.getVariablesOfType = function(type) {
 VariableMap.prototype.getVariableTypes = function(ws) {
   const variableMap = {};
   object.mixin(variableMap, this.variableMap_);
```
cpcallen, line 360: While you're modifying this function, please change it to [use `Object.assign` instead of `object.mixin`](#5914).

core/variable_map.js
```diff
@@ -358,9 +358,6 @@ VariableMap.prototype.getVariablesOfType = function(type) {
...
-  if (ws && ws.getPotentialVariableMap()) {
-    object.mixin(variableMap, ws.getPotentialVariableMap().variableMap_);
-  }
```
cpcallen, line 363: This eslint warning is a sign that in fact this code does need to be retained, but modified appropriately:
```JS
  if (ws) {
    Object.assign(variableMap, ws.getVariableMap().variableMap_);
  }
```

core/variables.js
cpcallen: (Unless we move the code to `WorkspaceSvg.prototype.getVariableById` as I propose above.)

core/variables.js
cpcallen: I concur about the change required: the entire second section of the function can be replace by:
```JS
  return workspace.createVariable(opt_name, opt_type, id);
```
(which is in part why I think this change _is_ worth doing).

cpcallen (changes requested):
I've looked at this PR in detail and, with some considerable assistance from @NeilFraser, I now believe I properly understand [the original bug](#5656), this PR, and the relevant parts of the underlying codebase.  Big props to you, @JerryHue for picking up this issue—even if it was maybe a little more thorny than you expected!

I think we should proceed with the proposed changes, though there are changes required as noted by @maribethb.  See inline for my detailed response to her comment.

BeksOmega:
I'm going to go ahead and close this as stale. Feel free to reopen if you're still interested in working on this! Or always feel free to reach out about other places you can contribute if you are interested =)

PR #5660, opened 2021-11-01
chore: Normalize ++x to x++.

NeilFraser:
There are only 10 instances of ++x in our codebase (not counting places where the return value is used), compared with over 500 instances of x++.  The style guide has no opinion on which to use, nor do I.  But the lack of consistency is making regex searches for bugs more difficult.


cpcallen (approved):
Slightly disappointed that there were no cases where pre/post increment made a difference.  I miss the days of `while(*p++ = *q++);`

Issue #5598, opened 2021-10-13
Modularize generators


rachel-fenichel:
Issue for tracking conversion of contents of `generators/` to `goog.module`.

Like blocks definitions, block generator files operate primarily by side effects, so this is related to #5189.

rachel-fenichel:
Done!

PR #5661, opened 2021-11-01
chore: update javascript block generators to const and let

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Updates javascript block generators to `const` and `let` in preparation for modularization. (Part of https://github.com/google/blockly/issues/5598)

### Proposed Changes

Replace `var` with `const` and `let`. Where needed, restructure code that was relying on hoisting and multiple declarations of a variable.

#### Behavior Before Change
No change.

### Test Coverage

Tested by generator tests.


generators/javascript/lists.js
```diff
@@ -82,61 +82,63 @@ Blockly.JavaScript['lists_indexOf'] = function(block) {
...
-        var code = list + '.pop()';
+        const code = list + '.pop()';
         return [code, Blockly.JavaScript.ORDER_MEMBER];
       } else if (mode === 'REMOVE') {
         return list + '.pop();\n';
       }
       break;
-    case ('FROM_START'):
-      var at = Blockly.JavaScript.getAdjusted(block, 'AT');
+    case ('FROM_START'): {
```
NeilFraser, line 114: Below on line 277 you solve the same problem with `let at1;` instead of creating extra scopes.  Is there a reason not to be consistent?

rachel-fenichel: Yes. If the variable will be used outside of that scope, I create it before the switch (with `let`). If it will only be used inside the case, I add brackets to create a new scope.

NeilFraser (approved):

PR #5675, opened 2021-11-03
chore: revert github action

alschmiedt:
### Proposed Changes
Reverts #5670 which was added for testing purposes.


cpcallen:
Yeah: we can take another pass at this later when we've worked out a better solution.

cpcallen (approved):

PR #5679, opened 2021-11-05
chore: add command to launch mocha tests

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves
Improves workflow for local testing

### Proposed Changes

Adds the command `npm run tests:mocha:interactive`, which launches a server and opens the mocha index page.

### Reason for Changes

Reduce friction in testing--when I see test failures in my mocha tests on the command line, my next move is usually to open the mocha test page so I can start debugging.

### Test Coverage

Ran command locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "start": "http-server ./ -o /tests/playground.html -c-1",
     "test": "tests/run_all_tests.sh",
     "test:generators": "tests/scripts/run_generators.sh",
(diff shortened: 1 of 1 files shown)


cpcallen (approved):
This is fine, but it would be even better if it would work even if one had already run `npm start`—either by running the second http server on a different port, or (better) by detecting the existing one and just opening the URL in one's browser.

PR #5680, opened 2021-11-05
fix: isPrime block should cast str as num

### LINKED ISSUE #5678: Change from == to === in code generators
**Describe the bug**

In the conversion from `==` to `===` in PR https://github.com/google/blockly/pull/5599, some strings of generated code were changed as well:
E.g., for checking if a number is prime:
https://github.com/google/blockly/blob/49c99377540a7bfd5ef612cb38663d754dedb195/generators/javascript/math.js#L161

Or for checking if a number is even: https://github.com/google/blockly/blob/49c99377540a7bfd5ef612cb38663d754dedb195/generators/javascript/math.js#L183

**Expected behavior**

I'm not sure if this change is intended or not because while using `===` is the new preferred style, it breaks backward compatibility.

**Additional context**

My students were in the process of correcting a bug in operator precedence and had based their changes on the original repository with `==`.  Once it changed to `===` some generator tests failed (https://github.com/RichDoherty/blockly/pull/10/checks#step:8:97).

NeilFraser:
Fixes #5678

### DIFF
--- a/tests/generators/golden/generated.js
+++ b/tests/generators/golden/generated.js
@@ -427,7 +427,7 @@ function test_constant() {
 
 function mathIsPrime(n) {
   // https://en.wikipedia.org/wiki/Primality_test#Naive_methods
-  if (n === 2 || n === 3) {
+  if (n == 2 || n == 3) {
     return true;
   }
   // False if n is NaN, negative, is 1, or not whole.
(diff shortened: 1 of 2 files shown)


tests/generators/golden/generated.js
```diff
@@ -427,7 +427,7 @@ function test_constant() {
 
 function mathIsPrime(n) {
   // https://en.wikipedia.org/wiki/Primality_test#Naive_methods
-  if (n === 2 || n === 3) {
```
rachel-fenichel, line 430: When did the generator test change in this direction?

NeilFraser: Here: https://github.com/google/blockly/pull/5599/files#diff-7910823a2f8f803e4f1e2aa5e7976728aa490ffe4c015d162e7f2e0533348a1b

This is the one place where functionality inadvertently changed.  But to trigger this change this one has to stuff a string into an untyped variable then pass the variable to the "is prime" math block:
https://blockly-demo.appspot.com/static/demos/code/index.html#384y9p

All the other changes in generated code are verified as being executionally NOPs.

rachel-fenichel (approved):

Issue #4353, opened 2020-10-06
Some orders in generators incorrect


moniika:
**Describe the bug**

In the generators for library blocks, some of the `ORDER` enums used are incorrect. 

In some cases, the mistake in order only causes extra parenthesis (only a problem for when looking at code for examples):
- if using an order for an operator stronger than the actual one for max strength (passed to valueToCode)
- if returning an order for an operator weaker than the actual one for min strength (for generator return)
- Certain accidental swaps of orders don't cause issues because of how close they are in order (commonly mixed up `ORDER_MEMBER` and `ORDER_FUNCTION_CALL`)

![image](https://user-images.githubusercontent.com/6621618/95247867-47d79b80-07cb-11eb-8d0d-d8489f9f4a33.png)

In other cases, the mistake in order causes syntax error in specific edge cases. This can be very tricky to catch because there are a lot of valid combinations.

![image](https://user-images.githubusercontent.com/6621618/95248094-a6047e80-07cb-11eb-9808-19f1e4072e75.png)

**Expected behavior**

Library blocks should use the correct order:
- There should be no syntax errors in code generated from Blockly blocks. 
- In the case where the mistakes don't cause syntax errors, mistakes in order are detrimental to others' understanding of generators and make the code unsuitable as example code. 

**Additional context**

Historically, there has been a lot of confusion with understanding the correct operator precedence to use when writing generator code. Part of what contributes to this confusion is that the value of the `ORDER` enums is inversely related to the strength (higher order and stronger precedence is a lower value).

moniika:
Manually reviewed colour generator code and found no errors in ORDER.

PR #5685, opened 2021-11-06
fix(generators): Fixed an operator precedence issue in the math_number_property generators to remove extra parentheses

RichDoherty:
## The basics

- [x] I branched from develop

### Partially Resolves

https://github.com/google/blockly/issues/4353 

### Proposed Changes

Fixes an issue in the math_number_property generator. Changed the order of the outerOrder argument to valueToCode to reflect that there are "is prime," "is positive," and "is negative" options. Corrected the return value order for "is positive" and "is negative" in JavaScript, PHP, and Dart.

#### Behavior Before Change

Extra parentheses were added when arithmetic blocks were placed in "is prime," "is positive," and "is negative" math_number_property blocks (excluding exponentiation) because ORDER_MODULUS/ORDER_MULTIPLICATIVE was being used instead of ORDER_NONE/ORDER_RELATIONAL. Extra parentheses were also added if you put "is positive" or "is negative" into a logic compare block because ORDER_EQUALITY was being used instead of ORDER_RELATIONAL. 

#### Behavior After Change

No extra parentheses are added.

### Reason for Changes

To remove unneeded parentheses in generated code.

### Test Coverage

Tested locally.

Tested on: 
* Desktop Chrome
* Desktop Firefox

### Documentation

None necessary.


RichDoherty:
@googlebot I signed it!

jeremyjacob123:
@googlebot I signed it!

jschanker:
@NeilFraser RE: the discussion from #5678, would we want to change the generated code for PHP (and update the tests) so that it uses `===` instead of `==` when code execution would remain unchanged as well?

generators/lua/math.js
```diff
@@ -127,14 +127,47 @@ Blockly.Lua['math_constant'] = function(block) {
...
+    PI: ['math.pi', Blockly.Lua.ORDER_HIGH],
+    E: ['math.exp(1)', Blockly.Lua.ORDER_HIGH],
+    GOLDEN_RATIO: ['(1 + math.sqrt(5)) / 2', Blockly.Lua.ORDER_MULTIPLICATIVE],
+    SQRT2: ['math.sqrt(2)', Blockly.Lua.ORDER_HIGH],
+    SQRT1_2: ['math.sqrt(1 / 2)', Blockly.Lua.ORDER_HIGH],
+    INFINITY: ['math.huge', Blockly.Lua.ORDER_HIGH]
+  };
+  return CONSTANTS[block.getFieldValue('CONSTANT')];
+};
+
```
jschanker, line 142: This code from the previous version should have been removed.

generators/dart/math.js
```diff
@@ -172,18 +172,35 @@ Blockly.Dart['math_constant'] = function(block) {
...
+  const numberToCheck = Blockly.Dart.valueToCode(block, 'NUMBER_TO_CHECK',
+      inputOrder) || '0';
   let code;
-  if (dropdown_property === 'PRIME') {
+  if (dropdownProperty === 'PRIME') {
     // Prime is a special case as it is not a one-liner test.
     Blockly.Dart.definitions_['import_dart_math'] =
         'import \'dart:math\' as Math;';
-    const functionName = Blockly.Dart.provideFunction_(
+        const functionName = Blockly.Dart.provideFunction_(
```
jschanker, line 203: This should only be indented four spaces.

LouisCatala:
@googlebot I fixed it

LouisCatala:
@googlebot I signed it!

cpcallen (changes requested):
Hi @RichDoherty,

This looks great.  There's just one small improvement I have to suggest (to be applied in several places); see below.

The main thing we'd like to see before merging this pull request added generator tests for the new code paths.  I suggest you add them on a separate branch (branched from `develop`), make sure they fail there, then cherry-pick them onto this branch (and make sure they pass here, obviously).

generators/lua/math.js
```diff
@@ -156,41 +176,20 @@ Blockly.Lua['math_number_property'] = function(block) {
...
-      const divisor = Blockly.Lua.valueToCode(block, 'DIVISOR',
-          Blockly.Lua.ORDER_MULTIPLICATIVE);
-      // If 'divisor' is some code that evals to 0, Lua will produce a nan.
-      // Let's produce nil if we can determine this at compile-time.
-      if (!divisor || divisor === '0') {
-        return ['nil', Blockly.Lua.ORDER_ATOMIC];
-      }
-      // The normal trick to implement ?: with and/or doesn't work here:
-      //   divisor == 0 and nil or number_to_check % divisor == 0
-      // because nil is false, so allow a runtime failure. :-(
```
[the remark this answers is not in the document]
jeremyjacob123: I'm a little confused by this comment. number_to_check % 0 doesn't cause a division by 0 runtime failure in Lua. This doesn't seem as if it is an issue with Blockly itself?

RichDoherty:
@cpcallen 

>This looks great. There's just one small improvement I have to suggest (to be applied in several places); see below.

Thank you for taking the time to review our pull request. 

>The main thing we'd like to see before merging this pull request added generator tests for the new code paths.

We added some block tests to account for extra parentheses and a test for the remainder when dividing by 0, but we weren't sure what to do with the Lua comment.
[https://github.com/google/blockly/pull/5685#discussion_r753639058](https://github.com/google/blockly/pull/5685#discussion_r753639058)

generators/lua/math.js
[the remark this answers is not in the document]
cpcallen: I don't know much Lua but a cursory check of running `print(1%0)` using [the online Lua demo](https://www.lua.org/cgi-bin/demo) gives a runtime error of `input:1: attempt to perform 'n%0'`, so AFAICT the originally-documented issue still exists, and so the documentation about it should be preserved.

tests/generators/math.xml
```diff
@@ -1204,20 +1331,20 @@
...
     </statement>
   </block>
-  <block type="procedures_defnoreturn" x="13" y="4038">
+  <block type="procedures_defnoreturn" x="46" y="4064">
     <field name="NAME">test change</field>
     <comment pinned="false" h="80" w="160">Tests the "change" block.</comment>
     <statement name="STACK">
       <block type="variables_set" inline="false">
-        <field name="VAR">varToChange</field>
+        <field name="VAR" id="vC|N=?5}X^+}FPTXGW`X">varToChange</field>
```
cpcallen, line 1339: Please remove unneeded variable `id`s.

tests/generators/math.xml
```diff
@@ -1887,9 +2014,9 @@
       </block>
     </statement>
   </block>
-  <block type="procedures_defnoreturn" y="5938" x="13">
+  <block type="procedures_defnoreturn" x="13" y="5938">
     <field name="NAME">test atan2</field>
-    <comment w="160" h="80" pinned="false">Describe this function...</comment>
+    <comment pinned="false" h="80" w="160">Describe this function...</comment>
```
cpcallen, line 2019: It would be preferable also to revert meaningless changes like these.

cpcallen (changes requested):
Thanks for adding tests.  They look good (but see some minor nit-picking about the XML input, below).

I see there are conflicts.  This is doubtless because we touched a lot of code in the generators last week, converting them from `goog.provide` to `goog.module`, so you'll need to merge `develop` into this branch (or rebase it on top of the current tip of `develop`).  Sorry to make more work for you!

generators/lua/math.js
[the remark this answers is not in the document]
RichDoherty: We added the comment back in, but when running `print(1%0)` on [https://repl.it/languages/lua](https://repl.it/languages/lua) the output is `-nan`. Although not official documentation, [https://rosettacode.org/wiki/Detect_division_by_zero#Lua](https://rosettacode.org/wiki/Detect_division_by_zero#Lua) seems to suggest that this is expected?

RichDoherty:
Made the suggested changes, but still not sure about the discrepancies for Lua.

cpcallen:
> Made the suggested changes, but still not sure about the discrepancies for Lua.

Please restore the comment that was on lines 186–188.

It looks like the Lua generator produces code that is very slightly buggy.  We should fix that, but it doesn't need to be fixed in this PR.  We do want to keep the comment pointing out the problem, though!

jeremyjacob123:
> Please restore the comment that was on lines 186–188.
>
> It looks like the Lua generator produces code that is very slightly buggy. We should fix that, but it doesn't need to be fixed in this PR. We do want to keep the comment pointing out the problem, though!

The comment from lines 186-188 was placed [here.](https://github.com/google/blockly/pull/5685/files#diff-024099ec49a554c64ca542911c66dbafcb76624414577ac1498e33a8f95e76b2R181-R183) Thanks!

cpcallen:
> > Please restore the comment that was on lines 186–188.
> > It looks like the Lua generator produces code that is very slightly buggy. We should fix that, but it doesn't need to be fixed in this PR. We do want to keep the comment pointing out the problem, though!
> 
> The comment from lines 186-188 was placed [here.](https://github.com/google/blockly/pull/5685/files#diff-024099ec49a554c64ca542911c66dbafcb76624414577ac1498e33a8f95e76b2R181-R183) Thanks!

Oh… somehow I overlooked that.  Mea culpa.  I apologise for wasting your time asking you to do something that you had already done!

PR #5692, opened 2021-11-08
fix: Don't try to set text fields to null on cancel (for master)

NeilFraser:
Mobile users get a window.prompt as an input, if they press the cancel button the return value is null.  Don't attempt to set the value of the field to null.

Caused errors in the custom note field which inherits from FieldTextInput.  Detected in Blockly Games Music.

This PR is for the master branch and includes a recompile.  The develop branch has changed enough that a cherrypick from develop to master won't work.  The bug in question represents a significant number of the errors being reported from Blockly Games.


rachel-fenichel:
Makes sense. Do you know where we introduced the change in behaviour?

rachel-fenichel (approved):

NeilFraser:
I don't believe the cancel button was ever respected.  It just isn't a code path we'd noticed.  Partly because we don't do a lot of testing on mobile.

PR #5695, opened 2021-11-08
fix: Don't throw if drag surface is empty.

NeilFraser:
Observed this error in Blockly Games Bird 6 (Chrome OS):
`Uncaught TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.`
~It has only shown up once, and I can't reproduce,~ but it looks like the drag surface was being cleared while empty.  Maybe some weird multi-touch race-condition on a Chromebook?

Open question: Leave the code as is so that the error is thrown and visible, or suppress it?

### DIFF
--- a/core/block_drag_surface.js
+++ b/core/block_drag_surface.js
@@ -252,11 +252,14 @@ BlockDragSurfaceSvg.prototype.getWsTranslation = function() {
  *     being moved to a different surface.
  */
 BlockDragSurfaceSvg.prototype.clearAndHide = function(opt_newSurface) {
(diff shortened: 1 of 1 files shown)


NeilFraser:
Update: This bug shows up a lot (now that I'm looking for it).  Chrome and Edge.  Maze and Bird (including simple levels without mutators).

rachel-fenichel (approved):

rachel-fenichel:
@NeilFraser please merge this if ready.

PR #5699, opened 2021-11-08
chore: move remaining function definitions out of blockly.js

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5026 

### Proposed Changes

Move several functions from `blockly.js` to `common.js`, but leave aliases behind.
- `svgSize`
- `resizeSvgContents`
- `hideChaff`
- `defineBlocksWithJsonArray`
- `isNumber`

These functions are all used from outside of Blockly but not from inside.

### Reason for Changes

@cpcallen requested these changes:

> The idea is that almost all modules are "pure" `goog.modules` that just have a bunch of `exports`.  All the construction of the Blockly namespace object (or rather, tree of namespace objects) is done in one place, and that file will NOT be used by anyone doing advanced compilation.
> (By "one place" I mean `blockly.js` and possibly helper modules like `requires.js`.)

### Documentation

@gonfunko possibly some JSDoc alias updates--please check.

### Additional Information

I considered adding items to the renamings script:
```
'Blockly': {
      exports: {
        svgSize: { module: 'Blockly.common' },
        resizeSvgContents: { module: 'Blockly.common' },
        hideChaff: { module: 'Blockly.common'},
        defineBlocksWithJsonArray: { module: 'Blockly.common' },
        isNumber: { module: 'Blockly.utils.math' },
      },
    },
```

but I left them out because I still want these names to be used through `Blockly` and not through `Blockly.common`.


rachel-fenichel:
After discussion with Chris I added renamings.

rachel-fenichel:
After discussion with Aaron I added or fixed aliases for functions that are now in common.

rachel-fenichel:
@cpcallen this is ready for review

core/common.js
```diff
@@ -152,3 +155,93 @@ exports.svgResize = svgResize;
...
+exports.resizeSvgContents = resizeSvgContents;
+
+/**
+ * Returns the dimensions of the specified SVG image.
+ * @param {!SVGElement} svg SVG image.
+ * @return {!Size} Contains width and height properties.
+ * @deprecated Use workspace.getCachedParentSvgSize. (2021 March 5)
+ * @alias Blockly.common.svgSize
+ */
+const svgSize = function(svg) {
```
cpcallen, line 177: Might this better live somewhere in `Blockly.utils`?

core/common.js
```diff
@@ -152,3 +155,93 @@ exports.svgResize = svgResize;
...
+ * @param {boolean=} opt_onlyClosePopups Whether only popups should be closed.
+ * @alias Blockly.common.hideChaff
+ */
+const hideChaff = function(opt_onlyClosePopups) {
+  // The deprecated name is `Blockly.hideChaff` because this function used to
+  // be declared in Blockly.js
+  deprecation.warn('Blockly.hideChaff', 'September 2021', 'September 2022');
+  getMainWorkspace().hideChaff(opt_onlyClosePopups);
+};
+exports.hideChaff = hideChaff;
```
cpcallen, line 202: I'd be OK leaving this in `blockly.js` as anyone who was using but wants to avoid importing all of Blockly can just do what it says and use `getMainWorkspace().hideChaff()`.

(Oh, and regardless of where it is, shouldn't the `deprecation.warn` include instructions to that effect?)

core/common.js
```diff
@@ -152,3 +155,93 @@ exports.svgResize = svgResize;
...
+  };
+};
+
+/**
+ * Define blocks from an array of JSON block definitions, as might be generated
+ * by the Blockly Developer Tools.
+ * @param {!Array<!Object>} jsonArray An array of JSON block definitions.
+ * @alias Blockly.common.defineBlocksWithJsonArray
+ */
+const defineBlocksWithJsonArray = function(jsonArray) {
```
cpcallen, line 223: Should this go in the `Blockly.Block` module?

cpcallen (approved):
Fine, I guess, but I did wonder if there might be better places to put a few of these things; see below.

core/common.js
cpcallen: (No, it shouldn't, since this is about block definitions not blocks.  Maybe `Blockly.Blocks`, though, that being where those definitions live.)

rachel-fenichel (commented):
Updated.

cpcallen (approved):
L(even better)TM.

PR #5706, opened 2021-11-11
chore: Move functions from utils

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5026, part of #5428

### Proposed Changes

Moves several functions from `utils.js` into new files: `utils/parsing.js` and `utils/svgMath.js`. Also adds renamings entries and updates all calls.

#### Behavior Before Change

No change

### Reason for Changes

@cpcallen requested these changes:

The idea is that almost all modules are "pure" goog.modules that just have a bunch of exports. All the construction of the Blockly namespace object (or rather, tree of namespace objects) is done in one place, and that file will NOT be used by anyone doing advanced compilation.
(By "one place" I mean blockly.js and possibly helper modules like requires.js.)

### Additional Information

Replaces two PRs from Beka: https://github.com/google/blockly/pull/5470 and https://github.com/google/blockly/pull/5484

Open questions:
- Should I include wrappers in `utils.js` that warn about deprecation, or just alias? Adding wrappers gives more info at runtime, so I want them if they won't cause problems.
- Should any of these be changed to be completely internal?
- Where do the remaining three functions go? (`getBlockTypeCounts`, `arrayRemove`, `runAfterPageLoad`)


gonfunko:
LGTM, I'd also be in favor of wrappers but defer to Chris on whether that causes any issues. I think we've added them in other cases where we've moved functions though.

gonfunko (approved):

PR #5709, opened 2021-11-12
chore: fix or ignore remaining lint

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/5160

### Proposed Changes

Fixes or ignore the remaining lint that blocks us from using `eslint-config-google`. After this, it's safe to add it as a dev dependency and turn it on in our eslintconfig.

- Now ignoring `prefer-rest-params` in mocha tests because of a pre-existing pattern.
- Now ignoring `guard-for-in` everywhere because of a pre-existing pattern.
- Fixed most violations of `no-invalid-this`.
  - Added `eslint-ignore-next-line` directives for two violations that I couldn't sort out.
  - Added TODOs for those violations, but I'll need to also file bugs and update the todos accordingly.
  - If you see how to just fix them, let me know and I'll update it in-place.
- Fixed violations of `prefer-spread`, mostly involving array concatenation.
- Ignored some missing jsdoc requests on constructors--adding jsdoc caused compiler failures because it can't take constructor annotations not at the top level.

### Reason for Changes

Get to google style for our lint config, I hope.

### Additional Information

The two issues I couldn't fix are in `blocks/procedures.js` and `core/contextmenu.js`.

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -537,11 +537,11 @@ Blockly.Blocks['procedures_mutatorarg'] = {
     // Hack: override showEditor to do just a little bit more work.
     // We don't have a good place to hook into the start of a text edit.
     field.oldShowEditorFn_ = field.showEditor_;
-    // TODO: Understand what this is in newShowEditorFn, and annotate properly.
+    /**
+     * @this {Blockly.FieldTextInput}
+     */
     const newShowEditorFn = function() {
-      /* eslint-disable-next-line no-invalid-this */
       this.createdVariables_ = [];
-      /* eslint-disable-next-line no-invalid-this */
       this.oldShowEditorFn_();
     };
     field.showEditor_ = newShowEditorFn;
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -537,8 +537,11 @@ Blockly.Blocks['procedures_mutatorarg'] = {
     // Hack: override showEditor to do just a little bit more work.
     // We don't have a good place to hook into the start of a text edit.
     field.oldShowEditorFn_ = field.showEditor_;
+    // TODO: Understand what this is in newShowEditorFn, and annotate properly.
     const newShowEditorFn = function() {
+      /* eslint-disable-next-line no-invalid-this */
       this.createdVariables_ = [];
+      /* eslint-disable-next-line no-invalid-this */
       this.oldShowEditorFn_();
     };
     field.showEditor_ = newShowEditorFn;
--- a/core/block.js
+++ b/core/block.js
@@ -817,7 +817,7 @@ Block.prototype.getDescendants = function(ordered) {
   let blocks = [this];
   const childBlocks = this.getChildren(ordered);
   for (let child, i = 0; (child = childBlocks[i]); i++) {
-    blocks = [...blocks, ...child.getDescendants(ordered)];
+    blocks.push(...child.getDescendants(ordered));
   }
   return blocks;
 };
--- a/core/block.js
+++ b/core/block.js
@@ -814,10 +814,10 @@ Block.prototype.setParent = function(newParent) {
  * @return {!Array<!Block>} Flattened array of blocks.
  */
 Block.prototype.getDescendants = function(ordered) {
-  let blocks = [this];
+  const blocks = [this];
   const childBlocks = this.getChildren(ordered);
   for (let child, i = 0; (child = childBlocks[i]); i++) {
-    blocks.push(...child.getDescendants(ordered));
+    blocks.push.apply(blocks, child.getDescendants(ordered));
   }
   return blocks;
 };
--- a/core/block.js
+++ b/core/block.js
@@ -814,10 +814,10 @@ Block.prototype.setParent = function(newParent) {
  * @return {!Array<!Block>} Flattened array of blocks.
  */
 Block.prototype.getDescendants = function(ordered) {
-  const blocks = [this];
+  let blocks = [this];
   const childBlocks = this.getChildren(ordered);
   for (let child, i = 0; (child = childBlocks[i]); i++) {
-    blocks.push.apply(blocks, child.getDescendants(ordered));
+    blocks = [...blocks, ...child.getDescendants(ordered)];
   }
   return blocks;
 };
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -60,7 +60,10 @@
         // Keep them for legacy reasons.
         "new-cap": ["off"],
         // Mostly use default rules for brace style, but allow single-line blocks.
-        "brace-style": ["error", "1tbs", { "allowSingleLine": true }]
+        "brace-style": ["error", "1tbs", { "allowSingleLine": true }],
+        // Blockly uses objects as maps, but uses Object.create(null) to
+        // instantiate them.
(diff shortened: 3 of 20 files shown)


crypt096 (approved):

blocks/procedures.js
```diff
@@ -537,8 +537,11 @@ Blockly.Blocks['procedures_mutatorarg'] = {
...
     // We don't have a good place to hook into the start of a text edit.
     field.oldShowEditorFn_ = field.showEditor_;
+    // TODO: Understand what this is in newShowEditorFn, and annotate properly.
     const newShowEditorFn = function() {
+      /* eslint-disable-next-line no-invalid-this */
       this.createdVariables_ = [];
+      /* eslint-disable-next-line no-invalid-this */
       this.oldShowEditorFn_();
     };
     field.showEditor_ = newShowEditorFn;
```
BeksOmega, line 547: My understanding is that because the function is invoked as a method, `this` should be bound to whatever object the function exists on ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#as_an_object_method)). So I think it should be `FieldTextInput` or whatever the particular derivative is. However, I don't think closure is going to like that because `oldShowEditorFn_` doesn't exist on `FieldTextInput`, only this kluged version of `FieldTextInput`.

Might be better to leave as is for now, and fix the hack by adding in a hook for the start of a text edit.

core/block.js
```diff
@@ -814,10 +814,10 @@ Block.prototype.setParent = function(newParent) {
  * @return {!Array<!Block>} Flattened array of blocks.
  */
 Block.prototype.getDescendants = function(ordered) {
-  const blocks = [this];
+  let blocks = [this];
   const childBlocks = this.getChildren(ordered);
   for (let child, i = 0; (child = childBlocks[i]); i++) {
-    blocks.push.apply(blocks, child.getDescendants(ordered));
+    blocks = [...blocks, ...child.getDescendants(ordered)];
```
NeilFraser, line 820: Can you check the performance delta of this?  Especially after being cross-compiled.  Usually (but not always) ES6 stuff is slower, and this looks like a critical area.  There's another similar change later in this PR too.

cpcallen: I suspect that this version will be slower simply because (unless the JIT is very clever) it creates a new object and copies all the entries over instead of appending to the existing one.

An alternative that might be both more readable and performant than the original is `blocks.push(blocks, ...child.getDescendants(ordered))`.

rachel-fenichel: Good call. 

TLDR: `blocks.push(...child.getDescendants(ordered)` is the move here.

Note that there's no need to use `push.apply` given that both objects are already arrays.

### Details

I benchmarked two ways: using [neil's blog](https://neil.fraser.name/news/2021/06/07/) and using [JSbench.me](https://jsbench.me/).

For each benchmark, I had the following setup code:
```
let array1 = [];
const array2 = [];
for (let i = 0; i < 1000; i++) {
  array1.push(i);
  array2.push(i + 2000);
}
```

#### Test cases

Then I tested three cases:
Spread only:
```
for (let j = 0; j < 1000; j++) {
  array1 = [...array1, ...array2];
}
```

Push + spread:
```
for (let j = 0; j < 1000; j++) {
  array1.push(...array2);
}
```

Concat:
```
for (let j = 0; j < 1000; j++) {
  array1 = array1.concat(array2);
}
```

Finally, I tried this with no looping in the actual test cases--just concatenating once.

#### Results

With both benchmarks, and for both looping and not, `[...arr1, ...arr2]` is at least an order of magnitude slower. `arr1.push(...arr2)` is faster than `arr1.concat(arr2)`, although the two sites give wildly different numbers for concat.

### Does it matter?
For this case: maybe. For the `renderList` case below: yes, absolutely, because that's a very hot path.

blocks/procedures.js
rachel-fenichel: I fixed the annotation based on that description.

core/block.js
rachel-fenichel: Fixed.

NeilFraser: Also relevant, Babel transpiles this code:

`  array1.push(...array2);`

into this code:

`  array1.push.apply(array1, array2);`

Which brings us back full circle to the original code.

However, that's only if the compiler can detect through static analysis that array2 is indeed an array.  If it can't figure this out, then it generates the following:

```
function _toConsumableArray(arr) {
  return (
    _arrayWithoutHoles(arr) ||
    _iterableToArray(arr) ||
    _unsupportedIterableToArray(arr) ||
    _nonIterableSpread()
  );
}

function _nonIterableSpread() {
  throw new TypeError(
    "Invalid attempt to spread non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method."
  );
}

function _unsupportedIterableToArray(o, minLen) {
  if (!o) return;
  if (typeof o === "string") return _arrayLikeToArray(o, minLen);
  var n = Object.prototype.toString.call(o).slice(8, -1);
  if (n === "Object" && o.constructor) n = o.constructor.name;
  if (n === "Map" || n === "Set") return Array.from(o);
  if (n === "Arguments" || /^(?:Ui|I)nt(?:8|16|32)(?:Clamped)?Array$/.test(n))
    return _arrayLikeToArray(o, minLen);
}

function _iterableToArray(iter) {
  if (
    (typeof Symbol !== "undefined" && iter[Symbol.iterator] != null) ||
    iter["@@iterator"] != null
  )
    return Array.from(iter);
}

function _arrayWithoutHoles(arr) {
  if (Array.isArray(arr)) return _arrayLikeToArray(arr);
}

function _arrayLikeToArray(arr, len) {
  if (len == null || len > arr.length) len = arr.length;
  for (var i = 0, arr2 = new Array(len); i < len; i++) {
    arr2[i] = arr[i];
  }
  return arr2;
}

array1.push.apply(array1, _toConsumableArray(array2));
```

NeilFraser: I've run `npm build` on this code, and each of the six instances of `...` get compiled to a variation of:

`b.push.apply(b,$jscomp.arrayFromIterable(d.getDescendants(a)))`

Same as Babel when it can't be certain that the argument is an array.

So, if this code is intended to be cross-compiled to ES5, then could be a problem since these look like functions which are performance-sensitive.

rachel-fenichel: Okay, I reverted it entirely and disabled that rule in `.eslintrc.json`

cpcallen: Seems like we ought to be able to improve the typing info and thereby allow Closure Compiler to better optimise this—but it would probably require advanced optimisation.

cpcallen: Update: unlike Babel, [`$jscomp.arrayFromIterable` fast-paths the case where its argument is an array](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/js/es6/util/arrayfromiterable.js), so there's a good chance that the performance of the compiled code (even absent advanced optimisations) will end up being basically the same as the `Array.prototype.push.apply` implementation once the JS engine has worked its magic—though it will vary depending on the competence of the JIT, of course.  No objection to continuing to use the straight-forward implementation here, but maybe a reminder to try to benchmark the actual code that we ship…

NeilFraser (approved):

PR #5716, opened 2021-11-17
chore: migrate generators to goog.module

rachel-fenichel:
## The basics

- [x] I branched from develop

### Additional Information

For Javascript generators I did one commit per file. After that I switched to one commit per language.
Mocha and generator tests pass, but node tests fail--debugging that is the next step.


cpcallen:
OK, this Closure Compiler once again kicking us because it doesn't like us renaming things in `blockly.js`, combined with trying to avoid setting up proper chunked compilation.

The error is:

```
TypeError: Blockly.Generator.Generator is not a constructor
    at /Users/cpcallen/src/blockly/dist/javascript_compressed.js:22:205
```
which is:
```JS
var module$contents$Blockly$JavaScript_JavaScript=new Blockly.Generator.Generator("JavaScript");
```
which is what the compiler outputs in response to the following lines excerpted from `generators/javascript.js`:
```JS
const {Generator} = goog.require('Blockly.Generator');

// ...

/**
 * JavaScript code generator.
 * @type {!Generator}
 */
const JavaScript = new Generator('JavaScript');

// ...

exports = JavaScript;
```

The immediate question is: why is it trying to instantiate `Blockly.Generator.Generator` instead of `module$exports$Blockly$Generator.Generator`?

At first I thought the issue might be the `require('Blockly');` line at the top of `javascript.js`—which should definitely go!—but removing it did not solve the problem.  It seems instead that it is basically the presence of `externs/generator_externs.js` in the compiler input, containing:

```JS
goog.provide('Blockly');
goog.provide('Blockly.Generator');
goog.provide('Blockly.inputTypes');
goog.provide('Blockly.utils.global');
goog.provide('Blockly.utils.object');
goog.provide('Blockly.utils.string');
goog.provide('Blockly.Names');
goog.provide('Blockly.Variables');
goog.provide('Blockly.Workspace');
goog.provide('Blockly.Block');
goog.provide('Blockly.Constants.Loops');

var Blockly;
```

Except for the first and last lines this has become a lie: the contents of the `Blockly.Generator` module is not available as `Blockly.Generator`; instead, the `Generator` export from the `Blockly.Generator` module is available as `Blockly.Generator`.  Great for backwards compatibility for our users, not so helpful here.

Removing the `goog.provide('Blockly.Generator');` line from the externs just ends up with the build failing with:

```
generators/javascript.js:15:20: ERROR - [JSC_MISSING_MODULE_OR_PROVIDE] Required namespace "Blockly.Generator" never defined.
  15| const {Generator} = goog.require('Blockly.Generator');
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 error(s), 0 warning(s)
```

Unless someone can find a way to tell the compiler that "module `Foo.Bar` has been compiled and is available via a `module$exports$Foo$Bar`" I think we are going to have to move to chunked compilation after all.

PR #5717, opened 2021-11-17
chore: enable eslint-config-google

### LINKED ISSUE #5160: Use new eslint config for es6 / goog.module

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Closes #5160 

### Proposed Changes

- Add `eslint-config-google` as a dev dependency.
- Extend `"google"` in our `.eslintrc.json`
- Fix a few remaining comma dangles.

#### Behavior Before Change

No change in behaviour.

#### Behavior After Change

No change in behaviour.

### Reason for Changes

New lint!
a) conforming to Google style
b) using the same style in both blockly and blockly-samples.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -76,6 +76,6 @@
         "exports": true
     },
     "extends": [
-        "eslint:recommended"
+        "eslint:recommended", "google"
     ]
 }
--- a/core/common.js
+++ b/core/common.js
@@ -195,7 +195,7 @@ exports.getBlockTypeCounts = getBlockTypeCounts;
  *     of jsonDef.
  */
 const jsonInitFactory = function(jsonDef) {
-  return function() {
+  return /** @this {Block} */ function() {
     this.jsonInit(jsonDef);
   };
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,220 +47,232 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.5.5",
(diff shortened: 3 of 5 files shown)


.eslintrc.json
```diff
@@ -76,6 +76,6 @@
         "exports": true
     },
     "extends": [
-        "eslint:recommended"
+        "eslint:recommended", "google"
```
alschmiedt, line 79: I'm just curious, it looks like in blockly samples we use "eslint-config-google" instead of "google". Is there a difference?

alschmiedt (approved):

.eslintrc.json
rachel-fenichel: That's a little suspicious, because this is how the readme says to do it.

Issue #5597, opened 2021-10-13
Modularize blocks


rachel-fenichel:
Issue for tracking conversion of blocks directory to `goog.module`.

Most of the files in `blocks/` don't need to export anything, because they install blocks on the blocks object--that is, the side effects are the important thing (related to #5189).

Some files put properties on `Blockly.Constants.SomeNamespace`. These are generally not use outside of the file. They are on the constants namespace to keep them out of the global namespace. They can just become local and not exported, and it's possible that some of the leaf namespace on `Blockly.Constants` will also not be needed.

PR #5721, opened 2021-11-19
chore(build): Use chunked compilation

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Apparently needed in order to resolve issues #5597 and #5598 (and to land PR #5716).

### Proposed Changes

* Add modules `Blockly.blocks.all` and `Blockly.<Generator>.all`.
* Use `closure-calculate-chunks` to enable Closure Compiler to compile `core/`, `blocks/` and `generators/` in a single pass while still producing the same output files.

### Reason for Changes

It is not possible for files in `blocks/` and `generators/` to access the `exports` object of `goog.modules` that are not compiled in the same compilation pass; the previous trick using `exports/generator_exports.js` only works if using names `provide`ed by `blockly.js`, which we wish to avoid doing because doing causes difficulty for advanced compilation due to confusion because of the renamings done in `blockly.js`.

### Test Coverage

None yet!

### Additional Information

This is a draft PR because the resulting `build/*_compressed.js` files don't (yet) have UMD wrappers.

To do (or at least consider):

* [X] Use gulp pipeline to build, instead of direct invocation of compiler only.
* [x] UMD wrappers.
  * It turns out that [`--chunk_wrapper` can provide a separate wrapper for each chunk](https://github.com/google/closure-compiler/wiki/Flags-and-Options#code-splitting).  Win!
* [X] ~~Support for `--closure-library` flag?~~  No, not needed.
* [x] Remove `closure/goog/base.js` from list of files to compile?
* [X] Verify modules in later chunks can get access to exports objects of modules in earlier chunks.
  * Aside: having the compiler output ES modules would be an obvious solution to this problem, but there [was a bug in Closure Compiler v20211006](https://github.com/google/closure-compiler/issues/3883) that broke `--chunk_output_type ES_MODULES`.  It's fixed in v20211107 but we've not updated yet.  We probably don't want to go this route anyway because ES modules containing ES5 code aren't actually ES5 compatible!
  * We _can_ use `--rename_prefix_namespace` to ensure that "globals" like `module$exports$Blockly$…` from one chunk can be accessed in later ones by adding a configurable prefix, like so: `NS.module$exports$Blockly$…`.  This requires using a wrapper to provide said namespace object.  [There's an example which puts the namespace object on `window`](https://github.com/google/closure-compiler/wiki/Chunk-output-for-dynamic-loading#using-global_namespace-as-the-chunk-output-type),  but an alternative would be to attach it to the `Blockly` object itself.


cpcallen:
OK, I'm going to split the originally intended scope of this PR into (at least) ~two~ three separate pieces:

1. This PR replaces the old build pipeline (except for `buildAdvancedCompilationTest`) with the new `buildCompiled` pipeline that uses chunked compilation.  This is an _approximately_ neutral change, with the following observed differences when diffing `blockly_compressed.js` (et al.):
   * The `blockly_compressed.js` goes from 925182 bytes to 956429 bytes.
   * This is probably at least in part due to the compiler being fed `closure/goog/base.js`, which it wasn't previously.
   * The copyright header from `base.js` is appearing in the output.
   * The compiler seems to be including a few extra polyfills (e.g. `Object.setPrototypeOf`) in `blockly_compressed.js`
   * On the other hand, it is no longer including polyfills in `blocks_compressed.js` etc!

0. The next PR will enable `--rename_prefix_namespace`, and make the needed adjustments in the wrappers to provide a namespace object.  After that we should be able to land outstanding `goog.module` migrations for blocks and generators.

0. One or more follow-up PR will remediate the issues mentioned in item 1 (where possible) and update `buildAdvancedCompilationTest` so the that the remaining vestiges of the old compile pipeline can be removed.

cpcallen:
Investigating test failures:

* Failure under node v12 is bizzare; it seems to begin with failing to find the build-in `fs` modue:
  ```
   [16:47:43] Starting 'buildCompiled'...
    internal/process/esm_loader.js:74
        internalBinding('errors').triggerUncaughtException(
                                  ^
    
    Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'fs' imported from /home/runner/work/blockly/blockly/node_modules/closure-calculate-chunks/lib/chunk-graph.js
        at packageResolve (internal/modules/esm/resolve.js:664:9)
        at moduleResolve (internal/modules/esm/resolve.js:705:18)
        at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:798:11)
        at Loader.resolve (internal/modules/esm/loader.js:100:40)
        at Loader.getModuleJob (internal/modules/esm/loader.js:246:28)
        at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:47:40)
        at link (internal/modules/esm/module_job.js:46:36) {
      code: 'ERR_MODULE_NOT_FOUND'
    }
  ```
* Failure on node v16 is due to the following line from `base.js`:
  ```JS
  goog.global =
      // Check `this` first for backwards compatibility.
      // Valid unless running as an ES module or in a function wrapper called
      //   without setting `this` properly.
      // Note that base.js can't usefully be imported as an ES module, but it may
      // be compiled into bundles that are loadable as ES modules.
      this ||
      // https://developer.mozilla.org/en-US/docs/Web/API/Window/self
      // For in-page browser environments and workers.
      self;
  ```
  which fails in strict mode because `this` will be undefined in the wrapper.

scripts/gulpfiles/build_tasks.js
```diff
@@ -418,6 +298,196 @@ function buildLangfiles(done) {
...
+ * To make @package behave less strictly in core/, this function can
+ * be used to as a gulp.rename filter, modifying the path object to
+ * flatten all files in core/** so that they're in the same directory,
+ * while ensuring that files with the same base name don't conflict.
+ *
+ * @param {{dirname: string, basename: string, extname: string}}
+ *     pathObject The path argument supplied by gulp.rename to its
+ *     callback.  Modified in place.
+ */
+function flattenCorePaths(pathObject) {
```
rachel-fenichel, line 436: Some files are multiple directories down (e.g. renderers). Does this run multiple times?

scripts/gulpfiles/build_tasks.js
```diff
@@ -418,6 +294,217 @@ function buildLangfiles(done) {
...
+    hide_warnings_for: 'node_modules',
+    externs: ['./externs/goog-externs.js', './externs/svg-externs.js'],
+    define: 'Blockly.VERSION="' + packageJson.version + '"',
+    chunk: chunkOptions.chunk,
+    chunk_wrapper: chunkOptions.chunk_wrapper,
+    // Don't supply the list of source files in chunkOptions.js as an
+    // option to Closure Compiler; instead feed them as input via gulp.src.
+  };
+  if (argv.debug || argv.strict) {
+    // Tempororary hack: only compile first chunk, because there are
```
rachel-fenichel, line 480: >.<

scripts/gulpfiles/build_tasks.js
rachel-fenichel: After reading the rest: no, it doesn't run multiple times. I think it needs to loop until it runs out of slashes.

cpcallen: I am pretty sure the code is correct.  After the call to `.split(path.sep)`, there will be no slashes in the `dirs` array.  `basename` is then reassembled using `.join`.

To check, I added an error to `core/renderers/zelos/measurables/inputs.js`, and the resulting error message was for `core/renderers-slash-zelos-slash-measurables-slash-inputs.js`.

(Note that we do not want to get rid of the slash after `core`, because we only want to expand the scope of `@package` within `core`, and not between `core` and any other directory.)

cpcallen:
In the interests of expediency, I'm going to fold the second and third, etc. PRs mentioned in my comment above.  Everything above this comment is the work described in item 1 on that list.  @rachel-fenichel has already reviewed most of it, but not commits eaeee39 or c11c67e (of which only the former is actually interesting).  The next commit will be item 2.

cpcallen:
The three commits since my previous comment cover the work described in item 2 on the list in my earlier comment.

rachel-fenichel (approved):

PR #5730, opened 2021-11-24
chore: moves the public loop mixin into its own file

alschmiedt:
## The basics

- [x] I branched from develop

### Proposed Changes
Moves the "CONTROL_FLOW_IN_LOOP_CHECK_MIXIN" mixin into its own file and re exports at the top level.

### Reason for Changes
This mixin is the only public mixin that we expect external developers might need to use. This is because in our comments we say that developers should modify the LOOP_TYPES property on it.

### Additional Information

There's a [related issue](https://github.com/mit-cml/appinventor-sources/issues/2032) on app inventor tracking how they use this mixin--they don't like the previous design either.


core/loop_mixin.js
```diff
@@ -0,0 +1,83 @@
...
+ * is contained in a loop. Otherwise a warning is added to the block.
+ * @mixin
+ * @augments Blockly.Block
+ * @readonly
+ */
+const CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
+    /**
+     * List of block types that are loops and thus do not need warnings.
+     * To add a new loop type add this to your code:
+     * Blockly.LoopMixin.CONTROL_FLOW_IN_LOOP_CHECK_MIXIN.LOOP_TYPES.push('custom_loop');
```
rachel-fenichel, line 29: Update comment to refer to the new location, and add to renamings file as well.

rachel-fenichel: And add a comment telling them to talk to us if they want to use this :)

alschmiedt: Ok added a comment asking them to comment on [this issue](https://github.com/mit-cml/appinventor-sources/issues/2032) if they need to use this. And I updated `Blockly.LoopMixin` to be `Blockly.loopMixin`.

rachel-fenichel (approved):

core/loop_mixin.js
```diff
@@ -0,0 +1,96 @@
...
+   * Please leave a comment on this github issue
+   * https://github.com/mit-cml/appinventor-sources/issues/2032 if you end up
+   * needing to use this.
+   */
+  LOOP_TYPES: [
+    'controls_repeat',
+    'controls_repeat_ext',
+    'controls_forEach',
+    'controls_for',
+    'controls_whileUntil',
```
cpcallen, line 46: I admit that I probably did not fully grok our recent discussion about this change—it was getting rather late here—but looking at this resulting PR I am completely perplexed about why this code has been moved to `core/`, since, for example, these block types are totally meaningless without the definitions given in `blocks/`.

Since modules in `generators/` require modules in `blocks/`, and since it would be easy to enable external code to `import {CONTROL_FLOW_IN_LOOP_CHECK_MIXIN} from 'blockly/blocks';` I don't really know why it wouldn't be better to leave this code in `blocks/loops.js` (but move it to the `Blockly.blocks.loops` module, and reexport from `Blockly.loops.all`)

rachel-fenichel: The idea that some blocks can contain break or continue, and others can't, seemed like a core idea that had gotten pushed out into the blocks. Although perhaps the controls block definition file should push each of these names onto the list as it defines the block, rather than having it hardcoded from the start.

PR #5731, opened 2021-11-24
chore(deps): Remove dependency on typescript-closure-tools

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Removes a dependency on typescript-closure-tools.
Comments out typings generation during test, since we are using a rube goldberg machine for typings this quarter--and will be on typescript after that.

#### Behavior Before Change

Build typings as a test, erroring only if typings completely failed.

#### Behavior After Change

No typings test.

### Reason for Changes

This tool can't handle `goog.module` so it hasn't been giving us useful information--it's just an extra dependency.

For Q4, typings will come from a series of hacks that make our code look like es6 modules. After that we will use the typescript compiler.

### Documentation

- [ ] File a bug to track re-enabling typings generation as a test.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,6 @@
         "selenium-standalone": "^8.0.3",
         "through2": "^4.0.2",
         "typescript": "^4.3.2",
-        "typescript-closure-tools": "^0.0.7",
         "webdriverio": "^7.0.3",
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
Closing in favor of https://github.com/google/blockly/pull/5776

PR #5741, opened 2021-11-29
chore: move some properties to constants.js

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Moves two properties, `VARIABLE_CATEGORY_NAME` and `PROCEDURE_CATEGORY_NAME`, from `internal_constants.js` to `constants.js`.

#### Behavior Before Change

Names were only exported at the top level.

#### Behavior After Change

Names are exported in constants, which means that generators can use them by importing constants.

### Reason for Changes

Needed to make generator migration (#5716) work, because it needs to use those two properties but not import all of Blockly.

### Test Coverage
Standard tests

### Documentation
No change

### Additional Information

I updated the aliases for jsdoc.
I left `VARIABLE_DYNAMIC_CATEGORY_NAME` alone because I don't have an immediate need to access it.


rachel-fenichel:
Closing in favor of #5743

Issue #5158, opened 2021-07-21
Break up and remove `core/internal_constants.js`


cpcallen:
### Background

#5140 is an important precursor to finishing the first pass of migrating `goog.provides` to `goog.module` by removing side effects from `core/constants.js`, which previously modified the `Blockly` object by adding a large number of assorted `@constant` properties to it.

In order not to advertise any new API surface, these constants have been redeclared in `core/internal_constants.js` (rather than `core/contstants.js`), which is marked `@package` to prevent their use outside of `core`.

### Proposal

* All constants should be moved out of this file and to the file(s) containing the code to which they relate.
* Where they are only used in a single module, they should become unexported, module-local constants in that file.
* Where they are intended to be user-configurable parameters, they should be renamed so as not to be `CONSTANT_CASE` and, to comply with [style guide rules forbidding mutable exports](https://google.github.io/styleguide/jsguide.html#es-module-exports), either be placed on an options object or replaced with pairs of get/set functions that modify an unexported local variable.
* Groups of related constants should become a suitably named `@enum` where appropriate.

### Specific recommendations

* `OPPOSITE_TYPE` should be moved to `connection_type.js`—and possibly also be renamed to make code that uses it is more readable—e.g., `let x = oppositeConnectionType[y]` instead of `… = OPPOSITE_TYPE[y]`.  (There's no rule that  _requires_ objects that happen never to be mutated to be declared in `CONSTANT_CASE`.)
* …

cpcallen:
Actually, come to think of it, we should probably also go over `core/contstants.js` and make sure that anything in it that has a more suitable home is moved, too.

alschmiedt:
Ok I did some work on this but I am planning on putting it down for now. The current state of this is:
- `constants.js` are meant for values we expect developers to be able to use but not change. However, there are only two values in here so this might change depending on what happens to those values.
- `config.js` are meant for values that we expect developers to be able to use and to change
- Any value that should actually be an internal constant should go in a module it relates to or its own module. See [sprites.js](https://github.com/google/blockly/blob/develop/core/sprites.js).

**All the values in internal_constants that still need a new home:** 

- `COLLAPSE_CHARS` - I believe this is a contender for being passed into the options struct since we already have a collapse option and allowing a developer to specify how many characters to allow in the same spot makes sense to me. However, I held off since we don't really have a good set of rules on what should/should not go into the options struct.
[...]

PR #5743, opened 2021-11-29
feat: clarify variable and procedure constants

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Needed to make generator migration (#5716) work.
Part of #5158 

### Proposed Changes

Split `Blockly.VARIABLE_CATEGORY_NAME` (and related constants for procedures and dynamic variables). This was being used in two ways: 
1. As a way to distinguish between types of entity names in generated code (e.g. procedures vs variables).
2. As a constant that could be used in toolbox XML to trigger the appropriate dynamic toolbox category.

To address 1, I added a `NameType` enum in `names.js`, and used it throughout `names.js`. This will also be used from the generators.

To address 2, I added `CATEGORY_NAME` properties in `variables.js`, `variables_dynamic.js`, and `procedures.js`. I used those properties in `workspace_svg.js`.

#### Behavior Before Change

Same name was being used two different ways

#### Behavior After Change

Names have been split.

### Reason for Changes

See above.

### Documentation

I will need to add deprecation warnings on the properties exported from `Blockly`, but only once I stop using these names in generators.


core/procedures.js
```diff
@@ -35,6 +35,16 @@ const {Workspace} = goog.require('Blockly.Workspace');
 goog.require('Blockly.Events.BlockChange');
 
 
+/**
+ * String for use in the "custom" attribute of a category in toolbox XML.
+ * This string indicates that the category should be dynamically populated with
+ * procedure blocks.
+ * @const {string}
+ * @alias Blockly.Procedures.PROCEDURE_CATEGORY_NAME
+ */
```
cpcallen, line 44: I think it might be worth having a "see also" referring to the `CATEGORY_NAME` in the other two modules (and the same there).

cpcallen (approved):
RIP "realm". 😢

core/procedures.js
rachel-fenichel: Done.

PR #5745, opened 2021-11-30
chore(build): Update compiler & linter input language to ES2020

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Also:

- Ensure that the `comma-dangle` rule will not be applied to function parameter lists (even when multi-line).

- Update `tests/node/.eslintrc.json` to make the environment node-specific and not pinned to es6.

### Reason for Changes

Proximately this is to allow use of spread syntax in object literals (`const foo = {...bar, baz: 'quux'}`).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -64,10 +64,17 @@
         // Blockly uses objects as maps, but uses Object.create(null) to
         // instantiate them.
         "guard-for-in": ["off"],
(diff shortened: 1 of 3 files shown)


cpcallen:
(Took a few tries to get the correct value for `language_in`, but I believe I have it now…)

rachel-fenichel (approved):

PR #5749, opened 2021-11-30
refactor: migrate `generators/dart.js` to `goog.module` syntax

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5598 

### Proposed Changes

Changes generators/dart.js to goog.module and fixes missing requires.


generators/dart.js
```diff
@@ -299,3 +300,5 @@ Blockly.Dart.getAdjusted = function(block, atId, opt_delta, opt_negate,
   }
   return at;
 };
+
+exports = Dart;
```
cpcallen, line 305: I think you have done it the preferable way since (as I understand it) mostly importers will be calling methods on the `Dart` object, but:

We could alternatively use named exports:

```suggestion
exports.Dart = Dart;
```

cpcallen (approved):

alschmiedt (approved):

generators/dart.js
rachel-fenichel: Yeah i went for default export because of how it's used in the block generator files.

Issue #5466, opened 2021-09-15
Move test helpers to core


alschmiedt:
**Is your feature request related to a problem? Please describe.**

We use test helpers found in dev tools to write tests in core. The fact that we were using an external package was causing pain while trying to do the goog_module conversion because we kept having to release dev-tools to get CI to pass. 

**Describe the solution you'd like**

When we have figured out multiple entry points in core we should move test helpers out of dev-tools and into core. Plugins that use the test helpers would import them from the blockly package that they already import.

**Describe alternatives you've considered**

**Additional context**

gonfunko:
This was done in #5969 but just never got closed.

PR #5736, opened 2021-11-29
refactor: Migrate `blocks/procedures.js` to goog.module syntax

cpcallen:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](
      https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] My code is presented in the form suggested in the [module
      conversion guide](https://github.com/google/blockly/issues/5026)
- [X] I have run `npm test`.

### Resolves

Part of #5597 

### Proposed Changes

Converts `blocks/procedures.js` to `goog.module` with ES6 `const`/`let`.

PR #5751, opened 2021-11-30
fix: procedure callers now remove inputs properly

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Fixes a bug where caller blocks would not update properly when an input was removed. 
This was found because of @BeksOmega tests in the block-plus-minus plugin. 🎊 
We should probably add a similar test in core. [[Filed issue](https://github.com/google/blockly/issues/5466)]

### Proposed Changes
Use `this.arguments_.length` instead of `itemCount_`. `itemCount_` is used in text & list blocks but not procedures. I think this was introduced when we were doing a [larger refactor](#5626) that included all the blocks.

#### Behavior Before Change
Remove an input on a procedure breaks the caller block.

#### Behavior After Change
Procedures work!

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -808,7 +808,7 @@ const procedureCall = {
       }
     }
     // Remove deleted inputs.
-    for (let i = this.itemCount_; this.getInput('ARG' + i); i++) {
+    for (let i = this.arguments_.length; this.getInput('ARG' + i); i++) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Please rebase to make sure you have this change: https://github.com/google/blockly/pull/5736

rachel-fenichel (approved):

PR #5757, opened 2021-12-01
Remove declare legacy serialization

alschmiedt:
## The basics

- [x] I have run `npm test`.

### Proposed Changes
Removes declareLegacyNamespace() from core/serialization* and re exports them on the Blockly object. 

### Additional Information
@cpcallen gave two suggestions for the best way to export these: 
```
const blocksSerialization = goog.require('Blockly.serialization.blocks');
exports.serialization = {
  blocks: blocksSerialization,
  …
};
```
or
```
exports.serialization = {
  blocks: {
    foo: blocksSerialization.foo
    bar: blocksSerialization.bar
  },
  …
};

```
I went with the first option because it seemed a bit easier to maintain. Adding a file every time is easier to remember than having to remember every time we add a function. But lmk if you have other opinions and we can discuss : )


cpcallen:
To summarise the investigation of the test failures:

Until now, in uncompiled mode, modules were being loaded in the following (partial) order: `Blockly.serialization.*`, `Blockly.Msg`, `Blockly`:
1. The `declareLegacyNamespace` calls in `core/seralizaiton/*.js` would create a global `Blockly` object (to put the `Blockly.serialization` namespace object on) and set the debug module loader's `goog.implicitNamespace_['Blockly']` to true.
0. The translation loading hack in `core/msg.js` would call `.declareLegacyNamespace` and add a `.Msg` property to `Blockly`, setting the value to that (`Blockly.Msg`) module's `exports` object.
0. Having finished loading `core/msg.js`, the debug module loader would notice that `.declareLegacyNamespace` had been called and call `goog.constructNamespace_('Blockly.Msg', …)`; `constructNamespace_` would notice that a `Blockly` object already existed, so it would not try to (redundantly) set `goog.implicitNamespace_['Blockly']` to true.
0. The `goog.module('Blockly')` call in `core/blockly.js` would check `goog.isProvided_('Blockly')`, which would return false because although a `Blockly` object existed `goog.implicitNamespace_['Blockly']` was true.

After @alschmiedt attempted to remove `declareLegacyNamespace` calls from `core/serialization/*.js` in this PR (#5757), there were no calls to this function before the one in `core/msg.js`, so what happened was:
1. The translation loading hack in `core/msg.js` would call `.declareLegacyNamespace` **and create a `Blockly` object ** and add a `.Msg` property to it.
0. Having finished loading `core/msg.js`, the debug module loader would notice that `.declareLegacyNamespace` had been called and call `goog.constructNamespace_('Blockly.Msg', …)`; `constructNamespace_` would notice that a `Blockly` object already existed, so it would not try to set `goog.implicitNamespace_['Blockly']` to true.
0. The `goog.module('Blockly')` call in `core/blockly.js` would check `goog.isProvided_('Blockly')`, which would return **true** because a `Blockly` object existed **and `goog.implicitNamespace_['Blockly']` was falsy** (actually: `undefiined`), causing `goog.module` to throw.

I've therefore filed PR #5768, after which (but before PR #5757) what will happen will be:
1. `core/serialization/*.js` will call `declareLegacyNamespace`; a `Blockly` object will be created (with `.serialization` property added) and `goog.implicitNamespace_['Blockly']` will be set to true.
0. `core/msg.js` will not call `declareLegacyNamespace`; it wile export a single `null`-prototype object as `exports.Msg`.
0. `core/blockly.js` will call `goog.module('Blockly')`; the check for `goog.isProvided_('Blockly') will succeed because `goog.implicitNamespace_['Blockly']` is true.
0. `core/blockly.js` will detect the existence of the `Blockly` object (because `'Blockly' in globalThis` is true) and do nothing special.
0. Having finished loading `core/blockly.js`, the debug module loader will call `goog.exportPath_` with the `overwriteImplicit` flag set to false; this will result in properties from the `exports` object (including `.Msg`) being copied on to the `Blockly` object created earlier as usual.

After both #5768 and #5757 have been merged, what will happen will be:

1. `core/serialization/*.js` will **not** call `declareLegacyNamespace`; `goog.implicitNamespace_['Blockly']` **will remain unset**.
0. `core/msg.js` will not call `declareLegacyNamespace`; it wile export a single `null`-prototype object as `exports.Msg`.
0. `core/blockly.js` will call `goog.module('Blockly')`; the check for `goog.isProvided_('Blockly') will succeed **because no `Blockly` global variable is defined**.
0. `core/blockly.js` **will _then_ detect that there is no `'Blockly' in globalThis` and create a `globalThis['Blockly'] = {Msg}`**.
0. Having finished loading `core/blockly.js`, the debug module loader will call `goog.exportPath_` with the `overwriteImplicit` flag set to false; this will result in properties from the `exports` object (including `.Msg`) being copied on to the `Blockly` object created earlier as usual; **`Blockly.Msg` will get overwritten but with the same value it was set to in the previous step**.

alschmiedt:
@cpcallen rebased and all tests are passing. PTAL when you have a second.

cpcallen (approved):

Issue #5759, opened 2021-12-01
Drag workspace failed, mouse sticky with the interface and cannot detached. Uncaught Error: Couldn't clear and hide the drag surface. A node was missing.


rotoava:
**Describe the bug**

Drag workspace failed, mouse sticky with the interface and cannot detached.

**To Reproduce**

1. Open \tests\playgrounds\iframe.html  in chrome
2. Drag the workspace (not blocks) and mouses up outside of the iframe
3. Moves the mouse back in the workspace and hover directly over the scrollbar
4. Click once on the scrollbar
5. Moves mouse and click on the workspace
6. See error

**Expected behavior**
     
The mouse can be restored and detached from the interface, 
and the workspace can be dragged and dropped again normally.

**Desktop**
 - OS: macOS, Windows
 - Browser: chrome

**Stack Traces**

```
workspace_drag_surface_svg.js:127 Uncaught Error: Couldn't clear and hide the drag surface. A node was missing.
    at Blockly.WorkspaceDragSurfaceSvg.clearAndHide (workspace_drag_surface_svg.js:127)
    at Blockly.WorkspaceSvg.resetDragSurface (workspace_svg.js:1306)
    at Blockly.Scrollbar.onMouseUpHandle_ (scrollbar.js:1072)
    at HTMLDocument.wrapFunc (browser_events.js:62)
```

PR #5764, opened 2021-12-01
fix: Uncaught Error: Couldn't clear and hide the drag surface

rotoava:
## The basics

- [x] I branched from develop

### Resolves

Issue #5759

#### Behavior Before Change

Drag workspace failed, mouse sticky with the interface and cannot detached.
#### Behavior After Change

Workspace can be dragged and dropped again normally.
### Reason for Changes
This change prevents the ```resetDragSurface``` function being called twice and raising the exception. 
It also maintains symmetry with the ```setupDragSurface``` function in terms of logic checks.

### Test Coverage

Tested on: 
Desktop Chrome

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1347,6 +1347,16 @@ WorkspaceSvg.prototype.resetDragSurface = function() {
     return;
   }
 
+  // This can happen if the user starts a drag, mouses up outside of the
+  // document where the mouseup listener is registered (e.g. outside of an
+  // iframe) and then moves the mouse back in the workspace and
+  // hover directly over the scrollbar, click once on the scrollbar,
+  // and then click on the workspace.
+  if (!this.isDragSurfaceActive_) {
+    console.warn('Tried to reset workspace drag surface twice.');
+    return;
+  }
+
   this.isDragSurfaceActive_ = false;
 
   const trans = this.workspaceDragSurface_.getSurfaceTranslation();


gonfunko:
Thanks for reporting this issue and sending the PR! I was able to repro as you described in the bug, and this does nicely prevent the JS error. I would like to try to address the root cause, which seems to be that we're not getting mouseup events from outside the frame/window - do you mind if we hold off on merging this for a bit while we look into that and prep for our upcoming release?

rotoava:
@gonfunko 
Thank you for your immediate response and understanding. It makes the most sense to carefully examine the root cause and find the best solution.
And the problem itself is only triggered by the user at a very low probability.

Also for this single issue (triggered by scrollbar), it would be an optional solution to check if ```this.workspace_.isDragSurfaceActive_``` is ```true``` in scrollbar.js file under the ```onMouseDownHandle_``` method to prevent the confusing dragging twice. But ```isDragSurfaceActive_``` doesn't seem to be exposed, and this only solves this with scrollbar. (in case anyone want to know this)

BeksOmega:
Hello! Thank you again for posting this, and I'm sorry we never got back to you :/ There has been a lot of churn on the codebase since this was posted, so I'm going to go ahead and close this PR. If you're still interested in working on/fixing this issue, feel free to open a PR! Or if you're ever interested in working on something else, always feel free to reach out =)

PR #5766, opened 2021-12-01
refactor: Migrate `blocks/text.js` to `goog.module` syntax

cpcallen:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](
      https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] My code is presented in the form suggested in the [module
      conversion guide](https://github.com/google/blockly/issues/5026)
- [X] I have run `npm test`.

### Resolves

Part of #5597 

### Proposed Changes

Converts `blocks/text.js` to `goog.module` with ES6 `const`/`let`.

### Additional Information

Have also done some common-property factoring as was done for `Blockly.blocks.procedures` in PR #5736.

PR #5768, opened 2021-12-01
refactor: `core/msg.js`: use named export, remove `declareLegacyNamespace`

### LINKED ISSUE #5614: Remove declareLegacyNamespace from msg.js

This issue requires more research into the correct solution.

Overall, the end goal is to simply be able to remove `goog.module.declareLegacyNamespace()` from `msg.js` without breaking translations. The current problem is that when we remove `declareLegacyNamespace` it will seal the object, making it impossible for us to add our translations as we have done in the past. 

**Additional context**
This will probably involve looking at how we want to do translations moving forward. Look at #5409 for more information.
This is also related to #5535 which tracks ongoing work removing declareLegacyNamespace from `requires.js` and `blockly.js`.

Removing all calls to `declareLegacyNamespace` is important because it means we will no longer have two Blockly objects hanging around. This will make reasoning about our code a lot easier and potentially help with some bugs that have been popping up.

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #5614.
Fixes failures loading in uncompiled mode caused by PR #5757.

### Proposed Changes

In `core/msg.js`:

* Remove the `goog.module.declareLegacyNamespace()` call.
* Use a null-prototype `@type {!Object<string>}` object as the messages dictionary, rather than using the `Blockly.Msg` exports object itself.
* Export it as a named export.
* Remove the translation-loading hack.

In `core/blockly.js`:

* Re-export the messages dictionary as `Blockly.Msg`.
* Provide a new, slightly more conservative granslation-loading hack.

### Reason for Changes

See "Resolves", above.

### Test Coverage

Passes `npm test`.

Tested on:
* Desktop Chrome


alschmiedt:
So it looks like the mocha tests are failing because `Blockly.serialization.blocks` and other serialization namespaces were undefined. 

I added in my change from [this PR ](https://github.com/google/blockly/pull/5757)to see if that would fix it. It seemed to fix that problem, but then runs into the error below: 
```
Cannot use 'in' operator to search for 'Blockly' in undefined 
```
with the offending line being 
```
if (!('Blockly' in globalThis['Blockly'])) {
```
in blockly.js

cpcallen:
> ```
> Cannot use 'in' operator to search for 'Blockly' in undefined 
> ```
> 
> with the offending line being
> 
> ```
> if (!('Blockly' in globalThis['Blockly'])) {
> ```
> 
> in blockly.js

You will know by now that I generally take the [belt-and-braces approach](https://www.phrases.org.uk/meanings/belt-and-braces.html) but in this case it would appear that my trouser-fastenings have inadvertently become entangled.

The condition had originally been `!globalThis['Blockly']` and had intended to change it to `!('Blockly' in globalThis)` but apparently I gave up half way through the edit.  (Probably thinking too much about dinner.)

> So it looks like the mocha tests are failing because `Blockly.serialization.blocks` and other serialization namespaces were undefined.
> 
> I added in my change from [this PR ](https://github.com/google/blockly/pull/5757)to see if that would fix it. It seemed to fix that problem, but then runs into the error below:

I didn't touch the `goog.module.declareLegacyNamespace` calls in `core/serialization/*.js` in this PR, so I'm not quite sure what the problem is.  Having fixed the error above I don't see any Mocha test failures locally or in CI.

alschmiedt (approved):

cpcallen:
Merging this after #5766 will break text blocks (due to import of `Msg` vs `{Msg}`).  Will fix before merging.

PR #5770, opened 2021-12-02
refactor: convert some block generators to goog.module

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5598

### Proposed Changes

Converts to goog.module with named requires, and fixes missingRequires, for:
- All files in `generators/php/`
- `generators/php.js`
- `generators/lua.js`

### Additional Information

▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▰▱▱▱▱▱▱ 80%


generators/php.js
```diff
@@ -34,7 +35,7 @@ Blockly.PHP = new Blockly.Generator('PHP');
  * accidentally clobbering a built-in object or function.
  * @private
```
maribethb, line 36: this is just calling a function, right? what exactly is being marked private? we shouldn't have anything listed private on the exported object

maribethb (approved):

generators/php.js
rachel-fenichel: Good point--the private tag doesn't make sense here. It's the same in the other language generators, so I'm going to remove it for all five generators in a commit in my next pr.

PR #5777, opened 2021-12-03
chore: Simplify NPM package wrappers, improve chunk wrapper generator

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Simplify the npm package module wrappers in `scripts/package/`:

* Slightly improve documentation for each file based on helpful explanations given by @samelhusseini.
* Remove redundant code---e.g., loading `javascript_compressed.js` creates and sets `Blockly.JavaScript` as a side effect, so there is no need to set `Blockly.JavaScript = BlocklyJavaScript` in `dist/javascript.js` (generated from `scripts/package/javascript.js`).
* Remove possibly harmful code---e.g., `Blockly.Msg` is initialised with a null-prototype object in `blockly_compressed.js` and that initial object should under no circumstances be replaced.
* Remove (or fix) downright misleading code---e.g., `dist/blocks.js` previously _appeared_ to replace `Blockly.Blocks` with an empty object, but in fact the `Blockly` name referred at that point to the exports object from `blocks_compressed.js`, which would randomly get a useless `{}`-valued `.Blocks` property tacked on to it; similarly, code in `dist/browser.js` (generated from `scripts/package/browser/index.js`) appeared to copy definitions from `BlocklyBlocks` to `Blockly.Blocks`, but the former would always be (the aforementioned) empty object, making this code ineffective.

Make several improvements to the `chunks` global and `chunkWrapper`
function in `scripts/gulpfiles/build_tasks.js`:

- Document the `chunk` definition format (and improve the names of of the documented properties).

- Replace the chunk `.namespace` property with two others:
  - `.exports` names the variable/property to be returned by the `factory` function, and which will be set on the global object if the module is loaded in a browser.
  - `.importAs` names the parameter that this chunk's exports value  is to be passed to the `factory` function of other chunks which depend on this one.  (This needs to be different because e.g. `Blockly.blocks` is not a valid parameter name.)

- Change the definition for the blocks chunk to export `Blockly.Blocks` (i.e., the block definition dictionary) as `blocks_compressed.js` did previous to PR #5721 (chunked compilation), rather than the (empty and soon to vanish) `Blockly.blocks` namespace object.

 This is a win for backwards compatibility, though it does mean that if we want to expose the `loopTypes` export from `blocks/loops.js` we will need to find a different way to do so.

### Test Coverage

`npm test` succeeds.


rachel-fenichel:
>  but in fact the Blockly name referred at that point to the exports object from blocks_compressed.js, which would randomly get a useless {}-valued .Blocks property tacked on to it;

:open_mouth:

scripts/gulpfiles/build_tasks.js
```diff
@@ -66,7 +66,25 @@ const NAMESPACE_OBJECT = '$';
...
+ *   (sans prefix) will be set in the global scope to that returned
+ *   value if the module is loaded in a browser.
+ * - .importAs: the name that this chunk's exports object will be
+ *   given when passed to the factory function of other chunks that
+ *   depend on it.  (Needs to be distinct from .exports since (e.g.)
+ *   "Blockly.blocks.all" is not a valid variable name.)
+ * - .factoryPreamble: code to override the default wrapper factory
+ *   function preamble.
+ * - .factoryPostamble: code to override the default wrapper factory
+ *   function postabmle.
```
rachel-fenichel, line 87: TIL postamble is a word. Thanks!

cpcallen: I actually had to check that myself.  I was very glad to discover that it was.

scripts/gulpfiles/build_tasks.js
```diff
@@ -283,8 +313,8 @@ function chunkWrapper(chunk) {
   const amdDeps = fileNames.join(', ');
   const cjsDeps = fileNames.map(f => `require(${f})`).join(', ');
   const browserDeps =
-      chunk.dependencies.map(d => `root.${d.namespace}`).join(', ');
-  const imports = chunk.dependencies.map(d => d.namespace).join(', ');
+      chunk.dependencies.map(d => `root.${d.exports}`).join(', ');
+  const imports = chunk.dependencies.map(d => d.importAs).join(', ');
```
rachel-fenichel, line 317: Every chunk previously had a `namespace` property, but not all chunks have an `importAs` property (in fact, only one does). Is that an issue?

cpcallen: Only chunks which are depended upon by other chunks need an `importAs`, and at the moment only the first chunk is depended upon by any others.

cpcallen: (Arguably I should add a robustness check here, such that if it discovers a chunk that has become a dependancy but has no `importAs` it will complain.  Feel free to open a bug about that if you like…)

rachel-fenichel: I see. Then this works fine here, but it's fragile.

rachel-fenichel: I'll file a low priority bug after I finish this review

scripts/gulpfiles/package_tasks.js
```diff
@@ -126,8 +126,8 @@ function packageBlockly() {
  */
 function packageBlocks() {
   return gulp.src('scripts/package/blocks.js')
-    .pipe(packageUMD('Blockly.Blocks', [{
-        name: 'Blockly',
+    .pipe(packageUMD('BlocklyBlocks', [{
```
rachel-fenichel, line 129: Why did you keep the `scripts/package/thing.js` files, while emptying them of code, instead of just cutting that step off the pipeline entirely? Is there a benefit to starting the pipeline with an actual file, or is it just serving to inject the license and comment about the wrapper?

scripts/gulpfiles/build_tasks.js
rachel-fenichel: After reading more: I expect that this code will continue to change into next quarter, particularly when we mess around with entry points, so I'll hold off on that bug.

rachel-fenichel (approved):
Approved, with the caveat that the packaging pipeline feels fragile and we're definitely going to go over it again next quarter.

scripts/gulpfiles/package_tasks.js
cpcallen: Mainly to try to not break anything, TBH, but yes: the documentation in the resulting files in `dist/` is helpful (and this improved documentation might have helped me considerably when I started trying to understand this mess some months ago).

I eventually think most of this should go away:

- Why not just copy `build/blockly_compressed.js` to `dist/blocklyjs` instead of having a separate wrapper module?
- Even where a wrapper module is needed, these ones are all so simple that having a build pipeline with a separate build function to put a UMD wrapper around each of module is absurdly heavyweight.  Surely we could just put pre-UMD-wrapped static files in `scripts/package` and just copy them into `dist/` unmodified?

But in the mean time I figured that feeding an effectively-empty file to the existing pipeline was the safest thing to do for the time being.

PR #5781, opened 2021-12-03
fix(tests): Fix & reenable the advanced compilation test

### LINKED ISSUE #5602: Re-enable compile test in run_all_tests.sh
```
$ npm run test:compile:advanced

> blockly@7.20211209.0-beta.0 test:compile:advanced
> gulp buildAdvancedCompilationTest

[11:22:24] Using gulpfile ~/Code/blockly/gulpfile.js
[11:22:24] Starting 'buildAdvancedCompilationTest'...
gulp-google-closure-compiler: tests/compile/test_blocks.js:15:0: ERROR - [JSC_MISSING_MODULE_OR_PROVIDE] Required namespace "Blockly.Blocks" never defined.
  15| goog.require('Blockly.Blocks');
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 error(s), 0 warning(s)

[11:22:26] 'buildAdvancedCompilationTest' errored after 2.35 s
[11:22:26] Error: gulp-google-closure-compiler: Compilation errors occurred
    at CompilationStream._compilationComplete (/Users/fraser/Code/blockly/node_modules/google-closure-compiler/lib/gulp/index.js:238:28)
    at /Users/fraser/Code/blockly/node_modules/google-closure-compiler/lib/gulp/index.js:208:14
```

Furthermore, in Blockly Games, nearly all games fail with:
```
compressed.js:289 Uncaught Error: Error when registering mutator "controls_if_mutator": Mutations must contain either XML hooks, or JSON hooks, or both
    at Hh (compressed.js:289)
    at Object.Fh.ut (compressed.js:286)
    at compressed.js:713
```

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #5602.

### Proposed Changes

* Fix advanced compilation test:

  - Fix loading of blocks in `tests/compile/main.js` caused by recent `goog.module`-ification work.
  - Fix problem caused by ADVANCED_OPTIMISATIONS renaming the `Msg` property on the fake `Blockly` object created by the translation-loading hack in `blockly.js`.

* Reenable advanced compilation test 

#### Behavior Before Change

Test was disabled and if run manually produced un-loadable code.

#### Behavior After Change

Test enabled, runs as part of `npm test`, and produces code that loads and works correctly.

### Reason for Changes

We like tests.

### Test Coverage

Tested on:
* `npm test`
* Desktop Chrome

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -742,7 +742,7 @@ exports.zelos = zelos;
 // Blockly.Msg module - so make sure it is, but only if there is not
 // yet a Blockly global variable.
 if (!('Blockly' in globalThis)) {
-  globalThis['Blockly'] = {Msg};
+  globalThis['Blockly'] = {'Msg': Msg};
 }
 
 // Temporary hack to copy accessor properties from exports to the
--- a/tests/compile/main.js
+++ b/tests/compile/main.js
@@ -11,16 +11,8 @@ goog.require('Blockly');
 goog.require('Blockly.geras.Renderer');
 goog.require('Blockly.VerticalFlyout');
 // Blocks
-goog.require('Blockly.Constants.Logic');
-goog.require('Blockly.Constants.Loops');
-goog.require('Blockly.Constants.Math');
-goog.require('Blockly.Constants.TestBlocks');
-goog.require('Blockly.Constants.Text');
-goog.require('Blockly.Constants.Lists');
-goog.require('Blockly.Constants.Colour');
-goog.require('Blockly.Constants.Variables');
-goog.require('Blockly.Constants.VariablesDynamic');
-goog.require('Blockly.blocks.procedures');
+goog.require('Blockly.blocks.all');
+goog.require('Blockly.blocks.testBlocks');
 
 Main.init = function() {
   Blockly.inject('blocklyDiv', {
--- a/tests/compile/test_blocks.js
+++ b/tests/compile/test_blocks.js
@@ -9,7 +9,7 @@
  */
 'use strict';
 
-goog.provide('Blockly.Constants.TestBlocks');
+goog.provide('Blockly.blocks.testBlocks');
 
 goog.require('Blockly');
 
--- a/tests/run_all_tests.sh
+++ b/tests/run_all_tests.sh
@@ -76,8 +76,8 @@ run_test_command "package" "npm run package"
 # Run Node tests.
 run_test_command "node" "./node_modules/.bin/mocha tests/node --config tests/node/.mocharc.js"
 
-# # Attempt advanced compilation of a Blockly app.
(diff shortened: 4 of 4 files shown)


core/blockly.js
```diff
@@ -742,7 +742,7 @@ exports.zelos = zelos;
 // Blockly.Msg module - so make sure it is, but only if there is not
 // yet a Blockly global variable.
 if (!('Blockly' in globalThis)) {
```
rachel-fenichel, line 744: Is there a case where there was already a Blockly global variable *and* we need to be able to refer to it by `Blockly.Msg` instead of by importing it by name?

rachel-fenichel (approved):

rachel-fenichel:
Changed to 'fix(tests)' and merged.

core/blockly.js
cpcallen: I'm not sure I understand the question, so I will say a few things in the hopes that one of them answers it:

* At this point, barring an oversight on our part, none of the code in `core/`, `blocks/` or `generators/` ever assumes the existence of a `Blockly` global variable.
* Aside from (optionally named) `goog.require` statements in tests, etc., the only code in that should create a `Blockly` global is this hack that tries to ensure that `gloablThis.Blockly.Msg` (and only that) exists, and the call to `declareLegacyNamespace` at the top of this file that will create a global `Blockly` variable when this file is loaded in uncompiled mode or (with help from the UMD wrapper) if `blockly_compressed.js` is imported with a `<script>` tag into the browser.
* Most "outside" code should be importing Blockly via either an ES `import` statement or, if using Closure modules/compiler, a `goog.require` (e.g., Blockly Games and the advanced compilation test (ACT)).
  * I'm actually slightly unclear about importing Blockly using `import` works outside of node, as I wasn't aware that browsers have a way of loading UMD modules in response to ES `import` statements, but I would generally expect that no globals would be created as a result unless the import was being done as a named import at the top (script) level, in which case the _importer_ is probably going to name the import `Blockly`.
  * For BG and the ACT, AVANCED_OPTIMIZATIONS results in everything (including `Blockly`) being renamed, so there a bunch of random 1- and 2-character variable names that are meaningless without a sourcemap.
* In either case, unless the importer has gone out of their way to ensure that the `Blockly` global variable exists and points at the export object of this (`blockly.js` / `Blockly`) module, trying to load translations via `<script>` will fail.  This means that the following excerpt from `tests/compile/index.html`:
  ```
    <script src="main_compressed.js"></script>
    <script src="../../msg/js/en.js"></script>
  ```
  would throw when `en.js` tried to do `Blockly.Msg["ADD_COMMENT"] = "Add Comment"` because `Blockly.Msg` was undefined (until I fixed the issue in the first commit (8370647) of this PR).

Issue #4959, opened 2021-06-28
Deleting comment when starting block drag gesture can cause rendering issue


jschanker:
**Describe the bug**

If the `?` comment button is removed from a block when starting a drag gesture, its child blocks may appear incorrectly displaced during the drag.

**To Reproduce**

1. Go to the Blockly Playground.
2. Import the following XML:
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="text_print" id="fpaqx?}yfPa/c`r|;NgZ" x="88" y="187">
    <value name="TEXT">
      <shadow type="text" id="6`d-*pZ`~Ey~A5h!ci#u">
        <field name="TEXT">abc</field>
      </shadow>
      <block type="math_arithmetic" id="^JLjGxS%Yh%Wo.egTOF{">
        <field name="OP">ADD</field>
        <comment pinned="true" h="80" w="160">abc</comment>
        <value name="A">
          <shadow type="math_number" id="(F`%1a`sD@*BIm.b3:xc">
            <field name="NUM">1</field>
          </shadow>
          <block type="math_number" id="mXPZWVVaPqHZR6Xv%]-S">
            <field name="NUM">123</field>
          </block>
        </value>
        <value name="B">
          <shadow type="math_number" id="0et:%d5L~O7|:]Z#29Mn">
            <field name="NUM">1</field>
          </shadow>
          <block type="math_number" id="o$XCP:?26fxG=$CJy9eO">
            <field name="NUM">123</field>
          </block>
        </value>
      </block>
    </value>
  </block>
</xml>
```
3. Enter the following in the Developer console:
```
workspace.addChangeListener(e => {
  if(e.type === Blockly.Events.BLOCK_DRAG) {
    workspace.getAllBlocks().find(block => block.type === "math_arithmetic").setCommentText(null)
  }
});
```
4. Drag the arithmetic block out of the print block, and notice the number blocks are displaced during the drag.

**Expected behavior**

The blocks should not appear displaced during dragging.

**Additional context**

To understand the use case where this could occur, I am using comments to provide feedback to users when they assemble blocks.   If they connected the wrong blocks, a comment appears to let them know as much.  The comments disappear when they disconnect the blocks.

alschmiedt:
Wow this really breaks. 

For whoever is going to take this issue the last release it looks like it was working in is 3.20200924.4.

jschanker:
[...]

Issue #5788, opened 2021-12-05
Deleting a block while dragging can cause major rendering issues that don't correct at end of drag


jschanker:
**Describe the bug**

If one or more of a block's ancestors is/are disposed of while dragging the block, the remaining root block's ancestors can be all over the place after deletion and upon completion of the drag.

**To Reproduce**

1. Go to the Blockly Playground at http://google.github.io/blockly/tests/playground.html.
2. Import the following XML:
```
<xml xmlns="https://developers.google.com/blockly/xml">
  <block type="controls_whileUntil" x="88" y="188">
    <field name="MODE">WHILE</field>
    <value name="BOOL">
      <block type="logic_boolean">
        <field name="BOOL">FALSE</field>
      </block>
    </value>
    <next>
      <block type="controls_repeat_ext">
        <value name="TIMES">
          <block type="math_number">
            <field name="NUM">123</field>
          </block>
        </value>
        <statement name="DO">
          <block type="text_print">
            <value name="TEXT">
              <block type="text">
                <field name="TEXT">A</field>
              </block>
            </value>
            <next>
              <block type="text_print">
                <value name="TEXT">
                  <block type="text">
                    <field name="TEXT">B</field>
                  </block>
                </value>
                <next>
                  <block type="text_print">
                    <value name="TEXT">
                      <block type="text">
                        <field name="TEXT">C</field>
                      </block>
                    </value>
                  </block>
                </next>
              </block>
            </next>
          </block>
        </statement>
      </block>
    </next>
  </block>
</xml>
```
3. Type the following in the Developer console:
```
workspace.addChangeListener(e => {
  if(e.type === Blockly.Events.BLOCK_DRAG) {
    setTimeout(() => workspace.getAllBlocks()
        .filter(block => block.type === 'text_print')
        .slice(1).forEach(block => block.dispose()), 5000);
  }
});
```
4. Drag the root repeat while block all around for at least 5 or so seconds.  Note the scattered blocks.

**Expected behavior**

The blocks render correctly in their appropriate places during and after the drag.

**Screenshots**

![screenshot (13)](https://user-images.githubusercontent.com/9159008/144738030-3ac9ce5d-2405-4a67-aec5-b236ed8b4539.png)

**Additional context**

This is a known issue and is one related to #4959, but I'm adding it as another one so I can reference it in a comment of a pull request I'm working on to allow for correct rendering during and after dragging.

rachel-fenichel:
[...]

PR #5789, opened 2021-12-05
Proper rendering of blocks during and after dragging

jschanker:
## The basics

- [x] I branched from develop

### Resolves

#4959 and #5788 (partially)

### Proposed Changes

* Allows methods that cause block displacement issues during/after dragging to request that the workspace render the potentially displaced blocks on relevant events up to and including the end of the drag.
* Allows comments to be removed/blocks to be disposed of while dragging workspace without causing block displacements.

#### Behavior Before Change

Removing comments/deleting blocks while dragging caused blocks to appear displaced.  (See: #4959 and #5788.)

#### Behavior After Change

Blocks no longer appear displaced during/after dragging.

### Reason for Changes

Performing certain actions while dragging blocks can cause the appearance of major block displacements so to avoid these problems, they may be deferred as follows: https://github.com/google/blockly/blob/6128f2dfd485ed2ff6d0be13b7e2558190678e2e/core/block_svg.js#L1088-L1099
This pull request represents an attempt to simplify code by allowing these types of function calls to be made during dragging.  At the same time, the hope is to do this without sacrificing performance by minimizing render calls to the blocks that may actually be affected during the events for which this can occur and only doing so on appropriate events during the drag.

### Test Coverage

Tested on Blockly Playground.

### Documentation

No extra documentation is necessary, but it may be helpful to specify how the added function `addToDragRenderQueue` can be used to eliminate some of these dragging rendering issues.

### Additional Information

This is very much a draft, but suggestions on whether this approach is useful or any other comments are very much welcome!

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -262,6 +262,26 @@ const WorkspaceSvg = function(
       this.options.rendererOverrides);
 
   /**
+   * The function that handles rendering of blocks added to queue for drag
+   *     events and removes itself from the change listeners and empties the
+   *     queue on the completion of a drag event.  
+   * @type {function(Blockly.Events.Abstract)}
+   * @private
+   */
+  this.renderOnDragListener_ = null;
+
+  /**
+   * The blocks to render on drag until next BlockDrag Event is ended.
+   * @type {!Array<
+   * {
+   *   block: !Blockly.BlockSvg,
+   *   bumpNeighbours: boolean
+   * }>}
+   * @private
+   */
+   this.blockDragRenderQueue_ = [];
+
+  /**
    * Cached parent SVG.
    * @type {SVGElement}
    * @private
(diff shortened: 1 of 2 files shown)


core/workspace_svg.js
```diff
@@ -1478,6 +1498,60 @@ WorkspaceSvg.prototype.render = function() {
...
+ *     to this queue.
+ * @param {boolean=} bumpNeighbours If true, calls bumpNeighbors on drag for
+ *     this block (and descendants if allDescendants is true).
+ */
+WorkspaceSvg.prototype.addToDragRenderQueue = function(block, addDescendants, bumpNeighbours) {
+  if (!this.isDragging()) {
+    return;
+  }
+
+  const blocks = (addDescendants ? block.getDescendants() : [block])
```
rachel-fenichel, line 1543: I think you can be a lot more discerning about which blocks get rendered, which may have an impact on performance for this change. Specifically, you shouldn't need to add descendants.

- When a set of blocks explodes during a drag, it's not that the leaf blocks were rendered wrong. It's that the parent block's rerender failed to update connection locations or, more frequently, recalculated connection locations wrong--e.g. by doubling offsets or misplacing the origin.

- Rendering "bubbles up", meaning that rendering a block will re-render its parents but not its children. So if you render a block and then one of its children, the child will render once but the parent and all blocks above it on the stack will render twice.

rachel-fenichel (commented):
I'd like to make sure I'm understanding this PR correctly:
- The goal is to enable programmatic changes mid-drag.
  - Currently this breaks by having block stacks "explode" so that blocks do not appear connected to each other, but still move in parallel.
- The mechanism is to add a list to the workspace, which tracks blocks that need to be rerendered if any block is deleted or if the drag ends.
- The list can be populated by request and is cleared when a non-start drag event is found.

Is all of that correct?

If yes, a couple of thoughts:
- The list can live on the block dragger instead of the workspace. That moves it closer to the relevant code, and also means that its state is automatically cleared at the end of the drag.
- Does order on the queue matter?
- A good way to test performance effects is to use the Spaghetti button in the playground or advanced playground, which generates a deep stack of nested blocks. Once that's loaded, you can open chrome dev tools and use the runtime performance tools to see how long mouse move handlers take during drags.
  - Here's a [tutorial](https://developer.chrome.com/docs/devtools/evaluate-performance/) on that part of dev tools. Other browsers may have equivalent tools but I'm not familiar with them.
  - In particular, you'll want to test dragging a block away from a stack; over a stack so that you pass near a valid connection; and back into a stack.
  - Baseline blocks performance on this test isn't great, so you'll be looking for a change in performance rather than raw numbers.

jschanker:
@rachel-fenichel Yes, your analysis of what I was trying to accomplish is correct.  Thanks for the comprehensive review.  This is very helpful!

In this draft pull request, I definitely wasn't being too discerning about what had to be rerendered.  And from what you wrote about rendering bubbling up, it seems rendering a stack of n blocks for large n would cause bad performance of &Omega;(n<sup>2</sup>) renders (a block m down in the stack would cause a chain of m renders?)  I haven't done much to understand Blockly rendering yet so I'll look into this more.  Empirically, I noticed that removing a comment icon while dragging could case a displacement during drag that resolves itself when the drag is complete.  But if you render when dragging to fix it but not at the completion, it's off when it finishes.  A quick look had made me think that tighten might have something to do with it.  I'll investigate more according to your recommendations.  Thanks again!

BeksOmega:
I think there might be some doc updates that can be done if this rad fix gets added. I remember telling people to check isDragging before making changes to the block, but I'm not sure if that info ever got added to the devsite.

BeksOmega:
I'm gonna go ahead and close this as stale. Feel free to reopen if you're still interested in working on this!  But just a heads up, we might be digging into the rendering code and doing some performance upgrades in the next quarter or two (I'm not exactly sure if that's part of the planned projects or not) so this area of the code might get a bit chaotic hehe.

jschanker:
Sorry for the delay on working on this; I definitely do want to return to this at some point!  I'm hoping to do so during the summer.

Issue #5812, opened 2021-12-10
Setting default saturation and value


rd--:
**Describe the bug**

I've been setting Blockly.HSV_SATURATION and Blockly.HSV_VALUE, as described at:

https://groups.google.com/g/blockly/c/kTIxjJa39TM/m/tESoM6wofzQJ

however this seems to have stopped working recently? 

I'm loading Blockly using: <script src="https://unpkg.com/blockly/blockly.min.js"></script>

Is there a new way to do this?

I see they're now internalConstants exported for "legacy reasons": 

https://github.com/google/blockly/blob/02a57122debaa4ca043312dec547944afae658c4/core/blockly.js#L627

There are no console errors, and Blockly.HSV_VALUE prints as the value it's been set to, but the colours of the blocks and in the toolbox don't reflect this.

This was sent to the forum, where it was suggested I should file a bug, though it's quite possible this is my mistake!  

https://groups.google.com/g/blockly/c/Zme8UGnh0Bg/m/pjAEQBjeBwAJ

**To Reproduce**

There is a version of the editor this affects at:
 
https://rohandrape.net/?t=blksc3&e=blksc3.html

There are no errors in the console, and Blockly.HSV_SATURATION and Blockly.HSV_VALUE print with the modified values, but the blocks drawn don't reflect this.

**Expected behavior**

Previously the blocks were recoloured.  I'm almost certain I didn't change anything related to this, but perhaps it was only working initially by chance.

**Screenshots**

N/A

**Desktop (please complete the following information):**

OS: Linux x1 5.14.0-4-amd64 #1 SMP Debian 5.14.16-1 (2021-11-03) x86_64 GNU/Linux

Browser: Chromium Version 90.0.4430.212 (Developer Build) built on Debian 11.0, running on Debian bookworm/sid (64-bit)

**Stack Traces**

N/A

**Additional context**

I'm quite new to Blockly, it's wonderful, this is quite possibly a beginner mistake!

rd--:
Below is a more minimal example.  I think this used to work?  Best, Rohan

````
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <script src="https://unpkg.com/blockly/blockly.min.js"></script>
  </head>
  <body>
  <div id="blocklyDiv" style="height: 480px; width: 600px;"></div>
  <xml xmlns="https://developers.google.com/blockly/xml" id="toolbox" style="display: none">
    <block type="logic_compare"></block>
[...]

PR #5821, opened 2021-12-14
fix: re-expose HSV_VALUE and HSV_SATURATION as settable properties on Blockly

rachel-fenichel:
## The basics

- [x] I branched from develop

However, this will need to be cherry-picked into a patch release.

### Resolves

Fixes the issue [reported](https://groups.google.com/g/blockly/c/Zme8UGnh0Bg) on the forum where devs can no longer set default saturation and value.

These properties were settable before goog.module-ification, and we [suggest setting them](https://developers.google.com/blockly/guides/create-custom-blocks/block-colour) in our documentation.

### Proposed Changes

- Move the two values from InternalConstants into utils/colour.js and add get and set functions.
- Update internal uses (only in utils/colour.js and utils/parsing.js)
- Add getter and setter pairs in blockly.js

Note that I marked the internal get and set functions package, because we may yet want to rearrange these.

#### Behavior Before Change

Could not set default value or saturation.

#### Behavior After Change

Can now set default value or saturation.

### Reason for Changes

Previously settable properties that are also actively in use.
Eventually we may want a more format setter for each of these properties, but this is an established pattern and it needs to continue to work.

### Test Coverage
Normal tests, plus I changed values in the playground and got this:
image

### Additional Information

This will need to be cherry-picked into a patch release.

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -444,6 +444,36 @@ Object.defineProperties(exports, {
       common.setSelected(newSelection);
     },
   },
+  /**
+   * The richness of block colours, regardless of the hue.
+   * Must be in the range of 0 (inclusive) to 1 (exclusive).
+   * @name Blockly.HSV_SATURATION
+   * @type {number}
+   * @suppress {checkTypes}
+   */
+  HSV_SATURATION: {
+    get: function() {
+      return utils.colour.getHsvSaturation();
+    },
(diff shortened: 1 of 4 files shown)


alschmiedt (approved):

alschmiedt:
Fixes #5812

PR #5832, opened 2021-12-16
chore(deps): updates the block-test devDependency

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Serialization category in test blocks is broken in the advanced playground since we are using an out of date version of block-test. 

It also causes problems when we remove the package-lock and try to do an `npm install` because the peer dependency of block-test does not match our current version.

### Proposed Changes
Update to the latest version of block-test.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "jsdom": "15.2.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


alschmiedt:
Closing in favor of #5836.

PR #5833, opened 2021-12-18
fit: Fixed the label closure on demo/blockfactory

ChrisJaunes:
## The basics

- [ ] I branched from develop

### Resolves
In demo/blockfactory, I see that the input tag is not closed in demo/blockfactory/index.html, and I fixed this

### DIFF
--- a/demos/blockfactory/index.html
+++ b/demos/blockfactory/index.html
@@ -143,9 +143,9 @@
         <div class="dropdown">
         <button id="button_importBlocks">Import Custom Blocks</button>
           <div id="dropdownDiv_importBlocks" class="dropdown-content">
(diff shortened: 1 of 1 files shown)


alschmiedt:
Nice catch and thanks for the contribution! 

It looks like we have multiple problems on these lines of code. Would you mind also going through these lines and removing `</input>`? Input is a void element and the closing tag is not necessary.

ChrisJaunes:
> Nice catch and thanks for the contribution!
> 
> It looks like we have multiple problems on these lines of code. Would you mind also going through these lines and removing `</input>`? Input is a void element and the closing tag is not necessary.

Thanks.
I have removed the </input>.

alschmiedt (approved):

alschmiedt:
Thanks!

PR #5839, opened 2021-12-21
chore: update PR template to include formatting info

alschmiedt:
## The basics

- [ ] I branched from develop

### Resolves
Part of #5604 was to find a way to tell developers what they should run when clang-format has failed. I am sure there is a way to run a GitHub action do it, but I am not sure we have the time to spend on it currently. I think having information on how to format the code in the PR will serve a similar purpose as commenting on the code. 

Even if we get the comment working, I still think we want to tell developers before their build fails that they need to run `npm run format`. 

### Proposed Changes
Adds run `npm run format` to our check list of tasks to finish before creating a PR. 

### Additional Information
The other option is to have a single command that includes lint, formatting and running the closure compiler to see any type problems.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,7 +10,7 @@
 - [ ] I branched from develop
 - [ ] My pull request is against develop
 - [ ] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
-
+- [ ] I ran `npm run format`
 ## The details
 ### Resolves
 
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
Q: Is the modified blockly style guide (previous checklist item) still relevant if we're relying on clang format now?

alschmiedt:
Good question!

It includes some information about TODOs and Jsdocs that are not covered by either clang format or eslint. However, I think we are getting closer and closer to the Google style guide and I would be very interested in not having to maintain this page if a developer can get just about all the information they need from the official Google style guide.

Also this made me realize we should also include lint in this list.

gonfunko (approved):

rachel-fenichel:
Is this ready to merge?

PR #5856, opened 2022-01-07
chore(deps): bump @blockly/block-test from 1.1.5 to 2.0.1 (#5836)

alschmiedt:
### Resolves
Clearing node_modules and removing the package-lock causes build errors because the block-test version is not updated correctly.

### Proposed Changes
Cherry pick #5836 from develop.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "jsdom": "15.2.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):
Does this mean we need a patch release?

alschmiedt:
It doesn't technically need to be released on npm but a patch on github would be nice so the tagged version has this fix.

Issue #5841, opened 2021-12-27
npm - en.d.ts not found.


johnduffynh:
Hi,
Error: ../../node_modules/blockly/index.d.ts:15:22 - error TS6053: File '.../my-workspace/node_modules/blockly/msg/en.d.ts' not found.
15 /// &lt;reference path="msg/en.d.ts" />

**To Reproduce**
```
ng new ws --create-application=true --routing=false --style=scss;
cd ws;
npm install blockly;
```
notice ws/node_modules/blockly/[index.d.ts](https://github.com/google/blockly/blob/master/typings/index.d.ts) references msg/en.d.ts.

Work around is to just add [en.d.ts](https://github.com/google/blockly/blob/master/typings/msg/en.d.ts) or remove reference.  I could not find where you do the packages to do a pull request.

Thanks for this great product!

alschmiedt:
Just a heads up we have released a [patch](https://github.com/google/blockly/releases/tag/7.20211209.2) for this and it is now available on [npm](https://www.npmjs.com/package/blockly).

PR #5846, opened 2022-01-02
fix: msg imports in typedefs.

yohcop:
Fixes #5841 for me.

## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

#5841 

### Proposed Changes

No visible changes, only compiler error fixes,

#### Behavior Before Change

Doesn't build.

#### Behavior After Change

Happy compilers.

Tested on:
* Desktop Chrome

But more importantly, I use bazel and typescript v4.5.4 to build.

PR #5858, opened 2022-01-07
fix: msg imports in type definitions

rachel-fenichel:
This is a copy of https://github.com/google/blockly/pull/5846 against master, which will fix https://github.com/google/blockly/issues/5841 in the next patch release.

When we released v7 we committed to making patch releases for typescript definition errors, since our pipeline for typings was convoluted this quarter.


rachel-fenichel:
And a hat tip to @yohcop for putting together the origin patch!

alschmiedt (approved):

PR #5868, opened 2022-01-11
feature: Replace text quotes and pilcrow with custom font

NeilFraser:
Changing from bitmaps to vectors significantly improves the visual quality of these blocks.  Also reduces the code size.

Once we have a custom font, this will allow other assets such as the dropdown arrow, the menu checkmark, the toolbox category arrows, and maybe more.

Old UI:

Screen Shot 2022-01-11 at 11 48 25

New UI:

Screen Shot 2022-01-11 at 11 48 16


NeilFraser:
Converting to a draft PR.  This change has exposed deeper pre-existing issues with text-length measurement.  Unsure how to fix.

rachel-fenichel:
Can we do a fixed-size vector image instead of a font? It would still scale better and wouldn't need to be measured.

NeilFraser:
@rachel-fenichel Let's address #5871 first, then reevaluate our position here.

BeksOmega:
I'm gonna go ahead and close this since it might be a few months before we get back around to it.

PR #5872, opened 2022-01-12
fix: typo of atan2 in japanese

tokoroten:
## The basics

- [x] I branched from develop

### Proposed Changes

fix Japanese typo.

#### Behavior Before Change

atan2 block show's `aran2` in japanese.

image

#### Behavior After Change

fix to `atan2`


maribethb:
Hi, thank you for the PR! I apologize for the delay, I was out last week and then I was looking into the right way to handle this.

In order to merge this, it needs two changes:

1. make the PR against the **translatewiki** branch instead of develop 
2. we need your commit message to look like: `fix: typo of atan2 in japanese` (`fix:` is the key element here, with the colon). You can change the commit message by following [these steps](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#rewriting-the-most-recent-commit-message), in short
    1. `git commit --amend`
    2. Force push to this branch

The reason for 1 is that the normal translation workflow involves going through TranslateWiki, so we need to make changes against their branch so it gets into their system. Any large translation changes would have to go through them entirely, but it can be a lot of overhead to get verified on their site just to fix a small change like this. So we can do small changes directly, but know that this change might get clobbered when translatewiki does their pushes, so if you see this isn't actually fixed after the next release we'll have to follow up with them.

And as for 2, we are now using[ conventional commits ](https://conventionalcommits.org)to help us organize our release notes better, and we are using tooling to enforce this for all commits. Sorry about the extra steps required, I am about to file a bug to add some info about conventional commits to our docs. 

Once you've done both of those, this looks good to merge! Thanks!

tokoroten:
Thanks for the advice.
A lot of conflicts have come up.
I will rewrite it from translatewiki.

tokoroten:
finally, I've fix it at translatewiki.
thank you for your advice.

https://translatewiki.net/wiki/Blockly:MATH_ATAN2_TITLE/ja

maribethb:
Awesome, thank you so much for contributing and fixing this issue!

Issue #5860, opened 2022-01-07
Where possible, convert to es6 classes.


rachel-fenichel:
Problematic classes (an incomplete list):

- [X] `generator.js` (Handled in #5928) 
  - Acts as both a [struct and a dict](https://github.com/google/closure-compiler/wiki/@struct-and-@dict-Annotations). We add properties after creation (like a dict) but also do . access (like a struct).
    - Can be fixed with the `@unrestricted` [annotation](https://github.com/google/closure-compiler/wiki/@struct-and-@dict-Annotations) on the class. --@rachel-fenichel 
  - Uses `Object.defineProperties` to add a property to the prototype.
- [X] `options.js`  (Handled in #5928)
  - Has a couple of functions (`setMetrics` and `getMetrics`) that don't always exist, but are declared for type purposes. Uses check whether these functions exist.
    - Declare these as data properties with `@type {function(....):... | undefined}`.  --@cpcallen
- [X] `shortcut_registry.js` (Handled in #5940)
  - Weird singleton pattern.
- [X] `contextmenu_registry.js` (Handled in #5940)
  - Weird singleton pattern.
- [X] `registry.js` (Handled in #5940)
- [X] `block_svg.js` (Handled in #5952)
  - Calls the super constructor late
- [X] `block.js` (Handled in #5952)
  - Has an unconverted subclass

### Toolbox
- [X] `toolbox/category.js` (Handled in #5947)
  - Has an unconverted subclass (`CollapsibleCategory`)
- [X] `toolbox/collapsible_category.js` (Handled in #5947)
  - I *think* the problem is that the superclass constructor uses the `registrationName` static property, which the subclass overrides. The property isn't overridden yet when the superclass constructor is called.
- [X] `toolbox/toolbox_item.js` (Handled in #5947)
  - Has an unconverted subclass (`ToolboxCategory`)

### Rendering:
- [X] `renderers/geras/path_object.js` (Handled in #5954)
  - Extends the base path object but doesn't call the super constructor. That's because the geras path object has to put a dark path in front of the main path, and z-index of svg objects is controlled by the order in which they are created.
- [X] `renderers/common/path_object.js` (Handled in #5954)
  - Has a non-converted subclass (see above)
- [X] `renderers/measurables/row.js` (Handled in #5957)
  - In code we add a few properties (`followsStatement`, `nextRightEdge`, and `rightAlignedDummyInput`) that don't really belong on this class. I could just add and describe them anyway, marked as package, if really necessary.
[...]

PR #5873, opened 2022-01-13
refactor: convert some classes to es6 classes

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5860 and part of #2892

### Proposed Changes

Converts event classes to ES6 classes.

As usual, this is best reviewed one commit at a time, and at the end I ran format and rebuilt tests/deps.js.

### Reason for Changes

Move to ES6 classes to reduce hassle when converting classes to Typescript.

### Test Coverage
Mocha tests, running build-debug to look for compiler errors, and opening up the playground to test when I changed the flyout classes in particular.


core/insertion_marker_manager.js
```diff
@@ -40,743 +40,746 @@ const {WorkspaceSvg} = goog.requireType('Blockly.WorkspaceSvg');
...
+ * could display.
+ * @enum {number}
  */
-InsertionMarkerManager.prototype.hideReplacementFade_ = function() {
-  this.fadedBlock_.fadeForReplacement(false);
-  this.fadedBlock_ = null;
+InsertionMarkerManager.PREVIEW_TYPE = {
+  INSERTION_MARKER: 0,
+  INPUT_OUTLINE: 1,
+  REPLACEMENT_FADE: 2,
```
cpcallen, line 779: This should probably be renamed `PreviewType`.  Also, I would either make it a `static` item in the `class` definition, or make it its own separate named export.

core/insertion_marker_manager.js
```diff
@@ -40,743 +40,746 @@ const {WorkspaceSvg} = goog.requireType('Blockly.WorkspaceSvg');
...
-  }
-  if (this.lastMarker_) {
-    result.push(this.lastMarker_);
-  }
-  return result;
-};
+InsertionMarkerManager.DUPLICATE_BLOCK_ERROR = 'The insertion marker ' +
+    'manager tried to create a marker but the result is missing %1. If ' +
+    'you are using a mutator, make sure your domToMutation method is ' +
+    'properly defined.';
```
cpcallen, line 791: Can this be an unexported `const`?

core/metrics_manager.js
```diff
@@ -94,387 +479,6 @@ MetricsManager.FixedEdges;
  *          }}
  */
 MetricsManager.UiMetrics;
```
cpcallen, line 485: Regarding this `@typedef` (and the previous three similar ones):

* I'd have suggested making these separate named exports, only it looks like we don't have _any_ named exports from this module.
* Can they be unexported `const`s?
* If not then I'd suggest they be exported, possibly after being factored out into a separate module that is `require`d both by this one any by any client code that needs these types.
* Otherwise perhaps they could be declared as `static` properties in the `class MetricsManager`  declaration

core/flyout_base.js
```diff
@@ -258,15 +258,13 @@ class Flyout extends DeleteArea {
...
-     * [block] ---- orthogonal to flyout boundary ----          |
-     * [block] \                                                |
-     * The angle is given in degrees from the orthogonal.
+     * Range of a drag angle from a flyout considered "dragging toward
+     * workspace". Drags that are within the bounds of this many degrees from
+     * the orthogonal line to the flyout edge are considered to be "drags toward
+     * the workspace". Example: Flyout Edge   Workspace [block] /  <-within this
+     * angle, drags "toward workspace" | [block] ---- orthogonal to flyout
+     * boundary ----          | [block] \ | The angle is given in degrees from
+     * the orthogonal.
```
cpcallen, line 267: Must disable clang-format's comment reflow here, somehow!

cpcallen (changes requested):
It was a lot easier to do this review once I figured out how to get GitHub's diff to ignore whitespace!

All very promising—but with a number of small issues to address.

core/flyout_vertical.js
```diff
@@ -36,352 +35,354 @@ goog.require('Blockly.constants');
...
-    this.targetWorkspace.recordDragTargets();
   }
-};
+}
+
+/**
+ * The name of the vertical flyout in the registry.
+ * @type {string}
+ */
+VerticalFlyout.registryName = 'verticalFlyout';
```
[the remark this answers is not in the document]
rachel-fenichel: The advanced compilation test failed on static properties, saying that it needs to be compiling from ESNEXT instead of ES2020. We're currently using the highest stable version for the closure compiler, so I'm going to leave it where it is.But I think that the ts compiler will sort this out.

rachel-fenichel (commented):
I ended up leaving static properties alone instead of moving the declaration into the class declaration. I tried to clean up all of my comments about it, but GitHub is playing hide and seek with my comments so I'm not sure I succeeded.

core/flyout_vertical.js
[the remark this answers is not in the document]
cpcallen: OK, fair enough.  The current version of closure-compiler does accept ES2021, but I don't think that includes static initialisers.

cpcallen (approved):

PR #5889, opened 2022-01-19
refactor!: remove unused constants

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
Part of #5158.

Specifically the three below parts of the issue: 

- All constants should be moved out of this file and to the file(s) containing the code to which they relate.
- Where they are only used in a single module, they should become unexported, module-local constants in that file.
- Consideration should be given as to whether they could be renamed to improve readability.

### Proposed Changes
This moves constants that were in internal_constants and were only used in a single file to be module-local constants in that file as described by the issue referenced above. 

In order to decide that these were not commonly used local constants I: 
- Checked our major partners (code.org, makecode, scratch, appInventor, blockly-games, blockly-samples) repositories to see if they were using or changing this value.
- I searched all of GitHub for these values and the word Blockly and then skimmed a few dozen repositories to see if they were being used/changed outside of the fork.

I will note that I was using GitHub's search feature which is mediocre at best.

#### Behavior After Change
The following values are no longer accessible on the Blockly object: 
- LINE_MODE_MULTIPLIER
- PAGE_MODE_MULTIPLIER
- BUMP_RANDOMNESS
- LONGPRESS
- SOUND_LIMIT

### Reason for Changes
Internal constants is a temporary file that was created with the move to goog.module. We want to place all of the constants in there in more appropriate files.

AFAICT these are not being used outside of Blockly and should in fact be internal constants.

### Additional Information
Just to note this is not changing whether these values can be set, only whether they should be read.

If people are uncomfortable with removing these from the public API, I am also fine with exporting these from the modules and placing getters with deprecation warnings on the Blockly object.


cpcallen:
> In order to decide that these were not commonly used local constants I:
> 
> * Checked our major partners (code.org, makecode, scratch, appInventor, blockly-games, blockly-samples) repositories to see if they were using or changing this value.
> * I searched all of GitHub for these values and the word Blockly and then skimmed a few dozen repositories to see if they were being used/changed outside of the fork.

That is impressively thorough!

cpcallen (approved):

PR #5890, opened 2022-01-20
fix: TypeScript exporting of the serialization functions

AndrewWyer-Dover:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/5891 Attempts to use `Blockly.serialization` fail within TypeScript environments as serialization is not exported within the type definition file.

### Proposed Changes

Cleans up the serialization type declarations and bundles them into the main Blockly export so that they are available to TypeScript users, in the same manner as Xml.

### Reason for Changes

Two of the serialization module declarations were distant from the others; moving them to be adjacent improves code maintainability.
The existing convention appears to use bundling and re-exporting, so a `serialization` module was introduced.
The new singular module is imported to blockly and then re-exported.

Note: I did **not** change the capitalization of the existing code, however it appears to be in conflict with precedent. I am unsure of policy / practice of this project in this area.

### Test Coverage

Tested on:
* Windows 10 64bit, using Visual Studio Code v1.63.2, using TypeScript v4.4.3

### Documentation

This change does not trigger documentation updates.

### DIFF
--- a/typings/blockly.d.ts
+++ b/typings/blockly.d.ts
@@ -1084,6 +1084,15 @@ declare module "serialization/workspaces" {
     } | undefined): void;
     import { Workspace } from "workspace";
 }
+declare module "serialization" {
+    import * as blocks from "serialization/blocks";
+    import * as exceptions from "serialization/exceptions";
+    import * as priorities from "serialization/priorities";
+    import * as registry from "serialization/registry";
+    import * as variables from "serialization/variables";
+    import * as workspaces from "serialization/workspaces";
+    export {blocks, exceptions, priorities, registry, variables, workspaces};
+}
 declare module "utils/size" {
     /**
      * @license
@@ -25545,6 +25554,7 @@ declare module "blockly" {
     import { RenderedConnection } from "rendered_connection";
     import { Scrollbar } from "scrollbar";
     import { ScrollbarPair } from "scrollbar_pair";
+    import * as serialization from "serialization";
     import * as ShortcutItems from "shortcut_items";
     import { ShortcutRegistry } from "shortcut_registry";
     import { TabNavigateCursor } from "keyboard_nav/tab_navigate_cursor";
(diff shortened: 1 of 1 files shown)


typings/blockly.d.ts
```diff
@@ -1036,6 +1036,63 @@ declare module "serialization/exceptions" {
...
+    /**
+     * Represents the state of a given variable.
+     * @typedef {{
+     *   name: string,
+     *   id: string,
+     *   type: (string|undefined)
+     * }}
+     * @alias Blockly.serialization.variables.State
+     */
+    export let State: any;
```
BeksOmega, line 1058: I don't think this let is actually necessary, but probably best to leave as is. Iirc it was just to make lint & closure happy, but seems to have ended up here.

BeksOmega (approved):
Yay thank you for the fix! LGTM :D But I'm no expert on typescript.

As for the capitalization, I believe this is correct. Most blockly exports are classes, which should be capitalized. But the serialization system just has namespaces for functions, which aren't capitalized.

Thank you again! Someone else with more typescript expertice should probably also give it a look. But I just wanted to tell you that I really appreciate you looking into this =)

typings/blockly.d.ts
AndrewWyer-Dover: I copy / pasted the block without editing. I agree it may not be necessary but wanted to keep my scope of changes limited.

cpcallen:
>> As for the capitalization, I believe this is correct. Most blockly exports are classes, which should be capitalized. But the serialization system just has namespaces for functions, which aren't capitalized.
> 
> As for the capitalization, I believe this is correct. Most blockly exports are classes, which should be capitalized. But the serialization system just has namespaces for functions, which aren't capitalized.

Beka is correct.  In fact, Blockly historically used UpperCamelCase for almost _all_ its namespace objects, not just the ones which were classes.  neilfraser@ might be able to shed some light on this (maybe the styleguide encouraged it a the time?)  New code should follow the current style guide (using lowerCamelCase unless the module exists primarily to provide a class of the same name).  At the moment, for backward-compatibility reasons we don't want to change the names of the properties on the `Blockly` object, and although we could now rename the `goog.module` names (since the `Blockly` property name is now independent of the module name) there's no good reason to as we will most probably soon move to ES (or TS) modules, which are loaded by path rather than module name.

cpcallen (approved):
Looks good to me!

Sorry for the delay in responding; I wasn't sure if we should accept changes to the `.d.ts` files directly, or try to fix the tooling that generates them.  But at the moment that tooling is a bit of a [Heath Robinson](https://en.wikipedia.org/wiki/W._Heath_Robinson) contraption and the consensus of the team is that manual fixes like this are helpful—so thanks!

AndrewWyer-Dover:
> Sorry for the delay in responding; I wasn't sure if we should accept changes to the `.d.ts` files directly, or try to fix the tooling that generates them. But at the moment that tooling is a bit of a [Heath Robinson](https://en.wikipedia.org/wiki/W._Heath_Robinson) contraption and the consensus of the team is that manual fixes like this are helpful—so thanks!

No worries; I had the same concern when I read the readme, but none of the items from that seemed to be still applicable (e.g., the script commands or referenced files).

Issue #5892, opened 2022-01-21
Set up typescript compilation


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**

Part of #5857. 

**Describe the solution you'd like**

- [ ]  Add a tsconfig.json.
- [ ]  Wire the typescript compiler to output into the build directory such that the closure compiler picks up the correct files.
- [ ]  Add gulp tasks and scripts in package.json to run the ts compiler and then the closure compiler.
- [ ]  Add gulp tasks and scripts in package.json to clean up generated files when needed.

cpcallen:
I think that this being marked "internal" is preventing me from adding as the linked issue for PR #6220.

cpcallen:
> I think that this being marked "internal" is preventing me from adding as the linked issue for PR #6220.

Maybe that is true but I have less confidence in this assertion, as marking the PR also as internal does not solve the problem.

cpcallen:
This is now basically complete, but I will leave open in case we decide that we do in fact want to do some post-`tsc` cleanup of the generated `.js` files (in particular, for file extensions on the imports).

PR #5894, opened 2022-01-21
chore(build): add a script to run the ts compiler and send the new files to closure compiler

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/5892

### Proposed Changes

- Adds a `tsconfig.json` telling the typescript compiler a few critical things: where the input code is, where the output code goes, and that it should read existing js files. 
- Adds a gulp task that runs the typescript compiler
- Adds a flag, `compileTs`, that controls where the build tasks look for source files. By default build uses the files in `core`. When `compileTs` is set, it uses the files in `build/ts/core`.
- Adds a script in `package.json` that runs the ts compiler and then runs build with the appropriate flag set.

Note that the `compileTs` flag is a hack designed to let me set up compilation without properly passing parameters through the build tasks. 
The default behaviour is currently to just run the closure compiler on the code in `core`. Ideally by the end of the quarter the default behaviour will be to run the closure compiler on the code generated by the typescript compiler, and this option will go away entirely. If possible I want to avoid having multiple compilation setups (with different source code) at the end of the quarter.

I did chat with Chris about this and we agreed it's a reasonable hack to get compilation going.

### Test Coverage
Tested by running locally and confirming that `deps` gets updated, and that I can open the playground and use the code (uncompiled) and open a demo and use the code (compiled).

### Additional Information

The tsconfig here does not enable generation of declarations (`.d.ts` files) because there are currently 85 errors that block declaration export.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -230,7 +230,7 @@ function buildDeps(done) {
       'node_modules/google-closure-library/closure/goog' :
       'closure/goog';
 
-  const coreDir = argv.compileTs ? TSC_OUTPUT_DIR + '/core' : 'core';
+  const coreDir = argv.compileTs ? path.join(TSC_OUTPUT_DIR, 'core') : 'core';
   const roots = [
     closurePath,
     coreDir,
@@ -358,7 +358,7 @@ return ${NAMESPACE_OBJECT}.${chunk.exports};
  */
 function getChunkOptions() {
   if (argv.compileTs) {
-    chunks[0].entry = TSC_OUTPUT_DIR + '/core/blockly.js';
+    chunks[0].entry = path.join(TSC_OUTPUT_DIR, chunks[0].entry);
   }
   const cccArgs = [
     '--closure-library-base-js-path ./closure/goog/base_minimal.js',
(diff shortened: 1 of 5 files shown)


scripts/gulpfiles/build_tasks.js
```diff
@@ -457,10 +467,19 @@ const pathSepRegExp = new RegExp(path.sep.replace(/\\/, '\\\\'), "g");
...
+      pathObject.dirname = 'build/ts/core';
+      pathObject.basename =
+          dirs.slice(3).concat(pathObject.basename).join('-slash-');
+    }
+  } else {
+    if (dirs[0] === 'core') {
+      pathObject.dirname = dirs[0];
+      pathObject.basename =
+          dirs.slice(1).concat(pathObject.basename).join('-slash-');
+    }
```
cpcallen, line 482: A shorter, more general version:
```suggestion
  const coreIndex = argv.compileTs ? 2 : 0;
  if (dirs[coreIndex] === 'core') {
    pathObject.dirname = path.join.apply(...dirs.slice(0, coreIndex + 1));
    pathObject.basename =
        dirs.slice(coreIndex + 1).concat(pathObject.basename).join('-slash-');
  }
```
Unfortunately neither version is totally safe (e.g. in the face of a hypothetical `blocks/x/core/` directory) but in practice this probably go away before any such directory gets created.

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    //"declaration": true,
+    // Types should go into this directory.
+    // Removing this would place the .d.ts files
+    // next to the .js files
+    //"declarationDir": "build/ts/declarations",
+
+    "outDir": "build/ts",
+    "module": "ES2015",
+    "moduleResolution": "node",
+    "target": "ES2015",
```
cpcallen, line 22: This should probably match the `language_in` specified for closure compiler:
```suggestion
    "target": "ES2020",
```
(`module` could also be updated to ES2020, which would give us dynamic `import`s, which might be useful for optional modules: auto-load features upon request!)

cpcallen (approved):
I've offered a few suggestions, but no objections to it going in as-is if you don't like them.

rachel-fenichel:
Github now lets me accept multiple suggested changes and commit them as a batch, which is super nice. I accepted all of yours.

Issue #5875, opened 2022-01-13
Proposal: make block modules side-effect free


cpcallen:
## Background

At the moment, the blocks modules (`Blockly.blocks.*`, in `blocks/`) generally do not have any exports; instead, importing one causes the created blocks definitions to be loaded into the block definitions dictionary (`Blockly.Blocks`, or more specifically the `Blocks` export of the `Blockly.blocks` module) as a side effect.  Various extensions are also registered (again as a side-effect) via the `register*` exports of the `Blockly.Extensions` module.

Additionally, library block modules are normally imported via a single `import 'blockly/blocks' statement, which imports the `Blockly.blocks.all` module, which in turn imports the rest of the individual `Blockly.blocks.*` modules.

This creates some problems for developers:

* Loading blocks via `import 'blockly/blocks'` is all-or-nothing: only developers using the Closure module system can use `goog.require` to load only a specific module (e.g. `Blockly.blocks.loops`).
* There is no way to load only certain blocks from an individual module.  At best you must load them all and then delete unwanted ones from `Blockly.Blocks` (as well as unregister any unwanted extensions).
* There is at present no way for developers who use ES Module `import` statements to access exports from individual block module (e.g., the `loopTypes` export from `Blockly.blocks.loops`)—see #5819.

## Proposed solution

Refactor the block modules in several steps, as follows:

1. Add a standard set of `exports` to the block modules:

   * `blocks`: a dictionary object (`@type {!Object<string, !Object>}`), keyed by block type, of the block definitions that that module creates.
   * `extensions`: a dictionary object (`@type {!Object<string, !Object>}`) keyed by extension name, of the extensions registered by the module.
   * Any other module-specific exports (e.g., `loopTypes`).

   The `blockly.blocks.all` module will provide `blocks` and `extensions` dictionaries that include all the entries form all of the individual block modules.  Additionally, it will re-export the exports objects from each of those modules using the corresponding names (e.g. `colour`, `lists`, `loops`, etc.)

   This will enable importers to see the full set of blocks (and extensions) that are loaded, as well as access any other exports provided by individual block modules (e.g., `loops.loopTypes`).

   This will be a breaking change to the exports provided by importing `blockly/blocks`:
   * At present `blockly/blocks` provides a default export of the whole `Blockly.Blocks` dictionary.
[...]

PR #5908, opened 2022-01-31
feature(blocks): Export block definitions

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #5875.

### Proposed Changes

Have each `Blockly.blocks.*` module export a `blocks` dictionary of the blocks defined in that module—including `Blockly.blocks.all`, which will export a `blocks` dictionary containing all the blocks defined in the `blocks/` directory, but not any developer-defined blocks added to `Blockly.Blocks` directly.

#### Behavior Before Change

* The `defineBlocksWithJsonArray` export from `Blockly.common` would convert a (possibly JSON-derived) POJsO array into block definitions (typed `!Object`) and then add them to the `Blocks` dictionary immediatley.

* `Blockly.blocks.*` modules create block definitions in the `Blocks` dictionary defined in `Blockly.blocks`, but do not export any block definitions.

#### Behavior After Change

* `defineBlocksWithJsonArray` is now a wrapper around two new API functions: `createBlockDefinitionsFromJsonArray`, which creates the `BlockDefinition` objects, and `defineBlocks` which adds them to the `Blocks` dictionary.  Either can be used on their own.

* `Blockly.blocks.*` modules create `BlockDefinitions` in a `blocks` export, which are then added to the the `Blocks` dictionary.

* `Blockly.blocks.all` creates a `blocks` export combing all the `blocks` exports of the other individual modules.

### Reason for Changes

This is a substantial step towards making it possible for block definition modules to be side-effect free.  In particular, now it is only the single call to `defineBlocks` at the end of each block module that touches the `Blocks` dictionary.

### Test Coverage

Passes `npm test` and a cursory check with Desktop Chrome.

### Documentation

We should probably mention the `blocks` exports in our developer documentation.

We may wish to mention `Blockly.common.createBlockDefinitionsFromJsonArray` and `Blockly.common.defineBlocks`.

### Additional Information

This PR only covers the `blocks` export mention in #5875; a separate PR will be needed to address extensions.

Ought we to reexport `createBlockDefinitionsFromJsonArray` and `defineBlocks` directly on the `Blockly` object, as we do for `defineBlocksFromJsonArray`?  I have elected not to for now.


core/blocks.js
```diff
@@ -16,11 +16,18 @@
 goog.module('Blockly.blocks');
 
 
+/**
+ * A block definition.  For now this very lose, but it can potentially
```
rachel-fenichel, line 20: this very lose -> this is very loose

rachel-fenichel (approved):

core/blocks.js
cpcallen: Oh bother.  GitHub didn't update the page to show the comment, only the approval.  🙄

Issue #5904, opened 2022-01-29
Convert to ES Modules


rachel-fenichel:
Closure compiler documentation:
- [ES6 modules and Closure interop](https://github.com/google/closure-compiler/wiki/ES6-modules-and-Closure-interop)
- [Migrating from `goog.modules` to ES6 modules](https://github.com/google/closure-compiler/wiki/Migrating-from-goog.modules-to-ES6-modules)

The specific steps that we need to take, pulled from the closure documentation:
**Migrating Closure modules with named exports**
For Closure modules that use named exports we recommend migrating files in place over 3 changes.

1. Edit the implementation file to be an ES module that calls `goog.declareModuleId` with the old Closure module name (i.e. the old argument to `goog.module`). This should allow other files to continue using `goog.require` on the Closure namespace. (Be sure to import Closure's `goog.js` file rather than use the global `goog` symbol!)
3. Edit all ES module files that are using `goog.require` on the module namespace to instead use ES `import`.
4. Remove the call to `goog.declareModuleId` in the migrated file once it is not being referenced by any `goog.module` or `goog.provide` files.

**Reasons to migrate**
- (As I understand it) the Typescript compiler cannot pick up and use the correct types for imports in the form of `const {Foo} = goog.require('Blockly.Foo')`.
- Tsickle assumes that the input files are ES modules.
- Tsickle relies on correct parsing by the typescript compiler, and chokes when the compiler chokes.
- The closure compiler understands ES modules, so *in theory* this change can be done without involving the typescript compiler--meeting my goal of being able to compiler with the closure compiler up until we've fully debugged our typescript compilation flow.

cpcallen:
Title tweaked: although they were introduced in ES6, they are more widely referred to simply as "ES modules" rather than "ES6 modules".

cpcallen:
> 1. (Be sure to import Closure's `goog.js` file rather than use the global `goog` symbol!)

Have you verified that this is necessary?

rachel-fenichel:
We may be able to skip directly to TypeScript, in which case we can close this entirely. @gonfunko is currently investigating this.

cpcallen:
This was done (for `core/`) as part of the TypeScript conversion.  Will track conversion of `blocks/` and `generators/` along with their conversion to TypeScript.

PR #5918, opened 2022-02-07
refactor(build): Stop flattening Blockly sources during compilation in preparation for move to ES modules

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Preparation for #5904

### Proposed Changes
This change disables the hoisting/flattening of Blockly sources files into a single directory during compilation and disables enforcement of `@package` (and `@private` as a side effect). This flattening breaks imports by relative paths (to be used once we convert to ES modules), so it needs to removed for now.


scripts/gulpfiles/build_tasks.js
```diff
@@ -173,7 +173,7 @@ function stripApacheLicense() {
  * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DiagnosticGroups.java#L113
  */
 var JSCOMP_ERROR = [
-  'accessControls',
+  // 'accessControls',
```
cpcallen, line 176: It turns out that this is now just a synonym for `visibility`, so it can just be deleted.

scripts/gulpfiles/build_tasks.js
```diff
@@ -216,7 +216,7 @@ var JSCOMP_ERROR = [
   'unusedPrivateMembers',
   'uselessCode',
   'untranspilableFeatures',
-  'visibility'
+  // 'visibility'
```
cpcallen, line 219: I think it would be good to add a comment about why this is commented out.  (Perhaps refer to the jsdoc for `flattenCorePaths` for background, and then just mention about temporarily disabling checking for `@package` due to ES by-path imports.)

scripts/gulpfiles/build_tasks.js
```diff
@@ -568,13 +568,14 @@ function buildAdvancedCompilationTest() {
     compilation_level: 'ADVANCED_OPTIMIZATIONS',
     entry_point: './tests/compile/main.js',
     js_output_file: 'main_compressed.js',
+    jscomp_warning: ['accessControls', 'visibility'],
```
cpcallen, line 571: As above: `accessControls` can be deleted.

scripts/gulpfiles/build_tasks.js
gonfunko: Done!

PR #5929, opened 2022-02-11
refactor!: moves the debug renderer out of core

### LINKED ISSUE #4307: Strip out the rendering debugger as it is now in @blockly/dev-tools

Clean up code in Blockly code referencing the rendering debugger as it is now in @blockly/dev-tools

**Additional context**

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#4307

### Proposed Changes
Strips out the debug rendering code since we are moving it completely out of core and in to @blockly/dev-tools.

#### Behavior Before Change
You could enable the debug renderer by calling the below: 
```
Blockly.blockRendering.startDebug();
```

#### Behavior After Change
The debug renderer is now available in [@blockly/dev-tools](https://www.npmjs.com/package/@blockly/dev-tools) and you can use it in the advanced playground.

### Reason for Changes
The debug renderer was placed in core before we had blockly-samples. Now that we have a better place for it to go, it should be included in [dev-tools](https://www.npmjs.com/package/@blockly/dev-tools).


core/renderers/common/block_rendering.js
```diff
@@ -51,19 +48,6 @@ const {Theme} = goog.requireType('Blockly.Theme');
...
-/**
- * Returns whether the debugger is turned on.
- * @return {boolean} Whether the debugger is turned on.
- * @alias Blockly.blockRendering.isDebuggerEnabled
- * @package
- */
-const isDebuggerEnabled = function() {
-  deprecation.warn(
-      'Blockly.blockRendering.isDebuggerEnabled()', 'September 2021',
-      'September 2022', 'Blockly.blockRendering.debug.isDebuggerEnabled()');
```
gonfunko, line 63: Are we OK removing this before the promised September 2022?

gonfunko (approved):

core/renderers/common/block_rendering.js
rachel-fenichel: I'm okay with removing it and changing these warnings to say where it went, then removing the warnings entirely after september 2022.

alschmiedt: Ok, I updated these messages as well as added back in `Blockly.blockRendering.debug` since those are what we recommended people switch to. Also just wanted to note that all these methods are technically @package and have been since they were created.

alschmiedt:
Ok I have been thinking about this PR and I think it is better to just leave in for right now. Leaving it half in is a bit of a pain, and we don't actually have a reason to take it out early other than me being impatient. So I am planning on leaving it in, keeping the issue to remove it open, and updating the deprecation warning to say that it is being completely removed from core in September.

Issue #5936, opened 2022-02-14
Comments are not added when a block is defined from a JSON array 


rhumbertgz:
**Describe the bug**

The function Block.prototype.jsonInit doesn't check for the key "commentText" as consequence comments added during the definition of a block from a JSON array are not added.

**To Reproduce**

The following code defines a dummy block with a comment

```
Blockly.defineBlocksWithJsonArray([
{
"type": "comment_block_2",
"message0": "%1",
"args0": [
{
"type": "field_input",
"name": "NAME",
"text": "default"
}
],
"colour": 230,
"commentText": "Comment demo", // Comment is NOT visible in the block
}
])
```

**Expected behavior**

![image](https://user-images.githubusercontent.com/10393245/153859205-083bb35c-20da-4498-b879-1e52c781e0ba.png)

**Screenshots**

The comment component of the blocks is not added, however declaring the block using JS solve the problem.

![image](https://user-images.githubusercontent.com/10393245/153859422-92f7f124-e449-41b2-97e8-a10f6eb68258.png)

**Desktop (please complete the following information):**
 - OS: MacOS 12.1
 - Browser Chrome 
 - Version 98.0.4758.80

gonfunko:
In talking with the team we're going to leave this as-is; the JSON and JS block definition approaches are not 1:1, and in some cases more advanced features like this are only available through JS. As an alternative, you may find that it works better to associate the comments with the block(s) in your toolbox definition; there, you can specify `icons: {comment: {text: "comment text goes here"}}` as part of specifying the available blocks, and since folks typically create blocks from the toolbox that would have the effect of causing all new instances dragged out to have the comment.

PR #5937, opened 2022-02-14
Add block's comment from JSON definition

rhumbertgz:
## The basics

- [ x] My pull request is against develop

### Resolves

Resolves issue [#5936](https://github.com/google/blockly/issues/5936)

### Proposed Changes

This pull requests checks the presence of the "commentText" key in a JSON Block definition and if such key exists add the comment to the block.

```
if (json['commentText'] !== undefined) {
    this.setCommentText(json['commentText']);
}
```

#### Behavior Before Change

 Comment is not added 
image
                                                                                                                                       

#### Behavior After Change

image

### Reason for Changes

Current implementation doesn't check the key "commentText" in the Json definition of a block

### Test Coverage

 * Desktop Chrome 
 * Desktop Firefox 
 * Desktop Safari
 * Desktop Opera

### Documentation

Not needed

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1113,6 +1113,9 @@ Blockly.Block.prototype.jsonInit = function(json) {
     var localizedValue = Blockly.utils.replaceMessageReferences(rawValue);
     this.setHelpUrl(localizedValue);
   }
+  if (json['commentText'] !== undefined) {
+    this.setCommentText(json['commentText']);
+  }
   if (goog.isString(json['extensions'])) {
     console.warn('JSON attribute \'extensions\' should be an array of ' +
       'strings. Found raw string in JSON for \'' + json['type'] + '\' block.');


BeksOmega (changes requested):
Hello @rhumbertgz Thank you so much for your PR and associated issue :D I think this is a lovely idea. Just two things that need to be resolved before this goes in:

1) Have this been tested with large amounts of text? I'm imagining it might get cut off because the size of the comment bubble isn't being modified. If it does get cut off, how would you feel about adding optional `commentWidth` and `commentHeight` properties as well? These would be a bit trickier because they might interact with headless mode (I'd have to check).
2) Could you add a call to `Blockly.utils.replaceMessageReferences` as in the above sections? This makes sure that the comment text can be localized if people want to do so.

Best wishes and thank you again for submitting this :D

[Edit: Also, would it be possible for you to side the CLA? Sadly your changes can't go in until we have permission in "writing"]

BeksOmega:
Hello! I'm going to go ahead and close this due to it being stale. But if you still want to work on this feel free to reopen/open a new pull request =)

PR #5940, opened 2022-02-14
refactor: convert registry files to es6 classes

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/5860

### Proposed Changes

- Change the shortcut registry and context menu registry to have internal `registry` variables that are exported under appropriate names.
- Add `reset` functions to the shortcut registry and context menu registry.
- Update tests as needed.
- Change the `Type` object in `registry.js` to an es6 class.

#### Behavior Before Change

Clearing the shortcut registry or context menu registry in tests was done by calling the constructor. 

#### Behavior After Change

Now clearing those registries is done by calling `reset`.

### Reason for Changes

Movement toward Typescript.

### Additional Information

Typescript warns that `keyMap_` and `registry_` and so on can be undefined, because I'm setting them in the `reset` function instead of directly in the constructor. Typescript has the [definite assignment assertion operator](https://www.typescriptlang.org/docs/handbook/2/classes.html#--strictpropertyinitialization) to deal with this.

### DIFF
--- a/core/contextmenu_registry.js
+++ b/core/contextmenu_registry.js
@@ -25,21 +25,101 @@ const {WorkspaceSvg} = goog.requireType('Blockly.WorkspaceSvg');
  * Class for the registry of context menu items. This is intended to be a
  * singleton. You should not create a new instance, and only access this class
  * from ContextMenuRegistry.registry.
- * @constructor
- * @private
- * @alias Blockly.ContextMenuRegistry
  */
-const ContextMenuRegistry = function() {
-  // Singleton instance should be registered once.
-  ContextMenuRegistry.registry = this;
+class ContextMenuRegistry {
+  /**
+   * @alias Blockly.ContextMenuRegistry
+   */
+  constructor() {
+    this.reset();
+  }
 
   /**
-   * Registry of all registered RegistryItems, keyed by ID.
-   * @type {!Object<string, !ContextMenuRegistry.RegistryItem>}
-   * @private
+   * Clear and recreate the registry.
    */
-  this.registry_ = Object.create(null);
-};
+  reset() {
+    /**
+     * Registry of all registered RegistryItems, keyed by ID.
+     * @type {!Object<string, !ContextMenuRegistry.RegistryItem>}
+     * @private
+     */
+    this.registry_ = Object.create(null);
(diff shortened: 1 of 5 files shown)


core/registry.js
```diff
@@ -73,28 +73,31 @@ exports.DEFAULT = DEFAULT;
 
 /**
  * A name with the type of the element stored in the generic.
- * @param {string} name The name of the registry type.
- * @constructor
  * @template T
  * @alias Blockly.registry.Type
```
maribethb, line 77: For the other classes you've moved the `@alias` down to the constructor as well. Not sure which one makes the JsDoc tool happier but I had been assuming it was on the constructor.

core/contextmenu_registry.js
```diff
@@ -90,85 +170,9 @@ ContextMenuRegistry.ContextMenuOption;
...
-    return a.weight - b.weight;
-  });
-  return menuOptions;
-};
-
-// Creates and assigns the singleton instance.
-new ContextMenuRegistry();
+ContextMenuRegistry.registry = new ContextMenuRegistry();
 
+exports.contextMenuRegistry = ContextMenuRegistry.registry;
```
maribethb, line 177: Why are we exporting this (and for shortcut registry) now? Is it just as a shorthand for `ContextMenuRegistry.registry`? I don't like introducing API/naming changes (also needs documentation changes if we intend developers to use this) buried in what seems like a refactoring change.

I think this is a good idea, it would match the fieldRegistry and class registry patterns, but also needs a few other updates like changing the internal usages and the jsdoc for the class explaining how to use the registry.

maribethb (approved):

core/contextmenu_registry.js
rachel-fenichel: Good point. I've backed it out and filed https://github.com/google/blockly/issues/5950 to track and/or discuss.

core/registry.js
rachel-fenichel: Done, for consistency. I'm not sure and slightly expect I'll have to do a cleanup pass when I get to API extractor; so far I've been ignoring the JsDoc tool.

Issue #5932, opened 2022-02-12
Latest patch (Q4 2021 Patch 3) fails to load


KlemenDEV:
**Describe the bug**

The latest patch (Q4 2021 Patch 3) fails to load

**To Reproduce**

Try to load Blockly as specified with examples.

**Resolution**

This seems to be fixed by changing line from https://github.com/google/blockly/blob/caba5966b90422d3068b4c154cc0203e33de0ae5/blocks_compressed.js#L10 

to

`root.Blockly.blocks = factory(root.Blockly);`

Taikono-Himazin:
Thanks to this, it works.
thank you.

alschmiedt:
Thanks for reporting! We are currently working on a fix for this, but in the meantime we have reverted the patch so version 7.20211209.4 should work as expected.

cpcallen:
O.K., I came up with a fix for this issue yesterday—for which I will submit a PR shortly—but I realised I was confused about why it should even have been necessary so in the mean time I advised @alschmiedt to submit #5938 to revert the release until I could be sure that my fix was _actually_ correct.

Here's the deal.  As noted in this (and other) bug reports, the error is indeed caused by the line
```JS
root.Blockly.blocks.all = factory(root.Blockly);
```
in `blocks_compressed.js`.

My first take was: of course this will fail, because while `blockly_compressed.js`, when loaded in a browser, creates a `root.Blockly` object that object won't have a `.blocks` property onto which could be tacked an `.all`.  I prepared a fix based on this.

But then, while verifying my fix, I noticed / remembered that, because of how `.declareLegacyNamespace` works (specifically, in this case, how it is implemented by Closure Compiler), the `factory` function _actually creates a `.blocks` property_ and initialises it with an empty object _and then additionally creates the `.blocks.all` property and initialises that too_!  So `root.Blockly.blocks.all = factory(...)` should in fact be doing nothing at all.  (If you look at the contents of `blocks_compressed.js`, note that `root.Blockly` will have previously been set to `$.Blocky` and `factory` returns `$.Blockly.blocks.all`.)

Unfortunately, while `root.Blockly.blocks.all = factory(root.Blockly)` ought to do nothing, it is actually throwing `TypeError`, _because `root.Blockly.blocks` is evaluated before the call to `factory` which would create the `.blocks` property_, and thus evaluates to `null`, and attempting to assign to `(null).all` is indeed a TypeError.

Here ends today's lesson on the order of evaluation of JavaScript expressions.

[...]

PR #5946, opened 2022-02-15
refactor!(blocks): Rename module `Blockly.blocks.all` to `Blockly.blocks`; former `Blockly.blocks` back to `Blockly.Blocks`

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

* Rename `Blockly.blocks` (`core/blocks.js`) back to `Blockly.Blocks` (as it was prior to PR #5515).
* Rename `Blockly.blocks.all` (`blocks/all.js`) to `Blockly.blocks` (`blocks/blocks.js`).

#### Behaviour Before Change

* Code requiring the blocks dictionary object does `const {Blocks} = goog.require('Blockly.blocks');`,
  * or uses a `script` tag to load `blockly_compressed.js` and then access the dictionary via `Blockly.Blocks`.
* Code loading library blocks does `goog.require('Blockly.blocks.all');`,
  * or uses a `<script>` tag to load `blocks_compressed.js` and then accesses exports via `Blockly.blocks.all`.

#### Behaviour After Change

* Code requiring the blocks dictionary object does <code>const {Blocks} = goog.require('Blockly.**B**locks');</code>
  * or uses a `script` tag to load `blockly_compressed.js` and then access the dictionary via `Blockly.Blocks` (unchanged).
* Code loading library blocks does `goog.require('Blockly.blocks');`,
  * or uses a `<script>` tag to load `blocks_compressed.js` and then accesses exports via <code>Blockly.**b**locks</code>.

There will be no changes visible when the blocks chunk is loaded via ES module `import` or CJS `require`.

### Reason for Changes

This change was originally created as an alternative fix for issue #5932.  PR #5945 is a better fix for the actual underlying cause of that issue, but I think it is worth considering doing these renamings anyway simply because of providing a more sensible path for the blocks modules exports object for those loading the compiled chunks directly.

### Test Coverage

Tested on:
* Desktop Chrome
* `npm test`

### Additional Information

Having the`core/blocks.js` module named `Blockly.Blocks` is less in accord with the styleguide, but more consistent with the (mis)capitalisation of its main export, the `.Blocks` dictionary object.

The need for naming of modules will shortly go away entirely when we convert from `goog.module` to ES modules.


cpcallen:
We could also consider renaming all of the `blocks/` modules to `Blockly.libraryBlocks.*`, or to `BlocklyBlocks.*` or even just  `blocks.*`, but in chat this option seemed to be slightly preferred.

rachel-fenichel:
After discussion with Chris:
- Rename to `Blockly.libraryBlocks` to make the module name make more sense when referred to directly.
- Find the right place to put documentation about how to access this through requires, through script tags, and through modules, since the correct code varies by context.

PR #5947, opened 2022-02-16
refactor: convert toolbox items to ES6 classes

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5860

### Proposed Changes

Converts toolbox item, category, and collapsible category to ES6 classes. ~~This involves adding a SENTINEL value very similar to the value used in fields so that we can make sure the toolbox category contents aren't parsed until all of the properties of the category instance have been defined.~~ This involves moving the parsing of category contents into the `init` method to ensure that all properties of the category instance have been defined before we try to modify them.

Should be no change in behavior *fingers crossed*.

### Reason for Changes

Moving to ES6 classes.

### Test Coverage

All unit tests pass.

### Documentation

N/A

### Additional Information

N/A


BeksOmega:
@maribethb Just wanted to give you a heads up, I requested your review because you're the most recent person to use this API, and I wanted to make sure I didn't mess things up for external peeps subclassing the categories.

scripts/gulpfiles/chunks.json
```diff
@@ -253,7 +253,6 @@
     "./core/extensions.js",
     "./core/block.js",
     "./core/utils/string.js",
-    "./core/utils/object.js",
```
maribethb, line 256: what's up with this change? is this the last file that is using object.inherits?

maribethb (commented):
Your PR title should say `refactor` instead of `refact`. I feel like the conventional commit linter should have caught this... I don't know why it didn't since `refact` is not in the approved list https://www.npmjs.com/package/@commitlint/config-conventional but  ¯\\\_(ツ)_/¯

core/toolbox/category.js
```diff
@@ -129,6 +126,8 @@ class ToolboxCategory extends ToolboxItem {
      */
     this.flyoutItems_ = [];
 
+    if (categoryDef === ToolboxCategory.SENTINEL) return;
```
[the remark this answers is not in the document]
alschmiedt: I think my preference is to call `parseContents_` in the init method. I think someone not calling init after creating a toolbox item is probably rare and goes against what we would generally expect. The toolbox item would already not work correctly if they do not call init, so the fact that it would break further if they don't call init seems ok to me. Definitely still a breaking change, but I think an acceptable one.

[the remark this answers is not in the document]
BeksOmega: Coolio. I'm going to attempt to implement calling `parseContents_` from `init`, because I agree that atm that seems like the best option.

scripts/gulpfiles/chunks.json
BeksOmega: No clue, something removed it automatically. But if you expand this code block down, it looks like it still exists [link](https://github.com/google/blockly/pull/5947/files#diff-75165818332bb6d85e32df0a855c805ce6c0fe94790fd66067719077b9f2c810L262). So I don't think I need to add it back in?

BeksOmega:
Should be ready for another look!

core/toolbox/collapsible_category.js
```diff
@@ -32,277 +31,277 @@ const {ToolboxSeparator} = goog.require('Blockly.ToolboxSeparator');
 
 /**
  * Class for a category in a toolbox that can be collapsed.
- * @param {!toolbox.CategoryInfo} categoryDef The information needed
- *     to create a category in the toolbox.
- * @param {!IToolbox} toolbox The parent toolbox for the category.
- * @param {ICollapsibleToolboxItem=} opt_parent The parent category or null if
- *     the category does not have a parent.
- * @constructor
  * @extends {ToolboxCategory}
```
maribethb, line 34: You don't need this extends either (and the one above still needs to be fixed as well)

core/toolbox/category.js
```diff
@@ -34,612 +34,628 @@ const {ToolboxItem} = goog.require('Blockly.ToolboxItem');
...
-  this.htmlDiv_ = null;
+  parseCategoryDef_(categoryDef) {
+    this.name_ = parsing.replaceMessageReferences(categoryDef['name']);
+    this.colour_ = this.getColour_(categoryDef);
+    object.mixin(
+        this.cssConfig_, categoryDef['cssconfig'] || categoryDef['cssConfig']);
+  }
+
+  /**
+   * @override
```
maribethb, line 189: Could you add a comment to this method saying it must be called immediately after creation? The super class's `init` is a no-op so it's good to be explicit here. Also I would suggest moving this up to be the first method after the constructor so it's easier to notice.

maribethb (approved):
Looks good although there were a few comments from round 1 that weren't addressed yet (make sure you update the PR title so that this gets categorized correctly in the release notes please!)

Also would you mind updating the PR description to reflect the latest changes? If we come back to look at this PR it could be confusing since we decided not to use the sentinel value.

Thanks!

BeksOmega:
Ok @maribethb :D PR title and description are updated, `@extends` annotations are removed, inline docs are updated. I think that's everything 🤞

rachel-fenichel:
> Your PR title should say `refactor` instead of `refact`. I feel like the conventional commit linter should have caught this... I don't know why it didn't since `refact` is not in the approved list https://www.npmjs.com/package/@commitlint/config-conventional but ¯\_(ツ)_/¯

The conventional commit linter only checks the shape of the commit message, unfortunately--it doesn't check against an approved list of types.

maribethb (approved):

PR #5953, opened 2022-02-23
refactor(blocks)!: Rename `Blockly.blocks.*` modules to `Blockly.libraryBlocks.*`

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

* Rename the `Blockly.blocks.all` module (`blocks/all.js`) to `Blockly.libraryBlocks` (`blocks/blocks.js`).
* Rename the other`Blockly.blocks.*` modules to `Blockly.libraryBlocks.*`.

#### Behaviour Before Change

* Code requiring the blocks dictionary object does `const {Blocks} = goog.require('Blockly.blocks');`,
  * or uses a `script` tag to load `blockly_compressed.js` and then access the dictionary via `Blockly.Blocks`.
* Code loading library blocks does `goog.require('Blockly.blocks.all');`,
  * or uses a `<script>` tag to load `blocks_compressed.js` and then accesses exports via `Blockly.blocks.all`.

#### Behaviour After Change

* Code requiring the blocks dictionary object is unchanged.
* Code loading library blocks does `goog.require('Blockly.libraryBlocks');`,
  * or uses a `<script>` tag to load `blocks_compressed.js` and then accesses exports via <code>Blockly.libraryBlocks</code>.

There will be no changes visible when the blocks chunk is loaded via ES module `import` or CJS `require`.

### Reason for Changes

An earlier version of this change was originally created as an alternative fix for issue #5932.  PR #5945 is a better fix for the actual underlying cause of that issue, but I proposed PR #5946 anyway simply because of providing a more sensible path for the blocks modules exports object for those loading the compiled chunks directly.  After discussion, this modified approach was settled on.

### Test Coverage

Tested on:
* Desktop Chrome
* `npm test`

### Additional Information

The need for naming of modules will shortly go away entirely when we convert from `goog.module` to ES modules.


rachel-fenichel (approved):

cpcallen:
Amended titles of commit & PR to fix conventional commit lint error.

PR #5957, opened 2022-02-23
refactor: convert row.js and base.js to ES6 classes

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Last of #5860

### Proposed Changes

There were some places where extra properties were getting assigned to `Row` instances. This moves them into WeakMaps instead because they are only used by certain subclass renderers.

I also added a lot of `instanceof` checks for row elements to fix type errors. But I left the `Type.isX()` calls as well. Not sure if we want to remove those or not.

Should be no change in behavior 🤞 

### Reason for Changes

The long winding journey to TypeScript

### Test Coverage

Sanity checked that all renderers look correct. 

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

N/A


core/renderers/measurables/row.js
```diff
@@ -28,207 +28,218 @@ const {Types} = goog.require('Blockly.blockRendering.Types');
...
+     * @type {boolean}
+     */
+    this.hasStatement = false;
+
+    /**
+     * TODO: Do not know what this does! FIX BEFORE MERGE.
+     * @type {number}
+     */
+    this.statementEdge = 0;
+
```
BeksOmega, line 127: @rachel-fenichel

BeksOmega:
I think removing that property assignment in the most recent commit is fine. It was added in #3740 to fix the bounding boxes of blocks, but I don't think it actually did that. I removed the assignment and bounding boxes still look fine:

image

core/renderers/common/drawer.js
```diff
@@ -149,7 +150,8 @@ class Drawer {
         this.outlinePath_ += this.constants_.OUTSIDE_CORNERS.topLeft;
       } else if (Types.isRightRoundedCorner(elem)) {
         this.outlinePath_ += this.constants_.OUTSIDE_CORNERS.topRight;
-      } else if (Types.isPreviousConnection(elem)) {
+      } else if (
+          Types.isPreviousConnection(elem) && elem instanceof Connection) {
```
cpcallen, line 154: I'm assuming that `Types.isPreviousConnection(elem)` implies that `elem instanceof Connection` and you're just helping the compiler figure that out, but if that's not so beware that this could result in a change of behaviour, where cases that would previously have matched this `if` clause could end up falling through and potentially matching later ones.

A slightly more conservative approach to adding a type check here would be:
```JS
      } else if (Types.isPreviousConnection(elem)) {
        if(!(elem instanceof Connection)) throw new TypeError(/*...*/);
```

(Here and various other places.)

cpcallen (commented):
Looks good to me.  I will leave final approval to @rachel-fenichel, once she has looked at the `statementEdge` question.

core/renderers/measurables/row.js
rachel-fenichel: image

I think it's what keeps the left sides of statement inputs aligned. In this case, adding an "else" adds padding to the right of each "do".

image

Otherwise it's controlling where the right side of the bar above statement inputs goes, to make sure it's aligned for all statement inputs.
image

BeksOmega: Based on my reading it was the former :D Thank you Rachel!

core/renderers/common/drawer.js
BeksOmega: I'm pretty sure that these `instanceof` checks shouldn't modify behavior, especially because we rely on properties of the checked class existing.

Honestly, I'm not sure why `Types` exists at all. It looks like it was [meant to unify a bunch of ad-hoc "type" checks](https://github.com/google/blockly/pull/2950/files). But it seems like `instanceof` works just as well.

rachel-fenichel (approved):

core/renderers/common/drawer.js
BeksOmega: Just bumping this to check your opinion before merging @cpcallen

cpcallen: LGTM.

PR #5968, opened 2022-03-01
feat: Backwards execution demo using JS-Interpreter

NeilFraser:
Also resolves existing issue with JS-Interpreter demo where changing the blocks no longer restarts the program.


maribethb:
Before I review this more fully, I think this should go into blockly-samples instead of core. The interpreter demo has already moved there and we have plans to delete most of the contents of demos/ in core already, so this would have to be moved there eventually. I don't think we need to also update the version in core since the work for demos should be done soon; we could go ahead and add a redirect to the samples page now if desired. 

I figure that if you open a new PR in samples it will probably be easier to do a full review with comments there so I will wait and haven't fully read the code. But I have a general question - was the acorn interpreter not used in the previous version of the demo? Are the files here pure copies of the acorn files, or are there changes you made? If they are pure copies, then would it be possible to source the files from npm or through a cdn like unpkg, so that we don't have to maintain a copy of them?

rachel-fenichel:
Closing in favor of https://github.com/google/blockly-samples/pull/1040

Issue #5972, opened 2022-03-03
Wrong type of block created with context menu option "create 'do something'"


gpfernandezflorio:
**Describe the bug**

The caller block created with the context menu option "create 'do something'" is always a "noreturn" no matters if the source block is a "defreturn" or a "defnoreturn"

**To Reproduce**

1. Create a procedures_defreturn block.
![image](https://user-images.githubusercontent.com/5846063/156472010-be6c8e4f-8996-4c02-b44c-3525ed31ef46.png)
2. Open the context menu and select "create 'do something'"
![image](https://user-images.githubusercontent.com/5846063/156472079-adc13cfc-3e94-4b14-b37a-ac4153c50035.png)
3. This is the result:
![image](https://user-images.githubusercontent.com/5846063/156472112-84fc9a64-a1bb-4164-9563-29f2e34d2e09.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome v98.0.4758.102

**Smartphone (please complete the following information):**

**Additional context**

The problem is that both procedures_defnoreturn and procedures_defreturn have the field callType_ assigned with "procedures_callreturn". After changing it from the console, it works.
![image](https://user-images.githubusercontent.com/5846063/156472706-03cd7158-89e7-4796-8573-4c2188c44667.png)

cpcallen:
This is now fixed (and I've added what I believe are suitable lables), but I will leave for oncall to complete triage and close.

PR #5974, opened 2022-03-03
fix(blocks): correct the `callType_` of `procedures_defreturn`

gpfernandezflorio:
## The basics

- [X] I branched from develop

### Resolves

https://github.com/google/blockly/issues/5972

### Proposed Changes

The field _callType__ was defined as 'procedures_callnoreturn' in PROCEDURE_DEF_COMMON.
Now it is defined separately in procedures_defnoreturn and in procedures_defreturn so each one defines it's own value for it.

#### Behavior Before Change

As shown in https://github.com/google/blockly/issues/5972, when selecting the option to create a caller of a returning procedure, the result was a procedures_callnoreturn block.

#### Behavior After Change

Now a procedures_callreturn is created.

### Reason for Changes

To fix the problem.

### Test Coverage

Tested on:
Desktop Chrome v98.0.4758.102 (Windows 11)

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -443,7 +443,6 @@ const PROCEDURE_DEF_COMMON = {
       }
     }
   },
-  callType_: 'procedures_callnoreturn',
 };
 
 blocks['procedures_defnoreturn'] = {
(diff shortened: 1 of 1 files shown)


cpcallen (approved):
Oh, nice catch.  This was totally my fault: when factoring out the common parts of the two blocks in PR #5736 I somehow failed to notice that they had different values.  Thanks for spotting the problem and providing the correct fix!

Issue #5975, opened 2022-03-03
Playground context menu missing "create" item in `develop`


cpcallen:
**Describe the bug**

I was attempting the reproduce steps for issue #5972 in `develop`, but was unable to to find the "create 'do something'" item in the `procedure_defreturn` block's context menu.  It is present in `master`.

**To Reproduce**

1. Check out `develop` branch.
2. `npm start`
3. Open the "Functions" section of the toolbox.
4. Drag a `procedure_defreturn` block (the one with a "return" input) to the workspace.
5. Right-click (or appropriate gesture) to open the context menu for this newly created block.

**Expected behaviour**

The last item in the menu will be "create 'do something'"

**Expected behaviour**

The menu ends with "Help".

**Screenshots**

See #5972, step 2 for expected.

Actual:
![Screenshot 2022-03-03 at 12 12 16](https://user-images.githubusercontent.com/5603958/156562511-b5293ac4-7815-410a-9d79-c6db5823ebda.png)

**Desktop (please complete the following information):**
 - OS: macOS 12.2.1
 - Browser: Chrome 99.0.4844.51
 - Version: `develop`, at 2a0e5475

**Additional context**

I wondered if this have been caused by my changes to playground loading in PR #5931, but `multi_playground.html` has the same problem despite not being converted yet.

cpcallen:
After a quick manual bisect, I see that the problem began when PR #5959 was merged.  @BeksOmega: any idea what might have caused this?

alschmiedt:
Fixed in #5976

PR #5976, opened 2022-03-03
fix: custom block context menus

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#5975 

### Proposed Changes

Moves the context menu property definition to the block instead of block svg.

#### Behavior Before Change

No custom context menus =(

#### Behavior After Change

Custom context menus!

### Reason for Changes

Any custom context menu function that was getting mixed into the block was getting overwritten by the block svg constructor. This fixes that.

### Test Coverage

Manually tested that the procedure context menus appear.

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

Thanks for the catch Christopher!


BeksOmega:
Per offline discussion, I've moved the properties back into BlockSvg (since they are rendered-specific) and changed the assignment to be identity (eg `this.decompose = this.decompose`) to avoid overwriting mixed-in properties.

cpcallen (commented):
As noted in our conversation:
* It's best to actually initialise any properties defined in a class constructor, to keep the shapes of the resulting objects consistent, but where a property is being redefined (to change its type) it is fine to just mention it without initialisation.
* It would be better not to move properties used only by `BlockSvg` to `Block`.
* Where a property in the subclass constructor is initialised to `undefined`, an initialisation like `this.prop = this.prop` will ensure it exists and is set to `undefined` if it was not set by the mixin.
* Changing the type of a superclass property in a subclass and user-defined mixins are both… problematic, as far as type-checking goes—but that is a problem for another ~~day~~ quarter.

tests/deps.js
```diff
@@ -143,7 +143,7 @@ goog.addDependency('../../core/registry.js', ['Blockly.registry'], [], {'lang':
...
 goog.addDependency('../../core/renderers/common/constants.js', ['Blockly.blockRendering.ConstantProvider'], ['Blockly.ConnectionType', 'Blockly.utils.Svg', 'Blockly.utils.colour', 'Blockly.utils.dom', 'Blockly.utils.object', 'Blockly.utils.parsing', 'Blockly.utils.svgPaths', 'Blockly.utils.userAgent'], {'lang': 'es6', 'module': 'goog'});
-goog.addDependency('../../core/renderers/common/debug.js', ['Blockly.blockRendering.debug'], [], {'lang': 'es6', 'module': 'goog'});
+goog.addDependency('../../core/renderers/common/debug.js', ['Blockly.blockRendering.debug'], ['Blockly.utils.deprecation'], {'lang': 'es6', 'module': 'goog'});
```
cpcallen, line 146: This change does not appear to have anything to do with this PR.  Is it just an update that was inadvertently omitted from a previous PR?  If so fine.

cpcallen (approved):

tests/deps.js
BeksOmega: Looks to be, yesh. Probably the PR that deprecated the debug renderer. Reran build:deps and there were no changes.

PR #5979, opened 2022-03-04
fix: move the dropdown div to a namespace instead of a class with only static properties

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Maybe finishing #5860 

### Proposed Changes

Changes the dropdown div to be a namespace instead of a class that only has static properties. This includes changing the module-local variables to conform to the styleguide.

Should be no behavior changes, because I am reexporting the module under its previous name.

### Reason for Changes

Workin' toward typescript.

### Test Coverage

All unit tests pass.

### Documentation

N/A

### Additional Information

N/A


core/dropdowndiv.js
```diff
@@ -163,7 +160,8 @@ DropDownDiv.positionToField_ = null;
  *        height:number
  * }}
  */
-DropDownDiv.BoundsInfo;
+let BoundsInfo;
+exports.BoundsInfo = BoundsInfo;
```
gonfunko, line 154: @cpcallen does this work as expected, or does the type annotation need to be on the export? I'm not quite sure how that all interacts.

core/dropdowndiv.js
```diff
@@ -387,18 +372,17 @@ const showPositionedByRect = function(
  */
 const show = function(
     owner, rtl, primaryX, primaryY, secondaryX, secondaryY, opt_onHide) {
-  owner_ = owner;
-  onHide_ = opt_onHide || null;
+  owner = owner;
```
gonfunko, line 375: Maybe make this newOwner to avoid shadowing?

BeksOmega: Whoops! Thought I caught all of these. I'll do another check.

core/dropdowndiv.js
BeksOmega: This is how [other type defs](https://github.com/google/blockly/blob/c9be8d984d6cc459917f2a777e03ba858f2e623d/core/positionable_helpers.js#L62) work I think. So hopefully should be good? If not I can go through and change those as well.

core/dropdowndiv.js
BeksOmega: Ok I think this was actually the last one!

gonfunko (approved):

core/dropdowndiv.js
cpcallen: I believe this is fine.  And it will soon be a single `export let …`, which is definitely fine as far as annotating it goes.

Issue #5857, opened 2022-01-07
Update Blockly codebase to Typescript


rachel-fenichel:
Steps for this update. These are not necessarily in order, and some will be broken out as separate bugs.

Old Tracking List:
- [x] Add a `tsconfig.json`. (#5892)
- [x] Wire the typescript compiler to output into the build directory such that the closure compiler picks up the correct files. (#5892)
- [x] Add gulp tasks and scripts in `package.json` to run the ts compiler and then the closure compiler. (#5892)
- [x] Add gulp tasks and scripts in `package.json` to clean up generated files when needed. (#5892)
- [x] #5904
- [x] #5960

Must happen pre merge:
- [x] Check all `@enums` for formatting issues. [beka]
  - [x] Fix `KeyCodes` in `core/utils/keycodes.ts`. [christopher]
- [x]  Check all `@typedef`s for lost JSDocs. [beka]
- [x] Export `@typedef`s that were static properties on classes [beka]
- [x] Move comments that migrants put in the wrong place [beka]
- [x] Remove remaining `declareLegacyNamespace` calls [christopher]
- [x] Migrate `blockly.js` to `blockly.ts` [christopher]
- [x] Get Blockly to pass CI. [TBD]

Can happen post merge:
- [x] Fix any `@protected` properties that have been changed to parameter properties [beka] 
- [x] Check that public parameter properties are correct [beka]
- [x] Investigate why `compose` is commented out in BlockSvg [beka]
- [ ] ~~Export interface types that were exported from `blockly.js` [beka]~~
- [X] Redefine properties that were defined using `Object.defineProperties` [christopher]
  - [X] Check that we've properly deprecated all of the above properties, including any in blocks or generators. [beka]
- [X] Find new places to put the properties that were defined using `Object.defineProperties` [beka]
- [X] Implement piping sourcemaps from tsc to closure compiler [christopher]
- [X] Get generating type defs working [beka]
  - [X] Rename instances of `AnyDuringMigration` in defs to `any`[beka]
- [X] Fix packaging sources (.ts instead of .js) [beka]
- [X] Get API Extractor Working
  - [x] Get Maribeth set up with a branch that builds [beka]
  - [X] Enable API Extractor. [maribeth]
  - [X] Migrate JSDoc to TSDoc. [TBD]
    - [ ] Move any docs attached to property assignments to property definitions [TBD]
  - [X] Figure out how to pipe the results of API Extractor to devsite [maribeth]
[...]

PR #5988, opened 2022-03-08
chore: makes types of divs more specific

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #5857

### Proposed Changes

- Add casts when calling `document.createElement('div')`.
- Update types for properties, return types, and parameters that can be more specific as a result.

#### Behavior Before Change

These were annotated as `Element`. The typescript compiler threw errors, saying that `style` and similar properties do not exist on `Element`.

#### Behavior After Change

Code works the same but tsc has fewer errors.

### Reason for Changes

Improve typing/fix errors.
`tsc` doesn't actually care about the casts when the divs are created and used immediately, because it's [clever](https://www.typescriptlang.org/docs/handbook/dom-manipulation.html#documentcreateelement). But `tsc` cares about having the correct types when the divs are saved or passed around, and the closure compiler needs the casts to agree that it's safe.

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -59,7 +59,7 @@ const inject = function(container, opt_options) {
   }
   const options =
       new Options(opt_options || (/** @type {!BlocklyOptions} */ ({})));
-  const subContainer = document.createElement('div');
+  const subContainer = /** @type {!HTMLDivElement} */ (document.createElement('div'));
   subContainer.className = 'injectionDiv';
   subContainer.tabIndex = 0;
   aria.setState(subContainer, aria.State.LABEL, Msg['WORKSPACE_ARIA_LABEL']);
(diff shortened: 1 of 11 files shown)


core/inject.js
```diff
@@ -59,7 +59,8 @@ const inject = function(container, opt_options) {
   }
   const options =
       new Options(opt_options || (/** @type {!BlocklyOptions} */ ({})));
-  const subContainer = document.createElement('div');
+  const subContainer =
+      /** @type {!HTMLDivElement} */ (document.createElement('div'));
```
cpcallen, line 63: If only it would work to type the variable rather than the initialiser:
```
  const /** !HTMLDivElement */ subContainer = document.createElement('div');
```
Alas…

cpcallen (approved):

PR #5993, opened 2022-03-11
refactor(build): Prepare UMD wrapper generation for transition to ES modules

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #5904.

### Proposed Changes

Makes various changes to the way UMD chunk wrappers are generated.  Specifically:

* Corrects erroneous assumption that `closure-calculate-chunks` outputs chunks in the same order as the `--entrypoint`s were specified on the command line.
* Introduces the configuration constant `NAMESPACE_PROPERTY` to control where the namespace object is saved on the exports object.
* Save the namespace object on every chunk exports object.
* Corrects erroneous assumption that chunks could have multiple dependencies; instead, they have a single parent chunk.
* Miscellaneous other updates to chunk wrapper generation.

#### Behaviour Before Change

The namespace object is saved only on the `.internal_` property of the first (`blockly`) chunk's exports object.

#### Behaviour After Change

The namespace object is saved on the `.__namespace__` property of every chunk's exports object.

### Reason for Changes

* Correction of erroneous code so as to avoid generating incorrect output if erroneous assumptions are violated in future.
* Refactor chunk wrapper generation to make it easier to modify to support ES modules, which will not have `goog.module.declareLegacyNamespace` calls.

### Test Coverage

* Passes `npm test`.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -459,14 +459,24 @@ function getChunkOptions() {
   // This is designed to be passed directly as-is as the options
   // object to the Closure Compiler node API, but we want to replace
   // the unhelpful entry-point based chunk names (let's call these
-  // "nicknames") with the ones from chunks.  Luckily they will be in
-  // the same order that the entry points were supplied in - i.e.,
-  // they correspond 1:1 with the entries in chunks.
+  // "nicknames") with the ones from chunks.  Unforutnately there's no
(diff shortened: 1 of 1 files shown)


cpcallen:
@gonfunko / @alschmiedt: please feel free to apply any suggested changes directly, and then go ahead and merge.

gonfunko (approved):

PR #5996, opened 2022-03-14
fix!: change paste to return the pasted thing to support keyboard nav

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on https://github.com/google/blockly-samples/issues/833

### Proposed Changes

Changes the paste functions on the clipboard and the workspace to return the pasted thing (block or workspace comment).

### Reason for Changes

Keyboard nav needs to grab the pasted thing to then move it to the correct spot. This gives keyboard nav a simple and safe API for doing so.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

I'm not sure if this is the thing we want to do or not, because it is a breaking change for anyone using `clipboard.paste(thing) === true`. I just thought posting a PR would be the best way to discuss.

(Personally I think this is the best API, but totally willing to change it for backwards compat reasons)


core/clipboard.js
```diff
@@ -38,13 +42,14 @@ exports.copy = copy;
 
 /**
  * Paste a block or workspace comment on to the main workspace.
- * @return {boolean} True if the paste was successful, false otherwise.
+ * @return {!BlockSvg|!WorkspaceCommentSvg|null} The pasted thing if the paste
```
gonfunko, line 45: This would probably be cleaner and more future-proof if you used ICopyable as the return type like in duplicate(). Probably want to do this in workspace_svg as well.

gonfunko (approved):
This looks good to me; I like the general approach and tend to agree it seems unlikely the API change would be problematic. Probably worth checking with the rest of the team to confirm folks are good with it though before submitting.

Issue #5404, opened 2021-08-23
[Blockly Developer Tools] Switching between tabs causes the workspace to be misaligned


BeksOmega:
**Describe the bug**

If you switch back and forth between the Block Factory tab and the other tabs, then the other tabs, the workspaces will disappear. I think this is because they're becoming misaligned.

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/demos/blockfactory/index.html
2. Switch to the Workspace Factory tab (if it asks you about unsaved work just hit 'ok')
3. Switch back to the Blocks Factory tab.
4. Switch to the Workspace Factory tab. Observe how the workspace has lost its grid.
5. Switch back to the Blocks Factory tab. Observe how the workspace is empty.

**Expected behavior**

No disappearing.

**Screenshots**

![us](https://user-images.githubusercontent.com/25440652/130465074-f1acafff-aefc-40fe-ad36-c4096e35582d.gif)

**Desktop (please complete the following information):**
 - OS: ChromeOS
 - Browser Chrome
 - Version 92

**Additional context**

Originally reported on the forums: https://groups.google.com/g/blockly/c/ys1bwF5WH6o/m/qG4IGBHHAQAJ

Apoorvgarg-creator:
@BeksOmega I would like to contribute to this. Can I work on the issue with you?

BeksOmega:
> @BeksOmega I would like to contribute to this. Can I work on the issue with you?

Sure thing @Apoorvgarg-creator ! I'm not actually working on this issue at the moment, but I'm happy to help in any way I can =)

To get started you'll want to [get set up for development](https://developers.google.com/blockly/guides/modify/development_setup#step_by_step) if you haven't already. Then open up the [block factory index.html](https://github.com/google/blockly/blob/develop/demos/blockfactory/index.html) file in your browser, and make sure you can reproduce the issue given the above steps. After that, you'll have do some debugging to figure out what the issue is, because I'm not exactly sure hehe. A good place to start is probably the [onTab](https://github.com/google/blockly/blob/64188ae27297b1ca23a3244cc987dfbfc6e98f34/demos/blockfactory/app_controller.js#L254) function, because that's what controls changing the what is shown and what is hidden.

We would definitely love help with this bug, but it may not be the best starter issue (because I'm not even sure what's going on!) If you get stuck or bored and still want to contribute, feel free to pick up a different issue, like #5273 or #5374 

And if you need help, always feel free to drop a message here! Best of luck :D

[...]

PR #5997, opened 2022-03-15
fix: Change the truthy tests of width and height in WorkspaceSvg.setCachedParentSvgSize

### LINKED ISSUE #5930: hiding blockly injection div can break blockly
If the div containing the Blockly workspace is hidden and `Blockly.svgResize` is called you end up in a state where the svg is forced to zero size and it will never some back.

The cause is that during the hidden `svgResize` `WorkspaceSvg.setCachedParentSvgSize`  fails to record the zero size as noted here [#5404(907989018)](https://github.com/google/blockly/issues/5404#issuecomment-907989018), but the SVG is still made 0x0.  Later when the div is un-hidden `svgResize` doesn't see that the size is growing since the cached value is bad, so the SVG is not set back to a visible size.

A real case of this is using [NGX-Blockly](https://github.com/roroettg/ngx-blockly) which implements a changable readonly input by having two workspaces with one being readonly.  That workspace is hidden/un-hidden to transition in and out of read-only mode.  If anything calls `Blockly.svgResize` while in read-write mode (in other words while the readonly workspace is in a hidden div) the readonly workspace becomes 0x0 and won't display again even when its div is un-hidden.  The problem manifests because `Blockly.svgResize` gets called in this case if the window is resized due to code in both NGX-Blockly and my own application.  In fact in my application it's worse because I also call `Blockly.svgResize` for several other reflow events, one of which happens ever time you go into read-write mode.

Fix:
The same as what seemed to be noticed in [#5404(907989018)](https://github.com/google/blockly/issues/5404#issuecomment-907989018): change the truthy tests of width and height in `WorkspaceSvg.setCachedParentSvgSize` to actual comparisons with null so that zero value can be saved into the cache.
```diff
 Blockly.WorkspaceSvg.prototype.setCachedParentSvgSize = function(width, height) {
   var svg = this.getParentSvg();
-  if (width) {
+  if (width != null) {
     this.cachedParentSvgSize_.width = width;
     // This is set to support the public (but deprecated) Blockly.svgSize method.
     svg.cachedWidth_ = width;
   }
-  if (height) {
+  if (height != null) {
     this.cachedParentSvgSize_.height = height;
     // This is set to support the public (but deprecated) Blockly.svgSize method.
     svg.cachedHeight_ = height;
```

I have verified that this does solve my problem, and looking around the rest of Blockly I don't see any places where `WorkspaceSvg.setCachedParentSvgSize` was intentionally not caching based on a non-truthy input other than null.

Note that if you use NGX-Blockly to see this then this fix alone is not enough, you also need to re-call `Blockly.svgSize` until [a different issue](https://github.com/roroettg/ngx-blockly/issues/78) is fixed over there.

markfinn:
## The basics

- [x] I branched from develop

### Resolves
  fixes #5930 and possibly #5404

### Proposed Changes
Change the truthy tests of width and height in WorkspaceSvg.setCachedParentSvgSize to actual comparisons with null/undefined so that zero value can be saved into the cache

#### Behavior Before Change

If the div containing the Blockly workspace is hidden and Blockly.svgResize is called you end up in a state where the svg is forced to zero size and it will never some back.

The cause is that during the hidden svgResize WorkspaceSvg.setCachedParentSvgSize fails to record the zero size as noted here https://github.com/google/blockly/issues/5404#issuecomment-907989018, but the SVG is still made 0x0. Later when the div is un-hidden svgResize doesn't see that the size is growing since the cached value is bad, so the SVG is not set back to a visible size.

The same as what seemed to be noticed in https://github.com/google/blockly/issues/5404#issuecomment-907989018: change the truthy tests of width and height in WorkspaceSvg.setCachedParentSvgSize to actual comparisons with null so that zero value can be saved into the cache.

#### Behavior After Change

Zero sizes are saved in to the cache so that they can be noticed later in a comparison when non-zero values become active.  Falsey values previously caused entries to not be recorded, but the actual use case is that null/undefined values should not be recorded.  Zero does not appear to have been an intentional "no update" trigger.

### Reason for Changes

To allow  hiding and unhiding of the blockly div to work correctly.

### Test Coverage

* Desktop Chrome
I've been using this change in chrome for several weeks

### Documentation

none

### Additional Information
none

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1244,13 +1244,13 @@ class WorkspaceSvg extends Workspace {
    */
   setCachedParentSvgSize(width, height) {
     const svg = this.getParentSvg();
-    if (width) {
+    if (width != null) {
       this.cachedParentSvgSize_.width = width;
       // This is set to support the public (but deprecated) Blockly.svgSize
       // method.
       svg.cachedWidth_ = width;
     }
-    if (height) {
+    if (height != null) {
       this.cachedParentSvgSize_.height = height;
       // This is set to support the public (but deprecated) Blockly.svgSize
       // method.


alschmiedt (approved):
lgtm, thanks for the fix!

PR #5998, opened 2022-03-15
chore: remove deprecation console log

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
There is a ton of console spam since I added the console warning for `isDebuggerEnabled`.

### Proposed Changes
Remove the deprecation warning from `isDebuggerEnabled`. If a developer is using the rendering debugger in core they will have to call `startDebugger` so they will get the deprecation warning there and the one in `isDebuggerEnabled` is redundant.

I also removed `display:none` for the debug rectangles from the playground because I don't believe displaying them is a problem since they are hidden by default and if we have this in core it should actually work.

#### Behavior Before Change
`isDebuggerEnabled` spammed the playground with a bunch of console warnings.

#### Behavior After Change
`isDebuggerEnabled` no longer prints out a warnings.

### Reason for Changes
Console spam.


core/renderers/common/debug.js
```diff
@@ -43,6 +38,8 @@ exports.isDebuggerEnabled = isDebuggerEnabled;
  * Turn on the blocks debugger.
  * @package
  * @alias Blockly.blockRendering.debug.startDebugger
+ * @deprecated September 2022. Use the rendering debugger in @blockly/dev-tools.
```
cpcallen, line 41: Checking a few other place where dates appear in `@deprecated` annotations, it appears that they are usually the date of the release in which the `@deprecated` tag was added, so shouldn't this be "March 2022"?

cpcallen (approved):

core/renderers/common/debug.js
alschmiedt: Yup, good call. Fixed.

Issue #6010, opened 2022-03-17
if-return blocks break undo stack when enabled/disabled due to moving


BeksOmega:
**Describe the bug**

When you move an if-return block in/out of a procedure block it gets enabled/disabled. This event does not have a group, which I assume is what's causing the issue.

**To Reproduce**

1. Go to the playground
2. Grab an if-return block and a procedure definition block
3. Drag the if-return block inside the procedure definition block.
4. Attempt to undo the move
5. Observe how you cannot undo

**Expected behavior**

You should be able to undo moving an if-return block in/out of a procedure definition

**Additional context**

Note that the break/continue block does actually work correctly.

BeksOmega:
Closed by #6018

PR #6018, opened 2022-03-22
fix: undo/redo for auto disabling if-return blocks

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#6010 

### Proposed Changes

Wrap the enable/disable event for the if-return in the same group as the initial move event, so the events get undone together. This is the same way the break/continue block handles this behavior.

#### Behavior Before Change

Undoing was not possible because the events were not properly grouped.

#### Behavior After Change

Now the events are properly grouped so undo and redo works.

### Reason for Changes

Undo and redo are helpful.

### Test Coverage

1) Drag an if-return from the workspace to a procedure. Undo and redo.
2) Drag an if-return from the workspace to a block inside a procedure. Undo and redo.
3) Drag an if-return from the workspace to a disabled block inside a procedure. Undo and redo. Enable the disabled block. Observe how the if-return is enabled.
1) Drag an if-return from a procedure to the workspace. Undo and redo.
2) Drag an if-return from a procedure to a block on the workspace. Undo and redo.
3) Drag an if-return from a procedure to a disabled block on the workspace. Undo and redo. Enable the disabled block. Observe how the if-return is not enabled.

 Tested on:
 * Desktop Chrome

### Additional Information

N/A

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -1139,7 +1139,7 @@ blocks['procedures_ifreturn'] = {
   /**
    * Called whenever anything on the workspace changes.
    * Add warning if this flow block is not nested inside a loop.
-   * @param {!AbstractEvent} e Change event.
+   * @param {!AbstractEvent} e Move event.
    * @this {Block}
    */
   onchange: function(e) {
--- a/blocks/loops.js
+++ b/blocks/loops.js
@@ -342,7 +342,7 @@ const CONTROL_FLOW_IN_LOOP_CHECK_MIXIN = {
   /**
    * Called whenever anything on the workspace changes.
    * Add warning if this flow block is not nested inside a loop.
-   * @param {!AbstractEvent} e Change event.
+   * @param {!AbstractEvent} e Move event.
    * @this {Block}
    */
   onchange: function(e) {


blocks/procedures.js
```diff
@@ -1139,11 +1139,12 @@ blocks['procedures_ifreturn'] = {
   /**
    * Called whenever anything on the workspace changes.
    * Add warning if this flow block is not nested inside a loop.
-   * @param {!AbstractEvent} _e Change event.
+   * @param {!AbstractEvent} e Change event.
```
alschmiedt, line 1142: Can you update this to be "Move event"?

blocks/procedures.js
```diff
@@ -1171,14 +1172,15 @@ blocks['procedures_ifreturn'] = {
...
-      if (!this.isInFlyout) {
-        this.setEnabled(true);
-      }
     } else {
       this.setWarningText(Msg['PROCEDURES_IFRETURN_WARNING']);
-      if (!this.isInFlyout && !this.getInheritedDisabled()) {
-        this.setEnabled(false);
-      }
+    }
+    if (!this.isInFlyout) {
```
alschmiedt, line 1178: Just curious why don't we need the `this.getInheritedDisabled` anymore?

alschmiedt (approved):

blocks/procedures.js
BeksOmega: I don't think we ever needed it. That would mean that if you dragged an if-return out of a procedure block, and into a disabled block, it wouldn't get disabled itself. Then you could re-enable the disabled parent, and the if-return would get enabled as well, which is not the behavior we want.

PR #6021, opened 2022-03-23
fix: advanced playground and playground 

### LINKED ISSUE #6019: Hosted advanced playground times out
**Describe the bug**
The hosted advanced playground times out. See #5557 for more information.

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html
2. Notice how it does not load.
**Expected behavior**
It should load.

**Additional context**
I originally did not fix the advanced playground because we were thinking about taking it out of core. Since we [decided to hold off on that](https://github.com/google/blockly/issues/5834#issuecomment-1057568156), I think it should be fixed so that it works when hosted.

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
#6019 and the fact that the playground was not working when hosted.

### Proposed Changes
- Changes the prepare.js script to use absolute path instead of relative path. This means that it will work with both the playground and the advanced playground. This makes it possible to fix #6019.
- Moving to the absolute path also fixed a bug that made it impossible to load the scripts from hosted sites. (We were using ../.. instead of just ../)
- Small fix for the test_theme that was not working.
#### Behavior Before Change
- Playground and advanced playground did not work when hosted.

#### Behavior After Change
- Advanced playground and playground now both work when hosted on appspot.
- Regular playground works when hosted on google.github.io. We would still need to upload the dev-tools/index.js file to the gh-pages branch for this to work on `google.github.io`.
### Reason for Changes
- If we are going to keep the advanced playground in core then it might as well work. 
- Hosted playgrounds working is necessary for testing.

### Test Coverage

- Tested on `google.github.io`
- Tested on appspot
- Tested locally

### DIFF
--- a/tests/playgrounds/advanced_playground.html
+++ b/tests/playgrounds/advanced_playground.html
@@ -3,29 +3,33 @@
 <head>
 <meta charset="utf-8">
 <title>Advanced Blockly Playground</title>
-<script src="../../blockly_uncompressed.js"></script>
-<script src="../../msg/messages.js"></script>
-<script src="../themes/test_themes.js"></script>
+<script src="prepare.js"></script>
 <script src="./screenshot.js"></script>
+<script src="../themes/test_themes.js"></script>
 <script src="../../node_modules/@blockly/dev-tools/dist/index.js"></script>
 <script src="../../node_modules/@blockly/theme-modern/dist/index.js"></script>
 
 <script>
-// Custom requires for the playground.
-// Rendering.
-goog.require('Blockly.minimalist.Renderer');
-goog.require('Blockly.Themes.Zelos');
+const IS_UNCOMPRESSED = Boolean(window.BlocklyLoader);  // See prepare.js
 
-// Other.
-goog.require('Blockly.WorkspaceCommentSvg');
-goog.require('Blockly.libraryBlocks');
-goog.require('Blockly.Dart.all');
-goog.require('Blockly.JavaScript.all');
-goog.require('Blockly.Lua.all');
-goog.require('Blockly.PHP.all');
-goog.require('Blockly.Python.all');
+if (IS_UNCOMPRESSED) {
+  // Custom requires for the playground.
+  // Rendering.
+  goog.require('Blockly.minimalist.Renderer');
+  goog.require('Blockly.Themes.Zelos');
+
+  // Other.
+  goog.require('Blockly.WorkspaceCommentSvg');
+  goog.require('Blockly.libraryBlocks');
+  goog.require('Blockly.Dart.all');
+  goog.require('Blockly.JavaScript.all');
+  goog.require('Blockly.Lua.all');
+  goog.require('Blockly.PHP.all');
+  goog.require('Blockly.Python.all');
+}
 </script>
-<script>
+<script type="module">
+import Blockly from './blockly.mjs';
 'use strict';
 
 function start() {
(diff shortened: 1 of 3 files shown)


tests/playgrounds/prepare.js
```diff
@@ -70,16 +70,30 @@
...
-    document.write('<script src="../../blockly_compressed.js"></script>');
-    document.write('<script src="../../msg/messages.js"></script>');
-    document.write('<script src="../../blocks_compressed.js"></script>');
-    document.write('<script src="../../dart_compressed.js"></script>');
-    document.write('<script src="../../javascript_compressed.js"></script>');
-    document.write('<script src="../../lua_compressed.js"></script>');
-    document.write('<script src="../../php_compressed.js"></script>');
-    document.write('<script src="../../python_compressed.js"></script>');
+    for (let i = 0; i < files.length; i++) {
+      document.write('<script src="/' + root + files[i] + '"></script>');
```
BeksOmega, line 96: For future: this seems like a good example of a place to use template strings

tests/playgrounds/advanced_playground.html
```diff
@@ -3,29 +3,33 @@
...
+  goog.require('Blockly.Themes.Zelos');
+
+  // Other.
+  goog.require('Blockly.WorkspaceCommentSvg');
+  goog.require('Blockly.libraryBlocks');
+  goog.require('Blockly.Dart.all');
+  goog.require('Blockly.JavaScript.all');
+  goog.require('Blockly.Lua.all');
+  goog.require('Blockly.PHP.all');
+  goog.require('Blockly.Python.all');
```
BeksOmega, line 28: I think these should be bundled in by `goog.bootstrap` in prepare.js, could you check if these can be removed?

tests/playgrounds/prepare.js
```diff
@@ -70,16 +70,30 @@
           });
         </script>`);
   } else {
+    // The below code is necessary for a few reasons:
+    // - We need an absolute path instead of relative path because the
+    //   advanced_playground the and regular playground are in different folders.
+    // - We need to get the root directory for blockly because it is
+    //   different for github.io, appspot and local.
```
BeksOmega, line 77: Thank you for this beautiful comment :D

tests/themes/test_themes.js
```diff
@@ -5,24 +5,22 @@
...
-goog.provide('Blockly.TestThemes');
-
 
 /**
  * A theme with classic colours but enables start hats.
  */
-Blockly.Themes.TestHats = Blockly.Theme.defineTheme('testhats', {
+const TestHatsTheme = Blockly.Theme.defineTheme('testhats', {
   'base': Blockly.Themes.Classic
 });
```
BeksOmega, line 14: This has side effects right? And that's how we can access the theme later? Just want to clarify.

BeksOmega (approved):

tests/themes/test_themes.js
alschmiedt: Yes, this has side effects. We register themes in their constructors. However, I believe in this case the `populateTestThemes` method is how we are adding the themes to the advanced playground.

tests/playgrounds/prepare.js
alschmiedt: So when writing this PR I remembered that we probably want at least the basic playground to work on ie11 to make testing easier. I don't think it currently will, but I didn't want to add another piece that we would have to go back and update.

BeksOmega: Ohhh gotcha gotcha. Sweet! That makes sense

Issue #1586, opened 2018-02-04
Demo index page redesign


Rgtemze:
User interface experience of HTML file (/demos/index.html), listing all the demos, could be improved as it looks very simple and non-attractive including navigation-bar, headers etc..  Same also holds for the demos themselves that could be reached from (/demos/index.html), for example, Resizable Demo (/demos/resizable/index.html) does not appear to attract users; however, Blockly Developer Tool (demos/blockfactory/index.html) seems to be good in shape. I would request that those who are experienced in UI provide us with a better design in those HTML files. 
Thank you.

AnmAtAnm:
While we whole-heartedly agree the index page needs some design attention, we'd like to keep the individual demo pages as simple as possible for the purposes for demonstration.  Perhaps iframes can be used to facilitate both goals.

All that said, we have no available cycles or design resources to address this. We welcome pull requests on this issue.

kushthedude:
@AnmAtAnm Should Can I work on it?

moniika:
Most of the demos are now on the [site](https://blockly-demo.appspot.com/static/demos/index.html), but there is still work do to there.

AnmAtAnm:
> Most of the demos are now on the [site](https://blockly-demo.appspot.com/static/demos/index.html), but there is still work do to there.

Isn't that the same site it has always been on?

steviedupree:
Does this still need to be worked on?

alschmiedt:
Hello!

Sorry, yes I think the wrong site was pasted. Here is the new [site](https://google.github.io/blockly-samples/). There is still some work to do with migrating over demos, but I think the new look probably satisfies this issue.
@rachel-fenichel to confirm.

rachel-fenichel:
@steviedupree there is still work to do on migration, if you are interested--specifically, for each demo that is now on blockly-samples, adding a redirect to the new location. There are probably between 10 and 15 demos that need to be properly redirected.

BeksOmega:
We now have a fancy looking demo site in the samples repo.

PR #6045, opened 2022-03-29
chore(demos): redirect some demos to gh-pages

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/1586

### Proposed Changes

Add redirects to the appengine folder so that some of the demos link to the blockly-samples github pages site instead of appengine.

#### Behavior Before Change

Links went to appengine.

#### Behavior After Change

Links go to github pages.

### Reason for Changes

These demos are duplicated on blockly-samples in the examples folder, and hosted on github pages. We don't need to have them in multiple locations. None of these demos save data, so they don't need to be on appengine.

### Test Coverage

Did not test yet, to avoid overriding the current site during testing. It's fine for this to land in Q2.
Need to check whether the icons for the demos still work on the index page.

### Documentation
None.

### Additional Information

Future work:
- Delete demo folders for the affected demos.
- Decide what to do about the demos that I didn't redirect.
- Move the code demo to blockly games.
- Update the index page to point directly to github pages, instead of through redirects.

To consider:
- We'll lose the ability to locally test some of these configurations by just rebuilding and opening demo pages. I'm okay with that, because the advanced playground lets us test configurations trivially, and the demos are generally part of the end-of-quarter release process rather than normal testing.
- There's no longer a direct link to the demo index page from developers.google.com/blockly. I'm not sure when we deleted it.

### DIFF
--- a/appengine/redirect.html
+++ b/appengine/redirect.html
@@ -61,6 +61,37 @@ if (loc.match('/apps/puzzle/')) {
   loc = loc.replace('/apps/', '/demos/');
 }
 
+// Demos without saved data were moved to Blockly Samples in 2021.
+if (loc.match('/demos/fixed/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/fixed-demo/';
+} else if (loc.match('/demos/resizable/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/resizable-demo/';
+} else if (loc.match('/demos/toolbox/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/toolbox-demo/';
+} else if (loc.match('/demos/maxBlocks/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/max-blocks-demo/';
+} else if (loc.match('/demos/generator/')) {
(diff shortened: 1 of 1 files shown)


maribethb (approved):

appengine/redirect.html
```diff
@@ -61,6 +61,37 @@
...
+} else if (loc.match('/demos/custom-dialogs/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/custom-dialogs-demo';
+} else if (loc.match('/demos/custom-fields/turtle/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/turtle-field-demo/';
+} else if (loc.match('/demos/custom-fields/pitch/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/pitch-field-demo/';
+} else if (loc.match('/demos/mirror/')) {
+  loc = 'https://google.github.io/blockly-samples/examples/mirror-demo/';
+}
+
```
cpcallen, line 94: Adding these here without also adding the relevant URLs to `app.yaml` has no effect.  Who knew?  ¯\\\_(ツ)\_/¯

PR #6049, opened 2022-03-30
fix: Right-align blocks in horizontal RTL flyouts.

### LINKED ISSUE #2309: Horizontal RTL Flyouts are Left Aligned
### Problem statement

Blocks in a horizontal RTL flyout are left aligned. This looks bad if there are only a few blocks in the flyout, or the flyout is zoomed out.

leftalignedtoolbox

### Expected Behavior

Blocks should be right-aligned.

### Actual Behavior

Blocks are left-aligned.

### Steps to Reproduce

1. Open the multiplayground.
2. Set it to simple toolbox mode
3. Navigate to the RTL Horizontal Toolboxes.
4. Zoom out. Observe how the blocks are left aligned.

I am assuming this is not browser specific.

### Additional Information

N/A

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
#2309

### Proposed Changes
This PR causes blocks/buttons in a horizontal flyout in an RTL workspace to be right-aligned when the width of the blocks is less than the width of the flyout.

#### Behavior Before Change
Blocks and buttons were left-aligned in the flyout.

### Test Coverage
Verified that test suite passes and manually verified behavior.

### DIFF
--- a/core/flyout_horizontal.js
+++ b/core/flyout_horizontal.js
@@ -28,7 +28,6 @@ const {Options} = goog.requireType('Blockly.Options');
 const {Rect} = goog.require('Blockly.utils.Rect');
 const {Scrollbar} = goog.require('Blockly.Scrollbar');
 
-
 /**
  * Class for a flyout.
  * @extends {Flyout}
@@ -42,6 +41,12 @@ class HorizontalFlyout extends Flyout {
   constructor(workspaceOptions) {
     super(workspaceOptions);
     this.horizontalLayout = true;
+    this.contentWidth_ = 0;
+    window.addEventListener('resize', () => {
+      if (this.RTL) {
+        this.reflowInternal_();
+      }
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


core/flyout_horizontal.js
```diff
@@ -42,6 +41,12 @@ class HorizontalFlyout extends Flyout {
   constructor(workspaceOptions) {
     super(workspaceOptions);
     this.horizontalLayout = true;
+    this.contentWidth_ = 0;
+    window.addEventListener('resize', () => {
+      if (this.RTL) {
+        this.reflowInternal_();
+      }
+    });
```
alschmiedt, line 49: I think this should use conditionalBind method. So something like: 

```
this.flyoutResizeHandler_ =
      browserEvents.conditionalBind(window, 'resize', null, function() {});
```
And then we need to dispose of it in the `dispose` method: 

```
browserEvents.unbind(this.flyoutResizeHandler_);
```

alschmiedt (approved):
Ok so my main concern here is that we already have a resize event listener on the workspace that calls flyout.position. Both flyout.position and flyout.reflowInternal update or use this.flyout.height_ which seems a bit fragile imo. I haven't been able to think of a better fix and it is pretty limited scope so I am fine with this going in as is however if you have any ideas feel free lmk : )

BeksOmega:
Gonna go ahead and close this. Can reopen during bug bash.

Issue #5580, opened 2021-10-03
Move internal dynamic categories to use JSON


BeksOmega:
**Is your feature request related to a problem? Please describe.**

The XML serializaation system is being iceboxed, so internal usages of the XML system should be moved to the new JSON system.

In particular, dynamic categories which are internal to Blockly should be changed to use JSON

**Describe the solution you'd like**

* Change the dynamic [variable](https://github.com/google/blockly/blob/a9f9086e03745f010b798b8fcb880d15bad40159/core/variables.js#L121) category to use json
* Change the dynamic [typed variable ](https://github.com/google/blockly/blob/a9f9086e03745f010b798b8fcb880d15bad40159/core/variables_dynamic.js#L43)category to use json
* Change the dynamic [procedure](https://github.com/google/blockly/blob/a9f9086e03745f010b798b8fcb880d15bad40159/core/procedures.js#L193) category to use json

**Describe alternatives you've considered**

N/A

**Additional context**

Note that currently we suport dynamic categories being defined in JSON or XML, but it's good to move the internal definitions to JSON.

bhsingla:
Hi team, I would love to work on this. Thanks

rachel-fenichel:
Hi @bhsingla! This is being worked on in the `blockly-badger` breakout room.

coolcoder03:
how can i work on this?

Apoorvgarg-creator:
Is this issue still available ?

BeksOmega:
Yep go for it @Apoorvgarg-creator !

Apoorvgarg-creator:
@BeksOmega  Thanks for Assigning it to me, Since it's a feature request and I am new to blockly, I would appreciate if you could give me a brief on how to get started with the feature ! What all should I know and how to test my features.
Thanks once again !

BeksOmega:
> @BeksOmega Thanks for Assigning it to me, Since it's a feature request and I am new to blockly, I would appreciate if you could give me a brief on how to get started with the feature ! What all should I know and how to test my features. Thanks once again !

Oh whoops, this shouldn't have been marked as a feature request!

[...]

PR #6050, opened 2022-03-30
fix: internal dynamic categories using JSON

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#5580 

### Proposed Changes

Changes the internal dynamic categories to use JSON instead of XML.

Also moves registration of button callback to the workpace so they only get registered once, instead of every time the dynamic category opens.

Should be no change in behavior 🤞 

### Reason for Changes

XML system is iceboxed.

### Test Coverage

* Testing adding variables (including sort order) for both typed and untyped variable categories.
* Tested adding procedures.
* Tested adding parameters to procedures.

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

Existing category generators could not be removed because they are public. Deprecation warnings have been added to them.


BeksOmega:
So I think the build is broken because `flyoutCategoryBlocks` is calling `generateVariableFieldDom` which is technically deprecated. But it needs to keep calling that for backwards compatibility, and I don't want to undeprecate it... so idk how to deal with this situation.

core/procedures.js
```diff
@@ -294,6 +298,67 @@ const flyoutCategory = function(workspace) {
 };
 exports.flyoutCategory = flyoutCategory;
 
+/**
+ * Construct the blocks required by the flyout for the procedure category.
+ * @param {!WorkspaceSvg} workspace The workspace containing procedures.
+ * @return {!Array<!Object>} Array of objects defining the contents of the
+ *     category.
+ */
+const flyoutCategoryJson = function(workspace) {
```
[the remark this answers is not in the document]
cpcallen: Actually, upon reading further I can see that `flyoutCategory` is used elsewhere too, so maybe it is best to be consistent.  I still think `Json` is probably not the best possible suffix, though.  🤔

core/procedures.js
```diff
@@ -212,9 +213,12 @@ exports.rename = rename;
  * Construct the blocks required by the flyout for the procedure category.
  * @param {!WorkspaceSvg} workspace The workspace containing procedures.
  * @return {!Array<!Element>} Array of XML block elements.
+ * @deprecated
```
cpcallen, line 216: What should be  used instead?  (Here and every other place this tag appears.)

core/variables.js
```diff
@@ -192,6 +199,123 @@ const flyoutCategoryBlocks = function(workspace) {
 };
 exports.flyoutCategoryBlocks = flyoutCategoryBlocks;
 
+/**
+ * Generate DOM objects representing a variable field.
+ * @param {!VariableModel} variableModel The variable model to
+ *     represent.
+ * @return {?Element} The generated DOM.
+ * @deprecated
```
cpcallen, line 207: Here especially I think specifying what should be used instead might help you answer the question you pose in your PR comment.

core/variables.js
```diff
@@ -192,6 +199,123 @@ const flyoutCategoryBlocks = function(workspace) {
...
+
+/**
+ * Construct the elements (blocks and button) required by the flyout for the
+ * variable category.
+ * @param {!WorkspaceSvg} workspace The workspace containing variables.
+ * @return {!Array<!Object>} Array of objects defining the contents of the
+ *     category.
+ * @package
+ */
+const flyoutCategoryJson = function(workspace) {
```
cpcallen, line 234: See above.

core/workspace_svg.js
```diff
@@ -488,21 +488,34 @@ class WorkspaceSvg extends Workspace {
     this.flyoutButtonCallbacks_ = Object.create(null);
 
     const Variables = goog.module.get('Blockly.Variables');
-    if (Variables && Variables.flyoutCategory) {
+    if (Variables) {
+      this.registerButtonCallback(
+          'CREATE_VARIABLE',
+          (button) => Variables.createVariableButtonHandler(
+              button.getTargetWorkspace()));
```
cpcallen, line 495: Why was this call  to `.registerButtonCallback` not needed previously?  (Here and below.)

cpcallen (changes requested):
Looks good except for some small details and the obvious issue of the build failure!

> So I think the build is broken because `flyoutCategoryBlocks` is calling `generateVariableFieldDom` which is technically deprecated. But it needs to keep calling that for backwards compatibility, and I don't want to undeprecate it... so idk how to deal with this situation.

Why is `generateVariableFieldDom`  deprecated, and what is the replacement?

* If it needs to be retained for backwards compatibility then it should probably not be deprecated.
* If it has a replacement, you should call that instead.
* If you really don't like either of those options and are feeling a bit chaotic neutral, you could just `@suppress` the diagnostic at the point that you make the call.

core/procedures.js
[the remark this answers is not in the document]
BeksOmega: I'm happy to change it to something else (I'm thinking of renaming the json version of `callbackFactory` in an upcoming PR). But if you want to retain it, I can't think of a better suffix than `Json` :/

core/workspace_svg.js
BeksOmega: It existed before, it was just called from `flyoutCategory` which would unnecessarily re-register the callback every time the flyout was opened. This is the proper place for the registrations.

BeksOmega:
Ah ha ha! Wrt the build failure, turns out it shouldn't actually be a problem. The calls that are making the build fail are coming from the custom context menu options (which I'm fixing next) so I can just suppress the failures for now.

I thought they were coming from `flyoutCategoryBlocks` which I also just deprecated, and I was like... idk how to tell closure this is ok. Turns out it already knows! (as it should)

Issue #6076, opened 2022-04-14
Dragging fails to complete for collapsed procedure blocks that have been deserialised from JSON


cmjcharlton:
**Describe the bug**

If I attempt to drag and drop a collapsed procedure block that has been loaded from JSON the drag fails to complete and the block gets stuck to the mouse cursor.

**To Reproduce**

1. Open the Blockly playground
2. Click on 'Functions' in the toolbox
3. Drag a procedure block into the workspace
4. Right-click and collapse the block
5. Click 'Save JSON'
6. Click 'Load'
7. Attempt to drag and drop the block

**Expected behavior**

The block should remain in the location where I release the mouse button and no errors should be reported in the developer console.

**Screenshots**

![image](https://user-images.githubusercontent.com/90400333/163387882-6a0ce88b-96ba-4fc9-bb9c-6b8c208edd84.png)

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser [Chrome, Firefox]
 - Version [Chrome 100.0.4896.88 (Official Build) (64-bit), Firefox 99.0.1 (64-bit)]

**Stack Traces**

```
coordinate.js:134 Uncaught TypeError: Cannot read properties of null (reading 'x')
    at Function.module$exports$Blockly$utils$Coordinate.Coordinate.sum (coordinate.js:134:29)
    at BlockDragger.module$exports$Blockly$BlockDragger.BlockDragger.dragIcons_ (block_dragger.js:432:44)
    at BlockDragger.module$exports$Blockly$BlockDragger.BlockDragger.drag (block_dragger.js:212:10)
    at module$exports$Blockly$TouchGesture.TouchGesture.module$exports$Blockly$Gesture.Gesture.startDraggingBlock_ (gesture.js:466:24)
    at module$exports$Blockly$TouchGesture.TouchGesture.module$exports$Blockly$Gesture.Gesture.updateIsDraggingBlock_ (gesture.js:397:12)
    at module$exports$Blockly$TouchGesture.TouchGesture.module$exports$Blockly$Gesture.Gesture.updateIsDragging_ (gesture.js:447:14)
    at module$exports$Blockly$TouchGesture.TouchGesture.module$exports$Blockly$Gesture.Gesture.updateFromEvent_ (gesture.js:288:12)
    at module$exports$Blockly$TouchGesture.TouchGesture.module$exports$Blockly$Gesture.Gesture.handleMove (gesture.js:555:10)
[...]

Issue #5146, opened 2021-07-21
Refactor block and blocksvg initialization


BeksOmega:
**Is your feature request related to a problem? Please describe.**

The block initialization system needs some work (putting it generously). The logic is currently stored in the [XML system](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L560) this is bad because one:
  1) It violates the single responsibility principle, introducing fragility into a high-stakes system.
  2) It makes it hard for outside devs to create blocks programmatically, because the initialization logic isn't surfaced through a good API. See the [docs](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/workspace_svg.js#L994) I added to `workspace.newBlock` a while ago.

Setting aside the API, the actual logic also needs some work.
  1) Headless and rendered blocks are handled separately using [type checks](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L560). This goes against the principles of polymorphism/inheritance. There should be an `init` function on both types which handles the logic of each.
  2) To work efficiently, this relies on calling `initSvg`  and then `render` the blocks from leaf blocks to parent blocks, *and* not allowing the `render` calls to bubble. See [here](https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L564). This is unclear and fragile.

**Describe the solution you'd like**

Both block types should have an `init` function which handles the logic for that type. This would:
  1) Follow the single responsibility principle by removing init logic from the XML system.
  2) Follow the principles of polymorphism/inheritance.
  3) Make life easier for outside devs, because they would only have to call one function. We might even be able to add the call to `workspace.newBlock`

I think it would be nice to have this `init` function recursively initialize and render child blocks (in such a way that there are no duplicate calls). But this may conflict with the fact that `render` calls generally bubble up to parents, and complicate the API.

**Describe alternatives you've considered**

N/A

**Additional context**

I really want to fix this so that I don't have to add this ugly init logic to the new serializer. But it's far too big to be in scope of my project :/

Note for external contributors that might want to take this on: This is a *big* project that would require *lots* of testing.

tweini:
I came form here: https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/serialization/blocks.js#L611
[...]

PR #6081, opened 2022-04-17
fix: dragging fails for collapsed blocks with Icons, which have been …

tweini:
…deserialized from JSON

## The basics
- [x] I branched from develop

## The details
Fix for "Dragging fails to complete for collapsed procedure blocks that have been deserialised from JSON #6076".
The bug affect every block with icons, not only procedures. So comments in a collapsed block will trigger this bug too.
Precondition is a deserialization from JSON.

### Resolves
[bug #6076](https://github.com/google/blockly/issues/6076)

### Proposed Changes
Provides an appropriate check if the Icon is visible, as this guarantees an initialized state of the Icon Object belonging to the block.
Not moving the not visible Icons should not have side effects, as expanding the block will render the icons so they get a proper state.
An alternative solution would be a simple null check, but this could chew future errors. The specific checks are imho the better way.

### Test Coverage
Manually tested with some blocks on a playground.

Tested on: This bug is not related to the client for obvious reasons.

### Documentation
There is a code comment in the PR explaining the changes and mentions the # of the issue

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -429,11 +429,7 @@ const BlockDragger = class {
     // Moving icons moves their associated bubbles.
     for (let i = 0; i < this.dragIconData_.length; i++) {
       const data = this.dragIconData_[i];
-      // Don't move hidden Icons as they can have not been initialized.
-      // fixes #6076
-      if (!this.draggingBlock_.isCollapsed() || !data.icon.collapseHidden) {
-        data.icon.setIconLocation(Coordinate.sum(data.location, dxy));
-      }
+      data.icon.setIconLocation(Coordinate.sum(data.location, dxy));
     }
   }
 
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -429,7 +429,11 @@ const BlockDragger = class {
     // Moving icons moves their associated bubbles.
     for (let i = 0; i < this.dragIconData_.length; i++) {
       const data = this.dragIconData_[i];
-      data.icon.setIconLocation(Coordinate.sum(data.location, dxy));
+      // Don't move hidden Icons as they can have not been initialized.
+      // fixes #6076
+      if (!this.draggingBlock_.isCollapsed() || !data.icon.collapseHidden) {
+        data.icon.setIconLocation(Coordinate.sum(data.location, dxy));
+      }
     }
   }
 
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -398,6 +398,12 @@ const appendPrivate = function(state, workspace, {
   loadInputBlocks(block, state);
   loadNextBlocks(block, state);
   initBlock(block, workspace.rendered);
+  // fixes #6076 JSO deserialization doesn't
+  // set .iconXY_ property so here it will be set
+  const icons = block.getIcons();
+  for (let i = 0; i < icons.length; i++) {
+    icons[i].computeIconLocation();
+  }
 
   return block;
 };
@@ -526,12 +532,6 @@ const loadIcons = function(block, state) {
       const blockSvg = /** @type {!BlockSvg} */ (block);
       setTimeout(() => blockSvg.getCommentIcon().setVisible(true), 1);
     }
-  }
-  // fixes #6076 JSO deserialization doesn't
-  // set .iconXY_ property so here it will be set
-  const icons = block.getIcons();
-  for (let i = 0; i < icons.length; i++) {
-    icons[i].computeIconLocation();
   }
 };
 
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -398,12 +398,6 @@ const appendPrivate = function(state, workspace, {
   loadInputBlocks(block, state);
   loadNextBlocks(block, state);
   initBlock(block, workspace.rendered);
-  // fixes #6076 JSO deserialization doesn't
-  // set .iconXY_ property so here it will be set
-  const icons = block.getIcons();
-  for (let i = 0; i < icons.length; i++) {
-    icons[i].computeIconLocation();
-  }
 
   return block;
 };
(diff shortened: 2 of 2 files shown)


BeksOmega:
Hello! Thank you for the contribution :D Were you able to find out why this was happening with JSON but not with XML?

tweini:
Hi Beka! Well, although I am a big fan of cereals, I didn't read your JSO serializer code, yet. Shall I?

tweini:
Okidoki, I managed to track down the bug in the JSO serializer. It misses to call Icon.computeIconLocation() [fyi: all calls to this method in entire codebase](https://github.com/google/blockly/search?q=computeIconLocation)

Some call stacks:

**Call stack when deserialized from xml:**
computeIconLocation (icon.js:176)
moveConnections (block_svg.js:800)
moveBy (block_svg.js:473)
domToWorkspace (xml.js:471)
load (playground.html:173)

**Call stack when deserialized from JSON:**
initBlock (blocks.js:618)
appendPrivate (blocks.js:400)
appendInternal (blocks.js:345)
append (blocks.js:308)
load (blocks.js:683)
load (workspaces.js:89)
load (playground.html:170)

And now the fun begins. Here I would place the computeIconLocation()-calls for the icons:

https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/serialization/blocks.js#L513

#2105 tells about an Icon API. Is there anything in progress?

So, before providing a new PR, I thought it will be better to sort out what is the best way to do it.

I am fine with contributing, and I am happy digging into the Blockly codebase 😀

tweini:
Pleasure is on my side @BeksOmega . Debugging is so much fun 😊 so I did further investigation.

The truth is not as simple as you proposes 😨 , so sorry.

Deserializing an Icon does not set proper location, so the .iconXY_ property remains _null_ although the the icon object is created and attached to the block.

Setting the icon location is a side effect on both serializers. While xml deserialization, the icon is drawn (really appears on the workspace!) without location here:
https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/block_svg.js#L290

This happens no matter the block is collapsed or not. If not collapsed, the icon gets its position with the call from the drawer L337 and is then properly drawn at its new position.

The uncollapsed block draws the icon as mentioned above. But its visibility state is maintained afterwards in block_svg L679 and then the drawer L111 hides it by setting display:none. The location at this point is _null_ and it is set (as a side effect) in block_svg L800.

I am sure you are aware of the problems, as I visited two issues from you during my investigation. For now, this bug must be patched, and a proper call to .computeIconLocation() would raise awareness for the problem in the JSO serializer.

The handling of Icons causes horrible things (citation from #2105 😄 ), which can't be unseen. And I think it is touching issue #5146 as well as .initSvg() does the mentioned call (see above, L290).

BeksOmega:
So what solution are you thinking of @tweini ?

tweini:
> I am sure you are aware of the problems, as I visited two issues from you during my investigation. For now, this bug must be patched, and a proper call to .computeIconLocation() would raise awareness for the problem in the JSO serializer.

Did change my PR with the new solution mentioned above. Hope my shared infos helps. I got a better understanding of the internals.

BeksOmega:
> > I am sure you are aware of the problems, as I visited two issues from you during my investigation. For now, this bug must be patched, and a proper call to .computeIconLocation() would raise awareness for the problem in the JSO serializer.
> 
> Did change my PR with the new solution mentioned above. Hope my shared infos helps. I got a better understanding of the internals.

Sweet! I couldn't tell if by:

> The truth is not as simple as you proposes 😨 , so sorry.

You meant it also needed a different solution hehe. Either way I really appreciate the added context about the internals!

core/block_dragger.js
```diff
@@ -429,7 +429,11 @@ const BlockDragger = class {
     // Moving icons moves their associated bubbles.
     for (let i = 0; i < this.dragIconData_.length; i++) {
       const data = this.dragIconData_[i];
-      data.icon.setIconLocation(Coordinate.sum(data.location, dxy));
+      // Don't move hidden Icons as they can have not been initialized.
+      // fixes #6076
+      if (!this.draggingBlock_.isCollapsed() || !data.icon.collapseHidden) {
```
BeksOmega, line 434: Do we still need this fix?

tweini: Nope, core/block_dragger.js remains untouched. I thought i got rid of the changes I made in my first commit 🤔

tweini: Wait! This code did trick my manual testing into false positive! More info in main thread... .

tweini:
As mentioned, I thought this code was removed. The fix fails if the icon is a mutator, as the mutator isn't serialized in the icon state of the serialization object. so we have to move it to the end. PR follows...

BeksOmega:
Heyo! So I think the issue with the build is that icons don't exist in headless mode (as discussed in https://github.com/google/blockly/issues/2105#issuecomment-885948249) If you move the `computeIconLocation` calls to [after this line](https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/serialization/blocks.js#L625) I think it should work.

Thanks for catching that issue with mutators! I wouldn't have seen that from looking at the code.

tweini:
Thanks for your suggestion, works like a charm. I ran the tests locally, and gladly the erroneous commit put this step back in scope 👓 . And I begin understanding what #2105 is about, as it is also referenced in you project cereal doc.
Hope all is fine, the next bug will be fixed later on. Was another deep dive into serialization internals 🤿

BeksOmega (approved):

BeksOmega:
Thank you again for the fix and all of the digging! :D

tweini:
@BeksOmega I am very happy 😃 , short celebration 🥳 , opened next pull request to prepare more happiness 😆 (and parties 🎉 )

Issue #6097, opened 2022-04-21
Additional type and cast cleanups


rachel-fenichel:
### Context

We enabled `strictMissingTypes` and `strictCheckProperties` in the closure build as a prerequisite for running MigranTS. Where possible we fixed type annotations properly, but in other places we need to do a refactor to make the types make sense--particularly where we tack properties onto SVG Elements to pass information around.

### Sub-issues
- [ ] `.tooltip` property on SVG Elements
- [ ] Touch/Gesture code uses `Event` and `PseudoEvent` instead of more specific types
  - [ ] `PointerEvents` are implemented on all of our supported browsers, so we can get rid of some of the touch code.
- [ ] `skew` and `translate` on SVG Elements in 

### See also
https://github.com/google/blockly/pull/6073
https://github.com/google/blockly/issues/5857

PR #6102, opened 2022-04-22
fix: add compose and decompose to block

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of migration to typescript.

### Proposed Changes

Add annotations for compose and decompose functions on block.
Change a cast from `*` to `?`.

### Reason for changes

Fixes several type errors blocking typescript conversion.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -179,6 +179,20 @@ class Block {
      */
     this.getDeveloperVariables = undefined;
 
+    /**
+     * An optional function that reconfigures the block based on the contents of
+     * the mutator dialog.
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
Technically the compiler *shouldn't* accept this, because the BlockSvg versions of these functions expect more restricted types - which is bad. But also technically we always use the different versions correctly so it's fine. We just can't communicate that to the compiler without casting all the time.

rachel-fenichel:
I'll add a note in https://github.com/google/blockly/issues/6097 to about the more specific types.

Issue #3725, opened 2020-03-06
Procedure callers are tightly coupled to mutator UI


BeksOmega:
**Describe the bug**

Procedure callers are tightly coupled to the mutator UI implementation. So there are some things that are impossible to do with a new mutator UI.

Specifically, you cannot reorder the parameters because the caller relies on blockIds from the mutator.

**To Reproduce**

Currently I have a branch of the blockly-samples repo that replicates this behavior. You can pull it [here](https://github.com/google/blockly-samples/compare/master...BeksOmega:demo/alternate-mutators) if you would like to replicate it.
1. Drag out a procedure def and a caller.
2. Add two params using the plus button.
2. Add number blocks to the param inputs.
3. Switch the order of the params using the arrows.
4. Observe how the param order switches but the blocks are not moved correctly.
![ParamOrder](https://user-images.githubusercontent.com/25440652/76047875-f4415e80-5f18-11ea-8566-bff96687fdad.gif)

**Expected behavior**

Caller should be UI-Agnostic.

**Additional context**

Problems originate in the procedure caller's [setProcedureParameters](https://github.com/google/blockly/blob/1a77a5e85938b9cd76a363f4a5b013b7a3324b69/blocks/procedures.js#L650) function.

Here is a demo pull request showing one method of fixing it: https://github.com/google/blockly/pull/3724
It is based on the idea that this can be fixed by having the procedure caller rely on variable ids instead of block ids.

BeksOmega:
So... demo fixes the problem, but makes another problem. I didn't realize that creating another variable each time the name was edited was the intended behavior (it prevents the user from accidently renaming things they didn't want to, or accidently coalescing variables). So back to the drawing board
(┏－‿ ◕)┏

rachel-fenichel:
> creating another variable each time the name was edited was the intended behavior (it prevents the user from accidently renaming things they didn't want to, or accidently coalescing variables). 

Did you figure this out because we have a test for it (seems unlikely) or just by exploring the code?

BeksOmega:
> > creating another variable each time the name was edited was the intended behavior (it prevents the user from accidently renaming things they didn't want to, or accidently coalescing variables).
> 
> Did you figure this out because we have a test for it (seems unlikely) or just by exploring the code?

Just by exploring.

BeksOmega:
Demo mk2: https://github.com/google/blockly/pull/3750

BeksOmega:
This will be fixed as part of #6526

PR #6103, opened 2022-04-22
fix: JSON deserialization fails (bug #6091) (collapsed procedure call…

### LINKED ISSUE #6091: JSON deserialization fails for collapsed procedure call blocks containing at least one input variable
**Describe the bug**

If I attempt to load a workspace from JSON containing a collapsed procedure call block with at least one input variable then this block is missing from the loaded workspace.

**To Reproduce**

1. Open the Blockly playground
2. Click 'Functions' in the toolbox
3. Drag a procedure block into the workspace
4. Open its mutator and add an input
5. Right-click the block and select 'Create "do something"'
6. Right-click the created procedure call block and collapse it
7. Click 'Save JSON'
8. Click 'Load'

**Expected behavior**

Both the procedure definition and call procedure blocks should be present in the loaded workspace.

**Screenshots**

Workspace to be saved:
image

Workspace after load:
image

 - OS: [Windows 10]
 - Browser [Chrome, Firefox]
 - Version [Chrome 100.0.4896.127 (Official Build) (64-bit), Firefox 99.0.1 (64-bit)]

**Stack Traces**

```
block.js:2022 Uncaught Error: Input not found: _TEMP_COLLAPSED_INPUT
    at BlockSvg.module$exports$Blockly$Block.Block.removeInput (block.js:2022:11)
    at BlockSvg.module$exports$Blockly$BlockSvg.BlockSvg.removeInput (block_svg.js:1376:27)
    at BlockSvg.module$exports$Blockly$BlockSvg.BlockSvg.updateCollapsed_ (block_svg.js:673:12)
    at BlockSvg.module$exports$Blockly$BlockSvg.BlockSvg.setCollapsed (block_svg.js:647:12)
    at BlockSvg.setProcedureParameters_ (procedures.js:739:10)
    at BlockSvg.loadExtraState (procedures.js:899:12)
    at module$contents$Blockly$serialization$blocks_loadExtraState (blocks.js:452:11)
    at module$contents$Blockly$serialization$blocks_appendPrivate (blocks.js:394:3)
    at module$exports$Blockly$serialization$blocks.appendInternal (blocks.js:345:17)
    at module$exports$Blockly$serialization$blocks.append (blocks.js:308:10)
```

**Additional context**

Although it involves collapsed blocks I don't believe that this is the exactly the same as issue #6076 as it appears to be related to extraState rather than icon visibility, however again the error does not appear to occur if saving/loading via XML. It also does not occur for either deserialization method if the procedure call block contains no input variables.

tweini:
… blocks)

## The basics

- [x] I branched from develop
- [x] BONUS: I ran `npm run test`

### Resolves

#6091

### Proposed Changes

Flipping the call order in core/serialization/blocks.js  of .loadAttributes() and .loadExtraState() does the trick.

#### Behavior Before Change

During deserialization of the call block, a very nasty bug begins here:

https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/blocks/procedures.js#L739

Because the collapsed attribute was set to true beforehand, this becomes a crime scene:
https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/block_svg.js#L642

this.collapsed_ is true, so the if-cond isn't met and super.setCollapsed(collapsed) is getting executed. This sets collapsed to false, so the condition if (!collapsed) is met and this.updateCollapsed_() gets executed.

updateCollapse_() maintains the logic of setting a dummy with _TEMP_COLLAPSED_INPUT (hence the error "block.js:2022 Uncaught Error: Input not found: _TEMP_COLLAPSED_INPUT")

#### Behavior After Change

For whom is interested in, the "right" call to updateCollapse_() will happen here:
https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/block_svg.js#L1741

As the block must get rendered collapsed.

### Reason for Changes

bug 🐛 

### Test Coverage

Existing tests passed. Manually tested both types "procedures_callnoreturn" and procedures_callreturn" in collapsed state, and observed changed behaviour in debugger.

**Potential Failures:**
As changing the order does change the deserialization of other "attributes":
https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/core/serialization/blocks.js#L426

I spotted the data-attribute as a potential "let's introduce a regression bug", as it shows up very early in xml deserialization:
https://github.com/google/blockly/blob/d4196c1105d3d7719e967edb0cecd0d7f76508c0/core/xml.js#L882

So it may be better to split up the .loadAttributes() method?

### Documentation

Maybe I should have left a comment in the code 😳 

### Additional Information

Yes, I will add a comment in the code. I will do this, after the code is reviewed.

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -721,10 +721,10 @@ const PROCEDURE_CALL_COMMON = {
     if (!mutatorOpen) {
       this.quarkConnections_ = {};
       this.quarkIds_ = null;
-    }
-    if (!paramIds) {
-      // Reset the quarks (a mutator is about to open).
-      return;
+    } else {
+      // fix #6091 - this call could cause an error when outside if-else
+      // expanding block while mutating prevents another error (ancient fix)
+      this.setCollapsed(false);
     }
     // Test arguments (arrays of strings) for changes. '\n' is not a valid
     // argument name character, so it is a valid delimiter here.
@@ -736,7 +736,6 @@ const PROCEDURE_CALL_COMMON = {
     if (paramIds.length !== paramNames.length) {
       throw RangeError('paramNames and paramIds must be the same length.');
     }
-    this.setCollapsed(false);
     if (!this.quarkIds_) {
       // Initialize tracking for this block.
       this.quarkConnections_ = {};
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -390,8 +390,8 @@ const appendPrivate = function(state, workspace, {
   const block = workspace.newBlock(state['type'], state['id']);
   block.setShadow(isShadow);
   loadCoords(block, state);
-  loadExtraState(block, state);
   loadAttributes(block, state);
+  loadExtraState(block, state);
   tryToConnectParent(parentConnection, block, state);
   loadIcons(block, state);
   loadFields(block, state);
--- a/core/serialization/blocks.js
+++ b/core/serialization/blocks.js
@@ -390,8 +390,8 @@ const appendPrivate = function(state, workspace, {
   const block = workspace.newBlock(state['type'], state['id']);
   block.setShadow(isShadow);
   loadCoords(block, state);
-  loadAttributes(block, state);
   loadExtraState(block, state);
+  loadAttributes(block, state);
   tryToConnectParent(parentConnection, block, state);
   loadIcons(block, state);
   loadFields(block, state);
--- a/tests/mocha/serializer_test.js
+++ b/tests/mocha/serializer_test.js
@@ -1674,12 +1674,52 @@ Serializer.Mutations.Procedure.Caller = new SerializerTestCase(
     '</mutation>' +
     '</block>' +
     '</xml>');
+Serializer.Mutations.Procedure.CollapsedProceduresCallreturn = new SerializerTestCase(
+    'CollapsedProceduresCallreturn',
+    '<xml xmlns="https://developers.google.com/blockly/xml">' +
+    '<variables>' +
+    '<variable id="aaaaaaaaaaaaaaaaaaaa">x</variable>' +
+    '</variables>' +
+    '<block type="procedures_defreturn" id="id******************" x="42" y="42">' +
+    '<mutation>' +
+    '<arg name="x" varid="aaaaaaaaaaaaaaaaaaaa"></arg>' +
+    '</mutation>' +
+    '<field name="NAME">do something</field>' +
+    '<comment pinned="false" h="80" w="160">Describe this function...</comment>' +
+    '</block>' +
+    '<block type="procedures_callreturn" id="id1*****************" collapsed="true" x="52" y="52">' +
+    '<mutation name="do something">' +
+    '<arg name="x"></arg>' +
+    '</mutation>' +
+    '</block>' +
+    '</xml>');
+Serializer.Mutations.Procedure.CollapsedProceduresCallnoreturn = new SerializerTestCase(
+    'CollapsedProceduresCallnoreturn',
+    '<xml xmlns="https://developers.google.com/blockly/xml">' +
+    '<variables>' +
+    '<variable id="aaaaaaaaaaaaaaaaaaaa">x</variable>' +
+    '</variables>' +
+    '<block type="procedures_defnoreturn" id="id******************" x="42" y="42">' +
+    '<mutation>' +
+    '<arg name="x" varid="aaaaaaaaaaaaaaaaaaaa"></arg>' +
+    '</mutation>' +
+    '<field name="NAME">do something</field>' +
+    '<comment pinned="false" h="80" w="160">Describe this function...</comment>' +
+    '</block>' +
+    '<block type="procedures_callnoreturn" id="id1*****************" collapsed="true" x="52" y="52">' +
+    '<mutation name="do something">' +
+    '<arg name="x"></arg>' +
+    '</mutation>' +
+    '</block>' +
+    '</xml>');
 Serializer.Mutations.Procedure.testCases = [
   Serializer.Mutations.Procedure.NoMutation,
   Serializer.Mutations.Procedure.Variables,
   Serializer.Mutations.Procedure.NoStatements,
   Serializer.Mutations.Procedure.IfReturn,
   Serializer.Mutations.Procedure.Caller,
+  Serializer.Mutations.Procedure.CollapsedProceduresCallreturn,
+  Serializer.Mutations.Procedure.CollapsedProceduresCallnoreturn,
 ];
 
 Serializer.Mutations.Procedure.Names = new SerializerTestSuite('Names');


BeksOmega:
So sadly we can't change the deserialization order :/ because that would be a breaking change.

Currently attributes are deserialized before extraState, which means that in a `loadExtraState` function, an external developer could be expecting `collapsed`, `disabled`, `x`, `y`, etc to be properly deserialized into the block. If we change the load order, we break that promise, and possibly break people's code. See also [the developer docs](https://developers.google.com/blockly/guides/configure/web/serialization#deserialization_order).

I hate adding if-statements to fix bugs, but do you think that adding an `if (this.rendered)` call around the `setCollapsed` in `setProcedureParameters_` would work? This way setting collapsed to false would be skipped while deserializing, so that this bug doesn't get triggered.

Also, would you be able to add a [new test case](https://github.com/google/blockly/blob/d2329f8cba661999cc7f03fcd6c0196770b05cdc/tests/mocha/serializer_test.js#L1677) for this? You would go to the playground, set up the correct scenario, export it to XML, and then add the stringified XML to that linked file. Unit tests wouldn't have worked for your previous PR because it required dragging to reproduce the bug, but here we can totally add a unit test =)

Thank you again for digging into this!

tweini:
Because I think this is a tip of an iceberg, this bug leads us to, I want to add my point of view:

If the developer relies on the 'collapsed" attributes, he would rely on the mutation is present in the block:

The root cause of this bug is, that the 'collapsed' attribute isn't a mere presentation attribute, it is literally a mutator, which mutates the block. This mutation creates the additional input "_TEMP_COLLAPSED_INPUT".

So, the flip side of this order could also be a developer, which not only relies on the correct "collapse" attribute present, but also on the entire **mutated** state of that block. (Fun fact: Blockly could be seen as a developer who was trapped, as it fails to deserialize correctly.)

So we could interpret that the breaking change was altering the order, as it was deserialzed in the very end in the xml serializer component.

I will review your proposal of fixing the bug. And I will do further investigation, as I hate adding if-statements to fix bugs as much as you do.

Testing is on my agenda. Thanks for nudging me.

BeksOmega:
> The root cause of this bug is, that the 'collapsed' attribute isn't a mere presentation attribute, it is literally a mutator, which mutates the block. This mutation creates the additional input "_TEMP_COLLAPSED_INPUT".

Yeah I've had the same frustration. It is unfortunate that collapsing was implemented by changing the block's inputs, rather than modifying the renderer to have a separate "collapsed" mode/state.

> So we could interpret that the breaking change was altering the order, as it was deserialzed in the very end in the xml serializer component.

I see what you're saying here too, but I don't think it was actually a breaking change. With the XML serializer we didn't actually make any promises about deserialization order. In the past XML was just deserialized in the order that nodes existed in, so developers couldn't rely on things being deserialized before other things. But this required duplicating a lot of extra state if you wanted to be safe, which bloated save file sizes. So we decided to make it explicit with the JSON serializer.

Plus the JSON serilializer was a new feature, so nothing about it could silently break developers' code. If we change the JSON deserialization order now that it's been released, it could.

---

Side note: I would say separate out the logic in `setProcedureParameters_` into multiple functions so that the `setCollapsed` call doesn't get hit during deserialization. This would avoid if-statements. But I might actually be tackling #3725 and other related procedure refactors soon, and I don't want you to have to do duplicate work.

Really appreciate hearing your thoughts. I'll be thinking about this more too.

tweini:
Appreciate your positive feedback. I am absolutely with you to decouple things, because side effects are causing all this sort of bugs.

Dealing with legacy code means thinking like a detective. And this led me to the question:

_Why the setCollapsed(false) (L739) call in the code?_

Removing this call is safe as long as there is no input connected to a variable, while being removed in collapsed state. An open mutator seems to be the precondition, where L739 makes sense, so we could append an else on the existing if at L721?

I tried it and ran the tests ✅ and the bug is gone. So, my assumption is, that L739 is an ancient bug fix, which has been placed too prominently. (Changing the role from detective to historian 😄 )

Out of curiousity: How is L727 reached. The precondition is if (!paramIds) and the comment says "Reset the quarks (a mutator is about to open)."?

Why are these parameters named "quarks"?

BeksOmega:
> _Why the setCollapsed(false) (L739) call in the code?_
> 
> Removing this call is safe as long as there is no input connected to a variable, while being removed in collapsed state. An open mutator seems to be the precondition, where L739 makes sense, so we could append an else on the existing if at L721?
> 
> I tried it and ran the tests ✅ and the bug is gone. So, my assumption is, that L739 is an ancient bug fix, which has been placed too prominently. (Changing the role from detective to historian 😄 )

I think that could definitely work! And from checking out the [git blame](https://github.com/google/blockly/commit/b36a88f42346e09db3cf4dfd4b682f808ca8b787), it does look like an [ancient bug fix](https://github.com/google/blockly/commit/b36a88f42346e09db3cf4dfd4b682f808ca8b787). Sadly it doesn't have any details (eg reproduction steps) attached. I imagine the bug was something like:

1) Create a procedure definition and caller.
2) Collapse the caller block.
3) Add an argument to the procedure definition.
4) Observe how the collapsed caller block is rendered incorrectly somehow?

So if I'd like to check that case before we merge anything.

Given the historical context: Do you think it would work to just remove the `setCollapsed` call entirely?

> Out of curiousity: How is L727 reached. The precondition is if (!paramIds) and the comment says "Reset the quarks (a mutator is about to open)."?

Hmmmm. Grrrrr. It looks like it actually isn't being hit anymore. Seems to be another [ancient bug fix with no docs](https://github.com/google/blockly/commit/e0ebce1b60c3a54075b245a65643a92b5695a31b).

(imho the procedure blocks really need to be rewritten from scratch, but its hard with such a small team)

> Why are these parameters named "quarks"?

Hahaha, good question! I think the idea was that the block is an "atom" that you are breaking down into its component pieces - ie quarks! It is mixing metaphors a bit when changing the block is already called a "mutation" though haha.

BeksOmega:
You know what, I bet the `setCollapsed` call was originally solving for: https://github.com/google/blockly/issues/3784 which has since been fixed.

So I'm leaning more and more toward just removing it.

tweini:
>(imho the procedure blocks really need to be rewritten from scratch, but its hard with such a small team)

Yep. As a tech guy I feel you. But managing technical debt is too complex. Also bugfixing is an investment to reduce technical debt, as the knowledge of system grows and the requirements for a rewrite become pretty clear. As long as the bugfix does real research and investigates root causes. (We definitely do.)

>Removing this call is safe as long as there is no input connected to a variable, while being removed in collapsed state. An open mutator seems to be the precondition, where L739 makes sense, so we could append an else on the existing if at L721?

As I started to name it "removing" let's state for the readers we mean "relocate" the setCollapsed() as mentioned above.

>>Out of curiousity: How is L727 reached. The precondition is if (!paramIds) and the comment says "Reset the quarks (a mutator is about to open)."?

>Hmmmm. Grrrrr. It looks like it actually isn't being hit anymore. Seems to be another [ancient bug fix with no docs](https://github.com/google/blockly/commit/e0ebce1b60c3a54075b245a65643a92b5695a31b).

It really can be safely removed, really deleted, as [the only caller of _setProcedureParameters__](https://github.com/google/blockly/search?q=setProcedureParameters_) makes it impossible to reach the code:

https://github.com/google/blockly/blob/2bf842ac65456e61d8c655daec99f8fbc75d2c29/blocks/procedures.js#L865

no side effects between L865 and L872 

https://github.com/google/blockly/blob/2bf842ac65456e61d8c655daec99f8fbc75d2c29/blocks/procedures.js#L872

So, we don't need to hesitate, but two pairs of eyes (eagle eyes, owl eyes, choose your weapon) are a good base for making decisions.

My next steps would be:

- Change the code with mentioned relocation
- Add a test case
- remove L725 - L728

Maybe it will take some days, maybe not. Stay tuned.

A requirement for the future rewrite we could be: "The Mutator should be able to mutate the call block without expanding it." But here we will definitely run in UX issues 😏 'cause the changes are "invisible", not revertible (as imho the undo isn't active while mutating.)

From a pure technical view it should definitely be possible, as it assures clean code as the rendered view and the data structure behind the block should be clearly separated.

Could be a cool exception from a YAGNI feature 😃

BeksOmega (approved):
LGTM! Thank you for the fix! And thank you for adding those unit tests, they look great =)

PR #6104, opened 2022-04-22
fix: component id should be on the prototype

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/5857

### Proposed Changes

Changes the `id` property on `i_component` to be on the prototype, and then updates things that implement `IComponent` accordingly.

The IDs were already there--created in the constructors for things that implement `IComponent`. `id` was not supposed to be at the top level, since it's not shared between instances of a class.

Also added `dispose` to `IBlockDragger`.

### Reason for Changes

Fix type errors to convert to typescript

### DIFF
--- a/core/drag_target.js
+++ b/core/drag_target.js
@@ -34,6 +34,19 @@ const {Rect} = goog.requireType('Blockly.utils.Rect');
  */
 class DragTarget {
   /**
+   * Constructor for DragTarget. It exists to add the id property and should not
+   * be called directly, only by a subclass.
+   */
+  constructor() {
+    /**
+     * The unique id for this component that is used to register with the
+     * ComponentManager.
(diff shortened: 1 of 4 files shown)


core/drag_target.js
```diff
@@ -33,6 +33,19 @@ const {Rect} = goog.requireType('Blockly.utils.Rect');
  * @alias Blockly.DragTarget
  */
 class DragTarget {
+  /**
+   * Constructor for DragTarget. It exists to add the id property and should not
```
BeksOmega, line 37: If this isn't meant to be constructed directly, should it be tagged `@abstract`?

BeksOmega (approved):

PR #6105, opened 2022-04-22
fix!: fix or silence type errors

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Progress on https://github.com/google/blockly/issues/5857

### Proposed Changes

Fix or silence remaining type errors exposed by enabling 'strictCheckTypes' and 'strictMissingProperties' flags for core code.

All changes resolve type errors. Where possible I made the types correct; when I couldn't, I added an item to track additional work in https://github.com/google/blockly/issues/6097. Most commits fix one error or type of error, so please review one commit at a time.

Things that blocked real fixes:
- Adding properties to SVG Elements
- Generic code
- "Optional" functions that are not part of any interface, and where it's not clear which interface they belong on.

#### Behavior Before Change

~45 errors in core with those flags enabled.

#### Behavior After Change

No errors in core with those flags enabled.

### Reason for Changes

These type errors were blocking conversion to TypeScript because the MigranTS tool wouldn't run with these errors in place.

### Test Coverage
Tested for build errors by enabling 'strictCheckTypes' and 'strictMissingProperties' in `build_tasks.js`.
Tested runtime errors with unit tests and some basic dragging around in the playground.

###
Breaking change
Renames `isBlockCreatable_` (on FlyoutBase) to `isBlockCreatable`. It was and still is `package`.


core/interfaces/i_flyout.js
```diff
@@ -198,4 +198,14 @@ IFlyout.prototype.position;
...
 
+/**
+ * Does this flyout allow you to create a new instance of the given block?
+ * Used for deciding if a block can be "dragged out of" the flyout.
+ * @param {!BlockSvg} block The block to copy from the flyout.
+ * @return {boolean} True if you can create a new instance of the block, false
+ *    otherwise.
+ * @package
+ */
+IFlyout.prototype.isBlockCreatable;
```
BeksOmega, line 209: Is removing the underscore on this name a breaking change?

I still don't understand our policy on external developers overriding `@package` properties in subclasses. But it seems like if this is part of the interface of a flyout, then renaming part of that interface is a breaking chnage.

core/interfaces/i_toolbox_item.js
```diff
@@ -89,4 +89,12 @@ IToolboxItem.prototype.dispose;
  */
 IToolboxItem.prototype.getClickTarget;
 
+/**
+ * Sets whether the category is visible or not.
+ * For a category to be visible its parent category must also be expanded.
+ * @param {boolean} isVisible True if category should be visible.
+ * @protected
+ */
+IToolboxItem.prototype.setVisible_;
```
BeksOmega, line 98: Is this supposed to exist here? Or on the collapsible toolbox item interface?

core/gesture.js
```diff
@@ -743,8 +745,10 @@ class Gesture {
    */
   doBubbleClick_() {
     // TODO (#1673): Consistent handling of single clicks.
-    this.startBubble_.setFocus && this.startBubble_.setFocus();
-    this.startBubble_.select && this.startBubble_.select();
+    if (this.startBubble_ instanceof WorkspaceCommentSvg) {
```
BeksOmega, line 748: Why are workspace comments bubbles? That seems like not what they are.

BeksOmega (approved):

core/interfaces/i_toolbox_item.js
rachel-fenichel: It has to move all the way up to IToolboxItem because of how it's being used. @alschmiedt said that "we allow any toolbox item to be in a collapsible category" and that it should be left as is for compatibility reasons.

core/gesture.js
rachel-fenichel: Notes for the curious:
I think it's just to make workspace comments similar to block comments. And block comments are icons/bubbles, just like mutators.
Bubbles sit on a different canvas than blocks, and always on top. They do that so that blocks on the workspace can never look like they're inside an open mutator.

core/interfaces/i_flyout.js
rachel-fenichel: Technically a breaking change, so I renamed the PR to `fix!`.

rachel-fenichel: And also added a comment on the breaking change to the PR description.

PR #4896, opened 2021-06-11
Moves toolbox cursor styling out of block_dragger

moniika:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

n/a

### Proposed Changes

Refactors logic for applying css style to toolbox out of block_dragger and into the toolbox.

PR #6116, opened 2022-04-27
fix: remove retired "update cursorstyle over toolbox"

tweini:
## The basics

- [x] I branched from develop

## The details

The Toolbox has been refactored and extends DeleteArea. As a consequence, the cursor update has been moved to the Toolbox:

https://github.com/google/blockly/blob/24704e0eff63a7e4bb7407d1c16307f3684ee7d2/core/toolbox/toolbox.js#L665

And there is an incorrect comment in IStyleable, as it refers to Toolbox, but it should refer more generally to "an object".

The latter change is somehow tangled to this issue, so it makes sense to bundle it here.

### Resolves

Unused code causes irritation. (I was irritated, so I took action to do this PR)

### Proposed Changes

There was old code which has been removed.

There are [no known references](https://github.com/google/blockly/search?q=updateToolboxStyle_)

#### Behavior After Change

No change. The "isDeletable()" check has been implemented in toolbox.js with "wouldDelete()", so no indication of lost functionality.

### Test Coverage

Manually tested. Cursor changes when block is dragged over toolbox.

### Documentation

No doc needed, change is trivial.

### Additional Information

Blockly Team is awesome 🤩

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -353,29 +353,6 @@ const BlockDragger = class {
   }
 
   /**
-   * Adds or removes the style of the cursor for the toolbox.
-   * This is what changes the cursor to display an x when a deletable block is
-   * held over the toolbox.
-   * @param {boolean} isEnd True if we are at the end of a drag, false
-   *     otherwise.
-   * @protected
-   */
-  updateToolboxStyle_(isEnd) {
-    const toolbox = this.workspace_.getToolbox();
-
-    if (toolbox) {
-      const style = this.draggingBlock_.isDeletable() ? 'blocklyToolboxDelete' :
-                                                        'blocklyToolboxGrab';
-
-      if (isEnd && typeof toolbox.removeStyle === 'function') {
-        toolbox.removeStyle(style);
-      } else if (!isEnd && typeof toolbox.addStyle === 'function') {
-        toolbox.addStyle(style);
-      }
-    }
-  }
-
-  /**
    * Fire a move event at the end of a block drag.
    * @protected
    */
--- a/core/interfaces/i_styleable.js
+++ b/core/interfaces/i_styleable.js
@@ -25,13 +25,13 @@ goog.module('Blockly.IStyleable');
 const IStyleable = function() {};
 
 /**
- * Adds a style on the toolbox. Usually used to change the cursor.
+ * Adds a style on an object. Usually used to change the cursor.
  * @param {string} style The name of the class to add.
  */
 IStyleable.prototype.addStyle;
 
 /**
- * Removes a style from the toolbox. Usually used to change the cursor.
+ * Removes a style from an object. Usually used to change the cursor.
  * @param {string} style The name of the class to remove.
  */
 IStyleable.prototype.removeStyle;


BeksOmega:
@maribethb It looks like the block dragger is included in the registry (meaning we expect people to create custom subclasses?)  And this function is protected, so does that make removing it technically a breaking change?

BeksOmega:
Just wanted to give you an update on this @tweini @maribethb is out of office this week, so it will be next week before I have info for you on this =)

maribethb:
Sorry I missed this before I went OOO!

Yeah, I guess this would *technically* be a breaking change - if someone registers their own block dragger class and calls this method which will no longer exist. That seems exceedingly unlikely to me, but we're already going to be doing another major release so there is no harm in throwing this up in the breaking changes section of the release notes.

So @tweini could you change the PR title to use the prefix `fix!:` (with the exclamation mark) and then add in a section in the PR description that says something along the lines of "Am I affected by this breaking change? probably not unless someone registers their own block dragger class and calls this method which will no longer exist" so that when people click on this from the release notes they can understand at a glance what has changed? Thanks! 


as a side note we should go through and change some of these methods from `protected` back to `private` - which would be a breaking change, but hopefully saves us in the future from having to make other trivial changes technically "breaking". we swung the pendulum too far when we made everything protected.

maribethb (approved):

tweini:
Hi,

got time for some investigations, because it seemed too hypothetical to me, that this is a breaking change:

The ability to register an own block dragger class given by this PR:

[hypothetical](https://github.com/google/blockly/pull/4893)

The cursor related code was added to toolbox.js with this PR one day later:

https://github.com/google/blockly/pull/4896

So the code which will be removed by my PR was unreachable since then. When someone would have overwritten this method, he would have noticed it doesn't work and he will have moved the changes to the new spot.

So bear with me. But there is a big difference between a hypothetical and a minimal chance.

Or another perspective is, that the code deletion was simply forgotten by #4896 and nobody will have noticed it.

Please don't get me wrong. Stability and Reliability as a goal of Blockly Team is absolutely honorable. But tagging this as a braking change is way too over engineered.

There will be harm in throwing this up in the breaking change section. Everybody who reads this wastes his time for an information which too hypothetical. It's like writing "Look up" because of falling pianos.

maribethb:
I definitely get where you're coming from wrt not wanting this to be a breaking change. However I still think we should mark it as one -

Let's say someone registers their own block dragger and they decide that they want to trigger the toolbox delete style in certain scenarios custom to their application, so in the block dragger they call this method which conveniently already exists for them, called `updateToolboxStyle_`. The fact that we don't ever call that method ourselves is not relevant here. It's that we were previously making that method available for people to use and if anyone does use it we would be breaking them. I would not want to break anyone without telling them that we've done so, hence my caution here. And even though I doubt anyone has called this method, in reality I don't know how others are using the block dragger.

I do understand the cognitive load there is when reading release notes for breaking changes, so that's why I want to add a section to the PR description to make it clear who has to care about this or not. We have similar other changes where it's only under a specific circumstance that a change would break anyone, but we still want to err on the side of caution.

BeksOmega:
I'm going to go ahead and close this one. If you're still interested in working on it, feel free to reopen!

PR #6122, opened 2022-04-28
feat: creates a new way to add shortcut key mappings

alschmiedt:
## The basics

- [x] I branched from develop

### Resolves
While I was writing a document talking about the different ways to customize Blockly I really wanted to say, "for the shortcut and context menu registries you simply create an object with all the necessary information and then register it." But that was not true for keyboard shortcuts because developers have to call `addKeyMapping` after registering their object. This was very annoying and since the fix is completely backwards compatible, this fixes that.

### Proposed Changes
- Add the ability to add a list of key codes to the shortcut object.
- Updates shortcut items to use this instead of `addKeyMapping`.
- Add the ability to allowCollisions from the shortcut object.

#### Behavior Before Change
In order to add a key code for a shortcut, developers would have to do:
```
ShortcutRegistry.registry.addKeyMapping('46', copyShortcut.name);
```

#### Behavior After Change
Now developers can do:
```
const copyShortcut = {
  ...
   keyCodes: ['46']
  ...
};
```
### Reason for Changes
This is cleaner in most situations.

Note: This does not cover the keyboard navigation use case where developers will want to update and change key mappings based on user input. In that case they would still need to use `addKeyMapping` in that situation.


BeksOmega (approved):
LGTM! Does this need docs changes as well? If so, could you file an issue for that?

alschmiedt:
Ok I added #6163 to add better documentation for shortcuts. I also updated the keyboard navigation codelab with this change. A PR for that is coming soon.

Issue #5914, opened 2022-02-02
Most uses of `object.mixin` should probably use `Object.assign` instead


cpcallen:
**Describe the bug**

There appear to be many places in the codebase where we use `object.mixin` (from `Blockly.utils.object`) when it would probably be better to use `Object.assign`.

Here's the implementation of `object.mixin`:

```JS
const mixin = function(target, source) {
  for (const x in source) {
    target[x] = source[x];
  }
};
```

Note that this will copy all enumerable properties from `source` to `target`, _including any inherited enumerable properties_.

In comparison, `Object.assign` copies only enumerable _own_ properties from the target object(s).

Since in most cases the source object is likely to be a simple string-keyed map, it is probably not normally the case that we intend or want to copy any inherited `enumerable` properties (including, for instance, any enumerable properties that an unwary developer might create on `Object.prototype`).

I therefore propose that uses of `object.mixin` be audited and, unless there is evidence that we _want_ to copy inherited properties, the call replaced with a call to `Object.assign` instead.

(I suspect this will result in 0 remaining uses of `object.mixin`, which can then be deprecated and removed.)

derwehr:
Hi, I'd like to get started and submit a PR for this issue, if no one is already working on it

BeksOmega:
> Hi, I'd like to get started and submit a PR for this issue, if no one is already working on it

Hello! Thank you for your interest :D Sadly at the moment we're working on a conversion to TypeScript, which may make this issue irrelevant.

Where we would really appreciate help (if you're interested) is actually in the samples repository, which adds cool new plugins and features to Blockly. Here is a [list of the help-wanted issues](https://github.com/google/blockly-samples/labels/help%20wanted).

If there's anything we can do to help you get started, don't hesitate to ask!

cpcallen:
Closed by #6138.

PR #6138, opened 2022-05-04
fix: replace `object.mixin` with `Object.assign`

derwehr:
## The basics

- [x] I branched from develop

### Resolves

#5914

### Proposed Changes

Replaces all uses of `object.mixin` with `Object.assign`

### Reason for Changes

see #5914

### Additional Information

Things I didn't do, because I was unsure if they're needed or wanted:
- update any documentation
- declare `object.mixin` as deprecated


BeksOmega (changes requested):
Thank you so much for posting this! And thank you for removing the `goog.require` calls as well.

I think we should definitely deprecate the `object.mixin` function, so that we don't accidentally use it in the future. [Here](https://github.com/google/blockly/blob/2bf842ac65456e61d8c655daec99f8fbc75d2c29/core/workspace_svg.js#L1719) is a good example of how to deprecate something.

1) Add an `@deprecated` tag recording what to use instead, and when the function was deprecated.
2) Add a call to [Blockly.utils.deprecation.warn](https://developers.google.com/blockly/reference/js/Blockly.utils.deprecation#.warn) with similar info. The `deletionDate` should be May 2023. 

Thank you again for posting this fix! We really appreciate it =)

BeksOmega (approved):
Perfect! Thank you again :D

Issue #6136, opened 2022-05-02
Add JSON serialization hooks to the lists_getIndex block


BeksOmega:
**Describe the bug**

One of the validators on the lists_getIndex block changes whether the block is a stack block (with next and previous connections) or a  value block (with an output connection). This information needs to be stored in a mutator so that the block is in the proper state before being connect to its parent block. See the [deserialization order documentation](https://developers.google.com/blockly/guides/configure/web/serialization#deserialization_order) for more info.

I didn't realize this originally, so I didn't add JSON hooks when updating the lists_getIndex block.

**Expected behavior**

The block should have a JSON serialization hook including data about whether it is a stack or value block.

**Additional context**

Currently the block serializes stringified XML as its extra state. When adding JSON hooks, we need to be careful to make sure it can load this data. I think we can just do an `if (typeof value === 'string')` check, but there should be a test for it.

If it isn't possible to add JSON hooks, then the comment [here](https://github.com/google/blockly/blob/7eebd78b5dbf0195f08cd817716bd802753e0216/blocks/lists.js#L449) should be removed and replaced with a more accurate comment.

This was originally reported by a lovely forum person [here](https://groups.google.com/g/blockly/c/4otFlbGCSNk).

gregdyke:
This looks like a good first issue for me, but I need a bit of clarification

- The block currently works, right? The difference will be that after, the serialized version will be a json object (faster to process and empty by default to save space?) rather than `"<mutation statement=\"true\" at=\"false\"></mutation>"`
- Could you clarify what is meant by "If it isn't possible to add JSON hooks, then the comment here should be removed and replaced with a more accurate comment.". In what case might it not be possible to add JSON hooks?

gregdyke:
Actually, while I'm asking questions, I realise I don't understand how/why the domToMutation xml is serialized in the JSON? I can't find a default saveExtraState implementation that calls domToMutation. And yet in the forum post, the author reports their json as including the mutation xml in the extra state?

cmjcharlton:
Looking at the code in /core/serialization/blocks.js the function saveExtraState:

https://github.com/google/blockly/blob/b92cbd3b915ad99659ed2ad58adf1a4950fbbf79/core/serialization/blocks.js#L168

falls back to saving a stringified version of the XML if saveExtraState() isn't defined for a block but mutationToDom() is.

When the block is loaded the function loadExtraState:

[...]

PR #6170, opened 2022-05-22
fix: json serialize lists_getIndex with json extraState (#6136)

gregdyke:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/6136

### Proposed Changes

- Add JSON serialization hooks to lists_getIndex (backward compatible with domMutation that was previously serialized)
- Move tests of blocks into mocha/blocks so that they are more easy to find

#### Behavior Before Change

- JSON serialization of lists_getIndex contained extraState as an xml mutation, serialized from domToXml
- Serialised JSON for blocks with extraState stored as an xml mutation is not automatically forward compatible when saveExtraState is added 

#### Behavior After Change

- JSON serialization hook for lists_getIndex to store extraState as a JSON object with an isStatement property
- If the lists_getIndex is not a statement, extraState is null
- JSON previously serialized with xml extraState is loaded via domToXml explicitly in lists_getIndex loadExtraState

### Reason for Changes

- It's more efficient in time and space to serialise the needed extra state as an object
- Only isStatement:true is needed (and only technically if there is a previous connection) as it's the only out of order deserialization
- There is no obligation for serialized extraState to be a json object, so we did not introduce generic handling of backward compatibility, preferring to duplicate explicit calling of domToXml

### Test Coverage

Added lists_test.js, covering
  - Basic (de)serialization
  - (De)serialization when there is a previous block (and loadExtraState is needed)
  - Backward compatible deserialization when there is a previous block stored as xml (and mutationToDom is needed)

### Documentation

It would probably be good to mention the backward compatibility strategy in the JSON serialization documentation, pointing to this as an example


core/serialization/blocks.js
```diff
@@ -441,14 +441,17 @@ const loadAttributes = function(block, state) {
...
+ * backward compatibility.
  * @param {!Block} block The block to set the extra state of.
  * @param {!State} state The state object to reference.
  */
 const loadExtraState = function(block, state) {
   if (!state['extraState']) {
     return;
   }
-  if (block.loadExtraState) {
+  if (block.loadExtraState && typeof state['extraState'] === 'object') {
```
BeksOmega, line 454: Could you move this check into the `loadExtraState` property of the `lists_getIndex` block instead?

Technically we don't require that the extraState be an object. So if someone has a custom block, where they've saved the state as a string, then this would break their implementation.

That does mean that if we have to do this upgrade for any other blocks, we will have to duplicate a bit of logic (which your current implementation avoids). But I still think moving this check is the better option for backwards compatibility reasons.

blocks/lists.js
```diff
@@ -452,10 +452,31 @@ blocks['lists_getIndex'] = {
...
+   * @return {?{isStatement: boolean}} The state of this block, ie whether it's
+   *     a statement.
+   */
+  saveExtraState: function() {
+    if (!this.outputConnection) {
+      return {
+        'isStatement': true,
+      };
+    }
+    return null;
```
BeksOmega, line 467: I appreciate the intelligent serialization you're doing here to avoid serializing unnecessary state :D Makes me so happy!

tests/mocha/lists_test.js
```diff
@@ -0,0 +1,113 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
BeksOmega, line 3:
```suggestion
 * Copyright 2022 Google LLC
```

Also, could you give this file a more descriptive name eg `list_blocks_test`?

tests/mocha/lists_test.js
```diff
@@ -0,0 +1,113 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+goog.module('Blockly.test.lists');
+
+const {runSerializationTestSuite} = goog.require(
+  'Blockly.test.helpers.serialization'
+);
```
BeksOmega, line 11: [I don't think you need to line wrap imports](https://google.github.io/styleguide/jsguide.html#es-module-imports). Can you try unwrapping and see if eslint is still happy?

BeksOmega (changes requested):
Overall looks good, and is backwards compatible to boot! Plus, tests!!!

One comment about backwards compatibility/organization, and a few nits =)

core/serialization/blocks.js
gregdyke: I've considered some alternatives (block having a method that says whether to use loadExtraState or domToMutation, throwing an exception in loadExtraState that triggers domToMutation being used).

None of them seem worth it, so I'll cope with some duplicated logic :)

blocks/lists.js
gregdyke: yay (I read your documentation! :))

tests/mocha/lists_test.js
```diff
@@ -0,0 +1,113 @@
...
+const {sharedTestSetup, sharedTestTeardown} = goog.require(
+  'Blockly.test.helpers.setupTeardown'
+);
```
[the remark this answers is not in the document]
gregdyke: I copy pasted from the logic_ternary_test

gregdyke:
Hi @BeksOmega 

Thanks for the quick review!

I addressed the backward compatibility strategy - I don't think there's a good generic way of doing it (I was forward thinking to fixing the other list blocks and even then, there isn't really a way)

I'm an irrational TDDer (I will generally try to do it regardless of evidence of whether it's a good thing or not :D). So I was glad to write tests (I'm proudest of going out of my way to use the test statement_block).

To be honest I found it quite tricky to figure out what gets tested where. I wonder if all the tests of blocks (blocks_ternary, possibly procedures and variables?) could go in a blocks folder (also should we add the implicit import of the block they are actually testing? It's a bit confusing to open a test file and there be no link to the actual source it is testing). That's also why I think sticking to lists_test.js is better. But if you're still not convinced, I'm happy to go with lists_blocks_test.js or block_lists_test.js?

I'm not quite sure about the commit message practices, so let me know if multiple commits in the PR is ok + whether the followup commits need better messages?

BeksOmega:
Ah ok, I see your point about the test file naming. Wrt to moving the tests into a blocks folder, do you mean a blocks folder within the mocha folder? That sounds like a fine idea to me =) But if it's going to complicate this PR, feel free to leave as is.

Wrt commit messages, it's a bit confusing, but we have information here. Basically if your PR only has a single commit, that commit needs to follow conventional commit structure. If your PR has multiple commits, then the title just needs to follow conventional commit structure.

I also noticed I forgot to say thank you in my previous review. So I'd like to say that I, and the team, are sincerely thankful for your contributions =) You rock!

Best wishes, Beka

BeksOmega (approved):
Thanks again for the contribution, and for the back-and-forth! Really appreciate it :D LGTM!

Issue #6176, opened 2022-05-26
Lists blocks serialize unnecessary xml mutations in JSON hooks


gregdyke:
The following blocks get their xml mutations serialized because that is the default behaviour without an explicit saveExtraState returning null. It is either unnecessary (or would be more efficiently stored as JSON, but I think unnecessary covers all the following)

- lists_setIndex
- lists_getSublist
- lists_split

The same applies to all blocks with the comment - could be addressed in this issue or separate issues?
```
    This block does not need JSO serialization hooks 
```

- procedures.js
- math.js
- text.js

**To Reproduce**

JSON Serialize any of the above blocks

**Expected behavior**

extraState in the form of a mutation should not be serialized (if necessary it should be serialized as json - it's likely it's not necessary)

BeksOmega:
Sounds awesome! Thank you for your interest in contributing more :D

Multiple PRs associated with this one issue sounds rad. Feel free to include the other blocks (math, text) in this issue as well. I'd recommend waiting on the `procedures_ifReturn` block though, because we might be messing with the procedures next quarter, and I don't want you to do work that gets written over hehe.

Thank you again! If you have any thoughts/questions always happy to discuss =)

BeksOmega:
Fixed by #6177

PR #6177, opened 2022-05-27
fix: json hooks for lists blocks not needing extra state

gregdyke:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/6176

### Proposed Changes

Don't serialize xml mutations to extra state in json serialization for

- lists_setIndex
- lists_getSubList
- lists_split

#### Behavior Before Change

xml mutations were serialized in json extra state and deserialized

#### Behavior After Change

xml mutations no longer serialized and are ignored on deserialization

### Reason for Changes

- not needed in JSON serialization
- better (de-)serialization performance

### Test Coverage

Added basic roundtrip (de)serialization tests, including with previous extra state

### DIFF
--- a/tests/mocha/blocks/lists_test.js
+++ b/tests/mocha/blocks/lists_test.js
@@ -106,4 +106,89 @@ suite('Lists', function() {
     ];
     runSerializationTestSuite(testCases);
   });
+
+  /**
+   * Test cases for serialization where JSON hooks should have null
+   * implementation to avoid serializing xml mutations in json.
+   * @param {!Object} serializedJson basic serialized json
+   * @param {!string} xmlMutation xml mutation that should be ignored/not reserialized in round trip
+   * @return {Array<SerializationTestCase>} test cases
+   */
+  function makeTestCasesForBlockNotNeedingExtraState_(serializedJson, xmlMutation) {
+    return [
+      {
+        title: 'JSON not requiring mutations',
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
Sorry for the delay, this looks great :D Thank you for the fix! I hate having ugly XML in the JSON. Tests look beautiful =)

tests/mocha/blocks/lists_test.js
```diff
@@ -106,4 +106,89 @@ suite('Lists', function() {
...
+        json: serializedJson,
+        assertBlockStructure: (block) => {
+          chai.assert.equal(block.type, serializedJson.type);
+        },
+      },
+      {
+        title:
+          'JSON with XML extra state',
+        json: {
+          ...serializedJson,
```
cpcallen, line 130: I was confused about why a build of a clean checkout of `develop` was resulting in changes to `tests/deps.mocha.js`.  It turns out that this use of `...` meant that `lists_test.js` went from being ES6 to ES2018, and `closure-make-deps` tries to update `testsd/deps.mocha.js` accordingly, but that that update was not included in the same commit.

Not worth fixing now, since PR #6205 is about to delete the deps files from the repo, but a useful reminder to us all to keep an eye out for changes to generated files and try make sure they get checked in at the same time as the change which caused them least we confuse our colleagues…

Issue #6169, opened 2022-05-18
code generated from combination of math_on_list and lists_create_empty blocks produces errors


nederland074:
**Describe the bug**

The code generated by "lists_create_empty" connected to "math_on_list" causes errors if the "math_on_list" block has its "operation" field set to "sum" or "average".

**To Reproduce**

1. Go to https://blockly-demo.appspot.com/static/demos/code/index.html#no6a4p or https://blockly-demo.appspot.com/static/demos/code/index.html#3braq6
2. Hit the run button.
3. See the popup window containing error message: "Program error: TypeError: reduce of empty array with no initial value"

**Expected behavior**

The code runs without any errors.

**Screenshots**

Both following block setups cause this error:
![blocks](https://user-images.githubusercontent.com/11175695/169037936-7608ddfb-4b23-4236-8701-1c4f58bd548f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Google Chrome 101.0.4951.64

BeksOmega:
Thanks for reporting this @nederland074 !

Extra repro details: Tested in all generators, and this only occurs with the JavaScript generator.

gregdyke:
If you put this as help wanted, it's likely it will be something I can take a look at fixing

rachel-fenichel:
Is this fixed?

maribethb:
fixed by #6178

PR #6178, opened 2022-05-29
fix: provide initial values to reduce functions in generated js

gregdyke:
## The basics

- [x] I branched from develop

### Resolves

https://github.com/google/blockly/issues/6169

### Proposed Changes

Generated code using reduce (average and sum) has initial value set to 0

#### Behavior Before Change

Runtime error when calling average or sum on an empty list in javascript

#### Behavior After Change

No runtime error

### Reason for Changes

Reduce without an initial value takes the 1st element in the array as the initial value in javascript -> needs an initial value for empty lists.

Note
- mean of an empty list still produces NaN (expected)
- std dev returns null for empty lists already so does not need an initial value passed to reduce

### Test Coverage

Adjusted existing tests to new generated code

### DIFF
--- a/generators/javascript/math.js
+++ b/generators/javascript/math.js
@@ -222,7 +222,7 @@ JavaScript['math_on_list'] = function(block) {
     case 'SUM':
       list = JavaScript.valueToCode(block, 'LIST',
           JavaScript.ORDER_MEMBER) || '[]';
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
LGTM! Thank you for the fix, and for updating tests :D

PR #6183, opened 2022-06-01
chore: update PR template for test requirements

BeksOmega:
## The basics

- [ ] I branched from develop

Note code! Needs to go into master.

### Resolves

N/A

### Proposed Changes

Makes the pull request template explicit about what testing information we want.

### Reason for Changes

Makes it easier for us to manually test things before release to prevent bugs. And makes it easier for contributors to provide information about what to test.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -35,10 +35,22 @@
 
 ### Test Coverage
 
-<!-- TODO: Please show how you have added tests to cover your changes,
-  -        or tell us how you tested it. For each systems you tested,
-  -        uncomment the systems in the list below.
+<!-- TODO: Please do one of the following:
+  -    * Create unit tests, and explain here how they cover your changes.
+  -    * List steps you used for manual testing, and explain how they cover
+  -      your changes.
+  -      For each system you manually tested, uncomment it below.
   -->
+
+<!-- I created tests for ... -->
+
+<!-- Steps you used to test:
+  -   1. Go to '...'
+  -   2. Click on '...'
+  -   3. Scroll down to '...'
+  -   4. See error
+  -->
+
 
 <!-- Tested on: -->
 <!-- * Desktop Chrome -->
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -35,11 +35,23 @@
...
+<!-- I created tests for ... -->
+
+<!-- Steps you used to test:
+  -   1. Go to '...'
+  -   2. Click on '...'
+  -   3. Scroll down to '...'
+  -   4. See error
+  -->
+
+
```
maribethb, line 54: I think you should consider removing this long list of browsers, nobody (including team members) actually uses it and the more things a person has to delete from the template, the less likely they are to actually fill out the pieces we want them to. Asking for the steps should be enough imo and it sounds like people agreed at our team meeting.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -35,11 +35,23 @@
...
-  -        or tell us how you tested it. For each systems you tested,
-  -        uncomment the systems in the list below.
+<!-- TODO: Please do one of the following:
+  -    * Create unit tests, and explain here how they cover your changes.
+  -    * List steps you used for manual testing, and explain how they cover
+  -      your changes.
+  -      For each system you manually tested, uncomment it below.
   -->
 
+<!-- I created tests for ... -->
```
maribethb, line 45: I don't know what you expect me to write here, can you either update this to be more specific or remove it? And similar to the below comment, I think we want to minimize the amount of things a person deletes or ignores from the template so consider removing this unless it's adding a lot of value.

maribethb (approved):

.github/PULL_REQUEST_TEMPLATE.md
BeksOmega: Sounds good. We will have to clarify for Divya which systems to test on. But that's probably better than just listing whatever someone happened to test with anyway.

PR #6203, opened 2022-06-09
fix!: remove backwards compatible get and set accessors

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Removes some get and set accessors (most of which were deprecated). 
- Blockly.alert
- Blockly.prompt
- Blockly.confirm
- Blockly.mainWorkspace
- Blockly.selected
- Blockly.HSV_SATURATION
- Blockly.HSV_VALUE
- Blockly.ContextMenu.currentBlock
- Blockly.Events.recordUndo
- Blockly.Tooltip.visible
- Blockly.Tooltip.DIV
- Blockly.WidgetDiv.DIV

#### Behavior Before Change

These accessors were available.

#### Behavior After Change

These accessor are not available.

### Reason for Changes

As we move to typescript, it's no longer possible to provide accessors directly on the exports object, so these have to be removed.

### Test Coverage

N/A

### How to Migrate

You can run the [blockly/migrate](https://www.npmjs.com/package/@blockly/migrate) script to automatically perform renamings in your files. If you are setting any of these values, you may need to do some manual edits to properly call the setter function.


cpcallen (commented):
This might be a good subject for a team discussion, but I lean towards leaving the ones in `blockly.js` alone until we are actually ready to convert that file, which we may find we do not want to do until a subsequent release as it entails some (drafted but not finished) changes to the UMD wrappers.

Actually, this brings up another point: we could monkey-patch in these accessors onto the malleable compiled ES5 code (doing this in one of the NPM package wrappers), allowing them to continue to work even though this would not be possible in uncompiled mode.

BeksOmega:
I think you're the only one with enough context to be able to write up alternatives and add it to the meeting @cpcallen

cpcallen:
Closing this as obsolete.  It was helpful, though, to make sure I didn't miss anything in #6260.

PR #6205, opened 2022-06-10
refactor(build): Preparation for building TypeScript

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #5892.

### Proposed Changes

* Update Closure Library `base.js` and also bring in `goog.js`.
* Move `deps.js` and `deps.mocha.js` from `test/` to `build/`.
* Always run `tsc` on `core/` and then run Closure Compiler against the output in `build/src/`.
* Change the `npm run build:deps` target to run `tsc` before `closre-make-deps`.
* General cleanup of `build_tasks.js` 

#### Behaviour Before Change

Build pipeline only runs `tsc` if you specify `--compileTs`.

#### Behaviour After Change

Build pipelines always runs `tsc`.

### Reason for Changes

Preparation for conversion to TypeScript.

### Test Coverage

* Passes `npm test`.
* Playground loads and appears fine.

No changes to manual testing expected.

### Additional Information

Note that although the build pipeline can now ingest `.ts` files, this PR does not included needed changes to the uncompiled loading mechanism used by the playgrounds in order to enable them to load the resulting ES modules.


gulpfile.js
```diff
@@ -30,7 +30,8 @@ module.exports = {
   buildLangfiles: buildTasks.langfiles,
   buildCompiled: buildTasks.compiled,
   buildAdvancedCompilationTest: buildTasks.advancedCompilationTest,
-  buildTs: buildTasks.buildTypescript,
+  buildJavaScript: buildTasks.javaScript,
+  buildJavaScriptAndDeps: gulp.series(buildTasks.javaScript, buildTasks.deps),
```
BeksOmega, line 34: Is there any reason to build javascript without building deps first? Do we need to export all three of these thigns, or can it be simplified?

package.json
```diff
@@ -42,6 +42,7 @@
     "license": "gulp checkLicenses",
     "lint": "eslint .",
     "package": "gulp package",
+    "prepare": "gulp buildJavaScriptAndDeps",
```
BeksOmega, line 45: If this is just an alias of `npm run build:deps` could you make it that instead?

package.json
```diff
@@ -24,9 +24,9 @@
     "build-strict-log": "npm run build:strict > build-debug.log 2>&1 && tail -3 build-debug.log",
     "build:compiled": "gulp buildCompiled",
     "build:compressed": "npm run build:compiled",
-    "build:deps": "gulp buildDeps",
+    "build:deps": "gulp buildJavaScriptAndDeps",
+    "build:js": "gulp buildJavaScript",
```
BeksOmega, line 28: Is this just for testing? If so, is it necessary to have it be part of the `package.json`?

scripts/gulpfiles/build_tasks.js
```diff
@@ -25,13 +25,18 @@ var closureDeps = require('google-closure-deps');
...
 var {getPackageJson} = require('./helper_tasks');
 
 ////////////////////////////////////////////////////////////
 //                        Build                           //
 ////////////////////////////////////////////////////////////
 
+/**
+ * Directory in which core/ can be found after passing through tsc.
+ */
+const CORE_DIR = path.join(TSC_OUTPUT_DIR, 'core');
```
BeksOmega, line 38: Is there a reason this is in this file and not in the config?

scripts/gulpfiles/config.js
```diff
@@ -18,18 +18,27 @@ var path = require('path');
...
-  TYPINGS_BUILD_DIR: path.join('build', 'typings'),
+// Dependencies file (for blockly_uncompressed.js):
+exports.DEPS_FILE = path.join(exports.BUILD_DIR, 'deps.js');
 
-  // Directory where typescript compiler output can be found.
-  // Matches the value in tsconfig.json: outDir
-  TSC_OUTPUT_DIR: path.join('build', 'ts'),
-};
+// Dependencies file (for blockly_uncompressed.js):
+exports.TEST_DEPS_FILE = path.join(exports.BUILD_DIR, 'deps.mocha.js');
```
BeksOmega, line 33: How does outputing these to the `build` directory relate to `build:checkin`? We no longer want them to be checked in by default?

BeksOmega (commented):
Generally looks good! Just lots of questions hehe

scripts/gulpfiles/chunks.json
```diff
@@ -9,265 +9,265 @@
...
+    "./build/src/core/variables.js",
+    "./build/src/core/utils/object.js",
+    "./build/src/core/events/events_abstract.js",
+    "./build/src/core/registry.js",
+    "./build/src/core/events/utils.js",
+    "./build/src/core/xml.js",
+    "./build/src/core/connection.js",
+    "./build/src/core/common.js",
+    "./build/src/core/blocks.js",
     "./closure/goog/base_minimal.js",
```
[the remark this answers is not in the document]
cpcallen: The change to compiling with `base.js` instead of `base_minimal.js` will be in the next PR (if I don't figure out a way to avoid it!)

gulpfile.js
cpcallen: As noted: `tsc` is relatively fast compared to `closure-make-deps`, and needs to be run after every `.ts` file change, while the latter is only needed after changes to imports.

package.json
cpcallen: Your question suggests that you think we should change this to
```
"prepare": "npm run build:deps",
```
and I was going to do that but upon reflection I sort of think that "buildJavaScriptAndDeps" is slightly more clear about what will actually be prepared.  What do you think?

BeksOmega: If you always want "prepare" to do the exact same thing "build:deps" does, then I think it should be an alias.

If "prepare" actually means something different from "build:deps" it just so happens that right now you want them to do the same thing, then I'm fine with having both commands have the same value.

I just want to make sure that neither command becomes broken cruft, because something gets changed in one place and not the other.

scripts/gulpfiles/build_tasks.js
```diff
@@ -270,12 +285,13 @@ function buildDeps(done) {
...
-  execSync(`closure-make-deps ${args} > tests/deps.js`, {stdio: 'inherit'});
+  execSync(`closure-make-deps ${args} > ${DEPS_FILE}`,
+           {stdio: 'inherit'});
 
   // Use grep to filter out the entries that are already in deps.js.
   const testArgs = testRoots.map(root => `--root '${root}' `).join('');
-  execSync(`closure-make-deps ${testArgs} | grep 'tests/mocha'` +
-      ' > tests/deps.mocha.js', {stdio: 'inherit'});
+  execSync(`closure-make-deps ${testArgs} | grep 'tests/mocha' ` +
+      `> '${TEST_DEPS_FILE}'`, {stdio: 'inherit'});
```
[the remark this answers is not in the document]
cpcallen: The single quotes protect against spaces in `TEST_DEPS_FILE`.  (Really I ought to also escape any single quotes and backslashes before interpolating too, but as this particular code is probably not _actually_ going to be subject to injection attacks I am willing to chance it.)  You are correct that `DEPS_FILE` should get quotes too.  Fixed.

scripts/gulpfiles/config.js
cpcallen: I was thinking you were asking if we were wanting to check in `deps*.js`, which I can confirm that (IMHO) we don't, and so `checkin:built` is not relevant, since it just checks in the `*_compressed.js` files.

But now I realise you might be trying to point out that it will _also_ unintentionally copy the `deps*.js` files out of `build/` (and, umm, also all of `build/src/` too, just for bad measure).  Ooops.

Though to be sure, "hey, do you realise that `checkin:built` will copy the deps files too?" would have been helpfully blunt in this case, because I nearly overlooked this despite your efforts.  😬

cpcallen: OK, fixed.

scripts/gulpfiles/build_tasks.js
```diff
@@ -672,11 +688,10 @@ const build = gulp.parallel(
...
   return gulp.src([
-    `${BUILD_DIR}/**.js`,
-    `${BUILD_DIR}/**.js.map`,
-    `${BUILD_DIR}/**/**.js`,
-    `${BUILD_DIR}/**/**.js.map`,
-  ]).pipe(gulp.dest('.'));
+    `${BUILD_DIR}/*_compressed.js`,
+    `${BUILD_DIR}/*_compressed.js.map`,
+    `${BUILD_DIR}/msg/js/*.js`,
+  ], {base: BUILD_DIR}).pipe(gulp.dest('.'));
```
BeksOmega, line 694: Does this put the messages correctly in the msg folder? Or does it dump all of them into the root directory?

BeksOmega (approved):
LGTM after all comments are resolved =)

package.json
cpcallen: > If you always want "prepare" to do the exact same thing "build:deps" does, then I think it should be an alias.
> 
> If "prepare" actually means something different from "build:deps" it just so happens that right now you want them to do the same thing, then I'm fine with having both commands have the same value.

I believe it is the latter.  At the moment, "build:deps" happens to do the things that I think we need to do to prepare to work in the Blockly repository (after installing npm packages, of course).  But this is mostly a coincidence;  in the future we might want `prepare` to do the same thing as `build`, or at least to additionally do `build:langfiles`.

> I just want to make sure that neither command becomes broken cruft, because something gets changed in one place and not the other.

Yes, very sensible.

scripts/gulpfiles/build_tasks.js
cpcallen: A question I learned the answer to the hard way (oh look my Blockly directory is suddenly full of spam!), so now the `{base: BUILD_DIR}` option is there to tell `gulp.dest` not to trim off anything more than `build/` from the output path for (any of, but particularly) the messages files.

PR #6184, opened 2022-06-01
fix: loading messages from script tags.

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

Fixes https://github.com/google/blockly/issues/6123, fixes https://github.com/google/blockly-samples/issues/1066 and fixes https://github.com/google/blockly-samples/issues/1065

Same as #6060, but hopefully this also works for requiring multiple CJS modules.

### Proposed Changes

See https://github.com/google/blockly/issues/6123#issuecomment-1144197138

Also changed the generated, but unwrapped, msg.js files to write to a new `messages` object, rather than `Blockly.Msg` to be clearer about what we're doing, and hopefully prevent errors in the future.

#### Behavior Before Change

Message loading from script tags does not work.

#### Behavior After Change

Message loading from script tags works.

### Reason for Changes

See https://github.com/google/blockly/issues/6123#issuecomment-1144197138

### Test Coverage

* Playground
* Playground after changing `prepare.js` to pull the `dist/msg/es.js` file instead of `msg/messages.js`
* Code demo in core as it currently exists.
* continuous-toolbox test page
* continuous-toolbox test  page after changing it to `import Fr from 'blockly/msg/fr'` and `Blockly.setLocale(Fr)`
* Mirror demo in samples
* Plane demo in samples
* React demo in samples
* React demo in samples after importing spanish and french, and setting the locale to spanish then french (so french translations are shown)

I couldn't find anything that actually uses CJS modules to be able to test.

### Documentation

N/A

### Additional Information

Will undraft after testing.

PR #6210, opened 2022-06-13
fix: message loading

BeksOmega:
Patch containing #6184 and rebuilt message files.


maribethb (approved):

maribethb:
@BeksOmega The rebuilt files were never updated in develop, if you have time could you make this change in develop as well? It makes master and develop different in confusing ways (develop is behind master)

Issue #6135, opened 2022-05-02
The append of a Dummy Input to a Block doesn't work


CualeMarke:
Hi, I'm using a block with a FieldDropDown and a Validator.
Every time the value of the dropdown changes I need to modify my block, so I can chose to show an input value, an input statement or a dummy input.
I've found that every time I try to append a Dummy input to the block, an error is displayed in the console and the input isn't appended.
The append of Value input and Statement input doesn't have any problem, it works perfectly.

To reproduce you can apply the code in the documentation (https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#creating_a_dropdown_validator), but instead of append a value/statement input you should append a dummy input.

Expected behaviour: the dummy input is added without any error

The displayed error is:
![image](https://user-images.githubusercontent.com/104019510/166258046-fdaa1a37-bb09-4c87-8590-7b0097773f6d.png)

I've tested the issue on:
 - OS: Windows 10
 - Browser: Google Chrome, Firefox, Opera Gx Corner

**Stack Traces**

```
Uncaught TypeError: Cannot read properties of undefined (reading 'width')
    at InputRow.module$exports$Blockly$blockRendering$InputRow.InputRow.measure (Ez?_t=1651501442:772:219)
    at RenderInfo.module$exports$Blockly$blockRendering$RenderInfo.RenderInfo.computeBounds_ (Ez?_t=1651501442:873:145)
    at RenderInfo.module$exports$Blockly$blockRendering$RenderInfo.RenderInfo.measure (Ez?_t=1651501442:856:159)
    at Renderer.module$exports$Blockly$blockRendering$Renderer.Renderer.render (Ez?_t=1651501442:892:352)
    at BlockSvg.module$exports$Blockly$BlockSvg.BlockSvg.render (Ez?_t=1651501442:1088:282)
    at BlockSvg.module$exports$Blockly$BlockSvg.BlockSvg.appendInput_ (Ez?_t=1651501442:1080:183)
    at BlockSvg.module$exports$Blockly$Block.Block.appendDummyInput (Ez?_t=1651501442:975:251)
    at updateShape (Ez?_t=1651501442:4928:19)
    at BlockSvg.onchange (Ez?_t=1651501442:4671:25)
    at WorkspaceSvg.module$exports$Blockly$Workspace.Workspace.fireChangeListener
```

BeksOmega:
[...]

PR #6211, opened 2022-06-14
fix: Check for undefined value in InputRow.measure

Zauzolkov:
## The basics

- [x] I branched from develop

### Resolves

N/A (no GitHub issue was opened)

### Proposed Changes

- Check for possible `undefined` value in `measure` method of `Blockly.blockRendering.InputRow` class.

#### Behavior Before Change

- There is no check for `undefined` value in `measure` method of `Blockly.blockRendering.InputRow` class.
- Execution is interrupted with error if there is `undefined` value in `this.elements` array (because of attempt to access a non-existent property): `Uncaught TypeError: elem is undefined`.

#### Behavior After Change

- `measure` method of `InputRow` will continue loop on `undefined` value instead of throwing error.

### Reason for Changes

This issue was discovered in process of development of custom Blockly-based application with integrated [block-plus-minus plugin](https://github.com/google/blockly-samples/tree/master/plugins/block-plus-minus).

In some cases, it causes endless freeze of entire Blockly workspace after clicking "plus" button (attempt to add user-defined argument) on "Procedure" block (with active "block-plus-minus" plugin).

Screenshot of "plus" button on block

Screenshot of JS console with error log

### Test Coverage

- Probably no new tests required.
- Built-in tests are successfully passed.

### Documentation

N/A

### Additional Information

This error can break the entire rendering process of workspace, it will infinitely throw `Uncaught TypeError: this.blockDragger_ is null` in console on any gesture or mouse move. The Blockly workspace becomes unusable after triggering this error.

Screenshot of console with bunch of errors

### DIFF
--- a/core/renderers/measurables/input_row.js
+++ b/core/renderers/measurables/input_row.js
@@ -59,6 +59,9 @@ class InputRow extends Row {
     let connectedBlockWidths = 0;
     for (let i = 0; i < this.elements.length; i++) {
       const elem = this.elements[i];
+      if (elem === undefined) {
+        continue;
+      }
       this.width += elem.width;
       if (Types.isInput(elem) && elem instanceof InputConnection) {
         if (Types.isStatementInput(elem) && elem instanceof StatementInput) {


gonfunko:
Thanks for reporting this! Is your application open source/otherwise available to take a look? I'm a bit confused about how an undefined value is getting in the elements array in the first place; the change works around that, but row.js declares the type of the elements array as `* @type {!Array<!Measurable>}`, ie undefined shouldn't be in there at all, and it would be great if we could address that root issue.

Zauzolkov:
I've managed to prepare a small example to reproduce this issue: https://codepen.io/zauzolkov/pen/ZErwXMG

Steps to reproduce:
- Add a procedure block
- Try to add some user-defined arguments
- Entire workspace is broken (notice error logs in console)

I found that the problem is present when **`thrasos` renderer** is used. Everything works well with `geras` and `zelos`. There is probably some problem with `thrasos` renderer.

Zauzolkov:
Probably related issue: #6135

Zauzolkov:
After some debugging… Looks like I found possible roots of this issue, where pushing of `undefined` value to `row.elements` is happening. Implementing a similar fix in `thrasos` renderer resolves this issue.

https://github.com/google/blockly/blob/master/core/renderers/geras/info.js#L140-L144
https://github.com/google/blockly/blob/master/core/renderers/thrasos/info.js#L85-L86

gonfunko:
Thanks a bunch for digging into this, I think you're exactly right that that's the root cause!

PR #6222, opened 2022-06-16
chore(tests): Drop build/test support for node.js v12

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Brings our test targets in line with the currently-supported versions of  node.js, less v18 which requires some non-trivial dependency updates.

### Proposed Changes

1. Update GitHub actions to remove v12 from our test matrix ~~and add v18 instead~~.
0. Delete no-longer-needed workarounds for `closure-calculate-chunks` (which requires v14) from `build_tasks.js`.
   * This has the effect of making it no longer possible to _build_ Blockly on node.js v12.  There is no reason to believe that it will no longer be possible to _run_ the `blockly` npm on node.js v12, though it is no longer guaranteed.

### Test Coverage

This is test coverage.  No changes to manual test procedures expected.

### Additional Information

Splitting out addition of test run on node.js v18 due to incompatibility with the older version of the `jsdom` NPM package we currently use (and we can't easily update because we load it in a non-standard way that does not work with  newer versions).


rachel-fenichel (approved):
Looks good. Also remove `scripts/gulpfiles/chunks.json` from the checked in files.

PR #6226, opened 2022-06-18
fix: Theme.name does not match registered name (#6186)

### LINKED ISSUE #6186: `Theme.name` does not match registered name
**Describe the bug**

When you create a new theme **that is not entirely lowercase**, the `.name` property of the theme does not match the name in the registry. The registry normalizes all names to lowercase, but the themes are not doing this.

**To Reproduce**

1. Open the playground source.
2. Edit the `start` function to include:
```js
const theme = Blockly.Theme.defineTheme('TEST', {});
console.log(theme);
console.log(Blockly.registry.getAllItems(Blockly.registry.Type.THEME));
```
3. Observe how the theme has the name 'TEST', but the item has the key 'test'

**Expected behavior**

The theme should also have the name 'test'

**Additional context**

This was found when Greg attempted to define a non-lowercase theme in the advanced playground.

derwehr:
## The basics

- [x] I branched from develop

### Resolves

#6186

### Proposed Changes

Add a line in defineTheme to convert its name to lower case.

#### Behavior Before Change

`Theme.name` did not match registered name, when it included upper case letters.

#### Behavior After Change

Now it does :tada:

### DIFF
--- a/core/theme.js
+++ b/core/theme.js
@@ -149,6 +149,7 @@ class Theme {
    * @return {!Theme} A new Blockly theme.
    */
   static defineTheme(name, themeObj) {
+    name = name.toLowerCase();
     const theme = new Theme(name);
     let base = themeObj['base'];
     if (base) {
--- a/tests/mocha/theme_test.js
+++ b/tests/mocha/theme_test.js
@@ -274,4 +274,11 @@ suite('Theme', function() {
           this.constants.validatedBlockStyle_(inputStyle), expectedOutput);
     });
   });
+
+  suite('defineTheme', function() {
+    test('Normalizes to lowercase', function() {
+      const theme = Blockly.Theme.defineTheme('TEST', {});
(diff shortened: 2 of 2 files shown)


cpcallen:
Thanks for this fix.  I added a question to the issue, but that is not why I have not merged this yet.

Rather, we're wondering if you would write a unit test for this change.  This would be _especially_ helpful at the moment as we're busy migrating `core/` to TypeScript in the `ts/migration` branch, and we want to make sure this change doesn't get lost when that branch gets merged into `develop`.  A unit test—safely outside the `core/` directory, would catch such an inadvertent regression.

derwehr:
Thanks for the review. I added a test for `defineTheme` normalizing to lowercase.

BeksOmega:
Looks perfect! Thank you again for the contribution @derwehr :D

PR #6231, opened 2022-06-21
Fix: inject function options dictionary has wrong type definition

andyanday33:
## The basics

- [x] I branched from develop

#### Behavior Before Change

The inject function was giving an error while trying to add a JSON toolbox inside an options object. Yet it was working fine after compiling to javascript.

error1

#### Behavior After Change

Typescript stopped showing any errors after the change. The function runs perfectly fine.

inject

### Reason for Changes

The inject function requires a JSON toolbox to include blocks inside the workspace. Since the function is giving an error when including a toolbox in options, it may be misleading for some.

### DIFF
--- a/typings/blockly.d.ts
+++ b/typings/blockly.d.ts
@@ -25686,7 +25686,7 @@ declare module "core/inject" {
      * @return {!WorkspaceSvg} Newly created main workspace.
      * @alias Blockly.inject
      */
-    export function inject(container: Element | string, opt_options?: (() => void) | undefined): WorkspaceSvg;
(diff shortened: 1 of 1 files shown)


cpcallen (approved):
Thanks for this.  The signature we published is definitely wrong (it looks like our tooling got confused about how the `BlocklyOptions` interface type is declared in the Closure type system) and your update is much more correct.

I'm going to go ahead and merge this (once test pass), though I should note that we're in the middle of a big migration from JS to TS, which (if successful) means that our next release will actually have typings generated from TypeScript sources.  Still: if that doesn't come to pass this PR will at least make sure that what we do release is a little less wrong than before!

PR #6232, opened 2022-06-21
fix: re-add `@package` annotations as `@internal` annotations

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

N/A

### Proposed Changes

Re-adds the `@package` tags lost during migration as `@internal` tags.

#### Behavior Before Change

No change in behavior.

#### Behavior After Change

No change in behavior.

### Reason for Changes

Don't make things that were package-internal public (over-exposing our API)

### Test Coverage

No manual testing or unit tests required.

I did write up a little python script to test that each file has the same amount of `@package` tags in JS and `@internal` tags in TS. There are only four files that are different, because they are fixes:
  * internal_constants has `@internal` tags on each exported member now.
  * field_textinput has an `@internal` tag on `initView`
  * utils/colour does not have `@internal` tags on unexported `let` members
  * i_path_obj has a new member (I guess Aaron added that?) which I added an `@internal` tag to

### Documentation

  N/A

### Additional Information

N/A


core/block.ts
```diff
@@ -190,6 +193,9 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @deprecated August 2019. Use getCommentText instead.
    */
   comment: string|Comment|null = null;
+  /**
+   * @internal
+   */
```
cpcallen, line 198: I'm not keen on this formatting: there's no reason for it to be three lines instead of one.

But then I had a look at the the original JS version I see that the tooling (or maybe @gonfunko??) has made a bit of a mess of the JSdocs for class fields: sometimes the comment has been attached to the field declaration at the top of the class, and sometimes it remains with the initialisation in the constructor.  Ugh.  That does not need to be dealt with in this PR, but I do think a follow-on to make things as consistent as possible would be desirable.

In the mean time, maybe these unnecessarily-multi-line JSDoc comments are OK where there's going to be a description added?  If so, this PR should probably be consistent—see contrary example below.

core/block_svg.ts
```diff
@@ -140,6 +141,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
   // Create core elements for the block.
   private svgGroup_: SVGGElement;
   style: BlockStyle;
+  /** @internal */
```
cpcallen, line 144: Here you've added a one-liner (which is fine, but inconsistent with the example above).

core/serialization/exceptions.ts
```diff
@@ -67,6 +67,10 @@ export class MissingConnection extends DeserializationError {
    *     'IF0', or 'next'.
    * @param block The block missing the connection.
    * @param state The state object containing the bad connection.
+   * @internal
+   * @internal
+   * @internal
+   * @internal
```
cpcallen, line 73: I think you might have been a little overzealous here.  (Were some of these meant to be attached to other methods?)

cpcallen (commented):
> I did write up a little python script to test that each file has the same amount of @package tags in JS and @internal tags in TS. There are only four files that are different, because they are fixes…

When I saw the title of the PR this was going to be the first thing I was going to ask if you'd checked.  Alas, you are as usual well ahead of me!

core/serialization/exceptions.ts
BeksOmega: Good catch! They were meant to be attached to the constructors of the other exceptions. Fixed, and did a quick search to make sure this didn't happen anywhere else.

cpcallen (approved):

PR #6237, opened 2022-06-23
chore: Restore @fileoverview comment locations

NeilFraser:


demos/blockfactory/app_controller.js
```diff
@@ -5,15 +5,14 @@
...
+ * Controller for the Blockly Factory
+ * @constructor
  */
 
 /**
- * Controller for the Blockly Factory
- * @constructor
+ * @fileoverview The AppController Class brings together the Block
+ * Factory, Block Library, and Block Exporter functionality into a single web
+ * app.
```
cpcallen, line 15: This appears to be overzealous application of a script.  We should revert.

cpcallen: (Just this one file, not the whole PR!)

NeilFraser: Oh, that's weird.  I thought this swap was in the other direction.

PR #6244, opened 2022-06-24
fix(build): Have `npm run prepare` do nothing when running in CI

cpcallen:
## The basics

- [X] I branched from ts/migration
- [X] My pull request is against ts/migration

### Proposed Changes

Have `npm run prepare` (which is done automatically by `npm install`) do nothing when running in CI.

We don't need to do any building, because `npm test` will build everything needed in the workflows in which it is run, and we don't want to build anything in other workflows because a `tsc` error would prevent those workflows from completing.

#### Behaviour Before Change

`npm run prepare` (and `npm install`) fail if there are `tsc` errors, causing the clang-formatter and lint workflows to abort unnecessarily.

#### Behaviour After Change

Only the workflows that run `npm test` should fail if there are `tsc` errors.

### Reason for Changes

More useful CI feedback.

### Test Coverage

Have a look at the CI runs for this PR!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,7 +31,7 @@ module.exports = {
   buildCompiled: buildTasks.compiled,
   buildAdvancedCompilationTest: buildTasks.advancedCompilationTest,
   buildJavaScript: buildTasks.javaScript,
-  buildJavaScriptAndDeps: gulp.series(buildTasks.javaScript, buildTasks.deps),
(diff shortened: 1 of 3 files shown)


cpcallen:
Yes, this seems to have worked.  Actual tests still fail, but lint and format checks get run (and pass).

BeksOmega (approved):

PR #6246, opened 2022-06-26
fix: Correct enum formatting, use merged `namespace`s for types that are class static members

cpcallen:
## The basics

- [X] I branched from ts/migration
- [X] My pull request is against ts/migration

### Resolves

* Some of the damage done by MigranTS to our files—in particular causing types declared as static members of classes such as `ContextMenuRegistry.*` to no longer be exported from the top-level Blockly module (`blockly.js`/`.ts`) because they:
  *  were `interface`s that had been converted to named exports and were no longer available via a `static` property at all, or
  *  were `enum`s that had been reexported under their original static name by assignment, which exports the enum dictionary object but not the associated type information, meaning that (e.g.) `ContextMenuRegistry.ScopeType` could not be used as a type name.

### Proposed Changes

* Fix formatting issues for `KeyCodes` enum in `core/utils/keycodes.ts`.
* Fix formatting and exporting of `ContextMenuRegistry` static types in `core/contextmenu_registry.ts`.
* Fix formatting and exporting of `Input.Align` in `core/input.ts`.
* Fix exporting of `Names.NameType` in `core/names.ts`.

### Reason for Changes

* Key codes actually matter—assuming sequential alignment is fragile in the face of additions/deletions!
* Readable comments matter.
* Maintaining backwards-compatibility matters.

### Additional Information

The use of merged `namespace`s violates the Google TS style guide, which forbids `namespace`, but as far as I can tell this is [the recommended way to model the common JavaScript pattern of types as static properties of a `class` constructor](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces-with-classes).  I do not advocate for ignoring this rule entirely—only making an exception where we need to maintain backwards compatibility with our existing API and this is the suggested approach.

N.B: this PR does not address all the `enum` formatting issues that need correction, nor all of the static-member types that have been moved-and-therefore-deleted by MigranTs.  It is instead intended mainly to be a model for (and opportunity to discuss the approach to) solving the issues noted.


cpcallen:
Reviews requested from @BeksOmega as she is the lead for the `ts/migration` branch, and additionally from @maribethb and @rachel-fenichel to solicit comments on the merged namespace approach.

BeksOmega:
Important info from Christopher:

> It looks like we need to export some of these static subtypes both as they have been (static properties on a class) for backwards compatibility, and also as named exports in their own right (as MigranTS has converted some of our code to expect them that way).  This created a little problem for enum types, about which I asked on Stack Overflow (though in the process of asking I found an answer that at least works): https://stackoverflow.com/questions/72763491/how-to-correctly-alias-enum-types-and-their-objects-in-typescript/72763492#72763492

core/contextmenu_registry.ts
```diff
@@ -125,33 +114,71 @@ export class ContextMenuRegistry {
...
+   */
+  ContextMenuRegistry.registry = new ContextMenuRegistry();
 }
 
-/**
- * Singleton instance of this class. All interactions with this class should be
- * done on this object.
- */
-ContextMenuRegistry.registry = new ContextMenuRegistry();
+export type ScopeType = ContextMenuRegistry.ScopeType;
```
BeksOmega, line 178: These are only available internally to Blockly core correct? We're not planning to export these from the top level?

core/input.ts
```diff
@@ -37,10 +37,9 @@ import {RenderedConnection} from './rendered_connection.js';
  * @alias Blockly.Input
  */
 export class Input {
-  static Align: AnyDuringMigration;
   private sourceBlock_: Block;
   fieldRow: Field[] = [];
-  align: number;
```
BeksOmega, line 43: If you decided to also export these as module exports, why change the name to use the namespace identifier? I.E. why use `Input.Align` instead of `Align`?

core/names.ts
```diff
@@ -32,8 +32,7 @@ import {Workspace} from './workspace.js';
  * @alias Blockly.Names
  */
 export class Names {
-  static NameType: AnyDuringMigration;
-  static DEVELOPER_VARIABLE_TYPE: AnyDuringMigration;
+  static DEVELOPER_VARIABLE_TYPE: NameType;
```
BeksOmega, line 35:
```suggestion
  static DEVELOPER_VARIABLE_TYPE: NameType.DEVELOPER_VARIABLE;
```

cpcallen:
> I do not want to export these enums from `blockly.js/ts` until we're done with the rest of the migration. Because I don't think we have time to really consider how we want people to access things right now.

Do you mean that you don't want to add, new separate reexports of these types to `blockly.js/ts` (at least for the moment)? 
 If so I absolutely agree.

But otherwise I'm confused about your question because we were already exporting them via static properties on their respective constructor functions, and the point of this PR is to ensure we continue to do so.

core/contextmenu_registry.ts
cpcallen: Not at the present time, no.

We could export them under their own names later if we want to.

Or alternatively we could remove these separate exports and change the code that MigranTS changed back to how it used to be (using the `static`-property versions of the types).

core/names.ts
BeksOmega: Ohhh woops gotcha.

> Oh, I realise I can probably write a fancier type spec here that declares that .DEVELOPER_VARIABLE_TYPE can only be one particular value of NameType, but I'm not sure it's worth being that specific.

That would be fun, but yeah I don't think it's worth being that specific either lol.

BeksOmega:
> > I do not want to export these enums from `blockly.js/ts` until we're done with the rest of the migration. Because I don't think we have time to really consider how we want people to access things right now.
> 
> Do you mean that you don't want to add, new separate reexports of these types to `blockly.js/ts` (at least for the moment)? If so I absolutely agree.
> 
> But otherwise I'm confused about your question because we were already exporting them via static properties on their respective constructor functions, and the point of this PR is to ensure we continue to do so.

I mean I don't want to add any new ways that external developers can access these properties - at least not right now. I want them to continue accessing them as static properties on the class. I don't want them to be accessible via the module.

I think we're actually on the same page there. Because right now in `blockly.js/ts` we're using destructuring imports to just get the class right? So unless we modify `blockly.js/ts`, the only way to access the enums is as static properties on the class. I might have been confused about this when I posted originally.

But then my question is, do we need the `export const E = C.E;` line if we don't plan to modify `blockly.js/ts` to export these from the top level?

BeksOmega (approved):
Going to approve based on the assumption that my new understanding is correct haha, so that you (@cpcallen ) can merge whenever you have time.

rachel-fenichel (approved):
This approach LGTM, although I understand the concerns about namespace merging.

cpcallen:
> I mean I don't want to add any new ways that external developers can access these properties - at least not right now. I want them to continue accessing them as static properties on the class.

Agreed.

> So unless we modify `blockly.js/ts`, the only way to access the enums is as static properties on the class.

Correct.

> But then my question is, do we need the `export const E = C.E;` line if we don't plan to modify `blockly.js/ts` to export these from the top level?

Unfortunately MigranTS seems to have (in many cases, including IIRC these ones) converted other modules to use the named export rather than the static property, so they need to "be accessible via the module" unless/until we go and undo those conversions.

> Going to approve based on the assumption that my new understanding is correct haha, so that you (@cpcallen ) can merge whenever you have time.

Thanks.  I did a merge of `ts/migration` into this branch locally to make sure it resolved all the tsc errors, and while it did it caused me to notice that it was producing new Closure Compiler `JSC_UNUSED_LOCAL_ASSIGNMENT` errors which I should address before submitting.  Alas: a task for tomorrow.

PR #6251, opened 2022-06-27
fix: remaining enums that weren't properly exported

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

N/A
Cleans up enums not covered by https://github.com/google/blockly/pull/6246

### Proposed Changes

Uses declaration merging to put enums that were defined on classes back where they are "supposed" to be.

#### Behavior Before Change

Should be no change in behavior (besides whether these types exist or not).

#### Behavior After Change

Should be no change in behavior (besides whether these types exist or not).

### Reason for Changes

Backwards compatibility.

### Test Coverage

No manual testing necessary.

### Documentation

N/A

### Additional Information

~I didn't include the `export const E = C.E;` line, because I don't believe we actually want these to be declared  (at the module level) in the exported JavaScript.~

Added. Christopher explained why they are necessary for enums [here](https://github.com/google/blockly/pull/6250#issuecomment-1169109089).


core/rendered_connection.ts
```diff
@@ -566,3 +547,25 @@ export class RenderedConnection extends Connection {
...
+   */
+  export enum TrackedState {
+    WILL_TRACK = -1,
+    UNTRACKED = 0,
+    TRACKED = 1,
+  }
+}
+
+export type TrackedState = RenderedConnection.TrackedState;
+export const TrackedState = RenderedConnection.TrackedState;
```
maribethb, line 571: dunno how you're handling formatting the new files, but at some point i assume a lint rule somewhere will complain about this lack of new line at end of file

maribethb (approved):

core/rendered_connection.ts
BeksOmega: Right now clang-format is properly running, but it looks like the eslint config was never updated to parse the typescript. Will fix the problems in this PR, and then investigate separately.

PR #6255, opened 2022-07-01
fix: HTML is not allowed in messages

NeilFraser:
Not a security issue, just gets printed literally.

### DIFF
--- a/msg/json/ia.json
+++ b/msg/json/ia.json
@@ -121,7 +121,7 @@
 	"MATH_SINGLE_TOOLTIP_NEG": "Retornar le negation de un numero.",
 	"MATH_SINGLE_TOOLTIP_LN": "Retornar le logarithmo natural de un numero.",
 	"MATH_SINGLE_TOOLTIP_LOG10": "Retornar le logarithmo in base 10 de un numero.",
-	"MATH_SINGLE_TOOLTIP_EXP": "Retornar <em>e</em> elevate al potentia de un numero.",
+	"MATH_SINGLE_TOOLTIP_EXP": "Retornar e elevate al potentia de un numero.",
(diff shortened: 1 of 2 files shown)


msg/json/ia.json
```diff
@@ -121,7 +121,7 @@
 	"MATH_SINGLE_TOOLTIP_NEG": "Retornar le negation de un numero.",
 	"MATH_SINGLE_TOOLTIP_LN": "Retornar le logarithmo natural de un numero.",
 	"MATH_SINGLE_TOOLTIP_LOG10": "Retornar le logarithmo in base 10 de un numero.",
-	"MATH_SINGLE_TOOLTIP_EXP": "Retornar <em>e</em> elevate al potentia de un numero.",
+	"MATH_SINGLE_TOOLTIP_EXP": "Retornar e elevate al potentia de un numero.",
```
cpcallen, line 124: I ran both of these through Translate to see if some quote marks would be helpful to disambiguate the variable e from the word e (meaning "and" in many languages).  In French it's fine.  I was unable to tell with this one, though, because Translate detected it as Latin rather than Interlingua.

Then I found [a working Interlingua <-> English translator](https://interlingua-translator.vercel.app/).  It gives somewhat different translations depending on whether it's fed the sentence with `e`, `'e'`, `"e"`, `x`, `'x'` or `"x"`, in a way that suggests that the meaning might be clearer if quote marks were used.

cpcallen (approved):

Zauzolkov:
This fix was reverted by automatic "Localisation update": https://github.com/google/blockly/commit/cf1d297fcf7efce6cf9a9c2e1fa4e065a5b4f726

Probably need to fix this at TranslateWiki.

PR #6257, opened 2022-07-01
chore: properly package .ts and .d.ts files

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

* Makes it so that .ts sources are included in the /dist (and by extension the package).
* Makes it so that generated .d.ts files are included in the /dist (and by extension the package).
* Removes useless imports from the hand written .d.ts files.
* Modifies the hand written .d.ts files to properly reference the generated ones where applicable.
* Adds a step to rename all of the references to `AnyDuringMigration` within the generated .d.ts files to `any`.

### Reason for Changes

* Prep for sourcemaps.
* Typings for peeps using typescript.
* Make what we're doing clearer.
* Typings for peeps using typescript.
* Without this step the typings files don't work, because they can't find a reference to `AnyDuringMigration`, and we can't import `any_aliases.d.ts` because it's not a module. Plus `any` is clearer for external devs.

### Test Coverage

I did the first step from #6174 but I need to do the rest of it, which is why this is going up as a draft.

### Additional Information

Docs for removing the useless imports:
  1) [Don't use \<references path\> in your declaration files.](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#-reference-path-)
  2) [Depending on a UMD library from a UMD library](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#dependencies-on-umd-libraries): "Do not use a /// <reference directive to declare a dependency to a UMD library!"
  3) [Consuming dependencies](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#consuming-dependencies). Dependencies are necessary when you actually use the types from dependencies. eg:

```
import * as moment from "moment";
function getThing(): moment;
 ```

### CODE REVIEW

**Comment by BeksOmega** [comment]
Tried linking this to a plugin, and I'm getting the error:
```
test_bundle.js:2449 Uncaught TypeError: Cannot read properties of undefined (reading 'global')
    at Object.eval (goog.js?67b9:37:28)
    at eval (blockly_compressed.js:13:35)
    at eval (blockly_compressed.js:16:3)
    at ../../../blockly/dist/blockly_compressed.js (test_bundle.js:641:1)
    at __webpack_require__ (test_bundle.js:2446:33)
    at fn (test_bundle.js:2658:21)
    at eval (blockly.js?f917:4:13)
    at eval (blockly.js?f917:2:3)
    at ../../../blockly/dist/blockly.js (test_bundle.js:631:1)
    at __webpack_require__ (test_bundle.js:2446:33)
```

Not sure what's up, need to debug

**Comment by BeksOmega** [comment]
Closing this until we resolve the build issues.

**Comment by BeksOmega** [comment]
Linking this to the https://github.com/johnnyoshika/blockly-typescript-libcheck repo continues to work, but in samples I'm not really getting intellisense.

E.g. if I type `Blockly.W` I get `WorkspaceSvg`, `we`, `wheel`, `wheelEvent_`. And notably no `Workspace`.

I'm not sure if this is a problem with the type defs (since the work in the other repo) or if vscode is just overwhelmed.

**Comment by BeksOmega** [comment]
I was linking the `scroll-options` plugin to Blockly, and intellisense was not working, so Maribeth suggested right clicking on `blockly/core` and selecting `Go to Definition` to get it to update. I did this, and it took me to `core.d.ts`... in `plugins/block-dynamic-connection` (the first plugin directory in `plugins`), which had the old type def files. I linked that plugin as well, and then the intellisense started working.

I don't think VSCode likes monorepos.

I tested both `blockly` and `blockly/core` and they both work as expected. `blockly` gets you `Blockly.Javascript`, and `Blockly.libraryBlocks`. While `blockly/core` does not.

**Comment by cpcallen** [comment]
Looks like this should be replaced by a PR targetting `develop` now.

File: tsconfig.json [hunk 1]
```diff
@@ -3,18 +3,21 @@
...
 
     // Enable the next few options for type declarations.
     // Generate d.ts files
-    //"declaration": true,
+    "declaration": true,
     // Types should go into this directory.
     // Removing this would place the .d.ts files
     // next to the .js files
-    //"declarationDir": "build/ts/declarations",
+    "declarationDir": "build/ts/declarations",
```
cpcallen, line 20: If we're not going to read this path from this file in `package_tasks.js`, I'd rather this be supplied via command line argument in the appropriate place in `buildJavaScript` in `build_tasks.js`, to avoid duplication.

Also: `ts/declarations/` is probably better than (the existing value of `TYPINGS_BUILD_DIR`) `typings/`, but unless you forsee generating some other files in the `ts/` subdirectory I'd go with `declarations/` or even just `d.ts/`.

scripts/gulpfiles/package_tasks.js
```diff
@@ -391,6 +392,8 @@ function packageDTS() {
     'typings/msg/msg.d.ts',
   ];
   return gulp.src(handwrittenSrcs, {base: 'typings'})
+      .pipe(gulp.src(`${BUILD_DIR}/ts/**/*.d.ts`))
```
cpcallen, line 395: Please use `TYPINGS_BUILD_DIR` from `scripts/gulpfiles/config.js`—or even better—read the path from `tsconfig.json`.

(There are a few other paths that we should read from there too.)

Failing that, at minimum use the same path that `tsconfig.json` does (i.e. in this case):

```suggestion
      .pipe(gulp.src(`${BUILD_DIR}/ts/declarations/**/*.d.ts`))
```

and delete the TYPINGS_BUILD_DIR declaration, which appears to be unused.

cpcallen (changes requested):
* Shouldn't this PR deleting (or at least considerably reducing the size of) `typings/blockly.d.ts`?

* Please remove any other vestiges of the old typings system—e.g. references to `checkin:typings` in `package.json` and `checkinTypings` in `gulpfile.js`.

tsconfig.json
BeksOmega: > but unless you forsee generating some other files in the `ts/` subdirectory

Hopefully soon enough we won't even have a `build/` directory! Webpack will just create a dist for us.

BeksOmega:
I decided to use the `TYPINGS_BUILD_DIR` defined in `config` since that seemed faster, and as far as I'm concerned this is all dead code walking.

cpcallen:
> as I'm concerned this is all dead code walking.

We'll see…

cpcallen (approved):

PR #6260, opened 2022-07-05
fix!: Move backwards-compatibility hacks to `main.js`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

 Prerequisite to #5904 and #5857.

### Proposed Changes

* Create a new `goog.module`, `core/main.js`, which becomes the entry point for the first chunk in compiled mode.  (It is not used in uncompiled mode.)  Move to this new module:
  * from `blockly.js`:
    * the hack to allow setting `globalThis.Blockly.Msg` via `<script>` tag
  * from `blockly.js`, `contextmenu.js`, `events/events.js`, `tooltip.js`, and `widgetdiv.js`:
    * the `Object.defineProperties` calls that provide backward-compatibility `get`/`set` accessors for mutable exports that were removed in the migration to `goog.module`.
* Delete the "temporary hack" that was needed to make the accessors available on the `Blockly` namespace object in uncompiled mode.
* Update the end date of the deprecation period for similar accessor properties in `generator.js`.  Since these are attached to the `Generator` class, which is a named export, instead of the the module `exports` object, they do not need to be moved out of hat module—but they did have a five-year deprecation period, which has been shortened to end with the 2022Q3 release to bring it in line with the period for the other accessors.

### Reason for Changes

It is not possible to use `Object.defineProperties` to add properties to a `Module` object, so this code can't be run in uncompiled mode.  It works in compiled mode only because Closure Compiler is set to output ES5, and the generated code does not contain any `Object.seal` calls for (what would be) Module objects.

### Test Coverage

* Passes `npm test`.
* Playground appears to work.

### Documentation

Release notes will need to contain information about what to use instead of the the accessor properties, especially for `.mainWorkspace`, `.selected`, `HSV_SATURATION` and `HSV_VALUE`, which were not previously marked as deprecated.

### Additional Information

Obsoletes PR #6203.

Does not add deprecation notices to the previously non-deprecated `Blockly..mainWorkspace`, `.selected`, `.HSV_SATURATION` and `.HSV_VALUE` accessors.  We want to do that before the delayed Q2 release but it is not a prerequisite to this move, which _is_ a prerequisite to merging `ts/migration` into `develop`.


NeilFraser (approved):

core/main.js
```diff
@@ -0,0 +1,291 @@
...
+   * provide alternatives to the modal browser window.
+   * @name Blockly.confirm
+   * @type {!function(string, function()=)}
+   * @deprecated Use Blockly.dialog.confirm / .setConfirm() instead.
+   *     (December 2021)
+   * @suppress {checkTypes}
+   */
+  confirm: {
+    set: function(newConfirm) {
+      deprecation.warn('Blockly.confirm', 'December 2021', 'December 2022');
```
BeksOmega, line 69: When we add deprecation warnings to `mainWorkspace` and friends, I think we should change this to be `'September 2022'`

PR #6264, opened 2022-07-12
fix: options parser should add trailing slash to 'media' if not present. (#6263)

### LINKED ISSUE #6263: options parser should add trailing slash to 'media' if not present.
But only if 'media' is not empty (since '' is not the same as '/').

Also, add deprecation warning to usage of 'path' option.  That's been deprecated for ~8 years.

derwehr:
## The basics

- [x] I branched from develop

### Resolves
#6263 

### Proposed Changes
Add a trailing slash to `media` option, if not present.
Add deprecated warning to `path` option

### Additional Information
I was unsure about the deletion date on the deprecation warning, so I just set it to 1 year from now.

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -15,6 +15,7 @@
  */
 goog.module('Blockly.Options');
 
+const deprecation = goog.require('Blockly.utils.deprecation');
 const idGenerator = goog.require('Blockly.utils.idGenerator');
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
This LGTM! Thanks for the fix :D

PR #6279, opened 2022-07-18
chore: use `import type` where possible

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

N/A

### Proposed Changes

Changes `import`s which were `goog.requireType`s in `goog.module`-land to `import type`s so that our dependency graph is simplified.

#### Behavior Before Change

No change in behavior.

#### Behavior After Change

No change in behavior.

### Reason for Changes

Simplify dependency graph.

### Additional Information

N/A


tsconfig.json
```diff
@@ -23,5 +23,9 @@
     "moduleResolution": "node",
     "target": "ES2020",
     "strict": true,
+    "outDir": "build",
```
cpcallen, line 26: Why specify this here?  Is the parameter supplied by `buildJavaScript` in `build_tasks.js` insufficient?

tsconfig.json
```diff
@@ -23,5 +23,9 @@
...
+
+    // This does not work understand enums so we cannot leave it enabled.
+    // "importsNotUsedAsValues": "error"
```
cpcallen, line 29: Great to document, but documentation is not great: "does not work understand"??

Also: is it just getting upset about how we reexport enums from `namespaces`, or is this feature somehow fundamentally broken in some way that the TypeScript maintainers should be appraised of?

cpcallen (approved):

tsconfig.json
BeksOmega: Removed, was testing with `tsc` instead of `npm run build:js`. But not necessary otherwise.

tsconfig.json
BeksOmega: Fixed, it is not a problem with `namespaces`. I believe Christopher is digging into the actual issue, and filing a bug.

BeksOmega: Added a comment linking to https://github.com/microsoft/TypeScript/issues/49974

PR #6281, opened 2022-07-22
fix: circular dependencies

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Built code not actually working, due to circular dependencies.

### Proposed Changes

Does various things to make the built code actually work.
  * Moves `getById` and `getAll` into common to break some dependency cycles.
  * Comments out imports of certain things being used for side effects to break dependency cycles. Basically I commented them out until it started working.
  * Creates a bunch of "stub" functions/methods that do not have implementations in the file they belong to (so that we can break dependency cycles). The implementations are monkey-patched in in blockly.ts
  * Creates internal versions of a bunch of functions that are added to the `TEST_ONLY` export of their respective modules so that they can be spied/stubbed in tests.
  * Changes `goog.js` to include actual definitions of its functions, instead of pointing to the global `goog` and assuming it exists. The closure compiler couldn't figure out how to order things to get this to work. The definitions in `goog.js` just call the global `goog`.
  * Commented out the `declareModuleId` error we were getting in uncompiled mode (using the debug module loader), so that we can run tests and the playground.

#### Behavior Before Change

Built code did not work.

#### Behavior After Change

Built code works!

### Reason for Changes

Things working is good.

### Documentation

N/A

### Additional Information

This is super messy because I was just trying to get things to work. Happy to clean up later, but wanted to put this up for discussion. Also happy to answer any questions about the weird things I am doing.

### Future work

* Investigate whether those commented-out imports are necessary or not.
* Get rid of any internal functions that were created for test stubbing.
* Figure out a way to run tests without commenting out the `declareModuleId` error. Eg use webpack or ESModule loading or something.


closure/goog/goog.js
```diff
@@ -36,48 +36,109 @@
...
+          '" has been loaded incorrectly.');
+    }
+    var exports = jscomp.require(jscomp.getCurrentModulePath());
+    goog.loadedModules_[namespace] = {
+      exports: exports,
+      type: goog.ModuleType.ES6,
+      moduleId: namespace
+    };
+  }
+};
```
cpcallen, line 134: Like `goog.require`, ``goog.declareModuleId` is is never needed in compiled mode—the compiler treats the call as declaration/pragma, and emits no code for it.  Unlike `require`, however, we (for now) definitely have calls to `declareModuleId` that actually have to happen in uncompiled mode.  But—contrary to my previous advice—I think it would be preferable to farm them out to the real implementation, if `tsc` and Closure Compiler will permit us:

```suggestion
export const declareModuleId = function(namespace) {
  if (window.goog && window.goog.declareModuleId) {
    window.goog.require.call(this, namespace);
  }
};
```

(Or `…(window.goog, namespace)…` if anything complains about the use of `this`.)

closure/goog/goog.js
```diff
@@ -36,48 +36,109 @@
...
+
 
 // Export select properties of module. Do not export the function itself or
 // goog.module.declareLegacyNamespace.
-export const module = {
-  get: goog.module.get,
-};
+// export const module = {
+//   get: goog.module.get,
+// };
```
cpcallen, line 141: You know… 🧐

One thing we haven't tried but could try as an alternative to the present PR is just to reintroduce the `goog.module.get` calls that were previously replaced with `import`s.  I'm not sure how happy `tsc` would be, but in principle that _should_ solve all our our circular dependency problems.  🤔🤔

core/procedures.ts
```diff
@@ -355,7 +356,7 @@ function mutatorChangeListener(e: Abstract) {
     return;
   }
   const workspaceId = e.workspaceId as string;
-  const workspace = Workspace.getById(workspaceId) as WorkspaceSvg;
+  const workspace = common.getWorkspaceById(workspaceId) as AnyDuringMigration as WorkspaceSvg;
```
cpcallen, line 359: Line length.

core/utils/xml.ts
```diff
@@ -95,3 +95,12 @@ export function domToText(dom: Node): string {
...
 }
+
+export function textToDom(text: string): Element {
+  const doc = textToDomDocument(text);
+  if (!doc || !doc.documentElement ||
+      doc.getElementsByTagName('parsererror').length) {
+    throw Error('textToDom was unable to parse: ' + text);
+  }
+  return doc.documentElement;
+}
```
cpcallen, line 106: Missing final `\n`.

core/workspace_svg.ts
```diff
@@ -954,11 +954,18 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
+    this.trashcan = WorkspaceSvg.newTrashcan(this);
     const svgTrashcan = this.trashcan.createDom();
     this.svgGroup_.insertBefore(svgTrashcan, this.svgBlockCanvas_);
   }
 
+  /**
+   * @internal
+   */
+  static newTrashcan(workspace: WorkspaceSvg): AnyDuringMigration {
+    return {};
```
cpcallen, line 966:
```suggestion
  static newTrashcan(workspace: WorkspaceSvg): Trashcan {
    throw new Error('unimplemented');
```

closure/goog/goog.js
```diff
@@ -36,48 +36,109 @@
...
+      try {
+        goog.debugLoader_.load_(namespace);
+      } finally {
+        goog.moduleLoaderState_ = moduleLoaderState;
+      }
+    }
+
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
BeksOmega: I don't think we use it in any of our TS files, but we do use it in the blocks and the generators, which are also compiled using the `goog.js` file.  SO I think we need to keep it? But I'll give removing it a shot as well.

closure/goog/goog.js
BeksOmega: Attempted this, and it built and ran, but it did not fix the uncompiled problems (ie mocha tests) so we probably won't dedicate more time to this solution.

core/utils/xml.ts
```diff
@@ -95,3 +95,12 @@ export function domToText(dom: Node): string {
   const oSerializer = new XMLSerializer();
   return oSerializer.serializeToString(dom);
 }
+
+export function textToDom(text: string): Element {
```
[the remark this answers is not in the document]
BeksOmega: Breaks a circular dependency between `xml.ts` and `variables.ts`

closure/goog/goog.js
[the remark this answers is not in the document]
BeksOmega: Oh I guess the thing is the blocks use the global `goog`? Not the one imported from the `goog.js` module?

core/procedures.ts
```diff
@@ -333,8 +334,8 @@ export function mutatorOpenListener(e: Abstract) {
     return;
   }
   const workspaceId = (bubbleEvent.workspaceId);
-  const block = Workspace.getById(workspaceId)!.getBlockById(
-                    bubbleEvent.blockId) as BlockSvg;
+  const block = common.getWorkspaceById(workspaceId)!.getBlockById(
+                    bubbleEvent.blockId) as AnyDuringMigration as BlockSvg;
```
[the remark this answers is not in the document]
BeksOmega: At some point my IDE was complaining about insufficient type overlap, but it seems that it doesn't care anymore.

core/variables.ts
```diff
@@ -175,7 +175,7 @@ export function flyoutCategoryBlocks(workspace: Workspace): Element[] {
       // assignable to parameter of type 'Node'.
       block.appendChild(
           generateVariableFieldDom(mostRecentVariable) as AnyDuringMigration);
-      const value = Xml.textToDom(
+      const value = utilsXml.textToDom(
```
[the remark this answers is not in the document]
BeksOmega: Yes, so I guess this change wasn't actually necessary to get the build to work. Will see what happens when I revert it.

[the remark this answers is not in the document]
BeksOmega: Seems like it was unnecessary! Thanks for the catch :D

closure/goog/goog.js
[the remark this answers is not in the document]
cpcallen: Correct: blocks and generators have not yet been migrated to ES/TS modules, so do not use `goog.js`, only the versions defined in `base.js`.

core/utils/xml.ts
cpcallen: Not your fault, but still missing a newline at the end of the file…

core/blockly.ts
```diff
@@ -572,17 +572,22 @@ export const PROCEDURE_CATEGORY_NAME: string =
...
+                                   Block {
+                                     return new Block(
+                                         this, prototypeName, opt_id);
+                                   }
+
+                               WorkspaceSvg.prototype.newBlock =
+    function(prototypeName: string, opt_id?: string):
+        BlockSvg {
+          return new BlockSvg(this, prototypeName, opt_id);
+        }
```
cpcallen, line 585: Go home clang-format, you're drunk.  (Line 581: WTF?)

Could the tool be persuaded to be slightly less toolish by manually inserting a newline after the `=` in both declarations?  Or… well, please: just see what you can do to mitigate this fiasco.

cpcallen (approved):
Approved but with nits.

cpcallen:
I will leave it to your judgement about whether you think this or the `goog.module.get` approach is the more promising one to go with.

core/procedures.ts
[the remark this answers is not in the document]
cpcallen: "complaining about insufficient type overlap" smells like a bug (albeit possibly in the editor rather than the code).

cpcallen (approved):
Test definitely run!  (Not pass… but very much a step in the right direction.)

maribethb (approved):
Overall thoughts

1) I don't love the patching approach, but it looks like that is what is going to unblock the circular dependency problem and allow us a path forward, so I can get over it.
2) I strongly dislike the TEST_ONLY solution to the mocking problem :( If this is the best path for unblocking us, then I can get over it for now but I really don't want this to stick around. I looked into it and it looks like there are some solutions for mocking modules like proxyquire, quibble, etc that allow us to basically sub in a fake module that we can stub. Alternatively I think if the tests used the compiled code we wouldn't run into this issue? Or some tests could be rearchitected to not have to stub as many things. Or if we make bigger changes to Blockly's architecture to fix some of the dependency cycles then we could also make some changes to allow for easier testing. I think in the near future after this is landed we should devote some time to really dig into current best practices for tests and improve our story on that front.
3) Can you update the PR description to add more information, since in the future this is likely where we'll trace these changes to. This can just be a link to the doc we have about work needed to unblock us, as long as that doc is up to date. Basically can you list the changes you had to make in this PR, and ideally give some indication as to whether there is any follow up work needed? For example as a follow up we should remove all of the dependencies that you've commented out here.

So I'm approving these changes, on the basis that they are the least-bad option we currently have, and acknowledging that some of these changes need follow up work.

core/dialog.ts
```diff
@@ -112,3 +112,12 @@ export function setPrompt(
...
 }
+
+function confirmInternal(
+    message: string, callback: (p1: boolean) => AnyDuringMigration) {
+  confirmImplementation(message, callback);
+}
+
+export const TEST_ONLY = {
+  confirmInternal,
+}
```
cpcallen, line 123: Missing `\n` at EOF.

cpcallen, line 123: This seems like avoidable redirection.  Do we need to ensure that we can correctly stub (and unstub) `confirm` even in the face of calls to `setConfirm`?

If not, why not just turn `confirmImplementation` into a property on `TEST_ONLY` and omit `confirmInternal` entirely?

core/utils/style.ts
```diff
@@ -292,3 +269,34 @@ export function getContainerOffsetToScrollInto(
...
+  style.display = originalDisplay;
+  style.position = originalPosition;
+  style.visibility = originalVisibility;
+
+  return new Size(offsetWidth, offsetHeight);
+}
+
+export const TEST_ONLY = {
+  getSizeInternal,
+}
```
cpcallen, line 302: Missing `\n` at EOF.  Does your editor have an option analogous to emacs's `require-final-newline`?

cpcallen (changes requested):
I would put the `…Internal` implementations directly after the original function they have hollowed out.  It's more readable, and makes for a cleaner diff.

(Unless you think the original function was in the wrong place, in which case move it to wherever you have put the `…Internal` verison.

core/dialog.ts
BeksOmega: Yeah I thought about just using the `confirmImplementation` here, but I wanted to make it clear that that was an actual feature for dev swapping in their own modal code, whereas we just need something to get tests to run.

Right now I prefer having the extra layer of indirection to indicate that we dealing with a non-ideal situation. I don't want to try to make good final decisions atm.

cpcallen: That seems reasonable.

core/blockly.ts
BeksOmega: Turned off clang for these lines.

core/clipboard.ts
```diff
@@ -63,9 +63,22 @@ export function paste(): ICopyable|null {
...
   copy(toDuplicate);
   const pastedThing = toDuplicate.toCopyData().source.paste(copyData!.saveInfo);
   copyData = oldCopyData;
   return pastedThing;
 }
+
+export const TEST_ONLY = {
+  duplicateInternal,
+  copyInternal,
+}
```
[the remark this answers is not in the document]
BeksOmega: Turned it on in my vscode settings. Looked into reporting a bug to clang, but I couldn't figure out which repo was the correct one. Might look into it again some other time.

VSCode settings: https://stackoverflow.com/questions/55374304/add-a-single-empty-line-as-the-last-line-in-the-code-clang-format

tests/bootstrap.js
```diff
@@ -85,13 +85,13 @@
...
-      'lua_compressed.js',
-      'php_compressed.js',
-      'python_compressed.js',
+      'build/blockly_compressed.js',
+      'build/blocks_compressed.js',
+      'build/dart_compressed.js',
+      'build/javascript_compressed.js',
+      'build/lua_compressed.js',
+      'build/php_compressed.js',
+      'build/python_compressed.js',
```
cpcallen, line 94: Did you see #6218, and the reason it is still a draft PR and not merged yet?

cpcallen (commented):
Whoop whoop!

tests/bootstrap.js
BeksOmega: Yeah this was committed accidentally :/ will revert.

BeksOmega: https://github.com/google/blockly/pull/6292

PR #6282, opened 2022-07-22
fix: bugs revealed by getting the built output working

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

N/A

### Proposed Changes

* Fixes extensions
* Removes a breaking change affecting toolbox subclasses

#### Behavior Before Change

* Extensions would throw an error for all mutators because it detected that `compose`/`decompose` had a value besides `undefined` (ie `null`)
* Toolbox subclasses would not build because `workspace_` had been renamed to `workspace`

#### Behavior After Change

* Extensions work
* Toolbox subclasses work

### Reason for Changes

Fixing regressions.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/registry.ts
+++ b/core/registry.ts
@@ -61,44 +61,31 @@ export const DEFAULT = 'default';
  * @alias Blockly.registry.Type
  */
 export class Type<T> {
-  // Type.CONNECTION_CHECKER = new Type('connectionChecker');
   static CONNECTION_CHECKER = new Type<IConnectionChecker>('connectionChecker');
 
-  // Type.CURSOR = new Type('cursor');
   static CURSOR = new Type<Cursor>('cursor');
 
-  // Type.EVENT = new Type('event');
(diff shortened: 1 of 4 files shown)


core/registry.ts
```diff
@@ -61,19 +61,44 @@ export const DEFAULT = 'default';
  * @alias Blockly.registry.Type
  */
 export class Type<T> {
+  // Type.CONNECTION_CHECKER = new Type('connectionChecker');
```
cpcallen, line 64: I'm not quite sure how we ended up her, but I can't really see point keeping these comments.

Though I think I'd move the `static` declarations below the non-static methods (i.e., to where these comments were previously) if possible, as I think it's slightly clearer to declare the type  before declaring all the instances.

cpcallen (approved):

PR #6300, opened 2022-08-02
fix: Made workspace non-nullable.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857

### Proposed Changes
The workspace property of Block/BlockSvg is non-nullable.

#### Behavior Before Change
The workspace property was nullable, even though it was de facto never null.

#### Behavior After Change
The workspace is no longer nullable and doesn't need a bunch of ! and ? when referring to it.

### Test Coverage
Ensured tests still pass and manually tested.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -221,9 +221,7 @@ export class Block implements IASTNodeLocation, IDeletable {
   type!: string;
   // Record initial inline state.
   inputsInlineDefault?: boolean;
-  // Setting this to null indicates that the block has been disposed. Must be
-  // nullable.
-  workspace: Workspace|null;
+  workspace: Workspace;
 
   /**
    * @param workspace The block's workspace.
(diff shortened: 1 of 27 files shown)


BeksOmega:
It looks like most of the references to `someBlock.workspace` that were checking for nullability were fixed, but the [one in connection](https://github.com/google/blockly/blob/da8e486995df8f0b7b08a086ffb1b4bd22b21963/core/connection.js#L584) was not. I'm not sure of the best way to find all of these... the only thing I can think of is doing searches for `!.*\.workspace` and `\(.*\.workspace` and then manually auditing them. Not sure of your thoughts, but I'd like us to do some kind of check/sweep for references.

gonfunko:
> \(.*\.workspace

I did exactly that, but must have missed this one. Updated accordingly!

BeksOmega (approved):

PR #6303, opened 2022-08-03
refactor: remove AnyDuringMigration from browser events

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Remove AnyDuringMigration from block drag surface.

#### Behavior Before Change

No behavior change.

#### Behavior After Change

No behavior change.

### Reason for Changes

Types are better than AnyDuringMigration.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/browser_events.ts
+++ b/core/browser_events.ts
@@ -185,6 +185,9 @@ export function bind(
  * @alias Blockly.browserEvents.unbind
  */
 export function unbind(bindData: Data): (e: Event) => void {
+  // Accessing an element of the last property of the array is unsafe if the
+  // bindData is an empty array. But that should never happen because developers
+  // should only pass Data from bind or conditionalBind.
   let callback = bindData[bindData.length - 1][2];
   while (bindData.length) {
     const bindDatum = bindData.pop();
--- a/core/browser_events.ts
+++ b/core/browser_events.ts
@@ -156,9 +156,9 @@ export function bind(
             e.changedTouches.length === 1) {
           // Map the touch event's properties to the event.
           const touchPoint = e.changedTouches[0];
-          // TODO: We are trying to make a touch event look like a mouse event,
-          //   which is not allowed, because it requires adding more properties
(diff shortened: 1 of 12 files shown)


core/browser_events.ts
```diff
@@ -151,13 +151,17 @@ export function bind(
...
+      const touchWrapFunc = (e: Event) => {
         // Punt on multitouch events.
-        if (e.changedTouches && e.changedTouches.length === 1) {
+        if (e instanceof TouchEvent && e.changedTouches &&
+            e.changedTouches.length === 1) {
           // Map the touch event's properties to the event.
           const touchPoint = e.changedTouches[0];
-          e.clientX = touchPoint.clientX;
-          e.clientY = touchPoint.clientY;
+          // TODO: We are trying to make a touch event look like a mouse event,
```
BeksOmega, line 160: Need feedback on this.

gonfunko: I feel like we should probably introduce our own common event, or maybe see if we can use PointerEvent? But fine to leave for a later PR I think.

core/browser_events.ts
```diff
@@ -181,16 +184,16 @@ export function bind(
  * @return The function call.
  * @alias Blockly.browserEvents.unbind
  */
-export function unbind(bindData: Data): Function {
-  let func;
+export function unbind(bindData: Data): (e: Event) => void {
+  let callback = bindData[bindData.length - 1][2];
```
gonfunko, line 188: This seems potentially unsafe if bindData is empty?

BeksOmega: It is, but that should never happen b/c people *should* only pass data provided by `bind` or `conditionalBind` to this function.

I think our options are (1) we continue assuming this will be fine, or (2) we modify the return type of the function to be nullable.

gonfunko (approved):

PR #6307, opened 2022-08-03
refactor: Remove uses of AnyDuringMigration

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857

### Proposed Changes
Replaces AnyDuringMigration with the actual type of various fields, in particular for the interfaces.

#### Behavior Before Change
Type information was lacking and typechecks were too loose.

#### Behavior After Change
More type information and accurate typechecks are present.

### Test Coverage
Ensured tests still pass.


core/workspace_svg.ts
```diff
@@ -895,7 +895,6 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
     }
 
     if (this.grid_) {
-      this.grid_.dispose();
       // AnyDuringMigration because:  Type 'null' is not assignable to type
       // 'Grid'.
       this.grid_ = null as AnyDuringMigration;
```
BeksOmega, line 900: You don't need to set things to null anymore, but you should generally call their dispose functions so that they can unregister event listeners if they have them.

core/shortcut_items.ts
```diff
@@ -69,7 +69,8 @@ export function registerDelete() {
     name: names.DELETE,
     preconditionFn(workspace) {
       const selected = common.getSelected();
-      return !workspace.options.readOnly && selected && selected.isDeletable();
+      return !!(
+          !workspace.options.readOnly && selected && selected.isDeletable());
```
BeksOmega, line 73: Could you explain why this is necessary? I'm not groking it.

BeksOmega (changes requested):
Generally LGTM, just want to bring that dispose call back I think.

core/workspace_svg.ts
gonfunko: All the dispose method in grid did was null out a field, so I just removed the method entirely since it would have been a no-op. Do we want to add it back in case somebody wants to subclass it and needs to do some cleanup?

core/shortcut_items.ts
gonfunko: It wasn't clear to me either, upon further inspection it's because of the `&& selected` and JS's delightful type coercion behavior - `true && null evaluates` to `null`, so it was possible for this to return `boolean|null`.

core/workspace_svg.ts
BeksOmega: Ah ok I didn't catch that the method was removed entirely. I don't think it's possible to subclass this in a useful way atm (because the type being constructed is [hard coded](https://github.com/google/blockly/blob/da8e486995df8f0b7b08a086ffb1b4bd22b21963/core/workspace_svg.js#L463)). So I think the code you've got right now is good. If for some reason we want to allow people to construct subclasses in the future we can always add it back.

And luckily `dispose` is package/internal, so this isn't a breaking change.

core/shortcut_items.ts
BeksOmega: Goootcha gotcha. I kind of prefer `!workspace.options.readOnly && !!selected && selected.isDeletable()` because that makes it clear which part we are coercing to a boolean. But feel free to merge as is (i.e. go with whatever you prefer).

BeksOmega (approved):

PR #6313, opened 2022-08-04
feat: updates accessors for getMainWorkspace and getSelected

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

* Adds an accessor for `getSelected` to `Blockly`
* Updates the setters and getters in `main.js` to point you to the accessors on `Blockly`.
* Updates some of the docs.

#### Behavior Before Change

No accessor

#### Behavior After Change

Accessor!

### Reason for Changes

We don't really want external developers to have to use `common`, since it was just used to break some circular dependencies.

### Test Coverage

N/A

### Documentation

We probably have documentation that still points to `mainWorkspace` and `selected` that should be updated.

### Additional Information

Removal date is September 2022 so that we can remove `main.js` as soon as possible.

### DIFF
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -315,13 +315,18 @@ export function hideChaff(opt_onlyClosePopups?: boolean) {
  * Returns the main workspace.  Returns the last used main workspace (based on
  * focus).  Try not to use this function, particularly if there are multiple
  * Blockly instances on a page.
- * @return The main workspace.
  * @see Blockly.common.getMainWorkspace
  * @alias Blockly.getMainWorkspace
  */
 // AnyDuringMigration because:  Property 'getMainWorkspace' does not exist on
 // type 'void'.
 export const getMainWorkspace = (common as AnyDuringMigration).getMainWorkspace;
+
+/**
+ * Returns the currently selected copyable object.
+ * @alias Blockly.common.getSelected
+ */
+export const getSelected = common.getSelected;
 
 /**
  * Define blocks from an array of JSON block definitions, as might be generated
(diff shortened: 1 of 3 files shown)


core/blockly.ts
```diff
@@ -315,14 +315,19 @@ export function hideChaff(opt_onlyClosePopups?: boolean) {
  * Returns the main workspace.  Returns the last used main workspace (based on
  * focus).  Try not to use this function, particularly if there are multiple
  * Blockly instances on a page.
- * @return The main workspace.
  * @see Blockly.common.getMainWorkspace
  * @alias Blockly.getMainWorkspace
  */
 // AnyDuringMigration because:  Property 'getMainWorkspace' does not exist on
 // type 'void'.
 export const getMainWorkspace = (common as AnyDuringMigration).getMainWorkspace;
```
gonfunko, line 323: Is this cast actually necessary or can it be removed?

core/main.js
```diff
@@ -85,9 +85,15 @@ Object.defineProperties(Blockly, {
    */
   mainWorkspace: {
     set: function(x) {
```
gonfunko, line 87: Here and for selected, the deprecation notice for set*() doesn't really make sense since it points you to the getter.

core/blockly.ts
BeksOmega: Removed all instances of AnyDuringMigration from `blockly.ts` except for the one in `populateProcedures` since I want to get rid of that as part of the shared procedures project anyway, and it's not external facing.

core/main.js
BeksOmega: True. The other option is we just don't include a suggested replacement in the set-case (since I don't want people to set the main workspace or selected object themselves). I thought it would be better to include the get method so that people have a little bit of info, but if it's confusing I'm more than happy to remove it.

gonfunko: Fair enough, I think you're right that the context is useful. Leaving as-is SGTM.

gonfunko (approved):

PR #6315, opened 2022-08-04
refactor: Remove uses of AnyDuringMigration and fix types in BlocklyO…

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857
### Proposed Changes
Removes uses of AnyDuringMigration from BlocklyOptions and Options, and fixes other typings.

### DIFF
--- a/core/blockly_options.ts
+++ b/core/blockly_options.ts
@@ -15,8 +15,9 @@
 import * as goog from '../closure/goog/goog.js';
 goog.declareModuleId('Blockly.BlocklyOptions');
 
-import {Theme} from './theme.js';
-import {Workspace} from './workspace.js';
+import type {Theme, ITheme} from './theme.js';
+import type {WorkspaceSvg} from './workspace_svg.js';
+import type {ToolboxDefinition} from './utils/toolbox.js';
 
 
 /**
@@ -26,7 +27,7 @@ import {Workspace} from './workspace.js';
 export interface BlocklyOptions {
   collapse?: boolean;
   comments?: boolean;
-  cs?: boolean;
+  css?: boolean;
   disable?: boolean;
   grid?: GridOptions;
   horizontalLayout?: boolean;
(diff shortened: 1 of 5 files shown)


core/blockly_options.ts
```diff
@@ -26,7 +27,7 @@ import {Workspace} from './workspace.js';
 export interface BlocklyOptions {
   collapse?: boolean;
   comments?: boolean;
-  cs?: boolean;
```
rachel-fenichel, line 29: :fearful:

core/options.ts
```diff
@@ -307,77 +252,43 @@ export class Options {
...
    */
-  private static parseZoomOptions_(options: AnyDuringMigration): ZoomOptions {
+  private static parseZoomOptions_(options: BlocklyOptions): ZoomOptions {
     const zoom = options['zoom'] || {};
     const zoomOptions = {} as ZoomOptions;
     if (zoom['controls'] === undefined) {
-      // AnyDuringMigration because:  Property 'controls' does not exist on type
-      // '{}'.
-      (zoomOptions as AnyDuringMigration).controls = false;
+      zoomOptions.controls = false;
```
rachel-fenichel, line 259: This says it used to be AnyDuringMigration because the property doesn't exist on type {}, but the declaration + initialization of `zoomOptions` didn't change. What changed to make the AnyDuringMigration unncessary?

Same question for zoom, grid, and move options.

gonfunko: AnyDuringMigration was introduced by the conversion script, but subsequent to that we've already done some work cleaning up types and such to get things to build. I'm guessing either the cast to ZoomOptions or the actual fields/types of ZoomOptions were fixed after the initial migration and provided the type information that was lacking at the time of conversion. It's also possible the migration script may have just been unsure of the type for whatever reason and erred on the side of caution.

rachel-fenichel (approved):

PR #6317, opened 2022-08-05
refactor: add better types to field configs

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Changes all of the field constructors to have actual (and accurate) types for their config parameters.

#### Behavior Before Change

Bad type checking =(

#### Behavior After Change

Good type checking =)

### Reason for Changes

Types are generally considered helpful for software libraries.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A


BeksOmega:
@btw17

btw17:
🎊

core/field.ts
```diff
@@ -1200,3 +1195,10 @@ export abstract class Field implements IASTNodeLocationSvg,
     }
   }
 }
+
+/**
+ * Extra configuration options for the base field.
+ */
+export interface Config {
```
gonfunko, line 1204: Maybe call this FieldConfig so its purpose is more evident when it's imported?

core/field_angle.ts
```diff
@@ -546,4 +523,41 @@ Css.register(`
...
+ */
+export enum Mode {
+  COMPASS = 'compass',
+  PROTRACTOR = 'protractor',
+}
+
+/**
+ * Extra configuration options for the angle field.
+ */
+export interface Config extends TextConfig {
```
gonfunko, line 550: FieldAngleConfig? I think in general making these clear/unique would be helpful, and same for the FromJsonConfigs.

core/field_dropdown.ts
```diff
@@ -686,14 +692,31 @@ export class FieldDropdown extends Field {
...
   height: number;
 }
 
+
+/**
+ * An individual option in the dropdown menu. The first element is the human-
+ * readable value (text or image), and the second element is the language-
+ * neutral value.
+ */
+export type MenuOption = [string | ImageProperties, string]
```
gonfunko, line 711: Semicolon

core/field_image.ts
```diff
@@ -248,15 +246,36 @@ export class FieldImage extends Field {
    * @nocollapse
    * @internal
    */
-  static fromJson(options: AnyDuringMigration): FieldImage {
+  static fromJson(options: FromJsonConfig): FieldImage {
+    if (!options.src || !options.width || !options.height) {
```
gonfunko, line 250: Would it work to just make the fields in the FromJsonConfig non-optional? I realize you'd still want the non-zero check though.

core/field_label.ts
```diff
@@ -131,4 +130,19 @@ export class FieldLabel extends Field {
...
 
-(FieldLabel.prototype as AnyDuringMigration).DEFAULT_VALUE = '';
\ No newline at end of file
+(FieldLabel.prototype as AnyDuringMigration).DEFAULT_VALUE = '';
+
+/**
+ * Config options for the label field.
+ */
+export interface Config extends BaseFieldConfig {
+  class
```
gonfunko, line 139: Formatting seems off - not sure if the formatter got confused by treating class as a keyword?

core/field_angle.ts
```diff
@@ -21,7 +21,7 @@ import * as Css from './css.js';
 import * as dropDownDiv from './dropdowndiv.js';
 import {Field} from './field.js';
 import * as fieldRegistry from './field_registry.js';
-import {FieldTextInput} from './field_textinput.js';
+import {Config as TextConfig, FieldTextInput} from './field_textinput.js';
```
btw17, line 24: Out of curiosity, why import from `'./field_textinput.js'` instead of `'./field_textinput'`? (Ditto for other imports)

I also think it might be worth renaming each **Config** interface to correspond to the file it belongs with. For example, renaming the **Config** from `'field_textinput'` to always be **TextConfig**. Even though all of the **Config** interfaces correspond to the `opt_config` key, they're each a distinct type of config which could be confusing if a developer is implementing multiple types of fields. (For example, my IDE might tell me at a glance that both `FieldAngle` and `FieldNumber` require **Config** and I'm super confused why I can't pass the same thing into both)

BeksOmega: > Out of curiosity, why import from './field_textinput.js' instead of './field_textinput'? (Ditto for other imports)

Closure compiler reasons =( If we just use './file' it's not smart enough to find them.

> Re: renamings

Done!

core/field_image.ts
BeksOmega: We cannot because then the `FieldImage`'s  `fromJson` becomes stricter than the supertype `Field`'s `fromJson`, breaking subtyping rules :/

core/field_angle.ts
```diff
@@ -546,4 +523,41 @@ Css.register(`
...
+ *   - wrap: 0
+ *   - round: 15
+ *
+ * Protractor specifies:
+ *   - clockwise: false
+ *   - offset: 0
+ *   - wrap: 0
+ *   - round: 15
+ */
+export enum Mode {
```
btw17, line 542: [nit] I tend to see and prefer to have interfaces/enums/types at the top of a file (defined before their use) instead of the bottom, although there's no functional problem with them being at the bottom. It's ultimately just a preference, though, so no big deal if you'd rather leave it as is.

core/field_image.ts
```diff
@@ -133,10 +129,12 @@ export class FieldImage extends Field {
...
    */
-  protected override configure_(config: AnyDuringMigration) {
+  protected override configure_(config: Config) {
     super.configure_(config);
-    this.flipRtl_ = !!config['flipRtl'];
-    this.altText_ = parsing.replaceMessageReferences(config['alt']) || '';
+    if (config.flipRtl) this.flipRtl_ = config.flipRtl;
+    if (config.alt) {
+      this.altText_ = parsing.replaceMessageReferences(config.alt);
+    }
```
btw17, line 137: Even though `altText_` is set to `''` by default, this would lead to a change in functionality if `configure_` was ever called multiple times. (Maybe it's only ever intended to be called once, though, so that would only happen if a class that extends `FieldImage` improperly calls `configure_` more than once with differt configs.) If it *is* valid for `configure_` to be called multiple times, though, then:

```ts
const config1 = { alt: "abc" };
const config2 = {};
... .configure_(config1); // Sets alt to 'abc'
... .configure_(config2); // Previously set alt to '', though it remains 'abc' now
```

If it's valid to call `configure_` multiple times, the boolean value `flipRtl` would also never allow setting `this.flipRtl_` to false.

core/field_label.ts
```diff
@@ -66,9 +65,9 @@ export class FieldLabel extends Field {
     this.setValue(opt_value);
   }
 
-  override configure_(config: AnyDuringMigration) {
+  override configure_(config: Config) {
```
btw17: I notice that `configure_` is protected in some classes, like `core/field_image.ts`, and not in others like here. Are these distinctions on purpose? (It looks like it's how the original js classes were, though perhaps it was a pre-existing problem)

If it's on purpose and `configure_` is intended to be called for this class, then line 70 has a minor functionality difference where including `class` on the first call of `configure_(...)` then removing `class` from the object in the follow-up call, the second call will not properly reset `this.class_` to `null` (or `undefined` by the previous functionality)

core/field_variable.ts
```diff
@@ -123,9 +123,9 @@ export class FieldVariable extends FieldDropdown {
    * Configure the field based on the given map of options.
    * @param config A map of options to configure the field based on.
    */
-  protected override configure_(config: AnyDuringMigration) {
+  protected override configure_(config: Config) {
     super.configure_(config);
-    this.setTypes_(config['variableTypes'], config['defaultType']);
+    this.setTypes_(config.variableTypes, config.defaultType);
```
btw17, line 128: Perhaps this is best suited for another PR, though if the input is trusted to match its type contract, `setTypes_` can be updated to remove some of its type checks.

One reason *not* to remove the runtime type checks would be if a JS project using Blockly improperly used the types, though I think it's reasonable to assume the type contract is properly fulfilled in a TS environment. Since this change will be a major version bump, it shouldn't accidentally break any projects.

core/field_image.ts
BeksOmega: Thanks for catching that! Luckily calling `configure_` multiple times is not a supported use case (really it's just supposed to be treated as part of the constructor, but it's been factored out because... I can't remember why, maybe JSON initialization). So I think we're good.

core/field_label.ts
BeksOmega: Fixed the visibility modifiers. Left the implementation the same because multiple calls are not supported.

core/field_variable.ts
BeksOmega: Going to leave this as a follow-up.

core/field_angle.ts
BeksOmega: We've decided as a team not to have an opinion on the ordering of top-level declarations. (some people wanted types at the top, and other people wanted the "main" class/function/etc at the top). So I'm going to go ahead and merge as is.

core/field_image.ts
```diff
@@ -248,15 +246,36 @@ export class FieldImage extends Field {
...
 
-(FieldImage.prototype as AnyDuringMigration).DEFAULT_VALUE = '';
\ No newline at end of file
+(FieldImage.prototype as AnyDuringMigration).DEFAULT_VALUE = '';
+
+/**
+ * Config options for the image field.
+ */
+export interface FieldImageConfig extends FieldConfig {
+  flipRtl?: boolean, alt?: string,
```
gonfunko, line 271: Should these be semicolons?

core/field_number.ts
```diff
@@ -17,7 +17,7 @@ goog.declareModuleId('Blockly.FieldNumber');
 
 import {Field} from './field.js';
 import * as fieldRegistry from './field_registry.js';
-import {FieldTextInput} from './field_textinput.js';
+import {FieldTextInputConfig as FieldTextInputConfig, FieldTextInput} from './field_textinput.js';
```
gonfunko, line 20: This renaming seems unnecessary?

gonfunko (approved):

PR #6333, opened 2022-08-10
chore: convert mocha tests and test helpers to esmodules

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Converts the mcoah tests to esmodules.

#### Behavior Before Change

No change in behavior.

#### Behavior After Change

No change in behavior.

### Reason for Changes

If we want to move to webpack, it is a good idea to convert the tests to esmodules. If we want to run the tests in node rather than webdriver, it is also a good idea to convert the tests to esmodules.

### Test Coverage

This is tests silly!

### Documentation

N/A

### Additional Information

There are still some `goog.declareModuleId` statements so that the `index.html` file can properly import these. And there are still `goog.require` statements because the blocks and generators have not yet been converted.

### DIFF
--- a/tests/mocha/blocks/lists_test.js
+++ b/tests/mocha/blocks/lists_test.js
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-goog.module('Blockly.test.lists');
+goog.declareModuleId('Blockly.test.lists');
 
 const {runSerializationTestSuite} = goog.require('Blockly.test.helpers.serialization');
 const {sharedTestSetup, sharedTestTeardown} = goog.require('Blockly.test.helpers.setupTeardown');
--- a/tests/mocha/blocks/lists_test.js
+++ b/tests/mocha/blocks/lists_test.js
@@ -7,9 +7,9 @@
 goog.declareModuleId('Blockly.test.lists');
 
 const {runSerializationTestSuite} = goog.require('Blockly.test.helpers.serialization');
-import {sharedTestSetup, sharedTestTeardown} from './test_helpers/setup_teardown.js';
+import {sharedTestSetup, sharedTestTeardown} from '../test_helpers/setup_teardown.js';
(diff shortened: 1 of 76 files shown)


tests/mocha/blocks/lists_test.js
```diff
@@ -4,12 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-goog.module('Blockly.test.lists');
+goog.declareModuleId('Blockly.test.lists');
 
 const {runSerializationTestSuite} = goog.require('Blockly.test.helpers.serialization');
```
rachel-fenichel, line 9: Why are the serialization helpers and ConnectionType requires not converted?

tests/mocha/blocks/logic_ternary_test.js
```diff
@@ -4,11 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-goog.module('Blockly.test.logicTernary');
+goog.declareModuleId('Blockly.test.logicTernary');
 
 const eventUtils = goog.require('Blockly.Events.utils');
```
rachel-fenichel, line 9: Event utils can be converted

tests/mocha/blocks/procedures_test.js
```diff
@@ -4,13 +4,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-goog.module('Blockly.test.procedures');
+goog.declareModuleId('Blockly.test.procedures');
 
 goog.require('Blockly');
```
rachel-fenichel, line 9: In one of the other files you converted the require for Blockly. Why not here?

BeksOmega:
I think the issue was that files in `mocha/blocks` were moved there in develop, but not ts/migration (which this was previously branched off of) so there were some errors with the rebase. Fixed.

rachel-fenichel (approved):

Issue #6325, opened 2022-08-08
Remove Internet Explorer support


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**

On June 15, 2022 Microsoft [announced](https://blogs.windows.com/windowsexperience/2022/06/15/internet-explorer-11-has-retired-and-is-officially-out-of-support-what-you-need-to-know/#:~:text=After%2025%2B%20years%20of%20helping,today%2C%20June%2015%2C%202022.) that "Internet Explorer 11 has retired and is officially out of support."

Supporting IE takes a decent amount of work and requires workarounds in both implementations and the build system. 

**Describe the solution you'd like**

- [x] Remove uses of `userAgent.IE` in Blockly core.
- [x] Remove `isIe` and `IE` from `useragent.ts`
- [x] Remove [workaround](https://github.com/google/blockly/blob/master/core/flyout_base.js#L594) in `flyout_base`
- [x] Remove IE from test matrix (internal documentation)
- [x] Remove IE from PR template
- [x] Remove workarounds for IE in `bootstrap.js`
- [ ] Remove IE workarounds from stripped-down `base.js`
- [x] #6328

**Describe alternatives you've considered**

Continue supporting IE until the heat death of the universe or 2029 (whichever comes sooner).

**Additional context**

This is a breaking change, although hopefully for very few end users.

rachel-fenichel:
Other places where we have workarounds for IE:
- `workspaceSvg.setBrowserFocus`
- `css.ts`: alternatives to `cursor: grabbing;`
- `xml.ts`: `blockToDom`

rachel-fenichel:
- `svg_math.ts`: `is3dSupported` looks like it was mostly a workaround for IE not supporting CSS transforms on SVG elements. The work here is to check whether the Firefox problem still repros and delete as much of this function as possible--meaning ideally collapse it down to `true` and then delete it all.

rachel-fenichel:
- `flyout_base.ts`:[ `positionAt_`](https://github.com/google/blockly/blob/master/core/flyout_base.js#L523) uses `setAttribute` on the SVG element for IE and Edge, which is [no longer necessary](https://caniuse.com/transforms2d).

rachel-fenichel:
- `utils/style.ts`: `getStyle`, `getComputedStyle`, and `getCascadedStyle`

rachel-fenichel:
- `utils/dom.ts`: `createSvgElement`

[...]

PR #6336, opened 2022-08-11
fix!: remove checks for IE and EdgeHTML in core

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/6325

### Proposed Changes

- Remove uses of `userAgent.IE`
- Remove export of `userAgent.IE`

Breaking change because of the removed export.

#### Behavior Before Change

Assorted checks for IE in our code.

#### Behavior After Change

No special behaviour for IE--but note that there are still special cases for Edge.

### Reason for Changes

No more IE support.

### Test Coverage

Built and ran tests.

### Additional Information

`isGecko` still uses whether IE has been found inside `useragent.js`

`isGecko = has('Gecko') && !isWebKit && !isIe && !isEdge;`

We should consider removing all exports from `useragent.js` that we aren't already using. User agent sniffing is highly not-recommended, and we can get rid of excess exports. It's technically a breaking change, but I think we want to do it in this release.

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -784,25 +784,19 @@ export abstract class Field implements IASTNodeLocationSvg,
     if (!this.borderRect_) {
       // Browsers are inconsistent in what they return for a bounding box.
       // - Webkit / Blink: fill-box / object bounding box
-      // - Gecko / Triden / EdgeHTML: stroke-box
+      // - Gecko: stroke-box
       const bBox = (this.sourceBlock_ as BlockSvg).getHeightWidth();
       const scale = (this.sourceBlock_.workspace as WorkspaceSvg).scale;
       xy = this.getAbsoluteXY_();
-      scaledWidth = bBox.width * scale;
-      scaledHeight = bBox.height * scale;
+      scaledWidth = (bBox.width + 1) * scale;
+      scaledHeight = (bBox.height +1) * scale;
 
       if (userAgent.GECKO) {
         xy.x += 1.5 * scale;
         xy.y += 1.5 * scale;
-        scaledWidth += 1 * scale;
-        scaledHeight += 1 * scale;
       } else {
-        if (!userAgent.EDGE) {
-          xy.x -= 0.5 * scale;
-          xy.y -= 0.5 * scale;
-        }
-        scaledWidth += 1 * scale;
-        scaledHeight += 1 * scale;
+        xy.x -= 0.5 * scale;
+        xy.y -= 0.5 * scale;
       }
     } else {
       const bBox = this.borderRect_.getBoundingClientRect();
--- a/core/field.ts
+++ b/core/field.ts
@@ -797,7 +797,7 @@ export abstract class Field implements IASTNodeLocationSvg,
         scaledWidth += 1 * scale;
         scaledHeight += 1 * scale;
       } else {
-        if (!userAgent.EDGE && !userAgent.IE) {
+        if (!userAgent.EDGE) {
           xy.x -= 0.5 * scale;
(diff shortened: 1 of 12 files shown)


maribethb (approved):
We should also investigate whether the Edge special cases are still needed - since Edge is now based on Chromium as of 2020 I would expect it to not. but that can be follow up work.

maribethb:
Also, I don't think this should be a chore. We don't have a strict definition, but generally I think chores shouldn't affect source code. In particular I think chores are left out of the release notes by some tools, or chores should be things we are comfortable leaving out of the release notes as they wouldn't affect external users.

So I would probably call this a fix even though it's not really a bug fix, but it's the closest thing. Or our docs say we can use the "deprecate" tag even though I'm not sure anyone has, and it only sort of fits this scenario.

core/field.ts
```diff
@@ -797,7 +797,7 @@ export abstract class Field implements IASTNodeLocationSvg,
         scaledWidth += 1 * scale;
         scaledHeight += 1 * scale;
       } else {
-        if (!userAgent.EDGE && !userAgent.IE) {
+        if (!userAgent.EDGE) {
```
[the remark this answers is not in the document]
rachel-fenichel: I looked into this more and concluded that we can [remove the checks for Edge](https://github.com/google/blockly/issues/6339) entirely, and that it makes sense to do so in the same PR since multiple of the checks are for both IE and Edge.

However, I need to load it in Firefox to check that it works after my changes, and I found that Firefox is broken.

core/field.ts
```diff
@@ -784,25 +784,19 @@ export abstract class Field implements IASTNodeLocationSvg,
...
       // - Webkit / Blink: fill-box / object bounding box
-      // - Gecko / Triden / EdgeHTML: stroke-box
+      // - Gecko: stroke-box
       const bBox = (this.sourceBlock_ as BlockSvg).getHeightWidth();
       const scale = (this.sourceBlock_.workspace as WorkspaceSvg).scale;
       xy = this.getAbsoluteXY_();
-      scaledWidth = bBox.width * scale;
-      scaledHeight = bBox.height * scale;
+      scaledWidth = (bBox.width + 1) * scale;
+      scaledHeight = (bBox.height +1) * scale;
```
maribethb, line 792: this is the line causing the clang error

core/utils/useragent.ts
```diff
@@ -63,18 +59,16 @@ function has(name: string): boolean {
...
 // Browsers.  Logic from:
 // https://github.com/google/closure-library/blob/master/closure/goog/labs/useragent/browser.js
-isIe = has('Trident') || has('MSIE');
-isEdge = has('Edge');
 // Useragent for JavaFX:
 // Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.44
 //     (KHTML, like Gecko) JavaFX/8.0 Safari/537.44
 isJavaFx = has('JavaFX');
-isChrome = (has('Chrome') || has('CriOS')) && !isEdge;
+isChrome = (has('Chrome') || has('CriOS'));
```
maribethb, line 66: do we use these for anything? maybe we should fully get rid of the ones we don't need at all

maribethb (approved):

rachel-fenichel:
Updated to fix Firefox by returning to `AnyDuringMigration` (fixes #6340). Please take a look.

Fully removing all of the user agent types we don't use sounds good as a followup PR, and matches with the goal of getting API changes out of the way this quarter.

BeksOmega (approved):

Issue #6319, opened 2022-08-07
Custom handling when the block is disposed


e-gloo:
**Is your feature request related to a problem? Please describe.**

If I understood correctly (and has it is in the [code](https://github.com/google/blockly/blob/master/core/field.js#L1148)), we cannot yet make a variable referenced by a field other than a FieldVariable. I needed a way to make 2 fields, a text field and a dropdown field, being linked somehow, so whenever we change the value in the text field, it is updated in the dropdown field at the same time, mainly when the value selected is the same as the value from the text field. In order to achieve this, when the block of the text field is created, it creates a variable with the same id as the block id, and the same name as the text field value. Then, when the value of the text field is updated, it renames the appropriate variable with the new value. From there, the block of the dropdown field listen when the variable is renamed, and then update the current value if needed.
Everything works just fine, except the block of the text field is disposed. As the block does not get the "delete" event of himself, it can't delete the related variable, and the value will never be updated in the dropdown field.

**Describe the solution you'd like**

There's already `init` function at the block definition, maybe we could have `destroy` function which would be called when the block is disposed, then we would have the same principle has a constructor/destructor.

**Describe alternatives you've considered**

Other ideas:

- Create a new type of extension that would be called when the block is disposed instead of when the block is created.
- Dispose the block after the `onchange` function is called which would let the current block get the event of it's deletion.

**Additional context**

Note that this is pretty much a workaround a field not being able to reference a variable. It would be better to get this feature as the variable's name would depend on the text field value, and calling `refreshVariableName` would update the dropdown field value. However, we would still need a way to delete the variable when the block of the text field is disposed.

Any other workarounds or ideas to handle this. The only thing I want to avoid if possible, is to handle this in the workspace change listener, as I would need to handle every disposed block in there.

Thank you.

BeksOmega:
Hello, thank you for filing this :D Would you be able to post your current code for handling this? I'm curious about why you're using a dropdown field with a variable, and not either a [variable field](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/variable) or a [dynamic dropdown field](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#dynamic_dropdowns).

> The only thing I want to avoid if possible, is to handle this in the workspace change listener, as I would need to handle every disposed block in there.

Could you elaborate on this? If you want to add extra functionality when blocks are disposed, event listeners are definitely the best way to do so =)

Best wishes,
--Beka

[...]

PR #6337, opened 2022-08-11
feat!: allow blocks to receive their own delete events

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#6319 

### Proposed Changes

Allows blocks to receive their own dispose events.

#### Behavior Before Change

Change listeners would be removed, and then dispose events would be fired (so blocks would not receive their own dispose events).

#### Behavior After Change

Dispose events are fired and then change listeners are removed (so blocks do receive their own dispose events).

### Reason for Changes

Sometimes blocks need to clean up state when they are disposed, this allows that to happen.

### Test Coverage

* Dispose events are properly received.
* Procedure caller blocks continue to delete themselves when their definitions are deleted.

### Documentation

It would be nice if we had some documentation about block lifecycle.

### Additional Information

@cpcallen I had to move the block tests back into the main mocha folder to get them to run. Would love to know what I need to do to get them to run from the `mocha/blocks` directory.

### Breaking Changes
This is a breaking change, because it could cause event listeners to be triggered when they were not previously. In particular, if you are using a deny list instead of an allow list. The recommended solution is to use an allow list in your change handler instead of a deny list.


tests/mocha/index.html
```diff
@@ -51,6 +51,7 @@
 	  // Test modules.
 	  'Blockly.test.astNode',
 	  'Blockly.test.blockChangeEvent',
+    'Blockly.test.blockDeleteEvent',
```
cpcallen, line 54: There appears to be a whitespace error here.  Are there some lurking `\t`s?

blocks/procedures.js
```diff
@@ -980,7 +980,7 @@ const PROCEDURE_CALL_COMMON = {
         Xml.domToWorkspace(xml, this.workspace);
         Events.setGroup(false);
       }
-    } else if (event.type === Events.BLOCK_DELETE) {
+    } else if (event.type === Events.BLOCK_DELETE && event.blockId != this.id) {
```
cpcallen, line 983: Strictly speaking this is a breaking change and should probably be mentioned in the PR description, even if it's very unlikely to ever cause anyone an issue.

tests/mocha/procedures_test.js
```diff
@@ -1033,6 +1032,7 @@ suite('Procedures', function() {
           });
         });
       });
+
```
cpcallen, line 1035: Spurious whitespace?

cpcallen (approved):
Approved subject to minor nits.

I'm not sure what's up with you having to move the test files in order for them to run.  Can you be more specific about how they were failing.  Were they just being ignored?

In the mean time: would it be better to rename the moved test files to `tests/mocha/blocks_*_test.js` (or `…/libraryBlocks_*_test.js`) to make it more clear what they are testing?  (E.g. Blockly core doesn't really have any "lists" that need testing…)

tests/mocha/index.html
BeksOmega: Lots!

blocks/procedures.js
BeksOmega: I don't think this is a breaking change. Before blocks did not receive their own events, so this case would have never been hit. Now I've added the `blockId` check so it continues to not be hit.

BeksOmega:
> I'm not sure what's up with you having to move the test files in order for them to run. Can you be more specific about how they were failing. Were they just being ignored?

They were just being ignored :/

>In the mean time: would it be better to rename the moved test files to tests/mocha/blocks_*_test.js (or …/libraryBlocks_*_test.js) to make it more clear what they are testing? (E.g. Blockly core doesn't really have any "lists" that need testing…)

I would prefer to fix whatever the issue is. I think they were actually originally prefixed with `block_` and moved into a separate directory for organizational purposes lol. But if there's no obvious solution I'm happy to rename them.

blocks/procedures.js
cpcallen: `onchange` is not private, so it could be called by code outside Blockly, which could pass in an arbitrary event.

I agree this is unlikely in practice, which is why I said "strictly speaking".  It's probably not useful to bump a major version number or to mention in the release notes, but it probably _is_ worth making sure that I am not the only one who has considered this scenario, and hence the comment.

cpcallen:
> > I'm not sure what's up with you having to move the test files in order for them to run. Can you be more specific about how they were failing. Were they just being ignored?
> 
> They were just being ignored :/
> 
> I would prefer to fix whatever the issue is. I think they were actually originally prefixed with `block_` and moved into a separate directory for organizational purposes lol. But if there's no obvious solution I'm happy to rename them.

Having noticed a bunch of resource not found errors in the console when running mocha tests in `develop`, I'm 95% sure the problem is files in `tests/moca/blocks/` trying to import from `'../../build/…` instead of `../../../build/…`.

cpcallen:
I had a quick look at the final version of this PR, and it LGTM, but the post-review commit history is so convoluted that I would not say I was exactly happy to see it submitted based on my earlier approval.  A judicious force-push to consolidate all the post-review changes into one or two easy-to-vet commits would definitely have made this less worrying.

PR #6350, opened 2022-08-15
chore(build): Only suppress expected warnings from closure-make-deps

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Modify `buildDeps` gulp target (in `build_tasks.js`) to only suppress expected warnings from `closure-make-deps`.

#### Behaviour Before Change

All `stderr` output from `closure-make-deps` redirected to `/dev/null`

#### Behaviour After Change

`stderr` output from `closure-make-deps` read by the gulp task and logged to console after filtering known-

### Reason for Changes

While preparing PR #6337, @BeksOmega discovered that our block tests (in `tests/mocha/blocks/`) were not being run.  This was due to invalid `import`s, where several of those files attempted to import from `../../build/…` instead of `../../../build/…`.  I was surprised that this did not provoke more complaints from various parts of our build/test infrastructure, and it occurred to me that `closure-make-deps` may have discovered the problem and reported it but we threw the error in the bit-bucket along with all the actually-spurious warnings it generates.

Thus I modified `buildDeps` to be more selective about what is suppresses.

### Additional Information

Alas it turns out that `closure-make-deps` does not in fact care if you attempt to `import` a file that does not actually exist.  Still, it is probably better that we be less heavy-handed in suppressing its warnings generally, so it seems worthwhile keeping the change I made.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -15,7 +15,7 @@ gulp.sourcemaps = require('gulp-sourcemaps');
 
 var path = require('path');
 var fs = require('fs');
-var execSync = require('child_process').execSync;
+const {exec, execSync} = require('child_process');
 var through2 = require('through2');
 
 const clangFormat = require('clang-format');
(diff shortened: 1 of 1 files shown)


scripts/gulpfiles/build_tasks.js
```diff
@@ -336,18 +336,47 @@ function buildDeps(done) {
...
-  const args = roots.map(root => `--root '${root}' `).join('');
-  execSync(
-      `closure-make-deps ${args} 2>/dev/null >'${DEPS_FILE}'`,
-      {stdio: 'inherit'});
+  function filterErrors(text) {
+    return text.split('\n')
+        .filter(
+            (line) => !/^WARNING /.test(line) ||
+                !(/Missing type declaration./.test(line) ||
+                  /illegal use of unknown JSDoc tag/.test(line)))
```
BeksOmega, line 344: Do you think we'll be dealing with this long enough that it would be useful to separate these out into an array?

scripts/gulpfiles/build_tasks.js
```diff
@@ -336,18 +336,47 @@ function buildDeps(done) {
...
+        (error, stdout, stderr) => {
+          console.warn(filterErrors(stderr));
+          if (error) {
+            reject(error);
+          } else {
+            resolve();
+          }
+        });
+  })).then(() => {
+    done();
```
BeksOmega, line 378: Is this fixing the `prepare` error?

BeksOmega (approved):
LGTM!

scripts/gulpfiles/build_tasks.js
cpcallen: I sincerely hope not.  But when I find myself a fourth regexp to this conditional then it will be time to reevaluate…

scripts/gulpfiles/build_tasks.js
cpcallen: Not as far as I know!

The switch to `Promise`s is to allow the non-`Sync` version of `exec` to be used (which in turn is so that we can capture `stderr` even when the child process does `exit(0)`).  We still need to make sure `done()` is called after all the rest of the work is done, though, as it was in the original version.  (We could of course alternatively return a vinyl stream, but that doesn't seem applicable for this task.)

PR #6352, opened 2022-08-16
fix(build): Fix sourcemaps

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #5857.

### Proposed Changes

* Enable `sourceMap` in `tsconfig.json`.  This alone is suffiicient to get functional,  per-`.ts`-file sourcemaps in `build/src/**` that work in uncompiled mode (i.e. in the playground / tests).  Moreover, no further changes are required for these to be ingested by gulp + Closure Compiler; the resulting sourcemaps—`build/*_compressed.js.map`—now correctly point at files in core/, blocks/, etc.  This works when packaged and also when doing local testing in compiled mode of the checked-in build products (i.e., after they are copied to the repository root by `checkinBuilt`).

* In order to get sourcemaps to work for local testing in compiled mode of the build products directly from `build/`, modify `buildCompiled` so that it creates symlinks in `build/` to `core/`, `blocks/` and `generators/`.

#### Behaviour Before Change

* No sourcemaps in uncompiled mode.
* In compiled mode (when packaged or checked in), sourcemaps point at `.js` files in `build/src/`.

#### Behaviour After Change

* Sourcemaps point at `.ts` files in `core/`, `blocks/` and `generators/`.

### Reason for Changes

Sourcemaps are more useful if they point at the original sources.

### Test Coverage

No manual testing changes anticipated.

### Documentation

No documentation changes expected.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -635,7 +635,7 @@ function buildCompiled() {
   for (const src of ['core', 'blocks', 'generators']) {
     const target = `${BUILD_DIR}/${src}`
     if (!fs.existsSync(target)) {
-      fs.symlinkSync(`../${dir}`, target);
+      fs.symlinkSync(`../${src}`, target);
     }
   }
 
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -630,6 +630,15 @@ function buildCompiled() {
     // option to Closure Compiler; instead feed them as input via gulp.src.
   };
 
+  // Symlink source dirs from build dir so that sourcemaps work in
+  // compiled-mode testing.
+  for (const src of ['core', 'blocks', 'generators']) {
+    const target = `${BUILD_DIR}/${src}`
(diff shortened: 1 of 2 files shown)


scripts/gulpfiles/build_tasks.js
```diff
@@ -630,6 +630,15 @@ function buildCompiled() {
     // option to Closure Compiler; instead feed them as input via gulp.src.
   };
 
+  // Symlink source dirs from build dir so that sourcemaps work in
+  // compiled-mode testing.
+  for (const src of ['core', 'blocks', 'generators']) {
+    const target = `${BUILD_DIR}/${src}`
+    if (!fs.existsSync(target)) {
+      fs.symlinkSync(`../${dir}`, target);
```
maribethb, line 638: This doesn't build, should it be `src`? if so then lgtm

cpcallen: Oh weird.  I fixed this locally and had thought I had done a force-push _before_ opening the PR, but evidently I forgot to `git add` the change before committing.  Oops!

PR #6355, opened 2022-08-17
fix!: remove unused or obsolete useragent checks

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Followup from https://github.com/google/blockly/pull/6336

### Proposed Changes

Remove exported useragent booleans for chrome, webkit, ipod, and tablet.

#### Behavior Before Change

No change.

#### Behavior After Change

No change.

### Reason for Changes

We weren't using these checks, and it's fragile for us to be maintaining and exporting them if we're not using them.

### Test Coverage
Standard tests. We'll want to test on iOS for the release, but it should only impact sound loading anyway.

### Additional Information

Developers who were using these values should read the MDN page on [browser detection using the user agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent). If there is no way to detect what you're looking for without checking the user agent, follow the instructions on that page to find the most targeted possible solution.

### DIFF
--- a/core/utils/useragent.ts
+++ b/core/utils/useragent.ts
@@ -25,8 +25,6 @@ let rawUserAgent: string;
 
 let isJavaFx: boolean;
 
-let isChrome: boolean;
-
 let isWebKit: boolean;
 
 let isGecko: boolean;
@@ -34,8 +32,6 @@ let isGecko: boolean;
 let isAndroid: boolean;
 
 let isIPad: boolean;
-
-let isIPod: boolean;
 
 let isIPhone: boolean;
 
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
I'm not sure if this counts as a chore or a fix. It addresses a requested follow-up from https://github.com/google/blockly/pull/6336.

BeksOmega (commented):
> Not a breaking change because we were not re-exporting userAgent from the top level before.

I'm not sure this is correct. It looks like we're exporting it on master:

* [utils is exported from blockly](https://github.com/google/blockly/blob/55a707662409a8829416e38e7655bb5d009aca39/core/blockly.js#L843)
* [utils exports userAgent](https://github.com/google/blockly/blob/55a707662409a8829416e38e7655bb5d009aca39/core/utils.js#L70)

rachel-fenichel:
Good catch. Changed to be a breaking change.

BeksOmega (approved):
LGTM! Please update the PR description with info about what people should do if they're relying on these things.

PR #6361, opened 2022-08-19
fix: TS errors on dependent projects with certain tsconfig settings (#6360)

### LINKED ISSUE #6360: TypteScript compilation errors in dependent project with certain tsconfig settings
I have a TS project that depends on blockly. When upgrading my blockly version to `9.0.0-beta.0`, I have the following errors when doing `tsc --watch` in my project:

    [11:27:56 AM] Starting compilation in watch mode...

    node_modules/blockly/core/block.ts:1416:22 - error TS2339: Property 'substring' does not exist on type 'any[]'.

    1416         text = (text.substring(0, opt_maxLength - 3) + '...') as
                              ~~~~~~~~~

    node_modules/blockly/core/block.ts:1420:5 - error TS2322: Type 'any[]' is not assignable to type 'string'.

    1420     return text;
             ~~~~~~~~~~~~

    node_modules/blockly/core/field_registry.ts:71:7 - error TS2352: Conversion of type 'Field' to type 'IRegistrableField' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
      Property 'fromJson' is missing in type 'Field' but required in type 'IRegistrableField'.

    71       registry.getObject(registry.Type.FIELD, options['type']) as
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    72           IRegistrableField |
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    73       null;
       ~~~~~~~~~~

      node_modules/blockly/core/interfaces/i_registrable_field.ts:30:3
        30   fromJson: fromJson;
             ~~~~~~~~
        'fromJson' is declared here.

    node_modules/blockly/core/generator.ts:162:17 - error TS2339: Property 'replace' does not exist on type 'any[]'.

    162     code = code.replace(/\n\s+$/, '\n');
                        ~~~~~~~

    node_modules/blockly/core/generator.ts:163:17 - error TS2339: Property 'replace' does not exist on type 'any[]'.

    163     code = code.replace(/[ \t]+\n/g, '\n');
                        ~~~~~~~

    node_modules/blockly/core/generator.ts:164:5 - error TS2322: Type 'any[]' is not assignable to type 'string'.

    164     return code;
            ~~~~~~~~~~~~

    node_modules/blockly/core/utils/toolbox.ts:428:3 - error TS2322: Type 'string | Element' is not assignable to type 'Element'.
      Type 'string' is not assignable to type 'Element'.

    428   return toolboxDef;

**To Reproduce**

Have a project that depends on blockly@9.0.0-beta.0 with the following tsconfig.json file. Note that the above errors cannot be bypassed even when setting `skipLibCheck` to `true`.

    {
        "compileOnSave": true,
        "compilerOptions": {
            "target": "es6",
            "module": "commonjs",
            "sourceMap": false,
            "removeComments": true,
            "listEmittedFiles": false,
            "newLine": "LF",
            "downlevelIteration" : true,
            "incremental": true,
            "jsx": "react",
            "resolveJsonModule": true,
            "lib": [
                "dom",
                "es2021",
                "WebWorker",
            ],
            "skipLibCheck": true,
            "baseUrl": ".",
            "paths": {
                "shared/*": ["src/shared/*"]
            },
        },
    }

An easier way to reproduce the same errors is to mess with the tsconfig.json file of the blockly project directly: (Changes are marked with ****). Then run `tsc --watch`.

    {
      "include": [
        "core/**/*",  // N.B.: also pulls in closure/goog/goog.js if needed.
        "closure/**/*",  // Just for ouptut directory structure.
      ],
      "exclude": [
        "core/blockly.js"
      ],
      "compilerOptions": {
        "allowJs": false, // **** set this to false
        "declaration": true,
        "sourceMap": true,
    
        "module": "ES2015",
        "moduleResolution": "node",
        "target": "ES2020",
        "strict": false, // **** set this to false
        
        // **** add the 'lib' and 'skipLibCheck' entries
        "lib": [
            "dom",
            "es2021",
            "WebWorker",
        ],
        "skipLibCheck": true,
      }
    }

**Desktop:**
 - OS: Ubuntu 22.04
-  tsc: 4.4.3

ktbytechibong:
## The basics

- [x] I branched from develop

### Resolves

resolves #6360

### Proposed Changes

Fairly minor refactoring to bypass some typescript errors appearing on dependent projects with certain tsconfig settings.

#### Behavior Before Change

N/A. No behavioral changes. 

#### Behavior After Change

N/A. No behavioral changes. 

### Test Coverage

Existing tests should cover my changes:
- run_node_test.js covers `Generator#workspaceToCode`
- toolbox_test.js has "parseToolbox" suite covers `parseToolboxTree`
- block_test.js has "toString" suite to test `Block#tostring`

Ran `npm run test` and all existing tests passed.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -1371,7 +1371,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @returns Text of block.
    */
   toString(opt_maxLength?: number, opt_emptyToken?: string): string {
-    const text = [];
+    const tokens = [];
     const emptyFieldPlaceholder = opt_emptyToken || '?';
 
     // Temporarily set flag to navigate to all fields.
@@ -1410,16 +1410,16 @@ export class Block implements IASTNodeLocation, IDeletable {
         case ASTNode.types.INPUT: {
           const connection = node.getLocation() as Connection;
           if (!node.in()) {
-            text.push(emptyFieldPlaceholder);
+            tokens.push(emptyFieldPlaceholder);
           } else if (shouldAddParentheses(connection)) {
-            text.push('(');
+            tokens.push('(');
           }
           break;
         }
         case ASTNode.types.FIELD: {
           const field = node.getLocation() as Field;
           if (field.name !== constants.COLLAPSED_FIELD_NAME) {
-            text.push(field.getText());
+            tokens.push(field.getText());
           }
           break;
         }
@@ -1437,7 +1437,7 @@ export class Block implements IASTNodeLocation, IDeletable {
           // If we hit an input on the way up, possibly close out parentheses.
           if (node && node.getType() === ASTNode.types.INPUT &&
               shouldAddParentheses(node.getLocation() as Connection)) {
-            text.push(')');
+            tokens.push(')');
           }
         }
         if (node) {
@@ -1452,28 +1452,28 @@ export class Block implements IASTNodeLocation, IDeletable {
     // Run through our text array and simplify expression to remove parentheses
     // around single field blocks.
     // E.g. ['repeat', '(', '10', ')', 'times', 'do', '?']
-    for (let i = 2; i < text.length; i++) {
-      if (text[i - 2] === '(' && text[i] === ')') {
-        text[i - 2] = text[i - 1];
-        text.splice(i - 1, 2);
+    for (let i = 2; i < tokens.length; i++) {
+      if (tokens[i - 2] === '(' && tokens[i] === ')') {
+        tokens[i - 2] = tokens[i - 1];
+        tokens.splice(i - 1, 2);
       }
     }
 
     // Join the text array, removing spaces around added parentheses.
-    let combinedText: string = text.reduce(function(acc, value) {
+    let text: string = tokens.reduce(function(acc, value) {
       return acc + (acc.substr(-1) === '(' || value === ')' ? '' : ' ') + value;
     }, '');
 
-    combinedText = combinedText.trim() || '???';
+    text = text.trim() || '???';
     if (opt_maxLength) {
       // TODO: Improve truncation so that text from this block is given
       // priority. E.g. "1+2+3+4+5+6+7+8+9=0" should be "...6+7+8+9=0", not
       // "1+2+3+4+5...". E.g. "1+2+3+4+5=6+7+8+9+0" should be "...4+5=6+7...".
-      if (combinedText.length > opt_maxLength) {
-        combinedText = combinedText.substring(0, opt_maxLength - 3) + '...';
+      if (text.length > opt_maxLength) {
+        text = text.substring(0, opt_maxLength - 3) + '...';
       }
     }
-    return combinedText;
+    return text;
   }
 
   /**
--- a/core/block.ts
+++ b/core/block.ts
@@ -1371,7 +1371,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @returns Text of block.
    */
   toString(opt_maxLength?: number, opt_emptyToken?: string): string {
-    let text = [];
+    const text = [];
     const emptyFieldPlaceholder = opt_emptyToken || '?';
 
     // Temporarily set flag to navigate to all fields.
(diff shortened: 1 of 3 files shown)


core/field_registry.ts
```diff
@@ -68,7 +68,7 @@ export function fromJson(options: AnyDuringMigration): Field|null {
  */
 function fromJsonInternal(options: AnyDuringMigration): Field|null {
   const fieldObject =
-      registry.getObject(registry.Type.FIELD, options['type']) as
+      registry.getObject(registry.Type.FIELD, options['type']) as unknown as
```
cpcallen, line 71: This smells bad.  From your bug report it looks like `tsc` is correctly deducing that `getObject` is returning a `Field`, and the error message (that `Field` is not assignable to `IRegisterableField` appears to be legit.

I think part of the problem may be that `IRegisterableField` is now incorrectly defined: [it was originally specifically _not_ an `@interface`](https://github.com/google/blockly/blob/d86cec9145edd36d7d0f89806371bc184116c948/core/interfaces/i_registrable_field.js#L27) because it's really supposed to be declaring a type that has a certain `static` method (i.e., property on the constructor function object) named `fromJson`, rather than declaring that it is a type with an _instance_ method (i.e., property on the `.prototype` object) named `from Json`, but this appears to have been erroneously converted to an `interface` by PR #6299.

So I think the fix here is one or more of the following:

1. Correct [the definition of `IRegisterableField`](https://github.com/google/blockly/blob/883d78d5a09a6f45b4982943e57c5e9dbd3fbc7e/core/interfaces/i_registrable_field.ts#L29), possibly by using [this decorator pattern](https://stackoverflow.com/a/43674389/4969945).
2. Correct [the definition of of `Type.FIELD`](https://github.com/google/blockly/blob/883d78d5a09a6f45b4982943e57c5e9dbd3fbc7e/core/registry.ts#L81) to use `IRegisterableField` instead of `Field`.
3. Add a runtime typecheck to make sure the `Field` returned by `registry.getObject` is actually an `IRegisterableField`—tricky, since interfaces are not concrete types that can be checked with `instanceof`, but possibly something like this might work:
   ```
   if (typeof Object.getPrototypeOf(fieldObject).constructor.fromJson !== 'function) {
     throw new TypeError('returned Field was not a IRegisterableField');
    }
   ```
   (I'm not sure `tsc` is clever enough to do the necessary inference, and in any case this only checks for the existence of a `fromJson` static method, not that method's signature.)

Maybe @maribethb can weigh in on what the correct type of the template parameter to `registry.Type.FIELD` should be; I'm weirded out by the lack of overlap between `Field` and `IRegisterableField` which seems like it ought to be a subclass of the former.

core/block.ts
```diff
@@ -1309,7 +1309,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @return Text of block.
    */
   toString(opt_maxLength?: number, opt_emptyToken?: string): string {
-    let text = [];
+    const text = [];
```
cpcallen, line 1312: I think I'd have a mall preference to rename this variable, and keep `text` as the name of the combined `string` variable.  But (unlike the similar case in `generator.ts` below) `lines` isn't an accurate name, so maybe you can suggest something?  But if no better idea surfaces then then the choice you've made here is fine.

cpcallen (changes requested):
Hello and thank you for sending us this useful pull request; it is appreciated!

It mostly looks great, but there is one place where I am hoping you might be able to suggest a better choice of name, and another where I think there is an actual error in our type definitions (maybe several errors) that should be fixed.

cpcallen:
Assigning to @maribethb for follow-up review as I will be on annual leave until 20th August.

core/field_registry.ts
ktbytechibong: Agreed that this is not ideal. I don't know enough about the code base to make a decision though. Will wait for suggestion.

ktbytechibong:
@btw17 commented that #6360 should be fixed with #6362. This PR will still clean up some code, but not nearly as useful after that fix is merged.

core/block.ts
cpcallen: Yes, `tokens` sounds good.

core/field_registry.ts
cpcallen: @maribethb, do you have an opinion on this?

BeksOmega: I think the best/simplest option for now is Option 3. Do a runtime check that the fieldObject (which is actually a field constructor) has a fromJson method on it, and throw a useful error if it does not. That way we know we at least have runtime safety.

We can discuss better solutions to defining the static method on a type later.

ktbytechibong:
Ok, addressed comments in the PR. Thanks.

BeksOmega (approved):
Thanks for the fixes! LGTM =) If/once CI passes I'll go ahead and merge this, and it will go into our next release!

Issue #6358, opened 2022-08-19
TypeScript issues with 9.0.0-beta.1


clementcontet:
**Describe the bug**
Following this discussion https://groups.google.com/g/blockly/c/beh5vO-hbLA, I've tested Blockly 9.0.0-beta.1 with my TypeScript project (https://github.com/Orange-OpenSource/super-coding-ball).

And I have the following issues:
1. I use the `field-slider` plugin and I had to force install it because it is not compatible with Blockly 9 (https://github.com/google/blockly-samples/blob/master/plugins/field-slider/package.json#L50)
2. ~~The compiler doesn't find the class `AnyDuringMigration`.~~
~~There is this line https://github.com/google/blockly/blob/develop/core/any_aliases.ts#L2 but it doesn't seem to be referenced.~~ Fixed with 9.0.0-beta.1
3. I can't localize Blockly: `import * as Fr from 'blockly/msg/fr';` and `Blockly.setLocale(Fr);` are not found in types
4. JavaScript generator is not working:
* If I use `import * as Blockly from 'blockly';` then `Blockly.JavaScript` is undefined
* If I use `import * as Javascript from 'blockly/javascript';` then `Javascript.workspaceToCode()` is not found
5. When I dispose the workspace when the Angular component is destoyed, I get: `TypeError: Workspace is headless.`
6. In `ITheme`, shouldn't `FontStyle` be optionnal if I don't want to override it?
7. Idem for `colourSecondary` and `colourTertiary` in `Blockly.Theme.BlockStyle`: previously I didn't set them and Blockly was fine
8. Idem for all fields of `Theme.ComponentStyle`; they are all nullable, so they should also be optional?

**Desktop:**
 - OS: Windows10
 - blockly 9.0.0-beta.1
 - typescript 4.5.5
 - tslint 6.1.3

ktbytechibong:
I also ran into issues 2, 3, and 4. Here's the tsconfig.json file of my project (which depends on blockly) if it helps.

    "compileOnSave": true,
    "compilerOptions": {
        "target": "es6",
        "module": "commonjs",
        "sourceMap": false,
        "removeComments": true,
        "listEmittedFiles": false,
        "newLine": "LF",
        "downlevelIteration" : true,
        "incremental": true,
        "jsx": "react",
        "resolveJsonModule": true,
        "lib": [
            "dom",
            "es2021",
            "WebWorker",
        ],
        "skipLibCheck": true,
    }

[...]

PR #6362, opened 2022-08-19
chore(build): use inline sources in sourcemaps; don't package sources separately

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

`tsc` compile failures for developers whose build pipeline is misconfigured in a way that causes it to ingest `.ts` source files from `node_modules/blockly/core/`—somehow without also ingesting `core/any_aliases.ts`—resulting in compile errors due to `AnyDuringMigration`.  (See e.g. item #2 of #6358.)

### Proposed Changes

Don't package the source `.ts` files; instead, include sources inline in the `.js.map` sourcemaps.

#### Behaviour Before Change

`.ts` files are copied from `core/`, `blocks/` and `generators/` to `dist/`.

Sourcemap (`.js.map`) files contain a two-byte `sourceRoot` property.

#### Behaviour After Change

`.ts` files are not packaged.

Sourcemap files contain a large `sourcesContent` property (~2MB for `blockly_compressed.js.map), containing all of the original source.

### Reason for Changes

@btw17 points out that it is [generally not advised to package `.ts` sources along with compiled code in NPM packages]https://stackoverflow.com/a/57534667/4969945), and empirically it does appear to be causing problem for some of our developers.

### Test Coverage

Manually verified that sourcemaps with inline sources work correctly in compiled and uncompiled mode.

### Additional Information

I sent email to @samelhusseini to get clarity about why sources were originally included in the Blockly NPM package.  I think it worth waiting for a response before merging this PR.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -674,8 +674,7 @@ function buildCompiled() {
       .pipe(gulp.sourcemaps.init())
       .pipe(compile(options))
       .pipe(gulp.rename({suffix: COMPILED_SUFFIX}))
-      .pipe(
-          gulp.sourcemaps.write('.', {includeContent: false, sourceRoot: './'}))
+      .pipe(gulp.sourcemaps.write('.'))
       .pipe(gulp.dest(BUILD_DIR));
 }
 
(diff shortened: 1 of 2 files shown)


maribethb (approved):
LGTM assuming we have sam's blessing. were you able to test the resulting sourcemaps work as expected in a plugin?

cpcallen:
> LGTM assuming we have sam's blessing.

I will assign to Rachel to submit when we have a reply (or she decides we don't need one).

> were you able to test the resulting sourcemaps work as expected in a plugin?

That is one configuration I did not test, but I did test loading in compiled mode which uses the same `.js.map` files so I'm 95% sure this will work.

PR #6363, opened 2022-08-19
chore: cleaned up several type cases in core/field.ts and impacted files

btw17:
## The basics

- [x] I branched from develop

### Resolves

This resolves several `AnyDuringMigration` type cases in `core/field.ts`.

### Proposed Changes

- Updates nullable types, when set, to be `<OriginalType>|null` to properly reflect its possible state
- Updates usage of those nullable types to assert they're available in `core/field.ts` and impacted descendants 
  - **Note:** Used the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), `someValue!.nowExists`
- Added type information for `callingClass` which uses Field's prototype
- Updates numbers which are expected to be strings to be cast to strings in `core/field.ts`

#### Behavior Before Change

- Less precise type information.
- `SVGTextElement.setAttribute` passed in numbers as attribute values instead of strings in `core/field.ts`

#### Behavior After Change

- More precise type information.
- `SVGTextElement.setAttribute` always passes in strings now in `core/field.ts`

### Reason for Changes

Simplifying the information space by offloading type juggling to the compiler.

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -72,9 +72,7 @@ export abstract class Field implements IASTNodeLocationSvg,
   protected value_: AnyDuringMigration;
 
   /** Validation function called when user edits an editable field. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'Function'.
-  protected validator_: Function = null as AnyDuringMigration;
+  protected validator_: Function|null = null;
 
   /**
    * Used to cache the field's tooltip value if setTooltip is called when the
@@ -87,44 +85,31 @@ export abstract class Field implements IASTNodeLocationSvg,
    * Holds the cursors svg element when the cursor is attached to the field.
    * This is null if there is no cursor on the field.
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGElement'.
-  private cursorSvg_: SVGElement = null as AnyDuringMigration;
+  private cursorSvg_: SVGElement|null = null;
 
   /**
    * Holds the markers svg element when the marker is attached to the field.
    * This is null if there is no marker on the field.
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGElement'.
-  private markerSvg_: SVGElement = null as AnyDuringMigration;
+  private markerSvg_: SVGElement|null = null;
 
   /** The rendered field's SVG group element. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGGElement'.
-  protected fieldGroup_: SVGGElement = null as AnyDuringMigration;
+  protected fieldGroup_: SVGGElement|null = null;
 
   /** The rendered field's SVG border element. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGRectElement'.
-  protected borderRect_: SVGRectElement = null as AnyDuringMigration;
+  protected borderRect_: SVGRectElement|null = null;
 
   /** The rendered field's SVG text element. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGTextElement'.
-  protected textElement_: SVGTextElement = null as AnyDuringMigration;
+  protected textElement_: SVGTextElement|null = null;
 
   /** The rendered field's text content element. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type 'Text'.
-  protected textContent_: Text = null as AnyDuringMigration;
+  protected textContent_: Text|null = null;
 
   /** Mouse down event listener data. */
   private mouseDownWrapper_: browserEvents.Data|null = null;
 
   /** Constants associated with the source block's renderer. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'ConstantProvider'.
-  protected constants_: ConstantProvider = null as AnyDuringMigration;
+  protected constants_: ConstantProvider|null = null;
 
   /**
    * Has this field been disposed of?
@@ -137,8 +122,7 @@ export abstract class Field implements IASTNodeLocationSvg,
   maxDisplayLength = 50;
 
   /** Block this field is attached to.  Starts as null, then set in init. */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type 'Block'.
-  protected sourceBlock_: Block = null as AnyDuringMigration;
+  protected sourceBlock_: Block|null = null;
 
   /** Does this block need to be re-rendered? */
   protected isDirty_ = true;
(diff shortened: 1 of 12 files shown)


BeksOmega (approved):
LGTM! It is frustrating that so many of these things have to be nullable, but I think that's a problem with the design of the field, not with this change.

One question: How would you feel about changing instances of `sourceBlock_!` to use `getSourceBlock()` instead (which already includes the assertion).

btw17:
> One question: How would you feel about changing instances of `sourceBlock_!` to use `getSourceBlock()` instead (which already includes the assertion).

I'm 100% for it! With that in mind, can I update `sourceBlock_` from `protected` to `private` as well? It would force those inheriting from `field` to use `getSourceBlock`, though since `sourceBlock_` isn't expected to be null most of the time anyway, it could save with some confusion on handling the null case.

btw17:
@BeksOmega all set! I updated `fieldGroup_!` to use `getSvgRoot()` as well. They're split into separate commits, so it should be pretty straightforward to verify.

core/field_angle.ts
```diff
@@ -198,7 +198,7 @@ export class FieldAngle extends FieldTextInput {
     // #2380)
     this.symbol_ = dom.createSvgElement(Svg.TSPAN, {});
     this.symbol_.appendChild(document.createTextNode('\u00B0'));
-    this.textElement_.appendChild(this.symbol_);
+    this.textElement_!.appendChild(this.symbol_);
```
BeksOmega, line 201: Discussed w/ Maribeth, and using `!.` in cases like this is fine b/c we know it was created in `initView`

BeksOmega (changes requested):
Thanks for your work on this dude!

core/field.ts
```diff
@@ -704,45 +686,32 @@ export abstract class Field implements IASTNodeLocationSvg,
     const constants = this.getConstants();
```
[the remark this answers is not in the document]
btw17: In the case of `getConstants()`, I actually think it's preferable *not* to throw an error since the return type properly informs the user that it could be null. (Unless an error needs to be thrown for it being something other than `ConstantProvider` or `null`?)

core/field.ts
```diff
@@ -275,7 +257,7 @@ export abstract class Field implements IASTNodeLocationSvg,
    * @return The block containing this field.
    */
   getSourceBlock(): Block {
-    return this.sourceBlock_;
```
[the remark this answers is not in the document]
btw17: Done! It'll end up being `The source block is null.` (or whatever else it is, like undefined). What do you think?

core/field.ts
```diff
@@ -614,7 +596,7 @@ export abstract class Field implements IASTNodeLocationSvg,
    * @return The group element.
    */
   getSvgRoot(): SVGGElement {
-    return this.fieldGroup_;
+    return this.fieldGroup_!;
```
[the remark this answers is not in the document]
btw17: Same error message as with source block!

core/field.ts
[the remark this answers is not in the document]
BeksOmega: Makes sense, sounds good =)

BeksOmega (approved):
Rebase/merge (your preference) to resolve conflicts, then I think this is gtg.

btw17:
My tests are broken locally even at `origin/develop` so I'm not able to verify this locally, though it looks like the tests are running into an issue with `getSvgRoot` throwing an error for falsy values of `fieldGroup_`. I'm assuming that could be fixed by reverting `getSvgRoot` to return `this.fieldGroup_!`, though this might be more of a sign that `getSvgRoot` should actually be returning `SVGElement|null`. What do you think @BeksOmega?

core/field_angle.ts
```diff
@@ -202,7 +202,7 @@ export class FieldAngle extends FieldTextInput {
     // #2380)
     this.symbol_ = dom.createSvgElement(Svg.TSPAN, {});
     this.symbol_.appendChild(document.createTextNode('\u00B0'));
-    this.textElement_.appendChild(this.symbol_);
+    this.getTextElement().appendChild(this.symbol_);
```
btw17, line 205: Hey @maribethb & @BeksOmega - I went ahead and created `getTextElement` and other wrapper functions to handle encapsulating the null error check. What do you think about utilizing it in this context, though? Earlier, it was mentioned using `this.textElement_!` made sense here since it was just defined in `initView`. The trade off here I think is 'runtime compute' vs 'developer compute', so to speak 😂

Runtime Compute - In cases like this where the null check is so minor, it would take an extreme number of runs for it to impact runtime performance. 

Developer Compute - It might be obvious that after `initView` the `textElement_` should be set, though it's just one less thing a developer has to juggle when looking through this code.

BeksOmega: I think either option makes sense in situations like this where we never expect developers to be interacting with these properties while they're null =)

BeksOmega (approved):
Just gonna wait for @maribethb 's quick lgtm as well.

core/field.ts
```diff
@@ -1217,19 +1228,20 @@ export abstract class Field implements IASTNodeLocationSvg,
    */
   setMarkerSvg(markerSvg: SVGElement) {
     if (!markerSvg) {
```
maribethb, line 1230: I think the parameter to this function needs to be nullable, otherwise this check doesn't make sense. some of our other `setMarkerSvg` methods have the same logic and the parameter is nullable there. I wonder if this was created from our closure code where we didn't always use `?` consistently

core/field_angle.ts
maribethb: We discussed this earlier this week and this is one of those cases where it's acceptable because it's easily verified that textElement is set directly beforehand in initView. The property is set and then used in the same method so there is no room for developer error - compare that to `fieldGroup_` where it could be set in multiple places and an error like forgetting to call an init method would result in the value being unexpectedly null.

There is no hard and fast rule I can give you but that's why this particular case the `!` is fine, though using the method is also fine.

core/field.ts
```diff
@@ -1101,7 +1113,7 @@ export abstract class Field implements IASTNodeLocationSvg,
    *
    * @returns Element to bind click handler to.
    */
-  protected getClickTarget_(): Element {
+  protected getClickTarget_(): Element|null {
```
maribethb, line 1116: this seems like one where if it's null, something has gone wrong and you could conceivably put the error here instead of at every calling location.

maribethb (approved):
in general, I would prefer more descriptive error messages that might give you a hint about what has done wrong like "fieldGroup_ is null. Has this field been initialized?" or "You may need to initialize this field first" or something, but that isn't always easy to tell. so don't necessarily need to change these, just something to think about whenever we add these.

core/field.ts
btw17: I actually did so initially, though tests broke when I did so.

BeksOmega:
@btw17 give me a ping when this is ready for another look =)

btw17:
> @btw17 give me a ping when this is ready for another look =)

All set now! Just updated this PR to include the latest updates from the `develop` branch.

BeksOmega (approved):

BeksOmega:
LGTM after the build passes! If you want to work on this I'll baby sit it to rerun the CI. But don't feel the need to work on it before Friday.

PR #6371, opened 2022-08-22
chore: Fix or individually disable no-this-alias lint rule violations.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857

### Proposed Changes
Fixes or individually disables violations of the no-this-alias lint rule. Where the violation was in the spirit of the rule (shenanigans with `this`) I fixed it by generally converting to arrow functions; in cases where the assignment was intentional (generally recursive traversals starting with `this`), I disabled the warning.

#### Behavior Before Change
Lint warnings for assigning `this` to a variable were always suppressed.

#### Behavior After Change
Lint warnings for assigning `this` to a variable are selectively suppressed but generally enabled.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -570,6 +570,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @return The block (if any) that surrounds the current block.
    */
   getSurroundParent(): this|null {
+    /* eslint-disable-next-line @typescript-eslint/no-this-alias */
     let block = this;
     let prevBlock;
     do {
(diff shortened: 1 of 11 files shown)


core/block.ts
```diff
@@ -570,6 +570,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @return The block (if any) that surrounds the current block.
    */
   getSurroundParent(): this|null {
+    /* eslint-disable-next-line @typescript-eslint/no-this-alias */
```
maribethb, line 573: it would be sweet if this rule were smart enough to only trigger if the alias was used inside a function. ah well.

maribethb (approved):
Thanks!

PR #6375, opened 2022-08-23
chore: remove underscores from private properties in grid.ts

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

General cleanup.

### Proposed Changes

Remove underscores on private variables and methods in `grid.ts` to match google TS style guide.
Remove unused property.

#### Behavior Before Change

No change.

#### Behavior After Change

No change.

### Reason for Changes

Match style guide; delete unused code.

### Test Coverage
Tested in the playground to make sure the grid still worked.

### Documentation
None

### DIFF
--- a/core/grid.ts
+++ b/core/grid.ts
@@ -32,11 +32,11 @@ export class Grid {
    * This should always be the same as the workspace scale.
    */
   private scale_ = 1;
-  private readonly spacing_: number;
-  private readonly length_: number;
-  private readonly line1_: SVGElement;
-  private readonly line2_: SVGElement;
-  private readonly snapToGrid_: boolean;
+  private readonly spacing: number;
+  private readonly length: number;
+  private readonly line1: SVGElement;
+  private readonly line2: SVGElement;
+  private readonly snapToGrid: boolean;
 
   /**
    * @param pattern The grid's SVG pattern, created during injection.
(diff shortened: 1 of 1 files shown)


core/grid.ts
```diff
@@ -32,11 +32,11 @@ export class Grid {
    * This should always be the same as the workspace scale.
    */
   private scale_ = 1;
```
maribethb, line 34: why not this one?

core/grid.ts
```diff
@@ -97,21 +97,21 @@ export class Grid {
   update(scale: number) {
     this.scale_ = scale;
     // MSIE freaks if it sees a 0x0 pattern, so set empty patterns to 100x100.
```
maribethb, line 99: this says it's for IE but it affects everything rn. seems like a reasonable default for everyone, so maybe update the comment? but you can also ignore; this isn't related

core/grid.ts
rachel-fenichel: Oops. Because I didn't push the commit that removed that property entirely--it was set but never used.

core/grid.ts
rachel-fenichel: Ignoring as unrelated but I added it to the list of places where we do things because of IE: https://github.com/google/blockly/issues/6325#issuecomment-1224648805

maribethb (approved):

PR #6376, opened 2022-08-23
chore: remove instances of AnyDuringMigration in icons code

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Removes instances of AnyDuringMigration and replaces them with better types where applicable.

Also makes some private things actually `private`.

#### Behavior Before Change

No change in behavior.

#### Behavior After Change

No change in bheavior.

### Reason for Changes

Types are fun!

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

I didn't fix the `Pid_` type in mutator because I still don't know how to properly type `setTimeout` return values.

### DIFF
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -76,7 +76,6 @@ export class Comment extends Icon {
 
   /** The top-level node of the comment text, or null if not created. */
   private paragraphElement_: SVGTextElement|null = null;
-  override bubble_: AnyDuringMigration;
 
   /** @param block The block associated with this comment. */
   constructor(block: BlockSvg) {
@@ -187,9 +186,7 @@ export class Comment extends Icon {
 
     setTimeout(textarea.focus.bind(textarea), 0);
 
-    // AnyDuringMigration because:  Type 'SVGForeignObjectElement | null' is not
-    // assignable to type 'SVGElement'.
-    return this.foreignObject_ as AnyDuringMigration;
+    return this.foreignObject_;
   }
 
   /** Add or remove editability of the comment. */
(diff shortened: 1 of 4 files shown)


core/comment.ts
```diff
@@ -307,8 +296,10 @@ export class Comment extends Icon {
       browserEvents.unbind(this.onInputWrapper_);
       this.onInputWrapper_ = null;
     }
-    this.bubble_.dispose();
-    this.bubble_ = null;
+    if (this.bubble_) {
+      this.bubble_.dispose();
+      this.bubble_ = null;
```
rachel-fenichel, line 301: Why do you still need to set `this.bubble_` to `null` here?

core/comment.ts
```diff
@@ -207,9 +204,7 @@ export class Comment extends Icon {
    * Resize the text area accordingly.
    */
   private onBubbleResize_() {
-    if (!this.isVisible()) {
-      return;
-    }
+    if (!this.bubble_) return;
```
rachel-fenichel, line 207: Why switch from `isVisible()` to `bubble_` for the check?

core/comment.ts
BeksOmega: This is just disposing of the whole bubble, not the icon. So we need to make sure to drop references to the bubble so it can be disposed.

core/comment.ts
BeksOmega: So typescript can know that `this.bubble_` is defined for the `this.bubble_.getBubbleSize()` call

rachel-fenichel (approved):

PR #6378, opened 2022-08-23
refactor: Remove more uses of AnyDuringMigration

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857

### Proposed Changes
Removes some uses of AnyDuringMigration and replaces them with real type annotations.


core/workspace_drag_surface_svg.ts
```diff
@@ -165,23 +159,19 @@ export class WorkspaceDragSurfaceSvg {
...
-    // to parameter of type 'string'.
-    this.SVG_.setAttribute('width', width as AnyDuringMigration);
-    // AnyDuringMigration because:  Argument of type 'number' is not assignable
-    // to parameter of type 'string'.
-    this.SVG_.setAttribute('height', height as AnyDuringMigration);
-    this.SVG_.appendChild(blockCanvas);
-    this.SVG_.appendChild(bubbleCanvas);
-    this.SVG_.style.display = 'block';
+    this.SVG.setAttribute('width', String(width));
+    this.SVG.setAttribute('height', String(height));
```
BeksOmega, line 171: The style guide doesn't have an opinion on this, but I prefer template literals since they take up less characters.

BeksOmega (approved):

core/workspace_drag_surface_svg.ts
gonfunko: Ah nice, I hadn't thought about that. There are other places using String()/toString(), so let's fix that up separately.

PR #6382, opened 2022-08-24
chore: remove AnyDuringMigration from events

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Removes `AnyDuringMigration` from all of the event files. Primarily by adding interface descriptions of the JSON serialization of the events.

#### Behavior Before Change

No change in behavior.

#### Behavior After Change

No change in behavior.

### Reason for Changes

Types are fun :D

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A


core/events/events_click.ts
```diff
@@ -60,9 +61,9 @@ export class Click extends UiBase {
    *
    * @returns JSON representation.
    */
-  override toJson(): AnyDuringMigration {
-    const json = super.toJson();
-    json['targetType'] = this.targetType;
+  override toJson(): ClickJson {
+    const json = super.toJson() as ClickJson;
+    json['targetType'] = this.targetType!;
```
maribethb, line 66: i don't understand why this one is safe. in the constructor it might be undefined since it's optional, so how do we know it's not undefined here? it seems like either we should set a default value or we should mark it optional in the json interface

core/events/events_comment_move.ts
```diff
@@ -67,15 +65,12 @@ export class CommentMove extends CommentBase {
    * called once.
```
maribethb, line 65: I can't leave a comment there but on line 57 the comment for `this.comment_` needs to be updated.

core/events/events_selected.ts
```diff
@@ -68,11 +69,16 @@ export class Selected extends UiBase {
...
-  override fromJson(json: AnyDuringMigration) {
+  override fromJson(json: SelectedJson) {
     super.fromJson(json);
     this.oldElementId = json['oldElementId'];
     this.newElementId = json['newElementId'];
   }
 }
 
+export interface SelectedJson extends AbstractEventJson {
+  oldElementId: string|null;
```
maribethb, line 80: similarly, most of the other event types seem to use `undefined` for missing properties, why do these use `null`?

core/events/events_marker_move.ts
```diff
@@ -96,4 +97,11 @@ export class MarkerMove extends UiBase {
   }
 }
 
+export interface MarkerMoveJson extends AbstractEventJson {
+  isCursor: boolean;
+  blockId: string|null;
```
maribethb, line 102: in the click events, `blockId` is optional rather than nullable, should we be consistent here?

core/events/events_viewport.ts
```diff
@@ -96,5 +96,12 @@ export class ViewportChange extends UiBase {
   }
 }
 
+export interface ViewportChangeJson extends AbstractEventJson {
+  viewTop: number;
```
maribethb, line 100: is it weird that if you create the event from js these are optional but not if you create it from json? what about allowing these to be optional but set them to 0 in `fromJson` just like in the constructor?

core/events/workspace_events.ts
```diff
@@ -57,15 +51,13 @@ export class FinishedLoading extends AbstractEvent {
    *
    * @returns JSON representation.
    */
-  override toJson(): AnyDuringMigration {
-    const json = {
-      'type': this.type,
-    };
+  override toJson(): FinishedLoadingJson {
+    const json = {'type': this.type} as FinishedLoadingJson;
```
maribethb, line 55: would `const json: FinishedLoadingJson = {'type': this.type}` allow you to avoid the cast?

core/events/workspace_events.ts
```diff
@@ -75,12 +67,18 @@ export class FinishedLoading extends AbstractEvent {
...
+  override fromJson(json: FinishedLoadingJson) {
     this.isBlank = false;
-    this.workspaceId = json['workspaceId'];
-    this.group = json['group'];
+    this.workspaceId = json['workspaceId'] || '';
+    this.group = json['group'] || '';
   }
 }
 
+export interface FinishedLoadingJson {
```
maribethb, line 77: why doesn't this one extend AbstractEventJson? since the class itself does extend AbstractEvent?

maribethb (commented):
if external developers will use the new interfaces i believe they all need to be re-exported from somewhere so that they are eventually exported from blockly.ts.

core/events/events_selected.ts
BeksOmega: The constructor accepts null, so unless we want to change that, these also have to accept null :/

Personally I think it would be a good idea to make all of the events consistently be `SomeType|undefined` but I didn't want to just do that without discussing.

core/events/events_viewport.ts
BeksOmega: I'm not sure that it's weird, because they're optional for different reasons. They're optional in the constructor so that you can create an event, and then call `fromJson` on it (which is a really ugly pattern, `fromJson` should be `static`, not an instance method).

core/events/workspace_events.ts
BeksOmega: Hmm I didn't extend it b/c `fromJson` and `toJson` don't call `super`, but I think they actually should? So I'll switch this to extend as well.

BeksOmega:
@maribethb I think this is ready for another look? But I did a lot of reorganizing, so it's definitely possible I missed one of your previous comments.

core/events/events_selected.ts
cpcallen: Aside: I prefer `null` to `undefined` as the 'empty' value of normally-object-valued properties and variables.  This is probably mostly just because I'm a (now slightly literally) greybeard C hacker, but there is a practical justification as well: since `undefined` in JS can also mean 'uninitialised' one can't be sure a value was actually set, whereas `null` shows that it was definitely initialised but is empty.  So encountering `undefined` unexpectedly tells you something has gone wrong.

It's also the convention used for [[Prototype]]: you can have a `null`-prototype object, but not an `undefined`-prototype one.

(I think the Closure type system designers must have felt similarly, too, because I note that it has a very concise syntax for indicating nullability, but no special syntax for defining undefinedability.)

BeksOmega: > (I think the Closure type system designers must have felt similarly, too, because I note that it has a very concise syntax for indicating nullability, but no special syntax for defining undefinedability.)

As opposed to typescript, which has the opposite (special syntax for undefinablility, and no special syntax for nullability).

cpcallen: > As opposed to typescript, which has the opposite (special syntax for undefinablility, and no special syntax for nullability).

True.  And fail to make a distinction between properties that do not exist and properties which exist but which have the value `undefined`.  😢

core/events/events_block_move.ts
```diff
@@ -124,14 +114,17 @@ export class BlockMove extends BlockBase {
...
     const workspace = this.getEventWorkspace_();
+    if (!this.blockId) {
+      throw new Error(
+          'The block ID is undefined. Either pass a block to ' +
+          'the constructor, or call fromJson');
+    }
     const block = workspace.getBlockById(this.blockId);
     const location = {} as BlockLocation;
     const parent = block!.getParent();
     if (parent) {
```
gonfunko, line 125: Maybe just make this `if (parent && block)` to avoid the non-null assertion?

core/events/events_block_base.ts
```diff
@@ -23,32 +23,39 @@ import {Abstract as AbstractEvent} from './events_abstract.js';
  * @alias Blockly.Events.BlockBase
  */
 export class BlockBase extends AbstractEvent {
-  override isBlank: AnyDuringMigration;
-  blockId: string;
-  override workspaceId: string;
+  override isBlank: boolean;
+  blockId?: string;
+  override workspaceId?: string;
```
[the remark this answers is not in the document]
BeksOmega: Removed!

core/events/events_block_base.ts
```diff
@@ -23,32 +23,39 @@ import {Abstract as AbstractEvent} from './events_abstract.js';
  * @alias Blockly.Events.BlockBase
  */
 export class BlockBase extends AbstractEvent {
-  override isBlank: AnyDuringMigration;
-  blockId: string;
-  override workspaceId: string;
+  override isBlank: boolean;
```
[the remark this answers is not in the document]
BeksOmega: I changed it to be `abstract` in `Abstract` but we still have to override it to set the default value.

core/events/events_var_base.ts
```diff
@@ -34,21 +33,27 @@ export class VarBase extends AbstractEvent {
...
+    if (!opt_variable) return;
 
-    /** The variable ID for the variable this event pertains to. */
-    this.varId = this.isBlank ? '' : opt_variable!.getId();
+    /** The variable id for the variable this event pertains to. */
+    this.varId = opt_variable.getId();
 
     /** The workspace identifier for this event. */
-    this.workspaceId = this.isBlank ? '' : opt_variable!.workspace.id;
+    this.workspaceId = opt_variable?.workspace.id;
```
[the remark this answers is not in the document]
BeksOmega: Thanks for the catch!

core/events/events_ui.ts
```diff
@@ -61,7 +57,7 @@ export class Ui extends UiBase {
    * @returns JSON representation.
    */
   override toJson(): AnyDuringMigration {
-    const json = super.toJson();
+    const json = super.toJson() as AnyDuringMigration;
```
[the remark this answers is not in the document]
BeksOmega: The UI event is deprecated, so I didn't update it.

core/events/events_comment_move.ts
```diff
@@ -150,8 +149,13 @@ export class CommentMove extends CommentBase {
     const target = forward ? this.newCoordinate_ : this.oldCoordinate_;
     // TODO: Check if the comment is being dragged, and give up if so.
     const current = comment.getXY();
-    comment.moveBy(target.x - current.x, target.y - current.y);
+    comment.moveBy(target!.x - current.x, target!.y - current.y);
```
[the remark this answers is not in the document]
BeksOmega: Done!

core/events/events_block_move.ts
BeksOmega: Reorganized this a bit and removed the non-null assertions.

BeksOmega:
This is ready for another look @gonfunko =)

gonfunko (approved):

PR #6383, opened 2022-08-24
refactor: Remove more uses of AnyDuringMigration.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857.

### Proposed Changes
Replaces some more uses of AnyDuringMigration with actual types.

### DIFF
--- a/core/block_animations.ts
+++ b/core/block_animations.ts
@@ -26,7 +26,7 @@ interface CloneRect {
 }
 
 /** PID of disconnect UI animation.  There can only be one at a time. */
-let disconnectPid: AnyDuringMigration = 0;
+let disconnectPid: ReturnType<typeof setTimeout>|null = null;
 
 /** SVG group of wobbling block.  There can only be one at a time. */
 let disconnectGroup: SVGElement|null = null;
@@ -199,7 +199,9 @@ function disconnectUiStep(group: SVGElement, magnitude: number, start: Date) {
  */
 export function disconnectUiStop() {
   if (disconnectGroup) {
-    clearTimeout(disconnectPid);
+    if (disconnectPid) {
+      clearTimeout(disconnectPid);
+    }
     disconnectGroup.setAttribute('transform', '');
     disconnectGroup = null;
   }
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -27,7 +27,7 @@ import type {Workspace} from './workspace.js';
  * @alias Blockly.Generator
  */
 export class Generator {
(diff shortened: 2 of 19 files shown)


core/generator.ts
```diff
@@ -433,10 +424,10 @@ export class Generator {
    */
   protected provideFunction_(desiredName: string, code: string[]|string):
       string {
-    if (!this.definitions_[desiredName]) {
+    if (!this.definitions_![desiredName]) {
```
BeksOmega, line 427: Could we just initialize these instead?

core/variables.ts
```diff
@@ -82,11 +82,10 @@ export function allDeveloperVariables(workspace: Workspace): string[] {
...
     let getDeveloperVariables = block.getDeveloperVariables;
-    if (!getDeveloperVariables &&
-        (block as AnyDuringMigration).getDeveloperVars) {
+    if (!getDeveloperVariables && (block as any).getDeveloperVars) {
       // August 2018: getDeveloperVars() was deprecated and renamed
       // getDeveloperVariables().
-      getDeveloperVariables = (block as AnyDuringMigration).getDeveloperVars;
+      getDeveloperVariables = (block as any).getDeveloperVars;
       if (!ALL_DEVELOPER_VARS_WARNINGS_BY_BLOCK_TYPE[block.type]) {
         console.warn(
```
BeksOmega, line 90: Can you replace this with a proper deprecate call w/ a date?

core/block_animations.ts
```diff
@@ -26,7 +26,7 @@ interface CloneRect {
 }
 
 /** PID of disconnect UI animation.  There can only be one at a time. */
-let disconnectPid: AnyDuringMigration = 0;
+let disconnectPid: ReturnType<typeof setTimeout>|null = null;
```
BeksOmega, line 29: It feels weird to change the type of this to conform to node stuff instead of browser stuff. But I guess these types are only seen internally, so it should be fine?

BeksOmega (approved):

core/block_animations.ts
gonfunko: It's both, it'll be browser types in a browser and Node types in Node. But yeah I don't think it should be a big deal, this field should probably be private anyway?

core/variables.ts
gonfunko: Will just remove it in a separate PR.

core/generator.ts
gonfunko: Done.

PR #6396, opened 2022-08-29
fix: remove some IE-specific code in dom and style

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6325 

### Proposed Changes

- Remove some IE-specific code related to setting and reading styles.
- Remove use of `getCascadedStyle` and mark it as deprecated.
- Simplify code in `getComputedStyle` by using `window` instead of `document.defaultView`
  - MDN documentation on [getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle#defaultview) says "In many code samples, getComputedStyle is used from the document.defaultView object. In nearly all cases, this is needless, as getComputedStyle exists on the window object as well. It's likely the defaultView pattern was a combination of folks not wanting to write a testing spec for window and making an API that was also usable in Java."
- Use `?.` to tighten up some style checks. There's future work here to sort out the `AnyDuringMigration`s and figure out whether we need to access `style` in a different way.

#### Behavior Before Change

No change in behaviour.

#### Behavior After Change

No change.

### Reason for Changes

This code ostensibly worked in IE, but that's irrelevant since we no longer transpile to ES5.

### Test Coverage
Opened the playground and clicked around in both Firefox and Chrome. Will need comprehensive testing in all browsers, along with other changes related to #6325.

### Documentation
None

### Additional Information

`getCascadedStyle` was probably not being used externally, but I'm going through the full deprecation process because of its visibility and lack of `@internal` marker.

### DIFF
--- a/core/utils/style.ts
+++ b/core/utils/style.ts
@@ -36,7 +36,7 @@ export function getSize(element: Element): Size {
  * Private version of getSize for stubbing in tests.
  */
 function getSizeInternal(element: Element): Size {
-  if (getStyle(element, 'display') !== 'none') {
+  if (getComputedStyle(element, 'display') !== 'none') {
     return getSizeWithDisplay(element);
   }
 
@@ -75,25 +75,6 @@ function getSizeWithDisplay(element: Element): Size {
 }
 
 /**
- * Cross-browser pseudo get computed style. It returns the computed style where
- * available. If not available it tries the inline style value.  It shouldn't be
- * called directly, see http://wiki/Main/ComputedStyleVsCascadedStyle for
- * discussion.
- *
- * Copied from Closure's goog.style.getStyle_
- *
- * @param element Element to get style of.
- * @param style Property to get (must be camelCase, not CSS-style).
- * @returns Style value.
- */
-function getStyle(element: Element, style: string): string {
-  // AnyDuringMigration because:  Property 'style' does not exist on type
-  // 'Element'.
-  return getComputedStyle(element, style) ||
-      (element as AnyDuringMigration).style?.[style];
-}
-
-/**
  * Retrieves a computed style value of a node. It returns empty string
  * if the property requested is an SVG one and it has not been
  * explicitly set (firefox and webkit).
@@ -110,7 +91,7 @@ export function getComputedStyle(element: Element, property: string): string {
   // element.style[..] is undefined for browser specific styles
   // as 'filter'.
   return (styles as AnyDuringMigration)[property] ||
-      styles.getPropertyValue(property) || '';
+      styles.getPropertyValue(property);
 }
 
 /**
--- a/core/utils/style.ts
+++ b/core/utils/style.ts
@@ -14,6 +14,7 @@
 import * as goog from '../../closure/goog/goog.js';
 goog.declareModuleId('Blockly.utils.style');
 
+import * as deprecation from './deprecation.js';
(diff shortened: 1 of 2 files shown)


core/utils/style.ts
```diff
@@ -109,17 +106,11 @@ function getStyle(element: Element, style: string): string {
...
-          styles.getPropertyValue(property) || '';
-    }
-  }
-
-  return '';
+  const styles = window.getComputedStyle(element);
+  // element.style[..] is undefined for browser specific styles
+  // as 'filter'.
+  return (styles as AnyDuringMigration)[property] ||
+      styles.getPropertyValue(property) || '';
```
BeksOmega, line 113:
```suggestion
      styles.getPropertyValue(property);
```

core/utils/style.ts
```diff
@@ -109,17 +106,11 @@ function getStyle(element: Element, style: string): string {
  * @alias Blockly.utils.style.getComputedStyle
  */
 export function getComputedStyle(element: Element, property: string): string {
```
BeksOmega, line 108: Why is this exported but `getStyle` is not? Do we want to deprecate this and then stop exporting it in v10?

BeksOmega (approved):

core/utils/style.ts
rachel-fenichel: I can't delete it because we use it in `flyout_button.ts`, but I can deprecate it and leave myself a note to make it `@internal` instead of deleting it in the next release.

rachel-fenichel: Actually, I don't like that because it means we have deprecation warnings in the browser for normal things (flyouts with buttons) for a full release.

BeksOmega: Can we export `getStyle` instead, and use that? (since it seems like the more general-purpose function anyway)

rachel-fenichel: But then we're just dropping one export to add another, which I don't like.

I'm looking into the uses of `getStyle` to see when one or the other gets used and whether I can simplify more.

BeksOmega: We ended up deleting `getStyle` since `element.getComputedStyle` returns a superset of the values on `element.style`, so the function would always short circuit before reaching the `||`. I.e. it wasn't doing anything different from `getComputedStyle`.

BeksOmega (approved):

rachel-fenichel:
Testing: before release, this should be tested in firefox in an iframe that starts hidden.

core/utils/style.ts
```diff
@@ -134,6 +106,9 @@ export function getComputedStyle(element: Element, property: string): string {
  * @alias Blockly.utils.style.getCascadedStyle
  */
 export function getCascadedStyle(element: Element, style: string): string {
+  deprecation.warn(
+      'Blockly.utils.style.getCascadedStyle', 'version 9.0.0',
+      'version 10.0.0');
```
cpcallen, line 111: I like this clear approach, but I note that it doesn't makie any promises about how long the deprecated function will be available for.  For example, though unlikely it's not inconceivable that a serious security bug could force a major version bump the day after 9.0 is release.

Given that:
* we don't think this is used externally, and
* we're not _obliged_ to remove deprecated features

I think that's fine, but perhaps be careful about adopting this style more generally until we are reasonably sure we can set and stick to a schedule for major version bumps.

Issue #6160, opened 2022-05-09
Remove workspace drag surface and block drag surface


BeksOmega:
**Describe the bug**

With the drag surface (~2-5 FPS) on https://github.com/google/blockly/commit/b92cbd3b915ad99659ed2ad58adf1a4950fbbf79:
![image](https://user-images.githubusercontent.com/25440652/167512135-69db8d73-676a-4d8b-a085-92c468703de7.png)

Without the drag surface (~10 FPS) https://github.com/google/blockly/commit/b92cbd3b915ad99659ed2ad58adf1a4950fbbf79:

![image](https://user-images.githubusercontent.com/25440652/167512192-fabee584-c533-4000-9f89-a8a078bc5689.png)

**To Reproduce**

1. Go to the playground.
2. Hit the 'Airstrike' button
3. Go to the performance tab of the chrome devtools.
4. Enable recording screenshots.
5. Set CPU Throttling to 6x.
6. Start a recording.
7. Drag the workspace left... right... and left again.
8. Stop the recording.

**Expected behavior**

Ideally we wouldn't drop frames while dragging. Unfortunately most of the costs of dragging occur in the SVG engine, which we don't have a ton of control over.

**Additional context**

I have been able to make this a lot faster by changing things from using `setAttribute` to using `<element>.style`. This lets the browser use its CSS optimizations.

On https://github.com/BeksOmega/blockly/tree/perf/ws-drag (~30 FPS):

![image](https://user-images.githubusercontent.com/25440652/167512322-c463305f-11e3-412a-a410-23392ea4c6cf.png)

But that branch needs *a lot* of cleanup. Eg this does not work in IE, because IE does not support CSS transforms on SVG elements. So we need some utils that will handle setting SVG locations in a generalized way.

BeksOmega:
I'm going to reopen this because the removal was reverted due to it being a breaking change, but we still want to remove it in the future.

rachel-fenichel:
We decided not to include this change in the March 2023 release because it's a breaking change. Instead we're holding it until we release v10.

When reapplying the fix in https://github.com/google/blockly/pull/6758, also apply https://github.com/google/blockly/pull/6874 to resolve https://github.com/google/blockly/issues/6852.

maribethb:
Note: We need to fix the scroll-options plugin when we release this as well.

BeksOmega:
Related to: https://github.com/google/blockly-samples/issues/1664

PR #6400, opened 2022-08-31
fix: stop using `is3dSupported`

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6325

### Proposed Changes

Stop checking whether `translate3d` is supported--it's fine on all browsers in 2022.
Deprecate `is3dSupported` in preparation for deleting it.

#### Behavior Before Change

Checked for support using a variety of browser prefixes, because there was not a standard way to check.

#### Behavior After Change

Always assume it's supported and act accordingly.

### Reason for Changes

Remove workarounds for old browser behaviour.

### Test Coverage

Checked in Chrome and Firefox.
I made the change in several pieces so that I could check functionality at an intermediate point.

### Documentation

None

### Additional Information

We might be able to make block and workspace drag surfaces always get built, or else decide they don't improve performance and delete them entirely. Either case would be nice since it would let us reduce the number of available code paths. But that's tricky because it requires performance analysis and because the block and drag surfaces are currently optional arguments to the (public) `WorkspaceSvg` constructor.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -196,8 +196,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
      * Whether to move the block to the drag surface when it is dragged.
      * True if it should move, false if it should be translated directly.
      */
-    this.useDragSurface_ =
-        svgMath.is3dSupported() && !!workspace.getBlockDragSurface();
+    this.useDragSurface_ = !!workspace.getBlockDragSurface();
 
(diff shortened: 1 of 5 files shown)


BeksOmega:
> ...or else decide they don't improve performance and delete them entirely... But that's tricky because it requires performance analysis

Already did the performance analysis: https://github.com/google/blockly/issues/6160 They do not improve performance :/

BeksOmega (approved):

PR #6401, opened 2022-08-31
chore: work on refactoring gesture code

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Removes `AnyDuringMigration` from the gesture code (mostly, I left it in where I think type guards are the correct solution, or where it deals with touch/mouse events being derpy).

Removes the different booleans tracking if a dragger exists, and instead just checks if the dagger exists.

#### Behavior Before Change

#### Behavior After Change

### Reason for Changes

Types === good?

### Test Coverage

Tested dragging things around and clicking things etc. Gestures appear to work, but who really knows.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/gesture.ts
+++ b/core/gesture.ts
@@ -53,33 +53,26 @@ export class Gesture {
    * pixels, with (0, 0) at the top left of the browser window (mouseEvent
    * clientX/Y).
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'Coordinate'.
-  private mouseDownXY_: Coordinate = null as AnyDuringMigration;
+  private mouseDownXY_ = new Coordinate(0, 0);
   private currentDragDeltaXY_: Coordinate;
 
   /**
    * The bubble that the gesture started on, or null if it did not start on a
    * bubble.
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'IBubble'.
-  private startBubble_: IBubble = null as AnyDuringMigration;
+  private startBubble_: IBubble|null = null;
 
   /**
    * The field that the gesture started on, or null if it did not start on a
    * field.
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type 'Field'.
-  private startField_: Field = null as AnyDuringMigration;
+  private startField_: Field|null = null;
 
   /**
    * The block that the gesture started on, or null if it did not start on a
    * block.
    */
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'BlockSvg'.
-  private startBlock_: BlockSvg = null as AnyDuringMigration;
+  private startBlock_: BlockSvg|null = null;
 
   /**
    * The block that this gesture targets.  If the gesture started on a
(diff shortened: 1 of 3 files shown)


core/gesture.ts
```diff
@@ -326,14 +301,12 @@ export class Gesture {
     if (!this.targetBlock_) {
       return false;
     }
-
     if (this.flyout_) {
```
rachel-fenichel, line 305: Update TSDoc: "Update this gesture to record whether a block is being dragged" --> "Check whether to start a block drag. If a block should be dragged, either from the flyout or in the workspace, create the necessary BlockDragger and start the drag."

core/gesture.ts
```diff
@@ -348,17 +321,18 @@ export class Gesture {
    * necessary WorkspaceDragger and starts the drag.
    */
   private updateIsDraggingWorkspace_() {
```
rachel-fenichel, line 324: Also update TSDoc here.

core/gesture.ts
```diff
@@ -424,6 +407,12 @@ export class Gesture {
     this.hasStarted_ = true;
 
     blockAnimations.disconnectUiStop();
+
```
rachel-fenichel, line 414: Do this check before setting `this.hasStarted_` or calling `disconnectUiStop`.

core/gesture.ts
```diff
@@ -495,11 +484,11 @@ export class Gesture {
    */
   handleMove(e: Event) {
     this.updateFromEvent_(e);
-    if (this.isDraggingWorkspace_) {
+    if (this.workspaceDragger_) {
```
rachel-fenichel, line 492: Ah I see, getting rid of the state-tracking booleans also gives you better type checks. Nice.

core/gesture.ts
```diff
@@ -944,14 +945,7 @@ export class Gesture {
...
   getCurrentDragger(): WorkspaceDragger|BubbleDragger|IBlockDragger|null {
-    if (this.isDraggingBlock_) {
-      return this.blockDragger_;
-    } else if (this.isDraggingWorkspace_) {
-      return this.workspaceDragger_;
-    } else if (this.isDraggingBubble_) {
-      return this.bubbleDragger_;
-    }
-    return null;
+    return this.blockDragger_ ?? this.workspaceDragger_ ?? this.bubbleDragger_;
```
rachel-fenichel, line 956: Nice.

core/gesture.ts
```diff
@@ -265,27 +239,30 @@ export class Gesture {
...
    */
   private updateIsDraggingFromFlyout_(): boolean {
-    if (!this.targetBlock_) {
+    if (!this.targetBlock_ ||
+        !this.flyout_?.isBlockCreatable(this.targetBlock_)) {
       return false;
     }
-    if (!this.flyout_.isBlockCreatable(this.targetBlock_)) {
-      return false;
+    if (!this.flyout_) {
```
rachel-fenichel, line 246: I think you don't need this check anymore. If flyout is undefined, it'll short-circuit on 243 on return false.

core/gesture.ts
BeksOmega: Originally had here so that it wouldn't be a change in behavior. But yeah, probably better to throw the error before we mess up the state of the gesture.

rachel-fenichel (approved):

PR #6402, opened 2022-08-31
chore: remove AnyDuringMigration from common renderer

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Removes `AnyDuringMigration` from the common renderer code.

#### Behavior Before Change

#### Behavior After Change

### Reason for Changes

Types are fun!

### Test Coverage

Tested that the geras and zelos renderers look correct.

### Documentation

N/A

### Additional Information

N/A


core/renderers/common/marker_svg.ts
```diff
@@ -214,10 +212,6 @@ export class MarkerSvg {
     }
   }
 
-  /**************************
-   * Display
-   **************************/
```
cpcallen, line 219: Doe this (and similar below) no longer convey any useful information?  Personally I find dividers like this fairly helpful.

core/renderers/common/marker_svg.ts
```diff
@@ -494,20 +482,20 @@ export class MarkerSvg {
    */
   protected positionPrevious_(
       width: number, markerOffset: number, markerHeight: number,
-      connectionShape: Notch|PuzzleTab) {
-    // AnyDuringMigration because:  Property 'pathLeft' does not exist on type
-    // 'Notch | PuzzleTab'.
+      connectionShape: Notch) {
```
cpcallen, line 485: This is a breaking change (unless we regard the whole `MarkerSvg` class as being project-internal, maybe.)

core/renderers/common/marker_svg.ts
```diff
@@ -572,7 +559,7 @@ export class MarkerSvg {
    *
    * @returns The object holding attributes to make the marker blink.
    */
-  protected getBlinkProperties_(): AnyDuringMigration {
+  protected getBlinkProperties_(): object {
```
cpcallen, line 562: `object` is definitely more specific than `any`, but might it be worth being more specific still—e.g., defining a suitable record type?

core/renderers/common/marker_svg.ts
```diff
@@ -444,15 +435,12 @@ export class MarkerSvg {
...
-    this.markerSvgLine_!.setAttribute('y', y as AnyDuringMigration);
-    // AnyDuringMigration because:  Argument of type 'number' is not assignable
-    // to parameter of type 'string'.
-    this.markerSvgLine_!.setAttribute('width', width as AnyDuringMigration);
+    if (!this.markerSvgLine_) {
+      throw new Error('createDom should be called before positioning the line');
+    }
+    this.markerSvgLine_.setAttribute('x', `${x}`);
+    this.markerSvgLine_.setAttribute('y', `${y}`);
+    this.markerSvgLine_.setAttribute('width', `${width}`);
```
cpcallen, line 443: 1Using template literals for conversion doesn't appear to be explicitly forbidden by the styleguide, but my reading is that [using `String()` is preferred](https://google.github.io/styleguide/tsguide.html#type-coercion).

```suggestion
    this.markerSvgLine_.setAttribute('x', String(x));
    this.markerSvgLine_.setAttribute('y', String(y));
    this.markerSvgLine_.setAttribute('width', String(width));
```

cpcallen, line 443: Using template literals for conversion doesn't appear to be explicitly forbidden by the styleguide, but my reading is that [using `String()` is preferred](https://google.github.io/styleguide/tsguide.html#type-coercion).

```suggestion
    this.markerSvgLine_.setAttribute('x', String(x));
    this.markerSvgLine_.setAttribute('y', String(y));
    this.markerSvgLine_.setAttribute('width', String(width));
```

core/renderers/common/marker_svg.ts
```diff
@@ -522,18 +510,13 @@ export class MarkerSvg {
...
-    // AnyDuringMigration because:  Argument of type 'number' is not assignable
-    // to parameter of type 'string'.
-    this.markerSvgRect_!.setAttribute('height', height as AnyDuringMigration);
+    if (!this.markerSvgRect_) {
+      throw new Error('createDom should be called before positioning the rect');
+    }
+    this.markerSvgRect_.setAttribute('x', `${x}`);
+    this.markerSvgRect_.setAttribute('y', `${y}`);
+    this.markerSvgRect_.setAttribute('width', `${width}`);
+    this.markerSvgRect_.setAttribute('height', `${height}`);
```
cpcallen, line 519: Coercion again:
```suggestion
    this.markerSvgRect_.setAttribute('x', String(x));
    this.markerSvgRect_.setAttribute('y', String(y));
    this.markerSvgRect_.setAttribute('width', String(width));
    this.markerSvgRect_.setAttribute('height', String(height));
```

cpcallen (approved):
I believe this should be marked as a breaking change (see below).

Otherwise LGTM subject to some minor nits.

core/renderers/common/marker_svg.ts
BeksOmega: I don't think the order they list options in implies preference. By that logic, we should be using `Boolean()` instead of `!!`. Personally I prefer template strings because they take up less characters.

core/renderers/common/marker_svg.ts
BeksOmega: I don't think this is a breaking change because passing a puzzle tab wouldn't have worked anyway (b/c it doesn't have a `.pathLeft` property. I think the types were just defined incorrectly before.

cpcallen: Fair enough!

core/renderers/common/marker_svg.ts
cpcallen: Yes, I would use `Boolean()` rather than `!!`.

I used to like writing very, very concise code too—but then I had to read some of the code I'd written a year or two later…

BeksOmega: I think we agree to disagree on this one hehe.

core/renderers/common/marker_svg.ts
BeksOmega: I don't think we actually want to be more specific. I think the intention of this is that external developers can override it to provide their own blink properties, which may look different. The code that calls this method doesn't actually inspect the values, it just passes them to `dom.createSvg`.

PR #6409, opened 2022-09-06
chore: remove AnyDuringMigration from the other renderers

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5857 

### Proposed Changes

Removes `AnyDuringMigration` from the non-common renderers.

#### Behavior Before Change

Should be no change in user-facing behavior.

#### Behavior After Change

Should be no change in user-facing behavior.

### Reason for Changes

Tyyypes

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Dependent on #6402

### DIFF
--- a/core/renderers/common/drawer.ts
+++ b/core/renderers/common/drawer.ts
@@ -379,7 +379,7 @@ export class Drawer {
    */
   protected positionStatementInputConnection_(row: Row) {
     const input = row.getLastInput();
-    if (input && input.connectionModel) {
+    if (input?.connectionModel) {
       let connX = row.xPos + row.statementEdge + input.notchOffset;
       if (this.info_.RTL) {
         connX *= -1;
--- a/core/renderers/common/drawer.ts
+++ b/core/renderers/common/drawer.ts
@@ -189,6 +189,7 @@ export class Drawer {
    */
   protected drawStatementInput_(row: Row) {
     const input = row.getLastInput();
+    if (!input) return;
     // Where to start drawing the notch, which is on the right side in LTR.
     const x = input.xPos + input.notchOffset + (input.shape as Notch).width;
 
@@ -378,7 +379,7 @@ export class Drawer {
    */
   protected positionStatementInputConnection_(row: Row) {
     const input = row.getLastInput();
-    if (input.connectionModel) {
+    if (input && input.connectionModel) {
       let connX = row.xPos + row.statementEdge + input.notchOffset;
       if (this.info_.RTL) {
         connX *= -1;
@@ -396,7 +397,7 @@ export class Drawer {
    */
   protected positionExternalValueConnection_(row: Row) {
     const input = row.getLastInput();
-    if (input.connectionModel) {
+    if (input && input.connectionModel) {
       let connX = row.xPos + row.width;
       if (this.info_.RTL) {
         connX *= -1;
--- a/core/renderers/common/info.ts
+++ b/core/renderers/common/info.ts
@@ -60,7 +60,7 @@ export class RenderInfo {
   RTL: boolean;
 
   /** The block renderer in use. */
-  protected readonly renderer_: Renderer;
+  protected readonly renderer_!: Renderer;
 
   /** The height of the rendered block, including child blocks. */
   height = 0;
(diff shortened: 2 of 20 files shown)


core/renderers/common/drawer.ts
```diff
@@ -378,7 +379,7 @@ export class Drawer {
    */
   protected positionStatementInputConnection_(row: Row) {
     const input = row.getLastInput();
-    if (input.connectionModel) {
+    if (input && input.connectionModel) {
```
rachel-fenichel, line 382: Can be `if (input?.connectionModel)`

core/renderers/common/info.ts
```diff
@@ -534,17 +534,17 @@ export class RenderInfo {
...
       lastSpacer.width += missingSpace / 2;
-    } else if (row.align === Align.RIGHT) {
+    } else if (row.align === Align.RIGHT && firstSpacer) {
       // Add padding at the beginning of the row.
       firstSpacer.width += missingSpace;
-    } else {
+    } else if (lastSpacer) {
       // Default to left-aligning.
       lastSpacer.width += missingSpace;
     }
```
rachel-fenichel, line 550: Add a final `else` to catch the case where none of the above were true and it couldn't be left-aligned. Currently this can add `missingSpace` to `row.width` without ever adding it to one of the spacers.

core/renderers/geras/drawer.ts
```diff
@@ -136,7 +136,7 @@ export class Drawer extends BaseDrawer {
 
   override positionStatementInputConnection_(row: Row) {
     const input = row.getLastInput();
-    if (input.connectionModel) {
+    if (input && input.connectionModel) {
```
rachel-fenichel, line 139: Also can use `?.`

core/renderers/measurables/top_row.ts
```diff
@@ -96,9 +95,8 @@ export class TopRow extends Row {
       const elem = this.elements[i];
       width += elem.width;
       if (!Types.isSpacer(elem)) {
-        // AnyDuringMigration because:  Property 'isHat' does not exist on type
-        // 'typeof Types'.
-        if ((Types as AnyDuringMigration).isHat(elem) && elem instanceof Hat) {
+        // AnyDof Types'.
```
rachel-fenichel, line 98: Delete extra comment.

core/renderers/minimalist/info.ts
```diff
@@ -28,6 +28,9 @@ import type {Renderer} from './renderer.js';
  * @alias Blockly.minimalist.RenderInfo
  */
 export class RenderInfo extends BaseRenderInfo {
+  // Exclamation is fine b/c this is assigned by the super constructor.
+  protected override renderer_!: Renderer;
```
rachel-fenichel, line 32: Then why does it need to be overridden here at all, instead of being `!`ed in `BaseRenderInfo`?

rachel-fenichel (commented):
I commented on the first commit, saw that it's covered by Christopher's review in another PR, and skipped forward to the new commits.

core/renderers/common/info.ts
```diff
@@ -507,9 +501,7 @@ export class RenderInfo {
         if (missingSpace > 0) {
           this.addAlignmentPadding_(row, missingSpace);
         }
-        // AnyDuringMigration because:  Property 'isTopOrBottomRow' does not
-        // exist on type 'typeof Types'.
-        if ((Types as AnyDuringMigration).isTopOrBottomRow(row)) {
```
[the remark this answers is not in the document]
BeksOmega: It was a const we cluged methods onto, but we turned it into an actual instance of a class. https://github.com/google/blockly/pull/6299/files#diff-a4de311e9527eae2835a5f0260a304a46c490e562922dece0b6e12d919d7b8c4

core/renderers/common/info.ts
BeksOmega: Added an `else { return; }`

core/renderers/minimalist/info.ts
BeksOmega: Here `Renderer` is the minimalist renderer, not the common renderer. So we're actually overriding it to change the type. We just also need to do the assertion b/c typescript can't figure out that its being set properly.

rachel-fenichel (approved):

PR #6411, opened 2022-09-07
fix: remove calls to addClass

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6325 

### Proposed Changes

Stop using `Blockly.utils.dom.addClass`. Instead, use `element.classList.add(className)`.

#### Behavior Before Change

No change.

#### Behavior After Change

No change.

### Reason for Changes

Our `addClass` function only existed because Internet Explorer couldn't handle access/modification of the `classList` for SVG elements.

### Test Coverage

Build and opened in playground.

### Documentation
None

### Additional Information

- Later work: deprecate `addClass`, followed by removal.
- The CSS configuration in the toolbox relied on accessing properties with strings: `(this.cssConfig_ as AnyDuringMigration)['containerClass']`. I switched to using dot-access, but need confirmation that it's okay to change it.
- The first change got the simple ones and the third change got ones where I was changing access modes.
- Other followup: same thing for `removeClass`, and removal of the import for `dom.ts` wherever possible.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -785,7 +785,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       (group as AnyDuringMigration).translate_ = '';
       (group as AnyDuringMigration).skew_ = '';
       common.draggingConnections.push(...this.getConnections_(true));
-      dom.addClass(this.svgGroup_ as Element, 'blocklyDragging');
+      this.svgGroup_.classList.add('blocklyDragging');
     } else {
       common.draggingConnections.length = 0;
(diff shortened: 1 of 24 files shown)


rachel-fenichel:
Tests are failing when the class name is not a constant--`add` fails with a `SyntaxError` if the token is empty.

rachel-fenichel:
After discussion with Neil: accessing the cssConfig class names still needs to be done with quoted strings because of how they are being set.

NeilFraser (approved):

PR #6414, opened 2022-09-08
fix: message types being incorrect

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #6358 

### Proposed Changes

Fixes the message files not being typed, and setLocale not being typed.

#### Behavior Before Change

Type errors.

#### Behavior After Change

No type errors!

### Reason for Changes

Type errors are sad.

### Test Coverage

Linked to samples and tested that type errors did not occur when doing:

```
import * as  Blockly from 'blockly/core';
import * as Fr from 'blockly/msg/fr';

Blockly.setLocale(Fr);
console.log(Fr.ADD_COMMENT);
```

### Documentation

N/A (I think this is how we currently tell developers to import the message definitions).

### Additional Information

Ideally @maribethb would be able to look at this since she dug into trying to type the message files before. But she's not around, so I might try to dig up anything she commited wrt this before. For now, just wanted to put up what I had so far.


BeksOmega:
Hmm I don't know how to handle `setLocale`. It definitely feels like it should be defined in `blockly.ts` (because it's defined in both the core and browser targets) but obviously the build isn't happy with this setup.

Going to leave it broken for now while other people give opinions.

BeksOmega:
Should be good for a look @gonfunko !

gonfunko (approved):

maribethb (approved):
discussed offline, summary:
1. test that setLocale actually works (not just the type checking passes, but the page is actually in french) (this should probably be an automated test in the future)
2. test that messages still work in script-tag scenario
3. add comment that setLocale is useless in script-tag scenario
4. file a bug for follow up to possibly improve this type. in the current type you'll get type errors if using custom translations. we can maybe get the best of both worlds by using a record type where you can specify specific strings, and then maybe we can add a union with just `string` type to account for custom translations, and that would hopefully still give us autocomplete for the built in ones. https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

PR #6418, opened 2022-09-09
fix: Deep copy options used by dropdown fields

NeilFraser:
Previously the options data structure passed to the dropdown constructor was sometimes copied (in the case of prefix/suffix trimming), but other times it was merely referenced.  This meant if options data was used to create one dropdown, then the options data was modified and a second dropdown was created, the first dropdown's options would mysteriously change as well.

This uncovered a broken test that was only passing because of a side effect of the dropdown option data getting modified.


rachel-fenichel:
First commit looks good, but somehow you've also picked up a second commit. Please rebase.

PR #6423, opened 2022-09-12
fix: theme types

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #6358 

### Proposed Changes

Makes properties of the theme interfaces optional (since they are mostly optional).

This was tricky, because after the theme gets passed to the renderer, all of the properties should be "filled in". But trying to record this side-effect based change with types was going to be... difficult. So it's easier just to check that the renderer properly initialized the properties.

#### Behavior Before Change

External developers had to provide properties to themes they didn't want to override.

Should be no end-user-facing change in behavior.

#### Behavior After Change

Now you only need to provide the properties you want to override.

Should be no end-user-facing change in behavior.

### Reason for Changes

Better dev experience.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -241,7 +241,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
    *
    * @returns #RRGGBB string.
    */
-  getColourSecondary(): string|null {
+  getColourSecondary(): string|undefined {
     return this.style.colourSecondary;
   }
 
@@ -250,7 +250,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
    *
    * @returns #RRGGBB string.
    */
-  getColourTertiary(): string|null {
+  getColourTertiary(): string|undefined {
     return this.style.colourTertiary;
   }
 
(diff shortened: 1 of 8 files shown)


cpcallen (commented):
I like the general approach, but dislike the repetitive `throw` statements.

My first thought was to just create a single function to do those checks, but then I realised that you could make it a type predicate for an `InitialisedTheme` type, and I don't think you even need to reference it explicitly in any of the functions which use it.  (Though this might create inefficiency by checking more than just the single field that the caller is about to use—if the JIT does not optimise the checks away reliably.)

BeksOmega:
Going to go ahead and merge due to release schedule. We can come back and add a type predicate later if desired.

maribethb:
I think an alternative way to solve this would be to change [this line](https://github.com/google/blockly/blame/64f4013b2c0b5c9fc8283e855444d3a7e1ea42ad/core/theme.ts#L20) to use `Partial<BlockStyles>` which allows all the properties to be optional, while in all the code that uses BlockStyle the properties will now be required.

as a side note if that approach works then we should use the same type in [this line](https://github.com/google/blockly/blob/64f4013b2c0b5c9fc8283e855444d3a7e1ea42ad/core/renderers/common/constants.ts#L661) as well

this is fine to fix as a follow up after this release though i think if we are crunched for time

PR #348, opened 2016-04-27
Re-enable events even when an exception is thrown

drigz:
This prevents a recoverable exception permanently disabling events. This could be caused by calling domToWorkspace() to load a program containing an invalid block type (eg because it was saved by an old version of Blockly).

To fix that case, it only needs to be added to xml.js:domToBlock. I added it everywhere else out of caution - however, in some places, it's not necessarily sensible, as there's other cleanup without which Blockly wouldn't work properly. @NeilFraser, do you think it's sensible to add it everywhere, or only where we predict an exception happening?

PR #6424, opened 2022-09-12
fix: disposing of a workspace that has overwritten shadows

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Part of #6358 

### Proposed Changes

Change where we set `block.disposed` to `true`. I audited all of the places where `disposed` is referenced, and I don't *think* this will break anything. But let's be honest, it probably will break something.

#### Behavior Before Change

If you disposed of a workspace that had an overwritten shadow block in it, the workspace would throw an error. See https://github.com/google/blockly/issues/6358#issuecomment-1244375899

No end-user-facing change in behavior.

#### Behavior After Change

No Error!

No end-user-facing change in behavior.

### Reason for Changes

Errors are sad =(

### Test Coverage

Manually tested that disposing of a workspace with the following blocks in it did not cause an error to be thrown:
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "math_single",
        "id": "f#E]:l@J;BJ;N(?;KQ5T",
        "x": 138,
        "y": 138,
        "fields": {
          "OP": "ROOT"
        },
        "inputs": {
          "NUM": {
            "shadow": {
              "type": "math_number",
              "id": "wr:jE3$@EZ(/{pNSq92C",
              "fields": {
                "NUM": 9
              }
            },
            "block": {
              "type": "math_number",
              "id": "5A]oz4Gd$y]!BUkH/i~@",
              "fields": {
                "NUM": 123
              }
            }
          }
        }
      }
    ]
  }
}
```

### Documentation

N/A

### Additional Information

Caused by https://github.com/google/blockly/pull/6300/files#diff-dd85563e074ebb74e4c34eee01568971430f3cb7a6b846feef8044ad8d5dcd61L566

The issue is that `workspace` used to be set to `null` before the blocks were disposed, so that would properly cause `createShadowBlock_` to short circuit. But because `createShadowBlock_` was now looking at `disposed` it wasn't short circuiting, because `disposed` was set after the blocks were disposed.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -319,7 +319,6 @@ export class Block implements IASTNodeLocation, IDeletable {
     if (this.disposed) {
       return;
     }
-    this.disposed = true;
 
     this.unplug(healStack);
     if (eventUtils.isEnabled()) {
@@ -339,6 +338,7 @@ export class Block implements IASTNodeLocation, IDeletable {
       this.workspace.removeTypedBlock(this);
       // Remove from block database.
       this.workspace.removeBlockById(this.id);
+      this.disposed = true;
 
       // First, dispose of all my children.
       for (let i = this.childBlocks_.length - 1; i >= 0; i--) {
--- a/core/block.ts
+++ b/core/block.ts
@@ -181,6 +181,13 @@ export class Block implements IASTNodeLocation, IDeletable {
   protected outputShape_: number|null = null;
 
   /**
+   * Is the current block currently in the process of being disposed?
+   *
+   * @internal
+   */
+  disposing = false;
+
+  /**
    * A string representing the comment attached to this block.
    *
    * @deprecated August 2019. Use getCommentText instead.
@@ -316,7 +323,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @suppress {checkTypes}
    */
   dispose(healStack: boolean) {
-    if (this.disposed) {
+    if (this.disposing || this.disposed) {
       return;
     }
 
@@ -338,7 +345,7 @@ export class Block implements IASTNodeLocation, IDeletable {
       this.workspace.removeTypedBlock(this);
       // Remove from block database.
       this.workspace.removeBlockById(this.id);
-      this.disposed = true;
+      this.disposing = true;
 
       // First, dispose of all my children.
       for (let i = this.childBlocks_.length - 1; i >= 0; i--) {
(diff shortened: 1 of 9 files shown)


BeksOmega:
I think it might actually be safest to add an internal `disposing` property, and replace all the places where we used to be checking `block.workspace == null` with that, so that we know the new behavior is 100% consistent.

But on the other hand, I hate using state to define functional logic.

So Imma sleep on it.

core/block.ts
```diff
@@ -356,7 +357,6 @@ export class Block implements IASTNodeLocation, IDeletable {
       }
     } finally {
       eventUtils.enable();
-      this.disposed = true;
```
gonfunko, line 359: Do you know where the exception this try/catch is wrapping is thrown? Do we still need this here in case it throws prior to disposed being set to true in the try?

tests/mocha/event_test.js
```diff
@@ -997,10 +997,6 @@ suite('Events', function() {
             this.eventsFireSpy, 0, Blockly.Events.BlockDelete,
             {oldXml: expectedOldXml, group: ''},
             workspaceSvg.id, expectedId);
-        assertNthCallEventArgEquals(
```
gonfunko, line 1000: Is this removal intentional?

core/block.ts
BeksOmega: Ha no. Back in 2016 a contributor just wrapped everything between an `Events.disable()` and an `Events.enable()` in a `try...finally` to make sure that events were always re-enabled: https://github.com/google/blockly/pull/348

tests/mocha/event_test.js
BeksOmega: Yes, they are the exact same call. Probably a rebase error.

core/block.ts
```diff
@@ -180,6 +180,13 @@ export class Block implements IASTNodeLocation, IDeletable {
   protected collapsed_ = false;
   protected outputShape_: number|null = null;
 
+  /**
+   * Is the current block currently in the process of being disposed?
+   *
+   * @internal
+   */
+  disposing = false;
```
gonfunko, line 188: Now that you're no longer checking the property directly, this should probably be made private vs @internal.

gonfunko (approved):

core/block.ts
BeksOmega: Done! Should be good for another look!

PR #6426, opened 2022-09-13
feat: return element from menu.render

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Removes a need for `?.`

### Proposed Changes

Return the div from `menu.render` so that it can be used immediately and guaranteed to be non-null in calling code.

#### Behavior Before Change

No return value, so you had to call `getElement`, which could return null.

### Reason for Changes
Noticed while working on `addClass`.

### DIFF
--- a/core/menu.ts
+++ b/core/menu.ts
@@ -86,7 +86,7 @@ export class Menu {
    *
    * @param container Element upon which to append this menu.
    */
-  render(container: Element) : HTMLDivElement {
+  render(container: Element): HTMLDivElement {
     const element = (document.createElement('div'));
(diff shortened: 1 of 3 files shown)


core/menu.ts
```diff
@@ -86,7 +86,7 @@ export class Menu {
    *
    * @param container Element upon which to append this menu.
    */
```
NeilFraser, line 88: No @returns?

rachel-fenichel: d'oh. fixed.

NeilFraser (approved):

PR #6428, opened 2022-09-13
fix: call remove instead of dom.removeNode

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6325
### Proposed Changes

Use `element.remove()` instead of our `dom.removeNode` function.

#### Behavior Before Change

No change

#### Behavior After Change

No change

### Reason for Changes

Get rid of code that worked around IE issues.


cpcallen (changes requested):
> #### Behavior Before Change
> No change

The automated tests beg to differ.

Issue #6165, opened 2022-05-16
Multiline field not correctly size in parent block when copy/paste


MrAndrewMal:
Describe the bug

![printscreen](https://user-images.githubusercontent.com/26713505/168569461-aa53d1f0-0be6-4265-b4a1-580a1a740378.png)

The multinline field is not sized correctly in the another block when copy/paste.

To Reproduce

Go to the playground.
Add on playground a block which can connect with multiline block.
Paste multiline block with a text in this block.
Copy/paste the parent block

**Expected behavior**

The background rect and block should show the text in full.

BeksOmega:
Closed by #6461

Issue #6071, opened 2022-04-11
MultilineInput - Not sized properly when workspace is loaded.


nanhishweta-26:
Hi Team,

As shown in the below screenshots, when the Workspace is cleared and loaded

Multiline block is not sized properly to show all the text within the block as seen in Image1. 

But, if we try to edit the block the multiline block is resized to accommodate the text as seen in Image 2. Also, afterwards it retains the size.

Image 1:
![image](https://user-images.githubusercontent.com/56015624/162752966-0c3386bd-a4ab-4b11-a1e6-4fc9b41b4fc0.png)

Image 2:
![image](https://user-images.githubusercontent.com/56015624/162753059-3c2f82fe-815a-4656-8c33-7eb1385c3a60.png)

Any help is much appreciated.

Thanks,
Shweta

maribethb:
Hi, I'm not able to reproduce this. Can you provide more detailed reproduction steps? How are you clearing and reloading the workspace?

Thanks!

nanhishweta-26:
This is the code to reproduce the bug. Please review and let us know if you need any more details.


`blocklyWorkspace.clear();

var workspaceJSONValue =
{
    "blocks": {
        "languageVersion": 0,
        "blocks": [
            {
                "type": "modify_form",
                "id": "4PMzIgAFVO2s_-m?$:1d",
                "x": 10,
                "y": 10,
                "inputs": {
                    "modify_form": {
                        "block": {
                            "type": "variables_set",
                            "id": "ud4J|#g9tWCPdL?gO,(K",
                            "fields": {
                                "VAR": {
                                    "id": "zFF)Ar@[ZdrIxEL}n)Ao"
                                }
                            },
                            "inputs": {
                                "VALUE": {
                                    "block": {
                                        "type": "text_multiline",
                                        "id": "|YCm`eZO3k?7g/dw0hb(",
                                        "fields": {
                                            "TEXT": "This is a very long message"
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        ]
    },
    "variables": [
        {
            "name": "Message",
            "id": "zFF)Ar@[ZdrIxEL}n)Ao"
        }
    ]
}

Blockly.serialization.workspaces.load(workspaceJSONValue, blocklyWorkspace);`

Juliandebruin:
Is there a workaround for now? I am having the same issue

shals1031:
Hi Blockly team

[...]

PR #6429, opened 2022-09-14
fix: Multiline field not correctly size in parent block when copy/paste or load.

yamadayutaka:
## The basics

- [x] I branched from develop

### Resolves

- #6165 
- #6071 

### Proposed Changes

When loading a block with input blocks or next blocks, render the blocks last.

#### Behavior Before Change

Multiline field not correctly size in parent block when copy/paste or load.

#### Behavior After Change

Multiline field correctly size in parent block when copy/paste or load.

### Reason for Changes

These issues are from v7.
So I modified the processing order to be similar for v6.

https://github.com/google/blockly/blob/6.20210701.0/core/xml.js#L564-L569

### DIFF
--- a/core/serialization/blocks.ts
+++ b/core/serialization/blocks.ts
@@ -390,13 +390,17 @@ export function appendInternal(
  * @param param1 parentConnection: If provided, the system will attempt to
  *     connect the block to this connection after it is created. Undefined by
  *     default. isShadow: The block will be set to a shadow block after it is
- *     created. False by default.
+ *     created. False by default. loadedBlocks: An array to store the loaded
+ *     blocks. Empty by default.
  * @returns The block that was just appended.
  */
 function appendPrivate(
     state: State, workspace: Workspace,
-    {parentConnection = undefined, isShadow = false}:
-        {parentConnection?: Connection, isShadow?: boolean} = {}): Block {
+    {parentConnection = undefined, isShadow = false, loadedBlocks = []}: {
+      parentConnection?: Connection,
+      isShadow?: boolean,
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@BeksOmega can you take a look at the serialization changes?

rachel-fenichel:
Hi @yamadayutaka--sorry for the delay, but the person who knows the serialization code best is currently out of office. We haven't forgotten about your change.

BeksOmega:
Hi @yamadayutaka ! Thank you so much for looking into this :D Sorry it took me so long to get back to you!

I was wondering, instead of refactoring the serialization order, could we try using [getFastTextWidth](https://github.com/google/blockly/blob/55a707662409a8829416e38e7655bb5d009aca39/core/utils/dom.js#L333) in [updateSize_](https://github.com/google/blockly/blob/55a707662409a8829416e38e7655bb5d009aca39/core/field_multilineinput.js#L286) for the FieldMultilineInput? This is what the [FieldTextInput uses](https://github.com/google/blockly/blob/55a707662409a8829416e38e7655bb5d009aca39/core/field.js#L739) (via Field), which is why it doesn't run into the same problem.

I haven't looked at the serialization-based solution in depth, but I would like to try fixing the field itself first if possible.

yamadayutaka:
Hi @BeksOmega,
Thank you for reviewing my PR.
I will try it.

yamadayutaka:
I've confirmed that what you suggested works as expected, so I created another PR.
#6461

BeksOmega:
Closing in favor of #6461

PR #6441, opened 2022-09-22
chore: Remove radix from parseInt, simplify Blockly.utils.dom methods, use Unicode characters.

NeilFraser:
parseInt guesses the radix if one was not provided.  While this is normally 10, a prefix of '0x' will switch it to hexadecimal -- this is fine.  But more problematic was that a prefix of '0' would switch it to octal.  Thus `parseint('018')` would return `1` since there's a leaning zero, and parsing would stop at the first invalid character (eight).  This caused a lot of subtle errors and as a result adding a radix used to be required.

However, years ago implicit octal detection was removed from parseInt.  Thus we no longer need to worry about it.

### DIFF
--- a/core/events/events_abstract.ts
+++ b/core/events/events_abstract.ts
@@ -42,7 +42,7 @@ export abstract class Abstract {
   /** @alias Blockly.Events.Abstract */
   constructor() {
     /**
-     * The event group id for the group this event belongs to. Groups define
+     * The event group ID for the group this event belongs to. Groups define
      * events that should be treated as an single action from the user's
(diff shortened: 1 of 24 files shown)


cpcallen:
> But more problematic was that a prefix of '0' would switch it to octal… However, years ago implicit octal detection was removed from parseInt.

Specifically, this was changed in ES5.

cpcallen (approved):
Please update PR title and description to include changes made by additional commits.

Please ensure squash commit gets new PR title.

PR #6475, opened 2022-10-04
chore(build): Remove build products from the Blockly repository

### LINKED ISSUE #6403: Remove build products from this repository
This is a tracking bug for work to update our build system so that we can remove all checked-in build products from the google/blockly repository.

## Background

The blockly repository contains not only source files but also build products, including the `*_compressed.js` (and `.map.js`) files.  This is very convenient for being able to quickly fire up (e.g.) the Blockly Playground after doing a `git clone`, but creates some problems:

* Previously, every build would potentially modify these files, creating a lot of confusion about what should and shouldn't be included in any commit.
  * This was fixed (for Closure Compiler output) by building to `build/` and adding the `npm run checkin` command.
* Now, we are in a situation where it is very easy to inadvertently be using stale checked-in build output rather than the freshly-generated copies.
  * This has occasionally caused quite a lot of confusion, particularly around running the playground in compiled mode (which uses the checked-in copies at least until #6218  is finished.
  * Additionally, there has been some desire to check in additional build product, such as the wrapped language files.

It is probably preferable not to have any build output checked into the repository, but to generate it when required.

## Outline Proposal
* Remove all generated files from the repository.
* Have `build/` contain only ~temporary build products that are not part of the final release—e.g.:
  * The `.js` files output from `tsc`.
  * The un-wrapped language files currently checked in at `msg/js/`.
* Have `dist/` contain (the only copy of) each "final" build product (e.g. `blockly_compressed.js`).  The contents of `dist/` would be the canonical place to load blockly from:
  * Test/playground code in the Blockly repository would load Blockly from this directory directly.
  * The contents of this directory will (continue to) be packaged as an NPM package.
  * The contents of this directory could additionally be attached to the GitHub release, to provide an easy way for non-NPM users to download a ready-to-use Blockly as a `.zip` file.
* The only exception will be test/playgrounds running in uncompiled mode (for edit/compile/test velocity) that would continue to to use files in `build/` where appropriate.
* `npm run build` would write some things that it currently writes to `build/` directly to `dist/`.
* `npm run package` would add whatever additional files (e.g. `package.json`) that are needed for release but not for testing.
* `npm run prepare`, which happens automatically when doing `npm install`, would be configured to do `npm run build` (or a suitable subset thereof) to ensure that previously checked-in build products have been built, allowing a "clone and go" approach to continue to work (just now "clone, install and go").
* Additionally, to reduce the potential for human error when building, packaging and publishing Blockly, each step of the build / package / publish process would be configured to run all the previous steps.  So for example in the following list each target includes all the previous targets (this is not a comprehensive list):
    * `build:js` (runs `tsc`)
    * `build:deps` (runs `closure-make-deps`)
    * `build:compiled` (runs `google-closure-compiler`; this target will thus become synonymous with `build` and be removed.)
    * `package` (copies files into `dist/`)
    * `publish` (publishes contents of `dist/`)
    This is already the case for `build:js` / `build:deps` / `build` targets, so this is just extending the same principle to the full set of primary targets.

## To Do list

* [x] Modify gulp targets so that each runs all the ones it depends on, to avoid any inconsistency in build/packaged files.
  * [x] Consider whether to rename gulp scripts to reflect the new arrangements.
    * For example, originally the `build:xyz` targets were all separate non-overlapping parts of the `build` target, but that will no longer be the case, so a different naming system might be advisable; perhaps e.g.  just `tsc` / `deps` / `bundle`.
* [x] Modify `package` target to run the `cleanBuildDir` and `cleanReleaseDir` targets (i.e. the same as `npm run clean`) before any of the other build steps to ensure that no extraneous files are inadvertently included in `dist/` and thus the package.
* [x] Modify `publish` script to run `npm ci` before invoking the gulp script to ensure that dependencies are at the specified version.
* [x] Modify `tests/run_all_tests.sh` to just run `npm run package` once at the start (for efficiency, since `package` will now invoke `build`).
* [x] Modify gulp targets to write to `dist/` instead of `build/` where appropriate.
  * [x] Remove obsolete copying code from `package` gup targets.
* [ ] Modify tests to load `dist/msg/en.js` instead of `msg/messages.js`.
* [x] Modify `tests/bootstrap.js` to load from `dist/` (#6218).
* [x] Remove  build products from the repository:
  * [x] `*_compressed.js` and `.js.map`
  * [x] `msg/js/*`
  * [ ] (any others?)
* [x] Update the procedures for pushing our GitHub Pages to include checking in the relevant build products to the `gh_pages` branch so that our GitHub Pages will continue to work.

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Resolves #6403

### Proposed Changes

* Divide gulp targets into three kinds: main sequence,  manually invokable, and script-only.  The first two categories automatically invoke their prerequisites.
  * Give some of the affected gulp targets shorter and more memorable names that could (depending on feedback on this PR) become their npm script names.
  * Have the `package` task invoke the `cleanBuildDir` (as well as `cleanPackageDir`) and `build` tasks.
  * Remove the `checkBuildDir` task as it is now redundant since a fresh build is done every time.
  * Have `npm test` run just the `package` target (with debug flags), since that runs the the `build` target automatically.
* Have any npm script that has external effects (e.g. publishing an npm package, pushing a new version to App Engine, or updating GitHub Pages) start by running `npm ci` to ensure that all dependencies are up-to-date with respect to `package-lock.json`.
  * This is done by npm and not a gulp script because `gulp` itself might need updating.  (So might `npm`, but that is less likely to make any difference to what gets published/pushed.)
* Have Closure Compiler write its output directly to `RELEASE_DIR` (`dist/`) since it comes out already UMD-wrapped.
* Have `build:langfiles` write to `build/msg/` instead of `build/msg/js`, for consistency with ultimate destination for wrapped files (`dist/msg/`).
* Modify `tests/bootstrap.js` to, when loading in compressed mode, use the freshly-built `dist/blockly_compressed.js` instead of the probably-stale checked in version in the repository root (and similarly for all the other chunks).
* Modify playgrounds to use `build/msg/en.js` instead of `msg/messages.js`.
  * All playgrounds load messages in a way that does not depend on the UMD wrapper.
* Modify demos, appengine deployment and GitHub pages release tasks to use files in `build/` and/or `dist/`.
* Remove all checked-in build products.

#### Behaviour Before Change

It was fairly easy to test or release outdated files by accident.

#### Behaviour After Change

It should be more or less impossible to inadvertently publish an out-of-date file.  Playgrounds etc. will only load the most recently-built versions of their Blockly dependencies.

### Reason for Changes

It is not generally considered good practice to check build products into a source code repository; there are other better means, such as npm packages and GitHub releases, to publish released "binaries".

### Documentation

There will need to be updates made to the release process documentation.


package.json
```diff
@@ -32,15 +32,15 @@
...
     "deployDemos": "gulp deployDemos",
     "deployDemos:beta": "gulp deployDemosBeta",
-    "format": "gulp clangFormat",
+    "format": "gulp format",
     "format:sortrequires": "gulp sortRequires",
     "generate:langfiles": "gulp generateLangfiles",
     "license": "gulp checkLicenses",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
+    "only:compile:advanced": "gulp onlyBuildAdvancedCompilationTest --debug",
```
BeksOmega, line 42: Not sure I like `only` as a prefix. Would `test:compile:advanced` work?

scripts/gulpfiles/build_tasks.js
```diff
@@ -753,17 +741,23 @@ function format() {
...
-  compiled: buildCompiled,
-  format: format,
-  checkinBuilt: checkinBuilt,
-  cleanBuildDir: cleanBuildDir,
-  advancedCompilationTest: buildAdvancedCompilationTest,
-}
+// Main sequence targets.  Each should invoke any immediate prerequisite(s).
+exports.cleanBuildDir = cleanBuildDir;
+exports.messages = buildLangfiles;
+exports.tsc = buildJavaScript;
```
BeksOmega, line 733: Is there any reason not to have the exported identifiers match the internal identifiers? Seems like an annoying level of indirection.

scripts/gulpfiles/git_tasks.js
```diff
@@ -94,7 +95,11 @@ function pushRebuildBranch(done) {
   done();
 }
 
-// Update github pages with what is currently in develop.
+/**
+ * Update github pages with what is currently in develop.
+ *
+ * Prerequisite: build.
```
BeksOmega: Prereq: package ?

package.json
```diff
@@ -44,16 +44,16 @@
     "package": "gulp package",
     "prepare": "gulp prepare",
     "prepareDemos": "gulp prepareDemos",
-    "publish": "gulp publish",
-    "publish:beta": "gulp publishBeta",
+    "publish": "npm ci && gulp publish",
+    "publish:beta": "npm ci && gulp publishBeta",
     "recompile": "gulp recompile",
     "release": "gulp gitCreateRC",
```
BeksOmega, line 49: What does this do? Does this also need an `npm ci`?

tests/run_all_tests.sh
```diff
@@ -56,7 +56,7 @@ fi
 
 # Run the full usual build process, checking to ensure there are no
 # closure compiler warnings / errors.
-run_test_command "build-debug" "npm run build-debug"
```
BeksOmega, line 59: Tested that this gives you the debug output? Just want to make sure it has the correct number of `--`s lol

scripts/gulpfiles/package_tasks.js
```diff
@@ -396,9 +386,8 @@ const package = gulp.series(
         packageLua,
         packageDart,
         packagePHP,
-        packageLocales,
         packageMedia,
-        packageUMDBundle,
+        gulp.series(packageLocales, packageUMDBundle),
```
BeksOmega, line 390: Why do these have to be in series now?

demos/mobile/android/app/build.gradle
```diff
@@ -28,16 +28,30 @@ task copyBlocklyHtmlFile(type: Copy) {
     rename('index.html', 'webview.html')
 }
 
-task copyBlocklyMoreFiles(type: Copy) {
```
BeksOmega, line 31: I didn't even know these demos existed lol

scripts/gulpfiles/appengine_tasks.js
```diff
@@ -14,9 +14,11 @@ var fs = require('fs');
 var rimraf = require('rimraf');
 var path = require('path');
 var execSync = require('child_process').execSync;
+const buildTasks = require('./build_tasks.js');
+const packageTasks = require('./package_tasks.js');
 
 var packageJson = require('../../package.json');
-
+var packageJson = require('../../package.json');
```
BeksOmega, line 21: Dupe?

scripts/gulpfiles/git_tasks.js
```diff
@@ -109,18 +109,20 @@ const updateGithubPages = gulp.series(
...
+  packageTasks.cleanReleaseDir,
   buildTasks.build,
-  buildTasks.checkin,
   function(done) {
-    execSync('git commit -am "Rebuild"', { stdio: 'inherit' });
-    execSync('git push ' + upstream_url + ' gh-pages --force', { stdio: 'inherit' });
+    execSync('git add build/msg/* dist/*_compressed.js*', {stdio: 'inherit'});
+    execSync('git commit -am "Rebuild"', {stdio: 'inherit'});
+    execSync('git push ' + upstream_url + ' gh-pages --force',
+        {stdio: 'inherit'});
```
BeksOmega, line 118: Very cool! Love this handling of the ghpages.

package.json
```diff
@@ -31,7 +31,6 @@
     "clean": "gulp clean",
     "clean:build": "gulp cleanBuildDir",
     "clean:release": "gulp cleanReleaseDir",
-    "checkin": "gulp checkin",
```
BeksOmega, line 34: Could we remove some more of these? e.g. now that everything properly calls prereqs, is there any reqson to call `clean:build` manually?

tests/run_all_tests.sh
cpcallen: Yes, tested.  It took a bit of work to get Closure Compiler to generate a diagnostic, actually, now that we are suppressing most type errors…

package.json
cpcallen: No; this only does some `git` commands.

package.json
cpcallen: The point is to have `test:compile:advanced` (which is intended to be invoked manually) do all required build steps (in this case: run `tsc` and `closure-make-deps`), but have `only:compile:advanced` (which is intended to be invoked only from `run_all_tests.js`) skip those steps because they'll have just been done in a previous step of the tests.

Luckily, with PR #6431 I think this can be eliminated, since the `test` task should be able to directly invoke the  `buildAdvancedCompilationTest` task.

scripts/gulpfiles/build_tasks.js
cpcallen: I agree that as-is this is annoying.  My plan was to update the names of the npm scripts to match, but hadn't actually done that yet.

Do you have an opinion about changing the names of the scripts to match the tasks, i.e.:

`npm run tsc` instead of `… build:javascript`, `… deps` instead of `build:deps`, etc.

  If you don't like these names then I will revert the task renames for now.

scripts/gulpfiles/package_tasks.js
cpcallen: Because `blockly.min.js` includes the (wrapped) version of `msg/en.js`.

scripts/gulpfiles/build_tasks.js
BeksOmega: Changing the names of the scripts to match the tasks sgtm! I would just like them to be matchy-matchy and descriptive, I don't care what direction changes in hehe.

cpcallen:
I believe I'm now done messing around with this.  Apologies for the slightly convoluted history; I changed my mind about teh contents of the last couple of commits several times…

BeksOmega (approved):
LGTM! One last thing: could you add a comment smewhere explaining the difference between messages and langfiles? Maybe in `gulpfile.js`?

PR #6488, opened 2022-10-05
feat: add interface definitions for procedure stuff

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Adds definitions for interfaces related to the shared procedures project.

### Reason for Changes

 Interfaces are useful for documenting intentions, and they also keep the code flexible.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A


core/interfaces/i_procedure_map.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+import * as goog from '../../closure/goog/goog.js';
+goog.declareModuleId('Blockly.IProcedureMap');
+
+import {IProcedureModel} from './i_procedure_model.js';
+
+/**
+ * A special collection of procedure models.
+ */
+export interface IProcedureMap {
```
cpcallen, line 20: Is there any reason not to just use `Map<string, IProcedureModel>` instead of this type?

That would seem to have various advantages: simplicity, principally, but also access to all of the excellent dynamic iteration facilities of `Map`, such as the feature that `for ([key, value] of map)` and `for value of Map.values())` correctly handling the case that items are deleted from the map during iteration.  (And if you really don't want that you can always `for (value of Array.from(…))`.)

cpcallen: OK, I've gone and re-read the relevant section of the design doc.  I understand now why you have defined this, but I think a better approach would be to define a (concrete, general-purpose) `ObservableMap` subclass of `Map` that adds hooks to all the normal `Map` methods, and use instances of that—using variables/properties/parameters typed as `Map<…>` to hide the observability where appropriate.

core/interfaces/i_parameter_model.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+import * as goog from '../../closure/goog/goog.js';
+goog.declareModuleId('Blockly.IParameterModel');
+
+import {VariableModel} from '../variable_model.js';
+
+/**
+ * A data model for a procedure.
+ */
+export interface IParameterModel {
```
cpcallen, line 20: I note that [the JS style guide still prohibits Hungarian notation](https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers).  I'd be interested to know what the TS Stylist's opinion about this is.

core/interfaces/i_procedure_model.ts
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Inserts a parameter into the list of parameters.
+   *
+   * To move a parameter, first delete it, and then re-insert.
+   */
+  insertParameter(parameterModel: IParameterModel, index: number):
+      IParameterModel;
+
+  /** Removes the parameter at the given index from the parameter list. */
+  deleteParameter(index: number): IProcedureModel;
```
cpcallen, line 33: Might it be safer for this to take the actual parameter object to be deleted?  Admittedly the implementation might then need to do an array scan internally, but it could avoid issues with unexpectedly interleaved updates to that array.

cpcallen (commented):
It's clear from the comments below that I skimmed a bit too much when reading your design doc previously.  I am not asserting that the design should necessarily be changed, but evidently I do think that it could use a bit more scrutiny (and in places justification).

core/interfaces/i_parameter_model.ts
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Sets the variable associated with the parameter model.
+   *
+   * Parameters have an identity (represented by the ID) which is separate from
+   * both their position in the parameter list, and the variable model they are
+   * associated with. The variable they are associated with can change over time
+   * as the human-readable parameter name is renamed.
+   */
+  setVariable(variable: VariableModel): IParameterModel;
```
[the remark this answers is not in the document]
cpcallen: Additionally: why is the return type not `void`?

core/interfaces/i_parameter_model.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * The interface for the data model of a procedure parameter.
+ *
+ * @namespace Blockly.IParameterModel
+ */
+import * as goog from '../../closure/goog/goog.js';
+goog.declareModuleId('Blockly.IParameterModel');
```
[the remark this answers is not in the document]
BeksOmega: Is there any reason not to add them, and then just remove them in bulk later? It seems like not adding them just creates the possibility to cause more confusion. Eg "why isn't this thing working!? ohhh it's one of the 2% of modules that aren't declaring their ID"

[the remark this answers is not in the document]
cpcallen: My thought was that if we don't add the declaration then we can't inadvertently add a `goog.require` that depends on it.  And we're definitely in the business of getting rid of `goog.require`s.

BeksOmega:
> * First of all I should declare up front that I have a general aversion to that style of interface.  There are some situations where it seems justified, but it adds considerable potential for confusion, especially if not all the methods on the class return `this`.

Cool cool, good to have context.

I understand how this could cause confusion, and I am amenable to changing the return type of `insertParameter`. However, in general, I think that a fluent interface is nice:

```
workspace.getProcedureMap()
    .addProcedure(new ObservableProcedureModel(workspace))
    .insertParameter(new ObservableParameterModel(workspace, new VariableModel(workspace)), 0)
```

as opposed to:
```
const proc = new ObservableProcedureModel(workspace);
proc.insertParameter(new ObservableParameterModel(workspace, new VariableModel(workspace)), 0);
workspace.getProcedureMap().addProcedure(proc);
```

I (and I think most people?) find that a fluent API is more readable, because each action leads logically to the next.

>   * E.g. at what point in the long chain of `obj.x().y().z()…` did we stop dealing with the original `obj`?  How would one guess that it was at `.insertParameter()`??

Like I said, I am amenable to changing `.insertParameter` but I'm not concerned about people "guessing" where the return type changes. IDEs will tell you the return types of functions, and will provide intellisense suggestions based on that.

This works similarly for our `.appendInput` and `.appendField` methods on `Block`.

> * Since it's easier to add a return value than take one away, I'd always err on the side of returning `void` until there is a compelling case to do otherwise.
>   * One might find that one wants to return something _other_ than `this`, after all, and then one is in real trouble!

In generally I appreciate being extremely flexible/cautious with the API. But in this case, I don't think that's a change that's likely to happen. E.g. I can't see a situation where I'd want to return something different from `addProcedure` or `setReturnType`.

> * I think it's better to write the non-fluent client code and only consider switching to a fluent interface.  (Maybe you already did this and have the result squirrelled away in another branch?)

No, that's what the design doc is for :P

core/interfaces/i_parameter_model.ts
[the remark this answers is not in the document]
BeksOmega: Ahhh ok that makes sense. Will remove o7

core/interfaces/i_procedure_map.ts
BeksOmega: I don't think having a general-purpose map makes sense, because the events that this will be firing are specific to procedures. However I think subclassing `Map` is fine. Generally I prefer composition over inheritance because I think inheritance confuses subtyping with code reuse. But in this case... I'd be fine with it.

core/interfaces/i_parameter_model.ts
BeksOmega: Yeee, but we've discussed this several times, and decided to go with it for consistency for now.

core/interfaces/i_procedure_model.ts
BeksOmega: I would rather force people to do an array scan externally if they feel the need, rather than performing one internally every time we do an operation.

cpcallen: I guess that's reasonable enough.  And the ordinal position of a parameter is (in most languages, at least) one of its most fundamental properties.

core/interfaces/i_parameter_model.ts
```diff
@@ -23,9 +23,9 @@ export interface IParameterModel {
   setName(name: string): IParameterModel;
 
   /**
-   * Sets the type(s) of this parameter to the given types.
+   * Sets the type of this parameter to the given type.
    */
-  setTypes(types: string[]): IParameterModel;
+  setType(types: string): IParameterModel;
```
cpcallen, line 28: Hmm.  Why are parameters now constrained to a single type?

core/interfaces/i_procedure_map.ts
cpcallen: > However I think subclassing Map is fine.

Are you planning to make the relevant changes in this PR (i.e., removing IProcedureMap or making it extend `Map`)?

core/interfaces/i_procedure_model.ts
```diff
@@ -0,0 +1,71 @@
...
+  /** Sets the human-readable name of the procedure. */
+  setName(name: string): IProcedureModel;
+
+  /**
+   * Inserts a parameter into the list of parameters.
+   *
+   * To move a parameter, first delete it, and then re-insert.
+   */
+  insertParameter(parameterModel: IParameterModel, index: number):
+      IParameterModel;
```
cpcallen, line 29: Did you decide about the return type here?  Given that one could put a bunch of fluent method calls on the ParameterModel in the first argument, I don't see a compelling reason to return the `IParameterModel`.  It seems to me more useful to be able to add multiple arguments fluently (though I suppose that in practice this would probably be done one-per-iteration in a loop).

core/interfaces/i_procedure_map.ts
BeksOmega: Done! Removed `IProcedureMap`

core/interfaces/i_parameter_model.ts
BeksOmega: Fixed.

core/interfaces/i_parameter_model.ts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * The interface for the data model of a procedure parameter.
+ *
+ * @namespace Blockly.IParameterModel
+ */
+
+import {VariableModel} from '../variable_model.js';
```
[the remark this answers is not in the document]
BeksOmega: Done.

core/interfaces/i_procedure_model.ts
BeksOmega: Discussed and decided on `IProcedureModel`

core/interfaces/i_parameter_model.ts
```diff
@@ -18,12 +18,12 @@ export interface IParameterModel {
   /**
    * Sets the name of this parameter to the given name.
    */
-  setName(name: string): IParameterModel;
+  setName(name: string): this;
```
cpcallen, line 21: Nice.

cpcallen (approved):

Issue #6017, opened 2022-03-22
Mutating children does not cause parents to bump neighbours


BeksOmega:
**Describe the bug**

Blockly tries to make sure blocks don't appear to be connected when they aren't by "bumping neighbours" when the shape of a block changes. Mutating a child block causes the shape of a parent block to change, but it doesn't trigger bumping.

**To Reproduce**

1. Go to the playground
2. Drag out some print blocks and some list create blocks, so that it looks like this:
![image](https://user-images.githubusercontent.com/25440652/159510651-749d4528-e2ae-4740-8609-0ee99e432685.png)

4. Add two new items to the list create block using the mutator.
5. Observe how bumping does not occur.
![image](https://user-images.githubusercontent.com/25440652/159510721-1efe140f-a182-450d-96f2-802127fe5da5.png)

**Expected behavior**

Bumping should occur so that blocks don't appear connected when they aren't.

**Additional context**

Similar mutator bumping issues: #76, #4448

PR #6538, opened 2022-10-12
fix: parent blocks not bumping neighbours

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#6017 

### Proposed Changes

makes it so that whenever a block in a block stack has `bumpNeighbours` called on it, all of the connections in the block stack bump.

We have to bump every single connection, because with custom renderers, we can't predict how blocks will be shaped based on other blocks. E.g. in geras higher sibling connections don't move, but in other renderers they might. So we have to trigger bumping on everything.

#### Behavior Before Change

Parent blocks would not bump neighbours when a child block was rerendered.

#### Behavior After Change

Parent blocks do bump neighbours when a child block is rerendered.

### Reason for Changes

Having blocks appear tto be connected (visually) without actually being connected (in terms of code generation) can cause confusing results and frustration for end-users. This helps eliminate that.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Also did some code cleanup of `bumpNeighbours` while I was in the area.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1565,9 +1565,12 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       }
 
       for (const neighbour of conn.neighbours(config.snapRadius)) {
+        // Don't bump away from things that are in our stack.
         if (neighbourIsInStack(neighbour)) continue;
+        // If both connections are connected, that's fine.
         if (conn.isConnected() && neighbour.isConnected()) continue;
 
+        // Always bump the inferior connection.
         if (conn.isSuperior()) {
           neighbour.bumpAwayFrom(conn);
         } else {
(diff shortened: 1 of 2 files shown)


core/block_svg.ts
```diff
@@ -1535,45 +1535,43 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
...
+    for (const conn of this.getConnections_(false)) {
+      if (conn.isSuperior()) {
+        // Recurse down the block stack.
+        conn.targetBlock()?.bumpNeighboursInternal();
       }
 
-      const neighbours = connection.neighbours(config.snapRadius);
-      for (let j = 0, otherConnection; otherConnection = neighbours[j]; j++) {
-        const renderedOther = otherConnection as RenderedConnection;
-        // If both connections are connected, that's probably fine.  But if
```
maribethb, line 1565: I think this comment in particular and probably the other ones are helpful context to leave in.

maribethb (approved):

PR #6542, opened 2022-10-13
fix: make getSourceBlock nullable again

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

https://groups.google.com/g/blockly/c/IxL-Px8GMXA/m/EK5g2VQ8BwAJ?utm_medium=email&utm_source=footer

### Proposed Changes

Makes getSourceBlock nullable again.

#### Behavior Before Change

No user-facing change in behavior.

#### Behavior After Change

No user-facing change in behavior.

### Reason for Changes

We were breaking people who were using getSourceBlock in callbacks that were called before the field was attached to a block. 

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -16,7 +16,7 @@ goog.declareModuleId('Blockly.FieldDropdown');
 
 import type {BlockSvg} from './block_svg.js';
 import * as dropDownDiv from './dropdowndiv.js';
-import {FieldConfig, Field} from './field.js';
+import {FieldConfig, Field, UnattachedFieldError} from './field.js';
 import * as fieldRegistry from './field_registry.js';
 import {Menu} from './menu.js';
 import {MenuItem} from './menuitem.js';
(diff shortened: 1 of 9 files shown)


core/field_dropdown.ts
```diff
@@ -217,16 +217,16 @@ export class FieldDropdown extends Field {
...
 
   /** Create a tspan based arrow. */
   protected createTextArrow_() {
     this.arrow_ = dom.createSvgElement(Svg.TSPAN, {}, this.textElement_);
     this.arrow_!.appendChild(document.createTextNode(
-        this.getSourceBlock().RTL ? FieldDropdown.ARROW_CHAR + ' ' :
-                                    ' ' + FieldDropdown.ARROW_CHAR));
-    if (this.getSourceBlock().RTL) {
+        this.getSourceBlock()?.RTL ? FieldDropdown.ARROW_CHAR + ' ' :
+                                     ' ' + FieldDropdown.ARROW_CHAR));
```
maribethb, line 228: ew clang why are we doing vertical alignment 

(no action needed just, ew clang)

core/procedures.ts
```diff
@@ -180,14 +180,21 @@ export function isNameUsed(
  * @alias Blockly.Procedures.rename
  */
 export function rename(this: Field, name: string): string {
+  const block = this.getSourceBlock();
+  if (!block) {
+    throw new Error(
+        'The field has not yet been attached to its input. ' +
```
maribethb, line 186: We're not doing this anywhere yet so maybe you don't want to do it willy-nilly.

but we could have a new error like UnattachedFieldError that is defined somewhere, with this message as default. Then you wouldn't have to keep repeating this string everywhere, and you can just `throw new UnattachedFieldError()` and even pass in the name of the field to the error message if you wanted.

maribethb (approved):

PR #6550, opened 2022-10-14
chore: added type information to `core/field_dropdown.ts`

btw17:
## The basics

- [x] I branched from develop

### Proposed Changes

Provided type information in place of `AnyDuringMigration` for `core/field_dropdown.ts`

#### Behavior Before Change

Less type information.

#### Behavior After Change

More type information.

### Reason for Changes

Cleaning up types that weren't included in the TS migration.

### Test Coverage

N/A

### Documentation

Yes - just the automatically generated documentation that utilizes the TS types.

### Additional Information

I added comments to the file changes with anything of note. There are a couple of spots where the implementation of a few functions has been updated, though it shouldn't impact the functionality.


core/field_dropdown.ts
```diff
@@ -114,30 +111,29 @@ export class FieldDropdown extends Field {
...
+      opt_config?: FieldConfig);
+  constructor(menuGenerator: Sentinel);
+  constructor(
+      menuGenerator: MenuGenerator|Sentinel, opt_validator?: Function,
+      opt_config?: FieldConfig) {
     super(Field.SKIP_SETUP);
 
-    // If we pass SKIP_SETUP, don't do *anything* with the menu generator.
-    if (menuGenerator === Field.SKIP_SETUP) {
+    if (isMenuGenerator(menuGenerator)) {
```
btw17, line 122: As noted in the comment for `isMenuGenerator`, there's a quirk with type guarding for `Sentinel` so I shifted the order around.

core/field_dropdown.ts
```diff
@@ -114,30 +111,29 @@ export class FieldDropdown extends Field {
...
+      if (Array.isArray(menuGenerator)) {
+        validateOptions(menuGenerator);
+        const trimmed = trimOptions(menuGenerator);
```
btw17, line 125: Before, `trimOptions` was mutating the deep copied menuGenerator and, when no `ImageProperties` labels were provided, setting a new value for `menuGenerator_`. Assuming there's not an edge case I'm missing where the mutation is needed, this update shifts setting the value for `menuGenerator_` to one point, and it should eliminate the need for a deep copy.

core/field_dropdown.ts
```diff
@@ -227,14 +223,9 @@ export class FieldDropdown extends Field {
         this.getSourceBlock().RTL ? FieldDropdown.ARROW_CHAR + ' ' :
                                     ' ' + FieldDropdown.ARROW_CHAR));
     if (this.getSourceBlock().RTL) {
-      // AnyDuringMigration because:  Argument of type 'SVGTSpanElement | null'
-      // is not assignable to parameter of type 'Node'.
-      this.getTextElement().insertBefore(
-          this.arrow_ as AnyDuringMigration, this.textContent_);
+      this.getTextElement().insertBefore(this.arrow_, this.textContent_);
```
btw17, line 226: `this.arrow_` apparently changed at some point since being given `AnyDuringMigration` since it worked when I just removed `AnyDuringMigration`.

btw17:
Looks like there were a few commits from the `master` branch I pulled over into this PR that weren't in `develop`. I'm assuming they'll be merged into `develop` in a bit, at which point I'll clean up this PR.

maribethb:
> Looks like there were a few commits from the `master` branch I pulled over into this PR that weren't in `develop`. I'm assuming they'll be merged into `develop` in a bit, at which point I'll clean up this PR.

Unfortunately this won't be a timely assumption, likely until our next release. It would be best to rebase onto develop.

btw17:
> > Looks like there were a few commits from the `master` branch I pulled over into this PR that weren't in `develop`. I'm assuming they'll be merged into `develop` in a bit, at which point I'll clean up this PR.
> 
> Unfortunately this won't be a timely assumption, likely until our next release. It would be best to rebase onto develop.

If `develop` won't be until the next major release, would it be fine to merge this into `master` instead of `develop` (once commits are cleaned up accordingly)? The idea is there wouldn't be a noticeable functional change here.

maribethb:
> would it be fine to merge this into `master` instead of `develop` (once commits are cleaned up accordingly)? 

Unfortunately no, that's just not how our workflow works. All PRs should go into develop as that is what we test with before merging a rollup of changes into master. Generally, master is equivalent with whatever code is currently on npm. Pushing something straight to master would break that assumption.

> If `develop` won't be until the next major release, 

`master` won't be merged into `develop` until the next release (major, minor, or patch*). That is another reason we can't push to master, because everyone else is working from develop. Someone else could change this line in develop, creating a merge conflict at the time of our next release.


* We don't always merge master back into develop when doing a cherry-pick patch release, but my point is that it's not just during major releases.

core/field_dropdown.ts
```diff
@@ -114,30 +113,29 @@ export class FieldDropdown extends Field {
    * @throws {TypeError} If `menuGenerator` options are incorrectly structured.
    */
   constructor(
-      menuGenerator: AnyDuringMigration[][]|Function|Sentinel,
-      opt_validator?: Function, opt_config?: FieldConfig) {
+      menuGenerator: MenuGenerator, opt_validator?: Function,
+      opt_config?: FieldConfig);
+  constructor(menuGenerator: Sentinel);
+  constructor(
```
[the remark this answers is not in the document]
btw17: For function overloading in TS, the final function's parameters need to support all of the previous functions' parameters. The final constructor supports `MenuGenerator|Sentinel` since those are the cases that need to be handled in code within the constructor. It's also ultimately the limit of what can be provided, so a user *could* provide both a `menuGenerator: Sentinel` and a validator.

The first two constructors here just guide what combinations should be provided. Without providing any parameters, it'll suggest providing a `MenuGenerator`. If a `Sentinel` is provided, it'll suggest not providing a validator or config. Since Sentinel is only for subclasses, it seems like most users won't want to provide it.

core/field_dropdown.ts
```diff
@@ -744,13 +669,89 @@ const IMAGE_Y_PADDING: number = IMAGE_Y_OFFSET * 2;
...
+
+  const shortest = utilsString.shortestStringLength(stringLabels);
+  const prefixLength = utilsString.commonWordPrefix(stringLabels, shortest);
+  const suffixLength = utilsString.commonWordSuffix(stringLabels, shortest);
+
+  if ((!prefixLength && !suffixLength) ||
+      (
+          // One or more strings will entirely vanish if we proceed.  Abort.
+          shortest <= prefixLength + suffixLength))
+    return {options: stringOptions};
```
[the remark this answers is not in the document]
btw17: Will do! I think originally I had the return on line 714, though the formatter probably bumped it to a new line.

core/field_dropdown.ts
```diff
@@ -433,18 +382,14 @@ export class FieldDropdown extends Field {
    *     (human-readable text or image, language-neutral name).
```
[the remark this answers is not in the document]
btw17: Before, an empty array or a generator that produces an empty array could also be provided. Good point, though - if we're assuming the array should be non-empty then we should throw an error if it isn't.

core/field_dropdown.ts
```diff
@@ -114,30 +113,29 @@ export class FieldDropdown extends Field {
...
     super(Field.SKIP_SETUP);
 
-    // If we pass SKIP_SETUP, don't do *anything* with the menu generator.
-    if (menuGenerator === Field.SKIP_SETUP) {
+    if (isMenuGenerator(menuGenerator)) {
+      if (Array.isArray(menuGenerator)) {
+        validateOptions(menuGenerator);
+        const trimmed = trimOptions(menuGenerator);
+        this.menuGenerator_ = trimmed.options;
+        this.prefixField = trimmed.prefix || null;
```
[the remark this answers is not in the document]
btw17: Originally, I tried keeping the resulting values identical to the previous implementation in mind of a minor release. Since this will probably be included in a major release, though, I think it might be worth making `prefixField` and `suffixField` accept undefined instead of null. `null` is in a weird spot where it's mostly symbolic now, perhaps conveying an explicit choice to provide a non-value rather than choosing to not provide a value.

I'll make the update in favor of undefined since it keeps the code more concise, though I can adjust either way.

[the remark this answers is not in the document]
btw17: Actually, updating `prefixField` and `suffixField` would require updating the `Field` parent class and impacted files so that'd be a pretty expansive change. Would it be fine to leave this as is for now? Ultimately I think those values probably should be just optional strings rather than string|null, though that's probably for a future PR.

core/field_dropdown.ts
```diff
@@ -725,6 +646,10 @@ export interface ImageProperties {
  */
 export type MenuOption = [string | ImageProperties, string];
 
+export type MenuGeneratorFunction = (this: FieldDropdown) => MenuOption[];
```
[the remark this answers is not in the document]
btw17: Done!

core/field_dropdown.ts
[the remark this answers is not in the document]
btw17: Fixed!

core/field_dropdown.ts
btw17: Done!

core/field_dropdown.ts
```diff
@@ -44,7 +44,7 @@ export class FieldDropdown extends Field {
    * height.
    */
   static MAX_MENU_HEIGHT_VH = 0.45;
-  static ARROW_CHAR: AnyDuringMigration;
+  static ARROW_CHAR: '▼'|'▾';
```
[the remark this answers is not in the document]
btw17: Good point - updated!

btw17:
I made most of the requested updates and cleaned up the branch to no longer include those commits from `master`. The only thing left outstanding in the discussion of `null` vs `undefined` and if there was any further thoughts about overloading the constructor.

core/field_dropdown.ts
```diff
@@ -433,18 +368,21 @@ export class FieldDropdown extends Field {
...
+    const options = (() => {
+      if (!this.menuGenerator_) return [];
+      if (Array.isArray(this.menuGenerator_)) return this.menuGenerator_;
+      if (opt_useCache && this.generatedOptions_) return this.generatedOptions_;
+
+      this.generatedOptions_ = this.menuGenerator_();
+      validateOptions(this.generatedOptions_);
       return this.generatedOptions_;
+    })();
+
```
btw17: Oh! I just noticed that `validateOptions` checks for a non-empty array. I was incorrect before in my assumption that the generator function and array could be empty. Updated again!

core/field_dropdown.ts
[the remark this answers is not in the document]
btw17: I just noticed that `validateOptions` checks for a non-empty array. I was incorrect before in my assumption that the generator function and array could be empty. Updated again!

core/field_dropdown.ts
```diff
@@ -44,7 +44,9 @@ export class FieldDropdown extends Field {
    * height.
    */
   static MAX_MENU_HEIGHT_VH = 0.45;
-  static ARROW_CHAR: AnyDuringMigration;
+
+  /** Android can't (in 2014) display "▾", so use "▼" instead. */
+  static ARROW_CHAR: string = userAgent.ANDROID ? '▼' : '▾';
```
BeksOmega, line 49: Pretty sure this isn't actually a problem anymore (I can view this file just fine on my android phone). Could you have this just use the small carrot string + remove the comment?

core/field_dropdown.ts
```diff
@@ -114,30 +113,27 @@ export class FieldDropdown extends Field {
    * @throws {TypeError} If `menuGenerator` options are incorrectly structured.
    */
   constructor(
-      menuGenerator: AnyDuringMigration[][]|Function|Sentinel,
-      opt_validator?: Function, opt_config?: FieldConfig) {
+      menuGenerator: MenuGenerator, opt_validator?: Function,
```
BeksOmega, line 116: nit: can you put each param on a separate line (here and below)? Related to: https://google.github.io/styleguide/jsguide.html#formatting-where-to-break and https://github.com/google/google-java-format/wiki/The-Rectangle-Rule

core/field_dropdown.ts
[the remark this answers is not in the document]
BeksOmega: Could you still move the comment so that which part is the predicate is a bit easier to parse? I think either put it right above the if-statement, or right about the return statement.

core/field_dropdown.ts
```diff
@@ -741,16 +652,90 @@ const IMAGE_Y_OFFSET = 5;
...
+                         Sentinel): menuGenerator is MenuGenerator {
+  return menuGenerator !== Field.SKIP_SETUP;
+}
+
+/**
+ * Factor out common words in statically defined options.
+ * Create prefix and/or suffix labels.
+ */
+function trimOptions(options: MenuOption[]):
+    {options: MenuOption[]; prefix?: string; suffix?: string;} {
```
BeksOmega, line 669: I reeeeally like this change :D

core/field_dropdown.ts
```diff
@@ -741,16 +652,90 @@ const IMAGE_Y_OFFSET = 5;
...
+
+/**
+ * Use the calculated prefix and suffix lengths to trim all of the options in
+ * the given array.
+ *
+ * @param options Array of option tuples:
+ *     (human-readable text or image, language-neutral name).
+ * @param prefixLength The length of the common prefix.
+ * @param suffixLength The length of the common suffix
+ * @returns A new array with all of the option text trimmed.
```
BeksOmega, line 720: Yay pure programming instead of mutation!

core/field_dropdown.ts
```diff
@@ -453,17 +388,11 @@ export class FieldDropdown extends Field {
...
+      |null {
     const options = this.getOptions(true);
-    for (let i = 0, option; option = options[i]; i++) {
-      // Options are tuples of human-readable text and language-neutral values.
-      if (option[1] === opt_newValue) {
-        isValueValid = true;
-        break;
-      }
-    }
+    const isValueValid = options.some((option) => option[1] === opt_newValue);
```
BeksOmega, line 394: Yes, love this change!

core/field_variable.ts
```diff
@@ -89,9 +88,7 @@ export class FieldVariable extends FieldDropdown {
      * An array of options for a dropdown list,
      * or a function which generates these options.
      */
-    // AnyDuringMigration because:  Type '(this: FieldVariable) => any[][]' is
-    // not assignable to type 'any[][] | ((this: FieldDropdown) => any[][])'.
-    this.menuGenerator_ = FieldVariable.dropdownCreate as AnyDuringMigration;
+    this.menuGenerator_ = FieldVariable.dropdownCreate as MenuGenerator;
```
BeksOmega, line 91: Mmm this is kind of unfortunate (having to cast with `as`), but I think this is the right thing to do here, b/c sadly expecting a `FieldVariable` means it isn't actually a subtype :/

core/field_dropdown.ts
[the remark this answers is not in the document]
BeksOmega: I think this combination of constructors makes sense!

BeksOmega (approved):
LGTM! I wrote down a few nits, but they're all optional in my mind, so I'm going ahead and approving this =) Will merge Tuesday after making sure Maribeth is cool with it.

core/field_dropdown.ts
btw17: Done!

core/field_dropdown.ts
btw17: Done!

core/field_dropdown.ts
btw17: I moved the comment to right above the return statement.

Glad you like the change! ⭐

core/field_dropdown.ts
btw17: Condensing code's incredibly satisfying. Glad you like the change! ⭐

core/field_variable.ts
btw17: Yeah, that was my thought as well. I explored a few different options, like updating `dropdownCreate` to accept generics, though in the end this made the most sense.

Ideally this could be refactored such that `this` isn't a parameter for the base function, though my goal was to avoid impacting the usage of each function.

btw17:
Pulled in this week's `develop` changes and rebased - should be good to go now!

maribethb (approved):
Thanks for making the changes and thanks to Beka for the additional review!

PR #6559, opened 2022-10-18
chore: fix mutator blocks jumping

### LINKED ISSUE #6541: Mutator stacks are broken in current master
On the playground:
https://blockly-demo.appspot.com/static/tests/playground.html

Open any block with a mutator featuring a stack of blocks (list, or if, or procedure).  Drag the blocks around and watch them fly apart.

Video:
https://youtu.be/OsaDGN7Z2_Y

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6541 

### Proposed Changes

Reverts changes made in #6301 that caused jumping.

### Reason for Changes

Bugs are bad.

### Test Coverage

Manually tested the repro steps from #6541 

### Documentation

N/A

### Additional Information

Thank you to Maribeth for the bisect + cherrypick idea!

Also for now I've just reverted the bug-causing changes. But I'm also going to file a bug for actually refactoring this system to be more testable + less stateful + not use private properties on svg elements.

### DIFF
--- a/core/block_animations.ts
+++ b/core/block_animations.ts
@@ -188,7 +188,11 @@ function disconnectUiStep(group: SVGElement, magnitude: number, start: Date) {
     skew = `skewX(${val})`;
     disconnectPid = setTimeout(disconnectUiStep, 10, group, magnitude, start);
   }
-  group.setAttribute('transform', skew);
+  (group as AnyDuringMigration).skew_ = skew;
+  group.setAttribute(
+    'transform',
+    (group as AnyDuringMigration).translate_ +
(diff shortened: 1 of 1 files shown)


maribethb (approved):
Spent a long time trying to understand the original change + this.

It looks like before the original change the `transform` attribute was set to skew + translate, and after that change it's only set to skew. I wonder if that's the whole problem?

Or part of the problem is block_svg is still using the `skew_` and `translate_` properties, so if we just don't set them here, things go weird? 

Either way this change LGTM and also filing a bug to fix this up is good.

PR #6562, opened 2022-10-18
chore: update procedure map tests to match the refactored API

### LINKED ISSUE #6561: Update procedure map tests to match the refactored API
### Description

The API designed in the [design doc](https://docs.google.com/document/d/1n4MxZQu49mniC7aUtZ1X4xvwWgpL3st4aDU25pwdaNc/edit?usp=sharing) was refactored in response to feedback on #6488 and #6489. The tests written before these (in  #6486) need to be updated to reflect the new design.

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #6561 

### Proposed Changes

🎵 Updates the tests 🎵

### Reason for Changes

Tests should test.

### Test Coverage

This is tests silly!

### Documentation

N/A

### Additional Information

Dependent on #6560

### DIFF
--- a/tests/mocha/procedure_map_test.js
+++ b/tests/mocha/procedure_map_test.js
@@ -73,7 +73,7 @@ suite('Procedure Map', function() {
             new Blockly.procedures.ObservableProcedureModel(this.workspace);
         this.procedureMap.addProcedure(procedureModel);
 
-        procedureModel.setReturnType(['return type 1', 'return type 2']);
+        procedureModel.setReturnTypes(['return type 1', 'return type 2']);
 
         chai.assert.isTrue(
             this.updateSpy.calledOnce, 'Expected an update to be triggered');
@@ -84,9 +84,9 @@ suite('Procedure Map', function() {
             new Blockly.procedures.ObservableProcedureModel(this.workspace);
         this.procedureMap.addProcedure(procedureModel);
 
-        procedureModel.setReturnType(['return type']);
+        procedureModel.setReturnTypes(['return type']);
         this.updateSpy.reset();
-        procedureModel.setReturnType(null);
+        procedureModel.setReturnTypes(null);
 
         chai.assert.isTrue(
             this.updateSpy.calledOnce, 'Expected an update to be triggered');
@@ -144,17 +144,16 @@ suite('Procedure Map', function() {
     });
 
     suite('parameter model updates', function() {
-      test('setting the variable model triggers an update', function() {
+      test('setting the name triggers an update', function() {
         const procedureModel =
             new Blockly.procedures.ObservableProcedureModel(this.workspace);
         this.procedureMap.addProcedure(procedureModel);
-        const parameterModel =
-            new Blockly.procedures.ObservableParameterModel(this.workspace);
+        const parameterModel = new Blockly.procedures.ObservableParameterModel(
+            this.workspace, 'test1');
         procedureModel.insertParameter(parameterModel);
         this.updateSpy.reset();
 
-        parameterModel.setVariable(
-            new Blockly.VariableModel(this.workspace, 'variable'));
+        parameterModel.setName('test2');
 
         chai.assert.isTrue(
             this.updateSpy.calledOnce, 'Expected an update to be triggered');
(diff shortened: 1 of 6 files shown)


tests/mocha/procedure_map_test.js
```diff
@@ -36,20 +36,28 @@ suite.skip('Procedure Map', function() {
...
 
       test('deleting a procedure triggers an update', function() {
-        const procedureModel =
-            new Blockly.procedures.ObservableProcedureModel(this.workspace);
-        this.procedureMap.addProcedure(procedureModel);
+        this.procedureMap.add(
+            new Blockly.procedures.ObservableProcedureModel(this.workspace));
 
-        this.procedureMap.deleteProcedure(procedureModel.getId());
+        this.procedureMap.delete([...this.procedureMap.values()][0].getId());
```
maribethb, line 60: why this change? the old way is more readable imo (here and below)

(i know the method name changed, but storing the procedureModel in a variable)

tests/mocha/procedure_map_test.js
```diff
@@ -8,18 +8,18 @@ import {sharedTestSetup, sharedTestTeardown} from './test_helpers/setup_teardown
...
-    this.procedureMap = this.workspace.getProcedureMap();
+    // this.procedureMap = this.workspace.getProcedureMap();
   });
 
   teardown(function() {
     sharedTestTeardown.call(this);
   });
 
-  suite('triggering block updates', function() {
+  suite.skip('triggering block updates', function() {
```
maribethb, line 22: just checking this is supposed to be skipped now? can you add a TODO comment saying when to unskip it?

maribethb (approved):

tests/mocha/procedure_map_test.js
BeksOmega: I was moving the tests to actually use the fluent API hehe. I'll move things around so that we're storing the relevant bit in a variable though.

(side question: Do you think it'd be more readable as `this.procedureMap.getProcedures()[0].getId()`? )

tests/mocha/procedure_map_test.js
BeksOmega: Yesh! The skip just got moved from the top-level down to here. Can add a comment o7

tests/mocha/procedure_map_test.js
```diff
@@ -184,4 +191,67 @@ suite.skip('Procedure Map', function() {
...
+
+  suite('backing variable to parameters', function() {
+    test(
+        'construction references an existing variable if available',
+        function() {
+          const variable = this.workspace.createVariable('test1');
+          const param = new Blockly.procedures.ObservableParameterModel(
+              this.workspace, 'test1');
+
+          console.log(param);
```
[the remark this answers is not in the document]
BeksOmega: Thanks for the catch!

tests/mocha/procedure_map_test.js
maribethb: yeah, that would be more readable, but i think storing it in a variable is best.

you can have a separate test that tests the getProcedures function and one that tests the chain of the fluent api.

the justification is that if one of those methods break, you ideally want only the tests related to it to break, and not every single one of the tests, because that would slow down finding out the actual cause of the breakage. 

I think these are relevant snippets from go/readable-tests:
- "Create local variables when nontrivial values are referenced multiple times"
- "Write tests as straight-line code that doesn't force the reader to do mental computations to verify the test's correctness"
- "Write a separate test for each behavior defined by a method"
- "Write tests that call multiple methods on a class only when that sequence of calls defines a well-specified behavior"

BeksOmega: Do you think it's useful to test the chain of the fluent API? Imo I think that's well specified by the interfaces, and additional tests would just end up being change detectors. But idk, I'm still learning how to write tests well. (also acknowledging that you could have just been using that as an example haha)

I just wanted to change these so that they matched how we want people to actually use the API, not necessarily /test/ the API if that makes sense.

And wrt your snippets, were there any other changes you wanted me to make?

BeksOmega: > but in terms of how people would use it, i think it's important that you stick to public APIs, but not important that everything you do in the test would necessarily make sense to do in production, if that makes sense.

Sweet, totally makes sense!

PR #6571, opened 2022-10-19
fix: Fix the compiler test.

NeilFraser:
The TypeScript compiler generates some code relating to enums that's not compatible with the Closure Compiler's advanced optimizations.  This PR adds a lightweight converter that fixes these enums and restores the functionality of this test.

Where the converter will live and what form it should take is a matter of future discussion.  But for now it's just in the tests/compile directory.

### DIFF
--- a/tests/compile/index.html
+++ b/tests/compile/index.html
@@ -3,8 +3,8 @@
 <head>
   <meta charset="utf-8">
   <title>Blockly: Advanced Compilation Test</title>
-  <script src="main_compressed.js"></script>
   <script src="../../msg/js/en.js"></script>
+  <script src="main_compressed.js"></script>
   <style>
     body {
       background-color: #fff;
@@ -22,7 +22,7 @@
 <body>
   <h1>Blockly: Advanced Compilation Test</h1>
 
-  <p>To run this test manually, run `npm run test:compile:advanced`
+  <p>To run this test manually, run `node tsick.js` and `npm run test:compile:advanced`
   from the command line, then open this file in your web browser.</p>
 
   <p>Measure the size of main_compressed.js (295kb as of October 2017), then reload
--- a/tests/compile/main.js
+++ b/tests/compile/main.js
@@ -16,6 +16,9 @@ const {inject} = goog.require('Blockly.inject');
 goog.require('Blockly.geras.Renderer');
 /** @suppress {extraRequire} */
 goog.require('Blockly.VerticalFlyout');
+/** @suppress {extraRequire} */
+const {Msg} = goog.require('Blockly.Msg');
+
 // Blocks
 /** @suppress {extraRequire} */
 goog.require('Blockly.libraryBlocks.logic');
@@ -30,6 +33,7 @@ goog.require('testBlocks');
 
 
 function init() {
+  Object.assign(Msg, window['Blockly']['Msg']);
   inject('blocklyDiv', /** @type {BlocklyOptions} */ ({
            'toolbox': document.getElementById('toolbox')
          }));
--- a/tests/compile/tsick.js
+++ b/tests/compile/tsick.js
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Lightweight conversion from tsc to Closure Compiler.
+ */
+'use strict';
+
+const fs = require('fs');
+
+const DIR = '../../build/src/core/';
+
+function spider(dir) {
+  const data = fs.readdirSync(dir, {withFileTypes: true});
+  for (const datum of data) {
+    if (datum.isDirectory()) {
+      spider(dir + datum.name + '/')
+    } else if (datum.name.endsWith('.js')) {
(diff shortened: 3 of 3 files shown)


tests/compile/index.html
```diff
@@ -22,7 +22,7 @@
 <body>
   <h1>Blockly: Advanced Compilation Test</h1>
 
-  <p>To run this test manually, run `npm run test:compile:advanced`
+  <p>To run this test manually, run `node tsick.js` and `npm run test:compile:advanced`
```
rachel-fenichel, line 25: Can you make the `test:compile:advanced` command also run `tsick.js`?

NeilFraser: I didn't integrate it here because I think it would make more sense to have the `build:js` command run `tsick.js`.  There is nothing disruptive in its changes.  Doing so would allow `build:compressed` to switch to advanced optimizations once we have an externs list.

That's why I'm proposing loosely storing tsick here for now, without integrating it into the flow.

rachel-fenichel: I'd rather integrate it directly now, so that running the test does the full workflow, and then possibly pulling it up into the `build:js` command down the line. It's relatively straightforward, and makes it harder to run tests in invalid ways.

tests/compile/main.js
```diff
@@ -30,6 +33,7 @@ goog.require('testBlocks');
 
 
 function init() {
+  Object.assign(Msg, window['Blockly']['Msg']);
```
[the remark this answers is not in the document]
cpcallen: I note that `tests/compile/index.js` loads `msg/js/en.js`, which is the pre-wrapped version that does not yet have the `Object.assign` built in.

I presume the reason it wasn't working as it did previously (loading blockly as part of `main_compressed.js` and then loading `en.js`) is because `main_compressed.js` does not define a `Blockly.Msg` property any more due to advanced compilation.

Previously here was a hack to work around this in `core/blockly.ts`, but a while ago I moved it to `core/main.js`.  It _looks_ like `core/main.js` is still getting presented to Closure Compiler by `buildAdvancedCompilationTest`.  @NeilFraser: can you verify whether the code contained therein actually being included in the output?

* If it is then I'm not sure why it's not having the desired effect.  Can you see if `globalThis.Blockly` is unexpectedly defined by that point?

* If not, does putting something like:
  ```JS
  if (!('Blockly' in globalThis)) {
    globalThis['Blockly'] = {};
  }
  if (!('Msg' in globalThis['Blockly'])) {
    globalThis['Blockly']['Msg'] = 'Msg': Blockly.Msg;
  }
  ```
  in `tests/compile/main.js` resolve the loading issue without needing to resort to the `Object.assign` kludge?  Not that this one is much nicer, but at least it is working around a known issue with advanced compilation.

tests/compile/tsick.js
```diff
@@ -0,0 +1,58 @@
...
+ * @fileoverview Lightweight conversion from tsc to Closure Compiler.
+ */
+'use strict';
+
+const fs = require('fs');
+
+const DIR = '../../build/src/core/';
+
+function spider(dir) {
+  const data = fs.readdirSync(dir, {withFileTypes: true});
```
cpcallen, line 17: I think `entries` or `dirEntries` (and later `entry` or `dirEntry`) would be a little more specific.

tests/compile/tsick.js
```diff
@@ -0,0 +1,58 @@
...
+    // m[0] looks like a bunch of lines in one of these two formats:
+    //   ScopeType["BLOCK"] = "block";
+    //   KeyCodes[KeyCodes["TAB"] = 9] = "TAB";
+    // We need to unquote them to look like one of these two formats:
+    //   ScopeType.BLOCK = "block";
+    //   KeyCodes[KeyCodes.TAB = 9] = "TAB";
+    const oldSnippet = m[0];
+    let newSnippet = oldSnippet.replace(/\["(\w+)"\]/g, '.$1');
+    // Add a comment so we don't keep trying to modify this enum.
+    newSnippet = newSnippet.replace(') {', ') {  // Converted by tsick.');
```
cpcallen, line 52: Judicious use of the [the stick`y` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) should obviate the need for this (but do make sure you're not creating a fresh RegExp object, and thereby resetting `.lastIndex`, for each iteration!)

tests/compile/index.html
cpcallen: The difficulty is that I will soon introduce a change that causes `test:compile:advanced` to run `build:js` automatically, preventing `tsick.js` from being run manually in between.

However, I think it better to commit this PR in (approximately) its current form and then I will integrate it into the pipeline as needed.

tests/compile/tsick.js
NeilFraser: The `y` flag is giving me mixed messages.  Here it works as expected:
```
const ry = /m/y;
'mmm'.match(ry);
'mmm'.match(ry);
```
But here it does not, returning null on the second call:
```
const ry = /m/y;
'm_m_m'.match(ry);
'm_m_m'.match(ry);
```

NeilFraser:
Closing, as this PR is replaced by #6638.

PR #6575, opened 2022-10-20
fix: refactors concrete implementations of the procedure data models

BeksOmega:
## The basics

- [] I branched from develop

### Resolves

No tracking issue for this, just what we discussed in internal chat / vc.

### Reason for Changes

Three separate things:
  * Expanding the IParameterModel interface
  * Removing support for return types from the ObervableProcedureModel
  * Adding an interface of the ProcedureMap (which includes our non-`Map` methods)

### Test Coverage

Relevant tests already existed.

### Documentation

N/A

### Additional Information

Dependent on #6570

### DIFF
--- a/core/procedures/observable_parameter_model.ts
+++ b/core/procedures/observable_parameter_model.ts
@@ -49,6 +49,20 @@ export class ObservableParameterModel implements IParameterModel {
   }
 
   /**
+   * Returns the name of this parameter.
+   */
+  getName(): string {
+    return this.variable.name;
+  }
+
+  /**
+   * Returns the types of this parameter.
+   */
+  getTypes(): string[] {
+    return [];
+  }
+
+  /**
    * Returns the unique language-neutral ID for the parameter.
    *
    * This represents the identify of the variable model which does not change
--- a/core/procedures/observable_procedure_model.ts
+++ b/core/procedures/observable_procedure_model.ts
@@ -56,8 +56,13 @@ export class ObservableProcedureModel implements IProcedureModel {
    * Pass null to represent a procedure that does not return.
    */
   setReturnTypes(types: string[]|null): this {
-    // TODO(#6516): Fire events.
+    if (types && types.length) {
+      throw new Error(
+        'The built-in ProcedureModel does not support typing. You need to ' +
+        'implement your own custom ProcedureModel.');
+    }
     this.returnTypes = types;
+    // TODO(#65): Fire events.
     triggerProceduresUpdate(this.workspace);
     return this;
   }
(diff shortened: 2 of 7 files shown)


BeksOmega:
Ok @maribethb I think this is ready for a look!

core/procedures/observable_procedure_model.ts
```diff
@@ -56,8 +58,13 @@ export class ObservableProcedureModel implements IProcedureModel {
...
    */
   setReturnTypes(types: string[]|null): this {
-    // TODO(#6516): Fire events.
+    if (types && types.length) {
+      throw new Error(
+          'The built-in ProcedureModel does not support typing. You need to ' +
+          'implement your own custom ProcedureModel.');
+    }
     this.returnTypes = types;
+    // TODO(#65): Fire events.
```
maribethb, line 67: nit: fix the todo number

core/procedures/observable_procedure_model.ts
```diff
@@ -56,8 +58,13 @@ export class ObservableProcedureModel implements IProcedureModel {
    * Pass null to represent a procedure that does not return.
```
maribethb, line 58: Can you add the information about how this isn't supported to the TsDoc for this function?  From the lens of someone viewing the reference doc, they should be able to tell that they shouldn't call this method (on this class) and need to implement their own.

core/procedures/observable_parameter_model.ts
```diff
@@ -48,6 +48,20 @@ export class ObservableParameterModel implements IParameterModel {
...
+   */
+  getName(): string {
+    return this.variable.name;
+  }
+
+  /**
+   * Returns the types of this parameter.
+   */
+  getTypes(): string[] {
+    return [];
```
maribethb, line 62: We're not supporting types in parameters either right? if so then I think it's mildly confusing that in the procedure it throws but here it returns empty. I think we should be consistent unless there is an outside reason this has to return empty.

also same comment applies about documenting the lack of support in this implementation

maribethb (approved):

core/procedures/observable_parameter_model.ts
BeksOmega: In both classes the setter throws but the getter returns a value, so I think it is consistent in that way. But I feel you about how it's weird.

I built it this way so that the procedure serializer could just work against the interfaces, rather than the concrete implementations. That way if people have to implement their own procedure data models they don't have to also implement their own serializer. But the utility of that is questionable.

I /prefer/ allowing the built-in serializer to work against the abstract types rather than the concrete ones because it's more future proof. But I think I'm fine with going the other way on this as well.

maribethb: ah no that makes sense, i was just looking at the setter for one and getter for another. not forcing people to write their own serializer seems useful. cool beans!

BeksOmega: ❄️🫘

Issue #6548, opened 2022-10-14
Remove underscores from private property and method names


rachel-fenichel:
**Is your feature request related to a problem? Please describe.**

Why: to match the current google style guide and as part of checking tests for accesses to private properties.

**Describe the solution you'd like**

Rename only things that are private or otherwise internal to a file.
Anything with an underscore that is public or protected should be left alone to avoid API churn.

**Describe alternatives you've considered**

Could also rename protected stuff, but that has to be handled on a case-by-case basis.

Alternatively, we could leave these alone because they aren't part of the external API anyway.

PR #6583, opened 2022-10-26
chore: remove some underscores in icon-related classes

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/6548

### Proposed Changes

Remove underscores from private property names and method names, where they weren't being used by tests.
Also cleans up some nullability indicators.

#### Behavior Before Change

No change.

#### Behavior After Change

No change.

### Reason for Changes

General cleanup, as a way of reading through code and understanding icon architecture.

### Test Coverage

Mocha tests, and also opened up the playground and played around with comments and mutators.

### Documentation

None.

### Additional Information

`comment.textarea_` and `comment.paragraphElement_` are being accessed directly in tests, so I left them alone.

### DIFF
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -213,14 +213,17 @@ export class Comment extends Icon {
    * the size of the bubble).
    */
   private resizeTextarea() {
+    if (!this.textarea_ || !this.foreignObject) {
+      return;
+    }
     const size = this.model.size;
     const doubleBorderWidth = 2 * Bubble.BORDER_WIDTH;
     const widthMinusBorder = size.width - doubleBorderWidth;
     const heightMinusBorder = size.height - doubleBorderWidth;
-    this.foreignObject!.setAttribute('width', `${widthMinusBorder}`);
-    this.foreignObject!.setAttribute('height', `${heightMinusBorder}`);
-    this.textarea_!.style.width = widthMinusBorder - 4 + 'px';
-    this.textarea_!.style.height = heightMinusBorder - 4 + 'px';
+    this.foreignObject.setAttribute('width', `${widthMinusBorder}`);
+    this.foreignObject.setAttribute('height', `${heightMinusBorder}`);
+    this.textarea_.style.width = widthMinusBorder - 4 + 'px';
+    this.textarea_.style.height = heightMinusBorder - 4 + 'px';
   }
 
   /**
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -85,7 +85,7 @@ export class Comment extends Icon {
    *
    * @param group The icon group.
    */
-  protected override drawIcon_(group: SVGGElement) {
+  protected override drawIcon_(group: Element) {
     // Circle.
     dom.createSvgElement(
         Svg.CIRCLE,
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -36,30 +36,30 @@ import {Svg} from './utils/svg.js';
  * @alias Blockly.Comment
  */
 export class Comment extends Icon {
-  private readonly model_: CommentModel;
+  private readonly model: CommentModel;
 
   /**
    * The model's text value at the start of an edit.
    * Used to tell if an event should be fired at the end of an edit.
    */
-  private cachedText_: string|null = '';
+  private cachedText: string|null = '';
 
   /** Mouse up event data. */
-  private onMouseUpWrapper_: browserEvents.Data|null = null;
+  private onMouseUpWrapper: browserEvents.Data|null = null;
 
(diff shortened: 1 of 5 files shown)


core/mutator.ts
```diff
@@ -236,65 +236,63 @@ export class Mutator extends Icon {
...
-      } else {
+      if (!this.getBlock().isEditable) {
         // Close any mutator bubble.  Icon is not clickable.
         this.setVisible(false);
-        if (this.iconGroup_) {
-          dom.addClass(this.iconGroup_, 'blocklyIconGroupReadonly');
-        }
+      }
+      if (this.iconGroup_) {
+        dom.removeClass(this.iconGroup_, 'blocklyIconGroupReadonly');
```
maribethb, line 244: seems like we are now always removing the class, but before we were adding it or removing it depending on the new editable state. I don't see where we're ever adding the class, is this change intended?

core/mutator.ts
```diff
@@ -236,65 +236,63 @@ export class Mutator extends Icon {
...
       }
     }
   }
 
   /** Resize the bubble to match the size of the workspace. */
-  private resizeBubble_() {
+  private resizeBubble() {
+    // If the bubble exists, the workspace also exists.
+    const ws = this.workspace as WorkspaceSvg;
```
maribethb, line 252: I like using the `!` instead of casting, the intention is more clear (with the cast, i scrolled back up to see if this was allowed to be a plain `Workspace` to understand why we needed the cast). does it work to `const ws = this.workspace!;`?

Alternatively I think `if (!this.workspace) return;` is even better since it's type safer and about the same amount of code. you could do the same below in `setVisible` to eliminate more non-null assertions

core/mutator.ts
```diff
@@ -314,70 +312,70 @@ export class Mutator extends Icon {
...
       if (tree) {
-        flyout!.init(this.workspace_!);
+        flyout!.init(this.workspace!);
         flyout!.show(tree);
       }
 
-      this.rootBlock_ = block.decompose!(this.workspace_!)!;
-      const blocks = this.rootBlock_!.getDescendants(false);
+      this.rootBlock = block.decompose!(this.workspace!)!;
+      const blocks = this.rootBlock!.getDescendants(false);
```
maribethb, line 328: you can eliminate the `!` here since you've already asserted above

core/mutator.ts
```diff
@@ -456,7 +454,7 @@ export class Mutator extends Icon {
       block.rendered = false;
 
       // Allow the source block to rebuild itself.
-      block.compose!(this.rootBlock_);
+      block.compose!(this.rootBlock);
```
maribethb, line 457: random aside: pretty sure we could have `block` say it is a type that describes a "Block where we know the compose property is present" since we know it actually must be present if we are calling this, using union types. that would be fancy and i'm only mentioning it here because i'm excited by how specific we can get with TS, not because i think we should actually change anything rn

core/warning.ts
```diff
@@ -143,8 +143,8 @@ export class Warning extends Icon {
    */
   getText(): string {
     const allWarnings = [];
-    for (const id in this.text_) {
-      allWarnings.push(this.text_[id]);
+    for (const id in this.text) {
```
maribethb, line 146: is this for loop different than `this.text.values()`? we probably just didn't use it before because of IE? (i know you didn't add this but just wondering)

core/mutator.ts
rachel-fenichel: And this is why we have code reviews! Nope, I misread and thought they were doing the same thing. Good catch.

rachel-fenichel (commented):
Done, and I also updated tests to use helpers for assertions.

core/comment.ts
```diff
@@ -143,82 +143,82 @@ export class Comment extends Icon {
...
-  private resizeTextarea_() {
-    const size = this.model_.size;
+  private resizeTextarea() {
+    const size = this.model.size;
     const doubleBorderWidth = 2 * Bubble.BORDER_WIDTH;
     const widthMinusBorder = size.width - doubleBorderWidth;
     const heightMinusBorder = size.height - doubleBorderWidth;
-    this.foreignObject_!.setAttribute('width', `${widthMinusBorder}`);
-    this.foreignObject_!.setAttribute('height', `${heightMinusBorder}`);
+    this.foreignObject!.setAttribute('width', `${widthMinusBorder}`);
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

maribethb (approved):

PR #6585, opened 2022-10-27
fix: Rename Generator to CodeGenerator

NeilFraser:
## The basics

- [X] I branched from develop

### Resolves

Naming collision between Blockly.Generator and ES6's Generator.

### Proposed Changes

Rename our Generator to CodeGenerator (it's a more descriptive name anyway).

#### Behaviour Before Change

TypeScript would get confused think that our code generators only had .next .return and .throw methods.

#### Behaviour After Change

TypeScript should be less grumpy.

### Reason for Changes

TypeScript is pretending to be Java.

### Test Coverage

I left the tests unchanged after the rename to verify that the old name still works.  Then I updated the tests to the new name.

### Documentation

We currently don't have any documentation about creating a new CodeGenerator.  This is a separate issue.

### Additional Information

In a separate commit on this PR I stripped out all the trailing whitespace from our codebase.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -160,7 +160,6 @@
             "jsdoc/check-param-names": ["off", {"checkDestructured": false}],
             // Allow any text in the license tag. Other checks are not relevant.
             "jsdoc/check-values": ["off"]
-                       
         }
     }]
 }
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -29,9 +29,9 @@ All submissions, including submissions by project members, require review. We
 use Github pull requests for this purpose.
 
 ### Browser compatibility
-We care strongly about making Blockly work on all browsers.  As of 2022 we 
+We care strongly about making Blockly work on all browsers.  As of 2022 we
 support Edge, Chrome, Safari, and Firefox.  We will not accept changes that only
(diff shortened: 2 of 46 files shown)


maribethb:
Drive-by comment: We could enable a [tslint rule](https://typescript-eslint.io/rules/no-shadow/) that yells at us when we're shadowing variables. There is an [option](https://eslint.org/docs/latest/rules/no-shadow#builtinglobals) you have to pass to have it yell about global objects.

Additional drive-by comment: We do have a [codelab](https://blocklycodelabs.dev/codelabs/custom-generator/index.html?index=..%2F..index#2) about generators.

BeksOmega (approved):
This LGTM!

I would also be in favor of enabling no-shadow before this gets merged =) Hopefully we never run into this kind of problem again 🤞 But imo, no harm in enabling it anyway!

NeilFraser:
We aren't actually shadowing anything.  Generator (ES6) is indeed an object, but not one that exists in the global namespace.  One can't execute `new Generator()`.  The word 'Generator' is just a name that the documentation uses.

However, TypeScript has this fantasy that 'Generator' is a global object, and gets all confused when it sees a user-created object of that name.  So we're actually seeing two separate issues with TypeScript here: 1) it should be able to keep track of what Blockly's Generator object is, and 2) it shouldn't even need to keep track of anything since there's no conflict.

PR #6596, opened 2022-11-01
test: add tests for the insertion marker manager

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Just improving test coverage.

### Proposed Changes

Add tests.

### Test Coverage

There are four test suites.

The first one checks that the constructor works when given various blocks and block stacks, and that the appropriate number of markers is created (i.e. one marker for the first block, and sometimes another for the last block in the stack).

The second suite checks logic controlling whether a block is deleted when the drag ends. I stubbed the capability manager and drag target to force it down various paths. `wouldDeleteBlock` is one of the few public methods available to get information out of the insertion marker manager.

The third and fourth suites check that the insertion marker manager can find an appropriate connection in the middle of a drag. I faked dragging the block above, below, left, and right from the stationary block, and then checked the connections on the marker to make sure it was connected in the correct place.

### Documentation

No documentation needed.

### Additional Information

These tests will hopefully catch any bugs in an upcoming cleanup of the insertion marker manager.

Live footage of me working on this PR: 
penguin-angry

### DIFF
--- a/tests/mocha/insertion_marker_manager_test.js
+++ b/tests/mocha/insertion_marker_manager_test.js
@@ -20,14 +20,14 @@ suite('Insertion marker manager', function() {
       sharedTestTeardown.call(this);
     });
 
-    function createBlocksAndManager(workspace, state) {
-        Blockly.serialization.workspaces.load(state, workspace);
-        const block = workspace.getBlockById('first');
-        const manager = new Blockly.InsertionMarkerManager(block);
-        return manager;
-    }
-
     suite('Create markers', function() {
+        function createBlocksAndManager(workspace, state) {
+            Blockly.serialization.workspaces.load(state, workspace);
+            const block = workspace.getBlockById('first');
+            const manager = new Blockly.InsertionMarkerManager(block);
+            return manager;
+        }
+
         test('One stack block', function() {
             const state = {
               'blocks': {
@@ -49,11 +49,11 @@ suite('Insertion marker manager', function() {
                 'blocks': {
                     'blocks': [
                     {
-                      'type': 'text_print',
+                      'type': 'stack_block',
                       'id': 'first',
                       'next': {
                         'block': {
-                          'type': 'text_print',
+                          'type': 'stack_block',
                           'id': 'second',
                         },
                       },
@@ -71,15 +71,15 @@ suite('Insertion marker manager', function() {
                 'blocks': {
                     'blocks': [
                     {
-                      'type': 'text_print',
+                      'type': 'stack_block',
                       'id': 'first',
                       'next': {
                         'block': {
-                          'type': 'text_print',
+                          'type': 'stack_block',
                           'id': 'second',
                           'next': {
                             'block': {
-                              'type': 'text_print',
+                              'type': 'stack_block',
                               'id': 'third',
                             },
                           },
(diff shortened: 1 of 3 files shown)


tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+            const componentManager = workspace.getComponentManager();
+            const stub = sinon.stub(componentManager, 'hasCapability');
+            return stub;
+        }
+
+        test('Over delete area: accepted', function() {
+            const dxy = new Blockly.utils.Coordinate(0, 0);
+            const stub = stubComponentManager(this.workspace);
+            stub.withArgs('fakeDragTarget',
+                Blockly.ComponentManager.Capability.DELETE_AREA).returns(true);
```
BeksOmega, line 165: nit: break at a higher syntactic level if possible. https://google.github.io/styleguide/jsguide.html#formatting-where-to-break

(Fixing the indentation may make this easier)

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+        test('No other blocks nearby', function() {
+            this.manager.update(new Blockly.utils.Coordinate(0, 0), null);
+            chai.assert.isFalse(this.manager.wouldConnectBlock());
+        });
+
+        test('Near other block: above', function() {
+            this.manager.update(new Blockly.utils.Coordinate(200, 190), null);
+            chai.assert.isTrue(this.manager.wouldConnectBlock());
+            const markers = this.manager.getInsertionMarkers();
+            chai.assert.equal(markers.length, 1);
```
BeksOmega, line 245: What are these assertions checking? These look like dupes of the 'Create markers' assertions, can they  be removed?

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+            const marker = markers[0];
+            chai.assert.isTrue(marker.previousConnection.isConnected());
+        });
+
+        test('Near other block: left', function() {
+            this.manager.update(new Blockly.utils.Coordinate(190, 200), null);
+            chai.assert.isTrue(this.manager.wouldConnectBlock());
```
BeksOmega, line 261: I'm confused about why this should be true.

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+            const stub = stubComponentManager(this.workspace);
+            stub.withArgs('fakeDragTarget',
+                Blockly.ComponentManager.Capability.DELETE_AREA).returns(true);
+            const fakeDragTarget = {
+                wouldDelete: sinon.fake.returns(true),
+                id: 'fakeDragTarget',
+            };
+            this.manager.update(dxy, fakeDragTarget);
+            chai.assert.isTrue(this.manager.wouldDeleteBlock());
+            chai.assert.isTrue(fakeDragTarget.wouldDelete.called);
```
BeksOmega, line 172: Is it a promise we've made to external developers that this method will get called? If yes, sweet! If not, I think this just makes the test more of a change-detector, and it should be removed.

BeksOmega (changes requested):
Generally looks really good! Unit tests make me so happy haha.

One high-level thing: How would you feel about renaming some stuff? I really like this convention, because I think it makes it very clear what each suite/test is doing:
  - suites: This tests '___' E.g.
    ```
    suite('Creating markers', function() { })
  - tests: This tests that '___' E.g.
    ```
    test('a single stack block results in a single marker block', function() { });
    ```

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+                          },
+                        },
+                      },
+                  },
+                ],
+              },
+            };
+            const manager = createBlocksAndManager(this.workspace, state);
+            const markers = manager.getInsertionMarkers();
+            chai.assert.equal(markers.length, 1);
```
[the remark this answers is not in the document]
rachel-fenichel: 6ad23110-bfe4-4972-9ce2-d61833d19ba4

If you have multiple row blocks in a row, we don't offer to connect to the end of the row.

I don't know why, though. Some things that come to mind:
- Value inputs/outputs often have type checks, so often it's not allowed anyway
- Distance from mouse pointer gets large fast (when compared to stacks)

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+        
+        function stubComponentManager(workspace) {
+            const componentManager = workspace.getComponentManager();
+            const stub = sinon.stub(componentManager, 'hasCapability');
+            return stub;
+        }
+
+        test('Over delete area: accepted', function() {
+            const dxy = new Blockly.utils.Coordinate(0, 0);
+            const stub = stubComponentManager(this.workspace);
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

tests/mocha/insertion_marker_manager_test.js
rachel-fenichel: Good point. I was using it to confirm to myself that the test was actually running correctly, but it doesn't need to stay

tests/mocha/insertion_marker_manager_test.js
rachel-fenichel: Checking that it's safe to call `markers[0].nextConnection.isConnected`. But I can remove the assertions on `markers.length` and assume that if I'm wrong the test will still fail on the next access. What do you think?

tests/mocha/insertion_marker_manager_test.js
rachel-fenichel: You have to get pretty close, but eventually the blocks will be close enough for a connection to be an option.
5428040e-e0a0-4c37-b469-3538cd8f2ed3


It would not be true if the blocks were taller. Then the distance between the top connection on the moving block and bottom connection on the stationary block would be too great, even if the blocks were directly on top of each other, for them to be compatible. I deliberately did not test exactly where the edge between valid and invalid is, since that would mostly be a test of the internal math.

tests/mocha/insertion_marker_manager_test.js
```diff
@@ -0,0 +1,329 @@
...
+                  },
+                ],
+              },
+            };
+            const manager = createBlocksAndManager(this.workspace, state);
+            const markers = manager.getInsertionMarkers();
+            chai.assert.equal(markers.length, 1);
+        });
+        
+        test('Two stack blocks', function() {
```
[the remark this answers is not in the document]
rachel-fenichel: aed92412-a07b-41e3-bd44-2345cadeeeaf

Two markers! I'll have to add a different test block for that.

And if the output + next block is the only block in the stack, it's one marker (I think).
3528aee7-1b85-4aa5-8ee2-5f90ebdffbde

rachel-fenichel (commented):
Re: naming convention: I'm not a huge fan, because it leads to really verbose names of the tests, especially when they're doing similar things. I'm mostly resigned to using the name of the test as a hint, but needing to read it to really understand what it's testing.

tests/mocha/insertion_marker_manager_test.js
BeksOmega: Sounds good to me!

tests/mocha/insertion_marker_manager_test.js
BeksOmega: So why are some of the assesrtions checking `wouldConnectBlock` but others are testing if the block is actually connected? Is that due to the "not testing the exact edge" thing?

rachel-fenichel: When approaching from the top or bottom, I'm confident that there is a correct answer: when approaching from the top, it should try to attach above. When approaching from the bottom, it should try to attach below.

When approaching from left or right, there's not obviously a correct answer, so I'm not asserting anything about it.

rachel-fenichel:
Okay I updated test names and also added tests for the row to stack block.

BeksOmega (approved):
LGTM

PR #6598, opened 2022-11-02
refactor: Migrate to PointerEvents

### LINKED ISSUE #6543: Switch to pointer events and remove custom event handling logic

Currently Blockly has complex code that tries to unify mouse and touch events. This is all custom code that predates the invention and subsequent adoption of `PointerEvent` by all major browsers. The custom code is complex and a source of bugs e.g. #6508 

Remove the custom code and switch to PointerEvents.

**Additional context**

Assigning to Aaron according to Q4 OKRs, please feel free to reassign if this changes.

gonfunko:
## The basics

- [x] I branched from develop

### Resolves

#6543 

### Proposed Changes
This PR updates Blockly to consistently use PointerEvents internally, rather than trying to normalize different input events by tacking properties onto them. It also cleans up related code and deprecates several obsolete functions.

### Test Coverage
Verified that the test suite passes, and confirmed via manual testing on desktop Safari, Chrome and Firefox, as well as mobile Safari on an iPad, that mouse/touch interactions (dragging, clicking, right clicking, scrolling, zooming) work as expected.


core/touch.ts
```diff
@@ -286,6 +287,7 @@ export function isTouchEvent(e: Event|PseudoEvent): boolean {
  * @alias Blockly.Touch.splitEventByTouches
  */
 export function splitEventByTouches(e: Event): Array<Event|PseudoEvent> {
+  deprecation.warn('splitEventByTouches()', 'version 10', 'version 11');
```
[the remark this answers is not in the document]
gonfunko: Done

core/touch.ts
```diff
@@ -206,18 +175,15 @@ export function getTouchIdentifierFromEvent(e: Event|PseudoEvent): string {
  *     identifier.
  * @alias Blockly.Touch.checkTouchIdentifier
  */
-export function checkTouchIdentifier(e: Event|PseudoEvent): boolean {
+export function checkTouchIdentifier(e: PointerEvent): boolean {
   const identifier = getTouchIdentifierFromEvent(e);
 
-  // if (touchIdentifier_) is insufficient because Android touch
-  // identifiers may be zero.
```
BeksOmega, line 217: No longer true?

core/gesture.ts
```diff
@@ -200,9 +200,9 @@ export class Gesture {
   }
 
   /**
-   * DO MATH to set currentDragDeltaXY_ based on the most recent mouse position.
+   * DO MATH to set currentDragDeltaXY_ based on the most recent pointer position.
```
BeksOmega: DO MATH lol

core/touch_gesture.ts
```diff
@@ -202,10 +197,10 @@ export class TouchGesture extends Gesture {
...
   /**
-   * Handle a touch start or pointer down event and keep track of current
+   * Handle a pointerdown event and keep track of current
    * pointers.
    *
-   * @param e A touch start, or pointer down event.
+   * @param e A pointerdown event.
    * @internal
    */
   handleTouchStart(e: PointerEvent) {
```
BeksOmega: Can `handleTouchStart` be absorbed into the `handleStart` method(s)?

core/comment.ts
```diff
@@ -149,11 +149,8 @@ export class Comment extends Icon {
     body.appendChild(textarea);
     this.foreignObject!.appendChild(body);
 
-    // Ideally this would be hooked to the focus event for the comment.
-    // However doing so in Firefox swallows the cursor for unknown reasons.
-    // So this is hooked to mouseup instead.  No big deal.
```
BeksOmega, line 154: No longer an issue?

core/gesture.ts
```diff
@@ -38,19 +38,19 @@ import type {WorkspaceSvg} from './workspace_svg.js';
 
 
 /**
- * Note: In this file "start" refers to touchstart, mousedown, and pointerstart
- * events.  "End" refers to touchend, mouseup, and pointerend events.
+ * Note: In this file "start" refers to pointerdown
+ * events.  "End" refers to pointerup events.
  */
-// TODO: Consider touchcancel/pointercancel.
+// TODO: Consider pointercancel.
```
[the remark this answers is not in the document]
gonfunko: I'm not clear what there actually is to consider; TOUCH_MAP already listens for pointercancel if you listen for mouseup or touchcancel; I suppose we could also listen for it if you listen for pointerup?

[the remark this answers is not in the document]
BeksOmega: Sweetness, well if you don't understand I don't think there's anyone on earth who understands, so we should just remove this TODO.

core/touch_gesture.ts
BeksOmega: In terms of inlining it.

core/blockly.ts
```diff
@@ -723,7 +722,6 @@ export {Toolbox};
 export {ToolboxCategory};
 export {ToolboxItem};
 export {ToolboxSeparator};
-export {TouchGesture};
```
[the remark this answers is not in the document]
BeksOmega: @maribethb for breaking change decisions.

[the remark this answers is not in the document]
maribethb: still a breaking change, unfortunately. haven't reviewed the rest of the PR, but if it's just a rename then we should be able to add an extra export statement of the old name with the new content, plus a comment saying to remove in version 10. I don't know if we can easily add an actual deprecation warning if we're just exporting under a new name but that's ok. Then please tag this PR as a deprecating change so we can call it out in release notes.

core/touch_gesture.ts
gonfunko: Given that it's called from multiple places I think it's probably preferable to leave as-is?

core/blockly.ts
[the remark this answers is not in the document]
BeksOmega: Reexporting sounds + looks good, I don't think there's anyway that could break people (even instanceof checks should still succeed :P ).

But thinking about this more: is there any reason for us to export `TouchGesture`? Or should we deprecate both of them? idk why an external dev would need to instantiate or reference a gesture.

core/touch_gesture.ts
BeksOmega: Sgtm

core/gesture.ts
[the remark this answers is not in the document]
rachel-fenichel: I'll spend a few minutes trying to figure out what this was

core/blockly.ts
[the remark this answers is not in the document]
gonfunko: It does include the static inProgress(), which I could see being generically useful, but I'm happy to defer to your/the rest of the team's judgement.

core/browser_events.ts
```diff
@@ -95,24 +88,6 @@ export function conditionalBind(
...
     node.addEventListener(name, wrapFunc, false);
     bindData.push([node, name, wrapFunc]);
-
-    // Add equivalent touch event.
-    if (name in Touch.TOUCH_MAP) {
-      const touchWrapFunc = (e: Event) => {
-        wrapFunc(e);
-        // Calling preventDefault stops the browser from scrolling/zooming the
-        // page.
-        const preventDef = !opt_noPreventDefault;
```
rachel-fenichel, line 105: Lint points out you're not using `opt_noPreventDefault`. Why is it no longer needed?

core/browser_events.ts
```diff
@@ -95,24 +88,6 @@ export function conditionalBind(
...
-        if (handled && preventDef) {
```
rachel-fenichel, line 106: `handled` is now unused and can be deleted.

core/field_angle.ts
```diff
@@ -278,9 +278,9 @@ export class FieldAngle extends FieldInput<number> {
     // a click handler on the drag surface to update the value if the surface
     // is clicked.
     this.clickSurfaceWrapper_ = browserEvents.conditionalBind(
-        circle, 'click', this, this.onMouseMove_, true, true);
+        circle, 'pointerdown', this, this.onMouseMove_, true, true);
```
rachel-fenichel, line 281: Does this preserve the behaviour?

rachel-fenichel: Same question for all `click` -> `pointerdown` conversions.

core/flyout_base.ts
```diff
@@ -615,7 +615,7 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
     }
 
     this.listeners_.push(browserEvents.conditionalBind(
-        (this.svgBackground_ as SVGPathElement), 'mouseover', this,
+        (this.svgBackground_ as SVGPathElement), 'pointerover', this,
```
rachel-fenichel, line 618: (no action needed): One day we'll just delete this, because it's a workaround for an IE issue that we no longer care about.

core/touch.ts
rachel-fenichel: Is the value required to be non-zero?

core/blockly.ts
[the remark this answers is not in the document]
rachel-fenichel: Keep it exported because removing it entirely would be a breaking change, not a deprecation.

core/gesture.ts
```diff
@@ -204,9 +239,10 @@ export class Gesture {
   }
 
   /**
-   * DO MATH to set currentDragDeltaXY_ based on the most recent mouse position.
+   * DO MATH to set currentDragDeltaXY_ based on the most recent pointer
```
rachel-fenichel, line 242: DO MATH is entirely because of this ebay posting: https://i.imgur.com/hLHVB.png

core/gesture.ts
```diff
@@ -398,22 +434,24 @@ export class Gesture {
...
-    if (browserEvents.isTargetInput(e)) {
-      this.cancel();
-      return;
-    }
-
+  doStart(e: PointerEvent) {
     if (!this.startWorkspace_) {
       throw new Error(
-          'Cannot start the gesture because the start ' +
+          'Cannot start the touch event becauase the start ' +
```
rachel-fenichel, line 445: touch event -> touch gesture

core/blockly.ts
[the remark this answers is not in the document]
BeksOmega: I definitely agree that the class should be exported under both names for now (for backwards compatibility). But I think long-term it'd be better to not export it at all (since idk of a use-case for it).

wrt `inProgress` I think `workspace.isDragging()` + `WidgetDiv.isVisible()` + etc covers that.

So I think we should remove both exports in v10.

core/gesture.ts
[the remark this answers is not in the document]
rachel-fenichel: According to MDN, pointercancel is supposed to handle cases where "the browser determines that there are unlikely to be any more pointer events, or if after the [pointerdown](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerdown_event) event is fired, the pointer is then used to manipulate the viewport by panning, zooming, or scrolling."

Possible cases include:
"
- A hardware event occurs that cancels the pointer activities. This may include, for example, the user switching applications using an application switcher interface or the "home" button on a mobile device.
- The device's screen orientation is changed while the pointer is active.
- The browser decides that the user started pointer input accidentally. This can happen if, for example, the hardware supports palm rejection to prevent a hand resting on the display while using a stylus from accidentally triggering events.
- The [touch-action](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) CSS property prevents the input from continuing."

It says that "Note: After the pointercancel event is fired, the browser will also send [pointerout](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerout_event) followed by [pointerleave](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerleave_event)."

So if we've already got pointerup, we don't need to worry about pointercancel, and we'll also already handle it directly. So yup, the TODO is obsolete.

core/field_angle.ts
gonfunko: Yes, I manually verified that all cases where this swap was made appear to be working correctly.

rachel-fenichel (approved):

PR #6599, opened 2022-11-03
fix(renamings): Small correction to renaming of `Blockly.utils.global`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

A small error in the entry for `Blockly.utils.global`, that failed to account for the fact that this (former) module had a named export.

### Proposed Changes

Adjust the entry to make it clear that `globalThis` was a named export, not the default export.

Also remove some unnecessary quoting of keywords.

#### Behaviour Before Change

It would have worked anyway.

#### Behaviour After Change

It will work correctly, and now for the right reason.

### Reason for Changes

I am pedantic.

### Additional Information

This change doesn't make any difference in practice to the behaviour of the renamings script, so there is no reason to push it to `master` immediately for the benefit of the renamings script.

### DIFF
--- a/scripts/migration/renamings.json5
+++ b/scripts/migration/renamings.json5
@@ -1339,8 +1339,13 @@
 
   '9.0.0': [
     {
-      'oldName': 'Blockly.utils.global',
-      'newPath': 'globalThis',
+      oldName: 'Blockly.utils.global',
+      exports: {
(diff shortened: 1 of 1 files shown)


maribethb (approved):
Sorry I missed this in my list of PRs.

PR #6602, opened 2022-11-03
chore: clean up insertion marker code

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6548 

### Proposed Changes

There are several rounds of changes in here. I did my best to have separate commits for basic naming changes, whitespace changes, and actual changes. I suggest reviewing one commit at a time.

The first bunch are fairly mechanical.
- Move documentation comments out of the constructor and onto property declarations. This should also improve tsdoc.
- Remove underscores from private property names
- Remove underscores from private properties and methods
- Format/lint

After that things get more complicated. The code previously had many checks for null that happened because a property could be null _at some point_, even though the relevant method was only called once and the relevant properties would never be null. Also, the information about the potential connection to make (the candidate connection) was stored in two separate properties, but hey always changed together. The code assumed that in multiple places before we had strict null checks. After turning on strict null checks, we needed lots of unnecessary checks.

How I resolved this:
- When no candidate is found, return `null` from `getCandidate` instead of a `CandidateConnection` object with `null` properties. As a result, when anything non-null is returned, I know the properties it contains are also non-null (and so does the compiler).
- Store the `CandidateConnection` as a single object, rather than storing the `localConnection` and `closestConnection` as separate properties. This means I don't have to coerce the compiler into believing that if one is non-null, the other is also non-null.
- Pass values into methods when I know they are non-null, instead of accessing them as properties. 

After that, I did assorted cleanup on the new code, including removing some errors that said things like "the code should never get here but the compiler insists that it could."

#### Behavior Before Change

No change.

#### Behavior After Change

No change.

### Reason for Changes

Part of general tidying made possible by knowing which properties are internal and which are external.

### Test Coverage
I wrote tests for this in #6596 and made sure they worked with the previous implementation and also worked after my changes.

I also opened up the advanced playground and manually looked at cases where the block would be inserted vs replaced, because the tests check whether a connection would be made, but not what happens to the previously connected block.

I don't have tests for inserting vs replacing because that's internal to the insertion marker. I could look at the state of the workspace after running `applyConnections`, though, if I trust my test setup enough. :thinking: 

### Documentation

None

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -51,7 +51,17 @@ const DUPLICATE_BLOCK_ERROR = 'The insertion marker ' +
  * @alias Blockly.InsertionMarkerManager
  */
 export class InsertionMarkerManager {
+
+  /**
+   * The top block in the stack being dragged.
+   * Does not change during a drag.
+   */
   private readonly topBlock_: BlockSvg;
+
+  /**
+   * The workspace on which these connections are being dragged.
+   * Does not change during a drag.
+   */
   private readonly workspace_: WorkspaceSvg;
 
   /**
@@ -67,6 +77,11 @@ export class InsertionMarkerManager {
    * Set in initAvailableConnections, if at all
    */
   private lastMarker_: BlockSvg|null = null;
+
+  /**
+   * The insertion marker that shows up between blocks to show where a block
+   * would go if dropped immediately.
+   */
   private firstMarker_: BlockSvg;
 
   /**
@@ -101,36 +116,24 @@ export class InsertionMarkerManager {
 
   /** The block being faded to indicate replacement, or null. */
   private fadedBlock_: BlockSvg|null = null;
+
+  /**
+   * The connections on the dragging blocks that are available to connect to
+   * other blocks.  This includes all open connections on the top block, as
+   * well as the last connection on the block stack. Does not change during a
+   * drag.
+   */
   private availableConnections_: RenderedConnection[];
 
   /** @param block The top block in the stack being dragged. */
   constructor(block: BlockSvg) {
     common.setSelected(block);
-
-    /**
-     * The top block in the stack being dragged.
-     * Does not change during a drag.
-     */
     this.topBlock_ = block;
 
-    /**
-     * The workspace on which these connections are being dragged.
-     * Does not change during a drag.
-     */
     this.workspace_ = block.workspace;
 
-    /**
-     * The insertion marker that shows up between blocks to show where a block
-     * would go if dropped immediately.
-     */
     this.firstMarker_ = this.createMarkerBlock_(this.topBlock_);
 
-    /**
-     * The connections on the dragging blocks that are available to connect to
-     * other blocks.  This includes all open connections on the top block, as
-     * well as the last connection on the block stack. Does not change during a
-     * drag.
-     */
     this.availableConnections_ = this.initAvailableConnections_();
   }
 
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -91,7 +91,7 @@ export class InsertionMarkerManager {
   private closestConnection: RenderedConnection|null = null;
 
   /**
-   * The connection that would connect to this.closestConnection_ if this
+   * The connection that would connect to this.closestConnection if this
    * block were released immediately. Updated on every mouse move. This is on
    * the top block that is being dragged or the last block in the dragging
    * stack.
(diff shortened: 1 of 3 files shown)


core/insertion_marker_manager.ts
```diff
@@ -460,150 +447,135 @@ export class InsertionMarkerManager {
...
+    if (!newCandidate) {
       return;
     }
 
+    const closest = newCandidate.closest;
+
     // Something went wrong and we're trying to connect to an invalid
     // connection.
-    if (closest === this.closestConnection_ ||
+    if (closest === this.activeCandidate?.closest ||
```
rachel-fenichel, line 471: I preserved this because I'm trying to keep the behaviour exactly the same, but I *don't know* what this is checking or why it's important.

core/insertion_marker_manager.ts
```diff
@@ -658,39 +630,39 @@ export class InsertionMarkerManager {
...
-  private hideInsertionMarker_() {
-    if (!this.markerConnection_) {
-      console.log('No insertion marker connection to disconnect');
-      return;
-    }
-
-    const imConn = this.markerConnection_;
-    const imBlock = imConn.getSourceBlock();
+  private hideInsertionMarker(markerConnection: RenderedConnection) {
+    const imBlock = markerConnection.getSourceBlock();
```
cpcallen, line 645: Why not just make this non-throwing, and then call it unconditionally?:
```suggestion
  private hideInsertionMarker() {
    if (!this.markerConnection) return;
    const imBlock = this.markerConnection.getSourceBlock();
```
…and likewise for `hideInsertionInputOutline` and `hideReplacementFade`.

core/insertion_marker_manager.ts
```diff
@@ -51,87 +60,79 @@ const DUPLICATE_BLOCK_ERROR = 'The insertion marker ' +
...
    */
-  private localConnection_: RenderedConnection|null = null;
+  private activeCandidate: CandidateConnection|null = null;
 
   /**
    * Whether the block would be deleted if it were dropped immediately.
    * Updated on every mouse move.
    */
-  private wouldDeleteBlock_ = false;
+  private wouldDeleteBlockInternal = false;
```
cpcallen, line 105: I don't like this property name, especially the `Internal` bit which suggests it is an internal implementation/alternative to the `wouldDeleteBlock` method, but it turns out not to be a method but a boolean.

Does the `wouldDeleteBlock` method, being marked `@internal` serve any useful purpose?  Could it be removed, this property renamed `wouldDeleteBlock`, and calls to the former method replaced with property accesses?  (There might even be a marginal performance benefit in avoiding method calls in what I am guessing is performance-sensitive code.)

If changes to `public @internal` properties are considered to be breaking (because someone might have ignored the `@internal` tag), then consider rename this to `deleteBlockOnDrop` or the like.

core/insertion_marker_manager.ts
```diff
@@ -333,51 +330,39 @@ export class InsertionMarkerManager {
...
-        console.error(
-            'Only one of localConnection_ and closestConnection_ was set.');
-      }
-    } else {  // No connection found.
+  private shouldUpdatePreviews(
+      newCandidate: CandidateConnection|null, dxy: Coordinate): boolean {
+    if (!newCandidate) {
       // Only need to update if we were showing a preview before.
-      return !!(this.localConnection_ && this.closestConnection_);
+      return !!this.activeCandidate;
```
cpcallen, line 342: Hooray for guard clauses.  But this and the following could have been one-liners.

cpcallen (changes requested):
The only mandatory change is removing the extra whitespace in the JSDoc for `hidePreview`.

core/insertion_marker_manager.ts
```diff
@@ -460,150 +447,135 @@ export class InsertionMarkerManager {
...
+    if (closest && closest.targetBlock() &&
+        this.workspace.getRenderer().shouldHighlightConnection(closest)) {
+      closest.unhighlight();
+    }
+    if (this.fadedBlock) {
+      this.hideReplacementFade(this.fadedBlock);
+    } else if (this.highlightedBlock) {
+      this.hideInsertionInputOutline(this.highlightedBlock);
+    } else if (this.markerConnection) {
+      this.hideInsertionMarker(this.markerConnection);
```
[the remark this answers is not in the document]
rachel-fenichel: Calling unconditionally seems worse to me because they *are* mutually exclusive, and I think it should be clear which of the actions is actually going to happen.

core/insertion_marker_manager.ts
rachel-fenichel: I went with `wouldDeleteOnDrop`.

re: property access vs performance benefit: it's essentially acting as a getter right now, which is nice because alternate implementations could change the implementation of `wouldDeleteBlock`. But when I say it out loud, that sounds a lot like designing for a problem we haven't hit. 

@maribethb my position on `public @internal` is that it's not breaking to change them, because during migration we removed `@internal` from things that were actually `public`, and anything left is actually `@internal`. Do you agree?

core/insertion_marker_manager.ts
```diff
@@ -443,14 +437,14 @@ export class InsertionMarkerManager {
    * @returns Whether dropping the block immediately would delete the block.
    */
   private shouldDelete(
-      candidate: CandidateConnection, dragTarget: IDragTarget|null): boolean {
+      candidate: CandidateConnection|null, dragTarget: IDragTarget|null): boolean {
```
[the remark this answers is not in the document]
rachel-fenichel: Done.

rachel-fenichel:
Okay, I've made the changes we discussed for removing previews. PTAL.

core/insertion_marker_manager.ts
maribethb: Agreed, things marked `@internal` should not be considered public and therefore you can change them without it being breaking.

I agree with Christopher's suggestion to remove the get method and just use a property. Re: alternate implementation, is there actually a way to have an alternate implementation (e.g. can you register a custom insertion marker manager?)? if not, then that is more reason why it's not a problem we need to solve, and if so then we actually need to be careful because this might be `protected` rather than `internal`?

rachel-fenichel: Done.

cpcallen (approved):

PR #6603, opened 2022-11-04
chore: add validator and field value types

### LINKED ISSUE #6505: Improve type information for the validator in `./core/field.ts`

`Function`, the type of `validator_` in `./core/field.ts`, doesn't provide the information needed to use it. Additionally, it's a banned type in **Blockly Samples** by default.

1. Upgrade `validator_`'s type in `./core/field.ts`
2. Update field classes in `./core` to reflect that change
3. Update impacted [Blockly Samples plugins](https://github.com/google/blockly-samples/tree/master/plugins)

**Additional context**

See the [field-slider plugin upgrade to TypeScript](https://github.com/google/blockly-samples/pull/1217#discussion_r990442714) in **Blockly Samples** to see the disabled eslint rule and the initial discussion about this issue.

btw17:
## The basics

- [x] I branched from develop

### Resolves

Resolves #6505

### Proposed Changes

- Major changes
  - Added type information for "validator" field functions
  - Updated `core/field.ts` to accept a generic type
    - NOTE: Sets `T = unknown` by default since functions using the base Field should only use T-impacted values in a way that works for all possible types.
  - Split `core/field_textinput.ts` partially into an abstract class `core/field_input.ts` which `core/field_textinput.ts` and `core/field_number.ts` now extend.
    - `FieldTextInput` assumes the values are strings, though `FieldNumber` (which uses number values) was previously a descendant of `FieldTextInput`
- Minor changes
  - Updated `core/field_registry.ts` to have type information reflected by its use
  - Updated `fieldFromJson_` to require `element.type` since `fieldRegistry.fromJson` would fail without it.
  - Added validator functions to main blockly export. (Alternatively, it might be fine just exporting only the base `FieldValidator`)
  - Updated `FieldNumber` to accept only numbers instead of strings and numbers

#### Behavior Before Change

Less precise type information.

#### Behavior After Change

More precise type information for `Field` and descendants of `Field`.

### Reason for Changes

To simplify utilizing `Field` classes. 

### Test Coverage

None

### Documentation

It should only be generated documentation related to type information.

#### Any vs Unknown

Regarding `any` vs `unknown`, there's a nuance in how they can be set between objects/values and functions/classes. Because of this, I think there are some cases where `any` is actually valid. See the code below snippet about the nuance.
- For values/objects which are *setting* `unknown`, you can provide whatever value to `unknown`.
- For functions/classes which are *using* `unknown`, you cannot replace it with functions/classes using anything less than `unknown`, although you can cast it to unknown.

As a result, I think functions, like `appendField`, which can be passed in a descendant of `Field`, like `FieldNumber`, are valid cases for using `Field<any>` parameters instead of forcing the implementation to provide `myFieldNumberClass as Field`. Any thoughts on this behavior?

```typescript
/*********/
export type GenericObject<T> = {
  value: T;
}

let strObj: GenericObject<string> = { value: 'hello' };
let unkObj: GenericObject<unknown> = { value: 'hello' };
let anyObj: GenericObject<any> = { value: 'hello' };
strObj = unkObj; // Error! An `unknown` value may not be a `string`, so can't set.
strObj = anyObj;
unkObj = strObj; // Okay! A `string` value can be an `unknown` value, so can set.
unkObj = anyObj;
/*********/
export type GenericFunction<T> = (value: T) => void;

let strFn: GenericFunction<string> = (val) => {};
let unkFn: GenericFunction<unknown> = (val) => {};
let anyFn: GenericFunction<any> = (val) => {};
strFn = unkFn; // Okay! If the function can do the thing with `unknown`, it can do the thing with `string`.
strFn = anyFn;
unkFn = strFn; // Error! Even if the function can do the thing with `string`, it may not be able to with `unknown`. 
unkFn = strFn as GenericFunction<unknown>; // Okay!
unkFn = anyFn;
/*********/
```


btw17:
I looked into updating some of the `AnyDuringMigration` types associated with the T value, though that spins off into a rabbit hole of minor changes. To keep this change as concise as possible, leaving it mostly as is here.

core/input.ts
```diff
@@ -90,7 +90,8 @@ export class Input {
    *     field again.  Should be unique to the host block.
    * @returns The index following the last inserted field.
    */
-  insertFieldAt(index: number, field: string|Field, opt_name?: string): number {
+  insertFieldAt(index: number, field: string|Field<unknown>, opt_name?: string):
```
BeksOmega: This should also be converted to a generic method, since it's public.

core/input.ts
```diff
@@ -75,7 +75,7 @@ export class Input {
    *     field again.  Should be unique to the host block.
    * @returns The input being append to (to allow chaining).
    */
-  appendField(field: string|Field, opt_name?: string): Input {
+  appendField(field: string|Field<any>, opt_name?: string): Input {
```
BeksOmega, line 78: I think this should be changed to a generic method, rather than `<any>` because `any`. My understanding is that using a generic method (`<T>`) means that the type is infered (external devs don't have to cast using `as Field`), but we still get the safety of using `unknown`.

We can convert the type to `Field` internally at the point we `splice` the field into the `fieldRow`.

core/interfaces/i_registrable_field.ts
```diff
@@ -14,8 +14,11 @@ goog.declareModuleId('Blockly.IRegistrableField');
 
 import type {Field} from '../field.js';
 
-
-type fromJson = (p1: object) => Field;
+/**
+ * This should be satisfied by any function that returns a Field of some type,
+ * so `any` is acceptable.
+ */
+type fromJson = (options: {}) => Field<any>;
```
BeksOmega, line 21: I think this whole file can actually just be deleted (since it's type info anyway, it shouldn't be binary-breaking, but @maribethb to confirm).

This is leftover from the typescript conversion, and we don't actually use the interface originally. Originally we had something in closure that would check if a /static/ `fromJson` method exists. But there's no equivalent way to check for static methods in typescript - so it just converted it to an erroneous interface.

core/field_textinput.ts
```diff
@@ -608,20 +39,16 @@ export class FieldTextInput extends Field {
   }
 }
 
+// TODO DISCUSSION: Should this be `field_textinput`?
 fieldRegistry.register('field_input', FieldTextInput);
```
BeksOmega, line 43: That would be a breaking change :/ So no.

core/dropdowndiv.ts
```diff
@@ -212,12 +212,13 @@ export function showPositionedByBlock(
  * @param opt_secondaryYOffset Optional Y offset for above-block positioning.
  * @returns True if the menu rendered below block; false if above.
  */
-export function showPositionedByField(
-    field: Field, opt_onHide?: Function,
+export function showPositionedByField<T>(
+    field: Field<T>, opt_onHide?: Function,
```
BeksOmega, line 216: Every other function in this namespace that's (1) exported, (2) not marked `@internal` and (3) references `Field` should also be updated to be generic. E.g. `showPositionedByBlock` above.

core/field_image.ts
```diff
@@ -12,21 +12,22 @@
...
-import {FieldConfig, Field} from './field.js';
+import {Field, FieldConfig, FieldValidator} from './field.js';
 import * as fieldRegistry from './field_registry.js';
 import * as dom from './utils/dom.js';
 import * as parsing from './utils/parsing.js';
 import type {Sentinel} from './utils/sentinel.js';
 import {Size} from './utils/size.js';
 import {Svg} from './utils/svg.js';
 
+export type FieldImageValidator = FieldValidator<string>;
```
BeksOmega, line 23: Unused.

core/field_multilineinput.ts
```diff
@@ -25,6 +25,7 @@ import {Svg} from './utils/svg.js';
 import * as userAgent from './utils/useragent.js';
 import * as WidgetDiv from './widgetdiv.js';
 
+export type FieldMultilineInputValidator = FieldTextInputValidator;
```
BeksOmega, line 28: Love the type alias!

BeksOmega (commented):
Generally love this =) Thanks for tackling this!

Some opinions on `<any>` vs `<unknown>` vs `<T>`, and a few other little things.

core/input.ts
btw17: Done!

core/input.ts
btw17: Done!

core/interfaces/i_registrable_field.ts
btw17: Done!

This is currently being used in `core/field_registry.ts` when registering a field class, which each field class does. The field to register is then being passed to `core/registry.ts` which looks like it *would* use the info but currently abandons it via `AnyDuringMigration`. Some notion of `fromJson` could later just be added to the core `Field` class, though.

core/dropdowndiv.ts
btw17: Done!

core/field_image.ts
btw17: Ope! Good catch! Looks like `FieldImage` isn't utilizing a validator type at all.

Another similar yet different case - `core/field_textinput` technically isn't using `FieldTextInputValidator` on its own, though that type works for the provided validator for the constructor. I went back and forth on if `Field` subclasses with validators should all have their own validator functions anyway, even if not called out in their constructors. In the case of `FieldTextInput`, the intellisense will show up as asking for `FieldInputValidator<string>`. It seems like an intuitive pattern to keep with, though - expecting the corresponding type in the corresponding class.

BeksOmega: > Another similar yet different case - core/field_textinput technically isn't using FieldTextInputValidator on its own, though 
that type works for the provided validator for the constructor.

Hmm, for that case, do you think you could just define a pass-through constructor that uses the `FieldTextInputValidator` type? I don't think there's any harm in doing so is there?

btw17: I think the trade off is that there's another point of maintenance for comments, though that's pretty minor. Some of the comments should probably be tailored to the individual class anyway - I'll go ahead and make the update!

btw17: done!

BeksOmega (approved):

Issue #6516, opened 2022-10-10
Add firing events to the procedure data models


BeksOmega:
### Problem Statement

The procedure data models need to fire events in response to change in their data so that devs can respond to these changes. E.g. sharing a procedure with another workspace in response to the creation of the procedure.

Important note: If there is no *actual* change to the procedure model (e.g. renaming "My Procedure" to "My Procedure") then no event should be fired.

https://docs.google.com/document/d/1n4MxZQu49mniC7aUtZ1X4xvwWgpL3st4aDU25pwdaNc/edit#heading=h.s7vgmsbcnh4t

PR #6593, opened 2022-10-31
test: firing procedure events

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

#6517 

### Proposed Changes

Adds unit tests for firing procedure events.

### Reason for Changes

Unit tests are good!

### Test Coverage

Tests for when all of the different types of tests should be fired.

### Documentation

N/A

### Additional Information

N/A

PR #6604, opened 2022-11-04
feat: add firing of procedure events.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

#6516 

### Proposed Changes

Adds firing procedure events to the procedure data models.

### Reason for Changes

Firing events appropriately is good :D

### Test Coverage

Enabled all events from #6593 

### Documentation

N/A

### Additional Information

I only want events to be fired when the procedure models are actually included in the procedure map, so I've added `startPublishing` and `stopPublishing` methods to the `ObservableProcedureModel` and `ObservableParameterModel` to signal that. But I also don't /love/ that solution (it just feels clunky).

`ObservableParameterModel`s also need to have a reference to their parent `IProcedureModel` to be able to properly construct their events, so I've added a `setProcedureModel` method. But theoretically we could also take the `IProcedureModel` in as part of the constructor, so that it's guarunteed to be non-null. That would just break the fluent API. I've added a TODO at the relevant place inline to discuss.


core/procedures/observable_parameter_model.ts
```diff
@@ -26,11 +30,16 @@ export class ObservableParameterModel implements IParameterModel {
    * Sets the name of this parameter to the given name.
    */
   setName(name: string): this {
-    // TODO(#6516): Fire events.
     if (name == this.variable.name) return this;
```
iamgiolaga, line 33:
```suggestion
    if (name === this.variable.name) return this;
```

rachel-fenichel (approved):

PR #6614, opened 2022-11-11
feat: refactor event serialization to use static fromJson method

### LINKED ISSUE #6607: Refactor event serialization

We support serializing events to JSON so that they can be sent over the wire and run in another instance of Blockly.

The current implementation has the deserialization method (`fromJson`) specified as an *instance method* on the event class. This requires constructing an instance of the event class, before you can actually deserialize it. Which in turn requires that every event support an empty constructor. Which in turn requires that every property of every event be nullable (nothing can be required at construction time).

Having every property of a class be nullable is not good TypeScript. It means you have to add lots of unnecessary non-null assertions or checks, which clutter up your main logic.

I would like to refactor the events system to support a *static* `fromJson` method, rather than an *instance* one. This would remove the need for an empty constructor, and nullable properties.

The `Blockly.Event.fromJson` function would be changed to check if the static method exists. If it does, then it would be used. If it doesn't, then we would fall back to the empty constructor + instance method strategy. This makes the refactor backwards compatible.

The detection logic would look something like this (as recommended by Christopher):
```
if (Object.getOwnPropertyDescriptors(C).m && typeof C.m === 'function')) {
  // Use static method
} else {
  // Fall back to empty constructor + instance method
}
```
Note that the above code doesn't mix quoted and dot accessors, so this is compatible with closure compiler.

In this proposal, existing events would maintain:
  * Their empty constructors
  * Their `fromJson` instance methods
  * The nullability of their properties.

Existing events would gain:
  * A `fromJson` static method

And new events (including the new procedure events) would have:
  * A `fromJson` static method
  * No empty constructors
  * No `fromJson` instance methods
  * No nullable properties

N/A

**Additional context**

N/A

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6607 

### Proposed Changes

Adds a new path to `Blockly.Events.fromJson` which uses a static `fromJson` method on the event classes to deserialize them.

Also adds static `fromJson` methods to all existing events.

### Reason for Changes

See #6607 

### Test Coverage

Adds round-trip serialization tests for all events, and updates change detectors in places where serialization was previously broken.

### Documentation

N/A

### Additional Information

I've marked the static `fromJson` methods as `@internal` since I'm not super hapy with the types, and external devs should be calling `Blockly.Events.fromJson` anyway.


core/events/events_block_base.ts
```diff
@@ -32,7 +34,7 @@ export class BlockBase extends AbstractEvent {
    */
   constructor(opt_block?: Block) {
     super();
-    this.isBlank = !!opt_block;
+    this.isBlank = !opt_block;
```
NeilFraser, line 37: Can we have a comment about this property?

NeilFraser (approved):

core/events/events_block_base.ts
BeksOmega: It has a docs comment [here](https://github.com/google/blockly/blob/6122ab22c4207d30681dfb6a3a96588419a9b980/core/events/events_abstract.ts#L28), do you think that's sufficient? I don't necessarily want to duplicate docs that might go out-of-date.

NeilFraser: Oh, cool.  Agreed, don't want to duplicate comments.

That said, that comment looks like Copilot wrote it.  `X.isY` -> "Whether or not X is Y" 🙄  I still have no idea what a 'blank' event means.

BeksOmega: Updated. How's that feel?

NeilFraser: Awesome!

PR #6615, opened 2022-11-11
feat(test): Miscellaneous improvements to `test_tasks.js`

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

* **feat**: Modify `runTestBlock` so that it can run any async task, not just ones that return a `Promise`, by using the `async-done` package (part of Gulp, and already an indirect dependency) to detect task completion.  Celebrate by renaming it to runTestTask.
* Create a `Tester` class to encapsulate the `runTestTask`, `reportTestResult` and `runAll` functions.
* Remove the `--silent` flag from npm scripts so as not to suppress syntax error in gulpfiles.
* Have the `test` task invoke the `buildAdvancedCompilationTest` task (via `onlyBuildAdvancedCompilationTest`, to skip already-run prerequisites) directly, rather than by running `npm`.

### Reason for Changes

* Generally cleaner code.
* Remove npm scripts that aren't intended to be used by humans.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@microsoft/api-extractor": "^7.29.5",
     "@typescript-eslint/eslint-plugin": "^5.33.1",
     "@wdio/selenium-standalone-service": "^7.10.1",
+    "async-done": "^2.0.0",
     "chai": "^4.2.0",
     "clang-format": "^1.6.0",
     "closure-calculate-chunks": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
     "release": "gulp gitCreateRC",
     "start": "concurrently -n tsc,server \"tsc --watch --preserveWatchOutput --outDir 'build/src' --declarationDir 'build/declarations'\" \"http-server ./ -s -o /tests/playground.html -c-1\"",
     "tsc": "gulp tsc",
-    "test": "gulp --silent test",
-    "test:generators": "gulp --silent testGenerators",
+    "test": "gulp test",
+    "test:generators": "gulp testGenerators",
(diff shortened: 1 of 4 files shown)


scripts/gulpfiles/test_tasks.js
```diff
@@ -32,48 +32,84 @@ const BOLD_GREEN = '\x1b[1;32m';
 const BOLD_RED = '\x1b[1;31m';
 const ANSI_RESET = '\x1b[0m';
 
-let failerCount = 0;
+class Tester {
```
BeksOmega, line 35: What's the advantage of encapsulating these in a class?

package.json
```diff
@@ -41,7 +41,6 @@
     "lint:fix": "eslint . --fix",
     "langfiles": "gulp langfiles",
     "minify": "gulp minify",
-    "only:compile:advanced": "gulp onlyBuildAdvancedCompilationTest --debug",
```
BeksOmega, line 44: Yay, goodbye!

BeksOmega (commented):
Slightly confused about the decision to make a Tester class. Otherwise generally LGTM, but want to understand that before approving.

scripts/gulpfiles/test_tasks.js
cpcallen: Mainly just to remove the `failerCount` global variable, and allow us to also count successes without adding a second global.

cpcallen:
There were merge conflicts, and it seems like I forgot a needed `require` from `test_tasks.js` (did I not test this locally?  I'm _sure_ I did, and yet…) so I did a rebase and force-push.

Just awaiting approval, assuming you think my justification for the `Tester` class is sufficient.

PR #6650, opened 2022-11-21
feat: add running procedure events

### LINKED ISSUE #6519: Implement running procedure events
### Problem Statement

Procedure events need to be:
  * Undoable
  * Redoable
  * Runnable in other workspaces

https://docs.google.com/document/d/1n4MxZQu49mniC7aUtZ1X4xvwWgpL3st4aDU25pwdaNc/edit#heading=h.s7vgmsbcnh4t

Related to #6518

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6519 

### Proposed Changes

Adds the ability to run procedure events backwards and forwards.

### Reason for Changes

This is necessary to support undo and redo, as well as sharing procedures between workspaces.

### Test Coverage

Tests were added in #6627 They are now enabled.

### Documentation

N/A

### Additional Information

N/A


core/events/events_procedure_enable.ts
```diff
@@ -9,9 +9,36 @@ import * as registry from '../registry.js';
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
+import {Workspace} from '../workspace.js';
+import {IProcedureModel} from '../interfaces/i_procedure_model.js';
```
rachel-fenichel: Can this be an `import type`?

core/events/events_procedure_create.ts
```diff
@@ -8,9 +8,32 @@ import * as registry from '../registry.js';
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
+import {Workspace} from '../workspace.js';
+import {IProcedureModel} from '../interfaces/i_procedure_model.js';
```
rachel-fenichel: `import type`?

core/events/events_procedure_delete.ts
```diff
@@ -9,9 +9,32 @@ import * as registry from '../registry.js';
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
+import {Workspace} from '../workspace.js';
+import {IProcedureModel} from '../interfaces/i_procedure_model.js';
```
rachel-fenichel: `import type?`

core/events/events_procedure_create.ts
```diff
@@ -4,21 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {IProcedureModel} from '../interfaces/i_procedure_model.js';
+import {ObservableProcedureModel} from '../procedures.js';
 import * as registry from '../registry.js';
+import {Workspace} from '../workspace.js';
```
rachel-fenichel, line 10: `import type`?

core/events/events_procedure_change_return.ts
```diff
@@ -5,19 +5,39 @@
...
+import type {IProcedureModel} from '../interfaces/i_procedure_model.js';
 import * as registry from '../registry.js';
-import {Workspace} from '../workspace.js';
+import type {Workspace} from '../workspace.js';
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
 
 
 export class ProcedureChangeReturn extends ProcedureBase {
```
cpcallen, line 16: Should ~~this class~~ these classes not have a JSDoc explaining its purpose?  It would make it easier for the unfamiliar reviewer…

(Documentation of the properties would be helpful too.)

core/events/events_procedure_change_return.ts
```diff
@@ -5,19 +5,39 @@
...
+
   constructor(
       workpace: Workspace, model: IProcedureModel,
       public readonly oldTypes: string[]|null) {
     super(workpace, model);
+
+    this.newTypes = model.getReturnTypes();
+  }
+
+  run(forward: boolean) {
```
cpcallen, line 28: Is it the convention that runnable events have a `forward` parameter, rather than `backward`?  I'd have thought that the default (`undefined` being falsey) should be to run the event forward…

core/events/events_procedure_change_return.ts
```diff
@@ -5,19 +5,39 @@
...
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
 
 
 export class ProcedureChangeReturn extends ProcedureBase {
+  type = eventUtils.PROCEDURE_CHANGE_RETURN;
+  private newTypes: string[]|null;
+
   constructor(
       workpace: Workspace, model: IProcedureModel,
```
cpcallen, line 21: I appreciate that changing the existing API is probably not sensible, but:

How does it work if you want to run an event from one workspace in another?  Wouldn't it make more sense for `workspace` to be a parameter to `run`?

core/events/events_procedure_create.ts
```diff
@@ -4,13 +4,36 @@
...
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
 
 
-export class ProcedureCreate extends ProcedureBase {}
+export class ProcedureCreate extends ProcedureBase {
+  type = eventUtils.PROCEDURE_CREATE;
+
+  constructor(workspace: Workspace, model: IProcedureModel) {
```
cpcallen, line 19: Isn't this change of signature (constructor now has required arguments) a breaking change?

core/events/events_procedure_delete.ts
```diff
@@ -5,13 +5,36 @@
...
 
 import {ProcedureBase} from './events_procedure_base.js';
 import * as eventUtils from './utils.js';
 
 
-export class ProcedureDelete extends ProcedureBase {}
+export class ProcedureDelete extends ProcedureBase {
+  type = eventUtils.PROCEDURE_DELETE;
+
+  constructor(workspace: Workspace, model: IProcedureModel) {
```
cpcallen, line 20: Breaking change?

core/events/events_procedure_change_return.ts
```diff
@@ -5,19 +5,39 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {IProcedureModel} from '../interfaces/i_procedure_model.js';
+import type {IProcedureModel} from '../interfaces/i_procedure_model.js';
```
cpcallen, line 8: 🎉

core/events/events_procedure_enable.ts
```diff
@@ -5,13 +5,42 @@
...
 
 
-export class ProcedureEnable extends ProcedureBase {}
+export class ProcedureEnable extends ProcedureBase {
+  type = eventUtils.PROCEDURE_ENABLE;
+
+  private oldState: boolean;
+  private newState: boolean;
+
+  constructor(workspace: Workspace, model: IProcedureModel) {
```
cpcallen, line 22: etc.

core/events/events_procedure_enable.ts
```diff
@@ -5,13 +5,42 @@
...
+    super(workspace, model);
+
+    this.oldState = !model.getEnabled();
+    this.newState = model.getEnabled();
```
cpcallen, line 26: Is it really necessary to store both?  Do you foresee in the future emitting events when the enable status "changes" from `true` to `true`, or from `false` to `false`?

tests/mocha/event_procedure_create_test.js
```diff
@@ -131,12 +130,6 @@ suite.skip('Procedure Create Event', function() {
             this.workspace.id);
       });
 
-      test.skip(
-          'a model is not deleted if no model with a matching ID exists',
-          function() {
-            // TODO: Do we want this to throw? warn? do nothing?
-          });
```
cpcallen, line 138: This seems like a useful test.  Why not implement it?

tests/mocha/event_procedure_delete_test.js
```diff
@@ -65,12 +64,6 @@ suite.skip('Procedure Delete Event', function() {
             {model},
             this.workspace.id);
       });
-
-      test(
-          'a model is not deleted if if nodel with a matching ID exists',
-          function() {
-            // TODO: Figure out what we want to do here.
-          });
```
cpcallen, line 73: Ditto here.

tests/mocha/event_procedure_parameter_delete_test.js
```diff
@@ -117,7 +116,7 @@ suite.skip('Procedure Parameter Delete Event', function() {
         const actualProc = this.createProcedureModel('test name', 'test id');
         this.procedureMap.add(actualProc);
 
-        event.run(true /* forward */);
+        event.run(false /* backward */);
```
cpcallen, line 119: While I don't see any rule barring what you've done here, the [styleguide seems to recommend](https://google.github.io/styleguide/jsguide.html#formatting-param-name-comments):
```suggestion
        event.run(/* forward= */ false);
```
(here and below)

core/events/events_procedure_change_return.ts
BeksOmega: Yep, it's defined on the abstract class: https://github.com/google/blockly/blob/485bd4dddf24f3c36db49565de10d8c0b0a3ba5d/core/events/events_abstract.ts#L115

core/events/events_procedure_change_return.ts
BeksOmega: To run an event in another workspace you have to serialize the event and deserialize it into the new workspace. The event system was originally meant for transmitting events over the wire, not running them in different workspaces on the same page. So no one thought about the usecase of passing a workspace to `run`. And at this point, yeah probably not worth changing the API.

core/events/events_procedure_create.ts
BeksOmega: These events/classes haven't been published, so no breaking change (here and below).

tests/mocha/event_procedure_create_test.js
BeksOmega: Do you have any opinions about what the behavior should actually be?

core/events/events_procedure_enable.ts
BeksOmega: I only included both for readability of the `run` function. Otherwise you have to do:

```
    if (forward) {
      procedureModel.setEnabled(this.newState);
    } else {
      procedureModel.setEnabled(!this.newState);
    }
```

Which I think is much more confusing than:

```
    if (forward) {
      procedureModel.setEnabled(this.newState);
    } else {
      procedureModel.setEnabled(this.oldState);
    }
```

core/events/events_procedure_change_return.ts
BeksOmega: Added docs! If you think there's more that could be added I'm happy to do more =)

core/events/events_procedure_create.ts
cpcallen: Whew!

core/events/events_procedure_enable.ts
cpcallen: I admit that having learned to program on a machine with 1kB of memory (and having done most of my university CS assignments on a machine with 100MB of RAM _shared between 60–100 simultaneous users_) might make me excessively concerned about wasting memory, but it does seem kind of distasteful to store the information twice just to make `run` very slightly more readable.

tests/mocha/event_procedure_create_test.js
cpcallen: Ideally: whatever constrains us the least in the future.  Unfortunately, whether we fail silently or throw an error _some_ dev is going to depend on that and will write code that breaks if we change it—but probably safer to throw.

Is there any precedent to handling similar situations in other events?

core/events/events_procedure_change_return.ts
cpcallen: Yes, even these simple (and probably from your point of view slightly redundant) docs are useful to this reader.

cpcallen (approved):

tests/mocha/event_procedure_create_test.js
BeksOmega: Ok I think what we actually want this to do is nothing. that's what the other events do. Plus it makes the most sense for having events be shared between workspaces.

Say you have two workspaces, wsA and wsB. When an event X is fired in wsA you want to run it in wsB. But when you run event X, it fires a new event Y in wsB. So your code tries to run Y in wsA. What we want is for that to just fail silently, and not fire an event.

BeksOmega: So I think that is actually covered by the following test. Does that look correct to you?

PR #6655, opened 2022-11-23
fix: blocks not receiving their own delete events

### LINKED ISSUE #6319: Custom handling when the block is disposed

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6319 

but actually this time.

### Proposed Changes

Makes it so that block change listeners unregister themselves when they receive the delete event for the block they are associated with. This ensures they actually receive that event (i.e. they're not getting removed too early).

#### Behavior Before Change

Blocks were not receiving their own delete events.

#### Behavior After Change

Blocks are receiving their own delete events.

### Reason for Changes

Sometimes you want to trigger cleanup / other operations when a block is deleted.

### Test Coverage

Fixed test to actually wait for the delay built-in to the events system. Before the tests were passing incorrectly because we'd stubbed the fire method to fire immediately.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/tests/mocha/event_block_delete_test.js
+++ b/tests/mocha/event_block_delete_test.js
@@ -10,9 +10,9 @@ import * as eventUtils from '../../build/src/core/events/utils.js';
 import {defineRowBlock} from './test_helpers/block_definitions.js';
 import {sharedTestSetup, sharedTestTeardown} from './test_helpers/setup_teardown.js';
 
-suite('Block Delete Event', function() {
+suite.only('Block Delete Event', function() {
   setup(function() {
-    sharedTestSetup.call(this);
+    sharedTestSetup.call(this, {useFakeTimers: false});
     defineRowBlock();
     this.workspace = new Blockly.Workspace();
   });
@@ -22,9 +22,9 @@ suite('Block Delete Event', function() {
   });
 
   suite('Receiving', function() {
-    test('blocks receive their own delete events', function() {
+    test('blocks receive their own delete events', function(done) {
       Blockly.Blocks['test'] = {
-        onchange: function(e) {},
+        onchange: function(e) { },
       };
       // Need to stub the definition, because the property on the definition is
       // what gets registered as an event listener.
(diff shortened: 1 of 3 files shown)


tests/mocha/event_block_delete_test.js
```diff
@@ -34,8 +34,10 @@ suite('Block Delete Event', function() {
       testBlock.dispose();
 
       const deleteClass = eventUtils.get(eventUtils.BLOCK_DELETE);
-      chai.assert.isTrue(spy.calledOnce);
-      chai.assert.isTrue(spy.getCall(0).args[0] instanceof deleteClass);
+      setTimeout(function() {
```
maribethb, line 37: Does it not work to pass `fireEventsNow: false` in the options, and then use `this.clock.tick()` instead of setting the timeout? This approach works but using the fake clock tick would be more standard imo.

maribethb (approved):

tests/mocha/event_block_delete_test.js
BeksOmega: Using the clock tick doesn't actually test what I want it to test. If I used the clock tick, we get a false positive with the old "fix" for the original bug. This makes sure that the event listener actually gets called with the delay.

maribethb: even if "fire events immediately" is false? that surprises me but this is fine too.

BeksOmega: Nah nah you're right. I thought about it more after I posted my reply, and the issue is ticking immediately after `fire` is called, not ticking in general.

PR #6657, opened 2022-11-25
fix: Add redirects to old demos.

NeilFraser:
There were added redirect.html, but not to app.yaml, and they were obviously never tested.


cpcallen:
Fixes a flaw in #6045.

cpcallen (approved):

PR #771, opened 2016-12-09
Split the scrollbar and flyout out into their own SVG elements. They

picklesrus:
are siblings of the workpsace SVG.  This paves the way to make performance
improvements to workspace dragging.

Note that this is very close to what's in the develop-drag-surface-ws-drag branch.  Main difference is:
- Changed to using translate() instead of translate3d to position the flyout and scrollbars since it is more compatible cross browser and there isn't much of perf gain there anyway.  That means I had to alter the regex code a bit in getRelativeXY

PR #6666, opened 2022-11-29
feat: make `getWorkspace`, `getFlyout`, and `getToolbox` public

NeilFraser:
Amusingly Flyout.isVisible and .setVisible are public, but there's no way to access a flyout given that getFlyout was not public.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -450,7 +450,6 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
    * Get the workspace inside the flyout.
    *
    * @returns The workspace inside the flyout.
-   * @internal
    */
   getWorkspace(): WorkspaceSvg {
     return this.workspace_;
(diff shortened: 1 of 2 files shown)


BeksOmega:
This LGTM but reassigned approval to Maribeth since this is a design decision. Also relevant forum link for context: https://groups.google.com/g/blockly/c/31Om9fkkAWA/m/fQuAg2-JBAAJ

maribethb (approved):
LGTM, but as Rachel said in the forum post, this is not just a documentation change, since we are formalizing these as part of the API. Even before TypeScript, these were marked `@package` so even though they have been used and recommended publicly they have never been actually declared as such.

So I would rename this PR: `feat: make getWorkspace, getFlyout, and getToolbox public` to make better release notes.

cpcallen:
Because I really enjoy admiring [Chesterton's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence): do we know why these were originally `@package` and not `@public`?  @rachel-fenichel

BeksOmega:
> Because I really enjoy admiring [Chesterton's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence): do we know why these were originally `@package` and not `@public`? @rachel-fenichel?

I looked at the original PR (#771) and couldn't find a reason.

rachel-fenichel:
Back then we didn't really talk about the flyout as a public API. There weren't any ways to customize it.

PR #6690, opened 2022-12-09
fix: set default field to any and fix validator function return type

btw17:
## The basics

- [x] I branched from develop

### Resolves

Although the default Field type `T` should eventually be `unknown`, set it to `any` so it does not break users with subclasses extending from `Field` in the meantime.

### Proposed Changes

- Update `core/field.ts` - `FieldValidator` to provide the correct return type.
- Update `core/field.ts` - `T = unknown` to `T = any`

#### Behavior Before Change

TypeScript users with classes extending from `Field` would have to fill in a type for `Field` or their code will break.

#### Behavior After Change

TypeScript users can now extend from `Field` without providing the value type.

### Reason for Changes

This is for introducing the Field type information in multiple stages.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Nope!

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -53,9 +53,9 @@ export type FieldValidator<T = any> = (value?: T) => T|null|undefined;
  * @alias Blockly.Field
  */
 export abstract class Field<T = any> implements IASTNodeLocationSvg,
-                                                    IASTNodeLocationWithBlock,
-                                                    IKeyboardAccessible,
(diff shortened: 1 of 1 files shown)


maribethb (approved):
LGTM, but can you call this a `fix` instead of a `chore`? Chores will now be excluded from the release notes and should only be for internal-facing things developers have no reason to care about. 

Thanks!

PR #6691, opened 2022-12-10
feat: Make 'workspace' param optional on load/save

NeilFraser:
Virtually 100% of calls to load and save (outside of tests) target the main workspace.  So do we need the developer to specify it every time?

Just a thought.  No big deal either way.


rachel-fenichel:
I strongly prefer having them specify the workspace. It means we don't count on assumptions about what's going on in the rest of the page.

NeilFraser:
Ok, withdrawn.

PR #6699, opened 2022-12-12
feat: Display a 'wait' cursor when opening trashcan

NeilFraser:
No visible change for regular case.  But if there's thousands of blocks in the trash, the visual feedback that something's happening is needed.  Otherwise the user will click in confusion, and that click will cause some random block to be spawned out of the trash.

### DIFF
--- a/core/trashcan.ts
+++ b/core/trashcan.ts
@@ -275,7 +275,13 @@ export class Trashcan extends DeleteArea implements IAutoHideable,
     const contents = this.contents_.map(function(string) {
       return JSON.parse(string);
     });
-    this.flyout?.show(contents);
+    // Trashcans with lots of blocks can take a second to render.
+    const blocklyStyle = this.workspace.getParentSvg().style;
+    blocklyStyle.cursor = 'wait';
+    setTimeout(() => {
+      this.flyout?.show(contents);
+      blocklyStyle.cursor = '';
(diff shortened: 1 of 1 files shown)


core/trashcan.ts
```diff
@@ -275,7 +275,13 @@ export class Trashcan extends DeleteArea implements IAutoHideable,
     const contents = this.contents_.map(function(string) {
       return JSON.parse(string);
     });
-    this.flyout?.show(contents);
+    // Trashcans with lots of blocks can take a second to render.
+    const blocklyStyle = this.workspace.getParentSvg().style;
+    blocklyStyle.cursor = 'wait';
+    setTimeout(() => {
```
gonfunko, line 281: I'm sure I'm missing something obvious here, but what's the purpose of the setTimeout? Is it to ensure the wait cursor takes effect/is shown before things get blocked by the call to show()?

NeilFraser: Yup, the browser doesn't update the cursor if it's busy doing other things.  The setTimeout allows the browser a chance to do UI housekeeping.  The bad news is that the minimum value will change between browsers and systems, meaning this is not 100% reliable.  The good news is that failure is not a big deal, in fact it is simply the status quo.

gonfunko (approved):

PR #6745, opened 2023-01-09
chore: refactor procedures namespace

### LINKED ISSUE #6525: Refactor the Procedures namespace
### Problem Statement

Blockly has a `Procedures` namespace, which is basically a catch-all for a bunch of procedure-related utilties, many of which aren't actually general-purpose. It would be nice to clean this up to shrink our public API.

### Proposed Solution

See: https://docs.google.com/document/d/1n4MxZQu49mniC7aUtZ1X4xvwWgpL3st4aDU25pwdaNc/edit#heading=h.o948m9ldldol

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6525 

### Proposed Changes

Refactors the procedures namespace to use the new procedure block interface methods instead of the old ones.

### Reason for Changes

Realized I actually did need to refactor this so that external devs don't have to implement the old `getProcedureDef` and `getProcedureCall` methods.

### Test Coverage

Passes all existing tests.

### Documentation

N/A

### Additional Information

Dependent on #6736

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -286,6 +286,10 @@ const procedureDefGetDefMixin = function() {
       return this.model_;
     },
 
+    isProcedureDef() {
+      return true;
+    },
+
     /**
      * Return all variables referenced by this block.
      * @return {!Array<string>} List of variable names.
@@ -995,6 +999,10 @@ const procedureCallerGetDefMixin = function() {
     getProcedureCall: function() {
       // The NAME field is guaranteed to exist, null will never be returned.
       return /** @type {string} */ (this.getFieldValue('NAME'));
+    },
+
+    isProcedureDef() {
+      return false;
     },
 
     /**
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -286,6 +286,11 @@ const procedureDefGetDefMixin = function() {
       return this.model_;
     },
 
+    /**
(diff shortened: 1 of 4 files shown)


gonfunko (approved):
Do we want to mark this as a breaking change? I realize the root blocks are updated to accommodate it, but since the interface changed it might still be worth calling out?

BeksOmega:
All the changes I made should be backwards compatible (that was my goal anyway). Do you see something I missed?

gonfunko:
> All the changes I made should be backwards compatible (that was my goal anyway). Do you see something I missed?

Just the change to the IProcedureBlock interface; theoretically somebody could have created a class that implements it and doesn't inherit from the core procedure blocks, in which case it would break I think? But I can easily be convinced that's enough of an edge case to not worry about.

BeksOmega:
> > All the changes I made should be backwards compatible (that was my goal anyway). Do you see something I missed?
> 
> Just the change to the IProcedureBlock interface; theoretically somebody could have created a class that implements it and doesn't inherit from the core procedure blocks, in which case it would break I think? But I can easily be convinced that's enough of an edge case to not worry about.

Ah sweet, that was only released in December, and according to our readme: "Once a new API is merged into master it is considered beta until the following release." so we get away with this one scot-free 😎

PR #6746, opened 2023-01-09
fix: proc json serialization

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

N/A

### Proposed Changes

Procedure parameters were not being rendered correctly when loading from JSON. This fixes that.

### Reason for Changes

Properly rendering parameters is important!

### Test Coverage

Added JSON serialization tests equivalent to the existing XML serialization tests.

### Documentation

N/A

### Additional Information

Dependent on #6745

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -591,6 +591,7 @@ const procedureDefMutator = {
       }
     }
 
+    this.doProcedureUpdate();
     this.setStatements_(state['hasStatements'] === false ? false : true);
 
     // Call mutate callers for backwards compatibility.
--- a/tests/mocha/blocks/procedures_test.js
+++ b/tests/mocha/blocks/procedures_test.js
@@ -1985,6 +1985,55 @@ suite('Procedures', function() {
     });
   });
 
+  suite('extra serialization test cases', function() {
+    test('definitions with parameters are properly rendered', function() {
+      Blockly.serialization.workspaces.load({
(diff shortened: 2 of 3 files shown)


blocks/procedures.js
```diff
@@ -430,7 +430,13 @@ const procedureDefUpdateShapeMixin = {
       const block = mutatorWorkspace.getBlockById(p.getId());
       if (!block) continue;  // Should not happen.
       if (block.getFieldValue('NAME') !== p.getName()) {
-        block.setFieldValue(p.getName(), 'NAME');
+        // The field is deterministic based on other events, no need to fire.
+        Events.disable();
```
cpcallen, line 434: Is this a breaking change?

cpcallen (commented):
> Procedure parameters were not being rendered correctly when loading from JSON. This fixes that.

Is there an issue for this?  Or can you be a bit more specific about what was broken and how this fixes it?

Specifically, what I see is that:

- You've disabled events in one place.
- You've renamed one method name (seemingly @internal but not declared as such) to remove one of two trailing underscores.
- You've added tests.

Presumably adding tests did not itself fix the problem, so it would be useful to understand how either of the first two items might.

BeksOmega:
The relevant change is the call to `doProcedureUpdate`!

cpcallen (approved):

PR #6750, opened 2023-01-10
fix: undoing pasting procedure callers that create defs

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

Records undo events for creating a new def to be associated with a deserialized (e.g. pasted) caller so that the events can be grouped with the creation of the caller.

### Reason for Changes

before undoing creating a caller that created its associated definition would not also delete the definition. Now it does (maintaining the behavior on master).

### Test Coverage

Manual testing.

### Documentation

N/A

### Additional Information

N/A


cpcallen (commented):
Moving `createDef_` in the same commit as it is modified makes it very hard to see what changed.

It appears that you've just added a `{recordUndo: true}`—is that correct?

Why move it?

BeksOmega:
Split into two commits, second commit has an explanation for why it moved =)

cpcallen (approved):

PR #6764, opened 2023-01-12
fix(deps): Don't use global variables for jsdom injection in `scripts/package/node/core.js` and `core/utils/xml.ts`

### LINKED ISSUE #6725: Don't use global variables in `scripts/package/node/core.js` and `core/utils/xml.ts`
**Describe the bug**

Due to the use of global variables to pass some (but not all) items obtained from the `jsdom` package from `scripts/package/node/core.js` to `core/utils/xml.ts`, XML serialisation (and therefore Blockly generally) is broken if:

1. Blockly is being loaded in node.js (or another runtime that does not have a native DOM implementation), and
2. more than one copy of Blocly is loaded.

Normally modules cannot be loaded more than once, but multiple copies of a module with different paths, or a single copy that is included in more than one (e.g. webpack) bundle, can defeat the usual deduping of loads to satisfy `import`s and `require`s.

**Specifically, due to this bug, the `@blockly/dev-scripts`'s `test` script, used by plugin developers (including us), is broken when used to test a project with more than one `test/*.mocha.js` test harness when using Blockly v9.2.0.**

**To Reproduce**

1. Clone `blockly-samples`.
2. Choose any plugin with more than one test script in its `test/` drirectory, where at least one of those tests creates a block.
3. Run `npm upgrade blockly` to bump `blockly` to 9.2.0.
4. Run `npm test`
5. Observe test errors.

**Expected behaviour**

Test should pass; no errors should occur.

**Stack Traces**

```
     TypeError: Failed to execute 'serializeToString' on 'XMLSerializer': parameter 1 is not of type 'Node'.
      at Object.exports.convert (webpack-internal:///./node_modules/jsdom/lib/jsdom/living/generated/Node.js:25:9)
      at XMLSerializer.serializeToString (webpack-internal:///./node_modules/jsdom/lib/jsdom/living/generated/XMLSerializer.js:111:23)
      at domToText$$module$build$src$core$utils$xml (webpack-internal:///./node_modules/blockly/blockly_compressed.js:74:333)
```

**Additional context**

See [this lengthy discussion](https://github.com/google/blockly-samples/pull/1452#issuecomment-1364442135) for more detail about the issue.

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6725.

### Proposed Changes

1. Introduce a mechanism for injecting arbitrary dependencies into modules, specifically in this case
to inject required bits of [jsdom](https://github.com/jsdom/jsdom)'s `Window` and `Document` implementations into `core/utils/xml.js` when running in node.js or other environments lacking a DOM.

0. Rename the `xmlDocument` local variable back to `document` (was renamed in #5461) so that the name used in this module corresponds to the [usual global variable](https://developer.mozilla.org/en-US/docs/Web/API/Window/document) it replaces.

0. Change the injection in `scripts/package/node/core.js` to use `injectDependencies` instead of `setDocument` + global variables for `DOMParser` and `XMLSerializer`; also eliminate apparently-unnecessary creation of a special
`Document` instance, using the default one supplied by jsdom instead.

0. Mark `getDocument` and `setDocument` as `@deprecated`, with suitable calls to `deprecation.warn()`.

   There are no remaining callers to either function within this repository—`setDocument` was only used by [the node.js wrapper](https://github.com/google/blockly/blob/develop/scripts/package/node/core.js), and apparently `getDocument` was never used anywhere—and we do not anticipate that either were used by external developers.

#### Behaviour Before Change

If a runtime included multiple separate copies of the blockly package and its dependency jsdom, then [XML serialisation and deserialisation would break](https://github.com/google/blockly-samples/pull/1452#issuecomment-1364442135).

#### Behaviour After Change

It is safe (but still not advisable) to load multiple separate copies of blockly and jsdom into the same runtime.

### Reason for Changes

Fix #6725; allow tests in blockly-samples to pass after updating to Blockly v9.x (e.g. in google/blockly-samples#1452).

### Test Coverage

* Passes `npm test`
* Tests in the branch for google/blockly-samples#1452 pass after using `npm link blockly` to include these changes.

No changes in manual test procedures anticipated.

### Documentation

No changes beyond automatic update including `@deprecation` notice.

### Additional Information

The `injectDependencies` function is intended to be a prototype for a generally applicable mechanism for dependency injection—for example, this approach could be used in other modules to break the circular dependencies that currently cause loading errors when trying to load Blockly via `import * as Blockly from './build/src/core/blockly.js'` without support from the Closure module tooling.  As such, it uses an options object and destructuring-with-defaults to facilitate injecting arbitrarily-named dependencies either one-at-a-time or (as here in `core.js`) in bulk.

### DIFF
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -27,13 +27,24 @@ import * as deprecation from './deprecation.js';
 let {document, DOMParser, XMLSerializer} = globalThis;
 
 /**
- * Inject dependencies.  Used by the Node.js wrapper for Blockly (see
- * scripts/package/node/core.js) to supply implementations from the
- * jsdom package instead.  While they may be set individually, it is
- * normally the case that all three should be supplied and sourced
- * from the same JSDOM instance.
+ * Inject implementations of document, DOMParser and/or XMLSerializer
+ * to use instead of the default ones.
  *
- * @param dependencies Options object contianing dependencies to set.
+ * Used by the Node.js wrapper for Blockly (see
+ * scripts/package/node/core.js) to supply implementations from the
+ * jsdom package instead.
+ *
+ * While they may be set individually, it is normally the case that
+ * all three will be sourced from the same JSDOM instance.  They MUST
+ * at least come from the same copy of the jsdom package.  (Typically
+ * this is hard to avoid satsifying this requirement, but it can be
+ * inadvertently violated by using webpack to build multiple bundles
+ * containing Blockly and jsdom, and then loading more than one into
+ * the same JavaScript runtime.  See
+ * https://github.com/google/blockly-samples/pull/1452#issuecomment-1364442135
+ * for an example of how this happened.)
+ *
+ * @param dependencies Options object containing dependencies to set.
  */
 export function injectDependencies(dependencies: {
   document?: Document,
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -14,6 +14,8 @@
 import * as goog from '../../closure/goog/goog.js';
 goog.declareModuleId('Blockly.utils.xml');
 
+import * as deprecation from './deprecation.js';
+
 
 /**
  * Injected dependencies.  By default these are just (and have the
@@ -57,9 +59,11 @@ export const NAME_SPACE = 'https://developers.google.com/blockly/xml';
  * Get the document object to use for XML serialization.
  *
  * @returns The document object.
+ * @deprecated No longer provided by Blockly.
  * @alias Blockly.utils.xml.getDocument
  */
 export function getDocument(): Document {
+  deprecation.warn('Blockly.utils.xml.getDocument', 'version 9', 'version 10');
   return document;
 }
 
@@ -67,9 +71,11 @@ export function getDocument(): Document {
  * Get the document object to use for XML serialization.
  *
  * @param xmlDocument The document object to use.
+ * @deprecated No longer provided by Blockly.
  * @alias Blockly.utils.xml.setDocument
  */
 export function setDocument(xmlDocument: Document) {
+  deprecation.warn('Blockly.utils.xml.setDocument', 'version 9', 'version 10');
   document = xmlDocument;
 }
 
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -16,19 +16,42 @@ goog.declareModuleId('Blockly.utils.xml');
 
 
 /**
+ * Injected dependencies.  By default these are just (and have the
+ * same types as) the corresponding DOM Window properties, but the
+ * Node.js wrapper for Blockly (see scripts/package/node/core.js)
+ * calls injectDependencies to supply implementations from the jsdom
+ * package instead.
+ */
+let {document, DOMParser, XMLSerializer} = globalThis;
+
+/**
+ * Inject dependencies.  Used by the Node.js wrapper for Blockly (see
+ * scripts/package/node/core.js) to supply implementations from the
+ * jsdom package instead.  While they may be set individually, it is
+ * normally the case that all three should be supplied and sourced
+ * from the same JSDOM instance.
+ *
+ * @param dependencies Options object contianing dependencies to set.
+ */
+export function injectDependencies(dependencies: {
+  document?: Document,
+  DOMParser?: typeof DOMParser,
+  XMLSerializer?: typeof XMLSerializer,
+}) {
+  ({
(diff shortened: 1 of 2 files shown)


core/utils/xml.ts
```diff
@@ -14,40 +14,69 @@
...
+ * package instead.
+ */
+let {document, DOMParser, XMLSerializer} = globalThis;
+
+/**
+ * Inject dependencies.  Used by the Node.js wrapper for Blockly (see
+ * scripts/package/node/core.js) to supply implementations from the
+ * jsdom package instead.  While they may be set individually, it is
+ * normally the case that all three should be supplied and sourced
+ * from the same JSDOM instance.
```
BeksOmega, line 34: Isn't setting thing from different JSDOM instances what caused us trouble in the first place? If so I think this inline documentation should be slightly more opinionated :P

scripts/package/node/core.js
```diff
@@ -18,9 +18,6 @@
 if (typeof globalThis.document !== 'object') {
   const {JSDOM} = require('jsdom');
   const {window} = new JSDOM(`<!DOCTYPE html>`);
-  globalThis.DOMParser = window.DOMParser;
-  globalThis.XMLSerializer = window.XMLSerializer;
-  const xmlDocument = Blockly.utils.xml.textToDomDocument(
-      `<xml xmlns="${Blockly.utils.xml.NAME_SPACE}"></xml>`);
-  Blockly.utils.xml.setDocument(xmlDocument);
+  Blockly.utils.xml.injectDependencies(window);
+  Blockly.utils.xml.injectDependencies({document: window.document});
```
BeksOmega, line 22: Why do you need to inject the document separately?

core/utils/xml.ts
```diff
@@ -14,40 +14,69 @@
...
- * @alias Blockly.utils.xml.NAME_SPACE
+ * @param dependencies Options object contianing dependencies to set.
  */
-export const NAME_SPACE = 'https://developers.google.com/blockly/xml';
+export function injectDependencies(dependencies: {
+  document?: Document,
+  DOMParser?: typeof DOMParser,
+  XMLSerializer?: typeof XMLSerializer,
+}) {
+  ({
```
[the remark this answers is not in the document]
cpcallen: Yes, because `{ /* ... */ } =` at the start of a statement gets parsed as a block followed by an (illegal) `=` token, rather than as a destructuring assignment.

core/utils/xml.ts
cpcallen: No, setting things using values _generated by separate copies of the `jsdom` package_ caused the trouble in the first place.  It's actually weird-but-fine to use separate _instances_ of the `JSDOM` class, provided they all came from the same (`===`) constructor.

But you have a point that the documentation could be more opinionated.

core/utils/xml.ts
```diff
@@ -14,40 +14,69 @@
...
 /**
  * Get the document object to use for XML serialization.
  *
  * @returns The document object.
+ * @deprecated No longer provided by Blockly.
  * @alias Blockly.utils.xml.getDocument
  */
 export function getDocument(): Document {
-  return xmlDocument;
+  deprecation.warn('Blockly.utils.xml.getDocument', 'version 9', 'version 10');
```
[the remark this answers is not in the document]
cpcallen: The long version is: "use the value you passed to `injectDependencies`, or if you didn't do that and are on node.js then just create a new `JSDOM` instance and use its `.window.document`, or if you're in a browser then just use `window.document` aka `globalThis.document` aka `document."

I'm not sure how to summarise that concisely, though, and I'm not sure it's worth bothering, because AFAICT this getter was created (as `document`) in #3126 for the sole purpose of being monkey-patchable by the node.js version of `core.js`.  It should have been deprecated when the only call to it (in this module) was removed in #5461, and I'm having trouble envisioning which "people" might have been using it and for what for…  Why would someone want to get a `Document` instance _from_ Blockly??

scripts/package/node/core.js
cpcallen: Ohh, oops.  This was left over from a bunch of lines I added to test various scenarios (injecting things one at a time, overwriting previous injections, etc.)  Good catch!

BeksOmega (approved):

maribethb (approved):

core/utils/xml.ts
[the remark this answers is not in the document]
rachel-fenichel: "what are you doing and can you not?"

PR #6773, opened 2023-01-13
fix(build): Fix event tests, improve `buildDeps`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

`npm test` output had become increasingly cluttered with miscellaneous WARNINGs from `closure-make-deps`.  Investigating this, it turned out that two recently-added test files were not being run at all.

### Proposed Changes

- Fix actual syntax errors in `import`s in `event_marker_move_test.js` and `event_selected.test.js`, which were preventing those tests from being run.

- Remove `suite.only` directives in those tests that would prevent all the other tests from running.

- Run `closure-make-deps` only once, instead of separately for `core/` and `tests/`.

- Specify a larger `exec` `maxBuffer` size, to ensure output and diagnostics are not truncated.

- Change `stderr` filtering in `buildDeps` to filter out bounded generics messages and blank lines.

- Attempt to suppress warnings in `stderr` output when `closure-make-deps` returns a non-zero exit code.

  Unfortunately, there seems to be a race condition which usually causes the `stderr` argument to the exec callback not to contain the complete output, so in that case print a helpful message.

- Have `buildDeps` just return a `Promise` instead of using a callback.

#### Behaviour Before Change

Two of the tests were not being run.

#### Behaviour After Change

All tests are run; `buildDeps` runs about twice as fast as before.

### Test Coverage

Improved.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -339,7 +339,7 @@ function buildDeps() {
   /**
    * Log unexpected diagnostics, after removing expected warnings.
    *
-   * @param {string} test Standard error output from closure-make-deps
+   * @param {string} text Standard error output from closure-make-deps
    */
   function log(text) {
     for (const line of text.split('\n')) {
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -309,71 +309,78 @@ function buildJavaScript(done) {
  *
  * Prerequisite: buildJavaScript.
  */
-function buildDeps(done) {
+function buildDeps() {
   const roots = [
     path.join(TSC_OUTPUT_DIR, 'closure', 'goog', 'base.js'),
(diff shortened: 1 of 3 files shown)


scripts/gulpfiles/build_tasks.js
```diff
@@ -309,71 +309,78 @@ function buildJavaScript(done) {
...
-    return text.split('\n')
-        .filter(
-            (line) => !/^WARNING /.test(line) ||
-                !(/Missing type declaration./.test(line) ||
-                  /illegal use of unknown JSDoc tag/.test(line)))
-        .join('\n');
+  /**
+   * Log unexpected diagnostics, after removing expected warnings.
+   *
+   * @param {string} test Standard error output from closure-make-deps
```
NeilFraser, line 342: s/test/text

NeilFraser (approved):
Other than the @param typo, LGTM.

PR #6782, opened 2023-01-18
refactor: Don't use skew_ and translate_ attributes on SVGs to animate blocks.

### LINKED ISSUE #6595: Refactor block animations to not store data on SVG elements
**Describe the bug**

Currently the block animations code stores data on properties of SVG element objects, which are also referenced in the main block code. This design has caused us problems in the past (namely https://github.com/google/blockly/issues/6541) because it is difficult to refactor the code when you're dealing with side effects on "private" properties that are triggered by other files.

**Suggested solution**

I think the easiest thing is just to add some public getters and setters for handling the skew and the translation (so that the block code and the block animation code can both mess with it). But there may also be a more elegant solution that does better at information hiding.

**Additional Info**

Related to #6559

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Fixes #6595

### Proposed Changes
This change stops setting skew_ and translation_ attributes on SVGs for blocks. It introduces an internal getter for the translation, and stores it in a proper field on BlockSvg.

skew_ was only being set to a non-'' value in one place, and then immediately read back to set the transform. Instead, the transform is simply setting using the directly-calculated skew value, so there's no need to carry it around and clear it at other points.

WorkspaceCommentSvg also maintained the skew_ and translate_ attributes, but they appear to be vestigial from a long-ago copy-paste and not actually used, so they were removed.

### DIFF
--- a/core/block_animations.ts
+++ b/core/block_animations.ts
@@ -28,8 +28,8 @@ interface CloneRect {
 /** PID of disconnect UI animation.  There can only be one at a time. */
 let disconnectPid: ReturnType<typeof setTimeout>|null = null;
 
-/** SVG group of wobbling block.  There can only be one at a time. */
-let disconnectGroup: SVGElement|null = null;
+/** The wobbling block.  There can only be one at a time. */
+let wobblingBlock: BlockSvg|null = null;
 
 
 /**
@@ -163,18 +163,18 @@ export function disconnectUiEffect(block: BlockSvg) {
     magnitude *= -1;
   }
   // Start the animation.
-  disconnectGroup = block.getSvgRoot();
-  disconnectUiStep(disconnectGroup, magnitude, new Date());
+  wobblingBlock = block;
+  disconnectUiStep(block, magnitude, new Date());
 }
 
 /**
  * Animate a brief wiggle of a disconnected block.
  *
- * @param group SVG element to animate.
+ * @param block Block to animate.
  * @param magnitude Maximum degrees skew (reversed for RTL).
  * @param start Date of animation's start.
  */
-function disconnectUiStep(group: SVGElement, magnitude: number, start: Date) {
+function disconnectUiStep(block: BlockSvg, magnitude: number, start: Date) {
   const DURATION = 200;  // Milliseconds.
   const WIGGLES = 3;     // Half oscillations.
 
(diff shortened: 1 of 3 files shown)


core/block_svg.ts
```diff
@@ -506,13 +516,8 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       this.workspace.getBlockDragSurface()!.translateSurface(
           newLoc.x, newLoc.y);
     } else {
-      (this.svgGroup_ as AnyDuringMigration).translate_ =
-          'translate(' + newLoc.x + ',' + newLoc.y + ')';
-      (this.svgGroup_ as AnyDuringMigration)
-          .setAttribute(
-              'transform',
-              (this.svgGroup_ as AnyDuringMigration).translate_ +
-                  (this.svgGroup_ as AnyDuringMigration).skew_);
```
BeksOmega, line 515: Why did you remove setting the skew value here? Won't this mess up the animation where the block wobbles when you disconnect it?

Or was the assignment in `setDragging` overwriting the skew value anyway, so it doesn't matter?

core/block_animations.ts
```diff
@@ -28,8 +28,8 @@ interface CloneRect {
 /** PID of disconnect UI animation.  There can only be one at a time. */
 let disconnectPid: ReturnType<typeof setTimeout>|null = null;
 
-/** SVG group of wobbling block.  There can only be one at a time. */
-let disconnectGroup: SVGElement|null = null;
+/** The wobbling block.  There can only be one at a time. */
+let wobblingBlock: BlockSvg|null = null;
```
BeksOmega, line 32: Thank you for the rename this is much better haha.

core/block_animations.ts
```diff
@@ -202,13 +200,12 @@ function disconnectUiStep(group: SVGElement, magnitude: number, start: Date) {
  * @internal
  */
 export function disconnectUiStop() {
-  if (disconnectGroup) {
+  if (wobblingBlock) {
```
BeksOmega, line 203: Optional nit: `if (!wobblingBlock) return;`

core/block_svg.ts
gonfunko: AFAICT no, because the disconnect animation calculates skew (and updates `transform` using it) in a 10ms loop, so it will continue to do so and override the lack of skew from here. Experimentally there was no visually perceptible difference and the animation was still shown, but please do double check me on this!

BeksOmega: Pulled it down and checked, and I agree looks fine =) Sweetness!

BeksOmega (approved):

core/block_animations.ts
gonfunko: Done.

PR #6785, opened 2023-01-20
chore: Remove some more uses of AnyDuringMigration

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
Part of #5857 

### Proposed Changes
Removes uses of AnyDuringMigration from workspace_comment_svg.ts, toolbox.ts, field_colour.ts, field_image.ts, and workspace.ts.

### DIFF
--- a/core/interfaces/i_selectable_toolbox_item.ts
+++ b/core/interfaces/i_selectable_toolbox_item.ts
@@ -61,3 +61,11 @@ export interface ISelectableToolboxItem extends IToolboxItem {
    */
   onClick(_e: Event): void;
 }
+
+/**
+ * Type guard that checks whether an IToolboxItem is an ISelectableToolboxItem.
+ */
+export function isSelectableToolboxItem(toolboxItem: IToolboxItem):
+    toolboxItem is ISelectableToolboxItem {
+  return toolboxItem.isSelectable();
+}
--- a/core/toolbox/toolbox.ts
+++ b/core/toolbox/toolbox.ts
@@ -29,6 +29,7 @@ import type {IDraggable} from '../interfaces/i_draggable.js';
 import type {IFlyout} from '../interfaces/i_flyout.js';
 import type {IKeyboardAccessible} from '../interfaces/i_keyboard_accessible.js';
 import type {ISelectableToolboxItem} from '../interfaces/i_selectable_toolbox_item.js';
+import {isSelectableToolboxItem} from '../interfaces/i_selectable_toolbox_item.js';
 import type {IStyleable} from '../interfaces/i_styleable.js';
 import type {IToolbox} from '../interfaces/i_toolbox.js';
 import type {IToolboxItem} from '../interfaces/i_toolbox_item.js';
@@ -100,7 +101,6 @@ export class Toolbox extends DeleteArea implements IAutoHideable,
    * Ex: [[node, name, func], [node, name, func]].
    */
   protected boundEvents_: browserEvents.Data[] = [];
-  override wouldDelete_: AnyDuringMigration;
 
   /** The workspace this toolbox is on. */
   protected readonly workspace_: WorkspaceSvg;
@@ -112,10 +112,9 @@ export class Toolbox extends DeleteArea implements IAutoHideable,
     this.workspace_ = workspace;
 
     /** The JSON describing the contents of this toolbox. */
-    // AnyDuringMigration because:  Type 'ToolboxInfo | { contents: never[]; }'
-    // is not assignable to type 'ToolboxInfo'.
-    this.toolboxDef_ = (workspace.options.languageTree || {'contents': []}) as
-        AnyDuringMigration;
+    this.toolboxDef_ =
+        (workspace.options.languageTree ||
+         {'contents': new Array<toolbox.ToolboxItemInfo>()});
 
     /** Whether the toolbox should be laid out horizontally. */
     this.horizontalLayout_ = workspace.options.horizontalLayout;
@@ -299,9 +298,8 @@ export class Toolbox extends DeleteArea implements IAutoHideable,
     if (!handled && this.selectedItem_) {
       // TODO(#6097): Figure out who implements onKeyDown and which interface it
       // should be part of.
-      const untypedItem = this.selectedItem_ as AnyDuringMigration;
-      if (untypedItem.onKeyDown) {
-        handled = untypedItem.onKeyDown(e);
+      if ((this.selectedItem_ as any).onKeyDown) {
+        handled = (this.selectedItem_ as any).onKeyDown(e);
       }
     }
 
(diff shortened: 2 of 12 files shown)


core/toolbox/toolbox.ts
```diff
@@ -112,10 +112,9 @@ export class Toolbox extends DeleteArea implements IAutoHideable,
     this.workspace_ = workspace;
 
     /** The JSON describing the contents of this toolbox. */
-    // AnyDuringMigration because:  Type 'ToolboxInfo | { contents: never[]; }'
-    // is not assignable to type 'ToolboxInfo'.
-    this.toolboxDef_ = (workspace.options.languageTree || {'contents': []}) as
-        AnyDuringMigration;
+    this.toolboxDef_ =
+        (workspace.options.languageTree ||
+         {'contents': new Array<toolbox.ToolboxItemInfo>()});
```
cpcallen, line 117: [TypeScript code must not use the Array constructor](https://google.github.io/styleguide/tsguide.html#array-constructor) (even though it seems pretty readable here).  It seems [`: TypeName` is preferred](https://google.github.io/styleguide/tsguide.html#type-assertions-and-object-literals), though I'm not completely clear how to action that advice here.

Also, single quotes around `contents` not needed.

core/toolbox/toolbox.ts
```diff
@@ -299,9 +298,8 @@ export class Toolbox extends DeleteArea implements IAutoHideable,
     if (!handled && this.selectedItem_) {
       // TODO(#6097): Figure out who implements onKeyDown and which interface it
       // should be part of.
-      const untypedItem = this.selectedItem_ as AnyDuringMigration;
-      if (untypedItem.onKeyDown) {
-        handled = untypedItem.onKeyDown(e);
+      if ((this.selectedItem_ as any).onKeyDown) {
+        handled = (this.selectedItem_ as any).onKeyDown(e);
```
cpcallen, line 302: It doesn't appear that `ISelectableToolboxItem` has an `.onKeyDown`, so I'm slightly unclear about how this is expected to work.  Does the interface need to be modified?  (Noting that it would be a breaking change.)

core/workspace.ts
```diff
@@ -172,22 +172,11 @@ export class Workspace implements IASTNodeLocation {
...
-    // AnyDuringMigration because:  Property 'offset' does not exist on type
-    // '(a: Block | WorkspaceComment, b: Block | WorkspaceComment) => number'.
-    return aXY.y +
-        (Workspace.prototype.sortObjects_ as AnyDuringMigration).offset *
-        aXY.x -
-        (bXY.y +
-         (Workspace.prototype.sortObjects_ as AnyDuringMigration).offset *
-             bXY.x);
+    const offset =
+        Math.sin(math.toRadians(Workspace.SCAN_ANGLE)) * (this.RTL ? -1 : 1);
```
cpcallen, line 176: Yes, this is definitely preferable to having an option property on the function object!

The only thing I wonder is if the `.offset` was ever being set to some other value by developers (sort of a monkey patch, given this function is `private`!)  e.g.: @NeilFraser, does Blockly Games ever set this (and can it stop, please?)

core/workspace_comment.ts
```diff
@@ -154,7 +154,7 @@ export class WorkspaceComment {
    *   This is not valid if the comment is currently being dragged.
    * @internal
    */
-  getXY(): Coordinate {
+  getRelativeToSurfaceXY(): Coordinate {
```
cpcallen, line 157: This rename would appear to be a breaking change to the API.  Is that intended?  If so, please update commit message and PR title accordingly.  If not, please (at most) deprecate `getXY` rather than removing it.

cpcallen (changes requested):
The main blocker is that this appears to contain a breaking change.

core/workspace.ts
NeilFraser: Blockly Games does not use `.offset`.  I don't recognize that property.

core/workspace_comment_svg.ts
```diff
@@ -867,17 +835,15 @@ export class WorkspaceCommentSvg extends WorkspaceComment implements
...
-    // to parameter of type 'string'.
-    this.foreignObject_!.setAttribute(
-        'width', size.width as AnyDuringMigration);
-    this.foreignObject_!.setAttribute(
-        'height', (size.height - topOffset).toString());
+    this.foreignObject_?.setAttribute('width', `${size.width}`);
+    this.foreignObject_?.setAttribute('height', `${size.height - topOffset}`);
     if (this.RTL) {
-      this.foreignObject_!.setAttribute('x', (-size.width).toString());
+      this.foreignObject_?.setAttribute('x', `${- size.width}`);
```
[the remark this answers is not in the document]
gonfunko: Yeah, I strongly prefer no space but clang-format kept wanting to fight me on it. Not sure if we can change our config to prevent that...

core/workspace_comment.ts
gonfunko: This method is `@internal`, so I was under the impression we can do whatever we want with it?

core/toolbox/toolbox.ts
gonfunko: As you noted, `: TypeName` doesn't work here; I wasn't able to find any other way to provide type information about an empty, inlined array literal. It might be possible to declare a variable and use that in place of the literal, but since this is only used if workspace.options.languageTree isn't set, I don't love that either. I think the Array constructor is the best bet, particularly since we're not passing any values to it here, so the unexpected behavior issue raised in the styleguide isn't applicable. Happy to change it if you have a specific suggestion/preference though!

core/workspace_comment_svg.ts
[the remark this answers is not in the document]
gonfunko: Indeed making this change causes the lint/format part of CI to fail, so I've reverted it.

core/workspace_comment.ts
cpcallen: I'd like to get @rachel-fenichel and/or @maribethb's take on that.  My understanding is that there are some cases where we have previously suggested using (previously) `@private` methods and now treat them as if they had been `@public` for the purpose of deciding what's a breaking change.

If this method is not one we've ever suggested people use then SGTM.

core/toolbox/toolbox.ts
cpcallen: I think that's a reasonable enough argument, though I expect at some time in the future this will produce a lint error (i.e., when the lint rules get specific enough to enforce the no-`new Array()` rule).

cpcallen (approved):
LGTM but please check about whether the `@internal` method really is sufficiently internal.

core/interfaces/i_selectable_toolbox_item.ts
```diff
@@ -61,3 +61,11 @@ export interface ISelectableToolboxItem extends IToolboxItem {
    */
   onClick(_e: Event): void;
 }
+
+/**
+ * Type guard that checks whether an IToolboxItem is an ISelectableToolboxItem.
+ */
+export function isSelectableToolboxItem(toolboxItem: IToolboxItem):
+    toolboxItem is ISelectableToolboxItem {
```
rachel-fenichel, line 69: i-dont-understand-is-it-modern

(I'll go look at the typescript docs--I know I've skimmed pass this section before.)

core/workspace_comment.ts
rachel-fenichel: I think this is safe--we rarely recommend devs do anything with workspace comments, and I don't remember us ever suggesting people use this. And we use the same name (getRelativeToSurfaceXY) for the same functionality in block/block_svg.

rachel-fenichel (approved):

PR #6810, opened 2023-02-01
feat: Add support for centering on a block itself vs its stack.

### LINKED ISSUE #5432: centerOnBlock centers on stack
The workspace method _centerOnBlock_ uses the block's _getHeightWidth_ to calculate its center point, but that returns the size of the block including the stack below it, so the center of the stack is used as focus point. The block that I want to highlight to the user can get moved off screen by this. Using the block's own width and height instead works fine.

To reproduce, load a long stack like this on the playground:
```xml
  <block type="controls_repeat_ext" id="|EJ:%,F@Xek)^?D/MKcJ" x="88" y="63">
    <value name="TIMES">
      <shadow type="math_number" id="4nvyR*s(VEkSM;qF,.}e">
        <field name="NUM">10</field>
      </shadow>
    </value>
    <statement name="DO">
      <block type="text_print" id="tQhTU%5$@Y4K[FCV=QGr">
        <value name="TEXT">
          <shadow type="text" id="0)JyE?fA`SNPe?f8|qM1">
            <field name="TEXT">abc</field>
          </shadow>
        </value>
        <next>
          <block type="controls_if" id="IzY7=mrt)3GJ;MzhWC%8">
            <mutation elseif="20"></mutation>
          </block>
        </next>
      </block>
    </statement>
  </block>
</xml>
```

Zoom in a bit and center on the print block with `Blockly.mainWorkspace.centerOnBlock('tQhTU%5$@Y4K[FCV=QGr')`

gonfunko:
## The basics

- [x] I branched from develop

### Resolves
#5432 

### Proposed Changes
Adds an optional argument to centerOnBlock() that, if set to true, will center the workspace on the block itself, rather than the default behavior of centering on that block's stack.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -1983,11 +1983,13 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
 
   /**
    * Scroll the workspace to center on the given block. If the block has other
-   * blocks stacked below it, the workspace will be centered on the stack.
+   * blocks stacked below it, the workspace will be centered on the stack,
+   * unless blockOnly is true.
    *
    * @param id ID of block center on.
+   * @param blockOnly True to center only on the block itself, not its stack.
    */
-  centerOnBlock(id: string|null) {
+  centerOnBlock(id: string|null, blockOnly?: boolean) {
     if (!this.isMovable()) {
       console.warn(
           'Tried to move a non-movable workspace. This could result' +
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
LGTM!

Was looking to see if we had any relevant tests for this code, and it seems our only test is actually broken (it calls `zoomToFit` instead of `centerOnBlock`): https://github.com/google/blockly/blob/7cbdf0d87e54fc9bab24e2ca6e9509e23ff1fd58/tests/mocha/workspace_svg_test.js#L208

This doesn't actually test the code you wrote :P But would you mind cleaning it up rq? If not that's cool, I can grab it!

gonfunko:
> 

Done!

PR #6818, opened 2023-02-03
refactor: Remove last remaining circular `import` in `core/`

### LINKED ISSUE #6817: Make it possible to load Blockly directly from `tsc`-generated ESMs
It should be possible to load a (working) version of Blockly by doing `import * as Blockly from 'build/src/core/blockly.js'` (in, for example, `tests/playground.html`)

## Background

Previous attempts to load Blockly from the `.js` files generated by `tsc` have failed because of module loading order issues, particularly (but perhaps not exclusively) due to cyclic `import`s.

Specifically, in the case where there are ECMAScript modules ('ESM') `a.js` and `b.js`, which define `class A` and `class B extends A` respectively, and `b.js` `import`s `a.js`, it is normally safe to `import` either or both of these modules in any order:
* If `a.js` is imported first, then `A` is defined; if `b.js` is subsequently imported then it `import`s `a.js` and gets the already-initialised `class A` from `a`'s already-extant`Module` object.
* If `b.js` is imported first, then it is fetched but before it is executed the interpreter notes that it `import`s `a.js`, so execution of `b` is postponed until `a.js` has been fetched and executed, then `b.js` is executed, and accesses `A` as before.

However, if there is an import cycle such that `a.js` also imports `b.js` (directly or indirectly), then a problem can occur:
* If `b.js` is imported first then it works as in the previous case, with the additional note that the execution of `a.js` is postponed temporarily to resolve its `import`s, one of which is `b.js`—but since `b.js` is already in the process of being loaded, no actual code is executed at this point, and `a.js` gets `b.js`'s empty `Module` object, which has no properties yet but will get them later once the loader unwinds the stack of imports back to `b.js`.  As long as `a.js` doesn't do a destructuring import, or try to use any properties exported by b _at load time_ (as opposed to later, in response to calls to exported functions/methods) then all is still well.
* On the other hand, if `a.js` is loaded first, its execution will be postponed to load `b.js`, and `b.js` will get `a.js`'s empty, uninitialised `Module` object, and this will cause a failure because `class B extends class A` throws `TypeError` if `B` is `undefined`—which it most certainly will be in this case because the `class A` declaration has not yet been evaluated because `a.js` has not been executed because it is waiting for its dependencies to finish loading first.

And it's not just `extends` declarations which are an issue: any use of either a destructuring `import` (`import {foo, bar baz} from …`) or attempting to use the value of a property on a `Module` object at load time (e.g. `import * as someModule from '…/some_module.js'; someModule.someFunction();`) will causes similar issues if the module that commits this 'infraction' is part of a dependency cycle and the cycle is loaded starting 'from the wrong place'.

Additionally:

* Prior to conversion from `goog.module` to ESM, `closure-make-deps` enforced a rule against circular `require`s because `goog.module`semantics do not allow circular `require`s, and presumably there is still some degree of enforcement of this within google3 because of the use of tsickle to convert TS to Closure-style JS, but the open-source version of `closure-make-deps` does not enforce this rule for `.js` files containing ESM.

## Possible approaches

A number of approaches to solving this issue have been mooted; they are not necessarily mutually-exclusive:

1. Refactoring what code lies in which source file, so as to be able to get rid of cycles in the import graph.
0. For each cycle, identify a canonical entry point and ensure that all modules which import any module from that cycle begin by importing the identified entry-point.
   * This would mean imposing rules on `import` statement order that might conflict with whatever our styleguide / tooling might prefer (at the moment the ordering is pretty arbitrary, a result of inconsistent changes made since we migrated to named `goog.require`s in the absence of any lint rules.)
   * Some modules might have to import an identified entry point module for load order reasons only, even though that import was unused.
   * It might be wise to add checks to ensure that the identified entry point had been imported first, by checking to see if the imported `Module` objects have defined properties: they should be defined in every module in a cycle except ones which import the identified entry point, which should see the entry-point's `Module` object having all exported properties `undefined` while loading.
0. Using the [internal module pattern](https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de)[1], where—either globally for all of Blockly, or locally for each distinct import cycle or interconnected set of cycles—there is a single "switchboard" module which imports and then reexports all of the other modules in the cycle in the correct order, and each module within the cycle imports only that switchboard module.
0. Using dependency injection to break the `import` cycles, as we have already done in some other places ([e.g. `Workspace.prototype.newBlock`](https://github.com/google/blockly/blob/299016e78d8d83e972f1947036c7f7c7dfde6c7b/core/workspace.ts#L510), injected [in `blockly.ts`](https://github.com/google/blockly/blob/75963ea58e65c6ef63eebeedfee99b59c14a3e0e/core/blockly.ts#L544))

[1] (what we have sometimes called a "megamodule", though i am unsure where this term came from as it does not seem to be used outside our team)

## Then, a miracle occurs…[*](https://www.researchgate.net/figure/Then-a-Miracle-Occurs-Copyrighted-artwork-by-Sydney-Harris-Inc-All-materials-used-with_fig2_302632920)

On 2023-02-03, even before @cpcallen finished writing the full description for this issue, he and @rachel-fenichel began investigating which files were involved in import cycles.  Although in the past there had been substantial numbers of files in cycles of up to 4–5 links circumference, it turned out that there was only one remaining cycle in `core/`: between `variables.ts` and `xml.ts`, which was easily solvable by moving the `textToDom` function from `core/xml.ts` to `core/utils/xml.ts`.

## Notes

* For now we will need to continue loading Blockly via `bootstrap.js` (and the Closure debug module loader) in tests and playgrounds, because the blocks and generators are still in `goog.module` format.  Separate bugs will be opened to track that work.
* We may wish to continue loading Blockly via a common script like `bootstrap.js` even once we have removed our dependency on the closure module loader for the other features offered, such as easy switching between compiled and uncompiled mode.

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6817.

### Proposed Changes

* Move `textToDom` from `core/xml.ts` to `core/utils/xml.ts`.
* Re-enable `goog.declareModuleId` multiple-call check in `closure/goog/base.js`.

### Reason for Changes

The `textToDom` function being in `core/xml.ts` is the cause for the last remaining circular `import` in `core/` (between `variables.ts` and `xml.ts`).  Moving it to `core/utils/xml.ts` makes sense anyway, since there is nothing Blockly-specific about this function.

Reenabling the assertion which checks to make sure that `goog.declareModuleId` is not called more than once in a module will also catch circular `import`s amongst ES modules, which are not detected by `closure-make-deps`).

### Test Coverage

Passes `npm test`.

No manual test changes expected.

### DIFF
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -119,7 +119,6 @@ export function createTextNode(text: string): Text {
  * @returns A DOM object representing the singular child of the document
  *     element.
  * @throws if the text doesn't parse.
- * @alias Blockly.utils.xml.textToDom
  */
 export function textToDom(text: string): Element {
   const doc = textToDomDocument(text);
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -113,6 +113,24 @@ export function createTextNode(text: string): Text {
 }
 
 /**
+ * Converts an XML string into a DOM structure.
+ *
+ * @param text An XML string.
+ * @returns A DOM object representing the singular child of the document
+ *     element.
+ * @throws if the text doesn't parse.
+ * @alias Blockly.utils.xml.textToDom
+ */
+export function textToDom(text: string): Element {
+  const doc = textToDomDocument(text);
+  if (!doc || !doc.documentElement ||
+      doc.getElementsByTagName('parsererror').length) {
+    throw Error('textToDom was unable to parse: ' + text);
+  }
+  return doc.documentElement;
+}
+
+/**
  * Converts an XML string into a DOM tree.
  *
  * @param text XML string.
--- a/appengine/storage.js
+++ b/appengine/storage.js
@@ -44,7 +44,7 @@ BlocklyStorage.restoreBlocks = function(opt_workspace) {
   var url = window.location.href.split('#')[0];
   if ('localStorage' in window && window.localStorage[url]) {
     var workspace = opt_workspace || Blockly.getMainWorkspace();
-    var xml = Blockly.Xml.textToDom(window.localStorage[url]);
+    var xml = Blockly.utils.xml.textToDom(window.localStorage[url]);
     Blockly.Xml.domToWorkspace(xml, workspace);
   }
 };
@@ -168,7 +168,7 @@ BlocklyStorage.monitorChanges_ = function(workspace) {
  */
 BlocklyStorage.loadXml_ = function(xml, workspace) {
   try {
-    xml = Blockly.Xml.textToDom(xml);
+    xml = Blockly.utils.xml.textToDom(xml);
   } catch (e) {
     BlocklyStorage.alert(BlocklyStorage.XML_ERROR + '\nXML: ' + xml);
     return;
--- a/blocks/lists.js
+++ b/blocks/lists.js
@@ -13,7 +13,6 @@
 goog.module('Blockly.libraryBlocks.lists');
 
 const xmlUtils = goog.require('Blockly.utils.xml');
-const Xml = goog.require('Blockly.Xml');
 const {Align} = goog.require('Blockly.Input');
 /* eslint-disable-next-line no-unused-vars */
 const {Block} = goog.requireType('Blockly.Block');
@@ -480,7 +479,7 @@ blocks['lists_getIndex'] = {
       this.updateStatement_(true);
     } else if (typeof state === 'string') {
       // backward compatible for json serialised mutations
-      this.domToMutation(Xml.textToDom(state));
+      this.domToMutation(xmlUtils.textToDom(state));
     }
   },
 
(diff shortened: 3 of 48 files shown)


cpcallen:
Leaving this as a draft for the moment because it turns out I missed  some calls to `textToDom` from outside `core/` that should also be updated to call the version in `utils/xml.ts`.

core/utils/xml.ts
```diff
@@ -112,6 +112,24 @@ export function createTextNode(text: string): Text {
...
 }
 
+/**
+ * Converts an XML string into a DOM structure.
+ *
+ * @param text An XML string.
+ * @returns A DOM object representing the singular child of the document
+ *     element.
+ * @throws if the text doesn't parse.
+ * @alias Blockly.utils.xml.textToDom
```
rachel-fenichel, line 122: You don't need `alias` anymore (and I'm working on deleting it everywhere else)

cpcallen: Done.

rachel-fenichel (approved):
LGTM once you fix conflicts.

PR #6827, opened 2023-02-07
fix: Fire deletion events when clearing variables.

NeilFraser:
https://groups.google.com/g/blockly/c/l_vUnrGSJ0M


koenvanwijk (commented):
I think the console.log still needs to be removed. Sorry you fixed in in a later commit. Please ignore.

cpcallen (approved):

Issue #1534, opened 2018-01-10
Clean up public APIs that should be calling setGroup()


RoboErikG:
### Problem statement

We aren't super consistent with when setGroup(true) and setGroup(false) get called in various APIs.

In theory, these should wrap the implementation of our public APIs (renames, moves, creations, etc.) but not appear in the internal implementation methods so that they can be chained without always checking if we're already in a group.

In practice we have checks and calls to setGroup() at a few layers, so unless we're confident we've cleaned up the majority of these cases we shouldn't remove calls, but we should make sure they at least check if there's already a group before setting/unsetting it.

PR #6829, opened 2023-02-07
fix: Don't clober event group when renaming vars

NeilFraser:
Also audit all existing event group commands and tweak a few of them where I think there's a potential issue.

See https://groups.google.com/g/blockly/c/dKizQbKb8vo


core/contextmenu_items.ts
```diff
@@ -279,7 +278,8 @@ export function registerDeleteAll() {
       }
       scope.workspace.cancelCurrentGesture();
       const deletableBlocks = getDeletableBlocks_(scope.workspace);
-      const eventGroup = idGenerator.genUid();
+      eventUtils.setGroup(true);
```
gonfunko, line 281: If the user chooses cancel on the dialog, deleteNext_ is never run and grouping is never turned off. deleteNext_ was already calling setGroup() with the passed-in ID - why is this change necessary?

core/workspace_svg.ts
```diff
@@ -1253,21 +1253,22 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
-
     let pastedThing;
-    // Checks if this is JSON. JSON has a type property, while elements don't.
-    if (state['type']) {
-      pastedThing = this.pasteBlock_(null, state as blocks.State);
-    } else {
-      const xmlBlock = state as Element;
-      if (xmlBlock.tagName.toLowerCase() === 'comment') {
-        pastedThing = this.pasteWorkspaceComment_(xmlBlock);
+    try {
```
gonfunko, line 1257: What was throwing an exception here?

core/contextmenu_items.ts
NeilFraser: This changes eliminates the dependency on idGenerator.  It also means that in Blockly there would be only one place which sets eventGroup -- so if we change the group format in the future, we won't need to change it in two places.

core/workspace_svg.ts
NeilFraser: If the clipboard is corrupted, the XML parser might throw.

gonfunko:
Does this also address #1534?

gonfunko (approved):

Issue #6828, opened 2023-02-07
Convert `blocks/` and `generators/` to TypeScript


cpcallen:
This is a tracking bug for the remaining work to migrate the Blockly codebase to TypeScript, following the successful migration of `core/`.

This is part of #5857, and will resolve #6248 and probably also #2995.

## Work to be done

* [x] Update `scripts/gulpfiles/build_tasks.js` to feed the contents of `blocks/` and `generators/` through `tsc`.
  * Done: #6836
* [x] Update `tests/bootstrap.js`, playgrounds, etc. to load blocks and generators from `build/src/` when loading uncompressed.
  * Turned out to be ~no work for now as they use `goog.require` so don't care about file location.
* [x] Create script to assist with migration (esp. of `import`s).
  * Done: #6837
* [ ] [OPTIONAL:] Refine type of `BlockDefinition` (in `core/blocks.ts`) to make defining blocks easier and preferably less error prone.
  * At the moment `BlockDefinition` is `AnyDuringMigration`; this _works_ but means that `tsc` provides no assistance in verifying block definitions—for example, it does not check to make sure that `Block` methods are being called (or overridden) correctly.
  * @cpcallen has [asked a question on Stack Overflow](https://stackoverflow.com/q/75403774/4969945) but this has not elicited an answer yet.
  * @btw17 has made some useful suggestions, including in particular defining `BlockDefinition` as `{[Property in keyof Block]?: Block[Property]} & {[key: string]: unknown}`, but this is not perfect: e.g. mixin methods do not get the correct type for `this`.
* [x] Create a model PR, migrating one file in `blocks/`, to verify that the above works as intended.
  * See #6900.

### `blocks/`

* [x] blocks.js (#7193)
* [x] colour.js (#6901)
* [x] lists.js (#6902)
* [x] logic.js (#7003)
* [x] loops.js (#6957)
* [x] math.js (#6900)
* [x] procedures.js (#7192 )
* [x] text.js (#6958)
* [x] variables.js (#7001)
* [x] variables_dynamic.js (#7001)

### `generators/`

* [ ] #7326
* [ ] #7401

* [x] dart.js
  * [x] dart/dart_generator.js
  * [x] dart/colour.js
  * [x] dart/lists.js
  * [x] dart/logic.js
  * [x] dart/loops.js
  * [x] dart/math.js
  * [x] dart/procedures.js
  * [x] dart/text.js
  * [x] dart/variables.js
  * [x] dart/variables_dynamic.js
[...]

PR #6836, opened 2023-02-09
feat(build): Run `tsc` on `blocks/` and `generators/`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6828

### Proposed Changes

Modify `build_tasks.js` so that `blocks/` and `generators/` are passed through `tsc` before being fed to `closure-make-deps`, Closure Compiler, etc.

#### Behaviour Before Change

Closure Compiler and debug module loader read from `blocks/` and `generators/` directly.

#### Behaviour After Change

`blocks/` and `generators/` are included in `tsc`'s input, and Closure Compiler and debug module loader read from `build/src/blocks/` and `build/src/generators/`.

### Reason for Changes

Allow conversion of blocks and generators to TypeScript.

### Test Coverage

Changes untested so far; leaving this PR as a draft until I've verified that everything actually works as expected.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -35,11 +35,6 @@ const {posixPath} = require('../helpers');
 ////////////////////////////////////////////////////////////
 
 /**
- * Directory in which core/ can be found after passing through tsc.
- */
-const CORE_DIR = path.join(TSC_OUTPUT_DIR, 'core');
-
-/**
  * Suffix to add to compiled output files.
  */
 const COMPILED_SUFFIX = '_compressed';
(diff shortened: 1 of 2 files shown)


cpcallen:
OK, my first pass at converting `blocks/math.js` to TypeScript—though not pretty!—is getting built fine, so I think the changes in this PR ready for review.

tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "include": [
     "core/**/*",  // N.B.: also pulls in closure/goog/goog.js if needed.
```
BeksOmega, line 3: Why does this one pull in goog.js if needed, but the other closure files need to be included explicitly?

scripts/gulpfiles/build_tasks.js
```diff
@@ -691,14 +682,13 @@ function buildAdvancedCompilationTest() {
...
-    TSC_OUTPUT_DIR + '/core/**/*.js',
-    'blocks/**/*.js',
-    'generators/**/*.js',
+    TSC_OUTPUT_DIR + '/**/*.js',
     'tests/compile/main.js',
     'tests/compile/test_blocks.js',
   ];
+  const ignore = [
+    TSC_OUTPUT_DIR + '/closure/goog/base.js',  // Use base_minimal.js only.
+  ];
```
BeksOmega, line 691: Optional nit: Personally I think it would be clearer to provide the two files explicitly (goog & base_minimal) rather than excluding base.js

tsconfig.json
cpcallen: Because apparently `closure-make-deps` will follow `import` statements.  (It can't do that for `goog.require`s, as they don't have filepaths.)  It might be, actually, that we could just include `core/blockly.ts` and that would be enough…

cpcallen: > Because apparently `closure-make-deps` will follow `import` statements. (It can't do that for `goog.require`s, as they don't have filepaths.) It might be, actually, that we could just include `core/blockly.ts` and that would be enough…

Sorry, this is totally wrong because brain was not engaged.  The correct answer is:

Because `tsc` will follow `import` statements.

And it _is_ sufficient to include only `blockly.ts`, `any_aliases.ts` and `main.js` (the three files that are not imported by any other file), but I think it is a little weird to do that.

scripts/gulpfiles/build_tasks.js
cpcallen: That was more or less how it was before—with us having to explicitly enumerate each of the subdirectories of `build/` (now including `blocks/` and `generators/`.  I think it's a lot tidier to say "compile everything output by `tsc`, except for this one specific file."  (Especially when that one specific file will likely go away before long.)

cpcallen:
@BeksOmega: Were there any other problems with this PR that you felt needed to be addressed before approving?

scripts/gulpfiles/build_tasks.js
BeksOmega: Ah ok, if it's theoretically going away soon that makes a lot of sense. SGTM!

BeksOmega (approved):
Nope just wanted to make sure I understood everything that was happening =) LGTM!

Issue #6786, opened 2023-01-21
Refactor rendering to use a queue


BeksOmega:
**Is your feature request related to a problem? Please describe.**

Currently we are doing a lot of [unnecessary rendering](https://github.com/google/blockly/issues/6130#issuecomment-1116686580) that doesn't actually get painted to the screen. Experimentally I've found an [~5x performance improvement](https://github.com/google/blockly/issues/6130#issuecomment-1397507224) from eliminating these extra renders.

The reason we're getting extra renders is that currently every time a something on a block is changed (e.g. it is connected, disconnected, an input is added, a field value is changed, etc) It gets immediately rerendered, and all of its parent blocks are also rerendered to accomodate that.

Related, we sometimes also set the `rendered` property on blocks to false to try to disable extra renders, but this leads to lots of bugs. This proposal would also fix that problem by eliminating the need to set `rendered` to false.

**Describe the solution you'd like**

Create a "queue" of blocks to be rendered, and then trigger that rendering from a [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) callback. The queue will actually be a tree of blocks that need to be rendered. And the rendering will be triggered depth-first (from leaf nodes to parent nodes) so that for each "batch of block changes" will only require each affected block to be rerendered once.

More specifically I think the API should look like the following.

* internal `RenderManagement` module
  * internal `function markDirty(block: BlockSvg)`
  * private `function triggerRenders()`
     Passed to the `requestAnimationFrame` callback

  * internal `class BlockTree`
     This is the tree of blocks that will be walked when rendering.
    * public `add(block:BlockSvg)`
    * public `remove(block: BlockSvg)`
    * public `has(block: BlockSvg): boolean`
    * public `getNodeFor(block:BlockSvg): Node`
    * public `getRootNode(): Node`
    * pulbic `clear()`

  * internal `class Node`
    * public `block?: BlockSvg`
       Optional so the root does not need tobe associated with a block
    * public `parent?: Node`
    * public `children: Node[]`

**Describe alternatives you've considered**

I considered making the BlockTree and Node class much more generic. E.g. so they would be programmed against some interface like:

```ts
interface INode {
  id: string;
  getParent(): INode;
}
```

But decided that since this logic is very specific to how we want to handle rendering, there wasn't much use in the additional abstraction. If we ever need a `BlockTree` for something else, we can always pull it out of the `RenderManagement` module and generalize it :P

**Implementation Plan**

[...]

PR #6859, opened 2023-02-21
fix: updating connections in the db multiple times

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Works on #6786 as a follow up to #6851 

### Proposed Changes

Separates out updating connection locations into a separate pass, specifically when using the render queue (normal immediate rendering is unaffected).

### Reason for Changes

This means that we are no longer recursively moving all of the child blocks every time we tighten a parent block (which was causing us to do a lot of duplicate work adding and removing things from the connection db). It also eliminates a heck-ton of calls to `getRelativeToSurfaceXY`, which is a pretty expensive function.

This makes rerendering ~ 2.5x faster when you're moving a bunch of connections. E.g. when you're wrapping a new if-block around the spaghetti stress test blocks.

### Test Coverage

Just manual testing =(

### Documentation

N/A

### Additional Information

Dependent on #6851

### DIFF
--- a/core/render_management.ts
+++ b/core/render_management.ts
@@ -4,11 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {BlockSvg} from './block_svg';
+import {BlockSvg} from './block_svg.js';
+import {Coordinate} from './utils/coordinate.js';
 
 
-const rootBlocks = new Set<BlockSvg>();
-const dirtyBlocks = new WeakSet<BlockSvg>();
+let rootBlocks = new Set<BlockSvg>();
+let dirtyBlocks = new WeakSet<BlockSvg>();
 let pid = 0;
 
 /**
@@ -43,11 +44,22 @@ function queueBlock(block: BlockSvg) {
  * Rerenders all of the blocks in the queue.
  */
 function doRenders() {
+  const workspaces = new Set([...rootBlocks].map((block) => block.workspace));
   for (const block of rootBlocks) {
+    // No need to render a dead block.
     if (block.isDisposed()) continue;
+    // Rendering will be triggered through the real root block.
+    if (block.getParent()) continue;
+
     renderBlock(block);
+    updateConnectionLocations(block, block.getRelativeToSurfaceXY());
+  }
+  for (const workspace of workspaces) {
+    workspace.resizeContents();
   }
 
+  rootBlocks = new Set();
+  dirtyBlocks = new Set();
   pid = 0;
 }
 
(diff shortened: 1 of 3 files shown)


core/render_management.ts
```diff
@@ -58,14 +70,29 @@ function doRenders() {
...
+}
+
+/**
+ * Updates the connection database with the new locations of all of the
+ * connections that are children of the given block.
+ *
+ * @param block The block to update the connection locations of.
+ * @param blockTL The top left of the given block in workspace coordinates.
+ */
+function updateConnectionLocations(block: BlockSvg, blockTL: Coordinate) {
```
maribethb, line 87: nit: would prefer a different name like "location" -- the fact that it's the top left of the block isn't particularly relevant here and `blockTL` is a fairly inscrutable name imo

core/render_management.ts
```diff
@@ -43,11 +44,22 @@ function queueBlock(block: BlockSvg) {
  * Rerenders all of the blocks in the queue.
  */
 function doRenders() {
+  const workspaces = new Set([...rootBlocks].map((block) => block.workspace));
   for (const block of rootBlocks) {
+    // No need to render a dead block.
     if (block.isDisposed()) continue;
+    // Rendering will be triggered through the real root block.
```
maribethb, line 51: Like we discussed offline, consider adding a comment explaining why a "root" block isn't a "real root block"

core/rendered_connection.ts
```diff
@@ -261,12 +273,26 @@ export class RenderedConnection extends Connection {
...
       const xy = svgMath.getRelativeXY(svgRoot);
-      block!.getSvgRoot().setAttribute(
-          'transform', 'translate(' + (xy.x - dx) + ',' + (xy.y - dy) + ')');
+      block!.translate(xy.x - dx, xy.y - dy);
       block!.moveConnections(-dx, -dy);
     }
   }
 
+  /**
+   * Moves the block son either side of this connection right next to eachother,
```
maribethb, line 282: s/block son/blocks on
s/eachother/each other

core/render_management.ts
```diff
@@ -43,11 +44,22 @@ function queueBlock(block: BlockSvg) {
...
+    if (block.getParent()) continue;
+
     renderBlock(block);
+    updateConnectionLocations(block, block.getRelativeToSurfaceXY());
+  }
+  for (const workspace of workspaces) {
+    workspace.resizeContents();
   }
 
+  rootBlocks = new Set();
```
maribethb, line 61: why not `rootBlocks.clear()` so you can keep rootBlocks a const?

maribethb (approved):

core/render_management.ts
BeksOmega: How do you feel about `blockOrigin`?

Personally I think the fact that it's the top left of the block is pretty important, but that may just be a me thing. Knowing that it's at the top left helps me visualize how all of the offsets in the positioning logic add together.

core/render_management.ts
BeksOmega: Tell me if this is better!

core/render_management.ts
maribethb: sgtm.

What I mean by not relevant is that if this coordinate were relative to the bottom right of the block instead, as long as all the coordinates were using the same relative corner, nothing in this method would change. It would be somewhere further down in the calls that cares about which specific corner we are looking at, in terms of bumping and positioning, that cares which corner it is. but in any case, `blockOrigin` seems good regardless :)

PR #6861, opened 2023-02-24
fix: remove update disabled calls

BeksOmega:
## The basics

- [x] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

\
Removes some unnecessary calls to `updateDisabled` and `applyColour`.

### Reason for Changes

`updateDisabled` is a very expensive operation because it recursively walks the block tree and applies styles to it. This moves color application to be triggered by `renderEfficiently` so we only have to do one walk of the block tree.

Everywhere where calls to `updateDisabled` were removed, there is a later call to `queueRender`. This means there is no change in behavior.

### Test Coverage

Onlymanual testing :/

### Documentation

N/A

### Additional Information

Dependent on #6860

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -511,7 +511,7 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
     if (!collapsed) {
       this.updateCollapsed_();
     } else if (this.rendered) {
-      this.render();
+      this.queueRender();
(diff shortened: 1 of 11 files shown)


BeksOmega:
Closing because this did not work as expected. Triggering the style updates later actually made this perform slower, because it caused style recalculations to occur twice.

PR #4184, opened 2020-08-17
Fix trashcan errors on delete event

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)

### Resolves

Closes #4048

### Proposed Changes

Dispose of lastMarker_ before reassigning it.

### Reason for Changes

The lastMarker_ was not being properly disposed when it was being reassigned by initAvailableConnections. This meant a block was getting left behind in the workspace's topBlocks_ array. So when you deleted all of the blocks, it would try to fire a delete event for the insertion marker. The insertion marker doesn't seem to serialize properly? so it would cause the trashcan to throw errors when it received the event.

### Test Coverage

Manually tested and could not reproduce #4048

Tested on:
 * Desktop Chrome

### Documentation

N/A

### Additional Information

N/A

PR #6875, opened 2023-03-01
fix: Don't repeatedly recreate the last marker block during a drag.

### LINKED ISSUE #6839: Insertion Marker Manager creates unused marker blocks.
### Check for duplicates

### Description

When a stack of block is dragged (even on an otherwise empty workspace), every pixel of drag triggers the creation of a new insertion marker block.  Despite that no insertion marker is shown on screen.

This is a performance issue.  It is also the source of unneeded calls to domToMutation.
Reported externally: https://groups.google.com/g/blockly/c/-eK657F9qCM

### Reproduction steps

1. Add a console.log('New Marker Block!') into the top of createMarkerBlock function of insertion_marker_manager.js
2. On an empty workspace, create a stack of any two statement blocks.
3. Open the console, and drag the stack of two blocks.

Expect: zero (or maybe one) instances of 'New Marker Block!'
Actual: one 'New Marker Block!' per pixel dragged.

### Stack trace

```shell
domToMutation	@	logic.js:332
createMarkerBlock	@	insertion_marker_manager.js:181
initAvailableConnections	@	insertion_marker_manager.js:239
updateAvailableConnections	@	insertion_marker_manager.js:102
getCandidate	@	insertion_marker_manager.js:290
update	@	insertion_marker_manager.js:148
drag	@	block_dragger.js:140
handleMove	@	gesture.js:466
wrapFunc	@	browser_events.js:59
```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

- [x] I branched from develop

### Resolves

Fixes  #6839

### Proposed Changes
#5722 introduced logic to update the available connections (and marker block) for a block being dragged as it was being dragged in order to make the marker block reflect changes to the connections of the block being dragged that happened during the drag. The insertion marker corresponding the first block in the stack being dragged was only regenerated when (a) it was shown and (b) its connections were inconsistent with those on the corresponding insertion marker block. However, the insertion marker corresponding to the last block in the stack was recreated every time the drag moved. This PR updates the insertion marker manager to only create the insertion marker block corresponding to the last block in the stack being dragged (a) at the beginning of the drag and (b) in the event of the same sort of mismatch during showing that would cause the insertion marker block corresponding to the first block in the stack being dragged to be regenerated.

#### Behavior Before Change
Numerous useless block creations.

#### Behavior After Change
Insertion marker blocks are only created on an as-needed basis.

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -133,6 +133,11 @@ export class InsertionMarkerManager {
     this.firstMarker = this.createMarkerBlock(this.topBlock);
 
     this.availableConnections = this.initAvailableConnections();
+
+    if (this.lastOnStack) {
+      this.lastMarker =
+          this.createMarkerBlock(this.lastOnStack.getSourceBlock());
+    }
   }
 
   /**
@@ -142,18 +147,8 @@ export class InsertionMarkerManager {
    */
   dispose() {
     this.availableConnections.length = 0;
-
-    eventUtils.disable();
-    try {
-      if (this.firstMarker) {
-        this.firstMarker.dispose();
-      }
-      if (this.lastMarker) {
-        this.lastMarker.dispose();
-      }
-    } finally {
-      eventUtils.enable();
-    }
+    this.disposeInsertionMarker(this.firstMarker);
+    this.disposeInsertionMarker(this.lastMarker);
   }
 
   /**
@@ -289,10 +284,8 @@ export class InsertionMarkerManager {
   }
 
   /**
-   * Populate the list of available connections on this block stack.  This
-   * should only be called once, at the beginning of a drag. If the stack has
-   * more than one block, this function will populate lastOnStack and create
-   * the corresponding insertion marker.
+   * Populate the list of available connections on this block stack. If the
+   * stack has more than one block, this function will also update lastOnStack.
    *
    * @returns A list of available connections.
    */
@@ -303,9 +296,6 @@ export class InsertionMarkerManager {
     if (lastOnStack && lastOnStack !== this.topBlock.nextConnection) {
       available.push(lastOnStack);
       this.lastOnStack = lastOnStack;
-      if (!this.lastMarker) {
-        this.lastMarker = this.createMarkerBlock(lastOnStack.getSourceBlock());
-      }
     }
     return available;
   }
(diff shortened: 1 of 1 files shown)


core/insertion_marker_manager.ts
```diff
@@ -303,15 +303,9 @@ export class InsertionMarkerManager {
...
       available.push(lastOnStack);
       this.lastOnStack = lastOnStack;
-      if (this.lastMarker) {
-        eventUtils.disable();
-        try {
-          this.lastMarker.dispose();
-        } finally {
-          eventUtils.enable();
-        }
+      if (!this.lastMarker) {
```
BeksOmega, line 306: Instead of doing this here, can you move it to the constructor & update the JSDoc for this function?

Like:
```
this.firstMarker = /* etc */;
this.initAvailableConnections();
if (this.lastOnStack) this.lastMarker = this.createMarkerBlock(/* etc */);
```

core/insertion_marker_manager.ts
```diff
@@ -561,8 +555,31 @@ export class InsertionMarkerManager {
...
-      insertionMarker = isLastInStack ? this.lastMarker : this.firstMarker;
+      if (isLastInStack && this.lastOnStack) {
+        if (this.lastMarker) {
+          eventUtils.disable();
+          try {
+            this.lastMarker.dispose();
+          } finally {
+            eventUtils.enable();
+          }
+        }
```
BeksOmega, line 566: Can you break this out into a sub procedure? E.g. `disposeInsertionMarker`

core/insertion_marker_manager.ts
```diff
@@ -561,8 +555,31 @@ export class InsertionMarkerManager {
...
+        insertionMarker = this.lastMarker;
+      } else {
+        if (this.firstMarker) {
+          eventUtils.disable();
+          try {
+            this.firstMarker.dispose();
+          } finally {
+            eventUtils.enable();
+          }
+        }
```
BeksOmega, line 578: Oof, were we just eating memory before?

BeksOmega (approved):

core/insertion_marker_manager.ts
gonfunko: I assume the GC would have cleaned up after us eventually since the marker would have become unreachable, but you introduced this logic in #4184 to fix an issue so I'm carrying it forward.

core/insertion_marker_manager.ts
gonfunko: Good call, done!

core/insertion_marker_manager.ts
gonfunko: Done.

Issue #6130, opened 2022-04-29
Improve performance of inserting blocks into stacks.


BeksOmega:
**Describe the bug**

Current performance on https://github.com/google/blockly/commit/7eebd78b5dbf0195f08cd817716bd802753e0216:
![Screenshot 2022-04-29 9 27 37 AM](https://user-images.githubusercontent.com/25440652/165985923-82f084b3-3ec7-42d9-a2f4-2771e582b829.png)

Time to execute endDrag = ~52 ms averaged over 5 trials.

Convert the following txt file to a json file and then load it into your Chrome Dev Tools to view the results:
[inserting-1.txt](https://github.com/google/blockly/files/8592304/inserting-1.txt)

**To Reproduce**

1. Go to the playground.
2. Load the following JSON:
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "controls_if",
        "inputs": {
          "DO0": {
            "block": {
              "type": "controls_if"
            }
          }
        }
      },
      {
        "type": "controls_if"
      }
    ]
  }
}
```
4. Go to the performance tab of the chrome devtools.
5. Enable recording screenshots.
6. Set CPU Throttling to 6x.
7. Start a recording.
8. Insert the single block into the stack of two blocks so that it looks like this:
![image](https://user-images.githubusercontent.com/25440652/165986799-a2eb876c-09ad-45be-8c19-f8447e3e63b0.png)

10. Stop the recording.

**Expected behavior**

Ideally, we wouldn't drop frames when inserting blocks. For this to be achieved, inserting needs to complete within 10ms.

**Additional context**

It looks like `hidePreview_` is doing a lot of work (~22ms), and triggering a lot of re-renders I'm hoping that can be cleaned up.

Similarly with `connect` (~16ms).

This one is going to take some more digging though.

BeksOmega:
So after some work on this, imo removing the extra render calls needs to be more of a project (with an actual design doc and whatnot). Currently our only way to delay/remove some of these render calls is to set `block.rendered = false` and `block.rendered = true` which is error prone, and has historically caused us problems as documented in #4288

Also going to link this to  #5375 as refactoring insertion markers will have to take into account when we trigger renders.

BeksOmega:
To clarify my previous comment: When we're inserting a block, we're actually "rendering" a lot of "stages" of connecting the blocks that never get drawn to the screen. If you use breakpoints you can step through and see them.

After first "dropping" the block:
[...]

PR #6876, opened 2023-03-02
fix: improve performance of connecting blocks

BeksOmega:
## The basics

- [x] I branched from develop

### Resolves

Related to #6130 

### Proposed Changes

Optimizes the performance of connecting blocks by:
  * Removing unnessary calls to `updateDisabled`
  * Only setting the parent once (when connecting) instead of twice (when connecting and disconnecting)

### Reason for Changes

Cuts off ~100ms from wrapping an if block around the spaghetti code test blocks.

### Test Coverage

Only manual testing :/

### Documentation

N/A

### Additional Information

Dependent on #6860

### DIFF
--- a/core/connection.ts
+++ b/core/connection.ts
@@ -93,7 +93,7 @@ export class Connection implements IASTNodeLocationWithBlock {
 
     // Make sure the childConnection is available.
     if (childConnection.isConnected()) {
-      childConnection.disconnect(false);
+      childConnection.disconnectInternal(false);
     }
 
     // Make sure the parentConnection is available.
@@ -104,7 +104,7 @@ export class Connection implements IASTNodeLocationWithBlock {
       if (target!.isShadow()) {
         target!.dispose(false);
       } else {
-        this.disconnect();
+        this.disconnectInternal();
         orphan = target;
       }
       this.applyShadowState_(shadowState);
@@ -245,11 +245,20 @@ export class Connection implements IASTNodeLocationWithBlock {
 
   /**
    * Disconnect this connection.
+   */
+  disconnect() {
+    this.disconnectInternal();
+  }
+
+  /**
+   * Disconnect two blocks that are connected by this connection.
    *
    * @param setParent Whether to set the parent of the disconnected block or
    *     not, defaults to true.
+   *     If you do not set the parent, ensure that a subsequent action does,
+   *     otherwise the view and model will be out of sync.
    */
-  disconnect(setParent = true) {
+  protected disconnectInternal(setParent = true) {
     const {parentConnection, childConnection} =
         this.getParentAndChildConnections();
     if (!parentConnection || !childConnection) {
--- a/core/connection.ts
+++ b/core/connection.ts
@@ -93,7 +93,7 @@ export class Connection implements IASTNodeLocationWithBlock {
 
     // Make sure the childConnection is available.
     if (childConnection.isConnected()) {
-      childConnection.disconnect();
+      childConnection.disconnect(false);
     }
 
     // Make sure the parentConnection is available.
@@ -244,7 +244,7 @@ export class Connection implements IASTNodeLocationWithBlock {
   }
 
   /** Disconnect this connection. */
-  disconnect() {
+  disconnect(setParent = true) {
     const {parentConnection, childConnection} =
         this.getParentAndChildConnections();
     if (!parentConnection || !childConnection) {
@@ -264,7 +264,7 @@ export class Connection implements IASTNodeLocationWithBlock {
       otherConnection.targetConnection = null;
     }
     this.targetConnection = null;
-    childConnection.getSourceBlock().setParent(null);
+    if (setParent) childConnection.getSourceBlock().setParent(null);
     if (event) {
       event.recordNew();
       eventUtils.fire(event);
--- a/core/connection.ts
+++ b/core/connection.ts
@@ -293,6 +293,9 @@ export class Connection implements IASTNodeLocationWithBlock {
   }
 
   /**
+   * Returns the parent connection (superior) and child connection (inferior)
+   * given this connection and the connection it is connected to.
+   *
    * @returns The parent connection and child connection, given this connection
    *     and the connection it is connected to.
    */
--- a/core/connection.ts
+++ b/core/connection.ts
@@ -245,65 +245,52 @@ export class Connection implements IASTNodeLocationWithBlock {
 
   /** Disconnect this connection. */
   disconnect() {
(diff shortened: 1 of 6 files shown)


core/insertion_marker_manager.ts
```diff
@@ -604,38 +599,16 @@ export class InsertionMarkerManager {
...
-      }
-
-      const previousBlockNextConnection =
-          markerPrev ? markerPrev.targetConnection : null;
-
-      imBlock.unplug(true);
-      if (previousBlockNextConnection && innerConnection) {
-        previousBlockNextConnection.connect(innerConnection);
-      }
+    if (!markerPrev?.targetConnection && !markerOutput?.targetConnection) {
```
BeksOmega, line 605: @rachel-fenichel I think this works, but could you PTAL and tell me if I'm missing anything?

BeksOmega:
@NeilFraser This is ready for a look! Might be advisable to review this commit-wise.

core/insertion_marker_manager.ts
rachel-fenichel: All of the complexity here is to handle a case that doesn't happen in Blockly but does happen in Scratch.

In Blockly if you drag a C-shaped block up a stack, it will attempt to insert, rather than wrap any of the blocks. When you get to the top of the stack, it'll offer to wrap the whole thing.
38bd759f-301d-4e7e-871e-c31cf749daff


In Scratch, it'll offer to wrap from the current position all the way to the end of the stack.

ddd008ab-cd13-4b8b-ae21-97a16cec7982

Disentangling the insertion marker when it's wrapping around blocks in the middle of the stack is handled by this code:
```
else if (markerConn.type === ConnectionType.NEXT_STATEMENT && !isNext) {
      // Inside of a C-block, first statement connection.
      const innerConnection = markerConn.targetConnection;
      if (innerConnection) {
        innerConnection.getSourceBlock().unplug(false);
      }

      const previousBlockNextConnection =
          markerPrev ? markerPrev.targetConnection : null;

      imBlock.unplug(true);
      if (previousBlockNextConnection && innerConnection) {
        previousBlockNextConnection.connect(innerConnection);
      }
    } 
```

This won't happen in base Blockly, because our implementation of `doDragChecks` in the `ConnectionChecker` doesn't allow it. But if you override `doDragChecks` you can definitely get back into this position, and Scratch's decision to do it was an active design choice, not an accident.

core/insertion_marker_manager.ts
```diff
@@ -180,19 +180,14 @@ export class InsertionMarkerManager {
    */
   applyConnections() {
     if (!this.activeCandidate) return;
-    // Don't fire events for insertion markers.
+    const {local, closest} = this.activeCandidate;
+    local.connect(closest);
```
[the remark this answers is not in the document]
BeksOmega: The events for this were fired before as well (the disable/enable only wraps `hidePreview`). I want this to happen first to eliminate duplicate calls to `setParent`.

core/insertion_marker_manager.ts
BeksOmega: Coolio, I'll write a unit test to document this behavior! and then see if I can implement this + still get the performance improvments.

BeksOmega: I found out that you can't achieve scratch-style insertion behavior behavior just by subclassing the connection checker and overriding `doDragChecks` :/ If you do you get this:

[Screen recording 2023-03-13 1.33.57 PM.webm](https://user-images.githubusercontent.com/25440652/224826433-d7b1e5d6-fc43-485f-84d3-bf5a5390b248.webm)

Scratch also overrides some logic in `connect` (which isn't possible in core blockly) to support inserting c-blocks. https://github.com/LLK/scratch-blocks/blob/66211340bc237a8fdc9017c8f2611768cb2fa3b4/core/connection.js#L150

I'm thinking that given this, it's best to keep the simplified insertion marker code, and file an issue to support scratch-style insertion. How does that sound to you @rachel-fenichel ?

rachel-fenichel: Sounds good. Link back to this conversation in the new issue, please.

BeksOmega: Filed https://github.com/google/blockly/issues/6898#issue-1626057110

core/connection.ts
```diff
@@ -243,67 +243,63 @@ export class Connection implements IASTNodeLocationWithBlock {
...
+   * @param setParent Whether to set the parent of the disconnected block or
+   *     not, defaults to true.
+   */
+  disconnect(setParent = true) {
+    const {parentConnection, childConnection} =
+        this.getParentAndChildConnections();
+    if (!parentConnection || !childConnection) {
       throw Error('Source connection not connected.');
     }
-    if (otherConnection.targetConnection !== this) {
```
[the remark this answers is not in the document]
BeksOmega: It seems a bit silly when we have the `connectReciprocally` method. If this check errors, afaict it just means we coded a bug (or someone messed with `targetConnection` directly, which is also our failing for making that public). Imo, these assertions should be covered by unit tests, not code. But I may be missing some context here.

core/connection.ts
```diff
@@ -243,67 +243,63 @@ export class Connection implements IASTNodeLocationWithBlock {
...
+
+    if (!childConnection.getSourceBlock().isShadow()) {
+      // If we were disconnecting a shadow, no need to spawn a new one.
+      parentConnection.respawnShadow_();
+    }
+
+    if (!eventGroup) eventUtils.setGroup(false);
+  }
+
+  /**
```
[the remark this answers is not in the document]
BeksOmega: Can do! Just wondering, do you want params and description to be generally required? Because we recently changed the lint config so we no longer require that with typescript.

core/connection.ts
```diff
@@ -243,67 +243,63 @@ export class Connection implements IASTNodeLocationWithBlock {
...
   }
 
-  /** Disconnect this connection. */
-  disconnect() {
-    const otherConnection = this.targetConnection;
-    if (!otherConnection) {
+  /**
+   * Disconnect this connection.
+   *
+   * @param setParent Whether to set the parent of the disconnected block or
```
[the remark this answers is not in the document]
BeksOmega: Added `protected disconnectInternal` so that the `setParent` param has been removed from the public API. Added docs to that new method.

rachel-fenichel (approved):

core/insertion_marker_manager.ts
cpcallen, line 1: `ConnectionType` is now unused, and you missed the lint warning (but now we can all enjoy it! 🤣)

PR #6877, opened 2023-03-03
fix: export Field-related types from Blockly

### LINKED ISSUE #6870: Re-export Field types from Blockly
### Check for duplicates

### Description

Certain properties were added to aid in typing fields correctly, but were not re-exported from Blockly, so they are not accessible outside of core. Known issues:

- FieldConfig in #6317 
- FieldValidator, FieldDropdownValidator, FieldInputValidator, FieldCheckboxValidator in #6603
    -  Also FieldColourValidator but not sure we need that given the move to a plugin

These properties must be exported from `blockly.ts` in order to be accessible outside of core. 

There was some debate about exporting them as top level properties (`Blockly.FieldConfig`) vs creating a new namespace to hold them. I think we should avoid creating new namespaces. I am ambivalent about creating a new module that exports these properties (similar to the utils module). It would be nice if we had better organization under the Blockly module, but it's confusing when the module structure doesn't match our directory structure, so the way you use these objects depends on if you are in or outside of core. For that reason, and given the current structure of fields, I would prefer to either move these helper type declarations to a new module (called `fields`?) and export them there, or just add them to the top level of Blockly.

Bad:
- Creating new TS namespaces (not recommended to mix namespaces and modules) => `Blockly.Field.FieldConfig`
- Declaring types in field.ts but exporting them from a different module that isn't blockly.ts (confusing)

Fine:
- exporting them from blockly.ts (bloats the blockly module) => `Blockly.FieldConfig`
- creating a new module, moving both the declarations and the export there, export this new module from `blockly.ts` => `Blockly.fields.FieldConfig`

@cpcallen any additional thoughts? 

### Reproduction steps

1. Try to use `Blockly.FieldConfig` from a plugin

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

_No response_

btw17:
## The basics

- [x] I branched from develop

### Resolves

Re-export Field types from Blockly
fixes #6870

### Proposed Changes

- Adds `FieldDropdownConfig` export
- Imports then re-exports `Field{Name}Config` and `Field{Name}FromJsonConfig` types from `./core/blockly.ts`

#### Behavior Before Change

- Config exports were not re-exported in `./core/blockly.ts`

#### Behavior After Change

- Config exports are re-exported in `./core/blockly.ts`

### Reason for Changes

- To make the included types available in the root export for Blockly

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

- `FieldDropdownConfig` is just a re-export of `FieldConfig`. Should we include this?
  - We currently have `FieldTextInputConfig` as a re-export of `FieldInputConfig` already, though the base `FieldInputConfig` isn't actually exported since `FieldInput` is only used internally
- **Note:** As mentioned above, the base `FieldInput` types aren't exported. It is not a registered field since it was created for internal use when splitting the old `FieldInput` to `FieldTextInput`. I don't think these types should be exported, though it's worth noting that this is deliberate.
- In `./core/blockly.ts`, think it's worthwhile to export `UnattachedFieldError` or `FieldProto`?
  - `FieldProto` isn't necessarily needed, though it could theoretically be used by and edge-case subclass that makes significant changes upon the base Field class

### DIFF
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -52,19 +52,19 @@ import {DragTarget} from './drag_target.js';
 import * as dropDownDiv from './dropdowndiv.js';
 import * as Events from './events/events.js';
 import * as Extensions from './extensions.js';
-import {Field, FieldValidator} from './field.js';
-import {FieldAngle, FieldAngleValidator} from './field_angle.js';
-import {FieldCheckbox, FieldCheckboxValidator} from './field_checkbox.js';
-import {FieldColour, FieldColourValidator} from './field_colour.js';
-import {FieldDropdown, FieldDropdownValidator, MenuGenerator, MenuGeneratorFunction, MenuOption} from './field_dropdown.js';
-import {FieldImage} from './field_image.js';
-import {FieldLabel} from './field_label.js';
+import {Field, FieldConfig, FieldValidator} from './field.js';
+import {FieldAngle, FieldAngleConfig, FieldAngleFromJsonConfig, FieldAngleValidator} from './field_angle.js';
+import {FieldCheckbox, FieldCheckboxConfig, FieldCheckboxFromJsonConfig, FieldCheckboxValidator} from './field_checkbox.js';
+import {FieldColour, FieldColourConfig, FieldColourFromJsonConfig, FieldColourValidator} from './field_colour.js';
+import {FieldDropdown, FieldDropdownConfig, FieldDropdownFromJsonConfig, FieldDropdownValidator, MenuGenerator, MenuGeneratorFunction, MenuOption} from './field_dropdown.js';
+import {FieldImage, FieldImageConfig, FieldImageFromJsonConfig} from './field_image.js';
+import {FieldLabel, FieldLabelConfig, FieldLabelFromJsonConfig} from './field_label.js';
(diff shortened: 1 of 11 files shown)


maribethb:
Changes so far lgtm, regarding the questions (which can be addressed in this PR if you like)

- `FieldInput` and its related objects should all be marked `@internal` if they are intentionally not exported. This removes them from the documentation.
- I think `FieldProto` should also be marked `@internal` and not exported, based on how we're currently using it. If needed someone can file a fr and ask us to change this in the future
- Yes, let's export `UnattachedFieldError`

btw17:
Should be all set @maribethb! I did as you suggested with a few other notes:
- I moved `Validator` types to the bottom of their files so match the convention with the `Config` files
- I removed a redundant comment line for the validator in `core/field.ts`
- I copied the validator function comment to each redeclared type
- I removed the export for `InputTypes` in `./core/field_inputs.ts` since it's not exported in `core/blockly.ts` and not used in any other files

maribethb (approved):

maribethb:
Note: Changed the PR title, because I think this should be reflected in release notes (chores are excluded)

Also updated PR comment to close the linked issue.

PR #6886, opened 2023-03-07
chore: Reduce delta on ports to blockly-samples

NeilFraser:
* Reduce usage of obsolete .keyCode property.
* Rename private properties/methods which violate eslint rules.
* Use arrays of bound events rather than individual properties.
* Improve typing info.

No user or developer visible changes.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,7 +16,7 @@
 ### Resolves
 
 <!-- TODO: What Github issue does this resolve? Please include a link. -->
-Fixes
+Fixes 
 
 ### Proposed Changes
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,7 +16,7 @@
 ### Resolves
 
 <!-- TODO: What Github issue does this resolve? Please include a link. -->
-Fixes 
+Fixes
 
 ### Proposed Changes
 
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -165,14 +165,6 @@ import {resizeSvgContents as realResizeSvgContents, WorkspaceSvg} from './worksp
 import * as Xml from './xml.js';
 import {ZoomControls} from './zoom_controls.js';
 
-declare global {
-  // By default TypeScript requires attributes to be strings.
-  // Add the ability to set a numeric attribute to an Element.
-  interface Element {
-    setAttribute(name: string, value: number): void;
-  }
-}
-
 /**
  * Blockly core version.
  * This constant is overridden by the build script (npm run build) to the value
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -165,6 +165,13 @@ import {resizeSvgContents as realResizeSvgContents, WorkspaceSvg} from './worksp
 import * as Xml from './xml.js';
 import {ZoomControls} from './zoom_controls.js';
 
+declare global {
+  // By default TypeScript requires attributes to be strings.
+  // Add the ability to set a numeric attribute to an Element.
+  interface Element {
+    setAttribute(name: string, value: number): void;
+  }
+}
 
 /**
  * Blockly core version.
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -165,6 +165,7 @@ import {resizeSvgContents as realResizeSvgContents, WorkspaceSvg} from './worksp
 import * as Xml from './xml.js';
 import {ZoomControls} from './zoom_controls.js';
 
+
 /**
  * Blockly core version.
  * This constant is overridden by the build script (npm run build) to the value
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -275,6 +275,7 @@ export class Comment extends Icon {
     for (const event of this.boundEvents) {
       browserEvents.unbind(event);
     }
+    this.boundEvents.length = 0;
     if (this.bubble_) {
       this.bubble_.dispose();
       this.bubble_ = null;
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -42,17 +42,12 @@ export class Comment extends Icon {
    */
   private cachedText: string|null = '';
 
-  /** Mouse up event data. */
-  private onMouseUpWrapper: browserEvents.Data|null = null;
-
-  /** Wheel event data. */
-  private onWheelWrapper: browserEvents.Data|null = null;
-
-  /** Change event data. */
-  private onChangeWrapper: browserEvents.Data|null = null;
-
-  /** Input event data. */
-  private onInputWrapper: browserEvents.Data|null = null;
+  /**
+   * Array holding info needed to unbind events.
+   * Used for disposing.
+   * Ex: [[node, name, func], [node, name, func]].
+   */
+  private boundEvents: browserEvents.Data[] = [];
 
   /**
    * The SVG element that contains the text edit area, or null if not created.
@@ -147,14 +142,14 @@ export class Comment extends Icon {
     body.appendChild(textarea);
     this.foreignObject!.appendChild(body);
 
-    this.onMouseUpWrapper = browserEvents.conditionalBind(
-        textarea, 'focus', this, this.startEdit, true);
+    this.boundEvents.push(browserEvents.conditionalBind(
+        textarea, 'focus', this, this.startEdit, true));
     // Don't zoom with mousewheel.
-    this.onWheelWrapper = browserEvents.conditionalBind(
+    this.boundEvents.push(browserEvents.conditionalBind(
         textarea, 'wheel', this, function(e: Event) {
           e.stopPropagation();
-        });
-    this.onChangeWrapper = browserEvents.conditionalBind(
+        }));
+    this.boundEvents.push(browserEvents.conditionalBind(
         textarea, 'change', this,
         /**
          * @param _e Unused event parameter.
@@ -165,15 +160,15 @@ export class Comment extends Icon {
                 this.getBlock(), 'comment', null, this.cachedText,
                 this.model.text));
           }
-        });
-    this.onInputWrapper = browserEvents.conditionalBind(
+        }));
+    this.boundEvents.push(browserEvents.conditionalBind(
         textarea, 'input', this,
         /**
          * @param _e Unused event parameter.
          */
         function(this: Comment, _e: Event) {
           this.model.text = textarea.value;
-        });
+        }));
 
     setTimeout(textarea.focus.bind(textarea), 0);
 
@@ -277,21 +272,8 @@ export class Comment extends Icon {
    * Dispose of the bubble.
    */
   private disposeBubble() {
-    if (this.onMouseUpWrapper) {
-      browserEvents.unbind(this.onMouseUpWrapper);
-      this.onMouseUpWrapper = null;
-    }
-    if (this.onWheelWrapper) {
-      browserEvents.unbind(this.onWheelWrapper);
-      this.onWheelWrapper = null;
-    }
-    if (this.onChangeWrapper) {
-      browserEvents.unbind(this.onChangeWrapper);
-      this.onChangeWrapper = null;
-    }
-    if (this.onInputWrapper) {
-      browserEvents.unbind(this.onInputWrapper);
-      this.onInputWrapper = null;
+    for (const event of this.boundEvents) {
+      browserEvents.unbind(event);
     }
     if (this.bubble_) {
       this.bubble_.dispose();
--- a/core/comment.ts
+++ b/core/comment.ts
@@ -210,8 +210,8 @@ export class Comment extends Icon {
     const doubleBorderWidth = 2 * Bubble.BORDER_WIDTH;
     const widthMinusBorder = size.width - doubleBorderWidth;
     const heightMinusBorder = size.height - doubleBorderWidth;
-    this.foreignObject.setAttribute('width', widthMinusBorder);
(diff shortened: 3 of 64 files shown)


core/field_angle.ts
```diff
@@ -69,49 +68,38 @@ export class FieldAngle extends FieldInput<number> {
...
+  /** The angle picker's gauge path depending on the value. */
+  gauge: SVGPathElement|null = null;
 
-  /**
-   * Serializable fields are saved by the serializer, non-serializable fields
-   * are not. Editable fields should also be serializable.
-   */
-  override SERIALIZABLE = true;
+  /** The degree symbol for this field. */
+  protected symbol: SVGTSpanElement|null = null;
```
BeksOmega, line 102: Protected properties are part of the public API :/ so the underscore should be re-added for now.

core/field_angle.ts
```diff
@@ -69,49 +68,38 @@ export class FieldAngle extends FieldInput<number> {
...
-  /** Surface mouse move event data. */
-  private moveSurfaceWrapper_: browserEvents.Data|null = null;
+  /** The angle picker's gauge path depending on the value. */
+  gauge: SVGPathElement|null = null;
 
-  /**
-   * Serializable fields are saved by the serializer, non-serializable fields
-   * are not. Editable fields should also be serializable.
-   */
-  override SERIALIZABLE = true;
```
BeksOmega, line 114: I think this removal should be reverted, but I might be missing something.

core/field_angle.ts
```diff
@@ -211,11 +199,15 @@ export class FieldAngle extends FieldInput<number> {
...
   }
 
-  /** Create the angle dropdown editor. */
-  private dropdownCreate_() {
+  /**
+   * Creates the angle dropdown editor.
+   *
+   * @returns The newly created slider.
+   */
+  private dropdownCreate() {
```
BeksOmega, line 210: Can you add a return type to this function now that it returns?

core/field_angle.ts
```diff
@@ -261,34 +253,28 @@ export class FieldAngle extends FieldInput<number> {
...
+        browserEvents.conditionalBind(
+            circle, 'pointerdown', this, this.onMouseMove, true));
+    this.boundEvents.push(
+        browserEvents.conditionalBind(
+            circle, 'pointermove', this, this.onMouseMove, true));
+    return svg;
   }
 
   /** Disposes of events and DOM-references belonging to the angle editor. */
   private dropdownDispose_() {
```
BeksOmega, line 269: Remove underscore.

core/field_angle.ts
```diff
@@ -302,9 +288,9 @@ export class FieldAngle extends FieldInput<number> {
    *
    * @param e Mouse move event.
    */
-  protected onMouseMove_(e: PointerEvent) {
```
BeksOmega, line 305: Protected, please revert.

core/field_colour.ts
```diff
@@ -113,46 +104,46 @@ export class FieldColour extends Field<string> {
...
-   *     string ('#rrggbb' format), or null to abort the change.Blockly.
-   * @param opt_config A map of options used to configure the field.
+   *     string ('#rrggbb' format), or null to abort the change.
+   * @param config A map of options used to configure the field.
    *     See the [field creation documentation]{@link
    * https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/colour}
    * for a list of properties this parameter supports.
    */
   constructor(
-      opt_value?: string|Sentinel, opt_validator?: FieldColourValidator,
```
BeksOmega, line 145: Can you make this update across all fields, or revert?

core/field_colour.ts
```diff
@@ -399,22 +394,22 @@ export class FieldColour extends Field<string> {
   private onMouseMove_(e: PointerEvent) {
```
BeksOmega: Remove underscore. Here and on all other `private` methods throughout this file..

core/field_multilineinput.ts
```diff
@@ -33,46 +32,44 @@ export class FieldMultilineInput extends FieldTextInput {
...
-  // AnyDuringMigration because:  Type 'null' is not assignable to type
-  // 'SVGGElement'.
-  textGroup_: SVGGElement = null as AnyDuringMigration;
+  textGroup: SVGGElement|null = null;
 
   /**
    * Defines the maximum number of lines of field.
    * If exceeded, scrolling functionality is enabled.
    */
-  protected maxLines_ = Infinity;
```
BeksOmega, line 44: Protected, please revert, here and on `isOverflowedY`.

BeksOmega (changes requested):
Changes generally look good to me! But I'm a bit confused about the intention of this PR. I would have either expected these things to be fixed across all of the fields in core, or just fixed after the fields were moved into samples.

NeilFraser:
> I'm a bit confused about the intention of this PR. I would have either expected these things to be fixed across all of the fields in core, or just fixed after the fields were moved into samples.

Blockly is a moving target, my attempts to move angle/colour/multiline to samples have been repeatedly disrupted by changes made in core.  So to simplify the move I'm back-porting as many changes as possible to core, that way the move becomes easier.

BeksOmega:
Resolved all the comments that are fixed. Just a few left! And they mostly discussion-y.

core/flyout_base.ts
```diff
@@ -799,7 +799,7 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
       blockInfo: toolbox.BlockInfo, gaps: number[], defaultGap: number) {
     let gap;
     if (blockInfo['gap']) {
-      gap = parseInt(blockInfo['gap'].toString());
+      gap = parseInt(String(blockInfo['gap']));
```
BeksOmega, line 802: d25dca438cfc02c163799f4f9e2cf3c9c718ce3c uses the string constructor instead of template litterals, but I'm not sure why. Are they semantically different?

core/comment.ts
```diff
@@ -277,21 +272,8 @@ export class Comment extends Icon {
...
-    }
-    if (this.onChangeWrapper) {
-      browserEvents.unbind(this.onChangeWrapper);
-      this.onChangeWrapper = null;
-    }
-    if (this.onInputWrapper) {
-      browserEvents.unbind(this.onInputWrapper);
-      this.onInputWrapper = null;
+    for (const event of this.boundEvents) {
+      browserEvents.unbind(event);
```
BeksOmega, line 276: I'm not sure we actually need to unbind these. I learned recently that listeners should be garbage collected when the element the listener is associated with is.

Similarly for the rest of commit e965089c28da5caf850b5f17cac89c2ca6f8b13b

core/field_multilineinput.ts
```diff
@@ -73,9 +73,14 @@ export class FieldMultilineInput extends FieldTextInput {
...
   }
 
+  /**
+   * Configure the field based on the given map of options.
+   *
+   * @param config A map of options to configure the field based on.
+   */
   protected override configure_(config: FieldMultilineInputConfig) {
     super.configure_(config);
-    if (config.maxLines_) this.setMaxLines(config.maxLines_);
```
BeksOmega: Glad you caught this!

core/field_colour.ts
```diff
@@ -176,6 +176,11 @@ export class FieldColour extends Field<string> {
     }
   }
 
+  /**
+   * Updates text field to match the colour/style of the block.
+   *
+   * @internal
```
BeksOmega, line 182: If this is overridable by external devs, it can't be tagged`@internal` because you can't override `@internal` things :/ This is a semantic difference from closure `@package` tags (which this almost certainly had before).

BeksOmega (changes requested):
Thanks for the continued work on this!

Just as a heads up, commits that contain orthogonal changes make it a lot more difficult for to catch things (e.g. renaming problems with `maxLines_` which I totally missed). So it would be really helpful for me if each commit contained just one type of change =)

core/comment.ts
NeilFraser: Historically the HTML world and the JavaScript world had separate garbage collectors which didn't talk to each other.  This meant if an element on one side linked to an element on the other side, the garbage collectors were powerless to clean up.  Thus one had to sever the links manually upon disposal.

Whether this is still true today is unknown to me.  Last month I ran a series of tests on this, but Chrome leaks memory so badly that it was impossible to determine how its garbage collector works.  And of course, we'd need to verify that _all_ supported browsers have upgraded garbage collectors before we could strip down our dispose functions.

core/field_colour.ts
NeilFraser: The function this overrides (`Field.applyColour`) is @internal.  Is that wrong too?  There are a lot of those in that file.

core/flyout_base.ts
NeilFraser: No difference.  We had a lovely diverse mix of:
1. `String(foo)`
2. `foo.toString()`
3. `'{$foo}'`
4. `foo + ''`

I really don't have any preferences other than getting rid of `foo.toString()` since that throws if foo is null or undefined.  This PR generally uses `'{$foo}'` for simple variables, and `String(foo.bar)` for anything more complicated (I find `'{$foo["bar"].toUpperCase()}'` to be harder to read).

If we want to standardize on something, that's the team's choice.  I just want to get rid of `.toString()` here.

core/field_variable.ts
```diff
@@ -416,14 +416,12 @@ export class FieldVariable extends FieldDropdown {
    * @param defaultType The type of the variable to create if this field's
    *     value is not explicitly set.  Defaults to ''.
    */
-  private setTypes_(variableTypes?: string[]|null, defaultType = '') {
+  private setTypes_(variableTypes: string[]|null = null, defaultType = '') {
```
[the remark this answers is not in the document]
BeksOmega: ahh ok I missed that you added the default parameter value. Sweetness.

core/comment.ts
BeksOmega: I tested this locally when I was working https://github.com/google/blockly/issues/6890. I ran the following steps, both with and without unbinding events:

1) Start a performance recording (with memory recording enabled).
2) Trigger the spaghetti test on the playground.
3) Force a GC.
4) Delete the top block in the spaghetti stack.
5) Force a GC.
6) Stop the performance recording.

This resulted in the same number of listeners after the final GC (both with and without unbinding).

All browsers ship with mark-and-sweep, so I can't imagine this is different cross browser. All of my searching also states that this is generally applicable (excepting IE ofc).

However, since this PR is so large and we want to get this merged before Friday. I think we can just leave as is and I can look into it as part of #6890

core/field_colour.ts
BeksOmega: I filed a separate issue for dealing with this, since it's a widespread problem: #6895

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,7 +16,7 @@
 ### Resolves
 
 <!-- TODO: What Github issue does this resolve? Please include a link. -->
-Fixes 
+Fixes
```
maribethb, line 19: The space is actually part of the template here, because you can just click and start typing the issue number.

core/blockly.ts
```diff
@@ -165,6 +165,13 @@ import {resizeSvgContents as realResizeSvgContents, WorkspaceSvg} from './worksp
 import * as Xml from './xml.js';
 import {ZoomControls} from './zoom_controls.js';
 
+declare global {
```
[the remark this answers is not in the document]
maribethb: I don't really think we should do this, though I could probably be convinced if we had good reasons.

I think declaring this as a new global would also affect anyone who pulls in our types, which may be undesirable for them. We had already casted everything to string, which is more explicit and reduces the possibility of surprises (such as mistakenly trying to set an attribute to `null` only to have it set to `"null"`). So I don't see a compelling reason to add this.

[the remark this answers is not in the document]
NeilFraser: This change strengthens our typing and does the opposite of what you describe.

1. This only opens the ability to set attributes to numbers (which comprise the majority of our calls to setAttribute).
2. This *reduces* the number of times we have to cast a value to a string, which means that if a 'number' is actually null or undefined, it will fail.
3. Casting everything to a string is much less explicit, since everything can be cast to a string.  This change removes massive numbers of unchecked, blind casts.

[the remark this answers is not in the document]
maribethb: What I meant by explicit is that anything passed to `setAttribute` will be implicitly cast to a string by the browser. Having to cast it to a string ourselves first is the explicit action we take. You don't have to know that the value you pass in will be cast to a string by the browser, because it's right there in our code. So I disagree with point 3, because the blind casts are still there, just hidden a layer deeper.

However, you're right that removing the casts in favor of checking for numbers (and possibly booleans) still provides us type-safety against unexpected `null` or `undefined` values which may otherwise just be cast to a string by our indiscriminate casting. 

I still don't really like that we would also be essentially opting in all of our downstream users into this change, and I don't know of a way around that.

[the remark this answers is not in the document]
cpcallen: > I still don't really like that we would also be essentially opting in all of our downstream users into this change, and I don't know of a way around that.

I agree that this `declare global` is very undesirable for exactly that reason.

What about providing (and consistently using) a helper `function setAttribute(Element, name: string, value: number)`—perhaps in `utils/dom.ts`?

.github/PULL_REQUEST_TEMPLATE.md
NeilFraser: What's hilarious is that I can't fix this in my IDE.  It strips all trailing spaces.  In the end I had to add an 'X' after 'Fixes', then use a hex editor to change the 'X' to a space.  This file is radioactive.

BeksOmega (approved):
LGTM!

PR #6894, opened 2023-03-14
fix: dispose performance

### LINKED ISSUE #6890: Improve performance of deleting blocks
### Check for duplicates

### Description

Deleting the top block of the spaghetti code stack on master (https://github.com/google/blockly/commit/4125fd31534dd9f450219259b9ccb3c1495605c9) takes ~1670ms, and `checkAndDelete` takes ~1000ms.

Screenshot 2023-03-09 12 06 55 PM

On  develop (https://github.com/google/blockly/commit/c0934216f8ece4261f21977bc89ea92e8483d468) it takes ~720 seconds, with `checkAndDelete` taking ~360ms.

### Reproduction steps

1. Go to the playground
2. Hit the "spaghetti" or "JSO spaghetti" test buttons.
3. Go to the performance tab of the chrome devtools.
4. Start a recording
5. Delete the top block in the spaghetti stack.
6. Stop the recording.

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

_No response_

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #6890 

### Proposed Changes

Improves performance of deleting blocks.

See #6890 for previous performance. The new performance is ~300ms total, and ~100ms for `checkAndDelete`.

### Reason for Changes

Speeeed

### Test Coverage

Manual testing + fixing failing tests.

Added one test to check that blocks do not receive their own delete events.

### Documentation

N/A

### Additional Information

We save an extra 20ms after #6876 is merged.

Best to review this commit-wise since there is some code reorganization.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -316,7 +316,10 @@ export class Block implements IASTNodeLocation, IDeletable {
     if (this.isDeadOrDying()) return;
 
     this.unplug(healStack);
-    eventUtils.fire(new (eventUtils.get(eventUtils.BLOCK_DELETE))(this));
+    if (Blockly.Events.isEnabled()) {
+      // Constructing the delete event is costly. Only perform if necessary.
+      eventUtils.fire(new (eventUtils.get(eventUtils.BLOCK_DELETE))(this));
+    }
     this.workspace.removeTopBlock(this);
     this.disposeInternal();
   }
--- a/core/block.ts
+++ b/core/block.ts
@@ -336,7 +336,7 @@ export class Block implements IASTNodeLocation, IDeletable {
       this.workspace.removeTypedBlock(this);
       this.workspace.removeBlockById(this.id);
       this.disposing = true;
-  
+
       [...this.childBlocks_].forEach((c) => c.disposeInternal());
       this.inputList.forEach((i) => i.dispose());
       this.inputList.length = 0;
--- a/core/block.ts
+++ b/core/block.ts
@@ -313,10 +313,8 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @suppress {checkTypes}
    */
   dispose(healStack: boolean) {
-    if (this.isDeadOrDying()) {
-      return;
-    }
-    // Terminate onchange event calls.
+    if (this.isDeadOrDying()) return;
+
     if (this.onchangeWrapper_) {
       this.workspace.removeChangeListener(this.onchangeWrapper_);
     }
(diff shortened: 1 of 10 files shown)


core/block.ts
```diff
@@ -313,44 +313,34 @@ export class Block implements IASTNodeLocation, IDeletable {
...
     if (eventUtils.isEnabled()) {
+      // Constructing the delete event is costly. Only perform if necessary.
       eventUtils.fire(new (eventUtils.get(eventUtils.BLOCK_DELETE))(this));
     }
-    eventUtils.disable();
+    this.workspace.removeTopBlock(this);
+    this.disposeInternal();
+  }
+
+  protected disposeInternal() {
```
rachel-fenichel, line 327: Add comment (standard documentation, + explaining what goes in `dispose` and what goes in `disposeInternal`).

core/block.ts
```diff
@@ -313,44 +313,34 @@ export class Block implements IASTNodeLocation, IDeletable {
...
+    if (this.isDeadOrDying()) return;
+
+    if (this.onchangeWrapper_) {
+      this.workspace.removeChangeListener(this.onchangeWrapper_);
```
rachel-fenichel, line 331: Will the existing change listener trigger on the call to `unplug` on line 318?

core/icon.ts
```diff
@@ -79,8 +79,11 @@ export abstract class Icon {
 
   /** Dispose of this icon. */
   dispose() {
-    dom.removeNode(this.iconGroup_);  // Dispose of and unlink the icon.
-    this.setVisible(false);           // Dispose of and unlink the bubble.
+    if (!this.getBlock().isDeadOrDying()) {
+      dom.removeNode(this.iconGroup_);
+    }
+    // TODO: Check if this is necessary.
```
rachel-fenichel, line 85: Nit: associate todo with an issue

core/field.ts
```diff
@@ -489,14 +489,11 @@ export abstract class Field<T = any> implements IASTNodeLocationSvg,
   dispose() {
     dropDownDiv.hideIfOwner(this);
     WidgetDiv.hideIfOwner(this);
-    Tooltip.unbindMouseEvents(this.getClickTarget_());
 
-    if (this.mouseDownWrapper_) {
-      browserEvents.unbind(this.mouseDownWrapper_);
+    if (!this.getSourceBlock()?.isDeadOrDying()) {
```
rachel-fenichel, line 493: Sweet, I like this optimization.

core/block_animations.ts
```diff
@@ -39,6 +39,9 @@ let wobblingBlock: BlockSvg|null = null;
  * @internal
  */
 export function disposeUiEffect(block: BlockSvg) {
+  // Disposing is going to take so long the animation won't playanyway.
```
rachel-fenichel, line 42: nit: space between "play" and "anyway"

core/block.ts
```diff
@@ -313,44 +313,34 @@ export class Block implements IASTNodeLocation, IDeletable {
...
-      // First, dispose of all my children.
-      for (let i = this.childBlocks_.length - 1; i >= 0; i--) {
-        this.childBlocks_[i].dispose(false);
-      }
-      // Then dispose of myself.
-      // Dispose of all inputs and their fields.
-      for (let i = 0, input; input = this.inputList[i]; i++) {
-        input.dispose();
-      }
+      [...this.childBlocks_].forEach((c) => c.disposeInternal());
```
rachel-fenichel, line 340: This used to use the original array and go backwards. Now it uses a copy and goes forward. Why the change?

rachel-fenichel (commented):
Overall looks good. I have some questions about how certain changes preserve old behaviour, but I'll try to find you to discuss in-person.

core/block.ts
BeksOmega: Added the removal to `dispose` as well, with explanatory comment.

core/block.ts
BeksOmega: So we discussed yesterday, and I said the issue was that `setParent(null)` updates the `childBlocks_` of the parent block. So calling `dispose` on children would mutate the array we were looping over. We could either deal with this by looping backwards (as previous) or by looping over a copy that doesn't get mutated.

However! I checked it again after we chatted and since we're no longer calling `unplug` on every block, we're also not calling `setParent(null)` on every block. So the array isn't being mutated at all, and we can just loop over it directly =)

core/icon.ts
BeksOmega: It is necessary because the bubble lives on the bubble canvas, and is not an SVG element of the block. It will not get automatically garbage collected, so we need to dispose of it separately.

rachel-fenichel (approved):

PR #6896, opened 2023-03-14
fix: toString performance

### LINKED ISSUE #4229: Collapsed block performance

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #4229 

### Proposed Changes

Increases peformance of `toString` 50x. Used to take ~50ms, now takes <1ms.

### Reason for Changes

Speeeeed.

### Test Coverage

Just manual testing.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -1400,91 +1400,15 @@ export class Block implements IASTNodeLocation, IDeletable {
    * Create a human-readable text representation of this block and any children.
    *
    * @param opt_maxLength Truncate the string to this length.
-   * @param opt_emptyToken The placeholder string used to denote an empty field.
+   * @param opt_emptyToken The placeholder string used to denote an empty input.
    *     If not specified, '?' is used.
    * @returns Text of block.
    */
   toString(opt_maxLength?: number, opt_emptyToken?: string): string {
-    const tokens = [];
-    const emptyFieldPlaceholder = opt_emptyToken || '?';
+    const tokens = this.toTokens(opt_emptyToken);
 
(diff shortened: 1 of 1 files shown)


core/block.ts
```diff
@@ -1400,21 +1399,51 @@ export class Block implements IASTNodeLocation, IDeletable {
...
+    for (let i = 2; i < tokens.length; i++) {
+      if (tokens[i - 2] === '(' && tokens[i] === ')') {
+        tokens[i - 2] = tokens[i - 1];
+        tokens.splice(i - 1, 2);
+      }
+    }
 
-    let node = ASTNode.createBlockNode(this);
-    const rootNode = node;
+    let prev = '';
```
maribethb, line 1419: Can you add the comment too? 

// Join the text array, removing the spaces around added parentheses.

maribethb (approved):

PR #6900, opened 2023-03-16
refactor(blocks): Migrate `blocks/math.js` to TypeScript

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6828.

### Additional Information

The types are not completely finalised, in particular because `BlockDefinition` is defined as `AnyDuringMigration`, and so it has been necessary in a few places to use `.!` and/or `as AnyDuringMigration`.  Hopefully it will be possible to address these in a future PR.

### DIFF
--- a/blocks/math.ts
+++ b/blocks/math.ts
@@ -494,10 +494,8 @@ const IS_DIVISIBLEBY_MUTATOR_MIXIN: BlockDefinition = {
  */
 const IS_DIVISIBLE_MUTATOR_EXTENSION = function(this: Block) {
   this.getField('PROPERTY')!.setValidator(
-      /**
-       * @param option The selected dropdown option.
-       */
-      function(this: FieldDropdown, option) {
+      /** @param option The selected dropdown option. */
+      function(this: FieldDropdown, option: string) {
         const divisorInput = (option === 'DIVISIBLE_BY');
         (this.getSourceBlock() as AnyDuringMigration)
             .updateShape_(divisorInput);
@@ -525,9 +523,9 @@ const LIST_MODES_MUTATOR_MIXIN: BlockDefinition = {
    */
   updateType_: function(this: Block, newOp: string) {
     if (newOp === 'MODE') {
-      this.outputConnection!.setCheck('Array');
+      this.outputConnection?.setCheck('Array');
     } else {
-      this.outputConnection!.setCheck('Number');
+      this.outputConnection?.setCheck('Number');
     }
   },
   /**
--- a/blocks/math.ts
+++ b/blocks/math.ts
@@ -493,16 +493,15 @@ const IS_DIVISIBLEBY_MUTATOR_MIXIN: BlockDefinition = {
  * property is "divisible by".
  */
 const IS_DIVISIBLE_MUTATOR_EXTENSION = function(this: Block) {
-  this.getField('PROPERTY')!.
-      setValidator(
-          /**
-           * @param option The selected dropdown option.
-           */
-          function(this: FieldDropdown, option) {
-            const divisorInput = (option === 'DIVISIBLE_BY');
-            (this.getSourceBlock() as AnyDuringMigration)
-                .updateShape_(divisorInput);
-          });
+  this.getField('PROPERTY')!.setValidator(
+      /**
+       * @param option The selected dropdown option.
+       */
+      function(this: FieldDropdown, option) {
+        const divisorInput = (option === 'DIVISIBLE_BY');
+        (this.getSourceBlock() as AnyDuringMigration)
+            .updateShape_(divisorInput);
+      });
 };
 
 Extensions.registerMutator(
--- a/blocks/math.ts
+++ b/blocks/math.ts
@@ -500,6 +500,7 @@ const IS_DIVISIBLE_MUTATOR_EXTENSION = function(this: DivisiblebyBlock) {
       function(this: FieldDropdown, option: string) {
         const divisorInput = (option === 'DIVISIBLE_BY');
         (this.getSourceBlock() as DivisiblebyBlock).updateShape_(divisorInput);
+        return undefined;  // FieldValidators can't be void.  Use option as-is.
       });
 };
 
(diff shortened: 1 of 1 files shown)


blocks/math.ts
```diff
@@ -500,20 +491,17 @@ const IS_DIVISIBLEBY_MUTATOR_MIXIN = {
...
-          function(option) {
-            const divisorInput = (option === 'DIVISIBLE_BY');
-            this.getSourceBlock().updateShape_(divisorInput);
-          });
+const IS_DIVISIBLE_MUTATOR_EXTENSION = function(this: Block) {
+  this.getField('PROPERTY')!.setValidator(
+      /**
+       * @param option The selected dropdown option.
+       */
+      function(this: FieldDropdown, option) {
```
gonfunko, line 500: Add a type for `option`?

blocks/math.ts
```diff
@@ -528,44 +516,39 @@ Extensions.register(
...
-   * @private
-   * @this {Block}
+   *
+   * @param newOp Either 'MODE' or some op than returns a number.
    */
-  updateType_: function(newOp) {
+  updateType_: function(this: Block, newOp: string) {
     if (newOp === 'MODE') {
-      this.outputConnection.setCheck('Array');
+      this.outputConnection!.setCheck('Array');
```
gonfunko, line 528: The styleguide doesn't quite forbid non-null assertions but does heavily discourage them. We probably should determine our policy here; my personal inclination in a situation like this is to use ? instead, although that admittedly can result in silent failures.

blocks/math.ts
```diff
@@ -577,13 +560,11 @@ const LIST_MODES_MUTATOR_MIXIN = {
...
  * modes operation (outputs a list of numbers).
- * @this {Block}
- * @package
  */
-const LIST_MODES_MUTATOR_EXTENSION = function() {
-  this.getField('OP').setValidator(function(newOp) {
-    this.updateType_(newOp);
-  }.bind(this));
+const LIST_MODES_MUTATOR_EXTENSION = function(this: Block) {
+  this.getField('OP')!.setValidator(function(this: Field, newOp: string) {
```
gonfunko, line 565: Could the validator be converted to an arrow function to avoid having to explicitly bind this/generally modernize it?

blocks/math.ts
cpcallen: Hmm.  It was type `{*}` before, so I guess `any`.  Or do you think `AnyDuringMigration` would be a better choice for now?

cpcallen: (I'm a little confused about where `option` comes from.  `.setValidator` takes a `FieldValidator`, which is just a one-argument function.…)

cpcallen: Ahh, no: `FieldDropdown extends Field<string>`, and `.setValidator` takes a `FieldValidator<T>`, which is a `(T) => T | null | undefined`.  So `option` here is a `string`.

blocks/math.ts
cpcallen: I say we can live with silent failures, and pray that whoever discovers this comment when debugging one in the future will forgive me…

blocks/math.ts
cpcallen: I don't think so, because the validator makes use of `this`, and if I understand correctly an arrow expression here would have `this` irrevocably bound to `LIST_MODES_MUTATOR_EXTENSION`, which is not what we want.

(If only block definitions were classes, and these could just be ordinary methods!)

cpcallen: (No, now that I think more carefully I think that's wrong: `this` would be `undefined`—still wrong.)

blocks/math.ts
cpcallen: Actually, I've changed my mind about this.  I think it best to avoid making any functional changes in this PR, but instead try to preserve existing behaviour exactly (and ideally have `tsc` generate `.js` that is identical to the original `.js` file once comments and whitespace are removed).

I agree that using `?.` would likely be preferable here, but I think such code-cleanup should be done in a separate PR.

blocks/math.ts
cpcallen: The more I look at this bit of code the weirder it is.  I think  I will rewrite it—but in a future PR.  For now just try to preserve existing behaviour exactly.

cpcallen:
I learned a bunch about how to more cleanly fix some type errors while working on `lists.ts`, and have applied those changes via two additional commits on this PR.

cpcallen:
One more update:

It turns out that the previous types for the mixins were completely wrong, for two reasons:

- They should be intersection types (which have the union of the  properties), not union types (which have the interseciton of the properties).
- The `*_MIXIN` objects were already declared as having type `BlockDefinition`, which is ultimately an alias for `any`.

TypeScript doesn't like (some) kinds of circularly defined types, so fixing the above necessitates declaring a few auxiliary types just to make the type checker happy - but the end result is excellent and caught an actual type error in the code.

blocks/math.ts
gonfunko: SGTM

blocks/math.ts
```diff
@@ -577,13 +563,13 @@ const LIST_MODES_MUTATOR_MIXIN = {
...
- * @this {Block}
- * @package
  */
-const LIST_MODES_MUTATOR_EXTENSION = function() {
-  this.getField('OP').setValidator(function(newOp) {
-    this.updateType_(newOp);
-  }.bind(this));
+const LIST_MODES_MUTATOR_EXTENSION = function(this: ListModesBlock) {
+  this.getField('OP')!.setValidator(
+      function(this: typeof LIST_MODES_MUTATOR_MIXIN|Block, newOp: string) {
```
gonfunko, line 569: Update to ListModesBlock?

blocks/math.ts
gonfunko: Can the type be added accordingly?

blocks/math.ts
cpcallen: Yes, I'd done that too!

gonfunko (approved):

cpcallen:
So, I forgot to push commit 8591f44 before asking you to look again.  Sorry.  Do feel free to have a look at again, because the declaration of the types of the mixins is much more correct but also much less pleasant.

PR #6902, opened 2023-03-17
refactor(blocks): Migrate `blocks/lists.js` to TypeScript

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6828.

### DIFF
--- a/blocks/lists.ts
+++ b/blocks/lists.ts
@@ -16,6 +16,8 @@ import * as fieldRegistry from '../core/field_registry.js';
 import * as xmlUtils from '../core/utils/xml.js';
 import {Align} from '../core/input.js';
 import type {Block} from '../core/block.js';
+import type {Connection} from '../core/connection.js';
+import type {BlockSvg} from '../core/block_svg.js';
 import type {BlockDefinition} from '../core/blocks.js';
 import {ConnectionType} from '../core/connection_type.js';
 import type {FieldDropdown} from '../core/field_dropdown.js';
@@ -114,9 +116,15 @@ export const blocks = createBlockDefinitionsFromJsonArray([
   },
 ]);
 
-type CreateWithBlock = Block|typeof blocks['lists_create_with'];
 
-blocks['lists_create_with'] = {
+/** Type of a 'lists_create_with' block. */
+type CreateWithBlock = Block&ListCreateWithMixin;
+interface ListCreateWithMixin extends ListCreateWithMixinType {
+  itemCount_: number;
+}
+type ListCreateWithMixinType = typeof LISTS_CREATE_WITH;
+
+const LISTS_CREATE_WITH = {
   /**
    * Block for creating a list with any number of elements of any type.
    */
@@ -126,7 +134,9 @@ blocks['lists_create_with'] = {
     this.itemCount_ = 3;
     this.updateShape_();
     this.setOutput(true, 'Array');
-    this.setMutator(new Mutator(['lists_create_with_item'], this));
+    this.setMutator(new Mutator(
+        ['lists_create_with_item'],
+        this as unknown as BlockSvg));  // BUG(#6905)
     this.setTooltip(Msg['LISTS_CREATE_WITH_TOOLTIP']);
   },
   /**
@@ -135,7 +145,7 @@ blocks['lists_create_with'] = {
    */
   mutationToDom: function(this: CreateWithBlock): Element {
     const container = xmlUtils.createElement('mutation');
-    container.setAttribute('items', this.itemCount_);
+    container.setAttribute('items', String(this.itemCount_));
     return container;
   },
   /**
@@ -146,7 +156,7 @@ blocks['lists_create_with'] = {
    */
   domToMutation: function(this: CreateWithBlock, xmlElement: Element) {
     const items = xmlElement.getAttribute('items');
-    if (items === null) throw new TypeError('element did not have items');
+    if (!items) throw new TypeError('element did not have items');
     this.itemCount_ = parseInt(items, 10);
     this.updateShape_();
   },
@@ -155,7 +165,7 @@ blocks['lists_create_with'] = {
    *
    * @return The state of this block, ie the item count.
    */
-  saveExtraState: function(): {itemCount: number} {
+  saveExtraState: function(this: CreateWithBlock): {itemCount: number} {
     return {
       'itemCount': this.itemCount_,
     };
@@ -165,7 +175,7 @@ blocks['lists_create_with'] = {
    *
    * @param state The state to apply to this block, ie the item count.
    */
-  loadExtraState: function(state: AnyDuringMigration) {
+  loadExtraState: function(this: CreateWithBlock, state: AnyDuringMigration) {
     this.itemCount_ = state['itemCount'];
     this.updateShape_();
   },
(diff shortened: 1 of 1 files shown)


blocks/lists.ts
```diff
@@ -138,53 +132,56 @@ blocks['lists_create_with'] = {
...
     this.updateShape_();
   },
   /**
    * Returns the state of this block as a JSON serializable object.
-   * @return {{itemCount: number}} The state of this block, ie the item count.
+   *
+   * @return The state of this block, ie the item count.
    */
-  saveExtraState: function() {
+  saveExtraState: function(): {itemCount: number} {
```
BeksOmega, line 158: Why doesn't this one need a `this:` parameter?

blocks/lists.ts
```diff
@@ -352,15 +352,14 @@ blocks['lists_getIndex'] = {
     const modeMenu = fieldRegistry.fromJson({
       type: 'field_dropdown',
       options: MODE,
-    });
+    }) as FieldDropdown;
```
BeksOmega, line 355: Here (and below) you use `as FieldDropdown` but above you use a conditional to check for defined-ness. Can you make these consistent?

blocks/lists.ts
```diff
@@ -947,49 +944,48 @@ blocks['lists_split'] = {
...
+      const inputBlock = inputConnection!.targetBlock();
       if (inputBlock) {
-        inputConnection.disconnect();
+        inputConnection!.disconnect();
         if (inputBlock.isShadow()) {
-          inputBlock.dispose();
+          inputBlock.dispose(false);
         } else {
           this.bumpNeighbours();
         }
```
BeksOmega, line 962: Not important right now, but I think all of this code is unnecessary. Any shadow should be disposed automatically from `setShadowDom(null)` and any child block should be disconnected automatically from the `setCheck` calls below.

Might be worth filing an issue about.

blocks/lists.ts
```diff
@@ -138,53 +132,56 @@ blocks['lists_create_with'] = {
...
-   * @param {!Workspace} workspace Mutator's workspace.
-   * @return {!Block} Root block in mutator.
-   * @this {Block}
+   *
+   * @param workspace Mutator's workspace.
+   * @return Root block in mutator.
    */
-  decompose: function(workspace) {
-    const containerBlock = workspace.newBlock('lists_create_with_container');
+  decompose: function(this: CreateWithBlock, workspace: Workspace): Block {
```
BeksOmega, line 178: Can you change the return to `ContainerBlock`?

blocks/lists.ts
```diff
@@ -352,15 +352,14 @@ blocks['lists_getIndex'] = {
...
-         * @this {FieldDropdown}
-         */
-        function(value) {
+        /** @param value The input value. */
+        function(this: FieldDropdown, value: string) {
           const isStatement = (value === 'REMOVE');
-          this.getSourceBlock().updateStatement_(isStatement);
+          (this.getSourceBlock() as GetIndexBlock)
+              .updateStatement_(isStatement);
+          return undefined;
```
BeksOmega, line 362: Why do you need to explicitly return undefined?

[Edit: Ah I see there are many places in this file that do this, so I assume it's just for consistency / clarity?]

BeksOmega (approved):

blocks/lists.ts
cpcallen: Only because the body does not actually reference `this`.  It could still have a `this:` though.  Do you think it is better to do so, for consistency?

BeksOmega: I thought it does reference `this.itemCount_`, which is why I was confused.

cpcallen: Oh, so it does.  I will correct (there are some other issues too.)

blocks/lists.ts
cpcallen: Because `.setValidator` does not accept `void` functions.  And I could add `: undefined` to the validator function expression, but then it complains that there is no `return` statement.  So I just added the return statement and let it infer the return type.

BeksOmega: Ah ok, can you file a bug for fixing `setValidator` to accept `void` functions? Because it definitely should. If no value (or a value of `undefined`) is returned from a validator, we consider the original value valid.

blocks/lists.ts
cpcallen: TODO added referencing #6920, containing note referencing your comment.

blocks/lists.ts
cpcallen: So, I wondered about that, but apparently in TypeScript a function that has its return type declared as `void` does not mean "no return statement" or "no return statement, or only returns `undefined`", but [instead means "caller promises to ignore any returned value"](https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-functions-returning-non-void-assignable-to-function-returning-void) which is obviously not true of validators.

If there is a way to tell TypeScript to infer that a function with no `return` statement is actually `(/*...*/) => undefined` rather than a `=> void` then that would be great, but I'm not aware of such.

BeksOmega: but but but... a function which has no return statement /is/ actually `(/* ... */) => undefined` lol

I guess this is why migrations are hard :P

blocks/lists.ts
```diff
@@ -114,9 +116,15 @@ export const blocks = createBlockDefinitionsFromJsonArray([
   },
 ]);
 
-type CreateWithBlock = Block|typeof blocks['lists_create_with'];
 
-blocks['lists_create_with'] = {
+/** Type of a 'lists_create_with' block. */
+type CreateWithBlock = Block&ListCreateWithMixin;
+interface ListCreateWithMixin extends ListCreateWithMixinType {
+  itemCount_: number;
```
BeksOmega, line 123: This is because this is added in the `init` function? Could we just define it in the mixin instead of defining it in the constructor?

Here & below.

BeksOmega (commented):
One comment, also question: did you want this to go into the upcoming release? If so, have you tested that the blocks actually function correctly?

blocks/lists.ts
cpcallen: I know.  I guess the problem is that `tsc` has to decide the return type of a `return`-less function.  Is it `void` or is it `undefined`?  It would be nice if it could be "OK, I'll treat it as `: void&undefined` (i.e., satisifies either) but there's probably some reason that's a bad idea.  Alternatively it could allow you to declare it as `: undefined` without actually having to have the `return` statement.  But alas it won't allow that either.   ¯\\\_(ツ)\_/¯

blocks/lists.ts
BeksOmega: I do believe that that is safe. We actually do that in the procedure blocks ([eg](https://github.com/google/blockly/blob/c688c9382b2878473b7072c20bd65ec40e27fe6b/blocks/procedures.js#L490)). All of the values get [mixed into the block](https://github.com/google/blockly/blob/c1e2956a318f459ffc4272b1fb9e115a2d9bb1a7/core/block.ts#L257) before the[ init method is called](https://github.com/google/blockly/blob/c1e2956a318f459ffc4272b1fb9e115a2d9bb1a7/core/block.ts#L264). So it's safe!

PR #6913, opened 2023-03-20
fix: overwriting flyout def with dynamic categories

### LINKED ISSUE #4443: Flyout.show changes flyout definition.
**Describe the bug**
When flyout.show() is given a custom category it changes the original toolbox tree found on workspace.options.languageTree.

**To Reproduce**
1. Create a toolbox with the following definition: 
```
var jsonToolbox = {
  "kind": "flyoutToolbox",
  "contents": [
    {
      "kind": "block",
      "type": "controls_if"
    },
    {
      "kind": "category",
      "custom": "VARIABLE"
    }
  ]
}
```
2. Run `Blockly.getMainWorkspace().options.languageTree` in the console.
3. Notice how the new toolbox definition now has a button instead of a custom category:
 ```
var jsonToolbox = {
  "kind": "flyoutToolbox",
  "contents": [
    {
      "kind": "block",
      "type": "controls_if"
    },
    {
      "kind": "button",
      "callback": "CREATE_VARIABLE"
    }
  ]
}
```
**Expected behavior**
The toolbox definition should still have a custom category.

**Additional context**
This bug is frustrating when creating a simple flyout with a custom category. In this case a user needs to be able to refresh their flyout when the contents in the custom category have changed. They should be able to do this by calling: 
`workspace.getFlyout().show(workspace.options.languageTree)`
However, this doesn't work because the custom category has been replaced by the button so the flyout will not update.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #4443

### Proposed Changes

Stops mutating the toolbox definition when parsing dynamic categories.

### Reason for Changes

See explanation in #4443 

### Test Coverage

Manual testing.

Can add unit testing after we confirm this is the behavior we actually want.

### Documentation

N/A

### Additional Information

Do we actually want flyout toolboxes to parse the dynamic categories? I kind of understand it in that you might want to embed variables / procedures in your flyout toolbox. But it also just seems.... weird?

Just want to confirm this before adding unit tests / merging.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -570,9 +570,9 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
       this.workspace_.removeChangeListener(this.reflowWrapper_);
       this.reflowWrapper_ = null;
     }
+    // Do NOT delete the blocks here.  Wait until Flyout.show.
+    // https://neil.fraser.name/news/2014/08/09/
   }
-  // Do NOT delete the blocks here.  Wait until Flyout.show.
-  // https://neil.fraser.name/news/2014/08/09/
 
   /**
    * Show and populate the flyout.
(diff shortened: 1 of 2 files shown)


cpcallen (approved):
This looks fine, though it took me far too long to understand what was going on: my brain misparsed things and matched the `}` following `return;` at the end of the first `if` block in `createItem` as  if it was the end of `createItem`, and I was deeply confused about why this helper function seemed to neither return a value nor have side effects.

I'm not sure if there's a good way to improve this code without breaking compatibility, but I think ideally there'd only be a single (top-level) function which recurses by calling itself.

Failing that, some comments noting that `createFlyoutInfo_` is more or less just a nonrecursive wrapper for `createInfo`, and that the latter does all the heavy lifting, might make this easier for newcomers to understand.

PR #6919, opened 2023-03-23
fix(fields): Make `Field.SKIP_SETUP` a `Symbol`; remove `class Sentinel`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

We introduced the `SKIP_SETUP` sentinel value when converting `Field` and its subclasses to ES6 `class` syntax, because `super` must be called before any other code in a subclass constructor, breaking the previous mechanism where subclasses would set some properties before calling their superclass constructor.

`SKIP_SETUP` was a singleton value of class Sentinel.

Recently, in PR #6639 @btw17 introduced the `isSentinel` type predicate to improve the typing of `Field`.  Unfortunately, there were some aspects of this change that were not very elegant:

- `isSentinel` was declared as a `static` method on `Field` (rather than on `Sentinel` itself).
- It only checks against the specific value `SKIP_SETUP`, rather than checking if the argument is `instanceof Sentinel` (though perhaps this is for efficiency?)

Additionally—as a result of the migration from ES6 to TS, and predating PR #6639—the signature for the `Field` constructor's first argument was `value: T|Sentinel`, with subclass constructors generally being `<some types>|Sentinel`.

This creates a small problem when attempting to port fields from core to plugins, because `Sentinel` is not reexported by `core/utils.ts` (and therefore not from `core/blockly.ts` either).

### Proposed Changes

The latter problem could be solved simply by reexporting `Sentinel`, or by having plugin constructors not accept `SKIP_SETUP` (though this potentially makes them more difficult to subclass), but neither is particularly desirable.

Instead, this PR proposes that we:

- Make `Field.SKIP_SETUP` a `Symbol`.
- Delete `Field.isSentinel`.
- Change the value argument to the `Field` constructor to accept `T|typeof Field.SKIP_SETUP`—where `typeof Field.SKIP_SETUP` is (like a literal type) a type that accepts just the single value `SKIP_SETUP`.
- Remove the `Sentinel` class and `core/utils/sentinel.ts`.

### Reason for Changes

To allow plug-in `Field` subclasses to use the same `SKIP_SETUP` mechanism as in-core fields.

### Test Coverage

Passes `npm test`.  No changes to manual testing expected.

### Documentation

No updates beyond automatic updates of generated reference docs.

### Additional Information

BREAKING CHANGE:

- Removes `Field.isSentinel`
  -  But this addition has not yet been published, so it can only break our own as-yet-unreleased code in samples.

- Changes the type of `Field.SKIP_SETUP` and the first argument of the `Field` constructor from `Sentinel` to `typeof SKIP_SETUP` (a unique Symbol)
  - But given that `Sentinel` has never been exported this should not break any actual external code.

For the above-noted reasons these changes should not necessitate a major version bump.


cpcallen:
@btw17: I can't formally request a review from you, but I would like your comments if you have any.

cpcallen:
Addendum: this PR also removes the `isMenuGenerator` type predicate from `core/field_dropdown.ts`; this was added in #6550 which has been released, but is unexported and therefore should be safe to remove.

rachel-fenichel:
This is a late change but I agree with merging it into the upcoming release:
- The change is largely about types, and is type-checked by the compiler
- The runtime behaviour is tested by our extensive field tests, which exercise the sentinel through subclassing
- This is a good use of `Symbol`s
- If we don't merge this, we'll need to make a breaking change to undo it in the following release
  - Merging it now is not breaking, because we haven't yet released the current version of the sentinel

btw17:
First off, I think making `Field.SKIP_SETUP` a `Symbol` is absolutely the desirable option! Using the `Sentinel` class always seemed strange to me, and I was under the impression previously that it needed to remain as that for legacy reasons.

tldr, though, I believe Field plugin subclasses should be fine as long as they call `Field.SKIP_SENTINEL` instead of `FieldSubclass.SKIP_SENTINEL`.

Clarifying my intent in #6639:
- I believe I made it a `static` method on `Field` so that it was delivered with `Field` wherever it was currently being used. Users were previously supplying `Field.SKIP_SENTINEL`, so I knew they would have this method in scope instead of having to import `Sentinel`, which they previously wouldn't have needed.
- Regarding checking the specific value, this was a deliberate choice. This behavior matched what was there previously (see code snippet below). All subclasses in the core codebase call specifically `Field.SKIP_SENTINEL` instead of `FieldSubclass.SKIP_SENTINEL`, which is always that exact single instance of `new Sentinel()` since `Field.SKIP_SENTINEL` is set to that value when the class is defined, long before its constructor is used. If calling `FieldSubclass.SKIP_SENTINEL` is what this is fixing, that would've already been broken.
```ts
// Before #6639
class Field<T> {
   static readonly SKIP_SETUP = new Sentinel();
   // ...
   constructor(value: T|Sentinel /*, ... */) {
      // ...
      if (opt_value === Field.SKIP_SETUP) {
        return;
      }
   }
   // ...
}

// After #6639
class Field<T> {
   static readonly SKIP_SETUP = new Sentinel();
   static isSentinel<T>(value: T|Sentinel): value is Sentinel {
      return value === Field.SKIP_SETUP;
   }
   // ...
   constructor(value: T|Sentinel /*, ... */) {
      // ...
      if (Field.isSentinel(opt_value)) return;
   }
   // ...
}
```

btw17 (commented):
Looked through it all - LGTM! (Although I don't have the permissions to give approval)

cpcallen:
I think this is now awaiting formal approval by @rachel-fenichel.

rachel-fenichel (approved):

rachel-fenichel:
Please fix the PR title (no !) before merging.

cpcallen:
> First off, I think making `Field.SKIP_SETUP` a `Symbol` is absolutely the desirable option! Using the `Sentinel` class always seemed strange to me, and I was under the impression previously that it needed to remain as that for legacy reasons.

It was introduced during the ES5->ES6 phase of the migration, and it is almost certainly my fault, though I am no longer exactly sure why I suggested creating a `Sentinel` class rather than using a `Symbol`; maybe I'd just temporarily forgotten about symbols (I have done vastly more ES5 work than ES6), or perhaps I was concerned about the limitations of the Closure type system which (as far as I am aware) has no equivalent of a literal type or unique symbol type: I think the constructor could have had `/** @param {T|Symbol} value */`, but I may have been worried that this could cause issues if someone wanted to define a `Symbol`-valued field.  (Not sure why I would have worried about that, though, because a sentinel symbol is no more difficult to pick out using `===` than a sentinel object is.)  Most likely it was just that I had used this [exact solution to type sentinel values in the Code City interpreter](https://github.com/google/CodeCity/blob/9739d31fdeb262788040b1f0d3c91313f8abf831/server/interpreter.js#L163).

> > tldr, though, I believe Field plugin subclasses should be fine as long as they call `Field.SKIP_SENTINEL` instead of `FieldSubclass.SKIP_SENTINEL`.

Actually, using `FieldSubclass.SKIP_SENTINEL` should be just fine (unless `FieldSubclass` redefines `SKIP_SENTINEL`) since it will inherit the value from the `Field` constructor.  (I actually tested this at one point while preparing this PR as an alternative to adding `import {Field} from './field.ts';` in one file that didn't otherwise need to import `Field`.)

> Clarifying my intent in #6639:
> 
> * I believe I made it a `static` method on `Field` so that it was delivered with `Field` wherever it was currently being used. Users were previously supplying `Field.SKIP_SENTINEL`, so I knew they would have this method in scope instead of having to import `Sentinel`, which they previously wouldn't have needed.

That's true, though any code that was likely to need `isSentinel` would probably also need `Sentinel` (in order to, for example, declare that a method can accept a `Sentinel` value to test).

> * Regarding checking the specific value, this was a deliberate choice. This behavior matched what was there previously (see code snippet below).


 All subclasses in the core codebase call specifically `Field.SKIP_SENTINEL` instead of `FieldSubclass.SKIP_SENTINEL`, which is always that exact single instance of `new Sentinel()` since `Field.SKIP_SENTINEL` is set to that value when the class is defined, long before its constructor is used. If calling `FieldSubclass.SKIP_SENTINEL` is what this is fixing, that would've already been broken.
> 
> ```ts
> // Before #6639
> class Field<T> {
>    static readonly SKIP_SETUP = new Sentinel();
>    // ...
>    constructor(value: T|Sentinel /*, ... */) {
>       // ...
>       if (opt_value === Field.SKIP_SETUP) {
>         return;
>       }
>    }
>    // ...
> }
> 
> // After #6639
> class Field<T> {
>    static readonly SKIP_SETUP = new Sentinel();
>    static isSentinel<T>(value: T|Sentinel): value is Sentinel {
>       return value === Field.SKIP_SETUP;
>    }
>    // ...
>    constructor(value: T|Sentinel /*, ... */) {
>       // ...
>       if (Field.isSentinel(opt_value)) return;
>    }
>    // ...
> }
> ```

cpcallen:
(Previous comment edited due to being inadvertently submitted before complete.  Check [hopefully final version here](https://github.com/google/blockly/pull/6919#issuecomment-1481817011).

PR #6929, opened 2023-03-28
chore: Don't call getConstants repeatedly

NeilFraser:


maribethb (approved):
LGTM, two notes

1. I think `getConstants` could be updated to just throw an error if the block is not rendered, is dead or dying, etc. as it's not valid to call it right then. Then we can say it does not return null, and eliminate all of these assertions.
2. You can do `const constants = this.getConstants()!` which means you don't have to do the assertion every time you reference `constants` and also eliminates the `!constants!` which I find unpleasant to look at... but up to you.

Not gonna insist on either of these, but I do think either one would be an improvement.

rachel-fenichel:
Would you say you don't want to call `getConstants` constantly?

(I'll see myself out)

BeksOmega:
I'm going to go ahead and close this as stale. Feel free to reopen!

PR #6944, opened 2023-04-01
feat: Parse newlines in JSON message as row separators.

### LINKED ISSUE #6809: Parse newlines in dummy inputs as row separators

Developers sometimes want multiple dummy inputs on separate lines for clarity/readability, but want the rest of the block to use inline inputs.

From https://github.com/google/blockly-samples/pull/1398

> This plugin is a proof-of-concept. I suggest that Blockly should eventually support putting \n characters in block JSON message0 to accomplish the same effect. 

**Additional context**

See https://github.com/google/blockly-samples/pull/1398 and https://github.com/google/blockly/issues/216

johnnesky:
## The basics

- [x] I branched from develop

### Resolves

Fixes #6809. Interprets newline characters in the JSON message as ~~dummies~~ a new subclass of Input called EndRowInput that force any following inputs to be rendered on a new row.

#### Behavior Before Change

Newline characters from the message are preserved, but render as regular spaces by the browser in a label field.

#### Behavior After Change

Newline characters are converted to special ~~dummy~~ end-row inputs that force the next input to be rendered on a new line. (Also, these dummies use the same alignment as the implicit last dummy.)

The common RenderInfo recognizes the new dummies and handles them appropriately.

### Reason for Changes

This allows more customization of block layout, including putting inline inputs on multiple rows.

### Test Coverage

Unit tests in `block_json_test.js` and `block_test.js` test token parsing and conversion to dummies. Rendering is not tested.

### Documentation

Information about using newline characters in JSON messages, and the new "input_end_row" JSON input type, could be added to: https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks


tests/mocha/block_json_test.js
```diff
@@ -65,6 +65,10 @@ suite('Block JSON initialization', function() {
       this.assertError(['test', 2, 'test'], 1,
           'Block "test": Message index %2 out of range.');
     });
+
+    test('Accepts newline token', function() {
```
BeksOmega, line 69: Nit: For new tests I prefer names that follow the format: *tests that* 'my test name here'.

Eg: 'newline characters are accepted', 'newlines are converted to dummy inputs with endOfRow set'. etc.

See here for reasoning: https://testing.googleblog.com/2014/10/testing-on-toilet-writing-descriptive.html

core/utils/parsing.ts
```diff
@@ -21,8 +21,8 @@ import * as colourUtils from './colour.js';
  *
  * @param message Text which might contain string table references and
  *     interpolation tokens.
- * @param parseInterpolationTokens Option to parse numeric
```
BeksOmega, line 24: Could we change this param to have a more descriptive name, given the new functionality? My  understanding is that if this is false, then we just parse message references and don't do anything else. So could we use `onlyReplaceMessageReferences = true`?

core/utils/parsing.ts
```diff
@@ -21,8 +21,8 @@ import * as colourUtils from './colour.js';
...
- *     interpolation tokens (%1, %2, ...) when true.
+ * @param parseInterpolationTokens Option to parse numeric interpolation
+ *     tokens (%1, %2, ...) and newline characters when true.
  * @returns Array of strings and numbers.
  */
 function tokenizeInterpolationInternal(
```
BeksOmega, line 28: Could you add tests for this new functionality to `tests/mocha/utils_test.js`? (sorry for the terrible file naming lol)

core/utils/parsing.ts
```diff
@@ -132,11 +141,11 @@ function tokenizeInterpolationInternal(
     tokens.push(text);
   }
 
-  // Merge adjacent text tokens into a single string.
+  // Merge adjacent text tokens into a single string (excluding newline tokens).
```
BeksOmega, line 144: In the case that we're just replacing message references, the `tokens` array is of length `1` correct? So this still properly returns a `mergedTokens` array of length `1` in that case?

Just want to make sure I understand what this is doing :P

core/block.ts
```diff
@@ -1684,7 +1684,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    *     with fields or inputs defined in the args array.
    * @param args Array of arguments to be interpolated.
    * @param lastDummyAlign If a dummy input is added at the end, how should it
```
BeksOmega, line 1686: Could this be renamed to something more descriptive?

core/input.ts
```diff
@@ -249,6 +255,31 @@ export class Input {
...
+    return this.endOfRow;
+  }
+
+  /**
+   * Set whether the input should be the last input on its row.
+   *
+   * @param endOfRow Whether the input should be the last input on its row.
+   * @returns The input being modified (to allow chaining).
+   */
+  setEndOfRow(endOfRow: boolean): Input {
```
BeksOmega, line 274: Thanks for filling out the API for this, looks great :D

BeksOmega (changes requested):
Generally LGTM! Mostly comments about renaming parameters :P

Wrt documentation: yes we should definitely document this. Could you file a docs issue for that (when you have time) and add it to our Q2 Github Project?

Also @maribethb the idea was that this doesn't force v10 (even though it's technically a breaking change in that it would change block rendering for people currently using \n) correct?

BeksOmega:
Another more design-related question (possibly for @maribethb ) Have we considered adding some more generic extra options to inputs that would allow them to be rendered differently? E.g. https://github.com/google/blockly/pull/3219/files#diff-e01c94caccf98f950fdbe393516b2595473c2e1f04e9f302ec62ed66549e2886R41

I believe this would allow App Inventor to implement their indented inputs (which it seems they still want) relatively easily externally.

core/utils/parsing.ts
johnnesky: The name onlyReplaceMessageReferences sounds like it would be inverting the current value of the parameter, in which case we'd have to find all the callers and change the argument they provide for the second parameter, which I guess we could do if this is a private API? Another thing we could do it just add a new third parameter called something like "tokenizeNewlines" which defaults to have the same value as the existing "parseInterpolationTokens" parameter.

core/utils/parsing.ts
johnnesky: I believe message references involving % characters will end up as separate elements in the tokens array above here, but this loop will those tokens via the buffer variable here, which will then become the only output token.

core/block.ts
```diff
@@ -1775,11 +1775,21 @@ export class Block implements IASTNodeLocation, IDeletable {
...
-        // AnyDuringMigration because:  Type '{ text: string; type: string; } |
-        // null' is not assignable to type 'string | number'.
-        element = this.stringToFieldJson_(element) as AnyDuringMigration;
-        if (!element) {
-          continue;
+        if (element === '\n') {
+          // Convert newline tokens to dummies with endOfRow enabled.
+          const newlineInput = {'type': 'input_dummy', 'endOfRow': true};
+          // Treat these as the "last" dummy for alignment purposes.
+          if (lastDummyAlign) {
```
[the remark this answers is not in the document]
johnnesky: I figured that it's likely that users would want the default alignment applied to the final row to also apply to other rows, and in the rare cases where that's not wanted users could just be more explicit, but it's true that this means that the name lastDummyAlign isn't really accurate anymore and implicitDummyAlign would be a more appropriate name. I'd be fine adding this alias to the JSON schema (but parsing it as the same property) if that's alright with everyone else.

core/renderers/geras/info.ts
```diff
@@ -360,7 +360,8 @@ export class RenderInfo extends BaseRenderInfo {
             row.width < prevInput.width) {
           rowNextRightEdges.set(row, prevInput.width);
         } else {
-          nextRightEdge = row.width;
+          // To avoid jagged right edges, use the maximum width.
+          nextRightEdge = Math.max(nextRightEdge, row.width);
```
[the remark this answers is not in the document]
johnnesky: Upon looking into it further, I think my fix isn't right either.

If I revert this change, the original geras renderer sometimes renders the newline feature like this, with a weird corner on the top of the block:
image

My change here fixes that... but also alters the behavior of rows that are separated by a statement input, using the widest non-statement row width so far (iterating upward through the rows):
image

Note that this geras override of the method alignRowElements_() only applies if the block's inputs are inline, in which case, up until now, all consecutive inline value inputs were expected to end up in a single row, possibly separated by statement rows. The statement rows are supposed to allow the inline value rows to have different widths (but at least as wide as the statement input connectors), which I assume is the entire point of this method override. (None of the other renderers override alignRowElements_().) 

The new feature is the possibility of multiple adjacent rows of inline value inputs, or even rows containing only text labels. Without a statement row to separate these rows, I assume they ought to be the same width... and the implicit top row that's inserted for rendering corners and a margin should also be the same width, but it seems to only be as wide as the minimum width of the first explicit row, hence the weird corner I was trying to fix.

Okay, I think I've figured out the fix for real this time, I'm using the old logic for statement rows and the new logic otherwise, and it renders like this:
image

core/utils/parsing.ts
BeksOmega: This is indeed module-private (since it's not exported) so we can do whatever we want with it =) I think either solution works here!

johnnesky: I added a separate parameter "tokenizeNewlines" to be explicit.

Precha24 (commented):
Good, thanks

johnnesky:
I've reimplemented this as a new subclass of Input, similar to DummyInput!

johnnesky:
Created an issue to add documentation: https://github.com/google/blockly/issues/7360

johnnesky:
Added input_end_row to the block factory demo.

demos/blockfactory/blocks.js
```diff
@@ -228,6 +228,24 @@ Blockly.Blocks['input_dummy'] = {
...
+  // End-row input.
+  init: function() {
+    this.jsonInit({
+      "message0": "end-row input",
+      "message1": FIELD_MESSAGE,
+      "args1": FIELD_ARGS,
+      "previousStatement": "Input",
+      "nextStatement": "Input",
+      "colour": 210,
+      "tooltip": "For adding fields at the end of a row with no " +
```
BeksOmega, line 241: Nit: Update tooltip.

tests/mocha/block_json_test.js
BeksOmega, line 1: Absolutely georgous tests =)

core/block.ts
```diff
@@ -1628,8 +1646,8 @@ export class Block implements IASTNodeLocation, IDeletable {
       this.interpolate_(
         json['message' + i],
         json['args' + i] || [],
-        // Backwards compatibility: lastDummyAlign aliases implicitDummyAlign.
-        json['lastDummyAlign' + i] || json['implicitDummyAlign' + i],
+        // Backwards compatibility: lastDummyAlign aliases implicitAlign.
+        json['lastDummyAlign' + i] || json['implicitAlign' + i],
```
BeksOmega, line 1650: Can we swap these so that 'implicitAlign' has a higher precedence? I think it's weird that if they both exist 'lastDummyAlign' gets applied to end row inputs instead of 'implicitAlign'. I would rather have people move to the cleaner new property.

Can you also update the block factory to use implicitAlign instead of lastDummyAlign?

BeksOmega (commented):
Generally LGTM! Just a few little things =)

core/inputs/end_row_input.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * @param name Language-neutral identifier which may used to find this input
+   *     again.
+   * @param block The block containing this input.
+   */
+  constructor(
+    public name: string,
+    block: Block,
+  ) {
```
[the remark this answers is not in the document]
johnnesky: Done.

core/renderers/common/info.ts
```diff
@@ -326,7 +327,7 @@ export class RenderInfo {
     } else if (input instanceof ValueInput) {
       activeRow.elements.push(new ExternalValueInput(this.constants_, input));
       activeRow.hasExternalInput = true;
-    } else if (input instanceof DummyInput) {
+    } else if (input instanceof DummyInput || input instanceof EndRowInput) {
       // Dummy inputs have no visual representation, but the information is
```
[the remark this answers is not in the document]
johnnesky: Done.

core/renderers/geras/info.ts
```diff
@@ -90,7 +91,7 @@ export class RenderInfo extends BaseRenderInfo {
     } else if (input instanceof ValueInput) {
       activeRow.elements.push(new ExternalValueInput(this.constants_, input));
       activeRow.hasExternalInput = true;
-    } else if (input instanceof DummyInput) {
+    } else if (input instanceof DummyInput || input instanceof EndRowInput) {
       // Dummy inputs have no visual representation, but the information is
```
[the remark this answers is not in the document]
johnnesky: Done.

tests/mocha/block_json_test.js
johnnesky: Thanks!

core/block.ts
johnnesky: Done.

demos/blockfactory/blocks.js
johnnesky: I updated the tooltips for both this and dummy inputs, please take a look.

core/renderers/geras/info.ts
```diff
@@ -379,8 +380,12 @@ export class RenderInfo extends BaseRenderInfo {
           row.width < prevInput.width
         ) {
           rowNextRightEdges.set(row, prevInput.width);
-        } else {
+        } else if (row.hasStatement) {
           nextRightEdge = row.width;
+        } else {
+          // To avoid jagged right edges along consecutive non-statement rows,
```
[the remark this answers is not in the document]
johnnesky: Done.

BeksOmega (approved):

johnnesky:
Created https://github.com/google/blockly-samples/issues/1854 to deprecate the renderer plugin that accomplished something similar.

PR #6952, opened 2023-04-04
fix: non-printable characters in XML

### LINKED ISSUE #4590: Blockly does not serialize non-printable characters correctly

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #4590 

### Proposed Changes

Makes it so that non-printable characters are properly escaped when serializing to XML, and that we do best-effort parsing when deserializing unescaped characters.

### Reason for Changes

App Inventor needs this to be pulled into core for them to be able to update.

### Test Coverage

Unit tests for serializing and deserializing.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -117,7 +117,7 @@ export function textToDom(text: string): Element {
     const oParser = new DOMParser();
     doc = oParser.parseFromString(text, 'text/html');
     if (!doc || !doc.body.firstChild ||
-        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
+        doc.body.firstChild.nodeName.toLowerCase() !== 'xml') {
       throw new Error(`DOMParser was unable to parse: ${text}`);
     }
     return doc.body.firstChild as Element;
@@ -145,12 +145,11 @@ export function textToDomDocument(text: string): Document {
  * @returns Text representation.
  */
 export function domToText(dom: Node): string {
-  const oSerializer = new XMLSerializer();
-  return sanitizeText(oSerializer.serializeToString(dom));
+  return sanitizeText(new XMLSerializer().serializeToString(dom));
 }
 
 function sanitizeText(text: string) {
   return text.replace(
       INVALID_CONTROL_CHARS,
-      (match) => `&#x${match.charCodeAt(0).toString(16)};`);
+      (match) => `&#x${match.charCodeAt(0)};`);
 }
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -13,14 +13,16 @@ import * as deprecation from './deprecation.js';
 let domParser: DOMParser = {
   parseFromString: function() {
     throw new Error(
-        'DOMParser was not properly injected using injectDependencies');
+        'DOMParser was not found in the global scope and was not properly ' +
+        'injected using injectDependencies');
   },
 };
 
 let xmlSerializer: XMLSerializer = {
   serializeToString: function() {
     throw new Error(
-        'XMLSerializer was not properly injected using injectDependencies');
+        'XMLSerializer was not foundin the global scope and was not properly ' +
+        'injected using injectDependencies');
   },
 };
 
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -32,8 +32,8 @@ let xmlSerializer: XMLSerializer = {
  * package instead.
  */
 let {document, DOMParser, XMLSerializer} = globalThis;
-domParser = new DOMParser();
-xmlSerializer = new XMLSerializer();
+if (DOMParser) domParser = new DOMParser();
+if (XMLSerializer) xmlSerializer = new XMLSerializer();
 
 /**
  * Inject implementations of document, DOMParser and/or XMLSerializer
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -57,6 +57,9 @@ export function injectDependencies(dependencies: {
  */
 export const NAME_SPACE = 'https://developers.google.com/blockly/xml';
 
+// eslint-disable-next-line no-control-regex
+const INVALID_CONTROL_CHARS = /[\x00-\x09\x0B\x0C\x0E-\x1F]/g;
+
 /**
  * Get the document object to use for XML serialization.
  *
@@ -108,10 +111,16 @@ export function createTextNode(text: string): Text {
  * @throws if the text doesn't parse.
  */
 export function textToDom(text: string): Element {
-  const doc = textToDomDocument(text);
+  let doc = textToDomDocument(text);
   if (!doc || !doc.documentElement ||
       doc.getElementsByTagName('parsererror').length) {
-    throw Error('textToDom was unable to parse: ' + text);
+    const oParser = new DOMParser();
+    doc = oParser.parseFromString(text, 'text/html');
+    if (!doc || !doc.body.firstChild ||
+        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
+      throw new Error(`DOMParser was unable to parse: ${text}`);
+    }
+    return doc.body.firstChild as Element;
   }
   return doc.documentElement;
 }
@@ -137,5 +146,11 @@ export function textToDomDocument(text: string): Document {
  */
 export function domToText(dom: Node): string {
   const oSerializer = new XMLSerializer();
-  return oSerializer.serializeToString(dom);
+  return sanitizeText(oSerializer.serializeToString(dom));
+}
+
+function sanitizeText(text: string) {
+  return text.replace(
+      INVALID_CONTROL_CHARS,
+      (match) => `&#x${match.charCodeAt(0).toString(16)};`);
 }
--- a/core/utils/xml.ts
+++ b/core/utils/xml.ts
@@ -118,16 +118,20 @@ export function createTextNode(text: string): Text {
  */
 export function textToDom(text: string): Element {
   let doc = domParser.parseFromString(text, 'text/xml');
-  if (!doc || !doc.documentElement ||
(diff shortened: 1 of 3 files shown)


core/utils/xml.ts
```diff
@@ -137,5 +146,11 @@ export function textToDomDocument(text: string): Document {
  */
 export function domToText(dom: Node): string {
   const oSerializer = new XMLSerializer();
-  return oSerializer.serializeToString(dom);
+  return sanitizeText(oSerializer.serializeToString(dom));
```
cpcallen, line 149: Nit: please fix hungarian notation (`oSerializer` -> `serializer`) while you're at it.

tests/mocha/serializer_test.js
```diff
@@ -1882,4 +1882,4 @@ const runSerializerTestSuite = (serializer, deserializer, testSuite) => {
 };
 
 runSerializerTestSuite(null, null, Serializer);
-runSerializerTestSuite((state) => state, (state) => state, Serializer);
+// runSerializerTestSuite((state) => state, (state) => state, Serializer);
```
cpcallen, line 1885: Uncomment?

core/utils/xml.ts
```diff
@@ -57,6 +57,9 @@ export function injectDependencies(dependencies: {
  */
 export const NAME_SPACE = 'https://developers.google.com/blockly/xml';
 
+// eslint-disable-next-line no-control-regex
+const INVALID_CONTROL_CHARS = /[\x00-\x09\x0B\x0C\x0E-\x1F]/g;
```
cpcallen, line 61: What about surrogate pairs?  Does `XMLSerializer` handle those correctly already?

cpcallen (commented):
I have so many questions about this, starting with: is `XMLSerialzer` really specced to return invalid XML which can't be parsed with `DOMParser`???

core/utils/xml.ts
```diff
@@ -137,5 +146,11 @@ export function textToDomDocument(text: string): Document {
...
 export function domToText(dom: Node): string {
   const oSerializer = new XMLSerializer();
-  return oSerializer.serializeToString(dom);
+  return sanitizeText(oSerializer.serializeToString(dom));
+}
+
+function sanitizeText(text: string) {
+  return text.replace(
+      INVALID_CONTROL_CHARS,
+      (match) => `&#x${match.charCodeAt(0).toString(16)};`);
```
NeilFraser, line 155: I'd recommend using decimal codes rather than hex.  Shorter (at least for smaller code points), faster, simpler.
`&#${match.charCodeAt(0)};`

core/utils/xml.ts
```diff
@@ -108,10 +111,16 @@ export function createTextNode(text: string): Text {
...
 export function textToDom(text: string): Element {
-  const doc = textToDomDocument(text);
+  let doc = textToDomDocument(text);
   if (!doc || !doc.documentElement ||
       doc.getElementsByTagName('parsererror').length) {
-    throw Error('textToDom was unable to parse: ' + text);
+    const oParser = new DOMParser();
+    doc = oParser.parseFromString(text, 'text/html');
+    if (!doc || !doc.body.firstChild ||
+        doc.body.firstChild.nodeName.toLowerCase() != 'xml') {
```
NeilFraser, line 120: Aren't we doing `!==` these days?

BeksOmega: Whoops, was pulled over from the previous PR

core/utils/xml.ts
BeksOmega: I just removed the const assignment.

cpcallen (changes requested):
Additional comments about `textToDom`, after discussion with Neil (but don't blame him if you don't like them!):

- AFAICT there's no reason to ever create more than one `DOMParser` (or `XMLSerializer`) instance, so I suggest having the module create a single module-local one and use it everywhere.
- `textToDomDocument` is only used by `textToDom` in this module (and nowhere else), so I suggest it be deprecated and have `textToDom` always use (the) `DOMParser` (instance) directly.
- I'm unclear about why `textToDom` tries parsing as HTML if parsing as XML fails.  That seems risky, and at least deserves an inline comment if not a note in the JSDoc.
- Normally I prefer handling errors at the top with an `if (fail) return;`, but in this case the error handlers end up getting nested, so I think it would be clearer to rewrite as:
  ```JS
  try xml
  if (xml win) return xml
  try html
  if (html win) return html
  throw 'I give up';
  ```

core/utils/xml.ts
```diff
@@ -145,12 +145,10 @@ export function textToDomDocument(text: string): Document {
...
-  const oSerializer = new XMLSerializer();
-  return sanitizeText(oSerializer.serializeToString(dom));
+  return sanitizeText(new XMLSerializer().serializeToString(dom));
 }
 
 function sanitizeText(text: string) {
   return text.replace(
-      INVALID_CONTROL_CHARS,
-      (match) => `&#x${match.charCodeAt(0).toString(16)};`);
+      INVALID_CONTROL_CHARS, (match) => `&#x${match.charCodeAt(0)};`);
```
cpcallen, line 153: Don't you need to lose the `x` too?

tests/mocha/serializer_test.js
```diff
@@ -708,10 +708,10 @@ Serializer.Fields.Variable.Types = new SerializerTestCase('Types',
 Serializer.Fields.Variable.Tabs = new SerializerTestCase('Tabs',
     '<xml xmlns="https://developers.google.com/blockly/xml">' +
     '<variables>' +
-    '<variable id="aaaaaaaaaaaaaaaaaaaa">line1	line2	line3</variable>' +
+    '<variable id="aaaaaaaaaaaaaaaaaaaa">line1&amp;#x9line2&amp;#x9line3</variable>' +
```
NeilFraser, line 711: Note that these tests still use hex.

core/utils/xml.ts
```diff
@@ -133,8 +138,9 @@ export function textToDom(text: string): Element {
  * @throws if XML doesn't parse.
  */
 export function textToDomDocument(text: string): Document {
-  const oParser = new DOMParser();
-  return oParser.parseFromString(text, 'text/xml');
+  deprecation.warn(
+      'Blockly.utils.xml.textToDomDocument', 'version 10', 'version 11');
```
cpcallen, line 146: Will our next release be v10?  If we release a 9.4 then this notice will actually be wrong.  🤔

BeksOmega: yep this can't go in until @maribethb + @rachel-fenichel confirm whether we're doing a v9 or a v10

BeksOmega:
@cpcallen Ok I think this is actually good for another look!

tests/mocha/xml_test.js
```diff
@@ -73,13 +76,29 @@ suite('XML', function() {
...
+      assertXmlDoc(dom);
+      chai.assert.equal(
+          dom.getElementsByTagName('block').length, 6, 'Block tags');
+    });
+
+    test('text with NCR Control characters are properly deserialized',
+        function() {
+          const dom = Blockly.utils.xml.textToDom('<xml>&#x1;</xml>');
+          assertXmlDoc(dom);
+          chai.assert.equal(dom.firstChild.textContent, ''); // u0001
```
cpcallen, line 92: It looks like the character between the `''`s not a space but actually U+0001 / `0x01` / ascii `SOH` / `^a`—or rather it _completely fails to look like that_, the necessitating the comment.  Please use an escape sequence:

```suggestion
          chai.assert.equal(dom.firstChild.textContent, '\x01');
```
or
```suggestion
          chai.assert.equal(dom.firstChild.textContent, '\u0001');
```

(Despite the styleguide rule about comments for escape sequences I don't think one is required here because it's reasonably clear you are checking the decoding, rather than caring about the particular character in question.)

tests/mocha/xml_test.js
```diff
@@ -73,13 +76,29 @@ suite('XML', function() {
...
-      chai.assert.equal(dom.nodeName, 'xml', 'XML tag');
-      chai.assert.equal(dom.getElementsByTagName('block').length, 6, 'Block tags');
+      assertXmlDoc(dom);
+      chai.assert.equal(
+          dom.getElementsByTagName('block').length, 6, 'Block tags');
+    });
+
+    test('text with NCR Control characters are properly deserialized',
+        function() {
+          const dom = Blockly.utils.xml.textToDom('<xml>&#x1;</xml>');
```
cpcallen, line 90: Please add a decimal NCR as well as the hex one, since the encoder uses decimal now (if I have understood correctly).

tests/mocha/xml_test.js
cpcallen: (A type in the earlier version of the first suggestion has been fixed.)

core/utils/xml.ts
BeksOmega: For history: we are moving to v10

cpcallen (approved):

PR #6958, opened 2023-04-07
refactor(blocks): Migrate `blocks/text.js` to TypeScript

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6828

### DIFF
--- a/blocks/text.ts
+++ b/blocks/text.ts
@@ -12,27 +12,22 @@ import * as goog from '../closure/goog/goog.js';
 goog.declareModuleId('Blockly.libraryBlocks.texts');
 
 import * as Extensions from '../core/extensions.js';
-import {Msg} from '../core/msg.js';
 import * as fieldRegistry from '../core/field_registry.js';
-/* eslint-disable-next-line no-unused-vars */
 import * as xmlUtils from '../core/utils/xml.js';
 import {Align} from '../core/input.js';
-/* eslint-disable-next-line no-unused-vars */
 import type {Block} from '../core/block.js';
-import type { BlockSvg } from '../core/block_svg.js';
-import type {BlockDefinition} from '../core/blocks.js';
+import type {BlockSvg} from '../core/block_svg.js';
 import {ConnectionType} from '../core/connection_type.js';
+import {FieldImage} from '../core/field_image.js';
+import {FieldDropdown} from '../core/field_dropdown.js';
+import {FieldTextInput} from '../core/field_textinput.js';
+import {Msg} from '../core/msg.js';
 import {Mutator} from '../core/mutator.js';
-/* eslint-disable-next-line no-unused-vars */
 import type {Workspace} from '../core/workspace.js';
 import {createBlockDefinitionsFromJsonArray, defineBlocks} from '../core/common.js';
-/** @suppress {extraRequire} */
 import '../core/field_multilineinput.js';
-/** @suppress {extraRequire} */
 import '../core/field_variable.js';
-import { FieldImage} from '../core/field_image.js';
-import { FieldDropdown } from '../core/field_dropdown.js';
-import { FieldTextInput } from '../core/field_textinput.js';
+import { Connection } from '../core/connection.js';
 
 
 /**
@@ -249,9 +244,9 @@ interface GetSubstringMixin extends GetSubstringType {
   WHERE_OPTIONS_1: Array<[string, string]>;
   WHERE_OPTIONS_2: Array<[string, string]>;
 }
-type GetSubstringType = typeof TEXT_GET_SUBSTRING_BLOCK;
+type GetSubstringType = typeof GET_SUBSTRING_BLOCK;
 
-const TEXT_GET_SUBSTRING_BLOCK = {
+const GET_SUBSTRING_BLOCK = {
   /**
    * Block for getting substring.
    */
@@ -284,6 +279,7 @@ const TEXT_GET_SUBSTRING_BLOCK = {
   /**
    * Create XML to represent whether there are 'AT' inputs.
    * Backwards compatible serialization implementation.
+   *
    * @returns XML storage element.
    */
   mutationToDom: function(this: GetSubstringBlock): Element {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@cpcallen fyi I made a start on the text blocks, but will need type help to disentangle some of the combinations of extensions and mixins.

blocks/text.ts
```diff
@@ -339,15 +334,15 @@ blocks['text_getSubstring'] = {
     const menu = fieldRegistry.fromJson({
       type: 'field_dropdown',
       options: this['WHERE_OPTIONS_' + n],
```
rachel-fenichel, line 336: @cpcallen What do you think I should do for this dynamic access?

blocks/text.ts
```diff
@@ -282,7 +279,7 @@ blocks['text_getSubstring'] = {
    * @return {!Element} XML storage element.
    * @this {Block}
    */
-  mutationToDom: function() {
+  mutationToDom: function(this: Block): Element {
     const container = xmlUtils.createElement('mutation');
     const isAt1 = this.getInput('AT1').type === ConnectionType.INPUT_VALUE;
```
rachel-fenichel, line 284: @cpcallen `this.getInput('AT')` is possible null, because it's added and removed dynamically. Is there a way to guarantee that it's present, or should I just use `!` to assert that it is non-null?

rachel-fenichel:
@cpcallen this is ready for review!

blocks/text.ts
```diff
@@ -802,24 +812,24 @@ const TEXT_JOIN_MUTATOR_MIXIN = {
...
     const connections = [];
     while (itemBlock) {
       if (itemBlock.isInsertionMarker()) {
-        itemBlock = itemBlock.getNextBlock();
+        itemBlock = itemBlock.getNextBlock()!;
         continue;
       }
-      connections.push(itemBlock.valueConnection_);
-      itemBlock = itemBlock.getNextBlock();
+      connections.push((itemBlock as JoinMutatorItemBlock).valueConnection_);
```
cpcallen, line 827: This cast should be avoidable.

blocks/text.ts
```diff
@@ -828,15 +838,15 @@ const TEXT_JOIN_MUTATOR_MIXIN = {
...
   },
   /**
    * Store pointers to any connected child blocks.
-   * @param {!Block} containerBlock Root block in mutator.
-   * @this {Block}
+   *
+   * @param containerBlock Root block in mutator.
    */
-  saveConnections: function(containerBlock) {
+  saveConnections: function(this: JoinMutatorBlock, containerBlock: Block) {
```
cpcallen, line 849: Again: a more specific type for `containerBlock` should be possible and might be useful.

rachel-fenichel: I'm neutral on the specific type for the container block. Do you want me to add it? My default is to leave it alone.

blocks/text.ts
```diff
@@ -724,76 +723,87 @@ const QUOTE_IMAGE_MIXIN = {
...
+/** Type of a block that has TEXT_JOIN_MUTATOR_MIXIN */
+type JoinMutatorBlock = Block&JoinMutatorMixin&QuoteImageMixin;
+interface JoinMutatorMixin extends JoinMutatorMixinType {}
+type JoinMutatorMixinType = typeof JOIN_MUTATOR_MIXIN;
+
+/** Type of a item block in the text_join_mutator bubble. */
+type JoinMutatorItemBlock = BlockSvg&JoinMutatorItemMixin;
+interface JoinMutatorItemMixin {
+  valueConnection_: Connection|null
+}
```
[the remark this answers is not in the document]
cpcallen: By "what the mixin is actually used for", I meant which blocks use it and (at a high level) what the purpose of the mutator is—e.g.: the're a mutator mixin in in `logic.ts` that is used by the `controls_if` block to add and remove `elseif` and `else` clauses.

blocks/text.ts
```diff
@@ -802,24 +812,24 @@ const TEXT_JOIN_MUTATOR_MIXIN = {
...
    * Reconfigure this block based on the mutator dialog's components.
-   * @param {!Block} containerBlock Root block in mutator.
-   * @this {Block}
+   *
+   * @param containerBlock Root block in mutator.
    */
-  compose: function(containerBlock) {
-    let itemBlock = containerBlock.getInputTargetBlock('STACK');
+  compose: function(this: JoinMutatorBlock, containerBlock: Block) {
+    let itemBlock = containerBlock.getInputTargetBlock('STACK') as BlockSvg;
```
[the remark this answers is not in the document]
cpcallen: That's a reasonable view, but perhaps the correct choice for now is to say it's `JoinMutatorItemBlock&BlockSvg`.

Overall I'd like to see the block definitions freed from any assumption of the existence of `BlockSvg` and `WorkspaceSvg`, though I suspect that may be tricky so long as Mutators are assumed to an Svg thing.  (Basically: separating model from view and controller).

blocks/text.ts
```diff
@@ -648,16 +651,15 @@ blocks['text_reverse'] = {
...
 
-/**
- * @mixin
- * @package
- * @readonly
- */
+/** Type of a block that has QUOTE_IMAGE_MIXIN */
+type QuoteImageMixinBlock = Block&QuoteImageMixin;
+interface QuoteImageMixin extends QuoteImageMixinType {}
+type QuoteImageMixinType = typeof QUOTE_IMAGE_MIXIN;
```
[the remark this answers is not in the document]
cpcallen: OK, but can we at least remove "Mixin" from `QuoteImageMixinBlock`?  Because otherwise (for consistency with the naming convention established everywhere else) the other two should be `QuoteImageMixinMixin` and `QuiteImageMixinMixinType`.

blocks/text.ts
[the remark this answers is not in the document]
cpcallen: (Attempt #N at writing a coherent reply comment here:)

I think casting the result of `getNextBlock()` is the correct thing to do in any case (at least pending adding an actual runtime type check in a future PR), but I don't think it matters very much for overall.

cpcallen (approved):
With one nit (remove `Mixin` from a `Block` type's name) this LGTM.

PR #6962, opened 2023-04-07
chore: remove underscores from private properties and methods for names and workspace comments

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of #6548

### Proposed Changes

Remove underscores from private property and method names. Fix uses in tests. Note that the tests are still accessing private properties. I decided not to add getters that would only exist for tests.

Stop setting properties to `null` in `dispose` when not needed.

#### Behavior Before Change

No change

### Reason for Changes

Match current google style guide.

### DIFF
--- a/core/workspace_comment_svg.ts
+++ b/core/workspace_comment_svg.ts
@@ -62,36 +62,37 @@ export class WorkspaceCommentSvg extends WorkspaceComment implements
   override workspace: WorkspaceSvg;
 
   /** Mouse up event data. */
-  private onMouseUpWrapper_: browserEvents.Data|null = null;
+  private onMouseUpWrapper: browserEvents.Data|null = null;
 
   /** Mouse move event data. */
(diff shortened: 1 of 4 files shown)


core/workspace_comment_svg.ts
```diff
@@ -905,56 +905,56 @@ export class WorkspaceCommentSvg extends WorkspaceComment implements
...
-  private disposeInternal_() {
-    this.textarea_ = null;
-    this.foreignObject_ = null;
-    this.svgRectTarget_ = null;
-    this.svgHandleTarget_ = null;
+  private disposeInternal() {
+    this.textarea = null;
+    this.foreignObject = null;
+    this.svgRectTarget = null;
+    this.svgHandleTarget = null;
```
BeksOmega, line 957: I don't think any of these nullifications are necessary. 🤞

BeksOmega (approved):
One comment about disposal, otherwise LGTM!

PR #6977, opened 2023-04-13
chore: remove underscores from private properties and methods in some fields

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Part of https://github.com/google/blockly/issues/6548

### Proposed Changes
Remove underscores from private property and method names. Fix uses in tests. In some cases, update tests to use existing getters.

### DIFF
--- a/tests/mocha/field_image_test.js
+++ b/tests/mocha/field_image_test.js
@@ -94,23 +94,23 @@ suite('Image Fields', function() {
       });
       test('JS Constructor', function() {
         const field = new Blockly.FieldImage('src', 10, 10, null, this.onClick);
-        chai.assert.equal(field.clickHandler_, this.onClick);
+        chai.assert.equal(field.clickHandler, this.onClick);
       });
       test('setOnClickHandler', function() {
         const field = new Blockly.FieldImage('src', 10, 10);
         field.setOnClickHandler(this.onClick);
-        chai.assert.equal(field.clickHandler_, this.onClick);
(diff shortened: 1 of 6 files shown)


tests/mocha/field_image_test.js
```diff
@@ -94,23 +94,23 @@ suite('Image Fields', function() {
...
         field.setOnClickHandler(this.onClick);
-        chai.assert.equal(field.clickHandler_, this.onClick);
+        chai.assert.equal(field.clickHandler, this.onClick);
       });
       test('Remove Click Handler', function() {
         const field = new Blockly.FieldImage('src', 10, 10, null, this.onClick);
         field.setOnClickHandler(null);
-        chai.assert.isNull(field.clickHandler_);
+        chai.assert.isNull(field.clickHandler);
       });
```
BeksOmega, line 108: Honestly not sure if these tests are worth maintaining given that `setOnClickHandler` literally just assigns a variable. (costs of doing updates like this might not be worth the benefits)

BeksOmega (approved):
LGTM! Just one question about whether the click handler tests are doing anything for us.

tests/mocha/field_image_test.js
rachel-fenichel: Yeah. It's testing all of the arguments, but not in a super useful way.

I've been taking notes as I changed fields, and this is a pattern across the field tests. I think the solution is to test that the click handler is actually getting called (possibly using the stuff Eric is working on?) and then delete the tests that check whether the property was set.

Issue #6988, opened 2023-04-18
block_svg: render isn't sync OK with the state after disable/enable ops


dimaaan21:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

There is a render mistake after disable/enable operations: if there is a disabled block at some level, all blocks views below it will unsync with the state

### Reproduction steps

1. Open playground (https://blockly-demo.appspot.com/static/tests/playground.html)
2. Paste in attached JSON to the textarea (or create hierarchy `if => [repeat, for, whileUntil, forEach]` manually)
3. Click "Load"
4. `if`-block: right-click => "Disable block"
5. `if`-block: right-click => "Enable block"
6. You'll see that blocks after `for`-block will be rendered as disabled but they're not (check JSON) – render bug
7. Reload page. Click "Load" again. Everything is OK

```json
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "controls_if",
        "id": "iB$?M1srl.oI.0f!0MJ6",
        "x": 13,
        "y": 13,
        "inputs": {
          "DO0": {
            "block": {
              "type": "controls_repeat_ext",
              "id": "IIi|I1fo%;+hBa-{~;t9",
              "inputs": {
                "TIMES": {
                  "shadow": {
                    "type": "math_number",
                    "id": "fD1HEU9xtQ.6#WWC7P=-",
                    "fields": {
                      "NUM": 10
                    }
                  }
                }
              },
              "next": {
                "block": {
                  "type": "controls_for",
                  "id": "Mg;C+1sKjF~dn1YOJVeV",
                  "enabled": false,
                  "fields": {
                    "VAR": {
                      "id": "+a5j@W.tyltn,x,_Ihmy"
                    }
                  },
                  "inputs": {
                    "FROM": {
                      "shadow": {
                        "type": "math_number",
                        "id": "N4^C~zr1!^~:{:^Ih+pP",
                        "fields": {
                          "NUM": 1
                        }
                      }
                    },
                    "TO": {
                      "shadow": {
                        "type": "math_number",
                        "id": "^^ev;%@.D}86**fl_[bc",
                        "fields": {
                          "NUM": 10
                        }
                      }
                    },
                    "BY": {
                      "shadow": {
                        "type": "math_number",
                        "id": "(Xka3$h[8tN7N7eM@*?a",
                        "fields": {
                          "NUM": 1
                        }
                      }
                    }
                  },
                  "next": {
                    "block": {
[...]

PR #6989, opened 2023-04-18
fix: block_svg disable/enable ops lead to unsync state and render

dimaaan21:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/6988


BeksOmega (commented):
Thanks for finding this bug! This fix looks like it works great :D But I have a recommendation for a slightly different approach.

It looks like the issue is that we need to update next blocks regardless of whether the current block has the proper `visuallyDisabled` state or not. But, we can still skip rendering all of the /inner/ children of a block to gain a little bit of performance.

So I think this would look something like:
```
if (this.visuallyDisabled === disabled) {
  this.getNextBlock()?.updateDisabled();
  return;
}
```

Does that look like it'll solve the bug(s) you're running into?

BeksOmega (approved):

PR #6990, opened 2023-04-19
feat: add support for appending custom inputs

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #2062 

### Proposed Changes

Adds an `appendInput` function, and support for constructing inputs from the registry.

### Reason for Changes

Support for custon inputs!

### Test Coverage

Added a test for constructing custom inputs from JSON.

### Documentation

N/A

### Additional Information

I reorganized some tests. Relevant changes are in commit 3e6db57 =)

I'd like to consider this the minimum viable API for handling custom inputs. It has several limitations:
  * No `fromJson` method for constructing inputs, meaning custom inputs cannot be further parameterized when constructing from a JSON block definition.
  * We force external developers to construct their own connections if they want them, which may be what we want, but signalling that by passing `null` to the connection constructor parameter is not ideal.
  * Our built-in inputs are still differentiated using a `type` property, which creates inconsistency between the built-in inputs and custom ones. We also have to pass an enum to the `Input` constructor, which feels kind of jank.

### DIFF
--- a/core/input_types.ts
+++ b/core/input_types.ts
@@ -19,5 +19,7 @@ export enum inputTypes {
   // A down-facing block stack.  E.g. 'if-do' or 'else'.
   STATEMENT = ConnectionType.NEXT_STATEMENT,
   // A dummy input.  Used to add field(s) with no input.
-  DUMMY = 5
+  DUMMY = 5,
+  // An unknown type of input defined by an external developer.
+  CUSTOM = 5
 }
--- a/core/input_types.ts
+++ b/core/input_types.ts
@@ -21,5 +21,5 @@ export enum inputTypes {
   // A dummy input.  Used to add field(s) with no input.
   DUMMY = 5,
   // An unknown type of input defined by an external developer.
-  CUSTOM = 5
+  CUSTOM = 6,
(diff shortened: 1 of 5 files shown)


core/input_types.ts
```diff
@@ -19,5 +19,7 @@ export enum inputTypes {
   // A down-facing block stack.  E.g. 'if-do' or 'else'.
   STATEMENT = ConnectionType.NEXT_STATEMENT,
   // A dummy input.  Used to add field(s) with no input.
-  DUMMY = 5
+  DUMMY = 5,
+  // An unknown type of input defined by an external developer.
+  CUSTOM = 5
```
NeilFraser, line 24: Maybe add a comment explaining why DUMMY and CUSTOM are both 5.
Also, are trailing commas are legal with TypeScript enums.  Isn't that desirable?

NeilFraser (approved):

core/input_types.ts
BeksOmega: Oh whoops actually meant to make it 6, thanks for the catch!

PR #6994, opened 2023-04-19
chore: clean up some drawer code lint

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

General cleanup: pull out some values in drawer code to avoid stacking ternaries too deeply.

### DIFF
--- a/core/renderers/zelos/drawer.ts
+++ b/core/renderers/zelos/drawer.ts
@@ -100,18 +100,17 @@ export class Drawer extends BaseDrawer {
       const precedesStatement = spacerRow.precedesStatement;
       const followsStatement = spacerRow.followsStatement;
       if (precedesStatement || followsStatement) {
-        const insideCornersObject =
-            this.constants_.INSIDE_CORNERS as InsideCorners;
-        const cornerHeight = insideCornersObject.rightHeight;
+        const insideCorners = this.constants_.INSIDE_CORNERS as InsideCorners;
+        const cornerHeight = insideCorners.rightHeight;
         const remainingHeight =
             spacerRow.height - (precedesStatement ? cornerHeight : 0);
         const bottomRightPath =
-            followsStatement ? insideCornersObject.pathBottomRight : '';
+            followsStatement ? insideCorners.pathBottomRight : '';
         const verticalPath = remainingHeight > 0 ?
             svgPaths.lineOnAxis('V', spacerRow.yPos + remainingHeight) :
             '';
         const topRightPath =
-            precedesStatement ? insideCornersObject.pathTopRight : '';
+            precedesStatement ? insideCorners.pathTopRight : '';
         // Put all of the partial paths together.
         this.outlinePath_ += bottomRightPath + verticalPath + topRightPath;
         return;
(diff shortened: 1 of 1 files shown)


core/renderers/zelos/drawer.ts
```diff
@@ -97,23 +97,23 @@ export class Drawer extends BaseDrawer {
     }
     if (Types.isSpacer(row)) {
       const spacerRow = row as SpacerRow;
-      if (spacerRow.precedesStatement || spacerRow.followsStatement) {
-        const cornerHeight =
-            (this.constants_.INSIDE_CORNERS as InsideCorners).rightHeight;
+      const precedesStatement = spacerRow.precedesStatement;
+      const followsStatement = spacerRow.followsStatement;
+      if (precedesStatement || followsStatement) {
+        const insideCornersObject =
```
maribethb, line 103: nit: `insideCorners` instead of `insideCornersObject` (object is implied from context / no hungarian-like notation) (here and below)

Also, are you sure this cast is necessary? `this.constants_.INSIDE_CORNERS` [seems](https://github.com/google/blockly/blob/develop/core/renderers/common/constants.ts#L376) like it should already be of the correct type?

maribethb (approved):

core/renderers/zelos/drawer.ts
rachel-fenichel: Oh, you're gonna hate this. It's the _common_ `InsideCorners` type, and this is casting it to the _zelos_ `InsideCorners` type.

maribethb: I do in fact hate this. I messed around in the playground and figured out [how](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqD2BPaBvAvgbgFBdJGAQvAHYCm0pAHgC6nEAmMCyKW070wAFqQO4AUASlSYso3OCjQAgsQCWAWzAhUbDtSQAuOEmxr2wEhGoAnAK7ANJoao53o1LnIgA6DcgC80Mrx2Ih2HZoYjh4UgAi8ADmFDR0jDLySiqsdgD0aXBOMM4OPNAADnKkwOS8pADkJuQKzhByxFH60BkULlEu0PAAbqQmJnL05IbExmDE1BAA+tAN0ABepCDwEGn0JmDlJgA00OP00ACSAHIAyofhAKJTAMIA8gBKx5cPp7PEC0sraSNjExDNVoAKzMxi6vX6g3IjlICgc8EK8Aa1HheRq8Gq0AgBRKcgAZnJgA5EDjmj0+gMhvDENoiGQ9HZfqYLFYbKl7OwIGYcdZBIEOXlnG4kNAvD5oHTSAFmsFmgVwP5hOyOY4he4XNw+NKgiF2EA) to avoid it. But sadly most of the other casts in this file can't be cleaned up the same way because they're not just casting things that are overridden within the renderer, they're like... overriding properties on blocks and inputs and we don't have `ZelosBlock` etc in which we can override things. So I'm disinclined to clean this up myself but feel free if you want :p

maribethb (approved):

PR #6996, opened 2023-04-20
feat: Add 'reason' field to move event

NeilFraser:
There are many types of move.  This addition allows one to detect what the reason for each move is.

### DIFF
--- a/core/events/events_block_move.ts
+++ b/core/events/events_block_move.ts
@@ -76,9 +76,9 @@ export class BlockMove extends BlockBase {
    *  'disconnect' -- Block got disconnected from another block.
    *  'create' -- Block created via XML.
    *  'cleanup' -- Workspace aligned top-level blocks.
-   * Reasons may become comma separated if move events merge ('drag,bump,snap').
+   * Event merging may create multiple reasons: ['drag', 'bump', 'snap'].
    */
-  reason?: string;
+  reason?: string[];
 
   /** @param opt_block The moved block.  Undefined for a blank event. */
   constructor(opt_block?: Block) {
@@ -204,7 +204,7 @@ export class BlockMove extends BlockBase {
    *
    * @param reason Why is this move happening?  'drag', 'bump', 'snap', ...
    */
-  setReason(reason: string) {
+  setReason(reason: string[]) {
     this.reason = reason;
   }
 
@@ -318,7 +318,7 @@ export interface BlockMoveJson extends BlockBaseJson {
   newParentId?: string;
   newInputName?: string;
   newCoordinate?: string;
-  reason?: string;
+  reason?: string[];
   recordUndo?: boolean;
 }
 
--- a/core/events/events_block_move.ts
+++ b/core/events/events_block_move.ts
@@ -61,10 +61,24 @@ export class BlockMove extends BlockBase {
   newInputName?: string;
 
   /**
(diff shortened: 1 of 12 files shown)


core/events/events_block_move.ts
```diff
@@ -61,11 +61,25 @@ export class BlockMove extends BlockBase {
...
+   *  'bump' -- Block got bumped away from an invalid connection.
+   *  'snap' -- Block got shifted to line up with the grid.
+   *  'inbounds' -- Block got pushed back into a non-scrolling workspace.
+   *  'connect' -- Block got connected to another block.
+   *  'disconnect' -- Block got disconnected from another block.
+   *  'create' -- Block created via XML.
+   *  'cleanup' -- Workspace aligned top-level blocks.
+   * Reasons may become comma separated if move events merge ('drag,bump,snap').
+   */
+  reason?: string;
```
maribethb, line 81: Can we make the list of reasons an enum? It would mean we get compile-time errors if we e.g. misspell `inbounds` or try to add a variation like `in-bounds`, plus give us autocomplete.

Also, can we have the reason property in the event be an array of reasons, instead of a single string with commas? That would be simpler for consumers of this property instead of having to parse the string into an array themselves.

core/events/events_block_move.ts
```diff
@@ -61,11 +61,25 @@ export class BlockMove extends BlockBase {
...
 
+  /**
+   * An explanation of what this move is for.  Known values include:
+   *  'drag' -- A drag operation completed.
+   *  'bump' -- Block got bumped away from an invalid connection.
+   *  'snap' -- Block got shifted to line up with the grid.
+   *  'inbounds' -- Block got pushed back into a non-scrolling workspace.
+   *  'connect' -- Block got connected to another block.
+   *  'disconnect' -- Block got disconnected from another block.
+   *  'create' -- Block created via XML.
```
maribethb, line 77: nit: `Block created via deserialization`? (since I assume this would also apply for blocks created json?)

core/events/utils.ts
```diff
@@ -290,6 +290,17 @@ export function filter(queueIn: Abstract[], forward: boolean): Abstract[] {
         lastEvent.newParentId = moveEvent.newParentId;
         lastEvent.newInputName = moveEvent.newInputName;
         lastEvent.newCoordinate = moveEvent.newCoordinate;
+        if (moveEvent.reason) {
+          if (lastEvent.reason) {
+            // Contatenate reasons into a comma separated string
```
maribethb, line 295: nit: Concatenate (if this comment still exists pending comment re: using an array)

core/events/events_block_move.ts
NeilFraser: I assumed the same, but there are no move events fired when deserializing from JSON, just when using XML.

core/events/events_block_move.ts
NeilFraser: I don't think an enum would allow developers to create new reasons.  Imagine an extension or application which, say, shuffled blocks around randomly to create [a puzzle](https://blockly.games/puzzle).  They should be able to create a custom reason (e.g. 'shuffle').

Changed all reasons to be arrays.

core/events/utils.ts
NeilFraser: Done.

maribethb (approved):

PR #7003, opened 2023-04-21
refactor(blocks): Migrate `blocks/logic.js` to TypeScript

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6828.

### Additional Info

Holding off on clang-formatting as it makes some very questionable choices, which I have raised a query about.


blocks/logic.ts
```diff
@@ -304,100 +314,107 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
...
+        }
+        const state = Object.create(null);
+        if (this.elseifCount_) {
+          state['elseIfCount'] = this.elseifCount_;
+        }
+        if (this.elseCount_) {
+          state['hasElse'] = true;
+        }
+        return state;
+      },
```
BeksOmega, line 365: Was this formatting intentional?

blocks/logic.ts
```diff
@@ -304,100 +314,107 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
...
    * Populate the mutator's dialog with this block's components.
-   * @param {!Workspace} workspace Mutator's workspace.
-   * @return {!Block} Root block in mutator.
-   * @this {Block}
+   *
+   * @param workspace Mutator's workspace.
+   * @return Root block in mutator.
    */
-  decompose: function(workspace) {
+  decompose: function(this: IfBlock, workspace: Workspace): Block {
```
BeksOmega, line 384: Should this return a `ContainerBlock` rather than a `Block`?

blocks/logic.ts
```diff
@@ -605,43 +626,43 @@ const LOGIC_COMPARE_ONCHANGE_MIXIN = {
...
  * "logic_compare" extension function. Adds type left and right side type
  * checking to "logic_compare" blocks.
- * @this {Block}
- * @readonly
  */
-const LOGIC_COMPARE_EXTENSION = function() {
+const LOGIC_COMPARE_EXTENSION = function(this: CompareBlock) {
   // Add onchange handler to ensure types are compatible.
   this.mixin(LOGIC_COMPARE_ONCHANGE_MIXIN);
 };
```
BeksOmega, line 633: Unrelated to types but: is this done to work around issues with mixing in methods that override built-in methods? We might want to add a comment documenting that.

blocks/logic.ts
```diff
@@ -291,11 +283,29 @@ Extensions.register(
...
+
+// Types for quarks defined in JSON.
+/** Type of a controls_if_if (if mutator container) block. */
+interface ContainerBlock extends Block {}
+
+/** Type of a controls_if_elseif or controls_if_else block. */
+interface ClauseBlock extends Block {
+  valueConnection_: Connection|null;
+  statementConnection_: Connection|null;
+}
```
BeksOmega, line 299: Should this technically be `Connection|null|undefined`?

blocks/logic.ts
```diff
@@ -550,31 +566,36 @@ const CONTROLS_IF_TOOLTIP_EXTENSION = function() {
 
 Extensions.register('controls_if_tooltip', CONTROLS_IF_TOOLTIP_EXTENSION);
 
+/** Type of a block that has LOGIC_COMPARE_ONCHANGE_MIXIN */
+type CompareBlock = Block&CompareMixin;
+interface CompareMixin extends CompareMixinType {
+  prevBlocks_: Array<Block|null>;
```
BeksOmega, line 572: I think this should also be `|undefined`

blocks/logic.ts
```diff
@@ -304,100 +314,107 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
...
+   *
+   * @param xmlElement XML storage element.
    */
-  domToMutation: function(xmlElement) {
-    this.elseifCount_ = parseInt(xmlElement.getAttribute('elseif'), 10) || 0;
-    this.elseCount_ = parseInt(xmlElement.getAttribute('else'), 10) || 0;
+  domToMutation: function(this: IfBlock, xmlElement: Element) {
+    const elseifAttribute = xmlElement.getAttribute('elseif');
+    if (elseifAttribute === null) {
+      throw new TypeError('xmlElement did not elseif attribute');
```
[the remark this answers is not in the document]
cpcallen: Guard clause completely removed, in line with plan to keep `tsc` output as close as possible to original `.js` files in first pass.

blocks/logic.ts
```diff
@@ -304,100 +314,107 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
...
+    }
+    this.elseifCount_ = parseInt(elseifAttribute, 10) || 0;
+    this.elseCount_ = parseInt(xmlElement.getAttribute('else')!, 10) || 0;
```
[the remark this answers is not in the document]
cpcallen: It's just asserting that `getAttribute()` does not return `null` or `undefined` to satisfy the signature of `parseInt`.  If `getAttribute` does return `null` or `undefined` then that gets passed to `parseInt` just as before (and of course `parseInt` will handle that fine, at least for certain definitions of fine that include "works as well as it did before the migration").

blocks/logic.ts
cpcallen: No.  Probably clang-format being a bit flighty.  We're not prettifying this directory (yet) so I've manually reformatted it to something more reasonable.

blocks/logic.ts
cpcallen: Good idea.  Done.

blocks/logic.ts
```diff
@@ -304,100 +314,107 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
...
+    let clauseBlock =
+        containerBlock.nextConnection!.targetBlock() as ClauseBlock | null;
     // Count number of inputs.
     this.elseifCount_ = 0;
     this.elseCount_ = 0;
-    const valueConnections = [null];
-    const statementConnections = [null];
-    let elseStatementConnection = null;
+    const valueConnections: Array<Connection|null> = [null];
+    const statementConnections: Array<Connection|null> = [null];
```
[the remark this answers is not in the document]
cpcallen: I'd rather not make any functional changes to the code, and I'm honestly not sure why it's written this way.  :-/  It does deserve a comment, though; maybe once I've had some tea I will be able to figure out what's going on here.

blocks/logic.ts
```diff
@@ -526,15 +541,16 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
 };
 
 Extensions.registerMutator(
-    'controls_if_mutator', CONTROLS_IF_MUTATOR_MIXIN, null,
+    'controls_if_mutator', CONTROLS_IF_MUTATOR_MIXIN,
+    null as unknown as undefined,  // BUG(#6920)
```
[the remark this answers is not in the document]
cpcallen: Huh.  Usually I only use BUG if there's actually a bug, and it's not clear that's the case here, so I'm not sure what past cpcallen was doing.

blocks/logic.ts
cpcallen: There's no problem overriding built-in methods with a JS-style block definition (i.e., a mixin that gets its properties copied on the the `Block` instance).  But the block that uses the `LOGIC_COMPARE_ONCHANGE_MIXIN` is defined using JSON, and I gather that the way we provided to mix arbitrary extra properties onto JSON-defined blocks is to use an extension like this.  (Which is probably sensible, since the extension can do all kinds of other things to the block, too.)

blocks/logic.ts
cpcallen: Hmm.  Looks like it, though I've annotated them as optional properties, which are equivalent (`valueConnection_?: Connection|null`) because that seems slightly clearer.  Unfortunately this causes a considerable number of knock-on type changes.  :-/

blocks/logic.ts
cpcallen: Agreed—but again, marked optional via `prevBlocks_?:` rather than `|undefined`, for clarity.

blocks/logic.ts
cpcallen: (OK, not _that_ considerable.)

cpcallen:
I believe this is now ready for re-review.

BeksOmega (approved):

PR #7020, opened 2023-04-27
chore: clean up lint in horizontal flyout

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

Cleans up `!` in the horizontal flyout in preparation for extracting the whole thing as a plugin. The style guide does not allow `!` in blockly-samples. Most of the fixes involved casting to `BlockSvg` or `WorkspaceSvg`.

#### Behavior Before Change

No changes

### Test Coverage
Checked that the horizontal flyout still works in the advanced playground.

### DIFF
--- a/core/flyout_horizontal.ts
+++ b/core/flyout_horizontal.ts
@@ -12,6 +12,7 @@
 import * as goog from '../closure/goog/goog.js';
 goog.declareModuleId('Blockly.HorizontalFlyout');
 
+import type {BlockSvg} from './block_svg.js';
 import * as browserEvents from './browser_events.js';
 import * as dropDownDiv from './dropdowndiv.js';
 import {Flyout, FlyoutItem} from './flyout_base.js';
@@ -23,6 +24,7 @@ import type {Coordinate} from './utils/coordinate.js';
 import {Rect} from './utils/rect.js';
 import * as toolbox from './utils/toolbox.js';
 import * as WidgetDiv from './widgetdiv.js';
+import type {WorkspaceSvg} from './workspace_svg.js';
 
 
 /**
@@ -82,7 +84,8 @@ export class HorizontalFlyout extends Flyout {
     if (!this.isVisible()) {
       return 0;
     }
-    const metricsManager = this.targetWorkspace!.getMetricsManager();
+    const targetWorkspace = this.targetWorkspace as WorkspaceSvg;
+    const metricsManager = targetWorkspace.getMetricsManager();
     const absoluteMetrics = metricsManager.getAbsoluteMetrics();
     const viewMetrics = metricsManager.getViewMetrics();
     const toolboxMetrics = metricsManager.getToolboxMetrics();
@@ -91,10 +94,10 @@ export class HorizontalFlyout extends Flyout {
     const atTop = this.toolboxPosition_ === toolbox.Position.TOP;
     // If this flyout is not the trashcan flyout (e.g. toolbox or mutator).
     // Trashcan flyout is opposite the main flyout.
-    if (this.targetWorkspace!.toolboxPosition === this.toolboxPosition_) {
+    if (targetWorkspace.toolboxPosition === this.toolboxPosition_) {
       // If there is a category toolbox.
       // Simple (flyout-only) toolbox.
-      if (this.targetWorkspace!.getToolbox()) {
+      if (targetWorkspace.getToolbox()) {
         if (atTop) {
           y = toolboxMetrics.height;
         } else {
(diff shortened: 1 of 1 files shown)


core/flyout_horizontal.ts
```diff
@@ -236,31 +244,31 @@ export class HorizontalFlyout extends Flyout {
 
     for (let i = 0, item; item = contents[i]; i++) {
       if (item.type === 'block') {
-        const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        const block = item.block as BlockSvg;
```
maribethb, line 247: Alternatively/safer:

if (!item.block) throw new Error(flyout item with type block must contain property block)

core/flyout_horizontal.ts
```diff
@@ -309,18 +317,18 @@ export class HorizontalFlyout extends Flyout {
     }
 
     const flyoutRect = this.svgGroup_.getBoundingClientRect();
-    // BIG_NUM is offscreen padding so that blocks dragged beyond the shown
+    // bigNum is offscreen padding so that blocks dragged beyond the shown
     // flyout area are still deleted.  Must be larger than the largest screen
     // size, but be smaller than half Number.MAX_SAFE_INTEGER (not available on
     // IE).
-    const BIG_NUM = 1000000000;
+    const bigNum = 1000000000;
```
maribethb, line 324: did you change this because the linter doesn't allow constant case in samples? if so, we should probably update that...

on the other hand, this doesn't meet the [requirements](https://google.github.io/styleguide/tsguide.html#identifiers-constants) of using constant case so it should be lower case anyway

core/flyout_horizontal.ts
```diff
@@ -82,7 +84,8 @@ export class HorizontalFlyout extends Flyout {
     if (!this.isVisible()) {
       return 0;
     }
-    const metricsManager = this.targetWorkspace!.getMetricsManager();
+    const targetWorkspace = this.targetWorkspace as WorkspaceSvg;
```
maribethb, line 87: do we even have flyouts in headless mode? i wonder if it makes more sense for the base flyout class to have the `targetWorkspace` use `WorkspaceSvg` instead of `Workspace`?

maribethb (approved):

core/flyout_horizontal.ts
rachel-fenichel: I changed it because it doesn't meet the requirements and it's been bugging me.

core/flyout_horizontal.ts
rachel-fenichel: I think this is actually validated in an earlier parsing step. There may be a better typing solution here so that we know that if type is `''block'`, it has the property `block`.

core/flyout_horizontal.ts
maribethb: cool, makes sense. i had to go look up the rules because i was worried the linter wasn't allowing it

core/flyout_horizontal.ts
maribethb: yeah, I think you can use the type guard method strategy:

```
function isBlock(item: FlyoutItem): item is BlockSvg {
  return item.type === 'block';
}
```


then check `isBlock(item)` instead of `item.type === block`

but after writing all that out, i wonder why bother checking the type at all, and just check `if item.block` directly?

rachel-fenichel: You mean with an instanceof check? I guess we could, as long as we assume that there will never be other cases where you have an item that has a block in it. Which is probably safe, or at least something we could deal with if/when it becomes an issue.

rachel-fenichel: Fair point though, that if the goal is to actually have type safety, casually casting doesn't do it.

core/flyout_horizontal.ts
rachel-fenichel: It is a `WorkspaceSvg` in the base flyout class. This is about whether it's `null`, not about whether it's a `Workspace` or a `WorkspaceSvg`.

core/flyout_horizontal.ts
maribethb: I don't think we need an instanceof check. Initially I thought we could use a type guard but I don't think that works because we need to assert `item.block` is a `BlockSvg`, not `item`.

I'm saying instead of checking `if (item.type === 'block')`, just check `if (item.block)`. If `item.block` is not null, we already know it must be a `BlockSvg` (without using an instanceof check, just because of the type of `FlyoutItem`). So I don't think we need the `item.type` at all here. The only two options for the type are button and block so I don't think we need to worry about other types also having a block property.

core/flyout_horizontal.ts
maribethb: IMO using `!` to assert non-null is just a special type of casting. If that's the only reason we're casting, I would prefer to just use `!` directly, as it's more clear about the intention, and tighter because it doesn't let you cast just whatever type into `WorkspaceSvg`. It only lets you remove `null` from the possible types. So I think using `!` is better than the general cast.

Therefore if we can't remove them, I would rather add them to the allowed style than change them to use looser casts.

Issue #6980, opened 2023-04-14
Document the rendering life cycle


BeksOmega:
### Location

A new Architecture section of the docs: https://github.com/google/blockly/issues/7236

### Type

- [X] Text
- [X] Image or Gif
- [X] Other

### Suggested content

Write up a doc explaining the rendering order (both for immediate rendering and queued rendering). Because right now the logic lives in lots of files and requires a lot of context.

### Additional context

N/A

PR #7024, opened 2023-04-27
fix: switch most remaining render calls to queueRender

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Was looking at https://github.com/google/blockly/issues/6980 and realized it would be easier to document if we just have one rendering path rather than two :P So I'm looking at trying to get rid of the old one.

### Proposed Changes

Looked through all of the remaining uses of `render` and switched any over to `queueRender` that could use `queueRender`.

Remaining uses of `render` after this PR include icons (because they need to know their locations immediately) and deserialization (because it's causing forced layouts that I need to investigate).

### Reason for Changes

Unification of rendering paths.

### Test Coverage

Manual poking at things.

### Documentation

N/A - but soon! hopefully 🤞 

### Additional Information

The plan for the icons is to add a function to the render management system that forces an immediate render of all queued renders.

### DIFF
--- a/blocks/procedures.js
+++ b/blocks/procedures.js
@@ -756,9 +756,6 @@ const PROCEDURE_CALL_COMMON = {
       this.quarkConnections_ = {};
       this.quarkIds_ = [];
     }
-    // Switch off rendering while the block is rebuilt.
-    const savedRendered = this.rendered;
-    this.rendered = false;
     // Update the quarkConnections_ with existing connections.
     for (let i = 0; i < this.arguments_.length; i++) {
       const input = this.getInput('ARG' + i);
@@ -797,11 +794,6 @@ const PROCEDURE_CALL_COMMON = {
           }
         }
       }
-    }
-    // Restore rendering and show the changes.
-    this.rendered = savedRendered;
-    if (this.rendered) {
-      this.render();
     }
   },
   /**
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1000,6 +1000,8 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       this.comment = null;  // For backwards compatibility.
     }
     if (this.rendered) {
+      // Icons must force an immediate render so that bubbles can be opened
+      // immedately at the correct position.
       this.render();
       // Adding or removing a comment icon will cause the block to change shape.
       this.bumpNeighbours();
@@ -1078,6 +1080,8 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       }
     }
     if (changedState && this.rendered) {
+      // Icons must force an immediate render so that bubbles can be opened
+      // immedately at the correct position.
       this.render();
       // Adding or removing a warning icon will cause the block to change shape.
       this.bumpNeighbours();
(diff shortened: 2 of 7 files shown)


core/block_svg.ts
```diff
@@ -1000,6 +1000,8 @@ export class BlockSvg extends Block implements IASTNodeLocationSvg,
       this.comment = null;  // For backwards compatibility.
     }
     if (this.rendered) {
+      // Icons must force an immediate render so that bubbles can be opened
```
maribethb, line 1003: Can we instead maybe add the comment icon in the `afterQueuedRenders` or whatever we called it? I don't actually know what I'm talking about here so feel free to ignore me if this is nonsense :p

otherwise: nit: (here and below) immediately typo

core/rendered_connection.ts
```diff
@@ -514,12 +514,7 @@ export class RenderedConnection extends Connection {
       return;
     }
     blockShadow.initSvg();
-    blockShadow.render(false);
-
-    const parentBlock = this.getSourceBlock();
-    if (parentBlock.rendered) {
-      parentBlock.queueRender();
-    }
+    blockShadow.queueRender();
```
maribethb, line 517: just double checking, will this also cause the parent block to be queued and that's why we got rid of the code below this?

blocks/procedures.js
```diff
@@ -798,11 +795,6 @@ const PROCEDURE_CALL_COMMON = {
         }
       }
     }
-    // Restore rendering and show the changes.
-    this.rendered = savedRendered;
-    if (this.rendered) {
-      this.render();
```
maribethb, line 804: do we not need to queue a render now?

maribethb (approved):
mostly questions so i understand how rendering works!

core/block_svg.ts
BeksOmega: Wrt your comment about `finishQueuedRenders`, that doesn't solve for this use case because the icon needs to be appended to the block before the render is triggered so the renderer can position it correctly.

But now that I'm thinking about this again, I'm going to investigate this one more time.

I was trying to solve for the use case of:
```
myBlock.setCommentText('yay a comment');
myBlock.getCommentIcon().setVisible(true);
```

I thought that for this to correctly position the bubble, the `setCommentText` method must:
  1) Append the comment to the block.
  2) Rerender the block to position the icon.
  3) Tell the icon what its new position is.

So that when `setVisible` is called, the icon has the correct information to position the bubble.

But thinking about it again, if we use the render queueing it should look like:
  1) Append the comment to the block.
  2) Open the bubble at the incorrect position.
  3) Rerender the block to position the icon.
  4) Tell the icon what it new position is.
  5) Update the position of the bubble for the new position of the icon.
  6) Actually paint the frame to the screen.

This should still be correct. But it was causing some unit tests to fail, so I need to remember why that was.

BeksOmega:
@maribethb I tried wrapping the `setVisible` calls so that the logic is triggered after the current render completes, but that means that external developers won't be able to access the contents of the bubble immediately (e.g. the mutator workspace).

I don't particularly like that breaking change, but on the other hand I don't really like external developers messing with the contents of the bubble anyway? So wanted to confirm before I go fix the 40 remaining tests :P

PR #7052, opened 2023-05-04
Prototype functional automated test 

ericblackmonGoogle:
## The basics

- [X ] I branched from develop

### Resolves

This is just a prototype for how our team might functionally test our features

### Proposed Changes

[Link](https://docs.google.com/document/d/1LYL6sJUQlM74XiSmAyRVadY1oMOwfqmsSpvn3CU9Khs/edit?resourcekey=0-ZU-Oy-In0vuAwFnRFwD3sQ#) to desgin doc

#### Behavior After Change

Some of blockly functionality is tested 

### Reason for Changes

To prototype functional testing for future features

### Test Coverage

These are test 

### Documentation

[Link](https://docs.google.com/document/d/1LYL6sJUQlM74XiSmAyRVadY1oMOwfqmsSpvn3CU9Khs/edit?resourcekey=0-ZU-Oy-In0vuAwFnRFwD3sQ#) to desgin doc

### Additional Information

N/A


tests/browser/test/basic_test.js
```diff
@@ -0,0 +1,117 @@
...
+    console.log('Starting webdriverio...');
+    browser = await webdriverio.remote(options);
+    console.log('Loading URL: ' + url);
+    await browser.url(url);
+    return browser;
+  });
+
+
+test('Drag Block Event', async function() {
+  const block = await browser.$('[data-id="test_block"]');
```
BeksOmega, line 57: Love being able to query based on the data-id 🗡️ 

I think we should consider making the IDs of blocks in flyouts consistent with the IDs of blocks they create in the workspace (e.g. generating a new flyout block with a new ID every time one is "used" to create a new block in the workspace), so that we can also use the data-id in flyout tests.

This would be a change in core though, and idk if it would work with block recycling.

BeksOmega (commented):
So cool :D Thank you for putting up this demo!

Also discussed offline: it would be really nice if we could do assertions against the state of the workspace rather than the state of the dom. We're going to play around with [execute](https://webdriver.io/docs/api/browser/execute/) and see if we can get something working!

tests/browser/test/basic_test.js
```diff
@@ -0,0 +1,117 @@
...
+ * Copyright 2019 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Node.js script to run Automated tests in Chrome, via webdriver.
+ */
+/* eslint-env node */
+/* eslint-env mocha */
+/* eslint-env webdriverio */
```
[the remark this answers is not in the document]
ericblackmonGoogle: Perfect that is a lot easier than doing it for every file!

tests/browser/test/basic_test.js
```diff
@@ -0,0 +1,117 @@
...
+test('Clicking on a Field', async function() {
+  const block = await browser.$('[data-id="test_block"]');
+  block.click();
+  const editText = await browser.$('#blocklyDiv > div > svg.blocklySvg > g > g.blocklyBlockCanvas > g.blocklyDraggable.blocklySelected > g.blocklyEditableText.editing');
+  const editTextClass = await editText.getAttribute('class');
+  chai.assert.isDefined(editTextClass.length);
+});
+
+test('Typing into a field', async function() {
+  await new Promise((resolve) => setTimeout(resolve, 2000)); // 2sec
```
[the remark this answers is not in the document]
ericblackmonGoogle: The DOM takes sometimes takes time settle before you can do new actions. When I noticed that happening I just put in a two second wait to slow it down a little.

tests/browser/test/basic_test.js
ericblackmonGoogle: That would be super helpful from a testing perspective!

tests/browser/test/basic_test.js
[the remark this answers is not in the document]
BeksOmega: Hm do you know if this is a Blockly specific problem? or have you seen other exmpales of using webdriver that also run into this?

I have an instinct that this is related to the new render queueing stuff, which only rerenders blocks at the end of the frame (for performance improvements).

tests/browser/test/connecting_blocks_test.js
```diff
@@ -0,0 +1,112 @@
...
+    const blocksConnected = await browser.execute(() => {
+      const parentBlock = Blockly.getMainWorkspace().getTopBlocks(false)[1];
+      if (childBlock.getChildren()) {
+        return true;
+      } else {
+        return false;
+      }
+   });
+
+    chai.assert.isTrue(blocksConnected);
```
BeksOmega, line 89: This is very cool! I like asserting based on the workspace state rather than the dom state a lot better.

When we finalize this, my preference is that this is the approach we go with!

tests/browser/test/basic_test.js
BeksOmega: @maribethb do you have an opinion / context on this from working on the continuous toolbox plugin?

maribethb: I don't think I was the one that added block recycling logic to continuous toolbox so I don't really have any additional context there. I might worry it would break tests where it's assumed the "same" block in the flyout always has the same id.

PR #7054, opened 2023-05-04
feat: add IIcon interface, and related interfaces

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

Work on #7036 

### Proposed Changes

Adds interfaces we need for defining icons.

### Reason for Changes

Interfaces are how we keep the relationships between internal blockly code and external code clear =)

### Test Coverage

N/A These are just interfaces, no functionality.

### Documentation

N/A

### Additional Information

N/A

PR #7050, opened 2023-05-03
chore: test icon management

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

Fixes #7034 

### Proposed Changes

Adds unit tests for adding, removing, and accessing icons.

### Reason for Changes

Tests are good :D

### Test Coverage

This is tests silly!

### Documentation

N/A

### Additional Information

N/A

PR #7059, opened 2023-05-05
feat: add implementations for adding, removing, and getting icons

### LINKED ISSUE #7037: Implement methods for adding, removing, and accessing icons
### Description

Part of #2105 

Implement methods on the block for dealing with icons:

* addIcon
* removeIcon
* getIcon
* getIcons
* hasIcon

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #7037 

### Proposed Changes

Adds `addIcon`, `removeIcon`, `hasIcon`, `getIcon`, and `getIcons`

### Reason for Changes

If you want to have custom icons, you need to be able to add them to blocks :P

### Test Coverage

Tests were added in #7050

### Documentation

N/A

### Additional Information

Dependent on #7050, #7053, #7054

The bad implementation of `getIcons` is necessary to maintain behavior while swapping over the icons system.


tests/mocha/block_test.js
```diff
@@ -1376,6 +1380,7 @@ suite('Blocks', function() {
         this.block.addIcon(new MockIconA());
         chai.assert.throws(
             () => {this.block.addIcon(new MockIconA());},
+            '',
```
maribethb, line 1383: I think neither of these is quite correct? https://www.chaijs.com/api/assert/#method_throws

tests/mocha/block_test.js
```diff
@@ -1420,30 +1425,32 @@ suite('Blocks', function() {
...
-            [iconA, iconB],
-            'Expected getIcon to return both icons');
-      });
+      test(
+          'all icons are returned from getIcons, in order of weight',
+          function() {
+            const iconA = new MockIconA();
+            const iconB = new MockIconB();
+            this.block.addIcon(iconA);
+            this.block.addIcon(iconB);
```
maribethb, line 1448: I would add these in the opposite order so that the test ensures they aren't just being returned in the order they were added.

maribethb (approved):

core/block.ts
```diff
@@ -2084,16 +2086,64 @@ export class Block implements IASTNodeLocation, IDeletable {
...
+
+  /**
+   * Removes the icon whose getType matches the given type iconType from the
+   * block.
+   *
+   * @param type The type of the icon to remove from the block.
+   * @return True if an icon with the given type was found, false otherwise.
+   */
+  removeIcon(type: string): boolean {
+    if (!this.hasIcon(type)) return false;
```
maribethb, line 2122: I sort of think removing an icon that doesn't exist should throw an error the same way adding one that already does throws. It feels more consistent. But I think you did it this way to be consistent with fields?

BeksOmega: [`removeField`](https://github.com/google/blockly/blob/9442b73d99fb63da4526c651ac41762e458b4ed7/core/inputs/input.ts#L145) optionally throws an error (which I think is a weird and bad pattern). I had it return a boolean to be more consistent with other ecmascript things (e.g. maps).

tests/mocha/block_test.js
BeksOmega: I think I'm gonna define a new error class for this. That seems like the correct-est + most useful thing to do.

maribethb (approved):

Issue #7045, opened 2023-05-03
Refactor the "mutator bubble" to be a MiniWorkspaceBubble


BeksOmega:
### Description

Part of #2105 

Refactor the bubble UI used by mutator icons into its own class so that it can easily be reused by multiple icons.

Issue #7044, opened 2023-05-03
Refactor the "warning bubble" to be a TextBubble


BeksOmega:
### Description

Part of #2105 

Make the bubble UI used by warning icons and non-editable comment icons into its own class so that it can easily be reused by multiple icons.

Issue #7043, opened 2023-05-03
Refactor the "comment bubble" to be a TextInputBubble


BeksOmega:
### Description

Part of #2105 

Make the bubble used by the comment icon into its own separate class, so that it can easily be reused by multiple icons.

Issue #7040, opened 2023-05-03
Refactor the Warning icon to conform to the IIcon interface


BeksOmega:
### Description

Work on #2105 

Warning icons should be almost entirely internal until we refactor how `setWarningText` works. See #7041 for details.

PR #7071, opened 2023-05-10
feat: bubble ownership

### LINKED ISSUE #7061: Refactor the built-in bubble class to define its own contents
We want to refactor how bubbles work so that they are defined more as a "view" that owns what it looks like. This makes it easy for us to have subclasses of `Bubble` that have specific views. E.g. a text bubble, a text input bubble, etc.

See also #7043, #7044, #7045

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7061 

### Proposed Changes

Refactors the bubble so that it owns its own view elements, instead of being passed them in the constructor.

### Reason for Changes

Clear separation between models and views.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Things may have to change as I subclass this to create the bubble subclasses, but I think this is basically gtg!


core/bubbles/bubble.ts
```diff
@@ -0,0 +1,558 @@
...
+  /**
+   * Sets the size of this bubble, including the border.
+   *
+   * @param size Sets the size of this bubble, including the border.
+   * @param relayout If true, reposition the bubble from scratch so that it is
+   *     optimally visible. If false, reposition it so it maintains the same
+   *     position relative to the anchor.
+   */
+  protected setSize(size: Size, relayout = false) {
+    // TODO: set size.
```
maribethb, line 154: Is this todo still relevant? if so could you make it more specific please

core/bubbles/bubble.ts
```diff
@@ -0,0 +1,558 @@
...
+  /** Sets the colour of the background and tail of this bubble. */
+  protected setColour(colour: string) {
+    this.colour = colour;
+    this.tail.setAttribute('fill', colour);
+    this.background.setAttribute('fill', colour);
+  }
+
+  /** Gets the emboss filter for this bubble. */
+  private getFilter(): {filter?: string} {
+    // TODO: Do we think this is actually still a problem??
```
maribethb, line 184: this was fun to look into.... i could only find a reference to this being a problem in java 8 in 2014. i'd say remove it

core/blockly.ts
```diff
@@ -26,7 +26,7 @@ import {BlockSvg} from './block_svg.js';
 import {BlocklyOptions} from './blockly_options.js';
 import {Blocks} from './blocks.js';
 import * as browserEvents from './browser_events.js';
-import {Bubble} from './bubble.js';
+import {Bubble} from './bubble_old.js';
```
maribethb, line 29: What is your plan for the `bubble_old` and `icon_old` classes? How long will those stick around?

Do you need to export the new bubble class here?

BeksOmega: They will stick around until all of the builtin bubbles + icons get converted to subclass the new classes.

So the old bubble will go away after #7045 , #7044, and #7043 are resolved.

And the old icon will go away after #7040, #7039, and #7038 are resolved.

maribethb: OK, makes sense. Can you add an issue then to check that we re-export the appropriate new classes from `blockly.ts` once everything is in final form? I don't want to miss that when we do the release.

core/bubbles/bubble.ts
```diff
@@ -0,0 +1,591 @@
...
+import {IBubble} from '../interfaces/i_bubble.js';
+import {ContainerRegion} from '../metrics_manager.js';
+import {Scrollbar} from '../scrollbar.js';
+import {Coordinate} from '../utils/coordinate.js';
+import * as dom from '../utils/dom.js';
+import * as math from '../utils/math.js';
+import {Rect} from '../utils/rect.js';
+import {Size} from '../utils/size.js';
+import {Svg} from '../utils/svg.js';
+import * as userAgent from '../utils/useragent.js';
```
maribethb, line 18: pls fix lint

core/utils/rect.ts
```diff
@@ -18,6 +18,9 @@ goog.declareModuleId('Blockly.utils.Rect');
  * Class for representing rectangular regions.
  */
 export class Rect {
+  public height;
```
maribethb, line 21: you don't need `public` since that's the default.

also, these only ever being set at construction is slightly weird since the other properties are also public and could be changed after the rectangle is constructed, and then the height and width would be incorrect 

and we do sometimes change them after construction e.g. https://github.com/google/blockly/blob/develop/core/workspace_svg.ts#L1901

so I get why you added them as a convenience but I wonder if they would be better as methods than properties, or maybe this is a good use case for a getter (though the style guide notes that transpiling getters results in bloated code)

maribethb (approved):

core/bubbles/bubble.ts
BeksOmega: I'm going to change this to an error so that it fails CI.

core/utils/rect.ts
BeksOmega: I'm going to change them to get-methods.

PR #7080, opened 2023-05-11
feat: text bubble

### LINKED ISSUE #7044: Refactor the "warning bubble" to be a TextBubble

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7044 

### Proposed Changes

Creates a TextBubble that visualizes text.

### Reason for Changes

We want bubbles to be pure views with a object oriented structure. This works toward that =)

### Test Coverage

Manually tested by hacking into the warning icon. Looks the same as bubbles did previously.

### Documentation

N/A

### Additional Information

Dependent on #7071

### DIFF
--- a/core/bubbles/text_bubble.ts
+++ b/core/bubbles/text_bubble.ts
@@ -1,0 +1,99 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Bubble} from './bubble.js';
+import {Coordinate} from '../utils/coordinate.js';
+import * as dom from '../utils/dom.js';
+import {Rect} from '../utils/rect.js';
+import {Size} from '../utils/size.js';
+import {Svg} from '../utils/svg.js';
+import {WorkspaceSvg} from '../workspace_svg.js';
+
+export class TextBubble extends Bubble {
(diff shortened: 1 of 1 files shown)


core/bubbles/text_bubble.ts
```diff
@@ -0,0 +1,99 @@
...
+    dom.removeNode(this.paragraph);
+    this.paragraph = this.stringToSvg(text, this.contentContainer);
+    this.updateBubbleSize();
+  }
+
+  /**
+   * Converts the given string into an svg containing that string,
+   * broken up by newlines.
+   */
+  private stringToSvg(text: string, container: SVGGElement) {
```
rachel-fenichel, line 46: You're calling `stringToSvg` with the same `container` in both cases (`this.contentContainer`) so you could remove the param and just access it inside the method.

(Optional change)

rachel-fenichel (approved):

core/bubbles/text_bubble.ts
BeksOmega: It's my preference to keep methods pure if possible. So I'm going to use as is, but if people prefer the other style we can always switch later since this is private.

PR #7095, opened 2023-05-17
chore: test for procedures

ericblackmonGoogle:
## The basics

- [x] I branched from develop
- [ x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [x ] I ran `npm run format` and `npm run lint`

### Resolves

Creates a simple test for procedures

### Proposed Changes

Creates a simple test for procedures

#### Behavior Before Change

Creates a simple test for procedures

#### Behavior After Change

Creates a simple test for procedures

### Reason for Changes

To expand on our automated test

### Test Coverage

It is a test
### Additional Information

N/A


ericblackmonGoogle:
Currently working on documentation for how to run test, but here is the documentation on how I built the prototype[prototype](https://docs.google.com/document/d/1LYL6sJUQlM74XiSmAyRVadY1oMOwfqmsSpvn3CU9Khs/edit?resourcekey=0-ZU-Oy-In0vuAwFnRFwD3sQ#heading=h.xn4emzo00mow)

tests/browser/test/procedure_test.js
```diff
@@ -0,0 +1,152 @@
...
+const path = require('path');
+const {posixPath} = require('../../../scripts/helpers');
+
+let browser;
+suite('Testing Connecting Blocks', function (done) {
+  // Setting timeout to unlimited as the webdriver takes a longer time to run than most mocha test
+  this.timeout(0);
+
+  // Setup Selenium for all of the tests
+  suiteSetup(async function () {
```
BeksOmega, line 22: Is this setup going to be the same for every test file? Might be worth moving it into its own helper method (in a separate PR).

tests/browser/test/procedure_test.js
```diff
@@ -0,0 +1,152 @@
...
+      );
+    } else {
+      // --disable-gpu is needed to prevent Chrome from hanging on Linux with
+      // NVIDIA drivers older than v295.20. See
+      // https://github.com/google/blockly/issues/5345 for details.
+      options.capabilities['goog:chromeOptions'].args.push('--disable-gpu');
+    }
+    // Use Selenium to bring up the page
+    const url =
+      'https://rachel-fenichel.github.io/blockly/demos/code/index.html';
```
BeksOmega, line 49: Can this be tested on the playground rather than the code demo? Perferably using the same local testing as the [playground test](https://github.com/google/blockly/blob/d3939f7a488c4bd7868237a7899a67e98bc0a79d/tests/browser/test/basic_playground_test.js#L47).

tests/browser/test/procedure_test.js
```diff
@@ -0,0 +1,152 @@
...
+    console.log('Starting webdriverio...');
+    browser = await webdriverio.remote(options);
+    console.log('Loading URL: ' + url);
+    await browser.url(url);
+    return browser;
+  });
+
+  test('Testing Procedure', async function () {
+    // Drag out first function
+    const functionButton = await browser.$('#blockly\\:9 > div.blocklyTreeRow');
```
BeksOmega, line 60: For "xButton" can you rename it to "xCategory"? We have a different thing where you can put buttons in a flyout, so the current naming is a bit confusing.

tests/browser/test/procedure_test.js
```diff
@@ -0,0 +1,152 @@
...
+    await new Promise((resolve) => setTimeout(resolve, 2000)); // 2 sec
+    functionButton.click();
+    let toReturnBlock = await browser.$(
```
BeksOmega, line 63: Similarly, for the blocks I think it would be a good idea to name them the same as the block type (where possible). E.g. 'proceduresDefReturn"

BeksOmega (commented):
I'm curious to see how stable or flakey this test ends up being. But that's something that can only be seen with time! So LGTM after the naming is fixed up =)

tests/browser/test/procedure_test.js
ericblackmonGoogle: There isn't a way, that I know of, to run the procedures. I think this is why all procedure testing uses 'https://rachel-fenichel.github.io/blockly/demos/code/index.html' since it has the run button built in

tests/browser/test/procedure_test.js
ericblackmonGoogle: Great call I can do that i follow up CL

tests/browser/test/procedure_test.js
ericblackmonGoogle: Will do, just to clarify. Is it ok to leave the names the same once they are in the workspace?

tests/browser/test/procedure_test.js
BeksOmega: @cpcallen Can we use the local code demo here instead of Rachel's hosted code demo?

BeksOmega (approved):

Issue #7082, opened 2023-05-12
Confirm custom icon exports are correct before release


BeksOmega:
Before doing the quarterly release, confirm that the exports related to custom icons are correct. In particular:
  * Ensure that everything is exported from the `Blockly.icons` namespace, and not exported individually.
  * Ensure that icon_old and bubble_old are not exported.
  * Ensure the renamings script is up to date.
  * Resolve https://github.com/google/blockly/pull/7096#issuecomment-1555285883

PR #7096, opened 2023-05-18
feat: mini workspace bubble

### LINKED ISSUE #7045: Refactor the "mutator bubble" to be a MiniWorkspaceBubble

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7045 

### Proposed Changes

Adds a new kind of bubble which is a mini workspace bubble.

### Reason for Changes

We want bubbles to be pure views with a object oriented structure. This works toward that =)

### Test Coverage

Manually tested by hacking into the warning icon. Looks the same as bubbles did previously.

This is the one I'm most nervous about since it has so much functionality that I can't test it all. But the simple case looks good! And I added validation to hopefully restrict peeps to those use cases.

If we want we can also mark the constructor `@internal` until we're more confident in it.

### Documentation

N/A

### DIFF
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -174,6 +174,7 @@ import {Menu} from './menu.js';
 import {MenuItem} from './menuitem.js';
 import {MetricsManager} from './metrics_manager.js';
 import {Msg, setLocale} from './msg.js';
+import {MiniWorkspaceBubble} from './bubbles/mini_workspace_bubble.js';
 import {Mutator} from './mutator.js';
 import {Names} from './names.js';
 import {Options} from './options.js';
@@ -440,6 +441,12 @@ WorkspaceCommentSvg.prototype.showContextMenu = function (
 };
 
 Mutator.prototype.newWorkspaceSvg = function (options: Options): WorkspaceSvg {
+  return new WorkspaceSvg(options);
+};
+
+MiniWorkspaceBubble.prototype.newWorkspaceSvg = function (
+  options: Options
+): WorkspaceSvg {
   return new WorkspaceSvg(options);
 };
 
--- a/core/bubbles/bubble.ts
+++ b/core/bubbles/bubble.ts
@@ -18,31 +18,31 @@ import {WorkspaceSvg} from '../workspace_svg.js';
 
 export abstract class Bubble implements IBubble {
   /** The width of the border around the bubble. */
-  static BORDER_WIDTH = 6;
+  static readonly BORDER_WIDTH = 6;
 
   /** Double the width of the border around the bubble. */
-  static DOUBLE_BORDER = this.BORDER_WIDTH * 2;
+  static readonly DOUBLE_BORDER = this.BORDER_WIDTH * 2;
 
   /** The minimum size the bubble can have. */
-  static MIN_SIZE = this.DOUBLE_BORDER;
+  static readonly MIN_SIZE = this.DOUBLE_BORDER;
 
   /**
    * The thickness of the base of the tail in relation to the size of the
    * bubble. Higher numbers result in thinner tails.
    */
-  static TAIL_THICKNESS = 1;
+  static readonly TAIL_THICKNESS = 1;
 
   /** The number of degrees that the tail bends counter-clockwise. */
-  static TAIL_ANGLE = 20;
+  static readonly TAIL_ANGLE = 20;
 
   /**
    * The sharpness of the tail's bend. Higher numbers result in smoother
    * tails.
    */
-  static TAIL_BEND = 4;
+  static readonly TAIL_BEND = 4;
 
   /** Distance between arrow point and anchor point. */
-  static ANCHOR_RADIUS = 8;
+  static readonly ANCHOR_RADIUS = 8;
 
   /** The SVG group containing all parts of the bubble. */
   private svgRoot: SVGGElement;
--- a/core/bubbles/bubble.ts
+++ b/core/bubbles/bubble.ts
@@ -551,11 +551,8 @@ export abstract class Bubble implements IBubble {
   }
 
   /**
-   * Move this bubble during a drag, taking into account whether or not there is
-   * a drag surface.
+   * Move this bubble during a drag.
    *
-   * @param dragSurface The surface that carries rendered items during a drag,
-   *     or null if no drag surface is in use.
    * @param newLoc The location to translate to, in workspace coordinates.
    * @internal
    */
--- a/core/bubbles/bubble.ts
+++ b/core/bubbles/bubble.ts
@@ -20,8 +20,11 @@ export abstract class Bubble implements IBubble {
   /** The width of the border around the bubble. */
   static BORDER_WIDTH = 6;
 
+  /** Double the width of the border around the bubble. */
+  static DOUBLE_BORDER = this.BORDER_WIDTH * 2;
+
   /** The minimum size the bubble can have. */
-  static MIN_SIZE = this.BORDER_WIDTH * 2;
+  static MIN_SIZE = this.DOUBLE_BORDER;
 
   /**
    * The thickness of the base of the tail in relation to the size of the
--- a/core/bubbles/mini_workspace_bubble.ts
+++ b/core/bubbles/mini_workspace_bubble.ts
@@ -7,19 +7,34 @@
 import {Abstract as AbstractEvent} from '../events/events_abstract.js';
 import type {BlocklyOptions} from '../blockly_options.js';
 import {Bubble} from './bubble.js';
-import type {Coordinate} from '../utils/coordinate.js';
+import {Coordinate} from '../utils/coordinate.js';
 import * as dom from '../utils/dom.js';
 import {Options} from '../options.js';
 import {Svg} from '../utils/svg.js';
 import type {Rect} from '../utils/rect.js';
 import {Size} from '../utils/size.js';
-import {WorkspaceSvg} from '../workspace_svg.js';
+import type {WorkspaceSvg} from '../workspace_svg.js';
 
 export class MiniWorkspaceBubble extends Bubble {
+  /**
+   * The minimum amount of change to the mini workspace view to trigger
+   * resizing the bubble.
+   */
+  private static readonly MINIMUM_VIEW_CHANGE = 10;
+
+  /** The root svg element containing the workspace. */
   private svgDialog: SVGElement;
+
+  /** The workspace that gets shown within this bubble. */
   private miniWorkspace: WorkspaceSvg;
+
+  /**
+   * Should this bubble automatically reposition itself when it resizes?
+   * Becomes false after this bubble is first dragged.
+   */
   private autoLayout = true;
 
+  /** @internal */
   constructor(
     workspaceOptions: BlocklyOptions,
     protected readonly workspace: WorkspaceSvg,
@@ -121,8 +136,10 @@ export class MiniWorkspaceBubble extends Bubble {
     const currSize = this.getSize();
     const newSize = this.calculateWorkspaceSize();
     if (
-      Math.abs(currSize.width - newSize.width) < 10 &&
-      Math.abs(currSize.height - newSize.height) < 10
+      Math.abs(currSize.width - newSize.width) <
+        MiniWorkspaceBubble.MINIMUM_VIEW_CHANGE &&
+      Math.abs(currSize.height - newSize.height) <
+        MiniWorkspaceBubble.MINIMUM_VIEW_CHANGE
     ) {
       // Only resize if the size has noticeably changed.
       return;
@@ -131,6 +148,7 @@ export class MiniWorkspaceBubble extends Bubble {
     this.svgDialog.setAttribute('height', `${newSize.height}px`);
     this.miniWorkspace.setCachedParentSvgSize(newSize.width, newSize.height);
     if (this.miniWorkspace.RTL) {
+      // Scroll the workspace to always left-align.
       this.miniWorkspace
         .getCanvas()
         .setAttribute('transform', `translate(${newSize.width}, 0)`);
@@ -152,7 +170,7 @@ export class MiniWorkspaceBubble extends Bubble {
   private calculateWorkspaceSize(): Size {
     const canvas = this.miniWorkspace.getCanvas();
     const bbox = canvas.getBBox();
-    let width = this.miniWorkspace.RTL ? -bbox.x : bbox.width;
+    let width = bbox.width + Bubble.DOUBLE_BORDER * 3;
     let height = bbox.height + bbox.y + Bubble.DOUBLE_BORDER * 3;
     const flyout = this.miniWorkspace.getFlyout();
     if (flyout) {
@@ -161,13 +179,8 @@ export class MiniWorkspaceBubble extends Bubble {
         .getMetricsManager()
         .getScrollMetrics();
       height = Math.max(height, flyoutScrollMetrics.height + 20);
-      if (this.miniWorkspace.RTL) {
-        width = Math.max(width, flyout.getWidth());
-      } else {
-        width += flyout.getWidth();
-      }
+      width += flyout.getWidth();
     }
-    width += Bubble.DOUBLE_BORDER * 3;
     return new Size(width, height);
   }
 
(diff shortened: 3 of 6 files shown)


core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.svgDialog = dom.createSvgElement(
+      Svg.SVG,
+      {
+        'x': Bubble.BORDER_WIDTH,
+        'y': Bubble.BORDER_WIDTH,
+      },
+      this.contentContainer
+    );
+    workspaceOptions.parentWorkspace = this.workspace;
+    this.miniWorkspace = this.newWorkspaceSvg(new Options(workspaceOptions));
```
cpcallen, line 42:
```suggestion
    this.miniWorkspace = new WorkspaceSvg(new Options(workspaceOptions));
```

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  /** @internal */
+  getWorkspace(): WorkspaceSvg {
+    return this.miniWorkspace;
+  }
+
+  /** Adds a change listener to the mini workspace. */
+  addWorkspaceChangeListener(listener: (e: AbstractEvent) => void) {
+    this.miniWorkspace.addChangeListener(listener);
+  }
```
cpcallen, line 76: What is the value of this method?  Why would a developer not just do `myMWB.getWorkspace().addChangeListener(…)`?  (Do we want to be able to monitor the adding of change listeners?  Seems futile given that the developer can just add it directly as noted.)

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+   * mini workspace.
+   */
+  private updateBubbleSize() {
+    if (this.miniWorkspace.isDragging()) return;
+
+    const currSize = this.getSize();
+    const newSize = this.calculateWorkspaceSize();
+    if (
+      Math.abs(currSize.width - newSize.width) < 10 &&
+      Math.abs(currSize.height - newSize.height) < 10
```
cpcallen, line 125: Should this delta value (`10`) be configurable, or at least a named constant?  Also, what units is it measured in?  10px seems like a pretty noticeable change.

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+      return;
+    }
+    this.svgDialog.setAttribute('width', `${newSize.width}px`);
+    this.svgDialog.setAttribute('height', `${newSize.height}px`);
+    this.miniWorkspace.setCachedParentSvgSize(newSize.width, newSize.height);
+    if (this.miniWorkspace.RTL) {
+      this.miniWorkspace
+        .getCanvas()
+        .setAttribute('transform', `translate(${newSize.width}, 0)`);
+    }
```
cpcallen, line 137: I'm not sure what's going on here.  A comment would be helpful.  I guess this is because the workspace needs to have its left edge rather than right edge moved, but I'm unclear on why the code uses only the new width rather than the difference between new and old.

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.miniWorkspace.recordDragTargets();
+  }
+
+  /**
+   * Calculates the size of the mini workspace for use in resizing the bubble.
+   */
+  private calculateWorkspaceSize(): Size {
+    const canvas = this.miniWorkspace.getCanvas();
+    const bbox = canvas.getBBox();
+    let width = this.miniWorkspace.RTL ? -bbox.x : bbox.width;
```
cpcallen, line 155: Similarly to above: it eludes me why only `x` is relevant in RTL.  Surely a width is a width?

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+    let height = bbox.height + bbox.y + Bubble.DOUBLE_BORDER * 3;
+    const flyout = this.miniWorkspace.getFlyout();
+    if (flyout) {
+      const flyoutScrollMetrics = flyout
+        .getWorkspace()
+        .getMetricsManager()
+        .getScrollMetrics();
```
cpcallen, line 162: Maybe I am beginning to regret my previous endorsement of Prettier.

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+      height = Math.max(height, flyoutScrollMetrics.height + 20);
+      if (this.miniWorkspace.RTL) {
+        width = Math.max(width, flyout.getWidth());
+      } else {
+        width += flyout.getWidth();
+      }
+    }
+    width += Bubble.DOUBLE_BORDER * 3;
```
cpcallen, line 170: Huh?  Why do we want to increase the width by 6 * border width?

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  /** @internal */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  newWorkspaceSvg(options: Options): WorkspaceSvg {
+    throw new Error(
+      'The implementation of newWorkspaceSvg should be ' +
+        'monkey-patched in by blockly.ts'
+    );
+  }
```
cpcallen, line 198:
```suggestion
```

core/flyout_base.ts
```diff
@@ -136,14 +137,14 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
    * Function that will be registered as a change listener on the workspace
    * to reflow when blocks in the flyout workspace change.
    */
-  private reflowWrapper: Function | null = null;
+  private reflowWrapper: ((e: AbstractEvent) => void) | null = null;
```
cpcallen, line 140: Hmm.  Technically this (and the below) is a breaking change, but I imagine that passing in a function that does not accept an `AbstractEvent` as its first parameter did not actually work previously, despite the sloppy signature.  Is that true?

core/mutator.ts
```diff
@@ -265,7 +265,7 @@ export class Mutator extends Icon {
     const canvas = this.workspace_.getCanvas();
     const workspaceSize = canvas.getBBox();
     let width = workspaceSize.width + workspaceSize.x;
-    let height = workspaceSize.height + doubleBorderWidth * 3;
+    let height = workspaceSize.height + doubleBorderWidth * 2;
```
cpcallen, line 268: Why this seemingly unrelated change to `Mutator`?

core/workspace.ts
```diff
@@ -668,7 +668,7 @@ export class Workspace implements IASTNodeLocation {
    * @param func Function to call.
    * @returns Obsolete return value, ignore.
    */
-  addChangeListener(func: Function): Function {
+  addChangeListener(func: (e: Abstract) => void): Function {
```
cpcallen, line 671: Ditto previous re: breaking change.

cpcallen (changes requested):
> We want bubbles to be pure views with a object oriented structure. This works toward that =)

I'm not quite sure what you mean by "pure views".

Am I correct in understanding that the current `Mutator` (really: block button with gear icon) will in future use a `MiniWorkspaceBubble` instead of whatever it currently uses?

> This is the one I'm most nervous about since it has so much functionality that I can't test it all. But the simple case looks good! And I added validation to hopefully restrict peeps to those use cases.

I was surprised that the mini workspace doesn't support scrolling, trash, etc.  Are there technical limitations why that is, or is this just policy?  If the latter, why?

> If we want we can also mark the constructor `@internal` until we're more confident in it.

That seems like a good idea, though not exporting it (as is presently the case) is probably just as effective.

cpcallen (changes requested):
> We want bubbles to be pure views with a object oriented structure. This works toward that =)

I'm not quite sure what you mean by "pure views".

Am I correct in understanding that the current `Mutator` (really: block button with gear icon) will in future use a `MiniWorkspaceBubble` instead of whatever it currently uses?

> This is the one I'm most nervous about since it has so much functionality that I can't test it all. But the simple case looks good! And I added validation to hopefully restrict peeps to those use cases.

I was surprised that the mini workspace doesn't support scrolling, trash, etc.  Are there technical limitations why that is, or is this just policy?  If the latter, why?

> If we want we can also mark the constructor `@internal` until we're more confident in it.

That seems like a good idea, though not exporting it (as is presently the case) is probably just as effective.

core/flyout_base.ts
BeksOmega: If you passed in a `() => void` that's fine because it's a subtype of `(anyParameters) => void`. If you passedin `(someTypeBesidesAnEvent) => void` then yeah that didn't work before. So this shouldn't break anything.

core/mutator.ts
BeksOmega: Accidental commit, thanks for the catch!

core/bubbles/bubble.ts
```diff
@@ -20,8 +20,11 @@ export abstract class Bubble implements IBubble {
   /** The width of the border around the bubble. */
   static BORDER_WIDTH = 6;
 
+  /** Double the width of the border around the bubble. */
+  static DOUBLE_BORDER = this.BORDER_WIDTH * 2;
```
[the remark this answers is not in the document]
BeksOmega: Ah whoops, these should both be readonly.

core/bubbles/mini_workspace_bubble.ts
BeksOmega: `getWorkspace` is internal because allowing them to arbitrarily mess with the workspace is a large promise that invites a lot of edge cases I don't have time to test. This creates a more minimal API.

core/bubbles/mini_workspace_bubble.ts
BeksOmega: Can add a constant. I don't think the change is necessarily "noticable" because we always have margin around the contents of the workspace. So if we don't resize for <10px there will just be slightly less margin than if we do resize.

core/bubbles/mini_workspace_bubble.ts
BeksOmega: I'll pull over the comment from `mutator`. I don't think the comment illuminates a lot, but yeah it does have to do with alignment.

core/bubbles/mini_workspace_bubble.ts
BeksOmega: This is just an arbitrary margin, pulled over from the existing mutator code.

BeksOmega:
> Am I correct in understanding that the current Mutator (really: block button with gear icon) will in future use a MiniWorkspaceBubble instead of whatever it currently uses?

Yes!

> I was surprised that the mini workspace doesn't support scrolling, trash, etc. Are there technical limitations why that is, or is this just policy? If the latter, why?

There are technical limitations related to how the workspace gets sized (and automatically resized). We can't support scrolling because the workspace should never be scrollable (always resizing to show the whole workspace). We can't support the trashcan because (currently) we don't take the size of the trashcan into account when resizing.

In the future we can always add support for these things! I just don't have time to implement that right now.

> That seems like a good idea, though not exporting it (as is presently the case) is probably just as effective.

I think we do want to export this for documentation purposes though. I'm pretty sure if it's not exported then API extractor won't include information about what other methods it has and what not. I'll mark it as internal just in case.

core/blockly.ts
```diff
@@ -443,6 +444,12 @@ Mutator.prototype.newWorkspaceSvg = function (options: Options): WorkspaceSvg {
   return new WorkspaceSvg(options);
 };
 
+MiniWorkspaceBubble.prototype.newWorkspaceSvg = function (
```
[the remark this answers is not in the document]
BeksOmega: Currently this is fine, but eventually we're going to have: WorkspaceSvg imports BlockSvg imports MutatorIcon imports MiniWorkspaceBubble imports WorkspaceSvg, and then it's going to break again. So might as well just add this from the beginning.

Also testing this with the WarningIcon caused the same problems.

Afaict this circular dependency is just an unfixable constraint if we want to provide miniworkspace bubbles :/

core/bubbles/mini_workspace_bubble.ts
BeksOmega: Also eludes me, this was pulled over from the mutator code, but I don't think it was working 100% correctly. E.g. having a RTL bubble in a LTR workspace would break (idk why you would want to do that, but that's beside the point :P) Tested unifying the width code and it looks good! Renders correctly in LTR and RTL with a flyout and without a flyout.

core/blockly.ts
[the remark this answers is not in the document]
cpcallen: Hmm.  That is unfortunate.  I cant' help but think that there must be a better way to break this circularity.  Like: couldn't `MutatorIcon` just fulfil an `IIcon` interface, and have `Block` not know any more specifics?

But fixing that is probably out-of-scope for this PR, so for now I approve of you using the existing pattern.

core/bubbles/mini_workspace_bubble.ts
cpcallen: > `getWorkspace` is internal

Oh, good point.

I'm sort of tempted to suggest having `getWorkspace` be public but return a `Proxy` that strictly limits what kind of messing they can do, but I will refrain.  (You can tell that I have spent too much time thinking about adversarial programming in JavaScript for Code City.)

Question, though: if they can add a listener, can the listener obtain a reference to the `Workspace` when an event fires?  (See previous parenthetical comment.)

core/bubbles/mini_workspace_bubble.ts
cpcallen: Isn't the border only 6px?  Or am I confused?

core/bubbles/mini_workspace_bubble.ts
BeksOmega: That's true, they can obtain a reference to the event via `Blockly.Workspace.getById` from the `workspaceId` property of the event.

core/bubbles/mini_workspace_bubble.ts
cpcallen: OK, I'm clearly confused between border and margin, but I think a comment (or a named constant) would be in order here too.

core/flyout_base.ts
cpcallen: Great.

core/bubbles/mini_workspace_bubble.ts
BeksOmega: The border is 6px, but we also have white space that exists around the workspace, which is what I meant by margin (sorry we don't really have language for this stuff so I'm just making it up on the spot lol). 
image

cpcallen:
> I think we do want to export this for documentation purposes though. I'm pretty sure if it's not exported then API extractor won't include information about what other methods it has and what not. I'll mark it as internal just in case.

Does API extractor include information about things marked `@internal` in its output?  And does it have any way to distinguish the type from the constructor for the type?  I'd sort of expect it would either show nothing about the type, or it would show the whole type as being `@internal`, which I think is not what you want.

But if you've checked that it is output as you intend then great.

cpcallen (approved):
Approved subject to due consideration given to remaining unresolved comments (changes at your discretion).

core/bubbles/mini_workspace_bubble.ts
BeksOmega: I was testing this again to make sure I caught everything, and it doesn't resize correctly as we create space between the flyout and the block. I'm just going to pull over the exact code from the existing mutator (since we know that works in common situtations). And file a bug for actually fixing / explaining this logic, and fixing the erroneous bits (e.g. RTL in LTR mode).

core/bubbles/mini_workspace_bubble.ts
BeksOmega: I think that a proxy would be the most correct solution to this, but I think this is the best compromise for now. Neither the current solution or a proxy would provide complete protection due to `Blockly.Workspace.getById` addressed above (i.e. someone can still always get to the workspace if they want to). But this does limit external developer's access.

If I had more time I would investigate a proxy, but for now I'm going to consider it out of scope.

BeksOmega:
> Does API extractor include information about things marked @internal in its output? ...

I will resolve this as part of #7082

core/bubbles/mini_workspace_bubble.ts
cpcallen: > If I had more time I would investigate a proxy, but for now I'm going to consider it out of scope.

The use of a `Proxy` was not really a serious suggestion (though it would certainly be a way to control access).  I think it would probably be better to invest the time in getting the flyout workspace to support all the regular features (including sub-bubbles!  🫧🫧🫧)

PR #7100, opened 2023-05-18
fix: work on calling icon hooks

### LINKED ISSUE #7098: Call new icon hooks in block rendering code
### Description

Actually call new hooks like `initView`, `applyColour`, `updateEditable`, and `updateCollapsed` at the appropriate places.

### LINKED ISSUE #7099: Modify gesture system to handle icon clicks
### Description

In the past the icons would handle click on `pointerup` obviating the need for them to use the gesture system. For code unification purposes we want to change them to use the gesture system like everything else (in particular fields) does.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7098

### Proposed Changes

Adds correctly calling the new icon hooks in a bunch of places.

### Reason for Changes

Actually get the code to work for testing refactoring icons.

### Test Coverage

Enabling tests yay!

### Documentation

N/A

### Additional Information

Two of the `initView` tests are still skipped because I need to fix #7099 first.

### DIFF
--- a/core/render_management.ts
+++ b/core/render_management.ts
@@ -153,6 +153,7 @@ function updateIconLocations(block: BlockSvg, blockOrigin: Coordinate) {
   for (const child of block.getChildren(false)) {
     updateIconLocations(
       child,
-      Coordinate.sum(blockOrigin, child.relativeCoords));
+      Coordinate.sum(blockOrigin, child.relativeCoords)
+    );
   }
 }
--- a/core/render_management.ts
+++ b/core/render_management.ts
@@ -5,6 +5,7 @@
  */
 
 import {BlockSvg} from './block_svg.js';
+import {isIcon} from './interfaces/i_icon.js';
 import {Coordinate} from './utils/coordinate.js';
 
 /** The set of all blocks in need of rendering which don't have parents. */
@@ -84,8 +85,9 @@ function doRenders() {
     if (block.getParent()) continue;
 
     renderBlock(block);
-    updateConnectionLocations(block, block.getRelativeToSurfaceXY());
-    updateIconLocations(block);
+    const blockOrigin = block.getRelativeToSurfaceXY();
+    updateConnectionLocations(block, blockOrigin);
+    updateIconLocations(block, blockOrigin);
   }
   for (const workspace of workspaces) {
     workspace.resizeContents();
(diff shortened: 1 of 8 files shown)


core/render_management.ts
```diff
@@ -138,12 +140,23 @@ function updateConnectionLocations(block: BlockSvg, blockOrigin: Coordinate) {
...
+    if (isIcon(icon)) {
+      icon.onLocationChange(blockOrigin);
+    } else {
+      // TODO (#7042): Remove old icon handling code.
+      icon.computeIconLocation();
+    }
   }
-  for (const child of block.getChildren(false)) {
-    updateIconLocations(child);
+  for (const conn of block.getConnections_(false)) {
```
maribethb, line 153: I don't understand this change -- doesn't `getChildren` already return the blocks connected via input and next connections? Using `getChildren` is more readable and already does some of the work like not finding previous connections and skipping connections with no blocks attached. I'm sure I'm missing something though.

core/renderers/common/drawer.ts
```diff
@@ -321,7 +328,13 @@ export class Drawer {
     if (this.info_.isInsertionMarker) {
       // Fields and icons are invisible on insertion marker.  They still have to
       // be rendered so that the block can be sized correctly.
-      svgGroup.setAttribute('display', 'none');
+      if (isIcon((fieldInfo as Icon).icon)) {
```
maribethb, line 331: 😵‍💫

maribethb (approved):

core/renderers/common/drawer.ts
BeksOmega: Haha I'll do another look at cleaning this up, but I think this method just needs to be split into some submethods

core/render_management.ts
BeksOmega: We could iterate the children, but we'd need to access their connections anyway to get the `relativeCoords` so we can do the fancy efficient math. So I think this is more readable given that constraint? but I can also check doing it the other way

maribethb: Isn't `target` === `child` (in the way the code was before)? 

So iiuc the code would be

```
for (const child of block.getChildren(false)) {
  updateIconLocations(target, Coordinate.sum(blockOrigin, child.relativeCoords))
}
```

core/renderers/common/drawer.ts
maribethb: I'm assuming that when you remove handling the old types some of these if statements will go away (though maybe not here since you didn't leave the comment about removing it?) so I tried not to comment on that aspect of them, but this one got me lol. I don't think you really need to change anything here unless you think it's actually possible to improve htis. but don't bother if it will be removed in 7042

core/render_management.ts
BeksOmega: Oh whoops you're right I was reading this wrong. I thought `target` was `targetConnection` not `targetBlock`. Can fix =)

PR #7105, opened 2023-05-20
feat: use new intermediate UI event for field edits, normal block change event for closing editor.

### LINKED ISSUE #2496: Make onFinishEditing_ Better
### Problem statement

Currently the text input field has a few little lines inside its widgetDispose_ function that call an onFinishedEditing_ function if it is defined. This is like a secret callback used by the [procedure mutator blocks](https://github.com/google/blockly/blob/9c4f9d6c1693c3c25160f008d3eaa2539127c9e8/blocks/procedures.js#L530) for deleting unnecessary variables.

This is all good, but there's probably a cleaner way to structure the code than directly adding a function to the field object.

### Proposals

A) Create an actual onFinishedEditing API. If one is created, it should probably apply to more than just the text input field (possibly every field). 

B) The procedures_mutatorarg block should clean up after itself as it creates new variables. I'm not sure how possible this is because I haven't looked into it.

### Additional Information

N/A

johnnesky:
One proposal to fix #2496. Creates a new event type specifically for intermediate field value changes, which is not recorded in undo history. Inhibits firing normal block change events for field value edits when triggered via a user input widget, and instead broadcasts this new event type. When the input widget is disposed, then a standard block change event will be fired with an oldValue corresponding to what the field's value was when the widget was opened.

Compare with: https://github.com/google/blockly/pull/7064/ and https://github.com/google/blockly/pull/7151

Unlike https://github.com/google/blockly/pull/7151 this version of the new event type inherits from UiBase instead of BlockBase. The difference is minor and only affects one file: events_block_field_intermediate_change.ts


core/field_input.ts
```diff
@@ -499,7 +521,20 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
    */
   private onHtmlInputChange_(_e: Event) {
+    // Intermediate value changes from user input are not confirmed until the
+    // user closes the editor, and may be numerous. Inhibit reporting these as
+    // normal block change events, and instead report them as special
+    // intermediate changes that do not get recorded in undo history.
+    const oldValue = this.value_;
+    eventUtils.disable();
     this.setValue(this.getValueFromEditorText_(this.htmlInput_!.value));
+    eventUtils.enable();
```
BeksOmega, line 531: I don't love the use of `disable` and `enable` to change the behavior here, because we're using the state of a separate system (the event system) to change the behavior of the super class (the `Field`). I think this makes the logic harder to follow/debug.

Do you think it would be possible / how would you feel about passing a flag to `setValue` instead? Similarly to your previous implementation: https://github.com/google/blockly/pull/7064/files#diff-a671fdac0ad6e251056d1b8aa6f3e2d746d3476e8945ff81624f5733dc4e28e4R996

tests/mocha/event_block_field_intermediate_change_test.js
```diff
@@ -0,0 +1,36 @@
...
+  setup(function () {
+    sharedTestSetup.call(this);
+    this.workspace = new Blockly.Workspace();
+  });
+
+  teardown(function () {
+    sharedTestTeardown.call(this);
+  });
+
+  suite('Serialization', function () {
```
BeksOmega, line 24: Yay tests :D

core/events/events_block_field_intermediate_change.ts
```diff
@@ -0,0 +1,143 @@
...
+import type {Block} from '../block.js';
+import type {BlockSvg} from '../block_svg.js';
+import * as registry from '../registry.js';
+import * as utilsXml from '../utils/xml.js';
+import {Workspace} from '../workspace.js';
+import * as Xml from '../xml.js';
+
+import {AbstractEventJson} from './events_abstract.js';
+import {UiBase} from './events_ui_base.js';
+import * as eventUtils from './utils.js';
```
BeksOmega, line 25: I thought I meant these lint warnings error but apparently I didn't -.- could you fix these before merging?

BeksOmega (commented):
Overal I like the implementation! Thanks for the back and forth on this :D

core/field_input.ts
```diff
@@ -72,6 +72,13 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
    */
   protected isTextValid_ = false;
 
+  /**
+   * The intial value of the field when the user opened an editor to change its
+   * value. When the editor is disposed, an event will be fired that uses this
+   * as the event's oldValue.
+   */
+  protected valueWhenEditorWasOpened_: string | T | null = null;
```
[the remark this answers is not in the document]
johnnesky: Yes. I don't know why it extends Field<string | T>, but that determines the type of value. Apparently all FieldInputs support string values even if the FieldInput subclass provides a different type for T.

core/events/events_block_field_intermediate_change.ts
johnnesky: Done!

core/field_input.ts
johnnesky: Done!

johnnesky:
Added a small change to procedures.ts so that the mutator recognizes the new event type and updates param names in response to intermediate edits.

BeksOmega:
Superseded by https://github.com/google/blockly/pull/7151

PR #7112, opened 2023-05-22
fix!: refactor warning icon

BeksOmega:
## The basics

- [X] I branched from develop
- [X] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

Fixes #7040 

### Proposed Changes

Refactors the warning icon to conform to the `IIcon` interface.

### Reason for Changes

Code unification!

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Dependent on #7110 

## Breaking changes / updating / upgrading

If you were previously modifying the warning icon directly, that is no longer available. Modifying warnings should be done through the `setWarningText` method on the block. E.g:
```
myBlock.setWarningText('some new warning text', 'an ID');
```

If you need to modify whether the bubble of the warning icon is visible or not, you can do:
```
myBlock.getIcon(Blockly.icons.WarningIcon.TYPE).setBubbleVisible(true);
```

If you have another use case that is not covered here, please [file a feature request](https://github.com/google/blockly/issues/new/choose) so we can work to find a solution for you =)

PR #7115, opened 2023-05-23
fix!: refactor mutator icon

### LINKED ISSUE #7039: Refactor the Mutator icon to conform to the IIcon interface
### Description

Part of #2105 

Mutators should have the following additional method:
  * getWorkspace

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7039 

### Proposed Changes

Refactors the mutator icon to conform to the `IIcon` interface.

### Reason for Changes

Code consistency :D

### Test Coverage

Updated tests! Also manual testing :D

### Documentation

N/A

### Additional Information

Dependent on #7110 

The `Blockly.Mutator` class has been renamed to `Blockly.icons.MutatorIcon`. If you use the blockly migration package, you can automatically rename this throughout your files by doing:

```
npx @blockly/migrate rename --from 9 --in-place ./path/to/my/files*
```

### Showing/hiding bubbles

The `setVisible` method has been renamed to `setBubbleVisible` for clarity. If you need to modify whether the bubble of the mutator icon is visible or not, you can do:

```js
myBlock.getIcon(Blockly.icons.MutatorIcon.TYPE).setBubbleVisible(true);
```

### Misc methods

The `updateBlockStyle` method is no longer available. Updating block styles is handled automatically, so you should be able to safely remove any references to this method. I removing this causes problems for your application, please [file a bug report](https://github.com/google/blockly/issues/new?assignees=&labels=issue%3A+bug%2C+issue%3A+triage&projects=&template=bug_report.yaml).

The `shouldIgnoreMutatorEvent` method is no longer available. This method was originally intended for internal use, and should not have been public. If you were using this method, we recommend duplicating the code from [here](https://github.com/google/blockly/pull/7115/files#diff-aeb28553e3f549c3f6fcdeb09b161c70732f73df152f7e7dccd36fe535fc49f0R275) into your own application.


core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+          oldExtraState,
+          newExtraState
+        )
+      );
+    }
+
+    eventUtils.setGroup(existingGroup);
+  }
+
+  /** @internal */
```
BeksOmega, line 305: After updating the tests and seeing where this is accessed, I'm wondering if it actually should be provided publicly. I just hate to provide it when I know all of the functionality is not supported.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+  static readonly TYPE = MUTATOR_TYPE;
+
+  /**
+   * The weight this icon has relative to other icons. Icons with more positive
+   * weight values are rendered farther toward the end of the block.
+   */
+  static readonly WEIGHT = 1;
+
+  /** The size of this icon in workspace-scale units. */
+  static readonly SIZE = 17;
```
BeksOmega, line 39: Make size class private instead of static

core/block_svg.ts
```diff
@@ -107,7 +107,7 @@ export class BlockSvg
   private warningTextDb = new Map<string, ReturnType<typeof setTimeout>>();
 
   /** Block's mutator icon (if any). */
-  mutator: Mutator | null = null;
```
BeksOmega, line 110: This should be deprecated.

cpcallen (changes requested):
Not quite finished reviewing all of `mutator_icon.ts`, but thought I should send you my comments so far.

cpcallen (changes requested):
Not quite finished reviewing all of `mutator_icon.ts`, but thought I should send you my comments so far.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+   * care about.
+   */
+  private shouldIgnoreMutatorEvent(e: Abstract) {
+    return (
+      e.isUiEvent ||
+      e.type === eventUtils.CREATE ||
+      (e.type === eventUtils.CHANGE &&
+        (e as BlockChange).element === 'disabled')
+    );
+  }
```
cpcallen, line 277: This doesn't seem to ned to be a method on `MutatorIcon`—certainly it could be a `static` method at least, or just a freestanding function.

Also, would `isIgnorableMutatorEvent` be a better name?

core/icons/mutator_icon.ts
cpcallen: I don't know what to think about this.  I can see arguments in both directions.  But now I'm wondering why `.miniWorkspaceBubble` is `private`.  Making it `protected` would aid subclassing—but I'd guessed it was public if I'd not gone and checked just now.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  /**
+   * Reconnects the given connection to the mutated input on the given block.
+   */
+  static reconnect(
+    connectionChild: Connection,
+    block: Block,
+    inputName: string
+  ): boolean {
```
cpcallen, line 317: Please provide JSDocs for parameters and return value.

(Actually, for all the methods in this class, but especially here were the meaning of the return value is not obvious.)

cpcallen: Also: my usual rule of thumb is that I only make a function `f` a static method on class `C` if it either takes one or more `C` instances as parameters, or returns a `C` instance, or otherwise at least mentions `C` somewhere in its body.  But this looks more like a random helper function that could be exported on its own (if indeed it is exported at all, though that might be useful for subclassing.)

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+      connectionParent.connect(connectionChild);
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Returns the parent workspace of a workspace that is inside a mutator,
+   * taking into account wither it is a flyout.
+   */
```
cpcallen, line 341: I think the terminology here needs to be updated.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+  static findParentWs(workspace: WorkspaceSvg): WorkspaceSvg | null {
+    let outerWs = null;
+    if (workspace && workspace.options) {
+      const parent = workspace.options.parentWorkspace;
+      // If we were in a flyout in a mutator, need to go up two levels to find
+      // the actual parent.
```
cpcallen, line 347: Is this dealing with the case that `workspace` is for the toolbox in a bubble?  Are there other doubly-nested workspaces to worry about?

(And is this function and it's usage correct in the situation of a toolbox-in-a-bubble on a  block that has a `MutatorIcon` _which is itself in a mutator bubble_?)

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Returns the parent workspace of a workspace that is inside a mutator,
+   * taking into account wither it is a flyout.
+   */
+  static findParentWs(workspace: WorkspaceSvg): WorkspaceSvg | null {
```
cpcallen, line 342: This seems to be called only from `blocks/procedures.js`; why not move it there—especially as it seems to have nothing to do with `MutatorIcon`?

cpcallen (commented):
Finished reviewing; had a few additional comments.

core/block_svg.ts
BeksOmega: I was in the middle of trying to get the comment icon refactored :P so I just left myself a note.

core/bubbles/bubble.ts
```diff
@@ -184,7 +184,7 @@ export abstract class Bubble implements IBubble {
   }
 
   /** Sets the colour of the background and tail of this bubble. */
-  protected setColour(colour: string) {
+  public setColour(colour: string) {
```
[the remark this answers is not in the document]
BeksOmega: Icons need to be able to set the colours of their bubbles to match their blocks, and I don't think there's any harm in making it configurable.

core/bubbles/mini_workspace_bubble.ts
```diff
@@ -196,6 +229,20 @@ export class MiniWorkspaceBubble extends Bubble {
     return new Size(width, height);
   }
 
+  /** Reapplies styles to all of the blocks in the mini workspace. */
+  updateBlockStyles() {
```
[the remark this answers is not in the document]
BeksOmega: If we update the styles of blocks in the main workspace, we want to update the styles of blocks in the mini workspace bubble as well. This update method is called by the icon, since the main workspace itself has no knowledge of mini workspace bubbles.

core/icons/icon_types.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+/** The type for a mutator icon. Used for registration and access. */
+export const MUTATOR_TYPE = 'mutator';
+
+/** The type for a warning icon. Used for registration and access. */
+export const WARNING_TYPE = 'warning';
+
+/** The type for a warning icon. Used for registration and access. */
+export const COMMENT_TYPE = 'comment';
```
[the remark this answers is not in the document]
BeksOmega: Hmmm, I added this just to get around circular dependencies, (before I was just accessing `MutatorIcon.TYPE`), so idk that it's a good idea to commit to this further.

I think my ideal is that `getIcon`, `removeIcon`, etc would take in a `Type` class like we use for the registry, so that the   methods can use/return the proper subclass types.

I would like to attempt refactoring that before the end of the quarter, but I don't know that I'll have time! So for now I'll leave the types the same, mark these constants as internal, make sure they're not exported, and file an issue for the above.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+   * The weight this icon has relative to other icons. Icons with more positive
+   * weight values are rendered farther toward the end of the block.
+   */
+  static readonly WEIGHT = 1;
+
+  /** The size of this icon in workspace-scale units. */
+  private readonly SIZE = 17;
+
+  /** The distance between the root block and the workspace edges. */
+  private readonly MARGIN = 16;
```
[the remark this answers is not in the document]
BeksOmega: I marked them as private because I didn't see a reason for them to be accessed outside of the class, as opposed to `WEIGHT` which could be used in the `getWeight` of other icons. E.g.

```js
class MyCustomIcon implements IIcon {
  getWeight() {
    return MutatorIcon.WEIGHT + 1;
  }

  // Other IIcon methods.
}
```

`WEIGHT` is not intended to be modifyable, which is why I have marked it as `readonly`.

Does the style guide have opinions on how to specify constants (I couldn't find anything)? I don't like taking up space on every instance, but I also don't like cluttering up our exports with useless properties.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+import {MiniWorkspaceBubble} from '../bubbles/mini_workspace_bubble.js';
+import {MUTATOR_TYPE} from './icon_types.js';
+import {Rect} from '../utils/rect.js';
+import {Size} from '../utils/size.js';
+import {Svg} from '../utils/svg.js';
+import type {WorkspaceSvg} from '../workspace_svg.js';
+
+export class MutatorIcon extends Icon implements IHasBubble {
+  /** The type string used to identify this icon. */
+  static readonly TYPE = MUTATOR_TYPE;
```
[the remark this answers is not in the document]
BeksOmega: See previous comment about dealing with types.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  /** The root block in the mini workspace. */
+  private rootBlock: BlockSvg | null = null;
+
+  /** The PID tracking updating the workkspace in response to user events. */
+  private updateWorkspacePid: ReturnType<typeof setTimeout> | null = null;
+
+  constructor(
+    private readonly flyoutBlockTypes: string[],
+    protected readonly sourceBlock: BlockSvg
```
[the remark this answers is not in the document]
BeksOmega: Switching the param order sgtm!

Wrt your other comment, `sourceBlock` is quite comment within our code base to denote "the block this doodad belongs to". E.g. inputs, fields, and connections use this consistently, so I think changing it for icons would end up being more confusing long-term.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,358 @@
...
+    return MutatorIcon.TYPE;
+  }
+
+  initView(pointerdownListener: (e: PointerEvent) => void): void {
+    if (this.svgRoot) return; // Already initialized.
+
+    super.initView(pointerdownListener);
+
+    // Square with rounded corners.
+    dom.createSvgElement(
```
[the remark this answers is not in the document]
BeksOmega: My understanding was that loading from a separate .svg file means more http requests or developers need to bundle their local `media` directory with our assets (since I don't think we bundle our own .svg assets).

But I'm not sure! For now I plan on keeping this logic the same, and if we want to refactor it separately, I can file an issue to track doing so.

core/icons/mutator_icon.ts
BeksOmega: Similar concerns about cluttering up our exports with static methods, but again no particular preference.

wrt `isIgnorableMutatorEvent` I tend to read things like: `if (this.shouldIgnoreMutatorEvent(e)) { doSomething(); }` as "if this (icon) should ignore (that) mutator event then do something" so "if this (event) is (an) ignorable mutator event" feels a bit strange to me.

But I'll go ahead and change it since `is`-prefixed boolean methods are more common!

core/icons/mutator_icon.ts
BeksOmega: Yeah I have no idea where this method should live. It is just a random helper method which is usually used with mutators.

I don't want to export it on `MutatorIcon` or `Blockly.icons` but idk where else to put it.

core/icons/mutator_icon.ts
BeksOmega: I will give this a look to see if this is procedure-specific, or if external developers could want to use it in a way that has nothing to do with procedures.

core/icons/mutator_icon.ts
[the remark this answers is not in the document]
BeksOmega: Ah whoops I forgot, the param order was actually to match the existing order for `Mutator`s so that we can use the renamings script to update people to the new `MutatorIcon`.

core/block_svg.ts
```diff
@@ -1035,26 +1035,16 @@ export class BlockSvg
...
-      mutator.createIcon();
-    }
-    if (this.rendered) {
-      // Icons must force an immediate render so that bubbles can be opened
-      // immedately at the correct position.
-      this.render();
-      // Adding or removing a mutator icon will cause the block to change shape.
-      this.bumpNeighbours();
-    }
+  override setMutator(mutator: MutatorIcon | null) {
```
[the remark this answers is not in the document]
BeksOmega: I'll rename this to `setMutatorIcon` and update the renamings database.

core/icons/mutator_icon.ts
BeksOmega: I think this is the only doubly nested workspace case.

core/icons/mutator_icon.ts
BeksOmega: I don't think this is procedure specific. Procedures use this to change something about the outer workspace in from a field validator on one of the mutator / mini workspace blocks, which could be done by non-procedure blocks.

Unless you can think of a third option for where to export it from, I think it makes more sense on `MutatorIcon` than it does in `procedures` (even though having it on `MutatorIcon` isn't great).

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,361 @@
...
+
+  getWeight(): number {
+    return MutatorIcon.WEIGHT;
+  }
+
+  getSize(): Size {
+    return new Size(this.SIZE, this.SIZE);
+  }
+
+  applyColour(): void {
```
BeksOmega, line 120: Call super.

core/icons/mutator_icon.ts
```diff
@@ -0,0 +1,361 @@
...
+  }
+
+  updateCollapsed(): void {
+    super.updateCollapsed();
+    if (this.sourceBlock.isCollapsed()) this.setBubbleVisible(false);
+  }
+
+  onLocationChange(blockOrigin: Coordinate): void {
+    super.onLocationChange(blockOrigin);
+    if (this.bubbleIsVisible()) {
```
BeksOmega, line 132: Remove check

core/icons/mutator_icon.ts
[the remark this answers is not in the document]
BeksOmega: In an ideal world I want to get https://github.com/google/blockly/issues/7126 in by the end of the quarter, so that it wouldn't be a breaking change. 

In the case that this is not possible, all of the `IIcon` implementations will have a `static readonly TYPE: string` property, which can be passed to `getIcon`, `removeIcon`, etc. I think this is a reasonable implementation, even if it's not ideal from a type perspective.

I don't think the advantage of having it be a `static readonly TYPE: IconType` in the case of not implementing #7126 by EOQ is worth taking the time to implement it.

So my plan is to leave as-is for now (since this gets everything to build) then complete #7126 after #7115, #7128, and #7112 are merged.

IDK if that answers your comment :P but hopefully it clarifies!

core/connection.ts
```diff
@@ -333,6 +333,36 @@ export class Connection implements IASTNodeLocationWithBlock {
...
+   *
+   * @param block The block to connect this connection to.
+   * @param inputName The name of the input to connect this connection to.
+   * @returns True if this connection was able to connect, false otherwise.
+   */
+  reconnect(block: Block, inputName: string): boolean {
+    // No need to reconnect if this connection's block is deleted.
+    if (this.getSourceBlock().isDeadOrDying()) return false;
+
+    const connectionParent = block.getInput(inputName)!.connection;
```
cpcallen, line 349: Wouldn't `?.` be safe here?  It seems to me that (especially now this is a method on `Connection` rather than just a local helper function) there's no guarantee there will be an input with the given name.

Of course perhaps you _want_ it to throw if the named input does not exist—but in that case I think an explicit check and `throw new SomeKindOfError('useful message')` would be in order.

core/connection.ts
```diff
@@ -333,6 +333,36 @@ export class Connection implements IASTNodeLocationWithBlock {
...
+    const currentParent = this.targetBlock();
+    if (
+      (!currentParent || currentParent === block) &&
+      connectionParent &&
+      connectionParent.targetConnection !== this
```
cpcallen, line 354: Can't this be:

```suggestion
      connectionParent?.targetConnection !== this
```

core/workspace.ts
```diff
@@ -804,6 +804,28 @@ export class Workspace implements IASTNodeLocation {
...
   }
 
+  /**
+   * Returns the root workspace of this workspace if the workspace has
+   * parent(s).
+   *
+   * E.g. workspaces in flyouts and mini workspace bubbles have parent
+   * workspaces.
+   */
+  getRootWorkspace(): Workspace | null {
```
cpcallen, line 814: Nit: might `getParentWorkspace` or `getSoureWorkspace` or `getOuterWorkspace` be a better name, since this is returns the next-enclosing workspace, not (as previously discussed) necessarily the root?

scripts/migration/renamings.json5
```diff
@@ -1447,5 +1447,9 @@
 
   'develop': [
     // New renamings go here!
+    {
+      oldName: 'Blockly.Mutator',
+      newName: 'Blockly.icons.MutatorIcon',
+    },
```
cpcallen, line 1453: This is total pedantry because it won't actually affect the behaviour of the renaming script as presently written, but since the export is being renamed as well as the module:

```suggestion
    {
      oldName: 'Blockly.Mutator',
      newName: 'Blockly.icons.MutatorIcon',
      exports: {
        Mutator: {
          newExport: 'MutatorIcon',
          oldPath: 'Blockly.Mutator',
          newPath: 'Blocky.icons.MutatorIcon',
        },
      },
    },
```

cpcallen (approved):
Just a few small nits which I will leave for you to resolve as you see fit.

core/connection.ts
BeksOmega: We need to evaluate `connectionParent` explicitly so that typescript knows it's defined within the body of the `if`.

core/workspace.ts
BeksOmega: `getParent` workspace doesn't quite work because for flyouts in a mini workspace bubble, their parent workspace is the workspace of the mini workspace bubble, but we actually return the workspace /containing/ the mini workspace bubble.

I think it's close enough to the truth to say "this always returns the root workspace" because further nesting than that is not supported. If we were to add support for further nesting (e.g. mini workspace bubbles inside eachother), updating this method would be part of that.

core/icons/mutator_icon.ts
BeksOmega: Haha turns out that `getWorkspace` is currently internal, so no reason to start providing it unless we get an ask.

PR #7116, opened 2023-05-24
fix: have inputs construct connections

### LINKED ISSUE #2062: Add support for custom inputs

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #2062 

### Proposed Changes

Makes it so that inputs must + can easily construct their own connections (for inputs that have connections).

#### Behavior Before Change

Inputs were passed their connections, except for custom inputs which weren't. It was difficult for custom inputs to construct their own connections, because they had to deal with rendered vs headless.

#### Behavior After Change

All inputs must construct their own connections. They can easily do this using the `protected makeConnection` method.

### Reason for Changes

This makes built-in inputs and custom inputs more consistent. It also makes custom inputs with connections easier to implement.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -1569,13 +1569,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @returns The input object created.
    */
   appendValueInput(name: string): Input {
-    return this.appendInput(
-      new ValueInput(
-        name,
-        this,
-        this.makeConnection_(ConnectionType.INPUT_VALUE)
-      )
-    );
+    return this.appendInput(new ValueInput(name, this));
   }
 
   /**
@@ -1587,13 +1581,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    */
   appendStatementInput(name: string): Input {
     this.statementInputCount++;
-    return this.appendInput(
-      new StatementInput(
-        name,
-        this,
-        this.makeConnection_(ConnectionType.NEXT_STATEMENT)
-      )
-    );
+    return this.appendInput(new StatementInput(name, this));
   }
 
   /**
@@ -1633,7 +1621,7 @@ export class Block implements IASTNodeLocation, IDeletable {
       false
     );
     if (!inputConstructor) return null;
-    return this.appendInput(new inputConstructor(name, this, null));
+    return this.appendInput(new inputConstructor(name, this));
   }
 
   /**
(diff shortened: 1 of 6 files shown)


core/block.ts
```diff
@@ -2310,8 +2298,9 @@ export class Block implements IASTNodeLocation, IDeletable {
    *
    * @param type The type of the connection to create.
    * @returns A new connection of the specified type.
+   * @internal
    */
-  protected makeConnection_(type: number): Connection {
+  public makeConnection_(type: number): Connection {
```
maribethb, line 2303: Doesn't need to be marked public since that's the default (same below)

core/inputs/input.ts
```diff
@@ -298,6 +293,10 @@ export class Input {
       this.connection.dispose();
     }
   }
+
+  protected makeConnection(type: number): Connection {
```
maribethb, line 297: Just curious, why isn't the type of `type` `ConnectionType`?

maribethb, line 297: Consider adding tsdoc for this method to explain to implementers of custom inputs when they should call this method.

maribethb (approved):
Reviewed since Rachel is ooo

core/inputs/input.ts
maribethb: upon rereading this sentence: @_@

BeksOmega: Fixed!

On the block version of this method we used to pass `inputTypes` to it sometimes, which isn't compatible with `ConnectionType`, so we had to use `number` instead. But since the code has been reorganized, that is no longer an issue =) Everything has been switched to use `ConnectionType`!

maribethb (approved):

PR #7117, opened 2023-05-24
fix!: move destroy earlier in the lifcycle

### LINKED ISSUE #6974: Move `destroy` hook to beginning of `dispose`
### Check for duplicates

### Problem

Sometimes the references that external developers want to clean up in a blocks `destroy` hook are keyed based on other properties of the block. E.g. field input values. Currently we call `destroy` at the very end of `dispose`, after fields (and etc) have already been cleaned up, so they can't use these values to clean up references.

### Request

Call `destroy` first (or nearer to the beginning of the cycle) before the rest of the block has been disposed.

### Alternatives considered

N/A

### Additional context

I don't /think/ this would be a breaking change. But I'd need to noodle on it a bit more to see if it would cause any major problems.

Originally reported by @mark-friedman related to the lexical variables plugin.

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #6974 

### Proposed Changes

Moves the `destroy` hook call up in the life cycle. 

#### Behavior Before Change

Before the destroy hook was called at the very end of disposal, so it couldn't depend on any of the other state of the block.

#### Behavior After Change

Now it is called after the block is disconnected from its parent, but before any of its elements or child blocks are disposed. So (e.g.) the `destroy` hook could depend on the value of the block's fields.

### Reason for Changes

If you want to destroy external state that is associated with field values, you need access to the field values in the `destroy` hook.

### Test Coverage

N/A

### Documentation

Docs need to be updated for v10.

### Additional Information

N/A

We don't know of any way that this could actually break people, since this is only giving folks access to more data. However, as a precaution we are marking this as a breaking change, so people can check the source code to see exactly what moved.


maribethb (approved):
Can you update the PR description in the bottom section you added to be speaking directly to people who view this PR from the release notes? Coming from that mindset, it's not clear what my action item is here. e.g.

If you implement the `destroy` hook, then note that this hook will now be run before the connections and fields associated with this block are destroyed. This is unlikely to actually require changes in your project unless your `destroy` hook was explicitly relying on the old timing.

BeksOmega:
Also @mark-friedman can you check that this timing would work for you?

mark-friedman:
> Also @mark-friedman can you check that this timing would work for you?

I'll check.

core/block.ts
```diff
@@ -343,23 +343,17 @@ export class Block implements IASTNodeLocation, IDeletable {
...
-      this.workspace.removeTypedBlock(this);
-      this.workspace.removeBlockById(this.id);
-      this.disposing = true;
-
-      this.childBlocks_.forEach((c) => c.disposeInternal());
-      this.inputList.forEach((i) => i.dispose());
-      this.inputList.length = 0;
-      this.getConnections_(true).forEach((c) => c.dispose());
-    } finally {
-      eventUtils.enable();
```
BeksOmega, line 357: I don't think we need to disable events any longer since the delete event is fired by `dispose` and not `disposeInternal`, which is what we use to recurse.

Said another way, only the top level block fires a delete event, so we have no need to disable events for child blocks.

Here is the context for originally disabling events: https://github.com/google/blockly/commit/c429949a4e449fc84e46f79e8eba55f5df9c06cb

BeksOmega:
I had to move things around to get tests to pass :/ Would you mind doing a re-review @maribethb ? No rush on this though.

mark-friedman:
> Also @mark-friedman can you check that this timing would work for you?

It looks like it would work.  Thanks!

mark-friedman:
Might I also suggest that you change the documentation for the `destroy` method so that it is clear to the developer that they can count on the block being pretty much intact when their callback is called?

maribethb (approved):

PR #7118, opened 2023-05-24
fix: return annotations

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

Reenables tag linting so that we don't commit `@return` annotations.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

Assigning christopher so he can check if removing the `@suppress` or `@nocollapse` tags breaks anything.


maribethb (approved):
I ran the advanced compilation test and everything worked without `@suppress`

also you can use @file instead of @fileoverview but the comments you removed aren't adding anything of value so that's fine

Issue #7126, opened 2023-05-31
Have `getIcons` take in a generic `Type` type that allows it to return a specific icon type


BeksOmega:
Currently the `getIcons` method returns an `IIcon` which means that if you want to retrieve a specific icon an call methods on it, you have to cast to that specific type of icon.

For example:
```
(block.getIcon(CommentIcon.TYPE) as CommentIcon).setBubbleSize(new Size(150, 150));
```

It would be much better if you could do:
```
block.getIcon(CommentIcon.TYPE).setBubbleSize(new Size(150, 150));
```

and have proper type checking.

This could be done by having the `.TYPE` property be a generic type, similar to how the registry works.

For example:

```
class CommentIcon {
  static TYPE = new IconType<CommentIcon>('comment');
}

class Block {
  getIcon<T>(type: IconType<T> | string): T;
}
```

cpcallen:
An alternative would be to use literal types for the argument, as [TypeScript does for certain DOM methods like `Document.createElement`](https://www.typescriptlang.org/docs/handbook/dom-manipulation.html#the-document-interface), where the return type depends on the actual literal value passed in.

I'm not sure that that is as extensible as Beka's proposal, though.

PR #7128, opened 2023-05-31
fix!: refactor comment icon

### LINKED ISSUE #7038: Refactor the Comment icon to conform to the IIcon interface
### Description

Part of #2105 

Comments should have the following additional methods:
  * setText
  * getText
  * setBubbleSize
  * getBubbleSize

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7038 

### Proposed Changes

Refactors the comment icon to conform to the `IIcon` interface.

### Reason for Changes

Code consistency :D

### Test Coverage

Updated tests! Also manual testing.

### Documentation

N/A

### Additional Information

N/A

The `Blockly.Comment` class has been renamed to `Blockly.icons.CommentIcon`. If you use the blockly migration package, you can automatically rename this throughout your files by doing:

```
npx @blockly/migrate rename --from 9 --in-place ./path/to/my/files*
```

### Access comment icons

The `Block.prototype.comment` property has been removed. If you need to access a comment icon directly you should use:

```js
const comment = myBlock.getIcon(Blockly.icons.CommentIcon.TYPE);
```

### Showing/hiding bubbles

The `setVisible` method has been renamed to `setBubbleVisible` for clarity. If you need to modify whether the bubble of the comment icon is visible or not, you can do:

```js
myBlock.getIcon(Blockly.icons.CommentIcon.TYPE).setBubbleVisible(true);
```


tests/mocha/block_test.js
```diff
@@ -1294,37 +1306,23 @@ suite('Blocks', function () {
         });
         test('Set While Visible - Editable', function () {
           this.block.setCommentText('test1');
-          const icon = this.block.getCommentIcon();
-          icon.setVisible(true);
+          const icon = this.block.getIcon(Blockly.icons.CommentIcon.TYPE);
```
maribethb, line 1309: why not use the `getCommentIcon` method? makes the test more readable imo, you can have a separate test for that method if you want to test it

tests/mocha/comment_test.js
```diff
@@ -31,8 +31,8 @@ suite('Comments', function () {
       Blockly.utils.xml.textToDom('<block type="empty_block"/>'),
       this.workspace
     );
-    this.comment = new Blockly.Comment(this.block);
-    this.comment.computeIconLocation();
+    this.comment = new Blockly.icons.CommentIcon(this.block);
+    // this.comment.computeIconLocation();
```
maribethb, line 35: should this be removed?

core/icons/comment_icon.ts
```diff
@@ -0,0 +1,326 @@
...
+  static readonly TYPE = COMMENT_TYPE;
+
+  /**
+   * The weight this icon has relative to other icons. Icons with more positive
+   * weight values are rendered farther toward the end of the block.
+   */
+  static readonly WEIGHT = 3;
+
+  /** The size of this icon in workspace-scale units. */
+  private readonly SIZE = 17;
```
maribethb, line 35: Why is the weight static but the size isn't? They seem pretty similar in concept to me.

core/icons/comment_icon.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  /** The default width in workspace-scale units of the text input bubble. */
+  private readonly DEFAULT_BUBBLE_WIDTH = 160;
+
+  /** The default height in workspace-scale units of the text input bubble. */
+  private readonly DEFAULT_BUBBLE_HEIGHT = 80;
```
maribethb, line 41: nit: I don't think non-static properties of a class meet the requirements to be constant case: "Global: Only symbols declared on the module level, static fields of module level classes, and values of module level enums, may use CONST_CASE."

core/icons/icon_types.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/** The type for a mutator icon. Used for registration and access. */
+export const MUTATOR_TYPE = 'mutator';
```
maribethb, line 8: Why do we need these defined here and in the respective file?

The only thing I can think of is that people would be allowed to register their own CommentIcon class for Blockly to use, and maybe it makes sense to provide a string they can use as the type for registration here so they don't have to use the one in `CommentIcon`. But a) if they can provide their own `CommentIcon` it should probably be a subclass of ours, so they could just rely on inheritance b) if developers are supposed to use this string, I think you need to export it from Blockly and I don't see where that's happening.

maribethb (commented):
please also fix the lint warnings (it won't let me comment near them?)

maribethb:
Also it looks like a lot of the refactoring isn't breaking because a lot of this was private to begin with. If the breaking changes are limited to the actual removal of `Blockly.Comment` (and `this.comment` in `Block` if you do that) then consider moving that to a separate PR for ease of understanding in release notes. I might have missed some of the breaking changes though so if there are others that would be hard to separate out totally fine to leave them together in one PR as that would be more clear in that case.

core/xml.ts
```diff
@@ -188,8 +190,9 @@ export function blockToDom(
 
   const commentText = block.getCommentText();
   if (commentText) {
-    const size = block.commentModel.size;
-    const pinned = block.commentModel.pinned;
+    const comment = block.getIcon(COMMENT_TYPE) as CommentIcon;
```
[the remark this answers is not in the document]
BeksOmega: See here for context about icon types: https://github.com/google/blockly/pull/7115#discussion_r1211754816

Essentially, I think `CommentIcon.TYPE` is the better API, but it breaks our build.

tests/mocha/comment_test.js
BeksOmega: Yes, thanks for the catch!

core/icons/comment_icon.ts
BeksOmega: See discussion here: https://github.com/google/blockly/pull/7115#discussion_r1205992681As per Christopher's suggestion, I'm going to make this module-const.

BeksOmega:
> Also it looks like a lot of the refactoring isn't breaking because a lot of this was private to begin with. If the breaking changes are limited to the actual removal of `Blockly.Comment` (and `this.comment` in `Block` if you do that) then consider moving that to a separate PR for ease of understanding in release notes. I might have missed some of the breaking changes though so if there are others that would be hard to separate out totally fine to leave them together in one PR as that would be more clear in that case.

Good idea! I'm going to:
1) Wait until all of the Icon refactor PRs are approved so that suggested changes from one can be pulled into the others.
2) Write up the breaking change messages.
3) See if any breakages would be better pulled into separate PRs (e.g. I think the deletions would be a good case like you suggest).
4) Merge the Icon refactor PRs.

core/block_svg.ts
```diff
@@ -906,45 +906,10 @@ export class BlockSvg
    *
    * @returns The comment icon attached to this block, or null.
    */
-  getCommentIcon(): Comment | null {
+  getCommentIcon(): CommentIcon | null {
     return this.commentIcon_;
```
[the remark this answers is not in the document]
BeksOmega: Chatted offline! Plan is to solve the readability problem via #7126

core/xml.ts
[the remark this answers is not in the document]
maribethb: iiuc, this one in particular isn't likely to break the build (you also import `CommentIcon` here so I would be confused about how) but the mutator icon is probably the one that broke it and you're doing this here to be consistent. is that right? if so, that's fine, just trying to understand what's happening.

maribethb (approved):

core/block.ts
```diff
@@ -2173,9 +2163,12 @@ export class Block implements IASTNodeLocation, IDeletable {
    * Returns the comment on this block (or null if there is no comment).
    *
    * @returns Block's comment.
+   * @deprecated Use getIcon. To be remove in v11.
    */
   getCommentText(): string | null {
-    return this.commentModel.text;
+    deprecation.warn('getCommentText', 'v10', 'v11', 'getIcon');
```
maribethb, line 2169: I find it strange we are deprecating `getCommentText` and not `setCommentText`, and would probably prefer to deprecate them at the same time, but awaiting your thoughts

core/xml.ts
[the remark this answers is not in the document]
BeksOmega: This one also creates circular dependencies :/ But I'm not sure why.

Wrt the import of `CommentIcon`, it doesn't create circular dependencies itself b/c it's only used as a type, which means it gets removed from the tsc output. I've changed it to `import type` to make this clear.

Accessing `CommentIcon.TYPE` means that you're using it as a value, which triggers the circular dependencies.

I'm imagining this is happening now (and not before) because `Xml` is importing `CommentIcon` directly, while before `Comment` was only imported by `BlockSvg`. But beyond that I'm not sure what the exact cycle is.

blocks/procedures.js
```diff
@@ -38,7 +38,7 @@ const {VariableModel} = goog.requireType('Blockly.VariableModel');
 const {Workspace} = goog.requireType('Blockly.Workspace');
 const {defineBlocks} = goog.require('Blockly.common');
 /** @suppress {extraRequire} */
-goog.require('Blockly.Comment');
+goog.require('Blockly.icons.CommentIcon');
```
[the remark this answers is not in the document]
BeksOmega: @cpcallen what is better for you here wrt merge conflicts?

core/block.ts
BeksOmega: Discussed offline, undeprecated `setCommentText`!

core/serialization/blocks.ts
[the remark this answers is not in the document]
BeksOmega: I also /actually/ changed it to error this time!

blocks/procedures.js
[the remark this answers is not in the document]
BeksOmega: Talked offline, we're stating the that goog module of the new icon matches the old icon to minimize conflicts.

PR #7132, opened 2023-06-05
feat: add types for accessing icons.

### LINKED ISSUE #7126: Have `getIcons` take in a generic `Type` type that allows it to return a specific icon type

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7126 

### Proposed Changes

Has `getIcons` and other relevant methods use an generic `IconType<T>` class to reference/retrieve icons so that we return icons of the correct type.

### Reason for Changes

Make typescript happy :D

### Test Coverage

Fixed tests.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -2228,7 +2228,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @param type The type of the icon to remove from the block.
    * @returns True if an icon with the given type was found, false otherwise.
    */
-  removeIcon(type: IconType<any>): boolean {
+  removeIcon(type: IconType<IIcon>): boolean {
     if (!this.hasIcon(type)) return false;
     this.getIcon(type)?.dispose();
     this.icons = this.icons.filter((icon) => !icon.getType().equals(type));
@@ -2239,7 +2239,7 @@ export class Block implements IASTNodeLocation, IDeletable {
    * @returns True if an icon with the given type exists on the block,
    *     false otherwise.
    */
-  hasIcon(type: IconType<any>): boolean {
+  hasIcon(type: IconType<IIcon>): boolean {
     return this.icons.some((icon) => icon.getType().equals(type));
   }
 
--- a/core/block.ts
+++ b/core/block.ts
@@ -49,6 +49,7 @@ import type {Workspace} from './workspace.js';
 import {DummyInput} from './inputs/dummy_input.js';
 import {ValueInput} from './inputs/value_input.js';
 import {StatementInput} from './inputs/statement_input.js';
+import {IconType} from './icons/icon_types.js';
 
 /**
  * Class for one block.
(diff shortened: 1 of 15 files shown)


core/block.ts
```diff
@@ -2227,28 +2228,33 @@ export class Block implements IASTNodeLocation, IDeletable {
...
     return true;
   }
 
   /**
    * @returns True if an icon with the given type exists on the block,
    *     false otherwise.
    */
-  hasIcon(type: string): boolean {
-    return this.icons.some((icon) => icon.getType() === type);
+  hasIcon(type: IconType<any>): boolean {
```
maribethb, line 2242: Can this (here and elsewhere) use `IIcon` instead of `any`? I am not sure if that's legal TS or not.

core/interfaces/i_icon.ts
```diff
@@ -13,7 +14,7 @@ export interface IIcon {
    * E.g. 'comment', 'warning', etc. This string should also be used when
```
maribethb, line 14: I think this comment needs to be updated

core/serialization/blocks.ts
```diff
@@ -709,7 +709,7 @@ function initBlock(block: Block, rendered: boolean) {
       if (isIcon(icon)) {
         icon.onLocationChange(blockSvg.getRelativeToSurfaceXY());
       } else {
-        icon.computeIconLocation();
+        // icon.computeIconLocation();
```
maribethb, line 712: Is this change supposed to be here?

core/xml.ts
```diff
@@ -190,7 +190,7 @@ export function blockToDom(
 
   const commentText = block.getCommentText();
   if (commentText) {
-    const comment = block.getIcon(COMMENT_TYPE) as CommentIcon;
+    const comment = block.getIcon(IconType.COMMENT) as CommentIcon;
```
maribethb, line 193: I think you should be able to remove the casts now right? (here and below)

maribethb (approved):

core/xml.ts
BeksOmega: Done!

core/serialization/blocks.ts
BeksOmega: Whoops, for a second I was gonna do this and #7141 together, but decided to split them out. Looks like this got left behind.

core/interfaces/i_icon.ts
BeksOmega: Done!

core/block.ts
BeksOmega: Done!

PR #7136, opened 2023-06-06
fix: gestures problems

### LINKED ISSUE #7134: Bug pinch zoom more than two fingers infinite loop
### Check for duplicates

### Description

When you are in a pinch zoom, and you have more than two fingers on the screen, an infinite loop occurs.

### Reproduction steps

1. Start a pinch zoom
2. Add a third finger

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

Android mobile, iOS mobile

### LINKED ISSUE #7137: Gesture: Block not released when changing app
### Check for duplicates

### Description

The selected block by the gesture is always selected when you press the home screen on android and ios. This is because "the touhcCancel" event is not taken into account.

### Reproduction steps

1. Drag a block
2. At the same time as moving the block (keep your finger pressed), press Android's home button with another finger
3. Reopen the browser

### Stack trace

```shell
N/A
```

### Screenshots

_No response_

### Browsers

Android mobile, iOS mobile

Plomging:
## The basics

- [x] I branched from develop

### Resolves
Fixes #7134
Fixes #7137

### Proposed Changes
Remove the problematic line of code that generates this loop

Add a listener to the event "touchcancel" who executes "handleUp" function

#### Behavior Before Change
Infinite loop because "handleTouchMove" call "handleMove" that call "handleTouchMove" and so on...

The block stays selected

#### Behavior After Change
Infinite loop removed.

The block is deselected like a "pointerup" occurs.

### Reason for Changes
The infinite loop causes the browser to crash

Impossible to use the interface after an app switch (android home button)

### Test Coverage
Manual tests

### Documentation
No

### Additional Information
No


core/gesture.ts
```diff
@@ -691,8 +700,6 @@ export class Gesture {
 
     if (this.isPinchZoomEnabled && this.cachedPoints.size === 2) {
       this.handlePinch(e);
-    } else {
-      this.handleMove(e);
```
BeksOmega, line 695: I think we still want to handle the move in the case where pinch to zoom is disabled! Could you add a check to this else that `this.cachedPoints.size < 3` so if there are 1 or two fingers, we still perform the drag?

BeksOmega (changes requested):
Generally LGTM! Just one fix =)

Thank you so much for putting up this PR! Really appreciate it :D

Precha24 (commented):
Good, thanks

BeksOmega:
Hi @Plomging ! Thanks for updating your branch! Now this just needs the change I suggested in my previous review, and then it should be good to go =)

core/gesture.ts
Plomging: Hi, for me the "handleMove" have nothing to do here in any case. Because if it's called, it automatically generates an infinite loop because "handleTouchMove" is called by "handleMove"... Here's an example with pinch zoom disabled as you proposed.

1. First finger down
    a. Calls "BindMouseEvents" => Making the next finger to call "handleStart"
    b. Calls "HandleTouchStart" => Register the finger in "cachedPoints"
2. Second finger down (without moving the first)
    a. Calls "handleStart" => (called because of the bindMouse call previously)
    b. Calls "HandleTouchStart" => Register the finger in "cachedPoints" and set isMultiTouch_ = true
3. Move a finger
    a. Calls "handleMove" (because of BindMouseEvents)
    b. Calls "handleTouchMove" because we are in multi touch and not dragging
    c. In your case (this.cachedPoints.size < 3 is true) so it calls "handleMove"
    d. In "handleMove" we have the same conditions as before, the multi touch condition is executed, it calls "handleMultiTouch" and so on... Infinite loop

BeksOmega:
Superceded by #7243

PR #7151, opened 2023-06-13
feat: use new intermediate block change event for field edits, normal block change event for closing editor. #7105

### LINKED ISSUE #2496: Make onFinishEditing_ Better

johnnesky:
Another proposal to fix #2496. Creates a new event type specifically for intermediate field value changes, which is not recorded in undo history. Inhibits firing normal block change events for field value edits when triggered via a user input widget, and instead broadcasts this new event type. When the input widget is disposed, then a standard block change event will be fired with an oldValue corresponding to what the field's value was when the widget was opened.

Compare with: https://github.com/google/blockly/pull/7064/ and https://github.com/google/blockly/pull/7105

Unlike https://github.com/google/blockly/pull/7105 this version of the new event type inherits from BlockBase instead of UiBase. The difference is minor and only affects one file: events_block_field_intermediate_change.ts


core/events/events_block_field_intermediate_change.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Notifies listeners when the value of a block's field has changed but the
+ * change is not yet complete, and is expected to be followed by a block change
+ * event.
+ */
+export class BlockFieldIntermediateChange extends BlockBase {
+  override type = eventUtils.BLOCK_FIELD_INTERMEDIATE_CHANGE;
+
+  // Intermediate events do not undo or redo.
```
BeksOmega, line 31: Can you add some info here about why intermediate events don't record undo? I think we'll be able to dig up your docs about this in the future, but might be nice to just give future people a brief summary!

core/field.ts
```diff
@@ -1042,9 +1042,12 @@ export abstract class Field<T = any>
    * than this method.
    *
    * @param newValue New value.
+   * @param fireChangeEvent Whether to fire a change event. Defaults to true.
+   *     Should usually be true unless the change will be reported some other
+   *     way.
```
BeksOmega, line 1047:
```suggestion
   *     way. E.g. an intermediate field change event.
```

core/field_angle.ts
BeksOmega, line 1: Can you add this change in samples as well? Since we're migrating the [angle field over there](https://github.com/google/blockly-samples/tree/master/plugins/field-angle) (we're leaving it in both places for a month or two to give people time to transition).

If not I can grab this since I'm testing samples this quarter anyway =)

BeksOmega (approved):
Thank you for the incredible amount of work on this John!! Just a few comments about inline comments, then LGTM =)

core/events/events_block_field_intermediate_change.ts
johnnesky: Done.

core/field.ts
johnnesky: Done.

core/field_angle.ts
johnnesky: Probably not until this is committed and blockly-samples depends on the latest version, but I created an issue to do so: https://github.com/google/blockly-samples/issues/1717

BeksOmega (approved):

PR #7155, opened 2023-06-13
chore!: remove angle field from core

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes https://github.com/google/blockly-samples/issues/1644

### Proposed Changes

Removes the angle field from core.

### Reason for Changes

The angle field has been moved into samples.

### Test Coverage

This may break the test blocks, but that needs changes in dev tools. Need to check.

If you are using the built in angle field, you will now need to import it from the [separate npm package](https://www.npmjs.com/package/@blockly/field-angle). This helps us keep core small and modular!

All references to `Blockly.FieldAngle` will need to be renamed to `FieldAngle`.

### Definitions

Here is an example of defining a block with an angle field in JavaScript:
```javascript
import * as Blockly from 'blockly';
import {FieldAngle} from '@blockly/field-angle';

Blockly.Blocks["test_field_angle"] = {
  init: function () {
    this.appendDummyInput()
      .appendField("angle: ")
      .appendField(new FieldAngle(90), "FIELDNAME");
  }
};
```

And here is an example of defining a block with an angle field in JSON:
```javascript
import * as Blockly from 'blockly';
import '@blockly/field-angle';

Blockly.defineBlocksWithJsonArray([
    {
        "type": "test_field_angle",
        "message0": "angle: %1",
        "args0": [
            {
                "type": "field_angle",
                "name": "FIELDNAME",
                "value": 50
            }
        ]
    }]);
```

### API Changes

TODO


tests/mocha/serializer_test.js
cpcallen, line 1: These tests don't seem to have been moved to the plugin.  Is that feasible?  Seems like it might be a good idea if so.

cpcallen (approved):

cpcallen:
Test failure looked at first glance like a timeout, so I kicked it but evidently that was not sufficient; I'll leave it to you to investigate further (but do ask if you need help.)

tests/mocha/serializer_test.js
BeksOmega: I'll file a bug about moving the serialization tests into samples!

PR #7159, opened 2023-06-13
fix: icon exports

### LINKED ISSUE #2105: Icon API

### LINKED ISSUE #7082: Confirm custom icon exports are correct before release

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #2105 
Fixes #7082 

### Proposed Changes

Makes sure that we are exporting the correct icon-related things.

### Reason for Changes

Exporting things is like our whole job :P

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

I updated https://github.com/google/blockly/pull/7060 to be a breaking change, and added docs about upgrading there.


scripts/migration/renamings.json5
```diff
@@ -1478,5 +1478,9 @@
         },
       },
     },
+    {
+      oldName: 'Blockly.Icon',
+      newName: 'Blockly.icons.Icon',
+    }
```
BeksOmega, line 1484: @cpcallen I think this is correct, b/c the export didn't get renamed, but if you could verify that would be appreciated!

cpcallen: There doesn't seem to be a goog.module ID for this module, so strictly speaking there's not a correct way to represent this in the renamings file as it is presently structured, but this will have the desired effect.

PR #7165, opened 2023-06-14
fix: input exports

### LINKED ISSUE #7083: Confirm custom input exports are correct before release
Before doing the quarterly release, confirm that the exports related to custom inputs are correct. In particular:
  * Ensure that everything is exported from the Blockly.inputs namespace, and not exported individually.
  * Ensure the renamings script is up to date.

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7083 

### Proposed Changes

Makes sure that we are exporting the correct input-related things.

### Reason for Changes

Exporting things is like our whole job :P

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

## Deprecations

If you were using the `Blockly.Input.Align` enum, it is now exported under `Blockly.inputs.Align`. Similarly these constants have moved:
  * `Blockly.ALIGN_LEFT` -> `Blockly.inputs.Align.LEFT`
  *  `Blockly.ALIGN_CENTER` -> `Blockly.inputs.Align.CENTER`
  *  `Blockly.ALIGN_RIGHT` -> `Blockly.inputs.Align.RIGHT`

If you were using the `Blockly.inputTypes` enum, it is now exported under `Blockly.inputs.inputTypes`.

You can use the [blockly migration script](https://www.npmjs.com/package/@blockly/migrate) to automatically perform these renames for you =)


scripts/migration/renamings.json5
```diff
@@ -1478,5 +1478,36 @@
...
+          newPath: 'Blockly.inputs.Align.CENTER',
+        },
+        'ALIGN_RIGHT': {
+          newPath: 'Blockly.inputs.Align.RIGHT',
+        },
+        'inputTypes': {
+          newModule: 'Blockly.inputs',
+        },
+      },
+    },
```
BeksOmega, line 1511: @cpcallen can you check these renamings for me as well?

core/inputs/input.ts
```diff
@@ -310,13 +310,17 @@ export namespace Input {
   /**
    * Enum for alignment of inputs.
    *
+   * @deprecated Use Blockly.inputs.Align. To be removed in v11.
    */
+  /** @suppress {deprecated} */
```
maribethb, line 315: Where is the new one? It looks like `Blockly.inputs.Align` currently would point to this enum here. That's going to make the reference docs confusing, if the new path you're telling them to use says it's deprecated.

Maybe we need a new `align.ts` file that exports the enum so that this one can truly be deprecated?

maribethb (commented):
Is there actually a breaking change in here? It looks like all deprecations to me.

Also in your PR description, "if you were using the `Align` enum" -- I think this should say "if you were using the `Blockly.Input.Align` enum" but want to double check that's correct

scripts/migration/renamings.json5
cpcallen: @BeksOmega: I do not see any change to the `goog.declareModueId()` calls, so none of these should have a `newName` at the module level—only new paths.

I will send you a PR to fix.

cpcallen: (Oh, this has not been merged yet, so I will add a commit.)

scripts/gulpfiles/build_tasks.js
```diff
@@ -231,6 +229,8 @@ const JSCOMP_ERROR = [
  * it's generally sufficient to remove them from JSCOMP_ERROR.
  */
 const JSCOMP_WARNING = [
+  'deprecated',
+  'deprecatedAnnotations',
```
cpcallen, line 233: This is going to lead to seeing warnings during build.  It that what we want?

More broadly, why do we have non-deprecated code calling `@deprecated` code?

cpcallen (commented):
I've added commit [6ac0570](https://github.com/google/blockly/pull/7165/commits/6ac057060188fc92c0239adb29a9965164c91bda) to tweak the renamings; please see the commit message and make sure that I have correctly understood your intent.

Also, see comment about Closure Compiler diagnostics below.

scripts/gulpfiles/build_tasks.js
maribethb: > More broadly, why do we have non-deprecated code calling @deprecated code?

We don't, in TypeScript. Beka made this a warning because it was previously an error. The line that Closure complains about is actually the line that declares the enum. The generated JS looks different from the TS, and causes this error. 

Adding a `@suppress` directly into the TsDoc already present on the enum did not cause closure to suppress the error either, but I don't know why. The only way we could get it to suppress the error was to add a new jsdoc comment containing only the `@suppress`, but that breaks the tsdoc generated reference docs as the real comment is no longer associated with the enum declaration. Would be interested in your thoughts on this problem, but in the meantime I suggested that calling deprecated code shouldn't be build-breaking IMO, but a warning seems appropriate.

BeksOmega:
Intention of the renamings was correct! Thank you for the fixes Christopher :D

maribethb (approved):

Issue #7086, opened 2023-05-12
Proposal: Library generator publication changes + supporting change to `CodeGenerator`


cpcallen:
This is the third of three issues created to discuss proposed changes to generators (and perhaps track implementation of those proposals, if agreed).  This issue concerns proposed changes to the way the generators in `generators/` are published (to form the basis of a standard for the publication of generators as plugins), plus a small breaking change to the `CodeGenerator` class in `core/generator.ts` to support this.  This proposal assumes that the principle parts of the proposals in #7084 and #7085 have been adopted, but is only partially contingent on that.

In code examples in these bugs, the word "language" / "Language" / "LANGUAGE" should be read as a metasyntactic variable standing in for some particular language like `JavaScript`. 

## Background

At the moment, the way per-block generator functions are created is via side effects (assigning to a property on a `CodeGenerator` instance), and ties them directly to a particular `CodeGenerator` instance (by calling other methods on that specific instance):

```JS
languageGenerator['block_type'] = function(block) {
  // Get code for some argument input:
  const argumentCode = languageGenerator.valueToCode(block, 'INPUT_NAME', languageGenerator.ORDER_FOO);
  const code = /* some code + */ argumentCode /* + more code */;
  const outputOrder = languageGenerator.ORDER_BAR;
  return [code, outputOrder];
};
```

This means that it is difficult to load just specific generator functions, and it is not generally safe to copy generator functions from one `CodeGenerator` instance to another.

### Option 1A: Pass  as parameter

The `CodeGenerator` instance could be passed as a parameter.  For backward compatibility it should be the second parameter:

```JS
languageGenerator['block_type'] = function(block: Block, generator: LanguageGenerator) {
  // Get code for some argument input:
  const argumentCode = generator.valueToCode(block, 'INPUT_NAME', generator.ORDER_FOO);
  const code = /* some code + */ argumentCode /* + more code */;
  const outputOrder = generator.ORDER_BAR;
  return [code, outputOrder];
};
```

### Option 1B: Pass as `this`

Alternatively the generator instance it could (via `Function.prototype.call`) be passed as `this`:

```JS
languageGenerator['block_type'] = function(this: LanguageGenerator, block: Block) {
  // Get code for some argument input:
  const argumentCode = this.valueToCode(block, 'INPUT_NAME', this.ORDER_FOO);
  const code = /* some code + */ argumentCode /* + more code */;
  const outputOrder = this.ORDER_BAR;
  return [code, outputOrder];
};
```

[...]

PR #7168, opened 2023-06-14
refactor(generators)!: Pass `this` `CodeGenerator` to individual generator functions

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Implements proposal 1 (option 1A) of #7086.

### Proposed Changes

* Have `CodeGenerator.prototype.blockToCode` pass `this` as the second argument to the per-block-type generator function when calling it.
* Rewrite all generator functions in `generators/*/*.js` to accept a `generator` argument as their second parameter, and use `generator` (instead of the closed-over `langGenerator` instance) when making recursive calls to (or accessing properties on) the generator.
* Remove a spurious duplicate generator function from `generators/javascript/procedures.js`.

#### Behaviour Before Change

Generator functions needed to be closures in order to access the generator instance they were being used by.  This meant they were tied to a single `CodeGenerator` instance.

#### Behaviour After Change

Generator functions can now be used by any `CodeGenerator` instance (provided it is of the correct subclass: obviously it does not make sense to use a JavaScript generator function with a `PythonGenerator` instance).

### Reason for Changes

See #7086.

### Test Coverage

Passes `npm test`.  No changes to manual test procedures expected.

### Documentation

We should update our generator documentation and codelabs to take advantage of the new features made available by this PR—but the changes are mostly non-breaking for existing code, so this probably does not need to happen immediately.

### Additional Information

**BREAKING CHANGE:**  This PR makes the generator functions we provide dependent on on the new generator parameter.  Although `CodeGenerator.prototype.blockToCode` has been modified to supply this automatically, so this change should not affect most developers, this change will be a breaking change where developers make direct calls to the provided generator functions without supplying the generator parameter.

Specifically, if a generator function contains a call of the form:
```JS
langGenerator.forBlock['block_type_1'] = function(block) {
  // ...
  const code = langGenrator.forBlock['block_type_2'](block);
  // ...
}
```
it will **must** be updated to accept and pass a second parameter as follows:
```JS
langGenerator.forBlock['block_type_1'] = function(block, generator) {
  // ...
  const code = genrator.forBlock['block_type_2'](block, generator);
  // ...
}
```

We **recommend** that other generator functions be updated similarly, but expect that existing code of the form
```JS
langGenerator.forBlock['block_type'] = function(block) {
  // ...
  const code = langGenrator.valueToCode(/*...*/);
  // ...
}
```
will continue to work without modification for the foreseeable future.

### DIFF
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -29,7 +29,10 @@ import {warn} from './utils/deprecation.js';
  *     or a [code, precedence] tuple (for value/expression blocks), or
  *     null if no code should be emitted for block.
  */
-export type BlockGenerator = (block: Block, generator: CodeGenerator) => [string, number] | string | null;
+export type BlockGenerator = (
+  block: Block,
+  generator: CodeGenerator
+) => [string, number] | string | null;
 
 /**
  * Class for a code generator that translates the blocks into a language.
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -24,11 +24,12 @@ import {warn} from './utils/deprecation.js';
  *
  * @see {@link https://developers.google.com/blockly/guides/create-custom-blocks/generating-code}
  * @param block The Block instance to generate code for.
+ * @param genearator The CodeGenerator calling the function.
  * @returns A string containing the generated code (for statement blocks),
  *     or a [code, precedence] tuple (for value/expression blocks), or
  *     null if no code should be emitted for block.
  */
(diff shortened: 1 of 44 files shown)


cpcallen:
Advice to reviewer: this PR contains several small, manual commits interspersed with large, automatically-generated ones.  You will almost certainly find it easier to review commit-by-commit.

BeksOmega (approved):

PR #7170, opened 2023-06-15
fix: prefer mp3 file for disconnect sound

simap:
The disconnect.mp3 entry should be first in the list, like the others. Otherwise it loads the .wav file when an mp3 would work.

## The basics

- [x] I branched from develop

### Resolves

The disconnect.wav file loads instead of the mp3.

### Proposed Changes

fix the order in the list of alternatives

### DIFF
--- a/core/inject.ts
+++ b/core/inject.ts
@@ -365,8 +365,8 @@ function loadSounds(pathToMedia: string, workspace: WorkspaceSvg) {
   );
   audioMgr.load(
     [
-      pathToMedia + 'disconnect.wav',
       pathToMedia + 'disconnect.mp3',
+      pathToMedia + 'disconnect.wav',
       pathToMedia + 'disconnect.ogg',
     ],
     'disconnect'


simap:
@maribethb, ooh, interesting history! Thank you for looking in to that. 

I have no interest in signing any contracts for a change this trivial, feel free delete this PR and re-fix or ignore. I was trying to trim down the external files a bit to embed this into a microcontroller, and noticed the .wav was used instead of the .mp3. I don't think anyone else will care about a handful of bytes.

The work y'all do with Blockly is awesome!

maribethb:
Fair enough! If you decide to contribute to any Google open source projects in the future, you'll have to sign the CLA then, but after you sign once you won't need to again as it'll be associated with the email on your git commits. We'll be here if you change your mind :)

NeilFraser:
For 'click' and 'delete' the mp3 files are smaller than the other formats.  However for 'disconnect' the wav file is smaller than the mp3.  So we simply prefer the smallest file.

PR #7173, opened 2023-06-15
refactor(generators): Restructure generator modules to contain side effects

### LINKED ISSUE #7086: Proposal: Library generator publication changes + supporting change to `CodeGenerator`

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7086

### Proposed Changes

Implements proposal 2 (option 2A) of #7086.  In particular:

1. `generators/<lang>.js` are moved to `generators/<lang>/<lang>_generator.js`.  Code creating `<lang>Generator` instances is removed from these file.
0. `generators/<lang>/*.js` (excluding `<lang>_generator.js`) are modified to declare and _individually_ export per-block generator functions, not referencing any particular `CodeGenerator` instance.  Calls to `<lang>Generator.addReservedWords()` re replaced with comments noting the reserved words needed by the generator functions in that file.
0. `generators/<lang>/all.js` are moved to `generators/<lang>.js` and modified to:

   - Reexport everything from `generators/<lang>/<lang>_generator.js`.
   - Create and export a `<Lang>Generator` instance named `<lang>Generator`.
   - Call `<langGenerator.addReservedWords()` to add reserved words mentioned in the block generator files (if any).
   - Install all of the block generator functions from `generators/<lang>/*.js` on `<lang>Generator.forBlock`.

   These file remain the entrypoint for the `<lang>_compressed.js` chunks.

Additionally, this PR:

- Updates the `chunks` declaration in `build_tasks.js` as accordingly.
- Updates `scripts/migration/renamings.json5` to record these updates plus update references from `Blockly.<Lang>` to `<lang>.<lang>Generator` (only applicable, and only effective, on code that imports generators as scripts.
- Fix two missing `export` statements for some `class <Lang>Generator`s.
- Fix some tests which imported generators improperly.

#### Behaviour Before Change

Generators work.

#### Behaviour After Change

Unchanged.

### Reason for Changes

See #7086.

### Test Coverage

Passes `npm test`.  No changes to manual test procedures expected.

### Documentation

No documentation updates needed.  This PR is an internal-only refactor.

### Additional Information

This change paves the way for better tree-shaking of generator functions, though the individual generator functions are not yet exposed in in the compiled chunks.  Projects that compile Blockly from source (such as Blockly Games) should be able to take advantage of this change immediately.

### Notes for Reviewer

Because of the file moves, it will probably be easier to review this PR commit-by-commit.


generators/dart.js
```diff
@@ -13,14 +13,32 @@
...
+
+/**
+ * Dart code generator instance.
+ * @type {!DartGenerator}
+ */
+export const dartGenerator = new DartGenerator();
+
+// Add reserved words.  This list should include all words mentioned
+// in RESERVED WORDS: comments in the imports above.
+dartGenerator.addReservedWords('Html,Math');
```
BeksOmega, line 37: Could you have the modules export arrays of reserved words? I imagine this would break the `Object.assign` below, but wanted to mention in case there was a way to work around that.

generators/dart/colour.js
```diff
@@ -11,18 +11,18 @@
...
 
-import {dartGenerator, Order} from './dart_generator.js';
+import {Order} from './dart_generator.js';
 
 
-dartGenerator.addReservedWords('Math');
+// RESERVED WORDS: 'Math'
 
-dartGenerator.forBlock['colour_picker'] = function(block, generator) {
+export function colour_picker(block, generator) {
```
BeksOmega, line 19: Because we're just doing `SIMPLE_OPTIMIZATIONS` these won't be renamed by closure correct?

BeksOmega (approved):

generators/dart.js
cpcallen: I actually thought about that.  It's doable: a particular export can be pulled out using a destructuring assignment, leaving the rest on a remainder object which could then be passed to `assign`.

I also looked into whether ES modules can have symbol-keyed exports.  Regular exports just become ordineary string-keyed properties on the `Module` object, so why not provide a way to have properties on the `Module` object that have a `Symbol` as their key?  (They in fact already have a property keyed by [Symbol.toStringTag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) so that they get reported as `[object Module]` by `Object.prototype.toString()`.)

Alas, it turns out that that there's no way to do that, and, due to the absence of symbol literals (symbols only really exist at runtime) combined with the way that exports are statically analysed, not really any good way for a future ES202x to make it possible.

I had another thought, though: what about putting an extra property on the function objects themselves?  That way each individual one carries with it the information needed to use it.  This could be scanned by `generator/<lang>.js` at installation time, or even in the `<Lang>Generator`'s `init` method.

Opinions welcome, though I'm inclined to leave implementing any of these options (well, except for the infeasible one) for a later PR, though, unless you think the current code unacceptable.

BeksOmega: Agree! best to leave for a separate PR. I'll respond to the rest of your questions laterz, just wanted you to know this isn't blocking =)

generators/dart/colour.js
cpcallen: Correct.

If we ever get `ADVANCED_OPTIMIZATIONS` working we'd need to rethink this.  One possibility would be to use the functions' built-in `.name` properties when installing them—though to avoid that _also_ being corrupted we might have to resort to writing something like:

```JS
export block_type(/*...*/) {/*...*/};
block_type.name = 'block_type';
```

Option 2B would avoid this potential future problem.

generators/dart.js
```diff
@@ -1,317 +1,44 @@
 /**
  * @license
- * Copyright 2014 Google LLC
+ * Copyright 2021 Google LLC
```
maribethb, line 3: minor: Why 2021?

generators/dart.js
maribethb: +1 that finding a way to export the reserved words would be useful in a follow-up. Putting an extra property on the function object makes me feel weird, but it would be nice if people who are importing individual functions couldn't forget this part.

generators/dart/colour.js
maribethb: If we ever got advanced compilation working, we'd definitely have to have some way of giving closure a list of things not to rename. This is not the first place we'd have an exported function that shouldn't be renamed. But I do not foresee that being in our roadmap anytime soon if ever.

generators/dart/dart_generator.js
```diff
@@ -0,0 +1,311 @@
...
+ * @fileoverview Helper functions for generating Dart for blocks.
+ * @suppress {checkTypes|globalThis}
+ */
+
+import * as goog from '../../closure/goog/goog.js';
+goog.declareModuleId('Blockly.Dart');
+
+import * as Variables from '../../core/variables.js';
+import * as stringUtils from '../../core/utils/string.js';
+// import type {Block} from '../../core/block.js';
```
maribethb, line 17: Can we clean up these commented out exports?

scripts/migration/renamings.json5
```diff
@@ -1510,5 +1512,61 @@
...
+        },
+      },
+    },
+    {
+      oldName: 'Blockly.Python',
+      exports: {
+        pythonGenerator: {
+          newModule: 'Blockly.Python.all',
+          oldPath: 'Blockly.Python',
+          newPath: 'python.pythonjavascriptGenerator',
```
maribethb, line 1567: I believe/hope this one is wrong :P

tests/mocha/generator_test.js
```diff
@@ -7,11 +7,11 @@
 goog.declareModuleId('Blockly.test.generator');
 
 import * as Blockly from '../../build/src/core/blockly.js';
-const {dartGenerator} = goog.require('Blockly.Dart');
-const {javascriptGenerator} = goog.require('Blockly.JavaScript');
-const {luaGenerator} = goog.require('Blockly.Lua');
-const {phpGenerator} = goog.require('Blockly.PHP');
-const {pythonGenerator} = goog.require('Blockly.Python');
+const {DartGenerator} = goog.require('Blockly.Dart.all');
```
maribethb, line 10: Why is this one different than the others? iiuc the `all` shouldn't be here since that one doesn't have the constructor.

generators/dart.js
cpcallen: > it would be nice if people who are importing individual functions couldn't forget this part.

🤔 We could include the reserved words in the name of the export:
```
export function math_arithmetic$RESERVED$Math(block, generator) {/* ... */}
```
😈

Filed #7187 for further discussion.

generators/dart.js
cpcallen: This was because `generators/dart.js` was created in 2014, but `generators/dart/all.js` was created in 2021, and the latter has been renamed to the former's name (while the former is now `generators/dart/dart_generator.js`).

generators/dart/dart_generator.js
cpcallen: The lines (in uncommented form) were generated from `goog.requireType` calls by the `js2ts` migration script.  Because `import type` is not a thing in plain JS, they can't be left in, but they will be needed when these files are migrated to TypeScript, so I thought it most sensible to just comment them out for now.

scripts/migration/renamings.json5
cpcallen: Yes, that is definitely wrong.  Fixed.

tests/mocha/generator_test.js
cpcallen: Good catch.  You are quite correct.  Fixed.

Issue #6920, opened 2023-03-24
Improve implementation and type safety of library blocks


cpcallen:
## Description

In the process of converting the library blocks (`blocks/*`) from JavaScript to TypeScript, the strategy for dealing with type errors unearthed by `tsc` (due to more specific types in the code and/or more assiduous type checking by `tsc` as compared to Closure Compiler) has been to try to suppress the type errors with casts (`as`, `!`, etc.) rather than make other code changes, in an effort to ensure that the `.js` code generated by `tsc` is as close as possible to the original—to ensure bug-for-bug compatibility.

Once the TypeScript migration is complete, we should revisit the library blocks implementation and remove such casts, replacing them with:

- more correct types, where the existing types are not quite correct, and/or
- type assertions and other runtime checks (rather than casts), and/or
- bugfixes, where the code is actually wrong.

### General issues throughout:

* Use of `!` (typically but not exclusively in the form (`!.`) to suppress could-be-null errors.
* Use of `as` casts (which in some cases may require making the return types of methods in core more specific)
* Certain list blocks unjustifiably assume they are instances of `BlockSvg` (tracked in #6905).

### In `blocks/lists.ts`

* `lists_getIndex`:
  * `updateStatement_` method calls `Block.prototype.unplug` with two arguments instead of one.

cpcallen:
# Improving block definition typing

The issue lists a large number of seemingly-unrelated issues, but many of them stem from the same root cause: a mismatch between the very simplistic mixin-object-based approach that Blockly uses for block definisions and mutator extensions, vs. [th more sophisticated but opinionated approach to mixins taken by TypeScript](https://www.typescriptlang.org/docs/handbook/mixins.html).

There are a number of ways to resolve this; probably the most intuitive would be to make block definitions all be subclasses of `Block`, but this would necessitate replacing `BlockSVG` with a view or block-renderer object ("favour composition over inheritance").  A less invasive approach would be to make block definitions and extensions into TypeScript-flavoured mixins, each with their own factory function.  Either approach would entail making changes to core to support a new, third mechanism for defining blocks.

Instead, I propose a possible approach to making block definitions and extension mixins more TypeScript-friendly and type-safe in a way does not require modifications core, but which has most of the advantages of the subclass and TS-mixin approaches.

Goals:

* To define blocks and extensions in ways that are as much like simply subclassing of `Block` as possible.
* To avoid having to do as much explicit (and potentially unsafe) casting.
[...]

PR #7192, opened 2023-06-21
refactor(blocks): Migrate `blocks/procedures.js` to TypeScript

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6828.

### DIFF
--- a/blocks/procedures.ts
+++ b/blocks/procedures.ts
@@ -4,11 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/**
- * @fileoverview Procedure blocks for Blockly.
- * @suppress {checkTypes|visibility}
- */
-
 import * as goog from '../closure/goog/goog.js';
 goog.declareModuleId('Blockly.libraryBlocks.procedures');
 
@@ -599,7 +594,7 @@ type FieldTextInputForArgument = FieldTextInput & {
 };
 
 const PROCEDURES_MUTATORARGUMENT = {
-    /**
+  /**
    * Mutator block for procedure argument.
    */
   init: function(this: ArgumentBlock) {
@@ -1229,7 +1224,7 @@ const PROCEDURES_IFRETURN = {
     }
     let legal = false;
     // Is the block nested in a procedure?
-    let block = this;
+    let block = this;  // eslint-disable-line @typescript-eslint/no-this-alias
     do {
       if (this.FUNCTION_TYPES.indexOf(block.type) !== -1) {
         legal = true;
--- a/blocks/procedures.ts
+++ b/blocks/procedures.ts
@@ -12,8 +12,6 @@
 import * as goog from '../closure/goog/goog.js';
 goog.declareModuleId('Blockly.libraryBlocks.procedures');
 
-/* eslint-disable-next-line no-unused-vars */
-import type * as AbstractEvent from '../core/events/events_abstract.js';
 import * as ContextMenu from '../core/contextmenu.js';
 import * as Events from '../core/events/events.js';
 import * as Procedures from '../core/procedures.js';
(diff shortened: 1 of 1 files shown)


blocks/procedures.ts
```diff
@@ -447,21 +465,21 @@ blocks['procedures_defnoreturn'] = {
   ...PROCEDURE_DEF_COMMON,
   /**
    * Block for defining a procedure with no return value.
-   * @this {Block}
    */
-  init: function() {
+  init: function(this: ProcedureBlock) {
```
BeksOmega: Can you make this one `ProcedureBlock&BlockSvg` to remove the cast below?

On the other hand, maybe that's more innaccurate and the problem is actually with types elsewhere (`setMutator`?).

blocks/procedures.ts
```diff
@@ -786,9 +845,10 @@ const PROCEDURE_CALL_COMMON = {
     // Reconnect any child blocks.
     if (this.quarkIds_) {
       for (let i = 0; i < this.arguments_.length; i++) {
-        const quarkId = this.quarkIds_[i];
+        const quarkId: string = this.quarkIds_[i];  // TODO(#6920)
         if (quarkId in this.quarkConnections_) {
-          const connection = this.quarkConnections_[quarkId];
+          // TODO(#6920): investigate claimed circular initialisers.
```
BeksOmega, line 899: What does "circular intialisers" refer to?

blocks/procedures.ts
```diff
@@ -76,10 +94,10 @@ const PROCEDURE_DEF_COMMON = {
   },
   /**
    * Update the display of parameters for this procedure definition block.
-   * @private
-   * @this {Block}
+   *
+   * @internal
```
BeksOmega, line 93: If the intention is to migrate `@private` methods to `@internal` methods please do this throughout the rest of the file!

blocks/procedures.ts
```diff
@@ -528,24 +546,29 @@ blocks['procedures_defreturn'] = {
...
+interface ContainerMixin extends ContainerMixinType {}
+type ContainerMixinType = typeof PROCEDURES_MUTATORCONTAINER;
+
+const PROCEDURES_MUTATORCONTAINER = {
   /**
    * Mutator block for procedure container.
-   * @this {Block}
    */
-  init: function() {
+  init: function(this: Block) {
```
BeksOmega: `this: ContainerBlock` ?

blocks/procedures.ts
```diff
@@ -555,36 +578,47 @@ blocks['procedures_mutatorcontainer'] = {
...
-    });
+    }) as FieldTextInputForArgument;
     field.setValidator(this.validator_);
     // Hack: override showEditor to do just a little bit more work.
     // We don't have a good place to hook into the start of a text edit.
-    field.oldShowEditorFn_ = field.showEditor_;
+    field.oldShowEditorFn_ = (field as AnyDuringMigration).showEditor_;
     /**
      * @this {FieldTextInput}
      */
```
BeksOmega, line 651: Delete unnecessary comment.

blocks/procedures.ts
```diff
@@ -555,36 +578,47 @@ blocks['procedures_mutatorcontainer'] = {
...
     const field = fieldRegistry.fromJson({
       type: 'field_input',
       text: Procedures.DEFAULT_ARG,
-    });
+    }) as FieldTextInputForArgument;
     field.setValidator(this.validator_);
     // Hack: override showEditor to do just a little bit more work.
     // We don't have a good place to hook into the start of a text edit.
-    field.oldShowEditorFn_ = field.showEditor_;
+    field.oldShowEditorFn_ = (field as AnyDuringMigration).showEditor_;
```
BeksOmega, line 648: Maybe add a TODO comment pointing at #6920  for this?

blocks/procedures.ts
cpcallen: From #6920:
> `setProcedureParameters_` unexpectedly required declarations of the types of the `quarkId` and `connection` local variables avoid "implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer" errors.  Since the definitions do not appear to be circular (though `connection` is initialised based on `quarkId`) this smells and probably warrants further investigation.

blocks/procedures.ts
cpcallen: Done.

blocks/procedures.ts
cpcallen: I could, but #6920 already mentions use of `as`, and I had been treating the "DuringMigration" part as making it implicitly clear that it was not OK.

I've  added a note to #6920 to check for `AnyDuringMigration` in particular, though.

blocks/procedures.ts
cpcallen: Done.  I've also gone back to check the others (there are a few missing in `text.ts`), which I will file a separate PR about.

blocks/procedures.ts
cpcallen: Done.  Obviously not necessary, but does help make things clearer.  (I think I had to go back and add `ContainerBlock` for the benefit of later block types, and didn't update this one.)

blocks/procedures.ts
cpcallen: That would be more consistent with how I had handled earlier cases.  (Ahh, the challenges of trying to finish a task one started more than a month earlier…)

Done—here and for `blocks['procedures_defreturn']` below.

BeksOmega (approved):

PR #7194, opened 2023-06-21
fix(blocks): Mark previously-`@private` methods `@internal`

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Where certain block mixin methods had been marked `@private`, and these annotations had been deleted while migrating the containing files to TypeScript, mark those methods as `@internal`.

#### Behaviour Before Change

These methods might have appeared in our `.d.ts` files and external documentation.

#### Behaviour After Change

These methods should not appear in our `d.ts` files or external documentation.

### Reason for Changes

The migration PRs were not intended to change any API.

### Test Coverage

Passes `npm test`.  No changes to manual testing required.

### Documentation

…may be sparser than before, intentionally.

### Additional Information

I'm not actually sure if these methods appear in our external documentation at the moment—I'd guess not.   But better to keep it that way.

### DIFF
--- a/blocks/logic.ts
+++ b/blocks/logic.ts
@@ -505,6 +505,8 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
   },
   /**
    * Modify this block to have the correct number of inputs.
+   *
+   * @internal
    */
   updateShape_: function (this: IfBlock) {
     // Delete everything.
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
One meta comment: Make sure to put `@private` and `@internal` in back ticks, otherwise we `@` people we don't want to :P

cpcallen:
> One meta comment: Make sure to put `@private` and `@internal` in back ticks, otherwise we `@` people we don't want to

Oops.  But also: whoever started using `@username` instead of `username@` gets a big ಠ_ಠ from me.

PR #7211, opened 2023-06-26
fix(demos): Update BlockFactory generator stub generator

cpcallen:
## The basics

- [X] I branched from develop

### Proposed Changes

Update the BlockFactory block generator function stub generator to apply recent changes in generator function best practices:

- Use `<language>Genarator` instead of `Blockly.<Language>`.
- Put generator functions in the `.forBlock` dictionary.
- Accept (and use) a second `generator` argument that is the calling CodeGenerator object.
- User `Order.ATOMIC` enum instead of `.ORDER_ATOMIC`.

Also:

- Prefix (e.g.) `javascriptGenerator` and `Order` with `javascript.`.
- Use template literals where useful.
- DRY up all the non-special field stub code generation.

### Reason for Changes

Ensure developer generator functions are as modern as possible.

### Additional Information

I added `language.` prefix to places where the stub code references `langaugeGenerator` and `Order`, as is necessary when loading the generators as a script (e.g. via a `<script>` tag), but we expect that many users will be doing `import {languageGenerator, Order} from 'blockly/language'`.  Should we provide a "use named imports" tickbox or something like that?

### DIFF
--- a/demos/blockfactory/factory_utils.js
+++ b/demos/blockfactory/factory_utils.js
@@ -80,10 +80,10 @@ FactoryUtils.getGeneratorStub = function(block, generatorLanguage) {
     return '  var ' + root + '_' + name;
   }
   // The makevar function lives in the original update generator.
-  var language = generatorLanguage;
+  var language = generatorLanguage.toLowerCase();
   var code = [];
-  code.push("Blockly." + language + "['" + block.type +
-            "'] = function(block) {");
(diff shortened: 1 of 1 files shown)


maribethb:
> Should we provide a "use named imports" tickbox or something like that?

This is one of the challenges I'm thinking about for the rewrite of developer tools. We don't need to do anything about it today for the version in core.

maribethb (approved):

PR #7272, opened 2023-07-10
chore: create test for undo block movements and editing a field

ericblackmonGoogle:
## The basics

- [x] I branched from develop

### Proposed Changes

Adding testing functionality for undoing block movement and editing fields

### DIFF
--- a/tests/browser/test/block_undo_test.js
+++ b/tests/browser/test/block_undo_test.js
@@ -1,0 +1,60 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Node.js script to run Automated tests in Chrome, via webdriver.
+ */
+
+const chai = require('chai');
+const {Key} = require('webdriverio');
+const {
+  testSetup,
+  testFileLocations,
+  getSelectedBlockElement,
+  getNthBlockOfCategory,
+  getBlockTypeFromCategory,
+  connect,
+} = require('./test_setup');
+
+let browser;
+suite('Testing Connecting Blocks', function (done) {
+  // Setting timeout to unlimited as the webdriver takes a longer time to run than most mocha test
+  this.timeout(0);
+
+  // Setup Selenium for all of the tests
+  suiteSetup(async function () {
+    browser = await testSetup(testFileLocations.playground);
+  });
+  test('Testing Field Edits ', async function () {
+    // Drag out first function
+    let proceduresDefReturn = await getBlockTypeFromCategory(
+      browser,
+      'Functions',
+      'procedures_defreturn'
+    );
+
+    //undo the block drag out
+    await proceduresDefReturn.dragAndDrop({x: 50, y: 20});
+    await browser.keys([Key.Ctrl, 'z']);
+
+    const blockOnWorkspace = await browser.execute(() => {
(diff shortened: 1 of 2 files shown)


tests/browser/test/field_edits_test.js
```diff
@@ -0,0 +1,74 @@
...
+  await workspace.click();
+  await new Promise((resolve) => setTimeout(resolve, 2000)); // 2 sec
+  // Get value of the number
+  const numericText = await browser
+    .$(
+      '#blocklyDiv > div > svg.blocklySvg > g > g.blocklyBlockCanvas > g.blocklyDraggable > g > text'
+    )
+    .getHTML();
+
+  chai.assert.isTrue(numericText.includes('1223'));
```
BeksOmega, line 73: Assuming that the second `2` is always going to be entered in the middle seems kind of fragile. Could we trigger the `delete` key first to remove the existing text, and then type in the new number? That feels like it might be more stable!

tests/browser/test/field_edits_test.js
```diff
@@ -0,0 +1,74 @@
...
+  // Cick on the workspace
+  const workspace = await browser.$('#blocklyDiv > div > svg.blocklySvg > g');
+  await workspace.click();
+  await new Promise((resolve) => setTimeout(resolve, 2000)); // 2 sec
+  // Get value of the number
+  const numericText = await browser
+    .$(
+      '#blocklyDiv > div > svg.blocklySvg > g > g.blocklyBlockCanvas > g.blocklyDraggable > g > text'
+    )
+    .getHTML();
```
BeksOmega, line 71: This is more of a high-level question, but in general I'm not sure what we care about more:
  * Whether the state of the UI is correct
  * Whether the state of Blockly is correct

Here we're checking that the UI (i.e. text) is correct. But we could also check that the state of Blockly (i.e. the output of `getFieldValue` is correct). I don't have an opinion about this! so  I think what you've got is fine :P Just wanted to call it out!

BeksOmega (commented):
Generally LGTM! Just a few nits + questions =)

tests/browser/test/block_undo_test.js
```diff
@@ -0,0 +1,71 @@
...
+    await testUndoBlock(-1);
+  });
+
+  // Teardown entire suite after test are done running
+  suiteTeardown(async function () {
+    await browser.deleteSession();
+  });
+});
+
+async function testUndoBlock(delta) {
```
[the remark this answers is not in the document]
ericblackmonGoogle: I put an enum in test setup that lets ltr represent 1 and rtl represent -1 so that its a little more readable

tests/browser/test/field_edits_test.js
ericblackmonGoogle: That makes sense! I have more familiar with interacting with the browser so I tend to check the UI more than Blockly, but can start using Blockly's methods more as well!

tests/browser/test/test_setup.js
```diff
@@ -73,6 +73,11 @@ const testFileLocations = {
   playground: 2,
 };
 
+const screenDirection = {
+  rtl: -1,
+  ltr: 1,
+};
```
cpcallen, line 79: This could have a JSDoc marking it as an `@enum` (as well as describing it, if useful).

Please use `CONSTANT_CASE` for the enum values (`LTR`, `RTL`).

PR #7279, opened 2023-07-11
chore: add tests for disabling blocks

### LINKED ISSUE #7277: Disable Function Testing

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7277 

### Proposed Changes

Adds tests for disabling blocks via the right click context menu.

### Reason for Changes

Testing is great!

### Test Coverage

This is tests :P

### Additional Information

N/A

### DIFF
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -131,7 +131,7 @@ suite('Disabling', function () {
   this.timeout(0);
 
   suiteSetup(async function () {
-    browser = await testSetup(testFileLocations.playground);
+    browser = await testSetup(testFileLocations.PLAYGROUND);
   });
 
   setup(async function () {
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -90,39 +90,33 @@ suite('Right Clicking on Blocks', function () {
 
   test('clicking the collapse option collapses the block', async function () {
     await contextMenuSelect(browser, this.block, 'Collapse Block');
-    const isCollapsed = await getIsCollapsed(browser, this.blockId);
-    chai.assert.isTrue(isCollapsed);
+    chai.assert.isTrue(await getIsCollapsed(browser, this.blockId));
   });
 
   // Assumes that
   test('clicking the expand option expands the block', async function () {
     await contextMenuSelect(browser, this.block, 'Expand Block');
-    const isCollapsed = await getIsCollapsed(browser, this.blockId);
-    chai.assert.isFalse(isCollapsed);
+    chai.assert.isFalse(await getIsCollapsed(browser, this.blockId));
   });
 
   test('clicking the disable option disables the block', async function () {
     await contextMenuSelect(browser, this.block, 'Disable Block');
-    const isDisabled = await getIsDisabled(browser, this.blockId);
-    chai.assert.isTrue(isDisabled);
+    chai.assert.isTrue(await getIsDisabled(browser, this.blockId));
   });
 
   test('clicking the enable option enables the block', async function () {
     await contextMenuSelect(browser, this.block, 'Enable Block');
-    const isDisabled = await getIsDisabled(browser, this.block.id);
-    chai.assert.isFalse(isDisabled);
+    chai.assert.isFalse(await getIsDisabled(browser, this.block.id));
   });
 
   test('clicking the add comment option adds a comment to the block', async function () {
     await contextMenuSelect(browser, this.block, 'Add Comment');
-    const commentText = await getCommentText(browser, this.block.id);
-    chai.assert.equal(commentText, '');
+    chai.assert.equal(await getCommentText(browser, this.block.id), '');
   });
 
   test('clicking the remove comment option removes a comment from the block', async function () {
     await contextMenuSelect(browser, this.block, 'Remove Comment');
-    const commentText = await getCommentText(browser, this.block.id);
-    chai.assert.isNull(commentText);
+    chai.assert.isNull(await getCommentText(browser, this.block.id));
   });
 
   // Teardown entire suite after test are done running
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -13,6 +13,8 @@ const {
   testSetup,
   testFileLocations,
   dragNthBlockFromFlyout,
+  dragBlockTypeFromFlyout,
+  connect,
   contextMenuSelect,
 } = require('./test_setup');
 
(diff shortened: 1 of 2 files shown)


tests/browser/test/basic_playground_test.js
```diff
@@ -76,16 +79,8 @@ suite('Right Clicking on Blocks', function (done) {
   test('Collapse', async function () {
     await browser.refresh();
     const block = await dragNthBlockFromFlyout(browser, 'Loops', 0, 20, 20);
-
-    const blockId = block.id;
-    let isCollapsed = await getIsCollapsed(browser, blockId);
-    chai.assert.isFalse(isCollapsed);
```
cpcallen, line 82: Why not check it's not collapsed before collappsing?

(And same below for disabling.)

tests/browser/test/basic_playground_test.js
```diff
@@ -95,42 +90,126 @@ suite('Right Clicking on Blocks', function (done) {
     await contextMenuSelect(browser, block, 'Collapse Block');
     await contextMenuSelect(browser, block, 'Expand Block');
```
cpcallen, line 91: Would it be worth verifying it is actually collapsed before expanding?  Or is that just duplicating the previous test?  Could you collapse the block via API for this test?  Otherwise maybe combine the tests since they seem to be unavoidably entwined.

And same below for disable/enable.

tests/browser/test/test_setup.js
```diff
@@ -232,11 +232,19 @@ async function dragBlockTypeFromFlyout(browser, categoryName, type, x, y) {
...
 async function contextMenuSelect(browser, block, itemText) {
-  await block.click({button: 2});
-  await browser.pause(200);
+  // Clicking will always happen in the middle of the block's bounds
+  // (including children) by default, which causes problems if it has holes
+  // (e.g. statement inputs).
+  const xOffset = Math.round((await block.getSize('width')) * 0.3);
+  const yOffset = Math.round((await block.getSize('height')) * 0.45);
+
+  await block.click({button: 2, x: -xOffset, y: -yOffset});
```
cpcallen, line 241: I'm a bit perplexed about the values you've chosen here.  If I am reading correctly, you're clicking 30% in from the right and 45% down from the top, it looks like it's optimised to be in the middle of the opening of C-shaped block, at least if it has any blocks inside it.  Wouldn't, say, 5% in from the left and somewhere >50% down from the top be a better bet?  Or small fixed offsets from the upper-left corner (chosen carefully to avoid any icons)?

tests/browser/test/basic_playground_test.js
BeksOmega: I think collapsing via API makes sense! although honestly I'm not really sure how to write a good integration test :P

tests/browser/test/test_setup.js
BeksOmega: The default click position is in the middle, so we're actually clicking 5% from the top and 20% from the left (this is why the offsets are negative). For some reason clicking closer to the left was being flakey, and sometimes clicked the workspace instead. Through experimentation, this seemed to work consistently!

I can expand on the comment to say where the click is occuring.

tests/browser/test/basic_playground_test.js
rachel-fenichel: I'm joining this conversation because Beka referred to it in comments on [my recent PR](https://github.com/google/blockly/pull/7284).  Personally, I don't see a lot of value in treating these as unit tests with a single assertion per test. The manual tests that they are modeled on do a sequence of actions, checking for failure at each step, in order to test that a workflow is working. While we *can* write these tests so that they are completely separated from each other, it seems to me this creates a lot of extra work and duplication of setup.

Personally, I would rather be able to describe the tests as a set of actions by a user, and have debugging involve stepping through those actions manually to find which one failed--even if that means that an individual test does not point you directly to the specific line that failed. For my right-click tests this could also take the form of doing all of the actions (collapse, expand, disable, enable, add comment, remove comment) one after another in a single test instead of doing each as a separate test with separate creation of the block under test.

rachel-fenichel: I'm not sure where this fits into formal models of testing, although in my head this is end-to-end testing.

tests/browser/test/basic_playground_test.js
BeksOmega: Reverted after talking with Rachel! She changed these to all chain.

BeksOmega:
@cpcallen ready for re-review =)

tests/browser/test/basic_playground_test.js
```diff
@@ -100,14 +103,14 @@ suite('Right Clicking on Blocks', function (done) {
 
   test('clicking the disable option disables the block', async function () {
     await contextMenuSelect(browser, this.block, 'Disable Block');
-    const isEnabled = await getIsEnabled(browser, this.blockId);
-    chai.assert.isFalse(isEnabled);
+    const isDisabled = await getIsDisabled(browser, this.blockId);
+    chai.assert.isTrue(isDisabled);
```
cpcallen, line 107: Any reason not to combine these into a single statement?
```suggestion
    chai.assert.isTrue(await getIsDisabled(browser, this.blockId));
```

tests/browser/test/basic_playground_test.js
```diff
@@ -100,14 +103,14 @@ suite('Right Clicking on Blocks', function (done) {
...
   });
 
   test('clicking the enable option enables the block', async function () {
     await contextMenuSelect(browser, this.block, 'Enable Block');
-    const isEnabled = await getIsEnabled(browser, this.block.id);
-    chai.assert.isTrue(isEnabled);
+    const isDisabled = await getIsDisabled(browser, this.block.id);
+    chai.assert.isFalse(isDisabled);
```
cpcallen, line 113: Ditto:
```suggestion
    chai.assert.isFalse(await getIsDisabled(browser, this.block.id))
```

tests/browser/test/basic_playground_test.js
```diff
@@ -127,3 +130,99 @@ suite('Right Clicking on Blocks', function (done) {
...
   });
 });
+
+suite('Disabling', function () {
+  // Setting timeout to unlimited as the webdriver takes a longer
+  // time to run than most mocha tests.
+  this.timeout(0);
+
+  suiteSetup(async function () {
+    browser = await testSetup(testFileLocations.playground);
```
cpcallen, line 140: This will break when merged to develop due to a refactor of `testFileLocations`.  Merge develop into this PR then fix:
```suggestion
    browser = await testSetup(testFileLocations.PLAYGROUND);
```

cpcallen (approved):

tests/browser/test/basic_playground_test.js
BeksOmega: I was trying to match the format that Rachel used.  But I'll go through and change all of 'em!

tests/browser/test/basic_playground_test.js
BeksOmega: I will rebase and fix!

tests/browser/test/basic_playground_test.js
rachel-fenichel: fwiw, I used separate lines because I like having `await` at the beginning of the line.

PR #7280, opened 2023-07-11
chore(tests): Add documentation to browser test helper functions

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Fixes lack of jsdoc.

### Proposed Changes

Change jsdoc rules for browser tests to not require types, because we don't have strong typing here (and do not intend to compile this code before running it).
Add jsdoc for functions in `test_helpers.js`.

#### Behavior Before Change

No change in test behaviour.

### Reason for Changes

Documentation!

### Test Coverage
These are tests.

### Documentation

This is documentation!

### Additional Information

I think I got the promises right but would appreciate a double-check.

### DIFF
--- a/tests/browser/test/test_setup.js
+++ b/tests/browser/test/test_setup.js
@@ -7,6 +7,10 @@
 /**
  * @fileoverview Node.js script to run automated functional tests in Chrome, via webdriver.
  * This file is to be used in the suiteSetup for any automated fuctional test.
+ * Note: In this file many functions return browser elements that can
+ * be clicked or otherwise interacted with through Selenium WebDriver. These
+ * elements are not the raw HTML and SVG elements on the page; they are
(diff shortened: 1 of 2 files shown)


tests/browser/test/test_setup.js
```diff
@@ -126,6 +142,13 @@ async function getCategory(browser, categoryName) {
   return category;
 }
 
+/**
+ * @param browser The active WebdriverIO Browser object.
+ * @param categoryName The name of the toolbox category to search.
+ * @param n Which block to select, indexed from the top of the category.
```
BeksOmega, line 148:
```suggestion
 * @param n Which block to select, 0-indexed from the top of the category.
```

Not sure if this is necessary since it's convention, but migth be worth clarifying!

BeksOmega (approved):

PR #7284, opened 2023-07-12
chore(tests): Add tests for adding and removing comments

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Resolves [#7276](https://github.com/google/blockly/issues/7276)

### Proposed Changes

- Switch to clearing the workspace after tests in this suite, instead of reloading the window before tests.
- Add tests for adding and removing comments through the right-click menu

### DIFF
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -70,6 +70,10 @@ suite('Testing Connecting Blocks', function (done) {
   });
 });
 
+/**
+ * These tests have to run together. Each test acts on the state left by the
+ * previous test, and each test has a single assertion.
+ */
 suite('Right Clicking on Blocks', function (done) {
   // Setting timeout to unlimited as the webdriver takes a longer time to run than most mocha test
   this.timeout(0);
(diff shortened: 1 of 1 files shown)


tests/browser/test/basic_playground_test.js
```diff
@@ -129,6 +136,27 @@ suite('Right Clicking on Blocks', function (done) {
     chai.assert.isTrue(isEnabled);
   });
 
+  test('Add Comment', async function () {
+    const block = await dragNthBlockFromFlyout(browser, 'Loops', 0, 20, 20);
+
+    const blockId = block.id;
```
BeksOmega, line 142: Since the `blockId` is only used once, I think it makes sense to just pass it to `getCommentText`!

tests/browser/test/basic_playground_test.js
```diff
@@ -129,6 +136,27 @@ suite('Right Clicking on Blocks', function (done) {
...
+
+    const commentText = await getCommentText(browser, blockId);
+    chai.assert.equal(commentText, '');
+  });
+
+  test('Remove Comment', async function () {
+    const block = await dragNthBlockFromFlyout(browser, 'Loops', 0, 20, 20);
+
+    const blockId = block.id;
+    await contextMenuSelect(browser, block, 'Add Comment');
```
BeksOmega, line 153: Christopher and I discussed in a different PR that since this is actually part of setup, it would be better to move it into the previous chunk of code (under `const block = await`) and use Blockly's APIs rather than the context menu.

tests/browser/test/basic_playground_test.js
```diff
@@ -73,8 +79,13 @@ suite('Right Clicking on Blocks', function (done) {
     browser = await testSetup(testFileLocations.playground);
   });
 
+  teardown(async function () {
+    await browser.execute(() => {
+      Blockly.getMainWorkspace().clear();
+    });
+  });
+
   test('Collapse', async function () {
```
BeksOmega, line 88: Could you also go through and update the test titles in this suite to be of the form "this tests that X" where X becomes the title? For example, this test could be come `test('clicking the collapse option collapses the block')`.

Here's an article about why descriptive test names are important! https://testing.googleblog.com/2014/10/testing-on-toilet-writing-descriptive.html

BeksOmega (approved):
LGTM! Just a few nits =) Feel free to merge after fixing!

tests/browser/test/basic_playground_test.js
rachel-fenichel: Done.

tests/browser/test/basic_playground_test.js
rachel-fenichel: Do you mean that I should use json to load a block in the correct location, with the correct value for enabled/comment/collapsed, rather than dragging in?

BeksOmega: In [my other PR Christopher suggested](https://github.com/google/blockly/pull/7279#discussion_r1261040532) using `browser.execute` to disable it. But you're right it does seem a bit weird to drag out the block using webdriver, but disable it using the API. So up to you! I'll fix my PR to match if necessary.

PR #7295, opened 2023-07-17
fix(generators): Changes to exports and access controls for TypeScript compatibility

### LINKED ISSUE #7283: TypeScript unhappy with 'blockly/javascript' exports
## TL;DR 

```ts
import { Order } from 'blockly/javascript';
```
Fails to compile without `// @ts-ignore` flag

### Temporary fix

Adding `// @ts-ignore` comment will allow it to compile and run without any noted errors
```ts
// @ts-ignore
import { Order } from 'blockly/javascript';
```

### Details

Didn't have time to look too deeply into the issue, but this is what I've dug up from a cursory look:

Looks like `Order` is not being exported in the `javascript.d.ts` file - https://github.com/google/blockly/blob/blockly-v10.0.1/typings/javascript.d.ts

Might be because typescript does not like the `export * from ...` syntax in original `javascript.js` source - https://github.com/google/blockly/blob/blockly-v10.0.1/generators/javascript.js

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7283.
Also partially fixes google/blockly-samples#1785.

### Proposed Changes

* Add missing declarations for `Order` enums to `typings/<language>.d.ts`. 

  This is a temporary measure until the generators are converted to TypeScript and the `.d.ts` files can be autogenerated.)

* Remove the `protected` declaration on the `provideFunction_` method and `FUNCTION_NAME_PLACEHOLDER_` property on `CodeGenerator` so they can be used from generator functions written in TypeScript.

  Previously generator functions were (ES5-style) methods on (Code)`Generator` instances, and Closure Compiler was apparently happy with this, but now these functions live in a separate dictionary and `tsc` is pickier.

#### Behaviour Before Change

It was not possible to write code generators in TypeScript that made use of `Order`, `provideFunction_` or `FUNCTION_NAME_PLACEHOLDER_`.  The former limitation had to be worked around by using the now-deprecated `ORDER_*` properties on the `CodeGenerator` instance instead.

#### Behaviour After Change

It is now possible to use all three features in TypeScript.

### Reason for Changes

Fix bugs.

### Test Coverage

Manually tested changes (to `javascript/javascript_generator.js` only) via `blockly-samples`' `examples/sample-app-ts`.

### Documentation

We should regenerate the documentation for `CodeGenerator` to remove the `protected` declaration.

It turns out we have no auto-generated reference documentation for anything in `generators/` so there is nothing needing updating on that front.  🙁

### Additional Information

This fix should be cherry-picked into a patch release ASAP.

### DIFF
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -49,7 +49,7 @@ export class CodeGenerator {
    * legitimately appear in a function definition (or comment), and it must
    * not confuse the regular expression parser.
    */
-  protected FUNCTION_NAME_PLACEHOLDER_ = '{leCUI8hutHZI4480Dc}';
+  FUNCTION_NAME_PLACEHOLDER_ = '{leCUI8hutHZI4480Dc}';
   FUNCTION_NAME_PLACEHOLDER_REGEXP_: RegExp;
 
   /**
@@ -471,7 +471,7 @@ export class CodeGenerator {
    * @returns The actual name of the new function.  This may differ from
    *     desiredName if the former has already been taken by the user.
    */
-  protected provideFunction_(
+  provideFunction_(
     desiredName: string,
     code: string[] | string
   ): string {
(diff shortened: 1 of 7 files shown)


BeksOmega (approved):
LGTM!

Issue #7068, opened 2023-05-09
Make icon bubble use the rendering queue


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Problem

Icon rendering currently has a special case carved out for it where it will cause an immediate render. This is because when a mutator bubble is shown with `setVisible` the bubble needs to know where the block is in order to show the bubble in the correct location.

The special case carved out for mutators is ugly, and not performant, because it doesn't wait for the next animation frame or use the render queue, it forces an immediate re-render.

### Request

Change `setVisible` to wait for the `afterQueuedRenders` promise before making the bubble actually visible. `setVisible` should also return a promise in case external developers (or the tests) need to wait on the promise before using the results of `setVisible` such as access to the mutator workspace. Note that this is a breaking change, but one that we think is acceptable because it would eliminate the special case for icons and make rendering more predictable and consistent. However, because this is a breaking change, a PR should only be submitted at times we've already agreed to do a breaking change next (like this quarter)

Note also there are a large number of tests in core that need to be updated to wait on the promise before checking that the bubble is positioned correctly.

### Additional context

https://github.com/google/blockly/pull/7024 for the start of this implementation, but backed out because the tests need to be fixed and we didn't want to hold up the rest of the change on this since Beka is busy.

BeksOmega:
Since we're not sure if we're going to do a major release this quarter or not, my plan is to is to add a function to the render management system that forces an immediate render of all queued renders. (As I originally suggested in #7024 ) This will allow us to use the new rendering system, but in such a way that it maintains the existing immediate-render behavior of the icons.

The next time we do a major release we can switch it to the fully correct implementation (the promises based one that Maribeth suggests).

But the immediate-render fix at least unblocks tearing out the remnants of the old system.

PR #7296, opened 2023-07-17
fix: have icons use the new render management system

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #7068 

### Proposed Changes

1) Adds a new method to the render management system to immediately trigger any queued renders. This should only be used in cases where it's necessary for backwards compatibility (like this one).
2) Switches icon methods over to use the render management system (with the immediate trigger).
3) Adds dummy promises to the return value of all `setBubbleVisible` methods.

### Reason for Changes

1) We need adding icons to trigger a render immediately so that if the bubbles are set to be visible immediately, they are positioned correctly. Eventually we want to move this to be promises-based (so setting the visibility will wait for any queued renders to finish first). But that is a breaking change.
2) This gets us one step closer to tearing out the old render management system.
3) This makes it easier for external devs to transition to the new promises based system.

### Test Coverage

All tests pass! Also quickly tested `block.setCommentText('test'); block.getIcon('comment').setBubbleVisible(true);` in the browser console to reassure myself that it was indeed working.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -59,7 +59,7 @@ import * as svgMath from './utils/svg_math.js';
 import {WarningIcon} from './icons/warning_icon.js';
 import type {Workspace} from './workspace.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
-import {queueRender} from './render_management.js';
+import * as renderManagement from './render_management.js';
 import * as deprecation from './utils/deprecation.js';
 import {IconType} from './icons/icon_types.js';
 
@@ -983,8 +983,8 @@ export class BlockSvg
       icon.initView(this.createIconPointerDownListener(icon));
       icon.applyColour();
       icon.updateEditable();
-      // TODO: Change this based on #7068.
-      this.render();
+      this.queueRender();
+      renderManagement.triggerQueuedRenders();
       this.bumpNeighbours();
     }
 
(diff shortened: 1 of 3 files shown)


core/block_svg.ts
```diff
@@ -1012,8 +1012,8 @@ export class BlockSvg
     if (type.equals(MutatorIcon.TYPE)) this.mutator = null;
 
     if (this.rendered) {
-      // TODO: Change this based on #7068.
-      this.render();
+      this.queueRender();
+      renderManagement.triggerQueuedRenders();
```
rachel-fenichel, line 1016: You can move the trigger into `queueRender` since you're always doing them together.

rachel-fenichel: I take it back--in a later PR you aren't doing them together.

BeksOmega: In this PR yes! But there are other places throughout the `BlockSvg` class where I don't trigger an immediate render. I.e. I use the render management system properly :P

cpcallen (approved):
Make sure that this PR still has a `deprecation` tag once it's been submitted: one of the bots seems to like to try to remove that tag.

Not for this PR, but: given similarity of code have you considered introducing a `BubbleIcon` intermediate class between `Icon` and `Comment`/`Mutator`/`WarningIcon`?

BeksOmega:
Maribeth actually requested on the next pull request that I remove the dummy promises, and we just cleanly break people later. So I've reverted that change and removed the deprecation notice!

PR #7306, opened 2023-07-19
fix: make deserialization use the new render management system

### LINKED ISSUE #7290: Make JSON deserialization use the render queue
### Check for duplicates

### Problem

JSON Deserialization currently still uses the old immediate rendering system, because I haven't had time to clean it up yet. 

### Request

Instead it should use the render queue, so we can fully tear out the old system.

### Alternatives considered

N/A

### Additional context

I think I can get away with not updating XML by using the same immediate-render function suggested in https://github.com/google/blockly/issues/7068#issuecomment-1635099700, but I'll know more once I dig into this!

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes #7290 

### Proposed Changes

Changed both the JSON and XML deserialization systems to use the render queue, but trigger an immediate render.

### Reason for Changes

Moving both sytems to use the render queue allows us to get rid of the old render management system. But I chose to make them trigger an immediate render (instead of waiting until the end of the frame) to ensure 100% backwards compatibility.

We could have people use the `finishQueuedRenders` method if they want to calculate things based on the size/position of the blocks after deserialization. However, as we discovered last time I was reorganizing stuff, that's technically a breaking change :/ And I think deserialization is a much more high-risk area for this kind of thing! (as opposed to like, adding an input to a block).

### Test Coverage

* Verified that there is no performance impact for deserialization. The spaghetti tests ran in the same amount of time before and after this change.
* Verified that copying and pasting properly renders the block.
* Verified that setting shadows programatically properly renders the block.

### Documentation

N/A

### Additional Information

N/A

Dependent on: https://github.com/google/blockly/pull/7296

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -36,6 +36,7 @@ import * as Variables from './variables.js';
 import {WorkspaceSvg} from './workspace_svg.js';
 import * as utilsXml from './utils/xml.js';
 import * as Xml from './xml.js';
+import * as renderManagement from './render_management.js';
 
 enum FlyoutItemType {
   BLOCK = 'block',
@@ -626,6 +627,8 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
     const parsedContent = toolbox.convertFlyoutDefToJsonArray(flyoutDef);
     const flyoutInfo = this.createFlyoutInfo(parsedContent);
 
+    renderManagement.triggerQueuedRenders();
+
     this.layout_(flyoutInfo.contents, flyoutInfo.gaps);
 
     if (this.horizontalLayout) {
@@ -770,7 +773,7 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
       ) as Element;
       block = this.getRecycledBlock(xml.getAttribute('type')!);
       if (!block) {
-        block = Xml.domToBlock(xml, this.workspace_);
+        block = Xml.domToBlockInternal(xml, this.workspace_);
       }
     } else {
       block = this.getRecycledBlock(blockInfo['type']!);
@@ -779,7 +782,10 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
           blockInfo['enabled'] =
             blockInfo['disabled'] !== 'true' && blockInfo['disabled'] !== true;
         }
-        block = blocks.append(blockInfo as blocks.State, this.workspace_);
+        block = blocks.appendInternal(
+          blockInfo as blocks.State,
+          this.workspace_
+        );
       }
     }
 
--- a/core/xml.ts
+++ b/core/xml.ts
@@ -23,6 +23,7 @@ import type {Workspace} from './workspace.js';
 import {WorkspaceComment} from './workspace_comment.js';
 import {WorkspaceCommentSvg} from './workspace_comment_svg.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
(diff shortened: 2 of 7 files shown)


core/flyout_base.ts
```diff
@@ -626,8 +627,11 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
     const parsedContent = toolbox.convertFlyoutDefToJsonArray(flyoutDef);
     const flyoutInfo = this.createFlyoutInfo(parsedContent);
 
+    renderManagement.triggerQueuedRenders();
```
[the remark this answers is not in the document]
maribethb: > Would it make sense to make show async in the future

Seems reasonable.

> If so I can set this up to return a resolved promise to make it easier for folks to transition.

by "this" do you mean the `show` method? similar to my comments below, I feel like that's a bit misleading right now. also it's kind of hard to communicate that they can go ahead and wait on a promise now. if at some point we do a breaking change where we make show async, then that's when it would make the most sense to announce the change, so idk what we'd get out of adding an immeditately-resolved promise right now

core/icons/comment_icon.ts
```diff
@@ -237,13 +237,24 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
     return this.bubbleVisiblity;
   }
 
-  setBubbleVisible(visible: boolean): void {
-    if (visible && (this.textBubble || this.textInputBubble)) return;
-    if (!visible && !(this.textBubble || this.textInputBubble)) return;
+  /**
+   * Sets whether the bubble is visible or not.
+   *
+   * @returns A promise that resolves when the the bubble has finished opening.
```
maribethb, line 243: This seems kind of misleading since you're not actually waiting on anything to finish. Seems like the `show*EditableBubble` methods would have to also return promises in order for this description to be accurate. Currently not sure of the value of adding these immediately-resolving promises. Is this the same thing you're proposing to do in the flyout.show method?

core/xml.ts
```diff
@@ -545,6 +547,27 @@ export function appendDomToWorkspace(
...
+ *
+ * This is defined internally so that it doesn't trigger an immediate render,
+ * which we do want to happen for external calls.
+ *
+ * @param xmlBlock XML block element.
+ * @param workspace The workspace.
+ * @returns The root block created.
+ * @internal
+ */
+export function domToBlockInternal(
```
maribethb, line 567: I get why you made `domToBlock` render immediately, but I don't love this name. It's not very descriptive, and it implies external developers would never want to call it. But in reality they might be perfectly fine not triggering an immediate render. What about `domToBlockAsync`? or something better than that in the same vein? Then we don't have to mark it as internal.

Or is your plan to eventually make the breaking change to `domToBlock`? in that case i don't care quite as much if this method is intended to be temporary

maribethb:
sorry didn't realize this was a draft

core/icons/comment_icon.ts
BeksOmega: I removed these in the previous PR, and I'm not planning to add promises to the flyout related stuff (based on your feedback)!

core/xml.ts
BeksOmega: My plan is indeed to eventually make the breaking change to `domToBlock` (unless there's an important reason we don't want to do that)! I chose the `domToBlockInternal` name to match the naming of the JSON serialization systsem.

maribethb (approved):

PR #7307, opened 2023-07-19
fix: make insertion markers use new render management system

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

Makes is so that insertion markers use the new render management system.
### Reason for Changes

Necessary for tearing out the ~new~ old render management system.

### Test Coverage

Manually tested that insertion markers appear and look correct.

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/7296

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -25,6 +25,7 @@ import type {IDragTarget} from './interfaces/i_drag_target.js';
 import type {RenderedConnection} from './rendered_connection.js';
 import type {Coordinate} from './utils/coordinate.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
+import * as renderManagement from './render_management.js';
 
 /** Represents a nearby valid connection. */
 interface CandidateConnection {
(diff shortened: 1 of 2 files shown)


core/workspace_svg.ts
```diff
@@ -1251,11 +1252,13 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
       for (let i = 0; i < imList.length; i++) {
-        imList[i].render(false);
+        imList[i].queueRender();
       }
     }
 
-    this.markerManager.updateMarkers();
+    renderManagement
+      .finishQueuedRenders()
+      .then(() => void this.markerManager.updateMarkers());
```
maribethb, line 1261: TIL about making non-leaky arrow functions

core/insertion_marker_manager.ts
```diff
@@ -608,18 +609,17 @@ export class InsertionMarkerManager {
...
 
-    if (imConn && closest) {
-      // Position so that the existing block doesn't move.
-      insertionMarker.positionNearConnection(imConn, closest);
-    }
-    if (closest) {
-      // Connect() also renders the insertion marker.
-      imConn.connect(closest);
-    }
+    // Position so that the existing block doesn't move.
```
maribethb, line 615: I can see why `imConn` has to exist (we throw an error above if it doesn't) but don't see where we're otherwise validating that `closest` exists. Maybe worth keeping the check? or am i missing where we're checking?

maribethb (approved):

maribethb:
> Reason for Changes
> Necessary for tearing out the **new** render management system.

😱 😛

core/insertion_marker_manager.ts
BeksOmega: `closest` is non-nullable! It is always guaranteed to be part of the `activeCandidate`.

PR #7308, opened 2023-07-19
fix: remove old `render` method

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

Removes the last vestiges of the old render management system.

### Reason for Changes

Less variants! 

### Test Coverage

All tests pass. Smoke tested the playground and it looks good!

Also tested that hiding and showing the workspace works properly.

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/7306 and https://github.com/google/blockly/pull/7307

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -78,6 +78,7 @@ import * as Xml from './xml.js';
 import {ZoomControls} from './zoom_controls.js';
 import {ContextMenuOption} from './contextmenu_registry.js';
 import * as renderManagement from './render_management.js';
+import * as deprecation from './utils/deprecation.js';
 
 /** Margin around the top/bottom/left/right after a zoomToFit call. */
 const ZOOM_TO_FIT_MARGIN = 20;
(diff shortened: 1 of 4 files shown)


core/workspace_svg.ts
```diff
@@ -1224,24 +1225,20 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
-      for (let i = blocks.length - 1; i >= 0; i--) {
-        blocks[i].markDirty();
-      }
-
-      this.render();
-      if (this.toolbox_) {
-        this.toolbox_.position();
-      }
-    } else {
+    if (!isVisible) {
```
rachel-fenichel, line 1228: Why is this one not needed? Setting the workspace visibility doesn't directly change the blocks or fields, so no rerender would be triggered.

Put another way: what case was this handling before?

BeksOmega: I have no idea what case this was handling before. But I can't think of a reason it's needed now! I manually tested showing and hiding the workspace, and everything looks correct.

BeksOmega: Dug into the blame, and the render existed originally when `setVisible` was implemented.: https://github.com/google/blockly/commit/e8ab929b6ae9f1d84962c91489a5f6a04792b728 So maybe itw as necessary at some point? Or maybe it was just added cautiously.

I tested adding block to the workspace while the workspace was hidden, and that also works and renders things correctly. So I think this is safe to remove!

rachel-fenichel (approved):

PR #7324, opened 2023-07-24
fix(tests): fix broken browser tests

rachel-fenichel:
## The basics

- [x] I branched from develop

### Proposed Changes

- Change offset math to offset by a fixed amount from top-left, rather than a percent
- Use correct input for one of the tests (`DO0` instead of `IF0`)
- Use a different category for one of the drag tests, because one of the blocks in the category was broken. That broken block is also fixed in a different PR/set of PRs.
- Switched from the 'landing' reporter, which the mocha documentation calls a 'gimmick', to one that prints the names of tests to the console as it runs.

#### Behavior Before Change

Tests worked for Beka but not for me

#### Behavior After Change

Hopefully tests work for everyone

### Reason for Changes

### DIFF
--- a/tests/browser/test/basic_block_test.js
+++ b/tests/browser/test/basic_block_test.js
@@ -14,7 +14,7 @@ const {
   testSetup,
   testFileLocations,
   getAllBlocks,
-  dragNthBlockFromFlyout
+  dragNthBlockFromFlyout,
 } = require('./test_setup');
 const {Key} = require('webdriverio');
 
@@ -34,13 +34,7 @@ suite('Basic block tests', function (done) {
 
   test('Drag three blocks into the workspace', async function () {
     for (let i = 1; i <= 3; i++) {
-      await dragNthBlockFromFlyout(
-        browser,
-        'Align',
-        0,
-        250,
-        50 * i
-      );
+      await dragNthBlockFromFlyout(browser, 'Align', 0, 250, 50 * i);
       chai.assert.equal(i, (await getAllBlocks(browser)).length);
     }
   });
--- a/tests/browser/test/basic_block_test.js
+++ b/tests/browser/test/basic_block_test.js
@@ -14,10 +14,7 @@ const {
   testSetup,
   testFileLocations,
   getAllBlocks,
-  getSelectedBlockElement,
-  switchRTL,
-  dragBlockTypeFromFlyout,
-  screenDirection,
+  dragNthBlockFromFlyout
 } = require('./test_setup');
 const {Key} = require('webdriverio');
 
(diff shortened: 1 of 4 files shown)


tests/browser/test/basic_block_test.js
```diff
@@ -37,14 +34,8 @@ suite('Basic block tests', function (done) {
...
-      await dragBlockTypeFromFlyout(
-        browser,
-        'Basic',
-        'test_basic_empty',
-        250,
-        50 * i
-      );
-      chai.assert.equal((await getAllBlocks(browser)).length, i);
+      await dragNthBlockFromFlyout(browser, 'Align', 0, 250, 50 * i);
+      chai.assert.equal(i, (await getAllBlocks(browser)).length);
```
BeksOmega, line 38: It should be the actual vaue and then the expecte value (swap le arguments).

tests/browser/test/basic_block_test.js
```diff
@@ -37,14 +34,8 @@ suite('Basic block tests', function (done) {
...
     for (let i = 1; i <= 3; i++) {
-      await dragBlockTypeFromFlyout(
-        browser,
-        'Basic',
-        'test_basic_empty',
-        250,
-        50 * i
-      );
-      chai.assert.equal((await getAllBlocks(browser)).length, i);
+      await dragNthBlockFromFlyout(browser, 'Align', 0, 250, 50 * i);
```
BeksOmega, line 37: Was there a reason this needed to be switched out?

tests/browser/test/basic_playground_test.js
```diff
@@ -182,7 +182,7 @@ suite('Disabling', function () {
         110,
         110
       );
-      await connect(browser, child, 'PREVIOUS', parent, 'IF0');
```
BeksOmega, line 185: Huh, I guess this was just getting it close enough that it was working? lol

BeksOmega (approved):
Works on my machine too!

tests/browser/test/basic_block_test.js
rachel-fenichel: One of the basic blocks was breaking (I think it's one that you fixed in dev-tools) and it was making testing difficult.

I want to have a different test to validate that every test block can be shown in the toolbox and dragged in, but this particular test doesn't actually care which block you use.

tests/browser/test/basic_block_test.js
rachel-fenichel: Done. My expectations were backwards!

tests/browser/test/basic_playground_test.js
rachel-fenichel: Yup.

PR #7328, opened 2023-07-24
chore(tests): use a shared chrome instance for all browser tests

### LINKED ISSUE #7325: Make browser tests use a single shared browser instance

rachel-fenichel:
## The basics
- [x] I branched from develop

### Resolves

Fixes #7325

### Proposed Changes

- Use mocha [root hooks](https://mochajs.org/#available-root-hooks) to do setup and teardown before all tests and after all tests.
  - Note that this means the tests won't behave nicely with the `--parallel` flag. That's fine: we weren't using it before either.
- Change the `testSetup` function to grab the shared browser instead of making a new one.
- Change all tests to store their `browser` instance on the test, rather than on a file-global variable. Update references as needed.
- Remove most suite teardowns, since all they did was close the browser.
- Change the block factory suite teardown to clear the workspace. This avoids a confirmation pop-up that was making tests hang.

#### Behavior Before Change

Every suite opened a new window and stole keyboard and mouse focus.

#### Behavior After Change

Only one window for all of the tests, so you can type while tests run.

### Reason for Changes

Make testing less annoying.


cpcallen:
> * Change all tests to store their `browser` instance on the test, rather than on a file-global variable. Update references as needed.

@rachel-fenichel: why?

rachel-fenichel:
Mostly so that it's clear when you have or haven't set up the browser for your particular test, either by refreshing or by loading a new page. If it's shared it will always work for everything but the first test in the file, even if the setup is wrong or missing.

PR #7331, opened 2023-07-27
feat: add basic pasters

### LINKED ISSUE #7336: Implement IPasters for copy paste
Implement an `IPaster` interface which allows us to easily add support for pasting new kinds of objects.

```ts
interface IPaster<T implements ICopyData> {
  paste(copyData: T, ws: workspace, coord?: Coordinate): ICopyable<T>;
}

```

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #7336 

### Proposed Changes

Add an IPaster interface and pasters for blocks and workspace comments.

### Reason for Changes

Work toward having separate pasters for separate objects so that we can easily add new pastable objects.

### Test Coverage

N/A - This is just using the serialization system so no need to test it separately. I'll add tests for the integration between the clipboard and the pasters when I get to that point! (I.e. to make sure the correct APIs are getting called).

### Documentation

N/A

### Additional Information

Currently these are not actually used anywhere :P I'm going to wire them in in a future PR.

### DIFF
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -1,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {BlockSvg} from '../block_svg.js';
+import {CopyData} from '../interfaces/i_copyable';
+import {IPaster} from '../interfaces/i_paster.js';
+import {State, append} from '../serialization/blocks';
+import {Coordinate} from '../utils/coordinate.js';
+import {WorkspaceSvg} from '../workspace_svg.js';
+
+export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
+  paste(
+    copyData: BlockCopyData,
+    workspace: WorkspaceSvg,
+    coordinate?: Coordinate,
+  ): BlockSvg | null {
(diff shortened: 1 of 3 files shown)


core/clipboard/block_paster.ts
```diff
@@ -0,0 +1,31 @@
...
+import {CopyData} from '../interfaces/i_copyable';
+import {IPaster} from '../interfaces/i_paster.js';
+import {State, append} from '../serialization/blocks';
+import {Coordinate} from '../utils/coordinate.js';
+import {WorkspaceSvg} from '../workspace_svg.js';
+
+export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
+  paste(
+    copyData: BlockCopyData,
+    workspace: WorkspaceSvg,
```
rachel-fenichel, line 17: Does the existing copy/paste system work with headless workspaces? If no, we can decide that working in headless is out of scope, and then this code gets to be simple. If yes, we have to decide whether to keep that or deliberately break it.

BeksOmega: Current code only works in rendered workspaces! `Block`s are not `ICopyable` only `BlockSvg`s are =) Same for workspace comments.

rachel-fenichel (approved):

Issue #7334, opened 2023-07-31
Variable block wrong shape for zelos


zfangqijun:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

After creating boolean and other variable blocks, the shape of the getter block and the shape of the input slot of the setter are incorrect.

### Reproduction steps

1. Create string, number, boolean variable blocks
2. In flyout, you can see it immediately
3. In the workspace, modify the variable options

Reproducible code：https://github.com/zfangqijun/blockly/commit/6dda0227e9d395c836618ab7261125efc05847bd

### Screenshots

<img width="488" alt="image" src="https://github.com/google/blockly/assets/51962507/dca996fb-07a5-4a35-b81f-a9a2ab85fa96">

<img width="444" alt="image" src="https://github.com/google/blockly/assets/51962507/16214fc4-f280-45b8-83c3-05fc94f6acaa">

PR #7335, opened 2023-07-31
fix(blocks): boolean variable block wrong shape for zelos

zfangqijun:
## The basics

- [x] I branched from develop

### Resolves

Fixes https://github.com/google/blockly/issues/7334

#### Behavior Before Change

image
image

#### Behavior After Change

image

### Reason for Changes

I think it's a bug 🤔️

### Additional Information

Is there a better way than rerender?

### DIFF
--- a/blocks/variables_dynamic.ts
+++ b/blocks/variables_dynamic.ts
@@ -25,6 +25,7 @@ import {
   defineBlocks,
 } from '../core/common.js';
 import '../core/field_label.js';
+import type {BlockSvg} from '../core/block_svg.js';
 
 /**
  * A dictionary of the block definitions provided by this module.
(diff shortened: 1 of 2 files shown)


blocks/variables_dynamic.ts
```diff
@@ -156,6 +157,10 @@ const CUSTOM_CONTEXT_MENU_VARIABLE_GETTER_SETTER_MIXIN = {
     } else {
       this.getInput('VALUE')!.connection!.setCheck(variableModel.type);
     }
+
+    if (this.rendered) {
+      (this as BlockSvg).queueRender();
+    }
```
BeksOmega, line 163: Instead of having this inside the individual block definition, you can override `setCheck` in the [rendered connection](https://github.com/google/blockly/blob/develop/core/rendered_connection.ts) class, and call `queueRender` there! This way whenever a connection check gets updated on any block, we'll be sure the shapes of the connections get updated =)

BeksOmega (changes requested):
Thank you for finding this bug and submitting a fix :D The change looks great, but I think there's actually a more general place we can put it!

blocks/variables_dynamic.ts
zfangqijun: Thank you for taking the time to review my PR and providing your feedback. I appreciate your suggestions. updated

BeksOmega (approved):
Sweet, LGTM! I'll get this merged after CI Passes =) Thank you for your hard work on this!!

Issue #7337, opened 2023-07-31
Refactor ICopyable for copy paste


BeksOmega:
Change `ICopyable` to be more generic (and use generics!)

```ts
interface ICopyable<T extends ICopyData> extends ISelectable {
  toCopyData(): T;
}

interface ICopyData {
  paster: string;
}
```

PR #7344, opened 2023-07-31
fix: refactor CopyData interface to have the correct structure

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Work on #7337 

### Proposed Changes

Changes the `CopyData` interface to `ICopyData` and changes it to only require the `paster: string;` property.

Updates everything using this type/downstream APIs to conform to the new interface.

### Reason for Changes

Working toward more flexible copy pasting.

### Test Coverage

Should be covered by existing tests!

### Documentation

N/A

### Additional Information

Best to review commit-wise due to renames =)

### DIFF
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -23,15 +23,17 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     if (!workspace.isCapacityAvailable(copyData.typeCounts!)) return null;
 
     if (coordinate) {
-      copyData.saveInfo['x'] = coordinate.x;
-      copyData.saveInfo['y'] = coordinate.y;
+      copyData.blockState['x'] = coordinate.x;
+      copyData.blockState['y'] = coordinate.y;
     }
-    return append(copyData.saveInfo, workspace, {recordUndo: true}) as BlockSvg;
+    return append(copyData.blockState, workspace, {
+      recordUndo: true,
+    }) as BlockSvg;
   }
 }
 
 export interface BlockCopyData extends ICopyData {
-  saveInfo: State;
+  blockState: State;
   typeCounts: {[key: string]: number};
 }
 
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -6,7 +6,7 @@
 
 import {BlockSvg} from '../block_svg.js';
 import {registry} from '../clipboard.js';
-import {CopyData} from '../interfaces/i_copyable.js';
+import {ICopyData} from '../interfaces/i_copyable.js';
 import {IPaster} from '../interfaces/i_paster.js';
 import {State, append} from '../serialization/blocks.js';
 import {Coordinate} from '../utils/coordinate.js';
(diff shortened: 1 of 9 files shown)


core/clipboard/block_paster.ts
```diff
@@ -22,15 +22,17 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
...
+      copyData.saveInfo['y'] = coordinate.y;
     }
-    return append(state, workspace, {recordUndo: true}) as BlockSvg;
+    return append(copyData.saveInfo, workspace, {recordUndo: true}) as BlockSvg;
   }
 }
 
-export interface BlockCopyData extends ICopyData {}
+export interface BlockCopyData extends ICopyData {
+  saveInfo: State;
```
cpcallen, line 34: Is this keeping the name `saveInfo` just for backwards-compatibility?  Seems like a good opportunity to rename it in each sub-interface to be more descriptive.

(Here and in `workspace_comment_paster.ts`.)

core/clipboard/block_paster.ts
```diff
@@ -22,15 +22,17 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
   ): BlockSvg | null {
     if (!workspace.isCapacityAvailable(copyData.typeCounts!)) return null;
 
-    const state = copyData.saveInfo as State;
```
cpcallen, line 25: Curiosity: are you removing this just because it isn't a very useful temporary binding, or is there a type-checking reason?

cpcallen (approved):
LGTM modulo comments below and some pedantry:

- Do the first two commits (at least) not contain breaking changes?  I note neither (nor the PR) is marked with "!" in the title nor "BREAKING CHANGE" in the description.
- I think I'd have titled most of these commits, and the PR itself, as `refactor:` rather than `chore:` or `fix:`.  
- > Changes the `CopyData` interface to `ICopyData`

  ಠ_ಠ for the usual style guide reasons (but I realise I am not going to win this).

core/clipboard/block_paster.ts
BeksOmega: Sounds good to me!

BeksOmega:
> * Do the first two commits (at least) not contain breaking changes?  I note neither (nor the PR) is marked with "!" in the title nor "BREAKING CHANGE" in the description.

The `CopyData` was only ever returned from an internal function. So since the type was inaccessible, I don't consider this a breaking change.

> * I think I'd have titled most of these commits, and the PR itself, as `refactor:` rather than `chore:` or `fix:`.

`refactor` is not one of our approved commit message prefixes :/ https://developers.google.com/blockly/guides/contribute/get-started/commits#type

The only approved ones are `chore`, `fix`, `feat`, `deprecate` or `release`.

> * > Changes the `CopyData` interface to `ICopyData`
>
>
>   ಠ_ಠ for the usual style guide reasons (but I realise I am not going to win this).

Haha I agree man. But consistency :shrug:

core/clipboard/block_paster.ts
BeksOmega: This used to be typed as `Object | Element` because workspace comments and blocks were returning the same "interface" so I needed to cast it to `State`. But since they are now using separate interfaces, there's no need to cast =)

PR #7348, opened 2023-08-01
feat: make ICopyable generic and update clipboard APIs

### LINKED ISSUE #7337: Refactor ICopyable for copy paste

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #7337

### Proposed Changes

* Updates the `ICopyable` interface to be generic.
* Updates the `clipboard` APIs to take in data instead of being stateful.
* Makes `paste` public. Everything else remains internal.

### Reason for Changes

* More informative typing.
* State is the worst.
* Necessary if we want people to paste things by passing in data.

### Test Coverage

Now that the methods aren't stateful anymore, I could test the API boundary between the clipboard and pasters :D

### Documentation

N/A

### Additional Information

N/A

## Deprecations

`Blockly.clipboard.copy` and `Blockly.clipboard.duplicate` have both been deprecated. These were marked `@internal` so if you are conforming to Blockly's API, you should not be accessing them anyway.

`Blockly.clipboard.copy` can be replaced with calling `toCopyData` on the element you want to copy.

`Blockly.clipboard.duplicate` can be replaced by calling `Blockly.clipboard.paste(myElement.toCopyData(), myWorkspace)`.

### DIFF
--- a/core/blockly.ts
+++ b/core/blockly.ts
@@ -140,7 +140,7 @@ import {ICollapsibleToolboxItem} from './interfaces/i_collapsible_toolbox_item.j
 import {IComponent} from './interfaces/i_component.js';
 import {IConnectionChecker} from './interfaces/i_connection_checker.js';
 import {IContextMenu} from './interfaces/i_contextmenu.js';
-import {ICopyable} from './interfaces/i_copyable.js';
+import {ICopyable, isCopyable} from './interfaces/i_copyable.js';
 import {IDeletable} from './interfaces/i_deletable.js';
 import {IDeleteArea} from './interfaces/i_delete_area.js';
 import {IDragTarget} from './interfaces/i_drag_target.js';
@@ -592,7 +592,7 @@ export {IComponent};
 export {IConnectionChecker};
 export {IContextMenu};
 export {icons};
-export {ICopyable};
+export {ICopyable, isCopyable};
 export {IDeletable};
 export {IDeleteArea};
 export {IDragTarget};
--- a/core/clipboard.ts
+++ b/core/clipboard.ts
@@ -24,6 +24,7 @@ let stashedWorkspace: WorkspaceSvg | null = null;
  * Copy a copyable element onto the local clipboard.
  *
  * @param toCopy The copyable element to be copied.
+ * @deprecated v11. Use `myCopyable.toCopyData()` instead. To be removed v12.
  * @internal
  */
 export function copy<T extends ICopyData>(toCopy: ICopyable<T>): T | null {
@@ -112,12 +113,21 @@ function pasteFromData<T extends ICopyData>(
  *
  * @param toDuplicate The element to be duplicated.
  * @returns The element that was duplicated, or null if the duplication failed.
+ * @deprecated v11. Use
+ *     `Blockly.clipboard.paste(myCopyable.toCopyData(), myWorkspace)` instead.
+ *     To be removed v12.
  * @internal
  */
 export function duplicate<
   U extends ICopyData,
   T extends ICopyable<U> & IHasWorkspace,
 >(toDuplicate: T): T | null {
+  deprecation.warn(
+    'Blockly.clipboard.duplicate',
+    'v11',
+    'v12',
+    'Blockly.clipboard.paste(myCopyable.toCopyData(), myWorkspace)',
+  );
   return TEST_ONLY.duplicateInternal(toDuplicate);
 }
 
@@ -128,12 +138,6 @@ function duplicateInternal<
   U extends ICopyData,
   T extends ICopyable<U> & IHasWorkspace,
 >(toDuplicate: T): T | null {
-  deprecation.warn(
-    'Blockly.clipboard.duplicate',
-    'v11',
-    'v12',
-    'Blockly.clipboard.paste(myCopyable.toCopyData(), myWorkspace)',
-  );
   const data = toDuplicate.toCopyData();
   if (!data) return null;
   return paste(data, toDuplicate.workspace) as T;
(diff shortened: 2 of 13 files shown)


core/clipboard.ts
```diff
@@ -12,52 +12,92 @@ import {BlockPaster} from './clipboard/block_paster.js';
...
 
 /**
  * Private version of copy for stubbing in tests.
  */
-function copyInternal(toCopy: ICopyable) {
-  copyData = toCopy.toCopyData();
-  source = (toCopy as any).workspace ?? null;
+function copyInternal<T extends ICopyData>(toCopy: ICopyable<T>): T | null {
+  const data = toCopy.toCopyData();
+  stashedCopyData = data; // Necessary for propery typing. This is why state sucks.
```
maribethb, line 37: Is this comment supposed to be about the line below it? I'm kinda confused about what you're saying is necessary for proper typing (?)

core/clipboard.ts
```diff
@@ -12,52 +12,92 @@ import {BlockPaster} from './clipboard/block_paster.js';
...
-let copyData: ICopyData | null = null;
+let stashedCopyData: ICopyData | null = null;
 
-let source: WorkspaceSvg | null = null;
+let stashedWorkspace: WorkspaceSvg | null = null;
 
 /**
  * Copy a block or workspace comment onto the local clipboard.
  *
  * @param toCopy Block or Workspace Comment to be copied.
```
maribethb, line 25: Update comment to match new comment in paste?

core/clipboard.ts
```diff
@@ -12,52 +12,92 @@ import {BlockPaster} from './clipboard/block_paster.js';
...
+/**
+ * Pastes the given data into the workspace, or the last copied ICopyable if
+ * no data is passed.
+ *
+ * @param copyData The data to paste into the workspace.
+ * @param workspace The workspace to paste the data into.
+ * @param coordinate The location to paste the thing at.
+ * @returns The pasted thing if the paste was successful, null otherwise.
+ */
+export function paste<T extends ICopyData>(
```
maribethb, line 72: I'm very curious how well api-documenter handles the function overloading.

core/clipboard.ts
```diff
@@ -68,23 +108,35 @@ export function paste(): ICopyable | null {
  *     duplication failed.
```
maribethb, line 108: Update comments here too?

core/clipboard.ts
```diff
@@ -68,23 +108,35 @@ export function paste(): ICopyable | null {
...
+  U extends ICopyData,
+  T extends ICopyable<U> & IHasWorkspace,
+>(toDuplicate: T): T | null {
+  const oldCopyData = stashedCopyData;
+  const oldWorkspace = stashedWorkspace;
+
+  const data = copy(toDuplicate);
+
+  // I hate side effects.
+  stashedCopyData = oldCopyData;
```
maribethb, line 131: Could you just call `toDuplicate.toCopyData()` instead of calling `copy()`? Then you wouldn't have to save and reset the stashed data.

core/clipboard.ts
```diff
@@ -68,23 +108,35 @@ export function paste(): ICopyable | null {
...
+  stashedWorkspace = oldWorkspace;
+
+  if (!data) return null;
+  return paste(data, toDuplicate.workspace) as T;
+}
+
+interface IHasWorkspace {
```
maribethb, line 138: I'm kind of confused by this. Shouldn't the duplicate method take in a `workspace` parameter like `paste` does? This method is already `@internal` so you could just update the callers to pass the workspace and not even worry about adding this particular bit of backwards compatibility.

BeksOmega: Can I just delete `duplicate` and `copy` since they're internal? I originally proposed maing them backwards compatible but deprecating them because I had forgotten they were internal by the time I wrote my design doc :P

But honestly I think it's better to handle this logic within the shortcuts/context menu options.

core/blockly.ts
```diff
@@ -592,7 +592,7 @@ export {IComponent};
 export {IConnectionChecker};
 export {IContextMenu};
 export {icons};
-export {ICopyable};
+export {ICopyable, isCopyable};
```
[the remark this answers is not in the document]
maribethb: Fair! sgtm

core/clipboard.ts
BeksOmega: Discussed offline. These are `@internal` but don't have underscores + are useful enough we think people are probably using them. We're going to deprecate v11, and remove v12.

core/clipboard.ts
BeksOmega: Yeshhh thank you!

core/clipboard.ts
BeksOmega: Sorry this was confusing: We have to assign to `data` instead of assigning to `stashedCopyData` directly because `stashedCopyData` loses type info. I just removed the comment cuz it's kinda misleading :P

core/clipboard.ts
```diff
@@ -12,79 +12,139 @@ import {BlockPaster} from './clipboard/block_paster.js';
...
+ *     `Blockly.clipboard.paste(myCopyable.toCopyData(), myWorkspace)` instead.
+ *     To be removed v12.
  * @internal
  */
-export function duplicate(toDuplicate: ICopyable): ICopyable | null {
+export function duplicate<
+  U extends ICopyData,
+  T extends ICopyable<U> & IHasWorkspace,
+>(toDuplicate: T): T | null {
+  deprecation.warn(
```
maribethb, line 125: would rather not add the deprecation warning until you've fixed the internal usages, but since this is going into a feature branch and not directly into the release i won't stop you

maribethb (approved):

core/clipboard.ts
BeksOmega: I have a draft PR fixing all internal usages :D https://github.com/google/blockly/pull/7352

PR #7350, opened 2023-08-01
chore(tests): cleanup and toolbox drag tests

### LINKED ISSUE #7327: Browser test: open all categories and drag out all blocks
### Check for duplicates

### Problem

_No response_

### Request

Add a browser test that loops over all toolbox categories, opening the category and dragging out each block from the flyout. This should test initialization for all of our built-in blocks.

This test should apply to all of our test toolboxes: simple, categories, and test blocks.

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #7346: Browser test helper: isBlockVisible
Add a helper to check whether a block is visible, meaning inside the bounds of the Blockly div.

Basic code looks like this:
```
const isVisible = await browser.execute((elem) => {
  const rect = elem.getBoundingClientRect()

  const blocklyDiv = document.getElementById('blocklyDiv');
  const blocklyRect = blocklyDiv.getBoundingClientRect();

  const vertInView = (rect.top >= blocklyRect.top) && (rect.bottom <= blocklyRect.bottom);
  const horInView = (rect.left >= blocklyRect.left) && (rect.right <= blocklyRect.right);

  return (vertInView && horInView)
}, flyoutBlock);
```

### LINKED ISSUE #7347: Browser test helper: scroll flyout
Add a helper to scroll the flyout using the scrollbar.

Basic code looks like this: 
```
/**
 * Find the flyout's scrollbar and scroll by the specified amount.
 * This makes several assumptions:
 *  - A flyout with a valid scrollbar exists, is open, and is in view.
 *  - The workspace has a trash can, which means it has a second (hidden) flyout.
 * @param xDelta How far to drag the flyout in the x direction. Positive is right.
 * @param yDelta How far to drag thte flyout in the y direction. Positive is down.
 */
async function scrollFlyout(browser, xDelta, yDelta) {
  // There are two flyouts on the playground workspace: one for the trash can
  // and one for the toolbox. We want the second one.
  // This assumes there is only one scrollbar handle in the flyout, but it could 
  // be either horizontal or vertical.
  await browser.pause(50);
  const selected = await browser.$$(`.blocklyFlyoutScrollbar`)[1].$(`rect.blocklyScrollbarHandle`);
  await browser.pause(50);
  await selected.dragAndDrop({x: xDelta, y: yDelta});
  await browser.pause(50);
}
```

Some error handling for a missing or invisible flyout scrollbar would be helpful.

rachel-fenichel:
## The basics

- [x] I branched from develop

### Resolves

Closes #7327 
Closes #7347 
Closes #7346

### Proposed Changes

In order of commits:
- Test cleanup: use helpers
- Test cleanup: miscellaneous
- Format
- Add test to drag out every block from the toolbox, with associated helpers
- Make toolbox drag tests work in RTL
- Lint

### Test Coverage

The toolbox drag tests open every category in the basic toolbox and test blocks toolbox, and drag out every (non-disabled) block.
**These tests take a long time (>100 seconds for the test blocks ones) so I have set them to be skipped right now.**
The verification step simply checks whether there is a single top block on the workspace after the drag. As written, these tests will tell us about some catastrophic failures but won't guarantee that everything about block initialization works. Having a human watch the tests, with `PAUSE_TIME` set to more than 50 ms, is the best way to actually verify that the blocks that are created look right. 

In particular, there is a block in the test blocks that has an invalid image URL. Blockly drops an error into the console for that, but the tests don't notice it. Ideally the tests would also monitor the browser console and forward any issues there.

### Additional Information

The helpers I added make some strong assumptions about the state of the workspace during the test. I documented the assumptions but did not try to make the helpers more general-purpose. We can generalize them if a specific need comes up.

### DIFF
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -46,24 +46,13 @@ suite('Testing Connecting Blocks', function () {
     this.browser = await testSetup(testFileLocations.PLAYGROUND);
   });
 
-  test('Testing Block Flyout', async function () {
-    const logicButton = await this.browser.$('#blockly-0');
-    logicButton.click();
-    const ifDoBlock = await this.browser.$(
-      '#blocklyDiv > div > svg:nth-child(7) > g > g.blocklyBlockCanvas > g:nth-child(3)',
-    );
-    await ifDoBlock.dragAndDrop({x: 20, y: 20});
-    await this.browser.pause(200);
-    const blockOnWorkspace = await this.browser.execute(() => {
-      const newBlock = Blockly.getMainWorkspace().getAllBlocks(false)[0];
-      if (newBlock.id) {
-        return true;
-      } else {
-        return false;
-      }
+  test('dragging a block from the flyout', async function () {
+    await dragBlockTypeFromFlyout(this.browser, "Logic", "controls_if", 20, 20);
+    const blockCount = await this.browser.execute(() => {
+      return Blockly.getMainWorkspace().getAllBlocks(false).length;
     });
 
-    chai.assert.isTrue(blockOnWorkspace);
+    chai.assert.equal(blockCount, 1);
   });
 });
 
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -46,7 +46,7 @@ suite('Testing Connecting Blocks', function () {
     this.browser = await testSetup(testFileLocations.PLAYGROUND);
   });
 
-  test('dragging a block from the flyout', async function () {
+  test('dragging a block from the flyout results in a block on the workspace', async function () {
     await dragBlockTypeFromFlyout(this.browser, 'Logic', 'controls_if', 20, 20);
     const blockCount = await this.browser.execute(() => {
       return Blockly.getMainWorkspace().getAllBlocks(false).length;
--- a/tests/browser/test/basic_playground_test.js
+++ b/tests/browser/test/basic_playground_test.js
@@ -47,7 +47,7 @@ suite('Testing Connecting Blocks', function () {
   });
 
   test('dragging a block from the flyout', async function () {
-    await dragBlockTypeFromFlyout(this.browser, "Logic", "controls_if", 20, 20);
+    await dragBlockTypeFromFlyout(this.browser, 'Logic', 'controls_if', 20, 20);
     const blockCount = await this.browser.execute(() => {
       return Blockly.getMainWorkspace().getAllBlocks(false).length;
     });
--- a/tests/browser/test/toolbox_drag_test.js
+++ b/tests/browser/test/toolbox_drag_test.js
@@ -145,7 +145,7 @@ async function openCategories(browser, categoryList, directionMultiplier) {
           await browser.keys(['\uE00C']);
           await browser.pause(PAUSE_TIME);
         } else {
-          let flyoutBlock = await browser.$(
+          const flyoutBlock = await browser.$(
             `.blocklyFlyout .blocklyBlockCanvas > g:nth-child(${3 + i * 2})`,
           );
           if (!(await elementInBounds(browser, flyoutBlock))) {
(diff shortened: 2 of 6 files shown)


tests/browser/test/toolbox_drag_test.js
```diff
@@ -0,0 +1,202 @@
...
+    const blockCount = await getBlockCount(browser, categoryName);
+
+    try {
+      for (let i = 0; i < blockCount; i++) {
+        const category = await getCategory(browser, categoryName);
+        await category.click();
+        if (await isBlockDisabled(browser, i)) {
+          // Unicode escape to close flyout.
+          await browser.keys(['\uE00C']);
+          await browser.pause(PAUSE_TIME);
```
BeksOmega, line 146: Can you use a `continue` here instead of wrapping the rest of the method in an `else`?

tests/browser/test/toolbox_drag_test.js
```diff
@@ -0,0 +1,202 @@
...
+          const topBlockCount = await browser.execute(() => {
+            return Blockly.getMainWorkspace().getTopBlocks(false).length;
+          });
+
+          if (topBlockCount != 1) {
+            failureCount++;
+            console.log(`fail: block ${i} in category ${categoryName}`);
+          }
+
+          // Clean up between blocks so they can't interact with each other
```
BeksOmega, line 167: Nit: End comment with a period.

(I know there are other places in this PR / the tests that don't conform. But for new stuff might as well)

BeksOmega (approved):
This generally LGTM! There are still lots of things about these tests I'd like to clean up :P but I don't want to block this on those! Feel free to ignore any nitpicky things about formatting/naming.

I am pretty confused about why the tests are skipped though. Is it just while we're building out the tests? I don't think there's anything wrong with them taking a long time, especially since we're only running them nightly.

tests/browser/test/basic_playground_test.js
```diff
@@ -46,24 +46,13 @@ suite('Testing Connecting Blocks', function () {
...
-    await ifDoBlock.dragAndDrop({x: 20, y: 20});
-    await this.browser.pause(200);
-    const blockOnWorkspace = await this.browser.execute(() => {
-      const newBlock = Blockly.getMainWorkspace().getAllBlocks(false)[0];
-      if (newBlock.id) {
-        return true;
-      } else {
-        return false;
-      }
+  test('dragging a block from the flyout', async function () {
```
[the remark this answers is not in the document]
rachel-fenichel: I like this as the basic "do any drags work" before getting into the others, because this is trivial to run fast. I don't have strong feelings on it though.

tests/browser/test/toolbox_drag_test.js
rachel-fenichel: I like `else` more than `continue` :)

tests/browser/test/toolbox_drag_test.js
rachel-fenichel: Done!

rachel-fenichel:
> I am pretty confused about why the tests are skipped though. Is it just while we're building out the tests? I don't think there's anything wrong with them taking a long time, especially since we're only running them nightly.

Yes, for while we're building out the tests. I think our teammates will quickly tire of running the tests locally if they always have to sit through how long these tests take. When we're not actively developing them + have sorted out some of the flakiness, I'll turn these on to run nightly.

rachel-fenichel:
@BeksOmega please feel free to leave more comments here telling me what else you would like to see cleaned up, so I can get to them on my return.

tests/browser/test/toolbox_drag_test.js
BeksOmega: I think for long term code health it's better to keep conditional blocks short: https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

For smaller things I can see how `else` is clearer. But in this case (especially since it is already wrapped in a `try...catch`) I would really prefer to de-dent it.

PR #7349, opened 2023-08-01
fix: bumping copied objects

BeksOmega:
## The basics

- [ ] I branched from develop
- [ ] My pull request is against develop
- [X] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [X] I ran `npm run format` and `npm run lint`

### Resolves

Fixes N/A

### Proposed Changes

Adds logic to bump pasted blocks out of the way of other blocks.
Adds logic to bump pasted comments out of the way of their original comment.

### Reason for Changes

Pasting blocks directly on top of other blocks or within the snap radius of other connections is confusing to the end user.

### Test Coverage

Added a test to make sure block create events are being fired.
Added tests to make sure the bumping is working correctly.

### Documentation

N/A

### Additional Information

Dependent on #7348

PR #7352, opened 2023-08-02
chore: upgrade keyboard shortcuts and context menus to use non-deprecated APIs

### LINKED ISSUE #7338: Refactor systems that use copy paste to use the new interfaces
Many systems (e.g. keyboard shortcuts and clipboard commands) within Blockly interact with the copy paste system using the old `ICopyable` interface that was less generic. These systems need to be refactored to conform to the new interfaces.

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes #7338 

### Proposed Changes

Upgrades keyboard shortcuts and context menu items to use `toCopyData` instead of `copy` and `duplicate`

### Reason for Changes

Deprecation warnings are sad :/

### Test Coverage

Manually tested that copying, pasting, cutting, and duplicating works for blocks and workspace comments.

Workspace comments only support duplicating.

### Documentation

N/A

### Additional Information

Dependent on #7349

### DIFF
--- a/core/contextmenu_items.ts
+++ b/core/contextmenu_items.ts
@@ -331,9 +331,10 @@ export function registerDuplicate() {
       return 'hidden';
     },
     callback(scope: Scope) {
-      if (scope.block) {
-        clipboard.duplicate(scope.block);
-      }
+      if (!scope.block) return;
+      const data = scope.block.toCopyData();
+      if (!data) return;
+      clipboard.paste(data, scope.block.workspace);
     },
     scopeType: ContextMenuRegistry.ScopeType.BLOCK,
     id: 'blockDuplicate',
--- a/core/shortcut_items.ts
+++ b/core/shortcut_items.ts
@@ -11,7 +11,7 @@ import {BlockSvg} from './block_svg.js';
 import * as clipboard from './clipboard.js';
 import * as common from './common.js';
 import {Gesture} from './gesture.js';
-import {isCopyable} from './interfaces/i_copyable.js';
+import {ICopyData, isCopyable} from './interfaces/i_copyable.js';
 import {KeyboardShortcut, ShortcutRegistry} from './shortcut_registry.js';
 import {KeyCodes} from './utils/keycodes.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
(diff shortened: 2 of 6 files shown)


core/contextmenu_items.ts
```diff
@@ -331,9 +331,10 @@ export function registerDuplicate() {
       return 'hidden';
     },
     callback(scope: Scope) {
-      if (scope.block) {
-        clipboard.duplicate(scope.block);
-      }
+      if (!scope.block) return;
+      const data = scope.block.toCopyData();
```
maribethb, line 335: small suggestion: `const data = scope.block?.toCopyData()` just to minimize the number of early returns. assuming typescript is smart enough to figure out that `scope.block` must still be defined in the paste call.

core/shortcut_items.ts
```diff
@@ -81,6 +81,9 @@ export function registerDelete() {
   ShortcutRegistry.registry.register(deleteShortcut);
 }
 
+let copyData: ICopyData | null = null;
```
maribethb, line 84: If you store the data here, would this mean you can't copy a block with the context menu and paste it using a keyboard shortcut? if so that would not be desirable and would be a good reason to leave the state in the clipboard module.

core/shortcut_items.ts
```diff
@@ -104,20 +107,18 @@ export function registerCopy() {
         !Gesture.inProgress() &&
         selected != null &&
         selected.isDeletable() &&
-        selected.isMovable()
+        selected.isMovable() &&
+        isCopyable(selected)
       );
     },
     callback(workspace, e) {
-      // Prevent the default copy behavior, which may beep or otherwise indicate
```
maribethb, line 111: I think this is actually helpful for context, since we had to add this to solve a specific bug (iirc, mac beeping at you when you used cmd+c). any reason you removed it?

core/shortcut_items.ts
BeksOmega: You can't copy blocks via the context menu, only duplicate them! And `copy` is internal, so external devs *shouldn't* have added a `copy` option to the context menu either.

core/contextmenu_items.ts
BeksOmega: So the `?.` doesn't work in this context :/

The `?.` operator short circuits an expression to evaluate to `undefined` instead of throwing a `TypeError` because you're trying to read a property of `undefined` or `null`. It doesn't actually assert that `scope.block` is not `undefined` or `null`.

core/shortcut_items.ts
BeksOmega: Nope will revert!

core/shortcut_items.ts
maribethb: This is surprising yet logical...

PR #7355, opened 2023-08-03
fix: have rendered blocks render themselves on construction

### LINKED ISSUE #5146: Refactor block and blocksvg initialization

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #5146

### Proposed Changes

Makes it so that BlockSvgs automatically render themselves when they are constructed.

### Reason for Changes

Before this change external devs needed to call `initSvg` and `render` on their blocks if they were instantiating them via `workspace.newBlock`. But I deprecated `render` recently. This makes it so calling `render` is no longer necessary.

### Test Coverage

Tested that loading the spaghetti code, and instantiating via `newBlock` both worked as expected.

### Documentation

N/A

### Additional Information

I decided to have these queue the render instead of rendering immediately. External devs calling `render` currently will trigger the immediate render, so this is non-breaking. And this way they can properly switch to using `finishQueuedRenders`.

### DIFF
--- a/blocks/lists.ts
+++ b/blocks/lists.ts
@@ -187,13 +187,11 @@ const LISTS_CREATE_WITH = {
     const containerBlock = workspace.newBlock(
       'lists_create_with_container',
     ) as ContainerBlock;
-    (containerBlock as BlockSvg).initSvg();
     let connection = containerBlock.getInput('STACK')!.connection;
     for (let i = 0; i < this.itemCount_; i++) {
       const itemBlock = workspace.newBlock(
         'lists_create_with_item',
       ) as ItemBlock;
-      (itemBlock as BlockSvg).initSvg();
       if (!itemBlock.previousConnection) {
         throw new Error('itemBlock has no previousConnection');
       }
--- a/blocks/logic.ts
+++ b/blocks/logic.ts
@@ -375,17 +375,14 @@ const CONTROLS_IF_MUTATOR_MIXIN = {
    */
   decompose: function (this: IfBlock, workspace: Workspace): ContainerBlock {
     const containerBlock = workspace.newBlock('controls_if_if');
-    (containerBlock as BlockSvg).initSvg();
     let connection = containerBlock.nextConnection!;
     for (let i = 1; i <= this.elseifCount_; i++) {
       const elseifBlock = workspace.newBlock('controls_if_elseif');
-      (elseifBlock as BlockSvg).initSvg();
       connection.connect(elseifBlock.previousConnection!);
       connection = elseifBlock.nextConnection!;
     }
     if (this.elseCount_) {
       const elseBlock = workspace.newBlock('controls_if_else');
-      (elseBlock as BlockSvg).initSvg();
       connection.connect(elseBlock.previousConnection!);
     }
     return containerBlock;
--- a/blocks/text.ts
+++ b/blocks/text.ts
@@ -792,13 +792,11 @@ const JOIN_MUTATOR_MIXIN = {
     const containerBlock = workspace.newBlock(
       'text_create_join_container',
     ) as BlockSvg;
-    containerBlock.initSvg();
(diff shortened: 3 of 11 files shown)


maribethb (commented):
This change itself seems fine, but it seems like #5146 still has more work to be done but this PR says it closes it. For example, you did not remove the TODO comment on the `initBlock` function in `serialization/blocks`. Which makes sense, because in #5146 you said you want to refactor the serialization system so that it does not need to do checking based on whether the block is rendered or not. Are you still planning on addressing that? If not could you remove the todo comment and if so could you update this PR so it doesn't say it's closing that issue?

Also, there are *a lot* of places in core where we still call `initSvg` and then `render` / `queueRender`. Are you planning to update those as well? If not in this PR, could you file an issue for that if one doesn't exist?

BeksOmega:
> This change itself seems fine, but it seems like #5146 still has more work to be done...

Ah yeah you're right, I should have marked this as "work on", not closing. I'm planning to punt on fully cleaning up `initBlock` because I don't fully remember why the `updateDisabled` or `onLocationChange` calls are necessary.

> Also, there are _a lot_ of places in core where we still call `initSvg` and then `render` / `queueRender`. Are you planning to update those as well? If not in this PR, could you file an issue for that if one doesn't exist?

Gosh darn it :P I was trying to do too many things at once yesterday. I checked for references on devsite but not other places in core haha.

So still need to:

- [ ] Clean up references to `initSvg` and then `queueRender` in core.
- [ ] Clean up references to `initSvg` and then `render` in other parts of this repo.
- [ ] Clean up references to `initSvg` and then `render` in samples.

BeksOmega:
This ran into problems because having the constructor call `initSvg` and `queueRender` means we can end up in a state where `initSvg` has been called but `rendered` is false. Our life cycle does not expect this to happen and it causes bugs.

So we're going to refactor `rendered` first and then make this change.

That means that for v10.1 I'm going to open up a different PR dedeprecating `render`.

PR #7356, opened 2023-08-03
deprecate: workspace paste

BeksOmega:
## The basics

- [ ] I branched from develop

### Resolves

Fixes N/A

### Proposed Changes

Deprecates the `paste` function on the workspace.

### Reason for Changes

This function is not maintainable. It is hard coded to deserialize block data and workspace comment data. It cannot be backwards compatibily updated to use the new paster system, because it doesn't takes in the save state instead of the `CopyData` directly.

I see no reason to maintain this when we have better functionality in `clipboard.paste`.

For now it will continue to work as it did before (no breaking changes). But it will be deleted in v12.

### Test Coverage

N/A This is just a deprecation :P

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -1295,6 +1295,12 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
   paste(
     state: AnyDuringMigration | Element | DocumentFragment,
   ): ICopyable<ICopyData> | null {
+    deprecation.warn(
+      'Blockly.WorkspaceSvg.prototype.paste',
+      'v11',
+      'v12',
+      'Blockly.clipboard.paste',
+    );
     if (!this.rendered || (!state['type'] && !state['tagName'])) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


core/workspace_svg.ts
```diff
@@ -1382,10 +1391,6 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
             for (let i = 0, connection; (connection = connections[i]); i++) {
               const neighbour = connection.closest(
                 config.snapRadius,
-                // TODO: This code doesn't work because it's passing an absolute
-                //     coordinate instead of a relative coordinate. Need to
-                //     figure out if I'm deprecating this function or if I
-                //     need to fix this.
```
cpcallen, line 1388: Any particular reason to delete these comments (here and blelow)?  Deciding to deprecate has (presumably) not caused the code to become correct…

cpcallen (approved):

core/workspace_svg.ts
BeksOmega: True! The TODO was to decide to fix it or not, and I decided not to so I deleted the TODO :P But I can leave a comment that the code is incorrect.

cpcallen: Good point re "TODO", but yes let's keep documentation of bugs until the buggy code is actually deleted.

PR #7364, opened 2023-08-08
feat: Insertion marker json deserialization 7316

### LINKED ISSUE #7316: Change insertion markers to use JSON deserialization when duplicating blocks
### Check for duplicates

### Problem

Insertion markers duplicate the block being dragged, and then tell the duplicate to be an insertion marker.

Currently the duplication duplicates (ha) a lot of code from the serialization system. We think this is because in the past the XML system didn't have a good way to just serialize and deserialize a single block (without children). But the JSON system does have this!

### Request

Change the insertion marker `createMarkerBlock` code over to use JSON serialization and deserialization.

### Alternatives considered

_No response_

### Additional context

Sparked by this discussion: https://groups.google.com/g/blockly/c/gcvynl88QLY/m/shFyX4x0BwAJ

varshneydevansh:
## The basics

- [x] I branched from develop

### Resolves

- Fixes #7316 

### Proposed Changes

Change the insertion marker `createMarkerBlock` code over to use **JSON serialization and deserialization**.

#### Behavior Before Change

In Blockly's XML serialization, child blocks are represented using nested XML elements. This means that if a block has child blocks connected to it, those child blocks will be represented as nested XML elements within the XML element representing the parent block. The structure of the XML serialization mirrors the structure of the blocks in the workspace.

#### Behavior After Change

The JSON serialization system in Blockly does something similar, but with a JSON structure instead of XML. The structure of the serialized data, whether XML or JSON, is designed to capture the hierarchical relationships between blocks and their connections.

### Reason for Changes

Insertion markers duplicate the block being dragged, and then tell the duplicate to be an insertion marker.

Currently the duplication duplicates (ha) a lot of code from the serialization system. We think this is because in the past the XML system didn't have a good way to just serialize and deserialize a single block (without children). But the JSON system does have this!

- Sparked by this discussion: https://groups.google.com/g/blockly/c/gcvynl88QLY/m/shFyX4x0BwAJ

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -232,49 +232,19 @@ export class InsertionMarkerManager {
    * @returns The insertion marker that represents the given block.
    */
   private createMarkerBlock(sourceBlock: BlockSvg): BlockSvg {
-    const imType = sourceBlock.type;
-
     eventUtils.disable();
     let result: BlockSvg;
     try {
-      result = this.workspace.newBlock(imType);
+      // 1. Serialize the source block to JSON
+      const blockJson = Blockly.serialization.blocks.save(sourceBlock);
+
+      // 2. Deserialize the JSON back to a block in the same workspace
+      result = Blockly.serialization.blocks.append(blockJson, this.workspace);
+
+      // Setting the result as an insertion marker
       result.setInsertionMarker(true);
-      if (sourceBlock.saveExtraState) {
-        const state = sourceBlock.saveExtraState();
-        if (state && result.loadExtraState) {
-          result.loadExtraState(state);
-        }
-      } else if (sourceBlock.mutationToDom) {
-        const oldMutationDom = sourceBlock.mutationToDom();
-        if (oldMutationDom && result.domToMutation) {
-          result.domToMutation(oldMutationDom);
-        }
-      }
-      // Copy field values from the other block.  These values may impact the
-      // rendered size of the insertion marker.  Note that we do not care about
-      // child blocks here.
-      for (let i = 0; i < sourceBlock.inputList.length; i++) {
-        const sourceInput = sourceBlock.inputList[i];
-        if (sourceInput.name === constants.COLLAPSED_INPUT_NAME) {
-          continue; // Ignore the collapsed input.
-        }
-        const resultInput = result.inputList[i];
-        if (!resultInput) {
-          throw new Error(DUPLICATE_BLOCK_ERROR.replace('%1', 'an input'));
-        }
-        for (let j = 0; j < sourceInput.fieldRow.length; j++) {
-          const sourceField = sourceInput.fieldRow[j];
-          const resultField = resultInput.fieldRow[j];
-          if (!resultField) {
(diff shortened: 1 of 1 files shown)


varshneydevansh:
The error - **Did you forget to signal async completion?**

is because am I missing fields with the modified serialization/de-serialization, or maybe I have to review the dependencies. Furthermore, I have not yet conducted any testing on this, as I am seeking confirmation whether these are the potential modifications that we intended.

core/insertion_marker_manager.ts
```diff
@@ -232,49 +221,19 @@ export class InsertionMarkerManager {
...
-          const sourceField = sourceInput.fieldRow[j];
-          const resultField = resultInput.fieldRow[j];
-          if (!resultField) {
-            throw new Error(DUPLICATE_BLOCK_ERROR.replace('%1', 'a field'));
-          }
-          resultField.setValue(sourceField.getValue());
-        }
-      }
+      // 1. Serialize the source block to JSON
+      const blockJson = Blockly.serialization.blocks.save(sourceBlock);
```
BeksOmega, line 228: When you're working in core you need to import the individual module instead of referencing the fully qualified name! So import `blocks` from `serialization/blocks.js`, and then call `blocks.save` (same for `append`)!

That should fix the build error you're getting =)

varshneydevansh: Got it. Making the changes.

varshneydevansh: I imported like this -`import {save, append} from '../core/serialization/blocks'`

But getting an error like this - 

image

image

image

However, if I am doing this, apparently the errors are also going away - 

image

varshneydevansh: I have also removed these two lines

```ts
      result.setCollapsed(sourceBlock.isCollapsed());
      result.setInputsInline(sourceBlock.getInputsInline());
      
```
 As the serialization/deserialization methods handle the collapsed state and inline inputs internally and manually setting them again would be redundant.

core/insertion_marker_manager.ts
```diff
@@ -232,49 +222,21 @@ export class InsertionMarkerManager {
...
       }
+      // 2. Deserialize the JSON back to a block in the same workspace
+      result = blocks.append(blockJson, this.workspace) as BlockSvg;
 
-      result.setCollapsed(sourceBlock.isCollapsed());
-      result.setInputsInline(sourceBlock.getInputsInline());
+      // Setting the result as an insertion marker
+      result.setInsertionMarker(true);
 
+      // Initialize the SVG for rendering
```
BeksOmega, line 239: Can you remove all of the comments here? I think the code is pretty self-explanatory!

BeksOmega (changes requested):
This looks great! Just one style change and then should be good to go =)

core/insertion_marker_manager.ts
varshneydevansh: done

varshneydevansh:
> This looks great! Just one style change and then should be good to go =)

I also wanted to know if we need to make any changes to the documents?

PR #7333, opened 2023-07-29
fix: removed X & Y from toolbox.ts and replaced movBy to moveTo

varshneydevansh:
## The basics

- [x] I branched from develop
- [x] My pull request is against develop
- [x] My code follows the [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide)
- [x] I ran `npm run format` and `npm run lint`

### Resolves

- Fixes #6280 

### Proposed Changes

The [BlockInfo type](https://github.com/google/blockly/blob/fa58cbaaf7456ab227c2214af5dd0e7025dc731c/core/utils/toolbox.ts#L19) should be updated to not include X and Y.
Blockly has two kinds of flyouts built-in, horizontal and vertical. You should update the layout logic in both of them.
The flyouts should be changed to use [moveTo](https://github.com/google/blockly/blob/fa58cbaaf7456ab227c2214af5dd0e7025dc731c/core/block_svg.ts#L414) [instead of moveBy](https://github.com/google/blockly/blob/61322294da70c770b32abb28b9d695af75f61a16/core/flyout_vertical.js#L250)

#### Behavior Before Change

The `BlockInfo type` was picking up the values of `x` and `y` henceforth the position of the Toolbox Element was getting changed.

#### Behavior After Change

After removing the `X` and `Y` from the `BlockInfo type` the Toolbox Element is now ignoring the value and introducing the chnage to the code in files 
- [core/flyout_horizontal.ts](https://github.com/google/blockly/pull/7333/files#diff-67129394789d0d6086c6c255000bb6db6c9a4ecfa66a3e3f8618a64b0f804d40)
- [core/flyout_vertical.ts](https://github.com/google/blockly/pull/7333/files#diff-b866da6441c5a41965db65339b45aaffa9036332a928bde9ba9e5f59823b14a3)

to use `moveTo` instead of `moveBy`

Ignoring the x and y properties on blocks would allow them to position correctly.

### Reason for Changes

Because, the blocks were using the `relative coordinates` instead of the `absolute coordinates`

### Test Coverage

Browser which I used is MS Edge Version 115.0.1901.188 (Official build) (64-bit)

### Additional Information

It's my first PR on this project, so I might make mistakes.

Issue #6858, opened 2023-02-21
Remove Closure library and module system from the Blockly codebase?


cpcallen:
This is a discussion + tracking bug for work to remove our remaining dependencies on the Closure library (of which `closure/goog/base.js` and `…/goog.js` are the only remaining pieces) and the Closure module system (`goog.module` / `goog.require`).

The existence of this bug does not denote that a decision to remove the Closure library from our codebase; it exists merely to document what would be involved in doing so and track what work is done in that direction, as well as to provide a place to discuss the merits of such work.

### Terminology
There are several related but separate Google products named "Closure":

* The [Closure module system](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide), also known as `goog.module`.
* The [Closure Library](https://developers.google.com/closure/library), which Blockly previously used for various purposes (including use of its UI widgets); most of it was removed prior to 2021 and its use in this repo has since PR #2976 been limited to supporting the closure module system.
* The [Closure Compiler](https://developers.google.com/closure/compiler), which we used as a bundler, chunker and minifier.
* Also [Closure templates](https://github.com/google/closure-templates), which not used in this repo.

### Overview, Motivation and Limitations

Use of the Closure module system ties us to `base.js` for uncompiled-mode loading and to Closure Compiler for bundling.  Since we'd like to try alternative bundlers such as Webpack, it would be good to remove our remaining use of the module system.

Even once we have fully removed use of the Closure module system we _may_ wish to retain use of the debug module loader (in `base.js`) at least in the short term because it can do useful things like load JS scripts (via `<script>` tags) _after_ ES modules have been `import`ed.

We will probably wish to continue to use Closure Compiler at least until an alternative bunding/minification tool has been proven to be a good alternative.  This is likely to entail our continued use of [`closure-calculate-chunks`](https://github.com/ChadKillingsworth/closure-calculate-chunks) which depends on [`closure-make-deps`](https://www.npmjs.com/package/google-closure-deps) which in turn expects the existence of at least a minimal `base.js`.

## Work already completed

(…as of the creation of this bug, late February 2023)

* Almost all of use of the Closure module system (`goog.module` etc.) in `core/` was retired with PR #6299 (see also issue #5857), but we still use it in various places, notably in `blocks/` and `generators/`.  
* We've broken all `import` cycles amongst modules in `core/`.
  * `closure-make-deps` enforces a rule against circular `goog.require`s because the semantics of Closure modules do not allow these.
[...]

PR #7380, opened 2023-08-11
refactor(tests): Introduce loading shims, use in playgrounds

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6858

### Proposed Changes

* Simplify implementation of `posixPath` in `scripts/helpers.js`

* Make safe to import `closure/goog/goog.js` in node.js.
  
  Modify the implementation of `declareModlueId` so that it can be safely (albeit ineffectively) imported and executed by node.js, which does not provide a window global.

* Introduce chunk loading shims:
  
  * Add a `buildShims` task to `build_tasks.js` that, for each chunk, creates a correspondingly-named `build/<chunk>.mjs` that will either (in uncompressed mode) `import` and re`export` that chunk's entry point module (e.g. `core/blockly.js`) or (in compressed mode) load `dist/<chunk>_compressed.js` using a `<script>` tag and then export the corresponding properties on the chunk's exports object.
  
  * Provide helper methods used by these shims in `tests/scripts/loading.mjs`, including code to detect whether to load in compressed or uncompressed mode.
  
  * Add a `quote()` function to `scripts/helpers.js`, used by `buildShims`.  This is copied from `tests/bootstrap_helper.js`, which will be removed in a later commit.

* Update `playground.html`, `advanced_playground.html` and `multi_playground.html` to use the new loading shims.

* Delete `playgrounds/shared_procedures.html`: shared procedure support was moved to a plugin and this should have been removed from core along with it.

#### Behaviour Before Change

The playgrounds used `tests/bootstrap.js` et al. to load Blockly and other needed scripts, ultimately depending on the debug module loader in `closure/goog/base.js` to do the actual loading.

Compressed/uncompressed selection was done by looking for localhost and/or `file:` in the URL, and could be overriden only by having the playground set an option on the `BLOCKLY_BOOTSTRAP_OPTIONS` global options object before loading `bootstrap.js`.

Once loaded, Blockly was made available via the `Blockly` global variable; generators via `javasscriptGenerator` (etc.) globals.

#### Behaviour After Change

The playgrounds can `import * as Blockly from '…/build/blockly.mjs';` (and similarly for blocks, generators)—almost as if loading it from the `blockly` NPM package.

It is possible to select compressed vs. uncompressed mode by appending a `?compressed=true` (or `false`, or `yes`/`no` or `1`/`0`) query parameter to the URL.

The mocha and generator tests still use `bootstrap.js` for now; they will be converted (and the bootstrap code removed) in a future PR.

### Reason for Changes

* Reduce dependency on Closure debug module loader.
* Make importing of Blockly in playgrounds more consistent with how this is done in `blockly-samples`.
* Maintain ability to choose compressed/uncompressed mode, with a more convenient way to override default choice.

### Test Coverage

Unchanged.  Playgrounds tested manually (resulting in incidental discovery of issue leading to revert of PR #7333 by PR #7375 and swift update of Blockly v10.1.0 with v10.1.1).

### Documentation

The `?compressed=` parameter should probably be documented somewhere, but I'm not sure where the best place would be.

### Additional Information

I  considered implementing the compressed/uncompressed switching by using [import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) to choose between two different modules:

* In compressed mode, the `'blockly'` module (e.g. `import … from 'blockly';`) would be mapped onto an ESM much like the shims created here, that would load the `<chunk>_compressed.js` and reexport the properties on its exports object, but slightly simplified by not have any support for loading in uncompressed mode.
* In uncompressed mode, `'blockly'` would be mapped directly to `build/src/core/blockly.js`.

However there are two reasons I did not go down that route:

* It would entail changing import maps based on `window.location`, and it wasn't completely clear how easy that would be to do.  Certainly it would entail having an initial `<script>` run to set up the correct import map before a `<script type="module">` could do the actual `import`s, whereas the approach chosen hides all the choosing logic in the shims themselves (or rather in `load.mjs`, which they all depend on).
* Import maps are a pretty new feature, [only recently supported by all the major browsers and not yet supported by all](https://caniuse.com/import-maps).

It may well make sense in a year or two to add a (static) import map so that `import * as Blockly from './path/to/build/blockly.mjs'` can be simplified to `import * as Blockly from 'blockly'`, while leaving the shims as they are.


tests/scripts/load.mjs
```diff
@@ -0,0 +1,140 @@
...
+export const ROOT = window.location.href.replace(/\/tests\/.*$/, '/');
+
+/**
+ * Decide whether to use compressed mode or not.
+ *
+ * Honours a "?compressed=" query parameter if present; otherwise uses
+ * uncompressed for when loading from local machine and compressed
+ * otherwise.  See issue #5557 for additional background.
+ *
+ * @return {boolean} True if shoudl load in compressed mode.
```
BeksOmega, line 46:
```suggestion
 * @return {boolean} True if should load in compressed mode.
```

tests/scripts/load.mjs
```diff
@@ -0,0 +1,140 @@
...
+/** @type {boolean} Load in compressed mode. */
+export const COMPRESSED = compressed();
+
+/**
+ * Load a chunk, either by importing its ESM entrypoint or loading the
+ * compressed chunk script.
+ *
+ * When loading in uncompressed mode, if scriptExports is a simple
+ * variable name (e.g. 'Blockly') then globalThis[scriptExports] will
+ * be set to the the chunk's Module object.  This attempts provide
```
BeksOmega, line 71:
```suggestion
 * be set to the the chunk's Module object.  This attempts to provide
```

tests/scripts/load.mjs
```diff
@@ -0,0 +1,140 @@
...
+ * backward compatibility with loading the compressed chunk as a
+ * script, where this is done by the compressed chunk's UMD wrapper.
+ * The compatibility is not complete, however: since Module objects
+ * are immutable, it is not possible to set (e.g.)
+ * Blockly.libraryBlocks.
+ *
+ * The intention is to allow the chunk to be accessed accessed either
+ * the returned Module / exports object (preferred) or via the global
```
BeksOmega, line 79:
```suggestion
 * The intention is to allow the chunk to be accessed via either
 * the returned Module / exports object (preferred) or via the global
```

tests/scripts/load.mjs
```diff
@@ -0,0 +1,140 @@
...
+    const script = document.createElement('script');
+    script.src = src;
+    script.onload = resolve;
+    script.onerror = reject;
+    document.getElementsByTagName('head')[0].appendChild(script);
+  });
+}
+
+/**
+ * Get the value referred to by a doted-itentifier path
```
BeksOmega, line 127:
```suggestion
 * Get the value referred to by a dotted-identifier path
```

scripts/gulpfiles/build_tasks.js
```diff
@@ -672,6 +674,56 @@ function buildCompiled() {
...
+  );
+
+  // Import each entrypoint module, enumerate its exports, and write
+  // a shim to load the chunk either by importing the entrypoint
+  // module or by loading the compiled script.
+  await Promise.all(chunks.map(async (chunk) => {
+    const modulePath = posixPath(chunk.moduleEntry ?? chunk.entry);
+    const scriptPath =
+        path.posix.join(RELEASE_DIR, `${chunk.name}${COMPILED_SUFFIX}.js`);
+    const shimPath = path.join(BUILD_DIR, `${chunk.name}.mjs`);
```
BeksOmega, line 705: How would you feel about changing this to `${chunk.name}.loader.mjs` to make it clearer within the playgrounds that we're importing a shim?

On one hand I like how the imports in the playgrounds are very similar to our imports in samples. But on the other hand, I feel like the current naming hides the fact that we actually are using a different thing that's doing different work.

BeksOmega (approved):
Mostly just typos :P and then one actual question!

I feel like this system fits in my head so much better than whatever the heck the debug module loader stuff is doing :P so happy!!

tests/scripts/load.mjs
cpcallen: I can dote on my identifier-paths if I want!

scripts/gulpfiles/build_tasks.js
cpcallen: That seems reasonable enough.

Issue #7326, opened 2023-07-24
Provide more user-friendly generator APIs


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

There are a bunch of methods and properties within the generator code that:
  1) Are protected, when they need to be public so they can be used by block-code generators.
  2) Have confusing names.

The following things are protected but used by generators:
  * quote_ (used by string block)
  * multiline_quote_ (used by multiline string block)
  * nameDB_ (also addressed in https://github.com/google/blockly/issues/6008)
  * definitions_ (we just cludge imports and user-defined procedures onto this object, you need to know its naming semantics to do this correctly)
  * scrub_ (needs to be called by procedure definition blocks to handle comments)

The `provideFunction_` method (which is only supposed to be used for developer-defined functions) is public, but has a sad underscore in the name :/

### Request

Provide an API that matches what external developers are actually trying to do.

I think it should look something like:
  * getVariableName
  * getProcedureName
  * quote
  * multilineQuote
  * addProcedure -> This should handle scrubbing and finding a name
  * addImport
  * addDeveloperFunction

### Alternatives considered

Make the protected methods public so they're actually usable. But stick with the less-than-ideal API.

### Additional context

Related to https://github.com/google/blockly/issues/6008 (should possibly supercede it)

maribethb:
It is probably worth adding some new functions (e.g. we already agreed on `getVariableName` and `getProcedureName`, and `addImport` or `addProcedure` might also make sense, to be discussed when we get to this change) but let's not rename the functions just to remove underscores. It creates a lot of churn for not a lot of benefit. They should be made public though.

cpcallen:
> let's not rename the functions just to remove underscores. It creates a lot of churn for not a lot of benefit.

If we're changing a bunch of other things in the API then I think we should provide (and document) non-underscored names—but there's no reason to cause churn: the existing underscored names can be kept indefinitely as aliases/wrappers.

maribethb:
Rachel & I discussed further and still think it's not worth the additional overhead and churn of adding aliases:
- It would add confusion when developers use blocks that contain a mix of the two styles
- Either we have dead code around forever, or we deprecate the old ones eventually and cause the unnecessary churn we want to avoid
- Sets a precedent for what we might do in the rest of Blockly, or else is confusing why we are only adding this special treatment to the `CodeGenerator` class.
[...]

PR #7392, opened 2023-08-14
feat: make generator quoting methods public

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #7326 

### Proposed Changes

Makes the `quote_` and `multiline_quote_` methods public.

### Reason for Changes

`quote_` and `multiline_quote_` were previously inaccessible to external developers. They can now be accessed inside block-code generators just like they are in core.

### Test Coverage

N/A

### Documentation

This is for some documentation :D

### Additional Information

N/A

### DIFF
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -37,7 +37,7 @@ export type BlockGenerator = (
 /**
  * Class for a code generator that translates the blocks into a language.
  */
-export abstract class CodeGenerator {
+export class CodeGenerator {
   name_: string;
 
   /** A dictionary of block generator functions keyed by block type. */
@@ -436,10 +436,6 @@ export abstract class CodeGenerator {
     const id = block.id.replace(/\$/g, '$$$$'); // Issue 251.
     return msg.replace(/%1/g, "'" + id + "'");
   }
-
-  abstract quote_(toQuote: string): string;
-
-  abstract multiline_quote_(toQuote: string): string;
 
   /**
    * Add one or more words to the list of reserved words for this language.
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -37,7 +37,7 @@ export type BlockGenerator = (
 /**
  * Class for a code generator that translates the blocks into a language.
  */
-export class CodeGenerator {
+export abstract class CodeGenerator {
   name_: string;
 
   /** A dictionary of block generator functions keyed by block type. */
@@ -169,10 +169,6 @@ export class CodeGenerator {
     codeString = codeString.replace(/[ \t]+\n/g, '\n');
     return codeString;
   }
-
-  // The following are some helpful functions which can be used by multiple
-
-  // languages.
 
   /**
    * Prepend a common prefix onto each line of code.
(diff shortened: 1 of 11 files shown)


core/generator.ts
```diff
@@ -37,7 +37,7 @@ export type BlockGenerator = (
 /**
  * Class for a code generator that translates the blocks into a language.
  */
-export class CodeGenerator {
+export abstract class CodeGenerator {
```
cpcallen, line 40: This is definitely a breaking change, and in particular one that makes the current state of the custom code generator codelab invalid (see google/blockly-samples#1858) as it creates an instance of `CodeGenerator` directly.  (And same for any developer-written custom code generators not updated with the changes announced at this year's summit.)

Although I think subclassing `CodeGenerator` is best practice, I don't think we should force developers to do so.

core/generator.ts
```diff
@@ -441,6 +437,10 @@ export class CodeGenerator {
     return msg.replace(/%1/g, "'" + id + "'");
   }
 
+  abstract quote_(toQuote: string): string;
```
cpcallen, line 440: If we can't make this `abstract`, could we instead a default implementation that will work adequately in most languages?  I note that at the moment `JavascriptGenerator`, `LuaGenerator` and `PhpGenerator` all use the same implementation (though it is wrong for at least JavaScript 🙁).

core/generator.ts
```diff
@@ -441,6 +437,10 @@ export class CodeGenerator {
...
+
+  abstract multiline_quote_(toQuote: string): string;
```
cpcallen, line 442: Admittedly this one is harder to provide a generic implementation for.  It could just default to using `quote` (and we could fix `quote` to handle newlines correctly), losing the nice formatting.

cpcallen (commented):
See also [my comment about underscores](https://github.com/google/blockly/issues/7326#issuecomment-1678157561).

core/generator.ts
BeksOmega: The `abstract` keyword only affects people using typescript right? They shouldn't be cludging on properties anyway, so I don't think this should affect them. (e.g. assigning directly to `scrub_` wouldn't work in typescript because it's protected)

I think we should force people to subclass `CodeGenerator` in the future, because it actually isn't useable on its own. At least as long as it's not breaking peeps more than we are anyway by (eventually) actually exporting types for our generators.

I may be misunderstanding what `abstract` does though, or the current state of how people are using `CodeGenerator` in the wild.

BeksOmega:
@cpcallen This is ready for rereview =)

core/generator.ts
maribethb: Discussed with Rachel and agreed that if you also fix the codelab and any documentation, it's reasonable to make this `abstract`. Since there's no way for typescript developers to be using this with correct types anyway, we also don't need to do a major version bump. We should call this out in release notes though / make sure the PR description explains the change.

I didn't discuss this with Rachel though so you can push back if you disagree, but I do think it might make sense to hold that particular change until we publish usable types for the generators we provide. That way TS devs can make all of their necessary changes at once.

BeksOmega: Filed an issue for this (https://github.com/google/blockly/issues/7401). Didn't mark it as `triage` since you already discussed. Linked to https://github.com/google/blockly/issues/6828

rachel-fenichel (approved):

rachel-fenichel:
> Makes CodeGenerator abstract.

It looks like this is no longer true, so please remove from the description before merging.

Issue #7157, opened 2023-06-13
Consider whether to provide accessors for methods relevant to the scroll options plugin


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

The scroll options plugin accesses some public + underscore properties of the workspace. We should consider whether to add public accessors for these, since there is obviously a use case for accessing them.

The first property is `svgGroup_`.

The second property is `svgBlockCanvas_`. This /is/ currently returned by the accessor `getBlockCanvas`, but `getBlockCanvas` is marked `@internal`, so it is inaccessible.

### Request

Consider adding a `getSvgGroup` method and making the `getBlockCanvas` method public.

### Alternatives considered

N/A

### Additional context

https://github.com/google/blockly-samples/pull/1666

maribethb:
Let's do it

gonfunko:
Closing as there is now a public `WorkspaceSvg.getSvgGroup()`, and, while `WorkspaceSvg.getBlockCanvas()` is still `@internal`, it just calls and returns `WorkspaceSvg.getCanvas()`, which is public.

PR #7394, opened 2023-08-14
feat: add accessor for the workspace svg group

BeksOmega:
## The basics

- [X] I branched from develop

### Resolves

Work on #7157 

### Proposed Changes

Adds public accessor for `svgGroup_`.

### Reason for Changes

This is necessary for the scroll options plugin to add a wheel listener, and for the multiselect plugin to add a mouse enter listener.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

I didn't make `getBlockCanvas` public because I realized the scroll options plugin is only using it to get the transform. I think we should either be able to use the metrics, or just `scrollX` and `scrollY` to do that instead.

App Inventor asked that I fix the `svgGroup_` issue this quarter so they don't need to worry about the multiselect plugin breaking (in this way) in the future.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -694,6 +694,15 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
   }
 
   /**
+   * Returns the SVG group for the workspace.
+   *
+   * @returns The SVG group for the workspace.
+   */
+  getSvgGroup(): Element {
+    return this.svgGroup_;
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
> I didn't make getBlockCanvas public because I realized the scroll options plugin is only using it to get the transform. I think we should either be able to use the metrics, or just scrollX and scrollY to do that instead.

Using metrics might work but be very slow--just something to keep an eye out for.

rachel-fenichel (approved):

Issue #7224, opened 2023-06-28
Use Blockly consistently in mocha tests


maribethb:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

In mocha test files, sometimes we use `Blockly.foo` and sometimes use `import {foo} from '../build/.......'; foo`

we shouldn't use both styles; that's just confusing.

Christopher says there are arguments to both. @cpcallen please elaborate in the comments :)

Without hearing those arguments, I would prefer using `Blockly.foo` because ideally the tests would use Blockly the same way users do and users don't import our individual files.

### Reproduction steps

1. 
2. 
3.

cpcallen:
### Background

Tests can ideally be divided into roughly three categories:
1. Unit tests, that test single functions/methods or perhaps single classes.  It is common to stub/mock the dependencies of the code under test in order to isolate it, which helps make the tests less brittle and allows them to execute more quickly.
2. Integration tests, in which multiple units of code (whole/multiple modules, perhaps) are tested to ensure that interactions between them are correct.  Dependencies outside the integration being tested might still be stubbed/mocked, but this is often less necessary since the integration being tested might include most or all of the dependencies.
3. System tests, in which an entire program/library is being tested, including all its dependencies.

In the case of system tests of a software library like Blockly, it behoves one to test against the API that one is shipping.  (If it's also a UI library, then one should additionally also be testing the UI—in our case the DOM elements Blockly creates and manages.)  So any kind of system tests we are doing ought to be against the exports of `core/blockly.ts`, i.e., (historically) `Blockly.*`.

In the case of unit tests, it is generally preferable to test against only the module containing the code being tested.  This enables the unit tests for a single module to be run without having to compile/load any other modules—most test frameworks have the ability to specify exactly what tests should be run, making it convenient to quickly run the tests for a particular module while working on it, without having to repeatedly run the entire test suite.  It also allows testing of interfaces that are not exposed by the public API (which for an app, rather than a library, is basically all of them).

(Aside: to facilitate keeping tests and tested code in sync, unit tests are often kept very close to the code being tested.  The Go convention is to keep the tests for `foobar.go` in `foobar_test.go` in the same directory, with the latter having full namespace access to the former without any explicit, as if it were part of the same file.  In scripting languages like Perl and Python, it is not unusual for libraries to include the unit tests _in the same file_ as the code under test, executed when the file is run as a script rather than imported as a module.)

[...]

PR #7406, opened 2023-08-17
refactor(tests): Use `import` instead of `goog.bootstrap` to load Blockly in mocha tests

cpcallen:
## The basics

- [X] I branched from develop

### Resolves

Part of #6858.

### Proposed Changes

* Have the `buildShims` task clean up after itself.

  We need to create a `build/package.json` file to allow node.js to load `build/src/core/blockly.js` and the other chunk entry points as ES modules (it otherwise forcibly assumes `.js` means CJS even if one is trying to `import` the file rather than `require` it, unless `package.json` says `{"type": "module"}`), but this interferes with `scripts/migration/js2ts` doing a `require('build/deps.js')`, since `deps.js` is _not_ an ES module.

* Temporarily add missing `import`s of `closure/goog/goog.js` to modules in `tests/mocha/` (see de9499b for rationale).

* Migrate `tests/mocha/index.html` to `import` the loader shims instead of using `bootstrap.js`.
  * It now uses `import` to load Blockly, the library blocks and the JavaScript generator via the shims create in PR #7380.  Because the mocha tests have always been run in uncompressed mode it could instead directly import the chunk entry point s (e.g. `build/src/core/blockly.js`), but using the shims allows us to optionally run the tests in compressed mode as well.  At the moment they do not all pass, because several tests directly import individual modules from `core/`, which is not compatible with `blockly_compressed.js`, but #7224 is tracking work that would resolve this problem.
  * It also now uses `import` to load the individual test files that were originally `goog.require()`ed and more recently loaded via `bootstrap.js`.
  * It uses `loadScript` (from `tests/scripts/load.mjs`) to load various additional scripts including `en.js`.

* Remove code from `build_tasks.js` that generates `tests/deps.mocha.js`, since it was only needed by the bootstrap code now deleted from `tests/mocha/index.html`.

* Remove `goog.declareModuleId` calls (and added `import`s) from all `tests/mocha/**/*.js`.

* Remove some dead code from `tests/mocha/test_helpers/serialization.js`.

#### Behaviour Before Change

Mocha tests run and 2994/2994 (100%) tests pass in uncompressed mode.

#### Behaviour After Change

Mocha tests run and 2994/2994 (100%) tests pass in uncompressed mode.

Moreover, tests can optionally be run in compressed mode, with 2174/2994 (97%) passing.

### Reason for Changes

Remove dependency on `bootstrap.js` so it and the Closure debug module loader can be deleted.

### Test Coverage

Unchanged.

### Additional Info

Due to the large number of files touched it may be easier to review commit-by-commit.

### DIFF
--- a/scripts/gulpfiles/build_tasks.js
+++ b/scripts/gulpfiles/build_tasks.js
@@ -691,10 +691,8 @@ async function buildShims() {
   // ESM, but fortunately we don't attempt to import or require this
   // file from node.js - we only feed it to Closure Compiler, which
   // uses the type information in deps.js rather than package.json.
-  await fsPromises.writeFile(
-    path.join(BUILD_DIR, 'package.json'),
-    '{"type": "module"}'
-  );
+  const TMP_PACKAGE_JSON = path.join(BUILD_DIR, 'package.json');
+  await fsPromises.writeFile(TMP_PACKAGE_JSON, '{"type": "module"}');
 
   // Import each entrypoint module, enumerate its exports, and write
   // a shim to load the chunk either by importing the entrypoint
@@ -723,6 +721,8 @@ ${Object.keys(exports).map((name) => `  ${name},`).join('\n')}
 );
 `);
   }));
+
+  await fsPromises.rm(TMP_PACKAGE_JSON);
 }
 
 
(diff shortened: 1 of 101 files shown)


rachel-fenichel:
This is a lot, but commit-by-commit it all makes sense. 

If you have not already done so, please make sure that everything still works/tests pass after running `npm run clean && npm run build`.

rachel-fenichel (approved):

PR #7413, opened 2023-08-19
feat: make ordered argument optional for workspace.getAllBlocks

randrei12:
If you're using typescript then you know how frustrating it is to always add a param when calling workspace.getAllBlocks(). That's why I think it's a good idea to make it optional. If a value is not passed, it will use `false` as default.

### DIFF
--- a/core/workspace.ts
+++ b/core/workspace.ts
@@ -206,7 +206,7 @@ export class Workspace implements IASTNodeLocation {
    * @param ordered Sort the list if true.
    * @returns The top-level block objects.
    */
-  getTopBlocks(ordered: boolean): Block[] {
+  getTopBlocks(ordered = false): Block[] {
     // Copy the topBlocks list.
     const blocks = new Array<Block>().concat(this.topBlocks);
     if (ordered && blocks.length > 1) {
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
This is a good idea, and just needs a few tweaks. 

First, just adding the `?` makes the parameter optional with a default value of `undefined`, per the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters). **Please switch to `getAllBlocks(ordered = false): Block[] {` to actually give a default value of `false`.**

Second, we'd actually like to do it for all four related methods in `workspace.ts` and their overrides in `workspace_svg.ts`:
- getAllBlocks
- getTopBlocks
- getTopComments
- getBlocksByType

**Please let me know if you want to make this change for all of the methods**. If not, I'll handle it as soon as I can merge your PR.

Finally, you'll need to **change your PR title to `feat: make ordered argument optional for workspace.getAllBlocks`** after you've added your next commit. You can just do that through the GitHub UI, or let me know if you're having issues with it.

Thanks!

randrei12:
I'll do the changes and also I can take care about the four functions.

PR #7414, opened 2023-08-19
refactor(tests): Migrate generator tests to `import` shims; delete `bootstrap.js`

cpcallen:
## The basics

### Resolves

Part of #6858.

### Proposed Changes

1.  Use shims instead of bootstrap to load Blockly:
    - Modify `tests/generators/index.html` to `import` the test shims instead of using `bootstrap.js` to load Blockly.
    - Modify `test/generators/webdriver.js` to have it wait for the workspace to exist before calling `loadSelected()`.  There was  previously a race which `index.html` had been winning, but now `webdriver.js` is winning (and the tests failing because   there is no workspace yet when `start()` is called.
0.  Delete `tests/bootstrap.js`, `bootstrap_helper.js`, and `bootstrap_done.mjs`.
    - Remove remaining references to `bootstrap.js`.
0.  Remove the `deps` npm script.  The `buildDeps` gulp task is now only needed by `buildCompiled`, not ever for runnning in uncompressed mode, so:
    - Remove the `deps` gulp task (and the `deps` npm script that runs it).
    - Have the `minify` task run `buildJavaScript` and `buildDeps` directly.
    - Additionally, the `buildAdvanceCompilationTest` target hasn't needed `deps.js` for some time (if ever), so skip having it run `buildDeps` entirely.
0.  Repatriate `const DEPS_FILE` to `build_tasks.js`, since it is no longer used anywhere else and so does not need to be declared in in common.js.

#### Behaviour After Change

Unchanged.

### Reason for Changes

Remove `bootstrap.js`.

### Test Coverage

Unchanged.

### Documentation

Need to remove mention of `npm run deps` from developer documentation.


tests/generators/index.html
```diff
@@ -4,20 +4,9 @@
...
-      'tests/generators/unittest_php.js',
-      'tests/generators/unittest_lua.js',
-      'tests/generators/unittest_dart.js',
-      'tests/generators/unittest.js',
-    ],
-  }
-</script>
-<script src="../bootstrap.js"></script>
-<script>
+// N.B.: This script depends on the following (module) script to load
```
rachel-fenichel, line 7: Nit: can you swap the order of the scripts so that the later one depends on the earlier one? It'll still need a comment about the dependency, but at least the ordering will be intuitive.

rachel-fenichel (approved):

tests/generators/index.html
cpcallen: Counterpoint: they are in the order they actually get executed in.  (As a rough approximation: all ordinary `<script>`s get executed before any `<script type="module">`s.)

Does that change your opinion about what order they should be in?

In addition to reordering, I could rewrite the main, plain `<script>` as a module script, but that is somewhat more work.

rachel-fenichel: Counterpoint is reasonable. Fine to leave as is.

PR #7435, opened 2023-08-26
feat: add added the ability to customize the color of  for each blocks

Grougalorasalar:
## The basics

### Resolves

Added the ability to customize the color of field_labels for each block. 
### Proposed Changes

Override applyColour for field_label and add colourFieldLabel in BlockStyle interface.

### Reason for Changes

This provides a wider range of customization options, using colors such as white or yellow with `field_labels` written in black, for example. If the block is disabled, the text color is white.

### Test Coverage

Change tests in theme_test.js file for include colourFieldLabel in expectedOutput. 

### Documentation

Update BlockStyle (add explanation of colourFieldLabel)  and FieldLabe (add explanation of applyColour ) .

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -246,6 +246,15 @@ export class BlockSvg
   }
 
   /**
+   * Get the field_label colour of a block.
+   *
+   * @returns #RRGGBB string.
+   */
+  getColourFieldLabel(): string | undefined {
+    return this.style.colourFieldLabel;
+  }
+
+  /**
    * Selects this block. Highlights the block visually and fires a select event
    * if the block is not already selected.
    */
--- a/core/field_label.ts
+++ b/core/field_label.ts
@@ -126,7 +126,10 @@ export class FieldLabel extends Field<string> {
     return new this(text, undefined, options);
   }
 
-  applyColour() {
+  /**
+   * Update field's textElement_ color.
+   */
+  override applyColour() {
     const style = (this.sourceBlock_ as BlockSvg).style;
     if (this.textElement_) {
       this.textElement_.style.fill = this.sourceBlock_?.isEnabled()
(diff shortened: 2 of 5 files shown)


BeksOmega:
Hello! I think you can do this with our current css selectors:

```css
/* replace geras-renderer.classic-theme with your theme and renderer */
.geras-renderer.classic-theme .blocklyDisabled .blocklyText {
  fill: #000;
}
```

Does that solve your use-case?

Grougalorasalar:
Yes, but if I would a black block and white block I need white text for the first block and black text for the second block. The goal of the PR is to have a large color palette for blocks, making them more legible by changing the text color. However, we could used this : https://www.w3.org/WAI/WCAG21/Techniques/general/G18 for automatic adaption text color but we have less possibilities color for the texts blocks.

BeksOmega:
> Yes, but if I would a black block and white block I need white text for the first block and black text for the second block. The goal of the PR is to have a large color palette for blocks, making them more legible by changing the text color. However, we could used this : https://www.w3.org/WAI/WCAG21/Techniques/general/G18 for automatic adaption text color but we have less possibilities color for the texts blocks.

That makes sense! But I think in general as a library we want to move away from doing styling changes in JavaScript and instead have people use CSS. This allows the browser to better optimize repaints, and makes our API more maintainable.

If we applied a css class to the block that was the block's [block-style](https://developers.google.com/blockly/guides/configure/web/appearance/themes#block_style) name, do you think you could achieve what you want with CSS?

BeksOmega:
> > Yes, but if I would a black block and white block I need white text for the first block and black text for the second block. The goal of the PR is to have a large color palette for blocks, making them more legible by changing the text color. However, we could used this : https://www.w3.org/WAI/WCAG21/Techniques/general/G18 for automatic adaption text color but we have less possibilities color for the texts blocks.
> 
> That makes sense! But I think in general as a library we want to move away from doing styling changes in JavaScript and instead have people use CSS. This allows the browser to better optimize repaints, and makes our API more maintainable.
> 
> If we applied a css class to the block that was the block's [block-style](https://developers.google.com/blockly/guides/configure/web/appearance/themes#block_style) name, do you think you could achieve what you want with CSS?

Heya @Grougalorasalar Still definitely interested in your thoughts on this!

PR #7473, opened 2023-09-08
refactor(build): Remove `closure-make-deps` and `closure-calculate-chunks`

### LINKED ISSUE #7469: Remove `closure-make-deps` and `closure-calculate-chunks`
Since each chunk is basically just the files in a particular subdirectory (`core/`, `blocks/`, `generators/*/`) we shouldn't need to use `closure-calculate-chunks` to compute the command line arguments for Closure Compiler—and if we don't need `closure-calculate-chunks` then we also no longer need `closure-make-deps`, since the former is the only remaining consumer of the output from the latter.

cpcallen:
## The basics

### Resolves

Fixes #7469 

### Proposed Changes

* Rewrite the `getChunkOptions` function to not use `closure-calculate-chunks`, but instead just chunk the input files (more or less) by subdirectory: first chunk is `core/`, second is `blocks/`, etc.

  This does make a material change to `blockly_compressed.js`, because we end up feeding several empty modules that contain only typescript interface declarations and which `tsc` compiles to `export {};` in the input to Closure Compiler
(`closure-calculate-chunks` is smart enough to notice that no other module depends on these), which results in ~1.7KiB of
superflous `var module$build$src$core$interfaces$i_ast_node_location_svg={};` declarations.  This can be avoided by filtering such empty modules out, but that has been left for a future commit.

  This adds the `glob` NPM package as a dev dependency, but `gulp` and several other existing dev dependencies already depend on it.

  Build time is sped up by about a factor of 3x, due to being able to skip the `buildDeps` step that was really slow:
  ```
  $ time npm run build
  # Before:
  real    0m24.410s
  user    0m16.010s
  sys     0m1.140s
  # After
  real    0m8.397s
  user    0m11.976s
  sys     0m0.694s
  ```

* Remove the `buildDeps` task.
* Remove the `$build$src` infix from munged paths.

  Closure Compiler renames module globals so that they do not clash when multiple modules are bundled together.  It does so by adding a `$$module$build$src$path$to$module` suffix (with the module object itself being named simply `$module$build$src$path$to$module`).

  By changing the `gulp.src` `base` option to be `build/src/` instead of `./` (referring to the repository root), Closure Compiler obligingly shortens all of these munged names by removing the `$build$src` infix, reducing the size of the compressed chunks by about 10%; `blockly_compressed.js` goes from 900595 to 816667 bytes.

* Compute module object munged name from entrypoint:
  - Add function `modulePath` to compute the munged name of the entrypoint  module from the entrypoint module filename, and use this instead of hard-coded `chunk.exports` paths.
  - Be more careful about when we are using posix vs. OS-specific  paths, especially `TSC_OUTPUT_PATH` which is regularly passed to `gulp.src`, which is documented as taking only posix paths.
  - Rename one existing variable `modulePath` -> `entryPath` to try to avoid confusion with the new `modulePath` function.

### Reason for Changes

Fewer dependencies, faster builds, simpler build config!

### Test Coverage

`npm test` passes; playground loads in compressed and uncompressed mode.


rachel-fenichel:
I'm excited about this one!

In the past we've had a few developers who exclusively use Windows report issues with our build system. Please tag in one or two of them to check that this still works on Windows.

scripts/gulpfiles/build_tasks.js
```diff
@@ -39,14 +40,14 @@ const {posixPath, quote} = require('../helpers');
...
 
 /**
- * Dependencies file (used by buildCompiled for chunking.
+ * Suffix to add to compiled output files.
  */
-const DEPS_FILE = path.join(BUILD_DIR, 'deps.js');
+const COMPILED_SUFFIX = '_compressed';
 
 /**
  * Name of an object to be used as a shared "global" namespace by
```
rachel-fenichel, line 53: Is this comment still accurate after your changes to chunking?

rachel-fenichel (approved):
I love this PR.

In addition to finding someone to test on Windows, it's worth packaging up a beta and testing it with examples and plugins to make sure nothing was magically depending on a detail of the build system.

scripts/gulpfiles/build_tasks.js
cpcallen: I believe so.

This PR changes only the way the command line arguments to Closure Compiler are computed, not the actual arguments themselves—with two exceptions:

1. As noted, the chunks now contain some additional `.js` files that are empty except for `export {}`, caused by `.ts` files containing only `interface` declarations.  `closure-calculate-chunks` would previously omit these because they are not imported from anywhere, and it knew this based on the `deps.js` file generated by `closure-make-deps`.  This adds a tiny bit of bloat to the `_compressed.js` files, which could be fixed by having the new `getChunks` function check be smarter:
   -  Minimal improved smartness: just omit files any file containing `export {}`.
   - Perfect smartness: recursively parse the entrypoint file and all the files it imports, and only include files seen (and not seen in a previous chunk).  But then we are essentially reimplementing `closure-make-deps` + `closure-calculate-chunks` (though we could probably make it much faster for our specific use case).
0. Within each chunk, the order in which the files are listed on the Closure Compiler command line has changed.  This probably has knock-on effects to the order in which code appears in the `_compressed.js` file.

As part of testing this I compared the Closure command line arguments before and after this PR; modulo trivial differences (file order within chunks, and paths starting from `/Users/cpcallen/src/blockly/` vs. `./`) the arguments were the same, and when I sorted the files the generated `_compressed.js` files were byte-for-byte identical except for the added `var module$core$interfaces$i_foo={};` declarations for each empty interface modules discussed above.

cpcallen:
> I love this PR.

Thanks!

> In addition to finding someone to test on Windows, it's worth packaging up a beta and testing it with examples and plugins to make sure nothing was magically depending on a detail of the build system.

I attempted to test this myself on Windows, but I don't know a lot about how Windows devs typically set up their machines.

cpcallen:
@yamadayutaka: You have been our main Windows developer when it comes build system testing.  Can you pull this change (now merged to `develop` and verify that `npm run clean ; npm test` succeeds on Windows?

Also, I now have access to a Windows machine but am not sure how to set it up in a way that best represents how developers like you work.  An explanation of how you have things set up (what did  you install, and from where?  Are you using WSL + bash or just Command Prompt?  etc. etc.) would be very useful.

yamadayutaka:
Hi @cpcallen, I tried running `npm test` in my environment and got the following error.

```
=======================================
== renamings
'tests' is not recognized as an internal or external command,
operable program or batch file.
Command failed: tests/migration/validate-renamings.mjs
FAILED: renamings
```

The `validate-renamings.mjs` is executed by shebang, but it does not seem to work on Windows.

If you change the `renamings` function as follows, it works fine on Windows.

```
function renamings() {
  return runTestCommand('renamings', 'node tests/migration/validate-renamings.mjs');
}
```

> Also, I now have access to a Windows machine but am not sure how to set it up in a way that best represents how developers like you work. An explanation of how you have things set up (what did you install, and from where? Are you using WSL + bash or just Command Prompt? etc. etc.) would be very useful.

I think it would be better to be able to build with PowerShell Propmpt instead of using WSL.

PR #7478, opened 2023-09-11
feat: Replace getValue check with getVariable for variable fields.

laurensvalk:
## The basics

### Resolves

Fixes `block.getVars` currently detecting only variables in `FieldVariable`, as indicated by the (now deleted) note:

```
// NOTE: This only applies to `FieldVariable`, a `Field<string>`
```

### Proposed Changes

There seems to be a hook intended for this already --- `getVariable()` --- but it isn't used everywhere.

The proposal is to replace `field.getValue()` check with `field.getVariable()?.getId()` or similar in `block.getVars` and related methods.

To make this work, also make `getVariable` available on fields. This matches `referencesVariables` and `refreshVariableName` prototypes being available on fields already.

### Reason for Changes

Getting a variable using `getValue` applies only to `FieldVariable`. There was a note to indicate this; this replaces it with the relevant check. This ensures that variables can be used with fields other than dropdowns, as seems to be the intention.

### Test Coverage

Added a test for `field_variable` to verify that `getValue` matches the `id` obtained from the variable.

### Documentation

`getVariable` is now available on fields.

### Additional Information

This is somewhere between a fix and a feature (I marked it as `feat` because of the added property). A use case could be a text field that declares a variable. For example to have a dedicated declare variable block instead of a "create variable" button.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -1104,7 +1104,7 @@ export class Block implements IASTNodeLocation, IDeletable {
   }
 
   /**
-   * Return all variables referenced by this block.
+   * Return id strings of all variables referenced by this block.
    *
    * @returns List of variable ids.
    */
@@ -1112,11 +1112,17 @@ export class Block implements IASTNodeLocation, IDeletable {
     const vars: string[] = [];
     for (let i = 0, input; (input = this.inputList[i]); i++) {
       for (let j = 0, field; (field = input.fieldRow[j]); j++) {
+        const variable = field.getVariable();
+        if (variable) {
+          vars.push(variable.getId());
+          continue;
+        }
+
+        // Backwards-compatible check used for fields that stored variable id
+        // as their value. Deprecated, implement Field.getVariable() instead.
+        // When this is removed, the continue keyword above can be removed too.
         if (field.referencesVariables()) {
-          const variable = field.getVariable();
-          if (variable) {
-            vars.push(variable.getId());
-          }
+          vars.push(field.getValue() as string);
         }
       }
     }
@@ -1132,8 +1138,9 @@ export class Block implements IASTNodeLocation, IDeletable {
   getVarModels(): VariableModel[] {
     const vars = [];
     for (const id of this.getVars()) {
+      // We have a variable at this point, but filter it through the workspace
+      // to be certain it's not just a potential variable.
       const model = this.workspace.getVariableById(id);
-      // Check if the variable exists (and isn't just a potential variable).
       if (model) {
         vars.push(model);
       }
@@ -1151,9 +1158,17 @@ export class Block implements IASTNodeLocation, IDeletable {
   updateVarName(variable: VariableModel) {
     for (let i = 0, input; (input = this.inputList[i]); i++) {
       for (let j = 0, field; (field = input.fieldRow[j]); j++) {
+        if (variable.getId() === field.getVariable()?.getId()) {
+          field.refreshVariableName();
+          continue;
+        }
+
+        // Backwards-compatible check used for fields that stored variable id
+        // as their value. Deprecated, implement Field.getVariable() instead.
+        // When this is removed, the continue keyword above can be removed too.
         if (
           field.referencesVariables() &&
-          variable.getId() === field.getVariable()?.getId()
+          variable.getId() === field.getValue()
         ) {
           field.refreshVariableName();
         }
@@ -1172,10 +1187,16 @@ export class Block implements IASTNodeLocation, IDeletable {
   renameVarById(oldId: string, newId: string) {
     for (let i = 0, input; (input = this.inputList[i]); i++) {
       for (let j = 0, field; (field = input.fieldRow[j]); j++) {
-        if (
-          field.referencesVariables() &&
-          oldId === field.getVariable()?.getId()
-        ) {
+        if (oldId === field.getVariable()?.getId()) {
+          field.setVariable(this.workspace.getVariableById(newId));
+          continue;
+        }
+
+        // Backwards-compatible check used for fields that stored variable id
+        // as their value. Deprecated, implement Field.getVariable() and
+        // Field.setVariable() instead. When this is removed, the continue
+        // keyword above can be removed too.
+        if (field.referencesVariables() && oldId === field.getValue()) {
           field.setValue(newId);
         }
       }
--- a/core/block.ts
+++ b/core/block.ts
@@ -51,6 +51,7 @@ import {EndRowInput} from './inputs/end_row_input.js';
 import {ValueInput} from './inputs/value_input.js';
 import {StatementInput} from './inputs/statement_input.js';
 import {IconType} from './icons/icon_types.js';
+import { deprecation } from './utils.js';
 
 /**
  * Class for one block.
@@ -1122,6 +1123,12 @@ export class Block implements IASTNodeLocation, IDeletable {
         // as their value. Deprecated, implement Field.getVariable() instead.
         // When this is removed, the continue keyword above can be removed too.
         if (field.referencesVariables()) {
+          deprecation.warn(
+            'Blockly.Field.referencesVariables',
+            'version 10',
+            'version 11',
+            'Blockly.Field.getVariable()',
+          );
           vars.push(field.getValue() as string);
         }
       }
@@ -1170,6 +1177,12 @@ export class Block implements IASTNodeLocation, IDeletable {
           field.referencesVariables() &&
           variable.getId() === field.getValue()
         ) {
+          deprecation.warn(
+            'Blockly.Field.referencesVariables',
+            'version 10',
+            'version 11',
+            'Blockly.Field.getVariable()',
+          );
           field.refreshVariableName();
         }
       }
@@ -1197,6 +1210,12 @@ export class Block implements IASTNodeLocation, IDeletable {
         // Field.setVariable() instead. When this is removed, the continue
         // keyword above can be removed too.
         if (field.referencesVariables() && oldId === field.getValue()) {
+          deprecation.warn(
+            'Blockly.Field.referencesVariables',
+            'version 10',
+            'version 11',
+            'Blockly.Field.setVariable()',
+          );
           field.setValue(newId);
         }
       }
--- a/core/block.ts
+++ b/core/block.ts
@@ -1113,8 +1113,10 @@ export class Block implements IASTNodeLocation, IDeletable {
     for (let i = 0, input; (input = this.inputList[i]); i++) {
       for (let j = 0, field; (field = input.fieldRow[j]); j++) {
         if (field.referencesVariables()) {
-          // NOTE: This only applies to `FieldVariable`, a `Field<string>`
-          vars.push(field.getValue() as string);
+          const variable = field.getVariable();
+          if (variable) {
+            vars.push(variable.getId());
+          }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


core/field.ts
```diff
@@ -1263,6 +1264,16 @@ export abstract class Field<T = any>
...
   }
 
+  /**
+   * Get the variable model if this field refers to one.
+   *
+   * @returns The field's variable, or null if none exists.
+   * @internal
+   */
+  getVariable(): VariableModel | null {
+    return null;
```
BeksOmega, line 1274: So as the code currently exists, this is a breaking change because we're requiring folks to implement the `getVariable()` method to return the variable they're referencing, which we didn't require before.

We can make this backwards compatible by having `getVariable` get the variable based on the value if `referencesVariables` is true.

Please add a unit test for this as well using a mock `Field` subclass.

core/field.ts
```diff
@@ -1263,6 +1264,16 @@ export abstract class Field<T = any>
     return false;
   }
 
+  /**
+   * Get the variable model if this field refers to one.
+   *
+   * @returns The field's variable, or null if none exists.
+   * @internal
```
BeksOmega, line 1271: If you want this to be overriden in subclasses, it should be public not internal.

core/field_variable.ts
```diff
@@ -312,7 +312,7 @@ export class FieldVariable extends FieldDropdown {
    * @returns The selected variable, or null if none was selected.
    * @internal
```
BeksOmega, line 313: Similarly `@internal` should be removed here.

core/block.ts
```diff
@@ -1164,7 +1159,10 @@ export class Block implements IASTNodeLocation, IDeletable {
   renameVarById(oldId: string, newId: string) {
     for (let i = 0, input; (input = this.inputList[i]); i++) {
       for (let j = 0, field; (field = input.fieldRow[j]); j++) {
-        if (field.referencesVariables() && oldId === field.getValue()) {
+        if (
+          field.referencesVariables() &&
```
BeksOmega, line 1163: Since we can expected `getVariable` to return null if the field doesn't reference any variables, probably best to deprecated `referencesVariables` as part of this change as well.

@rachel-fenichel I'm not certain what our policies are around deprecating things. I think we usually just make a judgement call. How do you feel about this one?

core/field.ts
```diff
@@ -1263,6 +1264,16 @@ export abstract class Field<T = any>
...
+   */
+  getVariable(): VariableModel | null {
```
BeksOmega, line 1273: Can you also change this to return an array instead of a single model, and update the other logic accordingly? If we're fixing up this code anyway might as well  make it more general :P

core/block.ts
```diff
@@ -1116,18 +1118,11 @@ export class Block implements IASTNodeLocation, IDeletable {
...
-          );
-          // Check if the variable actually exists (and isn't just a potential
-          // variable).
-          if (model) {
-            vars.push(model);
-          }
-        }
+    for (const id of this.getVars()) {
+      const model = this.workspace.getVariableById(id);
+      // Check if the variable exists (and isn't just a potential variable).
```
BeksOmega, line 1123: Hmm this makes me suspicious that the other code will actually fail if there is a potential variable. I'm not sure where these methods are being called from exactly though. Did you dig into this at all? If not I can take a look!

BeksOmega (changes requested):
This looks like a good change! But needs some modifications for it to be ready for release =)

core/block.ts
laurensvalk: > Since we can expected getVariable to return null if the field doesn't reference any variables, probably best to deprecated referencesVariables as part of this change as well.

Sounds good to me. I'll wait for your decision and update accordingly.

core/field.ts
laurensvalk: Thanks! My bad, coming from Python :) 

I was following this pattern from `referencesVariables`, which is `@internal` in `Field` and overridden by `FieldVariable`, also as `@internal`. I suppose we should drop it from there too, then.

core/field.ts
laurensvalk: Thanks. I suppose we could do something like the following, and deprecate this check if/when `referencesVariables` is dropped. I'm not entirely sure where to get the variable from, however. It might not live on the source block's workspace.

Or maybe we could handle backwards compatibility at the `getVars` stage where it was calling `getValue()`. Most other places in the API already seem to be correctly using `getVariable` already. I'll have a look.

```js
//  field.ts   //

  /**
   * Get the variable models this field refers to, if any.
   * 
   * @returns The field's variable, or null if none exists.
   */
  getVariables(): VariableModel[] {
    // Previously, when referencesVariables was true, we assumed getValue()
    // returned a variable ID. Keep allowing this for backwards compatiblity.
    if (this.referencesVariables()) {
      const variable = this.getSourceBlock()?.workspace.getVariableById(
        this.getValue() as string,
      )
      return variable ? [variable] : [];
    }

    // Fields don't reference variables by default. Subclasses are expected
    // to override this method if they do.
    return [];
  }
```

core/field.ts
laurensvalk: Sure, although `getVariable()` is already used in a few places, which might make it a breaking change too?

core/block.ts
laurensvalk: I think this part works the same as it did before. This is used, among others:
- getVarModels in procedures
- getVariableUsesById(id) / allUsedVarModels(ws)

core/block.ts
rachel-fenichel: That sounds right. Previously we needed the check because the string could be empty if it was just a different type of field, but the extra check is unnecessary when calling `getVariable` instead of `getValue`

core/block.ts
laurensvalk: > Hmm this makes me suspicious that the other code will actually fail if there is a potential variable.

With other code, did you mean `getVars`?

I think I see what you mean. I had kept the checks as they were to keep things simple, but we could:
- Use the "safer" `getVarModels()` instead of `getVars` where needed (or everywhere).
- Have `getVars` check that they are in the workspace just like `getVarModels()`, but this is a slight breaking change.

core/field.ts
laurensvalk: I think I'm going to keep it singular in round one :smile: Perhaps we can do a follow up commit for the plural case when everything else is sorted and squash it.

laurensvalk:
Updated:
- Implemented `setVariable` to complement `getVariable`, needed where Block used to call `setValue`, for the same reason.
- Dropped incorrect use of `@internal`.
- Added backwards compatibility in `Block.getVars` and friends to use `referencesVariables` when `getVariable` is not implemented.
- Added `@deprecated` to `referencesVariables`. Also added `To be removed in v??.` as seen on other methods. Not sure what the deprecation version should be.
- Stayed with singular `getVariable` instead of suggested `getVariables` for the moment a explained above.

core/field.ts
laurensvalk: Addressed, see follow-up notes in main PR conversation thread.

core/block.ts
```diff
@@ -1104,16 +1104,24 @@ export class Block implements IASTNodeLocation, IDeletable {
...
+        const variable = field.getVariable();
+        if (variable) {
+          vars.push(variable.getId());
+          continue;
+        }
+
+        // Backwards-compatible check used for fields that stored variable id
+        // as their value. Deprecated, implement Field.getVariable() instead.
+        // When this is removed, the continue keyword above can be removed too.
         if (field.referencesVariables()) {
```
BeksOmega, line 1124: Can you add a call to `deprecation.warn` in places where `referencesVariables` is called as well?

core/block.ts
BeksOmega: I have a few other PRs to dig into so I will circle back on this later today!

BeksOmega (changes requested):

core/field.ts
```diff
@@ -1280,12 +1282,19 @@ export abstract class Field<T = any>
...
   getVariable(): VariableModel | null {
     return null;
   }
 
+  /**
+   * Sets or clears the variable model.
+   *
+   * @param variable The variable to set or null to clear.
+   */
+  setVariable(_variable: VariableModel | null) {
```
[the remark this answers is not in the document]
laurensvalk: Right, we seem to have something more like this at the moment

```
            Field
         /         \
        /           \
FieldDropDown     FieldWithVariable*    // i.e. Would have been called "FieldVariable" if not already taken
       \            /
        \          /
        FieldVariable*                  // i.e. Would have been called "FieldVariableDropdown"  if not already named
```

So maybe it could be more like this?

```
Field
  |
  |
  |
FieldDropDown   FieldVariableMixin
  |            / 
  |           /
  |          /
  |         /
FieldVariable
```

There is probably a name for this, but I don't have a comp-sci background :smile: 

That way, perhaps more than `getVariable` and `setVariable` can be factored out.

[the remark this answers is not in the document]
BeksOmega: Haha yeah there's definitely a design flaw in our fields where we tightly coupled the data being modeled to the visual representation. But that is a larger problem :P

I think for now just factoring out `getVariables`, `setVariables`, and `refreshVariableName` into an interface is fine. I think all the other variable-model specific methods in `FieldVariable` are private!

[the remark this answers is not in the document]
laurensvalk: Thanks for taking the time on this review, I appreciate it.

I will do as you suggested and try to get back to you this week.

laurensvalk:
> Heya @laurensvalk as a heads up we're planning to do a release Thursday the 28th. Just wanted to give you the time line in case it's important to you that this goes into the next release! But no pressure if you're busy. Our next release after that will be ~mid November (6 weeks out from the 28th).

Thanks for the heads up!

And yes, unlike with most packages I haven't been able to use `patch-package` on Blockly, so I'll do my best to get it merged in properly, which is probably for the best anyway :smile:

laurensvalk:
> Heya @laurensvalk as a heads up we're planning to do a release Thursday the 28th. Just wanted to give you the time line in case it's important to you that this goes into the next release! But no pressure if you're busy. Our next release after that will be ~mid November (6 weeks out from the 28th).

Hi @BeksOmega - I will be away next week, so don’t wait for me. I’ll be happy to give this a look for a future release as applicable.

BeksOmega:
Heya @laurensvalk Are you still interested in working on this? or should I go ahead and close it?

laurensvalk:
Still interested! It's on my todo list. I am using a patch at the moment to pull this in but getting it upstream would be awesome.

BeksOmega:
Sorry I missed this! 

> I've added a commit where it sets the variable but also sets the `id` as its value for backwards compatibility.

That sounds perfect =)

> Is there a way to get more detailed output for failed `mocha` tests?

You can run `npm run test:mocha:interactive` locally. But atm there's no way to gt more info out of the CI.

BeksOmega:
> I can now add some context to this! Here's a demo of using fields with variables 🙂
> 
> The setup blocks have text fields that create typed variables. Other blocks have the usual dropdown, but without the rename and delete options, since this can be done on the setup blocks:
> 
>  varfield.mp4

Love the different connection shapes for setup vs normal logic!

BeksOmega:
Heya @laurensvalk are you still interested in working on this? I'll check back in a week and close it if not!

laurensvalk:
I had updated it with the `setValue` logic, but hadn't gotten around to the mocha tests yet.

I'm happy to close it and keep using the patch for now, and possibly give it a fresh go for future release. Thanks!

Issue #7426, opened 2023-08-23
Chasing reporter block around in zelos


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

_This bug was originally reported in #7412_ by @clementcontet

See the screenshot for the originally reported bug. 

I think the mechanism is: Sometimes when the insertion marker appears it moves the valid connection far enough away from the target connection that it becomes invalid, which makes the target block disconnect and get bumped away from the dragging block.

This is more likely to happen when using zelos because it changes the height of the input and moves the connection location down, while geras and thrasos leave the connection in the same location relative to the top left of the dragging block.

To verify: try the same drag with taller and shorter value blocks. The problem should become more visible the taller the value block is.

### Reproduction steps

1. 
2. 
3.

### Screenshots

https://github.com/google/blockly/assets/13686399/3c9ebe01-eb12-4388-bffb-f9bc81f424dc

clementcontet:
Hi Rachel,

Indeed I don't have the problem with small child blocks, or when switching to Geras.

But I found that even when the connection is done directly (without "chasing"), when the insertion marker appears, it appears at a place that makes the child block to be moved below.
Whereas in Geras, the insertion marker appears at the exact place that matches the child block current position.


https://github.com/google/blockly/assets/7261426/b1309d0b-dec1-47cf-9924-f00c822644ed

rachel-fenichel:
The fix would be to note the difference between the position of the connection when the value input is expanded vs the position of the connection when the value input is at its default size, and to shift the parent bock up and left by that difference.

But this would cause a different problem: the dragged block would have to jump relative to the mouse. I think that would be worse. What do you think?

clementcontet:
> the dragged block would have to jump relative to the mouse

We would first see the insertion marker of the block on its target position, right?
In that case, the "jump" would only occur when we drop the dragged block, and it would indeed be placed at the exact place of the insertion marker.

And I think we already see this "jump" with insertion marker:

https://github.com/google/blockly/assets/7261426/4fd56b87-3109-4655-9a2e-c47826310465

rachel-fenichel:
You're correct. Okay, we'll look at getting the marker to position so that the connecting block doesn't have to move.

rachel-fenichel:
@clementcontet If you want to fix this, take a look at the [`positionNearConnection`](positionNearConnection) function in `block_svg.ts`

clementcontet:
[...]

PR #7482, opened 2023-09-12
fix: insertion marker position when connection is resized (#7426)

clementcontet:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7426

### Proposed Changes

When positioning the insertion marker, take into account that its connection may have been resized.

### Reason for Changes

With these changes, when positioning the insertion marker, we ensure that the target connection will not move.

### Test Coverage

I tested in the playground with Zelos and Geras that insertion markers don't move target connections anymore, even if their connection have been resized.

### Documentation

N/A

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -614,12 +614,19 @@ export class InsertionMarkerManager {
     // Connect() also renders the insertion marker.
     imConn.connect(closest);
 
-    const originalOffsetToTarget = {x: closest.x - imConn.x, y: closest.y - imConn.y}
+    const originalOffsetToTarget = {
+      x: closest.x - imConn.x,
+      y: closest.y - imConn.y,
+    };
     const originalOffsetInBlock = imConn.getOffsetInBlock().clone();
     const imConnConst = imConn;
     renderManagement.finishQueuedRenders().then(() => {
       // Position so that the existing block doesn't move.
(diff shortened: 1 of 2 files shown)


core/insertion_marker_manager.ts
```diff
@@ -611,12 +611,22 @@ export class InsertionMarkerManager {
...
-    insertionMarker.positionNearConnection(imConn, closest);
     // Connect() also renders the insertion marker.
     imConn.connect(closest);
 
+    const originalOffsetToTarget = {
+      x: closest.x - imConn.x,
+      y: closest.y - imConn.y,
+    };
+    const originalOffsetInBlock = imConn.getOffsetInBlock().clone();
+    const imConnConst = imConn;
```
rachel-fenichel, line 622: nit: You don't need to assign this to another variable. You can just pass it directly.

rachel-fenichel (commented):
Logic looks good. There's one nit to fix, and then I'll merge it.

core/insertion_marker_manager.ts
clementcontet: Hum, no.
These values are modified during the render.
That's why I get their "original" value in constants before the render.

rachel-fenichel: Oh, I see.

rachel-fenichel (approved):

Issue #7484, opened 2023-09-12
contextMenuSelect clicks on the wrong block


rachel-fenichel:
`contextMenuSelect` in `tests/browser/test/test_setup.js` is supposed to click on the first bit of text on the block, which it finds with `const clickEl = block.$('.blocklyText')`. This works around a problem where clicking always happens in the center of the block's bounding rect, which may be  hole in the block.

This selector strategy works if the block has no children, but fails if the block has blocks connected to inputs. Due to the way that we construct our SVGs, the child block is earlier than the parent block's fields in the DOM.

A simple solution is to look for the last text element on the block, because we don't actually care *which* text element is selected as long as it's on the block. That works with blocks connected to inputs, but fails if there are more blocks in the stack: a block connected to the next connection is after the parent block in the DOM.

### Reproduction

I discovered this when updating webdriverio. There were other places in code where we simply clicked on the block's SVG group, rather than the first text element. After updating I got the [error](https://webdriver.io/docs/api/element/click/) "Element is not clickable at point (x, x). Other element would receive the click: ..."

I tried to resolve this by following the pattern in `contextMenuSelect` and found that it's wrong and just happens to work because many tests only use single blocks.

### Possible solutions

- Find a combination of selectors that accurately selects the first text element in the block's group.
- Drop into the actual browser using `browser.execute` and get an element that way.

In either case there should be a helper function `getClickTarget(block)` that can be used by both `contextMenuSelect` and code that needs to left-click.

PR #7485, opened 2023-09-12
fix(tests): contextMenuSelect sometimes clicks the wrong block

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7484

### Proposed Changes

- Add a helper function that tags the first label field on a block, then uses that to get that element as a Webdriverio Element.
- Change context menu helper function to take in a block ID instead of a block element.
- Use the helper function in all context menu tests and in block deletion tests.

### Reason for Changes

Fix issues with clicking when the block's path is occluded or has a hole in it.

### Test Coverage

Ran browser tests and watched for unexpected behaviour.

### Documentation

I added JSDoc

### Additional Information

- You could run into issues if you used this function twice, on different blocks, within the same test context. In that case the solution is to add a helper that removes the ID from the SVG element. This seems not worth adding unless/until it becomes an issue.
- There are other tests that click on blocks and have not previously had any issues. I left them in their current state (clicking on the block's SVG root). I don't see a benefit to updating them to use the more complicated helper function unless they break.
  - I could be convinced otherwise just for consistency.

### DIFF
--- a/tests/browser/test/delete_blocks_test.js
+++ b/tests/browser/test/delete_blocks_test.js
@@ -168,7 +168,7 @@ suite('Delete blocks', function (done) {
     const before = (await getAllBlocks(this.browser)).length;
     // Get first print block, click to select it, and delete it using context menu.
     const block = await getBlockElementById(this.browser, firstBlockId);
-    await contextMenuSelect(this.browser, block, 'Delete 2 Blocks');
+    await contextMenuSelect(this.browser, block.id, 'Delete 2 Blocks');
     const after = (await getAllBlocks(this.browser)).length;
     chai.assert.equal(
       before - 2,
--- a/tests/browser/test/delete_blocks_test.js
+++ b/tests/browser/test/delete_blocks_test.js
@@ -10,6 +10,7 @@ const {
   testFileLocations,
   getAllBlocks,
   getBlockElementById,
+  getClickableBlockElementById,
   contextMenuSelect,
   PAUSE_TIME,
 } = require('./test_setup');
@@ -132,13 +133,14 @@ suite('Delete blocks', function (done) {
     });
   });
 
-  test('Delete block using backspace key', async function () {
+  test.only('Delete block using backspace key', async function () {
     const before = (await getAllBlocks(this.browser)).length;
     // Get first print block, click to select it, and delete it using backspace key.
-    const block = (await getBlockElementById(this.browser, firstBlockId)).$(
-      '.blocklyPath',
-    );
-    await block.click();
+    // const block = (await getBlockElementById(this.browser, firstBlockId)).$(
+    //   '.blocklyPath',
+    // );
+    const clickEl = await getClickableBlockElementById(this.browser, firstBlockId);
+    await clickEl.click();
     await this.browser.keys([Key.Backspace]);
     const after = (await getAllBlocks(this.browser)).length;
     chai.assert.equal(
--- a/tests/browser/test/delete_blocks_test.js
+++ b/tests/browser/test/delete_blocks_test.js
@@ -133,12 +133,9 @@ suite('Delete blocks', function (done) {
     });
   });
 
-  test.only('Delete block using backspace key', async function () {
+  test('Delete block using backspace key', async function () {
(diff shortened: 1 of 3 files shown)


tests/browser/test/test_setup.js
```diff
@@ -428,18 +460,13 @@ async function dragBlockFromMutatorFlyout(browser, mutatorBlock, type, x, y) {
  * context menu item.
  *
  * @param browser The active WebdriverIO Browser object.
- * @param block The block to click, as an interactable element. This block must
+ * @param blockId The ID of the block to click. This block should
  *    have text on it, because we use the text element as the click target.
  * @param itemText The display text of the context menu item to click.
  * @return A Promise that resolves when the actions are completed.
  */
-async function contextMenuSelect(browser, block, itemText) {
```
BeksOmega, line 436: In general I think it's easiest if we pass the block element (which has an ID property) around to all of our helpers. This makes the code more consistent, and makes it easier for us to refactor the test helpers if we need to, because it's less likely we have to change test code as well.

tests/browser/test/delete_blocks_test.js
```diff
@@ -135,10 +136,8 @@ suite('Delete blocks', function (done) {
   test('Delete block using backspace key', async function () {
     const before = (await getAllBlocks(this.browser)).length;
     // Get first print block, click to select it, and delete it using backspace key.
-    const block = (await getBlockElementById(this.browser, firstBlockId)).$(
-      '.blocklyPath',
-    );
-    await block.click();
+    const clickEl = await getClickableBlockElementById(this.browser, firstBlockId);
+    await clickEl.click();
```
BeksOmega, line 140: Instead of calling this and then calling `click` could you create a `clickBlock` helper that does this and then removes the special ID? Then it cleans up after itself, and I think the code would be easier to read. If the `clickBlock` helper takes in the same options as `click` you could also use it from the context menu code.

tests/browser/test/test_setup.js
```diff
@@ -155,6 +155,38 @@ async function getBlockElementById(browser, id) {
   return elem;
 }
 
+/**
+ * Get a clickable element on the block. We can't always use the block's SVG root
+ * because clicking will always happen in the middle of the block's bounds
+ * (including children) by default, which causes problems if it has holes
+ * (e.g. statement inputs). Instead, this tries to get the first text field on the
+ * block. It falls back on the block's SVG root.
```
BeksOmega, line 163: Appreciate the detailed english description here!

tests/browser/test/test_setup.js
```diff
@@ -155,6 +155,38 @@ async function getBlockElementById(browser, id) {
...
+ * @param browser The active WebdriverIO Browser object.
+ * @param id The ID of the Blockly block to search for.
+ * @return A Promise that resolves to the text element of the first label
+ *     field on the block, or the block's SVG root if no label field was found.
+ */
+async function getClickableBlockElementById(browser, id) {
```
BeksOmega, line 169: Similarly would like to pass the block here instead of the ID.

tests/browser/test/test_setup.js
rachel-fenichel: This is true for cases where we're creating the blocks by dragging from the flyout but false for cases where we create blocks by loading from JSON or XML. In the tests above, all of the basic playground tests and some of the delete tests create blocks from the flyout, but the other delete tests load from XML and just use a block ID to find the blocks. As I see it, every time we have a handle on a block we also have a handle on the block's ID, but the opposite is not guaranteed to be true.

Thoughts?

BeksOmega: I think for those cases you can just get the block by ID using our helper for that.

tests/browser/test/delete_blocks_test.js
rachel-fenichel: Done.

tests/browser/test/test_setup.js
rachel-fenichel: Done.

rachel-fenichel:
Updated per your suggestions. PTAL.

BeksOmega (approved):

PR #7489, opened 2023-09-14
feat: Allow each extension to add a change handler.

laurensvalk:
## The basics

This is a draft PR for discussion; see notes below. Also still needs tests.

### Resolves

> >  I was having some difficulty "merging" several `onchange` handlers, i.e. adding one to a block that already uses one as part of another extension.
> >
> > _Originally posted by @laurensvalk in https://github.com/google/blockly/issues/7483#issuecomment-1718043598_        
>
> Yeah there's not a good way to do this at the moment :/ we have similar problems with other mixins.

_Originally posted by @BeksOmega in https://github.com/google/blockly/issues/7483#issuecomment-1718284852_
            

### Proposed Changes (and proposed next step)

Implement `Block.addOnChange`. Similar to `Block.setOnChange` but adds them to a list, and later calls them in the same order they were added.

`Block.setOnChange` is kept to preserve backwards compatibility, though some duplicate code could be merged. The handler set by `setOnChange` could always be called first (even if it was set later) for backwards compatibility.

In practice, this means users can easily enforce the order when they fill the `extensions[]` list on the block specification.

### Reason for Changes

When a block already has an `onChange` handler, calling `setOnChange` from another extension currently overrides it.

This makes it difficult to keep extensions small and local, i.e. an extension that deals with a field's final value as discussed in #7483, combined with some other extension on the block.

Similarly, it means `onChange` handlers can't easily be used with blocks that already use them internally, like those in block-shareable-procedures.

### Test Coverage

TBD

### Documentation

TBD

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -213,6 +213,17 @@ export class Block implements IASTNodeLocation, IDeletable {
   private onchangeWrapper_: ((p1: Abstract) => void) | null = null;
 
   /**
+   * Collection of unbound onchange handlers added with addOnChange().
+   */
+  private onChangeHandlers_: ((p1: Abstract) => void)[] = [];
+
+  /**
+   * A bound callback function that calls the onchange handlers added with
+   * addOnChange() when the parent workspace changes.
+   */
+  private onChangeHandlersCaller_: ((p1: Abstract) => void) | null = null;
+
+  /**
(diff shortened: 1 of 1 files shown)


laurensvalk:
It seemed useful more generally since you mentioned running into the same problem, but I'm happy to try it another way.

Issue #7488, opened 2023-09-13
Update webdriverio and remove selenium-standalone package


rachel-fenichel:
[As of  v8.14.0](https://webdriver.io/blog/2023/07/31/driver-management) webdriverio handles drivers internally. That means we will be able to get rid of `selenium-standalone` and `@wdio/selenium-standalone-service` once we update.

Issue #7494, opened 2023-09-16
Flaky test: children connected to statement inputs are disabled when the parent is disabled.


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

One of the browser tests is flaky in a way that suggests a race condition.

The test is: 
```
test(
    'children connected to statement inputs are disabled when the ' +
      'parent is disabled',
    async function () {
      const parent = await dragBlockTypeFromFlyout(
        this.browser,
        'Logic',
        'controls_if',
        10,
        10,
      );
      const child = await dragBlockTypeFromFlyout(
        this.browser,
        'Logic',
        'controls_if',
        110,
        110,
      );
      await connect(this.browser, child, 'PREVIOUS', parent, 'DO0');

      await contextMenuSelect(this.browser, parent, 'Disable Block');

      chai.assert.isTrue(await getIsDisabled(this.browser, child.id));
    },
  );

```

`contextMenuSelect` calls `clickBlock`, which causes this code to run:

```
  // In the test context, get the Webdriverio Element that we've identified.
  const elem = await browser.$(`#${findableId}`);
  await elem.click(clickOptions);
```

In some situations the right-click on the parent block silently fails. Then the test fails because the `Disable Block` option was not found.

Adding a `waitForExist` after clicking the block makes the flakiness go away for me but not for @NeilFraser (at last check). Running the tests with the browser log level set to anything above `warn` makes the flakiness go away for me. So does adding a `console.log` to print the clickable element after finding it but before clicking it. 

The combination of the three previous items makes me suspect a race condition. I have not been able to figure out why it occurs only in this case.

The [webdriverio click documentation](https://webdriver.io/docs/api/element/click) says "This issues a WebDriver click command for the selected element , which generally scrolls to and then clicks the selected element when no options are passed. When options object is passed it uses action class instead of webdriver click which give added capabilities like passing button type, coordinates etc. By default, when using options a release action command is send after performing the click action, pass option.skipRelease=true to skip this action."

In this code we are passing `{button: 2}` as an options object to `click`, so it's running a sequence of [actions](https://webdriver.io/docs/api/browser/action) instead of webdriver click. I have not tested passing `option.skipRelease=true` and then manually releasing the action. That's next on my debug list, but I don't see why that would fix it.

There are three tests in this suite, all of which have the same structure. Only this test regularly fails. On computers where it fails, changing the order of the tests or using `.only` to only run this test do not change the result.

### Reproduction steps

[...]

PR #7500, opened 2023-09-18
chore(deps): update webdriverio to 8.16 and remove selenium-standalone

rachel-fenichel:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7488

### Proposed Changes

- Update `webdriverio` from `v8.0.5` to `v8.16.7`
- Remove dependency on `@wdio/selenium-standalone-service`
- Remove dependency on `selenium-standalone`
- Remove `services: [['selenium-standalone']]` stanza from all webdriver test setup files
  - This code is duplicated in several different tests, but changing that is out of scope for this PR
- Remove `tests/scripts/selenium-config.js` because webdriverio is now in charge of downloading or finding the drivers
- Remove `tests/scripts/setup_osx_env.sh` because it just specified the location of chrome, which we no longer need
  - We actually aren't running tests on os X anyway (https://github.com/google/blockly/issues/2114).
- Add some `Math.round` calls to deal with non-integer pixel values.
  - Required because webdriverio got stricter and I missed pulling it into https://github.com/google/blockly/pull/7493

### Reason for Changes

[As of v8.14](https://webdriver.io/blog/2023/07/31/driver-management), webdriverio is in charge of finding, downloading, and setting up browser drivers.

### Test Coverage

Tested locally by building and running tests, and remotely by pushing it up to github and making sure all actions complete. I also ran `npm run test:browser` locally.

https://github.com/google/blockly/issues/7494 still applies sometimes

### Additional Information

One open question: which version of chrome should we test on? Which version of firefox?

https://webdriver.io/blog#say-goodbye-to-driver-services

Currently we aren't specifying the version. I could drop in stable to make it regularly update, or pin to a specific version to force consistency.


tests/browser/test/test_setup.js
```diff
@@ -367,8 +366,8 @@ async function connect(
   );
 
   const delta = {
-    x: targetLocation.x - draggedLocation.x,
-    y: targetLocation.y - draggedLocation.y,
+    x: Math.round(targetLocation.x - draggedLocation.x),
+    y: Math.round(targetLocation.y - draggedLocation.y),
```
cpcallen, line 370: Can you add an explanation for this change to the PR description please?  It seems unrelated to the config changes but… is needed to get certain tests to pass in certain circumstances?

cpcallen (approved):
> One open question: which version of chrome should we test on? Which version of firefox?
> 
> Currently we aren't specifying the version. I could drop in stable to make it regularly update, or pin to a specific version to force consistency.

* What is the effect of not specifying—does it just use whatever version one happens to have installed?  Since that was AFAIK what was happening before that seems reasonable for now.
* I think using a tag like `stable` is preferable to pinning to a particular version that will become increasingly obsolete (or create work for us to keep bumping).
* `canary` is probably more useful than `stable`, insofar as if Chrome were to break something in Blockly we'd find out about it in advance.

rachel-fenichel:
As far as I can tell the default is stable if not specified.

tests/browser/test/test_setup.js
rachel-fenichel: Done. As I added above:

Add some Math.round calls to deal with non-integer pixel values.
Required because webdriverio got stricter and I missed pulling it into https://github.com/google/blockly/pull/7493

Issue #7480, opened 2023-09-11
Modifying a block comment does not fire a block change event


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

If you type into a block comment, it does not fire a change event. It should.

### Reproduction steps

1. Run `npm run start` to open up the playground.
2. Enable workspace event logging by clicking the checkbox on the left.
3. Load the following JSON into the workspace. By pasting it into the textbox and hitting the "Load" button.
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "controls_if",
        "id": "GU6;.7|eVLgZ)A0#JvJv",
        "x": 62,
        "y": 87,
        "icons": {
          "comment": {
            "text": "a comment",
            "pinned": true,
            "height": 80,
            "width": 160
          }
        }
      }
    ]
  }
}
```
4) Open the browser console and clear it.
5) Type into the comment.
6) Observe how no events are fired.

### Additional Info

I'm going to add this as a "good first issue" because I think it is a good one, but I'm not tagging it with "ghc-osd" since we have enough now =)

## To Fix

In the comment icon's [onTextChange method](https://github.com/google/blockly/blob/07fec204bf66589625ad7816a248cf2ae9b5c3b6/core/icons/comment_icon.ts#L219
) fire a change event. You can work off of the example [here](https://github.com/google/blockly/blob/07fec204bf66589625ad7816a248cf2ae9b5c3b6/core/block.ts#L2195).

treypisano:
Hi, can I take a shot at resolving this?

BeksOmega:
@treypisano Go for it =) I've assigned you!

treypisano:
Hi @BeksOmega , I just had a few questions about this issue. I believe that I got the [onTextChange method](https://github.com/google/blockly/blob/07fec204bf66589625ad7816a248cf2ae9b5c3b6/core/icons/comment_icon.ts#L219) to fire the block change event, but there is an issue. This method is only firing when a comment is changed and then there is a MouseDown event, **not** when the text itself is changed. Should this be separated into another issue?

Also, I'm not sure if the old value of the comment is saved (pre-change), so I am just leaving BlockChange.oldValue as null at the moment.

Thanks

Issue #7432, opened 2023-08-24
First block from flyout matches ID and subsequent ones do not


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When you drag a block from the flyout for the first time, the block has the same ID as the block in the flyout. For subsequent blocks it has a different ID.

The first block has the same ID because we are serializing the flyout block and deserializing it into the main workspace (ids are included in serialization). Subsequent blocks do not have the same ID, because during deserialization the ID conflicts with the block that already exists in the workspace, so it gets a new ID.

This causes problems with unit tests because we have multiple blocks on the page that have the same `[data-id]` attribute.

Instead it would be nice if all blocks dragged from the flyout have a different ID than their originator.

We should add a `saveIds` parameter to `Blockly.serialization.blocks.save` that allows us to save blocks without IDs. If we save the blocks in the flyout like this and then deserialize them, we ensure that the IDs are always different.

### Reproduction steps

1. Run `npm run start` from the root of the repository.
2. Switch from "Category toolbox" to "Simple toolbox"
3. Drag the if-else block from the flyout.
4. Inspect the page for both the block in the flyout and the block on the main workspace. Observe how the `data-id` attribute is the same for both.

## To Fix

1) Run the reproduction steps and make sure you can reproduce the issue.
2) Add a new optional `addIds` parameter to the [save function](https://github.com/google/blockly/blob/344861698e65c3fc5f05db000c01cff2155f56ff/core/serialization/blocks.ts#L82) that defaults to `true`.
3) Change the logic to only save the Id if `addIds` is true.
4) Pass the parameter through all the recursive calls. Namely `saveInputBlocks` and `saveNextBlocks`.
5) Modify the [flyout code](https://github.com/google/blockly/blob/344861698e65c3fc5f05db000c01cff2155f56ff/core/flyout_base.ts#L1216) to pass `addIds: false` to `save`.
6) Run the reproduction steps again and make sure the blocks have different IDs.

bsampada7:
I would like to work on this, if I don't get assigned for the first one

KrautRock:
@BeksOmega I am interested in working on this, thanks!

rl-dhanush:
@BeksOmega I am interested in working on this, thanks !!

patelb268:
@maribethb I would like to work on these issue

Yitao-Wang01:
@BeksOmega I am interested in working on this, thanks!

[...]

PR #7505, opened 2023-09-19
fix: comment change event fires block change event

treypisano:
## The basics

### Resolves

Fixes [Issue #7840](https://github.com/google/blockly/issues/7480). 

### Proposed Changes

This Pull Request makes a change to the [onTextChange](https://github.com/google/blockly/blob/07fec204bf66589625ad7816a248cf2ae9b5c3b6/core/icons/comment_icon.ts#L219) method in the CommentIcon class. When this method is called, a block change event is fired.

In order to satisfy the [oldText](https://github.com/google/blockly/blob/develop/core/events/events_block_change.ts#L54) property of the BlockChange event, I added a private field called [oldText](https://github.com/treypisano/blockly/blob/block-change-event-fix/core/icons/comment_icon.ts#L58) to to the [CommentIcon](https://github.com/treypisano/blockly/blob/block-change-event-fix/core/icons/comment_icon.ts#L38) class. This is by default set to an empty string, and is set at the **end** of every time a change event is fired from the comment. That means that oldText will always either be an empty string, or the text from the last time a change event was fired.

### Reason for Changes

N/A
### Test Coverage

I tested this by putting manually creating a block and adding a comment, and then changing that comment, with log workspace events enabled. The onTextChange method then runs, and then logs the Event object to the console.

### Documentation

Yes, the oldText class field would need to be added to the _CommentIcon_ class.

### DIFF
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -155,10 +155,7 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
 
   /** Sets the text of this comment. Updates any bubbles if they are visible. */
   setText(text: string) {
-    let oldText: null | string = null;
-    if (this.bubbleIsVisible()) {
-      oldText = this.text;
-    }
+    const oldText = this.text;
     eventUtils.fire(
       new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
         this.sourceBlock,
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -38,7 +38,7 @@ const DEFAULT_BUBBLE_HEIGHT = 80;
 export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   /** The type string used to identify this icon. */
   static readonly TYPE = IconType.COMMENT;
-
+  
   /**
    * The weight this icon has relative to other icons. Icons with more positive
    * weight values are rendered farther toward the end of the block.
@@ -219,6 +219,15 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   onTextChange(): void {
     if (this.textInputBubble) {
       this.text = this.textInputBubble.getText();
+      eventUtils.fire(
+        new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
+          this.sourceBlock,
+          'comment',
+          null,
+          null,
+          this.text,
+        ),
+      );
     }
   }
 
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -218,14 +218,16 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
    */
   onTextChange(): void {
     if (this.textInputBubble) {
-      this.text = this.textInputBubble.getText();
+      const newText = this.textInputBubble.getText();
+      this.text = newText
+      const sourceBlock = this.sourceBlock
       eventUtils.fire(
         new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
-          this.sourceBlock,
+          sourceBlock,
           'comment',
           null,
           null,
-          this.text,
+          newText,
         ),
       );
     }
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -155,9 +155,9 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
 
   /** Sets the text of this comment. Updates any bubbles if they are visible. */
   setText(text: string) {
-    let oldText: null | string = null
+    let oldText: null | string = null;
     if (this.bubbleIsVisible()) {
-      oldText = this.text
+      oldText = this.text;
     }
     eventUtils.fire(
       new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
@@ -242,7 +242,7 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
             newText,
           ),
         );
-      }     
+      }
       this.text = newText;
     }
   }
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -232,8 +232,7 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   onTextChange(): void {
     if (this.textInputBubble) {
       const newText = this.textInputBubble.getText();
-      
-      if (this.textBubble) {
+      if (this.textInputBubble.getText() !== this.text) {
         eventUtils.fire(
           new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
             this.sourceBlock,
@@ -243,8 +242,7 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
             newText,
           ),
         );
-      }
-      
+      }     
       this.text = newText;
     }
   }
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -230,21 +230,21 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
    * the input bubble.
    */
   onTextChange(): void {
-    if (this.textInputBubble) {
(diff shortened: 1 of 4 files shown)


core/icons/comment_icon.ts
```diff
@@ -54,6 +54,9 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   /** The text of this comment. */
   private text = '';
 
+  /** The text of this comment before the most recent change event. */
+  private oldText = '';
```
BeksOmega, line 58: I don't think you need to add a separate property for this. You should be able to use the current text as the old text before assigning the new text.

Something like:
```
const newText = this.textInputBubble.getText();
eventUtils.fire(this.text, newText);  // Actually use the right signature :P
this.text = newText;
```

Could you try that and verify that it behaves correctly?

core/icons/comment_icon.ts
```diff
@@ -218,7 +221,18 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
    */
   onTextChange(): void {
     if (this.textInputBubble) {
-      this.text = this.textInputBubble.getText();
+      const newText = this.textInputBubble.getText();
```
BeksOmega, line 224: Just realized we should also update `setText` to fire an event. Could you do that as well?

BeksOmega (changes requested):
This is looking great! Thank you for the change :D Just a few fixes and then should be good to go!

core/icons/comment_icon.ts
treypisano: Yea, I can get both of these things resolved!

treypisano: Alright @BeksOmega, should be good to go!

BeksOmega: Oh hm my edit to that comment didn't save, sorry about that. `setText` fixes look great! But we also need to make a little change to [`setCommentText`](https://github.com/google/blockly/blob/12ac35829e11417e5acda818949fa35a15e9ac74/core/block.ts#L2224) to only fire the event if the icon is removed. Otherwise we're going to get duplicate events!

Once that's in, I'll give this a manual test, and hopefully should be good to go =)

Thank you for your work on this @treypisano !

treypisano: Ok I think I got it. Thanks for helping me out with this.

treypisano:
I spent some time looking into why these aren't passing. As far as I can tell, the reason the [first test](https://github.com/google/blockly/blob/4a6d4399d1c5d41f2632deb08e7badc59aae22b2/tests/mocha/block_test.js#L1336C1-L1336C1) is failing is because [`setCommentText`](https://github.com/google/blockly/blob/4a6d4399d1c5d41f2632deb08e7badc59aae22b2/core/block.ts#L2207C10-L2207C10) is setting the text of its child comment icon to the new text. 

Then, the event is triggered in [`onTextChange`](https://github.com/google/blockly/blob/4a6d4399d1c5d41f2632deb08e7badc59aae22b2/core/block.ts#L2207C10-L2207C10) except the oldValue that is being sent in the BlockChange event has already been set as the **new value** by `setCommentText`.  

Hope this helps.

treypisano:
Sure, I will look into that.

core/icons/comment_icon.ts
```diff
@@ -155,6 +155,19 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
 
   /** Sets the text of this comment. Updates any bubbles if they are visible. */
   setText(text: string) {
+    let oldText: null | string = null;
+    if (this.bubbleIsVisible()) {
+      oldText = this.text;
+    }
```
BeksOmega, line 161: I'm confused about why this is only set when the bubble is visible. Shouldn't the old text always be the current text?

core/icons/comment_icon.ts
```diff
@@ -218,7 +231,19 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
...
+          new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
+            this.sourceBlock,
+            'comment',
+            null,
+            this.text,
+            newText,
+          ),
+        );
+      }
+      this.text = newText;
```
BeksOmega, line 246: This can go inside of the above `if`. No reason to set it if it's allready the same!

It might be mor readable if you restructure this to use early returns.

```
if (!this.textInputBubble) return;

const newText = this.textInputBubble.getText();
if (this.text === newText) return;

eventUtils.fire(
  new (eventUtils.get(eventUtils.BLOCK_CHANGE))(
    this.sourceBlock,
    'comment',
    null,
    this.text,
    newText,
  ),
);
this.text = newText;
```

BeksOmega (changes requested):
This is lookin great! Just one question and one request and then I'll give this a manual test :D

Thank you for your hard work on this! Really appreciate the back and forth =)

core/icons/comment_icon.ts
treypisano: This is here to address the fact that the [tests](https://github.com/google/blockly/blob/develop/tests/mocha/block_test.js#L1277C11-L1277C11) are expecting the default value of `oldText` to be `null`, not an empty string. 

I believe this is due to the fact that they were written expecting that the Change Event would come from `setCommentText`, which [defaults to null](https://github.com/treypisano/blockly/blob/f8a134caef1aa57cbccc429efff8ec5a561c1b77/core/block.ts#L2206) if the comment doesn't exist yet. But now that the event is coming from `setCommentText`, the default `oldText` is `this.text`, which is `''`.

It might make sense to have the tests to expect `''` as a default `oldValue`, rather than `null`.

BeksOmega: Mmm darn ok. We can't change the `oldValue` to be `''` because external devs could be checking for an `oldValue` of `null` to tell if the comment was just added.

In that case, I don't think this code quite works. If you open the playground, enable event logging, add a block to the workspace, and then run the following two commands in the browser console:
```
Blockly.getMainWorkspace().getTopBlocks()[0].setCommentText('first text');
Blockly.getMainWorkspace().getTopBlocks()[0].setCommentText('second text');
```
You'll see that both events have an `oldValue` of `null`, when the second event should have an `oldValue` of `'first text'`.

So I think you need to fire the event for creating a comment from `setCommentText`. This probably means you'll need to call `Blockly.Events.disable` and `Blockly.Events.enable` around the call to `comment.setText`. But I'm not sure so you'll have to check!

Sorry for all the back and forth on this issue :/ I didn't realize it was going to be so complex when I filed it! but we want to make sure the behavior is correct for developers using the library =)

treypisano:
Alright, so I think that disabling events when `setText` is called from `setCommentText` should do the trick in terms of preventing duplicate events. So the event is now being called from `setCommentText` like it was originally. 

I manually tested out the workspace commands as well, they are firing the with the right values for `oldText` and `newText`.  Also, using `setText` on an icon itself is also working as expecting and firing a change event.

As far as the back and forth go, it's no problem! I'm learning a lot as I go along. I really appreciate the guidance as well.

treypisano:
Sorry about that, I forgot to format, and it seems that the event needs to be fired at the end of the function to work as desired

BeksOmega (approved):
Manually tested and this looks correct!

Tested:
* Calling `setCommenText` to add a comment icon.
* Calling `setCommentText` to modify a comment icon.
* Calling `setText` to modify a comment icon.
* Calling `setCommentText(null)` to remove a comment icon.

All events look correct.

Thank you so much @treypisano this looks great :D Let me know if you'd like me to point you to another issue. Otherwise you can check out our [good first issues](https://github.com/google/blockly/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22). I'll be marking some that were saved for OSD as public on Friday =)

treypisano:
Hi @BeksOmega , thanks for merging this. Not sure if this is the right place to ask, but if you have another issue you'd like to point me to that would be great.

BeksOmega:
Heya @treypisano Thanks for your interest :D This one seems like it might be a good fit: https://github.com/google/blockly/issues/7432

Issue #6976, opened 2023-04-13
RTL input field alignment is off


maribethb:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

See screenshot below. In RTL mode only fields are misaligned. I see this is true in the multi_playground going back to at least blockly-v9.2.1

### Reproduction steps

1. 
2. 
3. 

### Screenshots

<img width="280" alt="Screenshot 2023-04-13 at 11 34 07 AM" src="https://user-images.githubusercontent.com/8573958/231853085-89b64971-9556-4c60-a560-2b235fd7fb7b.png">

maribethb:
bisect says https://github.com/google/blockly/pull/6702 is the culprit. still looking into why this would matter only on RTL

maribethb:
This happens because we need to position the WidgetDiv (the HTMLTextInput) *after* the block is done rendering, which now happens on a delay due to the render queuing system.

So you can fix the issue by waiting until the render is done like so (in field_input):

```js
  private onHtmlInputChange_(_e: Event) {
    this.setValue(this.getValueFromEditorText_(this.htmlInput_!.value));
    requestAnimationFrame(() => {setTimeout(() => {
      this.resizeEditor_();
    }, 0);
  });
```

What I don't know is why you actually need to wait, since `resizeEditor` is already called during the `render` code. So just adding this bit of code is definitely inefficient, because we're resizing the editor twice, once during the render and once again immediately afterward. But when it's called during the `render` cycle, it has the wrong data and I'm not yet sure why, needs more investigation.

Also, this isn't the first issue caused by the general problem of needing to do additional rendering work after a block is rendered. So we might want to design a more elegant solution to that instead of just throwing more `setTimeout` calls whenever one of these pops up.

maribethb:
As for why the position values are different when doing it during render vs after, they are different all the way to the `[getBoundingClientRect](https://github.com/google/blockly/blob/develop/core/utils/style.ts#L119)` call inside `[getScaledBbox](https://github.com/google/blockly/blob/develop/core/field.ts#L871)`

I am not sure if this is because the `getBoundingClientRect` call is cached from the previous frame, or because the values genuinely change afterwards due to other rendering of the block. I read that you can invalidate the cache by writing to the dom before reading to it (note this is generally advised against however), so I tried that but still had the same issue, so I am guessing the latter. 

 Anyway gonna hold off on this further until we decide what to do about the broader issue.

Issue #6915, opened 2023-03-21
Soft keyboard issues on Android


maribethb:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

If you disable modal inputs on mobile, then trying to edit number or text fields on Android devices will fail. The soft keyboard will pop up and then close immediately.

The cause of this is that when Android opens the keyboard, it fires a resize event. When Blockly, hears a resize event, it will hideChaff. This includes closing the widget div, which is the html element that should have focus that the user is typing into. Since it goes away, so does the soft keyboard. This issue does not occur on iOS because iOS does not fire resize events when the keyboard appears. It is an overlay on top of the browser content and the user (or app) is responsible for scrolling the content if desired.

#### Proposed option
During the window resize window event listener, do not hide all the chaff. Only hide a subset of it:

- Do not hide widget div. This solves the immediate problem and matches what Scratch does. This requires adding logic to reposition the widget div during a resize, which I have borrowed from Scratch and tested on multiple devices.
- Do not hide dropdown div. We already have the logic to reposition it correctly (I've tested it) and this would prevent this bug from occurring if there was a dropdown div with an html input in it (which I don't know of any right now, but doesn't seem that far fetched, like a combobox)
- Tooltip: hide (though I cannot think of a way to resize a window while a tooltip is triggered, better safe than sorry)
- Registrable components: hide (it's a shame that there is no way for components to opt out of this hiding the way we're doing for the built-in divs, but that's a feature we could implement if somebody asks, for now, YAGNI). we should hide them by default because they may not know how to reposition themselves during a resize. and because they may have been designed knowing that they would be hidden whenever `hideChaff` is called.

#### Alternatives considered
If you search the web for this issue (Android firing a resize event when showing the keyboard) you'll find a few solutions:

- In the resize event handler, check if the currently focused element is an html input element. if so, then assume this is a spurious android resize triggered by the keyboard, then skip the rest of the resize logic. This doesn't work because you might legitimately resize the browser (or rotate your device) while the input is active, and if you did then things would be positioned badly
- On mobile devices, make the current `resize` handler an `orientationchange` handler instead. That way, nothing happens during the resize event that Android fires. This is not a good idea because as mobile devices improve there may be ways to change the browser or window size without rotating the device (e.g. a tablet that can use windowed apps). Also, it adds complexity to change the behavior between mobile and desktop.

Other options:

[...]

PR #7507, opened 2023-09-21
fix: widget positioning

### LINKED ISSUE #7491: Modifying the structure of a block does not trigger widgets to be repositioned
### Check for duplicates

### Description

If a warning icon is added or removed during field editing, the field is not rendered correctly.

related to the rendering queue and might be related to #7068 ?

### Reproduction steps

1. Open Block Factory
2. Drag two value input blocks into the inputs input. Notice how they each have a warning now due to conflicting names.
3. Change the name of one of the inputs. The warning disappears, and the underlying block is resized. But the text input html element is now in the wrong place because it did not move when the block was resized.
4. Click outside the field to end editing / close the widget div.
5. Change the name back to `NAME`. The warning reappears, causing the same problem again.

### Stack trace

_No response_

### Screenshots

Screenshot 2023-09-14 at 5 39 53 PM

### Browsers

_No response_

BeksOmega:
## The basics

### Resolves

Fixes #7491

### Proposed Changes

Reorganizes a bunch of code, and makes it so that fields get `onLocationChange` call backs.

### Reason for Changes

Makes it so that fields can reposition their widgets properly.

### Test Coverage

* Tested that https://github.com/google/blockly/issues/6976 does not reproduce
* Tested that changing the window size properly repositions the div, re: https://github.com/google/blockly/issues/6915
* Tested that #7491  does not reproduce
* Tested that we did not get a significant performance drop by dragging spaghetti blocks, and also typing into the spaghetti blocks. We might have gotten a slight performance dip from splitting up the connection update operations so they're not happening contiguously (the engine can't optimize as much). But all of this code is internal so I can circle back once I've had more time to investigate.

[old-dragging.txt](https://github.com/google/blockly/files/12691991/old-dragging.txt)
[new-dragging.txt](https://github.com/google/blockly/files/12691992/new-dragging.txt)

### Documentation

N/A

### Additional Information

Might want to hold until after release?

## Deprecations

- [ ] TODO

### DIFF
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -617,7 +617,6 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     }
     const div = WidgetDiv.getDiv();
     const bBox = this.getScaledBBox();
-    console.log(bBox);
     div!.style.width = bBox.right - bBox.left + 'px';
     div!.style.height = bBox.bottom - bBox.top + 'px';
 
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -252,6 +252,14 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
   }
 
   /**
+   * Notifies the field that it has changed locations. Moves the widget div to
+   * be in the correct place if it is open.
+   */
+  onLocationChange(): void {
+    if (this.isBeingEdited_) this.resizeEditor_();
+  }
+
+  /**
    * Updates the colour of the htmlInput given the current validity of the
    * field's value.
    *
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -33,7 +33,6 @@ import {Coordinate} from './utils/coordinate.js';
 import * as userAgent from './utils/useragent.js';
 import * as WidgetDiv from './widgetdiv.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
-import * as renderManagement from './render_management.js';
 import {Size} from './utils/size.js';
 
 /**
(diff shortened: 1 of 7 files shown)


BeksOmega:
Note to self to ping code.org people about this re: their flyout field if it gets approved. I think they can use this new callback to handle the width of their field less jankily. (Since we call `onLocationChange` after every render.)

core/field_input.ts
```diff
@@ -616,6 +617,7 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     }
     const div = WidgetDiv.getDiv();
     const bBox = this.getScaledBBox();
+    console.log(bBox);
```
BeksOmega, line 620: Remove.

core/field_input.ts
```diff
@@ -642,13 +644,13 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
 
-    bumpObjects.bumpIntoBounds(
+    const bumped = bumpObjects.bumpIntoBounds(
       this.workspace_!,
       this.workspace_!.getMetricsManager().getViewMetrics(true),
       block,
     );
 
-    this.resizeEditor_();
+    if (!bumped) this.resizeEditor_();
```
maribethb, line 653: Why do you only need to resize the editor if it wasn't bumped? Should it be the opposite?

BeksOmega: If we trigger a move of the block (the block gets bumped), it triggers `onLocationChange` for all fields, so the field gets bumped that way. If we don't move the block, we still need to reposition the widget because the block might be moving relative to the top-left of the window (even though it's not moving relative to the workspace origin).

Here's what it looks like if I remove the `if (!bumped) this.resizeEditor_();` line:
Screen recording 2023-10-26 9 31 51 AM

You can see that if we do bump, repositioning is handled, but if we don't it is not handled.

maribethb (approved):

Issue #7176, opened 2023-06-16
Add an API to programmatically scroll the workspace


BeksOmega:
🚧⚠️🚧 This issue is being saved for **Grace Hopper Open Source Day**. 🚧⚠️🚧 An event we're running to help girls and non-binary technologists make their first contributions to open source! Please see our list of [help wanted issues](https://github.com/google/blockly/issues?page=1&q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22+-label%3A%22ghc-osd%22) if you'd like to contribute! 

---

### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Sometimes you want to change how your users interact with scrolling the workspace. E.g. if you have a workspace that only supports vertical scrolling or only supports horizontal scrolling, you may want to add buttons that scroll instead of using the scrollbars.

Context: https://groups.google.com/g/blockly/c/klLbJ1aQqsA/m/lYjP5tyAAgAJ

### Request

Add an API for programatically scrolling the workspace. I'm not sure what the units should be (absolute vs relative, percentage vs workspace units, etc).

### Additional context

We currently have a scroll method on the workspace, but it is marked as `@internal`: https://github.com/google/blockly/blob/b8ad7d307f9f2abffe5306818155907287cfa634/core/workspace_svg.ts#L2179

## To Fix

1) Remove the `@internal` tag from the [scroll method](https://github.com/google/blockly/blob/b8ad7d307f9f2abffe5306818155907287cfa634/core/workspace_svg.ts#L2179).

maribethb:
Seems reasonable to make the `scroll` function public. We just need to remove the `@internal` tag from the documentation comment.

carthick:
Could you please assign me this issue.

maribethb:
Hi @carthick, I've assigned you the issue. You can find some information about how to contribute to Blockly [here](https://developers.google.com/blockly/guides/contribute/get-started). Thanks for your interest!

SophieNguyen113:
I would love to work on this issue

BeksOmega:
@SophieNguyen113 Go for it!

PR #7525, opened 2023-09-22
feat: Added an API to programmatically scroll the workspace

SophieNguyen113:
## The basics

### Resolves

Fixes 
https://github.com/google/blockly/issues/7176
### Proposed Changes

Added an API to programmatically scroll the workspace
Removed the @internal
All changes were in workspace_svg.ts file
### Reason for Changes

Make it easier for users to use the ToolBox and WorkSpace

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -2184,58 +2184,77 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
     return this.scale;
   }
 
+  // Add two new methods to WorkspaceSvg
+
+/**
+ * Scroll the workspace horizontally by a relative amount.
+ * @param {number} deltaX Pixels to scroll horizontally.
+ */
+workspace.scrollHorizontal(deltaX) {
+  // Use existing scroll method with updated x
+  this.scroll(this.scrollX + deltaX, this.scrollY); 
+}
+
+/** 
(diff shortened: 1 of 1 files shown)


core/workspace_svg.ts
```diff
@@ -2184,58 +2184,77 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
   }
 
-  /**
-   * Scroll the workspace to a specified offset (in pixels), keeping in the
-   * workspace bounds. See comment on workspaceSvg.scrollX for more detail on
-   * the meaning of these values.
-   *
-   * @param x Target X to scroll to.
-   * @param y Target Y to scroll to.
-   * @internal
```
BeksOmega, line 2194: Here's the tag we needed removed!

BeksOmega (commented):
Heya @SophieNguyen113 This change looks really cool! And we appreciate all the hard work you did on it :D

The change we were looking for was actually a bit smaller though. We want to remove the `@internal` tag from one of our existing functions so that developers can access it. We don't want to add new `scrollHorizontal` and `scrollVertical` methods right now, but maybe in the future!

Could you modify your change to just remove the `@internal` tag? You can look back at the issue for more info! https://github.com/google/blockly/issues/7176

Issue #6848, opened 2023-02-15
Paste block in viewport / near cursor


BeksOmega:
🚧⚠️🚧 This issue is being saved for **Grace Hopper Open Source Day**. 🚧⚠️🚧 An event we're running to help girls and non-binary technologists make their first contributions to open source! Please see our list of [help wanted issues](https://github.com/google/blockly/issues?page=1&q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22+-label%3A%22ghc-osd%22) if you'd like to contribute! 

---

### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When you paste a block, you expect the block to appear in the part of the workspace you are looking at. Usually either:
  * In the middle of the view
  * Near where your cursor is

Currently in Blockly, this is ignored, and blocks are always pasted near the original block that was copied.

### Reproduction steps

1. Copy a block to the clipboard using ctrl + c.
2. Scroll the workspace so that the original block is not in view.
3. Paste the bloc using ctrl + v.
4. Observe how the new pasted block is not in view, and was instead pasted near the original block.

### Additional Info

Pasting in the viewport may cause cascading block bumps for the pasted blocks, which would also be a bad user experience. I'm nopt sure if this occures with the current system or not.

Note: block bumping is when we move blocks away from other blocks connections so they don't accidentally appear to be connected.

## To Fix

1. Follow the reproduction steps to reproduce the bug.
2. Add a `getRelativeToSurfaceXY(): Coordinate` method to the [IMovable interface](https://github.com/google/blockly/blob/develop/core/interfaces/i_movable.ts).
3. Modify the [copy](https://github.com/google/blockly/blob/7e9d1eb3ba780f523863ba94b4af5cce36292295/core/shortcut_items.ts#L121) and [cut](https://github.com/google/blockly/blob/7e9d1eb3ba780f523863ba94b4af5cce36292295/core/shortcut_items.ts#L161) callbacks to store the result of `getRelativeToSurfaceXY` in a `copyCoords` variable.
4. Modify the [paste](https://github.com/google/blockly/blob/7e9d1eb3ba780f523863ba94b4af5cce36292295/core/shortcut_items.ts#L193) callback to:
    A. Run `copyWorkspace.getMetricsManager().getViewMetrics(true)` to get the coordinates of the viewport of the workspace.
    B. If the `copyCoords` are within the viewport, pass `copyCoords` to `clipboard.paste`.
[...]

Issue #752, opened 2016-11-21
Deletion of a highlighted block makes highlightBlock(id) stop working


ruslan-simonenko:
When highlighted block is deleted `highlightBlock` stops working.
```javascript
workspace.highlightBlock(id);
// delete that block
workspace.highlightBlock(anything); // Error: this.svgPath_ is undefined
workspace.highlightBlock(anything, true|false); // Works
```

NeilFraser:
Sorry about that.  My fault.  Fixed on develop branch in a66d274169540553c022b9f204a77b4589968db5.

(And if anyone on the team wants me to stop abusing my power with regards to pushing directly, just say so.)

PR #7531, opened 2023-09-22
fix: paste block in viewport/near cursor

adrijanag17:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/6848

### Proposed Changes

Pastes copied block in the center of viewport or near the original block depending on whether or not original is inside the viewport. 

### Reason for Changes

Copied blocks should not be pasted near original if original is not in viewport. The copied block should be visible in viewport.

### DIFF
--- a/core/interfaces/i_movable.ts
+++ b/core/interfaces/i_movable.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {Coordinate} from '../utils';
+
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
Hey @adrijanag17, this bug is already being worked on in https://github.com/google/blockly/pull/752.

If you comment on a different issue, Beka or Maribeth will get you assigned to avoid overlaps.

Issue #7523, opened 2023-09-22
Check for undefined in vertical flyout layout function


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

In [`layout_`](https://github.com/google/blockly/blob/osd/core/flyout_vertical.ts#L227) in `flyout_vertical.ts` we access the block with `const block = item.block`. The result is nullable, so all later uses of `block` have to have a [non-null assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-)

To fix:
- Check whether `block` is `undefined` after setting `const block = item.block`.
- `continue` if `block` doesn't exist.
- Remove `!` from after the remaining uses of `block` within the for loop.

Related to https://github.com/google/blockly/issues/7522

To get set up for this issue:
- Run `npm install` from the root of the repository.
- Run `npm run build` from the root of the repository.
- Run `npm run start` from the root of the repository to launch the test playground.

ivalkshfoeif:
Hi @rachel-fenichel, I'd like to work on this issue!

rashmi29:
Hi @rachel-fenichel , I would like to work on this issue. Could you please assign it to me?

11aparna91:
Hi @rachel-fenichel , I would like to work on this issue.

rachel-fenichel:
Hi all, and thanks for your interest! This is being worked on in the `blockly-elephant` breakout room.

PR #7535, opened 2023-09-22
fix : Check for undefined in vertical flyout layout function#7523

patelb268:
## The basics

- [X ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

## The details
In [layout_](https://github.com/google/blockly/blob/osd/core/flyout_vertical.ts#L227) in flyout_vertical.ts we access the block with const block = item.block. The result is nullable 

### Resolves
so all later uses of block have to have a [non-null assertion]

Fixes 
[Check for undefined in vertical flyout layout function#7523](https://github.com/google/blockly/issues/7523)
### Proposed Changes

Check whether block is undefined after setting const block = item.block.
continue if block doesn't exist.
Remove ! from after the remaining uses of block within the for loop.

### DIFF
--- a/core/flyout_vertical.ts
+++ b/core/flyout_vertical.ts
@@ -233,6 +233,9 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
+        if(block == undefined || block == null){
+          continue;
+        }
         const allBlocks = block!.getDescendants(false);
         for (let j = 0, child; (child = allBlocks[j]); j++) {
           // Mark blocks as being inside a flyout.  This is used to detect and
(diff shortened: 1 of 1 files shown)


core/flyout_vertical.ts
```diff
@@ -233,22 +233,25 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
+        if(block == undefined || block == null){
```
maribethb, line 236: You can simplify this line to just:

`if (block) {`

maribethb (changes requested):
Hi @patelb268,

Thank you for opening this pull request! It needs just a few things before it's ready to merge:
- Follow the google-cla bot's instructions to sign our CLA and refresh the check
- Address the comment left in the PR review below
- Run `npm run format` to automatically format the code to pass the rest of the CI checks

After that, things should be ready to merge! Thank you for your contribution!

maribethb:
Hi Prayag, another contributor also fixed this issue and we've merged that one. Sorry for the mixup, and thanks for the PR. I hope you enjoyed OSD!

PR #7536, opened 2023-09-22
fix: non-nullable check for block variable and removed ! in layout_ 

### LINKED ISSUE #7523: Check for undefined in vertical flyout layout function

rashmi29:
## The basics

### Resolves

Fixes #7523 

### Proposed Changes

- Check whether block is undefined after setting const block = item.block.
- continue if block doesn't exist.
- Remove ! from after the remaining uses of block within the for loop.

### Reason for Changes

Non nullable check was not present in code. That might break the code when variable has no value assigned.

### Test Coverage

I did npm run and tested all the buttons before and after and checked the entire functionality after my change.

### Documentation

No documentation required

### DIFF
--- a/core/flyout_vertical.ts
+++ b/core/flyout_vertical.ts
@@ -233,7 +233,7 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        if (block == null) {
+        if (!block) {
           continue;
         }
         const allBlocks = block.getDescendants(false);
--- a/core/flyout_vertical.ts
+++ b/core/flyout_vertical.ts
@@ -233,7 +233,7 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        if(block == undefined || block == null) {
+        if (block == undefined || block == null) {
           continue;
         }
         const allBlocks = block.getDescendants(false);
--- a/core/flyout_vertical.ts
+++ b/core/flyout_vertical.ts
@@ -233,29 +233,32 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        if(block == null || block == undefined) {
+          continue;
+        }
+        const allBlocks = block.getDescendants(false);
         for (let j = 0, child; (child = allBlocks[j]); j++) {
           // Mark blocks as being inside a flyout.  This is used to detect and
           // prevent the closure of the flyout if the user right-clicks on such
           // a block.
           child.isInFlyout = true;
         }
-        const root = block!.getSvgRoot();
-        const blockHW = block!.getHeightWidth();
-        const moveX = block!.outputConnection
+        const root = block.getSvgRoot();
+        const blockHW = block.getHeightWidth();
+        const moveX = block.outputConnection
           ? cursorX - this.tabWidth_
           : cursorX;
-        block!.moveBy(moveX, cursorY);
+        block.moveBy(moveX, cursorY);
 
         const rect = this.createRect_(
-          block!,
+          block,
           this.RTL ? moveX - blockHW.width : moveX,
           cursorY,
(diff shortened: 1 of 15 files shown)


rachel-fenichel:
Thanks for grabbing this issue!

Next steps:
- Run `npm run format` to fix some formatting issues and commit
- Sign the CLA, following the instructions from the bot above

I'll keep an eye on this PR and approve workflow runs as needed.

core/flyout_vertical.ts
```diff
@@ -233,29 +233,32 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        if(block == undefined || block == null) {
```
rachel-fenichel, line 236: This can be just `if (block) {`

rachel-fenichel (changes requested):

core/flyout_vertical.ts
```diff
@@ -233,29 +233,32 @@ export class VerticalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        if (block == undefined || block == null) {
```
rachanavishwanath, line 236: please squash your commits

rachanavishwanath, line 236: This check is redundant, block == null checks both null and undefined. We would both checks with strict equality i.e. block === undefined || block === null

rachel-fenichel: In this case `if (block)` will do fine, because we aren't trying to distinguish between the two cases.

rashmi29: If I just add if (block) than it shows "'block' is possibly 'undefined'." error. So, I added condition if (block == null) as suggested by rachna.

rachel-fenichel:
Hi @rashmi29,

I'm signing off for the weekend but I will take a look again on Monday. Thanks again for contributing, and have a great Grace Hopper if you're attending!

humera811:
nit: In the description section under `Resolves` you can add a link to the issue
```
  Resolves 
            fixes #7523 : Check for undefined in vertical flyout layout function 
```

core/flyout_vertical.ts
```diff
@@ -233,29 +233,32 @@ export class VerticalFlyout extends Flyout {
...
+        if (block == null) {
+          continue;
+        }
+        const allBlocks = block.getDescendants(false);
         for (let j = 0, child; (child = allBlocks[j]); j++) {
           // Mark blocks as being inside a flyout.  This is used to detect and
           // prevent the closure of the flyout if the user right-clicks on such
           // a block.
           child.isInFlyout = true;
         }
```
humera811, line 245: Suggestion:
```
if (block) {
          const allBlocks = block.getDescendants(false);
          for (let j = 0, child; (child = allBlocks[j]); j++) {
            // Mark blocks as being inside a flyout.  This is used to detect and
            // prevent the closure of the flyout if the user right-clicks on such
            // a block.
            child.isInFlyout = true;
          }
```
or we could also do 
```
if (!block) continue;
const allBlocks = block.getDescendants(false);
      for (let j = 0, child; (child = allBlocks[j]); j++) {
        // Mark blocks as being inside a flyout. This is used to detect and
        // prevent the closure of the flyout if the user right-clicks on such
        // a block.
        child.isInFlyout = true;
      }
```

maribethb (approved):
Looks great! Thank you so much for the PR!

PR #7537, opened 2023-09-22
fix: non-nullable check for block variable and removed ! in layout_ #7522

### LINKED ISSUE #7522: Check for undefined in horizontal flyout layout function
### Check for duplicates

### Description

In [`layout_`](https://github.com/google/blockly/blob/osd/core/flyout_horizontal.ts#L258) in `flyout_horizontal.ts` we access the block with `const block = item.block`. The result is nullable, so all later uses of `block` have to have a [non-null assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-)

To fix:
- Check whether `block` is `undefined` after setting `const block = item.block`.
- `continue` if `block` doesn't exist.
- Remove `!` from after the remaining uses of `block` within the for loop.

To get set up for this issue:
- Run `npm install` from the root of the repository.
- Run `npm run build` from the root of the repository.
- Run `npm run start` to launch the test playground.

priya-rjb:
## The basics

### Resolves

Fixes #7522 https://github.com/google/blockly/issues/7522

### Proposed Changes

Fixes flyout_horizontal.ts In [layout_] where we access the block with const block = item.block. The result was nullable, so all later uses of block have to have a [non-null assertion].
Screen Shot 2023-09-22 at 18 30 45

### Reason for Changes

To avoid writing the non-null assertion, to present the horizontal bar with the categories(blocks) in an appropriate manner. 

### Test Coverage

We changed the variable horizontalLayout to true in playground.html to test the changes.

### Documentation

We recommend updating the test coverage documentation in playground.html.

### DIFF
--- a/core/flyout_horizontal.ts
+++ b/core/flyout_horizontal.ts
@@ -267,11 +267,11 @@ export class HorizontalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        // CHANGE: Checked whether block is undefined:
-        if (typeof block === "undefined") {
+
+        if (block === undefined || block === null) { {
           continue;
         }
-        // CHANGE: Removed ! from after the remaining uses of block within the for loop.
+
         const allBlocks = block.getDescendants(false);
 
         for (let j = 0, child; (child = allBlocks[j]); j++) {
--- a/core/flyout_horizontal.ts
+++ b/core/flyout_horizontal.ts
@@ -268,7 +268,7 @@ export class HorizontalFlyout extends Flyout {
       if (item.type === 'block') {
         const block = item.block;
 
-        if (block === undefined || block === null) { {
+        if (block === undefined || block === null) {
           continue;
         }
 
--- a/core/flyout_horizontal.ts
+++ b/core/flyout_horizontal.ts
@@ -267,29 +267,35 @@ export class HorizontalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        //CHANGE: Checked whether block is undefined:
+        if (typeof block === "undefined") {
+          continue;
+        }
+        //CHANGE: Removed ! from after the remaining uses of block within the for loop.
+        const allBlocks = block.getDescendants(false);
+
         for (let j = 0, child; (child = allBlocks[j]); j++) {
           // Mark blocks as being inside a flyout.  This is used to detect and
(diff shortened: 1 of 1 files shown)


yuchiiii:
Hi Priya! I tried running prettier and changed the double quotes of undefined in the if statement should work.
Screen Shot 2023-09-22 at 4 19 20 PM

yuchiiii:
Also, our pull request title should be in this format. Thank you very much for revising this!!
Screen Shot 2023-09-22 at 4 16 40 PM

core/flyout_horizontal.ts
```diff
@@ -267,29 +267,35 @@ export class HorizontalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+        // CHANGE: Checked whether block is undefined:
+        if (typeof block === "undefined") {
```
johnnesky, line 271: This will work if block is undefined, but if block is null then this won't catch it. Just to be safe, I recommend using:

if (block === undefined || block === null) {

core/flyout_horizontal.ts
```diff
@@ -267,29 +267,35 @@ export class HorizontalFlyout extends Flyout {
...
+          continue;
+        }
+        // CHANGE: Removed ! from after the remaining uses of block within the for loop.
```
johnnesky, line 274: You don't need to describe code changes as comments in the code. Just describe what the code does as you've written it. Any such comments will continue to exist after your code is merged and should describe the code as it currently exists, not how it differs from previous versions.

johnnesky (changes requested):

core/flyout_horizontal.ts
yuchiiii: Hi John, could you please also grant me the access to push code? Thank you very much!

johnnesky (approved):

core/flyout_horizontal.ts
```diff
@@ -267,29 +267,35 @@ export class HorizontalFlyout extends Flyout {
     for (let i = 0, item; (item = contents[i]); i++) {
       if (item.type === 'block') {
         const block = item.block;
-        const allBlocks = block!.getDescendants(false);
+
+        if (block === undefined || block === null) { {
```
johnnesky, line 271: This PR isn't passing tests right now, and I suspect this double opening curly brace after the "if" may be the culprit!

johnnesky (changes requested):

johnnesky (approved):

johnnesky:
Excellent, thanks for your hard work!

PR #7550, opened 2023-09-28
fix(build): Fix sourcemaps, re-update metadata

cpcallen:
## The basics

### Resolves

Fixes broken sourcemaps, where (e.g.) `blockly_compressed.js.map` did not contain the sources files `core/**/*.ts` inline.

### Proposed Changes

* Do not attempt to strip `$build$src` from variable names in Closure Compiler output by renaming the input files.
* Update metatada to reflect the larger file sizes that result.

### Reason for Changes

Apparently some developers prefer to debug using the original source code rather than the actual JavaScript their browser is running.  Seems foolish but who am I to judge?

### Test Coverage

* Manually checked sourcemaps in playground (in compressed mode) and in a plugin (after `npm link`ing).
* `npm test` passes.


BeksOmega:
Reverts https://github.com/google/blockly/pull/7473/commits/06d78af6a403370ed933d75fbe523860c15b8aec of https://github.com/google/blockly/pull/7473 (subsequent changes mean the reversion is a bit different than the original changes)

BeksOmega (approved):

PR #7555, opened 2023-09-28
fix(generators): Modify type signature of `forBlock`

cpcallen:
## The basics

### Resolves

Part of #6828; solves a problem with the typing of the second (`generator`) argument to block generator functions.

### Proposed Changes

Make two changes to the `forBlock` dictionary on CodeGenerator objects:

- Change the type of the second argument to the special [polymorphic `this` type](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-types) (was CodeGenerator)

- Change the way `forBlock` is declared from using an index signature to using the `Record<>` template type.  The two should be equivalent but [the former provokes type errors when used with the 'this' type](https://stackoverflow.com/q/77197616/4969945).

Additionally, mark the now-unused (but exported) `BlockGenerator` type alias as `@deprecated` pending removal in the next major version.

### Reason for Changes

Having the `forBlock` dictionary typed `(block: Block, generator: CodeGenerator) => /*...*/` creates some difficulties when trying to write block generator functions for `CodeGenerator` subclasses like `Javascript Generator):

* If the generator function is typed to match, with a second argument of type `CodeGenerator`, then the body would need to cast to call methods defined only on the subclass—e.g.: `(generator as JavascriptGenerator).getAdjusted(/*...*/)`.
* If the generator is typed with second argument of the specific subclass (e.g. `JavascriptGenerator`) then the function itself must be cast `as (/*...*/, generator: CodeGenerator) => /*...*/` when installing it in the `.forBlock` dictionary.

### Test Coverage

`npm test` succeeds.

### Documentation

We should check any existing generator documentation for examples written in TypeScript, which will can be updated accordingly.

### Additional Information

Technically the change of signature is a breaking change, but it is unlikely to affect any developers in practice, since any block generator function `(block: Block, generator: CodeGenerator) => /*...*/` can be substituted for `(block: Block, generator: C extends CodeGenerator) => /*...*/`, so any existing generator functions can still be added to any `CodeGenerator` _or subclass_ instance's `.forBlock`.  The only situation in which this would be breaking is if a developer was obtaining block generator functions _from_ a subclass instance's `forBlock` dictionary and assuming they were `(block: Block generator: CodeGenerator) => /*...*/`.

### DIFF
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -19,8 +19,10 @@ import type {Workspace} from './workspace.js';
 import {warn} from './utils/deprecation.js';
 
 /**
- * Type declaration for per-block-type generator functions.
+ * Deprecated, no-longer used type declaration for per-block-type generator
+ * functions.
  *
+ * @deprecated
  * @see {@link https://developers.google.com/blockly/guides/create-custom-blocks/generating-code}
  * @param block The Block instance to generate code for.
  * @param genearator The CodeGenerator calling the function.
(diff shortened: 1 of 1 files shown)


maribethb (approved):
Change seems fine, but I'll note that I did not have any problems with either of these in my TS project:

```
const x = function(
    block: Blockly.Block,
    generator: JsonDefinitionGenerator
): string {
  // omitted
};

jsonDefinitionGenerator.forBlock['input'] = x;
```

or

```
jsonDefinitionGenerator.forBlock['input'] = function(
    block: Blockly.Block,
    generator: JsonDefinitionGenerator
): string {
  // omitted
};
```

So I'm not seeing any problem from the current system with using a subclass of `CodeGenerator` as the parameter type. But, this is still an improvement because there is a lot more information in the TsDoc which is shown when creating one of these functions.

This also works fine: 

```
import {BlockDefinition} from 'blockly/core/generator';

const x: BlockDefinition = function(
    block: Blockly.Block,
    generator: JsonDefinitionGenerator
): string {
  // omitted
};

jsonDefinitionGenerator.forBlock['input'] = x;
```

`BlockGenerator` was never re-exported from `blockly.ts` so it was never made accessible as part of the public API. It is accessible if you import it directly e.g. `import {BlockGenerator} from 'blockly/core/generator';` but we explicitly recommend against doing this. This is all just FYI, don't think it affects how we should proceed with this PR.

PR #7591, opened 2023-10-13
Throw an error for undefined messages in block definitions

### LINKED ISSUE #7589: Throw an error for undefined messages in block definitions
### Check for duplicates

### Description

The root cause of https://github.com/google/blockly/issues/7575 was that `Blockly.Msg['TEXT_PRINT_TITLE']` was undefined. The block was successfully created, however all of its inputs are missing because no message parsing occurred. **This should be an error state (when args0 is present but message0 is undefined)** but instead the block was created successfully, and you only get an error later, for example when trying to deserialize some state that has a block connected to an input that now doesn't exist.

The problematic code is here where https://github.com/google/blockly/blob/ca4fa7e4fb75b0f9ae8b1d878a2b18354ed5a82f/core/block.ts#L1663-L1664 ignores the error state of args0 being defined but message0 being undefined.

For a variety of reasons, `Blockly.Msg` being undefined is an unfortunately somewhat common problem for new developers so providing a good error message here would be useful.

### Reproduction steps

1. Add the following block definition to the playground above the `start()` call:
```
Blockly.defineBlocksWithJsonArray([
        {
          'type': 'test',
          // No `message0` when there should be one!
          'args0': [
            {
              'type': 'input_value',
              'name': 'INPUT'
            }
          ]
        }
      ])
```
1. Run `npm run start`.
2. Load the following JSON into the playground.
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "test",
        "inputs": {
          "INPUT": {
            "block": {
              "type": "text",
              "fields": {
                "TEXT": ""
              }
            }
          }
        }
      }
    ]
  }
}
```
3. Observe the error message:
```
blocks.ts:658 Uncaught Error: The block "test" block (id="3KvXrdD~jmobsK0$2K:T") is missing a(n) INPUT
connection
    at loadInputBlocks (blocks.ts:658:13)
    at appendPrivate (blocks.ts:463:3)
    at appendInternal (blocks.ts:404:13)
    at append (blocks.ts:359:17)
    at BlockSerializer.load (blocks.ts:789:7)
    at Module.load (workspaces.ts:82:38)
    at HTMLInputElement.load (playground.html:181:44)
```

0scvr:
See #7589

## The basics

### Resolves

Fixes #7589 

### Proposed Changes

Adding a check for `json["message0"] === undefined` at the start of `jsonInit` function.

### Reason for Changes

To fix current behavior which allows creation of blocks when messages is undefined.

### Additional Information

I wasn't actually able to reproduce the bug mentionned in #7589, so I added a check which should fix the issue if I understand correctly. How can I attempt to reproduce the issue, please ?

Also, I'd be happy to make any necessary changes. I know my current solution may not be what you had in mind.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -1641,6 +1641,18 @@ export class Block implements IASTNodeLocation, IDeletable {
       );
     }
 
+    // Validate that each arg has a corresponding message
+    let n = 0;
+    while (json['args' + n]) {
+      if (json['message' + n] === undefined) {
+        throw Error(
+          warningPrefix +
+            `args${n} must have a corresponding message (message${n}).`,
+        );
+      }
+      n++;
+    }
+
     // Set basic properties of block.
     // Makes styles backward compatible with old way of defining hat style.
     if (json['style'] && json['style'].hat) {


BeksOmega:
Hello! This looks like a great start =)

> How can I attempt to reproduce the issue, please ?

I've added reproduction steps to the original issue!

---

What we want it to do is check that for every `args#` a `message#` is defined. So you probably need to `while` loop through all of the args before `while` looping through all of the messages, checking that enough messages are defined!

If you could also add a unit test for this under [initialization](https://github.com/BeksOmega/blockly/blob/f0812dbcdfa34b1616a449a38ea19fc6d9e4833c/tests/mocha/block_test.js#L2461) that would be much appreciated =)

Thank you for your work on this! If you have any questions don't hesitate to ask!

core/block.ts
```diff
@@ -1641,6 +1641,22 @@ export class Block implements IASTNodeLocation, IDeletable {
...
+    let n = 0;
+    while (json['args' + n]) {
+      if (json['message' + n] === undefined) {
+        throw Error(
+          warningPrefix +
+            'args' +
+            n +
+            ' must have a corresponding message (message' +
+            n +
+            ').',
```
BeksOmega, line 1654: nit: Could you switch this to a [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) so that it's a bit easier to read?

BeksOmega (commented):
This looks great =) Just one little fix and then I'll get this merged!

BeksOmega (approved):
LGTM =)

PR #7618, opened 2023-10-30
fix: flyout id is different than first placed block

### LINKED ISSUE #7432: First block from flyout matches ID and subsequent ones do not

treypisano:
## The basics

- [ x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes  #7432 

### Proposed Changes

Adds a parameter,  `saveIds: boolean`, to the [`save`](https://github.com/google/blockly/blob/4ab8d000995a93eafbc29444abb76e535aedf134/core/serialization/blocks.ts#L82) function. I also made sure that the parameter is passed through the recursive calls.

### Reason for Changes

This ensures that when a block is dragged out, the ID of the newly placed block is different than the one in the flyout. 

### Test Coverage

1. Open up the flyout, inspect the if block and note the data-id attribute 
2. Drag the block out, note the data-id attribute
3. The data-ids are not the same

### DIFF
--- a/core/serialization/blocks.ts
+++ b/core/serialization/blocks.ts
@@ -123,12 +123,12 @@ export function save(
   if (addInputBlocks) {
     // AnyDuringMigration because:  Argument of type '{ type: string; id:
     // string; }' is not assignable to parameter of type 'State'.
-    saveInputBlocks(block, state as AnyDuringMigration, doFullSerialization);
+    saveInputBlocks(block, state as AnyDuringMigration, doFullSerialization, saveIds);
   }
   if (addNextBlocks) {
     // AnyDuringMigration because:  Argument of type '{ type: string; id:
     // string; }' is not assignable to parameter of type 'State'.
-    saveNextBlocks(block, state as AnyDuringMigration, doFullSerialization);
+    saveNextBlocks(block, state as AnyDuringMigration, doFullSerialization, saveIds);
   }
 
   // AnyDuringMigration because:  Type '{ type: string; id: string; }' is not
@@ -271,6 +271,7 @@ function saveInputBlocks(
   block: Block,
   state: State,
   doFullSerialization: boolean,
+  addIds: boolean,
 ) {
   const inputs = Object.create(null);
   for (let i = 0; i < block.inputList.length; i++) {
@@ -279,6 +280,7 @@ function saveInputBlocks(
     const connectionState = saveConnection(
       input.connection as Connection,
       doFullSerialization,
+      addIds
     );
     if (connectionState) {
       inputs[input.name] = connectionState;
@@ -302,6 +304,7 @@ function saveNextBlocks(
   block: Block,
   state: State,
   doFullSerialization: boolean,
+  saveIds: boolean
 ) {
   if (!block.nextConnection) {
     return;
@@ -309,6 +312,7 @@ function saveNextBlocks(
   const connectionState = saveConnection(
     block.nextConnection,
     doFullSerialization,
+    saveIds
   );
   if (connectionState) {
     state['next'] = connectionState;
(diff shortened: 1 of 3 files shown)


core/serialization/blocks.ts
```diff
@@ -270,6 +281,7 @@ function saveInputBlocks(
   block: Block,
   state: State,
   doFullSerialization: boolean,
+  addIds: boolean,
```
BeksOmega, line 284: Can you make this `saveIds` so that it's consistent?

BeksOmega (changes requested):
First look seems good =) I'll pull it down and give it a manual test in a bit!

Would you also be interested in writing a tiny unit test in our [block serialization suite](https://github.com/google/blockly/blob/4003093404785825d6dd009190cf312ce5f44654/tests/mocha/jso_serialization_test.js#L49) that tests that when `saveIds` is false, there aren't IDs?

Thank you for your work on this @treypisano :D

treypisano:
Sure, I'll write that test.

core/serialization/blocks.ts
```diff
@@ -86,20 +86,21 @@ export function save(
...
   } = {},
 ): State | null {
   if (block.isInsertionMarker()) {
     return null;
   }
-
   const state = {
     'type': block.type,
-    'id': block.id,
+    'id': saveIds ? block.id : null,
```
BeksOmega, line 103: We actually want this to be `undefined` instead of `null`. `null` gets stringified and takes up space in the save data, but `undefined` does not get serialized.

E.g. if you pass an object like the following to `JSON.stringify`:
```
{
  propertyA: null,
  propertyB: undefined
}
```
You get:
```
'{"propertyA": null}'
```

BeksOmega (changes requested):
Just one fix and then LGTM!

core/serialization/blocks.ts
treypisano: Got it, I hadn't thought of it that way

BeksOmega (approved):
LGTM! Thank you for your work on this @treypisano =)

PR #7619, opened 2023-10-30
fix: blocks being dragged behind toolbox

### LINKED ISSUE #7354: blocks behind toolbox while dragging to delete
### Check for duplicates

### Description

Dragging a blocklyDraggable element onto a blocklyToolboxDiv element will be obstructed by blocklyToolboxDiv, making it impossible to elevate the blocklyDraggable element level. Originally, raising the blocklyBlockDragSurface element level was sufficient. How can we solve this problem?

blocklyDraggable元素拖拽到blocklyToolboxDiv元素会被blocklyToolboxDiv遮挡，无法提升blocklyDraggable元素层级，原来通过提升blocklyBlockDragSurface元素层级即可，如何解决这个问题？

### Reproduction steps

Drag building blocks to toolbox

### Stack trace

_No response_

### Browsers

Chrome desktop

BeksOmega:
## The basics

### Resolves

Fixes #7354 

### Proposed Changes

Adds a layer manager to keep track of what layers we have (dragging layer, block layer, and bubble layer) and what order they should be in.

The new system should be entirely internal.

### Reason for Changes

Needed to add a dragging layer so that blocks could exist in front of the toolbox while dragging, so I just decided to implement a better layer management system at the same time. This prepares us to easily add new layers in the future if we need to.

There are still parts of blockly that are directly accessing the block and bubble layers, these can be moved to use the layer manager.

### Test Coverage

Added tests to make sure the layer manager properly handles adding new layers, and appending elements to the drag layer.

Manually tested that dragging elements and the worskpace seems to work properly, with the correct draw order.

### Documentation

N/A

### Additional Information

I know the issue said to just hard code a drag layer in there, but I didn't see a need to since I'm not working on anything else atm.

### DIFF
--- a/core/block_dragger.ts
+++ b/core/block_dragger.ts
@@ -30,6 +30,7 @@ import * as dom from './utils/dom.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
 import {hasBubble} from './interfaces/i_has_bubble.js';
 import * as deprecation from './utils/deprecation.js';
+import * as layers from './layers.js';
 
 /**
  * Class for a block dragger.  It moves blocks around the workspace when they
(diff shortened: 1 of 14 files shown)


NeilFraser (approved):

BeksOmega:
@NeilFraser Could you also pull this down and manually test it for me since this is kind of a dangerous area?

PR #7621, opened 2023-10-31
fix: prevent crash when clicking on a field to delete a block

### LINKED ISSUE #7587: Crash if a field click deletes a block
### Check for duplicates

### Description

If you have a block with a field that deletes the block*, Blockly will crash because the gesture handling code assumes the block continues to exist.

We can fix this in several ways:

1. Move `bringBlockToFront` to the beginning of [`doFieldClick`](https://github.com/google/blockly/blob/9c7c71ad879a62e285d97236e1c6907287463251/core/gesture.ts#L938) (so that it happens before the field has a chance to handle the click)
2. Have `bringBlockToFront` make sure `disposed` is false on the block before calling the `bringToFront` method. This would prevent trying to move a dead/dying block.

### Reproduction steps

1. Add the following block definition to the playground above the `start` call.
```js
Blockly.Blocks['test'] = {
        init: function() {
          this.setStyle('math_blocks');
          this.appendDummyInput()
              .appendField(new Blockly.FieldImage(
                  'media/arrow.png',
                  30,
                  30,
                  'test',
                  function() {
                    this.getSourceBlock().dispose();
                  }
              ));
        }
```
2. Run `npm run start` to start the playground.
3. Load the following JSON into the playground:
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "test"
      }
    ]
  }
}
``` 
4. Open the browser.
4. Click the arrow on the block.
5. Observe the error in the console.

### Stack trace

```shell
Uncaught TypeError: Cannot read properties of null (reading 'childNodes')
    at BlockSvg$$module$build$src$core$block_svg.bringToFront (block_svg.ts:1227:34)
    at Gesture$$module$build$src$core$gesture.bringBlockToFront_ (gesture.ts:940:25)
    at Gesture$$module$build$src$core$gesture.doFieldClick_ (gesture.ts:880:10)
    at Gesture$$module$build$src$core$gesture.handleUp (gesture.ts:591:14)
    at HTMLDocument.g (browser_events.ts:70:9)
```

### Additional Info

This should also be fixed for icons since, https://github.com/google/blockly/issues/7588 is fixed.

## To Fix

1) Run the reproduction steps above to make sure you can reproduce the issue.
2) Add a check in `bringToFront` that checks if `this.isDeadOrDying()`, and returns early if so.
3) Run the reproduction steps again to make sure the issue doesn't reproduce.

Apoorvgarg-creator:
## The basics

### Resolves

Fixes  #7587 

### Proposed Changes

The change is to check if the block is dead or is in dying state before accessing it in the ```bringToFront``` function.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1185,7 +1185,7 @@ export class BlockSvg
     let block: this | null = this;
     do {
       const isDeadOrDying: boolean = block.isDeadOrDying();
-      if(isDeadOrDying){
+      if (isDeadOrDying) {
         break;
       }
       const root = block.getSvgRoot();
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1183,8 +1183,7 @@ export class BlockSvg
   bringToFront(blockOnly = false) {
     /* eslint-disable-next-line @typescript-eslint/no-this-alias */
     let block: this | null = this;
-    const isDeadOrDying: boolean = block.isDeadOrDying();
-    if (isDeadOrDying) {
+    if (block.isDeadOrDying()) {
       return;
     }
     do {
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1184,6 +1184,10 @@ export class BlockSvg
     /* eslint-disable-next-line @typescript-eslint/no-this-alias */
     let block: this | null = this;
     do {
+      const isDeadOrDying: boolean = block.isDeadOrDying();
+      if(isDeadOrDying){
+        break;
+      }
       const root = block.getSvgRoot();
       const parent = root.parentNode;
       const childNodes = parent!.childNodes;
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -1184,6 +1184,10 @@ export class BlockSvg
     /* eslint-disable-next-line @typescript-eslint/no-this-alias */
     let block: this | null = this;
     do {
+      const isDeadOrDying: boolean = block.isDeadOrDying();
+      if (isDeadOrDying) {
```
BeksOmega, line 1188: If you're immediately checking it, there's no need to store the value in a temp variable!
```suggestion
      if (block.isDeadOrDying()) {
```

BeksOmega (changes requested):
Heya @Apoorvgarg-creator Thank you for your work on this :D

I gave it a look and I /think/ we can move the `isDeadOrDying` check out of the loop, so that it's at the top of the function. Can you give that a try for me and tell me if it works?

BeksOmega (approved):
This looks great! Thank you for the fix @Apoorvgarg-creator =)

PR #7634, opened 2023-11-06
fix: make autoclose toggleable for flyouts

### LINKED ISSUE #7506: Make `autoClose` in flyouts/toolboxes toggleable
### Check for duplicates

### Problem

We have a feature request in samples to make the continuous toolbox autoclose https://github.com/google/blockly-samples/issues/1922

We can't toggle this via the `constructor` or `init` because those are both called internally by Blockly. So we need to toggle it after the toolbox/flyout has been constructed.

Currently toolboxes/flyouts do not support changing the `autoClose` property after they have been initialized. You can assign to it because it is public, but the logic is not built to support this.

For example, if you change a flyout from always open to auto close, you [still get a delete area for the flyout](https://github.com/google/blockly/blob/3b234c7f87acab765e9b4a9bc6d9638ed229b087/core/flyout_base.ts#L504). If you change a flyout from auto close to always open, it [won't filter for capacity](https://github.com/google/blockly/blob/3b234c7f87acab765e9b4a9bc6d9638ed229b087/core/flyout_base.ts#L388).

### Request

Go through the flyout and make sure all of the functionality supports toggling `autoClose`. Add set methods and get methods for this property.

### Alternatives considered

N/A

### Additional context

Previous related requests:
* [[Improvement] Make flyout in toolbox without categories close when clicking outside it and autoClose=true](https://github.com/google/blockly/issues/1852)
* [Provide more support for non auto closing flyouts](https://github.com/google/blockly/issues/4378)

BeksOmega:
## The basics

### Resolves

Fixes #7506 

~Also https://github.com/google/blockly-samples/issues/1922 ? Need to check.~

### Proposed Changes

Makes it so that autoclose on flyouts is toggleable.

### Reason for Changes

People want to make the continuous toolbox autoclose. Also app inventor wants to control the flyout from an external component, so they use an autoclosing simple flyout.

### Test Coverage

Manually tested steps [here](https://github.com/google/blockly/issues/7506#issuecomment-1729531204).

Fixed up some change detector unit tests.

### Documentation

N/A

### Additional Information

N/A

Makes https://github.com/google/blockly-samples/issues/1922 possible to do further work on. Still needs fixes to `refreshSelection`

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -480,6 +480,11 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
     return this.workspace_;
   }
 
+  setAutoClose(autoClose: boolean) {
+    console.log('set');
+    this.autoClose = autoClose;
+  }
+
   /**
    * Is the flyout visible?
    *
@@ -624,6 +629,7 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
 
     // Parse the Array, Node or NodeList into a a list of flyout items.
     const parsedContent = toolbox.convertFlyoutDefToJsonArray(flyoutDef);
+    if (!parsedContent.length) return; // No need to show an empty flyout.
     const flyoutInfo = this.createFlyoutInfo(parsedContent);
 
     renderManagement.triggerQueuedRenders();
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -495,7 +495,12 @@ export abstract class Flyout
 
   /** Automatically hides the flyout if it is an autoclosing flyout. */
   autoHide(onlyClosePopups: boolean): void {
-    if (!onlyClosePopups && this.autoClose) this.hide();
+    if (
+      !onlyClosePopups &&
+      this.targetWorkspace.getFlyout(true) === this &&
+      this.autoClose
+    )
+      this.hide();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


core/flyout_horizontal.ts
```diff
@@ -382,22 +382,10 @@ export class HorizontalFlyout extends Flyout {
         }
       }
 
-      if (
```
maribethb, line 385: It makes way more sense for this logic to be in the metrics manager, so I like that change. But I'm wondering if it has the possibility to break people who depend on the current values the metrics manager returns / will they have to update their math to account or not account for the flyout now? or was the translate added by the flyout already accounted for in the metrics?

BeksOmega: The translation gets overwritten the next time the metrics are applied to the workspace. So we were usually overwriting this update anyway.

We definitely /could/ break people that are depending on the current metrics values. But it's unlikely to do that because we're essentially adding new functionality. We have two new possible states which are categories + always open flyout and no categories + hiding flyout, and we're properly reporting those metrics.

core/flyout_base.ts
```diff
@@ -426,7 +429,7 @@ export abstract class Flyout extends DeleteArea implements IFlyout {
    */
   dispose() {
     this.hide();
-    this.workspace_.getComponentManager().removeComponent(this.id);
+    this.targetWorkspace.getComponentManager().removeComponent(this.id);
```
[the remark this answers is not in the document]
BeksOmega: Yesh this was indeed a bug before.

maribethb (approved):

core/flyout_horizontal.ts
BeksOmega: We should definitely manually test whether this breaks the continuous flyout though. I created a testing task in the GH project and assigned it to Rachel since she's doing release stuffs.

Issue #7204, opened 2023-06-23
Support the dynamic connections plugin without hacky monkey-patching.


johnnesky:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

The `block-dynamic-connection` plugin monkey-patches `Blockly.InsertionMarkerManager`. It adds a new property, `pendingBlocks`, and overrides the `update` and `dispose` methods, but there's no facility to provide an alternative implementation of the interface so it has to fall back on monkey-patching. 

Additionally, my PR to migrate the plugin to TypeScript uses hacky syntax (square brackets) to access private methods, which maybe could be made protected instead. https://github.com/google/blockly-samples/pull/1719

### Request

Design and implement support for the dynamic connections plugin and similar plugins. The plugin logic can be modified too if necessary.

BeksOmega:
Related stuff about reindeer connections: https://github.com/microsoft/pxt-blockly/pull/287

BeksOmega:
Closed by https://github.com/google/blockly-samples/pull/2190

PR #7637, opened 2023-11-09
feat: add connection previewer interface

BeksOmega:
## The basics

### Resolves

Work on #7204 

### Proposed Changes

Adds an `IConnectionPreviewer` interface for later use.

### Reason for Changes

Working on making it possible to swap out the connection preview method.

### Test Coverage

No tests yet because this just defines a type not behavior.

### Documentation

N/A

### Additional Information

Merging this into a feature branch b/c I'm not sure if I'll be able to get it done in time for the release or not.


core/interfaces/i_connection_previewer.ts
```diff
@@ -0,0 +1,50 @@
...
+   * staticCon).
+   *
+   * @param draggedCon The connection on the block stack being dragged.
+   * @param staticCon The connection not being dragged that we are
+   *     connecting to.
+   * @param replacedBlock The block currently connected to the staticCon that
+   *     is being replaced.
+   */
+  previewReplacement(
+    draggedCon: RenderedConnection,
```
maribethb, line 27: nit (and you can disagree with me here) but I prefer "conn" as the abbreviation for "connection" if we're going to abbreviate it at all. apologies as i meant to leave this comment in your doc but hoping i've caught you early enough here

maribethb (approved):

core/interfaces/i_connection_previewer.ts
BeksOmega: Oh yeah totally good! I did the types in a separate PR partially so that it would be easier to bike shed :P So this is perf!

Issue #7600, opened 2023-10-20
Improve implementation and type safety of generators


cpcallen:
## Description

In the process of converting the generators (`generators/*/*`) from JavaScript to TypeScript, the strategy for dealing with type errors unearthed by `tsc` (due to more specific types in the code and/or more assiduous type checking by `tsc` as compared to Closure Compiler) has been to try to suppress the type errors with casts (`as`, `!`, etc.) rather than make other code changes, in an effort to ensure that the `.js` code generated by `tsc` is as close as possible to the original—to ensure bug-for-bug compatibility.

Once the TypeScript migration is complete, we should revisit the library blocks implementation and remove such casts, replacing them with:

- more correct types, where the existing types are not quite correct, and/or
- type assertions and other runtime checks (rather than casts), and/or
- bugfixes, where the code is actually wrong.

### General issues throughout:

* [ ] Use of `!` (typically but not exclusively in the form (`!.`) to suppress could-be-null errors.
* [ ] Use of `as` casts (which in some cases may require making the return types of methods in core more specific)
  *  Look out in particular for `as AnyDuringMigration`, which covers a multitude of sins including accessing private properties.

## See Also

Issue #6920 tracks similar work for the library blocks.

PR #7646, opened 2023-11-13
refactor(generators): Migrate Dart generators to TypeScript

cpcallen:
## The basics

### Resolves

Part of #6828.

### Proposed Changes

Migrate `generators/dart/*.js` and `generators/dart.js` to TypeScript.

Also make a few minor corrections to previously-migrated files in `generators/javascript/` and `generators/python/` for consistency (see commit d934aa1).

### Reason for Changes

Finish TS migration.

### Test Coverage

`npm test` passes.  No changes to manual test procedures required.

### Documentation

No changes to documentation yet, as there should be minimal changes to the generated code and no changes to the published typings (because during build we overwrite the generated `.d.ts` files with the ones from `typings/`).

### Additional Information

It may be easier to review this commit-by-commit, as the final format commit make a lot of noisy changes.


generators/dart/colour.ts
```diff
@@ -25,7 +25,9 @@ export function colour_picker(block: Block, generator: DartGenerator): [string,
 
 export function colour_random(block: Block, generator: DartGenerator): [string, Order] {
   // Generate a random colour.
-  generator.definitions_['import_dart_math'] =
+  // TODO(#7600): find better approach than casting to any to override
```
BeksOmega, line 33: Similarly to before (I think) would like this linked to #7326 because I think this is an API problem not just a type safety problem.

generators/javascript/logic.ts
```diff
@@ -63,7 +63,7 @@ export function logic_compare(
   block: Block,
   generator: JavascriptGenerator,
 ): [string, Order] {
-  // Dictionary of OP comparison operators and their implementations.
+  // Comparison operator.
```
BeksOmega, line 66: Is this a useful comment?

BeksOmega (approved):
Overall LGTM!

generators/dart/loops.ts
```diff
@@ -10,12 +10,14 @@
 
 // Former goog.module ID: Blockly.Dart.loops
 
-import {Order} from './dart_generator.js';
 import * as stringUtils from '../../core/utils/string.js';
+import type {Block} from '../../core/block.js';
+import type {DartGenerator} from './dart_generator.js';
 import {NameType} from '../../core/names.js';
+import {Order} from './dart_generator.js';
```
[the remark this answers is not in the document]
cpcallen: I think you should think again.  (I didn't _add_ any new uses of `Order`, but there are many existing ones.  This line appears to be an addition of a new import but in fact is just fixing import ordering.)

generators/dart/colour.ts
cpcallen: I believe you indicated you were statisfied with #7600 linking to #7326 about this topic (which it does).  Would you like the actual TODO to do so as well?

generators/javascript/logic.ts
cpcallen: The "Comparison operator" comment, which is here being restored having been inadvertently removed in a previous PR, is in a standard format that is common to almost all the generator functions.

Admittedly it should originaly have been a `/** */` JSDoc comment on the line above the `function` declaration.  I didn't want to try to fix all of them in this PR, so I've added that as a to-do item to #7600.

cpcallen:
Fixed the one definite error.  Please let me know if your other comments have been satisfactorily addressed (at least for the purposes of this PR).

generators/javascript/logic.ts
BeksOmega: Ohhh I didn't realize that this one was also a block-level comment. I thought it was about the dictionary itself.

generators/dart/colour.ts
BeksOmega: Oh haha nice I had forgotten. yep sg

BeksOmega:
LGTM!

PR #7647, opened 2023-11-13
refactor(generators): Migrate PHP generators to TypeScript

cpcallen:
## The basics

### Resolves

Part of #6828.

### Proposed Changes

Migrate `generators/php/*.js` and `generators/php.js` to TypeScript.

Also make a few minor corrections to previously-migrated files in `generators/javascript/` and `generators/python/` for consistency (see commit 40ede09).

### Reason for Changes

Finish TS migration.

### Test Coverage

`npm test` passes.  No changes to manual test procedures required.

### Documentation

No changes to documentation yet, as there should be minimal changes to the generated code and no changes to the published typings (because during build we overwrite the generated `.d.ts` files with the ones from `typings/`).

### Additional Information

It may be easier to review this commit-by-commit, as the final format commit make a lot of noisy changes.

N.B. that individual commit messages contain some additional information about each commit.


generators/php/php_generator.ts
```diff
@@ -242,65 +251,62 @@ export class PhpGenerator extends CodeGenerator {
...
+   * Generate code representing the specified value input, adjusted to take into
+   * account indexing (zero- or one-based) and optionally by a specified delta
+   * and/or by negation.
+   *
+   * @param block The block.
+   * @param atId The ID of the input block to get (and adjust) the value of.
+   * @param delta Value to add.
+   * @param negate Whether to negate the value.
+   * @param order The highest order acting on this value.
+   * @returns The adjusted value.
```
maribethb, line 269: Should this be "The adjusted value or code that generates the adjusted value"?

generators/php/php_generator.ts
```diff
@@ -242,65 +251,62 @@ export class PhpGenerator extends CodeGenerator {
...
    */
-  getAdjusted(block, atId, opt_delta, opt_negate, opt_order) {
-    let delta = opt_delta || 0;
-    let order = opt_order || this.ORDER_NONE;
+  getAdjusted(block: Block, atId: string, delta = 0, negate = false, order = Order.NONE): string | number {
```
maribethb: In its current state, it doesn't look like this function ever returns a number. I guess this is overridable though?

generators/php/php_generator.ts
```diff
@@ -242,65 +251,62 @@ export class PhpGenerator extends CodeGenerator {
...
+
+    // Easy case: no adjustments.
+    if (delta === 0 && !negate) {
+      return at;
+    }
+    // If the index is a naked number, adjust it right now.
+     if (stringUtils.isNumber(at)) {
+      at = String(Number(at) + delta);
+      if (negate) {
+        at = String(-Number(at));
```
maribethb, line 302: I guess `at` has to be a string for TS since `valueToCode` will return a string. but the converting back and forth between string and number is really ugly. would it look better if you used a different variable and only convert it to string when you're returning that value?

generators/php/text.ts
```diff
@@ -10,41 +10,45 @@
...
+export function text_multiline(block: Block, generator: PhpGenerator): [string, Order] {
   // Text value.
   const code = generator.multiline_quote_(block.getFieldValue('TEXT'));
   const order =
       code.indexOf('.') !== -1 ? Order.STRING_CONCAT : Order.ATOMIC;
   return [code, order];
 };
 
-export function text_join(block, generator) {
+export function text_join(block: Block, generator: PhpGenerator): [string, Order] {
```
maribethb: would it work to just type `block` as `JoinMutatorBlock` to begin with? haven't thought through the implications of this, just throwing it out there.

maribethb (approved):
Overall lgtm, minor questions and comments

generators/javascript/lists.ts
```diff
@@ -303,6 +303,7 @@ export function lists_getSublist(
   block: Block,
   generator: JavascriptGenerator,
 ): [string, Order] {
+  // Get sublist.
```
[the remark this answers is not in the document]
cpcallen: Agreed.  This is the fourth item in the to-do list in #7600.

generators/php/loops.ts
```diff
@@ -29,11 +32,11 @@ export function controls_repeat_ext(block, generator) {
   branch = generator.addLoopTrap(branch, block);
   let code = '';
   const loopVar =
-        generator.nameDB_.getDistinctName('count', NameType.VARIABLE);
+        generator.nameDB_!.getDistinctName('count', NameType.VARIABLE);
```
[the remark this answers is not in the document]
cpcallen: > I think we could improve this code by checking that nameDB_ exists and if not tell people they need to initialize the generator. This is a common mistake and it's actually not safe to assert this exists imo.

Agreed.  This is the first item in the to-do list in #7600—but I added a note specifically mentioning a check.
 
> Or we could add a `getDistinctName` to the Generator class, similar to [this one](https://github.com/google/blockly/blob/develop/core/generator.ts#L543-L544), which would do the check for us so we don't repeat it in every place and don't need to assert on the nameDB

That's an even better idea, which I also added to #7600.

generators/php/math.ts
```diff
@@ -10,31 +10,33 @@
...
   }
-  return [code, order];
+  return [String(number), number >= 0 ? Order.ATOMIC : Order.UNARY_NEGATION];
 };
 
-export function math_arithmetic(block, generator) {
+export function math_arithmetic(block: Block, generator: PhpGenerator): [string, Order] {
   // Basic arithmetic operators, and power.
-  const OPERATORS = {
+  const OPERATORS: Record<string, [string | null, Order]> = {
```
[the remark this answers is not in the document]
cpcallen: Oops!  This was just a copy-and-paste from one of the other generators, which did have `null`s in this lookup table.  Fixed.

generators/php/lists.ts
```diff
@@ -340,11 +344,19 @@ function ${generator.FUNCTION_NAME_PLACEHOLDER_}(&$list, $at, $value) {
   throw Error('Unhandled combination (lists_setIndex).');
 };
 
-export function lists_getSublist(block, generator) {
+export function lists_getSublist(block: Block, generator: PhpGenerator): [string, Order] {
   // Get sublist.
+  // Dictionary of WHEREn field choices and their CamelCase equivalents.
+  const wherePascalCase = {
```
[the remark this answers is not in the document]
cpcallen: Nope.  This again was an overzealous copy-and-paste (possibly overzealous `git diff | sed | git patch`, in particular) from the JavaScript generator.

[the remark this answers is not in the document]
cpcallen: Removed.  Good catch!

generators/php/php_generator.ts
cpcallen: I've gone with "The adjusted value or code that evaluates to it."—opting for "evaluates to" rather than "generates" to reduce confusion with generation of code by the code generators.

generators/php/php_generator.ts
cpcallen: I started carrying over the previous type declaration (`@return {string | number}` from `javascript_generator.js` and did the same in all the others, but in fact only the `PythonGenerator`'s (similar code but differently named) `getAdjustedInt` ever returns a number.

I think the correct course of action here is to:

- Fix this PR to have `getAdjusted` be declared as returning only a `string`.
- In a separate PR, do the same for the other `getAdjusted` implementations.

Then, since PythonGenerator's gratuitously-different `getAdjustedInt` _does_ actually return numbers and some callers depend on this fact, fix this by:
- Introducing a new a string-only `getAdjusted` method.
- Updating callers to use it (and work correctly when it returns `'0'` instead of `0`).
- Marking `getAdjustedInt` as deprecated and removing it in due course.

I have added this to the list in #7600.

cpcallen: (And yes: these changes would break any code that overrides `getAdjusted` with an implementation that returns numbers, if any such code exists and is implemented in TypeScript, but I do not think this egregiously worse than all the other random breakage we are likely to cause such developers just by dint issuing non-`any` generator typings.)

generators/php/php_generator.ts
cpcallen: A good suggestion.  I think this is best dealt with separately, however, so that all the implementations can be improved at the same time and perhaps made more consistent in the process, so I've again added it to the list in #7600.

generators/php/text.ts
cpcallen: That would be a great idea but alas it is not possible to put a `(block: JoinMutatorBlock, …) => …` into the `phpGenerator.forBlock` dictionary because that type is not assignable to `(block: Block, …) => …` (which is what the `forBlock` dictionary is declared as containing).

I've been thinking about this issue for a while, and I think the correct approach is to for the block definitions to provide [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) (for at least all those block type that declare additional properties) and then have the generator functions use those type predicates to narrow the type of `block` via runtime test (e.g. `if (!isJoinMutatorBlock(block)) throw new TypeError()`, after which `block` is understood to be a `JoinMutatorBlock`).

That obviously entails providing such predicates, and possibly changing the way the blocks types are declared, so I think this is out-of-scope for this PR.

maribethb (approved):

PR #7660, opened 2023-11-21
fix: Releases the event listeners that has not been released

ZoolMons:
## The basics

### Event listeners will not be disposed even workspace is cleared

Fixes 

### Reason for Changes
Will cause the memory leak.


BeksOmega:
Let me know when you've signed the CLA @ZoolMons and I'll rerun the automations =)

PR #7661, opened 2023-11-21
chore: modify readme about how we want people to acquire Blockly

### LINKED ISSUE #7624: Docs say blockly bundle comes with JS generator
### Location

https://www.npmjs.com/package/blockly

### Type

- [X] Text
- [ ] Image or Gif
- [ ] Other

### Suggested content

When doing `import * as Blockly from 'blockly';`

Should not say that you get the JS generator, because while we load it, there's no way for people to access it.

### Additional context

_No response_

BeksOmega:
## The basics

### Resolves

Fixes #7624 

### Proposed Changes

Modifies the README to hopefully be correct about how we want people to import things.

### Reason for Changes

Getting our code is important.

### Test Coverage

No functional changes!

### Documentation

This is documentation silly!

### Additional Information

N/A

### DIFF
--- a/scripts/package/README.md
+++ b/scripts/package/README.md
@@ -73,17 +73,20 @@ const {javascriptGenerator} = require('blockly/javascript');
 const En = require('blockly/msg/en');
 ```
 
-## Applying blocks and messages
+## Applying messages
 
-Once you have the library blocks and the message files, you also need to apply
-them.
+Once you have the message files, you also need to apply them.
 
 ```js
-Blockly.Blocks = libraryBlocks;
 Blockly.setLocal(En);
 ```
 
 For a full list of supported Blockly locales, see: [https://github.com/google/blockly/tree/master/msg/js](https://github.com/google/blockly/tree/master/msg/js)
+
+## Docs
+
+For more information about how to use Blockly, check out our
+[devsite](https://developers.google.com/blockly/guides/overview).
 
 ## License
 
--- a/scripts/package/README.md
+++ b/scripts/package/README.md
@@ -5,26 +5,10 @@ blocks together to build programs. All code is free and open source.
 
 The source for this module is in the [Blockly repo](http://github.com/google/blockly).
 
-## Installation
-
-You can install this package either via `npm` or `unpkg`.
-
-### npm
-
-```bash
-npm install blockly
-```
-
-### unpkg
-
-```html
-<script src="https://unpkg.com/blockly/blockly.min.js"></script>
-```
-
 ## Example Usage
 
 ```js
-import Blockly from 'blockly';
+import * as Blockly from 'blockly/core';
 Blockly.inject('blocklyDiv', {
     ...
 })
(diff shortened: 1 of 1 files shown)


scripts/package/README.md
```diff
@@ -34,44 +18,71 @@ Blockly.inject('blocklyDiv', {
...
 
-#### Blockly Languages
+Once you have the library blocks and the message files, you also need to apply
+them.
 
 ```js
-import * as Fr from 'blockly/msg/fr';
-Blockly.setLocale(Fr);
+Blockly.Blocks = libraryBlocks;
+Blockly.setLocal(En);
```
BeksOmega, line 83: Still not sure whether this is necessary or not.

scripts/package/README.md
```diff
@@ -34,44 +18,71 @@ Blockly.inject('blocklyDiv', {
...
 
-### Importing Blockly
+## Installation
 
-When you import Blockly with `import * as Blockly from 'blockly';` you'll get the default modules:
-Blockly core, Blockly built-in blocks, the JavaScript generator and the English lang files.
+You can install this package either via `npm` or `unpkg`.
 
-If you need more flexibility, you'll want to define your imports more carefully:
+### unpkg
```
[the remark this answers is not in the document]
BeksOmega: @maribethb

scripts/package/README.md
cpcallen:
```suggestion
Blockly.setLocale(En);
```
At the moment[*] mearly `import`ing (or `require`ing) `blockly/blocs` is sufficient to populate `Blockly.Blocks`, and trying to assign to that property does not change the dictionary object Blockly uses to look up block definitions, it merely causes it to no longer accessible to the developer (so they can't add custom block definitions or delete ones we've provided).

[*]: N.B. this is proposed to change—see #5875 and recent conversations re: best practices for publishing fields and blocks—but remains the case for the time being, and when it does change attempting to assign `Blockly.Blocks = /*...*/` will still be wrong.

As for whether the `setLocale` call (note final 'e') is required: for historical reasons this depends on whether a `Blockly` global variable is defined (and is pointing at the correct object).  If it is then the `setLocale` call is is not necessary; if it is not then it is.  To clarify further:

* When using a `<script>` tag to load `blockly_compressed.js` then `globalThis.Blockly` is defined and so merely loading `en.js` (as a script _or_ as a module) will set all the `Blockly.Msg[]`es.
  * A special case of this is loading `blockly.min.js` as a script, which includes both `blockly_compressed.js` and `en.js`
* When using `import` or `require`, `globalThis.Blockly` will not normally be defined by the UMD wrappers, but a developer could do this manually, e.g.:
  ```JS
  Blockly = require('blockly');  // N.B. no `const` so this sets a global.
  require('blockly/msg/en');  // Loads `Blockly.Msg` with no call to setLocale
  ```
  (This can also be done using `import` although it's harder to do by accident.)

BeksOmega:
> * Is it necessary to provide these instructions in `package.json`?  Could we not just refer people to devsite?  Or is the idea to have old versions of the NPM package have instructions that are correct for those versions?

I think most folks provide basic usage instructions with their package. It seems like a more use-friendly experience if they can at least load the code from the info on the readme. But pointing to devsite would be more maintainble, so I wouldn't mind just including al ink and no instructions.

> * Could we not eliminate `scripts/package/README.md` by combining it with the top-level `README.md`?  It's great that you're updating it here but I get the impression it is often overlooked and left with out-of-date instructions.  That's _slightly_ less likely to happen with the top-level `README.md`, and certainly having only one copy seems otherwise preferable.

The root-level readme talks a lot about our release cycle and things like that. I don't think it makes sense to include that here, but can be nice for people that want to contribute to the project. I would rather just add a link to devsite than include info in the top-level readme.

scripts/package/README.md
cpcallen: At the risk of getting lost in the weeds: it turns out the example I gave above is wrong:

BeksOmega: Ok so it sounds like:
  * For blocks, don't need to be assigned at the moment.
  * For languages, you do need to call `setLocale` unless you're loading from a script tag.

cpcallen: I concur.

maribethb:
I agree that it makes sense to keep this readme separate from the top-level one, however:

- It is easy for this to get out of date since it's not next to our other documentation.
- You still can't do anything with the package unless you also explain `Blockly.inject`, so as comprehensive as these are, it doesn't explain enough to start using blockly.
- All of this information should also be somewhere on devsite if it isn't already (the "getting the code" page probably needs to be updated and should probably contain very similar content).
- This page needs to link to devsite to point people at the documentation.


So given the above, I think we should either:
- put the bare minimum here (`npm install blockly`) and link to devsite for the rest
- update the devsite "get the code" page to be identical content to this and add a comment on the devsite page to remind authors to update this when they update that page

Thoughts?

maribethb:
Also meta question: Are we not recommending people use `import * as Blockly from 'blockly'` anymore? That's what the sample app does. They don't need to call `setLocale` if they do that.

BeksOmega:
> update the devsite "get the code" page to be identical content to this and add a comment on the devsite page to remind authors to update this when they update that page

This PR updates things to match the new get the code page (cl/577317152). I'll add a comment to tell people to update this when they update that.

> Also meta question: Are we not recommending people use `import * as Blockly from 'blockly'` anymore? That's what the sample app does. They don't need to call `setLocale` if they do that.

I'm not recommending that because it doesn't work great "batteries included" anymore since we changed how the javascript generators are exported.

I think it's clearer to just tell people to import all of the things they need individually.

PR #7693, opened 2023-12-05
Release dummy wheel listener on workspace dispose

### LINKED ISSUE #7674: Release wheel listener when disposing workspace.
### Check for duplicates

### Description

We have a dummy wheel listener we add to the document to work around a bug in safari. Currently this listener is  not getting disposed when the workspace is disposed.

## To Fix

1) Create a class member variable to store the [dummy listener](https://github.com/google/blockly/blob/3d3cd3fbb0c86b7c855baa592663d2989f921c2e/core/workspace_svg.ts#L790).
2) Remove the listener inside the WorkspaceSvg's `dispose` method.

ananta:
## The basics

### Resolves

Fixes #7674

### Proposed Changes
- Introduced a new class member dummyWheelListener in WorkspaceSvg to store the dummy wheel event listener.
- Modified the dispose method in WorkspaceSvg to remove the dummyWheelListener from the document body when the workspace is disposed.

### Reason for Changes
Previously, the dummy wheel event listener added to the document as a workaround for a Safari bug was not being properly disposed of. This could potentially lead to memory leaks and unintended behavior. By storing the listener in a class member, we can ensure it is correctly removed when the workspace is disposed, thus improving the robustness and cleanliness of the code.

### Test Coverage

- Manually tested to confirm that the dummyWheelListener is added and removed correctly.
- Ensured that the addition and removal of this listener have no side effects on the functionality of WorkspaceSvg.

### Documentation

- No major documentation updates required as these changes are internal and pertain to fixing a specific bug.

### Additional Information

N/A

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -248,6 +248,12 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
     | null = null;
 
   /**
+   * A dummy wheel event listener used as a workaround for a Safari scrolling issue.
+   * Set in createDom and used for removal in dispose to ensure proper cleanup.
+   */
+  private dummyWheelListener: (() => void) | null = null;
+
+  /**
    * In a flyout, the target workspace where blocks should be placed after a
    * drag. Otherwise null.
    *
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
This looks great =) Thank you for the fix! Once CI passes I'll get this merged.

PR #7698, opened 2023-12-06
feat: connection highlighting in geras and thrasos

BeksOmega:
## The basics

### Resolves

Work on #7204 

### Proposed Changes

Adds logic to geras and thrasos to highlight their connections.

Added logic to the base renderer to support code reuse.

Removed the interface since I added the logic to the base renderer.

### Reason for Changes

Needed to move highlighting logic to the renderer so it can differ on a renderer-by-renderer-basis.

### Test Coverage

Tested that connections in thrasos and geras are still properly highlighted.

### Documentation

N/A

### Additional Information

N/A


core/renderers/common/drawer.ts
```diff
@@ -440,4 +442,73 @@ export class Drawer {
...
+    }
+    const block = conn.getSourceBlock();
+    if (block.pathObject.addConnectionHighlight) {
+      block.pathObject.addConnectionHighlight(
+        conn,
+        path,
+        conn.getOffsetInBlock(),
+        block.RTL,
+      );
+    }
```
cpcallen, line 470: Can be simplified by use of optional chaining.  Will leave to your judgement as I'm not sure if it's more readable (though I suspect that with familiarity it will become so):
```suggestion
    block.pathObject.addConnectionHighlight?.(
      conn,
      path,
      conn.getOffsetInBlock(),
      block.RTL,
    );
```

core/renderers/common/info.ts
```diff
@@ -748,4 +748,24 @@ export class RenderInfo {
     this.startY = this.topRow.capline;
     this.bottomRow.baseline = yCursor - this.bottomRow.descenderHeight;
   }
+
+  /** Returns the connection measurable associated with the given connection. */
+  getMeasureableForConnection(conn: RenderedConnection): Connection | null {
+    if (
+      this.outputConnection &&
+      this.outputConnection.connectionModel === conn
```
cpcallen, line 756: Assuming that `conn` can never be `undefined` (which I think it can't since it's a `RenderedConnection`):
```suggestion
      this.outputConnection?.connectionModel === conn
```

core/renderers/common/renderer.ts
```diff
@@ -257,6 +257,28 @@ export class Renderer implements IRegistrable {
...
+
+    drawer.drawConnectionHighlightPath(conn);
+  }
+
+  /** Visually unhighlights the given connection, if it is highlighted. */
+  unhighlightConnection(conn: RenderedConnection): void {
+    const pathObject = conn.getSourceBlock().pathObject;
+    if (pathObject.removeConnectionHighlight) {
+      pathObject.removeConnectionHighlight(conn);
+    }
```
cpcallen, line 279:
```suggestion
   conn.getSourceBlock().pathObject.removeConnectionHighlight?.(conn);
```
(Assuming that `.getSourceBlock()` alwasy returns an object with a `.pathObject` property that is also an object, otherwise some more `.?`s are in order, though I imagine tsc would have told you already if so.)

core/renderers/common/renderer.ts
```diff
@@ -257,6 +257,28 @@ export class Renderer implements IRegistrable {
...
+   */
+  highlightConnection(conn: RenderedConnection): void {
+    const block = conn.getSourceBlock();
+
+    const info = this.makeRenderInfo_(block);
+    info.measure();
+    const drawer = this.makeDrawer_(block, info);
+
+    drawer.drawConnectionHighlightPath(conn);
+  }
```
cpcallen, line 272: Not sure why this couldn't be made idempotent.

scripts/migration/js2ts
```diff
@@ -80,72 +80,82 @@ for (const filename of filenames) {
...
+      name, // Name of named import of whole module (if applicable).
+      names, // {}-enclosed list of destructured imports.
+      type, // If truthy, it is a requireType not require.
+      moduleId, // goog.module ID that was goog.require()d.
+    ) {
+      const importPath = modulePaths[moduleId];
+      type = type ? ' type' : '';
+      if (!importPath) {
+        console.warn(
+          `Unable to migrate goog.require('${moduleId}') as no ES module path known.`,
```
cpcallen, line 107: Darn it, Prettier, you don't have to accept my approach to exceeding line length, but could you at least provide a better alternative?!

cpcallen (commented):
My main concern is about idempotency (or not) of some of the new methods introduced (see below).

Also not sure why Prettier is suddenly attacking `jst2s` especially as it appears you did not change anything in that script.

core/rendered_connection.ts
```diff
@@ -305,57 +293,15 @@ export class RenderedConnection extends Connection {
...
-        'd': steps,
-        'transform':
-          `translate(${offset.x}, ${offset.y})` +
-          (this.sourceBlock_.RTL ? ' scale(-1 1)' : ''),
-      },
-      this.sourceBlock_.getSvgRoot(),
-    );
+    const renderer = this.getSourceBlock().workspace.getRenderer();
+    renderer.unhighlightConnection(this);
+    renderer.highlightConnection(this);
```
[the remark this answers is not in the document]
BeksOmega: Ah whoops I think I actually fixed this so it is idempotent already. We early return in `PathObject.prototype.addConnectionHighlight` now. I just forgot to remove the `unhighlight` call. Thanks for the catch!

core/renderers/common/constants.ts
```diff
@@ -53,8 +53,8 @@ export interface PuzzleTab {
   type: number;
   width: number;
   height: number;
-  pathDown: string | ((p1: number) => string);
-  pathUp: string | ((p1: number) => string);
+  pathDown: string;
+  pathUp: string;
```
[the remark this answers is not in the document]
BeksOmega: The types were just incorrect. This never accepted `number => string`. If you passed that, your svg path would break.

core/renderers/common/drawer.ts
BeksOmega: Huh I didn't know you could do optional chaining in that location. I thought it was only for dot referencing, not method calls. Sweet!

scripts/migration/js2ts
cpcallen: Could you actually fix this, please?

Also, do you know why this file has been reformatted in this PR, and not earlier?

cpcallen (approved):
LGTM except for line length issue in `jst2ts`.

scripts/migration/js2ts
BeksOmega: I think since we have a system for autoformatting, it makes more sense to let it do the formatting, even if we don't 100% agree with it. This saves everybody time.

> Also, do you know why this file has been reformatted in this PR, and not earlier?

Not sure, maybe because it doesn't have a file extension it wasn't being detected previously? We did update to prettier 3.1 recently, but I can't find anything relevant in the [release notes](https://prettier.io/blog/2023/11/13/3.1.0.html#other-changes).

Issue #7649, opened 2023-11-14
Mocha test failures not reported by `test/mocha/webdriver.js`


cpcallen:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When running the mocha tests, our webdriver harness correctly reports the number of test failures but not any details.

### Reproduction steps

1. Modify any of the test files in `tests/mocha/*_test.js` such that one or more test will fail.  This can be done by changing one of the arguments to any `chai.assert.equal` call such that the assertion will fail.
2. Run `npm run build && node tests/mocha/webdriver.js` (or `npm test` if you're not in a hurry).
3. Observe output.

**Expected output**

Per code in `tests/mocha/webdriver.js`, the "Blockly Mocha Test Failures" section of the output is expected to be non-empty: it should contain either useful error messages from the controlled Chrome instance, or the string "There is at least one test failure, but no messages reported. Mocha may be failing because no tests are being run.".

**Actual output**

That section contains only blank lines (one per failure):
```
Loading URL: file:///Users/cpcallen/src/blockly/tests/mocha/index.html
============Blockly Mocha Test Failures================


============Blockly Mocha Test Summary=================
2 tests failed
============Blockly Mocha Test Summary=================
Mocha tests failed
```
Technically correct (is not empty, so does not get the "no messages" message), but not the best kind of correct.

### Stack trace

There are a number of stack traces in the browser console, but they all appear to be directly the result of tests that elicit errors rather than being related to this issue (e.g., mocha not being able to find the `#failureMessages` element to write to or other problems of that nature).

### Browsers

Chrome desktop

maribethb:
Pretty sure this used to work, the likely candidate for breaking this was I think we changed something about how mocha was being run with webdriver around the time we added the browser tests

btw17:
I'll take a look at this one!

btw17:
So the gist of the issue is that the message exists in the page, though it's beneath `#failureMessages` which has `display: none`. We're currently using `getText()` from `webdriverio` which only displays text of interactable elements.

https://webdriver.io/docs/api/element/getText

I'll get a workaround out soon.

PR #7706, opened 2023-12-08
fix: updated mocha tests to display error messages

btw17:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7649

### Proposed Changes

Instead of using `getText()`, which requires a visible element, I use `getHTML()` and strip out the `p` tag.
    - See https://webdriver.io/docs/api/element/getText and **note:** only visible elements will return display.
    - The root element with id `failureCount` has `display: none`, so the text content here never displays.

Previous output:
```
Loading URL: file:///usr/local/google/home/blaketw/git/blockly/blockly/tests/mocha/index.html
============Blockly Mocha Test Failures================

============Blockly Mocha Test Summary=================
1 tests failed
============Blockly Mocha Test Summary=================
Mocha tests failed
```

Updated  output:
```
Loading URL: file:///usr/local/google/home/blaketw/git/blockly/blockly/tests/mocha/index.html
============Blockly Mocha Test Failures================
"Comments Set/Get Bubble Size Set Size While Invisible" failed: expected 100 to equal 101
============Blockly Mocha Test Summary=================
1 tests failed
============Blockly Mocha Test Summary=================
Mocha tests failed
```

### Reason for Changes

To resolve https://github.com/google/blockly/issues/7649

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

- Not sure if `#failureCount` was previously visible, though another way to mitigate this would be to set it visible

### DIFF
--- a/tests/mocha/webdriver.js
+++ b/tests/mocha/webdriver.js
@@ -64,7 +64,7 @@ async function runMochaTestsInBrowser() {
     }
     for (const el of failureMessagesEls) {
       const messageHtml = await el.getHTML();
-      console.log(messageHtml.replace("<p>", "").replace("</p>", ""));
+      console.log(messageHtml.replace('<p>', '').replace('</p>', ''));
     }
   }
 
--- a/tests/mocha/webdriver.js
+++ b/tests/mocha/webdriver.js
@@ -63,7 +63,8 @@ async function runMochaTestsInBrowser() {
       console.log('There is at least one test failure, but no messages reported. Mocha may be failing because no tests are being run.');
     }
     for (const el of failureMessagesEls) {
-      console.log(await el.getText());
+      const messageHtml = await el.getHTML();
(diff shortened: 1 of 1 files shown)


tests/mocha/webdriver.js
```diff
@@ -63,7 +63,8 @@ async function runMochaTestsInBrowser() {
       console.log('There is at least one test failure, but no messages reported. Mocha may be failing because no tests are being run.');
     }
     for (const el of failureMessagesEls) {
-      console.log(await el.getText());
+      const messageHtml = await el.getHTML();
+      console.log(messageHtml.replace("<p>", "").replace("</p>", ""));
```
maribethb, line 67: nit: could you replace the double quotes with single quotes? it appears we don't format this file but that would match the rest of the file. Thanks!

btw17: Done!

maribethb (approved):

PR #7707, opened 2023-12-08
fix: JS random int block with non-ints.

NeilFraser:
Previously random_int(2.4, 2.6) would return 2.  This is wrong.  NaN is better; it's the same as random_int("hello", []).

Python, Lua, PHP all have their own native functions for random int, so we just use their native behaviours.  Dart is another matter (separate PR).

I think making this block more accurate is worth the code given that a LOT of kids use "random integer" for stuff.

Also re-flow the Math test blocks, they were overlapping. And the Dart interpreter link has changed URLs.

### DIFF
--- a/generators/javascript/math.ts
+++ b/generators/javascript/math.ts
@@ -414,11 +414,19 @@ export function math_random_int(
     'mathRandomInt',
     `
 function ${generator.FUNCTION_NAME_PLACEHOLDER_}(a, b) {
+  a = Number(a);
+  b = Number(b);
   if (a > b) {
     // Swap a and b to ensure a is smaller.
     var c = a;
     a = b;
     b = c;
+  }
+  a = Math.ceil(a);
+  b = Math.floor(b);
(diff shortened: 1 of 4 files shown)


BeksOmega:
This is a breaking change right?

NeilFraser:
Technically, yes.  But would you consider adding a guard clause to catch illegal input to a function a breaking change?

BeksOmega:
> Technically, yes. But would you consider adding a guard clause to catch illegal input to a function a breaking change?

For stuff written by SWEs no, but for stuff written by 3rd graders that may not know what they're doing, I'm a bit more cautious. I don't want to break kids' projects without telling them.

Personally, I think it's better for this to succeed with a dumb value rather than trigger NaN errors in the console that kids may not be able to access.

BeksOmega:
@NeilFraser did you have strong opinions on this? If not I'm going to close it on Friday.

PR #7716, opened 2023-12-14
chore: Resync develop's msgs from translatewiki.

NeilFraser:
These are manual changes to fix a large number of messages.  It's a roll-up of work to synchronize the two branches.  Translatewiki read from 'develop' and push to 'translatewiki'.


BeksOmega (approved):

BeksOmega:
Make sure to be careful when submit to use the PR title! I've missed this a few times recently.

PR #7727, opened 2023-12-18
fix(generators): Fix generator typings

### LINKED ISSUE #7652: Published generator typings are wrong
## Problem

At the moment we are shipping hand-written `.d.ts` files for the generators which are pretty deficient.  For example, `javacript.d.ts` (copied into the package from [`typings/javascript.d.ts`](https://github.com/google/blockly/blob/develop/typings/javascript.d.ts)) is essentially:

```TS

export enum Order {
  ATOMIC = 0,            // 0 "" ...
  // etc.
  NONE = 99,             // (...)
}

export declare const javascriptGenerator: any;
```

Note that `javascriptGenerator` is declared as `any`, and `JavascriptGenerator` is not declared at all.

This means that developers using generators from TypeScript are not getting the full benefit of TS's type checking of the generator classes / instances.

## Proposed Solution

Publish the `.d.ts` files produced by `tsc` instead of these handwritten versions.

* [x] #6828
* [x] Decide on what form the generator typings should be published in.
* [x] Modify `build_tasks.js` and `package_tasks.js` accordingly.
* [x] Remove unused files from `typings/*.js`.

### Discussion

* Should we publish individual `.d.ts` files in `dist/generators/**/*.d.ts` (as we do for files in core), with a wrapper that just imports them, or should we try to publish a single file per chunk that rolls these up?
  * There may be some difficult in publishing a roll up as it will need to reference `CodeGenerator`, which is currently defined in `core/`.

See also #5818.

cpcallen:
## The basics

### Resolves

Fixes #7652.

### Proposed Changes

* Modify `package_tasks.js` to include `tsc`-generated `.d.ts` files for the generators in `dist` when building the `blockly` NPM package.
* Replace the hand-written `.d.ts` files for the generator chunks with ones that reexport the types from the corresponding `tsc`-generated `.d.ts` file.

### Reason for Changes

Publish accurate typings for the generators.

### Additional Information

BREAKING CHANGE: Previously the `dartGenerator`, `luaGenerator`, `javascriptGenerator`, `phpGenerator`, and `pythonGenerator` exports (from the `blockly/dart`, `blockly/lua`, `blockly/javascript`, `blockly/php` and `blockly/python` entrypoints respectively) were being typed `any`.  Now their types are correct.  There is no change to their runtime behaviour, but developers who have written code in TypeScript that manipulates generators (adding custom block generator functions, for example) may find they need to correct type error in their code in order for it to compile correctly with Blockly v11.

Notably, block  generator functions to be installed in the `javascriptGenerator.forBlock` dictionary should be typed
```
(block: Blockly.Block, generator: javascript.JavascriptGenerator) => [string, javascript.Order] | string | null
```
—and similarly for each of the other four languages we support.

### DIFF
--- a/scripts/gulpfiles/package_tasks.js
+++ b/scripts/gulpfiles/package_tasks.js
@@ -348,7 +348,6 @@ function packageDTS() {
   return gulp.src(handwrittenSrcs, {base: 'typings'})
       .pipe(gulp.src(`${TYPINGS_BUILD_DIR}/**/*.d.ts`, {ignore: [
 	`${TYPINGS_BUILD_DIR}/blocks/**/*`,
-	`${TYPINGS_BUILD_DIR}/generators/**/*`,
       ]}))
       .pipe(gulp.replace('AnyDuringMigration', 'any'))
       .pipe(gulp.dest(RELEASE_DIR));
--- a/typings/dart.d.ts
+++ b/typings/dart.d.ts
@@ -4,25 +4,4 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export enum Order {
-  ATOMIC = 0,         // 0 "" ...
-  UNARY_POSTFIX = 1,  // expr++ expr-- () [] . ?.
-  UNARY_PREFIX = 2,   // -expr !expr ~expr ++expr --expr
-  MULTIPLICATIVE = 3, // * / % ~/
-  ADDITIVE = 4,       // + -
-  SHIFT = 5,          // << >>
-  BITWISE_AND = 6,    // &
-  BITWISE_XOR = 7,    // ^
-  BITWISE_OR = 8,     // |
-  RELATIONAL = 9,     // >= > <= < as is is!
-  EQUALITY = 10,      // == !=
-  LOGICAL_AND = 11,   // &&
-  LOGICAL_OR = 12,    // ||
-  IF_NULL = 13,       // ??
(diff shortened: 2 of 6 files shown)


scripts/gulpfiles/package_tasks.js
```diff
@@ -348,7 +348,6 @@ function packageDTS() {
   return gulp.src(handwrittenSrcs, {base: 'typings'})
       .pipe(gulp.src(`${TYPINGS_BUILD_DIR}/**/*.d.ts`, {ignore: [
 	`${TYPINGS_BUILD_DIR}/blocks/**/*`,
-	`${TYPINGS_BUILD_DIR}/generators/**/*`,
```
BeksOmega, line 351: Shouldn't this still exist if this is where the reexports are coming from?

cpcallen: It is being deleted from the `ignore:` clause, thus causing it to be included in what gets copied from `build/` to `dist/`.

BeksOmega: haha ok sweet sorry about that.

BeksOmega (approved):

cpcallen:
Darn it: this PR should have had a "!" in its title, and so especially should the squash commit, which didn't get a "BREAKING CHANGE:" text in the description.  I'm going to rename the PR in the hopes that that is enough to make the release notes come out correctly.  🤞

PR #7729, opened 2023-12-19
chore(deps)!: Bump JSDOM to v23, dropping node 16.

BeksOmega:
Rerolls #7667 to bump JSDOM which bumps our min version of node.

## Breaking changes

Drops support for Node v16.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
-        "jsdom": "22.1.0"
+        "jsdom": "23.0.0"
       },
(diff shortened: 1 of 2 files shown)


cpcallen (approved):
LGTM, but I wonder if we should start providing an [`engines`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) clause in `package.json`.

BeksOmega:
> LGTM, but I wonder if we should start providing an [`engines`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) clause in `package.json`.

Seems reasonable! Done.

PR #7730, opened 2023-12-19
fix!: have insertion markers use json deserialization

### LINKED ISSUE #7316: Change insertion markers to use JSON deserialization when duplicating blocks

BeksOmega:
## The basics

### Resolves

Fixes #7316

### Proposed Changes

Reverts https://github.com/google/blockly/pull/7430 https://github.com/google/blockly/pull/7429

Makes it so that insertion markers are created using JSON deserialization.

### Reason for Changes

From https://github.com/google/blockly/pull/7364

> Insertion markers duplicate the block being dragged, and then tell the duplicate to be an insertion marker.

> Currently the duplication duplicates (ha) a lot of code from the serialization system. We think this is because in the past the XML system didn't have a good way to just serialize and deserialize a single block (without children). But the JSON system does have this!

> Sparked by this discussion: https://groups.google.com/g/blockly/c/gcvynl88QLY/m/shFyX4x0BwAJ

### Test Coverage

All tests pass.

### Additional Information

N/A

## Breaking changes / To fix

Insertion-marker-ness is now set after any mutation is deserialized.

If you have a block whose behavior depeonds on whether it is an insertion marker or not, they should check `doFullSerialization` instead. Insertion markers pass `doFullSerialization` as false.

You can see the [shareable procedure blocks](https://github.com/google/blockly-samples/blob/4e28e47679990f5deddfe312347c7942d7f434da/plugins/block-shareable-procedures/src/blocks.ts#L487) for an example of round-tripping whether full serialization was enabled from `saveExtraState` to `loadExtraState`.

### DIFF
--- a/core/insertion_marker_manager.ts
+++ b/core/insertion_marker_manager.ts
@@ -17,7 +17,7 @@ import type {BlockSvg} from './block_svg.js';
 import * as common from './common.js';
 import {ComponentManager} from './component_manager.js';
 import {config} from './config.js';
-import * as constants from './constants.js';
+import * as blocks from './serialization/blocks.js';
 import * as eventUtils from './events/utils.js';
 import type {IDeleteArea} from './interfaces/i_delete_area.js';
 import type {IDragTarget} from './interfaces/i_drag_target.js';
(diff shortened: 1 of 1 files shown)


maribethb (approved):
LGTM, but please fix the description of the PR to include the directions for breaking changes directly instead of describing what would be in the release notes

PR #7743, opened 2024-01-03
feat!: modify icons to use the rendering queue

### LINKED ISSUE #7068: Make icon bubble use the rendering queue

BeksOmega:
## The basics

### Resolves

Fixes #7068 

### Proposed Changes

Makes it so the icons' `setBubbleVisible` methods wait for the current render to complete before showing their bubbles. 

### Reason for Changes

This way we know the location information is updated so the bubbles will show up in the right spot, but we're also not triggering an immediate render, which is bad for performance.

### Test Coverage

Updated tests to use promises.

### Documentation

Need to check docs on custom icons. Also needs changes in the release notes.

### Additional Information

N/A

## Breaking changes / To fix

If you are calling `setBubbleVisible` and then accessing the contents of the open bubble, you now need to wait for the promise returned by `setBubbleVisible` to resolve before accessing the contents.

```js
function openBubbleAndDoStuff(block) {
  const icon = block.getIcon(Blockly.icons.IconType.MUTATOR);
  icon.setBubbleVisible(true);
  icon.getWorkspace().doStuff();
}
```

Becomes

```js
async function openBubbleAndDoStuff(block) {
  const icon = block.getIcon(Blockly.icons.IconType.MUTATOR);
  await icon.setBubbleVisible(true);
  icon.getWorkspace().doStuff();
}
```

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1029,7 +1029,6 @@ export class BlockSvg
       icon.applyColour();
       icon.updateEditable();
       this.queueRender();
-      renderManagement.triggerQueuedRenders();
       this.bumpNeighbours();
     }
 
@@ -1058,7 +1057,6 @@ export class BlockSvg
 
     if (this.rendered) {
       this.queueRender();
-      renderManagement.triggerQueuedRenders();
       this.bumpNeighbours();
     }
     return removed;
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -22,6 +22,7 @@ import {Svg} from '../utils/svg.js';
 import {TextBubble} from '../bubbles/text_bubble.js';
 import {TextInputBubble} from '../bubbles/textinput_bubble.js';
 import type {WorkspaceSvg} from '../workspace_svg.js';
+import * as renderManagement from '../render_management.js';
 
 /** The size of the comment icon in workspace-scale units. */
 const SIZE = 17;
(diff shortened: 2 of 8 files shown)


maribethb (approved):
Could you go ahead and add the info to this PR description about what users will need to fix to accommodate this breaking change? I'm not sure if the auto-release notes will figure out how to link to this PR directly or not, but I'll assume everything in the v11 branch will get linked.

Issue #4288, opened 2020-09-17
Look into overloaded rendered field on block


alschmiedt:
**Is your feature request related to a problem? Please describe.**
Creating a field that forces a re render in the initModel method causes an error when mutating the block. (See example in #3458). This specific instance is fixed, but it is still plausible that someone could create a field that forces a re render which would cause this problem. To recreate Set this.[doValueUpdate_](https://github.com/google/blockly/blob/9ab6273e4c2fb0cd25004650d0020490294eed50/core/field_variable.js#L133) back to setValue.

The reason is that when we mutate the block, we are setting the rendered property on the block to false in an attempt to hold off rendering until the block has been fully created (not doing so will cause #4175). However, setting rendered = false causes the fields init methods to not get called. 

When the block is finally ready to be rendered it will go through and call the init methods on all the fields. Since the init methods have not been called before when we reach a field that forces a re render of the entire block(in the example  in #4175 this is the field_variable) the fields that have not yet been initialized will throw an error.

**Describe the solution you'd like**
All this points to our blocks rendered property being overloaded as described in [this comment](https://github.com/google/blockly/issues/4175#issuecomment-674250257) and at the [bottom of this comment](https://github.com/google/blockly/pull/3468#issuecomment-562292414). By setting some kind of flag for when we are making batch changes to a block we can avoid this issue. However, there are quite a few other issues that are around this that we should look into before deciding.

**Describe alternatives you've considered**

**Additional context**

BeksOmega:
Thanks for the issue! I think this is a good summary :D https://github.com/google/blockly/issues/1676 Also discusses some possible meanings / uses of `rendered` that might be useful for deciding how to approach this.

BeksOmega:
After v11 we will use `rendered` to check whether a block is a `Block` instance or a `BlockSvg` instance. `initialized` will be used to check if `initSvg` or (inclusive) `initModel` has been called.

PR #7745, opened 2024-01-04
fix: snap immediately

BeksOmega:
## The basics

### Resolves

Work on #5146  #4288 

### Proposed Changes

Makes it so that snapping to the grid happens immediately instead of after a delay.

### Reason for Changes

I am working on refactoring how block initialization works. As part of it, I need to have `bumpNeighbors` be triggered by the render management system, instead of being triggered by each individual change that might change the shape of the block.

But before `bumpNeighbors` gets called, I need to make sure the snap has happened. So I'm getting rid of the delay.

### Test Coverage

Manually tested that snapping works properly, and looks correct when dropping after dragging.

### Documentation

N/A

### Additional Information

Dependent on #7743

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -455,23 +455,12 @@ export class BlockSvg
 
   /** Snap this block to the nearest grid point. */
   snapToGrid() {
-    if (this.isDeadOrDying()) {
-      return; // Deleted block.
-    }
-    if (this.workspace.isDragging()) {
-      return; // Don't bump blocks during a drag.
-    }
-
-    if (this.getParent()) {
-      return; // Only snap top-level blocks.
-    }
-    if (this.isInFlyout) {
-      return; // Don't move blocks around in a flyout.
-    }
+    if (this.isDeadOrDying()) return;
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -455,23 +455,12 @@ export class BlockSvg
 
   /** Snap this block to the nearest grid point. */
   snapToGrid() {
-    if (this.isDeadOrDying()) {
-      return; // Deleted block.
-    }
-    if (this.workspace.isDragging()) {
-      return; // Don't bump blocks during a drag.
-    }
```
cpcallen, line 463: Is it safe to remove this check?

cpcallen (commented):
Looks like I should ultimately only need to review the second of the two commits in this PR, correct?

core/block_svg.ts
BeksOmega: Yeah there's no reason this should be called while the workspace is dragging. It only gets called when we click a block in the flyout to create it, or when we finish dragging a block.

PR #7747, opened 2024-01-04
fix!: `rendered` meaning

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/4288

### Proposed Changes

Makes it so that `rendered` is constant, and is only used to check whether the block is a `Block` or a `BlockSvg`.

### Reason for Changes

Makes the system easier to reason about and understand.

### Test Coverage

Fixed change detectors.

### Documentation

This will need docs. I have separate tasks filed for that.

### Additional Information

This PR creates a pretty significant performance degredation related to `bumpNeighbours` I have a separate fix in the works for that. But in an effort to keep my PRs small I'm putting that up separately.

There's also more work to be removing references to `initSvg` and `rendered` and cleaning up `initSvg`.

Dependent on #7745 

## Breaking changes / To fix

If you are using `rendered` to check whether the block is initialized or currently visible, do not do that. `rendered` should only be used to tell if the block is a `Block` instance or a `BlockSvg` instance.

If you are setting `rendered` to temporarily disable rendering and improve performance, that is no longer necessary because we perform rendering batching.

If you have another use case for accessing `rendered` that is not covered here, please let us know on the forum.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -662,11 +662,8 @@ export class BlockSvg
    * @internal
    */
   updateComponentLocations(blockOrigin: Coordinate) {
-    if (!this.rendered) {
-      // Rendering is required to lay out the blocks.
-      // This is probably an invisible block attached to a collapsed block.
-      return;
-    }
+    // TODO: make sure this doesn't have a performance impact
+    // when deserializing.
 
     if (!this.dragging) this.updateConnectionLocations(blockOrigin);
     this.updateIconLocations(blockOrigin);
@@ -801,7 +798,7 @@ export class BlockSvg
     Tooltip.dispose();
     ContextMenu.hide();
 
-    if (animate && this.rendered) {
+    if (animate) {
       this.unplug(healStack);
       blockAnimations.disposeUiEffect(this);
     }
@@ -1014,13 +1011,11 @@ export class BlockSvg
     if (icon instanceof WarningIcon) this.warning = icon;
     if (icon instanceof MutatorIcon) this.mutator = icon;
 
-    if (this.rendered) {
-      icon.initView(this.createIconPointerDownListener(icon));
-      icon.applyColour();
-      icon.updateEditable();
-      this.queueRender();
-      this.bumpNeighbours();
-    }
+    icon.initView(this.createIconPointerDownListener(icon));
+    icon.applyColour();
+    icon.updateEditable();
+    this.queueRender();
+    this.bumpNeighbours();
 
     return icon;
   }
@@ -1045,10 +1040,9 @@ export class BlockSvg
     if (type.equals(WarningIcon.TYPE)) this.warning = null;
     if (type.equals(MutatorIcon.TYPE)) this.mutator = null;
 
-    if (this.rendered) {
-      this.queueRender();
-      this.bumpNeighbours();
-    }
+    this.queueRender();
+    this.bumpNeighbours();
+
     return removed;
   }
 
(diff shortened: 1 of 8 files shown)


core/block_svg.ts
```diff
@@ -659,11 +662,8 @@ export class BlockSvg
    * @internal
    */
   updateComponentLocations(blockOrigin: Coordinate) {
-    if (!this.rendered) {
-      // Rendering is required to lay out the blocks.
-      // This is probably an invisible block attached to a collapsed block.
-      return;
-    }
+    // TODO: make sure this doesn't have a performance impact
```
maribethb, line 665: do you have a bug for this, or already addressing in the follow up PR you mentioned?

core/block_svg.ts
```diff
@@ -1276,13 +1256,8 @@ export class BlockSvg
    */
   override removeInput(name: string, opt_quiet?: boolean): boolean {
     const removed = super.removeInput(name, opt_quiet);
-
-    if (this.rendered) {
-      this.queueRender();
-      // Removing an input will cause the block to change shape.
```
maribethb, line 1282: nit: you may want to keep this comment explaining why we call bumpNeighbors (here and below, and maybe above where it wasn't before). I think it's moderately helpful... but I see you also removed it from other places so if you disagree that's fine

core/inputs/input.ts
```diff
@@ -123,7 +123,6 @@ export class Input {
 
     if (this.sourceBlock.rendered) {
```
maribethb, line 124: would it be possible to just check `instanceof BlockSvg` then you wouldn't need to cast either? circular dependencies are the only reason i can think of you wouldn't do that but i'm hoping that doesn't apply here

core/rendered_connection.ts
```diff
@@ -554,29 +550,10 @@ export class RenderedConnection extends Connection {
...
-    }
-    if (parentRendered && childRendered) {
-      if (
-        this.type === ConnectionType.NEXT_STATEMENT ||
-        this.type === ConnectionType.PREVIOUS_STATEMENT
-      ) {
-        // Child block may need to square off its corners if it is in a stack.
-        // Rendering a child will render its parent.
-        childBlock.queueRender();
-      } else {
```
maribethb, line 574: I'm curious why you got rid of this logic - just not enough of a performance boost to justify it? or was not rendering the child block in all cases a mistake?

core/block_svg.ts
```diff
@@ -126,7 +126,10 @@ export class BlockSvg
   style: BlockStyle;
   /** @internal */
   pathObject: IPathObject;
-  override rendered = false;
+
+  /** Is this block a BlockSVG? */
+  override readonly rendered = true;
```
maribethb, line 131: If the only purpose of this is to detect whether it's a `BlockSvg` I would rather just remove it. less chance of it being used mistakenly. an instanceof check for `BlockSvg` would be preferred.

core/block_svg.ts
BeksOmega: Thanks for the catch! I checked this but forgot to delete.

core/block_svg.ts
BeksOmega: In the next PR I move `bumpNeighbours` to be called as part of rendering. Whenever we rerender we should bump neighbours. So no need to have comments scattered in just some of the places.

core/inputs/input.ts
BeksOmega: It causes circular dependencies.

core/rendered_connection.ts
BeksOmega: Not rendering the child in all cases was a mistake. Since we added custom rendering, value input blocks may also change shape depending on if they're connected to a parent or not.

core/inputs/input.ts
maribethb: :(

PR #7748, opened 2024-01-05
fix: bump neighbours performance regression

BeksOmega:
## The basics

### Resolves

N/A

### Proposed Changes

Makes it so that `bumpNeighbours` is called as part of the rendering cycle.

### Reason for Changes

1) You no longer have to remember to call `bumpNeighbours` because it just happens whenever the block changes shape.
2) Alleviates a performance regression caused by https://github.com/google/blockly/pull/7747 . The regression was that `bumpNeighbours` was being called on every single block initialization action during deserialization, which meant a ton of `setTimeout` calls, which were slowing everything down. Making `bumpNeighbours` part of serialization gets rid of these `setTimeout` calls.

### Test Coverage

Manually tested.
Manually profiled performance.
Also covered by some existing unit tests (caught the issue with event grouping!)

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/7747

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -558,9 +558,10 @@ export class Block implements IASTNodeLocation, IDeletable {
   /**
    * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
    * connected should not coincidentally line up on screen.
+   * 
+   * @deprecated v11 - You should not need to call bump neighbours directly.
    */
   bumpNeighbours() {}
-  // noop.
 
   /**
    * Return the parent block or null if this block is at the top level. The
--- a/core/block.ts
+++ b/core/block.ts
@@ -558,7 +558,7 @@ export class Block implements IASTNodeLocation, IDeletable {
   /**
    * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
    * connected should not coincidentally line up on screen.
-   * 
+   *
    * @deprecated v11 - You should not need to call bump neighbours directly.
    */
   bumpNeighbours() {}
--- a/core/block.ts
+++ b/core/block.ts
@@ -558,8 +558,6 @@ export class Block implements IASTNodeLocation, IDeletable {
   /**
    * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
    * connected should not coincidentally line up on screen.
-   *
-   * @deprecated v11 - You should not need to call bump neighbours directly.
    */
   bumpNeighbours() {}
 
--- a/core/block_dragger.ts
+++ b/core/block_dragger.ts
@@ -287,13 +287,13 @@ export class BlockDragger implements IBlockDragger {
    */
   protected updateBlockAfterMove_() {
     this.fireMoveEvent_();
-    this.draggingBlock_.snapToGrid();
     if (this.draggedConnectionManager_.wouldConnectBlock()) {
       // Applying connections also rerenders the relevant blocks.
       this.draggedConnectionManager_.applyConnections();
     } else {
       this.draggingBlock_.queueRender();
     }
+    this.draggingBlock_.snapToGrid();
   }
 
   /** Fire a UI event at the end of a block drag. */
--- a/core/block_dragger.ts
+++ b/core/block_dragger.ts
@@ -287,13 +287,13 @@ export class BlockDragger implements IBlockDragger {
    */
   protected updateBlockAfterMove_() {
     this.fireMoveEvent_();
+    this.draggingBlock_.snapToGrid();
     if (this.draggedConnectionManager_.wouldConnectBlock()) {
       // Applying connections also rerenders the relevant blocks.
       this.draggedConnectionManager_.applyConnections();
     } else {
       this.draggingBlock_.queueRender();
     }
-    this.draggingBlock_.scheduleSnapAndBump();
   }
 
   /** Fire a UI event at the end of a block drag. */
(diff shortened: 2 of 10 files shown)


core/gesture.ts
```diff
@@ -963,7 +963,8 @@ export class Gesture {
           eventUtils.setGroup(true);
         }
         const newBlock = this.flyout.createBlock(this.targetBlock);
-        newBlock.scheduleSnapAndBump();
+        newBlock.snapToGrid();
```
maribethb, line 966: this feels suspicious to me... if we're telling external users they don't need to call `bumpNeighbors` directly, we shouldn't either when we're creating a new block in this manner.

I don't think we need it anyway. `createBlock` uses json serialization to `append` new block to the workspace. `append` will eventually queue a render on the block, which will now call these two things. so calling them here should be redundant, I think.

maribethb (approved):

core/block_dragger.ts
```diff
@@ -287,13 +287,13 @@ export class BlockDragger implements IBlockDragger {
    */
   protected updateBlockAfterMove_() {
     this.fireMoveEvent_();
+    this.draggingBlock_.snapToGrid();
```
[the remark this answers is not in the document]
BeksOmega: Done.

core/gesture.ts
BeksOmega: Ah you're right, this is actually a good case for someone calling this externally, so I've undeprecated it.

The problem here is that `createBlock` forces an immediate render (because we always force an immediate render when deserializing). But at the time the render is triggered, the blockj isn't guaranteed to be in the correct place, so the bump isn't guaranteed to be correct.

This would be fixed if `createBlock` just deserialized at the correct spot, but since it's overridable, that's not possible to guarantee.

It would also be possible to trigger a bumping after every block move, but this could easily break external developers' functions for "cleaning up" the workspace. I believe it would even break our built-in context menu option for cleanup, because we space blocks based on the minimum block height, not the bump distance.

So our only option is to snap to the grid after deserializing, and then trigger the bump manually.

maribethb: `createBlock` is currently internal, which... is one of the ones that is incorrect because it's also part of the `IFlyout` interface. But I think it might be fair to fix this in our implementation of `createBlock` and then add it as a requirement that anyone overriding `createBlock` needs to take into account? Because this is going into a major I feel comfortable adding that new constraint. We could even fix it in `placeNewBlock` and/or `positionNewBlock` and then make those methods protected instead of private, so that subclasses of `Flyout` could use them and use our fixed logic, making it pretty easy on the developer to get this right.

By "just deserialized at the correct spot" do you mean we'd calculate the position of where the block should go before deserializing (so add the correct coordinates to the json instead of calling `moveTo` afterwards)? If so I think the approach above could make sense.

Otherwise or if that actually doesn't make sense, then leaving it undeprecated makes sense as well.

BeksOmega:
Found an issue with insertion markers causing bumping that I have a few solutions for, but no clear winner. I'm going to let it sit in my brain over the long weekend and then come back to this tuesday.

core/gesture.ts
BeksOmega: So the specific case for `createBlock` could be worked around by making a breaking change, but I still don't think it's fixable in general.

Folks could still want to move blocks programmatically, and then trigger bumpin neighbours on them, and that requires the `bumpNeighbours` method to be public.

I don't think making the break in `createBlock` makes sense given that, but will await your opinions.

maribethb: Okay, that makes sense too. If we can't trigger it on every block move then it would need to be public regardless. I'm fine with leaving `createBlock` alone then.

core/block.ts
```diff
@@ -558,9 +558,10 @@ export class Block implements IASTNodeLocation, IDeletable {
   /**
    * Bump unconnected blocks out of alignment.  Two blocks which aren't actually
    * connected should not coincidentally line up on screen.
+   *
+   * @deprecated v11 - You should not need to call bump neighbours directly.
```
maribethb, line 562: This should also be removed then I assume? I'm not sure why this method even actually exists on non-rendered blocks in the first place, but... if we were going to deprecate this one it would need a different message referencing the fact that bumping doesn't make sense in non-rendered workspaces

core/block_svg.ts
```diff
@@ -1450,17 +1416,6 @@ export class BlockSvg
     }
   }
 
-  /**
-   * Schedule snapping to grid and bumping neighbours to occur after a brief
-   * delay.
-   *
-   * @internal
```
maribethb, line 1457: Unfortunately we use this in a plugin :/ This also seems like a likely function that people would be using even though it's marked internal. How would you feel about deprecating it instead and removing it at a future date?

maribethb (approved):

core/block_svg.ts
BeksOmega: How do you feel about making `scheduleSnapAndBump` public instead? Since we already established people may want to bump this after programatically moving their blocks.

core/block.ts
BeksOmega: It has to exist for type polymorphism reasons.

core/block_svg.ts
maribethb: That sounds great!

PR #7781, opened 2024-01-11
feat: connection highlight zelos

BeksOmega:
## The basics

### Resolves

Work on #7204 

Makes it so that zelos has the ability to highlight connections, so the eventual connection previewer system can call the renderer to highlight connections consistently (instead of sometimes calling the renderer, and sometimes doing other things).

### Test Coverage

Manually tested highlighting all three shapes of dynamic connections in zelos (hexagon, rounded, and square). Manually tested highlighting non-dynamic connections (puzzle tabs).

### Documentation

N/A

### Additional Information

N/A

### To Review

The drawer architecture is... a bit frustrating. All of the methods are stateful, so I had to pull out some logic for creating internal input paths into pure methods. This way the code could be reused.

Some of the "constants" for path shapes were also broken. Previously `pathDown()` was never being called, so no one noticed that it was busted. I fixed up the logic to actually work for all combinations of right or left side + drawing up or down.

[Edit: docs for context [https://developers.google.com/blockly/guides/create-custom-blocks/renderers/create-custom-renderers/connection-shapes#basic_shapes](https://developers.google.com/blockly/guides/create-custom-blocks/renderers/create-custom-renderers/connection-shapes#basic_shapes)]

### DIFF
--- a/core/renderers/zelos/constants.ts
+++ b/core/renderers/zelos/constants.ts
@@ -392,19 +392,20 @@ export class ConstantProvider extends BaseConstantProvider {
         blockHeight > maxHeight ? blockHeight - maxHeight : 0;
       const height = blockHeight > maxHeight ? maxHeight : blockHeight;
       const radius = height / 2;
+      const sweep = right === up ? '0' : '1';
       return (
         svgPaths.arc(
           'a',
-          '0 0,1',
+          '0 0,' + sweep,
           radius,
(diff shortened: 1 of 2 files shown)


cpcallen (approved):
I wish this was easier to understand (without extensive familiarity with the renderers) but alas I don't see any obvious improvements to suggest.

BeksOmega:
Thank you for the review Christopher! :D

PR #7792, opened 2024-01-18
feat: insertion marker previewer

BeksOmega:
## The basics

- [X] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Work on #7204 

### Proposed Changes + Reasons

Adds the insertion marker previewer so we can stop relying on the insertion marker manager (in a future PR).

### Test Coverage

Manually tested with the next PR that insertion markers and highlights properly appear:

- When inserting in stacks.
- When inserting around stacks.
- When inserting in rows.
- When inserting around rows.
- When replacing stacks.
- When replacing rows.

### Documentation

N/A

### Additional Information

N/A

## To Review

Recommend reveiewing comit-wise.

A lot of this code was adapted from the insertion marker manager, so it may be useful to look at that.

Please let me know if anything needs more docs.

PR #7793, opened 2024-01-18
feat: update block dragger to use connection previewer

BeksOmega:
## The basics

### Resolves

Work on #7204 

Switches the block dragger to use the connection previewer so we can actually swap the logic in and out.

### Test Coverage

See https://github.com/google/blockly/pull/7792

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/7792

## To review

Should be reviewed commit-wise. I did mess up one thing with deleting (blocks weren't actually getting deleted when dropped over delete areas), which is fixed in a further commit.

A lot of this code was adapted from the insertion marker manager, so it may be useful to look at that.

Let me know if anything needs more docs.

### DIFF
--- a/core/block_dragger.ts
+++ b/core/block_dragger.ts
@@ -21,7 +21,6 @@ import * as common from './common.js';
 import type {BlockMove} from './events/events_block_move.js';
 import * as eventUtils from './events/utils.js';
 import type {Icon} from './icons/icon.js';
-import {InsertionMarkerManager} from './insertion_marker_manager.js';
 import type {IBlockDragger} from './interfaces/i_block_dragger.js';
 import type {IDragTarget} from './interfaces/i_drag_target.js';
 import * as registry from './registry.js';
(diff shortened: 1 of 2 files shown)


core/block_dragger.ts
```diff
@@ -173,32 +191,178 @@ export class BlockDragger implements IBlockDragger {
...
+  private wouldDeleteBlock(
+    e: PointerEvent,
+    draggingBlock: BlockSvg,
+    delta: Coordinate,
+  ): boolean {
+    const dragTarget = this.workspace_.getDragTarget(e);
+    if (!dragTarget) return false;
+
+    const componentManager = this.workspace_.getComponentManager();
+    const isDeleteArea = componentManager.hasCapability(
```
NeilFraser, line 235: It's unfortunate that delete areas are hardcoded like this.  It would be nice if there were definable target areas, that would do something on their own if something were dragged into them.  E.g. delete areas, duplication areas, backpacks, stack fragmenters...  Thus the areas would instead return "I'll be handling this block when dropped, don't suggest anything".

NeilFraser (approved):

core/block_dragger.ts
BeksOmega: Yeah we have `onDrop` but we don't actually let drag targets handle the deleting... no idea why. But the block dragger needs to handle deleting for backwards compatibility :/

Issue #7780, opened 2024-01-10
block change event fired if you cancel editing text


maribethb:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

Now that we have intermediate change events, block change events are only fired when the user presses "enter" or otherwise "commits"/"finalizes" text entry in the angle and text input fields.

However, if a user cancels editing by pressing the "escape" key, a change event is fired with an incorrect "old" value. I think most likely no change event should be fired. But if one is fired, the "old" value and "initial" value should be the same.

### Solution

Pass `false` as the second argument [when we call `setValue` in the case of an "escape" keypress](https://github.com/google/blockly/blob/187e4cc3d7f932f571028f680674103359c1d257/core/field_input.ts#L543) so that the event is not fired. `setValue` takes an optional argument to control whether or not the event is fired. If we pass `false` for the argument, the final change event will not be fired.

Follow the reproduction steps again, but this time make sure when you press "escape" that no `BLOCK_CHANGE` event is fired.

### Reproduction steps

1.  Drag a block with a text input field onto a workspace.
2. Enable event logging using the checkbox in the playground, and open the developer console so you can see the events being logged.
3. Start typing into the text input field. See intermediate events being fired.
4. Hit enter and your text will be finalized. See the block change event being fired. The `oldValue` is the initial value before you started typing. The `newValue` is the value you have just typed.
5. Start typing into the text input field again.
6. Hit escape. A block change event will be fired, but this time the `oldValue` is the value that was in the editor just before you hit escape, and the `newValue` is the original value of the block before you started typing.

This `oldValue` for the event in this case does not make sense.

ebrahim95:
Can I be assigned this issue?

BeksOmega:
Go for it @ebrahim95 ! Let us know if you have any questions =)

BeksOmega:
Closed by #7794

PR #7794, opened 2024-01-19
fix: prevent block change event firing if editing is cancelled by using keypress

ebrahim95:
## The basics

### Resolves

 

Fixed the block change event firing. #7780 

### Proposed Changes

Passed false boolean  as a second argument to set_value() function located in the fields_input.ts file

### Reason for Changes

Because the old_value field contains an incorrect value.

### Test Coverage

I tested it out by doing the reproduction steps outlined in the issue #7780, and this time the old value was correct.


BeksOmega (approved):
This looks great! Thank you for the fix @ebrahim95 =)

PR #7795, opened 2024-01-19
fix: highlighting connections in zelos, also highlight connections moving

BeksOmega:
## The basics

### Resolves

Work on #7204 

### Proposed Changes + reason for changes.

Makes it so that zelos highlights connections in the same cases where the base renderer does (no more special casing).

To get this working, I also had to get updating connection highlight locations working. Because in zelos connecting an insertion marker changes the location of a connection. Now if a connection changes location while it is highlighted, the highlight changes location to match.

Now highlighting happens as part of the normal draw loop, so I removed the top-level APIs on the render class. This is not a breaking change because this code hasn't been released yet. I removed it to keep the API smaller.

### Test Coverage

1) Open the playground.
1) Drag a "logic_compare" block into the workspace.
2) Run the following code in the console.
```js
block = Blockly.getMainWorkspace().getTopBlocks()[0];
block.inputList[0].connection.highlight();
block.inputList[0].appendField('test');
```
3) Observe how the highlight is in the correct place.

Also tested with zelos and highlighting works correctly.

### Documentation

Probably need to add renderer docs for highlighting connections

### Additional Information

Dependent on #7793

### DIFF
--- a/core/renderers/common/path_object.ts
+++ b/core/renderers/common/path_object.ts
@@ -273,18 +273,10 @@ export class PathObject implements IPathObject {
     rtl: boolean,
   ) {
     if (this.connectionHighlights.has(connection)) {
-      const currPath = this.connectionHighlights
-        .get(connection)
-        ?.getAttribute('d');
-      const currOffset = this.highlightOffsets.get(connection);
-      if (
-        connectionPath === currPath &&
-        Coordinate.equals(offset, currOffset)
-      ) {
+      if (this.currentHighlightMatchesNew(connection, connectionPath, offset)) {
         return;
-      } else {
-        this.removeConnectionHighlight(connection);
       }
+      this.removeConnectionHighlight(connection);
     }
 
     const highlight = dom.createSvgElement(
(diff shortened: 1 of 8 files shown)


core/renderers/common/path_object.ts
```diff
@@ -40,12 +40,11 @@ export class PathObject implements IPathObject {
   constants: ConstantProvider;
   style: BlockStyle;
 
-  /**
-   * Highlight paths associated with connections.
-   *
-   * @protected
```
maribethb, line 46: Is taking this protected to private a breaking change? Or is this part of the unreleased code?

core/renderers/common/path_object.ts
```diff
@@ -273,7 +272,21 @@ export class PathObject implements IPathObject {
...
+      const currPath = this.connectionHighlights
+        .get(connection)
+        ?.getAttribute('d');
+      const currOffset = this.highlightOffsets.get(connection);
+      if (
+        connectionPath === currPath &&
+        Coordinate.equals(offset, currOffset)
+      ) {
+        return;
+      } else {
```
maribethb, line 285: Can remove this `else`. Also could make this a little clearer/more self documenting by extracting the `if` clause into a `highlightsMatch` function so

```
if (this.connectionHighlights.has(connection)) {
  if (this.highlightsMatch(connectionPath, offset, existingHighlight) return;
  this.removeConnectionHighlight(connection);
}

// ... rest of addConnectionHighlight ...


highlightsMatch(connectionPath, offset, existingHighlight) {
  const existingPath = existingHighlight.get(connection)?.getAttribute('d');
  const existingOffset = this.highlightOffsets.get(connection);
  return connectionPath === existingPath && Coordinate.equals(offset, existingOffset);
}
```

just a suggestion though :) feel free to disagree

maribethb (approved):

core/renderers/common/path_object.ts
BeksOmega: Much better! Done.

core/renderers/common/path_object.ts
BeksOmega: This is unreleased!

PR #7801, opened 2024-01-22
Add v11 breaking changes to the rc/v11 branch.

BeksOmega:
All of these changes have been individually reviewed.


cpcallen (approved):
I tweaked the title of this PR:

I'm presuming you're planning to rebase-and-merge rather than squash.  (A merge commit could also be used if it was useful to preserve the history of the `v11-breaks` branch, but here I think we're interested only in the individual commits.)  But it turns out that closing a PR via rebase-and-merge will create a new set of commits with slightly different hashes than the original, so even though this is _effectively_ like fast-forwarding the `rc/v11.0.0` branch to where `v11-breaks` is that is not actually what happens.

If you wanted to actually do the fast-forward then the way to do it would be to update your branches locally, do a `git merge --ff-only` to do the fast-forward, and then push the updated `rc/v11.0.0` branch.  (You might have to turn off branch protection temporarily.)

That might be worth doing to preserve the commit hashes of the commits on that branch if they are still the ones that were created by squashing the corresponding PRs, but if the branch has already been rebased since (some/most) of those PRs got squashed then there's not much to be lost by rebasing them again.

BeksOmega:
> If you wanted to actually do the fast-forward then the way to do it would be to update your branches locally, do a `git merge --ff-only` to do the fast-forward, and then push the updated `rc/v11.0.0` branch. (You might have to turn off branch protection temporarily.)
> 
> That might be worth doing to preserve the commit hashes of the commits on that branch if they are still the ones that were created by squashing the corresponding PRs, but if the branch has already been rebased since (some/most) of those PRs got squashed then there's not much to be lost by rebasing them again.

I'm going to rebase and merge. I don't think there's any /need/ to preserve the commit hashes, and I don't want to mess with branch protections. But good to know for the future that it is different than fast forwarding!

PR #7806, opened 2024-01-25
fix: export save and load procedure

BeksOmega:
## The basics

### Resolves

Fixes N/A

We need to export the save and load functions because the save and load functions deal with parameters, while the functions on the procedure classes do not.

The events in samples need to be able to serialize and deserialize procedures properly.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/serialization/procedures.ts
+++ b/core/serialization/procedures.ts
@@ -70,7 +70,7 @@ interface ParameterModelConstructor<ParameterModel extends IParameterModel> {
 /**
  * Serializes the given IProcedureModel to JSON.
  */
-function saveProcedure(proc: IProcedureModel): State {
+export function saveProcedure(proc: IProcedureModel): State {
   const state: State = proc.saveState();
   if (!proc.getParameters().length) return state;
   state.parameters = proc.getParameters().map((param) => param.saveState());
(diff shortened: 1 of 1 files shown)


maribethb (approved):
lgtm but why is this going directly into the rc/ branch and not develop (since it's not a breaking change)? My understanding is the rc branches are only used to cut releases so they should only contain cherry-picks if they aren't just a straight copy of develop.

BeksOmega:
> lgtm but why is this going directly into the rc/ branch and not develop (since it's not a breaking change)? My understanding is the rc branches are only used to cut releases so they should only contain cherry-picks if they aren't just a straight copy of develop.

Dependent on https://github.com/google/blockly/pull/7740 which only exists in the RC branch, because it is breaking.

BeksOmega:
> lgtm but why is this going directly into the rc/ branch and not develop (since it's not a breaking change)? My understanding is the rc branches are only used to cut releases so they should only contain cherry-picks if they aren't just a straight copy of develop.

Oh skimmed over the second part of your message. We're handling all the breaking changes in the rc branch this quarter so that we can do a minor release from develop mid quarter.

PR #7818, opened 2024-01-30
chore: fix docs for shouldHighlightConnection to say it might not be respected

BeksOmega:
## The basics

### Resolves

Fixes N/A

Updates docs to reflect that `shouldHighlightConnection` might not be respected by all connection previewers.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/renderers/common/renderer.ts
+++ b/core/renderers/common/renderer.ts
@@ -188,7 +188,10 @@ export class Renderer implements IRegistrable {
   }
 
   /**
-   * Determine whether or not to highlight a connection.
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
Are there any ways for devs to know which connection previewers care about this? If yes, add that information.

BeksOmega:
> Are there any ways for devs to know which connection previewers care about this? If yes, add that information.

Not besides checking the docs for that specific previewer :/ The one provided by default does, but we can't require external folks to respect it.

PR #7828, opened 2024-02-02
feat: make grid options togglable

### LINKED ISSUE #7827: Make grid length, spacing, and snapping toggleable
### Check for duplicates

### Problem

App Inventor wants to be able to toggle the visibility of the grid and whether snapping is enabled or not.

### Request

Add the following methods to the grid:
- setLength
- setSpacing
- setSnapToGrid

Also make `WorkspaceSvg.prototype.getGrid` public.

### Alternatives considered

Could force App Inventor to reinject the workspace when these things are toggled, but that seems like overkill.

### Additional context

This will mean that the state of the workspace could get out of sync with its inject options, but I don't think that's worth fixing.

BeksOmega:
## The basics

### Resolves

Fixes #7827 

Makes it so that grid options are toggleable so App Inventor can toggle them from their context menu items.

### Test Coverage

Manually tested that setting all of the options works as expected.

### Documentation

Covered by reference documentation.

### Additional Information

N/A

### DIFF
--- a/core/grid.ts
+++ b/core/grid.ts
@@ -20,11 +20,12 @@ import {GridOptions} from './options.js';
  * Class for a workspace's grid.
  */
 export class Grid {
-  private readonly spacing: number;
-  private readonly length: number;
+  private spacing: number;
+  private length: number;
+  private scale: number = 1;
   private readonly line1: SVGElement;
   private readonly line2: SVGElement;
-  private readonly snapToGrid: boolean;
+  private snapToGrid: boolean;
 
   /**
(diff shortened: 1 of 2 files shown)


BeksOmega:
> From the issue:
> 
> > This will mean that the state of the workspace could get out of sync with its inject options, but I don't think that's worth fixing.
> 
> Is this mooting the idea of modifying the properties of the `GridOptions` object? I'm not sure that would even be _desirable_ let alone "worth fixing".

Yesh not doing that.

PR #7834, opened 2024-02-06
fix: disposing of connection previewer

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

We were never actually calling dispose on the connection previewer. This worked fine because the default previewer's `dispose` method just calls `hidePreview`. But other previewers may need to do other work.

### Test Coverage

Manually spot-tested, everything seems good.

### Documentation

Discovered this when writing docs :P

### Additional Information

N/A

### DIFF
--- a/core/block_dragger.ts
+++ b/core/block_dragger.ts
@@ -382,6 +382,7 @@ export class BlockDragger implements IBlockDragger {
 
     blockAnimation.disconnectUiStop();
     this.connectionPreviewer.hidePreview();
+    this.connectionPreviewer.dispose();
 
     const preventMove =
       !!this.dragTarget_ &&


core/block_dragger.ts
```diff
@@ -382,6 +382,7 @@ export class BlockDragger implements IBlockDragger {
 
     blockAnimation.disconnectUiStop();
     this.connectionPreviewer.hidePreview();
+    this.connectionPreviewer.dispose();
```
cpcallen, line 385: If `.dispose` calls `.hidePreview`, can the latter call be omitted here?

cpcallen (approved):

core/block_dragger.ts
BeksOmega: It could be, but that would mean we're requiring a side effect (`.dispose` must hide any previews). Calling both separately makes the API cleaner and easier to document.

Issue #7864, opened 2024-02-16
Firing delete events before setting `disposing` to `true` causing issues errors in the dynamic connections


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

Currently we set `disposing` after we fire a delete event for the block (I'm not sure why this is). This causes issues for the dynamic connections plugin because the delete event triggers serialization, which triggers the block to be reconstructed due to historical reasons (the plugin was originally serialized with the wrong XML format, and we need to fix that). However the reconstruction fails because the renderer constants have already been disposed before it is triggered. So anything that references `getConstants()!` assuming it should exist during rendering triggers an error.

### Reproduction steps

1. Go to the [dynamic connections blocks plugin demo page](https://google.github.io/blockly-samples/plugins/block-dynamic-connection/test/index.html).
2. In the console run `Blockly.getMainWorkspace().dispose()`.
3. Observe the error

### Stack trace

```shell
blockly_compressed.js:708 Uncaught TypeError: Cannot read properties of null (reading 'FIELD_TEXT_BASELINE_CENTER')
    at FieldLabel$$module$build$src$core$field_label.createTextElement_ (blockly_compressed.js:708:211)
    at FieldLabel$$module$build$src$core$field_label.initView (blockly_compressed.js:897:1147)
    at FieldLabel$$module$build$src$core$field_label.init (blockly_compressed.js:706:494)
    at $.ValueInput$$module$build$src$core$inputs$value_input.insertFieldAt (blockly_compressed.js:900:189)
    at $.ValueInput$$module$build$src$core$inputs$value_input.appendField (blockly_compressed.js:899:555)
    at BlockSvg$$module$build$src$core$block_svg.addFirstInput (dynamic_text_join.ts:282:42)
    at BlockSvg$$module$build$src$core$block_svg.addItemInputs (dynamic_text_join.ts:264:24)
    at BlockSvg$$module$build$src$core$block_svg.finalizeConnections (dynamic_text_join.ts:219:10)
    at BlockSvg$$module$build$src$core$block_svg.mutationToDom (dynamic_text_join.ts:52:10)
    at blockToDom$$module$build$src$core$xml (blockly_compressed.js:167:332)
createTextElement_ @ blockly_compressed.js:708
initView @ blockly_compressed.js:897
init @ blockly_compressed.js:706
insertFieldAt @ blockly_compressed.js:900
appendField @ blockly_compressed.js:899
[...]

PR #7891, opened 2024-02-26
fix!: have disposing be true from start of dispose

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7864

### Proposed Changes

Makes it so `disposing` is set at the beginning of the `dispose` and `disposeInternal` methods.

### Reason for Changes

This way we can check `isDeadOrDying` in the dynamic connection plugin to get around issues with the dynamic connection plugin (see the linked issue for more detail).

### Test Coverage

Updated incorrect tests.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -776,7 +776,7 @@ export class BlockSvg
    * @param animate If true, show a disposal animation and sound.
    */
   override dispose(healStack?: boolean, animate?: boolean) {
-    if (this.isDeadOrDying()) return;
+    this.disposing = true;
 
     Tooltip.dispose();
     ContextMenu.hide();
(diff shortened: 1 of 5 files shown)


core/block_svg.ts
```diff
@@ -795,7 +795,7 @@ export class BlockSvg
    * E.g. does trigger UI effects, remove nodes, etc.
    */
   override disposeInternal() {
-    if (this.isDeadOrDying()) return;
+    this.disposing = true;
```
cpcallen, line 798: I am pretty sure that this one is redundant and can be removed, since you do a super call immediately and `Block.prototype.disposeInternal` also sets `this.disposing = true`.

cpcallen (approved):

core/block_svg.ts
BeksOmega: I agree, but to future-proof this (in case we ever add things before calling super) and make it extremely clear, I think it's reasonable to leave it in.

If this were more expensive I would remove it. But since it's cheap I think it's reasonable to leave.

PR #7914, opened 2024-03-07
feat: add comment view (for workspace comments, and block comments for partners)

### LINKED ISSUE #7905: Implement basic CommentView class

BeksOmega:
## The basics

### Resolves

Fixes #7905 

### Proposed Changes

Adds a comment view class which will later be used to 

### Test Coverage

Manually tested:
  * Typing into the block.
  * Setting the text programmatically.
  * Collapsing the block.
  * Collapsing the block programmatically.
  * Setting the block to be non-editable programatically (no way to do this manually).
  * Resizing the block.
  * Resizing the block programatically.
  * Moving the block programatically (no way to do this manually because that will be handled by whatever class is using the view).

### Documentation

There will need to be lots of docs later.

### Additional Information

Hopefully this is broken up in a way that's easy to review commit-wise 🤞 I do split some methods out into sub-procedures in later commits, but I think that's the only thing that majorly changes between earlier commits and later ones.


core/comments/comment_view.ts
```diff
@@ -47,103 +47,148 @@ export class CommentView implements IRenderedElement {
...
+    } = this.createTopBar(this.svgRoot, workspace));
+
+    ({foreignObject: this.foreignObject, textArea: this.textArea} =
+      this.createTextArea(this.svgRoot));
+
+    this.resizeHandle = this.createResizeHandle(this.svgRoot, workspace);
+
+    // TODO: Remove this comment before merging.
+    // I think we want comments to exist on the same layer as blocks.
+    workspace.getLayerManager()?.append(this, layers.BLOCK);
```
rachel-fenichel, line 110: The disadvantage of being on the same layer as blocks is that a block can go over the comment and look like it's inside it. That's a bigger issue with mutators bubbles, since they always hold blocks. I'm neutral on which one we should do for comments.

core/comments/comment_view.ts
```diff
@@ -18,28 +18,73 @@ import * as touch from '../touch.js';
...
+  /** The current size of the comment. */
   private size: Size = new Size(120, 100);
+
+  /** Whether the comment is collapsed or not. */
   private collapsed: boolean = false;
+
+  /** Whether the comment is editable or not. */
   private editable: boolean = true;
+
+  /** The current location of the comment. */
```
rachel-fenichel, line 57: Add note on coordinate system.

core/comments/comment_view.ts
```diff
@@ -0,0 +1,686 @@
...
+import {IRenderedElement} from '../interfaces/i_rendered_element.js';
+import {WorkspaceSvg} from '../workspace_svg.js';
+import * as dom from '../utils/dom.js';
+import {Svg} from '../utils/svg.js';
+import * as layers from '../layers.js';
+import * as css from '../css.js';
+import {Coordinate, Size, browserEvents} from '../utils.js';
+import * as touch from '../touch.js';
+
+// TODO: Before merging file an issue for calculating a better min size. We
```
rachel-fenichel, line 16: Check for TODOs and associate with issues before merging.

core/comments/comment_view.ts
```diff
@@ -174,6 +177,18 @@ export class CommentView implements IRenderedElement {
...
+    this.onSizeChange(oldSize, this.size);
+  }
+
+  private onSizeChange(oldSize: Size, newSize: Size) {
+    for (const listener of this.sizeChangeListeners) {
+      listener(oldSize, newSize);
+    }
+  }
+
+  addSizeChangeListener(listener: (oldSize: Size, newSize: Size) => void) {
```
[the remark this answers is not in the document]
rachel-fenichel: Same question for all listeners.

core/comments/comment_view.ts
BeksOmega: There is also a weird thing where when we deserialize, the comments will be deserialized after blocks, so they'll always end up on top after load even if they were behind blocks before.

core/comments/comment_view.ts
```diff
@@ -47,103 +47,148 @@ export class CommentView implements IRenderedElement {
       'class': 'blocklyComment blocklyEditable',
     });
 
-    this.topBar = dom.createSvgElement(
+    ({
+      topBar: this.topBar,
```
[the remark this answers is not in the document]
BeksOmega: Yesh! It is a little bit silly because if you try to reference the property in the creation method typescript doesn't realize you're doing a bad. (E.g. reference `this.textArea` in `createTextArea` throws a runtime error) But that's not specific to what I'm doing here, typescript just isn't smart enough to track references from the constructor.

core/comments/comment_view.ts
```diff
@@ -370,6 +377,27 @@ export class CommentView implements IRenderedElement {
   private truncateText(text: string): string {
     return text.length >= 12 ? `${text.substring(0, 9)}...` : text;
   }
+
+  private onDeleteDown(e: PointerEvent) {
```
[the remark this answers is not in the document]
BeksOmega: Can do, but I'll probably just give it the same format as blocks (`isDeadOrDying`) unless you have a different way you'd like it to look.

core/comments/comment_view.ts
```diff
@@ -0,0 +1,686 @@
...
+      Svg.RECT,
+      {
+        'class': 'blocklyCommentTopbar',
+        'x': 0,
+        'y': 0,
+      },
+      svgRoot,
+    );
+    // TODO: Before merging, does this mean to override an individual image,
+    // folks need to replace the whole media folder?
```
BeksOmega, line 140: Do you have thoughts on this @rachel-fenichel ?

core/comments/comment_view.ts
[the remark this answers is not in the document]
rachel-fenichel: That's fine. I just don't want to get to where we're inspecting whether a random SVG node is null as our way of checking for disposal >.<

Although that's no longer possible because of how you did initialization, so maybe it's fine to leave out `isDeadOrDying` and assume that we'll add it as soon as we have problems with events and disposal. Your call.

core/comments/comment_view.ts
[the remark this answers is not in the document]
BeksOmega: - [ ] Add tests

core/comments/comment_view.ts
rachel-fenichel: I don't know. Is replacing those images an immediate need, or a theoretical one?

tests/mocha/workspace_comment_test.js
```diff
@@ -9,259 +9,191 @@ import {
   sharedTestTeardown,
 } from './test_helpers/setup_teardown.js';
 
-suite('Workspace comment', function () {
+suite.only('Workspace comment', function () {
```
rachel-fenichel, line 12: Unexpected suite.only. Please fix.

rachel-fenichel (approved):

core/comments/comment_view.ts
BeksOmega: It's a need for spork blocks. For MakeCode we can just include their image in core.

BeksOmega: @maribethb  as well.

maribethb: Yeah, they do, but tbh they probably should anyway because the default one links to one hosted on app engine. I don't see this as being a concern, it's how the rest of media assets work. If we wanted to refactor that in the future that's a different discussion and shouldn't be folded into this work imo

BeksOmega: Ok I'll dig into what MakeCode is currently doing and chat with them if it seems like it might be a problem. Since we've been holding v11 until we can have stuff ready for MakeCode, I want to make sure it actually is ready, and they're not going to have to wait for a v12.

maribethb: What kind of problems are you thinking about? The media directory is already configurable via options, and they can copy the assets they want to keep the same into their own hosting solution (which is what we currently recommend, though not that loudly). I feel like I'm missing something here if you're worried about something needing a v12.

BeksOmega: You're right I don't think this would need a v12. If anything I would just make it possible to configure the assets individually (which would be a feat not breaking). Based on some investigation it looks like they're already hosting everything themselves so it shouldn't be an issue. I just want to make sure we have everything pinned down so we can actually say MakeCode has been migrated.

PR #7915, opened 2024-03-08
fix: comment min size

BeksOmega:
## The basics

### Resolves

Fixes N/A 

Makes it so we calculate the minimum size of the comment based on the size of the preview text. This way we'll never get preview text hanging off the edge of the comment.

Alternatively, I could have tried to truncate the preview text to fit within the available area, but this is super inefficient because it requires continual trial and error and layout recalculation. So instead of fitting the text to the area, I fit the area to the text.

Ideally the browser would just truncate the preview and add ellipses for us (then we could fit the text to the area), but SVG doesn't support that! So instead we have to do this complicated madness :P

### Test Coverage

Tested that the minimum size is properly set with different text width, and with or without the delete icon.

### Documentation

N/A

### Additional Information

This doesn't work exactly like MakeCode's setup. For them, they always set the width to a specific width when collapsing (the length of 12 `m` characters, plus the ellipse) which causes a bug where it overlaps their icons. Instead when collapsing, we keep the width of the comment the same, just hide the text area.

They are able to implement their behavior using the `onCollapse` listener to set the width explicitly. So this  isn't a blocker even though it's a difference in behavior.

~But writing this did make me realize that I might need to change the calculation for positioning the preview text slightly to do what they want, so I'll look into that in a follow up.~ Tested and the current setup should be fine.

### DIFF
--- a/core/comments/comment_view.ts
+++ b/core/comments/comment_view.ts
@@ -13,7 +13,6 @@ import * as css from '../css.js';
 import {Coordinate, Size, browserEvents} from '../utils.js';
 import * as touch from '../touch.js';
 
-const MIN_SIZE = new Size(100, 60);
 export class CommentView implements IRenderedElement {
   /** The root group element of the comment view. */
   private svgRoot: SVGGElement;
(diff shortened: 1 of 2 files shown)


core/utils/size.ts
```diff
@@ -43,4 +43,20 @@ export class Size {
     }
     return a.width === b.width && a.height === b.height;
   }
+
+  /**
+   * Returns a new size with the maximum width and heigh values out of both
```
rachel-fenichel, line 48: heigh -> height

core/comments/comment_view.ts
```diff
@@ -578,13 +622,19 @@ export class CommentView implements IRenderedElement {
...
+    this.updateTextPreview(this.text);
+    // Update size in case our minimum size increased.
+    this.setSize(this.size);
     // Loop through listeners backwards in case they remove themselves.
     for (let i = this.textChangeListeners.length - 1; i >= 0; i--) {
       this.textChangeListeners[i](oldText, this.text);
     }
   }
 
+  private updateTextPreview(text: string) {
```
rachel-fenichel, line 634: Add doc comment

core/comments/comment_view.ts
```diff
@@ -292,6 +291,51 @@ export class CommentView implements IRenderedElement {
...
+      width += 4; // Arbitrary margin after text.
+    }
+
+    // Arbitrary additional height.
+    const height = topBarSize.height + 20;
+
+    return new Size(width, height);
+  }
+
+  private calcDeleteMargin(topBarSize: Size, deleteSize: Size) {
```
rachel-fenichel, line 331: Add doc comment.

rachel-fenichel (approved):
A few nits on comments, then LGTM.

PR #7916, opened 2024-03-08
feat: add headless workspace comment class

### LINKED ISSUE #7906: Implement WorkspaceComment class

BeksOmega:
## The basics

### Resolves

Fixes #7906 

Adds a headless workspace comment class (just holds data, doesn't refer to any view elements) to support headless mode. 

### Test Coverage

No testing, because this is just a bunch of data storage methods. These methods will be covered by later serialization tests.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/comments/workspace_comment.ts
+++ b/core/comments/workspace_comment.ts
@@ -12,6 +12,15 @@ import * as idGenerator from '../utils/idgenerator.js';
 export class WorkspaceComment {
   public readonly id: string;
 
+  private text = '';
+  private size = new Size(120, 100);
+  private collapsed = false;
+  private editable = true;
+  private movable = true;
+  private deletable = true;
+  private location = new Coordinate(0, 0);
+  private disposed = false;
+
   constructor(
     protected readonly workspace: Workspace,
     id?: string,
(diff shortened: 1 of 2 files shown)


BeksOmega:
@cpcallen Since Maribeth seems to be maybe OOO would you be interested in taking this?

core/comments/workspace_comment.ts
```diff
@@ -0,0 +1,172 @@
...
+    return this.size;
+  }
+
+  /** Sets whether the comment is collapsed or not. */
+  setCollapsed(collapsed: boolean) {
+    this.collapsed = collapsed;
+  }
+
+  /** Returns whether the comment is collapsed or not. */
+  isCollapssed(): boolean {
```
maribethb, line 86: typo

core/comments/workspace_comment.ts
```diff
@@ -0,0 +1,172 @@
...
+  setEditable(editable: boolean) {
+    this.editable = editable;
+  }
+
+  /**
+   * Returns whether the comment is editable or not, respecting whether the
+   * workspace is read-only.
+   */
+  isEditable(): boolean {
+    return this.isOwnEditable() && !this.workspace.options.readOnly;
```
maribethb, line 100: swapping these two is slightly better in case the workspace is readonly and the user decides to compute the traveling salesman problem in the `ownEditable` function, but i'll acknowledge this is minute :P

the equivalent in `Block` also checks if the block has been disposed. do we need to do that here?

core/comments/workspace_comment.ts
```diff
@@ -0,0 +1,172 @@
...
+  getRelativetoSurfaceXY(): Coordinate {
+    return this.location;
+  }
+
+  /** Disposes of this comment. */
+  dispose() {
+    this.disposed = true;
+  }
+
+  /** Returns whether teh comment has been disposed or not. */
```
maribethb, line 168: typo `teh`

core/comments/workspace_comment.ts
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * @param workspace The workspace to construct the comment in.
+   * @param id An optional ID to give to the comment. If not provided, one will
+   *     be generated.
+   */
+  constructor(
+    protected readonly workspace: Workspace,
+    id?: string,
+  ) {
+    // TODO: Before merging, file issue to update getCommentById.
```
maribethb, line 51: you have some todos here fyi

maribethb (approved):

core/comments/workspace_comment.ts
BeksOmega: This is one thing that's bad about auto complete. I type it wrong once and then I just tab-complete wrong for the rest of time :P

core/comments/workspace_comment.ts
BeksOmega: I don't think we need to do that. Looks like in `Block` it was originally `(this.workspace && !this.workspace.options.readonly)` which turned into `!this.disposed && !this.workspace.options.readonly` which turned into `!this.isDeadOrDying() && !this.workspace.options.readonly`.

So I think it was originally just a type safety thing, and as we migrated code the logic got slightly confused :P

Theoretically something that is being disposed probably isn't editable? But it probably doesn't have a size or text either. I think it's fine to ignore this.

PR #7924, opened 2024-03-11
fix: Allow duplicate registration of values

### LINKED ISSUE #7804: Handle registering the same field multiple times
### Check for duplicates

### Problem

When custom blocks register themselves they also should register the custom fields that they need. This means that a file that defines multiple blocks may register a field multiple times if all of those blocks are installed.

### Request

Check whether the new field being registered under a specific name is the same as the field that was previously registered under that name, and act as a no-op if that's the case.

### Alternatives considered

_No response_

### Additional context

Part of https://github.com/google/blockly/issues/7226

NeilFraser:
Also removes unused partially-implemented case-sensitivity from registry.

Fixes #7804

### DIFF
--- a/core/registry.ts
+++ b/core/registry.ts
@@ -24,7 +24,6 @@ import type {ToolboxItem} from './toolbox/toolbox_item.js';
 import type {IPaster} from './interfaces/i_paster.js';
 import type {ICopyData, ICopyable} from './interfaces/i_copyable.js';
 import type {IConnectionPreviewer} from './interfaces/i_connection_previewer.js';
-import * as deprecation from './utils/deprecation.js';
 
 /**
  * A map of maps. With the keys being the type and name of the class we are
@@ -37,6 +36,15 @@ const typeMap: {
   };
 } = Object.create(null);
 export const TEST_ONLY = {typeMap};
+
+/**
+ * A map of maps.  Record the case-insensitive version of each name.
+ * e.g. {'field': {'field_angle': 'fIeLd_AnGle'}}
+ * This enables an error to be thrown if a name of conflicting case is used.
+ */
+const noCaseNameMap: {
+  [key: string]: {[key: string]: string}
+} = Object.create(null);
 
 /**
  * The string used to register the default class for a type of plugin.
@@ -123,7 +131,7 @@ export function register<T>(
     | AnyDuringMigration,
   opt_allowOverrides?: boolean,
 ): void {
-  type = normalizeName(`${type}`);
+  type = normalizeType(type);
   name = normalizeName(name);
   if (!registryItem) {
     throw Error('Can not register a null value');
@@ -132,10 +140,12 @@ export function register<T>(
   // If the type registry has not been created, create it.
   if (!typeRegistry) {
     typeRegistry = typeMap[type] = Object.create(null);
+    noCaseNameMap[type] = Object.create(null);
   }
 
   // Validate that the given class has all the required properties.
-  validate(type, registryItem);
+  validateItem(type, registryItem);
+  validateCase(type, name);
 
   // Don't throw an error if opt_allowOverrides is true.
   // Don't throw an error if the registryItem is not changing.
@@ -147,6 +157,7 @@ export function register<T>(
     throw Error(`Name "${name}" with type "${type}" already registered`);
   }
   typeRegistry[name] = registryItem;
+  noCaseNameMap[type][name.toLowerCase()] = name;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


core/registry.ts
```diff
@@ -265,15 +228,30 @@ function getItem<T>(
...
 }
 
+/**
+ * Normalize a string to be used as a key.  By default, lowercase and trimmed.
+ *
+ * @param name The key's name. (Ex. 'FiELd_AnGlE ')
+ * @returns A normalized string. (Ex. 'field_angle')
+ */
+function normalizeName(name: string): string {
+  // Do we need this type checking?  We are using TypeScript after all.
```
maribethb, line 247: While we're using TypeScript, a lot of developers aren't, so I've found that it reduces confusion when we check anyway and then issue good warnings. This only applies to developer-facing APIs, not internal-only methods. There are a few places where I wouldn't mind adding more (e.g. an FAQ in the forum is related to leaving off the `workspace` parameter when serializing so if we checked for that case and issued an error we could probably save ourselves some support burden)

That being said, this particular message isn't useful anymore. The original message told you it either had to be a string or had to be a `Blockly.registry.Type` class. So I would probably do that check in `register` that the name is valid, and you can eliminate it here.

core/registry.ts
```diff
@@ -316,18 +294,20 @@ export function getObject<T>(
...
- *     caseless keys). False by default.
+ * @param opt_cased Deprecated.
  * @param opt_throwIfMissing Whether or not to throw an error if we are unable
  *     to find the object. False by default.
  * @returns A map of objects with the given type, or null if none exists.
  */
 export function getAllItems<T>(
   type: string | Type<T>,
-  opt_cased?: boolean,
+  opt_cased?: any,
```
maribethb, line 304: Unfortunately after looking into it I'm no longer sure this is safe to remove, at least not without doing some work to make the serializer case-insensitive.

If someone registered a serializer with an uppercase name, then their existing serialized workspace data would have uppercase characters in it. After this change, if we tried to deserialize that data, the names wouldn't match and the data wouldn't load correctly. @BeksOmega can you confirm that? Would this be safe if we changed serialization to also lowercase the state keys when loading?

BeksOmega: Yeah I believe if someone has registered a serializer with capitals (which is would be reasonable if they were using `camelCase`) their data will no longer load. Should be relatively easy to add a test for that.

Given that, I have a small opinion that this is not worth changing.

rachel-fenichel:
I think this PR should be against `rc/v11.0.0` since it is necessary for the v11 release.

core/registry.ts
```diff
@@ -184,16 +168,32 @@ export function register<T>(
...
   }
 }
 
+/**
+ * Checks that the name hasn't been registered using a different case.
+ *
+ * @param type The type of the plugin. (e.g. Field, Renderer)
+ * @param name The plugin's name. (Ex. field_angle, geras)
+ */
+function validateCase(type: string, name: string) {
```
maribethb, line 190: I think this is now a breaking change in the other direction, we previously allowed you to ignore case (except for serializers, where the name you pass in is actually in the serialized output) so throwing an error now will break anyone who was relying on that. 

There might still be some cleanup that's possible here but I don't think we should change the behavior in either direction, in other words, we live with the somewhat weird state that most registries are case-insensitive except for the serializer. And the cleanup doesn't need to block the allow duplicate registration of values since that's required for v11 so there's some urgency there, unlike the cleanup.

NeilFraser: The current PR will break two cases:
1. Where someone is using our partial case-sensitivity to store different values for keys 'foo' and 'Foo'.
2. Where someone is using our partial case-insensitivity to get away with inconsistently switching between 'foo' and 'Foo'.
I'd argue that both of these behaviours are bugs.

It would be one thing if we could just leave this mess alone, but we need to build on top of it for V11.  This problem is going to contiune to compound, and we'll have a registry that behaves in random ways and is impossible to document.  I think if we don't straighten the registry out now, it's just going to get worse and worse.

maribethb: I don't think it's fair to call those behaviors bugs because we explicitly built support for them (i.e. the parameter you can pass to allow for case insensitivity). In hindsight, we may have made a different decision around this if we went back in time. But now, if we want to change it, we need to do a proper cleanup which would mean weighing the value of this breaking change against the burden of communicating it to users and forcing them to do work to change their code if they're affected by it. Part of that proper cleanup would be doing a real breaking change, communicating it effectively, and not having it be mixed with the feature of allowing duplicate registration.

I hear you that we're building on top of mess, but the actual change we need for v11 doesn't have anything to do with case sensitivity and is really a very small change buried in the much larger case sensitivity refactoring part of this PR, so I don't think that's a compelling enough reason to mix the two issues.

If you want to file an issue for the case insensitivity cleanup, we can triage it separately during our normal bug triage process. I would prefer you do just the duplicate registration of values in this PR, and wait to see how prioritization works out for the rest of the cleanup vs everything else we already have planned for the v11 release

NeilFraser:
I've written #7982 and #7983 which are the other two mutually-exclusive conflicting solutions.  You can commit whichever of these three you feel is best.

maribethb:
Fixed the original feature request in #7988 and opened https://github.com/google/blockly/issues/7987 for discussion of the general case-sensitivity issue.

PR #7893, opened 2024-02-26
chore: merge develop into v11

BeksOmega:
## The basics

- [X] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes N/A

### Proposed Changes + reasons

Fixes problems with CI on the v11 branch.

PR #7929, opened 2024-03-12
chore: merge develop into v11.0.0

BeksOmega:
Re-merges develop into v11, but this time we're actually going to use a merge commit so we can track where v11 is relative to develop. (as opposed to https://github.com/google/blockly/pull/7893 which squashed)


NeilFraser (approved):
Looks like all stuff that's already been reviewed, so I'm not looking too closely at it.

PR #7931, opened 2024-03-12
fix: comment have XML save and load new workspace comments classes

BeksOmega:
## The basics

### Resolves

Fixes N/A

Updates the XML system to save and load new workspace comment classes for a ✨ seamless transition ✨ 

### Test Coverage

Added round-trip tests for comments. This also covers the logic for all the headless setters and getters (as promised in a previous PR)

### Documentation

N/A

### Additional Information

Sorry about the derpy format for the serialization tests. But rewriting them to be mocha style isn't worth it.

When writing this I realized that JSON isn't handling RTL correctly. I used the same (bad) implementation here, then I'm going to fix it in my next PR.

### DIFF
--- a/core/xml.ts
+++ b/core/xml.ts
@@ -71,7 +71,7 @@ function saveWorkspaceComment(
   if (comment.isCollapsed()) elem.setAttribute('collapsed', 'true');
   if (!comment.isOwnEditable()) elem.setAttribute('editable', 'false');
   if (!comment.isOwnMovable()) elem.setAttribute('movable', 'false');
-  if (!comment.isOwnDeletable()) elem.setAttribute('deltable', 'false');
+  if (!comment.isOwnDeletable()) elem.setAttribute('deletable', 'false');
 
   return elem;
 }
--- a/core/xml.ts
+++ b/core/xml.ts
@@ -19,22 +19,21 @@ import * as utilsXml from './utils/xml.js';
 import type {VariableModel} from './variable_model.js';
 import * as Variables from './variables.js';
 import type {Workspace} from './workspace.js';
-import {WorkspaceComment} from './workspace_comment.js';
-import {WorkspaceCommentSvg} from './workspace_comment_svg.js';
-import type {WorkspaceSvg} from './workspace_svg.js';
+import {WorkspaceSvg} from './workspace_svg.js';
 import * as renderManagement from './render_management.js';
+import {WorkspaceComment} from './comments/workspace_comment.js';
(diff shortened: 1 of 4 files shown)


core/xml.ts
```diff
@@ -19,42 +19,63 @@ import * as utilsXml from './utils/xml.js';
...
 
+/** Serializes the given workspace comment to XML. */
+function saveWorkspaceComment(
+  comment: WorkspaceComment,
+  saveId = true,
+): Element {
+  const elem = utilsXml.createElement('comment');
+  if (saveId) elem.setAttribute('id', comment.id);
+
+  elem.setAttribute('x', `${comment.getRelativeToSurfaceXY().x}`);
```
maribethb, line 65: For `BlockSvg` this operation might be expensive because you have to look at a list of elements. Not sure how expensive it might be for ws comments in the current implementation, but might it be worth saving the value rather than calculate it twice in a row? If the values are cached in the class then it doesn't matter, but I haven't pulled down the branch to check

tests/mocha/serializer_test.js
```diff
@@ -1891,6 +2060,7 @@ const runSerializerTestSuite = (serializer, deserializer, testSuite) => {
         workspaces.load(deserializer(save), this.workspace);
       }
       const newXml = Blockly.Xml.workspaceToDom(this.workspace);
+      // console.log(Blockly.Xml.domToText(newXml));
```
maribethb, line 2063: Remove commented out code, probably

core/xml.ts
```diff
@@ -478,6 +491,34 @@ export function domToWorkspace(xml: Element, workspace: Workspace): string[] {
   return newBlockIds;
 }
 
+/** Deserializes the given comment state into the given workspace. */
+function loadWorkspaceComment(
```
maribethb, line 495: I am curious why not have this implemented as a function on the class itself, like it used to be? It seems like that would make it easier on people who are implementing custom subclasses, and it would make more sense organizationally IMO. And we do have a pattern of a `fromXml` method on fields, events, etc.

maribethb (approved):

core/xml.ts
BeksOmega: Yeah this just returns the saved location, no expense.

core/xml.ts
BeksOmega: No one is allowed to implement custom subclasses atm. We discussed that and decided YAGNI.

maribethb: Fair enough :)

maribethb (approved):

PR #7936, opened 2024-03-14
chore: Use .includes and .startsWith, not .indexOf

NeilFraser:
Easier to read than the diverse collection of `=== 0` and `!== -1` and `> -1` tests.


cpcallen (approved):
A useful improvment.  My eyes have gone crosseyed trying to check that all the conversions are correct, but I think they just might be.

NeilFraser:
The rule of thumb I use is that there should be a total of one `!` in either the new or the old.  And `> -1` counts as a `!`.

PR #7937, opened 2024-03-14
feat: allow overriding comment icons

### LINKED ISSUE #7910: Define CommentIcon interface

### LINKED ISSUE #7911: Add overriding comment icons

BeksOmega:
## The basics

### Resolves

Fixes #7910
Fixes #7911 

### Proposed Changes

Allows comment icons to be overriden, and constructed from `Block.prototype.setCommentText`.

### Reason for Changes

This will allow external partners to create their own comment icons (e.g. that use the new `CommentView`, while still using the built-in functionality like `setCommentText`, context menu options, etc.

### Test Coverage

Added tests that `setCommentText` properly constructs classes from the registry.

### Documentation

Yesh, we'll need to document the fact that you can do this somewhere.

### Additional Information

N/A

### DIFF
--- a/core/interfaces/i_comment_icon.ts
+++ b/core/interfaces/i_comment_icon.ts
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {IconType} from '../icons.js';
+import {IIcon, isIcon} from './i_icon.js';
+import {Size} from '../utils/size.js';
+import {IHasBubble} from './i_has_bubble.js';
+
+export interface ICommentIcon extends IIcon, IHasBubble {
+  setText(text: string): void;
+
+  getText(): string;
+
+  setBubbleSize(size: Size): void;
+
+  getBubbleSize(): Size;
+}
+
+/** Checks whether the given object is an ICommentIcon. */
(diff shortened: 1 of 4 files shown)


maribethb (approved):

core/interfaces/i_comment_icon.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  getBubbleSize(): Size;
+}
+
+/** Checks whether the given object is an ICommentIcon. */
+export function isCommentIcon(obj: Object): obj is ICommentIcon {
+  return (
+    isIcon(obj) &&
+    (obj as any)['setText'] !== undefined &&
+    (obj as any)['getText'] !== undefined &&
```
BeksOmega, line 27: Forgot to check for `setBubbleSize` and `getBubbleSize`

maribethb: I guess I thought this was intentional because if we're checking from `Block` then it could be headless so they don't need to care about bubble size necessarily. But I guess if they only care about headless they prooobably don't have a need to override the class to begin with?

BeksOmega: Headless ones also need to have these methods so that the size properly gets round-tripped through serialization.

I also didn't think they were necessary, and I actually left those methods out originally. I only added them to the interface when I ran into the compiler errors :P Which is why I forgot to add them to the type guard.

Issue #7715, opened 2023-12-13
Typescript error for dropdown field


pachtymi:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

When appending a dropdown field I get:
```
Argument of type 'FieldDropdown' is not assignable to parameter of type 'string | Field<string | undefined>'.
  Type 'FieldDropdown' is not assignable to type 'Field<string | undefined>'.
    Types of property 'validator_' are incompatible.
      Type 'FieldValidator<string> | null' is not assignable to type 'FieldValidator<string | undefined> | null'.
        Type 'FieldValidator<string>' is not assignable to type 'FieldValidator<string | undefined>'.
          Types of parameters 'newValue' and 'newValue' are incompatible.
            Type 'string | undefined' is not assignable to type 'string'.
              Type 'undefined' is not assignable to type 'string'.
```

The code, which is straight from the docs:
```
this.appendDummyInput()
                .appendField('drop down:')
                .appendField(new Blockly.FieldDropdown([
                    ['first item', 'ITEM1'],
                    ['second item', 'ITEM2']
                ]), 'FIELDNAME');
```

### Reproduction steps

1. Attempt to put a new FieldDropdown into an `appendField`

pachtymi:
A workaround I can do is 
```
this.appendDummyInput()
                .appendField('drop down:')
                .appendField(new Blockly.FieldDropdown([
                    ['first item', 'ITEM1'],
                    ['second item', 'ITEM2']
                ]) as Blockly.Field, 'FIELDNAME');
```
But this is less than ideal.

BeksOmega:
Looks like this might be a change with typescript 5.3 https://github.com/google/blockly/pull/7713

BeksOmega:
Christopher added notes

- https://github.com/google/blockly/pull/7713#issuecomment-1905870343
- https://github.com/google/blockly/pull/7713#issuecomment-1905885333

maribethb:
@btw17 would you be interested in looking at this one next? it's something we wanted to fix this quarter but haven't been able to prioritize. there's some more details in the PR comments beka linked above

btw17:
> @btw17 would you be interested in looking at this one next? it's something we wanted to fix this quarter but haven't been able to prioritize. there's some more details in the PR comments beka linked above

Looks interesting - I'll take a look @maribethb!

btw17:
Found the issue! Thank you for your patience with this - wasn't able to work on this the last few weeks.

[...]

PR #7939, opened 2024-03-15
fix: updated field_dropdown to properly infer its Field type with TS 5.3.3

btw17:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7715

### Proposed Changes

Update `FieldDropdown`'s type information for `doClassValidation` to match the template in `Field` so it properly infers `FieldDropdown` as `Field<string>` instead of `Field<string|undefined>`.

### Reason for Changes

To enable Blockly to update to TypeScript 5.3.3 and beyond.

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

None

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -408,9 +408,7 @@ export class FieldDropdown extends Field<string> {
    * @param newValue The input value.
    * @returns A valid language-neutral option, or null if invalid.
    */
-  protected override doClassValidation_(
-    newValue: string,
-  ): string | null | undefined;
+  protected override doClassValidation_(newValue: string): string | null;
   protected override doClassValidation_(newValue?: string): string | null;
   protected override doClassValidation_(newValue?: string): string | null {
     const options = this.getOptions(true);
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -408,6 +408,10 @@ export class FieldDropdown extends Field<string> {
    * @param newValue The input value.
    * @returns A valid language-neutral option, or null if invalid.
    */
+  protected override doClassValidation_(
+    newValue: string,
+  ): string | null | undefined;
+  protected override doClassValidation_(newValue?: string): string | null;
   protected override doClassValidation_(newValue?: string): string | null {
     const options = this.getOptions(true);
     const isValueValid = options.some((option) => option[1] === newValue);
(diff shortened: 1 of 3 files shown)


core/field_dropdown.ts
```diff
@@ -408,6 +408,10 @@ export class FieldDropdown extends Field<string> {
    * @param newValue The input value.
    * @returns A valid language-neutral option, or null if invalid.
    */
+  protected override doClassValidation_(
```
rachel-fenichel, line 411: Is this a bad diff, or are there three lines in a row declaring the same function? And if it's the latter, why is that necessary and allowed?

core/field_dropdown.ts
```diff
@@ -408,6 +408,10 @@ export class FieldDropdown extends Field<string> {
...
+    newValue: string,
+  ): string | null | undefined;
```
btw17, line 413: Just noticed a nitpick - even thoug the base class has `| undefined`, this class has never expected that and it's actually not needed here. (I confirmed locally.) Updating the slight tweak.

maribethb: class validators are allowed to return undefined: https://developers.google.com/blockly/guides/create-custom-blocks/fields/validators#return_values

I think we need to keep the `| undefined` to allow subclasses of Dropdown to continue returning `undefined`

rachel-fenichel (approved):

core/field_dropdown.ts
btw17: This is accurate. This is actually how we have it defined in `field.ts`, too:
https://github.com/google/blockly/blob/master/core/field.ts#L1183

What this does is allow inferring type info based on what you're supplying. Taking `Field`'s `doClassValidation` as an example, *implementation*-wise, `undefined` means to take new value as is instead of transforming it in some way. What happens is if you provide a `newValue` as a non-undefined value, it infers the first definition that matches.

```ts
  // Providing a non-undefined value allows returning `undefined` (which means to take that value as is)
  protected doClassValidation_(newValue: T): T | null | undefined;
  // Providing an undefined or possibly undefined value means you can't return `undefined` since
  // that would mean take the default, though default could be `undefined` which is not valid to set.
  protected doClassValidation_(newValue?: AnyDuringMigration): T | null;
  // This accepts all the implementations above and we handle the cases in code through this.
  protected doClassValidation_(
    newValue?: T | AnyDuringMigration,
  ): T | null | undefined {
    if (newValue === null || newValue === undefined) {
      return null;
    }

    return newValue as T;
  }
```

core/field_dropdown.ts
btw17: Sounds good! I'll make the update.

btw17: Done!

maribethb (approved):

PR #7654, opened 2023-11-17
refactor(generators): Migrate Lua generators to TypeScript

btw17:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Part of https://github.com/google/blockly/issues/6828.

### Proposed Changes

Migrate `generators/lua/*.js` and `generators/lua.js` to TypeScript.

### Reason for Changes

Finish TS migration.

### Test Coverage

`npm test` passes. No changes to manual test procedures required.

### Documentation

No changes to documentation yet, as there should be minimal changes to the generated code and no changes to the published typings (because during build we overwrite the generated `.d.ts` files with the ones from `typings/`).

### Additional Information

It may be easier to review this commit-by-commit, as the final format commit make a lot of noisy changes.

PR #7940, opened 2024-03-15
fix(generators): Make `scrub_` `public`

### LINKED ISSUE #2156: Make Generator.scrub_ public
### Problem statement

[Generator.scrub_](https://github.com/google/blockly/blob/ca17e4019530c044ac828de4b3c57515572c67f4/core/generator.js#L391) gets called by the [procedure generators](https://github.com/google/blockly/blob/7679d227df8f62a12316334ee708d4a9a800ef17/generators/javascript/procedures.js#L59) in each language. It's also overridden in some languages. It should probably be renamed scrub and made public.

### Expected Behavior

scrub_ should be made public since it is called by other bits of code.

### Actual Behavior

scrub_ is private and is called anyway.

cpcallen:
## The basics

### Resolves

Fixes #2156

### Proposed Changes

Make remove `protected` access modifier from `CodeGenerator.prototype.scrub_`, making it public.

### Reason for Changes

In PRs #7602, #7616, #7646, #7647 and #7654 the `@protected` access modifier on `scrub_` overrides subclasse of `CodeGenerator` not transcribed to the new typescript signature.  I was going to re-add it, but this breaks some of the procedure block generator functions which rely on it, and then @BeksOmega pointed out that this might be one of the CodeGenerator API functions which we had already decided should be public—and lo and behold I found #2156.

Per discussion amongst team, I am not renaming it to scrub at this time.

### Test Coverage

Passes `npm test`.

### Documentation

Documentation will need updating in the usual way.

### DIFF
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -578,7 +578,7 @@ export class CodeGenerator {
    * @param _opt_thisOnly True to generate code for only this statement.
    * @returns Code with comments and subsequent blocks added.
    */
-  protected scrub_(
+  scrub_(
     _block: Block,
     code: string,
     _opt_thisOnly?: boolean,
--- a/core/generator.ts
+++ b/core/generator.ts
@@ -578,11 +578,7 @@ export class CodeGenerator {
    * @param _opt_thisOnly True to generate code for only this statement.
    * @returns Code with comments and subsequent blocks added.
    */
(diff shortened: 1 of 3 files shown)


maribethb (approved):
lgtm pending format

PR #7945, opened 2024-03-18
fix: comment create and delete events

BeksOmega:
## The basics

### Resolves

Work on #7909  

### Proposed Changes

Changes all of the comment events to use new workspace comments classes instead of the old ones.

Updates the create and delete events to use JSON.

Updates the workspace comments classes to fire create and delete events.

### Test Coverage

Updates tests. Adds tests for firing the events. Manually tested undo and redo for create and delete events.

### Documentation

N/A - docs for events are just reference docs, which are automatically generated.

### Additional Information

N/A

### DIFF
--- a/core/comments/workspace_comment.ts
+++ b/core/comments/workspace_comment.ts
@@ -62,8 +62,6 @@ export class WorkspaceComment {
 
   private fireCreateEvent() {
     if (eventUtils.isEnabled()) {
-      // TODO: Before merging. In the old class, this is wrapped by a setGroup
-      //   call. Is there any reason that's actually necessary?
       eventUtils.fire(new (eventUtils.get(eventUtils.COMMENT_CREATE))(this));
     }
   }
--- a/core/comments/workspace_comment.ts
+++ b/core/comments/workspace_comment.ts
@@ -8,6 +8,7 @@ import {Workspace} from '../workspace.js';
 import {Size} from '../utils/size.js';
 import {Coordinate} from '../utils/coordinate.js';
 import * as idGenerator from '../utils/idgenerator.js';
+import * as eventUtils from '../events/utils.js';
 
 export class WorkspaceComment {
   /** The unique identifier for this comment. */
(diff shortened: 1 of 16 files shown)


core/events/events_comment_base.ts
```diff
@@ -102,12 +100,10 @@ export class CommentBase extends AbstractEvent {
...
     if (create) {
-      const xmlElement = utilsXml.createElement('xml');
-      if (!event.xml) {
-        throw new Error('Ecountered a comment event without proper xml');
+      if (!event.json) {
+        throw new Error('Encountered a comment event without proper json');
       }
-      xmlElement.appendChild(event.xml);
-      Xml.domToWorkspace(xmlElement, workspace);
+      comments.append(event.json, workspace);
```
NeilFraser, line 106: BTW, this is why I preferred the older more verbose `Blockly.Foo.bar` syntax.  Took me a while to figure out what the heck 'comments' was.  Variable?  Nope.  Parameter?  Nope.  Global?  Nope.  ...

NeilFraser (approved):

core/comments/workspace_comment.ts
```diff
@@ -56,7 +57,21 @@ export class WorkspaceComment {
     // TODO: File an issue to remove this once everything is migrated.
     workspace.addTopComment(this as AnyDuringMigration);
 
-    // TODO(7909): Fire events.
+    this.fireCreateEvent();
+  }
+
+  private fireCreateEvent() {
+    if (eventUtils.isEnabled()) {
+      // TODO: Before merging. In the old class, this is wrapped by a setGroup
```
BeksOmega, line 65: @NeilFraser could you check this for me? Wanted a second set of eyes.

core/events/events_comment_base.ts
BeksOmega: I would take that up with JavaScript :P

maribethb (approved):

core/comments/workspace_comment.ts
maribethb: I don't see any point to it. It would be a group with definitely only one event in it.

The other places where we use groups, there is at least the possibility of multiple events being fired in between where we set true and false. Here, I don't see any such possibility. So I think it's fine to remove.

PR #7953, opened 2024-03-20
feat(dragging): Create (new) `IDragger` and `IDraggable` interfaces

### LINKED ISSUE #7838: Create `IDragger` and `IDraggable` interfaces
```JS
interface IDragger {
  constructor(dragged: IDraggable);

  // Lets the dragger handle any drag startup.
  onDragStart(e: PointerEvent);

  // Lets the dragger calculate where the element should actually be moved to.
  onDrag(e: PointerEvent, totalDelta: Coordinate);

  // Lets the dragger handle any drag cleanup.
  onDragEnd(e: PointerEvent);
}
```

```JS
interface IDraggable {
  // Returns true if the element is currently movable. False otherwise.
  isMovable(e: PointerEvent): boolean;

  // Returns the current location of the draggable in workspace coordiantes.
  getLocation(): Coordinate;

  // Lets the element handle any drag startup (e.g. moving to the front of the workspace)
  startDrag(e: PointerEvent);

  // Lets the element handle moving the new location, and updating any visuals based
  // on that (e.g. connection previews).
  drag(newLoc: Coordinate, target: IDragTarget, e?: PointerEvent);

  // Lets the element handle any drag cleanup.
  endDrag(e: PointerEvent);
}
```

See PR #7846 for initial implementation.

cpcallen:
## The basics

### Resolves

Fixes #7838.

### Proposed Changes

- Rename existing `core/interfaces/i_draggable.ts` to `core/interfaces/i_draggable.old.ts` and update imports of it.
- Introduce new `IDragable` interface in `core/interfaces/i_draggable.ts`.
- Introduce new `IDragger` interface in `core/interfaces/i_dragger.ts`.

### Reason for Changes

API changes to support custom dragging.

### Test Coverage

Passes `npm test`.

### Additional Information

1) I wasn't sure about whether to make the `target` parameter of `IDraggable`'s `drag` method optional or nullable.  Opinions welcome.

2) I wasn't clear whether the `totalDelta` parameter of `IDragger`'s `onDrag` method was intended to be cumulative or not, but I supposed it was given that the non-cumulative drag distance is presumably available via the (non-optional) `PointerEvent` parameter.  (It's slightly unclear to me why we are making the gesture responsible for keeping track of the cumulative drag, but perhaps it has reasons to do so already.)


core/interfaces/i_draggable.ts
```diff
@@ -6,9 +6,47 @@
...
+   * Handles moving elements to the new location, and updating any
+   * visuals based on that (e.g connection previews for blocks).
+   *
+   * @param newLoc Coordinate to which the draggable has been dragged.
+   * @param target The drag target the pointer is over, if any.
+   * @param e PointerEvent that started the drag; could be used to
+   *     check modifier keys, etc.  May be missing when dragging is
+   *     triggered programatically rather than by user.
+   */
+  drag(newLoc: Coordinate, target: IDragTarget | null, e?: PointerEvent): void;
```
BeksOmega, line 42: The style guide doesn't have guidance about whether to use `null` or `undefined` in general. But I would definitely prefer one or the other within an interface. Personally I would switch `target` to be optional.

core/interfaces/i_dragger.ts
```diff
@@ -0,0 +1,33 @@
...
+   * @param e PointerEvent that started the drag.
+   */
+  onDragStart(e: PointerEvent): void;
+
+  /**
+   * Handles calculating where the element should actually be moved
+   * to.
+   *
+   * @param e PointerEvent that continued the drag.
+   * @param totalDelta Cumulative drag distance, in workspace
```
BeksOmega, line 22: Currently the equivalent function uses pixel units, not workspace units. That is probably what we want so we can support scrolling on dragging to the edge of the workspace in the scroll options plugin.

The reason we pass the total delta is because we don't actually start a drag until the mouse has already moved a certain distance, (e.g. we need to make sure when dragging from the fly out we're actually dragging toward the workspace). But we want the `IDraggable` to catch up with the mouse (otherwise it would be offset from where the user originally grabbed the `IDraggable`).

BeksOmega (approved):

core/interfaces/i_dragger.ts
cpcallen: JSDocs updated accordingly.

core/interfaces/i_draggable.ts
```diff
@@ -4,11 +4,63 @@
...
+  /** Returns the current location of the draggable in workspace coordinates. */
+  getLocation(): Coordinate;
+}
 
 /**
- * The interface for an object that can be dragged.
+ * Represents an object that can be dragged.
  */
-export interface IDraggable extends IDeletable {}
+export interface IDragStrategy {
```
BeksOmega, line 24: Curious why you separated this out into a separate interface. I don't think it makes sense to have `BlockSvg` and `RenderedWorkspaceComment` implement `IDragStrategy` since that's not what they are. In my mind are draggable things, they have drag strategies.

core/interfaces/i_draggable.ts
```diff
@@ -4,11 +4,63 @@
...
+   * @param newLoc Workspace coordinate to which the draggable has
+   *     been dragged.
+   * @param e PointerEvent that continued the drag; could be used to
+   *     look up any IDragTarget the pointer is over, check modifier
+   *     keys, etc.
+   * @param target The drag target the pointer is over, if any.  Could
+   *     be supplied as an alternative to providing a PointerEvent for
+   *     programatic drags.
+   */
+  drag(newLoc: Coordinate, e?: PointerEvent): void;
```
BeksOmega, line 48: If you want to structure the method signature like this, should the event be non-optional? Otherwise you basically have three function signatures:

```
drag(newLoc: Coordinate);
drag(newLoc: Coordinate, e: PointerEvent);
drag(newLoc: Coordinate, target: IDragTarget);
```

core/interfaces/i_draggable.ts
```diff
@@ -4,11 +4,63 @@
...
+
+  /**
+   * Handles moving elements to the new location, and updating any
+   * visuals based on that (e.g connection previews for blocks).
+   *
+   * @param newLoc Workspace coordinate to which the draggable has
+   *     been dragged.
+   * @param e PointerEvent that continued the drag; could be used to
+   *     look up any IDragTarget the pointer is over, check modifier
+   *     keys, etc.
```
BeksOmega, line 43: If provided, they /should/ look up the drag target and deal with it.

BeksOmega (changes requested):

core/interfaces/i_draggable.ts
cpcallen: The purpose is so that I can have `class BlockDragStrategy implements IDragStrategy` and have it tell me if I made any mistakes with my type signatures.

I agree that it "`IDraggable extends DragStrategy`" looks a bit funny, but other than omitting `IDragStrategy` entirely I don't see any way that `BlockSvg` and `RenderedWorkspaceComment` won't both satisfy it, regardless of whether they formally implement it or not.

core/interfaces/i_draggable.ts
cpcallen: That looks like what I intended.

If `e: PointerEvent` is non-optional then `target` would need to be nullable (or otherwise optional) to cover the case where you are doing a programmatic drag (no `PointerEvent`, unless maybe you synthesise one) and you don't happen to be over a drag target.

core/interfaces/i_draggable.ts
BeksOmega: Ah I missed that `IDraggable` was extending `IDragStrategy` this is fine!

cpcallen: Actually, there is a way to avoid that: the methods could be given different names in `IDragStrategy` than in `IDraggable`.  But that seems… gratuitous.

core/interfaces/i_draggable.ts
cpcallen: Good point.

BeksOmega (approved):

PR #7957, opened 2024-03-23
feat(dragging): Introduce `DragStrategy` and `BlockDragStrategy` [WIP]

### LINKED ISSUE #7840: Create `BlockDragStrategy`, implementing `IDraggable` for `BlockSvg`
Code that does work currently is in the `BlockDragger` class (in `core/block_dragger.ts`).

cpcallen:
## The basics

### Resolves

Fixes #7840.

### Proposed Changes

Introduce two new classes: the abstract `DragStrategy`, which is contains the parts of the drag strategy that are needed to deal with general dragging including interaction with delete areas, and `BlockDragStrategy`, which additionally handles connecting and disconnecting blocks.

### Reason for Changes

Most of the stuff that workspace comments will need re: drag strategy will also be needed by blocks.  (Bubbles might be a bit different if they can't be deleted.)

### Additional Information

* The `dragStart` methods are more or less complete; `drag` and `dragEnd` are not yet done.

* I wonder about the memory cost of keeping a `BlockDragStrategy` for ever block, especially when each one contains a connection previewer, etc.  Might we be able to reuse some of these, somehow?

* I don't think it will be necessary to remove the `couldConnect` parameter from `IDeleteArea`'s `wouldDelete` method, as was proposed in c1411dd44 (on Beka's `fix/would-delete` branch).

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -1,0 +1,127 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as blockAnimation from '../block_animations.js';
+import * as dom from '../utils/dom.js';
+import * as registry from '../registry.js';
+import * as eventUtils from '../events/utils.js';
+import type {BlockSvg} from '../block_svg.js';
+import {DragStrategy} from './drag_strategy.js';
+import type {IConnectionPreviewer} from '../interfaces/i_connection_previewer.js';
+import type {RenderedConnection} from '../rendered_connection.js';
+import type {WorkspaceSvg} from '../workspace_svg.js';
+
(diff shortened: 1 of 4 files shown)


cpcallen:
Updated this PR with slight revisions to previously-discussed commits, plus an additional commit implementing (most of) the drag method, again adapted from the existing code in `block_dragger.ts`.

Three outstanding issues I am aware of:

- I wasn't quite sure what the best way to compute the `delta` parameter used by `getConnectionCandidate`.  It seems like we may need to record the initial position in `startDrag`, thought that method does not take a location.
- There are some unused parameters in the hook functions on `DragStrategy` that are unused.  Lint wants these to be named with a leading underscore but I didn't want to change the JSDoc least implementers feel they have to use that naming convention when overriding.
- Still need to imlement `endDrag`.

PR #7965, opened 2024-03-26
fix: comment view styling

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

Was documenting customizing the view of workspace comments and realized there were a few things that needed fixing:
  * Setting colors on image elements via css doesn't actually work, so I removed that CSS variable.
  * We should have a default delete icon image.
  * The foldout icon image needed a better name than "arrow-dropdown"

### Test Coverage

Tested that all of the icons show up properly.

### Documentation

This was for documentation!

### Additional Information

N/A


core/comments/comment_view.ts
```diff
@@ -759,22 +758,18 @@ css.register(`
 .blocklyDeleteIcon {
   width: 20px;
   height: 20px;
-  display: none;
-  fill: var(--commentIconColour);
```
maribethb, line 763: This should work fine, I think the problem is that you're overriding this because the actual svg has "fill" as an inline attribute.

maribethb: Alright, that's annoying and I hate all of the weird tricks to make it work :(

maribethb (approved):

core/comments/comment_view.ts
BeksOmega: Yeah it sucks, I also tried using `<use>` tags instead, but those just don't show up at all.

PR #7969, opened 2024-03-28
feat!: Throw errors on missing inputs

### LINKED ISSUE #7665: `CodeGenerator.prototype.statementToCode` and `valueToCode` should throw when given an invalid input name
### Check for duplicates

### Problem

It is a common pattern for the code generators for blocks with statement inputs to look like:

```JS
export function blockWithStatementInput(block, generator) {
  const code = generator.statementToCode(block, 'STATEMENT_INPUT');
  return `/* prefix */ {\n${code}}\n /* postfix */`;
}
```

but if the given input name (in this case `'STATEMENT_INPUT'` does not correspond to one of the inputs on the block this fails silently, with `code` being set to `''`.  This is probably undesirable.

### Request

`CodeGenerator.prototype.statementToCode(block, name)` should throw an error if `block` does not have an input named `name`.

### Alternatives considered

As far as alternatives to this proposal: the obvious alternative is the status quo, which can leave developers confused about why their code is not working (see [a recent example in the forum](https://groups.google.com/g/blockly/c/V584R-nCan0)).

There are a few alternatives about _how_ this might be  achieved.  In particular, any of the following functions might be modified to detect this error and throw:

- `Block.prototype.getInput(name)` could throw if no input named `name` is found (currently returns `null` and is so documented).
- `Block.prototype.getInputTargetBlock(name)` could throw if no input named `name` is found (currently returns `null` and is so documented).
- `CodeGeneartor.prototype.blockToCode(block)` could throw if `block` is null (currently returns `''` and is so documented).
- `CodeGenerator.prototype.statementToCode(block, name)` could could throw if `block.getTargetBlock(name)` returns null (currently it carries on and calls `this.blockToCode(null)`, but no specific behaviour is documented).

I (@cpcallen) discussed this with @NeilFraser, who observed:

> Hmm, a getter returning null seems fine.  Whereas a 'toCode' function should throw in my opinion.  Getters are different from 'doers'.  I can't think of any case where we call `statementToCode` speculatively and expect '' to come back out.

My view is that `blockToCode` should throw when passed `null` (or any other non-`Block` value), but evidently the current behaviour is documented and relied upon: a trial implementation breaks 35 of the mocha tests currently.  Having only `statementToCode` throw is less helpful at catching developer bugs and would still be a breaking change, but it my be preferable as it is less disruptive, does not change _documented_ behaviour, and only causes seven mocha test failure.

### Additional context

_No response_

NeilFraser:
Previously generators could generate code from inputs that didn't exist and get back the empty string.  This silent failure was causing problems for diagnosing issues.  This PR changes the behaviour so that an error is thrown.

This will break generators which rely on the previous behaviour.  Several of our demo blocks needed editing to accommodate this change.

Resolves #7665

### DIFF
--- a/generators/dart/text.ts
+++ b/generators/dart/text.ts
@@ -128,10 +128,12 @@ export function text_charAt(
       return [code, Order.UNARY_POSTFIX];
     }
     case 'LAST':
-      at = 1;
-    // Fall through.
     case 'FROM_END': {
-      at = generator.getAdjusted(block, 'AT', 1);
+      if (where === 'LAST') {
+        at = 1;
+      } else {
+        at = generator.getAdjusted(block, 'AT', 1);
+      }
       const functionName = generator.provideFunction_(
         'text_get_from_end',
         `
@@ -140,7 +142,7 @@ String ${generator.FUNCTION_NAME_PLACEHOLDER_}(String text, num x) {
 }
 `,
       );
-      const code = functionName + '(' + text + ', ' + at + ')';
+      const code = `${functionName}(${text}, ${at})`;
       return [code, Order.UNARY_POSTFIX];
     }
     case 'RANDOM': {
--- a/generators/lua/text.ts
+++ b/generators/lua/text.ts
@@ -132,8 +132,6 @@ export function text_charAt(
   // Get letter at index.
   // Note: Until January 2013 this block did not have the WHERE input.
   const where = block.getFieldValue('WHERE') || 'FROM_START';
-  const atOrder = where === 'FROM_END' ? Order.UNARY : Order.NONE;
-  const at = generator.valueToCode(block, 'AT', atOrder) || '1';
   const text = generator.valueToCode(block, 'VALUE', Order.NONE) || "''";
   let code;
   if (where === 'RANDOM') {
@@ -154,6 +152,8 @@ end
     } else if (where === 'LAST') {
       start = '-1';
     } else {
+      const atOrder = where === 'FROM_END' ? Order.UNARY : Order.NONE;
+      const at = generator.valueToCode(block, 'AT', atOrder) || '1';
       if (where === 'FROM_START') {
         start = at;
       } else if (where === 'FROM_END') {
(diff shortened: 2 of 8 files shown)


generators/dart/text.ts
BeksOmega, line 1: Is this an unrelated change?

generators/lua/text.ts
BeksOmega, line 1: Is this an unrelated change?

BeksOmega (commented):
Generally LGTM! But I'm not groking the changes to the text files, so going to wait to approve until I know what that's doing :P

generators/dart/text.ts
NeilFraser: Nope, it's an example of the sort of block this PR will break.  This block mutates, sometimes it has an 'AT' input, sometimes not.  Previously the generator would just get `''` when generating from an input that doesn't exist.  Now it throws an error.

generators/lua/text.ts
NeilFraser: Nope, same as earlier change.  Same block, same input (AT), different language, and completely different code path.

generators/dart/text.ts
BeksOmega: Ah ok I see, the call to `getAdjusted` has been wrapped in the `else` so it only errors in the case where we do really expect there to be an input but it doesn't exist. Is that correct?

BeksOmega (approved):
Assuming my understanding is now correct, I'm going to go ahead and approve!

Didn't check if there are other blocks relying on the old behavior, so I'm trusting you caught all of 'em.

generators/dart/text.ts
NeilFraser: Yes, in the case of the dropdown being 'LAST' there is no 'AT' input, in the case of 'FROM_END' there is an 'AT' input.  Generators can no longer just try generating from an input that may or may not exist.

NeilFraser:
> Didn't check if there are other blocks relying on the old behavior, so I'm trusting you caught all of 'em.

These are all the ones I found.  But be aware that every generator (language) is different.  I can't completely guarantee that I caught every one.  And of course this will break everyone out there who has similar generators on blocks that mutate.

PR #7971, opened 2024-03-28
feat: have blocks use drag strategy

### LINKED ISSUE #7841: Add ability to set block drag strategy
Via (e.g.) `BlockSvg.prototype.setDragStratgegy(...)`.

Don't worry about setting workspace comment and bubble drag strategies for now, because YAGNI.

BeksOmega:
## The basics

### Resolves

Fixes #7841 

### Proposed Changes

 
Has the `BlockSvg` delegate to a `IDragStrategy` (which external devs can supply). This makes the blocks use the new drag system, and allows external devs to customize how blocks handle drags.

### Test Coverage

Tested manually with a following PR that updates the gesture system to use the dragger and raggables.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -36,7 +36,7 @@ import type {Input} from './inputs/input.js';
 import type {IASTNodeLocationSvg} from './interfaces/i_ast_node_location_svg.js';
 import type {IBoundedElement} from './interfaces/i_bounded_element.js';
 import type {ICopyable} from './interfaces/i_copyable.js';
-import type {IDraggable} from './interfaces/i_draggable.old.js';
+import type {IDragStrategy, IDraggable} from './interfaces/i_draggable.js';
 import {IIcon} from './interfaces/i_icon.js';
 import * as internalConstants from './internal_constants.js';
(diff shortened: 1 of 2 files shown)


core/block_svg.ts
```diff
@@ -154,6 +155,8 @@ export class BlockSvg
    */
   relativeCoords = new Coordinate(0, 0);
 
+  private dragStrategy: IDragStrategy = new BlockDragStrategy(this);
```
rachel-fenichel, line 158: Does this mean that there is one block drag strategy instantiated per block on the page?

BeksOmega: Yesh, but it shouldn't have much more memory than implementing them as methods on the block itself. If you're concerned about it I can do some profiling. Theoretically we could at Yet Another Registry to deal with this, but it would be pretty messy and confusing from the external dev perspective.

rachel-fenichel: Makes sense. I agree with not adding another registry.

rachel-fenichel (approved):

Issue #7844, opened 2024-02-07
Upgrade `Gesture` class to create a `Dragger` instance


BeksOmega:
This "upgrade" is mostly just simplifying the class to not have to worry about the difference between `BlockDragger`s and `BubbleDragger`s.

To construct the `IDragger`, the gesture system will need to get the draggable thing from `getSelected`, which means that `ISelectable` and `IDeletable` need some updates which can be found in the [design doc](https://docs.google.com/document/d/1EdKxQafwO_j1QmoTQrhfE2fwPLXAy5mm0yapPjQ1Oeo/edit?tab=t.0#heading=h.l8g6gmhlmn9i).

PR #7972, opened 2024-03-28
feat: have the gesture use a dragger for blocks

BeksOmega:
## The basics

### Resolves

Work on #7844

### Proposed Changes

Upgrades the gesture class to use a `IDragger` for dragging blocks instead of using a `IBlockDragger`. This conforms to our new design for handling dragging.

### Test Coverage

Manually tested with all of the previous PRs, and everything seems to wire together correctly! Once this is in I can hammer it more.

### Documentation

N/A

### Additional Information

As I write the other `IDraggable`s I'll migrate the gesture to pass those to the dragger instead of using special draggers. (i.e. the `workspaceDragger` and `bubbleDragger`)


core/gesture.ts
```diff
@@ -1235,9 +1230,6 @@ export class Gesture {
    * @internal
    */
   getInsertionMarkers(): BlockSvg[] {
```
rachel-fenichel, line 1232: I'm probably missing context, but: why does this function continue to exist but always return an empty list? Will you re-implement it later?

BeksOmega: Mmm I left this like this because I wasn't quite sure what to do with it, since the draggable doesn't have APIs for getting insertion markers, cuz that's super specific to blocks. But I think I can reimplements this by just walking the workspace and checking for any blocks that are flagged as insertion markers. I'll check where this is getting called from and whether that's going to be a performance killer or not.

BeksOmega: So it turns out that `getInsertionMarkers` is only called from `workspace.render`, so I just moved the relevant logic into there. Performance impact shouldn't be an issue, because we don't actually call that method. And if people are calling it and run into problems, I can give them more performant ways of triggering rerenders.

rachel-fenichel (approved):

PR #7973, opened 2024-03-28
chore: Remove colour blocks from code demo.

### LINKED ISSUE #7826: Verify code demo will not be impacted by moving blocks to plugins
The storage demo in `demos/storage` was created to demonstrate how to save and load Blockly programs from a web app, but is known to be used by members of the public as a way to provide a URL that will load a particular Blockly program.

As we are planning to delete some of the current blocks from this repository after moving them to the blockly-samples repository (see also #7226), this has the potential to result in some of the saved Blockly programs in the storage backend becoming un-loadable.

A quick look at the `demos/storage/index.html` suggests that this demo has provided only a very limited toolbox, and it does not appear that any of the provided blocks are on the to-delete list, but it would be worth verifying what blocks are actually used in stored programs, since programs may have been created with blocks not currently provided in the toolbox but still present (until now) in the block definitions in `blocks/`.

@NeilFraser may be able to assist with auditing the contents of the storage backend.

### LINKED ISSUE #7938: Update Blockly code demo to pull in colour blocks from plugins
### Check for duplicates

### Problem

Necessary cleanup after deleting fields from core. The code demo has fields and blocks that have been deleted, and we know that the code demo gets used for storing full programs.

### Request

The fix is to pull in and install the colour blocks from the new plugins. This has to happen after the blockly-samples `blocks_for_fields` branch is merged and the plugins are published.

### Alternatives considered

Delete the blocks from the demo. We shouldn't do this unless we can verify (through inspecting saved code) that they are never used. That verification is probably more work than just importing the blocks.

### Additional context

_No response_

NeilFraser:
The colour-picker is leaving core.

Removing the colour blocks means that any stored programs that use colour blocks will no longer load.  However, since the code demo has no visual output, the colour blocks serve no use.  An examination of the database shows no usable programs which use the colour blocks.

Resolves #7826
Resolves #7938

### DIFF
--- a/demos/code/index.html
+++ b/demos/code/index.html
@@ -350,44 +350,6 @@
       </block>
       <block type="lists_sort"></block>
     </category>
-    <category name="%{BKY_CATCOLOUR}" colour="%{BKY_COLOUR_HUE}">
-      <block type="colour_picker"></block>
-      <block type="colour_random"></block>
-      <block type="colour_rgb">
-        <value name="RED">
-          <shadow type="math_number">
-            <field name="NUM">100</field>
-          </shadow>
-        </value>
-        <value name="GREEN">
-          <shadow type="math_number">
-            <field name="NUM">50</field>
-          </shadow>
-        </value>
-        <value name="BLUE">
-          <shadow type="math_number">
-            <field name="NUM">0</field>
-          </shadow>
-        </value>
-      </block>
-      <block type="colour_blend">
-        <value name="COLOUR1">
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):

BeksOmega:
Mmm I didn't realize this was against develop. (same with https://github.com/google/blockly/pull/7969 ) It's fine cuz we gave up on doing a release before v11. But this should have been against the rc/v11 branch. My bad for not checking!

PR #7975, opened 2024-03-29
fix!: various drag-adjacent interfaces

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

Fixes up the `ISelectable` and `IDeletable` interfaces to confirm to the proposed design. Also switches everything over to the new `IDraggable` interfaces. This is a breaking change because `IDraggable` classes are no longer required to be `IDeletable` as well.

### Test Coverage

Manually tested that all the dragging and drag targets continue to work.

### Documentation

N/A

### Additional Information

N/A


core/shortcut_items.ts
```diff
@@ -105,7 +108,9 @@ export function registerCopy() {
         !workspace.options.readOnly &&
         !Gesture.inProgress() &&
         selected != null &&
+        isDeletable(selected) &&
```
rachel-fenichel, line 111: I see what it's doing, but this code feels weird: `isDeletable(selected) && selected.isDeletable()`. Without context I would have no idea what eat part of that is doing. Is there anything you can do with the name to make it clearer? Maybe `implementsDeletable(selected) && selected.isDeletable()`?

BeksOmega: Done!

rachel-fenichel: Now draggable, selectable, and deletable are inconsistent. Can you change all of them to `implementsX`?

If no because of other dependencies, better to have them consistently be `isX`.

BeksOmega: I think the issue is we already have a bunch of type guards that are `isX`, so unless we want to break all of those we're going to have inconsistency anyway. Is it better to have a group of type guards be different or just the one?

rachel-fenichel: Okay, then revert to `isDeletable`.

rachel-fenichel (approved):

PR #7998, opened 2024-04-03
fix: blocks not being returned to the correct layer if their drag is reverted with no parent

BeksOmega:
## The basics

### Resolves

Fixes N/A

### Proposed Changes

See PR title.

### Test Coverage

Tested that blocks that didn't have parents when their drag started are properly returned to the `BLOCK` layer.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/trashcan.ts
+++ b/core/trashcan.ts
@@ -697,6 +697,14 @@ export class Trashcan
     };
     return blockInfo;
   }
+
+  wouldDelete(_element: IDraggable): boolean {
+    return false;
+  }
+
+  shouldPreventMove(_dragElement: IDraggable): boolean {
+    return true;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


core/trashcan.ts
```diff
@@ -697,6 +697,14 @@ export class Trashcan
     };
     return blockInfo;
   }
+
+  wouldDelete(_element: IDraggable): boolean {
+    return false;
```
maribethb, line 702: is this supposed to be part of this change? shouldn't the traschcan be a delete area?

BeksOmega: Ah whoops yes! This was just for testing and accidentally got committed. Thank you for the catch.

maribethb (approved):

Issue #7979, opened 2024-03-29
ctrl+click does not open context menu in firefox on mac


maribethb:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

The context menu should open when you either right-click the ws or when you hold ctrl and click on the workspace.

In firefox on mac, the former is working but not the latter

I'm wondering if this is a bug in firefox. I opened the inspector and set a breakpoint on the `pointerdown` event listener (through firefox's developer tools, not a specific line of blockly code). If you right click on the workspace, you hit this breakpoint. If you are holding ctrl, you do not. That points to it never being fired by the browser which is not something we can fix on our end. I found this bug was opened 6 years ago and is still open: https://bugzilla.mozilla.org/show_bug.cgi?id=1504210 but I find it curious nobody has reported this before.  [Edit: we do have a proposed fix; see comments below.]

Mark of App Inventor originally reported it and says it's only happening on specific versions of firefox

### Reproduction steps

1. Open https://blockly-demo.appspot.com/static/tests/playground.html in firefox on mac
2. Hold Ctrl key
3. Click on workspace

### Browsers

Firefox desktop

maribethb:
confirmed with this fiddle i found on the internet: https://jsfiddle.net/y8ot42pa/

ctrl+click in chrome yields pointerdown, ctrl+click in firefox does not (it does give a mousedown though)

mark-friedman:
Note that while  ctrl+click in firefox on Mac doesn't generate a "pointerdown" event it **does** generate a "contextmenu" browser event (as does on Chrome).  Consequently, I wonder if maybe Blockly  should consider using the "contextmenu" browser event  for context menu triggering rather than the "pointerdown" event. I think that that might also simplify Blockly's browser event handling a bit.

I am amazed that no one in the last 6 years or so has noticed this while using Blockly with Firefox on a Mac laptop!

maribethb:
This probably wasn't happening until we switched to pointer events, so not a full 6 years. But 2.

We'd be happy to accept a PR for that idea, Mark, but are unlikely to be able to prioritize a fix for this ourselves in the near future. As long as @BeksOmega says this is not going to collide with her current gesture work.

BeksOmega:
I think this might be a blocker for App Inventor @mark-friedman does any of the info Maribeth found change prioritization?

mark-friedman:
> I think this might be a blocker for App Inventor @mark-friedman does any of the info Maribeth found change prioritization?

[...]

PR #8009, opened 2024-04-05
fix: fixed ctrl+click for right click on mac

btw17:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7979

### Proposed Changes

Updated in `browser_events.ts` the `isRightButton` to allow command + click on Mac.

There's two relevant documents for this issue:
- https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey
- https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey

I don't use a Mac personally, so this will need to be verified. When using **BrowserStack**, though, `ctrl` activates `metaKey`, implying that it's translated to the `command` key. There are two important notes from the documentation above:
- The `command` key gets translated to `metaKey`. If this is what the issue is really about, perfect!
- The `control` key is not detectable alongside a click event, so `ctrlKey` should never be `true` for a click event on any browser on Mac.

### Reason for Changes

To fix the ctrl + click issue on Mac.

### Test Coverage

None

### Documentation

None

### DIFF
--- a/core/browser_events.ts
+++ b/core/browser_events.ts
@@ -187,7 +187,7 @@ export function isTargetInput(e: Event): boolean {
  * @returns True if right-click.
  */
 export function isRightButton(e: MouseEvent): boolean {
-  if (e.ctrlKey && userAgent.MAC) {
+  if ((e.metaKey || e.ctrlKey) && userAgent.MAC) {
     // Control-clicking on Mac OS X is treated as a right-click.
     // WebKit on Mac OS X fails to change button to 2 (but Gecko does).
     return true;
--- a/core/browser_events.ts
+++ b/core/browser_events.ts
@@ -7,7 +7,6 @@
 // Former goog.module ID: Blockly.browserEvents
 
 import * as Touch from './touch.js';
-import * as userAgent from './utils/useragent.js';
 
 /**
  * Blockly opaque event data used to unbind events when using
@@ -187,13 +186,14 @@ export function isTargetInput(e: Event): boolean {
  * @returns True if right-click.
  */
 export function isRightButton(e: MouseEvent): boolean {
-  if ((e.metaKey || e.ctrlKey) && userAgent.MAC) {
-    // Control-clicking on Mac OS X is treated as a right-click.
-    // WebKit on Mac OS X fails to change button to 2 (but Gecko does).
-    return true;
-  }
   return e.button === 2;
 }
+
+// REVIEW NOTE: Not needed unless funneled through a common space, like
+// 'handleWsStart'. If we confirm this isn't needed, will remove.
+// export function isContextMenu(e: PointerEvent): boolean {
+//   return e.type === 'contextmenu';
+// }
 
 /**
  * Returns the converted coordinates of the given mouse event.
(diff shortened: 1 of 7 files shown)


BeksOmega:
@maribethb I'm passing this to you since I similarly don't have a Mac :P lmk if you'd rather I assign it to Christopher or Neil instead.

core/browser_events.ts
```diff
@@ -187,7 +187,7 @@ export function isTargetInput(e: Event): boolean {
  * @returns True if right-click.
  */
 export function isRightButton(e: MouseEvent): boolean {
-  if (e.ctrlKey && userAgent.MAC) {
+  if ((e.metaKey || e.ctrlKey) && userAgent.MAC) {
```
btw17, line 190: Of note, I left `e.ctrlKey` here in case there's a legacy Mac case that works with it. The comment mentions "Mac OS X", which implies from my brief wiki check (https://en.wikipedia.org/wiki/MacOS) that it's for `Mac OS X 10.7` or before, which is 2011 and before.  I can't test this with Browser Stack :)

maribethb: This doesn't work unfortunately.

What this does is make the context menu open if you press command+click, but that's not the behavior we want. We want it to open on ctrl+click (which mirrors the behavior elsewhere in the os). 

I think the information in the original issue is needed here. The pointerevent while holding ctrl on firefox is never fired, so this method is never even hit because no event listener is triggered. We need to listen for the `contextmenu` event to be fired. That should happen in all browsers (even chrome) so if we listen for the `contextmenu` event we should be able to remove some of the code that checks for right-click events.

Btw, the distinction between macOS and OS X is unlikely to be relevant. It was called OS X for so long that it just still gets called that even though it's out of date.

btw17: Does this work for you on other browsers as well? According to the `ctrlClick` documentation, this ctrl + click shouldn't work on any browser:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey

> On Macintosh keyboards, this key is labeled the control key. Also, note that on a Mac, a click combined with the control key is intercepted by the operating system and used to open a context menu, so ctrlKey is not detectable on click events.

maribethb: Yeah, it works fine. In Chrome if you ctrl click you get 3 events: `pointerdown`, `mousedown`, and `contextmenu`. In all 3 events, `ctrlKey` is true

In Firefox you only get `mousedown` and `contextmenu` events but `ctrlKey` is still true in both of them.

So MDN must be wrong. Maybe macOS changed this at some point.

btw17: Good to know. Bummer! Testing in Browserstack with firefox on `macOS14` works for me since it translates `ctrl` to `command`.

If `contextmenu` works, though, which is also what the `ctrlKey` documentation implied above and you confirmed, perhaps I can attach a contextmenu event listener on Mac and proceed from there.

maribethb: Actually, MDN might be correct that `ctrlKey` isn't detectable on `click` events. But we're not listening to `click` events, so that is irrelevant to us :)

btw17:
Please take a look when you get the chance @maribethb! I've left some comments throughout for review.

If the functionality is confirmed and we think edge cases are covered, I'll remove the excess comments.

maribethb:
As an update of what was discussed offline, this wasn't working either so we passed this bug off to Neil. Thanks for trying though Blake! :)

Issue #7890, opened 2024-02-26
Dropdown field arrow misplaced on Safari when using Thrasos renderer


mikeharv:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When using the Thrasos renderer on Safari, all arrows in dropdown fields appear too high. This is evidently because Safari does not inherit the property `dominant-baseline: "central"` from the parent element. 

Note: Code.org fixed this by overriding `createTextArrow_` to add the attribute to the SVG element: https://github.com/code-dot-org/code-dot-org/pull/55480

### Reproduction steps

Observe any dropdown field while using Safari with Thrasos.

### Screenshots

<img width="173" alt="image" src="https://github.com/google/blockly/assets/43474485/7a7109d3-5052-45fc-bc6c-dcf050b8a840">

maribethb:
Thanks for reporting! Would you be interested in putting up a PR for the fix y'all have? :)

PR #8017, opened 2024-04-08
fix: correct dropdown arrow placement for Thrasos on Safari

mikeharv:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7890

### Proposed Changes

Ensure that the dominant baseline for the arrow symbol on dropdown fields is 'center'. The exception is when a field's text element's dominant baseline should not be centered. Previously this was true for certain browsers (IE and Edge), but now it's only true for the Geras renderer. (https://github.com/google/blockly/blob/9e05d69d2accba5dca97b39281b322a2ec0961d3/core/renderers/geras/constants.ts#L15)

Screenshots are of the Advanced Playground, renderer set to Thrasos on Safari. No regressions found for other renderers or other browsers.

**Before:**
image

**After:**
image

### Reason for Changes

When using the Thrasos renderer on Safari, all arrows in dropdown fields appear too high. This is evidently because Safari does not inherit the property dominant-baseline: "central" from the parent element.

### Test Coverage

No test have been added.

### Documentation

I don't believe this warrants any updates to documentation.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -229,6 +229,9 @@ export class FieldDropdown extends Field<string> {
           : ' ' + FieldDropdown.ARROW_CHAR,
       ),
     );
+    if (this.getConstants()!.FIELD_TEXT_BASELINE_CENTER) {
+      this.arrow.setAttribute('dominant-baseline', 'central');
+    }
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
Thanks for fixing this!

PR #8027, opened 2024-04-13
 cp.js DOM text reinterpreted as HTML

Shivam7-1:
## The basics

By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.


BeksOmega:
Hello, Thank you for the contribution! But I don't believe there's a risk of xss here since we're applying a constant value, and not any user input.

PR #8037, opened 2024-04-16
fix: bumping comments into bounds

BeksOmega:
## The basics

- [X] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes N/A

### Proposed Changes

Fixes bumping comments into bounds so it actually works.

### Test Coverage

Manually tested that bumping into bounds works properly.

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/8036

PR #8038, opened 2024-04-16
chore: delete old comments

### LINKED ISSUE #8004: Fixup types on getCommentById and related

BeksOmega:
## The basics

### Resolves

Fixes #8004 

### Proposed Changes

Deletes the old comments implementations so now we just have the new ones :D

### Test Coverage

Project builds and tests pass.

### Documentation

N/A

### Additional Information

Dependent on https://github.com/google/blockly/pull/8037


core/bump_objects.ts
```diff
@@ -93,12 +93,15 @@ export function bumpIntoBoundsHandler(
   workspace: WorkspaceSvg,
 ): (p1: Abstract) => void {
   return (e) => {
+    console.log('got event');
```
gonfunko, line 96: Clean up the logging in this file?

core/workspace_svg.ts
```diff
@@ -2143,8 +2142,8 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
    *
    * @param comment comment to add.
    */
-  override addTopComment(comment: OldWorkspaceComment) {
-    this.addTopBoundedElement(comment as OldWorkspaceCommentSvg);
+  override addTopComment(comment: WorkspaceComment) {
```
gonfunko, line 2145: Are these just not being updated to RenderedWorkspaceComment for backwards compatibility?

BeksOmega: That would break type compatibility :/ You can't have a subclass that expects a more restricted type than the superclass does. We never actually pass the base class to this function, so from a functional perspective we should be able to do that. But typescript doesn't know that so it prevents us from defining the types that way.

This is all knock on effects of the way we've set up rendered vs headless workspaces :/

core/bump_objects.ts
BeksOmega: Ah my bad, forgot to clean these up when I rebased.

core/workspace_svg.ts
gonfunko: Oh yeah, missed the `override`, that makes sense. Thanks!

gonfunko (approved):

PR #8041, opened 2024-04-16
fix: Correct list's "find" and "get" blocks help.

NeilFraser:
Both blocks were using the same help URL.

Two unrelated results of recompiling messages:
* Allow unwanted 'qqq' messages to be reverted.
* 'constants.json' is unchanged, but sorted.

Reported externally here: https://groups.google.com/g/blockly/c/jR7rmdplSwg


msg/json/synonyms.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"CONTROLS_FOREACH_INPUT_DO": "CONTROLS_REPEAT_INPUT_DO", "CONTROLS_FOR_INPUT_DO": "CONTROLS_REPEAT_INPUT_DO", "CONTROLS_IF_ELSEIF_TITLE_ELSEIF": "CONTROLS_IF_MSG_ELSEIF", "CONTROLS_IF_ELSE_TITLE_ELSE": "CONTROLS_IF_MSG_ELSE", "CONTROLS_IF_IF_TITLE_IF": "CONTROLS_IF_MSG_IF", "CONTROLS_IF_MSG_THEN": "CONTROLS_REPEAT_INPUT_DO", "CONTROLS_WHILEUNTIL_INPUT_DO": "CONTROLS_REPEAT_INPUT_DO", "LISTS_CREATE_WITH_ITEM_TITLE": "VARIABLES_DEFAULT_NAME", "LISTS_GET_INDEX_INPUT_IN_LIST": "LISTS_INLIST", "LISTS_GET_SUBLIST_INPUT_IN_LIST": "LISTS_INLIST", "LISTS_INDEX_OF_INPUT_IN_LIST": "LISTS_INLIST", "LISTS_SET_INDEX_INPUT_IN_LIST": "LISTS_INLIST", "MATH_CHANGE_TITLE_ITEM": "VARIABLES_DEFAULT_NAME", "PROCEDURES_DEFRETURN_COMMENT": "PROCEDURES_DEFNORETURN_COMMENT", "PROCEDURES_DEFRETURN_DO": "PROCEDURES_DEFNORETURN_DO", "PROCEDURES_DEFRETURN_PROCEDURE": "PROCEDURES_DEFNORETURN_PROCEDURE", "PROCEDURES_DEFRETURN_TITLE": "PROCEDURES_DEFNORETURN_TITLE", "TEXT_APPEND_VARIABLE": "VARIABLES_DEFAULT_NAME", "TEXT_CREATE_JOIN_ITEM_TITLE_ITEM": "VARIABLES_DEFAULT_NAME"}
```
BeksOmega, line 1: Is there a reason this is formatted as one line?

NeilFraser: It's autogenerated by Python.  I've adjusted the Python script to pretty-print.

NeilFraser: Hmm, in retrospect, pretty-printing might not be a great idea.  By having these two files as one single line, it used to be obvious that they were autogenerated (i.e. do not edit manually).  Now they look editable.  And JSON doesn't have a comment syntax, so we can't add a "do not edit" line.

Whatever.  Either way is fine by me.  Just pointing out that although this looks and reads better, someone will eventually get burned by it (like I would have been a half hour ago had this been prettier).

BeksOmega: haha I thought they were actually manually edited which is why I was confused by them being one liners, so fair enough :P Maybe adding a comment that they are autogenerated would be helpful in either case?

NeilFraser: As I said, JSON doesn't support comments.  It's the biggest issue with the format.  :(

BeksOmega: > As I said, JSON doesn't support comments. It's the biggest issue with the format. :(

Ah my bad, missed that!

cpcallen (commented):
I concur that it would be better _not_ to pretty-print autogenerated JSON files.

NeilFraser:
Added a 'comment' in the JSON.

scripts/i18n/js_to_json.py
```diff
@@ -114,29 +114,22 @@ def main():
   write_files(args.author, args.lang, args.output_dir, results, False)
 
   # Create synonyms.json.
-  synonyms_sorted = sort_dict(synonyms)
```
cpcallen, line 117: Why remove the sort?

NeilFraser: It was replaced with `sort_keys=True`.

cpcallen (approved):

PR #8049, opened 2024-04-24
fix: Don't record undo events for enable/disable

### LINKED ISSUE #7951: In rc/v11.0.0 you can't redo drag changes that disable blocks
### Check for duplicates

### Description

The `controls_flow_statements` and the `procedures_ifreturn` blocks are automatically disabled if they are dragged outside of any loop or procedure definition, respectively. In the `rc/v11.0.0` branch, such a drag change can be undone, but not redone. In fact, if such a change is performed, all previous changes in the undo history cannot be redone.

Drag changes that enable blocks seem to be fine. Only drag changes that disable blocks cause the bug.

### Reproduction steps

1. Check out the `rc/v11.0.0` branch.
2. Run `npm run start` to view the playground.
3. Drag a `controls_flow_statements` block into the workspace. It will be disabled.
4. Drag a `controls_repeat_ext` block into the workspace separately.
5. Drag the `controls_flow_statements` block into the `controls_repeat_ext` block as a child. It will become enabled (At this point, the changes so far are undoable and redoable.)
6. Drag the `controls_flow_statements` block back out of the `controls_repeat_ext` block. It will become disabled again.
7. Undo.
8. Attempt to redo. It won't work.
9. You can undo a few more steps but you won't be able to redo them.

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

Chrome desktop

NeilFraser:
There's no need to record and replay these events since the change will happen automatically anyway.

Resolves #7951

### DIFF
--- a/blocks/loops.ts
+++ b/blocks/loops.ts
@@ -20,6 +20,7 @@ import {
   createBlockDefinitionsFromJsonArray,
   defineBlocks,
 } from '../core/common.js';
+import * as eventUtils from '../core/events/utils.js';
 import '../core/field_dropdown.js';
 import '../core/field_label.js';
 import '../core/field_number.js';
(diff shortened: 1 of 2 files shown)


BeksOmega:
I think it probably doesn't make sense to merge this into develop since John refactored the enable / disable system for v11. https://github.com/google/blockly/pull/7958

Can you check if the v11 branch could benefit from a similar change?

NeilFraser:
While the bug doesn't appear in develop, this PR does reduce the number of events in the undo/redo stack.  Up to you whether you want to accept or close.

cpcallen (approved):

PR #8050, opened 2024-04-24
fix(build): Reintroduce shims for subpackage entrypoints

cpcallen:
## The basics

### Proposed Changes

Reintroduce  shims (`dist/core.js`, `dist/blocks.js`, `dist/javascript.js`, etc.) for submodule entrypoints in top level of package.  Have them created by `package_tasks.js` directly, rather than being copied from `scripts/package/`.

Assumptions:

- Such bundlers will _completely_ ignore the exports declaration.
- The bundles are intended to be used in a browser—or at least not in node.js—so the core entrypoint never needs to route to `core-node.js`.  This is reasonable since there's little reason to bundle code for node.js, and node.js has supported the exports clause since at least v12, considerably older than any version of node.js we officially support.
- It suffices to provide only a CJS entrypoint (because we can only provide CJS or ESM, not both.  (We could in future switch to providing only an ESM entrypoint instead, though.)

### Reason for Changes

The v11 beta breaks builds for external developers using browserify.

This change should solve issues encountered by users of bundlers that don't support `exports` at all (e.g. [browserify](https://browserify.org/)) as well as ones that don't support it in certain circumstances (e.g., when using webpack's [`resolve.alias] configuration option](https://webpack.js.org/configuration/resolve/#resolvealias) to alias 'blockly' to 'node_modules/blockly', as we formerly did in most plugins, which causes webpack to ignore blockly's `package.json` entirely).

### Documentation

We should mention in next beta release notes that even bundlers that do not support `exports` should continue to work.

### Additional Information

This is in effect a partial rollback of PR #7822.

### DIFF
--- a/scripts/gulpfiles/package_tasks.js
+++ b/scripts/gulpfiles/package_tasks.js
@@ -111,6 +111,50 @@ function packageUMDBundle() {
       .pipe(gulp.dest(`${RELEASE_DIR}`));
 };
 
+
+/**
+ * This task creates shims for the submodule entrypoints, for the
+ * benefit of bundlers and other build tools that do not correctly
+ * support the exports declaration in package.json.  These shims just
+ * require() and reexport the corresponding *_compressed.js bundle.
+ *
+ * This should solve issues encountered by users of bundlers that don't
+ * support exports at all (e.g. browserify) as well as ones that don't
+ * support it in certain circumstances (e.g., when using webpack's
+ * resolve.alias configuration option to alias 'blockly' to
+ * 'node_modules/blockly', as we formerly did in most plugins, which
+ * causes webpack to ignore blockly's package.json entirely).
+ *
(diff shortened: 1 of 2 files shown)


BeksOmega (approved):
This looks good to me! Pulled it down and tested that the generated shims look correct.

Curious: does this support people using typescript + browserify? I think you mentioned something briefly about shimming the .d.ts but I don't think we fully discussed it at the team meeting.

cpcallen:
> Curious: does this support people using typescript + browserify? I think you mentioned something briefly about shimming the .d.ts but I don't think we fully discussed it at the team meeting.

Very conveniently there is already a `.d.ts` file for each of the new `.js` wrappers, so typescript type checking should work correctly, both for TSC and in the browser (even if the latter does obey the `exports` directive, which specifies the path to use.).

PR #8053, opened 2024-04-26
fix: prevent duplication of comment icon bubbles

riknoll:
Discussed this fix offline with @BeksOmega.

Fixes two related issues: 
* Comment bubbles were being duplicated when a workspace is loaded with a comment that is already expanded
* Comment bubbles were being created and orphaned whenever insertion markers were being created by the InsertionMarkerPreviewer

### DIFF
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -264,14 +264,17 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   }
 
   async setBubbleVisible(visible: boolean): Promise<void> {
-    if (visible && (this.textBubble || this.textInputBubble)) return;
-    if (!visible && !(this.textBubble || this.textInputBubble)) return;
-
+    if (this.bubbleVisiblity === visible) return;
     this.bubbleVisiblity = visible;
 
-    if (!this.sourceBlock.rendered || this.sourceBlock.isInFlyout) return;
-
     await renderManagement.finishQueuedRenders();
+
+    if (
+      !this.sourceBlock.rendered ||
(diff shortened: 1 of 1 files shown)


core/icons/comment_icon.ts
```diff
@@ -264,15 +264,16 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
...
 
     this.bubbleVisiblity = visible;
 
-    if (!this.sourceBlock.rendered || this.sourceBlock.isInFlyout) return;
+    if (!this.sourceBlock.rendered || this.sourceBlock.isInFlyout || this.sourceBlock.isInsertionMarker()) return;
 
     await renderManagement.finishQueuedRenders();
 
+    if (!this.sourceBlock.rendered || this.sourceBlock.isInFlyout || this.sourceBlock.isInsertionMarker()) return;
+
```
BeksOmega, line 276: Ah I missed that these checks were duplicated. It should be fine if you move the `await` line to be the first line in this function! Then you can remove the second check.

riknoll: fixed!

BeksOmega:
herm I will check why tests are sad.

BeksOmega:
Ok, top of the function makes round-tripping through serialization sad because we want to make sure `bubbleVisibility` gets set immediately, but before the `if (!this.sourceBlock.rendered || this.sourceBlock.isInFlyout || this.sourceBlock.isInsertionMarker()) return;` line is fine.

BeksOmega (approved):
Sorry somehow I missed your latest push. This looks great! Thank you so much for the fix and the back and forth :D

Issue #7966, opened 2024-03-27
input fields fire change events on invalid input before change is committed


maribethb:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

A text input field fires "intermediate field value change" events while the user is editing the input. When they press the "enter" key, their change is "committed" and we fire an actual `BLOCK_CHANGE` event with the old and new value.

However, if the value is rejected by a field validator, then a `BLOCK_CHANGE` event will be fired *before* the user commits their change. If a user does press enter while the text input contains an invalid value, then the value of the field will be reset to whatever it was before the user started typing at all.

In my opinion, this is a bug and not how these two events should behave. The goal of introducing the "intermediate" event was to allow developers to listen to events specifically that occur when the user commits or finalizes their change. Firing the final block change event before they commit the change seems wrong from that perspective. And really, firing the final block change event at all seems wrong, because if they do commit their change in that state, the value of the block doesn't change at all anyway, so the old and new value *should* match in theory.

### Reproduction steps

1. Create a block with a field validator that will sometimes reject the input
2. Enable event logging
3. Edit the block such that the validator you created will reject the input, but do not press enter
4. Notice that the field is still being edited, the background is red, and you get a `BLOCK_CHANGE` event fired even though you are still typing into the block

### Stack trace

```shell
This occurs because `doValueInvalid_` in `FieldInput` fires a `BLOCK_CHANGE` event.
```

PR #7991, opened 2024-04-02
feat!: change gestures to look at selected when dragging

BeksOmega:
## The basics

- [ ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes N/A

### Proposed Changes + reasons

Makes it so the gesture class looks at whatever is selected when it starts a drag. This will allow the multi select plugin to provide its own draggable for when multiple things are selected.

### Test Coverage

Manually tested and everything seems to be working! But this could definitely use some hammering.
* Dragging blocks
* Dragging child blocks
* Dragging the workspace
* Dragging flyouts
* Dragging blocks from the flyout
* Dragging on child blocks in the flyout
* Dragging blocks in mutators

### Documentation

N/A

### Additional Information

N/A

PR #8054, opened 2024-04-26
fix!: Only fire intermediate events when editing input with invalid text.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7966

### Proposed Changes

This propagates the `fireChangeEvent` parameter to the `doValueInvalid_` method, and uses it to conditionally inhibit the default behavior of firing a block change event when a field is assigned an invalid value.

I also made a significant change to the way `doValueInvalid_` reverts the field value. Previously, it copied the value from `this.htmlInput_!.getAttribute('data-untyped-default-value')`, which is always a string value. However, this doesn't really make sense for number fields, which usually contain a numerical value even when editing the associated text input widget. I replaced it with `this.valueWhenEditorWasOpened_` which is the exact value the field used to have.

Finally, I conditionally skip firing the block change event if the new (reverted) value was the same as the value was before the most recent (invalid) edit.

The intermediate field change events were already being fired correctly (in addition to the incorrect block change events) so I didn't need to change that.

### Test Coverage

I updated some existing unit tests to expect numerical values for numerical fields instead of string values when the validator returns null. I guess this is technically a breaking change.

### Documentation

N/A

### Additional Information

N/A

## Breaking changes / To fix

This affects the data type assigned to a field's value when an invalid value is provided. 

When a user selects a number field and types letters instead of numbers, the attempt to set the field's value to the user's input will fail, and instead the internal value of the field will be reset to what it used to be before the user started to edit it. However, the value assigned to the field in this case used to be a string containing numerals instead of the actual original number, whereas after this change the original number will be assigned instead.

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -1086,14 +1086,22 @@ export abstract class Field<T = any>
     }
 
     const classValidation = this.doClassValidation_(newValue);
-    const classValue = this.processValidation_(newValue, classValidation);
+    const classValue = this.processValidation_(
+      newValue,
+      classValidation,
+      fireChangeEvent,
+    );
     if (classValue instanceof Error) {
       doLogging && console.log('invalid class validation, return');
       return;
     }
 
     const localValidation = this.getValidator()?.call(this, classValue);
-    const localValue = this.processValidation_(classValue, localValidation);
+    const localValue = this.processValidation_(
+      classValue,
+      localValidation,
+      fireChangeEvent,
+    );
     if (localValue instanceof Error) {
       doLogging && console.log('invalid local validation, return');
       return;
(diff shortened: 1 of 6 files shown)


maribethb (approved):
LGTM. Can you add a section to the PR description that makes it clear to readers of the release notes why this is a breaking change and what they might need to do in response.  Here's an example in another PR https://github.com/google/blockly/pull/7991

maribethb:
Also note that the angle tests may need to be updated in the rc/v11 branch of blockly-samples. They've already been deleted from the rc/v11 branch of core because the fields themselves were removed.

johnnesky:
Thanks! I updated the description.

Actually I think the angle field tests in blockly-samples will be okay?

PR #8065, opened 2024-05-03
refactor: split out block serialization for creating new blocks from the flyout

gonfunko:
This PR adds a method for serializing a block to the flyout, which is used during the block dragging/creation process. This will allow subclasses to override it to customize the block that gets created on the target workspace.

This is needed by scratch-blocks to ensure that newly created blocks have different IDs from their source blocks in the flyout, to allow uniquely identifying blocks.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -1301,14 +1301,18 @@ export abstract class Flyout
     }
 
     // Clone the block.
-    const json = blocks.save(oldBlock) as blocks.State;
-    // Normallly this resizes leading to weird jumps. Save it for terminateDrag.
+    const json = this.serializeBlock(oldBlock);
+    // Normally this resizes leading to weird jumps. Save it for terminateDrag.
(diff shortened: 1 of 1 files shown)


BeksOmega:
This needs to go against the rc/v11 branch since we're not doing any releases from `develop` until we release v11.

BeksOmega (approved):
For posterity: This is definitely a kludge on top of an already bad API. But I don't think this is the time to dig into changing flyouts, because several of our major partners depend on various implementation details, and it would be really time consuming to untangle.

For context:
- This was the original fix where we made it so the first blocks dragged from a flyout would have a different ID. This is the functionality scratch needs for its reporter blocks. https://github.com/google/blockly/pull/7618
- This is the PR where we reverted that change, because it broke a bunch of tests for code.org. It seemed like fixing them was going to be so burdensome it would prevent them from updating. https://github.com/google/blockly/pull/7788

PR #8068, opened 2024-05-06
fix: Move right-clicking to 'contextmenu' event

### LINKED ISSUE #7979: ctrl+click does not open context menu in firefox on mac

NeilFraser:
Resolves #7979

Events are not consistent between platforms.  This needs testing on every platform we can get our hands on.

To do (takes about 20 seconds per browser):
* Right-click on the workspace, expect workspace context menu.
* Right-click on a block, expect block context menu (make sure it's not the workspace context menu).
* Repeat both the above with any alternative right-click gestures for your platform (on the Mac it is CTRL-click).
* Verify that left-click and drag works as expected on blocks and workspace.

- [x] Firefox Mac
- [x] Safari Mac
- [x] Chrome Mac
- [x] Opera Mac

- [x] Firefox Windows
- [x] Chrome Windows
- [x] Edge Windows

- [ ] Firefox Linux
- [ ] Chrome Linux

- [x] Mobile Android
- [ ] Mobile iOS

- [x] ChromeOS

Add any other browsers you have.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -219,7 +219,13 @@ export class BlockSvg
         svg,
         'pointerdown',
         this,
-        this.onMouseDown_,
+        this.onPointerdown_,
+      );
+      browserEvents.conditionalBind(
+        svg,
+        'contextmenu',
+        this,
+        this.onContextmenu_,
       );
     }
     this.eventsInit_ = true;
@@ -603,10 +609,23 @@ export class BlockSvg
    *
    * @param e Pointer down event.
    */
-  private onMouseDown_(e: PointerEvent) {
+  private onPointerdown_(e: PointerEvent) {
     const gesture = this.workspace.getGesture(e);
     if (gesture) {
       gesture.handleBlockStart(e, this);
+    }
+  }
+
+  /**
+   * Handle a contextMenu on an SVG block.
+   *
+   * @param e Pointer contextmenu event.
+   */
+  private onContextmenu_(e: PointerEvent) {
+    const gesture = this.workspace.getGesture(e);
+    if (gesture) {
+      gesture.handleBlockStart(e, this);
+      gesture.handleRightClick(e);
     }
   }
 
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -794,7 +794,7 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
         this.svgGroup_,
         'contextmenu',
         this,
-        this.onContextmenu_
+        this.onContextmenu_,
       );
       // This no-op works around https://bugs.webkit.org/show_bug.cgi?id=226683,
       // which otherwise prevents zoom/scroll events from being observed in
@@ -2505,7 +2505,7 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
       this.clearGesture();
       gesture = null;
     }
-    const isStart = (e.type === 'pointerdown' || e.type === 'contextmenu');
+    const isStart = e.type === 'pointerdown' || e.type === 'contextmenu';
     if (gesture) {
       if (isStart && gesture.hasStarted()) {
         console.warn('Tried to start the same gesture twice.');
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -787,8 +787,14 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
         this.svgGroup_,
         'pointerdown',
         this,
-        this.onMouseDown_,
+        this.onPointerdown_,
         false,
+      );
+      browserEvents.conditionalBind(
+        this.svgGroup_,
+        'contextmenu',
+        this,
+        this.onContextmenu_
       );
       // This no-op works around https://bugs.webkit.org/show_bug.cgi?id=226683,
       // which otherwise prevents zoom/scroll events from being observed in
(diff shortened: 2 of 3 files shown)


NeilFraser:
This PR may be tested here: https://neil.fraser.name/temp/blockly/tests/playground.html

BeksOmega:
Tested on chrome for chromeOS. Right click works, two finger tap works, and alt + tap works.

rachel-fenichel:
Tested on:
- Chrome on windows
  - Right-click with mouse
  - Two-finger tap
  - Two-finger click
  - Longpress on touchscreen
- Edge on windows (should be the same as Chrome)
  - Right-click with mouse
  - Two-finger tap
  - Two-finger click
  - Longpress on touchscreen
- Firefox on windows
  - Right-click with mouse
  - Two-finger tap
  - Two-finger click
  - Longpress on touchscreen


On Edge I once triggered the browser context menu instead of the Blockly context menu while using longpress on the touchscreen. I was unable to reproduce it.

NeilFraser:
There doesn't appear to be a long-press context menu in Chrome on Android (in Blockly or otherwise).  Other than the one for selected text.  So ... task failed successfully?

core/block_svg.ts
```diff
@@ -603,10 +609,23 @@ export class BlockSvg
    *
    * @param e Pointer down event.
    */
-  private onMouseDown_(e: PointerEvent) {
+  private onPointerdown_(e: PointerEvent) {
+    const gesture = this.workspace.getGesture(e);
+    if (gesture) {
+      gesture.handleBlockStart(e, this);
+    }
```
cpcallen, line 616:
```suggestion
    this.workspace.getGesture(e)?.handleBlockStart(e, this);
```

core/block_svg.ts
```diff
@@ -603,10 +609,23 @@ export class BlockSvg
...
+  /**
+   * Handle a contextMenu on an SVG block.
+   *
+   * @param e Pointer contextmenu event.
+   */
+  private onContextmenu_(e: PointerEvent) {
     const gesture = this.workspace.getGesture(e);
     if (gesture) {
       gesture.handleBlockStart(e, this);
+      gesture.handleRightClick(e);
```
cpcallen, line 628: I was trying to figure out if it makes sense to call _both_ `handeBlockStart` and `handleRightClick`, given that the latter `.dispose()`es the gesture: it seemed pointless to start a gesture if we already know we're going to show the context menu.  But if I understand correctly it's necessary: `handleBlockStart` sets `gesture.targetBlock`, which is used by `handleRightClick`.

This seems convoluted, perhaps enough to warrant some additional commentary here if it can't be simplified.

core/workspace_svg.ts
```diff
@@ -1604,10 +1610,23 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
    *
    * @param e Pointer down event.
    */
-  private onMouseDown_(e: PointerEvent) {
+  private onPointerdown_(e: PointerEvent) {
+    const gesture = this.getGesture(e);
+    if (gesture) {
+      gesture.handleWsStart(e, this);
+    }
```
cpcallen, line 1617:
```suggestion
    this.getGesture(e)?.handleWsStart(e, this);
```

core/workspace_svg.ts
```diff
@@ -1604,10 +1610,23 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
+  /**
+   * Handle a contextMenu on SVG drawing surface.
+   *
+   * @param e Pointer contextmenu event.
+   */
+  private onContextmenu_(e: PointerEvent) {
     const gesture = this.getGesture(e);
     if (gesture) {
       gesture.handleWsStart(e, this);
+      gesture.handleRightClick(e);
```
cpcallen, line 1629: See above.

cpcallen (approved):
LGTM but I'm not very familiar with DOM events or Blockly gestures, so perhaps worth requesting an additional review from someone who is.

core/block_svg.ts
rachel-fenichel: `handleRightClick` should be called from inside `doStart` in `gesture`, not directly from here. The gesture gets all of the information about what is going on with a click and decides what to do with it; the blocks and workspaces and fields just say that they've been hit.

`handleRightClick` is public, but only so that it could be called from `touch.ts:longStart`, and that is only called from inside `gesture:doStart`.

The code that checks if it should be a right click currently is:
```

    if (browserEvents.isRightButton(e)) {
      this.handleRightClick(e);
      return;
    }

    if (e.type.toLowerCase() === 'pointerdown' && e.pointerType !== 'mouse') {
      Touch.longStart(e, this);
    }

```

which happens from `handleBlockStart`. So a change to right-click handling should be done by changing that logic, not by adding an additional call into `handleRightClick` from outside of the gesture.

rachel-fenichel (changes requested):
See https://github.com/google/blockly/pull/8068#discussion_r1594457701

Issue #7882, opened 2024-02-22
Dialog methods don't allow you to assign subtypes


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

The [dialog methods](https://github.com/google/blockly/blob/develop/core/dialog.ts) (`setPrompt`, `setAlert`, `setConfirm`) don't allow you to assign subtype functions to those values, because of how they are being called.

For example, you can't provide a prompt function that has additional optional parameters, even though it is a subtype of the normal prompt function type, because we wrap the provided function in another function call.

```js
export function prompt(
  message: string,
  defaultValue: string,
  callback: (p1: string | null) => void,
) {
  promptImplementation(message, defaultValue, callback);
}
```

As you can see, `prompt` wraps `promptImplementation`, so any extra params passed to `prompt` don't get passed to `promptImplementation`.

### Suggested solution

Instead of assigning to `promptImplementation` we can assign to `prompt` directly.

I think this should work, but I might be missing something. Please let me know!

maribethb:
I think your solution is fine (I'm not sure why we have `promptImplementation` separated out but maybe for type reasons?

But I don't understand why this is needed. Either Blockly is the one calling prompt, in which case it's not going to pass any extra parameters, or the developer is calling their own prompt, which they could just do directly.

Can you clarify? Also, is there any need to prioritize this soon? If there is a good use case we can leave this as "help wanted" unless a partner has asked for this soon or something.

BeksOmega:
> I think your solution is fine (I'm not sure why we have `promptImplementation` separated out but maybe for type reasons?
> 
> But I don't understand why this is needed. Either Blockly is the one calling prompt, in which case it's not going to pass any extra parameters, or the developer is calling their own prompt, which they could just do directly.
>
> Can you clarify? Also, is there any need to prioritize this soon? If there is a good use case we can leave this as "help wanted" unless a partner has asked for this soon or something.

This is a blocker for spork block related things. Their prompts get created from within spork-blocks as well as spork-UI. I had the same thing about "just call your custom thing customly" but the cross-repo usage means it needs to live in the global namespace. So while there are other places one could put it, this just kinda makes the most sense, and  is a relatively simple type change for us. I think I already have the code in a branch, I just didn't want to put it up until there had been discussion.

maribethb:
[...]

PR #8084, opened 2024-05-07
feat: allow the dialog methods to take in extra parameters

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7882

### Proposed Changes

Makes it so the `setX` dialog methods assign directly to the property on the exported namespace tree instead of to an internal variable.

### Reason for Changes

Fixes a problem for Scratch where they didn't have a good place for their dialog code to live.

### Test Coverage

Added tests that assigned dialogmethods can take in extra parameters.

### Documentation

This is pretty niche so I don't think it needs explicit documentation.

### Additional Information

This doesn't have to go into v11, I just put it up against that in case it can be reviewed in time.

### DIFF
--- a/core/dialog.ts
+++ b/core/dialog.ts
@@ -16,13 +16,6 @@ let alertImplementation = function (
   }
 };
 
-let confirmImplementation = function (
-  message: string,
-  callback: (result: boolean) => void,
-) {
-  callback(window.confirm(message));
-};
-
 /**
  * Wrapper to window.alert() that app developers may override via setAlert to
  * provide alternatives to the modal browser window.
@@ -51,16 +44,12 @@ export function setAlert(alertFunction: (p1: string, p2?: () => void) => void) {
  * @param message The message to display to the user.
  * @param callback The callback for handling user response.
  */
-export function confirm(message: string, callback: (p1: boolean) => void) {
-  TEST_ONLY.confirmInternal(message, callback);
-}
-
-/**
- * Private version of confirm for stubbing in tests.
- */
-function confirmInternal(message: string, callback: (p1: boolean) => void) {
-  confirmImplementation(message, callback);
-}
+export let confirm = (
+  message: string,
+  callback: (confirmed: boolean) => void,
+) => {
+  callback(window.confirm(message));
+};
 
 /**
  * Sets the function to be run when Blockly.dialog.confirm() is called.
@@ -69,9 +58,12 @@ function confirmInternal(message: string, callback: (p1: boolean) => void) {
  * @see Blockly.dialog.confirm
  */
 export function setConfirm(
-  confirmFunction: (p1: string, p2: (p1: boolean) => void) => void,
+  confirmFunction: (
+    message: string,
+    callback: (confirmed: boolean) => void,
+  ) => void,
 ) {
-  confirmImplementation = confirmFunction;
+  confirm = confirmFunction;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


NeilFraser (approved):

cpcallen:
This PR is causing tests in blockly-samples repository to fail for `typed-variable-modal`—but seemingly only when I merge the `master` branch into `rc/v11.0.0` (see google/blockly-samples#2348) and then link that against a locally-built version of Blockly from this repo's `rc/v11.0.0` branch:

```
Error: Not implemented: window.alert
    at module.exports (webpack-internal:///./node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
    at Window.eval [as alert] (webpack-internal:///./node_modules/jsdom/lib/jsdom/browser/Window.js:864:7)
    at Object.alert$$module$build$src$core$dialog (webpack-internal:///../../../blockly/dist/blockly_compressed.js:575:605)
    at TypedVariableModal.onConfirm_ (webpack-internal:///./src/TypedVariableModal.js:205:56)
```

Linking to the commit before this PR landed prevents the test failures.
Running tests directly in the blockly-samples repository's `rc/v11.0.0` branch also does not exhibit test failures.

cpcallen:
After discussing with @BeksOmega, our working theory was that the test failure was being caused by [the call to sinon.stub()](https://github.com/google/blockly-samples/blob/337ce75ab72a74b6c9c2551f92442cbf848a5e90/plugins/typed-variable-modal/test/typed_variable_modal_test.mocha.js#L123) in the `typed-variable-modal` tests failing to successfully stub `Blockly.dialog.alert` now that that is `export let alert = …` rather than the previous `export function alert(…)`.

Having looked at `blockly_compressed.js`, I can confirm that this does appear to be probable.  Here is an extract from the compiled output for `core/dialog.ts`, rewriting for legibility:

```JS
var alert$$internal = (a,b) => {
  window.alert(a);
  b&&b()
},

module$build$src$core$dialog = {
  get alert(){ return alert$$internal },
  // ...
};
```
…where the object declared as `module$build$src$core$dialog` becomes `Blockly.dialog`.

In other words: Closure Compiler does correctly support `export let`, and does so by creating a `get` accessor method for the exported name.  It will actually do this for `export function`s too, since `function` declarations are equivalent to `var` and thus also mutable—though [it's clever enough to only emit `get` accessors for dynamic exports that are actually subject to subsequent assignment](https://closure-compiler-debugger.appspot.com/#input0%3Dexport%2520let%2520foo%2520%253D%2520'foo'%253B%250Aexport%2520let%2520bar%2520%253D%2520'bar'%253B%250Aexport%2520function%2520func1()%2520%257Breturn%2520'func1'%253B%257D%250Aexport%2520function%2520func2()%2520%257Breturn%2520'func2'%253B%257D%250Aexport%2520function%2520mutate()%2520%257B%250A%2520%2520foo%2520%253D%2520'bar'%253B%250A%2520%2520func1%2520%253D%2520function%2520func1a()%2520%257Breturn%2520'func1a'%253B%257D%253B%250A%257D%250A%26input1%26conformanceConfig%26externs%26refasterjs-template%26CHECK_TYPES%3Dtrue%26REWRITE_MODULES_BEFORE_TYPECHECKING%3Dtrue%26CLOSURE_PASS%3Dtrue%26PARSE_TYPE_ANNOTATIONS%3Dtrue%26PRETTY_PRINT%3Dtrue)

It's possible in principle for such an accessor to be stubbed, but Sinon would have to use `Object.defineProperty` rather than a simple assignment, and it probably doesn't.

Issue #7599, opened 2023-10-19
JSON deserialization does not fire variable create events for blocks


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

JSON deserialization does not fire variable create events for blocks. It does fire variable create events for serialized variables.

XML does properly fire variable create events.

### Reproduction steps

1. Run `npm run start` to run the playground.
2. Enable workspace event logging.
3. Load the following JSON into the playground:
```js
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "variables_set",
        "fields": {
          "VAR": {
            "id": "}XEzZCNb@$C$ZWc1%awz"
          }
        }
      }
    ]
  }
}
```

4. Observe how no variable create event is fired.

Apoorvgarg-creator:
@BeksOmega Can I work on this issue ?

BeksOmega:
This issue hasn't been triaged yet :/ but I can assign you to one of the [good first issues](https://github.com/google/blockly/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) or [help wanted issues](https://github.com/google/blockly/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)! Just comment on one and I'll get you assigned =)

BeksOmega:
## To Fix

1) Run the reproduction steps and make sure you can reproduce the problem.
2) Create a function in the [block serialization file](https://github.com/google/blockly/blob/develop/core/serialization/blocks.ts) that fires variable events like how the [XML serializer does](https://github.com/google/blockly/blob/3e75f4f6f6fdaf0d2c578a30b1725cd6d42a2767/core/xml.ts#L607).
3) Call that function from [`appendInternal`](https://github.com/google/blockly/blob/3e75f4f6f6fdaf0d2c578a30b1725cd6d42a2767/core/serialization/blocks.ts#L426) before the block create event is fired.
4) Run the reproduction steps and make sure you can no longer reproduce.

bsampada7:
Hi @BeksOmega , if you need this ticket off your hands, I would like to work on it

BeksOmega:
Go for it @bsampada7 =)

PR #8122, opened 2024-05-15
fix: JSON deserialization not firing variable create events for blocks

bsampada7:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7599

### Proposed Changes

Add code to fire VAR_CREATE event when a JSON is serialized and new variables are created while creating a block.

### Reason for Changes

Fixes the bug of variable create event not being fired, even though a variable is created when a block is created through JSON serialization.

### Test Coverage

Manually tested that the VAR_CREATE event fires when a variable is created during JSON serialization.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/serialization/blocks.ts
+++ b/core/serialization/blocks.ts
@@ -29,6 +29,7 @@ import {
 } from './exceptions.js';
 import * as priorities from './priorities.js';
 import * as serializationRegistry from './registry.js';
+import * as Variables from '../../core/variables.js';
 
 // TODO(#5160): Remove this once lint is fixed.
 /* eslint-disable no-use-before-define */
@@ -417,6 +418,7 @@ export function appendInternal(
   }
   eventUtils.disable();
 
+  const variablesBeforeCreation = workspace.getAllVariables();
   let block;
   try {
     block = appendPrivate(state, workspace, {parentConnection, isShadow});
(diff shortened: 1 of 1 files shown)


core/serialization/blocks.ts
```diff
@@ -425,6 +427,19 @@ export function appendInternal(
...
+    );
+    // Fire a VarCreate event for each (if any) new variable created.
+    for (let i = 0; i < newVariables.length; i++) {
+      const thisVariable = newVariables[i];
+      eventUtils.fire(
+        new (eventUtils.get(eventUtils.VAR_CREATE))(thisVariable),
+      );
+    }
+    // Block events come after var events, in case they refer to newly created
+    // variables.
```
BeksOmega, line 442: Would you mind separating this into a separate function?

BeksOmega (changes requested):
This looks great thank you for the fix! Just one organizational change then should be good to go =)

core/serialization/blocks.ts
bsampada7: Sure, just extracted it to a separate function

BeksOmega (approved):
Wonderful! Thank you so much for the fix :D

Issue #8137, opened 2024-05-16
Can't delete blocks that are being dragged by a shadow child


BeksOmega:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Description

See reproduction steps

### Reproduction steps

1. Load the following JSON
```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "controls_repeat_ext",
        "inputs": {
          "TIMES": {
            "shadow": {
              "type": "math_number",
              "fields": {
                "NUM": 10
              }
            }
          }
        }
      }
    ]
  }
}
```
2. Drag the block by the shadow block to the trashcan.
3. Observe how the trashcan does not open and the block cannot be deleted.

PR #8138, opened 2024-05-16
fix: dragging blocks by shadows to delete

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8137

### Proposed Changes

Reverts the old code for dragging shadows, which delegated to parents, and instead just select the parent from the beginning.

Click events still get fired on the shadow though because that's based on the start blocks not the selection.

### Test Coverage

Manually tested dragging and deleting by shadows.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -245,8 +245,8 @@ export class BlockSvg
 
   /** Selects this block. Highlights the block visually. */
   select() {
-    if (this.isShadow() && this.getParent()) {
-      this.getParent()!.select();
+    if (this.isShadow()) {
+      this.getParent()?.select();
       return;
     }
     this.addSelect();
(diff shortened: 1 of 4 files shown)


core/block_svg.ts
```diff
@@ -245,19 +245,16 @@ export class BlockSvg
...
     }
     this.addSelect();
   }
 
   /** Unselects this block. Unhighlights the blockv visually.   */
   unselect() {
-    if (this.isShadow()) {
-      this.getParent()?.unselect();
-      return;
-    }
```
cpcallen, line 260: Why was this code just deleted, and not replaced with:

```JS
    if (this.isShadow() && this.getParent()) {
      // Shadow blocks should not be selected.
      this.getParent()!.unselect();
      return;  // Or maybe fall through to this.removeSelect()?
    }
```

Doesn't the code as proposed in this PR create a seemingly-buggy asymmetry where where calling `select` and then `unselect` on a shadow block will leave the parent block selected?

PR #8141, opened 2024-05-17
fix: Zelos comments shouldn't be bold

### LINKED ISSUE #8061: Zelos comment text should not be bold
### Check for duplicates

### Description

See issue title.

### Reproduction steps

1. 
2. 
3.

### Stack trace

_No response_

### Screenshots

Good (old):
Screenshot 2024-05-01 3 23 43 PM
Bad (current):
Screenshot 2024-05-01 3 23 35 PM

### Browsers

_No response_

NeilFraser:
Adds a 'blocklyBubbleComment' class to the foreignobject element so that comments may be styled with CSS.

Resolves #8061


core/bubbles/textinput_bubble.ts
```diff
@@ -121,7 +121,7 @@ export class TextInputBubble extends Bubble {
 
     const body = document.createElementNS(dom.HTML_NS, 'body');
     body.setAttribute('xmlns', dom.HTML_NS);
-    body.className = 'blocklyMinimalBody';
+    body.className = 'blocklyMinimalBody blocklyBubbleComment';
```
BeksOmega, line 124: Text input bubbles are no longer comment specific, so I would prefer to avoid using `Comment` in the CSS class.

Instead of adding a new class you use the existing `blocklyTextArea` or `blocklyTextInputBubble`? And by "use" I mean delete this and change the CSS to reference them, because they're already added to some elements.

If not, can you change this to reference being a text input bubbles instead of a comment?

NeilFraser: Changed.  I'd initially been extremely nervous to do something like this because the CSS path `.blocklyTextInputBubble textarea` crosses a 'foreignobject' boundary.  All the browsers I've tested this on seem ok with this now.  But historically that would have been asking for trouble.

NeilFraser:
I've also added some rounding of needlessly precise constants on the resize UI.

core/bubbles/textinput_bubble.ts
BeksOmega: Oh good to know for future reference! I'll start checking that.

BeksOmega (approved):

PR #8146, opened 2024-05-17
fix: Restore resize corner UI

NeilFraser:
The corner UI for comment resizing got rewritten for V11.  This new UI is just two very short diagonal lines.  This isn't as clear, and a survey of other existant resizing UIs don't reveal any that are this minimal, and none that are on a rounded corner.  This PR restores the original look.

This is the V11 UI:

Screenshot 2024-05-17 at 17 30 34

This PR restores the original UI:

Screenshot 2024-05-17 at 17 35 40


BeksOmega:
This UI was to make it easier for Scratch and MakeCode to use this system, because they use the UI that I migrated to. I don't know that it actually makes that big of a difference for end-users, so I would kinda prefer to make it easier for our external dev friends.

NeilFraser:
Ok.

PR #8149, opened 2024-05-17
fix: delete area animation

### LINKED ISSUE #8098: Delete animation and sound don't play when dragging blocks to a delete area.
### Check for duplicates

### Description

Dragging a block to the trash or the toolbar deletes the block, but does not play the delete sound.  This is a regression in V11, it works properly on V10.  All browsers.

### Reproduction steps

1. Drag a block to the trash.

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

_No response_

BeksOmega:
## The basics

### Resolves

Fixes #8098 

### Proposed Changes

Hacks some things around to get delete animations to work. Everything should be internal. Please check!

### Documentation

Once we add an animation API we'll need to add those side effect requirements to the dragger docs.

### DIFF
--- a/core/layer_manager.ts
+++ b/core/layer_manager.ts
@@ -15,6 +15,8 @@ import {Coordinate} from './utils/coordinate.js';
 export class LayerManager {
   /** The layer elements being dragged are appended to. */
   private dragLayer: SVGGElement | undefined;
+  /** The layer elements being animated are appended to. */
+  private animationLayer: SVGGElement | undefined;
   /** The layers elements not being dragged are appended to.  */
   private layers = new Map<number, SVGGElement>();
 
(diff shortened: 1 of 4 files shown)


core/layer_manager.ts
```diff
@@ -48,6 +51,35 @@ export class LayerManager {
     return dom.createSvgElement(Svg.G, {}, svg);
   }
 
+  private createAnimationLayer(injectionDiv: Element) {
```
cpcallen, line 54: Please provide a suitable JSDoc for this (even if it's just an excerpt of the one from `appendToAnimationLayer`).

cpcallen (approved):

core/layer_manager.ts
BeksOmega: This is private so it doesn't need JSDoc. I think the method signature makes it pretty clear what it's doing.

PR #8150, opened 2024-05-17
fix: Only hide WidgetDiv if it is associated with the affected workspace.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly-samples/issues/2131

### Proposed Changes

This associates a workspace with the current `WidgetDiv`, so that when a workspace needs to clean up its chaff, it can detect whether the current `WidgetDiv` is associated with it and only clean it up if it's affected.

I added a workspace parameter to `WidgetDiv.show()` and `ContextMenu.show()`, and a `hideIfOwnerIsInWorkspace` method to `WorkspaceDiv`. I also updated some calls from `WidgetDiv.hide()` to `WidgetDiv.hideIfOwner(this)`.

### Reason for Changes

In https://github.com/google/blockly-samples/issues/2131 the minimap plugin calls `this.minimapWorkspace.zoomToFit()` in response to changes, which calls `WorkspaceSvg.hideChaff()` which calls the static class method `WidgetDiv.hide()` which unfocuses inputs in the main workspace because the input widget is shared between the main workspace and the minimap workspace. Intermediate change events do not trigger the issue, so most text input fields work fine, but standard BlockChange events trigger the issue and editing procedure names and parameters will update caller blocks in the same workspace which fires BlockChange events which will trigger the issue and immediately unfocus the input field after typing a single character.

### Test Coverage

I manually tested with the minimap plugin to make sure that procedures and procedure parameters could be renamed without immediately losing focus (and that basic widget hiding still seems to work). All existing unit tests pass.

### Documentation

N/A


NeilFraser (approved):
Looks ok to both me and Christopher.  Up to you whether you want to merge it now, or wait until after V11.

core/widgetdiv.ts
```diff
@@ -145,6 +161,19 @@ export function hideIfOwner(oldOwner: unknown) {
...
 }
+
+/**
+ * Destroy the widget and hide the div if it is being used by an object in the
+ * specified workspace.
+ *
+ * @param oldOwnerWorkspace The workspace that was using this container.
+ */
+export function hideIfOwnerIsInWorkspace(oldOwnerWorkspace: WorkspaceSvg) {
+  if (ownerWorkspace === oldOwnerWorkspace) {
```
maribethb, line 172: I think this should also hide if `ownerWorkspace` doesn't exist. That way for backwards compatibility if someone doesn't set the owner, we default to the old behavior of hiding it anyway.

johnnesky: Good idea! Done.

PR #8188, opened 2024-05-30
fix: change which element keydown is bound to from document to injection div

### LINKED ISSUE #8187: Keydown events should be bound to a workspace rather than a document
### Check for duplicates

### Description

Keydown events are currently bound to the entire HTML document.  This has, for example, the unfortunate effect that a copy/paste operation outside of the workspace can cause that operation also to take place on the workspace.  This is especially unfortunate, and confusing, when the workspace isn't even visible to the user.  The can happen in apps like App Inventor which has a UI design mode, where the Blockly workspace isn't shown, and a blocks mode, where the workspace is shown.  In that situation a copy/paste of something in the UI designer could cause a mysterious copy of a block to be created in the blocks editor.

### Reproduction steps

1. Run Blockly simple playground
2. Place a block on workspace
3. Select the block
4. Enter ctrl-C to copy the selected block
5. Place the browser pointer outside of the Blockly workspace
6. Enter ctrl-V
7. Notice that the block is pasted into the workspace

### Stack trace

_No response_

### Screenshots

_No response_

### Browsers

_No response_

mark-friedman:
## The basics

### Resolves

Fixes #8187

### Proposed Changes

Bind the keydown event to the workspace's injection div rather than the HTML document

### Reason for Changes

See #8187 for details.

### Test Coverage

I modified the unit tests which deal with keydown events.  I also tested it manually in Chrome, Firefox and Safari on a Mac.

### Documentation

The docstring for `bindDocumentEvents` was changed to reflect the change in the code.  I don't believe that that docstring is exposed in the reference documentation.

### Additional Information

In theory, this might be considered a breaking change, in that a developer might be depending on the current behavior, even though it is undocumented and possibly confusing to their users.  It is also possible that a developer might have some code which, for some reason, is expecting the event to be bound to the document.

### DIFF
--- a/core/inject.ts
+++ b/core/inject.ts
@@ -78,7 +78,6 @@ export function inject(
   });
 
   browserEvents.conditionalBind(subContainer, 'keydown', null, onKeyDown);
-  // subContainer.addEventListener('keydown', onKeyDown);
 
   return workspace;
 }
--- a/core/inject.ts
+++ b/core/inject.ts
@@ -77,7 +77,8 @@ export function inject(
     common.setMainWorkspace(workspace);
   });
 
-  subContainer.addEventListener('keydown', onKeyDown);
+  browserEvents.conditionalBind(subContainer, 'keydown', null, onKeyDown);
+  // subContainer.addEventListener('keydown', onKeyDown);
 
   return workspace;
 }
--- a/core/inject.ts
+++ b/core/inject.ts
@@ -77,6 +77,8 @@ export function inject(
     common.setMainWorkspace(workspace);
   });
 
+  subContainer.addEventListener('keydown', onKeyDown);
+
   return workspace;
 }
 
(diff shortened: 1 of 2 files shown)


cpcallen (commented):
This all seems pretty reasonable to me, but I'm soliciting additional views from the team as I'm not the most familiar with Blockly's keyboard event handling or indeed the DOM itself.

cpcallen:
Additional discussion questions for the team:

* On the topic of whether this is a breaking change: is it intended (i.e. a feature) that shortcuts work even if the focus is elsewhere in the document?  In particular…
* Does it present any usability issues if keyboard navigation depends on the user being able to focus on the div first?

BeksOmega:
> Does it present any usability issues if keyboard navigation depends on the user being able to focus on the div first?

I think we probably want this behavior. In my mind, keyboard nav should only apply when the user is focused on the div, so that the other parts of the site can also be navigable.

cpcallen (approved):
The consensus of the team is that this is a good change, and that it is indeed a fix and not a breaking change (even though it's possible that some developers could have been depending on the previous, undesirable behaviour).

Issue #8189, opened 2024-06-03
Disconnecting nested c-shaped blocks causes blocks to disappear


FlorianTschimben:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

![image](https://github.com/google/blockly/assets/138431573/9858ddb0-5db6-427a-bdad-b2f9f2ec32d3)
I created this Structure and when i drag the 2nd loop (count with i...) out of the outer loop, everything in it disappears, but the output code stays the same. 
![image](https://github.com/google/blockly/assets/138431573/caf10f61-928e-4c6b-91bf-9a7fb167989c)

### Reproduction steps
*Build from Inside out*
1. Recreate a structure with multiple loops in it like in the first picture
2. drag the second loop out of the first loop
3. watch the blocks disappear

### Stack trace

```shell
block_svg.ts:298  Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at BlockSvg$$module$build$src$core$block_svg.setParent (file:///C:/PSBlockly/blockly/blockly_compressed.js:1225:28)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnectInternal (file:///C:/PSBlockly/blockly/blockly_compressed.js:960:417)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnectInternal (file:///C:/PSBlockly/blockly/blockly_compressed.js:1204:479)
    at RenderedConnection$$module$build$src$core$rendered_connection.disconnect (file:///C:/PSBlockly/blockly/blockly_compressed.js:959:323)
    at BlockSvg$$module$build$src$core$block_svg.unplugFromStack_ (file:///C:/PSBlockly/blockly/blockly_compressed.js:977:134)
    at BlockSvg$$module$build$src$core$block_svg.unplug (file:///C:/PSBlockly/blockly/blockly_compressed.js:975:342)
    at InsertionMarkerPreviewer$$module$build$src$core$insertion_marker_previewer.hideInsertionMarker (file:///C:/PSBlockly/blockly/blockly_compressed.js:1387:315)
    at InsertionMarkerPreviewer$$module$build$src$core$insertion_marker_previewer.hidePreview (file:///C:/PSBlockly/blockly/blockly_compressed.js:1386:486)
[...]

PR #8203, opened 2024-06-10
fix: statement input blocks disappearing

BeksOmega:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8189

### Proposed Changes

The issue was we weren't properly clearing the state of the current candidate connection, so the preview was incorrect causing blocks to disconnect and disappear, because the place they were trying to be inserted was invalid.

To explain that slightly more linearly:
1) You start with a nested stack of 3 statement blocks, built from the inside out.
2) The middle block's current candidate connection is its child, because the state from when it connected to its child was never cleared.
3) You disconnect the middle block, and it finds the parent as the valid candidate, but the current candidate (its child) is better (I.e. closer) because its state was never cleared. So we use the current candidate instead.
4) The child blocks gets disconnected and set its parent to `null` so it can preview, but when that happens it tries to reorder its DOM position, which doesn't work because it's already on the drag layer, where it's not expected to be, causing the error from the original issue.

### Test Coverage

Manually tested.

### Documentation

N/A
### Additional Information

We should put out a patch for this.

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -99,6 +99,7 @@ export class BlockDragStrategy implements IDragStrategy {
 
     this.startLoc = this.block.getRelativeToSurfaceXY();
 
+    this.connectionCandidate = null;
     const previewerConstructor = registry.getClassFromOptions(
       registry.Type.CONNECTION_PREVIEWER,
       this.workspace.options,


rachel-fenichel:
The explanation makes sense, as does the fix. Thanks @FlorianTschimben for the repro steps.

In general, if there's other state the reset between drags, is there an advantage to doing it at the start of the next drag vs at the end of the previous drag? It feels like it's cleanup so I expected it to be at the end of a drag.

rachel-fenichel (approved):

BeksOmega:
> The explanation makes sense, as does the fix. Thanks @FlorianTschimben for the repro steps.
> 
> In general, if there's other state the reset between drags, is there an advantage to doing it at the start of the next drag vs at the end of the previous drag? It feels like it's cleanup so I expected it to be at the end of a drag.

There isn't any other state we need to set to `null` between drags. But there is a lot of state we collect at the start of the drag (e.g. the `startLoc`). So I decided to put it there. If we run into further trouble we can pull it out, but for a patch I'm just going to make the minimal change.

PR #8231, opened 2024-06-26
feat!: Add support for preserving block comment locations.

### LINKED ISSUE #8230: Preserve block comment positions when (de)serializing
### Check for duplicates

### Problem

Currently, the location of block comments in the workspace are not preserved/restored when (de)serializing them, unlike workspace comments.

### Request

To do this, we should:

* Extend `ICommentIcon` and its implementation with `setBubbleLocation(location: Coordinate): void` and `getBubbleLocation(): Coordinate` methods.
* Update `blockToDom()` to call `getBubbleLocation()` and persist the coordinate under `x` and `y` attributes in the block comment XML, and update `applyCommentTagNodes()` to move the comment to the specified location when deserializing. I realize we're not adding new functionality to XML serialization in general, but this particular change makes work on upgrading Scratch easier and is quite small/contained.
* Update `saveState()` and `loadState()` in `CommentIcon` to persist and restore the comment coordinates when using JSON serialization.
* Update the tests to exercise moving the comment and (de)serialization

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes  #8230

### Proposed Changes
This PR preserves the positions of block comments when (de)serializing a workspace.

### Reason for Changes
We allow users to position block comments arbitrarily, but while we preserve the location of workspace comments when (de)serializing the workspace, block comments have their positions lost. This is also needed to support the Scratch modernization effort, since Scratch persists block comment locations as well.

### Breaking Change
This PR modifies the ICommentIcon interface to have two additional methods: `setBubbleLocation(location: Coordinate): void` and `getBubbleLocation(): Coordinate | undefined`. If you have a class that implements ICommentIcon, you'll need to implement these methods as well. They should set and retrieve the location of the comment bubble/box, in workspace coordinates.

### DIFF
--- a/core/icons/comment_icon.ts
+++ b/core/icons/comment_icon.ts
@@ -58,6 +58,9 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   /** The size of this comment (which is applied to the editable bubble). */
   private bubbleSize = new Size(DEFAULT_BUBBLE_WIDTH, DEFAULT_BUBBLE_HEIGHT);
 
+  /** The location of the comment bubble in workspace coordinates. */
+  private bubbleLocation?: Coordinate;
+
   /**
    * The visibility of the bubble for this comment.
    *
@@ -149,7 +152,13 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
   }
 
   override onLocationChange(blockOrigin: Coordinate): void {
+    const oldLocation = this.workspaceLocation;
     super.onLocationChange(blockOrigin);
+    if (this.bubbleLocation) {
+      const newLocation = this.workspaceLocation;
+      const delta = Coordinate.difference(newLocation, oldLocation);
+      this.bubbleLocation = Coordinate.sum(this.bubbleLocation, delta);
+    }
     const anchorLocation = this.getAnchorLocation();
     this.textInputBubble?.setAnchorLocation(anchorLocation);
     this.textBubble?.setAnchorLocation(anchorLocation);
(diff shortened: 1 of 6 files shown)


core/icons/comment_icon.ts
```diff
@@ -259,6 +303,12 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
     }
   }
 
+  onBubbleLocationChange(): void {
+    if (this.textInputBubble) {
```
BeksOmega, line 307: Is this one supposed to only care about the textInputBubble? Or should it also care about textBubbles? I think they're movable even in read-only mode. What are your thoughts on how we should handle that?

BeksOmega (commented):
Logic LGTM assuming you tested that round-tripping the coords in RTL works correctly. Easy to typo that stuff.

Can you also:
1) Update the breaking change section of the PR description to tell people they'll need to implement these new methods if they're creating a custom comment icon.
2) File an issue to update the [docs on creating a custom comment icon](https://developers.google.com/blockly/guides/create-custom-blocks/icons/creating-custom-icons/override-built-in#comment_icon) to include this info?

core/icons/comment_icon.ts
gonfunko: I followed the model of `onSizeChange()`, which only concerns itself with the editable version. As you mentioned offline, presumably folks aren't serializing read-only workspaces, so I think this is fine to leave as-is.

BeksOmega (approved):

gonfunko:
> Logic LGTM assuming you tested that round-tripping the coords in RTL works correctly. Easy to typo that stuff.
Yep, verified roundtripping with each layout as well as going between them for both XML and JSON.
> 
> Can you also:
> 
> 1. Update the breaking change section of the PR description to tell people they'll need to implement these new methods if they're creating a custom comment icon.
Done
> 2. File an issue to update the [docs on creating a custom comment icon](https://developers.google.com/blockly/guides/create-custom-blocks/icons/creating-custom-icons/override-built-in#comment_icon) to include this info?
Sent you a CL updating the docs

PR #8240, opened 2024-07-02
fix: correctly report the pre-resize-drag size to workspace comment size change listeners

gonfunko:
…ize change listeners

## The basics

### Proposed Changes
This PR stores the workspace comment size at the start of a drag on the resize handle, and reports that value to size change listeners. Previously, because the comment was continuously resized during the drag, but size change events were only fired at the end, the "old" comment size calculated in the method that fires the size change event was identical to the new size, so size change listeners were misinformed.

### DIFF
--- a/core/comments/comment_view.ts
+++ b/core/comments/comment_view.ts
@@ -98,7 +98,7 @@ export class CommentView implements IRenderedElement {
 
   /** Whether this comment view has been disposed or not. */
   private disposed = false;
-  
+
   /** Size of this comment when the resize drag was initiated. */
   private preResizeSize?: Size;
 
@@ -525,7 +525,7 @@ export class CommentView implements IRenderedElement {
       e.stopPropagation();
       return;
     }
-    
+
     this.preResizeSize = this.getSize();
 
     // TODO(#7926): Move this into a utils file.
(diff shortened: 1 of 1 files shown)


core/comments/comment_view.ts
```diff
@@ -334,8 +337,12 @@ export class CommentView implements IRenderedElement {
    * elements to reflect the new size, and triggers size change listeners.
    */
   setSize(size: Size) {
-    const oldSize = this.size;
+    let oldSize = this.size;
     this.setSizeWithoutFiringEvents(size);
+    if (Size.equals(oldSize, this.size) && this.preResizeSize) {
```
BeksOmega, line 342: Why do we need to check that they're equal in addition to checking that the pre-resize size exists?

BeksOmega (approved):

core/comments/comment_view.ts
gonfunko: If they're different, presumably setSize isn't being called as a result of a drag, and we should defer to the normally-calculated oldSize.

gonfunko: Per offline discussion, removed this check since it is technically redundant.

BeksOmega (approved):

PR #8245, opened 2024-07-05
test(CI): Have GitHub use `npm ci` instead of `npm install`

cpcallen:
## The basics

### Proposed Changes

Have the GitHub `build` workflow, which runs our tests, use `npm ci` instead of `npm install` when installing dependencies.

### Reason for Changes

This should help ensure that CI catches PRs that update `package.json` without updating `package-lock.json` to match.

### Additional Information

Proposed in response to breakage caused by PR #8209.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,8 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Npm Install
-        run: npm install
+      - name: Npm Clean Install
+        run: npm ci
 
       - name: Linux Test Setup
         if: runner.os == 'Linux'


cpcallen:
OK the `build` workflow failed as desired. (since `package-lock.json` is mismatched against `package.json` at commit e01154c (upon which I created this branch).  Going to rebase this on the parent commit (which is not broken) to verify it passes there.

cpcallen:
OK, rebasing this onto a known-good commit did _not_ cause tests to start passing, because it turns out that the checkout workflow checks out a commit that's a merge commit merging the PR into `develop`, not the tip-of-PR-branch commit.  TIL.

Not sure when that changed but obviously it's a lot better because it tells you if your PR _will_ break `develop` when committed, not just whether it passes tests in isloation.

gonfunko (approved):

PR #8256, opened 2024-07-09
fix: export the layer constants.

gonfunko:
## The basics

## The details
This PR updates the layer constants to allow them to be referenced by third-party code.


BeksOmega (commented):
I think these have `@internal` tags that need to be removed as well.

gonfunko:
Whoops, good catch! Removed accordingly.

BeksOmega (approved):

PR #8284, opened 2024-07-10
feat: allow fetching block bounds not including children.

### LINKED ISSUE #7212: Access the block's height and width excluding child blocks
### Check for duplicates

### Problem

code.org wants to create a flyout field that is sized to the existing size of the block, so it takes up an entire input row. But currently there is no way to access with width of the block excluding children, so this isn't possible.

image

### Request

Add a way to access the block's height and width excluding child blocks. We cannot use `getHeightWidth` because that is already taken, but maybe we can use `getHeightWidthWithoutChildren`.

### Alternatives considered

N/A

### Additional context

N/A

gonfunko:
## The basics

### Resolves 

Fixes #7212

### Proposed Changes
This PR adds a `getBoundingRectangleWithoutChildren()` method to BlockSvg that returns the bounding box for the block alone, not including its children (for C blocks) or subsequent blocks in the stack.

### Reason for Changes
There was no public API to retrieve this information, and it's useful.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -407,6 +407,8 @@ export class BlockSvg
   moveTo(xy: Coordinate, reason?: string[]) {
     const curXY = this.getRelativeToSurfaceXY();
     this.moveBy(xy.x - curXY.x, xy.y - curXY.y, reason);
+    console.log(this.getBoundingRectangle());
+    console.log(this.getBoundingRectangleWithoutChildren());
   }
 
   /**
@@ -444,7 +446,7 @@ export class BlockSvg
    * @returns Object with coordinates of the bounding box.
    */
   getBoundingRectangle(): Rect {
-    return this.getBoundingRectangleWithWidth(this.getHeightWidth().width);
+    return this.getBoundingRectangleWithDimensions(this.getHeightWidth());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/block_svg.ts
```diff
@@ -436,16 +444,31 @@ export class BlockSvg
...
     if (this.RTL) {
-      left = blockXY.x - blockBounds.width;
+      left = blockXY.x - blockWidth;
       right = blockXY.x;
     } else {
       left = blockXY.x;
-      right = blockXY.x + blockBounds.width;
+      right = blockXY.x + blockWidth;
     }
     return new Rect(blockXY.y, blockXY.y + blockBounds.height, left, right);
```
BeksOmega, line 473: Does this still have the problem with the height changing depending on whether the block is selected or not?

gonfunko: Nope, results are consistent with the old `getBoundingRectangle()` method except when the block has children/next blocks.

core/block_svg.ts
```diff
@@ -407,6 +407,8 @@ export class BlockSvg
   moveTo(xy: Coordinate, reason?: string[]) {
     const curXY = this.getRelativeToSurfaceXY();
     this.moveBy(xy.x - curXY.x, xy.y - curXY.y, reason);
+    console.log(this.getBoundingRectangle());
+    console.log(this.getBoundingRectangleWithoutChildren());
```
BeksOmega, line 411: Remove logs plz.

BeksOmega (approved):

core/block_svg.ts
gonfunko: Done!

Issue #8291, opened 2024-07-10
Add a `blocklyFieldText` CSS class to fields' text elements


BeksOmega:
In the field's `createTextElement_` method, add a `blocklyFieldText` class to the [`class` property](https://github.com/google/blockly/blob/8323642d0eb677648f004f294208d2734c6e8d65/core/field.ts#L376).

https://github.com/google/blockly/blob/8323642d0eb677648f004f294208d2734c6e8d65/core/field.ts#L376

ga-fleury:
Hey @BeksOmega, can you assign this to me?

BeksOmega:
Closed by #8302

PR #8302, opened 2024-07-11
feat!: Add a blocklyFieldText CSS class to fields' text elements (#8291)

ga-fleury:
**feat!: Add a blocklyFieldText CSS class to fields' text elements (#8291)**

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -373,7 +373,7 @@ export abstract class Field<T = any>
     this.textElement_ = dom.createSvgElement(
       Svg.TEXT,
       {
-        'class': 'blocklyText',
+        'class': 'blocklyFieldText',
       },
       this.fieldGroup_,
     );


core/field.ts
```diff
@@ -373,7 +373,7 @@ export abstract class Field<T = any>
     this.textElement_ = dom.createSvgElement(
       Svg.TEXT,
       {
-        'class': 'blocklyText',
+        'class': 'blocklyFieldText',
```
BeksOmega, line 376:
```suggestion
        'class': 'blocklyText blocklyFieldText',
```

We want this class to be additional, not replacing! (Sorry I know this is different than some of the other issues)

BeksOmega (changes requested):

BeksOmega (approved):

Issue #8268, opened 2024-07-10
Add the block's type as a CSS class to the block's root SVG


BeksOmega:
## To Fix

In the [BlockSvg class' constructor](https://github.com/google/blockly/blob/17dfa462a357b553523393145c9d5045023e627b/core/block_svg.ts#L1924), call `dom.addClass` on the `this.svgGroup_` to add the `prototypeName` as a CSS class on the block.

The PR to fix this should be made against the `rc/v12.0.0` branch.

To work on this issue, please comment below asking to be assigned.

BeksOmega:
@ga-fleury is working on this.

Apocalypse96:
@BeksOmega I would love to solve this issue

sachan13harshit:
@BeksOmega can you pls assign it to me?

Aarya01Patil:
@BeksOmega Hey I would love to solve this issue

GarvitDadheech:
Hii @BeksOmega , Can you please assign this issue to me? I'd like to resolve this issue.

BarryByte:
@BeksOmega Please assign this to me.

CodeMaverick2:
@BeksOmega I would like to solve this

harsh-791:
i want to work on this issue @BeksOmega

BeksOmega:
This is open for work again @harsh-791 Are you still interested in it?

BeksOmega:
Nvm looks like @CodeMaverick2 is taking a swing =)

BeksOmega:
Closed by #8428

PR #8305, opened 2024-07-11
feat: Add the block's type as a CSS class to the block's root SVG (#8268)

ga-fleury:
Fixes: #8268 

**feat: Add the block's type as a CSS class to the block's root SVG (#8268)**

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -184,8 +184,6 @@ export class BlockSvg
     this.workspace = workspace;
     this.svgGroup_ = dom.createSvgElement(Svg.G, {});
 
-    dom.addClass(this.svgGroup_, prototypeName)
-
     /** A block style object. */
     this.style = workspace.getRenderer().getConstants().getBlockStyle(null);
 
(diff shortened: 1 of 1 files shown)


BeksOmega (approved):
This looks great! Thank you for the fix =) I'll get this merged once CI passes.

BeksOmega:
@ga-fleury , looks like some of the tests are failing. You can run them locally with `npm run test:mocha:interactive`. Can you take a look and try to figure out what's up?

BeksOmega (changes requested):
Marking this as changes requested

ga-fleury:
@BeksOmega Seems like the `BlockSvg` class is receiving empty `prototypeName` values somewhere, which means they wouldn't receive a CSS class. Adding a empty string check solves the issue, Should we pass a default class name when the value is empty?

BeksOmega:
> @BeksOmega Seems like the `BlockSvg` class is receiving empty `prototypeName` values somewhere, which means they wouldn't receive a CSS class. Adding a empty string check solves the issue (see below), Should we pass a default class name when the value is empty?
> 
> ```
>     if(prototypeName) {
>       dom.addClass(this.svgGroup_, prototypeName)
>     }
> ```

Yeah that LGTM! Thanks for digging into this :D

Normally I would say we should fix the tests instead, but there are other places where we're checking `if (prototypeName)`, so better to be cautious. One you send in the fix I'll give this another look!

ga-fleury:
@BeksOmega just pushed changes

BeksOmega (changes requested):
Can you run the same formatting fixes you did on the other PRs?

Issue #8310, opened 2024-07-11
Can't re-enable "disabled orphaned blocks" after migration to v11


clementcontet:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

My app disable automatically blocks that are orphaned (through `workspace.addChangeListener(Events.disableOrphans)`), and it is the only way to disable blocks (I have `options.disable = false`).

When loading a workspace that have been generated with v10, v11 automatically set the `disabledReasons` to `MANUALLY_DISABLED` (see https://github.com/google/blockly/issues/8095), but the users can't re-enable them (because of the `options.disable = false`).

I can't see a good solution to this issue :(

johnnesky:
Ah, I hadn't considered this case, thanks for the bug report!

I talked with Beka and she suggested that the https://www.npmjs.com/package/@blockly/disable-top-blocks plugin might help unblock your users. If you allow disabling in the workspace and add this plugin, users will be able to use the right-click context menu to enable and disable blocks but only when they're not currently orphaned.

Otherwise, you can wait for me to implement a workaround. I think that when any disabled reason is removed, and the workspace is configured to not allow manually disabling/enabling, and the block is currently marked as manually disabled, then the manually disabled reason should be removed in addition to the reason that initiated the change.

clementcontet:
Hi, thanks for the investigation and quick fix!
I'll wait for the next release 😀

PR #8354, opened 2024-07-12
fix: Enable blocks if user can't manually enable them.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8310

### Proposed Changes

When any disabled reason is removed, and the workspace is configured to not allow manually disabling/enabling, and a block is currently disabled for the manual reason, remove the manual reason in addition to whatever reason initiated the change.

### Reason for Changes

A legacy serialized workspace does not specify the reason why blocks are disabled. When loaded with blocks that had been disabled due to being orphaned but without a recorded reason, Blockly defaults to using the manually disabled reason. However, some workspaces are configured to disallow disabling/enabling blocks, in which case the user can't remove the manually disabled reason.

For the specific case where a workspace does not allow manually disabling or enabling blocks, and a block currently has the manually disabled reason, and some other disabled reason is being removed (e.g. the block is no longer orphaned), then removing the manually disabled reason will allow the block to be fully enabled again.

### Test Coverage

I haven't created any unit tests for this legacy workaround yet, but I did have to explicitly set the option `{disable: true}` in a test workspace to get one test method to pass since it was designed for the common case where users are allowed to manually disable blocks. 

I manually tested that the workaround behaves as intended by using the browser console to set `Blockly.getMainWorkspace().options.disable = false` after manually disabling a block.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/tests/mocha/blocks/procedures_test.js
+++ b/tests/mocha/blocks/procedures_test.js
@@ -25,7 +25,7 @@ import {defineRowBlock} from '../test_helpers/block_definitions.js';
 suite('Procedures', function () {
   setup(function () {
     sharedTestSetup.call(this, {fireEventsNow: false});
-    this.workspace = Blockly.inject('blocklyDiv', {});
+    this.workspace = Blockly.inject('blocklyDiv', {disable: true});
     this.workspace.createVariable('preCreatedVar', '', 'preCreatedVarId');
     this.workspace.createVariable(
       'preCreatedTypedVar',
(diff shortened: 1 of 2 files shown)


tests/mocha/blocks/procedures_test.js
```diff
@@ -25,7 +25,7 @@ import {defineRowBlock} from '../test_helpers/block_definitions.js';
 suite('Procedures', function () {
   setup(function () {
     sharedTestSetup.call(this, {fireEventsNow: false});
-    this.workspace = Blockly.inject('blocklyDiv', {});
+    this.workspace = Blockly.inject('blocklyDiv', {disable: true});
```
rachel-fenichel, line 28: Which test did this change target? Since it's in setup for a suite, I think it will impact multiple tests.

johnnesky: I've updated the test to only modify options in the one affected test method.

rachel-fenichel (approved):

PR #8357, opened 2024-07-14
fix!: Rename `blocklyTreeRow` and `blocklyToolboxCategory` CSS classes

### LINKED ISSUE #8345: Rename the `blocklyTreeRow` CSS class to `blocklyToolboxCategory`
Rename any references to `blocklyTreeRow` to `blocklyToolboxCategory`.

This change should be made against the `rc/v12.0.0` branch. This should be marked as a breaking change.

To work on this issue, comment below asking to be assigned.

Abhinav-Chdhary:
## The basics

## The details
- This commit renames the blocklyTreeRow CSS class to blocklyToolboxCategory.
- The commit was made against v12.0.0 branch
- It is a breaking change

### DIFF
--- a/core/toolbox/category.ts
+++ b/core/toolbox/category.ts
@@ -659,7 +659,7 @@ export type CssConfig = ToolboxCategory.CssConfig;
 
 /** CSS for Toolbox.  See css.js for use. */
 Css.register(`
-.blocklyTreeRow:not(.blocklyTreeSelected):hover {
+.blocklyToolboxCategory:not(.blocklyTreeSelected):hover {
   background-color: rgba(255, 255, 255, .2);
 }
 
@@ -671,7 +671,7 @@ Css.register(`
   margin: 1px 0 1px 5px;
 }
 
-.blocklyTreeRow {
+.blocklyToolboxCategory {
   height: 22px;
   line-height: 22px;
   margin-bottom: 3px;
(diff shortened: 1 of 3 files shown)


core/toolbox/category.ts
BeksOmega, line 1: There are some other references `blocklyTreeRow` in this file. Could you rename them as well?

BeksOmega (changes requested):
Thanks for digging into this! Just a few fixes then should be gtg.

core/toolbox/category.ts
Abhinav-Chdhary: Yes, let me know if this change works.

BeksOmega (changes requested):
Thanks for the fixes! But now it looks like some Mocha tests are failing. Could you look into what's going wrong? You can run them locally by doing `npm run test:mocha:interactive`.

BeksOmega (changes requested):
Ah I see the issue, this is my bad. We've created a conflict between the CSS class of the container and the CSS class of the row. Could you name the original uses of `blocklyToolboxCategory` to `blocklyToolboxCategoryContainer` as well? I think that should fix the problem!

Abhinav-Chdhary:
> Ah I see the issue, this is my bad. We've created a conflict between the CSS class of the container and the CSS class of the row. Could you name the original uses of `blocklyToolboxCategory` to `blocklyToolboxCategoryContainer` as well? I think that should fix the problem!

I have tried to do so. It passes all tests locally.

BeksOmega (approved):
Sweetness! Thank you for the back and forth, this looks beautiful =)

Issue #8242, opened 2024-07-03
Surface workspace comments' IDs in the DOM


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

The workspace comments' IDs should be in the dom so that the multiselect plugin can figure out which comments are being selected.

`getCommentById` should also be made public.

### Request

Add the comment's ID to a `data-id` attribute.

changminbark:
Hello,

I am currently tasked with updating the multiselect plugin to allow for support in the selection of other draggables. I wanted to ask whether it is possible to make it so that id's for all draggable objects and corresponding workspace getById() methods are implemented in Blockly to allow for access to them. Currently, the plugin uses the workspace getById() method to access the draggable objects so that we can call methods like drag(), isMovable(), etc. If there is a different way of tackling this issue or there is a reason why these ID related methods should not be exposed to the public, please let me know.

I will try to look into this issue, but I am not sure I will have the time to. If there is anyone else who can look into this, that would be helpful!

changminbark:
Another question I had regarding the workspace comments is: Why are the classes of the renderedWorkspaceComments of the 'blocklyComment blocklyEditable' classes and not the 'blocklyDraggable' class? Should it not include the 'blocklyDraggable' class if it is a draggable object?

mark-friedman:
> I wanted to ask whether it is possible to make it so that id's for all draggable objects and corresponding workspace getById() methods are implemented in Blockly to allow for access to them.

FYI, [this comment](https://github.com/mit-cml/workspace-multiselect/pull/62#issuecomment-2209411042) has some info that I think would make this fairly simple.

BeksOmega:
> Another question I had regarding the workspace comments is: Why are the classes of the renderedWorkspaceComments of the 'blocklyComment blocklyEditable' classes and not the 'blocklyDraggable' class? Should it not include the 'blocklyDraggable' class if it is a draggable object?

I split this off into a separate issue: https://github.com/google/blockly/issues/8253

changminbark:
I want to tackle this issue. Thank you!

changminbark:
I have fixed this in this PR: https://github.com/google/blockly/pull/8358

Issue #8253, opened 2024-07-08
Workspace comments are missing the `blocklyDraggable` CSS class


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

Workspace comments should have the `blocklyDraggable` class at the top level (where `blocklyComment`, `blocklyEditable`, and `blocklySelected` live). 

### Reproduction steps

1. Right click to add a comment to the workspace.
2. Use the inspector to see that it's missing the relevant class.

changminbark:
I would also like to tackle this issue

BeksOmega:
Go for it @changminbark =) Thanks for picking this up!

changminbark:
I have fixed this in this PR: https://github.com/google/blockly/pull/8358

changminbark:
I think this issue can now be closed. Thank you!

PR #8358, opened 2024-07-14
fix: added id to rendered ws comments; made getCommentById public; ad…

changminbark:
…ded blocklyDraggable class to ws comments

## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8242 and https://github.com/google/blockly/issues/8253

### Proposed Changes

This exposes the rendered workspace comment id in the DOM. It also makes the workspace getCommentById function public. It also added the blocklyDraggable class to the DOM of rendered workspace comments.

### Reason for Changes

This will help with the support of the implementation of workspace comments in the multiselect plugin. This is mainly needed as the multiselect plugin uses dragSelect to select the elements' DOM. The id in the DOM is then passed to a respective get[element]ById function to manipulate/work with the actual object (in this case workspace comments).

### Test Coverage

I tested this by looking at the HTML of the page and seeing whether it had the added classes and data-id field.

### Documentation

Documentation regarding rendered workspace comments having a data-id field may be added, but it would not be of the utmost concern.

### DIFF
--- a/core/comments/comment_view.ts
+++ b/core/comments/comment_view.ts
@@ -104,7 +104,7 @@ export class CommentView implements IRenderedElement {
 
   constructor(private readonly workspace: WorkspaceSvg) {
     this.svgRoot = dom.createSvgElement(Svg.G, {
-      'class': 'blocklyComment blocklyEditable',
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):
Thanks for your change!

PR #8377, opened 2024-07-16
feat: Add css classes from json block definitions

### LINKED ISSUE #8271: Add a `classes` property to the JSON block definitions to add CSS classes
Dependent on #8270

## To Fix

In the [BlockSvg class](https://github.com/google/blockly/blob/17dfa462a357b553523393145c9d5045023e627b/core/block_svg.ts#L1924), Override the `jsonInit` method to take the value of the `json['classes']` property, concatenate it if it is an array, and pass it to the `this.addClass` method.

The PR to fix this should be made against the `rc/v12.0.0` branch.

To work on this issue, please comment below asking to be assigned.

ishon19:
## The basics

## The details
This PR overrides the `jsonInit` method in the `BlockSvg` class to get the css class names from the `classes` field and apply the same using the `addClass` method. 

Fixes #8271

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1720,4 +1720,20 @@ export class BlockSvg
     traverseJson(json as unknown as {[key: string]: unknown});
     return [json];
   }
+
+  override jsonInit(json: AnyDuringMigration): void {    
+    super.jsonInit(json);
+      
+    if (json['classes']) {
+      let classesToAdd = '';
+
+      if (Array.isArray(json['classes'])) {        
+        classesToAdd = json['classes'].join(' ');
+      } else {        
+        classesToAdd = json['classes'];
+      }
+      
+      this.addClass(classesToAdd);
+    }
+  }
 }
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1721,7 +1721,7 @@ export class BlockSvg
     return [json];
   }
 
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -1720,4 +1720,20 @@ export class BlockSvg
...
+    if (json['classes']) {
+      let classesToAdd = '';
+
+      if (Array.isArray(json['classes'])) {        
+        classesToAdd = json['classes'].join(' ');
+      } else {        
+        classesToAdd = json['classes'];
+      }
+      
+      this.addClass(classesToAdd);
```
BeksOmega, line 1736: I think you can simplify this using a [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator).
```suggestion
      this.addClass(Array.isArray(json['classes']) ? json['classes'].join(' ') : json['classes']);
```

You'll probably need to run `npm run format` in the root directory after applying this change to handle the line wrapping.

BeksOmega (changes requested):

core/block_svg.ts
ishon19: I agree, thanks for pointing that out! Updated that and formatted using `npm run format`.

BeksOmega (approved):
Sweetness! Thank you for the fix this looks great =) I'll get this merged once CI passes.

Issue #8077, opened 2024-05-07
Refactor variable serialization to use the variable map


BeksOmega:

gonfunko:
(Actually uses the variable model since we moved save/load there).

PR #8388, opened 2024-07-17
refactor: update the variable interfaces.

gonfunko:
## The basics

### Proposed Changes
This PR revamps the new variable interfaces in response to issues encountered during implementation. Specifically:

* `save` and `load` are moved from IVariableMap to IVariableModel. In the case where somebody wants to add additional properties to their variables, this means they need only subclass/implement IVariableModel, and not also have to replace the variable map solely to handle serialization. 
* `getWorkspace()` is added to IVariableModel; the existing VariableModel had this property, and it's often necessary to determine which workspace a variable is associated with.
* `addVariable()` is added to IVariableMap. Again, this allows for simple variables-with-extra-fields without having to subclass the VariableMap in order to change the method signature of `createVariable`  to accomodate new properties; the custom variable can simply be instantiated, and then added to the vanilla VariableMap.

In order to get things to build, this PR updates VariableModel to conform to the new interface and register itself, and updates variable serialization to use the new save and load methods it introduces rather than hardcoding serialization.

### DIFF
--- a/core/interfaces/i_variable_map.ts
+++ b/core/interfaces/i_variable_map.ts
@@ -4,8 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {IVariableModel} from './i_variable_model.js';
-import {State} from '../serialization/variables.js';
+import type {IVariableModel, IVariableState} from './i_variable_model.js';
 
 /**
  * Variable maps are container objects responsible for storing and managing the
@@ -14,7 +13,7 @@ import {State} from '../serialization/variables.js';
  * Any of these methods may define invariants about which names and types are
  * legal, and throw if they are not met.
  */
-export interface IVariableMap<T extends IVariableModel, U extends State> {
+export interface IVariableMap<T extends IVariableModel<IVariableState>> {
   /* Returns the variable corresponding to the given ID, or null if none. */
   getVariableById(id: string): T | null;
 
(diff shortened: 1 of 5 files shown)


core/interfaces/i_variable_map.ts
```diff
@@ -60,13 +62,4 @@ export interface IVariableMap<T extends IVariableModel, U extends State> {
 
   /* Removes all variables from this variable map. */
   clear(): void;
-
-  /* Returns an object representing the serialized state of the variable. */
-  saveVariable(variable: T): U;
```
rachel-fenichel, line 65: Is this deletion a breaking change?

core/variable_model.ts
```diff
@@ -95,6 +96,30 @@ export class VariableModel implements IVariableModel {
...
+    const type = this.getType();
+    if (type) {
+      state['type'] = type;
+    }
+
+    return state;
+  }
+
+  static load(state: IVariableState, workspace: Workspace) {
+    // TODO(adodson): Once VariableMap implements IVariableMap, directly
```
rachel-fenichel, line 117: If this TODO will last past this week, make it a numbered issue.

core/interfaces/i_variable_map.ts
gonfunko: No, this interface was added a couple days ago in the V12 branch, it's not yet been released.

core/variable_model.ts
gonfunko: I intend to resolve it sooner than that :)

rachel-fenichel (approved):

gonfunko:
Fixed #8077

gonfunko:
Also fixed #8078.

Issue #8417, opened 2024-07-23
Unable to trigger Toolbox's Block delete style


zfangqijun:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

When the flyout is open, dragging a Block to the Toolbox does not trigger the Toolbox's Block deletion style(blocklyToolboxDelete)

### Reproduction steps

1. Blockly.getMainWorkspace().getFlyout().setAutoClose(false)
2. Open Flyout
3. Drag Block to Toolbox

### Screenshots

- Expected, autoClose is true
![blockly1 (online-video-cutter com)](https://github.com/user-attachments/assets/52dc24e8-4314-4f0a-8e66-4f7f3a7cef00)

- Unexpected, autoClose is false
![blockly2 (online-video-cutter com)](https://github.com/user-attachments/assets/74651b59-660a-4d59-960a-a012fc0d139b)

cpcallen:
Seems at least slightly related to #8413, though possibly different cause.

johnnesky:
The method `updateDragTarget` in dragger.ts calls `getDragTarget` from workspace_svg.ts, which iterates over `dragTargetAreas` and picks the first one it finds which contains the pointer event. When the flyout is closed, the toolbox is found, but when the flyout is open, the flyout's area overlaps the toolbox area, and the flyout appears to take precedence. 

The method `getClientRect` in flyout_vertical.ts includes this comment, which explains why its drag target area extends beyond the actual flyout's bounds in 3 directions:
```
    // BIG_NUM is offscreen padding so that blocks dragged beyond the shown
    // flyout area are still deleted.  Must be larger than the largest screen
    // size, but be smaller than half Number.MAX_SAFE_INTEGER (not available on
    // IE).
    const BIG_NUM = 1000000000;
```

When dragging over these areas, the mouse cursor is correctly updated to indicate that the block will be deleted regardless of whether the toolbox or the flyout is detected as the drag target, but the toolbox's CSS class is only updated if the flyout is closed and the toolbox is detected as the target.

One possible solution would be to make sure that the toolbox takes precedence in `getDragTarget` when both the toolbox and the flyout are present. The drag target areas are derived from the drag targets found using this query:
```
    const dragTargets = this.componentManager.getComponents(
      ComponentManager.Capability.DRAG_TARGET,
      true,
    );
```
[...]

PR #8418, opened 2024-07-23
fix: no trigger toolbox block deletion style

zfangqijun:
## The basics

### Description

When the flyout is open, dragging a Block to the Toolbox does not trigger the Toolbox's Block deletion style(blocklyToolboxDelete)

### Reproduction steps

1. Blockly.getMainWorkspace().getFlyout().setAutoClose(false)
2. Open Flyout
3. Drag Block to Toolbox

### Resolves

Fixes https://github.com/google/blockly/issues/8417

### Reason for Changes

I'm not sure if this change is appropriate 🤔

The reason is that when the flyout is open, getDragTarget always returns the flyout. Maybe a better way is to add a priority field to dragTargetAreas?

### DIFF
--- a/core/flyout_vertical.ts
+++ b/core/flyout_vertical.ts
@@ -320,7 +320,7 @@ export class VerticalFlyout extends Flyout {
 
     if (this.toolboxPosition_ === toolbox.Position.LEFT) {
       const width = flyoutRect.width;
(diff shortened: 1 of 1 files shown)


zfangqijun:
better https://github.com/google/blockly/pull/8432

Issue #6895, opened 2023-03-14
Audit `@internal` tags


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

We used to have `@package` tags back when we were in Closure land. Protected/public things tagged `@package` could be overridden by external developers, but the `@package` indicated that external developers weren't supposed to /call/ these things.

Now we no longer have `@package` instead we converted it `@internal`. `@internal` things aren't documented by API Extractor, and can't be overridden. This makes tagging things `@internal` semantically different from tagging things `@package`.

We should audit our `@internal` tags. A good place to start would be removing `@internal` tags from anything that is `protected`, since `@internal` doesn't allow something to be overridden.

### Reproduction steps

1. 
2. 
3.

PR #8426, opened 2024-07-24
chore: Removed @internal annotation from public Field methods.

gonfunko:
## The basics

### Resolves
Part of #6895

### Proposed Changes
This PR removes the `@internal` annotation from various Field methods that should be accessible by subclasses.


BeksOmega (approved):
This could probably go against develop if you want? No preference here.

gonfunko:
> This could probably go against develop if you want? No preference here.

Eh I'd have to rebase, I think I'll just stick with v12 for now.

Issue #8269, opened 2024-07-10
Add the block's style as a CSS class to the block's root SVG


BeksOmega:
## To Fix

In the [BlockSvg class' `setStyle` method](https://github.com/google/blockly/blob/17dfa462a357b553523393145c9d5045023e627b/core/block_svg.ts#L1103), call `dom.addClass` on the `this.svgGroup_` to add the `blockStyleName` as a CSS class on the block. You should also call `dom.removeClass` with `this.styleName_` to remove any previous style.

The PR to fix this should be made against the `rc/v12.0.0` branch.

To work on this issue, please comment below asking to be assigned.

BeksOmega:
@ga-fleury is working on this.

Shreshthaaa:
Hey @BeksOmega, I would like to work on this issue. Kindly assign this to me.

BeksOmega:
Hiya @Shreshthaaa =) Someone is already working on this issue! (I had to comment for them because they forgot.)

If you comment on an issue that isn't assigned to anyone (most of the ones in [this list](https://github.com/google/blockly/issues?q=is%3Aissue+is%3Aopen+no%3Aassignee+label%3A%22good+first+issue%22)). I can assign you to it!

Shreshthaaa:
@BeksOmega, I've commented on a few other issues as well. Could you kindly review and assign them to me? Thank you

Aarya01Patil:
@BeksOmega Hey can i work on this if its available to solve

GarvitDadheech:
Hey @BeksOmega , Can you please assign this issue to me? I'd love to solve this issue.

Shreshthaaa:
> @Shreshthaaa Are you still interested in working on this? It's now available.

Yeah sure, I'm interested in working on this.

Shreshthaaa:
Hey @BeksOmega, I have raised the [PR](https://github.com/google/blockly/pull/8436) for this issue. Kindly review.

omwagh28:
@BeksOmega I would like to solve this issue

BeksOmega:
Closed by #8436

PR #8436, opened 2024-07-27
feat: added block's style as a CSS class to block's root SVG

Shreshthaaa:
# Details :

Added the `blockStyleName` as a CSS class on the block. Also called `dom.removeClass` with `this.styleName_` to remove any previous style.

# Fixes  : 

#8269

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1162,6 +1162,10 @@ export class BlockSvg
       .getConstants()
       .getBlockStyle(blockStyleName);
 
+    if (!this.svgGroup_) {
+      throw new Error('SVG group is not defined');
+    }
+
     if (this.styleName_) {
       dom.removeClass(this.svgGroup_, this.styleName_);
     }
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -1161,7 +1161,10 @@ export class BlockSvg
       .getRenderer()
       .getConstants()
       .getBlockStyle(blockStyleName);
-    this.styleName_ = blockStyleName;
```
BeksOmega, line 1164: Can you re-add this after your check so that the style name is still being cached? I don't think your logic quite works without this :/

BeksOmega (changes requested):
One initial fix then I'll give it a more thorough check!

Shreshthaaa:
Hey @BeksOmega, I have done the needful. Kindly review.

core/block_svg.ts
```diff
@@ -1161,7 +1161,14 @@ export class BlockSvg
...
+
+    if (!this.svgGroup_) {
+      throw new Error('SVG group is not defined');
+    }
```
BeksOmega, line 1167: Can you remove this? The `svgGroup_` is non-nullable.

BeksOmega (changes requested):
Looks great! Just one fix then should be good to go =)

core/block_svg.ts
Shreshthaaa: Done.

BeksOmega (approved):
Sweet once CI passes I'll get this merged =)

Issue #7154, opened 2022-12-13
disable-top-blocks should not enable procedure callers if their associated definitions are disabled


BeksOmega:
**Describe the bug**

Procedure callers are supposed to be disabled if their associated definitions are disabled. However, this plugin automatically enables procedure caller blocks when they are attached to a start block, even if they are supposed to be disabled.

**To Reproduce**

1. Open the playground file, and add the follow code at the bottom of the `start` function.
```js
workspace.addChangeListener(Blockly.Events.disableOrphans);
```
3. Drag two procedure definitions into the workspace.
4. Disable the second one.
5. Drag a caller for the second one into the workspace.
6. Observe how it is disabled.
7. Attach the caller to the first procedure.
8. Observe how it is incorrectly enabled (incorrectly because its definition is disabled).

![image](https://github.com/google/blockly/assets/25440652/4589d23b-8d92-4053-9b5a-3b455dd4009b)

**Expected behavior**

Procedure callers should remain disabled if their definitions are disabled.

![image](https://github.com/google/blockly/assets/25440652/2a3e5bdd-e99d-4a11-a837-f3c3b170518f)

## To Fix

1) Run through the reproduction steps and make sure you can reproduce the issue.
2) In the [disableOrphans](https://github.com/google/blockly/blob/344861698e65c3fc5f05db000c01cff2155f56ff/core/events/utils.ts#L542) listener check if the block [isProcedureBlock](https://github.com/google/blockly/blob/344861698e65c3fc5f05db000c01cff2155f56ff/core/interfaces/i_procedure_block.ts#L19). If it is, make sure that its procedure model is enabled before enabling it.
1) Run the reproduction steps again and make sure the issue doesn't reproduce.

KrautRock:
@BeksOmega I am interested in working on this

Taru-Shukla:
I would like to work on this issue

BeksOmega:
Hiya @KrautRock Taru actually messaged me on slack a few minutes before you commented, so I'm going to assign them =)

If you'd like to find a different issue I can get you assigned!

Go for it @Taru-Shukla =)

harsh-791:
@BeksOmega I would like to work on this issue

harsh-791:
@BeksOmega Please review my pr and if any changes required do tell me plese.

BeksOmega:
Looks like this was already fixed by @johnnesky

PR #8452, opened 2024-08-01
fix: disableOrphans should not enable procedure callers if their associated definitions are disabled

harsh-791:
# Issue --> https://github.com/google/blockly/issues/7154

## Description:
This pull request addresses the issue detailed in [Issue #7154](https://github.com/google/blockly/issues/7154) by updating the `disableOrphans` function to comply with the latest Blockly API changes. Specifically, it replaces the deprecated `setEnabled` method with the recommended `setDisabledReason` method. This ensures that blocks are enabled or disabled correctly according to the updated API, maintaining the intended functionality without using deprecated methods.

## Changes Made:
- Replaced all instances of `child.setEnabled(true)` with `child.setDisabledReason(false, ORPHANED_BLOCK_DISABLED_REASON)` to enable blocks.
- Replaced all instances of `block.setEnabled(false)` with `block.setDisabledReason(true, ORPHANED_BLOCK_DISABLED_REASON)` to disable blocks.
- Ensured that the logic for handling procedure blocks and their enabled states is preserved.
- Maintained the conditions for when blocks should be enabled or disabled based on their parent status and workspace state.

## Impact:
These changes ensure that the block enabling and disabling functionality adheres to the latest Blockly standards, avoiding the use of deprecated methods and preventing potential issues in future versions.

### DIFF
--- a/core/events/utils.ts
+++ b/core/events/utils.ts
@@ -20,6 +20,7 @@ import type {BlockMove} from './events_block_move.js';
 import type {CommentCreate} from './events_comment_create.js';
 import type {CommentMove} from './events_comment_move.js';
 import type {ViewportChange} from './events_viewport.js';
+import {isProcedureBlock} from '../interfaces/i_procedure_block.js';
 
 /** Group ID for new events.  Grouped events are indivisible. */
 let group = '';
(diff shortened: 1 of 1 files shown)


BeksOmega:
Actually @harsh-791 were you able to reproduce the original issue?

harsh-791:
@BeksOmega Not exactly but tried going through other PR's raised and understaning the problem in them, maybe it was due to other's raising PR's in the wrong branch

BeksOmega:
> @BeksOmega Not exactly but tried going through other PR's raised and understaning the problem in them, maybe it was due to other's raising PR's in the wrong branch

hmm ok I think this might already be fixed actually. I will test in a bit.

BeksOmega:
Yeah sorry about that @harsh-791 it looks like @johnnesky has already fixed this issue. Thank you for taking a look at it though!

Issue #7428, opened 2023-08-24
Zelos hides disabled blocks placeholders


clementcontet:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When a block is disabled, Zelos doesn't show the placeholders (not sure about the correct name...), so it is very hard to connect disabled blocks together.

Strangely, the placeholders reappear when we open the toolbox.

Geras doesn't have this problem: placeholders are always visible.

NB: overriding style fixes the issue:
```
.blocklyDisabled > .blocklyOutlinePath {
  fill: darkgray !important;
}
```

### Screenshots

https://github.com/google/blockly/assets/7261426/ea55f3b6-9db5-4d2c-8706-d8e42858d463

### Browsers

Chrome desktop

surajguduru:
Hey @BeksOmega I would to work on this issue. Can you assign it to me ?

BeksOmega:
@surajguduru This isn't a good-first-issue which means it will require doing some investigation and learning the internals  of Blockly. Are you still interested in working on it?

surajguduru:
Yes. I would like to give it a shot.

BeksOmega:
Ok go for it @surajguduru

gonfunko:
Fixed by #8671.

PR #8468, opened 2024-08-06
Issue#7428

surajguduru:
## The basics

### Resolves

Fixes [#7428](https://github.com/google/blockly/issues/7428) - Zelos hides disabled blocks placeholders.

### Proposed Changes

This pull request addresses an issue where placeholders for disabled blocks were not visible in the Zelos renderer. The changes involve modifying the CSS to ensure that the placeholders are displayed consistently, similar to the behavior in the Geras renderer.

### Reason for Changes

The issue made it difficult for users to connect disabled blocks together, as the placeholders were not visible. This fix improves the usability of the Zelos renderer by ensuring that placeholders are always visible, even when blocks are disabled.

### Test Coverage

The changes were manually tested by:
- Disabling blocks in the Zelos renderer and confirming that the placeholders are visible.
- Comparing the behavior with the Geras renderer to ensure consistency.
- Testing across different browsers to confirm that the changes work as expected.

### Documentation

No additional documentation is needed, as this change only affects the internal CSS styling.

### Additional Information

None.

### DIFF
--- a/core/css.ts
+++ b/core/css.ts
@@ -242,6 +242,10 @@ let content = `
   display: none;
 }
 
+.blocklyDisabled > .blocklyOutlinePath {
+  fill: darkgray !important;
+}
+
 .blocklyInsertionMarker>.blocklyPath,
 .blocklyInsertionMarker>.blocklyPathLight,
 .blocklyInsertionMarker>.blocklyPathDark {
--- a/core/field_label.ts
+++ b/core/field_label.ts
@@ -74,6 +74,9 @@ export class FieldLabel extends Field<string> {
     if (this.class) {
       dom.addClass(this.getTextElement(), this.class);
     }
+    if (this.fieldGroup_) {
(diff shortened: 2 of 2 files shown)


core/css.ts
```diff
@@ -242,6 +242,10 @@ let content = `
   display: none;
 }
 
+.blocklyDisabled > .blocklyOutlinePath {
+  fill: darkgray !important;
+}
```
BeksOmega, line 247: I don't think you want to add this here. I think you want to modify the code in the zelos constants file `getCSS_`. You should be able to remove the `!important` declaration then.

And could you try just setting the stroke of the outline path instead of setting the fill? I think that might look nicer!

BeksOmega (changes requested):
Could you also perform a `git rebase` to remove the extra `field_lable.ts` commit from this branch?

surajguduru:
I will look into it.

PR #8533, opened 2024-08-16
fix: Detect if deleting shadow block affects selection highlight

johnnesky:
## The basics

### Resolves

Partially addresses: https://github.com/google/blockly-samples/issues/2426

### Proposed Changes

Detects whether deleting a block would delete the currently selected block, and if so, deselects it *before* severing connections.

### Reason for Changes

When a shadow block is selected, an ancestor of that block receives a visual highlight. If the shadow block is then deleted (e.g. because it was replaced with a real block via the shadow block converter plugin), then the highlighted block no longer has any relationship to any selected block. That highlight should go away, but currently it persists and may continue to persist while making further changes to the selection.

Note that `Block.dispose()` severs a connection, and `Block.disposeInternal()` also severs connections, so we need to determine whether any parent block needs to be updated before this happens. The subclass `BlockSvg` overrides these methods, so I chose to perform the detection and resolution in `BlockSvg.dispose()`.

### Test Coverage

Existing tests pass. I added new  tests for the new features.

### Documentation

N/A

### Additional Information

I think it feels unintuitive that, when using the shadow block converter plugin and the user edits a shadow block, the current selection gets cleared. As far as the user is concerned, nothing appears to have been deleted (since the real clone of the shadow block seamlessly replaces the shadow block) so it's weird that editing a field of a shadow block deselects the selection.

I propose that we fix this in the shadow block converter plugin by transferring the selection to the new real clone of the selected shadow block after replacing it. Since that's in the samples repo, it would have to be a separate PR. In the meantime, the shadow block converter plugin is already broken and I don't think this PR makes it worse.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -319,9 +319,18 @@ export class Block implements IASTNodeLocation {
     }
   }
 
-  descendsFrom(other: Block): boolean {
-    if (this === other) return true;
-    return this.getParent()?.descendsFrom(other) ?? false;
+  /**
+   * Returns true if this block is the same block as the given block or contains
+   * the provided block as a descendent.
+   *
+   * @param other the other block to compare to.
+   */
+  contains(other: Block | null): boolean {
+    while (other !== null) {
+      if (other === this) return true;
+      other = other.parentBlock_;
+    }
+    return false;
   }
 
   /**
--- a/core/block.ts
+++ b/core/block.ts
@@ -319,6 +319,11 @@ export class Block implements IASTNodeLocation {
     }
   }
 
+  descendsFrom(other: Block): boolean {
+    if (this === other) return true;
+    return this.getParent()?.descendsFrom(other) ?? false;
+  }
+
   /**
    * Dispose of this block.
    *
--- a/core/block.ts
+++ b/core/block.ts
@@ -320,20 +320,6 @@ export class Block implements IASTNodeLocation {
   }
 
   /**
-   * Returns true if this block is the same block as the given block or contains
-   * the provided block as a descendent.
-   *
-   * @param other the other block to compare to.
-   */
-  contains(other: Block | null): boolean {
-    while (other !== null) {
-      if (other === this) return true;
-      other = other.parentBlock_;
-    }
-    return false;
-  }
-
-  /**
    * Dispose of this block.
    *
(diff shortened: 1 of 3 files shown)


core/block.ts
```diff
@@ -319,6 +319,20 @@ export class Block implements IASTNodeLocation {
     }
   }
 
+  /**
+   * Returns true if this block is the same block as the given block or contains
+   * the provided block as a descendent.
```
BeksOmega, line 324: Can you comment here that this includes all next blocks? People are easily confused about descendants vs children.

BeksOmega (approved):
Looks lovely, just one docs addition =)

Thank you for the beautiful tests!

core/block.ts
johnnesky: Actually, that reminds me: the intent of this PR is to detect whether the selected block would get deleted if another block is deleted, but the next blocks of the originally deleted block do NOT get deleted (and the associated stack gets "healed"). I want to include the next blocks of inputs, but not the next blocks in the same stack as the deleted block. So this implementation probably isn't exactly right?

BeksOmega: Hmm my instinct is that it's not actually possible to run into that case since this is specifically a problem for shadow blocks being selected while highlighting another block, and you can't have non-shadows attached to shadows, so no healing would occur.

Basically the healing case is incompatible with the shadow selection case, iiuc. But it's a bit complex so I'm not 100%

johnnesky: On the other hand, if the deleted block comes between the selected shadow block and its highlighted ancestor, then deleting the block could sever the connection even if the selected block itself did not get deleted in the process. In which case, technically I think we should remove the highlight, delete the block and its inputs, then re-highlight an appropriate ancestor of the selected shadow block based on the new stack arrangement?

BeksOmega: Is it possible for the deleted block to come between the selected shadow block and its ancestor and the selected shadow block not get deleted?

johnnesky: Yeah, maybe I'm overthinking it.

However, I decided to remove the new `contains` function that I added, and manually inlined it into the one place I was using it. I figured it wasn't worth adding a new public function to the API that may be confusing and would need to be maintained indefinitely.

Issue #8327, opened 2024-07-11
Remove the `blocklyMenuItemHighlight` CSS class and use the `hover` CSS selector instead


BeksOmega:
1. Remove the `blocklyMenuItemHighlight` CSS class from the [menu item](https://github.com/google/blockly/blob/f8025a1bdd12025ecb8830159509148ff27ad3c8/core/menuitem.ts#L73).
2. Use the [`:hover` CSS pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) in [`css.ts`](https://github.com/google/blockly/blob/f8025a1bdd12025ecb8830159509148ff27ad3c8/core/css.ts#L481).

This change should be made against the `rc/v12.0.0` branch. This should be marked as a breaking change.

To work on this issue, comment below asking to be assigned.

Arun-cn:
I like to work on this issue

BeksOmega:
Go for it @Arun-cn =) Let me know if you have any questions!

Arun-cn:
Please inform me if any changes are required. Additionally, provide more context. Answer the question of why that change is necessary.

gonfunko:
Fixed by #8536.

PR #8536, opened 2024-08-19
fix!: Remove the blocklyMenuItemHighlight CSS class and use the hover

Arun-cn:
## The basics

### Resolves

Fixes Remove blocklyMenuItemHighlight CSS class and use :hover CSS selector #8327 

### Proposed Changes
Remove the blocklyMenuItemHighlight CSS class from the [menu item](https://github.com/google/blockly/blob/f8025a1bdd12025ecb8830159509148ff27ad3c8/core/menuitem.ts#L73).
Use the [:hover CSS pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover) in [css.ts](https://github.com/google/blockly/blob/f8025a1bdd12025ecb8830159509148ff27ad3c8/core/css.ts#L481).

### DIFF
--- a/core/css.ts
+++ b/core/css.ts
@@ -445,6 +445,10 @@ input[type=number] {
   cursor: inherit;
 }
 
+.blocklyMenuItem:hover {
+  background-color: rgba(0,0,0,.1);
+}
+
 /* State: hover. */
 .blocklyMenuItemHighlight {
   background-color: rgba(0,0,0,.1);
--- a/core/css.ts
+++ b/core/css.ts
@@ -449,11 +449,6 @@ input[type=number] {
   background-color: rgba(0,0,0,.1);
 }
 
-/* State: hover. */
-.blocklyMenuItemHighlight {
-  background-color: rgba(0,0,0,.1);
-}
-
 /* State: selected/checked. */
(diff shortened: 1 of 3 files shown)


core/menuitem.ts
```diff
@@ -185,15 +183,6 @@ export class MenuItem {
    */
   setHighlighted(highlight: boolean) {
```
BeksOmega, line 184: Can you actually delete this method entirely? It's internal and `this.highlight` is private, so removing it isn't a breaking change =)

BeksOmega (changes requested):
This looks much better! Just one fix then should be gtg =) Thank you for your continued work on this!

core/css.ts
```diff
@@ -445,6 +445,10 @@ input[type=number] {
   cursor: inherit;
 }
 
+.blocklyMenuItem:hover {
+  background-color: rgba(0,0,0,.1);
+}
+
 /* State: hover. */
```
BeksOmega, line 452: Sorry I missed this, could you delete this CSS as well?

BeksOmega (approved):
One last fix then LGTM!

BeksOmega:
Thank you again for all of your work on this!

Issue #8225, opened 2024-06-20
Issue with mirroring workspace


QuirkyCort:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

When all mirroring events from one workspace to another, I expect both workspaces to be identical. But when an "else if" is added to an "if" block, it is possible for the two workspace to differ.

The problem can be reproduced on https://google.github.io/blockly-samples/examples/mirror-demo/, and I have also verified it on my own application where I can confirmed that all events are being mirrored.

### Reproduction steps

1. Open https://google.github.io/blockly-samples/examples/mirror-demo/
2. Add an "if" block.
3. Add an "else" to the "if" block.
4. Add a "print" under the "else".
5. Add an "else if" to the "if" block.
6. The "print" remains under the "else" in the first workspace, but is disconnected in the second workspace.

maribethb:
fwiw this occurs in both v11.0.0 and v11.1.1 so it's not related to the disappearing input blocks.

cpcallen:
See also #7950.

cpcallen:
## Root cause analysis (part I)

Initially I knew the following (true) facts:

* When a `BlockChange` change is replayed in the secondary workspace, [`BlockChange.prototype.run`](https://github.com/google/blockly/blob/9a0619aa2a94d0acdccbac55b0d765f9b54e15b4/core/events/events_block_change.ts#L163) handles replaying `mutation`-element events by calling `loadExtraState` on the block.
* The implementation `CONTROLS_IF_MUTATOR_MIXIN.loadExtraState` calls `this.updateShape_()`.
* `CONTROLS_IF_MUTATOR_MIXIN.updateShape_` does not saving and restoring connections, unlike its sibling `.rebuildShape_`.

It therefore seemed logical that the solution would be to have `loadExtraState` instead call `this.rebuildShape_()` instead.

[...]

PR #8538, opened 2024-08-20
refactor(events): Introduce and use event type predicates

cpcallen:
## The basics

### Proposed Changes

* Introduce predicates for testing `Abstract` event subclasses based on their `.type` properties.  These are useful because there are places where it is not possible to use `instanceof <ClassConstructor>` tests for type narrowing due to load ordering issues that would be caused by the need to `import` (rather than just `import type`) the class
constructors in question.

* Simplify several sections of code by using these type predicates for type narrowing, and thereby avoiding the need for explicit casts.

### Reason for Changes

Cleanup and preparation for fixing #8225 et al.

### Test Coverage

* Passes `npm test`.
* Trashcan behaviour has been tested manually to ensure refactor has not broken anything.

### Additional Information

The type predicates are not currently reexported from `core/blockly.ts`.  This could be done by adding suitable `export … from …` statements to `core/events/events.ts`, but I've opted not to do this until there is a good reason to: in most cases it is probably preferable that code use `e instanceof EventClass` rather than `isEventClass(e)`, since the former is a stricter and less easily subverted type check, and code outside of `core/` can always access the event constructors without any cyclic `import` issues.

### DIFF
--- a/core/bump_objects.ts
+++ b/core/bump_objects.ts
@@ -13,7 +13,7 @@ import type {BlockMove} from './events/events_block_move.js';
 import type {CommentCreate} from './events/events_comment_create.js';
 import type {CommentMove} from './events/events_comment_move.js';
 import type {CommentResize} from './events/events_comment_resize.js';
-import type {ViewportChange} from './events/events_viewport.js';
+import {isViewportChange} from './events/predicates.js';
 import {BUMP_EVENTS, EventType} from './events/type.js';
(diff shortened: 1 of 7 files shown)


cpcallen:
> lgtm once you fix lint and formatting

Yeah: since I set up auto-sorting of imports, and this by default removes unused imports, any remaining unused imports now trigger both lint _and_ format failures.

PR #8539, opened 2024-08-21
fix(events): Simplify `filter` function, add new `enqueueEvent` function

### LINKED ISSUE #2037: controls_if: Mutation actions do not undo correctly

cpcallen:
## The basics

### Resolves

Fixes #2037, #8225

### Proposed Changes

* Remove the broken `BlockChange` event reordering code from `filter`.  Instead, do necessary event reordering in a new (and hopefully more robust and less buggy) `enqueueEvent` function called from `fireInternal` as events are fired.
* Since (as of #8537) we don't filter in undo any more, and filtering in reverse order is confusing and can cause problems, prepare to remove this opportunity for error by deprecating calling `filter` with `forward=false`.
* Simplify `filter` by having it consider only adjacent events in the queue for merging.  Previously any events in the queue could potentially be merged if they were of suitable (typically identical) `.type`, even if other events had occurred between them (with the sole exception of `BlockMove` events, which would only be merged with adjacent ones).  This could potentially result in replay failures during undo/redo/mirroring, though it is unknown whether any such problems occurred in practice.
* Use `for…of` loop where appropriate.
* Make filter reason-merging code more concise.
* Use arrow functions when calling `Array.prototype.filter`.

### Reason for Changes

* Remove code from `filter` that has been the source of multiple current and previous bugs, and actively breaks other attempts to fix those bugs (e.g. by emitting events in the correct order to start with).
* Simplify `filter` so it is easier to reason about its behaviour.

### Test Coverage

* Added unit tests for `enqueueEvent`.
* There are existing unit tests for `filter`, none of which fail due to the changes made in this PR.

### DIFF
--- a/core/events/utils.ts
+++ b/core/events/utils.ts
@@ -119,7 +119,7 @@ function fireInternal(event: Abstract) {
       setTimeout(fireNow, 0);
     }
   }
-  FIRE_QUEUE.push(event);
+  enqueueEvent(event);
 }
 
 /** Dispatch all queued events. */
@@ -135,6 +135,46 @@ function fireNow() {
       eventWorkspace.fireChangeListener(event);
     }
   }
+}
+
+/**
+ * Enqueue an event on FIRE_QUEUE.
+ *
+ * Normally this is equivalent to FIRE_QUEUE.push(event), but if the
+ * enqueued event is a BlockChange event and the most recent event(s)
+ * on the queue are BlockMove events that (re)connect other blocks to
+ * the changed block (and belong to the same event group) then the
+ * enqueued event will be enqueued before those events rather than
+ * after.
+ *
+ * This is a workaround for a problem caused by the fact that
+ * MutatorIcon.prototype.recomposeSourceBlock can only fire a
+ * BlockChange event after the mutating block's compose method
+ * returns, meaning that if the compose method reconnects child blocks
+ * the corresponding BlockMove events are emitted _before_ the
+ * BlockChange event, causing issues with undo, mirroring, etc.; see
+ * https://github.com/google/blockly/issues/8225#issuecomment-2195751783
+ * (and following) for details.
+ */
+function enqueueEvent(event: Abstract) {
+  if (isBlockChange(event) && event.element === 'mutation') {
+    let i;
+    for (i = FIRE_QUEUE.length; i > 0; i--) {
+      const otherEvent = FIRE_QUEUE[i - 1];
+      if (
+        otherEvent.group !== event.group ||
+        otherEvent.workspaceId !== event.workspaceId ||
+        !isBlockMove(otherEvent) ||
+        otherEvent.newParentId !== event.blockId
+      ) {
+        break;
+      }
+    }
+    FIRE_QUEUE.splice(i, 0, event);
+    return;
+  }
+
+  FIRE_QUEUE.push(event);
 }
 
 /**
@@ -177,6 +217,12 @@ function fireNow() {
  * but added considerable additional complexity and made it difficlut
  * to reason about how events are processed for undo/redo, so both the
  * call from undo and the post-processing code was removed.
+ *
+ * At the same time, the buggy code to reorder BlockChange events was
+ * replaced by a less-buggy version of the same functionality in a new
+ * function, enqueueEvent, called from fireInternal, thus assuring
+ * that events will be in the correct order at the time filter is
+ * called.
  *
  * @param queueIn Array of events.
  * @param forward True if forward (redo), false if backward (undo).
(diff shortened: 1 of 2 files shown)


cpcallen:
> ## The basics
> * [x]  I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)
> 
> ## The details
> ### Resolves
> Fixes #2037, #8225
> 
> ### Proposed Changes
> * Remove the broken `BlockChange` event reordering code from `filter`.  Instead, do necessary event reordering in a new (and hopefully more robust and less buggy) `enqueueEvent` function called from `fireInternal` as events are fired.
> * Since (as of [refactor(events): Don't filter events before undo #8537](https://github.com/google/blockly/pull/8537)) we don't filter in undo any more, and filtering in reverse order is confusing and can cause problems, prepare to remove this opportunity for error by deprecating calling `filter` with `forward=false`.
> * Simplify `filter` by having it consider only adjacent events in the queue for merging.  Previously any events in the queue could potentially be merged if they were of suitable (typically identical) `.type`, even if other events had occurred between them (with the sole exception of `BlockMove` events, which would only be merged with adjacent ones).  This could potentially result in replay failures during undo/redo/mirroring, though it is unknown whether any such problems occurred in practice.
> * Use `for…of` loop where appropriate.
> * Make filter reason-merging code more concise.
> * Use arrow functions when calling `Array.prototype.filter`.
> 
> ### Reason for Changes
> * Remove code from `filter` that has been the source of multiple current and previous bugs, and actively breaks other attempts to fix those bugs (e.g. by emitting events in the correct order to start with).
> * Simplify `filter` so it is easier to reason about its behaviour.
> 
> ### Test Coverage
> * Added unit tests for `enqueueEvent`.
> * There are existing unit tests for `filter`, none of which fail due to the changes made in this PR.

### Additional Info

Strictly speaking, making the `forward` parameter of `filter` default to `true` is a breaking change, as JS users would previously have observed it defaulting to falsy `undefined`.  My reasoning to overlook this is that I suspect very few developers use this function directly (to the point that I think we should consider deprecating its export), and the likelihood of any of them calling it with no second parameter, _intending that to request filtering in reverse order_, seems very remote.

Nevertheless the change adding a default could be removed if desired.

core/events/utils.ts
```diff
@@ -151,18 +162,21 @@ function fireNow() {
...
- * made it difficlut to reason about how events are processed for
- * undo/redo, so both the call from undo and the post-processing code
- * was later removed.
+ * but the mutation of events by this function was the cause of issue
+ * #7026 (note that events would combine differently in reverse order
+ * vs. forward order).  The originally-chosen fix for this was the
+ * addition (in PR #7069) of code to fireNow to post-filter the
+ * .undoStack_ and .redoStack_ of any workspace that had just been
+ * involved in dispatching events; this apparently resolved the issue
+ * but added considerable additional complexity and made it difficlut
```
gonfunko, line 217: nit: difficult

core/events/utils.ts
```diff
@@ -79,9 +79,20 @@ export type BumpEvent =
...
 
 /**
- * Create a custom event and fire it.
+ * Enqueue an event to be dispatched to change listeners.
  *
- * @param event Custom data for event.
+ * Notes:
+ *
+ * - Events are enqueued until a timeout, generally after rendering is
+ *   complete but or at the end of the current microtask, if not
```
gonfunko, line 87: nit: errant "but"

tests/mocha/event_test.js
```diff
@@ -1102,6 +1122,165 @@ suite('Events', function () {
...
+      event.newCoordinate = new Blockly.utils.Coordinate(0, 0);
+      return event;
+    }
+
+    function newConnectEvent(parent, child, inputName, workspaceId) {
+      const event = new Blockly.Events.BlockMove(child);
+      event.workspaceId = workspaceId;
+      event.oldParentId = undefined;
+      event.oldInputName = undefined;
+      event.oleCoordinate = new Blockly.utils.Coordinate(0, 0);
```
gonfunko, line 1145: nit: oldCoordinate

core/events/utils.ts
```diff
@@ -226,14 +229,17 @@ function enqueueEvent(event: Abstract) {
  *
  * @param queueIn Array of events.
  * @param forward True if forward (redo), false if backward (undo).
+ *     This parameter is deprecated: true is now the default and
+ *     calling filter with it false will in future not be supported.
```
gonfunko, line 237: nit: "with it set to"?

gonfunko (approved):

tests/mocha/event_test.js
cpcallen: I am concerned that this did not cause a test failure…

PR #8550, opened 2024-08-21
fix: Ensure immovable blocks are considered during workspace tidying

### LINKED ISSUE #6889: WorkspaceSvg.cleanUp() method doesn't account for immovable blocks
### Check for duplicates

### Description

If blocks are immovable when `cleanUp()` is called, they are skipped. Other blocks are moved into a column but ignoring any immovable blocks that might already be there: https://github.com/google/blockly/blob/4125fd31534dd9f450219259b9ccb3c1495605c9/core/workspace_svg.ts#L1812-L1830

This creates a situation where `cleanUp()` can actually make things look worse, covering up immovable blocks with movable ones.

### Reproduction steps

1. Add several blocks to the workspace.
2. Set one or more blocks to be immovable (e.g. `Blockly.selected.setMovable(false)`)
3. Select "Clean up blocks" from the workspace context menu.

### Stack trace

_No response_

### Screenshots

https://user-images.githubusercontent.com/43474485/224034976-c49bc4b7-3fba-44bd-b849-074abbc4fff6.mp4

image

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #6889

### Proposed Changes

This PR introduces a fix for #6889 (``WorkspaceSvg.cleanUp``) when a workspace contains an immovable block. Previously, immovable blocks were ignored in ``cleanUp`` which could lead to situations where movable blocks were stacked on top of the immovable blocks. The updated algorithm addresses this.

The old algorithm was essentially the following steps:
- Sort top-level blocks first by their y value then by their x (per [``getTopBlocks``](https://github.com/google/blockly/blob/9445adf8feb1f8431436c9ec61414df201b772a9/core/workspace.ts#L212) and ``Workspace``'s [sorting logic](https://github.com/google/blockly/blob/9445adf8feb1f8431436c9ec61414df201b772a9/core/workspace.ts#L170)).
- Position the first movable block at ``(0, 0)``.
- For each subsequent block:
  - If it's immovable, skip it.
  - If it's movable, position it at the previous block's ``y`` coordinate plus its height plus a minimum height spacing defined by the renderer's ``MIN_BLOCK_HEIGHT`` property. Also, left-align the block at an ``x`` value of ``0``.

The updated algorithm augments this by considering any immovable blocks that the new block's position might intersect with (by using ``Rect``'s [``intersects``](https://github.com/google/blockly/blob/9445adf8feb1f8431436c9ec61414df201b772a9/core/utils/rect.ts#L61) method), as so:
- Sort the top-level blocks in the same way before.
- Split the list of blocks into movable and immovable blocks.
- For each immovable block, compute its bounds ``Rect`` (to be used later).
- For each movable block, in order, compute its new position:
  - If it's the first block, assume a position of ``(0, 0)``. If not, assume it to be ``(0, y)`` where ``y`` is the previous block's position plus the previous block's height plus the minimum block spacing (as described above).
  - Enumerate all immovable block bounds and verify whether the block's new position would intersect with any of them. If it would, recompute ``y`` in the same way as the previous step except use the conflicting immovable block's position and size, rather than the previous movable block.
  - Repeat the last step until the new position no longer conflicts with any immovable blocks.

**NOTE**:

~This PR is also renaming ``WorkspaceSvg.cleanUp`` to ``WorkspaceSvg.tidyUp`` since 'clean up' is an overloaded term in the codebase and generally means "deinitialize." This PR does _not_ rename the context menu item since cleaning up the workspace has a different context for the user and probably makes equal sense to 'tidy' (so there wasn't an obvious reason to change it other than to keep the two aligned).~

This has been reverted after a discussion with the team--the preference is to avoid unnecessary code changes and this rename has few benefits.

### Reason for Changes

Trying to tidy up the workspace without considering immovable blocks leads to undesirable results. See previous behavior:

[Screen recording 2024-08-21 1.42.28 PM.webm](https://github.com/user-attachments/assets/70d3a70c-f695-4fe3-9990-b185db0048b1)

With the changes, the new behavior is:

[Screen recording 2024-08-21 1.40.37 PM.webm](https://github.com/user-attachments/assets/21467d29-deb4-450f-8ae0-747204f5f900)

This is a much better user experience. Note that the above was demoed by importing the following JSON into a local Blockly playground:

```json
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "math_number",
        "id": "block1",
        "x": 1,
        "y": 2,
        "fields": {
          "NUM": 1
        }
      },
      {
        "type": "math_number",
        "id": "block2",
        "x": 10,
        "y": 20,
        "movable": false,
        "fields": {
          "NUM": 2
        }
      },
      {
        "type": "math_number",
        "id": "block3",
        "x": 2,
        "y": 30,
        "fields": {
          "NUM": 3
        }
      },
      {
        "type": "controls_repeat_ext",
        "id": "block4",
        "x": 3,
        "y": 40,
        "inputs": {
          "TIMES": {
            "shadow": {
              "type": "math_number",
              "fields": {
                "NUM": 10
              }
            }
          },
          "DO": {
            "block": {
              "type": "controls_if",
              "inputs": {
                "IF0": {
                  "block": {
                    "type": "logic_boolean",
                    "fields": {
                      "BOOL": "TRUE"
                    }
                  }
                },
                "DO0": {
                  "block": {
                    "type": "text_print",
                    "inputs": {
                      "TEXT": {
                        "shadow": {
                          "type": "text",
                          "fields": {
                            "TEXT": "abc"
                          }
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      },
      {
        "type": "math_number",
        "id": "block5",
        "x": 20,
        "y": 200,
        "movable": false,
        "fields": {
          "NUM": 5
        }
      }
    ]
  }
}
```

### Test Coverage

``cleanUp`` had no functional tests, so a bunch were added (including for verifying existing behaviors). All of this function's new tests passed without the algorithmic changes except for the test ``"multiple block types immovable blocks are not moved"`` (which was verified to fail, as expected, without the algorithmic fixes).

Separately, ``Rect`` had some additional functionality and documentation added, along with an entire new test suite meant to thoroughly test all aspects of the class (since it plays a pivotal role in the new functionality, plus a bunch of other existing functionality throughout Blockly core).

### Documentation

It doesn't seem any documentation changes are needed at this time since the fix was entirely infrastructural and implementation-specific. Since this is a public API function, it's certainly possible some documentation work will be needed.

### Additional Information

The needs of ``cleanUp`` functional changes and its tests led to needing a bounding box class. This was added before ``Rect`` was discovered, and its intersection algorithm was independently derived. ``Rect``'s own intersection code is actually replaced in this PR with the derived version (the same logic, just inverted) with some documentation to logically explain why it works to help any future readers who come across it.

Separately, the workspace tests prefer to use JSON for block initialization vs. direct API access. This seems to be the recommended way to approach such test condition arrangement after discussions with other Blockly team members.

### DIFF
--- a/tests/mocha/rect_test.js
+++ b/tests/mocha/rect_test.js
@@ -119,7 +119,7 @@ suite('Rect', function () {
       assert.isTrue(areEqual, 'an instance and its clone should be equal');
     });
 
-    test('object should equal its clone', function () {
+    test('object should equal an exact explicit copy', function () {
       const rect1 = Blockly.utils.Rect.createFromPoint(
         this.createCoord(1, 2),
         23,
--- a/tests/mocha/rect_test.js
+++ b/tests/mocha/rect_test.js
@@ -13,7 +13,7 @@ import {
 suite('Rect', function () {
   setup(function () {
     sharedTestSetup.call(this);
-    this.createCoord = function(x, y) {
+    this.createCoord = function (x, y) {
       return new Blockly.utils.Coordinate(x, y);
     };
   });
@@ -22,7 +22,7 @@ suite('Rect', function () {
   });
 
   suite('Rect()', function () {
-    test('initializes properties correctly', function() {
+    test('initializes properties correctly', function () {
       const rect = new Blockly.utils.Rect(1, 2, 3, 4);
 
       assert.equal(rect.top, 1, 'top should be initialized');
@@ -33,8 +33,12 @@ suite('Rect', function () {
   });
 
   suite('createFromPoint()', function () {
-    test('initializes properties correctly', function() {
-      const rect = Blockly.utils.Rect.createFromPoint(this.createCoord(1, 2), 23, 45);
+    test('initializes properties correctly', function () {
+      const rect = Blockly.utils.Rect.createFromPoint(
+        this.createCoord(1, 2),
+        23,
+        45,
+      );
 
       assert.equal(rect.top, 2, 'top should be initialized');
       assert.equal(rect.bottom, 47, 'bottom should be initialized');
@@ -44,8 +48,12 @@ suite('Rect', function () {
   });
 
   suite('clone()', function () {
-    test('copies properties correctly', function() {
-      const rect = Blockly.utils.Rect.createFromPoint(this.createCoord(1, 2), 23, 45);
+    test('copies properties correctly', function () {
+      const rect = Blockly.utils.Rect.createFromPoint(
+        this.createCoord(1, 2),
+        23,
+        45,
+      );
 
       const clonedRect = rect.clone();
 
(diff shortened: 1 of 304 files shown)


tests/mocha/rect_test.js
```diff
@@ -0,0 +1,1668 @@
...
+        23,
+        45,
+      );
+
+      const areEqual = Blockly.utils.Rect.equals(rect, rect.clone());
+
+      assert.isTrue(areEqual, 'an instance and its clone should be equal');
+    });
+
+    test('object should equal its clone', function () {
```
BenHenning, line 122: Needs a different test name.

BenHenning (commented):
Self-reviewed changes.

BenHenning (commented):
Self-reviewed test rename.

BenHenning:
Looks like this is now ready for review.

gonfunko (approved):

BenHenning:
NB: Comments should be limited to 80 characters (ES lint doesn't catch this).

BenHenning (commented):
Self-reviewed changes renaming back to 'cleanUp'.

BenHenning:
@gonfunko could you PTAL per latest changes and let me know if the PR still looks good to you?

gonfunko (approved):

BenHenning:
Thanks @gonfunko!

mikeharv:
Thanks so much for this!

Issue #7926, opened 2024-03-11
Move logic from `startDrag` and `moveDrag` into a utils file


BeksOmega:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

The `startDrag` and `moveDrag` methods help you track the state of a drag (e.g. the total delta). Currently they exist on the workspace, and record state on the workspace, even though these methods are only used for resizing comments (block and workspace). E.g. block drags and workspace drags use other methods.

### Request

Pull the logic from these methods into a utils file. Unify with other code for recording drags if possible and useful. Have these methods defer to the utils file (since they are public and deleting them is a breaking change).

### Alternatives considered

Leave as is because it's not a big deal.

Apocalypse96:
Hey @BeksOmega I would love to solve this issue

BeksOmega:
Go for it @Apocalypse96 This should be merged into `develop` and it shouldn't include any breaking changes.

PR #8566, opened 2024-09-07
fix: Factor out workspace drag methods into utils.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/7926

### Proposed Changes

This PR transfers the logic from `WorkspaceSvg` methods `startDrag` and `moveDrag` into a new utils file. It further factors out some shared coordinate math into a private function. It removes the private property `dragDeltaXY` from `WorkspaceSvg` (is this a breaking change?) and replaces it with an internal map from workspace to delta.

### Reason for Changes

This feature is used by block comments and workspace comments, but is not generally used by workspace objects and thus is not really appropriate for `WorkspaceSvg`.

### Test Coverage

Manually dragging and resizing comments seems to work as expected. I don't believe any existing tests directly call these methods on a workspace, maybe some do indirectly, but FWIW all existing tests pass.

### Documentation

N/A

### Additional Information

Is removing a private property considered a breaking change?

### DIFF
--- a/core/utils/drag.ts
+++ b/core/utils/drag.ts
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as browserEvents from '../browser_events.js';
+import type {WorkspaceSvg} from '../workspace_svg.js';
+import {Coordinate} from './coordinate.js';
+
+const workspaceToDragDelta: WeakMap<WorkspaceSvg, Coordinate> = new WeakMap();
+
+/**
+ * Convert from mouse coordinates to workspace coordinates.
+ *
+ * @param workspace The workspace where the pointer event is occurring.
+ * @param e The pointer event with the source coordinates.
+ */
+function mouseToWorkspacePoint(
+  workspace: WorkspaceSvg,
+  e: PointerEvent,
+): SVGPoint {
+  const point = browserEvents.mouseToSvg(
+    e,
+    workspace.getParentSvg(),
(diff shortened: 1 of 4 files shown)


core/utils/drag.ts
```diff
@@ -0,0 +1,74 @@
...
+  e: PointerEvent,
+): SVGPoint {
+  const point = browserEvents.mouseToSvg(
+    e,
+    workspace.getParentSvg(),
+    workspace.getInverseScreenCTM(),
+  );
+  // Fix scale of mouse event.
+  point.x /= workspace.scale;
+  point.y /= workspace.scale;
```
cpcallen, line 30: Since I don't really know what a `ScreenCTM` is the following question is probably nonsense, but: can the post-`mouseToSvg` fixup be avoided by passing a better value as the third parameter of `mouseToSvg`?

And if so would such a change be beneficial?

(Side note: looking at the definitions of `mouseToSvg` and `getInverseScreenCTM`, it appears that it mightalternatively be possible to completely omit the third parameter to `mouseToSvg` with no change to the returned valuest—but perhaps that would be less efficient, since `getInverseScreenCTM` seems to do caching.)

cpcallen (approved):

core/utils/drag.ts
johnnesky: I think, in some sense, the "correct" thing to do would be to pass workspace.getCanvas() as the second argument, to derive the transform matrix from the block canvas which should incorporate the zoom. However, the SVG APIs that mouseToSvg uses require an SVGSVGElement (`<svg>`), whereas the block canvas is a `<g>` element.

PR #8571, opened 2024-09-11
fix: improve flyout performance

gonfunko:
## The basics

## The details
This PR improves performance of flyouts. Previously, flyouts were reflowing/repositioning their contents in response to every event fired by the flyout workspace; this was particularly problematic during populating the flyout, when blocks being added/removed could cause this to happen hundreds of times. Scrolling the flyout also triggered continuous relayouts due to viewport change events. Reflowing is now only triggered in response to block change events, which can legitimately require it when e.g. the user types into a field in a block in the flyout and the flyout needs to resize to accomodate the block's new dimensions.

Additionally, the flyout was also listening to all events from the main/target workspace, and running a handler to check for and disable blocks over the maximum allowed instances threshold. This has been updated to only run on block create/deletes, since these are the only events than can result in a block exceeding/falling back under this threshold.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -711,9 +711,6 @@ export abstract class Flyout
 
     this.filterForCapacity();
 
-    // Correctly position the flyout's scrollbar when it opens.
-    this.position();
-
     // Listen for block change events, and reflow the flyout in response. This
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):
So simple.

PR #8573, opened 2024-09-11
refactor: remove redundant flyout positioning.

gonfunko:
## The basics

## The details
This PR removes a redundant call to `Flyout.position()` in `WorkspaceSvg`. The code calls `Toolbox.position()` on the previous line, which [in turn calls `Flyout.position()`](https://github.com/google/blockly/blob/develop/core/toolbox/toolbox.ts#L723), so the flyout was needlessly being positioned twice.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -1050,9 +1050,6 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
     if (this.toolbox_) {
       this.toolbox_.position();
     }
-    if (this.flyout) {
-      this.flyout.position();
-    }
 
     const positionables = this.componentManager.getComponents(
       ComponentManager.Capability.POSITIONABLE,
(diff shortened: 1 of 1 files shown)


core/workspace_svg.ts
```diff
@@ -1050,9 +1050,6 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
     if (this.toolbox_) {
       this.toolbox_.position();
     }
-    if (this.flyout) {
```
rachel-fenichel, line 1053: You still need to handle the case where there's no toolbox but there is a flyout. So you want an else if, rather than just removing this.

rachel-fenichel (changes requested):

core/workspace_svg.ts
gonfunko: Ah good call, done1

rachel-fenichel (approved):

PR #8592, opened 2024-09-25
feat: Add support for inflating flyout separators.

gonfunko:
## The basics

### Proposed Changes
This PR adds a new `FlyoutSeparator` class to represent a gap between items in a flyout, and a corresponding flyout inflater for items of this type. This will allow flyouts to no longer have to track gaps and content in separate lists, and instead just use one list of content items, some of which are rendered blocks/buttons/labels/etc and some of which are invisible spacers that simply occupy space between other items.

### DIFF
--- a/core/flyout_separator.ts
+++ b/core/flyout_separator.ts
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import type {IBoundedElement} from './interfaces/i_bounded_element.js';
 import {Rect} from './utils/rect.js';
 
--- a/core/flyout_separator.ts
+++ b/core/flyout_separator.ts
@@ -1,0 +1,17 @@
+import type {IBoundedElement} from './interfaces/i_bounded_element.js';
+import {Rect} from './utils/rect.js';
+
+export class FlyoutSeparator implements IBoundedElement {
+  private x = 0;
+  private y = 0;
+  constructor(private gap: number) {}
+
+  getBoundingRectangle(): Rect {
+    return new Rect(this.y, this.y + this.gap, this.x, this.x + this.gap);
+  }
+
+  moveBy(dx: number, dy: number, _reason?: string[]) {
+    this.x += dx;
+    this.y += dy;
+  }
+}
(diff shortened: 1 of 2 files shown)


core/flyout_separator.ts
```diff
@@ -0,0 +1,17 @@
+import type {IBoundedElement} from './interfaces/i_bounded_element.js';
+import {Rect} from './utils/rect.js';
+
+export class FlyoutSeparator implements IBoundedElement {
```
rachel-fenichel, line 4: Add license stanza, here and in the other file.

rachel-fenichel, line 4: Add TSdoc, here and in the other file

core/flyout_separator.ts
```diff
@@ -0,0 +1,17 @@
...
+  private x = 0;
+  private y = 0;
+  constructor(private gap: number) {}
+
+  getBoundingRectangle(): Rect {
```
rachel-fenichel, line 9: If someone uses a really big gap value in a vertical flyout, can it cause the flyout to become excessively wide as a result of the gap being a square instead of a rectangle? If yes, I think we should have `gapX` and `gapY` be specified separately

core/separator_flyout_inflater.ts
```diff
@@ -0,0 +1,28 @@
...
+export class SeparatorFlyoutInflater implements IFlyoutInflater {
+  load(_state: Object, _flyoutWorkspace: WorkspaceSvg): IBoundedElement {
+    // This empty separator will be substituted by the one created by the
+    // flyout based on the value returned from gapForElement().
+    return new FlyoutSeparator(0);
+  }
+
+  gapForElement(state: Object, defaultGap: number): number {
+    const separatorState = state as SeparatorInfo;
+    const newGap = parseInt(String(separatorState['gap']));
```
rachel-fenichel, line 17: Use dot-access for `separatorState.gap` if possible.

rachel-fenichel (changes requested):

core/flyout_separator.ts
gonfunko: Done

gonfunko: Done

core/flyout_separator.ts
gonfunko: Yes; updated it to take an axis argument and be zero-width on the non-dominant axis.

core/separator_flyout_inflater.ts
gonfunko: As discussed offline I think dot access will get minified/obfuscated and break when JSON is passed in.

rachel-fenichel (approved):

PR #8608, opened 2024-10-01
feat: Allow specifying the placeholder text of workspace comments.

gonfunko:
## The basics

## The details
This PR adds a `setPlaceholderText()` method to `CommentView` and `RenderedWorkspaceComment` that allows specifying the placeholder text for a comment. It also updates the workspace comment creation contextual menu item to set the default text as a placeholder instead of literal text so that users don't have to manually delete/type over it to write their comment.

### DIFF
--- a/core/comments/comment_view.ts
+++ b/core/comments/comment_view.ts
@@ -684,6 +684,11 @@ export class CommentView implements IRenderedElement {
     this.onTextChange();
   }
 
+  /** Sets the placeholder text displayed for an empty comment. */
+  setPlaceholderText(text: string) {
+    this.textArea.placeholder = text;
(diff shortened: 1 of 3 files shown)


cpcallen (approved):
Seems like a good change.  The only thing I wonder is if we should rename `WORKSPACE_COMMENT_DEFAULT_TEXT` to `WORKSPACE_COMMENT_PLACEHOLDER_TEXT` (and tweak its description in `msg/messages.js`) for clarity.  But doing that should probably be held until we have a two-way connection with Translatewiki again, so could you file a bug about it?

gonfunko:
Sure, filed #8609.

Issue #8168, opened 2024-05-21
New dancing blocks with Zelos - blocks move when field edited


clementcontet:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Description

I have found another strange dancing blocks issue in my app, that I can reproduce in the playground.
I don't know if it is linked to https://github.com/google/blockly/issues/7898 ?

### Reproduction steps

See screencast

### Screenshots

https://github.com/google/blockly/assets/7261426/dd23247a-b3ac-451a-a68e-ff7f79373f69

### Browsers

Chrome desktop

BeksOmega:
Thanks for filing! I think this is related to bumping neighbours.

BeksOmega:
This also repros on 10.3

johnnesky:
Confirmed that the same bug occurs in the geras renderer.
https://github.com/user-attachments/assets/a68fdf48-f5ad-44a3-b210-e3dab0e7d430

PR #8613, opened 2024-10-07
fix: bump connected connections in a different direction

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8168

### Proposed Changes

When two connections that are compatible but are not connected to each other are close enough to trigger a bump, the choice of which connection to move and in what direction now depends on whether the inferior connection is connected to something else. Previously, if they were both movable, the inferior connection would always be bumped down and right. Now, if the inferior connection is connected, then it will remain in place and the superior connection will be bumped up and right instead.

I also refactored related logic for readability and codified the existing invariant that the parameter to the bump function is the superior connection.

### Reason for Changes

It's possible for two independent groups of connected blocks to have multiple neighboring pairs of connections that can bump each other. Typically, the group with the "inferior" connection is moved, but it's possible that both groups can have an inferior connection that would be bumped by the other group's connection. https://github.com/google/blockly/issues/8168 demonstrates such a scenario, where the `logic_operation` block's inferior output connection is bumped by the outer `controls_if`, which moves it downward and puts its superior input connection near the `logic_compare` block's output connection, resulting in the other group getting bumped. This all happens instantaneously, and results in both groups of blocks moving down and right, lining both them up again to get bumped the next time bumping is triggered.

This can occur when a group contains both an inferior connection and at least one superior connection. However, a group of blocks can only have one unconnected inferior connection, in the top left corner. All other inferior connections must be connected to a parent block in order for associated block to be part of the group. So most of the connections that are in danger of bumping neighboring connections are either superior connections or connected inferior connections. It is important that bumping either of these two kinds of connections does not cancel out the motion of previously bumped connections. 

Thus, I propose we bump them in orthogonal directions: superior up and right, inferior down and right.

A minor downside of this strategy is that it's generally preferable that superior connections end up to the left of inferior connections, to show a gap between the disconnected blocks. However, if the inferior block is already connected to another block, then there won't be any negative space in the gap anyway because it'll be filled by the other block.

### Test Coverage

All existing unit tests pass (except for 4 test methods that were already failing due to coordinates being slightly different when testing on my mac). I manually tested that the linked bug is avoided by this PR.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/rendered_connection.ts
+++ b/core/rendered_connection.ts
@@ -114,62 +114,90 @@ export class RenderedConnection extends Connection {
   }
 
   /**
-   * Move the block(s) belonging to the connection to a point where they don't
-   * visually interfere with the specified connection.
+   * Move the block(s) associated with either this connection or the provided
+   * connection such that there is a visually clear gap between them.
    *
-   * @param staticConnection The connection to move away from.
+   * @param superiorConnection The other connection. The provided connection
+   *     should be the superior connection and should not be connected to this
+   *     connection.
    * @internal
    */
-  bumpAwayFrom(staticConnection: RenderedConnection) {
+  bumpAwayFrom(superiorConnection: RenderedConnection) {
     if (this.sourceBlock_.workspace.isDragging()) {
       // Don't move blocks around while the user is doing the same.
       return;
     }
-    // Move the root block.
-    let rootBlock = this.sourceBlock_.getRootBlock();
-    if (rootBlock.isInFlyout) {
+    let offsetX =
+      config.snapRadius + Math.floor(Math.random() * BUMP_RANDOMNESS);
+    let offsetY =
+      config.snapRadius + Math.floor(Math.random() * BUMP_RANDOMNESS);
+    /* eslint-disable-next-line @typescript-eslint/no-this-alias */
+    const inferiorConnection = this;
+    const superiorRootBlock = superiorConnection.sourceBlock_.getRootBlock();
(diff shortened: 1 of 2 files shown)


johnnesky:
Whoops, I've noticed that this technique isn't foolproof, and can create situations where both groups of blocks move up and right instead. For example:

https://github.com/user-attachments/assets/19a786e9-c620-4aa3-a5fc-34b67b76aeb6

Perhaps a better strategy would be to treat connections differently depending on whether they are the "initiating" connection, i.e. they're part of the group that `BlockSvg.bumpNeighbours()` was called on. I can try this strategy next.

johnnesky:
I update the PR, and it seems to be more effective in preventing infinite bump cycles now.

core/rendered_connection.ts
```diff
@@ -117,59 +117,85 @@ export class RenderedConnection extends Connection {
...
-      // When reversing a bump due to an uneditable block, bump up.
-      dy = -dy;
-    }
-    if (rootBlock.RTL) {
-      dx =
-        staticConnection.x -
-        config.snapRadius -
-        Math.floor(Math.random() * BUMP_RANDOMNESS) -
-        this.x;
+    const selected = common.getSelected() == dynamicRootBlock;
```
gonfunko, line 190: Nit: probably want === here?

gonfunko (approved):

core/rendered_connection.ts
johnnesky: Done!

PR #8634, opened 2024-10-31
refactor: Move functions into FieldDropdown.

### LINKED ISSUE #6349: Enhance FieldDropdown's extendability

I would like to extend the [FieldDropdown](https://github.com/google/blockly/blob/master/core/field_dropdown.js) class to allow disabled options.
I was hoping to be able to simply override the `dropdownCreate_` function, to parse new option parameters in `content` as is done for image options, but it is currently very difficult to do so without copying most of the code as `validateOptions` is not a member of the class, so it cannot be adapted to take new parameters into consideration when validating the options.

Move `validateOptions` to be a static function (possible private) of the FieldDropdown class.

The only alternative I can think of to achieve what I need at the moment is to copy the constructor and `getOptions` of the class into the new class.

**Additional context**

None.

gonfunko:
## The basics

### Resolves

Fixes #6349 

### Proposed Changes
This PR moves several functions that were in field_dropdown.ts, not exported, and used only by FieldDropdown, to simply be members of the FieldDropdown class in order to allow subclasses to override them.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -766,22 +766,14 @@ export class FieldDropdown extends Field<string> {
       if (!Array.isArray(tuple)) {
         foundError = true;
         console.error(
-          'Invalid option[' +
-            i +
-            ']: Each FieldDropdown option must be an ' +
-            'array. Found: ',
-          tuple,
+          `Invalid option[${i}]: Each FieldDropdown option must be an array.
+          Found: ${tuple}`,
         );
       } else if (typeof tuple[1] !== 'string') {
         foundError = true;
         console.error(
-          'Invalid option[' +
-            i +
-            ']: Each FieldDropdown option id must be ' +
-            'a string. Found ' +
-            tuple[1] +
-            ' in: ',
-          tuple,
+          `Invalid option[${i}]: Each FieldDropdown option id must be a string. 
+          Found ${tuple[1]} in: ${tuple}`,
         );
       } else if (
         tuple[0] &&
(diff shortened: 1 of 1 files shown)


core/field_dropdown.ts
```diff
@@ -651,6 +660,140 @@ export class FieldDropdown extends Field<string> {
...
+    for (let i = 0; i < options.length; i++) {
+      const tuple = options[i];
+      if (!Array.isArray(tuple)) {
+        foundError = true;
+        console.error(
+          'Invalid option[' +
+            i +
+            ']: Each FieldDropdown option must be an ' +
+            'array. Found: ',
+          tuple,
```
cpcallen, line 763: Consider using template literals (here and below).

cpcallen (commented):
I admit that I am mildly concerned about arbitrarily increasing the size of our API, but this seems to be in aid of a good cause.

Is anything here breaking?  If not, why not do this in v11?  (If it is, then PR title needs updating.)

core/field_dropdown.ts
```diff
@@ -95,6 +95,15 @@ export class FieldDropdown extends Field<string> {
...
   override clickTarget_: SVGElement | null = null;
 
+  /**
+   * The y offset from the top of the field to the top of the image, if an image
+   * is selected.
+   */
+  protected IMAGE_Y_OFFSET = 5;
+
+  /** The total vertical padding above and below an image. */
+  protected IMAGE_Y_PADDING = this.IMAGE_Y_OFFSET * 2;
```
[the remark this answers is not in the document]
gonfunko: Done

core/field_dropdown.ts
gonfunko: Done

gonfunko:
> I admit that I am mildly concerned about arbitrarily increasing the size of our API, but this seems to be in aid of a good cause.
> 
> Is anything here breaking? If not, why not do this in v11? (If it is, then PR title needs updating.)

Nothing here is breaking - there are no changes to behavior, it's just moving functions that were inaccessible into the class. I suppose in theory if somebody had a subclass that happened to have added methods with these names there would be an issue, but that feels like a stretch. Rebased to develop as well.

cpcallen (approved):

PR #8646, opened 2024-11-04
fix: Fix crash when resizing page while editing a field.

### LINKED ISSUE #7809: Resize while editing field leads to crash
### Check for duplicates

### Description

A resize appears to lead to a crash if a field is being edited at the same time.

The reproduction steps below are a bit artificial, but it originally occurred during normal usage.

If you resize yourself, it usually doesn't happen because the field gets unselected. But it can happen due to an external resize while editing. In my case, I think my computer went to sleep/lock screen while editing. On opening again, I was greeted by the red screen :smile: 

### Reproduction steps

See video below:

1. Go to the default [advanced playground with Zelos](https://blockly-demo.appspot.com/static/tests/playgrounds/advanced_playground.html#renderer=zelos&debugEnabled=false).
2. Connect at least two blocks, such as two repeat blocks as shown below.
3. Edit a field. While you do, enforce an external resize. The easiest way is to toggle the device test built into the Chrome web dev tools, and then back again. If it isn't already, make sure it's set to one of the non-responsive devices (e.g. pick an iPhone from the dimension list).

### Stack trace

```shell
Uncaught Error: Block has parent
    at BlockSvg$$module$build$src$core$block_svg.moveBy (block_svg.ts:388:13)
    at bumpObjectIntoBounds$$module$build$src$core$bump_objects (bump_objects.ts:80:12)
    at FieldNumber$$module$build$src$core$field_number.repositionForWindowResize (field_input.ts:646:32)
    at repositionForWindowResize$$module$build$src$core$widgetdiv (widgetdiv.ts:270:42)
    at inject.ts:226:17
    at f (browser_events.ts:65:9)
moveBy @ block_svg.ts:388
bumpObjectIntoBounds$$module$build$src$core$bump_objects @ bump_objects.ts:80
repositionForWindowResize @ field_input.ts:646
repositionForWindowResize$$module$build$src$core$widgetdiv @ widgetdiv.ts:270
(anonymous) @ inject.ts:226
f @ browser_events.ts:65
```

### Screenshots

https://github.com/google/blockly/assets/12326241/0940e90c-b086-4a6b-aadd-ab63f26deb38

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #7809

### Proposed Changes
This PR fixes a crash that could occur when the window was resized while a field has focus/is being edited. Calling `moveBy` on a child block is invalid, but because FieldInput was bumping its parent block (potentially nested) during resizes, that invalid code path could be triggered. Instead, the root block of the FieldInput will be bumped into bounds if needed, which is safe. `resizeEditor_()` was also updated to wait for the active render to complete, since it needs to know the bounds of its parent block, which was causing rendering glitches.

### DIFF
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -28,6 +28,7 @@ import {
   UnattachedFieldError,
 } from './field.js';
 import {Msg} from './msg.js';
+import * as renderManagement from './render_management.js';
 import * as aria from './utils/aria.js';
 import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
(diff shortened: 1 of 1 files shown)


core/field_input.ts
```diff
@@ -630,22 +631,24 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
-    }
-    const div = WidgetDiv.getDiv();
-    const bBox = this.getScaledBBox();
-    div!.style.width = bBox.right - bBox.left + 'px';
-    div!.style.height = bBox.bottom - bBox.top + 'px';
+    renderManagement.finishQueuedRenders().then(() => {
+      const block = this.getSourceBlock();
+      if (!block) {
+        throw new UnattachedFieldError();
+      }
```
cpcallen, line 638:
```suggestion
      if (!block) throw new UnattachedFieldError();
```

core/field_input.ts
```diff
@@ -630,22 +631,24 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
-    const xy = new Coordinate(x, bBox.top);
+      // In RTL mode block fields and LTR input fields the left edge moves,
+      // whereas the right edge is fixed.  Reposition the editor.
+      const x = block.RTL ? bBox.right - div!.offsetWidth : bBox.left;
+      const xy = new Coordinate(x, bBox.top);
 
-    div!.style.left = xy.x + 'px';
-    div!.style.top = xy.y + 'px';
+      div!.style.left = xy.x + 'px';
+      div!.style.top = xy.y + 'px';
```
cpcallen, line 650: Does putting `x` and `bBox.top` into a `Coordinate` and then taking them back out again acutally do anything?

cpcallen (approved):
LGTM but two tweaks suggested if you feel like improving the old code while you're at it.

core/field_input.ts
gonfunko: Done

core/field_input.ts
gonfunko: Indeed it does not, removed.

PR #8648, opened 2024-11-06
feat!: Use CSS to specify field cursors.

### LINKED ISSUE #995: Dragging from certain fields doesn't show the right cursor.
Dragging from a dropdown or an icon (and possibly other fields) shows the default cursor instead of the grabbing hand.

Dragging from an editable text field shows the I-beam cursor instead of the grabbing hand.

If during a drag you exceed the boundaries of that field, it shows the correct cursor briefly.

### LINKED ISSUE #2884: Add a Generic CSS Class Config Option to Fields

It would be nice for outside developers to have more control over the on-block display of different fields, especially if they are trying to customize the look-and-feel of Blockly for different audiences.

The ability to pass a class string to the JS Constructor's opt_config property or the JSON Definition of a field would be one way to accomplish this. The string would then be appended to the field's fieldGroup_ during initialization.

* Giving each field type its own css class that could be appended to the fieldGroup_ would also go a long way towards customizability.
* Adding field customization to the themes API is also an option, but it seems like that would complicate the simplicity of themes.

**Open Problems**

**CSS API**
Currently I don't believe the CSS or DOM structure are considered to be part of Blockly's public API. Adding the ability to apply classes to fields would, in a way, make them part of the public API because fields' structures and css would have to become stable, so as not to break people.

The CSS should probably be reorganized before trying to complete this issue.

**Relation to Themes**
Themes make the colours of blockly dynamic, which is rad, but there's no way to access those colours from CSS (as far as I know). This could complicate things.

**Field Sizing**
Currently field sizing is handled programatically. This makes it less flexible when trying to impliment CSS. For example if an outside developer added a class to a colour field that made it taller or wider than default, it would not be handled correctly by the renderer.

**Editors**
It may also be useful for outside developers to style the editor of a field (angle fields come to mind as a good example).

### LINKED ISSUE #2885: Field Cursor Configuration

It is currently unclear if you should, and how you can change the cursor for fields.

E.g:
* Can/Should you change the cursor of a field type?
* Can/Should you change the cursor of a field instance?
* Can/Should you change it dynamically (e.g. based on if the field is editable)?

I think it would be the most flexible to say yes to all three.

* Some CURSOR properties already exist on field subclasses. One could also be added to the abstract field class. That way you won't have to dig through the css to figure out what the default cursor is.
* Add a cursor property to the opt_config of the abstract field class.  This would allow it to also be passed through the JSON Definition.
* Add a setCursor function that allows the cursor to be changed dynamically.

* No to all three.
* No to some and yes to some.

The abstract field class should maybe get a CURSOR property anyway though, for consistency.

**Additional context**

I've seen people setting the CURSOR property of field instances directly (on the forums), so I think there is demand for this.

This could also be handled by CSS instead of programmatically. Related: #2884

### LINKED ISSUE #8319: Remove the field `CURSOR` property
Remove the field `CURSOR` property, and instead assign cursors to fields in `css.ts`.

This is a breaking change.

gonfunko:
## The basics

### Resolves

* Fixes #8319
* Fixes #2885
* Fixes #2884
* Fixes #995

### Proposed Changes
This PR removes the `CURSOR` field from fields, and instead uses CSS to specify which cursor should be used for which fields. This directly fixes #8319. #2885 and #2884 are old bugs proposing adding options to inject classes into fields and refactoring the `CURSOR` field, which I believe are obsolete in light of the move to CSS. All core fields now add a class specifying the type of field, and custom field subclasses can also easily do this. As a result, users should be able to use CSS to customize the appearance of fields in general and their cursors in particular.

#995 was incidentally fixed here, since now that we're using CSS to specify cursors, it's trivial to specify the grabby cursor for fields when their parent block is being dragged.

### DIFF
--- a/core/css.ts
+++ b/core/css.ts
@@ -478,4 +478,17 @@ input[type=number] {
   z-index: 80;
   pointer-events: none;
 }
+
+.blocklyField {
+  cursor: default;
+}
+
+.blocklyInputField {
+  cursor: text;
+}
+
+.blocklyDragging .blocklyField,
+.blocklyDragging .blocklyIconGroup {
+  cursor: grabbing;
+}
 `;
--- a/core/field.ts
+++ b/core/field.ts
@@ -193,9 +193,6 @@ export abstract class Field<T = any>
    */
   SERIALIZABLE = false;
 
-  /** Mouse cursor style when over the hotspot that initiates the editor. */
-  CURSOR = '';
-
   /**
    * @param value The initial value of the field.
    *     Also accepts Field.SKIP_SETUP if you wish to skip setup (only used by
@@ -536,11 +533,9 @@ export abstract class Field<T = any>
     if (this.enabled_ && block.isEditable()) {
       dom.addClass(group, 'blocklyEditableField');
       dom.removeClass(group, 'blocklyNonEditableField');
-      group.style.cursor = this.CURSOR;
     } else {
       dom.addClass(group, 'blocklyNonEditableField');
       dom.removeClass(group, 'blocklyEditableField');
-      group.style.cursor = '';
     }
   }
 
--- a/core/field_checkbox.ts
+++ b/core/field_checkbox.ts
@@ -36,11 +36,6 @@ export class FieldCheckbox extends Field<CheckboxBool> {
   override SERIALIZABLE = true;
 
   /**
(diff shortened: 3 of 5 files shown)


cpcallen (approved):
That is an excellent ratio of bugs fixed to delta lines.

PR #8650, opened 2024-11-08
refactor!: Add setSelectedItem() to IToolbox.

### LINKED ISSUE #7273: Add support for selecting toolbox items by ID
### Check for duplicates

### Problem

Currently, IToolbox defines a `selectItemByPosition()` method, that allows for selecting a toolbox item by its index. There is however no straightforward way to select an item by type/kind; this came up in https://github.com/google/blockly-samples/pull/1751/files#diff-af9d92f020d53ffacb71e0519ecb2e9fa99c6b0d888131f7e7c5a32376df3553R71, where a toolbox item had to go hunting for itself in the toolbox definition in order to determine its index to select itself.

### Request

It would be great if IToolbox also allowed selecting an item by kind and/or ID, so that a known toolbox item could easily be selected without having to know (or find) where exactly in the toolbox it is.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #7273

### Proposed Changes
This PR updates the IToolbox interface by adding `setSelectedItem()`, to pair with the existing `getSelectedItem()`. This is a breaking change; implementers of this interface will need to add this method. Blockly's core Toolbox class already does implement this method, so subclasses of that will get it for free. The linked issue requested a method to select an item by ID, but since this (a) adds a setter to an existing getter (b) is already implemented in the core Toolbox class and (c) you'd need to have an item to get its ID anyway, I think this addresses the root need and is a bit cleaner.

### DIFF
--- a/core/interfaces/i_toolbox.ts
+++ b/core/interfaces/i_toolbox.ts
@@ -94,7 +94,7 @@ export interface IToolbox extends IRegistrable {
   setVisible(isVisible: boolean): void;
 
   /**
-   * Selects the toolbox item by it's position in the list of toolbox items.
+   * Selects the toolbox item by its position in the list of toolbox items.
    *
    * @param position The position of the item to select.
    */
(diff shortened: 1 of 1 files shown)


cpcallen (approved):

cpcallen:
Can I recommend you pull out the relevant parts of the PR description into a (thus formatted)

**BREAKING CHANGE: …** 

section, for ease of reference—and make sure that that section is included in the squash commit message?

gonfunko:
Sorry I missed this when I merged, but edited the PR description to call it out in its own section.

Issue #8655, opened 2024-11-12
Workspace shifts when deleting a block


gonfunko:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

https://groups.google.com/g/blockly/c/d-khS8yDMhk outlines the problem, but in short, when blocks are adjacent to the toolbox with some space to the right, and a block is dragged into the toolbox to delete it, the remaining blocks shift to the right.

### Stack trace

```shell

```

mateusslezinsky:
@gonfunko hello, my name is Mateus and I am getting started with contributions to this repository! 

I would like to give this issue a try, I have a solution in mind.

cpcallen:
Hi Mateus: if you have a potential fix then do please go ahead and send us a PR!  (Otherwise, if you are no longer interested or are not able to contribute, please unassign yourself so we know that you're not working on it any more.)

gonfunko:
Fixed by #8666.

PR #8666, opened 2024-11-18
fix: workspace shifts when deleting a block

mateusslezinsky:
## The basics

### Resolves

[Workspace shifts when deleting a block](https://github.com/google/blockly/issues/8655)

### Proposed Changes
Instead of resizing the workspace everytime a drag is finished, resizes when a block is deleted.

### Reason for Changes
To fix issue #8655.

### Test Coverage
Before changes:
[before_changes.webm](https://github.com/user-attachments/assets/4d7ce6af-7680-432e-a245-94e97da48fcb)

After changes:
[after_changes.webm](https://github.com/user-attachments/assets/14d03731-8e36-415e-ba02-ab4b140e0927)

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -358,7 +358,7 @@ export class BlockDragStrategy implements IDragStrategy {
    * Cleans up any state at the end of the drag. Applies any pending
    * connections.
    */
-  async endDrag(e?: PointerEvent): Promise<void> {
+  endDrag(e?: PointerEvent): void {
     if (this.block.isShadow()) {
       this.block.getParent()?.endDrag(e);
       return;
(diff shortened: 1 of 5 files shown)


gonfunko:
Hi, thank you for your interest and resolving this issue and for putting together this PR! While I see from your videos it stops the workspace movement, I think the original call to `setResizesEnabled()` is still necessary, because it pairs with the corresponding call in `startDrag()` to reenable workspace resizing after `startDrag()` disables it for the duration of the drag. In cases where the drag does not result in deletion, resizing would never be reenabled with this change. If you're interested in investigating further, I'd investigate the calls to `snapToGrid()` and `queueRender()`;  my suspicion is there may be a race condition with block deletion and the things (like snapping and rerendering) we'd do at the end of a normal non-deletion drag which are unnecessary if the block is being deleted but desirable (and could potentially result in adjusting the workspace bounds) in the case where a block is just being moved.

mateusslezinsky:
> Hi, thank you for your interest and resolving this issue and for putting together this PR! While I see from your videos it stops the workspace movement, I think the original call to `setResizesEnabled()` is still necessary, because it pairs with the corresponding call in `startDrag()` to reenable workspace resizing after `startDrag()` disables it for the duration of the drag. In cases where the drag does not result in deletion, resizing would never be reenabled with this change. If you're interested in investigating further, I'd investigate the calls to `snapToGrid()` and `queueRender()`; my suspicion is there may be a race condition with block deletion and the things (like snapping and rerendering) we'd do at the end of a normal non-deletion drag which are unnecessary if the block is being deleted but desirable (and could potentially result in adjusting the workspace bounds) in the case where a block is just being moved.

Hello! I have been debugging this issue and it indeed looks like a race condition. The reason for this is because `this.block.queueRender()` returns a Promise, which is async.

Marking `endDrag` method from `block_drag_strategy` does fixes the issue however. Should I choose this approach or attempt other fix?

mateusslezinsky:
After the latest changes, here is how it looks like:
[latest.webm](https://github.com/user-attachments/assets/5e6bb85c-d28a-4ee6-a501-b2a2a1e19fb1)

gonfunko:
Ah, this looks pretty in line with what I suspected, thank you for digging into it! Would you mind adjusting it to use `renderManagement.finishQueuedRenders().then(() => { // Everything after the if, from snapToGrid() to setGroup(false)  })` to avoid making the whole function async?

mateusslezinsky:
@gonfunko done. Now the method is no longer async and we have a reusable method that can be executed both sync/async.

gonfunko (approved):

gonfunko:
Looks good! Thanks for taking care of this and working through the back and forth, I and the rest of the Blockly team appreciate it!

PR #8667, opened 2024-11-21
feat: Add a generator for all fields on a block.

### LINKED ISSUE #5071: Create a javascript generator to iterate fields (not code generator)

There are a number of places in the codebase where we iterate all of the fields of the block. This is unfortunate because the way the data types are structure requires using a nested loop:
```javascript
for (let i = 0; i < block.inputList.length; i++) {
  const input = block.inputList[i];
  for (let j = 0; j < input.fieldRow.length; j++) {
    const field = input.fieldRow[j];
    // Do something with field.
  }
}
```
Which is not very readable.

It would be better to create a [javascript generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) which makes what you're doing more clear.

You could also add a method to the block which takes in a callback and iterates the fields:
```
Blockly.Block.forEachField = function(callback, thisArg) {
   // the code from before
  callback.call(thisArg, field);
}
```

**Additional context**

Some places this pattern would be useful:
* [Serializing fields](https://github.com/google/blockly/blob/e53168925d6b95a7458136a3cd3c1cda204540a1/core/xml.js#L133)
* [Initting models](https://github.com/google/blockly/blob/e53168925d6b95a7458136a3cd3c1cda204540a1/core/block.js#L432)
* [Updating editability](https://github.com/google/blockly/blob/e53168925d6b95a7458136a3cd3c1cda204540a1/core/block.js#L890)
* [Getting vars](https://github.com/google/blockly/blob/e53168925d6b95a7458136a3cd3c1cda204540a1/core/block.js#L1047)
* Etc

gonfunko:
## The basics

### Resolves

Fixes #5071

### Proposed Changes
This PR adds a `getFields()` [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) to `Block` that returns a generator that may be used to iterate through all the fields on the block. It also updates core to use this method in preference to nested for loops.


core/block.ts
```diff
@@ -1102,29 +1100,37 @@ export class Block implements IASTNodeLocation {
...
-        }
+    for (const field of this.getFields()) {
+      if (field.name === name) {
+        return field;
       }
     }
     return null;
   }
 
+  /**
```
rachel-fenichel, line 1111: Add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator to clarify that this is generator-in-javascript-language and not Blockly-Javascript-generator.

rachel-fenichel (approved):

core/block.ts
gonfunko: Done

Issue #8272, opened 2024-07-10
Refactor SVG filters to be singletons


BeksOmega:
We have certain filters, such as the disabled filter, that are created multiple times with unique IDs ([source code link](https://github.com/google/blockly/blob/be268e3bb4fbf2d6973ff468052ff6deba70ef91/core/inject.ts#L132)). This was necessary because of a browser bug that has since been fixed. We should refactor these so that they're singletons, so we can reference them from CSS.

Fixing this will also involve actually referencing them from CSS. 

This should be merged against the `rc/v12.0.0` branch.

See the [design doc](https://docs.google.com/document/d/1JLSuu0yqK9Yu9tx0zugBzlcA7qFcpX9bfHh15LDt29c/edit#heading=h.btkkrkm4t1ju) for more info.

jiya-singhal:
hey @BeksOmega
Can you assign this issue to me?

johnnesky:
Another concern is that there isn't an obvious place to put a shared global `defs` element. It has to be in an `svg` element to be effective, and `svg` elements are not supposed to go inside the `head` element, unlike css `style` elements. (Browsers would probably let us do it via JavaScript but if browsers directly load HTML that contains an `svg` in the `head` they would forcibly move it to the `body`.)

We could just put it at the beginning or end of the `body` element's children, and apply styles to visually hide the `svg` element such that it does not affect layout. We have to be careful, though, because using `display: none;` to hide an `svg` element also excludes its `defs` from being referenced elsewhere. (Which is another reason why we can't put the `defs` inside the `head` element which is usually hidden in this manner.) Instead, I propose we use the following style to hide the `svg` element:
```
<svg style="position: absolute !important; visibility: hidden !important; width: 0 !important; height: 0 !important; top: 0 !important; left: 0 !important;">
  <defs>...</defs>
</svg>
```

This would work, but we would have to hope that sneaking an `svg` element into the `body` won't mess up other expectations of the web app. For example, if the app has any CSS selectors like `body :first-child` then the hidden `svg` would steal that selector from another child. Similarly, if the app has any JavaScript that iterates over the children of the `body`, then our hidden `svg` element could confuse it.

[...]

Issue #1403, opened 2017-10-30
Disabled Blocks Pattern Colour Selection


asanzo:
### Problem statement

- [ ] Bug report
- [x] Feature request

### Expected Behavior

It'd be nice to be able to change the colour of the pattern in the disabled blocks.

In order to be able to do something like this:

```Javascript
Blockly.Msg.DISABLED_BACKGROUND_HUE = 0;
Blockly.Msg.DISABLED_PATTERN_HUE = 60;
```

### Actual Behavior

In some applications, the resultant colour of the disabled blocks makes the blocks difficult to see, or distinguish the borders and even read the labels:

![seleccion_008](https://user-images.githubusercontent.com/5421992/32189105-868398b0-bd88-11e7-8549-4b76ee94edaf.png)

### Steps to Reproduce

Just make a block disabled

### Stack Traces

Probably the line to modify should be around [here](https://github.com/google/blockly/blob/e93d44e8ff589b9bf21129d50d710030d9345504/core/inject.js#L159)

### Operating System and Browser

- [x] Desktop:
  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] Opera
  - [ ] IE 10+
  - [ ] IE 11
  - [ ] EDGE

Reproduced in Win 7/10 and Ubuntu/Mint

Thanks in advance for any respose :)

NeilFraser:
Note that you can easily override the look of disabled blocks with CSS.  For example:

.blocklyDisabled>.blocklyPath {
  fill: #f00;
}

gonfunko:
Closing as the CSS approach is the recommended way to do this. The `--blocklyDisabledPattern` CSS variable can also be overridden to reference a completely different SVG pattern for deeper control.

PR #8671, opened 2024-11-23
fix: Create CSS vars for SVG patterns.

johnnesky:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8272

We'll need a related PR in blockly-samples that adapts the screenshot plugin to be aware of CSS vars.

### Proposed Changes

Adds CSS variables scoped to the injection div referencing the SVG pattern ids so that CSS can use these consistent variable names instead of the random ids that were previously assigned to SVG attributes.

### Reason for Changes

So that renderer CSS can customize and override how SVG patterns are used.

### Test Coverage

I did not create any new tests, but all existing tests pass and I manually tested that the filters appear to be rendering correctly.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/bubbles/bubble.ts
+++ b/core/bubbles/bubble.ts
@@ -106,11 +106,7 @@ export abstract class Bubble implements IBubble, ISelectable {
     );
     const embossGroup = dom.createSvgElement(
       Svg.G,
-      {
-        'filter': `url(#${
-          this.workspace.getRenderer().getConstants().embossFilterId
-        })`,
-      },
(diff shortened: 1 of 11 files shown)


gonfunko (approved):

gonfunko:
I think this also fixes #1403?

johnnesky:
> I think this also fixes #1403?

Depends on what they wanted to do to the disabled pattern. I wouldn't close that yet without putting some thought into it.

PR #8675, opened 2024-11-27
chore: Migrate ESLint configuration file to new flat format.

gonfunko:
## The basics

### Resolves

Unblocks #8664 

### Proposed Changes
This PR updates the ESLint configuration file to the new flat format introduced in v8 that became mandatory in v9. This unblocks upgrading a number of dev dependencies.


core/variables.ts
```diff
@@ -202,7 +202,6 @@ export function generateUniqueNameFromOptions(
   let letterIndex = letters.indexOf(startChar);
   let potName = startChar;
 
-  // eslint-disable-next-line no-constant-condition
```
rachel-fenichel, line 205: This rule is still part of recommended. Does removing this line mean that you disabled no-constant-condition for all of Blockly, or that the recommended ruleset is no longer fully being applied?

gonfunko: Per https://eslint.org/docs/latest/rules/no-constant-condition#checkloops, the `allExceptWhileTrue` option for this rule is also on by default, so this no longer triggers.

rachel-fenichel (approved):

eslint.config.mjs
```diff
@@ -0,0 +1,308 @@
...
+      // Ensure there is a blank line between the body and any @tags,
+      // as required by the tsdoc spec (see #6353).
+      'jsdoc/tag-lines': ['error', 'any', {'startLines': 1}],
+    },
+  };
+}
+
+export default [
+  {
+    // Note: there should be no other properties in this object
```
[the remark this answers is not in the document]
gonfunko: Correct, per https://eslint.org/docs/latest/use/configure/ignore#ignoring-files the `ignores` key needs to be used by itself.

rachel-fenichel:
LGTM once package-lock conflicts are resolved.

PR #8682, opened 2024-12-02
chore: Remove underscores from private fields.

### LINKED ISSUE #6548: Remove underscores from private property and method names

gonfunko:
## The basics

### Resolves

Fixes #6548

### Proposed Changes
This PR removes underscores from private fields (found via a regex search for private.+_).

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -89,7 +89,7 @@ export class Block implements IASTNodeLocation {
    * Colour of the block as HSV hue value (0-360)
    * This may be null if the block colour was not set via a hue number.
    */
-  private hue_: number | null = null;
+  private hue: number | null = null;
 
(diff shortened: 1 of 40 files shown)


rachel-fenichel (approved):
LGTM.

There are 13 test files in this change, which means 13 files where we are accessing private properties in tests. Please take a quick pass through looking for places where those accesses could be replaced with existing getters, and either change or file a follow-up bug to change.

gonfunko:
Done!

PR #8687, opened 2024-12-05
fix: Fix bug that prevented keyboard navigation in flyouts.

gonfunko:
## The basics

## The details
This PR fixes an issue that caused `findNextLocationInFlyout()` to return invalid items. With the flyout changes in v12, spacers (and potentially custom user-defined items) are included in the list of flyout contents, but are not valid keyboard navigation locations. As such, simply taking the next/previous item relative to the current one is insufficient; that item may be inspected, but if it is of a non-navigable type, the subsequent item needs to be checked until a valid navigation location is found.

### DIFF
--- a/core/flyout_item.ts
+++ b/core/flyout_item.ts
@@ -1,0 +1,42 @@
+import type {IBoundedElement} from './interfaces/i_bounded_element.js';
+
+/**
+ * Representation of an item displayed in a flyout.
+ */
+export class FlyoutItem {
+  /**
+   * Creates a new FlyoutItem.
+   *
+   * @param element The element that will be displayed in the flyout.
+   * @param type The type of element. Should correspond to the type of the
+   *     flyout inflater that created this object.
+   * @param focusable True if the element should be allowed to be focused by
+   *     e.g. keyboard navigation in the flyout.
+   */
+  constructor(
+    private element: IBoundedElement,
+    private type: string,
+    private focusable: boolean,
+  ) {}
+
+  /**
+   * Returns the element displayed in the flyout.
+   */
+  getElement() {
+    return this.element;
+  }
+
+  /**
+   * Returns the type of flyout element this item represents.
+   */
+  getType() {
+    return this.type;
+  }
+
+  /**
+   * Returns whether or not the flyout element can receive focus.
+   */
+  isFocusable() {
+    return this.focusable;
+  }
+}
--- a/core/keyboard_nav/ast_node.ts
+++ b/core/keyboard_nav/ast_node.ts
@@ -17,8 +17,8 @@ import {BlockSvg} from '../block_svg.js';
 import type {Connection} from '../connection.js';
 import {ConnectionType} from '../connection_type.js';
 import type {Field} from '../field.js';
-import {FlyoutItem} from '../flyout_base.js';
 import {FlyoutButton} from '../flyout_button.js';
+import type {FlyoutItem} from '../flyout_item.js';
 import type {Input} from '../inputs/input.js';
 import type {IASTNodeLocation} from '../interfaces/i_ast_node_location.js';
 import type {IASTNodeLocationWithBlock} from '../interfaces/i_ast_node_location_with_block.js';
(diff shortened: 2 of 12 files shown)


core/keyboard_nav/ast_node.ts
```diff
@@ -388,7 +388,20 @@ export class ASTNode {
...
-    const resultIndex = forward ? currentIndex + 1 : currentIndex - 1;
+    let resultIndex = forward ? currentIndex + 1 : currentIndex - 1;
+    // The flyout may contain non-navigable elements like spacers or custom
+    // items. If the next/previous element is one of those, keep looking until a
+    // block or button is encountered.
+    while (
+      resultIndex >= 0 &&
+      resultIndex < flyoutContents.length &&
+      !(
+        flyoutContents[resultIndex].element instanceof BlockSvg ||
```
rachel-fenichel, line 399: Can you file a bug to create an `isNavigable` API or similar for flyout items? This will do for now but eventually those custom items will need to be navigable (a text entry box for searching being the most obvious case).

rachel-fenichel (approved):

core/keyboard_nav/ast_node.ts
gonfunko: Yes, this is why I was asking about ASTNode in the chat :) I think we probably just want to add that method to IASTNodeLocation; should I just go ahead and do that in this PR?

rachel-fenichel: I think so; @cpcallen to confirm. This would be in the v12 beta so we'll still have time to change it.

cpcallen: I think one of us might be confused, because it doesn't make sense to me for `IASTNodeLocation` to have an `isNavigable` method, because (in practice) an `IASTNodeLocation` is a thing _one has already navigated to_ (i.e., that is the value of a `.location` property of an `ASTNode`—so you'd have to have called `.isNavigable` on the possible location _before_ you create the `ASTNode` for it (which is the first time it definitely needs to fulfil the interface), and you won't have any reason to call `isNavigable` again because you already know it's navigable.  :-/

I think the method should probably be called `isFocusable` and be on `FlyoutItem`, but I'm not sure as I don't fully understand the flyout changes in v12.

gonfunko: Thank you, I was the confused one – I didn't realize the already navigated to aspect of `IASTNodeLocation`. As suggested, I've added `isFocusable()` to `FlyoutItem`. Since this is still only in beta and part of a larger breaking change, I took the liberty of refactoring the APIs to be a bit more consistent.

core/flyout_item.ts
```diff
@@ -0,0 +1,42 @@
...
+   * @param type The type of element. Should correspond to the type of the
+   *     flyout inflater that created this object.
+   * @param focusable True if the element should be allowed to be focused by
+   *     e.g. keyboard navigation in the flyout.
+   */
+  constructor(
+    private element: IBoundedElement,
+    private type: string,
+    private focusable: boolean,
+  ) {}
```
cpcallen, line 20: Is there some reasonably foreseeable scenario in which this class might be subclassed?  If not, it might be better to make the members (public) `readonly` and omit the `get*` methods.  The [styleguide](https://google.github.io/styleguide/tsguide.html#class-members) has this rule for `get`/`set` _accessors_ (`get foo()`) which I think seems like good advice also for ordinary getter methods (`getFoo()`).

> Getters and setters, also known as accessors, for class members may be used…
> 
> If an accessor is used to hide a class property, the hidden property may be prefixed or suffixed with any whole word, like internal or wrapped. When using these private properties, access the value through the accessor whenever possible. At least one accessor for a property must be non-trivial: do not define pass-through accessors only for the purpose of hiding a property. Instead, make the property public (or consider making it readonly rather than just defining a getter with no setter).

Of course if this class is likely to get subclassed then the current approach seems sound.

gonfunko: I could imagine a situation where somebody with a custom flyout might want to add additional metadata to the items in it, so perhaps? Happy to adjust if you think that's preferable though.

rachel-fenichel (approved):

PR #8695, opened 2024-12-13
fix: Fix unsafe cast in Input.setVisible().

### LINKED ISSUE #8694: `Input.prototype.setVisible` should not assume that `this.connection` is a `RenderedConnection`
[Bug report rewritten by @cpcallen.]

The `setVisible` method of `Input` does an unsafe cast `this.connection as RenderedConnection`.  Instead, the actual type of `this.connection` should be tested an the method shoud no-op if the connection is not a `RenderedConnection`.

Original description follows.
------

### Check for duplicates

### Description

It is not possible to activate the toolbox search bar and hide a input parameter of a block.

Within the function where my block and its inputfields get initialized i try to hide some inputfields with the method setVisible(false). But unfortunately this will cause an error.

The weird thing about that is, that this bug only occurs when the toolbox search bar is activated. 

### Reproduction steps

Error occurs during the initialization  of a block and its fields. 

An example of my code:

  Blockly.Blocks['customBlockName'] = {
      init: function (this: Blockly.Block) {
        this.appendDummyInput().appendField('customBlockName');

/* 
bunch of code
.......
......
.....
*/

const inputField: Blockly.Input = this.appendValueInput('inputFieldIdInput').setCheck('String').appendField('inputFieldId')

customAddBlockService.appendshadowBlock()

inputField.setVisible(fallse)  // --> This line of code will cuase an error

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8694

### Proposed Changes
This PR checks if the connection in Input.setVisible() is actually a rendered connection before assuming it is. If `setVisible()` was called while initing a block, connections may not have been fully set up and this would throw.

### DIFF
--- a/core/inputs/input.ts
+++ b/core/inputs/input.ts
@@ -20,7 +20,7 @@ import type {Connection} from '../connection.js';
 import type {ConnectionType} from '../connection_type.js';
 import type {Field} from '../field.js';
 import * as fieldRegistry from '../field_registry.js';
-import type {RenderedConnection} from '../rendered_connection.js';
+import {RenderedConnection} from '../rendered_connection.js';
 import {Align} from './align.js';
 import {inputTypes} from './input_types.js';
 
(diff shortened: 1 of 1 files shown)


rachel-fenichel (commented):
The PR description says that this will throw if the connection is the wrong type, but it'll actually just silently fail. Please update to throw.

gonfunko:
It said it *would* throw, ie that was the previous behavior, and it now no-ops aside from setting the visiblity status, which should be reflected when the block finishes initing.

rachel-fenichel:
Ah, I misread. Thanks.

rachel-fenichel (approved):

gonfunko:
> Ah, I misread. Thanks.

No worries, it was kind of an awkward phrasing on my part.

PR #8706, opened 2024-12-22
fix: treat media files as binary while packaging them

### LINKED ISSUE #8705: Media files corrupted during packaging
### Check for duplicates

### Description

If you look at the files inside the latest release tarball (blockly-11.2.0.tgz), the binaries in the media/ directory are different than the ones in the repository. Looking closer, they have a lot of `EF BF BD` byte sequences where they shouldn't, which is the UTF-8 representation of Unicode 'REPLACEMENT CHARACTER' (U+FFFD), suggesting that something is interpreting the binaries as Unicode, which obviously isn't right for MP3s etc.

Indeed that's what gulp's src() is doing unless you pass the `encoding: false` option (it was a breaking change in version 5.0.0).

### Reproduction steps

1. npm run package
2. diff media/click.mp3 dist/media/click.mp3

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

jfedor2:
## The basics

### Resolves

Fixes #8705

### Proposed Changes

Adds `encoding: false` option to `gulp.src()` in `packageMedia()` to treat files as binary.

### Reason for Changes

Without this the files are interpreted as UTF-8, which corrupts binaries like MP3s etc. This was a breaking change in gulp 5.0.0.

### Test Coverage

Manually tested with:
```
npm run package
diff -r media dist/media
```

### DIFF
--- a/scripts/gulpfiles/package_tasks.js
+++ b/scripts/gulpfiles/package_tasks.js
@@ -159,7 +159,7 @@ module.exports = require('./${bundle}');
  * This task copies all the media/* files into the release directory.
  */
 function packageMedia() {
-  return gulp.src('media/*')
+  return gulp.src('media/*', {encoding: false})
     .pipe(gulp.dest(`${RELEASE_DIR}/media`));
 };
 


gonfunko:
Thank you for identifying and resolving this issue, we really appreciate it!

gonfunko (approved):

PR #8723, opened 2025-01-07
fix: Don't create intermediate variables when renaming a procedure argument.

### LINKED ISSUE #8693: Intermediate variables are shown in flyout while procedure mutator field is being edited
### Check for duplicates

### Description

If you add an input to a procedure and rename it, the variables category in the flyout will display variables for each keystroke until the input field in the mutator bubble loses focus, at which point the intermediate variables disappear and only the variable corresponding to the input's final name remains.

### Reproduction steps

1. Drag out a procedure block
2. Open its mutator
3. Add a new input
4. Rename the input
5. While the input name field still has focus in the mutator bubble, open the Variables category of the toolbox
6. Observe that variables corresponding to each keystroke are present
7. Click outside of the input name field in the mutator bubble to cause it to lose focus
8. Reopen the Variables category, and observe that the intermediate state variables are gone, leaving only one corresponding to the final name of the input

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8693 

### Proposed Changes
This PR updates the procedure argument renaming logic to rename the variable being edited, rather than creating a new one for every keystroke and cleaning up the intermediate states when the field editor is dismissed. This prevents intermediate variables from appearing in the flyout when it is open while a procedure argument's name is being edited in the mutator.

### DIFF
--- a/blocks/procedures.ts
+++ b/blocks/procedures.ts
@@ -629,30 +629,49 @@ type ArgumentBlock = Block & ArgumentMixin;
 interface ArgumentMixin extends ArgumentMixinType {}
 type ArgumentMixinType = typeof PROCEDURES_MUTATORARGUMENT;
 
-// TODO(#6920): This is kludgy.
-type FieldTextInputForArgument = FieldTextInput & {
-  oldShowEditorFn_(_e?: Event, quietInput?: boolean): void;
-  createdVariables_: IVariableModel<IVariableState>[];
-};
+/**
+ * Field responsible for editing procedure argument names.
+ */
+class ProcedureArgumentField extends FieldTextInput {
+  /**
+   * Whether or not this field is currently being edited interactively.
+   */
+  editingInteractively = false;
+
+  /**
+   * The procedure argument variable whose name is being interactively edited.
+   */
+  editingVariable?: IVariableModel<IVariableState>;
+
+  /**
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Christopher said to also check one edge case: "When checking any fix for this, do test (or better still: add tests) to see what happens if one of the intermediate variable names clashes with another variable/parameter—e.g., if there is a variable `foo` and you rename a parameter from `f` to `food`."

Did you check it, and does it work?

gonfunko:
Yes and yes; behavior appears identical to the old version in that case. The pre-existing variable does not get deleted if you temporarily land on it but continue past it, and if you name the arg the same they coalesce.

rachel-fenichel (approved):
This looks good to me--I like that you are actually using the variable object to make sure that you can just change the active variable, instead of creating a pile of them and then going back to delete.

We have several versions of the procedure blocks now--particularly in [block-plus-minus](https://github.com/google/blockly-samples/blob/master/plugins/block-plus-minus/src/procedures.js) and [block-shareable-procedures](https://github.com/google/blockly-samples/tree/master/plugins/block-shareable-procedures). Please check whether you can apply the same fix there, and either apply it or file bugs to track that work.

PR #8731, opened 2025-01-13
fix: Listen for keyboard shortcuts when the widget or dropdown divs have focus.

### LINKED ISSUE #8725: [11.2.0] Keyboard shortcuts aren't handled correctly when a field editor is open
### Check for duplicates

### Description

In Blockly 11.2.0, it appears that no keyboard shortcuts are no longer working if a field editor is active. This means, for example, that if a dropdown field editor is visible, you can't press 'delete' to delete the block or 'escape' to call `hideChaff()`.

This was still working as of 11.1.1, so it's likely a regression caused by something here: https://github.com/google/blockly/releases/tag/blockly-v11.2.0

### Reproduction steps

1. Visit the Blockly playground
2. Drag out any block with a dropdown field and open the field editor
3. Press the escape key, observing that the field editor doesn't close. In fact, `onKeyDown` is never called.
Image

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8725

### Proposed Changes
This PR adds keydown listeners for the widget and dropdown divs and runs those events through the shortcut infrastructure. This allows keyboard shortcuts (e.g. esc, delete, copy/paste) to operate when a menu/dropdown is focused.

### DIFF
--- a/core/inject.ts
+++ b/core/inject.ts
@@ -77,6 +77,16 @@ export function inject(
   });
 
   browserEvents.conditionalBind(subContainer, 'keydown', null, onKeyDown);
+  browserEvents.conditionalBind(
+    dropDownDiv.getContentDiv(),
+    'keydown',
+    null,
+    onKeyDown,
+  );
+  const widgetContainer = WidgetDiv.getDiv();
+  if (widgetContainer) {
+    browserEvents.conditionalBind(widgetContainer, 'keydown', null, onKeyDown);
+  }
 
   return workspace;
 }


rachel-fenichel (approved):
LGTM. I took a look at onKeyDown to confirm that it handles cases where there are multiple workspaces; I think it does, because it calls `getMainWorkspace` and should get the mostly recently modified/interacted with workspace.

mikeharv:
Thank you for the quick fix! Could you all let us know when this will make its way into a new release? Code.org will be hanging out on 11.1.1 in the meantime.

gonfunko:
Sure thing! This is included in the just-released 11.2.1.

PR #8751, opened 2025-01-28
feat: add extension point for overriding Grid functionality

### LINKED ISSUE #8614: Add Grid extension point
### Check for duplicates

### Problem

Currently there is no mechanism I'm aware of for overriding the Grid class that renders the grid background for workspaces. Being able to extend this class would allow for more background options such as a repeating image or other custom patterns.

You can override the background today using CSS, but it won't scroll and scale with the workspace like the grid does.

### Request

Add an extension point that lets developers override the [Grid](https://github.com/google/blockly/blob/c8c4684d3beeec305fc52f3c4a7a3882455f537c/core/grid.ts#L23) class. I believe the [parsing code for grid options](https://github.com/google/blockly/blob/develop/core/options.ts#L312) would also need to be overridden; perhaps it could be moved to a static method on Grid rather than on Options.

### Alternatives considered

_No response_

### Additional context

In our [MakeCode for Minecraft editor](https://minecraft.makecode.com/#editor), we use an image as the background instead of a grid. Previously the code for this was in the microsoft/pxt-blockly fork but ever since we deprecated that repo we've lost the ability to do this in a clean manner.

As a temporary fix, I have implemented it with a [rather heinous monkey patch](https://github.com/microsoft/pxt/pull/10220).

Also, if this is a desirable change I would be willing to submit a PR!

riknoll:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8614

### Proposed Changes

Refactors the Grid class so that it can be overridden by a plugin. The static methods for [parsing grid options](https://github.com/google/blockly/blob/develop/core/options.ts#L312) and [creating the SVG pattern element](https://github.com/google/blockly/blob/3e4665e5b824041867b79f0cf8b657c5e6c32236/core/grid.ts#L216) have also been moved onto a new GridProvider class which can be passed in as a plugin in the BlocklyOptions object. The implementations of these functions are unchanged; they just got moved around.

Originally, I just wanted these to be static methods on the IGrid interface, but the typings for the registry's `getClass` functions don't really allow for static properties and I didn't want to tear all that up in this PR.

I would love some feedback on the typing of GridOptions here. My example use case for this is the [MakeCode Minecraft editor](https://minecraft.makecode.com/#editor) which overrides the grid with a background image that scrolls with the workspace. A plugin that provided that functionality would likely need to extend the GridOptions type with extra fields like the URL of the image, width/height, opacity, etc. However, the `grid` property of the BlocklyOptions object passed to inject is pretty strongly typed; the only way around it right now is to do a lot of casting to `any`.

### Reason for Changes

See the linked issue for reasoning!

### Test Coverage

Tested manually in the playground and by creating a test plugin locally. Happy to add unit tests, but I didn't see any for other extension points at a click glance.

### Additional Information

I can also add a plugin to blockly-samples that utilizes this change should this PR be approved.

### DIFF
--- a/core/interfaces/i_grid.ts
+++ b/core/interfaces/i_grid.ts
@@ -1,0 +1,116 @@
+import {BlocklyOptions} from '../blockly_options.js';
+import {GridOptions} from '../options.js';
+import {Coordinate} from '../utils';
+import type {IRegistrable} from './i_registrable.js';
+
+export interface IGrid {
+  /**
+   * Sets the spacing between the centers of the grid lines.
+   *
+   * This does not trigger snapping to the newly spaced grid. If you want to
+   * snap blocks to the grid programmatically that needs to be triggered
+   * on individual top-level blocks. The next time a block is dragged and
+   * dropped it will snap to the grid if snapping to the grid is enabled.
+   */
+  setSpacing(spacing: number): void;
+
+  /**
+   * Get the spacing of the grid points (in px).
+   *
+   * @returns The spacing of the grid points.
+   */
+  getSpacing(): number;
+
+  /** Sets the length of the grid lines. */
+  setLength(length: number): void;
+
+  /** Get the length of the grid lines (in px). */
+  getLength(): number;
+
+  /**
+   * Sets whether blocks should snap to the grid or not.
+   *
+   * Setting this to true does not trigger snapping. If you want to snap blocks
+   * to the grid programmatically that needs to be triggered on individual
+   * top-level blocks. The next time a block is dragged and dropped it will
+   * snap to the grid.
+   */
+  setSnapToGrid(snap: boolean): void;
+
+  /**
+   * Whether blocks should snap to the grid.
+   *
+   * @returns True if blocks should snap, false otherwise.
+   */
+  shouldSnap(): boolean;
+
+  /**
+   * Get the ID of the pattern element, which should be randomized to avoid
+   * conflicts with other Blockly instances on the page.
+   *
+   * @returns The pattern ID.
+   * @internal
+   */
+  getPatternId(): string;
+
+  /**
+   * Update the grid with a new scale.
+   *
+   * @param scale The new workspace scale.
+   * @internal
+   */
+  update(scale: number): void;
+
+  /**
+   * Move the grid to a new x and y position, and make sure that change is
+   * visible.
+   *
+   * @param x The new x position of the grid (in px).
+   * @param y The new y position of the grid (in px).
+   * @internal
+   */
+  moveTo(x: number, y: number): void;
+
+  /**
+   * Given a coordinate, return the nearest coordinate aligned to the grid.
+   *
+   * @param xy A workspace coordinate.
+   * @returns Workspace coordinate of nearest grid point.
+   *   If there's no change, return the same coordinate object.
+   */
+  alignXY(xy: Coordinate): Coordinate;
+}
+
+export interface IGridProvider extends IRegistrable {
+  /**
+   * Create the DOM for the grid described by options.
+   *
+   * @param rnd A random ID to append to the pattern's ID.
+   * @param gridOptions The object containing grid configuration.
+   * @param defs The root SVG element for this workspace's defs.
+   * @returns The SVG element for the grid pattern.
+   */
+  createDom(
+    rnd: string,
+    gridOptions: GridOptions,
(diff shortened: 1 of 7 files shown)


rachel-fenichel:
@gonfunko bump

core/interfaces/i_grid.ts
```diff
@@ -0,0 +1,116 @@
...
+  shouldSnap(): boolean;
+
+  /**
+   * Get the ID of the pattern element, which should be randomized to avoid
+   * conflicts with other Blockly instances on the page.
+   *
+   * @returns The pattern ID.
+   * @internal
+   */
+  getPatternId(): string;
```
gonfunko, line 54: Suggest removing this; it's quite low level, and its only caller in core is in WorkspaceSvg, which can be eliminated easily if we follow the approach I outlined in my top-level comment.

core/interfaces/i_grid.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Update the grid with a new scale.
+   *
+   * @param scale The new workspace scale.
+   * @internal
+   */
+  update(scale: number): void;
```
gonfunko, line 62: Let's make this `setScale()` and `getScale()` for consistency. The built-in `Grid` implementation can still call update() to adjust itself in response to changes in its configuration, but that doesn't need to be part of the interface (or require scale as an argument).

core/interfaces/i_grid.ts
```diff
@@ -0,0 +1,116 @@
...
+   * @returns True if blocks should snap, false otherwise.
+   */
+  shouldSnap(): boolean;
+
+  /**
+   * Get the ID of the pattern element, which should be randomized to avoid
+   * conflicts with other Blockly instances on the page.
+   *
+   * @returns The pattern ID.
+   * @internal
```
gonfunko, line 52: Throughout, I don't think we want @internal annotations in the interface; if we're committing it to the interface, that makes it public API.

gonfunko (commented):
Thank you for this change, and apologies for taking a while on the review! Broadly speaking making the grid swappable seems great, but I think there are some structural issues we might want to address while we're creating a new public API. In particular, I think the way things currently work in core is a bit silly, and the existing indirection complicates making this pluggable. However, I think the implementation can be adjusted to make the API cleaner without too much trouble.

As things stand, the flow is:
1. Grid options can be specified during the call to `inject()`
2. `inject()` calls `createDom()` (in inject.ts), which calls `Grid.createDom()` and stores the resulting SVG pattern on the `options` object
3. A `WorkspaceSvg` is constructed, which creates a new `Grid` object from the grid options specified in the call to `inject()` and the SVG pattern created and stashed on the `options` object in step 2

Instead, I think we want to:
1. Continue to specify grid options during the call to inject, and have those parsed out in options.ts as they currently are
2. Do nothing grid-related in the `createDom()` in inject.ts.
3. Have `WorkspaceSvg` continue to initialize a Grid in its constructor, but (a) get the class from the registry and (b) pass no arguments to the Grid constructor
4. In `WorkspaceSvg.createDom()`:
  - Call `this.grid.createDom()`, which is now an instance method, passing only the parsed grid options
  - Assign a random ID to the element returned from that method
  - Create a <defs> element, add the returned element as a child, and add the <defs> to the injection div (this currently happens in inject.ts)
  
This has the following benefits:
* Removes the need for Options.gridPattern, which is the only field there that isn't really a configuration option as such
* Removes the need for the `rnd`, `defs`, and (in v12) `injectionDiv` arguments to `Grid.createDom()`, which are quite low-level and require the implementation to follow a lot of contracts. Instead, implementers need only concern themselves with creating and returning an SVG pattern, and Blockly will handle its own internal bookkeeping.
* Makes `Grid.createDom()` an instance method, which works easier with interfaces as you noted
* Eliminates the need for `IGridProvider`
* Allows developers to add their own configuration methods to their implementations of IGrid, which they can use to configure properties, e.g. the image URL, size, opacity etc. that you noted. IMO this is strongly preferable to tacking arbitrary properties on to the `grid` injection option; that has the typing problems you described, and broadly speaking we don't really want to add options there in general. With this approach, after injection, developers can just call `workspace.getGrid().setWhateverConfigOption()`; those methods won't be part of the `IGrid` interface, but since they just injected their own version they can confidently cast the return type of `getGrid()` to their concrete implementation.

PR #8726, opened 2025-01-09
fix: Fix bug that preventing scrolling menu items into view.

gonfunko:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes #8724

### Proposed Changes
This PR fixes scrolling menu items into view when selected via the arrow keys. It also deprecates a number of now-unused, copied-from-Closure, wrapper-workarounds-for-browser-functionality-missing-from-IE-6 functions.

PR #8765, opened 2025-02-12
fix: Fix menu scrolling.

### LINKED ISSUE #8762: `scrollIntoView()` scrolls elements outside of menus
### Check for duplicates

### Description

In #8726 I updated menus to use `scrollIntoView()` to scroll to display items when the selection exceeded the ends of the menu. This works, but `scrollIntoView()` also attempts to scroll parent elements to better display the target element, which can cause the entire page to shift, particularly in cases where Blockly is embedded in an iframe. This behavior is apparently [under active reconsideration in the standards process](https://github.com/w3c/csswg-drafts/pull/11673), but in the meantime we should replace it with an alternative that doesn't have this side effect.

gonfunko:
## The basics

### Resolves

Fixes #8762

### Proposed Changes
This PR reverts #8726, which attempted to fix menu scrolling by using `scrollIntoView()`. Instead, the original scrolling utility function is reintroduced and used; the original issue was apparently that an intermediate div got introduced, so the parent element that was passed to the scrolling utility function wasn't actually the scrollable element.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -29,6 +29,7 @@ import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
 import * as parsing from './utils/parsing.js';
 import * as utilsString from './utils/string.js';
+import * as style from './utils/style.js';
 import {Svg} from './utils/svg.js';
 
 /**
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):
The original change said "This PR fixes scrolling menu items into view when selected via the arrow keys. It also deprecates a number of now-unused, copied-from-Closure, wrapper-workarounds-for-browser-functionality-missing-from-IE-6 functions."

Does this reversion mean that those functions were actually being useful instead of just working around IE6?

gonfunko:
Yes; I knocked over Chesterton's fence. The utility function handles scrolling, which I thought was obsolete given the existence of `scrollIntoView()`, without realizing that that was extremely aggressive.

iCloudys:
When the options of field_dropdown exceed 12, after I expand it, it automatically scrolls to the bottom, making it impossible for me to select other options.
#8804

PR #8767, opened 2025-02-13
feat!: Added support for separators in menus.

### LINKED ISSUE #8752: Add support for a dropdown menu separator
### Check for duplicates

### Problem

MakeCode inserts an empty dropdown item with role="separator" and custom styling into some dropdowns to separate values from actions. This dropdown item is not treated any differently as far as keyboard navigation is concerned. As a result, when moving up and down inside the dropdown, the menu separator takes the selection, albeit in a non-obvious way. See the video below for an example of this. Keyboard navigation should only visit the 4 'real' items in this dropdown, however, it also visits the menu separator as a 5th item.

https://github.com/user-attachments/assets/dd4118f0-b221-42d4-b5d2-9d7a438f6f1f

### Request

Please can you consider adding support for menu separators in dropdowns? Keyboard handling and aria roles for these separators can then be implemented correctly and centralized in Blockly. An alternative might be the ability to pass options for a menu item that could include removing it from other keyboard navigable items in some way.

### Alternatives considered

_No response_

### Additional context

I have opened a [PR in MakeCode](https://github.com/microsoft/pxt/pull/10364) in an attempt to fix this issue, however, it is very much a workaround added to an existing workaround.

gonfunko:
## The basics

### Resolves

Fixes #8752

### Proposed Changes
This PR adds support for separators in all Blockly menus, including contextual menus on the workspace, blocks and comments, and in dropdown menus spawned from `FieldDropdown`.

For contextual menus, separators can be added via the context menu registry in a similar manner to normal menu items:

```typescript
  const separator: RegistryItem = {
    separator: true,
    scopeType: ContextMenuRegistry.ScopeType.WORKSPACE,
    id: 'whatever-id-goes-here',
    weight: 2,
  };
  ContextMenuRegistry.registry.register(separator);
```

For `FieldDropdown`, separators can be added via JSON (or a generator function) by using the string literal `separator`:

```typescript
{
  'type': 'field_dropdown',
  'name': 'SOME_DROPDOWN_FIELD',
  'options': [
    ['Normal menu item', 'normal'],
    ['Other menu item', 'other'],
    'separator',
    ['Different menu item', 'different'],
  ],
},
```

### Breaking changes
In general existing code should continue to work, but in order to accomodate typings `ContextMenuOption` and `RegistryItem` are now `type`s instead of `interface`s. Core and our documentation suggest using these as object literals, as in the example above in this PR description, which will continue to work as-is, but if users have a class that `implements` either of these interfaces, the `implements` will need to be removed.

### DIFF
--- a/core/contextmenu_registry.ts
+++ b/core/contextmenu_registry.ts
@@ -87,21 +87,37 @@ export class ContextMenuRegistry {
     const menuOptions: ContextMenuOption[] = [];
     for (const item of this.registeredItems.values()) {
       if (scopeType === item.scopeType) {
-        const precondition = item.preconditionFn(scope);
-        if (precondition !== 'hidden') {
+        let menuOption:
+          | ContextMenuRegistry.CoreContextMenuOption
+          | ContextMenuRegistry.SeparatorContextMenuOption
+          | ContextMenuRegistry.ActionableContextMenuOption;
+        menuOption = {
+          scope,
+          weight: item.weight,
+        };
+
+        if (item.separator) {
+          menuOption = {
+            ...menuOption,
+            separator: true,
+          };
+        } else {
+          const precondition = item.preconditionFn(scope);
+          if (precondition === 'hidden') continue;
+
           const displayText =
             typeof item.displayText === 'function'
               ? item.displayText(scope)
               : item.displayText;
-          const menuOption: ContextMenuOption = {
+          menuOption = {
+            ...menuOption,
             text: displayText,
-            enabled: precondition === 'enabled',
             callback: item.callback,
-            scope,
-            weight: item.weight,
+            enabled: precondition === 'enabled',
           };
-          menuOptions.push(menuOption);
         }
+
+        menuOptions.push(menuOption);
       }
     }
     menuOptions.sort(function (a, b) {
@@ -134,9 +150,18 @@ export namespace ContextMenuRegistry {
   }
 
   /**
-   * A menu item as entered in the registry.
+   * Fields common to all context menu registry items.
    */
-  export interface RegistryItem {
+  interface CoreRegistryItem {
+    scopeType: ScopeType;
+    weight: number;
+    id: string;
+  }
+
+  /**
+   * A representation of a normal, clickable menu item in the registry.
+   */
+  interface ActionableRegistryItem extends CoreRegistryItem {
     /**
      * @param scope Object that provides a reference to the thing that had its
      *     context menu opened.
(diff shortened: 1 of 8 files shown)


core/contextmenu_registry.ts
```diff
@@ -134,27 +150,57 @@ export namespace ContextMenuRegistry {
...
+  interface CoreRegistryItem {
+    scopeType: ScopeType;
+    weight: number;
+    id: string;
+  }
+
+  /**
+   * A representation of a normal, clickable menu item in the registry.
+   */
+  interface ActionableRegistryItem extends CoreRegistryItem {
```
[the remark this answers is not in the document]
rachel-fenichel: Same for the contextmenuitem interface below.

core/field_dropdown.ts
```diff
@@ -748,28 +754,28 @@ export class FieldDropdown extends Field<string> {
...
+      if (!Array.isArray(option) && option !== FieldDropdown.SEPARATOR) {
         foundError = true;
         console.error(
-          `Invalid option[${i}]: Each FieldDropdown option must be an array.
-          Found: ${tuple}`,
+          `Invalid option[${i}]: Each FieldDropdown option must be an array or
+          the string literal 'separator'. Found: ${option}`,
         );
-      } else if (typeof tuple[1] !== 'string') {
+      } else if (typeof option[1] !== 'string') {
```
rachel-fenichel, line 764: Won't this fail if the option is just `FieldDropdown.SEPARATOR` instead of a tuple?

core/field_dropdown.ts
```diff
@@ -35,14 +36,10 @@ import {Svg} from './utils/svg.js';
  * Class for an editable dropdown field.
  */
 export class FieldDropdown extends Field<string> {
-  /** Horizontal distance that a checkmark overhangs the dropdown. */
```
rachel-fenichel, line 38: Why remove these two?

core/menu_separator.ts
```diff
@@ -0,0 +1,37 @@
...
+    aria.setRole(this.element, aria.Role.SEPARATOR);
+
+    return this.element;
+  }
+
+  /**
+   * Disposes of this separator.
+   */
+  dispose() {
+    this.element = null;
```
rachel-fenichel, line 35: Also remove from the document?

core/field_dropdown.ts
gonfunko: They're unreferenced in the entire codebase. I suppose this is technically a breaking change, but they feel very implementation detail-y. Happy to revert if you want though.

core/contextmenu_registry.ts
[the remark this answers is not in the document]
gonfunko: Done.

core/menu_separator.ts
gonfunko: Done.

rachel-fenichel (approved):

PR #8779, opened 2025-02-18
fix: Fix bug that prevented editing workspace comments on Firefox.

### LINKED ISSUE #8768: Workspace comments aren't editable in Firefox
### Check for duplicates

### Description

If you create a workspace comment in Firefox, it's not possible to edit or select its text. Works fine in other browsers, and block comments work fine in Firefox, with seemingly the same DOM structure.

### Browsers

Firefox desktop

gonfunko:
## The basics

### Resolves

Fixes #8768

### Proposed Changes
This PR fixes a bug that prevented editing the text of workspace comments on Firefox.

### DIFF
--- a/core/comments/rendered_workspace_comment.ts
+++ b/core/comments/rendered_workspace_comment.ts
@@ -208,8 +208,15 @@ export class RenderedWorkspaceComment
   private startGesture(e: PointerEvent) {
     const gesture = this.workspace.getGesture(e);
     if (gesture) {
-      gesture.handleCommentStart(e, this);
-      this.workspace.getLayerManager()?.append(this, layers.BLOCK);
+      const textArea = this.view.getTextArea();
+      if (e.target === textArea) {
+        // If the text area was the focus, don't allow this event to bubble up
+        // and steal focus away from the editor/comment.
+        e.stopPropagation();
+      } else {
+        gesture.handleCommentStart(e, this);
+        this.workspace.getLayerManager()?.append(this, layers.BLOCK);
+      }
       common.setSelected(this);
     }
   }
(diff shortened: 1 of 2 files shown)


core/comments/rendered_workspace_comment.ts
```diff
@@ -208,8 +208,15 @@ export class RenderedWorkspaceComment
   private startGesture(e: PointerEvent) {
     const gesture = this.workspace.getGesture(e);
     if (gesture) {
-      gesture.handleCommentStart(e, this);
-      this.workspace.getLayerManager()?.append(this, layers.BLOCK);
+      const textArea = this.view.getTextArea();
```
rachel-fenichel, line 211: Can you do this with `isTargetInput` from `browser_events.ts` instead?

gonfunko: Ooh, yes! Much nicer, thank you!

core/comments/rendered_workspace_comment.ts
```diff
@@ -208,8 +208,14 @@ export class RenderedWorkspaceComment
...
     const gesture = this.workspace.getGesture(e);
     if (gesture) {
-      gesture.handleCommentStart(e, this);
-      this.workspace.getLayerManager()?.append(this, layers.BLOCK);
+      if (browserEvents.isTargetInput(e)) {
+        // If the text area was the focus, don't allow this event to bubble up
+        // and steal focus away from the editor/comment.
+        e.stopPropagation();
+      } else {
+        gesture.handleCommentStart(e, this);
```
rachel-fenichel, line 216: Next question: why is this not already fixed by the other uses of `isTargetInput` in `inject.ts` and `gesture.ts`?

gonfunko: `handleCommentStart()` doesn't ever call `doStart()`, where `isTargetInput()` is checked; `handleWsStart` does, but by the time that's called the event has already bubbled up to the workspace and stolen focus away from the textarea, so it's too late. Likewise, inject.ts only checks it for keydown and contextmenu events, not clicks.

rachel-fenichel (approved):

PR #8781, opened 2025-02-19
fix: Fixes #8764 by moving the event grouping calls up to dragger.ts

### LINKED ISSUE #8764: Undo seems unstable with procedure blocks
### Check for duplicates

### Description

[This should maybe be part of #8560 but I didn't find an issue that exactly matches what I saw]

When I try to undo a move of a group of blocks that include a procedure call block, sometimes the procedure call block disappears, sometimes its parent disappears, sometimes everything is fine.
I didn't manage to find the exact conditions that trigger the bug :(

### Reproduction steps

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/d41a6bc3-c6ce-4834-a97a-7b95f3623efc

### Browsers

Chrome desktop

RoboErikG:
## The basics

### Resolves

Fixes #8764 

### Proposed Changes

Moves the event group management up to dragger.ts instead of being handled in the various *_drag_strategy.ts files. Conceptually, this makes it the responsibility of the top level dragger to manage event groups and ensures that the group wraps all side effects of the drag.

In vanilla Blockly this shouldn't have any effect, but is necessary for the multi-select plugin used by AppInventor.

Also fixes a bug where snapping to the grid adds a move event that isn't part of the group.

### Reason for Changes

AppInventor has a multi-select plugin that creates multiple BlockDragStrategy instances. This requires that endDrag be called on each one, which was causing the event group to be reset after the first one. Moving it up a level ensures all of the block drags finish before the event group is reset.

### Test Coverage

[Video of manual testing](https://github.com/user-attachments/assets/6091a93b-26f3-41d6-9918-a55f4791525d)

### Documentation

No documentation changes, but we should include it in the release notes that the behavior has changed and anyone overriding dragger.ts should make this change if they aren't calling through to the super.

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -62,8 +62,8 @@ export class BlockDragStrategy implements IDragStrategy {
    */
   private dragOffset = new Coordinate(0, 0);
 
-  /** Was there already an event group in progress when the drag started? */
-  private inGroup: boolean = false;
+  /** Used to persist an event group when snapping is done async. */
+  private originalEventGroup = '';
 
   constructor(private block: BlockSvg) {
     this.workspace = block.workspace;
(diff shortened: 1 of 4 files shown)


RoboErikG:
The failing test is the flakey connection test which was fixed in develop but hasn't been merged into V12 yet.

gonfunko (approved):

PR #8785, opened 2025-02-24
fix: Disallow and ignore x and y attributes for blocks in toolbox definitions.

### LINKED ISSUE #6280: Built-in flyouts position blocks with x and y values incorrectly
**Describe the bug**

If you define a toolbox where the block info has x and y properties, the blocks will be positioned incorrectly.

**To Reproduce**

1. Create a toolbox definition.
2. Add blocks to it
3. Give those blocks x and y properties (That are non-zero)
4. Observe how they are positioned incorrectly.

**Expected behavior**

The toolbox should ignore the x and y properties on blocks because it has its own gap/separator logic for positioning blocks.

**Additional context**

- The [BlockInfo type](https://github.com/google/blockly/blob/fa58cbaaf7456ab227c2214af5dd0e7025dc731c/core/utils/toolbox.ts#L19) should be updated to not include X and Y.
- Blockly has two kinds of flyouts built-in, horizontal and vertical. You should update the layout logic in both of them.
- The flyouts should be changed to use [moveTo](https://github.com/google/blockly/blob/fa58cbaaf7456ab227c2214af5dd0e7025dc731c/core/block_svg.ts#L414) [instead of moveBy](https://github.com/google/blockly/blob/61322294da70c770b32abb28b9d695af75f61a16/core/flyout_vertical.js#L250)

**Tips for Getting Started**
We have a lot of information about how to get started [here](https://developers.google.com/blockly/guides/contribute/get-started) including details about our commit message style and how to open a PR.

gonfunko:
## The basics

### Resolves

Fixes #6280

### Proposed Changes
This PR removes `x` and `y` from `BlockInfo`, and removes them if set before loading and laying out blocks in a flyout. Previously, these would cause blocks to be positioned incorrectly in the flyout, since the flyout repositions its contents relative to one another but assumes their initial origin is (0, 0). The XML codepath was not subject to this issue.

### DIFF
--- a/core/block_flyout_inflater.ts
+++ b/core/block_flyout_inflater.ts
@@ -101,8 +101,9 @@ export class BlockFlyoutInflater implements IFlyoutInflater {
       ) {
         blockDefinition['disabledReasons'] = [MANUALLY_DISABLED];
       }
-      // These fields used to be allowed and may still be present, but should
-      // now be ignored.
+      // These fields used to be allowed and may still be present, but are
+      // ignored here since everything in the flyout should always be laid out
+      // linearly.
       if ('x' in blockDefinition) {
         delete blockDefinition['x'];
       }
--- a/core/block_flyout_inflater.ts
+++ b/core/block_flyout_inflater.ts
@@ -101,6 +101,14 @@ export class BlockFlyoutInflater implements IFlyoutInflater {
       ) {
         blockDefinition['disabledReasons'] = [MANUALLY_DISABLED];
       }
+      // These fields used to be allowed and may still be present, but should
+      // now be ignored.
+      if ('x' in blockDefinition) {
+        delete blockDefinition['x'];
+      }
+      if ('y' in blockDefinition) {
+        delete blockDefinition['y'];
+      }
       block = blocks.appendInternal(blockDefinition as blocks.State, workspace);
     }
 
--- a/core/utils/toolbox.ts
+++ b/core/utils/toolbox.ts
@@ -24,8 +24,6 @@ export interface BlockInfo {
   disabledReasons?: string[];
   enabled?: boolean;
   id?: string;
-  x?: number;
-  y?: number;
   collapsed?: boolean;
   inline?: boolean;
   data?: string;


core/utils/toolbox.ts
```diff
@@ -24,8 +24,6 @@ export interface BlockInfo {
   disabledReasons?: string[];
   enabled?: boolean;
   id?: string;
-  x?: number;
-  y?: number;
```
BenHenning, line 28: What does member removal mean for backward compatibility for users upgrading to v12? I feel that I don't have a good enough context to know this already.

core/block_flyout_inflater.ts
```diff
@@ -101,6 +101,14 @@ export class BlockFlyoutInflater implements IFlyoutInflater {
       ) {
         blockDefinition['disabledReasons'] = [MANUALLY_DISABLED];
       }
+      // These fields used to be allowed and may still be present, but should
+      // now be ignored.
```
BenHenning, line 105: Is there anywhere else in the codebase that may use BlockInfo? I'm curious if there's a better location for this (like deserialization) that may be more comprehensive, or is this really the only place where it does and will ever matter?

BenHenning (commented):
Thanks @gonfunko! Just had a couple of comments looking for clarity.

Separately, can you confirm that the actual positioning logic no longer needs to be updated per #6280? It seems that it was fixed in #7333 but just wanted to double check.

gonfunko (commented):
Confirmed the positioning logic needs no change after this; it assumes items are initially at (0, 0) and moves them relative to that point, which was problematic when they were starting at not-(0, 0) due to the x and y properties being respected, but is no longer an issue.

core/block_flyout_inflater.ts
BenHenning: Ah I didn't actually fully have that context, then. Could you maybe update the comment here to perhaps explain that exact thing? Maybe something like:

```ts
      // These fields used to be allowed and may still be present, but should
      // now be ignored for the flyout since everything should always be laid
      // out linearly.
```

core/utils/toolbox.ts
BenHenning: Makes sense--thanks for clarfiying!

BenHenning:
Thanks @gonfunko. Just had one suggestion on clarifying a comment for clearer context, otherwise the PR LGTM.

core/block_flyout_inflater.ts
gonfunko: Done.

BenHenning (approved):

PR #8800, opened 2025-03-11
Revert "fix!: Remove the blocklyMenuItemHighlight CSS class and use the hover"

gonfunko:
Reverts google/blockly#8536

Using :hover is not sufficient, because menu items can also be focused by keyboard navigation, but need to be able to be targeted by a selector in order to be styled.

### DIFF
--- a/core/menuitem.ts
+++ b/core/menuitem.ts
@@ -12,6 +12,7 @@
 // Former goog.module ID: Blockly.MenuItem
 
 import * as aria from './utils/aria.js';
+import * as dom from './utils/dom.js';
 import * as idGenerator from './utils/idgenerator.js';
 
 /**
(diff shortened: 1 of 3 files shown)


core/menuitem.ts
```diff
@@ -175,6 +177,25 @@ export class MenuItem {
...
 
+  /**
+   * Highlights or unhighlights the component.
+   *
+   * @param highlight Whether to highlight or unhighlight the component.
+   * @internal
+   */
+  setHighlighted(highlight: boolean) {
+    this.highlight = highlight;
+    const el = this.getElement();
```
RoboErikG, line 188: I think I'm missing something, but why is the highlight class the only one you have to add or remove manually?  checked and enabled also set the class during creation but not when the setter is called.

RoboErikG: Tracking issue is fine

RoboErikG (approved):

core/menuitem.ts
gonfunko: Filed https://github.com/google/blockly/issues/8801

PR #8811, opened 2025-03-19
chore: Try another workflow permission.

gonfunko:
## The basics
This PR adds the `pull-requests: write` permission to assign_reviewers.yml. The docs say it needs this or `issues: write`, but I'm suspicious it may need both.

### DIFF
--- a/.github/workflows/assign_reviewers.yml
+++ b/.github/workflows/assign_reviewers.yml
@@ -9,14 +9,14 @@ name: Assign requested reviewers
 # N.B.: Runs with a read-write repo token.  Do not check out the
 # submitted branch!
 on:
-  pull_request:
+  pull_request_target:
     types: [review_requested]
-
-permissions: write-all
 
 jobs:
   requested-reviewer:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - name: Assign requested reviewer
         uses: actions/github-script@v7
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

rachel-fenichel (changes requested):
Please fix.

BenHenning:
I'm assuming you've already seen this: https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests.

https://docs.github.com/en/rest/issues/assignees?apiVersion=2022-11-28#add-assignees-to-an-issue suggests that the current permissions _should_ work. 

I'm having some difficulty backing this up, but can you try moving the permissions to be top-level instead of per-job? It shouldn't make a difference, but it would be nice to double check.

BenHenning:
I've tested this in https://github.com/BenHenning/blockly/blob/develop2/.github/workflows/assign_reviewers.yml using https://github.com/BenHenning/blockly/pull/1. You should be able to use `pull_request_target` with _just_ `pull-requests: write`. The documentation _is_ wrong in that it says `issues` OR `pull-requests` is needed. One **is** needed but it depends on the context.

Because of the nature of `pull_request_target` this won't actually pass until it's merged into develop (as that permission forces using the merged-in version even if you change it here per security reasons) which is why I tested against a base branch instead of within the PR.

BenHenning (approved):
Thanks @gonfunko! From my testing, this _should_ fix the check once merged in. :)

rachel-fenichel (approved):

BenHenning:
Also: this will need to be done on the v12 branch, too, if we want CI to not fail there.

PR #8830, opened 2025-03-28
fix: Clear the cursor when deleting the block it's on.

### LINKED ISSUE #8793: Clear cursor state in block deletion
### Check for duplicates

### Problem

_No response_

### Request

If the block that is deleted has a cursor pointing to it (at the workspace level), tell the cursor to clear itself from the block's `dispose` function.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8793

### Proposed Changes
This PR clears the cursor in BlockSvg.dispose() if the cursor is pointing to the block that is being deleted.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -808,6 +808,16 @@ export class BlockSvg
   override dispose(healStack?: boolean, animate?: boolean) {
     this.disposing = true;
 
+    if (
+      this.workspace
+        .getMarkerManager()
+        .getCursor()
+        ?.getCurNode()
+        ?.getLocation() === this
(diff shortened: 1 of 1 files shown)


maribethb:
Sorry Aaron, I was surprised by this behavior and double-checked with Rachel and she confirmed that the original issue is wrong. Rather than clear the node, we want to set it to something that makes sense (we can argue about what exactly that is later, but for now it should be whatever the behavior is in the experiment, i.e. setting it to the workspace or the previous block or connection). I added this info to #8793 as well.

PR #8833, opened 2025-03-31
Feature: Add a dropdown options setter

### LINKED ISSUE #1288: Create an API for setting menuGenerator_ on field_dropdown after creation
Currently, the "right" way to add a [dynamic menu](https://developers.google.com/blockly/guides/create-custom-blocks/dropdown-menus#dynamic_menu) to a Field is to pass the function in on creation. That's fine when your block definitions are in javascript, but not possible for JSON. You can currently create an extension, and create a new field to add to the block/replace an existing field when the block is created, but that's a little strange, because the structure of the resulting block won't really match the definition.

I think we need an API to set menuGenerator_ after creation, so it can be set on an extension without creating a new field.

RoboErikG:
## The basics

### Resolves

Fixes #1288 

### Proposed Changes

Adds a setOptions function to field_dropdown to allow changing the set of options or the generator for options after creation. 

### Reason for Changes

This makes it easier to define blocks with dynamic dropdowns in JSON.

### Test Coverage

Added a test to verify the selected option is updated after changing the set of options.

### Documentation

We should consider updating the [dropdown field docs](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#dynamic_menu) to include this as a simpler option for the JSON definitions.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -415,10 +415,8 @@ export class FieldDropdown extends Field<string> {
     } else {
       this.menuGenerator_ = menuGenerator;
     }
-    /**
-     * The currently selected option. The field is initialized with the
-     * first option selected.
-     */
+    // The currently selected option. The field is initialized with the
+    // first option selected.
     this.selectedOption = this.getOptions(false)[0];
     this.setValue(this.selectedOption[1]);
   }
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -136,26 +136,11 @@ export class FieldDropdown extends Field<string> {
     // If we pass SKIP_SETUP, don't do *anything* with the menu generator.
     if (menuGenerator === Field.SKIP_SETUP) return;
 
-    if (Array.isArray(menuGenerator)) {
-      this.validateOptions(menuGenerator);
-      const trimmed = this.trimOptions(menuGenerator);
-      this.menuGenerator_ = trimmed.options;
-      this.prefixField = trimmed.prefix || null;
-      this.suffixField = trimmed.suffix || null;
-    } else {
-      this.menuGenerator_ = menuGenerator;
-    }
-
-    /**
-     * The currently selected option. The field is initialized with the
-     * first option selected.
-     */
-    this.selectedOption = this.getOptions(false)[0];
+    this.setOptions(menuGenerator);
 
     if (config) {
       this.configure_(config);
     }
-    this.setValue(this.selectedOption[1]);
     if (validator) {
       this.setValidator(validator);
     }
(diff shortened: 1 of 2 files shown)


core/field_dropdown.ts
```diff
@@ -414,6 +399,30 @@ export class FieldDropdown extends Field<string> {
...
+    if (Array.isArray(menuGenerator)) {
+      this.validateOptions(menuGenerator);
+      const trimmed = this.trimOptions(menuGenerator);
+      this.menuGenerator_ = trimmed.options;
+      this.prefixField = trimmed.prefix || null;
+      this.suffixField = trimmed.suffix || null;
+    } else {
+      this.menuGenerator_ = menuGenerator;
+    }
+    /**
```
BenHenning, line 418: Nit: this shouldn't be a doc comment. I know you copied it over, but might as well fix it while here.

tests/mocha/field_dropdown_test.js
```diff
@@ -194,6 +194,14 @@ suite('Dropdown Fields', function () {
       this.field.setValue('B');
       assertFieldValue(this.field, 'B', 'b');
     });
+    test('After options change', function () {
```
BenHenning, line 197: I have a few suggestions:
1. Does it make sense to put the tests in the 'setValue' suite? I'd actually expect a new suite specifically for the new 'setOptions' function.
1. It would be nice if this test also described what the expected outcome was. It currently reads as "setValue After options change". Perhaps something like (per point (1) above): 'setOptions With array updates options' (though this could probably be clearer).
2. It seems like there are a bunch of other tests that might be useful to add:
    - Passing a generator instead of an array.
    - Passing options that would fail validateOptions.
    - Different scenarios of trimOptions (e.g. with prefix or suffix null, or other relevant behaviors).
    - Edge cases like passing an empty list.
    - Demonstrating that overriding options on an already filled list correctly updates it, and reset the selected item (as the documentation mentions).

BenHenning (commented):
Thanks @RoboErikG! Had some thoughts and suggestions, PTAL.

tests/mocha/field_dropdown_test.js
RoboErikG: Made it a separate suite and added tests for the generator, empty list, and a set of trimmable values.
For the other suggested tests:

- Most of the validateOptions paths are already tested by the constructorSuiteTests, so adding more here doesn't increase coverage.
- All of the valid calls to updateOptions demonstrate that it correctly updates since I'm checking that it was set to the first item in the new list.

BenHenning (approved):
Thanks, LGTM @RoboErikG!

PR #8834, opened 2025-03-31
refactor!: Backport LineCursor to core.

### LINKED ISSUE #8822: Make the `LineCursor` the only cursor implementation in core
### Check for duplicates

### Problem

_No response_

### Request

- [ ] Move the `LineCursor` implementation from keyboard-experimentation into core, replacing the contents of the existing `Cursor` class. 
- [ ] Remove extra installation code carried over from `LineCursor`
- [ ] Retain any extra code needed for the existing `super` calls

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #8823: Move `scrollBoundsIntoView` into core
### Check for duplicates

### Problem

_No response_

### Request

Move `scrollBoundsIntoView` to core from its current location in `workspace_utilities.ts` in the keyboard-experimentation code.

Call as needed from the cursor implementation in core (see #8822).

Make this function internal, not public.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #8822 
* Fixes #8823

### Proposed Changes
This PR migrates the `LineCursor` class from the blockly-keyboard-experimentation repo back to core, replacing the old `Cursor` class, which has now been removed.

### Breaking Changes
* `Cursor` has been removed in favor of `LineCursor`. In addition to being a different class (that takes a workspace as a constructor argument) the navigation behavior is also different. Developers may register a different cursor class in its place if need be.


core/workspace_svg.ts
```diff
@@ -2541,6 +2541,69 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
+   * In the case of small workspaces/large bounds, this function prioritizes
+   * getting the top left corner of the bounds into view. It also adds some
+   * padding around the bounds to allow the element to be comfortably in view.
+   *
+   * @internal
+   * @param bounds A rectangle to scroll into view, as best as possible.
+   */
+  scrollBoundsIntoView(bounds: Rect) {
+    if (Gesture.inProgress()) {
+      // This can cause jumps during a drag and it only suited for keyboard nav.
```
RoboErikG, line 2557: nit: s/it/is

core/workspace_svg.ts
```diff
@@ -2541,6 +2541,69 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
+      bounds.bottom <= viewport.bottom
+    ) {
+      // Do nothing if the block is fully inside the viewport.
+      return;
+    }
+
+    // Add some padding to the bounds so the element is scrolled comfortably
+    // into view.
+    bounds = bounds.clone();
+    bounds.top -= 10;
```
RoboErikG, line 2583: nit: magic number. Consider making the padding a param or at least a constant somewhere.

core/workspace_svg.ts
```diff
@@ -2541,6 +2541,69 @@ export class WorkspaceSvg extends Workspace implements IASTNodeLocationSvg {
...
+    bounds.left -= 10;
+    bounds.right += 10;
+
+    let deltaX = 0;
+    let deltaY = 0;
+
+    if (bounds.left < viewport.left) {
+      deltaX = viewport.left - bounds.left;
+    } else if (bounds.right > viewport.right) {
+      deltaX = viewport.right - bounds.right;
```
RoboErikG, line 2594: If the top left is within the viewport but the bounds are larger than the viewport will this jump to the bottom right of the bounds? I think you need a Math.min(bounds.left - viewport.left, viewport.right - bounds.right) to make sure it doesn't scroll the left boundary outside the viewport.

Same for the top one below.

core/keyboard_nav/line_cursor.ts
```diff
@@ -0,0 +1,680 @@
...
+import {ASTNode} from './ast_node.js';
+import {Marker} from './marker.js';
+
+/** Options object for LineCursor instances. */
+export type CursorOptions = {
+  /**
+   * Can the cursor visit all stack connections (next/previous), or
+   * (if false) only unconnected next connections?
+   */
+  stackConnections: boolean;
```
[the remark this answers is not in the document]
RoboErikG: Maybe file an issue to make sure it gets done before V12 is released? Changing it after release would be slightly annoying.

core/workspace_svg.ts
RoboErikG: I don't think setting it to zero will work in all cases. If the bounds are bigger than the viewport and are completely to the right of the viewport you wouldn't scroll them into view at all. I think you need the Math.min version to catch all the edge cases.

RoboErikG (approved):

core/keyboard_nav/line_cursor.ts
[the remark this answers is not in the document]
gonfunko: Done: #8856

core/workspace_svg.ts
gonfunko: The Math.min() version suffers from the same issue; since this is a bit of an edge case and the method is `@internal`, I filed #8855 to figure out the behavior we want and adjust this if needed. I also reverted the set-to-zero approach.

PR #8835, opened 2025-03-31
fix!: Tighten and correct typings on ASTNode

### LINKED ISSUE #8827: ASTNode createXXXNode return type/null handling
The `createXXXNode` static methods on `ASTNode` have signatures and behaviour like this:

```
  static createBlockNode(block: Block): ASTNode | null {
    if (!block) {
      return null;
    }
    return new ASTNode(ASTNode.types.BLOCK, block);
  }
```

So the types do not permit a null block, but the code accounts for the possibility and returns null. The return type accommodates this.

The keyboard nav plugin is full of code that takes reasonable care to get e.g. a block and then tries to convert it to an ASTNode. It then has to non-null assert the result except where it passes it directly to `setCurNode` (which fortunately accepts the null we know it not to be in practice).

@rachel-fenichel was concerned that there might be quite a few routes for null into this code and so we might need to change a lot if we changed the signature. But until the recent lint changes all the call sites were null asserting the result.

I'd be tempted to change the signature to

```
  static createBlockNode(block: Block): ASTNode {
    if (!block) {
      throw new Error("Null block"); // If we want to be defensive.
    }
    return new ASTNode(ASTNode.types.BLOCK, block);
  }
```

Or if we really want the current behaviour then you can model it more verbosely like in [this example](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgLIE8ByB7AJigbwCgBfIomAVxATGGxGQSgjkh3wAoA3ALmQDOYKKADmASn4YOEANwVqteo2at2eCD34hKAG12TkO-fKo06DJizYQZWwcLHIAPkb0GpWDS7f7kxZGRgGGQeZABeSN8DfyJAwJYwSihGY115QLIEiCSU-zIyIgB6IuQwdAAHCCIEBiEggQBpEGwAdxBMdwirNVsNTjTxGrqwBua2joZOv3CemzsAIhhsbAWhoA).

This isn't a big issue: now `setCurNode` accepts null there are only 3 ignores for the non-null lints that result (see [lint issue](https://github.com/google/blockly-keyboard-experimentation/issues/345)). I just wanted to share my thoughts because the signature still feels weird to me.

gonfunko:
## The basics

### Resolves

Fixes #8827

### Proposed Changes
This PR updates most of the static `ASTNode.create*Node()` methods to not check for null arguments (which their method signature prohibits) and not (claim to) return null when that only happened in the case of an invalid null argument. I also cleaned up several unsafe casts and non-null assertions and corrected some other typings.

### Breaking Changes
* The type signature of `ASTNode.getWsCoordinate()` now allows for null return types, which were always possible but not part of the return type before
* `createWorkspaceNode()` no longer allows null workspace or coordinate arguments; it would have just returned null had either of those been provided, so callers check for that and do whatever they would have done had the return value been null.

### DIFF
--- a/core/keyboard_nav/ast_node.ts
+++ b/core/keyboard_nav/ast_node.ts
@@ -724,7 +724,7 @@ export class ASTNode {
     const type = connection.type;
     if (
       type === ConnectionType.INPUT_VALUE ||
-      type === ConnectionType.NEXT_STATEMENT
+      (type === ConnectionType.NEXT_STATEMENT && connection.getParentInput())
     ) {
       const parentInput = connection.getParentInput();
       if (!parentInput) return null;
--- a/core/keyboard_nav/ast_node.ts
+++ b/core/keyboard_nav/ast_node.ts
@@ -708,10 +708,7 @@ export class ASTNode {
    * @param field The location of the AST node.
    * @returns An AST node pointing to a field.
    */
-  static createFieldNode(field: Field): ASTNode | null {
-    if (!field) {
-      return null;
-    }
+  static createFieldNode(field: Field): ASTNode {
     return new ASTNode(ASTNode.types.FIELD, field);
   }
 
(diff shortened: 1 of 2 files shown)


core/keyboard_nav/ast_node.ts
```diff
@@ -724,25 +716,14 @@ export class ASTNode {
...
+      type === ConnectionType.INPUT_VALUE ||
+      (type === ConnectionType.NEXT_STATEMENT && connection.getParentInput())
     ) {
-      // AnyDuringMigration because:  Argument of type 'Input | null' is not
-      // assignable to parameter of type 'Input'.
-      return ASTNode.createInputNode(
-        connection.getParentInput() as AnyDuringMigration,
-      );
+      const parentInput = connection.getParentInput();
+      if (!parentInput) return null;
```
rachel-fenichel, line 725: You shouldn't need to check if the parent input is null here, since you checked it in the condition.

rachel-fenichel (commented):
One more piece of cleanup, plus whatever is going on with the conventional commit labeler, and then good to go.

core/keyboard_nav/ast_node.ts
gonfunko: Done.

gonfunko:
The commit labeler is just busted here because this is targeting v12 and the action fix isn't there yet. I manually assigned the appropriate labels.

rachel-fenichel (approved):

PR #8854, opened 2025-04-02
Fix: Remove the collapsed block warning when expanding a block

### LINKED ISSUE #6020: Warnings for collapsed blocks does not update
**Describe the bug**
When you collapse a block and a child block has a warning we display a warning message on the block. This message isn't getting updated when the block is expanded.

**To Reproduce**

1. Go to the playground

2. Load the following xml
```
  <block type="controls_repeat_ext" id="?[R7O,[bcA,gyUuru9ji" x="62" y="63">
    <value name="TIMES">
      <shadow type="math_number" id="]sg4g$S5/$djRL^,HUBv">
        <field name="NUM">10</field>
      </shadow>
    </value>
    <statement name="DO">
      <block type="procedures_ifreturn" id="-*QyH~ToOCmVL#0E1ew@" disabled="true">
        <mutation value="1"></mutation>
      </block>
    </statement>
  </block>
</xml>
```
4. Collapse the parent block
5. Expand the parent block
6. See how the error message still exists even when the block is expanded

**Expected behavior**

**Screenshots**
Screen Shot 2022-03-22 at 2 07 45 PM

**Additional context**

RoboErikG:
## The basics

### Resolves

Fixes #6020 

### Proposed Changes

- Clears the collapsed block warning message when a block is expanded.
- When collapsing, checks if any children have a warning and if so adds the collapsed warnings warning.

### Reason for Changes

The collapsed block warning message was set if any block in a collapsed stack added a warning, but wasn't added if there already was a warning and was never removed making it sticky on the parent even after expanding and deleting the block with the warning.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -508,6 +508,23 @@ export class BlockSvg
   }
 
   /**
+   * Traverses child blocks to see if any of them have a warning.
+   *
+   * @returns true if any child has a warning, false otherwise.
+   */
+  private childHasWarning(): boolean {
+    const children = this.getChildren(false);
+    for (const child of children) {
+      if (child.getIcon(WarningIcon.TYPE)) {
+        return true;
+      } else if (child.childHasWarning()) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
    * Makes sure that when the block is collapsed, it is rendered correctly
    * for that state.
    */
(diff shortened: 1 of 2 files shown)


rachel-fenichel (commented):
Code looks fine, but please check if we have any related tests and add one for this case if possible.

RoboErikG:
Added tests, which required fixing the other bug where warning icons didn't show up right away when a block is collapsed.

core/block_svg.ts
```diff
@@ -507,6 +507,23 @@ export class BlockSvg
...
 
+  /**
+   * Traverses child blocks to see if any of them have a warning.
+   *
+   * @returns true if any child has a warning, false otherwise.
+   */
+  private childHasWarning(): boolean {
+    const children = this.getChildren(false);
+    for (const child of children) {
+      if (child.getIcon(WarningIcon.TYPE)) {
```
rachel-fenichel, line 518: Can be combined into `if (child.getIcon(WarningIcon.TYPE) || child.childHasWarning()) return true;`

rachel-fenichel (commented):
One quick change then lgtm. I'll manually add the pr: fix label in a sec.

rachel-fenichel (approved):

PR #8860, opened 2025-04-03
fix: Fixed the advanced toolbox, so it does not zoom with the workspace.

MaheshDhingra:
## The basics

### Resolves

Fixes the toolbox in advanced page, which was able to zoom while zooming into the workspace. A small contribution to start with, will continue to look for bugs and fixing them.

### Proposed Changes

It bumps the z-index correctly.

### Reason for Changes

I was able to zoom in the toolbox while zooming the taskbar, if i have clicked on the right black box/bar, and hover on the workspace and then zoom it. So, easy fix was to bump the z-index of it.

### Test Coverage

I have tested it in total of 7 times, as that's my lucky charm, even restarted the server twice to make sure things were okay!

### Documentation

No documentation needs to be updated.

### DIFF
--- a/core/toolbox/toolbox.ts
+++ b/core/toolbox/toolbox.ts
@@ -1093,7 +1093,7 @@ Css.register(`
   overflow-y: auto;
   padding: 4px 0 4px 0;
   position: absolute;
-  z-index: 70;  /* so blocks go under toolbox when dragging */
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Hi @MaheshDhingra, and sorry for the delay. The toolbox is supposed to zoom with the workspace so this is working as intended.

PR #8881, opened 2025-04-11
fix: Fix keyboard-driven menu scrolling.

### LINKED ISSUE #8880: Context menu items don't scroll into view when using keyboard
If you have a context menu with a max height or a short window, the context menu becomes scrollable (see screenshot). Context menu items selected with the keyboard are not scrolled into view.

Image

gonfunko:
## The basics

### Resolves

Fixes #8880

### Proposed Changes
This PR fixes scrolling of contextual menus using the arrow keys. It scrolls the menu item relative to the menu, whereas previously it was scrolling the menu item relative to the menu's parent element. This worked for dropdowns (which have an intermediate div between the menu and the container), but not for contextual menus, where the menu element lived directly in the WidgetDiv. This change broke keyboard scrolling of dropdown menus, but I adjusted their styling to work similarly to context menus. This in turn allowed for the removal of a manual scroll to move the selected dropdown menu item into view, which is now handled automatically by the menu.

### DIFF
--- a/core/css.ts
+++ b/core/css.ts
@@ -124,9 +124,6 @@ let content = `
 
 .blocklyDropDownContent {
   max-height: 300px;  /* @todo: spec for maximum height. */
-  overflow: auto;
-  overflow-x: hidden;
-  position: relative;
 }
 
 .blocklyDropDownArrow {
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):

microbit-robert:
@riknoll It's worth highlighting that this fix impacts MakeCode where some of these changes will need to be copied across or negated. We've quickly [patched this](https://github.com/microbit-matt-hillsdon/pxt/commit/19aac5b90cad750064fb246a7b189e7550358704), but I'm sure there are better approaches. Changes will be required when upgrading from v12.0.0-beta.3 to v12.0.0-beta4.

PR #8889, opened 2025-04-14
feat: Allow for HTML elements in dropdown field menus.

gonfunko:
## The basics

### Resolves

Fixes part of https://github.com/google/blockly-samples/issues/2482

### Proposed Changes
This PR adds support for arbitrary HTML elements in the menu of a `FieldDropdown`. This is mostly to allow them in `FieldGrid`, which is a `FieldDropdown` subclass, but also gives developers greater flexibility. This PR does not attempt to ensure reasonable bounds or interaction behavior for HTML elements in menus; provided elements should not be unreasonably large or expect to handle mouse/keyboard interaction, but this is the responsibility of developers to ensure.

When an HTML-element-based menu item is selected, a text representation of the element will be displayed in the field. In order of preference, the text is derived from (1) the `title` attribute of the element, (2) the `aria-label` attribute of the element, and (3) the `innerText` of the element.

This PR also updates several tests that omitted various fields from the `ImageProperties` interface; the interface and [documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown#image_dropdowns) state that all 4 properties are required, but the implementation was a bit loose on that.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -336,9 +336,9 @@ export class FieldDropdown extends Field<string> {
       const content = (() => {
         if (isImageProperties(label)) {
           // Convert ImageProperties to an HTMLImageElement.
-          const image = new Image(label['width'], label['height']);
-          image.src = label['src'];
-          image.alt = label['alt'] || '';
+          const image = new Image(label.width, label.height);
+          image.src = label.src;
+          image.alt = label.alt;
           return image;
         }
         return label;
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -334,7 +334,7 @@ export class FieldDropdown extends Field<string> {
 
       const [label, value] = option;
       const content = (() => {
-        if (typeof label === 'object') {
+        if (isImageProperties(label)) {
           // Convert ImageProperties to an HTMLImageElement.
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
@maribethb can you take a look at this one for me?

core/field_dropdown.ts
```diff
@@ -334,7 +334,7 @@ export class FieldDropdown extends Field<string> {
 
       const [label, value] = option;
       const content = (() => {
-        if (typeof label === 'object') {
+        if (isImageProperties(label)) {
           // Convert ImageProperties to an HTMLImageElement.
           const image = new Image(label['width'], label['height']);
```
maribethb, line 339: do you want to update these to use `.` access while you're here and updating other uses?

maribethb (approved):

core/field_dropdown.ts
gonfunko: Good call, done!

rpbourret:
This applies to issue 2482 in `blockly-samples` ([Generalise what can appear as field-grid-dropdown items](https://github.com/google/blockly-samples/issues/2482#top)), not 2482 in `blockly`, as is stated.

PR #8900, opened 2025-04-17
fix: Fix bug that caused flyout items under the mouse to be selected without movement.

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/216

### Proposed Changes
This PR updates the flyouts to select items on `pointermove` rather than `pointerenter`. This resolves an issue where, when using keyboard navigation, flyout items that happened to appear under the mouse when the flyout was opened via the keyboard would appear selected. This resulted in a confusing behavior where there could be two selections: the cursor, and the block incidentally under the mouse. Now, blocks will not be selected/highlighted unless the mouse is actually moved.

### DIFF
--- a/core/block_flyout_inflater.ts
+++ b/core/block_flyout_inflater.ts
@@ -222,7 +222,7 @@ export class BlockFlyoutInflater implements IFlyoutInflater {
     );
 
     blockListeners.push(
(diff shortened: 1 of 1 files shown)


BenHenning (approved):
Thanks @gonfunko! No concerns here.

PR #8875, opened 2025-04-09
[Do not merge] Test prototype focus manager incorporation

BenHenning:
Note that this includes changes from #8909, #8916, and #8920.

## The basics

- [ ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes asdf

### Proposed Changes

This is a prototype change. The details are explained in https://github.com/google/blockly-keyboard-experimentation/pull/415.

Issue #8771, opened 2025-02-14
Make key components focusable


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Many different Blockly components need to be updated in order to implementations of either ``IFocusableTree``, ``IFocusableNode``, or both, in order to properly integrate with the broader focus management system (see #8770) and the integration of this system.

### Request

The following Blockly components, at minimum, are expected to require updating:
- [ ] Toolbox categories and subcategories.
- [ ] Workspace.
- [ ] Comments.
- [ ] The flyout (such as for buttons and labels), and arbitrary items that can be added as of v12 (see #8687 for some context on how this affected the old keyboard navigation plugin).
- [ ] Blocks.
- [ ] Connections.
- [ ] Fields (may require further break-down).
- [ ] Icons.
- [ ] Mutators.
- [ ] Bubbles.
- [ ] Flyout buttons and labels.
- [ ] Zoom controls.
- [ ] The trashcan control and the trashcan flyout.
- [ ] Any custom flyout items with a visual representation.
- [ ] Other positionables, such as [search widgets](https://google.github.io/blockly-samples/plugins/workspace-search/test/index.html) if we're putting that in the top right corner
- [ ] Arbitrary modals (such as the confirmation dialog when deleting a variable that is used in multiple places).

### Alternatives considered

N/A -- see the technical design linked below for context on broad alternatives considered as part of #8770.

### Additional context

This relates to #8770 and the technical design laid out in https://github.com/google/blockly-keyboard-experimentation/issues/142#issuecomment-2657938253.

BenHenning:
If positionables are fully covered by implementations of `IPositionable` then making them focusable will definitely require introducing an API breakage. It has a UI element implied but not required by the interface.

Looking more broadly at `IComponent` we also have `IDragTarget` which looks like something that _could_ benefit from keyboard navigation interaction (maybe as a generalization of the current dragging support in the keyboard navigation plugin).

Separately there's `IAutoHideable` as the only other extension of `IComponent` in core Blocky. This one seems less important to make focusable since any important auto-hideables will already be focusable (like the flyout).

I think there are large open questions about making both `IPositionable` and `IDragTarget` focusable generally (or even `IComponent`) that could require additional design work.

BenHenning:
[...]

PR #8909, opened 2025-04-21
feat!: Introduce new focus tree/node functions.

### LINKED ISSUE #8770: Add support for centrally managing and changing DOM focus
### Check for duplicates

### Problem

Core Blockly doesn't currently have a good way to handle the following user flows:
- Ensuring element selection/navigation via the keyboard navigation plugin ties back to DOM focus (which is necessary screenreader support).
- Ensuring that currently focused elements have clear visual indicators (this sort of works today for elements that are selectable, but otherwise we rely on browser focus visual indicators which are very subtle or outright may not work in certain SVG contexts).
- Ensuring that distinct focus 'domains' (e.g. the workspace and toolbox) can represent 'passive' focus (that is, the last element within that logical context that held DOM focus but doesn't currently). This is especially useful for communicating which visual element will receive focus when returning to that context (such as navigating from the toolbox back to the workspace).
- Ensuring cases when focus context needs to be taken temporarily (such as for drop-downs) can be done in a clean way.

### Request

https://github.com/google/blockly-keyboard-experimentation/issues/142#issuecomment-2657938253 describes the high-level solution for this, but specifically this issue is meant to track adding (see also the sub-issues):

### Alternatives considered

The technical design above has a lot of context on alternative considerations.

### Additional context

The technical design above should have all important additional context. #8771 tracks more specific components that need to be updated in order for the system introduced with this issue to be useful. https://github.com/google/blockly-keyboard-experimentation/issues/142 tracks the work to integrate the new focus system within the keyboard navigation plugin.

BenHenning:
## The basics

### Resolves

Fixes #8770
Fixes part of #8771

### Proposed Changes

This introduces new callback methods for IFocusableTree and IFocusableNode for providing a basis of synchronizing domain state with focus changes. It also introduces support for implementations of IFocusableTree to better manage initial state cases, especially when a tree is focused using tab navigation.

FocusManager has also been updated to ensure functional parity between tab-navigating to a tree and using focusTree() on that tree. This means that tab navigating to a tree will actually restore focus back to that tree's previous focused node rather than the root (unless the root is navigated to from within the tree itself). This is meant to provide better consistency between tab and non-tab keyboard navigation.

### Reason for Changes

#8875 demonstrates future changes that will depend on the changes introduced in this PR. Much of this functionality is needed to ensure an easier time updating Workspace, Toolbox, and Flyout to be focusable trees along with their constituent nodes.

### Test Coverage

TODO

### Documentation

No documentation changes should be needed here beyond the ones already introduced with JS docs.

### Additional Information

Note that these changes originally came from #8875 and are required for later PRs that will introduce IFocusableNode and IFocusableTree implementations.

### DIFF
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -60,6 +60,7 @@ export class FocusManager {
   registeredTrees: Array<IFocusableTree> = [];
 
   private currentlyHoldsEphemeralFocus: boolean = false;
+  private lockFocusStateChanges: boolean = false;
 
   constructor(
     addGlobalEventListener: (type: string, listener: EventListener) => void,
@@ -89,7 +90,16 @@ export class FocusManager {
       }
 
       if (newNode) {
-        this.focusNode(newNode);
+        const newTree = newNode.getFocusableTree();
+        const oldTree = this.focusedNode?.getFocusableTree();
+        if (newNode === newTree.getRootFocusableNode() && newTree !== oldTree) {
+          // If the root of the tree is the one taking focus (such as due to
+          // being tabbed), try to focus the whole tree explicitly to ensure the
+          // correct node re-receives focus.
+          this.focusTree(newTree);
+        } else {
+          this.focusNode(newNode);
+        }
       } else {
         this.defocusCurrentFocusedNode();
       }
@@ -108,6 +118,7 @@ export class FocusManager {
    * certain whether the tree has been registered.
    */
   registerTree(tree: IFocusableTree): void {
+    this.ensureManagerIsUnlocked();
     if (this.isRegistered(tree)) {
       throw Error(`Attempted to re-register already registered tree: ${tree}.`);
     }
@@ -133,6 +144,7 @@ export class FocusManager {
    * this manager.
    */
   unregisterTree(tree: IFocusableTree): void {
+    this.ensureManagerIsUnlocked();
     if (!this.isRegistered(tree)) {
       throw Error(`Attempted to unregister not registered tree: ${tree}.`);
     }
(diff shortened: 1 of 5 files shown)


core/focus_manager.ts
```diff
@@ -89,7 +90,16 @@ export class FocusManager {
       }
 
       if (newNode) {
-        this.focusNode(newNode);
+        const newTree = newNode.getFocusableTree();
+        const oldTree = this.focusedNode?.getFocusableTree();
+        if (newNode === newTree.getRootFocusableNode() && newTree !== oldTree) {
```
BenHenning, line 95: Note to reviewer: this is a functionality change from before. These changes mean that tabbing to the root of a tree will always behave the same as focusTree(), that is, it will restore focus active focus back to the previous node if there was any. This introduces consistency with explicit focusTree() calls, but may not always be expected by the user as it will shift focus to something new. Note that the extra tree check is to ensure users can tab out of a tree (in which case they WILL go to the root and set that as the last known focus state of the tree which should generally stabilize the user experience since they explicitly navigated away).

BenHenning (commented):
Self-reviewed changes and made a few corrections.

BenHenning:
PTAL @rachel-fenichel.

BenHenning (commented):
Reviewed method re-introduction.

Apologies, I thought that `isFocusableTree` and `isFocusableNode` weren't needed.

BenHenning (commented):
Reviewed a fix for an existing FocusManager issue that I discovered downstream + a new test for it.

rachel-fenichel (approved):
LGTM after adding comments on some private functions.

core/focus_manager.ts
BenHenning: Assuming approval means this behavior change is okay. :)

rachel-fenichel: Yes. The behaviour change makes sense and is well explained.

BenHenning:
Going ahead and merging since I suspect the private method doc changes will be uncontroversial.

Issue #8915, opened 2025-04-21
Add tests for accessibility use cases


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

There aren't yet end-to-end tests for verifying the new keyboard navigation and accessibility use cases in Core Blockly (though they have started to be implemented for navigation in the keyboard navigation plugin repository: https://github.com/google/blockly-keyboard-experimentation/tree/main/test).

### Request

WebdriverIO tests should be added that verify both keyboard navigation & readout state as more of the keyboard navigation plugin and navigation functionality is moved into Core Blockly.

BenHenning:
Reminder to consider shadow block cases specifically as they relate to selection & focus per https://github.com/google/blockly/pull/8916#discussion_r2056747205.

BenHenning:
From #8923: make sure to add tests for drop-down & widget divs in how they manage their ephemeral focus.

BenHenning:
https://github.com/google/blockly/issues/8826#issuecomment-2860837080 another reference point.

BenHenning:
Note that I plan to address this issue in two pats:
1. Identifying and shoring up existing (regression) tests for cases that were found while working on v12-related features (especially around focus).
2. Identifying top-level classes of user flows (likely more from an end-to-end perspective) to ensure thorough accessibility testing. The latter should complement existing plans and tests within the keyboard navigation plugin repo, plus consider screen reader label outputs (though some of both of these can also be tested more thoroughly with unit tests--it's likely there will be two categories of new accessibility-specific tests to be added here).

rachel-fenichel:
A list of test cases for copy/paste behaviour: https://github.com/google/blockly-keyboard-experimentation/issues/103#issuecomment-2881792242

gonfunko:
Test for committing moves on keyboard shortcuts: https://github.com/google/blockly-keyboard-experimentation/pull/564#pullrequestreview-2879029449

gonfunko:
Closing now that v13 is out and we have decent test coverage for that; the one remaining open subissue can track that specific piece of work.

PR #8916, opened 2025-04-21
feat: Make WorkspaceSvg and BlockSvg focusable

### LINKED ISSUE #8913: Workspace.

### LINKED ISSUE #8914: Blocks.

BenHenning:
## The basics

### Resolves

Fixes #8913
Fixes #8914
Fixes part of #8771

### Proposed Changes

This updates `WorkspaceSvg` and `BlockSvg` to be focusable, that is, it makes the workspace a `IFocusableTree` and blocks `IFocusableNode`s.

Some important details:
- While this introduces focusable tree support for `Workspace` it doesn't include two other components that are obviously needed by the keyboard navigation plugin's playground: fields and connections. These will be introduced in subsequent PRs.
- Blocks are set up to automatically synchronize their selection state with their focus state. This will eventually help to replace `LineCursor`'s responsibility for managing selection state itself.
- The tabindex property for the workspace and its ARIA label have been moved down to the `.blocklyWorkspace` element itself rather than its wrapper. This helps address some tab stop issues that are already addressed in the plugin (via monkey patches), but also to ensure that the workspace's main SVG group interacts correctly with `FocusManager`.
- `WorkspaceSvg` is being initially set up to default to its first top block when being focused for the first time. This is to match parity with the keyboard navigation plugin, however the latter also has functionality for defaulting to a position when no blocks are present. It's not clear how to actually support this under the new focus-based system (without adding an ephemeral element on which to focus), or if it's even necessary (since the workspace root can hold focus).

### Reason for Changes

### Test Coverage

### Documentation

### Additional Information

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1836,9 +1836,7 @@ export class BlockSvg
 
   /** See IFocusableNode.onNodeFocus. */
   onNodeFocus(): void {
-    if (!this.isShadow()) {
-      common.setSelected(this);
-    }
+    common.setSelected(this);
   }
 
   /** See IFocusableNode.onNodeBlur. */
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -44,6 +44,8 @@ import {IContextMenu} from './interfaces/i_contextmenu.js';
 import type {ICopyable} from './interfaces/i_copyable.js';
 import {IDeletable} from './interfaces/i_deletable.js';
 import type {IDragStrategy, IDraggable} from './interfaces/i_draggable.js';
+import type {IFocusableNode} from './interfaces/i_focusable_node.js';
+import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {IIcon} from './interfaces/i_icon.js';
 import * as internalConstants from './internal_constants.js';
 import {MarkerManager} from './marker_manager.js';
@@ -76,7 +78,8 @@ export class BlockSvg
     IContextMenu,
     ICopyable<BlockCopyData>,
     IDraggable,
-    IDeletable
+    IDeletable,
+    IFocusableNode
 {
   /**
    * Constant for identifying rows that are to be rendered inline.
@@ -210,6 +213,7 @@ export class BlockSvg
 
     // Expose this block's ID on its top-level SVG group.
     this.svgGroup.setAttribute('data-id', this.id);
+    svgPath.id = this.id;
 
     this.doInit_();
   }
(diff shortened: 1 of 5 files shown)


BenHenning (commented):
Self-reviewed changes that aren't part of #8909.

BenHenning:
PTAL @gonfunko. FYI this includes changes from #8909 which can be ignored (there are no additional changes in those files in this PR).

gonfunko (approved):

core/block_svg.ts
```diff
@@ -1819,4 +1823,28 @@ export class BlockSvg
...
+  }
+
+  /** See IFocusableNode.getFocusableTree. */
+  getFocusableTree(): IFocusableTree {
+    return this.workspace;
+  }
+
+  /** See IFocusableNode.onNodeFocus. */
+  onNodeFocus(): void {
+    if (!this.isShadow()) {
```
rachel-fenichel, line 1839: What happens if it is a shadow block?

core/workspace_svg.ts
```diff
@@ -764,7 +769,19 @@ export class WorkspaceSvg
      *   <g class="blocklyBubbleCanvas"></g>
      * </g>
      */
-    this.svgGroup_ = dom.createSvgElement(Svg.G, {'class': 'blocklyWorkspace'});
+    this.svgGroup_ = dom.createSvgElement(Svg.G, {
+      'class': 'blocklyWorkspace',
+      // Only the main workspace should be tabbable.
```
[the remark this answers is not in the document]
BenHenning: That's correct--fixed!

BenHenning (commented):
Self-reviewed latest changes & followed up to review comments.

Note that #8909 has been merged so this PR is only the changes related to workspace & blocks. Subsequently, I realized I forgot to include the tab index change to make blocks focusable, so that's been fixed in the latest changes.

BenHenning:
PTAL @gonfunko if you would like to double check the latest changes.

PTAL @rachel-fenichel per your comments.

core/block_svg.ts
rachel-fenichel: I'm also not sure. I think call `setSelected` directly here and do some manual verification, then file a ticket to follow up with comprehensive tests for that case.

BenHenning: So far I'm not seeing any issues, so I don't think there's a particular concern to address.

I've made a note in #8915 regarding this specific case.

BenHenning (commented):
Self-reviewed latest brances change.

rachel-fenichel (approved):

BenHenning:
Thanks all for the reviews. :) Going ahead and merging (with squash this time...).

PR #8920, opened 2025-04-22
feat: Make toolbox and flyout focusable

### LINKED ISSUE #8918: Toolbox categories and subcategories.

### LINKED ISSUE #8919: The flyout.

BenHenning:
## The basics

### Resolves

Fixes #8918
Fixes #8919
Fixes part of #8771

### Proposed Changes

This updates several classes in order to make toolboxes and flyouts focusable:
- `IFlyout` is now an `IFocusableTree` with corresponding implementations in `FlyoutBase`.
- `IToolbox` is now an `IFocusableTree` with corresponding implementations in `Toolbox`.
- `IToolboxItem` is now an `IFocusableNode` with corresponding implementations in `ToolboxItem`.
- As the primary toolbox items, `ToolboxCategory` and `ToolboxSeparator` were updated to have -1 tab indexes and defined IDs to help `ToolboxItem` fulfill its contracted for `IFocusableNode.getFocusableElement`.

Each of these two new focusable trees have specific noteworthy behaviors behaviors:
- `Toolbox` will automatically indicate that its first item should be focused (if one is present), even overriding the ability to focus the toolbox's root (however there are some cases where that can still happen).
- `Toolbox` will automatically synchronize its selection state with its item nodes being focused.
- `FlyoutBase`, now being a focusable tree, has had a tab index of 0 added. Normally a tab index of -1 is all that's needed, but the keyboard navigation plugin specifically uses 0 for flyout so that the flyout is tabbable. This is a **new** tab stop being introduced.
- `FlyoutBase` holds a workspace (for rendering blocks) and, since `WorkspaceSvg` is already set up to be a focusable tree, it's represented as a subtree to `FlyoutBase`. This does introduce some wonky behaviors: the flyout's root will have passive focus while its contents have active focus. This could be manually disabled with some CSS if it ends up being a confusing user experience.
- Both `FlyoutBase` and `WorkspaceSvg` have built-in behaviors for detecting when a user tries navigating away from an open flyout to ensure that the flyout is closed when it's supposed to be. That is, the flyout is auto-hideable and a non-flyout, non-toolbox node has then been focused. This matches parity with the `T`/`Esc` flows supported in the keyboard navigation plugin playground.

One other thing to note: `Toolbox` had a few tests to update that were trying to reinit a toolbox without first disposing of it (which was caught by one of `FocusManager`'s state guardrails).

### Reason for Changes

### Test Coverage

### Documentation

### Additional Information


BenHenning (commented):
Self-reviewed changes not part of #8916 (though note that this includes `WorkspaceSvg` changes on top of that PR's changes).

BenHenning:
PTAL @maribethb. Note that files changed in #8916 and #8909 can be ignored except for `WorkspaceSvg` (since both this and #8916 make changes to that file).;

rachel-fenichel:
@BenHenning would you like me to take over reviewing this one since Maribeth is out of office?

BenHenning:
Yes please @rachel-fenichel but I will bring this up-to-date momentarily after #8916 has had a couple of updates, and will assign it over to you.

BenHenning:
PTAL @rachel-fenichel.

core/toolbox/toolbox.ts
```diff
@@ -1077,7 +1086,68 @@ export class Toolbox
...
+
+  /** See IFocusableTree.getRestoredFocusableNode. */
+  getRestoredFocusableNode(
+    previousNode: IFocusableNode | null,
+  ): IFocusableNode | null {
+    // Always try to select the first selectable toolbox item rather than the
+    // root of the toolbox.
+    if (!previousNode || previousNode === this) {
+      return this.getToolboxItems().find((item) => item.isSelectable()) ?? null;
+    } else return null;
```
rachel-fenichel, line 1123: No need for the `else`, just do a newline and `return null`

core/workspace_svg.ts
```diff
@@ -2618,6 +2642,67 @@ export class WorkspaceSvg
...
+    // If the flyout loses focus, make sure to close it.
+    if (this.isFlyout && this.targetWorkspace) {
+      // Only hide the flyout if the flyout's workspace is losing focus and that
+      // focus isn't returning to the flyout itself or the toolbox.
+      const flyout = this.targetWorkspace.getFlyout();
+      const toolbox = this.targetWorkspace.getToolbox();
+      if (flyout && nextTree === flyout) return;
+      if (toolbox && nextTree === toolbox) return;
+      if (toolbox) toolbox.clearSelection();
+      if (flyout && flyout instanceof Flyout) flyout.autoHide(false);
```
rachel-fenichel, line 2703: What else would `flyout` be besides an `instanceof Flyout`?

core/toolbox/toolbox.ts
BenHenning: Done.

core/workspace_svg.ts
BenHenning: It's `IFlyout` in this context per: https://github.com/google/blockly/blob/ac7fea11cffc4c445cedae77675290861ceec2f6/core/workspace_svg.ts#L1014

I think logically it can only ever be `Flyout` but TypeScript doesn't know that.

core/toolbox/toolbox.ts
```diff
@@ -1077,7 +1086,69 @@ export class Toolbox
...
+  onTreeFocus(
+    node: IFocusableNode,
+    _previousTree: IFocusableTree | null,
+  ): void {
+    if (node !== this) {
+      // Only select the item if it isn't already selected so as to not toggle.
+      if (this.getSelectedItem() !== node) {
+        this.setSelectedItem(node as IToolboxItem);
+      }
+    } else this.clearSelection();
```
rachel-fenichel, line 1147: Nit: use braces around the else

core/workspace_svg.ts
rachel-fenichel: I see. Makes sense.

rachel-fenichel (approved):
One nit then LGTM!

core/toolbox/toolbox.ts
BenHenning: Fixed!

BenHenning (commented):
Self-reviewed changes after merging #8916. I don't think another review round is necessary as the changes look clean.

BenHenning (commented):
Empty commit review.

BenHenning:
Thanks for the reviews! Going ahead and merging this to keep the downstream PRs simpler.

core/workspace_svg.ts
maribethb: > It's `IFlyout` in this context per:
> 
> https://github.com/google/blockly/blob/ac7fea11cffc4c445cedae77675290861ceec2f6/core/workspace_svg.ts#L1014
> 
> I think logically it can only ever be `Flyout` but TypeScript doesn't know that.

Using our base Flyout class is technically not required, only using the IFlyout interface, and I’m pretty sure one of the partners doesn’t use our base class. Instead you should use the IAutoHideable interface we have. Which our base class implements but isn’t required

PR #8923, opened 2025-04-22
feat: Make fields focusable

### LINKED ISSUE #8922: Fields (may require further break-down).

### LINKED ISSUE #8929: Arbitrary modals (such as the confirmation dialog when deleting a variable that is used in multiple places).

BenHenning:
## The basics

### Resolves

Fixes #8922
Fixes #8929
Fixes part of #8771

### Proposed Changes

This PR introduces support for fields to be focusable (and thus navigable with keyboard navigation when paired with downstream changes to `LineCursor` and the keyboard navigation plugin). This is a largely isolated change in how it fundamentally works:
- `Field` was updated to become an `IFocusableNode`. Note that it uses a specific string-based ID schema in order to ensure that it can be properly linked back to its unique block (which helps make the search for the field in `WorkspaceSvg` a bit more efficient). This could be done with a globally unique ID, instead, but all fields would need to be searched vs. just those for the field's parent block.
- The drop-down and widget divs have been updated to manage ephemeral focus with `FocusManager` when they're open for non-system dialogs (ephemeral focus isn't needed for system dialogs/prompts since those already take/restore focus in a native way that `FocusManager` will respond to--this may require future work, however, if the restoration causes unexpected behavior for users). This approach was done due to a suggestion from @maribethb as the alternative would be a more complicated breaking change (forcing `Field` subclasses to properly manage ephemeral focus). It may still be the case that certain cases will need to do so, but widget and drop-down divs seem to address the majority of possibilities.

**Important**: `Input`s are not explicitly being supported here. As far as I can tell, we can't run into a case where `LineCursor` tries to set an input node, though perhaps I simply haven't come across this case. Supporting `Fields` and `Connections` (per #8928) seems to cover the main needed cases, though making `Input`s focusable may be a future requirement.

### Reason for Changes

Note that #8929 isn't broadly addressed since making widget & drop down divs manage ephemeral focus directly addresses a large class of cases. Additional cases may arise where a plugin or Blockly integration may require additional effort to make keyboard navigation work for their field--this may be best addressed with documentation and guidance.

### Test Coverage

### Documentation

No new documentation is planned, however it may be prudent to update the field documentation in the future to explain how to utilize ephemeral focus when specifically building compatibility for keyboard navigation.

### Additional Information


core/field.ts
```diff
@@ -1401,6 +1435,30 @@ export abstract class Field<T = any>
...
+    return this.fieldGroup_;
+  }
+
+  /** See IFocusableNode.getFocusableTree. */
+  getFocusableTree(): IFocusableTree {
+    const block = this.getSourceBlock();
+    if (!block) {
+      throw new UnattachedFieldError();
+    }
+    // TODO: Remove the latter two casts once WorkspaceSvg is a focusable tree.
```
BenHenning, line 1452: Reminder to fix this & update workspace to integrate with fields correctly once #8916 is merged.

BenHenning: Since this PR is now chained off of the other branches, this has been fixed.

core/field.ts
```diff
@@ -191,6 +195,8 @@ export abstract class Field<T = any>
    */
   SERIALIZABLE = false;
 
+  private id_: string | null = null;
```
BenHenning, line 198: Add doc comment.

BenHenning (commented):
Self-reviewed changes not included from #8939.

core/field.ts
BenHenning: Addressed in latest changes.

BenHenning:
PTAL @rachel-fenichel. Feel free to use https://github.com/BenHenning/blockly/pull/5 to see the exact diff with #8939 which will hopefully make the review easier.

rachel-fenichel:
LGTM--the exact diff is not that large and matches the designed behaviour.

rachel-fenichel (approved):

BenHenning (commented):
Self-review confirming this has the same changes as https://github.com/BenHenning/blockly/pull/5.

BenHenning:

PR #8925, opened 2025-04-23
Fix: Add padding to the bounds before deciding if it should scroll

RoboErikG:
This makes it easier to scroll something into view with context by including the padding before checking if it should be scrolled into view.

## The basics

### Resolves

Fixes an issue uncovered in https://github.com/google/blockly-keyboard-experimentation/pull/451

### Proposed Changes

Add the padding to the bounds before checking if it's within the viewport.

### Reason for Changes

During a keyboard drag this method may be called more once. When that happens only the first call with default padding is executed.  During a constrained drag we'd like to increase the padding to give more context about the new location, so we should check with the padding to make sure later calls that try to move it further in bounds are respected.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -2581,23 +2581,23 @@ export class WorkspaceSvg
       rawViewport.left + rawViewport.width,
     );
 
-    if (
-      bounds.left >= viewport.left &&
-      bounds.top >= viewport.top &&
-      bounds.right <= viewport.right &&
-      bounds.bottom <= viewport.bottom
-    ) {
-      // Do nothing if the block is fully inside the viewport.
-      return;
-    }
-
-    // Add some padding to the bounds so the element is scrolled comfortably
+    // Add the padding to the bounds so the element is scrolled comfortably
     // into view.
     bounds = bounds.clone();
     bounds.top -= padding;
     bounds.bottom += padding;
     bounds.left -= padding;
     bounds.right += padding;
+
+    if (
+      bounds.left >= viewport.left &&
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
I am not sure this is always desirable. Blocks get scrolled into view by keyboard nav on selection, which can just be defaulted e.g. when you move to the workspace. There are often blocks near the origin. So this can mean the workspace jumps by padding unexpectedly.

Pretty sure it used to be this way and I changed it because it felt glitchy.

Whether you want to include the padding might depend on the action you're performing. In the context of a move it's much more reasonable.

core/workspace_svg.ts
```diff
@@ -2581,24 +2581,24 @@ export class WorkspaceSvg
...
+    bounds.right += padding;
+
     if (
       bounds.left >= viewport.left &&
       bounds.top >= viewport.top &&
       bounds.right <= viewport.right &&
       bounds.bottom <= viewport.bottom
     ) {
-      // Do nothing if the block is fully inside the viewport.
+      // Do nothing if the block with padding is fully inside the viewport.
```
gonfunko, line 2598: Nit while this comment is being edited: should be "bounds", not "block"

gonfunko (approved):

RoboErikG:
Good point @microbit-matt-hillsdon . An alternate solution would be to add it to the bounds before calling this method to force the move and leaving this method as is. I'll play around with it after @microbit-grace gets her change in.

RoboErikG:
Resolved this via https://github.com/google/blockly-keyboard-experimentation/pull/474 instead

Issue #8901, opened 2025-04-15
LineCursor can loop forward, but not back


microbit-robert:
This is in normally navigation mode only. Move mode works as expected.

We've patched this temporarily. Note, this patch also contains changes to remove the stack connection cursor positions. Disregard changes to validLineNode and validInLineNode. [Patch diff](https://github.com/google/blockly-keyboard-experimentation/compare/main...microbit-matt-hillsdon:blockly-keyboard-experimentation:line-cursor-patch).

Summary of relevant changes:
- Change the `prev` and `out` methods to call `getPreviousNode` with the `loop` param set to `true`.
- I also pass the `isValid` function to `getLastNode`, which is passed through `getNextNode` where it's called inside `getLastNode` (required to avoid errors on looping back).

gonfunko:
This has been incidentally fixed.

PR #8926, opened 2025-04-24
fix: LineCursor can loop forward, but not back

microbit-grace:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8901 

### Proposed Changes

- Use `getPreviousNode` in `prev` and `out` methods so that the `loop` parameter can be used and set as `true`.
- Add tests for loopback in `prev` and `out` methods.

### Reason for Changes

To fix https://github.com/google/blockly/issues/8901 so that cursor loops when keyboard navigating blocks backwards (up arrow keys, or left arrow keys) as it can forwards.

### Test Coverage

Added loop tests for `prev` and `out` in `cursor_test.ts`.

### DIFF
--- a/tests/mocha/cursor_test.js
+++ b/tests/mocha/cursor_test.js
@@ -125,6 +125,15 @@ suite('Cursor', function () {
       assert.equal(curNode.getLocation(), this.blocks.A.nextConnection);
     });
 
+    test('Prev - From first connection loop to last next connection', function () {
+      const prevConnection = this.blocks.A.previousConnection;
+      const prevConnectionNode = ASTNode.createConnectionNode(prevConnection);
+      this.cursor.setCurNode(prevConnectionNode);
+      this.cursor.prev();
+      const curNode = this.cursor.getCurNode();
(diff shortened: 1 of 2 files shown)


tests/mocha/cursor_test.js
```diff
@@ -133,6 +142,15 @@ suite('Cursor', function () {
       const curNode = this.cursor.getCurNode();
       assert.equal(curNode.getLocation(), this.blocks.E);
     });
+
+    test('Out - From first connection loop to last next connection', function () {
+      const prevConnection = this.blocks.A.previousConnection;
+      const prevConnectionNode = ASTNode.createConnectionNode(prevConnection);
+      this.cursor.setCurNode(prevConnectionNode);
+      this.cursor.prev();
```
gonfunko, line 150: Should this be `this.cursor.out()`?

gonfunko (approved):

tests/mocha/cursor_test.js
gonfunko: Thanks!

PR #8928, opened 2025-04-24
feat: Make connections focusable

### LINKED ISSUE #8930: Connections.

BenHenning:
## The basics

### Resolves

Fixes #8930
Fixes part of #8771

### Proposed Changes

This PR introduces support for connections to be focusable (and thus navigable with keyboard navigation when paired with downstream changes to `LineCursor` and the keyboard navigation plugin). This is a largely isolated change in how it fundamentally works:
- `RenderedConnection` has been updated to be an `IFocusableNode` using a new unique ID maintained by `Connection` and automatically enabling/disabling the connection highlight based on whether it's focused (per keyboard navigation).
- The way that rendering works here has changed: rather than recreating the connection's highlight SVG each time, it's only created once and updated thereafter to ensure that it correctly fits block resizes or movements. Visibility of the highlight is controlled entirely through display visibility and can now be done synchronously (which was a requirement for focusability as only displayed elements can be focused).
- This employs the same type of ID schema strategy as fields in #8923.

### Reason for Changes

### Test Coverage

### Documentation

### Additional Information


core/connection.ts
```diff
@@ -55,6 +56,8 @@ export class Connection implements IASTNodeLocationWithBlock {
   /** DOM representation of a shadow block, or null if none. */
   private shadowDom: Element | null = null;
 
+  id: string;
```
BenHenning, line 59: Add docs.

core/renderers/common/path_object.ts
```diff
@@ -268,37 +268,33 @@ export class PathObject implements IPathObject {
...
+  ): SVGElement {
+    const transformation =
+      `translate(${offset.x}, ${offset.y})` + (rtl ? ' scale(-1 1)' : '');
+
+    const previousHighlight = this.connectionHighlights.get(connection);
+    if (previousHighlight) {
+      // Since a connection already exists, make sure that its path and
+      // transform are correct.
+      previousHighlight.setAttribute('d', connectionPath);
+      previousHighlight.setAttribute('transform', transformation);
```
BenHenning, line 280: FYI @rachel-fenichel this bit confuses me. The connection seems to still magically update even if these aren't re-set across block movements or changes (despite not recreating the SVG). Either I'm doing something completely wrong here, or completely misunderstanding something because that doesn't sound correct to me.

BenHenning (commented):
Self-reviewed changes not overlapping with #8923.

core/connection.ts
BenHenning: Addressed in the latest changes.

BenHenning:
PTAL @rachel-fenichel. FYI that this does include a few changes over #8920 in order to make `FlyoutButton` focusable. Feel free to use https://github.com/BenHenning/blockly/pull/3 to see the exact diff with #8938 which will hopefully make the review easier.

@RoboErikG feel free to take a pass, but I believe Rachel plans to review the entire chain of PRs up to #8941.

core/renderers/common/path_object.ts
BenHenning: To be extra clear (since I realized now upon re-reading that this could've been worded better): if I remove these lines, the highlighting still seems to work and I'm having difficulty reasoning through the dataflow sufficiently well as to understand why that makes sense. I understand that the block paths basically get recreated upon movements and such, and that in turn means that the connection highlights are recomputed, but my other changes in this PR should prevent this block from being called if the path already exists. Unless the `Connection` objects themselves are being recreated (which I highly doubt since it would break focus), the original highlight transform shouldn't be updating. :)

rachel-fenichel: If you remove which lines?

BenHenning, line 280: The ones highlighted in this comment:

https://github.com/google/blockly/blob/896dbd7c11e532661fc6ab75ad759b604bbf38fb/core/renderers/common/path_object.ts#L279-L280

I.e. if we don't recompute the transform for an existing connection highlight, they still seem to magically go to the right place even if the block moves.

rachel-fenichel: Is it across block movements or blocks changing shape?
If it's about block movements, that's not surprising, because the transform is relative to the origin of the block's SVG and not to the rest of the page. Translating the block therefore also translates the connection highlights and all of the attached blocks.

BenHenning: Sorry--not at all sure why the comment started a new thread.

I think it might be the relative behavior, then. I tried a few box shape changing things and it didn't make a difference, but I probably didn't get the block to resize in the correct way to trigger a failure (since the connection highlight doesn't span the whole blog length like the marker does, so simply changing the width isn't quite sufficient to produce the case that would break the connection highlight w/o this update logic).

BenHenning: Continued in https://github.com/google/blockly/pull/8928#discussion_r2069354218. No clue why a new thread started.

BenHenning (commented):
Self-review to confirm that this seems to contain the same changes as https://github.com/BenHenning/blockly/pull/3.

rachel-fenichel (approved):

core/renderers/common/path_object.ts
BenHenning: Resolving since this is continued in the other thread.

BenHenning: Discussed over chat, but it largely seems that I may have missed cases where connection actually changes size (and am, admittedly, still slightly confused on why those cases _seem_ to work without this code in place). Nevertheless, it's definitely more correct to include this so continuing with that plan.

BenHenning:

PR #8933, opened 2025-04-25
Fix: Revert focus prs

RoboErikG:
## The basics

### Resolves

Fixes problems with focus and key handling introduced by #8916

### Proposed Changes

Revert two PRs that merged a portion of the focus behavior into Core.

### Reason for Changes

The changes broke some aspects of keyboard navigation being worked on in keyboard experimental

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -44,8 +44,6 @@ import {IContextMenu} from './interfaces/i_contextmenu.js';
 import type {ICopyable} from './interfaces/i_copyable.js';
 import {IDeletable} from './interfaces/i_deletable.js';
 import type {IDragStrategy, IDraggable} from './interfaces/i_draggable.js';
-import type {IFocusableNode} from './interfaces/i_focusable_node.js';
-import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {IIcon} from './interfaces/i_icon.js';
 import * as internalConstants from './internal_constants.js';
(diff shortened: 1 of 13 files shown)


rachel-fenichel (approved):
Reverting LGTM so that we can test and commit code in keyboard-experimentation today/early Monday.

@BenHenning the issues we saw were:
- `i` and `t` no longer worked from the workspace
- Focus was wrong when clicking on a block in the workspace

BenHenning:
> Reverting LGTM so that we can test and commit code in keyboard-experimentation today/early Monday.
> 
> @BenHenning the issues we saw were:
> 
> * `i` and `t` no longer worked from the workspace
> * Focus was wrong when clicking on a block in the workspace

Thanks for the added context @rachel-fenichel. I just tested this and it seems the first point was broken in #8920 (still investigating why). I don't understand what is meant by the second point. When I click on a block in the workspace, it seems to select find (and using arrow keys works correctly from that click point).

I did notice, however, that tabbing to the workspace causes the first selection to be wrong (and immediately goes passive). This was regressed in #8916.

RoboErikG:
Regarding the focus going passive, when I was investigating I found that the changes in core and in the kb experimentation repos were both trying to perform focus logic on the first click, which caused both blocklyActiveFocus (from core) and passiveBlockFocus (from kb exp) to get applied, leading to both styles being applied. I'm not sure if there were other effects as a result of that.

PR #8934, opened 2025-04-28
fix: Fix bug in `BlockSvg.prototype.setParent`

### LINKED ISSUE #8853: Hide connection previews in `revertDrag`
### Check for duplicates

### Problem

While implementing constrained drag in keyboard experimentation, I found that `revertDrag` in `BlockDragStrategy` does not call `connectionPreviewer.hidePreview`.

So far this has been safe because drags are only reverted when dropped onto an invalid location.

With the option to end a drag with escape, it's important to fully clean up in `revertDrag`.

### Request

Call `connectionPreviwer.hidePreview` in `revertDrag`. Should be a no-op if no preview was being shown.

### Alternatives considered

_No response_

### Additional context

_No response_

cpcallen:
## The basics

### Resolves

Fix an incorrect assumption in `setParent` when handling `.setParent(null)` calls.

Fixes #8853.

(But see comments on that bug for detail.)

### Proposed Changes

* In `BlockSvg.prototype.setParent`, when setting parent to `null`, check to see if the top-most `blocklyDragging` element is in fact a direct child of the canvass before attempting to insert the new top-level block ahead of it in the canvass's contents.
* In `BlockDragStragegy.prototype.revertDrag`, call `connectionPreviewer.hidePreview()` earlier, before attempting to reconnect the dragged block to its original parent.  Not strictly necessary but makes more sense.

### Reason for Changes

The topmost block whose root SVG element has the `blocklyDragging` class may not actually be a top-level block on the workspace, but the existing implementation assumed that was true and would throw an error if it wasn't.  This resulted in problems when reverting a drag if an insertion marker was displayed.

### Test Coverage

Adds two new tests for `setParent` (one of which actually caught a bug that existed in an earlier version of the fix in this PR!)

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -445,6 +445,9 @@ export class BlockDragStrategy implements IDragStrategy {
       return;
     }
 
+    this.connectionPreviewer!.hidePreview();
+    this.connectionCandidate = null;
+
     this.startChildConn?.connect(this.block.nextConnection);
     if (this.startParentConn) {
       switch (this.startParentConn.type) {
@@ -470,9 +473,6 @@ export class BlockDragStrategy implements IDragStrategy {
 
     this.startChildConn = null;
     this.startParentConn = null;
-
-    this.connectionPreviewer!.hidePreview();
-    this.connectionCandidate = null;
 
     this.block.setDragging(false);
     this.dragging = false;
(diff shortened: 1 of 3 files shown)


core/dragging/block_drag_strategy.ts
```diff
@@ -445,6 +445,9 @@ export class BlockDragStrategy implements IDragStrategy {
       return;
     }
 
+    this.connectionPreviewer!.hidePreview();
```
rachel-fenichel, line 448: Switch to `?.` instead of `!.` while you're at it.

rachel-fenichel (approved):
LGTM after minor fixes.

PR #8938, opened 2025-04-29
feat: Make WorkspaceSvg and BlockSvg focusable (roll forward)

### LINKED ISSUE #8913: Workspace.

### LINKED ISSUE #8914: Blocks.

BenHenning:
_Note: This is a roll forward of #8916 that was reverted in #8933. See Additional Information below._

## The basics

### Resolves

Fixes #8913
Fixes #8914
Fixes part of #8771

### Proposed Changes

This updates `WorkspaceSvg` and `BlockSvg` to be focusable, that is, it makes the workspace a `IFocusableTree` and blocks `IFocusableNode`s.

Some important details:
- While this introduces focusable tree support for `Workspace` it doesn't include two other components that are obviously needed by the keyboard navigation plugin's playground: fields and connections. These will be introduced in subsequent PRs.
- Blocks are set up to automatically synchronize their selection state with their focus state. This will eventually help to replace `LineCursor`'s responsibility for managing selection state itself.
- The tabindex property for the workspace and its ARIA label have been moved down to the `.blocklyWorkspace` element itself rather than its wrapper. This helps address some tab stop issues that are already addressed in the plugin (via monkey patches), but also to ensure that the workspace's main SVG group interacts correctly with `FocusManager`.
- `WorkspaceSvg` is being initially set up to default to its first top block when being focused for the first time. This is to match parity with the keyboard navigation plugin, however the latter also has functionality for defaulting to a position when no blocks are present. It's not clear how to actually support this under the new focus-based system (without adding an ephemeral element on which to focus), or if it's even necessary (since the workspace root can hold focus).
- `css.ts` was updated to remove `blocklyActiveFocus` and `blocklyPassiveFocus` since these have unintended highlighting consequences that aren't actually desirable yet. Instead, the exact styling for active/passive focus will be iterated in the keyboard navigation plugin project and moved to Core once finalized.

### Reason for Changes

### Test Coverage

### Documentation

### Additional Information

This was originally merged in #8916 but was reverted in #8933 due to https://github.com/google/blockly-keyboard-experimentation/issues/481. This actually contains no differences from the original PR except for `css.ts` which are documented above. It does employ a new merge strategy: all of the necessary PRs to move both Core and the plugin over to using `FocusManager` will be staged and merged in quick succession as ensuring the plugin works for each constituent change (vs. the final one) is quite complex. Thus, this PR *does* break the plugin, and won't be merged until its subsequent PRs are approved and also ready for merging.

Edit: See https://github.com/google/blockly/pull/8938#issuecomment-2843589525 for why this actually is being merged a bit sooner than originally planned. Keeping the original reasoning above for context.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -44,6 +44,8 @@ import {IContextMenu} from './interfaces/i_contextmenu.js';
 import type {ICopyable} from './interfaces/i_copyable.js';
 import {IDeletable} from './interfaces/i_deletable.js';
 import type {IDragStrategy, IDraggable} from './interfaces/i_draggable.js';
+import type {IFocusableNode} from './interfaces/i_focusable_node.js';
+import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {IIcon} from './interfaces/i_icon.js';
 import * as internalConstants from './internal_constants.js';
 import {MarkerManager} from './marker_manager.js';
@@ -76,7 +78,8 @@ export class BlockSvg
     IContextMenu,
     ICopyable<BlockCopyData>,
     IDraggable,
-    IDeletable
+    IDeletable,
+    IFocusableNode
 {
   /**
    * Constant for identifying rows that are to be rendered inline.
@@ -210,6 +213,7 @@ export class BlockSvg
 
     // Expose this block's ID on its top-level SVG group.
     this.svgGroup.setAttribute('data-id', this.id);
+    svgPath.id = this.id;
 
     this.doInit_();
   }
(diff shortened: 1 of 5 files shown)


BenHenning (commented):
Self-reviewed.

BenHenning:
PTAL @rachel-fenichel. FYI that this should be identical to #8916 with the addition of the `css.ts` changes.

@cpcallen feel free to take a pass, but Rachel has agreed to review the entire chain of PRs up to #8941.

rachel-fenichel (approved):
LGTM.

BenHenning:
Thanks @rachel-fenichel!

Holding off on merging this until https://github.com/google/blockly-keyboard-experimentation/pull/482 is reviewed & approved.

BenHenning:
Per a discussion this morning in stand-up, we're going to start merging PRs even though it breaks the keyboard navigation plugin. This is to ensure that changes later in the chain (i.e. #8941) have fewer changes to sync with similar ongoing work by @gonfunko and to ensure that there are fewer changes overall to reason about when considering other plugin or v12 changes over the next few development days.

https://github.com/google/blockly-keyboard-experimentation/pull/482#issuecomment-2843579597 makes me more confident in starting the merges now. Once my local changes are pushed to that branch, things should be relatively testable even there are likely to still be some regressions.

Issue #8910, opened 2025-04-21
Shore up testing coverage for FocusManager


BenHenning:
#8909 introduces a bunch of FocusManager changes that should be thoroughly tested.

I plan to do this myself, so I won't go into detail here on exactly what tests should be added.

PR #8941, opened 2025-04-29
feat: Update line cursor to use focus manager

### LINKED ISSUE #8940: Use FocusManager for LineCursor
### Check for duplicates

### Problem

`LineCursor` needs to rely on `FocusManager` rather than selection in order to tie together focus management and keyboard navigation (which is a critical requirement for the complete end-to-end accessibility experience).

### Request

This essentially requires that `LineCursor` use the current focused node by default and fall back to selection if that fails (though we will eventually be able to remove the selection bit entirely).

### Alternatives considered

_No response_

### Additional context

Note that this is essentially the tracking issue for Core changes to make https://github.com/google/blockly-keyboard-experimentation/issues/142 possible. There are also prerequisite changes (such as critical components from #8771) and follow-up changes (specifically in the keyboard navigation plugin) in order for these changes to fully work.

### LINKED ISSUE #8954: Cursor tests fail because focus manager is not reset correctly
### Check for duplicates

### Description

The tests in `cursor_test.js`, in the `Movement` suite, fail with traces like this:
```
‣
Error: No highlight SVG found corresponding to this connection.
    at RenderedConnection.getFocusableElement (file:///usr/local/google/home/bhenning/opensource/blockly/build/src/core/rendered_connection.js:556:15)
    at FocusManager.focusNode (file:///usr/local/google/home/bhenning/opensource/blockly/build/src/core/focus_manager.js:198:87)
    at LineCursor.setCurNode (file:///usr/local/google/home/bhenning/opensource/blockly/build/src/core/keyboard_nav/line_cursor.js:475:31)
    at Context.<anonymous> (cursor_test.js:129:19)
```

In most cases, the root cause is that the focus manager state is not correctly cleared during teardown between tests. The focus manager still holds a location and attempts to set passive focus on it, but fails because there is no corresponding svg element to update.

The issue can be resolved by adding a reset function in focus manager and using it during teardown:
`export function resetFocusManager() { focusManager = null;`

This implies an actual issue with focus manager cleanup that @BenHenning will investigate.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #8955: Cursor test fails: In - From output connection
### Check for duplicates

### Description

This test fails after @BenHenning's focus manager changes to the line cursor:
```
    test('In - From output connection', function () {
      const fieldBlock = this.blocks.E;
      const outputNode = ASTNode.createConnectionNode(
        fieldBlock.outputConnection,
      );
      this.cursor.setCurNode(outputNode);
      this.cursor.in();
      assert.equal(curNode.getLocation(), fieldBlock);
    });
```

The proximate cause is that output connections are not given highlight paths in `drawer.updateConnectionHighlights` because they live at the block level instead of in the rows.

In the keyboard experimentation plugin we sometimes (always?) use the block's node or the parent input connection's node, rather than placing the cursor on an output connection.

If the cursor is allowed to go to an output connection, `updateConnectionHighlights` needs to be updated.

If the cursor cannot ever be placed at an output connection, the direct call to `this.cursor.setCurNode(outputNode);` should throw an error.

### Local test updates
To make this test useful again, the initial cursor location can be block A's first input connection. Moving in from there should reach the field block node.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #8940
Fixes #8954
Fixes #8955

### Proposed Changes

This updates `LineCursor` to use `FocusManager` rather than selection (principally) as the source of truth.

### Reason for Changes

Ensuring that keyboard navigation works correctly with eventual screen reader support requires ensuring that ever navigated component is focused, and this is primarily what `FocusManager` has been designed to do. Since these nodes are already focused, `FocusManager` can be used as the primary source of truth for determining where the user currently has navigated, and where to go next.

Previously, `LineCursor` relied on selection for this purpose, but selection is now automatically updated (for blocks) using focus-controlled `focus` and `blur` callbacks. Note that the cursor will still fall back to synchronizing with selection state, though this will be removed once the remaining work to eliminate `MarkerSvg` has concluded (which requires further consideration on the keyboard navigation side viz-a-viz styling and CSS decisions) and once mouse clicks are synchronized with focus management.

Note that the changes in this PR are closely tied to https://github.com/google/blockly-keyboard-experimentation/pull/482 as both are necessary in order for the keyboard navigation plugin to correctly work with `FocusManager`.

Some other noteworthy changes:
- Some special handling exists for flyouts to handle navigating across stacks (per the current cursor design).
- `FocusableTreeTraverser` is needed by the keyboard navigation plugin (in https://github.com/google/blockly-keyboard-experimentation/pull/482) so it's now being exported.
- `FocusManager` had one bug that's now patched and tested in this PR: it didn't handle the case of the browser completely forcing focus loss. It would continue to maintain active focus even though no tracked elements now hold focus. One such case is the element being deleted, but there are other cases where this can happen (such as with dialog prompts).
- `FocusManager` had some issues from #8909 wherein it would overeagerly call tree focus callbacks and slightly mismanage the passive node. Since tests haven't yet been added for these lifecycle callbacks, these cases weren't originally caught (per #8910).
- `FocusManager` was updated to move the tracked manager into a static function so that it can be replaced in tests. This was done to facilitate changes to setup_teardown.js to ensure that a unique `FocusManager` exists _per-test_. It's possible for DOM focus state to still bleed across tests, but `FocusManager` largely guarantees eventual consistency. This change prevents a class of focus errors from being possible when running tests.
- A number of cursor tests needed to be updated to ensure that a connections are properly rendered (as this is a requirement for focusable nodes, and cursor is now focusing nodes). One test for output connections was changed to use an input connection, instead, since output connections can no longer be navigated to (and aren't rendered, thus are not focusable). It's possible this will need to be changed in the future if we decide to reintroduce support for output connections in cursor, but it seems like a reasonable stopgap. Huge thanks to @rachel-fenichel for helping investigate and providing an alternative for the output connection test.

**Current gaps** to be fixed after this PR is merged:
- The flyout automatically closes when creating a variable with with keyboard or mouse (I think this is only for the keyboard navigation plugin). I believe this is a regression from previous behavior due to how the navigation plugin is managing state. It would know the flyout should be open and thus ensure it stays open even when things like dialog prompts try to close it with a blur event. However, the new implementation in https://github.com/google/blockly-keyboard-experimentation/pull/482 complicates this since state is now inferred from `FocusManager`, and the flyout _losing_ focus will force it closed. There was a fix introduced in this PR to fix it for keyboard navigation, but fails for clicks because the flyout never receives focus when the create variable button is clicked. It also caused the advanced compilation tests to fail due to a subtle circular dependency from importing `WorkspaceSvg` directly rather than its type.
- The flyout, when it stays open, does not automatically update past the first variable being created without closing and reopening it. I'm actually not at all sure why this particular behavior has regressed. This is only relevant for the previously mentioned fix which is no longer part of this PR.

### Test Coverage

No new non-`FocusManager` tests have been added. It's certainly possible to add unit tests for the focusable configurations being introduced in this PR, but it may not be highly beneficial. It's largely assumed that the individual implementations should work due to a highly tested FocusManager, and it may be the case that the interactions of the components working together is far more important to verify (that is, the end user flows). The latter is planned to be tackled as part of #8915.

Some new `FocusManager` tests were added, but more are still needed and this is tracked as part of #8910.

### Documentation

No new documentation should be needed for these changes.

### Additional Information


BenHenning (commented):
Self-reviewed changes.

BenHenning:
PTAL @rachel-fenichel. Note that this is not mergeable yet, but it's at least in a good state for review. There's a 'current gaps' section in the PR description that explains the current items that will need to fixed before this can be merged. Aside from that, there may be additional changes required to Core as part of testing https://github.com/google/blockly-keyboard-experimentation/pull/482 which I'll likely pull into this PR to avoid another PR to be created. However, if you have strong opinions about getting this merged sooner (to reduce conflicts with @gonfunko's changes), I can pull out the variables change in this PR and focus on fixing the tests and merging it.

gonfunko (approved):

BenHenning (commented):
Self-reviewed latest changes. There's still 1 Mocha test failing yet in an unrelated test suite--investigating.

BenHenning (commented):
Self-reviewed correction to de-initialization. Tests should pass now.

BenHenning:
Mocha's passing, but advanced compilation in browser isn't. Digging.

BenHenning:
Turns out the failure was due to importing `WorkspaceSvg` directly in `variables.ts` rather than its type. This caused a circular dependency in the advanced compilation `main.js` which led to the 'healthCheck' function not being declared (leading the the obscure failure). Unfortunately the normal circular dependency routines didn't catch this and it produced a runtime failure. It's surprising this didn't trigger when testing the changes elsewhere, but it probably would've affected downstream Blockly projects so it's good that advanced compilation caught it.

BenHenning:
PTAL @gonfunko for the latest few changes or @rachel-fenichel. One approval will auto-merge the PR.

I want to get a double check on the test changes since it affects every Mocha test (though they are, fortunately, passing).

rachel-fenichel (approved):
LGTM with a potential follow-up down the line: as I understand it, you've changed the test code so that the focus manager won't keep state between tests. There is still a potential problem when a workspace has focus, the workspace is deleted, and the focus manager tries to mark the (dead) node as passively focused. I think you're going to need to fix that case in focus manager. If that's correct, please file a bug for it.

BenHenning:
> LGTM with a potential follow-up down the line: as I understand it, you've changed the test code so that the focus manager won't keep state between tests. There is still a potential problem when a workspace has focus, the workspace is deleted, and the focus manager tries to mark the (dead) node as passively focused. I think you're going to need to fix that case in focus manager. If that's correct, please file a bug for it.

I've filed #8957 for this case. I'm...not sure that this scenario is actually possible outside of tests (at least realistically). I definitely saw the same stack traces when running the tests, but I think the test environment creates a situation that's very different from a real production use case, specifically in how it manages its DOM. My tests so far in trying to deliberately break `FocusManager` in this way haven't worked, even in tests, because removing an actively focused DOM element will result in a focus event being fired and thus giving `FocusManager` an opportunity to clean up its state. I think a DOM needs to be abandoned but not removed to reproduce this issue which is highly contrived.

PR #8961, opened 2025-05-02
fix!: Remove alt+key commands

RoboErikG:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/258

### Proposed Changes

Remove the alt+key code version of commands

### Reason for Changes

Using alt is non-standard and may conflict with window commands, so we're removing it.

### Test Coverage

Removed the tests that expect alt to work.


rachel-fenichel (approved):
Change the pr title to `fix!` to indicate that this is a breaking change for release notes purposes.

PR #8981, opened 2025-05-02
fix: Synchronize gestures and focus

### LINKED ISSUE #8950: Returning ephemeral focus moves focus to the toolbox
### Check for duplicates

### Description

Sometimes returning ephemeral focus from a text editing field opens the flyout or moves focus to the most recently opened toolbox category.

### Reproduction steps

1. Open the blockly playground
2. Drag in a block with text fields (e.g. `1+1`)
3. Click on a text field
4. Click on the workspace
5. Observe that the flyout is not focused
6. Click on a text field
7.  Type text and hit `Enter` to confirm
8. Observe that the flyout opens and that the focus rectangle moves to a toolbox category
9. Click on the workspace to close the flyout
10. Click on the text field
11. Click on the workspace
12. Observe that the focus rectangle moves to a toolbox category

### Screenshots

Image

### LINKED ISSUE #8952: Cannot paste when a block is selected
### Check for duplicates

### Description

Selecting a block and using `ctrl+c` to copy and then `ctrl+v` to paste sometimes fails in core.

This is a regression and v12 blocking.

Expected behaviour: clicking on a block and hitting `ctrl+c` copies it. Successive `ctrl+v`s paste the block offset down and to the right.

Actual behaviour: on the first block, behaves correctly.
Clicking on another block and repeating the same actions does nothing: no visible copy, no visible paste, and no visible failure. The clipboard is never updated to contain new copied blocks.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #8971: Error caused when you keyboard navigate to a block, cut (CMD+X), paste (CMD+V), then undo (CMD+Z) via keyboard shortcuts.
### Check for duplicates

- [ ] Error caused when you keyboard navigate to a block, cut (CMD+X), paste (CMD+V), then undo (CMD+Z) via keyboard shortcuts.

### Description

Error caused when you keyboard navigate to a block, cut (CMD+X), paste (CMD+V), then undo (CMD+Z) via keyboard shortcuts:
Ben: unsure about this one. The stack trace is suggesting that undo is repeating the deletion which is causing the cursor to attempt to focus an old block that's no longer in the workspace. That's my interpretation of what would cause this failure, but it may be something else. The '???' is particularly interesting since that's the node being passed to focusNode. This is a particularly excellent find.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #8952
Fixes #8950
Fixes #8971

Fixes a bunch of other keyboard + mouse synchronization issues found during the testing discussed in https://github.com/google/blockly-keyboard-experimentation/pull/482#issuecomment-2846341307.

### Proposed Changes

This introduces a number of changes to:
- Ensure that gestures which change selection state also change focus.
- Ensure that ephemeral focus is more robust against certain classes of failures.

### Reason for Changes

There are some ephemeral focus issues that can come up with certain actions (like clicking or dragging) don't properly change focus. Beyond that, some users will likely mix clicking and keyboard navigation, so it's essential that focus and selection state stay in sync when switching between these two types of navigation modalities.

Other changes:
- Drop-down div was actually incorrectly releasing ephemeral focus before animated closes finish which could reset focus afterwards, breaking focus state.
- Both drop-down and widget divs have been updated to only return focus _after_ marking the workspace as focused since the last focused node should always be the thing returned.
- In a number of gesture cases selection has been removed. This is due to `BlockSvg` self-managing its selection state based on focus, so focusing is sufficient. I've also centralized some of the focusing calls (such as putting one in `bringBlockToFront` to ensure focusing happens after potential DOM changes).
- Similarly, `BlockSvg`'s own `bringToFront` has been updated to automatically restore focus after the operation completes. Since `bringToFront` can always result in focus loss, this provides robustness to ensure focus is restored.
- Block pasting now ensures that focus is properly set, however a delay is needed due to additional rendering changes that need to complete (I didn't dig deeply into the _why_ of this).
- Dragging has been updated to always focus the moved block if it's not in the process of being deleted.
- There was some selection resetting logic removed from gesture's `doWorkspaceClick` function. As far as I can tell, this won't be needed anymore since blocks self-regulate their selection state now.
- `FocusManager` has a new extra check for ephemeral focus to catch a specific class of failure where the browser takes away focus immediately after it's returned. This can happen if there are delay timing situations (like animations) which result in a focused node being deleted (which then results in the document body receiving focus). The robustness check is possibly not needed, but it help discover the animation issue with drop-down div and shows some promise for helping to fix the variables-closing-flyout problem.

Some caveats:
- Some undo/redo steps can still result in focus being lost. This may introduce some regressions for selection state, and definitely introduces some annoyances with keyboard navigation. More work will be needed to understand how to better redirect focus (and to what) in cases when blocks disappear.
- There are many other places where focus is being forced or selection state being overwritten that could, in theory, cause issues with focus state. These may need to be fixed in the future.
- There's a lot of redundancy with `focusNode` being called more than it needs to be. `FocusManager` does avoid calling `focus()` more than once for the same node, but it's possible for focus state to switch between multiple nodes or elements even for a single gesture (for example, due to bringing the block to the front causing a DOM refresh). While the eventual consistency nature of the manager means this isn't a real problem, it may cause problems with screen reader output in the future and warrant another pass at reducing `focusNode` calls (particularly for gestures and the click event pipeline).

### Test Coverage

This PR is largely relying on existing tests for regression verification, though no new tests have been added for the specific regression cases.

#8910 is tracking improving `FocusManager` tests which could include some cases for the new ephemeral focus improvements.

#8915 is tracking general accessibility testing which could include adding tests for these specific regression cases.

### Documentation

No new documentation is expected to be needed here.

### Additional Information

These changes originate from both #8875 and from a branch @rachel-fenichel created to investigate some of the failures this PR addresses. These changes have also been verified against both Core's playground and the keyboard navigation plugin's test environment.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -34,6 +34,7 @@ import type {BlockMove} from './events/events_block_move.js';
 import {EventType} from './events/type.js';
 import * as eventUtils from './events/utils.js';
 import {FieldLabel} from './field_label.js';
+import {getFocusManager} from './focus_manager.js';
 import {IconType} from './icons/icon_types.js';
 import {MutatorIcon} from './icons/mutator_icon.js';
 import {WarningIcon} from './icons/warning_icon.js';
@@ -1268,6 +1269,7 @@ export class BlockSvg
    *     adjusting its parents.
    */
   bringToFront(blockOnly = false) {
+    const previouslyFocused = getFocusManager().getFocusedNode();
     /* eslint-disable-next-line @typescript-eslint/no-this-alias */
     let block: this | null = this;
     if (block.isDeadOrDying()) {
@@ -1284,6 +1286,13 @@ export class BlockSvg
       if (blockOnly) break;
       block = block.getParent();
     } while (block);
+    if (previouslyFocused) {
+      // Bringing a block to the front of the stack doesn't fundamentally change
+      // the logical structure of the page, but it does change element ordering
+      // which can take automatically take away focus from a node. Ensure focus
+      // is restored to avoid a discontinuity.
+      getFocusManager().focusNode(previouslyFocused);
+    }
   }
 
   /**
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -60,7 +60,7 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     // Sometimes there's a delay before the block is fully created and ready for
     // focusing, so wait slightly before focusing the newly pasted block.
     const nodeToFocus: IFocusableNode = block;
-    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 1);
+    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 0);
     return block;
   }
 }
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -12,6 +12,7 @@ import * as eventUtils from '../events/utils.js';
 import {getFocusManager} from '../focus_manager.js';
 import {ICopyData} from '../interfaces/i_copyable.js';
 import {IPaster} from '../interfaces/i_paster.js';
+import * as renderManagement from '../render_management.js';
 import {State, append} from '../serialization/blocks.js';
 import {Coordinate} from '../utils/coordinate.js';
 import {WorkspaceSvg} from '../workspace_svg.js';
@@ -60,7 +61,9 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     // Sometimes there's a delay before the block is fully created and ready for
     // focusing, so wait slightly before focusing the newly pasted block.
     const nodeToFocus: IFocusableNode = block;
-    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 0);
+    renderManagement
+      .finishQueuedRenders()
+      .then(() => getFocusManager().focusNode(nodeToFocus));
     return block;
   }
 }
--- a/core/clipboard/block_paster.ts
+++ b/core/clipboard/block_paster.ts
@@ -5,10 +5,11 @@
  */
 
 import {BlockSvg} from '../block_svg.js';
-import * as common from '../common.js';
+import {IFocusableNode} from '../blockly.js';
 import {config} from '../config.js';
 import {EventType} from '../events/type.js';
 import * as eventUtils from '../events/utils.js';
+import {getFocusManager} from '../focus_manager.js';
 import {ICopyData} from '../interfaces/i_copyable.js';
 import {IPaster} from '../interfaces/i_paster.js';
 import {State, append} from '../serialization/blocks.js';
@@ -55,7 +56,11 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     if (eventUtils.isEnabled() && !block.isShadow()) {
       eventUtils.fire(new (eventUtils.get(EventType.BLOCK_CREATE))(block));
     }
-    common.setSelected(block);
+
+    // Sometimes there's a delay before the block is fully created and ready for
+    // focusing, so wait slightly before focusing the newly pasted block.
+    const nodeToFocus: IFocusableNode = block;
+    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 1);
     return block;
   }
 }
--- a/core/dragging/dragger.ts
+++ b/core/dragging/dragger.ts
@@ -133,8 +133,8 @@ export class Dragger implements IDragger {
     eventUtils.setGroup(false);
 
     if (!wouldDelete && isFocusableNode(this.draggable)) {
-      // Ensure focusable nodes that have finished dragging (but aren't being)
-      // deleted end with focus and selection.
+      // Ensure focusable nodes that have finished dragging (but aren't being
+      // deleted) end with focus and selection.
       getFocusManager().focusNode(this.draggable);
     }
   }
--- a/core/dragging/dragger.ts
+++ b/core/dragging/dragger.ts
@@ -8,11 +8,13 @@ import * as blockAnimations from '../block_animations.js';
 import {BlockSvg} from '../block_svg.js';
 import {ComponentManager} from '../component_manager.js';
 import * as eventUtils from '../events/utils.js';
+import {getFocusManager} from '../focus_manager.js';
 import {IDeletable, isDeletable} from '../interfaces/i_deletable.js';
 import {IDeleteArea} from '../interfaces/i_delete_area.js';
 import {IDragTarget} from '../interfaces/i_drag_target.js';
 import {IDraggable} from '../interfaces/i_draggable.js';
 import {IDragger} from '../interfaces/i_dragger.js';
+import {isFocusableNode} from '../interfaces/i_focusable_node.js';
 import * as registry from '../registry.js';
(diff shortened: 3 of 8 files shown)


core/clipboard/block_paster.ts
```diff
@@ -55,7 +56,11 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     if (eventUtils.isEnabled() && !block.isShadow()) {
       eventUtils.fire(new (eventUtils.get(EventType.BLOCK_CREATE))(block));
     }
-    common.setSelected(block);
+
+    // Sometimes there's a delay before the block is fully created and ready for
+    // focusing, so wait slightly before focusing the newly pasted block.
+    const nodeToFocus: IFocusableNode = block;
+    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 1);
```
BenHenning, line 63: Try 1.

core/block_svg.ts
```diff
@@ -1284,6 +1286,13 @@ export class BlockSvg
       if (blockOnly) break;
       block = block.getParent();
     } while (block);
+    if (previouslyFocused) {
```
BenHenning, line 1289: Document this in PR description.

core/field_dropdown.ts
```diff
@@ -769,7 +769,7 @@ export class FieldDropdown extends Field<string> {
       } else if (typeof option[1] !== 'string') {
         foundError = true;
         console.error(
-          `Invalid option[${i}]: Each FieldDropdown option id must be a string. 
+          `Invalid option[${i}]: Each FieldDropdown option id must be a string.
```
BenHenning, line 772: Revert this file.

BenHenning (commented):
Self-reviewed PR.

core/block_svg.ts
BenHenning: Done.

core/clipboard/block_paster.ts
BenHenning: Done in latest.

core/field_dropdown.ts
BenHenning: Done in latest.

core/gesture.ts
```diff
@@ -928,11 +941,7 @@ export class Gesture {
    * @param _e A pointerup event.
    */
   private doWorkspaceClick(_e: PointerEvent) {
-    const ws = this.creatorWorkspace;
```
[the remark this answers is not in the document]
BenHenning: Done.

BenHenning (commented):
Self-reviewed changes.

core/clipboard/block_paster.ts
```diff
@@ -55,7 +56,11 @@ export class BlockPaster implements IPaster<BlockCopyData, BlockSvg> {
     if (eventUtils.isEnabled() && !block.isShadow()) {
       eventUtils.fire(new (eventUtils.get(EventType.BLOCK_CREATE))(block));
     }
-    common.setSelected(block);
+
+    // Sometimes there's a delay before the block is fully created and ready for
+    // focusing, so wait slightly before focusing the newly pasted block.
+    const nodeToFocus: IFocusableNode = block;
+    setTimeout(() => getFocusManager().focusNode(nodeToFocus), 0);
```
rachel-fenichel, line 63: This delay may be related to the `bumpNeighbours` behaviour.

core/dragging/dragger.ts
```diff
@@ -129,6 +131,12 @@ export class Dragger implements IDragger {
       root.dispose();
     }
     eventUtils.setGroup(false);
+
+    if (!wouldDelete && isFocusableNode(this.draggable)) {
+      // Ensure focusable nodes that have finished dragging (but aren't being)
```
rachel-fenichel, line 136: "(but aren't being) deleted" -> "(but aren't being deleted)"

rachel-fenichel (commented):
Re: "changes have been verified", what did you do to manually verify?

core/clipboard/block_paster.ts
BenHenning: Hmm that seems really plausible, though I can't find a call to bump in this call path but I'm sure I'm missing large portions of what's happening here.

This does make me think that it might be nice to have a way to attach logic to run with the next rendering pass so that we can predictably run things that depend on rendering (vs. using a timeout and hoping it's correct).

core/dragging/dragger.ts
BenHenning: Nice catch--fixed!

BenHenning:
> Re: "changes have been verified", what did you do to manually verify?

I can re-run these steps with a video recording if that's helpful, but here's what I did:

**Core playground**
- Dragged to insert a block. While it was selected, attempted to copy and paste it several times. Paste worked. This demonstrates a fix to #8952.
  - Did the same thing with click insertion.
  - Did the same thing with an existing block that I then dragging to put it into the "dragged" state which broke pasting earlier in my repros.
- Added a binary operation block and edited one of the operators text fields. Saw the flyout did not reopen, demonstrating a fix for #8950.
  - Did the same for selecting an operator.
  - Repeated both of these (editing the field and changing the operator) for both dragging the block onto the workspace and clicking to add it.

**Keyboard nav environment**
- Many iterations on clicking different blocks and fields, then using keyboard navigation to ensure that state is correctly synchronized. Confirmed the same flows with just keyboard navigation to ensure that there weren't regressions (which happened a few times when trying to fix field input). This verified fixes for most of the selection issues brought up in the focus issues findings document.
- Verified both dragging and clicking to insert blocks onto the workspace to ensure the block maintains selection.
- Multiple iterations of dragging blocks between stacks, between blocks, and on/off stacks to ensure that they remain highlighted. Repeated some of these with keyboard movement to verify some of the cases in the focus issues findings document.
- Multiple iterations of copying and pasting blocks to ensure that pasting maintained the correct selection (new block only). Also tried cutting to verify that #8971 has also been fixed.

There may be other things I've tried, but I think this covers the majority. If there's anything else that you think is worth testing or re-testing, please let me know.

BenHenning:
PTAL @rachel-fenichel to the latest changes & my comment above.

rachel-fenichel:
Thanks for the list--that gives us things to turn into automated tests down the line, and means that you can ask other folks to reverify for you if issues pop up.

core/clipboard/block_paster.ts
rachel-fenichel:
```
/**
 * @returns A promise that resolves after the currently queued renders have
 *     been completed.
 */
export function finishQueuedRenders(): Promise<void> {
  // If there are no queued renders, return a resolved promise so `then`
  // callbacks trigger immediately.
  return afterRendersPromise ? afterRendersPromise : Promise.resolve();
}
```

You can find examples of this pattern in the codebase. I think it will do what you want.

BenHenning: I evidently didn't look hard enough for this exact mechanism. Nice suggestion--I updated the code and tested pasting again. Selection seems to correctly select the most recently pasted block as intended.

BenHenning (commented):
Self-reviewed latest.

BenHenning:
PTAL again @rachel-fenichel. I think everything's been addressed so far.

rachel-fenichel (approved):

Issue #8968, opened 2025-05-02
When navigating to focus on the colour input field, the parent node is selected


sappm01:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Description

When navigating to focus on the colour input field, the parent node is selected as well. Would have expected only the field to be outlined.
Ben: unsure how to fix, but need to investigate. This is almost certainly due to selection inconsistency.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

maribethb:
Fixed by #9004

Issue #8962, opened 2025-05-02
Remove MarkerSvg


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

MarkerSvg has essentially been entirely replaced by the capability of active/passive focus highlighting

### Request

To remove MarkerSvg.

PR #8991, opened 2025-05-05
fix!: remove MarkerSvg and uses

rachel-fenichel:
## The basics

### Resolves

Draft fix for https://github.com/google/blockly/issues/8962

### Proposed Changes
[Commit 1](01e1aa5e25c98dee76f444b964537868594650c3): Delete the actual drawer, subclasses, and uses
[Commit 2](1601319c820907e0c41a8bbc298fca50c6314f60): Delete residual SVG elements that were rendered through the cursor/marker flow instead of through the normal block flow.
Commits 3 and 4: lint and format.

### Reason for Changes

`MarkerSvg` and friends represented a way to draw the cursor that was outside of the standard rendering flow. These SVG elements were children of the standard block (field, workspace, etc) SVG elements, but were updated through explicit calls to `setCurNode` on the cursor.

Ben's focus changes updated focus rendering to use elements that are rendered/updated through the standard rendering flow. Where possible, focus is shown using CSS rather than by creating new elements.

### Additional Information

- keyboard-experimentation will need a [related PR](https://github.com/google/blockly-keyboard-experimentation/pull/490) to remove explicit calls to `hide()`. 
- The old keyboard navigation plugin will not compile after this change (if it even compiled before).
- `IASTNodeLocationSvg` can be removed as a follow-up, if it hasn't already been removed during refactors of `ASTNode`.


rachel-fenichel:
Does not fix https://github.com/google/blockly/issues/8968

core/workspace_svg.ts
```diff
@@ -1327,10 +1327,6 @@ export class WorkspaceSvg
       .flatMap((block) => block.getDescendants(false))
       .filter((block) => block.isInsertionMarker())
       .forEach((block) => block.queueRender());
-
-    renderManagement
-      .finishQueuedRenders()
```
cpcallen, line 1332: Should `render` no longer call `renderManagement.finishQueuedRenders()` at all?

(Maybe the queued renderers will be finished automatically, and the method call is not needed if you don't need the resulting promise?)

core/workspace_svg.ts
```diff
@@ -1327,10 +1327,6 @@ export class WorkspaceSvg
...
-      .then(() => void this.markerManager.updateMarkers());
```
cpcallen, line 1333: Should `render` no longer call `renderManagement.finishQueuedRenders()` at all?

(Maybe the queued renderers will be finished automatically, and the method call is not needed if you don't need the resulting promise?)

core/interfaces/i_ast_node_location_svg.ts
```diff
@@ -11,6 +11,4 @@ import type {IASTNodeLocation} from './i_ast_node_location.js';
 /**
  * An AST node location SVG interface.
  */
-export interface IASTNodeLocationSvg extends IASTNodeLocation {
-
-}
+export interface IASTNodeLocationSvg extends IASTNodeLocation {}
```
cpcallen, line 14: Could this interface not be deleted entirely, now that it is empty?

I suppose we could formally deprecate it first, but since anyone who was calling `setFooSvg` on an `IASTNodeLocationSvg` instance already has broken code I'm not sure it's worth it.

cpcallen (approved):
Opinion: I would (and now have) described issue #8962 as a task rather than a bug, so I'd have titled this PR and its main commits as `chore:` rather than `fix:`.

cpcallen:
For some reason GitHub hasn't run build, format or lint workflows, and I can't figure out how to kick it in to action.

core/interfaces/i_ast_node_location_svg.ts
rachel-fenichel: Filed as https://github.com/google/blockly/issues/9006

core/workspace_svg.ts
rachel-fenichel: Yes, queued renders will finish automatically, and the method call is specifically to be able to get a promise and run after the renders have finished.

rachel-fenichel:
Workflows ran and passed.

Issue #8828, opened 2025-03-28
Convert ASTNode to an interface


rachel-fenichel:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

AST Nodes represent locations in the Blockly workspace (AST = Abstract Syntax Tree).

They were originally created with the idea that all parts of the Blockly workspace are connected in an abstract syntax tree, because they represent code, and therefore can be traversed. Working with AST Nodes let us apply a variety of traversal algorithms (depth-first, breadth-first, etc) to make it possible to move a cursor through the tree using only arrow keys.

The first crack in this idea came when we allowed AST Nodes to represent locations on the workspace, with coordinates. These don't fit nicely into the idea of an AST.

In the context of keyboard navigation, we've found that there are a lot of different places that we might want a cursor to be: In the toolbox, moving the cursor moves selection between toolbox categories. In the flyout, it moves between stacks of blocks, buttons, and labels. On the workspace, we need to also consider comments (block and workspace) and icons.

A lot of the existing code in `ast_node.ts` consists of switch statements to figure out what type the node is and then do some type-specific work. In this sense it's not providing a useful abstraction.

### Request

- [ ] Design an appropriate interface
- [ ] Change ASTNode to an interface instead of a class, and use it in the cursor
- [ ] (Consider) Choose a new name (Location? Navigable?)
- [ ] Move code related to finding next and previous locations into the cursor class
- [ ] Delete old code
- [ ] Write or update cursor tests

### Alternatives considered

- Use `IFocusableNode` instead of creating a new interface. But this ties focus and navigation together too closely and may make multiselection difficult.
- Use `ISelectable` instead of creating a new interface. But not every cursor position is selectable (e.g. toolbox categories, connections)

### Additional context

We eventually want to support multi-selection. I think the cursor may be responsible for tracking that.

PR #8992, opened 2025-05-05
refactor!: Use navigation rulesets instead of ASTNode to control keyboard navigation.

gonfunko:
## The basics

### Resolves

Fixes #[8828](https://github.com/google/blockly/issues/8828)

### Proposed Changes
This PR should be reviewed commit-by-commit.

It introduces two new interfaces, `INavigable`, which represents an entity which can be navigated to using the keyboard, and `INavigationPolicy`, which represents a set of rules and logic that control where navigation should proceed to starting from a given `INavigable`. It makes the core navigable classes (`WorkspaceSvg`, `BlockSvg`, `RenderedConnection`, `Field`, `FlyoutButton`) conform to these, adds `INavigationPolicy` rulesets for each, updates the `LineCursor` to use them instead of `ASTNode`, and removes the corresponding logic from `ASTNode`.

Each workspace has its own `Navigator` instance which is responsible for associating navigation policies with classes. When the navigator is asked to navigate relative to a given `INavigable`, it asks it for its class, looks up the corresponding `INavigationPolicy`, and defers to that policy to determine the next entity that should receive focus.

Third-party developers may register their own NavigationPolicies with a workspace's Navigator to customize navigation behavior, either by replacing the built-in policies for the built-in types or by adding entirely new policies to allow custom navigable entities to be navigated to and from. These may subclass the built-in policies or being entirely standalone.

### Breaking Changes
* Much of `ASTNode` has been removed, and the class will be going away entirely. Use `INavigable` instead.


rachel-fenichel (commented):
A few questions about the deletion of the ASTNode tests:
- Did you check that the behaviours were still the same?
- Is there any way that you could repurpose those tests (with all of the enumerated cases) as tests for your new navigation strategies?

gonfunko:
I manually validated that the navigation behavior in the keyboard experiment was the same, and I've restored the ASTNode tests and updated them to use the Navigator instead. I removed tests related to stack nodes (because that concept no longer exists) and those that asserted you got null when you'd now loop around, and some exercising the create*Node methods and the like. The main in/out/next/prev logic tests are all passing.

rachel-fenichel (approved):
LGTM!

core/field.ts
```diff
@@ -1452,6 +1454,30 @@ export abstract class Field<T = any>
...
+    return (
+      this.isClickable() &&
+      this.isCurrentlyEditable() &&
+      !(this.getSourceBlock()?.isSimpleReporter() && this.isFullBlockField()) &&
+      this.getParentInput().isVisible()
+    );
+  }
+
+  /**
+   * Returns this field's class.
```
microbit-matt-hillsdon, line 1473: Belatedly, I think it would be good to clarify the difference between this and the [constructor property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor). When should they differ? E.g. is it OK/desirable not to override for some field subclasses? If it should always return the class/constructor, do we need this at all?

maribethb: Same question. Also, this is a breaking change, so could you please call it out in the PR description with clear instructions?

gonfunko: Unfortunately, the [constructor property is just typed as `Function`](https://github.com/microsoft/TypeScript/issues/3841), so it can't be used to retrieve the corresponding `INavigationPolicy` in a typesafe manner. Subclasses do not need to override `getClass()`, unless they want to allow for navigation policies targeting that specific subclass that are different from those targeting its parent class. In general, only subclasses that directly inherit from the abstract `Field` need to do anything, and they should just return their own class from `getClass`.

cpcallen: IMHO a breaking change adding a new `abstract` method to `Field` seems like a fairly high price to pay type safety.  Would it not be sufficient to add a cast to `Navigator.prototype.get`?:

```JS
  private get<T extends INavigable<T>>(
    key: T,
  ): INavigationPolicy<T> | undefined {
    return this.rules.get(key.constructor as new (...args: any) => T);
  }
```

I recognise that this would be a change the behaviour, insofar as (as written) one would need to register an `INavigationPolicy` for each subclass, but walking up the prototype chain if a policy isn't immediately found seems like a useful feature.

In fact to facilitate this I would suggest alternatively either:

- Using the `.prototype` objects, rather than the class constructor functions, as the keys for the `.rules` map.  Then get looks something like:
  ```JS
  private get(key) {
    while (key) {
      const policy = this.rules get(key);
      if (policy) return policy;
      key = Object.getPrototypeOf(key);
    }
    return undefined;
  }
  ```

- Or leverage property look-up to do that for you by putting the policy directly on a property of the prototype object.  Lookup becomes trivial:
  ```JS
  private get(key) {
    return key.navigationPolicy;
  }
  ```
  and to set this up one need only add a declaration to `Field` and then set some `.prototype` properties as needed:
  ```JS
  class Field {
    declare navigationPolicy?: INavigationPolicy;
  }

  Field.prototype.navigationPolicy = new FieldNavigationPolicy();
  ```

(Apologies for not having fully typed the above.)

BenHenning: Ideally we'd want to avoid child classes from having to understand what it means to implement a policy (otherwise it could put a lot more burden on implementations than simply implementing a `getClass`).

That being said, since prototypes should respect the same inheritance rules as class members, that approach is actually rather appealing, though I do have a few concerns:
- Constructing the policy class within the `Field`, `BlockSvg`, etc. may introduce very tricky circular dependencies (depending on what the import situation is for the policy class).
- Does the declaration provide a basis for non-prototype overloading of the value? It'd be a bit weird to programmatically provide a policy vs. only doing so through a prototype. Sure we can just always check the prototype in Core, but the fact that it could be set as a member seems unexpected.

For the earlier bit @cpcallen mentioned, does the `constructor` cast actually work? That could be a really clever way to replace `getClass` entirely while maintaining type safety. I would sort of expect `T.constructor` to always conform to `new (...args: any) => T` unless there is weird monkeypatching shenanigans going on. Only other case I would be unsure about is constructor inheritance, but I'd expect that to "just work" even for polymorphism.

PR #8995, opened 2025-05-06
fix: Add private to variableChangeCallback

RoboErikG:
## The basics

### Resolves

Fixes 

### Reason for Changes

Add private keyword to variableChangeCallback. This was requested for #8980 but didn't make it in because auto-merge was enabled.


gonfunko (approved):

maribethb:
It doesn't need to be `@internal` if it's `private`. But this is like P5 to fix.

PR #8997, opened 2025-05-06
fix: Improve missing node resiliency

### LINKED ISSUE #8994: Deleting a variable throws an exception
### Check for duplicates

### Description

When deleting a variable, a console error is thrown.

### Reproduction steps

1. Create new variable from the toolbox.
2. Add the variable to the workspace.
3. Right click on the variable and click to delete it.
4. Open the developer console to see the stack trace thrown.

### Stack trace

```shell
Uncaught Error: Attempting to focus node which isn't recognized by its parent tree: ???.
    at FocusManager.focusNode (focus_manager.ts:254:13)
    at focus_manager.ts:347:18
```

### Screenshots

_No response_

### Browsers

Chrome desktop

BenHenning:
## The basics

### Resolves

Fixes #8994

### Proposed Changes

This removes an error that was previously thrown by `FocusManager` when attempting to focus an invalid node (such as one that's been removed from its parent).

### Reason for Changes

https://github.com/google/blockly/issues/8994#issuecomment-2855447539 goes into more detail. While this error did cover legitimately wrong cases to try and focus things (and helped to catch some real problems), fixing this 'properly' may become a leaky boat problem where we have to track down every possible asynchronous scenario that could produce such a case. One class of this is ephemeral focus which had robustness improvements itself in #8981 that, by effect, caused this issue in the first place. Holistically fixing this with enforced API contracts alone isn't simple due to the nature of how these components interact.

This change ensures that there's a sane default to fall back on if an invalid node is passed in. Note that `FocusManager` was designed specifically to disallow defocusing a node (since fallbacks can get messy and introduce unpredictable user experiences), and this sort of allows that now. However, this seems like a reasonable approach as it defaults to the behavior when focusing a tree explicitly which allows the tree to fallback to a more suitable default (such as the first item to select in the toolbox for that particular tree). In many cases this will default back to the tree's root node (such as the workspace root group) since sometimes the removed node is still the "last focused node" of the tree (and is considered valid for the purpose of determining a fallback; tree implementations could further specialize by checking whether that node is still valid).

### Test Coverage

Some new tests were added to cover this case, but more may be useful to add as part of #8910.

### Documentation

No documentation needs to be added or updated as part of this (beyond code documentation changes).

### Additional Information

This original issue was found by @RoboErikG when testing #8995. I also verified this against the keyboard navigation plugin repository.

### DIFF
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -227,7 +227,7 @@ export class FocusManager {
   }
 
   /**
-   * Focuses DOM input on the selected node, and marks it as actively focused.
+   * Focuses DOM input on the specified node, and marks it as actively focused.
    *
    * Any previously focused node will be updated to be passively highlighted (if
    * it's in a different focusable tree) or blurred (if it's in the same one).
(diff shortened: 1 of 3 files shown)


BenHenning (commented):
Self-reviewed changes (needed a couple of follow-up fixes).

BenHenning:
@rachel-fenichel also assigning to you in case you have time to review this before @gonfunko.

rachel-fenichel (approved):

PR #9004, opened 2025-05-06
feat!: Make everything ISelectable focusable

### LINKED ISSUE #8820: Update cursor state in `setSelected`
### Check for duplicates

### Problem

_No response_

### Request

Update `Blockly.common.setSelected` to set the cursor position if the selected item can have a cursor on it--and unset the cursor position if necessary.

Test by linking against keyboard-experimentation.

Eventually this will also interact with focus.

Related to [#8796](https://github.com/google/blockly/issues/8796)

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #8998: Clicking the workspace does not remove focus or selection from workspace comments
### Check for duplicates

### Description

Clicking a workspace comment selects it and allows it to be edited, but clicking away to the workspace does not remove focus or selection from the comment.

As a result, dragging the workspace after editing a workspace comment moves the comment instead of the workspace.

Clicking on a block correctly moves focus/selection.

### Reproduction steps

In the screencast below, I

- Dragged a block into the workspace.
- Dragged the workspace, which correctly translated (see scrollbars).
- Right-clicked to create a workspace comment
- Clicked away from the comment to the workspace (click is not rendered in the gif)
- Dragged the workspace
- Observed that the comment translated relative to the workspace background and block, rather than the entire workspace translating

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

### LINKED ISSUE #9000: Comments.

### LINKED ISSUE #9001: Icons.

### LINKED ISSUE #9002: Make mutator bubbles focusable and navigable
See parent issue.

### LINKED ISSUE #9003: Bubbles.

BenHenning:
## The basics

### Resolves

Fixes #8820
Fixes #9000
Fixes #9001
Fixes #9003
Fixes #8998
Fixes part of #8771

### Proposed Changes

At a high level, this PR makes everything that's selectable (that is, implements `ISelectable`) now focusable by making `ISelectable` extend `IFocusableNode`. The brings with it a bunch of new possibilities, also introduced in this PR:
- A change to how selection is managed in `common` to be completely derived from focus (since we now guarantee that selection auto-updates for `ISelectable`s when they are focused).
- An update to ensure that icons and bubbles are now focusable themselves. They are not yet navigable via the keyboard, but that should be a straightforward next step now that they've been made focusable.
  - Note that this means a specific change to make both `IIcon` and `IBubble` extensions of `IFocusableNode`. Both imply having visual elements which is why these root interfaces have been updated (unlike in other cases like `Block` and `Connection`).
- Replacements for all `common.setSelected()` calls to now use `getFocusManager().focusNode()`. Since selection is represented by focus, focusing with auto-selection is logically equivalent to the old `common.setSelected()` call (except for clearing a selection state).
- `LineCursor` no longer needs to fall back to selection since it's completely covered by focus cases.
- Identifying comments, bubbles, and icons in `WorkspaceSvg` has been done somewhat generically to provide a reasonable pathway for custom implementations to properly hook into workspace focus.

### Reason for Changes

Fundamentally, these changes are needed to simplify selection logic as well as add missing components to the focus system (which will then allow them to become navigable and fully functional for keyboard navigation).

Some key design decisions that were made:
- `ISelectable` and `common.setSelected()`/`common.getSelected()` have been kept even though they are trivially removable currently. This is because we still want to maintain selectable state and an API to represent it for when multi selection can be designed into core Blockly. `common.setSelected()` is still marked as `@internal`.
- A new `canBeFocused` method has been added to `IFocusableNode`.
  - This provides a means for components to participate in the focus system without being obligated to provide a focusable element. Since we can't easily prevent this from being a dynamic property, it should more or less work so long as the contract documented is followed.
  - This may be able to replace `INavigable.isNavigable()`.
- The `SELECTED` event firing logic has been pulled into a new `common` method. It's now called directly by `ISelectable` implementations and, being that it's `@internal`, shouldn't be used outside of core Blockly. There may well be a better location for this to live in the long-term, but that's a design decision better settled as part of multi-select.
- `BlockSvg` selection is now directly attached to shadow blocks rather than being forced to delegate to the block's parent. Instead, this is being checked at the places where it could matter (namely for gestures).
  - `BlockSvg`'s dispose logic no longer needs to care about selections since they're tied to focus, and focus should automatically adjust when a block is removed from the workspace.
- `Bubble` has a bit of tricky changes to its constructor as a result of needs from `TextInputBubble`. Specifically, for reasons I am not 100% sure of it seems that `TextInputBubble`'s `textarea` must be focused in order to receive input (this is not the case for `RenderedWorkspaceComment` despite very similar implementations). Thus, `Bubble` provides a way to customize the focusable element that it uses (but otherwise defaults to its root group).
  - Note that some of the `bringToFront()` logic was simplified since it can be tied to focus being received, now.
  - Some of the extra focus logic in `TextInputBubble` is no longer necessary. `FocusManager` provides fairly strong guarantees for restoring focus for certain DOM operations (though this can get complicated as is seen for `BlockSvg.bringToFront()`).
- `IHasBubble` has been updated to require being able to return the current bubble (which is used by `WorkspaceSvg` to find a focusable bubble in response to DOM changes).
- #8998 is fixed due to `RenderedWorkspaceComment` now being focusable (and thus unfocusable to return focus to the workspace itself for gesture management). Note that there was one change in `LayerManager` to avoid unnecessary appends (which can cause a cycle in `FocusManager` for focused elements when the operation is called in `onFocusNode`). Some effort could be made to break the cycle in `FocusManager` itself (which is tricky), but this approach has a net benefit to avoid an operation that simply isn't needed.
- `WorkspaceDragger` has had its selection/focus logic on drag start removed due to it being unnecessary. See: https://github.com/google/blockly/pull/9004#discussion_r2080304185.

Caveats:
- This PR doesn't change where `BlockSvg`'s `addSelect`/`removeSelect` are used directly. It's assumed these are still needed, but they may be incidentally obsolete if there's any part of the dataflow that forces focus (since that will then force selection state, anyway).
- This will cause a failure in the keyboard navigation plugin until https://github.com/google/blockly-keyboard-experimentation/pull/502 is submitted (due to the breaking API change being introduced here).
- Focus events are a bit spammy now as they'll send a deselect between selection events (i.e. selecting something new will involve a deselect event being fired first). This is the existing behavior for blocks (but is relatively new since that behavior was introduced when `BlockSvg` was updated to auto-select on focus) and is new for other selectables.
- The changes to selection here and in #8938 mean that `common.getSelected()` will return null for cases when there was previously a selection but not active focus (such as right clicking on a block).
- `RenderedConnection` technically depends on the shadow block selection logic that was removed, however it's actually quite challenging to implement that there. `RenderedConnection` cannot import `BlockSvg` directly, and fixing this import ordering could potentially be quite complex. That being said, it seems that the selection management there is actually no longer needed due to focus automatically driving selection states. See https://github.com/google/blockly/pull/9004#discussion_r2078738690 for much more details on why.
- One test has been removed from the block tests since it was directly verifying unhighlighting a shadow's parent when disposing the shadow child and there's no strong analogue to change this (as the behavior no longer exists).
- This doesn't solve #9002 despite the mutator icon now being focusable and it containing a workspace that supports focus. Some nested focus logic needs to be implemented in `WorkspaceSvg` in order to ensure cross-workspace hand-off works correctly, though this might partly come down to a usability and navigation question that should be solved outside this PR.

### Test Coverage

Tests have not been added here, but there are legitimate cases for future testing:
- The `FocusManager` behaviors should be tested as part of #8910.
- The various behavior changes for selection, bubbles, icons, and workspace comments should be tested more thoroughly as part of #8915.

Manually testing has been detailed in this comment: https://github.com/google/blockly/pull/9004#issuecomment-2864468390.

### Documentation

As part of the broader focus management documentation, we will likely want to describe how to account for focus support when implementing custom icons and bubbles (though the base classes are meant to mostly solve these as-is) when wanting to maintain keyboard navigation support.

### Additional Information

None currently.


core/common.ts
```diff
@@ -86,38 +91,58 @@ export function setMainWorkspace(workspace: Workspace) {
...
+export function getSelected(): ISelectable | null {
+  const focused = getFocusManager().getFocusedNode();
+  if (focused && isSelectable(focused)) return focused;
+  return null;
+}
 
 /**
- * Returns the currently selected copyable object.
+ * @internal
  */
```
rachel-fenichel, line 104: Add a comment: returns the selected block, or the first non-shadow parent if the selected block was a shadow.

rachel-fenichel:
This looks directionally correct to me.

core/common.ts
BenHenning: Thanks for the reminder--I definitely will add a comment. Just added the '@internal' to make it clear for draft review.

core/block_svg.ts
```diff
@@ -892,8 +867,7 @@ export class BlockSvg
     this.disposing = true;
     super.disposeInternal();
 
-    if (common.getSelected() === this) {
-      this.unselect();
+    if (getFocusManager().getFocusedNode() === this) {
```
maribethb, line 870: It looks like you’re missing the equivalent “defocus” call, is that intentional?

core/common.ts
```diff
@@ -7,7 +7,12 @@
 // Former goog.module ID: Blockly.common
 
 import type {Block} from './block.js';
-import {ISelectable} from './blockly.js';
+import {
+  BlockSvg,
+  ISelectable,
+  getFocusManager,
+  isSelectable,
+} from './blockly.js';
```
maribethb, line 15: These probably shouldn’t be imported from blockly.js but from their real files. Vs code will do the wrong thing sometimes unfortunately.

core/common.ts
```diff
@@ -86,38 +91,58 @@ export function setMainWorkspace(workspace: Workspace) {
 }
 
 /**
- * Currently selected copyable object.
+ * Returns the current selection.
  */
-let selected: ISelectable | null = null;
+export function getSelected(): ISelectable | null {
+  const focused = getFocusManager().getFocusedNode();
```
maribethb, line 97: I am curious about how well this will work. 

One scenario is that in context menu items sometimes the callback would use getSelected, because the block is still selected when the menu is open. But the block won’t be focused when the menu is open. So we would need to call this out as a breaking change. I am curious if there are other situations where a block would have been selected but not (active) focused and I’m not sure I can think of any, maybe in some widget div scenarios

core/icons/mutator_icon.ts
```diff
@@ -7,6 +7,7 @@
 // Former goog.module ID: Blockly.Mutator
 
 import type {BlockSvg} from '../block_svg.js';
+import {IBubble} from '../blockly.js';
```
maribethb, line 10: This can be import type, right? Here and elsewhere. Also, make sure you import it from the interface file and not from blockly.js

core/workspace_svg.ts
```diff
@@ -2766,7 +2768,33 @@ export class WorkspaceSvg
       return null;
     }
 
-    return this.getBlockById(id) as IFocusableNode;
+    // Search for a specific block.
+    const block = this.getBlockById(id) as IFocusableNode;
```
maribethb, line 2772: Do you still need this cast?

core/block_svg.ts
```diff
@@ -265,20 +265,14 @@ export class BlockSvg
 
   /** Selects this block. Highlights the block visually. */
   select() {
-    if (this.isShadow()) {
```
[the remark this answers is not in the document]
maribethb: I found it later in the review

core/bubbles/textinput_bubble.ts
```diff
@@ -131,11 +130,21 @@ export class TextInputBubble extends Bubble {
     this.locationChangeListeners.push(listener);
   }
 
+  private static createTextArea(): HTMLTextAreaElement {
+    const textArea = document.createElementNS(
+      dom.HTML_NS,
+      'textarea',
+    ) as HTMLTextAreaElement;
+    textArea.className = 'blocklyTextarea blocklyText';
+    // textArea.setAttribute('tabindex', '-1');
```
BenHenning, line 139: Remove.

core/bubbles/textinput_bubble.ts
```diff
@@ -149,22 +158,16 @@ export class TextInputBubble extends Bubble {
...
     body.appendChild(textArea);
     inputRoot.appendChild(body);
 
     this.bindTextAreaEvents(textArea);
-    setTimeout(() => {
-      textArea.focus();
-    }, 0);
+    // setTimeout(() => {
+    //   textArea.focus();
+    // }, 0);
```
BenHenning, line 168: Remove.

core/clipboard/workspace_comment_paster.ts
```diff
@@ -49,7 +49,7 @@ export class WorkspaceCommentPaster
     if (eventUtils.isEnabled()) {
       eventUtils.fire(new (eventUtils.get(EventType.COMMENT_CREATE))(comment));
     }
-    common.setSelected(comment);
+    getFocusManager().focusNode(comment);
```
BenHenning, line 52: Double check this works.

core/common.ts
BenHenning: I think the scenario you describe would be broken as of #8938 not this change, but it's equally relevant. It's definitely the case now that right clicking will not maintain block selection since the selection highlight is cleared on focus loss.

It's something worth discussing, I think. Does something really have selection if it doesn't have input focus? My assumption is no which is why the change moved in this direction. If the answer should be yes, then we need to approach the auto-selection differently.

core/contextmenu.ts
```diff
@@ -289,7 +289,7 @@ export function callbackFactory(
     if (eventUtils.isEnabled() && !newBlock.isShadow()) {
       eventUtils.fire(new (eventUtils.get(EventType.BLOCK_CREATE))(newBlock));
     }
-    common.setSelected(newBlock);
+    getFocusManager().focusNode(newBlock);
```
BenHenning, line 292: Verify that this works.

core/contextmenu_items.ts
```diff
@@ -631,7 +631,7 @@ export function registerCommentCreate() {
           workspace,
         ),
       );
-      common.setSelected(comment);
+      getFocusManager().focusNode(comment);
```
BenHenning, line 634: Verify that this works.

core/gesture.ts
```diff
@@ -346,7 +346,7 @@ export class Gesture {
     // to return the new block.
     if (this.flyout) this.updateIsDraggingFromFlyout();
 
-    const selected = common.getSelected();
+    const selected = this.getSelectedBlock();
```
BenHenning, line 349: Verify that the specific case where this matters with shadow blocks still works.

core/icons/comment_icon.ts
```diff
@@ -338,6 +339,10 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
     );
   }
 
+  getBubble(): IBubble | null {
```
BenHenning, line 342: Add docs.

core/bubbles/textinput_bubble.ts
```diff
@@ -131,11 +130,21 @@ export class TextInputBubble extends Bubble {
     this.locationChangeListeners.push(listener);
   }
 
+  private static createTextArea(): HTMLTextAreaElement {
```
BenHenning, line 133: Add docs.

core/icons/icon.ts
```diff
@@ -144,4 +155,22 @@ export abstract class Icon implements IIcon {
...
+  /** See IFocusableNode.getFocusableElement. */
+  getFocusableElement(): HTMLElement | SVGElement {
+    const svgRoot = this.svgRoot;
+    if (!svgRoot) throw new Error('Attempting to focus uninitialized icon.');
+    return svgRoot;
+  }
+
+  /** See IFocusableNode.getFocusableTree. */
+  getFocusableTree(): IFocusableTree {
+    return this.sourceBlock.workspace as WorkspaceSvg;
```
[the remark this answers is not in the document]
BenHenning: Is there an easy way to check whether an icon is headless? Now that we have `canBeFocused()` we can actually do this somewhat properly (even if there's a bit of an inversion here, though that's already true given that this class manages an SVG root).

That all being said, I get that `IIcon`s can be headless, but can `Icon`s? It seems from the design that this abstract class is meant for non-headless cases (the new invisible icons in the keyboard navigation plugin notwithstanding).

core/icons/mutator_icon.ts
```diff
@@ -203,6 +204,10 @@ export class MutatorIcon extends Icon implements IHasBubble {
     );
   }
 
+  getBubble(): IBubble | null {
```
BenHenning, line 207: Add docs.

core/icons/warning_icon.ts
```diff
@@ -197,6 +198,10 @@ export class WarningIcon extends Icon implements IHasBubble {
     );
   }
 
+  getBubble(): IBubble | null {
```
BenHenning, line 201: Add docs.

core/interfaces/i_focusable_node.ts
```diff
@@ -59,6 +59,21 @@ export interface IFocusableNode {
...
+   * Note that special care must be taken if this dynamically changes its return
+   * value over the lifetime of the node as certain environment conditions could
+   * affect the focusability of this node's DOM element (such as whether the
+   * element has a positive or zero tabindex). Also, changing from a true to a
+   * false value while the node holds focus would not change the current focus
+   * of the node, and may result in some of the node's functions being called
+   * later on when defocused (since it was previously considered focusable at
+   * the time of being focused).
+   */
+  canBeFocused(): boolean;
```
BenHenning, line 76: Update docs to include a returns, and reconsider the caveats in `INavigable.isNavigable` to see if some of those should be moved here.

core/interfaces/i_has_bubble.ts
```diff
@@ -4,12 +4,16 @@
...
+import {IBubble} from './i_bubble';
+
 export interface IHasBubble {
   /** @returns True if the bubble is currently open, false otherwise. */
   bubbleIsVisible(): boolean;
 
   /** Sets whether the bubble is open or not. */
   setBubbleVisible(visible: boolean): Promise<void>;
+
+  getBubble(): IBubble | null;
```
BenHenning, line 16: Add docs.

core/interfaces/i_has_bubble.ts
```diff
@@ -4,12 +4,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {IBubble} from './i_bubble';
```
BenHenning, line 7: `import type`, instead.

core/interfaces/i_selectable.ts
```diff
@@ -7,11 +7,17 @@
 // Former goog.module ID: Blockly.ISelectable
 
 import type {Workspace} from '../workspace.js';
+import {IFocusableNode, isFocusableNode} from './i_focusable_node.js';
 
 /**
  * The interface for an object that is selectable.
+ *
+ * Implementations are genreally expected to use their implementations of
```
BenHenning, line 15: 'generally'

core/rendered_connection.ts
```diff
@@ -559,21 +555,6 @@ export class RenderedConnection
...
-    // path could be partially obscured by a new block inserted after it in the
-    // DOM.
-    const selection = common.getSelected();
-    const selectedBlock =
-      (selection === parentBlock && parentBlock) ||
-      (selection === childBlock && childBlock);
-    if (selectedBlock) {
-      selectedBlock.removeSelect();
-      selectedBlock.addSelect();
-    }
```
BenHenning, line 575: Check whatever cases this can occur to see if it still works as expected.

core/rendered_connection.ts
```diff
@@ -198,15 +197,12 @@ export class RenderedConnection
...
     // Raise it to the top for extra visibility.
-    const selected = common.getSelected() === dynamicRootBlock;
-    if (!selected) dynamicRootBlock.addSelect();
     if (dynamicRootBlock.RTL) {
       offsetX = -offsetX;
     }
     const dx = staticConnection.x + offsetX - dynamicConnection.x;
     const dy = staticConnection.y + offsetY - dynamicConnection.y;
     dynamicRootBlock.moveBy(dx, dy, ['bump']);
-    if (!selected) dynamicRootBlock.removeSelect();
```
BenHenning, line 209: Verify that these changes still work correctly.

core/workspace_dragger.ts
```diff
@@ -57,9 +57,8 @@ export class WorkspaceDragger {
    * @internal
    */
   startDrag() {
-    if (common.getSelected()) {
-      common.getSelected()!.unselect();
-    }
+    // Focus the workspace root to indicate that it's now selected for dragging.
+    getFocusManager().focusNode(this.workspace);
```
BenHenning, line 61: Test workspace dragging.

core/workspace_svg.ts
BenHenning: I do not! Good catch--removed in the latest changes.

BenHenning (commented):
Self-reviewed all changes so far.

core/block_svg.ts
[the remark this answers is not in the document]
BenHenning: It was actually missing for a bit, but it's also now documented in the PR description hopefully to make things a bit more explicit and not require discovery.

core/common.ts
BenHenning: Yep, this was definitely VS Code in a bunch of places. I've fixed this here, and updated elsewhere to use more `import type` since it largely seems better for reducing circular dependency risks.

core/icons/mutator_icon.ts
BenHenning: Yep! Haven't fixed this yet, but also made notes to do so in my self-review. Will fix in a follow-up change.

core/common.ts
```diff
@@ -86,38 +91,58 @@ export function setMainWorkspace(workspace: Workspace) {
...
+  return null;
+}
 
 /**
- * Returns the currently selected copyable object.
+ * @internal
  */
-export function getSelected(): ISelectable | null {
-  return selected;
+export function getSelectedBlock(): BlockSvg | null {
```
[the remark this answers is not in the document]
BenHenning: I don't like it either, and neither did the node tests which fell over due to the circular dependency bringing `BlockSvg` into `common` introduced. :)

We also, unfortunately, can't specialize `common.getSelected()` in this way for the same reason. I've moved it to what I _think_ is the only place now where it's needed (gestures), but there's some testing in the only other place that _could_ need it (`RenderedConnection`) to verify. I really hope that this is the case since otherwise this PR is about to get a lot more complicated.

BenHenning:
Despite the large number of self-comments, I think this actually could go through a real review now. I've tested a bunch of the obvious happy paths around interacting with icons, workspace comments, and bubbles (mutators/comments/warnings) to make sure everything seems sane, and I've checked the keyboard navigation plugin briefly to make sure nothing obvious is broken by this PR (though there's still an `INavigable` issue there so it's not a perfect test).

There are, unfortunately, a lot of other things that still need to be tested here and I've made notes for them. I'll need a second review pass once I address the remaining and any new comments, but I'm hoping the code delta between now and then won't be large (barring any real issues found from the remaining testing yet to be conducted).

core/icons/icon.ts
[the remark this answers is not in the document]
maribethb: If there’s an icon in a block definition, and then you load that kind of block into a headless workspace, the block still has an icon. That’s what I mean by headless icon (not meaning an invisible icon in a rendered block)

core/bubbles/bubble.ts
```diff
@@ -97,6 +101,7 @@ export abstract class Bubble implements IBubble, ISelectable {
     public readonly workspace: WorkspaceSvg,
     protected anchor: Coordinate,
     protected ownerRect?: Rect,
+    private overriddenFocusableElement?: SVGElement | HTMLElement,
```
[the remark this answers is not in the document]
BenHenning: Done in latest.

core/bubbles/textinput_bubble.ts
BenHenning: Done in latest.

core/contextmenu_items.ts
BenHenning: This seems to be working correctly. How I verified: I added a workspace comment and verified that its border was highlighted (indicating that it was selected). Removing this line also avoids the auto-selection upon creation. Focusing a workspace comment does NOT mean it can be edited for text input.

All of this seems to match existing behavior.

core/gesture.ts
BenHenning: First of all, this was wrong and needed to be fixed. It was always assuming a block here but there could also be a workspace comment, so it's been adjusted.

For blocks, this...is interesting. As far as I can tell, we don't actually need the specialization for shadow blocks and, in fact, it may be less desirable.

Here's the current state:

Screen recording 2025-05-08 11 35 48 AM

Here's the new behavior with shadow blocks being considered (notice the highlight on the shadow when starting to drag--this is new)--this is the current implementation in this PR:

Screen recording 2025-05-08 11 33 51 AM

Here's the new behavior without shadow blocks being considered:

Screen recording 2025-05-08 11 32 49 AM

The last seems the most ideal since it maintains the shadow block selection which is in-line with what we want now. However, it is a bit messy in that the shadow can't easily be selected _normally_ except through dragging.

@rachel-fenichel do you have any thoughts or opinions on this?

core/icons/comment_icon.ts
BenHenning: Done in latest.

core/icons/mutator_icon.ts
BenHenning: Done in latest.

core/icons/warning_icon.ts
BenHenning: Done in latest.

core/bubbles/textinput_bubble.ts
BenHenning: Done & revised `createEditor`'s docs in latest.

core/gesture.ts
rachel-fenichel: What happens if you start a drag from a shadow block and _then_ copy-paste? As long as that doesn't crash I think I'm good with it. I expect a crash or other fail if you try to paste a shadow block at the top level, which is illegal.

There is obviously an inconsistency in that copy-paste will act on the parent block, and https://github.com/google/blockly-keyboard-experimentation/issues/208 already asked for that to be changed.

I am okay with the highlight being tighter (on the shadow block instead of on the parent), because it will only be confusing in the copy/paste case and we will fix that in a follow-up.

However, I don't understand the distinction between "shadow blocks being considered" and "shadow blocks not being considered".

core/interfaces/i_focusable_node.ts
BenHenning: Copied over the main blob from `INavigable` and adjusted it to better tailor toward focus, and provide accessibility-specific advice. I did drop the 'Field'-specific example since it would be contradicted by my additional consideration that even disabled or read-only fields may still benefit from being focusable. I also mentioned invisible elements since that's actually the only case today where we actually disable focusability (for the keyboard navigation plugin a la https://github.com/google/blockly-keyboard-experimentation/pull/502.

core/interfaces/i_has_bubble.ts
BenHenning: Done in latest.

core/interfaces/i_has_bubble.ts
BenHenning: Done in latest.

core/interfaces/i_selectable.ts
BenHenning: Done in latest.

core/workspace_dragger.ts
BenHenning: Working dragging with selection still works as expected, but it turns out this line isn't actually needed at all. I think this is due to some redundancy in gesture handling, specifically this line:

https://github.com/google/blockly/blob/561129ac612662e8b77a12556265eae348aac054/core/gesture.ts#L770

I verified this by adding a block, selecting it, then dragging outside the block to verify that the workspace position both moves and the block is de-selected. I verified that this line is no longer necessary by removing it and re-checking the behavior, and then commenting out the above the line and verifying that line is actually the one that ensures deselection as expected.

Interestingly, the line above is necessary. Removing that but keeping the line in `WorkspaceDragger` regresses the behavior (i.e. by keeping the block highlighted even when finishing a workspace drag). I've opted to remove the line from `WorkspaceDragger` entirely since it's unnecessary.

I also verified this against `develop` to ensure that the behavior is the same as before.

core/rendered_connection.ts
BenHenning: See https://github.com/google/blockly/pull/9004/files/b7c6aca41e1ccb85b65bb3c7eb211e484c2ecc70#r2078738690.

core/rendered_connection.ts
BenHenning: Note that this testing will include verifying https://github.com/google/blockly/pull/9004#discussion_r2078738887 as well to keep things in one location for `RenderedConnection`.

BenHenning: I've dug back to https://github.com/google/blockly/commit/02b2c219ebab921b3c725f1e8ac20ebc80e29202 as the original introducing of the selection correction logic. I'm not sure what the original bug's repro steps were, but this fix is very old and predates gestures plus bring-to-front logic (at least in its current form). I'm now suspecting that the selection bits for bumping have been made obsolete with other bring-to-front mechanisms such that this hasn't been needed for some time. No amount of bumping seems to get in the way of the selected block or the selection itself (both always seem to stay above the bumped block), so I do think it's safe to remove this logic now.

BenHenning: Going back to `connect_`, I think I know what's happening here and verifying would be tricky, but I think this fine to logic out based on the observation that everything is working fine without the selection bits in.

With the latest changes, the data flow roughly looks like this:
- User sends pointer event to click on a block.
- Gestures mark that block as focused.
- The block, upon receiving focus, auto select themselves (and if a previous block was focused, it removes its selection).
- The process of removing/adding a selection bumps it back to the top in the same way as the old logic here did.

Thus, this shouldn't be needed since the various efforts to focus the block being moved should generally always result with its selection appearing on top.

BenHenning (commented):
Self-reviewed latest changes and working to close open conversation threads that have had their trivial fixes addressed.

core/common.ts
BenHenning: This has been addressed--marking as resolved to keep open only those threads that may need additional discussion.

core/block_svg.ts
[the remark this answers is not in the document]
BenHenning: This has been addressed--marking as resolved to keep open only those threads that may need additional discussion.

core/common.ts
BenHenning: This has been addressed--marking as resolved to keep open only those threads that may need additional discussion.

core/icons/mutator_icon.ts
BenHenning: This has been addressed--marking as resolved to keep open only those threads that may need additional discussion.

core/workspace_svg.ts
BenHenning: This has been addressed--marking as resolved to keep open only those threads that may need additional discussion.

core/common.ts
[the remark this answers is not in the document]
BenHenning: Can confirm that the selection logic for `RenderedConnection` seems like it can correctly be removed in full per https://github.com/google/blockly/pull/9004#discussion_r2078738690.

core/common.ts
maribethb: I think logically it's fine, but it is a big behavior change that the block isn't selected anymore while you're interacting with non-Blockly parts of the UI (e.g. context menu or dev console) so we just need to be super clear about that in our communication.

core/icons/icon.ts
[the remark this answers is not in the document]
maribethb: But to be clear I don't have a solution to this and I don't think it's the end of the world since I can't imagine how this method could be called if we weren't dealing with a rendered workspace. You could probably check if the source block is a `BlockSvg` which I think would then know that its `workspace` property is a `WorkspaceSvg`. But then you'd have to throw an error if it's not a `BlockSvg` since you otherwise have to return *something*. Idk, I don't necessarily have a problem with the cast, just wanted to make the original comment in case we did have a `RenderedIcon` class I was forgetting about, but we don't

core/common.ts
BenHenning: What's the best way for me to indicate that bit (per being super clear)?

core/icons/icon.ts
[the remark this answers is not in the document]
BenHenning: Got it. In that case, it seems that something like this is necessary. Do you have a preference between the `WorkspaceSvg` cast vs. a `BlockSvg` cast as you suggested? Happy to do either.

core/focus_manager.ts
```diff
@@ -232,11 +232,16 @@ export class FocusManager {
...
+   * **Important**: If the provided node is not able to be focused (e.g. its
+   * canBeFocused() method returns false), it will be ignored and any existing
+   * focus state will remain unchanged.
+   *
    * @param focusableNode The node that should receive active focus.
    */
   focusNode(focusableNode: IFocusableNode): void {
     this.ensureManagerIsUnlocked();
     if (this.focusedNode === focusableNode) return; // State is unchanged.
+    if (!focusableNode.canBeFocused()) return; // This node can't be focused.
```
maribethb, line 244: Would it make sense to log a warning in this case? This isn't something you would intentionally ever do right?

core/gesture.ts
```diff
@@ -357,6 +357,16 @@ export class Gesture {
     }
   }
 
+  private getSelectedBlock(): BlockSvg | null {
```
maribethb: Can you add some docs explaining the purpose of this please

core/gesture.ts
BenHenning: For the distinction, see this change: https://github.com/google/blockly/pull/9004/commits/0a6e716795c659a5516504bb49eda91498bb14c7. This is the the last gif above and is now pushed to latest (without the change it's the middle gif).

For copying and pasting, it doesn't crash. The behavior is basically:
- If you copy and paste with nothing copied with the shadow block now selected, nothing happens.
- If you copy and paste another block then try again with the shadow block, it will paste the previously copied block (since it seems the shadow block fails to copy, not to paste).

core/interfaces/i_focusable_node.ts
maribethb: I think I would probably remove "disabled, read-only" from the list of examples for why you might want this to return false, as those cases should still be focusable for screenreader users most likely. I think it really is just the purely visual decoration / no visual representation case.

core/layer_manager.ts
```diff
@@ -122,7 +122,12 @@ export class LayerManager {
     if (!this.layers.has(layerNum)) {
       this.createLayer(layerNum);
     }
-    this.layers.get(layerNum)?.appendChild(elem.getSvgRoot());
+    const childElem = elem.getSvgRoot();
+    if (this.layers.get(layerNum)?.lastChild !== childElem) {
```
maribethb, line 126: This seems weird, since `append` doesn't remove the element from elsewhere, I can't imagine when this case would occur. Wouldn't it cause a problem if you tried to `append` a child to a layer where that child already existed?

If it is somehow intentional before that you could have the same element appended multiple times, it feels like you might be breaking things by disallowing that.

I don't have a good understanding of this code so I could be wrong, but this feels suspicious.

maribethb (approved):

core/bubbles/bubble.ts
```diff
@@ -211,8 +224,7 @@ export abstract class Bubble implements IBubble, ISelectable {
   /** Brings the bubble to the front and passes the pointer event off to the gesture system. */
```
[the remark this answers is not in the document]
BenHenning: Done in latest.

core/focus_manager.ts
BenHenning: Good idea--added.

core/gesture.ts
BenHenning: This has been removed in the latest changes.

core/layer_manager.ts
BenHenning: See the first note in https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild which is actually what was happening. We would append the child which would trigger another focus, that would trigger an `onNodeFocus` callback for `RenderedWorkspaceComment`, which would append the child, and thus repeat ad infinitum.

core/interfaces/i_focusable_node.ts
BenHenning: I agree, though I was trying to maintain some parity with the existing documentation.

@gonfunko and @rachel-fenichel do you have any thoughts as this documentation would be giving explicit guidance to implementations?

core/layer_manager.ts
maribethb: Oh got it, thanks!

core/common.ts
maribethb: Update the PR description of this PR (and/or any others you've made that are breaking) to include a `## Breaking Changes` section that explains 1) who is broken by the change 2) what they need to do in response to the breaking change.

So for the issue about the block not being selected anymore, provide workarounds like using the (now passively-) focused block instead of selected block, using the scope for context menu items and shortcuts instead of selected, etc.

For this PR, you might need other details like what new methods need to be implemented if any for bubbles, etc.

Then remind me to call this out in the handwritten release notes we'll attach to the github release/forum announcement as well.

core/icons/icon.ts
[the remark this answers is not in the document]
maribethb: I'm fine with it as is

PR #9011, opened 2025-05-07
fix: Fix regressions in `Field`.

### LINKED ISSUE #9005: `Field` regressions due to #8738
### Check for duplicates

### Description

PR #8738 fixed issue #8738, but introduced a number of unexpected changes to fields:

1.  Remove of `Field.NBSP` broke the multiline text input field in a trivial but somewhat gratuitous way.
0.  Removal of [code that ensures field display text is at least one character long](https://github.com/gonfunko/blockly/blob/c68d6451b736827eb0f0efa425427a488cd576a8/core/field.ts#L982) caused empty inputs to become quite narrow:
    * Before:
      Image
    * After:
      Image
0.  Removal of [code that replaces normal spaces with non-breaking spaces](https://github.com/gonfunko/blockly/blob/c68d6451b736827eb0f0efa425427a488cd576a8/core/field.ts#L990) causes fields with multiple consecutive spaces to be mis-rendered:
    * Before:
       Image
    * After:
      Image

### Reproduction steps

1. Open playground.
2. Drag any block with a `FieldInput` to the workspace.
3. Enter the empty string.
4. Enter text with multiple spaces.

### Additional observations

* The change to make empty label fields zero-width appears to be correct and should be retained.
* It's pretty clear that `FieldInput` at should have the previous multiple-space-retention behaviour.  It's less clear whether ordinary `Field`s should too, but it looks like there will be a rendering bug if they don't.
* There doesn't seem to be any reason that empty `FieldInput`s should be revert to the previous, slightly wider width:
  * It's a larger target, thus easier for pointer users to hit.
  * Making them narrower does not fulfil any part of the request in #8738.

gonfunko:
## The basics

### Resolves

Fixes #9005

### Proposed Changes
This PR fixes several regressions in `Field` introduced by an earlier change:
* It restores the `Field.NBSP` constant, which was in fact in use
* It makes empty input fields have a reasonable minimum width
* It prevents sequential spaces from collapsing when rendering a field

### DIFF
--- a/core/field.ts
+++ b/core/field.ts
@@ -88,6 +88,9 @@ export abstract class Field<T = any>
    */
   DEFAULT_VALUE: T | null = null;
 
+  /** Non-breaking space. */
+  static readonly NBSP = '\u00A0';
+
   /**
    * A value used to signal when a field's constructor should *not* set the
    * field's value or run configure_, and should allow a subclass to do that
@@ -110,7 +113,28 @@ export abstract class Field<T = any>
    * field is not yet initialized. Is *not* guaranteed to be accurate.
    */
   private tooltip: Tooltip.TipInfo | null = null;
-  protected size_: Size;
+
+  /** This field's dimensions. */
+  private size: Size = new Size(0, 0);
+
+  /**
+   * Gets the size of this field. Because getSize() and updateSize() have side
+   * effects, this acts as a shim for subclasses which wish to adjust field
+   * bounds when setting/getting the size without triggering unwanted rendering
+   * or other side effects. Note that subclasses must override *both* get and
+   * set if either is overridden; the implementation may just call directly
+   * through to super, but it must exist per the JS spec.
+   */
+  protected get size_(): Size {
+    return this.size;
+  }
+
+  /**
+   * Sets the size of this field.
+   */
+  protected set size_(newValue: Size) {
+    this.size = newValue;
+  }
 
   /** The rendered field's SVG group element. */
   protected fieldGroup_: SVGGElement | null = null;
@@ -973,6 +997,8 @@ export abstract class Field<T = any>
       // Truncate displayed string and add an ellipsis ('...').
       text = text.substring(0, this.maxDisplayLength - 2) + '…';
     }
+    // Replace whitespace with non-breaking spaces so the text doesn't collapse.
+    text = text.replace(/\s/g, Field.NBSP);
     if (this.sourceBlock_ && this.sourceBlock_.RTL) {
       // The SVG is LTR, force text to be RTL by adding an RLM.
       text += '\u200F';
--- a/core/field_image.ts
+++ b/core/field_image.ts
@@ -27,7 +27,6 @@ export class FieldImage extends Field<string> {
    * of the field.
    */
   private static readonly Y_PADDING = 1;
-  protected override size_: Size;
   protected readonly imageHeight: number;
 
   /** The function to be called when this field is clicked. */
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -101,6 +101,26 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
   override SERIALIZABLE = true;
 
   /**
+   * Sets the size of this field. Although this appears to be a no-op, it must
+   * exist since the getter is overridden below.
+   */
+  protected override set size_(newValue: Size) {
+    super.size_ = newValue;
+  }
+
+  /**
+   * Returns the size of this field, with a minimum width of 14.
+   */
+  protected override get size_() {
+    const s = super.size_;
+    if (s.width < 14) {
+      s.width = 14;
+    }
+
+    return s;
+  }
+
+  /**
    * @param value The initial value of the field. Should cast to a string.
    *     Defaults to an empty string if null or undefined. Also accepts
    *     Field.SKIP_SETUP if you wish to skip setup (only used by subclasses
--- a/core/field_variable.ts
+++ b/core/field_variable.ts
@@ -49,7 +49,6 @@ export class FieldVariable extends FieldDropdown {
    * dropdown.
    */
   variableTypes: string[] | null = [];
-  protected override size_: Size;
 
   /** The variable model associated with this field. */
   private variable: IVariableModel<IVariableState> | null = null;


core/field.ts
```diff
@@ -110,7 +113,28 @@ export abstract class Field<T = any>
...
+  protected get size_(): Size {
+    return this.size;
+  }
+
+  /**
+   * Sets the size of this field.
+   */
+  protected set size_(newValue: Size) {
+    this.size = newValue;
+  }
```
cpcallen, line 137: Styleguide says "At least one accessor for a property must be non-trivial: do not define "pass-through" accessors only for the purpose of hiding a property."

I gather you are doing this in order to be able to write `super.size_ = newValue` in `FieldInput`'s `get size_` accessor,  but I think a better approach might be to have `updateSize_` enforce a minimum width (possibly in `Field` but preferably in `FieldInput`).  (Alas, `updateSize_` isn't written in a way that makes it easy to override for this particular purpose.)

core/field_image.ts
```diff
@@ -27,7 +27,6 @@ export class FieldImage extends Field<string> {
    * of the field.
    */
   private static readonly Y_PADDING = 1;
-  protected override size_: Size;
```
cpcallen, line 30: What was this even doing??

cpcallen:
(Restoration of `.NBSP` and space-to-NBSP anti-collapse code all LGTM, though.)

core/field.ts
gonfunko: As you noted, `updateSize_()` is not well-suited to being overridden, and short of introducing a `setSizeWithoutSideEffects()` method I couldn't see a way to refactor it to be more amendable. Likewise, overriding `getSize()` (or calling it from `updateSize_()` instead of accessing the `size_` property directly) is problematic due to its side effects of rerendering the field. Hence this seemed like the least-bad option, as it preserves backwards compatibility and allows for subclasses to adjust the size as they need. If you have an idea for how `updateSize_()` might be refactored I'd certainly be open to that though!

core/field_image.ts
gonfunko: Nothing AFAICT; I suspect it was introduced by the automated tooling during the TS migration.

core/field.ts
cpcallen: I had to think about this for a bit, but I think I see a way to do this without needing to have accessors on `Field`.

The obvious approach is to have the `get` and `set` accessors on the subclasses that want to enforce a minimum width only, and leave `Field` alone.  As you doubtless noticed, the reason this doesn't work is that the `Field` constructor sets `.size_`, and that results in (e.g.) `FieldInput` instances having a `.size_` property that is found before the `get size_`/`set size_` accessors on `FieldInput.prototype`.

But [if the `Field` constructor either didn't set that property if there were accessors](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDKCWAvAptA3gKGtYB7AdhAC4BOArsEbidABQAOZARiPMNAB4Bc0+ZAtk2QkANNEYs20AJ48+g4QEp0AXwDcGFRgyhIMAApgSyfEXRZoEJMh4IUG7HkKkKVErWWZs2APQ-oACIEAORm8PjwRPBgrBCo8ABmlsy6UNAAJsgJ4cgwAObIRD5xZmDAwLkQ1BAAdBbYRAAW8LVWKNAA-B0AvLzIAO5w1rQADCIjig7QWlo64GkAws0g6djIHEQm6QZGJmZe4iTwAG5gm5bWAJKmwvgx0L34A0Moo+OTFn7QAAJEEAC08Dy+GoNmgwTayGC0BaGSyOVWkGgYEOuHowiI0hh+Bw8xgwUMxlMwTETDIYRguGOwiO6UyOMawniOOCS3gK2hSLAOLKFSg1Dq2AKZkhHnQxiIZBIOKaLRqkOum2lMTUWmwJQur0hnllrSuN2VIAemuQqs0QA), like so:
```JS
class FieldInput {
  constructor() {
    if (!this.size_) {
      this.size_ = new Size(0, 0);
    }
  }
}
```
or [if the `FieldInput` constructor removed it](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDKCWAvAptA3gKGtYB7AdhAC4BOArsEbidABQAOZARiPMNAB4Bc0+ZAtk2QkANNEYs20AJ48+g4QEp0AXwDcGFRgyhIMAApgSyfEXRZoEJKgC8vZAHc412gAYRrxRq07wUaADCABbwIAAm2MgcRCZhBkYmZpjY9CTwAG5gMZbWAJKmwvhgINB2+I7OKG4eXhZ4hKQUVCS0ysnYlmT0wq0aHdAA9APQAAJEEAC08ADm+NTIPAAqQai43SRg+BG4AGbQYNAA5GHIIMgxh9BrwlnU0PxkxNBCV-RE8ATFAHQW2Cdn2SIIQgXysKD60C02CGo3GU1m8x4hzByEu8BgJx28HKEUg+3EJGuJCI0mg2Jwfhgh0MxlMhzETDIZnRV3SwjSYRO+GgK2MZO5h2CoTClzxm32wGAyCg1B+2Gm5xyVTaxiIZBI3KB6NBeQKGuKaihlkVKNoKLaWpBKPyMX1JTsKMNmiAA), like so:
```JS
class FieldInput {
  constructor() {
    super();
    delete this.size_;
  }
}
```
then I think it's possible to have the behaviour you have here without the mostly-useless accessors on `Field.prototype`.  Both approaches are a bit kludgy, and the `delete` approach _might_ have some performance consequences (it's generally preferable not to change the shape of an object, though here at least it's being done at construction time).  I'm genuinely not sure either is clearly preferable to your proposal, but I offer them for your consideration.

cpcallen: Additionally, I think that even if you keep the `get`/`set` accessors on `Field` it might be preferable to not call them from the corresponding accessors on `FieldInput`, either by making `size` protected (so that that `FieldInput`'s accessors can use it directly) or by having `FieldInput` keep it's own separate `Size` property.  A pair of accessors _ought_ to be interchangeable with a data property, in my view, but you have revealed that that isn't always the case.  I'd like to at least not tie ourselves too tightly to using accessors everywhere.

core/field.ts
```diff
@@ -110,7 +113,28 @@ export abstract class Field<T = any>
    * field is not yet initialized. Is *not* guaranteed to be accurate.
    */
   private tooltip: Tooltip.TipInfo | null = null;
-  protected size_: Size;
+
+  /** This field's dimensions. */
+  private size: Size = new Size(0, 0);
```
cpcallen, line 118: Style guide suggests something like `sizeInternal`:
```suggestion
  private sizeInternal: Size = new Size(0, 0);
```

core/field.ts
maribethb: I prefer the mostly-useless accessors (which most fields can just ignore) over the kludgy approaches mentioned above.

core/field.ts
maribethb: I am confused about this, why? it's already `private`, why would we also add the word `internal`

maribethb (approved):

core/field.ts
cpcallen: Quoting [the Styleguide section on get and set accesors says](go/tsjs-style#classes-getters-and-setters):
> If an accessor is used to hide a class property, the hidden property _may_ be prefixed or suffixed with any whole word, like `internal` or `wrapped`.

They do not provide a rational, but I suspect the reason is to avoid clashing with the name of the accessors.  In this case the accessors are named `size_` (which, with the trailing underscore, is contrary to current naming policy) but if it were `size` then the internal property would need to be named something else.

Issue #9007, opened 2025-05-07
Clearing the potentialVariableMap when clearing flyout thwarts block recycling


riknoll:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

Here is the corresponding makecode issue for this: https://github.com/microsoft/pxt-microbit/issues/6181

MakeCode has recently moved to block inflaters for its flyout caching; we maintain a cache of recycled blocks and use those when creating new elements for our toolbox flyout. Many of our toolbox categories create variables like so:

![Image](https://github.com/user-attachments/assets/10e10f12-3146-4b48-8323-6315883e1018)

These variables are added to the workspace's `potentialVariableMap`, which [gets cleared](https://github.com/google/blockly/blob/bb76d6e12c92230dce4ca5f9c3842c7739788f8e/core/flyout_base.ts#L771) after all of the flyout items are disposed when clearing the flyout. The code for clearing variable maps also deletes all usages of the variables in the map, however the logic seems to exclude potential variables from that deletion which is good because otherwise it would dispose of all the blocks we are recycling.

This generally works as intended except when you drag one of those blocks with a variable out of the flyout. That creates a new variable in the workspace with the same ID as the potential variable. As a result, the variable map clearing code no longer filters out the potential variable usages and deletes all of the cached blocks from the workspace.

I currently have a workaround for this where I remap the variables when recycling blocks if I detect that a variable with the same ID appears in the non-potential variable map of the workspace. Ideally, however, the potential variable and the non-potential variable should have different IDs, or the variable map clearing code should know to skip usages in the flyout.

### Reproduction steps

Reproducing this bug requires you to have a block inflater that recycles blocks like in MakeCode, so here are some steps to reproduce on a past version of MakeCode's micro:bit editor.

1. go to https://makecode.microbit.org/v7.1.30
2. create a new project
3. in the toolbox, expand the "Advanced" section and open the Arrays category
4. drag the "set list to ..." block out of the flyout
5. Open the Arrays category again
6. See the missing blocks

### Stack trace

```shell
No stack trace
```

### Screenshots

see https://github.com/microsoft/pxt-microbit/issues/6181

riknoll:
for reference, here is my PR in makecode that works around this issue:

https://github.com/microsoft/pxt/pull/10566

maribethb:
@riknoll we would take a fix for this if you want to send one, since you have repro steps you can test it with more easily. we are unlikely to get to this for v12 otherwise

PR #9025, opened 2025-05-12
fix: Don't fire events for changes to potential variables

### LINKED ISSUE #9018: Flyout might be being re-rendered too much
### Check for duplicates

### Description

In v12 while testing something else I noticed this warning: `Ignoring non-existant field name in block text_append` in the console. It occurs many many times as long as the flyout is open. So whatever code that triggers that message is being called way too many times.

In v12 you get the message continuously while the flyout is open.
In v11 you get the message only once per flyout open.

It seems this only happens in the advanced playground, so it might be a bad interaction with the advanced playground and something needs fixing there rather than in core

we should probably not have a non-existent field in the built in block, either, but that seems less pressing.

### Reproduction steps

1. Open the advanced playground (not the basic one)
2. Open the `Text` category of the flyout while the console is open
3. Close it and open it again3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

RoboErikG:
## The basics

### Resolves

Fixes #9018 

### Proposed Changes

Adds a param to variable_map to declare it as a potential map and doesn't fire events if it's a potential map.

### Reason for Changes

The potential map firing delete events was causing the toolbox to refresh which deleted potential variables. This repeated indefinitely while the toolbox was open to a category with a variable.

### DIFF
--- a/core/variable_map.ts
+++ b/core/variable_map.ts
@@ -46,8 +46,15 @@ export class VariableMap
     Map<string, IVariableModel<IVariableState>>
   >();
 
-  /** @param workspace The workspace this map belongs to. */
-  constructor(public workspace: Workspace) {}
+  /**
+   * @param workspace The workspace this map belongs to.
+   * @param potentialMap True if this holds variables that don't exist in the
+   *  workspace yet.
+   */
+  constructor(
+    public workspace: Workspace,
+    public potentialMap = false,
+  ) {}
 
   /** Clear the variable map.  Fires events for every deletion. */
   clear() {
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
Also fixes https://github.com/google/blockly-keyboard-experimentation/issues/513

rachel-fenichel:
@riknoll https://github.com/google/blockly/issues/9007 may be related, depending on whether you're looking at the event stream for this.

rachel-fenichel (approved):

PR #9028, opened 2025-05-12
fix: Fix failing browser tests Part 1

### LINKED ISSUE #8953: Browser tests fail with "element not interactable"
### Check for duplicates

### Description

```
/**
 */
  await browser.execute(
    (blockId, newElemId) => {
            return;
          }
        }
      }
    },
    block.id,
    findableId,
  );

  }, findableId);
}
```

### Reproduction steps

1. Check out `rc/v12.0.0`
2. Build
3. Run `npm run test:browser`

### Stack trace

```shell
Right Clicking on Blocks
```

### Screenshots

_No response_

### Browsers

_No response_

RoboErikG:
## The basics

### Resolves

Fixes #8953 

### Proposed Changes

This is still in progress. Changes so far resolve about half of the failing tests.

- Any attempts to click a block will scroll the block into view first.
- Any attempts to click a collapsed block will target the block directly and skip finding a text field.
- Clicking a block is updated to be done by ID and used consistently.
- Updates some positioning on drags from the toolbox to keep in bounds.
- Updates some positioning on drags from the toolbox to avoid early connections.


RoboErikG:
Moved PR to develop now that V12 is out.

PR #9030, opened 2025-05-12
fix: don't visit connections with the cursor.

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/464

### Proposed Changes
This PR updates the navigation policies to not visit connections, but instead proceed directly to the connected block/field. The connection navigation policy has been left in place for now, but is mostly unreachable (outside of a connection being selected in some cases when a block is deleted). This also makes siblings/parents/children semantically accurate for blocks, fields and the workspace, and allows simplifying the `LineCursor`.

### DIFF
--- a/core/keyboard_nav/block_navigation_policy.ts
+++ b/core/keyboard_nav/block_navigation_policy.ts
@@ -4,10 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type {BlockSvg} from '../block_svg.js';
+import {BlockSvg} from '../block_svg.js';
+import type {Field} from '../field.js';
 import type {INavigable} from '../interfaces/i_navigable.js';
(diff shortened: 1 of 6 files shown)


RoboErikG (approved):
Overall looks good and I think it can be submitted after rebasing and doing one more sanity check since I don't think you can test this in the toolbox without the recent fixes to toolboxes (I tested with a rebase, but please double check).

I also found a few issues not directly related to this change that I'll file bugs for:

- You can't add disabled blocks to the workspace from the toolbox (so break can't be added)
- I think if you could add a break getPreviousSibling wouldn't return it
- Empty inputs can't be navigated to which feels wrong, especially with the list blocks, but was mentioned in another thread for discussion

gonfunko:
One issue I realized this created is resolved by https://github.com/google/blockly-keyboard-experimentation/pull/516; PTAL at that as well.

PR #9031, opened 2025-05-12
fix: Improve robustness of `IFocusableNode` uses

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/515

### Proposed Changes

This adds `canBeFocused()` checks to all the places that could currently cause problems if a node were to return `false`.

### Reason for Changes

This can't introduce a problem in current Core and, in fact, most of these classes can never return `false` even through subclasses. However, this adds better robustness and fixes the underlying issue by ensuring that `getFocusableElement()` isn't called for a node that has indicated it cannot be focused.

### Test Coverage

I've manually tested this through the keyboard navigation plugin. However, there are clearly additional tests that would be nice to add both for the traverser and for `WorkspaceSvg`, both likely as part of resolving #8915.

### Documentation

No new documentation changes should be needed here.

### Additional Information

This is fixing theoretical issues in Core, but a real issue tracked by the keyboard navigation plugin repository.

### DIFF
--- a/core/utils/focusable_tree_traverser.ts
+++ b/core/utils/focusable_tree_traverser.ts
@@ -32,13 +32,15 @@ export class FocusableTreeTraverser {
    * @returns The IFocusableNode currently with focus, or null if none.
    */
   static findFocusedNode(tree: IFocusableTree): IFocusableNode | null {
-    const root = tree.getRootFocusableNode().getFocusableElement();
+    const rootNode = tree.getRootFocusableNode();
+    if (!rootNode.canBeFocused()) return null;
+    const root = rootNode.getFocusableElement();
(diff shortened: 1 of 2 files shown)


BenHenning:
Feel free to take a look @rachel-fenichel if you have a free moment--I don't think this PR needs to be specifically reviewed by anyone in particular and I suspect that it's a largely non-controversial change.

BenHenning (commented):
Self-reviewed changes.

BenHenning:
Will be affected a bit by #9033 so waiting for that to be merged, first.

gonfunko (approved):

PR #9032, opened 2025-05-12
fix: Ensure cursor syncs with more than just focused blocks

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/499

### Proposed Changes

This ensures that non-blocks which hold active focus correctly update `LineCursor`'s internal state.

### Reason for Changes

This is outright a correction in how `LineCursor` has worked up until now, and is now possible after several recent changes (most notably #9004). #9004 updated selection to be more explicitly generic (and based on `IFocusableNode`) which means `LineCursor` should also properly support more than just blocks when synchronizing with focus (in place of selection), particularly since lots of non-block things can be focusable.

What's interesting is that this change isn't strictly necessary, even if it is a reasonable correction and improvement in the robustness of `LineCursor`. Essentially everywhere navigation is handled results in a call to `setCurNode` which correctly sets the cursor's internal state (with no specific correction from focus since only blocks were checked and we already ensure that selecting a block correctly focuses it).

### Test Coverage

It would be nice to add test coverage specifically for the cursor cases, but it seems largely unnecessary since:
1. The main failure cases are test-specific (as mentioned above).
2. These flows are better left tested as part of broader accessibility testing (per #8915).

This has been tested with a cursory playthrough of some basic scenarios (block movement, insertion, deletion, copy & paste, context menus, and interacting with fields).

### Documentation

No new documentation should be needed here.

### Additional Information

This is expected to only affect keyboard navigation plugin behaviors, particularly plugin tests.

It may be worth updating `LineCursor` to completely reflect current focus state rather than holding an internal variable. This, in turn, may end up simplifying solving issues like #8793 (but not necessarily).


BenHenning (commented):
Self-reviewed changes.

BenHenning:
@gonfunko I'm curious whether you have any thoughts on this PR, as well, since it's largely touching on navigable things (which may also change with your plans to maybe remove `INavigable` entirely).

gonfunko:
Will look in depth later, but just posted https://github.com/google/blockly/pull/9033 which almost merges the two, leaving INavigable solely for `getClass()`; if we remove that, we could just have both the cursor and the focus infrastructure operate on `IFocusableNode`s, which would be nice.

gonfunko:
INavigable is gone in favor of IFocusableNode in https://github.com/google/blockly/pull/9037

BenHenning:
Will update this once #9037 is merged since it will need to be essentially completely redone (but should be very straightforward to do).

BenHenning (commented):
Self-reviewed the newer version of this that doesn't use `INavigable`.

BenHenning:
FYI I will test this once https://github.com/google/blockly-keyboard-experimentation/pull/520 is merged since I can't verify this against the current keyboard navigation plugin. However, I don't expect there to be any changes in behavior after #9037 has been merged.

maribethb (approved):

BenHenning:
Verified against the latest tip-of-tree keyboard navigation plugin and this seems to be working well. Going ahead and merging.

Issue #8943, opened 2025-04-30
Make all flyout items (with visual representation) focusable


BenHenning:
Arbitrary flyout items that can be added as of v12 (see #8687 for some context on how this affected the old keyboard navigation plugin).

BenHenning:
#8939 addresses the key `FlyoutButton` case, but a longer-term approach may be to change `FlyoutItem` to force an `IFocusableNode` as the element. This would be a breaking change unless it were done in a passive "it _could_ be focusable" way.

gonfunko:
I don't think we can do that because not all flyout items are rendered, and therefore focusable - notably, gaps are FlyoutItems, but they aren't rendered and obviously shouldn't be focusable. They are bounded, however.

BenHenning:
Hmm. In that case, either we'd need to have a special case version of `FlyoutItem` or, at minimum, update some documentation somewhere to explain how to make an item properly focusable for keyboard navigation (i.e. hook it up like `FlyoutButton`). The bigger concern is that "just do what `FlyoutButton` does" won't actually work as advice, however, since there's a lookup component (which is done in `WorkspaceSvg` in this case) to actually make it work correctly.

How much are we expecting custom flyout items that _are_ focusable and aren't part of Core? If the answer's more than zero, then we'll need to think carefully about how to make `Workspace`'s find function composable and customizable for external configuration (I have some ideas on how this might work).

rachel-fenichel:
After #9004 lands I think the state will be:
- Blocks, block stacks, buttons, and labels are all focusable
- Separators are not focusable
- Arbitrary `FlyoutItems` are navigable but not focusable

Making `FlyoutItem` implement `IFocusableNode` will be a breaking change and is thus required for v12.

@BenHenning @gonfunko please confirm whether the above are correct.

gonfunko:
That sounds correct.

BenHenning:
Agreed--making `FlyoutItem` implement `IFocusableNode` is probably the cleanest approach now that focusability can be made optional.

PR #9033, opened 2025-05-12
refactor: Make INavigable extend IFocusableNode.

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/8943

### Proposed Changes
This PR makes `INavigable` extend `IFocusableNode`, which is useful for converting between focused items and elements in the cursor traversal. It also moves `isNavigable()` onto `INavigationPolicy` instead of `INavigable` to allow it to be overridden/configured for classes (like `BlockSvg`, `WorkspaceSvg` and `RenderedConnection`) that are effectively not subclassable.

It may be possible to remove `INavigable` entirely pending the outcome of the discussion in https://github.com/google/blockly/pull/8992#discussion_r2079376896, but this change will be desirable in any case.

### DIFF
--- a/core/navigator.ts
+++ b/core/navigator.ts
@@ -59,8 +59,9 @@ export class Navigator {
     const result = this.get(current)?.getFirstChild(current);
     if (!result) return null;
     // If the child isn't navigable, don't traverse into it; check its peers.
-    if (!this.get(result)?.isNavigable(result))
+    if (!this.get(result)?.isNavigable(result)) {
       return this.getNextSibling(result);
+    }
     return result;
   }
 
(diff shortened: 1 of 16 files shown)


core/navigator.ts
```diff
@@ -59,7 +59,8 @@ export class Navigator {
     const result = this.get(current)?.getFirstChild(current);
     if (!result) return null;
     // If the child isn't navigable, don't traverse into it; check its peers.
-    if (!result.isNavigable()) return this.getNextSibling(result);
+    if (!this.get(result)?.isNavigable(result))
```
BenHenning, line 62: Nit: please add braces since this breaks to a new line.

Ditto for lines 89-90 and 105-106 below.

BenHenning (approved):
Thanks @gonfunko. Just one nit, otherwise LGTM!

core/navigator.ts
gonfunko: Good catch, I...do not care for that formatter behavior. Fixed!

PR #9034, opened 2025-05-12
fix: Ensure selection stays when dragging blocks, comments, and bubbles

### LINKED ISSUE #9027: Dragging blocks stops their selection highlight
### Check for duplicates

### Description

Attempting to drag a block results in the block's selection highlight disappearing until dragging finishes (at which point it's restored).

### Reproduction steps

1. Open the playground for v12.
2. Insert any block.
3. Click to drag the block around.
4. Observe that selection highlight is missing during dragging.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9027
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/521

### Proposed Changes

Ensure that a block & workspace comment being dragged are properly focused mid-drag.

### Reason for Changes

Focus seems to be lost due to the element being moved to the drag layer, so re-focusing the element ensures that it remains both actively focused and selected while dragging. This applies to both blocks and workspace comments, and theoretically bubbles (though that's only for an event basis since they don't actually have a selection highlight).

The regressions were likely caused when block and comment selection was moved to be fully synced based on active focus.

Note that there are also changes for the selection block path. It was noticed when trying to fix https://github.com/google/blockly-keyboard-experimentation/issues/521 that since a selection highlight is a complete copy of a block's path it was also copying the ID and CSS state for that block. It just so happened that the block was actually receiving passive focus at the time of re-render that generated the selection highlight and thus that state took precedence in the new path object. This was likely due to rendering happening while moving the block over to the drag layer (and thus during its brief window of having lost focus). The fix is to simply ensure that all focus-related attributes are properly removed from the cloned object.

There is one other location where a block's path is cloned: `block_animations.ts` for disposing of the block. I suspect this won't actually require the additional attribute removal being done for the selection highlight, so this case hasn't been fixed. My local testing of block deletion seems to suggest it looks correct, but there could well be edge cases that I haven't considered.

### Test Coverage

This has been manually verified in Core's simple playground and in the keyboard navigation plugin's test environment.

It would be helpful to add a new test case for the underlying problem (i.e. ensuring that the block holds focus mid-drag) as part of resolving #8915.

### Documentation

No new documentation should need to be added.

### Additional Information

This was found during the development of https://github.com/google/blockly-keyboard-experimentation/pull/511.

### DIFF
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -14,7 +14,6 @@ import {ConnectionType} from '../connection_type.js';
 import type {BlockMove} from '../events/events_block_move.js';
 import {EventType} from '../events/type.js';
 import * as eventUtils from '../events/utils.js';
-import {getFocusManager} from '../focus_manager.js';
 import {IConnectionPreviewer} from '../interfaces/i_connection_previewer.js';
 import {IDragStrategy} from '../interfaces/i_draggable.js';
 import * as layers from '../layers.js';
@@ -121,10 +120,6 @@ export class BlockDragStrategy implements IDragStrategy {
     }
     this.block.setDragging(true);
     this.workspace.getLayerManager()?.moveToDragLayer(this.block);
-
-    // Since moving the block to the drag layer will cause it to lose focus,
-    // ensure it regains focus (to enable the block's selection highlight).
-    getFocusManager().focusNode(this.block);
   }
 
   /**
@@ -399,10 +394,6 @@ export class BlockDragStrategy implements IDragStrategy {
         .getLayerManager()
         ?.moveOffDragLayer(this.block, layers.BLOCK);
       this.block.setDragging(false);
-
-      // Since moving the block off the drag layer will cause it to lose focus,
-      // ensure it regains focus (to enable the block's selection highlight).
-      getFocusManager().focusNode(this.block);
     }
 
     if (this.connectionCandidate) {
--- a/core/dragging/block_drag_strategy.ts
+++ b/core/dragging/block_drag_strategy.ts
@@ -399,6 +399,10 @@ export class BlockDragStrategy implements IDragStrategy {
         .getLayerManager()
         ?.moveOffDragLayer(this.block, layers.BLOCK);
       this.block.setDragging(false);
+
+      // Since moving the block off the drag layer will cause it to lose focus,
+      // ensure it regains focus (to enable the block's selection highlight).
+      getFocusManager().focusNode(this.block);
     }
 
     if (this.connectionCandidate) {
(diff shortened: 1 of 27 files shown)


BenHenning (commented):
Self-reviewed changes (added some additional comment lines for clarity).

rachel-fenichel:
Is there a bug for "At the time of the PR being opened, this couldn't be tested in the test environment for the experimental keyboard navigation plugin since there's a navigation connection issue there that needs to be resolved to test movement"?

core/dragging/block_drag_strategy.ts
```diff
@@ -120,6 +121,10 @@ export class BlockDragStrategy implements IDragStrategy {
     }
     this.block.setDragging(true);
     this.workspace.getLayerManager()?.moveToDragLayer(this.block);
+
+    // Since moving the block to the drag layer will cause it to lose focus,
+    // ensure it regains focus (to enable the block's selection highlight).
+    getFocusManager().focusNode(this.block);
```
rachel-fenichel, line 127: Question: should this be done in `moveToDragLayer` (and `moveOffDragLayer`) instead of in the drag strategy?

BenHenning: Good idea. This actually made me realize that workspace comments had the same problem, so this shift actually fixes those, too.

BenHenning (commented):
Self-reviewed the latest (basically a new PR at this point).

BenHenning:
> Is there a bug for "At the time of the PR being opened, this couldn't be tested in the test environment for the experimental keyboard navigation plugin since there's a navigation connection issue there that needs to be resolved to test movement"?

I don't think there was a bug but https://github.com/google/blockly-keyboard-experimentation/pull/516 fortunately fixed the underlying issue so I was able to properly test this with keyboard navigation & fix another bug that I found: https://github.com/google/blockly-keyboard-experimentation/issues/521.

I've updated both the code and PR description accordingly, PTAL.

tests/mocha/layering_test.js
```diff
@@ -80,7 +89,7 @@ suite('Layering', function () {
   });
 
   suite('dragging', function () {
-    test('moving an element to the drag layer adds it to the drag group', function () {
+    test.only('moving an element to the drag layer adds it to the drag group', function () {
```
rachel-fenichel, line 92: Reset to `test(` so you're not skipping all the other tests.

BenHenning: I forget this far too often until I do my self-review. :) Thanks & removed in latest.

BenHenning (commented):
Self-reviewed latest.

BenHenning:
Looks like tests are now both running and passing. PTAL @rachel-fenichel.

rachel-fenichel (approved):

PR #9037, opened 2025-05-13
refactor: Remove INavigable in favor of IFocusableNode.

gonfunko:
## The basics

### Proposed Changes
This PR gets rid of the `INavigable` interface, since it was reduced to just `getClass()`, and there was [significant objection](https://github.com/google/blockly/pull/8992#discussion_r2079376896) to that. Instead, we now act directly on `IFocusableNode` instances. As a side effect, this also makes navigation policies Just Work for `Field` subclasses, and opens up the possibility of removing the cursor's `curNode` in favor of asking the focus manager for the currently focused thing, although I'll address that in a followup change. Registering navigation policies has been moved out of the cursor and into the `Navigator`.


core/interfaces/i_navigation_policy.ts
```diff
@@ -57,4 +56,13 @@ export interface INavigationPolicy<T> {
    * @returns True if this element should be included in keyboard navigation.
    */
   isNavigable(current: T): boolean;
+
+  /**
+   * Returns whether or not this navigation policy corresponds to the type of
+   * the given object.
+   *
+   * @param current An instance to check whether this policy applies to.
+   * @returns True if the given object is of a type handled by this policy.
```
BenHenning, line 65: A nit but I don't think this actually returns a boolean, I think it returns `current` coerced to type `T` or `undefined` if coercion fails per the predicate.

I'm not 100% certain on this and I find the TypeScript documentation unclear, but I recall debugging guards like this before and being surprised that their return values weren't actually booleans.

core/keyboard_nav/flyout_navigation_policy.ts
```diff
@@ -21,15 +21,17 @@ export class FlyoutNavigationPolicy<T> implements INavigationPolicy<T> {
   constructor(
     private policy: INavigationPolicy<T>,
     private flyout: IFlyout,
-  ) {}
+  ) {
+    console.log('make flyout nav policy');
```
BenHenning, line 25: Probably want to remove this debug line.

Ditto for the others in this file.

core/navigator.ts
```diff
@@ -18,35 +22,41 @@ export class Navigator {
...
    */
-  private get<T extends INavigable<T>>(
-    key: T,
-  ): INavigationPolicy<T> | undefined {
-    return this.rules.get(key.getClass());
+  private get(
+    current: IFocusableNode,
+  ): INavigationPolicy<typeof current> | undefined {
+    for (const rule of this.rules) {
+      if (rule.isApplicable(current)) {
```
BenHenning, line 54: I wonder if ordering will matter here if a policy is over-zealous on which types it can claim.

core/navigator.ts
```diff
@@ -18,35 +22,41 @@ export class Navigator {
...
    *     undefined if no ruleset has been registered for the object's class.
    */
-  private get<T extends INavigable<T>>(
-    key: T,
-  ): INavigationPolicy<T> | undefined {
-    return this.rules.get(key.getClass());
+  private get(
+    current: IFocusableNode,
+  ): INavigationPolicy<typeof current> | undefined {
+    for (const rule of this.rules) {
```
BenHenning, line 53: Optional, but could also be replaced with a find, e.g.:

```typescript
return this.rules.find((rule) => rule.isApplicable(current));
```

core/flyout_navigator.ts
```diff
@@ -0,0 +1,24 @@
...
+import {FlyoutSeparatorNavigationPolicy} from './keyboard_nav/flyout_separator_navigation_policy.js';
+import {Navigator} from './navigator.js';
+
+export class FlyoutNavigator extends Navigator {
+  constructor(flyout: IFlyout) {
+    super();
+    this.rules.push(
+      new FlyoutButtonNavigationPolicy(),
+      new FlyoutSeparatorNavigationPolicy(),
+    );
```
BenHenning, line 19: Would it perhaps be cleaner to introduce something like a protected init method that takes the list of rules to initialize? Then this navigator could fully override the behavior and very specifically indicate which rules should be present (rather than only being able to add onto them, or needing to overwrite the list).

This also would avoid needing to over-expose the field.

core/blockly.ts
```diff
@@ -429,7 +429,7 @@ Names.prototype.populateProcedures = function (
 };
 // clang-format on
 
-export * from './interfaces/i_navigable.js';
+export * from './flyout_navigator.js';
```
BenHenning, line 432: Is anything actually being exported from here?

core/keyboard_nav/line_cursor.ts
```diff
@@ -14,120 +14,27 @@
...
-
-  /**
-   * Registers default navigation policies for Blockly's built-in types with
-   * this cursor's workspace.
-   */
-  protected registerNavigationPolicies() {
-    const navigator = this.workspace.getNavigator();
-
-    const blockPolicy = new BlockNavigationPolicy();
-    if (this.workspace.isFlyout) {
```
BenHenning, line 95: I like the refactor to use a separate navigator for flyout--that seems much cleaner.

BenHenning (approved):
Thanks @gonfunko! Had some thoughts.

Approving to unblock, but I'm happy to take another pass. Didn't look at the flyout navigation policy super closely since it appeared to have a bunch of unfinished code int.

core/keyboard_nav/flyout_navigation_policy.ts
rachel-fenichel, line 25: nit: remove old console.log

core/keyboard_nav/flyout_navigation_policy.ts
```diff
@@ -49,20 +51,29 @@ export class FlyoutNavigationPolicy<T> implements INavigationPolicy<T> {
    * @param current The flyout item to navigate from.
    * @returns The flyout item following the given one.
    */
-  getNextSibling(current: T): INavigable<unknown> | null {
+  getNextSibling(current: T): IFocusableNode | null {
+    console.log('flyout next');
```
rachel-fenichel, line 55: Remove console.log

rachel-fenichel: Here and in the rest of the file.

rachel-fenichel (approved):
Approved mod some deletion of logs.

This feels solid.

core/interfaces/i_navigation_policy.ts
```diff
@@ -3,8 +3,7 @@
  * Copyright 2025 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
-
```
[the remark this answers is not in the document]
gonfunko: Done

core/keyboard_nav/flyout_navigation_policy.ts
gonfunko: Done

core/blockly.ts
gonfunko: Yes, the `FlyoutNavigator` class, which developers may wish to subclass.

core/keyboard_nav/flyout_navigation_policy.ts
gonfunko: Done

core/keyboard_nav/flyout_navigation_policy.ts
gonfunko: Done

core/navigator.ts
gonfunko: Done

core/navigator.ts
gonfunko: It could, but I suspect in practice that would be rare, and could be worked around by a navigator subclass that registers different rules (perhaps just subclassing and overriding `isApplicable()` to be more restrictive)

core/interfaces/i_navigation_policy.ts
gonfunko: I agree the docs are unclear on this, but I think `boolean` is more useful to both folks reading and implementing than the perhaps technically correct value; ultimately the method needs to return a boolean (if you're implementing it) and you can rely on it being a truthy value if you call it.

BenHenning: I used the language 'whether' in this case which I suppose is slightly more ambiguous (perhaps beneficially):

```typescript
/**
 * Determines whether the provided object fulfills the contract of
 * IFocusableNode.
 *
 * @param object The object to test.
 * @returns Whether the provided object can be used as an IFocusableNode.
 */
export function isFocusableNode(object: any | null): object is IFocusableNode {
  return (
    object &&
    'getFocusableElement' in object &&
    'getFocusableTree' in object &&
    'onNodeFocus' in object &&
    'onNodeBlur' in object &&
    'canBeFocused' in object
  );
}
```

Not sure that's actually better. Note I don't have strong opinions here, so if you prefer the explicit 'True' that seems reasonable to me.

core/flyout_navigator.ts
gonfunko: Attempted this, but TS/JS initialization rules make it complicated. Punting for now.

PR #9045, opened 2025-05-13
fix: Ensure BlockSvg is uniquely focusable on the page

### LINKED ISSUE #9043: Block IDs can repeat across multiple uses
### Check for duplicates

### Description

Block IDs are not guaranteed to be re-generated when inserting a block from the flyout, and can even be duplicated in other contexts (such as with the minimap plugin--see https://github.com/google/blockly-samples/issues/2512).

Note that this issue here is _not_ that IDs are reused, it's that the focus system can have multiple nodes with identical IDs. This can result in unpredictable or outright incorrect behavior (see https://github.com/google/blockly-keyboard-experimentation/issues/521 for one example of what can happen when IDs are duplicated).

The fix here isn't to change core behavior but, rather, to change what the focus system relies upon.

### Reproduction steps

1. Open the simple test playground for the latest version of v12.
2. Add a block from the flyout.
3. Observe that the block's ID in the workspace is the same as it was in the flyout.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9043
Fixes https://github.com/google/blockly-samples/issues/2512

### Proposed Changes

This replaces using BlockSvg's own ID for focus management since that's not guaranteed to be unique across all workspaces on the page.

### Reason for Changes

Both https://github.com/google/blockly-samples/issues/2512 covers the user-facing issue in more detail, but from a technical perspective it's possible for blocks to share IDs across workspaces. One easy demonstration of this is the flyout: the first block created from the flyout to the main workspace will share an ID. The workspace minimap plugin just makes the underlying problem more obvious.

The reason this introduces a breakage is due to the inherent ordering that `FocusManager` uses when trying to find a matching tree for a given DOM element that has received focus. These trees are iterated in the order of their registration, so it's quite possible for some cases (like main workspace vs. flyout) to resolve such that the behavior looks correct to users, vs. others (such as the workspace minimap) not behaving as expected.

Guaranteeing ID uniqueness across all workspaces fixes the problem entirely.

### Test Coverage

This has been manually tested in core Blockly's simple test playground and in Blockly samples' workspace minimap plugin test environment (linked against this change). See the new behavior for the minimap plugin:

[Screen recording 2025-05-13 4.31.31 PM.webm](https://github.com/user-attachments/assets/d2ec3621-6e86-4932-ae85-333b0e7015e1)

Note that this is a regression to v11 behavior in that the blocks in the minimap now show as selected.

This has been verified as working with the latest version of the keyboard navigation plugin (tip-of-tree). Keyboard-based block operations and movement seem to work as expected.

For automated testing this is expected to largely be covered by future tests added as part of resolving #8915.

### Documentation

No public documentation changes should be needed, though `IFocusableNode`'s documentation has been refined to be clearer on the uniqueness property for focusable element IDs.

### Additional Information

There's a separate open design question here about whether `BlockSvg`'s descendants should use the new focus ID vs. the block ID. Here is what I consider to be the trade-off analysis in this decision:

|                        | Pros                                            | Cons                                                                         |
|------------------------|-------------------------------------------------|------------------------------------------------------------------------------|
| Use `BlockSvg.id`      | Can use fast `WorkspaceSvg.getBlockById`.       | `WorkspaceSvg.lookUpFocusableNode` now uses 2 different IDs.                 |
| Use `BlockSvg.focusId` | Consistency in IDs use for block-related focus. | Requires more expensive block look-up in `WorkspaceSvg.lookUpFocusableNode`. |

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -55,6 +55,7 @@ import type {IPathObject} from './renderers/common/i_path_object.js';
 import * as blocks from './serialization/blocks.js';
 import type {BlockStyle} from './theme.js';
 import * as Tooltip from './tooltip.js';
+import {idGenerator} from './utils.js';
 import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
 import {Rect} from './utils/rect.js';
@@ -210,7 +211,9 @@ export class BlockSvg
 
     // Expose this block's ID on its top-level SVG group.
     this.svgGroup.setAttribute('data-id', this.id);
-    svgPath.id = this.id;
+
+    // The page-wide unique ID of this Block used for focusing.
+    svgPath.id = idGenerator.getNextUniqueId();
 
     this.doInit_();
   }
--- a/core/interfaces/i_focusable_node.ts
+++ b/core/interfaces/i_focusable_node.ts
@@ -19,11 +19,13 @@ export interface IFocusableNode {
    * - blocklyActiveFocus
    * - blocklyPassiveFocus
    *
-   * The returned element must also have a valid ID specified, and unique to the
-   * element relative to its nearest IFocusableTree parent. It must also have a
-   * negative tabindex (since the focus manager itself will manage its tab index
-   * and a tab index must be present in order for the element to be focusable in
-   * the DOM).
+   * The returned element must also have a valid ID specified, and unique across
+   * the entire page. Failing to have a properly unique ID could result in
+   * trying to focus one node (such as via a mouse click) leading to another
+   * node with the same ID actually becoming focused by FocusManager. The
+   * returned element must also have a negative tabindex (since the focus
+   * manager itself will manage its tab index and a tab index must be present in
+   * order for the element to be focusable in the DOM).
    *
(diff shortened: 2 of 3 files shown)


BenHenning:
PTAL @maribethb though be aware that there's an open design question that I'd like input on, and additional testing to do once https://github.com/google/blockly-keyboard-experimentation/pull/520 is merged.

maribethb (approved):
I think I would use the first option in your design question. The reason being that it might be possible to have globally unique block IDs in the future and then we could just revert this change.

We actually tried to have globally unique block ids in a previous release, but code.org was relying on the current behavior in their tests so it became a blocking/breaking change for them. I think Mike said he was working on their tests such that that would no longer be a blocker. See https://github.com/google/blockly/issues/7432 for historic context

As an aside, I think the correct behavior for the minimap would be to select the block in the main workspace when a block in the minimap is selected and I wonder if that's now possible with navigation policies?

BenHenning:
Confirming that this works well against the keyboard navigation plugin tip-of-tree.

PR #9046, opened 2025-05-13
feat: Add CSS classes for more focus states.

### LINKED ISSUE #9044: Introduce CSS classes for a variety of focus states
### Check for duplicates

### Problem

Some various states of styling in the keyboard navigation plugin are difficult today, see: https://github.com/google/blockly-keyboard-experimentation/pull/511.

### Request

Many of the 'has' styling cases (and some of the other existing, unsolved problems) could benefit from having a few new CSS classes that properly represent different state situations:
- `blocklyTreeHasActiveFocus` for cases when a tree has a node with active focus that isn't the tree root.
- `blocklyTreeHasPassiveFocus` for cases when a tree has a node with passive focus that isn't the tree root.
- `blocklySubtreeHasActiveFocus` for cases when a tree has a subtree (such as the flyout's workspace) has a node with active focus that isn't the subtree root.
- `blocklySubtreeHasPassiveFocus` for cases when a tree has a subtree (such as the flyout's workspace) has a node with passive focus that isn't the subtree root.
- `blocklyFieldHasEditorOpen` for cases when a field is being actively edited.

### Alternatives considered

Using the `:has()` CSS pseudo-class, however this requires a version of Safari that may introduce compatibility issues for Blockly & Blockly applications. It also may not be sufficient to solve all known styling problems that have been identified in the keyboard navigation plugin.

### Additional context

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9044

### Proposed Changes

TODO: Finish this PR description.


BenHenning:
NB: This isn't completed yet as subtrees still need to be added, and a lot more testing needs to be conducted.

I'm also unlikely to be comfortable merging this without new focus manager tests as the edge cases here can be quite complex.

BenHenning:
I was originally pushing to try and get this into v12, though while that has become increasingly less likely as v12 approaches launch I also no longer want to. The CSS changes for fields and trees are relatively benign, but adding the new CSS classes for subtrees actually introduces non-trivial risks:
- It involves adding a lot more potentially very expensive (but hopefully not) tree crawls for every node change in order to properly calculate parent trees that are inheriting an active/passive node.
- The act of interacting with all these trees could stir up new problems if there are certain conditions where trees are partly deleted or in a bad state. While those bugs could already exist today, this logic would increase the likelihood of hitting them and introducing a catastrophic failure.

While I actually expect the overall changes to end up being safe and sufficiently performant, the risk is too high to risk trying to rush this into v12.

BenHenning:
I think this actually implements all of the CSS classes that we need per #9044 except for the subtree ones.

Those are...a bit hard to implement correctly. I'm actually considering inverting the way subtrees are provided so that they are actually indicated through registration. This will probably require us to do cycle checking, but the current pattern doesn't prevent that and we don't currently check for it. :)

The advantage of explicit registration is that it would make it much easier to notify parents when something changes in a subtree, including multiple levels of updates, such as active/passive focus changing, i.e., this PR. Since we already require trees to register, this isn't a big difference in management for callers and, to some extent, actually makes things a bit easier for ensuring that unregistered subtrees aren't accidentally represented.

That's a pretty big change, and we'll need to file a follow-up bug to fully remove the nested trees function in v13. For now I think we can deprecate the function and remove all calls to it (plus update its documentation accordingly). Basically no one should be using subtrees outside core, anyway.

Issue #8821, opened 2025-03-27
Tab between editable fields


rachel-fenichel:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Problem

https://github.com/google/blockly/pull/8803 removes the explicit creation of a `TabNavigateCursor`, because tab is now used to move within the page instead of within Blockly.

However, it's fairly standard on the web for tab to navigate to the next editable field _when you are already editing_.

### Request

@kmcnaught [suggested](https://github.com/google/blockly/issues/8792#issuecomment-2743211735) "How about a middle ground of navigating to the next editable field within the same block and opening the editor, falling back to doing nothing if there are no more editable fields in that block? I can see that being able to tab to the next field is a nicer flow than having to Esc and arrow multiple times."

Reimplement tabbing between editable input fields, per the suggestion above. Put this code in the `LineCursor` instead of creating a new cursor just-in-time. Wire up to the appropriate tab listener.

### Alternatives considered

Remove tabbing between fields entirely.

BenHenning:
One possible approach once the constituent pieces of #8875 are merged: when a field editor is opened, update all of the source block's field focusable elements to have a tab index of 0 (which makes them automatically tabbable), and reset them to -1 on the field editor closing.

rachel-fenichel:
Current status (as far as I know): @gonfunko and @BenHenning have a plan to implement this with a tab listener rather than setting tab indices.

Not yet in progress, but is required for v12.

gonfunko:
Yes, that's the idea; a tab listener delegating to the navigator.

PR #9049, opened 2025-05-14
fix: Reenable support for tabbing between fields.

gonfunko:
## The basics

### Resolves

Fixes  https://github.com/google/blockly/issues/8821

### Proposed Changes
This PR re-adds support for tabbing between editable fields to correct a regression. This is a bit complicated because, in Zelos, navigation is to blocks rather than fields for simple reporters, but it appears to work correctly in both Zelos and non-Zelos renderers.

### DIFF
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -582,6 +582,38 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
       );
       WidgetDiv.hideIfOwner(this);
       dropDownDiv.hideWithoutAnimation();
+    } else if (e.key === 'Tab') {
+      e.preventDefault();
+      const cursor = this.workspace_?.getCursor();
+      const target = e.shiftKey
+        ? cursor?.getPreviousNode(
+            this,
+            (node) =>
+              (node instanceof FieldInput ||
+                (node instanceof BlockSvg && node.isSimpleReporter())) &&
+              node !== this.getSourceBlock(),
+            false,
+          )
+        : cursor?.getNextNode(
+            this,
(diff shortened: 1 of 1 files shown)


core/field_input.ts
```diff
@@ -582,6 +582,38 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
       );
       WidgetDiv.hideIfOwner(this);
       dropDownDiv.hideWithoutAnimation();
+    } else if (e.key === 'Tab') {
+      e.preventDefault();
+      const cursor = this.workspace_?.getCursor();
+      const target = e.shiftKey
+        ? cursor?.getPreviousNode(
+            this,
+            (node) =>
```
rachel-fenichel, line 591: Optional: these predicates are the same and could be pulled out into a helper function to make this more readable.

rachel-fenichel:
Can get tests added as part of https://github.com/google/blockly/issues/8915

core/field_input.ts
```diff
@@ -582,6 +583,28 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
+        : cursor?.getNextNode(this, isValidDestination, false);
+      target =
+        target instanceof BlockSvg && target.isSimpleReporter()
+          ? target.getFields().next().value
+          : target;
+
+      if (target instanceof FieldInput) {
+        WidgetDiv.hideIfOwner(this);
+        dropDownDiv.hideWithoutAnimation();
+        target.showEditor();
```
BenHenning, line 606: Does this correctly focus the new field, too?

PR #9056, opened 2025-05-15
chore: Rollup of 2025 Q2 updates from TranslateWiki

gonfunko:
## The basics

### Proposed Changes
Update messages in msg/json/ from TranslateWiki.


msg/json/en.json
```diff
@@ -370,7 +369,6 @@
 	"PROCEDURES_DEFNORETURN_PROCEDURE": "do something",
 	"PROCEDURES_BEFORE_PARAMS": "with:",
 	"PROCEDURES_CALL_BEFORE_PARAMS": "with:",
-	"PROCEDURES_CALL_DISABLED_DEF_WARNING": "Can't run the user-defined function '%1' because the definition block is disabled.",
```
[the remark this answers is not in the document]
gonfunko: Done.

msg/json/constants.json
```diff
@@ -1,12 +1 @@
-{
```
maribethb, line 1: surprised that this got reformatted, but don't see any problems with it

msg/json/en.json
```diff
@@ -1,7 +1,7 @@
 {
 	"@metadata": {
 		"author": "Ellen Spertus <ellen.spertus@gmail.com>",
-		"lastupdated": "2024-04-16 23:19:53.668551",
+		"lastupdated": "2023-12-08 18:42:04.679586",
```
maribethb, line 4: this is also a really weird change lol but assuming it's just from the way we just straight up replace this file instead of actually merging it

msg/json/en.json
```diff
@@ -291,7 +291,7 @@
 	"LISTS_ISEMPTY_TITLE": "%1 is empty",
 	"LISTS_ISEMPTY_TOOLTIP": "Returns true if the list is empty.",
 	"LISTS_INLIST": "in list",
-	"LISTS_INDEX_OF_HELPURL": "https://github.com/google/blockly/wiki/Lists#finding-items-in-a-list",
+	"LISTS_INDEX_OF_HELPURL": "https://github.com/google/blockly/wiki/Lists#getting-items-from-a-list",
```
maribethb, line 294: I think the previous version of this is more correct, so it's possible this entire file should just be reverted as this is somehow putting back an old change (hence updating the date backwards and using a worse link)

msg/json/synonyms.json
```diff
@@ -1,22 +1 @@
-{
```
maribethb, line 1: another suspiciously reformatted file... I think this is one that I manually reverted in https://github.com/google/blockly/pull/8162 so you may want to do the same thing

maribethb (commented):
I don't see any more problematic deletiions, but I think you might want to revert some of the files a la https://github.com/google/blockly/pull/8162

msg/json/en.json
gonfunko: Reverted as per your second comment.

msg/json/en.json
gonfunko: Done

msg/json/synonyms.json
gonfunko: Done

maribethb (approved):

PR #9062, opened 2025-05-15
fix: Limit `LineCursor` & focus syncing

### LINKED ISSUE #526: Toolbox flyout navigation becomes permanently broken after trying to select a FlyoutButton
### Check for duplicates

### Description

Attempting to navigate to a flyout button with keyboard navigation permanently breaks flyout navigation.

### Reproduction steps

1. Check out 8a9e9d98372ada7228ae7bc5297bc6ba10b7b7f4.
2. Ensure that it's updated to use v12 beta.7 and has been correctly `npm install`'d and `npm ci`'d.
3. Start the local test environment with `npm run start`.
4. Open the test environment locally and tab navigate to the toolbox.
5. Use down arrow keys until the Variables category is selected.
6. Press the 'right' arrow key to move focus to the flyout.
7. Press the 'left' arrow key to move focus back to the toolbox.
8. Attempt to navigate to any toolbox category flyout and notice that the highlight no longer works. It's also not actually hidden, it's truly not selecting (since blocks cannot be inserted anymore).

### Stack trace

```shell
N/A
```

### Screenshots

Video repro:

https://github.com/user-attachments/assets/20f17791-2b3a-4ba4-8f53-600d15bfd8b2

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/526

### Proposed Changes

This limits synchronizing in `LineCursor` from `FocusManager` to just nodes that have a corresponding block.

### Reason for Changes

Limiting the synchronizing in this way ensures that navigation can't enter a bad state. The reason for why this is needed is explained in https://github.com/google/blockly-keyboard-experimentation/issues/526#issuecomment-2885117998.

Longer term it would maybe be ideal to do one or both of the following:
- Figure out ways of making navigation a bit more robust (perhaps on the keyboard navigation side) such that if cursor _is_ in a bad state there's some way to recover (rather than ending up permanently broken).
- Remove `Marker`'s internal state in favor of always relying on `FocusManager`'s state to cover the cases where there can be automatic focus shifting.

### Test Coverage

This was manually tested with the keyboard navigation plugin and verified to ensure that both https://github.com/google/blockly-keyboard-experimentation/issues/526 and https://github.com/google/blockly-keyboard-experimentation/issues/499 are (still) working as expected. Some basic testing was done with the core simple playground with the developer console open to ensure there weren't any expected failures.

Automated testing cases would be better addressed as part of resolving #8915.

### Documentation

No new documentation is needed here.

### Additional Information

This behavior is expected to only affect the keyboard navigation plugin.

### DIFF
--- a/core/keyboard_nav/line_cursor.ts
+++ b/core/keyboard_nav/line_cursor.ts
@@ -377,7 +377,7 @@ export class LineCursor extends Marker {
     // Ensure the current node matches what's currently focused.
     const focused = getFocusManager().getFocusedNode();
     const block = this.getSourceBlockFromNode(focused);
-    if (!block || block.workspace === this.workspace) {
+    if (block && block.workspace === this.workspace) {
       // If the current focused node corresponds to a block then ensure that it
       // belongs to the correct workspace for this cursor.
       this.setCurNode(focused);


BenHenning (commented):
Self-reviewed change.

BenHenning:
Assigning over to you @rachel-fenichel mainly in case you have any high-level thoughts.

rachel-fenichel (approved):

PR #9054, opened 2025-05-15
fix: Make clickable but non-editable fields navigable.

gonfunko:
## The basics

- [x] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves
This PR fixes an issue that prevented fields that were clickable/interactable but not editable from being visited via keyboard navigation.

PR #9063, opened 2025-05-15
fix: Ensure `FieldImage` is clickable when appropriate

BenHenning:
## The basics

### Resolves

Fixes part of https://github.com/google/blockly-keyboard-experimentation/issues/525

### Proposed Changes

This introduces an `isClickable` override for `FieldImage` that only allows images to be clickable when they have a click handler.

### Reason for Changes

Without this change it's possible to navigate to non-clickable images (like the quotes in certain text blocks) since #9054. This is both a confusing experience and causing a test failure in the keyboard navigation plugin: https://github.com/google/blockly-keyboard-experimentation/pull/527#issuecomment-2885234898.

### Test Coverage

New tests have been added for `FieldImage.isClickable()` and have been specifically verified such that that the following new tests correctly fail without the fix in place:
- `For attached, enabled, editable field without click handler returns false`
- `For attached, enabled, editable field with removed click handler returns false`

The first one perfectly covers the case that caused the regression, so these tests should help ensure the regression doesn't occur again.

Additionally, the tests in https://github.com/google/blockly-keyboard-experimentation/pull/527 have been verified against this PR via linking and verified to pass.

I've manually tested the scenarios involving quotes (since that was the direct failure case) in the keyboard navigation plugin test environment and placed some similar blocks in core Blockly's simple playground (while checking console logs) to verify nothing obvious is broken.

### Documentation

No new documentation changes should be needed.

### Additional Information

This is a fix suggested over chat by @gonfunko when helping to investigate why the new test was found failing. Note that this failure is a discrepancy between beta.7 and the full 12.0.0 release.

### DIFF
--- a/core/field_image.ts
+++ b/core/field_image.ts
@@ -213,6 +213,17 @@ export class FieldImage extends Field<string> {
   }
 
   /**
+   * Check whether this field should be clickable.
+   *
+   * @returns Whether this field is clickable.
+   */
+  isClickable(): boolean {
+    // Images are only clickable if they have a click handler and fulfill the
+    // contract to be clickable: enabled and attached to an editable block.
+    return super.isClickable() && !!this.clickHandler;
+  }
+
+  /**
    * If field click is called, and click handler defined,
    * call the handler.
    */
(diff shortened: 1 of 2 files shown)


BenHenning (commented):
Self-reviewed changes (quick spot check).

RoboErikG (approved):

PR #9079, opened 2025-05-20
feat: Automatically manage focus tree tab indexes

### LINKED ISSUE #523: Extra toolbox tab position
1. In the latest demo (built locally) tab from the Renderer `select` to the toolbox.
2. Try to shift-tab back
3. Note you need to do it twice to get back and the first tab stop cannot be seen

I'm not convinced there are two meaningful positions here and there are not [in the last deployed demo](https://google.github.io/blockly-keyboard-experimentation/) at the time of writing.

### LINKED ISSUE #547: Focus is not moving properly from flyout to workspace in one circumstance
### Check for duplicates

### Description

When focussed on a toolbox category, if you move right, then hit Tab, focus moves to the main workspace but the focus manager seems unaware that anything changed (the flyout isn't closed, and class names are not updated).

### Reproduction steps

1. Tab to the toolbox which opens the flyout for the first category
2. Use ArrowRight to move focus into the flyout which focuses the first block
3. Hit the Tab key
4. Browser focus moves to the main workspace but the focus manager seems unaware that anything happened

### Interestingly the following situations work as expected:

1. Tab to the toolbox which opens the flyout for the first category
2. **Hit the Tab key to focus the flyout**
3. Hit the Tab key to focus the main workspace

or

1. Tab to the toolbox which opens the flyout for the first category
2. Use ArrowRight to move focus into the flyout which focuses the first block
3. **Use the ArrowDown key to move focus to the next block**
4. Hit the Tab key to focus the main workspace

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/78d2622e-c806-47c9-960a-7533355c4c94

### Browsers

_No response_

### LINKED ISSUE #8965: Shift+Tab after keyboard navigating to a block doesn’t open flyout as expected
### Check for duplicates

### Description

Shift+Tab after keyboard navigating to a block doesn’t open flyout as expected. There seems to be an extra workspace tab stop that causes the workspace outline to appear.

Ben: this is not a simple fix, and probably needs discussion. The workspace is a browser-level tab stop that we have some control on overriding to switch focus, but it's unclear if the user wants to be able to navigate back to the workspace. If we want to provide that functionality, it has to stay as a tabstop. We do override first-time focus on the workspace to select the first block (if one is present), but it is possible to navigate back to the workspace via shift-tab as described.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #8970: Shift tabbing out of the toolbox to outside the workspace/toolbox/flyout does not close the flyout
### Check for duplicates

### Description

Shift tabbing out of the toolbox to outside the workspace/toolbox/flyout does not close the flyout. I think it should.
Ben: should be easy to fix by updating Toolbox's onTreeBlur behavior.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #8978: Tabbing to a freshly opened flyout doesn’t automatically select the first block
### Check for duplicates

### Description

Tabbing to a freshly opened flyout doesn’t automatically select the first block.
Not sure if we want to be consistent with the workspace and automatically select the first block. That’s the current demo’s behaviour.
Ben: this should be easy to fix with a change to Workspace's getRestoredFocusableNode or via Flyout function.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #8965
Fixes #8978
Fixes #8970
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/523
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/547
Fixes part of #8910

### Proposed Changes

Fives groups of changes are included in this PR:
1. Support for automatic tab index management for focusable trees.
2. Support for automatic tab index management for focusable nodes.
3. Support for automatically hiding the flyout when back navigating from the toolbox.
4. A fix for `FocusManager` losing DOM syncing that was introduced in #9082.
5. Some cleanups for flyout and some tests for previous behavior changes to `FocusManager`.

### Reason for Changes

Infrastructure changes reasoning:
- Automatically managing tab indexes for both focusable trees and roots can largely reduce the difficulty of providing focusable nodes/trees and generally interacting with `FocusManager`. This facilitates a more automated navigation experience.
- The fix for losing DOM syncing is possibly not reliable, but there are at least now tests to cover for it. This may be a case where a `try{} finally{}` could be warranted, but the code will stay as-is unless requested otherwise.

`Flyout` changes:
- `Flyout` no longer needs to be a focusable tree, but removing that would be an API breakage. Instead, it throws for most of the normal tree/node calls as it should no longer be used as such. Instead, its workspace has been made top-level tabbable (in addition to the  main workspace) which solves the extra tab stop issues and general confusing inconsistencies between the flyout, toolbox, and workspace.
- `Flyout` now correctly auto-selects the first block (#9103 notwithstanding). Technically it did before, however the extra `Flyout` tabstop before its workspace caused the inconsistency (since focusing the `Flyout` itself did not auto-select, only selecting its workspace did).

Important caveats:
- `getAttribute` is used in place of directly fetching `.tabIndex` since the latter can apparently default to `-1` (and possibly `0`) in cases when it's not actually set. This is a very surprising behavior that leads to incorrect test results.
- Sometimes tab index still needs to be introduced (such as in cases where native DOM focus is needed, e.g. via `focus()` calls or clicking). This is demonstrated both by updates to `FocusManager`'s tests as well as toolbox's category and separator. This can be slightly tricky to miss as large parts of Blockly now depend on focus to represent their state, so clicking either needs to be managed by Blockly (with corresponding `focusNode` calls) or automatic (with a tab index defined for the element that can be clicked, or which has a child that can be clicked).

Note that nearly all elements used for testing focus in the test `index.html` page have had their tab indexes removed to lean on `FocusManager`'s automatic tab management (though as mentioned above there is still some manual tab index management required for `focus()`-specific tests).

### Test Coverage

New tests were added for all of the updated behaviors to `FocusManager`, including a new need to explicitly provide (and reset) tab indexes for all `focus()`-esque tests. This also includes adding new tests for some behaviors introduced in past PRs (a la #8910).

Note that all of the new and affected conditionals in `FocusManager` have been verified as having at least 1 test that breaks when it's removed (inverted conditions weren't thoroughly tested, but it's expected that they should also be well covered now).

Additional tests to cover the actual navigation flows will be added to the keyboard experimentation plugin repository as part of https://github.com/google/blockly-keyboard-experimentation/pull/557 (this PR needs to be merged first).

For manual testing, I mainly verified keyboard navigation with some cursory mouse & click testing in the simple playground. @rachel-fenichel also performed more thorough mouse & click testing (that yielded an actual issue that was fixed--see discussion below).

The core webdriver tests have been verified to have seemingly the same existing failures with and without these changes.

All of the following new keyboard navigation plugin tests have been verified as failing without the fixes introduced in this branch (and passing with them):
- `Tab navigating to flyout should auto-select first block`
- `Keyboard nav to different toolbox category should auto-select first block`
- `Keyboard nav to different toolbox category and block should select different block`
- `Tab navigate away from toolbox restores focus to initial element`
- `Tab navigate away from toolbox closes flyout`
- `Tab navigate away from flyout to toolbox and away closes flyout`
- `Tabbing to the workspace after selecting flyout block should close the flyout`
- `Tabbing to the workspace after selecting flyout block via workspace toolbox shortcut should close the flyout`
- `Tabbing back from workspace should reopen the flyout`
- `Navigation position in workspace should be retained when tabbing to flyout and back`
- `Clicking outside Blockly with focused toolbox closes the flyout`
- `Clicking outside Blockly with focused flyout closes the flyout`
- `Clicking on toolbox category focuses it and opens flyout`

### Documentation

No documentation changes are needed beyond the code doc changes included in the PR.

### Additional Information

An additional PR will be introduced for the keyboard experimentation plugin repository to add tests there (see test coverage above). This description will be updated with a link to that PR once it exists.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -22,7 +22,6 @@ import {FlyoutItem} from './flyout_item.js';
 import {FlyoutMetricsManager} from './flyout_metrics_manager.js';
 import {FlyoutNavigator} from './flyout_navigator.js';
 import {FlyoutSeparator, SeparatorAxis} from './flyout_separator.js';
-import {getFocusManager} from './focus_manager.js';
 import {IAutoHideable} from './interfaces/i_autohideable.js';
 import type {IFlyout} from './interfaces/i_flyout.js';
 import type {IFlyoutInflater} from './interfaces/i_flyout_inflater.js';
@@ -323,8 +322,6 @@ export abstract class Flyout
       .getThemeManager()
       .subscribe(this.svgBackground_, 'flyoutOpacity', 'fill-opacity');
 
-    getFocusManager().registerTree(this, true);
-
     return this.svgGroup_;
   }
 
@@ -406,7 +403,6 @@ export abstract class Flyout
     if (this.svgGroup_) {
       dom.removeNode(this.svgGroup_);
     }
-    getFocusManager().unregisterTree(this);
   }
 
   /**
@@ -972,13 +968,12 @@ export abstract class Flyout
 
   /** See IFocusableNode.getFocusableElement. */
   getFocusableElement(): HTMLElement | SVGElement {
-    if (!this.svgGroup_) throw new Error('Flyout DOM is not yet created.');
-    return this.svgGroup_;
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableNode.getFocusableTree. */
   getFocusableTree(): IFocusableTree {
-    return this;
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableNode.onNodeFocus. */
@@ -989,31 +984,29 @@ export abstract class Flyout
 
   /** See IFocusableNode.canBeFocused. */
   canBeFocused(): boolean {
-    return true;
+    return false;
   }
 
   /** See IFocusableTree.getRootFocusableNode. */
   getRootFocusableNode(): IFocusableNode {
-    return this;
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableTree.getRestoredFocusableNode. */
   getRestoredFocusableNode(
     _previousNode: IFocusableNode | null,
   ): IFocusableNode | null {
-    return null;
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableTree.getNestedTrees. */
   getNestedTrees(): Array<IFocusableTree> {
-    return [this.workspace_];
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableTree.lookUpFocusableNode. */
   lookUpFocusableNode(_id: string): IFocusableNode | null {
-    // No focusable node needs to be returned since the flyout's subtree is a
-    // workspace that will manage its own focusable state.
-    return null;
+    throw new Error('Flyouts are not directly focusable.');
   }
 
   /** See IFocusableTree.onTreeFocus. */
@@ -1023,14 +1016,7 @@ export abstract class Flyout
   ): void {}
 
   /** See IFocusableTree.onTreeBlur. */
-  onTreeBlur(nextTree: IFocusableTree | null): void {
-    const toolbox = this.targetWorkspace.getToolbox();
-    // If focus is moving to either the toolbox or the flyout's workspace, do
-    // not close the flyout. For anything else, do close it since the flyout is
-    // no longer focused.
-    if (toolbox && nextTree === toolbox) return;
-    if (nextTree === this.workspace_) return;
-    if (toolbox) toolbox.clearSelection();
-    this.autoHide(false);
+  onTreeBlur(_nextTree: IFocusableTree | null): void {
+    throw new Error('Flyouts are not directly focusable.');
   }
 }
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -966,12 +966,20 @@ export abstract class Flyout
     return null;
   }
 
-  /** See IFocusableNode.getFocusableElement. */
+  /**
+   * See IFocusableNode.getFocusableElement.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   getFocusableElement(): HTMLElement | SVGElement {
     throw new Error('Flyouts are not directly focusable.');
   }
 
-  /** See IFocusableNode.getFocusableTree. */
+  /**
+   * See IFocusableNode.getFocusableTree.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   getFocusableTree(): IFocusableTree {
     throw new Error('Flyouts are not directly focusable.');
   }
@@ -987,24 +995,40 @@ export abstract class Flyout
     return false;
   }
 
-  /** See IFocusableTree.getRootFocusableNode. */
+  /**
+   * See IFocusableNode.getRootFocusableNode.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   getRootFocusableNode(): IFocusableNode {
     throw new Error('Flyouts are not directly focusable.');
   }
 
-  /** See IFocusableTree.getRestoredFocusableNode. */
+  /**
+   * See IFocusableNode.getRestoredFocusableNode.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   getRestoredFocusableNode(
     _previousNode: IFocusableNode | null,
   ): IFocusableNode | null {
     throw new Error('Flyouts are not directly focusable.');
   }
 
-  /** See IFocusableTree.getNestedTrees. */
+  /**
+   * See IFocusableNode.getNestedTrees.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   getNestedTrees(): Array<IFocusableTree> {
     throw new Error('Flyouts are not directly focusable.');
   }
 
-  /** See IFocusableTree.lookUpFocusableNode. */
+  /**
+   * See IFocusableNode.lookUpFocusableNode.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   lookUpFocusableNode(_id: string): IFocusableNode | null {
     throw new Error('Flyouts are not directly focusable.');
   }
@@ -1015,7 +1039,11 @@ export abstract class Flyout
     _previousTree: IFocusableTree | null,
   ): void {}
 
-  /** See IFocusableTree.onTreeBlur. */
+  /**
+   * See IFocusableNode.onTreeBlur.
+   *
+   * @deprecated v12: Use the Flyout's workspace for focus operations, instead.
+   */
   onTreeBlur(_nextTree: IFocusableTree | null): void {
     throw new Error('Flyouts are not directly focusable.');
   }
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -308,7 +308,6 @@ export abstract class Flyout
     // hide/show code will set up proper visibility and size later.
     this.svgGroup_ = dom.createSvgElement(tagName, {
       'class': 'blocklyFlyout',
-      'tabindex': '0',
     });
     this.svgGroup_.style.display = 'none';
     this.svgBackground_ = dom.createSvgElement(
@@ -324,7 +323,7 @@ export abstract class Flyout
       .getThemeManager()
       .subscribe(this.svgBackground_, 'flyoutOpacity', 'fill-opacity');
 
-    getFocusManager().registerTree(this);
+    getFocusManager().registerTree(this, true);
 
     return this.svgGroup_;
   }
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -17,7 +17,18 @@ import {FocusableTreeTraverser} from './utils/focusable_tree_traverser.js';
  */
 export type ReturnEphemeralFocus = () => void;
 
+/**
+ * Represents an IFocusableTree that has been registered for focus management in
+ * FocusManager.
+ */
 class TreeRegistration {
+  /**
+   * Constructs a new TreeRegistration.
+   *
+   * @param tree The tree being registered.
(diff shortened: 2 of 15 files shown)


rachel-fenichel:
@BenHenning what are the next steps for this to be mergeable?

BenHenning:
> @BenHenning what are the next steps for this to be mergeable?

@rachel-fenichel it can probably go into review now (and I will do that later today). I want to spend a bit of time seeing if I can get automatically managed node tab indexes working before fully giving up on that.

Edit: Actually there is one bug I'm re-remembering now that I'm testing this locally that still needs to be fixed yet (sometimes the flyout doesn't auto-collapse).

BenHenning:
The following scenario breaks still:
- Tab-navigate to the workspace
- Press 'T' to open the flyout (via toolbox focus)
- Press 'right' to navigate to the flyout
- Press 'tab'
- Observe that there's a weird in-between state with the flyout still focused but the workspace having an outline (local debugging confirms that the workspace svg holds DOM focus but the `FocusManager` still thinks the flyout block should hold active focus)

This is related to https://github.com/google/blockly-keyboard-experimentation/issues/547.

BenHenning:
Uh oh. `FocusManager` can get stuck in a state updating the focused node which basically disables DOM syncing. 🙃 This could cause all sorts of interesting issues. I'll fix it in this PR alongside the main issue.

BenHenning:
NB: This should be fully working now, but tests & some clean-up need to come yet.

I also think that I'm going to be a bit stricter on tab management in `FocusManager`. Right now trees need to make themselves focusable but non-tabbable if they disable auto management, but nodes no longer need to. This inconsistency feels weird and a bit bad--it's preferable just to have `FocusManager` be entirely responsible for managing tabs, I think.

Edit: Actually, I just realized that `focusNode` already automatically solves this, so setting a tree's root to a `-1` tab index is simply redundant. Cool.

BenHenning:
Decided to include a fix for #8970 since I was in the neighborhood.

BenHenning:
From @rachel-fenichel's testing:

> Interacting with the workspace and then clicking on a toolbox category opens the first category instead of the one I clicked

core/toolbox/category.ts
```diff
@@ -225,7 +225,6 @@ export class ToolboxCategory
    */
   protected createContainer_(): HTMLDivElement {
     const container = document.createElement('div');
-    container.tabIndex = -1;
```
BenHenning, line 228: This is the cause of the bug @rachel-fenichel found.

The problem comes from the fact that there's no automatic focusing when clicking on categories now, whereas there was before. This results in there not being a focused node upon click, so toolbox defaults to its first category when it receives focus (since it has no child receive focus).

It's actually correct to put this line back as the fix. I'll add a line comment here & for separator to try and make it extra clear why this is needed.

BenHenning: Fixed in latest.

core/toolbox/category.ts
```diff
@@ -225,6 +225,8 @@ export class ToolboxCategory
    */
   protected createContainer_(): HTMLDivElement {
     const container = document.createElement('div');
+    // Ensure that the category has a tab index to ensure it receives focus when
+    // clicked (since clicking isn't managed by the toolbox).
     container.tabIndex = -1;
```
BenHenning, line 230: Maybe consider adding a test for this & the separator cases in core.

BenHenning (commented):
Self-reviewed (mostly a spot check for anything obvious).

BenHenning:
Things left to do in/for this PR:
- [x] Add the tests mentioned above for toolbox. Added 1 new test in https://github.com/google/blockly-keyboard-experimentation/pull/557.
- [x] Fix behaviors to get the keyboard nav tests to pass. Fix was simple: need to use `focusNode()` now instead of `.focus()`.
- [x] Add new keyboard nav tests. Done in https://github.com/google/blockly-keyboard-experimentation/pull/557.
- [x] Verify core webdriver tests with these changes. This is tricky since they're still failing at tip-of-tree (I see 28-29 pass on each run without these changes). The results appear the same with these changes. Unfortunately some of the failing tests are related to toolbox categories so we can't test those effectively.
- [x] ~Add video demonstrations of each of the linked bugs now being fixed.~ (skipping since the behavior can be observed via the tests).
- [x] Fix that clicking out of the injection div doesn't auto-close the `Flyout` when it's focused (but does work fine when toolbox is focused). Found by @microbit-robert. Fixed this without too much trouble, but still need to add a test for it. Tests have been added to keyboard nav and are now tracked as part of that item above.
- [x] Address feedback & get back into review (after everything else is done).

BenHenning:
I think this can be reviewed now, though please note there will need to probably be 1 more pass once all of the above items are addressed.

core/flyout_base.ts
```diff
@@ -990,31 +984,29 @@ export abstract class Flyout
...
   /** See IFocusableNode.canBeFocused. */
   canBeFocused(): boolean {
-    return true;
+    return false;
   }
 
   /** See IFocusableTree.getRootFocusableNode. */
   getRootFocusableNode(): IFocusableNode {
-    return this;
+    throw new Error('Flyouts are not directly focusable.');
```
maribethb, line 1004: Can you add a deprecation warning to these so we can remove them in v13?

core/focus_manager.ts
```diff
@@ -147,7 +185,16 @@ export class FocusManager {
...
-    return this.registeredTrees.findIndex((reg) => reg === tree) !== -1;
+    return !!this.lookUpRegistration(tree);
+  }
+
+  /**
+   * Returns the TreeRegistration for the specified tree, or null if the tree is
+   * not currently registered.
+   */
+  private lookUpRegistration(tree: IFocusableTree): TreeRegistration | null {
+    const index = this.registeredTrees.findIndex((reg) => reg.tree === tree);
```
maribethb: Why are you using `findIndex` instead of `find` since you're returning the element anyway, not the index?

Though I'm wondering if this should be a map of registered trees to IDs and not an array. Does the order in which trees are registered matter?

core/focus_manager.ts
```diff
@@ -240,11 +300,15 @@ export class FocusManager {
...
    *
+   * Note that this may update the specified node's element's tabindex to ensure
+   * that it can be properly read out by screenreaders while focused.
+   *
    * @param focusableNode The node that should receive active focus.
    */
   focusNode(focusableNode: IFocusableNode): void {
     this.ensureManagerIsUnlocked();
-    if (!this.currentlyHoldsEphemeralFocus) {
+    const mustRestoreUpdatingNodeFlag = !this.currentlyHoldsEphemeralFocus;
```
maribethb: nit: don't use the word `flag` in the name, we already know it's a boolean

core/focus_manager.ts
```diff
@@ -448,14 +521,29 @@ export class FocusManager {
...
+      if (treeIsTabManaged) {
+        // If this node's tree has its tab auto-managed, ensure that it's no
+        // longer tabbable now that it holds active focus.
+        tree.getRootFocusableNode().getFocusableElement().tabIndex = -1;
+      }
     }
     node.onNodeFocus();
     this.lockFocusStateChanges = false;
 
+    // Only overwrite the tab index if it isn't a tree root that's auto managed.
```
maribethb: I find this confusing, could you explain it and consider rewriting the comment?

You say you're going to overwrite the tab index, but you only set it if there wasn't a tab index attribute to start with. It's hard for me to call that overwriting. I think you have to add this because you can't focus an element that doesn't have a tabindex, and you removed the tabindex from a bunch of the html for core blockly elements like fields. Is that right? If so, that would be helpful context to put in the comment instead of saying "overwrite"


On a separate confusion, it would help if you explained why you don't set it if it *is* a tree root that's auto managed. Is the only reason because you already set it to -1 above?

BenHenning:
NB: The plan to fix #9105 is no longer part of this PR. It's not blocking MakeCode (since they use their own flyout implementation), and the fix is actually likely to be a bit tricky with the way we automatically open/close the flyout based on focus.

core/toolbox/category.ts
BenHenning: Decided to add a test in kbe instead as it was much easier (ran into a lot of problems with click/focus in toolbox's test unfortunately). Didn't bother testing separator as it seems less important.

core/flyout_base.ts
BenHenning: Done in latest.

core/focus_manager.ts
BenHenning: No reason not to use `find` honestly--updated.

As for order, it doesn't matter\* and the size of the array will be quite small (there aren't many trees). I'm also not keen on relying on the stability of the tree's root's element ID for registration correctness (though it should technically be reliable per contracts).

\* The order can theoretically matter, but only if trees are violating their contracts.

core/focus_manager.ts
BenHenning: Ah nice catch--surprised I did that. Fixed!

core/focus_manager.ts
BenHenning: This is, admittedly, a very confusing comment. :)


> ... I think you have to add this because you can't focus an element that doesn't have a tabindex, and you removed the tabindex from a bunch of the html for core blockly elements like fields. Is that right? If so, that would be helpful context to put in the comment instead of saying "overwrite"

This is correct. It's documented in `IFocusableNode` (where I think it's much more important to explain), but I completely agree that 'overwrite' is the wrong terminology here and the context is important. I've tried to clarify the different cases that this bit affects, and why.

> On a separate confusion, it would help if you explained why you don't set it if it _is_ a tree root that's auto managed. Is the only reason because you already set it to -1 above?

Yep, this is the reason, and I've attempted to also clarify it in the latest changes.

Hopefully this is much more clear now.

BenHenning (commented):
Self-reviewed all the latest changes.

BenHenning:
With https://github.com/google/blockly-keyboard-experimentation/pull/557 in place and all the tasks listed above addressed (including reviewer comments), I think this is ready again for review. Fortunately, there weren't many changes needed here.

PTAL @maribethb.

maribethb (approved):

BenHenning:
Thanks @maribethb for the review, and @rachel-fenichel for helping with testing!

Issue #9081, opened 2025-05-21
`FocusManager` can get out of sync with the DOM with Firefox and Safari


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

https://github.com/google/blockly-keyboard-experimentation/issues/87 has much more context, but essentially it's possible for `FocusManager` to desync with the DOM's focus specifically in Firefox and Safari in cases when a focused (possibly any) element is moved to a new location in the DOM. Chrome will fire a focus change event in these cases, but Firefox/Safari do not (or at least not in the same scenarios).

### Reproduction steps

See https://github.com/google/blockly-keyboard-experimentation/issues/87 for repro steps, though the fixed introduced in #9082 has to be temporarily removed in order to observe the problem.

### Stack trace

None

### Browsers

Firefox desktop, Safari desktop

maribethb:
@BenHenning Do you have any further thoughts on this issue? Is #9082 sufficient to close it? It feels like this is a bug in Firefox/Safari (that no focus lost event is fired) and I'm not sure what we could do on our end to work around it beyond what you've already done in 9082

BenHenning:
@maribethb I was originally envisioning that we could do some sort of broader focus and DOM tracking to synthesize our own focus events in a cross-browser manner (to account for things like lost DOM elements correctly). That would simplify what `FocusManager` needs to do, but in itself would be complex (particularly doing it in a way that doesn't introduce large performance issues).

If you're happy with the slightly hacky solution from #9082 for the long run then I agree this can be closed.

PR #9082, opened 2025-05-21
fix: Update `focusNode` to self correct focus

### LINKED ISSUE #87: firefox can't focus code editor
The current code (13bc49401) had some last minute hacks to (a) remove an extraneous tab stop before the editor and (b) change the focus indicator for tabbing to the editor. I presume (a) is buggy (i.e. 3100903e6) because I can't tab to the code editor on Firefox. I'm not sure what the behaviour was before those 2 commits. 

Here's a [30 second loom video](https://www.loom.com/share/229d67cc07ad4aea8babea1d9a11c05e) comparing the two behaviours

Firefox 132.0 (64-bit) on Windows 11.

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/87

### Proposed Changes

This updates `FocusManager.focusNode()` to automatically defocus its internal state if it detects that DOM focus (per `document.activeElement`) doesn't match its own internal focus.

It also updates `FocusManager` to avoid duplicate self calls to `focusNode()`.

### Reason for Changes

This is a robustness improvement for `focusNode` that is nice to keep as a "if all else fails" mechanism, but it's currently a hacky workaround to https://github.com/google/blockly-keyboard-experimentation/issues/87. #9081 is tracking introducing a long-term fix for the desynchronizing problem, but that's likely to be potentially much harder to solve and this at least introduces a reasonable correction.

From a stability perspective, it seems likely that there are multiple classes of failures covered by this fix. Essentially the browser behavior difference in Firefox and Safari over Chrome is that the former do not fire a focus change event when a focused element is removed from the DOM (leading to `FocusManager` getting out of sync). There may be other such cases when a focus event isn't fired, so this robustness improvement at least ensures eventual consistency so long as `focusNode()` is called (and, fortunately, that's done a lot now).

While this is a nice robustness improvement, it's not a perfect replacement for a real fix. For the time between `FocusManager` getting out of sync and `focusNode` getting called, `getFocusedNode` will _not_ match the actual element holding focus. The primary class of issues known is when a DOM element is being moved, and in these cases `focusNode` _is_ called. If there are other such unknown cases where a desync can happen, **`getFocusedNode()` will remain wrong until a later `focusNode()` call**.

Note one other change: originally implemented but removed in earlier PRs for `FocusManager`, this change also includes ensuring `focusNode()` isn't called multiple times for a single request to focus a node. Current logic results in a call to `focusNode()` calling a node's `focus()` which then processes a second call to `focusNode()` (which is fully executed because `FocusManager.focusedNode` isn't updated until after the call to `focus()`). This doesn't actually correct any state, but it's more efficient and provides some resilience against potential logic issues from calling node/tree callbacks multiple times (which was observed up to 3 times in some cases).

### Test Coverage

This has been tested via the keyboard navigation experimental plugin's test environment (with Firefox), plus new unit tests. Note the new test for directly verifying desyncing logic is contrived, but it should be perfectly testing the exact scenario that's being observed on Firefox/Safari. A separate test was added for the existing behavior of focusing a different node still correcting `FocusManager` state even if it was desynced (the bug only happens for the same node being refocused).

New tests were also added for the various lifecycle callbacks (to ensure they aren't called multiple times).

All of the new tests were verified to fail without the two fixes in place (they were verified in isolation), minus the test for focusing a second node when desynced (since that should pass regardless of the new fixes).

Some basic simple playground testing was done, as well, just to verify nothing obvious was broken around selection, gestures, and copy/paste.

### Documentation

No new documentation should be needed here.

### Additional Information

This wasn't explicitly tested in Safari since I only have access to Chrome and Firefox, but I will ask someone else on the team to verify this for me after merging if it isn't checked sooner.

### DIFF
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -240,6 +240,15 @@ export class FocusManager {
    */
   focusNode(focusableNode: IFocusableNode): void {
     this.ensureManagerIsUnlocked();
+    // Double check that state wasn't desynchronized in the background. Context:
+    // https://github.com/google/blockly-keyboard-experimentation/issues/87.
+    const currentFocusedElement = this.focusedNode?.getFocusableElement();
+    if (this.focusedNode && currentFocusedElement !== document.activeElement) {
+      // Ensure the node has actually been defocused since the DOM defocused it.
+      // In some cases this may result in an immediate re-focus, but that's fine
+      // since this logic at least guarantees eventual consistency.
+      this.defocusCurrentFocusedNode();
+    }
     if (this.focusedNode === focusableNode) return; // State is unchanged.
     if (!focusableNode.canBeFocused()) {
       // This node can't be focused.
@@ -292,6 +301,12 @@ export class FocusManager {
       this.activelyFocusNode(nodeToFocus, prevTree ?? null);
     }
     this.updateFocusedNode(nodeToFocus);
+    if (!this.currentlyHoldsEphemeralFocus) {
+      // Focus the actual element. This is done last since it can create a loop
+      // back to focusNode() (which will be short-circuited by the early exit
+      // logic above).
+      nodeToFocus.getFocusableElement().focus();
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


BenHenning (commented):
Self-reviewed changes.

maribethb (approved):

PR #9085, opened 2025-05-22
fix: Fix bug that prevented using keyboard shortcuts when the DropDownDiv is open.

### LINKED ISSUE #522: Escape no longer reliably closes field editors
In local builds of the demo, it seems to work only for the first colour picker and never for dropdowns.

The [pre-focus-changes version currently deployed](https://google.github.io/blockly-keyboard-experimentation/) works reliably in this respect.

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/522

### Proposed Changes
This PR moves the attachment of the global keyboard shortcut handler to the `DropDownDiv` and `WidgetDiv` into those respective modules, rather than adding the listener in `inject()`. This reduces spooky action at a distance, and fixes a bug caused by a change to the `DropDownDiv` to make it recreate its div between showings in order to guarantee that its state is fully reset. Because the key handler was attached elsewhere, this being unsafe wasn't noticed. Now, the listener is added every time the div is created, and the logic to attach it has been moved into the relevant modules themselves to hopefully avoid being overlooked in a similar manner in the future.

### DIFF
--- a/core/common.ts
+++ b/core/common.ts
@@ -8,11 +8,13 @@
 
 import type {Block} from './block.js';
 import {BlockDefinition, Blocks} from './blocks.js';
+import * as browserEvents from './browser_events.js';
 import type {Connection} from './connection.js';
 import {EventType} from './events/type.js';
 import * as eventUtils from './events/utils.js';
 import {getFocusManager} from './focus_manager.js';
 import {ISelectable, isSelectable} from './interfaces/i_selectable.js';
+import {ShortcutRegistry} from './shortcut_registry.js';
 import type {Workspace} from './workspace.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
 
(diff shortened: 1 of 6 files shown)


core/common.ts
```diff
@@ -310,4 +312,31 @@ export function defineBlocks(blocks: {[key: string]: BlockDefinition}) {
...
 }
 
+/**
+ * Handle a key-down on SVG drawing surface. Does nothing if the main workspace
+ * is not visible.
+ *
+ * @internal
+ * @param e Key down event.
+ */
+// TODO (https://github.com/google/blockly/issues/1998) handle cases where there
```
BenHenning: This appears out of date. Is it still relevant? If so, perhaps we should file a new issue and update this accordingly.

core/widgetdiv.ts
```diff
@@ -69,11 +70,19 @@ export function createDom() {
   if (document.querySelector('.' + containerClassName)) {
     containerDiv = document.querySelector('.' + containerClassName);
   } else {
-    containerDiv = document.createElement('div') as HTMLDivElement;
+    containerDiv = document.createElement('div');
     containerDiv.className = containerClassName;
   }
+  if (!containerDiv) return;
```
BenHenning: Is this case actually possible or just TS not knowing that it isn't? The code above seems to guarantee the div will exist, so perhaps the code could be rearranged to ensure that TS is aware of the value being present in order to obviate this check.

BenHenning (commented):
Thanks @gonfunko! The fix makes a lot of sense, and does seem to improve robustness overall.

Is it possible to test the failure cases for both divs and verify that the fix addresses the issue? I'm not sure how difficult that would be, but this seems like a thing that's very subtle and could potentially break again without some sort of regression check.

core/common.ts
gonfunko: Removed.

core/widgetdiv.ts
gonfunko: Mostly TS not knowing; refactored so that it does.

BenHenning (approved):
Thanks @gonfunko! LGTM.

PR #9086, opened 2025-05-22
feat: Add more ephemeral overrides for drop-downs.

### LINKED ISSUE #9078: Add manageEphemeralFocus parameter to dropdown show functions
### Check for duplicates

### Description

For field editors where MakeCode uses both the widget div and the dropdown div in combination, [this ephemeral focus error](https://github.com/google/blockly/blob/develop/core/focus_manager.ts#L320-L323) is raised. Unlike the [Blockly samples field slider plugin](https://github.com/google/blockly-samples/blob/master/plugins/field-slider/src/field_slider.ts#L102-L107), in MakeCode, focus is moved to the HTML input, not the dropdown, so the ephemeral focus handling needs to be on the widget, not the dropdown div.

Please can you add an extra parameter for manageEphemeralFocus to [showPositionedByField](https://github.com/google/blockly/blob/135da402ef0035ce5a44d4a5d5d73d576e92bc82/core/dropdowndiv.ts#L222) and [showPositionedByBlock](https://github.com/google/blockly/blob/135da402ef0035ce5a44d4a5d5d73d576e92bc82/core/dropdowndiv.ts#L197), which can then be passed to [showPositionedByRect](https://github.com/google/blockly/blob/135da402ef0035ce5a44d4a5d5d73d576e92bc82/core/dropdowndiv.ts#L278). At the moment, manageEphemeralFocus defaults to true and no callers pass this arg in core.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

Example of the type of block affected, note the focus on the HTML input on opening:
Image

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9078
Fixes part of #8915 (new tests)

### Proposed Changes

Exposes the ability to disable ephemeral focus management for drop-down divs that are shown using `showPositionedByBlock` or `showPositionedByField`. Previously, this was only supported via `show`, but the former methods are also used externally.

This allows the underlying issue reported by #9078 to be fixed downstream for cases when both the widget and drop-down divs are opened simultaneously.

This PR also introduces tab indexes for both widget and drop-down divs (which were noticed missing when adding tests). This is because, currently, taking ephemeral focus on for a node that doesn't have a tab index will do nothing. This fix is useful for future screen reader work, and doesn't have obvious impacts on existing core or keyboard navigation behaviors (per testing and reasoning).

### Reason for Changes

Exposing the ability to disable ephemeral focus management for all public API entrypoints for showing the divs is crucial for providing the maximum flexibility when downstream apps use both the widget and drop-down divs together. This should ensure that all of these cases can be correctly managed in the same way as https://github.com/google/blockly-samples/pull/2521.

### Test Coverage

This introduces a bunch of new tests that were missing originally for both widget and drop-down div (including specifically verifying ephemeral focus). As part of the drop-down div tests, it also introduces actual positioning logic. This isn't great, but it's somewhat reasonable and robust against page changes (since the actual mocha results can move where the elements will end up on the page).

These changes have also been manually tested with both the core simple playground and the keyboard experiment plugin's test environment with no noticed regressions in either. The plugin's tests have also been run against these changes to ensure no new breakages have been introduced.

### Documentation

No documentation changes beyond the code ones introduced in this PR should be needed.

### Additional Information

The new tests may actually act as a basis for avoiding the test backdoor that's used today for the positioning tests for drop-down div tests. This doesn't replace those existing tests nor does it cover other behaviors and entrypoints that would be worth testing, but testing ephemeral focus is a nice improvement (especially in the context of what this PR is fixing).

### DIFF
--- a/core/dropdowndiv.ts
+++ b/core/dropdowndiv.ts
@@ -122,6 +122,7 @@ export function createDom() {
   }
   div = document.createElement('div');
   div.className = 'blocklyDropDownDiv';
+  div.tabIndex = -1;
   const parentDiv = common.getParentContainer() || document.body;
   parentDiv.appendChild(div);
 
@@ -192,6 +193,11 @@ export function setColour(backgroundColour: string, borderColour: string) {
  * @param block Block to position the drop-down around.
  * @param opt_onHide Optional callback for when the drop-down is hidden.
  * @param opt_secondaryYOffset Optional Y offset for above-block positioning.
+ * @param manageEphemeralFocus Whether ephemeral focus should be managed
+ *     according to the drop-down div's lifetime. Note that if a false value is
+ *     passed in here then callers should manage ephemeral focus directly
+ *     otherwise focus may not properly restore when the widget closes. Defaults
+ *     to true.
  * @returns True if the menu rendered below block; false if above.
  */
 export function showPositionedByBlock<T>(
@@ -199,10 +205,12 @@ export function showPositionedByBlock<T>(
   block: BlockSvg,
   opt_onHide?: () => void,
   opt_secondaryYOffset?: number,
+  manageEphemeralFocus: boolean = true,
 ): boolean {
   return showPositionedByRect(
     getScaledBboxOfBlock(block),
     field as Field,
+    manageEphemeralFocus,
     opt_onHide,
     opt_secondaryYOffset,
   );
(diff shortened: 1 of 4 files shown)


BenHenning (commented):
Self-review (spot check). I could add some comments for the tab indexes, but I won't bother unless other changes end up being needed.

rachel-fenichel (approved):

PR #9092, opened 2025-05-24
fix: Skip invisible inputs in the field navigation policy

### LINKED ISSUE #524: Navigation visits non-visible inputs
I think we had this working as of https://github.com/google/blockly/pull/8948 but since the move away from ASTNode the navigation sequence for this block moving right is the two visible fields, an invisible position, and then the mutator is skipped. 

As in many MakeCode blocks, the mutator shows or hides an input between the last field and the mutator.

https://github.com/user-attachments/assets/35643274-1af1-4f54-aff6-e01ba3a9fb1b

It looks like there might be two problems:
- we're back to visiting invisible inputs
- separately, it looks like we no longer visit mutators (you can see this in blocks with no hidden inputs like MakeCode's if)

RoboErikG:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/524
image

### Proposed Changes

Skips over hidden inputs when finding the next/prev node in field_navigation_policy.

### Reason for Changes

If you had a field before or after an input that had visible set to false the hidden input could be navigated to.

### Test Coverage

Added tests to cover next and previous for this case.

### DIFF
--- a/core/keyboard_nav/field_navigation_policy.ts
+++ b/core/keyboard_nav/field_navigation_policy.ts
@@ -48,15 +48,14 @@ export class FieldNavigationPolicy implements INavigationPolicy<Field<any>> {
     let fieldIdx = input.fieldRow.indexOf(current) + 1;
     for (let i = curIdx; i < block.inputList.length; i++) {
       const newInput = block.inputList[i];
-      if (!newInput.isVisible()) {
-        continue;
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@cpcallen please review in your AM tomorrow so we can merge this tomorrow.

cpcallen (approved):

Issue #9098, opened 2025-05-27
Improve cut/copy/paste flexibility


cpcallen:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

At the moment the cut and copy shortcuts are fairly inflexible.  In order to cut or copy a focusable, it must:

1.  Be an `ICopyable`, so that it is possible to get it's copy data.
2.  Be an`IDraggable` and `IDeletable`.  This is reasonable as we want to ensure that any pasted copies can be manipulated/deleted.
3.  Return `true` from both `.isMovable` and `.isDeletable`, for the same reason.

These criteria are currently hard-coded into the `preconditionFn` of the cut and copy shortcuts.

Criteria 3 is necessary because the copy will get the same `.movable`/`.deletable` bits as the source item, so if the source isn't currently movable/deletable the pasted copy won't be either.

Relatedly, because [shadow blocks' `.isMovable`](https://github.com/google/blockly/blob/cc9384ae87aeb25c47c158261fe1dc8d381240a4/core/block.ts#L827) and [`.isDeletable`](https://github.com/google/blockly/blob/cc9384ae87aeb25c47c158261fe1dc8d381240a4/core/block.ts#L794) always return false, making shadows non-copyable.

We know that developers would like to allow their users to copy shadow blocks (see e.g.  #9084).  It's possible to do so via a kludgy workaround in the shortcuts themselves, but some relatively small changes to the Blockly API allow a more general approach would be beneficial not only for this scenario but for others as well.  For example:

* It should be possible to have fixed starting blocks that can be copied (e.g. [Blockly Games's Maze's initial "move forward" block](https://blockly.games/maze?lang=en)).
* It might be desirable that pasted copies of disabled blocks be enabled.
* It would be useful if it were possible for custom copyable items to have other attributes that differ on the copy than on the original—e.g., a unique identifier or, in a multi-user scenario, a created-by-user attribute.

### Request

I propose the following changes:

* Add an `.isCopyable()` method to `ICopyable`.
  * To avoid this being a breaking change, the method could be made an optional member of the interface in v12, possibly being promoted to non-optional in a future major version.  The cut/copy shortcuts would fall back to the current criteria for any `ICopyable` not implementing this new method.
* Provide implementations of `isCopyable` on `Block` and `WorkspaceComment`:
  * For backwards compatibility, the default implementation should mostly  replicate the current copyability criteria, so that e.g. fixed start blocks do not become copyable unexpectedly.
[...]

Issue #9084, opened 2025-02-18
Cannot copy a shadow block


rachel-fenichel:
> Thanks for the explanation and demo, that makes a lot of sense. I can't help but favour the plugin's behaviour of turning a shadow block into a real block on edit - since that shows intent from the user to make something concrete. Are there situations or applications in which this wouldn't make sense, or previous discussions I can refer back to learn about the competing needs?
> 
> There are downstream consequences of this behaviour on copy and paste, which don't make up with my expectations. I would have assumed you'd be able to copy and paste the contents of an oval, like the text in a "write with shadow" block
> 
> ![Screenshot. Write with shadow (item variable). Write with shadow "how are you". The block containing "how are you" is selected. ](https://github.com/user-attachments/assets/9a374b70-2e66-4354-b0e1-c55b140a9289)
> 
> I would naively argue that these blocks should be real, and draggable, and copyable. If they are not, then we definitely need some user feedback when they try to copy a block, because currently copy will silently fail and you'll end up surprised by pasting the last thing in the copy buffer.  

 _Originally posted by @kmcnaught in [#132](https://github.com/google/blockly-keyboard-experimentation/issues/132#issuecomment-2662820703)_

maribethb:
Note that with google/blockly-keyboard-experimentation#501 this behavior should be implemented in core

rachel-fenichel:
"Should be" meaning new code needs to be added in core, or meaning that the only remaining work is verification?

maribethb:
needs to be added in core

cpcallen:
I've put together [a working proof-of-concept of a change a developer could make to enable copying shadow blocks](https://github.com/cpcallen/blockly/commit/5ac1b850a1c24b0bda72a0a32a084b0375d2c3d9).

I've also filed #9098 to track a better long-term fix for this issue, and am closing this issue in favour of that one.

cpcallen:
Elaborating on my previous comment with some additional background information:

This issue was discussed by the Blockly team.  We observed:

* Being able to copy shadow blocks is a reasonable feature request that we would like to fulfill.
* There are some obvious parallel situations where it might be desirable for non-shadow blocks that are not currently copyable (because they are not movable and/or not deletable) to be copied and pasted, provided the copies were movable and deletable.
* We do not currently have the needed machinery to paste blocks that are different than the originally-copied block (e.g. to make a copy of a non-movable block movable).
[...]

PR #9107, opened 2025-05-28
fix: Make cut/copy/paste work consistently and as expected

### LINKED ISSUE #546: Can no longer use keyboard actions on blocks in the flyout
### Check for duplicates

### Description

At one point, we were able to use Ctrl + Enter to open the context menu on blocks in the flyout, and Ctrl + C to copy blocks in the flyout (the latter automatically closed the flyout). This is no longer case. This feels inconsistent with mouse users who can still open the context menu via right click. Copy is still not available when opening the context menu via mouse, but perhaps it should be?

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

RoboErikG:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/546

This is a dependency for https://github.com/google/blockly-keyboard-experimentation/pull/556 

### Proposed Changes

Updates the preconditions for cut/copy/paste to be more aware of flyouts and to use the following criteria:
- Allow copying any blocks that could be moved, deleted, and edited once pasted (isOwn<X>able)
- Allow cutting any blocks that are currently movable and deletable (is<X>able).
- Allow pasting onto editable workspaces when there's copydata available.

### Reason for Changes

You couldn't copy a block in the toolbox and then paste it into the workspace and when paste was enabled was inconsistent.

### DIFF
--- a/core/shortcut_items.ts
+++ b/core/shortcut_items.ts
@@ -213,7 +213,9 @@ export function registerCopy() {
         : targetWorkspace;
       if (!targetWorkspace) return false;
 
-      targetWorkspace.hideChaff();
+      if (!focused.workspace.isFlyout) {
+        targetWorkspace.hideChaff();
+      }
       copyData = focused.toCopyData();
       copyWorkspace = targetWorkspace;
       copyCoords =
--- a/core/shortcut_items.ts
+++ b/core/shortcut_items.ts
@@ -130,8 +130,7 @@ function isCopyable(
     isIDeletable(focused) &&
     focused.isOwnDeletable() &&
     isDraggable(focused) &&
-    focused.isOwnMovable() &&
-    focused.isOwnEditable()
+    focused.isOwnMovable()
   );
 }
 
@@ -164,8 +163,7 @@ function isCuttable(focused: IFocusableNode): boolean {
     isIDeletable(focused) &&
     focused.isDeletable() &&
     isDraggable(focused) &&
-    focused.isMovable() &&
-    focused.isEditable()
+    focused.isMovable()
   );
 }
 
--- a/core/shortcut_items.ts
+++ b/core/shortcut_items.ts
@@ -101,7 +101,7 @@ let copyWorkspace: WorkspaceSvg | null = null;
 let copyCoords: Coordinate | null = null;
 
 /**
- * Determine if a focusable node can be copied using cut or copy.
+ * Determine if a focusable node can be copied.
  *
  * Unfortunately the ICopyable interface doesn't include an isCopyable
  * method, so we must use some other criteria to make the decision.
@@ -110,8 +110,8 @@ let copyCoords: Coordinate | null = null;
  * - It must be an ICopyable.
  * - So that a pasted copy can be manipluated and/or disposed of, it
  *   must be both an IDraggable and an IDeletable.
- * - Additionally, both .isMovable() and .isDeletable() must return
- *   true (i.e., can currently be moved and deleted).
+ * - Additionally, both .isOwnMovable() and .isOwnDeletable() must return
+ *   true (i.e., the copy could be moved and deleted).
  *
  * TODO(#9098): Revise these criteria.  The latter criteria prevents
  * shadow blocks from being copied; additionally, there are likely to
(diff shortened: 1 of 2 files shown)


core/shortcut_items.ts
```diff
@@ -124,12 +124,48 @@ let copyCoords: Coordinate | null = null;
...
   focused: IFocusableNode,
 ): focused is ICopyable<ICopyData> & IDeletable & IDraggable {
+  if (!(focused instanceof BlockSvg)) return false;
+  return (
+    isICopyable(focused) &&
+    isIDeletable(focused) &&
+    focused.isOwnDeletable() &&
+    isDraggable(focused) &&
+    focused.isOwnMovable() &&
+    focused.isOwnEditable()
```
maribethb: Is Editable actually required? That seems like a new requirement that didn't exist before. If the block isn't editable (i.e. fields can't be changed) but it is movable and deletable, i think it should still be able to be copied (especially if we didn't have this restriction before). Was this solving a specific problem?

maribethb (commented):
Generally I think this makes sense but it would be nice if there were more test cases specifically covering the kinds of blocks that weren't copyable before this and now are.

lgtm pending question about editable though.

core/shortcut_items.ts
RoboErikG: Whoops. Good catch. I'll remove that check from here and isCuttable.

maribethb:
Sorry, just merged my change so you got the merge conflict, please be sure to add the ephemeralFocusTaken check as well. Thanks!

maribethb (approved):

core/shortcut_items.ts
```diff
@@ -127,6 +127,40 @@ let copyCoords: Coordinate | null = null;
...
+ *
+ * TODO(#9098): Revise these criteria.  The latter criteria prevents
+ * shadow blocks from being copied; additionally, there are likely to
+ * be other circumstances were it is desirable to allow movable /
+ * copyable copies of a currently-unmovable / -copyable block to be
+ * made.
+ *
+ * @param focused The focused object.
+ */
+function isCuttable(focused: IFocusableNode): boolean {
```
cpcallen, line 162: The reason I created `isCopyable`, above, was to ensure that cut and copy used _common_ criteria for deciding if something can be cut/copied, with cut being allowed for anything that can be copied and is additionally deletable—AFAICT there is no circumstance where it should be possible to cut something but _not_ copy it.

So `isCuttable` should call `isCopyable` (not just `isICopyable`), and return false if it does.  Or… we could just have the cut `preconditionFn` just check `isCopyable(focused) && focused.isDeletable()`, as it previously did.

core/shortcut_items.ts
```diff
@@ -127,6 +127,40 @@ let copyCoords: Coordinate | null = null;
...
   focused: IFocusableNode,
 ): focused is ICopyable<ICopyData> & IDeletable & IDraggable {
+  if (!(focused instanceof BlockSvg)) return false;
+  return (
+    isICopyable(focused) &&
+    isIDeletable(focused) &&
+    focused.isOwnDeletable() &&
+    isDraggable(focused) &&
+    focused.isOwnMovable()
+  );
```
cpcallen, line 137: This change makes things that we decided we did not want to make copyable copyable.

This includes shadow blocks, but also blocks that have custom `isDeletable` and `isMovable` methods.

While I was working on #9084 we decided we did not want to do that in v12.

I'm not necessarily averse to making some things copyable (which were not previously copyable), but It is in my opinion a mistake to have done so without introducing an `isCopyable` method on `ICopyable` so that block definition authors can have more fine-grained control.

core/shortcut_items.ts
```diff
@@ -151,29 +185,45 @@ export function registerCopy() {
     name: names.COPY,
     preconditionFn(workspace, scope) {
       const focused = scope.focusedNode;
+      if (!(focused instanceof BlockSvg)) return false;
```
cpcallen, line 188: This prevents anything except blocks—e.g., workspace comments—from being copied.

core/shortcut_items.ts
```diff
@@ -151,29 +185,45 @@ export function registerCopy() {
...
+        ? workspace.targetWorkspace
+        : workspace;
       return (
-        !workspace.isReadOnly() &&
-        !Gesture.inProgress() &&
         !!focused &&
-        isCopyable(focused) &&
-        !getFocusManager().ephemeralFocusTaken()
+        !!targetWorkspace &&
+        !targetWorkspace.isReadOnly() &&
```
cpcallen, line 196: It's not clear to me that the main workspace being readonly should prevent blocks from being copied from the flyout.

(It also isn't clear to me that it should prevent blocks from being copied from the main workspace either—but that was the previous behaviour so leaving it as-is for the time being is fine.)

cpcallen (commented):
There are a number of significant issues with this PR, including making a change that we had decided not to make as well as breaking copying of workspace comments.

I think it should be (mostly) reverted, and a different stab taken at making it possible to copy blocks from the workspace, by ascertaining what change had made them uncopybable and addressing that as narrowly as possible.  Further increasing the range of things that can be copied should be left until we decide we want to implement the proposed changes described in #9098 (or a comparable alternative approach).

PR #9134, opened 2025-06-09
fix: Add isCopyable to the ICopyable interface and use it

### LINKED ISSUE #9129: Make comments copyable via adding an (optional) isCopyable method to ICopyable
There are IMHO a number of significant issues with #9107, including:

* Making it possible to copy blocks even if their `.isDeletable()` or `.isMovable()` methods return false, in contradiction with our previous decision not to do that in v12 (see #9084).
* Creating an `isCuttable` helper function that is not guaranteed to be in sync with `isCopyable` (thus contradicting the purpose of having factored `isCopyable` out of the cut and copy `preconditionFn`s in in the first place), and
* Making it impossible to cut or copy workspace comments.

I think #9107 should be reverted (except for new tests), and a different approach taken to resolving google/blockly-keyboard-experimentation#546—ideally by ascertaining what change caused flyout blocks to have become uncopyable in the the first place and narrowly addressing that.

Further increasing the range of things that can be copied is highly desirable but should be left until we decide we want to implement the proposed changes described in #9098 (or some comparable alternative approach).

_Originally posted by @cpcallen in https://github.com/google/blockly/pull/9107#pullrequestreview-2902220302_

RoboErikG:
## The basics

### Resolves

Fixes #9129 

### Proposed Changes

Adds an isCopyable method to the ICopyable interface and cleans up the cut/copy shortcut logic.
- Allows copying any block or ws comment that is internally movable (isOwnMovable) and deletable (isOwnDeletable).
- Allows cutting any block or ws comment that is copyable and deletable on the current workspace (isDeletable).
- Makes paste act on the workspace that was passed into the callback instead of the one set during cut/copy.
- Adds tests for cutting and for ws comments.

### Reason for Changes

The logic for cut/copy/paste was out of date with several assumptions about where actions could take place. A first pass on updating the logic had broken copying/cutting ws comments. This fixes the direct problem and does further cleanup and testing of the cut/copy/paste behavior.

### Test Coverage

Added new tests for cutting and for ws comments.

### DIFF
--- a/core/interfaces/i_copyable.ts
+++ b/core/interfaces/i_copyable.ts
@@ -22,13 +22,6 @@ export interface ICopyable<T extends ICopyData> extends ISelectable {
    * @returns True if it can currently be copied.
    */
   isCopyable?(): boolean;
-
-  /**
-   * Whether this instance is currently cuttable.
-   *
-   * @returns True if it can currently be cut.
-   */
-  isCuttable?(): boolean;
 }
 
 export namespace ICopyable {
--- a/core/interfaces/i_copyable.ts
+++ b/core/interfaces/i_copyable.ts
@@ -17,7 +17,8 @@ export interface ICopyable<T extends ICopyData> extends ISelectable {
   toCopyData(): T | null;
 
   /**
-   * Whether this instance is currently copyable.
+   * Whether this instance is currently copyable. The standard implementation
+   * is to return true if isOwnDeletable and isOwnMovable return true.
    *
    * @returns True if it can currently be copied.
    */
--- a/core/interfaces/i_copyable.ts
+++ b/core/interfaces/i_copyable.ts
@@ -15,6 +15,20 @@ export interface ICopyable<T extends ICopyData> extends ISelectable {
    * @returns Copy metadata.
    */
   toCopyData(): T | null;
+
+  /**
+   * Whether this instance is currently copyable.
+   *
+   * @returns True if it can currently be copied.
+   */
+  isCopyable?(): boolean;
+
+  /**
+   * Whether this instance is currently cuttable.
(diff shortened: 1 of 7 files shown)


core/interfaces/i_copyable.ts
```diff
@@ -15,6 +15,13 @@ export interface ICopyable<T extends ICopyData> extends ISelectable {
    * @returns Copy metadata.
    */
   toCopyData(): T | null;
+
+  /**
+   * Whether this instance is currently copyable.
+   *
```
cpcallen: Might it be worth documenting what the behaviour is if this method is not present—or, since that description might get out of date, at least providing a pointer to the relevant implementation?

core/shortcut_items.ts
```diff
@@ -100,74 +94,43 @@ export function registerDelete() {
...
-    isICopyable(focused) &&
-    isIDeletable(focused) &&
-    focused.isOwnDeletable() &&
-    isDraggable(focused) &&
-    focused.isOwnMovable()
-  );
+function isCopyable(focused: IFocusableNode): boolean {
+  if (!isICopyable(focused) || !isIDeletable(focused) || !isDraggable(focused))
+    return false;
+  if (focused.isCopyable !== undefined) {
```
cpcallen: The `!== undefilned` test is… OK but kind of mid.  I'd probably opt to be concise:
```suggestion
  if (focused.isCopyable) {
```
which has the advantage of throwing an error if a developer does something unexpected—or be thorough:
```suggestion
  if (typeof focused.isCopyable === 'function') {
```
which should reduce the chance of breaking any existing unexpected implementations.

cpcallen (approved):

PR #9135, opened 2025-06-11
feat: Run keyboard plugin tests in CI

BenHenning:
## The basics

### Resolves

N/A (no tracking issue)

### Proposed Changes

Introduces a GitHub actions CI workflow to run the webdriver IO tests from https://github.com/google/blockly-keyboard-experimentation as part of core Blockly's CI.

### Reason for Changes

Since development on the plugin is going to continue for many months yet, this ensures that behavioral changes in core Blockly don't inadvertently break the plugin.

Note that this shouldn't be made a blocking workflow since there may be cases where it's necessary to break the plugin before a change to the plugin itself can be introduced to then fix it (as this has happened many times in the past). However, the CI check is forced signal to both author and reviewer as to whether their change affects the plugin without having to manually check the test suite.

### Test Coverage

N/A -- Verifying that the CI workflow runs is sufficient.

### Documentation

No documentation changes are needed here.

### Additional Information

Nothing.

### DIFF
--- a/.github/workflows/keyboard_plugin_test.yml
+++ b/.github/workflows/keyboard_plugin_test.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: 'google/blockly-keyboard-experimentation'
-          ref: 'develop'
+          ref: 'main'
           path: blockly-keyboard-experimentation
 
       - name: Use Node.js 20.x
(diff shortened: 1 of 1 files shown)


BenHenning:
The new tests may be slightly flaky and require some occasional re-runs. Probably still work introducing them in CI even with that.

BenHenning:
/cc @rachel-fenichel and @maribethb as well in case you have strong opinions on this.

If everyone's happy with it, feel free to merge while I'm out.

RoboErikG (approved):
@maribethb @rachel-fenichel the config LGTM, but up to you if you want to submit today or leave for next week.

maribethb (approved):

PR #9142, opened 2025-06-13
refactor: Make the cursor use the focus manager for tracking the current node.

### LINKED ISSUE #584: Workspace context menu does not open when focusing an empty workspace (first time only)

gonfunko:
## The basics

### Resolves
This PR fixes https://github.com/google/blockly-keyboard-experimentation/issues/584, but more generally updates the cursor to no longer track its own state wrt the actively focused element, but instead defer to the `FocusManager`. This also allows Marker and MarkerManager to be trivially deleted, but I'm leaving them for now to avoid breaking changes.

### DIFF
--- a/core/keyboard_nav/line_cursor.ts
+++ b/core/keyboard_nav/line_cursor.ts
@@ -17,7 +17,6 @@ import {BlockSvg} from '../block_svg.js';
 import {Field} from '../field.js';
 import {getFocusManager} from '../focus_manager.js';
 import type {IFocusableNode} from '../interfaces/i_focusable_node.js';
-import {isFocusableNode} from '../interfaces/i_focusable_node.js';
 import * as registry from '../registry.js';
 import {WorkspaceSvg} from '../workspace_svg.js';
 import {Marker} from './marker.js';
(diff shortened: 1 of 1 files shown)


core/keyboard_nav/line_cursor.ts
```diff
@@ -374,17 +373,8 @@ export class LineCursor extends Marker {
    *
    * @returns The current field, connection, or block the cursor is on.
    */
-  override getCurNode(): IFocusableNode | null {
-    // Ensure the current node matches what's currently focused.
-    const focused = getFocusManager().getFocusedNode();
-    const block = this.getSourceBlockFromNode(focused);
-    if (block && block.workspace === this.workspace) {
```
rachel-fenichel, line 381: This change is kind of interesting--it means that it's now possible for the cursor to return a block (or field, or connection, etc) that is on a different workspace from the one that the cursor belongs to. Will the other cursor code handle that case correctly?

gonfunko: At least experimentally ~yes, because this is the case for mutators, and that's working with this change. I think we may need to think through this all a bit more thoroughly, because I *think* what we want is for things to be tied to the root workspace, but they're currently tied to `Workspace` in general.

rachel-fenichel: Okay. Can you check if we have an issue to track that work, and add if not? This seems like a place where the failure mode is a game of whack-a-mole. It's a good topic for a deep dive at a team meeting.

I think it's okay to go ahead with this change since the previous behaviour could trivially get weirdly out of sync, and this at least makes it consistent with the focus manager.

rachel-fenichel: Hang tight, thinking more on this one before I approve.

rachel-fenichel (approved):
Okay, go for it.

core/keyboard_nav/line_cursor.ts
gonfunko: Filed https://github.com/google/blockly-keyboard-experimentation/issues/592 to track.

PR #9150, opened 2025-06-17
refactor(interfaces): Make type predicates more robust

cpcallen:
## The basics

### Proposed Changes

Make type predicates for interface types more robust:

* Always use `typeof … === 'function'` when testing for presence of methods.
* Test for presence of all non-optional members.
* Accept input of type `any` but always check that the value is neither `null` nor `undefined`.

### Reason for Changes

Because many of these interfaces define developer-supplied types, making the corresponding type predicates more exacting will help ensure that we do not inadvertently accept a value that only partially conforms to the interface.

For example, given some object
```JS
const obj = {
  a: 0,
  b: 'hello',
  c() {
    console.log('hello');
  }
  d: undefined,
  // There is no e.
}
```
the typical tests we use will all return true for almost any non-missing value:
```JS
obj.a !== undefined  // true
obj.b !== undefined  // true
obj.c !== undefined  // true
obj.d !== undefined  // false
obj.e !== undefined  // false
```
while a lazier test might catch a primitive being supplied where a method was expected, at least if the primitive is falsey:
```JS
Boolean(obj.a)  // FALSE
Boolean(obj.b)  // true
Boolean(obj.c)  // true
Boolean(obj.d)  // false
Boolean(obj.e)  // false
```
whereas using `typeof` will ensure that, at least for primitive- or method-valued members of the interface, the type is approximately as expected:
```JS
typeof obj.a === 'number'  // true

typeof obj.a === 'string'  // false

typeof obj.a === 'function'  // false
typeof obj.b === 'function'  // false
typeof obj.c === 'function'  // true
typeof obj.d === 'function'  // false
typeof obj.e === 'function'  // false
```

Notably, the only values whose `typeof` is `'function'` that are _not_ callable as a method are ES6 class constructors (e.g., `class C {}`), which need to be invoked with `new`.

### Test Coverage

Passes `npm test`.

### Additional Information

One possible argument against this change might be that the type predicates will be slightly more expensive to invoke—though I suspect that in most situations the JIT will optimise away most of the actual checks.  Likely only use of these predicates inside hotspot inner loops is likely to be of any measurable effect, but it would be useful if we had some benchmarks so we could actually test to see what effect if any this change might have.

### DIFF
--- a/core/interfaces/i_autohideable.ts
+++ b/core/interfaces/i_autohideable.ts
@@ -23,5 +23,5 @@ export interface IAutoHideable extends IComponent {
 
 /** Returns true if the given object is autohideable. */
 export function isAutoHideable(obj: any): obj is IAutoHideable {
-  return typeof obj.autoHide === 'function';
+  return obj && typeof obj.autoHide === 'function';
 }
--- a/core/interfaces/i_autohideable.ts
+++ b/core/interfaces/i_autohideable.ts
@@ -23,5 +23,5 @@ export interface IAutoHideable extends IComponent {
 
 /** Returns true if the given object is autohideable. */
 export function isAutoHideable(obj: any): obj is IAutoHideable {
-  return obj.autoHide !== undefined;
+  return typeof obj.autoHide === 'function';
 }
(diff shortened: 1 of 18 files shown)


cpcallen:
Marking this as draft for now because it turns out that some of our own test mocks do not conform to their specified types!

maribethb (approved):

PR #9151, opened 2025-06-17
feat: Add support for keyboard navigation into mutator workspaces.

gonfunko:
## The basics

### Proposed Changes
This PR lays the groundwork for adding supporting for navigating into mutator workspaces via keyboard navigation. Specifically, it:

* Updates `MiniWorkspaceBubble` to resize/reposition more intelligently when a keyboard move is in progress
* Updates the `IconNavigationPolicy` to navigate into `MutatorIcon`s' workspaces
* Updates `WorkspaceSvg` to include its child mutator workspaces as nested trees for focus handling purposes.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -41,6 +41,7 @@ import type {FlyoutButton} from './flyout_button.js';
 import {getFocusManager} from './focus_manager.js';
 import {Gesture} from './gesture.js';
 import {Grid} from './grid.js';
+import {MutatorIcon} from './icons/mutator_icon.js';
 import {isAutoHideable} from './interfaces/i_autohideable.js';
 import type {IBoundedElement} from './interfaces/i_bounded_element.js';
 import {IContextMenu} from './interfaces/i_contextmenu.js';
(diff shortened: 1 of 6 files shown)


core/workspace_svg.ts
```diff
@@ -2710,7 +2711,22 @@ export class WorkspaceSvg
...
-    return [];
+    const nestedWorkspaces = this.getAllBlocks()
+      .map((block) => block.getIcons())
+      .flat()
+      .filter(
+        (icon): icon is MutatorIcon =>
+          icon instanceof MutatorIcon && icon.bubbleIsVisible(),
+      )
+      .map((icon) => icon.getBubble()?.getWorkspace())
+      .filter((workspace) => !!workspace);
```
cpcallen, line 2722: This seems like a very expensive operation.  Is it not the case that there is at most one visible bubble containing a workspace at a time?

cpcallen (approved):

core/workspace_svg.ts
cpcallen: OK, I spoke too soon: we do support having multiple workspace bubbles open simultaneously.  Still: it seems like we should be able to find them without having to travers all the icons on all the blocks on the workspace.

gonfunko: Agreed, I don't care for this at all; unfortunately I'm not aware of another way of achieving the goal here, and there's similar code in `lookupFocusableNode()`. We might need a way for bubbles/icons to register themselves with the workspace, I'll file an issue to explore that.

gonfunko: https://github.com/google/blockly/issues/9162

PR #9172, opened 2025-06-26
fix: clear touch identifier on comment text area pointerdown

### LINKED ISSUE #9108: touch identifier not always cleared properly when interacting with comments
### Check for duplicates

### Description

when interacting with comments, the touch identifier for the current gesture doesn't always get cleared properly. this doesn't particularly matter in most browsers because they reuse pointerid values for different pointer events. in my testing, most browsers use 1 for the primary pointer and only use higher numbers when more than one touch is active. as a result the touchidentifier might not be cleared but the next pointer event will also have a pointerid of 1 (despite it being a separate gesture) and blockly will continue to process gesture events.

i say "most" browsers because ios safari is *not* most browsers. it assigns a unique random number for each pointerid so failing to clear the identifier when the pointer is lifted will prevent all future gestures from executing

also worth noting that the prevailing behavior of a pointerid of "1" is not actually part of [the spec](https://www.w3.org/TR/pointerevents/#dom-pointerevent-pointerid), so it's probably a good idea to make sure this is cleared anyhow

i worked around this in makecode in a fairly hacky manner since i believe a proper fix will require changing the comment view class in blockly: https://github.com/microsoft/pxt/pull/10613

### Reproduction steps

1. open the blockly playground in ios safari (or in the ipad simulator from xcode)
2. add a block to the workspace
3. add a comment to the workspace
4. click on the comment and edit its text
5. click on the workspace
6. try to move the block. note that it is now "stuck" and all interaction will fail until you click something that happens to clear the touch identifier

### Stack trace

```shell
none
```

### Screenshots

none

### Browsers

iOS mobile

### LINKED ISSUE #9170: Touch-specific issues affecting comments
Using `npm start` on develop as of f4dbea0a65ad4808f8318f3bbe6aa5000b24c235.

## Workspace comments

https://github.com/user-attachments/assets/ca0f1af9-8f78-4087-93dc-eba532dae9f3

1. Using a touch device or dev tools emulating one, add a workspace comment (long press for menu)
2. Focus the text area
3. Delete the comment via the icon
4. Note it just moves focus
5. Try again, nothing happens
6. Try to select a block, selection doesn't move

This seems to have been an issue in v11 too.

(6) also happens if you skip 3/4/5 provided you did (2).

## Block comments

https://github.com/user-attachments/assets/ecbe1c39-8863-44c3-8770-e4da9b817b16

1. Using a touch device or dev tools emulating one, add a block comment
2. Focus the text area
3. Try to select the block, selection doesn't move

By repeatedly tapping back and forth I can sometimes get selection to move but it seems in a broken state from then on.

I couldn't reproduce this in v11 (you need a second block to try as selection remains on the block when the comment takes focus).

riknoll:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/9170
Fixes #9108 

### Proposed Changes

Clear the cached touch identifier inside the pointerdown event handler inside the `CommentView`

### Reason for Changes

The pointerdown event handler inside of `CommentView` uses `conditionalBind` to bind the event listener. Inside that event listener, it then calls `stopPropagation` to prevent the event from bubbling. As a result a touch identifier is registered but it never gets cleared which prevents all further pointer events from firing on the workspace

### Test Coverage

Tested in playground locally.

### Additional Information

@microbit-matt-hillsdon FYI

### DIFF
--- a/core/bubbles/textinput_bubble.ts
+++ b/core/bubbles/textinput_bubble.ts
@@ -176,6 +176,7 @@ export class TextInputBubble extends Bubble {
     // Don't let the pointerdown event get to the workspace.
     browserEvents.conditionalBind(textArea, 'pointerdown', this, (e: Event) => {
       e.stopPropagation();
+      touch.clearTouchIdentifier();
     });
 
     browserEvents.conditionalBind(textArea, 'change', this, this.onTextChange);
--- a/core/comments/comment_editor.ts
+++ b/core/comments/comment_editor.ts
@@ -8,6 +8,7 @@ import * as browserEvents from '../browser_events.js';
 import {getFocusManager} from '../focus_manager.js';
 import {IFocusableNode} from '../interfaces/i_focusable_node.js';
 import {IFocusableTree} from '../interfaces/i_focusable_tree.js';
+import * as touch from '../touch.js';
 import * as dom from '../utils/dom.js';
 import {Size} from '../utils/size.js';
 import {Svg} from '../utils/svg.js';
@@ -80,6 +81,7 @@ export class CommentEditor implements IFocusableNode {
         // and steal focus away from the editor/comment.
         e.stopPropagation();
         getFocusManager().focusNode(this);
+        touch.clearTouchIdentifier();
       },
     );
 


microbit-matt-hillsdon:
I think an equivalent fix is needed for the block comment case in `textinput_bubble.ts`.

maribethb (approved):

Issue #9171, opened 2025-06-26
Use strongly typed identifier for focusable nodes


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Using strings for uniquely identifying `IFocusableNode`s is introducing magic strings (https://en.wikipedia.org/wiki/Magic_string) which can and have caused problems.

### Request

While it would be nice if TypeScript supported `final` and sealed classes, a reasonable alternative might be to introduce a new class, say `FocusableNodeId` which takes an enum for type and a unique ID (probably automatically managed using `idgenerator.ts` with a custom equality check function.

It's already been tested that attaching arbitrary objects to elements works just fine, so the system would use a new attribute (perhaps `focusId`) instead of `id` and utilize this object type, instead.

### Alternatives considered

There are other approaches that might be nice to consider. Two main drawbacks with enums:
1. They result in a single location for all node types, even across trees (which can be a bit weird as it's creating a sort of global repository for all things focus node).
2. They must be updated for new node types, and are inherently not extensible for new types of nodes (though this may actually be preferred since we don't really want people creating custom nodes or trees outside of core Blockly).

### Additional context

It's worth noting that this change can be done _internally_ without it being breaking (though it will require some sort of clever interoperability solution in order to ensure plugins and custom components continue to work before they can be updated).

One nice thing is that technically the contract for `IFocusableNode` can change to provide the ID via its element without it being a compile-time error. It will introduce a runtime error, however, for custom components outside of core Blockly (so this will need to be carefully considered). However, `IFocusableTree` can **not** be updated without introducing a breaking change, so that will need to wait to v13.

One interoperable solution might be to introduce a translation layer between the current magic strings and the new ID type. This could allow us to update all logic in trees, the traverser, and `FocusManager` to use the new IDs without actually requiring any API changes or even changes in nodes (though we could design the system to support both legacy magic strings in nodes and the new type since they are different attributes). We could even introduce new functions entirely for the tree traverser to keep it simple (we can't change it directly since it's exported). Then, v13 would involve updating the actual APIs and removing support for the legacy IDs entirely.

BenHenning:
NB: #9173 actually ran into a legitimate issue with the magic string IDs being used today and led to introducing two new block ID checks to catch at least one class of problems that can arise from using these.

BenHenning:
NB: #9155 is another example of how strongly typed IDs can result in a safer and more robust API.

BenHenning:
[...]

PR #9173, opened 2025-06-26
fix: Ensure focus changes when tabbing fields

### LINKED ISSUE #578: Tabbing through input fields doesn't move the cursor
### Check for duplicates

### Description

When editing a field (number / text), passive focus is shown around that field and you can use 'Tab' to move to the next input field. When moving between inputs, passive focus is not updated. When you hit 'Enter' or 'Escape', the focus can arguably be returned to the wrong input as it's always returned to the input you started on.

### Reproduction steps

Please see video below for reproduction.

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/df57e515-80e7-4ce4-9690-b8a58dbe72ad

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/578
Fixes part of #8915

### Proposed Changes

Ensures fields update focus to the next field when tabbing between field editors. The old behavior can be observed in [#578](https://github.com/google/blockly-keyboard-experimentation/issues/578) and the new behavior can be observed here:

[Screen recording 2025-06-25 1.39.28 PM.webm](https://github.com/user-attachments/assets/e00fcb55-5c20-4d5c-81a8-be9049cc0d7e)

### Reason for Changes

Having focus reset back to the original field editor that was opened is an unexpected experience for users. This approach is better.

Note that there are some separate changes added here, as well:
- Connections and fields now check if their block IDs contain their indicator prefixes since this will all-but-guarantee focus breaks for those nodes. This is an excellent example of why #9171 is needed.
- Some minor naming updates for `FieldInput`: it was incorrectly implying that key events are sent for changes to the `input` element used by the field editor, but they are actually `InputEvent`s per https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event.

### Test Coverage

New tests were added for field editing in general (since this seems to be missing), including tabbing support to ensure the fixes originally introduced in #9049.

One new test has been added specifically for verifying that focus updates with tabbing. This has been verified to fail with the fix removed (as have all tabbing tests with the tabbing code from #9049 removed).

Some specific notes for the test changes:
- There's a slight test dependency inversion happening here. `FieldInput` is being tested with a specific `FieldNumber` class via real block loading. This isn't ideal, but it seems fine given the circumstances (otherwise a lot of extra setup would be necessary for the tests).
- The workspace actually needs to be made visible during tests in order for focus to work correctly (though it's reset at the end of each test, but this may cause some flickering while the tests are running).
- It's the case that a bunch of tests were actually setting up blocks incorrectly (i.e. not defining a must-have `id` property which caused some issues with the new field and connection ID validation checks). These tests have been corrected, but it's worth noting that the blocks are likely still technically wrong since they are not conforming to their TypeScript contracts.
- Similar to the previous point, one test was incorrectly setting the first ID to be returned by the ID generator as `undefined` since (presumably due to a copy-and-paste error when the test was introduced) it was referencing a `TEST_BLOCK_ID` property that hadn't been defined for that test suite. This has been corrected as, without it, there are failures due to the new validation checks.
- For the connection database checks, a new ID is generated instead of fixing the block ID to ensure that it's always unique even if called multiple times (otherwise a block ID would need to be piped through from the calling tests, or an invalid situation would need to be introduced where multiple blocks shared an ID; the former seemed unnecessary and the latter seemed nonideal).
- There are distinct Geras/Zelos tests to validate the case where a full-block field should have its parent block, rather than the field itself, focused on tabbing. See this conversation for more context: https://github.com/google/blockly/pull/9173#discussion_r2172921455.

### Documentation

### Additional Information

Nothing to add.


BenHenning (commented):
Spot checked changes.

BenHenning (commented):
Spot-checked test fix changes.

core/field_input.ts
```diff
@@ -614,6 +615,7 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
       if (target instanceof FieldInput) {
         WidgetDiv.hideIfOwner(this);
         dropDownDiv.hideWithoutAnimation();
+        getFocusManager().focusNode(target);
```
gonfunko: Drive by comment: In the case of full-block fields, I think we actually want to focus the parent block for consistency.

core/connection.ts
```diff
@@ -83,6 +83,12 @@ export class Connection {
     public type: number,
   ) {
     this.sourceBlock_ = source;
+    if (source.id.includes('_connection')) {
+      console.warn(
+        `Connection ID indicator is contained in block ID. This may cause ` +
+          `problems with focus: ${source.id}.`,
+      );
+    }
```
cpcallen: Maybe I am misunderstanding something, but is it not an error for a block to have an ID containing `'_connection'`?  And if it is an error, should this not be `throw new TypeError` (or at least `console.error`) instead of merely `console.warn`?

Also, are we not testing for this in the `Block` constructor?

core/field_input.ts
```diff
@@ -83,8 +84,8 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
   /** Key down event data. */
   private onKeyDownWrapper: browserEvents.Data | null = null;
 
-  /** Key input event data. */
-  private onKeyInputWrapper: browserEvents.Data | null = null;
+  /** Input element input event data. */
+  private onInputWrapper: browserEvents.Data | null = null;
```
cpcallen, line 88: Since you're working on improving naming:

I don't think a variable of type `Data` should be called `wrapper`, since the latter strongly implies (to me at least) that the value of the variable will be a callable function—which AFAICT this is not.  Maybe:

```suggestion
  private onInputBinding: browserEvents.Data | null = null;
```

core/field_input.ts
cpcallen: Yes, agreed.  (And note that whether a field is a full-block field or not depends on the renderer.)

tests/mocha/field_textinput_test.js
```diff
@@ -23,11 +23,50 @@ import {
 } from './test_helpers/setup_teardown.js';
 
 suite('Text Input Fields', function () {
-  setup(function () {
+  setup(async function () {
```
cpcallen: I don't think this needs to be an async function: its body contains no occurrences of `await`.

cpcallen (changes requested):

core/connection.ts
BenHenning: I don't see any specific validation for ID in `Block` or `BlockSvg`'s constructors, but maybe there's somewhere subtle where it happens? We could check for the field & connection failure cases there, instead.

It probably is an error, I was just hesitant to make it actually throw. I'm fine with doing that, however. Updated. That being said, I'm using `Error`, not `TypeError`. The type (string) is technically correct which makes this a value, not type, incompatibility.

I'm less certain about moving the checks elsewhere because these failure cases actually came up from weirdly set up tests. We have a BUNCH of tests that set up invalid blocks by defining them directly as JSON objects that don't actually define required properties like `id`. I fixed the ones that were relevant here, but that does mean that the constructor isn't guaranteed to be called (and thus this check could miss real failures in tests even if it works correctly for production code).

core/field_input.ts
BenHenning: Did a quick audit and it seems we call the returned object any of the following:
- Wrapper (here and three other files)
- Event (one file)
- Listener (three files)
- Handler (two files)

I'm not opposed to fixing the naming, but this introduces a new type and thus technically increases the inconsistency. I can rename all of the cases where it's used, but I lean toward doing that in a follow-up PR to keep it separate (in case this needs to be reverted for some reason).

Does that sound good to you? If so, do you have a strong opinion on 'binding' being the "one true name suffix" we should use everywhere? (I think 'binding' is reasonable, for what it's worth, though I might lean a bit more toward 'eventBinding' to be extra clear).

tests/mocha/field_textinput_test.js
BenHenning: I think this is a remnant from when I previously had a call to `finishQueuedRenders` here. Nice catch--removed.

core/field_input.ts
BenHenning: Didn't know about full-block fields before this. Fortunately, the block being tested here already uses full-block fields in Zelos, so I updated the tests to verify behavior specifically with Zelos and Geras separately. I confirmed that one of the Zelos tests fails without the new `focusNode` line in place.

BenHenning (commented):
Spot-checked latest changes.

BenHenning:
CI is passing and I think everything should now be addressed. PTAL @cpcallen.

core/field_input.ts
cpcallen: I agree consistency is good, but am happy for you to leave it to a subsequent PR.

I have no strong preference about the exact naming, though given my background in C and Go I generally prefer shorter names than some of our erstwhile Java-programming colleagues.

cpcallen (approved):

core/connection.ts
cpcallen: If I understand correctly #9171 should address the problem that these checks are trying to guard against, but if that's not the case would you open an issue to ensure that we have suitable checks (in constructor, deserialiser, etc.) to ensure that invariants about identifiers are maintained?

BenHenning: That will fix it. Do you still think the checks are needed in the meantime?

Will follow up separately so that I can merge this PR.

cpcallen: Further checks are not needed for this PR.

I will leave it to you to decide if there is cause to add other interim checks before the matter can be addressed by #9171.

BenHenning: In that case, I'll probably hold off but I'll make note of this conversation in that thread.

PR #9174, opened 2025-06-26
fix: Short-circuit node lookups for missing IDs

### LINKED ISSUE #9155: short circuit `findFocusableNode` if id is null/empty
This is the only place that we call `lookUpFocusableNode`: https://github.com/google/blockly/blob/97ffea73becea41465e7c4e757fab44a2ded10df/core/utils/focusable_tree_traverser.ts#L89

I _suspect_ this is happening for all the cases `document.body` gets focus (which is a lot, including in these editable textarea cases for reasons I'm not quite sure of). This is actually another argument in favor of moving toward typed IDs since it would be more explicit why an ID is missing, versus just being empty.

We should patch it in the traverser, I think, rather than here. In fact, we can probably short-circuit that entire function and just return `null` if the ID is empty for the provided element (since `IFocusableNode` requires a valid, unique ID to be present which means it can't be an empty string/left undefined).

_Originally posted by @BenHenning in https://github.com/google/blockly/pull/9154#discussion_r2155512964_

`lookUpFocusableNode` is doing a bunch of unnecessary work when it's called with an empty string id (which it is many times) -- easy performance win to not do that work if id is empty.

BenHenning:
## The basics

### Resolves

Fixes #9155

### Proposed Changes

In cases when an ID is missing for an element passed to `FocusableTreeTraverser.findFocusableNodeFor()`, always return `null`.

Additionally, the new short-circuit logic exposed that `Toolbox` actually wasn't being set up correctly (that is, its root element was not being configured with a valid ID). This has been fixed.

### Reason for Changes

These are cases when a valid node should never be matched (and it's technically possible to incorrectly match if an `IFocusableNode` is set up incorrectly and is providing a focusable element with an unset ID). This avoids the extra computation time of potentially calling deep into `WorkspaceSvg` and exploring all possible nodes for an ID that should never match.

Note that there is a weird quirk with `null` IDs actually being the string `"null"`. This is a side effect of how `setAttribute` and attributes in general work with HTML elements. There's nothing really that can be done here, so it's now considered invalid to also have an ID of string `"null"` just to ensure the `null` case is properly short-circuited.

Finally, the issue with toolbox being configured incorrectly was discovered with the introducing of a new hard failure in `FocusManager.registerTree()` when a tree with an invalid root element is registered. From testing there are no other such trees that need to be updated.

A new warning was also added if `focusNode()` is used on a node with an element that has an invalid ID. This isn't a hard failure to follow the convention of other invalid `focusNode()` situations. It's much more fragile for `focusNode()` to throw than `registerTree()` since the former generally happens much earlier in a page lifecycle, and is less prone to dynamic behaviors.

### Test Coverage

New tests were added to validate the various empty ID cases for `FocusableTreeTraverser.findFocusableNodeFor()`, and to validate the new error check for `FocusManager.registerTree()`.

### Documentation

No new documentation should be needed.

### Additional Information

Nothing to add.


BenHenning (commented):
Spot checked changes.

BenHenning:
Looks like this is a slightly less simple fix than I was hoping. Core tests are passing, but the nav tests might have issues (perhaps indicating a setup issue). Will need to dig.

BenHenning (commented):
Spot checked latest fixes.

core/focus_manager.ts
```diff
@@ -349,8 +356,12 @@ export class FocusManager {
     // happens when calls to focusNode() interleave with asynchronous clean-up
     // operations (which can happen due to ephemeral focus and in other cases).
     // Fall back to a reasonable default since there's no valid node to focus.
+    const focusableNodeElement = focusableNode.getFocusableElement();
+    if (!focusableNodeElement.id || focusableNodeElement.id === 'null') {
+      console.warn('Trying to focus a node that has an invalid ID.');
```
RoboErikG: Should this clear isUpdatingFocusedNode and return early? It looks like the only time the rest of this method would do anything is if the element is a root node.

RoboErikG (approved):
One suggestion, otherwise LGTM.

core/focus_manager.ts
BenHenning: I don't think that's correct, but perhaps I'm misunderstanding you.

If there's an invalid ID, then `findFocusedNode` will return null causing the condition on line 369 to fail, thus leading to the node to focus falling back to the tree's defaults (which can be one of several nodes depending on the tree and its state).

I'll add a line comment here to make it a bit more clear that it's intentionally falling through.

BenHenning (commented):
Self-reviewed latest.

BenHenning:
@RoboErikG I'll go ahead and merge this, but per the conversation thread please let me know if I missed something and if there's something else you think needs to be addressed here.

PR #9181, opened 2025-06-27
Use 4-space indention for Python

vanillajonathan:
As per PEP-8.

## The basics

### Proposed Changes

Change the indention for Python code from 2 spaces to 4 spaces.

### Reason for Changes

The convention for indention in the Python community is [4 spaces as per PEP-8](https://peps.python.org/pep-0008/#indentation).

### DIFF
--- a/generators/python/python_generator.ts
+++ b/generators/python/python_generator.ts
@@ -76,6 +76,8 @@ export class PythonGenerator extends CodeGenerator {
     [Order.LOGICAL_OR, Order.LOGICAL_OR],
   ];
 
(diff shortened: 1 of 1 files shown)


gonfunko:
Thanks for the PR! While you're certainly correct about the PEP 8 guidance, in talking with the team we'd like to leave the current behavior to avoid disruption for folks who are using the current generator, and particularly since this is already user-configurable; running `Blockly.Python.INDENT = '    '` before generating code will cause it to use 4 spaces without having to change the core implementation or bother with a subclass.

PR #9182, opened 2025-06-27
feat: Add support for keyboard navigation in/to workspace comments.

### LINKED ISSUE #9167: Shortcuts to move focus from textarea to workspace comment do not always work
#9154 added Escape and Cmd/Ctrl+Enter shortcuts to move focus from a workspace comment textarea to the comment, allowing you to get out of the text area.

However, they only work for comments that have been focused (the comment itself, not the text area).

Repro:

1. Click in an existing workspace comment textarea or even tab into the textarea, [e.g. via this demo](https://test-comments.blockly-keyboard-experimentation.pages.dev/?scenario=comments&toolbox=toolbox&renderer=zelos)
2. Press escape or Cmd+Enter
3. Nothing happens, focus is still on the text area

Cause: The keydown handler calls `this.svgRoot.focus()` but at this point the SVG root is a `<g>` element without a tabindex so cannot be focused. New comment or comments you've clicked on have tabindex=-1 so those cases work.

Ideally this would be fixed in the context of a broader plan for keyboard navigation of comments that e.g. eliminates the comment tab stops in favour of navigation more similar to block navigation (at least when using the keyboard nav plugin). But providing some route back to blocks once you've moved to a comment is also important.

gonfunko:
## The basics

### Resolves

Fixes #9167 
Fixes part of https://github.com/google/blockly-keyboard-experimentation/issues/614

### Proposed Changes
This PR adds support for keyboard navigation in and through workspace comments. Specifically, it:

* Factors out comment icons into their own classes that support being focused
* Adds navigation policies for workspace comments and icons
* Improves the Rect and WorkspaceSvg APIs to have better typings/additional convenience functions to support this

A (small) corresponding PR will still be needed in the keyboard-experimentation repo.

With this change, workspace comments are part of the keyboard nav flow; with a comment focused, Enter will focus its editor, Escape will return focus to the comment as a whole, and the collapse/delete icons may be reached by right-arrowing in from a focused workspace comment. Up and down include workspace comments as well as blocks in the list of locations that they visit.


core/comments/comment_icon.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import type {IFocusableNode} from '../interfaces/i_focusable_node.js';
+import {Rect} from '../utils/rect.js';
+import type {WorkspaceSvg} from '../workspace_svg.js';
+import type {RenderedWorkspaceComment} from './rendered_workspace_comment.js';
+
+/**
+ * Actionable icon displayed on a comment's top bar.
+ */
+export abstract class CommentIcon implements IFocusableNode {
```
maribethb: this is a confusing name because `CommentIcon` already exists to refer to the block icon that adds a block comment. maybe `CommentBarIcon`? Possibly calling these icons at all is confusing, and we should just call them buttons.

core/comments/comment_icon.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * SVG image displayed by this icon.
+   */
+  protected abstract readonly icon: SVGImageElement;
+
+  /**
+   * Creates a new CollapseCommentIcon instance.
```
maribethb: this tsdoc is for the wrong class

core/comments/comment_icon.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Returns the dimensions of this icon in SVG units.
+   *
+   * @param includeMargin True to include the margin when calculating the size.
+   * @returns The size of this icon.
+   */
+  getSize(includeMargin = false): Rect {
+    const bounds = this.icon.getBBox();
+    const rect = Rect.from(bounds);
+    if (includeMargin) {
```
maribethb: Since `getMargin` is calculating itself based on the container, and then you're undoing that math here... can you just return the stats for the `container` instead of doing this math?

core/comments/comment_icon.ts
```diff
@@ -0,0 +1,103 @@
...
+      const margin = this.getMargin();
+      rect.left -= margin;
+      rect.top -= margin;
+      rect.bottom += margin;
+      rect.right += margin;
+    }
+    return rect;
+  }
+
+  /** Returns the margin in SVG units surrounding this icon. */
```
maribethb: is SVG units a thing we refer to elsewhere? I thought we just used workspace and pixel. Is SVG something different?

core/keyboard_nav/block_navigation_policy.ts
```diff
@@ -143,21 +145,29 @@ function getBlockNavigationCandidates(
...
+ * @returns The first element in the stack offset by `delta` relative to the
+ *     current element's stack, or the last element in the stack offset by
+ * `delta` relative to the current element's stack when navigating backwards.
  */
-export function navigateStacks(current: BlockSvg, delta: number) {
-  const stacks = current.workspace.getTopBlocks(true);
-  const currentIndex = stacks.indexOf(current.getRootBlock());
+export function navigateStacks(
+  current: IFocusableNode,
+  workspace: WorkspaceSvg,
```
maribethb: Is it a breaking change to add this parameter here? Could you just get the workspace from `current.workspace`? It seems like that's the only value you ever pass in.

If it stays, it needs tsdoc

core/comments/comment_icon.ts
gonfunko: Done.

core/keyboard_nav/block_navigation_policy.ts
gonfunko: Technically yes, but it was introduced in 12.1, almost assuredly has no third-party callers and per [the docs](https://developers.google.com/blockly/guides/programming/using_blockly_apis#annotations-public) we reserve the right to modify new APIs in the subsequent release. I can mark this PR as breaking if you want though? And added the docs.

core/comments/comment_icon.ts
gonfunko: The container is wider than the bar button, so we'd still need to munge it vs just returning the container's bounding box if we're including margins. I think the current approach is clearer than taking the container's bounds and manipulating their width?

core/comments/comment_icon.ts
gonfunko: Renamed to `CommentBarButton`.

core/comments/comment_icon.ts
gonfunko: I thought it was different, but `FlyoutMetricsManager.getBoundingBox()` uses it and refers to it as workspace coordinates, so updated accordingly.

core/comments/comment_icon.ts
maribethb: Yeah you're right, I missed the width calculation was still needed. thanks

core/comments/comment_bar_button.ts
```diff
@@ -0,0 +1,105 @@
...
+      const margin = this.getMargin();
+      rect.left -= margin;
+      rect.top -= margin;
+      rect.bottom += margin;
+      rect.right += margin;
+    }
+    return rect;
+  }
+
+  /** Returns the margin in SVG units surrounding this button. */
```
maribethb: This one still says SVG units instead of workspace units

core/keyboard_nav/block_navigation_policy.ts
maribethb: no, don't want to do v13 next week :) I think it's fine if you need to add it, but is there a reason you can't just use `current.workspace` within the body of the method since that's the only value you ever pass it? I guess `IFocusableNode` doesn't guarantee that it's something that has a `workspace` property... does this method work if you change the parameter `current` to be of type `ISelectable` instead? that would cover blocks and workspace comments and I _think_ nothing else gets passed in.

maribethb (approved):

core/comments/comment_bar_button.ts
gonfunko: Done.

core/keyboard_nav/block_navigation_policy.ts
gonfunko: Ooh good call, yes the lack of `.workspace` was why I did it this way, but `ISelectable` works nicely. Thanks!

PR #9183, opened 2025-06-27
fix: Fix more browser tests

### LINKED ISSUE #8953: Browser tests fail with "element not interactable"

RoboErikG:
## The basics

This fixes the remaining browser tests

### Resolves

Fixes [#8953](https://github.com/google/blockly/issues/8953)

### Proposed Changes

Lots of fixes to the browser tests


tests/browser/test/delete_blocks_test.mjs
```diff
@@ -194,7 +194,8 @@ suite('Delete blocks', function (done) {
     );
   });
 
-  test('Redo block deletion', async function () {
+  // TODO(#9029) enable this test once deleting a block doesn't lose focus
```
gonfunko: This was fixed in https://github.com/google/blockly/pull/9195, can this be reenabled now?

tests/browser/test/test_setup.mjs
```diff
@@ -295,6 +298,55 @@ export async function getBlockTypeFromCategory(
...
+  // In the browser context, find the element that we want and give it a findable ID.
+  await browser.execute(
+    (blockType, newElemId) => {
+      const block = Blockly.getMainWorkspace()
+        .getFlyout()
+        .getWorkspace()
+        .getBlocksByType(blockType)[0];
+      if (!block.isCollapsed()) {
+        for (const input of block.inputList) {
+          for (const field of input.fieldRow) {
```
gonfunko: There's a recently-ish introduced `getFields()` that returns a generator that might be more convenient here, but not a blocker.

tests/browser/test/test_setup.mjs
```diff
@@ -448,6 +500,9 @@ export async function switchRTL(browser) {
  */
 export async function dragNthBlockFromFlyout(browser, categoryName, n, x, y) {
   const flyoutBlock = await getNthBlockOfCategory(browser, categoryName, n);
+  while (!(await elementInBounds(browser, flyoutBlock))) {
+    await scrollFlyout(browser, 0, 50);
+  }
```
gonfunko: Could you use `WorkspaceSvg.scrollBoundsIntoView(flyoutBlock.getBoundingRectangleWithoutChildren())` rather than scrolling repeatedly by a fixed amount until it's on screen?

tests/browser/test/test_setup.mjs
```diff
@@ -474,15 +529,44 @@ export async function dragBlockTypeFromFlyout(
...
 ) {
-  const flyoutBlock = await getBlockTypeFromCategory(
+  const flyoutBlock = await getDraggableBlockElementByType(
     browser,
     categoryName,
     type,
   );
+  while (!(await elementInBounds(browser, flyoutBlock))) {
+    await scrollFlyout(browser, 0, 50);
+  }
```
gonfunko: Here as well

tests/browser/test/test_setup.mjs
gonfunko: This would seemingly allow you to get rid of `elementInBounds`, since it should be safe to do unconditionally.

tests/browser/test/toolbox_drag_test.mjs
```diff
@@ -57,28 +57,29 @@ const testCategories = [
...
+    return Blockly.getMainWorkspace()
+      .getFlyout()
+      .getWorkspace()
+      .getTopBlocks(false)[i].type;
+  }, n);
 
-    return vertInView && horInView;
-  }, element);
+  // Unicode escape to close flyout.
+  await browser.keys(['\uE00C']);
```
gonfunko: Can this be `webdriverio.Key.Escape`?

tests/browser/test/delete_blocks_test.mjs
RoboErikG: Done.

tests/browser/test/test_setup.mjs
RoboErikG: Done.

tests/browser/test/test_setup.mjs
RoboErikG: It took some reworking of a few of the test methods, but does seem to be cleaner and work better overall. =)

tests/browser/test/toolbox_drag_test.mjs
RoboErikG: Done here and elsewhere.

RoboErikG (commented):
Updated based on comments

gonfunko (approved):

PR #9196, opened 2025-07-01
chore: Merge develop into translatewiki

gonfunko:
This PR merges develop into the translatewiki branch in order to bring the localizable messages up to date.

This should be submitted as a MERGE COMMIT.


BenHenning:
Just checking @gonfunko: did you resolve all of the conflicts by just taking the msg files from the `translatewiki` branch instead of any from `develop`? I'm assuming there were conflicts because we changed the `msg` files at some point independently of the generated Translatewiki changes, and we don't need to pull those since the strings we pull from Translatwiki should be the source of truth.

msg/json/en.json
```diff
@@ -394,5 +397,26 @@
 	"WORKSPACE_ARIA_LABEL": "Blockly Workspace",
 	"COLLAPSED_WARNINGS_WARNING": "Collapsed blocks contain warnings.",
 	"DIALOG_OK": "OK",
-	"DIALOG_CANCEL": "Cancel"
```
cpcallen, line 397: It appears that `DIALOG_CANCEL` was inadvertently removed during merge conflict resolution: AFAICT it still exists in `msg/messages.js` and AFAIUI that means it should also exist here.

cpcallen (approved):
OK, it took me a while to figure out how to review this, but eventually I diffed the `dev-into-wiki` branch against the commit it is pulling in from `develop` (e5804e7)—which should give us exactly the differences we expect to pull back in to `develop` next time we update it from `translatewiki`—and amazingly enough it appears that this enormous PR is practically perfect in every way, with just one single-line merge conflict resolution mistake.

msg/json/en.json
gonfunko: Good catch, thanks! Added back in.

PR #9200, opened 2025-07-02
fix: Prevent loss of focus when deleting a workspace comment.

gonfunko:
## The basics

### Resolves
This PR fixes a bug that could cause focus to be lost in keyboard navigation mode when deleting a workspace comment via its comment bar button.

### DIFF
--- a/core/comments/delete_comment_bar_button.ts
+++ b/core/comments/delete_comment_bar_button.ts
@@ -5,6 +5,7 @@
  */
 
 import * as browserEvents from '../browser_events.js';
+import {getFocusManager} from '../focus_manager.js';
 import * as touch from '../touch.js';
 import * as dom from '../utils/dom.js';
 import {Svg} from '../utils/svg.js';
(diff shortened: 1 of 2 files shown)


core/comments/delete_comment_bar_button.ts
```diff
@@ -98,5 +99,6 @@ export class DeleteCommentBarButton extends CommentBarButton {
 
     this.getParentComment().dispose();
     e?.stopPropagation();
+    getFocusManager().focusNode(this.workspace);
```
BenHenning, line 102: Is this covered by the tests from https://github.com/google/blockly-keyboard-experimentation/pull/625? If not, is it possible to test this (either in core or there; we have a bunch of tests now that verify `focusNode` behavior outside of `FocusManager`'s tests)?

microbit-matt-hillsdon:
Just a quick comment to say that it would be great to get this one in.

core/comments/delete_comment_bar_button.ts
gonfunko: Added a test to verify this.

BenHenning (approved):

PR #9202, opened 2025-07-03
feat: Use pure CSS for selection highlights

### LINKED ISSUE #8942: Use CSS for selection state
### Check for duplicates

### Problem

Selection is currently done using a custom SVG filter: https://github.com/google/blockly/blob/ada01da23ce84f359436c06181ca4ba41f5253c6/core/renderers/zelos/constants.ts#L675

This makes customization difficult especially when combined with other possible states (such as active focus).

### Request

It would be preferable for selection to be styled using pure CSS such that it can be combined with other possible states.

### Alternatives considered

_No response_

### Additional context

This has been explored in #8875 and it's not obvious exactly how to solve it. Some possibilities that have been explored as well as discussed with @rachel-fenichel:
- Use a stroke to simulate a border highlight. This causes a cutoff due to the way blocks are stacked atop each other.
- Expand the height of a block so that its stroke wouldn't be cut off, however this could cause a visual jittering that may be distracting when switching between blocks (especially with keyboard navigation).
- Create and maintain a custom highlight path that always renders after the block stack (and thus on top per the SVG spec) where `blocklySelected` can then apply a stroke property and control visibility of this special highlight element. It's still tricky, however, as this element requires custom lifecycle management and it can't be a child of the block's `g` element (which means `blocklySelected` has to get applied to a different element tree than it does today).

BenHenning:
## The basics

### Resolves

Fixes #8942

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -55,7 +55,6 @@ import type {IPathObject} from './renderers/common/i_path_object.js';
 import * as blocks from './serialization/blocks.js';
 import type {BlockStyle} from './theme.js';
 import * as Tooltip from './tooltip.js';
-import {idGenerator} from './utils.js';
 import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
 import {Rect} from './utils/rect.js';
@@ -211,9 +210,6 @@ export class BlockSvg
 
     // Expose this block's ID on its top-level SVG group.
     this.svgGroup.setAttribute('data-id', this.id);
-
-    // The page-wide unique ID of this Block used for focusing.
-    svgPath.id = idGenerator.getNextUniqueId();
 
     this.doInit_();
   }
--- a/core/renderers/common/path_object.ts
+++ b/core/renderers/common/path_object.ts
@@ -10,6 +10,7 @@ import type {BlockSvg} from '../../block_svg.js';
 import type {Connection} from '../../connection.js';
 import {RenderedConnection} from '../../rendered_connection.js';
 import type {BlockStyle} from '../../theme.js';
+import {idGenerator} from '../../utils.js';
 import {Coordinate} from '../../utils/coordinate.js';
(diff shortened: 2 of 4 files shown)


BenHenning:
This is a possible solution to #8942 that is a completely different approach from those. It still relies on an extra path structure (not unlike option (3) in the issue and current highlights), except it uses it as a clip path. This is a clever solution I learned from https://alexwlchan.net/2021/inner-outer-strokes-svg/ that essentially produces an inner-only stroke so long as the stroke width is doubled.

This actually works _very_ well, producing exactly the size of stroke we want without the block overlaps causing any issues.

However, there's now a small 1px gap due to a 2px reduction in size (before blocks would have a 1px stroke border on both sides, resulting in them overlapping by 1px) in Zelos. I haven't yet figured out how to eliminate this gap since it requires changes to block placement or path generation (not sure which yet).

Separately, I put this in common since the idea of using simple CSS for tweaking selection highlight behavior probably means we want the same solution for all renderers. This seems to look correct in Thrasos, but quite wrong in Geras (I'm assuming because of the offset dark path, but I haven't dug yet).

@rachel-fenichel I assume you are likely to have strong opinions about this--could you PTAL when you are next able to? It'd be helpful to get some thoughts on:
- Whether this is a sufficiently clean approach.
- Whether we should isolate this to Zelos or continue with the approach to generalize it.
- How to eliminate the 1px gap that exists now due to the default stroke borders also inherently becoming inner borders with the new clipping in place.

Edit: Also, I realized while working on this PR that Thrasos and Geras both have their CSS-based selection highlights partially cut off at the bottom of the blocks, as well, since they rely on stroke. I think it's just move noticeable in Zelos for whatever reason, so it may not be unreasonable to just remove the filter bits from Zelos and move it over to the exact same CSS selection highlight and just be fine with the slight cut-off. The clip approach introduced here ensures an even border around the block which does look much nicer, but there's a fair amount of work remaining to get it working properly for all three renderers.

PR #9204, opened 2025-07-04
revert: "fix: Auto close drop-down divs on lost focus"

### LINKED ISSUE #9203: Recent serious focus regressions
1. Closing the context menu with escape or via an action no longer returns focus to the the block. 
    - Only actions like duplicate that explicitly focus a block are usable on the context menu via keyboard now.
1. Confirming or cancelling a field edit (e.g. text/number) via enter/escape no longer returns focus to the block

In both cases focus moves to the body.

Both of these can be seen in the playground by clicking first then using keyboard to confirm/cancel, but the most serious impact is to users of the keyboard navigation plugin. [My deployment using the latest Blockly can be found here](https://blockly-keyboard-experimentation.pages.dev/).

This was introduced by 4c78c1d4a31fd2474737a3068c11d11881ae954e.

cpcallen:
Reverts google/blockly#9175.

Fixes #9203.


microbit-matt-hillsdon:
LGTM. Tested via blockly dev server and keyboard nav demo npm linked with a local build of this branch. I'd already tested MakeCode with an equivalent git revert.

cpcallen:
N.B. I have enabled auto-squash on this PR.

maribethb (approved):

PR #9211, opened 2025-07-07
fix: Improve workspace comment keyboard navigation behavior. 

gonfunko:
## The basics

## The details
This PR addresses two points of feedback raised in https://github.com/google/blockly-keyboard-experimentation/pull/625:

* The workspace comment textarea has its tabindex set to -1, preventing tabbing into it. It remains accessible via keyboard navigation.
* When workspace comments are focused, they will be scrolled into view on the workspace.

### DIFF
--- a/core/comments/comment_editor.ts
+++ b/core/comments/comment_editor.ts
@@ -52,6 +52,7 @@ export class CommentEditor implements IFocusableNode {
       dom.HTML_NS,
       'textarea',
     ) as HTMLTextAreaElement;
+    this.textArea.setAttribute('tabindex', '-1');
     dom.addClass(this.textArea, 'blocklyCommentText');
     dom.addClass(this.textArea, 'blocklyTextarea');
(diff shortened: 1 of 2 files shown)


maribethb (approved):
lgtm when tests pass. though i don't think the current approach to scrolling into view is very nice, that is a long term problem to fix.

do you think we need an interface for "Scroll-into-view-able"?

or should these components do this automatically in their "onNodeFocus" methods? is there a time when a node is focused that we *wouldn't* want to scroll them into view?

these are just rhetorical questions for now. i'll file an issue for them.

gonfunko:
Thanks! I don't think the test failures are actually related to this PR - they pass locally, and nothing in this would make Blockly.getMainWorkspace() not be a function.

maribethb:
> Thanks! I don't think the test failures are actually related to this PR - they pass locally, and nothing in this would make Blockly.getMainWorkspace() not be a function.

fair, i didn't investigate what the failures actually were

PR #9216, opened 2025-07-08
fix: Auto-close widget divs on lost focus

### LINKED ISSUE #563: Ephemeral focus issue clicking outside injection div

BenHenning:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/563

### Proposed Changes

This expands the functionality introduced in #9213 to also include widget divs.

### Reason for Changes

MakeCode makes use of widget div in several field editors, so the issues described in https://github.com/google/blockly-keyboard-experimentation/issues/563 aren't fully mitigated with #9213 alone.

This PR essentially adds the same support for auto-closing as drop-down divs now have, and enables this functionality by default.

Note the drop-down div change: it was missed in #9123 that the API change for drop-down div's `show` function is actually API-breaking, so this updates that API to be properly backward compatible (and reverts one test change that makes use of it).

The `FocusManager` change actually corrects an implementation issue from #9123: not updating the tracked focus status before calling the callback can result in focus being inadvertently restored if the callback triggers returning focus due to a lost focus situation. This was wrong for drop-down divs, too, but it's harder to notice there because the dismissal of the drop-down div happens on a timer (which means there's sufficient time for `FocusManager`'s state to correct prior to attempting to return from the ephemeral focus state).

Demonstration of fixed behavior (since the inline number editor uses a widget div):

[Screen recording 2025-07-08 2.12.31 PM.webm](https://github.com/user-attachments/assets/7c3c7c3c-224c-48f4-b4af-bde86feecfa8)

### Test Coverage

New widget div tests have been added to verify the new parameter and auto-close functionality.

The `FocusManager` test was updated to account for the new, and correct, behavior around the internal tracked ephemeral focus state.

Note that some `tabindex` state has been clarified and cleaned up in the test index page and `FocusManager`. It's fine (and preferable) for ephemeral-used elements to always be focusable rather than making them dynamically so (which avoids state bleed across test runs which was happening one of the new tests).

https://github.com/google/blockly-keyboard-experimentation/pull/649 includes additional tests for validating widget behaviors.

### Documentation

No new documentation should be needed here--the API documentation changes should be sufficient.

One documentation update was made in `dropdowndiv.ts` that corrects the documentation parameter ordering.

### Additional Information

Nothing further to add.

### DIFF
--- a/core/dropdowndiv.ts
+++ b/core/dropdowndiv.ts
@@ -346,8 +346,8 @@ function showPositionedByRect(
     secondaryX,
     secondaryY,
     manageEphemeralFocus,
-    autoCloseOnLostFocus,
     opt_onHide,
+    autoCloseOnLostFocus,
   );
 }
 
@@ -382,8 +382,8 @@ export function show<T>(
   secondaryX: number,
   secondaryY: number,
   manageEphemeralFocus: boolean,
-  autoCloseOnLostFocus: boolean,
   opt_onHide?: () => void,
+  autoCloseOnLostFocus?: boolean,
 ): boolean {
   owner = newOwner as Field;
   onHide = opt_onHide || null;
--- a/core/dropdowndiv.ts
+++ b/core/dropdowndiv.ts
@@ -366,9 +366,9 @@ function showPositionedByRect(
  * @param primaryY Desired origin point y, in absolute px.
  * @param secondaryX Secondary/alternative origin point x, in absolute px.
  * @param secondaryY Secondary/alternative origin point y, in absolute px.
- * @param opt_onHide Optional callback for when the drop-down is hidden.
  * @param manageEphemeralFocus Whether ephemeral focus should be managed
  *     according to the widget div's lifetime.
+ * @param opt_onHide Optional callback for when the drop-down is hidden.
  * @param autoCloseOnLostFocus Whether the drop-down should automatically hide
  *     if it loses DOM focus for any reason.
  * @returns True if the menu rendered at the primary origin point.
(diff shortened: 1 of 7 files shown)


BenHenning (commented):
Self-checked and fixed a few things, and added a couple more tests.

BenHenning:
FYI: I'm going to update https://github.com/google/blockly-keyboard-experimentation/pull/649 to include tests for the field editor to integrate test the widget changes, and then I'm going to try and verify all samples that might be affected by both widget and drop-down div changes.

microbit-matt-hillsdon:
Did a quick test of this in MakeCode and it looks promising. Will test further tomorrow (don't block merge on my feedback).

Fields that use both widget + dropdown only close widget but I think that's probably on us to sort as it's the widget that takes the focus. It seems to do no real harm and we can likely address in MakeCode. You might have similar scenarios (maybe in other direction?) in samples.

BenHenning:
> Did a quick test of this in MakeCode and it looks promising. Will test further tomorrow (don't block merge on my feedback).
> 
> Fields that use both widget + dropdown only close widget but I think that's probably on us to sort as it's the widget that takes the focus. It seems to do no real harm and we can likely address in MakeCode.

Thanks for the quick check @microbit-matt-hillsdon!

Drop-downs + widgets have been historically tricky with ephemeral focus, and this definitely adds to that. It may be the case you simply would turn off the auto behavior for those situations, I suspect, and handle it manually (which isn't ideal since catching all the cases where focus is lost is annoying and precisely why `FocusManager` is doing it).

If there's an API change we can make to simplify this, please do let me know.

gonfunko (approved):

Issue #9221, opened 2025-07-09
Rethink ephemeral focus changes that relate to dropdown/widget div


microbit-matt-hillsdon:
We've been testing the widget/dropdown div changes from #9213 and #9216 and unfortunately we think there is too much breakage in MakeCode without a clear route to address for Friday's release.

Please can you revert the widget div PR and consider carefully whether to also revert the dropdown div PR? We lean towards reverting both and reconsidering the approach.

This will reinstate focus quirks for mixed mode scenarios where you click outside the workspace when a field edit is in progress and then tab back into the toolbox/workspace. Clicking on the workspace or in MakeCode, using Cmd+B, 4 will fix it. We prefer a mixed mode bug with a workaround to a core keyboard nav flow bug.

## Problems in blockly-samples field slider

Mainly just to illustrate it's not entirely a MakeCode thing and may affect others.

1. Drag out e.g. this block

<img width="118" height="56" alt="Image" src="https://github.com/user-attachments/assets/5ce432c2-6663-4f2d-9021-7ec29c945f31" />

2. Click to edit the field
3. Click in the number input
4. Enter to confirm

Note that the block is not selected when it normally would be. This is because focus was returned when you clicked outside the dropdown even though a widget was still open.

This is caused by #9213 

You can see the problem more clearly by adding the keyboard nav plugin to the test page and using keyboard interactions except step 3.
 
## Problems in MakeCode

They are essentially the same but more prominent and with significant impact on the keyboard navigation feature. We have several field editors which augment number field with a dropdown part to implement a combobox experience or an associated slider similar to field-slider but where the input can also be interacted with.

For keyboard navigation we use up/down to move between these two parts which works well.

But now moving away from the widget part removes the widget part entirely leaving a stranded dropdown. The user experience is that they cannot move back up to the widget.

If the user does use escape to close the dropdown part that they're stuck in then focus is not correctly returned and we end up with passive focus on the field.

This is caused by #9216 (because it's the widget taking ephemeral focus). 

### Example: combo box

https://github.com/user-attachments/assets/eac1b289-6ab5-40b6-9f61-58607561b7ba

### Example: slider 

https://github.com/user-attachments/assets/c7473c7c-6b0c-42e1-8c36-00d4458250a2

## Potential solutions

We've tried showing the widget part with autoCloseOnFocusLost false:
- this means the widget doesn't close which reinstates the ability to move back and forth
- pressing enter in the dropdown part still results in passive focus on the field
[...]

PR #9222, opened 2025-07-09
fix: Revert drop down and widget div PRs

BenHenning:
## The basics

- [x] ~I~ @microbit-matt-hillsdon and @microbit-robert [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes part of #9221 (or, rather, puts things back into a less broken state)

### Proposed Changes

This reverts #9213 and #9216.

### Reason for Changes

The "fixed" world is more broken than before, or at least the breakages from before are better understood.

Fundamentally, there are several classes of edge cases for ephemeral focus with widget and drop-down divs that may not have single solutions. A more carefully thought out solution is required at this point, and there isn't sufficient time to safely land such a change ahead of the 12.2 release. It's preferable to revert the attempts at fixing these classes of issues since the previous breakages were harder to hit than the new ones.

#9221 is tracking longer term fixes.

### Test Coverage

N/A -- This is a revert.

### Documentation

N/A -- This is a revert.

### Additional Information

This PR was produced with two `git revert` commands and has no other changes.

https://github.com/google/blockly-keyboard-experimentation/issues/563 will need to be reopened after this PR merges since the problems that issue tracked will be reintroduced.


BenHenning:
PTAL @gonfunko but also please hold on merging until @microbit-matt-hillsdon has a chance to confirm the revert.

core/dropdowndiv.ts
```diff
@@ -366,11 +354,9 @@ function showPositionedByRect(
  * @param primaryY Desired origin point y, in absolute px.
  * @param secondaryX Secondary/alternative origin point x, in absolute px.
  * @param secondaryY Secondary/alternative origin point y, in absolute px.
+ * @param opt_onHide Optional callback for when the drop-down is hidden.
```
gonfunko, line 357: Do we want to re-reorder this to be the last arg?

core/dropdowndiv.ts
```diff
@@ -719,6 +693,7 @@ export function hideWithoutAnimation() {
     onHide();
     onHide = null;
   }
+  clearContent();
```
gonfunko, line 696: Is reordering this worth keeping independent of the other reverted changes?

gonfunko (approved):

core/dropdowndiv.ts
BenHenning: The reordering fixes an actual new issue introduced by the ephemeral changes in this PR, and doesn't seem to cause problems with ephemeral focus prior to these changes (even though it's logically in the wrong place).

It's probably safer to keep this part unchanged since the previous state of the divs was better tested and understood without any changes from these PRs.

core/dropdowndiv.ts
BenHenning: Probably, but keeping as-is just to keep the revert true and unchanged. We're likely to want to reapply these changes in the future.

microbit-matt-hillsdon:
Looks good, thanks.

I tested MakeCode locally linked to a checkout of this PR and as expected we have https://github.com/google/blockly-keyboard-experimentation/issues/563 back but the combined widget/dropdown div field editors are working well again for keyboard users.

We also speculatively tested an equivalent pair of git reverts earlier today and all seemed well.

I've also updated [our preview deployment of MakeCode to include this PR](https://preview.review-pxt.pages.dev/#editor) so you can try it there too (enable Keyboard Controls via settings cog or first tab stop). Simple test program as JavaScript:

```javascript
input.onButtonPressed(Button.A, function () {
    led.plot(2, 3)
    basic.pause(500)
})
```

BenHenning:
Thanks @microbit-matt-hillsdon and @microbit-robert for help with the testing.

Going ahead and merging since this puts Blockly back to an expected good (enough) state for the release.

PR #9227, opened 2025-07-10
feat: Add support for keyboard navigation to/from block comments.

### LINKED ISSUE #9166: add keyboard listeners to finish block comment editing
### Check for duplicates

### Problem

_No response_

### Request

in #9154 we added keyboard event listeners for escape and ctrl+command keys to finish editing a comment and move focus to the parent node. we should do the same for block comments. less clear to me what the parent node for block comments is - the block? the comment icon?

if we also add a shortcut for something like "enter" to start editing a comment, we should add the same thing to block comments as well. this may be done as part of https://github.com/google/blockly-keyboard-experimentation/issues/614

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9166 and part of https://github.com/google/blockly-keyboard-experimentation/issues/636.

### Proposed Changes
This PR adds support for keyboard navigation to/from block comments. To that end:

* `TextInputBubble` is refactored to reuse `CommentEditor` (also used by workspace comments) to handle text editing
* Bubbles have been updated to close themselves when Escape is pressed
* Several additional navigation policies have been added.

A corresponding change will be needed in keyboard-experiment, and tests will be added there.

### DIFF
--- a/core/bubbles/bubble.ts
+++ b/core/bubbles/bubble.ts
@@ -11,6 +11,7 @@ import {getFocusManager} from '../focus_manager.js';
 import {IBubble} from '../interfaces/i_bubble.js';
 import type {IFocusableNode} from '../interfaces/i_focusable_node.js';
 import type {IFocusableTree} from '../interfaces/i_focusable_tree.js';
+import type {IHasBubble} from '../interfaces/i_has_bubble.js';
 import {ISelectable} from '../interfaces/i_selectable.js';
 import {ContainerRegion} from '../metrics_manager.js';
 import {Scrollbar} from '../scrollbar.js';
@@ -107,6 +108,7 @@ export abstract class Bubble implements IBubble, ISelectable, IFocusableNode {
     protected anchor: Coordinate,
     protected ownerRect?: Rect,
     overriddenFocusableElement?: SVGElement | HTMLElement,
+    protected owner?: IHasBubble & IFocusableNode,
   ) {
     this.id = idGenerator.getNextUniqueId();
     this.svgRoot = dom.createSvgElement(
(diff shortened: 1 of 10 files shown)


core/bubbles/bubble.ts
```diff
@@ -106,6 +108,7 @@ export abstract class Bubble implements IBubble, ISelectable {
     protected anchor: Coordinate,
     protected ownerRect?: Rect,
     overriddenFocusableElement?: SVGElement | HTMLElement,
+    protected owner?: IHasBubble & IFocusableNode,
```
RoboErikG, line 112: Nit: add new parameter to docs

core/bubbles/textinput_bubble.ts
```diff
@@ -130,58 +123,6 @@ export class TextInputBubble extends Bubble {
...
-
-    this.bindTextAreaEvents(textArea);
-
-    return inputRoot;
-  }
-
-  /** Binds events to the text area element. */
-  private bindTextAreaEvents(textArea: HTMLTextAreaElement) {
-    // Don't zoom with mousewheel; let it scroll instead.
-    browserEvents.conditionalBind(textArea, 'wheel', this, (e: Event) => {
```
RoboErikG, line 173: Verify that we either don't need this or add it to the comment editor

core/bubbles/bubble.ts
gonfunko: Done.

core/bubbles/textinput_bubble.ts
gonfunko: Good catch, factored it out to the comment editor.

RoboErikG (approved):

PR #9228, opened 2025-07-10
fix: Focus the first element in flyouts.

### LINKED ISSUE #9103: Flyout only default focuses blocks
### Check for duplicates

### Description

Navigating to the flyout with tab navigation with the keyboard experimentation plugin results in only the first block of the flyout being auto-focused rather than the first item.

### Reproduction steps

1. Check out the latest versions of core Blockly and the keyboard experimentation plugin.
2. Link core Blockly to the keyboard experimentation plugin.
3. Start the keyboard experimentation plugin test environment playground.
4. Tab navigate to the toolbox.
5. Use arrow keys to navigate to 'Variables'.
6. Create a variable.
7. Refocus the toolbox (there's currently a bug wherein focus is completely lost after creating a variable).
8. Arrow key navigate back to 'Variables'.
9. Press 'right'.
10. Observe that the 'set variable' block is selected first rather than 'Create variable'.

The same can be observed via the 'misc' category, but that requires a fix for https://github.com/google/blockly-keyboard-experimentation/issues/241 in order to navigate (since otherwise it's impossible to navigate to a category's flyout).

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

Chrome desktop

gonfunko:
## The basics

### Resolves

Fixes #9103

### Proposed Changes
This PR returns the first focusable node of a flyout workspace when `getRestoredFocusableNode()` is called, regardless of whether that element is a block or some other type of entity.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -2729,12 +2729,15 @@ export class WorkspaceSvg
       const flyout = this.targetWorkspace?.getFlyout();
       if (this.isFlyout && flyout) {
         // Return the first focusable item of the flyout.
-        for (const flyoutItem of flyout.getContents()) {
-          const elem = flyoutItem.getElement();
-          if (isFocusableNode(elem) && elem.canBeFocused()) {
-            return elem;
-          }
-        }
+        return (
+          flyout
+            .getContents()
+            .find((flyoutItem) => {
+              const element = flyoutItem.getElement();
+              return isFocusableNode(element) && element.canBeFocused();
+            })
+            ?.getElement() ?? null
+        );
       }
       return this.getTopBlocks(true)[0] ?? null;
     } else return null;
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -2726,6 +2726,16 @@ export class WorkspaceSvg
     previousNode: IFocusableNode | null,
   ): IFocusableNode | null {
     if (!previousNode) {
+      const flyout = this.targetWorkspace?.getFlyout();
+      if (this.isFlyout && flyout) {
+        // Return the first focusable item of the flyout.
+        for (const flyoutItem of flyout.getContents()) {
+          const elem = flyoutItem.getElement();
+          if (isFocusableNode(elem) && elem.canBeFocused()) {
+            return elem;
+          }
+        }
+      }
       return this.getTopBlocks(true)[0] ?? null;
     } else return null;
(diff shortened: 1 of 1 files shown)


core/workspace_svg.ts
```diff
@@ -2726,6 +2726,16 @@ export class WorkspaceSvg
     previousNode: IFocusableNode | null,
   ): IFocusableNode | null {
     if (!previousNode) {
+      const flyout = this.targetWorkspace?.getFlyout();
```
maribethb, line 2729: I know it's been said before, but

😵‍💫

core/workspace_svg.ts
```diff
@@ -2726,6 +2726,16 @@ export class WorkspaceSvg
...
+      if (this.isFlyout && flyout) {
+        // Return the first focusable item of the flyout.
+        for (const flyoutItem of flyout.getContents()) {
```
maribethb:
```
flyout.getContents().find((flyoutItem) => {
  const elem = flyoutItem.getElement();
  return isFocusableNode(elem) && elem.canBeFocused());
});
```

core/workspace_svg.ts
```diff
@@ -2726,6 +2726,16 @@ export class WorkspaceSvg
...
+          const elem = flyoutItem.getElement();
+          if (isFocusableNode(elem) && elem.canBeFocused()) {
+            return elem;
+          }
+        }
+      }
       return this.getTopBlocks(true)[0] ?? null;
```
maribethb, line 2742: there's probably an argument to be made that this is too specific even for non-flyout workspaces. what if a workspace comment is first? seems like workspaces could really do with a "getTopLevelNavigableThingies" type of method. Which I think would also have been used in the Stack navigation next/prev actions instead of hardcoding blocks + comments. Just food for thought, don't have to change this here

maribethb (approved):

core/workspace_svg.ts
gonfunko: Yeah...

core/workspace_svg.ts
gonfunko: Done.

core/workspace_svg.ts
gonfunko: Agreed, there is `getTopBoundedElements()`, but the typing doesn't work without casting. Leaving for now but will keep this in mind.

PR #9231, opened 2025-07-11
fix: Fix order of arguments in `IVariableMap.createVariable()`.

### LINKED ISSUE #9218: bug: `IVariableMap.createVariable` type definition is wrong
### Check for duplicates

### Description

The type definition for `IVariableMap.createVariable` is incorrect.
The arguments `id` and `type` are reversed.

https://github.com/google/blockly/blob/bea183d85da6465cd0511fc48f2e612b99b6c1ff/core/interfaces/i_variable_map.ts#L46

https://github.com/google/blockly/blob/bea183d85da6465cd0511fc48f2e612b99b6c1ff/core/variable_map.ts#L221

### Reproduction steps

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9218

### Proposed Changes
This PR flips the names of the second and third arguments to `IVariableMap.createVariable()` to reflect the ordering used in the built-in `VariableMap` class.

### NOT-QUITE-BREAKING CHANGE
This PR changes the API, but (a) the previous interface was incorrect relative to the actual implementation provided, and the types of the arguments did not change. As a result, this will not result in build failures, but if you were using the old argument order your code wasn't doing what you intended.

### DIFF
--- a/core/interfaces/i_variable_map.ts
+++ b/core/interfaces/i_variable_map.ts
@@ -43,7 +43,7 @@ export interface IVariableMap<T extends IVariableModel<IVariableState>> {
    * Creates a new variable with the given name. If ID is not specified, the
    * variable map should create one. Returns the new variable.
    */
-  createVariable(name: string, id?: string, type?: string | null): T;
+  createVariable(name: string, type?: string, id?: string | null): T;
 
   /* Adds a variable to this variable map. */
   addVariable(variable: T): void;


cpcallen (approved):
I think it might be worth adding a "NOT-QUITE-BREAKING CHANGE" note to the PR description, noting this is technically an API change but (a) the interface did not previously reflect the implementation, and (b) as parameters' types are not changing it (perhaps unfortunately) cannot cause any compilation failures.

Issue #9217, opened 2025-07-08
Fix and re-enable the 'Open toolbox categories' test and move them into their own test runner


RoboErikG:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Description

The 'Open toolbox categories' suite of tests was disabled as dragging most of the blocks out is currently failing. These take long enough to run that they also need to be moved into their own test runner so we can run them less frequently than other tests.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

RoboErikG:
Tests should be passing after samples gets published this week. Remaining work is to pull out the open toolbox categories suite into its own test runner and configure browser tests to run automatically. Can probably run most browser tests as part of pre-submit and do the open toolbox categories suite once a week since it takes longer than all the rest.

PR #9238, opened 2025-07-14
fix: Fix toolbox categories tests

### LINKED ISSUE #8953: Browser tests fail with "element not interactable"

RoboErikG:
This fixes the the toolbox categories tests except for dragging out the four empty statement blocks, which is fixed by https://github.com/google/blockly-samples/pull/2580

Tests will continue to fail until the latest version of samples is published.

## The basics

### Resolves

Fixes #8953 and half of #9217 

### Proposed Changes

Fixes and re-enables the toolbox categories test.

### Reason for Changes

Tests were broken.

### DIFF
--- a/tests/browser/test/toolbox_drag_test.mjs
+++ b/tests/browser/test/toolbox_drag_test.mjs
@@ -179,9 +179,9 @@ async function openCategories(browser, categoryList, directionMultiplier) {
   chai.assert.equal(failureCount, 0);
 }
 
-// TODO (#9217) These take too long to run and are very flakey. Need to find a
-// better way to test whatever this is trying to test.
-suite.skip('Open toolbox categories', function () {
+// TODO (#9217) These take too long to run and are very flakey. Need to pull
+// these out into their own test runner.
+suite('Open toolbox categories', function () {
   this.timeout(0);
 
   test('opening every toolbox category in the category toolbox in LTR', async function () {
--- a/tests/browser/test/toolbox_drag_test.mjs
+++ b/tests/browser/test/toolbox_drag_test.mjs
@@ -11,7 +11,7 @@
 import * as chai from 'chai';
 import {Key} from 'webdriverio';
 import {
-  dragBlockTypeFromFlyout,
+  getBlockTypeFromCategory,
   getCategory,
   PAUSE_TIME,
   screenDirection,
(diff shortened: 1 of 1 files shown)


gonfunko (approved):

cpcallen:
N.B. that:

* this PR was merged rather than squashed, and
* the merge commit (7e47bf4) did not follow the conventional commits convention, and
* one of the constituent commits (55bd1af) also did not follow the conventional commits convention.

This may result in some problems with the automatically-generated release notes next time we release.

Making a merge commits instead of squashing is an easy mistake to make (as unfortunately GitHub remembers one's previous choice here, rather than remembering a per-PR selection, and we _do_ make some merge commits deliberately during release).  It is likewise easy to fail to notice that the merge commit is not conventional, since normally squash commits take the (hopefully conventional) PR title automatically.  Unfortunately both of the above mistakes cannot be prevented by CI, since it can't check commit messages (or types) before the commit is made.

But AIUI Ci _is_ supposed to be checking both PR titles and the commit messages of the constituent commits, and I'm not quite sure why the unconventional commit in the PR branch was not flagged.  @maribethb: do we no longer check constituent commits?

maribethb:
This won’t affect release notes and won’t have any negative impact.

For core, release notes are generated based on the labels that are added to
PRs by the conventional label action. This PR was labeled correctly because
the labels are based on the PR title. We select the version number for a
release manually in core so this also does not impact that process.

I believe the lint action only checks the PR title unless there is only one
commit. If there is one commit it will check that that commit does follow
conventional commits.

In samples, it’s a little different because the version numbers are
determined automatically. But even in that case it’s difficult to break
things because if you merge a PR, as long as the constituent commits were
titled correctly, the release would be fine. If however everything were
labeled as “fix” except the one commit that should be “feat” or breaking
change, then the version number could be incorrect if there were no other
correctly labeled feat or breaking change commits.

TLDR: it’s fine, this does not impact anything.

Issue #9126, opened 2025-06-02
`Flyout.setAutoClose(false)` does not work as expect


QianNangong:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

I'm working on a project that need the flyout to be always shown, but I have two problems on it.

1. If I set `flyout.setAutoClose(false)`, I will be there unless I drag a block to workspace and it just close.
2. If I make flyout shown, if I click the blank space inside flyout and then click the blank space inside the workspace, it closes.

### Reproduction steps

Common Steps:
1. Setup workspaces
```ts
const ws = Blockly.inject(blocklyDiv, {toolbox});

let flyout;
if (flyout = ws.getToolbox()?.getFlyout() as Blockly.Flyout) {
  flyout.setAutoClose(false);
}
```
3. Open flyout

First issue:
1. Drag any block into workspace, then the flyout is hidden unexpectly

Second issue:
1. Click the blank space in the flyout
2. Click the blank space in the workspace, then the flyout is hidden unexpectly

### Stack trace

```shell

```

### Screenshots

First Issue:
![Image](https://github.com/user-attachments/assets/52524dda-5852-40fc-b8a1-faace5c9a9c1)
Second Issue:
![Image](https://github.com/user-attachments/assets/188db715-777a-44da-ab85-4880cb41830e)

### Browsers

Chrome desktop

PR #9245, opened 2025-07-17
fix: Make non-autoclosing flyouts stay open.

gonfunko:
## The basics

### Resolves

Fixes #[9126](https://github.com/google/blockly/issues/9126)

### Proposed Changes
This PR fixes the behavior for non-autoclosing flyouts. As part of the recent focus changes, flyouts were more or less forced to always autoclose. This is because `Toolbox.clearSelection()` winds up calling `this.setSelectedItem(null)` -> `this.updateFlyout(*, null)`, and the latter unconditionally hides the flyout if the new item is `null`. Now, in situations where the toolbox is losing focus, its `autoHide` method is called, which gates the call to `clearSelection()` on a check for whether the toolbox's flyout is in autoclose mode. The explicit call to `Flyout.autoHide` in `WorkspaceSvg` has been removed because it's redundant as the prior call to `autoHide()` on the toolbox will wind up invoking it.

### DIFF
--- a/core/toolbox/toolbox.ts
+++ b/core/toolbox/toolbox.ts
@@ -22,10 +22,7 @@ import '../events/events_toolbox_item_select.js';
 import {EventType} from '../events/type.js';
 import * as eventUtils from '../events/utils.js';
 import {getFocusManager} from '../focus_manager.js';
-import {
-  isAutoHideable,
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):
Thanks for adding tests.

PR #9250, opened 2025-07-23
Fix: pointercancel event is not handled

### LINKED ISSUE #9249: Fix not handled pointercancel event
### Check for duplicates

### Description

When I use the multi-touch gesture on a mobile device, there may be a 'pointercancel' event fired under some situations. In this situation, the Gesture instance is not disposed because the code is waiting for the 'pointerup' event, and the following events are all affected by the Gesture.

We can make the 'pointerup' event listener to also listen to the 'pointercancel' event, and I will try to commit a PR to fix this problem.

### Reproduction steps

1. Try to use two fingers to pinch the workspace and hold the fingers on the screen
2. Try to fire a 'pointercancel' event, for example, move the two fingers very close
3. Release the fingers, and try to move the workspace by a finger, you can find that the moving action is abnormal

### Stack trace

```shell
None
```

### Screenshots

https://github.com/user-attachments/assets/972f969e-182d-4b7f-861a-0e5b22888cae

### Browsers

_No response_

nianxy:
## The basics

### Resolves

Fixes 
#9249 

### Proposed Changes

Listen to the 'pointercancel' event

### Reason for Changes

The 'pointercancel' event is not handled, which leads to the abnormal

### Test Coverage

Manually tested.

### Documentation

None

### Additional Information

None

### DIFF
--- a/core/touch.ts
+++ b/core/touch.ts
@@ -46,6 +46,7 @@ export const TOUCH_MAP: {[key: string]: string[]} = {
   'mouseup': ['pointerup', 'pointercancel'],
   'touchend': ['pointerup'],
   'touchcancel': ['pointercancel'],
+  'pointerup': ['pointerup', 'pointercancel'],
 };
 
 /** PID of queued long-press task. */


maribethb:
Thank you for filing this PR and reporting the issue it goes with!

This is a creative approach that will essentially bind any registered `pointerup` events to also fire when a `pointercancel` event is received as well. But I think if a `pointercancel` event is received, we don't necessarily want to fire all the same events that would fire on a true `pointerup`. We should definitely cancel the in-progress gesture, but since this affects any event that is registered through `conditionalBind` we should be cautious as it could have unintended effects. The `pointercancel` event may be triggered when "The browser decides that the user started pointer input accidentally. This can happen if, for example, the hardware supports palm rejection to prevent a hand resting on the display while using a stylus from accidentally triggering events." ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event)) and if that were to happen it's possible we'd want to abort a move rather than finish it, for example.

cc @rachel-fenichel for more input on this because you have more experience with touch handling.

nianxy:
Yes, my changes are risky. 

My App users are children, and they often use their fingers to control the blocks on the pad. I have received many feedback  saying that the blocks sometimes move unexpectedly. I watched their screen records attached, and I'm sure the problem is caused by the unhandled `pointercancel` event.

In order to quickly fix this problem, I temporarily patched my code in my App. I listen to the `pointercancel` event and emit a `pointerup` event to the workspace when I receive a `pointercancel` event. After doing this about a week ago, I haven't received any report on this problem till now.

Since I haven't fully read the code of Blockly, I think my PR may be inconsiderate. If so, I'm looking forward to your advice on how to fix the problem more sensibly.

maribethb (approved):
Thank you so much for your patience. After discussing this internally with the team, we think this is ok to try given that this is how things work for mouse users already, and we don't have an established way to abort an in-progress gesture other than completing it. Thank you for contributing and testing the fix!

PR #9258, opened 2025-07-28
fix: Fix displaying HTML elements in `FieldDropdown`.

### LINKED ISSUE #9247: Problem using HTML elements in a dropdown field
### Check for duplicates

### Description

Using an HTML element in a dropdown field results in the warning, `"Can't get text for existing dropdown option. If you're using HTMLElement dropdown options in node, ensure you're using jsdom-global or similar."` and the use of the field's value rather than its `title`, `ariaLabel`, or `innerText` as is intended. The HTML element is not displayed.

The problem occurs in `trimOptions`, where [this ternary statement](https://github.com/google/blockly/blob/c92314d4101e1a9a8844552c34f0c317bbf1db07/core/field_dropdown.ts#L714) replaces the HTML element (`label`) with `{...label}`. The resulting object does not have type `HTMLElement`, so the option fails the test `option instanceof HTMLElement` in [`getText_`](https://github.com/google/blockly/blob/c92314d4101e1a9a8844552c34f0c317bbf1db07/core/field_dropdown.ts#L646).

The solution is probably to change the `false` outcome of the ternary statement from `{...label}` to `label`. The spread was previously necessary when we weren't strict about requiring image options to have all four properties. The addition of `isImageProperties` now requires this (see [`validateOptions`](https://github.com/google/blockly/blob/c92314d4101e1a9a8844552c34f0c317bbf1db07/core/field_dropdown.ts#L799), so the only thing that drops through are `HTMLElements`.

### Reproduction steps

See https://jsfiddle.net/rpbourret/18z6p2vf/32/ or run the following code:
```JS
Blockly.Blocks["headings_dropdown"] = {
  init: function() {
    var input = this.appendDummyInput()
        .appendField("heading");
    var options = [
        [h1, "H1VALUE"],
    ];
    input.appendField(new Blockly.FieldDropdown(options), "HEADING");
  }
};

const toolbox = {
  kind: "flyoutToolbox",
  contents: [
    {
      kind: "block",
      type: "headings_dropdown",
    },
  ],
}

const workspace = Blockly.inject("blocklyDiv", {
  toolbox: toolbox,
})
```

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9247

### Proposed Changes
This PR fixes `FieldDropdown` to not attempt to spread `HTMLElement` menu options, which caused later checks for `instanceof HTMLElement` to fail. This allows HTML elements to be properly displayed in dropdown fields.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -699,25 +699,30 @@ export class FieldDropdown extends Field<string> {
     prefix?: string;
     suffix?: string;
   } {
-    let hasImages = false;
+    let hasNonTextContent = false;
     const trimmedOptions = options.map((option): MenuOption => {
-      if (option === FieldDropdown.SEPARATOR) return option;
+      if (option === FieldDropdown.SEPARATOR) {
+        hasNonTextContent = true;
+        return option;
+      }
 
       const [label, value] = option;
       if (typeof label === 'string') {
         return [parsing.replaceMessageReferences(label), value];
       }
 
-      hasImages = true;
+      hasNonTextContent = true;
       // Copy the image properties so they're not influenced by the original.
       // NOTE: No need to deep copy since image properties are only 1 level deep.
       const imageLabel = isImageProperties(label)
         ? {...label, alt: parsing.replaceMessageReferences(label.alt)}
-        : {...label};
+        : label;
       return [imageLabel, value];
     });
 
-    if (hasImages || options.length < 2) return {options: trimmedOptions};
+    if (hasNonTextContent || options.length < 2) {
+      return {options: trimmedOptions};
+    }
 
     const stringOptions = trimmedOptions as [string, string][];
     const stringLabels = stringOptions.map(([label]) => label);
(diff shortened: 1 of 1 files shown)


core/field_dropdown.ts
```diff
@@ -699,25 +699,30 @@ export class FieldDropdown extends Field<string> {
     prefix?: string;
     suffix?: string;
   } {
-    let hasImages = false;
+    let hasNonTextContent = false;
     const trimmedOptions = options.map((option): MenuOption => {
-      if (option === FieldDropdown.SEPARATOR) return option;
+      if (option === FieldDropdown.SEPARATOR) {
+        hasNonTextContent = true;
```
maribethb, line 705: This change means that dropdowns with separators won't go through the trimming/prefixing process. Is that intended? It looks like those were pretty new so maybe it never worked before, but conceptually these kinds of lists (if they only contain text and separators) could still go through the trimming process so I'm not sure if we want this or not.

gonfunko: Correct; I believe they would have blown up at runtime, because without this change the `const stringOptions = trimmedOptions as [string, string][];` cast below is unsafe/inaccurate, as a separator is only a single string, not an array of two strings. The whole trimming thing feels very weird to me in general, but given that it's been around forever I realize we need to keep it.

maribethb: Ok got it! 

It's definitely weird but I was convinced of the utility by some i18n use cases. Basically it's easier to translate whole phrases so it's nice if they're all together in the options, but it's nicer to have the prefix factored out when a user is actually selecting an option.

maribethb (approved):

PR #9261, opened 2025-07-29
fix: Make workspace-multiselect plugin workable with Blockly v12

### LINKED ISSUE #103: cannot be used normally in Blockly 12
### Check for duplicates

### Description

error info: 
```
ERROR TypeError: Cannot read properties of null (reading 'canBeFocused')
    at FocusManager$$module$build$src$core$focus_manager.focusNode (blockly_compressed.js:4674:41)
    at Object.setSelected$$module$build$src$core$common [as setSelected] (blockly_compressed.js:402:62)
    at MultiselectControls.updateMultiselect (multiselect_controls.js:390:24)
    at multiselect.js:371:18
    at Set.forEach (<anonymous>)
    at Multiselect.eventListenerAll_ (multiselect.js:370:31)
```

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

HollowMan6:
## The basics

### Resolves

Fixes https://github.com/mit-cml/workspace-multiselect/issues/103

### Proposed Changes

Workspace-multiselect plugin is in the process of updating to Blockly v12:
- https://github.com/mit-cml/workspace-multiselect/pull/106

However, some changes are necessary to get Blockly working properly with the plugin. The fundamental issue here is that Blockly v12 is buggy at this stage to properly support a custom and dummy IDraggable (`MultiselectDraggable` for the plugin's case), as Blockly constantly wants to change the focus to the actual block via hardcoded `getFocusManager().focusNode` calls that are scattered all around the codebase. Setting `MultiselectDraggable` directly via `Blockly.common.setSelected` doesn't work as well, and we will need to call `Blockly.getFocusManager().updateFocusedNode` directly.

I'm sure the changes in this PR now will break some native Blockly features here and there. Feel free to modify on top of this PR directly, and I'll help test out to see if it still works with the plugin.

### Reason for Changes

Necessary changes to make workspace-multiselect plugin workable with Blockly v12

### Test Coverage

N/A

### Documentation

N/A

### Additional Information

https://groups.google.com/g/blockly/c/7RSn9cXAulA

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -300,9 +300,12 @@ export class BlockSvg
 
     const oldXY = this.getRelativeToSurfaceXY();
     const focusedNode = getFocusManager().getFocusedNode();
-    const restoreFocus = this.getSvgRoot().contains(
-      focusedNode?.getFocusableElement() ?? null,
-    );
+    let element = focusedNode?.getFocusableElement() ?? null;
+    // Make sure the element is of Node type
+    if (!(element instanceof Node)) {
+      element = null;
+    }
+    const restoreFocus = this.getSvgRoot().contains(element);
     if (newParent) {
       (newParent as BlockSvg).getSvgRoot().appendChild(svgRoot);
       // appendChild() clears focus state, so re-focus the previously focused
@@ -870,11 +873,12 @@ export class BlockSvg
     // If this block (or a descendant) was focused, focus its parent or
     // workspace instead.
     const focusManager = getFocusManager();
-    if (
-      this.getSvgRoot().contains(
-        focusManager.getFocusedNode()?.getFocusableElement() ?? null,
-      )
-    ) {
+    let element = focusManager.getFocusedNode()?.getFocusableElement() ?? null;
+    // Make sure the element is of Node type
+    if (!(element instanceof Node)) {
+      element = null;
+    }
+    if (this.getSvgRoot().contains(element)) {
       let parent: BlockSvg | undefined | null = this.getParent();
       if (!parent) {
         // In some cases, blocks are disconnected from their parents before
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -320,13 +320,18 @@ export class FocusManager {
       this.isUpdatingFocusedNode = true;
     }
 
-    // Double check that state wasn't desynchronized in the background. See:
-    // https://github.com/google/blockly-keyboard-experimentation/issues/87.
-    // This is only done for the case where the same node is being focused twice
-    // since other cases should automatically correct (due to the rest of the
-    // routine running as normal).
-    const prevFocusedElement = this.focusedNode?.getFocusableElement();
-    const hasDesyncedState = prevFocusedElement !== document.activeElement;
+    // // TODO: This will break multiselect plugin as it will make the
+    // // selected `MultiselectDraggable` (has already assigned to `this.focusedNode`)
+    // // to lose focus (get changed to the actual focused one managed by Blockly core).
+    // // Double check that state wasn't desynchronized in the background. See:
+    // // https://github.com/google/blockly-keyboard-experimentation/issues/87.
+    // // This is only done for the case where the same node is being focused twice
+    // // since other cases should automatically correct (due to the rest of the
+    // // routine running as normal).
+    // const prevFocusedElement = this.focusedNode?.getFocusableElement();
+    // const hasDesyncedState = prevFocusedElement !== document.activeElement;
+    const hasDesyncedState = false;
+
     if (this.focusedNode === focusableNode && !hasDesyncedState) {
       if (mustRestoreUpdatingNode) {
         // Reenable state syncing from DOM events.
--- a/core/layer_manager.ts
+++ b/core/layer_manager.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {getFocusManager} from './focus_manager.js';
 import type {IFocusableNode} from './interfaces/i_focusable_node.js';
 import {IRenderedElement} from './interfaces/i_rendered_element.js';
 import * as layerNums from './layers.js';
@@ -104,11 +103,14 @@ export class LayerManager {
   moveToDragLayer(elem: IRenderedElement & IFocusableNode) {
     this.dragLayer?.appendChild(elem.getSvgRoot());
 
-    if (elem.canBeFocused()) {
-      // Since moving the element to the drag layer will cause it to lose focus,
-      // ensure it regains focus (to ensure proper highlights & sent events).
-      getFocusManager().focusNode(elem);
-    }
+    // // TODO: This will break multiselect plugin, as multiselect plugin will
+    // // implicitly call this method as well, we should not change the focus to individual
+    // // block and move away from `MultiselectDraggable` at this time.
+    // if (elem.canBeFocused()) {
+    //   // Since moving the element to the drag layer will cause it to lose focus,
+    //   // ensure it regains focus (to ensure proper highlights & sent events).
+    //   getFocusManager().focusNode(elem);
+    // }
   }
 
   /**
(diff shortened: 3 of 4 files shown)


core/block_svg.ts
```diff
@@ -300,9 +300,12 @@ export class BlockSvg
...
     const oldXY = this.getRelativeToSurfaceXY();
     const focusedNode = getFocusManager().getFocusedNode();
-    const restoreFocus = this.getSvgRoot().contains(
-      focusedNode?.getFocusableElement() ?? null,
-    );
+    let element = focusedNode?.getFocusableElement() ?? null;
+    // Make sure the element is of Node type
+    if (!(element instanceof Node)) {
+      element = null;
+    }
```
cpcallen, line 307: AFAICT this check does nothing (and I'm surprised that TSC does not complain), because `IFocusableNode`'s `getFocusableElement` method is typed as returning `HTMLElement | SVGElement` and both `HTMLElement` and `SVGElement` inherit from `Node`, so the only time the body of the `if` statement will be executed is if `element` is already `null`.

If some focusable node's `.getFocusableElement()` returns something which is not an `instanceof Node` then it has violated the interface definition—unless I have greatly misunderstood something about how the DOM works, at any rate.

core/block_svg.ts
```diff
@@ -870,11 +873,12 @@ export class BlockSvg
...
-    if (
-      this.getSvgRoot().contains(
-        focusManager.getFocusedNode()?.getFocusableElement() ?? null,
-      )
-    ) {
+    let element = focusManager.getFocusedNode()?.getFocusableElement() ?? null;
+    // Make sure the element is of Node type
+    if (!(element instanceof Node)) {
+      element = null;
+    }
```
cpcallen, line 880: Ditto here.

core/focus_manager.ts
```diff
@@ -320,13 +320,18 @@ export class FocusManager {
...
+    // // selected `MultiselectDraggable` (has already assigned to `this.focusedNode`)
+    // // to lose focus (get changed to the actual focused one managed by Blockly core).
+    // // Double check that state wasn't desynchronized in the background. See:
+    // // https://github.com/google/blockly-keyboard-experimentation/issues/87.
+    // // This is only done for the case where the same node is being focused twice
+    // // since other cases should automatically correct (due to the rest of the
+    // // routine running as normal).
+    // const prevFocusedElement = this.focusedNode?.getFocusableElement();
+    // const hasDesyncedState = prevFocusedElement !== document.activeElement;
+    const hasDesyncedState = false;
```
cpcallen, line 333: The multiselect plugin will _definitely_ need to be able to operate without causing focus desynchronisation.

core/block_svg.ts
```diff
@@ -300,9 +300,12 @@ export class BlockSvg
...
     const focusedNode = getFocusManager().getFocusedNode();
-    const restoreFocus = this.getSvgRoot().contains(
-      focusedNode?.getFocusableElement() ?? null,
-    );
+    let element = focusedNode?.getFocusableElement() ?? null;
+    // Make sure the element is of Node type
+    if (!(element instanceof Node)) {
+      element = null;
+    }
+    const restoreFocus = this.getSvgRoot().contains(element);
```
cpcallen, line 308: It's possible that the existing check, reproduced here, is overly-broad: I note that `restoreFocus` will get set to true even if `this` merely _contains_ `focusedNode` (rather than _is_ `focusedNode`).

I also wonder whether the focus manipulation in `appendChild`, which this section of code is working around, could itself be either eliminated or made more adept, such that this workaround would not be needed at all.

core/layer_manager.ts
```diff
@@ -104,11 +103,14 @@ export class LayerManager {
...
-      // ensure it regains focus (to ensure proper highlights & sent events).
-      getFocusManager().focusNode(elem);
-    }
+    // // TODO: This will break multiselect plugin, as multiselect plugin will
+    // // implicitly call this method as well, we should not change the focus to individual
+    // // block and move away from `MultiselectDraggable` at this time.
+    // if (elem.canBeFocused()) {
+    //   // Since moving the element to the drag layer will cause it to lose focus,
+    //   // ensure it regains focus (to ensure proper highlights & sent events).
+    //   getFocusManager().focusNode(elem);
```
cpcallen, line 112: It's not obvious to me why moving an element to the drag layer will cause it to lose focus.  Maybe that is the problem that should be fixed, so this section of code is not needed.

core/layer_manager.ts
```diff
@@ -119,11 +121,14 @@ export class LayerManager {
...
-      getFocusManager().focusNode(elem);
-    }
+    // // TODO: This will break multiselect plugin, as multiselect plugin will
+    // // implicitly call this method as well, we should not change the focus to individual
+    // // block and move away from `MultiselectDraggable` at this time.
+    // if (elem.canBeFocused()) {
+    //   // Since moving the element off the drag layer will cause it to lose focus,
+    //   // ensure it regains focus (to ensure proper highlights & sent events).
+    //   getFocusManager().focusNode(elem);
+    // }
```
cpcallen, line 131: Ditto here.

core/utils/dom.ts
```diff
@@ -76,10 +76,16 @@ export function createSvgElement<T extends SVGElement>(
  */
 export function addClass(element: Element, className: string): boolean {
   const classNames = className.split(' ');
-  if (classNames.every((name) => element.classList.contains(name))) {
+  if (
+    classNames.every(
+      (name) => !!element.classList && element.classList.contains(name),
```
cpcallen, line 81: [MDN tells me](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) that `Element`'s `.classList` will always be a [`DOMTokenList`](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList), even if the element has no class attribute, so I cannot see why the nullish check might be needed here.

core/utils/dom.ts
```diff
@@ -90,7 +96,9 @@ export function addClass(element: Element, className: string): boolean {
  * @param classNames A string of one or multiple class names for an element.
  */
 export function removeClasses(element: Element, classNames: string) {
-  element.classList.remove(...classNames.split(' '));
+  if (element.classList) {
```
cpcallen, line 99: Ditto here.

core/utils/dom.ts
```diff
@@ -104,10 +112,16 @@ export function removeClasses(element: Element, classNames: string) {
  */
 export function removeClass(element: Element, className: string): boolean {
   const classNames = className.split(' ');
-  if (classNames.every((name) => !element.classList.contains(name))) {
+  if (
+    classNames.every(
+      (name) => !element.classList || !element.classList.contains(name),
```
cpcallen, line 117: Ditto here.

core/block_svg.ts
HollowMan6: > If some focusable node's .getFocusableElement() returns something which is not an instanceof Node then it has violated the interface definition

For now the `MultiselectDraggable` is dummy and doesn't have an actual corresponding DOM object, that's why this check is added here, not sure if there's a better way for handling this one.

https://github.com/mit-cml/workspace-multiselect/pull/106/commits/1c9b6d5d0a7a3573ff5a7ee2095d32a04c4668f5#diff-a6ef3f5605a5577d348743aa99d160ec6523f07f18c754f34e638513878515a5R47-R49

core/focus_manager.ts
HollowMan6: As explained:

> Currently, the workspace-multiselect plugin acts as an adapter. It maintains its own multiple selection set (state), which keeps track of currently selected blocks. At the Blockly side, this is implemented as if there's a global `MultiselectDraggable` (which implements IDraggable) for each workspace, and current design for the plugin is that **this instance of `MultiselectDraggable` should always be selected (in a dummy way as it won't create any new DOM object) when we have multiple blocks selected**, so that the plugin can receive corresponding actions and pass all the actions to the blocks in the multiple selection set (state). It was working well in previous versions, but now in v12, the current code for `getFocusManager().focusNode` seems to break this completely, as it keeps unselecting `MultiselectDraggable`.

This change is an attempt to address the `MultiselectDraggable` unselecting issue, as we don't want the actual focus to get synchronised when in multiselect mode, although I'm aware that this will cause issues with other Blockly features.

core/block_svg.ts
HollowMan6: For now, the MultiselectDraggable is dummy and doesn't have an actual corresponding DOM object. If this is properly addressed, I would assume this change will already be eliminated.

core/utils/dom.ts
HollowMan6: Similar reason as for now, the MultiselectDraggable is dummy and doesn't have an actual corresponding DOM object.

core/layer_manager.ts
HollowMan6: It's because that we should keep `MultiselectDraggable` selected when we have multiple blocks selected, and when we do the dragging, we drag multiple blocks one by one by calling the method here. This code will make the single exact block that is in the dragging process get selected, instead of keeping `MultiselectDraggable` selected.

DaryeDev:
:(

PR #9265, opened 2025-08-01
fix:  Check for existing event group in cleanUp()

### LINKED ISSUE #9264: Workspace `cleanUp()` method breaks event group
### Check for duplicates

### Description

The workspace `cleanUp()` function creates a new event group for itself, but doesn't check whether a group already exists. This seems an oversight, since other places in code do have this check when relevant.

### Reproduction steps

1. Write a function that creates an event group, does something observable, calls `workspace.cleanUp()`, and then does something observable. 
2. Run this function in playground.
3. Undo. Observe that only part of the actions are undone.

Expected behavior: the whole action is undone as one step.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

michaela-mm:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/9264

### Reason for Changes

Add a check into `workspace.cleanUp` whether an event group already exists, and resets it at the end.

### Test Coverage

Checked reproduction steps in playground.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -1674,7 +1674,10 @@ export class WorkspaceSvg
   /** Clean up the workspace by ordering all the blocks in a column such that none overlap. */
   cleanUp() {
     this.setResizesEnabled(false);
-    eventUtils.setGroup(true);
+    const existingGroup = eventUtils.getGroup();
+    if (!existingGroup) {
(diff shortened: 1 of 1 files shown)


gonfunko (approved):

gonfunko:
Thanks for the bug report and fix, we appreciate it!

Issue #8207, opened 2024-06-11
Use ARIA in core Blockly


rachel-fenichel:
### Check for duplicates

- [X] I have searched for similar issues before opening a new one.

### Problem

A key part of accessibility is [screen reader support](https://abilitynet.org.uk/factsheets/introduction-screen-readers). 

To do this, we need to use [ARIA](https://www.w3.org/WAI/standards-guidelines/aria/). From that website:
"WAI-ARIA provides a framework for adding attributes to identify features for user interaction, how they relate to each other, and their current state. WAI-ARIA describes navigation techniques to mark regions and common Web structures as menus, primary content, secondary content, banner information, and other types of Web structures."

ARIA is not the same as keyboard navigation and shortcut handling, but the two problems are closely coupled and we may need to make decisions about both together at times.

### Request

Use correct ARIA roles, states, and properties in:
- [Fields](https://github.com/google/blockly/issues/8205)
- Toolbox and flyout (and non-block sub-elements)
- Trashcan
- Zoom controls
- Widgets
- Scrollbars

### Additional context

Related issues, which we may want to close or update:
- https://github.com/google/blockly/issues/3370
- https://github.com/google/blockly/issues/2844

Major screenreaders we will want to test with (these may behave in slightly different ways but should all work):
- JAWS
- NVDA
- VoiceOver
- ChromeVox

We may also want to test with:
- TalkBack

BenHenning:
This work is now tracked more specifically by #9305 (ARIA support) and #9303 (auditing screen reader support across different screen readers).

Issue #3370, opened 2019-10-30
Integrating screen readers and keyboard navigation


alschmiedt:
**Is your feature request related to a problem? Please describe.**
Should look at how keyboard navigation interacts with screen readers. 
- Focus should go to  toolbox when navigation opens the toolbox.
- How fields deal with focus 
- Should loose focus when we click on something out of the workspace
I am sure there are more I am just not aware of yet.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

BenHenning:
Note from https://github.com/google/blockly-keyboard-experimentation/issues/523#issuecomment-2901951222 that this work should include performing an NVDA test, particularly a page-level point-in-time analysis since our use of tabindexes is pretty complicated and could have unintended side effects.

This analysis should probably wait until more of the ARIA context is correct since that will presumably have a big impact on the context and how it's represented.

BenHenning:
Note: I'm treating this as the general "add screen reader support into Blockly" for the sake of work tracking.

rachel-fenichel:
@BenHenning Is this tracking issue still useful, or should we close it in favor of the smaller issues that have already been broken out?

maribethb:
I don't think this tracking bug is still super useful, the entire project board is sort of tracking this issue.

BenHenning:
Discussed with @maribethb over chat. I plan to use this as a general view for the initial screen reader integration into Blockly (i.e. v13), and also a view of all of the work that went into the experimental changes (as a visual parallel to the screen reader document). I'll be tweaking the issue organization over the next few weeks since it's not urgent.

This contrasts a bit from the project board because that also includes bugs and sometimes non-screen reader issues that don't quite fit into the actual "introduce screen reader" project.

gonfunko:
Closing now that v13 is out the door.

PR #9262, opened 2025-07-29
feat!: Temp branch for screen reader concepts [DO NOT MERGE]

BenHenning:
This is a temporary staging branch for core changes that will (hopefully) eventually be entirely moved to https://github.com/google/blockly-keyboard-experimentation/pull/684.

Note that this may also serve as a reference for future longer-term design changes to core to accommodate screen reader feature requirements.

This will be closed once https://github.com/google/blockly-keyboard-experimentation/pull/684 is ready for review.

PR #9280, opened 2025-08-06
feat: Add initial support for screen readers (experimental)

BenHenning:
## The basics

### Resolves

Fixes part of #8207
Fixes part of #3370

### Proposed Changes

This introduces initial broad ARIA integration in order to enable at least basic screen reader support when using keyboard navigation.

Largely this involves introducing ARIA roles and labels in a bunch of places, sometimes done in a way to override normal built-in behaviors of the accessibility node tree in order to get a richer first-class output for Blockly (such as for blocks and workspaces).

### Reason for Changes

ARIA is the fundamental basis for configuring how focusable nodes in Blockly are represented to the user when using a screen reader. As such, all focusable nodes requires labels and roles in order to correctly communicate their contexts.

The specific approach taken in this PR is to simply add labels and roles to all nodes where obvious with some extra work done for `WorkspaceSvg` and `BlockSvg` in order to represent blocks as a tree (since that seems to be the best fitting ARIA role per those available: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles). The custom work specifically for blocks includes:
- Overriding the role description to be 'block' rather than 'tree item' (which is the default).
- Overriding the position, level, and number of sibling counts since those are normally determined based on the DOM tree and blocks are not laid out in the tree the same way they are visually or logically (so these computations were incorrect). This is also the reason for a bunch of extra computation logic being introduced.

One note on some of the labels being nonsensical (e.g. 'DoNotOverride?'): this was done intentionally to try and ensure _all_ focusable nodes (that can be focused) have labels, even when the specifics of what that label should be aren't yet clear. More components had these temporary labels until testing revealed how exactly they would behave from a screen reader perspective (at which point their roles and labels were updated as needed). The temporary labels act as an indicator when navigating through the UI, and some of the nodes can't easily be reached (for reasons) and thus may never actually need a label. More work is needed in understanding both what components need labels and what those labels should be, but that will be done beyond this PR.

### Test Coverage

No tests are added to this as it's experimental and not a final implementation.

The keyboard navigation tests are failing due to a visibility expansion of `connectionCandidate` in `BlockDragStrategy`. There's no way to avoid this breakage, unfortunately. Instead, this PR will be merged and then https://github.com/google/blockly-keyboard-experimentation/pull/684 will be finalized and merged to fix it. There's some additional work that will happen both in that branch and in a later PR in core Blockly to integrate the two experimentation branches as part of #9283 so that CI passes correctly for both branches.

### Documentation

No documentation is needed at this time.

### Additional Information

This work is experimental and is meant to serve two purposes:
- Provide a foundation for testing and iterating the core screen reader experience in Blockly.
- Provide a reference point for designing a long-term solution that accounts for all requirements collected during user testing.

This code should never be merged into `develop` as it stands. Instead, it will be redesigned with maintainability, testing, and correctness in mind at a future date (see https://github.com/google/blockly-keyboard-experimentation/discussions/673).

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -56,6 +56,7 @@ import * as blocks from './serialization/blocks.js';
 import type {BlockStyle} from './theme.js';
 import * as Tooltip from './tooltip.js';
 import {idGenerator} from './utils.js';
+import * as aria from './utils/aria.js';
 import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
 import {Rect} from './utils/rect.js';
@@ -168,6 +169,8 @@ export class BlockSvg
   /** Whether this block is currently being dragged. */
   private dragging = false;
 
+  public currentConnectionCandidate: RenderedConnection | null = null;
+
   /**
    * The location of the top left of this block (in workspace coordinates)
    * relative to either its parent block, or the workspace origin if it has no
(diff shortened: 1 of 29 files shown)


BenHenning:
There might need to be a bit of branch-fu in order to make the tests work correctly between this new long-lived core branch and the equivalent version in the keyboard navigation repo.

BenHenning:
Next steps here:
- [x] File an issue to track this work. (Edit: Using #8207 since it already exists)
- [x] Audit the remaining TODOs that were removed from https://github.com/google/blockly-keyboard-experimentation/pull/684 to make sure nothing was missed.
- [x] Audit against the original experimentation PR to ensure other core-specific things can be re-added (since it's easier without the monkey patching), per #9262.
- [x] Figure out a plan for getting cross-repo CI working.
- [x] Finalize names for the base branches (maybe we should add '-experimental'?).
- [x] Get CI passing on this PR and get it into review.

core/block_svg.ts
```diff
@@ -215,9 +218,69 @@ export class BlockSvg
     // The page-wide unique ID of this Block used for focusing.
     svgPath.id = idGenerator.getNextUniqueId();
 
+    aria.setState(svgPath, aria.State.ROLEDESCRIPTION, 'block');
+    aria.setRole(svgPath, aria.Role.TREEITEM);
+    this.recomputeAriaLabel();
+    svgPath.tabIndex = -1;
+    this.currentConnectionCandidate = null;
+
     this.doInit_();
```
microbit-matt-hillsdon: The aria bit needs to be after `doInit_()` for the block to have fields.

BenHenning: Nice catch. The dataflow of when things happen has changed each time the code was moved, so I missed this and somehow missed the voice output being incomplete for blocks.

rachel-fenichel (approved):

PR #9281, opened 2025-08-06
fix: Drag immovable and shadow blocks along with their parent.

### LINKED ISSUE #678: move mode for stacks of blocks with unmovable children
### Check for duplicates

### Description

it's possible to have stacks of blocks that are all dragged together by having some child blocks marked as not movable, so that they stick to the parent block when they're dragged with the mouse. Using move mode on the parent block should behave the same as dragging with the mouse, so that the entire stack is moved together.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/678

### Proposed Changes
This PR fixes a bug that could cause a dragged block to disconnect from its immovable/shadow next sibling(s), rather than bring them along for the ride. This could happen in core while holding command/control to drag a single block vs the entire stack, and in the keyboard experiment repo when entering move mode for a block with such siblings. Now, in single-block-drag or keyboard-experiment move mode, moving a "single" block will bring along all contiguous subsequent siblings that are immovable or shadows, and, if stack healing is on, connect the previous block and the first normal next-sibling around the group of blocks that was detached.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -501,22 +501,32 @@ export class Block {
       // Detach this block from the parent's tree.
       this.previousConnection.disconnect();
     }
-    const nextBlock = this.getNextBlock();
-    if (opt_healStack && nextBlock && !nextBlock.isShadow()) {
-      // Disconnect the next statement.
-      const nextTarget = this.nextConnection?.targetConnection ?? null;
-      nextTarget?.disconnect();
-      if (
-        previousTarget &&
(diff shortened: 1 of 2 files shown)


maribethb (approved):
nice! I thought this was going to be more complicated.

PR #9288, opened 2025-08-11
refactor: Make focusable elements responsible for scrolling themselves into bounds.

### LINKED ISSUE #9212: scroll components into view when they're focused
currently, the `LineCursor` manually calls `workspace.scrollBoundsIntoView` whenever the focused node is specifically a block, field, or workspace comments. it makes instanceof checks for these. this is clunky and not extensible.

should these components do this automatically in their "onNodeFocus" methods? is there a time when a node is focused that we *wouldn't* want to scroll them into view?

if there is and we can't do it in the `onNodeFocus` method, do we need an interface for "is-scroll-into-bounds-able" so that the linecursor can do it without making instanceof checks?

_Originally posted by @maribethb in https://github.com/google/blockly/pull/9211#pullrequestreview-2995327647_

gonfunko:
## The basics

### Resolves

Fixes #9212

### Proposed Changes
This PR updates `LineCursor` to no longer attempt to scroll elements into view when they are focused. Instead, this responsibility is delegated to the elements themselves in the `onNodeFocus()` callback. This simplifies the implementation of `LineCursor` by not tying it to specific kinds of elements, and means that additional focusable elements (whether first or third party) can adopt this behavior.

### DIFF
--- a/core/rendered_connection.ts
+++ b/core/rendered_connection.ts
@@ -26,6 +26,7 @@ import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {hasBubble} from './interfaces/i_has_bubble.js';
 import * as internalConstants from './internal_constants.js';
 import {Coordinate} from './utils/coordinate.js';
+import {Rect} from './utils/rect.js';
 import * as svgMath from './utils/svg_math.js';
 import {WorkspaceSvg} from './workspace_svg.js';
 
@@ -644,6 +645,14 @@ export class RenderedConnection
   /** See IFocusableNode.onNodeFocus. */
   onNodeFocus(): void {
     this.highlight();
+    const highlight = this.findHighlightSvg();
+    if (!highlight) return;
+    const bbox = highlight.getBBox();
+    // Y coordinate appears to be the middle of the connection.
+    const y = this.y - bbox.height / 2;
+    const bounds = new Rect(y, y + bbox.height, this.x, this.x + bbox.width);
+
+    this.getSourceBlock().workspace.scrollBoundsIntoView(bounds);
   }
 
   /** See IFocusableNode.onNodeBlur. */
@@ -656,12 +665,12 @@ export class RenderedConnection
     return true;
   }
 
-  private findHighlightSvg(): SVGElement | null {
+  private findHighlightSvg(): SVGPathElement | null {
     // This cast is valid as TypeScript's definition is wrong. See:
     // https://github.com/microsoft/TypeScript/issues/60996.
     return document.getElementById(this.id) as
       | unknown
-      | null as SVGElement | null;
+      | null as SVGPathElement | null;
   }
 }
 
--- a/core/rendered_connection.ts
+++ b/core/rendered_connection.ts
@@ -26,7 +26,6 @@ import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {hasBubble} from './interfaces/i_has_bubble.js';
 import * as internalConstants from './internal_constants.js';
 import {Coordinate} from './utils/coordinate.js';
-import {Rect} from './utils/rect.js';
 import * as svgMath from './utils/svg_math.js';
 import {WorkspaceSvg} from './workspace_svg.js';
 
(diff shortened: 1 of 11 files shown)


core/rendered_connection.ts
```diff
@@ -644,6 +645,14 @@ export class RenderedConnection
   /** See IFocusableNode.onNodeFocus. */
   onNodeFocus(): void {
     this.highlight();
+    const highlight = this.findHighlightSvg();
+    if (!highlight) return;
+    const bbox = highlight.getBBox();
+    // Y coordinate appears to be the middle of the connection.
```
BenHenning:
```suggestion
    // Y coordinate is in the middle of the connection.
```

Otherwise this reads as a discovery rather than a description.

BenHenning (commented):
Thanks! I think it makes a lot of sense to approach it this way. I worry a little about it being easy to forget to auto-scroll in `onNodeFocus` implementations, but I don't have a better suggestion on the approach.

Is it possible to add tests for the different components to verify that they auto-scroll correctly onto the screen? I realize that might be a lot to test, but if it's important to work then we probably should ensure that it stays working, if possible.

core/rendered_connection.ts
gonfunko: It kind of is a discovery; I can't find anywhere that this is documented, but experimentally it checked out. I'm hesitant to state it as a fact.

gonfunko:
Added browser tests to ensure that focusing elements moves the workspace such that the element's bounds are visible.

tests/browser/test/basic_playground_test.mjs
```diff
@@ -199,3 +199,224 @@ suite('Disabling', function () {
     },
   );
 });
+
+suite.only('Focused nodes are scrolled into bounds', function () {
```
BenHenning: `only` needs to be removed here.

BenHenning: Just to check: do each of these tests fail without the new auto-scrolling logic in place for each of the changed `onNodeFocus` methods? Just wanted to double check since I find these types of tests sometimes difficult to get the initial condition failing as expected when trying to verify the fix.

core/rendered_connection.ts
BenHenning: I worry that this will make it more confusing when trying to reason about this code in the future.

I dug into this a bit, too, and it surprises me. I think bbox is supposed to be providing an upper left position, not centered (though I have seen some weird bbox behaviors in SVGs outside of Blockly). If I understand the documentation correctly, though, bbox isn't affected by transformations which might be skewing things here).

I wonder: since the highlight SVG is just a clone of the block's path with some small adjustments, would `svg_math`'s `getRelativeXY` work for this? That might be a nicer approach if so since then it'll at least be consistent with blocks.

BenHenning (approved):
Thanks @gonfunko! Approving to unblock, but had a few follow-ups: PTAL.

tests/browser/test/basic_playground_test.mjs
gonfunko: Yes, I made sure they all failed without the changes and just reverified that.

gonfunko: Done, thanks!

core/rendered_connection.ts
gonfunko: I concur; converting to using the source block's position plus the relative XY of the highlight path yielded the same results as the bounding box, such that navigating backwards to an open input connection ~mostly scrolls it into view, but not along the vertical axis:

Screenshot 2025-08-15 at 1 12 09 PM

I suppose I'm happy that they're the same, but I was also hoping that would resolve this. Another possibility might be to just scroll the source block itself into view?

BenHenning: So something weird is happening here, then, with the highlight SVG. If it's possible to scroll the source block into view, that probably does sound cleaner and more consistent to how we'd solve other similar "part of the block is off screen" problems.

gonfunko: Done.

BenHenning (approved):
Really nice tests, and I think the latest solution is much cleaner.

PR #9291, opened 2025-08-12
fix: Fix bug that caused inadvertent scrolling when the `WidgetDiv` was shown.

### LINKED ISSUE #9246: Input fields and context menus may scroll workspace out of view
### Check for duplicates

### Description

We are using Blockly for multiple small inline exercises on a large page. When I click into an input field inside a block, my browser scrolls away, to the top or to the bottom of the page, depending on the exact setup. Something similar happens with the right-click context menu: The browser scrolls before the menu is even shown, causing it to appear in the wrong place.

### Reproduction steps

Minimal example that shows the problem:

```html
<!DOCTYPE html>
<title>Test</title>
<script src="https://unpkg.com/blockly@12.1.0/blockly_compressed.js"></script>
<script src="https://unpkg.com/blockly@12.1.0/blocks_compressed.js"></script>
<script src="https://unpkg.com/blockly@12.1.0/msg/en.js"></script>

<div id="blocklyDiv" style="width: 100%; height: 500px"></div>
<div style="height: 2000px"></div>

<script>
    Blockly.inject('blocklyDiv', {
        toolbox: {
            contents: [
                {
                    kind: 'block',
                    type: 'math_number'
                }
            ]
        }
    });
</script>
```

Click into the input field for the number block or right-click the block.

The bug seems to have appeared in 12.1.0; if I use 12.0.0, the problem is gone.

### Stack trace

```shell

```

### Screenshots

Screenshot after right click:

Image

### Browsers

Chrome desktop, Firefox desktop

gonfunko:
## The basics

### Resolves

Fixes #9246

### Proposed Changes
This PR updates the `FocusManager` to not automatically scroll focused nodes into view (the default browser behavior). This was causing shifts in the top-level page when the `WidgetDiv` was shown, and may have caused inadvertent scrolling in other contexts as well.

### DIFF
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -309,6 +309,8 @@ export class FocusManager {
    * Note that this may update the specified node's element's tabindex to ensure
    * that it can be properly read out by screenreaders while focused.
    *
+   * The focused node will not be automatically scrolled into view.
+   *
    * @param focusableNode The node that should receive active focus.
    */
   focusNode(focusableNode: IFocusableNode): void {
@@ -423,6 +425,8 @@ export class FocusManager {
    * the returned lambda is called. Additionally, only 1 ephemeral focus context
    * can be active at any given time (attempting to activate more than one
    * simultaneously will result in an error being thrown).
+   *
+   * This method does not scroll the ephemerally focused node into view.
    */
   takeEphemeralFocus(
     focusableElement: HTMLElement | SVGElement,
@@ -439,7 +443,7 @@ export class FocusManager {
     if (this.focusedNode) {
       this.passivelyFocusNode(this.focusedNode, null);
     }
-    focusableElement.focus();
+    focusableElement.focus({preventScroll: true});
 
     let hasFinishedEphemeralFocus = false;
     return () => {
@@ -574,7 +578,7 @@ export class FocusManager {
     }
 
     this.setNodeToVisualActiveFocus(node);
-    elem.focus();
+    elem.focus({preventScroll: true});
   }
 
   /**
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -426,7 +426,7 @@ export class FocusManager {
    * can be active at any given time (attempting to activate more than one
    * simultaneously will result in an error being thrown).
    *
-   * This method does not scroll the ephemerally focused node into view.
+   * This method does not scroll the ephemerally focused element into view.
    */
   takeEphemeralFocus(
     focusableElement: HTMLElement | SVGElement,
--- a/core/interfaces/i_focusable_node.ts
+++ b/core/interfaces/i_focusable_node.ts
@@ -59,6 +59,9 @@ export interface IFocusableNode {
    * they should avoid the following:
    * - Creating or removing DOM elements (including via the renderer or drawer).
    * - Affecting focus via DOM focus() calls or the FocusManager.
+   *
+   * Implementations should consider scrolling themselves into view here; that
+   * is not handled by the focus manager.
    */
   onNodeFocus(): void;
 
--- a/core/interfaces/i_focusable_node.ts
+++ b/core/interfaces/i_focusable_node.ts
@@ -60,8 +60,8 @@ export interface IFocusableNode {
    * - Creating or removing DOM elements (including via the renderer or drawer).
    * - Affecting focus via DOM focus() calls or the FocusManager.
    *
-   * Implementations should consider scrolling themselves into view here; that
-   * is not handled by the focus manager.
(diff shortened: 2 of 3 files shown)


core/focus_manager.ts
```diff
@@ -423,6 +425,8 @@ export class FocusManager {
    * the returned lambda is called. Additionally, only 1 ephemeral focus context
    * can be active at any given time (attempting to activate more than one
    * simultaneously will result in an error being thrown).
+   *
+   * This method does not scroll the ephemerally focused node into view.
```
BenHenning:
```suggestion
   * This method does not scroll the ephemerally focused element into view.
```

Nit on terminology for consistence.

core/focus_manager.ts
```diff
@@ -439,7 +443,7 @@ export class FocusManager {
     if (this.focusedNode) {
       this.passivelyFocusNode(this.focusedNode, null);
     }
-    focusableElement.focus();
+    focusableElement.focus({preventScroll: true});
```
BenHenning, line 446: For this & the change below: is it possible to have a test for each of these? It has very subtle down-stream effects, for even a change detector-esque check would be a good idea, I think.

core/interfaces/i_focusable_node.ts
```diff
@@ -59,6 +59,9 @@ export interface IFocusableNode {
    * they should avoid the following:
    * - Creating or removing DOM elements (including via the renderer or drawer).
    * - Affecting focus via DOM focus() calls or the FocusManager.
+   *
+   * Implementations should consider scrolling themselves into view here; that
```
BenHenning:
```suggestion
   * Implementations may consider scrolling themselves into view here; that
```

We probably don't want to be prescriptive here since, as I understand it, most nodes don't actually care to do this.

Edit: Ah, looking at #9288 it's more obvious that many nodes need to do this. I'm fine with keeping the current terminology if you prefer, but I think other make sense.

Ron may also have thoughts on this wording.

BenHenning (changes requested):
Thanks @gonfunko! This makes sense. I had a couple of thoughts--PTAL.

core/interfaces/i_focusable_node.ts
gonfunko: Done.

core/focus_manager.ts
gonfunko: Done.

core/focus_manager.ts
gonfunko: Added a browser test, because that was the only way I could find to exercise this where the problem was reproducible. It invokes the contextual menu, although for reasons I couldn't determine I had to right-left-right click to get the page to jump pre-fix. That does mean that the block itself should be focused from the left click and the contextual menu should get ephemeral focus, so both codepaths wind up getting exercised at least.

tests/browser/test/basic_playground_test.mjs
```diff
@@ -101,6 +101,24 @@ suite('Right Clicking on Blocks', function () {
     await contextMenuSelect(this.browser, this.block, 'Remove Comment');
     chai.assert.isNull(await getCommentText(this.browser, this.block.id));
   });
+
+  test.only('does not scroll the page', async function () {
```
maribethb: you've got a stray `only` here

gonfunko: Whoops, good catch, thank you!

tests/browser/test/basic_playground_test.mjs
```diff
@@ -101,6 +101,24 @@ suite('Right Clicking on Blocks', function () {
     await contextMenuSelect(this.browser, this.block, 'Remove Comment');
     chai.assert.isNull(await getCommentText(this.browser, this.block.id));
   });
+
+  test('does not scroll the page', async function () {
```
BenHenning: Just to check: does this fail as expected when either of the new `preventScroll` changes are reverted (since you mentioned it calls through to both paths)?

core/focus_manager.ts
BenHenning: Gotcha--I left a follow-up comment about verifying the fix but I think it's sensible if it does manage to inadvertently verify both paths.

BenHenning (approved):
Thanks @gonfunko! Approving to unblock, though I did leave one comment just to double check that the test fails as expected.

tests/browser/test/basic_playground_test.mjs
gonfunko: It does when the ephermal focus change is reverted; added another test to explicitly verify the active focus codepath (and verified that fails without its corresponding change).

PR #9327, opened 2025-08-20
fix: Correct the alignment of narrow text in input fields.

### LINKED ISSUE #9069: The position of `text` elements in blocks is wrong
### Check for duplicates

### Description

The position of `text` elements in blocks is wrong. It's actually the amount of transform on the parent `g` element of the `text` element. They used to be central, but are now rendered slightly to the left. 

It is very noticeable when you start and stop editing the field.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/71fca41f-08ee-4bae-be07-aaad88ab96a7

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9069

### Proposed Changes
This PR fixes a bug that caused narrow text values in Zelos input fields to be slightly misaligned horizontally from the center. Input fields have a minimum width, and if the measured width of the text was less than that, it was positioned incorrectly. I also adjusted a couple places that were manually setting the `width` attribute of the field size and bypassing the minimum size enforcement as a result; this wasn't a root cause, but could lead to tricky bugs down the line.

### DIFF
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -46,6 +46,11 @@ import type {WorkspaceSvg} from './workspace_svg.js';
 type InputTypes = string | number;
 
 /**
+ * The minimum width of an input field.
+ */
+const MINIMUM_WIDTH = 14;
+
+/**
  * Abstract class for an editable input field.
  *
  * @typeParam T - The value stored on the field.
@@ -115,8 +120,8 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
    */
   protected override get size_() {
     const s = super.size_;
-    if (s.width < 14) {
-      s.width = 14;
+    if (s.width < MINIMUM_WIDTH) {
+      s.width = MINIMUM_WIDTH;
     }
 
     return s;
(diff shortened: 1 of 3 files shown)


core/field_input.ts
```diff
@@ -732,6 +737,23 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     return true;
   }
 
+  /**
+   * Position a field's text element after a size change.  This handles both LTR
+   * and RTL positioning.
+   *
+   * @param xOffset x offset to use when positioning the text element.
```
RoboErikG: Consider renaming xMargin and clarifying in comment that this is added to both sides of the element. Since you're adding it to both sides and centering the text element it's not exactly an offset.

RoboErikG (approved):
One comment on naming/comments but otherwise LGTM.

core/field_input.ts
gonfunko: Done.

PR #9356, opened 2025-09-09
fix: Improve representation of `FlyoutButtons` in screenreaders.

### LINKED ISSUE #9294: FlyoutButton has a hardcoded aria-label of "Button"
### Check for duplicates

### Description

Using https://google.github.io/blockly-keyboard-experimentation/screenreader/ flyout labels and buttons read out poorly because of hardcoded incorrect label text.

### Reproduction steps

2. Note that the text "This is a label" in category Misc has role="button" and aria-label="Button" 
3. Note that the button "Create Variable" has the same aria label (role is good here)

They both get read as "button Button" or similar across screen readers.

(2) is particularly important in MakeCode as all categories have a label at the top.

As the text is correct already likely we can just remove the aria-labels in both cases and test with screen readers (OK in VoiceOver at least).

For the label maybe role=heading with the correct heading level passed from the application as Blockly doesn't have sufficient context to set the level correctly. In MakeCode there are headings for the category as a whole and for subsections of the category. But any role that resulted in the label text being read out would be a step forward.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9294

### Proposed Changes
This PR updates `FlyoutButton`s to use either a `button` or `treeitem` ARIA role based on whether they are acting as buttons or labels. The actual text will be read out in both cases, and it will be denoted as a button when it is one.

### Limitations
At least on VoiceOver, the screenreader focus ring targets the whole flyout for the button variant, and the selection status is not spoken. Neither is an issue for labels.

### DIFF
--- a/core/flyout_button.ts
+++ b/core/flyout_button.ts
@@ -416,13 +416,10 @@ export class FlyoutButton
     const xy = this.getPosition();
     const bounds = new Rect(xy.y, xy.y + this.height, xy.x, xy.x + this.width);
     this.workspace.scrollBoundsIntoView(bounds);
-    aria.setState(this.svgGroup, aria.State.SELECTED, true);
   }
 
   /** See IFocusableNode.onNodeBlur. */
-  onNodeBlur(): void {
-    aria.setState(this.svgGroup, aria.State.SELECTED, false);
-  }
+  onNodeBlur(): void {}
 
   /** See IFocusableNode.canBeFocused. */
   canBeFocused(): boolean {
--- a/core/flyout_button.ts
+++ b/core/flyout_button.ts
@@ -114,12 +114,17 @@ export class FlyoutButton
     this.id = idGenerator.getNextUniqueId();
     this.svgGroup = dom.createSvgElement(
       Svg.G,
-      {'id': this.id, 'class': cssClass},
+      {'id': this.id, 'class': cssClass, 'tabindex': '-1'},
       this.workspace.getCanvas(),
     );
 
-    aria.setRole(this.svgGroup, aria.Role.BUTTON);
-    aria.setState(this.svgGroup, aria.State.LABEL, 'Button');
+    // Set the accessibility role. All child nodes will be set to `presentation`
+    // to avoid extraneous "group" narration on VoiceOver.
+    if (this.isFlyoutLabel) {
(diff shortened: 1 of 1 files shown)


core/flyout_button.ts
```diff
@@ -406,10 +416,13 @@ export class FlyoutButton
     const xy = this.getPosition();
     const bounds = new Rect(xy.y, xy.y + this.height, xy.x, xy.x + this.width);
     this.workspace.scrollBoundsIntoView(bounds);
+    aria.setState(this.svgGroup, aria.State.SELECTED, true);
```
maribethb: I don't think `aria-selected` makes sense here, maybe that's why VO doesn't read it out. I think `aria-selected` is supposed to be used for like, custom radio buttons or something like that where they have a selected state. But here are you just using it to show that it's focused? I don't think you need to set the selected state for that. It being focused is enough. These buttons don't have state.

I found an issue that seems to imply that VO would respect aria-pressed for buttons, but I don't think that's appropriate either for the same reasons.

core/flyout_button.ts
```diff
@@ -135,6 +140,7 @@ export class FlyoutButton
         },
         this.svgGroup!,
       );
+      aria.setRole(shadow, aria.Role.PRESENTATION);
```
maribethb, line 143: I hate that you have to do this but I don't know enough to suggest anything better. There is *very* little out there in terms of mark up for VO specifically.

core/flyout_button.ts
gonfunko: That makes sense; I added it for consistency with blocks, which do get read out as "selected", but I think you're right and that behavior is the bug – obviously it's selected because the screenreader is reading it out! Removed accordingly.

core/flyout_button.ts
gonfunko: Agreed. Happy to revisit down the line if we find a better alternative though.

core/flyout_button.ts
maribethb: For blocks I could see it being useful in cases where selection != focus (e.g. for multiselect) so we might have to find a way to overcome the fact that VO won't let you use aria-selected for arbitrary roles. But, that is a different problem and I don't think it applies to flyout buttons regardless.

maribethb (approved):

Issue #9299, opened 2025-08-14
Introduce better block labeling for screen readers


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Screen reader experimentation (https://github.com/google/blockly-keyboard-experimentation/discussions/673) currently has a fairly poor experience for reading out blocks that relies on generating a human-readable sentence using a block's field labels. This doesn't account for inputs or work in cases when blocks lack a field label.

### Request

The ARIA label for blocks needs to be able to be explicitly provided by custom blocks, likely a custom block property. If this property isn't defined, then the label should fall back to being generated using a block's field labels.

However, other aspects and possible improvements need to be considered:
- Dynamic labels (i.e. cases when a block's label depends on dynamic state and can change over time). Both the computation and updating of these needs to be figured out.
- Templating labels (which may actually be a solution for dynamic labels), e.g. accounting for inputs and maybe connections.
- Whether aspects of `Block.toString()` can be used to provide a richer automatic label for cases when a custom block lacks a label.

Note that this issue also tracks updating all built-in blocks to include new labels under the new system introduced here.

### Alternatives considered

Other options:
- Only use `toString()`, though this can produce a fairly verbose and redundant description since it includes children in the output.
- Use a custom generator to produce block-specific strings. This would require a lot of block-specific specialization that would result in more wiring and overall code, though it may still be a viable option for internationalization in the long-term.
- Introduce a utility that infers a label from the block type.

### Additional context

This solution was largely determined in a team discussion this week with contributions from essentially the whole team.

BenHenning:
There have been a lot of questions and challenges around whether label templating is actually needed for blocks. As it stands, it probably isn't for now and is something we can add later on (it's likely easier to add this in a future major release of core than to remove it).

As such, we will lean on field labels to compute the block label. Field labels should already be reordered to account for internationalization (which was a benefit of templating).

The main drawbacks that will need mitigation:
- Some edge cases like the demo draw emoji block need additional context to understand the field value. This will be done using contextualized fields (see https://github.com/google/blockly/issues/9301#issuecomment-3272237369).
- Blocks don't have human-friendly names to use today, so this will need to be added. To start, this will be done using a central manager class for experimentation (that will be designed to be extensible), but in the long-term will probably take a data-driven design (e.g. by adding a new human-readable name field to the block definition).
[...]

PR #9357, opened 2025-09-10
Improve block labeling for screen readers

microbit-robert:
This PR is a quick prototype so that we can test changes to screenreader output based on recent discussions. The changes are not intended to represent any real attempt at implementation and don't concern themselves with real issues going forward such as translation, etc. Block summaries are created by reading value-inputs and fields in place, and recursively. The changes also attempt to determine the block shape, the number of fields/inputs, and number of children if applicable. Some code has also been moved so that the aria labels for text and dropdown fields update when their values are changed.

The readout is likely biased to the micro:bit team's preferences, but this can be easily changed to test different outputs. I may have also missed some suggestions, please shout if that's the case. There are likely scenarios where the readout doesn't make sense. I've tested the scenario referenced in the link below with VoiceOver and NVDA, and the result doesn't feel like an unreasonable readout, but thoughts very welcome.

Deployment: https://sr-fluent-alt.blockly-keyboard-experimentation.pages.dev/?scenario=navigationTestBlocks&toolbox=toolbox&renderer=zelos

~~I've no intention on filling out detailed information below as there is no intention to merge any of these changes.~~

## The basics

### Resolves

Partially addresses https://github.com/google/blockly/issues/9299

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -218,12 +218,11 @@ export class BlockSvg
     // The page-wide unique ID of this Block used for focusing.
     svgPath.id = idGenerator.getNextUniqueId();
 
-    aria.setState(svgPath, aria.State.ROLEDESCRIPTION, 'block');
-    aria.setRole(svgPath, aria.Role.TREEITEM);
     svgPath.tabIndex = -1;
     this.currentConnectionCandidate = null;
 
     this.doInit_();
+    this.computeAriaRole();
   }
 
   private recomputeAriaLabel() {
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -268,7 +306,7 @@ export class BlockSvg
       const firstSibling: BlockSvg = surroundParent.getChildren(false)[0];
       const siblings: BlockSvg[] = [firstSibling];
       let nextSibling: BlockSvg | null = firstSibling;
-      while ((nextSibling = nextSibling.getNextBlock())) {
+      while ((nextSibling = nextSibling?.getNextBlock())) {
```
microbit-robert, line 309: Prevents error when dropping "random colour" or similar blocks into the workspace. Has unknown consequences, but not being able to do this was hindering testing.

gonfunko:
@microbit-robert would you be open to merging this, at least the block parts? I think it would be helpful to have a basic implementation that we can then file more specific bugs against, particularly while we're still trying to figure out the appropriate level of verbosity and such.

BenHenning:
> @microbit-robert would you be open to merging this, at least the block parts? I think it would be helpful to have a basic implementation that we can then file more specific bugs against, particularly while we're still trying to figure out the appropriate level of verbosity and such.

This can be considered as fixing part of #9299 (though there may be some work past this, I think).

microbit-robert:
As requested this PR has been modified with changes isolated to `block_svg.ts`. Please be aware that these changes will overwrite a [relatively recent PR](https://github.com/google/blockly/pull/9361), and some of these changes may need reinstating.

core/block_svg.ts
```diff
@@ -235,29 +234,67 @@ export class BlockSvg
...
+      // The type is poorly inferred here.
+      // If currentBlock is null, we can't enter this while loop...
+      currentBlock = currentBlock!.getNextBlock();
+    }
+
+    let blockTypeText = 'block';
+    if (this.isShadow()) {
+      blockTypeText = 'input block';
+    } else if (this.outputConnection) {
+      blockTypeText = 'replacable block';
```
gonfunko, line 265: I believe these two clauses are reversed; blocks with an output connection are input blocks, and shadow blocks get replaced when another block is dragged atop them.

gonfunko (approved):

PR #9358, opened 2025-09-10
Update bug_report.yaml

sappm01:
Added in text area field for users to provide more details regarding priority.

## The basics

### Resolves

Fixes


maribethb:
Formatted locally since Mike generated this in the web editor so can't run the formatter.

Had to add the pipe character because otherwise the colons in the value trip up the yaml parser. I chose `|` over `>` because it should preserve the newlines I chose in the placeholder text which probably makes it easier to read.

cpcallen (approved):

PR #9361, opened 2025-09-12
fix: Use field ARIA labels for field-only blocks.

### LINKED ISSUE #9316: Screen reader output for full block fields
### Check for duplicates

### Description

The screen reader output for full block fields is currently unhelpful. The field roles are on an element that isn't selected or even a descendent of what is focused and so aren't of any use for the screen reader output.

### Reproduction steps

2. Navigate to any full block field
3. The screen reader will announce something generic rather than descriptive. See screenshot below.

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9316

### Proposed Changes
This PR makes blocks containing full-block fields derive their ARIA label from the label of their child field. Previously, these blocks did not have a label at all. The current field labels seem a bit suboptimal, but if they are updated that change will propagate through.

Because block labels are derived from their field labels, and Field's `getFocusableElement()` can throw if called early during initialization, I wrapped the access in a try-catch and moved the recalculation of the label into `initSvg()`, a bit later in the initialization sequence than the block constructor.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -241,12 +241,14 @@ export class BlockSvg
       // Shadow blocks are best represented directly by their field since they
       // effectively operate like a field does for keyboard navigation purposes.
       const field = Array.from(this.getFields())[0];
-      let label = 'Unknown?';
       try {
-        label =
-          aria.getState(field.getFocusableElement(), aria.State.LABEL) ?? label;
-      } catch {}
-      return label;
+        return (
+          aria.getState(field.getFocusableElement(), aria.State.LABEL) ??
+          'Unknown?'
+        );
+      } catch {
+        return 'Unknown?';
+      }
     }
 
     const fieldLabels = [];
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -239,15 +236,19 @@ export class BlockSvg
...
       // Shadow blocks are best represented directly by their field since they
       // effectively operate like a field does for keyboard navigation purposes.
       const field = Array.from(this.getFields())[0];
-      return (
-        aria.getState(field.getFocusableElement(), aria.State.LABEL) ??
-        'Unknown?'
-      );
+      try {
+        return (
+          aria.getState(field.getFocusableElement(), aria.State.LABEL) ??
```
maribethb, line 246: I think instead of trying to copy the label, this is what aria-labelledby is for? So we should put the id of the field's element here instead. Does that make any of what you were talking about earlier easier?

gonfunko: Seemingly so! Alas, https://issues.chromium.org/issues/40807024 exists, so I think we may be stuck with this.

maribethb: :/

maribethb (approved):

PR #9363, opened 2025-09-12
fix: Avoid creating redundant a11y tree nodes when selecting a block.

### LINKED ISSUE #9295: Focusing a treeitem in the flyout causes another treeitem to appear
### Check for duplicates

### Description

Focusing a treeitem in the flyout causes another treeitem to appear, perhaps because the block's path is cloned to indicate focus/selection.

### Reproduction steps

2. Navigate the Functions flyout (because it's short) while watching it in the accessibility tree in Chrome
3. Note that another treeitem pops into existence when you navigate to a block with the keyboard or (easier to see) hover over it with the mouse

I suspect the code clones or otherwise creates a similar path for the focus highlight but this clone will need only role=presentation.

I think this is a source of incorrect counting in the flyout tree in screen reader output. The functions flyout is often read has having 4 items when it has only 3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9295

### Proposed Changes
This PR updates the Zelos `PathObject` selection code to not clone the block's path to create a selection outline, but instead create a new path with only the intended attributes duplicated. This prevents duplicate nodes from becoming part of the accessibility tree by marking this path as `role="presentation"` rather than bringing over their existing role and ARIA attribute values.

### DIFF
--- a/core/renderers/zelos/path_object.ts
+++ b/core/renderers/zelos/path_object.ts
@@ -8,8 +8,8 @@
 
 import type {BlockSvg} from '../../block_svg.js';
 import type {Connection} from '../../connection.js';
-import {FocusManager} from '../../focus_manager.js';
 import type {BlockStyle} from '../../theme.js';
+import * as aria from '../../utils/aria.js';
 import * as dom from '../../utils/dom.js';
 import {Svg} from '../../utils/svg.js';
 import {PathObject as BasePathObject} from '../common/path_object.js';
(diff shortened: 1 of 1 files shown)


BenHenning (approved):
This clone has caused me multiple problems in the past, so this is a much cleaner and future-proof way to address it. Thanks @gonfunko!

Issue #9313, opened 2025-08-15
Discussion of the application of roles such as listbox, checkbox, and textbox


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

It feels like roles such as listbox, checkbox and textbox may be incorrectly applied in the workspace. Elements with the listbox role work to some extent in that Space / Enter opens the dropdown, but the down arrow does not open the dropdown and typing characters does not open the dropdown or change its value. Typing when focused on something with the role of textbox has no effect. You have to press Enter  / Space before focus is actually moved into an element that supports entering text.

Are elements that currently have these roles more like buttons to enable the editing of a field and should they have different semantics?

### Reproduction steps

1. Go to the [screen reader test page](https://google.github.io/blockly-keyboard-experimentation/screenreader/?scenario=navigationTestBlocks&toolbox=toolbox&renderer=zelos)
2. Navigate to the width field of the create canvas block
3. The screen reader will announce "edit text" as a result of the textbox role being applied to the element, but at this point, typing characters to change the value of the field has no effect. Is this misleading?

### Stack trace

```shell

```

BenHenning:
There seem to be two parts to this issue:
1. A general question of whether some of the roles are set up incorrectly, particularly for fields.
2. There are some expected default behaviors like typing-to-search in a combo box.

(1) is definitely the case, and is the premise for #9301 and its children (thus I'm marking this a child of that issue, though other sibling issues need to be resolved before this can be directly addressed).

(1) and (2) will both need to be reevaluated once fields have correct roles setup to determine if there's anything more to do here. It may be the case that some of this "just works" once the roles are correct, but I'm doubtful (at least for (2)) which means further discussion and work will be needed.

BenHenning:
#9384 is working to address basically all of the role and ARIA structure inconsistencies that #9301 is tracking. I still need to reevaluate the solution in the framing of (2) above, but I still suspect that the shortcuts will _not_ work out-of-the-box due to Blockly overriding basically all keyboard behaviors. The important question here is how much these shortcuts behave as part of the reader vs. expected to be handled by the application as part of the ARIA specification (and if there are any cases where it's reasonable for us to intentionally _not_ follow the specification).

BenHenning:
I've had some time to think about this issue more specifically, and here are my current thoughts:

[...]

Issue #9304, opened 2025-08-15
Audit and reduce complexity of accessibility node tree


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Ongoing screen reader experimentation work (https://github.com/google/blockly-keyboard-experimentation/discussions/673) has resulted in a lot of changes to the accessibility node tree. This means that there are many extra nodes, incorrectly placed nodes, or incorrectly represented nodes in the tree.

### Request

The accessibility node tree should be configured such that it contains the exact structure desired for screen reader representation. Much of this is covered in other ARIA work so this issue covers ensuring that, once the core functionality is completed, that the structure of the tree is exactly as desired (that is, all nodes are intentionally placed and correct with no extra, invalid, or missing nodes).

Part of this should ensure hierarchy: some nodes may require adding `aria-owns` attributes in places to ensure that ownership is correctly represented.

Part of this should also including auditing all existing ARIA attributes and ensuring each one is correctly and intentionally set (since some existing attributes in core Blockly may require changes).

### Alternatives considered

The only alternative is allowing for excess nodes to be present. This isn't desirable since it could result in some screen readers incorrectly representing the structure of the page, potentially confusing users.

rachel-fenichel:
@BenHenning I've moved this to an in-progress P1 because you already have a PR open to do some of it, and i don't want to let that PR get too stale.

BenHenning:
Fixed in #9449 (well enough anyway).

Issue #9301, opened 2025-08-14
Introduce better field labeling and accessibility for screen readers


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

Screen reader experimentation (https://github.com/google/blockly-keyboard-experimentation/discussions/673) currently has somewhat insufficient labeling and context support for fields.

Specifically, block fields may lack type context (e.g. proper ARIA setup for drop-downs), or may not have a sufficiently clear described context for custom fields (e.g. colour picker).

### Request

Field accessibility largely comes down to three things:
1. Supporting a descriptive label for what the field actually represents along with its value.
2. Providing additional context for how the field works semantically (and provide some hint for how its editor will work).
3. Ensuring that field editors are themselves accessible.

(1) can be solved by introducing an expectation for a new getter for `Field` that can dynamically determine a good fitting label for the field. This getter can be provided in custom fields without an override, and is thus fine to introduce without custom versions or complicated NPM linking in plugins (since the getter will simply go unused outside of the screen reader experimentation work). Some specific considerations here:
- When does the field need to recompute its label and possibly update its properties? Some invariants/assumptions need to be established in the getter's or class's contract.
- Custom getters in plugins should be tracked so that they can be removed or changed if the long-term solution doesn't end up using them in the same way.
- How should the getter be defaulted? This should largely be on a class-by-class basis, but the specific field value may need to be considered as well.

(2) needs to be solved by introducing correct ARIA roles and, in some cases, elements on a field-by-field basis. #8205 has additional details for implementing ARIA for different fields. This is tracked by the following issues (may not be comprehensive):
- #8206
- #8210
- #8255
- #8211
- #8254
- #8212
- #8213

(3) needs to be done, at minimum, for any external plugins that the keyboard navigation plugin playground uses, plus for all built in fields. The latter are covered by the issues mentioned in (2), and the former are covered as follows:
- To be done in a follow-up issue as discovered.

### Alternatives considered

One alternative considered was introducing a utility that can determine a label based on the specific class, though this is a dependency inversion and is largely tricky to manage (even in an experiment) vs. using a more inheritance-based approach (which in this case is a bit easier to work with).

### Additional context

This solution was largely determined in a team discussion this week with contributions from essentially the whole team.

BenHenning:
Outside of the obvious roles & labels that are being tracked by the child items of this issue, the main two gaps to figure out for fields are:
[...]

Issue #1595, opened 2018-02-08
Use <input id="number" type="number"> for field_number editor


AnmAtAnm:
### Problem statement

FieldNumber editors should provide the correct `<input>` attributes to help touch/mobile device onscreen keyboards.

### Expected Behavior

Number fields editors should use `<input id="number" type="number" ...>`.

Need to verify the change does not have unexpected impact on non-touch platforms.

### Actual Behavior

`field_number` currently use the following for editor:

    <input class="blocklyHtmlInput" spellcheck="true" value="0" style="font-size: 11pt;">

alschmiedt:
Tried fixing this in #2607 by setting the htmlInput_.type to number. However, on mobile it shows a prompt editor so the htmlInput_ is undefined and breaks. More information on this issue can be found here: #2659.

maribethb:
There is no good way to do this, unfortunately. The current behavior is that you should be able to enter any value that JavaScript can convert to a number with `Number(value)` (after we do our transformations). This includes `Infinity` and hexadecimal numbers.

The default mobile experience is to use modal inputs instead of the text input anyway. It's only pretty recently that we supported disabling these inputs in core and even more recently that that experience worked well enough on all mobile browsers for anyone to turn it on. So largely this doesn't even matter, because if you're using modal inputs, which use `window.prompt` there is no way to control the kind of keyboard shown.


But if you are disabling the modal inputs then:

- `type="number"` doesn't work because the browser is too strict about what a number is. It doesn't let you enter any non-digit text so you cannot enter Infinity or hexadecimal numbers, for example.
- `inputMode="numeric"` is the preferred approach, and it works well on desktop and iOS. The browser will still let you enter any text. On iOS it will show the number keypad by default but it will still let you switch to the alpha keyboard so you can still enter `Infinity`. However, on Android, it will show the number keyboard but **not** let you switch to alpha.
- None of the input modes supported by android let you show the number keyboard **and** switch to alpha. 

Thus, on Android, there would not be any way to both show the number keyboard and allow users to enter values like `Infinity`. 

Instead, we could solve this by providing a specific mobile-friendly number input as a plugin, maybe with a popout numpad widget, and a separate `Infinity` math block. Developers who come across this issue and want this feature and don't care about Infinity or hexadecimal should feel free to subclass the number field and add `inputMode="numeric"` to the HTML.

PR #9384, opened 2025-09-24
feat: Fix ARIA roles and setup for fields (experimental)

### LINKED ISSUE #8206: ARIA for checkbox fields
See the parent issue #8205 for more information about the general problem.

This issue is for discussing and agreeing on expected behaviour, and then tracking implementation.

Tasks:
- [x] Apply the correct [ARIA roles and properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to the DOM elements that represent the field
- [x] Provide [accessible names and descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) to the DOM elements that represent the field
- [x] Make sure the field works both during editing and during display (not editing)
- [x] Describe the correct keyboard shortcuts to move into and out of editing mode when on a single instance of the field (e.g. enter, space, and esc)
  - Implementation may happen in the keyboard nav code or repository

Resources:
- ARIA Authoring Practices Guide: [Checkbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/)
- APG [two-state checkbox example](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/)
- [Blockly checkbox field documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/checkbox)

### LINKED ISSUE #8210: ARIA for dropdown fields
See the parent issue #8205 for more information about the general problem.

This issue is for discussing and agreeing on expected behaviour, and then tracking implementation.

Tasks:
- [x] Apply the correct [ARIA roles and properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to the DOM elements that represent the field
- [x] Provide [accessible names and descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) to the DOM elements that represent the field
- [x] Make sure the field works both during editing and during display (not editing)
- [x] Describe the correct keyboard shortcuts to move into and out of editing mode when on a single instance of the field (e.g. enter, space, and esc)
  - Implementation may happen in the keyboard nav code or repository

Resources:
- ARIA Authoring Practices Guide: [Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- APG [select-only combobox example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)
- [Blockly dropdown field documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/dropdown)

### LINKED ISSUE #8211: ARIA for text input fields
See the parent issue #8205 for more information about the general problem.

This issue is for discussing and agreeing on expected behaviour, and then tracking implementation.

Tasks:
- [x] Apply the correct [ARIA roles and properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to the DOM elements that represent the field
- [x] Provide [accessible names and descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) to the DOM elements that represent the field
- [x] Make sure the field works both during editing and during display (not editing)
- [x] Describe the correct keyboard shortcuts to move into and out of editing mode when on a single instance of the field (e.g. enter, space, and esc)
  - Implementation may happen in the keyboard nav code or repository

Resources:
- [ARIA: textbox role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role)
- [Blockly text input field documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/text-input)

### LINKED ISSUE #8212: ARIA for number input fields
See the parent issue #8205 for more information about the general problem.

This issue is for discussing and agreeing on expected behaviour, and then tracking implementation.

Tasks:
- [x] Apply the correct [ARIA roles and properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to the DOM elements that represent the field
- [x] Provide [accessible names and descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) to the DOM elements that represent the field
- [x] Make sure the field works both during editing and during display (not editing)
- [x] Describe the correct keyboard shortcuts to move into and out of editing mode when on a single instance of the field (e.g. enter, space, and esc)
  - Implementation may happen in the keyboard nav code or repository

Resources:
- [ARIA: textbox role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role)
- [ARIA: slider role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role)
  - May be relevant because we allow min and max
  - However, most number inputs are not sliders
- [ARIA Authoring Practices Guide: Slider Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/slider/)
- [WCAG 2.0 Understanding Docs: Input Assistance](https://www.w3.org/WAI/WCAG22/Understanding/input-assistance.html) 
- [Blockly number field documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/number)

### LINKED ISSUE #8213: ARIA for variable fields
See the parent issue #8205 for more information about the general problem.

This issue is for discussing and agreeing on expected behaviour, and then tracking implementation.

Variable fields are mostly just dropdown fields (https://github.com/google/blockly/issues/8210) with a data source that Blockly owns. It is possible that they will not need additional work. 

Tasks:
- [x] Apply the correct [ARIA roles and properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to the DOM elements that represent the field
- [x] Provide [accessible names and descriptions](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) to the DOM elements that represent the field
- [x] Make sure the field works both during editing and during display (not editing)
- [x] Describe the correct keyboard shortcuts to move into and out of editing mode when on a single instance of the field (e.g. enter, space, and esc)
  - Implementation may happen in the keyboard nav code or repository

Resources:
- ARIA Authoring Practices Guide: [Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
- APG [select-only combobox example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)
- [Blockly variable field documentation](https://developers.google.com/blockly/guides/create-custom-blocks/fields/built-in-fields/variable)

### LINKED ISSUE #8254: ARIA for label fields
Parent Issue: https://github.com/google/blockly/issues/8205

### LINKED ISSUE #8255: ARIA for image fields
Parent Issue: https://github.com/google/blockly/issues/8205

BenHenning:
## The basics

### Resolves

Fixes #8206
Fixes #8210
Fixes #8213
Fixes #8255
Fixes #8211
Fixes #8212
Fixes #8254
Fixes part of #9301
Fixes part of #9304

### Proposed Changes

This PR completes the remaining ARIA roles and properties needed for all core fields. Specifically:
- #8206: A better name needed to be used for the checkbox value, plus there was an ARIA property missing for actually representing the checkbox state. The latter needed to be updated upon toggling the checkbox, as well. These changes bring checkbox fields in compliance with the ARIA checkbox pattern documented here: https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/.
- #8210: This one required a lot of changes in order to adapt to the ARIA combobox pattern documented here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/. Specifically:
  - Menus needed to have a unique ID that's also exposed in order to link the combobox element to its menu when open.
  - ARIA's `activedescendant` proved very useful in ensuring that the current dropdown selection is correctly read when the combobox has focus but its menu isn't opened.
  - The default properties available for options (label and value) aren't very good for readout, so a custom ARIA property was added for much clearer option readouts. This is only demonstrated for the math arithmetic block for now.
  - The text element is normally hidden for ARIA but it's useful in conjunction with `activedescendant` to represent the current value selection.
  - Images have been handled here as well (partly as part of #8255) by leveraging their alt text for readouts. This actually seems to work quite well both for current value and selection.
- #8213: Much of the improvements here come from the combobox (`FieldDropdown`) improvements explained above. However one additional bit was done to provide an explicit 'Variable <name>' readout for the purpose of clarity. This demonstrates some contextualization of the value of the field which may be a generally useful pattern to copy in other field contexts.
- #8255: Image fields have been refined since they were redundantly specifying 'image' when an `image` ARIA role is already being used. Now only the alt text is supplied along with the role context. Note that images need special handling since they can sometimes be navigable (such as when they have click handlers).
- #8211: Text input fields have had their labeling improved like all other fields, and the field's value is now exposed via its `text` element since this will show up as a `StaticText` node in the accessibility tree and automatically be read as part of the field's value.
- #8212: This gets the same benefits as the previous point since those improvements were included for both text and number input. However, existing `valuemin` and `valuemax` ARIA properties have been removed. It seems these are really only useful when introducing a slider mechanism (see https://www.w3.org/WAI/ARIA/apg/patterns/slider/) and from testing seems to not really be utilized for the basic text input that `FieldNumber` currently uses. It may be the case that this is a better pattern to use in the future, but it's more likely that other custom fields could benefit from more specific patterns like slider rather than `FieldNumber` being changed in that way.
- #8254 and part of #9304: Field labels have been completely removed from the accessibility node tree since they can never be navigated to (as #8254 explains all labels will be included as part of the block's ARIA label itself for readout parity with navigation options).

Note that it doesn't cover external fields (such as those supplied in blockly-samples), nor does it fully set up the infrastructure work for those. Ultimately that work needs to happen as part of #9301.

Beyond the role work above, this PR also introduces some fundamental work for #9301. Specifically:
- It demonstrates how block definitions could be used to introduce accessibility label customizations (in this case for the options of the arithmetic operator block's drop-down field, plus the block itself).
- It sets up some central label computation for all fields, though more thought is needed on whether this is sufficient for custom fields outside of core Blockly and on how to properly contextualize labels for field values. Core Blockly's fields are fairly simple for representing values which is why that aspect of #9301 didn't need to be solved in this PR. Note that the field labeling here is being used to improve all of the fields above, but also it tries to aggressively fall back to the _next best_ label to be used (though it's possible to run out of options which is why fields still need contextually-specific fallbacks).

### Reason for Changes

Generally the initial approach for implementing labels is leveraging as specific ARIA roles as exist to directly represent the element. This PR is completing that work for all of core Blockly's built-in fields, and laying some of the groundwork for generalizing this support for custom fields.

Having specific roles does potentially introduce inconsistencies across screen readers (though should improve consistency across sites for a single screen reader), and expectations for behaviors (like shortcuts) that may need to be ignored or only partially supported (#9313 is discussing this).

### Test Coverage

Only manual testing has been completed since this is experimental work.

Video demonstrating most of the changes:

[Screen recording 2025-10-01 4.05.35 PM.webm](https://github.com/user-attachments/assets/c7961caa-eae0-4585-8fd9-87d7cbe65988)

### Documentation

N/A -- Experimental work.

### Additional Information

This has only been tested on ChromeVox.


blocks/math.ts
```diff
@@ -54,12 +54,13 @@ export const blocks = createBlockDefinitionsFromJsonArray([
...
         'type': 'field_dropdown',
         'name': 'OP',
+        'optAriaName': 'Arithmetic operation',
         'options': [
-          ['%{BKY_MATH_ADDITION_SYMBOL}', 'ADD'],
-          ['%{BKY_MATH_SUBTRACTION_SYMBOL}', 'MINUS'],
-          ['%{BKY_MATH_MULTIPLICATION_SYMBOL}', 'MULTIPLY'],
-          ['%{BKY_MATH_DIVISION_SYMBOL}', 'DIVIDE'],
-          ['%{BKY_MATH_POWER_SYMBOL}', 'POWER'],
+          ['%{BKY_MATH_ADDITION_SYMBOL}', 'ADD', 'Plus'],
```
BenHenning, line 59: NB: Will need be added to messages.

BenHenning:
Note to self: maybe check shortcuts since that could allow #9313 to possibly be addressed, too.

BenHenning:
Keyboard nav test failure is happening due to https://github.com/google/blockly-keyboard-experimentation/blob/30c5286ee0910c2795c1c93bbf440ed5ab1b3953/test/webdriverio/test/block_comment_test.ts#L39 being a bit fragile with directly using a generated ID that can easily change (such as in this PR since more IDs are being generated). That test will need to be updated outside of this PR.

blocks/math.ts
BenHenning: Not planning to resolve this as part of this PR--it can be covered with #9301 (if even then since localizing these is likely unimportant for upcoming user tests).

BenHenning (commented):
Self-review.

core/field.ts
```diff
@@ -199,6 +199,10 @@ export abstract class Field<T = any>
   /** The unique ID of this field. */
   private id_: string | null = null;
 
+  private configuredAriaName: string | null = null;
```
maribethb: bikeshed over names: I don't think these should use the prefix "configured"

core/field.ts
```diff
@@ -1418,7 +1440,10 @@ export abstract class Field<T = any>
  * Extra configuration options for the base field.
  */
 export interface FieldConfig {
+  type: string;
+  name?: string;
   tooltip?: string;
+  optAriaName?: string;
```
maribethb: Don't use the prefix "opt", it's already optional in the type (and tooltip for example is optional but doesn't have a prefix).

core/field_dropdown.ts
```diff
@@ -198,12 +199,28 @@ export class FieldDropdown extends Field<string> {
...
     }
 
+    this.recomputeAria();
+  }
+
+  private recomputeAria() {
+    if (!this.fieldGroup_) return; // There's no element to set currently.
     const element = this.getFocusableElement();
-    aria.setRole(element, aria.Role.LISTBOX);
+    aria.setRole(element, aria.Role.COMBOBOX);
```
maribethb, line 208: I find the aria spec really confusing here. Our dropdown fields are not comboboxes in the traditional sense, because they do not allow you to type in them. They are not "combo" boxes, they are just listboxes.

But reading [this page](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) sort of does imply that it is a "non-editable" combo box. I would just call that a listbox personally. At the end there is some criteria that distinguishes comboboxes and listboxes, but unfortunately our proposal doesn't fit either category cleanly:

- You can abort the selection with escape, and the value is not set immediately, only when you press enter (combobox)
- The user's choice is not presented in an editable field when the list is closed; it can't be copied to the clipboard (listbox)

I feel that listbox would fit the "spirit" of the dropdown better, and it would be confusing if users expect a combobox to allow them to type.

core/field_image.ts
```diff
@@ -164,7 +164,7 @@ export class FieldImage extends Field<string> {
     aria.setState(
```
maribethb, line 166: If the image field has a click handler then its role should be set as "button" instead.

core/field_input.ts
```diff
@@ -176,13 +178,13 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
+    aria.setState(this.textElement_, aria.State.HIDDEN, false);
+
     const element = this.getFocusableElement();
     aria.setRole(element, aria.Role.TEXTBOX);
-    aria.setState(
-      element,
-      aria.State.LABEL,
-      this.name ? `Text ${this.name}` : 'Text',
-    );
+    aria.setState(element, aria.State.LABEL, this.getAriaName() ?? 'Field');
```
maribethb: text makes more sense than "field" imo as a default, with the caveat that as stated elsewhere, I think not reading anything would be better than reading this. The role already says it is a text box. "Field" doesn't mean anything to users is likely to confuse them.

core/field_number.ts
```diff
@@ -296,14 +295,11 @@ export class FieldNumber extends FieldInput<number> {
   protected override widgetCreate_(): HTMLInputElement {
     const htmlInput = super.widgetCreate_() as HTMLInputElement;
 
-    // Set the accessibility state
     if (this.min_ > -Infinity) {
       htmlInput.min = `${this.min_}`;
-      aria.setState(htmlInput, aria.State.VALUEMIN, this.min_);
```
maribethb, line 302: When sighted users are typing in a value, if it's invalid, the field is red. VI users should get the same feedback somehow (that's sort of a different question about field editors though) and being upfront about the min/max seems better. Why remove it? It is probably more common for number input fields that have a min/max to be implemented as sliders, but I don't think that's a good enough reason to remove it here if a number input does set it. The Slider field plugin inherits from this field, so doing so here probably sets it for the slider field too.

core/field.ts
```diff
@@ -272,6 +283,17 @@ export abstract class Field<T = any>
     this.id_ = `${block.id}_field_${idGenerator.getNextUniqueId()}`;
   }
 
+  getAriaName(): string | null {
+    return (
```
maribethb, line 279: I think this is fine to prototype with but I question if this is a good fallback system. I think in many cases, not reading anything would be better than reading the field's name or type. I don't think it's useful information. Especially if we go the route of providing shortcuts that would read more information. Imagine you have a shortcut that, while you're focused on a field, would read the entire block's text. That would give you more context than the name or type of the field.

Or, you could check if the block has one or more label fields immediately preceding this one. If it does, then we could say that this field is aria-labelledby* those label fields. This would also need prototyping and looking at some examples of custom blocks to make sure this is sensible but I think it would on average give better results

*if you ignore the fact that aria-labelledby does not work in chrome on mac with treeitems currently (may not be an issue as we go away from treeitems)

maribethb (approved):

maribethb:
Also, if  there's any way you could attach a video demo to this PR, that would help a lot, as I find it difficult sometimes to know exactly what's being read out based on the roles and state. Like somewhere the actual value should be read out, which doesn't seem to be explicitly in any of the manually-set roles, so the video would help me make sense of how it all fits together. Or, in lieu of that, writing out a couple of examples.

core/field_number.ts
BenHenning: Because as far as I can tell, screen readers ignore this property for the role we're using here. The spec suggests that valuemin is for a range widget (per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemin) though it _is_ supposed to be used for `<input>`s with type `number`. However, we aren't using `number` here which means it defaults to `text` and thus `valuemin` seems ignored.

Changing the actual type to `number` seemed a lot more significant--do we actually want to do that?

core/field_input.ts
BenHenning: Text is a good suggestion--will update. Ideally this case is never triggered, but it is apparently theoretically possible.

core/field_image.ts
BenHenning: Will update, though I'm not sure we'll ever use a non-button version of this since I think we can only navigate to clickable images (so otherwise the image's context, i.e. alt text, will be included in its block label).

core/field_dropdown.ts
BenHenning: So I think I fully disagree withi this. The very beginning of the combobox pattern page states this:

> A [combobox](https://w3c.github.io/aria/#combobox) is an input widget that has an associated popup. The popup enables users to choose a value for the input from a collection. The popup may be a [listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/), [grid](https://www.w3.org/WAI/ARIA/apg/patterns/grid/), [tree](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/), or [dialog.](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)

The idea that it can be attached to an input for typing seems auxiliary and that the primary pattern difference between combobox and listbox is whether the list of options is collapsed. Comboboxes hide their options and listboxes don't. From that definition, Blockly's use is very clearly a combobox.

I also checked with Gemini and it seems to suggest the same thing and this particular bit stood out to me:

> The role="listbox" is meant for the container that holds the list of options (role="option"). A listbox is typically presented as an always-visible, scrollable list. While a listbox can be used as the popup part of another widget, applying the listbox role to your entire dropdown widget would be incorrect, as it fails to describe the interactive "show/hide" nature of the controlling button.

This aligns with the current implementation since we actually do use a listbox: the modal that we show with the options is the listbox of options, but the core widget used as a field is a combobox that an activate that pop-up listbox. This seems to very strongly fit the pattern explained in the documentation.

core/field.ts
BenHenning: Yeah...I did notice this inconsistency halfway through and have no issues with changing it.

core/field.ts
BenHenning: I didn't want to spend too much mental effort thinking about names, but in that vain I'm also fine just removing the prefix. Proper thought needs to go into design for the long-term solution, anyway.

core/field_dropdown.ts
BenHenning: Also: if it helps, the drop-down widgets on for the kbe playground page are comboboxes in the accessibility tree (and as they don't have typeable input they seem to be perfect analogues to the dropdown field widget). They are simple `select` elements in the markup.

core/field_number.ts
gonfunko: I think I looked at making it `number`, but unfortunately we support `Infinity`, which input type=number does not IIRC

core/field_dropdown.ts
maribethb: I don't think it's that clearcut according to common definitions of comboboxes, compare https://designsystem.digital.gov/components/select/ to https://designsystem.digital.gov/components/combo-box/

Our dropdowns are meant to replicate the html `select` element, which have an aria role of "listbox". That's where my confusion comes from. I think this page is a bit confusing because most pages aren't building their own plain dropdowns, you only need to build custom comboboxes, that's the lens this is written through.

But I don't feel strongly enough about this to block it. I think it's fine to keep this and revisit if we get feedback that users expect to be able to type into that field.

core/field_number.ts
maribethb: No, we don't want to do that. The "number" type includes too many restrictions, see [this issue](https://github.com/google/blockly/issues/1595#issuecomment-1533953067) for more discussion.

I think it would be fine to leave these in in case subclasses do override and set type = number, or use a slider like in the field slider. Is there a downside if they're ignored for text inputs? Having them here would make subclasses more likely to be more correct.

core/field_dropdown.ts
BenHenning: > Our dropdowns are meant to replicate the html `select` element, which have an aria role of "listbox". That's where my confusion comes from. I think this page is a bit confusing because most pages aren't building their own plain dropdowns, you only need to build custom comboboxes, that's the lens this is written through.

I don't think this is correct, though. As I mentioned in my previous message we use `<select>` for the tooolbox, scenario, and renderer selections on the KBE playground (see https://google.github.io/blockly-keyboard-experimentation/screenreader) and if you inspect those in the accessibility node tree they are Comboboxes:

image

and

image

Interestingly those don't actually use listbox at all, but they do use a `MenuListPopup` for the options (which I'm unsure how to replicate and may well be a built-in exclusive role, but that's probably unimportant for us).

core/field_number.ts
BenHenning: That's an interesting point (the idea that it could make subclasses more likely to be correct). I don't disagree with that, but it does feel rather wishful vs. explicitly designing in the requirement. Perhaps we could build in a mode for `FieldNumber` where it can actually be type `number` and support `minvalue`/`maxvalue`. In fact with the right combination of settings `Infinity` shouldn't even be allowed (perhaps ditto for hex numbers, but that depends on if that's a configurable option).

This isn't something I'll fight hard for, but at the same time I'm not keen on setting what amounts to invalid ARIA properties. I actually lean toward expecting subclass implementations to implement their ARIA properties correctly (since that's what we expect for field editors, anyway).

core/field.ts
BenHenning: This are great points, and as discussed over chat I think it makes sense for us to lean toward less context (per other ongoing discussions this week) which means removing all context from potential readouts. I'll do that in a follow-up PR so that I can think about it holistically and compare it against other ongoing improvements to block labeling (which will be largely affected by it).

core/field_dropdown.ts
maribethb: Sorry, I think you're right. The spec says it should only be a listbox if multiselect is on or if the size is > 1 (number of options shown to user at once) which doesn't apply here.

I think I agree with [Wikipedia's definition](https://en.wikipedia.org/wiki/Combo_box) and [this commenter](https://github.com/w3c/html-aam/issues/46) in this very long thread that listbox makes more sense than combobox, and that "read-only combobox" to describe a dropdown without a textbox is a **horrible** naming choice as it implies that you can't change the value of the select. But apparently my beef is with the html standards body, and not you. Carry on!

core/field_number.ts
maribethb: Alright, fair enough. Getting this stuff right is really tricky so I think it's good if we can set things up for developers as much as possible, but I understand where you're coming from too.

I do think at some point it would make sense for us to provide alternative number blocks that don't support hex, use the correct type or inputtype, and also a separate Infinity block. And we could set these values correctly in those blocks if so, which reduces the risk that developers don't set their roles correctly.

core/field_dropdown.ts
BenHenning: That...was a rabbit hole I didn't know exist and I really appreciate you finding it. It actually never occurred to me that the name combobox implies there must be a text-editable component to it. I'm quite used to combobox being used synonymously with drop-down, and only now do I realize my folly in not questioning the 'combo' part of that name.

It seems there's no true good answer for how to resolve the confusion and that thread simply led to them deciding that combobox was the better fit between the two choices and now it's standard (if I'm reading it correctly). I suppose we just need to live with it, then, and hope users don't get too confused with it. We at least benefit from consistency here I suppose.

core/field.ts
BenHenning: Ended up just saving the configuration itself and fetching it (to avoid a conflict with `name`).

core/field.ts
BenHenning: Fixed in latest.

core/field_dropdown.ts
BenHenning: Resolving this since it seems like we probably need to stick with combobox for correctness and consistency.

core/field.ts
BenHenning: I noted this in https://github.com/google/blockly/issues/9301#issuecomment-3358468069 so resolving this.

core/field_number.ts
BenHenning: Added a TODO to track this in code so that we reconsider it when evaluating long-term design choices for screen reader support.

core/field_input.ts
BenHenning: Updated in latest.

core/field_image.ts
BenHenning: Updated in latest.

BenHenning (commented):
Self-reviewed latest changes.

> Also, if there's any way you could attach a video demo to this PR, that would help a lot, as I find it difficult sometimes to know exactly what's being read out based on the roles and state. Like somewhere the actual value should be read out, which doesn't seem to be explicitly in any of the manually-set roles, so the video would help me make sense of how it all fits together. Or, in lieu of that, writing out a couple of examples.

I've been trying to minimize videos since they can be time consuming (especially doing good ones), but I did add a quickly made one to the PR description outlining most of the changes in the PR which hopefully helps add some context. I realize it's difficult especially for a PR like this that's affecting so much at one time.

BenHenning:
Everything that was addressed seems largely non-controversial, so I don't think another review pass is needed. The more tricky comments were either closed without further action needed, or tracked in new locations so that they can be addressed outside of this PR.

Issue #9293, opened 2025-08-13
Screen reader issues with toolbox role=tree implementation


microbit-matt-hillsdon:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Work Summary

- [ ] aria-activedescendant is on the wrong element
- [x] grouping structure not correct
- [ ] remove aria-selected=false on the unselected elements
- [ ] investigate and remove "owns" if possible

### Description

Using https://google.github.io/blockly-keyboard-experimentation/screenreader/

Two problems prevent the tree in the demo being accurately represented in the accessibility tree:
1. aria-activedescendent is on the wrong element. It’s on the child of the tree (div.blocklyToolboxCategoryGroup), not the tree itself. It should move to the tree.
2. the grouping structure is not correct
    - CollapsibleCategory should be role=treeitem (Misc in the demo)
    - its contentsContainer should have role=group

Another minor point from reviewing the code: [MDN docs are clear that we should not set aria-selected=false on all the non-selected thing in a single select tree](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role#:~:text=In%20single%2Dselect%20trees%2C%20aria%2Dselected%20is%20set%20to%20true%20for%20the%20selected%20tree%20items%20and%20it%20is%20not%20present%20on%20any%20other%20tree%20item%20in%20the%20tree.).

I am also unsure that the work done to maintain "owns" is necessary (here and elsewhere). The MDN documentation is unclear, but I think it should only be needed if you need to "own" non-descendent nodes. Did you find it to be needed based on practical experience with a screen reader?

### Reproduction steps

1. Use https://google.github.io/blockly-keyboard-experimentation/screenreader/
2. Note that items in the toolbox tree are e.g. 2/8 but there are 9
3. Note that Misc itself is just read as a group not as a tree item

### Stack trace

```shell

```

BenHenning:
Great points--this is a nice issue to track all the bits you've mentioned.

I agree that we can probably remove `aria-owns`, but it will need testing. I don't remember exactly what steps led to that being added, and I need to recheck the DOM structure to make sure it lines up correctly. One confusing bit is exactly how groups and items need to be laid out in a tree in order for its structure to be correctly determined, and I think there were some extra elements that needed to be ignored in the toolbox structure (but that may not actually be correct).

BenHenning:
[...]

PR #9394, opened 2025-10-01
fix: Toolbox & Flyout ARIA positions (experimental)

### LINKED ISSUE #9386: Screen reader - toolbox and flyout position and count readout inconsistencies
### Check for duplicates

### Description

When navigating the toolbox, and at least one flyout, the treeitem / block position and total counts are inconsistent and can be incorrect. See the screen recording below, noting the change from the "p5 blocks" toolbox item, into the "Misc" and "A subcategory" tree items. Also note the readout while navigating through the flyout items when inside the "p5 blocks" flyout.

### Reproduction steps

2. With a screen reader switched on, note the inconsistencies in item position and count while navigating to the last toolbox items, and flyout items.

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/380c8f2a-e0ee-4e38-8947-990c31aec931

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9386
Fixes part of #9293

### Proposed Changes

Addresses #9386 through a number of changes:
- Ensures the flyout contents are reevaluated for ARIA changes whenever they themselves change (since previously `WorkspaceSvg` only recomputed its ARIA properties when one of its blocks self-registered which doesn't account for labels or buttons).
- Collapsible categories are now correctly wrapped by a group (since groups of tree items must be in a parent group).
-  Updates toolbox ARIA computations to be on content changes rather than having items self-specify recomputing the ARIA properties. This mitigates an issue with collapsible categories not updating the toolbox contents and thus being omitted.
- Introduced a separate pathway for computing tree info for flyout workspaces (vs. for the main workspace) in order to account for `FlyoutButton`s.
- Updated `FlyoutButton` to use a nested structure of `treeitem` then `button` in order to actually count correctly in the tree and still be an interactive button. The readout experience is actually better now on ChromeVox, as well, since it reads out _both_ 'Button' and 'Tree item' which is interesting. It seems screen readers are designed to look for this pattern but it only works if set up in a very particular way.

### Reason for Changes

Most of the changes here fixed incidental problems noticed while trying to address #9386 (such as the variables category not correctly accounting for the 'Create variable' button in the count, or not having the correct labels). Much of the count issues in the original issue were caused by a combination of missing some flyout items, and trying to compute the labels too early (i.e. before the categories were fully populated). 

### Test Coverage

Since this is an experimental change, no new tests were added.

### Documentation

No documentation changes are directly needed here.

### Additional Information

None.

### DIFF
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -52,10 +52,7 @@ import type {IBoundedElement} from './interfaces/i_bounded_element.js';
 import {IContextMenu} from './interfaces/i_contextmenu.js';
 import type {IDragTarget} from './interfaces/i_drag_target.js';
 import type {IFlyout} from './interfaces/i_flyout.js';
-import {
-  isFocusableNode,
-  type IFocusableNode,
-} from './interfaces/i_focusable_node.js';
+import {type IFocusableNode} from './interfaces/i_focusable_node.js';
 import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
 import {hasBubble} from './interfaces/i_has_bubble.js';
 import type {IMetricsManager} from './interfaces/i_metrics_manager.js';
@@ -2745,10 +2742,7 @@ export class WorkspaceSvg
         return (
           flyout
             .getContents()
-            .find((flyoutItem) => {
-              const element = flyoutItem.getElement();
-              return isFocusableNode(element) && element.canBeFocused();
-            })
+            .find((flyoutItem) => flyoutItem.getElement().canBeFocused())
             ?.getElement() ?? null
         );
       }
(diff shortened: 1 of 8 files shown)


BenHenning (commented):
Self-review.

core/workspace_svg.ts
```diff
@@ -2947,10 +2947,39 @@ export class WorkspaceSvg
   }
 
   recomputeAriaTree() {
-    // TODO: Do this efficiently (probably incrementally).
-    this.getTopBlocks(false).forEach((block) =>
-      this.recomputeAriaTreeItemDetailsRecursively(block),
-    );
+    // Flyout workspaces require special arrangement to account for items.
```
gonfunko, line 2940: This is fine for now, but note that flyouts can contain arbitrary user-defined kinds of elements, not just buttons and blocks.

gonfunko (approved):

core/workspace_svg.ts
BenHenning: Great point. I've added a TODO to specifically mention this.

BenHenning (commented):
Self-review latest.

BenHenning:
I don't think the latest change needs another review pass since it's just adding a TODO and cleaning up redundant `isFocusableNode` calls in `WorkspaceSvg`. Will go ahead and merge this once CI is done running.

Issue #9387, opened 2025-09-25
Some contribution ideas for Blockly (typings, procedures, docs)


heliacer:
Hi all, I have taken quite a liking to Blockly as I’ve been actively using it on my dream project called Graplet, and it has helped in many ways to think bigger since I first stumbled upon it (a long time ago btw). That’s why I would love to return the favour and contribute in some specific parts that have been bugging me for a while now - and I’d really love to see them in Blockly :>

I’d love to implement typings for the JSON Block Definition. It currently is `any`, passed into `defineBlocksWithJsonArray` Having clear optional identifiers with their corresponding types would help out a bunch, instead of relying on unpredictable string literal properties.

I’ve also been thinking about built-in stuff, such as variable and procedure blocks, and modifying them is still a bit of a hassle. There are many overrides one has to do, as well as unregistering/reregistering, and that kinda feels off.

Having the option to disable all built-in blocks would make things much easier - for example, to create a custom procedure model (e.g. one where parameters are not global variables, but value blocks that can be dragged out of the procedure definition block as a source, kinda similar to how Scratch’s “Custom Blocks” work).

The slight tradeoff would be that many commonly used blocks, such as `math_number`, would have to be recreated - but honestly, I think that’s completely worth it. In fact, I’d say it’s worth wiping the built-in stuff entirely for that tradeoff, if the option is disabled (eg. `useBuiltIn: false`) 

The docs on creating custom procedures are also a bit misleading. They have incomplete/false code snippets, and figuring out the actual way to implement ObservableProcedureModel is quite the challenge…

I hope some of those ideas and possible contributions align with Blockly’s ideals. I would really love to work on this and see it live in action :D (and thanks for reading that big message wall up until here lol).

heliacer:
Hey, quick correction to my first post:
When I mentioned `BlockDefinition`, what I really meant was the JSON block definition (the one passed into `defineBlocksWithJsonArray`).

`BlockDefinition` itself is the object with functions (like when you call `defineBlocks({ ... })`). But for the JSON shape, it’s currently still typed as `AnyDuringMigration`. That’s the part I was actually talking about - having a proper type/interface there (instead of `any`) would make manual block creation a lot smoother, with clear optional identifiers and their types instead of relying on string literals.

maribethb:
[...]

PR #9402, opened 2025-10-03
JSON Block Definition interface

heliacer:
## The basics

### Resolves

Implements JSON Block Definition interface proposed in #9387

### Proposed Changes
Introduces a new `JsonBlockDefinition` in `core/interfaces/`

- Added `core/interfaces/i_json_block_definition.ts` with `JsonBlockDefinition` and `BlockArg` definitions
- Migrated `Block.jsonInit`, `jsonInitFactory`, `defineBlocksWithJsonArray` and `createBlockDefinitionsFromJsonArray` to use the new interface
- Migrated `interpolate` to use the new `BlockArg` type
- Updated all built-in block JSON definitions in `blocks/` to explicitly set `'output': null` to match the default behaviour in the block factory
- Removed legacy `AnyDuringMigration` usage for JSON block definitions

### Reason for Changes

Makes it easier to write valid JSON Block definitions without having to always rely on the block factory

### Test Coverage

I ran the unit test suite and manually loaded the playground with the updated blocks to ensure they still render and connect correctly

### Documentation
Not needed

### Additional Information

I found another issue while testing (unrelated to this) in the default Playground, when `Categories: (typed variables)` is selected, the colour category errors out because of an invalid `colour_picker` definition


heliacer:
@gonfunko  bump, will this be considered? 👀

gonfunko:
Yes, sorry for the delay on this! Will try to get you an initial review this week, thank you for the ping on it.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,97 @@
...
+  width: number;
+  height: number;
+  alt?: string;
+  flipRtl?: boolean | 'TRUE' | 'FALSE';
+}
+
+interface FieldVariableArg extends CommonArg {
+  type: 'field_variable'
+  variable: string | null
+}
```
gonfunko: Each of the Field classes defines and exports a `FieldXYZFromJsonConfig` interface, which I think should be used in place of these to avoid duplication/skew.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface JsonBlockDefinition {
+  type?: string;
+  style?: string;
+  colour?: string | number;
+  output: string | string[] | null;
```
gonfunko: What was the motivation for making `output` mandatory while allowing `previousStatement` and `nextStatement` to be omitted entirely? In principle they should all be optional, but if there's code that's relying on `output` being there I think that would be the better thing to fix.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,97 @@
...
+  outputShape?: number;
+  inputsInline?: boolean;
+  tooltip?: string;
+  helpUrl?: string;
+  extensions?: string[];
+  mutator?: string;
+  enableContextMenu?: boolean;
+  suppressPrefixSuffix?: boolean;
+
+  [key: `message${number}`]: string | undefined;
```
gonfunko, line 49: This is clever! TIL about this TS mechanism

gonfunko (commented):
This looks promising; added a couple notes. I'd like to run this by my colleagues before merging, but as you may have seen we're transitioning from Google to the Raspberry Pi Foundation so it may be a couple weeks until things are settled.

core/interfaces/i_json_block_definition.ts
heliacer: ohh, didn't know these configs existed 👀

```ts
interface FieldInputArg extends FieldTextInputFromJsonConfig {
  name: string
  type: 'field_input'
}

interface FieldNumberArg extends FieldNumberFromJsonConfig {
  name: string
  type: 'field_number'
}
// ...
```
wdyt 🤔

core/interfaces/i_json_block_definition.ts
heliacer: Good question LOL, well I had thought that the block factory had always included the output, whether it is null or not.
Turns out I was completely false, so leaving it optional is obviously better 🙂

and no, there shouldn't be code relying on the output -  `jsonInit` intentionally checks if `output` is not undefined, just like it does with previous/nextStatement

heliacer:
converted this to draft, will implement @gonfunko's review feedback sooner or later :>

core/interfaces/i_json_block_definition.ts
heliacer: @gonfunko bump again, hope this finds you well since there was all the migration from google to rpf and stuff 👀 
I would definitely want to include your suggested changes, but before I do, can you run this and lmk if it works? 
Thanks 😃

gonfunko: Thanks for the bump, and yes, we're all moved over and all is well. The approach to the interfaces you proposed in your prior comment makes sense to me, so updating it along those lines sounds good.

core/interfaces/i_json_block_definition.ts
gonfunko: No worries! Let's just make it optional then.

heliacer:
The following new changes have been made:
- implemented @gonfunko 's idea to use the existing `FieldXYZFromJsonConfig ` interfaces
- removed unnecessary `output: null` which have been added by mistake

blocks/variables.ts
```diff
@@ -6,21 +6,21 @@
 
 // Former goog.module ID: Blockly.libraryBlocks.variables
 
-import type {Block} from '../core/block.js';
+import type { Block } from '../core/block.js'
```
gonfunko: We do want the semicolons here; I saw that Copilot flagged a few other things, you should be able to just do `npm run format` to get everything aligned with project conventions.

blocks/variables.ts
```diff
@@ -60,18 +60,19 @@ export const blocks = createBlockDefinitionsFromJsonArray([
     ],
     'previousStatement': null,
     'nextStatement': null,
+    'output': null,
```
gonfunko: Output isn't needed here; this is the setter, which is just a statement block without any output value.

blocks/variables_dynamic.ts
```diff
@@ -59,20 +58,21 @@ export const blocks = createBlockDefinitionsFromJsonArray([
         'name': 'VALUE',
       },
     ],
+    'output': null,
```
gonfunko, line 61: As with the non-dynamic variant, the setter block shouldn't have an output.

core/block.ts
```diff
@@ -1732,87 +1733,78 @@ export class Block {
     }
 
     // Set basic properties of block.
-    // Makes styles backward compatible with old way of defining hat style.
-    if (json['style'] && json['style'].hat) {
-      this.hat = json['style'].hat;
-      // Must set to null so it doesn't error when checking for style and
-      // colour.
-      json['style'] = null;
-    }
```
gonfunko, line 1741: We need to leave this backfilling of hats in place. I don't think that `style: string | {hat: string}` should be in the types (`style` can just be typed as `string` as you have it presently), but we should still check and fix it here, casting as necessary to allow checking for `style.hat`.

core/block.ts
```diff
@@ -1732,87 +1733,78 @@ export class Block {
...
 
     // Interpolate the message blocks.
     let i = 0;
-    while (json['message' + i] !== undefined) {
+    while (json[`message${i}`] !== undefined) {
       this.interpolate(
-        json['message' + i],
-        json['args' + i] || [],
-        // Backwards compatibility: lastDummyAlign aliases implicitAlign.
-        json['implicitAlign' + i] || json['lastDummyAlign' + i],
```
gonfunko, line 1758: This backwards compatibility shim also needs to stay. As with hats, the types shouldn't reflect the old behavior, but we should still check for it here.

core/block.ts
```diff
@@ -1826,12 +1818,12 @@ export class Block {
    * @param json Structured data describing the block.
    * @param warningPrefix Warning prefix string identifying block.
    */
-  private jsonInitColour(json: AnyDuringMigration, warningPrefix: string) {
-    if ('colour' in json) {
-      if (json['colour'] === undefined) {
+  private jsonInitColour(json: JsonBlockDefinition, warningPrefix: string) {
+    if (json.colour) {
```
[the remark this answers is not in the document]
gonfunko: Please revert to using the `in` check.

core/common.ts
```diff
@@ -13,6 +13,7 @@ import type {Connection} from './connection.js';
 import {EventType} from './events/type.js';
 import * as eventUtils from './events/utils.js';
 import {getFocusManager} from './focus_manager.js';
+import {JsonBlockDefinition} from './interfaces/i_json_block_definition.js';
```
gonfunko: This can be `import type`.

core/block.ts
```diff
@@ -54,6 +54,7 @@ import * as idGenerator from './utils/idgenerator.js';
 import * as parsing from './utils/parsing.js';
 import {Size} from './utils/size.js';
 import type {Workspace} from './workspace.js';
+import { BlockArg, JsonBlockDefinition } from './interfaces/i_json_block_definition.js'
```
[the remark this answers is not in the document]
gonfunko: Please update this to `import type` as Copilot suggested.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import {FieldCheckboxFromJsonConfig} from '../field_checkbox.js';
+import {FieldDropdownFromJsonConfig} from '../field_dropdown';
+import {FieldImageFromJsonConfig} from '../field_image';
+import {FieldNumberFromJsonConfig} from '../field_number';
+import {FieldTextInputFromJsonConfig} from '../field_textinput';
+import {FieldVariableFromJsonConfig} from '../field_variable';
+
+export interface JsonBlockDefinition {
+  type?: string;
```
gonfunko: `type` should be required; we can't do anything with a block definition that doesn't specify what type of block it's defining.

gonfunko (commented):
Thanks! If you could run `npm run format` and commit the results that should take care of aligning the formatting with our standards and reduce the diff size substantially.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,95 @@
...
+  helpUrl?: string;
+  extensions?: string[];
+  mutator?: string;
+  enableContextMenu?: boolean;
+  suppressPrefixSuffix?: boolean;
+
+  [key: `message${number}`]: string | undefined;
+  [key: `args${number}`]: BlockArg[] | undefined;
+  [key: `implicitAlign${number}`]: string | undefined;
+}
```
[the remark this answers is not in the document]
heliacer: will do

core/interfaces/i_json_block_definition.ts
heliacer: agree, but there are snippets like these which would no longer be accepted, and would require an additional `type` even though it is is given by assigning `blocks` itself:
```ts
blocks['text_reverse'] = {
  /**
   * Block for reversing a string.
   */
  init: function (this: Block) {
    this.jsonInit({
      'message0': Msg['TEXT_REVERSE_MESSAGE0'],
      'args0': [
        {
          'type': 'input_value',
          'name': 'TEXT',
          'check': 'String',
        },
      ],
      'output': 'String',
      'inputsInline': true,
      'style': 'text_blocks',
      'tooltip': Msg['TEXT_REVERSE_TOOLTIP'],
      'helpUrl': Msg['TEXT_REVERSE_HELPURL'],
    });
  },
};
```

BenHenning:
Since @gonfunko is already looking at this I don't think I need to. Seems like the PR assigner rotation thing kinda goofed during the transition, but not a big deal.

core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,114 @@
...
+interface InputStatementArg {
+  type: 'input_statement';
+  name?: string;
+  check?: string | string[];
+}
+interface InputDummyArg {
+  type: 'input_dummy';
+  name?: string;
+}
```
maribethb, line 82: There's also an end-row input that should probably be accounted for

maribethb (commented):
Thank you for the pull request! It's great to make some progress in this area. I think there's more work before this is ready to go in, however.

1. This is a promising start for typing the built-in block definitions, but we need developers to be able to customize it more. For example, if a developer has defined a custom input or field, currently with this typing there is no way for them to provide a valid typed json object because `BlockArg` type is not going to include their custom object. We need a solution for this.
2. Because of (1) I would like to see a test case in `tests/typescript` that registers a custom block using json that uses the custom field already in that test suite (`FieldMitosis`)
3. This is going to be a breaking change for typescript developers if `jsonInit` and related methods **require** their arguments to conform to this type. I think the best path forward would be to include this interface now but leave the arguments to the various init/define functions as `AnyDuringMigration`, so that developers can go ahead and update their code to use the new type when they want without being required to do so. In the next major release we could then make the breaking change of requiring the types. This would also let us soft-launch the types and test/change them before we require their use.
4. You've changed from bracket access to dot access in some of the block initialization code, but I think you should revert those changes. The reason for the bracket access is because Blockly is compatible with closure compiler's advanced compilation mode, which does a lot of property renames. The bracket access says not to rename those properties, which is important if your block definitions are not also compiled with closure compiler. Overall I think block definitions are likely to be compiled, unlike, say, the json from a serialized workspace, where it's important we continue to use bracket access. And, I would guess that approximately nobody except blockly-games (which does not even use this version of blockly) is using advanced compilation anyway. At some point we may decide that we don't care about maintaining that compatibility. But for now, I think there is some risk in removing this (however small) and not much upside. So I would recommend that you revert those changes. They are not really relevant to this change anyway.

heliacer:
tests for the custom input / fields will be added soon :D

heliacer:
FYI: This is still being worked on (there was a break in-between due to some personal things)
The final changes, including @maribethb's review requests will be squashed into one commit :D

What's left: create support for the custom fields (wip)

Issue #9312, opened 2025-08-15
Improve aria-label for mutator workspaces


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

Mutator workspaces have role="group" and aria-label="Bubble". In VoiceOver this is read out as "Bubble, group". It would be nice if this label could be more user-friendly / descriptive. One of the reasons for this labelling to be made clear is that triggering the mutator workspace via the mutator icon also moves focus into the mutator workspace, which makes it more important for the user to know where they are / have moved to.

### Reproduction steps

1. Go to the [screen reader test page](https://google.github.io/blockly-keyboard-experimentation/screenreader/?scenario=navigationTestBlocks&toolbox=toolbox&renderer=zelos)
2. Navigate to the one of the if block mutator icons and open it with Enter or Space
3. The screen reader will announce something like "If, selected, Bubble, group" - there might be some variation between different screen readers.

### Stack trace

```shell

```

BenHenning:
It seems like the bubble is being focused before the workspace, so there may need to be different labels depending on the context or we need to consider simplifying the navigation flow here so that it goes straight to the workspace.

Separately, the workspace's label can be made a bit clearer in the mutator context. I didn't spend much time on mutators because I wasn't sure what we wanted the terminology to be for them, yet, but we can at least fix on being consistent with the term 'mutator' and adjust that term in particular once we have more testing data.

If we have to keep the bubble and workspace as separate navigation items, then we could at least distinguish between them with 'Mutator bubble' and 'Mutator workspace' until we can further refine based on user feedback. I suspect both the terms 'bubble' and 'mutator' are going to be confusing and need adjustments.

maribethb:
The bubble should be a separate navigation item for 2 reasons:

- Mutators can remain open on the page, and you need to be able to continue navigating "around" the bubble without getting trapped inside the workspace. See https://github.com/google/blockly-keyboard-experimentation/issues/723 for where we don't separately focus the bubble and you can get "trapped" inside the comment. If we didn't focus the bubble separately, the same problem would apply here.
- Bubbles are movable with the mouse and we may wish to make them movable with the keyboard as well, and they would need to be focused in order to do that.

I think "mutator" shouldn't be a user-facing term and would use something like "edit this block" but not sure how to turn that into a noun, since "block editor" is ambiguous and is what I would suggest for the default workspace region label :)

BenHenning:
[...]

PR #9412, opened 2025-10-08
feat: Remove main workspace tree/treeitem & use top-level regions (experimental)

### LINKED ISSUE #428: Add the ARIA: region role to the root of the Blockly workspace.

### LINKED ISSUE #9395: Remove positional and level information for blocks in screen reader output
### Check for duplicates

### Problem

Discussion this week suggested that screen reader output should be simplified and not need positional information.

Separately, `treeitem`s have caused compatibility issues on Safari with VoiceOver.

### Request

Remove `Workspace`'s `treeitem` usage for blocks, though the version for `Flyout` can maybe be kept (more thought is needed there).

### Alternatives considered

_No response_

### Additional context

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9395
Fixes https://github.com/google/blockly-keyboard-experimentation/issues/428
Fixes part of #9312

### Proposed Changes

Removes all `tree` and `treeitem` roles from non-flyout workspaces, and uses `region` for top-level workspace, toolbox, and flyout roles.

Demonstration of changes:

[Screen recording 2025-10-08 10.37.10 AM.webm](https://github.com/user-attachments/assets/a42084b2-06fa-48a2-9661-710c630f3ba0)

(This shows the reintroduction of block role description, the new regions, the removal of tree/tree items for blocks, and an attempt at fixing mutators though with apparent issues).

### Reason for Changes

Recent feedback from micro:bit panelists and accessibility experts suggests that the tree-based approach may be adding too much contextual information to the point of not actually being helpful (and potentially even being harmful from a UX perspective). Additionally, VoiceOver has mixed support for trees and tree items and it may be very difficult or impossible to work around these and continue with those roles.

The new plan is to test with less explicit context and add some more context through block labeling (particularly for C-shaped blocks), which is ongoing work in #9299. Longer term it may be necessary to add a more 'verbose' readout mode, or add some sort of 'where am I?' shortcut to re-add the contextual positioning information (so the code removed in this PR may be useful for future referencing).

Some specific notes on regions:
- Regions provide explicit readout context (at least in ChromeVox) usually when entered and exited. This actually makes it possible to leverage both regions and trees for context (in the case of toolboxes and flyouts) which is actually quite clean.
- Mutators are still very fuzzy. Mutators are being made a region since they are technically distinct workspaces from the main workspace, but the readouts for them on ChromeVox aren't working as expected which is why #9312 isn't being fully closed here. More work is needed to get mutators to have a good readout experience.

Note also that this reintroduces the 'block' role description for the main workspace and is leveraging `figure` (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/figure_role) as the base role for blocks. This _seems_ like the best fit as blocks are not inherently buttons, and overwriting the role description is supposed to provide the correct context to readers. This may well need to change if more incompatibilities are discovered (such as with VoiceOver).

### Test Coverage

None since this is an experimental change.

### Documentation

None since this is an experimental change.

### Additional Information

None

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -285,40 +285,16 @@ export class BlockSvg
   private computeAriaRole() {
     if (this.isSimpleReporter()) {
       aria.setRole(this.pathObject.svgPath, aria.Role.BUTTON);
-    } else {
-      // This isn't read out by VoiceOver and it will read in the wrong place
-      // as a duplicate in ChromeVox due to the other changes in this branch.
-      // aria.setState(
-      //   this.pathObject.svgPath,
-      //   aria.State.ROLEDESCRIPTION,
-      //   'block',
-      // );
+    } else if (this.workspace.isFlyout) {
       aria.setRole(this.pathObject.svgPath, aria.Role.TREEITEM);
-    }
-  }
-
-  collectSiblingBlocks(surroundParent: BlockSvg | null): BlockSvg[] {
-    // NOTE TO DEVELOPERS: it's very important that these are NOT sorted. The
-    // returned list needs to be relatively stable for consistent block indexes
-    // read out to users via screen readers.
-    if (surroundParent) {
-      // Start from the first sibling and iterate in navigation order.
-      const firstSibling: BlockSvg = surroundParent.getChildren(false)[0];
-      const siblings: BlockSvg[] = [firstSibling];
-      let nextSibling: BlockSvg | null = firstSibling;
-      while ((nextSibling = nextSibling?.getNextBlock())) {
-        siblings.push(nextSibling);
-      }
-      return siblings;
     } else {
-      // For top-level blocks, simply return those from the workspace.
-      return this.workspace.getTopBlocks(false);
+      aria.setState(
+        this.pathObject.svgPath,
+        aria.State.ROLEDESCRIPTION,
+        'block',
+      );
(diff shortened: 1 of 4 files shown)


core/block_svg.ts
```diff
@@ -285,42 +285,18 @@ export class BlockSvg
...
-    } else {
-      // This isn't read out by VoiceOver and it will read in the wrong place
-      // as a duplicate in ChromeVox due to the other changes in this branch.
-      // aria.setState(
-      //   this.pathObject.svgPath,
-      //   aria.State.ROLEDESCRIPTION,
-      //   'block',
-      // );
+    } else if (this.workspace.isFlyout) {
       aria.setRole(this.pathObject.svgPath, aria.Role.TREEITEM);
```
maribethb, line 289: I think it's kind of weird for the flyout workspaces to still be treeitems. The categories in a toolbox make sense as trees, but blocks in a flyout should probably be treated like blocks in the workspace, in my opinion.

maribethb (approved):

core/block_svg.ts
BenHenning: Yeah that seems right. It seems maybe useful to treat it like a list of some kind, but we may be stretching the definition of list item here. I can just go back to the idea of no list entirely and see what testing brings up.

gonfunko:
Talked with @rachel-fenichel and decided to merge now to unblock further work.

PR #9414, opened 2025-10-09
fix(ci): pass PR title to commitlint via environment variable

AdnaneKhan:
## The basics

The commitlint workflow currently uses the PR title via actions context variable which can be used to pass code. Fortunately the GITHUB_TOKEN is scoped down, but this can turn into a problem later if a privileged workflow (such as one that has contents:write) consumes from the cache. This PR fixes that risk.

### Proposed Changes

Sanitize PR title via env var.

### Reason for Changes

Fix actions injection.

### Test Coverage

N/A

### DIFF
--- a/.github/workflows/conventional-label.yml
+++ b/.github/workflows/conventional-label.yml
@@ -7,6 +7,8 @@ name: commit lint & label
 jobs:
   lint:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


gonfunko (approved):

gonfunko:
Thanks for the catch and fix!

PR #9416, opened 2025-10-10
fix: Improve narration and navigation of C-shaped blocks.

### LINKED ISSUE #9341: Up/down navigation on value blocks skips to next statement block
### Check for duplicates

### Description

The current navigation policy system is based on (and intended to emulate) the previous `LineCursor` implementation in google/blockly-keyboard-experimentation, which in turn was intended to emulate the behaviour of the cursor in a text editor:

* Using left and right would visit every possible node ("character"), in pre-order traversal of the block forest, while
* Using up and down would skip to a node on the previous or next "line".

Currently the navigation policy implements this very well for statement blocks, but not so well for value blocks (notably: ones with external inputs): when visiting a value block connected to the external input of another value block, where there are additional blocks connected to the same parent above/below, the up and down keys move not to the next value block in the given direction, but to the next statement block.

### Reproduction steps

1. Add a value block with multiple external inputs to the workspace.
2. Add an additional value block to each of the external inputs  of the first block.
3. Select one of the additional blocks.
4. Press up or down arrow key.

### Screenshots

Note that in this screen capture I use only the up and down arrow keys:

https://github.com/user-attachments/assets/06ca5043-6b12-4932-b371-6f1b7baeb990

### Browsers

Chrome desktop

### LINKED ISSUE #9407: screenreader improvements for statement input blocks
### Check for duplicates

### Problem

_No response_

### Request

When a next connection without child blocks is focused (see google/blockly-keyboard-experiment#754), the screenreader label should reflect that the current location is the end of a statement input. exact terminology TBD, pick something for us to debate over in testing!

When the first block of a statement input is focused, announce that it's the first block of a statement input. May need additional context such as any labels that precede the statement input.

May need special case text for empty inputs.

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #9415: Visit connection locations for c-shaped blocks
### Check for duplicates

### Problem

_No response_

### Request

- Visit next connections that don't have child blocks (when inside statement inputs)
- Visit value inputs between statement inputs (possibly needs some exploration to make this feel natural, e.g. do you visit a value input that's after a statement input? visit all of them but the first? etc.)

This is what we think based on discussion only, not necessarily the final decision for behavior. We need to do something sensible and it's ok if you find an edge case and address it while implementing this. The goal is to have a starting point for discussion, not to adhere strictly to what I've typed above :)

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #9415
* Fixes #9407 
* Fixes #9341

### Proposed Changes
This PR makes several changes to improve the keyboard navigation and screenreader narration of blocks, particularly C-shaped blocks and external inputs. Specifically, it:

* Visits empty statement inputs via keyboard navigation, and reads them out
* Visits the last (empty) next connection in a stack of blocks inside of a statement input, and reads it out as ending the statement input
* Visits blocks or empty value inputs of clauses in between statement inputs, e.g. "else if <>" conditions. The first ("if") condition is not visited when navigating down from the C/E shaped block, because its contents are read out as part of the top-level block
* Adjusts the readout for the first block in a statement input to note that it is beginning a new scope
* Makes the down arrow in keyboard navigation behave more naturally when a block connected to an external input is focused, by moving to the next block attached to the parent's next external input

https://github.com/user-attachments/assets/28f290eb-30cc-40c4-aeda-32de89bc2be5

### DIFF
--- a/core/keyboard_nav/block_navigation_policy.ts
+++ b/core/keyboard_nav/block_navigation_policy.ts
@@ -125,6 +125,13 @@ function getBlockNavigationCandidates(
   for (const input of block.inputList) {
     if (!input.isVisible()) continue;
     candidates.push(...input.fieldRow);
+    if (
+      input.connection?.type === ConnectionType.NEXT_STATEMENT &&
+      !input.connection.targetBlock()
+    ) {
+      candidates.push(input.connection as RenderedConnection);
+    }
+
     if (input.connection?.targetBlock()) {
       const connectedBlock = input.connection.targetBlock() as BlockSvg;
       if (input.connection.type === ConnectionType.NEXT_STATEMENT && !forward) {
@@ -140,6 +147,15 @@ function getBlockNavigationCandidates(
     } else if (input.connection?.type === ConnectionType.INPUT_VALUE) {
       candidates.push(input.connection as RenderedConnection);
     }
+  }
+
+  if (
+    block.nextConnection &&
+    !block.nextConnection.targetBlock() &&
+    (block.lastConnectionInStack(true) !== block.nextConnection ||
+      !!block.getSurroundParent())
+  ) {
+    candidates.push(block.nextConnection);
   }
 
   return candidates;
--- a/core/keyboard_nav/block_navigation_policy.ts
+++ b/core/keyboard_nav/block_navigation_policy.ts
@@ -124,28 +124,34 @@ function getBlockNavigationCandidates(
 
   for (const input of block.inputList) {
     if (!input.isVisible()) continue;
-    candidates.push(...input.fieldRow);
-    if (
-      input.connection?.type === ConnectionType.NEXT_STATEMENT &&
-      !input.connection.targetBlock()
-    ) {
-      candidates.push(input.connection as RenderedConnection);
-    }
 
-    if (input.connection?.targetBlock()) {
-      const connectedBlock = input.connection.targetBlock() as BlockSvg;
-      if (input.connection.type === ConnectionType.NEXT_STATEMENT && !forward) {
+    candidates.push(...input.fieldRow);
+
+    const connection = input.connection as RenderedConnection | null;
+    if (!connection) continue;
+
+    const connectedBlock = connection.targetBlock();
+    if (connectedBlock) {
+      if (connection.type === ConnectionType.NEXT_STATEMENT && !forward) {
         const lastStackBlock = connectedBlock
           .lastConnectionInStack(false)
           ?.getSourceBlock();
         if (lastStackBlock) {
+          // When navigating backward, the last block in a stack in a statement
+          // input is navigable.
           candidates.push(lastStackBlock);
         }
       } else {
+        // When navigating forward, a child block connected to a statement
+        // input is navigable.
         candidates.push(connectedBlock);
       }
-    } else if (input.connection?.type === ConnectionType.INPUT_VALUE) {
-      candidates.push(input.connection as RenderedConnection);
+    } else if (
+      connection.type === ConnectionType.INPUT_VALUE ||
+      connection.type === ConnectionType.NEXT_STATEMENT
+    ) {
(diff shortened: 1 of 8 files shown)


core/keyboard_nav/block_navigation_policy.ts
```diff
@@ -125,6 +125,13 @@ function getBlockNavigationCandidates(
   for (const input of block.inputList) {
     if (!input.isVisible()) continue;
     candidates.push(...input.fieldRow);
+    if (
+      input.connection?.type === ConnectionType.NEXT_STATEMENT &&
```
rachel-fenichel: Nit: can pull `input.connection` out into a local variable (and do the cast) at line 124 to make this and the following code more concise.

core/keyboard_nav/block_navigation_policy.ts
```diff
@@ -125,6 +125,13 @@ function getBlockNavigationCandidates(
   for (const input of block.inputList) {
     if (!input.isVisible()) continue;
     candidates.push(...input.fieldRow);
+    if (
```
rachel-fenichel: Add a comment explaining which case this is. (Empty statement input, I believe)

core/keyboard_nav/block_navigation_policy.ts
```diff
@@ -142,6 +149,15 @@ function getBlockNavigationCandidates(
     }
   }
 
+  if (
```
rachel-fenichel, line 158: Add a comment explaining which case this is (last connection in a substack, I believe)

core/rendered_connection.ts
```diff
@@ -334,7 +335,30 @@ export class RenderedConnection
...
+      } else if (
+        this.type === ConnectionType.INPUT_VALUE &&
+        this.getSourceBlock().statementInputCount
+      ) {
+        aria.setState(
+          highlightSvg,
+          aria.State.LABEL,
+          `${this.getParentInput()?.getFieldRowLabel()}`,
+        );
+      }
```
rachel-fenichel, line 363: Shouldn't there be an else clause with the prior code (`aria.setState(highlightSvg, aria.State.LABEL, 'Open connection');`)? These two cases don't cover everything.

core/keyboard_nav/line_cursor.ts
```diff
@@ -298,6 +305,82 @@ export class LineCursor extends Marker {
...
+            // If the candidate's parent uses inline inputs, disallow the
+            // candidate; it follows that it must be on the same row as its
+            // parent.
+            if (candidate.outputConnection?.targetBlock()?.getInputsInline()) {
+              return false;
+            }
+
+            const candidateParents = new Set<BlockSvg>();
+            let parent = candidate.getParent();
+            while (parent) {
```
rachel-fenichel: Break out the code get building the parent sets into helpers.

rachel-fenichel (commented):
One functionality question + some cleanup/commenting requests.

core/keyboard_nav/line_cursor.ts
gonfunko: Done.

core/rendered_connection.ts
gonfunko: Done.

core/keyboard_nav/block_navigation_policy.ts
gonfunko: Done.

core/keyboard_nav/block_navigation_policy.ts
gonfunko: Done.

core/keyboard_nav/block_navigation_policy.ts
gonfunko: Done.

rachel-fenichel (approved):

PR #9424, opened 2025-10-15
fix: Miscellaneous improvements for screenreader support.

### LINKED ISSUE #9296: Aria label for blocks includes non-visible inputs and doesn't update when visibility changes
### Check for duplicates

### Description

MakeCode has blocks that use icons/images to allow their structure to be changed, e.g. 

Collapsed:
Image

Expanded:
Image

The icons toggle the visibility of inputs.

Aria labels set by the add-screen-reader-support-experimental branch are set in the constructor and never updated. They also don't take into account the visibility of the inputs.

As blocks are generally mutable there may be other points at which the aria label needs to update. One option might be to update it last thing, before render.

### Reproduction steps

1. Use the MakeCode deployment with screen reader support at https://sr-preview.review-pxt.pages.dev/
2. Example block is in Leds category
3. Note aria label initially includes text you cannot see

Or open this shared project which has some example blocks that change: https://sr-preview.review-pxt.pages.dev/#pub:_110DWocKLWdW

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #9316: Screen reader output for full block fields

### LINKED ISSUE #9420: Screen reader - dropdown output bug
### Check for duplicates

### Description

The screen reader output for dropdown fields is limited to the text shown for the current value for VoiceOver and NVDA. Only ChromeVox provides information that informs the user that these are dropdowns.

Block example:
Image

Screen reader output for the field highlighted above:
ChromeVox: field_dropdown first Collapsed Combo box
VoiceOver: first, group
NVDA: first, section
JAWS: end group, first

This change in output is the result of setting the `aria-hidden` value of the text element to `false` - see [this changed line in the linked PR](https://github.com/google/blockly/pull/9384/files#diff-af2458fcf0c5427418a18bfcc195a5f27fb6177b212fdc523f31c077819035f1R646). While this improves the output with ChromeVox, it degrades the readout for other screen readers to the point where it's unclear that this field is a dropdown / combo box. We will check the state of JAWS soon.

When restoring aria-hidden to the text element, the readouts for VoiceOver and NVDA change to the following:
VoiceOver: list box pop-up collapsed, combo box
NVDA: field_dropdown combo box collapsed opens list

This provides context but is now missing the current value.

As an aside, dropdown fields and input(?) blocks that are full block fields with a dropdown field (see image below) are treated differently in terms of their readout. In the case below, all screen readers output "micro, input block, button". I'm not sure whether the plan is to align the readout for these cases, or treat them separately? Can we use aria-haspopup to make the case below clearer? It looks like the correct semantics are in an SVG group sibling to the button at the moment which doesn't take focus. Is it possible to surface this information some way?

Image

It would be really valuable if the the first part of this could be addressed in time for our initial user testing scheduled for early November.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #9423: Screen reader - integrate field name into aria-label
### Check for duplicates

### Problem

For full block fields, the screen reader output doesn't include the field name. This is similar to focusing on a text input that doesn't have a name attribute or label. It would be really beneficial to have this added context for users.

### Request

Is it possible to include the current field name in the aria-label for full block and other fields so we can construct [value, field name, other role info]? The field name is often in the DOM as the aria-label for a sibling element to the one that takes focus. This may share some detail with https://github.com/google/blockly/issues/9420.

The current readout for the block below is "micro, input block, button". It would be more informative if it was possible to include the current field name so we get construct [value, field name, role info]. In this case, it would be "micro, TEXT, input block, button". There are more complex cases in MakeCode where the field name would be even more valuable.

Image

As these field names are not typically human-friendly, it would be great if there was a way to provide text that could be used in place of the default field name to improve this readout.

### Alternatives considered

_No response_

### Additional context

It would be good if we were able to user test with something like this in early November, but appreciate that the timeline is tight.

### LINKED ISSUE #9425: Screen reader - begin [name] connection for functions (and other blocks)
### Check for duplicates

### Description

When moving into the scope of a function, the "begin [name] connection" readout is missing its name. Calling `parentInput.getFieldRowLabel()` works where the statement input is inline with text such as the `setup` and `draw` blocks, as well as blocks with the text "do", but doesn't work for functions, or many blocks in MakeCode where the text is not in the same row as the statement input, e.g., onStart.

Image

Image

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #9420
* Fixes #9423
* Fixes #9316 (again)
* Fixes #9296
* Fixes #9425

### Proposed Changes
This PR is a grab bag of improvements to the screenreader experience. Specifically:

* Shadow blocks are now noted as "replaceable blocks" and input blocks are noted as such; this was previously reversed.
* Disabled and collapsed blocks are noted as such
* Hidden fields/inputs are excluded from the description of a block
* The count of child blocks is accurate for multi-statement-input blocks, e.g. if/else if/else.
* ARIA labels for blocks are computed lazily on focus, instead of multiple times upfront during the rendering process.
* Simple reporter blocks/full block fields have had their readout improved. Previously, they were described as buttons, regardless of the nature of the field. Now, the block returns its single child field's focusable element as its own focusable element. This means that, from the perspective of the focus manager/keyboard navigation, the block is the focused element (as has always been the case) but the HTML element with focus is that of the field, along with its ARIA role and attributes, so from the screenreader's perspective the field is focused and described as if it were a normal field.
* "Begin X" and "End X" labels for statement input connections/blocks now look back to the previous input line to generate a description of the clause that is beginning/ending, useful for blocks that do not include a label on the same row as the statement input itself
* Extraneous "group" readout on VoiceOver when navigating menu items has been removed
* Text input fields now use `role="button"`, since they must be activated before they can be edited, which the old `role="textbox"` did not make clear.
* `FieldDropdown.recomputeAria()` has been made `protected`, which unblocks fixing `FieldColour`'s readout.
* Fields now include their name in the ARIA label. By default this is pretty bad as this is not a human-readable value, but an `ariaName` may (and should) be included in the field's config to specify a useful name/context.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -226,7 +226,12 @@ export class BlockSvg
     this.computeAriaRole();
   }
 
-  private recomputeAriaLabel() {
+  /**
+   * Updates the ARIA label of this block to reflect its current configuration.
+   *
+   * @internal
+   */
+  recomputeAriaLabel() {
     if (this.isSimpleReporter()) {
       const field = Array.from(this.getFields())[0];
       if (field.isFullBlockField() && field.isCurrentlyEditable()) return;
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -227,6 +227,11 @@ export class BlockSvg
   }
 
   private recomputeAriaLabel() {
+    if (this.isSimpleReporter()) {
+      const field = Array.from(this.getFields())[0];
+      if (field.isFullBlockField() && field.isCurrentlyEditable()) return;
+    }
+
     aria.setState(
       this.getFocusableElement(),
       aria.State.LABEL,
@@ -240,32 +245,40 @@ export class BlockSvg
       ? ` ${inputCount} ${inputCount > 1 ? 'inputs' : 'input'}`
       : '';
 
-    let currentBlock: Block | null = null;
+    let currentBlock: BlockSvg | null = null;
     let nestedStatementBlockCount = 0;
-    // This won't work well for if/else blocks.
-    this.inputList.forEach((input) => {
+
+    for (const input of this.inputList) {
       if (
         input.connection &&
         input.connection.type === ConnectionType.NEXT_STATEMENT
       ) {
-        currentBlock = input.connection.targetBlock();
+        currentBlock = input.connection.targetBlock() as BlockSvg | null;
+        while (currentBlock) {
+          nestedStatementBlockCount++;
+          currentBlock = currentBlock.getNextBlock();
+        }
       }
-    });
-    // The type is poorly inferred here.
-    while (currentBlock as Block | null) {
-      nestedStatementBlockCount++;
-      // The type is poorly inferred here.
-      // If currentBlock is null, we can't enter this while loop...
-      currentBlock = currentBlock!.getNextBlock();
     }
 
     let blockTypeText = 'block';
     if (this.isShadow()) {
-      blockTypeText = 'input block';
-    } else if (this.outputConnection) {
       blockTypeText = 'replacable block';
+    } else if (this.outputConnection) {
+      blockTypeText = 'input block';
     } else if (this.statementInputCount) {
       blockTypeText = 'C-shaped block';
+    }
+
+    const modifiers = [];
+    if (!this.isEnabled()) {
+      modifiers.push('disabled');
+    }
+    if (this.isCollapsed()) {
+      modifiers.push('collapsed');
+    }
+    if (modifiers.length) {
+      blockTypeText = `${modifiers.join(' ')} ${blockTypeText}`;
     }
 
     let prefix = '';
@@ -298,9 +311,7 @@ export class BlockSvg
   }
 
   private computeAriaRole() {
-    if (this.isSimpleReporter()) {
-      aria.setRole(this.pathObject.svgPath, aria.Role.BUTTON);
-    } else if (this.workspace.isFlyout) {
+    if (this.workspace.isFlyout) {
       aria.setRole(this.pathObject.svgPath, aria.Role.TREEITEM);
     } else {
       aria.setState(
@@ -335,8 +346,6 @@ export class BlockSvg
     if (!svg.parentNode) {
       this.workspace.getCanvas().appendChild(svg);
     }
-    // Note: This must be done after initialization of the block's fields.
-    this.recomputeAriaLabel();
     this.initialized = true;
   }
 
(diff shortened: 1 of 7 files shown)


microbit-robert:
> * Simple reporter blocks/full block fields have had their readout improved. Previously, they were described as buttons, regardless of the nature of the field. Now, the block returns its single child field's focusable element as its own focusable element. This means that, from the perspective of the focus manager/keyboard navigation, the block is the focused element (as has always been the case) but the HTML element with focus is that of the field, along with its ARIA role and attributes, so from the screenreader's perspective the field is focused and described as if it were a normal field.

I presume this means that it is expected that custom fields need to add their own aria attributes as appropriate? If so, is there a default / fallback that can be used?

gonfunko:
> > * Simple reporter blocks/full block fields have had their readout improved. Previously, they were described as buttons, regardless of the nature of the field. Now, the block returns its single child field's focusable element as its own focusable element. This means that, from the perspective of the focus manager/keyboard navigation, the block is the focused element (as has always been the case) but the HTML element with focus is that of the field, along with its ARIA role and attributes, so from the screenreader's perspective the field is focused and described as if it were a normal field.
> 
> I presume this means that it is expected that custom fields need to add their own aria attributes as appropriate? If so, is there a default / fallback that can be used?

Yes, custom fields should set attributes as appropriate, but assuming they inherit from e.g. `FieldInput` or another concrete `Field` subclass, they will have a reasonable default without doing anything custom.

rachel-fenichel (approved):
Changes look good, and I'm glad label calculation is now being done lazily.

Please put a comment on this PR describing what tests you think should be added to validate these changes. (Or link to an issue with that information.)

gonfunko:
Things that would be nice to test, assuming we keep all this:

* Hidden inputs are excluded from block labels
* Hiding/unhiding an input updates the label
* Collapsed blocks are labelled as such
* Disabled blocks are labelled as such
* Toggling collapsed/disabled updates the label
* The count of child blocks for multiple-statement-input blocks is correct
* Input fields have `role="button"`
* Blocks return their field's focusable element for simple reporter full block field blocks
* `getFieldRowLabel()` returns an appropriate label for statement inputs both with and without preceding fields on the same row

PR #9426, opened 2025-10-20
fix: improve file path handling in theme generator script

mmvarma-pro:
## The basics

- [x] I have read https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change

### Resolves

Improve file path handling in theme generator script

### Proposed Changes

This PR enhances the `create_blockStyles.py` script to handle input files from any location by:

1. Using Path objects from the pathlib module to properly handle file paths
2. Generating output files in the same directory as the input file
3. Creating output filenames with a "new_" prefix while preserving the original stem name

### Reason for Changes

Previously, the script would only work correctly when run from the same directory as the input file. This change makes the script more robust by:

- Ensuring output files are always created in the same directory as the input file
- Using pathlib for cross-platform path handling
- Maintaining a consistent naming convention for output files

This improves developer experience when using the theme generator from different working directories.

### Test Coverage

The changes were tested manually by:
- Running the script with input files in the same directory
- Running the script with input files in different directories
- Verifying that output files are created in the expected locations
- Confirming that file content is correctly written

### Documentation

No documentation updates are needed as this is an internal improvement to the script's functionality.

### Additional Information

This change maintains backward compatibility with existing usage patterns while adding more flexibility for different file locations.


BenHenning (approved):
Thanks @mmvarma-pro! I don't have a lot of familiarity with the workings of the script, but I also suspect few people on the team do since it hasn't been changed in a long time. The change here makes sense to me and seems like a really solid improvement. I'm happy to accept this change.

Issue #8786, opened 2025-02-25
Allow accessing the underlying menu from ContextMenu


gonfunko:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Request

Currently, the ContextMenu module provides no way to get at the underlying Menu instance, which may be convenient. This would have been useful in https://github.com/google/blockly-keyboard-experimentation/pull/245 for example (which should be revisited when this is addressed). A getter should be added to ContextMenu to allow for this.

PR #9433, opened 2025-10-22
Tashee core 8786

tashee:
## The basics

### Resolves

Fixes issue #8786 

### Proposed Changes

Add getter to ContextMenu

### Reason for Changes

Currently, the ContextMenu module provides no way to get at the underlying Menu instance, which may be convenient.

### Test Coverage

Tests to be added separately.

### Documentation

N/A

### Additional Information

N/A

### DIFF
--- a/core/contextmenu.ts
+++ b/core/contextmenu.ts
@@ -66,7 +66,6 @@ export function setCurrentMeny(menu: Menu | null) {
   menu_ = menu;
 }
 
-
 /**
  * Gets Menu instance.
  *
--- a/core/contextmenu.ts
+++ b/core/contextmenu.ts
@@ -58,14 +58,6 @@ export function setCurrentBlock(block: Block | null) {
  */
 let menu_: Menu | null = null;
 
-export function getCurrentMenu(): Menu | null {
-  return menu_;
-}
-
-export function setCurrentMeny(menu: Menu | null) {
-  menu_ = menu;
-}
-
 /**
  * Gets Menu instance.
  *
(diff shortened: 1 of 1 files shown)


BenHenning:
Thanks @tashee! Apologies for the delayed response here. Some requests:
- The lint & label workflow is failing. I think this is due to your PR title being incorrectly formatted. Can you please fix that to match other PRs? E.g. 'fix: <Brief explanation fix>' is a great format to use.
- Could you perhaps go ahead and add tests for the new getter in this PR? It's generally better to add tests with the code change so that we can help verify that it works and not forget to add them later. :)
- Could you please sync with the latest `develop` changes as well? That should fix the welcome contributor workflow.

The keyboard navigation webdriver tests should be unrelated flakes from what I can tell.

core/contextmenu.ts
```diff
@@ -58,6 +58,15 @@ export function setCurrentBlock(block: Block | null) {
  */
 let menu_: Menu | null = null;
 
+/**
+ * Gets Menu instance.
+ *
+ * @returns menu instance.
```
BenHenning, line 64: One thing that would be really helpful to include is context on the menu's lifecycle. Specifically:
- When is it defined? (Relative to other functions called here)
- When is `null` returned? What does that mean/what can the caller infer from receiving a null value?
- Is it valid to save the returned value or could that cause problems (e.g. if the menu is later closed)?

These would be great to clarify in the documentation here for a more solid defined API contract for callers.

gonfunko:
Closing as the underlying issue has been addressed.

PR #9437, opened 2025-10-24
fix: Make up/previous navigation consistent with down/next.

gonfunko:
## The basics

### Resolves

Fixes https://github.com/google/blockly-keyboard-experimentation/issues/762

### Proposed Changes
This PR modifies the locations that are visited when navigating using the up arrow key to match those that are visited by the down arrow key, but in reverse. Specifically, when navigating backwards, we now:

* Visit the last (empty) next connection in a block stack within a statement input
* Do not visit the first (empty or filled) value input in a block with statement inputs, because this results in redundant screenreader output with the block itself. These inputs/blocks are also skipped when navigating with the down arrow for the same reason
* Do not focus the rightmost nested block in a statement block; instead, focus on the root statement block itself.

### DIFF
--- a/core/keyboard_nav/block_navigation_policy.ts
+++ b/core/keyboard_nav/block_navigation_policy.ts
@@ -137,9 +137,9 @@ function getBlockNavigationCandidates(
           .lastConnectionInStack(false)
           ?.getSourceBlock();
         if (lastStackBlock) {
-          // When navigating backward, the last block in a stack in a statement
-          // input is navigable.
-          candidates.push(lastStackBlock);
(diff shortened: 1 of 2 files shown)


maribethb (approved):

maribethb:
I think a similar set of tests is failing on my PR against this branch. Not sure what's up with them, although at least this PR is plausible that some of the tests need updating.

gonfunko:
Fixed the tests in https://github.com/google/blockly-keyboard-experimentation/pull/766, which should fully pass once this is merged.

PR #9439, opened 2025-10-24
fix: Don't filter out falsey input values when generating ARIA labels.

### LINKED ISSUE #9431: [Screen reader] Label filtering on falsey values
### Check for duplicates

### Description

The [label filtering in `recomputeAriaLabel` in `field_input.ts`](https://github.com/google/blockly/blob/add-screen-reader-support-experimental/core/field_input.ts#L194) accidentally filters out values such as `0`. The resulting screen reader output is simply "NUM" for number inputs where the value is `0` instead of "0, NUM".

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9431

### Proposed Changes
This PR updates the logic for generating ARIA labels for `FieldInput` to not omit falsey values like `false` or `0`, which can be valid input values that should be read out.

### DIFF
--- a/core/field_input.ts
+++ b/core/field_input.ts
@@ -191,7 +191,7 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
 
     const element = this.getFocusableElement();
     const label = [this.getValue(), this.getAriaName()]
-      .filter((item) => !!item)
+      .filter((item) => item !== null)
       .join(', ');
 
     aria.setState(element, aria.State.LABEL, label);


BenHenning (approved):
Nice catch @microbit-robert and thanks @gonfunko for the quick fix!

PR #9445, opened 2025-10-28
feat: Speech synthesis alternative experiment [DO NOT MERGE]

BenHenning:
## The basics

### Resolves

Fixes 

### Proposed Changes

Attempt to demonstrate using the web speech API as an alternative for reading out labels instead of using a screen reader.

The mode can be activated using 'ctrl+alt+s' or 'command+alt+s'.

Some problems/limitations:
- There's a bit of a delay before speech utterance begins, but then it's consistent.
- Roles are not read correctly, so additional role context is ignored.
- This does **not** seem to work alongside a reader. I made some effort to try and make it so that elements are ignored when the speech synthesis mode is enabled, but this reveals two issues:
  - (1) The speech synthesis API seems to be completely disabled when a reader is enabled (at least for ChromeVox).
  - (2) Removing all ARIA information and event setting an element to 'presentation' does NOT stop its readout if it has a negative tabindex and is focused. Instead, the browser seems to fall back to some default accessibility node representation which then leads to an unhelpful readout (like 'Graphics Symbol' for a block's path object).

### DIFF
--- a/core/focus_manager.ts
+++ b/core/focus_manager.ts
@@ -6,6 +6,7 @@
 
 import type {IFocusableNode} from './interfaces/i_focusable_node.js';
 import type {IFocusableTree} from './interfaces/i_focusable_tree.js';
+import * as aria from './utils/aria.js';
 import * as dom from './utils/dom.js';
 import {FocusableTreeTraverser} from './utils/focusable_tree_traverser.js';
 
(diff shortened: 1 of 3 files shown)


BenHenning:
Closing this since it won't actually be viable (per the PR description) but keeping the branch for now in case anyone wants to try and play around with this for whatever testing reason.

Issue #9307, opened 2025-08-15
Design long-term technical solution for screen reader support


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

This tracks the work for phase 3 of https://github.com/google/blockly-keyboard-experimentation/discussions/673, specifically: writing and finalizing a technical implementation plan for long-term screen reader support in Blockly.

### Request

Write a TDD that provides a long-term technical implementation plan for screen reader support by leveraging the experimentation and findings from phase 2 of https://github.com/google/blockly-keyboard-experimentation/discussions/673.

### Alternatives considered

The alternative would be not writing a TDD for screen reader support. That could be a viable option, but it's unlikely since it seems probable that screen reader support will require strong contractual enforcement in most Blockly components which has far-reaching implications for users of v13 and beyond. These considerations need careful evaluation by the team, and that's a bit more easily done with a document.

BenHenning:
Note that the final plan will need to include auditing and making accessible all fields provided in `blockly-samples`.

PR #9446, opened 2025-10-29
fix: Miscellaneous screen reader issues (experimental)

### LINKED ISSUE #9301: Introduce better field labeling and accessibility for screen readers

### LINKED ISSUE #9312: Improve aria-label for mutator workspaces

### LINKED ISSUE #9313: Discussion of the application of roles such as listbox, checkbox, and textbox

BenHenning:
## The basics

### Resolves

Fixes #9301
Fixes #9312
Fixes #9313
Fixes part of #9304

### Proposed Changes

This introduces a variety of specific changes to resolve several issues for screen reader work, including introducing fundamental support for field labeling.

Specifically:
- Field labels have been simplified to only use their custom defined ARIA name otherwise they are null (and thus should be ignored for readout purposes) which wraps up the remaining high-level work for #9301 (#9450 tracks more specific follow-up work to improve upon what's been established at this point). The PR also introduces an ARIA override for number inputs in math blocks so that the readout is correct for them.
- Bubble labeling is more explicit now which is useful for mutators (#9312), warnings, and comments. The general improvement for bubbles wraps up the remaining work for #9313 as well since the core issue was resolved in #9351. By default a bubble has no ARIA label.
- #9304 is partly being addressed here with the change to field images: they are no longer being added to the accessibility node tree unless they are actually navigable (that is, clickable). Part of #9304's goal is to remove extraneous nodes.
- Finally, a typo was fixed for 'replaceable blocks' since these were not reading out correctly. This was noticed in passing and isn't directly related to the other issues.

### Reason for Changes

This PR is largely being used as a basis for one particularly significant issue: #9301. Field labeling has undergone several iterations over the past few months and the team seems comfortable sticking with a "do as little as possible" approach when determining the label, thus justifying the need for expecting more specific customization (i.e. #9450). To this end it's important to be clear that getting fields to a good state is not actually "done" but the need to track it as a large incomplete thing has ended. Note that one important part of #9301 was updating field plugins to be accessible--this largely seems unnecessary as-is as it will be completely dependent on the needs of future user tests. The long-term plan will need to account for making all fields in `blockly-samples` accessible (per #9307).

Some of the terminology used here (e.g. for bubbles) will likely need to change after user testing, but it's important to establish that _something_ correct is communicated even if the terminology may require scaffolding and/or refinement.

It's important to note that while non-clickable field images are no longer in the node graph, their ARIA presence still exists as part of the fluent block labeling solution. That is, `FieldImage`'s alt text is used as part of constructing a fluent block label (sometimes to confusing effect--see #9452).

### Test Coverage

No tests needed since these are experimental changes and do not change existing test behaviors.

### Documentation

No documentation changes are needed for these experimental changes.

### Additional Information

None.


BenHenning (commented):
Finished self-review.

BenHenning:
This is a bit smaller of a change so I decided not to add a video. However if that is especially important for review I am happy to make one @maribethb.

rachel-fenichel (approved):
LGTM. Please check and confirm that the failing tests are due to the current test setup and not new issues.

BenHenning:
> LGTM. Please check and confirm that the failing tests are due to the current test setup and not new issues.

Thanks! Confirmed. The reported failures exactly match 22905ad5958c1190b5cc7c508ecc5c44e44a7d30 so it's setup/the existing test, not this PR.

Issue #9457, opened 2025-11-03
[Screen reader] The "Begin ..." label is missing information in some cases


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

For c-shaped blocks that don't have text on the statement input line and have value inputs in the previous field row, `getFieldRowLabel` is only returning part of the block text. Please see the example below, where "Begin pressed" is output, but it would be better if the output was "Being on button A pressed".

<img width="541" height="280" alt="Image" src="https://github.com/user-attachments/assets/b53b4baf-5fdc-4344-8982-d2c4a2ea91f3" />

Other blocks that have text on the statement input:

<img width="521" height="332" alt="Image" src="https://github.com/user-attachments/assets/2c86e4ca-585b-4f93-ae9e-f70d2ef633d1" />

or more simple c-shaped blocks without inputs:

<img width="528" height="259" alt="Image" src="https://github.com/user-attachments/assets/3738cf43-d07e-46c5-bc67-a90dae040525" />

are readout correctly.

### Reproduction steps

1. Go to https://screen-reader-preview.review-pxt.pages.dev/#editor - this is a version of MakeCode that has the latest screen reader work integrated
2. Add an "on button pressed" block and navigate to it's statement input connection
3. Observe the partial "Begin pressed" output

### Stack trace

```shell

```

microbit-robert:
We intend to open a PR for this.

BenHenning:
[...]

PR #9458, opened 2025-11-03
fix: create more comprehensive field row labels

microbit-robert:
## The basics

### Resolves

Fixes https://github.com/google/blockly/issues/9457

### Proposed Changes

This PR uses all preceding inputs to create a more comprehensive field row label.

There is a scenario where this fix won't do exactly the right thing. In the case where a block has multiple branches (e.g., an `if` block) and also doesn't have inline text for the statement input (e.g., there is no "do"), then the entire block to the current input index is going to be output, which might be something like "Begin if true then else if false...". Thoughts on this scenario are welcome. In such situations, `getFieldRowLabel` could be overridden by the developer, but I wonder if a better default can be provided.

### DIFF
--- a/core/inputs/input.ts
+++ b/core/inputs/input.ts
@@ -324,9 +324,17 @@ export class Input {
     if (!fieldRowLabel) {
       const inputs = this.getSourceBlock().inputList;
       const index = inputs.indexOf(this);
-      if (index > 0) {
-        return inputs[index - 1].getFieldRowLabel();
-      }
+      const precedingInputs = inputs.slice(0, index);
+      return precedingInputs
+        .flatMap((input) => {
+          const fields = input.fieldRow.map((field) => {
+            if (!field.isVisible()) return undefined;
+            return [field.getText() ?? field.getValue()];
+          });
+          return fields;
+        })
+        .filter(Boolean)
+        .join(' ');
     }
     return fieldRowLabel;
   }


core/inputs/input.ts
```diff
@@ -324,9 +324,17 @@ export class Input {
     if (!fieldRowLabel) {
       const inputs = this.getSourceBlock().inputList;
       const index = inputs.indexOf(this);
-      if (index > 0) {
```
BenHenning, line 327: This seems necessary to keep (at least, `slice` will probably misbehave without this check). That being said, I'm not sure it's even possible for this scenario to logically happen in practice, but it's definitely possible just from an API perspective.

core/inputs/input.ts
```diff
@@ -324,9 +324,17 @@ export class Input {
...
+      const precedingInputs = inputs.slice(0, index);
+      return precedingInputs
+        .flatMap((input) => {
+          const fields = input.fieldRow.map((field) => {
+            if (!field.isVisible()) return undefined;
+            return [field.getText() ?? field.getValue()];
+          });
+          return fields;
+        })
+        .filter(Boolean)
```
BenHenning, line 336: I've actually never seen this syntax 'hack'. Is it effectively the same as `.filter((value) => !!value)`? Which, I suppose, is simply filtering out empty and null/undefined entries.

BenHenning (commented):
Thanks @microbit-robert! I have a couple of comments but I also want to follow up on the issue itself.

core/inputs/input.ts
microbit-robert: `[1, 2, 3].slice(0, 0)` returns `[]`, so this should do no harm, but I'm happy to keep the check in.

core/inputs/input.ts
microbit-robert: Yes, this is exactly what it does in terms of API and in practice.

BenHenning:
@microbit-robert can this be closed now since #9457 is resolved or are there still tasks to do here? (just wanted to double check)

microbit-robert:
@BenHenning Yes, I think this can be closed. This might resurface in a similar or different form when it comes to considering how move mode works with if / else blocks and how moving into different branches is announced, but that can be a new issue if required.

PR #9470, opened 2025-11-14
feat: field label improvements (experimental)

### LINKED ISSUE #9450: All fields should have ARIA label overrides
### Check for duplicates

### Problem

Without context there's not really any indication as to what the value means for a particular field when using a screen reader.

### Request

In order to ensure that block and field readouts provide sufficient context, all built-in shadow block and field input cases should have contextual overrides (like the math arithmetic block).

### Alternatives considered

_No response_

### Additional context

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9450


BenHenning:
This needs a bunch of explanation and work yet. It largely adds good naming for fields and drop-downs for all but the dynamic cases (since those are trickier and may need some additional work to make it possible).

It also splits field labels into type and values and largely keeps the type out of the block readout unless using new shortcuts.

Two new shortcuts:
- ctrl/command + shift + I: prints out a verbose readout of the current block
- ctrl/command + shift + P: prints out a verbose readout of the current block's parent

These are meant to provide initial possibilities as part of addressing https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764.

BenHenning:
From feedback: ensure fields always read out their type + value (verbose) when focused.

core/field.ts
```diff
@@ -271,8 +271,51 @@ export abstract class Field<T = any>
...
+  getAriaTypeName(): string | null {
+    return this.config?.ariaTypeName ?? null;
+  }
+
+  /**
+   * Gets a an ARIA-friendly label representation of this field's value.
+   *
+   * @returns An ARIA representation of the field's value.
+   */
+  abstract getAriaValue(): string;
```
microbit-robert: Could this not be abstract and have a default implementation that returned an empty string (or null)? I found myself going through and returning empty strings for quite a few custom fields in order to get up and running with this change, and having a default implementation would reduce the overhead.

BenHenning: I was on the fence of going with a default implementation but this gives strong incentive to. Within core Blockly we more or less need to explicitly decide what to provide here in all cases so I didn't have enough context to decide on a default value.

Could you maybe provide some context on why so many of your custom fields have an empty string? I'd expect that to be atypical in general since I expect that basically all fields have a non-default value (except maybe text input when nothing has yet been inputted).

BenHenning:
The shortcut bits have been moved out of this branch and into #9481 (which is now merged into the experimental branch). Bringing this up-to-date to focus on specifically the label improvements.

BenHenning:
Hmm I inadvertently undid my changes. :)

Edit: They should now be back. I still need to go through and double check that all fields have been accounted for, and add support for blocks defined through functions rather than JSON.

PR #9478, opened 2025-11-20
fix: Remove extra image ARIA properties for drop-down fields (experimental)

### LINKED ISSUE #9461: [Screen reader] Remove aria for field_dropdown image element
### Check for duplicates

### Description

In https://github.com/RaspberryPiFoundation/blockly/pull/9384, `role` and `aria-label` were added to the image element (if applicable) for `field_dropdown`. This causes the dropdown to be read as 'label, image / graphic' rather than providing list-box information and using the aria-label of the parent `g` element. I think [these attributes](https://github.com/RaspberryPiFoundation/blockly/blob/add-screen-reader-support-experimental/core/field_dropdown.ts#L628-L629) should probably be removed so the more informative aria information is used.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

MakeCode examples with VoiceOver below.

If the image has a role and aria-label:
Image

If the image does not have a role and aria-label:
Image

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9461

### Proposed Changes

Remove image-specific ARIA properties (role and label) for the current selected item in `FieldDropdown`.

### Reason for Changes

This intentionally changes a behavior introduced in #9384 because the alternative ARIA behavior works better. As the images in #9461 show the actual focus outline when the current drop-down value has ARIA properties goes to that element (separate from active focus in this case) since the screen reader will read out those properties. However, this makes the reader lose context on the combo box itself.

This basically goes away from leveraging active descendant (though that's still necessary for correct combo box implementation) and just forces the label in a way that keeps the combo box context. Text-based combo boxes already do this, so this change simply brings image-based items in alignment with text combo box behaviors.

### Test Coverage

No new automated tests are needed for this experimental change. Manual testing was done via core Blockly's advanced playground (using the 'test blocks' toolbox).

### Documentation

No documentation changes are needed.

### Additional Information

None.

### DIFF
--- a/core/field_dropdown.ts
+++ b/core/field_dropdown.ts
@@ -624,9 +624,6 @@ export class FieldDropdown extends Field<string> {
       const element = this.getFocusableElement();
       aria.setState(element, aria.State.ACTIVEDESCENDANT, imageElement.id);
     }
-
-    aria.setRole(imageElement, aria.Role.IMAGE);
-    aria.setState(imageElement, aria.State.LABEL, imageJson.alt);
   }
 
   /** Renders the selected option, which must be text. */


BenHenning:
PTAL @rachel-fenichel.

rachel-fenichel (approved):
LGTM.

I appreciate that you've got all of the infrastructure set up and now these changes are just a line or two.

BenHenning (commented):
Self-reviewed spot check to make sure only the changes I want are in the PR. :)

PR #9480, opened 2025-11-20
fix: Remove clickable image labels from block summary (experimental)

### LINKED ISSUE #9452: [Screen reader] consider excluding instances of FieldImage from block aria labels
### Check for duplicates

### Description

While integrating the latest screen reader work into MakeCode, we found that the fluent block readout included the aria labels from instances of FieldImage in the block. In the MakeCode case, these are buttons to add / remove `else` and `else if` clauses. While we want these aria labels when navigating to the mutator buttons themselves, it was confusing and overloading to have these labels included in the overall block output. 

Examples below (these use a tweaked block readout, but the impact of the FieldImage labels is evident regardless).

With FieldImage labels (the more clauses you have, the worse it gets):
Image

Without FieldImage labels:
Image

We don't know how other consumers of Blockly use FieldImage, so excluding by default might not be appropriate.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9452

### Proposed Changes

This change filters out `FieldImage` labels from `BlockSvg`'s ARIA summary when they are clickable.

### Reason for Changes

The discussions in #9452 provide more context but largely this change seems to be a net positive for users:
- Clickable images are less likely to provide block-identifying context that makes them worthy of being in the block's label (unlike `FieldLabel` or non-interactive `FieldImage`s), whereas including them can create additional confusion due to extra, possibly unnecessary, context being added to the block label that could be determined based on navigation discovery.
- It creates consistency with clickable icons (such as for opening mutator bubbles) which are also not included as part of the block label.

Behavior without change:

[Screen recording 2025-11-19 5.36.06 PM.webm](https://github.com/user-attachments/assets/716f7f61-6297-4770-9717-24f1f04c072c)

Behavior with change (notice the shorter block readout for the 'Image click handler' block):

[Screen recording 2025-11-19 5.34.32 PM.webm](https://github.com/user-attachments/assets/d645093b-2819-473a-bec8-c6214e925729)

### Test Coverage

No additional automated testing is needed as part of this change. Manual testing was conducted using core Blockly's advanced playground with the 'test blocks' toolbox (to test with the 'Image click handler' block).

### Documentation

### Additional Information

None.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -33,6 +33,7 @@ import {BlockDragStrategy} from './dragging/block_drag_strategy.js';
 import type {BlockMove} from './events/events_block_move.js';
 import {EventType} from './events/type.js';
 import * as eventUtils from './events/utils.js';
+import {FieldImage} from './field_image.js';
 import {FieldLabel} from './field_label.js';
 import {getFocusManager} from './focus_manager.js';
 import {IconType} from './icons/icon_types.js';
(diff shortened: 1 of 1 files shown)


BenHenning (commented):
Self-review spot check.

BenHenning:
PTAL @rachel-fenichel.

rachel-fenichel (approved):

Issue #9460, opened 2025-11-06
[Screen reader] Review of aria-live region announcements


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

Screen readers behave differently when it comes to announcing content in aria-live regions. This is a review of the toast and 'move mode' announcements as of [commit 74a36d7](https://github.com/google/blockly/commit/74a36d77692de4e282fdffc4dfb4e7112f740edc), so includes the change to make the move mode live region assertive. We can test VoiceOver, NVDA, JAWS and ChromeVox, so these are reviewed below, and we are happy to test changes going forward. It will be important to check the live region behaviour against any other screen reader that is being targeted.

### Summary

As things stand:
- ChromeVox performs perfectly with the current implementation and the best out of those tested
- JAWS and NVDA perform well announcing move mode but do not announce toast messages or repeated message
- VoiceOver performs the worst having some issues dropping move mode announcements while block labels are being announced (this is a regression from changing the live region to assertive)

### NVDA

- When inserting a block from the toolbox, move mode is correctly announced
- Move mode generally works, however:
    - If the text in the aria-live region is the same as the previous text, the message is not announced (this is a known issue with screen readers)
- Move announcements correctly interrupt themselves while moving a block around
- If a move is triggered while the block aria-label is being readout, move mode announcement correctly interrupts
- On accepting a move, the move announcement is truncated
- Toast messages are not announced at all

### JAWS

Behaviour matches that of NVDA except for:
- When a block on the workspace is put into move mode, the block label is announced before move mode as opposed to NVDA which just announces the move

### ChromeVox

- When inserting a block from the toolbox, move mode is correctly announced. If a toast message is shown at this point, the toast is announced before move mode. It might be preferable to have the move mode message first.
- If the text in the aria-live region is the same as the previous text, the message is correctly announced 
- Move announcements correctly interrupt themselves while moving a block around. Toast messages and move mode messages also correctly interrupt each other.
- If a move is triggered while the block aria-label is being readout, move mode announcement correctly interrupts the block readout
- On accepting a move, the move announcement continues to be read out in full
- Toast messages are correctly announced, however, the readout includes the toast close button which seems irrelevant. The role of “Status” is also announced, which I don’t think adds anything. 

### VoiceOver

- When inserting a block from the toolbox, move mode is not announced. Note that it is announced after the block readout when the live region is set to “polite”.
- Move mode generally works, however:
[...]

PR #9481, opened 2025-11-26
feat: Add verbosity shortcuts (experimental)

BenHenning:
## The basics

### Resolves

Fixes part of https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764
Fixes part of #9450 (infrastructure needs)

### Proposed Changes

Introduces support for two new "where am I?" shortcuts for helping to provide location context for users:
- `I`: re-reads the current selected block with full verbosity (i.e. also includes the block's field types with their values in the readout).
- `shift+I`: reads the current selected block's parent with full verbosity.

Note that this includes some functional changes to `Field` to allow for more powerful customization of a field's ARIA representation (by splitting up value and type), though a field's value defaults potentially to null which will be ignored in the final ARIA computed label. This seems necessary per the discussion here: https://github.com/RaspberryPiFoundation/blockly/pull/9470/files#r2541508565 but more consideration may be needed here as part of #9307.

Some limitations in the new shortcuts:
- They will not read out anything if a block is not selected (e.g. for fields and icons).
- They read out input blocks when the input block is selected.
- They cannot read out anything while in move mode (due to the behavior here in the plugin which automatically cancels moves if an unknown shortcut is pressed: https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/blob/a36f3662b05c2ddcd18bde8745777fff8dc3df31/src/actions/mover.ts#L166-L191).
- The readout is limited by the problems of dynamic ARIA announcements (per #9460).

### Reason for Changes

https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764 provides context on the specific needs addressed here.

### Test Coverage

Self tested. No new automated tests needed for experimental work.

### Documentation

No new documentation needed for experimental work.

### Additional Information

This was spun out of #9470 with the intent of getting shortcuts initially working checked in even if the entirety of the experience is incomplete.


BenHenning (commented):
Quick self-review spot check.

BenHenning (commented):
Post-conflict merge spot check.

BenHenning:
PTAL @maribethb. I'm not considering this a full completion of https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764 for the caveats listed in the PR description, but it seems perhaps reasonable to check this in as-is and iterate to address the remaining needs of the issue (which may require changes both here & in the plugin).

Note I'll also revisit #9470 after this is merged since that's lower priority (and duplicates some of the code here so it's easier to pipeline it).

At least 2 of the 4 WebdriverIO failures on Mac seem to be existing. The other 2 look a bit different, but I don't think this is actually a new failure since the Linux tests aren't failing (and I'd expect them to in such a case).

core/shortcut_items.ts
```diff
@@ -386,6 +389,71 @@ export function registerRedo() {
   ShortcutRegistry.registry.register(redoShortcut);
 }
 
+/**
+ * Registeres a keyboard shortcut for re-reading the current selected block's
```
maribethb: nit: Registers (typo)

core/shortcut_items.ts
```diff
@@ -386,6 +389,71 @@ export function registerRedo() {
...
+export function registerReadFullBlockSummary() {
+  const i = ShortcutRegistry.registry.createSerializedKey(KeyCodes.I, null);
+  const readFullBlockSummaryShortcut: KeyboardShortcut = {
+    name: names.READ_FULL_BLOCK_SUMMARY,
+    preconditionFn(workspace) {
+      return (
+        !workspace.isDragging() &&
+        !getFocusManager().ephemeralFocusTaken() &&
+        !!getFocusManager().getFocusedNode() &&
+        getFocusManager().getFocusedNode() instanceof BlockSvg
```
maribethb, line 406: This shortcut should work for fields, connections, and icons too. The idea is that you can get information about the block without having to change your focus. This can help you figure out what to type into a field, for example.

I know you have a lot in flight right now though so it may make more sense to come back to this as a follow up (or pass to Aaron) in order to at least get the base version submitted.

maribethb (approved):
sorry this was a stray comment I didn't mean to leave

core/shortcut_items.ts
BenHenning: Yep I agree, and that's part of why this isn't fully resolving https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764. I'll reference this comment there so that this context isn't lost, and I'm happy for @gonfunko to take over that issue if he has bandwidth to continue chipping away at the other parts of it.

core/shortcut_items.ts
BenHenning: Done.

BenHenning (commented):
Spot checked changes.

BenHenning:
Failures are unrelated. Since there's more open work to do I won't be addressing your second comment @maribethb but I did note it in the main tracking issue for the shortcut so that it gets addressed. I'm hoping merging this earlier will allow it to potentially be tested in interim (or, at least, available for testing).

PR #9483, opened 2025-11-26
feat: Add labels for `Toolbox` and `Flyout` content containers (experimental)

### LINKED ISSUE #9464: [Screen reader] Trees should have aria-labels or aria-labelledby attributes
### Check for duplicates

### Description

The toolbox tree and flyout tree are missing these attributes which I think are required to follow best practice. For most screen readers you can get enough context from the region change, but for some reason, JAWS does not announce the region change and the trees themselves don’t have labels. I have tried changing the JAWS settings (general verbosity and web page specific verbosity) to no avail. This change makes the NVDA readout more verbose, but I think this is necessary to support anyone using JAWS. 

It’s particularly bad in MakeCode without the tree labels as we move from the Basic category to the Basic flyout where the first item is the Basic `FlyoutLabel`, which is very confusing.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

BenHenning:
## The basics

### Resolves

Fixes #9464

### Proposed Changes

Adds ARIA labels for `Toolbox` and `Flyout` ('s `Workspace`) containers to complement their existing region labels.

### Reason for Changes

In some cases certain screen readers (like NVDA) don't read out region labels. Previously, only the region labels provided context for currently being within the toolbox or flyout, so adding additional labels on the ARIA `tree` containers themselves ensures that these contexts are always read if the tree's context is read.

This adds redundancy with the region output if both are read, but ChromeVox at least adds some delay between finishing the tree's context and beginning the region's context and this seems to mesh well together. The fully read context now feels like: "here's an item, it's the toolbox tree, and that tree is within the toolbox region." Even though both regions only contain a single tree it doesn't feel dimorphic.

### Test Coverage

No new automated tests are needed for this experimental change.

This has been manually tested with ChromeVox to demonstrate the dual tree + region output for both toolbox and flyout:

[Screen recording 2025-11-26 2.52.22 PM.webm](https://github.com/user-attachments/assets/aa43ae81-9da6-4c79-b0fc-120146892aae)

### Documentation

### Additional Information

This was only tested on ChromeVox.

### DIFF
--- a/core/toolbox/toolbox.ts
+++ b/core/toolbox/toolbox.ts
@@ -226,6 +226,11 @@ export class Toolbox
       contentsContainer.style.flexDirection = 'row';
     }
     aria.setRole(contentsContainer, aria.Role.TREE);
+    aria.setState(
+      contentsContainer,
+      aria.State.LABEL,
+      Msg['TOOLBOX_ARIA_LABEL'],
+    );
     return contentsContainer;
   }
 
--- a/core/workspace_svg.ts
+++ b/core/workspace_svg.ts
@@ -801,6 +801,7 @@ export class WorkspaceSvg
     if (this.isFlyout) {
       // Use the block canvas as the primary tree parent for flyout blocks.
       aria.setRole(this.svgBlockCanvas_, aria.Role.TREE);
+      aria.setState(this.svgBlockCanvas_, aria.State.LABEL, ariaLabel);
     } else {
       browserEvents.conditionalBind(
         this.svgGroup_,


BenHenning (commented):
Spot check on code.

Mac CI failures _seem_ to be existing but there are more than I'm used to seeing. I don't see why this change would cause any new issues, and the tests are still passing on Linux so this is likely not affecting those tests.

BenHenning:
PTAL @maribethb.

/cc @microbit-robert in case you wanted to re-check this on the micro:bit side.

microbit-robert:
@BenHenning LGTM

maribethb (approved):

BenHenning:
Thanks @maribethb! Going ahead and merging this since the CI failures seem to be unrelated to this change.

Issue #9456, opened 2025-11-03
[Screen reader] Missing aria details for shadow blocks in some cases


microbit-robert:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

There is currently a [check for whether a block is a simple reporter with a full block field](https://github.com/google/blockly/blob/add-screen-reader-support-experimental/core/block_svg.ts#L235-L238), however, this code only checks the first field in the input. There are instances in MakeCode where it is not the first field that is the full block field. Please see the example below where the first field is just the text "melody" and the second field is the full block field.

<img width="405" height="380" alt="Image" src="https://github.com/user-attachments/assets/6846ebe5-db26-45ff-8622-164251864778" />

I think that at least for shadow blocks, the check needs to extend to all of the fields. This same change would also have to be applied to [`getFocusableElement` where the same check is made](https://github.com/google/blockly/blob/add-screen-reader-support-experimental/core/block_svg.ts#L1965-L1971).

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

BenHenning:
I'm assuming that in this case 'melody' is a field label. We could possibly extend the definition of simple reporter (at least for the sake of screen reader & keyboard navigation, not sure about generally, but maybe that makes sense) to be a block that contains exactly one editable field rather than having exactly one field that's editable. I think that will address the problem here, but some additional considerations need to be made if it's affecting more than keyboard nav.

BenHenning:
This is mostly fixed from #9484 but there's still a change needed in the plugin.

BenHenning:
Fully addressed in https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/pull/773.

PR #9484, opened 2025-11-27
feat: Expand single field block labeling (experimental)

BenHenning:
## The basics

- [ ] ~I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)~ (I can't really due to the nature of this change--I have to rely on tests and external testing)

### Resolves

Fixes part of #9456 (remainder will be in a change in the keyboard navigation plugin)

### Proposed Changes

Introduce and use new `Block` function for retrieving a configurably constrained singleton field for a given block. The constraints allow for some level of configuring (such as whether to isolate to only full or editable blocks). The existing simple reporter function has been retrofitted to use this new function, instead.

### Reason for Changes

This expanded support fixes the underlying use case.

Separately, this change reveals two noteworthy details:
1. There's inconsistency in the codebase as to when the singleton field needs to be editable, a full-block field, both, and neither. It would be ideal to make this consistent. Interestingly, the documentation for `isSimpleReporter` seems to have been wrong since it wasn't actually fulfilling its contract of returning an editable field (this has been retained for callsites except where the check was already happening).
2. There's a possible recursion case now possible between `getSingletonFullBlockField` and `isFullBlockField` due to `FieldInput`'s `isFullBlockField` depending on `isSimpleReporter`. Ideally this would be changed in the future to avoid that potential recursion risk (possibly as part of #9307).

### Test Coverage

No new automated tests are needed for this experimental work. Manual testing mainly comprised of cursory navigation and readout checks for single-field blocks to make sure nothing breaks. More thorough testing is difficult in core since the specific situation of multiple fields don't have a corresponding block to use in the playground to verify.

Automated tests are also being heavily relied on for correctness since all of the nuance behind the simple reporter cases would require a deeper testing pass.

### Documentation

No new documentation needed for this experimental work.

### Additional Information

None.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -966,8 +966,13 @@ export class Block {
    * @param mustBeEditable Whether the evaluated field must be editable.
    * @internal
    */
-  isSimpleReporter(mustBeFullBlock: boolean = false, mustBeEditable: boolean = false): boolean {
-    return this.getSingletonFullBlockField(mustBeFullBlock, mustBeEditable) !== null;
+  isSimpleReporter(
+    mustBeFullBlock: boolean = false,
+    mustBeEditable: boolean = false,
+  ): boolean {
+    return (
+      this.getSingletonFullBlockField(mustBeFullBlock, mustBeEditable) !== null
+    );
   }
 
   /**
@@ -981,7 +986,10 @@ export class Block {
    * @returns The only full-block, maybe editable field of this block, or null.
    * @internal
    */
-  getSingletonFullBlockField(mustBeFullBlock: boolean, mustBeEditable: boolean): Field<any> | null {
+  getSingletonFullBlockField(
+    mustBeFullBlock: boolean,
+    mustBeEditable: boolean,
+  ): Field<any> | null {
     if (!this.outputConnection) return null;
     for (const input of this.inputList) if (input.connection) return null;
     const matchingFields = Array.from(this.getFields()).filter((field) => {
--- a/core/block.ts
+++ b/core/block.ts
@@ -963,6 +963,7 @@ export class Block {
 
   /**
    * @returns True if this block is a value block with a full block field.
+   * @param mustBeFullBlock Whether the evaluated field must be 'full-block'.
    * @param mustBeEditable Whether the evaluated field must be editable.
    * @internal
    */
@@ -976,7 +977,7 @@ export class Block {
   }
 
   /**
-   * Determies and returns the only field of this block, or null if there isn't
+   * Determines and returns the only field of this block, or null if there isn't
    * one and this block can't be considered a simple reporter. Null will also be
    * returned if the singleton block doesn't match additional criteria, if set,
    * such as being full-block or editable.
--- a/core/block.ts
+++ b/core/block.ts
@@ -962,16 +962,34 @@ export class Block {
   }
 
   /**
-   * @returns True if this block is a value block with a single editable field.
+   * @returns True if this block is a value block with a full block field.
+   * @param mustBeEditable Whether the evaluated field must be editable.
    * @internal
    */
-  isSimpleReporter(): boolean {
-    if (!this.outputConnection) return false;
+  isSimpleReporter(mustBeFullBlock: boolean = false, mustBeEditable: boolean = false): boolean {
+    return this.getSingletonFullBlockField(mustBeFullBlock, mustBeEditable) !== null;
+  }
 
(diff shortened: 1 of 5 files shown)


BenHenning:
There seems to be one legitimate deletion test failure caused by this. Exact cause is unclear and will require investigation.

BenHenning:
@microbit-robert if you get a chance could you check to see whether this fix actually fixes micro:bit's use case? It may be well worth me looking to double check there aren't any playground blocks that we can use for this (or perhaps I could add one), but I do at least want to ensure that the fix actually does what it's supposed to per the original issue.

microbit-robert:
Thanks @BenHenning. This fixes the micro:bit use case when combined with a change in the keyboard-navigation plugin.

I think that [`block.isSimpleReporter` in `tryShowFullBlockFieldEditor`](https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/blob/main/src/actions/enter.ts#L352-L353) needs `true, true` passing in, so that "Enter" actually opens the dropdown.

BenHenning:
> Thanks @BenHenning. This fixes the micro:bit use case when combined with a change in the keyboard-navigation plugin.
> 
> I think that [`block.isSimpleReporter` in `tryShowFullBlockFieldEditor`](https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/blob/main/src/actions/enter.ts#L352-L353) needs `true, true` passing in, so that "Enter" actually opens the dropdown.

Nice catch. I'll need to think about how to coordinate the changes here, but that does make sense. Thanks for confirming as well!

BenHenning:
> There seems to be one legitimate deletion test failure caused by this. Exact cause is unclear and will require investigation.

This corresponded to the following failure:

```
  1) Deleting Blocks
       Cutting stranded block selects top block:
     AssertionError: expected null to equal 'p5_setup_1'
      at Context.<anonymous> (file:///home/runner/work/blockly/blockly/blockly-keyboard-experimentation/test/webdriverio/test/dist/delete_test.js:165:25)
      at Generator.next (<anonymous>)
      at fulfilled (file:///home/runner/work/blockly/blockly/blockly-keyboard-experimentation/test/webdriverio/test/dist/delete_test.js:9:58)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Which hasn't recurred. We might have a flaky test on Linux, too. Will need to keep an eye out for this.

BenHenning:
Given that the navigation plugin change can be done separately from this, it's reasonable for this to go into review now and for that to be a follow-up change once this is merged.

PTAL @maribethb or @gonfunko.

core/block.ts
```diff
@@ -962,16 +962,42 @@ export class Block {
   }
 
   /**
-   * @returns True if this block is a value block with a single editable field.
+   * @returns True if this block is a value block with a full block field.
+   * @param mustBeEditable Whether the evaluated field must be editable.
```
gonfunko, line 967: Nit: Add docs for the `mustBeFullBlock` arg.

core/block.ts
```diff
@@ -962,16 +962,42 @@ export class Block {
...
+      this.getSingletonFullBlockField(mustBeFullBlock, mustBeEditable) !== null
+    );
+  }
 
-    for (const input of this.inputList) {
-      if (input.connection || input.fieldRow.length > 1) return false;
-    }
-    return true;
+  /**
+   * Determies and returns the only field of this block, or null if there isn't
```
gonfunko: Nit: Determines

gonfunko (approved):

core/block.ts
BenHenning: Done.

core/block.ts
BenHenning: Done.

BenHenning:
Thanks @gonfunko! Going ahead and merging this once CI is done (probably ignoring the failing OS X tests since they should be unrelated).

BenHenning:
Huh. Hit the same deletion failure in the Ubuntu tests. Gonna try running this locally a few times to see if I can get some consistency on a local failure.

BenHenning:
There definitely seems to be a flake here which isn't great to see, particularly since it's the first time I've noticed it and it's happened twice in this PR. Will try re-running on CI. I had 10 local passes of the test so it doesn't seem very easy to repro, or it's environmental.

BenHenning:
Passed this time so going ahead and merging.

PR #9501, opened 2025-12-02
feat: Identify root stack blocks in ARIA label

gonfunko:
## The basics

### Resolves

Fixes #[9434](https://github.com/RaspberryPiFoundation/blockly/issues/9434)

### Proposed Changes
This PR adds "Begin stack" to the ARIA label of root blocks in the workspace, so that navigating to a new root block identifies it as the start of a stack. Although the bug also requested this for the end of stacks, the technically-last block in a stack often doesn't pass the vibes test in terms of user expectations (e.g. a deeply nested input block or statement), so I've omitted that for now. Navigating past it will of course call out "Begin stack", so it would be a bit redundant anyway. We can revisit if testing shows it would be valuable however.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -301,6 +301,10 @@ export class BlockSvg
       prefix = `${parentInput.getFieldRowLabel()} `;
     }
 
+    if (this.getRootBlock() === this) {
+      prefix = 'Begin stack, ' + prefix;
+    }
+
     let additionalInfo = blockTypeText;
     if (inputSummary && !nestedStatementBlockCount) {
       additionalInfo = `${additionalInfo} with ${inputSummary}`;


core/block_svg.ts
```diff
@@ -301,6 +301,10 @@ export class BlockSvg
       prefix = `${parentInput.getFieldRowLabel()} `;
     }
 
+    if (this.getRootBlock() === this) {
+      prefix = 'Begin stack, ' + prefix;
```
maribethb, line 305: This should be in a Msg I think

We haven't been super consistent with using messages but I think it makes sense to, I know we are doing it in some places even in this experimental branch

maribethb (approved):

core/block_svg.ts
maribethb: Nevermind, after viewing the rest of your PRs, you're doing stuff that's going to be semi-complicated with translations, so I think it's fine to plan to do a more robust translation pass before we merge this branch to main.

gonfunko: Yeah, all the labels are hardcoded for the moment so I'm definitely anticipating we'll need to clean that up once we lock in what we actually want them to be.

PR #9502, opened 2025-12-02
fix: Don't include count of inputs in block ARIA label

### LINKED ISSUE #9491: Don't include the number of inputs in block ARIA labels
Currently, the number of inputs is included in blocks' ARIA labels. Instead, the label should simply note that the block "has inputs" if it does, and not include a count.

gonfunko:
## The basics

### Resolves

Fixes #9491

### Proposed Changes
This PR removes the count of input blocks from blocks' ARIA labels, and instead notes simply "has inputs" if that is the case.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -244,7 +244,12 @@ export class BlockSvg
 
   private computeAriaLabel(): string {
     const {blockSummary, inputCount} = buildBlockSummary(this);
-    const inputSummary = inputCount >= 1 ? 'has inputs' : '';
+    let inputSummary = '';
+    if (inputCount > 1) {
+      inputSummary = 'has inputs';
+    } else if (inputCount === 1) {
+      inputSummary = 'has input';
+    }
 
     let currentBlock: BlockSvg | null = null;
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -244,9 +244,12 @@ export class BlockSvg
...
   private computeAriaLabel(): string {
     const {blockSummary, inputCount} = buildBlockSummary(this);
-    const inputSummary = inputCount
-      ? ` ${inputCount} ${inputCount > 1 ? 'inputs' : 'input'}`
-      : '';
+    let inputSummary = '';
+    if (inputCount > 1) {
+      inputSummary = 'has inputs';
+    } else if (inputCount === 1) {
+      inputSummary = 'has input';
```
maribethb, line 251: we are going to be sad later because we don't support plural translations :/

(no action needed, this is just me wishing we had more robust i18n systems)

maribethb (approved):

core/block_svg.ts
gonfunko: I am already sad :( But yeah, that would be good to address at some point.

PR #9505, opened 2025-12-03
feat: Add comma-separation to ARIA labels

grega:
## The basics

### Resolves

Fixes: https://github.com/RaspberryPiFoundation/blockly/issues/9494

### Proposed Changes

Adds comma separation between "labels" and "inputs", results in aria-labels being transformed such as:

```
count with i from 1 to 10 by 1 do, C-shaped block with  4 inputs
->
count with, i, from, 1, to, 10, by, 1, do, C-shaped block with  4 inputs
```

And:

```
create list with item i repeated 5 times, input block with  2 inputs
->
create list with item, i, repeated, 5, times, input block with  2 inputs
```

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -243,7 +243,7 @@ export class BlockSvg
   }
 
   private computeAriaLabel(): string {
-    const {blockSummary, inputCount} = buildBlockSummary(this);
+    const {commaSeparatedSummary, inputCount} = buildBlockSummary(this);
     const inputSummary = inputCount
       ? ` ${inputCount} ${inputCount > 1 ? 'inputs' : 'input'}`
       : '';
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -2041,57 +2041,114 @@ export class BlockSvg
...
+    }
+  }
+  flushLabels();
+
+  // comma separate label runs and inputs
+  const commaSeparatedSummary =
+    spokenParts.length <= 1
+      ? (spokenParts[0] ?? '')
+      : spokenParts[0] + ', ' + spokenParts.slice(1).join(', ');
+
```
gonfunko: Perhaps I'm missing something but it seems like this could just be `const commaSeparatedSummary = spokenParts.join(', ');`?

gonfunko:
This looks great, thanks!

core/block_svg.ts
grega: Yep you're right, I was initially thinking the first element in the array might need to be treated differently (especially if it's the only element), but `.join(', ')` already handles that in exactly the same way with no need for the rest. I'll push a change up...

grega: Would you prefer me to rebase so we have a single `feat: ...` commit?

gonfunko (approved):

PR #9507, opened 2025-12-03
feat: Use custom ARIA roledescriptions for different block types

### LINKED ISSUE #9489: Use `aria-roledescription` to specify block types
Currently, block types are included as part of the ARIA label, and the `aria-roledescription` is simply "block". Instead, we want to specify the block type as one of the following in `aria-roledescription`:

"container block" (C-shaped blocks)
"statement block"
"value block"

It should also be possible for block authors to override/customize these names, to allow e.g. specifying the data type of value blocks or features like hat blocks that are unique to the application.

gonfunko:
## The basics

### Resolves

Fixes #9489

### Proposed Changes
This PR updates the `aria-roledescription` to specify the kind of block, rather than simply "block", with the kind specified in the label. This results in less redundant output and uses hopefully clearer names for the different block types. I also refactored the ARIA label generation code since it had become an unmaintainable mess of string manipulation.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -274,6 +274,9 @@ export class BlockSvg
     if (this.isCollapsed()) {
       labelComponents.push('collapsed');
     }
+    if (this.isShadow()) {
+      labelComponents.push('replaceable');
+    }
 
     if (inputCount > 1) {
       labelComponents.push('has inputs');
@@ -307,9 +310,7 @@ export class BlockSvg
    * @returns The ARIA roledescription for this block.
    */
   protected getAriaRoleDescription() {
-    if (this.isShadow()) {
-      return 'replaceable block';
-    } else if (this.outputConnection) {
+    if (this.outputConnection) {
       return 'value block';
     } else if (this.statementInputCount) {
       return 'container block';
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -243,70 +243,75 @@ export class BlockSvg
   }
 
   private computeAriaLabel(): string {
-    const {commaSeparatedSummary, inputCount} = buildBlockSummary(this);
-    let inputSummary = '';
-    if (inputCount > 1) {
-      inputSummary = 'has inputs';
-    } else if (inputCount === 1) {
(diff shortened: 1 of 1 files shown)


core/block_svg.ts
```diff
@@ -243,73 +243,78 @@ export class BlockSvg
...
+  /**
+   * Returns the ARIA role description for this block.
+   *
+   * Block definitions may override this method via a mixin to customize
+   * their role description.
+   *
+   * @returns The ARIA roledescription for this block.
+   */
+  protected getAriaRoleDescription() {
+    if (this.isShadow()) {
```
maribethb: I wonder if `replaceable` should be added to the specific type? Like "replaceable value block"? Or should `replaceable` stay in the aria label and not in the role description?

cc @BenHenning for thoughts

maribethb (approved):
Thanks, this refactor is great

core/block_svg.ts
BenHenning: Replaceable was meant to hint at it being a shadow more than its type, so I think I agree with this. If custom block types facilitate special labels then it could still work for shadow blocks (if that ever actually makes sense).

Separately, should this be protected? I assumed we never designed `Block` functionality for inheritance since Blockly doesn't support custom `BlockSvg` class implementations. I'm wondering if this needs to be something overridable in the block definition instead.

gonfunko: Great callout Maribeth, and thanks for confirming this makes sense to you as well Ben. To clarify, do we want "replaceable" in the label or the role description? I'm inclined towards the label personally, in the same part as we call out the disabled/collapsed modifiers since it seems to be the same kind of thing.

maribethb: Yes I think the label makes sense alongside the other modifiers, but defer to Ben if he thinks differently

gonfunko: Updated to mention it in the label. @BenHenning PTAL as well. Also I think block definitions are able to override protected methods via mixins?

gonfunko: Going to go ahead and merge, but if you have concerns Ben please do flag/reopen.

Issue #9496, opened 2025-12-02
Disable keyboard navigation looping


gonfunko:
Currently, navigating through blocks with the keyboard will eventually reach the "end" of the workspace and loop around to the first block stack, forever. Instead, when the end is reached, navigation should stop and an audible cue should be emitted. The audio portion may require some experimentation and validation to identify an approach that is compatible/audible in conjunction with a screenreader.

BenHenning:
Note that this changes the behavior for regular keyboard navigation users, too. I wonder if this necessitates adding jumping shortcuts for stack navigation or moving to the top/bottom stack of the workspace to allow for the same looping behavior to exist without it being default enabled.

gonfunko:
See #9498 for the shortcuts :) My inclination here is to allow toggling looping on the `LineCursor` instance to keep backwards compatibility; then the keyboard experiment can just disable it, but we can keep the current behavior for other users, and if we decide we want it back it's just a method call away to revert.

kmcnaught:
Just to flag that we didn't used to have looping on keyboard navigation, and it was a big confusion in user testing, particularly in ambiguous/confusing layouts like this (everyone pressed right arrow, nothing happened, and eventually most admitted defeat, possibly after trying down arrow to no avail too). Adding looping means that _something_ always happens, so even if the ordering is a bit weird you get some feedback and find a way back to where you want. 

<img width="1422" height="589" alt="empty forever block to the left of a taller on start stack, centre aligned so that the forever block is lower and technically after the tall stack" src="https://github.com/user-attachments/assets/dc51501e-7bd1-4e2e-9cd5-71b504c8cdc7" />

gonfunko:
Noted! With this layout and the current navigation right/down does navigate you to the next stack, even with looping turned off. The looping toggle is also effectively a one-liner, so this should be easy to revert if testing suggests that's needed or even if we want to vary the behavior contextually.

kmcnaught:
@gonfunko I think you misunderstood the example. I can still replicate the issue in [screenreader demo](https://raspberrypifoundation.github.io/blockly-keyboard-experimentation/screenreader/) (assuming that's the correct deployment?) with the following code. With the cursor in the position shown in the screenshot, `right arrow` takes you to the icon field first, and then `down arrow`/`right arrow` both result in an error tone and no position change.

[...]

PR #9511, opened 2025-12-04
feat: Make navigation looping configurable

gonfunko:
## The basics

### Resolves

Fixes part of #9496

### Proposed Changes
This PR makes the looping behavior in `LineCursor` configurable. Although `getPreviousNode()` and `getNextNode()` already allowed configuring this, `in()`, `out()`, `prev()` and `next()` did not. This PR keeps the existing behavior, but will allow the keyboard experiment to disable looping and play an alert when the end of the navigable tree is reached.

### DIFF
--- a/core/keyboard_nav/line_cursor.ts
+++ b/core/keyboard_nav/line_cursor.ts
@@ -42,6 +42,9 @@ export class LineCursor extends Marker {
   /** Locations to try moving the cursor to after a deletion. */
   private potentialNodes: IFocusableNode[] | null = null;
 
+  /** Whether or not navigation loops around when reaching the end. */
+  private navigationLoops = true;
+
   /**
    * @param workspace The workspace this cursor belongs to.
    */
@@ -64,7 +67,7 @@ export class LineCursor extends Marker {
     const newNode = this.getNextNode(
       curNode,
       this.getValidationFunction(NavigationDirection.NEXT),
-      true,
+      this.getNavigationLoops(),
     );
 
     if (newNode) {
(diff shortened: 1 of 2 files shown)


core/keyboard_nav/line_cursor.ts
```diff
@@ -221,9 +224,19 @@ export class LineCursor extends Marker {
     isValid: (p1: IFocusableNode | null) => boolean,
     loop: boolean,
   ): IFocusableNode | null {
-    if (!node || (!loop && this.getLastNode() === node)) return null;
+    const originalLoop = this.getNavigationLoops();
+    this.setNavigationLoops(loop);
```
maribethb, line 229: this whole rigamarole is unfortunate, is the loop parameter in getNextNode already released? should we just deprecate it? or potentially should get navigationLoops policy be made part of the navigation policy instead and it can pass that value into here? (This suggestion may not make sense, I forget which order these calls happen in, let me know if you want me to dig in further to find a proposal that might make more sense)

gonfunko: Yeah I don't super love it, but yes that argument is public already, hence this stack-y thing. I think the cursor feels like the right level to control this (the navigation policies should just be given X, what's the next/previous thing? with filtering and larger-scale behavior delegated to the cursor), but the existence of this argument does complicate it.

maribethb: Can you add a comment with a TODO so we can come back to this when we're designing the eventual "real" solution? The argument should probably just be removed at that point

maribethb (approved):

core/keyboard_nav/line_cursor.ts
gonfunko: Done, and yeah if we're comfortable doing that that SGTM.

PR #9514, opened 2025-12-04
fix: Don't include 'begin' in label for blocks in the first statement input

### LINKED ISSUE #9487: Remove "begin <parent block label>" narration from first child block in a C-shaped block
Currently, the first statement block nested within a C-shaped block has its ARIA description prefixed with "begin <label of the C-shaped block>". This is redundant since the C-shaped block's label would have just been read out when it previously had focus, and should be removed. For statement blocks inside a statement input that is not the first statement input (e.g. inside of an "else if" clause), this should remain, as the label/context is otherwise not available.

gonfunko:
## The basics

### Resolves

Fixes #9487

### Proposed Changes
This PR removes "Begin <parent block label>" from the ARIA label for blocks that are attached to the first statement input in their parent. The information this provides would have been previously read out when the parent block was focused.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -252,7 +252,15 @@ export class BlockSvg
     const parentInput = (
       this.previousConnection ?? this.outputConnection
     )?.targetConnection?.getParentInput();
-    if (parentInput && parentInput.type === inputTypes.STATEMENT) {
+    if (
+      parentInput &&
+      parentInput.type === inputTypes.STATEMENT &&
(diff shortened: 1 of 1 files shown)


gonfunko:
This also fixes #9457.

maribethb (approved):
This change does what the issue asks, but I'm not totally convinced the issue should ask for the thing (I left a comment there). If nobody else objects to this behavior this lgtm though

PR #9521, opened 2025-12-09
feat: Beep when navigating across nesting levels

### LINKED ISSUE #9500: Narrate nesting level when navigating between blocks
When navigating between blocks at different levels of nesting, the nesting level of the new block should be added to its ARIA label or otherwise read out to indicate its position in the hierarchy. If this is infeasible, this information could be added to the verbose readout of the block instead.

gonfunko:
## The basics

### Resolves

Fixes #9500

### Proposed Changes
This PR plays a beep when navigating between block nesting levels, increasing in pitch the more nested the destination block is. Currently this is done unconditionally; when we eventually need to merge this branch back into `main` we'll probably need to figure out a way to do it just in a screenreader context or at least make it configurable.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -1983,6 +1983,22 @@ export class BlockSvg
   }
 
   /**
+   * Returns how deeply nested this block is in parent C-shaped blocks.
+   *
+   * @internal
+   * @returns The nesting level of this block, starting at 0 for root blocks.
+   */
+  getNestingLevel(): number {
+    // Don't consider value blocks to be nested.
+    if (this.outputConnection) {
+      return this.getParent()?.getNestingLevel() ?? 0;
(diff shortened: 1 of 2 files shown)


BenHenning (approved):
This is fantastic. Is there a theoretical limit to how many levels we can go before the sound becomes too high pitched for the average person to hear? I am wondering a bit if we should start with lower tones.

Otherwise I think this is a great solution and I'm eager to see how it does with testing.

gonfunko:
Uh, assuming 20kHz is the limit of human hearing, and the 40Hz steps this is using, that'd be 490 levels of nesting. I assume it would get gratingly high pitched before it got out of audible range, but...maybe that's a useful incentive to not nest your loops 100 levels deep.

PR #9522, opened 2025-12-09
feat: Include parent category in verbose block descriptions

### LINKED ISSUE #9499: Include blocks' categories in their verbose description
When a verbose readout of a block is requested, the block's parent category name should be included on a best-effort basis. For blocks without a category, it should be omitted; for blocks in multiple categories, one may be arbitrarily chosen, or perhaps fall back to block/category color as a tiebreaker.

gonfunko:
## The basics

### Resolves

Fixes #9499 

### Proposed Changes
This PR adds blocks' parent category to their verbose description in order to better orient users. By default each category is searched through to see if the block is part of it; if no match is found (e.g. if the block is part of a dynamic category) then matching falls back to checking the block's color against each category's color.

### DIFF
--- a/core/toolbox/category.ts
+++ b/core/toolbox/category.ts
@@ -382,6 +382,14 @@ export class ToolboxCategory
   }
 
   /**
+   * Returns the colour of this category.
+   * @internal
+   */
+  getColour() {
+    return this.colour_;
+  }
+
+  /**
    * Gets the HTML element that is clickable.
    * The parent toolbox element receives clicks. The parent toolbox will add an
    * ID to this element so it can pass the onClick event to the correct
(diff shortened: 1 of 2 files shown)


core/toolbox/category.ts
```diff
@@ -381,6 +381,14 @@ export class ToolboxCategory
     return '';
   }
 
+  /**
+   * Returns the colour of this category.
```
BenHenning, line 385: Nit: line fix to address for this block.

BenHenning (approved):
Thanks @gonfunko! This generally LGTM though obviously there are CI fixes to address likely due to the circular import. If you'd like me to take another review pass after those are fixed then please ping me and I can take a quick look.

BenHenning (approved):
Nice workaround. :)

PR #9528, opened 2025-12-11
feat: Make `Flyout` an ARIA `list` (experimental)

### LINKED ISSUE #9495: Refactor the ARIA structure of the flyout
The flyout's ARIA structure should be updated as follows:

* The flyout's role should become `menu`
* Labels should have their role set to `heading`
* Each set of items between a label should be wrapped in an element with role `group`
* All other flyout items should have their role set to `menuitem`

BenHenning:
## The basics

### Resolves

Fixes #9495

### Proposed Changes

Changes a bunch of ARIA role & label management to ensure that `Flyout` acts like a list rather than a tree.

### Reason for Changes

`Flyout`s are always hierarchically flat so it doesn't make sense to model them as a tree. Instead, a menu is likely a better fit per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role:

> A `menu` generally represents a grouping of common actions or functions that the user can invoke. The `menu` role is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application. Submenus, also known as pop-up menus, also have the role `menu`.

However, there are important caveats that need to be considered and addressed:
- As discussed below, menus introduce some unexpected compatibility issues with VoiceOver so this PR presently uses `list` and `listitem`s as a slightly more generic enumerating alternative for menus.
- Menus (and to some extent lists) are stricter\* than trees in that they seem to impose a requirement that `menuitem`s cannot contain interactive elements (they are expected to be interactive themselves). This has led to a few specific changes:
  - Block children are now hidden when in the flyout (since they aren't navigable anyway).
  - Flyout buttons are themselves now the `menuitem` rather than their container parent, and they no longer use the role button.
- Menus aren't really expected to contain labels but it isn't inherently disallowed. This is less of an issue with lists.
- Because everything must be a `listitem` (or a few more specific alternatives) both blocks and buttons lack some context. Since not all `Flyout` items can be expected to be interactive, buttons and blocks have both had their labels updated to include an explicit indicator that they are buttons and blocks, respectively. Note that this does possibly go against convention for buttons in particular but it seems fine since this is an unusual (but seemingly correct) utilization of a `list` element.
- To further provide context on blocks, the generated label for blocks in the `Flyout` is now its verbose label rather than the more compact form.

\* This is largely a consequence of a few specific attributes of `menuitem` and `menu`s as a whole and very likely also applies to `tree`s and `treeitem`s (and `list`s and `listitems`s). However, now seemed like a good time to fix this especially in case some screen readers get confused rather than ignore nested interactive controls/follow semantic cloaking per the spec.

Demo of it working on VoiceOver (per @gonfunko -- note this was the `menu` variant rather than the `list` variant of the PR):

Screen Recording 2025-12-11 at 2 50 30 PM

### Test Coverage

This has been manually tested with ChromeVox. No automated tests are needed as part of this experimental change.

### Documentation

No new documentation changes are needed for this experimental change.

### Additional Information

None.


BenHenning:
@microbit-robert does this work for micro:bit? Will get this into review if you don't see any issues (and after #9449 goes in). It seems to be working well for me in ChromeVox and @gonfunko tested it on VoiceOver and didn't spot any issues so I'm hoping this addresses the root problem.

BenHenning:
PTAL @gonfunko. Apologies that this includes the changes in #9449. I will be merging this after that PR for sure, and if you don't review this before it's merged I'll make sure to update the conversation thread here with a note once the PR diff is clean.

microbit-robert:
I have the following issue on Chrome, Firefox and Safari:
- When navigating up from a statement block inside an if block, the VoiceOver cursor doesn't follow the focused element and the if block label is not read out. **Update**: the [addition of this line](https://github.com/RaspberryPiFoundation/blockly/pull/9528/files#diff-bb0de0069443e7577f62ffa96e6b51c5fafd06703f50c3eed6eb616ac87467bfR282) appears to be the cause of the VO cursor becoming decoupled from the focused element during workspace navigation. Removing it addresses this problem.

In Safari and Firefox, the changes work well and the flyout navigation seems good.

I get some very odd VoiceOver issues when testing these changes **on Chrome** only.

- When pressing 'T' to move into the toolbox, the toolbox categories are not read out, even when switching between them with up/down (this is what I experienced when I quickly attempted this change). **Update**: so, remove [this line](https://github.com/RaspberryPiFoundation/blockly/pull/9528/files) and everything works fine (but this is obviously incorrect). You can also change the role to generic or listbox or presumably almost anything else and it will work. I'm baffled. 
    - Chrome (and other Chromium-based browsers) are important to us and we recommend Chrome for MakeCode (and for user testing) because it supports WebUSB which makes downloading a MakeCode program to a micro:bit easy. Safari does not support WebUSB.

I reset my VoiceOver settings to the defaults. I'll be even more confused if these issues can't be reproduced by anyone else - specifically the Chrome only issue. An alternative approach that does seem to work in Chrome, Firefox and Safari is to replace menu and menuitem with listbox and listitem.

Screen recording using VoiceOver and Chrome:
https://github.com/user-attachments/assets/14638a8c-7e03-42cb-ae4a-846fe928f3b3

I'll try and track down what changes specifically affect things in VoiceOver - comment has been updated.

gonfunko:
@microbit-robert I can repro the Safari issue, but the Chrome/VoiceOver one works fine for me even with the same block layout and key sequence as shown in your demo video oddly enough.

core/flyout_base.ts
```diff
@@ -708,7 +708,11 @@ export abstract class Flyout
       }
     }
 
-    return this.normalizeSeparators(contents);
+    // TODO: Optimize this.
+    return this.normalizeSeparators(contents).map((item) => {
+      // aria.setRole(item.getElement().getFocusableElement(), aria.Role.MENUITEM);
+      return item;
+    });
```
gonfunko: This appears to be a no-op at present, remove or uncomment the role setting?

core/block_svg.ts
```diff
@@ -265,16 +268,17 @@ export class BlockSvg
...
-        ) {
-          childElemIds.push(connection.getFocusableElement().id);
-        }
       }
+
+      const nextConnection = this.nextConnection as RenderedConnection | null;
+      if (
+        nextConnection &&
+        nextConnection.canBeFocused() &&
+        nextConnection.type === ConnectionType.NEXT_STATEMENT
```
gonfunko: I think the check for `type === ConnectionType.NEXT_STATEMENT` is redundant given that we got the connection from the `nextConnection` field?

gonfunko (approved):

core/flyout_base.ts
BenHenning: Uh, this was a placeholder for testing and I did forget to remove it. Thanks for calling it out!

core/block_svg.ts
BenHenning: Good point--will remove this.

BenHenning:
Thanks @microbit-robert! That is very puzzling. I'm not opposed to switching to `listitem` and I was considering that originally. I think to some extent a list is even more generic than a menu, and the specification isn't super clear on when to pick one versus the other so I don't think it's strictly _wrong_ to use list here, instead (that is, I'm not sure the extra semantic benefits of menu are worth risking bugs).

BenHenning:
> * When navigating up from a statement block inside an if block, the VoiceOver cursor doesn't follow the focused element and the if block label is not read out. **Update**: the [addition of this line](https://github.com/RaspberryPiFoundation/blockly/pull/9528/files#diff-bb0de0069443e7577f62ffa96e6b51c5fafd06703f50c3eed6eb616ac87467bfR282) appears to be the cause of the VO cursor becoming decoupled from the focused element during workspace navigation. Removing it addresses this problem.

This is an issue with #9449 and is actually rather concerning. I wasn't expecting the `aria-owns` declarations to actually impact behaviors, they were meant to just fix up the hierarchy. It's rather annoying that the VO cursor isn't following the focus given that's basically its job--I will need to try and figure out which specific association is breaking it since the hierarchy is more, not less, correct than before.

BenHenning:
I was semi expecting the hierarchy changes from #9449 to cause more stability in the flyout for this PR but that may not actually be the case. I will try to decouple those changes so that this can go in independently as this is a much more pressing problem for the user test than #9449 (which is really just a stabilizing PR despite the regression).

core/block_svg.ts
BenHenning: Also, moving comment to #9449 where it's actually needed.

BenHenning (commented):
Quick self-review after detaching #9449.

BenHenning (commented):
Self-reviewed commit that changes this to a list (and makes the branch name a liar :) ).

BenHenning:
@gonfunko could you perhaps take one more quick pass on the PR now that #9449 is detached and there are some other changes here?

@microbit-robert is this working well enough to merge or are more changes needed? I can do those on Monday if so.

gonfunko (approved):

microbit-robert:
@BenHenning This now works well for me, thanks! There is some interesting numbering going on in VoiceOver with the list items in the flyout for full block fields that exist on their own (i.e., the first block in the Math flyout), but I think this is something that could be looked at separately. I would be happy if this got merged as is.

BenHenning:
> @BenHenning This now works well for me, thanks! There is some interesting numbering going on in VoiceOver with the list items in the flyout for full block fields that exist on their own (i.e., the first block in the Math flyout), but I think this is something that could be looked at separately. I would be happy if this got merged as is.

Could you file an issue for that separately @microbit-robert? Would be helpful to look at with a bit more context.

core/block_svg.ts
BenHenning: Addressed in #9449.

Issue #9411, opened 2025-10-08
Finalize prototype dynamic ARIA states for moving and inserting blocks


BenHenning:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Problem

The screen reader experimental branch doesn't currently do a great job of representing block movement and insertion which is likely to lead to confusion in upcoming user studies.

### Request

More discussion is needed on what the ideal state should be here, but at minimum:
- It should be clear to the user whether they are inserting or moving a block.
- It should be clear to the user which block they are inserting or moving.
- It should be clear to the user where the block is being added in the workspace.

BenHenning:
@rachel-fenichel, @maribethb, and @gonfunko: I wasn't around for discussions last week when I think you were all discussing move mode behaviors. Is there anything that needs to be updated in the requested behaviors before this begins being worked on per that discussion?

microbit-robert:
We've made some quick changes to the move mode announcements to better facilitate user testing, so I'll share the approach we've taken so far. In general, we've lost the use of "child of" in favour of what we believe is more user friendly language. When moving references a block (rather than an input), we use a minimal readout of the block which excludes its status, inputs and child count to keep the readout short. When moving to an input in a block, we do the same, but keep the number of inputs in the output as it's likely useful to know that you might have more than one input to choose from. 

When moving to the first position inside a c-shaped block:
<img width="645" height="337" alt="Image" src="https://github.com/user-attachments/assets/ccc0d068-9fbe-471d-a09d-7eff9042c993" />

When moving after a statement block:
<img width="629" height="317" alt="Image" src="https://github.com/user-attachments/assets/9327914e-fd53-400d-816d-cc7ce8b7b079" />

When moving before a statement block:
<img width="640" height="271" alt="Image" src="https://github.com/user-attachments/assets/a3cb7266-b2a2-4e00-aef3-072cf11472f1" />

When moving to an input, we use the lowercased input name to refer to the input, not its current value. To support MakeCode, we've actually added a public label field to `Input` and given it a `setLabel` method as the input name isn't always that good - we use the input name as a fallback:
<img width="623" height="323" alt="Image" src="https://github.com/user-attachments/assets/345d2425-401b-483f-ac17-172e3dae8050" />

When inserting or putting a block in move mode when there are no constrained positions (currently reads out nothing) we've opted for a simple announcement:
[...]

PR #9529, opened 2025-12-11
feat: Better unconstrained announcements (experimental)

BenHenning:
## The basics

### Resolves

Fixes part of #9411

### Proposed Changes

Introduces specific directional language when moving a block unconstrained, e.g. 'moved block up' also accounting for theoretical diagonal movement (which is only possible with mouse dragging currently).

### Reason for Changes

#9411 includes context on the fact that unconstrained movement has very unclear output currently since it simply outputs coordinates which is unhelpful even for a fully sighted user. Relative language provides actual context on what's happening with the action.

Future iterations of this may benefit from adding proximity context (e.g. getting near another block or distance to nearest block) perhaps in a verbose readout shortcut.

### Test Coverage

Manually tested on ChromeVox with both keyboard movement and mouse (mouse is very noisy). No automated tests needed.

### Documentation

No new docs needed.

### Additional Information

None.

### DIFF
--- a/core/block_svg.ts
+++ b/core/block_svg.ts
@@ -2150,13 +2150,13 @@ enum CoordinateShift {
   MOVE_SOUTHEAST,
   MOVE_SOUTHWEST,
   MOVE_NORTHWEST,
-  STAY_STILL
+  STAY_STILL,
 }
 
 enum AxisShift {
   SMALLER,
   SAME,
-  LARGER
+  LARGER,
 }
 
 interface BlockSummary {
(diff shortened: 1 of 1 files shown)


BenHenning:
@microbit-robert FYI I decided to spend a few minutes creating this even though it's not marked as a blocker. I'm assuming that unconstrained movement can happen even accidentally so making some of that language better seems like an easy win.

gonfunko (approved):

BenHenning:
Thanks @gonfunko! Going ahead and merging this since it seems like a net positive, though happy to revert if it ends up not being wanted. :)

PR #9531, opened 2025-12-17
fix: Make left and right arrow keys only navigate within the current block/comment

gonfunko:
## The basics

### Proposed Changes
This PR updates the behavior of the left and right arrow keys to only navigate within the current block/workspace comment, rather than traversing through the entire workspace. This is in response to feedback from screen reader user testing. Up and down may still be used to navigate from block to block and across stacks.

### DIFF
--- a/core/keyboard_nav/line_cursor.ts
+++ b/core/keyboard_nav/line_cursor.ts
@@ -345,7 +345,28 @@ export class LineCursor extends Marker {
     switch (direction) {
       case NavigationDirection.IN:
       case NavigationDirection.OUT:
-        return () => true;
+        return (candidate: IFocusableNode | null) => {
+          const candidateBlock = this.getSourceBlockFromNode(candidate);
+          const currentBlock = this.getSourceBlock();
+
+          // Preventing escaping the current block/comment/etc by:
+          // Disallow moving from a node with a block to a non-block node (other than a block comment editor)
+          // Disallow moving from a non-block node to a block node
+          // Disallow moving to the workspace
+          if (
+            (currentBlock && !candidateBlock) ||
+            (!currentBlock && candidateBlock) ||
+            candidate === this.workspace
+          ) {
+            return false;
+          }
+
+          if (!candidateBlock || !currentBlock) return true;
+
+          const currentParents = this.getOutputParents(currentBlock);
+          const candidateParents = this.getOutputParents(candidateBlock);
+          return candidateParents.intersection(currentParents).size > 0;
+        };
       case NavigationDirection.NEXT:
       case NavigationDirection.PREVIOUS:
         return (candidate: IFocusableNode | null) => {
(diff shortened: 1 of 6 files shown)


tests/mocha/cursor_test.js
```diff
@@ -163,126 +163,13 @@ suite('Cursor', function () {
...
       const prevConnection = this.blocks.A.previousConnection;
       const prevConnectionNode = prevConnection;
       this.cursor.setCurNode(prevConnectionNode);
       this.cursor.out();
       const curNode = this.cursor.getCurNode();
-      assert.equal(curNode, this.blocks.D.nextConnection);
-    });
-  });
-
-  suite('Multiple statement inputs', function () {
```
BenHenning, line 176: Are all of these test scenarios not possible anymore with the new navigation logic? It may be worth keeping the test but demonstrating that the directional moves don't actually change the position per the new logic.

core/keyboard_nav/line_cursor.ts
```diff
@@ -452,6 +473,24 @@ export class LineCursor extends Marker {
     return parents;
   }
 
+  /**
+   * Returns a set of all of the parent blocks of the given block.
```
BenHenning: Maybe mention it includes the block itself, as well.

core/keyboard_nav/line_cursor.ts
```diff
@@ -345,7 +345,28 @@ export class LineCursor extends Marker {
...
+            candidate === this.workspace
+          ) {
+            return false;
+          }
+
+          if (!candidateBlock || !currentBlock) return true;
+
+          const currentParents = this.getOutputParents(currentBlock);
+          const candidateParents = this.getOutputParents(candidateBlock);
+          return candidateParents.intersection(currentParents).size > 0;
```
BenHenning, line 374: So this inadvertently allows for navigating to the same block again, though I'm assuming that's fine (it would have been `true` before, as well, and I assume that can't actually 
happen higher up in the callstack).

I am admittedly having some difficulty understanding the purpose of the intersection logic--I suspect a brief comment would be really helpful here. My presumption is that this prevents jumping to the next/previous parent sibling because there won't be a common parental hierarchy, but I'm not entirely convinced by that. If you have a node within a c-block within another c-block, won't the outer c-block be a common parent when navigating between nodes in the innermost c-block? Perhaps I'm misunderstanding the graph direction specifically with output connections, though.

BenHenning (commented):
Thanks @gonfunko! Had a few comments, one of which is specifically looking for some clarity.

BenHenning:
Also, the keyboard nav tests look to be real failures but that's unsurprising since this is a bit change in left/right navigation.

tests/mocha/cursor_test.js
gonfunko: I was going to say no since the point of these was to verify navigation between statement blocks nested in a parent C/E-shaped block, but that can still be verified using next/previous (up/down) instead of in/out (left/right), so updated them accordingly. Thanks for flagging!

core/keyboard_nav/line_cursor.ts
gonfunko: Done.

core/keyboard_nav/line_cursor.ts
gonfunko: Yes, revisiting is fine; if looping is enabled that's allowed, and if it's not the looping check prevents it. Added a comment on the intersection logic, but the key point is that `getOutputParents()` only returns parents connected to the block's output connection, not its previous connection, so a C-shaped block will never be included since those are next connections, not value/input connections.

BenHenning: Ah, the new comment is especially helpful. Thanks!

BenHenning (approved):
Thanks @gonfunko. This LGTM!

PR #9532, opened 2025-12-18
Add screen reader support experimental (DO NOT MERGE)

BenHenning:
## The basics

### Resolves

Fixes


core/block_svg.ts
```diff
@@ -266,12 +390,14 @@ export class BlockSvg
   select() {
     this.addSelect();
     common.fireSelectedEvent(this);
+    aria.setState(this.getFocusableElement(), aria.State.SELECTED, true);
```
BenHenning, line 393: Curious if this is actually needed.

core/field.ts
```diff
@@ -328,6 +388,11 @@ export abstract class Field<T = any>
     this.bindEvents_();
     this.initModel();
     this.applyColour();
+
+    const id = this.sourceBlock_?.isSimpleReporter(true, true)
+      ? idGenerator.getNextUniqueId()
+      : `${this.sourceBlock_?.id}_field_${idGenerator.getNextUniqueId()}`;
+    this.fieldGroup_.setAttribute('id', id);
```
BenHenning, line 395: I think this isn't needed anymore but it needs to be checked.

core/field.ts
```diff
@@ -1416,7 +1482,10 @@ export abstract class Field<T = any>
  * Extra configuration options for the base field.
  */
 export interface FieldConfig {
+  type: string;
+  name?: string;
```
BenHenning, line 1486: Aren't used anymore.

core/menuitem.ts
```diff
@@ -74,10 +75,12 @@ export class MenuItem {
 
     const content = document.createElement('div');
     content.className = 'blocklyMenuItemContent';
+    aria.setRole(content, aria.Role.PRESENTATION);
     // Add a checkbox for checkable menu items.
     if (this.checkable) {
       const checkbox = document.createElement('div');
       checkbox.className = 'blocklyMenuItemCheckbox ';
+      aria.setRole(checkbox, aria.Role.PRESENTATION);
```
BenHenning, line 83: Shouldn't be needed with the parent being set to presentation.

core/flyout_base.ts
```diff
@@ -309,6 +310,9 @@ export abstract class Flyout
       'class': 'blocklyFlyout',
     });
     this.svgGroup_.style.display = 'none';
+    // Ignore the svg root in the accessibility tree since is is not focusable.
+    aria.setRole(this.svgGroup_, aria.Role.PRESENTATION);
```
BenHenning, line 314: Probably wrong since presentation will be ignored if it's used on a parent of focusable elements.

core/flyout_base.ts
```diff
@@ -334,6 +338,9 @@ export abstract class Flyout
   init(targetWorkspace: WorkspaceSvg) {
     this.targetWorkspace = targetWorkspace;
     this.workspace_.targetWorkspace = targetWorkspace;
+    if (this.targetWorkspace.isMutator) {
+      aria.setRole(this.workspace_.getFocusableElement(), aria.Role.GENERIC);
```
BenHenning, line 342: Curious why this is actually needed.

core/toolbox/separator.ts
```diff
@@ -63,6 +64,9 @@ export class ToolboxSeparator extends ToolboxItem {
       dom.addClass(container, className);
     }
     this.htmlDiv = container;
+
+    aria.setRole(this.htmlDiv, aria.Role.SEPARATOR);
```
BenHenning, line 68: This isn't needed because the separator is skipped with keyboard navigation.

core/block_svg.ts
```diff
@@ -168,6 +172,8 @@ export class BlockSvg
   /** Whether this block is currently being dragged. */
   private dragging = false;
 
+  public currentConnectionCandidate: RenderedConnection | null = null;
```
BenHenning, line 175: Should definitely be able to removed I think.

core/block_svg.ts
BenHenning: Since we aren't using `treeitem` anymore this won't actually get used. `listitem` doesn't seem to use it, either. `listbox` does, though, but it seems that `listitem` is sufficient for our uses in the flyout. This can be removed.

core/block_svg.ts
```diff
@@ -215,7 +221,124 @@ export class BlockSvg
...
+  }
+
+  private computeAriaLabelForFlyoutBlock(): string {
+    return `${this.computeAriaLabel(true)}, block`;
+  }
+
+  computeAriaLabel(
+    verbose: boolean = false,
+    minimal: boolean = false,
+    currentBlock: this | undefined = undefined,
```
BenHenning, line 255: @gonfunko I know that this code has changed a lot but, as far as I can tell, this is always `undefined` now. Does that match your analysis?

core/block_svg.ts
```diff
@@ -215,7 +221,124 @@ export class BlockSvg
...
+      this.previousConnection ?? this.outputConnection
+    )?.targetConnection?.getParentInput();
+    if (
+      parentInput &&
+      parentInput.type === inputTypes.STATEMENT &&
+      // The first statement input is redundant with the parent block's label.
+      parentInput !==
+        parentInput
+          .getSourceBlock()
+          .inputList.find((input) => input.type === inputTypes.STATEMENT)
```
BenHenning, line 273: @gonfunko would this be a given if the parent is coming from a `previousConnection`? Is there a case where a child block can be a next connection but not a statement, or an input connection but a statement? My mental model is that next/previous are always statement inputs and input/output are always value inputs.

Just thinking a bit about how this can be simplified but I don't want to miss obvious edge cases.

core/inputs/input.ts
```diff
@@ -303,6 +306,31 @@ export class Input {
...
+        return `${label} ${field.getValue()}`;
+      }, '')
+      .trim();
+    if (!fieldRowLabel) {
+      const inputs = this.getSourceBlock().inputList;
+      const index = inputs.indexOf(this);
+      if (index > 0) {
+        return inputs[index - 1].getFieldRowLabel();
+      }
+    }
```
BenHenning, line 330: @gonfunko this actually surprises me a bit. In the context of deriving a block label this definitely makes sense, but in isolation it's a bit surprising (and inverts responsibility a bit). It seems valid to me for an input to lack a field row label if it simply doesn't have any fields. Do you remember what the use case was for this bit of logic?

core/inputs/input.ts
```diff
@@ -303,6 +306,31 @@ export class Input {
...
+   *
+   * Generally this consists of the labels/values of the preceding fields, and
+   * is intended for accessibility descriptions.
+   *
+   * @internal
+   * @returns A description of this input's row on its parent block.
+   */
+  getFieldRowLabel(): string {
+    const fieldRowLabel = this.fieldRow
+      .reduce((label, field) => {
```
BenHenning, line 320: Could probably use `map` and `join`, instead, to avoid concatenation.

core/inputs/input.ts
gonfunko: I believe this was for the statement inputs on MakeCode's if blocks, which are on a row with no text (whereas Blockly's built in ones have "do"), so they'd be read out as "begin if true then" rather than just "begin"
Screenshot 2026-03-05 at 12 32 53 PM

core/block_svg.ts
```diff
@@ -1858,4 +2008,324 @@ export class BlockSvg
...
+    isNestedInput: boolean = false,
+  ): SummarySegment[] {
+    return block.inputList.flatMap((input) => {
+      const fields: SummarySegment[] = input.fieldRow
+        .filter((field) => {
+          if (!field.isVisible()) return false;
+          if (field instanceof FieldImage && field.isClickable()) {
+            return false;
+          }
+          return true;
```
BenHenning, line 2206: Could potentially be refined using `canBeFocused()` (and clean up `FieldNavigationPolicy`'s logic a bit).

core/block_svg.ts
gonfunko: I believe the invariant you're describing is correct, but that's not what this is checking; `parentInput.getSourceBlock().inputList.find((input) => input.type === inputTypes.STATEMENT)` is just a too-clever-for-its-own good way to get the first statement input on a block, to make sure that's not our parent (ie our parent is the second or third+ statement input)

core/block_svg.ts
```diff
@@ -1858,4 +2008,324 @@ export class BlockSvg
...
+
+        const targetBlock = input.connection.targetBlock();
+        if (targetBlock) {
+          const nestedSegments = recursiveInputSummary(
+            targetBlock as BlockSvg,
+            true,
+          );
+
+          if (targetBlock === currentBlock) {
+            nestedSegments.unshift({kind: 'label', text: 'Current block: '});
```
BenHenning, line 2237: Similarly @gonfunko I'm assuming this is never used now?

core/block_svg.ts
gonfunko: I changed the type and tried building, and that flagged that it's used here: https://github.com/RaspberryPiFoundation/blockly/pull/9532/changes#diff-8a5d4586670bfd492dc8c9e4b9dc38e3101065c8c60f6aacd10f87b7c22fc76eR465

core/inputs/input.ts
BenHenning, line 320: @gonfunko also curious about your take on something: couldn't this be filtered on specifically non-navigable fields? I actually assume we'd probably want this to match the 'field run' from BlockSvg's recursive label builder since it seems a bit odd to me that they can logically differ.

gonfunko: It could but I'm not sure that's desirable, e.g. for if statements I think reading out the condition, when that's a navigable/editable field, would still be good?

core/block_svg.ts
```diff
@@ -1858,4 +2008,324 @@ export class BlockSvg
...
+        input.connection &&
+        input.connection.type === ConnectionType.INPUT_VALUE
+      ) {
+        if (!isNestedInput) {
+          inputCount++;
+        }
+
+        const targetBlock = input.connection.targetBlock();
+        if (targetBlock) {
+          const nestedSegments = recursiveInputSummary(
```
BenHenning, line 2231: This is always field parts b/c nested inputs should always be omitted.

core/block_svg.ts
BenHenning: Good catch. I need to think more about shortcut announcements anyway, but yeah it seems that there's probably some level of hierarchical representation that's going to be needed I suspect.

core/inputs/input.ts
BenHenning: Yep you're right. I think there's a way to get both, though, and I will be trying to capture this in my doc.

core/inputs/input.ts
BenHenning: Good point, I wasn't sure if it was for that case and I admittedly hadn't checked the blame yet. This probably should be done at the callsite, instead, and will make note of that in the doc.

core/block_svg.ts
BenHenning: Ah got it--thanks!

core/inputs/input.ts
```diff
@@ -193,6 +193,9 @@ export class Input {
         child.getSvgRoot().style.display = visible ? 'block' : 'none';
       }
     }
+    if (this.sourceBlock.rendered) {
+      (this.sourceBlock as BlockSvg).recomputeAriaLabel();
+    }
```
BenHenning, line 198: Why is this needed?

core/inject.ts
```diff
@@ -54,6 +55,9 @@ export function inject(
     dom.addClass(subContainer, 'blocklyRTL');
   }
 
+  // Ignore the subcontainer in aria since it is not focusable
+  aria.setRole(subContainer, aria.Role.PRESENTATION);
```
BenHenning, line 59: This isn't correct. `PRESENTATION` can't be used for a parent like this.

Issue #8775, opened 2025-02-16
Remove dependency on `jsdom` (make it a peer dependency instead)


jogibear9988:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

JSDom is not needed in browser environments, but is still added cause the package json lists it as a dependency.
This is a problem if I want to build an electron app, cause it needs all dependencies, and JSDom then add 100 other ones

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

cpcallen:
This is something we would like to do but we need to do some planning to make sure it is done correctly; unfortunately we're not going to be able to do this in the immediate future.

Entkenntnis:
In the mean time, in my project I'm using this override to avoid installing `jsdom`:

```
  "overrides": {
    "jsdom": "npm:dry-uninstall"
  }
```

Hint: first remove blockly, run npm install, then add the override and reinstall blockly, otherwise the lock file is not properly updated

rachel-fenichel:
I'm retagging as triage so we can discuss whether this goes into v13.

PR #9536, opened 2025-12-19
fix!: Remove event dependencies on XML

### LINKED ISSUE #9253: Deserializing from JSON still calls XML serialization
### Check for duplicates

### Description

XML serialization requires certain DOM APIs, because it calls `document.createElementNS`. However, today, many environments exist for running JS that don't have a `document` object. We use jsdom to make this work in node, but there are still environments where using jsdom is not desired or feasible.

Using JSON serialization instead *should* be a reasonable substitute for devs operating in environments without the DOM. However, doing so still calls into the XML system, and I don't think this should be necessary. The chain is `Blockly.serialization.blocks.append` -> `new Blockly.Events.BlockCreate` (not directly as events are registrable) -> `Blockly.Xml.blockToDomWithXY`

The `BlockCreate` event serializes the created block into both json and xml. But the `run` method for the event only ever uses the `json` property, so the `xml` is created but never used. I think this was done for backwards compatibility reasons, where event listeners might use the xml.

 Removing this would be a breaking change, but wouldn't affect most people unless they have event listeners that use the xml. Those could be updated to use json or if your events really depend on the xml being present, then you can register a replacement class for `BlockCreate` that does it. So I think this should be possible to remove in a major release.

As a workaround, if you are using headless workspaces in an environment without dom, try calling `Blockly.Events.disable` before you deserialize json into the workspace.

See this [forum post](https://groups.google.com/g/blockly/c/ljVsUSSVpQ8) for a motivating example.

### Reproduction steps

1. 
2. 
3.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9253

### Proposed Changes
This PR removes dependencies on XML from the events system. This allows for headless use of Blockly in environments without DOM support, e.g. server-side NodeJS. The original issue only identified `BlockCreate` as being problematic, but `BlockDelete`, `CommentCreate` and `CommentDelete` have the same issue and have also been addressed.

### Breaking Change Details
This change is breaking for systems that depend on monitoring the XML-related fields in these events. Generally, the best migration path is to update code to use the equivalent JSON-serialized form stored in the events instead. Alternatively, you can subclass these events, add the XML-related fields in your subclass, and register your event subclasses with the Blockly registry, overriding the default implementations.

### DIFF
--- a/core/events/events_block_create.ts
+++ b/core/events/events_block_create.ts
@@ -14,9 +14,7 @@
 import type {Block} from '../block.js';
 import * as registry from '../registry.js';
 import * as blocks from '../serialization/blocks.js';
-import * as utilsXml from '../utils/xml.js';
 import {Workspace} from '../workspace.js';
-import * as Xml from '../xml.js';
 import {BlockBase, BlockBaseJson} from './events_block_base.js';
 import {EventType} from './type.js';
 import * as eventUtils from './utils.js';
@@ -27,9 +25,6 @@ import * as eventUtils from './utils.js';
  */
 export class BlockCreate extends BlockBase {
   override type = EventType.BLOCK_CREATE;
-
-  /** The XML representation of the created block(s). */
-  xml?: Element | DocumentFragment;
 
   /** The JSON respresentation of the created block(s). */
   json?: blocks.State;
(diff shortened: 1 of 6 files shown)


rachel-fenichel:
Does this mean we can get rid of our jsdom dependency?

rachel-fenichel (approved):

gonfunko:
I think it would still be necessary if you were intentionally using XML serialization (and probably in other situations too), but #8775 is tracking making it a peer dependency at least.

PR #9538, opened 2025-12-22
fix: Don't select shadow blocks on click

### LINKED ISSUE #9180: shadow block selection is inconsistent
### Check for duplicates

### Description

In Blockly v11, shadow blocks couldn't be selected. Clicking on a shadow block would always select the parent block, and the shadow block would never have visual selection. In v12, shadow blocks *usually* can't be selected by mouse, except for one case I've found which is probably a mistake.

### Temporary selection visual flash
1. Drag a "length of text" block out of the toolbox
2. Click on the shadow text block (not inside the html input element)
3. In v11, the parent block is selected immediately. In v12, the shadow block is selected while the pointer is down, but the parent block becomes selected on pointerup. When you perform a normal click, this results in a quick flash where the selection is very briefly the shadow block.

### Dragging
1. Drag of a "length of text" block out of the toolbox
2. Click and drag on the shadow text block (not inside the html input element). This causes shadow + attached parent to move together.
3. In v11, the selection is on the parent block the entire time. In v12, the shadow block will be selected from the pointerdown before the pointermove. Once you start dragging, selection moves to the parent block. But, when you finish dragging and pointerup, **selection moves back to the shadow block**. This is the only way I've been able to get the shadow block selected by mouse. This lets you copy/paste the shadow block (will be pasted as a real block) but otherwise I can't actually select one.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9180

### Proposed Changes
This PR fixes a bug that could cause shadow blocks to temporarily or permanently become selected when clicked on. Clicking on a block triggers a block start gesture, which correctly handles shadow blocks when determining what to select (their parent block). Click events bubble up to the workspace and trigger a workspace start gesture as well though, and that was focusing the start block, which is sometimes different than the target block that the block gesture start handler resolves. I updated the check for "nothing else should be focused" to check `targetBlock` rather than `startBlock`, and removed the explicit call focus the target/start block, since that would have already happened in the block start gesture handler.

I verified that shadow blocks still become selected via keyboard navigation, that clicking a shadow block focuses its parent block, that workspace comments and bubbles are still selectable, and that clicking the workspace focuses the workspace.

### DIFF
--- a/core/gesture.ts
+++ b/core/gesture.ts
@@ -777,13 +777,11 @@ export class Gesture {
     this.setStartWorkspace(ws);
     this.mostRecentEvent = e;
 
-    if (!this.startBlock && !this.startBubble && !this.startComment) {
+    if (!this.targetBlock && !this.startBubble && !this.startComment) {
       // Ensure the workspace is selected if nothing else should be. Note that
       // this is focusNode() instead of focusTree() because if any active node
       // is focused in the workspace it should be defocused.
       getFocusManager().focusNode(ws);
-    } else if (this.startBlock) {
-      getFocusManager().focusNode(this.startBlock);
     }
 
     this.doStart(e);
--- a/tests/mocha/gesture_test.js
+++ b/tests/mocha/gesture_test.js
@@ -13,6 +13,7 @@ import {
   sharedTestTeardown,
 } from './test_helpers/setup_teardown.js';
 import {dispatchPointerEvent} from './test_helpers/user_input.js';
+import {getProperSimpleJson} from './test_helpers/toolbox_definitions.js';
 
 suite('Gesture', function () {
   function testGestureIsFieldClick(block, isFieldClick, eventsFireStub) {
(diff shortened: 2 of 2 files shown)


rachel-fenichel:
> I verified that shadow blocks still become selected via keyboard navigation, that clicking a shadow block focuses its parent block, that workspace comments and bubbles are still selectable, and that clicking the workspace focuses the workspace.

Please add automated tests for these cases.

BenHenning (commented):
Nice find. Is it possible to test the specific scenarios that were breaking?

Edit: Asked this 8 minutes after Rachel did. :)

gonfunko:
Added a test to verify that clicked shadow blocks don't become selected. Ready for another look @BenHenning

BenHenning (approved):
Thanks for adding the new test, this LGTM!

PR #9539, opened 2025-12-22
fix: Fix bug that prevented the first block change event in a flyout from being dispatched

### LINKED ISSUE #9486: Flyout workspace cannot receive first BlockChange event
### Check for duplicates

### Description

When we change a block in flyout (when autoClose is false), the first change event is not received in listener, while the later change events are received correctly.

I add a logging breakpoint in devtools at `Blockly.Workspace.fireChangeListener`, it print out the change event correctly, but the listener isn't called.

### Reproduction steps

1. Run Blockly advanced playground, with the following configuration:
  - options > toolbox: simple (I use simple toolbox since it won't auto close and it is visible at first, then i can change the block in the flyout, e.g. inputs)
  - actions > logging > log flyout events: true
2. Change any fields or inputs.
3. Check devtool, BlockChange event is not printed.
4. Do step 2 again.
5. This time we can see BlockChange in devtool.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

Chrome desktop

gonfunko:
## The basics

### Resolves

Fixes #9486

### Proposed Changes
This PR adds an `inhibitReflowWrapper` instance variable to the base `Flyout` class and uses it to prevent the reflow wrapper from running in preference to unregistering and reregistering the reflow wrapper listener. The latter approach was mutating the listener array such that other listeners could be missed when dispatching events.

### DIFF
--- a/core/flyout_base.ts
+++ b/core/flyout_base.ts
@@ -136,6 +136,12 @@ export abstract class Flyout
   private reflowWrapper: ((e: AbstractEvent) => void) | null = null;
 
   /**
+   * If true, prevents the reflow wrapper from running. Used to prevent infinite
+   * recursion.
+   */
+  private inhibitReflowWrapper = false;
+
+  /**
    * List of flyout elements.
    */
   protected contents: FlyoutItem[] = [];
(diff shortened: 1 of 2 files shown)


BenHenning (commented):
This seems like a very reasonable fix. Is it possible to add a test for this or is it too contrived?

gonfunko:
Done!

rachel-fenichel (approved):

PR #9553, opened 2026-01-07
chore: Update repo URLs in messages

### LINKED ISSUE #9551: Update URLs in messages to reference RPF instead of Google
The messages contain quite a few URLs, mostly for wiki links used for the help menu on blocks, that still reference various Google URLs. These should be updated to reference the equivalent Raspberry Pi Foundation URL.

gonfunko:
## The basics

### Resolves

Fixes #9551

### Proposed Changes
This PR updates URLs of the Blockly repo in the messages files to point to the new RPF location.


maribethb (commented):
I think this is going to cause conflicts when we next pull in translate wiki updates, because the translate wiki sync is not bidirectional. updating the files for other languages here does not populate them in translatewiki.

since github handles the wiki redirection alright, I think it's fine to let the language owners update the links themselves whenever they get to it.

So I think you should only [update the messages file](https://developers.google.com/blockly/guides/contribute/core/add_localization_token) and run the script to update the autogenerated files, then just let translatewiki handle the other languages.

gonfunko:
Ah yeah, that's a good point. Reverted the commit, updated just msg/messages.js and ran `npm run messages` that updated qqq.json and en.json.

maribethb (approved):

PR #9554, opened 2026-01-07
chore: Disable macOS keyboard experiment CI

gonfunko:
## The basics
This PR disables running the keyboard-experimentation CI test under macOS, since this has been failing for months and the same test under Ubuntu is working fine. Until we can resolve the failure, this is just noise and makes it harder to spot legitimate regressions or failures.


maribethb (approved):
:(

PR #9555, opened 2026-01-08
fix: Annotate potentially blocking event listeners with passivity

### LINKED ISSUE #9542: Non-passive wheel event listener warning when both move.wheel and zoom.wheel are enabled
### Check for duplicates

### Description

When using Blockly, a Chrome console warning appears about a non-passive event listener being added to a scroll-blocking wheel event.

Image

This warning seems to occur when both move.wheel and zoom.wheel are enabled in the Blockly.inject options. Functionally, the workspace behaves as expected (mouse wheel scrolls the workspace, and Ctrl + mouse wheel zooms), and there are no noticeable performance issues. However, the warning is consistently logged, and I’m concerned about potential future implications.

I’d like to confirm whether this behavior is expected and whether there is a recommended way to avoid this warning.

### Reproduction steps

1. Use Blockly v12
2. Inject a Blockly workspace with both move.wheel and zoom.wheel enabled:
```js
Blockly.inject(blocklyDiv, {
  move: {
    scrollbars: true,
    drag: true,
    wheel: true,
  },
  zoom: {
    wheel: true,
  },
});
```
3. Open Chrome DevTools and check the console.
4. Scroll the mouse wheel over the workspace or use Ctrl + mouse wheel to zoom.
5. Observe the non-passive event listener warning in the console.

### Priority

**Low**
- There is no functional or performance issue observed.
- This is mainly a console warning and a potential future concern rather than a blocking bug.

### Stack trace

```shell
[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event.
Consider marking event handler as 'passive' to make the page more responsive.
See https://www.chromestatus.com/feature/5745543795965952

L @ blockly-IxHXmmBh.js:1
createDom @ blockly-IxHXmmBh.js:661
init @ blockly-IxHXmmBh.js:657
init @ blockly-IxHXmmBh.js:757
...
```

### Screenshots

Image

### Browsers

Chrome desktop

gonfunko:
## The basics

### Resolves

Fixes #9542

### Proposed Changes
This PR updates `(conditional)Bind` to take an additional optional parameter corresponding to the `addEventListener` [`options` argument](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options). This allows us to specify whether `touchstart` and `wheel` events are or are not passive. Passive events are those which will never call `preventDefault()`; in that case, performance can be improved because the browser can run the event handler asynchronously, whereas if there is a possibility of `preventDefault()` being called, the browser needs to run the whole event handler before scrolling or responding to the touch in case the event handler prevents it, which causes jank commensurate with the complexity of the event handler.

Chrome has started notifying about this in the console, so this also cleans that up. There is still some logging about `touchstart` in the advanced playground, but it seems that that's coming from one of its dependencies.


core/browser_events.ts
```diff
@@ -33,6 +33,18 @@ const LINE_MODE_MULTIPLIER = 40;
  */
 const PAGE_MODE_MULTIPLIER = 125;
 
+/**
+ * Options to control the behavior of bound event listeners. Based on
+ * https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options
+ * and should be kept consistent with web standards as they evolve.
+ */
+export type BindOptions = {
```
maribethb: what about 

```
type BindOptions = Parameters<typeof addEventListener>[2];
```

the magic number is not ideal, but the chances that the parameter order for `addEventListener` changes are essentially 0 so I think it's better than us having to update for a new option being added.

I might even consider inlining this into the function signature rather than exporting the type, if someone needs it they can create it the same way. or if you export it then consider naming it `EventListenerOptions` instead. if we did want to add an options object to the actual bind function then you've used the name that would be least confusing for that. you can't simply shove more options in here since it has to match the options passed to `addEventListener`

maribethb (approved):

core/browser_events.ts
gonfunko: Thanks for flagging this, I dug a bit more and turns out there's actually a `AddEventListenerOptions` type known to TS that's used for the actual argument to `addEventListener()`. Updated to just use that.

maribethb: oh, sweet! nice find

PR #9559, opened 2026-01-09
fix: Fix typo in unregisterWorkspace function name

suk-6:
## The basics

### Resolves

Fixes 

### Proposed Changes

Fix typo in unregisterWorkspace function name


maribethb:
Incredible, this typo has been here for at least 4 years undetected.

Unfortunately changing it directly like this would be a breaking change, in case anyone's using the old name. We need to:

- Add a new, correctly spelled function, like you've done here.
- Keep the old misspelled one and just have it call the new one.
- Add a [deprecation warning](https://github.com/RaspberryPiFoundation/blockly/blob/main/core/utils/deprecation.ts#L23) to the old one that it will be removed in v13.
- Add a [rename to the migration database](https://github.com/RaspberryPiFoundation/blockly/blob/main/scripts/migration/renamings.json5) so that folks can automatically update their code when migrating to v13.
- Update the internal usages of the function.

Let us know if you're still interested in fixing this with these additional steps. Thanks!

suk-6:
Added `unregisterWorkpace` to the migration database. Is this the right approach?

scripts/migration/renamings.json5
```diff
@@ -1595,5 +1595,16 @@
     },
   ],
 
+  '13.0.0': [
+    {
+      oldName: 'Blockly.common',
+      exports: {
+        unregisterWorkpace: {
+          newExport: 'unregisterWorkspace'.
```
maribethb:
```suggestion
          newExport: 'unregisterWorkspace',
```

maribethb (commented):
Nice, thank you! One suggestion then lgtm.

suk-6:
As the repository changed to a monorepo, I merged the main branch and made changes to core/workspace.ts.

maribethb:
Thank you again for this PR, I've merged your commits in #9936 :)

PR #9560, opened 2026-01-09
fix: Modernize audio playback

### LINKED ISSUE #3536: Preloading sounds have become noticeable in Chrome
We need to preload and cache all the sounds, otherwise when we play a sound the first instance will play a second late.  The [Audio](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio) API has a 'preload' attribute that's designed to provide exactly this.  However, when tested back in 2014, this did not work across all browsers.  So we chose to [play all sounds at minimum volume](https://github.com/google/blockly/blob/master/core/workspace_audio.js#L109) upon page load.  At a volume of just 1%, this didn't used to be audible.

However, this preloading is now audible.  Crank your volume up, and load any of the [demo pages](https://blockly-demo.appspot.com/static/demos/index.html) in Chrome (at least on a Mac).  On page load, one can clearly hear the click and delete sounds.  Despite the code, there's no way that this is 1% volume.

I think it's time that Blockly take another look at sounds and determine if there are better ways of preloading sounds in the majority of browsers.  We will probably need to keep the hacky low-volume preload technique for IE, but at a minimum I'd expect that Chrome and FF should have proper preloading abilities.  Also be aware that iOS can apparently only play one sound.  This is a swamp of browser-specific gotchas.

Surfaced due to https://github.com/google/blockly-games/issues/171

### LINKED ISSUE #4538: Connecting/disconnecting block audio does not work
**Describe the bug**
The audio for connecting and disconnecting a block does not work on ios.

**To Reproduce**

1. Go to the playground
2. Try to connect to blocks
**Notice how no sound is emitted**

**Expected behavior**
It should make the connecting blocks sound.

**Screenshots**

 - Device: iPhone6, iPhone X
 - OS: iOS 12, iOS14
 - Browser: Safari

**Stack Traces**

```
Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```

**Additional context**

### LINKED ISSUE #5798: First user interaction on multiplayground on iOS triggers angry clacking sounds
**Describe the bug**
The first user interaction (tap) on the multiplayground (https://google.github.io/blockly/tests/multi_playground.html) triggers a bunch of angry clacking sounds.

**To Reproduce**
1. Visit https://google.github.io/blockly/tests/multi_playground.html on an iOS device
2. Tap on a flyout category
3. Listen to the angry clacking

**Expected behavior**
The first user interaction is silent.

**Device**
 - Device: 11" iPad Pro
 - OS: iOS 15.1
 - Browser: Safari

gonfunko:
## The basics

### Resolves

* Fixes #3536
* Fixes #4538
* Fixes #5798

### Proposed Changes
This PR modernizes Blockly's audio playback to remove decade-old hacks:

* Preloading has been improved by fetching the audio file when it is loaded and storing its bytes, rather than creating `HTMLAudioElement` instances and playing them on page load to try to trick the browser into fetching and caching them. Because we're no longer relying on playing `HTMLAudioElement`s, various hacks to do this in response to user interaction in order to prevent the browser suppressing it are also unnecessary. This incidentally fixes several situations whereby the preloading/playback could be heard on pageload, because we no longer play sounds until we actually intend to play them.

* The OGG and WAV files have been removed; [MP3 is supported by all modern browsers](https://caniuse.com/mp3), so these fallbacks were just wasting bandwidth.

* Rather than creating and cloning audio tags based on useragent, we now just use the WebAudio API to play sounds.

### Testing
I manually verified that block connection, disconnection and delete sounds play properly on Safari, Mobile Safari (in the simulator), Chrome and Firefox. I also verified that bulk deletion after e.g. "Scatter" in the playground plays lots of delete sounds in quick succession properly.


BenHenning (approved):
This is a fantastic cleanup. I have nothing to suggest.

PR #9565, opened 2026-01-12
fix!: Normalize Zelos connection indicators

### LINKED ISSUE #3659: Zelos does not support input outlines around filled inputs - causes preview weirdness
**Describe the bug**

Zelos does not allow you to trigger an input outline around an input that has been filled.

This means that `Blockly.zelos.Renderer.prototype.getConnectionPreviewMethod` has to return a logically incorrect value (`REPLACMENT_FADE`) to get the desired behavior.

This only works because the replacement fade appears the same as an input outline (in zelos specifically). If they ever stop being the same, this will break.

**Expected behavior**

Zelos should support input outlines around filled inputs. This way the `Blockly.zelos.Renderer.prototype.getConnectionPreviewMethod` could return the correct value.

**Additional context**

I think all of the above information is correct, but I'm not super familiar with zelos so I could be wrong.

[TODO link](https://github.com/google/blockly/blob/e2b510662cf60e4dba1e228f3b83549ea694e951/core/renderers/zelos/renderer.js#L129)

### LINKED ISSUE #9338: Provide connection previews when connecting a moving value block's input to an existing block (zelos)
### Check for duplicates

### Problem

When dragging a value block with external inputs, if the block's inputs are moved within radius of an unconnected value output on the workspace a connection preview will appear if using geras or thrasos:

https://github.com/user-attachments/assets/da346c8b-039a-49d1-8d36-f519ab4f3f9a

For some reason, that is not not if using zelos, even though all the same connections are possible:

https://github.com/user-attachments/assets/18696ed6-843e-403b-b2c8-a11c97450ce5

### Request

Given that blocks will still connect even absent a connection preview, it would seem to be preferable to be consistent and alway showing a connection preview if there is a connection candidate.

### Alternatives considered

I note that Scratch and MakeCode both seem to forbid connecting a moving block via one of its value inputs, probably by not having any blocks with external inputs (since inline inputs do not connect in this way).

### Additional context

I note that issue #3659 also concerns differences between zelos and the other renderers about how they handle inputs, though as it seems to relate to filled inputs (rather than unconnected ones) I am not sure it is directly relevant.

gonfunko:
## The basics

### Resolves

* Fixes #3659
* Fixes #9338

### Proposed Changes
This PR cleans up Zelos' display of proposed connection indicators. Previously, proposed input connections indicators had three possible appearances:

Screenshot 2026-01-12 at 3 06 10 PM (2)

Fuzzy, when an existing block would be replaced

Screenshot 2026-01-12 at 3 06 04 PM (2)

Normal, when a block would be connected to an empty input

Screenshot 2026-01-12 at 3 05 59 PM (2)

Skinny, same circumstances as above but when the highlight ring happened to be z-ordered below the fill for the input slot.

Likewise, next/previous connection indicators could appear normal, or partially obscured by the insertion marker depending on the whims of z-ordering:

Screenshot 2026-01-12 at 3 07 12 PM (2)
Screenshot 2026-01-12 at 3 07 09 PM (2)

Now:

* The "replacement fade" (fuzzy) filter has been entirely eliminated; the connection highlight for replacing a block appears the same as the highlight for inserting a block into an empty input connection
* Connection highlights are always brought to the front, so the skinny/obscured appearances are eliminated. This also normalizes them and makes them consistently prominent for Geras/Thrasos.
* Input connection highlights are shown in Zelos on both the dragging and stationary block, as appropriate; previously, if the block being dragged was aligned in such a way that it was eligible to connect to an input on a stationary block, there was no visual indication of this.
* `BlockSvg.highlightShapeForInput()` has been removed, as it was internal and had no callers.

### Breaking Changes
* `REPLACEMENT_GLOW_COLOUR`, `REPLACEMENT_GLOW_SIZE`, and `replacementGlowFilterId` have been removed from `Blockly.zelos.ConstantProvider`. References to these fields should be removed; the appearance of the connection indicator can be styled with CSS targeting `.blocklyHighlightedConnectionPath`.
* `IPathObject.updateReplacing()` has been added; this method is optional and the root `PathObject` defines it by simply toggling a class, so this should not strictly be breaking, but custom renderers may wish to implement/override it.


BenHenning:
I haven't looked at this in depth quite yet, but how are highlights now consistently brought to the front to fix the z-ordering issue, and does this introduce any problems with keyboard navigation? I remember digging into this before and I found that the standard `bringToFront` will break focus since it changes the DOM and many DOM operations just completely reset focus back to the document body. I think we built some resilience to this failure type in places (and maybe in `FocusManager` itself) but I would need to dig to confirm.

Edit: Ah, #8981 was the one that may have made this work. I'm still curious about the first part, though, because I don't see `bringToFront` used anywhere.

packages/blockly/core/rendered_connection.ts
```diff
@@ -332,6 +332,7 @@ export class RenderedConnection
     const highlightSvg = this.findHighlightSvg();
     if (highlightSvg) {
       highlightSvg.style.display = '';
+      highlightSvg.parentElement?.appendChild(highlightSvg);
```
BenHenning, line 334: This _might_ cause focus loss problems. I'm also curious why this is necessary--shouldn't the highlight SVG already be properly parented since it exists in the DOM at this stage?

Or is this the means to try and 'bring it to front'?

BenHenning (commented):
Other than my specific concerns on focus this seems like an excellent simplification.

packages/blockly/core/rendered_connection.ts
gonfunko: Yes, this is the means to bring it to the front. Experimentally, there don't appear to be any focus issues, I think because the highlight SVG is not itself the thing that has focus; I was able to move blocks using keyboard nav and Zelos and it behaved as expected.

BenHenning: Wait that's slightly confusing as well. The highlight SVG should definitely hold focus but perhaps this case isn't hit today because we can't actually navigate to the connections themselves. I wonder if focus would still behave correctly if we temporarily re-added connections to possible navigable locations.

BenHenning:
Sorry this dropped off my radar @gonfunko. Just followed up to the comment thread.

It also seems like CI is failing--might be worth taking a look in case there are any issues with the PR changes.

packages/blockly/core/rendered_connection.ts
gonfunko: I don't think that makes a difference; the highlight SVG is the focusable element for the connection, but this is run in two situations: when the connection is being highlighted as a potential connection point, in which case the focused thing is the block being dragged, or in response to onNodeFocus() being called, which the focus manager invokes prior to actually focusing the element, so the DOM shuffle happens before focus gets assigned.

gonfunko:
@BenHenning No worries, I also replied in the thread just now. CI failures are across all of v13, the closure compiler dep just needs to be updated.

packages/blockly/core/rendered_connection.ts
BenHenning: I think I've run into issues with some of these DOM elements being sort of slightly delayed, but if you don't notice any problems then I think this is fine to go in as-is.

BenHenning (approved):
Awesome fix overall, thanks @gonfunko!

packages/blockly/core/rendered_connection.ts
gonfunko: It seems to be fine; I'll go ahead and merge, this is going into v13 so we'll get lots of testing on it before it goes out.

PR #9568, opened 2026-01-13
fix: Fix bug that prevented redoing adding empty block comments

### LINKED ISSUE #6708: Undo and redo fail to re-add comment to block
**Describe the bug**

* Redo won't re-add a comment to a block that was removed by undoing the addition.
* Undo won't re-add a comment to a block that was removed by using the Remove Comment context menu item.

**To Reproduce failure**

Steps to reproduce the redo fail behaviour:
1. Go to blockly playground.
2. From the Logic category, add an "if" block to the workspace.
3. Right-click on the new block, choose "add comment"
4. Press undo (⌘Z on macOS)
5. Press redo (⌘⇧Z on macOS)

Steps to reproduce the undo fail behaviour: steps 1-3 as above, then
4. Right-click on the new block, choose "remove comment"
4. Press undo (⌘Z on macOS)

**Expected behaviour**

* In the first case, the comment indicator (?) should disappear when undo is pressed, and then reappear when redo is pressed.
* In the second case, the comment indicator (?) should disappear when "remove comment" is selected, and then reappear when undo is pressed.

**Actual behaviour**

The comment indicator disappears and does not reappear.

 - OS: macOS 12.6.1
 - Browser: Firefox 107.0.1 (64-bit), Chrome 108.0.5359.98
 - Version: develop @ 3be3d4a (i.e., 9.2.0 release candidate)

**Stack Traces**

No console ouptut.

**Additional context**

By adding additional items to the undo stack it can be observed that the action is undone/redone (i.e., it takes up one step of a sequence of undo/redo steps), but without observable effect.

Adding text to the comment before deleting it causes it to be correctly restored.

gonfunko:
## The basics

### Resolves

Fixes #6708

### Proposed Changes
This PR fixes a bug that prevented redoing adding a comment to a block. By default comments are added as `''`, which is falsey, so when the comment text was set it fell back to null. I switched it to `??` to only fall back in the case of `undefined`.

### DIFF
--- a/tests/mocha/comment_test.js
+++ b/tests/mocha/comment_test.js
@@ -167,4 +167,27 @@ suite('Comments', function () {
       assertBubbleLocation(this.comment, 100, 100);
     });
   });
+  suite('Undo/Redo', function () {
+    test('Adding a comment can be undone', function () {
+      const block = this.workspace.newBlock('empty_block');
+      block.setCommentText('');
+      assert.isNotNull(block.getIcon(Blockly.icons.IconType.COMMENT));
+      assert.equal(block.getCommentText(), '');
+
+      this.workspace.undo(false);
+
+      assert.isUndefined(block.getIcon(Blockly.icons.IconType.COMMENT));
+      assert.isNull(block.getCommentText());
+    });
+
+    test('Adding a comment can be redone', function () {
+      const block = this.workspace.newBlock('empty_block');
+      block.setCommentText('');
+      this.workspace.undo(false);
+      this.workspace.undo(true);
+
+      assert.isNotNull(block.getIcon(Blockly.icons.IconType.COMMENT));
+      assert.equal(block.getCommentText(), '');
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


tests/mocha/comment_test.js
```diff
@@ -167,4 +167,27 @@ suite('Comments', function () {
       assertBubbleLocation(this.comment, 100, 100);
     });
   });
+  suite('Undo/Redo', function () {
+    test('Adding a comment can be undone', function () {
```
BenHenning: Nit here & below: maybe clarify that these are empty comments (since that's what's necessary to trigger the actual failure).

It might be nice to also add an undo and undo/redo test for non-empty comments just for more comprehensive coverage.

BenHenning (approved):
Thanks @gonfunko! One small suggestion to improve test clarity and another for adding a bit more test coverage, but otherwise this looks good.

tests/mocha/comment_test.js
gonfunko: Done!

PR #9570, opened 2026-01-14
fix: Make `MenuItem` methods toggle classes immediately

### LINKED ISSUE #8801: CSS classes/state aren't updated when menu items are modified
### Check for duplicates

### Description

The setters in MenuItem only toggle the backing field, but don't update the CSS classes attached to the menu item's DOM element, other than `setHighlighted()`. They should do so to allow for dynamically changing the state of menu items while the menu is being displayed, without having to close and reopen it.

gonfunko:
## The basics

### Resolves

Fixes #8801 

### Proposed Changes
This PR makes the various methods on `MenuItem` that toggle properties like checking, enabling, RTL, etc immediately update the state of the menu item without requiring the entire menu to be torn down and recreated to reflect the new state. It also adds tests for `MenuItem`.

### DIFF
--- a/core/menuitem.ts
+++ b/core/menuitem.ts
@@ -256,6 +256,13 @@ export class MenuItem {
     this.actionHandler = fn.bind(obj);
   }
 
+  /**
+   * Adds or removes the checkmark indicator on this menu item.
+   * The indicator is present even if this menu item is not checked, as long
+   * as it is checkable; its visibility is controlled with CSS.
+   *
+   * @param add True to add the checkmark indicator, false to remove it.
+   */
   private toggleCheckbox(add: boolean) {
     if (add) {
       if (
--- a/core/menuitem.ts
+++ b/core/menuitem.ts
@@ -198,10 +198,12 @@ export class MenuItem {
    */
   setHighlighted(highlight: boolean) {
     this.highlight = highlight;
-    this.getElement()?.classList.toggle(
-      'blocklyMenuItemHighlight',
-      this.highlight,
-    );
+    if (this.isEnabled()) {
+      this.getElement()?.classList.toggle(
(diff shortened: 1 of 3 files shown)


core/menuitem.ts
```diff
@@ -186,15 +198,10 @@ export class MenuItem {
...
-    const el = this.getElement();
-    if (el && this.isEnabled()) {
-      const name = 'blocklyMenuItemHighlight';
-      if (highlight) {
-        dom.addClass(el, name);
-      } else {
-        dom.removeClass(el, name);
-      }
-    }
+    this.getElement()?.classList.toggle(
```
maribethb: previously we were checking `this.isEnabled()` as well, should that check remain?

core/menuitem.ts
```diff
@@ -243,4 +255,33 @@ export class MenuItem {
...
   }
+
+  /**
+   * Adds or removes the checkmark indicator on this menu item.
+   * The indicator is present even if this menu item is not checked, as long
+   * as it is checkable; its visibility is controlled with CSS.
+   *
+   * @param add True to add the checkmark indicator, false to remove it.
+   */
+  private toggleCheckbox(add: boolean) {
```
maribethb: the method name makes it sound like you're toggling whether the box is checked or not. what about `toggleHasCheckbox()`  or similar?

maribethb (approved):

core/menuitem.ts
gonfunko: Good catch, yes! Added a test for this as well.

core/menuitem.ts
gonfunko: Done

maribethb (approved):

PR #9571, opened 2026-01-14
feat: Add dynamic keycode for primary modifier key

### LINKED ISSUE #8836: Add a custom key code that means "control on windows and command on mac"
### Check for duplicates

### Problem

For keyboard shortcuts that use "control" on windows and "command" on mac, the author of the shortcut has to manually register both keys (and blockly calls the "command" key "meta" which is potentially confusing as windows also has a separate "meta" key)

### Request

Add a custom key code to the list of allowable modifier keys. If this key code is registered to a shortcut, blockly will automatically use the appropriate platform-specific key.

Name tbd: `CTRL_CMD` ?

This is inspired by the [webdriver custom keycode](https://webdriver.io/docs/api/browser/keys/) that has this behavior.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8836

### Proposed Changes
This PR adds `KeyCodes.CTRL_CMD`, which dynamically resolves to either Control or Command depending on the current platform. This allows keyboard shortcuts to be registered with the expected modifier key on all platforms, and without also being registered for other platforms' modifier key.


core/utils/keycodes.ts
```diff
@@ -151,4 +153,8 @@ export enum KeyCodes {
   // indicates a hardware/bios problem.
   // http://en.community.dell.com/support-forums/laptop/f/3518/p/19285957/19523128.aspx
   PHANTOM = 255,
+
+  // The primary modifier key on the current platform, i.e. Command on Apple
+  // platforms and Control elsewhere.
+  CTRL_CMD = userAgent.MAC || userAgent.IPHONE || userAgent.IPAD ? 91 : 17,
```
BenHenning: Could this be changed to reference `CTRL` and `MAC_WK_CMD_LEFT` directly rather than repeating the constant values directly?

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -158,87 +158,42 @@ suite('Keyboard Shortcut Items', function () {
         'hideChaff',
       );
     });
-    const testCases = [
```
BenHenning, line 161: For here & below: should we perhaps keep these? It seems useful to actually test that both types of shortcuts work on their respective systems (though I assume this will require adapting the parametrization to support updating the user agent so that the correct key code is selected?

Or, are you able to confirm that both ctrl and cmd are tested based on whether these tests are run on Linux or OS X?

BenHenning (commented):
Thanks @gonfunko! It's a nice simplification. I have a couple of suggestions, PTAL.

packages/blockly/tests/mocha/shortcut_items_test.js
gonfunko: They should be both tested on the respective platforms, but we don't actually run CI under macOS (see #2114). I'm also a bit stuck on how to actually deal with the user agent; Sinon can't stub the fields directly because ES Modules, and the useragent utility resolves the user agent once on load and stores the results in the non-stubbable constants, so even if you change navigator.userAgent you'd need to do a full page reload for that to be captured I think.

core/utils/keycodes.ts
gonfunko: Done

packages/blockly/tests/mocha/shortcut_items_test.js
BenHenning: That is rather unfortunate, but I see there aren't a lot of options here then. Got it--thanks for explaining!

BenHenning (approved):
Sorry, this dropped off my radar.

Thanks @gonfunko. LGTM!

PR #9573, opened 2026-01-15
fix: Make the undo/redo API more ergonomic

### LINKED ISSUE #8167: Improve readability and maintainability of undo method
### Check for duplicates

### Problem

The `Workspace.undo()` method takes one bool parameter named "redo" which, if false, performs the undo action as one would expect of an "undo()" method, and if true, performs a redo of the last action that was reverted using `undo()`.
Since the name of the parameter is not visible when calling the method (e.g. `undo(false)`), this seems unintuitive and decreases readability and maintainability of Blockly code.
See [Blockly reference](https://developers.google.com/blockly/reference/js/blockly.workspace_class.undo_1_method).

### Request

* Option 1:
Instead of just having one `undo()` method and a bool parameter, introduce a parameterless `redo()` method and remove the parameter of `undo()`. When using Blockly, the code would call `myWorkspace.undo()` or `myWorkspace.redo()` without any parameters.
In order to avoid duplicate code, internally this could call a "performUndoRedo" method and pass a parameter that determines if undo or redo should be performed. Ideally, the type of this parameter would be an enum, e.g. "UndoRedoAction", with two elements, e.g. "UndoRedoAction.UNDO" and "UndoRedoAction.REDO".
This could be made backward compatible by making the bool parameter optional, instead of removing it, and using *false* as the default value.

* Option 2:
Use the internal implementation with the enum described above, but make it the official implementation. That means, the code using Blockly would call `myWorkspace.performUndoRedo(UndoRedoAction.UNDO)` or `myWorkspace.performUndoRedo(UndoRedoAction.REDO)`.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8167

### Proposed Changes
This PR updates `Workspace.undo()` to have a default argument of `false`, corresponding to undo, and adds a tiny `Workspace.redo()` wrapper method. This allows users to simply `.undo()` or `.redo()` without having to remember or correctly reason about the argument, and preserves backwards compatibility.


packages/blockly/core/workspace.ts
```diff
@@ -762,6 +762,13 @@ export class Workspace {
     }
   }
 
+  /**
+   * Redoes the previous action.
+   */
+  redo() {
```
BenHenning, line 768: Small suggestion: it would be nice to have a few new tests specifically for this method (since it's now part of the class's public API which means it has its own contract that should be enforced). Presumably these could just be a copy of the redo `undo` tests, but I also don't think they should replace those since `undo`'s contract isn't changing (other than the new defaulting behavior and I don't think that needs to necessarily be tested).

BenHenning (approved):
Thanks @gonfunko! Another nice simplification. Just had one suggestion on improving testing but otherwise this LGTM.

PR #9574, opened 2026-01-15
refactor!: Deprecate `Block.getVars()`

### LINKED ISSUE #1498: Deprecate getVars and renameVar and remove usages in core
https://github.com/google/blockly/pull/1495/files makes these unnecessary, but there are still some uses in the generators and procedure blocks.  Fix up all uses and clean up code.

gonfunko:
## The basics

### Resolves

Fixes #1498

### Proposed Changes
This PR deprecates `Block.getVars()` in favor of `Block.getVarModels()`, and removes `@internal` from the latter. This method is more flexible as it returns the entire variable model objects rather than just their IDs (or, in the case of procedure blocks, their names). `renameVar` mentioned in the bug was removed long ago.

### Breaking Changes
If you are calling `Block.getVars()`, you should call `Block.getVarModels()` instead. To get just the IDs, `Block.getVarModels().map(model => model.getId())` should be identical to the old behavior.

### DIFF
--- a/core/block.ts
+++ b/core/block.ts
@@ -50,6 +50,7 @@ import * as registry from './registry.js';
 import * as Tooltip from './tooltip.js';
 import * as arrayUtils from './utils/array.js';
 import {Coordinate} from './utils/coordinate.js';
+import * as deprecation from './utils/deprecation.js';
 import * as idGenerator from './utils/idgenerator.js';
 import * as parsing from './utils/parsing.js';
 import {Size} from './utils/size.js';
@@ -1139,9 +1140,16 @@ export class Block {
   /**
    * Return all variables referenced by this block.
    *
+   * @deprecated v13: Use Blockly.Block.getVarModels().map(m => m.getId())
    * @returns List of variable ids.
    */
   getVars(): string[] {
+    deprecation.warn(
+      'Blockly.Block.getVars()',
+      'v13',
+      'v14',
+      'Blockly.Block.getVarModels().map(model => model.getId())',
+    );
     const vars: string[] = [];
     for (const field of this.getFields()) {
       if (field.referencesVariables()) {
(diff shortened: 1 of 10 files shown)


core/block.ts
```diff
@@ -1139,9 +1140,16 @@ export class Block {
...
    *
+   * @deprecated v13: Use Blockly.Block.getVarModels().map(m => m.getId())
    * @returns List of variable ids.
    */
   getVars(): string[] {
+    deprecation.warn(
+      'Blockly.Block.getVars()',
+      'v13',
+      'v14',
+      'Blockly.Block.getVarModels().map(model => model.getId())',
```
maribethb, line 1151: Would be cool if our renaming script actually like, compiled the code and found references to this method and could rename them. I was going to suggest adding this to the renaming script but that doesn't actually help since it's never going to be called with the full namespace, only as `block.getVars()` where `block` is a user-defined name. I wonder if we could write a vscode plugin or something that did this. obviously not relevant to this PR :P

tests/mocha/blocks/procedures_test.js
```diff
@@ -1973,20 +1973,26 @@ suite('Procedures', function () {
...
             function assertArgs(argArray) {
               assert.equal(
-                this.defBlock.getVars().length,
+                this.defBlock.getVarModels().length,
                 argArray.length,
                 'Expected the def to have the right number of arguments',
               );
               for (let i = 0; i < argArray.length; i++) {
-                assert.equal(this.defBlock.getVars()[i], argArray[i]);
+                assert.equal(
```
maribethb: these tests could just use deepEqual instead of having to check the length and then each member individually

maribethb (approved):

tests/mocha/blocks/procedures_test.js
gonfunko: Done

core/block.ts
gonfunko: Yeah that would be fancy. This also feels like a thing our LLM overlords could handle pretty effectively.

maribethb (approved):

PR #9576, opened 2026-01-20
fix: Improve flyout behavior

### LINKED ISSUE #8547: Remove explicit variable handling from flyout
Flyouts used to be hardcoded to handle blocks and labels, now we have the ability to have arbitrary items in flyouts with the FlyoutInflater design (see design doc here [Arbitrary items in flyouts](https://docs.google.com/document/d/1Fmw309-JTh0db8JNm3GCFVH-vfuBdpsyGsGHNUIAC9M/edit?tab=t.0#heading=h.xia80uwom2zh))

The flyout base class still does some special handling for variables, in particular it fires variable creation events manually since the events are disabled during flyout creation. The doc linked above says:

Currently the flyout has some logic to fire variable creation events after the flyout is populated, because it disables events while it is being populated. I don’t think disabling the events is actually necessary, because it only changes the order of undoable events relative to UI events (not undoable events relative to each other). But this should be verified during actual implementation.

This may or may not be related to #9486 (not actually related but whatever solution we have for this might impact the solution to that or vice versa)

### LINKED ISSUE #8769: Dragging in from the flyout sometimes bumps unexpectedly
### Check for duplicates

### Description

Image

If the root block is at or near 0, 0 and another (not yet understood) set of conditions is met, a bump occurs when dragging a block out of the flyout. It also occurs when creating a block from the flyout with the keyboard experimentation plugin.

This happens because of a render that is triggered in `serialization.blocks.append` when called from `flyout_base.placeNewBlock` and can be resolved by switching to `blocks.appendInternal` to avoid the immediate render, but we want to understand the issue more before making that change.

### Reproduction steps

Open the playground and load this JSON:

```
{
  "blocks": {
    "languageVersion": 0,
    "blocks": [
      {
        "type": "controls_repeat_ext",
        "id": "$~GZGETMKa80_E_cnRR4",
        "x": 0,
        "y": 0,
        "inputs": {
          "TIMES": {
            "shadow": {
              "type": "math_number",
              "id": "~#{8T0:+h23YhNbVGTF+",
              "fields": {
                "NUM": 10
              }
            }
          }
        }
      }
    ]
  }
}
```

Then drag the workspace (not the block) to offset it away from the flyout for visibility.
Then open the Logic category and slowly drag out an if block.

Notice that the repeat block is bumped sideways as the if block appears.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #9239: It's harder to drag out blocks if the flyout has a scrollbar
Chrome/Mac:
 
1. Load https://blockly-demo.appspot.com/static/tests/playground.html
2. Ensure your browser window is tall enough that the flyout for Logic doesn't have a scrollbar
3. Drag an "if" block from "Logic" ensuring your initial drag is roughly vertically down, as you might do if headed to the bottom left of the visible workspace under the flyout
4. Resize your browser window so that the "Logic" category has a scrollbar.
5. Attempt the same drag as (3) and note it doesn't drag the block at all unless you change your motion to move more to the right or left while still on the block.

I think this is an interaction with a click/hold/move to scroll feature. If I repeat the same steps but click/hold and drag upward you can see the scrolling. But it's very confusing when you don't appreciate the feature exists and you unintentionally try to scroll down. It's not a feature desktop mouse users typically expect.

In practice I encounter this in MakeCode with toolbox categories that always have a scrollbar on typical displays, rather than because I use Blockly in weirdly small windows. In MakeCode the first "Basic" category has relatively few blocks and no scrollbar on my screen but other categories typically have enough to scroll, so I encounter this different feel/technique to the drag regularly.

This also reproduces on Android/Chrome where a scrollbar is much more likely.

Would it be better for the scroll behaviour to only apply if dragging the flyout background? At least for mouse interactions, not sure about touch, especially if you're continuing a scroll in a long flyout.

https://github.com/user-attachments/assets/bcf366b3-927f-4e30-ba41-17339ecd9f7f

gonfunko:
## The basics

### Resolves

* Fixes #8547
* Fixes #8769
* Fixes #9239

### Proposed Changes
This PR is a bit of a grab bag of flyout related fixes. Specifically, it:

* Removes a bunch of special handling of variable and block creation events when dragging a block out of the flyout. Experimentally this still seems to work fine, including undo, and including blocks that reference potential variables, and all tests still pass.
* Avoids a race condition that could result in inadvertent block bumps when dragging out a new block by calling `appendInternal()` instead of `append()`, the latter of which triggers an immediate (and unnecessary) render before the state of things is fully combobulated.
* Changes the behavior such that dragging on a block will not ever drag-to-scroll the flyout. The flyout can still be scrolled by clicking and dragging on the flyout workspace background, but clicks targeting blocks will always be treated as a drag on the block.

### DIFF
--- a/packages/blockly/core/flyout_base.ts
+++ b/packages/blockly/core/flyout_base.ts
@@ -38,7 +38,6 @@ import * as dom from './utils/dom.js';
 import * as idGenerator from './utils/idgenerator.js';
 import {Svg} from './utils/svg.js';
 import * as toolbox from './utils/toolbox.js';
-import * as Variables from './variables.js';
 import {WorkspaceSvg} from './workspace_svg.js';
 
 /**
@@ -813,44 +812,16 @@ export abstract class Flyout
    * @internal
    */
   createBlock(originalBlock: BlockSvg): BlockSvg {
-    let newBlock = null;
-    eventUtils.disable();
-    const variablesBeforeCreation = this.targetWorkspace
-      .getVariableMap()
-      .getAllVariables();
-    this.targetWorkspace.setResizesEnabled(false);
     try {
-      newBlock = this.placeNewBlock(originalBlock);
+      return this.placeNewBlock(originalBlock);
     } finally {
-      eventUtils.enable();
-    }
+      this.targetWorkspace.hideChaff();
 
-    // Close the flyout.
-    this.targetWorkspace.hideChaff();
-
-    const newVariables = Variables.getAddedVariables(
-      this.targetWorkspace,
-      variablesBeforeCreation,
-    );
-
-    if (eventUtils.isEnabled()) {
-      eventUtils.setGroup(true);
-      // Fire a VarCreate event for each (if any) new variable created.
-      for (let i = 0; i < newVariables.length; i++) {
-        const thisVariable = newVariables[i];
-        eventUtils.fire(
-          new (eventUtils.get(EventType.VAR_CREATE))(thisVariable),
-        );
+      // Close the flyout.
+      if (this.autoClose) {
+        this.hide();
       }
-
-      // Block events come after var events, in case they refer to newly created
-      // variables.
-      eventUtils.fire(new (eventUtils.get(EventType.BLOCK_CREATE))(newBlock));
     }
-    if (this.autoClose) {
-      this.hide();
-    }
-    return newBlock;
   }
 
   /**
@@ -890,7 +861,7 @@ export abstract class Flyout
     const json = this.serializeBlock(oldBlock);
     // Normally this resizes leading to weird jumps. Save it for terminateDrag.
     targetWorkspace.setResizesEnabled(false);
-    const block = blocks.append(json, targetWorkspace) as BlockSvg;
+    const block = blocks.append(json, targetWorkspace, {recordUndo: true}) as BlockSvg;
 
     this.positionNewBlock(oldBlock, block);
 
--- a/packages/blockly/core/flyout_base.ts
+++ b/packages/blockly/core/flyout_base.ts
@@ -812,16 +812,24 @@ export abstract class Flyout
    * @internal
    */
   createBlock(originalBlock: BlockSvg): BlockSvg {
-    try {
-      return this.placeNewBlock(originalBlock);
-    } finally {
-      this.targetWorkspace.hideChaff();
-
-      // Close the flyout.
-      if (this.autoClose) {
-        this.hide();
-      }
+    const targetWorkspace = this.targetWorkspace;
+    targetWorkspace.hideChaff();
+    const svgRootOld = originalBlock.getSvgRoot();
+    if (!svgRootOld) {
+      throw Error('oldBlock is not rendered');
     }
+
+    // Clone the block.
+    const json = this.serializeBlock(originalBlock);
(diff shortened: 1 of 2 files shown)


packages/blockly/core/flyout_base.ts
```diff
@@ -813,44 +812,16 @@ export abstract class Flyout
...
   createBlock(originalBlock: BlockSvg): BlockSvg {
-    let newBlock = null;
-    eventUtils.disable();
-    const variablesBeforeCreation = this.targetWorkspace
-      .getVariableMap()
-      .getAllVariables();
-    this.targetWorkspace.setResizesEnabled(false);
     try {
-      newBlock = this.placeNewBlock(originalBlock);
+      return this.placeNewBlock(originalBlock);
```
maribethb: If the block is not successfully created then now the flyout is closed, where previously the event utils would be enabled but then the method would crash (without closing flyout)

I think if a block isn't ultimately created for whatever reason (not sure how likely this method is to throw an error in reality) it would be better for the flyout not to close. 

i don't think you need to use a `try` anymore since we don't have to clean up the event utils. if `placeNewBlock` errors then just let it error. so you can hideChaff before you return the new block without relying on the `finally` block.

packages/blockly/core/flyout_base.ts
```diff
@@ -813,44 +812,16 @@ export abstract class Flyout
...
-    if (eventUtils.isEnabled()) {
-      eventUtils.setGroup(true);
-      // Fire a VarCreate event for each (if any) new variable created.
-      for (let i = 0; i < newVariables.length; i++) {
-        const thisVariable = newVariables[i];
-        eventUtils.fire(
-          new (eventUtils.get(EventType.VAR_CREATE))(thisVariable),
-        );
+      // Close the flyout.
+      if (this.autoClose) {
```
maribethb: Doesn't hiding the chaff on the target workspace already close the flyout?

maribethb (approved):

packages/blockly/core/flyout_base.ts
gonfunko: Indeed it does, this was left over from before, but removed accordingly!

packages/blockly/core/flyout_base.ts
gonfunko: Good catch, updated; this left this method as nothing but `hideChaff()` and calling through to `placeNewBlock()`, so I inlined `placeNewBlock()` here as it was private and had no other callsites.

packages/blockly/core/flyout_base.ts
```diff
@@ -812,16 +812,24 @@ export abstract class Flyout
...
-      return this.placeNewBlock(originalBlock);
-    } finally {
-      this.targetWorkspace.hideChaff();
-
-      // Close the flyout.
-      if (this.autoClose) {
-        this.hide();
-      }
+    const targetWorkspace = this.targetWorkspace;
+    targetWorkspace.hideChaff();
```
maribethb: I think you want to do this right before return, so that if there is an error the flyout isn't closed

maribethb (approved):
otherwise lgtm

packages/blockly/core/flyout_base.ts
gonfunko: Done!

PR #9579, opened 2026-01-22
chore: Make JSDom a peer/dev dependency on NodeJS only

### LINKED ISSUE #8775: Remove dependency on `jsdom` (make it a peer dependency instead)

gonfunko:
## The basics

### Resolves

Fixes #8775 

### Proposed Changes
This PR removes JSDom as a normal dependency, and adds the latest version as a peer and dev dependency. It also uses the `browser` field in package.json to indicate to bundlers and the like that it should be omitted for builds targeting a browser, rather than Node-based, deployment environment.

### DIFF
--- a/packages/blockly/package.json
+++ b/packages/blockly/package.json
@@ -123,7 +123,7 @@
     "eslint-plugin-prettier": "^5.2.1",
     "glob": "^11.0.1",
     "globals": "^16.0.0",
-    "google-closure-compiler": "^20260225.0.0",
+    "google-closure-compiler": "^20260315.0.0",
     "gulp": "^5.0.0",
     "gulp-concat": "^2.6.1",
     "gulp-gzip": "^1.4.2",
--- a/packages/blockly/package.json
+++ b/packages/blockly/package.json
@@ -16,6 +16,9 @@
   "author": {
     "name": "Neil Fraser"
   },
+  "browser": {
+    "jsdom": false
+  },
   "scripts": {
     "build": "gulp build",
     "build-debug": "gulp build --verbose --debug",
@@ -133,6 +136,7 @@
     "gulp-sourcemaps": "^3.0.0",
     "gulp-umd": "^2.0.0",
     "http-server": "^14.0.0",
+    "jsdom": "27.4.0",
     "json5": "^2.2.0",
     "markdown-tables-to-json": "^0.1.7",
     "mocha": "^11.3.0",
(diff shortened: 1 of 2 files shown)


packages/blockly/package.json
```diff
@@ -135,6 +138,7 @@
     "gulp-sourcemaps": "^3.0.0",
     "gulp-umd": "^2.0.0",
     "http-server": "^14.0.0",
+    "jsdom": "27.4.0",
```
BenHenning, line 139: Gemini thinks the upgraded version here is likely causing the build failures since this was part of the CI logs:

```
2026-01-22T00:00:56.5787404Z npm warn EBADENGINE Unsupported engine {
2026-01-22T00:00:56.5787981Z npm warn EBADENGINE   package: 'jsdom@27.4.0',
2026-01-22T00:00:56.5788484Z npm warn EBADENGINE   required: { node: '^20.19.0 || ^22.12.0 || >=24.0.0' },
2026-01-22T00:00:56.5788980Z npm warn EBADENGINE   current: { node: 'v18.20.8', npm: '10.8.2' }
2026-01-22T00:00:56.5789308Z npm warn EBADENGINE }
```

(I was curious but didn't actually want to debug).

Can we avoid the version upgrade for now?

BenHenning (approved):
Thanks @gonfunko! The change looks reasonable, though perhaps we may need to drop the version upgrade for now. Left one comment on that per the failing CI.

Edit: Also, it's very cool that the new `package.json` now implies Blockly has zero dependencies (though it's hard to be sure without scanning the package lock file).

packages/blockly/package.json
gonfunko: I think we actually just want to drop the old Node versions for v13, but we should have a team chat about how aggressive to be about that.

BenHenning: I think this makes a lot of sense. It seems completely reasonable to drop everything but the LTS versions at the time of a release cut, but I'm not sure how popular that would be. As of now that means dropping 18, 19, and 20 (since it will be EOL before v13 is cut). I think one good argument is that patch releases become harder when supporting deprecated Node versions.

BenHenning:
I assume that the 18.x element should be removed from the test matrix as part of upgrading the min Node version?

gonfunko:
Yes, see https://github.com/RaspberryPiFoundation/blockly/pull/9600.

PR #9580, opened 2026-01-22
refactor: Use `IContextMenu` to dispatch right clicks

### LINKED ISSUE #8842: Update gesture code to check for `IContextMenu`
### Check for duplicates

### Problem

_No response_

### Request

The current [gesture code currently](https://github.com/google/blockly/blob/212f13af083004368abb5c5f4e46da6a41995b70/core/gesture.ts#L725) checks a specific set of objects and calls the `showContextMenu` function for the object if it is what was right-clicked on.

Instead, the gesture code should:

- Check what is currently focused (this assumes that #8840 is complete)
- Check if the currently focused thing implements `IHasContextMenu`
- If it does, call its `showContextMenu` function and stop propagation of the event. If it doesn't, do nothing and don't stop propagation since the click wasn't handled.
- The current calls to `bringBlockToFront` and `hideChaff` should be moved into their respective objects' implementations of `showContextMenu` since the handling depends on that specific object, and there's no reason for the gesture code to know about it.

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #8842

### Proposed Changes
This PR updates `Gesture` to check for conformance to `IContextMenu` when dispatching right clicks, rather than hard coding specific kinds of elements.

Although `Gesture` is still fairly closely tied to specific elements, in general this does open a path for things to just implement `IContextMenu` to get context menu support. Clicks bubble up from e.g. fields to blocks to the workspace, with each item informing Gesture of a click in turn, and the workspace handler is what ultimately triggers the right click handler. Thus, as long as an element (a) gains focus on (right)click, (b) is a child of the workspace, and (c) implements `IContextMenu`, `Gesture` should now trigger display of the context menu when it is right clicked.

While I was here, I also cleaned up the bringing-block-to-front behavior. This was being triggered in various block sub-elements (fields, icons), but because of the bubbling behavior described above, clicks on those will ultimately reach the parent block, so performing bring-to-front when the block becomes the target block covers all these cases, for both left and right clicks.

### DIFF
--- a/packages/blockly/core/gesture.ts
+++ b/packages/blockly/core/gesture.ts
@@ -27,6 +27,7 @@ import * as eventUtils from './events/utils.js';
 import type {Field} from './field.js';
 import {getFocusManager} from './focus_manager.js';
 import type {IBubble} from './interfaces/i_bubble.js';
+import {hasContextMenu} from './interfaces/i_contextmenu.js';
 import {IDraggable, isDraggable} from './interfaces/i_draggable.js';
 import {IDragger} from './interfaces/i_dragger.js';
 import type {IFlyout} from './interfaces/i_flyout.js';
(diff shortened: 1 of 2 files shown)


packages/blockly/core/gesture.ts
```diff
@@ -878,12 +869,6 @@ export class Gesture {
       );
     }
 
-    // Note that the order is important here: bringing a block to the front will
-    // cause it to become focused and showing the field editor will capture
-    // focus ephemerally. It's important to ensure that focus is properly
-    // restored back to the block after field editing has completed.
-    this.bringBlockToFront();
```
BenHenning, line 885: Just to check, did you validate this bit is no longer needed? I.e. that using keyboard nav to end field editing correctly returns focus to the block? (I think that was the case that necessitated this originally).

BenHenning (approved):
Thanks @gonfunko! Just one comment to double check one of the `bringBlockToFront` removals that I don't think is covered by tests, but otherwise this LGTM.

packages/blockly/core/gesture.ts
gonfunko: AFAICT behavior is unchanged; if you focus a field via keyboard, hit enter, change it, and hit enter, the field gets focus when editing is committed. If you click on a field, make changes, and hit enter, the parent block gets focus, which is consistent with the behavior I see on the hosted keyboard experiment demo page.

BenHenning: Thanks for double checking. It's possible other changes made this unnecessary, as well, or maybe it was never necessary at the time. Either way, removing it sounds good.

PR #9582, opened 2026-01-23
fix: Fix bug that caused flyout buttons to retain their active appearance on mobile

### LINKED ISSUE #6029: Flyout buttons are not unhighlighted after being pressed on mobile
**Describe the bug**

If you click a flyout button on mobile, it gets highlighted, and then never gets unhighlighted.

**To Reproduce**

1. Open the playground on mobile
2. Add a variable using the "create variable" button in the variable flyout.
3. Observe how the button stays highlighted even after you've used it.

**Expected behavior**

The button should lose its highlight.

**Additional context**

Usually the button loses its highlight when you stop hovering, but there's no hover on mobile.

gonfunko:
## The basics

### Resolves

Fixes #6029

### Proposed Changes
This PR fixes a bug that caused flyout buttons to remain visually highlighted after being tapped on mobile. This is because mobile browser apply :hover styles on :active, but don't stop applying them until something else gets focus. Now, the :hover style is limited to devices that actually support hovering via a media query, and an :active style has been added for mobile, which automatically gets removed post-tap.

### DIFF
--- a/packages/blockly/core/flyout_button.ts
+++ b/packages/blockly/core/flyout_button.ts
@@ -427,7 +427,13 @@ Css.register(`
   fill: #666;
 }
 
-.blocklyFlyoutButton:hover {
+@media (hover: hover) {
+  .blocklyFlyoutButton:hover {
+    fill: #aaa;
+  }
+}
+
+.blocklyFlyoutButton:active {
   fill: #aaa;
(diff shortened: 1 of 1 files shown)


BenHenning (approved):
Nice fix.

PR #9587, opened 2026-02-12
fix: Fix running browser tests

gonfunko:
## The basics

### Resolves
This PR fixes `npm run test:browser`, which had been failing with `Error: No test files found: "test"`. I'm not entirely clear why, but it seems like Mocha was not running from the path we were `cd`ing to, I assume due to either a version change or the monorepo work. In any case, the browser tests now work correctly.


maribethb (approved):
Checked out the PR and it works for me locally 👍

Issue #8960, opened 2025-05-02
Refactor `WorkspaceSvg.prototype.isDragging` to call `Mover.prototype.isMoving(this)`


cpcallen:
PR #8959 added support for treating keyboard moves as if they were drags in `WorkspaceSvg.prototype.isDragging`.  At present this is done by having a set method, `.setKeyboardMoveInProgress`, that is called from `Mover.prototype.startMove`, `.finishMove` and `.abortMove`, but there is some possibility that the state of `.isKeyboardMoveInProgress` could get out of sync with the `Mover` state.

It would be better to have `.isDragging` call `mover.isMoving(this)` directly, to check whether `this` workspace is one of the keys present in the `Mover`'s `.moves` map.  Unfortunately there is not at present any easy way for a `WorkspaceSvg` to get access to the (should-but-not-currently-guaranteed-to-be singleton) `Mover` instance.

It seems likely that `Mover` will eventually become part of core (and possibly be refactored to be per-Workspace rather than a singleton), at which point this is likely to be much easier.

PR #9593, opened 2026-02-17
feat!: Allow moving blocks, comments and bubbles using the keyboard

gonfunko:
## The basics

### Proposed Changes
This PR backports the support for moving blocks, workspace comments, and bubbles using the keyboard from the [blockly-keyboard-experimentation](https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation) repo to core Blockly. These interactions have been proved out through user testing, and are an important foundational aspect of making Blockly more accessible. Toward this end, the following changes have been made:

* The `IDragger`, `IDraggable` and `IDragStrategy` interfaces have been revised to put less of an emphasis on mouse events and clean up some other issues that using them in practice has revealed. See the Breaking Changes section below for more details.
* A new global singleton `KeyboardMover` class has been introduced. An object of this class is used to coordinate keyboard-driven movement between keyboard shortcuts and the dragging system. Although Blockly is generally moving away from global singletons in favor of per-workspace objects, KeyboardMover is one because (a) only one drag can be in progress at a given time, (b) drags can cross workspaces, for example dragging out of the flyout into the main workspace and (c) this opens the door for cross-workspace drags more broadly, or one toolbox that serves multiple workspaces.
* Keyboard shortcuts for entering and exiting move mode and moving in each cardinal direction have been introduced. These are mapped to M (start move), arrow keys (move in the corresponding direction), Enter (commit an in-progress move) and Escape (revert an in-progress move).
  * For blocks in particular, keyboard-driven moves move them between available connection locations by default. Blocks may be freely moved around the workspace by pressing Command (Apple platforms) or Control (Windows, Linux, ChromeOS, etc) in conjunction with the arrow keys.

### Future Work
At present, it is not possible to insert blocks from the flyout using the keyboard. This is now feasible, but in the interest of avoiding scope creep and preventing this PR from becoming even larger I will include it in a followup PR.

Additionally, the remaining functionality from the blockly-keyboard-experimentation repo still needs to be backported to core.

### Test Coverage
Tests have been added that exercise each kind of built-in movable entity (blocks, workspace comments, and bubbles). Additionally, the test suite for constrained movement of blocks from the blockly-keyboard-experimentation repo has been backported.

### Breaking Changes
If your application has custom code that implements the `IDragger`, `IDraggable`, or `IDragStrategy` interfaces, you may need to make some changes. Specifically:
* `IDragStrategy.startDrag()`, `IDraggable.startDrag()` and `IDragger.onDragStart()` need to return an `IDraggable & ISelectable & IBoundedElement` that the rest of the drag process will act on. Typically, `IDraggable`s should just return themselves. In cases where you want a drag to act on a different element than the one the drag started on, you can return the object you actually want the drag to act on.
* All drag-related methods that previously accepted a `PointerEvent` now need to handle receiving a `PointerEvent | KeyboardEvent`. In keyboard-driven drags, this will be a `KeyboardEvent`.
* Classes that implement `IDragger` now need to implement `IDragger.onDragRevert()`. This method differs from `onDragEnd()` in that it is invoked when a drag is cancelled or reverted by the user; generally, the item being dragged should be moved back to its starting position. If a drag is reverted, `onDragEnd()` will still be invoked after `onDragRevert()` has been called, so common cleanup can be reliably handled there.
* The event arguments to methods in `IDragger` have been made optional, consistent with `IDraggable` and `IDragStrategy`. Your code should not rely on them being present, although they typically will be for both mouse- and keyboard-driven drags.


gonfunko:
This fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/786

gonfunko:
This also fixes #8960

packages/blockly/core/bubbles/mini_workspace_bubble.ts
```diff
@@ -153,11 +154,7 @@ export class MiniWorkspaceBubble extends Bubble {
    * are dealt with by resizing the workspace to show them.
    */
   private bumpBlocksIntoBounds() {
-    if (
-      this.miniWorkspace.isDragging() &&
-      !this.miniWorkspace.keyboardMoveInProgress
-    )
-      return;
+    if (this.miniWorkspace.isDragging() && !KeyboardMover.isMoving()) return;
```
maribethb: this took me a second to make sure i'm parsing this correctly. we don't want to do this for mouse drags but do want to do it for keyboard drags? if so can you add a comment i think to confirm that it's intentional

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -246,25 +267,56 @@ export class BlockDragStrategy implements IDragStrategy {
   }
 
   /** Moves the block and updates any connection previews. */
-  drag(newLoc: Coordinate): void {
-    if (this.block.isShadow()) {
-      this.block.getParent()?.drag(Coordinate.sum(newLoc, this.dragOffset));
-      return;
+  drag(newLoc: Coordinate, e?: PointerEvent | KeyboardEvent): void {
+    this.moveMode =
+      e instanceof KeyboardEvent && !e.altKey
```
maribethb: minor nit here, in the PR description you say it's control. i believe alt is what we want, so don't change the code, but please update the PR description.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -246,25 +267,56 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      this.block.moveDuringDrag(newLoc);
     }
-
-    this.block.moveDuringDrag(newLoc);
     this.updateConnectionPreview(
       this.block,
       Coordinate.difference(newLoc, this.startLoc!),
     );
+
+    // Handle the case when an unconstrained drag found a connection candidate.
```
maribethb: this also handles constrained drags right? this comment is confusing if so.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -246,25 +267,56 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      this.searchNode = neighbour;
+      if (this.moveMode === MoveMode.CONSTRAINED) {
+        // Position the moving block down and slightly to the right of the
+        // target connection.
+        this.block.moveDuringDrag(
+          new Coordinate(neighbour.x + 10, neighbour.y + 10),
+        );
+      }
+    } else {
+      // Handle the case when unconstrained drag was far from any candidate.
```
maribethb: this comment is also confusing because you have a check for a constrained move inside this block.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -125,6 +132,34 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      this.connectionCandidate = this.createInitialCandidate();
+      const neighbour = this.updateConnectionPreview(
+        this.block,
+        new Coordinate(0, 0),
+      );
+      if (neighbour) {
+        let offset: Coordinate;
+        if (neighbour.type === ConnectionType.PREVIOUS_STATEMENT) {
+          const origin = this.block.getRelativeToSurfaceXY();
+          offset = new Coordinate(origin.x + 10, origin.y - 10);
```
maribethb: I think the 10 should be a protected constant in this class so it's changeable, and because you use it elsewhere as well (in the candidate connection logic of `drag`)

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -125,6 +132,34 @@ export class BlockDragStrategy implements IDragStrategy {
     this.getVisibleBubbles(this.block).forEach((bubble) => {
       this.workspace.getLayerManager()?.moveToDragLayer(bubble, false);
     });
+
+    // For keyboard-driven moves, cache a list of valid connection points for
+    // use in constrained moved mode.
+    if (e instanceof KeyboardEvent) {
+      for (const topBlock of this.block.workspace.getTopBlocks(true)) {
+        this.allConnections.push(...this.getAllConnections(topBlock));
```
maribethb, line 142: I wonder if it makes sense to do this at the start of any move. I'm envisioning someone doing constrained mouse moves in the future (could be a cool touch interface where you can only drag blocks to valid connections maybe with the reindeer connection though i have no idea how that would work in practice). But I guess we could add that if anyone asks for it. Might be expensive to add prematurely.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -520,4 +595,172 @@ export class BlockDragStrategy implements IDragStrategy {
...
+   * Create a candidate representing where the block was previously connected.
+   * Used to render the block position after picking up the block but before
+   * moving during a drag.
+   *
+   * @returns A connection candidate representing where the block was at the
+   *     start of the drag.
+   */
+  private createInitialCandidate(): ConnectionCandidate | null {
+    this.searchNode =
+      /* this.startPoint ?? */ this.startParentConn ?? this.startChildConn;
```
maribethb: nit: stray comment I think?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -0,0 +1,314 @@
...
+import {ShortcutRegistry} from '../shortcut_registry.js';
+import {Coordinate} from '../utils/coordinate.js';
+import {KeyCodes} from '../utils/keycodes.js';
+import {MoveIndicator} from './move_indicator.js';
+
+/**
+ * Cardinal directions in which a move can proceed.
+ */
+export enum Direction {
+  NONE = 0,
```
maribethb: why does only NONE have a value? is this a typescript magic i don't know?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Identifier for a keyboard shortcut that commits the in-progress move.
+ */
+const COMMIT_MOVE_SHORTCUT = 'commitMove';
+
+/**
+ * Class responsible for coordinating keyboard-driven moves with the workspace
+ * and dragging system.
+ */
+class KeyboardMoverImplementations {
```
maribethb: why `Implementations` and not just `KeyboardMover`?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -0,0 +1,314 @@
...
+  /**
+   * Object responsible for dragging workspace elements in response to move
+   * commands.
+   */
+  protected dragger?: IDragger;
+
+  /**
+   * The object that is currently being moved.
+   */
+  protected draggable?: IDraggable & IBoundedElement & ISelectable;
```
maribethb: the fact that you've repeated this a couple places makes me wonder if the IDraggable interface should just be updated to include IBoundedElement and ISelectable so you can keep saying just IDraggable. Is there now any use of IDraggable that doesn't have to be these other two as well?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -0,0 +1,314 @@
...
+    // keyboard shortcuts that commits the move before allowing the real
+    // shortcut to proceed. This avoids all kinds of fun brokenness when
+    // deleting/copying/otherwise acting on a element in move mode.
+    const shortcutKeys = Object.values(ShortcutRegistry.registry.getRegistry())
+      .flatMap((shortcut) => shortcut.keyCodes)
+      .filter((keyCode) => {
+        return (
+          keyCode &&
+          ![
+            KeyCodes.RIGHT,
```
maribethb, line 136: I think this it potentially worth rethinking to make it more flexible. I can see people wanting to add shortcuts that work in move mode. For example the `i` shortcut should possibly work in move mode.

An easy change would be to move this list of "shortcuts that are allowed to work in move mode" to a protected class member as well, so that it can be customized. But it's a little odd that it works based off the keycode and not the name of the shortcut.

packages/blockly/core/hints.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param force Set to show it even if previously shown.
+ */
+export function showUnconstrainedMoveHint(
+  workspace: WorkspaceSvg,
+  force = false,
+) {
+  const modifier =
+    userAgent.MAC || userAgent.IPAD || userAgent.IPHONE
+      ? Msg['OPTION_KEY']
+      : Msg['CONTROL_KEY'];
```
maribethb, line 28: this should be alt

packages/blockly/core/shortcut_items.ts
```diff
@@ -386,6 +396,144 @@ export function registerRedo() {
   ShortcutRegistry.registry.register(redoShortcut);
 }
 
+/**
+ * Registers keyboard shortcuts for keyboard-driven movement of workspace
+ *elements.
```
maribethb: super duper nit: space after the asterisk (surprised a linter / formatter did not fix this)

packages/blockly/tests/mocha/test_helpers/move_test_blocks.js
```diff
@@ -0,0 +1,528 @@
+/**
```
maribethb, line 1: Consider changing the name of this file since it currently only contains helpers related to move mode tests? there's already other test_helper files that contain similar code so `fixtures` feels a little too generic unless you envision more things going here.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Yeah, it auto-increments from that for the subsequent values. Although it also defaults to 0, so I should just remove this.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Because we only export a single instance of this class under the name `KeyboardMover` to enforce it being a singleton, so the class itself had to have a different name. Dropped the plural though, that was inadvertently included.

packages/blockly/core/shortcut_items.ts
gonfunko: Done

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Removed

packages/blockly/core/keyboard_nav/keyboard_mover.ts
maribethb: What we do in some places is `KeyboardMover` the class and `keyboardMover` the global. An example is `keyboardNavigationController`. Or for `FocusManager` we have `getFocusManager`. It might be better to stay consistent with the lowercase name for the global object, but I could be convinced otherwise if we have other examples of globals being UpperCamel.

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Yes, this and the next one were copypasta from keyboard-experiment. Updated!

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Yes, rewrote + clarified.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
maribethb: makes sense. follow up question, should the DragDisposition enum use the same pattern then or is there a reason that one starts at 1?

packages/blockly/core/bubbles/mini_workspace_bubble.ts
gonfunko: Other way around, but yes – good call on the comment, added!

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Yeah good call, in order to be draggable things kind of have to be selectable and have dimensions/a visual representation, so may as well formalize that especially while we're already making breaking changes. Updated!

packages/blockly/tests/mocha/test_helpers/move_test_blocks.js
gonfunko: Done

packages/blockly/core/dragging/block_drag_strategy.ts

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: I think I'd rather hold off for now in the interest of performance. We certainly can change it later easily enough.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Updated it such that we export `class KeyboardMover` which has a `static mover = new KeyboardMover()` field. This follows the ContextMenuItemRegistry pattern, which I think is the variant of this that I like best. Does that seem reasonable?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Are you thinking that we'd also make this registerable?

packages/blockly/core/keyboard_nav/keyboard_mover.ts
maribethb: yep!

packages/blockly/core/keyboard_nav/keyboard_mover.ts
maribethb: discussed offline, will not make it registrable since it's the analog to Gesture which is also not registrable. You can customize a lot of the behavior in drag strategies. Aaron will add a method to add more keyboard shortcuts to the allow list so that keyboard shortcuts can exempt themselves from this behavior.

packages/blockly/core/hints.ts
maribethb: discussed offline, it can't be alt on windows, so it's control. but there's no reason it can't be command on mac, so we update it to command on mac and ctrl on windows so it's consistent. there weren't any other places we were using alt anyway, so this makes more sense. if this poses an issue to anyone please raise it in an issue

maribethb (commented):
everything so far lgtm, will review again for the changes we discussed offline

packages/blockly/core/dragging/block_drag_strategy.ts

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Done; added methods to get and set the list of allowlisted shortcuts, and a test. Also switched everything that was `protected` to `private`, since there's no useful ability to subclass this if we're not making it registerable.

maribethb (approved):

PR #9598, opened 2026-02-24
fix: Remove unsafe non-null assertions

### LINKED ISSUE #9597: Uncaught exception in toolbox onClick_
### Check for duplicates

### Description

I have a plugin which adds a sub-element into the toolbox which contains an id. When clicking on this element, `getToolboxItemById` returns null but throws `Cannot read properties of null (reading 'isSelectable')`.

### Reproduction steps

1. Create element in toolbox which has an id (or call onClick_ on element that isn't in the toolbox)
2. Error is thrown

### Priority

_No response_

### Stack trace

```shell
TypeError: Cannot read propertyies of null (reading 'isSelectable')
```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9597

### Proposed Changes
This PR removes a variety of unsafe non-null assertions in the toolbox.

### DIFF
--- a/packages/blockly/core/toolbox/toolbox.ts
+++ b/packages/blockly/core/toolbox/toolbox.ts
@@ -11,7 +11,6 @@
  */
 // Former goog.module ID: Blockly.Toolbox
 
-// Unused import preserved for side-effects. Remove if unneeded.
 import {BlockSvg} from '../block_svg.js';
 import * as browserEvents from '../browser_events.js';
 import * as common from '../common.js';
@@ -192,7 +191,7 @@ export class Toolbox
     aria.setRole(this.contentsDiv_, aria.Role.TREE);
     container.appendChild(this.contentsDiv_);
 
-    svg.parentNode!.insertBefore(container, svg);
+    svg.parentNode?.insertBefore(container, svg);
 
     this.attachEvents_(container, this.contentsDiv_);
     return container;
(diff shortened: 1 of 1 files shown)


BenHenning (commented):
So I think a lot of the changes make sense because they're simplifying existing states. However I'm less certain about some of the other switches to `?.` syntax. Some of these seem like cases where the value should absolutely be present. I'm wondering if all of these cases are actually understood as to why the value can be missing. If so, it's definitely worth adding tests and comments to explain those. Otherwise I feel like the forced un-nulling is actually a useful mechanism to catch incorrect developer assumptions (which then leads to overall safety, not just potentially silencing the failure).

What are your thoughts @gonfunko?

gonfunko:
The main issue here (and throughout Blockly in general) is that lots of things are not actually initialized in the constructor, but rather in some init/createDom method. That means they have to be treated as nullable, but de facto aren't. That's the case here as well; `flyout`, `HtmlDiv`, and `contentsDiv_` are all initialized by `init()`, which is called from `Blockly.inject()`. Being called directly from `inject()` is about as close to a guarantee of "this will always happen" as we can get, and I think just having ?. to satisfy TS is adequate.

The bug that led to this PR is a bit different in that we do allow adding arbitrary items to the toolbox. We're already only selecting them if they're selectable, so I think only doing that in the case that we were able to successfully look up the item in the first place is fine, and certainly preferable to throwing.

BenHenning (approved):
That context is helpful @gonfunko. I think it's probably worth adding tests for these cases, though, because that's a way to explicitly say "we support these use cases" (where the cases include ones when the toolbox is being correctly interacted with before fully initialization). I won't block my review on that, though, but it's worth considering to strengthen Toolbox's contract.

PR #9599, opened 2026-02-24
fix: Focus nearest neighbor when deleting a focused block

### LINKED ISSUE #9585: Deleting a top-level selected block scrolls the workspace to the top left
### Check for duplicates

### Description

When you delete the selected block and there is no containing parent block for the focus manager to select (ie the block is top-level), the manager defaults to focusing whatever block is in the top-left of the workspace which scrolls the workspace to top left.

This behavior is very annoying for large workspaces because it means constantly having to re-scroll to wherever you are in your project whenever you delete something. For example, if you want to delete three blocks that are top level in the bottom right of your workspace, you're going to have to scroll all the way from the top left to the bottom right three times.

The expected behavior imo would be to focus whatever node is closest to the block's original location before being dragged/deleted.

### Reproduction steps

1. Open the Blockly playground
2. Place a block in the top-left of the workspace and scroll to the right so that the block is no longer visible
3. Place a few other blocks on the workspace
4. Delete one of these blocks
5. Watch as the workspace is scrolled away from where you were working

See screenshots below for a recording of me reproducing this issue

### Priority

This is primarily an issue for our MakeCode Arcade editor, where large blocks projects are very common. I expect this will affect most of our users, so we consider this a high priority issue.

I haven't worked in the focus manager code before, so I can't speak to the effort involved but the solution at least seems clear.

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9585

### Proposed Changes
This PR adjusts the focus behavior when a focused block is deleted. Previously, deleting a focused block would focus the workspace, which would in turn focus the first (left/topmost)  block on the workspace. This could cause large jumps in scroll position. Now, when a focused block is deleted, it moves focus to the closest remaining block, resulting in a smaller (or no) workspace movement.

The tests were partially LLM-generated in response to a list of testcases I specified. I reviewed, modified them, and manually added an additional testcase.

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -932,9 +932,12 @@ export class BlockSvg
       } else {
         const nearestNeighbour = this.getNearestNeighbour();
         if (nearestNeighbour) {
-          setTimeout(() => focusManager.focusNode(nearestNeighbour), 0);
+          focusManager.focusNode(nearestNeighbour);
         } else {
-          setTimeout(() => focusManager.focusTree(this.workspace), 0);
+          setTimeout(() => {
+            if (!this.workspace.rendered) return;
+            focusManager.focusTree(this.workspace);
+          }, 0);
         }
       }
     }
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -864,6 +864,32 @@ export class BlockSvg
   }
 
   /**
+   * Returns the closest live block to this one, if any.
+   */
+  private getNearestNeighbour() {
+    if (!this.workspace.rendered) return null;
+
+    const blocks = this.workspace
+      .getAllBlocks(false)
+      .filter((block) => !block.isDeadOrDying());
+    let nearestNeighbour = null;
+    let closestDistance = Number.MAX_SAFE_INTEGER;
+    const self = this.getRelativeToSurfaceXY();
+    for (const block of blocks) {
+      const other = block.getRelativeToSurfaceXY();
+      const distance = Math.sqrt(
+        Math.pow(other.x - self.x, 2) + Math.pow(other.y - self.y, 2),
+      );
+      if (distance < closestDistance) {
+        nearestNeighbour = block;
+        closestDistance = distance;
+      }
+    }
+
+    return nearestNeighbour;
+  }
+
+  /**
    * Dispose of this block.
    *
    * @param healStack If true, then try to heal any gap by connecting the next
@@ -904,7 +930,12 @@ export class BlockSvg
       if (parent) {
         focusManager.focusNode(parent);
       } else {
-        setTimeout(() => focusManager.focusTree(this.workspace), 0);
+        const nearestNeighbour = this.getNearestNeighbour();
+        if (nearestNeighbour) {
+          setTimeout(() => focusManager.focusNode(nearestNeighbour), 0);
+        } else {
+          setTimeout(() => focusManager.focusTree(this.workspace), 0);
+        }
       }
     }
 
--- a/packages/blockly/tests/mocha/block_test.js
+++ b/packages/blockly/tests/mocha/block_test.js
@@ -3050,5 +3050,13 @@ suite('Blocks', function () {
         'Focus should not change when an unfocused block is deleted',
       );
     });
+
+    test('Disposing a workspace with a focused block succeeds', function () {
+      Blockly.getFocusManager().focusNode(this.workspace.getTopBlocks(false)[0]);
+      this.workspace.dispose();
+      this.clock.runAll();
+
+      // No assert, this just shouldn't throw.
+    });
   });
 });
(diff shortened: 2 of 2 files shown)


packages/blockly/core/block_svg.ts
```diff
@@ -904,7 +930,12 @@ export class BlockSvg
       if (parent) {
         focusManager.focusNode(parent);
       } else {
-        setTimeout(() => focusManager.focusTree(this.workspace), 0);
+        const nearestNeighbour = this.getNearestNeighbour();
+        if (nearestNeighbour) {
+          setTimeout(() => focusManager.focusNode(nearestNeighbour), 0);
```
BenHenning: Is there any case where there may be a sequence of deletions such that this triggers an attempt to focus the next closest node before it's deleted? Might be worth adding a test case to see if that's actually possible and, if so, make this a bit more robust against that.

packages/blockly/tests/mocha/block_test.js
```diff
@@ -2946,4 +2946,94 @@ suite('Blocks', function () {
...
+      this.workspace = Blockly.inject('blocklyDiv');
+      const firstBlock = this.workspace.newBlock('stack_block');
+      firstBlock.moveBy(-500, -500);
+    });
+
+    test('Deleting the sole block on the workspace focuses the workspace', function () {
+      const block = this.workspace.getTopBlocks(false)[0];
+      Blockly.getFocusManager().focusNode(block);
+      block.dispose();
+      this.clock.runAll();
```
BenHenning, line 2961: Is this actually needed here & below?

packages/blockly/tests/mocha/block_test.js
```diff
@@ -2946,4 +2946,94 @@ suite('Blocks', function () {
...
+      );
+    });
+
+    test('Bulk deleting blocks does not focus another dying block', function () {
+      for (let i = 0; i < 50; i++) {
+        this.workspace.newBlock('stack_block');
+      }
+
+      Blockly.getFocusManager().focusNode(
+        this.workspace.getTopBlocks(false)[0],
```
BenHenning: This relates to my other comment: perhaps double check that the nearest neighbor is still actually doing what it seems like and selecting and trying to focus a block before it's disposed. It may be the case that something else (maybe the focus system or another part of disposal) is robust against this failure case.

packages/blockly/tests/mocha/block_test.js
```diff
@@ -2946,4 +2946,94 @@ suite('Blocks', function () {
...
+      const focused = Blockly.getFocusManager().getFocusedNode();
+      assert.equal(
+        focused,
+        blockRight,
+        'Focus should move to the closest remaining block (blockRight at (100, 0))',
+      );
+    });
+
+    test('Bulk deleting blocks does not focus another dying block', function () {
+      for (let i = 0; i < 50; i++) {
```
BenHenning: Idea can probably be tested with fewer blocks (for perforamnce). Maybe just have something like 5?

packages/blockly/tests/mocha/block_test.js
```diff
@@ -2946,4 +2946,94 @@ suite('Blocks', function () {
...
+    test('Deleting an unfocused block does not change focus', function () {
+      const a = this.workspace.getTopBlocks(false)[0];
+      const b = this.workspace.newBlock('stack_block');
+      this.workspace.newBlock('stack_block');
+
+      Blockly.getFocusManager().focusNode(a);
+      b.dispose();
+      this.clock.runAll();
+
+      assert.equal(
```
BenHenning: Here & elsewhere: I suggest using `strictEqual` instead since there can be problems with the deep equal for connected blocks (circular references are possible), and it shouldn't actually be necessary for verifying focus.

BenHenning (changes requested):
Thanks @gonfunko! Largely LGTM but had a few thoughts--PTAL.

packages/blockly/core/block_svg.ts
gonfunko: That is a possibility, and is why `getNearestNeighbour()` filters the blocks by `isDeadOrDying()`, so that it only returns a block that isn't about to be disposed of. This is also the purpose of the test that makes 50 blocks and clears the workspace.

packages/blockly/tests/mocha/block_test.js
gonfunko: Yes, this test file mocks the clock and this gets the various async focus things to happen.

packages/blockly/tests/mocha/block_test.js
gonfunko: Oh nice, that's a big improvement. We should probably take a pass to update all the tests to do this where relevant to avoid locking things up when there's a failure.

packages/blockly/tests/mocha/block_test.js
gonfunko: Done

packages/blockly/tests/mocha/block_test.js
gonfunko: Adjusted this is a bit to verify that it does not focus dying blocks.

gonfunko:
This also fixes #9601

BenHenning (approved):
Thanks @gonfunko!

PR #9600, opened 2026-02-24
chore!: Update supported Node versions

gonfunko:
## The basics

### Proposed Changes
This PR updates the Node version support matrix for v13 to be Node 22 and 24. Node 18 is EOL and Node 20 will be EOL in April, which is likely before Blockly v13 will be released. We may want to add Node 26 depending on when the release occurs, but it's not yet available. Although Blockly itself doesn't need this, it seems the general approach in the ecosystem is to keep these within the LTS range, and this will allow us to update more dependencies that we'd otherwise be stuck on old versions of because they have dropped support for Node versions we're still treating as supported.


.github/workflows/browser_test.yml
```diff
@@ -6,7 +6,7 @@ name: Run browser manually
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 6 * * 1' # Runs every Monday at 06:00 UTC
+    - cron: "0 6 * * 1" # Runs every Monday at 06:00 UTC
```
maribethb: any particular reason for this change? we use single quotes elsewhere in this file and generally

maribethb (approved):

.github/workflows/browser_test.yml
gonfunko: Autoformatter did it 🤷‍♂️ Reverted though.

PR #9603, opened 2026-02-26
fix: Fix a bug that could cause menus to scroll incorrectly on mouseover

gonfunko:
## The basics

### Proposed Changes
This PR resolves an issue that caused lengthy dropdown menus to jump/scroll and momentarily display the wrong item as selected on mouseover. Scratch reported this in the variable dropdown, but it affected core as well.

### DIFF
--- a/packages/blockly/core/css.ts
+++ b/packages/blockly/core/css.ts
@@ -456,6 +456,7 @@ input[type=number] {
   height: 16px;
   position: absolute;
   width: 16px;
+  display: none;
 }
 
 .blocklyMenuItemSelected .blocklyMenuItemCheckbox {
@@ -463,6 +464,7 @@ input[type=number] {
   float: left;
   margin-left: -24px;
   position: static;  /* Scroll with the menu. */
+  display: block;
 }
 
 .blocklyMenuItemRtl .blocklyMenuItemCheckbox {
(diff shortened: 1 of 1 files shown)


packages/blockly/core/css.ts
```diff
@@ -456,13 +456,15 @@ input[type=number] {
   height: 16px;
   position: absolute;
   width: 16px;
+  display: none;
```
BenHenning: Why is `none` needed here? That's surprising.

gonfunko: The checkboxes are always present in the DOM, but are only shown when `.blocklyMenuItemSelected` is applied below. Since they're absolutely positioned here, they're laid out atop the menu items like so:

Image

That means that when the mouse moves over one, the menu attempts to resolve which item it's a child of, but their position doesn't correspond to their corresponding menu items, so the one that's being hovered over might actually be a child of an earlier menu item than where the mouse is, hence the jank. I just deleted this entire rule and moved height/width into the selected case below to make it clearer; now they're still always present in the DOM, but have no bounds (and hence won't be mouse move targets, or be absolutely positioned in such a way that they aren't children of the menu item they're displayed above) which still solves the problem and is hopefully clearer.

BenHenning (approved):

BenHenning:
Thanks @gonfunko!

PR #9607, opened 2026-03-02
chore(deps): Update Closure compiler to the latest version

gonfunko:
## The basics

### Resolves
This PR updates the version of the Closure compiler in use on the v13 branch to the latest, hopefully fixing CI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -739,12 +739,14 @@
       }
     },
     "node_modules/google-closure-compiler": {
-      "version": "20260114.0.0",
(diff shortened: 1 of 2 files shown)


maribethb (approved):

maribethb:
I wonder if it has to do with the node versions supported by the older package. the previous version in package.json claims it works with node ">=18" so there's no upper bound set, but perhaps something else under the hood caused it to not be compatible. that's the only relevant difference I can think of with v13. hopefully this fixes it Once And For All (TM)

PR #9610, opened 2026-03-03
refactor!: Move responsibility for block creation out of flyouts

gonfunko:
## The basics

### Proposed Changes
This PR moves block creation out of the flyout and into the `BlockDragStrategy`. This simplifies the code overall and results in a better separation of concerns: flyouts are capable of holding arbitrary kinds of objects now, so having them tightly coupled to blocks is suboptimal. Now, duplicating blocks when dragged out of the flyout is handled by the `BlockDragStrategy`, and the flyout is concerned only with the display and layout of its contents, whatever they may be. As a result of this change, it's now possible to copy blocks out of the flyout using the keyboard alone by selecting a block in the flyout and hitting `m` to enter move mode, which copies the block to the main workspace and puts it in move mode, exactly the same as a mouse-driven drag would.

### Breaking Changes
* `createBlock()`, `isDragTowardWorkspace()`, and `isBlockCreatable()` have been removed from `IFlyout`. While this is not breaking from a building standpoint, it is behaviorally breaking if you have an alternate flyout implementation that customized those methods. `isDragTowardWorkspace()` was already a no-op, so no action needs to be taken other than removing it. Customizations that were made to `createBlock()` and `isBlockCreatable()` can be achieved by creating a `BlockDragStrategy` subclass that overrides `isMovable()` and/or `swapTargetBlock()` and using a mixin or extension to apply that drag strategy to your blocks.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -123,9 +123,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   /**
    * Returns a block to use for the current drag operation in place of the
-   * current block, or undefined if the current block should be used.
+   * current block, or undefined if the current block should be used. This may
+   * create and return a newly instantiated block when e.g. dragging from a
+   * flyout.
    */
-  protected swapTargetBlock() {
+  protected getTargetBlock() {
     if (this.block.isShadow()) {
       const parent = this.block.getParent();
       if (parent) {
@@ -157,9 +159,9 @@ export class BlockDragStrategy implements IDragStrategy {
    * from any parent blocks.
    */
   startDrag(e?: PointerEvent | KeyboardEvent) {
-    const alterateTarget = this.swapTargetBlock();
-    if (alterateTarget) {
-      return alterateTarget.startDrag(e);
+    const alternateTarget = this.getTargetBlock();
+    if (alternateTarget) {
+      return alternateTarget.startDrag(e);
     }
 
     this.dragging = true;
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -122,10 +122,8 @@ export class BlockDragStrategy implements IDragStrategy {
   }
 
   /**
-   * Returns a block to use for the current drag operation in place of the
-   * current block, or undefined if the current block should be used. This may
-   * create and return a newly instantiated block when e.g. dragging from a
-   * flyout.
+   * Returns the block to use for the current drag operation. This may create
+   * and return a newly instantiated block when e.g. dragging from a flyout.
    */
   protected getTargetBlock() {
     if (this.block.isShadow()) {
@@ -153,7 +151,7 @@ export class BlockDragStrategy implements IDragStrategy {
       }
     }
 
-    return null;
+    return this.block;
   }
 
   /**
@@ -162,7 +160,7 @@ export class BlockDragStrategy implements IDragStrategy {
    */
   startDrag(e?: PointerEvent | KeyboardEvent) {
     const alternateTarget = this.getTargetBlock();
-    if (alternateTarget) {
+    if (alternateTarget !== this.block) {
       return alternateTarget.startDrag(e);
     }
 
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -25,8 +25,10 @@ import * as layers from '../layers.js';
 import * as registry from '../registry.js';
 import {finishQueuedRenders} from '../render_management.js';
 import type {RenderedConnection} from '../rendered_connection.js';
+import * as blocks from '../serialization/blocks.js';
 import {Coordinate} from '../utils.js';
 import * as dom from '../utils/dom.js';
+import * as svgMath from '../utils/svg_math.js';
 import type {WorkspaceSvg} from '../workspace_svg.js';
 
 /** Represents a nearby valid connection. */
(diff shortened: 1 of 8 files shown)


gonfunko:
This fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/679

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -95,17 +97,71 @@ export class BlockDragStrategy implements IDragStrategy {
...
+
+    return null;
+  }
+
   /**
    * Handles any setup for starting the drag, including disconnecting the block
    * from any parent blocks.
    */
   startDrag(e?: PointerEvent | KeyboardEvent) {
+    const alterateTarget = this.swapTargetBlock();
```
maribethb: typo: alternate (if this stays)

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -95,17 +97,71 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      screenCoordinate,
+    );
+    newBlock.moveTo(workspaceCoordinates);
+  }
+
+  /**
+   * Returns a block to use for the current drag operation in place of the
+   * current block, or undefined if the current block should be used.
+   */
+  protected swapTargetBlock() {
```
maribethb: we discussed offline why this is the way it is:

`startDrag` on the dragger returns the thing being dragged, to allow for cases just like this. since the block drag strategy is the thing that ultimately decides which thing is being dragged, it has to have a method for doing that logic.

we decided to call it `getTargetBlock` (or something similar) which feels a little simpler and more generic. It might be a little unexpected that a `get` method can create a new block (as it does in the flyout case) so should be explained in the tsdoc.

packages/blockly/core/gesture.ts
```diff
@@ -896,17 +855,11 @@ export class Gesture {
...
-        }
-        const newBlock = this.flyout.createBlock(this.targetBlock);
-        newBlock.snapToGrid();
-        newBlock.bumpNeighbours();
-
-        // If a new block was added, make sure that it's correctly focused.
-        getFocusManager().focusNode(newBlock);
-      }
+
+      // Perform a zero-length drag to copy the block out of the flyout.
```
maribethb: It's really weird that Gesture does this I think. A couple of thoughts:

- I know some people were doing custom block creation shenanigans where they would have a block look different in the flyout than it did in the main workspace, to make the flyout look simpler. I don't know how this works, are they doing it with fancy block definitions (that change based on whether the block is in a flyout) or were they doing it with fancy flyouts that override `flyout.createBlock`? If the latter, now they'll have to override the drag strategy instead which is maybe unexpected. Is there a place, maybe on the block class which can be customized with a mixin, where we can have the logic for "create a new block of this type from the flyout"?
- If we did have that, it could also be called from the keyboard shortcut that handles pressing "enter" on a block in the flyout (which currently calls `flyout.createBlock` as well)
- otherwise, the Enter shortcut is also going to have to create one of these fake drags which is also undesirable.
- Then, as to the fact that gesture is for some reason handling creating a block on click, maybe blocks should just do this themselves in response to a block click? if they're on a flyout and get a click event, then they could clone themselves onto the main workspace without involving the dragger at all? I know you're just recreating the logic that was here but the old logic was weird too.

Let me know your thoughts here!

packages/blockly/tests/mocha/gesture_test.js
```diff
@@ -102,9 +102,10 @@ suite('Gesture', function () {
 
   test('Clicking on shadow block does not select it', function () {
     const flyout = this.workspace.getFlyout(true);
-    flyout.createBlock(
+    Blockly.KeyboardMover.mover.startMove(
```
maribethb: Is this just a convoluted way to get a block with a shadow block onto the workspace? why not do that via serializing a block instead?

packages/blockly/core/gesture.ts
gonfunko: What if we leaned on the serialization system? Here (and on enter) we could serialize the block, then deserialize that data on the main workspace. For the case where people want to do block shenanigans, they could just define a `saveExtraState()` that adds a "savedFromFlyout" key, and check for that in `loadExtraState()` (and that the block's workspace isn't presently a flyout) and perform their shenanigans. Or maybe just check whether the workspace is or isn't a flyout in `loadExtraState()` and not bother with the key in the first place if they just have flyout vs non-flyout version. That will also Just Work for dragging/moving out of the flyout, since `getTargetBlock()` does its thing via serialization on the flyout and deserialization on the main workspace.

gonfunko: I also agree that it's weird that `Gesture` handles block clicks, but it seems like `Gesture` handling all clicks is how things work in general, so I'd be hesitant to change it for just this one case.

packages/blockly/tests/mocha/gesture_test.js
gonfunko: Yep, this was the closest parallel to what it was doing before, but updated to use serialization.

packages/blockly/core/dragging/block_drag_strategy.ts

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Renamed accordingly

packages/blockly/core/gesture.ts
maribethb: While Gesture does handle all the clicks, it usually does so by just calling a 'handleClick' method on the thing being clicked (or something similar), the in-depth block creation logic including checking if the block is allowed to be created, etc. is what seems out of place. 

The serialization system does seem like the correct place to handle that kind of block creation shenanigans, you're right.

So I guess my question is where does it make sense to call the block serialization system? You already do it in the dragger code, does it make more sense to replicate that both here and in the enter shortcut? Or have gesture and enter shortcut create a fake drag? Or have one method that all 3 of those places can call?

gonfunko: Fair, in general I'd kind of like to introduce an interface along the lines of "handleClick" since it'd be very helpful to have for pressing Enter in keyboard nav mode without having to do a bunch of instanceof checks. For now, WDYT about switching to using serialization here, but refactoring where click handling happens (and adding support for the Enter shortcut) in its own PR, perhaps with the new interface in that as well?

maribethb: Sounds reasonable!

gonfunko: Done

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -123,9 +123,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   /**
    * Returns a block to use for the current drag operation in place of the
-   * current block, or undefined if the current block should be used.
+   * current block, or undefined if the current block should be used. This may
```
maribethb: I do think for clarity that it would be better if the method returned the current block instead of undefined, and then you check if (this.block === this.getTargetBlock())

maribethb (approved):

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Oh right, I forgot we'd talked about that. Updated!

PR #9611, opened 2026-03-03
feat!: Allow using Blockly in web components/shadow DOM

### LINKED ISSUE #1114: shadow dom causes "Uncaught Error: container is not in current document."

gonfunko:
## The basics

### Resolves

Fixes #1114

### Proposed Changes
This PR adds support for using Blockly inside of the shadow DOM and web components.

Largely, this is a matter of injecting CSS into the document or shadow DOM root as appropriate, rather than creating `<style>` tags and injecting them into `<head>`. Additionally, the global floaty elements (tooltips, input fields, widget div, dropdown div) needed some adjustments to their positioning logic; they had generally all been assuming that they lived in a global div at the root of the document, but this was already not necessarily the case. We've allowed specifying a parent element for these, but if that had bounds smaller than the document things could get wonky. Now, the parent container defaults to returning the injection div unless another value is specified, so these elements are all contained within that (and the shadow DOM/web component, if Blockly is used in one), and that element's position in the page is taken into account when determining the bounds for tooltips, dropdowns, etc.

I also added an additional playground/test harness (with LLM assistance) that embeds Blockly inside a `<div>` as usual, but also defines a web component and uses that to embed a separate instance on the same page.

### Reason for Changes
This has been a long-standing request, and web components have become more widespread and popular. We also have some partner applications that could benefit from this.

### Breaking Changes
* `getParentContainer()` returns the injection div if another element has not been specified. This behavior is likely fine, but if you were using this method you should either explicitly set the container you want or verify that your use is compatible with the injection div.
* `Blockly.Css.inject()` requires that the element into which Blockly/the CSS should be injected be specified.
* The `tagName` argument has been removed from `ConstantProvider.createDom()` and `ConstantProvider.injectCss_()`, since CSS is no longer added via a `<style>` tag. If you have a custom renderer or `ConstantProvider` or call these methods, ensure that the `tagName` argument is removed from your implementation/calls.

### DIFF
--- a/packages/blockly/core/css.ts
+++ b/packages/blockly/core/css.ts
@@ -19,7 +19,7 @@ export function register(cssContent: string) {
   if (typeof window === 'undefined' || !window.CSSStyleSheet) return;
 
   const sheet = new CSSStyleSheet();
-  sheet.replaceSync(cssContent);
+  sheet.replace(cssContent);
   registeredStyleSheets.push(sheet);
 }
 
@@ -40,7 +40,7 @@ export function inject(
   hasCss: boolean,
   pathToMedia: string,
 ) {
-  if (!hasCss) {
+  if (!hasCss || typeof window === 'undefined' || !window.CSSStyleSheet) {
     return;
   }
 
@@ -54,7 +54,7 @@ export function inject(
   const cssContent = content.replace(/<<<PATH>>>/g, mediaPath);
 
   const sheet = new CSSStyleSheet();
-  sheet.replaceSync(cssContent);
+  sheet.replace(cssContent);
   root.adoptedStyleSheets.push(sheet);
 
   registeredStyleSheets.forEach((sheet) => root.adoptedStyleSheets.push(sheet));
--- a/packages/blockly/core/css.ts
+++ b/packages/blockly/core/css.ts
@@ -6,7 +6,8 @@
 
 // Former goog.module ID: Blockly.Css
 /** Has CSS already been injected? */
-let injected = false;
+const injectionSites = new WeakSet<Document | ShadowRoot>();
+const registeredStyleSheets: Array<CSSStyleSheet> = [];
 
 /**
  * Add some CSS to the blob that will be injected later.  Allows optional
@@ -15,10 +16,11 @@ let injected = false;
  * @param cssContent Multiline CSS string or an array of single lines of CSS.
  */
 export function register(cssContent: string) {
-  if (injected) {
-    throw Error('CSS already injected');
-  }
-  content += '\n' + cssContent;
+  if (typeof window === 'undefined' || !window.CSSStyleSheet) return;
+
+  const sheet = new CSSStyleSheet();
+  sheet.replaceSync(cssContent);
+  registeredStyleSheets.push(sheet);
 }
 
 /**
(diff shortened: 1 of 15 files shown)


packages/blockly/core/renderers/common/constants.ts
```diff
@@ -1121,26 +1122,22 @@ export class ConstantProvider {
...
-    this.cssNode = document.getElementById(cssNodeId) as HTMLStyleElement;
-    const text = cssArray.join('\n');
-    if (this.cssNode) {
-      // Already injected, update if the theme changed.
-      this.cssNode.firstChild!.textContent = text;
+  protected injectCSS_(root: Document | ShadowRoot, selector: string) {
+    if (
+      typeof window === 'undefined' ||
+      !window.CSSStyleSheet ||
+      injectionSites.has(root)
```
maribethb: will this return early now if the theme changes? i think the previous code allowed you to re-inject if the theme changed? not super familiar with this code though so could be misreading

packages/blockly/core/css.ts
```diff
@@ -28,37 +30,40 @@ export function register(cssContent: string) {
...
  */
-export function inject(hasCss: boolean, pathToMedia: string) {
-  // Only inject the CSS once.
-  if (injected) {
-    return;
-  }
-  injected = true;
-  if (!hasCss) {
+export function inject(
+  container: HTMLElement,
```
maribethb, line 39: again not too familiar with this code, but is it possible to have container be optional and the default value be `getParentContainer`?

maribethb (approved):
I don't know enough about web components so this seems plausible enough, but a couple questions.

packages/blockly/core/renderers/common/constants.ts
gonfunko: Great catch, yes! Updated it to be keyed by selector, so it only injects once per theme, and confirmed that fixes it.

packages/blockly/core/css.ts
gonfunko: We could, but this is only called from Blockly.inject(), and is explicitly a no-op if called again. At the point `Blockly.inject()` calls it, the workspace doesn't know its injection div yet, so `getParentContainer()` wouldn't return it.

maribethb (approved):

microbit-robert:
@maribethb @gonfunko  In the CSS cascade algorithm, adoptedStyleSheets are ordered after document.styleSheets, which affects precedence. As a result, where a Blockly class style was previously overridden by app-specific CSS without additional specificity by relying on document order, these overrides will no longer take effect. Consuming applications may need to bump the specificity of their overrides to restore custom styles.

This change will likely require an update to documentation.

Update: Possible additional CSS changes required due to inherited styles from the widget / dropdown div etc now being inside of the injectionDiv.

PR #9613, opened 2026-03-05
feat: Add a JSON Block definition interface

heliacer:

This is a follow-up of #9402 :D (please read)

**Changes:**
- Add custom field/input support as requested by @maribethb
- Add tests in `tests/typescript` for field mitosis

The custom fields are done by augmenting, as shown in the jsdoc example.
Also I'm not too proud of the documentation strings in general but I'm leaving it as-is now.

lmk if I need to change something in order to merge :>

### DIFF
--- a/packages/blockly/core/interfaces/i_json_block_definition.ts
+++ b/packages/blockly/core/interfaces/i_json_block_definition.ts
@@ -1,0 +1,126 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {FieldCheckboxFromJsonConfig} from '../field_checkbox.js';
+import {FieldDropdownFromJsonConfig} from '../field_dropdown';
+import {FieldImageFromJsonConfig} from '../field_image';
+import {FieldNumberFromJsonConfig} from '../field_number';
+import {FieldTextInputFromJsonConfig} from '../field_textinput';
+import {FieldVariableFromJsonConfig} from '../field_variable';
+import {Align} from '../inputs/align.js';
+
+/**
+ * Defines the JSON structure for a block definition.
+ *
+ * @example
+ * ```typescript
+ * const blockDef:  JsonBlockDefinition = {
+ *   type: 'custom_block',
+ *   message0: 'move %1 steps',
+ *   args0: [
+ *     {
+ *       'type': 'field_number',
+ *       'name': 'INPUT',
+ *     },
+ *   ],
+ *   previousStatement: null,
+ *   nextStatement: null,
+ * };
+ * ```
+ */
+export interface JsonBlockDefinition {
+  type: string;
+  style?: string | null;
+  colour?: string | number;
+  output?: string | string[] | null;
+  previousStatement?: string | string[] | null;
+  nextStatement?: string | string[] | null;
+  outputShape?: number;
+  inputsInline?: boolean;
+  tooltip?: string;
+  helpUrl?: string;
+  extensions?: string[];
+  mutator?: string;
+  enableContextMenu?: boolean;
+  suppressPrefixSuffix?: boolean;
+
+  [key: `message${number}`]: string | undefined;
+  [key: `args${number}`]: JsonBlockArg[] | undefined;
+  [key: `implicitAlign${number}`]: string | undefined;
+}
+
+export type JsonBlockArg =
+  | InputValueArg
+  | InputStatementArg
+  | InputDummyArg
+  | InputEndRowArg
+  | FieldInputArg
+  | FieldNumberArg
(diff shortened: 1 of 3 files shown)


heliacer:
hi, I noticed the review assignment changed,
@maribethb pinging in case you have time to review this.
thanks :D

packages/blockly/blocks/loops.ts
```diff
@@ -124,7 +124,6 @@ export const blocks = createBlockDefinitionsFromJsonArray([
       {
         'type': 'field_variable',
         'name': 'VAR',
-        'variable': null,
```
maribethb, line 127: The change to this file should not be made. You are changing the block definitions in unexpected ways.

packages/blockly/blocks/variables_dynamic.ts
```diff
@@ -38,7 +38,6 @@ export const blocks = createBlockDefinitionsFromJsonArray([
         'variable': '%{BKY_VARIABLES_DEFAULT_NAME}',
       },
     ],
-    'output': null,
```
maribethb, line 41: The change to this file should also be reverted. You've removed an output from one block and added it to a different one unexpectedly. This will change the function and shape of the blocks.

packages/blockly/core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,144 @@
...
+  [key: `args${number}`]: JsonBlockArg[] | undefined;
+  [key: `implicitAlign${number}`]: string | undefined;
+}
+
+/**
+ * A map for custom args.
+ *
+ * @example
+ * ```typescript
+ * declare module 'blockly/core/interfaces/i_json_block_definition' {
```
maribethb: This is an interesting idea, but I don't think it's a pattern we want to add here. Using `declare module` to augment the types might work, but we'd probably want to implement it in a different way so that developers don't have to use this specific filepath which is more of an implementation detail.

I think for now, it would be best to have custom fields/inputs/args use `unknown` type. While this isn't as helpful as having a fully typed interface, it would let developers create their own types for their own internal use but pass them to Blockly as unknown, which is an acceptable compromise at this point.

packages/blockly/core/interfaces/i_json_block_definition.ts
```diff
@@ -0,0 +1,144 @@
...
+  type: 'field_image';
+  name?: string;
+}
+
+interface FieldVariableArg extends FieldVariableFromJsonConfig {
+  type: 'field_variable';
+  name?: string;
+}
+
+export type FieldsAlign = 'LEFT' | 'RIGHT' | 'CENTRE';
```
maribethb: We might want to use `Blockly.inputs.Align` for this instead of redeclaring these strings.

packages/blockly/tests/typescript/src/field/json_block_custom_args.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {defineBlocksWithJsonArray} from 'blockly-test/core';
+import type {JsonBlockDefinition} from 'blockly-test/core/interfaces/i_json_block_definition';
+
+import './different_user_input';
+
+declare module 'blockly-test/core/interfaces/i_json_block_definition' {
```
maribethb: As discussed above, if you have the internal Blockly types use `unknown` for custom fields, you should still be able to declare and use this type locally (so without the `declare module` bit).

maribethb (commented):
Thank you so much for your patience while we reviewed this. This needed some consultation from our resident TS expert and after chatting with him I think we want to go in a different direction for custom fields for now. But the rest of this is mostly solid, thank you for your work on it.

heliacer:
hi @maribethb, 
I've made the changes you requested.

for the custom fields (and inputs), I provided an `UnknownArg` which looks like this

```ts
interface UnknownArg {
  type: string
  [key: string]: unknown
}
```

this preserves existing typings, so intellisense continues to work as expected.
If you'd start typing "pr" it would suggest you `previousStatement` and so on. I personally think that this change is really helpful, since it doesn't require you to always rely on the block factory if you already know how the basic json block definition looks like.

Thanks :D & Have a good day

maribethb (approved):
Thanks, this is looking great. We really appreciate your effort on this and your patience while we got it reviewed. We'll call this out in the review notes and encourage other folks to try them out so that we can enable them by default in a future release.

heliacer:
> Thanks, this is looking great. We really appreciate your effort on this and your patience while we got it reviewed. We'll call this out in the review notes and encourage other folks to try them out so that we can enable them by default in a future release.

sounds great :D

PR #9615, opened 2026-03-06
feat: Add keyboard shortcut to focus the workspace

gonfunko:
## The basics

### Proposed Changes
This PR backports the W shortcut from the keyboard-experiment repo that moves focus to the workspace. It slightly refactors registration of shortcuts to define a separate registration method for keyboard navigation-related shortcuts, so that these may be more easily enabled or removed. Tests were LLM generated and heavily modified manually.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -38,6 +38,14 @@ export enum names {
   UNDO = 'undo',
   REDO = 'redo',
   MENU = 'menu',
+  FOCUS_WORKSPACE = 'focus_workspace',
+  START_MOVE = 'start_move',
+  FINISH_MOVE = 'finish_move',
+  ABORT_MOVE = 'abort_move',
+  MOVE_UP = 'move_up',
+  MOVE_DOWN = 'move_down',
+  MOVE_LEFT = 'move_left',
+  MOVE_RIGHT = 'move_right',
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/blockly/core/shortcut_items.ts
```diff
@@ -523,6 +531,33 @@ export function registerShowContextMenu() {
...
+      if (target) {
+        return resolveWorkspace(target);
+      }
+    }
+    return workspace.getRootWorkspace() ?? workspace;
+  };
+
+  const contextMenuShortcut: KeyboardShortcut = {
+    name: names.FOCUS_WORKSPACE,
+    preconditionFn: () => true,
```
maribethb: I think we probably need to not do this mid-move/drag

maribethb (approved):

packages/blockly/core/shortcut_items.ts
gonfunko: The mover already handles it via the register-a-shortcut-for-every-other shortcut hack that commits the move.

maribethb: I think you should still exclude dragging, as right now in the plugin if you press w during move it highlights the workspace (during mouse drag)

gonfunko: Ah good call, done!

maribethb (approved):

PR #9629, opened 2026-03-11
feat: Add support for going to the beginning of a block

### LINKED ISSUE #9628: Center on start block
### Check for duplicates

### Problem

At the moment, it is possible to center the block only to the middle of it with the stack centering option.

### Request

By adding a parameter and making edits, it will be possible to achieve the behavior so that the block is centered in the middle of the workspace. This will be useful when the block has child blocks and it stretches a lot.

### Alternatives considered

_No response_

### Additional context

_No response_

bbrudastyy:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9628

### Proposed Changes

Allows you to center a block at its beginning.

### Reason for Changes

Centering takes place in the center of the block, which is inconvenient, for example, when switching to defining a procedure.
image

You can use the changes to go to the beginning.
image

### Test Coverage

The changes do not affect existing tests and do not require writing new ones.

### Documentation

JSDoc has been updated
https://developers.google.com/blockly/reference/js/blockly.workspacesvg_class.centeronblock_1_method?hl=it

### Additional Information

No

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1819,6 +1819,36 @@ export class BlockSvg
     return this.dragStrategy.isMovable();
   }
 
+  /** Returns whether the block fully fits within the boundaries of the workspace */
+  isBlockFullyInBounds(): boolean {
+    let blockLeft;
+    let blockRight;
+    const {left, top, width, height} = this.workspace
+      .getMetricsManager()
+      .getViewMetrics(true);
+
+    const xy = this.getRelativeToSurfaceXY();
+    const {width: blockWidth, height: blockHeight} = this.getHeightWidth();
+
+    if (this.RTL) {
+      blockLeft = xy.x - blockWidth;
+      blockRight = xy.x;
+    } else {
+      blockLeft = xy.x;
+      blockRight = xy.x + blockWidth;
+    }
+
+    const blockTop = xy.y;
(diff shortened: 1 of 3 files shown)


gonfunko (commented):
Thanks for the PR! This seems very useful, but I think it would be better if this was automatic rather than requiring the caller to request this behavior. Would you be open to revising this change to always center on the top of the block if the block being centered is taller than the current viewport? It would be good to have a test or two as well.

bbrudastyy:
> Thanks for the PR! This seems very useful, but I think it would be better if this was automatic rather than requiring the caller to request this behavior. Would you be open to revising this change to always center on the top of the block if the block being centered is taller than the current viewport? It would be good to have a test or two as well.

Hi, you want to center the block to the beginning at the xy coordinates.y only if the procedure definition does not fit in the workspace viewport? If it fits, then leave xy.y + heightWidth.height / 2?

gonfunko:
Yes, if the block fits within the viewport it should be centered (current Blockly behavior), but if it's too tall to fit completely in view then the top of the block should be centered vertically in the workspace (like you change does). Thanks!

bbrudastyy:
I have placed the block definition in the block_svg file. In the future, you may need to https://github.com/RaspberryPiFoundation/blockly/issues/9640

PR #9630, opened 2026-03-11
feat: Move mode for stacks of blocks

mikeharv:
## The basics

### Resolves

Fixes first part of https://github.com/RaspberryPiFoundation/blockly/issues/9625

### Proposed Changes

Adds a new shortcut to enable move mode with Shift+M. Updates block drag strategy so that keyboard events including shift will move the full stack of blocks (similar to a normal pointer drag).

### Reason for Changes

Shift+M should activate move mode for the focused block and its children (the same set of blocks that would be moved if you use the mouse)

### Test Coverage

Add new tests to check that disconnecting blocks works as expected when turning on move mode. We check each of three blocks in a stack using both the standard and modified shortcut.

### DIFF
--- a/packages/blockly/tests/mocha/keyboard_movement_test.js
+++ b/packages/blockly/tests/mocha/keyboard_movement_test.js
@@ -35,6 +35,13 @@ suite('Keyboard-driven movement', function () {
     workspace.getInjectionDiv().dispatchEvent(event);
   }
 
+  function startMoveStack(workspace) {
+    const event = createKeyDownEvent(Blockly.utils.KeyCodes.M, [
+      Blockly.utils.KeyCodes.SHIFT,
+    ]);
+    workspace.getInjectionDiv().dispatchEvent(event);
+  }
+
   function moveUp(workspace, modifiers) {
     const event = createKeyDownEvent(Blockly.utils.KeyCodes.UP, modifiers);
     workspace.getInjectionDiv().dispatchEvent(event);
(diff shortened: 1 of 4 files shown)


packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -407,6 +414,103 @@ suite('Keyboard-driven movement', function () {
     testExemptedShortcutsAllowed();
   });
 
+  suite('to disconnect blocks', function () {
+    setup(function () {
+      // Clear workspace and build 3-block stack.
+      this.workspace.clear();
```
gonfunko: The workspace should have been set up afresh before each test, the clearing shouldn't be necessary.

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -407,6 +414,103 @@ suite('Keyboard-driven movement', function () {
...
+      assert.equal(this.block1.isDragging(), false);
+      assert.equal(this.block2.isDragging(), true);
+      assert.equal(this.block3.isDragging(), false);
+      cancelMove(this.workspace);
+    });
+
+    test('from bottom block - Detaches single block', function () {
+      Blockly.getFocusManager().focusNode(this.block3);
+      startMove(this.workspace);
+      assert.isNull(this.block3.nextConnection.targetBlock());
```
gonfunko: The previous connection is more relevant to verify here since it was already the last block in the stack

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -407,6 +414,103 @@ suite('Keyboard-driven movement', function () {
...
+      assert.equal(this.block1.isDragging(), false);
+      assert.equal(this.block2.isDragging(), true);
+      assert.equal(this.block3.isDragging(), true);
+      cancelMove(this.workspace);
+    });
+
+    test('from bottom block - Detaches single-block stack from bottom', function () {
+      Blockly.getFocusManager().focusNode(this.block3);
+      startMoveStack(this.workspace);
+      assert.isNull(this.block3.nextConnection.targetBlock());
```
gonfunko: As with the single block case I'd check previousConnection here

gonfunko (approved):
Just a couple minor nits but otherwise looks good, thanks!

packages/blockly/tests/mocha/keyboard_movement_test.js
mikeharv: Good idea. I also added a `previousConnection` check for `block2` for each of the "middle block" tests.

PR #9634, opened 2026-03-16
feat!: Add support for keyboard navigation

gonfunko:
## The basics

### Proposed Changes
This PR backports the core support for keyboard navigation of the workspace, toolbox and flyout from the blockly-keyboard-experimentation repo. It also refactors things to provide a more cohesive API for interacting with navigation. The general aspirational approach is:

* The various `*_navigation_policy.ts` files specify, for each kind of focusable/navigable entity in Blockly, what its parent, first child, and previous/next siblings are, from a "just the facts"/AST perspective.
* `Navigator` provides methods to query the "just the facts" relationships between elements, but also provides methods that incorporate business logic to make navigation order in response to the arrow keys make sense/feel right/pass the vibe test.
* `FlyoutNavigator` and `ToolboxNavigator` customize the business logic for navigation in a flyout and toolbox context, respectively, and operate on a set of navigation policies relevant to focusable/navigable entities that exist in those contexts
* Every `IFocusableTree` provides a `Navigator` instance which should be used to handle navigation within that tree
* The actual shortcuts for the arrow keys are effectively one-liners that delegate to the `Navigator` for the active focus tree

This means that keyboard navigation does not have to keep track of the current state/context as in the keyboard-experimentation repo, because that maps precisely to the currently focused tree, which the `FocusManager` already keeps track of.

I backported the navigation business logic from the add-screen-reader-support-experimental branch, where up/down move between "rows" and left/right navigate within the current "row". 

### Future Work
* Navigation within the toolbox needs to be adjusted to accomodate horizontal layouts and the various start/end toolbox positions
* Some CSS rules need to be backported to make it clearer where focus is
* The navigation policies should be cleaned up a bit to remove some vestiges of business logic
* Additional tests for navigation behavior should be added; the suite from the keyboard-experimentation repo was backported with LLM assistance and manual review, but additional cases should be exercised, particularly with looping enabled
* Enter/space to act on the focused element still needs to be backported

### Breaking Changes
* `MarkerManager`, `Marker`, and `LineCursor` have all been removed, along with their accessors. If you were interacting with one of these classes, `Navigator` should be close to a drop-in replacement.
* `IFocusableTree` must implement `getNavigator()`, which should return a `Navigator` instance or subclass appropriate for navigating the tree
* `Flyout` no longer implements `IFocusableNode` or `IFocusableTree`. The flyout's workspace continues to implement those interfaces, and should be used in place of the flyout itself.
* Changes have been made to several toolbox-related interfaces. If you have custom toolbox-related classes that do not inherit from those provided in Blockly, you may need to adjust your implementation to ensure that:
    * `IToolbox` provides a `getToolboxItems()` method that returns an array of all items
    * `IToolboxItem` provides a `getParentToolbox()` method that returns a reference to its containing toolbox


packages/blockly/core/interfaces/i_toolbox.ts
```diff
@@ -9,7 +9,7 @@
 import type {ToolboxInfo} from '../utils/toolbox.js';
 import type {WorkspaceSvg} from '../workspace_svg.js';
 import type {IFlyout} from './i_flyout.js';
-import type {IFocusableTree} from './i_focusable_tree.js';
+import {type IFocusableTree} from './i_focusable_tree.js';
```
maribethb: should this change be here? it's different than all the others

gonfunko:
This fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/241

packages/blockly/core/keyboard_nav/navigation_policies/flyout_navigation_policy.ts
```diff
@@ -26,11 +26,11 @@ export class FlyoutNavigationPolicy<T> implements INavigationPolicy<T> {
   /**
    * Returns null to prevent navigating into flyout items.
```
maribethb: The jsdoc no longer matches the implementation and I can't tell which one should be correct.

Or is the behavior the same but you moved the responsibility of returning null to the policy? I would still probably clarify the jsdoc in that case.

packages/blockly/core/keyboard_nav/navigation_policies/flyout_navigation_policy.ts
```diff
@@ -96,7 +108,13 @@ export class FlyoutNavigationPolicy<T> implements INavigationPolicy<T> {
    * @returns True if the given flyout item can be focused.
    */
   isNavigable(current: T): boolean {
-    return this.policy.isNavigable(current);
+    return (
+      this.policy.isNavigable(current) &&
+      this.flyout
```
maribethb: curious why this is needed - what scenarios are coming up where a flyout item isn't included in the flyout contents?

packages/blockly/core/keyboard_nav/navigation_policies/toolbox_item_navigation_policy.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export class ToolboxItemNavigationPolicy
+  implements INavigationPolicy<IToolboxItem>
+{
+  /**
+   * Returns the first child of the given toolbox item.
+   *
+   * @param current The toolbox item to return the first child of.
+   * @returns The child item of a collapsible toolbox item, or the flyout
+   *     workspace for non-collapsible flyout items.
```
maribethb: this comment doesn't seem to match reality - it returns null, not the workspace. i'm not sure why it would return the workspace?

packages/blockly/core/keyboard_nav/navigators/navigator.ts
```diff
@@ -0,0 +1,498 @@
...
+    new WorkspaceCommentNavigationPolicy(),
+    new CommentBarButtonNavigationPolicy(),
+    new BlockCommentNavigationPolicy(),
+    new CommentEditorNavigationPolicy(),
+  ];
+
+  /** Whether or not navigation loops around when reaching the end. */
+  protected navigationLoops = false;
+
+  protected relativeNode: IFocusableNode | null = null;
```
maribethb: can you add jsdoc for this please?

packages/blockly/core/keyboard_nav/navigators/navigator.ts
```diff
@@ -0,0 +1,498 @@
...
+          const currentParents = currentBlock.getOutputParents();
+          const candidateParents = candidateBlock.getOutputParents();
+          // If we're navigating from a block (or nested element) to a block
+          // (or nested element), ensure that we're not crossing a statement
+          // block boundary (i.e. moving to a next or previous block vertically)
+          // by verifying that the two blocks in question are either the same
+          // or have a common parent accessible only by traversing output
+          // connections, meaning that they are part of the same row.
+          return (
+            (candidateParents as any).intersection(currentParents).size > 0
```
maribethb: what's the reason for the cast here and is there a way we can eliminate it?

packages/blockly/core/keyboard_nav/navigators/navigator.ts
```diff
@@ -0,0 +1,498 @@
...
+   * @param direction The direction in which the user is navigating.
+   * @returns A function that takes a proposed navigation candidate and returns
+   *     true if navigation should be allowed to proceed to it, or false to find
+   *     a different candidate.
+   */
+  getValidationFunction(
+    direction: NavigationDirection,
+  ): (node: IFocusableNode) => boolean {
+    switch (direction) {
+      case NavigationDirection.IN:
```
maribethb, line 398: There is a LOT of logic in this function. I was expecting more of this logic to live in the navigation policies themselves rather than have so much business logic in the navigator. It's confusing to me that you have some logic here but some similar-seeming business logic about allowable possible navigation stops in the block navigation policy for example.

Possibly we should discuss this over video call because of the complexity


If this logic has to stay here I think it's worth investing in it to make this as readable as possible. Maybe start by splitting up the in/out vs next/prev into different named functions to remove one layer of logic?

packages/blockly/core/keyboard_nav/navigators/navigator.ts
```diff
@@ -0,0 +1,498 @@
...
+          // or have a common parent accessible only by traversing output
+          // connections, meaning that they are part of the same row.
+          return (
+            (candidateParents as any).intersection(currentParents).size > 0
+          );
+        };
+      case NavigationDirection.NEXT:
+      case NavigationDirection.PREVIOUS:
+        return (candidate: IFocusableNode | null) => {
+          if (
```
maribethb: This needs more comments explaining, or maybe it would be more readable if you split up the conditions into separate statements.
yes it's more redundant but the super nested conditions are difficult to parse and probably need individual comments with reasoning

packages/blockly/core/keyboard_nav/navigators/navigator.ts
```diff
@@ -0,0 +1,498 @@
...
+  getSourceBlockFromNode(node: IFocusableNode | null): BlockSvg | null {
+    if (node instanceof BlockSvg) {
+      return node;
+    } else if (node instanceof Field) {
+      return node.getSourceBlock() as BlockSvg;
+    } else if (node instanceof RenderedConnection) {
+      return node.getSourceBlock();
+    } else if (node instanceof Icon) {
+      return node.getSourceBlock() as BlockSvg;
+    }
```
maribethb, line 522: worth adding a `if typoeof node.getSourceBlock === function return node.getSourceBlock()` to sort of make this work as if we had a `IBelongsToBlock` interface or something? Or worth formalizing that as an interface? potentially as a followup.

packages/blockly/core/flyout_base.ts
```diff
@@ -212,6 +210,7 @@ export abstract class Flyout
     // Keep the workspace visibility consistent with the flyout's visibility.
     this.workspace_.setVisible(this.visible);
     this.workspace_.setNavigator(new FlyoutNavigator(this));
+    this.workspace_.keyboardAccessibilityMode = true;
```
maribethb: I think we probably need to remove this constant? It's not used anywhere that I can tell? We shouldn't need both it and the keyboardNavigationController.

packages/blockly/core/keyboard_nav/navigation_policies/flyout_navigation_policy.ts
gonfunko: The thing being checked isn't necessarily a flyout item; this prevents things that aren't top-level flyout items (e.g. icons, fields, connections) from becoming focused by indicating that they aren't navigable.

packages/blockly/core/flyout_base.ts
gonfunko: Removed

packages/blockly/core/keyboard_nav/navigation_policies/flyout_navigation_policy.ts
gonfunko: Removed this file.

packages/blockly/core/interfaces/i_toolbox.ts
gonfunko: No, reverted.

packages/blockly/core/keyboard_nav/navigation_policies/toolbox_item_navigation_policy.ts
gonfunko: Updated, originally the logic that's now in the navigator was here.

packages/blockly/core/keyboard_nav/navigators/navigator.ts
gonfunko: Removed it instead!

packages/blockly/core/keyboard_nav/navigators/navigator.ts
gonfunko: Removed in the refactor, but it's because JS didn't add `intersection()` to the built in `Set` until recently and TS doesn't know about it yet.

packages/blockly/core/keyboard_nav/navigators/navigator.ts
gonfunko: Removed in favor of the new row ID approach.

packages/blockly/core/keyboard_nav/navigators/navigator.ts
gonfunko: Removed in favor of the row ID approach.

packages/blockly/core/keyboard_nav/navigation_policies/connection_navigation_policy.ts
```diff
@@ -140,7 +119,22 @@ export class ConnectionNavigationPolicy
...
+    // Empty next connections on block stacks inside of a C shaped block are
+    // navigable.
+    if (current.type === ConnectionType.NEXT_STATEMENT) {
+      if (current.targetBlock()) return false;
+
+      const rootBlock =
+        current.getSourceBlock().getRootBlock() ?? current.getSourceBlock();
+      if (current === rootBlock.lastConnectionInStack(false)) return false;
+
+      return true;
```
maribethb, line 133: do you still need to check if `current.canBeFocused()` in this case? i'm not sure of the scenarios in which that might be false

packages/blockly/core/keyboard_nav/navigation_policies/flyout_button_navigation_policy.ts
```diff
@@ -54,6 +54,16 @@ export class FlyoutButtonNavigationPolicy
...
   }
 
+  /**
+   * Returns the row ID of the given flyout button.
+   *
+   * @param current The flyout button to retrieve the row ID of.
+   * @returns The row ID of the given flyout button.
+   */
+  getRowId(current: FlyoutButton) {
+    return current.getButtonText();
```
maribethb: this feels fragile, there's already an `id` property in `FlyoutButton` so maybe add a getter for that instead.

packages/blockly/core/inputs/input.ts
```diff
@@ -314,4 +314,37 @@ export class Input {
   protected makeConnection(type: ConnectionType): Connection {
     return this.sourceBlock.makeConnection_(type);
   }
+
+  /**
+   * Returns an ID for the visual "row" this input is part of.
+   *
+   * @internal
```
maribethb, line 321: I don't think this should be internal since it would be called from a navigation policy which is externally modifiable. right?

but if it's not internal, you can still add a disclaimer that the value only holds meaning for keyboard navigation and the value is not guaranteed to be stable, to discourage people from relying on this to do shenanigans

packages/blockly/core/interfaces/i_navigation_policy.ts
```diff
@@ -44,6 +44,19 @@ export interface INavigationPolicy<T> {
...
+   * of. All elements on the same visual row should share the same ID. For
+   * example, icons share their parent block's row ID, as do inline connected
+   * blocks or value inputs. Statement inputs, external inputs, or blocks
+   * connected to one another's previous or next connections form distinct
+   * visual rows and should have distinct row IDs.
+   *
+   * @param current The element to return the row ID of.
+   * @returns The row ID of the given element.
+   */
+  getRowId(current: T): string;
```
maribethb, line 58: the one remaining thing I find odd about this is that things that aren't on blocks also have this, like workspace comments. but i can't think of a different way to do this that's better

maribethb (approved):

packages/blockly/core/interfaces/i_navigation_policy.ts
gonfunko: In theory the bar button icons could be one row and the editor field another; it's not currently set up that way, but I don't think it's too much of a stretch. And the comment itself is of course a form of row.

packages/blockly/core/inputs/input.ts
gonfunko: Yeah, I was on the fence about this. You can create your own NavigationPolicy, subclass the existing ones, change the set that a Navigator uses, or subclass Navigator. There isn't inherently a reason to need this method though; we don't provide any ability to add new kinds of entities to blocks, where the block version of this might be needed, and you could for example subclass the Field or Connection navigation policies which call this method but just not override getRowId() in those, although again I don't think there's any reason to now that the navigation policies themselves don't really contain business logic. Navigator just calls getRowId() on the navigation policy, not this method, so you can even get it there with public API. If you make a new entity and put it in the flyout you'd need to specify a row ID in your navigation policy, but that would be your own synthetic ID, not an input or block one. So I think @internal is fine, especially since it's a sign not a cop? In general I figured we could start it off that way, and if someone came to us with a compelling reason why they needed it we could remove it. WDYT?

packages/blockly/core/keyboard_nav/navigation_policies/connection_navigation_policy.ts
gonfunko: I think in practice the current approach is safe, but that's more defensive and probably strictly more correct, so updated accordingly.

packages/blockly/core/keyboard_nav/navigation_policies/flyout_button_navigation_policy.ts
gonfunko: Done!

packages/blockly/core/inputs/input.ts
maribethb: I was thinking the case where someone registers their own input navigation policy, they would need to call this.   But that's fine, I think you're right it can start this way until needed.

maribethb (approved):

PR #9641, opened 2026-03-19
fix: visit all connection candidates in move mode

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9631

---

## Proposed Changes

- Fixes keyboard navigation skipping valid connection candidates (now matches mouse behavior)
- Replaces incremental traversal with a precomputed list of valid connection pairs
- Ensures deterministic, complete, and cyclic traversal
- Improves traversal order by reordering local connections for more natural right/down movement (configurable with a `natural` flag)
- Adds `initialConnectionPair` so traversal starts from the block’s original connection
- Refines constrained-move positioning for more consistent visual alignment

---

## Reason for Changes

Keyboard navigation previously missed valid connection targets.

This PR makes traversal complete, predictable, and easier to reason about.

---

## Test Coverage

- Updated tests to include previously skipped candidates (now expected).
- Manually tested keyboard navigation on complex block structures.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -31,14 +31,16 @@ import * as dom from '../utils/dom.js';
 import * as svgMath from '../utils/svg_math.js';
 import type {WorkspaceSvg} from '../workspace_svg.js';
 
-/** Represents a nearby valid connection. */
-interface ConnectionCandidate {
+/** Represents a valid pair of connections between the dragging block and a block on the workspace. */
+interface ConnectionPair {
   /** A connection on the dragging stack that is compatible with neighbour. */
   local: RenderedConnection;
-
   /** A nearby connection that is compatible with local. */
   neighbour: RenderedConnection;
+}
 
+/** Represents a nearby valid connection. */
+interface ConnectionCandidate extends ConnectionPair {
   /** The distance between the local connection and the neighbour connection. */
   distance: number;
 }
@@ -65,6 +67,8 @@ export class BlockDragStrategy implements IDragStrategy {
    */
   private startChildConn: RenderedConnection | null = null;
 
+  private initialConnectionPair: ConnectionCandidate | null = null;
+
   private startLoc: Coordinate | null = null;
 
   private connectionCandidate: ConnectionCandidate | null = null;
@@ -73,11 +77,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   private dragging = false;
 
-  /** Where a constrained movement should start when traversing the tree. */
-  private searchNode: RenderedConnection | null = null;
+  private localConnections: RenderedConnection[] | null = null;
+  private workspaceConnections: RenderedConnection[] | null = null;
 
   /** List of all connections available on the workspace. */
-  private allConnections: RenderedConnection[] = [];
+  private allConnectionPairs: ConnectionPair[] = [];
 
   /** The current movement mode. */
   private moveMode = MoveMode.UNCONSTRAINED;
@@ -169,6 +173,7 @@ export class BlockDragStrategy implements IDragStrategy {
 
     this.startLoc = this.block.getRelativeToSurfaceXY();
 
+    this.initialConnectionPair = null;
     this.connectionCandidate = null;
     const previewerConstructor = registry.getClassFromOptions(
       registry.Type.CONNECTION_PREVIEWER,
@@ -187,6 +192,8 @@ export class BlockDragStrategy implements IDragStrategy {
     if (this.shouldDisconnect(healStack)) {
       this.disconnectBlock(healStack);
     }
+
+    this.localConnections = this.getLocalConnections(this.block, true);
     this.block.setDragging(true);
     this.workspace.getLayerManager()?.moveToDragLayer(this.block);
     this.getVisibleBubbles(this.block).forEach((bubble) => {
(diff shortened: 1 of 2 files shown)


packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -65,6 +67,8 @@ export class BlockDragStrategy implements IDragStrategy {
    */
   private startChildConn: RenderedConnection | null = null;
 
+  private initialConnectionPair: ConnectionCandidate | null = null;
```
gonfunko: Is this necessary? It seems to only be set in `storeInitialConnections()`, which gets called by `disconnectBlock()` which in turn gets called by `startDrag()`. `startDrag()` then just assigns this value to `connectionCandidate`, so it seems like `storeInitialConnections()` could just assign directly to `connectionCandidate` and this field could be removed?

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -73,11 +77,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   private dragging = false;
 
-  /** Where a constrained movement should start when traversing the tree. */
-  private searchNode: RenderedConnection | null = null;
+  private localConnections: RenderedConnection[] | null = null;
+  private workspaceConnections: RenderedConnection[] | null = null;
```
gonfunko: This also appears to not be needed – seemingly it's only used in `cacheAllConnectionPairs()`, and could be replaced with a local variable there?

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -73,11 +77,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   private dragging = false;
 
-  /** Where a constrained movement should start when traversing the tree. */
-  private searchNode: RenderedConnection | null = null;
+  private localConnections: RenderedConnection[] | null = null;
```
gonfunko: I'd kind of prefer to just fetch this dynamically in the two places it's used; I'm not adverse to instance variables in general, but we have a lot of different connection permutations right now that make it difficult to reason about things, vs just one list of pairs.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -297,17 +331,56 @@ export class BlockDragStrategy implements IDragStrategy {
...
+
+    this.startParentConn = parentTargetConn;
+    if (localParentConn && parentTargetConn) {
+      this.initialConnectionPair = {
+        local: localParentConn,
+        neighbour: parentTargetConn,
+        distance: 0,
+      };
+    } else {
+      // If there is no parent conenction and we are moving a single block,
```
gonfunko: nit: connection

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -545,12 +630,37 @@ export class BlockDragStrategy implements IDragStrategy {
    * Includes any connections on the dragging block, and any last next
    * connection on the stack (if one exists).
    */
-  private getLocalConnections(draggingBlock: BlockSvg): RenderedConnection[] {
+  private getLocalConnections(
+    draggingBlock: BlockSvg,
+    natural: boolean,
```
gonfunko: I think we just always want the natural behavior

mikeharv: Sounds good. I wasn't sure if we might want to make it configurable or A/B testable later. For example, the natural behavior could be good for sighted users but less intuitive for screen reader users. Easy enough to add in again if we ever want it!

packages/blockly/core/dragging/block_drag_strategy.ts
mikeharv: Good point. I confirmed it's basically redundant so I refactored to remove it.

packages/blockly/core/dragging/block_drag_strategy.ts
mikeharv: That makes sense. My original intent with caching `localConnections` was to avoid recomputing them on every drag frame, since `getConnectionCandidate` can be hit pretty frequently during pointer moves. Happy to fetch dynamically if the readability outweighs any slight performance dips. Presumably it's worked this way for a while.

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Thanks!

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Thanks!

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: That's a good point, I think since it's just the immediate connections on the dragging block it should be fine, but worth keeping in mind if we do run into any performance issues.

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Makes sense! I'd ask Maribeth for her thoughts as well, but typically I prefer to shy away from configuration options without a clear and present need as anything we add to the API surface is something we're committing to maintain ~forever. Granted this method is private so that's not directly a concern here, but as a general principle that's how I tend to approach things. We also don't have any mechanism for A/B testing since we're operating solely at the library level and intentionally don't have any sort of metrics/analytics since that would affect every partner/team/product that uses Blockly and their actual end users.

gonfunko (approved):

packages/blockly/core/dragging/block_drag_strategy.ts
mikeharv: Good points! Thanks for explaining.

PR #9650, opened 2026-03-23
feat: Add keyboard shortcut for disconnecting the selected block

gonfunko:
## The basics

### Proposed Changes
This PR backports the block disconnect keyboard shortcut (X) from the keyboard-experimentation repo. It also adds a Shift X variant, that, consistent with move mode, disconnects the focused block and the rest of the stack following it, rather than just the focused block by itself.


BenHenning (approved):
Nothing to add, it looks good. Thanks @gonfunko!

PR #9653, opened 2026-03-26
feat: aria live region for announcements

mikeharv:
This adds an aria live region for announcing updates on Blockly. This was partially backported out of the screen-reader branch.

## The basics

API verified by testing announcement of toast changes via uncommitted code.

### Resolves https://github.com/RaspberryPiFoundation/blockly/issues/9632

Fixes 

### Proposed Changes

During `inject()`, we will create a displayed-but-user-invisible `<div/>` with id `blocklyAriaAnnounce` which is configured as an aria live region with passive priority. Other parts of Blockly will be able to call `announceDynamicAriaState` with text to add to this region to be read by assistive technologies. An asynchronous timeout is used to manage replacing outdated pending announcements with the latest update.

### Reason for Changes

This will give us a centralized place to announce dynamic state changes and information that cannot otherwise be conveyed.

### Test Coverage

Added ARIA live region tests covering initialization, accessibility attributes, delayed announcements, repeated message handling, last-write-wins behavior, and delayed updates to role and assertiveness to prevent regressions.

### Documentation

N/A

### DIFF
--- a/packages/blockly/core/inject.ts
+++ b/packages/blockly/core/inject.ts
@@ -55,9 +55,6 @@ export function inject(
     dom.addClass(subContainer, 'blocklyRTL');
   }
 
-  // Ignore the subcontainer in aria since it is not focusable.
-  aria.setRole(subContainer, aria.Role.PRESENTATION);
-
   containerElement!.appendChild(subContainer);
   const svg = createDom(subContainer, options);
 
@@ -82,7 +79,7 @@ export function inject(
     common.globalShortcutHandler,
   );
 
-  aria.createLiveRegion(subContainer);
+  aria.initializeGlobalAriaLiveRegion(subContainer);
 
   return workspace;
 }
--- a/packages/blockly/core/inject.ts
+++ b/packages/blockly/core/inject.ts
@@ -17,6 +17,7 @@ import {Options} from './options.js';
 import {ScrollbarPair} from './scrollbar_pair.js';
 import * as Tooltip from './tooltip.js';
 import * as Touch from './touch.js';
+import * as aria from './utils/aria.js';
 import * as dom from './utils/dom.js';
 import {Svg} from './utils/svg.js';
 import * as WidgetDiv from './widgetdiv.js';
@@ -54,6 +55,9 @@ export function inject(
     dom.addClass(subContainer, 'blocklyRTL');
   }
 
+  // Ignore the subcontainer in aria since it is not focusable.
+  aria.setRole(subContainer, aria.Role.PRESENTATION);
+
   containerElement!.appendChild(subContainer);
   const svg = createDom(subContainer, options);
 
@@ -77,6 +81,8 @@ export function inject(
     null,
     common.globalShortcutHandler,
   );
+
+  aria.createLiveRegion(subContainer);
 
   return workspace;
 }
--- a/packages/blockly/core/utils/aria.ts
+++ b/packages/blockly/core/utils/aria.ts
@@ -18,7 +18,7 @@ const ROLE_ATTRIBUTE = 'role';
  * ARIA state values for LivePriority.
  * Copied from Closure's goog.a11y.aria.LivePriority
  */
-export enum LivePriority {
+export enum LiveRegionAssertiveness {
   // This information has the highest priority and assistive technologies
   // SHOULD notify the user immediately. Because an interruption may disorient
   // users or cause them to not complete their current task, authors SHOULD NOT
@@ -31,6 +31,23 @@ export enum LivePriority {
   // the next graceful opportunity, such as at the end of speaking the current
   // sentence or when the users pauses typing.
   POLITE = 'polite',
+}
+
+/**
+ * Customization options that can be passed when using `announceDynamicAriaState`.
+ */
+export interface DynamicAnnouncementOptions {
+  /** The custom ARIA `Role` that should be used for the announcement container. */
+  role?: Role;
+
+  /**
+   * How assertive the announcement should be.
+   *
+   * Important*: It was found through testing that `ASSERTIVE` announcements are
+   * often outright ignored by some screen readers, so it's generally recommended
+   * to always use `POLITE` unless specifically tested across supported readers.
+   */
+  assertiveness?: LiveRegionAssertiveness;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


packages/blockly/core/inject.ts
```diff
@@ -54,6 +55,9 @@ export function inject(
     dom.addClass(subContainer, 'blocklyRTL');
   }
 
+  // Ignore the subcontainer in aria since it is not focusable.
+  aria.setRole(subContainer, aria.Role.PRESENTATION);
```
BenHenning: This should be omitted. It's actually wrong on the experimental branch and for fairly nuanced reasons (essentially it's invalid to mark an element as having the `presentation` role when any of its children are actually interactive for the screen reader; doing so will cause this element to get a default role instead).

packages/blockly/core/utils/aria.ts
```diff
@@ -156,3 +181,85 @@ export function setState(
...
+ * Callers should use this judiciously. Over-announcing can reduce usability,
+ * so this should primarily be used for dynamic states or information that
+ * cannot be conveyed through standard ARIA semantics.
+ *
+ * @param text The text to announce to the user.
+ * @param options Configuration options for the announcement.
+ * @param options.assertiveness The ARIA live region priority
+ * @param options.role Optional ARIA role to apply to the live region before
+ */
+export function announceDynamicAriaState(
```
BenHenning, line 266: For this & the other changes in this file: is it possible perhaps to adapt the API the one proposed in the WIP screen reader design document? Specifically this section:

https://docs.google.com/document/d/1-q0paWfV9qYWoZKWb9BpcfYOy3xYZtvcLyuuD61qHXw/edit?tab=t.0#bookmark=id.a3l9k4e1tah4

That was written after analyzing the approach taken in the experimental branch, considering micro:bit's changes, and other longer-term considerations. If there's an aspect of the design that doesn't work then it'd be great to discuss (e.g. via chat and/or comments directly on the doc) so that the doc and code end up being close to the same.

BenHenning (commented):
Thanks @mikeharv! Had some initial thoughts--PTAL.

Separately for testing, I think it would still be a good idea, if possible, to try and add at least some fundamental tests for the new announcement behavior. Some things which are hopefully testable and could be quite valuable to avoid downstream issues:
- The live region is `polite`.
- The live region is displayed (i.e. not `display: none` since this will then be ignored by screen readers).
- That subsequent announcements even with the same text are unique from the last (e.g. due to the additional non-spoken character being toggled).
- The general behavior of the region correctly announcing.

It would also be helpful to test that the region updates its `role` and value after a delay but I'm not sure how easy that one is to check. This directly fixes a somewhat tricky-to-debug issue that micro:bit found and fixed, so a regression test for it is definitely a nice-to-have (and not something I think we can assume works correctly just from other tested functionality that eventually uses dynamic announcements).

packages/blockly/core/utils/aria.ts
mikeharv: Hi Ben. Thanks for the link and comments. I've pushed an update that attempts to better align to the document. PTAL and let me know what you think.

packages/blockly/core/utils/aria.ts
```diff
@@ -156,3 +198,75 @@ export function setState(
...
+ * over-announce information that can be inferred from other sources on the page,
+ * so this ought to be used only when certain context cannot be easily determined
+ * (such as dynamic states that may not have perfect ARIA representations or
+ * indications).
+ *
+ * @param text The text to read to the user.
+ * @param options Custom options to configure the announcement. This defaults to no
+ *    custom `Role` and polite assertiveness.
+ */
+
```
BenHenning: I think this blank newline needs to be removed to make jslint happy (since it may not realize that the JSDoc comment is actually associated with this function).

packages/blockly/tests/mocha/index.html
```diff
@@ -159,6 +159,7 @@
       import {javascriptGenerator} from '../../build/javascript.loader.mjs';
 
       // Import tests.
+      import './aria_live_region_test.js';
```
BenHenning: Probably should just be 'aria_test.js' for consistency with the other test suites (usually we just append '_test' for the test file name and instead use `suite`s within the test files to distinguish categories of behaviors being covered).

packages/blockly/core/utils/aria.ts
```diff
@@ -156,3 +198,75 @@ export function setState(
...
+  const ariaAnnouncementContainer = liveRegionElement;
+  const {assertiveness = LiveRegionAssertiveness.POLITE, role = null} =
+    options || {};
+
+  clearTimeout(ariaAnnounceTimeout);
+  ariaAnnounceTimeout = setTimeout(() => {
+    // Clear previous content.
+    ariaAnnouncementContainer.replaceChildren();
+    setState(ariaAnnouncementContainer, State.LIVE, assertiveness);
+    setRole(ariaAnnouncementContainer, role);
```
BenHenning, line 287: I think we may need to check if `role` is non-null here since, as-is, this will actually clear the current `Role` rather than keeping its default setting (which is what the API is documenting). Probably a good thing to double check via a test, as well, if there isn't already one (that a missing role doesn't clear the existing role set for the container).

Or perhaps put a different way: we should be keeping the `STATUS` `Role` if a custom one isn't provided. Another interesting test might be to try three subsequent announcements: one that doesn't set a role, one that does, and then one that doesn't to validate that it does correctly switch between `STATUS`, custom, and back to `STATUS` as expected.

packages/blockly/core/utils/aria.ts
```diff
@@ -156,3 +198,75 @@ export function setState(
...
+) {
+  if (!liveRegionElement) {
+    throw new Error('ARIA live region not initialized.');
+  }
+  const ariaAnnouncementContainer = liveRegionElement;
+  const {assertiveness = LiveRegionAssertiveness.POLITE, role = null} =
+    options || {};
+
+  clearTimeout(ariaAnnounceTimeout);
+  ariaAnnounceTimeout = setTimeout(() => {
```
BenHenning, line 283: Probably a good idea to drop a line comment above here to explain why the timeout is needed/what specific purpose it serves (since I think timeouts can generally be construed as a bit of a code smell, or at least something scary to change, without explicit explanation for why they're needed).

packages/blockly/core/utils/aria.ts
```diff
@@ -156,3 +198,75 @@ export function setState(
...
+  if (liveRegionElement && document.contains(liveRegionElement)) {
+    return;
+  }
+  const ariaAnnouncementDiv = document.createElement('div');
+  ariaAnnouncementDiv.textContent = '';
+  ariaAnnouncementDiv.id = 'blocklyAriaAnnounce';
+  dom.addClass(ariaAnnouncementDiv, 'hiddenForAria');
+  setState(ariaAnnouncementDiv, State.LIVE, LiveRegionAssertiveness.POLITE);
+  setRole(ariaAnnouncementDiv, Role.STATUS);
+  ariaAnnouncementDiv.setAttribute('aria-atomic', 'true');
```
BenHenning: This can be added as an explicit `State` value for cleanliness, I think. We're trying to avoid manually using `setAttribute` with anything ARIA-related and instead funneling through this utility for better type checking and auditing.

packages/blockly/tests/mocha/aria_test.js
```diff
@@ -0,0 +1,141 @@
...
+  test('live region is created', function () {
+    assert.isNotNull(this.liveRegion);
+  });
+
+  test('live region has polite aria-live', function () {
+    assert.equal(this.liveRegion.getAttribute('aria-live'), 'polite');
+  });
+
+  test('live region has atomic true', function () {
+    assert.equal(this.liveRegion.getAttribute('aria-atomic'), 'true');
```
BenHenning, line 33: I think you can use `assert.isTrue()` for this.

packages/blockly/tests/mocha/aria_live_region_test.js
```diff
@@ -0,0 +1,141 @@
...
+
+  test('live region has atomic true', function () {
+    assert.equal(this.liveRegion.getAttribute('aria-atomic'), 'true');
+  });
+
+  test('live region has status role by default', function () {
+    assert.equal(this.liveRegion.getAttribute('role'), 'status');
+  });
+
+  test('live region is visually hidden but not display none', function () {
```
BenHenning: Maybe check its position? Not sure how best to represent this but technically the test isn't actually validating the "is visually hidden" part yet.

BenHenning (changes requested):
Thanks @mikeharv! Great tests and changes overall! I had a few suggestions, PTAL.

packages/blockly/core/utils/aria.ts
BenHenning: This looks great. I had some follow-ups that I left in the latest review pass. Thanks!

packages/blockly/tests/mocha/aria_test.js
mikeharv: Not automatically, because `getAttribute` returns the string `'true'`. I can either keep it like this or coerce the attribute value to a boolean (ex. with `JSON.parse()`).

packages/blockly/tests/mocha/aria_live_region_test.js
mikeharv: Oh right, I hesitated on this one but forgot to come back around to it. I'll update it double-check the expected CSS, including position. Open to other ideas!

packages/blockly/core/utils/aria.ts
mikeharv: Sounds good. If we always want to keep `STATUS` if a custom `Role` isn't provided, I think we can also go back to the simpler version of `setRole` that doesn't expect `null` in the first place.

packages/blockly/core/utils/aria.ts
```diff
@@ -169,12 +174,8 @@ export enum State {
  * @param element DOM node to set role of.
  * @param roleName Role name.
  */
-export function setRole(element: Element, roleName: Role | null) {
```
BenHenning: This part actually seems correct and worth keeping (and also worth updating the JSDoc for the function). We do want an ability to clear the role, we just don't need to do it as part of the status stuff below.

packages/blockly/core/utils/aria.ts
BenHenning: I think that might still be worth keeping for the general ability to clear the role (which might be useful elsewhere), and I've left a new comment to cover that.

packages/blockly/tests/mocha/aria_test.js
BenHenning: Ah, nah this is fine. I missed that the type being evaluated is a string.

packages/blockly/tests/mocha/aria_live_region_test.js
BenHenning: The latest version is effectively a change detector test but it seems reasonable given that there aren't much better ways to test it. We could probably do an intersection check against the window to see if it's outside those bounds as an alternative option, but I think the current approach will do the job of regression detecting just fine.

BenHenning (approved):
Thanks @mikeharv! Generally happy with this, just had one follow-up. Going ahead and approving to unblock you but feel free to assign back to me if you want me to look at any follow-up changes.

packages/blockly/core/utils/aria.ts
mikeharv: Thanks Ben! I re-added the ability to clear the role, along with a direct `removeRole` function, similar to Closure (https://google.github.io/closure-library/api/goog.a11y.aria.html)

PR #9670, opened 2026-03-30
chore: Remove the keyboard-navigation plugin from the advanced playground

gonfunko:
## The basics

### Resolves
Now that keyboard navigation is being backported into core, the keyboard navigation plugin is unnecessary and actively conflicts with the built-in behavior in the advanced playground. This PR removes the dependency on the plugin in favor of the built-in behavior, which is available by default in the advanced playground, as well as the normal and multi variants.


mikeharv (approved):
🧹

PR #9671, opened 2026-03-31
chore(docs): migrate to Docusaurus and GitHub Pages

### LINKED ISSUE #9543: Migrate documentation site
Tracking issue for docs migration

I'm going to pass off this work so I'll document the progress I've made in this issue.

The work I've done so far is [here](https://github.com/maribethb/blockly/tree/docs).
There is an internal doc with some handoff info from cybage I'll share internally as needed.

I'll share what I've done in these commits, and then what I see as the remaining work. My suggestion would be to `git reset --soft HEAD~1` on my docs branch so that you keep the commits I've made that are distinct commits but can finish the rest of the files in another order that makes more logical sense.

## My approach

The cybage branch was [here](https://github.com/Joe-Davis-Blockly/blockly/tree/cybage-blockly/docs). This represents a working site if you ignore the fact that it's a whole npm project inside the blockly npm project. That is, while the docs site works, the blockly package it's contained in does not. And there are many places where the repository and branch names are hardcoded for the wrong value, or other assumptions have been made that don't hold up in reality. 

One problem is that initially the rest of the blockly source code was deleted from the fork, which makes it incredibly difficult to simply rebase these changes onto the current `main` branch of blockly. And while the commit history tells a bit of a story here, there are no pull requests where longer form discussion can be seen to explain some of the choices made. Therefore, I did not feel the git history is particularly worth preserving, and I soft reset the history to edit and commit the files in an order I felt would be more useful going forward.

Also, I deleted the committed reference docs because I think it makes more sense to generate these on-demand at the time the site is created. Otherwise, reference docs would need to be regenerated on every commit to the codebase, which is cumbersome. But I didn't go back to fix the process to generate them yet. See the notes on remaining work about this.

## Completed work

- Checkout the [cybage-blockly branch](https://github.com/Joe-Davis-Blockly/blockly/tree/cybage-blockly)
- Run `git reset --soft main` to keep all the changes but reset the git history
- Remove the `docs/reference/js` folder and add `docs/reference` to .gitignore since the reference docs will be auto-generated.
- Remove the `i18n` directory since we're not going to localize anything and this was all examples or test changes
- Revert some files that were clearly leftover example files or unused
- Revert some changes to script files that were irrelevant to docs changes
- Commit all the `docs/docs` files (codelabs, guides, and publications)
- Commit all the `docs/static` files (static resources like images)
    - Note that there are two folders here: `img` and `images`. This needs to be resolved in a future commit. 
- Revert `docs/theme/Footer` files 
    - These were [swizzled](https://docusaurus.io/docs/swizzling) but as far as I can tell no changes were made so we don't need them. If we did need to make a change, we could probably swizzle a more specific component instead of this one which ejects about 8 different react components. It's a lot of code that now we would have to maintain, all for some reason that I can't detect.
- Revert `docs/theme/DocSidebarCategory` files
    - These were swizzled as well. There are some changes but I don't know why. I think it may have to do with messing with the sidebar for the reference docs, so it's possible we need to revisit this when we come back around to the reference docs. For now, reverted.
- Commit google analytics tracking code
	- docs/src/theme/Root/index.js
	- docs/src/utils/tracking.js
	- Note that I have only cursorily reviewed this code and I don't totally understand if we actually need all of it. how much do we actually care what code people are copying? are we going to use this?
	- Also not entirely sure this is set up correctly, because the plugin docs say it is to be used in conjuction with the standard gtag plugin but i don't think that is configured
	- Will come back to this in a future commit, for now, no changes to these two files
	- Talked to joe and the person that wrote this warrantied the work so if it doesn't work we can get him to look at it, and we can ask him questions.
	- There is also some information in the handoff guide, but I'm not sure that what's written in the guide matches the code we have here.
- At this point I paused to turn blockly into a monorepo, then rebased the branch on that change. Moved the docs directory to be under `packages/docs` and also committed the basic workspace files (package.json) etc.

That's the end of the "structured" commits I made.

The following changes were made but not committed separately from the final "misc" commit that I think should be soft reset:

- Rename `tutorialSidebar` to `codelabsSidebar`
- In `homepagefeatures/index.js`: a global community of "passion" developers -> "passionate"

## Reference docs

This is a big enough chunk of work I'm calling it out separately.

For backstory, we started using api-extractor and api-documenter a couple of years ago. You can see some [context here](https://github.com/RaspberryPiFoundation/blockly/pull/6506). We had been using an older script that generated reference docs but it didn't work with typescript. We fixed a bunch of issues with the tsdoc in the source code, work that would have had to be done regardless of which tool we picked to generated the final docs pages.

We considered using TypeDoc but the team liked how api-documenter felt better, and we had some ambitions to use api-extractor to set up tests on our public API, which still hasn't happened. api-documenter did have some annoyances we worked around.

In the intervening years, it's become more clear that api-documenter has some problems that are unlikely to be fixed as we are still having to patch them and there's been no movement on the corresponding issue. Also, while they have an example site that uses docusaurus, it's not updated for docusaurus v3/4 and there's no solution for nicely generating the table of contents. There is a docusaurus plugin for api-documenter, but it hasn't been updated in the past 4 years and is unlikely to work with docusaurus v3/4 though I haven't actually tried it. The work that cybage did to try to get the generated docs working with docusaurus is fragile and may be a lot of work to maintain.

### Solution

I think it's worth trying to use TypeDoc now. There's a [docusaurus plugin](https://typedoc-plugin-markdown.org/plugins/docusaurus/quick-start) for it which is well-maintained and seems popular. I think it might be relatively plug-and-play given we've already done the work to fix the issues with our TsDoc in the source code.

### Alternative

Problems with current approach:
  - In the cybage branch, the reference docs were generated once and then committed.
  - In the desired approach, the reference docs will be generated when the site is built, and not committed. 
  - You can run `npm run docs` from within the `blockly` subpackage in order to generate the reference docs. The script probably needs to be changed to put them in the correct docs folder.
  - in `sidebars.js` I commented out the correct sidebar and put something else just to make the code compile since I had deleted all the reference docs. You'll want to fix that when you get here.
  - I'm not convinced by the code in `docs_tasks.js` but I haven't reviewed it all yet. I wrote the original code here to turn the `blockly.md` file generated by api-documenter into a `toc.yaml` file used by devsite for the table of contents. But the needs here are different.
  - I am not sure why they're doing all this work to convert these files to mdx. I think this is likely to be fragile and if we do need to actually do this it needs to be more clearly documented why. But I'd really prefer not to and just leave the files without doing all this transformation. Why can't they stay as markdown interpreted with commonmark instead of mdx? We already have the markdown mode set to `detect`.
  - also not sure why they're extracting the name like this, but it's not great. It seems like everything will be Sentencecasewithnospaces for long method names that are originally camelCase, which is confusing. 
  - When the files are generated they don't have any front matter; this code could be simpler if it didn't assume that there may or may not be any front matter present. This works if we don't commit the reference files and just generate them anew every time, which I think makes more sense.
  - there is an npm package that claims it can generate the sidebar file and some other stuff for using api-documenter with docusaurus. It is worth looking into whether we should use that (noting it hasn't been updated in a long time) or at least look into what it does to see if we're doing something similar or we should copy parts of it.
 - update api-documenter and api-extractor both to latest, update the api-documenter patch, and make sure in package.json we're no longer locking the package version (add the `^` back). also check if we have a dependabot config that says to stop updating this dependency, if so remove it. We couldn't update before because of the markdown in tables thing. But docusaurus is supposed to be able to handle that.

## Remaining work

- [x] the GitHub action needs to be changed (https://github.com/grega/blockly/commit/2b0e4c856a14c7355b5b31351ff7c747d6456895)
    - [x] The docs should be uploaded to a subdirectory of the GitHub pages like /docs.
    - [x] we don't want to update the site automatically because docs will be updated before the relevant changes are released. This should be manual and eventually tied into the release of blockly
    - [x] Wasn't sure why the action needed `fetch-depth: 0` at first, but it's because docusaurus is configured to display the last updated date
    - [x] hasn't been updated for monorepo structure and uses wrong branch name
- [x] toc.yaml seems incomplete - should be deleted? It doesn't seem like it's being used (https://github.com/grega/blockly/commit/15307fdc113d176749fccc33ec5f1aa7f5b20eb2)
- [x] ClassBlock seems weird to me. Do you really need a whole react component just to slap a class on something? Anyway, we tend to use the same classes over and over so it might be better to just have specific components for those use cases instead of this super generic component. (https://github.com/grega/blockly/commit/8ae59744db72926a39a125bfef50201bca8d1e84)
    - [x] Like for CodelabImages, we already have an Image component so why not use that instead? (https://github.com/grega/blockly/commit/50edae16ee3daa0e41dc44ff7ae78958530423a4)
    - [x] For displayTableHeader is it possible to just use a better css selector instead of having to supply a class name? We don't need to make tables look exactly like they used to. And I haven't looked into what these styles are doing to know if they're needed in the first place (https://github.com/grega/blockly/commit/ec19340c069358743577187ec1750a8d59dcd700)
    - [x] for compareBetter / compareWorse, we can use the built-in callout styles, there's a way to [specify a custom title](https://docusaurus.io/docs/markdown-features/admonitions#specifying-title) on them which is fine. We don't need to match the existing style exactly (https://github.com/grega/blockly/commit/97a3e414e9d9c777bccb6f5d21368ff6e1ec2933)
- [x] should `CodelabsCards` be in its own directory with the style.module.css file? because that's just in the top level components directory but the css seems only relevant to the codelabs (https://github.com/grega/blockly/commit/ec0b74b5d27f32fafe9a641e30a4490ff10136c6)
- [x] many of the css files have a bunch of commented out styles that should be cleaned up (https://github.com/grega/blockly/commit/9273512081cc2efe4fa3d6542c64c458f899fce8)
- [x] why is there both `css/custom.css` and `pages/index.module.css`? Not sure the difference between them and why some styles are in one vs the other
  - @grega: `css/custom.css` is for global custom styles, `pages/index.module.css` is for custom styles scoped to the homepage
- [x] custom.css needs some work. A ton of commented out styles. The theme variables haven't been changed except the primary color. Not sure all the footer styles need to be there (previously the footer was ejected/swizzled so maybe that's why they're here, but I unswizzled it so hopefully we can delete a lot of these) (https://github.com/grega/blockly/commit/350404844cdb9533f92a05b96113be4113f158c8)
- [x] consider removing the `/js` part of the reference docs path. We kept that to match the format of the old reference docs, where we had js, iOS, and android reference docs. But we haven't had those in a long time. Now is a good time to get rid of it. We can redirect from developers.google.com/blockly/reference/js/* to blockly.com/docs/reference/* (https://github.com/grega/blockly/commit/eb6f6c30e55d4d2a15170219f49b9433b91aa78d)
- [x] update `docusaurus.config.js` to correct values for repo and branch names, consider removing irrelevant comments (https://github.com/grega/blockly/commit/50591e03e07fe0d4947d780f8fafb39c4f3abc38)
- [x] update `robots.txt` to allow search engines to index it (https://github.com/grega/blockly/commit/ef3cea8fe4e2294b8ccea20d1c7e55ad8f0ccb63)
- [x] update .`prettierignore` and add prettier config so that handwritten md files get formatted but reference docs don't. all the code and config files in `docs/` package should also be formatted (https://github.com/grega/blockly/commit/231a9b1c8db98c14c50a9f0cba99261d91da8c76 and https://github.com/grega/blockly/commit/45168a66bbcea2c901096064a636d330664bb2f1)
- [x] why is there both `static/img` and `static/images`. Pick one. The former has some stuff we probably don't need in it (https://github.com/grega/blockly/commit/8839c16e0b78f2d2fd8af5646e28c14423569d9d)
- [ ] set up redirects from devsite
- [ ] set up GitHub pages to be served at blockly.com/docs
- [ ] make sure google analytics works (in particular we might need to configure the gtag info. Maribeth/Joe can give access to ga site)
- [ ] make sure the algolia search works, it's currently set up to index the wrong url.
- [ ] consider converting all this code to typescript (might not be worth it, but at least deserves some consideration)
- [x] fix sidebar rendering for generated reference docs
  - [x] also gitignore the `_reference.js` file generated by the reference docs build task? (it was originally checked in but I (@grega) can't see any reason to keep it)

(Edit: @grega turned list above into checkbox / task list)

grega:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9543

### Proposed Changes

Demo site: https://grega.github.io/blockly/

I've copied the task list from [9543](https://github.com/RaspberryPiFoundation/blockly/issues/9543) below, for clarity.

### Test Coverage

n/a

### Documentation

This is that 🙂 

### Tasks completed

- [x] GitHub Action updated (https://github.com/grega/blockly/commit/2b0e4c856a14c7355b5b31351ff7c747d6456895)
    - [x] docs uploaded to a subdirectory of the GitHub pages
    - [x] manual deployment of docs site (through GitHub Action workflow dispatch config)
    - [x] hasn't been updated for monorepo structure and uses wrong branch name
- [x] toc.yaml removed (https://github.com/grega/blockly/commit/15307fdc113d176749fccc33ec5f1aa7f5b20eb2)
- [x] ClassBlock removed (https://github.com/grega/blockly/commit/8ae59744db72926a39a125bfef50201bca8d1e84)
    - [x] CodelabImages migrated to use Image component (https://github.com/grega/blockly/commit/50edae16ee3daa0e41dc44ff7ae78958530423a4)
    - [x] fix displayTableHeader class application (https://github.com/grega/blockly/commit/ec19340c069358743577187ec1750a8d59dcd700)
    - [x] compareBetter / compareWorse replaced with Docusaurus admonitions (https://github.com/grega/blockly/commit/97a3e414e9d9c777bccb6f5d21368ff6e1ec2933)
- [x] `CodelabsCards` component moved to own dir (https://github.com/grega/blockly/commit/ec0b74b5d27f32fafe9a641e30a4490ff10136c6)
- [x] general CSS clean up (https://github.com/grega/blockly/commit/9273512081cc2efe4fa3d6542c64c458f899fce8)
- [x] custom.css cleanup and correct setting / use of color varialbes (https://github.com/grega/blockly/commit/350404844cdb9533f92a05b96113be4113f158c8)
- [x] `/js` removed from reference docs path (https://github.com/grega/blockly/commit/eb6f6c30e55d4d2a15170219f49b9433b91aa78d)
- [x] update `docusaurus.config.js` to correct values for repo and branch names, consider removing irrelevant comments (https://github.com/grega/blockly/commit/50591e03e07fe0d4947d780f8fafb39c4f3abc38)
- [x] update `robots.txt` to allow search engines to index it (https://github.com/grega/blockly/commit/ef3cea8fe4e2294b8ccea20d1c7e55ad8f0ccb63)
- [x] update .`prettierignore` and add prettier config so that handwritten md files get formatted but reference docs don't. all the code and config files in `docs/` package should also be formatted (https://github.com/grega/blockly/commit/231a9b1c8db98c14c50a9f0cba99261d91da8c76 and https://github.com/grega/blockly/commit/45168a66bbcea2c901096064a636d330664bb2f1)
- [x] de-duplicate `static/img` and `static/images` (https://github.com/grega/blockly/commit/8839c16e0b78f2d2fd8af5646e28c14423569d9d)
- [x] fix sidebar rendering for generated reference docs
  - [x] also gitignore the `_reference.js` file generated by the reference docs build task
- [x] set up GitHub pages to be served at blockly.com/docs (handled through env var) (https://github.com/grega/blockly/commit/bb89514ddae55771cf486cc563d133020fb32ff8)

### Tasks remaining

Are these best handled in this PR, or can / should they be done separately?

- [ ] set up redirects from devsite
- [ ] make sure google analytics works (in particular we might need to configure the gtag info. Maribeth/Joe can give access to ga site)
- [ ] make sure the algolia search works
- [ ] consider converting all this code to typescript (might not be worth it, but at least deserves some consideration)


grega:
@mikeharv might be best for Maribeth and I to review where this is at first, there's quite a bit going on 🙂

maribethb:
Overall this looks really great, thank you for your work on this! The things left in remaining tasks can def wait until after this is merged (also some of them will be handled outside github)

Remaining thoughts/questions:

- I'm also stumped by the QA failures. It seems like mocha couldn't even load the test file? I have no idea... I'm assuming these commands work locally?
- Will probably file a follow up bug to tackle the reference docs section of [this bug](https://github.com/RaspberryPiFoundation/blockly/issues/9543), you got it working well enough with api-documenter, but I am still skeptical of how much work it is to translate these all into mdx and I don't really understand why they're doing all of this manual transformation. Worth looking at how much work it would be to switch to TypeDoc but can happen later
- The biggest question I have is about how deploying to github pages works when we already deploy something else there. Will pages like the [playground](https://raspberrypifoundation.github.io/blockly/packages/blockly/tests/playground.html) continue to work? We have a different workflow that pushes directly to the gh-pages branch, I don't know how that interacts with also using the gh-pages action. I think ideally the docs would be at `raspberrypifoundation.github.io/blockly/docs` and the other existing stuff would be at its current location. you could test this locally by running `npm run updateGithubPages:staging` from `packages/blockly` in your fork which will deploy our normal github pages to your fork

maribethb:
I realized that I know these two different publishing processes will not work together because the gulp script we currently use for gh-pages force pushes to that branch and overwrites the existing contents there. So this will definitely need some more thought to reconcile those two processes

but we could submit the rest of this PR and work on that as a follow up especially if i need to pick this back up

grega:
Finally getting to the bottom of the test failures - mostly to do with some implicit differences in npm behaviour what with the two `package.json` files within the respective `packages/` dirs, related to hoisting & deduping, and installing packages to a root `node_modules` dir (shared between both packages). I've applied what I think are (hopefully) sensible fixes, but happy to look to tackle this differently if needed.

The other, related, issue with linting was the `packages/docs/package.json` installing different versions of eslint and Prettier to what the `packages/blockly/package.json` was defining, so I've tightened up the version pinning there to avoid future conflicts (but suspect this could be further improved / de-duped somehow; comment below...).

~~Another fix incoming and hopefully will be good to go...~~ All done 🎉 

And yep, that clash with the two workflows deploying to GH Pages will need a bit of a rethink, happy to talk it through once this PR is wrapped up.

package.json
```diff
@@ -23,6 +23,10 @@
     "@commitlint/cli": "^20.1.0",
     "@commitlint/config-conventional": "^20.0.0"
   },
+  "overrides": {
+    "eslint": "9.36.0",
+    "prettier": "3.6.2"
+  },
```
grega, line 29: Assuming this is an acceptable approach, would it benefit from some documentation perhaps? (ie. how to handle shared dependencies across packages)

packages/blockly/tests/mocha/webdriver.js
```diff
@@ -8,8 +8,32 @@
...
 const webdriverio = require('webdriverio');
+const fs = require('fs');
+const path = require('path');
 const {posixPath} = require('../../scripts/helpers');
 
+/**
+ * Ensure browser test imports that use ../../node_modules/* continue to work
+ * when npm hoists dependencies to the repository root node_modules dir
+ */
+function ensureWorkspaceNodeModulesLinks() {
```
maribethb, line 19: oh boy. this is a can of worms. we probably shouldn't be importing things directly out of node modules? Like, we should be letting the node module resolution algorithm handle this instead of us hardcoding the path?

obviously not going to fix that as part of this PR so thank you for this little patch. i have long suspected our mocha tests are not set up ideally and this is one more bit of evidence of that but i have never sat down to try to figure out how to do it more correctly

package.json
maribethb: hmm. I think this is okay, but I thought node was supposed to be able to handle differing versions in different subpackages. 

But I wonder if for lint and format it makes more sense to just have these at the top level only anyway, and just lint the entire repository at once? i'll have to check how other monorepos do this but this seems acceptable for now to unblock progress

maribethb (approved):

packages/blockly/tests/mocha/webdriver.js
grega: > we should be letting the node module resolution algorithm handle this

Yep agreed.

Shall I get an issue written up for tackling this? (along with one also for the prettier/eslint handling?)

maribethb: Sure, that would be great! Thanks! I think we can merge this one in the meantime

grega: Got it, will sort those, and get this merged...

Thanks for looking through all of this!

grega: https://github.com/RaspberryPiFoundation/blockly/issues/9735

package.json
grega: https://github.com/RaspberryPiFoundation/blockly/issues/9736

PR #9672, opened 2026-04-01
feat!: announce toasts via shared ARIA live region

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9667

### Proposed Changes

This PR updates the Toast utility to use the shared `announceDynamicAriaState()` live region instead of managing its own `aria-live` region on each toast.

Changes include:
- Removed `role="status"` and `aria-live` from the toast DOM element.
- Added a call to `announceDynamicAriaState()` when a toast is shown.
- Passed the toast message, status role, and configurable assertiveness to the shared live region.
- Removed the `Toast.Assertiveness` enum and replaced usage with `aria.LiveRegionAssertiveness`.

### Breaking Change
This work removes `Blockly.Toast.Assertiveness`. Users who have depended upon this will want to switch to the equivalent `Blockly.utils.aria.LiveRegionAssertiveness`.

### Reason for Changes

Previously, each toast managed its own `aria-live` region. This could lead to duplicate or inconsistent screen reader announcements and duplicated ARIA logic across components.

Using the shared `announceDynamicAriaState()` utility centralizes live region management, ensures consistent announcement behavior, and aligns toast announcements with other dynamic ARIA announcements in Blockly.

### Test Coverage

Tests were updated to reflect the new architecture:
- Verified that toast messages are announced through the shared live region.
- Verified that the live region is assigned the `status` role.
- Verified that assertiveness defaults to `polite` and respects the `assertiveness` option when provided.
- Verified that the toast element itself no longer has `aria-live` or `role="status"` to prevent duplicate announcements.


packages/blockly/core/toast.ts
```diff
@@ -45,7 +45,7 @@ export interface ToastOptions {
    * How prominently/interrupting the readout of the toast should be for
    * screenreaders. Corresponds to aria-live and defaults to polite.
    */
-  assertiveness?: Toast.Assertiveness;
+  assertiveness?: aria.LiveRegionAssertiveness;
```
BenHenning, line 48: This is technically an API breakage (at least for TS typing) since toast utilities are exported as part of the public Blockly API. I think the underlying JS may end up being the same but it would still introduce build breakages when upgrading to v13.

@gonfunko and @maribethb do you have any concerns with introducing this breakage? It's a trivial change but I'm assuming the more "proper" way to do this would be to deprecate the old field and introduce a new one that uses the other type.

It would obviously be easier just to keep the current type and translate between them.

packages/blockly/tests/mocha/toast_test.js
```diff
@@ -97,33 +98,48 @@ suite('Toasts', function () {
...
+      this.liveRegion.getAttribute('role'),
+      Blockly.utils.aria.Role.STATUS,
+    );
+    assert.equal(
+      this.liveRegion.getAttribute('aria-live'),
+      Blockly.utils.aria.LiveRegionAssertiveness.POLITE,
     );
   });
-
+
```
BenHenning: Extra space to delete for the linter to be happy.

If it helps I usually always run the following command before I push any changes:

```sh
npm run format && npm run lint:fix && npm run format:check && npm run lint
```

This runs both auto fix eslint and auto fix Prettier runs, then runs both to check to make sure they are passing (since it's sometimes possible for the auto-fix to miss stuff).

BenHenning (changes requested):
Thanks @mikeharv! Changes look good though I do have one concern that will need to be addressed before this can be merged. PTAL.

packages/blockly/tests/mocha/toast_test.js
mikeharv: Thanks for this tip!

packages/blockly/core/toast.ts
maribethb: I think this is ok, Toast was newly added so it's in our grace period for changes, and we can add this to the renaming script to get updated automatically. Just mark this as a breaking change and note in the PR description.

I definitely don't want to introduce a new field under a different name (that is introducing a bigger breaking change later for no benefit), I think the actual way to do this backwards-compatibly would have the type be `Toast.Assertiveness | aria.LiveRegionAssertiveness` but I don't think it's necessary in this case.

BenHenning: Ah, I always forget about the union typing thing. Good point, but sounds good if we're fine with this being breaking.

BenHenning (approved):
Thanks @mikeharv! No other concerns from me, this looks good!

PR #9673, opened 2026-04-01
feat: Add keyboard shortcut to perform an action on the currently focused element

### LINKED ISSUE #9620: Enter/Space on workspace should not open the toolbox
### Check for duplicates

### Description

Users found it confusing and kept accidentally opening the toolbox. Enter/space when focused on workspace should toast a hint instructing to use the arrow keys to navigate instead.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #9624: enter hint for blocks refinement
### Check for duplicates

### Description

if you press enter on a block, you get a hint toast telling you to press / to see shortcuts. If the block has inputs (or anything else that would be navigated to with the right arrow) the hint should tell you to use the right arrow to navigate the block.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #9624
* Fixes #9620

### Proposed Changes
This PR backports the Enter keyboard shortcut from the keyboard-experimentation repo that performs an action on the currently focused element. It adds an optional `performAction()` method to IFocusableNode which may be adopted by classes to gain support for Enter acting upon them.

### DIFF
--- a/packages/blockly/core/bubbles/mini_workspace_bubble.ts
+++ b/packages/blockly/core/bubbles/mini_workspace_bubble.ts
@@ -6,6 +6,7 @@
 
 import type {BlocklyOptions} from '../blockly_options.js';
 import {Abstract as AbstractEvent} from '../events/events_abstract.js';
+import {getFocusManager} from '../focus_manager.js';
 import {KeyboardMover} from '../keyboard_nav/keyboard_mover.js';
 import {Options} from '../options.js';
 import {Coordinate} from '../utils/coordinate.js';
@@ -286,5 +287,13 @@ export class MiniWorkspaceBubble extends Bubble {
       'The implementation of newWorkspaceSvg should be ' +
         'monkey-patched in by blockly.ts',
     );
+  }
+
+  /**
+   * Handles the user acting on this bubble via keyboard navigation by focusing
+   * the mutator workspace.
+   */
+  performAction() {
+    getFocusManager().focusTree(this.getWorkspace());
   }
 }
--- a/packages/blockly/core/bubbles/textinput_bubble.ts
+++ b/packages/blockly/core/bubbles/textinput_bubble.ts
@@ -279,6 +279,14 @@ export class TextInputBubble extends Bubble {
   getEditor() {
     return this.editor;
   }
+
+  /**
+   * Handles the user acting on this bubble via keyboard navigation by focusing
+   * the comment editor.
+   */
+  performAction() {
+    getFocusManager().focusNode(this.getEditor());
+  }
 }
 
 Css.register(`
--- a/packages/blockly/core/hints.ts
+++ b/packages/blockly/core/hints.ts
@@ -74,6 +74,8 @@ export function clearMoveHints(workspace: WorkspaceSvg) {
  */
 export function showHelpHint(workspace: WorkspaceSvg) {
   const shortcut = getShortActionShortcut('list_shortcuts');
+  if (!shortcut) return;
+
   const message = Msg['HELP_PROMPT'].replace('%1', shortcut);
   const id = helpHintId;
   Toast.show(workspace, {message, id});
--- a/packages/blockly/core/hints.ts
+++ b/packages/blockly/core/hints.ts
@@ -6,11 +6,15 @@
 
 import {Msg} from './msg.js';
 import {Toast} from './toast.js';
+import {getShortActionShortcut} from './utils/shortcut_formatting.js';
 import * as userAgent from './utils/useragent.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
 
 const unconstrainedMoveHintId = 'unconstrainedMoveHint';
 const constrainedMoveHintId = 'constrainedMoveHint';
+const helpHintId = 'helpHint';
+const blockNavigationHintId = 'blockNavigationHint';
+const workspaceNavigationHintId = 'workspaceNavigationHint';
 
 /**
  * Nudge the user to use unconstrained movement.
@@ -61,4 +65,38 @@ export function showConstrainedMovementHint(workspace: WorkspaceSvg) {
 export function clearMoveHints(workspace: WorkspaceSvg) {
   Toast.hide(workspace, constrainedMoveHintId);
   Toast.hide(workspace, unconstrainedMoveHintId);
+}
+
+/**
+ * Nudge the user to open the help.
+ *
+ * @param workspace The workspace.
+ */
+export function showHelpHint(workspace: WorkspaceSvg) {
+  const shortcut = getShortActionShortcut('list_shortcuts');
+  const message = Msg['HELP_PROMPT'].replace('%1', shortcut);
+  const id = helpHintId;
+  Toast.show(workspace, {message, id});
+}
+
+/**
+ * Tell the user how to navigate inside blocks.
+ *
+ * @param workspace The workspace.
+ */
+export function showBlockNavigationHint(workspace: WorkspaceSvg) {
+  const message = Msg['KEYBOARD_NAV_BLOCK_NAVIGATION_HINT'];
+  const id = blockNavigationHintId;
+  Toast.show(workspace, {message, id});
+}
+
+/**
+ * Tell the user how to navigate inside the workspace.
+ *
+ * @param workspace The workspace.
+ */
+export function showWorkspaceNavigationHint(workspace: WorkspaceSvg) {
+  const message = Msg['KEYBOARD_NAV_WORKSPACE_NAVIGATION_HINT'];
+  const id = workspaceNavigationHintId;
+  Toast.show(workspace, {message, id});
 }
--- a/packages/blockly/core/utils/shortcut_formatting.ts
+++ b/packages/blockly/core/utils/shortcut_formatting.ts
@@ -28,7 +28,7 @@ export function getShortActionShortcut(action: string): string {
 
 /**
  * Find the relevant shortcuts for the given action for the current platform.
- * Keys are returned in a long user facing format.
+ * Keys are returned in a long user facing format, e.g. "Command ⌘ Option ⌥ C"
  *
  * @internal
  * @param action The action name, e.g. "cut".
@@ -52,7 +52,7 @@ const shortModifierNames: Record<string, string> = {
 
 /**
  * Find the relevant shortcuts for the given action for the current platform.
- * Keys are returned in a user facing format.
+ * Keys are returned in a short user facing format, e.g. "⌘ ⌥ C"
  *
  * This could be considerably simpler if we only bound shortcuts relevant to the
  * current platform or tagged them with a platform.
--- a/packages/blockly/core/utils/shortcut_formatting.ts
+++ b/packages/blockly/core/utils/shortcut_formatting.ts
@@ -1,0 +1,138 @@
+/**
+ * @license
+ * Copyright 2026 Raspberry Pi Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Msg} from '../msg.js';
+import {ShortcutRegistry} from '../shortcut_registry.js';
+import * as userAgent from './useragent.js';
+
+/**
+ * Find the primary shortcut for this platform and return it as single string
+ * in a short user facing format.
+ *
+ * @internal
+ * @param action The action name, e.g. "cut".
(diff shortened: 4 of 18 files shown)


packages/blockly/core/bubbles/textinput_bubble.ts
```diff
@@ -279,6 +279,14 @@ export class TextInputBubble extends Bubble {
   getEditor() {
     return this.editor;
   }
+
+  /**
+   * Handles the user acting on this bubble via keyboard navigation by focusing
+   * the comment editor.
+   */
+  performAction() {
```
mikeharv, line 287: Any reason we're adding this to these `Bubble` subclasses instead of the base class itself?

packages/blockly/core/utils/shortcut_formatting.ts
```diff
@@ -0,0 +1,138 @@
...
+    shortcut.some(
+      (key) => 'Meta' === key || 'Alt' === key || 'Control' === key,
+    ),
+  );
+  const chosen = hasModifiers ? [currentPlatform[0]] : currentPlatform;
+  return chosen.map((shortcut) => {
+    return shortcut
+      .map((maybeNumeric) =>
+        Number.isFinite(+maybeNumeric)
+          ? String.fromCharCode(+maybeNumeric)
```
mikeharv, line 112: Could we document the expected numeric range for these strings? Are they always assumed to be character codes for letters/numbers?

packages/blockly/core/bubbles/mini_workspace_bubble.ts
```diff
@@ -287,4 +288,12 @@ export class MiniWorkspaceBubble extends Bubble {
         'monkey-patched in by blockly.ts',
     );
   }
+
+  /**
+   * Handles the user acting on this bubble via keyboard navigation by focusing
+   * the mutator workspace.
+   */
+  performAction() {
+    getFocusManager().focusTree(this.getWorkspace());
```
mikeharv, line 297: Is there a particular reason to add this to two `Bubble` subclasses as opposed to just the base class?

packages/blockly/core/utils/shortcut_formatting.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Find the relevant shortcuts for the given action for the current platform.
+ * Keys are returned in a long user facing format.
+ *
+ * @internal
+ * @param action The action name, e.g. "cut".
+ * @returns The formatted shortcuts as individual keys.
+ */
+export function getLongActionShortcutsAsKeys(action: string): string[][] {
```
mikeharv, line 37: Do we expect to use this for screen reader output?

packages/blockly/core/utils/shortcut_formatting.ts
```diff
@@ -0,0 +1,138 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Msg} from '../msg.js';
+import {ShortcutRegistry} from '../shortcut_registry.js';
+import * as userAgent from './useragent.js';
+
+/**
+ * Find the primary shortcut for this platform and return it as single string
+ * in a short user facing format.
```
mikeharv, line 13: For this and the function below it, could we document a simple example of each to make it obvious how the short/long formats differ?

packages/blockly/core/utils/shortcut_formatting.ts
```diff
@@ -0,0 +1,138 @@
...
+  'Meta': '⌘',
+  'Alt': userAgent.APPLE ? '⌥' : Msg['ALT_KEY'],
+};
+
+/**
+ * Find the relevant shortcuts for the given action for the current platform.
+ * Keys are returned in a user facing format.
+ *
+ * This could be considerably simpler if we only bound shortcuts relevant to the
+ * current platform or tagged them with a platform.
```
mikeharv, line 58: Since it is pretty complicated, could we documented the sorting priority up here too? I think it's like this:
1. Normalize modifier order (Meta, Control, Alt, Shift).
2. Prefer shortcuts that use Meta over Alt/Control variants.
3. Filter shortcuts to the current platform.
4. Prefer simpler shortcuts (fewer keys).
5. If any shortcuts use modifiers, assume they are platform variants and return only the best one.

packages/blockly/core/hints.ts
```diff
@@ -62,3 +66,37 @@ export function clearMoveHints(workspace: WorkspaceSvg) {
...
   Toast.hide(workspace, unconstrainedMoveHintId);
 }
+
+/**
+ * Nudge the user to open the help.
+ *
+ * @param workspace The workspace.
+ */
+export function showHelpHint(workspace: WorkspaceSvg) {
+  const shortcut = getShortActionShortcut('list_shortcuts');
```
mikeharv, line 76: Should we handle the potential case where `shortcut` is an empty string? I think it's possible that the shortcut could have been unregistered which would lead to a help prompt with a nonsensical message.

packages/blockly/msg/messages.js
```diff
@@ -1695,4 +1695,10 @@ Blockly.Msg.KEYBOARD_NAV_CONSTRAINED_MOVE_HINT = 'Use the arrow keys to move, th
 Blockly.Msg.KEYBOARD_NAV_COPIED_HINT = 'Copied. Press %1 to paste.';
 /** @type {string} */
 /// Message shown when an item is cut in keyboard navigation mode.
-Blockly.Msg.KEYBOARD_NAV_CUT_HINT = 'Cut. Press %1 to paste.';
\ No newline at end of file
+Blockly.Msg.KEYBOARD_NAV_CUT_HINT = 'Cut. Press %1 to paste.';
+/** @type {string} */
+/// Message shown when a user presses Enter with a navigable block focused.
```
mikeharv: Will translators need any more context? For example, is it still the right arrow key on LTR workspaces? (Genuine question I don't know the answer to.)

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -19,8 +19,8 @@ import {createKeyDownEvent} from './test_helpers/user_input.js';
 suite('Keyboard Shortcut Items', function () {
   setup(function () {
     sharedTestSetup.call(this);
-    const toolbox = document.getElementById('toolbox-categories');
-    this.workspace = Blockly.inject('blocklyDiv', {toolbox});
+    const toolbox = document.getElementById('toolbox-test');
+    this.workspace = Blockly.inject('blocklyDiv', {toolbox, renderer: 'zelos'});
```
mikeharv, line 24: Intentional to change this whole test suite to use Zelos?

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -553,22 +553,6 @@ suite('Keyboard Shortcut Items', function () {
   });
 
   suite('Focus Toolbox (T)', function () {
-    setup(function () {
```
mikeharv, line 631: Did this have a purpose for this test suite at some point?

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -799,4 +783,216 @@ suite('Keyboard Shortcut Items', function () {
...
+
+        await bubbleShown;
+        assert.strictEqual(
+          Blockly.getFocusManager().getFocusedNode(),
+          icon.getBubble(),
+        );
+      }
+    });
+
+    test('Triggers flyout button actions', function () {
```
mikeharv, line 1209: Good set of new tests!

mikeharv (commented):
This looks great. Thanks for your patience with my questions as some of this code is still somewhat new to me.

packages/blockly/core/bubbles/mini_workspace_bubble.ts
gonfunko: Yes, they're doing different things; the mini workspace bubble focuses its workspace and the text input bubble focuses its comment editor. Bubble is an abstract class and there's no common behavior, so each subclass does its own thing.

packages/blockly/core/bubbles/textinput_bubble.ts
gonfunko: Yes, see the other reply on the workspace bubble.

packages/blockly/core/utils/shortcut_formatting.ts
gonfunko: Yes, they're always character codes. The isFinite thing isn't to avoid `Infinity`, just to check if a string is secretly a number.

packages/blockly/core/hints.ts
gonfunko: That is a possibility, I'm not sure how we might helpfully handle it though. End users wouldn't be able to do anything about it, I suppose we could just not show the toast? Or did you have something else in mind?

packages/blockly/tests/mocha/shortcut_items_test.js
gonfunko: Yes, in order to test the behavior of full-block fields, which don't exist in other renderers. All the other tests continue to work fine with it, so I just changed it globally for the file.

packages/blockly/tests/mocha/shortcut_items_test.js
gonfunko: Yes, this block is in the old categories toolbox these tests were using, but is not in the test toolbox we're now using, so we don't need to define it any more.

packages/blockly/core/utils/shortcut_formatting.ts
gonfunko: Not exactly, at present it's not used for anything but I just brought back this whole file verbatim. It will be used in the dialog that displays all the available shortcuts, which isn't screenreader per se but would/can be read by a screenreader.

packages/blockly/core/utils/shortcut_formatting.ts
gonfunko: Done!

packages/blockly/core/utils/shortcut_formatting.ts
gonfunko: I think it's a bit better suited to inline documentation? The point is basically to just do the right/expected thing, such that the method returns what you'd expect. I don't think there's a lot of value in explaining it twice, especially since this is an internal-only method.

packages/blockly/msg/messages.js
gonfunko: I think we're OK, ideally this would interpolate the actually-bound shortcut like some of the others but that's something we can address later I think.

packages/blockly/core/utils/shortcut_formatting.ts
mikeharv: Aha! Makes sense. It seemed intentional, but I was having trouble guessing at the intention.

packages/blockly/core/hints.ts
mikeharv: I think it's reasonable that if the shortcut isn't available because it was unregistered, we could just return early and not show the toast. No toast is better than one that says "Press for help on keyboard controls". The only other option I can think of would be a fallback string that just says "invalid action" or something if we don't have a shortcut to suggest for help.

packages/blockly/tests/mocha/shortcut_items_test.js
mikeharv: Makes sense! Worth adding a comment to explain this?

mikeharv (approved):
Thanks for answering all of my questions. I think my biggest concern is the possibility of creating a toast with a broken/incomplete message if the shortcut has gone missing. This is hopefully more of an edge, but perhaps worth tracking with a new issue so we can remember to solve it at some point.

packages/blockly/core/hints.ts
gonfunko: Added an early return in that case. I also disabled the test as it now fails since the shortcut to show the list of shortcuts doesn't yet exist.

packages/blockly/tests/mocha/shortcut_items_test.js
gonfunko: Done!

PR #9677, opened 2026-04-02
feat: i shortcut on workspace gives overview

mikeharv:
## The basics

### Resolves

Fixes https://github.com/issues/assigned?issue=RaspberryPiFoundation%7Cblockly%7C9622#:~:text=give%20workspace%20overview-,%239622,-Edit

### Proposed Changes

This adds a new 'i' shortcut for workspaces. If the workspace is focused, the shortcut will trigger an announcement of the current total of block stacks and workspace comments. New translation strings were added to facilitate communicating the number of stacks and comments in a grammatically accurate way. The new strings were intentionally phrased to front-load the most important information (ie. actual quantity) over contextual description (ie. "on the workspace").
See tests for sample announcement strings created using the new messages.

### Reason for Changes

From linked issue:
>when focused on workspace, i shortcut should tell how many stacks of blocks, how many comments, etc are on the workspace. stacks are more useful than total block count.

### Test Coverage

The main new test performs a round trip through the different permutations of messages. The other test checks that the precondition logic works as expected (the workspace must be focused).

### Documentation

N/A

### Additional Information

I expect we'll want to make sure this new shortcut gets added to the keyboard navigation help menu/shortcuts list once that has been ported over.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -19,7 +19,9 @@ import {type IDraggable, isDraggable} from './interfaces/i_draggable.js';
 import {type IFocusableNode} from './interfaces/i_focusable_node.js';
 import {Direction, KeyboardMover} from './keyboard_nav/keyboard_mover.js';
 import {keyboardNavigationController} from './keyboard_navigation_controller.js';
+import {Msg} from './msg.js';
 import {KeyboardShortcut, ShortcutRegistry} from './shortcut_registry.js';
+import {aria} from './utils.js';
 import {Coordinate} from './utils/coordinate.js';
 import {KeyCodes} from './utils/keycodes.js';
 import {Rect} from './utils/rect.js';
@@ -53,6 +55,7 @@ export enum names {
   NAVIGATE_UP = 'up',
   NAVIGATE_DOWN = 'down',
   DISCONNECT = 'disconnect',
+  INFORMATION = 'information',
 }
 
 /**
@@ -635,20 +638,20 @@ export function registerArrowNavigation() {
   }
 }
 
+const resolveWorkspace = (workspace: WorkspaceSvg) => {
+  if (workspace.isFlyout) {
+    const target = workspace.targetWorkspace;
+    if (target) {
+      return resolveWorkspace(target);
+    }
+  }
+  return workspace.getRootWorkspace() ?? workspace;
+};
+
 /**
  * Registers keyboard shortcut to focus the workspace.
  */
 export function registerFocusWorkspace() {
-  const resolveWorkspace = (workspace: WorkspaceSvg) => {
-    if (workspace.isFlyout) {
-      const target = workspace.targetWorkspace;
-      if (target) {
-        return resolveWorkspace(target);
-      }
-    }
-    return workspace.getRootWorkspace() ?? workspace;
-  };
-
   const focusWorkspaceShortcut: KeyboardShortcut = {
     name: names.FOCUS_WORKSPACE,
     preconditionFn: (workspace) => !workspace.isDragging(),
@@ -687,6 +690,55 @@ export function registerFocusToolbox() {
     keyCodes: [KeyCodes.T],
   };
   ShortcutRegistry.registry.register(focusToolboxShortcut);
+}
+
+/**
+ * Registers keyboard shortcut to get count of block stacks and comments.
+ */
+export function registerWorkspaceOverview() {
+  const shortcut: KeyboardShortcut = {
+    name: names.INFORMATION,
+    preconditionFn: (workspace, scope) => {
+      const focused = scope.focusedNode;
+      return focused === workspace;
+    },
+    callback: (_workspace) => {
+      const workspace = resolveWorkspace(_workspace);
+      const stacks = workspace.getTopBlocks().length;
+      const comments = workspace.getTopComments().length;
+
+      // Build base string with block stack count.
+      let baseMsgKey;
+      if (stacks === 0) {
+        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_ZERO';
+      } else if (stacks === 1) {
+        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_ONE';
+      } else {
+        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_MANY';
+      }
+
+      // Build comment suffix.
+      let suffix = '';
+      if (comments > 0) {
+        suffix = Msg[
+          comments === 1
+            ? 'ARIA_WORKSPACE_COMMENTS_ONE'
+            : 'ARIA_WORKSPACE_COMMENTS_MANY'
+        ].replace('%1', String(comments));
+      }
+
+      // Build final message.
+      const msg = Msg[baseMsgKey]
+        .replace('%1', String(stacks))
+        .replace('%2', suffix);
+
+      aria.announceDynamicAriaState(msg);
+
+      return true;
+    },
+    keyCodes: [KeyCodes.I],
+  };
+  ShortcutRegistry.registry.register(shortcut);
 }
 
 /**
@@ -745,5 +797,13 @@ export function registerKeyboardNavigationShortcuts() {
   registerDisconnectBlock();
 }
 
+/**
+ * Registers keyboard shortcuts used to announce screen reader information.
+ */
+export function registerScreenReaderShortcuts() {
+  registerWorkspaceOverview();
+}
+
 registerDefaultShortcuts();
 registerKeyboardNavigationShortcuts();
+registerScreenReaderShortcuts();
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -707,32 +707,32 @@ export function registerWorkspaceOverview() {
     },
     callback: (_workspace) => {
       const workspace = resolveWorkspace(_workspace);
-      const stacks = workspace.getTopBlocks().length;
-      const comments = workspace.getTopComments().length;
+      const stackCount = workspace.getTopBlocks().length;
+      const commentCount = workspace.getTopComments().length;
 
       // Build base string with block stack count.
       let baseMsgKey;
-      if (stacks === 0) {
-        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_ZERO';
-      } else if (stacks === 1) {
-        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_ONE';
+      if (stackCount === 0) {
+        baseMsgKey = 'WORKSPACE_CONTENTS_BLOCKS_ZERO';
+      } else if (stackCount === 1) {
+        baseMsgKey = 'WORKSPACE_CONTENTS_BLOCKS_ONE';
       } else {
-        baseMsgKey = 'ARIA_WORKSPACE_BLOCKS_MANY';
+        baseMsgKey = 'WORKSPACE_CONTENTS_BLOCKS_MANY';
       }
 
       // Build comment suffix.
       let suffix = '';
-      if (comments > 0) {
+      if (commentCount > 0) {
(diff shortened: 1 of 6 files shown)


packages/blockly/core/shortcut_items.ts
```diff
@@ -745,5 +797,13 @@ export function registerKeyboardNavigationShortcuts() {
...
+/**
+ * Registers keyboard shortcuts used to announce screen reader information.
+ */
+export function registerScreenReaderShortcuts() {
+  registerWorkspaceOverview();
+}
+
 registerDefaultShortcuts();
 registerKeyboardNavigationShortcuts();
+registerScreenReaderShortcuts();
```
BenHenning, line 882: I think we actually want to keep this out, for now. The idea is we'll use an optional plugin to enable these based on user prompting.

packages/blockly/msg/messages.js
```diff
@@ -1695,4 +1695,23 @@ Blockly.Msg.KEYBOARD_NAV_CONSTRAINED_MOVE_HINT = 'Use the arrow keys to move, th
...
+/// ARIA live region message announcing there are no blocks in the workspace, optionally including comments.
+/// \n\nParameters:\n* %2 - optional phrase announcing comments, including leading space
+Blockly.Msg.ARIA_WORKSPACE_BLOCKS_ZERO = 'No blocks%2 in workspace.';
+/** @type {string} */
+/// ARIA live region phrase appended when there are multiple workspace comments.
+/// \n\nParameters:\n* %1 - the number of comments (integer greater than 1)
+Blockly.Msg.ARIA_WORKSPACE_COMMENTS_MANY = ' and %1 comments';
+/** @type {string} */
+/// ARIA live region phrase appended when there is exactly one workspace comment.
+Blockly.Msg.ARIA_WORKSPACE_COMMENTS_ONE = ' and one comment';
```
BenHenning: Nit: please re-add EOF line terminator.

packages/blockly/core/shortcut_items.ts
```diff
@@ -689,6 +692,55 @@ export function registerFocusToolbox() {
...
+  const shortcut: KeyboardShortcut = {
+    name: names.INFORMATION,
+    preconditionFn: (workspace, scope) => {
+      const focused = scope.focusedNode;
+      return focused === workspace;
+    },
+    callback: (_workspace) => {
+      const workspace = resolveWorkspace(_workspace);
+      const stacks = workspace.getTopBlocks().length;
+      const comments = workspace.getTopComments().length;
```
BenHenning: Nit: Perhaps use `stackCount` and `commentCount` for a bit more clarity on what's being contained in this variables (since they read a bit as arrays at the moment).

packages/blockly/msg/json/en.json
```diff
@@ -420,5 +420,10 @@
 	"KEYBOARD_NAV_UNCONSTRAINED_MOVE_HINT": "Hold %1 and use arrow keys to move freely, then %2 to accept the position",
 	"KEYBOARD_NAV_CONSTRAINED_MOVE_HINT": "Use the arrow keys to move, then %1 to accept the position",
 	"KEYBOARD_NAV_COPIED_HINT": "Copied. Press %1 to paste.",
-	"KEYBOARD_NAV_CUT_HINT": "Cut. Press %1 to paste."
+	"KEYBOARD_NAV_CUT_HINT": "Cut. Press %1 to paste.",
+	"ARIA_WORKSPACE_BLOCKS_MANY": "%1 stacks of blocks%2 in workspace.",
+	"ARIA_WORKSPACE_BLOCKS_ONE": "One stack of blocks%2 in workspace.",
+	"ARIA_WORKSPACE_BLOCKS_ZERO": "No blocks%2 in workspace.",
+	"ARIA_WORKSPACE_COMMENTS_MANY": " and %1 comments",
+	"ARIA_WORKSPACE_COMMENTS_ONE": " and one comment"
```
BenHenning: I'm finding this a bit hard to grok, and I suspect a translator might as well. Perhaps we could be slightly repetitive to simplify things. What if we avoided the double concatenation and had four different variables for the label:
- "%1 stacks of blocks and %2 comments in workspace."
- "One stack of blocks and %2 comments in workspace."
- "%1 stacks of blocks and one comment in workspace."
- "One stack of blocks and one comment in workspace."

Separately, I'm not sure about the `Msg` name here. These aren't actual ARIA labels or properties, they're just strings that happen to be used as a status message to trick the screen reader into announcing them I do recognize that we will want to standardize on this, though, so perhaps @maribethb might have a suggestion. One thought might be to do something like: `WORKSPACE_MANY_BLOCKS_AND_MANY_COMMENTS_SCREEN_READER_ANNOUNCEMENT_LABEL` but I'm not sure sure just how specific we want to get with these.

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -552,6 +552,81 @@ suite('Keyboard Shortcut Items', function () {
...
+        this.injectionDiv.dispatchEvent(keyEvent);
+        // Wait for the live region to update after the event.
+        this.clock.tick(11);
+        // The announcement may include an additional non-breaking space.
+        assert.include(this.liveRegion.textContent, expected);
+      };
+      this.liveRegion = document.getElementById('blocklyAriaAnnounce');
+    });
+
+    test('Announces block stacks and comments', function () {
```
BenHenning: I suggest splitting these up a bit. The Blockly team doesn't exactly follow the arrange-act-assert structure strictly in the codebase but it's a good baseline to follow particularly for longer tests like this as more granular failures can be more informative when investigating.

BenHenning (commented):
Thanks @mikeharv! Took a first pass and has some comments, PTAL.

Note that I'm not requesting changes in case you want to switch to a different reviewer since I'm actually out until next Wednesday.

packages/blockly/core/shortcut_items.ts
mikeharv: Ah, okay. With this shortcut, I was thinking it reasonable to always include since it won't really do much if you're using a screen reader.

maribethb: I don't think this is quite true. You shouldn't need a plugin just to enable a feature that's already in core.

We haven't really fleshed out exactly what will be enabled by default yet, but I think everything that doesn't impact the experience for non-screenreader users should be enabled by default, and then a select few things are added via a settings toggle (the sounds as you navigate the workspace, disable looping are the ones I can think of). That way a screenreader user who doesn't discover the additional features still gets as much utility as possible.

That being said feel free to keep this method and not call it yet, we can call it or not later.

packages/blockly/msg/json/en.json
maribethb: Yeah, I think going with the 4 separate messages is more straightforward for translators. I might add a 5th one which is "Workspace is empty" for that case. You mentioned you wrestled with a couple different options, what made you not choose this one?

Agree I wouldn't use `Aria` in the name, but I think we can come up with a shorter name than Ben's suggestion. The fact that it's primarily a label for screenreader announcement can be explained in the message comment, the same way we do for e.g. "Tooltip" and "Dropdown menu option". So I would suggest something like `WORKSPACE_CONTENTS_MANY_BLOCKS_MANY_COMMENTS` etc

maribethb: Actually, we discussed a bit more in real time and one of the things that the original proposal gives us that the 4 separate comments doesn't is the ability to leave off information about comments entirely. We actually don't include workspace comments by default in Blockly, as you have to enable the context menu item that lets you add them separately. Some products don't use them at all (e.g. code.org). So it would be confusing for those users to always be told that there are no workspace comments when they aren't even aware that those might exist.

So I think if we add more examples in the message description that the original design is workable. Also, Ben, had you been able to read the message descriptions when you commented here? Since you commented on en.js you don't see the descriptions in that file, but a translator would see them in the console.

packages/blockly/core/shortcut_items.ts
BenHenning: The problem I see is that this still uses up the keyboard binding, and I’m assuming makes populating the keyboard navigation help menu a bit trickier since it presumably needs to account for only showing these when the “screen reader mode” thing is enabled. My thinking was gating the actual registration of the shortcuts makes both cases easier, but perhaps you’re thinking of this differently?

packages/blockly/msg/json/en.json
BenHenning: I admittedly hadn’t looked at the descriptions since I reviewed this a bit quickly and figured they would change if we changed the format itself.

I agree with the benefit of omitting comments though the current implementation doesn’t make this distinction. We could perhaps just not include the word ‘comments’ if there are zero to account for that case and still benefit from having the four simpler messages. Could that work?

Pedagogically I don’t think specifying 0 comments is actually all that useful since comments are very much a secondary or tertiary user flow/concept.

mikeharv: >Pedagogically I don’t think specifying 0 comments is actually all that useful since comments are very much a secondary or tertiary user flow/concept.

Agreed. That's actually the main motivator of the current implementation here.

The current implementation skips the phrase `and n comments` entirely if there aren't any. Whether there are comments or not, we still need to be able to report that there are zero/one/many stacks. There could also be zero/one/many stacks. (Note that it's advantageous to avoid grouping 'zero' with the plural version, because not all languages work that way.)
I believe these are the permutations we'd need to support, assuming we're aligned with dropping the comments part entirely when there are zero:
- `[0, 0]` "Workspace is empty" (or "No blocks on workspace")
- `[1, 0]` "1 stack of blocks on workspace"
- `[2+,0]` "2 stack of blocks on workspace"
- `[0, 1]` "No blocks and 1 comment on workspace"
- `[1, 1]` 1 stack of block and 1 comment on workspace"
- `[2+,1]` "2 stacks of blocks and 1 comment on workspace"
- `[0, 2+]` "No blocks and 2 comments on workspace"
- `[1, 2+]` "1 stack of blocks and 2 comments on workspace"
- `[2+,2+]` "2 stacks of blocks and 2 comments on workspace"

It's not clear to me how we could do this with four simpler messages.

packages/blockly/core/shortcut_items.ts
maribethb: I am writing a document now about what we'll do about default options, so we don't need to debate it in this PR :) I'll share it with you when ready.

packages/blockly/msg/json/en.json
maribethb: The current implementation does make this distinction, as you can leave off the "and x comments" message.

> We could perhaps just not include the word ‘comments’ if there are zero to account for that case and still benefit from having the four simpler messages. Could that work?

That's what the current implementation does, but I'm not sure how you could do it with the messages you suggested earlier. Can you clarify?

edit: sorry did not see Mike's comment above when I wrote this

BenHenning: Yeah you’re both right that this can’t be done easier. Well I’m certainly not going to be insistent on it being simpler since what we’ll be doing for block labels is going to be much more complex than this and translators hopefully will be able to figure it out. Splitting up seems reasonable.

packages/blockly/core/shortcut_items.ts
```diff
@@ -692,6 +695,55 @@ export function registerFocusToolbox() {
   ShortcutRegistry.registry.register(focusToolboxShortcut);
 }
 
+/**
+ * Registers keyboard shortcut to get count of block stacks and comments.
+ */
+export function registerWorkspaceOverview() {
```
maribethb, line 701: I think that when we backport the existing `i` shortcut (which currently works only on blocks, and only in the screenreader experiment branch), it should be merged with this shortcut, so that there's only one "i" shortcut. That will make it easier to show in the keyboard shortcut help menu as it only needs to appear once.

But doesn't need to be changed for this PR, can be changed at the time we backport it. Since that will happen before v13 is merged I'm not worried about changing the name of the exported function here.

maribethb (approved):

PR #9680, opened 2026-04-03
fix: add group when cutting workspace comments

### LINKED ISSUE #9164: events not grouped correctly for cut/paste
### Check for duplicates

### Description

When a block or workspace comment (or other `ICopyable` thing) is cut or pasted, several events are fired at once:

for cut
- Selected (old: block id, new: null)
- Delete block or delete comment

for paste, or duplicate context menu option
- selected (old: whatever was previously selected, new: id of pasted thing)
- Create block or create comment

These two events should be grouped since they are part of the same user action. They are only grouped for the case of cutting a block specifically, because the cut callback calls a block's `checkAndDelete` method which handles event grouping. For the case of cutting other non-block types, or pasting anything including blocks, or using the duplicate context menu item, there is no group id set on the events fired in those cases.

### Reproduction steps

1. Enable event logging in the playground
2. Cut a block with ctrl+x and notice the events fired have a group id
3. Cut a workspace comment with ctrl-x and notice they don't

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

lizschwab:
## The basics

- [ x ] I [validated my changes](https://developers.google.com/blockly/guides/contribute/core#making_and_verifying_a_change)

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9164

### Proposed Changes

Adds an event group when the `cut` action is performed from a keyboard shortcut for a non-block entity.
image

### Reason for Changes

Without this functionality, events fired from the `cut` keyboard shortcut cannot be grouped together.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -217,7 +217,9 @@ export function registerCut() {
       if (focused instanceof BlockSvg) {
         focused.checkAndDelete();
       } else if (isIDeletable(focused)) {
+        eventUtils.setGroup(true);
         focused.dispose();
+        eventUtils.setGroup(true);
       }
       return !!copyData;
     },
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -219,7 +219,7 @@ export function registerCut() {
       } else if (isIDeletable(focused)) {
         eventUtils.setGroup(true);
         focused.dispose();
-        eventUtils.setGroup(true);
+        eventUtils.setGroup(false);
       }
       return !!copyData;
     },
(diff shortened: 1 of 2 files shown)


packages/blockly/core/shortcut_items.ts
```diff
@@ -217,7 +217,9 @@ export function registerCut() {
       if (focused instanceof BlockSvg) {
         focused.checkAndDelete();
       } else if (isIDeletable(focused)) {
+        eventUtils.setGroup(true);
         focused.dispose();
+        eventUtils.setGroup(true);
```
mikeharv: Should this second call be `setGroup(false)`? I see a similar pattern used in `registerDelete()` and in `registerCommentDelete()` (from `contextmenu_items.ts`).

lizschwab: Whoops, you're correct. Looks like I missed that when I had to re-apply the changes after fixing a git issue I was having. Thanks for catching that!

maribethb (approved):

PR #9683, opened 2026-04-07
feat: Add aria APIs to Field base class

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9654

### Proposed Changes

This PR introduces new ARIA helper APIs on the base `Field` class to support accessible label generation:

- Adds:
  - `getAriaTypeName(): string | null`
  - `getAriaValue(): string | null`
  - `computeAriaLabel(includeTypeInfo?: boolean): string`
- Extends `FieldConfig` with optional `ariaTypeName` to allow block definitions to configure ARIA type labels
- Implements safe default behavior in the base class:
  - `getAriaTypeName` returns configured value (if present)
  - `getAriaValue` falls back to `getValue()` stringification
  - `computeAriaLabel` composes type/value with optional verbosity
- Adds unit tests covering:
  - Default behavior
  - Config-driven type names
  - Label composition logic
  - Null/undefined edge cases
  - Subclass override behavior

These changes establish a consistent foundation for ARIA labeling across all fields.

### Reason for Changes

Blockly currently lacks a standardized way for fields to expose ARIA-friendly labels. 

### Test Coverage

Added a new Aria test suite covering:

- `getAriaTypeName`
  - Default (`null`)
  - JS constructor config
  - JSON config
- `getAriaValue`
  - Strings, numbers
  - `null` and `undefined`
- `computeAriaLabel`
  - Value-only output
  - Type + value composition
  - Type-only cases
  - Empty output when both are missing
- Subclass overrides
  - Custom `getAriaValue`
  - Custom `getAriaTypeName`
  - Combined overrides

Tests ensure correct fallback behavior and verify that subclasses can override these methods as intended.

### Documentation

Inline JSDoc comments were added for all new APIs.

No external documentation updates yet.

### Additional Information

Follow-up PRs are planned to adopt the new APIs across built-in fields.

### DIFF
--- a/packages/blockly/core/field.ts
+++ b/packages/blockly/core/field.ts
@@ -98,7 +98,8 @@ export abstract class Field<T = any>
   /** Validation function called when user edits an editable field. */
   protected validator_: FieldValidator<T> | null = null;
 
-  protected ariaTypeName_: string | null = null;
+  /** The ARIA-friendly label representation of this field's type. */
+  protected ariaTypeName: string | null = null;
 
   /**
    * Used to cache the field's tooltip value if setTooltip is called when the
@@ -253,7 +254,7 @@ export abstract class Field<T = any>
       this.setTooltip(parsing.replaceMessageReferences(config.tooltip));
     }
     if (config.ariaTypeName) {
-      this.ariaTypeName_ = config.ariaTypeName;
+      this.ariaTypeName = config.ariaTypeName;
     }
   }
 
@@ -315,7 +316,7 @@ export abstract class Field<T = any>
    *     unspecified.
    */
   getAriaTypeName(): string | null {
-    return this.ariaTypeName_;
+    return this.ariaTypeName;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/blockly/core/field.ts
```diff
@@ -300,6 +306,88 @@ export abstract class Field<T = any>
...
+
+  /**
+   * Gets an ARIA-friendly label representation of this field's value.
+   *
+   * Note that implementations should generally always override this value to
+   * ensure a non-null value is returned since the default implementation relies
+   * on 'getValue' which may return null, and a null return value for this
+   * function will prompt ARIA label generation to skip the field's value
+   * entirely when there may be a better contextual placeholder to use, instead,
+   * specific to the field.
```
maribethb, line 330: I think it might be useful here to include some examples of potential overrides, to more clearly illustrate the point. Something like:

> For example, a text input field may have a value of `null` if no text has been entered. To avoid hiding this field from screenreaders, implementations should ensure that if the value is `null`, this function would return an appropriate, localized, value such as "Empty text"

We'll have written docs eventually explaining what custom field authors need to do but having it here would probably help too.

maribethb (approved):

PR #9696, opened 2026-04-10
feat: Add basic support for generating ARIA labels and roles for blocks

### LINKED ISSUE #9685: Implement Block aria labeling
### Check for duplicates

### Problem

_No response_

### Request

Implement the logic that assembles the aria label for blocks

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9685

### Proposed Changes
This PR adds basic support for adding ARIA roles and labels to blocks. It backports the existing behavior from the screenreader branch, with heavy refactoring to improve readability and maintainability.

### Future Work
Labeling will be substantially improved by field subclasses customizing their labeling. Those changes will be automatically incorporated; this PR uses the newly-introduced APIs to get appropriate field labels.

Additionally, label generation should be tweaked once Inputs have support for overriding labeling to use those APIs.

When support for detailed readout keyboard shortcuts and "where am I" shortcuts are added, methods to compute labels for the full tree leading up to a block (specified in the design doc) should be added.


packages/blockly/core/block_aria_composer.ts
```diff
@@ -0,0 +1,319 @@
...
+ *
+ * The returned label will be specialized based on whether the block is part of a
+ * flyout.
+ *
+ * @internal
+ * @param block The block for which an ARIA representation should be created.
+ * @param verbosity How much detail to include in the description.
+ * @returns The ARIA representation for the specified block.
+ */
+export function computeARIALabel(
```
maribethb: I think this should be `computeAriaLabel` to match existing APIs, if you know of other ones where they're all caps we prob want to audit that before v13 release. 

same for other methods in this file.

style guide: https://google.github.io/styleguide/tsguide.html#camel-case

packages/blockly/core/utils/aria.ts
```diff
@@ -276,6 +276,15 @@ export enum State {
   VALUEMIN = 'valuemin',
 }
 
+/**
+ * Used to control how verbose generated a11y labels are.
+ */
+export enum Verbosity {
+  TERSE,
+  NORMAL,
```
maribethb: not to bikeshed too hard but i kinda think i prefer "standard" over "normal" as it feels less judgmental

packages/blockly/core/block_aria_composer.ts
```diff
@@ -0,0 +1,319 @@
...
+ * `lookback` attribute is specified, all of the fields on the row immediately
+ * above the Input will be used instead.
+ *
+ * Returns undefined if no fields precede the given Input.
+ *
+ * @internal
+ * @param input The Input to compute a description/context label for.
+ * @param lookback If true, will use fields on the previous row to compute a
+ *     label for the given input if it has no fields itself.
+ * @returns An accessibility label for the given input, or undefined if one
```
maribethb: This doesn't match reality, should this method join the array and return it as string?

packages/blockly/core/block_aria_composer.ts
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * Returns a label indicating that the block is a shadow block.
+ *
+ * @internal
+ * @param block The block to generate a label for.
+ * @returns A label indicating that the block is a shadow (if it is), otherwise
+ *     undefined.
+ */
+function getReplaceableLabel(block: BlockSvg) {
```
maribethb: I think it might be slightly better to name this `getShadowBlockLabel` to keep this consistent with the blockly library's vocabulary. the fact that the translation uses "replaceable" can be noted in a comment but is a bit more subject to change if for example we got user testing results that showed some other word made way more sense to people. but internally we're going to keep using shadow block. it's more explainable with the existing APIs.

packages/blockly/msg/messages.js
```diff
@@ -1724,3 +1724,49 @@ Blockly.Msg.KEYBOARD_NAV_BLOCK_NAVIGATION_HINT = 'Use the right arrow key to nav
...
+/// a statement block, i.e. that it has a next or previous connection.
+Blockly.Msg.BLOCK_LABEL_STATEMENT = 'statement';
+/** @type {string} */
+/// Part of an accessibility label for a block that indicates that it is
+/// a container block, i.e. that it has one or more statement inputs.
+Blockly.Msg.BLOCK_LABEL_CONTAINER = 'container';
+/** @type {string} */
+/// Part of an accessibility label for a block that indicates that it is
+/// a value block, i.e. that it has an output connection.
+Blockly.Msg.BLOCK_LABEL_VALUE = 'value';
```
maribethb: nit: add trailing newline (why doesn't our formatter do this????)

packages/blockly/core/block_svg.ts
```diff
@@ -1986,4 +1993,12 @@ export class BlockSvg
...
   }
+
+  /**
+   * Updates the ARIA label, role and roledescription for this block.
+   */
+  private recomputeARIAAttributes() {
+    aria.setState(this.getSvgRoot(), aria.State.LABEL, computeARIALabel(this));
+    configureARIARole(this);
+  }
 }
```
maribethb, line 2015: I think BlockSvg needs a method called `getAriaLabel` that internally calls the aria composer. It has to be public because we'll call it from various places such as shortcuts and plugins may also want to call it (imagine someone writes a plugin that adds some accessibility feature for blocks i can't think of right now). adding the method in BlockSvg will allow us to keep the aria composer fully internal as it will be totally transparent to anyone outside the library.

packages/blockly/core/block_aria_composer.ts
gonfunko: Doing that loses some of the commas in the full block readout. I think I'd kind of prefer to have this always return a list, and not export it?

packages/blockly/core/block_aria_composer.ts
gonfunko: Done

packages/blockly/core/utils/aria.ts
gonfunko: Done

packages/blockly/core/block_aria_composer.ts
gonfunko: Done

packages/blockly/msg/messages.js
gonfunko: Done

packages/blockly/core/block_svg.ts
gonfunko: Done

packages/blockly/core/block_aria_composer.ts
maribethb: That's fine, can you just update the jsdoc though, because right now it doesn't return a label or undefined

packages/blockly/core/block_svg.ts
```diff
@@ -1986,4 +1993,23 @@ export class BlockSvg
     // All other blocks are their own row.
     return this.id;
   }
+
+  /**
+   * Updates the ARIA label, role and roledescription for this block.
+   */
+  private recomputeARIAAttributes() {
```
maribethb: This should also be `recomputeAriaAttributes` per naming guidelines

packages/blockly/core/block_aria_composer.ts
gonfunko: Done!

maribethb (approved):

packages/blockly/core/block_svg.ts
gonfunko: Done

PR #9697, opened 2026-04-10
fix: Add default aria role to svg and group elements

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9664

### Proposed Changes

Adds `role=generic` to svg elements of type [g](https://github.com/RaspberryPiFoundation/blockly/blob/main/packages/blockly/core/utils/svg.ts#L50) and type [svg](https://github.com/RaspberryPiFoundation/blockly/blob/main/packages/blockly/core/utils/svg.ts#L64)

### Reason for Changes

This change ensures that browsers and screen readers do not attempt to associate additional semantic information with these elements.

### Test Coverage

Added unit tests which cover the following scenarios:
 - Creating svg elements with the name or nameTag 'g' or 'svg' results in the generic role being added
 - Passing in a specific role will override the default 'generic' role
 - Other element types are not affected by these changes

### DIFF
--- a/packages/blockly/core/utils/dom.ts
+++ b/packages/blockly/core/utils/dom.ts
@@ -6,6 +6,7 @@
 
 // Former goog.module ID: Blockly.utils.dom
 
+import * as aria from './aria.js';
 import type {Svg} from './svg.js';
 
 /**
@@ -56,6 +57,17 @@ export function createSvgElement<T extends SVGElement>(
   opt_parent?: Element | null,
 ): T {
   const e = document.createElementNS(SVG_NS, `${name}`) as T;
+  /**
+   * For svg and group (g) elements, we set the role to generic so that they are ignored by assistive technologies.
+   */
+  if (
+    name === 'svg' ||
+    name === 'g' ||
+    e.tagName === 'svg' ||
+    e.tagName === 'g'
+  ) {
+    aria.setRole(e, aria.Role.GENERIC);
+  }
   for (const key in attrs) {
     e.setAttribute(key, `${attrs[key]}`);
   }
--- a/packages/blockly/core/utils/dom.ts
+++ b/packages/blockly/core/utils/dom.ts
@@ -7,7 +7,7 @@
 // Former goog.module ID: Blockly.utils.dom
 
 import * as aria from './aria.js';
-import type {Svg} from './svg.js';
+import {Svg} from './svg.js';
 
 /**
  * Required name space for SVG elements.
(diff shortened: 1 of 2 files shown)


packages/blockly/core/utils/dom.ts
```diff
@@ -56,6 +57,17 @@ export function createSvgElement<T extends SVGElement>(
   opt_parent?: Element | null,
 ): T {
   const e = document.createElementNS(SVG_NS, `${name}`) as T;
+  /**
+   * For svg and group (g) elements, we set the role to generic so that they are ignored by assistive technologies.
+   */
+  if (
+    name === 'svg' ||
```
gonfunko: Rather than string literals, it might be preferable to use `Svg.G.tagName` and `Svg.SVG.tagName` (from utils/svg.ts)

packages/blockly/tests/mocha/utils_test.js
```diff
@@ -433,6 +433,51 @@ suite('Utils', function () {
...
+          Blockly.utils.Svg.G,
+          {},
+        );
+        const g = Blockly.utils.dom.createSvgElement('g', {});
+        assert.equal(svgG.getAttribute('role'), 'generic');
+        assert.equal(g.getAttribute('role'), 'generic');
+      });
+      test('svg elements of type svg have the generic role by default', function () {
+        const svgSvg = Blockly.utils.dom.createSvgElement(
+          Blockly.utils.Svg.G,
```
gonfunko: This should be `Blockly.utils.Svg.SVG` I think?

gonfunko (approved):
Just a couple minor nits but looks good, thanks!

packages/blockly/tests/mocha/utils_test.js
lizschwab: Yep, thanks for catching that!

packages/blockly/core/utils/dom.ts
```diff
@@ -6,7 +6,8 @@
 
 // Former goog.module ID: Blockly.utils.dom
 
-import type {Svg} from './svg.js';
+import * as aria from './aria.js';
+import {Svg} from './svg.js';
```
lizschwab, line 10: In order to use `Svg` as a value below I needed to change the import statement. This does not appear to affect anything, but I wanted to highlight it as I'm unsure of the reasons we were previously importing it as a type.

packages/blockly/core/utils/dom.ts
```diff
@@ -56,6 +57,17 @@ export function createSvgElement<T extends SVGElement>(
   opt_parent?: Element | null,
 ): T {
   const e = document.createElementNS(SVG_NS, `${name}`) as T;
+  /**
+   * For svg and group (g) elements, we set the role to generic so that they are ignored by assistive technologies.
+   */
+  if (
+    name === Svg.SVG.toString() ||
```
lizschwab, line 64: `Svg.SVG.tagName` is private, but `toString()` returns the `tagName`.

packages/blockly/core/utils/dom.ts
gonfunko: Generally we just import the type alone if it's being used strictly for TS typings, e.g. never instantiated/having any fields accessed. It's fine to import the full thing when we are actually using it though.

packages/blockly/core/utils/dom.ts
gonfunko: Ah good point, but yes this is fine, thanks!

gonfunko (approved):

PR #9699, opened 2026-04-10
fix: Fix bug that caused blocks inserted via Enter to not attach

gonfunko:
## The basics

### Resolves
This PR fixes a bug where inserting a block from the flyout using Enter (but not M) would insert it in move mode but with no potential candidate connections, meaning it could only be positioned via unconstrained move mode. The behavior is now identical whether you're using M or Enter. I also updated a test to ensure that inserting a block with Enter results in an insertion marker being displayed, meaning that the insertion is in constrained move mode.


mikeharv (approved):
Good find! Thanks for adding coverage too.

PR #9701, opened 2026-04-10
chore!: remove unused statuses

mikeharv:
Reverts RaspberryPiFoundation/blockly#9700

This is a resubmission of https://github.com/RaspberryPiFoundation/blockly/pull/9698 with the only difference being that it correctly marks the breaking change.

### DIFF
--- a/packages/blockly/core/utils/aria.ts
+++ b/packages/blockly/core/utils/aria.ts
@@ -123,19 +123,13 @@ export enum State {
    *
    * Value: one of {true, false}.
    */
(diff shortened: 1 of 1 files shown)


gonfunko:
### Breaking Change

If you were using a removed `aria.State`, we recommend using normal HTML/JS APIs instead of the Blockly utility functions and enums, e.g. `foo.setAttribute('aria-label', 'accessibility label goes here')`.

gonfunko (approved):

PR #9718, opened 2026-04-13
fix: initialize orphaned blocks on failed connection check during des…

### LINKED ISSUE #9266: XML deserializes weirdly on failed connection

periperidev:
Body:

  ## The basics

  ## The details
  ### Resolves

  Fixes https://github.com/google/blockly/issues/9266

  ### Proposed Changes

  When a child block fails to connect to its parent during deserialization (e.g. due to a type check mismatch), the block
  becomes an invisible "ghost" — it exists in the workspace data but has no SVG.

  This PR fixes both deserialization paths:

  - **XML path** (`core/xml.ts`): After initializing descendants, detect orphaned blocks (new top-level blocks that aren't
  descendants of the top block) and call `initSvg()` on them.
  - **JSON path** (`core/serialization/blocks.ts`): Log a warning instead of throwing `BadConnectionCheck` for non-shadow
  blocks, allowing `initBlock()` to run. Shadow blocks still throw.

  Both paths mark orphaned blocks as disabled via `setDisabledReason(true, 'orphaned_connection_check')`.

  ### Reason for Changes

  When block definitions change (e.g. an input type changes from `Long` to `Int`), saved workspaces may contain blocks with
  stale connection types. Currently these blocks become invisible ghosts that trigger validation errors. After this fix, they
  appear as visible, disabled, detached top-level blocks.

  ### Test Coverage

  - 2 new tests in `xml_test.js` — rendered and headless workspace orphan initialization
  - Updated "Bad checks" test in `jso_deserialization_test.js` to assert orphan behavior instead of throw
  - 1 new rendered workspace SVG test in `jso_deserialization_test.js`
  - All existing mocha tests pass (0 failures)

### DIFF
--- a/packages/blockly/core/serialization/blocks.ts
+++ b/packages/blockly/core/serialization/blocks.ts
@@ -601,18 +601,27 @@ function tryToConnectParent(
 
   if (!connected) {
     const checker = child.workspace.connectionChecker;
-    throw new BadConnectionCheck(
-      checker.getErrorMessage(
-        checker.canConnectWithReason(childConnection, parentConnection, false),
-        childConnection,
-        parentConnection,
-      ),
-      parentConnection.type === inputTypes.VALUE
-        ? 'output connection'
-        : 'previous connection',
-      child,
-      state,
+    const reason = checker.getErrorMessage(
+      checker.canConnectWithReason(childConnection, parentConnection, false),
+      childConnection,
+      parentConnection,
     );
+    if (child.isShadow()) {
+      throw new BadConnectionCheck(
+        reason,
+        parentConnection.type === inputTypes.VALUE
+          ? 'output connection'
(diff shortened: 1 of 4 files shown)


maribethb:
Hi, thank you for this PR. While we appreciate your work on it, we don't think this is the correct approach for this issue. Throwing an error upon trying to load invalid serialized state is correct, and the XML system should be changed to do this rather than the JSON system changing to avoid throwing. There is more reasoning in [this comment](https://github.com/RaspberryPiFoundation/blockly/issues/9266#issuecomment-4253893302). Feel free to respond in the bug if you have more questions. Thanks!

PR #9722, opened 2026-04-13
fix: Fix positioning of move marker on blocks

### LINKED ISSUE #9702: [v13.0.0-beta.0] Move mode icon position issues
### Check for duplicates

### Description

https://github.com/RaspberryPiFoundation/blockly/pull/9699 addresses the more fundamental issue with inserting a block from the flyout via keyboard. However, the move mode icon is initially positioned on the top-left of the block. On the first arrow key press (even if this doesn't move the block, i.e., no constrained positions) moves the icon to the top-right of the block.

For blocks already on the workspace, the move icon's horizontal position is dependent on the widest block in the stack that is being moved which can leave it detached from the block that the move was triggered on. See screen shot below.

### Reproduction steps

1. Check out the Blockly v13 tag
2. `npm start` from the relevant dir
3. Insert a block from the flyout via keyboard and observe move mode icon position
4. Arrow key press
5. Observe move mode icon position

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots
Image

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9702

### Proposed Changes
This PR resolves two issues with move indicator placement:
* The initial placement now waits for a render to complete to avoid the position changing
* For blocks, the indicator is anchored to the rightmost point of the topmost block

### DIFF
--- a/packages/blockly/core/keyboard_nav/keyboard_mover.ts
+++ b/packages/blockly/core/keyboard_nav/keyboard_mover.ts
@@ -3,14 +3,16 @@
  * Copyright 2026 Raspberry Pi Foundation
  * SPDX-License-Identifier: Apache-2.0
  */
+import type {BlockSvg} from '../block_svg.js';
 import type {IDraggable} from '../interfaces/i_draggable.js';
 import type {IDragger} from '../interfaces/i_dragger.js';
+import type {IFocusableNode} from '../interfaces/i_focusable_node.js';
 import * as registry from '../registry.js';
+import * as renderManagement from '../render_management.js';
 import {ShortcutRegistry} from '../shortcut_registry.js';
 import {Coordinate} from '../utils/coordinate.js';
 import {KeyCodes} from '../utils/keycodes.js';
 import {MoveIndicator} from './move_indicator.js';
-
 /**
  * Cardinal directions in which a move can proceed.
  */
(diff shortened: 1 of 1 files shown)


packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -263,12 +265,68 @@ export class KeyboardMover {
    * Repositions the move indicator to the corner of the item being moved.
    */
   private repositionMoveIndicator() {
-    const bounds = this.draggable?.getBoundingRectangle();
-    if (!bounds) return;
+    renderManagement.finishQueuedRenders().then(() => {
+      let bounds = this.draggable?.getBoundingRectangle();
+      if (
+        this.draggable &&
```
mikeharv, line 271: Is it possible that `this.draggable` could change before the async promise resolves? It might be worth using a local variable to capture this before the async call to be sure we're looking at the right block.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
```diff
@@ -263,12 +265,68 @@ export class KeyboardMover {
...
+
+    // Work backwards from the rightmost block; deeply nested value blocks do
+    // not have the same y position as their parent because they are visually
+    // depicted as being inside of it. Keep working up the parent block
+    // hierarchy until one is found with the same y position as the starting
+    // block, meaning is is the rightmost top-level value block in the same row
+    // as the starting block.
+    while (
+      targetBlock?.getBoundingRectangleWithoutChildren().getOrigin().y !==
+      block.getBoundingRectangleWithoutChildren().getOrigin().y
```
mikeharv: nit: Could we cache this before the loop so we don't have to recalculate each time through the loop?

mikeharv: We could also make a `getY` helper to possibly make this more readable.

mikeharv (approved):
A couple small suggestions, but this looks like a good solution!

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: No, it's only mutable by starting a move, and you can't do that while a move is in progress.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
gonfunko: Split it out of the loop, which also improved the readability.

packages/blockly/core/keyboard_nav/keyboard_mover.ts
mikeharv: Oh right. Thanks for explaining.

mikeharv (approved):

Issue #9497, opened 2025-12-02
Clarify which input a block is being moved into


gonfunko:
Currently, when a block is in constrained move mode, potential destinations are identified by reading out the proposed-parent block's label. In cases where the parent block has multiple (value or statement) inputs however, the input in question is not identified, which makes it challenging to place blocks. This should be improved, perhaps by using the field row label, index of the input or another method.

BenHenning:
From @microbit-robert:

> This is specifically a move mode issue, where it's not clear at the moment which branch of an `if` statement you are moving a block into. I can provide a diff of what I've done here shortly.
>
> These are the changes we've made for move mode for user testing in isolation which includes handling the specific branch for if / else blocks [https://github.com/RaspberryPiFoundation/blockly/compare/add-screen-reader-support-experimental...microbit-matt-hillsdon:blockly:move-mode-changes](https://www.google.com/url?q=https://github.com/RaspberryPiFoundation/blockly/compare/add-screen-reader-support-experimental...microbit-matt-hillsdon:blockly:move-mode-changes&sa=D&source=docs&ust=1764882845570470&usg=AOvVaw1BtY6AnULU9iI8hcfABXQe)
>
> This actually works better for MakeCode's if blocks than the Blockly test page if blocks. As the MakeCode blocks don't have the "do" text preceding the statement inputs, we use the text on the previous row. In the Blockly test page case, all 'else if' statements are labelled as "do", which can still be confusing.

joreneng:
Adding a note that this was also a pain point with all of the user testers in this testing round.

jversloot:
This was also an issue in our user tests with users moving their focus into a block and onto number inputs, and only being read out the value of the number, and not a title of what field they are focused on. Not sure if this is the same issue as when in move mode so I wanted to bring it up but can log it separately if it is different.

maribethb:
Also relevant to #9686

PR #9731, opened 2026-04-17
feat: Screenreader announcements for move mode

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9686

### Proposed Changes

This PR introduces ARIA live region announcements for moving blocks (both keyboard-driven move mode and standard mouse/pointer drags).

Key changes include:
- Added `computeMoveLabel` to generate more descriptive and context-aware move announcements.
- Introduced dynamic preposition selection (`before`, `after`, `inside`, `around`) based on connection types.
- Added disambiguation logic for cases where multiple compatible connections exist, including:
  - Input label subsets (e.g. distinguishing between `if`, `else if`, `else` statement inputs)
  - Existing field-based labels (e.g. operator text like `=`)
  - Index-based fallbacks for unlabeled inputs (e.g. value inputs on `text_join` blocks)
- Added `BlockSvg.getStackBlocksCountLabel` to announce stack size when moving multiple connected "stack blocks".
- Integrated the new announcement logic into `BlockDragStrategy`.
- Updated `Field.getAriaValue` to use `getText` instead of `getValue`, improving readout of some custom fields.
- Updated `getParentInputLabel` to ignore parent blocks if they are the insertion marker for the moving block
- Refactored `getInputLabels` to use a new `Input.getLabel` method which now also correctly ignore insertion markers.

If a block is moving without a candidate connection we will simply announce `moving [block label] on workspace`. It was decided that omitting cardinal directions is beneficial here as the user likely knows which arrow keys they have pressed (or which way they are dragging).
If a user cancels a movement, this is also announced. There is not currently an announcement when a movement is completed (ie. `endDrag()`).
Other announcement labels follow this format:
`Moving [block label] [disambiguation info] *preposition* [target block label] [disambiguation info]`
Disambiguation info is of course optional, as is the block label for the moving block if the user has already been moving (in `drag()` as opposed to `startDrag()`). Only the preposition and the target connection's block label are always included. "Before" and "after" are exclusively reserved for stacked blocks while "inside" and "around" are use for input/output connections and statement inputs.

Examples:
|`startDrag()`|`drag()`|
|---|---|
| image | image |
| `moving [print “ def ”] after [print “ abc ”]`<br>(Candidate found on start of move) | `moving [] before [print “ abc ”]`<br>(Block label omitted for successive moves) |
| image | image |
| `moving [if, do] inside [repeat 10, times, do] []` | `moving [] [] around [repeat 10, times, do]` |

|image|
|---|
|```moving [2 stack blocks] inside [if false, do, else if true, do, else if false or false, do, else] [else if false or false, do]```|

Note: Sometimes the block labels feel exceedingly verbose, especially when we need to iterate a portion to disambiguate the connection. As a follow-up, I think it could be worth being able to compute a terser label for a block with multiple connections.

### Reason for Changes

This provides an improved version of screen reader announcements for move mode that was based on micro:bit's testing. Key improvements include:
- correct prepositional descriptions
- terser announcements (often omitting superfluous block labels)
- disambiguation labels for multiple compatible connections with a target block.

### Test Coverage

Added and extended tests in the keyboard movement test suite to verify ARIA announcements.

Coverage includes:
- Announcements when starting a move (single block and multi-block stacks)
- Correct prepositions for different connection types (`before`, `after`, `inside`, `around`)
- Announcements when no connection candidate is available (workspace movement)
- Disambiguation behavior when multiple compatible connections exist:
  - Statement inputs (e.g. `if`, `else if`, `else`)
  - Value inputs using (e.g. in `logic_compare` blocks)
  - Fallback to input index for unlabeled inputs (e.g. in `text_join`)
- Ensures announcements include expected labels and prepositions and exclude unnecessarily verbose labels.

Tests use real keyboard movement flows and assert against the ARIA live region output.

### Documentation

TSDocs were added/updated as needed.

### Additional Information

Tests were designed to be resilient to minor wording or localization changes by asserting substrings rather than full string equality. This should make the tests less brittle as we potentially iterate on block and field labels.


packages/blockly/core/field.ts
```diff
@@ -323,31 +323,27 @@ export abstract class Field<T = any>
...
-    const value = this.getValue();
-
-    if (value === null || value === undefined) {
+    const fieldText = this.getText();
+    if (fieldText === 'null' || fieldText === '') {
       return null;
     }
-
-    return String(value);
+    return this.getText();
```
gonfunko: Probably should just return `fieldText` here?

packages/blockly/core/block_aria_composer.ts
```diff
@@ -15,6 +16,18 @@ import {
 import {Msg} from './msg.js';
 import {Role, setRole, setState, State, Verbosity} from './utils/aria.js';
 
+/**
+ * Prepositions to use when describing the relationship between two blocks based
+ * on their connection types.
+ */
+export enum ConnectionPreposition {
```
gonfunko, line 23: Minor nit: maybe make `UNKNOWN` first so it gets mapped to 0 and becomes falsey?

packages/blockly/core/block_aria_composer.ts
```diff
@@ -172,21 +186,58 @@ function getBeginStackLabel(block: BlockSvg) {
...
+ *
+ * @internal
+ * @param block The block to retrieve a list of field/input labels for.
+ * @param input The input that defines the end of the subset.
+ * @returns A list of field/input labels for the given block.
+ */
+export function getInputLabelsSubset(block: BlockSvg, input: Input): string[] {
+  const inputIndex = block.inputList.indexOf(input);
+  if (inputIndex === -1) {
+    throw new Error('Input not found on block.');
```
gonfunko: Might be handy to include the IDs of the input and the block in this

packages/blockly/core/block_aria_composer.ts
```diff
@@ -172,21 +186,58 @@ function getBeginStackLabel(block: BlockSvg) {
...
+  }
+
+  const targetType = input.type;
+
+  // Find the last index before inputIndex with the same type
+  let startIndex = 0;
+  for (let i = inputIndex - 1; i >= 0; i--) {
+    if (block.inputList[i].type === targetType) {
```
gonfunko: I'm a little unclear on the root goal of this, but...should this be a `!==`? Otherwise it seems like if we start from the last input of a given type in a sequence of that type, the preceding one's type will match, and we'll break out with startIndex set to the index of the input this method was called with?

packages/blockly/core/block_aria_composer.ts
```diff
@@ -250,6 +301,122 @@ function getParentToolboxCategoryLabel(block: BlockSvg) {
...
+    (describeNeighbour &&
+      neighbourInput &&
+      getInputLabelsSubset(neighbourBlock, neighbourInput).join(', ')) ||
+    '';
+
+  return announcementTemplate
+    .replace('%1', `[${blockLabel}]`)
+    .replace('%2', `[${localConnLabel}]`)
+    .replace('%3', `[${neighbourBlockLabel}]`)
+    .replace('%4', `[${neighbourConnLabel}]`);
```
gonfunko: Is the inclusion of the brackets here intentional or a debugging artifact?

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -939,6 +940,232 @@ suite('Keyboard-driven movement', function () {
...
+            assert.include(text, expected);
+            const index = text.indexOf(expected);
+            text =
+              text.slice(0, index) +
+              text.slice(index + expected.toString().length);
+          });
+        };
+        this.getBlockLabel = (block) =>
+          block.getAriaLabel(Blockly.utils.aria.Verbosity.TERSE);
+        this.element.dispose();
```
gonfunko: Is this necessary? I think element is only set inside of other suites?

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -939,6 +940,232 @@ suite('Keyboard-driven movement', function () {
...
+        this.block1 = this.workspace.newBlock('draw_emoji');
+        this.block1.initSvg();
+        this.block1.render();
+      });
+
+      test('announces simple block moving on workspace', async function () {
```
gonfunko: This and the next test don't appear to actually be async

gonfunko:
This also fixes #9497

packages/blockly/tests/mocha/keyboard_movement_test.js
mikeharv: changing to `this.workspace.clear()` higher up

packages/blockly/core/block_aria_composer.ts
mikeharv: Will add `input.name` to the message.

packages/blockly/core/block_aria_composer.ts
```diff
@@ -213,19 +211,13 @@ export function getInputLabels(block: BlockSvg): string[] {
 export function getInputLabelsSubset(block: BlockSvg, input: Input): string[] {
   const inputIndex = block.inputList.indexOf(input);
   if (inputIndex === -1) {
-    throw new Error('Input not found on block.');
+    throw new Error(`Input with name "${input.name}" not found on block.`);
```
gonfunko: Can you add the block ID as well?

gonfunko (approved):

PR #9732, opened 2026-04-17
fix: Improve navigation and movement looping behavior

### LINKED ISSUE #9706: [v13.0.0-beta.0] Workspace position for blocks with constrained move positions
### Check for duplicates

### Description

When moving a block with constrained move positions using keyboard, the block now visits the workspace. When the block visits the workspace:

- the KEYBOARD_NAV_UNCONSTRAINED_MOVE_HINT is always shown
- the block doesn't actually move which is quite confusing
    - blocks in the background may move as a result of the connection preview no longer being shown, however, in some cases this is very subtle
    - in the video below, it is unclear why the initial move position is different when the block is first 'popped out'

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/83d27cb1-3ada-4722-87f2-f537fe4b0b05

### Browsers

_No response_

### LINKED ISSUE #9707: [v13.0.0-beta.0] Keyboard looping configuration
### Check for duplicates

### Description

The current keyboard navigation behaviour is set to **not** loop, but this is not expected to be the default behaviour.

In contrast, the constrained move mode does loop. The looping behaviour should be consistent across navigation and move mode.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

* Fixes #9706
* Fixes #9707

### Proposed Changes
This PR makes several improvements around keyboard-driven movement and navigation:

* Looping now defaults to on
* The prompt to try unconstrained move mode when a block is moved to the workspace as a top-level block in constrained mode is now only shown once per session
* When blocks are moved to top-level blocks during constrained mode, their position is offset from their prior position to make it clearer that the proposed drop location/state has changed
* Constrained move mode now respects the keyboard navigation looping setting

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -456,7 +456,7 @@ export class BlockDragStrategy implements IDragStrategy {
       // suggest using unconstrained mode to arbitrarily position the block if
       // we're in keyboard-driven constrained mode.
       if (this.moveMode === MoveMode.CONSTRAINED) {
-        showUnconstrainedMoveHint(this.workspace, true);
+        showUnconstrainedMoveHint(this.workspace);
         this.workspace.getAudioManager().playErrorBeep();
       }
     }
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -811,7 +811,7 @@ export class BlockDragStrategy implements IDragStrategy {
   /**
    * Get the nearest valid candidate connection in traversal order.
    *
-   * @param direction The cardinal direction in which the block is being moved.
+   * @param delta The distance the block has moved since this drag began.
    * @returns A candidate connection and radius, or null if none was found.
    */
   findTraversalCandidate(delta: Coordinate): ConnectionCandidate | null {
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -601,15 +601,23 @@ export class BlockDragStrategy implements IDragStrategy {
     delta: Coordinate,
   ): ConnectionCandidate | null {
     if (this.moveMode === MoveMode.CONSTRAINED) {
-      const direction = this.getDirectionToNewLocation(
-        Coordinate.sum(this.startLoc!, delta),
-      );
-      return this.findTraversalCandidate(direction);
+      return this.findTraversalCandidate(delta);
     }
 
     // If we do not have a candidate yet, we fallback to the closest one nearby.
+    return this.getClosestCandidate(this.block, delta);
+  }
+
+  /**
+   * Returns the closest connection candidate for the given block.
+   *
+   * @param block The block to find a connection for.
+   * @param delta The distance the block has traveled since dragging began.
+   * @returns The closest available connection candidate, if any.
+   */
+  private getClosestCandidate(block: BlockSvg, delta: Coordinate) {
     let radius = this.getSearchRadius();
-    const localConns = this.getLocalConnections(this.block);
+    const localConns = this.getLocalConnections(block);
     let candidate: ConnectionCandidate | null = null;
 
     for (const conn of localConns) {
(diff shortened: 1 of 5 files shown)


packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -456,7 +456,7 @@ export class BlockDragStrategy implements IDragStrategy {
       // suggest using unconstrained mode to arbitrarily position the block if
       // we're in keyboard-driven constrained mode.
       if (this.moveMode === MoveMode.CONSTRAINED) {
-        showUnconstrainedMoveHint(this.workspace, true);
+        showUnconstrainedMoveHint(this.workspace);
```
maribethb, line 462: Is it possible to differentiate this case from the case where a block was already on the workspace? There are two scenarios in this else clause:

1. You are doing constrained move of a block that has multiple positions to go to, and the current position happens to be the workspace stop, but if you keep moving the block in constrained mode (with looping on, or move back the way you came), the block would in fact move somewhere else.
2. You are doing constrained move of a block that has nowhere else to go, so constrained move does not do anything because there are no other connection points.

You've improved the behavior by only showing the hint once, but I think that ideally scenario 1 would not show the hint at all. If this is too complicated to do in this PR, I would be fine submitting this as an improvement and filing a bug to follow up with this.

Edit: kept reading the rest of the PR and you are doing some work related to differentiating these, so maybe you can reuse that work here

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -479,6 +479,37 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      const bounds = root?.getBoundingRectangle();
+      if (!bounds) return;
+
+      let destination: Coordinate;
+      switch (direction) {
+        case Direction.LEFT:
+        case Direction.UP:
+          destination = new Coordinate(
+            bounds.getOrigin().x,
+            bounds.getOrigin().y - 20 - draggingBlock.getHeightWidth().height,
```
maribethb: this magic number should probably be a constant, don't we already have one for the amount a block moves in unconstrained mode? i would use that

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -807,9 +862,63 @@ export class BlockDragStrategy implements IDragStrategy {
...
+   * position), and a block that just entered move mode as a top-level block,
+   * and should therefore still be able to move to another connection point
+   * even if looping is disabled.
+   *
+   * @param block The block to check.
+   * @param direction The current dragging direction.
+   * @returns True if the block is at the start or end of its possible positions
+   *     on the workspace.
+   */
+  private isInTerminalPosition(
```
maribethb, line 887: The amount of work you're doing to figure this out does make me wonder if it means there's a better approach than including the workspace in the list of possible places to go, but you'd have to talk to Mike about how he implemented that to see if y'all can work out something better.

maribethb (approved):
I think it is worth thinking through if there is a possible better approach to handle adding the workspace as a "stop" in constrained move, since it has added complexity here. But if you think about it and don't have any better ideas, this LGTM.

packages/blockly/core/dragging/block_drag_strategy.ts
mikeharv: Unfortunately, I can’t see an alternative way to do this without scanning workspace connectivity. The cached state doesn’t include any notion of "terminal", so it has to be inferred at runtime from connectivity rather than read directly from state.

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Done

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Used the offset constant we already had in this class

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Yeah, I agree this is more complex than I'd like but I don't see any better alternatives unfortunately.

maribethb (approved):

PR #9737, opened 2026-04-20
fix!: Improve keyboard navigation of icons and bubbles

### LINKED ISSUE #9720: [v13.0.0-beta.0] Mutator workspaces issues with keyboard controls
### Check for duplicates

### Description

When opening mutator workspaces, all of the text inside the workspace is highlighted in yellow. The mutator workspace doesn't have the right focus at this point. Arrow keys move keyboard focus in the background, see https://github.com/RaspberryPiFoundation/blockly/issues/9719. Escape does not close the mutator workspace. 

If you open the mutator workspace, and then hit Tab, then it does get the right kind of focus. At this point, everything works correctly inside the mutator workspace, but one can still not close it with Escape.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/f291b72a-f3c5-458b-b4b8-95c81d6c4347

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9720

### Proposed Changes
This PR improves the interaction of keyboard navigation with icons and bubbles. Now, for all kinds of built-in bubbles:

* Open bubbles are treated as part of the first row of the block, and are navigable with left/right
* When a bubble is focused, M may be used to move it, Escape may be used to close it, and (for mutator and comment bubbles) Enter may be used to move to its mutator workspace/comment editor
* Mutator bubbles don't highlight the text of blocks in the mutator toolbox when they are focused

### Breaking Changes
* `Blockly.BlockCommentNavigationPolicy` has been replaced with a more-generic `Blockly.BubbleNavigationPolicy`, which should be used in place of it.


packages/blockly/tests/mocha/keyboard_navigation_test.js
```diff
@@ -14,6 +14,8 @@ import {
 } from './test_helpers/setup_teardown.js';
 import {createKeyDownEvent} from './test_helpers/user_input.js';
 
+config.showDiff = false;
```
maribethb: curious did you mean to leave this off or was this just from debugging something?

maribethb (approved):

packages/blockly/tests/mocha/keyboard_navigation_test.js
gonfunko: Nope this was from debugging so the tests don't lock up the browser when an assert fails. Removed, thanks for flagging!

PR #9744, opened 2026-04-21
feat: `FieldInput` ARIA

mikeharv:
## The basics

- [x] I validated my changes

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9656

### Proposed Changes

Implements foundational ARIA support for input fields, including text input and number fields.

Key changes:
- Implemented default ARIA behavior in `FieldInput`:
  - Provides a default type (`"input field"`) if none is specified with the block definition
  - Ensures a non-empty ARIA value (uses localized placeholder for empty strings)
  - Adds a private `recomputeAriaContext()` helper to:
    - Set `role="button"`
    - Set `aria-label` using `computeAriaLabel`
    - Prefix labels with localized "Edit" text when editable
    - Hide fields in flyouts via `aria-hidden`
- Wired ARIA updates to lifecycle events:
  - `initView`
  - `doValueUpdate_`
  - `doValueInvalid_`
- Updated field text elements to be `aria-hidden` since they are represented via ARIA labels
- Added new message strings:
  - `FIELD_LABEL_EMPTY`
  - `FIELD_LABEL_EMPTY_INPUT`
  - `ARIA_TYPE_FIELD_INPUT`
  - `FIELD_LABEL_EDIT_PREFIX`

### Reason for Changes

This change establishes a baseline pattern in `FieldInput` so that fields have meaningful default accessibility behavior without requiring every implementation to solve the same problems independently. It ensures that ARIA labeling stays in sync with field state and that fields communicate their interactivity in a predictable way.

### Test Coverage

Added unit tests for ARIA behavior on:
- `FieldTextInput`
- `FieldNumber`

Tests cover:
- Role assignment (`button`)
- Visibility handling (`aria-hidden` in flyouts)
- Default and updated ARIA labels
- Presence/absence of "Edit" prefix based on editability
- Label updates when field values change

Manual testing was also performed in a rendered workspace to verify:
- ARIA labels update correctly when editing fields
- Screen readers announce expected values

### Documentation

New methods include TSDocs.

### Additional Information

- This PR establishes the base pattern for ARIA support; additional field types (e.g. dropdowns, checkboxes, images) will implement these APIs in follow-up changes.

### DIFF
--- a/packages/blockly/core/field_input.ts
+++ b/packages/blockly/core/field_input.ts
@@ -811,7 +811,7 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     return text;
   }
 
- /**
+  /**
    * Gets an ARIA-friendly label representation of this field's type.
    *
    * Implementations are responsible for, and encouraged to, return a localized
@@ -824,7 +824,7 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     return this.ariaTypeName || Msg['ARIA_TYPE_FIELD_INPUT'];
   }
 
- /**
+  /**
    * Gets an ARIA-friendly label representation of this field's value.
    *
    * Implementations are responsible for, and encouraged to, return a localized
--- a/packages/blockly/core/field_input.ts
+++ b/packages/blockly/core/field_input.ts
@@ -811,15 +811,26 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
     return text;
   }
 
-  /**
-   * Provide a default type if none is specified.
+ /**
+   * Gets an ARIA-friendly label representation of this field's type.
+   *
+   * Implementations are responsible for, and encouraged to, return a localized
+   * version of the ARIA representation of the field's type.
+   *
+   * @returns An ARIA representation of the field's type or a default if it is
+   *     unspecified.
    */
   override getAriaTypeName(): string | null {
     return this.ariaTypeName || Msg['ARIA_TYPE_FIELD_INPUT'];
   }
 
-  /**
-   * Text inputs should always expose a meaningful value, even when empty.
+ /**
+   * Gets an ARIA-friendly label representation of this field's value.
+   *
+   * Implementations are responsible for, and encouraged to, return a localized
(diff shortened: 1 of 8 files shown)


packages/blockly/core/field_input.ts
```diff
@@ -807,6 +810,57 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
+
+    if (this.getSourceBlock()?.isInFlyout) {
+      aria.setState(focusableElement, aria.State.HIDDEN, true);
+      return;
+    }
+
+    aria.setState(focusableElement, aria.State.HIDDEN, false);
+    // The button role is intended to indicate to users that the field has an
+    // editing mode that can be activated.
+    focusableElement.setAttribute('role', 'button');
```
gonfunko: Probably good to use `aria.setRole()` here for consistency?

packages/blockly/msg/json/en.json
```diff
@@ -468,5 +468,8 @@
 	"ANNOUNCE_MOVE_AROUND": "moving %1 %2 around %3",
 	"ANNOUNCE_MOVE_TO": "moving %1 %2 to %3 %4",
 	"ANNOUNCE_MOVE_CANCELED": "Canceled movement",
-	"FIELD_LABEL_EMPTY": "empty"
+	"FIELD_LABEL_EMPTY": "empty",
+	"FIELD_LABEL_EMPTY_INPUT": "empty",
```
gonfunko: What's the difference between these two?

gonfunko (approved):

packages/blockly/msg/json/en.json
mikeharv: There isn't a difference, except that one is for the base field class. I wasn't sure if we should keep them distinct in case they diverge, for example if the new one changes to something like "empty input". Removed redundant message.

packages/blockly/core/field_input.ts
mikeharv: Good call!

PR #9745, opened 2026-04-21
fix: Don't close the flyout when creating a variable using keyboard nav

### LINKED ISSUE #9122: Creating a variable closes the flyout

gonfunko:
## The basics

### Resolves

Fixes #9122

### Proposed Changes
This PR updates Blockly's `dialog` utility to use a `<dialog>` element instead of the built-in browser alert modals. This allows for more consistent focus management, since the dialog is just part of the DOM rather than an application-level modal. Secondarily, it adds an event listener to the flyout workspace in the create variable button callback handler that identifies and focuses the newly-created variable block for convenience now that the flyout no longer closes.

### DIFF
--- a/packages/blockly/core/dialog.ts
+++ b/packages/blockly/core/dialog.ts
@@ -249,9 +249,8 @@ function displayDialog(
 
   form.appendChild(buttonRow);
 
-  const focusedNode = getFocusManager().getFocusedNode();
   let restoreFocus: (() => void) | undefined;
-  if (focusedNode && activeDialogCount === 0) {
+  if (activeDialogCount === 0) {
     restoreFocus = getFocusManager().takeEphemeralFocus(dialog);
   }
 
--- a/packages/blockly/core/dialog.ts
+++ b/packages/blockly/core/dialog.ts
@@ -252,9 +252,7 @@ function displayDialog(
   const focusedNode = getFocusManager().getFocusedNode();
   let restoreFocus: (() => void) | undefined;
   if (focusedNode && activeDialogCount === 0) {
-    restoreFocus = getFocusManager().takeEphemeralFocus(
-      focusedNode.getFocusableElement(),
-    );
+    restoreFocus = getFocusManager().takeEphemeralFocus(dialog);
   }
 
   activeDialogCount++;
--- a/packages/blockly/core/dialog.ts
+++ b/packages/blockly/core/dialog.ts
@@ -6,15 +6,24 @@
 
 // Former goog.module ID: Blockly.dialog
 
+import {getFocusManager} from './focus_manager.js';
+import {Msg} from './msg.js';
 import type {ToastOptions} from './toast.js';
 import {Toast} from './toast.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
 
+/** Supported types of dialogs. */
+enum DialogType {
+  ALERT = 1,
+  CONFIRM,
+  PROMPT,
+}
+
+/** Tally of the number of dialogs currently on screen. */
+let activeDialogCount = 0;
+
 const defaultAlert = function (message: string, opt_callback?: () => void) {
-  window.alert(message);
-  if (opt_callback) {
-    opt_callback();
-  }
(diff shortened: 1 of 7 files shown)


packages/blockly/core/dialog.ts
```diff
@@ -165,3 +172,111 @@ export function setToast(
...
   toastImplementation = toastFunction;
 }
+
+/**
+ * Displays a dialog, potentially prompting for user input, and invokes the
+ * provided callback with the response.
+ */
+function displayDialog(
+  ...[type, message, callback, defaultValue]:
+    | [
```
maribethb, line 182: I wonder if function overloading would be more readable than this? you'd be able to have better docs on the parameters as well.

packages/blockly/core/dialog.ts
```diff
@@ -23,7 +32,7 @@ const defaultConfirm = function (
   message: string,
   callback: (result: boolean) => void,
 ) {
-  callback(window.confirm(message));
+  displayDialog(DialogType.CONFIRM, message, callback, undefined);
```
maribethb, line 35: Are you sure that ephemeral focus shouldn't be marked taken and then released in the callback? I know for the flyout use case you're somewhat manually restoring focus so that might be fine, but in the general case in order to restore focus correctly I think you might want this? Plus, some of the shortcuts specifically check if ephemeral focus is taken in their precondition

maribethb (commented):
1 question about ephemeral focus, and then also want to double check there's no behavioral change for developers who are calling our prompt functions right? so this isn't breaking?

packages/blockly/core/dialog.ts
gonfunko: I tried that initially but TS can't infer the types of the callback arguments based on the type of dialog, whereas with this approach it can.

packages/blockly/core/dialog.ts
gonfunko: It is taken in the implementation of `displayDialog()` (line 255). I just tweaked that to take it for the dialog vs the focused element since I think that's the design intent and may be helpful for the focus-lost-outside-the-div situation.

gonfunko:
Correct, there should be no functional behavior difference - same API calls work, the contents are just displayed in our custom dialog vs the browser modal. And it's still possible for developers to replace our implementation with their own.

packages/blockly/core/dialog.ts
```diff
@@ -165,3 +172,109 @@ export function setToast(
...
+    cancel.value = CANCEL;
+    buttonRow.appendChild(cancel);
+  }
+
+  form.appendChild(buttonRow);
+
+  const focusedNode = getFocusManager().getFocusedNode();
+  let restoreFocus: (() => void) | undefined;
+  if (focusedNode && activeDialogCount === 0) {
+    restoreFocus = getFocusManager().takeEphemeralFocus(dialog);
```
maribethb, line 254: thanks, missed it here. this makes sense to me. `dialog` is the correct thing that should be taking ephemeral focus. so i think you can leave off the condition of if there is already a focused node. the focus manager should handle it correctly either way.

maribethb (approved):

packages/blockly/core/dialog.ts
gonfunko: Whoops yeah missed removing that, thanks!

gonfunko:
This also fixed #9106.

Issue #8868, opened 2025-04-07
Clicking outside of Blockly area doesn't update the field value inside the block


Joost-Jens-Luminis:
### Check for duplicates

- [x] I have searched for similar issues before opening a new one.

### Description

When the user is editing a field, and clicks outside of the blockly area, the CHANGE event is not fired.

I expect the change event to be fired, or for there to be a way to request blockly to lose focus of the text field and update it's value in the block. The behaviour should be similar to how Blockly works if you click inside the workspace.

### Reproduction steps

1. There is a block on in the Workspace with a text field.
2. The user begins editing the text field (it select the field)
3. When the user is done editing, it clicks outside the blockly area (for example on a button)

Result:
The text field is still shown as being edited and the internal block is not yet updated with the new value. No CHANGE even is fired either.

### Stack trace

```shell

```

Teruk15:
Hi @Joost-Jens-Luminis, is this still open? Can I work on this?

Joost-Jens-Luminis:
@Teruk15 I have a workaround which works for me. But it would be better if this was natively/automatically supported in Blockly.
Far be it for me to tell you what to do though :)

maribethb:
Hi @Teruk15 this bug isn't a great choice at the moment because it's affected by some in-progress work we're doing with focus management. I'd recommend checking the `help wanted` label :)

maribethb:
@BenHenning This is related to https://github.com/google/blockly-keyboard-experimentation/issues/563 and our discussion around it fyi. Further support for making the behavior between clicking the workspace & clicking outside it consistent.

BenHenning:
Confirming #9175 doesn't directly fix this, but I think it makes it possible to fix with some additional work.

Once #9175 is merged it will be possible for widget div to be updated to automatically dispose of itself when focus is lost (in much the same way #9175 updates drop-down div to auto-hide). I think doing that will cause the field editor to, in turn, close and update the field result (and fire the change event).

rachel-fenichel:
Marking for re-triage as this needs verification of whether the issue still exists.

maribethb:
My understanding is the same work that #9175 did for dropdown divs needs to happen for widget divs as well in order to close this

maribethb:
@gonfunko can you check if your PR closes this bug too?

PR #9749, opened 2026-04-22
fix: Improve focus handling when clicking outside injection div

gonfunko:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/563

### Proposed Changes
This PR improves focus management when focus transitions out of the injection div, either via click or tab, while the `WidgetDiv`, `DropDownDiv`, or both are open. Previously, this would get focus management into a bad wedged state. Now, the `WidgetDiv`/`DropDownDiv`(s) will be dismissed, the element that spawned them will get passive focus, and focus will transition to whatever the clicked/tabbed-to element is. When focus returns to the workspace, the element that spawned the `WidgetDiv`/`DropDownDiv` will regain active focus.

At a high level, this works by:
* Listening for `focusout` events on the div that holds the `WidgetDiv`/`DropDownDiv`
* When a `focusout` event is received, checks to make sure that the newly-focused element is actually outside of that div
* If so, notifies the `WidgetDiv` and `DropDownDiv`, which will (a) resign ephemeral focus if they hold it, while instructing the focus manager to *not* actively focus the previously-focused element, and (b) hide themselves
* Keeping tabs on the appropriate state, in the case of multiple workspaces and multiple `*Div`s, as focus moves around and the `*Div`s hide and show.

I manually verified that this worked for dropdowns, menus, and field editors with a single playground and in the multi-playground, and also that behavior is as expected for `FieldAngle` which uses both the `DropDownDiv` and the `WidgetDiv` simultaneously. The latter will need some slight tweaks (in general, for fields that use both, `WidgetDiv` should take ephemeral focus in preference to `DropDownDiv` so that the field editor is focused initially), but focus can move between the field editor and angle picker, both can be interacted with, and when the value is committed or focus leaves the workspace, the field becomes passively/actively focused as appropriate.

I have not added unit tests because (a) it would be a pain but more importantly (b) it'd be very useful to make sure that this is robust and works for MakeCode given the previous churn that attempts to resolve this issue have caused before investing the effort to put together a test suite given point (a).

### DIFF
--- a/packages/blockly/core/common.ts
+++ b/packages/blockly/core/common.ts
@@ -87,7 +87,7 @@ export function getMainWorkspace(): Workspace {
 export function setMainWorkspace(workspace: Workspace) {
   mainWorkspace = workspace;
   if (workspace.rendered) {
-    getFocusManager().setQuasiModalFocusRoot(
+    getFocusManager().setPopoverFocusRoot(
       (workspace as WorkspaceSvg).getInjectionDiv(),
     );
   }
--- a/packages/blockly/core/common.ts
+++ b/packages/blockly/core/common.ts
@@ -86,6 +86,11 @@ export function getMainWorkspace(): Workspace {
  */
 export function setMainWorkspace(workspace: Workspace) {
   mainWorkspace = workspace;
+  if (workspace.rendered) {
+    getFocusManager().setQuasiModalFocusRoot(
+      (workspace as WorkspaceSvg).getInjectionDiv(),
+    );
+  }
 }
 
 /**
--- a/packages/blockly/core/focus_manager.ts
+++ b/packages/blockly/core/focus_manager.ts
@@ -15,7 +15,7 @@ import {FocusableTreeTraverser} from './utils/focusable_tree_traverser.js';
  *
  * See FocusManager.takeEphemeralFocus for more details.
  */
-export type ReturnEphemeralFocus = () => void;
+export type ReturnEphemeralFocus = (restoreFocus?: boolean) => void;
 
 /**
  * Represents an IFocusableTree that has been registered for focus management in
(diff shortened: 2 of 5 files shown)


packages/blockly/core/focus_manager.ts
```diff
@@ -15,7 +15,7 @@ import {FocusableTreeTraverser} from './utils/focusable_tree_traverser.js';
  *
  * See FocusManager.takeEphemeralFocus for more details.
  */
-export type ReturnEphemeralFocus = () => void;
+export type ReturnEphemeralFocus = (restoreFocus?: boolean) => void;
```
maribethb, line 22: can you add info about the restoreFocus param here, including why you might want to set it to false vs true

packages/blockly/core/common.ts
```diff
@@ -86,6 +86,11 @@ export function getMainWorkspace(): Workspace {
  */
 export function setMainWorkspace(workspace: Workspace) {
   mainWorkspace = workspace;
+  if (workspace.rendered) {
+    getFocusManager().setQuasiModalFocusRoot(
```
maribethb: as discussed i don't think we can use this name sorry :(

I think "popover" fits: https://developer.apple.com/design/human-interface-guidelines/popovers

gemini suggests "portal" which is apparently a react term for a control that's rendered outside its parent dom hierarchy, but that might be too niche of a reference. it also suggests "overlay"

packages/blockly/core/focus_manager.ts
gonfunko: Done

packages/blockly/core/common.ts
gonfunko: Done

gonfunko:
This also fixes #8868

maribethb (approved):

PR #9755, opened 2026-04-23
feat: Added keyboard shortcut for displaying tooltip

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/447

### Proposed Changes

Adds a keyboard shortcut for displaying the tooltip.

### Reason for Changes

Adding the ability to display the tooltip from a keyboard shortcut increases accessibility for keyboard users.

### Test Coverage

Added unit tests around basic functionality, ensuring an existing tooltip does not block a new tooltip from popping up, and ensuring it does not display while dragging is happening.

### Additional Information

Screenshots:
Screenshot 2026-04-24 at 1 48 43 PM
Screenshot 2026-04-24 at 1 48 07 PM

[ ] TODO: Update the key mapping to something that makes sense.
The key mapping in this PR is a placeholder; which actual key(s) to use is something that still needs to be decided. I picked Ctrl+J as it wasn't currently in use. I saw that VS Code uses Ctrl+K followed by Ctrl+I but I didn't see an obvious way to register this in the code and didn't want to spend extra time on it before we even know if we need to do so.

### DIFF
--- a/packages/blockly/core/tooltip.ts
+++ b/packages/blockly/core/tooltip.ts
@@ -8,6 +8,7 @@
 
 import * as browserEvents from './browser_events.js';
 import * as common from './common.js';
+import type {IFocusableNode} from './interfaces/i_focusable_node.js';
 import * as blocklyString from './utils/string.js';
 import type {WorkspaceSvg} from './workspace_svg.js';
 
@@ -365,8 +366,12 @@ export function hide() {
 
 /**
  * Display the tooltip for a given target.
+ *
+ * @internal
+ * @param target The node upon which the tooltip should be displayed.
+ * @param workspace The target node's workspace.
  */
-export function display(target: AnyDuringMigration, workspace?: WorkspaceSvg) {
+export function display(target: IFocusableNode, workspace?: WorkspaceSvg) {
   // If the target is not the same element currently displaying a tooltip, hide
   // the existing tooltip and set the target as our element.
   if (element !== target) {
@@ -383,6 +388,14 @@ export function display(target: AnyDuringMigration, workspace?: WorkspaceSvg) {
     // user gesture, such as a click or drag.
     return;
   }
+
+  // Set the position to just below the element with horizontal alignment based
+  // on the target's RTL/LTR orientation.
+  const targetRect = target.getFocusableElement().getBoundingClientRect();
+  const rtl = element.RTL;
+  lastX = rtl ? targetRect.x + targetRect.width : targetRect.x;
+  lastY = targetRect.y + targetRect.height;
+
   show(workspace);
 }
 
--- a/packages/blockly/core/tooltip.ts
+++ b/packages/blockly/core/tooltip.ts
@@ -364,6 +364,29 @@ export function hide() {
 }
 
 /**
+ * Display the tooltip for a given target.
+ */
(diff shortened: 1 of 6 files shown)


packages/blockly/core/tooltip.ts
```diff
@@ -363,6 +363,29 @@ export function hide() {
   }
 }
 
+/**
+ * Display the tooltip for a given target.
+ */
+export function display(target: AnyDuringMigration, workspace?: WorkspaceSvg) {
```
lizschwab: The `show()` function is relying on some checks already having been made, so I added this new function rather than exporting that one.

gonfunko: Maybe make this `@internal` and use `IFocusableNode` as the type for `target`? If we're not expecting third party callers to use this it's probably better to keep it locked down for now.

packages/blockly/tests/mocha/shortcut_items_test.js
```diff
@@ -1435,4 +1435,55 @@ suite('Keyboard Shortcut Items', function () {
       assert.deepEqual(oldBounds, newBounds);
     });
   });
+
```
gonfunko, line 1438: Throughout, `assert.isTrue()`/`assert.isFalse()` might make the intent a bit clearer.

packages/blockly/core/tooltip.ts
gonfunko: Would be nice to add `@param` lines to the docstring too.

gonfunko (approved):

packages/blockly/core/tooltip.ts
```diff
@@ -363,6 +364,41 @@ export function hide() {
...
+    // user gesture, such as a click or drag.
+    return;
+  }
+
+  // Set the position to just below the element with horizontal alignment based
+  // on the target's RTL/LTR orientation.
+  const targetRect = target.getFocusableElement().getBoundingClientRect();
+  const rtl = element.RTL;
+  lastX = rtl ? targetRect.x + targetRect.width : targetRect.x;
+  lastY = targetRect.y + targetRect.height;
```
lizschwab, line 397: This is new. I discovered while testing my changes that it was basing the location of the tooltip on the mouse position. This calculates it based on the target element.

gonfunko (approved):

microbit-robert:
Regarding the shortcut, cmd J while using VoiceOver on mac works to open the tooltip, but it also triggers VoiceOver to output "jump to selection". Incidental, but when I was trying this out, I immediately tried cmd J to toggle the tooltip closed again, but you need Escape. No strong view either way.

lizschwab:
@microbit-robert Thanks for bringing that to our attention. This is a placeholder until we can find a better solution. I did not run into this issue myself when testing with VO, but we'll be replacing the shortcut with something else once it's been further discussed anyway so I'm not too concerned. Let us know if you have input on that; we've discussed Ctrl/Cmd + K, Ctrl/Cmd + I, but that is more complex than we'd like, Ctrl/Cmd + Q, but that is the quit application shortcut in Mac, and Ctrl/Cmd + F1, but we've been avoiding function keys up to this point. Happy to hear any suggestions!

PR #9766, opened 2026-04-24
feat: `FieldDropdown` ARIA

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9657

---

### Proposed Changes

This PR add the ARIA implementation of `FieldDropdown` .

Key changes include:

- Adds correct ARIA role and state management to the dropdown trigger element:
  - Ensures the field’s focusable element has `role="button"`
  - Updates `aria-expanded` dynamically based on whether the dropdown menu is open
  - Sets `aria-haspopup="listbox"` to indicate the popup type
  - Sets `aria-controls` to reference the unique ID of the dropdown menu
  - Dropdown fields are hidden from the accessibility tree when rendered inside flyouts

- Introduces a Blockly-unique ID for the dropdown menu root element so it can be referenced reliably via `aria-controls`

- Improves menu item accessibility structure:
  - Adds support for ARIA labels on menu options
  - Ensures non-essential visual-only/redundant content is not exposed to screen readers by marking it as presentation

- Enhances ARIA label computation for dropdown options:
  - Explicit ARIA label override (if provided) takes highest priority
  - Image-based options use `ariaLabel`, then `alt` as fallback
  - HTMLElement-based options use accessible properties in priority order (ARIA label, title, then visible text)
  - String-based options are used directly
  - Final fallback produces a human-readable representation of the option (including indexed fallback where necessary)

---

### Reason for Changes

This change establishes a baseline pattern in `FieldDropdown` so that fields have meaningful default accessibility behavior without requiring every implementation to solve the same problems independently. It ensures that ARIA labeling stays in sync with field state and that fields communicate their interactivity in a predictable way.

---

### Test Coverage

This change is covered by an expanded suite of ARIA-focused unit tests for `FieldDropdown`.

Simple Dropdown
- Block has field type name in ARIA label
- Focusable element has role of button
- Hidden when in a flyout
- Does not have aria-expanded when dropdown is closed
- Has aria-expanded when dropdown is open
- Has aria-haspopup of listbox
- Has aria-controls that matches the ID of the dropdown menu
- Has placeholder ARIA label by default
- setValue updates ARIA label

Dropdown with Option ARIA labels
- Option ARIA labels are included in field ARIA label
- Option ARIA labels are included in field ARIA label when value is changed

Dropdown with image options
- Image alt text is included in ARIA label
- Image ARIA label is prioritized over alt text

Dropdown with HTMLElement options
- Explicit ARIA label overrides all other label sources
- HTMLElement ariaLabel prioritized over other properties
- HTMLElement title is used when ariaLabel is missing
- HTMLElement innerText is used as final fallback
- Empty label falls back to option index

---

### Documentation

New methods include TSDocs.

---

### Additional Information

Further follow-up work will be required to update built-in Blockly block definitions to fully utilize improved ARIA option labeling. This could to updating some tests here as block definitions change.

### DIFF
--- a/packages/blockly/core/field_dropdown.ts
+++ b/packages/blockly/core/field_dropdown.ts
@@ -25,6 +25,7 @@ import * as fieldRegistry from './field_registry.js';
 import {Menu} from './menu.js';
 import {MenuSeparator} from './menu_separator.js';
 import {MenuItem} from './menuitem.js';
+import {Msg} from './msg.js';
 import * as aria from './utils/aria.js';
 import {Coordinate} from './utils/coordinate.js';
 import * as dom from './utils/dom.js';
@@ -97,6 +98,8 @@ export class FieldDropdown extends Field<string> {
 
   /** The total vertical padding above and below an image. */
   protected static IMAGE_Y_PADDING = FieldDropdown.IMAGE_Y_OFFSET * 2;
+
+  isInitialized: boolean = false;
 
   /**
    * @param menuGenerator A non-empty array of options for a dropdown list, or a
@@ -197,6 +200,8 @@ export class FieldDropdown extends Field<string> {
       dom.addClass(this.fieldGroup_, 'blocklyField');
       dom.addClass(this.fieldGroup_, 'blocklyDropdownField');
     }
+    this.recomputeAriaContext();
+    this.isInitialized = true;
   }
 
   /**
@@ -295,6 +300,7 @@ export class FieldDropdown extends Field<string> {
     }
 
     this.applyColour();
+    aria.setState(this.getFocusableElement(), aria.State.EXPANDED, true);
   }
 
   /** Create the dropdown editor. */
@@ -306,6 +312,11 @@ export class FieldDropdown extends Field<string> {
     const menu = new Menu();
     menu.setRole(aria.Role.LISTBOX);
     this.menu_ = menu;
+    aria.setState(
+      this.getFocusableElement(),
+      aria.State.CONTROLS,
+      this.menu_.getId(),
+    );
 
     const options = this.getOptions(false);
     this.selectedMenuItem = null;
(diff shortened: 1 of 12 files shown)


packages/blockly/core/menuitem.ts
```diff
@@ -43,6 +43,9 @@ export class MenuItem {
   private actionHandler: ((obj: this, menuSelectEvent: Event) => void) | null =
     null;
 
+  /** The unique ID for this menu item. */
+  private id: string | null = null;
```
gonfunko: I think you could just assign it to `getNextUniqueId()` here and not have to have it be nullable/handle doing that in the getter?

packages/blockly/core/menu.ts
```diff
@@ -62,6 +63,9 @@ export class Menu {
   /** ARIA name for this menu. */
   private roleName: aria.Role | null = null;
 
+  /** The menu's ID. */
+  private id: string | null = null;
```
gonfunko: As with `MenuItem` I think you could directly assign the ID here?

packages/blockly/core/field_dropdown.ts
```diff
@@ -98,6 +99,8 @@ export class FieldDropdown extends Field<string> {
   /** The total vertical padding above and below an image. */
   protected static IMAGE_Y_PADDING = FieldDropdown.IMAGE_Y_OFFSET * 2;
 
+  isInitialized: boolean = false;
```
gonfunko, line 106: A comment explaining the purpose of this might be helpful

packages/blockly/tests/mocha/field_dropdown_test.js
```diff
@@ -325,4 +325,261 @@ suite('Dropdown Fields', function () {
...
+      });
+      test('Has placeholder ARIA label by default', function () {
+        const label = this.focusableElement.getAttribute('aria-label');
+        assert.include(label, 'true');
+      });
+      test('setValue updates ARIA label', function () {
+        const initialLabel = this.focusableElement.getAttribute('aria-label');
+        assert.include(initialLabel, 'true');
+        this.field.setValue('FALSE');
+        console.log(this.field.getOptions());
```
gonfunko: Remove

gonfunko (approved):

Issue #9041, opened 2025-05-13
The trashcan control and the trashcan flyout.


BenHenning:

BenHenning:
Trashcan flyout doesn't work as-is today, even with additional work in the keyboard navigation plugin to support it as a proper state. Without digging in more detail I suspect that this may require changes to `WorkspaceSvg` and `Flyout` base to support the case of the flyout being part of the trash can experience. Separately, there may be some work needed for navigation to work here (but I actually expect that to "just work" once the focus bits are corrected).

There's also some oddity in navigating to the trashcan flyout when it's open, but I think that's largely because it's not closing/auto-closing as expected with keyboard navigation. I suspect we'll largely want to treat this similar to the toolbox flyout in how and when it auto-hides.

BenHenning:
Updating the issue to include making the trashcan and its flyout workspace navigable (including navigating into/out of the trashcan flyout).

rachel-fenichel:
@gonfunko does the trash flyout work after your flyout changes?

gonfunko:
It does!

rachel-fenichel:
And is there a way for a user to get into and out of that flyout with keyboard navigation?

gonfunko:
Not entirely, you have to use the mouse to click the trash to open it. It is tabbable once you've done that though, and you can insert from it to restore a block from the trash.

Issue #9040, opened 2025-05-13
Zoom controls.


BenHenning:

BenHenning:
Updating the issue to include making zoom controls navigable, as well.

PR #9768, opened 2026-04-24
fix: Make trashcan and zoom controls accessible

### LINKED ISSUE #9042: Other positionables, such as [search widgets](https://google.github.io/blockly-samples/plugins/workspace-search/test/index.html) if we're putting that in the top right corner

gonfunko:
## The basics

### Resolves

Fixes #9042

### Proposed Changes
This PR makes the trashcan and zoom controls accessible via the keyboard. It does this by making them tabstops rather than part of the navigation order, since having to navigate through them with the arrow keys repeatedly would be annoying and frequently unneeded. When focused via the tab key, Enter/Space can be used to act upon them in the same way as a click. Each control has its ARIA role set to `button` and an appropriate label set. When the trash can is acted upon, the first block in its flyout will be focused.


gonfunko:
This also fixes #9040

gonfunko:
This fixes #9041 as well

mikeharv (approved):
Looks like a good upgrade. I like the cleaner separation of zoom actions into discrete components. Tab-stop approach and ARIA roles/labels all make sense. It seems well aligned to the current mouse behavior.

PR #9772, opened 2026-04-29
feat: ARIA for other field classes

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9659 and https://github.com/RaspberryPiFoundation/blockly/issues/9658

### Proposed Changes

This PR add the ARIA implementation of `FieldCheckbox`, `FieldImage`, `FieldLabel`, and `FieldVariable`. It also makes some minor adjustments to `FieldNumber` and other more general changes to how ARIA methods are used.

#### General updates
- `computeAriaLabel` was updated to `includeTypeInfo = true` by default. Currently, we only exclude type info for move mode announcements, which makes it the clear exception rather than the rule.
- As a result of the above change, type info is now intentionally included in the labels for a field's focusable element.
- Changed implementations of `recomputeAriaContext()` from `private` to `protected`. We expect develops may need to extend this method in their custom field subclasses.
- Added several new translatable strings for describing field ARIA type names and field values.

#### `FieldVariable`
- Use a qualified version of the variable as the field's ARIA value, e.g. `'Variable "i"'`
- Update the renaming option to include the variable name, e.g. `'Rename the "i" variable'`

#### `FieldCheckbox`
- Sets the field's focusable element to have the role checkbox.
- Set its `aria-checked` state based on whether it's been checked.
- Sets `aria-hidden` if in the flyout.
- Returns an ARIA value of 'checked' or 'not checked' for use with block ARIA labels.
- Sets the field's label to its provided ARIA type name or 'Checkbox'. (The field label does not include 'checked' or 'unchecked' because this is already provided by the `aria-checked` state.)
- Recomputes its ARIA state when its checked state changes.

#### `FieldImage`
- Provides its image's alt text as its ARIA value by default
- Sets its role to either button or presentation, depending on whether it has a click handler. (Only clickable image fields are navigable.)
- Sets its label to include its type name (or "image") and alt text.
- Recomputes its ARIA state when its value changes.

#### `FieldLabel`
- Set itself to always be `aria-hidden` since it should not be navigable.

#### `FieldNumber`
- Removed its custom `aria-valuemin` and `aria-valuemax` logic as it likely never worked. (These properties are for range widgets like sliders.)

### Reason for Changes

This change continues the baseline pattern (https://github.com/RaspberryPiFoundation/blockly/pull/9683, https://github.com/RaspberryPiFoundation/blockly/pull/9744, https://github.com/RaspberryPiFoundation/blockly/pull/9766 ) so that fields have meaningful default accessibility behavior without requiring every implementation to solve the same problems independently. It ensures that ARIA labeling stays in sync with field state and that fields communicate their interactivity in a predictable way.

### Test Coverage

Added/updated unit tests for ARIA behavior on checkbox, dropdown, image, label, number, textinput, and variable fields.

Tests cover:
- Role assignment
- Visibility handling
- Default and updated ARIA labels
- Label updates when field values change
- Type name included in the ARIA label

Notes:
- Dropdown field ARIA label checks were relaxed from `equal` to `include` because the label now also includes the ARIA type name.
- `FieldVariable` benefits from the ARIA logic in `FieldDropdown`, so those tests were written to reinforce that alignment. 
- An existing test was updated to confirm that variable rename/delete option labels include the variable name.

### DIFF
--- a/packages/blockly/core/field_label.ts
+++ b/packages/blockly/core/field_label.ts
@@ -77,8 +77,8 @@ export class FieldLabel extends Field<string> {
     }
     if (this.fieldGroup_) {
       dom.addClass(this.fieldGroup_, 'blocklyLabelField');
+      aria.setState(this.fieldGroup_, aria.State.HIDDEN, true);
     }
-    this.recomputeAriaContext();
   }
 
   /**
@@ -127,16 +127,6 @@ export class FieldLabel extends Field<string> {
     // `this` might be a subclass of FieldLabel if that class doesn't override
     // the static fromJson method.
     return new this(text, undefined, options);
-  }
-
-  /**
-   * Recomputes the ARIA role and label for this field.
-   */
-  protected recomputeAriaContext(): void {
-    const element = this.getFocusableElement();
-    if (!element) return;
-
-    aria.setState(element, aria.State.HIDDEN, true);
   }
 }
 
--- a/packages/blockly/core/field_label.ts
+++ b/packages/blockly/core/field_label.ts
@@ -14,6 +14,7 @@
 
 import {Field, FieldConfig} from './field.js';
 import * as fieldRegistry from './field_registry.js';
+import {aria} from './utils.js';
 import * as dom from './utils/dom.js';
 import * as parsing from './utils/parsing.js';
 
@@ -77,6 +78,7 @@ export class FieldLabel extends Field<string> {
     if (this.fieldGroup_) {
       dom.addClass(this.fieldGroup_, 'blocklyLabelField');
     }
+    this.recomputeAriaContext();
   }
 
   /**
@@ -125,6 +127,16 @@ export class FieldLabel extends Field<string> {
     // `this` might be a subclass of FieldLabel if that class doesn't override
     // the static fromJson method.
     return new this(text, undefined, options);
+  }
+
+  /**
+   * Recomputes the ARIA role and label for this field.
+   */
+  protected recomputeAriaContext(): void {
+    const element = this.getFocusableElement();
+    if (!element) return;
+
+    aria.setState(element, aria.State.HIDDEN, true);
   }
 }
 
--- a/packages/blockly/core/utils/aria.ts
+++ b/packages/blockly/core/utils/aria.ts
@@ -202,6 +202,18 @@ export enum State {
    * Value:one of {true, false, undefined}.
    */
   SELECTED = 'selected',
+  /**
+   * See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemax.
+   *
+   * Value: a number representing the maximum allowed value for a range widget.
+   */
+  VALUEMAX = 'valuemax',
+  /**
+   * See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemin.
+   *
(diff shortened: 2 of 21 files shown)


packages/blockly/msg/messages.js
```diff
@@ -139,7 +139,7 @@ Blockly.Msg.REDO = 'Redo';
 Blockly.Msg.CHANGE_VALUE_TITLE = 'Change value:';
 /** @type {string} */
 /// dropdown choice - When the user clicks on a variable block, this is one of the dropdown menu choices.  It is used to rename the current variable.  See [https://github.com/RaspberryPiFoundation/blockly/wiki/Variables#dropdown-menu https://github.com/RaspberryPiFoundation/blockly/wiki/Variables#dropdown-menu].
-Blockly.Msg.RENAME_VARIABLE = 'Rename variable...';
+Blockly.Msg.RENAME_VARIABLE = 'Rename the "%1" variable';
```
gonfunko, line 142: I think it would be safer to add this as a new string since it adds a placeholder where there was none - I'm not sure if interpolating throws or otherwise has issues if a placeholder is/isn't present as expected, but might be best to err on the side of caution?

packages/blockly/core/utils/aria.ts
```diff
@@ -370,6 +358,7 @@ export function announceDynamicAriaState(
   text: string,
   options?: DynamicAnnouncementOptions,
 ) {
+  console.log(text);
```
gonfunko: Remove

packages/blockly/core/utils/aria.ts
```diff
@@ -202,18 +202,6 @@ export enum State {
...
-   *
-   * Value: a number representing the maximum allowed value for a range widget.
-   */
-  VALUEMAX = 'valuemax',
-  /**
-   * See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuemin.
-   *
-   * Value: a number representing the minimum allowed value for a range widget.
-   */
-  VALUEMIN = 'valuemin',
```
gonfunko, line 216: Removing this is a breaking change, I think it's best to just leave them here even if we're not using them in `FieldNumber` any more.

packages/blockly/core/field_label.ts
```diff
@@ -126,6 +128,16 @@ export class FieldLabel extends Field<string> {
...
   }
+
+  /**
+   * Recomputes the ARIA role and label for this field.
+   */
+  protected recomputeAriaContext(): void {
+    const element = this.getFocusableElement();
+    if (!element) return;
+
+    aria.setState(element, aria.State.HIDDEN, true);
```
gonfunko: Can this just go in `createTextElement_`?

gonfunko (approved):

packages/blockly/core/utils/aria.ts
mikeharv: Reverted!

packages/blockly/core/field_label.ts
mikeharv: Moved to `initView()` as this class wasn't already extending `createTextElement_`.

packages/blockly/msg/messages.js
mikeharv: Interpolation is done via `.replace()` which doesn't throw if there's a problem. We discussed this and agreed to leave it as a replacement for an existing string.

PR #9783, opened 2026-04-30
feat: Bubble ARIA methods

mikeharv:
This updates `Bubble` with new API changes:
- `type AriaLabelProvider`
- `recomputeAriaContext()`
- `setAriaLabelProvider()`
- `getAriaLabel()`

These new methods ensure that `Bubble`s have a valid ARIA label and the default role `group`. To use, `Bubble` owners (such as `Icon`s) are responsible computing the context at the correct times and providing an `AriaLabelProvider` to override the default `"Bubble"` label.

## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9661

### Proposed Changes

Bubble will be updated with new API changes:

```
/**
 * Represents a function that, when called, can provide a custom ARIA string to
 * represent a bubble, or null if the default fallback should be used. See
 * setAriaLabelProvider for more context.
 */
export type AriaLabelProvider = (bubble: Bubble) => string | null;

/**
 * Recomputes the ARIA label and role for this bubble. This is automatically called
 * during initialization, but implementations may find it useful to call this if
 * the bubble's label should be changed.
 *
 * Bubbles use a default non-specific label unless they're customized otherwise
 * which is the responsibility of the bubble's owner rather than bubble
 * implementations. Customization can be done via setAriaLabelProvider.
 */
protected recomputeAriaContext()

/**
 * Sets a custom ARIA label provider for this bubble, or null if it should be reset
 * to use the default method.
 *
 * Bubbles do not compute ARIA labels specifically to their implementation since
 * they can be rather general-purpose. Instead, owners of the specific bubble
 * instance (such as an icon) are responsible for defining custom label providers
 * for their bubbles.
 *
 * Note that calling this isn't sufficient for it to actually be used.
 * recomputeAriaContext will likely also need to be called to actually apply the
 * custom label to the bubble's focusable element.
 */
setAriaLabelProvider(provider: AriaLabelProvider | null)

/**
 * Returns the ARIA label to use for this bubble based on the provider set via
 * setAriaLabelProvider. This will return null if the provider is absent or
 * returns null.
 *
 * @returns The ARIA label to use for this bubble, or null to use a default.
 */
getAriaLabel(): string | null
```

The documentation strings explain the contracts of these methods, but it's expected that implementations will never need to override any of these values. `Bubble` owners (such as `Icon`s) are responsible for actually setting up the `Bubble` instance's ARIA context to be correct (e.g. by calling `recomputeAriaContext` as the correct time). All bubbles are treated as role `group` unless `recomputeAriaContext` is overridden. The focusable element's label is set to the value returned by `getAriaLabel` (or a default `'Bubble'` if no `AriaLabelProvider` is configured or it returns `null`, to ensure that at least some human-readable label is used).

### Reason for Changes

This provides the groundwork that is necessary to give built-in bubble owners (such as mutators, warnings, and comments) reasonable ARIA labels. This will also make it possible for developers to do the same for their own `Bubble` implementations.

### Test Coverage

To start, this just tests that the three main types of bubbles we support have the correct role and _a_ label. It also tests, via comments, that a label provider works as expected. A follow-up task will define those providers and test that they are working as expected.

### DIFF
--- a/packages/blockly/core/bubbles/bubble.ts
+++ b/packages/blockly/core/bubbles/bubble.ts
@@ -823,7 +823,7 @@ export abstract class Bubble
       if (typeof this.ariaLabelProvider === 'string') {
         return this.ariaLabelProvider;
       } else if (typeof this.ariaLabelProvider === 'function') {
-      return this.ariaLabelProvider(this);
+        return this.ariaLabelProvider(this);
       }
     }
     return null;
--- a/packages/blockly/core/bubbles/bubble.ts
+++ b/packages/blockly/core/bubbles/bubble.ts
@@ -27,11 +27,11 @@ import {Svg} from '../utils/svg.js';
 import {WorkspaceSvg} from '../workspace_svg.js';
 
 /**
- * Represents a function that, when called, can provide a custom ARIA string to
- * represent a bubble, or null if the default fallback should be used. See
- * setAriaLabelProvider for more context.
+ * Represents a either a string or a function that, when called, can provide a
+ * custom ARIA string to represent a bubble, or null if the default fallback
+ * should be used. See setAriaLabelProvider for more context.
  */
-export type AriaLabelProvider = (bubble: Bubble) => string | null;
+export type AriaLabelProvider = string | ((bubble: Bubble) => string | null);
 
 /**
  * The abstract pop-up bubble class. This creates a UI that looks like a speech
@@ -820,7 +820,11 @@ export abstract class Bubble
    */
   getAriaLabel(): string | null {
     if (this.ariaLabelProvider) {
+      if (typeof this.ariaLabelProvider === 'string') {
+        return this.ariaLabelProvider;
+      } else if (typeof this.ariaLabelProvider === 'function') {
       return this.ariaLabelProvider(this);
+      }
     }
     return null;
   }
--- a/packages/blockly/core/bubbles/bubble.ts
+++ b/packages/blockly/core/bubbles/bubble.ts
@@ -16,6 +16,7 @@ import type {IHasBubble} from '../interfaces/i_has_bubble.js';
 import {ISelectable} from '../interfaces/i_selectable.js';
 import {ContainerRegion} from '../metrics_manager.js';
 import {Scrollbar} from '../scrollbar.js';
+import {aria} from '../utils.js';
 import {Coordinate} from '../utils/coordinate.js';
 import * as dom from '../utils/dom.js';
 import * as idGenerator from '../utils/idgenerator.js';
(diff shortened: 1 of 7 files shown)


packages/blockly/core/bubbles/bubble.ts
```diff
@@ -25,6 +26,13 @@ import {Size} from '../utils/size.js';
 import {Svg} from '../utils/svg.js';
 import {WorkspaceSvg} from '../workspace_svg.js';
 
+/**
+ * Represents a function that, when called, can provide a custom ARIA string to
+ * represent a bubble, or null if the default fallback should be used. See
+ * setAriaLabelProvider for more context.
+ */
+export type AriaLabelProvider = (bubble: Bubble) => string | null;
```
maribethb: I would probably update this to allow either a string or a function that returns a string. Then it will match patterns used elsewhere (in block for the role description, and in input for the aria label). Those two places didn't exist when the design was originally written for this.

packages/blockly/core/bubbles/bubble.ts
```diff
@@ -759,4 +773,55 @@ export abstract class Bubble
...
+   */
+  protected recomputeAriaContext(): void {
+    const element = this.getFocusableElement();
+    if (!element) return;
+
+    aria.setRole(element, aria.Role.GROUP);
+
+    const label = this.getAriaLabel()?.trim();
+
+    aria.setState(element, aria.State.LABEL, label ? label : 'Bubble');
```
maribethb, line 794: The `Bubble` string would need to be localized. I do wonder if there's a better default we could use, because "bubble" is so meaningless to end users. But I can't think of another one that's better. At least the built-in use cases of bubble will be able to have better defaults so this would only occur if someone is implementing a custom icon that launches a custom bubble and doesn't bother to set a correct label.

mikeharv: Good catch, thanks! I agree that it's meaningless, but weighed the trade offs in the same way as you.

packages/blockly/core/bubbles/bubble.ts
mikeharv: Seems like a good idea. Added.

packages/blockly/tests/mocha/mutator_test.js
```diff
@@ -84,4 +84,27 @@ suite('Mutator', function () {
       );
     });
   });
+  suite('ARIA', function () {
+    setup(async function () {
+      this.workspace = Blockly.inject('blocklyDiv', {});
+      defineMutatorBlocks();
+      const block = createRenderedBlock(this.workspace, 'xml_block');
```
gonfunko: Could just use controls_if here to not have to mess around with loading/unloading custom definitions

gonfunko (approved):

packages/blockly/tests/mocha/mutator_test.js
gonfunko, line 98: Can remove this too presumably?

PR #9785, opened 2026-04-30
feat: Display keyboard shortcuts in context menus

### LINKED ISSUE #9739: associated keyboard shortcut for context menu items
### Check for duplicates

### Problem

some context menu items also have shortcut items associated with them, and in those cases we want to display the keyboard shortcut in the menu. to make the process of creating the context menu easier, we can add an api to associate the context menu item with a keyboard shortcut

### Request

- add an optional `associatedKeyboardShortcut` to `ContextMenuItem` interface (feel free to debate name)
- should be a string, the name of a keyboard shortcut as it exists in the registry
- if the kb shortcut does not exist in the registry at the time we access the shortcut name, then log a warning but then otherwise proceed as if the property were undefined
- in the context menu creation, move the logic that currently adds the keyboard shortcut combination to the display text to check if there is an associated keyboard shortcut or not, and if so display the preferred key combination (see #9740)
- update the relevant context menu items in core to associate their keyboard shortcuts

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9739

### Proposed Changes
This PR displays keyboard shortcuts associated with context menu items in the context menu. Context menu items that perform the same action as a keyboard shortcut should add the name the shortcut is registered under to their `associatedKeyboardShortcut` field to inform Blockly of the relationship between them.

### DIFF
--- a/packages/blockly/core/contextmenu.ts
+++ b/packages/blockly/core/contextmenu.ts
@@ -335,15 +335,17 @@ function makeMenuitem(
 }
 
 /**
- * Returns a keyboard shortcut for the given context menu item, if any.
+ * Returns a textual representation of the keyboard shortcut for the given
+ * context menu item, if any.
  *
  * @param option The context menu item to retrieve a keyboard shortcut for.
- * @returns The keyboard shortcut registered under the same ID as the context
- *     menu item, if any.
+ * @returns A textual representation of the keyboard shortcut registered under
+ *     the name stored in the menu option's `associatedKeyboardShortcut` field,
+ *     if any.
  */
 function getKeyboardShortcut(
   option: ContextMenuOption | LegacyContextMenuOption,
 ): string {
-  if (!('id' in option)) return '';
-  return getShortcutKeysShort(option.associatedKeyboardShortcut ?? option.id);
+  if (!('id' in option) || !option.associatedKeyboardShortcut) return '';
+  return getShortcutKeysShort(option.associatedKeyboardShortcut);
 }
--- a/packages/blockly/core/contextmenu.ts
+++ b/packages/blockly/core/contextmenu.ts
@@ -344,21 +344,6 @@ function makeMenuitem(
 function getKeyboardShortcut(
   option: ContextMenuOption | LegacyContextMenuOption,
 ): string {
-  // Map built-in menu item keys to shortcut keys. These should be the same
-  // going forward, but this preserves backwards compatibility for long-standing
-  // menu items/shortcuts.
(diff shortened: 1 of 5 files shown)


gonfunko:
Refactored this to use `associatedKeyboardShortcut` as described in #9739.

packages/blockly/core/contextmenu_registry.ts
```diff
@@ -188,6 +190,7 @@ export namespace ContextMenuRegistry {
     displayText: ((p1: Scope) => string | HTMLElement) | string | HTMLElement;
     separator?: never;
     preconditionFn: (p1: Scope, menuOpenEvent: Event) => string;
+    associatedKeyboardShortcut?: string;
```
maribethb, line 199: can you add a comment explaining this pleas

packages/blockly/core/contextmenu.ts
```diff
@@ -302,3 +304,46 @@ export function callbackFactory(
...
+  }
+
+  return option.text;
+}
+
+/**
+ * Returns a keyboard shortcut for the given context menu item, if any.
+ *
+ * @param option The context menu item to retrieve a keyboard shortcut for.
+ * @returns The keyboard shortcut registered under the same ID as the context
```
maribethb: I think we should just have the associatedKeyboardShortcut, not also use the matching id.

1. only having one path to do this
2. that path being less magical and more api-based
3. it would be a breaking change, people might already have a shortcut item named the same as a context menu item but not want them associated (e.g. if they have manually added the keyboard keys in the display text and wouldn't want them to appear twice)

the return value is currently a bit wrong and will be more wrong if you make this change so please update that as well

maribethb (commented):
can you please also update the PR description to avoid referencing the compatibility table you refactored out, it makes the release notes confusing

packages/blockly/core/contextmenu_registry.ts
gonfunko: Done

packages/blockly/core/contextmenu.ts
gonfunko: Done

maribethb (approved):

microbit-robert:
@gonfunko A head's up that this causes aria-labels for context menu items to be '[object HTMLDivElement]' due to [this fallback option of using String(this.content)](https://github.com/RaspberryPiFoundation/blockly/blob/v13/packages/blockly/core/menuitem.ts#L114-L118) which is now getting passed an HTML element instead of a string.

PR #9789, opened 2026-04-30
feat: Added support for custom input  ARIA labels

lizschwab:
## The basics

- [x] I [validated my changes]

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9752

### Proposed Changes

Adds a method to set a custom ARIA label to inputs when creating from javascript or when creating from JSON along with support for using that ARIA label if available or using the default if unavailable.

### Reason for Changes

This allows developers to set their own ARIA labels for inputs when creating blocks.

### Test Coverage

Unit tests are included. Manual testing was performed by updating the blocks defined in the playground and ensuring they are assigned the correct ARIA labels.

### DIFF
--- a/packages/blockly/tests/mocha/block_test.js
+++ b/packages/blockly/tests/mocha/block_test.js
@@ -1896,88 +1896,26 @@ suite('Blocks', function () {
       });
 
       suite('ARIA', function () {
-        suite('Bubble', function () {
-          setup(async function () {
-            this.block.setWarningText('Warning Text');
-            this.block.initSvg();
-            this.block.render();
-            const icon = this.block.getIcon(Blockly.icons.WarningIcon.TYPE);
-            icon.performAction();
-            await Blockly.renderManagement.finishQueuedRenders();
+        setup(async function () {
+          this.block.setWarningText('Warning Text');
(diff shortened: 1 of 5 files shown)


packages/blockly/tests/mocha/block_test.js
```diff
@@ -1896,26 +1896,88 @@ suite('Blocks', function () {
...
+              'group',
+            );
+          });
         });
-        test('Bubble has ARIA role of group', async function () {
-          assert.equal(
-            this.bubble.focusableElement.getAttribute('role'),
-            'group',
-          );
+        suite('Input', function () {
```
mikeharv: Could we consider moving this to a new ARIA suite at the bottom of `input_test` instead of here? The existing suite above is here because this block is all nested under `Block>Icon management>Warning Icons`. 
I'm also considering moving the above suite to new file specifically for warnings, but I don't think you need to worry about that.

lizschwab: Ah, gotcha. I didn't notice where this was nested, just that it was the only ARIA test group I found. I'll move it.

maribethb (approved):

PR #9805, opened 2026-05-01
feat: Icon ARIA

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9662

### Proposed Changes

Updates `Icon` and its subclasses with the following API changes:

```
/**
 * Recomputes the ARIA label and role for this icon. This is automatically called
 * during initialization, but implementations may find it useful to call this if
 * the icon's label should be changed.
 */
protected recomputeAriaContext()

/**
 * Returns the ARIA label to use for this icon (defaults to null). Note that this
 * method will only be called during initialization by default, so dynamic changes
 * to the icon's ARIA label need to be applied by calling recomputeAriaContext.
 *
 * @returns The ARIA label to use for this icon, or null to use a default.
 */
protected getAriaLabel(): string | null
```

This follows a similar pattern to [`Bubble`s](https://github.com/RaspberryPiFoundation/blockly/pull/9783) including defaulting the label to `'Icon'` if no override is provided. By default, `Icon`s are given the role `button` as it best fits their interactive nature since it's expected that `Icons` are always clickable. Developers are thus expected to use non-clickable `FieldImage`s when they want to include non-clickable graphics as part of a block. 

Icons are also responsible for configuring their corresponding Bubbles to have correct ARIA labels (with translation support). This is done by providing an `AriaLabelProvider` string or function to the `Bubble`. This is done because `Bubble` and its subtypes do not have any particular awareness of the context in which they are being used.

Icons for Mutators, Warnings, and Comments have a new `getAriaLabel` override that returns a localized message pertaining to opening its bubble or closing it. The bubble itself will be labeled either as 'Block editor workspace' (mutator) or the full bubble text preceded by `'Warning: '`/`'Comment: '`.

In order to ensure that labels for comment bubbles are updated, a change listener was added to `TextInputBubble`. Note that read-only workspaces use `TextBubble`, which does need to have a similar mechanism due to users not being able to type. Warnings are also effectively read-only.

### Reason for Changes

This gives built-in bubble owners (mutators, warnings, and comments) correct dynamically updated ARIA labels and roles for their icons and bubbles. This will also make it possible for developers to do the same for their own `Icon` implementations.

### Test Coverage

A new test suite was added for `Icon` which tests roles and labels for generic icons. For mutators, warnings, and comments, existing suites were updated check label providers, expected roles for icons and bubbles, and dynamic updates from toggling the bubble visibility and updating comment text.

### DIFF
--- a/packages/blockly/core/bubbles/textinput_bubble.ts
+++ b/packages/blockly/core/bubbles/textinput_bubble.ts
@@ -116,6 +116,11 @@ export class TextInputBubble extends Bubble {
     this.editor.addTextChangeListener(listener);
   }
 
+  /** Removes the given listener from the list of text change listeners. */
+  removeTextChangeListener(listener: () => void) {
+    this.editor.removeTextChangeListener(listener);
+  }
+
   /** Adds a change listener to be notified when this bubble's size changes. */
   addSizeChangeListener(listener: () => void) {
     this.sizeChangeListeners.push(listener);
@@ -289,6 +294,16 @@ export class TextInputBubble extends Bubble {
    */
   performAction() {
     getFocusManager().focusNode(this.getEditor());
+  }
+
+  /**
+   * Dispose of this bubble.
+   */
+  dispose() {
+    super.dispose();
+    this.removeTextChangeListener(() => {
+      this.recomputeAriaContext();
+    });
   }
 }
 
--- a/packages/blockly/core/bubbles/textinput_bubble.ts
+++ b/packages/blockly/core/bubbles/textinput_bubble.ts
@@ -85,6 +85,9 @@ export class TextInputBubble extends Bubble {
     this.contentContainer.appendChild(this.editor.getDom());
     this.resizeGroup = this.createResizeHandle(this.svgRoot, workspace);
     this.setSize(this.DEFAULT_SIZE, true);
+    this.addTextChangeListener(() => {
+      this.recomputeAriaContext();
+    });
   }
 
   /** @returns the text of this bubble. */
--- a/packages/blockly/core/bubbles/textinput_bubble.ts
+++ b/packages/blockly/core/bubbles/textinput_bubble.ts
@@ -63,6 +63,9 @@ export class TextInputBubble extends Bubble {
   /** View responsible for supporting text editing. */
   private editor: CommentEditor;
 
+  private readonly textChangeListener = () => {
+    this.recomputeAriaContext();
+  };
   /**
    * @param workspace The workspace this bubble belongs to.
    * @param anchor The anchor location of the thing this bubble is attached to.
@@ -85,9 +88,7 @@ export class TextInputBubble extends Bubble {
     this.contentContainer.appendChild(this.editor.getDom());
     this.resizeGroup = this.createResizeHandle(this.svgRoot, workspace);
     this.setSize(this.DEFAULT_SIZE, true);
-    this.addTextChangeListener(() => {
-      this.recomputeAriaContext();
-    });
+    this.addTextChangeListener(this.textChangeListener);
   }
 
   /** @returns the text of this bubble. */
@@ -114,11 +115,6 @@ export class TextInputBubble extends Bubble {
   /** Adds a change listener to be notified when this bubble's text changes. */
   addTextChangeListener(listener: () => void) {
     this.editor.addTextChangeListener(listener);
-  }
-
-  /** Removes the given listener from the list of text change listeners. */
-  removeTextChangeListener(listener: () => void) {
-    this.editor.removeTextChangeListener(listener);
   }
 
   /** Adds a change listener to be notified when this bubble's size changes. */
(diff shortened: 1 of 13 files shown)


packages/blockly/core/bubbles/textinput_bubble.ts
```diff
@@ -85,6 +85,9 @@ export class TextInputBubble extends Bubble {
     this.contentContainer.appendChild(this.editor.getDom());
     this.resizeGroup = this.createResizeHandle(this.svgRoot, workspace);
     this.setSize(this.DEFAULT_SIZE, true);
+    this.addTextChangeListener(() => {
+      this.recomputeAriaContext();
+    });
```
gonfunko: This should be removed in dispose() as well.

packages/blockly/core/icons/comment_icon.ts
```diff
@@ -336,6 +344,9 @@ export class CommentIcon extends Icon implements IHasBubble, ISerializable {
         'comment',
       ),
     );
+    if (this.isInitialized) {
+      this.recomputeAriaContext();
+    }
```
gonfunko: Might be better to just check if `this.svgRoot` is non-null rather than introducing another instance variable?

packages/blockly/core/icons/icon.ts
```diff
@@ -75,6 +77,8 @@ export abstract class Icon implements IIcon, IContextMenu {
     );
     (this.svgRoot as any).tooltip = this;
     tooltip.bindMouseEvents(this.svgRoot);
+    console.log(this.svgRoot);
```
gonfunko: Remove

packages/blockly/tests/mocha/block_test.js
```diff
@@ -1897,26 +1897,56 @@ suite('Blocks', function () {
...
+        function getFocusableAriaLabel(iFocusable) {
+          return iFocusable.getFocusableElement().getAttribute('aria-label');
+        }
         test('Bubble has ARIA label', async function () {
           assert.isTrue(
             this.bubble.focusableElement.hasAttribute('aria-label'),
           );
         });
+        test('Bubble has working ARIA label provider', function () {
+          const label = this.bubble.focusableElement.getAttribute('aria-label');
```
gonfunko: Could use the `getFocusableAriaLabel()` helper function here?

packages/blockly/tests/mocha/comment_test.js
```diff
@@ -216,31 +216,54 @@ suite('Comments', function () {
...
+    test('Bubble has ARIA label', async function () {
       assert.isTrue(this.bubble.focusableElement.hasAttribute('aria-label'));
     });
-    test('Bubble has ARIA role of group', function () {
+    test('Bubble has working ARIA label provider', function () {
+      const label = getFocusableAriaLabel(this.bubble);
+      assert.include(label, 'Comment');
+      assert.include(label, 'test text');
+    });
+    test('Bubble has ARIA role of group', async function () {
```
gonfunko: This test doesn't appear to actually be async

gonfunko: Could also use the helper function here

packages/blockly/tests/mocha/mutator_test.js
```diff
@@ -88,20 +88,51 @@ suite('Mutator', function () {
...
       sharedTestTeardown.call(this);
     });
 
-    test('Bubble has ARIA label', async function () {
-      assert.isTrue(this.bubble.focusableElement.hasAttribute('aria-label'));
+    test('Bubble has working ARIA label provider', async function () {
+      const label = getFocusableAriaLabel(this.bubble);
+      assert.equal(label, 'Block editor workspace');
     });
     test('Bubble has ARIA role of group', async function () {
```
gonfunko: Not async

packages/blockly/tests/mocha/comment_test.js
mikeharv: This test is getting the role, not the label. So I'd either need to add a second helper or parameterize the existing one.

gonfunko: Ah, yeah fine to leave as is!

packages/blockly/core/bubbles/textinput_bubble.ts
```diff
@@ -290,6 +295,16 @@ export class TextInputBubble extends Bubble {
   performAction() {
     getFocusManager().focusNode(this.getEditor());
   }
+
+  /**
+   * Dispose of this bubble.
+   */
+  dispose() {
+    super.dispose();
+    this.removeTextChangeListener(() => {
```
gonfunko: This needs to be the same object as was added, otherwise it's a no-op. I don't think it needs to be split out into a standalone method either, it can just live in `dispose`.

gonfunko (approved):

PR #9806, opened 2026-05-01
feat: Insert blocks at focus point

### LINKED ISSUE #9758: [v13.0.0-beta.1] Blocks inserted from the flyout are always placed on the workspace
### Check for duplicates

### Description

Blocks inserted from the flyout using keyboard are always placed in the top left of the workspace, regardless of where keyboard focus was (or passive focus is) on the workspace. E.g., if I am focused on a function block in the workspace, then insert a statement block from the flyout, the statement block is not placed (previewed) inside the function block.

Blocks sometimes have a connection preview, but this is coincidental and requires compatible blocks to already be positioned in the top left of the workspace where the new block from the flyout will be placed.

Please see the videos below.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

Coincidental connection preview:
https://github.com/user-attachments/assets/e198d38a-101d-4354-97a5-5a44c39d5009

No connection preview - block placed on workspace instead of inside function block
https://github.com/user-attachments/assets/23e476c2-1ccf-4cdd-be15-a91a85035043

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9758

### Proposed Changes
This PR reintroduces support for inserting blocks from the flyout at or near the currently focused node in the workspace. If a connection is focused, and the chosen block can connect to it, it will be initially moved to/proposed for connection at that point; otherwise, the focused node's parent block will be identified, and, assuming there is one, the inserted block will be proposed for connection to the first compatible connection on that block. A final attempt will be made to connect the inserted block to the focused block's parent block's input connection, after which the inserted block will just be initially proposed as a top-level block.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -15,6 +15,7 @@ import {ConnectionType} from '../connection_type.js';
 import type {BlockMove} from '../events/events_block_move.js';
 import {EventType} from '../events/type.js';
 import * as eventUtils from '../events/utils.js';
+import {FocusManager} from '../focus_manager.js';
 import {showUnconstrainedMoveHint} from '../hints.js';
 import type {IBubble} from '../interfaces/i_bubble.js';
 import type {IConnectionPreviewer} from '../interfaces/i_connection_previewer.js';
@@ -26,7 +27,7 @@ import * as layers from '../layers.js';
 import {Msg} from '../msg.js';
 import * as registry from '../registry.js';
 import {finishQueuedRenders} from '../render_management.js';
-import type {RenderedConnection} from '../rendered_connection.js';
+import {RenderedConnection} from '../rendered_connection.js';
 import * as blocks from '../serialization/blocks.js';
 import {Coordinate} from '../utils.js';
 import * as aria from '../utils/aria.js';
@@ -243,10 +244,6 @@ export class BlockDragStrategy implements IDragStrategy {
     }
 
     this.block.setDragging(true);
-    this.workspace.getLayerManager()?.moveToDragLayer(this.block);
-    this.getVisibleBubbles(this.block).forEach((bubble) => {
-      this.workspace.getLayerManager()?.moveToDragLayer(bubble, false);
-    });
 
     // For keyboard-driven moves, cache a list of valid connection points for
     // use in constrained moved mode.
@@ -254,7 +251,6 @@ export class BlockDragStrategy implements IDragStrategy {
       this.cacheAllConnectionPairs();
 
       // Scooch the block to be offset from the connection preview indicator.
-      this.block.moveDuringDrag(this.startLoc);
       const neighbour = this.updateConnectionPreview(
         this.block,
         new Coordinate(0, 0),
@@ -262,10 +258,11 @@ export class BlockDragStrategy implements IDragStrategy {
       if (neighbour) {
         let offset: Coordinate;
         if (neighbour.type === ConnectionType.PREVIOUS_STATEMENT) {
-          const origin = this.block.getRelativeToSurfaceXY();
           offset = new Coordinate(
-            origin.x + this.BLOCK_CONNECTION_OFFSET,
-            origin.y - this.BLOCK_CONNECTION_OFFSET,
+            neighbour.x,
+            neighbour.y -
+              this.block.getHeightWidth().height -
+              this.BLOCK_CONNECTION_OFFSET,
           );
         } else {
           offset = new Coordinate(
@@ -275,7 +272,14 @@ export class BlockDragStrategy implements IDragStrategy {
         }
         this.block.moveDuringDrag(offset);
       }
+    } else {
+      this.block.moveDuringDrag(this.startLoc);
     }
+
+    this.workspace.getLayerManager()?.moveToDragLayer(this.block);
+    this.getVisibleBubbles(this.block).forEach((bubble) => {
+      this.workspace.getLayerManager()?.moveToDragLayer(bubble, false);
+    });
 
     this.announceMove(true);
     return this.block;
@@ -388,7 +392,7 @@ export class BlockDragStrategy implements IDragStrategy {
    * the initial connection pair is also used as the first connection candidate.
    */
   private storeInitialConnections(healStack: boolean) {
-    // Prioritze the block's parent connection (output or previous) if one exists.
+    // Prioritize the block's parent connection (output or previous) if one exists.
     let localParentConn: RenderedConnection | null = null;
     let parentTargetConn: RenderedConnection | null = null;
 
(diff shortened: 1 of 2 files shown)


packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -584,6 +589,8 @@ export class BlockDragStrategy implements IDragStrategy {
         ? currCandidate
         : newCandidate;
     this.connectionCandidate = candidate;
+    console.log('updated');
+    console.log(this.connectionCandidate);
```
mikeharv: Remove

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -1049,4 +1056,74 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      .getNavigator()
+      .getSourceBlockFromNode(passiveNode);
+    if (!passiveBlock) return null;
+
+    const passiveBlockConnections = passiveBlock.getConnections_(false);
+    let passiveConnection: RenderedConnection | null = null;
+
+    // If the passively focused node is a connection, return it if it is
+    // compatible with the dragged block.
+    if (passiveBlockConnections.includes(passiveNode as any)) {
```
mikeharv, line 1089: Could we check whether `passiveNode instanceof RenderedConnection` instead of casting `as any`? At that point, I think we'd also be able to safely infer that it's one of `passiveBlockConnection` because they have same source block.

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -554,6 +573,124 @@ suite('Keyboard-driven movement', function () {
...
+        this.workspace.getInjectionDiv().dispatchEvent(esc);
+      });
+
+      test('initially moves the block to the previously-focused input connection', function () {
+        const ifBlock = this.workspace.newBlock('controls_if');
+        ifBlock.initSvg();
+        ifBlock.render();
+
+        const inputConnection = ifBlock.getInput('IF0').connection;
+        Blockly.getFocusManager().focusNode(inputConnection);
```
mikeharv, line 624: Would it be worth also testing for a block with multiple similar inputs, e.g. focusing `'IF1`?

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -554,6 +573,124 @@ suite('Keyboard-driven movement', function () {
...
+        const down = createKeyDownEvent(Blockly.utils.KeyCodes.DOWN);
+        this.workspace.getInjectionDiv().dispatchEvent(down);
+        const enter = createKeyDownEvent(Blockly.utils.KeyCodes.ENTER);
+        this.workspace.getInjectionDiv().dispatchEvent(enter);
+
+        const movingBlock = Blockly.getFocusManager().getFocusedNode();
+        const candidate = movingBlock.getDragStrategy().connectionCandidate;
+
+        assert.equal(candidate.local, movingBlock.outputConnection);
+        assert.equal(candidate.neighbour, ifBlock.getInput('IF0').connection);
```
mikeharv, line 671: Similar to previous comment, this test could be slightly more valuable if the neighbour block was configured to have an `else if` branch.

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -554,6 +573,124 @@ suite('Keyboard-driven movement', function () {
...
+        ifBlock.render();
+
+        const statementConnection = ifBlock.getInput('DO0').connection;
+        Blockly.getFocusManager().focusNode(statementConnection);
+        const t = createKeyDownEvent(Blockly.utils.KeyCodes.T);
+        this.workspace.getInjectionDiv().dispatchEvent(t);
+        const enter = createKeyDownEvent(Blockly.utils.KeyCodes.ENTER);
+        this.workspace.getInjectionDiv().dispatchEvent(enter);
+
+        const movingBlock = Blockly.getFocusManager().getFocusedNode();
```
mikeharv: Calling these `printBlock`, `negateBlock` etc. would make it easier to read/visualize these tests without having having to cross-reference the toolbox definition way at the top of the file.

packages/blockly/tests/mocha/keyboard_movement_test.js
```diff
@@ -554,6 +573,124 @@ suite('Keyboard-driven movement', function () {
...
+        const candidate = movingBlock.getDragStrategy().connectionCandidate;
+
+        assert.equal(candidate.local, movingBlock.outputConnection);
+        assert.equal(candidate.neighbour, compare.getInput('A').connection);
+
+        const esc = createKeyDownEvent(Blockly.utils.KeyCodes.ESC);
+        this.workspace.getInjectionDiv().dispatchEvent(esc);
+      });
```
mikeharv, line 735: The tests that are here look solid. I think we can cut down on a lot of repeated code by using the parent suite's `moveDown`, `endMove` (maybe rename this one), and `cancelMove`. We could also create a new helper for T.

A few other scenarios come to mind:
- Multiple-input neighbour blocks (like using `IF1`, mentioned above)
- No compatible connections, falling back to top-level insertion

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -388,7 +392,7 @@ export class BlockDragStrategy implements IDragStrategy {
    * the initial connection pair is also used as the first connection candidate.
    */
   private storeInitialConnections(healStack: boolean) {
-    // Prioritze the block's parent connection (output or previous) if one exists.
+    // Prioritize the block's parent connection (output or previous) if one exists.
```
mikeharv, line 395: Thanks!

mikeharv (commented):
Pulled and tested. Seems to work great!

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: That's not the purpose for the cast, `passiveBlockConnections` is an array of `RenderedConnection`, so if it includes `passiveNode`, we know it is one; unfortunately without this, TS whinges that "Argument of type 'IFocusableNode' is not assignable to parameter of type 'RenderedConnection'.", stubbornly refusing to consider the possibility that the IFocusableNode `passiveNode` may in fact be a `RenderedConnection` (which implements `IFocusableNode`!). Since this is all verifiably safe I just casted to any to satisfy the compiler; doing an `instanceof` requires a full import of `RenderedConnection`, and IIRC that leads to an import cycle.

packages/blockly/core/dragging/block_drag_strategy.ts

packages/blockly/tests/mocha/keyboard_movement_test.js
gonfunko: Added one to that effect

packages/blockly/core/dragging/block_drag_strategy.ts
mikeharv: Ah, got it! I think that's totally reasonable then.

mikeharv (approved):

PR #9807, opened 2026-05-01
chore: add documentation on accessibility of custom fields

rachel-fenichel:
### Resolves
Part of https://github.com/RaspberryPiFoundation/blockly/issues/9787

### Proposed Changes

Add information about accessibility and ARIA to the pages on extending a built-in field and creating a custom field.

### Reason for Changes

Documentation!

### DIFF
--- a/packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
+++ b/packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
@@ -23,7 +23,7 @@ To create a new field, do the following:
 1.  [Handle disposal of event listeners](#disposing) (UI disposal is handled for
     you).
 1.  [Implement value handling](#value-handling).
-1.  [Add a text-representation of your field's value, for accessibility](#text).
+1.  [Add a text-representation of your field's value](#text).
 1.  Add additional functionality such as:
     - [An editor](#creating-an-editor).
     - [On-block display updates](#updating-the-on-block-display).
@@ -166,6 +166,25 @@ The requirements of a field's on-block display are:
 - All DOM elements must be children of the field's `fieldGroup_`. The field
   group is created automatically.
 - All DOM elements must stay inside the reported dimensions of the field.
+- The root DOM element has an appropriate 
+[ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Techniques).
+
+
+### Accessibility
+
+You are responsible for ensuring that your custom fields are WCAG compliant and
+correctly implement `IFocusableNode`. That includes having an appropriate ARIA
+role for your on-block display and making your field editor WCAG compliant.
+
+In general, the on-block display for a clickable field should have the ARIA
+role `button` to indicate that an action will take place when it is clicked or
+activated.
+
+When the on-block display is focused, its focusable DOM element will have the 
+`blocklyActiveFocus` or `blocklyPassiveFocus` CSS class added. Blockly defaults
+to showing a yellow outline around the focused field. You may override this
+behaviour as long as you show [visible focus](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html).
+
 
 See the
 [Rendering](/guides/create-custom-blocks/fields/customizing-fields/creating#updating-the-on-block-display)
@@ -180,11 +199,13 @@ field's degree symbol) you can append the symbol element (usually contained in a
 
 ### Input events
 
-By default fields register tooltip events, and mousedown events (to be used for
-showing
-[editors](/guides/create-custom-blocks/fields/customizing-fields/creating#creating-an-editor)).
-If you want to listen for other kinds of events (e.g. if you want to handle
-dragging on a field) you should override the field's `bindEvents_` function.
+By default fields register tooltip events, mousedown events, and some keyboard
+events. Users can open the field 
+[editor](/guides/create-custom-blocks/fields/customizing-fields/creating#creating-an-editor)
+by tapping or clicking on the field, or by navigating to the field with the
+keyboard and pressing `Enter`. If you want to listen for other kinds of events 
+(e.g. if you want to handle dragging on a field) you should override the field's
+`bindEvents_` function.
 
 ```js
 bindEvents_() {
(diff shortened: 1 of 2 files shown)


packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
```diff
@@ -166,6 +166,25 @@ The requirements of a field's on-block display are:
...
+
+### Accessibility
+
+You are responsible for ensuring that your custom fields are WCAG compliant and
+correctly implement `IFocusableNode`. That includes having an appropriate ARIA
+role for your on-block display and making your field editor WCAG compliant.
+
+In general, the on-block display for a clickable field should have the ARIA
+role `button` to indicate that an action will take place when it is clicked or
+activated.
```
mikeharv, line 181: Is this the only example because it's known to the most common?

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
```diff
@@ -424,12 +445,26 @@ getText() {
 }
 ```
 
+All fields must have an ARIA type and value, which are used to construct the
+ARIA label. The ARIA label is generally of the form `<ARIA type>: <ARIA value>`.
+The ARIA value defaults to the result of calling `getText` on the field. You 
+must ensure that the result is meaningful when read with a screen reader. To 
+this end, if getAriaValue returns an invalid representation (i.e. `null` or 
+empty string) then it should be instead represented using a localization of the 
+text empty string to convey this to the user.
```
mikeharv: Nit: Add quotes around "empty string" for clarity?

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
```diff
@@ -72,3 +72,28 @@ static fromJson(options) {
...
+The ARIA label is generally of the form `<ARIA type>: <ARIA value>`. The ARIA
+value defaults to the result of calling `getText` on the field. You must ensure
+that the result is meaningful when read with a screen reader. To this end, if 
+`getAriaValue` returns an invalid representation (i.e. `null` or empty string)
+then it should be instead represented using a localization of the text 
+`empty string` to convey this to the user.
+
+When extending a built-in field you will inherit its ARIA type and value logic.
+You can override the inherited ARIA logic by overriding any of `getAriaTypeName`,
+`getAriaValue`, and `computeAriaLabel`.
```
mikeharv, line 93: In general, do expect people to need to override `computeAriaLabel`? It's really just building the `<ARIA type>: <ARIA value>` string (or just `<ARIA value>` when type info is excluded). While it's possible to change how this works, I would think most developers would only need to override the other two.

mikeharv: I also wonder if it would be overkill to include an example here of when you might want to override, such as making things more specific... Example: `'Angle'` instead of `'Number'` for type or `n + ' degrees'` for clarifying the meaning of the value.

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
```diff
@@ -72,3 +72,28 @@ static fromJson(options) {
...
+behaviour as long as you show [visible focus](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html).
+
+All fields must have an ARIA type and value, which are used to construct the
+[ARIA label](/guides/create-custom-blocks/fields/anatomy-of-a-field#aria-label). 
+The ARIA label is generally of the form `<ARIA type>: <ARIA value>`. The ARIA
+value defaults to the result of calling `getText` on the field. You must ensure
+that the result is meaningful when read with a screen reader. To this end, if 
+`getAriaValue` returns an invalid representation (i.e. `null` or empty string)
+then it should be instead represented using a localization of the text 
+`empty string` to convey this to the user.
```
mikeharv: This is all factual, but I feel like we could make it a bit clearer. The word "it" on line 94 is particularly ambiguous, but I think we could generally tighten the responsibility here too:

> `getAriaValue` must return a human-readable, screen reader–appropriate representation of the field’s current value. By default, it returns the result of `getText`, which is not guaranteed to be informative for assistive technologies.
> You must override `getAriaValue` when `getText` might not produce a meaningful spoken representation (for example, when possible values includes symbols, abbreviations, or non-textual representations such as icons or colors).
> If `getAriaValue` would return an empty or non-informative value, it must instead return a localized string indicating that the value is empty.

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
rachel-fenichel: Yes. Are there other roles that you think are worth including? Checkbox is the only other one that came to mind.

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
rachel-fenichel: Do we have any built-in fields that do anything interesting for type or value that I could link to?

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/creating.mdx
rachel-fenichel: Updated to match the phrasing you suggested for extending a field.

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
rachel-fenichel: Updated.

rachel-fenichel (commented):
I updated the ARIA value phrasing. Please let me know if it's clearer this way.

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
mikeharv: `FieldVariable` implements [`getAriaValue`](https://github.com/RaspberryPiFoundation/blockly/pull/9772/changes#diff-a1c77d44795744d952e2e0a7518ed4af9a63840df8ac10e9d849b7cae4cf708a) to give a label like `"Variable 'i'"` instead of just `"i"`.

Lots of blocks have [custom labels for dropdown field options](https://github.com/RaspberryPiFoundation/blockly/pull/9880/changes),  especially in the math category.

Function definitions are interesting for [type name](https://github.com/RaspberryPiFoundation/blockly/pull/9902/changes#diff-833d44778d8d2c1421ce53f7d5ce902ceee07d8c154c70ff017c3bfedd358bcb), as they'll read `"function name: do something"` instead of `"text: do something"`.

mikeharv (approved):

packages/docs/docs/guides/create-custom-blocks/fields/customizing-fields/extending.mdx
rachel-fenichel: I'm going to leave this as-is, but possible add a link to your talk after it's published

PR #9812, opened 2026-05-04
fix: Added comma between labels for fields when composing ARIA label strings

### LINKED ISSUE #9791: [v13.0.0-beta.2] Block label commas inconsistent
### Check for duplicates

### Description

In the previous implementation, commas in the aria-label were used to wrap inputs in order to make these more pronounced to help screen reader users infer that these are in fact inputs. Commas in the current version are either missing, or can occur in unexpected places, perhaps before the label attached to an input. Please see the "show number" block below for an example of the latter.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

Image

Details of the output aside, I would expect commas to appear in this output as follows: "to, dropdown: Variables 'item', append text, image: "text: empty image:', has input, command"

Image

I would expect: "show number, number: 0, has input, command" whereas the comma currently occurs in between "show" and "number".

### Browsers

_No response_

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9791

### Proposed Changes

Adds a comma between inputs when joining together the array of input labels.

Screenshot:

Screenshot 2026-05-04 at 12 11 09 PM

### Reason for Changes

This change will add a comma between labels for fields when composing the ARIA label string. This signals to the screenreader that it should pause, which gives screenreader users valuable information about the composition of the block.

### Test Coverage

N/A

### DIFF
--- a/packages/blockly/tests/mocha/input_test.js
+++ b/packages/blockly/tests/mocha/input_test.js
@@ -369,17 +369,11 @@ suite('Inputs', function () {
       assert.equal(labelA, labelB);
     });
     test('Field labels are comma separated', function () {
-      this.block
-        .appendDummyInput('NAME')
-        .appendField(new Blockly.FieldTextInput('text'), 'NAME');
-      this.block
-        .appendDummyInput('VALUE')
-        .appendField(new Blockly.FieldNumber('number'), 'VALUE');
+      this.block.appendDummyInput().appendField('first').appendField('second');
 
       const label = this.block.getAriaLabel();
-      const inputLabels = this.block.inputList.map((input) => input.getLabel());
 
-      assert.include(label, inputLabels.join(', '));
+      assert.include(label, 'first, second');
     });
   });
 });
--- a/packages/blockly/tests/mocha/input_test.js
+++ b/packages/blockly/tests/mocha/input_test.js
@@ -368,5 +368,18 @@ suite('Inputs', function () {
       // AriaLabelProvider and without setting the provider (the default label)
       assert.equal(labelA, labelB);
     });
+    test('Field labels are comma separated', function () {
+      this.block
+        .appendDummyInput('NAME')
+        .appendField(new Blockly.FieldTextInput('text'), 'NAME');
+      this.block
+        .appendDummyInput('VALUE')
+        .appendField(new Blockly.FieldNumber('number'), 'VALUE');
+
+      const label = this.block.getAriaLabel();
+      const inputLabels = this.block.inputList.map((input) => input.getLabel());
+
+      assert.include(label, inputLabels.join(', '));
+    });
   });
 });
--- a/packages/blockly/core/inputs/input.ts
+++ b/packages/blockly/core/inputs/input.ts
@@ -413,7 +413,7 @@ export class Input {
         );
       }
     }
-    return labels.join(' ');
+    return labels.join(', ');
   }
 
   /**


mikeharv (commented):
Considering that this is a deliberate change, and the behavior has evidently flipped a couple times already, could we add a test?

packages/blockly/tests/mocha/input_test.js
```diff
@@ -368,5 +368,18 @@ suite('Inputs', function () {
       // AriaLabelProvider and without setting the provider (the default label)
       assert.equal(labelA, labelB);
     });
+    test('Field labels are comma separated', function () {
```
mikeharv, line 371: Thanks for adding a test. Unfortunately, I think it's giving a false positive as it is currently written. In fact, it passes without your core change. This is because each of your test inputs below only have a single field. It would work better to model a single input with multiple fields, since `getLabel()` composes from all labels within the input.
For example, you could do:
```
const input = this.block.appendDummyInput().appendField('first').appendField('second');
assert.include(input.getLabel(), 'first, second');
```

Using label fields like this also makes it clear that multiple labels being comma-separated is intentional. If we decided to make an exception for these later, we'd want to expand this to several tests covering different combinations of field types on the same input.

packages/blockly/tests/mocha/input_test.js
```diff
@@ -368,5 +368,18 @@ suite('Inputs', function () {
...
+        .appendDummyInput('NAME')
+        .appendField(new Blockly.FieldTextInput('text'), 'NAME');
+      this.block
+        .appendDummyInput('VALUE')
+        .appendField(new Blockly.FieldNumber('number'), 'VALUE');
+
+      const label = this.block.getAriaLabel();
+      const inputLabels = this.block.inputList.map((input) => input.getLabel());
+
+      assert.include(label, inputLabels.join(', '));
```
mikeharv: Using the same `.join(', ')` here in the test feels like it's going to give a weaker signal than explicit string comparison, since it's very similar to the way we currently implement this in core. If other parts of the implementation broke, this could continue to pass.

packages/blockly/tests/mocha/input_test.js
```diff
@@ -368,5 +368,18 @@ suite('Inputs', function () {
       // AriaLabelProvider and without setting the provider (the default label)
       assert.equal(labelA, labelB);
     });
+    test('Field labels are comma separated', function () {
+      this.block
+        .appendDummyInput('NAME')
+        .appendField(new Blockly.FieldTextInput('text'), 'NAME');
+      this.block
+        .appendDummyInput('VALUE')
+        .appendField(new Blockly.FieldNumber('number'), 'VALUE');
```
mikeharv: The first constructor arg in `Field` classes is generally going to be an optional but valid default value. `FieldNumber` expects a number (or string that can be cast to a number), so this value `'number'` ends up equivalent to passing `null`.

packages/blockly/tests/mocha/input_test.js
```diff
@@ -368,5 +368,18 @@ suite('Inputs', function () {
       // AriaLabelProvider and without setting the provider (the default label)
       assert.equal(labelA, labelB);
     });
+    test('Field labels are comma separated', function () {
+      this.block
+        .appendDummyInput('NAME')
+        .appendField(new Blockly.FieldTextInput('text'), 'NAME');
```
mikeharv: I think it's fine to omit names for fields and inputs in tests (e.g. `'NAME'`) if we don't need to reference them again later.

mikeharv (approved):

PR #9820, opened 2026-05-06
fix: Adjust keyboard navigation of external inputs

gonfunko:
## The basics

## The details
This PR adjusts the behavior of keyboard navigation on blocks, particularly those with external inputs. Previously, every empty external input or block connected to an external input was treated as being on its own row, so that pressing the up/down arrows would move focus to the next "row". For example, in this screenshot, pressing up would focus the empty "if true" input,  then the "if false" input on the most-deeply-nested ternary block, and so forth.
Screenshot 2026-05-06 at 9 19 13 AM

Now, rows are demarcated solely by (a) next/previous connections (b) separate block stacks, and (c) statement inputs. All nested blocks/inputs/fields inside a block that has a next/previous connection, is a root block, or is part of a clause of a block that has statement inputs are treated as being on its same row. Thus, up arrow in the screenshot above would take you to the previous block stack. Nested inputs/blocks may be visited via the left and right arrows.

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1978,9 +1978,9 @@ export class BlockSvg
 
   /**
    * Returns an ID for the visual "row" this block is part of. A "row" is
-   * bounded by a previous/next connection or a block stack boundary; all
-   * blocks/inputs nested inside of a block that is a root block or has a next/
-   * previous connection are conceptually part of its same row.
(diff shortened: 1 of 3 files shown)


maribethb (approved):
suggestion for the tsdoc: s/visual/logical when you describe "rows" since they aren't really visual anymore but more a logical grouping of things

PR #9827, opened 2026-05-07
fix: Improve ARIA fallback generation for HTML-based menu items

gonfunko:
## The basics

### Resolves

Fixes @microbit-robert's comment in #9785 

### Proposed Changes
This PR improves the fallback ARIA label generation of HTML-based contextual menu items. Previously, menu items that didn't have an explicit ARIA label set were wrapped in `String()`, which produced `[object HTMLDivElement]` or the like. Now, when the menu item is HTML-based, its `textContent` field is used.

### DIFF
--- a/packages/blockly/core/menuitem.ts
+++ b/packages/blockly/core/menuitem.ts
@@ -114,7 +114,12 @@ export class MenuItem {
   getAriaLabel(): string {
     // This fallback should only be hit by Context Menu items as all
     // FieldDropdown options should have an ARIA label.
-    return this.ariaLabel || String(this.content);
+    return (
+      this.ariaLabel ||
+      (typeof this.content === 'string'
+        ? String(this.content)
+        : this.content.textContent)
+    );
   }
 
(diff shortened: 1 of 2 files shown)


packages/blockly/core/menuitem.ts
```diff
@@ -114,7 +114,12 @@ export class MenuItem {
   getAriaLabel(): string {
     // This fallback should only be hit by Context Menu items as all
     // FieldDropdown options should have an ARIA label.
-    return this.ariaLabel || String(this.content);
+    return (
+      this.ariaLabel ||
+      (typeof this.content === 'string'
+        ? String(this.content)
```
maribethb: nit: if the type is already string you prob don't need to cast it to string

maribethb (approved):

packages/blockly/core/menuitem.ts
gonfunko: Oh great point, just left that as it was but indeed it's unneeded. Removed!

PR #9845, opened 2026-05-08
feat!: Added additional keyboard shortcuts for context menu

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/622

### Proposed Changes

Adds context menu shortcuts for Shift+F10 and the Menu key.

### Reason for Changes

These are common shortcuts for the context menu in Windows.

### Test Coverage

Added unit tests for validating the context menu appears from a block, from a workspace comment, and from the workspace when pressing these key combinations. Tested manually (my keyboard does have a menu button).

### Breaking Changes

New keyboard shortcuts have been registered for Shift+F10 and the Menu/Context Menu key. This will cause an error if you try to register another shortcut with these same keys. If you already have one or both of these keyboard shortcuts registered for something in your code, you will need to either change which key(s) your shortcut uses or remove the new registration(s) to keep your existing shortcut(s).

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -558,13 +558,18 @@ export function registerMovementShortcuts() {
 }
 
 /**
- * Keyboard shortcut to show the context menu on ctrl/cmd+Enter.
+ * Keyboard shortcut to show the context menu on ctrl/cmd+Enter, shift+F10, and
+ * the menu key.
  */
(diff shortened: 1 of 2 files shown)


mikeharv (approved):
Changes look good. Thanks for adding test coverage. I think this may need to be marked as a breaking change, in case developers have already assigned the new shortcuts to other functionality.

PR #9846, opened 2026-05-08
feat: use custom aria labels for inputs with move announcements

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9836

### Proposed Changes

This PR allows move announcements to use `Input.ariaLabelProvider` when disambiguating connection candidates. 

image

If a provider is not provided, or one returns an empty string, we will fallback on the computed label using either the field labels for the input, or, if it's a statement input, a subset of inputs.

image

### Test Coverage

A new test ensures that a move announcement `disambiguates with custom input labels`. 

I also updated a few related tests to assert using string literals rather than duplicate computation logic that lived in the test file. This should make the tests easier to read and more likely to catch accidental regressions.

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -205,11 +205,7 @@ export function getInputLabels(
 ): string[] {
   return block.inputList
     .filter((input) => input.isVisible())
-    .map((input) =>
-      input.getAriaLabelText() !== null
-        ? input.getAriaLabelText()!
-        : input.getLabel(verbosity),
-    );
+    .map((input) => input.getAriaLabelText() ?? input.getLabel(verbosity));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/blockly/core/block_aria_composer.ts
```diff
@@ -205,11 +205,7 @@ export function getInputLabels(
 ): string[] {
   return block.inputList
     .filter((input) => input.isVisible())
-    .map((input) =>
-      input.getAriaLabelText() !== null
-        ? input.getAriaLabelText()!
-        : input.getLabel(verbosity),
-    );
+    .map((input) => input.getAriaLabelText() ?? input.getLabel(verbosity));
```
mikeharv, line 208: This doesn't change anything functionally, but using nullish is safe because it expresses the intended “use this unless it's null” logic directly, while avoiding both a duplicate method call and an unnecessary non-null assertion.

lizschwab: Interesting, I was under the impression that nullish coalescing was unsafe here because empty string evaluates to falsy but looks like the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) says that's not the case. Thanks, I wanted to do that to begin with. :D

lizschwab (approved):
lgtm

PR #9847, opened 2026-05-11
fix: skip empty label fields and dummy/end row inputs for move announcements

mikeharv:
## The basics

### Resolves

Fixes 

### Proposed Changes

This updates the way unlabeled inputs are counted, in order to skip dummy and end row inputs.
It also updates `FieldLabel` to allow for empty string ARIA labels (which are then ignored for input/block).
The **before** screenshot below shows incorrect input counting due to the presence of a dummy input. It also shows an `'empty'` label due to a blank field that is used to assist in laying out the block.
image

After these changes, the input index is reported in a more intuitive way and the 'empty' label field is skipped entirely:
image

### Reason for Changes

These changes are important in order to make sure that inputs are described more helpfully, and that superfluous block layout information is not conveyed to assistive technologies.

### Test Coverage

New tests were added to ensure we do not count dummy and end row inputs when announcing moves involving unlabeled inputs.

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -209,7 +209,7 @@ export function getInputLabels(
 ): string[] {
   return block.inputList
     .filter((input) => input.isVisible())
-    .map((input) => input.getLabel(verbosity, input.getAriaLabelText()));
+    .map((input) => input.getAriaLabelText() ?? input.getLabel(verbosity));
 }
 
 /**
@@ -246,7 +246,7 @@ function getInputLabelsSubset(block: BlockSvg, input: Input): string[] {
     .filter((input) => input.isVisible())
     .map(
       (input) =>
-        input.getLabel(Verbosity.TERSE, input.getAriaLabelText()) ||
+        input.getLabel(Verbosity.TERSE) ||
         Msg['INPUT_LABEL_INDEX'].replace(
           '%1',
           (input.getIndex() + 1).toString(),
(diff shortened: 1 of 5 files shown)


packages/blockly/core/field_label.ts
```diff
@@ -81,6 +81,42 @@ export class FieldLabel extends Field<string> {
...
+   * a separate ARIA property.
+   *
+   * Unlike other built-in fields, FieldLabel does return an empty string when its
+   * value is empty. This is because empty labels are sometimes used for layout
+   * purposes.
+   *
+   * @param includeTypeInfo Whether to include the field's type information in
+   *     the returned label, if available.
+   */
+  computeAriaLabel(includeTypeInfo: boolean = true): string {
```
lizschwab: I believe this should be `override computeAriaLabel` as this class extends `Field` class, which already has a `computeAriaLable` function.

lizschwab (approved):
Once the `override` issue is addressed this is good to merge

packages/blockly/core/block_aria_composer.ts
```diff
@@ -224,11 +228,7 @@ export function getInputLabels(
  * @param input The input that defines the end of the subset.
  * @returns A list of field/input labels for the given block.
  */
-export function getInputLabelsSubset(
-  block: BlockSvg,
-  input: Input,
-  verbosity = Verbosity.STANDARD,
-): string[] {
+function getInputLabelsSubset(block: BlockSvg, input: Input): string[] {
```
lizschwab, line 256: Per [Maribeth's comment](https://raspberrypifoundation.slack.com/archives/C09KAC5AY7N/p1778531603039799?thread_ts=1778531209.958339&cid=C09KAC5AY7N) I just want to call out that this is a change to the signature and could be breaking. I'm not sure whether that makes this a breaking change.

mikeharv: This is internal and new with `v13`, so I don't believe it could be considered breaking.

lizschwab: That makes sense. Resolving conversation :slightly_smiling_face:

PR #9863, opened 2026-05-13
feat!: Improve context announcement keyboard shortcuts

gonfunko:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/764

### Proposed Changes
This PR backports the Shift-I keyboard shortcut that announces the block hierarchy down to the current block, and backports the regular I keyboard shortcut behavior that announces a verbose description of the current block when one is focused.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -870,13 +870,18 @@ export function registerReadExtendedInformation() {
           // The current block was already read out earlier if it has an output
           // connection.
           toAnnounce.push(
-            `Current block: ${computeAriaLabel(block, aria.Verbosity.TERSE, false)}`,
+            Msg['CURRENT_BLOCK_ANNOUNCEMENT'].replace(
+              '%1',
+              computeAriaLabel(block, aria.Verbosity.TERSE, false),
+            ),
           );
         }
 
-        aria.announceDynamicAriaState(`Parent blocks: ${toAnnounce.join(',')}`);
+        aria.announceDynamicAriaState(
+          Msg['PARENT_BLOCKS_ANNOUNCEMENT'].replace('%1', toAnnounce.join(',')),
+        );
       } else {
-        aria.announceDynamicAriaState('Current block has no parent');
+        aria.announceDynamicAriaState(Msg['NO_PARENT_ANNOUNCEMENT']);
       }
       return true;
     },
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -6,6 +6,7 @@
 
 // Former goog.module ID: Blockly.ShortcutItems
 
+import {computeAriaLabel} from './block_aria_composer.js';
 import {BlockSvg} from './block_svg.js';
 import * as clipboard from './clipboard.js';
 import {RenderedWorkspaceComment} from './comments.js';
@@ -63,6 +64,7 @@ export enum names {
   NEXT_STACK = 'next_stack',
   PREVIOUS_STACK = 'previous_stack',
   INFORMATION = 'information',
+  EXTENDED_INFORMATION = 'extended_information',
   PERFORM_ACTION = 'perform_action',
   DUPLICATE = 'duplicate',
   CLEANUP = 'cleanup',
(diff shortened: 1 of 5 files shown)


packages/blockly/core/shortcut_items.ts
```diff
@@ -762,19 +764,32 @@ export function registerFocusToolbox() {
...
+        const description = computeAriaLabel(
+          block,
+          aria.Verbosity.LOQUACIOUS,
+          true,
+        );
+        aria.announceDynamicAriaState(description);
+        return true;
+      }
+
+      // Fall back to describing the state of the workspace.
```
maribethb: I think this shouldn't be a fallback, it should be if the workspace is focused

I would rearrange the logic here to separate this out into helper functions

if (workspace is focused)
  readWorkspaceInfo();
else if (block or part of a block is focused)
  readBlockInfo();
else
  return false because we didn't do anything for this shortcut

packages/blockly/core/shortcut_items.ts
```diff
@@ -811,6 +826,66 @@ export function registerWorkspaceOverview() {
...
+        toAnnounce.reverse();
+        if (!block.outputConnection?.isConnected()) {
+          // The current block was already read out earlier if it has an output
+          // connection.
+          toAnnounce.push(
+            `Current block: ${computeAriaLabel(block, aria.Verbosity.TERSE, true)}`,
+          );
+        }
+
+        aria.announceDynamicAriaState(`Parent blocks: ${toAnnounce.join(',')}`);
```
maribethb: All these strings need to be localized

packages/blockly/core/shortcut_items.ts
maribethb, line 831: please double check that this string doesn't reference only getting info about workspaces

packages/blockly/core/shortcut_items.ts
```diff
@@ -811,6 +826,66 @@ export function registerWorkspaceOverview() {
...
+      } else {
+        aria.announceDynamicAriaState('Current block has no parent');
+      }
+      return true;
+    },
+    keyCodes: [shiftI],
+    displayText: () => Msg['SHORTCUTS_EXTENDED_INFORMATION'],
```
maribethb, line 896: does this already exist? you didn't add it here

packages/blockly/core/shortcut_items.ts
```diff
@@ -811,6 +826,66 @@ export function registerWorkspaceOverview() {
...
+      // from there because the outputs of a block are read anyway, so we don't
+      // need to repeat them.
+      let startBlock = block;
+      while (startBlock.outputConnection?.isConnected()) {
+        startBlock = startBlock.getParent()!;
+      }
+
+      if (startBlock !== block) {
+        toAnnounce.push(
+          computeAriaLabel(startBlock, aria.Verbosity.TERSE, true),
```
maribethb: I think you want to pass false, here and elsewhere, so that custom labels for inputs are not used. if they're present, they'd replace the actual text on the block which is usually undesirable.

we pretty much only want the custom labels for inputs in move mode or if connections are focused

maribethb (commented):
this is a feat, not a fix, and it's a breaking change because you're registering a new keyboard shortcut

packages/blockly/core/shortcut_items.ts
gonfunko: Done

packages/blockly/core/shortcut_items.ts
gonfunko: Done

packages/blockly/core/shortcut_items.ts
gonfunko: Confirmed, it's just "Announce information"

packages/blockly/core/shortcut_items.ts
gonfunko: Added, thanks!

packages/blockly/core/shortcut_items.ts
gonfunko: Done

packages/blockly/msg/json/qqq.json
```diff
@@ -1,18 +1,4 @@
 {
```
maribethb, line 1: normally i manually undo this change but i vote we just keep it off and see what happens

maribethb (approved):

PR #9869, opened 2026-05-13
feat!: Add keyboard shortcut to toggle screenreader mode

### LINKED ISSUE #9861: shortcut for enabling "screenreader optimized mode"
### Check for duplicates

### Problem

_No response_

### Request

add a shortcut that when toggled on:

1. disables navigation looping
2. enables the nesting audio cues

and when toggled off does the opposite

see https://docs.google.com/document/d/1Is3u6-4uavVAFLCnUrYfR2PS8eVgccl3qXjnKtPe7IY/edit?tab=t.0#heading=h.kwr4p1x0g24 for slightly more details

note that in core the shortcut simply enables these and announces it has done so in the aria live region. but there is no UI to accompany this

### Alternatives considered

_No response_

### Additional context

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9861

### Proposed Changes
This PR adds a keyboard shortcut bound to Command Option Z that toggles screenreader mode on or off. With screenreader mode on, navigation looping is disabled and navigating between blocks of different nesting levels plays tones whose frequency corresponds to the level of nesting.

Additionally, error boops are now played when reaching the end of navigation, regardless of whether screenreader mode is on or off.

### DIFF
--- a/packages/blockly/core/keyboard_navigation_controller.ts
+++ b/packages/blockly/core/keyboard_navigation_controller.ts
@@ -12,7 +12,8 @@
 export class KeyboardNavigationController {
   /** Whether the user is actively using keyboard navigation. */
   private isActive = false;
-  private levelChangeBeepsEnabled = false;
+  /** Whether to play audio cues when navigating between scope levels. */
+  private scopeChangeAudioCuesEnabled = false;
   /** Css class name added to body if keyboard nav is active. */
   private activeClassName = 'blocklyKeyboardNavigation';
 
@@ -54,16 +55,16 @@ export class KeyboardNavigationController {
    * Sets whether or not audio cues should be played when keyboard navigation
    * transitions between blocks of different nesting levels.
    */
-  setLevelChangeBeepsEnabled(enabled: boolean) {
-    this.levelChangeBeepsEnabled = enabled;
+  setScopeChangeAudioCuesEnabled(enabled: boolean) {
+    this.scopeChangeAudioCuesEnabled = enabled;
   }
 
   /**
    * Returns whether or not audio cues should be played when keyboard navigation
    * transitions between blocks of different nesting levels.
    */
-  getLevelChangeBeepsEnabled() {
-    return this.levelChangeBeepsEnabled;
+  getScopeChangeAudioCuesEnabled() {
+    return this.scopeChangeAudioCuesEnabled;
   }
 
   /** Adds or removes the css class that indicates keyboard navigation is active. */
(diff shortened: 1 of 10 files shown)


packages/blockly/core/keyboard_navigation_controller.ts
```diff
@@ -12,6 +12,7 @@
 export class KeyboardNavigationController {
   /** Whether the user is actively using keyboard navigation. */
   private isActive = false;
+  private levelChangeBeepsEnabled = false;
```
mikeharv: There's a few different `Beep` functions in properties here, and I'm wondering if those names are maybe too implementation specific. I wonder if it might be clear to name the concept like `AudioCues` rather than the specific chosen sound effect.

Similarly, `level` on its own feels slightly vague. I wonder if there's a way for the property to more clearly signify we're talking about nested block levels/depth.

packages/blockly/core/shortcut_items.ts
```diff
@@ -1107,6 +1122,45 @@ export function registerShowTooltip() {
...
+
+  let enabled = false;
+
+  const toggleScreenreader: KeyboardShortcut = {
+    name: names.TOGGLE_SCREENREADER,
+    preconditionFn: () => true,
+    callback: (workspace) => {
+      enabled = !enabled;
+      keyboardNavigationController.setLevelChangeBeepsEnabled(enabled);
+      workspace.getNavigator().setNavigationLoops(!enabled);
```
mikeharv, line 1148: Since this changes functionality for all users, not just screen readers, would it be appropriate to also show a toast explaining what changed? Otherwise there's no visible feedback or way to understand why navigation behavior has changed (or how to toggle it back).

mikeharv (commented):
These changes look good. My naming comment is a bit nit-picky, but I mention it because now is the easiest time to get it right. Open to your thoughts!

Let me know what you think about adding a toast as well.

packages/blockly/core/shortcut_items.ts
gonfunko: Great idea, switched to a toast (which automatically announces its contents) and added a note about how to turn it on/off.

packages/blockly/core/keyboard_navigation_controller.ts
gonfunko: Updated to "ScopeChangeAudioCue" rather than "LevelChangeBeep" throughout

packages/blockly/msg/json/qqq.json
mikeharv, line 15: Re-add?

mikeharv (approved):

PR #9873, opened 2026-05-14
feat: Adjacent label fields are combined when composing ARIA labels

### LINKED ISSUE #9855: coalesce adjacent label fields
### Check for duplicates

### Problem

_No response_

### Request

some blocks visually have adjacent labels that are actually part of different inputs. e.g. first part of the string belongs to a dummy input and the second part belongs to a value input. this is due to various reasons with block design.

currently, all fields have their screenreader output separate by commas. but if two label fields are visually adjacent then there are extra commas.

we want to remove the extra commas between adjacent label fields (in other words, coalesce adjacent label fields into one "piece" of the screenreader output)

for the purpose of this issue a label field is any subclass of `FieldLabel` and "adjacent" means in subsequent field position on the block but respecting "rows" (e.g. a label field at the end of an end-row input is not adjacent to a label field in a statement input on the following row)

Example block definition:

```js
Blockly.Blocks['device_show_number'] = {
  init: function () {
    this.appendDummyInput()
      .appendField('show');

    this.appendValueInput('number')
      .setCheck('Number')
      .appendField('number');

    this.setNextStatement(true);

    this.setColour('#1e90ff');
  }
};
```

### Alternatives considered

_No response_

### Additional context

_No response_

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9855

### Proposed Changes

When composing an ARIA label, adjacent field labels will now be combined together so that they no longer have commas separating them.

### Reason for Changes

The way some custom blocks are composed, multiple field labels are sometimes generated which are visually part of the same label but which were being separated by commas in the ARIA label. This fixes it so that the screen reader and visual experience are closer aligned.

### Test Coverage

Added and updated unit tests and manually tested with a custom block in the playground.


packages/blockly/core/block_aria_composer.ts
```diff
@@ -140,9 +145,34 @@ export function computeFieldRowLabel(
...
+          adjacentFieldLabels.push(field.computeAriaLabel(includeTypeInfo));
+          return [];
+        } else if (adjacentFieldLabels.length >= 1) {
+          // There is at least one adjacent FieldLabel before this one but none
+          // after. Combine the FieldLabels into one string.
+          adjacentFieldLabels.push(field.computeAriaLabel(includeTypeInfo));
+          const label = adjacentFieldLabels.join(' ');
+          adjacentFieldLabels = [];
+          return label;
+        } else {
```
gonfunko: This else can just be removed in favor of falling through to the return at the end of the `flatMap` callback

gonfunko (approved):

packages/blockly/core/block_aria_composer.ts
lizschwab: Yeah, I went back and forth on that one. The way it is now makes it clear that it's intentional, but it's also unnecessary because it will fall through. I'll update it :).

PR #9902, opened 2026-05-19
feat: Add custom input labels to built in blocks

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9660

### Proposed Changes

This adds custom ARIA labels to a bunch built-in blocks, and also adds custom ARIA type names to fields on built in procedure blocks.

#### Input Labels
To determine which blocks needed input labels, a full audit was performed on all built-in blocks that show up in the Categories toolbox of the public playground.

The rule of thumb here was to address blocks where the default computed label, based on block fields, was either missing or unclear. For example, the inputs on the `logic_compare` block would otherwise be labeled as "input 1" and "equals" (or other dropdown value ARIA label). These are changed to "first value" and "second value".
image

The decision to use terms like "condition", "number", "text", and "value" is based on the expected block. While I tried to avoid being overly opinionated, it's impossible to come up with a set of labels all developers will think work equally well. Inputs labels can be specified/overridden either by:
1. providing a `ariaLabelText` (string) with the block definition
2. calling `input.setAriaLabelProvider` (with a string or function) after the block has been created.

The most interesting label providers here are for the input for `controls_whileUntil` and `lists_split`. These label will change depending on the selected value in the blocks `"MODE"` dropdown . The provider for `controls_whileUntil` is specified with an extension because the block definition is in JSON. This is a familiar pattern that other Blockly developers may already be familiar with.

image
image

The proposed changes are shown visually here:
https://docs.google.com/document/d/1VTi1tXK3QrmOIqoDcZ6kjWGO9DzEiiW_CvDxRV4hlsA/edit?tab=t.0#heading=h.xkgll8dimyia

#### Field Type Names
Typically each field subclass specifies its own safe default type name, such as "number" or "dropdown".  In the vast majority of cases, customizing these for specific built-in blocks is not needed. The only ones that seemed worth touching were the name fields for procedure definition and argument blocks. 

imageimage

These were squeezed into the same branch to avoid merge conflicts from competing branches both adding new messages ahead of cutting a new Blockly release.

### Test Coverage

Existing tests were updated to use the new labels. Tests that specifically checks the behavior of unlabeled inputs needed to be updated to remove the `ariaLabelProvider` from blocks in the test instance.


packages/blockly/msg/messages.js
```diff
@@ -1959,7 +1959,7 @@ Blockly.Msg.KEYBOARD_NAV_BLOCK_NAVIGATION_HINT = 'Use %1 to navigate inside of b
 /// Message shown when a user presses Enter with the workspace focused.
 Blockly.Msg.KEYBOARD_NAV_WORKSPACE_NAVIGATION_HINT = 'Use the arrow keys to navigate.';
 /** @type {string} */
-/// Message shown when a user presses Enter with a flyout label (heading) focused.
+/// Message shown when a user presses Enter with a flyout label (heading) focused. Placeholder %1 is the keyboard shortcut for navigating to the next heading.
```
mikeharv, line 1962: This was mistakenly removed during a bad merge conflict.

https://github.com/RaspberryPiFoundation/blockly/pull/9874/changes#r3269203999

packages/blockly/msg/json/en.json
```diff
@@ -522,6 +522,55 @@
 	"INPUT_LABEL_STATEMENT": "command position",
 	"INPUT_LABEL_END_STATEMENT": "End %1",
 	"INPUT_LABEL_EMPTY": "Empty",
+	"INPUT_LABEL_CONDITION": "condition",
+	"INPUT_LABEL_CONDITION_A": "first condition",
+	"INPUT_LABEL_CONDITION_B": "second condition",
+	"INPUT_LABEL_CONDITION_UNTIL": "condition to stop repeating",
```
maribethb: I think I might just call these both "condition" here. you can always press `i` to hear the full block text again, so there's no need to go overly specific.

"condition to stop repeating" is easy to misread as "thing you want to stop doing" instead of "reason you will stop", that's why I would just call it condition.

packages/blockly/msg/json/en.json
```diff
@@ -522,6 +522,55 @@
...
+	"INPUT_LABEL_NUMBER": "number",
+	"INPUT_LABEL_NUMBER_A": "first number",
+	"INPUT_LABEL_NUMBER_B": "second number",
+	"INPUT_LABEL_NUMBER_TO_CHECK": "number to check",
+	"INPUT_LABEL_NUMBER_LIST": "list of numbers",
+	"INPUT_LABEL_MATH_DIVIDEND": "dividend",
+	"INPUT_LABEL_MATH_DIVISOR": "divisor",
+	"INPUT_LABEL_MATH_CHANGE_BY": "amount to change by",
+	"INPUT_LABEL_MATH_CONSTRAIN_VALUE": "number to constrain",
+	"INPUT_LABEL_NUMBER_MIN": "minimum",
```
maribethb, line 539: I think labels that largely repeat text that is already present before the input on the block, should probably be removed. For example if you don't have a custom label then the input will already read "maximum of list" so min/max might be redundant.

whereas if the text does not come before the input then the default label is not as readable so the custom label with more context is more useful

packages/blockly/msg/json/en.json
```diff
@@ -522,6 +522,55 @@
...
+	"INPUT_LABEL_TEXT_POSITION": "letter position",
+	"INPUT_LABEL_TEXT_START_POSITION": "start position",
+	"INPUT_LABEL_TEXT_END_POSITION": "end position",
+	"INPUT_LABEL_TEXT_PROMPT_MESSAGE": "message",
+	"INPUT_LABEL_VARIABLES_SET": "value to set",
+	"INPUT_LABEL_LISTS_CREATE_WITH_ITEM": "value %1",
+	"INPUT_LABEL_LISTS_REPEAT_ITEM": "value to repeat",
+	"INPUT_LABEL_LISTS_REPEAT_NUM": "number of times to repeat",
+	"INPUT_LABEL_LISTS_TO_CHECK": "list to check",
+	"INPUT_LABEL_LISTS_VALUE_TO_FIND": "value to find",
```
maribethb: this is another one where i think the default label might be better ("find [first] occurrence of item, input position")

maribethb (approved):
there are a couple i would tweak but overall looks good!

packages/blockly/msg/json/en.json
mikeharv: These strings aren't actually used for list blocks. I agree with this feedback for the `math_constrain` block as it has suitable enough labels already. However, for `math_random_int` the default labels end up as `"random integer from"` and `"to"`, so I'm going to leave them for that block.

PR #9903, opened 2026-05-19
fix: Improve initial block placement

### LINKED ISSUE #9878: [v13.0.0-beta.4] Initial placement of top-level blocks inserted by keyboard
### Check for duplicates

### Description

The keyboard nav plugin has reasonable initial placement of top-level blocks inserted via keyboard, e.g.

Image

The Blockly core implementation just makes a pile in the top left, perfectly and confusingly overlapped.

Interestingly, if I also add the other type of function available in the playground I get a separate pile below:

Image

There are many of each block above.

This is a big step back for clients like MakeCode where the program is structured via top-level event handlers.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9878

### Proposed Changes
This PR improves the placement of newly-inserted blocks when using keyboard navigation. It's a near-exact backport of the logic from the keyboard-experimentation repo, with the adjustment of increasing the bounds of existing blocks to avoid placing the new block within snapping range. Generally new blocks will be placed wherever free space is available while scanning across the workspace row by row.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -112,10 +112,9 @@ export class BlockDragStrategy implements IDragStrategy {
   /**
    * Positions a cloned block on its new workspace.
    *
-   * @param oldBlock The flyout block that was cloned.
    * @param newBlock The new block to position.
    */
-  private positionNewBlock(oldBlock: BlockSvg, newBlock: BlockSvg) {
+  private positionNewBlock(newBlock: BlockSvg) {
     const workspace = newBlock.workspace;
     const initialY = 10;
     const initialX = 10;
@@ -215,7 +214,7 @@ export class BlockDragStrategy implements IDragStrategy {
           },
         ) as BlockSvg;
         eventUtils.setRecordUndo(false);
-        this.positionNewBlock(this.block, newBlock);
+        this.positionNewBlock(newBlock);
         eventUtils.setRecordUndo(true);
 
         return newBlock;
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -93,6 +93,11 @@ export class BlockDragStrategy implements IDragStrategy {
 
   protected readonly BLOCK_CONNECTION_OFFSET = 10;
 
+  /**
+   * How far in from the edges of the workspace to position newly placed blocks.
+   */
+  protected readonly WORKSPACE_MARGIN = 10;
+
   constructor(private block: BlockSvg) {
     this.workspace = block.workspace;
   }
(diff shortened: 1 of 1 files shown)


packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -112,19 +112,83 @@ export class BlockDragStrategy implements IDragStrategy {
...
-    const workspaceCoordinates = svgMath.screenToWsCoordinates(
-      newBlock.workspace,
-      screenCoordinate,
-    );
-    newBlock.moveDuringDrag(workspaceCoordinates);
+  private positionNewBlock(newBlock: BlockSvg) {
+    const workspace = newBlock.workspace;
+    const initialY = 10;
+    const initialX = 10;
+    const xSpacing = 80;
```
mikeharv, line 128: What would think about replacing both `initialX` and `initialY` with a single `WORKSPACE_MARGIN` or similar, to help explain the magic numbers? 
`xSpacing` could also maybe be clearer as `NEW_BLOCK_HORIZONTAL_SPACE`, although I realize that's also pretty lengthy. I'm pointing these out because I had to read through all the code below before I could understand the purpose of these.

packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -112,19 +112,83 @@ export class BlockDragStrategy implements IDragStrategy {
...
+      .filter((block) => block.id !== newBlock.id);
+    const allBlockBounds = filteredTopBlocks.map((block) => {
+      const bounds = block.getBoundingRectangle();
+      // Expand the bounds to avoid the new block being placed within snapping
+      // distance.
+      return new Rect(
+        bounds.top - 30,
+        bounds.bottom + 30,
+        bounds.left - 30,
+        bounds.right + 30,
```
mikeharv: Is this padding based on `snapRadius`? Is that a configurable value? Wondering if we should avoid `30` as a magic number here. At the least, could we use a named constant?

mikeharv (approved):

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Done, added a comment for xSpacing

packages/blockly/core/dragging/block_drag_strategy.ts
gonfunko: Yes, updated accordingly.

mikeharv (approved):

microbit-robert:
@gonfunko I think this has negative impacts on dragging blocks from the flyout with the mouse.

gonfunko:
Ah good catch, working on a fix forward

PR #9905, opened 2026-05-20
chore(docs): switch documentation deployment target

grega:
## The basics

### Proposed Changes

Switches documentation deployment target from GitHub Pages to Cloudflare Pages.

Preview build: https://blockly-docs.pages.dev

### Reason for Changes

Ensures that the [existing GH Pages workflow](https://github.com/grega/blockly/blob/main/.github/workflows/update-gh-pages.yml) can continue to run without needing modification (ie. without needing to figure out how to deploy both sets of assets to GH Pages without overwriting each other).

Prior to running this workflow for the first time, two secrets will need adding to this repository:

- `CLOUDFLARE_ACCOUNT_ID`
- `CLOUDFLARE_API_TOKEN`

I can provide values for each.

### DIFF
--- a/packages/docs/static/_redirects
+++ b/packages/docs/static/_redirects
@@ -1,0 +1,10 @@
+# Cloudflare Pages Redirects
+# Syntax: source destination [status_code]
+# https://developers.cloudflare.com/pages/configuration/redirects/
+
+# Legacy /reference/js/* paths redirect to canonical /reference/* URLs
+/reference/js/*   /reference/:splat   302
+
+# Strip .md/.mdx extensions if someone lands on the raw filename
+/reference/*.md   /reference/:splat   302
+/reference/*.mdx  /reference/:splat   302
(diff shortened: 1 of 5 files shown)


grega:
@maribethb has the context here @lizschwab, so feel free to pass on the review if you prefer!

packages/docs/static/_redirects
```diff
@@ -0,0 +1,10 @@
+# Cloudflare Pages Redirects
+# Syntax: source destination [status_code]
+# https://developers.cloudflare.com/pages/configuration/redirects/
+
+# Legacy /reference/js/* paths redirect to canonical /reference/* URLs
+/reference/js/*   /reference/:splat   302
+
+# Strip .md/.mdx extensions if someone lands on the raw filename
+/reference/*.md   /reference/:splat   302
+/reference/*.mdx  /reference/:splat   302
```
grega, line 10: I used `302` throughout as I'm generally hesitant about applying truly _permanent_ redirects with `301` (not that all user-agents necessarily stick to the spec here, though some will certainly cache it with no expiry).

That said, the `.md` and `.mdx` redirects feel in particular like a `301` would be fine.

lizschwab (approved):

packages/docs/static/_redirects
maribethb: I think 301 is probably fine, but I'm happy to make sure everything is set up correctly before we start claiming that our redirects are permanent!

maribethb (approved):

PR #9908, opened 2026-05-20
fix: CSS back to <style> for Safari/cascade reasons

### LINKED ISSUE #9876: [v13 beta] Consider Safari verison baseline for v13 vs CSS injection change
### Check for duplicates

### Description

v13 beta changes the browser baseline in two ways that I can see:

- :has CSS pulled in from the keyboard navigation plugin, this is expected and was discussed last year ([requires Safari 15.4](https://caniuse.com/?search=%3Ahas)), it's possible there are more involved ways to the same end. No strong view from me but would be good to cover in release notes.
- Constructable stylesheets + adoptedStylesheets call added on #9611, [this requires Safari 16.4](https://caniuse.com/?search=adoptedstylesheets) which is an unexpected baseline raise for us that will affect MakeCode's supported versions and require raising minimum versions for apps. This is also a start-up crash rather than degrade.

Is it viable to fall back to the old approach for #9611 with feature detection / error handling? I can take a look at a PR for this if that's an acceptable path forward.

### Reproduction steps

1. Use Safari 15, e.g. via browserstack
2. Load v13 branch playground
3. Note no workspace and logged error.

### Priority

I think this is pretty straightforward to address if there's the willingness to maintain a fallback path.

It's possible you could push apps to use a polyfill instead. I haven't tested any with this scenario.

### Stack trace

```shell
[Error] TypeError: Illegal constructor
	CSSStyleSheet (blockly_compressed.js:134:245)
	register$$module$build$src$core$css (blockly_compressed.js:134:245)
	(anonymous function) (blockly_compressed.js:1422:482)
	(anonymous function) (blockly_compressed.js:10)
	Global Code (blockly_compressed.js:12)
```

### Screenshots

Image

I tested Safari desktop but more worried about edu iPads in practice.

### Browsers

Safari desktop

microbit-matt-hillsdon:
Reverts the storage mechanism introduced in #9611 (constructable stylesheets via `adoptedStyleSheets`) while keeping the per-root injection-site tracking that #9611 added for shadow-DOM support.

Motivations:

- Safari 15.4 compatibility. `new CSSStyleSheet()` and `adoptedStyleSheets` require Safari 16.4+
- Cascade order. `adoptedStyleSheets` apply after `<style>`/`<link>` elements in the document, so Blockly's defaults silently overrode host stylesheets. Prepending a `<style>` to the head (or to the shadow root) restores the pre-#9611 behavior where any author stylesheet declared later wins on specificity ties.

Trade-offs:

- Per-shadow-root CSS text is duplicated rather than shared via a single adopted sheet object. Negligible for typical use.
- `Css.register()` calls made after the first `inject()` no longer reach already-injected roots (same as #9611's behavior); subsequent `inject()` calls into other roots still pick them up. Web-component consumers can legitimately register late, so this is preferred to reinstating the pre-#9611 throw.

@gonfunko this is one possible take with the downsides noted above. Another option is to use both mechanisms (tag and adoptedStylesheets) either based on feature detection or based on the root. Then to fix the cascade issue consistently use "`@layer` blockly" for Blockly's CSS. That will require auditing of the !important use in Blockly as they would become [impossible to override](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer#:~:text=As%20noted%20in%20the%20above%20diagram). If relying on feature detection note it is not sufficient to check for CSSStyleSheet being defined as that significantly predates constructable stylesheets.

Please feel free to make changes to this PR and/or replace it with one of your own.

## The basics

~~I didn't format as it made a large unrelated diff~~ whoops, I had outdated prettier, all good now.

### Resolves

Fixes #9876

### Test Coverage

I checked the web component playground added in the previous PR.

### DIFF
--- a/packages/blockly/core/css.ts
+++ b/packages/blockly/core/css.ts
@@ -5,9 +5,8 @@
  */
 
 // Former goog.module ID: Blockly.Css
-/** Has CSS already been injected? */
 const injectionSites = new WeakSet<Document | ShadowRoot>();
-const registeredStyleSheets: Array<CSSStyleSheet> = [];
+const registeredCss: Array<string> = [];
 import * as userAgent from './utils/useragent.js';
 
 /**
@@ -17,11 +16,7 @@ import * as userAgent from './utils/useragent.js';
  * @param cssContent Multiline CSS string or an array of single lines of CSS.
  */
 export function register(cssContent: string) {
-  if (typeof window === 'undefined' || !window.CSSStyleSheet) return;
-
-  const sheet = new CSSStyleSheet();
-  sheet.replace(cssContent);
-  registeredStyleSheets.push(sheet);
+  registeredCss.push(cssContent);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


gonfunko (approved):
Thanks for putting this together, looks reasonable to me but I'll let Maribeth have a look before merging.

gonfunko (approved):
Thanks for putting this together, looks reasonable to me but I'll let Maribeth have a look before merging.

maribethb (approved):

PR #9911, opened 2026-05-20
fix: Don't prepend "begin stack" to blocks inserted from the flyout

microbit-robert:
## The basics

### Resolves
Omits the "begin stack" label to blocks inserted from the flyout in move mode. At this point, the block can be positioned anywhere. It is arguably only a stack once is has been placed on the workspace in a certain position.

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9802

### Proposed Changes
Use the `block.isDragging` method to determine whether a block is inserted on the workspace in move mode.

### Reason for Changes
This reduces the amount of information in the screen reader output which should make it easier to parse for screen reader users. Arguably improves the label accuracy.

### Test Coverage
None

### Documentation
No

### Additional Information
N/A

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -71,7 +71,9 @@ export function computeAriaLabel(
     }
   }
   return [
-    verbosity >= Verbosity.STANDARD && getBeginStackLabel(block),
+    verbosity >= Verbosity.STANDARD &&
+      !block.isDragging() &&
+      getBeginStackLabel(block),
     getParentInputLabel(block),
     ...getInputLabels(block, verbosity, useCustomInputLabels),
     verbosity === Verbosity.LOQUACIOUS && getParentToolboxCategoryLabel(block),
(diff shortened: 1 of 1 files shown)


microbit-robert:
Hmm, moving back to draft.

It feels like `block.workspace.isDragging()` should cover this case in the `getBeginStackLabel` function, but does not. What is the difference between `block.workspace.isDragging()` and `block.isDragging()`?

microbit-robert:
Ok. `KeyboardMover.mover.isMoving()` is false at the time the block label is calculated. Presumably this is because the block is place on the workspace and then put into move mode.

I propose switching `block.workspace.isDragging()` for `block.isDragging()` which is true at the time the label is calculated. This address removing the "being stack" prefix for blocks inserted from the flyout and for other blocks already on the workspace put into move mode. Is there any reason we can't do this?

maribethb:
`workspace.isDragging` is I think if the workspace itself is being dragged. I think that was just straight up a mistake in the original PR, whoops. Thanks for catching this!

maribethb (approved):

PR #9916, opened 2026-05-21
fix: Block labels should not use custom input labels (nested inputs)

microbit-robert:
## The basics

### Resolves

Builds on https://github.com/RaspberryPiFoundation/blockly/pull/9886 to ensure that block labels do not use custom input labels if the block contains nested value blocks, etc. This change also ensures commas are consistent between inputs for nested value blocks, etc.

### Proposed Changes

Push `useCustomLabels` param through to the input `getLabel` method.

### Reason for Changes

Current behaviour: Nested block has focus, block label is as expected:
Screenshot 2026-05-19 at 11 09 14

Current behaviour: Parent block has focus, block label for nested block is using custom input labels:
Screenshot 2026-05-19 at 11 09 25

This changes fixes scenario 2, so that the output is now:
Screenshot 2026-05-21 at 12 29 00

This seems to be missing a comma between "0 y", but will investigate this separately. Update: I pushed a change to fix this as well since it is trivial.

### Test Coverage

None added.

### Documentation

No

### Additional Information

N/A

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -308,7 +308,7 @@ export function getInputLabels(
 
   return inputsToLabel.map((input) => {
     const customLabel = useCustomLabels ? input.getAriaLabelText() : null;
-    return customLabel ?? input.getLabel(verbosity);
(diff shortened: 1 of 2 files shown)


mikeharv (approved):
Thanks for the fix!

PR #9921, opened 2026-05-21
feat: Reduce build size by enabling `assume_function_wrapper` build flag

### LINKED ISSUE #5795: Enable `--assume_function_wrapper` Closure Compiler option

A feature request that will help with the problem of the size of `blockly_compressed.js` having grown substantially over the last two quarters.

Enable the `--assume_function_wrapper` option in the `options` object passed to `compile` by `buildCompiled` in `scripts/gulpfiles/build_tasks.js`.  This would (as of this writing) reduce the size of `blockly_compressed.js` from 907 to 623 KiB, and `blockly_compressed.js.gz` from 168 to 151 KiB.

Enabling `ADVANCED_OPTIMIZATIONS` would have an even greater effect, but would (absent additional work to prevent this) result in most of the export properties (and class `.prototype` properties) being renamed, rendering our API hopelessly broken.

**Additional context**

The one issue to be resolved is that `--assume_function_wrapper` causes the top-level `Blockly` export (i.e., the one created by the call to `goog.module.declareLegacyNamespace()` in `core/blockly.js`) to be renamed from `$.Blockly` to `$.X` (where `X` is, by chance literally an X, but could be any short 1–2 letter name), breaking the UMD wrapper.

We will need to determine how best to arrange for the UMD wrapper to return the `Blockly` exports object despite this renaming.

### LINKED ISSUE #7551: Remove `$build$src` from variable names in compressed chunks
### Check for duplicates

### Problem

In commit 06d78af of PR #7473 I made a change that reduced the size of the compressed chunks by shortening the lengths of the generated variable names by shortening the pathnames fed to Closure Compiler.

Unfortunately this broke sourcemaps because it caused the gulp-sourcemaps plugin to omit the inline source from the generated `<chunk>_compressed.js.map` files, so the change was rolled back in pr #7550.

It should be possible to shorten the file paths fed to Closure Compiler, and thus the size of the compressed chunks, without breaking sourcemaps, but there was insufficient time to finish investigating how. 

### Request

Update the `buildCompiled` gulp task (`npm run minify`) to shorten the filenames fed to Closure Compiler—without breaking sourcemaps.

### Alternatives considered

_No response_

### Additional context

`DEBUG='gulp-sourcemaps:*' npm run minify` will run the task with debugging enabled for the sourcemaps package, which looks like it should make this task much easier.

gonfunko:
## The basics

- [X] @cpcallen validated the LLM's changes.

### Resolves

* Fixes #5795
* Fixes #7551

### Proposed Changes
**This is a largely LLM-generated change**. It does seemingly work, and cuts the size of blockly_compressed.js from 1093 KB (216 KB zipped) to 624 KB (176 KB zipped). I split it into two commits; the first is just a formatter run, the second has the actual changes. 

Christopher then added additional commits to clean up the LLM-generated changes and make them more human-readable.

### DIFF
--- a/packages/blockly/core/blockly.ts
+++ b/packages/blockly/core/blockly.ts
@@ -234,8 +234,6 @@ import {ZoomControls} from './zoom_controls.js';
  * This constant is overridden by the build script (npm run build) to the value
  * of the version in package.json. This is done by the Closure Compiler in the
  * buildCompressed gulp task.
- * For local builds, you can pass --define='Blockly.VERSION=X.Y.Z' to the
- * compiler to override this constant.
  *
  * @define {string}
  */
--- a/packages/blockly/scripts/gulpfiles/build_tasks.mjs
+++ b/packages/blockly/scripts/gulpfiles/build_tasks.mjs
@@ -196,7 +196,7 @@ const CHUNK_EXPORTS_DIR = 'chunk_exports';
  * @param {{name: string}} chunk
  * @return {string}
  */
-function chunkExportPath(chunk) {
+function chunkExporterPath(chunk) {
   return path.posix.join(CHUNK_EXPORTS_DIR, `${chunk.name}_export.js`);
 }
 
@@ -217,12 +217,12 @@ async function buildChunkExporters() {
 
   await Promise.all(
     chunks.map(async (chunk) => {
-      const exportFile = chunkExportPath(chunk);
+      const filename = chunkExporterPath(chunk);
       const importPath = posixPath(
-        path.posix.relative(path.posix.dirname(exportFile), chunk.entry),
+        path.posix.relative(path.posix.dirname(filename), chunk.entry),
       );
       await fsPromises.writeFile(
-        path.join(TSC_OUTPUT_DIR, exportFile),
+        path.join(TSC_OUTPUT_DIR, filename),
         `/** @fileoverview @suppress {undefinedVars} */
 
 import * as exports from '${importPath}';
@@ -567,7 +567,7 @@ function getChunkOptions() {
     const files = globs
       .flatMap((glob) => globSync(glob, {cwd: TSC_OUTPUT_DIR_POSIX}))
       .map((file) => path.posix.join(TSC_OUTPUT_DIR_POSIX, file));
-    files.push(path.posix.join(TSC_OUTPUT_DIR_POSIX, chunkExportPath(chunk)));
+    files.push(path.posix.join(TSC_OUTPUT_DIR_POSIX, chunkExporterPath(chunk)));
     chunkOptions.push(
       `${chunk.name}:${files.length}` +
         (chunk.parent ? `:${chunk.parent.name}` : ''),
--- a/packages/blockly/scripts/gulpfiles/build_tasks.mjs
+++ b/packages/blockly/scripts/gulpfiles/build_tasks.mjs
@@ -16,7 +16,6 @@ import sourcemaps from 'gulp-sourcemaps';
 import {execSync} from 'child_process';
 import * as fs from 'fs';
 import * as fsPromises from 'fs/promises';
-import {finished} from 'node:stream/promises';
 import * as path from 'path';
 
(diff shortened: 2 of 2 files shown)


lizschwab (approved):

cpcallen:
It took a while to remind myself how the code in `build_tasks.mjs` works, and a bit longer to remind myself how the code it _generates_ works, but after fairly carefully inspecting the `*_compressed.js` minified bundles I believe the changes made by the LLM are basically correct, and that the bundled chunks should work properly.

A few notes about Closure Compiler's behaviour:

* Even with only `SIMPLE_OPTIMIZATIONS`, the compiler was already renaming method parameters and local variables to short names (`a`, `b`, `c`, etc.).  That's great.
* Additionally, `SIMPLE_OPTIMIZATIONS` causes Closure Compile to replace many property accesses with variables—e.g. instead of using property `foo.bar.baz`, it will mostly use `var foo$bar$baz` internally instead.
  * In order to ensure that this "optimisation" doesn't break external client code, it will emit bookkeeping code like `foo.bar.baz=foo$bar$baz` as necessary.
  * The long variable names and extra bookkeeping actually bloat the output somewhat—and this bloat got worse when we moved to `goog.module`s and later to ES modules, as the compiler now gives all those variables a `module$build$src$…` prefix (hence #7551).
  * Unfortunately, the one bit of bookkeeping code that the compiler doesn't generate is actually the topmost chunk-level export—i.e., it would be very convenient if the `blockly_compressed.js` chunk included a `var Blockly = module$build$src$core$blockly;`, and then the factory function in the chunk wrapper for this chunk could just `return Blockly;`.  Fortunately since the mapping from soure filename to internal module variable name was fixed it was easy to do compute this name using the `modulePath` function, and just have the wrapper `return module$build$src$core$blockly;`.
* Enabling `--assume-function-wrapper` allows the compiler to aggressively rename basically all these variables (as well as related properties on the namespace object), giving them very short (1–3 character) names instead—e.g., where it was already replacing use of `foo.bar.baz` with `var foo$bar$baz`, it will now use something like `var p` instead.  This makes the compressed chunks much smaller!
  * Fortunately (and unlike with `ADVANCED_OPTIMIZATIONS`) Closure Compiler continues to emit the extra bookkeeping, so there will be a statement `foo.bar.baz=p;` to ensure that external code (i.e., project importing Blockly) still work as expected.
* The major problem with turning on `--assume function-wrapper` is that it causes the compiler to rename `module$build$src$core$blockly` to something short but arbitrary (currently `R`, as it happens), making it difficult to find and return the correct object from the wrapper's factory function.

To work around this, the LLM proposes adding an extra source file for each chunk to the compiler's input, that imports the chunk and saves the resulting exports (module) object to a well-known location—in this case, on the namespace object that was introduced in #5721.  Specifically, for the core chunk, a file `blockly_exports.js` contains:

```JS
import * as exports from `…/core/blockly.js`;
$['__chunkExport__'] = exports;
```
would be inserted into the source files for the chunk, and then the chunk wrapper's factory function would `return $['__chunkExport__'];`.

This does indeed work.  My main critiques are:

- Each chunk uses the same property on `$`, overwriting the value from the previous chunk.  Since the property is only used as temporary storage within the factory function that's not a bug, but it did make it harder to understand the code generated by the compiler.
- It is a bit confusing that these extra source files are called "chunk exports", when that name sounds like it is referring the the chunk's export object.

I have therefore taken the liberty of pushing a few additional commits to this PR, which:

- Use a separate property, `$['__chunk_' + chunk.name]`, for each chunk's export object.
- Consistently call the files that create these properties "chunk exporters".
- Tweak the code in the chunk wrapper so that these properties are not accessed with `[]` in some cases and `.` in others, just in case anyone tries feeding the compiled chunks back in to Closure Compiler.
- Improve naming, code comments and code ordering to improve readability.

I have manually checked to make sure the resulting compiled chunks load locally in the playground.

maribethb (approved):
Thanks Christopher, your changes and review here are invaluable!

I also did additional manual verification of the compressed files before merging.

PR #9935, opened 2026-05-26
feat: Enable blocks during drags

### LINKED ISSUE #9910: [13.0.0-beta.5] Drag or keyboard move of a disabled block should render as enabled during drag/move
### Check for duplicates

### Description

In v12 dragging a disabled block remains disabled. But in v12 + keyboard nav plugin dragging or keyboard moving a disabled block enables it for the duration of the drag/move, which we did to improve the understanding of the connection previews.

Plugin implementation: https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/blob/main/src/disabled_blocks.ts#L17-L44

I think a case can be made for just doing this across the board but will check with the MakeCode team later today. Perhaps an explicitly disabled block in Blockly should be treated differently.

In MakeCode the typical scenario is a block that's disabled because it's on the workspace but should be moved to a top-level event block or value input. It's hard to reproduce that exactly in the playground.

### Reproduction steps

1.
2.
3.

### Priority

Looking for a stance initially: was this missed or intentionally dropped? I think it's all public API so it could be an application concern, but consistency might be good for something like this. Was the effect on drags intended or just a side effect of getting it right for keyboard moves?

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9910

### Proposed Changes
This PR enables disabled blocks during a drag and re-disables them once the drag terminates, whether by being committed or cancelled. This applies to both mouse- and keyboard-driven drags.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -92,6 +92,12 @@ export class BlockDragStrategy implements IDragStrategy {
   /** Used to persist an event group when snapping is done async. */
   private originalEventGroup = '';
 
+  /**
+   * Map from block IDs to reason(s) why it was disabled, used to restore
+   * disabled state post-drag.
+   */
+  private lastBlockDisabledReasons: Map<string, Set<string>> = new Map();
+
   protected readonly BLOCK_CONNECTION_OFFSET = 10;
 
   /**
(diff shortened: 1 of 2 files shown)


maribethb (approved):
pls fix pr title but LGTM

PR #9952, opened 2026-06-02
fix: Fix styling and focusing of connections in shadow DOM

gonfunko:
## The basics

### Proposed Changes
This PR fixes two related issues:

* Connections could not be navigated to when Blockly was used in the shadow DOM, because `document` was used to find the highlight SVG, and that refers to the document outside the scope of the shadow root. Now, the shadow root or document, as appropriate, is used.
* Some focus-related styling was not displayed correctly inside the shadow DOM, because the `blocklyKeyboardNavigation` class was being added to `document.body`. Now, it is added to the parent of the injection div, which exists when Blockly is injected in or outside of the shadow DOM.

### DIFF
--- a/packages/blockly/core/rendered_connection.ts
+++ b/packages/blockly/core/rendered_connection.ts
@@ -719,9 +719,10 @@ export class RenderedConnection
   private findHighlightSvg(): SVGPathElement | null {
     // This cast is valid as TypeScript's definition is wrong. See:
     // https://github.com/microsoft/TypeScript/issues/60996.
-    return document.getElementById(this.id) as
-      | unknown
-      | null as SVGPathElement | null;
+    const root = this.getSourceBlock().getSvgRoot().getRootNode() as
+      | ShadowRoot
+      | HTMLDocument;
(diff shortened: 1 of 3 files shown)


packages/blockly/core/rendered_connection.ts
```diff
@@ -719,9 +719,10 @@ export class RenderedConnection
   private findHighlightSvg(): SVGPathElement | null {
     // This cast is valid as TypeScript's definition is wrong. See:
     // https://github.com/microsoft/TypeScript/issues/60996.
-    return document.getElementById(this.id) as
-      | unknown
-      | null as SVGPathElement | null;
+    const root = this.getSourceBlock().getSvgRoot().getRootNode() as
+      | ShadowRoot
+      | HTMLDocument;
+    return root.getElementById(this.id) as SVGPathElement | null;
```
maribethb, line 725: surprised to find you don't need the `as unknown` part of the cast here, as implied by the linked issue. you could remove the comment with the typescript issue since that's the case.

maribethb (approved):

PR #9958, opened 2026-06-05
fix: allow nav into collapsed block row via right arrow

microbit-matt-hillsdon:
This supports e.g. expand/collapse icons added outside Blockly.

MakeCode has an expand FieldImageNoText on collapsed blocks that wasn't otherwise navigable (inserted into COLLAPSED_INPUT_NAME):

```javascript
input.appendField(new FieldImageNoText(image, 24, 24, "Expand", () => {
    this.setCollapsed(false)
}, false));
```

To support this:
1. the input that shares the block row id takes into account visibility
2. we drop the short circuit for collapsed blocks and rely on the filtering already in place to filter out the collapsed content

Also filter icons whose updateCollapsed() hides them via display:none; without this they remain in the candidate list and the navigator can focus e.g. an invisible cog for a mutator workspace.

## The basics

### Resolves

Fixes 

### Reason for Changes

We considered a custom navigation policy here but I think others will likely hit this too.

### Test Coverage

Added webdriver test. Note that I see lots of failures that predate my change.

### DIFF
--- a/packages/blockly/core/inputs/input.ts
+++ b/packages/blockly/core/inputs/input.ts
@@ -375,7 +375,13 @@ export class Input {
   getRowId(): string {
     const inputs = this.getSourceBlock().inputList;
 
-    // The first input in a block has the same ID as its parent block.
+    // The first visible input shares the block's row id; this also covers
+    // the collapsed-input placeholder, since every other input is hidden.
+    if (this === inputs.find((i) => i.isVisible())) {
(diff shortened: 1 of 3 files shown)


microbit-matt-hillsdon:
Back to draft as I think there's a real test failure here.

microbit-matt-hillsdon:
> Back to draft as I think there's a real test failure here.

Sorted the failing test and added a few more.

maribethb (approved):

PR #9962, opened 2026-06-09
fix: only show paste hint toast on keyboard copy/cut

microbit-matt-hillsdon:
This differs from the keyboard nav plugin but I think it's correct for an on-by-default world as it's very weird to encounter this toast on a touch device with and not helpful for mouse usage (e.g. in MakeCode's context menu).

## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9961

(newly raised, check the team agree)

### Proposed Changes

Gate the cut/copy toasts on the event being a KeyboardEvent (could be context menu via keyboard or the shortcut).


mikeharv (approved):

mikeharv:
Makes sense! Thanks for filing and fixing.

PR #9972, opened 2026-06-10
fix: position dragging block above/below all blocks

mikeharv:
## The basics

### Proposed Changes

When a block is moved to the workspace in constrained move mode, we will now move it above or below all blocks on the workspace.
image

### Reason for Changes

Previously, the fallback placement (above or below the block) was based on the bounding rectangle of a single “root” connected block. This could lead to the dragging block being placed on top of other blocks.

image

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -665,7 +665,7 @@ export class BlockDragStrategy implements IDragStrategy {
         if (!bounds) return;
 
         const blockRects = draggingBlock.workspace
-          .getTopBlocks(true)
+          .getTopBlocks()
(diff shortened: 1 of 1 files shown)


packages/blockly/core/dragging/block_drag_strategy.ts
```diff
@@ -664,25 +664,31 @@ export class BlockDragStrategy implements IDragStrategy {
         const bounds = root?.getBoundingRectangle();
         if (!bounds) return;
 
+        const blockRects = draggingBlock.workspace
+          .getTopBlocks(true)
```
maribethb: do you need this to be sorted? you're already finding the minimum left coordinate so i don't think it does.

maribethb (approved):

maribethb (approved):

PR #9977, opened 2026-06-11
fix: Do not bump neighbours mid-drag

mikeharv:
## The basics

### Resolves

Fixes 

### Proposed Changes

This makes me two adjustments to how blocks are positioned during `startDrag`, at the start of a keyboard move:
1. Dragging blocks will no longer be bumped or have their neighbours bumped. If a top block is moved off of its stack, the other blocks will stay put; similar to command+dragging a single block off the top of a stack.

image
image

2. Newly dragging blocks will use the same offset/positioning logic as blocks that are mid-drag. We offset the location so that it reveals both the insertion marker and candidate connecting block. If the dragging block is becoming a child, we offset to the lower right. If the dragging block is becoming a parent, we offset to the upper left.
imageimage

### Reason for Changes

As of the latest v13 beta.9, pressing M to move a block causes the selected block and any neighbour to move in unexpected ways due to bumping. The preview position of the dragging block was also inconsistent with later moves within the same "drag". 

image
image

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1615,6 +1615,7 @@ export class BlockSvg
     if (
       this.isDeadOrDying() ||
       this.workspace.isDragging() ||
+      this.isDragging() ||
       root.isInFlyout
     ) {
       return;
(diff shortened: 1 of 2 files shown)


maribethb (approved):
Thanks!

PR #9981, opened 2026-06-12
chore: Use default renderer for thrasos

mikeharv:
## The basics

### Proposed Changes

Updates the playground to use the default renderer, thrasos as of v13.

### Reason for Changes

With Blockly v13, Thrasos is the new default renderer. Explicitly setting the renderer would override the default.

### Additional Information

We should also update the default value of the renderer dropdown in the Advanced Playground's options list (in samples).

Geras:
image

Thrasos:
image


maribethb (approved):
LGTM but I would call this a chore because it does not affect the published release.

PR #9983, opened 2026-06-12
chore: Add ARIA-related messages for samples

gonfunko:
## The basics

### Proposed Changes
This PR adds a variety of localizable strings that will be used in ARIA-related attributes for various plugins in blockly-samples.


maribethb (approved):
can you make this against `main` instead, we've already merged the v13 branch.

PR #9984, opened 2026-06-12
docs: add clarification to mutators

### LINKED ISSUE #7981: Include Tutorial about MutatorIcon
### Location

https://developers.google.com/blockly/guides/create-custom-blocks/extensions

### Type

- [X] Text
- [ ] Image or Gif
- [ ] Other

### Suggested content

The guide [Extensions](https://developers.google.com/blockly/guides/create-custom-blocks/extensions) has never mentioned MutatorIcon. Thus, when users want to write a block like 'text_join' or 'list_create_with', they might find their block missing the left-most gear icon through which the mutator UI can be opened. The issue could be commonplace.
According to the guide, the decompose and compose functions provide the container block at the top of the mutator workspace. But the guide fails to tell readers that MutatorIcon provides information about the FlyOut of that workspace. The knowledge system is incomplete and we confuse how 'text_join_item' and 'list_create_with_item' are specified as choices in the FlyOut.
Hopefully a paragraph about MutatorIcon can be added to the guide [Extensions](https://developers.google.com/blockly/guides/create-custom-blocks/extensions). 

### Additional context

I sincerely appologize for my disturbance. Please forgive me if I am making mistakes as this is the first time for me to open an issue at blockly. I could not find a contribute button at the guide page, and I have no permission to post conversations at Google Groups. (I tried to post conversations, but after clicking Post Message, nothing happened and I could not find my conversation anywhere.) Sorry again if I am submitting an issue at the wrong place.

zspriggs:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/7981 

### Proposed Changes

Adds a section on mutators to the guide on [Extensions](https://docs.blockly.com/guides/create-custom-blocks/define/extensions/) which provides a brief overview of mutators and directs to the more in-depth [mutators doc](https://docs.blockly.com/guides/create-custom-blocks/mutators/). 

Also clarifies how the default UI works and makes a brief mention of the MutatorIcon class to provide clarification on when the user might need to access it directly. 

### Reason for Changes

Current docs are confusing on how to get the gear icon to be visible and how MutatorIcon class fits in.


packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
  "extensions": ["my_mixin"],
 }
 ```
+### Mutators
+
+Mutators are a type of mixin that allow you to add extra details about the state 
+of a block. If the full state of your block that can't be captured by the 
+block's existing properties, you may need a mutator. By default, Blockly
```
mikeharv, line 97: Maybe: "...add extra state for a block, especially state that cannot be represented using the block's standard fields or properties."

packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
...
+### Mutators
+
+Mutators are a type of mixin that allow you to add extra details about the state 
+of a block. If the full state of your block that can't be captured by the 
+block's existing properties, you may need a mutator. By default, Blockly 
+represents mutators with a small gear icon on the block. When the user 
+clicks the icon, a mutator UI and the user can select options that affect the 
+shape and/or behavior of the block.
+
+For instance, the builtin ```list_create_with``` block uses a mutator. The
```
mikeharv: nit: built-in

Also, can we use single ` instead of triple ```?

packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
...
+mutator allows for the user to change the shape of the block and allows the
```
mikeharv: I think I would say that "the mutator tracks this item count and updates the block's shape accordingly."

packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
...
+block to keep track of how many items are in the list. 
+
+![A list_create_with block with its mutator bubble open. This lets users add 
+more items to the list.](/images/list-create-with-mutator.png)
```
mikeharv: Since the purpose of the image is to highlight its parts, I think the alt text should describe that the icon/button and IU/bubble are identified in the image.

packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
...
+
+When you create a mutator, you'll need to define the certain functions that
```
mikeharv: nit: Could just be "When you create a mutator, you define functions that..."

packages/docs/docs/guides/create-custom-blocks/define/extensions.mdx
```diff
@@ -89,3 +89,27 @@ extension.
...
+specify how the mutator should serialize and deserialize the state of the block,
+as well as how the mutator should compose and decompose the block's shape in the
+UI. As with any mixin, you'll also need to register the mutator before you can 
+reference it in your block's JSON definition.
```
mikeharv, line 112: nit: I'd probably replace "your" with "a" here, since the same mutator can theoretically be used for any number of blocks.

packages/docs/docs/guides/create-custom-blocks/mutators.mdx
```diff
@@ -322,3 +326,13 @@ var helper = function () {
...
+
+### MutatorIcon class
+
+Blockly's default mutator UI is implemented in the `MutatorIcon` class. In most 
+cases, you can use the default mutator UI that Blockly provides or create your 
+own custom UI without instantiating `MutatorIcon` directly. If you are extending
+Blockly itself to add new mutator features, you might want to extend the 
+`MutatorIcon` class, but note that this will not always be needed. You can 
+look at the [MutatorIcon](https://docs.blockly.com/reference/blockly.icons_namespace.mutatoricon_class/#iconsmutatoricon-class)
+reference documentation for more details about this class.
```
mikeharv, line 338: I think it's worth mentioning that `MutatorIcon` manages the bubble UI when a mutator is defined for a block, and that custom mutator UIs can be implemented without instantiating or extending `MutatorIcon`. I think there isn't actually a consensus on whether people should use the default UI, so I'd avoid suggesting to use it "in most cases".

mikeharv (approved):

PR #9995, opened 2026-06-15
fix: nested category with blocks in parent 

### LINKED ISSUE #9392: Nested category with blocks in parent should collapse when dragging block
### Check for duplicates

### Problem

_No response_

### Request

Currently, when a parent toolbox category contains blocks alongside nested categories, reopening the parent category after dragging a block requires a double click, since the category is not collapsed automatically. This creates unnecessary friction.

**Expected behavior:**

- If the parent category contains blocks: collapse the parent after a block is dragged, so reopening the nested category only requires a single click.
- If the parent category contains no blocks: keep the current behavior.

More context in this [group discussion](https://groups.google.com/g/blockly/c/prcvorFdIuU/m/i0PGMgSRAQAJ).

### Reproduction Steps

1. Create sample app: `npx @blockly/create-package app hello-world`
2. Add a nested toolbox category with blocks in the parent category
3. Open the nested category
4. Drag a block onto the workspace
5. Try reopening the parent category (Requires double click)  

Example toolbox config:

```
    {
      kind: 'category',
      name: 'Nested',
      categorystyle: 'text_category',
      contents: [
        {
          kind: 'block',
          type: 'text',
        },
        {
          kind: 'block',
          type: 'text_join',
        },
        {
          kind: 'category',
          name: 'Nested Level 2',
          categorystyle: 'text_category',
          contents: [
            {
              kind: 'block',
              type: 'text_join',
            },
            {
              kind: 'block',
              type: 'text_indexOf',
            },
            {
              kind: 'block',
              type: 'text_getSubstring',
            },
            {
              kind: 'block',
              type: 'text_changeCase',
            }
          ]
        }
      ]
    }
  ],
};
```

https://github.com/user-attachments/assets/f3b92558-3c09-4722-b76d-fadbec7cdabe

zspriggs:
## The basics

### Resolves

Fixes #9392 

### Proposed Changes

Adjusts handling of toolboxes when a collapsible category has its own flyout items. Now, when a block is dragged out of an expanded parent, the categories all remain expanded even if the flyout automatically closes. When the user clicks back into the parent category, the parent flyout opens without collapsing the child category. Another click on the parent category will collapse the category and flyout. 

### Test Coverage

Includes 3 unit tests for covering different combinations of flyout/category open. Note that the category closed/flyout open combination was not included bc it shouldn't be possible to enter this state with the current setup. 

### Additional Information

Note that the updated behavior doesn't precisely match the expected behavior described in the original issue, this is an intentional choice made after discussing the issue!

### DIFF
--- a/packages/blockly/core/toolbox/collapsible_category.ts
+++ b/packages/blockly/core/toolbox/collapsible_category.ts
@@ -237,44 +237,29 @@ export class CollapsibleToolboxCategory
   }
 
   override onClick(_e: Event) {
-    //TODO: figure out what the behavior should be, then properly refactor
-    //Separate the logic for category expansion and the flyout
-    const parentToolbox = this.getParentToolbox();
-    const thisFlyoutVisible = parentToolbox.getSelectedItem() === this && parentToolbox.getFlyout()?.isVisible();
-    const isExpanded = this.isExpanded();
+    // Check for special case where the category has its own flyout items. 
+    if(this.flyoutItems_.length > 0) {
+      const parentToolbox = this.getParentToolbox();
+      const thisFlyoutVisible = 
+        parentToolbox.getSelectedItem() === this &&
+        parentToolbox.getFlyout()?.isVisible();
 
-    // if expanded && flyout showing, then close both.
-    if(isExpanded && thisFlyoutVisible) {
-      this.parentToolbox_.getFlyout()?.setVisible(false);
-      this.setExpanded(false);
-      return;
+      if(this.expanded_ && !thisFlyoutVisible) {
+        this.toggleFlyout();
+        return;
+      }
     }
 
-    // if expanded && flyout not showing, then just open flyout. 
-    if(isExpanded && !thisFlyoutVisible) {
-      this.parentToolbox_.getFlyout()?.setVisible(true);
-      return;
-    }
-
-    // if not expanded && flyout showing, then just expand OR close flyout??? 
-    // as it currently stands the code makes this one imposssible
-    if(!isExpanded && thisFlyoutVisible) {
-     //this.parentToolbox_.getFlyout()?.setVisible(false);
-      this.setExpanded(true);
-      return;
-    }
-
-    // if not expanded && flyout not showing, then open both
-    if(!isExpanded && !thisFlyoutVisible) {
-      this.setExpanded(true);
-      this.parentToolbox_.getFlyout()?.setVisible(true);
-      return;
-    }
-    //this.toggleExpanded();
+    this.toggleExpanded();
   }
 
+  /** Toggles the parent toolbox's flyout. */
   toggleFlyout() {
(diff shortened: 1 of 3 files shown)


packages/blockly/core/toolbox/collapsible_category.ts
```diff
@@ -237,9 +237,27 @@ export class CollapsibleToolboxCategory
...
+        this.setFlyout(true);
+        return;
+      }
+    }
+
     this.toggleExpanded();
   }
 
+  /** Sets the visibility of the parent toolbox's flyout. */
+  setFlyout(isVisible: boolean) {
```
gonfunko: This would be a bit clearer as `setFlyoutVisible()`, and should probably be `private`.

packages/blockly/tests/mocha/toolbox_test.js
```diff
@@ -261,6 +261,71 @@ suite('Toolbox', function () {
...
+        this.toolbox.render(collapsibleCategoryWithFlyout);
+        this.flyout = this.toolbox.getFlyout();
+        this.parentCategory = this.toolbox.getToolboxItems()[0];
+      });
+
+      test('if category collapsed and flyout hidden, click should uncollapse and open flyout', function () {
+        this.parentCategory.setExpanded(false);
+
+        this.parentCategory.onClick(new Event('click'));
+        this.toolbox.selectItemByPosition(0);
```
gonfunko: Does `this.parentCategory.getClickTarget().click()` work vs simulating the flow? If so that might be a bit nicer/less fragile.

packages/blockly/tests/mocha/toolbox_test.js
```diff
@@ -261,6 +261,71 @@ suite('Toolbox', function () {
...
+
+        assert.isTrue(this.parentCategory.isExpanded());
+        assert.isTrue(this.flyout.isVisible());
+      });
+      test('if category expanded and flyout hidden, click should open flyout', function () {
+        this.parentCategory.setExpanded(true);
+        this.flyout.getWorkspace().hideChaff();
```
gonfunko: Assuming the goal of this is to hide the flyout, I don't think it's doing that; the flyout's workspace is the workspace that displays the blocks in the flyout, and its `ComponentManager` doesn't know about the existence of the flyout. This should work if you call `hideChaff()` on the flyout's `targetWorkspace`, which is the main workspace, although it'd probably be a bit clearer to just do `this.flyout.hide()`.

packages/blockly/tests/mocha/toolbox_test.js
zspriggs: Unfortunately it doesn't work,`this.parentCategory.getClickTarget().click()` ends up having no effect on the state of the category or the flyout. If you have any other ideas let me know, I agree that it feels weird to simulate it like this

gonfunko (approved):

PR #9996, opened 2026-06-16
fix: Coalesce adjacent field labels

### LINKED ISSUE #9855: coalesce adjacent label fields

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9855

### Proposed Changes

Coalesces adjacent field labels, taking into account whether the labels are on the same visual row. 

### Reason for Changes

Blocks built in ways where multiple labels are next to one another were computed as being comma separated when building out the block's ARIA label. This is confusing for users as these labels are treated as a single string for sighted users while being read as separate strings for screenreader users.

### Test Coverage

Unit tests have been added to cover the edge cases around this change.

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -318,8 +318,84 @@ export function getInputLabels(
       );
     }
   }
+  let adjacentFieldLabels: Array<string> = [];
+  return inputsToLabel.flatMap((input, index, inputs) => {
+    const inputLabel = input.getLabel(verbosity);
+    if (
+      index < inputs.length - 1 &&
+      endsWithFieldLabel(input) &&
+      beginsWithFieldLabel(inputs[index + 1]) &&
+      // getVisualRowId(input) === getVisualRowId(inputs[index + 1])
+      !isEndOfRow(input, inputs[index + 1])
+    ) {
+      // This input ends with a field label and the next one begins with one.
+      // They are also on the same visual row. We want to combine these, so we
+      // add this one to the list for later handling. Value inputs are
+      // excluded from this as the value should be separated.
+      adjacentFieldLabels.push(inputLabel);
+      return [];
+    } else if (beginsWithFieldLabel(input) && adjacentFieldLabels.length > 0) {
+      // There is at least one adjacent FieldLabel before this one but none
+      // after. Combine the FieldLabels into one string.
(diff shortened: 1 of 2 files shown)


packages/blockly/core/block_aria_composer.ts
```diff
@@ -318,8 +318,83 @@ export function getInputLabels(
...
+      return label;
+    }
+    return inputLabel;
+  });
+}
+
+/**
+ * Returns whether an input's list of visible fields begins with a FieldLabel
+ *
+ * @internal
```
maribethb: You don't need to mark these as internal because they already aren't exported from this file

packages/blockly/tests/mocha/input_test.js
```diff
@@ -379,19 +379,80 @@ suite('Inputs', function () {
...
+            .appendField('third')
+            .appendField('fourth');
+
+          const label = this.block.getAriaLabel();
+
+          assert.include(label, 'first second,');
+          assert.include(label, 'third fourth');
+        });
+        test('Statement inputs are not combined', function () {
+          const statementBlock = Blockly.Xml.domToBlock(
```
maribethb: prefer using json for these kinds of tests (even if older existing tests use xml).

maribethb (approved):

PR #10032, opened 2026-06-18
fix: Bring blocks to the front when focused via keyboard

### LINKED ISSUE #9907: [13.0.0-beta.5] Bring focused node to the front when using keyboard controls
### Check for duplicates

### Description

This is not a regression from the keyboard navigation plugin.

When using a mouse, blocks that are selected with the mouse are brought to the front of the workspace. When navigating using keyboard controls, selected blocks are not brought to the front. This can result in some regrettable visuals (see screen grab below).

Is it possible to bring blocks / comments / bubbles etc to the front when they are selected via keyboard controls? It looks like workspace comments already have the desired behaviour.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/d38f4084-d80f-462e-badc-74023ea4d062

### Browsers

_No response_

microbit-robert:
## The basics

### Resolves

Brings blocks to the front when focused by keyboard navigation.

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9907

### Proposed Changes

Moves out a separate `moveSvgRootToFront` method from the existing `bringToFront` method and calls it inside `onNodeFocus` for blocks and rendered connections. The `moveSvgRootToFront` is safe to call inside the `onNodeFocus` callbacks.

### Reason for Changes

Aligns mouse and keyboard behaviour. Avoids situations when focused blocks are visually hidden by other blocks on top of them when using keyboard navigation.

### Test Coverage

Not covered.

### Documentation

N/A

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1385,10 +1385,11 @@ export class BlockSvg
     do {
       const root = block.getSvgRoot();
       const parent = root.parentNode;
-      const childNodes = parent!.childNodes;
+      if (!parent) return;
+      const childNodes = parent.childNodes;
       // Avoid moving the block if it's already at the bottom.
       if (childNodes[childNodes.length - 1] !== root) {
-        parent!.appendChild(root);
+        parent.appendChild(root);
       }
       if (blockOnly) break;
       block = block.getParent();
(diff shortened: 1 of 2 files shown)


mikeharv (approved):
Thanks for the fix!

gonfunko:
@microbit-robert This appears to break VoiceOver + Safari when navigating between blocks; generally the newly-focused block doesn't actually get read out.

PR #10033, opened 2026-06-18
fix: Keep workspace stack count fresh instead of hiding it on Mac/VoiceOver

microbit-matt-hillsdon:
Screen readers read the workspace's region label when its contents are focused, but updateAriaLabel only ran on the workspace node's own focus. When a child block took focus on first entry the label was never refreshed, so a stale stack count (e.g. "0 stacks") was announced. The previous workaround dropped the count entirely on Apple.

Refresh the label in onTreeFocus, which fires whenever focus enters the workspace tree (including via a child), before the element is focused, so the announced count is current. This lets us restore the informative stack count on all platforms.

## The basics

### Resolves

This fixes only part 1 of this issue as described in this comment https://github.com/RaspberryPiFoundation/blockly/issues/9885#issuecomment-4719643883

### Proposed Changes

Refresh the label in onTreeFocus, which fires whenever focus enters the workspace tree (including via a child), before the element is focused, so the announced count is current. This lets us restore the informative stack count on all platforms.

### Reason for Changes

This isn't Mac specific, other screen readers also output region context when focusing  nodes in the region.

### DIFF
--- a/packages/blockly/core/workspace_svg.ts
+++ b/packages/blockly/core/workspace_svg.ts
@@ -745,17 +745,6 @@ export class WorkspaceSvg
    * Updates the label on the workspace to reflect the number of top-level stacks in the workspace.
    */
   private updateAriaLabel() {
-    if (userAgent.APPLE) {
-      // VoiceOver is reading this label inappropriately, so don't show the
-      // stack count because it might be inaccurate.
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
This orphans WORKSPACE_LABEL_PLAIN but it seems prudent to keep it around.

maribethb (approved):

PR #10037, opened 2026-06-18
feat: Add relevant things to full-block field labels

### LINKED ISSUE #9864: add "begin stack" label and other relevant things to full-block field labels
### Check for duplicates

### Description

blocks that are full-block field blocks get special cased in block aria composer, the special case should include adding things like "begin stack", "disabled", the custom input name, etc if they are relevant to the block

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9864

### Proposed Changes

This updates how ARIA labels are computed for field classes that support full-block fields (`FieldInput` and `FieldDropdown`). When `recomputeAriaContext()` is called for these fields, we were update the `aria-label` using the full block's label, which includes a lot of additional information. This requires re-inserting the expected field label (with type name) to the right place with the block label.

Examples:

Focusing full-block field, number top block
image
Focusing full-block field, child
image
Information shortcut, shadow
image
Focusing full-block field, dropdown top block
image
Information shortcut
image
Focusing full-block field
image
Information shortcut, disabled
image
Extended Information shortcut
image

Labels are also refreshed automatically when block context changes: `BlockSvg` delegates aria recomputation to the full-block field on connect/disconnect (`setParent`) and after drag ends (`setDragging(false)`), with tests for those paths.

### Reason for Changes

Blocks that are full-block field blocks get special cased in block aria composer, but the special case didn't include adding things like "begin stack", "disabled", the custom input name, etc if they are relevant to the block.

### Test Coverage

Tests were added that check expected labels for focused blocks similar to those screenshotted above.


packages/blockly/core/field_dropdown.ts
```diff
@@ -940,7 +941,14 @@ export class FieldDropdown extends Field<string> {
     if (!shouldCustomize) return false;
 
     const focusableElement = this.getFocusableElement();
-    const label = this.computeAriaLabel(true);
+    let label = this.computeAriaLabel(true);
+    if (this.isFullBlockField()) {
+      // Full block fields get a more detailed label that includes the block's label
+      label = computeAriaLabel(this.getSourceBlock() as BlockSvg).replace(
+        this.computeAriaLabel(false),
+        label,
```
mikeharv: This admittedly feels a bit janky, but it's needed because `getInputLabels` will use `includeTypeInfo = false` for standard verbosity.

packages/blockly/core/field_input.ts
```diff
@@ -855,8 +856,35 @@ export abstract class FieldInput<T extends InputTypes> extends Field<
...
+      const fullBlockLabel = computeAriaLabel(
+        this.getSourceBlock() as BlockSvg,
+      ).replace(this.computeAriaLabel(false), label);
+      if (requiresEditableLabel) {
+        const labels = fullBlockLabel.split(', ');
+        const beginStackLabel = getBeginStackLabel(
+          this.getSourceBlock() as BlockSvg,
+        );
+
+        // Insert "Edit" after "Begin stack" if found, otherwise at start.
```
mikeharv, line 880: This is vibes based, but it felt odd to read/hear "Edit Begin stack, number:10..."

microbit-robert:
I've done some initial testing with these changes in MakeCode. When I refresh the page, I see a lot of full block numbers as block inputs that get the "begin stack" prefix when I think they shouldn't. If I delete the block, then hit undo, the "begin stack" prefix is removed and the input name is added. I'm not sure if there's a timing issue here - perhaps the full block field labels are calculated on load before they are attached to block they are in?

https://github.com/user-attachments/assets/a460b996-86c5-4671-b68d-219200978eea

microbit-robert:
This is a really daft example, but it's more likely to happen in MakeCode where we have full block text inputs. If the value of the input matches the input name, then weird things happen to the output. The order of elements is switched and repeated? I don't really know what's happening here.

https://github.com/user-attachments/assets/9063adf0-f7d4-4739-b73f-fc974fba6f00

microbit-robert:
Is it worth adding input names to fields as well? Nice to know what P0 and P1 relate to when I focus them for this block for example?

Screenshot 2026-06-22 at 13 05 01

microbit-robert:
I ran into another text mangling example and it does seem to depend on the value of the input. I was able to reproduce in the Blockly playground running locally. Some kind of replacement problem maybe? I haven't investigated.

Update: Ahh [it's this code](https://github.com/mikeharv/blockly/blob/full-block-fields/packages/blockly/core/field_input.ts#L868-L870). For this example it amounts to `"value 1, a".replace("a", "number: a")` - it replaces the first "a" in "value".

I've raised https://github.com/RaspberryPiFoundation/blockly/issues/10050 separately.

https://github.com/user-attachments/assets/32fa588a-1c84-4d8c-b4aa-8c325a2bab98

mikeharv:
Hi @microbit-robert. Thanks for the thorough testing here! I've updated this branch with fixes that should remove the timing problems and text mangling. Please feel free to take another look!

microbit-robert:
> Hi @microbit-robert. Thanks for the thorough testing here! I've updated this branch with fixes that should remove the timing problems and text mangling. Please feel free to take another look!

@mikeharv Looks good to me now in terms of behaviour. Thanks!

maribethb (approved):

PR #10061, opened 2026-06-24
feat: ctrl/cmd+v pastes near focused block

### LINKED ISSUE #9091: Pasting with ctrl-v should paste near the currently selected block
### Check for duplicates

### Problem

Pasting with ctrl-v tries to place the block in the same position as the originally copied block, then progressively bumps the pasted block down and to the right until it reaches an unambiguous position, meaning that it doesn't overlap any other block exactly and is not in the snap radius of any other connections.

If the user copied in one location with the keyboard and then moved to a different location to paste, the paste should occur near their cursor instead of near the copied block.

### Request

Use the location of the focused block or connection as the start point for the placement algorithm in `block_paster.ts`.

### Alternatives considered

Per discussion in https://github.com/google/blockly-keyboard-experimentation/issues/480, an alternative is to connect to the selected location on paste. See the other issue for a discussion of why we are not doing that, and the user-facing issues it would create.

### Additional context

Sibling of https://github.com/google/blockly/issues/9088
Remaining work needed for https://github.com/google/blockly-keyboard-experimentation/issues/480

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/9091

### Proposed Changes

This updates the `'paste'` shortcut to use a currently focused block's position instead of the position included in the `copyData`. If a block's child is focused (e.g., connection, comment editor, etc.), the source block's position is used. If the workspace is focused, there is no change to the behavior.

### Reason for Changes

Previously, if the user copied in one location with the keyboard and then moved to a different block, the paste would occur near the original copied block's location.

### Test Coverage

Added new relevant tests.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -297,6 +297,27 @@ export function registerCut() {
 }
 
 /**
+ * Returns workspace coordinates to use as the paste origin when focus is on a
+ * block or one of its children (fields, connections, icons, block comments).
+ *
+ * @param focusedNode The node that currently has focus.
+ * @param targetWorkspace The workspace where the paste will occur.
+ * @returns The parent block's location in workspace coordinates, or undefined
+ *     if focus is not on a block or one of its children.
+ */
+function getPasteOriginFromFocus(
+  focusedNode: IFocusableNode | null,
+  targetWorkspace: WorkspaceSvg,
+): Coordinate | undefined {
+  const block = targetWorkspace
+    .getNavigator()
+    .getSourceBlockFromNode(focusedNode);
+  if (!block) return undefined;
+
+  return block.getRelativeToSurfaceXY();
+}
+
+/**
  * Keyboard shortcut to paste a block on ctrl+v, cmd+v, or alt+v.
  */
 export function registerPaste() {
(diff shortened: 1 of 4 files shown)


packages/blockly/core/shortcut_items.ts
```diff
@@ -355,6 +356,16 @@ export function registerPaste() {
         return !!clipboard.paste(copyData, targetWorkspace, mouseCoords);
       }
 
+      // If the focused node is a block, paste relative to that block's position.
```
maribethb: I think this comment is very slightly inaccurate, it's if the focus node is a block or part of a block (e.g. connection or field)

packages/blockly/core/shortcut_items.ts
```diff
@@ -355,6 +356,16 @@ export function registerPaste() {
...
+      const block = targetWorkspace
+        .getNavigator()
+        .getSourceBlockFromNode(focusedNode);
+      const pasteOrigin = block?.getRelativeToSurfaceXY();
+      if (pasteOrigin) {
+        return !!clipboard.paste(copyData, targetWorkspace, pasteOrigin);
+      }
```
maribethb, line 367: it would be super cool if when:

1. a connection is focused
2. the copied block is compatible with that connection

we connected the block to the connection automatically

but that could be a follow up feature too :)

maribethb (approved):

packages/blockly/core/shortcut_items.ts
mikeharv: I think that was discussed in https://github.com/RaspberryPiFoundation/blockly-keyboard-experimentation/issues/480. There were concerns that auto-connecting on paste could have surprising effects (popping out blocks, unexpected nesting, etc.), and that people often use copy/paste to reuse a block elsewhere rather than insert it at the current location.

PR #10067, opened 2026-06-25
chore: use none ARIA role consistently instead of presentation

### LINKED ISSUE #9993: consistently use role of none instead of role of presentation
### Check for duplicates

### Description

we use role of presentation and role of none interchangeably. they are synonyms, but it would be less confusing if we consistently used one or the other. we've decided "none" is the more aptly named one and should update usages of "presentation" to "none"

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

dejoski:
## Summary
- add `aria.Role.NONE` for the ARIA `none` role
- replace internal uses of the presentation role with the none role
- update role assertions in the affected mocha tests

Closes #9993

## Validation
- npm ci
- npm run lint --workspace blockly
- npm run tsc --workspace blockly
- git diff --check

### DIFF
--- a/packages/blockly/core/block_aria_composer.ts
+++ b/packages/blockly/core/block_aria_composer.ts
@@ -87,7 +87,7 @@ export function computeAriaLabel(
  * @param block The block to set ARIA role and roledescription attributes on.
  */
 export function configureAriaRole(block: BlockSvg) {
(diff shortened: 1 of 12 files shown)


mikeharv (approved):
Thanks for cleaning this up for us!

PR #10073, opened 2026-06-26
fix: Improve VoiceOver experience when using Safari

### LINKED ISSUE #10066: Issues with VoiceOver and Safari
### Check for duplicates

### Description

https://github.com/RaspberryPiFoundation/blockly/pull/10032 introduced an [issue](https://github.com/RaspberryPiFoundation/blockly/pull/10032#issuecomment-4793362803) where where VoiceOver doesn't output anything when moving focus from block to block. I was going to suggest simply reverting this change. However, when testing in Safari with [commit f466d4](https://github.com/RaspberryPiFoundation/blockly/commit/f466d4daa74c421aae5e9174fc3feb59055a5fb6) reverted, I also noticed that rendered connections don't play nicely with VoiceOver and Safari. This might due to the rendered connection highlight being `display: none` until it is focused. These are flakey at best in Safari with VoiceOver and ignored completely at worst. See screen grab below.

If Safari needs to be supported, both of these issues need addressing.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

Issue with rendered connections in Safari remains with [commit f466d4](https://github.com/RaspberryPiFoundation/blockly/commit/f466d4daa74c421aae5e9174fc3feb59055a5fb6)  reverted:
https://github.com/user-attachments/assets/dff3834f-8e00-41d7-97de-c77f354c63d7

### Browsers

_No response_

microbit-robert:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10066

### Proposed Changes

This PR consists of many small changes that combine to improve the VoiceOver in Safari experience:

- Do not append the element to the DOM that is about to be focused (causes VoiceOver cursor / focus decoupling). Instead move other elements before it. This is applied to blocks and rendered connections
- Reorder blocks / connections inside requestAnimationFrame to avoid VoiceOver cursor flickers
- Compute basic aria role and roledescription for rendered connections at draw time. This ensures that the element is focusable and understood by Safari to be focusable. The aria-label is still only calculated on focus.
- Ensure that rendered connections are always visible. Toggling the display style causes real issues with Safari and VoiceOver. Instead, a class is used to manage the stroke highlighting and toggled where display was previously.
- Ensures that the move mode icon stays at the front after the above changes

### Reason for Changes

Maintains the '[bring to the front'](https://github.com/RaspberryPiFoundation/blockly/pull/10032) fix for blocks focused via keyboard and improves VoiceOver in Safari. 

### Test Coverage

Manual testing in Safari with VoiceOver. Existing tests still pass.

### Documentation

None

### Additional Information

This change would really benefit from testing with different screen readers on different browsers to ensure that are no regressions.

### DIFF
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1397,6 +1397,7 @@ export class BlockSvg
       return;
     }
     requestAnimationFrame(() => {
+      if (this.dragging) return;
       /* eslint-disable-next-line @typescript-eslint/no-this-alias */
       let block: this | null = this;
       do {
--- a/packages/blockly/core/block_svg.ts
+++ b/packages/blockly/core/block_svg.ts
@@ -1393,25 +1393,28 @@ export class BlockSvg
    * @internal
    */
   moveSvgRootToFront(blockOnly = false) {
-    /* eslint-disable-next-line @typescript-eslint/no-this-alias */
-    let block: this | null = this;
-    if (block.isDeadOrDying()) {
+    if (this.isDeadOrDying()) {
       return;
     }
-    do {
-      const root = block.getSvgRoot();
-      const parent = root.parentNode;
-      if (!parent) return;
-      const childNodes = parent.childNodes;
-      // Avoid moving the block if it's already at the bottom.
-      if (childNodes[childNodes.length - 1] !== root) {
-        while (root.nextSibling) {
-          parent.insertBefore(root.nextSibling, root);
+    requestAnimationFrame(() => {
+      /* eslint-disable-next-line @typescript-eslint/no-this-alias */
+      let block: this | null = this;
+      do {
+        if (block.isDeadOrDying()) return;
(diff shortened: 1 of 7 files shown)


gonfunko (approved):
Thanks for this! I kicked the tires in Chromevox and NVDA and it seems OK there as well. I did notice Voiceover is now reading an incorrect count for the number of blocks in the flyout, but on net this is still a big improvement.

microbit-robert:
@gonfunko Good catch. I suspect this is because the rendered connections are present in the flyout and get included in the count now as they have roles. Probably fixable by having the role setting conditional on not being in a flyout, or ensuring that the block parent aria-owns its rendered connections (similar to icons). I won't have time to look at this until Monday, but happy for someone else to pick it up in the meantime.

gonfunko:
I'll merge this as-is for now and take a look at that myself.

PR #10074, opened 2026-06-26
docs: refactor getting-started codelab to be based on sample app

zspriggs:
## The basics

### Proposed Changes

- Refactors the getting-started codelab to be based on the existing sample app rather than the custom music-maker app.
- The content has remained the same where possible, with slightly more elaboration throughout on what the copy & paste code is actually doing. 
- The codelab goes over largely the same topics, but focuses more on important Blockly concepts than configuring a specific page. 
- Adds the sample app README to the getting started docs, so that the existence of the sample app is more visible and the overview can be referenced from the codelabs.


gonfunko:
This also fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/1752

packages/docs/docs/codelabs/getting-started/complete-code/src/blocks/text.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
```
maribethb: Here and elsewhere, please update copyright to 2026 and Raspberry Pi Foundation

(old files stay as they are, but new files get new info)

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
```diff
@@ -66,42 +76,56 @@ To do this you call the function `Blockly.inject(container, options)`, which tak
...
+`Blockly.inject` allows you to specify some options regarding what your workspace and toolbox look like. 
+In this case, the only configuration option you selected was `toolbox` which lets you add a toolbox definition to your instance of Blockly. 
 
-- `toolbox`: A JavaScript object which defines the toolbox for the editor.
+The `options` struct offers many additional options which give you significant control over your Blockly's instance.
+
+For example: 
 - `scrollbars`: Whether to show scrollbars in the workspace.
 - `horizontalLayout`: Whether to display the toolbox horizontally or vertically in the workspace.
 - `toolboxPosition`: Whether to show the toolbox at the top or bottom of the workspace.
```
maribethb: it's only top or bottom if we were using the horizontal layout. in this case, we're not, so this is kind of confusing. you also might want to explain the default values for these since it's not obvious horizontalLayout will be false, scrollbars will be true, etc.

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
```diff
@@ -66,42 +76,56 @@ To do this you call the function `Blockly.inject(container, options)`, which tak
...
+- `maxBlocks`: The maximum number of blocks that may be created in the workspace. 
 
-The `options` struct gives you significant control over your Blockly instance. You can pass options to set Blockly's theme, modify scrolling behaviour, set the renderer, and more. For more information, head over to Blockly's developer site and check out the [configuration](/guides/configure/configuration_struct#the-options-dictionary) section.
+You can also pass options to set Blockly's theme, modify scrolling behaviour, set the renderer, and more. For more information, see the [configuration](/guides/configure/configuration_struct#the-options-dictionary) documentation.
```
maribethb: For people new to blockly, the theme and renderer probably don't mean anything. I would either link to the documentation for those topics, or make this more general and say you can control the appearance of blockly.

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
```diff
@@ -66,42 +76,56 @@ To do this you call the function `Blockly.inject(container, options)`, which tak
...
 <Image
-  src="/images/codelabs/getting-started/workspace_with_loop.png"
-  alt="image"
+  src="/images/codelabs/getting-started/full_page_first_toolbox.png"
+  alt="A webpage with a space for generated code, a Blockly toolbox with two blocks in it, and a Blockly workspace."
   className="codelabImage"
-/>
-
-Drag and drop the available loop block to the workspace to test it out.
+/>
```
maribethb, line 132: I would add some instructions to drag the blocks around so they can see it already works with basic interactions.

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
```diff
@@ -8,73 +8,83 @@ import Image from '@site/src/components/Image';
...
 ## 6. Create a custom block
 
-Since this is a music maker app, we want a block that plays sounds. We could create one block per sound, but instead we will create a single block with a dropdown to select which note to play:
-
-<Image
-  src="/images/codelabs/getting-started/play_sound_block.png"
-  alt="image"
-  className="codelabImage"
-/>
+Now, you'll add a custom block to the workspace.
```
maribethb: I think I would add a sentence explaining why we're using custom blocks. Something like, Blockly ships with several basic blocks for writing numbers, strings, and loops. But the real power of a Blockly application comes from your custom blocks. You can define blocks that are related to your application, such as one that plays a song, makes a robot move on the screen, or reports the temperature from a connected sensor.

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
```diff
@@ -8,73 +8,83 @@ import Image from '@site/src/components/Image';
...
-<Image
-  src="/images/codelabs/getting-started/play_sound_block.png"
-  alt="image"
-  className="codelabImage"
-/>
+Now, you'll add a custom block to the workspace. 
 
 In Blockly, a _block definition_ describes how a block looks and behaves. This includes its text, colour, and shape. It may also include which other blocks it can connect to.
 
 Blocks can be defined in either JavaScript or JSON. The developer site has a full article on [how to define a block](/guides/create-custom-blocks/define/block-definitions#how-to-create-block-definitions).
```
maribethb, line 16: It might be nice to link to the block factory here as well as a really visual way of showing how a block can be customized. edit: i see that's linked at the end of this page, will leave it up to you if that's fine or maybe better to link to it earlier.

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
```diff
@@ -8,73 +8,83 @@ import Image from '@site/src/components/Image';
...
+```
+
+This block definition defines a block called `add_text` that has one `string` value input called `TEXT`. There is a message on the block that says "Add text" before the input. 
+The `add_text` block also has a previous connection and a next connection, allowing it to be stacked vertically. 
+
+The page on [Block structure in JSON](/guides/create-custom-blocks/define/structure-json) is a good resource for understanding the parts of a JSON block definition. 
+
+### Create Block Definition
+
+After defining the JSON definition for the block, we'll need to create the block definition.
```
maribethb: This is kind of a confusing sentence, possibly because it uses the word "define/definition" too much haha. Maybe:

Now that we've written the JSON definition for the block, we'll need to turn that into the full block definition that Blockly can register later. To do that, we call `createBlockDefinitionsFromJsonArray` which takes the raw JSON object we wrote and turns it into the JavaScript block definition Blockly will use. We save that definition as an object called `blocks` which we will use in the next step.

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
```diff
@@ -8,73 +8,83 @@ import Image from '@site/src/components/Image';
...
+The block is set up, now you need to register it. Add the following import statement and code to the top of the `index.js` file:
 
-```html
-<script src="scripts/sound_blocks.js"></script>
+```js
+import {blocks} from './blocks/text';
 ```
 
-Your sound block definitions must come after importing Blockly and before the other imports, since you will use Blockly functions in this file, and you will be using functions from this file in later files. Your imports should now look like this:
+Just after the import statements, insert the following code in `index.js`, which adds your custom block to the list of available blocks:
```
maribethb: Be slightly more explicit and say uses the block definition we saved in the previous step to add the custom block to the list of available blocks.

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
```diff
@@ -8,73 +8,83 @@ import Image from '@site/src/components/Image';
...
+// This file has no side effects!
+export const blocks = Blockly.common.createBlockDefinitionsFromJsonArray([
+  addText,
 ]);
 ```
 
-Add a script tag to `index.html` to include your new block definition:
+### Register the custom block
+
+The block is set up, now you need to register it. Add the following import statement and code to the top of the `index.js` file:
```
maribethb, line 73: nit: this codelab is inconsistent with using you vs we. I don't think this is the biggest deal but it may be worth looking through the existing codelabs and seeing if we have a preference for one over the other (then maybe documenting that standard somewhere for future codelab writers)

packages/docs/docs/codelabs/getting-started/create-base-app.mdx
```diff
@@ -0,0 +1,39 @@
+---
+description: How to set up the base for your app.
+---
+
+# Getting started with Blockly
```
maribethb, line 5: I think it might be better to explain how to import Blockly in this step instead of at the inject a workspace stage. then you have some dedicated space to explaining the import statements and you can also explain the difference between `import * as Blockly from 'blockly';` and `import * as Blockly from 'blockly/core';` because different files in the app use different versions and previously that wasn't explained.

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
```diff
@@ -66,42 +76,56 @@ To do this you call the function `Blockly.inject(container, options)`, which tak
...
 
+Copy and paste the full code into your `index.js`: 
 ```js
-Blockly.inject('blocklyDiv', {
-  toolbox: toolbox,
-  scrollbars: false,
-  horizontalLayout: true,
-  toolboxPosition: 'end',
-});
+import * as Blockly from 'blockly';
```
maribethb, line 93: see note further down but I think this would be best to explain/set up in step 4.

packages/docs/docs/codelabs/getting-started/create-base-app.mdx
```diff
@@ -0,0 +1,39 @@
...
+    </div>
+  </body>
+</html>
+```
+
+### Examine the CSS
+
+The css for the app is already included in the `index.css` file. This codelab includes basic styling for the webpage. 
+
+If you would like to change the theme of Blockly, you can check out the [customizing your themes codelab](/codelabs/theme-extension-identifier/codelab-overview/).
```
maribethb: would briefly explain what a theme is and consider linking to the theme blocks in addition to codelab

packages/docs/docs/codelabs/getting-started/generate-javaScript-code.mdx
```diff
@@ -8,43 +8,74 @@ import Image from '@site/src/components/Image';
 
 ## 8. Generate JavaScript code
 
-Now that each button can be configured with its own Blockly workspace, the next thing we want to do is to generate JavaScript code from each workspace.
+The next thing we want to do is to generate JavaScript code.
```
maribethb: Here is another place I would recommend tying back into the "why" of the app -- that blocks on their own don't *do* anything, but we can make them generate code that does, and then run that code.

packages/docs/docs/codelabs/getting-started/generate-javaScript-code.mdx
```diff
@@ -8,43 +8,74 @@ import Image from '@site/src/components/Image';
 
 ## 8. Generate JavaScript code
 
-Now that each button can be configured with its own Blockly workspace, the next thing we want to do is to generate JavaScript code from each workspace.
+The next thing we want to do is to generate JavaScript code. 
 
 This generated code will be run by the browser, effectively executing the blocks set up in the Blockly workspace.
```
maribethb: I very much dislike this sentence (I know it's not yours) because the idea of "executing" blocks is a very common misconception that is hard to dispel for whatever reason. We should be clearer here that blocks are turned into strings, and if those strings are code, they can be executed (and executing the code is outside the scope of what Blockly cares about, and is something the application has to care about.)

Consider stealing some of the language from here: https://docs.blockly.com/guides/create-custom-blocks/code-generation/overview/ or linking to that page which I think explains the process much better

packages/docs/docs/codelabs/getting-started/generate-javaScript-code.mdx
```diff
@@ -8,43 +8,74 @@ import Image from '@site/src/components/Image';
...
-  return 'MusicMaker.queueSound(' + value + ');\n';
+import {Order} from 'blockly/javascript';
+
+export const forBlock = Object.create(null);
+
+forBlock['add_text'] = function (block, generator) {
+  // Generate the code for the input block.
+  const text = generator.valueToCode(block, 'TEXT', Order.NONE) || "''";
+  // Create the function for this block.
+  const addText = generator.provideFunction_(
```
maribethb, line 50: This block uses an advanced topic for code generators (the provideFunction call) and it may be worth calling that out or pointing people to the codelab for generators.

packages/docs/docs/codelabs/getting-started/generate-javaScript-code.mdx
```diff
@@ -8,43 +8,74 @@ import Image from '@site/src/components/Image';
...
-As previously mentioned, you can define your imports more carefully to get a [different generator](https://www.npmjs.com/package/blockly#blockly-generators).
+By default, Blockly can generate code from blocks for five languages: Dart, JavaScript, Lua, PHP, and Python. 
+You can also create a custom language generator to generate code in a different language, which the [custom generator codelab](/codelabs/custom-generator/codelab-overview/) describes. 
 
 ### Add a block generator
 
 When Blockly generates JavaScript code for blocks in a workspace, it translates each block into code. By default, it knows how to translate all library-provided default blocks into JavaScript code. However, for any custom blocks, we need to specify our own translation functions. These are called _block generators_.
 
-Add the following code to the bottom of `scripts/sound_blocks.js`:
+Find the `javascript.js` file in `src/generators/`. This is where you will tell Blockly's JavaScript generator how to generate code for the add text block.
```
maribethb: nit: this step creates the block generator function, but it doesn't register it with Blockly's JavaScript generator yet. I think you should reference that step more clearly below, so I would change this one to read something like:

Find the `javascript.js` file in `src/generators/`. This is where you'll write the block generator function for the `add_text` block.

packages/docs/docs/codelabs/getting-started/generate-javaScript-code.mdx
```diff
@@ -8,43 +8,74 @@ import Image from '@site/src/components/Image';
...
+function addText(text) {
+  // Add text to the output area.
+  const outputDiv = document.getElementById('output');
+  const textEl = document.createElement('p');
+  textEl.innerText = text;
+  outputDiv.appendChild(textEl);
+}
+
+addText('Hello, World!');
 ```
```
maribethb, line 88: Add a step here to register the block code generator with the javascript generator. be explicit that's what the object.assign is doing.

packages/docs/docs/codelabs/getting-started/run-generated-code.mdx
```diff
@@ -8,52 +8,100 @@ import Image from '@site/src/components/Image';
...
-try {
-  eval(code);
-} catch (error) {
-  console.log(error);
-}
+Object.assign(javascriptGenerator.forBlock, forBlock);
 ```
 
-### The full code
+### Write the function
```
maribethb: Set up code execution environment, or something more descriptive because it's not clear what "the function" is

packages/docs/docs/codelabs/getting-started/run-generated-code.mdx
```diff
@@ -8,52 +8,100 @@ import Image from '@site/src/components/Image';
...
 
-The end result should look like this:
+The `runCode` function uses Blockly's javascriptGenerator to generate the code for the blocks on the workspace. 
+Then it displays the code in the codeDiv, resets the outputDiv, and evaluates the JavaScript code.
```
maribethb, line 19: Make it clear that executing the code is the responsibility of the application. Blockly just turns blocks into code. What you do with that code is up to you. For this sample app, we'll display the code in a div, and we'll use `eval` to run it.

packages/docs/docs/codelabs/getting-started/save-load-workspace.mdx
```diff
@@ -6,48 +6,101 @@ description: How to save and load your workspace.
 
 ## 7. Save/load workspace
 
-You now have a Blockly workspace that appears when editing a button. The user can use this to write code, but there's no link between the code the user has written and an individual button on the page.
-
-Once the button behavior is defined by the user, it needs to be saved for later use. The saved code must be per button, since buttons can be programmed to play different sounds.
+You now have a Blockly workspace. However, if you reload the page, you'll find that any blocks you've dragged into the workspace will disappear.
+Your users likely won't want their blocks to vanish, but you can fix this by saving the state of the workspace in local browser storage.
```
maribethb: nit: saving to local storage is just one way to handle storage and probably not the one most real apps would use. 

Your users likely won't want their blocks to vanish, but you can fix this by saving the state of the workspace and loading it back in next time they visit the site. For this codelab, we'll save that state to local browser storage.

.gitignore
```diff
@@ -9,3 +9,4 @@ build-debug.log
 *.komodoproject
 /nbproject/private/
 tsdoc-metadata.json
+.vscode
```
maribethb: nit: add the trailing newline

maribethb (commented):
Sorry it's a lot of comments but overall the structure is great and I've mostly suggested some wording changes or minor reorganization!

packages/docs/docs/codelabs/getting-started/complete-code/src/blocks/text.js
zspriggs: I wasn't sure whether to keep the old license since I just copied the exact same code from the sample app. It's a new file wrt this codelab, but not a new file for Blockly. Happy to update but just want to double check that's what I should do (since the same will be the case for all the other codelabs)

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
zspriggs: Done, I also realized it might be helpful to give a shout to keyboard nav here so I added that as well

packages/docs/docs/codelabs/getting-started/create-a-custom-block.mdx
zspriggs: I left it at the bottom but added another reference to it at the end so they don't forget it exists

packages/docs/docs/codelabs/getting-started/create-a-blockly-workspace.mdx
zspriggs: Yes good point, I fixed this! Also, taking a closer look at this this made me realize that some of the default options are different between flyout & category toolboxes. With the goal being consistency across codelabs, I think that not clarifying this could cause some confusion when someone moves on to the real sample app after this codelab and sees a trash can and scrollbars when those things weren't in the workspace in this codelab. There's a few ways I could address this but I ended up just adding a couple lines at the end of the codelab about the difference, lmk if that looks good

packages/docs/docs/codelabs/getting-started/complete-code/src/blocks/text.js
maribethb: yeah I think I would still update it.

packages/docs/docs/codelabs/getting-started/create-base-app.mdx
```diff
@@ -35,5 +35,28 @@ Copy and paste the following HTML into your `index.html` file.
...
+``` 
+
+Importing Blockly this way includes the core Blockly library and the builtin blocks. 
+However, sometimes you may only want to import the core Blockly library. In that case, you could import Blockly using:
+
+```js
+import * as Blockly from 'blockly/core';
+```
+
+This codelab will go over generators in the [generate JavaScript code](/codelabs/getting-started/generate-javascript-code) step, but for now just note that generators will need to be imported separately, like so:
```
maribethb: It might be easier to just say "We'll learn how to import generators in step ##" and leave the actual import statement for then.

But I'm also not sure if they know what a generator is at this point, so maybe best to leave it off entirely?

packages/docs/docs/codelabs/getting-started/create-base-app.mdx
```diff
@@ -35,5 +35,28 @@ Copy and paste the following HTML into your `index.html` file.
...
+
+At various points throughout this codelab, you'll have to import Blockly in different ways. 
+Since you already installed Blockly using npm, you can freely import Blockly into your files without an additional step. 
+Add the following import statement to `index.js`:
+
+```js
+import * as Blockly from 'blockly';
+``` 
+
+Importing Blockly this way includes the core Blockly library and the builtin blocks.
```
maribethb: what about:

Importing Blockly this way includes the core Blockly library, the built-in blocks, and some other pieces you need but we won't discuss yet, like the JavaScript [code generator](link to step whatever) and the English [translations](link to article)

then you could delete the part below about importing the generator

maribethb (approved):

PR #10078, opened 2026-06-26
fix: detect delete areas using pointer coordinates

### LINKED ISSUE #10076: Bad delete area detection when zoomed
### Check for duplicates

### Description

When zooming in on the workspace, blocks that are not touching a delete area, but are near one, can be deleted. This is most evident in mutator workspaces due to their size.

### Reproduction steps

1. Zoom in a few steps on the Blockly Playground workspace.
2. Add a `controls_if` block to the workspace and open its mutator.
3. Drag a `controls_if_elseif` block from the mutator flyout to the stack.

An insertion marker is shown as well as delete styling on the cursor. Releasing the block causes it to be deleted as it is connected.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/375e7db3-8d5c-4a17-9c1e-e45278393b44

### Browsers

_No response_

### LINKED ISSUE #10077: Blocks move using top left corner as origin instead of pointer coordinates
### Check for duplicates

### Description

When blocks are dragged on the workspace, the decision to delete them (or other actions like moving them to a backpack) is determined based on the block's `getRelativeToSurfaceXY()` coordinates. This creates unexpected behavior if the block is dragged from somewhere further right on the block.

### Reproduction steps

1. Using a mouse, click a block's middle or right side to begin a drag.
2. Move the block such that it's left edge slightly overlaps the toolbox without the cursor overlapping the block. (The block would be deleted.)
3. Move the cursor over the trashcan without touching the top-left corner of the block to the trashcan. (The block would not be deleted).

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

The mouse cursor could not be captured in these screenshots, unfortunately, but it was position at the `'to'` label at the middle of the block.

This block would be deleted if dropped here:

Image

This block would not be deleted if dropped here:

Image

### Browsers

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10077

### Proposed Changes

- `Dragger`: Delete-area detection during pointer drags now uses the cursor position (`PointerEvent.clientX` / `clientY`) instead of the block’s workspace origin (`getRelativeToSurfaceXY()`). Block movement is unchanged; only hit testing for delete styling, drag-target hooks, and drop/delete on `onDragEnd` follow the pointer.
- `WorkspaceSvg.getDragTarget`: Accepts only `PointerEvent` once again.
- Keyboard drags: `onDrag` skips delete-area logic when the event is not a `PointerEvent`; `onDragEnd` commits without delete-area checks.

This effectively reverts one particular change introduced by https://github.com/RaspberryPiFoundation/blockly/pull/9593

### Reason for Changes

Previously, deletion could trigger when the block origin overlapped a delete area even if the cursor was elsewhere, for example, dragging a block over the workspace while its top-left corner still intersected the toolbox column. Now, deletion depends on where the user is pointing, not where the block’s origin happens to be.

This conveniently seems to resolve https://github.com/RaspberryPiFoundation/blockly/issues/10076 as well, likely due to `getRelativeToWorkspaceXY()` returning the wrong sort of coordinates for zoomed workspaces. (I did not bother to root cause this, but did confirm it's no longer happening.)

### Test Coverage

New mocha suite simulating realistic drags with delete areas.

- Drag to trashcan center - delete styling applied; block disposed on drop.
- Drag to toolbox center - same behavior for toolbox delete area.
- Origin overlaps toolbox, cursor outside - block origin moved slightly into toolbox delete rect; cursor stays outside; no delete styling; block survives `onDragEnd`.


packages/blockly/core/workspace_svg.ts
```diff
@@ -1480,17 +1480,13 @@ export class WorkspaceSvg
   /**
    * Returns the drag target the pointer event is over.
    *
-   * @param e Pointer move event or a workspace coordinate.
+   * @param e Pointer move event.
    * @returns Null if not over a drag target, or the drag target the event is
    *     over.
    */
-  getDragTarget(e: PointerEvent | Coordinate): IDragTarget | null {
```
gonfunko, line 1487: Here especially, and kinda throughout, I think this really should have always taken a coordinate; I'd suggest leaving it as-is. Also this is technically breaking. In general I think only the actual event listener should handle events, and any methods called by it or its callees should take specific/abstracted arguments, rather than handing events around all over the place and picking one or two properties out of them.

packages/blockly/core/dragging/dragger.ts
```diff
@@ -107,31 +107,35 @@ export class Dragger implements IDragger {
...
+      // For keyboard events, we don't check for a drag target or delete area. Just commit the drag.
+      this.draggable.endDrag(e, DragDisposition.COMMIT);
+      if (isFocusableNode(this.draggable)) {
+        // Ensure focusable nodes end drag with focus and selection.
+        getFocusManager().focusNode(this.draggable);
+      }
+      return;
+    }
+
+    let wouldDelete = false;
```
gonfunko: Why was this moved back here and made a `let`? It seems like it's still only assigned in one place and read right after that down below?

packages/blockly/tests/mocha/dragger_test.js
```diff
@@ -0,0 +1,162 @@
...
+suite('Dragger', function () {
+  /**
+   * @param {!Blockly.BlockSvg} block The block to measure.
+   * @returns {{x: number, y: number}} Viewport coordinates at the block center.
+   */
+  function blockCenterClient(block) {
+    const bounds = block.getSvgRoot().getBoundingClientRect();
+    return {
+      x: (bounds.left + bounds.right) / 2,
+      y: (bounds.top + bounds.bottom) / 2,
```
gonfunko: This only maths if it's at (0,0) right? and same for `rectCenterClient` below

packages/blockly/core/dragging/dragger.ts
mikeharv: Good catch. There was a point within my refactoring that this made sense, but it ended up looking more like the original version than I anticipated at one point. Will revert.

packages/blockly/core/workspace_svg.ts
mikeharv: That's fair. I went with this change because it was essentially restoring it to how it worked a few months ago (https://github.com/RaspberryPiFoundation/blockly/pull/9593/files#diff-c4d6416d6519f5631c8a310a588345baac4285ff0f2b850012f29a783c83f9deL1434-L1439). I agree though that it does feel cleaner for this function to generally work with the actual coordinates and not a full event. I reverted this change and updated `Dragger` accordingly.

packages/blockly/tests/mocha/dragger_test.js
gonfunko: Disregard, I was apparently way under-caffeinated for this review. Sorry!

gonfunko (approved):

PR #10096, opened 2026-07-02
fix: delete-area hit testing at non-default zoom (#10095)

### LINKED ISSUE #10095: Deleting blocks via dragging does not work anymore when zoom isn't on default
### Check for duplicates

### Description

Whenever the zoom is on the default value I can delete blocks by dragging them into the trashcan or the toolbox.

But when I zoom in or out to any extent, dragging the block on either the toolbox or the trashcan, it doesn't gets deleted. It just drops back onto the workspace. 
The trashcan opening animation doesn't show as well in that state but the cursor still changes to the red cross (shown in the screenshot below).

This bug didn't appear in the v12 Versions but did happen in 3.0.0 and 3.0.1 - at least on my side

### Reproduction steps

1. create a workspace with zoom capabilities 
2. zoom in or out
3. try to delete a block by dragging it onto the trashcan or toolbox

### Priority

often occurs in the workflow

### Stack trace

```shell
No error messages appearing
```

### Screenshots

Image

### Browsers

Appeared in firefox and chrome, likely universal

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10095

### Proposed Changes

Hit-test delete areas now us client/viewport coordinates (`clientX`/`clientY`) instead of a client-to-workspace-to-client round-trip

### Reason for Changes

After #10078, delete styling would still appear at non-default zoom but blocks were not reliably deleted and the trashcan lid did not animate. The coordinate round-trip misaligned hit tests with `getBoundingClientRect()` at certain zoom levels, but using client coordinates fixes that. 

### Test Coverage

- Extended `dragger_test.js` with generated trashcan/toolbox delete tests at zoom 0.7 and 1.5 (delete style, trashcan lid, disposal)
- Generated extra mutator workspace tests with a zoomed parent workspace (delete on pointer-over-area; no false positive on origin-only overlap). Unrelated to bug report, but felt useful enough to add.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
     "@commitlint/config-conventional": "^21.0.2"
   },
   "overrides": {
-    "eslint": "10.5.0",
-    "prettier": "3.8.4"
+    "eslint": "10.3.0",
+    "prettier": "3.8.3"
   },
   "scripts": {
     "test": "npm run test --ws --if-present",
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
     "@commitlint/config-conventional": "^21.0.2"
   },
   "overrides": {
-    "eslint": "10.3.0",
-    "prettier": "3.8.3"
+    "eslint": "10.5.0",
+    "prettier": "3.8.4"
   },
   "scripts": {
     "test": "npm run test --ws --if-present",
--- a/packages/blockly/core/dragging/dragger.ts
+++ b/packages/blockly/core/dragging/dragger.ts
@@ -16,7 +16,6 @@ import type {IDragger} from '../interfaces/i_dragger.js';
 import {isFocusableNode} from '../interfaces/i_focusable_node.js';
 import * as registry from '../registry.js';
 import {Coordinate} from '../utils/coordinate.js';
-import {screenToWsCoordinates} from '../utils/svg_math.js';
 
 export class Dragger implements IDragger {
   protected startLoc: Coordinate;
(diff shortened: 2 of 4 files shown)


mikeharv:
CI on this PR is failing because `main` has a `package.json` / `package-lock.json` override mismatch (`npm ci` missing `eslint@10.3.0` / `prettier@3.8.3`). That is unrelated to the delete-area fix.

I opened a separate chore PR to fix `main` for everyone: #10097

I rebased `zoom-delete` onto that fix so Node.js CI should pass here once the re-run completes.

lizschwab (approved):
Looks good once we get CI happy again

microbit-matt-hillsdon:
I think this PR is correct and a net simplification but I was puzzled about the mechanism behind the bug as I felt the coordinates should roundtrip ok.

After poking around with Claude I think this is what went wrong:

1. Dragger.onDrag computes coordinate once and hands the same object to wouldDeleteDraggable(coordinate) and updateDragTarget(coordinate) (and onDragEnd does something similar).
2. Each getDragTarget(coordinate) calls wsToScreenCoordinates(this, coordinate).
3. wsToScreenCoordinates does workspaceCoordinates.scale(ws.scale), and **Coordinate.scale() is mutating**
4. So the first call scales the shared object by ws.scale in place; the second call receives an already-×scale coordinate, scales it again, and lands outside the delete area.
5. At scale = 1 it's a no-op so all's well. At any other zoom it makes a mess.

This explains why the red-cross cursor still shows (call 1, wouldDeleteDraggable, sees the correct coordinate) but things controlled by later calls go wrong (trashcan lid).

From a non-Blockly-team perspective it seems a footgun  that wsToScreenCoordinates mutates its argument. This seems like it might be an easy fix. I feel the same way about Coordinate's fluent-but-mutating API but that is a much more disruptive change.

microbit-matt-hillsdon:
@mikeharv, interested in what you think about the [mutability in wsToScreenCoordinates](https://github.com/RaspberryPiFoundation/blockly/pull/10096#issuecomment-4875463937), worth some clean-up?

PR #10115, opened 2026-07-08
docs: refactor toolbox codelab to use sample app

zspriggs:
## The basics

### Proposed Changes

Refactors the "Customizing a Blockly toolbox" codelab to use the sample app as starter code, rather than custom starter code. This is one of the more extensive refactors, because the original codelab used an XML toolbox, but the sample app uses a JSON toolbox, so the codelab had to be changed accordingly in a few different places. 

There are also a few other changes in this refactor that were made for consistency and/or clarity:
* Moved setup of custom class from `Setup` step to new `Make a Custom Category` step. This makes this codelab more consistent with other codelabs, which have very minimal `Setup` steps that don't include any coding. 
* Added call to `super()` in `setSelected()` override and explained why. The old codelab set the ARIA state in the override without really explaining it, so I changed it to use `super` for slightly better form and added an explanation.
* Changed Blockly logo to toolbox emoji. Previously the Blockly logo svg was shipped with the starter code, but since the starter code is now the sample app, I switched it to a toolbox emoji so that users don't have to deal with a separate download.


packages/docs/docs/codelabs/custom-toolbox/codelab-overview.mdx
```diff
@@ -40,8 +40,9 @@ The code samples are written in ES6 syntax. You can find the code for the [compl
...
 - A text editor.
 - Basic knowledge of HTML, CSS, and JavaScript.
+- node.js and npm installed ([instructions](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)).
+- Comfort using the command line/terminal.
 - Basic understanding of the [Blockly toolbox](/guides/configure/toolboxes/toolbox).
 
 Throughout various parts of this codelab we will be talking about [toolbox definitions](/guides/configure/toolboxes/category?tab=xml).
-The toolbox definition can be written in XML or JSON. We will be using an XML
-toolbox definition that can be found in the provided code.
+The toolbox definition can be written in XML or JSON. We will be using a JSON toolbox definition, which is the preferred format.
```
gonfunko: I'm kind of inclined to just not mention XML or say anything here; while it is true that it remains supported, it's solely for backwards compatibility and really shouldn't be used for anything new.

packages/docs/docs/codelabs/custom-toolbox/setup.mdx
```diff
@@ -8,92 +8,34 @@ import Image from '@site/src/components/Image';
...
 ## 2. Setup
 
-### Download the sample code
+### Create the application
 
-You can get the sample code for this codelab by either downloading the zip here:
+Use the [`npx @blockly/create-package app`](https://www.npmjs.com/package/@blockly/create-package) command to create a standalone application that contains a sample setup of Blockly, including custom blocks and a display of the generated code and output.
 
-[Download zip](https://github.com/RaspberryPiFoundation/blockly/archive/main.zip)
+1. Run `npx @blockly/create-package app custom-toolbox-codelab` in your terminal. This will create a Blockly application in the folder `custom-toolbox-codelab`.
```
gonfunko, line 15: It'd be nice to add the `--typescript` flag here to generate a TS starter app (this means the finished version will need updating to TS, sorry!). I think TS is sufficiently adopted at this point that that should be fine, and the benefits in terms of typechecking can be especially useful for beginners who aren't familiar with the codebase and are working through a codelab.

packages/docs/docs/codelabs/custom-toolbox/setup.mdx
```diff
@@ -8,92 +8,34 @@ import Image from '@site/src/components/Image';
...
-
-If you downloaded the source as a zip, unpacking it should give you a root folder named `blockly-main`.
-
-The relevant files are in `packages/docs/docs/codelabs/custom-toolbox`. There are two versions of the app:
-
-- `starter-code/`: The starter code that you'll build upon in this codelab.
-- `complete-code/`: The code after completing the codelab, in case you get lost or want to compare to your version.
-
-Each folder contains:
+The initial application has a toolbox definition included in already included in the `toolbox.js` file.
```
gonfunko: "has a toolbox definition already included in"

packages/docs/docs/codelabs/custom-toolbox/add-a-custom-category.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+# Customizing a Blockly toolbox
+
+## 3. Make a custom category
+
+### Define and register a custom category
+
+To start, create a file named `custom_category.js` in the `src` directory.  
+
+Then, add an `import` statement to `index.js` for this new file:
```
gonfunko: This might be a little clearer as "in" than "to" - I initially read it as wanting you to do `import './index.js'` in the new file.

packages/docs/docs/codelabs/custom-toolbox/add-a-custom-category.mdx
```diff
@@ -0,0 +1,76 @@
...
+import * as Blockly from 'blockly';
+
+class CustomCategory extends Blockly.ToolboxCategory {
+  /**
+   * Constructor for a custom category.
+   * @override
+   */
+  constructor(categoryDef, toolbox, opt_parent) {
+    super(categoryDef, toolbox, opt_parent);
+  }
```
gonfunko: Since this constructor is just a passthrough to super, maybe just omit it entirely?

packages/docs/docs/codelabs/custom-toolbox/add-a-custom-category.mdx
```diff
@@ -0,0 +1,76 @@
...
+}
+```
+
+After defining your category you need to tell Blockly that it exists.
+Register your category by adding the below code to the end of `custom_category.js`:
```
gonfunko: Best practice these days is to export a `registerCustomCategory` function and call that, rather than just automatically registering the thing, so that imports are side-effect-free and so that you don't get duplicate registrations if e.g. something else depends on the category and you also wind up importing that.

packages/docs/docs/codelabs/custom-toolbox/change-the-look-of-a-category.mdx
```diff
@@ -56,11 +68,11 @@ similar to the below toolbox.
...
 
-Create a file named `toolbox_style.css` in the same directory as `index.html`
-and include it in `index.html`:
+Create a file named `toolbox_style.css` in the same directory as `index.js`.
+Then, import the new file in `index.js`:
 
-```
-<link rel="stylesheet" href="toolbox_style.css">
+```js
+import './toolbox_style.css';
```
gonfunko, line 75: I'm assuming this is some sort of Webpack magic, maybe worth adding a note about that?

packages/docs/docs/codelabs/custom-toolbox/change-the-look-of-a-selected-category.mdx
```diff
@@ -37,15 +38,55 @@ setSelected(isSelected){
...
+     // Set the text back to white.
+     labelDom.style.color = 'white';
+   }
+}
+```
+
+:::note 
+Check the parent implementation whenever you override a Blockly method. This
+will help ensure that you are properly implementing non-visual processes that are 
+important.
```
gonfunko: Maybe word this a little stronger, something to the effect of always review the superclass's implementation and call `super()` unless you have a good reason not to.

packages/docs/docs/codelabs/custom-toolbox/add-an-icon-to-your-category.mdx
```diff
@@ -6,38 +6,52 @@ import Image from '@site/src/components/Image';
...
 # Customizing a Blockly toolbox
 
-## 5. Add an icon to your category
+## 6. Add an icon to your category
 
 We are going to add an icon to our "Logic" category by adding an icon library to
 our `index.html` file, and setting the appropriate CSS class on our category definition.
 
-To start, we are going to grab an icon library and add it to `index.html`:
+To start, we are going to grab an icon library and paste it to `index.html`, inside
```
gonfunko, line 14: Nit: s/to/into

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
```diff
@@ -15,18 +15,19 @@ For this example, we are going to create a toolbox label.
...
 
-```html
-<script src="toolbox_label.js"></script>
+```js
+import './toolbox_label.js';
 ```
 
-Create a class in `toolbox_label.js` that extends `Blockly.ToolboxItem`
-and register it.
+Create a class in `toolbox_label.js` that extends `Blockly.ToolboxItem` and register it.
```
gonfunko: Here too I'd recommend omitting the passthrough constructor and putting the registration code in its own exported function, and calling that from index.js

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
```diff
@@ -91,11 +99,11 @@ Next, we are going to return this element:
   }
 ```
 
-If you open the `index.html` file you should see a label above your first category.
+If you refresh the app page you should see a label above your first category.
```
gonfunko, line 119: Does this work OK with keyboard nav?

packages/docs/docs/codelabs/custom-toolbox/setup.mdx
zspriggs: You make a really good point, but if we're switching this codelab to TS then I'd want to make the other codelabs use TS for consistency, which should probably be its own PR since it would also affect codelabs that I haven't been refactoring. I think we should discuss this more though, since you're right that typechecking would be helpful

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
zspriggs: Ah, good catch, I just tested and it does not. It essentially stops you from being able to cycle the toolbox, bc when you hit the top or bottom you attempt to focus the label, which throws a console error since it doesn't have a valid id. I'll look into this.

zspriggs: Okay, fixed the keyboard navigation. I also added some better aria handling so that the label can be actually seen and read off by the screenreader, but a side effect of that was that I had to make it focusable as well. I added a `roledescription` that says "Label" to indicate that this isn't a proper category, but I'm still not 100% sure if this is the ideal way to handle this. The only other thing that seemed to "work" was just hiding the label, but then that makes it invisible to screenreaders.

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
zspriggs: Done, I also omitted the constructor as you recommended for the custom category since it also just calls super

packages/docs/docs/codelabs/custom-toolbox/change-the-look-of-a-category.mdx
zspriggs: Yes it is webpack, I just added a note

packages/docs/docs/codelabs/custom-toolbox/setup.mdx
gonfunko: Yeah agreed, fine to leave as-is for now and we can take a TS conversion pass later if folks are onboard with that.

packages/docs/docs/codelabs/custom-toolbox/complete-code/src/index.js
```diff
@@ -12,13 +12,18 @@ import { save, load } from './serialization';
 import { toolbox } from './toolbox';
 import './index.css';
 import './toolbox_style.css';
-import './custom_category.js';
-import './toolbox_label.js';
+import { CustomCategory, registerCustomCategory } from './custom_category.js';
+import { ToolboxLabel, registerToolboxLabel } from './toolbox_label.js';
```
gonfunko: The classes themselves don't seem to be used and could be removed in favor of just importing the registration functions

packages/docs/docs/codelabs/custom-toolbox/add-a-custom-category.mdx
```diff
@@ -12,39 +12,44 @@ import Image from '@site/src/components/Image';
...
+    true,
+  );
+}
+```
+
+Finally, you'll need to import the new category, and registration function, in `index.js`.
+To register the custom category, call the `registerCustomCategory()` function in `index.js`, just after registering the blocks and generator:
+
+```js
+import { CustomCategory, registerCustomCategory } from './custom_category.js';
```
gonfunko: Here too, no need to import the class

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
```diff
@@ -16,31 +16,38 @@ For this example, we are going to create a toolbox label.
...
+    ToolboxLabel,
+  );
+}
 ```
 
+As with the custom category, you'll also need to import the `ToolboxLabel` and registration function in `index.js`.
+To register the toolbox label, call the `registerToolboxLabel()` function in `index.js`, just after registering the custom category:
+
+```js
+import { ToolboxLabel, registerToolboxLabel } from './toolbox_label.js';
```
gonfunko: And here

packages/docs/docs/codelabs/custom-toolbox/complete-code/src/toolbox_label.js
```diff
@@ -7,14 +7,15 @@
...
-    super(toolboxItemDef, parentToolbox);
-  }
-
   /** @override */
   init() {
     // Create the label.
     this.label = document.createElement('label');
+    // Set ARIA role and description, so that screenreaders can see and read the label
+    this.label.setAttribute('role', 'treeitem');
+    this.label.setAttribute('aria-roledescription', 'label');
```
gonfunko: Optional, but might be worth using `Blockly.utils.aria` here

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
```diff
@@ -92,6 +92,9 @@ First, we are going to add an init method to the `ToolboxLabel` class that will
   init() {
     // Create the label.
     this.label = document.createElement('label');
+    // Set ARIA role and description, so that screenreaders can see and read the label
+    this.label.setAttribute('role', 'treeitem');
+    this.label.setAttribute('aria-roledescription', 'label');
```
gonfunko: The aria utils would be good here too

packages/docs/docs/codelabs/custom-toolbox/change-the-look-of-a-category.mdx
gonfunko: Thanks!

packages/docs/docs/codelabs/custom-toolbox/adding-a-custom-toolbox-item.mdx
gonfunko: Thanks! That seems OK for now, this is kind of a contrived example anyway – as long as it's not actively breaking the experience that seems adequate.

packages/docs/docs/codelabs/custom-toolbox/complete-code/src/index.js
zspriggs: Done, and adjusted in the .mdx files too

packages/docs/docs/codelabs/custom-toolbox/complete-code/src/toolbox_label.js
zspriggs: Good call, done here and in the .mdx file

packages/docs/docs/codelabs/custom-toolbox/complete-code/src/toolbox_label.js
```diff
@@ -12,8 +12,12 @@ class ToolboxLabel extends Blockly.ToolboxItem {
     // Create the label.
     this.label = document.createElement('label');
     // Set ARIA role and description, so that screenreaders can see and read the label
-    this.label.setAttribute('role', 'treeitem');
-    this.label.setAttribute('aria-roledescription', 'label');
+    Blockly.utils.aria.setRole(this.label, 'treeitem');
```
gonfunko: treeitem/label correspond to Blockly.utils.aria.ROLE/Blockly.utils.aria.STATE enums, here and in the other file.

gonfunko (approved):
One minor nit but feel free to submit after, thanks!

PR #10128, opened 2026-07-10
chore: move formatting to root level

### LINKED ISSUE #7761: Set up formatting

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/7761

### Proposed Changes

This change moves the following to the root level:
 - prettier config & ignore
 - gitignore
 - format cli commands

### Reason for Changes

This change helps to get the workspace ready for moving the plugins over to the monorepo.

### Test Coverage

No test coverage. I validated my changes by inserting formatting errors in the core and docs workspaces and ensuring prettier still catches them and doesn't catch anything it shouldn't.

Before my changes:
Screenshot 2026-07-09 at 4 39 28 PM

After my changes:
Screenshot 2026-07-10 at 1 36 17 PM


.prettierignore
```diff
@@ -0,0 +1,48 @@
...
+
+# Core: Demos, scripts, misc
+packages/blockly/node_modules/*
+packages/blockly/demos/*
+packages/blockly/appengine/*
+packages/blockly/externs/*
+packages/blockly/closure/*
+packages/blockly/scripts/gulpfiles/*
+packages/blockly/CHANGELOG.md
+packages/blockly/PULL_REQUEST_TEMPLATE.md
```
gonfunko, line 33: I realize these were copied from the old config file, but a variety of these no longer exist (in particular closure, externs, tests/blocks, tests/jsunit, tests/screenshot, tests/workspace_svg, and possibly others?) Optional, but would be nice to clean this up and remove them (and anything else obsolete in other config files)

.prettierignore
```diff
@@ -0,0 +1,48 @@
...
+
+# Core: Don't bother formatting JavaScript files we're about to migrate:
+packages/blockly/generators/**/*.js
```
gonfunko, line 36: This migration is done, and this rule can also be removed.

gonfunko (approved):

.prettierignore
lizschwab: I always like an opportunity to remove cruft! I'll compare these to the directories and remove anything that's no longer present.

PR #10129, opened 2026-07-10
fix: Consolidate select events

### LINKED ISSUE #9087: Recursive block selects firing with v12
### Check for duplicates

### Description

With v12.0.0, Blockly seems to be firing far more `Blockly.Events.SELECTED` events than normal. This is leading to performance issues.

### Reproduction steps

1. Visit any v12 demo page, such as https://google.github.io/blockly-samples/plugins/field-grid-dropdown/test/index.html
2. Using the console, add a simple event listener to log block selection events. For example:
```
Blockly.getMainWorkspace().addChangeListener(e=>{if(e.type===Blockly.Events.SELECTED){console.log(e);}})
```
3. Drag a block out of the toolbox and connect it to another block.
4. Inspect the console. You will see over a dozen selection events.

This does not happen with v11, such as at https://blockly-demo.appspot.com/static/tests/playground.html

### Stack trace

```shell

```

### Screenshots

Image

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #9087

### Proposed Changes
This PR consolidates and normalizes select events. Much of the situation the bug describes was already addressed, I suspect through improvements to the focus manager. There were however still several classes of duplicative and confusing select events:

* When blocks were inserted from the flyout, they had the same ID as the block in the flyout. This is legal, because the main workspace and flyout workspace are separate, but resulted in select events being fired where the `oldElementId` and `newElementId` were the same, because the `oldElementId` was that of the block in the flyout, which was initially selected when a drag started, and the `newElementId` was that of the newly-created block in the main workspace, which was the same ID. Now, IDs are not preserved when inserting blocks from the flyout (this was already the case for subsequent insertions of the same block, because you can't have two blocks with the same ID on the same workspace).
* Select events are fired both when an element loses selection and when it gains it. This means that if block A was selected, then block B was clicked, block A would fire a select event from A to undefined (when it lost selection) and block B would fire a select event from A to B (when it gained selection). These are now consolidated in the event stream to one select event from A to B. Note that we can't just not fire events on selection loss, because you can e.g. click the workspace, which would cause a selected block to become unselected, but not cause anything else to gain selection. The workspace would gain focus, but selection events are for selection, which is a different concept.
* Moving elements in the DOM causes them to lose focus, which Blockly works around in several places by manually reassigning focus post-DOM move. If a selected block was e.g. brought to the front, this would cause it to fire a select event from itself to itself. These are now filtered out of the event stream.

I manually verified that the event stream for the actions described in the bug looks reasonable, and added unit tests to verify the consolidation/filtering logic described above, and ensured that they failed without these changes.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -232,7 +232,7 @@ export class BlockDragStrategy implements IDragStrategy {
     } else if (this.block.isInFlyout && this.block.workspace.targetWorkspace) {
       const rootBlock = this.block.getRootBlock();
 
-      const json = blocks.save(rootBlock);
+      const json = blocks.save(rootBlock, {saveIds: false});
       if (json) {
         const newBlock = blocks.appendInternal(
           json,
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -232,7 +232,7 @@ export class BlockDragStrategy implements IDragStrategy {
     } else if (this.block.isInFlyout && this.block.workspace.targetWorkspace) {
       const rootBlock = this.block.getRootBlock();
 
-      const json = blocks.save(rootBlock, {saveIds: false});
+      const json = blocks.save(rootBlock);
       if (json) {
         const newBlock = blocks.appendInternal(
           json,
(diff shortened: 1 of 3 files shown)


mikeharv (changes requested):
The selection event fixes make sense to me, but I think losing the ids of flyout blocks is potentially a breaking change that warrants more internal discussion. Could we remove that change for now?

gonfunko:
Reverted that change for now.

mikeharv (approved):

PR #10130, opened 2026-07-10
fix: Fix regression in trashcan lid behavior

### LINKED ISSUE #10112: Trashcan doesn't animate
### Check for duplicates

### Description

It was reported before but the issues have been closed.
Since v13, the trashcan doesn't animate.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #10112

### Proposed Changes
This PR fixes some regressions in the behavior of the trashcan. In v12, when a block was dragged over the trashcan, its lid would open to indicate that the block could be dropped there. In v13, this only happened if the trashcan already contained one or more items.

Additionally, in v12, hovering over an empty trashcan did not saturate its color, but in v13 it does.

This PR addresses both issues to align v13 with the v12 behavior: dragging a block over an empty (or full) trashcan will saturate it and animate the lid open, and merely hovering over an empty trashcan will not saturate it. Additionally, the keyboard/screenreader behavior was made more consistent with the mouse: when the trashcan is empty, it is marked as `aria-disabled`, since clicking/pressing enter on it has no effect, and it is not saturated when focused and empty. When empty, the trashcan's label indicates that, rather than saying "Open trash", which is impossible. When full, the label updates to "Open trash".

### DIFF
--- a/packages/blockly/core/trashcan.ts
+++ b/packages/blockly/core/trashcan.ts
@@ -172,7 +172,12 @@ export class Trashcan
     );
 
     aria.setRole(this.svgGroup, aria.Role.BUTTON);
-    aria.setState(this.svgGroup, aria.State.LABEL, Msg['OPEN_TRASH']);
+    aria.setState(
+      this.svgGroup,
+      aria.State.LABEL,
+      Msg['ARIA_LABEL_TRASH_EMPTY'],
+    );
+    aria.setState(this.svgGroup, aria.State.DISABLED, true);
 
     let clip;
     const rnd = String(Math.random()).substring(2);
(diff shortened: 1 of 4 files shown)


packages/blockly/core/trashcan.ts
```diff
@@ -569,6 +581,8 @@ export class Trashcan
     }
 
     this.svgGroup?.classList.add(TRASH_FULL);
```
mikeharv: Nit: Safe to remove `?` now?

mikeharv (approved):

packages/blockly/core/trashcan.ts
gonfunko: Done!

Issue #2249, opened 2019-02-01
Filter out insertion markers in getDescendants and getChildren


rachel-fenichel:
### Problem statement
Insertion markers are returned when you query for a block's descendants/children, which is unexpected.

### Expected Behavior

Blockly manages insertion markers internally, and developers calling our external APIs don't need to check for insertion markers.

### Actual Behavior

Insertion markers are filtered out at workspace.getAllBlocks but not at lower levels.

### Additional Information
Most of the work here is in figuring out where this filtering needs to happen.
See https://github.com/google/blockly/pull/2243

gonfunko:
Closing as #10156 obviates this problem by making insertion markers not be blocks at all.

PR #10156, opened 2026-07-17
feat: Rewrite insertion markers to improve performance and fix memory leaks

### LINKED ISSUE #5375: Refactor insertion markers to not be blocks

Insertion markers cause lots of bugs in Blockly. This is because they are just blocks with an extra flag set to true, instead of being their own independent type. As such, anywhere that deals with blocks has to explicitly filter out insertion markers, and if we forget, we get bugs.

Here is a list of bugs related to this problem with insertion markers:
  * #2216
  * #2249
  * #2322
  * #2354
  * #2948
  * #4048
  * #4485 
  * #4636 
  * #5374 

There may be more that I didn't catch.

Refactor the insertion markers to not be blocks. They should be their own type that does not come up in getDescendants or getChildren or getAllBlocks calls. They should have connections (or some connection-like interface) but they should not have fields or icons.

Basically, we should use the type system to make sure these errors never occur in the first place, instead of checking flags and chasing bugs.

**Issues**
One big issue with this would be figuring out how to make the insertion marker have the correct shape. Right now it works by:
  1) Creating a block
  2) Setting all of the field values so they have the correct widths
  3) Rendering the block (since the rendering system is tightly coupled to the BlockSvg type)
  4) Hiding all of the fields so the user can't see them
  5) Setting the block's color to grey so it looks like an "insertion marker" (even though it's just a block with a flag)

The rendering system would need to be decoupled from the BlockSvg type before we can refactor insertion markers.

gonfunko:
## The basics

### Resolves

Fixes #5375 and #2249 once old behavior is removed.

### Proposed Changes
This PR rewrites Blockly's insertion markers to be SVG groups/paths on the workspace, and not actual blocks. This fixes a long-standing leaky abstraction, and more importantly significantly improves performance when attempting to connect blocks to large stacks. It also incidentally fixes an issue where all insertion marker blocks were being leaked, causing Blockly's memory usage to grow indefinitely every time an insertion marker was shown.

### Reason for Changes
Drawing insertion markers, particularly in large stacks of blocks, could be extremely slow. This is because, every time an insertion marker was shown (which was on every move), the dragging block would be cloned, the proposed connection point would be disconnected from its parent if any, and the cloned block would be connected to the proposed connection point. Connecting/disconnecting results in large DOM shuffles, because blocks are nested within each other in the DOM. This in turn caused a bunch of layout thrashing and redraw. Additionally, every insertion marker ever created was retained forever.

Now, when an insertion marker is rendered, the dragging block has its connections updated to the proposed connection (without using the connect() method, which shuffles the DOM/fires events/etc – the connection's `targetConnection` field is just set), is rerendered to extract its path, has the path used as the insertion marker, and is rerendered in its original state. This rendering involves only the block being dragged and the block it is being connected to, and doesn't involve any DOM modification outside of the blocks having their paths set and the insertion marker being added at the root of the workspace, which don't involve shifting and invalidating the layout of large DOM trees.

For users using the built-in renderers, this change will default on, and is not breaking. There are minor behavioral breaking changes possible for users with custom renderers, so this revised behavior defaults off for non-built-in renderers. Those users may opt in by setting `Blockly.InsertionMarkerPreviewer.useFastInsertionMarkers = true`. This will become the default behavior in v14.

This change significantly improves performance when e.g. dragging a new block into the top of the "spaghetti" block stack created in the playground in Safari, Firefox and Chrome.

Old:
https://github.com/user-attachments/assets/9c983a6c-ae7f-4e01-bcea-da9171e49242

New:
https://github.com/user-attachments/assets/f36e4fbd-1983-4ee8-951c-568f897894ce

Old profile (red flags indicate jank/dropped frames; note also continuously increasing node and listener counts):
Screenshot 2026-07-17 at 3 00 34 PM

New profile (note lack of jank, and stair-step node/listener count as GC runs):
Screenshot 2026-07-17 at 3 01 57 PM

### Test Coverage
Existing tests continue to pass. I manually tested in Chrome, Safari and Firefox on Mac. I verified that insertion markers were correctly displayed when attaching blocks horizontally and vertically at the top/bottom/start/end of a stack/row, splicing them into the middle of a stack, and wrapping statement blocks around a stack, in Thrasos and Zelos, with mouse drags and keyboard-driven drags.

### DIFF
--- a/packages/blockly/core/insertion_marker.ts
+++ b/packages/blockly/core/insertion_marker.ts
@@ -1,0 +1,165 @@
+/**
+ * @license
+ * Copyright 2026 Raspberry Pi Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type {BlockSvg} from './block_svg.js';
+import * as renderManagement from './render_management.js';
+import {RenderedConnection} from './rendered_connection.js';
+import {Coordinate} from './utils/coordinate.js';
+import * as dom from './utils/dom.js';
+import {Size} from './utils/size.js';
+import {Svg} from './utils/svg.js';
+
+/**
+ * Visual representation of a mid-drag block if it were to be connected.
+ */
+export class InsertionMarker {
+  /** The current size of the insertion marker, in workspace units. */
+  private size: Size = new Size(0, 0);
+
+  /** The DOM element representing the insertion marker. */
+  private marker?: SVGGElement;
+
+  /** The static connection to which the insertion marker is attached. */
+  private parentConnection?: RenderedConnection;
+
+  /**
+   * Returns the current size of the insertion marker in workspace units.
+   */
+  getHeightWidth() {
+    return this.size;
+  }
+
+  /**
+   * Displays an insertion marker representing the block structure if
+   * `draggingConnection` were to be connected to `staticConnection`.
+   *
+   * @param staticConnection The proposed connection on the stationary block.
+   * @param draggingConnection The proposed connection on a block being dragged.
+   */
+  show(
+    staticConnection: RenderedConnection,
+    draggingConnection: RenderedConnection,
+  ) {
+    this.parentConnection = staticConnection;
+    this.parentConnection.attachInsertionMarker(this);
+
+    const connectingBlock = draggingConnection.getSourceBlock();
+
+    // Save all the existing connections pairs on the dragged block.
+    const oldConnections = new Map<
+      RenderedConnection,
+      RenderedConnection | null
+    >();
+    for (const connection of connectingBlock.getConnections_(false)) {
+      oldConnections.set(connection, connection.targetConnection);
+      connection.targetConnection = null;
+    }
+
+    // Temporarily connect the dragging and static connections and render the
+    // dragging block to steal its size and path for use by the insertion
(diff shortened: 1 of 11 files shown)


packages/blockly/core/insertion_marker.ts
```diff
@@ -0,0 +1,165 @@
...
+    const oldConnections = new Map<
+      RenderedConnection,
+      RenderedConnection | null
+    >();
+    for (const connection of connectingBlock.getConnections_(false)) {
+      oldConnections.set(connection, connection.targetConnection);
+      connection.targetConnection = null;
+    }
+
+    // Temporarily connect the dragging and static connections and render the
```
maribethb, line 61: For readability I might extract this to another method so that when you're trying to know what `show` does you just need to know that we have some way to get the size and path info.

maribethb (approved):

packages/blockly/core/insertion_marker.ts
gonfunko: Discussed in chat, but that would need the connection stashing/unstashing as well, and the offset calculation needs to happen before they're unstashed, and then it's pretty much the whole method body :/

PR #10162, opened 2026-07-19
fix: throw on invalid XML connections

### LINKED ISSUE #9266: XML deserializes weirdly on failed connection

jianchang56:
## The basics

### Resolves

Fixes #9266

### Proposed changes

XML deserialization now checks whether connecting a child block to its serialized parent succeeds. If `Connection.connect()` returns `false`, deserialization throws a `TypeError` containing the developer-facing message from the workspace connection checker.

Regression tests cover failed connections for both value inputs and next statements.

### Reason for changes

Previously, `domToBlockHeadless` ignored the boolean result of `parentConnection.connect()`. When connection checks rejected serialized XML, the child block remained in the workspace without being connected to the root block. Because only descendants of the root block were initialized, this produced an uninitialized and potentially invisible "ghost" block.

This change follows the behavior agreed on in the issue discussion: invalid serialized state should fail during deserialization, matching the existing JSON deserialization behavior. It deliberately does not recover, disable, reposition, or otherwise correct invalid blocks, and it does not change JSON deserialization. This is the key difference from the approaches in #9715 and #9718, which were closed because they attempted to preserve invalid children as top-level blocks.

### Test coverage

- Added XML regression coverage for an incompatible value input connection.
- Added XML regression coverage for an incompatible next-statement connection.
- `gulp build --debug`
- Full browser Mocha suite: 0 failures
- Targeted browser Mocha regression tests: 0 failures
- ESLint on changed files
- Prettier check on changed files
- `git diff --check`

### Documentation

No documentation changes are required because this fixes error handling for invalid legacy XML and does not change a supported serialization format or public API.

### DIFF
--- a/packages/blockly/core/xml.ts
+++ b/packages/blockly/core/xml.ts
@@ -1004,22 +1004,37 @@ function domToBlockHeadless(
 
   // Connect parent after processing mutation and before setting fields.
   if (parentConnection) {
+    let childConnection: Connection;
     if (connectedToParentNext) {
       if (block.previousConnection) {
-        parentConnection.connect(block.previousConnection);
+        childConnection = block.previousConnection;
       } else {
         throw TypeError('Next block does not have previous statement.');
       }
     } else {
       if (block.outputConnection) {
-        parentConnection.connect(block.outputConnection);
+        childConnection = block.outputConnection;
       } else if (block.previousConnection) {
-        parentConnection.connect(block.previousConnection);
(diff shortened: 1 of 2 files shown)


gonfunko (approved):
Thanks for the PR, this looks good!

PR #10168, opened 2026-07-22
chore(deps): update jsdom peer dependency version range

Carlosamouco:
### Reason for Changes

Blockly currently declares jsdom peer support as ^27.4.0, which restricts consumers to the 27.x major line. However the project itself is developed and tested with jsdom@29.1.1. Expanding the peer dependency range to >=27.4.0 <30.0.0 aligns the declared compatibility with the versions currently used by the project while still protecting against future breaking changes in jsdom 30.


maribethb (approved):
Thanks!

PR #10171, opened 2026-07-22
fix: Fix bug that caused some connections to refuse to connect when dragging from the toolbox

### LINKED ISSUE #10145: With Zelos "connection point" is not computed correctly for a block that comes from toolbox
### Check for duplicates

### Description

With Zelos, when I get a block directly from the toolbox, the "connection point" (sorry if I don't have the correct vocabulary!) is not correctly computed.

When I drop the same block and drag it again from workspace, the computation is correct.

PS: it doesn't seem to be a regression from v13.

### Reproduction steps

_No response_

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/f0342cde-be09-42d9-b192-cd4efaa20f4c

### Browsers

_No response_

gonfunko:
## The basics

### Resolves

Fixes #10145

### Proposed Changes
This PR fixes a bug that caused only some connections to be connectable when dragging a block out of the toolbox. The new block wasn't being rendered before connections were cached, so all of its connections were cached as if they were located at (0, 0) on the block. Now, the block is immediately rendered, so connections get their correct offsets and the connection cache is correctly populated.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -242,6 +242,7 @@ export class BlockDragStrategy implements IDragStrategy {
           },
         ) as BlockSvg;
         eventUtils.setRecordUndo(false);
+        newBlock.render();
         this.positionNewBlock(this.block, newBlock);
         eventUtils.setRecordUndo(true);
 


mikeharv (approved):
Nice!

PR #10178, opened 2026-07-23
chore: fix monorepo CI build for dev-tools

### LINKED ISSUE #7763: Pull in `dev-tools` and make sure it works with core and themes plugins

lizschwab:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/7763, https://github.com/RaspberryPiFoundation/blockly/issues/10144

### Proposed Changes

- Fixes the workspace declaration for dev-tools so that it is correctly added to the monorepo workspaces
- Adds building all workspaces and running the docs script from blockly so that build can run properly on github actions.

### Reason for Changes

The dev-tools plugin was incorrectly specified in the package.json. Also, adding additional build steps ensures we no longer break docs when adding or updating dependencies.

### Test Coverage

No new tests were added. I validated this change locally by running the workflow steps (npm ci, npm run build, npm run test) locally after removing all the node modules, dist, build, and the docs reference directories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "packages/plugins/theme-highcontrast",
     "packages/plugins/theme-modern",
     "packages/plugins/theme-tritanopia",
-    "packages/dev-tools",
+    "packages/plugins/dev-tools",
     "packages/docs"
   ],
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "test": "npm run test --ws --if-present",
     "lint": "npm run lint --ws --if-present",
     "lint:fix": "npm run lint:fix --ws --if-present",
-    "build": "npm run package --workspace=blockly && npm run build --ws --if-present",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "test": "npm run test --ws --if-present",
     "lint": "npm run lint --ws --if-present",
     "lint:fix": "npm run lint:fix --ws --if-present",
-    "build": "npm run package --workspace=blockly && npm run build --ws --if-present",
+    "build": "npm run package --workspace=blockly && npm run docs --workspace=blockly && npm run build --ws --if-present",
```
lizschwab, line 50: I have plans to split out building docs from building everything else, but I'm going to wait until I have Nx working as I feel I'll have better control there. For now, individual packages can be built from their respective workspaces and from the top level will build everything.

gonfunko (approved):

PR #10206, opened 2026-07-28
fix: Flyout toolbar changes block position

### LINKED ISSUE #9992: Flyout of toolbox that has sub-categories AND sub-blocks at the same time pushes and glitches blocks in workspace
Flyout of toolbox that has sub-categories AND sub-blocks at the same time pushes and glitches blocks in workspace.

I originally thought https://github.com/RaspberryPiFoundation/blockly/issues/9392 covered this, but it does not, so opening a new ticket.

The issue can also be observed in https://github.com/RaspberryPiFoundation/blockly/issues/9392, the main post video in seconds 8-9.

You can also see the problem in the video below (when I expand Math category):

Image

As you can see, the problem does not happen with categories with only-sub categories, e.g. "Block procedures" but does happen when opening "Math" category in toolbox that has "Vectors" sub-category and some blocks at the same time.

```xml
<xml id="toolbox">
    <category name="Demo1">
        <block type="coord_x"/>
        <block type="coord_y"/>
        <block type="coord_z"/>
		<category name="Vectors">
			<block type="coord_vector"/>
		</category>
    </category>
    <category name="Demo2">
        <block type="coord_x"/>
        <block type="coord_y"/>
        <block type="coord_z"/>
    </category>
    <category name="Demo3">
		<category name="Vectors 1">
			<block type="coord_vector"/>
		</category>
		<category name="Vectors 2">
			<block type="coord_vector"/>
		</category>
    </category>
</xml>
```

The problem in the "demo toolbox" above will only happen with Demo1 category, Demo2 and Demo3 work fine and don't shift blocks to the side.

TannerGabriel:
## The basics

### Resolves

Fixes #9992

### Proposed Changes

Only add the flyout width/height to the workspace origin if the flyout is not autoClose.

### DIFF
--- a/packages/blockly/core/toolbox/toolbox.ts
+++ b/packages/blockly/core/toolbox/toolbox.ts
@@ -722,13 +722,13 @@ export class Toolbox
       this.toolboxPosition === toolbox.Position.LEFT
         ? workspace.scrollX +
           rect.width +
-          (flyout?.isVisible() ? flyout.getWidth() : 0)
+          (flyout && !flyout.autoClose ? flyout.getWidth() : 0)
         : workspace.scrollX;
     const newY =
       this.toolboxPosition === toolbox.Position.TOP
         ? workspace.scrollY +
           rect.height +
-          (flyout?.isVisible() ? flyout.getHeight() : 0)
+          (flyout && !flyout.autoClose ? flyout.getHeight() : 0)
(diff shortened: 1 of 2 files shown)


mikeharv (approved):
Thanks for the fix, and for adding the regression test too!

Issue #9735, opened 2026-04-20
Stop importing node modules directly in Mocha webdriver tests


grega:
When completing the recent documentation work (https://github.com/RaspberryPiFoundation/blockly/pull/9671#discussion_r3087501107) it became apparent that the webdriver tests are directly importing from the `node_modules` dir:

https://github.com/RaspberryPiFoundation/blockly/blob/main/packages/blockly/tests/mocha/webdriver.js#L19-L22

The above is a patch for when multiple `node_modules` dirs exist (ie. one in root, and one in each of the `packages/**/` directories, but the core issue remains: packages should have their paths resolved by the Node environment directly and not be hardcoded.

maribethb:
this is also an issue in the playground files which load script src directly out of hardcoded node_modules paths.

PR #10208, opened 2026-07-29
test: Add support for running the Mocha test suite under Node

gonfunko:
## The basics

### Proposed Changes
This PR adds support for running the existing Mocha test suite under Node, via `npm run test:mocha:node`, as well as in the browser via the existing `npm run test:mocha:interactive`. This change was LLM-assisted. The vast majority of the test suite was unchanged; a single-digit number of tests that depended upon specific layout details or focus event behavior were moved to the browser test suite (`npm run test:browser`). Additionally, the tests now import sinon/chai/etc by Node module resolution, rather than by relative path to the node_modules folder. Reviewing commit-by-commit is recommended.

### Reason for Changes
Running the tests in the browser is slow, tricky in CI, and confusing to LLMs. It also means we have very poor test coverage of running Blockly headless/server-side, since before this the only test we had that ran under Node was a minimal smoketest of serialization. Now, 3956 testcases run and pass under both Node and within the browser. Several changes were made to achieve this:

* The test files were all updated to import Node modules by name, rather than relative path to the node_modules folder. These imports are resolved by Node when running under Node. Browsers do not implement Node's module resolution algorithm, so specifying a path remains necessary there; however, we now use an [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap) and imports to load the various Node modules when running in the browser.
* Some markup and test fixtures were factored out to more easily share a single source of truth between the Node and browser Mocha test harnesses.
* A few bugs in the tests were surfaced and fixed by this process.
* A single-digit number of test cases that depended on focus and layout specifics were moved out of the Mocha suite and into the browser suite, so that the full set of Mocha tests runs regardless of browser vs node runtime. The dragger-related tests were manually migrated, and the focus and dropdowndiv ones were LLM-migrated, with extensive review. 
* The old Node smoketest and tests for the colour field were removed.
* A harness for running tests under Node and a minimal mock for some mostly-SVG-related behavior that JSDom doesn't implement were added. These (node-setup.mjs and svg_layout_stubs.js) are the main LLM-generated pieces. I have reviewed them and cut down on the chattiness of the comments, although I left a fair bit where they were actually useful for understanding.

### Test Coverage
`npm run test:mocha:interactive`, `npm run test:mocha:node`, and `npm run test:browser` all pass, exercising the Mocha suite in the browser, in Node, and the browser-specific suite in the browser.


lizschwab (approved):

packages/blockly/scripts/gulpfiles/test_tasks.mjs
```diff
@@ -263,6 +264,22 @@ async function metadata() {
...
 }
 
+/**
+ * Generates tests/mocha/test-modules.generated.mjs,
+ * the list of tests imported by the browser harness (tests/mocha/index.html).
+ * Keeping it generated from a glob means new *_test.js files are picked up
+ * automatically and the browser and Node runners stay in sync.
+ * @return {Promise} Asynchronous result.
+ */
+function generateMochaIndex() {
```
maribethb, line 274: woot, keeping this up to date was a footgun

packages/blockly/tests/mocha/test_helpers/svg_layout_stubs.js
```diff
@@ -0,0 +1,334 @@
+/**
+ * @license
+ * Copyright 2026 Raspberry Pi Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Geometry stubs for running the Mocha suite under JSDom.
+ *
+ * JSDom implements the SVG DOM but not SVG layout: it provides no `getBBox`,
```
maribethb, line 10: There are a couple libraries that do this, e.g. https://www.npmjs.com/package/svgdom

which would just be a devDependency for the tests so we wouldn't have to maintain these mocks. I don't really want to maintain this kind of thing if I don't have to...

packages/blockly/tests/mocha/index.html
```diff
@@ -4,7 +4,22 @@
...
 
-    <link href="../../node_modules/mocha/mocha.css" rel="stylesheet" />
+    <!-- Resolve the bare/subpath specifiers used by the test modules so the
+         same import statements work in both the browser and under Node. The
+         "#core/" and "#generators/" prefixes mirror the "imports" field in
+         package.json (used by the Node runner). -->
+    <script type="importmap">
+      {
+        "imports": {
+          "chai": "../../node_modules/chai/index.js",
```
maribethb, line 14: Do we have to do this? This doesn't really fix #9735 as much as it moves the problem to one file instead of in every test file. This is still hardcoding the node_modules location instead of letting node do its job and find the package itself.

maribethb: I'm fine with doing this for now, but it doesn't fix either of the two issues linked from the issue description and I would prefer you not close them so we can look into actually fixing them in the future. because this approach requires us to keep the hack described in #9735

packages/blockly/tests/mocha/node-setup.mjs
```diff
@@ -0,0 +1,226 @@
...
+    useFakeTimers({...config, shouldAdvanceTime: true});
+  return sandbox;
+};
+
+// Install geometry stubs (getBBox/getCTM/getScreenCTM/createSVGPoint, canvas
+// text measurement) that JSDom does not implement.
+installSvgLayoutStubs(window);
+
+// Load Blockly, the standard blocks and the JavaScript generator, and expose
+// them as globals, mirroring what tests/mocha/index.html does in the browser.
```
maribethb, line 178: Half the test files also import Blockly themselves. Not sure if this creates the two-blockly problem. it hasn't bitten us yet?? but just flagging this here as probably something we should decide - should we import it globally or make each test import it themselves. I think this might be covered by https://github.com/RaspberryPiFoundation/blockly/issues/7224

anyway, this isn't a new problem, just also creating the same problem we have in browser as for node, so nothing to do here.

packages/blockly/tests/mocha/index.html
maribethb: I have done a little bit of research into this and have some ideas if you want to hear them we can chat over video. but they shouldn't block this PR.

maribethb (approved):
mostly informational comments, but imo i don't think either of the bugs you marked as fixed should be closed yet.

packages/blockly/tests/mocha/test_helpers/svg_layout_stubs.js
gonfunko: Will take a look at this for a followup!

packages/blockly/tests/mocha/node-setup.mjs
gonfunko: Noted, and yeah, I think I vaguely prefer each test importing it, but agreed that's tracked by the issue you linked.

PR #10212, opened 2026-07-30
test: Fix the browser tests

gonfunko:
## The basics

## The details
This PR fixes the browser tests now that the devtools have been removed from the playground. Two tests were using the test blocks version of the playground, which depends on the devtools, but for no actual reason; I modified them to use the normal toolbox. The other tests are those that drag out every block in the test block toolbox; I've skipped those for now, and we can either reenable them later or move them into a test suite that lives in devtools.


lizschwab (approved):

gonfunko:
Filed #10213 to track reenabling the skipped ones.

PR #10214, opened 2026-07-30
fix: keep tall field corners inside round caps with zelos

### LINKED ISSUE #2021: Bitmap field renders outside of block bounds in zelos
### Check for duplicates

### Component

field-bitmap

### Description

The bitmap field renders outside the bounds of its block if you're using the zelos renderer.

### Reproduction steps

1. Go to https://google.github.io/blockly-samples/plugins/field-bitmap/test/index.html#renderer=zelos&debugEnabled=false
2. Observe how the fields in the flyout are rendered outside of their blocks.

### Stack trace

```shell
N/A
```

### Screenshots

image

### Additional info

If the size of the field is being reported correctly, this may actually be a bug in core.

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly-samples/issues/2021

### Proposed Changes

For round blocks using the Zelos renderer, negative (inset) cap spacing now uses height-aware circle clearance instead of a fixed centerline-only pad. Tall rectangular fields (images, bitmaps) no longer clip through the rounded ends.

bitmap field after
image field after

This also widens an existing tight-nesting special case from `FieldTextInput` to `FieldInput`, so number fields get the same treatment as text inputs (about 2px narrower on simple number reporters).  This isn't related to the reported bug.

### Reason for Changes

With the Zelos renderer, round connections reserve a full rounded cap width on each side. Content is then pulled into those caps so it nests inside the curve rather than sitting in a rectangular box beside the cap.
If a tall rectangular field on the left/right edge of a pill-shaped block, the field’s corners stick out through the curved cap outline. This only happened when the field is tall enough relative to the cap radius (roughly >30). Normal short fields/icons looked fine. The old code treated every field like round centerline content, and under-padded tall rectangles inside round caps.

bitmap field before
image field before

Note that bug report was for bitmap fields specifically, but the bug applies to all potentially tall rectangular fields.

### Test Coverage

- Added `tests/mocha/zelos_info_test.js`. The test renders a 75×75 image on a single-field block and checks that the total horizontal pad is above the old centerline-only pad (24px) and below full-cap width (84px).
- Manually verified tall bitmap and image reporters in the Zelos playground (before/after screenshots above), plus a simple `math_number` block for the `FieldInput` nesting tweak (nearly identical, now 2px narrower). No changes observed on any other field types.

### DIFF
--- a/packages/blockly/tests/mocha/zelos_info_test.js
+++ b/packages/blockly/tests/mocha/zelos_info_test.js
@@ -22,7 +22,8 @@ suite('Zelos RenderInfo', function () {
           {
             'type': 'field_image',
             'name': 'IMG',
-            'src': 'https://blockly-demo.appspot.com/static/tests/media/a.png',
+            // Layout uses width/height only; src need not resolve.
+            'src': 'about:blank',
             'width': 75,
             'height': 75,
             'alt': 'A',
--- a/packages/blockly/tests/mocha/zelos_info_test.js
+++ b/packages/blockly/tests/mocha/zelos_info_test.js
@@ -22,8 +22,7 @@ suite('Zelos RenderInfo', function () {
           {
             'type': 'field_image',
             'name': 'IMG',
-            'src':
-              'https://blockly-demo.appspot.com/static/tests/media/a.png',
+            'src': 'https://blockly-demo.appspot.com/static/tests/media/a.png',
             'width': 75,
             'height': 75,
             'alt': 'A',
--- a/packages/blockly/tests/mocha/zelos_info_test.js
+++ b/packages/blockly/tests/mocha/zelos_info_test.js
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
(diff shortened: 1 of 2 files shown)


packages/blockly/tests/mocha/zelos_info_test.js
```diff
@@ -0,0 +1,55 @@
...
+    this.workspace = Blockly.inject('blocklyDiv', {renderer: 'zelos'});
+    Blockly.defineBlocksWithJsonArray([
+      {
+        'type': 'tall_round_reporter',
+        'message0': '%1',
+        'args0': [
+          {
+            'type': 'field_image',
+            'name': 'IMG',
+            'src': 'https://blockly-demo.appspot.com/static/tests/media/a.png',
```
gonfunko: Is the src necessary/does it need to be valid? If so, something not on the old App Engine site would be good. But would be ideal if the width/height are adequate and this can be about:blank or something.

gonfunko (approved):

gonfunko (approved):

PR #10217, opened 2026-07-31
fix: Fix some memory leaks

gonfunko:
## The basics

### Proposed Changes
This PR fixes some memory leaks. These were largely related to (a) not unregistering tooltip listeners or clearing the `.tooltip` property on DOM elements, which various Blockly-related classes were stuffed into, sometimes creating cycles and (b) missing clearing an element in the block flyout inflater.

### Reason for Changes
RAM is expensive, and cleaning up your own mess is nice.

### DIFF
--- a/packages/blockly/core/field.ts
+++ b/packages/blockly/core/field.ts
@@ -661,6 +661,13 @@ export abstract class Field<T = any>
       dom.removeNode(this.fieldGroup_);
     }
 
+    const clickTarget = this.getClickTarget_();
+    if (clickTarget) {
+      Tooltip.unbindMouseEvents(clickTarget);
+      delete (clickTarget as any).tooltip;
+    }
+    this.tooltip = null;
+
     this.disposed = true;
   }
 
--- a/packages/blockly/core/icons/icon.ts
+++ b/packages/blockly/core/icons/icon.ts
@@ -81,8 +81,12 @@ export abstract class Icon implements IIcon, IContextMenu {
   }
 
   dispose(): void {
-    tooltip.unbindMouseEvents(this.svgRoot);
-    dom.removeNode(this.svgRoot);
+    if (this.svgRoot) {
+      tooltip.unbindMouseEvents(this.svgRoot);
+      delete (this.svgRoot as any).tooltip;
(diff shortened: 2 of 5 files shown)


packages/blockly/core/field.ts
```diff
@@ -661,6 +661,13 @@ export abstract class Field<T = any>
       dom.removeNode(this.fieldGroup_);
     }
 
+    const clickTarget = this.getClickTarget_();
+    if (clickTarget) {
+      Tooltip.unbindMouseEvents(clickTarget);
+      delete (clickTarget as any).tooltip;
+    }
+    this.tooltip = null;
```
lizschwab, line 669: Is there a reason this sets the tooltip to null but the others set it to an empty string?

packages/blockly/core/icons/icon.ts
```diff
@@ -81,8 +81,12 @@ export abstract class Icon implements IIcon, IContextMenu {
   }
 
   dispose(): void {
-    tooltip.unbindMouseEvents(this.svgRoot);
-    dom.removeNode(this.svgRoot);
+    if (this.svgRoot) {
+      tooltip.unbindMouseEvents(this.svgRoot);
+      delete (this.svgRoot as any).tooltip;
```
lizschwab, line 86: For all of these delete (foo as any).tooltip, is it possible to throw? Would they ever be undefined?

lizschwab (approved):
Just a couple of questions. I don't think either is enough to hold back an approval, but take a look.

packages/blockly/core/field.ts
gonfunko: Yes, the tooltip field is nullable here, but not on the others. Either way will prevent cycles or leaks – I'd prefer to null it out everywhere, but don't want to change the API.

packages/blockly/core/icons/icon.ts
gonfunko: Icon and Field, which can be headless, both check for non-nullness of the thing on which .tooltip lives. BlockSvg doesn't, but it also will always have an SVG path. The delete keyword is documented as being a no-op if the property being deleted doesn't exist, so this should all be safe.

PR #10226, opened 2026-08-03
test: Use local media URL for tests

gonfunko:
## The basics

### Proposed Changes
This PR specifies a local media path for all Blockly injection sites across the tests. This reduces unnecessary network requests when tests run in a browser, and avoids making actual across-the-internet network requests to the default media URL when running tests. This in turn resolves the test failures in #10223. This change was LLM-generated, and I reviewed it to ensure the changes were correctly made (but a second pair of eyes is still appreciated!).

### DIFF
--- a/packages/blockly/tests/mocha/test_helpers/setup_teardown.js
+++ b/packages/blockly/tests/mocha/test_helpers/setup_teardown.js
@@ -23,7 +23,6 @@ const TEST_MEDIA_PATH = '../../media/';
  */
 export const DEFAULT_INJECT_OPTIONS = Object.freeze({
   media: TEST_MEDIA_PATH,
-  sounds: false,
 });
 
 /**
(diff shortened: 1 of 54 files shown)


packages/blockly/tests/mocha/test_helpers/setup_teardown.js
```diff
@@ -7,6 +7,25 @@
...
+/**
+ * Default options for Blockly.inject in tests. Media is loaded from the local
+ * checkout and sounds are disabled so that tests never fetch assets over the
+ * network. Spread this to add or override options, e.g.
+ *
+ *     Blockly.inject('blocklyDiv', {...DEFAULT_INJECT_OPTIONS, rtl: true});
+ */
+export const DEFAULT_INJECT_OPTIONS = Object.freeze({
+  media: TEST_MEDIA_PATH,
+  sounds: false,
```
maribethb: where did we land with this? i'd personally really like to have sounds in tests, especially since this only matters now when you intentionally run them in the browser and not just on every test run.

gonfunko: Seems like the media URL fix alone is sufficient, reverted the removal of sounds.

maribethb (approved):

PR #10231, opened 2026-08-05
chore: update version and publish workflow

### LINKED ISSUE #7765: Configure the release process

lizschwab:
## The basics

### Resolves 

Fixes #7765 

### Proposed Changes

Add Lerna to the monorepo and update the publish github action to use lerna for versioning and publishing.

### Reason for Changes

This is part of the ongoing monorepo project. Leveraging Lerna sets us up to be able to build and publish packages in the correct order.

### Test Coverage

No tests were added. I tested this locally by testing the version logic locally with the different variable combinations. Additional tests will be run prior to these changes being used for live publishing.


.github/workflows/publish.yml
```diff
@@ -54,146 +55,73 @@ jobs:
...
+              VERSION_COMMAND="${VERSION_COMMAND} --conventional-graduate"
+            fi
+          else
+            VERSION_COMMAND="${VERSION_COMMAND} --conventional-prerelease --preid beta"
           fi
-          VERSION=$(node -p "require('./package.json').version")
-          if [[ "${VERSION}" == *"-beta."* ]]; then
-            npm version prerelease --preid=beta --no-git-tag-version
+          VERSION_COMMAND="${VERSION_COMMAND} --conventional-commits --no-git-tag-version --yes"
+          if [ "${DRY_RUN}" = "true" ]; then
```
lizschwab, line 93: This is a change from what I showed you the other day. I moved dry run into here so that we're using the same version command that would be applied with the same inputs.

.github/workflows/publish.yml
```diff
@@ -54,146 +55,73 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+          VERSION_OVERRIDE: ${{ inputs.version_override }}
           DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
           REF_NAME: ${{ github.ref_name }}
-          RELEASE_TYPE: ${{ steps.bump.outputs.release_type }}
-          VERSION_OVERRIDE: ${{ inputs.version_override }}
+          DRY_RUN: ${{ inputs.dry_run}}
         run: |
-          set -euo pipefail
+          VERSION_COMMAND="npx lerna version"
```
lizschwab, line 80: I'm essentially doing a string builder and adding the appropriate flags depending on the inputs and which branch we're working from.

.github/workflows/publish.yml
```diff
@@ -54,146 +55,73 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SKIP_VERSIONING: ${{ inputs.skip_versioning }}
         run: |
-          DIST_TAG="${{ github.ref_name == github.event.repository.default_branch && 'latest' || 'beta' }}"
-          echo "Dry run: would publish version ${{ steps.version.outputs.version }} to npm dist-tag: ${DIST_TAG}"
-          if [ "${{ github.ref_name }}" != "${{ github.event.repository.default_branch }}" ]; then
-            echo "GitHub release would be created as prerelease."
+          if [ "${SKIP_VERSIONING}" = "true" ]; then
+            npx lerna publish from-package --yes --loglevel silly
```
lizschwab: I'll remove the --loglevel silly once we're happy with the results

.github/workflows/publish.yml
```diff
@@ -54,146 +55,73 @@ jobs:
...
-        run: npm pack ./dist
-
       - name: Create GitHub release
-        working-directory: packages/blockly
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          TARBALL="blockly-${{ needs.version.outputs.version }}.tgz"
+          VERSION=$(node -p "require('.packages/blockly/package.json').version")
+          TARBALL="blockly-*${VERSION}.tgz"
```
lizschwab: I'm not 100% sure on this. I did a test of packing locally and all of the file names matched this pattern, so I think this will work.

packages/blockly/package.json
```diff
@@ -23,6 +23,9 @@
     "#core/*": "./build/src/core/*",
     "#generators/*": "./build/src/generators/*"
   },
+  "publishConfig": {
```
lizschwab, line 26: This was the per-package configuration I was able to find for the publish command. https://github.com/lerna/lerna/blob/main/libs/commands/publish/README.md#per-package-configuration

.gitignore
```diff
@@ -11,6 +11,7 @@ build-debug.log
 /nbproject/private/
 tsdoc-metadata.json
 .vscode
+CHANGELOG.md
```
lizschwab: As far as I can tell from the Lerna documentation, there's no per-package configuration for version. I'm hoping adding the root level and blockly package changelog files to gitignore will prevent them from being published.

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
+            if [[ "${VERSION}" == *"-beta."* ]]; then
+              VERSION_COMMAND="${VERSION_COMMAND} --conventional-graduate"
+            fi
+          else
+            VERSION_COMMAND="${VERSION_COMMAND} --conventional-prerelease --preid beta"
           fi
-          VERSION=$(node -p "require('./package.json').version")
-          if [[ "${VERSION}" == *"-beta."* ]]; then
-            npm version prerelease --preid=beta --no-git-tag-version
+          VERSION_COMMAND="${VERSION_COMMAND} --conventional-commits --no-git-tag-version --yes"
```
maribethb: I think including `no-git-tag-version` is problematic here. Since you're using `lerna publish from-git` it's looking for the git tag corresponding to this release, but you've skipped creating it. I think we need to create it here so remove this flag. you might want it at the same place you have `--no-push` though.

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
-          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
-          echo "Version: $VERSION"
-
-      - name: Dry run summary
-        if: ${{ inputs.dry_run }}
+      - name: Publish
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SKIP_VERSIONING: ${{ inputs.skip_versioning }}
+          DIST_TAG: ${{ steps.dist.output.dist_tag }}
```
maribethb: I believe this needs to be `outputs` not `output` for the syntax to work https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
-
-      - name: Create tarball
-        working-directory: packages/blockly
-        run: npm pack ./dist
-
       - name: Create GitHub release
-        working-directory: packages/blockly
         env:
           GH_TOKEN: ${{ github.token }}
+          DIST_TAG: ${{ steps.dist.output.dist_tag }}
```
maribethb: same outputs typo here

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
-
-      - name: Publish to npm
-        working-directory: packages/blockly/dist
-        run: npm publish --tag "${NPM_DIST_TAG}" --verbose
-
-      - name: Create tarball
-        working-directory: packages/blockly
-        run: npm pack ./dist
-
       - name: Create GitHub release
```
maribethb, line 125: this step shouldn't run if dry-run is true

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
-        id: version
-        working-directory: packages/blockly
-        run: |
-          VERSION=$(node -p "require('./package.json').version")
-          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
-          echo "Version: $VERSION"
-
-      - name: Dry run summary
-        if: ${{ inputs.dry_run }}
+      - name: Publish
```
maribethb, line 108: this step also shouldn't run if dry run is true.

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
       - name: Create GitHub release
-        working-directory: packages/blockly
         env:
           GH_TOKEN: ${{ github.token }}
+          DIST_TAG: ${{ steps.dist.output.dist_tag }}
         run: |
-          TARBALL="blockly-${{ needs.version.outputs.version }}.tgz"
-          if [ "${{ github.ref_name }}" != "${{ github.event.repository.default_branch }}" ]; then
-            gh release create "blockly-v${{ needs.version.outputs.version }}" "$TARBALL" \
+          VERSION=$(node -p "require('.packages/blockly/package.json').version")
```
maribethb: probably is supposed to be `./packages` not `.packages` right?

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
+          TARBALL="blockly-*${VERSION}.tgz"
```
maribethb: this tarball isn't going to exist right now because you removed the "create tarball" step

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
       - name: Install dependencies
-        if: ${{ !inputs.skip_versioning }}
         run: npm ci
 
-      - name: Determine version bump
-        id: bump
-        if: ${{ !inputs.skip_versioning && inputs.version_override == '' }}
+      - name: Build core package
         working-directory: packages/blockly
+        run: npm run package
```
maribethb: This is when the publishable copy of `./packages/blockly/dist/package.json` gets copied / created out of the real one right?

If so then I think lerna would publish the wrong version to npm. because lerna is only going to update `./packages/blockly/package.json` in the version step, which happens after this, and therefore the one in `dist` which actually gets published would still be the old version.

you could test this locally by running `npm run package`, then the version command with --no-push, and see if `dist/package.json` gets updated but I really don't think lerna will do that since it's not until the publish step that it cares about the `dist` directory at all.

I think it should be safe to resolve this by just moving this step after the version step.

.github/workflows/publish.yml
```diff
@@ -33,19 +33,18 @@ permissions:
 jobs:
   ci:
     uses: ./.github/workflows/build.yml
-
-  version:
+  
+  publish:
```
maribethb, line 37: we should add the guard that we have in samples that doesn't bother to run any part of this if the owner isn't RaspberryPiFoundation

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
+          else
+            VERSION_COMMAND="${VERSION_COMMAND} --conventional-prerelease --preid beta"
           fi
-          VERSION=$(node -p "require('./package.json').version")
-          if [[ "${VERSION}" == *"-beta."* ]]; then
-            npm version prerelease --preid=beta --no-git-tag-version
+          VERSION_COMMAND="${VERSION_COMMAND} --conventional-commits --no-git-tag-version --yes"
+          if [ "${DRY_RUN}" = "true" ]; then
+            VERSION_COMMAND="${VERSION_COMMAND} --no-push"
+            RELEASE_VERSIONS=$(eval "$VERSION_COMMAND")
```
maribethb: I think instead of capturing this into a variable and then printing the variable one line later it might be cleaner to just eval the command? i worry capturing the output as a variable might do something weird to the formatting and there's probably going to be a lot of info since we have log-level silly on at the moment.

.github/workflows/publish.yml
```diff
@@ -54,146 +53,84 @@ jobs:
...
         env:
-          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
-          REF_NAME: ${{ github.ref_name }}
-          RELEASE_TYPE: ${{ steps.bump.outputs.release_type }}
+          GH_TOKEN: ${{ github.token }}
           VERSION_OVERRIDE: ${{ inputs.version_override }}
+          DRY_RUN: ${{ inputs.dry_run}}
+          DIST_TAG: ${{ steps.dist.outputs.dist_tag }}
         run: |
-          set -euo pipefail
```
maribethb, line 78: I think you want to keep this to make sure we can see error output and that an error in lerna causes the step to fail (my bash is kinda rusty tho)

.gitignore
maribethb: makes sense, my only question is confirming this doesn't gitignore all the CHANGELOG.md files, I am pretty sure it would. should this be `/CHANGELOG.md` so it's just the root level one?

lizschwab: Good catch

.github/workflows/publish.yml
lizschwab: yeah, I wasn't 100% sure we still wanted that. I'll update it.

.github/workflows/publish.yml
lizschwab: you are correct, good catch!

.github/workflows/publish.yml
lizschwab: The publish step should create the tarball. I found that info here: https://github.com/lerna/lerna/tree/main/libs/commands/publish#lifecycle-scripts

.github/workflows/publish.yml
lizschwab: yeah, good catch!

.github/workflows/publish.yml
maribethb: i'm pretty sure this will be in a temp directory though and not exist at the package root where you're looking for it here.

.github/workflows/publish.yml
lizschwab: I took away the log-level silly on version, but that makes sense. I can just have it eval instead of storing in a variable. I tested it locally and it works, but the output is a little different. When we store it in the variable we get this (ignore the lack of version number change; there are no changes in my local):
```
Dry run: would publish the following versions to npm dist-tag: latest

Changes:
 - blockly: 13.3.0 => 13.3.0
 - @blockly/block-dynamic-connection: 13.3.0 => 13.3.0
 - @blockly/block-plus-minus: 13.3.0 => 13.3.0
 - @blockly/block-shareable-procedures: 13.3.0 => 13.3.0
 - @blockly/block-test: 13.3.0 => 13.3.0
 - @blockly/workspace-content-highlight: 13.3.0 => 13.3.0
 - @blockly/continuous-toolbox: 13.3.0 => 13.3.0
 - @blockly/plugin-cross-tab-copy-paste: 13.3.0 => 13.3.0
 - @blockly/create-package: 13.3.0 => 13.3.0
 - @blockly/dev-scripts: 13.3.0 => 13.3.0
 - @blockly/dev-tools: 13.3.0 => 13.3.0
 - @blockly/disable-top-blocks: 13.3.0 => 13.3.0
 - @blockly/field-angle: 13.3.0 => 13.3.0
 - @blockly/field-bitmap: 13.3.0 => 13.3.0
 - @blockly/field-colour: 13.3.0 => 13.3.0
 - @blockly/field-colour-hsv-sliders: 13.3.0 => 13.3.0
 - @blockly/field-date: 13.3.0 => 13.3.0
 - @blockly/field-dependent-dropdown: 13.3.0 => 13.3.0
 - @blockly/field-grid-dropdown: 13.3.0 => 13.3.0
 - @blockly/field-multilineinput: 13.3.0 => 13.3.0
 - @blockly/field-slider: 13.3.0 => 13.3.0
 - @blockly/fixed-edges: 13.3.0 => 13.3.0
 - @blockly/migrate: 13.3.0 => 13.3.0
 - @blockly/plugin-modal: 13.3.0 => 13.3.0
 - sample-app: 13.3.0 => 13.3.0 (private)
 - sample-app-typescript: 13.3.0 => 13.3.0 (private)
 - @blockly/plugin-scroll-options: 13.3.0 => 13.3.0
 - @blockly/shadow-block-converter: 13.3.0 => 13.3.0
 - @blockly/plugin-strict-connection-checker: 13.3.0 => 13.3.0
 - @blockly/suggested-blocks: 13.3.0 => 13.3.0
 - @blockly/theme-dark: 13.3.0 => 13.3.0
 - @blockly/theme-deuteranopia: 13.3.0 => 13.3.0
 - @blockly/theme-hackermode: 13.3.0 => 13.3.0 (private)
 - @blockly/theme-highcontrast: 13.3.0 => 13.3.0
 - @blockly/theme-modern: 13.3.0 => 13.3.0
 - @blockly/theme-tritanopia: 13.3.0 => 13.3.0
 - @blockly/toolbox-search: 13.3.0 => 13.3.0
 - @blockly/plugin-typed-variable-modal: 13.3.0 => 13.3.0
 - @blockly/workspace-backpack: 13.3.0 => 13.3.0
 - @blockly/workspace-minimap: 13.3.0 => 13.3.0
 - @blockly/plugin-workspace-search: 13.3.0 => 13.3.0
 - @blockly/zoom-to-fit: 13.3.0 => 13.3.0

```
If we eval instead we get this:
```
Dry run: would publish the following versions to npm dist-tag: latest
lerna notice cli v9.0.7
lerna info current version 13.3.0
lerna info Assuming all packages changed
lerna WARN version Skipping working tree validation, proceed at your own risk

Changes:
 - blockly: 13.3.0 => 13.3.0
 - @blockly/block-dynamic-connection: 13.3.0 => 13.3.0
 - @blockly/block-plus-minus: 13.3.0 => 13.3.0
 - @blockly/block-shareable-procedures: 13.3.0 => 13.3.0
 - @blockly/block-test: 13.3.0 => 13.3.0
 - @blockly/workspace-content-highlight: 13.3.0 => 13.3.0
 - @blockly/continuous-toolbox: 13.3.0 => 13.3.0
 - @blockly/plugin-cross-tab-copy-paste: 13.3.0 => 13.3.0
 - @blockly/create-package: 13.3.0 => 13.3.0
 - @blockly/dev-scripts: 13.3.0 => 13.3.0
 - @blockly/dev-tools: 13.3.0 => 13.3.0
 - @blockly/disable-top-blocks: 13.3.0 => 13.3.0
 - @blockly/field-angle: 13.3.0 => 13.3.0
 - @blockly/field-bitmap: 13.3.0 => 13.3.0
 - @blockly/field-colour: 13.3.0 => 13.3.0
 - @blockly/field-colour-hsv-sliders: 13.3.0 => 13.3.0
 - @blockly/field-date: 13.3.0 => 13.3.0
 - @blockly/field-dependent-dropdown: 13.3.0 => 13.3.0
 - @blockly/field-grid-dropdown: 13.3.0 => 13.3.0
 - @blockly/field-multilineinput: 13.3.0 => 13.3.0
 - @blockly/field-slider: 13.3.0 => 13.3.0
 - @blockly/fixed-edges: 13.3.0 => 13.3.0
 - @blockly/migrate: 13.3.0 => 13.3.0
 - @blockly/plugin-modal: 13.3.0 => 13.3.0
 - sample-app: 13.3.0 => 13.3.0 (private)
 - sample-app-typescript: 13.3.0 => 13.3.0 (private)
 - @blockly/plugin-scroll-options: 13.3.0 => 13.3.0
 - @blockly/shadow-block-converter: 13.3.0 => 13.3.0
 - @blockly/plugin-strict-connection-checker: 13.3.0 => 13.3.0
 - @blockly/suggested-blocks: 13.3.0 => 13.3.0
 - @blockly/theme-dark: 13.3.0 => 13.3.0
 - @blockly/theme-deuteranopia: 13.3.0 => 13.3.0
 - @blockly/theme-hackermode: 13.3.0 => 13.3.0 (private)
 - @blockly/theme-highcontrast: 13.3.0 => 13.3.0
 - @blockly/theme-modern: 13.3.0 => 13.3.0
 - @blockly/theme-tritanopia: 13.3.0 => 13.3.0
 - @blockly/toolbox-search: 13.3.0 => 13.3.0
 - @blockly/plugin-typed-variable-modal: 13.3.0 => 13.3.0
 - @blockly/workspace-backpack: 13.3.0 => 13.3.0
 - @blockly/workspace-minimap: 13.3.0 => 13.3.0
 - @blockly/plugin-workspace-search: 13.3.0 => 13.3.0
 - @blockly/zoom-to-fit: 13.3.0 => 13.3.0

lerna info auto-confirmed 
lerna info execute Skipping git tag/commit
lerna info execute Skipping git push
lerna info execute Skipping releases
lerna info getChangelogConfig Auto-prefixing conventional-changelog preset "conventionalcommits"
lerna info getChangelogConfig Successfully resolved preset "conventional-changelog-conventionalcommits"
lerna success version finished
```
So storing it is a bit cleaner but it's not harmful to have the extra logging in there.

.github/workflows/publish.yml
```diff
@@ -33,19 +33,20 @@ permissions:
...
-
-  version:
+  
+  publish:
     needs: ci
     runs-on: ubuntu-latest
-    timeout-minutes: 30
-    outputs:
-      version: ${{ steps.version.outputs.version }}
+    # Don't try to publish from a fork of RaspberryPiFoundation/blockly-samples.
```
maribethb: nit: s/blockly-samples/blockly

.github/workflows/publish.yml
maribethb: I think you just need to add back the `npm pack` step we used to have (just make sure you pack the correct directory) to create the tarball. Having published using lerna a lot on the command line I'm pretty confident that the tarball does not automatically exist at this location after the publish is complete, otherwise i would have had to be deleting them from my working tree all the time.

.github/workflows/publish.yml
maribethb: ok one more nit, is the star here doing anything? i think it should just be "blockly-${VERSION}.tgz". it's just one file that we're uploading. otherwise, lgtm

maribethb (approved):

PR #10257, opened 2026-08-11
fix: wrong css prop

SimonShiki:
## The basics

### Resolves

None

### Proposed Changes
Correct wrong css prop name


lizschwab (approved):
Good catch! Thanks for submitting this.

PR #10268, opened 2026-08-12
fix: Clamp transforms to sensible levels of precision

gonfunko:
## The basics

### Proposed Changes
This PR clamps text width and workspace translation to integer values, and workspace scale to 3 decimals. Previously these were unbounded, typically leading to 14 digits of decimal precision, which is obviously unnecessary, and more computationally intensive.

### Test Coverage
All tests pass, and I visually inspected the blocks and zooming.


mikeharv (approved):
It might be worth calling out that this could potentially result in tiny visual differences, especially for applications that rely upon reference images for automated regression testing.

PR #10273, opened 2026-08-14
feat: Relax access restrictions on `BlockDragStrategy`

gonfunko:
## The basics

### Proposed Changes
This PR switches a number of fields and methods in `BlockDragStrategy` from `private` to `protected`.

### Reason for Changes
`BlockDragStrategy` can reasonably be expected to be subclassed, and the existing access restrictions made this challenging. The members that have been changed are those that I'm reasonably confident are stable, would likely be needed by subclasses, and have not-hideous APIs. These changes would ease or enable implementation of several drag strategies in Scratch, for e.g. duplicating blocks on drag, using a parent block as a drag target instead of the clicked block, and always using a new ID when dragging blocks out of the flyout to the main workspace.

### DIFF
--- a/packages/blockly/core/dragging/block_drag_strategy.ts
+++ b/packages/blockly/core/dragging/block_drag_strategy.ts
@@ -64,7 +64,7 @@ enum MoveMode {
 }
 
 export class BlockDragStrategy implements IDragStrategy {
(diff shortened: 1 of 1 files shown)


mikeharv (approved):
Seems very reasonable!

PR #10274, opened 2026-08-14
chore: added nx and updated scripts to use nx

### LINKED ISSUE #10233: Configure Nx
### Check for duplicates

### Problem

We're currently relying on the workspaces array order in the monorepo root package.json to determine build order. Nx can do this for us if we configure it properly.

### Request

This is part of the [monorepo project](https://github.com/google/blockly/issues/7755).

Configure Nx so that it handles the build order appropriately and update the npm workspaces list to ["packages/blockly", "packages/plugins/*"] rather than explicitly listing the packages in build order. We should probably be able to remove the package step from the build command in the root package.json as well once Nx is in place.

Test to make sure everything works correctly when running ci/build/test/lint/format.

### Alternatives considered

_No response_

### Additional context

_No response_

lizschwab:
## The basics

### Resolves

Fixes [10233](https://github.com/RaspberryPiFoundation/blockly/issues/10233)

### Proposed Changes

- Adds Nx to the monorepo
- updates build and test scripts at the root to use nx to allow for paralell build and test
- adds a clean script to the root
- splits build out into 3 separate scripts, defaulting to building everything but docs (we do not need to build docs for publish)
- leverages Nx to automate the steps needed to build/start/test in blockly and dev-tools
- adds a new`build:advanced` script to blockly that launches the advanced playground from dev-tools while watching for changes to blockly (in the future we plan to move this back into the playground as a link the way it used to be)

### Reason for Changes

- allows for paralell test and build runs
- removes unneeded docs build from publish workflow
- makes starting and building packages easier


packages/blockly/scripts/gulpfiles/package_tasks.mjs
```diff
@@ -175,6 +176,7 @@ function packageJSON(done) {
   const json = JSON.parse(JSON.stringify(getPackageJson()));
   // Remove unwanted entries.
   delete json['scripts'];
+  delete json['nx'];
```
lizschwab, line 179: Probably not necessary, but I figured if we're removing scripts from the `/dist` copy, we should remove nx, too.

packages/blockly/package.json
```diff
@@ -26,6 +26,29 @@
   "publishConfig": {
     "directory": "dist"
   },
+  "nx": {
+    "implicitDependencies": ["!@blockly/block-test"],
```
lizschwab, line 30: This prevents the circular dependency with block-test from preventing nx from determining build order. We don't need it for build, just test.

packages/blockly/package.json
```diff
@@ -44,6 +67,7 @@
     "package": "gulp pack",
     "prepareDemos": "gulp prepareDemos",
     "start": "npm run build && concurrently -n tsc,server \"tsc --watch --preserveWatchOutput --outDir \"build/src\" --declarationDir \"build/declarations\"\" \"http-server ./ -s -o /tests/playground.html -c-1\"",
+    "start:advanced": "npm run package && npx nx run @blockly/dev-tools:build && npm run start --workspace=@blockly/dev-tools && npx nx watch --projects=@blockly/dev-tools --includeDependencies -- nx run-many -t build -p \\$NX_PROJECT_NAME --exclude=@blockly/dev-tools",
```
lizschwab, line 70: We'll revisit adding this back in as a link on the regular playground later.

package.json
```diff
@@ -76,17 +36,21 @@
     "eslint-plugin-prettier": "^5.5.6",
     "globals": "^17.6.0",
     "lerna": "^9.0.7",
+    "nx": "23.1.0",
     "prettier": "3.9.5",
     "prettier-plugin-organize-imports": "^4.3.0"
   },
   "scripts": {
-    "test": "npm run test --ws --if-present",
+    "test": "npx nx run-many -t test --exclude=blockly && npx nx run blockly:test",
```
lizschwab, line 44: If we run blockly tests in paralell with other tests, they fail to find targets. I'm guessing this has to do with the gulp tasks as it's when it's running the gulp tasks prior to launching the browser that these errors happen. For now, we can run everything else in paralell and run blockly tests by themselves. In the future we can look at how to run these all in paralell.

Default for nx is to run 3 processes in paralell. We can adjust that if we want but I figure start with the default and see how CI runners handle that.

maribethb (approved):

PR #10285, opened 2026-08-18
fix: Fix bug that could cause dragged blocks to connect to and delete others

### LINKED ISSUE #10281: Dropping blocks into the flyout can delete the blocks behind it
### Check for duplicates

### Description

This is https://github.com/scratchfoundation/scratch-blocks/issues/3654 but it also happens in Blockly

When dragging a block into the flyout, it can connect to the blocks behind it and cause them to also be deleted. I would expect that only the dragged block is deleted

A similar bug can also happen when dragging blocks to the trash can icon at the bottom right corner, if something is behind the trash can

### Reproduction steps

1. Go to the [Blockly Playground](https://raspberrypifoundation.github.io/blockly/packages/blockly/tests/playground.html)
2. Place some blocks close to the flyout
3. Place some blocks far away from the flyout
4. Scroll the workspace so that blocks go behind the flyout
5. Drag a block to the top of a block behind the flyout
6. See that not only the block you dragged gets deleted, but also the block(s) that it connects to

### Priority

Work effort: idk

Workarounds:
- if the bug already happened: undo (the problem with this is that it doesn't work if you notice it too late)
- to prevent the bug: only use the keyboard and the context menu for deleting blocks

### Stack trace

```shell

```

### Screenshots

https://github.com/user-attachments/assets/702f048b-749b-4555-924d-972aecd0a1cf

### Browsers

Chrome desktop, Android mobile

gonfunko:
## The basics

### Resolves

Fixes #10281

### Proposed Changes
This PR fixes a bug where dragged blocks that were dropped onto a delete area could connect to other blocks on the workspace just before they were deleted, causing the non-dragged blocks to be deleted as well.

### Test Coverage

I added a browser test to exercise this scenario, and verified that it failed without the change in place.

### DIFF
--- a/packages/blockly/tests/browser/test/dragger_test.mjs
+++ b/packages/blockly/tests/browser/test/dragger_test.mjs
@@ -391,7 +391,6 @@ suite('Dragging into a delete area', function () {
 
         // Move the existing block near the trash so that the dragged block will
         // attempt to connect on top of it.
-        existingBlock.moveTo(new Blockly.utils.Coordinate(0, 0));
         existingBlock.moveTo(
           new Blockly.utils.Coordinate(
             deleteRectCenter.x - workspace.toolbox.getClientRect().right - 30,
--- a/packages/blockly/tests/browser/test/dragger_test.mjs
+++ b/packages/blockly/tests/browser/test/dragger_test.mjs
@@ -372,4 +372,51 @@ suite('Dragging into a delete area', function () {
       });
     });
   });
+
+  test('does not delete blocks already on the workspace within the connection radius of a dragged block dropped on a delete area', async function () {
+    const {blockHasDeleteStyle, blockIsDeadOrDying, existingBlockDeleted} =
+      await this.browser.execute(() => {
+        const workspace = Blockly.getMainWorkspace();
+
+        const existingBlock = workspace.newBlock('controls_repeat');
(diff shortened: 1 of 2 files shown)


packages/blockly/tests/browser/test/dragger_test.mjs
```diff
@@ -372,4 +372,51 @@ suite('Dragging into a delete area', function () {
...
+        const draggingBlock = workspace.newBlock('controls_if');
+        draggingBlock.initSvg();
+        draggingBlock.render();
+
+        const deleteRect = workspace.trashcan.getClientRect();
+        const deleteRectCenter = rectCenterClient(deleteRect);
+
+        // Move the existing block near the trash so that the dragged block will
+        // attempt to connect on top of it.
+        existingBlock.moveTo(new Blockly.utils.Coordinate(0, 0));
```
mikeharv: Is this move necessary (maybe to set the default transform)? If so, could we add a comment about why it's here?

mikeharv (approved):

packages/blockly/tests/browser/test/dragger_test.mjs
gonfunko: No, just leftover from debugging. Removed this line and confirmed the test still fails without the change and passes with it.

PR #10309, opened 2026-08-21
test: Refactor mocha test infrastructure

### LINKED ISSUE #5468: Look into using imports for mocha tests
**Background**
In #5467 we moved to getting our mocha and chai dependencies from the node_modules folder instead of unpkg. This means that we can more easily control what version of our dependencies we are using through the `package.json` file. This works, but we are now wondering if the better way to do this is to use import statements for our dependencies instead of script tags. 

I did a little bit of research on this and it looks like the [official example](https://mochajs.org/#running-mocha-in-the-browser) for testing in the browser uses script tags. 

I think the open questions here are: 
- Does using es6 imports over a script tag provide us any benefit?
- How do es6 imports and goog_module work together? Or can they work together?

### LINKED ISSUE #9735: Stop importing node modules directly in Mocha webdriver tests

gonfunko:
## The basics

### Resolves

* Fixes #5468
* Fixes #9735

### Proposed Changes
This PR refactors the Mocha test infrastructure. This change was LLM assisted with extensive manual review and edits. Reviewing commit by commit is recommended.

At a high level, this PR makes CI run the Mocha tests under node, and makes `npm run test:mocha:interactive` more convenient for day to day use by devs. More specifically:

* A bunch of tooling that served to communicate tests results back from the browser to the CLI has been removed; when tests are run under Node in CI, that all just gets outputted directly, and now that the interactive in-browser version is meant for humans, and doesn't need to copy the test results from the browser to the shell.
* Removing that infrastructure removes the need for Webdriver's involvement in the in-browser tests.
* When running the Mocha tests under Node, passing test output is suppressed and only failures are reported.
* The in-browser test page now opens in your default browser, and the full test suite passes under Chrome, Firefox and Safari.
* Edits to tests and Blockly's source both hot-reload the in-browser test suite.
* Tests are found dynamically and bundled with sinon, chai, and locally-built Blockly, so dependencies are resolved via node rather than hardcoded paths to node_modules. This bundling/compilation step also lays the groundwork for writing core tests in Typescript. This did require a new dev dependency on esbuild, but (a) that has no transitive dependencies itself and (b) is very widely used, so I think it's reasonable.
* Various dead code and a gulp-related NPM package have been removed, as they existed in test_tasks.js only to support an old test that measured Blockly's build size and was removed some time ago.

Mocha is copied from its resolved path in node_modules into the build/test directory and loaded from there. Mocha distributes pre-built browser versions, and their [own guidance](https://mochajs.org/running/browsers/) indicates that these should be used. Because the path is resolved, we don't hardcode node_modules anywhere. 

### Reason for Changes
This improves the speed and reliability of CI, since it no longer runs the test in a browser, improves developer experience when running tests with hot reload, improves reliability of tests (once we start moving them to Typescript), and removes a lot of complex code and hacks (importing modules by path).

### Test Coverage
Tests continue to pass, both in the browser and in Node.


packages/blockly/scripts/prepare_mocha_bundle.mjs
```diff
@@ -0,0 +1,43 @@
...
+  );
+}
+
+await fs.mkdir(OUT_DIR, {recursive: true});
+// Resolve and copy Mocha's prebuilt browser distribution.
+await fs.copyFile(
+  require.resolve('mocha/mocha.js'),
+  path.join(OUT_DIR, 'mocha.js'),
+);
+// Write out the esbuild entrypoint that imports the test files.
```
maribethb, line 42: I'm not gonna lie, I can't say I love the fact that we are adding a *third* compiler to our stack... but at least this one seems relatively lightweight compared to the webpack dance we're doing with testing samples.

maribethb (approved):
I have some slight hesitations about this but the end result is probably worth it and overall a net simplification.

packages/blockly/scripts/prepare_mocha_bundle.mjs
gonfunko: Yeah agreed, I may investigate whether switching to esbuild from webpack is viable/beneficial.

PR #10316, opened 2026-08-24
feat(workspace-search): Add screen reader support

### LINKED ISSUE #10324: Add screen reader support to the workspace search plugin
### Check for duplicates

### Component

WorkspaceSearch

### Problem

This plugin has keyboard navigation, but doesn't have screen reader support yet. 

### Request

Add support for screen readers that describes:

- How workspace search works (e.g., it is currently esc to focus the found block and this is not obvious)
- Announce when matching blocks are found (perhaps we have to announce the currently matching block - unclear)
- Is focusing the found block when the user presses Esc enough to understand that the workspace search input has closed and focus has now moved to the block you searched for?

### Alternatives considered

_No response_

### Additional context

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10324

### Proposed Changes

Add screen reader support to `@blockly/plugin-workspace-search`:

- Marked the search bar as a search landmark and use a search input
- Localized `aria-label`s on the input (including instructions around keyboard usage) and buttons
- Live-region announcements for the current match and for no matches

Initial:
image
After typing search term (matches found):
image
Pressing Enter to cycle matches:
image
Pressing Escape to focus matched block:
image
No matching blocks:
image

Core changes:
- Added `Role.SEARCH` to Aria utils and `WORKSPACE_SEARCH_*` messages

### Test Coverage

Unit tests cover attributes, names, and announcements.

### Documentation

Updated the plugin readme to add how Escape changes focus.

### DIFF
--- a/packages/plugins/workspace-search/src/workspace_search.ts
+++ b/packages/plugins/workspace-search/src/workspace_search.ts
@@ -35,7 +35,7 @@ export class WorkspaceSearch implements Blockly.IPositionable {
   /**
    * The placeholder text for the search bar input.
    */
-  private textInputPlaceholder = 'Search';
+  private textInputPlaceholder = Blockly.Msg['WORKSPACE_SEARCH_PLACEHOLDER'];
 
   /**
    * A list of blocks that came up in the search.
@@ -148,6 +148,7 @@ export class WorkspaceSearch implements Blockly.IPositionable {
 
     this.htmlDiv = document.createElement('div');
     Blockly.utils.dom.addClass(this.htmlDiv, 'blockly-ws-search');
+    Blockly.utils.aria.setRole(this.htmlDiv, Blockly.utils.aria.Role.SEARCH);
 
     const searchContainer = document.createElement('div');
     Blockly.utils.dom.addClass(searchContainer, 'blockly-ws-search-container');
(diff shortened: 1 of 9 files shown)


packages/plugins/workspace-search/src/workspace_search.ts
```diff
@@ -447,18 +463,20 @@ export class WorkspaceSearch implements Blockly.IPositionable {
...
    */
   open() {
     this.setVisible(true);
-    this.inputElement?.focus();
-    this.inputElement?.select();
-    if (this.searchText) {
-      this.searchAndHighlight(this.searchText);
+    if (this.inputElement) {
+      this.inputElement.value = this.searchText;
+      this.inputElement.focus();
```
maribethb, line 476: I wonder if you had this take ephemeral focus, could you then update the passive focus in the workspace when the matches are navigated? That would let you get the focus outlines on the matched block. You'd probably still want the live region announcements because a lot of the shortcuts to read blocks don't work when ephemeral focus is taken, but it would probably improve the experience for sighted users

maribethb (approved):

packages/plugins/workspace-search/src/workspace_search.ts
mikeharv: I'm going to address this and some other improvements in a follow-up PR.

mikeharv: https://github.com/RaspberryPiFoundation/blockly/pull/10391

PR #10320, opened 2026-08-25
test: Reenable a variety of skipped tests

gonfunko:
## The basics

### Proposed Changes
This PR removes `.skip()` from various tests. These are some combination of (a) tests that were passing fine but were just disabled (b) placeholder/dummy tests that were removed, (c) tests that needed some minor tweaks and (d) a suite of workspace comment tests written against a very old API that were nonetheless reasonable, and which I updated to exercise the otherwise-untested headless workspace comments.

### Test Coverage
Is better now.


packages/blockly/tests/mocha/blocks/procedures_test.js
```diff
@@ -615,12 +615,6 @@ suite('Procedures', function () {
       );
     });
 
-    test.skip(
-      'renaming a variable such that you get a parameter ' +
```
maribethb, line 619: fair enough to remove this but we probably should have a test for this behavior assuming we ever decided what the behavior should be lol, might be worth siccing the llm on it

packages/blockly/tests/mocha/rendered_workspace_comment_test.js
```diff
@@ -0,0 +1,198 @@
+/**
+ * @license
+ * Copyright 2024 Google LLC
```
maribethb, line 3: new file should get new header

packages/blockly/tests/mocha/workspace_comment_test.js
```diff
@@ -1,197 +1,209 @@
 /**
  * @license
- * Copyright 2024 Google LLC
+ * Copyright 2020 Google LLC
```
maribethb, line 3: this is a confusing change?

maribethb (approved):

packages/blockly/tests/mocha/rendered_workspace_comment_test.js
gonfunko: It's not a new file, just renamed – previously there was "old_workspace_comment_test" and "workspace_comment_test"; old was renamed to workspace_comment_test since it exercises the headless regular workspace comments, and workspace_comment_test was renamed rendered_workspace_comment_test, since that's what it was exercising. That's also the cause of the copyright year change the diff shows in the other file.

packages/blockly/tests/mocha/workspace_comment_test.js
gonfunko: Per the other comment it's an artifact of renaming, not a real change

packages/blockly/tests/mocha/blocks/procedures_test.js
gonfunko: Yeah, I think the deciding what should happen is the issue. Presently it just does nothing (the variable name goes through, and becomes the same variable as the param) which I think is fine?

PR #10391, opened 2026-08-31
feat(workspace-search): UI Improvements

### LINKED ISSUE #10402: workspace-search UI Improvements
### Check for duplicates

### Problem

Follow-up to https://github.com/RaspberryPiFoundation/blockly/issues/10324

The current UI is hard to parse. Blocks matching the search query are black and the current match is grey. This screen shot shows highlighting the second match for "if":

Image

### Request

Maribeth suggested having the search input element take ephemeral focus so that the matched blocks could have focus outlines, making them more obvious to sighted users. See: https://github.com/RaspberryPiFoundation/blockly/pull/10316#discussion_r3881936456

Additionally, it is worth exploring other improvements to the UI. Inspiration could be taken from other search UI that indicates multiple visual matches simultaneously.

Image

### Alternatives considered

_No response_

### Additional context

_No response_

mikeharv:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10402

#### Core
`BlockDragStrategy`:
- `getTargetBlock()` will position new blocks before setting them to dragging. This ensures that we can detect potential connection with blocks dragged from the flyout after a workspace scroll.

`WorkspaceSvg`:
- `onTreeBlur()` will correctly close the flyout workspace if ephemeral focus is taken (e.g. by the search bar) if the user is going to the target workspace. Previously ephemeral focus always kept the flyout open because it was generally expected to be temporary overlay UI like `WidgetDiv`, `DropdownDiv` or `Dialog`.

#### `WorkspaceSearch`
- Search now takes ephemeral focus on the input while the bar is open. The callback from `takeEphemeralFocus` is stored and called when that overlay goes away. This approach is basically the same as how we do it in `WidgetDiv`, `DropdownDiv`, and `Dialog`, except that we need to use an instance property in order to potentially support multiple workspaces.
- Now implements `IAutoHideable`. This is necessary in order to dismiss the search bar when the user does something like a click a block or workspace.
- `searchAndHighlight` now bookmarks the current match by block id, not by the object in `this.blocks`. During `close()`, that list is empty, so it falls back to `lastHighlighted`. On re-`open()`, it looks the block up and keeps that match if it still matches. `open()` now passes `this.preserveSelected` (always true unless overridden) so that path actually runs.
- New SVG styling methods were created to toggle classes on/off. A new `getNonMatchingBlocks()` is used to get the inverse of `this.blocks` so that we can dim them.

##### CSS changes
The screenshots below were all captured after searching for "if".
image

|  | Before | After |
|-|-|-|
| `blockly-ws-search-highlight`<br>All blocks that match the search query | image<br>`fill: #000;` | image<br>No styling (Class remains for apps to potentially continue overriding)|
| `blockly-ws-search-highlight`<br>The current match | image<br>`fill: grey;` | image<br>`filter: saturate(3) drop-shadow();`|
| `blockly-ws-search-greyed`<br>Non-matching | image<br>N/A (New class) | image<br>`fill: lightgrey; stroke: dimgrey;`|
| Practical example with many blocks. Current match is the top-most inner `if true do` block. | image | image|

### Reason for Changes

The cur

### Test Coverage
- Searching a term with multiple matches highlights them, marks the first as current, and does not grey matches.
- A search with no matches greys non-matching blocks.
- Cycling moves the current match to the next result.
- Closing focuses the matched block, or the workspace if there is no match. Closing after a miss still focuses the last found block.
- Reopening after changing focus restores the search query and the previous match, and keeps the cursor in the search field.
- If the previous match was deleted, reopen falls back to the first remaining match.

### Documentation

README updated.

### DIFF
--- a/packages/plugins/workspace-search/src/workspace_search.ts
+++ b/packages/plugins/workspace-search/src/workspace_search.ts
@@ -554,7 +554,7 @@ export class WorkspaceSearch
     }
   }
 
-  /** 
+  /**
    * Checks whether the search bar is currently visible.
    *
    * @returns true if the search bar is visible, false otherwise.
--- a/packages/plugins/workspace-search/src/workspace_search.ts
+++ b/packages/plugins/workspace-search/src/workspace_search.ts
@@ -11,7 +11,9 @@ import {injectSearchCss} from './css';
 /**
  * Class for workspace search.
  */
-export class WorkspaceSearch implements Blockly.IPositionable {
+export class WorkspaceSearch
+  implements Blockly.IPositionable, Blockly.IAutoHideable
+{
   /**
    * The unique id for this component.
    */
@@ -50,6 +52,11 @@ export class WorkspaceSearch implements Blockly.IPositionable {
    * the scroll position.
    */
   private lastHighlighted: Blockly.BlockSvg | null = null;
+
+  /**
+   * Callback to FocusManager to return ephemeral focus when search closes.
+   */
+  private returnEphemeralFocus: Blockly.ReturnEphemeralFocus | null = null;
 
   /**
    * Index of the currently "selected" block in the blocks array.
@@ -96,7 +103,10 @@ export class WorkspaceSearch implements Blockly.IPositionable {
     this.workspace.getComponentManager().addComponent({
       component: this,
       weight: 0,
-      capabilities: [Blockly.ComponentManager.Capability.POSITIONABLE],
+      capabilities: [
+        Blockly.ComponentManager.Capability.POSITIONABLE,
+        Blockly.ComponentManager.Capability.AUTOHIDEABLE,
+      ],
     });
     injectSearchCss();
     this.createDom();
@@ -111,6 +121,7 @@ export class WorkspaceSearch implements Blockly.IPositionable {
    * to prevent memory leaks.
    */
   dispose() {
+    this.releaseEphemeralFocus(false);
     for (const event of this.boundEvents) {
       Blockly.browserEvents.unbind(event);
     }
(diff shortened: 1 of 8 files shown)


packages/plugins/workspace-search/src/workspace_search.ts
```diff
@@ -473,25 +494,62 @@ export class WorkspaceSearch implements Blockly.IPositionable {
     this.setVisible(true);
     if (this.inputElement) {
       this.inputElement.value = this.searchText;
-      this.inputElement.focus();
+      const focusManager = Blockly.FocusManager.getFocusManager();
+      if (!this.returnEphemeralFocus && !focusManager.ephemeralFocusTaken()) {
+        this.returnEphemeralFocus = focusManager.takeEphemeralFocus(
+          this.inputElement,
+        );
+      } else {
```
maribethb: when would we hit this case?
- if the ephemeral focus callback is already set here (in which case it seems like something might have gone wrong if we're somehow re-focusing it without having destroyed that callback? should we just reset it in that case?)
- if ephemeral focus is already taken by something else (in which case might we be breaking something else by forcibly removing focus from it and potentially not letting the other thing have the chance to release ephemeral focus?)

not sure that either of my two analyses is actually correct so don't take my word for it, this just felt surprising to me

maribethb (approved):

packages/plugins/workspace-search/src/workspace_search.ts
mikeharv: Good question. The focus() call was in the original version, but we don’t need it once we’re taking ephemeral focus. Your point about ephemeral focus already being taken by something else is a good one. I think we’d realistically only hit this branch if Cmd+F was pressed while search is already open, and then we don’t need to do anything anyway. I’ve removed that branch and double-checked that everything still works.

PR #10392, opened 2026-08-31
test: Remove logspam when running test suite

gonfunko:
## The basics

### Proposed Changes
This PR avoids several thousand lines of errors and warnings in the devtools console when running the Mocha test suite. There are now less than a dozen warnings from when tests actually try to do naughty things.

These changes fall into several buckets:
* Only register the p5 blocks once
* Make the test workspace scrollable to avoid lots of whining from the bumpIntoBounds handler about events not having groups during teardown
* Check if unregistering is actually needed during test teardown
* Move up the shared event listener teardown to before workspace teardown

There are also a few core changes. While these mostly show up in a test context, they all could happen during actual use, so I think the changes are justified:

* The block paster focuses the pasted block after a delay, and now checks to make sure that it hasn't been disposed of in between pasting and that callback firing
* `FieldInput.resizeEditor_` is more defensive instead of force unwrapping the `WidgetDiv`
* The `WorkspaceAudio` manager is more defensive and only attempts to play audio when not suspended by the browser

### Reason for Changes
Less logspam makes debugging new tests easier and increases the likelihood that important log messages from the test suite are noticed and acted upon.

### DIFF
--- a/packages/blockly/tests/mocha/keyboard_navigation_test.js
+++ b/packages/blockly/tests/mocha/keyboard_navigation_test.js
@@ -139,7 +139,9 @@ suite('Keyboard navigation on Blocks', function () {
       toolbox: toolbox,
       renderer: 'zelos',
     });
-    Blockly.common.defineBlocks(p5blocks);
+    if (!('p5_setup' in Blockly.Blocks)) {
+      Blockly.common.defineBlocks(p5blocks);
+    }
     Blockly.serialization.workspaces.load(navigationTestBlocks, this.workspace);
     for (const block of this.workspace.getAllBlocks()) {
       block.initSvg();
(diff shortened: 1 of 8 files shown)


packages/blockly/tests/mocha/keyboard_navigation_test.js
```diff
@@ -139,7 +139,9 @@ suite('Keyboard navigation on Blocks', function () {
       toolbox: toolbox,
       renderer: 'zelos',
     });
-    Blockly.common.defineBlocks(p5blocks);
+    if (!('p5_setup' in Blockly.Blocks)) {
+      Blockly.common.defineBlocks(p5blocks);
```
maribethb: I think ideally these tests that register blocks should unregister them at the end of their tests

maribethb (approved):

packages/blockly/tests/mocha/keyboard_navigation_test.js
gonfunko: Done!

PR #10396, opened 2026-09-01
test: Reenable the drag-out-every-test-block browser tests

### LINKED ISSUE #10213: Reenable or move test block browser tests into devtools
There are two tests in Blockly's browser test suite (`npm run test:browser`) that exercise dragging out every block in the test blocks toolbox provided by the devtools plugin. We should either move those two tests ("opening every toolbox category in the test toolbox in LTR" and "opening every toolbox category in the test toolbox in RTL") into the devtools plugin, or reenable them after adjusting the infrastructure such that they can be run from core again.

gonfunko:
## The basics

### Resolves

Fixes #10213

### Proposed Changes
This PR reenables the two browser tests that exercise dragging out every block from the block-test plugin into the workspace. This change was LLM-assisted; the path for the block-test plugin is dynamically resolved from the monorepo, and the source gets executed inside the WebDriver browser session to define the blocks and toolbox. This avoids hardcoding paths and ensures that the latest local version of the test blocks is loaded and used.

I also updated the `npm run test:browser` command to build the plugin, rather than implicitly relying on that having been done.

### Reason for Changes
These tests have caught some bugs before, so it's useful to have them enabled.

### Test Coverage
Is better.


maribethb (approved):

packages/blockly/package.json
```diff
@@ -87,7 +87,7 @@
     "start": "npm run package && npx nx run @blockly/dev-tools:build && concurrently -n watch,devtools-server,core-server \"npx nx watch --projects=@blockly/dev-tools --includeDependencies -- npm run build\" \"npx nx run @blockly/dev-tools:start --no-open\" \"http-server ./ -p 8081 -s -o /tests/playground.html -c-1\"",
     "tsc": "gulp tsc",
     "test": "gulp test",
-    "test:browser": "npx mocha --config tests/browser/.mocharc.js",
+    "test:browser": "npx nx run @blockly/block-test:build && npx mocha --config tests/browser/.mocharc.js",
```
maribethb: I think there should be a way to have nx do this automatically, like we tell it that the `test:browser` target depends on the build for block-test. that way we still get caching and stuff and you don't replicate the nx config in the script definition. @lizschwab to research/advise please :)

Issue #10298, opened 2026-08-20
Allow deploying documentation without updating reference docs


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Problem

Right now deploying documentation always regenerates the reference docs and updates them as well. This is problematic if `main` contains changes that haven't been released yet, as those unreleased APIs will show up in reference docs. But at the same time, we need flexibility to update the guides in between releases.

### Request

Don't update the reference docs when deploying docs (unless we've just done a release / may need to manually specify)

Current flow:
1. api-extractor creates an api.json file (gitignored)
2. api-documenter uses that to create >4000 mdx files (gitignored)
3. docusaurus uses those + handwritten guides to create docs site
4. docs site hosted on cloudflare pages

I think that the way Cloudflare pages works is that you upload everything at once, there's no partial update mechanism where if you just don't change some files they stay the same. So I see two paths forward:

### Option 1: Commit api.json (preferred)

As part of the release process, commit the api.json file that api-extractor outputs (possibly with a few additional steps that modify the api.json file such as `removeRenames` in the docs_tasks). Only do this for `latest` tagged releases, not betas

This file becomes a committed history of our public API shape, which is something we've been interested in anyway since part of the cool thing about using api-extractor is being able to tell when a PR modifies the public API for example.

Then the docs deploy process can generate the reference docs using api-documenter from the committed api.json file instead of re-running api-extractor on the current codebase.

### Option 2: Checkout released version of blockly for the reference docs

As part of the release process, find the latest version tag, check that out, run `npm run docs` on that, and then use those generated reference docs instead of the ones from the current codebase

This would also be fine, but it lacks some of the advantages of the first one:
- no committed api.json file (which again is something we've discussed doing anyway)
- you'd be running the checked-out version of `npm run docs` on it so if you wanted to do something like fix a bug in the generated mdx or update the api-documenter patch, you'd have to wait until the next release. with option 1 you can use an updated api-documenter process but keep using the old api.json surface without doing a release

### option 3: commit all the generated reference docs (not recommended)

just including this here to document all the options, we could commit all the reference docs to the repo instead of generating them at documentation deploy time. but that would also still need to be done only at release time to avoid the original problem of documenting unreleased APIs, so it is just a more cumbersome version of option 1 that requires us to commit a bunch of files that may change just for cosmetic reasons in between releases.

Issue #7244, opened 2023-07-02
Method reference docs are missing descriptions


BeksOmega:
### Location

e.g. https://developers.google.com/blockly/reference/js/blockly.icons_namespace.commenticon_class.md

### Type

- [ ] Text
- [ ] Image or Gif
- [X] Other

### Suggested content

The reference docs only put method descriptions if the direct JSDoc has a description. This means that we are missing descriptions if:
  1) We are overriding a function - the description should be inheritted but is not.
  2) We only have an `@return` tag with no description - the `@return` tag should be used as the description.

### Additional context

There are several solutions to this problem:

1) Patch API documenter.
    We already [patch API documenter](https://github.com/google/blockly/blob/68eb38889a2e8e97fda2fa61efb80efb26b9030b/patches/%40microsoft%2Bapi-documenter%2B7.22.4.patch) so we could also patch it to fix this issue.

2) Fork API documenter.
    They state in their [dev docs](https://api-extractor.com/pages/setup/custom_docs/#forking-the-api-documenter-project) that "The api-documenter project was designed to serve as a code sample", so continually patching sample code feels like not the happy path. So it might also be worth it to just create our own documenter, so we don't have to keep [maintaining patches](https://github.com/google/blockly/pull/7075).

3) Use someone else's API documenter, such as  [firebase's](https://www.npmjs.com/package/@firebase/api-documenter).

4) Change the lint config to require descriptions.

maribethb:
Let's talk about this before you do any work on it.

PR #10398, opened 2026-09-02
feat: generate API docs with TypeDoc

### LINKED ISSUE #10110: Add source code links to API docs
### Location

_No response_

### Type

- [ ] Text
- [ ] Image or Gif
- [ ] Other

### Suggested content

API docs do not currently link back to source code. 

It would potentially improve the usefulness of the API docs if links back to source code were added (at least at the class level) so that there's easy access to the specific code being referenced. API extractor already generates a `fileUrlPath` in the JSON that it initially creates when making the docs, so we would just need to use it. 

### Additional context

_No response_

### LINKED ISSUE #10303: stop using gulp to do string manipulation in docs tasks
### Check for duplicates

### Description

docs_tasks are hardly gulp tasks, only 2 things use streams and those can be replaced. 

most of the functions here are doing string manipulation to rename things, get the titles out of links, etc. ideally those functions would have their own unit tests so we don't break the docs one day, but right now it would be a nightmare to unit test because they're all wrapped up in gulp.

just convert this to a normal node script.

- `convertToMdx` is at least the gulpiest function we have, but it can still be done with just regex/replace and then changing the file suffix in the output path.
- `removeRenames` is removing the string `_2` from the api.json file. this can also just be a function.
- pretty much everything else can just be a function that takes in strings and outputs strings. easier to unit test.

right now, we are doing 4 separate passes over each of 4000 files. so each file has to be opened, modified, rewritten 4 times. but we can collapse that into 1 pass per file to do all the transforms at once instead of split up like this.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

### LINKED ISSUE #10321: Switch to TypeDoc for generating API docs
### Check for duplicates

### Request

I've been looking into switching to [TypeDoc](https://typedoc.org/) after @maribethb [mentioned it](https://github.com/RaspberryPiFoundation/blockly/issues/10110#issuecomment-4906133603) and I think it would help resolve a number of API documentation issues. 

Pros: 
1. We currently have to patch API documenter, which is not ideal for maintenance reasons.
1. We have to do a lot of file/string manipulation to API documenter's output. A lot of this is API documenter specific and would go away. There's also a [TypeDoc plugin for Docusaurus](https://typedoc-plugin-markdown.org/plugins/docusaurus) that we could use.
1. #7244 would be partially resolved by switching to TypeDoc, as it is able to inherit descriptions. 
1. #10110 would be resolved by switching to TypeDoc, since it includes links automatically. 
1. #10298 would not be any harder with TypeDoc, and might actually be better. You can still follow Option 1, because TypeDoc supports JSON generation and you can generate the docs from there. TypeDoc's links point to a specific commit, so if you had changes to a method that weren't released, it would automatically point to whatever commit it was when you generated the JSON.

Cons: 
1. It's hard to get the format of the API docs to be the same as they were with API documenter. Either we a) write/maintain code to get everything to look as similar as possible or b) adopt a different format for the API docs, which would introduce a new learning curve for developers who are used to our API docs as is. 
1. TypeDoc has different naming conventions compared to API documenter, so all of our API reference links would be different. In our own guides, we can convert the links, but we would need to consider developers who might have API docs bookmarked. We could:
a) add a script to convert the TypeDoc names into the API documenter format
b) provide redirects, but we'd need an aggressive number of them
c) break everybody's API links
1. I've only just started exploring this so there could be other cons 

If we were to convert, we ideally could do it before putting work into the open docs pipeline issues (#10303 and #10298) but that would mean putting those issues on hold for a little while.

### Alternatives considered

_No response_

### Additional context

_No response_

zspriggs:
## The basics

### Resolves

Fixes #10321
Fixes #10110 since TypeDoc automatically includes source links.
Fixes #10303 by removing gulp docs tasks entirely, since they are no longer needed.

### Proposed Changes

Adds TypeDoc: 
- Add `typedoc.json` with config options
- Add TypeDoc plugin to `docusaurus.config.js`
- Add TypeDoc and relevant plugins to `package.json` in the `docs/` folder
- Add `packages/docs/typedoc-member-index.mjs` which adds indexes for methods and properties & a custom router:
  - `MarkdownThemeContext` extension: The TypeDoc Docusaurus plugins allow for an index to be generated only if members have their own pages. After team discussion, it was decided that it would be nice to have a top-level index, but also have the full descriptions on the same page. The `typedoc-member-index.mjs` makes the those index sections (indexed sections are properties and methods) and inserts them into the page body. 
  - `MemberRouter` extension: There's a weird quirk with the way that the markdown plugin handles namespaces specifically, where it prepends /blockly to the URL. For example, the URL for a class is something like `/reference/classes/Block/` whereas the URL for a namespace ends up being `/reference/blockly/namespaces/Block/`. Since we are changing the URL structure, I wanted the new URLs to be as intuitive to figure out as possible, so I also added a custom router to `typedoc-member-index.mjs` that adjusts the handling of namespaces so that the extra /blockly is removed. 
  - This code was LLM assisted

Removes api-documenter/api-extractor:
- Remove `api-documenter` `api-extractor` `patch-package` and `postinstall`. The latter two are no longer needed because the `api-documenter` patch was the only one we had. 
- Remove `patches` folder bc again, the `api-documenter` patch was our only one
- Remove `docs_tasks.mjs` entirely, because this was all post-processing of API documenter's output, which is no longer needed. Similarly removed `api-extractor.json`.

One of the key differences between TypeDoc and API documenter is that TypeDoc hooks into Docusaurus and generates the reference docs automatically whenever Docusaurus loads. This means: 
- You no longer need to `npm run docs` in `blockly` in order to run the docs site.
- The docs are no longer dependent at all on the build output of `blockly`. TypeDoc just looks at the source `.ts` files and generates from there (whereas API documenter was looking at the compiled `.d.ts` files). Docs `package.json` and `deploy_docusaurus.yml` have been adjusted accordingly. 

### Reason for Changes

As described in #10321, api-documenter/api-extractor do not work super well for our use case. We have to patch api-documenter and maintain lots of post-processing to get the docs functioning. Switching to TypeDoc removes much of this overhead, as well as giving us some useful features for free (like source code links, easier Ctrl+f page searching, and more detailed property/method descriptions).

### Test Coverage

If you want to see the format of the new reference docs, you can pull down this branch and run `npm install` to get the TypeDoc install and then `npm run start` in the `packages/docs` folder to run the site. 

### Documentation

Included an update to docs README.

The format of the API docs links has changed, which affects our guides. In the interest of not making a massive diff, I will fix those links in a separate PR. In the meantime, docs should *not* be deployed until after the link fix is also merged, because the deployment of just this PR would break all the links.

### Additional Information

Follow-up work that is affected by this PR: 
#10298 
#7244 is not resolved by this conversion, but should be made a bit easier to do bc we won't have to add another patch to api-documenter


packages/docs/package.json
```diff
@@ -5,41 +5,20 @@
...
-            "projects": "blockly",
-            "target": "package"
-          },
-          {
-            "projects": "blockly",
-            "target": "docs"
-          }
-        ]
-      },
+      "start": {},
```
lizschwab:
```suggestion
```

lizschwab: We don't actually need to define targets here if we're not telling nx to do anything special with them.

packages/docs/package.json
```diff
@@ -5,41 +5,20 @@
...
       "build": {
+        "dependsOn": [],
```
lizschwab:
```suggestion
```
Same as above; we can just leave this out if it's empty.

lizschwab (approved):
I added some suggestions for minor changes, but this looks good.

PR #10404, opened 2026-09-03
fix: align Home/End/Page Down shortcuts with arrow-key navigation

### LINKED ISSUE #10186: Additional shortcuts via registerNavigationShortcuts (Page Up / Down & Home / End)
### Check for duplicates

### Description

We have been catching up with the behaviour of Page Up / Down and Home / End as we didn't ship this in MakeCode. I think we have found a couple of bugs, but also have thoughts on the behaviour in some areas as well.

### Bugs

When a container end statement position is focused, pressing End moves focus into the previous block's end statement position. In this case, I'd expect End to do nothing at all.

https://github.com/user-attachments/assets/55fc8e03-f118-4d11-9c5e-aa869caa5d04

When a container block is focused, pressing End moves focus to the container end statement position, but this position is in the wrong place. It focuses an impossible position between the top inside edge of the container and the top of the statement block inside. This is normally unreachable with Up / Down.

https://github.com/user-attachments/assets/cfdc57da-68ba-4311-a75d-0bc952140469

### Thoughts on behaviour

Sometimes it feels logical, and sometimes it feels unpredictable. We think it would make more sense if it followed conventions used for [keyboard controls in a grid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/grid_role#keyboard_interactions), or put another way, would move focus to where the arrow keys normally would if they were held down while scoped to the current stack. Details below:

- Page Up - moves to the first block in the stack accessible by using the Up arrow. Effectively what pressing the Up arrow multiple times would do, except scoped to the currently focused stack. No behaviour change.
- Page Down - moves to the last block in the stack accessible by using the Down arrow. Effectively what pressing the Down arrow multiple times would do, except scoped to the currently focused stack. Behaviour change - Page Down no longer traverses / moves focus right into block inputs.
- Home - moves to the left-most focusable position in the block accessible by using the Left arrow which is (almost?) always the block itself.
- End - moves to the right-most focusable position in the block accessible by using the Right arrow. Lots of behaviour change. E.g., When focused on a statement & container block, End would no longer focus the end statement position.
- Ctrl/Cmd + Home - moves to the first focusable position on the workspace (excluding the workspace itself). No behaviour change.
- Ctrl/Cmd + End - moves to the last focusable position on the workspace. Behaviour change - this doesn't just move to the last block, but moves to the last input. In a non-looping world, this would mimic what holding the Down arrow would do until you hear the end of program beep, followed by holding the Right arrow until you hear the end of program beep. 

If the above is applied, then Home / End always work, even if you are currently focused on any input or field in a block. Let's say the second input of 3. End moves to the third input, home moves to the parent block.

These shortcuts should respect ltr and rtl layouts.

### Reproduction steps

N/A - see bug section of the above.
Use the [Blockly live demo](https://raspberrypifoundation.github.io/blockly/packages/blockly/demos/code/index.html) and paste `Blockly.ShortcutItems.registerNavigationShortcuts()` in the devtools console to turn on these shortcuts.

### Priority

Work effort: likely significant, but depends on the changes ultimately made.
Impact: low in a way as this feature is opt-in, but potentially has lots of added value to applications that make use of it. We think that some changes are required before MakeCode can opt-in to these additional navigation shortcuts.

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

anishtsudo:
## The basics

### Resolves

Fixes https://github.com/RaspberryPiFoundation/blockly/issues/10186

### Proposed Changes

Update the opt-in `registerNavigationShortcuts()` so Home, End, Page Down, and Ctrl/Cmd+End follow the same destinations as holding arrow keys, scoped to the current block, stack, or workspace.

- Home moves to the first in-block node (usually the block itself).
- End moves to the last in-block node on the current row, not the statement end.
- Page Down moves to the last Down-reachable node in the stack and no longer enters inline value inputs.
- Ctrl/Cmd+End moves to the last focusable workspace node (last stack, then Down, then In).
- Page Up and Ctrl/Cmd+Home are unchanged.

This also stops End on a container end-statement from moving focus, and stops End on a focused container block from landing on an unreachable statement connection.

### Reason for Changes

MakeCode needs these shortcuts before opting in. End previously focused the last input connection, which was wrong for C-shaped blocks.

### Test Coverage

Added/updated mocha tests in `shortcut_items_test.js` and `navigation_test.js` covering both End bugs, Page Down not entering inline inputs, and Ctrl+End targeting the last focusable node.

Mocha (Chrome): Jump shortcuts + navigator helpers 26 passed; NavigationFunctions 73 passed.

### Documentation

Updated the optional-shortcuts table in `packages/docs/docs/guides/configure/keyboard-nav.mdx`.

### Additional Information

These shortcuts remain opt-in via `Blockly.ShortcutItems.registerNavigationShortcuts()`. They use the same In/Out/Down walks as arrow keys, so LTR/RTL follow existing navigator direction handling.

### DIFF
--- a/packages/blockly/core/shortcut_items.ts
+++ b/packages/blockly/core/shortcut_items.ts
@@ -35,9 +35,9 @@ import {type IFocusableNode} from './interfaces/i_focusable_node.js';
 import {isSelectable} from './interfaces/i_selectable.js';
 import type {IToolbox} from './interfaces/i_toolbox.js';
 import {Direction, KeyboardMover} from './keyboard_nav/keyboard_mover.js';
+import type {Navigator} from './keyboard_nav/navigators/navigator.js';
 import {keyboardNavigationController} from './keyboard_navigation_controller.js';
 import {Msg} from './msg.js';
-import {RenderedConnection} from './rendered_connection.js';
 import {KeyboardShortcut, ShortcutRegistry} from './shortcut_registry.js';
 import * as Tooltip from './tooltip.js';
 import {aria} from './utils.js';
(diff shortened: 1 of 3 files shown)


mikeharv (changes requested):
Thanks for picking this up! We will need a few changes before this can merge. None of them are a rewrite of the destination rules, which look good.

1. Navigator API:
We try to keep new Navigator methods `private` or `@internal` unless we know an app needs to call them. These helpers are only used from `shortcut_items.ts`, so making them public would lock us into supporting them as API. Several of the new methods are the same while-loop with a different step and stop condition. Could they be combined to a single private method? Some shortcut logic has been moved out of `shortcut_items` that should probably move back. `Navigator` methods shouldn't need to be aware of the scoping of Home/End shortcuts.

2. Tests:
Could you assert the actual destination node rather than `getLastNodeInBlock()` / `getLastNodeInStack()` as the expected value? Otherwise the tests will keep passing if the helper is wrong, as long as the shortcut still calls it. The older tests in this file are a good model. 

3. Rebase (due to merge conflicts)
Note that this will change some of the tests you are modifying as we've stopped using `getFocusedNodeStub` in favor of the real `focusNode()` (see https://github.com/RaspberryPiFoundation/blockly/pull/10388).

Issue #10305, opened 2026-08-20
stop using gulp for closure compiler tasks


maribethb:
### Check for duplicates

- [ ] I have searched for similar issues before opening a new one.

### Description

We are using gulp to invoke closure-compiler and do some other tasks related to building, like renaming some files. However, this has a few problems:

1. gulp is barely maintained
2. our gulp scripts are difficult to reason about/modify
3. nx may or may not actually be cacheing the build output correctly or efficiently
4. the google-closure-compiler gulp plugin we're using is updated frequently but doesn't seem to be officially supported by google, and nobody can answer questions about it apparently. it also breaks every ~month or so requiring us to update to the newest version which has no appreciable new features or value to us other than unborking CI. unfortunately this may or may not be solved by this move because it's the same npm package we'd be using, but we'd switch to the plain node implementation instead of the gulp plugin.
We can remove gulp from the build_tasks with the following replacements:

- a bunch of functions that are already async or just doing some work and calling gulp's `done` at the end: just regular functions, stay the same / drop the `done` callback
- `stripApacheLicense`: currently using gulp-replace stream that modifies the files in-stream so this intermediate step isn't written anywhere. we can modify the tsc output in place instead (and write to disk)
    - note that we currently call this on the advanced_compilation_test's source files, but we don't need to because who cares if those test files still have the apache header instead of 4 newlines? so it's okay for this to move out of the current place and just act on the tsc outputs instead of everything that's passed to cc.
- `compile` uses the google-closure-compiler gulp plugin. instead, we can just use the native node API. we may need to add our own error handling to make a closure compiler exit the script with error, instead of relying on the gulp plugin to cause a stream error.
- gulp-sourcemaps plugin: remove and just pass the sourcemap flags to closure compiler directly.
- `buildCompiled` is the complicated step. it currently has 6 stages:
    1. `stripApacheLicense`: discussed above, rewrite in place
    2. `sourcemaps.init()`: discussed above, just drop this
    3. `compile`: discussed above
    4. `rename` to add `_compressed` to filenames: just rename the files on disk afterwards, then update the names in the sourcemaps after they're generated
    5. `sourcemaps.write`: discussed above, handled by passing the `--create-source-map` flag to cc. since this would be generated in step 3 now, need to handle renaming the files to compressed in the sourcemaps as well.
    6. `gulp.dest` is where all of this is finally written to disk, instead we'd be doing it as we go 
[...]

PR #10415, opened 2026-09-09
chore: stop using gulp as a test task runner

### LINKED ISSUE #10306: stop using gulp as a test task runner
### Check for duplicates

### Description

We're using gulp as effectively a task runner in `npm run test` which is not a thing gulp is particularly good at doing. It has caused numerous problems repeatedly:

- gulpfiles are difficult to maintain/modify/reason about
- we've been bit by multiple errors that get swallowed and are a pain to debug because the actual issue is swallowed by gulp somewhere, leading to layers of additional logging that are just adding more potential spots for confusion and failure
- results are not cacheable with nx due to the way we've set things up so currently the blockly tests have to run after all the plugin tests
- `lint` has special handling to make it not run twice in CI

gulp is not a task runner. we basically rolled our own so that we could do things like have all the tests run and not fail until the end. however, now we're actually using a task runner, nx, which can do a better job for us.

steps:

1. make each task its own named npm script
2. accurately declare the nx graph to name which tasks depend on others, their inputs and outputs so they can be cached correctly, etc. needs care to get right to avoid cache problems.
3. run everything with `npx nx run-many` to run all the commands at once. use `--nx-bail=false` to have it keep running even after something fails.

things to be aware of:
- if we have cacheing enabled in ci, make sure the cache is actually invalidated in between running on node versions 22 and 24
- nx doesn't have the cute github step summary we hand-made, so either get used to nx's output or re-write this or see if someone has a plugin for this
- using colons in the script names forces us to use the project/target flags instead of nx's default `project:target` syntax so it might be better to just rename the scripts while we're here.

### Reproduction steps

1. 
2. 
3.

### Priority

_No response_

### Stack trace

```shell

```

### Screenshots

_No response_

### Browsers

_No response_

lizschwab:
## The basics

- [x] I [validated my changes](https://docs.blockly.com/guides/contribute/core/#make-and-verify-a-change)
- [x] This needs https://github.com/RaspberryPiFoundation/blockly/pull/10406 to merge first before it can be merged

### Resolves

Fixes #10306 

### Proposed Changes

Removes the gulp test task file and uses nx instead to orchestrate which tests are run in which order. A couple of the test tasks live in the gulp build tasks file, which will be addressed with #10305, so those were not updated.

### Reason for Changes

Using gulp for this was causing issues where nx was unable to cache test results and had to run blockly tests separately from other tests. These can now be run in parallel.


packages/blockly/tests/scripts/test_helpers.mjs
```diff
@@ -0,0 +1,75 @@
...
+  });
+}
+
+/**
+ * Helper method for running test command.
+ * @param {id} id The test command id/name.
+ * @param {string} command Command line to run.
+ * @return {Promise} Asynchronous result.
+ */
+export async function runTestCommand(id, command) {
```
gonfunko: Is this used by anything?

packages/blockly/tests/generators/test_generators.mjs
gonfunko, line 1: Now that this is just being run by node, do any of the methods or constants in it need to be exported?

packages/blockly/scripts/gulpfiles/test_tasks.mjs
```diff
@@ -1,305 +0,0 @@
...
- * Copyright 2022 Google LLC
- * SPDX-License-Identifier: Apache-2.0
- */
-
-/**
- * @fileoverview Gulp tasks to test.
- */
-/* eslint-env node */
-
-import asyncDone from 'async-done';
```
gonfunko, line 12: I think this was the only use of async-done, we may be able to drop that dep?

gonfunko: Oh wait nevermind it moved to the new file, disregard!

packages/blockly/tests/scripts/test_helpers.mjs
lizschwab: I was using it but removed it in favor of just running the tests that were using it directly in node. I'll remove it.

packages/blockly/tests/generators/test_generators.mjs
lizschwab: Good catch, I don't think they need to be exported.

packages/blockly/scripts/gulpfiles/test_tasks.mjs
lizschwab: We actually don't need it; I realized the tasks I had using it weren't adding anything of value now that we don't have the pretty summary.

gonfunko (approved):

